PR #4132, opened 2021-06-28
Fix documentation for SSD and RetinaNet

### LINKED ISSUE #4106: Documentation confusing on whether SSD and RetinaNet count background as class object
The documentation for the `SSD` class mentions that we should not count the background as an object class when passing the number of classes as a parameter to instantiate an SSD object.

https://github.com/pytorch/vision/blob/959666891589c35e6d225943253f523cffbae4cc/torchvision/models/detection/ssd.py#L144

However, further down in the same file, an SSD object is instantiated in a function that explicitly says that the background **should** be counted as an object class, but this is not taken into account in the code (i.e. I did not see `num_classes` be decremented by one when creating the SSD object).

https://github.com/pytorch/vision/blob/959666891589c35e6d225943253f523cffbae4cc/torchvision/models/detection/ssd.py#L589

Here is the documentation for this function, which says we should include the background in the number of classes.

https://github.com/pytorch/vision/blob/959666891589c35e6d225943253f523cffbae4cc/torchvision/models/detection/ssd.py#L563

This is confusing. Should we or should we not count the background as an object class when instantiating the SSD? In either case, how should object classes be ID'd during training?

As an example, with Faster RCNN, the background is counted as an object class (with ID 0 reserved for it) and actual object classes are identified during training starting from ID 1. What should be the procedure for SSD?

I have also opened [a topic](https://discuss.pytorch.org/t/do-we-have-to-count-the-background-as-an-object-class-for-ssd/124927) in the forums, since this is both a personal question of mine as well as a possible issue in the docs (or the code).

datumbox:
Fixes #4106

### DIFF
--- a/torchvision/models/detection/retinanet.py
+++ b/torchvision/models/detection/retinanet.py
@@ -256,7 +256,7 @@ class RetinaNet(nn.Module):
             It should contain an out_channels attribute, which indicates the number of output
             channels that each feature map has (and it should be the same for all feature maps).
             The backbone should return a single Tensor or an OrderedDict[Tensor].
-        num_classes (int): number of output classes of the model (excluding the background).
+        num_classes (int): number of output classes of the model (including the background).
         min_size (int): minimum size of the image to be rescaled before feeding it to the backbone
         max_size (int): maximum size of the image to be rescaled before feeding it to the backbone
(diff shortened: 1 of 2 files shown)


NicolasHug (approved):
thanks!

PR #4133, opened 2021-06-28
README: update torch/torchvision support table

adamjstewart:
I'm happy to keep updating this table, but maybe we should add something to the release instructions that remind people to update this table.


fmassa (approved):
Thanks for the PR!

Yes, we should keep updating the table in our release runbook.

cc @NicolasHug for visibility

Issue #4136, opened 2021-06-30
GoogleNet is secretly transforming input


celidos:
Hello!

I recently noticed that I might be doing image normalization twice in my experiments.
The [documentation](https://pytorch.org/vision/0.8/models.html#torchvision.models.googlenet) says that the default value of the `transform_input` parameter is False.

So when calling 
```python3
model = torchvision.models.googlenet(pretrained=True)
```

I would probably expect the model not to do any input transformations, but accidentally it does ([permalink](https://github.com/pytorch/vision/blob/a83b9a17e441e6d77e9d59ca14d75fb5ba8c31f6/torchvision/models/googlenet.py#L41)) until you directly specify `transform_input=False`. So in case of `pretrained=True` and not-specified `transform_input` model suddenly sets its value to True:

```python3
    if pretrained:
        if 'transform_input' not in kwargs:
            kwargs['transform_input'] = True
```

It is confusing for me. This thing is only happens in GoogleNet.

NicolasHug:
Thanks for the report @celidos , the documentation is wrong, is should indicate that the default is True.

#4137 should fix this.


@fmassa, do you remember why the `transform_input` and `aux_logits` parameters are passed as kwargs? It looks like we could just add them  as regular parameters (we can make them keyword-only if we want to) ?

I feel llike we should avoid kwargs unless we really need to, as they obfuscate the documentation, like here. Also, in the googlenet code we're modifying the kwargs dictionary inplace, and as a user I would find this fairly unexpected.

fmassa:
Hi,

The situation is a bit complicated, and I think we should improve the documentation indeed.

The problem is that the pre-trained weights from Inception and GoogleNet were converted from TF, which have a different input normalization.

In order to make the models compatible with the rest of torchvision, we added this `transform_input` argument.

This argument can be seen as an internal implementation detail, which gets enabled if you load the default pre-trained weights that we provide (which were converted from the original implementation in TF).

So if you are training your model from scratch and you are using the default imagenet mean / std values, then you don't need to change anything.

celidos:
fix 

https://github.com/pytorch/vision/pull/4137

can possibly create mirror problem: if `pretrained=False`, you would expect for model to transform input by default, but GoogleNet class has default parameter `transform_input=False`, therefore it will not perform transformation.

How can this be brought to a common style and not cause misunderstandings?

fmassa:
[...]

PR #4137, opened 2021-06-30
Fix googlenet and inception docstring for transform_input parameter

NicolasHug:
Fixes https://github.com/pytorch/vision/issues/4136

Default isn't always False, it depends on `pretrained`


celidos:
comment on https://github.com/pytorch/vision/issues/4136

NicolasHug:
You're right, thanks, I updated the comment to be ````Default: True if ``pretrained=True``, else False````

datumbox (approved):
LGTM, thanks!

NicolasHug:
Sorry, I messed up the confict merge - I'll merge https://github.com/pytorch/vision/pull/5089 with the same changes instead instead

Issue #2950, opened 2020-11-03
In transforms.Resize, tensor interpolate is not the same as PIL resize.


hjinlee88:
## 🐛 Bug

Resize supports tensors by F.interpolate, but the behavior is not the same as Pillow resize.
https://github.com/pytorch/vision/blob/f95b0533243dfbc901b5ed5f5db28a5a46bdb699/torchvision/transforms/functional.py#L309-L312

## To Reproduce

```python
import urllib
from PIL import Image
from torchvision import transforms
from matplotlib import pyplot as plt

size = 112
img = Image.open(urllib.request.urlopen("https://pytorch.org/tutorials/_static/img/tv_tutorial/tv_image01.png"))

tensor_interpolate = transforms.Compose([transforms.ToTensor(), transforms.Resize(size), transforms.ToPILImage()])
pillow_resize = transforms.Compose([transforms.Resize(size)])

plt.subplot(311)
plt.imshow(img)
plt.title("original")
plt.subplot(312)
plt.imshow(tensor_interpolate(img))
plt.title("tensor interpolate")
plt.subplot(313)
plt.imshow(pillow_resize(img))
plt.title("pillow resize")
plt.show()
```
![image](https://user-images.githubusercontent.com/30526247/97950992-f6c3c300-1ddb-11eb-85d4-a511d58e3b50.png)

## Expected behavior

Both should have the same or nearly identical output.
Perhaps, it needs blur before interpolate.

## Environment
I installed pytorch using the following command:
conda install pytorch torchvision -c pytorch

python collect_env.py
Collecting environment information...
PyTorch version: 1.7.0
Is debug build: True
CUDA used to build PyTorch: 11.0
ROCM used to build PyTorch: N/A

OS: Microsoft Windows 10 Home
GCC version: (MinGW.org GCC-8.2.0-3) 8.2.0
Clang version: Could not collect
CMake version: version 3.18.2

Python version: 3.8 (64-bit runtime)
Is CUDA available: True
CUDA runtime version: 10.0.130
GPU models and configuration: GPU 0: GeForce RTX 2060
Nvidia driver version: 456.38
cuDNN version: C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v10.0\bin\cudnn64_7.dll
HIP runtime version: N/A
MIOpen runtime version: N/A

Versions of relevant libraries:
[pip3] numpy==1.19.2
[pip3] torch==1.7.0
[pip3] torchvision==0.8.1
[...]

PR #4138, opened 2021-06-30
Minor additions to Resize docs

NicolasHug:
Just some minor edits to the docs which will hopefully explicitly clarify that using `antialias` will lead to closer outputs for PIL and tensors. This is motivated by https://github.com/pytorch/vision/issues/2950#issuecomment-871189566

CC @vfdev-5


vfdev-5 (approved):
Thanks @NicolasHug , looks good to me.

PR #4139, opened 2021-06-30
NOMRG Estimate time gain of not running CPU tests on GPU machines

NicolasHug:


NicolasHug:
"Run test" part of the CI jobs:

For linux, 16min -> 8min = 8min gain
For Windows, 21min  -> 10min = 11min gain

**In both cases this amounts for ~50% of the "Run test" parts, and about ~30% of the whole CI job (including spin up, installations, etc)**

The unittest_windwos_gpu used to be the slowest test job taking about 10min longer than all the other ones:
image

Now, we've got those 10 minutes back. **So potentially this saves 10 mins of developer time on every single commit.**

image

PR #4141, opened 2021-07-01
Support float in `torchvision.io.video.read_video`

bryant1410:
`start_pts` can be `float` or `int`, depending on the value of `pts_unit`. [`float` typing considers both](https://stackoverflow.com/a/50928627/1165181).

### DIFF
--- a/torchvision/io/video.py
+++ b/torchvision/io/video.py
@@ -236,7 +236,7 @@ def _align_audio_frames(
 
 
 def read_video(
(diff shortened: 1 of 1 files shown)


NicolasHug:
Thanks for the PR @bryant1410 

Looks like start_tps can also be a `Fraction`, according to the docstring. And it looks like we could update end_tps too.

bryant1410:
Good catch!

bryant1410:
The failing tests are unrelated to this change, right?

NicolasHug (approved):
LGTM, thanks @bryant1410 !
Yes, the failures are unrelated

PR #4142, opened 2021-07-01
update mobilenetv3 for better performance

Sharpiless:


NicolasHug:
Hi @Sharpiless , thanks for the PR. Could you please describe your intended change here? What kind of performance gain do you observe?
I'm not sure we can just replace dropout with a batchnorm unfortunately, as clearly this would lead to different predictions.

Sharpiless:
> Hi @Sharpiless , thanks for the PR. Could you please describe your intended change here? What kind of performance gain do you observe?
> I'm not sure we can just replace dropout with a batchnorm unfortunately, as clearly this would lead to different predictions.

I find there's a performance gap between mobilenetv3 in vision and this repo : https://github.com/xiaolai-sqlai/mobilenetv3
The main difference is the final classifier，and I tested the later one on five datasets：

<html xmlns:v="urn:schemas-microsoft-com:vml" xmlns:o="urn:schemas-microsoft-com:office:office" xmlns:x="urn:schemas-microsoft-com:office:excel" xmlns="http://www.w3.org/TR/REC-html40">
<head>

<meta name=Generator content="Microsoft Excel">

<style>

</style>
</head>
<body>



dataset | torchvision | improved
-- | -- | --
DTD | 70.3 | 71.54255
256 | 85.35 | 84.78513
bird | 80.54 | 81.66164
car | 78.98 | 80.98318
dog | 82.66 | 83.82924



</body>

</html>

NicolasHug:
Thanks for the details,

FWIW, the original MobilenetV3 paper https://arxiv.org/pdf/1905.02244.pdf does mention a dropout of 0.2-0.8. So while https://github.com/xiaolai-sqlai/mobilenetv3 may lead to slightly better results, it might not be as faithful to the original implementation.

Perhaps this is something that @datumbox can comment on once he's back

datumbox:
@Sharpiless Thanks for the PR.

As Nicholas said, TorchVision's implementation follows closely the paper and the original implementation. Also please note that even if modifying the network gives you better results on some small datasets, the MobileNetV3 architecture is optimised for ImageNet.

My advice is to close the PR and let those users who want to introduce additional layers or remove dropout from the classifier to do it on their side.

PR #4058, opened 2021-06-15
enable windows cuda test on vision

mszhanyi:
Fix #4006

1. The major root cause is the environment variables not set correctly for newly installed cuda.
2. nvjpeg and nvjpeg_dev are not in the package list. It is indispensable for vision
3. VS integration directory is different for cuda 10 and cuda 11
4. Job executor name is different from pytorch repo.

PS. the install_cuda_compatible_cmath is out of date. remove them all.

PR #4145, opened 2021-07-01
Enable Window CUDA Test

mszhanyi:
1. Per Nikita's suggert, split  install_cuda_compatible_cmath from #4058. 
2. Because it'd better to check nvjpeg before building, so I don't plan to split nvjpeg change to another PR
```
   if [ ! -f "${CUDA_PATH}\include\nvjpeg.h" ]
    then
        echo "nvjpeg does not exist"
        exit 1
    fi
```


mszhanyi:
@NicolasHug it looks recent pytorch change trigger a regression error.
https://app.circleci.com/pipelines/github/pytorch/vision/9376/workflows/2b47b22b-53f3-4d1f-be8e-5693b0861a74/jobs/689087

I install the release pytorch version, the build can pass.
https://github.com/mszhanyi/vision/blob/5a298742f9b03b658be211c47425d9262aae58d3/packaging/build_cmake.sh#L31-L32

NicolasHug:
@mszhanyi Indeed, this was reported in https://github.com/pytorch/pytorch/issues/61036 and hopefully should be fixed soon!

malfet:
This PR seems to break number of existing build configs, for example [cmake_linux_gpu](https://app.circleci.com/pipelines/github/pytorch/vision/10274/workflows/26212e6f-f41c-4179-a923-fe1d279defec/jobs/777300) now fails, because CUDA could not be find in "C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v10.2" folder

packaging/build_cmake.sh
```diff
@@ -45,6 +45,8 @@ else
     yum install -y libpng-devel libjpeg-turbo-devel
 fi
 
+source .circleci/unittest/windows/scripts/set_cuda_envs.sh
+
```
malfet, line 49: This would be executed from both Windows and Linux build environment, which seems wrong

malfet (changes requested):
Please fix build failures and then request re-review.
Also, I would strongly recommend to separate updating CUDA versions and enabling the builds into a separate PRs.
For example, PR enabling Windows CUDA testing should not change packaging scripts, as as packaging TorchVision with CUDA currently works correctly, isn't it?

Issue #4146, opened 2021-07-01
Error with PIL 8.3.0: __array__() takes 1 positional argument but 2 were given


slettner:
## 🐛 Bug

The `torchvision.transformer` function `ToTensor` raise a `TypeError` when receiving a image loaded with `PIL`.
The error origins from a call to `np.array` and contains the message: 
`__array__() takes 1 positional argument but 2 were given`

## To Reproduce

```
import io
import requests
import torchvision.transforms as T
from PIL import Image

resp = requests.get('https://picsum.photos/200')
img = Image.open(io.BytesIO(resp.content))

preprocess1 = T.Compose([
   T.ToTensor(),
])

x = preprocess1(img)
print(x.shape)
```

Pip Freeze Log:
```
certifi==2021.5.30
chardet==4.0.0
idna==2.10
numpy==1.21.0
Pillow==8.3.0
requests==2.25.1
torch==1.9.0
torchvision==0.10.0
typing-extensions==3.10.0.0
urllib3==1.26.6
```

Full Traceback:
```
Traceback (most recent call last):
  File "/Users/sebastianlettner/Desktop/tmp/test.py", line 13, in <module>
    x = preprocess1(img)
  File "/Users/sebastianlettner/Desktop/tmp/.venv/lib/python3.8/site-packages/torchvision/transforms/transforms.py", line 60, in __call__
    img = t(img)
  File "/Users/sebastianlettner/Desktop/tmp/.venv/lib/python3.8/site-packages/torchvision/transforms/transforms.py", line 97, in __call__
    return F.to_tensor(pic)
  File "/Users/sebastianlettner/Desktop/tmp/.venv/lib/python3.8/site-packages/torchvision/transforms/functional.py", line 129, in to_tensor
    np.array(pic, mode_to_nptype.get(pic.mode, np.uint8), copy=True)
TypeError: __array__() takes 1 positional argument but 2 were given
```

## Expected behavior

I expect that the conversion works and I reveive a torch.Tensor with shape (200, 200, 3)

## Environment
```
Collecting environment information...
PyTorch version: 1.9.0
Is debug build: False
CUDA used to build PyTorch: None
ROCM used to build PyTorch: N/A

OS: macOS 10.15.7 (x86_64)
GCC version: Could not collect
Clang version: 12.0.0 (clang-1200.0.32.29)
CMake version: Could not collect
Libc version: N/A

[...]

PR #3808, opened 2021-05-10
NOMRG Current CI status

NicolasHug:
This is just a dummy PR to see which jobs are green and which aren't
(Useful to decide whether failures are related for newly-created PRs)

Feel free to merge with master every now and then to keep it up to date

Please avoid commenting about new failures below. If there are failures, it's best to open a new issue and reference this PR in the issue. Also please do not push fixes to this branch, open a new PR instead.

PR #4148, opened 2021-07-01
Explicitly exclude PIL 8.3.0 from compatible dependencies

NicolasHug:
Closes https://github.com/pytorch/vision/issues/4146

CC @pmeier

### DIFF
--- a/.circleci/unittest/linux/scripts/environment.yml
+++ b/.circleci/unittest/linux/scripts/environment.yml
@@ -12,6 +12,6 @@ dependencies:
   - ca-certificates
   - pip:
     - future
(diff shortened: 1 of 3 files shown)


pmeier (approved):
Thanks!

pmeier (changes requested):
I think there are few more places where this needs to be pinned. At least some CI jobs still install `8.3.0`

NicolasHug:
Should be good now, unittests jobs are passing. The same binary jobs failures are present in https://github.com/pytorch/vision/pull/3808, so they're unrelated.

pmeier (approved):
Stamp

PR #4165, opened 2021-07-12
Catch ValueError due to recent change to torch.testing.assert_close

NicolasHug:
With https://github.com/pytorch/pytorch/pull/61031 now merged, we should catch `ValueError`s.

This should fix most of the currently failing tests.

CC @pmeier :) !


pmeier (approved):
LGTM, thanks @NicolasHug!

NicolasHug:
Thanks for the reviews, I'll look into the rest of the (unrelated) failures in a separate PR

PR #4167, opened 2021-07-12
Don't check transparency channel for pil >= 8.3

NicolasHug:
png decoding tests are  [currently failing](https://app.circleci.com/pipelines/github/pytorch/vision/9425/workflows/4e50f823-4378-4f55-8d0f-981d9c92dcbf/jobs/694580). Looking into it, the issue appeared in pillow 8.3 and I reported it in https://github.com/python-pillow/Pillow/issues/5593#issuecomment-878244910. The problem comes from PIL not properly setting the transparency channel of some images.

This is a temporary workaround (we ignore the check in those conditions) which should be removed once the fix is published in pillow.

As a bonus this uses `torch.testing.assert_close` instead of `torch.allclose`

### DIFF
--- a/test/test_image.py
+++ b/test/test_image.py
@@ -7,7 +7,7 @@ from pathlib import Path
 import pytest
 import numpy as np
 import torch
-from PIL import Image
+from PIL import Image, __version__ as PILLOW_VERSION
 import torchvision.transforms.functional as F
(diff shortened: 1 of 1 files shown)


fmassa (approved):
Thanks!

PR #4172, opened 2021-07-13
Added missing typing annotations in datasets/video_utils

frgfm:
Following up on #2025, this PR adds missing typing annotations in datasets/video_utils.py.

Any feedback is welcome!

cc @pmeier


pmeier:
I remember there was some issue preventing me from adding these annotations earlier. I think it was missing annotations in `torchvision.io` since we depend on that here. @fmassa is it fine to add these now? 

@frgfm the `mypy` failures are related:

```
torchvision/datasets/kinetics.py:137: error: Argument 4 to "VideoClips" has
incompatible type "Optional[float]"; expected "Optional[int]"  [arg-type]
                frame_rate,
                ^
Found 1 error in 1 file (checked 110 source files)
```

pmeier (changes requested):
@frgfm the `mypy` failures are related:

```
torchvision/datasets/kinetics.py:137: error: Argument 4 to "VideoClips" has
incompatible type "Optional[float]"; expected "Optional[int]"  [arg-type]
                frame_rate,
                ^
Found 1 error in 1 file (checked 110 source files)
```

torchvision/datasets/video_utils.py
```diff
@@ -55,17 +61,17 @@ class _VideoTimestampsDataset(object):
...
+    def __len__(self) -> int:
         return len(self.video_paths)
 
-    def __getitem__(self, idx):
+    def __getitem__(self, idx: int) -> Tuple[List[int], Optional[float]]:
         return read_video_timestamps(self.video_paths[idx])
 
 
-def _collate_fn(x):
+def _collate_fn(x: Any) -> Any:
```
pmeier, line 74: Can we make this more concrete?

```suggestion
from typing import TypeVar

T = TypeVar("T")

def _collate_fn(x: T) -> T:
```

Without that, the output of `collate_fn` will always be `Any` and thus effectively loosing us the information we have about the input type.

torchvision/datasets/video_utils.py
```diff
@@ -131,16 +137,16 @@ def __init__(
...
         self.video_pts = []
         self.video_fps = []
 
         # strategy: use a DataLoader to parallelize read_video_timestamps
         # so need to create a dummy dataset first
         import torch.utils.data
 
-        dl = torch.utils.data.DataLoader(
-            _VideoTimestampsDataset(self.video_paths),
+        dl: torch.utils.data.DataLoader = torch.utils.data.DataLoader(
```
pmeier, line 148: Why do we need to annotate `dl` here?

torchvision/datasets/video_utils.py
```diff
@@ -157,23 +163,23 @@ def _compute_frame_pts(self):
...
+    def metadata(self) -> Dict[str, Any]:
         _metadata = {
             "video_paths": self.video_paths,
             "video_pts": self.video_pts,
             "video_fps": self.video_fps,
         }
         return _metadata
 
-    def subset(self, indices):
+    def subset(self, indices: List[int]) -> Any:
```
pmeier, line 182:
```suggestion
    def subset(self, indices: List[int]) -> "VideoClips":
```

torchvision/datasets/video_utils.py
```diff
@@ -215,12 +227,12 @@ def compute_clips_for_video(video_pts, num_frames, step, fps, frame_rate):
             warnings.warn("There aren't enough frames in the current video to get a clip for the given clip length and "
                           "frames between clips. The video (and potentially others) will be skipped.")
         if isinstance(idxs, slice):
-            idxs = [idxs] * len(clips)
+            idxs = [idxs] * len(clips)  # type: ignore[assignment]
```
pmeier, line 230: Why do we need the ignore here? Doesn't that mean that the return annotation of `VideoClips._resample_video_idx` is wrong?

torchvision/datasets/video_utils.py
```diff
@@ -215,12 +227,12 @@ def compute_clips_for_video(video_pts, num_frames, step, fps, frame_rate):
...
         else:
             idxs = unfold(idxs, num_frames, step)
-        return clips, idxs
+        return clips, idxs  # type: ignore[return-value]
```
pmeier, line 233: Why do we need to ignore here? Doesn't that mean that the return annotation is wrong?

torchvision/datasets/video_utils.py
```diff
@@ -198,7 +204,13 @@ def subset(self, indices):
...
 
     @staticmethod
-    def compute_clips_for_video(video_pts, num_frames, step, fps, frame_rate):
+    def compute_clips_for_video(
+        video_pts: torch.Tensor,
+        num_frames: int,
+        step: int,
+        fps: int,
+        frame_rate: Optional[int] = None
+    ) -> Tuple[torch.Tensor, Union[List[slice], torch.Tensor]]:
```
pmeier, line 213: Maybe?

```suggestion
    ) -> Tuple[torch.Tensor, Union[List[int], torch.Tensor]]:
```

torchvision/datasets/video_utils.py
```diff
@@ -329,29 +341,29 @@ def get_clip(self, idx):
...
-            info = _probe_video_from_file(video_path)
-            video_fps = info.video_fps
+            _info = _probe_video_from_file(video_path)
+            video_fps = _info.video_fps
             audio_fps = None
 
-            video_start_pts = clip_pts[0].item()
-            video_end_pts = clip_pts[-1].item()
+            video_start_pts = int(clip_pts[0].item())
+            video_end_pts = int(clip_pts[-1].item())
```
pmeier, line 349: The output is already `int` if you call `.item()`. Unfortunately, `mypy` doesn't know that. In order to make the intent more clear here, i.e. "we want to make the static type checker happy", let's do

```suggestion
            from typing import cast
            video_start_pts = cast(int, clip_pts[0].item())
            video_end_pts = cast(int, clip_pts[-1].item())
```

pmeier (commented):
Thanks @frgfm for the update! I have some comments inline.

torchvision/datasets/video_utils.py
frgfm: mypy is throwing me errors otherwise:
```
torchvision/datasets/video_utils.py:150: error: Need type annotation for 'dl'  [var-annotated]
            dl = torch.utils.data.DataLoader(
                 ^
```

torchvision/datasets/video_utils.py
frgfm: but that would be incorrect?
```python
if step.is_integer():
            # optimization: if step is integer, don't need to perform
            # advanced indexing
            step = int(step)
            return slice(None, None, step)
```
makes it a slice if not a Tensor

torchvision/datasets/video_utils.py
pmeier: Weird, but ok.

torchvision/datasets/video_utils.py
pmeier: Now that we differentiate between `_idxs` (nit: should be `_idcs`) and `idxs` that might be true. Let's wait and see if `mypy` is happy.

frgfm: Looks like it's all good! Let me know if you want me to fix the typos (but there are several `idxs` in the files)

pmeier:
@datumbox Since the annotations heavily depend on `torchvision.io._video_utils` that will be annotated in #4173, let's finish that one first. Afterwards we can rebase and see if `mypy` still happy with the current implementation.

frgfm:
> @datumbox Since the annotations heavily depend on `torchvision.io._video_opts` that will be annotated in #4173, let's finish that one first. Afterwards we can rebase and see if `mypy` still happy with the current implementation.

Just updated the PR @pmeier :+1:

torchvision/datasets/video_utils.py
```diff
@@ -320,22 +326,22 @@ def get_clip(self, idx):
             end_pts = clip_pts[-1].item()
             video, audio, info = read_video(video_path, start_pts, end_pts)
         else:
-            info = _probe_video_from_file(video_path)
-            video_fps = info.video_fps
+            _info = _probe_video_from_file(video_path)
```
pmeier, line 329: Since we now have

https://github.com/pytorch/vision/blob/59baae99b3e1bd2e43b2c3aebf44505c5cdf1c90/mypy.ini#L6

this renaming is probably not needed anymore. Could you check?

torchvision/datasets/video_utils.py
```diff
@@ -371,10 +377,10 @@ def __getstate__(self):
         video_pts = [x.to(torch.int64) for x in self.video_pts]
         # video_pts can be an empty list if no frames have been decoded
         if video_pts:
-            video_pts = torch.cat(video_pts)
+            video_pts = torch.cat(video_pts)  # type: ignore[assignment]
```
pmeier, line 380: Same as above.

torchvision/datasets/video_utils.py
```diff
@@ -2,20 +2,23 @@
...
-from typing import List
+from typing import Any, Dict, List, Optional, Callable, Union, Tuple, TypeVar, cast
 
 import torch
 from torchvision.io import (
     _probe_video_from_file,
     _read_video_from_file,
     read_video,
     read_video_timestamps,
+    VideoMetaData,
```
pmeier, line 13: This seems to be unused as picked up by [`flake8`](https://app.circleci.com/pipelines/github/pytorch/vision/12677/workflows/919bab15-00d3-4f44-8b48-77dd7454a216/jobs/1005060).

pmeier (commented):
Three minor things left.

torchvision/datasets/video_utils.py
frgfm: I did and unfortunately it's still an issue

torchvision/datasets/video_utils.py
frgfm: same, I'm still having trouble here

torchvision/datasets/video_utils.py
frgfm: my bad!

torchvision/datasets/video_utils.py
pmeier: It seems even with `allow_redefinition`, `mypy` is [stricter than I thought](https://mypy.readthedocs.io/en/stable/config_file.html?highlight=allow_redefinition#confval-allow_redefinition):

> Allows variables to be redefined with an arbitrary type, **as long as the redefinition is in the same block and nesting level as the original definition.** [emphasis mine]

Since that is not the case here, we need to keep your fix.

torchvision/datasets/video_utils.py
pmeier: Same as above.

torchvision/datasets/video_utils.py
```diff
@@ -198,29 +200,32 @@ def subset(self, indices):
...
         if fps is None:
             # if for some reason the video doesn't have fps (because doesn't have a video stream)
             # set the fps to 1. The value doesn't matter, because video_pts is empty anyway
             fps = 1
         if frame_rate is None:
             frame_rate = fps
         total_frames = len(video_pts) * (float(frame_rate) / fps)
-        idxs = VideoClips._resample_video_idx(int(math.floor(total_frames)), fps, frame_rate)
+        _idxs = VideoClips._resample_video_idx(int(math.floor(total_frames)), fps, frame_rate)
         video_pts = video_pts[idxs]
```
pmeier, line 214:
```suggestion
        video_pts = video_pts[_idxs]
```

pmeier (approved):
LGTM when CI is happy. Thanks a lot @frgfm!

torchvision/datasets/video_utils.py
```diff
@@ -198,29 +200,32 @@ def subset(self, indices):
...
+        video_pts: torch.Tensor, num_frames: int, step: int, fps: int, frame_rate: Optional[int] = None
+    ) -> Tuple[torch.Tensor, Union[List[slice], torch.Tensor]]:
         if fps is None:
             # if for some reason the video doesn't have fps (because doesn't have a video stream)
             # set the fps to 1. The value doesn't matter, because video_pts is empty anyway
             fps = 1
         if frame_rate is None:
             frame_rate = fps
         total_frames = len(video_pts) * (float(frame_rate) / fps)
-        idxs = VideoClips._resample_video_idx(int(math.floor(total_frames)), fps, frame_rate)
```
prabhat00155, line 209: What's the point of this change in variable name?

torchvision/datasets/video_utils.py
```diff
@@ -131,16 +133,16 @@ def __init__(
...
         self.video_fps = []
 
         # strategy: use a DataLoader to parallelize read_video_timestamps
         # so need to create a dummy dataset first
         import torch.utils.data
 
-        dl = torch.utils.data.DataLoader(
-            _VideoTimestampsDataset(self.video_paths),
+        dl: torch.utils.data.DataLoader = torch.utils.data.DataLoader(
+            _VideoTimestampsDataset(self.video_paths),  # type: ignore[arg-type]
```
prabhat00155, line 145: Why type ignore here?

torchvision/datasets/video_utils.py
```diff
@@ -320,22 +325,22 @@ def get_clip(self, idx):
             end_pts = clip_pts[-1].item()
             video, audio, info = read_video(video_path, start_pts, end_pts)
         else:
-            info = _probe_video_from_file(video_path)
```
prabhat00155, line 323: Again, why are you adding `_` in front of variable names?

torchvision/datasets/video_utils.py
```diff
@@ -371,10 +376,10 @@ def __getstate__(self):
         video_pts = [x.to(torch.int64) for x in self.video_pts]
         # video_pts can be an empty list if no frames have been decoded
         if video_pts:
-            video_pts = torch.cat(video_pts)
+            video_pts = torch.cat(video_pts)  # type: ignore[assignment]
```
prabhat00155, line 379: Were you getting warnings here? Why this type ignore?

prabhat00155 (commented):
Thanks @frgfm. I have a few comments.

torchvision/datasets/video_utils.py
pmeier: See https://github.com/pytorch/vision/pull/4172/files/b8887ec9449ba5e8719ab9cf1c44c375d9ecb22e#r679697283.

torchvision/datasets/video_utils.py
pmeier: We are overwriting it with a different type below and `mypy` doesn't like that. See https://github.com/pytorch/vision/pull/4172/files/b8887ec9449ba5e8719ab9cf1c44c375d9ecb22e#r752061997

torchvision/datasets/video_utils.py
pmeier: Before this branch `video_opts` is a `List[Tensor]`, but now becomes a `Tensor`.

pmeier (commented):
@prabhat00155 All the question can be answered with "Because we are using a static analyzer on dynamic code". Usually when we encounter weirdness, it is either because `mypy` is unnecessarily strict or we actually have spaghetti / unsafe code that should be refactored. Given that the latter is usually not obvious, we opted to simply silence `mypy` where it complains, since the code works as is. 

If you want to see all the errors, you can checkout this PR locally, revert the changes you had questions about and run `mypy`.

prabhat00155 (approved):

PR #4173, opened 2021-07-13
Added typing annotations to io/_video_opts

frgfm:
Following up on #2025, this PR adds missing typing annotations in io/_video_opts.py.

Any feedback is welcome!


pmeier (changes requested):
Thanks for the PR @frgfm. Unfortunately this breaks `torch.jit`. Relevant test failure from `test_probe_video_from_memory_script - test.test_video_reader.TestVideoReader`

```
E           RuntimeError: 
E           Unknown type name 'np.ndarray':
E             File "/root/project/torchvision/io/_video_opt.py", line 432
E           def _probe_video_from_memory(
E               video_data: Union[torch.Tensor, np.ndarray],
E                                               ~~~~~~~~~~ <--- HERE
E           ) -> VideoMetaData:
E               """
```

frgfm:
> Thanks for the PR @frgfm. Unfortunately this breaks `torch.jit`. Relevant test failure from `test_probe_video_from_memory_script - test.test_video_reader.TestVideoReader`
> 
> ```
> E           RuntimeError: 
> E           Unknown type name 'np.ndarray':
> E             File "/root/project/torchvision/io/_video_opt.py", line 432
> E           def _probe_video_from_memory(
> E               video_data: Union[torch.Tensor, np.ndarray],
> E                                               ~~~~~~~~~~ <--- HERE
> E           ) -> VideoMetaData:
> E               """
> ```

Oh, should I remove the typing for this function only then @pmeier?

pmeier:
The problem here is that `torch.jit` currently does not support the `Union` type. There is a PR pytorch/pytorch#53180 that adds support, but it has not been landed yet. 

I think the way forward is to try to eliminate the `Union` and try to get `mypy` to live with it. If that is not possible than yes, we need to remove the annotations again since in general `torch.jit` > `mypy`.

frgfm:
Apart from the `Union` issue with `torch.jit`, I checked locally and mypy is working perfectly.
I had to ignore assignment I couldn't get to work at the end of the file, and if you have an idea how to get around the jit issue, I'm all ears @pmeier :)

torchvision/io/_video_opt.py
```diff
@@ -24,21 +24,20 @@
 # simple class for torch scripting
 # the complex Fraction class from fractions module is not scriptable
 class Timebase(object):
-    __annotations__ = {"numerator": int, "denominator": int}
-    __slots__ = ["numerator", "denominator"]
+    __annotations__: Dict[str, Type[int]] = {"numerator": int, "denominator": int}
+    __slots__: List[str] = ["numerator", "denominator"]
```
pmeier, line 28: Are these necessary? Can't `mypy` infer them automatically?

torchvision/io/_video_opt.py
```diff
@@ -24,21 +24,20 @@
 # simple class for torch scripting
 # the complex Fraction class from fractions module is not scriptable
 class Timebase(object):
-    __annotations__ = {"numerator": int, "denominator": int}
-    __slots__ = ["numerator", "denominator"]
+    __annotations__: Dict[str, Type[int]] = {"numerator": int, "denominator": int}
```
pmeier, line 27: Shouldn't that be

```suggestion
    __annotations__: Dict[str, Type] = {"numerator": int, "denominator": int}
```

Otherwise we can only have integer attributes, right?

torchvision/io/_video_opt.py
```diff
@@ -24,21 +24,20 @@
...
+        numerator: int,
+        denominator: int,
+    ) -> None:
         self.numerator = numerator
         self.denominator = denominator
 
 
 class VideoMetaData(object):
-    __annotations__ = {
+    __annotations__: Dict[str, Any] = {
```
pmeier, line 40:
```suggestion
    __annotations__: Dict[str, Type] = {
```

torchvision/io/_video_opt.py
```diff
@@ -517,7 +531,7 @@ def get_pts(time_base):
     return vframes, aframes, _info
 
 
-def _read_video_timestamps(filename, pts_unit="pts"):
+def _read_video_timestamps(filename: str, pts_unit: str = "pts") -> Tuple[List[Fraction], Optional[float]]:
```
pmeier, line 534:
```suggestion
def _read_video_timestamps(
        filename: str, pts_unit: str = "pts"
) -> Tuple[Union[List[int], List[Fraction]], Optional[float]]:
```

In addition, you need to put `pts: Union[List[int], List[Fraction]]` before the first assignment of it. Doing that, we can drop the ignores.

torchvision/io/_video_opt.py
```diff
@@ -530,8 +544,8 @@ def _read_video_timestamps(filename, pts_unit="pts"):
         video_time_base = Fraction(
             info.video_timebase.numerator, info.video_timebase.denominator
         )
-        pts = [x * video_time_base for x in pts]
+        pts = [x * video_time_base for x in pts]  # type: ignore[misc]
```
pmeier, line 547:
```suggestion
        pts = [x * video_time_base for x in pts]
```

torchvision/io/_video_opt.py
```diff
@@ -530,8 +544,8 @@ def _read_video_timestamps(filename, pts_unit="pts"):
...
 
     video_fps = info.video_fps if info.has_video else None
 
-    return pts, video_fps
+    return pts, video_fps  # type: ignore[return-value]
```
pmeier, line 551:
```suggestion
    return pts, video_fps
```

pmeier (commented):
> Apart from the `Union` issue with `torch.jit`

Until this is fixed upstream, there is nothing we can do about it.

> I had to ignore assignment I couldn't get to work at the end of the file

I've answered inline with some more comments.

torchvision/io/_video_opt.py
frgfm: Correct, but I thought this was the purpose of the annotation attribute :man_shrugging:

torchvision/io/_video_opt.py
frgfm: That might be true, but I thought it would be better for future contributors that class attributes would have specified types. Do you think we should remove this? (going for the minimal so that mypy is happy I mean)

pmeier: > Do you think we should remove this?

IMO, yes. Since this is not a public interface, we should let `mypy` infer whatever it can.

pmeier (commented):
Hey @frgfm this is looking almost ready to go! The CI is complaining about one more issue that pops up, because we no longer ignore errors in `torchvision.io`. You can either ignore it for now

```ini
[mypy-torchvision.io]

ignore_errors = True
```

or fix it, since it is a fairly small fix. I'm ok with both. If you choose the latter, you need to fix the annotation of `VideoReader.__iter__`. It should be `Iterator[T]` with `T` being whatever `__next__` returns.

frgfm:
> Hey @frgfm this is looking almost ready to go! The CI is complaining about one more issue that pops up, because we no longer ignore errors in `torchvision.io`. You can either ignore it for now
> 
> ```ini
> [mypy-torchvision.io]
> 
> ignore_errors = True
> ```
> 
> or fix it, since it is a fairly small fix. I'm ok with both. If you choose the latter, you need to fix the annotation of `VideoReader.__iter__`. It should be `Iterator[T]` with `T` being whatever `__next__` returns.

I went for the 2nd option, let's hope the CI is pleased :)

pmeier (approved):
Thanks a lot @frgfm!

torchvision/io/_video_opt.py
```diff
@@ -113,8 +117,11 @@ def _fill_info(vtimebase, vfps, vduration, atimebase, asample_rate, aduration):
     return meta
 
 
-def _align_audio_frames(aframes, aframe_pts, audio_pts_range):
-    # type: (torch.Tensor, torch.Tensor, List[int]) -> torch.Tensor
+def _align_audio_frames(
+    aframes: torch.Tensor,
+    aframe_pts: torch.Tensor,
```
datumbox, line 122: Are we sure that's a Tensor and not a tuple of ints?

torchvision/io/_video_opt.py
```diff
@@ -438,15 +447,20 @@ def _probe_video_from_memory(video_data):
     return info
 
 
-def _convert_to_sec(start_pts, end_pts, pts_unit, time_base):
+def _convert_to_sec(start_pts: float, end_pts: float, pts_unit: str, time_base: Fraction) -> Tuple[float, float, str]:
```
datumbox, line 450: It seems that `start_pts`, `end_pts`, `*pts*` switch between ints and floats in different methods. It's worth checking that in all signatures where we declare them as ints are indeed always ints. That might be easier to do with a debugger.

cc @prabhat00155

torchvision/io/_video_opt.py
frgfm: Well tracing it back, it looks like so https://github.com/pytorch/vision/blob/7947fc8fb38b1d3a2aca03f22a2e6a3caa63f2a0/torchvision/csrc/io/video_reader/video_reader.h#L29

`aframe_pts` is one of the outputs of this function

torchvision/io/_video_opt.py
frgfm: As mentioned in the other comment, there is no clear dtype in https://github.com/pytorch/vision/blob/7947fc8fb38b1d3a2aca03f22a2e6a3caa63f2a0/torchvision/csrc/io/video_reader/video_reader.h#L29 and considering it's not casted afterwards, I'm not sure how to answer this :man_shrugging:

datumbox: Sorry I missed the other comment. I understand it's not straightforward to identify the types. That's why I think finding out what they are and annotating the code-base is useful. Nevertheless if we fail to annotate them correctly (say they are floats and we mark them as ints) it's going to be really confusing. 

Can we confirm the types of the various pts vars by running the unit-tests and putting a debugger to observe their materialized types?

torchvision/io/_video_opt.py
```diff
@@ -438,15 +447,20 @@ def _probe_video_from_memory(video_data):
...
         end_pts = float(end_pts * time_base)
         pts_unit = 'sec'
     return start_pts, end_pts, pts_unit
 
 
-def _read_video(filename, start_pts=0, end_pts=None, pts_unit="pts"):
+def _read_video(
+    filename: str,
+    start_pts: int = 0,
+    end_pts: Optional[float] = None,
```
prabhat00155, line 461: `end_pts` can be int, float or Fraction. So, this should be Optional[Union[float, Fraction]].

torchvision/io/_video_opt.py
```diff
@@ -438,15 +447,20 @@ def _probe_video_from_memory(video_data):
...
         start_pts = float(start_pts * time_base)
         end_pts = float(end_pts * time_base)
         pts_unit = 'sec'
     return start_pts, end_pts, pts_unit
 
 
-def _read_video(filename, start_pts=0, end_pts=None, pts_unit="pts"):
+def _read_video(
+    filename: str,
+    start_pts: int = 0,
```
prabhat00155, line 460: `start_pts` can be int, float or Fraction, so this should be `Union[float, Fraction]`.

torchvision/io/_video_opt.py
frgfm: Apart from if I'm mistaking, `Union` are not yet allowed to ensure torchscript is still working :/

pmeier: pytorch/pytorch#64234 added support for `Union`'s so feel free to use them now!

frgfm: Holy cow, and I wasn't notified ?? :joy:
Let's clean all those annotations then :smile: I'll deal with it shortly!

frgfm: My apologies, I forgot about the "shortly" along the way, I just dealt with it :sweat_smile:

datumbox:
Unfortunately you got a few [linter](https://app.circleci.com/pipelines/github/pytorch/vision/12616/workflows/6d20bfcf-d107-4e8f-b437-fb7b9e68962a/jobs/999168) and [typing](https://app.circleci.com/pipelines/github/pytorch/vision/12616/workflows/834a45c5-5be5-4449-ac2e-9884935f7b56/jobs/999214) failures.

torchvision/io/_video_opt.py
prabhat00155: Not sure this has been resolved.

torchvision/io/_video_opt.py
```diff
@@ -495,13 +514,17 @@ def get_pts(time_base):
...
+    filename: str,
+    pts_unit: str = "pts"
+) -> Tuple[Union[List[int], List[Fraction]], Optional[float]]:
     if pts_unit == "pts":
         warnings.warn(
             "The pts_unit 'pts' gives wrong results and will be removed in a "
             + "follow-up version. Please use pts_unit 'sec'."
         )
 
+    pts: Union[List[int], List[Fraction]]
```
prabhat00155, line 527: Why this line? Type should be clear from the return type of `_read_video_timestamps_from_file()`.

torchvision/io/_video_opt.py
```diff
@@ -495,13 +514,17 @@ def get_pts(time_base):
...
     pts, _, info = _read_video_timestamps_from_file(filename)
```
prabhat00155, line 528: I am bit confused here. You have annotated `_read_video_timestamps_from_file()` above as returning a tuple with its first element being `List[int]`. However, you have annotated `pts` as:
```
pts: Union[List[int], List[Fraction]]
```

torchvision/io/_video_opt.py
frgfm: Oh my bad, I thought it was a bug on github and that was a duplicate for the "end_pnts" comment!

torchvision/io/_video_opt.py
frgfm: So the redefinition of pts on line 527 doesn't please mypy, so I type it upfront to avoid issues over there

torchvision/io/_video_opt.py
frgfm: Yes, because if we don't go in the for loop: it's indeed a `List[int]` as you mentioned
But if you go in the for loop, it becomes a `List[Fraction]` to the best of my understanding :+1: 

If that's not correct, happy to change it!

prabhat00155 (approved):
Thanks @frgfm!

torchvision/io/__init__.py
```diff
@@ -132,7 +132,7 @@ def __next__(self) -> Dict[str, Any]:
             raise StopIteration
         return {"data": frame, "pts": pts}
 
-    def __iter__(self) -> Iterator["VideoReader"]:
+    def __iter__(self) -> "VideoReader":
```
pmeier, line 135:
```suggestion
    def __iter__(self) -> Iterator[Dict[str, Any]]:
```

torchvision/io/_video_opt.py
```diff
@@ -45,7 +44,7 @@ class VideoMetaData:
         "audio_duration": float,
         "audio_sample_rate": float,
     }
-    __slots__ = [
+    __slots__: List[str] = [
```
pmeier, line 47: Does `mypy` complain here if we drop the annotation? It should be able to deduce it automatically.

pmeier (approved):
Two minor things left from my side. LGTM, when the other open comments are resolved and CI is happy. Thanks a lot @frgfm!

torchvision/io/_video_opt.py
frgfm: Oh you're correct, I think I remember I had to in the beginning, but that problem seems to have vanished (many commits on main since then I guess)

PR #4177, opened 2021-07-14
Cleanup test suite related to `torch.testing.assert_close`

pmeier:
This

1. moves `assert_equal` from `test/_assert_utils.py` to `test/common_utils.py` since after pytorch/pytorch#58918 we need to keep it permanently and
2. removes the `check_stride=False` since this became the default in pytorch/pytorch#60637.


NicolasHug (approved):
Thanks @pmeier ! Happy about the `check_stride` default change :)

There was one last occurrence left, I just pushed it and will merge when green!

PR #4178, opened 2021-07-14
[DO NOT MERGE] introduce `isort` and `black` as autoformatters

### LINKED ISSUE #4174: Use black?
Formatting the codebase with [black](https://github.com/psf/black) would help in a few ways:
- avoid review nitpicks regarding style
- make signatures easier to read (when type annotations are involved, things can get *really* ugly)
- avoid PRs where contributors IDE automatically re-format the  imports etc, and we always have to tell them to revert such unrelated changes
- help avoiding the CI linting issues
- overall style consistency

Cons:

- Adds noise to `git blame`, namely 1 extra click. **This is for GitHub only**, locally nothing is changed (see below)
- Adds a lot of vertical length to the code. It can be a bit unsettling/ugly at first, but you get used to it
- a bit more setup/work for new contributors, but black is becoming more and more of a standard anyway so people are mostly used to it, and pre-commit hooks are very easy to setup.
- will create conflicts with ongoing PRs. Not really a big deal, there aren't too many PRs and resolving conflicts should be fairly simple
- As discovered with @pmeier, this forces to add a new `--no-build-isolation` to all `pip install --editable` commands  [1], as done in https://github.com/pytorch/vision/pull/4178/commits/3d1cd1daf7412b2b7f5150d6ea47ce9bb109d1ee. This isn't a big deal either as long as we provide good guidelines, and we recommend `conda` over `pip` anyway

Using black has already been discussed in https://github.com/pytorch/vision/issues/2025, with general positive sentiment towards adoption. The main concern seems to be that `black` brakes `git blame`, but it's actually not the case: we can avoid breaking `git glame` thanks to the [ignore-revs-file option](https://black.readthedocs.io/en/stable/guides/introducing_black_to_your_project.html#avoiding-ruining-git-blame) (unfortunately, GitHub [doesn't support it yet](https://github.community/t/support-ignore-revs-file-in-githubs-blame-view/3256), so the Github blame may still be a bit noisy).

[1] The reason for this nonsense is that `black` relies ([only](https://github.com/psf/black/issues/683#)) on a `pyproject.toml` file for its configuration. This file was initially intended to declare [build-time dependencies](https://www.python.org/dev/peps/pep-0518/) (something **completely** unrelated to code formatting, but oh well), and it can make pip behave differently in editable mode https://github.com/psf/black/issues/683#issuecomment-487306157, as illustrated in https://app.circleci.com/pipelines/github/pytorch/vision/9457/workflows/4007ced4-684c-498d-8c78-f6ee83853a7d/jobs/697466 . We need the [--no-build-isolation](https://pip.pypa.io/en/stable/cli/pip/#pep-517-and-518-support) flag to preserve `pip`'s sane behaviour, now that this file is added.

----

Adopting black would involve:
- a black-formatting PR (duh). Might consider make the PR author a bot or some dummy account.
- A PR that adds the `ignore-revs-file` file, referencing the previous PR for `git-blame` to be preserved.
- some instructions in the contributing guide to use `black`, edit the "install from source" docs with `pip install -e .` to use the `--no-build-isolation` flag, and optionally setup a pre-commit hook as done e.g. in [pandas](https://pandas.pydata.org/docs/development/contributing_codebase.html#pre-commit)
- some instructions e.g. as a sticky issue to resolve potential merge conflicts
- a new CI check. This could just be part of the current job that checks for flake8.

I'm happy to work on this if we agree to rely on `black` from now on.

CC @fmassa @pmeier

pmeier:
Closes #4174. For now this is just a proof of concept PR. **DO NOT MERGE** this until we actually want to use `black` and `isort`.

Only `pyproject.toml` and `setup.cfg` need review. All the other changes were made by the autoformatters or to accommodate `flake8`. 

Todo:

- [x] Add `black` and `isort` to CI
- [ ] Add `black` and `isort` as pre-commit hooks


torchvision/ops/poolers.py
```diff
@@ -1,11 +1,11 @@
...
+from typing import Dict, List, Optional, Tuple, Union
 
+import torch
 import torchvision
-from torchvision.ops import roi_align
+from torch import Tensor, nn
 from torchvision.ops.boxes import box_area
 
-from typing import Optional, List, Dict, Tuple, Union
+from .roi_align import roi_align
```
pmeier, line 8: Our current implementation relies on a specific import order in `torchvision/ops/__init__.py` since we have `torchvision/ops/roi_align.py` as well as a `roi_align` function. Thus, depending on the current stage of the imports `from torchvision.ops import roi_align` gives you either the module or the function.

In the future we should avoid name conflicts between modules and functions. For example, we could rename the module to `_roi_align.py` since all its functionality is available through `torchvision.ops`.

setup.cfg
```diff
@@ -9,7 +9,13 @@ max-line-length = 120
 
 [flake8]
 max-line-length = 120
-ignore = F401,E402,F403,W503,W504,F821
+ignore = E203,E402,F403,W503,W504,F821
+per-file-ignores =
```
NicolasHug, line 13: What is the reason for this change, in particular switching to ignoring F401 on a per-file basis?

gallery/plot_visualization_utils.py
```diff
@@ -140,18 +136,34 @@ def show(imgs):
 # boat class:
 
 sem_classes = [
-    '__background__', 'aeroplane', 'bicycle', 'bird', 'boat', 'bottle', 'bus',
-    'car', 'cat', 'chair', 'cow', 'diningtable', 'dog', 'horse', 'motorbike',
-    'person', 'pottedplant', 'sheep', 'sofa', 'train', 'tvmonitor'
+    "__background__",
```
NicolasHug, line 139: Let's tell `black` to ignore the `gallery` folder. We don't want readers to have to scroll for such narrative-style documentation.

NicolasHug (commented):
Thanks @pmeier !

Considering the sequence of changes that will need to be done (in particular the doc updates, the pre-commit hook and the CI check), I would suggest to do all  this in a separate branch first. We'll then be able to merge everything in master in one go, while keeping  the development in separate PRs without worrying too much about the current state of master.

setup.cfg
pmeier: F401 is unused imports. They were bugging me for some time. In some modules like `__init__.py` though we simply import stuff for accessibility. In these files we simply ignore the unused imports.

pmeier: E203 is whitespace around slice syntax. `black` and `flake8` disagree on this and thus we ignore it to be able to use `black`.

gallery/plot_visualization_utils.py
pmeier: You mean `isort`, right? Telling `black` ignore it will not make a difference for the imports.

NicolasHug: I meant `black` (as it can introduce a lot of vertical space), but we should avoid applying `isort` there too :)

setup.cfg
```diff
@@ -9,7 +9,13 @@ max-line-length = 120
...
+    __init__.py: F401, F403, F405
```
NicolasHug, line 14: F403 seems to be a global ignore above, maybe we can remove it?

setup.cfg
NicolasHug: Thanks for the details, sounds good. I feel like we should clean up those `import *` eventually... but let's leave it for future PRs, might be a good-first-issue thing

setup.cfg
pmeier: F403 will be raised if `flake8` finds `from foo import *`. We should remove this from the global ignores and only allow this in the files below.

pmeier: There are none besides in the files in that list. Thus, I think we should go forward with my plan.

NicolasHug: I agree with the plan, but F403 is still part of the global `ignore` list right now

pmeier: Sorry, misunderstood your intent.

pyproject.toml
```diff
@@ -0,0 +1,40 @@
+[build-system]
+requires = [
+  "setuptools >= 40.9.0",
+  "wheel",
+]
+build-backend = "setuptools.build_meta"
```
NicolasHug, line 6: Could you detail why we need this new `[build-system]` section?

pmeier: Just a test to get the CI builds running. Locally everything works so I need to debug on CI. From the failing tests I get that this is not the issue. Trying `--no-build-isolation` next as you suggested offline.

pmeier:
Superseded by #4384.

PR #4179, opened 2021-07-14
Don't install torchvision as editable mode when not needed

NicolasHug:
I'm not sure we really need to install torchvision in editable mode for the CI jobs, as sources are not edited whatsoever in those jobs


pmeier (approved):
LGTM, thanks @NicolasHug!

Issue #4183, opened 2021-07-15
MNIST Download Raises 'UserWarning: The given NumPy array is not writeable'


JAEarly:
## 🐛 Bug

When downloading MNIST (using torchvision/datasets/mnist.py), the following UserWarning is raised:

```
UserWarning: The given NumPy array is not writeable, and PyTorch does not support non-writeable tensors. This means you can write to the underlying (supposedly non-writeable) NumPy array using the tensor. You may want to copy the array to protect its data or make it writeable before converting it to a tensor. This type of warning will be suppressed for the rest of this program. (Triggered internally at  ..\torch\csrc\utils\tensor_numpy.cpp:180.)
  return torch.from_numpy(parsed.astype(m[2], copy=False)).view(*s)
```

Changing line 498 in torchvision/datasets/mnist.py from
` return torch.from_numpy(parsed.astype(m[2], copy=False)).view(*s) `
to
` return torch.from_numpy(parsed.astype(m[2])).view(*s) `
removes the error.

Checking the Torch documentation (https://pytorch.org/docs/stable/generated/torch.from_numpy.html) isn't too helpful.
There is no record of  the `copy=False` arg.

## To Reproduce

A simple test of:
```
from torchvision.datasets import MNIST
MNIST('data/', train=True, download=True)
```
raises the warning.
Removing `copy=False` as discussed above then stops the warning from being raised.

## Expected behavior

The warning should not be raised.
As the warning is suppressed after this happens, it may not be raised elsewhere when it's actually important. 

## Environment

**Script output:**
PyTorch version: 1.9.0+cu102
Is debug build: False
CUDA used to build PyTorch: 10.2
ROCM used to build PyTorch: N/A

OS: Microsoft Windows 10 Enterprise
GCC version: Could not collect
Clang version: Could not collect
CMake version: Could not collect
Libc version: N/A

Python version: 3.8.6rc1 (tags/v3.8.6rc1:08bd63d, Sep  7 2020, 23:10:23) [MSC v.1927 64 bit (AMD64)] (64-bit runtime)
Python platform: Windows-10-10.0.19041-SP0
Is CUDA available: True
CUDA runtime version: 10.2.89
GPU models and configuration: GPU 0: GeForce GTX 1650
Nvidia driver version: 462.31
cuDNN version: Could not collect
HIP runtime version: N/A
MIOpen runtime version: N/A

Versions of relevant libraries:
[pip3] numpy==1.21.0
[...]

PR #4184, opened 2021-07-15
Fixed MNIST Download Raises 'UserWarning: The given NumPy array is not writeable'

JAEarly:
Fix as discussed in #4183


NicolasHug (approved):
Thanks @JAEarly

PR #4185, opened 2021-07-15
[ONNX] Remove conditional import of onnxruntime

garymm:
The comment indicates that this was put in place to support older
versions of python, but torchvision now requires python >=3.6.

Instead of having the test pipeline invoke test_onnx.py and just have
it skip all the tests contained therein, instead configure the pipeline
to not invoke test_onnx.py. This is better because otherwise people
may be given the false impression that the onnx tests are running and
passing, when in fact they're being skipped.

### DIFF
--- a/test/test_onnx.py
+++ b/test/test_onnx.py
@@ -1,13 +1,6 @@
-# onnxruntime requires python 3.5 or above
-try:
-    # This import should be before that of torch
-    # see https://github.com/onnx/onnx/issues/2394#issuecomment-581638840
-    import onnxruntime
-except ImportError:
-    onnxruntime = None
-
 from common_utils import set_rng_seed, assert_equal
 import io
+import pytest
 import torch
 from torchvision import ops
 from torchvision import models
@@ -20,11 +13,13 @@ from torchvision.models.detection.faster_rcnn import FastRCNNPredictor, TwoMLPHe
 from collections import OrderedDict
 from typing import List, Tuple
 
-import pytest
 from torchvision.ops._register_onnx_ops import _onnx_opset_version
 
+# In environments without onnxruntime we prefer to
+# invoke all tests in the repo and have this one skipped rather than fail.
+onnxruntime = pytest.importorskip("onnxruntime")
 
-@pytest.mark.skipif(onnxruntime is None, reason='ONNX Runtime unavailable')
+
 class TestONNXExporter:
     @classmethod
     def setup_class(cls):


.circleci/unittest/linux/scripts/run_test.sh
```diff
@@ -7,4 +7,7 @@ conda activate ./env
 
 export PYTORCH_TEST_WITH_SLOW='1'
 python -m torch.utils.collect_env
-pytest --cov=torchvision --junitxml=test-results/junit.xml -v --durations 20 test --ignore=test/test_datasets_download.py
+# test_onnx requires additional packages to be installed and has its own pipeline.
+pytest --cov=torchvision --junitxml=test-results/junit.xml -v --durations 20 test \
+  --ignore=test/test_datasets_download.py \
+  --ignore=test/test_onnx.py
```
NicolasHug, line 13: Let's remove this: I think it's actually valuable to see the tests being skipped in the CI: it shows they exist, and that they're skipped. If anything, it shows that the `skipif` logic actually works as expected.

As you said they're being run by another CI run anyway so there's no risk of missing them.

NicolasHug (commented):
Thanks @garymm , LGTM with a minor comment: I think we don't need to hide the tests that are skipped in the CI

.circleci/unittest/linux/scripts/run_test.sh
NicolasHug: Sorry, I forgot that part.

If you wish to remove the try/except, we should be able to convert it into `importorskip`: https://docs.pytest.org/en/latest/how-to/skipping.html#skipping-on-a-missing-import-dependency

garymm: Nice. Done.

test/test_onnx.py
```diff
@@ -1,13 +1,11 @@
...
-except ImportError:
-    onnxruntime = None
-
 from common_utils import set_rng_seed, assert_equal
 import io
+import pytest
+# In environments without onnxruntime we prefer to
+# invoke all tests in the repo and have this one skipped rather than fail.
+# This import should be before that of torch
+# see https://github.com/onnx/onnx/issues/2394#issuecomment-581638840
```
NicolasHug, line 7: Thanks @garymm , looks good. Do you know if this comment is still relevant? The original issue was closed so I feel like we should be able to remove the comment and do that `importskip` lower down, where it will be easier to find.

Do you mind trying that, and if there are still issue, we can put it back?

garymm: Thanks for checking this. Done.

NicolasHug (approved):
Thanks @garymm !

PR #4196, opened 2021-07-20
add `pytest.ini`

### LINKED ISSUE #4194: Add a `pytest.ini`
## 🚀 Feature

Add a `pytest.ini`.

## Motivation

IMO, the default `pytest` configuration is horrible and leads to enormous tracebacks that carry almost no information. [Example](https://app.circleci.com/pipelines/github/pytorch/vision/9503/workflows/395d34f3-1f69-4c04-bfed-fff13e77f403/jobs/702140/tests#failed-test-0)

## Pitch

Let's add a `pytest.ini` to the root of the project configuring the output to a more usable state. We can use [PyTorch core as base](https://github.com/pytorch/pytorch/blob/master/pytest.ini).

## Alternatives

If we for some reason do not want to add this for everyone, we still should add the appropriate flags to our CI runs.

cc @seemethere @pmeier

pmeier:
Closes #4194.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -3,7 +3,7 @@ addopts =
     # show summary of all tests that did not pass
     -ra
     # Make tracebacks shorter
-    --tb=short
+    --tb=native
     # enable all warnings
     -Wd
 testpaths =
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,10 @@
+[pytest]
+addopts =
+    # show summary of all tests that did not pass
+    -ra
+    # Make tracebacks shorter
+    --tb=short
+    # enable all warnings
+    -Wd
+testpaths =
+    test


pytest.ini
```diff
@@ -0,0 +1,10 @@
+[pytest]
+addopts =
+    # show summary of all tests that did not pass
+    -ra
+    # Make tracebacks shorter
+    --tb=short
```
NicolasHug, line 6: pytorch uses native. I'm not too familiar with it, do you have a preference?

NicolasHug (approved):
Thanks @pmeier , LGTM, feel free to self merge. We can always adjust in the future, should we want to

pytest.ini
pmeier: Depends. `native` might get better tracebacks for non-python errors. `short` is even more concise for default Python tracebacks. No strong opinion.

PR #4032, opened 2021-06-10
Port test_internet.py to pytest and add new pytest-mock dependency

NicolasHug:
This PR ports the `test_internet.py` tests to pytest.

It also adds the `pytest-mock` test-time dependency. I'll comment below for why this is probably preferable, but I'm happy to hear y'all thoughts on this.

PR #4197, opened 2021-07-20
Mock redirection logic for tests

pmeier:
Fixes recent CI failures like [this](https://app.circleci.com/pipelines/github/pytorch/vision/9503/workflows/395d34f3-1f69-4c04-bfed-fff13e77f403/jobs/702140/tests). Two of our tests depend on the Caltech servers, which are down for some days #4188. This PR removes the dependency and mocks out the redirection behavior.

### DIFF
--- a/test/test_datasets_utils.py
+++ b/test/test_datasets_utils.py
@@ -29,7 +29,7 @@ def patch_url_redirection(mocker, redirect_url):
     def patched_opener(*args, **kwargs):
         yield Response(redirect_url)
 
-    return mocker.patch("torchvision.datasets.utils.urllib.request.urlopen", new=patched_opener)
+    return mocker.patch("torchvision.datasets.utils.urllib.request.urlopen", side_effect=patched_opener)
 
 
 class TestDatasetsUtils:
@@ -37,19 +37,27 @@ class TestDatasetsUtils:
         url = "https://url.org"
         expected_redirect_url = "https://redirect.url.org"
 
-        patch_url_redirection(mocker, expected_redirect_url)
+        mock = patch_url_redirection(mocker, expected_redirect_url)
 
         actual = utils._get_redirect_url(url)
         assert actual == expected_redirect_url
+
+        assert mock.call_count == 2
+        call_args_1, call_args_2 = mock.call_args_list
+        assert call_args_1[0][0].full_url == url
+        assert call_args_2[0][0].full_url == expected_redirect_url
 
     def test_get_redirect_url_max_hops_exceeded(self, mocker):
         url = "https://url.org"
         redirect_url = "https://redirect.url.org"
 
-        patch_url_redirection(mocker, redirect_url)
+        mock = patch_url_redirection(mocker, redirect_url)
 
         with pytest.raises(RecursionError):
             utils._get_redirect_url(url, max_hops=0)
+
+        assert mock.call_count == 1
+        assert mock.call_args[0][0].full_url == url
 
     def test_check_md5(self):
         fpath = TEST_FILE
--- a/test/test_datasets_utils.py
+++ b/test/test_datasets_utils.py
@@ -34,8 +34,8 @@ def patch_url_redirection(mocker, redirect_url):
 
 class TestDatasetsUtils:
     def test_get_redirect_url(self, mocker):
-        url = "http://www.vision.caltech.edu/visipedia-data/CUB-200-2011/CUB_200_2011.tgz"
-        expected_redirect_url = "https://drive.google.com/file/d/1hbzc_P1FuxMkcabkgn9ZKinBwW683j45/view"
+        url = "https://url.org"
+        expected_redirect_url = "https://redirect.url.org"
 
         patch_url_redirection(mocker, expected_redirect_url)
 
@@ -43,8 +43,8 @@ class TestDatasetsUtils:
         assert actual == expected_redirect_url
 
     def test_get_redirect_url_max_hops_exceeded(self, mocker):
-        url = "http://www.vision.caltech.edu/visipedia-data/CUB-200-2011/CUB_200_2011.tgz"
-        redirect_url = "https://drive.google.com/file/d/1hbzc_P1FuxMkcabkgn9ZKinBwW683j45/view"
+        url = "https://url.org"
+        redirect_url = "https://redirect.url.org"
 
         patch_url_redirection(mocker, redirect_url)
 
--- a/test/test_datasets_utils.py
+++ b/test/test_datasets_utils.py
@@ -10,6 +10,7 @@ from torch._utils_internal import get_file_path_2
 from urllib.error import URLError
 import itertools
 import lzma
+import contextlib
 
 from common_utils import get_tmp_dir
 from torchvision.datasets.utils import _COMPRESSED_FILE_OPENERS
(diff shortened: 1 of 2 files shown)


test/test_internet.py
```diff
@@ -9,22 +9,61 @@
...
+        class PatchedOpener:
+            def __init__(self, request_or_url, *args, **kwargs):
+                self._request_or_url = request_or_url
+
+            def __enter__(self):
+                url = (
+                    self._request_or_url.full_url
+                    if isinstance(self._request_or_url.full_url, Request)
+                    else self._request_or_url.full_url
+                )
```
NicolasHug, line 34: Looks like this is `self._request_or_url.full_url` in all cases?
Also, maybe we can slightly simplify some logic consididering we know we'll always pass an URL, not a Request object?

test/test_internet.py
```diff
@@ -9,22 +9,61 @@
...
 import warnings
 from urllib.error import URLError
+from urllib.request import Request
 
 import torchvision.datasets.utils as utils
 from common_utils import get_tmp_dir
 
 
-class TestDatasetUtils:
+@pytest.fixture
```
NicolasHug, line 18: do we need it to be a fixture so that we can use `mocker`?

test/test_internet.py
```diff
@@ -9,22 +9,61 @@
...
+def patch_url_redirection(mocker):
+    class Response:
```
NicolasHug, line 20: could we use the Response class from requests (if there's such a thing?)

test/test_internet.py
```diff
@@ -9,22 +9,61 @@
...
+                url = (
+                    self._request_or_url.full_url
+                    if isinstance(self._request_or_url.full_url, Request)
+                    else self._request_or_url.full_url
+                )
+
+                if url == urls[-1]:
+                    redirect_url = url
+                else:
+                    redirect_url = urls[urls.index(url) + 1]
```
NicolasHug, line 39: looks like we only ever call this with 2 urls. Could we simplify this to reflect that fact? Also above maybe we could just be `def factory(url, redirect_url):`?

NicolasHug (commented):
Thanks @pmeier , looks good, I'm trying to see if we could simplify things a bit :)

test/test_internet.py
pmeier: Yeah sorry. I have a habit of making stuff more generic and in turn complicated in places where it is not needed.

test/test_internet.py
pmeier: Yes. We could also define a regular function and pass `mocker` though. No strong opinion here.

test/test_internet.py
pmeier: Same as https://github.com/pytorch/vision/pull/4197#discussion_r673915005.

test/test_internet.py
pmeier: Apart from the [docs stating](https://docs.python.org/3/library/http.client.html#http.client.HTTPResponse):

> Not instantiated directly by user.

we do not need anything apart from the `url` attribute. Thus, in the spirit of the other suggestions, I would keep it a custom object until we need something more.

pmeier: And if you mean the `requests` package by

> Response class from requests

than no. Since we are using `urllib` here, classes from `requests` will be as foreign as a custom object.

NicolasHug: I got a bit mixed up, I meant use the class which is returned by `urllib.request.urlopen`. But if that class is any "harder" to instanciate than the custom `Response` you wrote, I agree it's not worth it

test/test_internet.py
```diff
@@ -9,22 +9,40 @@
...
+
 
-    def test_get_redirect_url(self):
+class TestDatasetUtils:
+    def test_get_redirect_url(self, mocker):
         url = "http://www.vision.caltech.edu/visipedia-data/CUB-200-2011/CUB_200_2011.tgz"
-        expected = "https://drive.google.com/file/d/1hbzc_P1FuxMkcabkgn9ZKinBwW683j45/view"
+        expected_redirect_url = "https://drive.google.com/file/d/1hbzc_P1FuxMkcabkgn9ZKinBwW683j45/view"
+
+        patch_url_redirection(mocker, expected_redirect_url)
```
NicolasHug, line 35: If I understand the logic correctly, should we now ensure that the mocker/urlopen was called exactly twice, once with the first URL and once with the second URL?

Also, since we don't need to use real URLs anymore, perhaps we could just create obviously fake ones like `url = "some_url"` and `expected_redirect_url = "some_other_url"`?

pmeier: > If I understand the logic correctly, should we now ensure that the mocker/urlopen was called exactly twice, once with the first URL and once with the second URL?

I can add that, yes.

> Also, since we don't need to use real URLs anymore, perhaps we could just create obviously fake ones like `url = "some_url"` and `expected_redirect_url = "some_other_url"`?

Good point.

NicolasHug:
BTW now that these tests don't require an internet connection, we can take them away from this file and put back where they were prior to https://github.com/pytorch/vision/pull/4032 :)

test/test_internet.py
pmeier: After some more thought, I think we should not test the mocked `urlopen` further.

1. Since the call actually returns a context manager, we would need to create our custom mock instead of relying on the `new` kwarg. AFAIK there is no easy way to do it.
2. We shouldn't really care. As long as `_get_redirect_url` returns the correct URL, we shouldn't really care about the internals.

NicolasHug: We can do it by replacing `new` with `side_effect`, this will return a MagicMock (with a call_count attribute that we can validate) instead of the `patched_opener`  function:

```patch
@ test/test_datasets_utils.py:32 @ def patch_url_redirection(mocker, redirect_url):
    def patched_opener(*args, **kwargs):
        yield Response(redirect_url)

-    return mocker.patch("torchvision.datasets.utils.urllib.request.urlopen", new=patched_opener)
+    return mocker.patch("torchvision.datasets.utils.urllib.request.urlopen", side_effect=patched_opener)


class TestDatasetsUtils:
@ test/test_datasets_utils.py:40 @ class TestDatasetsUtils:
        url = "https://url.org"
        expected_redirect_url = "https://redirect.url.org"

-        patch_url_redirection(mocker, expected_redirect_url)
+        mocked = patch_url_redirection(mocker, expected_redirect_url)

        actual = utils._get_redirect_url(url)
+      assert mocked.call_count == 2
        assert actual == expected_redirect_url
```

Regarding 2: I would agree that we shouldn't have to care about the internals, but I feel like we're already concerned with internals by creating a Response object with a url attribute. So we might as well fully validate the logic, IMHO.

Anyway, no strong opinion, we can also merge as-is :)

test/test_datasets_utils.py
```diff
@@ -19,7 +20,36 @@
...
+def patch_url_redirection(mocker, redirect_url):
+    class Response:
+        def __init__(self, url):
+            self.url = url
+
+    @contextlib.contextmanager
+    def patched_opener(*args, **kwargs):
+        yield Response(redirect_url)
+
+    return mocker.patch("torchvision.datasets.utils.urllib.request.urlopen", new=patched_opener)
```
NicolasHug, line 32: nit: if we don't validate the mocked function, we don't need a return here

NicolasHug (approved):
Thanks a lot!

PR #4200, opened 2021-07-21
Ignore direnv files

garymm:
See https://direnv.net/.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,7 @@ gen.yml
 
 ## Xcode User settings
 xcuserdata/
+
+# direnv
+.direnv
+.envrc


NicolasHug:
Thanks for the PR @garymm 

Would it be feasible for you to rely on a custom set of gitignore rules instead? Something like https://stackoverflow.com/questions/6117109/is-it-possible-to-have-a-custom-gitignore-read-only-access

garymm:
Seems like widely used tools are already included in .gitignore (e.g. vscode files).
But please close this PR if this is unacceptable and I'll try the work-around.

NicolasHug:
vscode is the default at fb so that's why it's here. But yeah there is stuff like .idea/ .orig and *.venv that are more user specific (and that shouldn't be here IMHO, but OK, let's merge)

NicolasHug:
tagging it as CI for lack of better tag

garymm:
Thanks!

Issue #4112, opened 2021-06-24
UCF101: Dataloader Fail on assertion


YanCote:
## 🐛 Bug

When loading UCF101 with different value of frames_per_clip and step_between_clips, it often yield frames_per_clip + 1 images which results in the following assertion to fail:
	
```
assert len(video) == self.num_frames, "{} x {}".format(
    video.shape, self.num_frames
```

## Code To Reproduce

```
import torch
from torchvision.datasets import UCF101

ucf_loc = "/dataset/ucf"
ucf_data_dir = f"{ucf_loc}/UCF101/UCF-101"
ucf_label_dir = f"{ucf_loc}/ucfTrainTestlist"
frames_per_clip = 5
step_between_clips = 1
num_workers = 4

def custom_collate(batch):
    filtered_batch = []
    for video, _, label in batch:
        filtered_batch.append((video, label))
    return torch.utils.data.dataloader.default_collate(filtered_batch)

if __name__ == '__main__':
    # create train loader (allowing batches and other extras)
    test_dataset = UCF101(ucf_data_dir, ucf_label_dir, frames_per_clip=frames_per_clip,
                           step_between_clips=step_between_clips, train=False, transform=None, num_workers=num_workers)
    test_loader = torch.utils.data.DataLoader(test_dataset, batch_size=8, shuffle=True,
                                               collate_fn=custom_collate,num_workers=num_workers)

    for i, (video, label) in enumerate(test_loader):
        print(video.size())
        print(label)
```

### Steps to reproduce the behavior:

1. Download [UCF101](https://www.crcv.ucf.edu/data/UCF101.php)
2. Modify ucf_loc and run the code included above

### stack trace

Original Traceback (most recent call last):
  File "/opt/miniconda3/envs/p38_ucf/lib/python3.8/site-packages/torch/utils/data/_utils/worker.py", line 287, in _worker_loop
    data = fetcher.fetch(index)
  File "/opt/miniconda3/envs/p38_ucf/lib/python3.8/site-packages/torch/utils/data/_utils/fetch.py", line 44, in fetch
    data = [self.dataset[idx] for idx in possibly_batched_index]
  File "/opt/miniconda3/envs/p38_ucf/lib/python3.8/site-packages/torch/utils/data/_utils/fetch.py", line 44, in <listcomp>
    data = [self.dataset[idx] for idx in possibly_batched_index]
[...]

PR #4204, opened 2021-07-22
UCF101 Sketchy Fix 

bjuncek:
Providing a quick-fix for #4112 by exposing `pts_unit` to videoclips object.


prabhat00155 (commented):
Thanks @bjuncek. Can you add a unit test as well?

PR #3143, opened 2020-12-08
Removed all backward methods from header files

datumbox:
As discussed in #3135 keeping the backward methods public is unnecessary, so this PR removes them.

PR #2926, opened 2020-10-28
PSROIPool + Dispatcher + Autocast + Code Cleanup

datumbox:
part of #2797 and #2796

Implements Dispatcher + Autocast for PSROIPool.

I'm also doing some clean up on the original code such as:
- Remove primitive const declaration from method names.
- Using references when possible.
- Fix variable naming.

PR #2898, opened 2020-10-26
Port DeformConv to use the Dispatcher and support Autocast

datumbox:
Fixes #2833 

part of #2797 and #2796

PR #2366, opened 2020-06-29
Port roi_align to actually use dispatcher

ezyang:

PR #4205, opened 2021-07-23
Remove unnecessary dispatcher/registration code

NicolasHug:
I don't really understand the initial goal for these registrations, so I might be missing something. But it looks like we can remove them.

### DIFF
--- a/torchvision/csrc/ops/autograd/deform_conv2d_kernel.cpp
+++ b/torchvision/csrc/ops/autograd/deform_conv2d_kernel.cpp
@@ -121,67 +121,6 @@ class DeformConv2dFunction
   }
 };
 
-// TODO: There should be an easier way to do this
-class DeformConv2dBackwardFunction
-    : public torch::autograd::Function<DeformConv2dBackwardFunction> {
- public:
-  static torch::autograd::variable_list forward(
-      torch::autograd::AutogradContext* ctx,
-      const torch::autograd::Variable& grad,
-      const torch::autograd::Variable& input,
-      const torch::autograd::Variable& weight,
-      const torch::autograd::Variable& offset,
-      const torch::autograd::Variable& mask,
-      const torch::autograd::Variable& bias,
-      int64_t stride_h,
-      int64_t stride_w,
-      int64_t pad_h,
-      int64_t pad_w,
-      int64_t dilation_h,
-      int64_t dilation_w,
-      int64_t groups,
-      int64_t offset_groups,
-      bool use_mask) {
-    at::AutoDispatchBelowADInplaceOrView g;
-    auto result = detail::_deform_conv2d_backward(
-        grad,
-        input,
-        weight,
-        offset,
-        mask,
-        bias,
-        stride_h,
-        stride_w,
-        pad_h,
-        pad_w,
-        dilation_h,
-        dilation_w,
-        groups,
-        offset_groups,
-        use_mask);
-
-    auto grad_input = std::get<0>(result);
-    auto grad_weight = std::get<1>(result);
-    auto grad_offset = std::get<2>(result);
-    auto grad_mask = std::get<3>(result);
-    auto grad_bias = std::get<4>(result);
-
-    return {
-        grad_input,
-        grad_weight,
-        grad_offset,
-        grad_mask,
-        grad_bias,
-    };
-  }
-
-  static torch::autograd::variable_list backward(
-      torch::autograd::AutogradContext* ctx,
-      const torch::autograd::variable_list& grad_output) {
-    TORCH_CHECK(0, "double backwards on deform_conv2d not supported");
-  }
-};
-
 at::Tensor deform_conv2d_autograd(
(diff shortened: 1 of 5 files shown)


datumbox (changes requested):
These registrations should not be removed to allow C++ programs access the backwards passes. Please check the original PRs where I introduced them for more context why they are necessary.

NicolasHug:
> These registrations should not be removed to allow C++ programs access the backwards passes

Thanks for the context. I think this should be documented, and at the very least it should be tested against, if this is a feature we want to explicitely support. Such doc / testing would have saved me a lot of time.

Also, I might be missing something but it seems somewhat contradictory with this PR https://github.com/pytorch/vision/pull/3143 were it is said that we don't want the backward passes to be public.

> Please check the original PRs where I introduced them for more context why they are necessary

I did, but could not find anything relevant. Would you mind pointing me towards those?

NicolasHug:
For example this was added in https://github.com/pytorch/vision/pull/2926 for  `ps_roi_align`, but I can't find any reference to this in the PR or in the related issues. Same for DeformConv in https://github.com/pytorch/vision/pull/2898.

datumbox:
@NicolasHug Unfortunately I got limited internet access at the moment and it's a bit hard to find all the PRs/issues where it was decided to support backwards. I just saw that this PR removes lots of code which we probably want to keep and wanted to flag this so that we don't accidentally merge and break stuff. If @fmassa is around, perhaps he can give you the background easier.
 
Concerning documentation and testing, that's definitely an area that the C++ codebase can be improved. The aforementioned PRs targeted upgrading the ops code to use the latest dispatcher mechanism, so we used the existing testing infra. Feel free to propose ways to improve the code.

fmassa:
I had a chat with Nicolas about this earlier today.

Those functions used to be there for properly supporting double-backwards, see https://github.com/pytorch/vision/pull/2366/files#r447547554 for more context.

Given that we don't provide a double-backwards implementation for our operators, it should be possible for PyTorch to automatically register this function somehow. That's why there is this `TODO` in the code.

Except if something changed in PyTorch since last year, I think those functions might be needed (at least for double-backwards), although I would have expected that we could have compilation errors without them.

cc @ezyang for more insights on if we still need those for double-backwards or if PyTorch already handles it for us

NicolasHug:
It's highly likely that I'm missing something but the only way I'm able to hit the "double backward on ... not supported" on `master` is by doing:

```py
import torch
from torch.autograd import grad
import torchvision
from torchvision.ops import roi_align

t = torch.rand(2, 3, 4, 5, requires_grad=True)
boxes = torch.rand(2, 5, requires_grad=True)

out = roi_align(t, boxes, output_size=1)

grad_t = grad(out.sum(), t, create_graph=True)[0]
grad(grad_t.sum(), boxes, allow_unused=True)
```

i.e. asking for `d/dboxes (droi_align / dt)`, which doesn't seem to make much sense.

If we ask for `d/dt (droi_align/dt)`, which seems to be a more likely use-case:

```py
grad_t = grad(out.sum(), t, create_graph=True)[0]
grad(grad_t.sum(), t, allow_unused=True)
```

we get `None`.

In this PR's branch, however, both second-order derivatives will fail with:

```
RuntimeError: element 0 of tensors does not require grad and does not have a grad_fn
```

If the original goal was to avoid a segfault in such cases, maybe we could still remove this code: getting an error might actually be better than getting `None`?

ezyang:
OK, so there's like, a bunch of stuff here.

First, you don't get an error for `d/dt (droi_align/dt)` because you passed `allow_unused=True`. Of course if you don't pass that kwarg, you get the error that "One of the differentiated Tensors appears to not have been used in the graph."

Second, this error is correct. The derivative for roi_align has no dependence on the input in question. Intuitively, you can see this is the case because all roi_align forwards is doing is indexing (OK, interpolating, really) out points from within the RoI. The gradient contribution to these points is independent of what the values in the points are (which would have been the case if we were doing a multiply, e.g.). So of course autograd is going to complain in this case.

Third, while it's true d/dboxes (droi_align / dt) probably doesn't make too much sense by itself, my favorite "ordinary" use of second order derivatives is gradient penalty. So imagine a standard use of MaskRCNN, but with gradient penalty (let's forget, for a moment, whether or not gradient penalty is actually useful in this case). The loss will involve computations from roi_align, which will in turn rely on bounding boxes from the Region Proposal Network, which itself contains a bunch of weights. To penalize large gradients on the weights in the RPN, we would have to differentiate through roi_align with respect to dboxes.

Finally, when @soulitzer nails https://github.com/pytorch/pytorch/issues/62032 we will be allowed to get rid of this redundant code; this is not done yet, however!

NicolasHug:
Thanks for the details @ezyang  and @datumbox @fmassa  for your input.
I'll close this PR then, and I'll keep an eye on https://github.com/pytorch/pytorch/issues/62032 to properly get rid of these once we have a better support mechanism.

PR #4208, opened 2021-07-26
[1/2] Added backward pass on CPU for interpolation with anti-alias option

vfdev-5:
Description:
- Added backward pass on CPU for interpolation with anti-alias option
  - bilinear 2d
  - bicubic 2d
- Added tests on CPU only
- Renamed internal forward methods to reflect "2d" only implementation
- Fixed int64 -> int in CUDA forward pass kernel

cc @fmassa

### DIFF
--- a/test/test_functional_tensor.py
+++ b/test/test_functional_tensor.py
@@ -1,3 +1,4 @@
+from functools import partial
 import itertools
 import os
 import colorsys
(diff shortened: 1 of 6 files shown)


test/test_functional_tensor.py
```diff
@@ -578,6 +579,52 @@ def test_assert_resize_antialias(interpolation):
...
+def test_interpolate_antialias_backward(dt, size, interpolation):
+
+    # temporarily hard-code device as CPU, CUDA support will be done later
+    device = "cpu"
+
+    if dt == torch.float16 and device == "cpu":
+        # skip float16 on CPU case
+        return
+
+    torch.manual_seed(12)
```
fmassa, line 594: In general, should we set the seed of the tests in a more automated way via pytest?

test/test_functional_tensor.py
```diff
@@ -578,6 +579,52 @@ def test_assert_resize_antialias(interpolation):
...
+            result = forward_op(i, size, False)
+            ctx.save_for_backward(i, result)
+            return result
+
+        @staticmethod
+        def backward(ctx, grad_output):
+            i, result = ctx.saved_tensors
+            ishape = i.shape
+            oshape = result.shape[2:]
+            return backward_op(grad_output, oshape, ishape, False)
```
fmassa, line 615: This is fine for now. I suppose the next step would be to move those functions to PyTorch so that we nave native autograd support?

torchvision/csrc/ops/cpu/interpolate_aa_kernels.cpp
```diff
@@ -475,6 +498,151 @@ void _ti_upsample_bicubic2d_kernel_impl(
...
+              grad_output_data[c * output_slice_size + oh * output_width + ow];
+
+          for (size_t y = 0; y < ysize; y++) {
+            for (size_t x = 0; x < xsize; x++) {
+              *input_indexr(c, ymin + y, xmin + x) +=
+                  wx[x] * wy[y] * grad_output_value;
+            }
+          }
+        }
+      }
```
fmassa, line 603: This implementation is good and follows what we have discussed previously (i.e., follows the backwards op in PyTorch's interpolate).

I would like you to think if there would be a way of generalizing this for nd cases that doesn't involve a lot of copy-pasting. This will be specially important when we move the function to PyTorch, as we will also support 3d and 5d cases.

fmassa (approved):
Implementation looks good to me, thanks!

Some parts are not yet binded (like autograd for the ops), but I suppose this can be left for the future.

test/test_functional_tensor.py
NicolasHug: It's fine for now as this is what we're doing almost everywhere else
Once we're finished with pytest porting I'll look into ways to improve the RNG handling in our tests.
One thing I'm wondering is: does `torch.manual_seed(12)` leak the rng for the rest of the tests?

vfdev-5: > does torch.manual_seed(12) leak the rng for the rest of the tests?

I think yes

NicolasHug: that's pretty bad :)

but yeah, it's OK for now to use the old pattern.

datumbox: it certainly does leak it and it's bitten us quite a few times in the past.

Issue #3945, opened 2021-06-02
Port test/test_transforms.py to pytest


NicolasHug:
Currently, most tests in [test/test_transforms.py](https://github.com/pytorch/vision/blob/master/test/test_transforms.py) rely on `unittest.TestCase`. Now that we support `pytest`, we want to remove the use of the `unittest` module.

For a similar issue see https://github.com/pytorch/vision/issues/3956

### Instructions

There are many tests in this file, so I bundled them in multiple related groups below. If you're interested in working on this issue, please comment below with "I'm working on `<group X>`, `<group Y>`, etc..." so that others don't pick the same tests as you do. Feel free to pick as many groups as you wish, but please don't submit more than 2 groups per PR in order to keep the reviews manageable. Before picking a group, make sure it wasn't picked by another contributor first. Thanks!!

### How to port a test to pytest

Porting a test from `unittest` to pytest is usually fairly straightforward. For a typical example, see https://github.com/pytorch/vision/pull/3907/files:

- take the test method out of the  `Tester(unittest.TestCase)` class and just declare it as a function
- Replace `@unittest.skipIf` with `pytest.mark.skipif(cond, reason=...)`
- remove any use of `self.assertXYZ`.
  - Typically `assertEqual(a, b)` can be replaced by `assert a == b` when a and b are pure python objects (scalars, tuples, lists), and otherwise we can rely on `assert_equal` which is already used in the file.
  - `self.assertRaises` should be replaced with the `pytest.raises(Exp, match=...):` context manager, as done in https://github.com/pytorch/vision/pull/3907/files. Same for warnings with `pytest.warns`
  - `self.assertTrue` should be replaced with a plain `assert`
- When a function uses for loops to tests multiple parameter values, one should use`pytest.mark.parametrize` instead, as done e.g. in https://github.com/pytorch/vision/pull/3907/files.
- It may make sense to keep related tests within a single class. For example here, the tests in group A could be grouped into a `TestToPILImage` class, the tests in group N could be in `TestPad`, etc. Not all groups need a dedicated class though, it's on a case-by-case basis.

CC @ShrillShrestha and @zhiqwang in case you're interested in this too :)

-----

### Groups:

- [x] Groups A: https://github.com/pytorch/vision/pull/3959
  - `test_1_channel_ndarray_to_pil_image`
  - `test_1_channel_tensor_to_pil_image`
  - `test_2_channel_ndarray_to_pil_image`
  - `test_2_channel_tensor_to_pil_image`
[...]

PR #4215, opened 2021-07-28
Port test/test_datasets.py to use pytest

sallysyw:
Following the discussion in [#3945](https://github.com/pytorch/vision/issues/3945), we should also port `test/test_dataset` to use `pytest`.

In this PR, we replace all the `self.assertXYZ` with python `assert`.

TODO: Replace the `self.subTest()` in these files and finally remove the `unittest.TestCase` inheritance.

### DIFF
--- a/test/test_datasets.py
+++ b/test/test_datasets.py
@@ -178,7 +178,7 @@ class Caltech101TestCase(datasets_utils.ImageDatasetTestCase):
                 actual = type(combined_target)
                 expected = type(individual_target)
                 assert (
-                    actual == expected
+                    actual is expected
                 ), "Type of the combined target does not match the type of the corresponding individual target: "
                 f"{actual} is not {expected}",
 
@@ -580,7 +580,7 @@ class CelebATestCase(datasets_utils.ImageDatasetTestCase):
                 actual = type(combined_target)
                 expected = type(individual_target)
                 assert (
-                    actual == expected
+                    actual is expected
                 ), "Type of the combined target does not match the type of the corresponding individual target: "
                 f"{actual} is not {expected}",
 
--- a/test/test_datasets.py
+++ b/test/test_datasets.py
@@ -16,6 +16,7 @@ import zipfile
 import PIL
 import datasets_utils
 import numpy as np
+import pytest
 import torch
 import torch.nn.functional as F
 from torchvision import datasets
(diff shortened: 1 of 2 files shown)


test/test_datasets.py
```diff
@@ -167,23 +168,19 @@ def test_combined_targets(self):
...
                 actual = type(combined_target)
                 expected = type(individual_target)
-                self.assertIs(
-                    actual,
-                    expected,
-                    f"Type of the combined target does not match the type of the corresponding individual target: "
-                    f"{actual} is not {expected}",
-                )
+                assert (
+                    actual == expected
```
NicolasHug, line 181: according to the original test, I think this should be

```suggestion
                    actual is expected
```

test/test_datasets.py
```diff
@@ -573,33 +570,29 @@ def test_combined_targets(self):
...
```
NicolasHug, line 583: here as well, this should be an identity check instead of an equality check

NicolasHug (approved):
Thanks @sallysyw , contrags on your first torchvision PR!
This looks great, there are just 2 assertions that need a minor change. I'll approve anyway, feel free to self-merge once the changes are addressed

NicolasHug:
Indeed the next step will be to fully remove the use of `unittest.TestCase` and thus remove the `subTest` parts, which I think is fine.

It'd be nice to change the `test_all_configs` decorator into a pytest parametrization. @pmeier do you foresee any major complication in doing that?

test/test_datasets.py
sallysyw: Thanks for pointing it out!

sallysyw:
Some CI tests are failing but they don't seem related - is it safe to merge? @NicolasHug

pmeier:
> It'd be nice to change the test_all_configs decorator into a pytest parametrization. @pmeier do you foresee any major complication in doing that?

IIRC, you can't use `pytest.mark.parametrize` on the test methods of a `unittest.TestCase`. So we would also need to change that class. But then, we are currently using the `TestCase.setUpClass` method

https://github.com/pytorch/vision/blob/b29ed34f46ba328b46f3aadc5ac15f2a9417a3b0/test/datasets_utils.py#L387-L392

which we would have to emulate another way.

NicolasHug:
> IIRC, you can't use pytest.mark.parametrize on the test methods of a unittest.TestCase

That's correct, and getting rid of `unittest` altogether is our end goal here :). Regarding the setup, we'll probably just have to rename it into a `setup_class` method instead: https://docs.pytest.org/en/6.2.x/xunit_setup.html#class-level-setup-teardown

PR #4216, opened 2021-07-28
Add utility to draw keypoints

### LINKED ISSUE #3365: Add utilites to plot Keypoints
## 🚀 Feature

We added utilities to plot bounding boxes #2556 #2785 and utility to plot semantic segmentation masks #3272 #3330 #3820
I think this is probably the last plotting utility we need.

## Motivation
Same as before, we need these to make visualization and post-processing easier.
Also Easily support torchvision models.

## Pitch
Use only PIL to draw key points. Also, keep the function signature consistent with previous utilities.

A prototype implementation is here!

```
@torch.no_grad()
def draw_keypoints(
    image: torch.Tensor,
    keypoints: torch.Tensor,
    colors: Optional[List[Union[str, Tuple[int, int, int]]]] = None,
    radius: Optional[int] = 5,
    connect : Optional[bool] = False,
    plot_ids Optional[bool] = False, 
    font: Optional[str] = None,
    font_size: int = 10
) -> torch.Tensor:

"""
Args:
        image (Tensor): Tensor of shape (C x H x W) and dtype uint8.
        keypoints (Tensor): The ketpoints from Keypoint RCNN
        colors (List[Union[str, Tuple[int, int, int]]]): List containing the colors of keypoints. The colors can
            be represented as `str` or `Tuple[int, int, int]`.
       radius (int): Radius of circles drawn for keypoints.
       connect (bool): If True connects the visible keypoints.
       plot_ids (bool): If True plots the visible keypoints ids on the image.
        font (str): A filename containing a TrueType font. If the file is not found in this filename, the loader may
            also search in other directories, such as the `fonts/` directory on Windows or `/Library/Fonts/`,
            `/System/Library/Fonts/` and `~/Library/Fonts/` on macOS.
        font_size (int): The requested font size in points.
"""

```
Radius to determine how big the key point should be. This is essential as torchvision models suppose flexible image sizes.
The font is to plot the keypoint id. Sometimes in models like PoseNet, we would like to know which keypoint ids are visible.
Might help in debugging if particular keypoint does not work well.

The output of Keypoint RCNN is as follows

```
boxes (FloatTensor[N, 4]): the ground-truth boxes in [x1, y1, x2, y2] format,
with values of x between 0 and W and values of y between 0 and H

labels (Int64Tensor[N]): the class label for each ground-truth box

keypoints (FloatTensor[N, K, 3]): the K keypoints location for each of the N instances,
in the format [x, y, visibility], where visibility=0 means that the keypoint is not visible.

```

I think we should plot only the keypoints which are visible,

## Additional context

The boxes can be plotted using `draw_bounding_boxes`. Users can use both these utils together to plot both boxes and keypoints. I don't think so `draw_keypoints` should internally call / draw boxes.

Probably we need to give a demo run at it

oke-aditya:
Closes #3365 

Finally! (everything went well in my life!) I hope I can complete the implementation soon!
I will post the outputs of completed prototype here!

- [x] Adds code
- [x] Adds docs
~~- [ ] Adds example to gallery.~~ Next PR
- [x] Adds tests


oke-aditya:
Here is the first dry run! This is super dirty and possibly needs few more iterations :smile: 

I'm trying to replicate something similar to [this](https://colab.research.google.com/github/tugstugi/dl-colab-notebooks/blob/master/notebooks/TorchvisionPersonKeypoint.ipynb) colab notebook.

Keypoint outputs are slightly tricky! Note that the outputs of keypoint detection models are

`keypoints (FloatTensor[N, K, 3])`: the locations of the predicted keypoints, in [x, y, v] format.
N -> Number of Keypoints
K -> Each keypoint output
3 -> X, Y, Visibility.

Just like utiltities. This utlity too draws all the possible keypoints on the image. 
Threshold needs to be processed by the user.

Here are some of the current outputs.

(Image 1)

(Indian road :heart: )

example3


(Some Image taken from Pascal VOC examples)


I still need to implement plotting labels and some functionality to join the points (plot lines with connections)
I'm not sure if color can be provided by the user or we should generate from a color palette.

I have kept the label option, since some users might be interested in labeling the keypoints, (E.g. by the keypoint numbers or by names).

A review at this point would be great!

cc @datumbox @NicolasHug

torchvision/utils.py
```diff
@@ -300,6 +301,70 @@ def draw_segmentation_masks(
     return out.to(out_dtype)
 
 
+@torch.no_grad()
+def draw_keypoints(
+    image: torch.Tensor,
+    keypoints: torch.Tensor,
+    labels: Optional[List[str]] = None,
```
fmassa, line 308: Instead of providing `labels` (which, for the COCO dataset that we consider we only have one category), I would instead provide a `connectivity` argument, which dictates which points should be linked against which other point.

This can be an optional argument, and if provided will connect the different keypoints together (and could thus replace the `connect` boolean argument as well.

Here is an implementation that I had [for it for the "person" class](https://github.com/facebookresearch/maskrcnn-benchmark/blob/57eec25b75144d9fb1a6857f32553e1574177daf/demo/predictor.py#L413-L471)

torchvision/utils.py
```diff
@@ -300,6 +301,70 @@ def draw_segmentation_masks(
...
+    font_size: int = 10
+) -> torch.Tensor:
+
+    """
+    Draws Keypoints on given RGB image.
+    The values of the input image should be uint8 between 0 and 255.
+
+    Args:
+        image (Tensor): Tensor of shape (3, H, W) and dtype uint8.
+        keypoints (Tensor): Tensor of shape (num_keypoints, K, 3) the K keypoints location for each of the N instances,
```
fmassa, line 322: instead of `num_keypoints`, I would maybe call it `num_instances`, as it can be confusing compared to `K` (which is the number of keypoints per instance)

Also, the `3` here is a bit weird, as we don't use it for now. Maybe I would just check that the tensor has at least 2 dimensions, and the code ignores everything after the 2nd dimension. Thoughts?

torchvision/utils.py
```diff
@@ -300,6 +301,70 @@ def draw_segmentation_masks(
...
+        labels (List[str]): List containing the labels for each Keypoint.
+        colors (Union[List[Union[str, Tuple[int, int, int]]], str, Tuple[int, int, int]]): List containing the colors
+            or a single color for all the keypoints.
+            The colors can be represented as `str` or `Tuple[int, int, int]`.
+        radius (int): Integer denoting radius of keypoint.
+        connect (bool): If True. It connects all the visible keypints.
+        font (str): A filename containing a TrueType font. If the file is not found in this filename, the loader may
+            also search in other directories, such as the `fonts/` directory on Windows or `/Library/Fonts/`,
+            `/System/Library/Fonts/` and `~/Library/Fonts/` on macOS.
+        font_size (int): The requested font size in points.
```
fmassa, line 333: If we don't provide labels, we don't need to provide font either

fmassa (commented):
Sorry for missing this PR @oke-aditya !

I've made a few comments, let me know what you think.

torchvision/utils.py
oke-aditya: Hi @fmassa . Thanks a lot for review :smiley: 

The idea behind labels is if person wants to explicitly mark the keypoint number.
Particularly useful in counting the keypoints, or validating performance of all the keypoints. (Maybe a keypoint number X is performing very poorly? User would need to label these to come to know.

Also it might be particularly useful in face detection etc models. 

image

Maybe users want to do something similar to above. 

About the connectivity point, I think you are right. connectivity can be `Optional[Tuple[Tuple[int, int]]]` denoting which two points should be connected. So that we can link them up if provided.

I think the utility should be slightly generic and not limited to COCO. Let me know what you think

The person implementation is very nice. (maskrcnn-benchmark is probably one of the best libraries)
I think I can post images side by side how maskrcnn implementation and the current one (when finalized a bit more) looks like.

torchvision/utils.py
oke-aditya: Sure, it is `num_instances`. I will change that.

3 is bit weird, but that makes the API simple, users can directly pass (X, Y, visibility) the format of torchvision keypoints.

We can do visibility based thresholding, to use the third dimension. (this might restrict the API)
Or simply discard as you suggested and leave the above part to end-user.

Let me know your thoughts. I will refactor accordingly.

oke-aditya:
Bonjour @fmassa !

I hope I have implemented the basic functionality as per requirements.
The keypoints can be connected by a Optional connectivity arg which is `Tuple[Tuple[int, int]]`. Each tuple denoting the 
keypoint ids to be connected.

There are two small issues

1. Colors: -

Well color palette is always the issue with visualization utils and it's back here too!
We need colors for keypoints as well as lines to join them (if connectivity is True) 

For keypoints colors. Should we
By default generate unique colors for every keypoint.
    or
 Allow users to pass into colors as one string for all keypoints as well as colors for each keypoints.

For line colors. Should we
Again randomly generate unique colors for every line
  or
Allow users to pass into colors one string for all lines as well as colors for each line.
  or
Use the same colors that were passed to keypoints

2. Labels: -

As I mentioned earlier, it is useful to keep labels, but again plotting lot of labels can lead to messy image.

3. Visibility: -

We can leave it to user, or have an argument to threshold it.

Here are few outputs! The colors are awful

Figure_1

Figure_2

Figure_3

Figure_4

Figure_5

Figure_6

Would love to hear thoughts from @datumbox and @NicolasHug too :smiley:

oke-aditya:
Code to reproduce.

```
def show(imgs):
    if not isinstance(imgs, list):
        imgs = [imgs]
    fix, axs = plt.subplots(ncols=len(imgs), squeeze=False)
    for i, img in enumerate(imgs):
        img = img.detach()
        img = F.to_pil_image(img)
        axs[0, i].imshow(np.asarray(img))
        axs[0, i].set(xticklabels=[], yticklabels=[], xticks=[], yticks=[])
    plt.show()


model = keypointrcnn_resnet50_fpn(pretrained=True)
model = model.eval()

IMAGE_PATH = 'demo_im6.jpg'
image_tensor2 = read_image(IMAGE_PATH)

image = Image.open(IMAGE_PATH)
image_tensor = to_tensor(image)
output = model([image_tensor])[0]

kpts = output['keypoints']
res = draw_keypoints(image_tensor2, kpts, colors="red", connectivity=((0, 1), (1, 2), (2, 3), (3, 4)))
show(res)

res = draw_keypoints(image_tensor2, kpts, colors="red")
show(res)
```

oke-aditya:
@fmassa can you please have a look. I will have more bandwidth over next few weeks. (I can contribute bit more 😅)

oke-aditya:
Hey @NicolasHug @datumbox, @fmassa  maybe this is missed. :smiley: We can either go ahead or continue on this after 0.11 release?

datumbox:
@oke-aditya I've asked Francisco to pick this up as he is more familiar with the keypoints model. He is currently OOO but he will get back to you.

oke-aditya:
No problem 👍

oke-aditya:
Apologies for excessive pings, but I think @fmassa is back :smiley:

torchvision/utils.py
fmassa: I think for keypoints (given that they are fairly small) labelling them might not be very readable, while using a color-code might be easier to spot. This begs the question if the color should represent the instance or the keypoint-id in the instance.

My 2 cents is that we can pass a color tensor that is of size `(num_instances, num_keypoints, 3)`, and that internally we can broadcast to the full shape if the user only passed `(num_keypoints, 3)` or `(num_instances, 1, 3)` or `(3)`.

Thoughts?

fmassa (commented):
Hi @oke-aditya 

Very sorry for the delay in responding!

I've made a couple of comments.

- I think that the `visibility` shouldn't be taken into account as of now (as we don't even discard invisible keypoints), plus I'm not sure if this type of information is readily available on most datasets (but needs to be verified).
- the colors can be left fairly generic so that we can internally handle both single-color, different color per instance, different color per keypoint, or different color per keypoint-per-instance.

Let me know what you think

torchvision/utils.py
oke-aditya: I agree about visibility. Plus the logic of thresholding / choosing to plot if visible should be left to user.
So we can skip the third dimension.

torchvision/utils.py
oke-aditya: Yes, we do not use labels too for drawing segmentation masks. And a good color combination can be sufficient for visualizing keypoints. 

We don't use tensors for colors, in other utlitites so maybe we shouldn't use here.
I suggest colors to be either "str" (One color for all keypoints) List[str] specifying color for each the keypoints ids. 
None, so that we generate random colors from palette. 

Specifying color per instance might be hard and users would need to dynamically adjust this utlity, as number of detected instances would differ. We don't use num_instances in any other utility and would prefer to stick with keypoint ids.

torchvision/utils.py
oke-aditya: Removing labels as suggested.

torchvision/utils.py
```diff
@@ -300,6 +301,83 @@ def draw_segmentation_masks(
...
+    width: int = 3,
+) -> torch.Tensor:
+
+    """
+    Draws Keypoints on given RGB image.
+    The values of the input image should be uint8 between 0 and 255.
+
+    Args:
+        image (Tensor): Tensor of shape (3, H, W) and dtype uint8.
+        keypoints (Tensor): Tensor of shape (num_instances, K, 2) the K keypoints location for each of the N instances,
```
oke-aditya, line 320: I'm unsure should it be (num_instances, K, 2) or (num_instances, K, 3) or maybe any dimension can be passed.
We aren't using that dimension anyway so it doesn't hurt.

torchvision/utils.py
```diff
@@ -300,6 +301,83 @@ def draw_segmentation_masks(
...
+
+    for kpt_inst in img_kpts:
+        for kpt in kpt_inst:
+            x1 = kpt[0] - radius
+            x2 = kpt[0] + radius
+            y1 = kpt[1] - radius
+            y2 = kpt[1] + radius
+
+            if isinstance(colors, str) or isinstance(colors, tuple):
+                draw.ellipse([x1, y1, x2, y2], fill=colors, outline=None, width=0)
```
oke-aditya, line 361: If we decide to pass tuple containing colors from each keypoint. It would be simply the ith index of tuple.

If we decide for colors per instance. It would need refactor.

torchvision/utils.py
```diff
@@ -300,6 +301,83 @@ def draw_segmentation_masks(
...
+
+        if connectivity:
+            for connection in connectivity:
+                start_pt_x = kpt_inst[connection[0]][0]
+                start_pt_y = kpt_inst[connection[0]][1]
+
+                end_pt_x = kpt_inst[connection[1]][0]
+                end_pt_y = kpt_inst[connection[1]][1]
+
+                draw.line(
```
oke-aditya, line 371: What should be the line color? Should there be a parameter for line color? Or should it be same as color of keypoints?
Currently it is white which looks bad.

test/test_utils.py
```diff
@@ -248,5 +258,64 @@ def test_draw_segmentation_masks_errors():
         utils.draw_segmentation_masks(image=img, masks=masks, colors=bad_colors)
 
 
+def test_draw_keypoints_vanilla():
```
oke-aditya, line 261: We just do like boxes, save an image and check after loading if expected image was same as saved image.

oke-aditya (commented):
Few little points to address. It looks almost there to merge before the release :smile:

torchvision/utils.py
fmassa: Let's document it as `(num_instances, K, 2)`

torchvision/utils.py
fmassa: Yes, I propose that we handle this with a slight change.
We convert the colors list to a `Tensor`, perform the proper reshape + broadcast so that `colors` is always `(num_instances, K, 3)` tensor, and then convert it back to a list (if needed).

torchvision/utils.py
fmassa: In the interest of getting this merged soon, let's leave it white for now, and in the future create an issue to enable this to be configurable.

test/test_utils.py
fmassa: Ok with me

torchvision/utils.py
oke-aditya: Sounds good.

torchvision/utils.py
oke-aditya: I need little bit more clarity in colors. I'm assuming we are not going to specify color for every keypoint id.

Say we have 2 instances and 3 keypoints.

So what are choices possible to user?

```
# Specify one single color for all keypoints of all instances
colors = "red"

# Specify one single color for all keypoints of each instance.
colors = ["red", blue"]

# Specify color for every keypoint of every instance.

colors = [["red", "blue", "green"], ["red", "blue", "green"]]

```
Note that string can be replaced by RGB Tuple too.

Am I aligning with your thoughts?

torchvision/utils.py
```diff
@@ -300,6 +301,83 @@ def draw_segmentation_masks(
...
+            x1 = kpt[0] - radius
+            x2 = kpt[0] + radius
+            y1 = kpt[1] - radius
+            y2 = kpt[1] + radius
+
+            if isinstance(colors, str) or isinstance(colors, tuple):
+                draw.ellipse([x1, y1, x2, y2], fill=colors, outline=None, width=0)
+
+        if connectivity:
+            for connection in connectivity:
```
oke-aditya, line 364: One small point I forgot to mention.

connectivity is Tuple[Tuple[int, int]] which assumes that keypoint ids are numbered from `0`. 
To join keypoint no 0 and and 1 we pass `connectivity=((0, 1), )`

Is it fine to use 0 based indexing? or should we use 1 based?

oke-aditya:
Hi!

I gave another run and got the outputs for COCO models.
I have a script ready which can become the gallery example! I think it's good to leave it as a new PR as it will need a more thorough reviewing.

Here are the outputs.

boy_kpt1

Connecting the skeleton with connectivity.

boy_kpt2

dog_kpt


I joined them randomly nothing special

dog_kpt2

Picture Credits to MS COCO images.

The output looks bit bad without line_color, I think we shouldn't keep it white and merge it.

oke-aditya:
@fmassa can you please have a review. :)

torchvision/utils.py
fmassa: Hi @oke-aditya 

This logic is what I proposed. I think using broadcasting is easier to implement, and handle extra cases like if the user wants all the keypoints to have the same color, but they being different colors for each instance.

They would only need to do the following:
```python
# there are N keypoints, and 3 instances
colors = [["red"], ["green"], ["blue"]]
```

The implementation I provided above should already handle those cases without problems, let me know if you disagree.

oke-aditya: Somehow I'm on the edge here. :sweat_smile: The above case is handled by current implementation.
user could do. 

```
# there are N keypoints, and 3 instances.
# The first instance gets red, second gets green and third blue

colors = ["red", "blue", "green"]
```

Probably I should have attached one run with this.
To implement with numpy broadcasting vs this one. I'm not really sure which is easier (I felt current one is easier)

I had few doubts on the numpy implementation

`colors=[["red"], ["green"], ["blue"]]`

Does this mean there are 3 instances with multiple keypoints, and all have same color or does it mean that there are 3 instances and all have just 1 keypoint.
It's a bit ambiguous.

Where as with the current one it's bit straightforward,

`colors=["red", "blue", "green"]` for having 3 instances with same color for all keypoints
`colors=[["red"], ["blue"], ["green"]` for having 3 instances having color with just 1 keypoint.

Is there any edge case the current implementation does not handle well? I feel that either code are fine.

datumbox: FYI, I would love to use this utility on the Multi-pretrained weights example for Keypoint RCNN which is still pending. I don't want to express opinion over what's best here as I'm not too familiar with the KP model but I would love to see this unblocked. Francisco has a ton of experience with the specific model but if Oke has still concerns perhaps it's worth simplifying the utility (for example support 1 colour only) and open an issue to extend to all? Also I would like to see the PR containing an example on the Gallery but I see that's planned for follow up.

fmassa: @oke-aditya 
> Does this mean there are 3 instances with multiple keypoints, and all have same color or does it mean that there are 3 instances and all have just 1 keypoint.

Both are possible with broadcasting :-)

I think my proposal might be the "transposed" of yours. My specification is that we broadcast the colors to be of the same shape as the keypoints (so `[N, k]`). You can call `colors.T` in my proposal (after broadcasting) to make it compatible with yours.

oke-aditya: :sweat_smile: Do you suggest refactoring the current one? Or staying with the same :sweat_smile: 
I will leave the decision to you @fmassa :)  Seems like we have 2 valid and equivalent solutions.

datumbox: @oke-aditya Because it's been over 3 months since this PR has started (and that's not your problem, it's on us; thankfully you are a regular contributor and have understanding!), I think we should wrap it up and merge the potion for which there is consensus. As long as we ensure there will be no BC-breaking modification, we can always revisit it on a follow up PR.

In that direction, I recommend 2 ways to move forward:
1. If you have no major concerns about the API, the user-friendliness etc. Adopt @fmassa's proposal. 
2. If you are not convinced about the proposal, just support a single colour (1 string param) for all keypoints. 

In both cases, I would prefer if you added an example in the gallery folder. You can do it either here or on a follow up shortly after.

Let's wrap this up as this is a great addition and the community is missing out.

oke-aditya: I can go ahead with @fmassa's proposal :) I'm convinced by his idea.  Let me give a run with his idea and should we face not any bugs / problems. We can merge it

I will write up a gallery example in follow up PR very soon. I do have the code ready (which needs slight change for the adopting)
 
I guess there are a few aspects that need to be worked upon which won't be BC Breaking. 
Supporting Line color, not drawing with default white if None is passed (Similar to #4528).
Also should we plot the lines above the circle or vice-versa. I will create a ticket where we can track down these.

torchvision/utils.py
```diff
@@ -300,6 +300,77 @@ def draw_segmentation_masks(
...
+    elif image.size()[0] != 3:
+        raise ValueError("Pass an RGB image. Other Image formats are not supported")
+
+    if keypoints.ndim != 3:
+        raise ValueError("keypoints must be of shape (num_instances, K, 2)")
+
+    ndarr = image.permute(1, 2, 0).numpy()
+    img_to_draw = Image.fromarray(ndarr)
+    draw = ImageDraw.Draw(img_to_draw)
+    out_dtype = torch.uint8
```
datumbox, line 347:
```suggestion
```

torchvision/utils.py
```diff
@@ -300,6 +300,77 @@ def draw_segmentation_masks(
...
+
+                end_pt_x = kpt_inst[connection[1]][0]
+                end_pt_y = kpt_inst[connection[1]][1]
+
+                draw.line(
+                    ((start_pt_x, start_pt_y), (end_pt_x, end_pt_y)),
+                    width=width,
+                )
+
+    return torch.from_numpy(np.array(img_to_draw)).permute(2, 0, 1).to(dtype=out_dtype)
```
datumbox, line 371:
```suggestion
    return torch.from_numpy(np.array(img_to_draw)).permute(2, 0, 1).to(dtype=torch.uint8)
```

datumbox (approved):
@oke-aditya LGTM, thanks!

Concerning the extra colours, I think there are a few options to detect and handle it so that the broadcasting can still work. At any case, I think we can discuss these changes on an issue dedicated to extending the functionality of the util. 

We should add a good example with a proper picture on the gallery to showcase how to call the keypoint model and visualize the predictions. Since the utility requires preprocessing the model output (cleaning up, removing extra columns etc), we should provide a working example that users can follow.

I'll let @fmassa have the final word on this as he has been involved on the review since the beginning.

oke-aditya:
Sure Vasilis ! I do have the code for gallery example ready which works with the keypoint_rcnn model. I will send a follow-up PRa and open up a ticket for additional functionality once this is merged.

fmassa (approved):
LGTM, let's get this merged as is (i.e., without the handling of multiple colors).

We can discuss about multiple colors in an issue.

fmassa:
Thanks for all your work @oke-aditya !

PR #4220, opened 2021-07-29
Add typings to FasterRCNN, MaskRCNN and RetinaNet

oke-aditya:
I'm unsure of few types. I will tag them.

I will go 3 files at a time, otherwise the diff would be too big to review.

cc @pmeier


torchvision/models/detection/retinanet.py
```diff
@@ -374,14 +426,23 @@ def __init__(self, backbone, num_classes,
         self._has_warned = False
 
     @torch.jit.unused
-    def eager_outputs(self, losses, detections):
+    def eager_outputs(
+        self,
+        losses: Dict[str, Tensor],
+        detections: List[Dict[str, Tensor]],
+    ):
```
oke-aditya, line 433: Bit confused here on the type of output.

torchvision/models/detection/retinanet.py
```diff
@@ -455,7 +521,11 @@ def postprocess_detections(self, head_outputs, anchors, image_shapes):
 
         return detections
 
-    def forward(self, images, targets=None):
+    def forward(
+        self,
+        images: List[Tensor],
+        targets: Optional[Dict[str, Tensor]] = None,
+    ):
```
oke-aditya, line 528: Here as well. I'm unsure of type of the output.

torchvision/models/detection/retinanet.py
```diff
@@ -40,19 +40,33 @@ class RetinaNetHead(nn.Module):
...
         self.regression_head = RetinaNetRegressionHead(in_channels, num_anchors)
 
-    def compute_loss(self, targets, head_outputs, anchors, matched_idxs):
+    def compute_loss(
+        self,
+        targets: List[Dict[str, Tensor]],
+        head_outputs: Dict[str, Tensor],
+        anchors: List[Tensor],
+        matched_idxs: List[Tensor],
+    ) -> Dict[Tensor]:
```
oke-aditya, line 59: I did a mistake here, will correct it.

oke-aditya (commented):
Hey @pmeier @frgfm :smile: would be great if you have a look

torchvision/models/detection/faster_rcnn.py
```diff
@@ -141,24 +142,41 @@ class FasterRCNN(GeneralizedRCNN):
...
-                 bbox_reg_weights=None):
+    def __init__(
+        self,
+        backbone: nn.Module,
+        num_classes: Optional[int] = None,
+        # transform parameters
+        min_size: int = 800,
+        max_size: int = 1333,
+        image_mean: Optional[Tuple[float]] = None,
+        image_std: Optional[Tuple[float]] = None,
```
datumbox, line 153: I think it should be `List[float]` instead of `Tuple[float]`?

Similar to:
https://github.com/pytorch/vision/blob/3e7653c8c4b5eb0c854251a0fb86be18de55bc68/torchvision/models/detection/ssd.py#L169

torchvision/models/detection/faster_rcnn.py
```diff
@@ -141,24 +142,41 @@ class FasterRCNN(GeneralizedRCNN):
...
+        box_head: Optional[nn.Module] = None,
+        box_predictor: Optional[nn.Module] = None,
+        box_score_thresh: float = 0.05,
+        box_nms_thresh: float = 0.5,
+        box_detections_per_img: int = 100,
+        box_fg_iou_thresh: float = 0.5,
+        box_bg_iou_thresh: float = 0.5,
+        box_batch_size_per_image: int = 512,
+        box_positive_fraction: float = 0.25,
+        bbox_reg_weights: Optional[Tuple[float]] = None
```
datumbox, line 178: I think this should be either `List[float]` or `Tuple[float, ...]`. See https://docs.python.org/3/library/typing.html#typing.Tuple

Not sure if the latter is supported on JIT; I usually use the first.

torchvision/models/detection/mask_rcnn.py
```diff
@@ -149,26 +150,45 @@ class MaskRCNN(FasterRCNN):
...
-                 mask_roi_pool=None, mask_head=None, mask_predictor=None):
+    def __init__(
+        self,
+        backbone: nn.Module,
+        num_classes: Optional[int] = None,
+        # transform parameters
+        min_size: int = 800,
+        max_size: int = 1333,
+        image_mean: Optional[Tuple[float]] = None,
+        image_std: Optional[Tuple[float]] = None,
```
datumbox, line 161: Same as above.

torchvision/models/detection/mask_rcnn.py
```diff
@@ -149,26 +150,45 @@ class MaskRCNN(FasterRCNN):
...
+        # RPN parameters
+        rpn_anchor_generator: Optional[nn.Module] = None,
```
datumbox, line 163: I believe that this needs to be `Optional[AnchorGenerator]` because of how the parent class works. See:
https://github.com/pytorch/vision/blob/183a722169421c83638e68ee2d8fc5bd3415c4b4/torchvision/models/detection/faster_rcnn.py#L169

torchvision/models/detection/retinanet.py
```diff
@@ -268,6 +310,8 @@ class RetinaNet(nn.Module):
             maps.
         head (nn.Module): Module run on top of the feature pyramid.
             Defaults to a module containing a classification and regression module.
+        proposal_matcher (nn.Module): Module that matches the proposals generated by groudn truth and predictions.
```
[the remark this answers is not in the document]
datumbox: The proposal_matcher is not of type `nn.Module`. It's currently of type object:
https://github.com/pytorch/vision/blob/183a722169421c83638e68ee2d8fc5bd3415c4b4/torchvision/models/detection/_utils.py#L227

Not sure if we could change this on a separate PR first without breaking existing code but worth having a look.

torchvision/models/detection/retinanet.py
```diff
@@ -314,18 +352,26 @@ class RetinaNet(nn.Module):
...
-                 topk_candidates=1000):
+    def __init__(
+        self,
+        backbone: nn.Module,
+        num_classes: int,
+        # transform parameters
+        min_size: int = 800,
+        max_size: int = 1333,
+        image_mean: Optional[Tuple[float]] = None,
+        image_std: Optional[Tuple[float]] = None,
```
datumbox, line 363: Same here.

torchvision/models/detection/retinanet.py
```diff
@@ -314,18 +352,26 @@ class RetinaNet(nn.Module):
...
+        # Anchor parameters
+        anchor_generator: Optional[AnchorGenerator] = None,
+        head: Optional[nn.Module] = None,
+        proposal_matcher: Optional[nn.Module] = None,
```
datumbox, line 367: Same as above.

torchvision/models/detection/retinanet.py
datumbox: Yeah... this is an ugly workaround for JIT. See here for the output type:
https://github.com/pytorch/vision/blob/3e7653c8c4b5eb0c854251a0fb86be18de55bc68/torchvision/models/detection/ssd.py#L216-L217

torchvision/models/detection/retinanet.py
datumbox: Similar to above, see here:
https://github.com/pytorch/vision/blob/3e7653c8c4b5eb0c854251a0fb86be18de55bc68/torchvision/models/detection/ssd.py#L284-L285

torchvision/models/detection/retinanet.py
```diff
@@ -564,8 +624,14 @@ def forward(self, images, targets=None):
...
 
 
-def retinanet_resnet50_fpn(pretrained=False, progress=True,
-                           num_classes=91, pretrained_backbone=True, trainable_backbone_layers=None, **kwargs):
+def retinanet_resnet50_fpn(
+    pretrained: bool = False,
+    progress: bool = True,
+    num_classes: int = 91,
+    pretrained_backbone: bool = True,
+    trainable_backbone_layers: bool = None,
```
datumbox, line 632:
```suggestion
    trainable_backbone_layers: Optional[bool] = None,
```

datumbox (commented):
Thanks @oke-aditya.

I had a first pass and added a few comments. Let me know what you think. Please note that my current setup does not allow me to do thorough reviews, so I will leave it for someone else from the team to approve (or wait until I'm properly back. :) )

torchvision/models/detection/faster_rcnn.py
```diff
@@ -141,24 +142,41 @@ class FasterRCNN(GeneralizedRCNN):
...
+        rpn_post_nms_top_n_test: int = 1000,
+        rpn_nms_thresh: float = 0.7,
+        rpn_fg_iou_thresh: float = 0.7,
+        rpn_bg_iou_thresh: float = 0.3,
+        rpn_batch_size_per_image: int = 256,
+        rpn_positive_fraction: float = 0.5,
+        rpn_score_thresh: float = 0.0,
+        # Box parameters
+        box_roi_pool: Optional[MultiScaleRoIAlign] = None,
+        box_head: Optional[nn.Module] = None,
```
frgfm, line 169: `box_head: Optional[RPNHead] = None` I believe

torchvision/models/detection/mask_rcnn.py
```diff
@@ -149,26 +150,45 @@ class MaskRCNN(FasterRCNN):
...
+        rpn_head: Optional[nn.Module] = None,
```
frgfm, line 164: same here : `Optional[RPNHead]` I think

torchvision/models/detection/faster_rcnn.py
```diff
@@ -141,24 +142,41 @@ class FasterRCNN(GeneralizedRCNN):
...
+        box_predictor: Optional[nn.Module] = None,
```
frgfm, line 170: `Optional[FastRCNNPredictor]` I think

torchvision/models/detection/mask_rcnn.py
```diff
@@ -149,26 +150,45 @@ class MaskRCNN(FasterRCNN):
...
+        rpn_nms_thresh: float = 0.7,
+        rpn_fg_iou_thresh: float = 0.7,
+        rpn_bg_iou_thresh: float = 0.3,
+        rpn_batch_size_per_image: int = 256,
+        rpn_positive_fraction: float = 0.5,
+        rpn_score_thresh: float = 0.0,
+        # Box parameters
+        box_roi_pool: Optional[nn.Module] = None,
+        box_head: Optional[nn.Module] = None,
+        box_predictor: Optional[nn.Module] = None,
```
frgfm, line 178: cf. comments on faster_rcnn

torchvision/models/detection/mask_rcnn.py
```diff
@@ -149,26 +150,45 @@ class MaskRCNN(FasterRCNN):
...
+        box_detections_per_img: int = 100,
+        box_fg_iou_thresh: float = 0.5,
+        box_bg_iou_thresh: float = 0.5,
+        box_batch_size_per_image: int = 512,
+        box_positive_fraction: float = 0.25,
+        bbox_reg_weights: Optional[Tuple[float]] = None,
+        # Mask parameters
+        mask_roi_pool: Optional[MultiScaleRoIAlign] = None,
+        mask_head: Optional[nn.Module] = None,
+        mask_predictor: Optional[nn.Module] = None
```
frgfm, line 190: `Optional[MaskRCNNPredictor]` I guess

torchvision/models/detection/mask_rcnn.py
```diff
@@ -149,26 +150,45 @@ class MaskRCNN(FasterRCNN):
...
+        box_nms_thresh: float = 0.5,
+        box_detections_per_img: int = 100,
+        box_fg_iou_thresh: float = 0.5,
+        box_bg_iou_thresh: float = 0.5,
+        box_batch_size_per_image: int = 512,
+        box_positive_fraction: float = 0.25,
+        bbox_reg_weights: Optional[Tuple[float]] = None,
+        # Mask parameters
+        mask_roi_pool: Optional[MultiScaleRoIAlign] = None,
+        mask_head: Optional[nn.Module] = None,
```
frgfm, line 189: `Optional[MaskRCNNHeads]` over here

torchvision/models/detection/retinanet.py
```diff
@@ -314,18 +352,26 @@ class RetinaNet(nn.Module):
...
+        backbone: nn.Module,
+        num_classes: int,
+        # transform parameters
+        min_size: int = 800,
+        max_size: int = 1333,
+        image_mean: Optional[Tuple[float]] = None,
+        image_std: Optional[Tuple[float]] = None,
+        # Anchor parameters
+        anchor_generator: Optional[AnchorGenerator] = None,
+        head: Optional[nn.Module] = None,
```
frgfm, line 366: I'm not sure here but maybe it's `Optional[RetinaNetHead]`

torchvision/models/detection/retinanet.py
frgfm: small typo I think: `Optional[int]`

frgfm (changes requested):
Thanks for the help!
You forgot typing annotations on the `forward` methods of most classes of `faster_rcnn.py` module, I added comments about things that weren't already pointed out by @datumbox!

oke-aditya:
I have combined the PRs. I will address all the above comments :smiley:

torchvision/models/detection/retinanet.py
datumbox: damn! thanks for flagging.

torchvision/models/detection/retinanet.py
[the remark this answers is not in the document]
oke-aditya: Ok I have kept it as `Optional[object]`. Let's think about changing in a new PR.

oke-aditya (commented):
Let me know the thoughts

torchvision/models/detection/retinanet.py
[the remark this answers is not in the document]
pmeier: Since everything is an `object` in Python, this is equivalent to `Any`. Why not use `Optional[Matcher]`?

pmeier (commented):
Overall looks good, thanks @oke-aditya. I have a few comments inline. 

Apart from them, there is the larger issue of annotating inputs with `nn.Module` rather than a more precise type. In general, we should make the input types as lenient and the output types as strict as possible. Meaning if an `nn.Module` has all teh interfaces we need within our implementation, we shouldn't force a specific type as input. On the other hand, if we return two possible subtypes of `nn.Module`, it would be better to return `Union[Foo, Bar]` instead of `nn.Module`. The former case carries more information for static type checkers to work with.

Note that we need to relax that general rule a bit, since we operate on `torch.jit > mypy`. For example, in some cases `Sequence` would be a much better annotation than `List` or `Tuple`, but since `torch.jit` does not (yet?) support `Sequence` we can't use it. You can run our test suite to see if there are any issues with this.

torchvision/models/detection/faster_rcnn.py
oke-aditya: I'm unsure of what you mean @pmeier . I'm new to type hints.
I think trying to be very strict in typing by custom classes might be too much. If `nn.Module` satisfies the purpose I think it should be good enough?

torchvision/models/detection/keypoint_rcnn.py
```diff
@@ -256,7 +285,10 @@ def __init__(self, in_channels, num_keypoints):
         self.up_scale = 2
         self.out_channels = num_keypoints
 
-    def forward(self, x):
+    def forward(
+        self,
+        x,
+    ):
```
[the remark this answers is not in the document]
oke-aditya: I'm unsure of types here. My guess is `x: Tensor` and return is also a `Tensor`

oke-aditya (commented):
I think we should discuss a bit more :smile: 
Sorry I'm a bit new to typing.

oke-aditya:
Also can someone let me know where and why CI is failing. I tried to debug, but couldn't catch it :disappointed: 
I think it's due to `roi_heads.py`. Since [this](https://github.com/pytorch/vision/pull/4220/commits/1ba33ce96427084e0df1be37588141781250cd86) commit passed the CI.

frgfm:
> Also can someone let me know where and why CI is failing. I tried to debug, but couldn't catch it 
> I think it's due to `roi_heads.py`. Since [this](https://github.com/pytorch/vision/pull/4220/commits/1ba33ce96427084e0df1be37588141781250cd86) commit passed the CI.

@oke-aditya what do you mean? :thinking: 
For mypy, you can (and should) run it locally to make sure the PR works
For the unittests, if you don't want to run them locally, you can check the logs of the CI to see if it's related to your modifications :+1:

oke-aditya:
@frgfm the current CI failure is due to my modifications. 😅 Hence need little help in debugging.

frgfm:
> @frgfm the current CI failure is due to my modifications.  Hence need little help in debugging.

Oh I see! I suggest breaking the PR into several ones. There are too many modifications (hundreds of LoC) to effectively review or debug in my opinion :sweat_smile:

oke-aditya:
I'm undoing `roi_heads.py`. In this PR let's tackle teh detection models only. Mypy seems to trigger lot of errors!

torchvision/models/detection/transform.py
```diff
@@ -117,16 +126,19 @@ def forward(self,
                 targets[i] = target_index
 
         image_sizes = [img.shape[-2:] for img in images]
-        images = self.batch_images(images, size_divisible=self.size_divisible)
+        batched_images = self.batch_images(images, size_divisible=self.size_divisible)
```
oke-aditya, line 129: instead of 
```
images = self.batch_images(images, size_divisible=self.size_divisible)
```
I did
```
batched_images = self.batch_images(images, size_divisible=self.size_divisible)
```

This avoided the mypy error as it now is able to guess the type.

oke-aditya:
Here is a brief summary of where mypy does not agree with the code.

1. Inplace modifications of `List[Tensor]` which confuses mypy.

Mypy always thinks we are operating on list of tensors. e.g when we re assign self.var = cat(sel.var, 0)
It think var still is `List[Tensor]`. A possible fix is to rename the variables after assignment.
I tried it once and it has worked. I have commented it in `transforms.py` file

These are mostly the errors associate with `List[Tensor]` and operation not being compatible.

2. Can't use Union for JIT.

E.g. When we return in `eager_outputs` methods of retinanet and ssd. This method is `@jit,unused`.
We use `Tuple[Dict[str, Tensor], List[Dict[str, Tensor]]]` 
 
But we actually return `Union[Tuple[Dict[str, Tensor], List[Dict[str, Tensor]]]]]`.
This probably is done because JIT does not support Union. I think so there is no workaround.

3. backbone out channels: -

Well something that seems so natural isn't so for mypy (I think life too is same)
Naturally We expect  this to return int,
```
out_channels = backbone.out_channels
```
But actually it doesn't! Mypy thinks this is `Union[Tensor, Module]`

Also backbone is probably indexable (We can do backbone[-1]) But I think mypy doesn't think so.

Let me know what you all think

docs/Makefile
```diff
@@ -6,7 +6,7 @@ ifneq ($(EXAMPLES_PATTERN),)
 endif
 
 # You can set these variables from the command line.
-SPHINXOPTS    = -W -j auto $(EXAMPLES_PATTERN_OPTS)
+SPHINXOPTS    = -j auto $(EXAMPLES_PATTERN_OPTS)
```
NicolasHug, line 9: Hi @oke-aditya , we should not remove the `-W` flag. What warnings were you getting?

oke-aditya: Oh no I commited wrong files.

oke-aditya:
Hi all! How do we go about this?

One option is to add typing only in this PR and keep the CI green. Let mypy ignore all the types.
In a follow-up PR fixup where mypy does not agree.

Le me know what you all think?

torchvision/models/detection/ssd.py
```diff
@@ -197,9 +198,9 @@ def __init__(self, backbone: nn.Module, anchor_generator: DefaultBoxGenerator,
         self.proposal_matcher = det_utils.SSDMatcher(iou_thresh)
 
         if image_mean is None:
-            image_mean = [0.485, 0.456, 0.406]
+            image_mean = (0.485, 0.456, 0.406)
```
pmeier, line 201: Why did you change the type? Shouldn't `List[float]` also be ok?

torchvision/models/detection/transform.py
```diff
@@ -18,7 +18,7 @@ def _get_shape_onnx(image: Tensor) -> Tensor:
 @torch.jit.unused
 def _fake_cast_onnx(v: Tensor) -> float:
     # ONNX requires a tensor but here we fake its type for JIT.
-    return v
+    return v  # type: ignore
```
pmeier, line 21: What is the error? Can we use
```suggestion
    return cast(float, v)
```

torchvision/models/detection/faster_rcnn.py
pmeier: > If nn.Module satisfies the purpose I think it should be good enough?

Yes. As a rule of thumb, the input types should be as loose as possible and the output types as strict as possible. 

After some digging, I think 

https://github.com/pytorch/vision/blob/fe3118c51b487faf1aa59861ffd08b5658f41254/torchvision/models/detection/roi_heads.py#L753

is the only time we do anything with the object passed as `box_head`. Since it is possible to call an `nn.Module`, using it as annotation is fine.

---

In other cases @frgfm's comment might have some merit in other cases: for example, if we at some point access `box_head.foo`, `nn.Module` is not sufficient as annotation, since it doesn't define a `foo` attribute. In such a case we have two options:
1. Annotate with a custom class such as `RPNHead` that has this attribute.
2. Create a  `typing.Protocol` that defines the interface we are looking for. In the example above the object passed to `box_head` needs to be callable and define an attribute `foo`.

Especially if users are allowed to pass custom objects, I would always prefer 2. over 1. because it doesn't require any changes on the user side. I don't know if this works with torchscript though.

Let's discuss this if the need for something like this arises.

pmeier (commented):
> Hi all! How do we go about this?

Sorry, this PR slipped under my radar.

> One option is to add typing only in this PR and keep the CI green. Let mypy ignore all the types.
In a follow-up PR fixup where mypy does not agree.

I dislike ignoring errors. Is there anything that is blocking right now, or does this just need more work?

torchvision/models/detection/ssd.py
datumbox: Hmmm. I believe it's because he defines the type as `Tuple[float, ...]` instead of `List[float]`. I think over the library we use a lot of different ways for this, perhaps it would be best to keep one and stick with it. I think `Tuple[float, ...]` is only used in a handful of methods at `functional_pil.py` while `List[float]` is more common.

torchvision/models/detection/transform.py
datumbox: No, this is needs to be as-is. :( 

It's a nasty workaround to deal with the fact that ONNX requires the value to be a Tensor (so that it's part of its computational graph). If you cast it, it will no longer work...

pmeier: Are you sure? `from typing import cast` is a no-op at runtime. It only exists to help `mypy`.

datumbox: Sounds good then, give it a try and if the test on ONNX does not fail then we can definitely keep it. Perhaps even edit the comment and explain that the cast is no-op at runtime and it's there only to help `mypy`.

torchvision/models/detection/mask_rcnn.py
```diff
@@ -185,7 +206,7 @@ def __init__(self, backbone, num_classes=None,
                 sampling_ratio=2)
 
         if mask_head is None:
-            mask_layers = (256, 256, 256, 256)
+            mask_layers = (56, 256, 256, 256)
```
oke-aditya, line 209: My god a bad mistake. Hope I don't foget to fix this.

torchvision/models/detection/ssd.py
oke-aditya: Hmm, `List[float]` too would have worked fine. I will try changing it once, I don't recall the error. 

Anyway about tuple.
Since we are not modifying `image_mean` or `image_std`, it can be good to leave as tuple. As such tuple are immutable so a good standard choice. Tuple can save a bit of memory (too little but yeah).

oke-aditya:
Also note this comment contains my analysis of the current CI failure.
Feel free to let me know from these

https://github.com/pytorch/vision/pull/4220#issuecomment-891279910

torchvision/models/detection/ssd.py
pmeier: The best choice would be `Sequence[float]` since we do not care about the actual type. But that is not supported by torchscript. Thus, I would simply go with whatever we had or in case you add something new go with what we use the most throughout the code base.

pmeier:
> 1. Inplace modifications of `List[Tensor]` which confuses mypy.
> Mypy always thinks we are operating on list of tensors. e.g when we re assign self.var = cat(sel.var, 0)
> It think var still is `List[Tensor]`. A possible fix is to rename the variables after assignment.
> I tried it once and it has worked. I have commented it in `transforms.py` file above
> 
> These are mostly the errors associate with `List[Tensor]` and operation not being compatible.

Yes, the solution is to rename the variables, since they don't have the same type. In case `mypy` cannot infer the correct type, we also need to better annotate the callable that transforms the list.

> 2. Can't use Union for JIT.
> 
> 
> E.g. When we return in `eager_outputs` methods of retinanet and ssd. This method is `@jit,unused`.
> We use `Tuple[Dict[str, Tensor], List[Dict[str, Tensor]]]`
> 
> But we actually return `Union[Tuple[Dict[str, Tensor], List[Dict[str, Tensor]]]]]`.
> This probably is done because JIT does not support Union. I think so there is no workaround.

Is this still true in the light of #4360? 

> 3. backbone and out channels errors: -
> 
> 
> Well something that seems so natural isn't so for mypy (I think life too is same)
> Naturally We expect this to return int,
> 
> ```
> out_channels = backbone.out_channels
> ```
> 
> But actually it doesn't! Mypy thinks this is `Union[Tensor, Module]`

The problem is that `backbone` is annotated as `nn.Module` and this does not have an `out_channels` attribute. `nn.Module` is set up in way that `mypy` will infer `Union[Tensor, Module]` in case you access something unknown. Two solutions:

1. Annotate `backbone` more precisely either through the actual class that is used. Alternatively, use a protocol as described in https://github.com/pytorch/vision/pull/4220#discussion_r707928797.
2. Wrap the attribute access in a `cast`, i.e. `out_channels = cast(int, backbone.out_channels)` to tell `mypy` the correct type.

oke-aditya:
> Is this still true in the light of #4360?

Well no. I think Union issue isn't anymore.

torchvision/models/detection/transform.py
oke-aditya: Seems that cast() is working fine for ONNX,

torchvision/models/detection/ssd.py
oke-aditya: Let's make a convention of using `List[float]`. I'm editing the code to keep consistency with List[float]

torchvision/models/detection/transform.py
```diff
@@ -222,8 +237,8 @@ def batch_images(self, images: List[Tensor], size_divisible: int = 32) -> Tensor
 
     def postprocess(self,
                     result: List[Dict[str, Tensor]],
-                    image_shapes: List[Tuple[int, int]],
```
oke-aditya, line 225: Here is the problem, We have colliding types.
We are Defining this as `Tuple[int, int]` here but down in `resize_boxes` we use List[int]
I'm not sure whether we should keep `List[int]` or `Tuple[int, int]` we would need to modify the code accordingly.

Currently due to this, we have CI failure.

 @pmeier and @datumbox

oke-aditya: Thought the convention seems to be using `List[List[int]]` everywhere

torchvision/models/detection/transform.py
```diff
@@ -160,12 +175,12 @@ def resize(self,
             return image, target
 
         bbox = target["boxes"]
-        bbox = resize_boxes(bbox, (h, w), image.shape[-2:])
+        bbox = resize_boxes(bbox, [h, w], list(image.shape[-2:]))
```
oke-aditya, line 178: This is necessary due to type being defined below as List[int]

oke-aditya (commented):
Hi! I think we need some assumption in typing
Due to `resize_keypoints` and `resize_boxes` operations defined [here](https://github.com/pytorch/vision/pull/4220/files#diff-915d3fa993285c902d3bb51970cd119a3a24cfcf6a587aaa5f227bd1213a3beeL254)

We can't annotate using either `Tuple[int, int]` or `List[int]`.
I think we need to go ahead with either and refactor accordingly.

torchvision/models/detection/generalized_rcnn.py
```diff
@@ -31,16 +37,23 @@ def __init__(self, backbone, rpn, roi_heads, transform):
         # used only on torchscript mode
         self._has_warned = False
 
-    @torch.jit.unused
-    def eager_outputs(self, losses, detections):
-        # type: (Dict[str, Tensor], List[Dict[str, Tensor]]) -> Union[Dict[str, Tensor], List[Dict[str, Tensor]]]
+    def eager_outputs(
```
oke-aditya, line 40: Now JIT supports union, so we can remove the decorator and add Union

torchvision/models/detection/keypoint_rcnn.py
```diff
@@ -55,10 +57,10 @@ class KeypointRCNN(FasterRCNN):
             If box_predictor is specified, num_classes should be None.
         min_size (int): minimum size of the image to be rescaled before feeding it to the backbone
         max_size (int): maximum size of the image to be rescaled before feeding it to the backbone
-        image_mean (Tuple[float, float, float]): mean values used for input normalization.
+        image_mean (List[float, float, float]): mean values used for input normalization.
```
oke-aditya, line 60: Well we have decided it to be `List[int]` let me know if this is fine.

torchvision/models/detection/mask_rcnn.py
```diff
@@ -149,34 +151,53 @@ class MaskRCNN(FasterRCNN):
...
+    ) -> None:
 
         assert isinstance(mask_roi_pool, (MultiScaleRoIAlign, type(None)))
 
         if num_classes is not None:
             if mask_predictor is not None:
                 raise ValueError("num_classes should be None when mask_predictor is specified")
 
-        out_channels = backbone.out_channels
+        out_channels = cast(int, backbone.out_channels)
```
oke-aditya, line 200: We need to cast a @pmeier suggested, this works fine with mypy

torchvision/models/detection/retinanet.py
```diff
@@ -259,15 +295,17 @@ class RetinaNet(nn.Module):
         num_classes (int): number of output classes of the model (including the background).
         min_size (int): minimum size of the image to be rescaled before feeding it to the backbone
         max_size (int): maximum size of the image to be rescaled before feeding it to the backbone
-        image_mean (Tuple[float, float, float]): mean values used for input normalization.
+        image_mean (List[float, float, float]): mean values used for input normalization.
```
oke-aditya, line 298: Again same we have decided for List.

torchvision/models/detection/retinanet.py
```diff
@@ -373,16 +420,23 @@ def __init__(self, backbone, num_classes,
...
         self._has_warned = False
 
-    @torch.jit.unused
-    def eager_outputs(self, losses, detections):
-        # type: (Dict[str, Tensor], List[Dict[str, Tensor]]) -> Tuple[Dict[str, Tensor], List[Dict[str, Tensor]]]
+    def eager_outputs(
+        self,
+        losses: Dict[str, Tensor],
+        detections: List[Dict[str, Tensor]],
+    ) -> Union[Dict[str, Tensor], List[Dict[str, Tensor]]]:
```
oke-aditya, line 427: Using Union for eager output.

torchvision/models/detection/roi_heads.py
```diff
@@ -461,11 +461,10 @@ def _onnx_paste_masks_in_image_loop(masks, boxes, im_h, im_w):
     return res_append
 
 
-def paste_masks_in_image(masks, boxes, img_shape, padding=1):
-    # type: (Tensor, Tensor, Tuple[int, int], int) -> Tensor
+def paste_masks_in_image(masks: Tensor, boxes: Tensor, img_shape: List[int], padding: int = 1) -> Tensor:
```
oke-aditya, line 464: I'm only typing this function for now, will follow up this file in new PR.
This needed as this raises error otherwise.
Again here the `img_shape` is `Tuple[int, int]` which cannot  be as `resize_boxes` and `resize_keypoints` `use List[int]`
We need consistency here.

torchvision/models/detection/ssd.py
```diff
@@ -164,27 +164,39 @@ class SSD(nn.Module):
...
         self.backbone = backbone
-
         self.anchor_generator = anchor_generator
-
         self.box_coder = det_utils.BoxCoder(weights=(10., 10., 5., 5.))
 
         if head is None:
             if hasattr(backbone, 'out_channels'):
-                out_channels = backbone.out_channels
+                # Other models use int as out_channel, we need compatibility.
```
oke-aditya, line 191: Some patchy workaround for mypy.
Since in other models we use `out_channels` from backbone as int, and in this case we need `List[int]`.
We can handle the single int case by simply converting to list. Bit consistency for the user.

torchvision/models/detection/ssd.py
```diff
@@ -164,27 +164,39 @@ class SSD(nn.Module):
...
+                # Also out_channels needs a runtime cast to int for mypy.
+                out_channels_calc = cast(int, backbone.out_channels)
+                if isinstance(out_channels_calc, int):
+                    out_channels = [out_channels_calc]
+                elif isinstance(out_channels_calc, list):
+                    out_channels = out_channels_calc
+                else:
+                    raise ValueError("Backbone should have out_channels either int or List[int] ")
             else:
                 out_channels = det_utils.retrieve_out_channels(backbone, size)
```
oke-aditya, line 202: The below assert is bit ugly.
We could instead 
`raise ValueError("Length Specified output_channels for backbone = x, while the length of aspect ratios of Achor generator are y. Which do not match.")`

torchvision/models/detection/ssd.py
```diff
@@ -212,9 +224,12 @@ def __init__(self, backbone: nn.Module, anchor_generator: DefaultBoxGenerator,
         # used only on torchscript mode
         self._has_warned = False
 
-    @torch.jit.unused
-    def eager_outputs(self, losses: Dict[str, Tensor],
-                      detections: List[Dict[str, Tensor]]) -> Tuple[Dict[str, Tensor], List[Dict[str, Tensor]]]:
+    def eager_outputs(
```
oke-aditya, line 227: Using union as supported by JIT

torchvision/models/detection/transform.py
```diff
@@ -18,13 +18,18 @@ def _get_shape_onnx(image: Tensor) -> Tensor:
 @torch.jit.unused
 def _fake_cast_onnx(v: Tensor) -> float:
     # ONNX requires a tensor but here we fake its type for JIT.
-    return v
+    # cast is no-op at runtime and it's there only to help mypy.
+    return cast(float, v)
```
oke-aditya, line 22: This worked.

oke-aditya (commented):
A small changelog, we aren't done yet but would say halfway there.
Mypy now complains slightly less.

torchvision/models/detection/generalized_rcnn.py
```diff
@@ -31,16 +37,23 @@ def __init__(self, backbone, rpn, roi_heads, transform):
...
 
         return detections
 
-    def forward(self, images, targets=None):
-        # type: (List[Tensor], Optional[List[Dict[str, Tensor]]]) -> Tuple[Dict[str, Tensor], List[Dict[str, Tensor]]]
+    def forward(
+        self,
+        images: List[Tensor],
+        targets: Optional[List[Dict[str, Tensor]]] = None,
+    ) -> Tuple[Dict[str, Tensor], List[Dict[str, Tensor]]]:
```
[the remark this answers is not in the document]
pmeier: You don't need to nest `Union`'s: `Union[Tuple[losses,detection], Union[losses,detection]]` is equal to `Union[losses, detection, Tuple[losses, detection]]`. I say, let's retype it.

torchvision/models/detection/ssd.py
pmeier: 1. If `out_channels` can be `int` or `List[int]`, shouldn't we also cast it as `Union[int, List[int]]`?
2. From the old code I get that `out_channels` is expected to be a `List[int]` so adding the possibility of it being `int` is a runtime change. Let's just annotate in this PR and only change stuff in follow-up PRs.

torchvision/models/detection/ssd.py
pmeier: Agreed. Plain `assert`'s should be avoided. Feel free to send a follow-up PR.

mypy.ini
```diff
@@ -32,6 +32,10 @@ ignore_errors = True
 
 ignore_errors = True
 
+[mypy-torchvision.models.detection.ssdlite]
+
+ignore_errors = True
+
```
pmeier: Will you do this in a follow-up PR?

torchvision/models/detection/keypoint_rcnn.py
pmeier: You mean `List[float]`, right? Plus, `List` only has a single inner type.

```suggestion
        image_mean (List[float]): mean values used for input normalization.
```

torchvision/models/detection/keypoint_rcnn.py
```diff
@@ -55,10 +57,10 @@ class KeypointRCNN(FasterRCNN):
...
             They are generally the mean values of the dataset on which the backbone has been trained
             on
-        image_std (Tuple[float, float, float]): std values used for input normalization.
+        image_std (List[float, float, float]): std values used for input normalization.
```
pmeier, line 63:
```suggestion
        image_std (List[float]): std values used for input normalization.
```

torchvision/models/detection/roi_heads.py
pmeier: Since this file is currently ignored by `mypy`, let's not introduce annotations here.

torchvision/models/detection/rpn.py
```diff
@@ -60,16 +65,26 @@ def forward(self, x):
         return logits, bbox_reg
 
 
-def permute_and_flatten(layer, N, A, C, H, W):
-    # type: (Tensor, int, int, int, int, int) -> Tensor
+def permute_and_flatten(
+    layer: Tensor,
+    N: int,
+    A: int,
```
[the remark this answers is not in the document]
pmeier: Good point. What is passed in as `A` in the places where this is called?

torchvision/models/detection/transform.py
pmeier: > Thought the convention seems to be using `List[List[int]]` everywhere

If that is the convention, let's also roll with it here. Although I would prefer `Tuple[int, int]`, because an image size is always of length 2 and not arbitrary.

pmeier (commented):
Hey @oke-aditya, I did another pass. I have some comments inline. This gets harder and harder to review due to the sheer amount of changes. I suggest we split this into multiple PRs for each file for the "final" review, if we are mostly happy here.

torchvision/models/detection/keypoint_rcnn.py
oke-aditya: Well, here is the catch. By annotating as `List[float]`.  The end-user will not know how many numbers to pass in.
Hence I propose to type with `Tuple[float, float, float]` and refactor code accordingly.

Also we can make using Tuple a Convention, where we know the number of elements.
E.g. case is this, where image_mean, image_std will be of length 3; the below case where image_size will be length 2.

torchvision/models/detection/keypoint_rcnn.py
oke-aditya: See my comment below for using Tuple.

torchvision/models/detection/roi_heads.py
oke-aditya: Well I have to annotate it due to the `img_shape` which should be Tuple[int, int], Otherwise mypy will compain in other places.

torchvision/models/detection/rpn.py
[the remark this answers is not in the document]
oke-aditya: Here is block of code that uses this.
```
for box_cls_per_level, box_regression_per_level in zip(
        box_cls, box_regression
    ):
        N, AxC, H, W = box_cls_per_level.shape
        Ax4 = box_regression_per_level.shape[1]
        A = Ax4 // 4
        C = AxC // A
        box_cls_per_level = permute_and_flatten(
            box_cls_per_level, N, A, C, H, W
        )
        box_cls_flattened.append(box_cls_per_level)

        box_regression_per_level = permute_and_flatten(
            box_regression_per_level, N, A, 4, H, W
        )
        box_regression_flattened.append(box_regression_per_level)
```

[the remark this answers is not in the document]
oke-aditya: We would need to keep A or refactor accordingly. Though not in scope of this PR.

torchvision/models/detection/ssd.py
oke-aditya: I will open an issue later to track these followups, there will be many I think

torchvision/models/detection/transform.py
oke-aditya: Ok so I will refactor to `Tuple[int, int]`. This might be bc breaking as it changes the old annotation of `List[int]` and users would get error if they were not passing List of length 2.

oke-aditya (commented):
HI! Feel free to add your thoughts. I get that this PR modfiies lot of files which makes reviewing a harder as we progress.
Once we finalize, I will send PR for each file individually!

torchvision/models/detection/transform.py
pmeier: Why would they get an error? IF we only change the annotations, the runtime behavior should not change.

torchvision/models/detection/keypoint_rcnn.py
oke-aditya: Hi!, once we get a consensus on using Tuple for finite element cases, I will refactor the code accordingly. :smile:

pmeier: > By annotating as `List[float]`. The end-user will not know how many numbers to pass in.

True, and this is why `List` is not a good type for this. Still, `List` only has a single inner type, so `List[float, float, float]` is simply a wrong annotation. If you use that inside the code instead of the docstring you will get an error.

> Also we can make using Tuple a Convention, where we know the number of elements.
E.g. case is this, where image_mean, image_std will be of length 3; the below case where image_size will be length 2.

I agree in general. Is this here such a case though? Are we somehow enforcing three channels or does the number of elements in `image_mean` and `image_std` just has to match the number of channels?

pmeier:
> I get that this PR modfiies lot of files which makes reviewing a harder as we progress. Once we finalize, I will send PR for each file individually!

I would prefer to split sooner than later.

pmeier (commented):
> I get that this PR modfiies lot of files which makes reviewing a harder as we progress. Once we finalize, I will send PR for each file individually!

I would prefer to split sooner than later.

oke-aditya:
I can split it, bit since the files are highly coupled, it makes hard to annotate in multiple PRs.

pmeier:
> I can split it, bit since the files are highly coupled, it makes hard to annotate in multiple PRs.

Start with the one that has no dependencies and work up the chain:

1. `image_list.py`
2. `_utils.py`
3. `anchor_utils.py`
4. `roi_heads.py`
5. `transform.py`
6. `rpn.py`
...

torchvision/models/detection/keypoint_rcnn.py
oke-aditya: I think we assume that channels are 3. We don't directly support any other channels without editing the backbone. So it might be safe assumption. I think there maybe a case where user has a backbone with different no of channels (say 2 or 4) and is trying to build FRCNN. Though I'm not very sure, as I haven't constructed a backbone with different channels.

Maybe Vasilas can chip in his thoughts, he knows it better cc @datumbox

datumbox: My 2 cents: It's true that out of the box we don't offer a method that builds a backbone with channels !=3. Nevertheless the API of the FRCNN does support this as one can pass an arbitrary backbone which could support arbitrary number of channels. I would really prefer it if we could use `Tuple[float, ...]` because that's the true type (immutable tuple of arbitrary many floats) but based on your latest findings that's not supported by JIT. So the next best thing that we got is `List[float]` as Philip suggested.

oke-aditya (commented):
A small review, I will split this PR soon. I have substantially reduced mypy errors.

I think we can start with _utils and utility files, these would help making this PR easier.

oke-aditya:
One of the problems in typing utils and leading errors to mypy are because of #4389, as per discussion there, we can't use nn.Module to type the detection utils, so it some errors for mypy remain.

oke-aditya:
Using μfmt and setting mypy setting to allow redefinition has made this PR lot easier !!!!

Current error count has decreased to.

```
Found 32 errors in 8 files (checked 128 source files)
```

If I recall well the error count was close to 100 in the beginning. We are getting there.

I'm sorry that It's bit ugly work from my side by not splitting the PR and tackling easier files first. I have been very slow on working this PR :( .

pmeier:
> I'm sorry that It's bit ugly work from my side by not splitting the PR and tackling easier files first. I have been very slow on working this PR :( .

No worries, this PR is not blocking anything. We are grateful that you are picking up the work! Just note, that I'm not going to review this PR any further before the split. It just takes to much time to read all the discussions over and over again to get up to speed.

oke-aditya:
Hi all. I'm closing this PR and today I will split this up into multiple ones. Each file will unblock only the particular file which mypy needs to address. That is mypy will ignore all the other files that aren't going to be modified by each independent PR.

PR #4223, opened 2021-07-29
Added typing annotations to ``models/detection`` ``generalized_rcnn``, ``keypoint_rcnn``, ``rpn``

oke-aditya:
Followup of #2025

I will tag typings where I have a few doubts.

cc @pmeier


datumbox:
Generally it's good to split PRs to avoid making them too long and complex. In this very specific case, I suspect we might have duplicate comments from #4220 applying in this PR. 

If you decide to keep them separate please have a look on the 1st PR and pass any relevant comments here as well. :)

oke-aditya:
No problem. I will keep everything in the first PR. :smile: I will close this.

PR #4224, opened 2021-07-29
Added typing annotations to io/__init__

frgfm:
Following up on #2025, this PR adds missing typing annotations in `io/__init__.py`.

Note to reviewers: I'm not sure about the return types when a method returns `self` :man_shrugging: 

Any feedback is welcome!


torchvision/io/__init__.py
```diff
@@ -126,10 +127,10 @@ def __next__(self):
             raise StopIteration
         return {"data": frame, "pts": pts}
 
-    def __iter__(self):
+    def __iter__(self) -> Iterator['VideoReader']:
         return self
 
-    def seek(self, time_s: float):
+    def seek(self, time_s: float) -> 'VideoReader':
```
datumbox, line 133: Could be replaced with `VideoReader` once we drop Python 3.6 support. See [here](https://stackoverflow.com/a/33533514).

torchvision/io/__init__.py
```diff
@@ -126,10 +127,10 @@ def __next__(self):
             raise StopIteration
         return {"data": frame, "pts": pts}
 
-    def __iter__(self):
+    def __iter__(self) -> Iterator['VideoReader']:
```
datumbox, line 130: See [mypy doc](https://mypy.readthedocs.io/en/stable/protocols.html#iterable-t).

datumbox (commented):
@frgfm I changed the return type for self. See my commit.

torchvision/io/__init__.py
oke-aditya: Nice. 
I learn something new everyday because of review from torchvision maintainers 😇

There are couple of more places where I have used this workaround in typing in torchvision. I will keep a note of this point.

frgfm: @datumbox yup fully agree! But for now (this PR), I guess we keep the string version?

datumbox: yes that's right, nothing we can do for now.

datumbox:
@frgfm could you please merge master to clear the CI errors?

frgfm:
@datumbox all errors were cleared after the previous merge :ok_hand: 
I just merged main so have everything ready for review & I guess merge :+1:

datumbox (approved):
LGTM, thanks!

PR #4228, opened 2021-07-29
Added missing arg description in lraspp_mobilenet_v3_large

frgfm:
This PR adds a missing argument description for `lraspp_mobilenet_v3_large`.

Any feedback is welcome!

### DIFF
--- a/torchvision/models/segmentation/segmentation.py
+++ b/torchvision/models/segmentation/segmentation.py
@@ -181,6 +181,7 @@ def lraspp_mobilenet_v3_large(pretrained=False, progress=True, num_classes=21, *
             contains the same classes as Pascal VOC
         progress (bool): If True, displays a progress bar of the download to stderr
         num_classes (int): number of output classes of the model (including the background)
+        aux_loss (bool): If True, it uses an auxiliary loss
     """
     if kwargs.pop("aux_loss", False):
         raise NotImplementedError('This model does not use auxiliary loss')


torchvision/models/segmentation/segmentation.py
```diff
@@ -181,6 +181,7 @@ def lraspp_mobilenet_v3_large(pretrained=False, progress=True, num_classes=21, *
             contains the same classes as Pascal VOC
         progress (bool): If True, displays a progress bar of the download to stderr
         num_classes (int): number of output classes of the model (including the background)
+        aux_loss (bool): If True, it uses an auxiliary loss
```
NicolasHug, line 184: thanks for the PR @frgfm 

If you look just below we have

```py
    if kwargs.pop("aux_loss", False):
        raise NotImplementedError('This model does not use auxiliary loss')
```

So documenting `aux_loss` is actually more confusing than anything. We should instead indicate how the `kwargs` argument is used

frgfm: Oh my bad, I misread, I thought the arg was declared outside of kwargs!

NicolasHug: We'll  be happy to consider a Pr for kwargs though, if you feel like it :) !

frgfm: Mmmh, what do you have in mind?

Something like saying whose kwargs are those and mentioning the special case of aux_loss?

PR #4230, opened 2021-07-29
Import squeeze from symbolic_opset11

jiafatom:
`_register_custom_op` set `_onnx_opset_version=11`, so we need import `squeeze`, etc from `symbolic_opset11`.

### DIFF
--- a/torchvision/ops/_register_onnx_ops.py
+++ b/torchvision/ops/_register_onnx_ops.py
@@ -8,7 +8,8 @@ _onnx_opset_version = 11
 def _register_custom_op():
     from torch.onnx.symbolic_helper import parse_args, scalar_type_to_onnx, scalar_type_to_pytorch_type, \
         cast_pytorch_to_onnx
-    from torch.onnx.symbolic_opset9 import select, unsqueeze, squeeze, _cast_Long, reshape
+    from torch.onnx.symbolic_opset9 import _cast_Long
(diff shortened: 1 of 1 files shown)


torchvision/ops/_register_onnx_ops.py
```diff
@@ -8,7 +8,8 @@
 def _register_custom_op():
     from torch.onnx.symbolic_helper import parse_args, scalar_type_to_onnx, scalar_type_to_pytorch_type, \
         cast_pytorch_to_onnx
-    from torch.onnx.symbolic_opset9 import select, unsqueeze, squeeze, _cast_Long, reshape
+    from torch.onnx.symbolic_opset9 import _cast_Long
```
NicolasHug, line 11: Thanks for the PR @jiafatom , I'm wondering, why leaving this one to opset9?

jiafatom: There is no `_cast_Long` in `torch.onnx.symbolic_opset11`. Not every function can be found in upper opsets `torch.onnx.symbolic_opset11`. We only updated if there is spec change.

jiafatom: @NicolasHug Can we merge the PR? thanks!

NicolasHug (approved):
Thanks @jiafatom !

PR #4233, opened 2021-07-30
Pytest for test_videoapi.py and test_video_reader.py

vmoens:
I also removed the common_utils.py/TestCase class which -- as far as i can tell - wasn't used anywhere. This is probably the most contentious point of this PR and needs proper reviewing!

### DIFF
--- a/test/test_video_reader.py
+++ b/test/test_video_reader.py
@@ -2,7 +2,7 @@ import collections
 import itertools
 import math
 import os
-import unittest
+import pytest
 from fractions import Fraction
 
 import numpy as np
@@ -12,7 +12,7 @@ from numpy.random import randint
 from torchvision import set_video_backend
 from torchvision.io import _HAS_VIDEO_OPT
 from common_utils import PY39_SKIP, assert_equal
-import pytest
+
 
 try:
     import av
@@ -564,8 +564,6 @@ class TestVideoReader:
                 audio_timebase_num,
                 audio_timebase_den,
             )
-            #     min_dimension, min(tv_result[0].size(1), tv_result[0].size(2))
-            # )
             assert min_dimension==min(tv_result[0].size(1), tv_result[0].size(2))
 
     def test_read_video_from_file_rescale_max_dimension(self):
@@ -606,8 +604,6 @@ class TestVideoReader:
                 audio_timebase_num,
                 audio_timebase_den,
             )
-            #     max_dimension, max(tv_result[0].size(1), tv_result[0].size(2))
-            # )
             assert max_dimension==max(tv_result[0].size(1), tv_result[0].size(2))
 
     def test_read_video_from_file_rescale_both_min_max_dimension(self):
@@ -648,10 +644,6 @@ class TestVideoReader:
                 audio_timebase_num,
                 audio_timebase_den,
             )
-            #     min_dimension, min(tv_result[0].size(1), tv_result[0].size(2))
-            # )
-            #     max_dimension, max(tv_result[0].size(1), tv_result[0].size(2))
-            # )
             assert min_dimension == min(tv_result[0].size(1), tv_result[0].size(2))
             assert max_dimension == max(tv_result[0].size(1), tv_result[0].size(2))
 
@@ -830,10 +822,6 @@ class TestVideoReader:
                         * float(atimebase[0])
                         / float(atimebase[1])
                     )
-                    #     aframes.size(0),
-                    #     int(duration * asample_rate.item()),
-                    #     delta=0.1 * asample_rate.item(),
-                    # )
                     assert abs(aframes.size(0)-int(duration * asample_rate.item()))<0.1 * asample_rate.item()
 
     @PY39_SKIP
--- a/test/test_video_reader.py
+++ b/test/test_video_reader.py
@@ -295,12 +295,12 @@ class TestVideoReader:
 
         # check if pts of video frames are sorted in ascending order
         for i in range(len(vframe_pts) - 1):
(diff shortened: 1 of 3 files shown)


test/test_videoapi.py
```diff
@@ -178,23 +166,23 @@ def test_accurateseek_middle(self):
...
 
-        self.assertTrue(metadata["subtitles"]["duration"] is not None)
+        assert (metadata["subtitles"]["duration"] is not None)
         os.remove(video_path)
 
 
-if __name__ == "__main__":
-    unittest.main()
+if __name__ == '__main__':
+    pytest.main([__file__])
```
NicolasHug, line 188: Looks like this is missing a blank line at the end of the file

test/test_video_reader.py
```diff
@@ -282,45 +281,46 @@ def check_separate_decoding_result(self, tv_result, config):
         video_duration = vduration.item() * Fraction(
             vtimebase[0].item(), vtimebase[1].item()
         )
-        self.assertAlmostEqual(video_duration, config.duration, delta=0.5)
+        assert abs(video_duration-config.duration)<=0.5
```
NicolasHug, line 284: have you tried using `pytest.approx` here? https://docs.pytest.org/en/6.2.x/reference.html#pytest-approx

I think we should be able to use it everywhere `assertAlmostEqual` was used

test/test_video_reader.py
```diff
@@ -1187,7 +1175,7 @@ def test_probe_video_from_memory(self):
 
     def test_probe_video_from_memory_script(self):
         scripted_fun = torch.jit.script(io._probe_video_from_memory)
-        self.assertIsNotNone(scripted_fun)
+        assert scripted_fun
```
NicolasHug, line 1178: in order to be slightly more explicit about what we don't want scripted_fun to be, let's use `assert scripted_fun is not None`

test/test_video_reader.py
```diff
@@ -1258,8 +1246,7 @@ def test_audio_present_pts(self):
                     set_video_backend(backend)
                     _, audio, _ = io.read_video(
                         full_path, start_offset, end_offset, pts_unit='pts')
-                    self.assertGreaterEqual(audio.shape[0], 1)
-                    self.assertGreaterEqual(audio.shape[1], 1)
+                    assert all(audio.shape[:2])
```
NicolasHug, line 1249: here as well it will probably be more obvious to include the `>= 1` part in the assertion

test/test_video_reader.py
```diff
@@ -1275,9 +1262,8 @@ def test_audio_present_sec(self):
                     set_video_backend(backend)
                     _, audio, _ = io.read_video(
                         full_path, start_offset, end_offset, pts_unit='sec')
-                    self.assertGreaterEqual(audio.shape[0], 1)
-                    self.assertGreaterEqual(audio.shape[1], 1)
+                    assert all(audio.shape[:2])
```
NicolasHug, line 1265: same here

test/test_video_reader.py
```diff
@@ -1275,9 +1262,8 @@ def test_audio_present_sec(self):
...
 
 
-if __name__ == "__main__":
-    unittest.main()
+if __name__ == '__main__':
+    pytest.main([__file__])
```
NicolasHug, line 1269: needs a new line here as well :)

test/test_videoapi.py
```diff
@@ -178,23 +166,23 @@ def test_accurateseek_middle(self):
                 for frame in video_reader:
                     middle_num_frames += 1
 
-                self.assertTrue(middle_num_frames < num_frames)
-                self.assertAlmostEqual(middle_num_frames, num_frames // 2, delta=1)
+                assert (middle_num_frames < num_frames)
```
NicolasHug, line 169: Here and in various places, we don't need the surrounding parenthesis. It's a nit though, feel free not to address!

NicolasHug (commented):
Thanks for the PR @vmoens ! I made a few minor comments, but this looks good!

Also there are a few linting issue, which you can find by looking at the `ci/circleci: python_lint` job below.

You did well getting rid of `TestCase`, I'm glad we're finally done with it :)

test/test_video_reader.py
```diff
@@ -1209,7 +1199,7 @@ def test_read_video_from_memory_scripted(self):
         audio_timebase_num, audio_timebase_den = 0, 1
 
         scripted_fun = torch.jit.script(io._read_video_from_memory)
-        self.assertIsNotNone(scripted_fun)
+        assert scripted_fun
```
NicolasHug, line 1202:
```suggestion
        assert scripted_fun is not None
```

test/test_videoapi.py
```diff
@@ -178,23 +167,23 @@ def test_accurateseek_middle(self):
...
-                self.assertAlmostEqual(middle_num_frames, num_frames // 2, delta=1)
+                assert middle_num_frames < num_frames
+                assert middle_num_frames == approx(num_frames // 2, abs=1)
 
                 video_reader.seek(duration / 2)
                 frame = next(video_reader)
                 lb = duration / 2 - 1 / md[stream]["fps"][0]
                 ub = duration / 2 + 1 / md[stream]["fps"][0]
-                self.assertTrue((lb <= frame["pts"]) & (ub >= frame["pts"]))
+                assert (lb <= frame["pts"]) & (ub >= frame["pts"])
```
NicolasHug, line 177: It's not originally from this PR, but should this be `and` instead of `&`?

NicolasHug (approved):
Thanks a lot @vmoens !

I made 2 minor comments but LGTM, feel free to self merge once addressed

PR #4234, opened 2021-07-30
 Added typing annotations to transforms/functional_pil

oke-aditya:
Following up on #2025, this PR adds missing typing annotations in functional_pil.py

Any feedback is welcome!

cc @frgfm @pmeier


torchvision/transforms/functional_pil.py
```diff
@@ -121,7 +126,13 @@ def adjust_gamma(img, gamma, gain=1):
 
 
 @torch.jit.unused
-def pad(img, padding, fill=0, padding_mode="constant"):
+def pad(
+    img: Image.Image,
+    padding: Union[int, List, Tuple],
```
oke-aditya, line 131: Not sure if `List[int]` would suffice.

torchvision/transforms/functional_pil.py
```diff
@@ -121,7 +126,13 @@ def adjust_gamma(img, gamma, gain=1):
...
+    fill: Union[str, int, Tuple] = 0,
```
oke-aditya, line 132: A bit unsure about this too.

torchvision/transforms/functional_pil.py
```diff
@@ -261,7 +290,13 @@ def _parse_fill(fill, img, name="fillcolor"):
 
 
 @torch.jit.unused
-def affine(img, matrix, interpolation=0, fill=None):
+def affine(
+    img: Image.Image,
+    matrix,
```
oke-aditya, line 295: I'm not sure what matrix is. I think it's `Any`

torchvision/transforms/functional_pil.py
```diff
@@ -280,7 +323,13 @@ def rotate(img, angle, interpolation=0, expand=False, center=None, fill=None):
 
 
 @torch.jit.unused
-def perspective(img, perspective_coeffs, interpolation=Image.BICUBIC, fill=None):
+def perspective(
+    img: Image.Image,
+    perspective_coeffs: float,
+    interpolation=Image.BICUBIC,
```
oke-aditya, line 329: How do we type interpolation :question:

torchvision/transforms/functional_pil.py
```diff
@@ -308,28 +357,28 @@ def to_grayscale(img, num_output_channels):
...
-def posterize(img, bits):
+def posterize(img: Image.Image, bits: int) -> Image.Image:
     if not _is_pil_image(img):
         raise TypeError('img should be PIL Image. Got {}'.format(type(img)))
     return ImageOps.posterize(img, bits)
 
 
 @torch.jit.unused
-def solarize(img, threshold):
+def solarize(img: Image.Image, threshold: int) -> Image.Image:
```
oke-aditya, line 374: Please verify. I think it's int. (after digging into PIL code)

torchvision/transforms/functional_pil.py
```diff
@@ -308,28 +357,28 @@ def to_grayscale(img, num_output_channels):
...
-def invert(img):
+def invert(img: Image.Image) -> Image.Image:
     if not _is_pil_image(img):
         raise TypeError('img should be PIL Image. Got {}'.format(type(img)))
     return ImageOps.invert(img)
 
 
 @torch.jit.unused
-def posterize(img, bits):
+def posterize(img: Image.Image, bits: int) -> Image.Image:
```
oke-aditya, line 367: Please verify. I think it's int. (after digging into PIL code)

oke-aditya (commented):
Little help would be nice! I'm not sure about few

torchvision/transforms/functional_pil.py
```diff
@@ -196,15 +207,28 @@ def pad(img, padding, fill=0, padding_mode="constant"):
...
         raise TypeError('img should be PIL Image. Got {}'.format(type(img)))
 
     return img.crop((left, top, left + width, top + height))
 
 
 @torch.jit.unused
-def resize(img, size, interpolation=Image.BILINEAR, max_size=None):
+def resize(
+    img: Image.Image,
+    size: Union[Sequence, int],
```
pmeier, line 227:
```suggestion
    size: Union[Sequence[int], int],
```

torchvision/transforms/functional_pil.py
```diff
@@ -196,15 +207,28 @@ def pad(img, padding, fill=0, padding_mode="constant"):
...
+    interpolation=Image.BILINEAR,
```
pmeier, line 228:
```suggestion
    interpolation: int = Image.BILINEAR,
```

torchvision/transforms/functional_pil.py
```diff
@@ -242,7 +266,12 @@ def resize(img, size, interpolation=Image.BILINEAR, max_size=None):
 
 
 @torch.jit.unused
-def _parse_fill(fill, img, name="fillcolor"):
+def _parse_fill(
+    fill: Union[int, float, List, Tuple],
```
pmeier, line 270:
```suggestion
    fill: Union[numbers.Number, List[numbers.Number], Tuple[numbers.Number, ...]],
```

torchvision/transforms/functional_pil.py
pmeier: As `int`. See https://github.com/python-pillow/Pillow/blob/f518149a1bbe5ebea9d0ca41612e3e6748349ace/src/PIL/Image.py#L184-L190

```suggestion
    interpolation: int = Image.BICUBIC,
```

torchvision/transforms/functional_pil.py
```diff
@@ -271,7 +306,15 @@ def affine(img, matrix, interpolation=0, fill=None):
...
 
 @torch.jit.unused
-def rotate(img, angle, interpolation=0, expand=False, center=None, fill=None):
+def rotate(
+    img: Image.Image,
+    angle: float,
+    interpolation: int = 0,
+    expand: bool = False,
+    center: Optional[Tuple[int, int]] = None,
+    fill: Optional[Union[int, float, List, Tuple]] = None,
```
pmeier, line 315: Please apply the same as I suggested above here and anywhere else.

torchvision/transforms/functional_pil.py
pmeier: Yep, [seems to be good](https://pillow.readthedocs.io/en/stable/reference/ImageOps.html#PIL.ImageOps.posterize):

> bits – The number of bits to keep for each channel (1-8).

torchvision/transforms/functional_pil.py
pmeier: [Seems to be good](https://pillow.readthedocs.io/en/stable/reference/ImageOps.html?highlight=posterize#PIL.ImageOps.solarize)

> threshold – All pixels above this greyscale level are inverted.

Although I'm not sure if `float` would also be possible for floating point images (mode `F`).

pmeier (commented):
Hey @oke-aditya and sorry for the delay. I have some comments inline.

torchvision/transforms/functional_pil.py
```diff
@@ -104,7 +104,12 @@ def adjust_hue(img, hue_factor):
 
 
 @torch.jit.unused
-def adjust_gamma(img, gamma, gain=1):
+def adjust_gamma(
+    img: Image.Image,
+    gamma: float,
+    gain: float = 1,
```
datumbox, line 110:
```suggestion
    gain: float = 1.0,
```

oke-aditya (commented):
I think I have fixed the bugs.
Should we unblock this file from mypy.ini ?

torchvision/transforms/functional_pil.py
```diff
@@ -271,7 +306,15 @@ def affine(img, matrix, interpolation=0, fill=None):
...
 
 @torch.jit.unused
-def rotate(img, angle, interpolation=0, expand=False, center=None, fill=None):
+def rotate(
+    img: Image.Image,
+    angle: float,
+    interpolation: int = 0,
+    expand: bool = False,
+    center: Optional[Tuple[int, int]] = None,
+    fill: Optional[Union[numbers.Number, List[numbers.Number], Tuple[numbers.Number, ...]]] = 0,
```
oke-aditya, line 315: Here

torchvision/transforms/functional_pil.py
```diff
@@ -280,7 +323,13 @@ def rotate(img, angle, interpolation=0, expand=False, center=None, fill=None):
 
 
 @torch.jit.unused
-def perspective(img, perspective_coeffs, interpolation=Image.BICUBIC, fill=None):
+def perspective(
+    img: Image.Image,
+    perspective_coeffs: float,
+    interpolation: int = Image.BICUBIC,
+    fill: Optional[Union[numbers.Number, List[numbers.Number], Tuple[numbers.Number, ...]]] = 0,
```
oke-aditya, line 330: Here

torchvision/transforms/functional_pil.py
```diff
@@ -261,7 +290,13 @@ def _parse_fill(fill, img, name="fillcolor"):
 
 
 @torch.jit.unused
-def affine(img, matrix, interpolation=0, fill=None):
+def affine(
+    img: Image.Image,
+    matrix: List[float],
+    interpolation: int = 0,
+    fill: Optional[Union[numbers.Number, List[numbers.Number], Tuple[numbers.Number, ...]]] = 0,
```
oke-aditya, line 297: Here

oke-aditya (commented):
Just marking the lines where error occured

oke-aditya:
Hmmm tests seem to be failing on `.pad()` function. 
Not very sure why it occured, but I think that has easy fix by passing args as `padding=pad`.

Also another point I noticed that. 

The functional tensor has similar [padding code](https://github.com/pytorch/vision/blob/master/torchvision/transforms/functional_tensor.py#L410), and I think that it accepts `float`, `int` as well as List and Tuples of them.
But the typing seems to be only `List[int]`.

I think that the typing for Functional PIL and Functional Tensor should be same. Since that's how we [test them](https://github.com/pytorch/vision/blob/master/test/test_functional_tensor.py#L849) as well.

torchvision/transforms/functional_pil.py
```diff
@@ -121,7 +126,13 @@ def adjust_gamma(img, gamma, gain=1):
 
 
 @torch.jit.unused
-def pad(img, padding, fill=0, padding_mode="constant"):
+def pad(
+    img: Image.Image,
+    padding: Union[int, List[int], Tuple[int, ...]],
+    fill: Optional[Union[numbers.Number, List[numbers.Number], Tuple[numbers.Number, ...]]],
```
pmeier, line 132: 1. `mypy` treats `int` as virtual subtype of `float`. So if we substitute `numbers.Number` with `float`, we should be good even if we set the default argument to `0`.
2. Tests fail here, because we missed the ` = 0` default argument.

pmeier (commented):
> We have stumbled upon [mypy/#3186](https://github.com/python/mypy/issues/3186) issue because of which mypy still complains.
> It thinks `int` is not a number and hence we cannot assign `0` as default arg.

Good catch. I've added a fix inline.

> Hmmm tests seem to be failing on `.pad()` function.
> Not very sure why it occured, but I think that has easy fix by passing args as `padding=pad`.

Fix inline.

> The functional tensor has similar [padding code](https://github.com/pytorch/vision/blob/master/torchvision/transforms/functional_tensor.py#L410), and I think that it accepts `int` as well as List and Tuples of them.
> (Basically same as `Union[int, List[int], Tuple[int, ...]],`)
> 
> But the typing seems to be only `List[int]`.
> 
> I think that the typing for Functional PIL and Functional Tensor should be same. Since that's how we [test them](https://github.com/pytorch/vision/blob/master/test/test_functional_tensor.py#L849) as well.

Yes the signatures and in turn the annotations should be the same. Would you send a follow-up PR after this one is merged?

torchvision/transforms/functional_pil.py
oke-aditya: Yeah I just spotted a minute ago too. :smile: (Coincidence)

I will cover these small fixes up and in a follow up PR ensure that the annotations between `functional_pil` and `functional_tesnor ` match properly. Not just on the pad case.

oke-aditya:
@datumbox @pmeier 
I think so CI is green enough to merge this. Probably all the failures are unrelated.

Hopefully, Good to go!

torchvision/transforms/functional_pil.py
```diff
@@ -121,7 +126,13 @@ def adjust_gamma(img, gamma, gain=1):
 
 
 @torch.jit.unused
-def pad(img, padding, fill=0, padding_mode="constant"):
+def pad(
+    img: Image.Image,
+    padding: Union[int, List[int], Tuple[int, ...]],
+    fill: Optional[Union[float, List[float], Tuple[float, ...]]] = 0,
```
datumbox, line 132: This is defined as int on the tensor equivalent method, though it supports both. Do we want them aligned?

torchvision/transforms/functional_pil.py
```diff
@@ -196,15 +207,28 @@ def pad(img, padding, fill=0, padding_mode="constant"):
...
         raise TypeError('img should be PIL Image. Got {}'.format(type(img)))
 
     return img.crop((left, top, left + width, top + height))
 
 
 @torch.jit.unused
-def resize(img, size, interpolation=Image.BILINEAR, max_size=None):
+def resize(
+    img: Image.Image,
+    size: Union[Sequence[int], int],
```
datumbox, line 227: The tensor API uses List instead of Sequence. Do we want them aligned?

torchvision/transforms/functional_pil.py
```diff
@@ -242,7 +266,12 @@ def resize(img, size, interpolation=Image.BILINEAR, max_size=None):
 
 
 @torch.jit.unused
-def _parse_fill(fill, img, name="fillcolor"):
+def _parse_fill(
+    fill: Optional[Union[float, List[float], Tuple[float, ...]]],
```
datumbox, line 270: This is also affected if we turn the above to int.

torchvision/transforms/functional_pil.py
```diff
@@ -261,7 +290,13 @@ def _parse_fill(fill, img, name="fillcolor"):
 
 
 @torch.jit.unused
-def affine(img, matrix, interpolation=0, fill=None):
+def affine(
+    img: Image.Image,
+    matrix: List[float],
+    interpolation: int = 0,
+    fill: Optional[Union[float, List[float], Tuple[float, ...]]] = 0,
```
datumbox, line 297: FYI strangely the tensor API defines this as float as well.

torchvision/transforms/functional_pil.py
```diff
@@ -261,7 +290,13 @@ def _parse_fill(fill, img, name="fillcolor"):
 
 
 @torch.jit.unused
-def affine(img, matrix, interpolation=0, fill=None):
+def affine(
+    img: Image.Image,
+    matrix: List[float],
+    interpolation: int = 0,
```
datumbox, line 296: Perhaps we should us Image.NEAREST instead of the numeric here?

torchvision/transforms/functional_pil.py
```diff
@@ -271,7 +306,15 @@ def affine(img, matrix, interpolation=0, fill=None):
 
 
 @torch.jit.unused
-def rotate(img, angle, interpolation=0, expand=False, center=None, fill=None):
+def rotate(
+    img: Image.Image,
+    angle: float,
+    interpolation: int = 0,
```
datumbox, line 312: same here

datumbox (commented):
@oke-aditya I checked this again. Left a couple of comments to get your input. It's mainly around whether we want to align the API between PIL and Tensor or if we want to keep the two separate. No strong opinions here. @pmeier Let me also know what you think please.

I know this PR turned out to be lengthy so if you had enough we can address them later. Cheers! :)

torchvision/transforms/functional_pil.py
oke-aditya: I think we want float in tensor equivalent method to align this.

torchvision/transforms/functional_pil.py
oke-aditya: Yes, I think we need to have Sequence in tensor, to align. But probably JIT does not support it.

torchvision/transforms/functional_pil.py
oke-aditya: Hmm, something stranger is that in the tensor API we pass a string.
I'm not sure how would we align it. (again probably JIT?)

torchvision/transforms/functional_pil.py
oke-aditya: Actually the tensor API defines it to be only `List[float]` and not all the three, again something which isn't aligned or probably has an issue.

torchvision/transforms/functional_pil.py
oke-aditya: We could change it as you suggested, but again this is inconsistent with functional_tensor.

oke-aditya (commented):
Hey @datumbox ! Here are my thoughts.

There are probably multiple places where we have misaligned with functional_tensor.

My idea was to follow-up this with a new PR. (#4282 ).


Let me know what you think!

torchvision/transforms/functional_pil.py
datumbox: no strong opinion here. I leave it up to you.

torchvision/transforms/functional_pil.py
datumbox: It's the union that is not supported. I just highlight that here the type (float vs int) is actually aligned. Which is weird cause we don't seem to keep things aligned. I probably agree that it's the tensor that needs changing though.

torchvision/transforms/functional_pil.py
datumbox: So the reason why I mention this is because on other methods in this file you use the "enum" while in this case you use the integer equivalent (see `perspective` for example). I think the enum will be clearer.

torchvision/transforms/functional_pil.py
datumbox: I mean something different. See the next comment.

torchvision/transforms/functional_pil.py
oke-aditya: Aah cool, I get it. I will rewrite with Enum :smile:

datumbox: yeah sorry I should have been clearer. OK ping me when this is done and I'll approve. Thanks for the effort and sorry for the back and forth. :)

oke-aditya:
I think this should be do it! @datumbox can you just cross check if I have kept the modes `BILINEAR` `CUBIC` and `NEAREST` correct as per old definitions?

torchvision/transforms/functional_pil.py
oke-aditya: I just verified Sequence is not supported by JIT. And yes Union too is not supported.

https://pytorch.org/docs/stable/jit_language_reference.html#supported-type

datumbox (approved):
LGTM, thank you @oke-aditya for your contribution and for your patience. :) 

I checked the modes and they look fine to me.

@pmeier Do you mind also having a look that we are good to go? Once I get your stamp we can merge.

torchvision/transforms/functional_pil.py
pmeier: Yes, `torch.jit` does not support `Sequence`. Since we also have a `Union` here, we can't align them anyway. Thus, I think it is ok to go with `Sequence`.

pmeier (approved):
> can you just cross check if I have kept the modes `BILINEAR` `CUBIC` and `NEAREST` correct as per old definitions?

Seems to be good, even replacing an old `0` with `NEAREST` :+1:

LGTM, thanks a lot @oke-aditya!

PR #4237, opened 2021-07-31
Fixed typing annotations of torchvision/ops

frgfm:
Following up on #2025, this PR fixes typing annotations in transforms/ops, and edits the `mypy.ini` to stop ignoring errors from this module.

Any feedback is welcome!

### DIFF
--- a/torchvision/ops/poolers.py
+++ b/torchvision/ops/poolers.py
@@ -5,7 +5,7 @@ import torchvision
 from .roi_align import roi_align
 from torchvision.ops.boxes import box_area
 
-from typing import Optional, List, Dict, Tuple, Union, Sequence
+from typing import Optional, List, Dict, Tuple, Union
 
 
 # copying result_idx_in_level to a specific index in result[]
@@ -134,7 +134,7 @@ class MultiScaleRoIAlign(nn.Module):
         self.featmap_names = featmap_names
         self.sampling_ratio = sampling_ratio
         self.output_size = (output_size, output_size) if isinstance(output_size, int) else tuple(output_size)
-        self.scales: Optional[Sequence[float]] = None
+        self.scales: Optional[List[float]] = None
         self.map_levels = None
         self.canonical_scale = canonical_scale
         self.canonical_level = canonical_level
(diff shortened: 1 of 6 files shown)


torchvision/ops/poolers.py
```diff
@@ -131,12 +131,10 @@ def __init__(
...
     ):
         super(MultiScaleRoIAlign, self).__init__()
-        if isinstance(output_size, int):
-            output_size = (output_size, output_size)
         self.featmap_names = featmap_names
         self.sampling_ratio = sampling_ratio
-        self.output_size = tuple(output_size)
-        self.scales = None
+        self.output_size = (output_size, output_size) if isinstance(output_size, int) else tuple(output_size)
+        self.scales: Optional[Sequence[float]] = None
```
oke-aditya, line 137: I think JIT is not happy here! From circle CI [build](https://app.circleci.com/pipelines/github/pytorch/vision/9650/workflows/638e29aa-f199-426b-9c15-ea0106ab0170/jobs/715732).

```
def infer_scale(self, feature: Tensor, original_size: Sequence[int]) -> float:
                                                          ~~~~~~~~~~~~~ <--- HERE
        # assumption: the scale is of the form 2 ** (-k), with k integer
        size = feature.shape[-2:]
'MultiScaleRoIAlign.infer_scale' is being compiled since it was called from 'MultiScaleRoIAlign.setup_scales'
  File "/root/project/torchvision/ops/poolers.py", line 179
```

I think JIT does not support Sequence,

torchvision/ops/poolers.py
```diff
@@ -2,10 +2,10 @@
 from torch import nn, Tensor
 
 import torchvision
-from torchvision.ops import roi_align
+from .roi_align import roi_align
```
oke-aditya, line 5: I would like to know why.

oke-aditya (commented):
Hope you don't mind!

torchvision/ops/poolers.py
frgfm: mypy was confused with the import (since there is a `roi_align.py` file)
Since we're in the correct folder, switching to relative import cleared that!

datumbox:
@frgfm Thanks for the PR and apologies for taking so long time to provide feedback.

One concern I have is that mypy limits you too much and all of your workarounds are done not necessarily to improve the quality of the code but to please mypy. Perhaps in this specific case, ignoring mypy would be preferable. Let me know your thoughts.

frgfm:
> @frgfm Thanks for the PR and apologies for taking so long time to provide feedback.
> 
> One concern I have is that mypy limits you too much and all of your workarounds are done not necessarily to improve the quality of the code but to please mypy. Perhaps in this specific case, ignoring mypy would be preferable. Let me know your thoughts.

Hi there!
Happy to go with what you think is best :ok_hand: Do you mean completely ignoring `torchvision.ops`? Or ignoring some parts where mypy requires some edits?

datumbox:
@frgfm Let's ignore the `torchvision.ops` for now. There are simply too many ignores at the moment.

Issue #4231, opened 2021-07-30
"No module named lmdb" error encountered using LSUN dataset


import-antigravity:
## 🐛 Bug

I'm aware that the code for downloading the LSUN dataset is currently not implemented. I downloaded it manually, but then when I try to use the torchvision dataset I get the following error:

```
  File ".../python3.8/site-packages/torchvision/datasets/lsun.py", line 18, in __init__
    import lmdb
ModuleNotFoundError: No module named 'lmdb'
```

## To Reproduce

1. Download LSUN dataset manually and extract it
2. Attempt to use the `LSUN` object

## Expected behavior

No error

## Environment

Please copy and paste the output from our
[environment collection script](https://raw.githubusercontent.com/pytorch/pytorch/master/torch/utils/collect_env.py)
(or fill out the checklist below manually).

```
[pip3] numpy==1.20.3
[pip3] pytorch-fid==0.2.0
[pip3] pytorch-lightning==1.2.10
[pip3] torch==1.8.1
[pip3] torchmetrics==0.2.0
[pip3] torchvision==0.9.1
[conda] blas                      2.109                       mkl    conda-forge
[conda] blas-devel                3.9.0                     9_mkl    conda-forge
[conda] cudatoolkit               10.1.243             h036e899_8    nvidia
[conda] ffmpeg                    4.3                  hf484d3e_0    pytorch
[conda] libblas                   3.9.0                     9_mkl    conda-forge
[conda] libcblas                  3.9.0                     9_mkl    conda-forge
[conda] liblapack                 3.9.0                     9_mkl    conda-forge
[conda] liblapacke                3.9.0                     9_mkl    conda-forge
[conda] mkl                       2021.2.0           h726a3e6_389    conda-forge
[conda] mkl-devel                 2021.2.0           ha770c72_390    conda-forge
[conda] mkl-include               2021.2.0           h726a3e6_389    conda-forge
[conda] numpy                     1.20.3           py38h9894fe3_1    conda-forge
[conda] pytorch                   1.8.1           py3.8_cuda10.1_cudnn7.6.3_0    pytorch
[conda] pytorch-fid               0.2.0                    pypi_0    pypi
[conda] pytorch-lightning         1.2.10             pyhd8ed1ab_0    conda-forge
[conda] torchmetrics              0.2.0                    pypi_0    pypi
[conda] torchvision               0.9.1                py38_cu101    pytorch
```

[...]

PR #4239, opened 2021-08-02
Add docs to LSUN dataset to install lmdb package

NicolasHug:
This PR indicates that `lmdb` should be installed to use the LSUN dataset class

Addresses https://github.com/pytorch/vision/issues/4231#issuecomment-889847142


fmassa (approved):
Thanks!

PR #4248, opened 2021-08-04
Add CUDA-11.3 builds to torchvision

malfet:
Also mark  `.cicrleci/config.yml` as autogenerated in `.gitattributes`


janeyx99 (approved):
seems like CI passes, but :/ about introducing binary conda 11.3 errors

Issue #4247, opened 2021-08-04
Use numpy RandomState objects in tests instead of drawing from numpy's global RNG


NicolasHug:
We're using `np.random` in a few places in our  tests:

```
(pt) ➜  vision git:(master) ✗ git grep np.random
test/common_utils.py:    np.random.seed(seed)
test/test_datasets.py:        labels = np.random.randint(0, self._VERSION_CONFIG["num_categories"], size=num_images).tolist()
test/test_image.py:        pixels = np.random.rand(*shape) > 0.5
test/test_image.py:        pixels = np.random.rand(*shape) > 0.5
test/test_transforms.py:        ndarray = np.random.randint(low=0, high=255, size=(height, width, channels)).astype(np.uint8)
test/test_transforms.py:        ndarray = np.random.rand(height, width, channels).astype(np.float32)
test/test_transforms.py:            trans(np.random.rand(1, height, width).tolist())
test/test_transforms.py:            trans(np.random.rand(height))
test/test_transforms.py:            trans(np.random.rand(1, 1, height, width))
test/test_transforms.py:        np_arr = np.random.randint(0, 255, (32, 32, 3), dtype=np.uint8)
test/test_transforms.py:        input_data = np.random.randint(low=0, high=255, size=(height, width, channels)).astype(np.uint8)
test/test_transforms.py:        input_data = torch.as_tensor(np.random.rand(channels, height, width).astype(np.float32))
test/test_transforms.py:            trans(np.random.rand(1, height, width).tolist())
test/test_transforms.py:            trans(np.random.rand(1, height, width))
test/test_transforms.py:    x_np = np.random.randint(0, 256, x_shape, np.uint8)
test/test_transforms.py:    x_np = np.random.randint(0, 256, x_shape, np.uint8)
test/test_transforms_video.py:            trans(np.random.rand(numFrames, height, width, 1).tolist())
```

this is using the global numpy RNG, which isn't ideal. We should instead instanciate a `RandomState` object in each test to contain the RNG into a single local object. For example, instead of calling `np.random.randint(...)` we should do:

```py
np_rng = np.random.RandomState(0)  # or whatever seed you might prefer
np_rng.randint(...)
```

Same for all calls to `np.random.xyz()`. This local `np_rng` object is only alive during the function call and thus doesn't leak or alter the global numpy RNG.

[...]

PR #4250, opened 2021-08-04
using `np.random.RandomState(seed)` instead of `np.random.seed(seed)`

vmoens:
closing #4247 

This PR makes use of `np.RandomState()` locally in the tests instead of the previous `np.random.seed()`, so as to contain numpy's  RNG locally to test functions instead of leaking the RNG seed globally.


test/test_datasets.py
```diff
@@ -21,6 +21,8 @@
 import torch.nn.functional as F
 from torchvision import datasets
 
+random_state_numpy = np.random.RandomState(0)
```
NicolasHug, line 24: we should create a local RandomState object in each test function   instead of having a global one in each file. Otherwise, tests are still dependent on each other withing a single module.

Also, no strong opinion on that but `np_rng` would be shorter and still be a descriptive name, so I'd suggest to use that instead

torchvision/models/inception.py
```diff
@@ -123,7 +125,9 @@ def __init__(
                     import scipy.stats as stats
                     stddev = m.stddev if hasattr(m, 'stddev') else 0.1
                     X = stats.truncnorm(-2, 2, scale=stddev)
-                    values = torch.as_tensor(X.rvs(m.weight.numel()), dtype=m.weight.dtype)
+                    values = torch.as_tensor(
+                        X.rvs(m.weight.numel(), random_state=random_state_numpy),
```
NicolasHug, line 129: oh, during our convo I had missed that the `rvs` was within the model's code, I thought it was just in the test. That's a bigger problem than I thought

@fmassa, googlenet and inception call scipy's rvs methods, which draw from numpy's RNG.

That's fairly unexpected I think. Shouldn't we just be relying on pytorch's RNG? I'm thinking of the following workarounds:

- add a new np_random_state parameter to the constructor to control that RNG
- rely on torch instead of numpy to draw samples from a truncated normal.

I think the second would make much more sense, although I don't know how easy this will be. WDYT?


As a temporary workaround we could use a pytest fixture that sets numpy's RNG and restores it as e.g. in https://gist.github.com/VictorDarvariu/6cede9c79900c6215b5f848993d283c6, but ugh

torchvision/models/inception.py
```diff
@@ -69,7 +70,8 @@ def __init__(
         aux_logits: bool = True,
         transform_input: bool = False,
         inception_blocks: Optional[List[Callable[..., nn.Module]]] = None,
-        init_weights: Optional[bool] = None
+        init_weights: Optional[bool] = None,
+        random_state_numpy: np.random.RandomState = np.random.RandomState(),
```
NicolasHug, line 74: if we ever add a new parameter (I'm not sure we should for now, but we might have to), the default should be `None`

NicolasHug (commented):
Thanks @vmoens , this looks good

This rvs issue is unfortunate, I commented below, let's see what @fmassa thinks

torchvision/models/inception.py
fmassa: Hum, we don't pass RNG in any of torchvision functions as of now, so I would rather not do it here, as it would involve a larger discussion.

torchvision/models/inception.py
fmassa: > That's fairly unexpected I think. Shouldn't we just be relying on pytorch's RNG? I'm thinking of the following workarounds:

We should, but PyTorch didn't implement `trunc_normal` back when we first implemented this model.
It seems now that it's implemented in https://github.com/pytorch/pytorch/pull/32397 , so we could replace it to use PyTorch's implementation. It should be fairly straightforward, but it would be good to check if the PyTorch sampling is much slower than scipy's or not, and to make this change in a separate PR.

NicolasHug: @vmoens Would you like a submit a PR to change googlenet and inception to rely on `torch.nn.init.trunc_normal_` instead?

I think we can keep this one on hold until then

vmoens: Sure let me do this: 
- keep `np.random.seed(0)` in set_seed for now in this PR
- do a new PR where `np.random.seed(0)` is taken away and we use `trunc_normal_` for inception

NicolasHug: To keep matters separate, we should aim at removing all numpy seedings in one go, so the first step might unnecessary. We'll need to get rid of the `rvs` calls first

test/test_image.py
```diff
@@ -273,8 +273,9 @@ def test_write_file_non_ascii():
 ])
 def test_read_1_bit_png(shape):
     with get_tmp_dir() as root:
+        np_rng = np.random.RandomState(0)
```
NicolasHug, line 276: as a very minor nit (nitpick =  feel free not to address): the declaration of `np_rng` doesn't need to be within the context manager. Same below

test/test_models.py
```diff
@@ -193,7 +192,7 @@ def _check_fx_compatible(model, inputs):
 # the _test_*_model methods.
 _model_params = {
     'inception_v3': {
-        'input_shape': (1, 3, 299, 299)
+        'input_shape': (1, 3, 299, 299),
```
NicolasHug, line 195: I realize this is due to previous changes that were reverted, but in general we try to avoid unrelated changes, as it obfuscates git blame. Same for the removal of `import sys` above, which on his own is actually relevant (and doesn't hurt git blame), but having lots of those in a single PR makes review more difficult. Would you mind reverting those changes?

test/test_transforms.py
```diff
@@ -200,18 +200,20 @@ class TestToTensor:
     def test_to_tensor(self, channels):
         height, width = 4, 4
         trans = transforms.ToTensor()
+        np_rng = np.random.RandomState(0)
```
NicolasHug, line 203: I think we can remove this one,  as you're already declaring another RandomState below

test/test_transforms.py
```diff
@@ -225,22 +227,25 @@ def test_to_tensor(self, channels):
     def test_to_tensor_errors(self):
         height, width = 4, 4
         trans = transforms.ToTensor()
+        np_rng = np.random.RandomState(0)
```
NicolasHug, line 230: Let's remove this one too

NicolasHug (approved):
Thanks @vmoens !! I have some minor comments below, but I'll approve now so you can address and merge once ready

test/test_image.py
vmoens: so is the rest of the test :p strictly speaking, only `image_path = os.path.join(root, f'test_{shape}.png')` needs to be there

test/test_models.py
vmoens: I agree i'll revert these!

test/test_transforms.py
vmoens: my bad yeah i missed that one

test/test_transforms.py
vmoens: I think they come from a rather chaotic `git revert` :) should have checked before pushing though

test/test_image.py
NicolasHug: Well we do need `image_path` to be valid, so technically we could only remove the last 2 lines. But yeah, it's a nit anyway :)

Issue #4249, opened 2021-08-04
What Kitti dataset are we supporting?


NicolasHug:
We added support for the Kitti dataset in https://github.com/pytorch/vision/pull/3640 but it's not clear which one we are vendoring, as there are many different versions of the dataset : http://www.cvlibs.net/datasets/kitti/index.php

@pmeier @fmassa @prabhat00155

NicolasHug:
After a bit of trial and error, looking at the URLs it seems to be the "left color images" dataset for object detection

PR #4251, opened 2021-08-04
Indicate which Kitti dataset we're vendoring

NicolasHug:
Closes https://github.com/pytorch/vision/issues/4249


fmassa (approved):
Thanks @NicolasHug !

PR #4253, opened 2021-08-05
Added torchvision maintainers guide

NicolasHug:
This PR adds the torchvision maintainers guide with different guidelines regarding public/private API etc., as discussed internally


NicolasHug:
I'm converting the PR to draft, as some minor edits are to be expected

NicolasHug:
I updated the doc following our previous discussion @mthrok @nairbv @dgenzel2 @fmassa 

I believe this is ready to be merged now.

maintainer_guide.md
```diff
@@ -0,0 +1,80 @@
...
+When a public-facing feature needs to change, we try to deprecate, but sometimes
+a BC-breaking change is possible or even warranted. Typically in the case of
+bugs, a failure is better than incorrect results. To decide whether to deprecate
+or to introduce a breaking change:
+
+1. Check whether X does something purely internal VS something that a reasonable
+   user might be interested in.
+2. Look for usages on Github for usages of X. This is often difficult due to the
+   forks, copy-pastes, and name conflicts.
+3. Look on FBcode where we can get a cleaner (but also biased) signal of the
```
datumbox, line 65: References to FBcode feel a bit strange here. Not all contributors can do this, so this probably needs to be rephrased to cover these cases (for example "sync with a FB contributor to do this for you". Quite common advice given also on PyTorch core).

maintainer_guide.md
```diff
@@ -0,0 +1,80 @@
...
+truth, but rather as a handy reference guide.
+
+### What is public?
+
+* In the torchvision folder:
+    * For Python, the [usual
+      convention](https://www.python.org/dev/peps/pep-0008/#public-and-internal-interfaces)
+      applies: anything that has a leading underscore in its name or in its path
+      is private. The rest is public, even if it’s not properly documented, and
+      even if it’s not exposed in an `__init__` file. While we discourage use of
```
mthrok, line 14: > The rest is public, even if it’s not properly documented, and even if it’s not exposed in an `__init__` file. 

The use of word "public" in this statement feels to me giving out impression that these interfaces are BC-guaranteed.

If the goal of this section is to encourage developers and maintainers to think twice about the backward compatibility of such interfaces, I think we can focus on reality more.

```
Interfaces excluded from documentation and `__init__` attributes are not guaranteed for backward compatibility. (For reasons such as residuals from early stage of the project) Therefore, users should not rely on them, (or users find them useful are encourage to file an issue to make them into a public interface) however, maintainers must still consider backwards compatibility when making changes to such potentially in-use code.
```

maintainer_guide.md
```diff
@@ -0,0 +1,80 @@
...
+      on torchvision ops. FYI PyTorch as BC API checks for its ops.
+* The test folder is not importable and is **private.** Even fbcode projects
+  should not rely on it. 
+* The references folder serves as documentation and is treated as such: it’s
+  “private” in the sense that there are no BC guarantees here. Breaking changes
+  are possible, but we should make sure that the tutorials are still running
+  properly, and that their intended narrative is preserved (by e.g. checking
+  outputs, etc.).
+* The rest of the folders (build, android, ios) are private and have no BC
+  breaking guarantees.
```
mthrok, line 28: > and have no BC breaking guarantees.

Isn't it "no backward compatibility guarantees"?

maintainer_guide.md
NicolasHug: > The use of word "public" in this statement feels to me giving out impression that these interfaces are BC-guaranteed.

Yes, unfortunately they are. I believe this is what we all agreed on during our internal discussions.

maintainer_guide.md
NicolasHug: I added

> if you're not a fb employee, ask someone to do that for you.

But just to clarify, the audience of this document is the torchvision maintainers (fb employees + quantsight), not the general pool of contributors - although they might find some relevant info here too. fb and quantsight employees all know what fbcode is, which is why we're using the term "fbcode" as well above without further details

datumbox: Quantisight engineers wouldn't have access to FBcode either, so they still need to pair with a FB employee to do such checks.

Perhaps then the guide should leave in a public Google Docs (for transparency reasons) but mainly shared across the torchvision maintainers? Adding it inside the repo if it does not target the general pool of contributors might be confusing. Thoughts?

NicolasHug: It's very common for projects to include maintainers-only info in their repo / docs:

- https://scikit-learn.org/stable/developers/maintainer.html
- https://pandas.pydata.org/pandas-docs/stable/development/maintaining.html

There's usually no reason not to do so, and it help build trust / transparency. It's also easier to have everything at the same place.

nairbv: There's some subtle difference in whether we think of this as "an internal doc, posted publicly" or "a public explanation of our process"

I'm onboard with being transparent. If we're being transparent we can use language that's accessible to the readers to whom we're being transparent. E.g. if it's an internal only doc then terms like `fbcode` which reference a particular file path make sense. If we're being transparent so external contributors can see and understand our process, it seems slightly nicer to write out things like "facebook internal code" instead of referencing the directory name or internal abbreviations.

In some cases a statement like "We should/will/must check for..." might feel more inclusive than an imperative command directed at the reader ("Check for...") so that there's less of an implication that the reader is a FB employee.

though, I do notice examples of the imperative form in the referenced maintainer guides.

maintainer_guide.md
nairbv: What I remember from our discussion was that we want to communicate that maintainers should consider and follow appropriate steps for BC-breaking changes to these kinds of APIs, but that we should avoid implying a particular guarantee to users.

What a user should expect is a different question we could address elsewhere, but we can write our maintainer instructions without additional implications for users. I don't think we have any absolute "guarantees" of backwards compatibility, just that we expect maintainers to follow this process when breaking BC.

Elsewhere we can communicate different standards for backwards compatibility, which might consider whether the API is documented. Our actual communicated-to-users level of BC commitment might include differences for major/minor releases vs nightlies/master, documented vs undocumented APIs, etc.

maintainer_guide.md
NicolasHug: > whether we think of this as "an internal doc, posted publicly" or "a public explanation of our process"

I think of it as both, with a stronger component on the first definition. It *may* be useful as a public explanation of our process in some rare cases, but it's not the main purpose.

Should non-maintainers read this document, I think it's fair to assume they would understand what "fbcode" means in this context, the name is fairly self-explanatory.

In pytorch core, there are many explicit references to fbcode:

```
git grep -i fbcode | wc -l
127
```

I'm happy to edit the document if you really insist on removing internal terms and language, but in all honesty, I don't think it will matter :)

maintainer_guide.md
NicolasHug: I agree with your comment @nairbv 

Regarding of the use of the term "public", I'm not sure why this is so controversial: we're simply following pep8 here.

To clarify my point above:


> > The use of word "public" in this statement feels to me giving out impression that these interfaces are BC-guaranteed.
> 
> Yes, unfortunately they are. I believe this is what we all agreed on during our internal discussions.


An interface  can be either public or private. From pep8, an interface is private if and only if it has a leading underscore, and it's public otherwise. Private interfaces can be broken without BC considerations. Public interfaces cannot be broken without BC considerations.

What those *considerations* are is **not** the purpose of this section. They are instead (unfortunately only partially described) below.

As an important datapoint that I already mentionned, scikit-learn and numpy both consider that all public interfaces follow the same deprecation cycles, whether those were initially intended to be effectively public or not.

Issue #3562, opened 2021-03-12
[RFC] New datasets to torchvision


oke-aditya:
## 🚀 Feature

This is a proposal to add more highly cited datasets. Thanks to [papers with code datasets](https://paperswithcode.com/datasets) which made this search easy. 

## Motivation

These datasets are used quite frequently and would provide benefits to both researchers as well as people who work in computer vision. I'm not sure of the citation metric, but we can verify the count of papers once.

## Pitch

The following datasets can be considered. Papers are reported as per the last 5 years count on papers with code. They can be inaccurate, feel free to edit. I'm also adding previously 

- [ ] [KITTI](https://paperswithcode.com/dataset/kitti) 1709 papers
- [ ] [LFW Labeled Faces in Wild](https://paperswithcode.com/dataset/lfw) 640 papers
- [ ] [Market-1501](https://paperswithcode.com/dataset/market-1501) 492 papers
- [ ] [MPII Human Pose](https://paperswithcode.com/dataset/mpii)
- [ ] [VGGFace2](https://paperswithcode.com/dataset/vggface2-1) Earlier requested in #1193 #2910 [Here](https://www.robots.ox.ac.uk/~vgg/data/vgg_face/vgg_face_dataset.tar.gz) is tar.gz file. Hopefully we can add it
- [ ] [MovingMNIST](http://www.cs.toronto.edu/~nitish/unsupervised_video/)  Perviously approved in #2676  #2690.
- [ ] [iNaturalist](https://www.inaturalist.org/) #3292 
- [ ] [LVIS](https://www.lvisdataset.org) 

Probably, we should think and add these, one by one. Also support downloading, not just loading of the dataset.

## Additional context

I would be very glad to send PRs, but I would need help.

fmassa:
Hi,

This is exactly our current idea, thanks for bringing it up.

I agree with all the aforementioned proposals. One thing to mention as well is that there is an ongoing effort to provide new dataset abstractions in PyTorch via DataPipes https://github.com/pytorch/pytorch/issues/49440.

While this doesn't block us providing new datasets, it is good to keep in mind that we might in the future revisit the way we implement datasets.

seyeeet:
related to  this issue, it can also be useful if pytorch can store this datasets on their storage and provide link to download them.
e.g. there are a lot of issues with downloading imagenet and other large datasets, im not sure if licensing can be problematic, but it would be super useful

pmeier:
@seyeeet 

[...]

PR #4255, opened 2021-08-05
Added LFW Dataset

ABD-01:
Hi!
This pr is with reference to https://github.com/pytorch/vision/issues/3562

I added the dataset class for [LFW Labeled Faces in Wild](https://paperswithcode.com/dataset/lfw)

Please do comment for the required changes, and other things to be done.
Thank You


pmeier (commented):
Hey @ABD-01 and thanks for the PR. Sorry about the delay as I was on vacation.

I have a few minor comments inline and only reviewed `LFW_People`, since most of the comments also apply to `LFW_Pairs`. Apart from that I have three more comments:

1. The two datasets have a very similar structure. Since they return different things, I think having two separate classes is fine to avoid polymorphism. Still, I wonder if it makes sense to have a common superclass, e.g. `_LFW` that groups the common behavior such as downloads.
2. We have automated tests for datasets located in [test/test_datasets.py](https://github.com/pytorch/vision/blob/master/test/test_datasets.py). Can you please add tests for these datasets as well? The main challenge is to provide "fake data", i.e. the (extracted) data that would be present after `.download()` is completed. It should resemble the structure of the original dataset, but only provide a few dummy images. Let me know if you need help with that.
3. Please also add the datasets to the list in [`torchvision/datasets/__init__.py`](https://github.com/pytorch/vision/blob/master/torchvision/datasets/__init__.py) as well to our documentation in [`docs/source/datasets.rst`](https://github.com/pytorch/vision/blob/master/docs/source/datasets.rst). Otherwise they are not discoverable.

torchvision/datasets/lfw.py
```diff
@@ -0,0 +1,260 @@
...
+    download_url_prefix = "http://vis-www.cs.umass.edu/lfw/"
+
+    file_dict = {
+        'lfw': ("lfw.tgz", "a17d05bd522c52d84eca14327a23d494"),
+        'lfw_funneled': ("lfw-funneled.tgz", "1b42dfed7d15c9b2dd63d5e5840c86ad"),
+        'lfw-deepfunneled': ("lfw-deepfunneled.tgz", "68331da3eb755a505a502b5aacb3c201")
+    }
+
+    peopleDevTrain = " peopleDevTrain.txt"
+    peopleDevTest = " peopleDevTest.txt"
```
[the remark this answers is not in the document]
ABD-01: It was to download using those filenames but I wrote it like this:
```py
download_url(f"{self.download_url_prefix}peopleDev{self.split}.txt", self.root)
```
So yes, they aren't used anywhere, I'll remove these lines.

torchvision/datasets/lfw.py
```diff
@@ -0,0 +1,260 @@
...
+
+        if self.target_transform is not None:
+            target = self.target_transform(target)
+
+        return img, target
+
+    def loader(self, path: str) -> Image.Image:
+        with open(path, 'rb') as f:
+            img = Image.open(f)
+            return img.convert('RGB')
```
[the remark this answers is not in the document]
ABD-01: No there's no such reason, I wrote with considering the style of the `torchvision.datasets.Dataset` class. I'll move it in __getitem__ class.
Thanks for pointing that out.

ABD-01:
Hi, I have made new commits as per the comments.
The tests are remaining so I'll add them, could you please guide me on how to proceed with that?

torchvision/datasets/lfw.py
```diff
@@ -0,0 +1,218 @@
...
+        if not self._check_integrity():
+            raise RuntimeError('Dataset not found or corrupted.' +
+                               ' You can use download=True to download it')
+
+        self.images_dir = os.path.join(self.root, images_dir)
+
+    def loader(self, path: str) -> Image.Image:
+        with open(path, 'rb') as f:
+            img = Image.open(f)
+            return img.convert('RGB')
```
pmeier, line 50: I see where you getting at with that, but I'm not sure if it is worth it. At the very least we should make it private so other people do not depend on it.
```suggestion
    def _loader(self, path: str) -> Image.Image:
        with open(path, 'rb') as f:
            img = Image.open(f)
            return img.convert('RGB')
```

torchvision/datasets/lfw.py
```diff
@@ -0,0 +1,218 @@
...
+
+    def _check_integrity(self):
+        fpath = os.path.join(self.root, self.filename)
+        fname = os.path.join(self.root, f"{self.view}Dev{self.split}.txt")
```
pmeier, line 54: Can't we also have checksums for these files? Should be only four distinct files, right?

torchvision/datasets/lfw.py
```diff
@@ -0,0 +1,218 @@
...
+                for num in range(1, int(num_imgs) + 1):
+                    img = os.path.join(images_dir, identity, f"{identity}_{num:04d}.jpg")
+                    if os.path.exists(img):
+                        data.append(img)
+                        targets.append(cls)
+
+        return cls_to_names, data, targets
+
+    def __len__(self):
+        return len(self.data)
```
pmeier, line 121: This can be moved into `_LFW`.

pmeier (commented):
Hey @ABD-01 and thanks for the continuing effort. I have some minor comments inline. Apart from them, two more things:

1. Please also add the new datasets to our documentation as shortly outlined in https://github.com/pytorch/vision/pull/4255#pullrequestreview-732488800
2. I have not yet reviewed the `_get_people` and `_get_pairs` methods, as they really depend on the data format, which I don't know yet. However, this will be solved by the next point:

> The tests are remaining so I'll add them, could you please guide me on how to proceed with that?

Sure! You need to create two test cases, `LFWPeopleTestCase` and `LFWPairsTestCase`, in  [`tess_datasets.py`](https://github.com/pytorch/vision/blob/main/test/test_datasets.py) that inherit from 

https://github.com/pytorch/vision/blob/7947fc8fb38b1d3a2aca03f22a2e6a3caa63f2a0/test/datasets_utils.py#L180

Please read the docstring to get a better understanding what needs to be done. 

The absolute minimum that you need to provide, is an implementation of the `inject_fake_data` method. It gets passed a `tmp_dir` that needs to contain all files in the correct structure that would be present in `root` if I would instantiate the dataset with `LFWPeople(root, download=True)`. If the content changes depending on other keyword arguments, you can query them from the passed `config` dict, i.e. `config["train"]`. If you overwrite the class attribute `ADDITIONAL_CONFIGS` for example with 

```python
ADDITIONAL_CONFIGS = datasets_utils.combinations_grid(
    train=(True, False), image_set=("original", "funneled", "deepfunneled"),
)
```

you will get automate testing for all combinations. 

Of course the files provided by `inject_fake_data` musn't be the original files, but rather mock ups of them, i.e. "fake data". For example, you will want to create a `peopleDevTrain.txt` file, that only contains a few lines. For the accompanying images, you can use convenience utilities such as 

https://github.com/pytorch/vision/blob/7947fc8fb38b1d3a2aca03f22a2e6a3caa63f2a0/test/datasets_utils.py#L666

or 

https://github.com/pytorch/vision/blob/7947fc8fb38b1d3a2aca03f22a2e6a3caa63f2a0/test/datasets_utils.py#L702

For the LFW dataset, I propose you add a `_LFWTestCase` that handles the injection of fake data so that the actual test cases only need to handle their special stuff.

---

If all of the above is "too much", we can also do it step by step. Implement something and if you hit a roadblock, push the commit and ping me so I can take a look at it and guide you more precisely.

torchvision/datasets/lfw.py
```diff
@@ -0,0 +1,218 @@
...
+        image_set: str,
+        view: str,
+        transform: Optional[Callable] = None,
+        target_transform: Optional[Callable] = None,
+        download: bool = False,
+    ):
+        super(_LFW, self).__init__(os.path.join(root, self.base_folder),
+                                   transform=transform, target_transform=target_transform)
+
+        image_set = verify_str_arg(image_set.lower(), 'image_set', self.file_dict.keys())
```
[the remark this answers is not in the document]
ABD-01: Yes, that will be nice. I'll do the changes.

torchvision/datasets/lfw.py
ABD-01: Yes
1. peopleDevTrain.txt
2. peopleDevTest.txt
3. pairsDevTrain.txt
4. pairsDevTest.txt

I'll try to add md5 checksum for these files as well.

ABD-01:
Hey, sorry for the delay for this update (had some college work).
I have added the tests for [LFWPeopleTestCase](https://github.com/ABD-01/vision/blob/add_lfw/test/test_datasets.py#L1793-L1824) and used it as super class for [LFWPairsTestCase](https://github.com/ABD-01/vision/blob/add_lfw/test/test_datasets.py#L1827-L1852)

ABD-01:
I have updated the tests as all cases were failing before.
Now only the feature types test is failing and I can't figure out why is this happening.
```bash
======================================================================
FAIL: test_feature_types (__main__.LFWPairsTestCase) (idx=0, image_set='deepfunneled', train=True)
----------------------------------------------------------------------
  File "/root/vision/test/datasets_utils.py", line 553, in test_feature_types
    assert isinstance(feature, expected_feature_type)
AssertionError
```

pmeier (commented):
> Now only the feature types test is failing and I can't figure out why is this happening.

The test is not implemented in a way that lets you check the types inside of a container. By providing `(PIL.Image.Image, PIL.Image.Image)` as first element of the `FEATURE_TYPES` class attribute, the test will check if the first element of a returned sample is either a `PIL.Image.Image` **or** a `PIL.Image.Image`. Since it is tuple, the test fails. Weirdly, the introspection of `pytest` is "disabled" and this is why you don't get a proper error message (cc @NicolasHug ).

To fix this, you would need to define `FEATURE_TYPES = (tuple, int)`. But then we should discuss if this is the way to go. What is the reason you want to return the image pair in a tuple rather than individually, i.e. `return img1, img2, target`? We do it like that in our `PhotoTour` dataset:

https://github.com/pytorch/vision/blob/5ef75d7bc666c9af3dd05b62b075df4c23ca467c/torchvision/datasets/phototour.py#L118

Apart from that, IMO the PR is 95% done. Thanks a lot @ABD-01! I have one more minor comment inline.

ABD-01:
Thanks for the Review.

> What is the reason you want to return the image pair in a tuple rather than individually, i.e. return `img1, img2, target`?

We can do it like this as well. I guess that'll be better way. I was returning a tuple as we needed a pair of images and a label whether they are same or not.

Apart form that I have a small doubt, there's a `10 fold validation` ([pairs](http://vis-www.cs.umass.edu/lfw/pairs.txt) and [people](http://vis-www.cs.umass.edu/lfw/people.txt)) view of dataset, should I add it??
The problem is that the format is similar but not same, so how can we add that?
Can we have like: split: `Train`, `Test`, `10-fold`. And for `10-fold` use different function to get data?

pmeier:
> Apart form that I have a small doubt, there's a `10 fold validation` ([pairs](http://vis-www.cs.umass.edu/lfw/pairs.txt) and [people](http://vis-www.cs.umass.edu/lfw/people.txt)) view of dataset, should I add it??

Since you want to contribute this dataset, I'm guessing you want to use it for research or something related, right? If you know the field, is the fold variant used in other publications? If not, I'm fine to  table this for now and only add it if we get requests for it. If you want to add it, that would also be ok for me.

> The problem is that the format is similar but not same, so how can we add that? 
> Can we have like: split: `Train`, `Test`, `10-fold`. And for `10-fold` use different function to get data?

What format are you talking about?  Isn't the `10-fold` variant just a different selection of images rather than the train / test split?

test/test_datasets.py
```diff
@@ -1790,5 +1790,77 @@ def test_targets(self):
...
+
+    def _create_images_dir(self, root, idir, split):
+        idir = os.path.join(root, idir)
+        os.makedirs(idir, exist_ok=True)
+        num_pairs = 7  # effectively 7*2 = 14
+        with open(pathlib.Path(root) / f"pairsDev{split}.txt", "w") as file:
+            file.write(str(num_pairs))
+            for i in range(num_pairs):
+                name1 = datasets_utils.create_random_string(random.randint(5, 7)) + '_' \
+                    + datasets_utils.create_random_string(random.randint(4, 7))
```
pmeier, line 1847: By putting this in a method, e.g. `_create_image_id`, we could avoid some duplication. These lines appear four times in total for both test cases.

test/test_datasets.py
```diff
@@ -1790,5 +1790,77 @@ def test_targets(self):
...
+                no1, no2 = random.randint(1, 100), random.randint(1, 100)
+                file.write(f"\n{name1}\t{no1}\t{no2}")
+                datasets_utils.create_image_folder(idir, name1, lambda _: f"{name1}_{no1:04d}.jpg", 1, 250)
```
pmeier, line 1850: Same here for the `lambda`, which appears five times total.

pmeier (commented):
Two more nitpicks after I've looked at the tests again. Not blocking, but would improve the code quality.

ABD-01:
> What format are you talking about? Isn't the `10-fold` variant just a different selection of images rather than the train / test split?

Yes it's almost same, just that it is divided in 10 parts, whose each part can be viewed as `LFWPairs` or `LFWPeople` class

test/test_datasets.py
ABD-01: Yeah! That'd be great. Thanks for pointing it out.

pmeier:
> Yes it's almost same, just that it is divided in 10 parts, whose each part can be viewed as `LFWPairs` or `LFWPeople` class

Ok, but that means that each fold is similar to a split, right? If yes, instead of `train: bool` we could have `split: str`, which takes either `"train"`, `"test"`, as well as `fold_X` where `X` is the number of the split you want to select. What do you think?

Still, let's first answer if we want to go that route in this PR.

ABD-01:
> Ok, but that means that each fold is similar to a split, right? If yes, instead of `train: bool` we could have `split: str`, which takes either `"train"`, `"test"`, as well as `fold_X` where `X` is the number of the split you want to select. What do you think?
> 
> Still, let's first answer if we want to go that route in this PR.

I will try to add that functionality, if works out easily then fine, else we are good to go with the current code.

ABD-01:
Hey!, I did some [updates](https://github.com/pytorch/vision/pull/4255/commits/754951786cdcaf29c3ab16ea673db9e4e24a56f2?diff=split)
I will add the tests in short time

test/test_datasets.py
```diff
@@ -1790,5 +1790,78 @@ def test_targets(self):
                 assert item[6] == i // 3
 
 
+class LFWPeopleTestCase(datasets_utils.DatasetTestCase):
+    DATASET_CLASS = datasets.LFWPeople
+    FEATURE_TYPES = (PIL.Image.Image, int)
+    ADDITIONAL_CONFIGS = datasets_utils.combinations_grid(
+        train=(True, False),
```
pmeier, line 1797: Currently the CI is failing because with this the test is trying to create `LFWPeople(..., train=True, ...)` which is no longer possible. This should be 
```suggestion
        split=("train", "test", "10fold"),
```
right?

test/test_datasets.py
```diff
@@ -1790,5 +1790,78 @@ def test_targets(self):
...
+    }
+
+    def inject_fake_data(self, tmpdir, config):
+        tmpdir = pathlib.Path(tmpdir) / "lfw-py"
+        os.makedirs(tmpdir, exist_ok=True)
+        split = "Train" if config["train"] else "Test"
+        return dict(
+            num_examples=self._create_images_dir(tmpdir, self._IMAGES_DIR[config["image_set"]], split),
+            split=split
+        )
```
pmeier, line 1813: This needs to be adapted as well.

test/test_datasets.py
```diff
@@ -1790,5 +1790,78 @@ def test_targets(self):
...
+            datasets_utils.create_image_folder(idir, name, lambda n: f"{name}_{n+1:04d}.jpg", no, 250)
+            num_examples += no
+        with open(pathlib.Path(root) / f"peopleDev{split}.txt", "w") as f:
+            f.writelines(flines)
+
+        return num_examples
+
+    def _create_random_id(self):
+        return datasets_utils.create_random_string(random.randint(5, 7)) + '_' \
+            + datasets_utils.create_random_string(random.randint(4, 7))
```
pmeier, line 1834: Nit: let's try to avoid backslashes where ever possible

```suggestion
    def _create_random_id(self):
        part1 = datasets_utils.create_random_string(random.randint(5, 7))
        part2 = datasets_utils.create_random_string(random.randint(4, 7))
        return f"{part1}_{part2}"
```

pmeier (commented):
Hey @ABD-01, I have a hint about the failing CI as well as one nitpick inline. 

Apart from that, there are two things left for you to do at the moment: address https://github.com/pytorch/vision/pull/4255#discussion_r693680577 and https://github.com/pytorch/vision/pull/4255#discussion_r693683762. If that is done and tests are passing, I'll do one final pass, because I haven't reviewed the people / pair extraction yet. Please ping me if the PR is otherwise ready.

torchvision/datasets/lfw.py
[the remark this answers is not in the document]
ABD-01: Is there are way to add the `extra_repr` of super class in it's own instance??
Because the method is getting over-written in subclasses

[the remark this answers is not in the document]
pmeier: I would do something like 

```python
    def extra_repr(self):
        return f"split={self.split}, image_set={self.image_set}"
```

in the super class, and simply extend it in the subclasses:

```python
    def extra_repr(self):
        return super().extra_repr() + f", foo={self.foo}"
```

torchvision/datasets/lfw.py
ABD-01: The md5 for these files (`{'pairs'|'people'}{''|'DevTrain'|'DevTest'}.txt`) is not provided on their website, so should I use the values calculated using [`calculate_md5`](https://github.com/pytorch/vision/blob/main/torchvision/datasets/utils.py#L53) function?

Ig rest all is done.

ABD-01:
@pmeier The PR is ready !!

torchvision/datasets/lfw.py
```diff
@@ -0,0 +1,260 @@
...
+            puts it in root directory. If dataset is already downloaded, it is not
+            downloaded again.
+
+    """
+
+    def __init__(
+        self,
+        root: str,
+        split: str = "train",
+        image_set: str = "funneled",
```
pmeier, line 118: This is out of sync with the docstring.
```suggestion
        image_set: str = "original",
```

torchvision/datasets/lfw.py
```diff
@@ -0,0 +1,260 @@
...
+            n_folds, s = (int(lines[0]), 1) if self.split == "10fold" else (1, 0)
+
+            for fold in range(n_folds):
+                n_lines = int(lines[s])
+                people = [line.strip().split("\t") for line in lines[s + 1: s + n_lines + 1]]
+                s += n_lines + 1
+                for i, (identity, num_imgs) in enumerate(people):
+                    for num in range(1, int(num_imgs) + 1):
+                        img = self._get_path(identity, num)
+                        if os.path.exists(img):
```
pmeier, line 142: Is there a possibility an image does not exist unless the data is corrupted?

torchvision/datasets/lfw.py
```diff
@@ -0,0 +1,260 @@
...
+            puts it in root directory. If dataset is already downloaded, it is not
+            downloaded again.
+
+    """
+
+    def __init__(
+        self,
+        root: str,
+        split: str = "10fold",
+        image_set: str = "funneled",
```
pmeier, line 202: Same here

```suggestion
        split: str = "train",
        image_set: str = "original",
```

torchvision/datasets/lfw.py
```diff
@@ -0,0 +1,260 @@
...
+                n_folds, n_pairs = 1, int(lines[0])
+            s = 1
+
+            for fold in range(n_folds):
+                matched_pairs = [line.strip().split("\t") for line in lines[s: s + n_pairs]]
+                unmatched_pairs = [line.strip().split("\t") for line in lines[s + n_pairs: s + (2 * n_pairs)]]
+                s += (2 * n_pairs)
+                for pair in matched_pairs:
+                    img1, img2, same = self._get_path(pair[0], pair[1]), self._get_path(pair[0], pair[2]), 1
+                    if os.path.exists(img1) and os.path.exists(img2):
```
pmeier, line 229: Same question.

torchvision/datasets/lfw.py
```diff
@@ -0,0 +1,260 @@
...
+                        pair_names.append((pair[0], pair[0]))
+                        data.append((img1, img2))
+                        targets.append(same)
+                for pair in unmatched_pairs:
+                    img1, img2, same = self._get_path(pair[0], pair[1]), self._get_path(pair[2], pair[3]), 0
+                    if os.path.exists(img1) and os.path.exists(img2):
```
pmeier, line 235: Same question.

test/test_datasets.py
```diff
@@ -1801,5 +1801,87 @@ def test_targets(self):
...
+    ADDITIONAL_CONFIGS = datasets_utils.combinations_grid(
+        split=('10fold', 'train', 'test'),
+        image_set=('original', 'funneled', 'deepfunneled')
+    )
+    _IMAGES_DIR = {
+        "original": "lfw",
+        "funneled": "lfw_funneled",
+        "deepfunneled": "lfw-deepfunneled"
+    }
+    _split = {'10fold': '', 'train': 'DevTrain', 'test': 'DevTest'}
```
pmeier, line 1816: This could get a more expressive name, since this doesn't get you a split back, but takes one. Mabye 
```suggestion
    _file_id = {'10fold': '', 'train': 'DevTrain', 'test': 'DevTest'}
```

pmeier (commented):
Hey @ABD-01! This really looks good. I have some very minor comments inline as well as one question. Otherwise this is good to go!

torchvision/datasets/lfw.py
ABD-01: No. The downloaded files are completely fine but that just if incase someone alters/deletes/renames the images

torchvision/datasets/lfw.py
ABD-01: Yeah, same. There's almost zero chances that dataset will be corrupted, but again these are just incase someone alters the structure

torchvision/datasets/lfw.py
pmeier: That would make matters worse: now the dataset would silently ignore some files and thus changing the dataset. Results that were trained on that might not be reproducible. So either we throw an error in case the file is not there, or we simply don't perform this check. In case the file is gone, the error will be expressive.

ABD-01: Oh yes, you are right.
So I'll remove the check.
Or an alternative: 
create a list `self.missing = []`
```py
if os.path.exists(img):
    data.append(img)
    targets.append(self.class_to_idx[identity])
else:
   self.missing.append(img)
```
and in extra_repr
```py
def extra_repr(self) -> str:
    rpr =  super().extra_repr() + "\nClasses (identities): {}".format(len(self.class_to_idx))
    if len(self.missing) > 0:
        rpr + f"\nImages missing: {len(self.missing)}"
    return rpr
```
Let me know your thoughts, I'll update accordingly.

pmeier: I appreciate the thoroughness, but that again would hide the information in some sense. I'm guessing most users won't do `str(dataset)` and examine the result before they start training. Since we don't have any such checks in the other datasets, we can simply remove it.

torchvision/datasets/lfw.py
pmeier: Shouldn't we use `"original"` as the default? To me the other ones sound like special cases and thus are not good candidates for a default value.

torchvision/datasets/lfw.py
```diff
@@ -0,0 +1,260 @@
...
+        download (bool, optional): If true, downloads the dataset from the internet and
+            puts it in root directory. If dataset is already downloaded, it is not
+            downloaded again.
+
+    """
+
+    def __init__(
+        self,
+        root: str,
+        split: str = "10fold",
```
pmeier, line 117: Same here. Shouldn't `"train"` be the default?

ABD-01: * I've seen `funneled` used more than `original` (which is mostly used when paper uses their own face alignment framework). [And also [sklearn's lfw](https://scikit-learn.org/stable/modules/generated/sklearn.datasets.fetch_lfw_people.html) uses funneled as default]
* And 10fold is kept default as the [results on LFW](http://vis-www.cs.umass.edu/lfw/results.html) are provided on 10 fold view

torchvision/datasets/lfw.py
ABD-01: Okay Sir 👍

pmeier (approved):
This is great, thanks a lot @ABD-01!

fmassa (approved):
Thanks!

PR #4256, opened 2021-08-05
Use torch instead of scipy for random initialization of inception and googlenet weights

vmoens:
weights generated with `truncnorm` from scipy are conditioned by np global seed (or a RandomState). We want to avoid setting `np.random.seed` in our tests and also we don't want to pass a RandomState object in the model args. As such, we move to generating weights with `torch.nn.init.trunc_normal_` instead, which is conditioned by `torch.manual_seed`.


NicolasHug (commented):
Thanks for the PR @vmoens ,

I realize this is still WIP but I'm seeing lots of changed pkl files: according to https://app.circleci.com/pipelines/github/pytorch/vision/9719/workflows/9ad29d07-4475-418c-88e9-54e8f09b8c5c/jobs/721968 it seems that only one test was failing, so we probably don't need to update all of those.

Also it seems like we still have a typing error for `float(m.stddev)`

NicolasHug (approved):
Thanks @vmoens !

I tagged as BC breaking because the results might change slightly, but it's fine as we don't have guarantees w.r.t. randomness

datumbox:
@vmoens Thanks for the PR. The change looks useful. Though it's correctly marked as BC-breaking, hopefully the two methods don't produce significantly different results and the average user should be OK.

@NicolasHug It might be worth confirming that a model initialized under the new scheme does not diverge. A rudimental check in this case would be to run the model for 1-2 epochs and confirm that the loss decreases on the new branch. Is this something we ran already or plan to run on the future?

NicolasHug:
@vmoens would you mind trying @datumbox 's suggestion, just to make sure?

Issue #4270, opened 2021-08-12
annotation_path parameter in torchvision.datasets.UCF101 is not clear.


damtharvey:
## 📚 Documentation

Please describe what kind of files should be in annotation_path, and what the files should contain. It is not obvious.

cc @pmeier

fmassa:
Hi,

It's the folder that contain the following files:
```
classInd.txt  testlist01.txt  testlist02.txt  testlist03.txt  trainlist01.txt  trainlist02.txt  trainlist03.txt 
```
Which describes how the train / test split are selected.

Here is an example of what is in `trainlist01.txt`:
```
head annotation/trainlist01.txt
```
gives
```
ApplyEyeMakeup/v_ApplyEyeMakeup_g08_c01.avi 1
ApplyEyeMakeup/v_ApplyEyeMakeup_g08_c02.avi 1
ApplyEyeMakeup/v_ApplyEyeMakeup_g08_c03.avi 1
ApplyEyeMakeup/v_ApplyEyeMakeup_g08_c04.avi 1
ApplyEyeMakeup/v_ApplyEyeMakeup_g08_c05.avi 1
ApplyEyeMakeup/v_ApplyEyeMakeup_g09_c01.avi 1
ApplyEyeMakeup/v_ApplyEyeMakeup_g09_c02.avi 1
ApplyEyeMakeup/v_ApplyEyeMakeup_g09_c03.avi 1
ApplyEyeMakeup/v_ApplyEyeMakeup_g09_c04.avi 1
ApplyEyeMakeup/v_ApplyEyeMakeup_g09_c05.avi 1
```

I believe I've answered your question and as such I'm closing this issue. If you think the documentation can be improved in some way, please let us know or even better, send a pull request with proposed improvements.

damtharvey:
In case someone doesn't know you don't have to make this yourself, these files can be found in https://www.crcv.ucf.edu/wp-content/uploads/2019/03/UCF101TrainTestSplits-RecognitionTask.zip.

fmassa:
@bjuncek aren't those files available in the original download link for the datasets?

damtharvey:
The official web page is https://www.crcv.ucf.edu/data/UCF101.php, which contains the official download link https://www.crcv.ucf.edu/data/UCF101/UCF101.rar, which contains only directories of video files. There are various annotations available, including the action recognition annotations that are expected in annotation_path.

fmassa:
[...]

PR #4275, opened 2021-08-13
Update UCF101 docstring

bjuncek:
Addressing #4270

### DIFF
--- a/torchvision/datasets/ucf101.py
+++ b/torchvision/datasets/ucf101.py
@@ -14,7 +14,10 @@ class UCF101(VisionDataset):
     UCF101 is an action recognition video dataset.
     This dataset consider every video as a collection of video clips of fixed size, specified
     by ``frames_per_clip``, where the step in frames between each clip is given by
-    ``step_between_clips``.
+    ``step_between_clips``. The dataset itself can be downloaded from `here
+    <https://www.crcv.ucf.edu/data/UCF101/UCF101.rar>`; annotations that ``annotation_path``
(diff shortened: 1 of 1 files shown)


bjuncek:
Unitest failing seems unrelated.

torchvision/datasets/ucf101.py
```diff
@@ -14,7 +14,10 @@ class UCF101(VisionDataset):
     UCF101 is an action recognition video dataset.
     This dataset consider every video as a collection of video clips of fixed size, specified
     by ``frames_per_clip``, where the step in frames between each clip is given by
-    ``step_between_clips``.
+    ``step_between_clips``. The dataset itself can be downloaded from `here
+    <https://www.crcv.ucf.edu/data/UCF101/UCF101.rar>`; annotations that ``annotation_path``
+    should be pointing to can be downloaded from `here
```
NicolasHug, line 19: I think in general it's best practice not to directly link to a downloadable file, but instead explain how to download it. The link might change and break, but the instructions should be a bit more robust. For example, WDYT of saying something like:

> The annotation files contains the train/test splits information and can correspond to the "Train/Test Splits for Action Recognition on UCF101 data set" link on the home page of the dataset.

NicolasHug: Also, in the description for `annotation_path` I think we should  say something like

> see docstring above for download instructions of these files

NicolasHug (approved):
Thanks @bjuncek , I have some minor comments but LGTM anyway!

Issue #4278, opened 2021-08-16
CI failing due to fate repo being down


NicolasHug:
https://fate.ffmpeg.org/fate-suite is down and we rely on it in our tests: https://github.com/pytorch/vision/blob/master/test/test_videoapi.py#L34:L34

which are currently broken https://app.circleci.com/pipelines/github/pytorch/vision/9783/workflows/f63f033a-d999-4c67-a1a1-477c3805cf20/jobs/729094

Hopefully this will resolve on its own

cc @pmeier

PR #4280, opened 2021-08-16
Replace get_tmp_dir() with tmpdir fixture in pytest

ccongge:
Per title. 

- Replaced get_tmp_dir() in tests with tmpdir fixture
- Moved get_tmp_dir() last remaining caller to test/dataset_util.py (as it is not ported to pytest yet)

### DIFF
--- a/test/test_datasets_samplers.py
+++ b/test/test_datasets_samplers.py
@@ -13,7 +13,7 @@ from torchvision.datasets.samplers import (
 from torchvision.datasets.video_utils import VideoClips, unfold
 from torchvision import get_video_backend
 
-from common_utils import assert_equal
+from common_utils import get_tmp_dir, assert_equal
 
 
 @contextlib.contextmanager
--- a/test/test_datasets_samplers.py
+++ b/test/test_datasets_samplers.py
@@ -13,34 +13,33 @@ from torchvision.datasets.samplers import (
 from torchvision.datasets.video_utils import VideoClips, unfold
 from torchvision import get_video_backend
 
-from common_utils import get_tmp_dir, assert_equal
+from common_utils import assert_equal
 
 
 @contextlib.contextmanager
-def get_list_of_videos(num_videos=5, sizes=None, fps=None):
-    with get_tmp_dir() as tmp_dir:
-        names = []
-        for i in range(num_videos):
-            if sizes is None:
-                size = 5 * (i + 1)
-            else:
-                size = sizes[i]
-            if fps is None:
-                f = 5
-            else:
-                f = fps[i]
-            data = torch.randint(0, 256, (size, 300, 400, 3), dtype=torch.uint8)
-            name = os.path.join(tmp_dir, "{}.mp4".format(i))
-            names.append(name)
-            io.write_video(name, data, fps=f)
+def get_list_of_videos(tmpdir, num_videos=5, sizes=None, fps=None):
+    names = []
+    for i in range(num_videos):
+        if sizes is None:
+            size = 5 * (i + 1)
+        else:
+            size = sizes[i]
+        if fps is None:
+            f = 5
+        else:
+            f = fps[i]
+        data = torch.randint(0, 256, (size, 300, 400, 3), dtype=torch.uint8)
+        name = os.path.join(tmpdir, "{}.mp4".format(i))
+        names.append(name)
+        io.write_video(name, data, fps=f)
 
-        yield names
+    yield names
 
 
 @pytest.mark.skipif(not io.video._av_available(), reason="this test requires av")
 class TestDatasetsSamplers:
-    def test_random_clip_sampler(self):
-        with get_list_of_videos(num_videos=3, sizes=[25, 25, 25]) as video_list:
+    def test_random_clip_sampler(self, tmpdir):
+        with get_list_of_videos(tmpdir, num_videos=3, sizes=[25, 25, 25]) as video_list:
             video_clips = VideoClips(video_list, 5, 5)
             sampler = RandomClipSampler(video_clips, 3)
             assert len(sampler) == 3 * 3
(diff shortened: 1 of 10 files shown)


NicolasHug:
Thanks a lot for the PR @ccongge ! This looks great so far, LMK when it's ready for a proper review.
We can safely ignore the fate-related CI issues: https://github.com/pytorch/vision/issues/4278

NicolasHug:
>  There are still references for get_tmp_dir(), but they are not used in pytest tests but as util functions.

We can safely ignore the `get_tmp_dir()` that's in `torchvision/datasets/imagenet.py`. However it looks like there are still some occurrences in the test folder in this PR branch:

```
(pt) ➜  vision git:(replace_get_tmp_dir) ✗ git grep get_tmp_dir test
test/common_utils.py:def get_tmp_dir(src=None, **kwargs):
test/datasets_utils.py:from common_utils import get_tmp_dir, disable_console_output
test/datasets_utils.py:        with get_tmp_dir() as tmpdir:
test/test_datasets_samplers.py:from common_utils import get_tmp_dir, assert_equal
test/test_datasets_samplers.py:    with get_tmp_dir() as tmp_dir:
test/test_datasets_video_utils.py:from common_utils import get_tmp_dir, assert_equal
test/test_datasets_video_utils.py:    with get_tmp_dir() as tmp_dir:
test/test_transforms_tensor.py:    get_tmp_dir,
test/test_transforms_tensor.py:    with get_tmp_dir() as tmp_dir:
```

test/test_datasets_samplers.py
```diff
@@ -13,34 +13,33 @@
...
-        yield names
+    yield names
 
 
 @pytest.mark.skipif(not io.video._av_available(), reason="this test requires av")
 class TestDatasetsSamplers:
-    def test_random_clip_sampler(self):
-        with get_list_of_videos(num_videos=3, sizes=[25, 25, 25]) as video_list:
+    def test_random_clip_sampler(self, tmpdir):
+        with get_list_of_videos(tmpdir, num_videos=3, sizes=[25, 25, 25]) as video_list:
```
NicolasHug, line 42: I think that `get_list_of_videos` was written as a context manager because it was relying on `get_tmp_dir()` which was itself a context manager.

But now that we're getting rid of `get_tmp_dir()`, I believe we should be able to convert `get_list_of_videos()` into a simple function: we can remove the `@contextlib.contextmanager` decorator and change `yield` into a `return`. And here, we could just do 

```py
video_list = get_list_of_videos(tmpdir, num_videos=3, sizes=[25, 25, 25])
```

and indent everything to the left

test/test_datasets_video_utils.py
```diff
@@ -6,28 +6,27 @@
...
-                f = 5
-            else:
-                f = fps[i]
-            data = torch.randint(0, 256, (size, 300, 400, 3), dtype=torch.uint8)
-            name = os.path.join(tmp_dir, "{}.mp4".format(i))
-            names.append(name)
-            io.write_video(name, data, fps=f)
-
-        yield names
+def get_list_of_videos(tmpdir, num_videos=5, sizes=None, fps=None):
```
NicolasHug, line 13: Oh, I'm discovering now that this function is duplicated. Let's get rid of both context manager (in this file and the other one above) and instead put the function version that I suggested in `common_utils.py`.

test/test_transforms_tensor.py
```diff
@@ -109,61 +107,63 @@ def _test_op(func, method, device, fn_kwargs=None, meth_kwargs=None, test_exact_
...
-def test_random(func, method, device, fn_kwargs, match_kwargs):
-    _test_op(func, method, device, fn_kwargs, fn_kwargs, **match_kwargs)
-
-
 @pytest.mark.parametrize('device', cpu_and_gpu())
 class TestColorJitter:
 
     @pytest.mark.parametrize('brightness', [0.1, 0.5, 1.0, 1.34, (0.3, 0.7), [0.4, 0.5]])
-    def test_color_jitter_brightness(self, brightness, device):
+    @pytest.fixture()
```
NicolasHug, line 114: We should remove all  the newly added `@pytest.fixture()` decorators, as this transforms the test into a fixture, and  the test doesn't get run

test/test_transforms_tensor.py
```diff
@@ -85,13 +84,12 @@ def _test_class_op(method, device, meth_kwargs=None, test_exact_match=True, **ma
     batch_tensors = _create_data_batch(height=23, width=34, channels=3, num_samples=4, device=device)
     _test_transform_vs_scripted_on_batch(f, scripted_fn, batch_tensors)
 
-    with get_tmp_dir() as tmp_dir:
-        scripted_fn.save(os.path.join(tmp_dir, f"t_{method.__name__}.pt"))
+    scripted_fn.save(os.path.join(tmp_dir, f"t_{method.__name__}.pt"))
```
NicolasHug, line 87: hm, the structure of the tests in this file is slightly awkward (for various historical reasons), and it makes it a bit cumbersome to pass the `tmpdir` fixture from the actual `test_...()` function all the way down to here.

So I would instead suggest to actually just rely on our custom `get_tmp_dir()` function for this and revert those bits. I'll try to find a cleaner solution in the future, potentially refactoring further those tests. This means that we should put back `get_tmp_dir()` in the `common_utils.py` file (sorry for the extra work!)

test/test_transforms_tensor.py
```diff
@@ -309,11 +311,10 @@ def test_x_crop(fn, method, out_length, size, device):
 
 
 @pytest.mark.parametrize('method', ["FiveCrop", "TenCrop"])
-def test_x_crop_save(method):
+def test_x_crop_save(method, tmpdir):
```
NicolasHug, line 314: Just to clarify: for these functions that directly use the `tmpdir` fixture and do not indirectly call `_test_class_op` of `_test_op`, it's nice to have `tmpdir` and not to rely on `get_tmp_dir`, so no need to revert the changes for these :)

NicolasHug (commented):
Thanks a lot @ccongge , this look great! I made a few comments below but this is in good shape already

test/test_datasets_samplers.py
ccongge: Thanks, edited. :)

test/test_datasets_video_utils.py
ccongge: Thanks, edited. :)

test/test_transforms_tensor.py
ccongge: Thanks, edited. :)

test/test_transforms_tensor.py
ccongge: Yes. That is also what I was wondering. I moved get_tmp_dir() back to common_utils.py but still replaced other get_tmp_dir() with tmpdir fixsure except the one in _test_class_op().

test/test_transforms_tensor.py
ccongge: Yes, per above comment it is edited.

NicolasHug (approved):
Thanks a lot @ccongge ! LGTM

PR #4283, opened 2021-08-17
Fix unnecessary backbone weights download on lraspp_mobilenet_v3_large

### LINKED ISSUE #4274: a little problem when using some pretrained models
## 🐛 Bug

A little problem.

I used some pretrained models to do object detection. 

However, when i used models whose name include others models' name, such as 'fasterrcnn_mobilenet_v3_large_320_fpn', (its name includes the name of model 'mobilenet_v3_large'),  it will download the weight file of the short name models. 

For example, when i used model 'fasterrcnn_mobilenet_v3_large_320_fpn', whenever the pretrained attribute is True or not, the weight file of model 'mobilenet_v3_large' will be downloaded.  

This problem also happen in the models such as 'maskrcnn_resnet50_fpn' and many other models.

## To Reproduce

it's easy to reproduce. For example:
```bash
from torchvision.models.detection import *
model = fasterrcnn_mobilenet_v3_large_320_fpn()
```
excute the code above, the weight file of model "mobilenet_v3_large" will be  downloaded.
you can change the model name to other this kind of model names.

cc @fmassa @vfdev-5 @pmeier

datumbox:
Fixes #4274

### DIFF
--- a/torchvision/models/segmentation/segmentation.py
+++ b/torchvision/models/segmentation/segmentation.py
@@ -186,6 +186,8 @@ def lraspp_mobilenet_v3_large(pretrained=False, progress=True, num_classes=21, *
         raise NotImplementedError('This model does not use auxiliary loss')
 
     backbone_name = 'mobilenet_v3_large'
+    if pretrained:
+        kwargs["pretrained_backbone"] = False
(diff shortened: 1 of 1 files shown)


fmassa (approved):
Thanks!

PR #4284, opened 2021-08-17
Skip test if connection issues on fate

datumbox:
The Fate server of FFmpeg has been giving us headaches the last 2 weeks. Instead of disabling temporarily the test, I'm taking the approach to skip it if a connection issue occurs.

### DIFF
--- a/test/test_videoapi.py
+++ b/test/test_videoapi.py
@@ -2,6 +2,7 @@ import collections
 import os
 import pytest
 from pytest import approx
+import urllib
 
 import torch
 import torchvision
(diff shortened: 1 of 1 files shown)


test/test_videoapi.py
```diff
@@ -177,7 +178,10 @@ def test_accurateseek_middle(self):
                 assert (lb <= frame["pts"]) and (ub >= frame["pts"])
 
     def test_fate_suite(self):
-        video_path = fate("sub/MovText_capability_tester.mp4", VIDEO_DIR)
+        try:
+            video_path = fate("sub/MovText_capability_tester.mp4", VIDEO_DIR)
+        except (urllib.error.URLError, ConnectionError) as error:
```
NicolasHug, line 183: Could you please add a TODO or a FIXME so we remember to remove this eventually? Otherwise LGTM, thanks

NicolasHug (approved):

PR #4285, opened 2021-08-17
Make torch version check numeric

jithunnair-amd:
String comparison messes up in cases such as `torch.__version__ = 1.10.blah`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -147,7 +147,9 @@ def get_extensions():
     )
 
     is_rocm_pytorch = False
-    if torch.__version__ >= '1.5':
+    TORCH_MAJOR = int(torch.__version__.split('.')[0])
+    TORCH_MINOR = int(torch.__version__.split('.')[1])
+    if TORCH_MAJOR > 1 or (TORCH_MAJOR == 1 and TORCH_MINOR >= 5):
         from torch.utils.cpp_extension import ROCM_HOME
         is_rocm_pytorch = True if ((torch.version.hip is not None) and (ROCM_HOME is not None)) else False
 


datumbox (approved):
LGTM, thanks! Will merge once the CI completes.

setup.py
```diff
@@ -147,7 +147,9 @@ def get_extensions():
     )
 
     is_rocm_pytorch = False
-    if torch.__version__ >= '1.5':
+    TORCH_MAJOR = int(torch.__version__.split('.')[0])
+    TORCH_MINOR = int(torch.__version__.split('.')[1])
+    if TORCH_MAJOR > 1 or (TORCH_MAJOR == 1 and TORCH_MINOR >= 5):
```
fmassa, line 152: Should we be [using the newly introduced `TorchVersion` instead](https://github.com/pytorch/pytorch/blob/1022443168b5fad55bbd03d087abf574c9d2e9df/torch/torch_version.py#L13)?

fmassa: Created an issue to track this down https://github.com/pytorch/vision/issues/4296

ABD-01: If we want to use the new [`torch.__version__` as a `TorchVersion` object](https://github.com/pytorch/pytorch/blob/1022443168b5fad55bbd03d087abf574c9d2e9df/torch/torch_version.py#L13) , then just reverting this pr will do the job.
Or can I create a new pr for just one line?

NicolasHug: >  Or can I create a new pr for just one line?

Yup :)

ABD-01: Done.
Thanks Sir

PR #4290, opened 2021-08-18
`masks_to_bounding_boxes` op

### LINKED ISSUE #3960: Ops to convert `masks` to `boxes`
## 🚀 Feature
A simple `torchvision.ops` to convert Segmentation masks to bounding boxes.

## Motivation

This has a few use-cases.

1. This makes it easier to use semantic segmentation datasets for object detection.
The pipeline can be easier. Also the bounding boxes are represented as `xyxy `in `torchvision.ops` as a convention.
So probably convert masks to `xyxy` format.

2. The other use case is to make it easier in comparing performance of segmentation model vs detection model.
Let's Say that the detection model performs well for segmentation dataset. Then it would be better to go ahead with detection models as it is faster in real-time use-cases than to train a segmentation model.

New Pipeline

```
from torchvision.ops import masks_to_boxes, box_convert

class SegmentationToDetectionDataset(Dataset):
    def __getitem__(self, idx):
          boxes_xyxy = masks_to_boxes(segmentation_masks)

         # Now for any change of boxes to COCO Format.
          boxes_xywh = box_convert(boxes_xyxy, in_fmt="xyxy", out_fmt="xywh")
          return boxes_xywh
```

## Pitch

Port the [masks_to_boxes](https://github.com/ashkamath/mdetr/blob/fdee8c50d7bcf2ad09cc0d6b783a8333720e4048/util/box_ops.py#L63) function from mDeTR.

[masks_to_boxes](https://github.com/facebookresearch/detr/blob/a54b77800eb8e64e3ad0d8237789fcbf2f8350c5/util/box_ops.py#L64) was also used in DeTR.

## Alternatives

The above function assumes masks of shape `(N, H, W)` -> `num_masks, Height, Width`. A floating tensor.
IIRC, we used a boolean tensor in `draw_segmentation_masks` (After Nicolas refactored). So perhaps we should be using boolean tensor? Though I see no particular use case of this util being only valid for instance segmentation.

## Additional context

I can port this, we perhaps need a few tests to ensure it works fine.
Especially test for float16 overflow.

cc @datumbox @NicolasHug

0x00b1:
This (draft) pull request resolves #3960. I created a draft to kickoff new contributor on-boarding (e.g. CLA).

I'm working on a gallery example now. I'll also add test against different `dtypes`.


oke-aditya:
Sorry for an early poke at this PR.

I think it would be nice to place the test in [test_ops.py](https://github.com/pytorch/vision/blob/master/test/test_ops.py)

I'm not sure of how it should be tested. My initially were we could manually create boolean masks and test them like the ones done in [utils.draw_segmentation_masks test](https://github.com/pytorch/vision/blob/master/test/test_utils.py#L175)?

Also the code can be kept directly in [boxes](https://github.com/pytorch/vision/blob/master/torchvision/ops/boxes.py) ? No strong opinion here though.

0x00b1:
Hi, @oke-aditya! I appreciate any and all feedback! 😄 My code organization was my preference but I am more than happy to adopt whatever convention you or any other maintainers prefer.

I will parametrize the `dtype` in the fixtures but other than that I think the unit test does its job. If you're curious, I used the `random_shapes` function from `skimage.draw` to create the fixtures. I was the original author of that function and its purpose was to solve the exact issue of writing unit tests for object localization methods. Ideally, in the future, if `torchvision` starts adding more of these types of operators it would be nice to port over that function and similar data generators to write these types of tests. It would also allow for some basic fuzzing.

oke-aditya:
> if torchvision starts adding more of these types of operators it would be nice to port over that function and similar data generators from scikit-image to simplify writing these types of tests

Not sure if they fall into same category, but there too we hardcoded some of the boxes and masks for images to test them.
Actually we have bunch of tests, from [here](https://github.com/pytorch/vision/blob/master/test/test_utils.py#L93) (these are operators used for plotting boxes, masks) Also a many operators like `box_iou`, `box_area`, where the boxes are predefined (but these are mostly mathematical so probably ok). You can have a look at these and I would be glad to hear your thoughts!

My initial thought was that this function too would be tested in similar way instead of drawing random shapes / boxes on an image.

P.S. I'm just a small contributor to the library (also a novice developer) so please don't mind.

test/test_masks_to_bounding_boxes.py
```diff
@@ -0,0 +1,43 @@
...
+import pytest
+import torch
+
+import torchvision.ops
+
+ASSETS_DIRECTORY = os.path.join(os.path.dirname(os.path.abspath(__file__)), "assets")
+
+
+@pytest.fixture
+def labeled_image() -> torch.Tensor:
```
NicolasHug, line 14: Just FYI, we don't really need / use type annotations in the tests

test/test_masks_to_bounding_boxes.py
```diff
@@ -0,0 +1,43 @@
...
+    with PIL.Image.open(os.path.join(ASSETS_DIRECTORY, "labeled_image.png")) as image:
+        return torch.tensor(numpy.array(image, numpy.int))
+
+
+@pytest.fixture
+def masks() -> torch.Tensor:
```
NicolasHug, line 20: unless we're planning on using t his fixture in other tests, I would suggest to not make it a fixture. It could just be within the `test_masks_to_bounding_boxes`

test/test_masks_to_bounding_boxes.py
```diff
@@ -0,0 +1,43 @@
...
+            image.seek(index)
+
+            frames[index] = numpy.array(image)
+
+        return torch.tensor(frames)
+
+
+def test_masks_to_bounding_boxes(masks):
+    expected = torch.tensor(
+       [[ 127.,    2.,  165.,   40. ],  # noqa: E121, E201, E202, E241
```
NicolasHug, line 34: we should avoid the redundant `.` at the end, i.e. `127.` should just be `127`. We can set the `dtype` argument of the tensor if needed.

Also, let's avoid all these flake8 issues and just format the array as expected.

torchvision/ops/_masks_to_bounding_boxes.py
```diff
@@ -0,0 +1,26 @@
+import torch
+
+
+def masks_to_bounding_boxes(masks: torch.Tensor) -> torch.Tensor:
```
NicolasHug, line 4: Since all other utils use the term `box` instead of `bounding_box`, let's call this `masks_to_boxes` instead
(although I would personally prefer the way you did it :)  )

torchvision/ops/_masks_to_bounding_boxes.py
```diff
@@ -0,0 +1,26 @@
...
+    """Compute the bounding boxes around the provided masks
+    The masks should be in format [N, H, W] where N is the number of masks, (H, W) are the spatial dimensions.
+    Returns a [N, 4] tensors, with the boxes in xyxy format
```
NicolasHug, line 7: Would you mind writing a full docstring with a Parameters and Return section, e.g. like in https://github.com/pytorch/vision/blob/master/torchvision/ops/boxes.py


I would also suggest to move this function into the `boxes.py` file instead of creating a new file

torchvision/ops/_masks_to_bounding_boxes.py
```diff
@@ -0,0 +1,26 @@
...
+    """
+    if masks.numel() == 0:
+        return torch.zeros((0, 4), device=masks.device)
+
+    h, w = masks.shape[-2:]
```
NicolasHug, line 12: Do you think it would be useful to also allow for tensors of shape (H, W), i.e. let the function accept a single mask instead of a batch?

test/test_masks_to_bounding_boxes.py
```diff
@@ -0,0 +1,43 @@
...
+
+@pytest.fixture
+def labeled_image() -> torch.Tensor:
+    with PIL.Image.open(os.path.join(ASSETS_DIRECTORY, "labeled_image.png")) as image:
+        return torch.tensor(numpy.array(image, numpy.int))
+
+
+@pytest.fixture
+def masks() -> torch.Tensor:
+    with PIL.Image.open(os.path.join(ASSETS_DIRECTORY, "masks.tiff")) as image:
```
NicolasHug, line 21: Do you think it would be possible to write a test without the need for new images and hard-coded coordinates?

Ideally, we could generate random masks and have a super simple version of `masks_to_boxes` which we could use as the reference implementation?

NicolasHug (commented):
Thanks a lot for the PR @0x00b1 and @oke-aditya for the review. I made a few comments and will look at the rest once this isn't draft anymore. Thanks for the initiative of writing a gallery example!!

torchvision/ops/_masks_to_bounding_boxes.py
RylanSchaeffer: It would be useful to me if the tensors could have shape (H, W, D) or (batch, H, W, D) i.e. 3D images.

0x00b1:
@datumbox Yep! I was on vacation (it was lovely). I'll finish this today or tomorrow. Thanks, @NicolasHug for the comments!

RylanSchaeffer:
Can you generalize this to 3D images?

datumbox:
@RylanSchaeffer it's definitely worth discussing it on a new issue. I would prefer if we did this on a separate PR to avoid blocking it for longer.

RylanSchaeffer:
@datumbox I have two opinions. On one hand, I agree that not blocking this PR is good. On the other, a half solution means a fix to the complete problem will probably be delayed.

torchvision/ops/_masks_to_bounding_boxes.py
oke-aditya: I think it should be Tensor of [N, H, W] since all the other ops accept tensor of shape [N, 4].
Also (H, W) is a special case of [1, H, W] ? I don't think so we handle such cases where N = 1 differently for other ops.

NicolasHug:
> a half solution means a fix to the complete problem will probably be delayed

This PR isn't half a solution, it's a complete solution to a complete problem: 2D images.

3D images are a different problem which we'll be happy to tackle at a future time once this PR is merged, as @datumbox suggested

RylanSchaeffer:
> This PR isn't half a solution, it's a complete solution to a complete problem: 2D images.

That's a strange way to think about things. Imagine someone submitted a cross entropy loss implementation for a single axis. By your metric, it's a complete solution to a complete problem i.e. 1-dimensional classification. But the **real** problem is more general than this. For us, the real problem is: given a N-dimensional segmentation mask, how to convert the mask to N-dimensional bounding boxes?

Look at the cross entropy loss implementation: it works for arbitrary dimensions, not just one.

image

NicolasHug:
@RylanSchaeffer , just because we can generalize a problem doesn't mean that one problem is less "real" or "complete" than the other. 2D masks are a normal use-case that a lot of people have and solving this will be valuable on its own.

When it comes to software development, a merged PR that solves one problem is worth more than an unmerged PR that solves 2 problems.

Again, we'll be happy to consider an extension if the use-case is compelling.

RylanSchaeffer:
Ok issue opened! https://github.com/pytorch/vision/issues/4339

torchvision/ops/_masks_to_bounding_boxes.py
0x00b1: @NicolasHug IMO we should have a complimentary op for transforming labeled images (i.e. connected components like the image returned by functions like `scipy.ndimage.label` or `skimage.measure.label`) like `labeled_image_to_boxes`. What do you think?

0x00b1: @RylanSchaeffer Good idea! I'll adapt it to 3D (or even n-D). But I suppose that can come in a future PR.

0x00b1:
@RylanSchaeffer I agree that this should come in a future PR. But don't worry, 3D or n-D images are something I care about too so I'm more than happy to do that work.

torchvision/ops/boxes.py
```diff
@@ -297,3 +297,36 @@ def generalized_box_iou(boxes1: Tensor, boxes2: Tensor) -> Tensor:
     areai = whi[:, :, 0] * whi[:, :, 1]
 
     return iou - (areai - union) / areai
+
+
+def masks_to_boxes(masks: torch.Tensor) -> torch.Tensor:
+    """
+    Compute the bounding boxes around the provided masks
+
+    Returns a [N, 4] tensors, with the boxes in (X, Y, X, Y) format.
```
oke-aditya, line 306: nit: Probably instead of (X, Y, X, Y) you can write
```
 Both sets of boxes are expected to be in ``(x1, y1, x2, y2)`` format with
``0 <= x1 < x2`` and ``0 <= y1 < y2
```
as this is written in other docstrings in ops.

test/test_masks_to_bounding_boxes.py
0x00b1: @NicolasHug a friendly bump

0x00b1:
@NicolasHug This is nearly ready. I just need to add the code to the gallery example. Unfortunately I cannot get Sphinx to build correctly. I get the following error which I cannot reproduce in the REPL. Any ideas?

```sh
/Users/allengoodman/Documents/com/github/0x00b1/vision/gallery/plot_scripted_tensor_transforms.py failed to execute correctly: Traceback (most recent call last):
  File "/Users/allengoodman/Documents/com/github/0x00b1/vision/gallery/plot_scripted_tensor_transforms.py", line 52, in <module>
    dog1 = read_image(str(Path('assets') / 'dog1.jpg'))
  File "/Users/allengoodman/Documents/com/github/0x00b1/vision/torchvision/io/image.py", line 222, in read_image
    data = read_file(path)
  File "/Users/allengoodman/Documents/com/github/0x00b1/vision/torchvision/io/image.py", line 42, in read_file
    data = torch.ops.image.read_file(path)
  File "/Users/allengoodman/Documents/com/github/0x00b1/vision/venv/lib/python3.7/site-packages/torch/_ops.py", line 60, in __getattr__
    op = torch._C._jit_get_operation(qualified_op_name)
RuntimeError: No such operator image::read_file
```

datumbox:
@0x00b1 It seems you have not compiled the C++ ops of TorchVision. You basically need to run from the root for the project:
`python setup.py develop`

See this guide for [details](https://github.com/pytorch/vision/blob/main/CONTRIBUTING.md#development-installation).

gallery/plot_repurposing_annotations.py
```diff
@@ -69,5 +38,31 @@ def plot(imgs, with_orig=True, row_title=None, **imshow_kwargs):
...
 # For example, the masks to bounding_boxes operation can be used to transform masks into bounding boxes that can be
 # used in methods like Faster RCNN and YOLO.
-padded_imgs = [T.Pad(padding=padding)(orig_img) for padding in (3, 10, 30, 50)]
-plot(padded_imgs)
+
+image, labels = skimage.draw.random_shapes((512, 256), 4, min_size=32, multichannel=False)
+
+labeled_image = skimage.measure.label(image, background=255)
+
+labeled_image = skimage.img_as_ubyte(labeled_image)
```
datumbox, line 46: Could we avoid introducing a `skimage` dependency on the docs? It seems we could do something like this using pytorch.

0x00b1: I'm using it for `random_shapes`. I suppose I could use the text fixture images.

gallery/plot_repurposing_annotations.py
```diff
@@ -0,0 +1,68 @@
...
+segmentation methods into bounding boxes used by object detection methods).
+"""
+
+import matplotlib.patches
+import matplotlib.pyplot
+import numpy
+import skimage.draw
+import skimage.measure
+import torch
+import torchvision.ops.boxes
```
oke-aditya, line 17: In practice shouldn't import the file. (It's correct theoretically, but we should import the only function needed)

We could just do
```suggestion
from torchvision.ops import masks_to_boxes
```

gallery/plot_repurposing_annotations.py
```diff
@@ -0,0 +1,68 @@
...
+# For example, the masks to bounding_boxes operation can be used to transform masks into bounding boxes that can be
+# used in methods like Faster RCNN and YOLO.
+
+image, labels = skimage.draw.random_shapes((512, 256), 4, min_size=32, multichannel=False)
+
+labeled_image = skimage.measure.label(image, background=255)
+
+labeled_image = skimage.img_as_ubyte(labeled_image)
+
+masks = numpy.zeros((len(labels), *labeled_image.shape), numpy.ubyte)
```
oke-aditya, line 48: Just a doubt.
Why not `masks = torch.zeros((len(labels), *labeled_image.shape), numpy.ubyte)`

We wouldn't nee need to convert to tensors later.

gallery/plot_repurposing_annotations.py
```diff
@@ -0,0 +1,68 @@
...
+
+for index in numpy.unique(labeled_image):
+    masks[index - 1] = numpy.where(labeled_image == index, index, 0)
```
oke-aditya, line 51: Can we try `torch.where` ?

gallery/plot_repurposing_annotations.py
```diff
@@ -0,0 +1,68 @@
...
+
+figure = matplotlib.pyplot.figure()
+
+a = figure.add_subplot(121)
+b = figure.add_subplot(122)
+
+a.imshow(labeled_image)
+b.imshow(labeled_image)
+
+for bounding_box in bounding_boxes.tolist():
```
oke-aditya, line 63: Hmm you could use `torchvision.utils.draw_bounding_boxes` probably?
You wouldn't need to convert `bounding_boxes.tolist()`. It can accept tensors. Also you won't need `matplotlib.patches`, etc.

test/test_masks_to_boxes.py
```diff
@@ -0,0 +1,34 @@
+import os.path
+
+import PIL.Image
+import numpy
+import torch
+
+import torchvision.ops
+import torchvision.ops.boxes
```
oke-aditya, line 8: I think you only need 1 import.
```suggestion
from torchvision.ops import masks_to_boxes.
```

oke-aditya (commented):
Some small thoughts from my side

gallery/plot_repurposing_annotations.py
0x00b1: I am an idiot.

gallery/plot_repurposing_annotations.py
oke-aditya: It directly accepts bounding box of format which `masks_to_boxes` have converted to!

gallery/plot_repurposing_annotations.py
0x00b1: I am an idiot x2 (but I believe it should be `masks = torch.zeros((len(labels), *labeled_image.shape), dtype=torch.int)`).

gallery/plot_repurposing_annotations.py
oke-aditya: Well no sir, it happens with everyone. If you are hurt by my review. I'm bit sorry too.

gallery/plot_repurposing_annotations.py
oke-aditya: (It happens with everyone. 
I remember those days when I couldn't compare 3 variables properly :sweat_smile: Happens with all)

gallery/plot_repurposing_annotations.py
```diff
@@ -0,0 +1,75 @@
...
+import os.path
+
+import PIL.Image
+import matplotlib.patches
+import matplotlib.pyplot
+import numpy
+import torch
+import torchvision.ops.boxes
+
+ASSETS_DIRECTORY = "../test/assets"
```
oke-aditya, line 19: Actually there is gallery assets folder.
We shouldn't use the tests folder I think

https://github.com/pytorch/vision/tree/main/gallery/assets

0x00b1: I didn't want to duplicate assets.

oke-aditya: Correct. But it would be nice to seperate tests from gallery. Let's hear from others too!

torchvision/ops/boxes.py
```diff
@@ -297,3 +297,35 @@ def generalized_box_iou(boxes1: Tensor, boxes2: Tensor) -> Tensor:
...
+
+    Returns:
+        Tensor[N, 4]: bounding boxes
+    """
+    if masks.numel() == 0:
+        return torch.zeros((0, 4))
+
+    n = masks.shape[0]
+
+    bounding_boxes = torch.zeros((n, 4), dtype=torch.int)
```
datumbox, line 321: Add this to ensure the new Tensor is created on the right device:
```suggestion
    bounding_boxes = torch.zeros((n, 4), dtype=torch.int, device=masks.device)
```

gallery/plot_repurposing_annotations.py
```diff
@@ -0,0 +1,75 @@
...
+# A nice property of masks is that they can be easily repurposed to be used in methods to solve a variety of object
+# localization tasks.
+#
+# Masks to bounding boxes
+# ~~~~~~~~~~~~~~~~~~~~~~~
+# For example, the masks to bounding_boxes operation can be used to transform masks into bounding boxes that can be
+# used in methods like Faster RCNN and YOLO.
+
+with PIL.Image.open(os.path.join(ASSETS_DIRECTORY, "masks.tiff")) as image:
+    frames = numpy.zeros((image.n_frames, image.height, image.width), int)
```
datumbox, line 46: My understanding is that all of these numpy operations can be written in pytorch. I think it's important to use pytorch in our own examples.

test/test_masks_to_boxes.py
```diff
@@ -0,0 +1,34 @@
...
+import torch
+
+from torchvision.ops import masks_to_boxes
+
+ASSETS_DIRECTORY = os.path.join(os.path.dirname(os.path.abspath(__file__)), "assets")
+
+
+def test_masks_to_boxes():
+    with PIL.Image.open(os.path.join(ASSETS_DIRECTORY, "masks.tiff")) as image:
+        frames = numpy.zeros((image.n_frames, image.height, image.width), int)
```
datumbox, line 14: Same here

gallery/plot_repurposing_annotations.py
```diff
@@ -0,0 +1,75 @@
...
+segmentation methods into bounding boxes used by object detection methods).
+"""
+import os.path
+
+import PIL.Image
+import matplotlib.patches
+import matplotlib.pyplot
+import numpy
+import torch
+from torchvision.ops.boxes import masks_to_boxes
```
oke-aditya, line 17:
```suggestion
from torchvision.ops import masks_to_boxes
```

gallery/plot_repurposing_annotations.py
```diff
@@ -0,0 +1,75 @@
...
+
+ASSETS_DIRECTORY = "../test/assets"
+
+matplotlib.pyplot.rcParams["savefig.bbox"] = "tight"
+
+####################################
+# Masks
+# --------------------------------------
```
oke-aditya, line 25:
```suggestion
# -----
```

gallery/plot_repurposing_annotations.py
```diff
@@ -0,0 +1,75 @@
...
+# one object. For example, if your input image has the dimensions 224 x 224 and has four annotated objects the shape
+# of your masks annotation has the following shape:
+#
+#       (4, 224, 224).
+#
+# A nice property of masks is that they can be easily repurposed to be used in methods to solve a variety of object
+# localization tasks.
+#
+# Masks to bounding boxes
+# ~~~~~~~~~~~~~~~~~~~~~~~
```
oke-aditya, line 41:
```suggestion
# ----------------------------------------
```

oke-aditya (commented):
Hi @0x00b1 Allen!

Just the build docs CI needs to go green. Then I think it would be good to go!

https://app.circleci.com/pipelines/github/pytorch/vision/10607/workflows/bc1a73b1-8a36-40a7-80ce-0ed6f6dd5fe9/jobs/806889/steps

gallery/plot_repurposing_annotations.py
```diff
@@ -0,0 +1,75 @@
+"""
+=======================
+Repurposing annotations
+=======================
+
+The following example illustrates the operations available in :ref:`the torchvision.ops module <ops>` for repurposing
```
oke-aditya, line 6: After some debugging I found out the reason for `build_docs` CI failure. The problem is `torchvision.ops` does not have a nice index on right side (basically a html link to #ops like transforms has). This causes CI failure.

We need to remove the ref, and it will work fine.
I tried running it locally. I could build the gallery example. It looks nice.

```suggestion
The following example illustrates the operations available in the torchvision.ops module for repurposing
```

oke-aditya (commented):
Hey Allen you need to add docs to `docs/ops.rst` where you can use.

```
.. autofunction:: masks_to_boxes
```

This will add docs for this code.

test/test_masks_to_boxes.py
datumbox: @0x00b1 Could make the same changes you did on the galleries also here?
https://github.com/pytorch/vision/pull/4290/files#diff-26925685a7a896911607f04c21b03d9b2c029a0a9530c7006f01c1a994ab8448R45-R53

Aka replace numpy with torch.

datumbox (commented):
@0x00b1 sorry for the back and forth. Adding an operator is possibly one of the most complex things as one needs to add many things across many files. I think we are almost there to merge. Let me summarize the comments that I think remain unresolved:

1. Address the docs failure as described here: https://github.com/pytorch/vision/pull/4290/files#r712457642
2. We missed one use of numpy vs torch. Just copy paste what you got on the examples and we should be good to go. https://github.com/pytorch/vision/pull/4290/files#r712884767
3. Add the masks_to_boxes in docs as described here: https://github.com/pytorch/vision/pull/4290#pullrequestreview-759023390

gallery/plot_repurposing_annotations.py
0x00b1: Nice! I appreciate the debugging.

0x00b1:
> Hey Allen you need to add docs to `docs/ops.rst` where you can use.
> 
> ```
> .. autofunction:: masks_to_boxes
> ```
> 
> This will add docs for this code.

Nice catch! Fixed.

0x00b1:
> @0x00b1 sorry for the back and forth. Adding an operator is possibly one of the most complex things as one needs to add many things across many files. I think we are almost there to merge. Let me summarize the comments that I think remain unresolved:
> 
> 1. Address the docs failure as described here: https://github.com/pytorch/vision/pull/4290/files#r712457642
> 2. We missed one use of numpy vs torch. Just copy paste what you got on the examples and we should be good to go. https://github.com/pytorch/vision/pull/4290/files#r712884767
> 3. Add the masks_to_boxes in docs as described here: [#4290 (review)](https://github.com/pytorch/vision/pull/4290#pullrequestreview-759023390)

It's no problem dude! I sincerely appreciate your and @oke-aditya's patience! In the future, it might be worth investigating whether someone should add a cookiecutter or cookiecutter-like method for generating op scaffolding.

0x00b1:
@datumbox OK. Everything has been addressed. Hopefully we don't see any CI failures! I would also be more than happy to squash these commits down.

torchvision/ops/boxes.py
```diff
@@ -297,3 +297,35 @@ def generalized_box_iou(boxes1: Tensor, boxes2: Tensor) -> Tensor:
...
 
     return iou - (areai - union) / areai
+
+
+def masks_to_boxes(masks: torch.Tensor) -> torch.Tensor:
+    """
+    Compute the bounding boxes around the provided masks
+
+    Returns a [N, 4] tensor.  Both sets of boxes are expected to be in ``(x1, y1, x2, y2)`` format with
+    ``0 <= x1 < x2`` and ``0 <= y1 < y2.
```
datumbox, line 307:
```suggestion
    ``0 <= x1 < x2`` and ``0 <= y1 < y2``.
```

datumbox (approved):
LGTM, thanks a lot @0x00b1. Congrats on your first contribution. :)

NicolasHug:
Hi @0x00b1 ,

Thank you for the great work on this PR!
I'm sorry I wasn't able to make a last pass, I think I missed the point where this PR got un-drafted https://github.com/pytorch/vision/pull/4290#pullrequestreview-734734416.

I only have 2 remaining comments at this point:

- would it be possible to not rely on a hard-coded image for the tests, as asked in https://github.com/pytorch/vision/pull/4290#discussion_r692723546 (sorry again I missed your ping)? It would help make the tests more robust, also avoid storing files in the repo which can end up bloated. The PR was merged already so it will be included anyway I guess.
- Would it be possible to use the `draw_segmentation_masks` and `draw_bounding_box` utilities in the example, as suggested by @oke-aditya in https://github.com/pytorch/vision/pull/4290#discussion_r711102218? It would likely simplify the example and trim it down to its essential part: the new `masks_to_boxes` operator, instead of having lots of plotting code. It would also help users discover these plotting tools that they might come useful in other scenarios.

Would you or @oke-aditya be interested in a follow-up PR with these? The first point might be a bit trickier, but the second one should be reasonably simple.  We can do them in separate PRs. Thanks!

oke-aditya:
I'm fine with either. I will leave choice to @0x00b1

oke-aditya:
Also another thought about the gallery example.
Another example that can be added is to show how a simple Segmentation dataset can be rewritten to detection dataset.
As pointed out in #3960. This is a common use of `masks_to_boxes`.

Adding this example will help users to convert PenFudan / Panopatic datasets easily to detection.
This might look simple, but let's keep an example to help users.

```
from torchvision.ops import masks_to_boxes, box_convert

class SegmentationToDetectionDataset(Dataset):
    def __getitem__(self, idx):
          boxes_xyxy = masks_to_boxes(segmentation_masks)

         # Now for any change of boxes to COCO Format.
          boxes_xywh = box_convert(boxes_xyxy, in_fmt="xyxy", out_fmt="xywh")
          return boxes_xywh
```

torchvision/ops/boxes.py
```diff
@@ -297,3 +297,35 @@ def generalized_box_iou(boxes1: Tensor, boxes2: Tensor) -> Tensor:
...
+
+    Returns:
+        Tensor[N, 4]: bounding boxes
+    """
+    if masks.numel() == 0:
+        return torch.zeros((0, 4))
+
+    n = masks.shape[0]
+
+    bounding_boxes = torch.zeros((n, 4), device=masks.device, dtype=torch.int)
```
oke-aditya, line 321: My initial thought was dtype should be `torch.float`. Since all other ops follow float dtype.

datumbox: Agreed, also the above zeros needs to have a device:
`torch.zeros((0, 4), device=masks.device)`

 Could you please send a PR that fixes these 2 issues? The rest of the doc/test improvements discussed here can happen on a separate PR.

PR #4291, opened 2021-08-19
remove dependency for dtype getters

pmeier:
The dtype getters currently exposed under `torch.testing` will be deprecated / removed in the upcoming PyTorch releases pytorch/pytorch#63554.

### DIFF
--- a/test/common_utils.py
+++ b/test/common_utils.py
@@ -102,11 +102,11 @@ def cycle_over(objs):
 
 
 def int_dtypes():
-    return torch.testing.integral_types()
+    return (torch.uint8, torch.int8, torch.int16, torch.int32, torch.int64)
 
 
 def float_dtypes():
-    return torch.testing.floating_types()
+    return (torch.float32, torch.float64)
 
 
 @contextlib.contextmanager


test/common_utils.py
```diff
@@ -102,11 +102,11 @@ def cycle_over(objs):
 
 
 def int_dtypes():
-    return torch.testing.integral_types()
+    return (torch.uint8, torch.int8, torch.int16, torch.int32, torch.int64)
 
 
 def float_dtypes():
-    return torch.testing.floating_types()
+    return (torch.float32, torch.float64)
```
pmeier, line 109: Should we include `torch.float16` and `torch.bfloat16` here?

datumbox: Sometimes I had to check float16 (see [example](https://github.com/pytorch/vision/blob/90a2402b11b58705884b564b7bfce325f1c7d7c7/test/test_ops.py#L953-L958)) in the test to ensure the operator implementation did not overflow. We don't use the specific method there but maybe it's going to be useful?

NicolasHug: I would say no as they originally weren't part of it, and I think float16 can be a bit tricky to deal with sometimes. We already skip a bunch of float16 test instances.

NicolasHug (approved):
Thanks for the PR @pmeier and for the detailed context in the linked issue!

datumbox (approved):
LGTM, thanks!

test/common_utils.py
pmeier: We could just put them in there and see how many failures we get.

datumbox: It's likely that the universe will break. We could take the approach of adding it "strategically" when needed. Up to you.

Issue #3911, opened 2021-05-25
[RFC] TorchVision with Batteries included - Phase 1


datumbox:
## 🚀 Feature

_Note: To track the progress of the project check out [this board](https://github.com/pytorch/vision/projects/1)._

Add popular primitives (Losses, Schedulers, Data Augmentations, Operators etc) which are often used to reproduce SOTA references and new popular highly accurate models with pre-trained weights to TorchVision.

## Motivation

Though TorchVision currently includes many common building blocks necessary for training CV models, it currently lacks popular primitives which are often used to reproduce SOTA. Some of these primitives are part of our reference scripts (Data utils, transforms etc) because previously did not want to commit to a specific API. Others are part of libraries from the broader ecosystem. Additionally, it does not provide some of the newer, popular architectures which currently achieve good results in a variety of vision tasks.

Adding support of such primitives and models to TorchVision will give a “batteries included” experience to its users. Researchers will be able to do SOTA research and reproduce papers by using common building blocks rather than rewriting their own while industry users will be able to adapt easier the models in their domains using SOTA techniques.

## Pitch

The addition of primitives should be done in several phases, iterating between trying to reproduce SOTA recipes, identifying accuracy gaps and implementing the necessary methods to close them. The progress of this project is tracked on [this board](https://github.com/pytorch/vision/projects/1).

During phase 1, add to TorchVision the following primitives and models:

- Losses #2980:
  - [x] [LabelSmoothing Loss](https://github.com/pytorch/pytorch/issues/7455) https://github.com/pytorch/pytorch/pull/63122 
  - [x] [SoftTarget CrossEntropy](https://github.com/pytorch/pytorch/issues/11959) https://github.com/pytorch/pytorch/pull/61044
- Schedulers:
  - [x] [ChainedScheduler](https://github.com/pytorch/pytorch/pull/26423#discussion_r329976246) https://github.com/pytorch/pytorch/pull/63491 https://github.com/pytorch/pytorch/pull/63457 https://github.com/pytorch/pytorch/pull/65034
  - [x] ConstantLR and LinearLR for [warmup](https://github.com/pytorch/pytorch/pull/60836) https://github.com/pytorch/pytorch/pull/64395
  - [x] [SequentialLR](https://github.com/pytorch/vision/issues/4281) https://github.com/pytorch/pytorch/pull/64037 https://github.com/pytorch/pytorch/pull/65035
- Models #2707:
[...]

PR #4293, opened 2021-08-19
Add EfficientNet Architecture in TorchVision

### LINKED ISSUE #980: [Feature Request] Add EfficientNet
https://ai.googleblog.com/2019/05/efficientnet-improving-accuracy-and.html

It seems like EfficientNet is faster and better than most the other nets out there currently. Would be awesome to see it torchvision soon! :)

datumbox:
Fixes #980


torchvision/models/efficientnet.py
```diff
@@ -0,0 +1,236 @@
...
+
+        # project
+        layers.append(ConvBNActivation(cnf.expanded_channels, cnf.out_channels, kernel_size=1, norm_layer=norm_layer,
+                                       activation_layer=nn.Identity))
+
+        self.block = nn.Sequential(*layers)
+        self.stochastic_depth = StochasticDepth(stochastic_depth_prob, "row")
+        self.out_channels = cnf.out_channels
+
+    def forward(self, input: Tensor, drop_rate: float = 0.0) -> Tensor:
```
datumbox, line 82:
```suggestion
    def forward(self, input: Tensor) -> Tensor:
```

rvandeghen:
Hi,

Do you know when this will be released ?

datumbox:
@rvandeghen No concrete date yet but as you can see I'm working on it. It's still early days and the architecture needs to be verified for all variants, add pre-trained weights etc.

Subscribe to this PR to get notified when it's merged.

datumbox:
The failing tests are not related.

rwightman:
@datumbox thanks for putting the mention of weight sources in doc file and weight links comment, the B0-B4 took some time on my part to train. The B0, B3, and B4 that I trained are all better than the original TF baseline, RandAugment, and AdvProp + AA weights on ImageNet-1k, the B1 and B2 are earlier attempts that are decent, but not to the same level, I didn't follow up on since I rarely use those model sizes.

Re those B5-B7 weights, I believe those are just the weights from the original tensorflow models (https://github.com/tensorflow/tpu/tree/master/models/official/efficientnet) as ported by Luke. They need assymetric TF 'SAME' padding for full performance and will suffer reduced accuracy without. That drop will vary as you change the input size and have different padding offsets through the model relative to what SAME would produce. That can be more problematic using them as backbones for say obj detection / segmentation, esp if you freeze some of the layers.

I'm always open to training B5-B7/B8 but would need some sort of business arrangement + someone elses compute resources as it'd would be a significant commitment in time and compute to iterate and produce good weights.

datumbox:
@rwightman No worries, thanks for your work on it. :)

Indeed B5-B7 require the "same" padding. We were debating whether to add code to handle this in our implementation but we decided not to because a) the degradation is very small (0.1-0.2 points), b) it's slower and c) as you said it can be resolved by retraining the models on the future. For now, users who want to faithfully reproduce the models can use your implementation.

Concerning B1 and B2, yes I noticed that they can be further optimized but I think the models are good enough for now. We are happy to review them once the #3911 epic is completed as it will add lots of standard utils in TorchVision to achieve SOTA results. Some of them are added directly in PyTorch (LabelSmoothing, Warmup schedulers etc), so it might be feasible to reduce the amount of code you maintain on timm and help you focus on bringing more models.

PR #3123, opened 2020-12-04
Implement all AutoAugment transforms + Policies

datumbox:
Fixes #3050 

This PR implements the necessary transforms for AutoAugment:

- [x] posterize
- [x] solarize
- [x] equalize
- [x] sharpness
- [x] autocontrast
- [x] invert

It also implements the following policies:

- [x] ImageNet AutoAugment Policy
- [x] CIFAR10 AutoAugment Policy
- [x] SVHN AutoAugment Policy

PR #4301, opened 2021-08-20
Add StochasticDepth implementation

### LINKED ISSUE #4298: Implement a Stochastic Depth operator
## 🚀 Feature
Implement the Stochastic Depth as described in [Deep Networks with Stochastic Depth](https://arxiv.org/abs/1603.09382) to randomly drop residual branches in residual architectures.

The operator should support 2 modes, dropping the entire batch as described on the original paper and dropping individual rows from the batch as utilized on the [EfficientNet paper](https://arxiv.org/pdf/1905.11946.pdf). We should provide both a functional and a Layer form.

datumbox:
Fixes #4298

### DIFF
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -2,6 +2,7 @@ from common_utils import needs_cuda, cpu_and_gpu, assert_equal
 import math
 from abc import ABC, abstractmethod
 import pytest
+import random
 
 import numpy as np
 
@@ -12,6 +13,11 @@ from torch.autograd import gradcheck
 from torch.nn.modules.utils import _pair
 from torchvision import ops
 from typing import Tuple
+
+try:
+    from scipy import stats
+except ImportError:
+    stats = None
 
 
 class RoIOpTester(ABC):
@@ -998,6 +1004,35 @@ class TestGenBoxIou:
                                        [279.2440, 197.9812, 1189.4746, 849.2019]], dtype=dtype)
             expected = torch.tensor([[1.0, 0.9933, 0.9673], [0.9933, 1.0, 0.9737], [0.9673, 0.9737, 1.0]])
             gen_iou_check(box_tensor, expected, tolerance=0.002 if dtype == torch.float16 else 1e-3)
+
+
+class TestStochasticDepth:
+    @pytest.mark.skipif(stats is None, reason="scipy.stats not available")
+    @pytest.mark.parametrize('mode', ["batch", "row"])
+    @pytest.mark.parametrize('p', [0.2, 0.5, 0.8])
+    def test_stochastic_depth(self, mode, p):
+        random.seed(42)
+        batch_size = 5
+        x = torch.ones(size=(batch_size, 3, 4, 4))
+        layer = ops.StochasticDepth(mode=mode, p=p).to(device=x.device, dtype=x.dtype)
+        layer.__repr__()
+
+        trials = 250
+        num_samples = 0
+        counts = 0
+        for _ in range(trials):
+            out = layer(x)
+            non_zero_count = out.sum(dim=(1, 2, 3)).nonzero().size(0)
+            if mode == "batch":
+                if non_zero_count == 0:
+                    counts += 1
+                num_samples += 1
+            elif mode == "row":
+                counts += batch_size - non_zero_count
+                num_samples += batch_size
+
+        p_value = stats.binom_test(counts, num_samples, p=p)
+        assert p_value > 0.0001
 
 
 if __name__ == '__main__':
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -13,11 +13,6 @@ from torch.nn.modules.utils import _pair
 from torchvision import ops
 from typing import Tuple
 
-try:
-    from scipy import stats
-except ImportError:
-    stats = None
-
 
 class RoIOpTester(ABC):
     dtype = torch.float64
@@ -1006,10 +1001,10 @@ class TestGenBoxIou:
 
 
 class TestStochasticDepth:
-    @pytest.mark.skipif(stats is None, reason="scipy.stats not available")
     @pytest.mark.parametrize('p', [0.2, 0.5, 0.8])
     @pytest.mark.parametrize('mode', ["batch", "row"])
     def test_stochastic_depth(self, mode, p):
+        stats = pytest.importorskip("scipy.stats")
         batch_size = 5
         x = torch.ones(size=(batch_size, 3, 4, 4))
         layer = ops.StochasticDepth(p=p, mode=mode).to(device=x.device, dtype=x.dtype)
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -1008,13 +1008,13 @@ class TestGenBoxIou:
 
 class TestStochasticDepth:
     @pytest.mark.skipif(stats is None, reason="scipy.stats not available")
-    @pytest.mark.parametrize('mode', ["batch", "row"])
     @pytest.mark.parametrize('p', [0.2, 0.5, 0.8])
+    @pytest.mark.parametrize('mode', ["batch", "row"])
     def test_stochastic_depth(self, mode, p):
         random.seed(42)
         batch_size = 5
         x = torch.ones(size=(batch_size, 3, 4, 4))
-        layer = ops.StochasticDepth(mode=mode, p=p).to(device=x.device, dtype=x.dtype)
+        layer = ops.StochasticDepth(p=p, mode=mode).to(device=x.device, dtype=x.dtype)
         layer.__repr__()
 
         trials = 250
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -2,7 +2,6 @@ from common_utils import needs_cuda, cpu_and_gpu, assert_equal
 import math
 from abc import ABC, abstractmethod
 import pytest
-import random
 
 import numpy as np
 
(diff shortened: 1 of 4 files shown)


test/test_ops.py
```diff
@@ -1000,5 +1006,34 @@ def gen_iou_check(box, expected, tolerance=1e-4):
             gen_iou_check(box_tensor, expected, tolerance=0.002 if dtype == torch.float16 else 1e-3)
 
 
+class TestStochasticDepth:
+    @pytest.mark.skipif(stats is None, reason="scipy.stats not available")
```
NicolasHug, line 1010: As a slightly cleaner pattern we could use: `pexpect = pytest.importorskip("pexpect")` in the test body https://docs.pytest.org/en/latest/how-to/skipping.html#summary

But is there a reason we should not expect stats to be available here? I think we rely on scipy unconditionally in other tests

test/test_ops.py
```diff
@@ -1000,5 +1006,34 @@ def gen_iou_check(box, expected, tolerance=1e-4):
...
+    @pytest.mark.parametrize('mode', ["batch", "row"])
+    @pytest.mark.parametrize('p', [0.2, 0.5, 0.8])
+    def test_stochastic_depth(self, mode, p):
+        random.seed(42)
```
NicolasHug, line 1014: Shouldn't this be the torch RNG that we want to fix? I see calls to `torch.rand` in `stochastic_depth()`

torchvision/ops/stochastic_depth.py
```diff
@@ -0,0 +1,56 @@
...
+    Implements the Stochastic Depth from `"Deep Networks with Stochastic Depth"
+    <https://arxiv.org/abs/1603.09382>`_ used for randomly dropping residual
+    branches of residual architectures.
+
+    Args:
+        input (Tensor[N, ...]): The input tensor or arbitrary dimensions with the first one
+                    being its batch i.e. a batch with ``N`` rows.
+        mode (str): ``"batch"`` or ``"row"``.
+                    ``"batch"`` randomly zeroes the entire input, ``"row"`` zeroes
+                    randomly selected rows from the batch.
```
NicolasHug, line 16: Do I understand correctly that with "batch" the zero-ing is done element-wise, e.g. a row can have zero and non-zero elements, whereas with "row", some rows would be entirely zeroed?

NicolasHug: As a follow-up I'm wondering if there could be a more descriptive name instead of "row". I don't know in which context this would be used but for example if we know the input is always a batch of feature maps, this could be "featuremap" instead or something like that?

Again, depending on the context in which this is used, this comment may not make sense. 

Context: I'm unfamiliar with the use of the term "row" in a non-2D array, but perhaps it's a common generalization?

torchvision/ops/stochastic_depth.py
```diff
@@ -0,0 +1,56 @@
...
+
+    survival_rate = 1.0 - p
+    if mode == "batch":
+        keep = torch.rand(size=(1, ), dtype=input.dtype, device=input.device) < survival_rate
+    elif mode == "row":
+        keep = torch.rand(size=(input.size(0),), dtype=input.dtype, device=input.device) < survival_rate
+        keep = keep[(None, ) * (input.ndim - 1)].T
+    else:
+        raise ValueError("mode has to be either 'batch' or 'row', but got {}".format(mode))
+    return input / survival_rate * keep
```
NicolasHug, line 36: The `training` param doesn't seem to be used. Should we only devide by `survival_rate` when it's True?

NicolasHug (commented):
Thanks @datumbox ! I made a brief review with some questions, but this looks good

torchvision/ops/stochastic_depth.py
fmassa, line 36: nit: here are some potential micro-optimizations, as we don't need to divide the whole input by `survival_rate`, but we can do it only in the `keep`. This is similar to what is done in the [PyTorch implementation of dropout](https://github.com/pytorch/pytorch/blob/913c1f83f49f9e1e2a494186cc0069d780cee852/aten/src/ATen/native/Dropout.cpp#L62)

```suggestion
    if mode not in ["batch", "row"]:
        raise ValueError("mode has to be either 'batch' or 'row', but got {}".format(mode))
    size = [1] * input.ndim
    if mode == "row":
       size[0] = input.shape[0]
   noise = torch.empty(size, dtype=input.dtype, device=input.device)
   noise = noise.bernoulli_(survival_rate).div_(survival_rate)
   return input * noise
```

What I like about this implementation is that it makes it even clearer what the differences are between the different modes. But it's a minor difference in clarity, so I'm not set to stone on my proposal.

torchvision/ops/stochastic_depth.py
```diff
@@ -0,0 +1,56 @@
+import torch
+from torch import nn, Tensor
+
+
+def stochastic_depth(input: Tensor, mode: str, p: float, training: bool = True) -> Tensor:
```
fmassa, line 5: What do you think about making the signature be
```python
input: Tensor, p: float, mode: str, training: bool = True
```
instead? I think having `p` come first might be more natural?

torchvision/ops/stochastic_depth.py
```diff
@@ -0,0 +1,56 @@
...
+    else:
+        raise ValueError("mode has to be either 'batch' or 'row', but got {}".format(mode))
+    return input / survival_rate * keep
+
+
+class StochasticDepth(nn.Module):
+    """
+    See :func:`stochastic_depth`.
+    """
+    def __init__(self, mode: str, p: float) -> None:
```
fmassa, line 43: Same comment here regarding making `p` come first

fmassa (approved):
Looks great, thanks a lot!

I have some small comments, let me know what you think. None of those are merge-blocking, so I'm approving it.

Also, I think you need to add an entry in https://github.com/pytorch/vision/blob/master/docs/source/ops.rst for it to appear in the documentation

test/test_ops.py
datumbox: That's what I thought as well but then I saw the refactored version of the transforms and followed the same approach:
https://github.com/pytorch/vision/blob/3fa2055b5a4807a8ffc577c53da180382cafd17c/test/test_transforms.py#L522-L524

Do you think this was a mistake introduced during the refactoring?

torchvision/ops/stochastic_depth.py
datumbox: We use it, check few lines up. And no the formula is correct.

torchvision/ops/stochastic_depth.py
datumbox: I also don't like the name. That's the best we came up with Brian and Francisco. If you find better names, happy to replace them. 

Currently batch drops the entire batch while row drops random records of the batch.

test/test_ops.py
datumbox: I'm using the same pattern we use here:
https://github.com/pytorch/vision/blob/3fa2055b5a4807a8ffc577c53da180382cafd17c/test/test_transforms.py#L17-L20

https://github.com/pytorch/vision/blob/3fa2055b5a4807a8ffc577c53da180382cafd17c/test/test_transforms.py#L512

Perhaps we could create a ticket to refactor all of them?

test/test_ops.py
NicolasHug: It looks like these lines were introduced in https://github.com/pytorch/vision/pull/3123

Perhaps I'm missing what this is supposed to do then: are we trying to fix the randomness of `StochasticDepth`, or the randomness of something else, and what is this "something else"?

As far as I understand the test procedure in itself (the call to scipy.stats) is deterministic?

torchvision/ops/stochastic_depth.py
NicolasHug: From what I can tell we only use it here:

```py
    if not training or p == 0.0:
        return input
```

which makes me wonder whether this parameter is needed? If it's intended, I would suggest to update the docstring to:
> training: apply dropout if is ``True``. If ``False``, this transform is a no-op. Default: ``True``

test/test_ops.py
NicolasHug: sure, but we might as well start using his already :)
We started using it in https://github.com/pytorch/vision/blob/master/test/test_onnx.py#L20 to skip an entire test file

test/test_ops.py
datumbox: > It looks like these lines were introduced in #3123

This is before #3123. See [here](https://github.com/pytorch/vision/blame/4eab7a67f9f134c810eb7d1977e3ef9aa3fa1b26/test/test_transforms.py#L435-L445). Perhaps the intention was not to fix at all the rand on torch and only work on PIL/numpy randomness. I'll check if I can completely remove it.

> As far as I understand the test procedure in itself (the call to scipy.stats) is deterministic?
Yes, that's correct.

torchvision/ops/stochastic_depth.py
datumbox: I'm following the same convention as in PyTorch's Dropout:
https://github.com/pytorch/pytorch/blob/master/torch/nn/modules/dropout.py#L58

NicolasHug: Oh right, I forgot `self.training` gets set automatically. I'm wondering whether the current docstring is accurate though:

> training: apply dropout if is ``True``. Default: ``True``

Should `dropout` be something else?

test/test_ops.py
datumbox: It seems it passes without fixing the seed, so I removed it. If we see flakiness we can set the seed for both all randgens.

datumbox:
Thanks both for the fast review. I believe I addressed all comments. I'll merge after the tests pass.

PR #3597, opened 2021-03-23
Use FX to have a more robust intermediate feature extraction

fmassa:

PR #4302, opened 2021-08-21
Add FX feature extraction as an alternative to intermediate_layer_getter

alexander-soare:
Motivation is similar to https://github.com/pytorch/vision/pull/3597. I just happened to pick up on this while working on timm: https://github.com/rwightman/pytorch-image-models/pull/800

This builds on https://github.com/pytorch/vision/pull/3597 with a few adjustments:
- Keep track of leaf ops as well as leaf modules, giving the user even finer control
- Handle reuse of same op/module using an incremented _{int} suffix on qualified node names. This is automatically done by fx for repeated ops, but not for modules.
- Add utility for printing node qualified names of a model.
- Add wrapper module which converts the input model to a graph module but keeps the input model's non-parametric properties.

A caveat, and possibly a dealbreaker that we should figure out how to handle:
- [x] Static control flow is frozen into place when the GraphModule is produced. This is especially not a good thing for control flow that depends on the state of `model.training`

Other things I will do once we are ready to commit:
- [x] Tests
- [x] More documentation


alexander-soare:
@fmassa just wanted to make sure this topic stays active. I have some checkpoints above explaining what I think needs to be done to close this out. Can you please help me add what's needed there?

I believe you mentioned you'd chat to the fx team about the train/eval problem so I'll be on standby to pick it back up based on the outcome of that.

fmassa:
Thanks a ton for the PR, let me know when you'll want me to have a closer look at it.

For the points you brought, I have a few questions:

> Handle reuse of same op/module using an incremented _{int} suffix on qualified node names. This is automatically done by fx for repeated ops, but not for modules.

Maybe I'm missing something here, but is it even possible to have two modules with the same name? If this is for handling things like `layer3.2` being equal to `layer3`, is the best way to increment the node names (i.e., creating a new node) instead of having a one-to-many mapping from node to module?

> Add wrapper module which converts the input model to a graph module but keeps the input model's non-parametric properties.

Should this be a new `nn.Module`, or should we instead re-use the same module returned by FX and add the missing attributes there? What's the benefits of wrapping it in a `nn.Module`? The current module wrapper doesn't have the same state-dict as the original model, which makes it a bit annoying to be seamlessly used to load state-dicts etc.

> Static control flow is frozen into place when the GraphModule is produced. This is especially not a good thing for control flow that depends on the state of model.training

I've discussed about this with @jamesr66a . His view was that this was a special case of the more general re-specialization problem (like for specializing over shape / ranks of Tensors), and as of now FX is not going to be handling it explicitly, but nothing blocks us from having our own tooling to do this, and I think we should do it.

A naive way would be to trace the model twice (one in `training=True` and one with `training=False`) and have a wrapper `nn.Module` be returned that does dispatch to one of the two traces depending on the training mode.

A naive solution is something in the lines of
```python
class WrappedModel(nn.Module):
    def __init__(self, model, ...):
        super().__init__()
        mode = model.training
        self.train_model = trace(model.train(), ...)
        self.eval_model = trace(model.eval(), ...)
        model.train(mode)

    def forward(self, *args, **kwargs):
        if self.training:
            return self.train_model(*args, **kwargs)
        else:
            return self.eval_model(*args, **kwargs)
```
but ideally we wouldn't duplicate parameters nor change the namespace of the parameters.

> Also, do you think there's an appropriate place to add more docs than what's in the docstrings?

Yes, I think we should either make a new doc page under https://github.com/pytorch/vision/tree/main/docs/source or use the [new galleries](https://github.com/pytorch/vision/tree/main/gallery) to better illustrate how this functionality can be used.
But this can come once we have finalized the APIs / etc, and can come as a follow-up PR.

> I have some checklist items above explaining what I think needs to be done to close this out. Can you please help me add what's needed there?

The points I've made above might still require iterating a bit on some of the aspects implemented here (but please let me know if I'm missing something!). Also, adding tests will be important to validate that the code behave as expected (including dummy `nn.Module` that exercise the `train` / `eval` differences if this gets implemented now).

From my perspective, we can break this PR into multiple ones if it facilitate things for you:
- get an initial version that works for all models in torchvision and handles the issues I mentioned in the first point in https://github.com/pytorch/vision/pull/3597#issuecomment-898386181 . It would also be good to know what percentage of `timm` models are covered with this initial version (including what are most common failure modes), so we can plan accordingly next steps
- extend support for train / eval differences
- gallery with examples
- fix some of the long-tail models to make the code more robust
- (maybe?) upstream the feature to PyTorch

Thoughts?

alexander-soare:
@fmassa thanks for your input! I've made some good progress with the train/eval problem but I'm out of time for this week and will move onto tests/docs next week. Might be worth waiting till then before you have a closer look.

But to answer the points that you already have:

> Maybe I'm missing something here, but is it even possible to have two modules with the same name?

I'm talking about something like: user defines `self.relu = nn.ReLU()` then uses it twice in some `forward` method. Without my modification, the qualnames would both come out as `walk.path.to.relu`. With my modification we get: `walk.path.to.relu` and `walk.path.to.relu_1`. So it's possible to disambiguate if we want to choose these as return nodes.

> ... is the best way to increment the node names (i.e., creating a new node) instead of having a one-to-many mapping from node to module?

FX creates the new nodes and the node to module mapping, and I'm thinking we leave that behaviour as it is. All we're doing is making a node to string mapping, because we've made it such that the strings follow a nicer convention.

> Should this be a new nn.Module, or should we instead re-use the same module returned by FX and add the missing attributes there? 

I agree. I just deleted it and incorporated the desired behaviour into `build_feature_graph_net`.

> A naive way would be to trace the model twice ... but ideally we wouldn't duplicate parameters nor change the namespace of the parameters.

Good suggestion. That's what I did + some gymnastics to not duplicate params or change the namespace.

fmassa:
Hi @alexander-soare 

Thanks for the clarifications, makes sense!

Let's get this first version of the PR merged into torchvision this week. Just tests + docs should be good for now.

> Would it make sense to put a version of those guidelines in the docs and gradually phase them out as/if core FX updates solve those problems

Good question, I think it might be good to have some high-level guidelines in the documentation as well. I'm not sure FX will be able to capture `assert` without using `torch._assert`, but I'd feel a bit reluctant to propose users to rely on a private API.

I'll do a first pass on the PR today with some more high-level comments.

torchvision/models/_utils.py
```diff
@@ -64,3 +73,395 @@ def forward(self, x):
...
+        model, graphs['train'], graphs['eval'], class_name=name)
+
+    # Keep non-parameter model properties for reference
+    for attr_str in model.__dir__():
+        attr = getattr(model, attr_str)
+        if (not attr_str.startswith('_')
+                and attr_str not in graph_module.__dir__()
+                and not ismethod(attr)
+                and not isinstance(attr, (nn.Module, nn.Parameter))):
+            setattr(graph_module, attr_str, attr)
```
fmassa, line 462: What are some of the use-cases that you had in mind for this part of the code?
We are not copying methods (which is ok), but for example if a particular family of `nn.Module` has methods like `no_weight_decay` or `forward_features`, they won't get propagated.

What should be the "guarantees" that we provide when copying some of the attributes?

torchvision/models/_utils.py
```diff
@@ -64,3 +73,395 @@ def forward(self, x):
...
+    Args:
+        model (nn.Module): model on which we will extract the features
+        return_nodes (Union[List[name], Dict[name, new_name]])): either a list
+            or a dict containing the names (or partial names - see note above)
+            of the nodes for which the activations will be returned. If it is
+            a `Dict`, the keys are the qualified node names, and the values
+            are the user-specified keys for the graph module's returned
+            dictionary. If it is a `List`, it is treated as a `Dict` mapping
+            node specification strings directly to output names.
+        tracer_kwargs (Dict): a dictionary of keywork arguments for
```
fmassa, line 341: nit: have to update the doc with the new `train` / `eval` arguments

fmassa (commented):
This is looking great, thanks a lot @alexander-soare !

I think we are almost ready to get this merged.

Also, I was wondering if we should add it in a different location in torchvision that is not private, maybe `torchvision.models.utils` or `torchvision.models.feature_extraction` or something like that?

torchvision/models/_utils.py
alexander-soare: @fmassa The use case I had in mind was for when there are properties that hold useful information. For example `model.out_features` is a common one in the `timm` library. 

Hmm I'll think about propagating methods a little more. I guess my original worry was that the methods might make use of parameters, and not all parameters will necessarily exist in the final graph module.

alexander-soare:
@fmassa 

> Also, I was wondering if we should add it in a different location in torchvision that is not private, maybe torchvision.models.utils or torchvision.models.feature_extraction or something like that?

That  makes sense. Will drop that in a commit. But first a question, would you sat that's also true for  the original `IntermediateLayerGetter`?

torchvision/models/_utils.py
alexander-soare: > What should be the "guarantees" that we provide when copying some of the attributes?

Maybe you saw, but I have this in the docstring:
>An attempt is made to keep all non-parametric properties of the original
    model, but existing properties of the constructed `GraphModule` are not
    overwritten.

If you read that coming in fresh would it be sufficient or would you still have questions?

fmassa: @alexander-soare 

> Hmm I'll think about propagating methods a little more. I guess my original worry was that the methods might make use of parameters, and not all parameters will necessarily exist in the final graph module.

I agree with you, I think it's better for now to avoid dealing with methods, as it can have many issues as you pointed out.
My question as more if we should restrict ourselves to a fixed set of properties to copy (like only numbers).

Also, in the case you pointed out, `model.out_features`, won't give the expected values for the new model, as we can extract features from arbitrary locations of the model (which can have a different number of channels).
This by itself could be a reason why we might not want to propagate meta-information, but I'm open for discussion.

(As a side note, I think we should stop relying on `out_features` for finetuning, and instead use `LazyModule` instead.

fmassa:
Thanks for the changes @alexander-soare !

> I also moved IntermediateLayerGetter there and made sure to reroute references to it. Let me know if there are any issues with that and I can revert it.

I've made `IntermediateLayerGetter` private because of all the limitations it had. The way I see things, `build_feature_graph_net` entirely replaces the need for `IntermediateLayerGetter`, so I would be just fine keeping it in `_utils.py` for now, and maybe issuing a deprecation warning (but let's do this in a separate PR), and not put it in the documentation neither (that's why it isn't documented).

I'm bikeshedding here, but I'm thinking if we can find another name for `build_feature_graph_net` and `print_graph_node_qualified_names`. Here is my reasoning:
- `print_graph_node_qualified_names` has a bit too specific nomenclature, and I'm not sure if FX uses `qualified_names` much in its documentation. Maybe something like `get_graph_node_names` might be better?
- `build_feature_graph_net` is a bit weird to me as well, maybe `get_submodel` or `get_intermediate_execution_graph` or something like that?

torchvision/models/feature_extraction.py
```diff
@@ -0,0 +1,574 @@
...
+          the `tracer_kwargs`.
+        - Use `torch.matmul(tensor_a, tensor_b)` instead of `tensor_a @
+          tensor_b`.
+
+    For further information on FX see the
+    `torch.fx documentation <https://pytorch.org/docs/stable/fx.html>`_.
+
+    Args:
+        model (nn.Module): model on which we will extract the features
+        return_nodes (Optional[Union[List[str], Dict[str, str]]]): either a `List`
```
NicolasHug, line 386: We should avoid writing type-annotations in the docstrings: they're usually much less readable than a simple plain English description, and annotations are targeted for machines, not humans. In this case this could just be something like

```suggestion
        return_nodes (list or dict, optional) either a `List`
```

and the details of the expected types can be described below. This is how we usually try to document stuff: https://pytorch.org/vision/stable/transforms.html#torchvision.transforms.RandomCrop

test/test_backbone_utils.py
```diff
@@ -1,5 +1,9 @@
+import unittest
```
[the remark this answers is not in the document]
NicolasHug: Let's not make this a nit but a requirement instead please 😇

torchvision/models/feature_extraction.py
datumbox: I remember you raising this in the discussion at #2025. Not sure we reached an agreement to replace `List[str]` with `list` in the discussion. Personally I'm happy with the precise definition but I acknowledge there is a certain degree of subjectivity on what's readable.

NicolasHug: @datumbox I believe these are 2 unrelated discussions.

Our discussion in #2025 was about whether having type annotations or not. I'm still firmly against it, but consensus wasn't reached.

This here however is about how we specify the types of the parameters in docstrings. We should never ever use the type annotation syntax for that.

NicolasHug: @datumbox it looks like you reacted with a "thumbs down" emoji to my comment above; would you mind elaborating? What part do you disagree with?

datumbox: @NicolasHug I don't agree with your "never ever" statement but at the same time I don't think it's worth over analyzing it. Sometimes people agree to disagree and move on. :)

torchvision/models/_utils.py
alexander-soare: I didn't think the `out_features` example through :sweat_smile: . I get you. I'm just going to drop this altogether actually. Better to leave the responsibility with the user if they want to propagate anything across, and we can learn from feature requests if there are any.

(didn't know about `LazyModule` - awesome!)

torchvision/models/feature_extraction.py
```diff
@@ -0,0 +1,574 @@
...
+          hold a reference to that module's forward method.
+        - Likewise, you may turn functions into leaf functions by passing a
+          list of `autowrap_functions` as one of the `tracer_kwargs` (see
+          example below).
+        - Some inbuilt Python functions can be problematic. For instance,
+          `int` will raise an error during tracig. You may wrap them in your
+          own function and then pass that in `autowrap_functions` as one of
+          the `tracer_kwargs`.
+        - Use `torch.matmul(tensor_a, tensor_b)` instead of `tensor_a @
+          tensor_b`.
```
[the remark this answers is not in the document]
alexander-soare: https://github.com/pytorch/pytorch/issues/64483 for ref

torchvision/models/feature_extraction.py
alexander-soare: Just making sure to acknowledge this. I had a look at the the other examples in the repo and found a bit of a mix of both approaches. So I'm happy to leave it the way I have it now, unless a clear decision is made. Resolving this for now. And thanks both for the feedback!

NicolasHug: would you mind pointing to the current occurrences of annotation syntax in the docstrings? If there are any we should update them

alexander-soare: @NicolasHug this was just with a random click through of ctrl+F "Args:" in my IDE. Found an example for you: https://github.com/pytorch/vision/blob/e95db7aa09ac012dfbdb5db8a24ba8e5804d55b5/torchvision/utils.py#L162

alexander-soare:
@fmassa thanks for the continued support on this! I've sorted it all out in the latest commit. Two more things

**Thing 1:**

About the naming. Cool, I've changed them. I went for `build_feature_extractor` which I hope is also in line with your motivation (abstracting from specific nomenclature). If not, I have no problem at all changing it to whatever works.

I also dropped the use of the term "qualified node name" in the parts that are visible to the user. I do have one bikesheddy concern about this though. Calling it just a "node name" might not be specific enough, and may clash with other notions. An actual fx Node already has some sort of notion of a name attached to it. In fact, fx docs even mention qualified node names. But that's slightly different from the node naming convention I've put together. BUT, happy to go ahead with this PR as is, and leave this as a note.

**Thing 2**:

> I would like to pull the PR locally and play a bit with it once you mark the PR ready for review.

Maybe it's ready for you to go ahead and do that - and in fact that would be welcome at this point. But before marking as ready, I wanted to have a go at training a detection model as the final test. Hoping to do that on the weekend.

torchvision/models/feature_extraction.py
alexander-soare: @NicolasHug sorted

torchvision/models/feature_extraction.py
```diff
@@ -0,0 +1,507 @@
...
+        suppress_diff_warning (bool): whether to suppress a warning when there
+            are discrepancies between the train and eval version of the graph.
+            Defaults to False.
+
+    Examples::
+
+        >>> # Feature extraction with resnet
+        >>> model = torchvision.models.resnet18()
+        >>> # extract layer1 and layer3, giving as names `feat1` and feat2`
+        >>> model = torchvision.models._utils.build_feature_extractor(
```
oke-aditya, line 352:
```suggestion
        >>> model = torchvision.models.feature_extraction.build_feature_extractor(
```

Also If we are to make `feature_extraction` public as a stable feature.
It would need an import in the `__init__` file.
Of course the import would then change to
```suggestion
        >>> model = torchvision.models.build_feature_extractor(
```

docs/source/index.rst
```diff
@@ -32,6 +32,7 @@ architectures, and common image transformations for computer vision.
    :caption: Package Reference
 
    datasets
+   feature_extraction
```
oke-aditya, line 35: It looks like `feature_extraction` is a submodule or a `folder` in torchvision, but I think the files are placed under `models/feature_extraction`

oke-aditya:
Hi, I just had a very small look at this. I was very eager for this feature. 
Probably there is small confusion in placing the files?
Or maybe I have misunderstood something.

torchvision/models/feature_extraction.py
alexander-soare: @oke-aditya thanks! Sorted and will push soon

docs/source/index.rst
alexander-soare: Wasn't keen on putting it in the models docs as that seems to have a consistent theme of talking about models, how to load them, and their metrics. And I'm a noob with auto-docs so didn't know if I could nest it in they table of contents but still have its own page. Any suggestions?

alexander-soare: Re. your comment below:

> Probably there is small confusion in placing the files?

Yes probably :)

oke-aditya: I'm pretty much noob too :smile: Probably the maintainers here would tell you better solution for docs.

alexander-soare:
Trained ImageNet classification with pretrained resnet18 + concatenated FPN features with both `IntermediateLayerGetter` and `build_feature_extractor`. Trained each for 1 epoch printing loss values (6 decimal places) for approx 100 intervals. No difference between the two feature extraction methods.

@fmassa this covers off my "Thing 2" [above](https://github.com/pytorch/vision/pull/4302#issuecomment-912534011) so from my end this is ready for review.

docs/source/feature_extraction.rst
```diff
@@ -0,0 +1,126 @@
...
+          # Get a resnet50 backbone
+          m = resnet50()
+          # Extract 4 main layers (note: you can also provide a list for return
+          # nodes if the keys and the values are the same)
+          self.body = build_feature_extractor(
+              m, return_nodes=['layer1', 'layer2', 'layer3', 'layer4'])
+          # Dry run to get number of channels for FPN
+          inp = torch.randn(1, 3, 224, 224)
+          with torch.no_grad():
+              out = self.body(inp)
```
fmassa, line 104: nit: unfortunately this can have bad side-effects if the model has batch-norm, as the batch-norm statistics can be messed up (and in previous versions of PyTorch, forwarding a 1-sized batch to BatchNorm would raise an error).

Not sure we can do much better now, apart from changing the `FeaturePyramidNetwork` to infer `in_channels_list`.

fmassa (commented):
I've tried the PR locally and it worked great, thanks!

I've just noted one potential (unwanted?) side-effect of the prefix matching, let me know what you think.

Also, the test failures can be fixed with some of the comments I made.

I think after this batch of comments is addressed we are good to merge, thanks a ton @alexander-soare !

docs/source/feature_extraction.rst
alexander-soare: Ah yeah, I just put the batch size as 2 then. In any case, I'd expect that the user advanced enough to use this feature will probably take the example with a grain of salt.

torchvision/models/feature_extraction.py
```diff
@@ -0,0 +1,510 @@
...
+        available_nodes = list(tracer.node_to_qualname.values())
+        # FIXME We don't know if we should expect this to happen
+        assert len(set(available_nodes)) == len(available_nodes), \
+            "There are duplicate nodes! Please raise an issue https://github.com/pytorch/vision/issues"
+        # Check that all outputs in return_nodes are present in the model
+        for query in mode_return_nodes[mode].keys():
+            # To check if a query is available we need to check that at least
+            # one of the available names starts with it up to a .
+            if not any([re.match(rf'^{query}.?', n) is not None
+                        for n in available_nodes]):
```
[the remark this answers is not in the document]
alexander-soare: Nice catch! Pushing a fix soon

fmassa:
@alexander-soare also, following some discussion we had, maybe using `create_feature_extractor` would be better than `build_feature_extractor` (the `build` might look like the builder pattern, which is not what we are doing here).
Would you mind changing it as well?

torchvision/models/feature_extraction.py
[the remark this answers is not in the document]
fmassa: @lw proposed something along the lines of
```python
re.match(rf'^{query}($|(?=\.))', n) 
```
can you double-check? This is a python-specific extension called positive lookahead.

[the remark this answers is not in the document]
lw: The positive lookahead only matters if you don't want to consume the string. But here you just check whether there's a match, hence that doesn't matter. You could just use
```
re.match(rf'^{query}($|\.)', n) 
```
Note that `.` in regexes has a very special meaning thus if you want a literal period you always need to escape it.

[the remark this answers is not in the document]
alexander-soare: Ah just saw this. Went with @lw's approach.

alexander-soare:
@fmassa that's all sorted. Thanks!

torchvision/models/__init__.py
```diff
@@ -8,8 +8,8 @@
 from .mobilenet import *
 from .mnasnet import *
 from .shufflenetv2 import *
-from .efficientnet import *
```
fmassa, line 11: Can you add this like back?

alexander-soare: Yep on it :) Sorry!

alexander-soare: Nope, there's more to do. Need to add a leaf module that can't be traced through from effnet. Will send it soon

alexander-soare:
@fmassa see me comment above about your last push. Will pull your changes and tack on the last bit

fmassa:
@alexander-soare should we change the implementation of EfficientNet so that it can be properly traced?

Anyway, we can for now skip testing efficientnets, and tackle that in a follow-up PR?

cc @datumbox

fmassa:
@alexander-soare don't bother about fixing EfficientNet, we will do it ourselves. For now, can you just disable efficientnets in the testing of FX, so that we can move forward? Something like filtering `efficientnet` from the `get_classification_models` should be enough

fmassa:
@alexander-soare FYI we are changing `StochasticDepth` so that it is FX-traceable, see https://github.com/pytorch/vision/pull/4372 for the fix.

Anyway, we can get your diff merged once tests finish (lint is failing btw), and then revert the latest changes so that we test `StochasticDepth` as well

alexander-soare:
> For now, can you just disable efficientnets in the testing of FX, so that we can move forward? 

@fmassa yeah this does make sense. Was just trying to make sure there wasn't another hidden reason for the issue. So as it stands, Stochastic depth is treated as a leaf module, and I can revert it once you're ready.

fmassa:
@alexander-soare there are still some minor test failures following your last changes, do you mind fixing it so that we can get this merged?

alexander-soare:
@fmassa sorted - introduced it last time... This is the one though. I can feel it

fmassa:
Thanks a ton for all your work @alexander-soare !

alexander-soare:
@fmassa was a pleasure. And thank you again for all the support!

torchvision/models/feature_extraction.py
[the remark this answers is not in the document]
fmassa: FYI this has been fixed in PyTorch already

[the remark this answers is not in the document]
alexander-soare: @fmassa yep I removed it at the time

Issue #4286, opened 2021-08-18
Addition of OHEM sampler


rush2406:
## 🚀 Feature
Addition of OHEM (online hard example mining) sampler

## Motivation

Currently, a random sampler is being used in RoIHeads. However, other samplers have been shown to perform better than random sampling and heuristics. Thus, this motivates the need to introduce other samplers for example OHEM.

## Pitch

There can be flag/option introduced to use a particular sampler. Based on the flag, the corresponding sampler would be used.

datumbox:
I agree it would be useful. The key challenge is providing a good generic API what can be used across multiple use-cases without sacrificing speed.

I faced something similar when I was writing the implementation of the SSD model which includes a Hard negative sampling component on the loss. My original reaction was to create a sampler similar to [BalancedPositiveNegativeSampler](https://github.com/pytorch/vision/blob/master/torchvision/models/detection/_utils.py#L11) nevertheless it became clear that writing a [custom implementation](https://github.com/pytorch/vision/blob/3e7653c8c4b5eb0c854251a0fb86be18de55bc68/torchvision/models/detection/ssd.py#L268-L276) in the SSD loss that exploits precomputed stats was much more efficient. I'm sure that with some thought one could provide a good enough generic implementation but it's worth discussing first the potential approaches to cover the different use-cases.

rush2406:
@datumbox probably I could create a pull request for OHEM. Based on that, the changes which are required for including new samplers would become clearer?

datumbox:
@rush2406 If you would like to start an experimental PR to investigate options that's totally fine. It's a great way to speed up conversations and might even be merged if it has no breaking changes. I do suspect that it will be BC-breaking though and hence why I marked this proposal with "discussion needed".

fmassa:
Hi @rush2406 

Thanks for the proposal!

Correct me if I'm wrong, but have OHEM been used in newer models since its introduction, or have something like `focal_loss` replaced most of its usages nowadays?

rush2406:
Hi @fmassa 
As far I as know, focal loss is usually used in the 1 stage detectors to counter imbalance between foreground and background proposals which is not an issue in 2 stage detectors. OHEM would be more useful for 2 stage detectors like faster rcnn.

rush2406:
@datumbox @fmassa could you please review #4305 and provide feedback?

fmassa:
Hi @rush2406 

[...]

PR #4305, opened 2021-08-22
Changes related to OHEM sampler

rush2406:
Contains changes related to the addition of new samplers such as OHEM

### DIFF
--- a/torchvision/models/detection/_utils.py
+++ b/torchvision/models/detection/_utils.py
@@ -78,11 +78,11 @@ class BalancedPositiveNegativeSampler(object):
 
 class OhemPositiveNegativeSampler(object):
     """
-    This class samples all positiveand negative proposals
+    This class samples all positive and negative proposals
     """
 
     def __init__(self):
-        
+
         pass
 
     def __call__(self, matched_idxs):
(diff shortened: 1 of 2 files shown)


datumbox:
@rush2406 Thanks for the contribution.

I think your PR showcases the problem we briefly discussed on the issue #4286. The current API is problematic and there is no much one can do without potentially redesigning it. To clarify this is not a problem of your approach, it's an issue of the current code structure that does not allow you do put things where one would expect them to be. Here are some examples:
- The `OhemPositiveNegativeSampler` is not actually doing the hard negative sampling cause it's missing the loss information. Instead we select everything and the real sampling is elsewhere. This is quite misleading.
- The `mining()` is forced to do everything. It estimates the loss, it does the negative sampling and runs NMS. This means it can be misaligned with other methods such as the losses and the post-processing.

Due to the above issues when I implemented the SSD, I ended up putting the upsampling directly in the loss estimation. This implementation choice was not highlighted on the review or related issues. I think this PR showcases the issue and can help us make the right decisions on the future when we redesign the API.

rush2406:
@datumbox thanks for the feedback. The points that you have mentioned are relevant. I guess it would be useful to have something like a base sampler from which other custom samplers could easily be extended similar to [this](https://github.com/open-mmlab/mmdetection/tree/master/mmdet/core/bbox/samplers)

Issue #4296, opened 2021-08-20
Avoid string comparisons for PyTorch version comparison


fmassa:
## 🚀 Feature

PyTorch now exposes [`torch.__version__` as a `TorchVersion` object](https://github.com/pytorch/pytorch/blob/1022443168b5fad55bbd03d087abf574c9d2e9df/torch/torch_version.py#L13), which allows for better version comparisons.
We might be able to simplify things like https://github.com/pytorch/vision/pull/4285/files#r692734821

oke-aditya:
I can probably dig into this once I get time.

nivedwho:
The new method you suggested only works from 1.10.0 torch version right? In that case would it make sense to use it here, since earlier versions would throw an error? I am new to this library so please correct me if I am wrong.

fmassa:
@nivedwho modifying this in torchvision would be fine as latest torchvision depends on latest PyTorch. For projects that depend on both torchvision and PyTorch and are pinned to an older version, `TorchVersion` might not be available so another solutions (specific to those projects) would need to be used.

PR #4307, opened 2021-08-23
Avoided string comparisons for PyTorch version comparison

ABD-01:
Fixes https://github.com/pytorch/vision/issues/4296

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -147,9 +147,7 @@ def get_extensions():
     )
 
     is_rocm_pytorch = False
-    TORCH_MAJOR = int(torch.__version__.split('.')[0])
-    TORCH_MINOR = int(torch.__version__.split('.')[1])
-    if TORCH_MAJOR > 1 or (TORCH_MAJOR == 1 and TORCH_MINOR >= 5):
+    if torch.__version__ >= "1.5":
         from torch.utils.cpp_extension import ROCM_HOME
         is_rocm_pytorch = True if ((torch.version.hip is not None) and (ROCM_HOME is not None)) else False
 


setup.py
```diff
@@ -147,9 +147,7 @@ def get_extensions():
     )
 
     is_rocm_pytorch = False
-    TORCH_MAJOR = int(torch.__version__.split('.')[0])
-    TORCH_MINOR = int(torch.__version__.split('.')[1])
-    if TORCH_MAJOR > 1 or (TORCH_MAJOR == 1 and TORCH_MINOR >= 5):
+    if torch.__version__ >= "1.5":
```
NicolasHug, line 150: Thanks for the PR @ABD-01 , 

I believe the original point of https://github.com/pytorch/vision/issues/4296 was to rely on the TorchVersion object: https://github.com/pytorch/pytorch/blob/1022443168b5fad55bbd03d087abf574c9d2e9df/torch/torch_version.py#L13, instead of just reverting the changes introduced in https://github.com/pytorch/vision/pull/4285

ABD-01: Got it. 
In this repository there's version comparison only once and that is in [setup.py](https://github.com/pytorch/vision/blob/main/setup.py).
So I guess, there's no other file that need changes.

vadimkantorov: Does `torch.__version__` object support comparisons to tuples? If yes, it would look much less error-prone: `if torch.__version__ >= (1, 5)` - at least it would be clear for code reader that some magic in `torch.__version__` is happening and that it's not just a string.

There are also similar problems with torchvision itself: https://github.com/HoBeom/detr/blob/master/util/misc.py#L21

fmassa: According to [the documentation](https://github.com/pytorch/pytorch/blob/1022443168b5fad55bbd03d087abf574c9d2e9df/torch/torch_version.py#L22-L32), `TorchVersion` supports comparing to strings, tuples and `Version` objects. Maybe the safest / most robust [would be to use `Version`](https://github.com/pytorch/pytorch/blob/1022443168b5fad55bbd03d087abf574c9d2e9df/torch/torch_version.py#L6), but it is worth checking more carefully

fmassa:
Waiting on https://github.com/pytorch/pytorch/pull/63848 to be merged before we merge this

ABD-01:
> Waiting on [pytorch/pytorch#63848](https://github.com/pytorch/pytorch/pull/63848) to be merged before we merge this

Cool. Thanks

nivedwho:
@fmassa The other PR is merged and `TorchVersion` should work now. But this PR does not use `TorchVersion` and hence does not really address the issue you mentioned right?

ABD-01:
It uses [`torchversion`](https://github.com/pytorch/pytorch/blob/master/torch/torch_version.py#L11), with new methods `__lt__()`,`__ge__()` and `__le__()` this will work perfect when performing operation: 'greater than equals to' which [currently](https://github.com/pytorch/vision/blob/master/setup.py#L150) is done by splitting str and comparing the converted ints.

ABD-01:
Thank You 😊

PR #4306, opened 2021-08-23
Prep change for changing def branch to main

malfet:
Renames default branch in CI configs an in doc references to `main`

PR #4308, opened 2021-08-23
fix dependency table

pmeier:
#4306 broke the dependency table in our README. This fixes that and adds some more constraints that weren't being tracked properly.


NicolasHug (approved):
Thanks Philip, LGTM

I'm wondering what the reason is for specifying upper bounds on the Python version: aren't all 3.X versions backward-compatible anyway?

pmeier:
We only build binaries for the versions I've given in the table. So if you are on Python 3.9, you cannot `pip install` anything before `torchvision<0.8.2`.

PR #4311, opened 2021-08-24
[FBcode->GH] Port quantize_val and dequantize_val into torchvision to avoid at::native and android xplat incompatibility

datumbox:
Summary:
This diff ports `quantize_val` and `dequantize_val` from at::native to torchvision because native kernels are incompatible with android xplat builds (see D30234056).

This should only be temporary until we find a way to move those functions out of at::native, or until the at::native / android incompatibility disappears.

Reviewed By: fmassa

Differential Revision: D30393619

fbshipit-source-id: 18b7b1b349ad9a24088a120e23da7535f7fa7ddc


datumbox:
This PR is a sync from FBcode. I suggest merging it even if it breaks master due to clang formatting issues and fix ASAP on a follow up PR. This will simplify the process of syncing the change back to FBcode later on.

fmassa (approved):
Thanks!

PR #3389, opened 2021-02-12
Implementing multithreaded video decoding

bjuncek:
Addressing issue #3388 on implementing multithreaded video decoding. 

I attach the _sequential_ reading speeds with multiple number of threads compared to openCV on test videos in torchvision. Note that 0 denotes disabled multithreading. 

download

download (1)

Issue #3388, opened 2021-02-12
[video] multithreaded video decoding


bjuncek:
## 🚀 Feature
At the moment, video decoding is forced to be single threaded, which is limiting the performance to an extent.
We want to enable users to enable multithreaded decoding if they want/need to.

## Motivation

At the moment, video decoding is forced to be single threaded, which is limiting the performance to an extent. Both Decord and OpenCV use multithreaded decoding. 

## Pitch

Enabling `VideoReader(path, stream, num_threads)` so that users can choose to use multithreaded decoding.

cc @bjuncek

bjuncek:
Closed via #3389

PR #4330, opened 2021-08-27
[VideoReader] FFMPEG Multithreading, take2

bjuncek:
Addressing issue #3388 on implementing multithreaded video decoding, and superseeding #3389.


torchvision/io/__init__.py
```diff
@@ -107,7 +111,7 @@ def __init__(self, path, stream="video"):
                 + "ffmpeg (version 4.2 is currently supported) and"
                 + "build torchvision from source."
             )
-        self._c = torch.classes.torchvision.Video(path, stream)
+        self._c = torch.classes.torchvision.Video(path, stream, num_threads)
```
prabhat00155, line 114: num_threads is not defined anywhere. Should it be an argument in `__init__()`?

torchvision/csrc/io/decoder/stream.cpp
```diff
@@ -49,7 +51,51 @@ int Stream::openCodec(std::vector<DecoderMetadata>* metadata) {
...
+    codecCtx_->thread_count = num_threads;
+
+  } else {
+    // otherwise set sensible defaults
+    // with the special case for the different MPEG4 codecs
+    // that don't have threading context functions
+    // TODO: potentially automate this using native c++ function lookups
+    if (strcmp(codecCtx_->codec->name, "mpeg4") == 0 &&
+        codecCtx_->codec_type == 0) {
+      if (codecCtx_->codec_tag == 1684633208) {
```
prabhat00155, line 77: What is this magic number here?

torchvision/csrc/io/decoder/stream.cpp
```diff
@@ -49,7 +51,51 @@ int Stream::openCodec(std::vector<DecoderMetadata>* metadata) {
                << ", avcodec_alloc_context3 failed";
     return AVERROR(ENOMEM);
   }
+  // multithreading heuristics
+  // this is a dumb conservative limit; ideally we'd use 
+  // int max_threads = at::get_num_threads(); but this would cause 
+  // fb sync to fail as it would add dependency to ATen to the decoder API
+  int max_threads = 12;
```
prabhat00155, line 58: Should we do a `#define` for this instead of using a magic number inside a function?

prabhat00155:
Thanks @bjuncek! Could you fix the linter/format errors?

torchvision/csrc/io/decoder/stream.cpp
bjuncek: I suppose we can yes; is it customary to do that in the header file?

torchvision/csrc/io/decoder/stream.cpp
bjuncek: This is codec signature for videos from HMDB and some from UCF.
In principle, we can avoid this by checking if the codec implements threading context function, but I'm  not sure how to do it; I'll ask around.

torchvision/io/__init__.py
bjuncek: Ah yes - forgot to add this

torchvision/csrc/io/decoder/stream.cpp
prabhat00155: Okay sure, the alternative solution sounds better.

torchvision/csrc/io/decoder/stream.cpp
prabhat00155: Do you suspect max_threads will be used elsewhere? Perhaps declaring it a `const` might be better(could be under namespace ffmpeg.

prabhat00155 (commented):
Thanks @bjuncek! Could you also resolve the merge conflicts and linter errors?

torchvision/csrc/io/decoder/stream.cpp
bjuncek: No, not really. At least not in the same context.

bjuncek: I've defined `max_threads` as the protected const int in the header file until we find a better way of solving this.

torchvision/csrc/io/decoder/__init__.py
```diff
@@ -0,0 +1,208 @@
+import torch
```
prabhat00155, line 1: I thought `csrc` directory is for C++ code, why are you adding `__init__.py` here?

torchvision/csrc/io/decoder/__init__.py
```diff
@@ -0,0 +1,208 @@
...
+        return True
+
+
+else:
+
+    def _has_video_opt() -> bool:
+        return False
+
+
+class VideoReader:
```
prabhat00155, line 47: Isn't this already defined in [`torchvision/io/__init__.py`](https://github.com/pytorch/vision/blob/972ca657416c5896ba6e0f5fe7c0d0f3e99e1087/torchvision/io/__init__.py#L47).

torchvision/csrc/io/decoder/video.h
```diff
@@ -0,0 +1,58 @@
+#pragma once
```
prabhat00155, line 1: This code looks same as https://github.com/pytorch/vision/blob/main/torchvision/csrc/io/video/video.h.

torchvision/csrc/io/decoder/video.cpp
```diff
@@ -0,0 +1,349 @@
+#include "video.h"
```
prabhat00155, line 1: Is this a copy of https://github.com/pytorch/vision/blob/main/torchvision/csrc/io/video/video.cpp? What are we trying to do here?

bjuncek:
CLosing and sending new one as cleanup

PR #4338, opened 2021-08-31
Refactor AutoAugment to support more augmentations

datumbox:
Refactor the AutoAugment method so that more techniques can share code easier.

Moving code from #4221

cc @vfdev-5 @datumbox


fmassa (approved):
Thanks!

PR #4341, opened 2021-09-01
Explicitely store a distance value that is reused

camilodlt:
I don't see a reason to calculate the value twice for each distance. 

Knowing that this code is going to be called at every epoch and that probably there is not a compiler that optimizes this (+ also code clarity), I think is best to store the value for x and y.

### DIFF
--- a/torchvision/models/detection/_utils.py
+++ b/torchvision/models/detection/_utils.py
@@ -215,11 +215,11 @@ class BoxCoder(object):
         pred_ctr_y = dy * heights[:, None] + ctr_y[:, None]
         pred_w = torch.exp(dw) * widths[:, None]
         pred_h = torch.exp(dh) * heights[:, None]
-        
-        # Distance from center to box's corner. 
+
+        # Distance from center to box's corner.
         c_to_c_h = torch.tensor(0.5, dtype=pred_ctr_y.dtype, device=pred_h.device) * pred_h
         c_to_c_w = torch.tensor(0.5, dtype=pred_ctr_x.dtype, device=pred_w.device) * pred_w
-        
+
         pred_boxes1 = pred_ctr_x - c_to_c_w
         pred_boxes2 = pred_ctr_y - c_to_c_h
         pred_boxes3 = pred_ctr_x + c_to_c_w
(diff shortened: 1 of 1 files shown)


torchvision/models/detection/_utils.py
```diff
@@ -215,11 +215,15 @@ def decode_single(self, rel_codes, boxes):
...
         pred_h = torch.exp(dh) * heights[:, None]
-
-        pred_boxes1 = pred_ctr_x - torch.tensor(0.5, dtype=pred_ctr_x.dtype, device=pred_w.device) * pred_w
-        pred_boxes2 = pred_ctr_y - torch.tensor(0.5, dtype=pred_ctr_y.dtype, device=pred_h.device) * pred_h
-        pred_boxes3 = pred_ctr_x + torch.tensor(0.5, dtype=pred_ctr_x.dtype, device=pred_w.device) * pred_w
-        pred_boxes4 = pred_ctr_y + torch.tensor(0.5, dtype=pred_ctr_y.dtype, device=pred_h.device) * pred_h
+        
+        # Distance from center to box's corner. 
+        c_to_c_w= torch.tensor(0.5, dtype=pred_ctr_y.dtype, device=pred_h.device) * pred_h
+        c_to_c_h= torch.tensor(0.5, dtype=pred_ctr_x.dtype, device=pred_w.device) * pred_w
```
NicolasHug, line 221: Flake8 + fix

```suggestion
        c_to_c_h = torch.tensor(0.5, dtype=pred_ctr_y.dtype, device=pred_h.device) * pred_h
        c_to_c_w = torch.tensor(0.5, dtype=pred_ctr_x.dtype, device=pred_w.device) * pred_w
```

NicolasHug:
Looks like there are still some minor linting issues:

```
./torchvision/models/detection/_utils.py:218:1: W293 blank line contains whitespace
./torchvision/models/detection/_utils.py:219:48: W291 trailing whitespace
./torchvision/models/detection/_utils.py:222:1: W293 blank line contains whitespace
```

NicolasHug (approved):
@datumbox thanks for the review, for such PR that are simple and where there's no rush, it's probably best to give the contributor a chance to address the review comments first.

LGTM then

PR #4342, opened 2021-09-01
Fix Caltech101 docstring renderring

NicolasHug:
Rendering of its docstring is currently broken https://pytorch.org/vision/stable/datasets.html#torchvision.datasets.Caltech101

cc @pmeier


datumbox (approved):
stamping to unblock

pmeier (approved):
Thanks!

PR #4348, opened 2021-09-01
Adding RandAugment implementation

datumbox:
Partially implements #3817

Inspired from work started by @SamuelGabriel at #4221

cc @vfdev-5 @datumbox

### DIFF
--- a/torchvision/transforms/autoaugment.py
+++ b/torchvision/transforms/autoaugment.py
@@ -7,7 +7,7 @@ from typing import List, Tuple, Optional, Dict
 
 from . import functional as F, InterpolationMode
 
-__all__ = ["AutoAugmentPolicy", "AutoAugment"]
+__all__ = ["AutoAugmentPolicy", "AutoAugment", "RandAugment"]
 
 
 def _apply_op(img: Tensor, op_name: str, magnitude: float,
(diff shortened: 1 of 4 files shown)


torchvision/transforms/autoaugment.py
```diff
@@ -239,3 +239,87 @@ def forward(self, img: Tensor) -> Tensor:
...
+        self.num_magnitude_bins = num_magnitude_bins
+        self.interpolation = interpolation
+        self.fill = fill
+
+    def _augmentation_space(self, num_bins: int, image_size: List[int]) -> Dict[str, Tuple[Tensor, bool]]:
+        return {
+            # op_name: (magnitudes, signed)
+            "ShearX": (torch.linspace(0.0, 0.3, num_bins), True),
+            "ShearY": (torch.linspace(0.0, 0.3, num_bins), True),
+            "TranslateX": (torch.linspace(0.0, 150.0 / 331.0 * image_size[0], num_bins), True),
```
datumbox, line 278: @SamuelGabriel I noticed in your implementation you restrict the [maximum value of Translate to 14.4 pixel](https://github.com/automl/trivialaugment/blob/master/aug_lib.py#L382). I couldn't find a reference of that on the RandAugment paper. I was hoping if you could provide some background info?

torchvision/transforms/autoaugment.py
```diff
@@ -178,9 +178,9 @@ def _get_transforms(
         else:
             raise ValueError("The provided policy {} is not recognized.".format(policy))
 
-    def _get_magnitudes(self, num_bins: int, image_size: List[int]) -> Dict[str, Tuple[Tensor, bool]]:
+    def _augmentation_space(self, num_bins: int, image_size: List[int]) -> Dict[str, Tuple[Tensor, bool]]:
```
datumbox, line 181: Even thought this is a private method, I decided to use the terminology of the TrivialAugment paper as I think it describes better what we get back (combination of permitted ops and magnitudes) for the given transform.

torchvision/transforms/autoaugment.py
```diff
@@ -239,3 +239,87 @@ def forward(self, img: Tensor) -> Tensor:
...
+        magnitude (int): Magnitude for all the transformations.
+        num_magnitude_bins (int): The number of different magnitude values.
+        interpolation (InterpolationMode): Desired interpolation enum defined by
+            :class:`torchvision.transforms.InterpolationMode`. Default is ``InterpolationMode.NEAREST``.
+            If input is Tensor, only ``InterpolationMode.NEAREST``, ``InterpolationMode.BILINEAR`` are supported.
+        fill (sequence or number, optional): Pixel fill value for the area outside the transformed
+            image. If given a number, the value is used for all bands respectively.
+        """
+
+    def __init__(self, num_ops: int = 2, magnitude: int = 9, num_magnitude_bins: int = 30,
```
datumbox, line 263: - N=2, M=9 are the best ImageNet values for ResNet50 (see A.2.3 on paper).
- num_magnitude_bins=30 because the majority of the experiments on the paper used this value. Weirdly section A.2.3 mentions trying the level 31 for EfficientNet B7.

datumbox (commented):
Below I highlight a few bits of the implementation that I think is noteworthy. 

@SamuelGabriel Given your work on the area I would love it if you can provide your input once you are back.

torchvision/transforms/autoaugment.py
```diff
@@ -239,3 +239,87 @@ def forward(self, img: Tensor) -> Tensor:
...
+            img (PIL Image or Tensor): Image to be transformed.
+        Returns:
+            PIL Image or Tensor: Transformed image.
+        """
+        fill = self.fill
+        if isinstance(img, Tensor):
+            if isinstance(fill, (int, float)):
+                fill = [float(fill)] * F.get_image_num_channels(img)
+            elif fill is not None:
+                fill = [float(f) for f in fill]
```
fmassa, line 303: nit: might be worth putting this in a helper function, or push it directly to `_apply_op` maybe?

torchvision/transforms/autoaugment.py
```diff
@@ -239,3 +239,87 @@ def forward(self, img: Tensor) -> Tensor:
 
     def __repr__(self) -> str:
         return self.__class__.__name__ + '(policy={}, fill={})'.format(self.policy, self.fill)
+
+
+class RandAugment(torch.nn.Module):
```
fmassa, line 244: I wonder if it makes sense to inherit from `AutoAugment` and override only the `_augmentation_space` method?

fmassa (approved):
Looks great, thanks!

I haven't checked the bin values for the augmentation space, but the rest is good with me.

I've made a couple of minor comments, none of which are merge blocking

torchvision/transforms/autoaugment.py
datumbox: Good call, I'll add this in a helper method for now. This can move to a base class once we nail the API.

torchvision/transforms/autoaugment.py
datumbox: Yes indeed. It's a direct copy-paste. The only reason I didn't make it static or inherit from AutoAugment is because I think we haven't nailed the API of the base class yet. I was thinking of keeping only the public parts visible and make changes once we add a couple of methods.

torchvision/transforms/autoaugment.py
datumbox: The JIT is giving me headaches if I move it on ops. I'll add a TODO to remove duplicate code once we have a base class.

torchvision/transforms/autoaugment.py
SamuelGabriel: I am actually not sure what is the best strategy to follow here. Any idea? The same problem arises for AutoAugment actually. Should we ask the authors or focus only on 32x32 images or only on ImageNet? For TA it is simpler we use the same setting across datasets.

SamuelGabriel (commented):
To me it looks good, besides the mentioned problems of the augmentation space.

datumbox (commented):
@SamuelGabriel Thanks a lot for your review. I'll fix the issues you highlighted on a separate PR as this is already merged.

PR #4361, opened 2021-09-04
added serialization note to models docs

### LINKED ISSUE #2939: Add notes about serialization to models docs
## 📚 Documentation

Following https://github.com/pytorch/vision/issues/2923#issuecomment-719650811 

> it would be useful to add a note in the models documentation pointing to https://pytorch.org/docs/stable/notes/serialization.html and saying that we guarantee compatibility with serialized state-dict, but not if the model is entirely serialized.

saswatpp:
Fixes #2939

A note about backward compatibility with respect to model serialization was added to the `torchvision.models` documentation. 

Since the note is not specific to classification/segmentation tasks, hence it was added in the beginning.

### DIFF
--- a/docs/source/models.rst
+++ b/docs/source/models.rst
@@ -7,6 +7,15 @@ different tasks, including: image classification, pixelwise semantic
 segmentation, object detection, instance segmentation, person
 keypoint detection and video classification.
 
+.. note ::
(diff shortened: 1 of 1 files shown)


vfdev-5 (approved):
Thanks for the PR @saswatpp

datumbox (approved):
LGTM, thanks.

Issue #2923, opened 2020-10-28
Cannot use Inception3 model created in previous version in PyTorch 1.6+ (Vision 0.7+)


Stehsegler:
## 🐛 Bug

I am completely new to PyTorch and currently familiarizing myself with a project someone else created (and which unfortunately is not open source so I cannot share any code or exported model). With this background, I hope my report is concise enough for you to reproduce my problem.

As far as I understand, model loading in PyTorch should be backwards compatible, i.e. I can use models created in earlier versions of PyTorch in later versions. However, for me this does not seem to be the case for an InceptionV3 network created somewhere around PyTorch 0.4 when I try to run inference with it in PyTorch 1.6 and 1.7 (Torch Vision 0.7.0 and 0.8.0, respectively).

## To Reproduce

1. Load Inceptionv3 model created in earlier PyTorch version (e.g. PyTorch 0.4) in PyTorch 1.6 or higher using `torch.load()`.
1. Switch to evaluation mode using `model.eval()`.
1. Run inference: `output = model(input)`.

Result: A `ModuleAttributeError` is thrown saying `'Inception3' object has no attribute 'maxpool1'`. Stack trace:
```
Traceback (most recent call last):
  *** project-specific call hierarchy cut out ***
    output = model(input)
  File "/opt/conda/lib/python3.8/site-packages/torch/nn/modules/module.py", line 727, in _call_impl
    result = self.forward(*input, **kwargs)
  File "/opt/conda/lib/python3.8/site-packages/torchvision/models/inception.py", line 199, in forward
    x, aux = self._forward(x)
  File "/opt/conda/lib/python3.8/site-packages/torchvision/models/inception.py", line 143, in _forward
    x = self.maxpool1(x)
  File "/opt/conda/lib/python3.8/site-packages/torch/nn/modules/module.py", line 778, in __getattr__
    raise ModuleAttributeError("'{}' object has no attribute '{}'".format(
torch.nn.modules.module.ModuleAttributeError: 'Inception3' object has no attribute 'maxpool1'
```

## Expected behavior

Inference runs without errors.

## Environment

I'm using the docker image `pytorch/pytorch:1.7.0-cuda11.0-cudnn8-runtime`.

Output from the [environment collection script](https://raw.githubusercontent.com/pytorch/pytorch/master/torch/utils/collect_env.py):
```
Collecting environment information...
[...]

PR #4363, opened 2021-09-04
Fix note about model serialization BC guarantees

NicolasHug:
This is a follow up to https://github.com/pytorch/vision/pull/4361, I might be missing something but according to the original https://github.com/pytorch/vision/issues/2923#issuecomment-719650811, BC is **not** guaranteed for fully serialized models:

> we guarantee compatibility with serialized state-dict, but not if the model is entirely serialized.

Also minor change from single ticks to double backticks which I believe was the intended formatting.

### DIFF
--- a/docs/source/models.rst
+++ b/docs/source/models.rst
@@ -9,10 +9,10 @@ keypoint detection and video classification.
 
 .. note ::
     Backward compatibility is guaranteed for loading a serialized 
-    `state_dict` to the model created using old PyTorch version. 
+    ``state_dict`` to the model created using old PyTorch version. 
     On the contrary, loading entire saved models or serialized 
-    `ScriptModules` (seralized using older versions of PyTorch) 
-    will preserve the historic behaviour. Refer to the following 
+    ``ScriptModules`` (seralized using older versions of PyTorch) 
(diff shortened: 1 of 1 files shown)


datumbox (approved):
Yes, you are correct the previous PR should have read `will not`.

saswatpp:
Yes, I agree that preservation of historic behaviour is "limited" when a fully serialised model is loaded. Thanks for fixing.

NicolasHug:
Thanks for the fast reviews!

@datumbox  should it be "may not" (as currently written) or "will not"? The distinction is subtle but probably important :)

datumbox:
@NicolasHug I think your proposal is fine. No need to be nitpicking. :)

FYI the language that PyTorch typically uses for that is "it's not guaranteed to be BC". See the documentation on cross version reproducibility [here](https://pytorch.org/docs/stable/notes/randomness.html).

NicolasHug:
Ok, thanks for the details, I wanted to make sure.

I wouldn't say it's a nitpick. In one case the sentence would mean "this is never backward-compatible", and in the other case it would mean "this is maybe BC, maybe not". While the conclusion for the user will be the same (that they should not assume BC), this kind of distinction is quite important when writing docs IMO.

Issue #4360, opened 2021-09-03
Use Union in Torchscript


oke-aditya:
### 🚀 The feature

Use Union in Torchscript.

https://github.com/pytorch/pytorch/pull/64234

### Motivation, pitch

Now that Union typing support is merged. We can rewrite the workarounds which don't support Union.
E.g we can now proceed with #4323 

There maybe more places so we can track them in this Issue too.

NicolasHug:
That's great news, on top of #4323, a top priority regarding this would be to fix the annotations that are currently incorrect because we were lacking Union support in the past

oke-aditya:
I think it is supported but the docs haven't updated the change?

https://pytorch.org/docs/master/jit_language_reference.html#supported-type

frgfm:
> I think it is supported but the docs haven't updated the change?
> 
> https://pytorch.org/docs/master/jit_language_reference.html#supported-type

@oke-aditya https://github.com/pytorch/pytorch/pull/68435 should fix that :+1:

oke-aditya:
So nice of you @frgfm Francois. Merci beaucoup!!

PR #4369, opened 2021-09-06
Add typing in GeneralizedRCNNTransform

datumbox:


torchvision/models/detection/transform.py
```diff
@@ -86,10 +85,9 @@ def __init__(self, min_size, max_size, image_mean, image_std, size_divisible=32,
         self.fixed_size = fixed_size
 
     def forward(self,
-                images,       # type: List[Tensor]
-                targets=None  # type: Optional[List[Dict[str, Tensor]]]
-                ):
-        # type: (...) -> Tuple[ImageList, Optional[List[Dict[str, Tensor]]]]
+                images: List[Tensor],
+                targets: Optional[List[Dict[str, Tensor]]] = None
+                ) -> Tuple[ImageList, Optional[List[Dict[str, Tensor]]]]:
```
oke-aditya, line 90: Hey @datumbox !
We don't need to hack this anyomore!
Unoin is supported by Torchscript #4360 

We can use `Union[ImageList, Optional[List[Dict[str, Tensor]]]`

I'm currently working to use this and possibly finish #4220

oke-aditya: Aah found it https://github.com/pytorch/vision/pull/4220#discussion_r679218040 yeah 
I commented on wrong code.

datumbox: I'm not sure this is related to the problem you describe. This is not a workaround, could you clarify if your comment is still valid here?

oke-aditya: Yeah its' invalid. My bad

datumbox: no worries :) thanks for checking though

fmassa (approved):
Thanks!

PR #4372, opened 2021-09-06
Resolving tracing problem on StochasticDepth iterator

datumbox:
Fixing issue reported at https://github.com/pytorch/vision/pull/4302#issuecomment-913674132

PR #4373, opened 2021-09-06
Make StochasticDepth FX-compatible

fmassa:
Follow-up from https://github.com/pytorch/vision/pull/4372

Let's just entirely wrap `stochastic_depth`, we can consider it as an elementary operator


datumbox (approved):
LGTM, thanks!

Issue #3992, opened 2021-06-07
Re-enable or remove test_cpp_models.py?


NicolasHug:
Currently, all the tests in `test_cpp_models.py` are unconditionally skipped:

```py
@unittest.skipIf(
    sys.platform == "darwin" or True,
    "C++ models are broken on OS X at the moment, and there's a BC breakage on master; "
    "see https://github.com/pytorch/vision/issues/1191")
```

(note the `or True`).

I tried running them on my laptop and they all passed except for a few of the `mnasnet` tests that failed with `RuntimeError: Given groups=32, expected weight to be divisible by 32 at dimension 0, but got weight of size [[40, 1, 3, 3]] instead` , which is probably fairly easy to fix. BTW, I'm on OS X so the ` sys.platform == "darwin"` can probably be removed too.

Should we fix and re-activate these tests? Or perhaps they are dedundant with the ones in `test_models.py` now? The tests ran in 2+ minutes for me, so reactivating them will have a significant impact on CI usage, which is not negligible.

CC @fmassa 

cc @pmeier

fmassa:
Those tests test the C++ API of the models in torchvision. The current way we did to test them is not ideal, as we expose the C++ models to Python to compare the results of the baseline implementation in Python.

In principle we should enable the C++ tests back, but given that the C++ API of torchvision is in a state which it is not maintained (we don't enforce adding C++-equivalent models to the new Python models added), I'm tempted to not spend too much effort with it. If it's a quick fix, then great, but let's make sure we don't ship our torchvision binaries with the C++ models, and that it is only compiled in the `unittests`.

Additionally, it would be good to know who is the current maintainer of the C++ API of PyTorch

NicolasHug:
@fmassa , now that https://github.com/pytorch/vision/pull/4375 is merged and that the cpp API is deprecated, should we just remove the tests altogether? They're always skipped at the moment (and have been for a while)

NicolasHug:
I don't see this file anymore, we probably removed it

PR #4375, opened 2021-09-06
Deprecate the C++ vision::models namespace

datumbox:
We should consider dropping the C++ `vision::models` namespace for the following reasons:
1. They are not currently being maintained.
2. They represent an old attempt to support Vision models in C++. This was superseded by JIT. All of the models we propose to remove are fully jit-traceable and jit-scriptable.
3. Several of the new classification models are missing (MobileNetV3, EfficientNets etc). Segmentation and Detection models are not implemented.

This PR deprecates the entire namespace by adding warnings. It also updates the code examples to show how to integrate models using JIT.

### DIFF
--- a/examples/cpp/hello_world/main.cpp
+++ b/examples/cpp/hello_world/main.cpp
@@ -1,25 +1,44 @@
 #include <iostream>
+#include <torch/script.h>
 #include <torch/torch.h>
 #include <torchvision/vision.h>
-#include <torchvision/models/resnet.h>
 
-int main()
-{
-  auto model = vision::models::ResNet18();
-  model->eval();
+int main() {
+  torch::DeviceType device_type;
+  device_type = torch::kCPU;
+
+  torch::jit::script::Module model;
+  try {
+    std::cout << "Loading model\n";
+    // Deserialize the ScriptModule from a file using torch::jit::load().
+    model = torch::jit::load("resnet18.pt");
+    std::cout << "Model loaded\n";
+  } catch (const torch::Error& e) {
+    std::cout << "error loading the model\n";
+    return -1;
+  } catch (const std::exception& e) {
+    std::cout << "Other error: " << e.what() << "\n";
+    return -1;
+  }
+
+  // TorchScript models require a List[IValue] as input
+  std::vector<torch::jit::IValue> inputs;
 
   // Create a random input tensor and run it through the model.
-  auto in = torch::rand({1, 3, 10, 10});
-  auto out = model->forward(in);
-
-  std::cout << out.sizes();
+  inputs.push_back(torch::rand({1, 3, 10, 10}));
+  auto out = model.forward(inputs);
+  std::cout << out << "\n";
 
   if (torch::cuda::is_available()) {
     // Move model and inputs to GPU
-    model->to(torch::kCUDA);
-    auto gpu_in = in.to(torch::kCUDA);
-    auto gpu_out = model->forward(gpu_in);
+    model.to(torch::kCUDA);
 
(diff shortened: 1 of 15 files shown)


examples/cpp/hello_world/main.cpp
```diff
@@ -1,25 +1,44 @@
 #include <iostream>
+#include <torch/script.h>
```
NicolasHug, line 2: on top of updating the example (which is great) I think we should also update the C++ API part of our README: https://github.com/pytorch/vision#c-api

NicolasHug (commented):
Thanks for the PR @datumbox , I have a few remarks / questions

> They are not currently being maintained.

Somewhat related is https://github.com/pytorch/vision/issues/3992. It shouldn't take too much effort to put it back on track but it might not be needed after all

> They represent an old attempt to support Vision models in C++. This was superseded by JIT. All of the models we propose to remove are fully jit-traceable and jit-scriptable.

It seems however that relying on jit for this involves a slightly more cumbersome workflow: namely, one has to first export a model from Python so that it can then be used in C++. This means that you can the exported models also needs to be shipped as part of the binaries I guess. What do we think about this?

Also, is there a similar effort on the pytorch side? Quoting @fmassa 's https://github.com/pytorch/vision/issues/3992#issuecomment-857502031

> it would be good to know who is the current maintainer of the C++ API of PyTorch

Finally, with torchscript being currently unbundled to different other packages, will there be a different alternative for this in the future? (say torch::deploy or something else)

datumbox:
@NicolasHug Thanks for the comments. 

Unfortunately, maintaining the namespace requires significant more effort than one would think. The reason, most of the newer models are missing is because they rely on operators that are only available in Python. Moreover offering a C++ API for only the old classification models makes very little sense. Please note that this initiative is coordinated with Francisco and discussed with Greg and Brian. If you have more concerns about this, let's chat offline.

NicolasHug:
Thanks @datumbox ,

I would still be interested in having my questions above answered. Perhaps there is a document that I could read?
If I were a user relying on the current C++ API, I think these questions would be relevant to me, so maybe there's value in having them publicly as well?

datumbox:
@NicolasHug Sounds good. We can gather references related. to the above and you are welcome to lead the efforts to make them presentable publicly so that the community can benefit.

NicolasHug:
Thanks for the proposal @datumbox but since I'm not involved in this work, I don't think it should be up to me

fmassa:
Hi,

To address some points brought up by @NicolasHug comments:

> They represent an old attempt to support Vision models in C++. This was superseded by JIT. All of the models we propose to remove are fully jit-traceable and jit-scriptable.

The torchscript and C++ APIs are two different ways of getting your model into C++, and they target different users.
There are use-cases where having your full training / evaluation script in C++ is preferable (i.e., in very low-latency cases for RL), so the C++ API might be a better fit in there (and this is why we added the C++ models to begin with).

> This means that you can the exported models also needs to be shipped as part of the binaries I guess. What do we think about this?

I believe this is fine and what is done by the rest of the community. If you imagine that you have your binary that runs your application, and that this binary takes a serialized model as an argument, then shipping this extra file is very reasonable (and we don't need to recompile the app if we switch models).

> Finally, with torchscript being currently unbundled to different other packages, will there be a different alternative for this in the future? (say torch::deploy or something else)

Yes, I would expect that we will have a 3rd option to enable models to run in C++, which is similar to the torchscript solution but running entirely with Python.

NicolasHug:
Thanks you for addressing my questions @fmassa

PR #4376, opened 2021-09-07
fix fx example so that it works with MaskRCNN

alexander-soare:
The MaskRCNN example worked up to creating the model but the forward pass would have failed because of mismatching fpn/anchor generation specs. Now fixed.


fmassa (approved):
Thanks!

PR #3656, opened 2021-04-09
Remove pandas dependency for CelebA dataset

carolineechen:
Remove the runtime dependency on `pandas` for CelebA, using Python's builtin `csv` module instead.

closes #3650

PR #4377, opened 2021-09-07
fix splitting CelebA dataset + corresponding test

VassilisCN:
While working on the CelebA dataset, I noticed that when splitting the dataset (into 'train'/'valid'/'test' or take it as a whole 'all') only the target (the attributes) were coherent with the split. The input images were always the same, the only thing changing was their number. I traced the code and noticed that the mask that was created for picking the correct samples for the split was only applied on the targets (the attributes) and not the input images. Therefore I fixed the above issue by applying the mask on the input images as well. Also, I added a corresponding test function that checks if each individual split ('train', 'valid', 'test') merged together would create the same input images as if taking the whole dataset ('all'). This test Fails without the aforementioned fix. 

cc @pmeier

### DIFF
--- a/test/test_datasets.py
+++ b/test/test_datasets.py
@@ -605,7 +605,7 @@ class CelebATestCase(datasets_utils.ImageDatasetTestCase):
             with self.create_dataset(split=split) as (dataset, _):
                 merged_imgs_names += dataset.filename
 
-        assert(merged_imgs_names == all_imgs_names)
+        assert merged_imgs_names == all_imgs_names
 
 
 class VOCSegmentationTestCase(datasets_utils.ImageDatasetTestCase):
--- a/test/test_datasets.py
+++ b/test/test_datasets.py
@@ -512,7 +512,7 @@ class CelebATestCase(datasets_utils.ImageDatasetTestCase):
         return dict(num_examples=num_images_per_split[config["split"]], attr_names=attr_names)
 
     def _create_split_txt(self, root):
-        num_images_per_split = dict(train=3, valid=2, test=1)
+        num_images_per_split = dict(train=3, valid=2, test=2)
 
         data = [
             [self._SPLIT_TO_IDX[split]] for split, num_images in num_images_per_split.items() for _ in range(num_images)
--- a/test/test_datasets.py
+++ b/test/test_datasets.py
@@ -595,6 +595,18 @@ class CelebATestCase(datasets_utils.ImageDatasetTestCase):
         with self.create_dataset() as (dataset, info):
             assert tuple(dataset.attr_names) == info["attr_names"]
 
+    def test_images_names_split(self):
+        splits = ["train", "valid", "test"]
+        with self.create_dataset(split='all') as (dataset, _):
+            all_imgs_names = dataset.filename
+
+        merged_imgs_names = []
+        for split in splits:
+            with self.create_dataset(split=split) as (dataset, _):
+                merged_imgs_names += dataset.filename
+
+        assert(merged_imgs_names == all_imgs_names)
+
 
 class VOCSegmentationTestCase(datasets_utils.ImageDatasetTestCase):
     DATASET_CLASS = datasets.VOCSegmentation
--- a/test/test_datasets.py
+++ b/test/test_datasets.py
@@ -596,9 +596,8 @@ class CelebATestCase(datasets_utils.ImageDatasetTestCase):
             assert tuple(dataset.attr_names) == info["attr_names"]
 
     def test_images_names_split(self):
-        splits = ["train", "valid", "test"]
         with self.create_dataset(split='all') as (dataset, _):
-            all_imgs_names = dataset.filename
+            all_imgs_names = set(dataset.filename)
 
         merged_imgs_names = set()
         for split in ["train", "valid", "test"]:
--- a/test/test_datasets.py
+++ b/test/test_datasets.py
@@ -600,10 +600,10 @@ class CelebATestCase(datasets_utils.ImageDatasetTestCase):
         with self.create_dataset(split='all') as (dataset, _):
             all_imgs_names = dataset.filename
 
-        merged_imgs_names = []
-        for split in splits:
+        merged_imgs_names = set()
+        for split in ["train", "valid", "test"]:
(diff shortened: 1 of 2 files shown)


test/test_datasets.py
```diff
@@ -595,6 +595,18 @@ def test_attr_names(self):
...
+        splits = ["train", "valid", "test"]
+        with self.create_dataset(split='all') as (dataset, _):
+            all_imgs_names = dataset.filename
+
+        merged_imgs_names = []
+        for split in splits:
+            with self.create_dataset(split=split) as (dataset, _):
+                merged_imgs_names += dataset.filename
+
+        assert(merged_imgs_names == all_imgs_names)
```
pmeier, line 608: `assert` is a statement, so we don't need the parentheses:

```suggestion
        assert merged_imgs_names == all_imgs_names
```

test/test_datasets.py
```diff
@@ -595,6 +595,18 @@ def test_attr_names(self):
...
 
+    def test_images_names_split(self):
+        splits = ["train", "valid", "test"]
+        with self.create_dataset(split='all') as (dataset, _):
+            all_imgs_names = dataset.filename
+
+        merged_imgs_names = []
+        for split in splits:
+            with self.create_dataset(split=split) as (dataset, _):
+                merged_imgs_names += dataset.filename
```
pmeier, line 606: 1. This relies on the fact that the images of the splits are back to back, i.e. the ordering is important. Since we actually don't want to test this here, we might as well use a set to ignore the ordering.
2. Nitpick: since `splits` is only used once, I would prefer to inline it or at least move it closer to this loop


```suggestion
        merged_imgs_names = set()
        for split in ["train", "valid", "test"]:
            with self.create_dataset(split=split) as (dataset, _):
                merged_imgs_names.update(dataset.filename)
```

test/test_datasets.py
```diff
@@ -595,6 +595,18 @@ def test_attr_names(self):
         with self.create_dataset() as (dataset, info):
             assert tuple(dataset.attr_names) == info["attr_names"]
 
+    def test_images_names_split(self):
+        splits = ["train", "valid", "test"]
+        with self.create_dataset(split='all') as (dataset, _):
+            all_imgs_names = dataset.filename
```
pmeier, line 601:
```suggestion
        with self.create_dataset(split='all') as (dataset, _):
            all_imgs_names = set(dataset.filename)
```

torchvision/datasets/celeba.py
```diff
@@ -99,7 +99,12 @@ def __init__(
 
         mask = slice(None) if split_ is None else (splits.data == split_).squeeze()
 
-        self.filename = splits.index
+        if mask == slice(None):  # if split == all
+            self.filename = splits.index
+        elif len(torch.nonzero(mask)) == 1:  # check for the case of a single sample
+            self.filename = [splits.index[torch.nonzero(mask)]]
```
pmeier, line 105: How can this branch be triggered? Each split should always have more than one image right? Or is the fake data in the tests setup in a way that one split only has one image? If so, let's change the test data to make it more realistic.

pmeier (commented):
Hey @VassilisCN. Thanks for pointing this out. This is a regression from #3656. I have a few minor comments inline.

torchvision/datasets/celeba.py
VassilisCN: Yes this is affected by the fake data, but I thought the more general case would not hurt. But since the more realistic scenario is not having a single sample for any of the splits (specially since the splits are already predetermined by the creators of the dataset), we can just change the fake data for the test case to be more than 1.

pmeier: We should always adapt our tests to fit the implementation and not the other way around. I appreciate the concern for generality, but the dataset is fixed, so if there is nothing triggering this case now, there won't be anything later.

test/test_datasets.py
VassilisCN: Agreed

test/test_datasets.py
VassilisCN: Agree as to it is the better way for its purposes.

torchvision/datasets/celeba.py
VassilisCN: I removed the single sample case in the latest commit, and changed the fake data for the 'test' to be 2.

test/test_datasets.py
```diff
@@ -512,7 +512,7 @@ def inject_fake_data(self, tmpdir, config):
         return dict(num_examples=num_images_per_split[config["split"]], attr_names=attr_names)
 
     def _create_split_txt(self, root):
-        num_images_per_split = dict(train=3, valid=2, test=1)
+        num_images_per_split = dict(train=3, valid=2, test=2)
```
pmeier, line 515: Let's us a different value for each split to make sure everything works as expected.

```suggestion
        num_images_per_split = dict(train=4, valid=3, test=2)
```

torchvision/datasets/celeba.py
```diff
@@ -99,7 +99,10 @@ def __init__(
 
         mask = slice(None) if split_ is None else (splits.data == split_).squeeze()
 
-        self.filename = splits.index
+        if mask == slice(None):  # if split == all
```
pmeier, line 102:
```suggestion
        if mask == slice(None):  # if split == "all"
```

pmeier (approved):
Thanks @VassilisCN for the fix!

NicolasHug (approved):
thanks @VassilisCN and @pmeier !

mohammad-brdrn:
I've not gone through the source codes but I did an experiment in Torchvision V 0.12.0.dev20211118+cu102. It seems to me that the index for choosing the image and the annotation don't match. For annotations, I checked the 'gender'. gender[1] is for image[0], gender[2] is for image[1], ... (image[0] misses an gender annotation).

I did the experiments in google cloab using torchvision V 0.12.0.dev20211118+cu102.

I went through my own codes and didn't find any bugs. is it the case or should I check again?

pmeier:
@mohammad-brdrn I'm not sure what you are doing. "gender" is not one of the attributes that CelebA provides. Could you post a snippet with unexpected or wrong behavior?

PR #4379, opened 2021-09-07
Adding Mixup and Cutmix

### LINKED ISSUE #3817: [RFC] New Augmentation techniques in Torchvison

datumbox:
Partially resolves #3817

I'm adding Mixup and Cutmix in References instead in the Transforms to allow the investigation of the [new Transforms API](https://github.com/pmeier/torchvision-datasets-rework/) to conclude. The transforms were tested using unit-tests which can be found in [earlier commit](https://github.com/pytorch/vision/pull/4379/files/b5bf8fcce294910c56ef30022ce63be32f635cc8#diff-b91f9cf240d6656aaa67507d11139f5b8be05e9789b412544b5d515075141c2dR724-R796). Prior to moving them in the Transforms area, we should consider introducing a base-class and eliminate duplicate code between the two implementations.


torchvision/transforms/transforms.py
```diff
@@ -2036,19 +2030,19 @@ def forward(self, batch: Tensor, target: Tensor) -> Tuple[Tensor, Tensor]:
...
+        if self.mixup_alpha <= 0.0:
             use_mixup = False
         else:
-            use_mixup = self._cutmix_dist is None or torch.rand(1).item() >= self.cutmix_p
+            use_mixup = self.cutmix_alpha <= 0.0 or torch.rand(1).item() >= self.cutmix_p
 
         if use_mixup:
             # Implemented as on mixup paper, page 3.
-            lambda_param = self._mixup_dist.sample().item()
+            lambda_param = float(torch._sample_dirichlet(torch.tensor([self.mixup_alpha, self.mixup_alpha]))[0])
```
datumbox, line 2040: The above is the JIT-friendly version of the following: 
```python
torch.distributions.Beta(self.mixup_alpha, self.mixup_alpha).sample()
```

Dirichlet is the generalization of Beta for multiple variables.

torchvision/transforms/transforms.py
```diff
@@ -1953,3 +1953,126 @@ def forward(self, img):
...
+        target_flipped = target.flip(0)
+
+        if self.mixup_alpha <= 0.0:
+            use_mixup = False
+        else:
+            use_mixup = self.cutmix_alpha <= 0.0 or torch.rand(1).item() >= self.cutmix_p
+
+        if use_mixup:
+            # Implemented as on mixup paper, page 3.
+            lambda_param = float(torch._sample_dirichlet(torch.tensor([self.mixup_alpha, self.mixup_alpha]))[0])
```
datumbox, line 2040: The above is the JIT-friendly version of the following: 
```python
torch.distributions.Beta(self.mixup_alpha, self.mixup_alpha).sample()
```

Dirichlet is the generalization of Beta for multiple variables.

torchvision/transforms/transforms.py
```diff
@@ -1953,3 +1953,126 @@ def forward(self, img):
...
+            Set to 0.0 to turn off. Default value is 0.0.
+        label_smoothing (float): the amount of smoothing using when one-hot encoding.
+            Set to 0.0 to turn off. Default value is 0.0.
+        inplace (bool): boolean to make this transform inplace. Default set to False.
+    """
+
+    def __init__(self, num_classes: int,
+                 p: float = 1.0, mixup_alpha: float = 1.0,
+                 cutmix_p: float = 0.5, cutmix_alpha: float = 0.0,
+                 label_smoothing: float = 0.0, inplace: bool = False) -> None:
```
datumbox, line 1982: I understand that the choice of offering `inplace` support won't make everyone happy. The reason I decided to support it is:
- It is more performant in terms of memory and speed.
- Most the tensor-only transforms on torchvision typically support `inplace` operations (see `Normalize` and `RandomErasing`). 
- The ClassyVision version of this method [supported](https://github.com/facebookresearch/ClassyVision/blob/8e3d7587e4102cf308ee77498fa0ff395058bc1b/classy_vision/dataset/transforms/mixup.py#L284) inplace operations and thus to ensure we are equally performant we should support it as well.

torchvision/transforms/transforms.py
```diff
@@ -1953,3 +1953,126 @@ def forward(self, img):
...
+            batch = batch.clone()
+            # target = target.clone()
+
+        target = self._smooth_one_hot(target)
+        if torch.rand(1).item() >= self.p:
+            return batch, target
+
+        # It's faster to flip the batch instead of shuffling it to create image pairs
+        batch_flipped = batch.flip(0)
+        target_flipped = target.flip(0)
```
datumbox, line 2031: This is a widely adopted optimization [[1](https://github.com/rwightman/pytorch-image-models/blob/a2727c1bf78ba0d7b5727f5f95e37fb7f8866b1f/timm/data/mixup.py#L26), [2](https://github.com/facebookresearch/ClassyVision/blob/8e3d7587e4102cf308ee77498fa0ff395058bc1b/classy_vision/dataset/transforms/mixup.py#L34)] which exploits the fact that the training batches are already randomized. Thus creating pairs by "folding" the batch in half instead of shuffling it is much faster. 

This introduces a limitation for supporting batches of odd size (which is usually avoided anyway as it's slower than when a power of two). There are workarounds but none of them produces as clean code as this. We could also allow odd batch sizes with the only caveat that the middle record will be "combined with itself" and thus the transform has no effect. In that case we can just throw a warning to the user.

torchvision/transforms/transforms.py
```diff
@@ -1953,3 +1953,126 @@ def forward(self, img):
...
+            r_y = torch.randint(H, (1,))
+
+            r = 0.5 * math.sqrt(1.0 - lambda_param)
+            r_w_half = int(r * W)
+            r_h_half = int(r * H)
+
+            x1 = int(torch.clamp(r_x - r_w_half, min=0))
+            y1 = int(torch.clamp(r_y - r_h_half, min=0))
+            x2 = int(torch.clamp(r_x + r_w_half, max=W))
+            y2 = int(torch.clamp(r_y + r_h_half, max=H))
```
datumbox, line 2058: All the int/float casts in this section are necessary to please JIT.

test/test_transforms_tensor.py
```diff
@@ -715,3 +718,78 @@ def test_gaussian_blur(device, meth_kwargs):
...
+    for test_file in [("encode_jpeg", "grace_hopper_517x606.jpg"), ("fakedata", "logos", "rgb_pytorch.png")]:
+        fullpath = (os.path.dirname(os.path.abspath(__file__)), 'assets') + test_file
+        img = read_image(get_file_path_2(*fullpath))
+        images.append(F.resize(img, [224, 224]))
+    dataset = TensorDataset(torch.stack(images).to(torch.float32), torch.tensor([0, 1]))
+
+    # Use mixup in the collate
+    mixup = T.RandomMixupCutmix(2, cutmix_alpha=1.0, mixup_alpha=1.0, label_smoothing=0.1)
+    dataloader = DataLoader(dataset, batch_size=2,
+                            collate_fn=lambda batch: mixup(*(torch.stack(x) for x in zip(*batch))))
```
datumbox, line 784: Might be an overkill to use a DataLoader here, but the intention is to use the transform the same way we will suggest to the users to use it.

datumbox (commented):
Highlighting some "interesting" parts below to provide context. The PR is not yet ready.

torchvision/transforms/transforms.py
```diff
@@ -1953,3 +1953,126 @@ def forward(self, img):
 
     def __repr__(self):
         return self.__class__.__name__ + '(p={})'.format(self.p)
+
+
+class RandomMixupCutmix(torch.nn.Module):
```
fmassa, line 1958: High-level comment: should we split this function in two, doing `MixUp` and `CutMix`?

torchvision/transforms/transforms.py
```diff
@@ -1953,3 +1953,126 @@ def forward(self, img):
...
+        self.cutmix_alpha = cutmix_alpha
+        self.label_smoothing = label_smoothing
+        self.inplace = inplace
+
+    def _smooth_one_hot(self, target: Tensor) -> Tensor:
+        N = target.shape[0]
+        device = target.device
+        v = torch.full(size=(N, 1), fill_value=1 - self.label_smoothing, device=device)
+        return torch.full(size=(N, self.num_classes), fill_value=self.label_smoothing / self.num_classes,
+                          device=device).scatter_add_(1, target.unsqueeze(1), v)
```
fmassa, line 2000: Do we need to have this function be label-smoothing aware? I wonder if label smoothing in PyTorch can natively support this use-case https://github.com/pytorch/pytorch/pull/63122/files#diff-0b104a4612220d611b2c5d76272c28ddb21efbd8f9b9ae29d7f5b6c9e2a9be1fR480

torchvision/transforms/transforms.py
fmassa: Another option as well is instead of calling `flip` to just call `roll` -- this way the center element will not be matched with itself.

torchvision/transforms/transforms.py
fmassa: I'm still a bit skeptical about adding this in-place operation, as most of the operators in here are actually not performing any in-place operation, so the gains of having an inplace flag might not be that large. But we might need to benchmark to be able to know for sure

fmassa (commented):
Thanks a ton for the PR!

Adding a few high-level comments, let me know what you think

torchvision/transforms/transforms.py
datumbox: It's an option but there are caveats:
1. Mixup and Cutmix are not compatible and must not be combined one after the other. 
2. It's very common to switch between them randomly and many recipes use this. Separating them would require us to find another way to randomize between them. 
2. ClassyVision's implementation has them together so ours wont be a replacement after its deprecation.

torchvision/transforms/transforms.py
datumbox: Yes we can. It's mathematically equivalent and it can help us decouple the two which is great.

From my original discussions with @jbschlosser, I understood that his plan was to create the `label_smoothing` argument only for the simple case where the inputs are labels and not probabilities. But then it seems after they changed plan on the implementations, they ended adding support also when the provided vectors are probabilities. This is great and simplifies things.

torchvision/transforms/transforms.py
datumbox: I don't see why we can't support an in-place operation if one can be implemented correctly. For now I plan to move the entire implementation on the references script so that we have time to think this through before we commit to the API, so we don't have to solve this now. But overall I think inplace is a valid optimization and provided it can be implemented properly, there is no reason not do get the extra speed improvements.

torchvision/transforms/transforms.py
datumbox: This is definitely an option. The only possible caveat is the behaviour wont be aligned with the one of Classy's but to be frank this is not the kind of change that is likely to cause real issues.

torchvision/transforms/transforms.py
datumbox: @neerajprad FYI this where the JIT issue was that we discussed.

torchvision/transforms/transforms.py
datumbox: As discussed offline with @fmassa we are going to review the use of `inplace` before moving this from references to transforms.

torchvision/transforms/transforms.py
datumbox: As discussed with @fmassa offline, we decided to separate these two and introduce another transform that switches between the 2 randomly.

datumbox (commented):
I took care of splitting the methods. Below I highlight other interesting bits that were not there on the previous review.

torchvision/transforms/transforms.py
```diff
@@ -1953,3 +1965,169 @@ def forward(self, img):
...
+        elif not batch.is_floating_point():
+            raise TypeError('Batch dtype should be a float tensor. Got {}.'.format(batch.dtype))
+        elif target.dtype != torch.int64:
+            raise TypeError("Target dtype should be torch.int64. Got {}".format(target.dtype))
+
+        if not self.inplace:
+            batch = batch.clone()
+            # target = target.clone()
+
+        target = torch.nn.functional.one_hot(target, num_classes=self.num_classes).to(dtype=torch.float32)
```
fmassa, line 2018: This is fine for now, but for the future we might want to make this a "safe_one_hot` if the tensor is already one-hot encoded (for example if doing distillation from a teacher network, the targets are the output of the teacher model)

test/test_transforms.py
```diff
@@ -1311,7 +1311,8 @@ def test_random_choice():
             transforms.Resize(15),
             transforms.Resize(20),
             transforms.CenterCrop(10)
-        ]
+        ],
+        [1 / 3, 1 / 3, 1 / 3]
```
fmassa, line 1315: Is this modification necessary? I thought that the default argument (`None`) should make things compatible?

references/classification/train.py
```diff
@@ -165,10 +166,21 @@ def main(args):
     train_dir = os.path.join(args.data_path, 'train')
     val_dir = os.path.join(args.data_path, 'val')
     dataset, dataset_test, train_sampler, test_sampler = load_data(train_dir, val_dir, args)
+
+    collate_fn = None
+    num_classes = len(dataset.classes)
```
fmassa, line 171: not for now, but this exposes a limitation of our current datasets, which is that we don't consistently enforce a way of querying the number of classes in a dataset. The dataset refactoring work from @pmeier will address this

torchvision/transforms/transforms.py
```diff
@@ -515,9 +516,20 @@ def __call__(self, img):
...
-        return t(img)
+    def __init__(self, transforms, p=None):
+        super().__init__(transforms)
+        if p is not None and not isinstance(p, Sequence):
+            raise TypeError("Argument transforms should be a sequence")
+        self.p = p
+
+    def __call__(self, *args):
+        t = random.choices(self.transforms, weights=self.p)[0]
+        return t(*args)
```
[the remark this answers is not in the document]
fmassa: This is good, thanks!

fmassa (approved):
LGTM, thanks!

I've made a few minor comments, none of which are merge-blocking.

test/test_transforms.py
datumbox: Yes it's not necessary. It's just to show how this would work if we wanted to pass probabilities. If `None` is passed (which is the default argument) it assumes automatically equal probs.

test/test_transforms_tensor.py
```diff
@@ -715,3 +719,78 @@ def test_gaussian_blur(device, meth_kwargs):
...
+        t(torch.randint(256, (32, 3, 60, 60), dtype=torch.uint8), torch.randint(10, (32, )))
+    with pytest.raises(TypeError, match="Target dtype should be torch.int64."):
+        t(torch.rand(32, 3, 60, 60), torch.randint(10, (32, ), dtype=torch.int32))
+
+
+@pytest.mark.parametrize('device', cpu_and_gpu())
+@pytest.mark.parametrize('transform, expected', [
+    (T.RandomMixup, [60.77401351928711, 0.5151033997535706]),
+    (T.RandomCutmix, [70.13909912109375, 0.525851309299469])
+])
```
datumbox, line 770: @NicolasHug Any concerns of passing both the method and the expected value like this?

torchvision/transforms/transforms.py
datumbox: I was indeed considering adding support for more types of targets later on in follow up PRs, but thinking of it again, there is very little reason not to do a simple check and do this safely as you suggest. So I'm adding this now and we can further extend it on the future.

test/test_transforms_tensor.py
NicolasHug: Sorry I missed your comment
No concern at all, this is a very common and useful use of the parametrization!

references/classification/train.py
pmeier: With #4432, you will be able to do 

```python
info = torchvision.datasets.info(name)
info.categories
```

where `categories` is a list of strings in which the index corresponds to the label.

references/classification/train.py
```diff
@@ -273,6 +285,8 @@ def get_args_parser(add_help=True):
     parser.add_argument('--label-smoothing', default=0.0, type=float,
                         help='label smoothing (default: 0.0)',
                         dest='label_smoothing')
+    parser.add_argument('--mixup-alpha', default=0.0, type=float, help='mixup alpha (default: 0.0)')
+    parser.add_argument('--cutmix-alpha', default=0.0, type=float, help='cutmix alpha (default: 0.0)')
```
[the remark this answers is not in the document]
xiaohu2015: @datumbox nn.CrossEntropyLoss dose not work when you use mixup or cutmix, because the traget  shape is (N, K), rather (N,)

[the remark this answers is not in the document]
datumbox: I think this was added at https://github.com/pytorch/pytorch/pull/63122

This should be available on the latest stable version of pytorch. See [doc](https://pytorch.org/docs/stable/generated/torch.nn.CrossEntropyLoss.html).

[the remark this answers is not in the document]
xiaohu2015: @datumbox ok, thanks.

Issue #4346, opened 2021-09-01
Add Exponential Moving Average (EMA) support on our classification reference script


datumbox:
### 🚀 The feature

Add optional EMA support in our classification reference scripts. 

The solution needs to take care of:
- The EMA creation: We should use the `torch.optim.swa_utils.AveragedModel` util to build this similarly to how it's described [in the documentation](https://pytorch.org/docs/stable/optim.html#stochastic-weight-averaging).
- Checkpointing: Ideally store the weights of the ema model in a `"model_ema"` key on the `state_dict`
- Update the training loops: make the necessary calls to update the weights, provide stats on validation dataset etc.

The reference script should updated in a BC way to accept a decay param for the EMA. The default value of the decay should be 0, to indicate that the EMA should be turned off. The solution should also explore ways to keep the implementation memory efficient and avoid exhausting the GPU memory.

### Motivation, pitch

Most of SOTA models use EMA to get a few extra accuracy points for free. TorchVision should include how to do this on our reference scripts in order to help users build better models.

cc @datumbox

PR #4381, opened 2021-09-07
Added Exponential Moving Average support to classification reference script

prabhat00155:
Resolves https://github.com/pytorch/vision/issues/4346.

cc @datumbox

### DIFF
--- a/references/classification/train.py
+++ b/references/classification/train.py
@@ -205,7 +205,7 @@ def main(args):
 
     model_ema = None
     if args.model_ema:
-        model_ema = utils.ExponentialMovingAverage(model, device=device, decay=args.model_ema_decay)
+        model_ema = utils.ExponentialMovingAverage(model_without_ddp, device=device, decay=args.model_ema_decay)
 
     if args.resume:
         checkpoint = torch.load(args.resume, map_location='cpu')
--- a/references/classification/train.py
+++ b/references/classification/train.py
@@ -17,7 +17,8 @@ except ImportError:
     amp = None
 
 
-def train_one_epoch(model, criterion, optimizer, data_loader, device, epoch, print_freq, apex=False):
+def train_one_epoch(model, criterion, optimizer, data_loader, device, epoch,
+                    print_freq, apex=False, model_ema=None):
     model.train()
     metric_logger = utils.MetricLogger(delimiter="  ")
     metric_logger.add_meter('lr', utils.SmoothedValue(window_size=1, fmt='{value}'))
@@ -45,11 +46,14 @@ def train_one_epoch(model, criterion, optimizer, data_loader, device, epoch, pri
         metric_logger.meters['acc5'].update(acc5.item(), n=batch_size)
         metric_logger.meters['img/s'].update(batch_size / (time.time() - start_time))
 
+    if model_ema:
+        model_ema.update_parameters(model)
 
-def evaluate(model, criterion, data_loader, device, print_freq=100):
+
+def evaluate(model, criterion, data_loader, device, print_freq=100, log_suffix=''):
     model.eval()
     metric_logger = utils.MetricLogger(delimiter="  ")
-    header = 'Test:'
+    header = f'Test: {log_suffix}'
     with torch.no_grad():
         for image, target in metric_logger.log_every(data_loader, print_freq, header):
             image = image.to(device, non_blocking=True)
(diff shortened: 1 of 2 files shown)


references/classification/train.py
```diff
@@ -306,6 +320,12 @@ def get_args_parser(add_help=True):
     parser.add_argument('--world-size', default=1, type=int,
                         help='number of distributed processes')
     parser.add_argument('--dist-url', default='env://', help='url used to set up distributed training')
+    parser.add_argument(
+        '--model-ema', action='store_true',
```
datumbox, line 324: Is there a reason why you introduce an extra param instead of setting the default of `--model-ema-decay` to 0.0? In that case your EMA will be as if you run the normal validation and thus you could omit it.

references/classification/train.py
```diff
@@ -199,12 +203,18 @@ def main(args):
         model = torch.nn.parallel.DistributedDataParallel(model, device_ids=[args.gpu])
         model_without_ddp = model.module
 
+    model_ema = None
+    if args.model_ema:
+        model_ema = utils.ExponentialMovingAverage(model, device=device, decay=args.model_ema_decay)
```
datumbox, line 208: We might need to pass the `model_without_ddp` version of the model here.

datumbox (commented):
Thanks @prabhat00155, overall it looks good. I left a couple of comments for your consideration. Let me know what you think.

Also do you plan to investigate how we can avoid loading two models on the GPU?

references/classification/utils.py
```diff
@@ -161,6 +161,25 @@ def log_every(self, iterable, print_freq, header=None):
...
+    is used to compute the EMA.
+    """
+    def __init__(self, model, decay, device='cpu', name='ExponentialMovingAverage'):
+        ema_avg = (lambda avg_model_param, model_param, num_averaged:
+                   decay * avg_model_param + (1 - decay) * model_param)
+        super().__init__(model, device, ema_avg)
+        self._name = name
+
+    @property
+    def name(self):
```
NicolasHug, line 178: I'm just wondering, what is the reason for having a `name` attribute, and why do we need it to be a `property`?
IDK if this is relevant but we can access an object's class name with `obj.__class__.__name__`

NicolasHug (commented):
Thanks @prabhat00155 , I made some minor comments / questions

references/classification/utils.py
```diff
@@ -161,6 +161,25 @@ def log_every(self, iterable, print_freq, header=None):
         print('{} Total time: {}'.format(header, total_time_str))
 
 
+class ExponentialMovingAverage(torch.optim.swa_utils.AveragedModel):
+    """Maintains moving averages of model parameters using an
```
NicolasHug, line 165: Doc nits that don't really matter since we don't render the docs, but good to fix if we ever move this in the public API. Also good to know for future PRs:

- the first sentence should be in one line according to pep8. When this isn't respected, sphinx can create rendering issue (typically when displaying the summary of a function / class)
- since sphinx relies on rst format (not markdown like in github), the single backticks should be double backticks.
- I think there's a missing `_` at the end of the link below, although I haven't checked

references/classification/train.py
NicolasHug: I don't have a strong opinion on this, but I cant think of 2 reasons to have both --model-ema-decay and  the --model-ema flag as currently done:

- the flag allows for --model-ema-decay to have a good default of 0.99 which hopefully should work fine for most models (that might actually not be true in practice though, IDK). Without this, users would have to come up with a value themselves.
- it may not be obvious for inexperienced users that `ema_decay == 0 => no EMA`. The flag helps disambiguate whether EMA is used at all

Again, no strong opinion. Both of these could actually be solved with a few sentences in the parameter description.

prabhat00155: I guess @NicolasHug explained it well. Having `--model-ema` allows us to have a good default which users may choose to use. It also makes it simpler to use(not having to worry about choosing an ema_decay in your first run).

prabhat00155:
> Thanks @prabhat00155 , I made some minor comments / questions
> 
> Also, it'd be good to validate the new code somehow. We don't really have tests for the references but perhaps it would be relevant to report the result of a model relying on the --model_avg option to compare to the baseline?

Thanks @NicolasHug! I ran tests locally(on CPU) on a toy dataset and on aws cluster for 5 epochs. I will run it to completion to verify the results before merging this PR.

prabhat00155:
> Thanks @prabhat00155, overall it looks good. I left a couple of comments for your consideration. Let me know what you think.
> 
> Also do you plan to investigate how we can avoid loading two models on the GPU?

Thanks @datumbox! Yes, I was thinking of doing that in a follow-up PR.

datumbox:
@prabhat00155 sounds good to me. 

The only thing worth addressing here is to pass the non parallel model in ema (see my earlier comment). I believe what you have now will fail to handle properly the checkpoints on a multi gpu setup. Worth confirming by doing two epochs on AWS and resuming from checkpoint. If it works you can leave as is.

Everything else are nits that can be done later on separate PRs.

references/classification/utils.py
prabhat00155: Yeah, I think it's unnecessary, thanks!

references/classification/utils.py
prabhat00155: Thanks, updated them.

datumbox (approved):
@prabhat00155 As we discussed offline, this is a great contribution so let's unblock the merge and push further investigations on follow up PRs.

references/classification/train.py
prabhat00155: Updated it.

prabhat00155:
> @prabhat00155 As we discussed offline, this is a great contribution so let's unblock the merge and push further investigations on follow up PRs.

Makes sense, thanks!

Issue #3430, opened 2021-02-23
Paper behind Keypoint R-CNN implementation in torchvision


FraPochetti:
## 📚 Documentation

Hi team,
I can't seem to find the paper behind the [Keypoint R-CNN](https://pytorch.org/vision/0.8/models.html#keypoint-r-cnn) implementation in `torchvision`.
I'd like to better understand and study the architecture and I was looking for some support in the literature, please.
Thanks

fmassa:
Hi,

Sorry for missing this. The paper is "Mask R-CNN" https://arxiv.org/abs/1703.06870 , see Section 5.

@prabhat00155 can you add an entry to the detection models with the paper references?

PR #4383, opened 2021-09-08
Added paper references to detection models

prabhat00155:
Discussion: https://github.com/pytorch/vision/issues/3430#issuecomment-915038902

### DIFF
--- a/torchvision/models/detection/faster_rcnn.py
+++ b/torchvision/models/detection/faster_rcnn.py
@@ -300,6 +300,8 @@ def fasterrcnn_resnet50_fpn(pretrained=False, progress=True,
     """
     Constructs a Faster R-CNN model with a ResNet-50-FPN backbone.
 
+    Reference: `"Faster R-CNN: Towards Real-Time Object Detection with Region Proposal Networks" <https://arxiv.org/abs/1506.01497>`_.
+
     The input to the model is expected to be a list of tensors, each of shape ``[C, H, W]``, one for each
     image, and should be in ``0-1`` range. Different images can have different sizes.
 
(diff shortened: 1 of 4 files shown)


NicolasHug (approved):
Linter is failing with the error below, but other than that LGTM, thanks @prabhat00155 

./torchvision/models/detection/faster_rcnn.py:303:121: E501 line too long (134 > 120 characters)

torchvision/models/detection/faster_rcnn.py
```diff
@@ -300,6 +300,8 @@ def fasterrcnn_resnet50_fpn(pretrained=False, progress=True,
     """
     Constructs a Faster R-CNN model with a ResNet-50-FPN backbone.
 
+    Reference: `"Faster R-CNN: Towards Real-Time Object Detection with Region Proposal Networks" <https://arxiv.org/abs/1506.01497>`_. # noqa
```
NicolasHug, line 303: instead of adding `# noqa` we should just break this line to be shorter, it won't affect the html rendering

prabhat00155:
> Linter is failing with the error below, but other than that LGTM, thanks @prabhat00155
> 
> ./torchvision/models/detection/faster_rcnn.py:303:121: E501 line too long (134 > 120 characters)

Thanks @NicolasHug, yeah ignoring linter on that line by adding # noqa as it's a hyperlink.

torchvision/models/detection/faster_rcnn.py
prabhat00155: Oh, will do that then. Thanks!

PR #4384, opened 2021-09-08
add `ufmt` as code formatter

pmeier:
Supersedes #4178. Instead of using `isort` and `black` individually, [`ufmt`](https://ufmt.omnilib.dev/en/stable/) uses [`usort`](https://usort.readthedocs.io/en/stable) and `black` in an atomic way. 

To keep this PR reviewable, I'm only going to pull the trigger on all formatting changes if the PR is approved otherwise.

This PR is blocked for now until `ufmt>=1.3` is released.


pyproject.toml
```diff
@@ -0,0 +1,16 @@
+[tool.usort]
+
+[tool.usort.known]
+
+first_party = ["torch", "torchvision"]
```
NicolasHug, line 5: If I understand the "first-party" feature correctly, this means the `torch` imports will be done last, right?

For torchvision it's not an issue but for torch we actually want it to be the first import, because `import torch` does a bunch of stuff in the background like resolving dynamic links and other magic stuff

pyproject.toml
```diff
@@ -0,0 +1,16 @@
...
+[tool.black]
+
+line-length = 120
+target-version = ["py36"]
+
+[tool.ufmt]
+
+excludes = [
+    "gallery",
+]
```
NicolasHug, line 16: Nit: missing new line here and in setup.cfg as well

torchvision/ops/poolers.py
```diff
@@ -2,7 +2,7 @@
 from torch import nn, Tensor
 
 import torchvision
-from torchvision.ops import roi_align
+from .roi_align import roi_align
```
NicolasHug, line 5: just curious, was this flagged somehow?

pyproject.toml
```diff
@@ -0,0 +1,16 @@
+[tool.usort]
```
NicolasHug, line 1: Do we need this empty section?

NicolasHug (commented):
Sweet, thanks @pmeier ! I made a quick pass, the strategy sounds good.

Do we know when the ufmt release will be out?

torchvision/ops/poolers.py
pmeier: See https://github.com/pytorch/vision/pull/4178#discussion_r669382906. In short: we relied on a specific order of the imports due to a naming conflict. Without this fix we need to exclude these imports from being sorted.

pyproject.toml
pmeier: > If I understand the "first-party" feature correctly, this means the torch imports will be done last, right?

What this does is move `torch` into the first party category. By default only local packages, e.g. `torchvision`, go into that block. Have a look at the [sorting documentation](https://usort.readthedocs.io/en/latest/guide.html#sorting) for an example of the blocks.

> `import torch` does a bunch of stuff in the background like resolving dynamic links and other magic stuff

Does that mean we need to always import `torch` before other PyTorch derivatives like `torchvision` or does it even need to be imported before all other third party packages?

pyproject.toml
```diff
@@ -0,0 +1,31 @@
...
+
+categories = [
+    "future",
+    "standard_library",
+    "third_party",
+    "pytorch",
+    "first_party",
```
pmeier, line 8: I've split the imports a little further. This way `torch` will always get imported before we import `torchvision`, but after the standard library and other third-party packages. If we later add dependencies to other PyTorch derivatives, we can simply add another section for them to make sure they are also imported after `torch`.

pyproject.toml
```diff
@@ -0,0 +1,31 @@
...
+
+[tool.usort.known]
+
+pytorch = ["torch"]
+first_party = ["torchvision"]
+local_utils = [
+    "_utils_internal",
+    "common_utils",
+    "dataset_utils",
+]
```
pmeier, line 20: These are local utilities in our `test/` folder. Since its not a package, i.e. `tests/__init__.py` is missing, `usort` does not pick up on the fact that these are local dependencies and sorts them into the third party category.

setup.cfg
```diff
@@ -9,7 +9,13 @@ max-line-length = 120
...
 [flake8]
 max-line-length = 120
-ignore = F401,E402,F403,W503,W504,F821
+ignore = E203, E402, W503, W504, F821
+per-file-ignores =
+    __init__.py: F401, F403, F405
+    ./hubconf.py: F401
+    torchvision/models/mobilenet.py: F401, F403
+    torchvision/models/quantization/mobilenet.py: F401, F403
+    test/smoke_test.py: F401
```
pmeier, line 18: See https://github.com/pytorch/vision/pull/4178#discussion_r669383913 for details on this change.

pmeier (commented):
The lint workflow is failing, because the version of `ufmt` that we need. Otherwise this should be good to go. 

@NicolasHug: in #4178 we talked about adding pre-commit hooks for this. I have a PR up that adds this to `ufmt`: omnilib/ufmt#15. Since they are not necessary to run `ufmt`, IMO we can add them in the future.

pyproject.toml
NicolasHug: Let's add this comment in the file to provide context for future us :)

Also to avoid conflicting it with other utils, can we change the name to `local_test_utils`? Alternatively we could just add these into the "first-party" category?

torchvision/ops/poolers.py
NicolasHug: There might have been a reason for that initially, but I'm wondering why we even use absolute imports within the `torchvision` package. Every torchvision improt should be relative there. Perhaps @fmassa knows?

(anyway, this is for another PR, let's leave it as is :)  )

pyproject.toml
NicolasHug: > Does that mean we need to always import torch before other PyTorch derivatives like torchvision or does it even need to be imported before all other third party packages?

I think importing torch before torchvision (and other torchxyz) stuff should be enough. But who knows what `import torch` does... I've seen some weird stuff going on. In anycase this should be enough, if it's not, tests will hopefully fail to let us know

NicolasHug (commented):
Thanks @pmeier 

> @NicolasHug: in #4178 we talked about adding pre-commit hooks for this. I have a PR up that adds this to ufmt: omnilib/ufmt#15. Since they are not necessary to run ufmt, IMO we can add them in the future.

I'm a bit confused, do we need pre-commit hooks in the ufmt repo in order to have pre-commit hooks in torchvision?

pyproject.toml
pmeier: > Alternatively we could just add these into the "first-party" category?

We could, but if you don't have a strong opinion about this, I would leave them in an extra category. IMO, first party and local should be separated, because the latter indicates that this cannot be import the "normal" way. In fact, `isort` has a section for local imports where all the relative imports go. Unfortunately, `usort` does not have that. I've asked about it in facebookexperimental/usort#30.

pmeier:
> I'm a bit confused, do we need pre-commit hooks in the ufmt repo in order to have pre-commit hooks in torchvision?

Just to be sure: we are talking about hooks added by the [`pre-commit` framework](https://pre-commit.com/) to add these hooks and not provide a homebrew solution, right?

If yes, we don't need `pre-commit` hooks in the `ufmt` config, but a configuration telling `pre-commit` how to use the `ufmt` repository. The confusing thing about this is probably the file names:
- `.pre-commit-config.yaml` defines what hooks should be run before every commit (this is the file we need)
- `.pre-commit-hooks.yaml` defines what hooks the repository provides in case it is used by `pre-commit` (this is the file `ufmt` needs)

NicolasHug:
Oh OK, thanks. I didn't know the "upstream" repo (ufmt) had to be configured as well for commit hooks to work.

I feel like we might want to wait for commit hooks to be available though. It will probably make the transition smoother for most contributors / developers, and we will just be able to run `pre-commit run --all-files` to make sure to avoid any potential future change

pmeier:
@NicolasHug I've added support for `pre-commit` hooks in #4387. If that is landed and `ufmt` supports being used as one, we only need to add a few lines to the `.pre-commit-config.yaml` file.

pmeier:
Blocked by omnilib/ufmt#13 and omnilib/ufmt#15. After these PRs are in we also need to wait for the next release, which I'm hoping will be quick, since we added some important functionality.

pyproject.toml
```diff
@@ -0,0 +1,36 @@
...
+    "first_party",
+]
+
+# This flag is needed to group torchvision into 'pytorch_derivatives' rather than 'first_party'
+first_party_detection = false
+
+[tool.usort.known]
+
+pytorch = ["torch"]
+pytorch_derivatives = ["torchvision"]
```
pmeier, line 18: This might seem excessive now, but we will probably depend on `torchdata` (not yet released) in the near future.

NicolasHug: Do we need a new category name for those? It seems that we could add `torchdata` to what used to be `first_party`?

I feel like having `first_party` as the internal test utils below doesn't really fit the intended definition of the "first_party" category.

pmeier: It's a matter of style. By default `first_party` groups the auto-detected package including all relative imports. I like to separate relative from absolute imports, because the former usually indicates that something is either not publicly exposed or at a different point. `isort` has a separate category for `local`, whereas `usort` has not. 

I've asked about this in facebookexperimental/usort#30 and the current implementation was the suggested workaround.

1. Do we want to spit absolute and relative imports? If not, your suggestion is the way to go.
2. Can we live with this workaround or should we send a patch to `usort`?

NicolasHug: > Do we want to spit absolute and relative imports?

yes, I would think so. But I would assume that the relative imports would always be separate from the absolute ones, even within the "first_party" group?
So IIUC the only difference will be that there will be a blank line between these, which IMHO isn't necessary

pmeier:
The PRs I sent to `ufmt` were landed so this PR is almost ready to go. As soon as `ufmt` has its next release (@jreese is there a timeline for that?) I'll push the changes. I have them locally by running `ufmt`'s `main` branch as reference. 

Since this PR will touch (almost?) any `*.py` file in our repository, we will probably get merge conflicts for every open PR. I'll comment an upgrade guide, so we can point contributors here.

pmeier:
If you are reading this, you were probably prompted to fix the code format of your PR. Here is what you do:

```shell
$ git checkout main
$ git pull
$ git checkout $MY_FEATURE_BRANCH
$ git merge -X ours main
$ pre-commit run --all-files
$ git commit -am "fix code format"
$ git push
```

If you don't have `pre-commit` installed, you can install it with `pip install pre-commit` or `conda install -c conda-forge pre-commit`. To avoid running this manually in the future, you can `pre-commit install` some hooks that will be run everytime you `git commit` something in this repository.

jreese:
I published a 1.3.0 release a few minutes ago. Cheers

pmeier:
@NicolasHug I've just realized that we also use local utilities in our references, e.g.

https://github.com/pytorch/vision/blob/85982ac695e78af80bf59cd9c855e1729b7376f5/references/classification/train.py#L10-L16

I think we should broaden our `local_test_utils` to `local_utils` and include all of them. Otherwise they will be sorted into the `third_party` category, which I feel is especially confusing for the reference scripts.

pmeier:
The next commit will apply the code formatting to the whole repo. Due to the sheer amount of changes, this PR will become "unreviewable". If you still want to do it, here is a list of files, that were manually edited:

- [`.circleci/config.yml.in`](https://github.com/pmeier/vision/blob/ufmt/.circleci/config.yml.in)
- [`.github/workflows/tests-schedule.yml`](https://github.com/pmeier/vision/blob/ufmt/.github/workflows/tests-schedule.yml)
- [`.pre-commit-config.yaml`](https://github.com/pmeier/vision/blob/ufmt/.pre-commit-config.yaml)
- [`pyproject.toml`](https://github.com/pmeier/vision/blob/ufmt/pyproject.toml)
- [`setup.cfg`](https://github.com/pmeier/vision/blob/ufmt/setup.cfg)
- [`torchvision/ops/poolers.py`](https://github.com/pmeier/vision/blob/ufmt/torchvision/ops/poolers.py)

datumbox:
@pmeier shall we postpone until the CI is not red? Currently it's completely toasted.

pmeier:
@datumbox 

> shall we postpone until the CI is not red? Currently it's completely toasted.

Yes, let's not merge this before.

datumbox:
> I think we should broaden our local_test_utils to local_utils and include all of them. Otherwise they will be sorted into the third_party category, which I feel is especially confusing for the reference scripts.

Yeap it happened:
https://github.com/pytorch/vision/pull/4384/files#diff-32b42103e815b96c670a0b5f0db055fe63f10fc8776ccbb6aa9b61a6940abba0R5

datumbox:
@pmeier Concerning the use of pre-commit hooks, could you clarify if the intention is to use them to validate VS to modify the code automatically?

NicolasHug:
> could you clarify if the intention is to use them to validate VS to modify the code automatically?

They can do both :)
At least for `black`, not for `flake8` which requires human intervention.

I'm planning to present all of this during our next meeting on Thurs. We very likely won't merge this until next week because I need to make sure all goes well on the fbcode side, etc.

datumbox:
> They can do both :)

Yes I know. I think making large scale refactoring to the code prior commit without human approval is risky. Bugs can fly under the radar. It's part of the reason why the process on FBcode is to run manually the linter formatters either via the IDE or via console. I think running the pre-commit hook for validation makes sense.

> We very likely won't merge this until next week because I need to make sure all goes well on the fbcode side, etc.

Agreed, this needs to be merged carefully. The plan is good (green CI, checks on FBcode side, comms before merge etc).

pmeier:
> Yes I know. I think making large scale refactoring to the code prior commit without human approval is risky.

If a hook makes changes, you will need to re-add and -commit, so the change is happening automatically, but you still need to manually approve it. If any hook fails, the actual commit command is never executed.

datumbox:
@pmeier This clarifies my question, thanks.

We can fine tune the developer experience/flow and stream line it but the approach looks reasonable. We can adopt and test the proposal and tweak it if need be.

NicolasHug:
On top of what Philip mentionned, commit hooks are purely optional. One can definitely preserve the more manual workflow of committing without hooks and applying the linting manually.

.circleci/regenerate.py
```diff
@@ -14,68 +14,78 @@
...
+from jinja2 import select_autoescape
 
 
 PYTHON_VERSIONS = ["3.6", "3.7", "3.8", "3.9"]
 
 RC_PATTERN = r"/v[0-9]+(\.[0-9]+)*-rc[0-9]+/"
 
 
-def build_workflows(prefix='', filter_branch=None, upload=False, indentation=6, windows_latest_only=False):
+def build_workflows(prefix="", filter_branch=None, upload=False, indentation=6, windows_latest_only=False):
```
prabhat00155, line 29: Is the a way to configure ufmt to avoid making this `' -> "` change?

.circleci/unittest/linux/scripts/run-clang-format.py
```diff
@@ -122,13 +113,12 @@ def run_clang_format_diff_wrapper(args, file):
     except DiffError:
         raise
     except Exception as e:
-        raise UnexpectedError('{}: {}: {}'.format(file, e.__class__.__name__,
-                                                  e), e)
+        raise UnexpectedError("{}: {}: {}".format(file, e.__class__.__name__, e), e)
```
prabhat00155, line 116: f-string?

.circleci/unittest/linux/scripts/run-clang-format.py
```diff
@@ -95,11 +88,9 @@ def list_files(files, recursive=False, extensions=None, exclude=None):
 def make_diff(file, original, reformatted):
     return list(
         difflib.unified_diff(
-            original,
-            reformatted,
-            fromfile='{}\t(original)'.format(file),
-            tofile='{}\t(reformatted)'.format(file),
-            n=3))
+            original, reformatted, fromfile="{}\t(original)".format(file), tofile="{}\t(reformatted)".format(file), n=3
```
prabhat00155, line 91: f-string?

.circleci/unittest/linux/scripts/run-clang-format.py
```diff
@@ -153,17 +143,10 @@ def run_clang_format_diff(args, file):
...
-            encoding='utf-8')
-    except OSError as exc:
-        raise DiffError(
-            "Command '{}' failed to start: {}".format(
-                subprocess.list2cmdline(invocation), exc
-            )
+            invocation, stdout=subprocess.PIPE, stderr=subprocess.PIPE, universal_newlines=True, encoding="utf-8"
         )
+    except OSError as exc:
+        raise DiffError("Command '{}' failed to start: {}".format(subprocess.list2cmdline(invocation), exc))
```
prabhat00155, line 149: f-string?

.circleci/regenerate.py
NicolasHug: I don't think so, using double quotes is black's convention

pmeier: `black` is the tool ding this. If we don't want it, we could put a `skip_string_normalization = true` line into its section in the `pyproject.toml`.

.circleci/unittest/linux/scripts/run-clang-format.py
pmeier: If we want this everywhere, there is [`flynt`](https://pypi.org/project/flynt/) to do this conversion automatically. We could also add it to our auto formatters to enforce it.

.circleci/regenerate.py
prabhat00155: I feel this is adding a number of unnecessary changes, given pep8 doesn't recommend one over the other.

.circleci/unittest/linux/scripts/run-clang-format.py
prabhat00155: Got it, thanks!

.circleci/regenerate.py
NicolasHug: Unfortunately this isn't something we can change, as the internal `arc lint` would force these back to double quotes

prabhat00155: I thought we were not using the default internal linter. Can't we configure that the same way as pointed above by Philip?

NicolasHug: I don't think it's possible, but even if it were I wouldn't recommend it. It's better to stick to the usual convention as much as possible. The only one we don't properly respect is the line length, for historical reasons. Are you mostly concerned about git blame noise?

prabhat00155: I have the following concerns here:
1. Unnecessary changes.
2. I am not convinced with the idea of forcing `"` over '`'. This should be a user choice given that pep8 doesn't recommend either. Also, this is not the usual convention. Line length is something recommend by pep8.
3. Python by default uses `'` for representing string:
```
>>> a = "abc"
>>> a
'abc'
```
4. `'` uses 1 key stroke against 2 for `"`.

NicolasHug: regarding 4, you're free to use single quotes if you want and let `black` automatically format them :)

For the rest, I understand your point, but one of the goal of using `black` is precisely to avoid that kind of discussion about  style preferences. It may not be perfect and suit everyone's taste (it doesn't suit mine either), but at least it's a standard that we can all automatically comply to.

NicolasHug:
Sync is looking good now (it should look like D31379666 and the few diffs below).

I'll merge this when green. We can address the flake8 issues later, they're almost all unused imports

NicolasHug:
Tests are green. Let's get this in.
🥵
May the syncing gods be with me

NicolasHug:
A million thanks to you @pmeier for your help

Issue #4386, opened 2021-09-10
maskrcnn_resnet50_fpn use _save_for_lite_interpreter error


vitansoz:
## To Reproduce

1.
model = torchvision.models.detection.maskrcnn_resnet50_fpn(num_classes=24,pretrained=False).to(device)
model.load_state_dict(checkpoint['model'])
model.eval()
2.
script_model = torch.jit.script(model)
opt_model = optimize_for_mobile(script_model)
opt_model._save_for_lite_interpreter("mask_rcnn_1.pt")
3.
RuntimeError: __torch__ types other than torchbind (__torch__.torch.classes)are not supported in lite interpreter. Workaround: instead of using arbitrary class type (class Foo()), define a pytorch class (class Foo(torch.nn.Module)).
![SharedScreenshot](https://user-images.githubusercontent.com/45025677/132150714-0b7dcb03-30bb-4024-a37e-63dda279f93d.jpg)

## Environment

Please copy and paste the output from our
[environment collection script](https://raw.githubusercontent.com/pytorch/pytorch/master/torch/utils/collect_env.py)
(or fill out the checklist below manually).

You can get the script and run it with:
```
wget https://raw.githubusercontent.com/pytorch/pytorch/master/torch/utils/collect_env.py
# For security purposes, please check the contents of collect_env.py before running it.
python collect_env.py
```

PyTorch version: 1.9.0+cpu
Is debug build: False
CUDA used to build PyTorch: None
ROCM used to build PyTorch: N/A

OS: Microsoft Windows 10 专业版
GCC version: Could not collect
Clang version: Could not collect
CMake version: version 3.18.1
Libc version: N/A

Python version: 3.6.8 (tags/v3.6.8:3c6b436a57, Dec 24 2018, 00:16:47) [MSC v.1916 64 bit (AMD64)] (64-bit runtime)
Python platform: Windows-10-10.0.19041-SP0
Is CUDA available: False
CUDA runtime version: No CUDA
GPU models and configuration: No CUDA
Nvidia driver version: No CUDA
cuDNN version: No CUDA
HIP runtime version: N/A
MIOpen runtime version: N/A

Versions of relevant libraries:
[pip3] numpy==1.18.5
[pip3] torch==1.9.0
[pip3] torchaudio==0.9.0
[pip3] torchvision==0.10.0
[conda] Could not collect

## Additional context

cc @fmassa @vfdev-5 @pmeier

datumbox:
[...]

PR #4389, opened 2021-09-10
Convert classes used for Detection to nn.Modules

datumbox:
I'm converting some of the existing classes used to do computations for Detection models to nn.Modules. It's unclear why they were structured like this, could be because they were ported from DETR in the past (perhaps @fmassa can provide more info). 

This PR was originally done to investigate https://github.com/pytorch/vision/issues/4386#issuecomment-916751913. Though it does not resolves it (might be partial solution), this conversion enables us to drop the `__annotation__` declarations.

I also tried to convert the `LevelMapper` but it's modification is not straightforward as it modifies arguments post compilation and creates JIT issues. Worth checking this on a separate PR.

### DIFF
--- a/torchvision/models/detection/_utils.py
+++ b/torchvision/models/detection/_utils.py
@@ -2,13 +2,13 @@ import math
 import torch
 
 from collections import OrderedDict
-from torch import Tensor
+from torch import Tensor, nn
 from typing import List, Tuple
 
 from torchvision.ops.misc import FrozenBatchNorm2d
 
 
-class BalancedPositiveNegativeSampler(object):
+class BalancedPositiveNegativeSampler(nn.Module):
     """
     This class samples batches, ensuring that they contain a fixed proportion of positives
     """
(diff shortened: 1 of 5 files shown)


torchvision/models/detection/_utils.py
```diff
@@ -126,7 +127,7 @@ def encode_boxes(reference_boxes, proposals, weights):
     return targets
 
 
-class BoxCoder(object):
+class BoxCoder(nn.Module):
```
datumbox, line 130: This class with its encode/decode methods does not play nice with the nn.Module's forward() approach. Not sure if we should convert it but it will allow us to drop its declarations to __annotations__

torchvision/models/detection/_utils.py
```diff
@@ -266,14 +268,18 @@ def __init__(self, high_threshold, low_threshold, allow_low_quality_matches=Fals
...
         assert low_threshold <= high_threshold
         self.high_threshold = high_threshold
         self.low_threshold = low_threshold
         self.allow_low_quality_matches = allow_low_quality_matches
 
-    def __call__(self, match_quality_matrix):
+    def forward(self, match_quality_matrix):
+        return self._forward_impl(match_quality_matrix)
+
+    def _forward_impl(self, match_quality_matrix):
```
datumbox, line 282: Workaround for overwriting the forward on inheriting classes. We do the same on quantization.

datumbox:
I'm going to close this PR until it's clearer to us whether we should convert Classes to Modules OR if there will be native support of Objects in JIT.

Issue #4238, opened 2021-08-01
Unable to reproduce classification accuracy using the reference scripts


netw0rkf10w:
## 🐛 Bug

I have been trying to reproduce the [reported](https://pytorch.org/vision/stable/models.html) 79.312% accuracy on ImageNet of `resnext101_32x8d` using the [reference scripts](https://github.com/pytorch/vision/tree/master/references/classification), but I could obtain only 75%-76%. I try two different trainings on 64 GPUs: 

- 16 nodes of 4 V100 GPUs
- 8 nodes of 8 V100 GPUs

but obtained similar results.

## To Reproduce

Clone the master branch of `torchvision`, then `cd vision/references/classification` and submit a training to 64 GPUs with arguments `--model resnext101_32x8d --epochs 100`.

The training logs (including std logs) are attached for your information: [log.txt](https://github.com/pytorch/vision/files/6912644/log.txt) and
[resnext101_32x8d_reproduced.log](https://github.com/pytorch/vision/files/6912661/resnext101_32x8d_reproduced.log)

## Expected behavior

Final top-1 accuracy should be around 79%.

## Environment

 - PyTorch / torchvision Version (e.g., 1.0 / 0.4.0): 1.8.1
 - OS (e.g., Linux): Linux
 - How you installed PyTorch / torchvision (`conda`, `pip`, source): `pip`
 - Python version: 3.8
 - CUDA/cuDNN version: 10.2
 - GPU models and configuration: V100

cc @vfdev-5

netw0rkf10w:
**Update:** I tried ResNet-50 and observed the same issue, 73.3% final accuracy instead of [76.130%](https://pytorch.org/vision/stable/models.html). Log file FYI: [log.txt](https://github.com/pytorch/vision/files/6913649/log.txt).

netw0rkf10w:
**Update:** I was hypothesizing that the issue was caused by recent changes in the code, so I tried reverting back to version [115d2eb](https://github.com/pytorch/vision/commit/115d2eb7f0280af85e5661e6a24b57a308721e13#diff-32529de4b69665fa82545850335c27c41875abdf9234a7c75476cf00cda4f83b) but then obtained again 73.3% for ResNet-50. Std logs: [resnet50_115d2eb.log](https://github.com/pytorch/vision/files/6915580/resnet50_115d2eb.log).

@fmassa @datumbox Would it be possible for you to check this please?

fmassa:
Hi @netw0rkf10w 

[...]

PR #4390, opened 2021-09-10
Clarification for training resnext101_32x8d on ImageNet

netw0rkf10w:
Resovles #4238.


references/classification/README.md
```diff
@@ -40,12 +40,16 @@ python -m torch.distributed.launch --nproc_per_node=8 --use_env train.py\
...
 ```
 python -m torch.distributed.launch --nproc_per_node=8 --use_env train.py\
     --model resnext101_32x8d --epochs 100
 ```
 
+Note that the above command corresponds to a single node with 8 GPUs. If you use
+a different number of GPUs and/or a different batch size, then the learning rate
+should be scaled accordingly. For example, the pretrained model provided by
+`torchvision` was trained on 8 nodes, each with 8 GPUs (for a total of 64 GPUs),
+with `--batch_size 16` and `--lr 0.4`.
```
NicolasHug, line 52: Small suggestion, maybe this will help illustrate your point:

```suggestion
with `--batch_size 16` and `--lr 0.4`, instead of the current defaults
which are respectively batch_size=32 and lr=0.1
```

NicolasHug (approved):
Thanks @netw0rkf10w , LGTM with a minor suggestion, feel free to ignore

references/classification/README.md
netw0rkf10w: Thanks, @NicolasHug! I've merged your suggestion.

PR #4393, opened 2021-09-12
Use a pre-compiled regex to check video DivX build

bryant1410:
May give an edge when reading video during a data loading, as there are potentially 2 regex compilations that are saved and some memory in saving some unnecessary groups.

### DIFF
--- a/torchvision/io/video.py
+++ b/torchvision/io/video.py
@@ -49,6 +49,8 @@ def _av_available() -> bool:
 _CALLED_TIMES = 0
 _GC_COLLECTION_INTERVAL = 10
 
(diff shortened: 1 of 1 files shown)


NicolasHug (commented):
Thanks for the PR @bryant1410 , from a quick look the changes look good (although regexes are always tricky to review), but it looks like the CI is failing with a segault in `test_probe_video_from_file`. I updated https://github.com/pytorch/vision/pull/3808 to figure out whether this is related or not, would you mind taking a look?

EDIT: hm nvm, looks like they're pre-existing issues

mszhanyi:
Could you rerun cmake_windows_gpu, it should be passed now.

PR #4394, opened 2021-09-13
Overwrite the existing dlls while unzipping

mszhanyi:
In vision, the build & test are running on Windows GPU machines.
So far, The default cuda driver is 10.1 in circleCI image.
So there's alreay gpu_driver_dlls, we should allow the old dlls could be overwrited by the downloaded dlls(10.2).

Else, there's the error
image
https://app.circleci.com/pipelines/github/pytorch/vision/10311/workflows/c0f4755c-4254-40cd-ad67-418617a7817f/jobs/781019


mszhanyi:
@malfet @NicolasHug
It's the first step to enable CUDA tests in vision.
 Could you please review this PR as soon as possible. I could submit the following PRs.

There're only 4 mac workflows failure. I think it's a known issue and has nothing with this PR

mszhanyi:
For 10.2, if the dlls are not overwrited.
the cuda driver version and cuda runtime version will not be consistent
image

As mentioned in https://github.com/pytorch/vision/pull/4058#issuecomment-862532742
There's always `No CUDA Runtime is found,...` when running `python setup.py install` with cuda 10.2.

I'll make the copying logic only for 10.2 in another PR.

malfet (approved):

PR #4358, opened 2021-09-03
Update unstable docs from master to main

oke-aditya:
Resolves #4356 

- [x] Add .gitignore
- [x] Edits versions.html and Readme
- [x] Create symbolic link for master to main.

Note that this PR goes to gh-pages branch and not to the master branch.

PR #4396, opened 2021-09-13
Change versions.html to main

### LINKED ISSUE #4356: Migrate Docs from master to main
### 📚 The doc issue

Inorder to view the latest docs Users should NOT navigate to https://pytorch.org/vision/master/index.html

but instead navigate to

https://pytorch.org/vision/main/index.html

This happens because Main is not visible in the version selector, 
image

Simple fix is to edit https://github.com/pytorch/vision/blob/gh-pages/versions.html#L21  
To include main instead of master.
Note that this is in the gh-pages (deployed over GitHub) branch.

I can probably send a PR!

oke-aditya:
Small fix in versions.html To reflect main instead of master.

Closes #4356 (Please close the issue after merging this, it's fully resolved)


NicolasHug (approved):
Looks like we missed this one in https://github.com/pytorch/vision/pull/4358

Thanks @oke-aditya !

PR #4398, opened 2021-09-13
Improve custom cuda installation

mszhanyi:
This PR is to improve custom cuda install script and envs, 
To pass the tests, the installed binaries and environment variables could be spitted  

1. Custom installed cuda environment variables were not set.
    It caused CMake could find the proper package.

2. cuda_install.bat was copied from builder repo. It's a little out of date.
    nvjpeg is necessary for vision build, but pytorch script doesn't include it.


mszhanyi:
@malfet @NicolasHug 
So far, the work for cmake_windows_gpu is done.

cuda tests are still skipped  because custom cuda installation isn't included in the workflow of unittest_windows_gpu 
It'll be finished in another PR.

packaging/windows/internal/cuda_install.bat
```diff
@@ -187,7 +193,12 @@ start /wait setup.exe -s %ARGS%
 popd
 
 echo Installing VS integration...
-xcopy /Y "%SRC_DIR%\temp_build\cuda\CUDAVisualStudioIntegration\extras\visual_studio_integration\MSBuildExtensions\*.*" "C:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\Common7\IDE\VC\VCTargets\BuildCustomizations"
+if "%CUDA_VER_MAJOR%" == "10" (
+    xcopy /Y "%SRC_DIR%\temp_build\cuda\CUDAVisualStudioIntegration\extras\visual_studio_integration\MSBuildExtensions\*.*" "C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Microsoft\VC\v160\BuildCustomizations"
+)
+if "%CUDA_VER_MAJOR%" == "11" (
+    xcopy /Y "%SRC_DIR%\temp_build\cuda\visual_studio_integration\CUDAVisualStudioIntegration\extras\visual_studio_integration\MSBuildExtensions\*.*" "C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Microsoft\VC\v160\BuildCustomizations"
+)
```
malfet, line 201: Can you please add a comment, why those are needed?

packaging/windows/internal/cuda_install.bat
```diff
@@ -171,12 +177,12 @@ goto cuda_common
...
 if not exist "%SRC_DIR%\temp_build\NvToolsExt.7z" (
-    curl -k -L https://www.dropbox.com/s/9mcolalfdj4n979/NvToolsExt.7z?dl=1 --output "%SRC_DIR%\temp_build\NvToolsExt.7z"
+    curl -k -L https://ossci-windows.s3.us-east-1.amazonaws.com/builder/NvToolsExt.7z --output "%SRC_DIR%\temp_build\NvToolsExt.7z"
     if errorlevel 1 exit /b 1
 )
 
-if not exist "%SRC_DIR%\temp_build\gpu_driver_dlls.7z" (
-    curl -k -L "https://drive.google.com/u/0/uc?id=1injUyo3lnarMgWyRcXqKg4UGnN0ysmuq&export=download" --output "%SRC_DIR%\temp_build\gpu_driver_dlls.zip"
+if not exist "%SRC_DIR%\temp_build\gpu_driver_dlls.zip" (
+    curl -k -L "https://ossci-windows.s3.us-east-1.amazonaws.com/builder/additional_dlls.zip" --output "%SRC_DIR%\temp_build\gpu_driver_dlls.zip"
```
malfet, line 185: Can it be done in separate PR?

.circleci/unittest/windows/scripts/set_cuda_envs.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
+
+if [ "${CU_VERSION:-}" != "cpu" ] ; then
+    if [[ ${#CU_VERSION} -eq 5 ]]; then
+        CUDA_VERSION="${CU_VERSION:2:2}.${CU_VERSION:4:1}"
+    fi
```
malfet, line 6: Please print an error, if it is not the case
```suggestion
    if [[ ${#CU_VERSION} -eq 5 ]]; then
        CUDA_VERSION="${CU_VERSION:2:2}.${CU_VERSION:4:1}"
    else
        echo "Unrecognized CU_VERSION format:", ${CU_VERSION}
        exit 1
    fi
```

.circleci/unittest/windows/scripts/set_cuda_envs.sh
```diff
@@ -0,0 +1,40 @@
...
+
+    # It's a log to see if CU_VERSION exists, if not, we use environment CUDA_VERSION directly
+    echo "Using CUDA $CUDA_VERSION as determined by CU_VERSION $CU_VERSION"
+
+    version="$(python -c "print('.'.join(\"${CUDA_VERSION}\".split('.')[:2]))")"
```
malfet, line 11: I'm having a hard time understanding what this statement is doing. CUDA_VERSION is already MAJOR.MINOR, why we need to split and join it again?

.circleci/unittest/windows/scripts/set_cuda_envs.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
+
+if [ "${CU_VERSION:-}" != "cpu" ] ; then
```
malfet, line 3: Nit: instead of having an if statement for the entire script, just exist early in case of CPU build
```suggestion
if [ "${CU_VERSION:-}" == "cpu" ] ; then
  exit 0
fi
```

malfet (changes requested):
I though I've already reviewed a PR that replaces dropbox.com with ossci-windows buckets. Can you please turn it into a separate one?
And the same about nvjpeg installation: it's not needed for packaging right now, is it? (it passes the CI) So why are you saying it is a mandatory feature?

mszhanyi:
> I though I've already reviewed a PR that replaces dropbox.com with ossci-windows buckets. Can you please turn it into a separate one?
> And the same about nvjpeg installation: it's not needed for packaging right now, is it? (it passes the CI) So why are you saying it is a mandatory feature?
I think If packaging had smoke tests, it would fail. @NicolasHug , is nvjpeg needed for packing?
And the main purpose of this PR is to improve the cuda install script, which is shared by unittest and packaging.
So In the next PR, I could only add some change for unittest workflow.
To detect this failure as early as possible, I add a check in set_cuda_env. Or the CI would fail in build stage.
https://github.com/pytorch/vision/blob/7947fc8fb38b1d3a2aca03f22a2e6a3caa63f2a0/setup.py#L325.

packaging/windows/internal/cuda_install.bat
mszhanyi: I read through the [Nvidia doc](https://docs.nvidia.com/cuda/cuda-installation-guide-microsoft-windows/index.html#build-customizations). the BuildCustomizaitons is for VC project. Since we only support ninja to build, it's useless now.
The old code had a error because the directory is changed in Visual Studio 2019, but it won't break the CI.
So, I'll just remove the old code.

mszhanyi: @peterjc123 for double check.

packaging/windows/internal/cuda_install.bat
mszhanyi: OK, I'll do it in another PR.

.circleci/unittest/windows/scripts/set_cuda_envs.sh
mszhanyi: I remember I came across the passed CUDA_VERSION  is 11.2.2. (MAJOR.MINOR.UPDATE)
So I use [:2] to ensure the version number I use is MAJOR.MINOR

.circleci/unittest/windows/scripts/set_cuda_envs.sh
mszhanyi: @malfet 
The set_cuda_envs.sh is used for unittest_windows_gpu and cmake_windows_gpu.
In unittest_windows_gpu, the passed environment name is CUDA_VERSION (10.2)
https://github.com/pytorch/vision/blob/fe3118c51b487faf1aa59861ffd08b5658f41254/.circleci/config.yml#L742

In cmake_windows_gpu, the passed environment name is cu_version (cu102)
https://github.com/pytorch/vision/blob/fe3118c51b487faf1aa59861ffd08b5658f41254/.circleci/config.yml#L1704

To compatible with these 2 workflows, CU_VERSION(cmake_windows_gpu) is checked first, if no CU_VERSION, I'll use CUDA_VERSION (unittest_windows_gpu).

It looks a little ugly. Maybe we could make the refactory in the future.

.circleci/unittest/windows/scripts/set_cuda_envs.sh
mszhanyi: To avoid confusion, I remove it. MAJOR.MINOR.UPDATE should be a incorrect input.

.circleci/unittest/windows/scripts/set_cuda_envs.sh
mszhanyi: updated

Issue #3410, opened 2021-02-17
Create reset_parameters method for ResNet blocks and models


vadimkantorov:
This would ease re-initialization some blocks that are not meant to be pretrained/frozen:
https://github.com/pytorch/vision/blob/master/torchvision/models/resnet.py#L188-L203

prachigarg23:
Hi, I'd like to work on this feature. I have a good understanding of the ResNet code but this will be my first attempt at open source contribution. To clarify, the reset_parameter method should be able to reset/initialise parameters for any layer or block of layers passed right? Since that would serve the purpose of someone who has to customize the random init or freezing/pretraining of specific layers in the architecture? Can you please guide me on whether the reset parameters method needs to be written for a single layer or a set of layers? I can think of these changes after creating a reset_parameters method and moving the mentioned code into it: 

`class ResNet(nn.Module):

    def __init__(
        self,
        block: Type[Union[BasicBlock, Bottleneck]],
        layers: List[int],
        num_classes: int = 1000,
        zero_init_residual: bool = False,
        groups: int = 1,
        width_per_group: int = 64,
        replace_stride_with_dilation: Optional[List[bool]] = None,
        norm_layer: Optional[Callable[..., nn.Module]] = None
    ) -> None:
        super(ResNet, self).__init__()
        if norm_layer is None:
            norm_layer = nn.BatchNorm2d
        self._norm_layer = norm_layer

        self.inplanes = 64
        self.dilation = 1
        self.zero_init_residual = zero_init_residual

        if replace_stride_with_dilation is None:
            # each element in the tuple indicates if we should replace
            # the 2x2 stride with a dilated convolution instead
            replace_stride_with_dilation = [False, False, False]
        if len(replace_stride_with_dilation) != 3:
            raise ValueError("replace_stride_with_dilation should be None "
                             "or a 3-element tuple, got {}".format(replace_stride_with_dilation))
        self.groups = groups
        self.base_width = width_per_group
        self.conv1 = nn.Conv2d(3, self.inplanes, kernel_size=7, stride=2, padding=3,
                               bias=False)
        self.bn1 = norm_layer(self.inplanes)
        self.relu = nn.ReLU(inplace=True)
        self.maxpool = nn.MaxPool2d(kernel_size=3, stride=2, padding=1)
        self.layer1 = self._make_layer(block, 64, layers[0])
        self.layer2 = self._make_layer(block, 128, layers[1], stride=2,
[...]

PR #3182, opened 2020-12-16
Add MobileNetV3 Architecture in TorchVision

datumbox:
Partially fixes #1676

Depends and cherrypicks commits from #3177

PR #4403, opened 2021-09-14
Add RegNet Architecture in TorchVision

### LINKED ISSUE #2655: RegNet in torchvision ?
## 🚀 Feature
Add RegNet trunks in torchvision

## Motivation
RegNets were proposed in this paper https://arxiv.org/pdf/2003.13678.pdf, they're showing very interesting performance and speed. They have been open sourced [already](https://github.com/facebookresearch/pycls), but are not usable in a straightforward way for people used to having reference models in torchvision. Another implementation is available in [ClassyVision](https://github.com/facebookresearch/ClassyVision/blob/master/classy_vision/models/regnet.py) (I'm a co-author of this one), but it does not cover all use cases.

## Pitch
Start from the ClassyVision RegNet support and implement RegNets in torchvision.

## Alternatives
Let users use RegNets from external implementations

## Additional context
This has been discussed with @pdollar, one of the RegNet authors. CC @fmassa 

cc @vfdev-5

kazhang:
Resolves #2655


torchvision/models/regnet.py
```diff
@@ -0,0 +1,785 @@
...
+
+    Args:
+        pretrained (bool): If True, returns a model pre-trained on ImageNet
+        progress (bool): If True, displays a progress bar of the download to stderr
+    """
+    params = RegNetParams(depth=23, w_0=320, w_a=69.86, w_m=2.0, group_width=168,
+                          use_se=False, **kwargs)
+    return _regnet("regnet_x_32gf", params, pretrained, progress, **kwargs)
+
+# TODO(kazhang): Add RegNetZ_500MF and RegNetZ_4GF
```
kazhang, line 785: RegNetZ from paper [Fast and Accurate Model Scaling](https://arxiv.org/abs/2103.06877) and is available in Classy.
We would need ConvBNReLU in before the last FC layer. I will address this in the next commit.

kazhang:
Publish to gather early feedback.
cc @fmassa @datumbox

torchvision/models/regnet.py
```diff
@@ -0,0 +1,785 @@
...
+        bottleneck_multiplier: float = 1.0,
+        stem_type: Callable[..., nn.Module] = SimpleStemIN,
+        stem_width: int = 32,
+        block_type: Callable[..., nn.Module] = ResBottleneckBlock,
+        activation: Callable[..., nn.Module] = nn.ReLU,
+        use_se: bool = True,
+        se_ratio: float = 0.25,
+        bn_epsilon: float = 1e-05,
+        bn_momentum: float = 0.1,
+        num_classes: int = 1000,
```
datumbox, line 405: The majority of the constructors in the models of vision receive many of these params (such as `num_classes`) in the constructor. See [here](https://github.com/pytorch/vision/blob/main/torchvision/models/googlenet.py#L68).

datumbox (commented):
@kazhang thanks for the contribution!

You didn't specify if you wanted us to focus on the ML side or on the architectural side of the code. Given that in the past you have implemented this model on ClassyVision, below I focus only on high-level comments about the architecture. Hope it's OK.

My comments below can be grouped in 2 categories:
1. I highlight whenever you introduce a block for which we have one or many copies. This is in order to discuss whether we want to share an existing module. In the past we favoured keeping the models with minimum inheritance so that people can copy-paste them and edit them easily, but this does not scale very well.
2. I let you know when you use a different pattern/idiom than the one usually found in vision. All patterns have pros/cons, your usages are valid and tastes are highly subjective. So I think this will boil down to the question "Do we want to keep the entire codebase consistent?"

Happy to provide more detailed input once we discuss these. Let me know your thoughts, happy to jump in a call to speed this up as well and later summarize in this PR.

torchvision/models/regnet.py
kazhang: I'm also considering whether we should have this kind of params container. If so, what should be in the params container?
Would it make more sense if we only keep the params that are used in `get_expanded_params` in the container and move the rest(e.g. num_classes, stem_type, block_type, activation) to model constructor?

torchvision/models/regnet.py
```diff
@@ -0,0 +1,785 @@
...
+        )
+
+    def forward(self, x: Tensor) -> Tensor:
+        x_squeezed = self.avgpool(x)
+        x_excited = self.excitation(x_squeezed)
+        x_scaled = x * x_excited
+        return x_scaled
+
+
+class BasicTransform(nn.Sequential):
```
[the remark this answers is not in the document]
kazhang: > Could we reuse the ConvBNActivation layer from here?

Do you mean replacing line 79-81 with `ConvBNActivation`? or inherit from `ConvBNActivation` and add `Conv2D`, final_bn?

> passing all the layers on the constructor rather than attaching them on self.

I don't have a strong preference here. With currently implementation, we could divide layers into blocks (a, b, c), see `BottleneckTransform`. I assume the layer name would be slightly more readable when extracting features (e.g. `block4-4.f.final_bn`)

torchvision/models/regnet.py
```diff
@@ -0,0 +1,785 @@
...
+    """
+    Squeeze and excitation layer from
+    `"Squeeze-and-Excitation Networks" <https://arxiv.org/pdf/1709.01507>`_.
+    """
+
+    def __init__(
+        self,
+        in_channels: int,
+        reduction_ratio: Optional[int] = 16,
+        reduced_channels: Optional[int] = None,
```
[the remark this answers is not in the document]
kazhang: I agree that we should reduce the code duplication. In fact, we're not using `reductiton_ratio` in the model implementation. After removing it, the only difference between this and the one in [efficentnet](https://github.com/pytorch/vision/blob/main/torchvision/models/efficientnet.py#L34)  seems to be the activation. Maybe we could make the efficientnet one extendible. The question is should we do it before or after this PR :)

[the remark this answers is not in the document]
datumbox: Since you agree, I would rather if we do it earlier than later (because of the upcoming release). I also like your idea of marking the class with `_` to indicate it's private, let's do that on efficientnet (though I would like to bring up in our discussion the developer API). Could you send a PR where you extend the efficientnet class to support different activations so that you could reuse it?

torchvision/models/regnet.py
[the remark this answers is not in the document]
datumbox: > Do you mean replacing line 79-81 with ConvBNActivation? or inherit from ConvBNActivation and add Conv2D, final_bn?

I was thinking of replacing. That would also mean that you would be able to extend the original `ConvBNActivation` to meet your needs (in a BC manner) to make it reusable.

> With currently implementation, we could divide layers into blocks (a, b, c)

I personally prefer passing it on the constructor but that's just my preference. @fmassa could you weight in here?

torchvision/models/regnet.py
```diff
@@ -0,0 +1,785 @@
...
+            nn.BatchNorm2d(width_out, eps=bn_epsilon, momentum=bn_momentum),
+            activation,
+            nn.Conv2d(width_out, width_out, 3, stride=1, padding=1, bias=False),
+        )
+
+        self.final_bn = nn.BatchNorm2d(width_out, eps=bn_epsilon, momentum=bn_momentum)
+        self.depth = 2
+
+
+class ResStemCifar(nn.Sequential):
```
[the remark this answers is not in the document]
datumbox: Sounds good. I spotted a couple of other classes that are never used in the implementation. I would remove these as well unless you plan to create networks that actually use them. What we can do is move them to the examples to show to people how they could implement different blocks.

torchvision/models/regnet.py
```diff
@@ -0,0 +1,785 @@
...
+    """Residual basic block: x + F(x), F = basic transform."""
+
+    def __init__(
+        self,
+        width_in: int,
+        width_out: int,
+        stride: int,
+        bn_epsilon: float,
+        bn_momentum: float,
+        activation: nn.Module,
```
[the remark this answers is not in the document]
datumbox: Aha! Sorry I didn't realize, thanks for the clarification.

I think we avoid passing and reusing the same activation instance of an activation function because:
1. Some activation methods might potentially learn parameters (like [PReLU](https://pytorch.org/docs/stable/generated/torch.nn.PReLU.html#torch.nn.PReLU)) so at each place we might need different instances so that they don't share weights.
2. The Quantization API will have issues fusing the operators. We typically use the nn.* versions of them so that they can fuse them easier. 

As far as I see, you also make use of the `Callable[..., nn.Module]` [elsewhere](https://github.com/pytorch/vision/pull/4403/files/617b833a12c7f08d42b8b13d2037ddcf8075cb2c#diff-62577d15011d998701098e5e567a78bd117257deeeb2a23b72e550e1f6b14279R400), would you mind taking the same approach here?

torchvision/models/regnet.py
datumbox: Very good points.

The latter approach worked very nicely form me in the past (aka passing the rest on constructor). I also started constructing the parameter objects in the building methods and passing them on the constructor (instead of adding them on a method of the class). That allowed me to parameterize and change the architecture of the network at will with minimum modifications on the Model class itself.

Now to be clear, the above approach is not taken universally in the codebase but rather followed on latest models such as MobileNetV3 and EfficientNets. So perhaps this is something we should discuss and decide here for going forwards. No strong opinions on my side. @fmassa any thoughts around this?

torchvision/models/regnet.py
```diff
@@ -0,0 +1,785 @@
...
+        x = self.stem(x)
+        x = self.trunk_output(x)
+
+        x = self.avgpool(x)
+        x = x.flatten(start_dim=1)
+        x = self.fc(x)
+
+        return x
+
+    def _init_weights(self) -> None:
```
vadimkantorov, line 571: I propose a different semi-standard name for this function: `reset_parameters()` + making it public for resetting the parameters if needed

datumbox: You are right to say right now we don't offer a good way to achieve this and that we probably should create a mechanism. I believe you have create a feature request for this at #3410.

This is worth discussing separately from this PR because the specific idiom proposed is not currently used in vision. In practice in all cases, the models initialize their weights inside the constructor.

vadimkantorov: A related discussion is at https://github.com/pytorch/pytorch/issues/29523

vadimkantorov: Guidance here also proposes to always using nn.init methods: https://pytorch.org/tutorials/prototype/skip_param_init.html#updating-modules-to-support-skipping-initialization

torchvision/models/regnet.py
```diff
@@ -0,0 +1,785 @@
...
+            3,  # width_in
+            params.stem_width,
+            params.bn_epsilon,
+            params.bn_momentum,
+            activation,
+        )
+
+        current_width = params.stem_width
+
+        self.trunk_depth = 0
```
fmassa, line 519: Are those meta-informations used somewhere in Classy?

torchvision/models/regnet.py
```diff
@@ -0,0 +1,785 @@
...
+
+        return x
+
+    def _init_weights(self) -> None:
+        # Performs ResNet-style weight initialization
+        for m in self.modules():
+            if isinstance(m, nn.Conv2d):
+                # Note that there is no bias due to BN
+                fan_out = m.kernel_size[0] * m.kernel_size[1] * m.out_channels
+                m.weight.data.normal_(mean=0.0, std=math.sqrt(2.0 / fan_out))
```
fmassa, line 577: Let's use `torch.nn.init` as @vadimkantorov pointed out, with the proper parameters to make everything compatible

torchvision/models/regnet.py
```diff
@@ -0,0 +1,785 @@
...
+        group_widths_min = [min(g, w_bot) for g, w_bot in zip(group_widths, widths)]
+
+        # Compute the adjusted widths so that stage and group widths fit
+        ws_bot = [_make_divisible(w_bot, g) for w_bot, g in zip(widths, group_widths_min)]
+        stage_widths = [int(w_bot / b) for w_bot, b in zip(ws_bot, bottleneck_ratios)]
+        return stage_widths, group_widths_min
+
+
+class RegNet(nn.Module):
+    def __init__(self, params: RegNetParams) -> None:
```
fmassa, line 503: I think I would prefer to break this down into multiple different arguments, like `(stem, stages)` or something like that

torchvision/models/regnet.py
```diff
@@ -0,0 +1,785 @@
...
+        activation = params.activation(inplace=True)
+
+        # Ad hoc stem
+        self.stem = params.stem_type(
+            3,  # width_in
+            params.stem_width,
+            params.bn_epsilon,
+            params.bn_momentum,
+            activation,
+        )
```
fmassa, line 515: I think it is a bit more generic to let the user pass the fully-materialized `stem` in the constructor, so that we don't get fixed on the particular arguments that are necessary to construct the `stem`

torchvision/models/regnet.py
```diff
@@ -0,0 +1,785 @@
...
+        The second step is to compute the parameters per stage,
+        taking into account the skip connection and the final 1x1 convolutions.
+        We use the fact that the output width is constant within a stage.
+        """
+
+        QUANT = 8
+        STRIDE = 2
+
+        # Compute the block widths. Each stage has one unique block width
+        widths_cont = np.arange(self.depth) * self.w_a + self.w_0
```
fmassa, line 450: not super important, but would it be possible to use `torch` functions instead of `np` functioned whenever possible?

fmassa (commented):
Thanks a ton for the PR, this is looking good!

I did an initial pass and I have a few high-level comments.

Maybe we could have a joint chat with Vasilis to try and converge on some specific points brought by our comments, what do you think?

torchvision/models/regnet.py
```diff
@@ -0,0 +1,785 @@
...
+
+        self.a = nn.Sequential(
+            nn.Conv2d(width_in, width_out, 3, stride=stride, padding=1, bias=False),
+            nn.BatchNorm2d(width_out, eps=bn_epsilon, momentum=bn_momentum),
+            activation,
+            nn.Conv2d(width_out, width_out, 3, stride=1, padding=1, bias=False),
+        )
+
+        self.final_bn = nn.BatchNorm2d(width_out, eps=bn_epsilon, momentum=bn_momentum)
+        self.depth = 2
```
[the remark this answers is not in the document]
kazhang: It was used to calculate model depth(i.e. number of blocks) for classy vision specific. The attribute has been removed from Classy but apparently those are not. I will remove them.

torchvision/models/regnet.py
vadimkantorov: (also I found that VGGNet uses `_initialize_weights`)

torchvision/models/regnet.py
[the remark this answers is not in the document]
kazhang: @mannatsingh are `ResStemIN`, `VanillaBlock`, `ResBasicBlock` actively/commonly used in RegNet model construction?

torchvision/models/regnet.py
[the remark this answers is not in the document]
datumbox: Is the depth param that you removed from here, different form the depth param that still exists in `RegNetParams`?

torchvision/models/regnet.py
```diff
@@ -0,0 +1,708 @@
...
+        group_width: int,
+        bottleneck_multiplier: float = 1.0,
+        stem_type: Callable[..., nn.Module] = SimpleStemIN,
+        stem_width: int = 32,
+        block_type: Callable[..., nn.Module] = ResBottleneckBlock,
+        activation: Callable[..., nn.Module] = nn.ReLU,
+        use_se: bool = True,
+        se_ratio: float = 0.25,
+        bn_epsilon: float = 1e-05,
+        bn_momentum: float = 0.1,
```
datumbox, line 337: Shall we pass the norm_layer also here? (see [this](https://github.com/pytorch/vision/blob/main/torchvision/models/resnet.py#L155-L160)). This is useful because it will allow us to pass other layers (such as [FrozenBN](https://github.com/pytorch/vision/blob/af228fa89bd4386816607c9d5de77f1304316983/torchvision/models/detection/backbone_utils.py#L52)) in detection.

kazhang: Sounds good.

torchvision/models/regnet.py
```diff
@@ -0,0 +1,708 @@
...
+        activation_layer: Callable[..., nn.Module],
+    ) -> None:
+        super().__init__(
+            ConvBNActivation(width_in, width_out, kernel_size=7, stride=2,
+                             norm_layer=norm_layer, activation_layer=activation_layer),
+            nn.MaxPool2d(3, stride=2, padding=1),
+        )
+
+
+class SimpleStemIN(ConvBNActivation):
```
datumbox, line 102: Not currently used.

torchvision/models/regnet.py
```diff
@@ -0,0 +1,708 @@
...
+        width_in: int,
+        width_out: int,
+        norm_layer: Callable[..., nn.Module],
+        activation_layer: Callable[..., nn.Module],
+    ) -> None:
+        super().__init__(width_in, width_out, kernel_size=3, stride=2,
+                         norm_layer=norm_layer, activation_layer=activation_layer)
+
+
+class VanillaBlock(nn.Sequential):
```
datumbox, line 116: Not currently used.

torchvision/models/regnet.py
```diff
@@ -0,0 +1,708 @@
...
+            a=nn.Sequential(
+                ConvBNActivation(width_in, width_out, kernel_size=3, stride=stride,
+                                 norm_layer=norm_layer, activation_layer=activation_layer),
+                nn.Conv2d(width_out, width_out, 3, stride=1, padding=1, bias=False),
+            ),
+            final_bn=norm_layer(width_out),
+        ))
+
+
+class ResStemIN(nn.Sequential):
```
datumbox, line 85: Not currently used.

torchvision/models/regnet.py
```diff
@@ -0,0 +1,708 @@
...
+        self.stem_type = stem_type
+        self.block_type = block_type
+        self.activation = activation
+        self.norm_layer = partial(nn.BatchNorm2d, eps=bn_epsilon, momentum=bn_momentum)
+        self.stem_width = stem_width
+        self.use_se = use_se
+        self.se_ratio = se_ratio if use_se else None
+        self.num_classes = num_classes
+
+    def get_expanded_params(self):
```
kazhang, line 357: Reason to keep this API in `RegNetParams`:
Here we calculate the width and depth for each block in `get_expanded_params`, but In model scaling use case, we want to pass in pre-computed blocks widths, depths as Lists. See [AnyNet in Classy](https://github.com/facebookresearch/ClassyVision/blob/main/classy_vision/models/anynet.py#L367-L371). I'm thinking for model scaling, user could inherit `RegNetParams` and override the `get_expanded_params` method to directly return the block settings.
cc @mannatsingh

torchvision/models/regnet.py
kazhang, line 85: @mannatsingh could you please comment on whether we need `ResStemIN`, `VanillaBlock` and `ResBottleneckLinearBlock`? they're not currently used in predefined model architectures.

Stanislav1975 (approved):

mannatsingh (commented):
High level comment for the group - do we want to support the following - 
- [Fast and Accurate Model Scaling](https://arxiv.org/abs/2103.06877) proposes a way to scale models. Once a model is scaled, it cannot be defined by `w_a`, `w_m`, etc. anymore and instead needs to be specified by the stage widths, group widths, etc. which means the `RegNetParams` API does not suffice. We already have models like these being trained currently with multiple downstream applications (happy to share details internally). In Classy Vision we had an `AnyNetParams` API for this (https://github.com/facebookresearch/ClassyVision/blob/b9ea86b39cc469867aabb76ada33bf5a8530eb19/classy_vision/models/anynet.py#L364), which `RegNetParams` derived from.
- The [Designing Network Design Spaces](https://arxiv.org/abs/2003.13678) paper specified AnyNets as fairly general networks and `RegNetX`, `RegNetY`, `RegNetZ` are instantiations of these. Which is why we have multiple block types - a user can specify very general networks with these settings. 

IMO at the very least we should support instantiating scaled models. Supporting general AnyNets might be out of scope for torchvision. Answering these two questions is important to answer some of the comments in code currently!

cc @pdollar

torchvision/models/regnet.py
mannatsingh: I think something like this is quite important! But also, if we want to support it, we should do it fully instead of allowing users to override it with their code - this can get messy.

torchvision/models/regnet.py
mannatsingh: Yes, these blocks aren't currently used for any RegNet version! See my high level comment for this :)

torchvision/models/regnet.py
```diff
@@ -0,0 +1,672 @@
...
+            # beginning of the block
+            width_se_out = int(round(se_ratio * width_in))
+            layers["se"] = SqueezeExcitation(
+                input_channels=w_b,
+                squeeze_channels=width_se_out,
+                activation=activation_layer,
+            )
+
+        layers["c"] = nn.Conv2d(w_b, width_out, 1, stride=1, padding=0, bias=False)
+        layers["final_bn"] = norm_layer(width_out)
```
kazhang, line 165: @mannatsingh / @blefaudeux does the layer name(a, b, c final_bn) have special meanings? otherwise we could  fuse "c" and "final_bn" into one `ConvBN` op.

torchvision/models/regnet.py
```diff
@@ -0,0 +1,688 @@
...
+        x = self.fc(x)
+        return x
+
+
+class RegNet(nn.Module):
+    def __init__(
+        self,
+        stem: nn.Module,
+        blocks: nn.Module,
+        classifier: nn.Module,
```
kazhang, line 450: In the most recent commit, I change the model constructor to take layer instances instead of layer constructor. In this way, user could construct their own stem or classifier outside of model. The downside is that model constructor along cannot build the model, user would also need helper functions like `_make_stem`, `_make_blocks`.
I haven't seen many patterns like this in TorchVision, so I wanna give it a try and gather feedback.

torchvision/models/regnet.py
```diff
@@ -0,0 +1,688 @@
...
+                nn.init.zeros_(m.bias)
+            elif isinstance(m, nn.Linear):
+                nn.init.normal_(m.weight, mean=0.0, std=0.01)
+                nn.init.zeros_(m.bias)
+
+
+def _regnet(arch: str, block_params: BlockParams, pretrained: bool, progress: bool, **kwargs: Any) -> RegNet:
+    norm_layer = kwargs["norm_layer"] if "norm_layer" in kwargs else partial(nn.BatchNorm2d, eps=1e-05, momentum=0.1)
+    activation = kwargs["activation"] if "activation" in kwargs else nn.ReLU
+    num_classes = kwargs["num_classes"] if "num_classes" in kwargs else 1000
```
kazhang, line 485: maybe `kwargs.get("num_classes", 1000)` is better.

blefaudeux: slightly less redundant ? taste and colours but I would prefer that indeed

torchvision/models/regnet.py
kazhang: @mannatsingh alternatively, we could change the constructor to take list of depths, widths, group_widths etc. and have a class method that take `w_0`, `w_a`, `w_m` to compute the lists.

torchvision/models/regnet.py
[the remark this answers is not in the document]
kazhang: They are different. The depth param in `RegNetParams` refers to number of blocks(`AnyStage`) in **the model**. The deleted depth refers to the number of layers inside **a block**.

kazhang:
> High level comment for the group - do we want to support the following -
> 
>     * [Fast and Accurate Model Scaling](https://arxiv.org/abs/2103.06877) proposes a way to scale models. Once a model is scaled, it cannot be defined by `w_a`, `w_m`, etc. anymore and instead needs to be specified by the stage widths, group widths, etc. which means the `RegNetParams` API does not suffice. We already have models like these being trained currently with multiple downstream applications (happy to share details internally). In Classy Vision we had an `AnyNetParams` API for this (https://github.com/facebookresearch/ClassyVision/blob/b9ea86b39cc469867aabb76ada33bf5a8530eb19/classy_vision/models/anynet.py#L364), which `RegNetParams` derived from.
> 
>     * The [Designing Network Design Spaces](https://arxiv.org/abs/2003.13678) paper specified AnyNets as fairly general networks and `RegNetX`, `RegNetY`, `RegNetZ` are instantiations of these. Which is why we have multiple block types - a user can specify very general networks with these settings.
 
We could support both scaled and regular regnet model if we init the `RegNetParams` with lists of params and have a class method to compute list of params.


Regarding to various block type,  I'm leaning towards no including unused block types. From engineering point of view, if a component is not actively used within the code base, it is likely to be ignored in regular maintenance and eventually abandoned. Additionally, it's much easier to add a component than deleting it in a BC manner.

torchvision/models/regnet.py
kazhang: I saw a couple places where we assume the model takes `num_classes`: [reference script](https://github.com/datumbox/vision/blob/main/references/classification/train.py#L189), [test_model](https://github.com/datumbox/vision/blob/main/test/test_models.py#L444). classifier might have to be constructed inside model then.

datumbox: I think both ways have pros and cons and when working on a new code-base it's a matter of preference. Perhaps even what you propose can be better.

Unfortunately given that the rest of the code-base uses the other idiom, diverging can cause issues. The main issue that comes to mind is when the backbone is incorporated to Object Detection and Segmentation models. Diverging means you will need to do this differently. This increases complexity and could potentially confuse users.

To avoid delaying the PR, I would recommend switching back to callable layers and using `num_classes` similar to all other models. This will allow you to eliminate some of the helper methods/classes that you introduce.. Then open a RFC where you record pros/cons of the change and investigate how we potentially do this in a BC manner. I think the models area can use a facelift, so let's just discuss/organize this there.

torchvision/models/regnet.py
```diff
@@ -0,0 +1,688 @@
...
+        # Init weights and good to go
+        self.reset_parameters()
+
+    def forward(self, x: Tensor) -> Tensor:
+        x = self.stem(x)
+        x = self.blocks(x)
+        x = self.classifier(x)
+        return x
+
+    def reset_parameters(self) -> None:
```
datumbox, line 467: Let's not commit to a public method yet:
```suggestion
    def _reset_parameters(self) -> None:
```

torchvision/models/regnet.py
datumbox: nit: I would probably pass these as named parameters similar to [other builder methods](https://github.com/pytorch/vision/blob/main/torchvision/models/mobilenetv3.py#L188).

torchvision/models/regnet.py
datumbox: I would probably make this private:
```suggestion
    def _get_expanded_params(self):
```

torchvision/models/regnet.py
datumbox: I agree with the recent https://github.com/pytorch/vision/pull/4403#issuecomment-924434522 by @kazhang that any class not used should be removed. We can always add them back later if we add more models.

torchvision/models/regnet.py
```diff
@@ -0,0 +1,688 @@
...
+    ) -> None:
+        super().__init__()
+
+        # Use skip connection with projection if shape changes
+        self.proj_block = (width_in != width_out) or (stride != 1)
+        if self.proj_block:
+            self.proj = nn.Conv2d(
+                width_in, width_out, 1, stride=stride, padding=0, bias=False
+            )
+            self.bn = norm_layer(width_out)
```
datumbox, line 205: Nit: This can be replaced with `ConvBNActivation` that uses the identity.

torchvision/models/regnet.py
```diff
@@ -0,0 +1,688 @@
...
+        norm_layer: Callable[..., nn.Module],
+        activation_layer: Callable[..., nn.Module],
+    ) -> None:
+        super().__init__(OrderedDict(
+            a=nn.Sequential(
+                ConvBNActivation(width_in, width_out, kernel_size=3, stride=stride,
+                                 norm_layer=norm_layer, activation_layer=activation_layer),
+                nn.Conv2d(width_out, width_out, 3, stride=1, padding=1, bias=False),
+            ),
+            final_bn=norm_layer(width_out),
```
datumbox, line 41: Nit: This could also be replaced with `ConvBNActivation` BUT knowing that it is the last bn is beneficial as this allows us to do easily [0 gamma initialization](https://arxiv.org/abs/1812.01187). Unfortunately given that this is not supported by any of the other models, we would need to find a different generic solution to support the method. So I would be in favour of being consistent and replacing it.

torchvision/models/regnet.py
```diff
@@ -0,0 +1,688 @@
...
+        *args,
+        **kwargs,
+    ) -> None:
+        super().__init__()
+        self.proj_block = (width_in != width_out) or (stride != 1)
+        if self.proj_block:
+            self.proj = nn.Conv2d(
+                width_in, width_out, 1, stride=stride, padding=0, bias=False
+            )
+            self.bn = norm_layer(width_out)
```
datumbox, line 130: Nit: this can be replaced with a ConvBNActivation that uses identity.

torchvision/models/efficientnet.py
```diff
@@ -32,15 +32,21 @@
 
 
 class SqueezeExcitation(nn.Module):
-    def __init__(self, input_channels: int, squeeze_channels: int):
+    def __init__(
```
datumbox, line 35: Keep your new version from #4448

torchvision/models/regnet.py
datumbox: Nit: This could be replaced by ConvBNActivation. See my other comment on why `final_bn` can be useful to know.

datumbox (commented):
Thanks for the updates @kazhang. I think we are getting very close, I really like the changes.

I left couple of comments, feel free to ignore the nits as we can address them optionally on a separate PR. Let me know what you think.

mannatsingh:
> We could support both scaled and regular regnet model if we init the RegNetParams with lists of params and have a class method to compute list of params.

> Regarding to various block type, I'm leaning towards no including unused block types. From engineering point of view, if a component is not actively used within the code base, it is likely to be ignored in regular maintenance and eventually abandoned. Additionally, it's much easier to add a component than deleting it in a BC manner.

Agree on both ascounts! Thanks Kai!

torchvision/models/regnet.py
```diff
@@ -0,0 +1,574 @@
...
+        super().__init__()
+
+        if stem_type is None:
+            stem_type = SimpleStemIN
+        if norm_layer is None:
+            norm_layer = nn.BatchNorm2d
+        if block_type is None:
+            block_type = ResBottleneckBlock
+        if activation is None:
+            activation = nn.ReLU
```
datumbox, line 295: I agree with @kazhang's earlier comment that this looks unnecessarily busy. It's true that in most places in vision, the idiom used is to pass `None` for the default layer and then assign a value. That seems not great IMO.

@fmassa any specific historical reason why we did this? Any concerns adopting Kai's proposal and do:
```python
        stem_type: Callable[..., nn.Module] = SimpleStemIN,
        block_type: Callable[..., nn.Module] = ResBottleneckBlock,
        norm_layer: Callable[..., nn.Module] = nn.BatchNorm2d,
        activation: Callable[..., nn.Module] = nn.ReLU,
```

torchvision/models/regnet.py
```diff
@@ -0,0 +1,574 @@
...
+                        bottleneck_multiplier,
+                        block_params.se_ratio,
+                        stage_index=i + 1,
+                    ),
+                )
+            )
+
+            current_width = width_out
+
+        self.trunk_output = nn.Sequential(OrderedDict(blocks))
```
datumbox, line 336: Nit: In other models this is called `features`

torchvision/models/regnet.py
```diff
@@ -0,0 +1,574 @@
...
+
+        self.avgpool = nn.AdaptiveAvgPool2d((1, 1))
+        self.fc = nn.Linear(in_features=current_width, out_features=num_classes)
```
datumbox, line 339: Nit: If some models (not all), we have a classifier attribute ([see here](https://github.com/pytorch/vision/blob/main/torchvision/models/mobilenetv3.py#L154-L160)).

datumbox (approved):
LGTM, thanks @kazhang.

The comments I left are optional nits or discussion points, none of them are blocking. IMO you should be able to merge this once you add the weights.

What's the plan for the pre-trained models? There are lots of new variants being added, do you foresee having weights for all of them prior the release? Alternatively, do you propose releasing the methods without the weights (it's not common but it [has been done before](https://github.com/pytorch/vision/blob/main/torchvision/models/mnasnet.py#L11-L18))? Another option is to offer builder methods only for models that have weights.

docs/source/models.rst
```diff
@@ -94,6 +109,20 @@ These can be constructed by passing ``pretrained=True``:
...
+    regnet_y_8gf = models.regnet_y_8gf(pretrained=True)
+    regnet_y_16gf = models.regnet_y_16gf(pretrained=True)
+    regnet_y_32gf = models.regnet_y_32gf(pretrained=True)
+    regnet_x_400mf = models.regnet_x_400mf(pretrained=True)
+    regnet_x_800mf = models.regnet_x_800mf(pretrained=True)
+    regnet_x_1_6gf = models.regnet_x_1_6gf(pretrained=True)
+    regnet_x_3_2gf = models.regnet_x_3_2gf(pretrained=True)
+    regnet_x_8gf = models.regnet_x_8gf(pretrained=True)
+    regnet_x_16gf = models.regnet_x_16gf(pretrained=True)
+    regnet_x_32gf = models.regnet_x_32gf(pretrained=True)
```
datumbox, line 125: Before merge remember to remove from this list all models for which we don't offer weights. :)

datumbox:
@kazhang I did a final check and the PR looks good to go for me. We can follow up for the nits and todos on another PR.

Two quick questions:
- Your implementation contains classes such as `SimpleStemIN`, `BottleneckTransform` etc which I understand are not declared on the `__all__` variable. This is a very common pattern in TorchVision BTW. I assume that you consider these classes not part of the public API but still made it not fully private (by prepending an `_`) because other developers might be interested in using them. Can you confirm this is the case? 
- Have you validated all deployed pre-trained weights by running the references script with the `--test-only --pretrained` arguments? This is a final step we usually do just prior merging to confirm the accuracies reported on docs are accurate and that we are good to go. See the bottom of the description of PR https://github.com/pytorch/vision/pull/3182 for an example.

kazhang:
@datumbox yes, `SimpleStemIN`, `BottleneckTransform` are common building blocks in RegNet, other developers might want to use them when exploring other variants of RegNet.

kazhang:
I've tested the pretrained model on ImageNet:
| model | Acc@1 | Paper Acc@1 | Acc@5 | command |
| --- | --- | --- | --- | --- |
| regnet_y_400mf | 74.024 | 74.1 | 91.68 | srun -p $TRAIN_QUEUE --cpus-per-task=96 --gpus-per-node=8 python -m torch.distributed.launch --nproc_per_node=8 --use_env train.py --model regnet_y_400mf --batch-size 128 --test-only --pretrained |
| regnet_y_800mf | 76.42 | 76.3 | 93.136 | srun -p $TRAIN_QUEUE --cpus-per-task=96 --gpus-per-node=8 python -m torch.distributed.launch --nproc_per_node=8 --use_env train.py --model regnet_y_800mf --batch-size 128 --test-only --pretrained |
| regnet_y_8gf | 79.966 | 79.9 | 95.1 | srun -p $TRAIN_QUEUE --cpus-per-task=96 --gpus-per-node=8 python -m torch.distributed.launch --nproc_per_node=8 --use_env train.py --model regnet_y_8gf --batch-size 64 --test-only --pretrained |
| regnet_x_400mf | 72.83 | 72.7 | 90.95 | srun -p $TRAIN_QUEUE --cpus-per-task=96 --gpus-per-node=8 python -m torch.distributed.launch --nproc_per_node=8 --use_env train.py --model regnet_x_400mf --batch-size 128 --test-only --pretrained |
| regnet_x_800mf | 75.19 | 75.2 | 92.418 | srun -p $TRAIN_QUEUE --cpus-per-task=96 --gpus-per-node=8 python -m torch.distributed.launch --nproc_per_node=8 --use_env train.py --model regnet_x_800mf --batch-size 128 --test-only --pretrained |
| regnet_x_8gf | 79.324 | 79.3 | 94.694 | srun -p $TRAIN_QUEUE --cpus-per-task=96 --gpus-per-node=8 python -m torch.distributed.launch --nproc_per_node=8 --use_env train.py --model regnet_y_8gf --batch-size 64 --test-only --pretrained |

Issue #4391, opened 2021-09-10
Investigate Exponential Moving Average result in classification script


prabhat00155:
### 🐛 Describe the bug

Verify that the EMA added to classification script in https://github.com/pytorch/vision/pull/4381 works as expected.

### Versions

N/A

cc @datumbox

datumbox:
The script might not be working as expected in all use-cases. Here is the results I get by running the following:
```
$ PYTHONPATH=$PYTHONPATH:`pwd` python -u run_with_submitit.py --ngpus 8 --nodes 1 --model mobilenet_v3_large --epochs 600 --opt rmsprop --batch-size 128 --lr 0.064 --wd 0.00001 --lr-step-size 2 --lr-gamma 0.973 --auto-augment ta_wide --random-erase 0.2 --model-ema
Submitted job_id: 46058186


$ grep "Acc@1" ./46058186/46058186_0_log.out
 * Acc@1 5.300 Acc@5 15.888
 * Acc@1 0.126 Acc@5 0.490
 * Acc@1 16.598 Acc@5 37.364
 * Acc@1 0.120 Acc@5 0.496
 * Acc@1 24.016 Acc@5 47.174
 * Acc@1 0.122 Acc@5 0.502
 * Acc@1 27.202 Acc@5 51.712
 * Acc@1 0.114 Acc@5 0.494
 * Acc@1 33.306 Acc@5 59.140
 * Acc@1 0.112 Acc@5 0.498
 * Acc@1 34.138 Acc@5 60.178
 * Acc@1 0.108 Acc@5 0.512
 * Acc@1 36.414 Acc@5 62.486
 * Acc@1 0.104 Acc@5 0.528
 * Acc@1 37.536 Acc@5 64.172
 * Acc@1 0.102 Acc@5 0.516
 * Acc@1 40.744 Acc@5 66.738
 * Acc@1 0.100 Acc@5 0.528
 * Acc@1 40.974 Acc@5 67.410
 * Acc@1 0.100 Acc@5 0.516
 * Acc@1 40.950 Acc@5 67.206
 * Acc@1 0.102 Acc@5 0.494
 * Acc@1 41.506 Acc@5 67.552
 * Acc@1 0.104 Acc@5 0.500
 * Acc@1 39.652 Acc@5 65.506
 * Acc@1 0.106 Acc@5 0.494
 * Acc@1 44.440 Acc@5 70.518
 * Acc@1 0.104 Acc@5 0.500
[...]

PR #4406, opened 2021-09-14
Added update_parameters to EMA to fix calculation

prabhat00155:
Resolves https://github.com/pytorch/vision/issues/4391.

Output logs: 
[resnext101_32x8d_training_log_ema_fix.txt](https://github.com/pytorch/vision/files/7160747/resnext101_32x8d_training_log_ema_fix.txt)
[resnet18_training_log_ema_fix.txt](https://github.com/pytorch/vision/files/7160751/resnet18_training_log_ema_fix.txt)

cc @datumbox

### DIFF
--- a/references/classification/utils.py
+++ b/references/classification/utils.py
@@ -172,6 +172,17 @@ class ExponentialMovingAverage(torch.optim.swa_utils.AveragedModel):
                    decay * avg_model_param + (1 - decay) * model_param)
         super().__init__(model, device, ema_avg)
 
+    def update_parameters(self, model):
+        for p_swa, p_model in zip(self.module.state_dict().values(), model.state_dict().values()):
+            device = p_swa.device
+            p_model_ = p_model.detach().to(device)
+            if self.n_averaged == 0:
+                p_swa.detach().copy_(p_model_)
+            else:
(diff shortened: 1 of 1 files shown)


datumbox (approved):
LGTM, thanks Prabhat.

As @kazhang this is an issue on PyTorch core. Ideally the base class should include all state not just the params. We use a similar approach in other averaging schemes (see [this](https://github.com/pytorch/vision/blob/main/references/classification/utils.py#L271)) so this is aligned to what we've seen in the past. Though this workaround will do for our use-case, I think it's still worth raising it on PyTorch core and either correcting it or introducing better control over which params we should consider.

prabhat00155:
> this

Makes sense. Will follow-up on this.

prabhat00155:
> LGTM, thanks Prabhat.
> 
> As @kazhang this is an issue on PyTorch core. Ideally the base class should include all state not just the params. We use a similar approach in other averaging schemes (see [this](https://github.com/pytorch/vision/blob/main/references/classification/utils.py#L271)) so this is aligned to what we've seen in the past. Though this workaround will do for our use-case, I think it's still worth raising it on PyTorch core and either correcting it or introducing better control over which params we should consider.

Makes sense. Will follow-up on this.

PR #4407, opened 2021-09-14
Updated model-ema-decay default to 0.9

prabhat00155:
Updated default value of model-ema-decay to a less extreme value.

cc @datumbox


datumbox (approved):
LGTM. I confirm I've tested this value on my side and it's reasonable for most models and learning rate schemes.

PR #4408, opened 2021-09-14
Update the metrics output on reference scripts

datumbox:
Update the metrics outputs on reference scripts to make it easier to discriminate between normal and EMA validation when we use `grep`.

cc @datumbox

### DIFF
--- a/references/classification/train.py
+++ b/references/classification/train.py
@@ -71,8 +71,7 @@ def evaluate(model, criterion, data_loader, device, print_freq=100, log_suffix='
     # gather the stats from all processes
     metric_logger.synchronize_between_processes()
 
-    print('{header} Acc@1 {top1.global_avg:.3f} Acc@5 {top5.global_avg:.3f}'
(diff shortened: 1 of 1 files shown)


prabhat00155 (approved):
Thanks @datumbox, looks good to me.

references/classification/train.py
```diff
@@ -71,8 +71,8 @@ def evaluate(model, criterion, data_loader, device, print_freq=100, log_suffix='
     # gather the stats from all processes
     metric_logger.synchronize_between_processes()
 
-    print(' * Acc@1 {top1.global_avg:.3f} Acc@5 {top5.global_avg:.3f}'
-          .format(top1=metric_logger.acc1, top5=metric_logger.acc5))
+    print('{header} Acc@1 {top1.global_avg:.3f} Acc@5 {top5.global_avg:.3f}'
```
prabhat00155, line 74: Should we use fstrings?

Issue #4402, opened 2021-09-14
`unittest_macos_cpu` segfault while running test_io.py


malfet:
### 🐛 Describe the bug

See https://app.circleci.com/pipelines/github/pytorch/vision/10319/workflows/7e41f9a8-f949-4ae7-ba1b-a3265692426d/jobs/781701 for example

The difference between passing and previously failing cases, is that in passing `torchvisoin` was built without fmmpeg support, which can be seen in https://app.circleci.com/pipelines/github/pytorch/vision/10289/workflows/0b74f684-8ba1-416b-9b7e-0448031b5138/jobs/778860?invite=true#step-108-65

And this is how backtrace looks like:
```
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x30)
  * frame #0: 0x0000000138965436 libc++.1.0.dylib`std::__1::basic_ostream<char, std::__1::char_traits<char> >::operator<<(unsigned int) + 230
    frame #1: 0x000000013951c911 video_reader.so`ffmpeg::Decoder::openStreams(this=0x00007ffeefbf67e0, metadata=0x00007ffeefbf65b0 size=0) at decoder.cpp:441:73 [opt]
    frame #2: 0x000000013951c1f5 video_reader.so`ffmpeg::Decoder::init(this=0x00007ffeefbf67e0, params=0x00007ffeefbf64d0, in=<unavailable>, metadata=<unavailable>)>&&, std::__1::vector<ffmpeg::DecoderMetadata, std::__1::allocator<ffmpeg::DecoderMetadata> >*) at decoder.cpp:388:8 [opt]
    frame #3: 0x000000013950befc video_reader.so`vision::video_reader::(anonymous namespace)::probeVideo(isReadFile=<unavailable>, input_video=<unavailable>, videoPath=<unavailable>) at video_reader.cpp:463:28 [opt]
    frame #4: 0x000000013950d8ef video_reader.so`vision::video_reader::probe_video_from_file(videoPath=<unavailable>) at video_reader.cpp:661:10 [opt]

```

### Versions

Collecting environment information...
PyTorch version: 1.10.0.dev20210913
Is debug build: False
CUDA used to build PyTorch: None
ROCM used to build PyTorch: N/A

OS: macOS 10.15.5 (x86_64)
GCC version: Could not collect
Clang version: 12.0.0 (clang-1200.0.32.2)
CMake version: Could not collect
[...]

Issue #4392, opened 2021-09-10
Video support in torchvision


prabhat00155:
This is to keep track of the major work items for video support in TorchVision in this half.

- [x] Building torchvision with ffmpeg.
- [x] Updating the documentation to clearly state the steps to use video api.
- [x] Benchmarking performance.
- [x] [Video GPU Decoding](https://github.com/pytorch/vision/issues/2439)

vadimkantorov:
It would be nice that ffmpeg support features two README items/examples:
1. How to install a working torchvision + ffmpeg combination from conda
2. How to compile torchvision from sources with ffmpeg (with two ffmpeg options: one is installable from conda; another is ffmpeg also built from sources in the most basic variant, for decoding it should be enough)

bjuncek:
Regarding benchmarking:
I've done the first steps towards it in this [repo](https://github.com/bjuncek/video_reader_benchmark).

The main for me are that:
1) in straight-up video reading, it is not obviously clear that GPU decoding is actually faster than CPU decoding [1]. In chat with people, it seems like there is benefit of GPU decoding in end-to-end pipelines where decoded frames can be directly consumed by the GPU. But that raises the questions whether we can actually support this bc our transforms are afaik done on CPUs, so you'd not gain much. Similar things have been observed by Mike from pyAV as well in their docs [2]

![test](https://user-images.githubusercontent.com/6318056/135487200-7fdbfed0-38f0-47d9-85d1-3d296fb60afa.png)

2) As with our VideoReader and older `read_video` APIs, each come with their set of advantages and trade-offs. You can simply check the `demo.ipynb` in the repo above to see that different readers have a different approach to how the API should look like. It is also useful to note that some libraries (namely decord), actually count on the _approximations_ in order to implement the VideoReader API that allowes getting of the _k_-th frame. Many filetypes (for example stock HMDB51 videos) do not contain `extradata` metadata; when approximating number of frames, we found that up to 4% of the frames can be missed/duplicated in these situations. 




[1] note, these comparisons were done on the machine with above-average CPU and what used to be quite competitive GPU. Running it on different hardware would probably provide more results. 

[2] https://pyav.org/docs/develop/overview/about.html : see section "Unsuported features"

vadimkantorov:
(I think there is also a push for more and more transforms supporting GPU and relying on just PyTorch ops)

vadimkantorov:
[...]

PR #4405, opened 2021-09-14
Disabling the video tests that seg fault

datumbox:
The specific tests cause the CI to seg fault on MacOS. Skipping it temporarily until https://github.com/pytorch/vision/issues/4402 is resolved, to enable the execution of the remaining tests.

cc @pmeier

PR #4410, opened 2021-09-14
Ignore platforms other than Linux for ffmpeg

prabhat00155:
Resolves https://github.com/pytorch/vision/issues/4402.
Part of https://github.com/pytorch/vision/issues/4392.


setup.py
```diff
@@ -351,6 +351,8 @@ def get_extensions():
 
     ffmpeg_exe = distutils.spawn.find_executable('ffmpeg')
     has_ffmpeg = ffmpeg_exe is not None
+    if sys.platform != 'linux':
+        has_ffmpeg = False
```
datumbox, line 355: For my info, do we need to disable also for Windows?

datumbox (approved):
LGTM @prabhat00155 in principle. I'm approving and I'll let you merge once you are happy with the CI and close the superseded PR at #4405.

setup.py
prabhat00155: I tried it on Windows once and saw some failure. After fixing that, I'll re-enable it here.

PR #4411, opened 2021-09-14
Warmup schedulers in References

### LINKED ISSUE #4281: Update reference scripts to use the "Batteries Included" utils
## 🚀 Feature
As part of the "Batteries Included" initiative (#3911) we are adding a number of new utils that can be used to produce SOTA results. Once those utils are landed, we should update our reference scripts to use them. 

More specifically we need to:
- [ ] Update the reference scripts to use PyTorch's new [linear warmup](https://github.com/pytorch/pytorch/pull/60836) scheduler.
  - The reference scripts should be updated to let users define the `warmup_method` (if any), the `warmup_iters` and the `warmup_factor`. The warmup scheduler should be chained with other existing schedulers.
  - The Object Detection recipe needs to be BC compatible and replace our [custom linear warmup](https://github.com/pytorch/vision/blob/dc42f933f3343c76727dbfba6e4242f1bcb8e1a0/references/detection/engine.py#L20-L24) approach with the one from PyTorch.
  - Similarly the Video Classification recipe needs also to be BC compatible and replace our [customer scheduler](https://github.com/pytorch/vision/blob/master/references/video_classification/scheduler.py#L5).
  - All other recipes (Classification and Segmentation), should be updated to optionally use warm up. The addition should be BC compatible and turned off by default.
- [x] Update the Classification reference to use PyTorch's new label smoothing implementation.
  - We will support the case of label smoothing by using the upcoming `label_smoothing` argument of CrossEntropyLoss (https://github.com/pytorch/pytorch/pull/63122). 
  - A new `label_smoothing` argument should be introduced on the [script](https://github.com/pytorch/vision/blob/c2ab0c59f42babf9ad01aa616cd8a901daac86dd/references/classification/train.py#L165). To ensure BC, its default value should be 0.0 (no label smoothing).

datumbox:
Resolves #4281

Adds warmup on the following recipes:
- [x] classification
- [x] detection
- [x] segmentation
- [x] video classification

This PR maintains the location where we call `scheduler.step()` in each recipe. Segmentation and Video classification do it on the iteration level, Classification does it on the epoch level and Detection does it in a hybrid.

Though doing it on the iteration level provides more flexibility, doing so will have slight effects on the reproducibility of existing models. These effects should be minor and largely overshadowed by other differences across runs (such as the randomness of the initialization scheme). The only reason I'm not doing the switch here is because it requires extra work which I'm deferring for when we will start retraining the models using the new utils of Batteries Included.


references/segmentation/train.py
```diff
@@ -133,9 +133,30 @@ def main(args):
         params_to_optimize,
         lr=args.lr, momentum=args.momentum, weight_decay=args.weight_decay)
 
-    lr_scheduler = torch.optim.lr_scheduler.LambdaLR(
+    iters_per_epoch = len(data_loader)
+    main_lr_scheduler = torch.optim.lr_scheduler.LambdaLR(
         optimizer,
-        lambda x: (1 - x / (len(data_loader) * args.epochs)) ** 0.9)
+        lambda x: (1 - x / (iters_per_epoch * (args.epochs - args.lr_warmup_epochs))) ** 0.9)
```
fmassa, line 139: Created an issue to track if we can now use a stock PyTorch scheduler for this https://github.com/pytorch/vision/issues/4438

fmassa (approved):
LGTM, thanks!

Just double-checking, did you run the new schedulers in a loop to compare before / after results?

datumbox:
@fmassa Thanks!

Yes I did. They match really close (to the 5th decimal) most of the times. For reference here are the scripts used to test this:

```
Segmentation:
python -m torch.distributed.launch --nproc_per_node=2 --use_env train.py --dataset coco --model lraspp_mobilenet_v3_large --epochs 7 --lr-warmup-epochs 2

Classification:
python -m torch.distributed.launch --nproc_per_node=8 --use_env train.py --model mobilenet_v3_large --data-path /datasets01/tinyimagenet/081318/ --epochs 100 --lr-scheduler cosineannealinglr --lr 0.004 --lr-warmup-epochs 10

Detection:
python -m torch.distributed.launch --nproc_per_node=8 --use_env train.py --dataset coco --model ssdlite320_mobilenet_v3_large --epochs 2 --lr-scheduler cosineannealinglr --lr 0.1 --batch-size 12

Video:
python -m torch.distributed.launch --nproc_per_node=8 --use_env train.py --data-path=/private/home/vvryniotis/kinetics400 --train-dir=train --val-dir=val --batch-size=1 --cache-dataset --epochs 10 --lr-warmup-epochs 2 --lr-milestones 4 6 8
```

PR #4412, opened 2021-09-15
Register ops to AutocastCPU

CaoE:
* modify the directory structure: moved the autocast files from torchvision/csrc/ops/autocast/ to torchvision/csrc/ops/autocast/cuda

* add the cpu directory under the autocast directory;

* register deform_conv2d, nms, ps_roi_align, ps_roi_pool, roi_align, and roi_pool to AutocastCPU.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -133,9 +133,10 @@ def get_extensions():
         os.path.join(extensions_dir, "ops", "*.cpp")
     )
     source_cpu = (
-        glob.glob(os.path.join(extensions_dir, "ops", "autograd", "*.cpp"))
-        + glob.glob(os.path.join(extensions_dir, "ops", "cpu", "*.cpp"))
-        + glob.glob(os.path.join(extensions_dir, "ops", "quantized", "cpu", "*.cpp"))
+        glob.glob(os.path.join(extensions_dir, 'ops', 'autograd', '*.cpp')) +
+        glob.glob(os.path.join(extensions_dir, 'ops', 'autocast', 'cpu', '*.cpp')) +
+        glob.glob(os.path.join(extensions_dir, 'ops', 'cpu', '*.cpp')) +
+        glob.glob(os.path.join(extensions_dir, 'ops', 'quantized', 'cpu', '*.cpp'))
     )
 
     is_rocm_pytorch = False
@@ -163,7 +164,7 @@ def get_extensions():
     else:
         source_cuda = glob.glob(os.path.join(extensions_dir, "ops", "cuda", "*.cu"))
 
-    source_cuda += glob.glob(os.path.join(extensions_dir, "ops", "autocast", "*.cpp"))
+    source_cuda += glob.glob(os.path.join(extensions_dir, 'ops', 'autocast', 'cuda', '*.cpp'))
 
     sources = main_file + source_cpu
     extension = CppExtension
--- a/setup.py
+++ b/setup.py
@@ -133,10 +133,10 @@ def get_extensions():
         os.path.join(extensions_dir, "ops", "*.cpp")
     )
     source_cpu = (
-        glob.glob(os.path.join(extensions_dir, 'ops', 'autograd', '*.cpp')) +
-        glob.glob(os.path.join(extensions_dir, 'ops', 'autocast', 'cpu', '*.cpp')) +
-        glob.glob(os.path.join(extensions_dir, 'ops', 'cpu', '*.cpp')) +
-        glob.glob(os.path.join(extensions_dir, 'ops', 'quantized', 'cpu', '*.cpp'))
+        glob.glob(os.path.join(extensions_dir, "ops", "autograd", "*.cpp"))
+        + glob.glob(os.path.join(extensions_dir, "ops", "autocast", "cpu", "*.cpp"))
+        + glob.glob(os.path.join(extensions_dir, "ops", "cpu", "*.cpp"))
+        + glob.glob(os.path.join(extensions_dir, "ops", "quantized", "cpu", "*.cpp"))
     )
 
     is_rocm_pytorch = False
@@ -164,7 +164,7 @@ def get_extensions():
     else:
         source_cuda = glob.glob(os.path.join(extensions_dir, "ops", "cuda", "*.cu"))
 
-    source_cuda += glob.glob(os.path.join(extensions_dir, 'ops', 'autocast', 'cuda', '*.cpp'))
+    source_cuda += glob.glob(os.path.join(extensions_dir, "ops", "autocast", "cuda", "*.cpp"))
 
     sources = main_file + source_cpu
     extension = CppExtension
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -65,11 +65,11 @@ class RoIOpTester(ABC):
         gradcheck(func, (x,))
         gradcheck(script_func, (x,))
 
-    @pytest.mark.parametrize('device', cpu_and_gpu())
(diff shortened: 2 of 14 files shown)


test/test_ops.py
```diff
@@ -71,6 +71,12 @@ def test_autocast(self, x_dtype, rois_dtype):
         with torch.cuda.amp.autocast():
             self.test_forward(torch.device("cuda"), contiguous=False, x_dtype=x_dtype, rois_dtype=rois_dtype)
 
+    @pytest.mark.parametrize('x_dtype', (torch.float, torch.half))
+    @pytest.mark.parametrize('rois_dtype', (torch.float, torch.half))
+    def test_autocast_cpu(self, x_dtype, rois_dtype):
```
NicolasHug, line 76: instead of creating a new test, maybe we could just parametrize over the device with the `cpu_and_gpu()` function?
Since the context manager is device-dependent, we could just set it in the code, like

```py
cm = torch.cpu.amp.autocast if device == 'cpu' else torch.cuda.amp.autocast
with cm():
    self.test_forward(torch.device(device), contiguous=False, x_dtype=x_dtype, rois_dtype=rois_dtype)
```

We could do the same for the rest of the newly introduced tests

NicolasHug (commented):
Thanks @CaoE , I just took a brief look. For now I'm just curious, what is you use-case for supporting autocast on CPU?

test/test_ops.py
CaoE: Yeah, good idea, I will modify it like this.

CaoE:
Because we want to train nlp models on cpu servers, but torchvision ops like nms will report errors when using BFloat16.

test/test_ops.py
```diff
@@ -514,12 +516,22 @@ def test_nms_cuda(self, iou, dtype=torch.float64):
...
-    @needs_cuda
+    @pytest.mark.parametrize('device', cpu_and_gpu())
     @pytest.mark.parametrize("iou", (.2, .5, .8))
-    @pytest.mark.parametrize("dtype", (torch.float, torch.half))
-    def test_autocast(self, iou, dtype):
-        with torch.cuda.amp.autocast():
-            self.test_nms_cuda(iou=iou, dtype=dtype)
+    @pytest.mark.parametrize("dtype", (torch.bfloat16, torch.half))
+    def test_autocast(self, device, iou, dtype):
+        def test_nms_cpu(iou, dtype):
```
NicolasHug, line 523: Instead of creating this new one here, do you think we could just rely on `test_nms_ref` instead?
It doesn't accept a `dtype` parameter so we could add it if it's relevant, or alternatively we can just define `test_fn` as a partial function.

No strong opinion on this, we can also leave as is, even though it's a bit unfortunate that we have to define a new `test_nms_cpu` here

NicolasHug (approved):
Thanks @CaoE , this LGTM provided that the tests go green. There is a minor linting issue for now, would you mind fixing it? 

`test/test_ops.py:524:17: E117 over-indented`

I will let @fmassa take a look before merging, also possibly @datumbox since you're more familiar with the src structure etc.

setup.py
```diff
@@ -170,7 +171,7 @@ def get_extensions():
     else:
         source_cuda = glob.glob(os.path.join(extensions_dir, 'ops', 'cuda', '*.cu'))
 
-    source_cuda += glob.glob(os.path.join(extensions_dir, 'ops', 'autocast', '*.cpp'))
+    source_cuda += glob.glob(os.path.join(extensions_dir, 'ops', 'autocast', 'cuda', '*.cpp'))
```
datumbox, line 174: Buck will need changes due to this. Might be worth bringing the PR in FBcode prior merging to ensure it does not break anything.

datumbox (approved):
The changes on the C++ overall look good to me. I would advise testing them on FBcode and updating buck before merging to ensure we won't break anything internally.

My only concern, and it's not linked to this PR, is the amount of duplicate and boilerplate code we are forced to add in the repo to handle the registrations.

@ezyang I know that a year ago you've been working on improving the op registration. Is there a better way to do it?

test/test_ops.py
CaoE: Creating this new one here just to check whether the input is converted to float type. Is your suggestion to use `test_nms_ref` instread of `test_nms_cpu` here?  It's all ok for me. Thank you so much for your suggestion @NicolasHug

NicolasHug: >  test_nms_ref instread of test_nms_cpu here?

Yes :)

CaoE: Ok, I will modify it later.

fmassa:
If I understand it correctly, the only change between the CPU and CUDA folders is:
- the `autocast` passes the CPU device
- the registration mechanism registers it to `AutocastCPU` instead of `Autocast`

Is there any other difference, or do we expect it to potentially diverge more over time?

test/test_ops.py
```diff
@@ -64,12 +64,13 @@ def func(z):
...
 
-    @needs_cuda
+    @pytest.mark.parametrize('device', cpu_and_gpu())
     @pytest.mark.parametrize('x_dtype', (torch.float, torch.half))
     @pytest.mark.parametrize('rois_dtype', (torch.float, torch.half))
-    def test_autocast(self, x_dtype, rois_dtype):
-        with torch.cuda.amp.autocast():
-            self.test_forward(torch.device("cuda"), contiguous=False, x_dtype=x_dtype, rois_dtype=rois_dtype)
+    def test_autocast(self, device, x_dtype, rois_dtype):
+        cm = torch.cpu.amp.autocast if device == 'cpu' else torch.cuda.amp.autocast
```
ezyang, line 71: Someone should file a bug asking for torch.amp.autocast that accepts a device argument lol

ezyang:
> @ezyang I know that a year ago you've been working on improving the op registration. Is there a better way to do it?

I think the most direct way to reduce boilerplate is if you can reuse the templates/macros that are used inside PyTorch core to setup autocasting. I haven't checked if you are doing unusual autocasting so that these aren't applicable. We don't have codegen support for autocasting so that's right out, and no one has written a boxed fallback for autocasting (maybe someone should!)

datumbox:
@CaoE Are we good to merge?

CaoE:
> @CaoE Are we good to merge?

yes, I have nothing else to commit.

datumbox:
Awesome, let us run some additional tests on the internal FB infra and we will merge right after.

datumbox (changes requested):
Marking this as "changes needed" to avoid accidental merges prior testing it on FBcode.

@CaoE No further action is needed on your side, you are good to go. Just give us a bit more time to test things :)

fmassa (changes requested):
Marking as "request changes" so that we don't forget to answer my questions on the past post.
While we can have duplicate code here initially, it would be good for us to understand if we plan to keep those two sets of files practically identical or if further changes that can lead to divergence in the CPU / CUDA files are expected to happen.

This is important because we can think about follow-up work that will refactor this redundancy out, improving maintainance cost in the future, but only if we know that the changes will be kept to the ones I pointed out in my previous comment.

CaoE:
> If I understand it correctly, the only change between the CPU and CUDA folders is:
> 
> * the `autocast` passes the CPU device
> * the registration mechanism registers it to `AutocastCPU` instead of `Autocast`

Yes.

> Is there any other difference, or do we expect it to potentially diverge more over time?

If a new op is added to vision, it may also need to be registered to `AutocastCPU` and `Autocast` respectively. There should be no change in other parts.

Issue #3367, opened 2021-02-10
Segmentation fault for Python 3.9 for video


seemethere:
## To Reproduce

Build torchvision for the latest version using python 3.9 with the latest pytorch nightly, run `pytest -v test/test_video.py`.

Output:
```
vision39 ❯ pytest -v test/test_video.py
================================================================================ test session starts =================================================================================
platform linux -- Python 3.9.1, pytest-6.2.2, py-1.10.0, pluggy-0.12.0 -- /home/eliuriegas/miniconda3/envs/vision39/bin/python
cachedir: .pytest_cache
rootdir: /home/eliuriegas/work/vision
plugins: cov-2.11.1
collected 3 items                                                                                                                                                                    

test/test_video.py::TestVideo::test_metadata PASSED                                                                                                                            [ 33%]
test/test_video.py::TestVideo::test_read_video_tensor Fatal Python error: Segmentation fault

Current thread 0x00007f618a834740 (most recent call first):
  File "/home/eliuriegas/work/vision/torchvision/io/__init__.py", line 123 in __next__
  File "/home/eliuriegas/work/vision/test/test_video.py", line 296 in test_read_video_tensor
  File "/home/eliuriegas/miniconda3/envs/vision39/lib/python3.9/unittest/case.py", line 550 in _callTestMethod
  File "/home/eliuriegas/miniconda3/envs/vision39/lib/python3.9/unittest/case.py", line 593 in run
  File "/home/eliuriegas/miniconda3/envs/vision39/lib/python3.9/unittest/case.py", line 653 in __call__
  File "/home/eliuriegas/miniconda3/envs/vision39/lib/python3.9/site-packages/_pytest/unittest.py", line 321 in runtest
  File "/home/eliuriegas/miniconda3/envs/vision39/lib/python3.9/site-packages/_pytest/runner.py", line 162 in pytest_runtest_call
  File "/home/eliuriegas/miniconda3/envs/vision39/lib/python3.9/site-packages/pluggy/callers.py", line 187 in _multicall
  File "/home/eliuriegas/miniconda3/envs/vision39/lib/python3.9/site-packages/pluggy/manager.py", line 78 in <lambda>
  File "/home/eliuriegas/miniconda3/envs/vision39/lib/python3.9/site-packages/pluggy/manager.py", line 87 in _hookexec
  File "/home/eliuriegas/miniconda3/envs/vision39/lib/python3.9/site-packages/pluggy/hooks.py", line 289 in __call__
[...]

PR #4417, opened 2021-09-15
Skip building torchvision with ffmpeg when python==3.9

prabhat00155:
Building torchvision with ffmpeg 4.3 causes seg fault with python 3.9. Initial debugging shows crash in ffmpeg's sws_scale(). Until this issue is resolved, it's better to avoid building torchvision with ffmpeg for python 3.9.

cc @pmeier

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -351,10 +351,10 @@ def get_extensions():
 
     ffmpeg_exe = distutils.spawn.find_executable('ffmpeg')
     has_ffmpeg = ffmpeg_exe is not None
-    # Building torchvision with ffmpeg on MacOS or with Python 3.9 causes
-    # crash. See the following GitHub issues for more details.
-    # https://github.com/pytorch/pytorch/issues/65000
-    # https://github.com/pytorch/vision/issues/3367
+    # FIXME: Building torchvision with ffmpeg on MacOS or with Python 3.9
+    # FIXME: causes crash. See the following GitHub issues for more details.
+    # FIXME: https://github.com/pytorch/pytorch/issues/65000
+    # FIXME: https://github.com/pytorch/vision/issues/3367
     if sys.platform != 'linux' or (
             sys.version_info.major == 3 and sys.version_info.minor == 9):
         has_ffmpeg = False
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -351,7 +351,12 @@ def get_extensions():
 
     ffmpeg_exe = distutils.spawn.find_executable('ffmpeg')
     has_ffmpeg = ffmpeg_exe is not None
-    if sys.platform != 'linux':
+    # Building torchvision with ffmpeg on MacOS or with Python 3.9 causes
```
NicolasHug, line 354: Should we add a FIXME or a TODO comment? It makes it easy to find out all outstanding issues with a simple `git grep command`

NicolasHug (approved):
Thank you @prabhat00155 , I think this is indeed the safest option for now. LGTM. Do you think we should reopen https://github.com/pytorch/vision/issues/3367?

setup.py
```diff
@@ -351,7 +351,12 @@ def get_extensions():
...
+    # crash. See the following GitHub issues for more details.
+    # https://github.com/pytorch/pytorch/issues/65000
+    # https://github.com/pytorch/vision/issues/3367
+    if sys.platform != 'linux' or (
+            sys.version_info.major == 3 and sys.version_info.minor == 9):
```
NicolasHug, line 359: BTW I think we can remove the ffmpeg version checks below as well, right?

setup.py
prabhat00155: Yeah, makes sense.

setup.py
prabhat00155: https://github.com/pytorch/vision/pull/4413 does that. After merging this PR, #4413 should be good to merge.

fmassa:
One thing it would be good to understand is why only Python 3.9 is failing. IIUC the python version has no influence on the ffmpeg version we are installing, so maybe this is an issue with our binding code (torchbind?)

prabhat00155:
> One thing it would be good to understand is why only Python 3.9 is failing. IIUC the python version has no influence on the ffmpeg version we are installing, so maybe this is an issue with our binding code (torchbind?)

Yeah, makes sense. I've created this issue: https://github.com/pytorch/vision/issues/4430 to track this, will debug this further.

PR #4418, opened 2021-09-15
Improve FX node naming

alexander-soare:
For now this is just a draft for feedback.

The motivation is as follows:

Currently I can do something like:

```python
import torch
from pprint import pprint

class SubModule(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.relu = torch.nn.ReLU()
        
    def forward(self, x):
        x = x + 1
        x = self.relu(x)
        return x


    def __init__(self):
        super().__init__()
        self.submodule = SubModule()
        self.relu = torch.nn.ReLU()
        
    def forward(self, x):
        x = self.submodule(x)
        x = x + 1
        x = x + 1
        x = self.relu(x)
        x = self.relu(x)
        return x

model = MyModule()

train_nodes, _ = get_graph_node_names(model)
pprint(train_nodes)
```

which will give `['x', 'submodule.add', 'submodule.relu', 'add_1', 'add_2', 'relu', 'relu_1']`

As intended, the `relu` gets an incremented postfix **within the scope** of the root module. BUT the `add` get's incremented over a global scope. The reason for the discrepancy is that the module postfix mechanism was separately added here, but the function postfix was adopted from core `torch.fx`.

This behaviour is not great because you get things like this (example from `timm.models.xcit`).

```
['cls_attn_blocks.0.attn.matmul_48', 'cls_attn_blocks.1.attn.matmul_50']
```
where really it would be nicer to have:

```
['cls_attn_blocks.0.attn.matmul', 'cls_attn_blocks.1.attn.matmul']
```
and because we're referring to a different scope, there's no ambiguity about which node is being referred to. To reiterate, the desired behaviour is implemented for call_module nodes.

Going back to my toy example, the proposed change would produce:

```
['x', 'submodule.add', 'submodule.relu', 'add', 'add_1', 'relu', 'relu_1']
```


alexander-soare:
A note on this, it's not something I definitely want to be merged. More of a request for discussion on whether we should. If we agree  that it should, I'll tidy it up and double check the regex before it's ready.

fmassa:
Hi,

So, I think the name scheme for FX is not bond to be stable across different PyTorch versions (@jamesr66a can correct me on this though).

Given that we are mapping prefixes to our nodes corresponding to the module they come from, I think it makes sense to do what you suggest and also specify the counter on the local scope level.
So I like the proposed approach here and would be happy to have it move forward.
One thing to keep in mind is that we would need a bit more documentation for the naming scheme as we would deviate a bit more from `torch.fx` naming conventions (but I think in a good way)

alexander-soare:
@fmassa thanks for the feedback. Soon I'll go ahead and close it out with docs.

alexander-soare:
@fmassa ready for review whenever is convenient. Btw I don't suppose those test failures are related to this?

torchvision/models/feature_extraction.py
```diff
@@ -297,58 +301,55 @@ def create_feature_extractor(
...
 
-    A note on node specification: For the purposes of this feature extraction
-    utility, a node name is specified as a `.` seperated path walking the
-    hierarchy from top level module down to leaf operation or leaf module. For
-    instance `blocks.5.3.bn1`. The keys of the `return_nodes` argument should
-    point to either a node's name, or some truncated version of it. For
-    example, one could provide `blocks.5` as a key, and the last node with
-    that prefix will be selected. :func:`get_graph_node_names` is a useful
-    helper function for getting a list of node names of a model.
+    Desired output nodes must be specified as a ``.`` seperated
```
fmassa, line 304:
```suggestion
    Desired output nodes must be specified as a ``.`` separated
```

docs/source/feature_extraction.rst
```diff
@@ -30,6 +30,39 @@ The `torch.fx documentation <https://pytorch.org/docs/stable/fx.html>`_
...
 the inner workings of the symbolic tracing.
 
+.. _about-node-names:
+
+**About Node Names**
+
+In order to specify which nodes should be output nodes for extracted
+features, one should be familiar with the node naming convention used here
+(which differs slightly from that used in ``torch.fx``). A node name is
+specified as a ``.`` seperated path walking the module hierarchy from top level
```
fmassa, line 40:
```suggestion
specified as a ``.`` separated path walking the module hierarchy from top level
```

fmassa (approved):
Sorry for the delay in reviewing, this slipped through my review list.

Changes LGTM, let's rebase and see if the failures disappear

PR #4433, opened 2021-09-16
Add prototype namespace and datasets

pmeier:


torchvision/prototype/datasets/__init__.py
```diff
@@ -0,0 +1,5 @@
+from ._home import *
```
mthrok, line 1: I thought asterisk import is prohibited by `flake8`. Are there such many symbols to be imported here?

pmeier: > I thought asterisk import is prohibited by `flake8`. 

Yes, but we blanket ignore `F403`:

https://github.com/pytorch/vision/blob/1140ecf237c8287a315fd293e0a1bb5f710e115a/setup.cfg#L12

I'm in the process of fixing that, but I think that is ok in `__init__.py` files under the condition that the modules define an `__all__`. This way all the information about what is exposed is contained in the module itself.

> Are there such many symbols to be imported here?

Nope, but this way the file stays clean. Following my advice above the namespace also stays clean. The only issue with the style used here is that by looking at the `__init__.py` we don't know where something is defined.

pmeier: After some more thought, I think you are right. We should be explicit there.

pmeier:
@mthrok We are going to split this PR up into more reviewable chunks. Do you want me to cc you there?

mthrok:
> @mthrok We are going to split this PR up into more reviewable chunks. Do you want me to cc you there?

sure, thanks.

PR #4436, opened 2021-09-17
make test folder a package

pmeier:
This solves two issues:

1. Without this we cannot import other modules within subfolders, e.g. using anything from `test/common_utils.py` in `test/prototype/*`. With this change, we can do `from test.common_utils import foo` from anywhere in our test suite.
2. `pytest` requires unique module names in non-package folders. That means, we cannot have `test/protototype/datasets/test_utils.py` if `test/test_utils.py` already exists. See [this CI run](https://app.circleci.com/pipelines/github/pytorch/vision/10495/workflows/0305f2ed-1620-47ce-9ada-2872e9ac6f35/jobs/796347) from #4433 for an example.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -1,4 +1,4 @@
-from common_utils import IN_CIRCLE_CI, CIRCLECI_GPU_NO_CUDA_MSG, IN_FBCODE, IN_RE_WORKER, CUDA_NOT_AVAILABLE_MSG
+from test.common_utils import IN_CIRCLE_CI, CIRCLECI_GPU_NO_CUDA_MSG, IN_FBCODE, IN_RE_WORKER, CUDA_NOT_AVAILABLE_MSG
 import torch
 import pytest
 
--- a/test/datasets_utils.py
+++ b/test/datasets_utils.py
@@ -19,7 +19,7 @@ import torch
 import torchvision.datasets
 import torchvision.io
 
-from common_utils import get_tmp_dir, disable_console_output
+from test.common_utils import get_tmp_dir, disable_console_output
 
 
 __all__ = [
--- a/test/test_backbone_utils.py
+++ b/test/test_backbone_utils.py
@@ -12,7 +12,7 @@ from torchvision.models._utils import IntermediateLayerGetter
 
 import pytest
 
-from common_utils import set_rng_seed
+from test.common_utils import set_rng_seed
 
 
 def get_available_models():
--- a/test/test_datasets.py
+++ b/test/test_datasets.py
@@ -14,7 +14,7 @@ import xml.etree.ElementTree as ET
 import zipfile
 
 import PIL
-import datasets_utils
+from test import datasets_utils
 import numpy as np
 import pytest
 import torch
--- a/test/test_datasets_samplers.py
+++ b/test/test_datasets_samplers.py
@@ -13,7 +13,7 @@ from torchvision.datasets.samplers import (
 from torchvision.datasets.video_utils import VideoClips, unfold
 from torchvision import get_video_backend
 
-from common_utils import get_list_of_videos, assert_equal
+from test.common_utils import get_list_of_videos, assert_equal
 
 
 @pytest.mark.skipif(not io.video._av_available(), reason="this test requires av")
--- a/test/test_datasets_video_utils.py
+++ b/test/test_datasets_video_utils.py
@@ -6,7 +6,7 @@ import pytest
 from torchvision import io
 from torchvision.datasets.video_utils import VideoClips, unfold
 
-from common_utils import get_list_of_videos, assert_equal
+from test.common_utils import get_list_of_videos, assert_equal
 
 
 class TestVideo:
--- a/test/test_datasets_video_utils_opt.py
+++ b/test/test_datasets_video_utils_opt.py
@@ -1,6 +1,6 @@
 import unittest
 from torchvision import set_video_backend
-import test_datasets_video_utils
+from test import test_datasets_video_utils
 
 # Disabling the video backend switching temporarily
 # set_video_backend('video_reader')
--- a/test/test_functional_tensor.py
+++ b/test/test_functional_tensor.py
@@ -14,7 +14,7 @@ import torchvision.transforms.functional as F
 import torchvision.transforms as T
 from torchvision.transforms import InterpolationMode
 
-from common_utils import (
+from test.common_utils import (
     cpu_and_gpu,
     needs_cuda,
     _create_data,
--- a/test/test_image.py
+++ b/test/test_image.py
@@ -9,7 +9,7 @@ import numpy as np
 import torch
 from PIL import Image, __version__ as PILLOW_VERSION
 import torchvision.transforms.functional as F
-from common_utils import needs_cuda, assert_equal
+from test.common_utils import needs_cuda, assert_equal
 
 from torchvision.io.image import (
     decode_png, decode_jpeg, encode_jpeg, write_jpeg, decode_image, read_file,
--- a/test/test_io.py
+++ b/test/test_io.py
@@ -9,7 +9,7 @@ from torchvision import get_video_backend
 import warnings
 from urllib.error import URLError
 
-from common_utils import assert_equal
+from test.common_utils import assert_equal
 
 
 try:
--- a/test/test_io_opt.py
+++ b/test/test_io_opt.py
@@ -1,6 +1,6 @@
 import unittest
 from torchvision import set_video_backend
-import test_io
+from test import test_io
 
 
 # Disabling the video backend switching temporarily
--- a/test/test_models.py
+++ b/test/test_models.py
@@ -1,8 +1,8 @@
 import os
 import io
 import sys
-from common_utils import map_nested_tensor_object, freeze_rng_state, set_rng_seed, cpu_and_gpu, needs_cuda
-from _utils_internal import get_relative_path
+from test.common_utils import map_nested_tensor_object, freeze_rng_state, set_rng_seed, cpu_and_gpu, needs_cuda
+from test._utils_internal import get_relative_path
 from collections import OrderedDict
 import functools
 import operator
--- a/test/test_models_detection_anchor_utils.py
+++ b/test/test_models_detection_anchor_utils.py
@@ -1,5 +1,5 @@
 import torch
-from common_utils import assert_equal
+from test.common_utils import assert_equal
 from torchvision.models.detection.anchor_utils import AnchorGenerator, DefaultBoxGenerator
 from torchvision.models.detection.image_list import ImageList
 import pytest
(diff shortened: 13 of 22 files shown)


test/conftest.py
```diff
@@ -1,4 +1,4 @@
-from common_utils import IN_CIRCLE_CI, CIRCLECI_GPU_NO_CUDA_MSG, IN_FBCODE, IN_RE_WORKER, CUDA_NOT_AVAILABLE_MSG
+from test.common_utils import IN_CIRCLE_CI, CIRCLECI_GPU_NO_CUDA_MSG, IN_FBCODE, IN_RE_WORKER, CUDA_NOT_AVAILABLE_MSG
```
NicolasHug, line 1: Shouldn't all imports withing the `test` folder be relative imports?

NicolasHug (approved):
Thanks @pmeier , this LGTM but I'll have to import this into fbcode before merging to make sure this doesn't toast the internal infra.... I'll get back to you

test/conftest.py
pmeier: Do we have a rule for when to use relative vs. absolute imports I should follow? Otherwise it makes no difference. Can change if you like.

NicolasHug: I've always used only-relative imports for imports within a package... I don't even remember why.
If anything, it should make the import sorting more straightforward?

fmassa (approved):

NicolasHug (changes requested):
Just marking so we don't accidentally merge before checking fbcode. Also as Philip noted on slack the onnx import failures seem legit

pmeier:
The problem lies in the fact that the test is invoked with `python test/test_onnx.py`. While this invokes `pytest` internally

https://github.com/pytorch/vision/blob/a2b4c652e1a0c8678e951c6c50174615d6ba101f/test/test_onnx.py#L500-L501

the `test` package is never imported and thus importing from submodules of it (regardless of absolute or relative) fails.

So the question is, if we want to retain the ability to run tests by invoking the module with `python` directly? After our migration to `pytest` I see no point. In the example above it functions as drop-in replacement: `pytest test/test_onnx.py`

fmassa:
Let's just assume that tests are run with `pytest` only then

mthrok:
I recommend to introduce another directory in `test` as a top-level unit test module for the following reasons.
1. Having `torchvision` and `test` (the top level test directory) forces to have `torchvision` from the repository root precedence when `import torchvision`.
This is fine for the regular development flow, but sometimes, there are occasions one wants to run unit tests on the installations from binary distribution, such as nightly and release candidates. In this case, having test module top directory inside of `test` directory comes handy because `cd test` and the tests will run on installed versions, instead of source directory. Otherwise one has to delete `torchvision` directory first to run tests.
2. There is a standard module called [`test`](https://docs.python.org/3/library/test.html), though it is not really used. I do not know if the existing test frameworks use this module, but having `test/__init__.py` and invoking it from the top level might cause a problem.

NicolasHug:
These are good points @mthrok , thanks. Pytest actually has a few guidelines on this: https://docs.pytest.org/en/6.2.x/goodpractices.html#choosing-a-test-layout-import-rules

Their recommendation would be to move `torchvision` into `src/torchvision` instead.

This makes me wonder whether this is worth it or not. Would it be possible to have a `test` folder within the `prototype` package instead? I guess you wouldn't be able to access the same common test utils though

pmeier:
> This makes me wonder whether this is worth it or not. Would it be possible to have a test folder within the prototype package instead? I guess you wouldn't be able to access the same common test utils though

Yes, I think it is possible. Currently, I don't need access to anything, but that might be a problem in the future.

mthrok:
> Their recommendation would be to move `torchvision` into `src/torchvision` instead.

I considered doing so for audio since it would not be difficult but I was afraid that it might give degraded developer experience to folks work on multiple repos.

I also see that [data](https://github.com/facebookexternal/torchdata/pull/141) is following the pattern.
It would be best, if all PyTorch projects were structured that way from the beginning.

mthrok:
> > This makes me wonder whether this is worth it or not. Would it be possible to have a test folder within the prototype package instead? I guess you wouldn't be able to access the same common test utils though
> 
> Yes, I think it is possible. Currently, I don't need access to anything, but that might be a problem in the future.

With that approach, I assume that CI test script needs to be updated so as to run the tests in CI. Prototype features can come and go. I think it's best to reduce the number of places one need to update.

pmeier:
> With that approach, I assume that CI test script needs to be updated so as to run the tests in CI. Prototype features can come and go. I think it's best to reduce the number of places one need to update.

We could have a separate job to run `pytest torchvision/prototype` similar to our onnx and hub tests. Maybe we can even only trigger it if something in there changes. In any case if the prototype test fails on an unrelated PR, we know that the error can be ignored.

NicolasHug:
I'm not sure I understand the concern @mthrok , is this about test discovery?

I think we would just need to change

```diff
-pytest --cov=torchvision --junitxml=test-results/junit.xml -v --durations 20 test --ignore=test/test_datasets_download.py
+pytest --cov=torchvision --junitxml=test-results/junit.xml -v --durations 20 test torchvision --ignore=test/test_datasets_download.py
```

in `run_test.sh`

pmeier:
> I think we would just need to change

Nope, we don't need to change anything to run the "normal" test suite. `pytest` is set up to only look for tests in `test/` so we don't need to ignore `torchvision/`:

https://github.com/pytorch/vision/blob/1140ecf237c8287a315fd293e0a1bb5f710e115a/pytest.ini#L9-L10

NicolasHug:
I'm assuming that we want all CIs to run both the tests in `test` and in `torchvision/whatever/test`?

That's why I'm saying we just need to add `torchvision` explicitly so that the CI doesn't omit any of the `torchvision/whatever/test` folders.

Or better then, edit `testpaths` to also include torchvision and we don't need to change anything else.

pmeier:
There are three ways forward here:

1. Make `test/` a package, which would solve both issues that prompted this PR. As @mthrok noted, this will prohibit us from using the [`test` package](https://docs.python.org/3/library/test.html), but since the documentation reads: 

    > The test package is meant for internal use by Python only. It is documented for the benefit of the core developers of Python. Any use of this package outside of Python’s standard library is discouraged as code mentioned here can change or be removed without notice between releases of Python.

    we shouldn't be using that anyway.
2. `pytest` also supports a layout in which the tests exist alongside the source, e.g. `torchvision/prototype/datasets/test/*`. That would solve the issue of duplicated names, but would prohibit us from using anything from common test utilities unless we also move them inside `torchvision`.
3. We could also add a new folder `test_prototype` that would have the same effect and downsides as 2.

For all three variants we can have a separate prototype workflow with minimal effort.

mthrok:
> For all three variants we can have a separate prototype workflow with minimal effort.

I think 1 is the minimal effort for the long term. It does not prevent the prototype test from using the common test utilities, it does not require any change to the CI, and it has an added benefit of running the tests for package installations.

I feel that 2. is going against the already established code organization, without justification why it has to be that way. And one has to remember to enforce it every time a new prototype is added, why the tests for prototype have to be placed alongside of code, unlike the rest of the code. Once you have a good justification of why it has to be that way, you probably need to document it somewhere.

NicolasHug:
Something that I missed earlier and that @pmeier clarified in our chat is that the main goal here is to separate our regular tests from the prototype tests, because the prototype tests are much more subject to fail than the rest, and we want to separate these signals in the CI.

With that in mind, I'm not sure I see how option 1. really solves this?

pmeier:
> With that in mind, I'm not sure I see how option 1. really solves this?

If we go for a `test/prototype/` folder, the only thing we would need to do, is to add another `--ignore` [here](https://github.com/pytorch/vision/blob/74ac47e3c3944be1e91281cb9685f907d9177512/.circleci/unittest/linux/scripts/run_test.sh#L10) and in all other `run_test.sh` scripts that point directly to `test/`. The new workflow would simply invoke `pytest test/prototype`.

NicolasHug:
Thanks @pmeier 

> I feel that 2. is going against the already established code organization, without justification why it has to be that way. And one has to remember to enforce it every time a new prototype is added, why the tests for prototype have to be placed alongside of code, unlike the rest of the code. Once you have a good justification of why it has to be that way, you probably need to document it somewhere.

I understand the constraint, but this seems to apply to all 3 proposed solutions? None of them completely follow the current code organization I would say.


My personal (decreasing) order of preference would be 3 -> 2 -> 1. Option 1. would require to create an additional folder, whether it's `test/something/` or `src/torchvision` which seems more disruptive that the rest (in terms of git blame implications, changes to different workflows, fbcode, etc.)

pmeier:
> Option 1. would require to create an additional folder

Every option will add an folder:

1. `test/prototype/`
2. `torchvision/prototype/test/` (or nested deeper)
3. `test_prototype/`

> or `src/torchvision`

Note that this will not solve the issue we are trying to solve here, since it only changes how `torchvision` is imported, but not the test collection.

NicolasHug:
> Every option will add an folder:

yes but option 1., i.e. making `test` a package, will require to either move `torchvision` to  `src/torchvision` (https://github.com/pytorch/vision/pull/4436#issuecomment-921754441), or add another directory in `test` (https://github.com/pytorch/vision/pull/4436#issuecomment-921732337).

This is the extra directory that I'm taking about

pmeier:
I've had an offline discussion with @NicolasHug about his concerns. The takeaway is, that as long as we install `torchvision` in `--editable` / `develop` mode, it makes no difference if we import `torchvision` from our local folder or from the `site-packages`. 

Since we do that in CI and also recommend this in our contributing guide, option 1. can be adapted without changes. This PR already implements it and minus all the unrelated CI failures it works out fine.

pmeier:
Unfortunately, `--import-mode=importlib` implies that no local dependencies are allowed. There are some ideas to change that in pytest-dev/pytest#8964, but no consensus or PR yet. I think we could fairly easy write a small plugin that simply imports `test/` before we start collection. Otherwise, I stand by my order of preference above.

fmassa:
I propose a different option to move forward here: keep the hierarchy flat, add test files like `test_prototype_datasets.py`, and use decorators to skip those tests if `torchvision.prototype` is not available.

fmassa (changes requested):
See my previous comment

datumbox:
Thanks everyone for the input on this. It's great to see that lots of thought was put on the solution. I wonder if we can make the necessary changes and unblock this PR? 

I'm not going to pitch in on which direction we will follow as I think you got enough options. I just want to make sure that the prototype area has an agreed initial way to support tests because we plan to introduce more soon. If we all agree, we could adopt the quick an easy approach mentioned by Francisco at https://github.com/pytorch/vision/pull/4436#issuecomment-925830399 and then provide a better approach as we improve our test suite?

PR #4439, opened 2021-09-17
Fix size_t issues across JPEG versions and platforms

datumbox:
This PR fixes (again) the size_t issues we've been facing due to JPEG. It seems that overnight, the version on windows changed from `9b-hb83a4c4_2` to `9d-h2bbff1b_0`. It seems that the Windows version has fixed the issue from 9d>= and aligned itself with the ones of the rest of the platforms.

To ensure that our patch works with all versions, I pinned and tested jpeg on the following versions:
- [x] jpeg <=9b
- [x] jpeg >=9d (same as unpinned)

We skip tests for `jpeg ==9c` because we get `Unsatisfiable dependencies` [errors](https://app.circleci.com/pipelines/github/pytorch/vision/10529/workflows/b0ca6994-710c-4b03-8ae6-6db3d0b4712a/jobs/799540) in conda.

### DIFF
--- a/torchvision/csrc/io/image/cpu/encode_jpeg.cpp
+++ b/torchvision/csrc/io/image/cpu/encode_jpeg.cpp
@@ -15,11 +15,7 @@ torch::Tensor encode_jpeg(const torch::Tensor& data, int64_t quality) {
 #else
 // For libjpeg version <= 9b, the out_size parameter in jpeg_mem_dest() is
 // defined as unsigned long, whereas in later version, it is defined as size_t.
-// For windows backward compatibility, we define JpegSizeType as different types
-// according to the libjpeg version used, in order to prevent compilation
-// errors.
-#if defined(_WIN32) || !defined(JPEG_LIB_VERSION_MAJOR) || \
-    JPEG_LIB_VERSION_MAJOR < 9 ||                          \
+#if !defined(JPEG_LIB_VERSION_MAJOR) || JPEG_LIB_VERSION_MAJOR < 9 || \
     (JPEG_LIB_VERSION_MAJOR == 9 && JPEG_LIB_VERSION_MINOR <= 2)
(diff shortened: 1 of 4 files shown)


torchvision/csrc/io/image/cpu/encode_jpeg.cpp
```diff
@@ -15,11 +15,7 @@ torch::Tensor encode_jpeg(const torch::Tensor& data, int64_t quality) {
 #else
 // For libjpeg version <= 9b, the out_size parameter in jpeg_mem_dest() is
 // defined as unsigned long, whereas in later version, it is defined as size_t.
-// For windows backward compatibility, we define JpegSizeType as different types
-// according to the libjpeg version used, in order to prevent compilation
-// errors.
-#if defined(_WIN32) || !defined(JPEG_LIB_VERSION_MAJOR) || \
```
fmassa, line 21: No more Windows checks? Ok with me, but IIUC there used to be discrepancies between Windows / Linux libjpeg that we might need to take into account?

fmassa (approved):
Approving conditioned on green CI.

I have a question about removing Windows that might be good to double-check

torchvision/csrc/io/image/cpu/encode_jpeg.cpp
datumbox: The situation [seems](https://app.circleci.com/pipelines/github/pytorch/vision/10521/workflows/d7d80f03-bef3-4e0f-a587-bc274097ed25/jobs/798797) to have been resolved in later versions of Windows (9d>=) hence blunt windows filter needs to go. I'll pin different versions of JPEG in a bit to see what exactly is the right condition we should use here.

datumbox:
The `cmake_windows_gpu` is unrelated, see #4440.

PR #4441, opened 2021-09-17
add prototype image folder dataset

pmeier:
This is the precursor of the new style datasets. Example:

```python

dataset, categories = datasets.from_image_folder(...)
dataset = dataset.shuffle().decode_sample(datasets.decoder.pil, "image")

for sample in dataset:
    ...
```

cc @pmeier


pmeier:
@fmassa Should we have a `prototype` label for these PRs to easily tell them apart from other stuff for the release notes?

torchvision/prototype/datasets/decoder.py
```diff
@@ -0,0 +1,40 @@
...
+
+import numpy as np
+import PIL.Image
+import torch
+from torch.utils.data import IterDataPipe, functional_datapipe
+
+__all__ = ["pil"]
+
+
+@functional_datapipe("decode_sample")
```
fmassa, line 13: I've mentioned this to @VitalyFedyunin in the past, but I think that the fluent API should be used with care.

The reason is that things like
```python
import torch.utils.data
import torchvision.prototype

generic_datapipe.decode_sample(...)
```
works but
```python
import torch.utils.data

generic_datapipe.decode_sample(...)
```
doesn't can be very surprising to users, and it is very very hard to spot what is wrong / missing.

IMO I'm not sure this is a pattern we would like to encourage outside of PyTorch core abstractions.
Thoughts?

pmeier:
One problem I see is that we cannot decode until after the `.shuffle()` call: it will read `buffer_size` samples ([defaults to 10000](https://github.com/pytorch/pytorch/blob/9157a2889f9e17dda1be6b0b137ab1801ba253a9/torch/utils/data/datapipes/iter/combinatorics.py#L79)) into memory before a single sample is returned to the user. 

Given that in most of the cases the user will want to shuffle the dataset, I'm wondering if we even want to provide an option for decoding in the future API. If we don't we need to figure out a good way to pass on the information which keys of the sample need to be decoded from the dataset to the user.

torchvision/prototype/datasets/_folder.py
```diff
@@ -0,0 +1,44 @@
...
+    for label, category in enumerate(categories):
+        category_dp = FileLister(str(root / category), recursive=True)
+        category_dp = FileLoader(category_dp)
+        category_dp = Mapper(
+            category_dp,
+            _collate,
+            fn_kwargs=dict(label=label, category=category),
+        )
+        category_dps.append(category_dp)
+    return Concater(*category_dps), categories
```
fmassa, line 44: Doesn't this make shuffling much harder, because the elements will be sampled from the same category in order? Or does the shuffler have additional knowledge that knows how to handle those use-cases in a special way?

torchvision/prototype/datasets/_folder.py
```diff
@@ -0,0 +1,44 @@
...
+def _collate(
+    data: Tuple[str, io.IOBase],
+    *,
+    label: int,
+    category: str,
+) -> Dict[str, Any]:
+    path, image_file = data
+    return dict(
+        path=path,
+        image=image_file,
```
fmassa, line 22: nit: maybe calling this `image` might be misleading because the image hasn't been decoded yet?

fmassa (commented):
That's a great start, thanks!

I have a few questions for my understanding, let me know what you think

fmassa:
> @fmassa Should we have a prototype label for these PRs to easily tell them apart from other stuff for the release notes?

Yes, good idea, can you create one?

torchvision/prototype/datasets/decoder.py
pmeier: Agreed. Apart from the availability, this is also hard to discover. Unless you grep for `decode_sample`, you probably won't find the `decode_sample` datapipe.

datumbox:
I'm brining up a question I raised previously concerning the name of this space. A few were proposed over time, is `prototype` the selected term across all domain libs?

fmassa:
@datumbox torchaudio follows `prototype` as well, and torchtext had traditionally used `experimental`. IIRC we had an agreement to use `prototype` in a past discussion

torchvision/prototype/datasets/_folder.py
pmeier: > Doesn't this make shuffling much harder, because the elements will be sampled from the same category in order?

Yes, but I have no idea how to improve this. Maybe we can read one sample from each directory before we start at the top again. Or we implement a `RandomChoser` which picks one element at random from multiple datapipes.

fmassa: Can't we just have a single `FileLister` which is recursive? This way it makes the job of a `Shuffler` much easier, as FileLister could implement a specific property that allows for perfect shuffling for those family of pipes.

@VitalyFedyunin @ejguan thoughts?

torchvision/prototype/datasets/_folder.py
pmeier: Valid question. IMO decoding should be an in-place operation and since in most cases there will be some decoding at some point, I feel like "image" is ok as key. Plus, otherwise the user would need to specify the name of the decoded feature together with the feature to be decoded, e.g.

```python
dataset.decode_sample(datasets.decoder.pil, ("image_file", "image")
```

torchvision/prototype/datasets/_folder.py
pmeier: > Can't we just have a single `FileLister` which is recursive?

We can, yes. As far as I see, that would make sorting harder, since we now need to extract the label and category from each path instead of just setting them. I'll implement it that way and benchmark.

torchvision/prototype/datasets/_folder.py
fmassa: We can revisit this in the early future, so let's not block on this.

torchvision/prototype/datasets/decoder.py
fmassa: So let's remove this from torchvision and be more verbose on our end, wdyt?

pmeier:
Workflow now looks like:

```python

dataset, categories = datasets.from_image_folder(...)
decode = datasets.decoder.decode_sample(datasets.decoder.pil, "image")

dataset = dataset.shuffle().map(decode)

for sample in dataset:
    ...

```

torchvision/prototype/datasets/decoder.py
pmeier: See https://github.com/pytorch/vision/pull/4441#issuecomment-921851202

torchvision/prototype/datasets/_folder.py
pmeier: If we use

```python
class RandomPicker(IterDataPipe):
    def __init__(self, *datapipes):
        self.datapipes = datapipes

    def __iter__(self):
        non_exhausted = [iter(dp) for dp in self.datapipes]
        while non_exhausted:
            dp = random.choice(non_exhausted)
            try:
                yield next(dp)
            except StopIteration:
                non_exhausted.remove(dp)
```

instead of `Concater` we get a pseudo-shuffled dataset with minimal overhead.

VitalyFedyunin:
Hi! As we likely going to have shuffle topic again, I captured notes in the doc file. Feel free to comment/feedback. https://docs.google.com/document/d/15RzQtCMl2FDtu9loZH5a-wxHDPnSpT0LRT6wMjR1p-U/edit

torchvision/prototype/datasets/datapipes.py
```diff
@@ -0,0 +1,20 @@
+"""This is a temporary module for datapipes that are needed now, but should be absorbed by torchdata in the future."""
+
+import random
+from typing import Any
+
+from torch.utils.data import IterDataPipe
+
+
+class RandomPicker(IterDataPipe):
```
VitalyFedyunin, line 9: https://github.com/facebookexternal/torchdata/blob/7881a93c6f83aa5c6a129c2123a42cf23c7dddd9/torchdata/datapipes/iter/__init__.py#L29 might help

torchvision/prototype/datasets/_folder.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def from_image_folder(
+    root: Union[str, pathlib.Path], *, pseudo_shuffle: bool = True
+) -> Tuple[IterDataPipe, List[str]]:
+    root = pathlib.Path(root).expanduser().resolve()
+    categories = sorted({item.name for item in os.scandir(root) if item.is_dir})
+    category_dps = []
+    category_dp: IterDataPipe
+    for label, category in enumerate(categories):
```
VitalyFedyunin, line 38: This is interesting approach to avoid scanning twice to obtain list of categories.

torchvision/prototype/datasets/datapipes.py
pmeier: I'll test when `torchdata` is public.

torchvision/prototype/datasets/_folder.py
pmeier: Interesting as in good or bad interesting?

pmeier:
I've refactored the loading process to be able to shuffle directly after the files are enumerated. The workflow now looks like this:

```python

dataset = datasets.from_image_folder(root)
```

By default, the images will now be shuffled with an infinite buffer, i.e. a true random permutation. To turn this off one can pass `datasets.from_image_folder(..., shuffler=None)`.

`datasets.from_image_folder(root)` is now equivalent to the current `datasets.ImageFolder(root, transform=transforms.ToTensor())`. Testing this on Caltech256 data, leads to almost identical warm-up times and the datapipe implementation is even a little faster for completely consuming the dataset.

torchvision/prototype/datasets/_folder.py
```diff
@@ -0,0 +1,74 @@
...
+from torch.utils.data import IterDataPipe
+from torch.utils.data.datapipes.iter import FileLister, Concater, FileLoader, Mapper, Shuffler
+
+from torchvision.prototype.datasets.decoder import pil
+
+
+__all__ = ["from_data_folder", "from_image_folder"]
+
+# pseudo-infinite buffer size until a true infinite buffer is supported
+INFINITE = 1_000_000_000
```
pmeier, line 18: @VitalyFedyunin @ejguan We need the ability to specify an infinite buffer, i.e. `buffer_size=None` for every datapipe that implements a buffer.

torchvision/prototype/datasets/decoder.py
```diff
@@ -0,0 +1,12 @@
...
+import numpy as np
+import PIL.Image
+import torch
+
+__all__ = ["pil"]
+
+
+def pil(file: io.IOBase, mode="RGB") -> torch.Tensor:
+    image = PIL.Image.open(file).convert(mode.upper())
+    return torch.from_numpy(np.array(image, copy=True)).permute((2, 0, 1))
```
fmassa, line 12: FYI this could be replaced by `torchvision.transforms.functional.pil_to_tensor`

torchvision/prototype/datasets/decoder.py
```diff
@@ -0,0 +1,12 @@
...
+
+import numpy as np
+import PIL.Image
+import torch
+
+__all__ = ["pil"]
+
+
+def pil(file: io.IOBase, mode="RGB") -> torch.Tensor:
+    image = PIL.Image.open(file).convert(mode.upper())
```
fmassa, line 11: Do we want to close the file, or doesn't it make a difference? Maybe using a context manager here?

fmassa (approved):
Let's get this merged.

I have some comments that can be addressed in a follow-up PR

torchvision/prototype/datasets/decoder.py
pmeier: IIRC, we don't need to. `.convert()` does this automatically.

torchvision/prototype/datasets/_folder.py
```diff
@@ -0,0 +1,81 @@
...
+        data=data,
+        label=label,
+        category=category,
+    )
+
+
+def from_data_folder(
+    root: Union[str, pathlib.Path],
+    *,
+    shuffler: Optional[Callable[[IterDataPipe], IterDataPipe]] = lambda dp: Shuffler(dp, buffer_size=INFINITE),
```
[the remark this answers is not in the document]
pmeier: I think not shuffling by default is ok. Although everyone will probably turn it on to do anything useful with the dataset, we now also require users to turn it on manually.

Issue #4333, opened 2021-08-29
[RFC] API For Common Layers In Torchvision


oke-aditya:
Picks up from discussion in https://github.com/pytorch/vision/pull/4293#discussion_r696471536

## 🚀 Feature

API for Commonly used Layers in building models.

## Motivation

A huge code duplication is involved in building very basic blocks for large neural networks. Some of these blocks can be standardized and re-used. Also these could be offered to end user as an API so that downstream libraries can build models easily.

E.g. for duplication are SqueezeExcitation, ConvBNRelu, etc.

## Pitch

Create an API called `torchvision.nn` or `torchvision.layers`
Our implementations need to be generic but not locked to certain activation functions or channels, etc.
These can be simply classes based on `nn.Module`.

An example

```
class ConvNormAct(nn.Module):
    def __init__(
        self,
        in_planes: int,
        out_planes: int,
        kernel_size: int = 3,
        stride: int = 1,
        groups: int = 1,
        dilation: int = 1,
        norm_layer: Optional[Callable[..., nn.Module]] = None,
        activation_layer: Optional[Callable[..., nn.Module]] = None,
    ) -> None: 

        super().__init__()
        #  Thsi could be even a nn.Sequential Block instead.
        padding = (kernel_size - 1) // 2 * dilation
        self.conv1 = nn.Conv2d(in_planes, out_planes, kernel_size, stride, padding, dilation, groups, bias=False)
        self.norm_layer = norm_layer
        self.activation = activation_layer

    def forward(x: Tensor):
         x = self.conv1(x)
         if self.norm_layer is not None:
              x = self.norm_layer(x)
         x = self.activation(x)
        return x

```

User can use this as

```
from torchvision import layers
l1 = layers.ConvNormAct(3, 10, norm_layer=nn.BatchNorm2d(), activation_layer=nn.Relu(inplace=True))
dummy_input = torch.randn(3, 224, 224)
out = l1(dummy_input)
```

Also layers can be mixed into new custom models.

E.g.

```
class SimpleModel(nn.Module):
    def __init__():
        self.l1 = layers.ConvNormAct(3, 10, activation_layer=nn.Relu(inplace=True))
        self.l2  = layers.ConvNormAct(10, 3, activation_layer=nn.Relu(inplace=True))

    def forward(x: Tensor):
        x = self.l1(x)
        x = self.l2(x)
        return x
```

### We have torchvision.ops then why layers? 

[...]

PR #4448, opened 2021-09-19
Allow custom activation in SqueezeExcitation of EfficientNet

kazhang:
It would be great to reuse SqueezeExcitation in RegNet implementation(#4403).

cc @datumbox

### DIFF
--- a/torchvision/models/efficientnet.py
+++ b/torchvision/models/efficientnet.py
@@ -37,18 +37,20 @@ class SqueezeExcitation(nn.Module):
         input_channels: int,
         squeeze_channels: int,
         activation: Callable[..., nn.Module] = nn.ReLU,
+        scale_activation: Callable[..., nn.Module] = nn.Sigmoid,
     ) -> None:
         super().__init__()
         self.fc1 = nn.Conv2d(input_channels, squeeze_channels, 1)
         self.fc2 = nn.Conv2d(squeeze_channels, input_channels, 1)
         self.activation = activation()
+        self.scale_activation = scale_activation()
 
     def _scale(self, input: Tensor) -> Tensor:
         scale = F.adaptive_avg_pool2d(input, 1)
         scale = self.fc1(scale)
         scale = self.activation(scale)
         scale = self.fc2(scale)
-        return scale.sigmoid()
+        return self.scale_activation(scale)
 
     def forward(self, input: Tensor) -> Tensor:
         scale = self._scale(input)
--- a/torchvision/models/efficientnet.py
+++ b/torchvision/models/efficientnet.py
@@ -32,15 +32,23 @@ model_urls = {
 
 
 class SqueezeExcitation(nn.Module):
-    def __init__(self, input_channels: int, squeeze_channels: int):
+    def __init__(
+        self,
+        input_channels: int,
+        squeeze_channels: int,
+        activation: Optional[Callable[..., nn.Module]] = None,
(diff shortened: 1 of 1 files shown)


torchvision/models/efficientnet.py
```diff
@@ -32,15 +32,23 @@
...
+        self,
+        input_channels: int,
+        squeeze_channels: int,
+        activation: Optional[Callable[..., nn.Module]] = None,
+    ) -> None:
         super().__init__()
         self.fc1 = nn.Conv2d(input_channels, squeeze_channels, 1)
         self.fc2 = nn.Conv2d(squeeze_channels, input_channels, 1)
+        if activation is None:
+            activation = nn.SiLU
```
datumbox, line 45: Given the original Squeeze-and-Excitation [paper](https://arxiv.org/pdf/1709.01507.pdf) uses ReLU, perhaps that's a better default option for when we move it out to the `ops` namespace?

torchvision/models/efficientnet.py
```diff
@@ -32,15 +32,23 @@
...
+        self.activation = activation(inplace=True)
```
datumbox, line 46: I would probably drop `inplace` from here because not all activation methods in PyTorch support the attribute (for example `Sigmoid`).

torchvision/models/efficientnet.py
```diff
@@ -32,15 +32,23 @@
...
 
     def _scale(self, input: Tensor) -> Tensor:
         scale = F.adaptive_avg_pool2d(input, 1)
         scale = self.fc1(scale)
-        scale = F.silu(scale, inplace=True)
+        scale = self.activation(scale)
         scale = self.fc2(scale)
         return scale.sigmoid()
```
datumbox, line 53: [MobileNetV3](https://github.com/pytorch/vision/blob/main/torchvision/models/mobilenetv3.py#L35) uses `hardsigmoid` here, so we probably need another option to allow someone replacing sigmoid with another activation.

torchvision/models/efficientnet.py
```diff
@@ -32,15 +32,23 @@
 
 
 class SqueezeExcitation(nn.Module):
```
datumbox, line 34: I'm still on the fence on whether we should mark this class with `_` to explicitly denote it's private (I believe that's what you did on your other PR) or leave as is for when we move it to `ops`.  Thoughts?

Would you consider making this PR replace ALL existing SE layers? If yes, we could move this straight to `ops`. Happy to help with converting existing models in a BC manner.

datumbox (commented):
Thanks for the PR @kazhang. This is much needed and will allow us to reuse the layer in multiple places. I left a couple of comments, let me know what you think.

torchvision/models/efficientnet.py
kazhang: > I'm still on the fence on whether we should mark this class with _ to explicitly denote it's private
I'm considering to make it public since it has been published for a while and still popular (https://paperswithcode.com/method/squeeze-and-excitation-block)

 > Would you consider making this PR replace ALL existing SE layers? If yes, we could move this straight to ops. Happy to help with converting existing models in a BC manner.
I would like to get the consensus in #4333 first and execute carefully. I'm thinking to merge this PR first to avoid the third SE implementation and then consolidate the two into one.

datumbox (approved):
LGTM, thanks @kazhang!

PR #4461, opened 2021-09-22
[NO_MERGE] Test if 0.10 release is failing

datumbox:


NicolasHug:
The pytorch version that's being used is the nightly, which explains why the hub tests are failing

> .Using cache found in /home/circleci/.cache/torch/hub/pytorch_vision_main

torchhub relies on the main branch since https://github.com/pytorch/pytorch/pull/64364 was merged, but the test (which should be updated) still thinks it's the master branch.

Shouldn't we be testing 0.10 against the 1.9.1 release instead? https://github.com/pytorch/pytorch/releases/tag/v1.9.1

PR #4463, opened 2021-09-22
Build with ffmpeg only on Linux

prabhat00155:


bjuncek:
Hmm. 
wouldn't we want to be able to build it on a macOS? 
If I understand correctly, the linking segfault is gone when pytorch is compiled from source; maybe we can just add that to the documentation? 

Or even better, maybe add a flag that would be off by default (on a mac or not) so that we can do 
`WITH_FFMPEG=1 python setup.py isntall`

prabhat00155:
> Hmm.
> wouldn't we want to be able to build it on a macOS?
> If I understand correctly, the linking segfault is gone when pytorch is compiled from source; maybe we can just add that to the documentation?
> 
> Or even better, maybe add a flag that would be off by default (on a mac or not) so that we can do
> `WITH_FFMPEG=1 python setup.py isntall`

This PR is against the release branch.

Issue #4430, opened 2021-09-16
read_video_from_file() causes seg fault with Python 3.9


prabhat00155:
### 🐛 Describe the bug

https://app.circleci.com/pipelines/github/pytorch/vision/10411/workflows/5783f083-26e2-4209-b9c6-976de289d637/jobs/791355

```
Program received signal SIGSEGV, Segmentation fault.
0x00007f87b3031d22 in ?? () from /root/project/env/lib/libswscale.so.5
Missing separate debuginfos, use: debuginfo-install glibc-2.17-324.el7_9.x86_64
(gdb) bt
#0  0x00007f87b3031d22 in ?? () from /root/project/env/lib/libswscale.so.5
#1  0x00007f87b3030be4 in ?? () from /root/project/env/lib/libswscale.so.5
#2  0x00007f87b3006af4 in sws_scale () from /root/project/env/lib/libswscale.so.5
#3  0x00007f87b4ab1932 in ffmpeg::(anonymous namespace)::transformImage(SwsContext*, unsigned char const* const*, int*, ffmpeg::VideoFormat, ffmpeg::VideoFormat, unsigned char*, unsigned char**, int*) [clone .isra.9] ()
   from /root/project/torchvision/video_reader.so
#4  0x00007f87b4ab1bd9 in ffmpeg::VideoSampler::sample(unsigned char const* const*, int*, ffmpeg::ByteStorage*) () from /root/project/torchvision/video_reader.so
#5  0x00007f87b4ab2ac7 in ffmpeg::VideoStream::copyFrameBytes(ffmpeg::ByteStorage*, bool) () from /root/project/torchvision/video_reader.so
#6  0x00007f87b4aaea9a in ffmpeg::Stream::decodePacket(AVPacket const*, ffmpeg::DecoderOutputMessage*, bool, bool*) () from /root/project/torchvision/video_reader.so
#7  0x00007f87b4aaa183 in ffmpeg::Decoder::processPacket(ffmpeg::Stream*, AVPacket*, bool*, bool*) () from /root/project/torchvision/video_reader.so
#8  0x00007f87b4aaa80a in ffmpeg::Decoder::getFrame(unsigned long) () from /root/project/torchvision/video_reader.so
#9  0x00007f87b4aafe81 in ffmpeg::SyncDecoder::decode(ffmpeg::DecoderOutputMessage*, unsigned long) () from /root/project/torchvision/video_reader.so
[...]

PR #4465, opened 2021-09-22
Updated ffmpeg channel to fix seg fault

prabhat00155:
Resolves https://github.com/pytorch/vision/issues/4430.
ffmpeg 4.3(latest) from pytorch channel causes seg fault. This PR replaces pytorch channel with conda-forge which works fine.


NicolasHug:
I could be wrong but  I believe the reason we don't rely on the conda-forge channel is because this would defacto make torchvision's binaries gpl-licensed.

The ffmpeg in the pytorch channel is stripped down from the GPL-licensed code.

prabhat00155:
> I could be wrong but I believe the reason we don't rely on the conda-forge channel is because this would defacto make torchvision's binaries gpl-licensed.
> 
> The ffmpeg in the pytorch channel is stripped down from the GPL-licensed code.

Does that mean we can't use it for CI? Sorry, I don't know much about how these licenses work.

NicolasHug:
> Does that mean we can't use it for CI?

We technically can use it, as in, there's no licensing issue there. The licensing issue is about what ffmpeg version we can link against in our binaries: we can't link against conda-forge.
So testing against conda-forge wouldn't give us the signal we're looking for: we should test with the ffmpeg version that our binaries will be linked against.

NicolasHug:
> we should test with the ffmpeg version that our binaries will be linked against.

Maybe my comment doesn't make much sense since the actual ffmpeg code that's eventually executed can be swapped.

But this makes me wonder why we even have a custom ffmpeg in the pytorch channel then 🤔
Sorry for the noise @prabhat00155 , perhaps @fmassa can provide more actionable feedback here

andfoy:
Also, FFmpeg 4.3 has some errors that are in a assembly section of their code, so it is not a good idea to do that. After a discussion that we had earlier this month, it was decided that the FFmpeg "linking" symbols will be provided via a proxy package that is GPL and links statically against GPL-flavored FFmpeg, so rendering our own FFmpeg outdated.

prabhat00155:
It still installs ffmpeg from pytorch channel:
```
(base) prabhatroy-mbp:notebooks prabhatroy$ ssh -p 64538 34.228.52.20
(base) [root@334f751efa90 ~]# conda activate /root/project/env
(/root/project/env) [root@334f751efa90 ~]# conda list | grep ffmpeg
ffmpeg                    4.3                  hf484d3e_0    pytorch
(/root/project/env) [root@334f751efa90 ~]# 

```

bjuncek:
Just my two cents here to get @prabhat00155 unblocked:

1. At the moment we shouldn't (and don't IIUC) link against any FFMPEG at the moment; the FFMPEG from `pytorch` channel causes segfaults, and the `conda-forge` version of 4.3 (4.3.2, _NOT_ the 4.3.1 that segfaults) has been stable so far. 

2. @NicolasHug  at the moment we don't (and shouldn't) link FFMPEG binaries until we verify that the proposal @andfoy mentions (IIRC Nikita said he will write a prototype) works. 
So for now, our default is saying to the customers: "please install your own ffmpeg (however licensed you want) and rebuild from source. This _needs_ to be changed, and then we can update CI's accordingly.

**tl;dr** enabling this only makes sure that our tests are running with the version of FFMPEG that most people will be using, regardless of the licensing. When we have a better solution, we can reflect that in CI as well.

bjuncek (approved):
Approving to unblock @prabhat00155 and further testing; should be revisited once we find a way to link against correct FFMPEG

PR #4469, opened 2021-09-23
Rewrite test and fix masks_to_boxes implementation

oke-aditya:
Fix the comments suggested in https://github.com/pytorch/vision/pull/4290#discussion_r714552490


test/test_ops.py
```diff
@@ -1000,6 +1002,43 @@ def gen_iou_check(box, expected, tolerance=1e-4):
...
+
+            return masks
+
+        image = _get_image()
+        expected = torch.tensor([[127, 2, 165, 40], [2, 50, 44, 92], [56, 63, 98, 100], [139, 68, 175, 104],
+                                 [160, 112, 198, 145], [49, 138, 99, 182], [108, 148, 152, 213]])
+
+        for dtype in [torch.int16, torch.int32, torch.int64]:
+            masks = torch.zeros((image.n_frames, image.height, image.width), dtype=dtype)
+            masks = _create_masks(image, masks)
```
oke-aditya, line 1032: One we know to create masks better way, without the image we can fix this,

torchvision/ops/boxes.py
```diff
@@ -301,24 +301,24 @@ def generalized_box_iou(boxes1: Tensor, boxes2: Tensor) -> Tensor:
 
 def masks_to_boxes(masks: torch.Tensor) -> torch.Tensor:
     """
-    Compute the bounding boxes around the provided masks
+    Compute the bounding boxes around the provided masks.
 
-    Returns a [N, 4] tensor.  Both sets of boxes are expected to be in ``(x1, y1, x2, y2)`` format with
+    Returns a [N, 4] tensor containing bounding boxes. The boxes are in ``(x1, y1, x2, y2)`` format with
```
oke-aditya, line 306: Small fix to docstring.

oke-aditya (commented):
Small comments for rew

datumbox (approved):
@oke-aditya Thanks for the PR. LGTM.

On the future, we could potentially do a couple of iterations to improve the tests of this operator. We should also look into the fact that the gallery uses an asset from tests, which is not great.

test/test_ops.py
```diff
@@ -1000,6 +1002,43 @@ def gen_iou_check(box, expected, tolerance=1e-4):
...
+            masks_box_check(masks, expected)
+
+        image2 = _get_image()
+        for dtype in [torch.float16, torch.float32, torch.float64]:
```
NicolasHug, line 1036: considering the expected tensor is just integer values and we use check_dtype=False, do we really need to test against all these dtypes? It seems to me that just torch.float would be enough?

I'm also not even sure the tensor we end-up with indeed has the desired dtype, it will depend on what numpy gives I think?

test/test_ops.py
```diff
@@ -1000,6 +1002,43 @@ def gen_iou_check(box, expected, tolerance=1e-4):
...
+    def test_masks_box(self):
+        def masks_box_check(masks, expected, tolerance=1e-4):
+            out = ops.masks_to_boxes(masks)
+            torch.testing.assert_close(out, expected, rtol=0.0, check_dtype=False, atol=tolerance)
+
+        # Check for int type boxes.
+        def _get_image():
+            assets_directory = os.path.join(os.path.dirname(os.path.abspath(__file__)), "assets")
+            mask_path = os.path.join(assets_directory, "masks.tiff")
+            image = Image.open(mask_path)
```
NicolasHug, line 1015: maybe keep the context manager for the image to be closed automatically?

torchvision/ops/boxes.py
NicolasHug: good find!

test/test_ops.py
```diff
@@ -1000,6 +1002,43 @@ def gen_iou_check(box, expected, tolerance=1e-4):
             gen_iou_check(box_tensor, expected, tolerance=0.002 if dtype == torch.float16 else 1e-3)
 
 
+class TestMasksToBoxes:
+    def test_masks_box(self):
+        def masks_box_check(masks, expected, tolerance=1e-4):
+            out = ops.masks_to_boxes(masks)
```
NicolasHug, line 1008: maybe assert that out is of float dtype, since this is the fix you just made in this PR

NicolasHug (approved):
Thanks @oke-aditya , some minor comments but LGTM

I agree with @datumbox 's box comment about not relying on the test directory in the galleries

test/test_ops.py
```diff
@@ -1000,6 +1002,43 @@ def gen_iou_check(box, expected, tolerance=1e-4):
             gen_iou_check(box_tensor, expected, tolerance=0.002 if dtype == torch.float16 else 1e-3)
 
 
+class TestMasksToBoxes:
```
0x00b1, line 1005: You're not using any class features so you can likely move the methods to the module namespace.

test/test_ops.py
oke-aditya: Hmm, this function returns the image, a with context will close it and make it unavailable for outer function to test?

test/test_ops.py
NicolasHug: We grouped some tests in this file in classes because they belong to the same operators, but it doesn't mean we should do it with everything. In this case there's only one test, so we don't really need a class. It's not a huge deal though.

>  It would be nice if the file is rewritten fully using pytest, calling fixtures, etc

It's heavily relying on pytest parametrization already :)
What fixtures are you thinking of?

test/test_ops.py
oke-aditya: I think yes, for safety as we do in other ops too E.g. have a look at TestGenBoxIou, TestBoxIou

I will defer to @datumbox

test/test_ops.py
oke-aditya: Yes, but the tests above E,g TestBoxIou, TestBoxArea do one test in a class. I would prefer code consistency.

fixture should be both the_functions, rather than calling it, pytest way is to use fixture and pass as param.

test/test_ops.py
NicolasHug: It's different I think.

`TestGenBoxIou` doesn't make any assumption on the input/output's dtype so it's supposed to be checked against all possibilities.

Here, the output is hardcoded to be a float.

Also the expected values aren't floats

oke-aditya: Hmm, correct the reason is it uses `_upcast` to type cast properly. While that is used to prevent int16 and float16 overflow.
It might not happen here, So probably, we should check only for int.

test/test_ops.py
oke-aditya: Yes, done.

test/test_ops.py
oke-aditya: Checking only float as mentioned.

test/test_ops.py
NicolasHug: > Yes, but the tests above E,g TestBoxIou, TestBoxArea do one test in a class. I would prefer code consistency.

Fair enough :)

> fixture should be both the_functions, rather than calling it, pytest way is to use fixture and pass as param.

Hm, I'm not sure I agree here. Fixtures are helpful to have a setup / teardown mechanism, and to have automatically cached function calls, which is useful e.g. if you need the same expensive resource across multiple tests in a class, or in a test module. When none of these conditions are met, there's no reason to use a fixture instead of a simple function

test/test_ops.py
datumbox: The logic of running multiple floats is necessary for ops that do math operations that can over/under-flow. Here it's not going matter too much because all operations seem fine. This operator is so lightweight that I don't think there is a significant speed benefits from removing the multiple checks.

It's definitely not hurting us having them so if someone was to remove it on a future PR, I don't see an issue. Similarly, if one was to make a case to run always all floats through all ops, I would equally be fine.

test/test_ops.py
```diff
@@ -1000,6 +1002,39 @@ def gen_iou_check(box, expected, tolerance=1e-4):
...
+                frame = np.array(image)
+                masks[index] = torch.tensor(frame)
+
+            return masks
+
+        image = _get_image()
+        expected = torch.tensor([[127, 2, 165, 40], [2, 50, 44, 92], [56, 63, 98, 100], [139, 68, 175, 104],
+                                 [160, 112, 198, 145], [49, 138, 99, 182], [108, 148, 152, 213]], dtype=torch.float)
+
+        image2 = _get_image()
```
NicolasHug, line 1031: we don't need this one, do we? I think it's possible to just use `image.n_frames, image.height, image.width` and remove `image2`?

oke-aditya:
I think I should have addressed all the issues :smile: Thanks for the detailed review

datumbox (approved):
LGTM, still. :)

I think the PR is a net positive and since it fixes the device bug, we should merge it. Happy to review follow up PRs that further improve the tests of the operator.

oke-aditya:
I think I will followup to make gallery example a bit more robust. Also see if we can do with a PNG image and PNG mask (like in PenFudan) instead of a tiff.
Also try to use our plotting utilities to do the plotting job.

NicolasHug:
Thanks @oke-aditya ! Looking forward to the follow-ups.

@datumbox should we remove the "bug" label for this PR and just use the same labels as for https://github.com/pytorch/vision/pull/4290?

If we keep "bug" there's a high change this will end up as an entry in the bugfix section in the changlog, and this won't be accurate since this is something that hadn't been released yet.

datumbox:
@NicolasHug Hmm... TLDR: Sure thing, go ahead.

The slightly longer response is that the label is accurate in a sense that this PR fixes a bug. I do agree with you that it shouldn't be reported on the changelog as we are not fixing a bug of the previous release. So I think there might be an issue with how we use the tags as there are conflicting needs for example release notes require tagging new things as bug, while tracking better engineering initiatives requires tagging all. I think that's a good topic for DAPI or Vision weekly as we are going to bump on it again and again.

NicolasHug:
I'll remove it for this one, but I agree this is something we could discuss and improve

PR #4472, opened 2021-09-23
Add JIT tests for box_area, box_iou, genaralized_box_iou

oke-aditya:
They are JIT scriptable and compatible, but it would be nice to test it I think.

cc @NicolasHug @datumbox


test/test_ops.py
```diff
@@ -999,6 +1015,14 @@ def gen_iou_check(box, expected, tolerance=1e-4):
             expected = torch.tensor([[1.0, 0.9933, 0.9673], [0.9933, 1.0, 0.9737], [0.9673, 0.9737, 1.0]])
             gen_iou_check(box_tensor, expected, tolerance=0.002 if dtype == torch.float16 else 1e-3)
 
+    def test_giou_jit(self):
+        box_tensor = torch.tensor([[0, 0, 100, 100], [0, 0, 50, 50], [200, 200, 300, 300]], dtype=torch.float)
```
oke-aditya, line 1019: I remember adding these and calculating by hand an year ago. Great that they are still used! :smile:

oke-aditya:
Hi @datumbox is this worth checking and adding (extra testing doesn't hurt in my opinion)
But let me know, I am happy to close otherwise :smiley:

datumbox (approved):
LGTM, thanks for adding them.

Perhaps on a future PR we can rethink how we split the tests. If we didn't add them on separate classes, we could easier share code and use `parametrize` to pass the operator.

fmassa (approved):
Thanks!

As @datumbox mentioned, we can probably refactor those tests to avoid some code duplication.

I've created an issue to track this down https://github.com/pytorch/vision/issues/4500

PR #4473, opened 2021-09-23
add API for new style datasets

pmeier:
This PR is the second precursor for the new style datasets. The biggest change is that we are moving away from map-like datasets, i.e. with `__getitem__`, to iter-like datasets. This has a lot of advantages (and of course also some disadvantages) that I'm not going to list here, but rather in a post when all of this is officially announced.

## How do we use the new style datasets?

For users the biggest change is that datasets are no longer instantiated from specific classes, but `torchvision.prototype.datasets` API now has five top-level functions:

1. `home()`: In contrast to the current behavior, the new style datasets will handle storing the data automatically. `home()` returns the root of the tree, which defaults to `~/.cache/torch/datasets/vision`. By passing a path to `home()` the root can be moved to any other location.
2. `list()`: Returns a list of all available datasets. 
3. `info(name)`: Returns static info about a dataset like the name, homepage, categories, or the available options to load the dataset.
4. `load(name, **kwargs)`: Return the given dataset as datapipe, in which each sample is a dictionary containing all the information.
5. `register()`: This is used to register datasets (more on what this actually is later) with our API. Since `datasets.load()` is a thin convenience wrapper, we might also not expose this and only use it for builtin datasets (which are *not* part pf this PR).

So the workflow that currently looks like this

```python
from torchvision.datasets import MNIST

dataset = MNIST(root="/path/to/root", train=False)
```

will now look like

```python

dataset = datasets.load("mnist", split="test")
```
(note that the example cannot be run yet, because there are no built-in datasets in this PR)

## How do we implement the new style datasets?

There are three closely connected objects that are all importable from `torchvision.prototype.datasets.utils`. The first two will only be used in most cases:

- `DatasetInfo`: Static information about the dataset. See `datasets.info()` above.
- `DatasetConfig`: Namespace containing the configuration of the dataset, for example which split to use.

The last object, `Dataset`, is actually the one that actually holds the information about how to load the dataset. It has three abstract methods that need to be overwritten:

1. `info` (property): Returns the `DatasetInfo` with the static information about the dataset. This is for example used to instantiate the `DatasetConfig` from the additional parameters in `datasets.load()`
2. `resources`: Returns a list of `OnlineResource`'s that need to be present to load the current `config`. They will be downloaded and checked automatically.
3. `_make_datapipe`: The heart of the implementation. The method receives a list of datapipes, which correspond to the output of `resources`, the current config, as well as optionally a shuffler and decoder. Implementation is highly dataset specific, so examples for this will only be showcased in follow-up PRs that add some fundamental datasets.

cc @pmeier @mthrok


torchvision/prototype/datasets/utils/_dataset.py
```diff
@@ -0,0 +1,218 @@
...
+    body_too_long = (len(prefix) + len(body) + len(postfix)) > line_length
+    multiline_body = len(str(body).splitlines()) > 1
+    if not (body_too_long or multiline_body):
+        return prefix + body + postfix
+
+    body = textwrap.indent(to_str(",\n"), " " * 2)
+    return f"{prefix}\n{body}\n{postfix}"
+
+
+class DatasetConfig(Mapping):
```
pmeier, line 48: This class looks a lot like overkill, but it is actually not:

1. Since `lambda`'s are "not welcome" in datapipes that take a callable (a warning will be thrown), we often need to use `functools.partial(..., config=config)`. To do that `config` has to be hashable. Hence, we implement the `__hash__` and `__eq__` methods and fail in case `__getitem__` or `__delitem__` are invoked.
2. For convenience it is much nicer to use a `DatasetConfig` as namespace as a dictionary, so we additionally implement `__getattr__`, `__setattr__`, and `__delattr__`.

torchvision/prototype/datasets/utils/_dataset.py
```diff
@@ -0,0 +1,218 @@
...
+            valid_args = self._valid_options[name]
+
+            if arg not in valid_args:
+                raise ValueError(
+                    add_suggestion(
+                        f"Invalid argument '{arg}' for option '{name}' of dataset {self.name}.",
+                        word=arg,
+                        possibilities=valid_args,
+                    )
+                )
```
pmeier, line 158: We can move this to a dedicated function, so we have the ability to override the default option checking.

torchvision/prototype/datasets/utils/_resource.py
```diff
@@ -0,0 +1,54 @@
...
+class OnlineResource:
+    def __init__(self, url: str, *, sha256: str, file_name: str) -> None:
+        self.url = url
+        self.sha256 = sha256
+        self.file_name = file_name
+
+    def to_datapipe(self, root: Union[str, pathlib.Path]) -> IterDataPipe:
+        path = (pathlib.Path(root) / self.file_name).expanduser().resolve()
+        # FIXME
+        return FileLoader(IterableWrapper((str(path),)))
```
pmeier, line 36: This is greatest weak point right now. Functionality to download stuff and verify the checksum is present in `torchdata`, but is not public yet.

torchvision/prototype/datasets/_home.py
```diff
@@ -0,0 +1,20 @@
+import os
+import pathlib
+from typing import Optional, Union
+
+from torch.hub import _get_torch_home
+
+HOME = pathlib.Path(_get_torch_home()) / "datasets" / "vision"
+
+
+def home(home: Optional[Union[str, pathlib.Path]] = None) -> pathlib.Path:
```
fmassa, line 10: FYI the function name being equal to the argument could be misleading (recursive function?)

torchvision/prototype/datasets/utils/_dataset.py
```diff
@@ -0,0 +1,220 @@
...
+    def __repr__(self) -> str:
+        return make_repr(type(self).__name__, self.items())
+
+
+class DatasetInfo:
+    def __init__(
+        self,
+        name: str,
+        *,
+        categories: Union[int, Sequence[str], str, pathlib.Path],
```
fmassa, line 101: For the future: this assumes that all we care about for datasets are the categories. But sometimes datasets don't have categories (i.e., for depth estimation), or it might have more info that we might want to expose (meta-categories).

If we assume that all datasets will return dicts, we could instead let this return the types for each one of the outputs of the dataset. But we can postpone the discussion to the future

fmassa (approved):
I made a few comments but they can be addressed in follow-up PRs.

Thanks!

torchvision/prototype/datasets/_api.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+DATASETS: Dict[str, Dataset] = {}
+
+
+def register(dataset: Dataset) -> None:
+    DATASETS[dataset.name] = dataset
+
+
+def list() -> List[str]:
```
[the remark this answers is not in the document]
pmeier: Ouch, you are right. I'll fix that. Is it ok to have it in the `datasets` namespace though?

torchvision/prototype/datasets/_home.py
pmeier: Yeah, let's just rename this to `root` or something like that.

torchvision/prototype/datasets/_api.py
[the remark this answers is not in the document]
fmassa: yes, I think it's ok

PR #4479, opened 2021-09-24
Removed tests executing deprecated F_t.center/five/ten_crop methods

vfdev-5:
Description:
- Removed tests executing deprecated F_t.center/five/ten_crop methods
- Suppressed deprecation warnings while running test/test_functional_tensor.py

For example, on `main`: https://app.circleci.com/pipelines/github/pytorch/vision/10708/workflows/d8abe45c-a6c3-4bcb-8679-1d1f30858d2f/jobs/816425?invite=true#step-108-3325
```
test/test_functional_tensor.py::test_assert_image_tensor[center_crop-args8-cpu]
  /root/project/torchvision/transforms/functional_tensor.py:244: UserWarning: This method is deprecated and will be removed in future releases. Please, use ``F.center_crop`` instead.
    "This method is deprecated and will be removed in future releases. "

test/test_functional_tensor.py::test_assert_image_tensor[five_crop-args9-cpu]
  /root/project/torchvision/transforms/functional_tensor.py:268: UserWarning: This method is deprecated and will be removed in future releases. Please, use ``F.five_crop`` instead.
    "This method is deprecated and will be removed in future releases. "

test/test_functional_tensor.py::test_assert_image_tensor[ten_crop-args10-cpu]
  /root/project/torchvision/transforms/functional_tensor.py:295: UserWarning: This method is deprecated and will be removed in future releases. Please, use ``F.ten_crop`` instead.
    "This method is deprecated and will be removed in future releases. "
```


fmassa (approved):

NicolasHug (commented):
If the goal is to remove the warnings, maybe we could just specifically suppress them with a pytest decorator? https://docs.pytest.org/en/6.2.x/warnings.html#pytest-mark-filterwarnings

NicolasHug:
Thanks for the details @vfdev-5 
As a general principle I think we should not remove tests until we properly remove what's being tested, whether it's deprecated or not (if it's deprecated and not tested, we might as well just remove it...)

But I agree that if `_assert_image_tensor` was the only test left, it's not even worth keeping anyway.

fmassa:
@NicolasHug those functions are there only for BC reasons. They can be entirely implemented in a generic way in the `functional` namespace, as they just dispatch to `crop`.

That wasn't the case in their first implementation, so we kept those functions in the codebase for BC, but we can safely remove the tests

PR #4484, opened 2021-09-25
Rewrite gallery example for masks to boxes.

oke-aditya:
Not done yet, I will ping once it's ready for review.


oke-aditya:
I think I have completed it. Now the tutorial includes converting masks to boxes and
1. torchvision.io to read images and masks.
2. Plotting masks and boxes using torchvision.utils
3. Using FasterRCNN over converted boxes.
4. A simple SegmentationToDetection dataset

Feel free to review @NicolasHug @0x00b1

NicolasHug:
thanks @oke-aditya , I'll take a look tomorrow

oke-aditya:
Here is the link to rendered example :smile: 

https://822905-73328905-gh.circle-artifacts.com/0/docs/auto_examples/plot_repurposing_annotations.html#

gallery/plot_repurposing_annotations.py
```diff
@@ -7,69 +7,190 @@
...
+
+
+from torchvision.io import read_image
+from torchvision.ops import masks_to_boxes
+
+img_path = os.path.join(ASSETS_DIRECTORY, "FudanPed00054.png")
+mask_path = os.path.join(ASSETS_DIRECTORY, "FudanPed00054_mask.png")
+img = read_image(img_path)
+mask = read_image(mask_path)
+mask = F.convert_image_dtype(mask, dtype=torch.float)
```
NicolasHug, line 70: Do we really need to convert the mask into a float value here? It looks like we could keep it as integer values which seems more natural?

Eventually the masks will need to be converted into boolean masks so there's no reason to keep track of 3 different dtypes, we could just have the original dtype (int) and the end dtype (bool)

gallery/plot_repurposing_annotations.py
```diff
@@ -7,69 +7,190 @@
...
+
+#########################
+# Here the masks are represented as a PNG Image, with floating point values.
+# Each pixel is encoded as different colors, with 0 being background
```
NicolasHug, line 74:
```suggestion
# Each pixel is encoded as different colors, with 0 being background.
```

gallery/plot_repurposing_annotations.py
```diff
@@ -7,69 +7,190 @@
...
+# Notice that that dimensions of image and mask match.
```
NicolasHug, line 75:
```suggestion
# Notice that the dimensions of image and mask match.
```

gallery/plot_repurposing_annotations.py
```diff
@@ -7,69 +7,190 @@
...
+# Now the masks are boolean tensor.
+# The first dimension in this case 3. Signifies number of instance, that is 3 people in the image.
+# The other two dimensions are height and width, which match the dimensions of the image.
+# The boolean tensors represent if the pixel belong to the segmentation mask of the image.
+
+print(masks.size())
+print(masks)
+
+####################################
+# Let's visualize an image and it's corresponding segmentation masks.
```
NicolasHug, line 102:
```suggestion
# Let's visualize an image and its corresponding segmentation masks.
```

gallery/plot_repurposing_annotations.py
```diff
@@ -7,69 +7,190 @@
...
+
+
+####################################
+# Converting Segmentation Dataset to Detection Dataset
+# ----------------------------------------------------
+#
+# With this utility it becomes very simple to convert a segmentation dataset to a detection dataset.
+# With this we can now use a segmentation dataset to train a detection model.
+# One can similarly convert panoptic dataset to detection dataset.
+# You can use the below code for Dataset creation in PenFudan Detection Tutorial
```
NicolasHug, line 153: Is this the tutorial you're referring to? https://pytorch.org/tutorials/intermediate/torchvision_tutorial.html

Maybe we could link to it?

NicolasHug (commented):
Thanks @oke-aditya , I made a few minor comments but this looks great.


Since the example currently only shows how to repurpose masks into bounding boxes, should we change the title of the example to 

"Repurposing masks into bounding boxes"

?

I feel like this would be more specific and perhaps more enticing for the reader? We could always change the title later when we add more stuff (we can keep the file name as is)

gallery/plot_repurposing_annotations.py
oke-aditya: Yes, I just added a comment to say that it would work with floating dtype too. Maybe users wouldn't know that.

gallery/plot_repurposing_annotations.py
oke-aditya: Linked :smiley:

oke-aditya:
Hi @NicolasHug feel free to have a review.

gallery/plot_repurposing_annotations.py
```diff
@@ -1,75 +1,204 @@
...
-=======================
-
-The following example illustrates the operations available in the torchvision.ops module for repurposing object
-localization annotations for different tasks (e.g. transforming masks used by instance and panoptic segmentation
+=====================================
+Repurposing masks into bounding boxes
+=====================================
+
+The following example illustrates the operations available in the
+`torchvision.ops <https://pytorch.org/vision/stable/ops.html>`_ module for repurposing
```
NicolasHug, line 7: instead of using hard links we should instead use the sphinx directives which will preserve the current doc version that's being browsed. In this case this should be

```suggestion
:mod:`torchvision.ops` module for repurposing
```

gallery/plot_repurposing_annotations.py
```diff
@@ -1,75 +1,204 @@
...
+
+from torchvision.io import read_image
+
+img_path = os.path.join(ASSETS_DIRECTORY, "FudanPed00054.png")
+mask_path = os.path.join(ASSETS_DIRECTORY, "FudanPed00054_mask.png")
+img = read_image(img_path)
+mask = read_image(mask_path)
+
+# You can use floating point masks as well
+# mask = F.convert_image_dtype(mask, dtype=torch.float)
```
NicolasHug, line 72: I'm not sure this comment is needed to be honest. Is there a case where users would want to convert the masks to float values?

I'm suggesting an alternative comment below

gallery/plot_repurposing_annotations.py
```diff
@@ -1,75 +1,204 @@
...
+
+############################
+
+# We get the unique colors, as these would be the object ids.
+obj_ids = torch.unique(mask)
+
+# first id is the background, so remove it.
+obj_ids = obj_ids[1:]
+
+# split the color-encoded mask into a set of boolean masks.
```
NicolasHug, line 91:
```suggestion
# split the color-encoded mask into a set of boolean masks.
# Note that this snippet would work as well if the masks were float values instead of ints.
```

gallery/plot_repurposing_annotations.py
```diff
@@ -1,75 +1,204 @@
...
+masks = mask == obj_ids[:, None, None]
+
+########################
+# Now the masks are a boolean tensor.
+# The first dimension in this case 3. Which signifies the number of instances, that there are 3 people in the image.
```
NicolasHug, line 96: I think this might be a bit simpler:

```suggestion
# The first dimension in this case 3 and denotes the number of instances: there are 3 people in the image.
```

gallery/plot_repurposing_annotations.py
```diff
@@ -1,75 +1,204 @@
...
+# The other two dimensions are height and width, which are equal to the dimensions of the image.
+# For each instance, the boolean tensors represent if the particular pixel belong to the segmentation mask of the image.
```
NicolasHug, line 98:
```suggestion
# For each instance, the boolean tensors represent if the particular pixel belongs to the segmentation mask of the image.
```

gallery/plot_repurposing_annotations.py
```diff
@@ -1,75 +1,204 @@
...
+from torchvision.ops import masks_to_boxes
+
+boxes = masks_to_boxes(masks)
+print(boxes.size())
+print(boxes)
+
+####################################
+# As the shape denotes, there are 3 boxes and in ``(xmin, ymin, xmax, ymax)`` format.
+# These can be visualized very easily with :func:`~torchvision.utils.draw_bounding_boxes` utility
+# provided in `torchvision.utils <https://pytorch.org/vision/stable/utils.html>`_.
```
NicolasHug, line 129:
```suggestion
# provided in :mod:`torchvision.utils`.
```

gallery/plot_repurposing_annotations.py
```diff
@@ -1,75 +1,204 @@
...
+
+####################################
+# Converting Segmentation Dataset to Detection Dataset
+# ----------------------------------------------------
+#
+# With this utility it becomes very simple to convert a segmentation dataset to a detection dataset.
+# With this we can now use a segmentation dataset to train a detection model.
+# One can similarly convert panoptic dataset to detection dataset.
+# You can use the below code for Dataset creation in
+# `PenFudan Detection Tutorial <https://pytorch.org/tutorials/intermediate/torchvision_tutorial.html>`_.
```
NicolasHug, line 161: minor suggestion:

```suggestion
# Here is an example where we re-purpose the dataset from the
# `PenFudan Detection Tutorial <https://pytorch.org/tutorials/intermediate/torchvision_tutorial.html>`_.
```

NicolasHug (commented):
Thanks @oke-aditya , I made another pass with minor suggestions, LGTM when addressed :)

gallery/plot_repurposing_annotations.py
oke-aditya: :mod:`torchvision.ops` seems to not render a HTML link.

image

while the previous one rendered a link, which can be redirected.

image

I would prefer a link, since that can be clicked.

gallery/plot_repurposing_annotations.py
oke-aditya: See my comment above, it will not render a link similarly.

gallery/plot_repurposing_annotations.py
oke-aditya: Users won't need to convert the masks, I just wanted to demonstrate that they can use float masks if it is needed.
I'm fine with either, let me know what you think,

oke-aditya:
Hey @NicolasHug two small things to clear, then we would be good to merge.

gallery/plot_repurposing_annotations.py
NicolasHug: Isn't it actually a link? It should be. If it's not, maybe try ``` :mod:`ops` ``` then. 
Otherwise let's do what we did here: https://github.com/pytorch/vision/blob/main/gallery/plot_transforms.py#L6-L7
 and https://raw.githubusercontent.com/pytorch/vision/main/docs/source/transforms.rst (see the `.. _transforms:` at the top )

gallery/plot_repurposing_annotations.py
NicolasHug: If they don't need to convert the masks then I think it's best not to suggest that they could do it :)

oke-aditya: Cool!

gallery/plot_repurposing_annotations.py
oke-aditya: Done, finally the transforms trick, worked. :ref: has worked and we now get redirect links.

oke-aditya:
We should be good to merge!! @NicolasHug

NicolasHug (approved):
Thanks a lot @oke-aditya !!
LGTM, I'll merge tomorrow after checking the rendered docs :)

oke-aditya:
Here is the link for final docs. :smiley: 

https://826691-73328905-gh.circle-artifacts.com/0/docs/auto_examples/plot_repurposing_annotations.html

NicolasHug:
Thanks again @oke-aditya !!

oke-aditya:
I learn a lot by contributing. Thanks to you guys. 😀

PR #4487, opened 2021-09-27
Replace MobileNetV3's SqueezeExcitation with EfficientNet's one

### LINKED ISSUE #4455: Replace SqueezeExcitation of MobileNetV3 with the one of EfficientNet
### 🚀 The feature

Investigate if it's possible to replace the `SqueezeExcitation` of MobileNetV3 in a BC manner and use the one introduced on `EfficientNet` at #pull.

### Motivation, pitch

This will reduce the amount of different SE implementations that we have and open the possibility of moving the layer on a separate area where it can be reused.

### Alternatives

_No response_

### Additional context

_No response_

cc @datumbox

datumbox:
Fixes #4455

All validation stats of models remain the same:

<details><summary>mobilenet_v3_large</summary>

```
torchrun --nproc_per_node=2 train.py --model mobilenet_v3_large --test-only --pretrained

Main Branch:
Test:  Acc@1 74.042 Acc@5 91.340

PR:
Test:  Acc@1 74.042 Acc@5 91.340

```
</details>

<details><summary>mobilenet_v3_small</summary>

```
torchrun --nproc_per_node=2 train.py --model mobilenet_v3_small --test-only --pretrained

Main Branch:
Test:  Acc@1 67.668 Acc@5 87.402

PR:
Test:  Acc@1 67.668 Acc@5 87.402

```
</details>

<details><summary>quantized mobilenet_v3_large</summary>

```
python -u train_quantization.py --device cpu --model mobilenet_v3_large --test-only

Main Branch:
Test:  Acc@1 73.004 Acc@5 90.858

PR:
Test:  Acc@1 73.004 Acc@5 90.858

```
</details>

<details><summary>ssd300_vgg16</summary>

```
torchrun --nproc_per_node=2 train.py --dataset coco --model ssd300_vgg16 --pretrained --test-only

Main Branch:
IoU metric: bbox
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.251
 Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=100 ] = 0.415
 Average Precision  (AP) @[ IoU=0.75      | area=   all | maxDets=100 ] = 0.262
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.055
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.268
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.435
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=  1 ] = 0.239
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets= 10 ] = 0.344
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.365
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.088
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.406
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.602

PR:
IoU metric: bbox
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.251
 Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=100 ] = 0.415
 Average Precision  (AP) @[ IoU=0.75      | area=   all | maxDets=100 ] = 0.262
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.055
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.268
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.435
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=  1 ] = 0.239
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets= 10 ] = 0.344
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.365
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.088
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.406
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.602

```
</details>

<details><summary>lraspp_mobilenet_v3_large</summary>

```
torchrun --nproc_per_node=2 train.py --dataset coco --model lraspp_mobilenet_v3_large --pretrained --test-only

Main Branch:
global correct: 91.2
average row correct: ['94.5', '84.3', '69.5', '72.8', '57.7', '42.0', '77.0', '57.0', '90.4', '36.1', '76.0', '60.8', '81.4', '78.9', '81.0', '87.6', '51.3', '83.9', '62.2', '84.2', '56.1']
IoU: ['90.2', '69.2', '57.7', '58.5', '47.8', '35.7', '69.5', '47.1', '79.1', '29.6', '62.6', '34.2', '65.5', '63.4', '70.0', '76.8', '30.1', '61.9', '46.8', '70.6', '49.1']
mean IoU: 57.9

PR:
global correct: 91.2
average row correct: ['94.5', '84.3', '69.5', '72.8', '57.7', '42.0', '77.0', '57.0', '90.4', '36.1', '76.0', '60.8', '81.4', '78.9', '81.0', '87.6', '51.3', '83.9', '62.2', '84.2', '56.1']
IoU: ['90.2', '69.2', '57.7', '58.5', '47.8', '35.7', '69.5', '47.1', '79.1', '29.6', '62.6', '34.2', '65.5', '63.4', '70.0', '76.8', '30.1', '61.9', '46.8', '70.6', '49.1']
mean IoU: 57.9

```
</details>


torchvision/models/quantization/mobilenetv3.py
```diff
@@ -107,13 +110,13 @@ def _mobilenet_v3_model(
         torch.quantization.prepare_qat(model, inplace=True)
 
         if pretrained:
-            _load_weights(arch, model, quant_model_urls.get(arch + '_' + backend, None), progress)
+            _load_weights(arch, model, quant_model_urls.get(arch + '_' + backend, None), progress, False)
```
datumbox, line 113: Earlier versions of the SqueezeExcite class used `F.adaptive_avg_pool2d()` and `F.hardsigmoid()` instead of their `nn.Module` equivalents. Using the latter are advised as because QAT can further optimize them.

Loading the old weights, is still possible but the QAT bits of the above two layers will be missing. Passing `strict=false` allows us to use the previous weights and achieve the same accuracy.

torchvision/models/mobilenetv3.py
```diff
@@ -18,25 +19,16 @@
 }
 
 
-class SqueezeExcitation(nn.Module):
-    # Implemented as described at Figure 4 of the MobileNetV3 paper
+class SqueezeExcitation(SElayer):
```
datumbox, line 22: No longer used, marked for deprecation but inherits from the efficientnet one.

torchvision/models/mobilenetv3.py
```diff
@@ -1,12 +1,13 @@
...
 
 from functools import partial
 from torch import nn, Tensor
-from torch.nn import functional as F
-from typing import Any, Callable, Dict, List, Optional, Sequence
+from typing import Any, Callable, List, Optional, Sequence
 
 from .._internally_replaced_utils import load_state_dict_from_url
-from torchvision.models.mobilenetv2 import _make_divisible, ConvBNActivation
+from .efficientnet import SqueezeExcitation as SElayer
```
datumbox, line 9: We can try moving it to a separate place in a separate PR.

datumbox (commented):
Highlighting some interesting bits of the implementation.

torchvision/models/quantization/mobilenetv3.py
```diff
@@ -79,11 +81,12 @@ def _load_weights(
     model: QuantizableMobileNetV3,
     model_url: Optional[str],
     progress: bool,
+    strict: bool
```
kazhang, line 84: nit: default to True, which is desired in most of cases.

torchvision/models/quantization/mobilenetv3.py
```diff
@@ -16,16 +17,17 @@
 }
 
 
-class QuantizableSqueezeExcitation(SqueezeExcitation):
+class QuantizableSqueezeExcitation(SElayer):
     def __init__(self, *args: Any, **kwargs: Any) -> None:
+        kwargs["scale_activation"] = nn.Hardswish
         super().__init__(*args, **kwargs)
         self.skip_mul = nn.quantized.FloatFunctional()
```
kazhang, line 24: Should we be concerned about BC here? what if some downstream callsites construct the module with `squeeze_factor`?

kazhang (commented):
LGTM overall. Thanks for working on this!
I only have a question on quantizable module BC.

torchvision/models/quantization/mobilenetv3.py
datumbox: Indeed we always pass strict=True. This here is out of necessity. See my earlier comment at https://github.com/pytorch/vision/pull/4487#discussion_r717030133

torchvision/models/quantization/mobilenetv3.py
datumbox: Good question. An option would be to keep the old class, remove its usages and deprecate it and introduced a new one similar to what I did here:
https://github.com/pytorch/vision/blob/496940877e3cb5164408a6a4b06a65364734d43b/torchvision/models/mobilenetv3.py#L22-L26

I had a quick look both on [Github](https://github.com/search?p=1&q=QuantizableSqueezeExcitation&type=Code) and on FBcode and I can't find any direct usages of this class outside the intended in this module. In addition the module is undocumented and not publicly exposed to `__all__`, hence I would personally be OK offering this level of BC. Thoughts?

torchvision/models/quantization/mobilenetv3.py
kazhang: I was referring to setting a default value here like `strict: bool=True` so that you don't need to explicitly set strict=True in line 119.
BTW, for readability, I always use name argument if the value is primitive.

kazhang (approved):
Thanks for consolidating the SE layers!

torchvision/models/quantization/mobilenetv3.py
datumbox: Oh I gotcha, sounds good. I'll make the change.

torchvision/models/quantization/mobilenetv3.py
```diff
@@ -79,11 +81,12 @@ def _load_weights(
     model: QuantizableMobileNetV3,
     model_url: Optional[str],
     progress: bool,
+    strict: bool = True
 ) -> None:
     if model_url is None:
         raise ValueError("No checkpoint is available for {}".format(arch))
     state_dict = load_state_dict_from_url(model_url, progress=progress)
-    model.load_state_dict(state_dict)
+    model.load_state_dict(state_dict, strict=strict)
```
fmassa, line 89: Instead of passing `strict=False` here to be able to get the pre-trained weights loaded, another option would be to implement a `-load_from_state_dict` method in `QuantizableSqueezeExcitation`, adding a `v2` to it and handling the missing fields.

An example on how we did it in the past can be seen in https://github.com/pytorch/vision/blob/a9d710a8ae9df7922981ffc79ec5102a8a927352/torchvision/ops/misc.py#L69-L85

This could be more robust, specially if an user has its own checkpoint based on their finetuned quantized `mobilenetv3`.

fmassa (approved):
Thanks for the PR!

I've left one comment which I think would be a better way of handling the BC in the quantized model.

I'm approving the PR now, as I would be ok merging the PR as it currently stands.

torchvision/models/quantization/mobilenetv3.py
datumbox: Hmm, interesting. Thanks for the suggestion. 

Let me provide a bit of context around this. Here I use a cheap workaround of passing `strict=False`. I don't really like this "solution" but I did this for 3 reasons:
1. It helps me avoid introducing a Versioning mechanism before the Model Versioning RFC is properly discussed. 
2. It's unclear to me how the new FX quantization API will affect the `Quantizable*` classes, so I didn't want to change anything but the bare minimum.
3. We are very close to the 0.11 release and I want to ensure no multiple SE layers will be released. I also wanted to avoid introducing complex versioning mechanisms that I won't be able to take back.

But as I really don't like workarounds. Let's chat about your proposal offline and see whether we will implement it now or after the release.

datumbox: I synced with Francisco offline and did a quick investigation to implement the proposal (see https://github.com/pytorch/vision/pull/4487/commits/b491fa26fd7638fed05665fcf40a51f42734122a). Unfortunately it seems that once the patch is applied, the accuracy of the model deteriorates significantly:
`Test:  Acc@1 0.120 Acc@5 0.534`

I'm pretty sure this can be done, but it requires more investigation. I propose to merge the workaround and defer the investigation for the future as this is an implementation detail that can be changed in a BC-manner.

PR #4490, opened 2021-09-27
Bump main trunk version to 0.12

malfet:


prabhat00155 (commented):
I think you missed https://github.com/pytorch/vision/blob/f483e71b512e0f9dd5d6eb531fb602581ffb39a6/packaging/build_cmake.sh#L18.

Issue #4488, opened 2021-09-27
Use stable sort in nms


shuokay:
### 🚀 The feature

Please use stable sort in nms since torch1.9 has supported stable sort.

### Motivation, pitch

Why we need stable sort:
In quantization model, many score are equal due to quantization, so when using unstable sort, the result of nms is non-reproducible.

datumbox:
It's true that the sort component of NMS has been the source of many headaches. The fact that the algorithm is unstable means that we get different order depending on platform/python/system config. This leads to us being unable to properly test those algorithms and we often resort to ugly hacks/workarounds:
https://github.com/pytorch/vision/blob/2e0949e22be6dceef5d91101e49ade1fe8a0689b/test/test_models.py#L579-L592

PR #4491, opened 2021-09-28
[nms]: replace unstable sort with stable sort

shuokay:
Use stable sort in nms since pytorch1.9 has supported stable sort. The discussion is in https://github.com/pytorch/vision/issues/4488

### DIFF
--- a/torchvision/csrc/ops/cpu/nms_kernel.cpp
+++ b/torchvision/csrc/ops/cpu/nms_kernel.cpp
@@ -27,7 +27,8 @@ at::Tensor nms_kernel_impl(
 
   at::Tensor areas_t = (x2_t - x1_t) * (y2_t - y1_t);
 
-  auto order_t = std::get<1>(scores.sort(0, /* descending=*/true));
+  auto order_t =
+      std::get<1>(scores.sort(0, /* descending=*/true, /*stable*/ true));
 
   auto ndets = dets.size(0);
   at::Tensor suppressed_t = at::zeros({ndets}, dets.options().dtype(at::kByte));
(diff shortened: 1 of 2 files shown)


NicolasHug:
Thanks for the PR @shuokay 

It looks like the changes are causing a few test failures: https://app.circleci.com/pipelines/github/pytorch/vision/10788/workflows/50fb9b2b-7f7b-47e3-8d9c-81357e4ad183/jobs/824433


Also for completeness I think we should update the quantized nms as well https://github.com/pytorch/vision/blob/main/torchvision/csrc/ops/quantized/cpu/qnms_kernel.cpp


That being said, it's important to note that this would be a BC breaking change, and I'm personally not sure that changing downstream users' results is warranted in this case. I think it's worth considering adding a new `stable_sort=False` parameter though.

datumbox:
> It looks like the changes are causing a few test failures:
> this would be a BC breaking change

I agree we should investigate the test failures and understand why it's failing.

If the change only affects the order (as it should) then I wouldn't consider this a breaking change. This is because the order of items produced by the current sorting algorithm depends on platform/OS specific characteristics, so the outcome is pretty random and can't be relied upon. It has also been a source of major pain for us in the testing of detection models.

One thing we should check though before merging if there is any speed degradation on NMS. We should isolate NMS as much as possible and measure before/after to understand the tradeoff.

fmassa:
I'm fine getting this PR merged (once the quantized kernels are also updated). If anything, using stable sort would be more natural for users and make it consistent across OSes.

As @datumbox pointed out though, it would be good to understand what are the speed implications by making this change. Can you perform some benchmarks on tensors of varying sizes (from 100 to up to 6k boxes) to give a better picture of the trade-offs here?

PR #4496, opened 2021-09-28
Switch testing to torch_test rather than torch_nightly

malfet:


malfet:
`pytorch_test` channel /`whl/test/` folder contains binaries from `release/X.YY` branch, whereas `nightly` contains binaries from from default branch. Goal is to test `torchvision` release branch against respective release candidate rather than against nightly builds

NicolasHug:
Thanks for clarifying @malfet . But does that mean that we will only be able to test torchvision against the latest stable (or RC) release?
 
Considering the pace at which these stable/RC releases are issued by pytorch core, it looks like we would be lagging quite a bit, which in some cases might hinder our development speed I believe. We would definitely not complain about the improved stability though :p 

Perhaps this is something that we should discuss on Thursday in the torchvision meeting @fmassa @datumbox @prabhat00155 @vmoens 

As a minor note, we currently tell users users to use pytorch-nightly to buildl from source, so I assume we would need to update these instructions as well if we ever switch.

datumbox:
Thanks for the ping.

My understanding is that this PR will change torch_test on the 0.11 release branch, not main. Main will continue pointing to the nightly. Could you please that this is the intention?

malfet:
@NicolasHug this PR targets `release/0.11` branch rather than `main`, i.e. the restriction would only apply to release branches

NicolasHug:
Ooh, thanks, I missed that. That sounds good then!

PR #3032, opened 2020-11-19
Support for image with no annotations in RetinaNet

datumbox:
Fixes #2800

Merge after #3031, coauthored with @fmassa.

To ensure that the patch in this branch does not affect the performance of the model I retrained it and compared it with master.

### Branch

<details><summary>retinanet_resnet50_fpn + no annotations support: boxAP=36.4</summary>

```
0: IoU metric: bbox
0:  Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.364
0:  Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=100 ] = 0.558
0:  Average Precision  (AP) @[ IoU=0.75      | area=   all | maxDets=100 ] = 0.384
0:  Average Precision  (AP) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.190
0:  Average Precision  (AP) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.401
0:  Average Precision  (AP) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.492
0:  Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=  1 ] = 0.314
0:  Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets= 10 ] = 0.503
0:  Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.542
0:  Average Recall     (AR) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.342
0:  Average Recall     (AR) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.584
0:  Average Recall     (AR) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.690
```
</details>

### Master

<details><summary>retinanet_resnet50_fpn: boxAP=36.4</summary>

```
0: IoU metric: bbox
0:  Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.364
0:  Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=100 ] = 0.558
0:  Average Precision  (AP) @[ IoU=0.75      | area=   all | maxDets=100 ] = 0.381
0:  Average Precision  (AP) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.193
[...]

PR #4497, opened 2021-09-28
Prevent tests from leaking their respective RNG

NicolasHug:
This PR adds a new autouse pytest fixture to prevent each test from leaking its RNG consumption to other tests.

We've had strange errors and major headaches because of this in the past https://github.com/pytorch/vision/pull/3032#issuecomment-734829336

After a bit of thoughts and discussions with @pmeier and @vfdev-5, I don't think there's a better alternative than relying on an autouse fixture. It's a bit implicit, but that's the best we can do considering that pytorch only offers `manual_seed()` to contrrol the RNG, which leaks to the entire program execution.

cc @pmeier


test/test_transforms_tensor.py
```diff
@@ -714,6 +714,7 @@ def test_random_apply(device):
 @pytest.mark.parametrize('channels', [1, 3])
 def test_gaussian_blur(device, channels, meth_kwargs):
     tol = 1.0 + 1e-10
+    torch.manual_seed(12)
```
NicolasHug, line 717: I had to add this to prevent the test from failing (see https://app.circleci.com/pipelines/github/pytorch/vision/10830/workflows/946973f0-bad1-48ed-aaf3-f3720ab5f56a/jobs/828462).

If anything, this shows that the PR is working as expected and that this test is a bit flaky, and sensible to the `_create_data()` call.

To confirm I parametrized it over 100 random seeds, and I got 8 failures over 1000+ test instances. Each time, just 1 pixel was off. Considering the low failure rate, I think it's fine to keep the manual seeding here.

test/conftest.py
```diff
@@ -80,3 +82,26 @@ def pytest_sessionfinish(session, exitstatus):
     # To avoid this, we transform this 5 into a 0 to make testpilot happy.
     if exitstatus == 5:
         session.exitstatus = 0
+
+
+@pytest.fixture(autouse=True)
+def prevent_leaking_rng():
```
NicolasHug, line 88: This is pretty much the same as `freeze_rng_state()` https://github.com/pytorch/vision/blob/main/test/common_utils.py#L86

`autouse=True` means that this fixture will be used by every single test automatically.

vfdev-5 (approved):
Cool ! LGTM

pmeier:
Nicolas and I had an offline discussion about this a while back and I wanted to also give my opinion here: 

IMO, this is not needed. There should be only two types of tests for this purpose
1. The test uses random data.
2. The test uses fixed data, that we generate by setting the random seed and afterwards generate "random" data.

In case 1. we don't need a fixed RNG state, because it shouldn't matter. If it does, the test actually belongs to category 2. and should be fixed.

In case 2. the RNG state doesn't matter since we overwrite the seed anyway. 

@NicolasHug Is there an actual issue that this will fix?

fmassa (approved):
LGTM, thanks!

PyTorch does provide a `generator` that we can pass to (some) functions, but last time I checked it didn't work for CUDA (but that was a long time ago).

Your solution seems ok with me, thanks!

NicolasHug:
@pmeier I think that https://github.com/pytorch/vision/pull/3032#issuecomment-734829336 and the very fact that `test_gaussian_blur()` broke clearly show that this is in fact needed.

> IMO, this is not needed. There should be only two types of tests for this purpose

While you're right in principle, the reality much less simple unfortunately. We have tests for which the RNG is not entirely controlled and in fact depends on other tests. `test_gaussian_blur()` was one of them. It's likely that there are others (and if they're flaky they'll pop up soon, which is a good thing).

NicolasHug:
Also, and possibly most importantly:

> 1. The test uses random data.


Before this PR, we had no such tests: all tests were using a pre-defined RNG state that was leaked by the previous tests that were run before. There was almost zero test that was relying on the global pytorch RNG that would change across executions. The only tests that did were the tests that were run before the first test that called `manual_seed()`.

pmeier:
Maybe I misunderstood the purpose. From your comments I get that you want this to detect flaky tests that we don't realize are there, because they pass until the the order of execution changes and the get different "random" data. Is that correct?

NicolasHug:
> detect flaky tests that we don't realize are there, because they pass until the the order of execution changes and the get different "random" data. Is that correct?

This is indeed a positive consequence of the changes, but that's not the main intent.

The main intent is that any test `x` should never be dependent on any other test `y`. I agree with you that in an ideal world we would never write such tests, but the reality is that we often do, whether we know it or not. The main reason being that it's not always obvious to figure out where randomness is involved.

PR #4499, opened 2021-09-29
add torchdata as prototype dependency

pmeier:


pmeier:
Blocked by pytorch/data#20.

torchvision/prototype/datasets/__init__.py
```diff
@@ -1,3 +1,13 @@
+try:
+    import torchdata
+except (ModuleNotFoundError, TypeError) as error:
```
pmeier, line 3: The `torchdata` library hosted on PyPI fails for me at import with an `TypeError`. This is why we can aggregate both failure cases here.

torchvision/prototype/datasets/__init__.py
```diff
@@ -1,3 +1,13 @@
...
+    raise ModuleNotFoundError(
+        "`torchvision.prototype.datasets` depends on PyTorch's `torchdata` (https://github.com/pytorch/data). "
+        "You can install it with `pip install git+https://github.com/pytorch/data.git`. "
```
fmassa, line 6: Do you need to add `@main` in the uri ? You've added it in the other commands

torchvision/prototype/datasets/__init__.py
fmassa: Chatted with @pmeier , this is a bug in the `torchdata` from PyPI (which is unrelated to the `torchdata` we are depending on.

torchvision/prototype/datasets/__init__.py
pmeier: Nope, if you don't specify a branch / tag / ..., `pip` uses the default branch. I'll remove it in the other places to be consistent.

fmassa (approved):
Thanks! Only have one quick comment

torchvision/prototype/datasets/__init__.py
pmeier: If you do `pip install torchdata`, `python -c "import torchdata"` fails with 

```
TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases
```

PR #4504, opened 2021-09-29
Moving common layers to ops

### LINKED ISSUE #4333: [RFC] API For Common Layers In Torchvision

datumbox:
Partially resolves #4333

In this PR:
- [x] Move the commonly used `_make_divisible()` from MobileNetV2 to `models._utils`.
- [x] Deprecate the old `ConvBNReLU` and `ConvBNActivation` classes in favour of a new `ConvNormActivation` added in `ops.misc`.
- [x] Moved `SqueezeExcitation` to `ops.misc` and slightly modified the class to use only nn.Modules instead of functional.
- [x] Update RegNet implementation.
- [x] Validate that the accuracy of all affected models remains the same: https://www.internalfb.com/phabricator/paste/view/P460968703


torchvision/ops/misc.py
```diff
@@ -97,3 +97,56 @@ def forward(self, x: Tensor) -> Tensor:
...
+class SqueezeExcitation(torch.nn.Module):
+    def __init__(
+        self,
+        input_channels: int,
+        squeeze_channels: int,
+        activation: Callable[..., torch.nn.Module] = torch.nn.ReLU,
+        scale_activation: Callable[..., torch.nn.Module] = torch.nn.Sigmoid,
+    ) -> None:
+        super().__init__()
+        self.avgpool = torch.nn.AdaptiveAvgPool2d(1)
```
datumbox, line 137: A tiny change from @kazhang's original implementation. It's best to always use nn modules as it works better with quantization (at least the old API).

oke-aditya:
Can we move stochastic depth from ops.boxes to `ops.misc.py` ? I think we should avoid making it public until we decide about #4333 ?

datumbox:
@oke-aditya thanks for the comment.

Note that the StochasticDepth layer lives in its own module at `ops.stochastic_depth`. Including stuff on `ops.misc` doesn't make them less public AFAIK.

The historical convention for where to include things seems to be that layers with functional and class components end up on their own modules while those that are simple and use standard building convolutional blocks end up in `ops.misc`.

fmassa (approved):
LGTM, thanks!

PR #4507, opened 2021-09-30
NOMRG [ao_migration] torchvision: torch.quantization -> torch.ao.quantization

NicolasHug:
Summary:
This changes the imports in the `caffe2/torch/ao/nn` to include the new import locations.

```
codemod -d pytorch/vision --extensions py 'torch.quantization' 'torch.ao.quantization'
```

Differential Revision: D31302368


NicolasHug:
This is just an export of D31302368 to see what breaks

PR #4510, opened 2021-09-30
add prototypes for `Caltech(101|256)` datasets

pmeier:
cc @pmeier @mthrok @bjuncek


setup.py
```diff
@@ -495,7 +495,7 @@ def run(self):
         # Package info
         packages=find_packages(exclude=('test',)),
         package_data={
-            package_name: ['*.dll', '*.dylib', '*.so']
+            package_name: ['*.dll', '*.dylib', '*.so', '*.categories']
```
pmeier, line 498: We need to some file type here, so that these file types are packaged with everything else. They are plain text files, but I've opted to give them a "custom" extension to not accidentally include anything else.

torchvision/prototype/datasets/_builtin/caltech.py
```diff
@@ -0,0 +1,209 @@
+import io
+import pathlib
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union
+import re
+
+import numpy as np
```
fmassa, line 6: Do we need numpy here? You only use it to cast the array to `int64`, which you can also do in numpy with a string (or directly add a `dtype` in `torch.as_tensor` / `torch.tensor`)

torchvision/prototype/datasets/_builtin/caltech.py
```diff
@@ -0,0 +1,209 @@
...
+        key, image_data, ann_data = data
+        category, _ = key
+        image_path, image_buffer = image_data
+        ann_path, ann_buffer = ann_data
+
+        label = self.info.categories.index(category)
+
+        image = decoder(image_buffer) if decoder else image_buffer
+
+        ann = read_mat(ann_buffer)
```
fmassa, line 98: Does `ann` contain anything else or just `box_coord` and `obj_contour`?

torchvision/prototype/datasets/_builtin/caltech.py
```diff
@@ -0,0 +1,209 @@
...
+
+        images_dp = TarArchiveReader(images_dp)
+        images_dp = Filter(images_dp, self._is_not_background_image)
+        # FIXME: add this after https://github.com/pytorch/pytorch/issues/65808 is resolved
+        # images_dp = Shuffler(images_dp, buffer_size=INFINITE_BUFFER_SIZE)
+
+        anns_dp = TarArchiveReader(anns_dp)
+        anns_dp = Filter(anns_dp, self._is_ann)
+
+        dp = KeyZipper(
```
fmassa, line 129: For my understanding: is this efficient compared to our current Caltech implementation based on the old-style datasets?

torchvision/prototype/datasets/_builtin/caltech.py
```diff
@@ -0,0 +1,209 @@
...
+
+    def generate_categories_file(self, root: Union[str, pathlib.Path]) -> None:
+        dp = self.resources(self.default_config)[0].to_datapipe(pathlib.Path(root) / self.name)
+        dp = TarArchiveReader(dp)
+        dir_names = {pathlib.Path(path).parent.name for path, _ in dp}
+        categories = [name.split(".")[1] for name in sorted(dir_names)]
+        create_categories_file(HERE, self.name, categories)
+
+
+if __name__ == "__main__":
```
fmassa, line 204: Question: do we want to keep this for all datasets?

fmassa (approved):
LGTM, thanks!

The only minor thing I would change is the `np` import, otherwise good to merge. Do you prefer to change that in a follow-up PR or directly here?

torchvision/prototype/datasets/_builtin/caltech.py
pmeier: Unfortunately, we do. The `box_coord` array has dtype uint16, which `torch.as_tensor` cannot handle. So either we use numpy to convert to a common dtype first, or we could also convert the values to a list and build a tensor from that. What would you prefer? In any case I'll add a comment explaining what is going on.

torchvision/prototype/datasets/_builtin/caltech.py
pmeier: Nope.

torchvision/prototype/datasets/_builtin/caltech.py
pmeier: I'm not sure how to answer if this is efficient or not. 

Without the `Shuffler` we could use `Zipper` here and this is basically what we are doing now. If both datapipes, i.e. `images_dp` and `anns_dp`, are already aligned, `KeyZipper` only adds the overhead of evaluating `key_fn` and `ref_key_fn` for every sample.

As an alternative we could put the `Shuffler` after we use a `Zipper` and remove the `KeyZipper` all together.

torchvision/prototype/datasets/_builtin/caltech.py
pmeier: The idea behind this is that we should keep track of how we generated the category files. With this it is easy to regenerate if we need to. Just call `python -m torchvision.prototype.datasets._builtin.caltech`. Maybe we could also do an aggregation to be able to generate "everything", but I would keep that for the future.

torchvision/prototype/datasets/_builtin/caltech.py
pmeier: Just `box_coord` and `obj_contour`?

PR #4512, opened 2021-09-30
add prototype datasets for MNIST and variants

pmeier:


torchvision/prototype/datasets/_builtin/mnist.py
```diff
@@ -0,0 +1,396 @@
...
+        start: Optional[int]
+        stop: Optional[int]
+        if config.split == "test10k":
+            start = 0
+            stop = 10000
+        else:  # config.split == "test50k"
+            start = 10000
+            stop = None
+
+        return Slicer(dp, start=start, stop=stop)
```
ejguan, line 396: Is there way we can untackle `test10k` and `test50k` at the resource stage?

ejguan (commented):
I only have one concern for `QMNIST` as shown below

torchvision/prototype/datasets/utils/_internal.py
```diff
@@ -102,3 +107,65 @@ def __init__(self, datapipe: IterDataPipe[D], start: int = 0) -> None:
...
     def __iter__(self) -> Iterator[Tuple[int, D]]:
         yield from enumerate(self.datapipe, self.start)
+
+
+class CompressionType(enum.Enum):
+    GZIP = "gzip"
+    LZMA = "lzma"
+
+
+class Decompressor(IterDataPipe[Tuple[str, io.IOBase]]):
```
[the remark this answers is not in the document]
pmeier: > We may take reference from it for TorchData in the future. 

Please do so for every `IterDataPipe` in this file. I've only put them there if I thought they are general enough to be used by multiple datasets / torchdata.

torchvision/prototype/datasets/_builtin/mnist.py
pmeier: Unfortunately, no. The earliest would be in the `MNISTFileReader` that could enumerate the chunks and discard the unused chunks directly. In fact, that would improve performance since otherwise we also decode the chunks that we throw away later. I'll send a patch.

pmeier: The latest commits include this change

ejguan (approved):
LGTM with one comment about FC for web stream

PR #4513, opened 2021-09-30
make mypy more strict for prototype datasets

pmeier:
Instead of retrofitting more strictness for `mypy` later, we can start off with strict settings.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -4,6 +4,22 @@ files = torchvision
 show_error_codes = True
 pretty = True
 
+[mypy-torchvision.prototype.*]
+
+; untyped definitions and calls
+disallow_untyped_defs = True
+
+; None and Optional handling
+no_implicit_optional = True
+
+; warnings
+warn_unused_ignores = True
+warn_return_any = True
+warn_unreachable = True
+
+; miscellaneous strictness flags
+allow_redefinition = True
+
 [mypy-torchvision.io._video_opt.*]
 
 ignore_errors = True
--- a/torchvision/prototype/datasets/_builtin/caltech.py
+++ b/torchvision/prototype/datasets/_builtin/caltech.py
@@ -84,7 +84,10 @@ class Caltech101(Dataset):
         return category, id
 
     def _collate_and_decode_sample(
-        self, data, *, decoder: Optional[Callable[[io.IOBase], torch.Tensor]]
+        self,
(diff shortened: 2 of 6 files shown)


mypy.ini
```diff
@@ -4,6 +4,22 @@ files = torchvision
...
+; None and Optional handling
+no_implicit_optional = True
+
+; warnings
+warn_unused_ignores = True
+warn_return_any = True
+warn_unreachable = True
+
+; miscellaneous strictness flags
+allow_redefinition = True
```
NicolasHug, line 21: Are these the default values for those options?
If they're not the default, do we have a strong reason to use them instead of the defaults? Is this going to be clearly beneficial to the code-base and to us as developers?

pmeier: > Are these the default values for those options?

Nope.

> If they're not the default, do we have a strong reason to use them instead of the defaults? Is this going to be clearly beneficial to the code-base and to us as developers?

Let's go through them one by one:

- `disallow_untyped_defs`: by default `mypy` simply accepts untyped functions and uses `Any` for the input and output annotations. If our ultimate goal is to declare `torchvision` typed, we should make sure that we don't miss some functions. This flag enforces that.

- `no_implicit_optional`: By default `mypy` allows this:

  ```python
  def foo(bar: int = None) -> int:
      pass
  ```
  With this option enabled, it has to be
  ```python
  def foo(bar: Optional[int] = None) -> int:
      pass
  ```
  Given that `None` is a valid input, we should also explicitly mention it in the annotation.

- `warn_unused_ignores`: Sometimes we use `# type: ignore` directives on stuff that is actually wrong in other libraries. For example pytorch/pytorch#65998 will make some ignore directives obsolete that are needed now. Without this flag, we would never know.

- `warn_return_any`: If a function does something with dynamic types, `mypy` usually falls back to treating the output as `Any`. This will warn us if something like this happened, but we specified a more concrete output type. 

- `warn_unreachable`: This is more a test functionality, as `mypy` will now warn us if some code is unreachable. For example, with this flag set, `mypy` will warn that the `if` branch is unreachable.

  ```python
  def foo(bar: str) -> str:
      if isinstance(bar, int):
          bar = str(bar)
      return bar
  ```

- `allow_redefinition`: See #4531. If we have this globally, we can of course remove it here.

Apart from `warn_return_any` and `warn_unreachable` I think these flags are clearly beneficial. For the other two, they were beneficial for me in the past, but I can others object to them.

PR #4514, opened 2021-09-30
add prototype dataset for CelebA

pmeier:
cc @pmeier @mthrok @bjuncek


pmeier:
`mypy` will be happy after either #4513 or #4531 is merged.

torchvision/prototype/datasets/_builtin/celeba.py
```diff
@@ -0,0 +1,204 @@
...
+
+
+class CelebACSVParser(IterDataPipe):
+    def __init__(
+        self,
+        datapipe,
+        *,
+        has_header,
+        delimiter=" ",
+        skipinitialspace=True,
```
NicolasHug, line 32: I understand this is originally from the `csv` module, but this is inconsistent with the rest of our parameter naming convention. So unless we have a strong reason to keep this, I would suggest to change to `skip_initial_space`

Same for `other_fmt_params` below

torchvision/prototype/datasets/_builtin/celeba.py
```diff
@@ -0,0 +1,200 @@
...
+            if self.has_header:
+                # The first row is skipped, because it only contains the number of samples
+                next(file)
+
+                # Empty field names are filtered out, because some files have an extr white space after the header
+                # line, which is recognized as extra column
+                fieldnames = [name for name in next(csv.reader([next(file)], **self._fmtparams)) if name]
+                # Some files do not include a label for the image ID column
+                if fieldnames[0] != "image_id":
+                    fieldnames.insert(0, "image_id")
```
ejguan, line 49: This is super annoying that we can't use `DictReader`.
Since we have three datapipes with header, could we hard-code the `fieldnames` for each one and use `CSVDictReader(dp, skip_lines=2, fieldnames=[...])`

torchvision/prototype/datasets/_builtin/celeba.py
```diff
@@ -0,0 +1,200 @@
...
+        }
+        anns_dp: IterDataPipe
+        partial_anns_dps: Sequence[IterDataPipe]
+        anns_dp, *partial_anns_dps = [
+            Mapper(dp, self._name_ann, fn_kwargs=dict(name=name)) for name, dp in anns_dps.items()
+        ]
+        partial_anns_dp: IterDataPipe
+        for partial_anns_dp in partial_anns_dps:
+            anns_dp = KeyZipper(anns_dp, partial_anns_dp, lambda data: data[0], buffer_size=INFINITE_BUFFER_SIZE)
+            anns_dp = Mapper(anns_dp, self._collate_partial_anns)
```
ejguan, line 144: I will add an issue to data repo about the KeyZipper. This is a super anti-pattern design for stacked KeyZipper.

ejguan (approved):
LGTM. Just want to discuss if we can eliminate `CelebACSVParser` and use `CSVDictParser`

torchvision/prototype/datasets/_builtin/celeba.py
pmeier: We could, but there is more to it. Note that we also need to `map` the output so that it is a tuple with the image id first and the remaining row second. At this point we would probably have a more elaborate implementation bending everything to use the default building blocks than to just write our own. 

I agree it is annoying, but writing and understanding this custom parser is not hard so I feel its warranted.

torchvision/prototype/datasets/_builtin/celeba.py
pmeier: Let me see if the files are perfectly aligned. If yes, we could use a `Zipper` instead, which is able to handle more than 2 datapipes at once.

pmeier: They are perfectly aligned :tada: The latest commits only need two `KeyZippers` to zip the splits, images, and annotations. We can't get around them at the moment.

torchvision/prototype/datasets/_builtin/celeba.py
```diff
@@ -0,0 +1,184 @@
...
+        *,
+        has_header,
+    ):
+        self.datapipe = datapipe
+        self.has_header = has_header
+        self._fmtparams = dict(delimiter=" ", skipinitialspace=True)
+
+    def __iter__(self):
+        for _, file in self.datapipe:
+            file = (line.decode() for line in file)
```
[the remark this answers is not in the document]
pmeier: Yeah, that would be a lot better. I'll send a patch.

PR #4522, opened 2021-10-01
Tweak docs for better understanding of what spatial_scale does

alexander-soare:
Very minor thing and possibly just my opinion but I found the docs for spatial scale in the roi_pool variants misleading. Conceptually we are mapping the box co-ordinates to the input co-ordinates, not the other way around. So if `scale_factor = 0.5` we might consider that we are doing the mapping box -> box * 0.5. I only figured this out because I felt suspicious enough to open up a console and try it with a toy feature map. For argument's sake, you _could_ write it the other way around, but then you'd have to invert the scale factor for it to be semantically correct.

Also added an example to make it super clear.

Happy to get other opinions. If agreed, also happy to have this snuck in with a diff PR if convenient.


NicolasHug (approved):
Thanks @alexander-soare , I agree that this direction feels more natural, and the example is of great help here. Just for ref, in the code we indeed map from box coordinate to input coordinate:

```cpp
roi_start_w = offset_rois[1] * spatial_scale - offset;
```

`offset_rois` is in the "box coord space" and `roi_start_w` is in "input feature map coord space"

PR #4524, opened 2021-10-01
Remove prototype folder from release

fmassa:
Prototype features are not present in the release binaries


NicolasHug (approved):

pmeier (approved):
LGTM. For the future we should probably find a mechanism to exclude the folder by setting some environment variable. I know @mthrok had some ideas about this, but I can't find the comment.

fmassa:
Yes. IIRC @mthrok hasn't implemented this feature (only had the proposal for implementing it), so for now manually removing it is the recommended way IIRC

datumbox (approved):

mthrok:
> Yes. IIRC @mthrok hasn't implemented this feature (only had the proposal for implementing it), so for now manually removing it is the recommended way IIRC

Yup, that's the case.

PR #4525, opened 2021-10-01
add prototype for VOC dataset

pmeier:
cc @pmeier @mthrok @bjuncek


datumbox (approved):
LGTM, thanks.

torchvision/prototype/datasets/_builtin/voc.py
```diff
@@ -0,0 +1,141 @@
...
+                split=("train", "val", "test"),
+                year=("2012",),
+                task=("detection", "segmentation"),
+            ),
+        )
+
+    def resources(self, config: DatasetConfig) -> List[OnlineResource]:
+        if config.year == "2012":
+            if config.split == "train":
+                archive = HttpResource(
```
ejguan, line 51: I was looking to `HttpResource`. I am guessing I need to update `on_disk_cache` before you start to implement `HttpResource` to download and cache the Dataset in the `to_datapipe` method.

torchvision/prototype/datasets/_builtin/voc.py
```diff
@@ -0,0 +1,141 @@
...
+            functools.partial(self._classify_archive, config=config),  # type: ignore[arg-type]
+            drop_none=True,
+            buffer_size=INFINITE_BUFFER_SIZE,
+        )
+
+        split_dp = Filter(split_dp, self._is_in_folder, fn_kwargs=dict(name=self._SPLIT_FOLDER[config.task]))
+        split_dp = Filter(split_dp, path_comparator("name", f"{config.split}.txt"))
+        # TODO: replace the .map() call with decode=True in LineReader when
+        #  https://github.com/pytorch/data/issues/28 is resolved
+        split_dp = LineReader(split_dp).map(bytes.decode, input_col=1)
```
ejguan, line 129: nit: PR is landed and Issue is resolved.

torchvision/prototype/datasets/_builtin/voc.py
```diff
@@ -0,0 +1,141 @@
...
+        split_dp = Shuffler(split_dp, buffer_size=INFINITE_BUFFER_SIZE)
+
+        dp = split_dp
+        for level, data_dp in enumerate((images_dp, anns_dp)):
+            dp = KeyZipper(
+                dp,
+                data_dp,
+                key_fn=getitem(*[0] * level, 1),
```
ejguan, line 137: It seems there is another thing we could improve for `KeyZipper` is how to combine the value together. Currently, the result is`(split, image), ann`

IMO, Not sure if this is possible if it's better or doable to become `split, [image, ann]`.

ejguan (approved):
LGTM with some comments to discuss.

torchvision/prototype/datasets/_builtin/voc.py
pmeier: Right now the download functionality is not enabled and we rely on the files already being present in the right directory.

torchvision/prototype/datasets/_builtin/voc.py
ejguan: I have the same preference between these two approaches. Then only concern is that buffer is going to blow up if a DataPipe has the reverse order compared to the other DataPipe (buffer becomes M x N, where M represents the number of DataPipes and N represents the number of element). If we stack multiple `KeyZipper`, the same scenario as above that a single DataPipe has the reverse order would cause buffer size N rather than M x N.

pmeier: Zipping multiple datapipes shouldn't be the norm, so I feel like a warning in the documentation should be sufficient.

PR #4527, opened 2021-10-02
Fix annotation of draw_segmentation_masks

oke-aditya:
`draw_segmentation_masks` also supports passing single str containing color name or single str containing color hex or tuple containing RGB values as all colors for masks.

A simple example is our gallery :smiley: 

https://pytorch.org/vision/master/auto_examples/plot_repurposing_annotations.html#sphx-glr-auto-examples-plot-repurposing-annotations-py

I have updated the tests too which all check all the above cases.

cc @NicolasHug


oke-aditya:
I think that we should generate a color palette for `draw_bounding_boxes` too instead of drawing default white color.
If people really want white color boxes (Yuk) they can pass `color ="white"`

test/test_utils.py
```diff
@@ -197,6 +200,9 @@ def test_draw_segmentation_masks(colors, alpha):
     if colors is None:
         colors = utils._generate_color_palette(num_masks)
 
+    if isinstance(colors, str) or isinstance(colors, tuple):
```
NicolasHug, line 203:
```suggestion
    if colors is None:
        colors = utils._generate_color_palette(num_masks)
    elif isinstance(colors, str) or isinstance(colors, tuple):
```

torchvision/utils.py
```diff
@@ -288,18 +289,17 @@ def draw_segmentation_masks(
...
-        color = torch.tensor(color, dtype=out_dtype)
-        colors_.append(color)
+        colors_.append(torch.tensor(color, dtype=out_dtype))
 
     img_to_draw = image.detach().clone()
     # TODO: There might be a way to vectorize this
-    for mask, color in zip(masks, colors_):
-        img_to_draw[:, mask] = color[:, None]
+    for mask, color_tensor in zip(masks, colors_):
+        img_to_draw[:, mask] = color_tensor[:, None]
```
NicolasHug, line 297: In order to preserve git blame I would suggest to revert these changes, as I don't think they significantly improve readability either

torchvision/utils.py
```diff
@@ -242,8 +242,9 @@ def draw_segmentation_masks(
         masks (Tensor): Tensor of shape (num_masks, H, W) or (H, W) and dtype bool.
         alpha (float): Float number between 0 and 1 denoting the transparency of the masks.
             0 means full transparency, 1 means no transparency.
-        colors (list or None): List containing the colors of the masks. The colors can
-            be represented as PIL strings e.g. "red" or "#FF00FF", or as RGB tuples e.g. ``(240, 10, 157)``.
+        colors: Union[List[Union[str, Tuple[int, int, int]]], str, Tuple[int, int, int]]: List containing the colors
+            of the masks or single color for all masks. The colors can be represented as
+            PIL strings e.g. "red" or "#FF00FF", or as RGB tuples e.g. ``(240, 10, 157)``.
             When ``masks`` has a single entry of shape (H, W), you can pass a single color instead of a list
```
NicolasHug, line 248: let's just remove this sentence below then, it doesn't add much to the description now

> When ``masks`` has a single entry of shape (H, W), you can pass a single color instead of a list with one element

torchvision/utils.py
```diff
@@ -242,8 +242,9 @@ def draw_segmentation_masks(
         masks (Tensor): Tensor of shape (num_masks, H, W) or (H, W) and dtype bool.
         alpha (float): Float number between 0 and 1 denoting the transparency of the masks.
             0 means full transparency, 1 means no transparency.
-        colors (list or None): List containing the colors of the masks. The colors can
-            be represented as PIL strings e.g. "red" or "#FF00FF", or as RGB tuples e.g. ``(240, 10, 157)``.
+        colors: Union[List[Union[str, Tuple[int, int, int]]], str, Tuple[int, int, int]]: List containing the colors
```
NicolasHug, line 245: We should avoid using the annotation style in the docstrings, as they can be very hard to read and understand, especially this one. In this case we can keep things simple along with a good explanation in the description as you already wrote:
 
```suggestion
        colors (color or list of colors, optional): List containing the colors
```

NicolasHug (commented):
Thanks @oke-aditya , this looks good, I made a few comments below

torchvision/utils.py
oke-aditya: Problem is mypy. To satisfy mypy I had to change the name.

torchvision/utils.py
oke-aditya: We don't write this way in draw_bounding_boxes.
Do you want me to update there as well?

NicolasHug: it would be nice, thanks :)

torchvision/utils.py
NicolasHug: what is mypy complaining about?

oke-aditya: Mypy complains that `color` isn't a tensor, as we are re-using the name color.
Here is the exact error.

```
torchvision/utils.py:292: error: Incompatible types in assignment (expression
has type "Tensor", variable has type "Union[str, Tuple[int, int, int]]") 
[assignment]
            color = torch.tensor(color, dtype=out_dtype)
                    ^
torchvision/utils.py:298: error: Invalid index type "Tuple[slice, None]" for
"Union[str, Tuple[int, int, int]]"; expected type "Union[int, slice]"  [index]
            img_to_draw[:, mask] = color[:, None]
                                   ^
torchvision/utils.py:298: error: Invalid tuple index type (actual type
"Tuple[slice, None]", expected type "Union[int, slice]")  [misc]
            img_to_draw[:, mask] = color[:, None]
                                   ^
torchvision/utils.py:298: error: Incompatible types in assignment (expression
has type "Union[str, Any]", target has type
"Union[Tensor, Union[int, float, bool]]")  [assignment]
            img_to_draw[:, mask] = color[:, None]
                                   ^
Found 4 errors in 1 file (checked 127 source files)

Exited with code exit status 1

```

The easiest fix was to rename, rather than ignoring mypy errors.

NicolasHug: Ok, let's leave it as-is then.

@pmeier what is your take on this?

pmeier: Given that this is one of the most frustrating things when working with `mypy`, we could simply put `allow_redefinition = True` in our `mypy` config. I also proposed doing this in #4513, albeit only for the prototype datasets.

NicolasHug: done https://github.com/pytorch/vision/pull/4531/files

torchvision/utils.py
```diff
@@ -159,9 +159,9 @@ def draw_bounding_boxes(
             the boxes are absolute coordinates with respect to the image. In other words: `0 <= xmin < xmax < W` and
             `0 <= ymin < ymax < H`.
         labels (List[str]): List containing the labels of bounding boxes.
-        colors (Union[List[Union[str, Tuple[int, int, int]]], str, Tuple[int, int, int]]): List containing the colors
-            or a single color for all of the bounding boxes. The colors can be represented as `str` or
-            `Tuple[int, int, int]`.
+        colors: (color list of colors, optional): List containing the colors
```
NicolasHug, line 162:
```suggestion
        colors: (color or list of colors, optional): List containing the colors
```

torchvision/utils.py
```diff
@@ -242,10 +242,10 @@ def draw_segmentation_masks(
         masks (Tensor): Tensor of shape (num_masks, H, W) or (H, W) and dtype bool.
         alpha (float): Float number between 0 and 1 denoting the transparency of the masks.
             0 means full transparency, 1 means no transparency.
-        colors (list or None): List containing the colors of the masks. The colors can
-            be represented as PIL strings e.g. "red" or "#FF00FF", or as RGB tuples e.g. ``(240, 10, 157)``.
-            When ``masks`` has a single entry of shape (H, W), you can pass a single color instead of a list
-            with one element. By default, random colors are generated for each mask.
+        colors: (color list of colors, optional): List containing the colors
```
NicolasHug, line 245:
```suggestion
        colors: (color or list of colors, optional): List containing the colors
```

NicolasHug (commented):
Thanks @oke-aditya , LGTM with a last comment

NicolasHug (approved):
thanks, I'll merge when green

NicolasHug:
@oke-aditya we just started formatting the code with black - you'll get a few merge conflicts as you already noticed. Don't worry about it, I'll take care of it.

we'll soon publish proper instructions on how to merge these conflicts, but until then feel free to ignore them :)

sorry for the unexpected conflicts!

torchvision/utils.py
```diff
@@ -243,10 +243,10 @@ def draw_segmentation_masks(
         masks (Tensor): Tensor of shape (num_masks, H, W) or (H, W) and dtype bool.
         alpha (float): Float number between 0 and 1 denoting the transparency of the masks.
             0 means full transparency, 1 means no transparency.
-        colors (list or None): List containing the colors of the masks. The colors can
-            be represented as PIL strings e.g. "red" or "#FF00FF", or as RGB tuples e.g. ``(240, 10, 157)``.
-            When ``masks`` has a single entry of shape (H, W), you can pass a single color instead of a list
-            with one element. By default, random colors are generated for each mask.
+        colors: (color or list of colors, optional): List containing the colors
```
oke-aditya, line 246:
```suggestion
        colors (color or list of colors, optional): List containing the colors
```

NicolasHug:
@oke-aditya  if that's OK I'll merge https://github.com/pytorch/vision/pull/4531 first so we can remove the renaming from this PR, and then I'll merge it. Sounds good?

oke-aditya:
Yeah works fine :smiley:

NicolasHug:
Done! Thanks @oke-aditya

PR #4533, opened 2021-10-04
Relative import for roi_align

NicolasHug:
This is needed for https://github.com/pytorch/vision/pull/4384 and I think I need it to be a separate PR because it messes up with the sync tool


pmeier (approved):
Thanks!

Issue #4506, opened 2021-09-30
Fix flaky tests that are recently popping up


NicolasHug:
Since https://github.com/pytorch/vision/pull/4497 was merged, we're observing a few tests that start randomly failing.

Before https://github.com/pytorch/vision/pull/4497, these tests were almost always using the same RNG state, which was set in a test that was run earlier in the test execution suite. Now that all tests are properly independent and that the RNG doesn't leak, these tests run with a new RNG at each execution, and if they're unstable they might fail.

(Note: this is a good thing; it's better to know that they fail now rather than when submiting an unrelated PR, which is what happened in https://github.com/pytorch/vision/pull/3032#issuecomment-734829336)

For each of these tests we should find out whether the flakyness is severe or not. A simple solution is to parametrize the test over 100 or 1000 random seeds and check the failure rate. If the failure rate is reasonable we can just set a seed with `toch.manual_seed()`. If not, we should try to fix the test and make it more robust.

The list of tests so far is:

- [ ]  [test_random_apply](https://app.circleci.com/pipelines/github/pytorch/vision/10854/workflows/959a7090-eba7-4bfb-bc85-aaedbb408241/jobs/830887)
- [ ]  [test_stochastic_depth[row-0.2] - test.test_ops.TestStochasticDepth](https://app.circleci.com/pipelines/github/pytorch/vision/10854/workflows/959a7090-eba7-4bfb-bc85-aaedbb408241/jobs/830859)
- [ ]  [test_randomperspective_fill[L]](https://app.circleci.com/pipelines/github/pytorch/vision/10859/workflows/9d9ed379-d861-4752-bf9e-81dd9378e0d9/jobs/831307)
- [ ]  [test_randomperspective_fill[RGB]](https://app.circleci.com/pipelines/github/pytorch/vision/10859/workflows/9d9ed379-d861-4752-bf9e-81dd9378e0d9/jobs/831307)
- [ ]  [test_randomperspective_fill[F]](https://app.circleci.com/pipelines/github/pytorch/vision/10859/workflows/9d9ed379-d861-4752-bf9e-81dd9378e0d9/jobs/831307)
- [ ]  [test_random_vertical_flip](https://app.circleci.com/pipelines/github/pytorch/vision/10859/workflows/9d9ed379-d861-4752-bf9e-81dd9378e0d9/jobs/831307)
[...]

PR #4535, opened 2021-10-04
Fix all flake8 issues

NicolasHug:
This fixes all outstanding flake8 issues that popped up after https://github.com/pytorch/vision/pull/4384 (https://github.com/pytorch/vision/pull/4384#issuecomment-933474723)

Most of it is just removing unused imports. I stayed conservative (i.e. keep the import with a `# noqa`) when I wasn't sure.

It also ignores the `line-too-long` errors from flake8: those are taken care of by black now anyway, but there can be minor conflicts in the case e.g. of https://github.com/pytorch/vision/blob/main/torchvision/models/detection/ssdlite.py#L21:L21, so we just keep `flake8` quite and trust `black` in those cases.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -8,8 +8,9 @@ license_file = LICENSE
 max-line-length = 120
 
 [flake8]
+# note: we ignore all 501s (line too long) anyway as they're taken care of by black
 max-line-length = 120
-ignore = E203, E402, W503, W504, F821
+ignore = E203, E402, W503, W504, F821, E501
 per-file-ignores =
     __init__.py: F401, F403, F405
     ./hubconf.py: F401
--- a/test/test_datasets_video_utils_opt.py
+++ b/test/test_datasets_video_utils_opt.py
@@ -1,7 +1,6 @@
 import unittest
 
 import test_datasets_video_utils
-from torchvision import set_video_backend
 
 # Disabling the video backend switching temporarily
 # set_video_backend('video_reader')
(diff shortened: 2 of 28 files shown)


test/test_datasets_video_utils_opt.py
```diff
@@ -1,7 +1,6 @@
 import unittest
 
 import test_datasets_video_utils
-from torchvision import set_video_backend
```
pmeier, line 4: I think commenting it would be better, because this will be needed if we remove the "temporary" issue below.

test/test_io_opt.py
```diff
@@ -1,7 +1,7 @@
 import unittest
 
 import test_io
-from torchvision import set_video_backend
+from torchvision import set_video_backend  # noqa
```
pmeier, line 4: In any case we should be consistent. You removed this line in my comment above, although the setup is the same.

pmeier (approved):
Some minor comments, otherwise LGTM!

setup.cfg
```diff
@@ -8,8 +8,9 @@ license_file = LICENSE
 max-line-length = 120
 
 [flake8]
+# note: we ignore all 501s (line too long) anyway as they're taken care of by black
```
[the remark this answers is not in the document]
NicolasHug: this comment just refers to the line below which is 

>max-line-length = 120

So the comment essentially says that the line is useless, but I think it's worth keeping anyway

NicolasHug:
The `test_frozenbatchnorm2d_eps` errors are unrelated and are tracked in https://github.com/pytorch/vision/issues/4506

PR #4518, opened 2021-10-01
add prototype for SBD dataset

pmeier:
cc @pmeier @mthrok @bjuncek

PR #4537, opened 2021-10-05
add prototype for SBD dataset

pmeier:
Redo of #4518, because I messed up the branch

cc @pmeier @mthrok @bjuncek


torchvision/prototype/datasets/_builtin/sbd.py
```diff
@@ -0,0 +1,179 @@
...
+            )
+        return Mapper(dp, self._collate_and_decode_sample, fn_kwargs=dict(config=config, decoder=decoder))
+
+    def generate_categories_file(self, root: Union[str, pathlib.Path]) -> None:
+        dp = self.resources(self.default_config)[0].to_datapipe(pathlib.Path(root) / self.name)
+        dp = TarArchiveReader(dp)
+        dp = Filter(dp, path_comparator("name", "category_names.m"))
+        dp = LineReader(dp)
+        dp = Mapper(dp, bytes.decode, input_col=1)
+        lines = tuple(zip(*iter(dp)))[1]
```
ejguan, line 160: This is seems also a part `Mapper` can improve by only keeping the result of function and dropping all other columns.

ejguan: Maybe achieved by adding the `__getitem__` for `IterDataPipe` to select columns.

torchvision/prototype/datasets/_builtin/sbd.py
```diff
@@ -0,0 +1,179 @@
...
+                dp,
+                data_dp,
+                key_fn=getitem(*[0] * level, 1),
+                ref_key_fn=path_accessor("stem"),
+                buffer_size=INFINITE_BUFFER_SIZE,
+            )
+        return Mapper(dp, self._collate_and_decode_sample, fn_kwargs=dict(config=config, decoder=decoder))
+
+    def generate_categories_file(self, root: Union[str, pathlib.Path]) -> None:
+        dp = self.resources(self.default_config)[0].to_datapipe(pathlib.Path(root) / self.name)
```
ejguan, line 155: As resources are just fixed files, just wondering why this function is attached to Dataset class. Do we expect users would call this method in the future to generate the category files?

ejguan (approved):
Thanks for writing the prototype.

torchvision/prototype/datasets/_builtin/sbd.py
pmeier: Note that user will never see a `Dataset` object during "normal" operation. The public API `torchvision.prototype.datasets.load` returns a datapipe

https://github.com/pytorch/vision/blob/b1ef19639ee845902ecf60930757d028f1e1491c/torchvision/prototype/datasets/_api.py#L74

torchvision/prototype/datasets/_builtin/sbd.py
ejguan: The main goal is to somehow align the API with TorchArrow DataFrame since DataFrame normally can select column by `df['x']`.

PR #4538, opened 2021-10-05
fix import order for torchvision.prototype.datasets

pmeier:
This was broken by #4384, but not detected since we don't have tests for the prototype functionality yet.

### DIFF
--- a/torchvision/prototype/datasets/__init__.py
+++ b/torchvision/prototype/datasets/__init__.py
@@ -7,10 +7,9 @@ except (ModuleNotFoundError, TypeError) as error:
         "Note that you cannot install it with `pip install torchdata`, since this is another package."
     ) from error
 
-
 from . import decoder, utils
+from ._home import home
 
 # Load this last, since some parts depend on the above being loaded first
-from ._api import register, _list as list, info, load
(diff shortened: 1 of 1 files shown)


torchvision/prototype/datasets/__init__.py
```diff
@@ -7,10 +7,9 @@
         "Note that you cannot install it with `pip install torchdata`, since this is another package."
     ) from error
 
-
 from . import decoder, utils
+from ._home import home
 
 # Load this last, since some parts depend on the above being loaded first
-from ._api import register, _list as list, info, load
+from ._api import register, _list as list, info, load  # usort: skip
```
NicolasHug, line 14: Just curious, what's the reason we need to load these after _home?

NicolasHug (approved):
LGTM when green, thanks

torchvision/prototype/datasets/__init__.py
pmeier: To avoid cyclic imports. Importing anything from `_api` here will try to import `home`:

https://github.com/pytorch/vision/blob/0c0a6a448a001b91e0d12f50225b68372bed2354/torchvision/prototype/datasets/_api.py#L6

PR #4541, opened 2021-10-05
add prototype for HMDB51 dataset

pmeier:
The idea is that the dataset will always return the complete video. For the common use case of working with clips of a fixed number of frames, I propose we load the `dataset` with `decoder=None` and provide a custom `IterDataPipe` (follow-up PR) that handles all of this.

cc @pmeier @mthrok @bjuncek


torchvision/prototype/datasets/decoder.py
```diff
@@ -1,11 +1,23 @@
...
+def pil(buffer: io.IOBase, *, mode: str = "RGB") -> Dict[str, Any]:
+    return dict(image=pil_to_tensor(PIL.Image.open(buffer).convert(mode.upper())))
+
+
+def av(buffer: io.IOBase, **read_video_kwargs: Any) -> Dict[str, Any]:
+    with unittest.mock.patch("torchvision.io.video.os.path.exists", return_value=True):
+        return dict(
+            zip(
+                ("video", "audio", "video_meta"),
+                read_video(buffer, **read_video_kwargs),  # type: ignore[arg-type]
```
fmassa, line 21: I want us to step back and think a bit more about this. We have the opportunity to improve on the video datasets now that we are using IterDataPipes.

PR #4544, opened 2021-10-06
Fix bug on EMA n_averaged estimation

datumbox:
The bug doesn't affect the accuracy of the technique but it messes up the counter of how many updates we did to the weights. Fixing to cherry-pick on the upcoming release branch.


NicolasHug (approved):
stamping

prabhat00155 (approved):
Looks good, thanks @datumbox!

PR #4545, opened 2021-10-06
Fix bug on EMA n_averaged estimation. (#4544)

datumbox:
Cherrypicking #4544 to the release branch.

cc @datumbox


prabhat00155 (approved):
Thanks @datumbox!

PR #4551, opened 2021-10-06
[ao_migration] torchvision: torch.quantization -> torch.ao.quantization

z-a-f:
Summary:
This changes the imports in the `pytorch/vision` to include the new import locations.

```
codemod -d pytorch/vision --extensions py 'torch.quantization' 'torch.ao.quantization'
```

Differential Revision: D31302368


z-a-f:
OK, I am closing the internal diff as well, assuming that @NicolasHug will take care of the AO migration within the torchvision :)

NicolasHug:
@z-a-f my intent with #4554 was to ease the Pr submission process, but if there are any additional changes I believe I might not be the best suited person for this. I have zero knowledge of the ao module.

Would you mind following up on my comment in #4554, namely:

> I just did some brutal sed -i "s/.../..." so I might have done something wrong. @z-a-f please double-check that these are the intended changes :)

and 

> @z-a-f , are the functions / classes in toch.ao.quantization exactly the same as in torch.quantization, or should we expect some different results? Some of the changes here are editing the training references, so we need to make sure that everything still works as before in terms of model accuracy etc.

PR #4554, opened 2021-10-07
Use torch.ao.quantization instead of torch.quantizaiton

NicolasHug:
Same as #4551 but without black conflicts

I just did some brutal `sed -i "s/.../..."` so I might have done something wrong. @z-a-f please double-check that these are the intended changes :)


NicolasHug:
We expect the torchvision main branch to work with pytorch nightly, we don't need it to be compatible with previous (or even currently stable) versions

z-a-f:
The reason https://app.circleci.com/pipelines/github/pytorch/vision/11088/workflows/c3d06578-6137-4d13-8f38-7342e02c2148/jobs/853515 is failing is because there is a dependency on 1.11.dev09+.

datumbox:
I tried to investigate why this is failing. It seems that [cmake on Windows with GPU](https://app.circleci.com/pipelines/github/pytorch/vision/12463/workflows/3a637bff-870b-4607-adcc-11b5ec54cb13/jobs/984472) installs a very old nightly of PyTorch (`1.11.0.dev20210929`) that possibly doesn't contain the necessary change on the quantization namespace. 

It's worth noting that the [Windows Cmake with CPU](https://app.circleci.com/pipelines/github/pytorch/vision/12463/workflows/3a637bff-870b-4607-adcc-11b5ec54cb13/jobs/984492) picks up today's nightly (`1.11.0.dev20211111`) and works without issues.

Upon further investigation, it seems that the last available release of win-64 with CUDA 10.2 happened on [20210929](https://anaconda.org/pytorch-nightly/pytorch). Since Cmake builds using CUDA 10.2, we are having issues.

datumbox:
This should now hopefully be resolved after #4945

Issue #4540, opened 2021-10-05
Replace IntermediateLayerGetter with feature_extraction.create_feature_extractor()


datumbox:
### 🚀 The feature

A new feature extraction utility based on FX was introduced by #4302. It allows us to extract internal outputs of a module in a simpler way and it should be a drop-in replacement of the old `IntermediateLayerGetter` class.

Unfortunately the `create_feature_extractor()` util is not currently used anywhere on vision. We should investigate using it to replace some of the old approaches. Here is a list of candidates:
- [ ] Replace `IntermediateLayerGetter` calls in Segmentation models [[1](https://github.com/pytorch/vision/blob/5f0edb97b46e5bff71dc19dedef05c5396eeaea2/torchvision/models/segmentation/segmentation.py#L63), [2](https://github.com/pytorch/vision/blob/5f0edb97b46e5bff71dc19dedef05c5396eeaea2/torchvision/models/segmentation/segmentation.py#L119)]
- [ ] Replace `IntermediateLayerGetter` calls in Detection models [[3](https://github.com/pytorch/vision/blob/5f0edb97b46e5bff71dc19dedef05c5396eeaea2/torchvision/models/detection/backbone_utils.py#L37)]
- [ ] Rewrite model surgery from the `SSDFeatureExtractorVGG` [[4](https://github.com/pytorch/vision/blob/5f0edb97b46e5bff71dc19dedef05c5396eeaea2/torchvision/models/detection/ssd.py#L441)] and `SSDLiteFeatureExtractorMobileNet` modules [[5](https://github.com/pytorch/vision/blob/5f0edb97b46e5bff71dc19dedef05c5396eeaea2/torchvision/models/detection/ssdlite.py#L125-L129)].

Extra care needs to be given to ensure that the old weights of pre-trained models can still load.

### Motivation, pitch

Using the latest FX-based utility will help us future proof our code-base.

kingjuno:
@datumbox can i work on this issue?

datumbox:
@kingjuno There are a few limitations of using the FX-based approach (losing class info) which means replacing the SSD ones might not work. I'm currently OOO, but I can look into if this still makes sense once I'm back.

kingjuno:
ok

datumbox:
[...]

PR #4563, opened 2021-10-07
Use FX feature extractor for segm model

vfdev-5:
Description:
- Use FX feature extractor for segm model

Related to https://github.com/pytorch/vision/issues/4540

cc @datumbox


vfdev-5:
@datumbox OK, sounds good to remove `use_fe` option. Removed it.

Issue #4245, opened 2021-08-04
Gallery example backlinks aren't shown on the master documentation


NicolasHug:
The backlinks are those "This function/class is used in example XYZ" parts in the docstrings. They're properly present in the [stable docs](https://pytorch.org/vision/stable/transforms.html#transforms-on-pil-image-and-torch-tensor) but they don't appear on the [master docs](https://pytorch.org/vision/master/transforms.html#transforms-on-pil-image-and-torch-tensor) anymore.

Possibly due to a sphinx-gallery version... I'll need to look into this

PR #4565, opened 2021-10-07
Fix sphinx-gallery backlinks

NicolasHug:
Trying to fix https://github.com/pytorch/vision/issues/4245 by disabling the cache

Disabling the cache isn't ideal but it's not a big deal right now, the build_docs CI job only takes 10 minutes so it's not what's bringing us down.

I'll follow-up after the release to properly fix this and re-enable the cache

### DIFF
--- a/torchvision/prototype/datasets/benchmark.py
+++ b/torchvision/prototype/datasets/benchmark.py
@@ -1,0 +1,403 @@
+import argparse
+import contextlib
+import copy
+import inspect
+import itertools
+import os
+import os.path
+import pathlib
+import shutil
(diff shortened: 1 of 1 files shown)


NicolasHug:
Looks like this works: the backlinks exist on this PR https://857563-73328905-gh.circle-artifacts.com/0/docs/utils.html

whereas they don't in this recent one https://github.com/pytorch/vision/pull/3808: https://857412-73328905-gh.circle-artifacts.com/0/docs/utils.html

let's merge this ASAP and to include it in the release?

PR #4574, opened 2021-10-08
Added docs for reference scripts

NicolasHug:
Some basic docs for the references scripts. Any additional thing we should mention here?


NicolasHug:
Docs looking good, merging, thanks for the reviews!

Issue #4577, opened 2021-10-08
Make Dropout rates configurable on all Model Architectures


datumbox:
### 🚀 The feature

Introduce an extra `dropout` parameter to all model constructors that contain a Dropout layer on the classifier:
- [x] [AlexNet](https://github.com/pytorch/vision/blob/aaee8ff1f75941dd2be82a180826a052965ce84c/torchvision/models/alexnet.py#L37-L40)
- [x] [GoogleNet](https://github.com/pytorch/vision/blob/aaee8ff1f75941dd2be82a180826a052965ce84c/torchvision/models/googlenet.py#L122)
- [x] [InceptionV3](https://github.com/pytorch/vision/blob/aaee8ff1f75941dd2be82a180826a052965ce84c/torchvision/models/inception.py#L118)
- [x] [MobileNetV2](https://github.com/pytorch/vision/blob/aaee8ff1f75941dd2be82a180826a052965ce84c/torchvision/models/mobilenetv2.py#L164)
- [x] [MobileNetV3](https://github.com/pytorch/vision/blob/aaee8ff1f75941dd2be82a180826a052965ce84c/torchvision/models/mobilenetv3.py#L203)
- [x] [SqueezeNet](https://github.com/pytorch/vision/blob/aaee8ff1f75941dd2be82a180826a052965ce84c/torchvision/models/squeezenet.py#L80)
- [x] [VGG](https://github.com/pytorch/vision/blob/aaee8ff1f75941dd2be82a180826a052965ce84c/torchvision/models/vgg.py#L42-L45)

The introduction should happen in a backwards compatible manner, maintaining the same probability values as before by default.

### Motivation, pitch

Many of the supported model architectures contain dropout layers with hardcoded probability values. 

Unfortunately there are cases where using Dropout is not desirable or its p value needs to be adjusted. Though this can be achieved with model surgery, it's also quite straightforward to add on the constructor of he models a single parameter to give better control. In fact, some of the models already support this. See:
https://github.com/pytorch/vision/blob/aaee8ff1f75941dd2be82a180826a052965ce84c/torchvision/models/mnasnet.py#L98

https://github.com/pytorch/vision/blob/aaee8ff1f75941dd2be82a180826a052965ce84c/torchvision/models/efficientnet.py#L153

[...]

PR #4580, opened 2021-10-09
Added Dropout parameter to Models Constructors

ABD-01:
Hi!
This pr is with reference to https://github.com/pytorch/vision/issues/4577

I have added the dropout parameter to following model constructors

* [X] AlexNet
* [X] GoogleNet
 * [X] InceptionV3
 * [X] MobileNetV2
 * [X] MobileNetV3
 * [x] SqueezeNet
 * [X] VGG

For GoogleNet there's dropout used in [auxiliary inception net class](https://github.com/pytorch/vision/blob/aaee8ff1f75941dd2be82a180826a052965ce84c/torchvision/models/googlenet.py#L287) so I've added a parameter named `dropout_aux` for that, but let me know if that is not needed.

Please do comment for the required changes, and other things to be done.
Thank You

cc @datumbox


oke-aditya:
It's slightly subjective, I would suggest waiting a bit for maintainers to have a look. :sweat_smile:

datumbox:
@ABD-01 Just use named arguments aka `dropout= dropout_aux)`

ABD-01:
Thank you very much! Always happy to learn 🐣

PR #4585, opened 2021-10-10
update python syntax >=3.6

Borda:
In fact, I was just adding a new hook to pre-commit `pyupgrade` which generated all these changes
also I have added `python_requires=">=3.6",` to the seup.py

An alternative would be enable pre-commit bot and then it would be clear that most of the changes comes just from the pre-commit hooks, see https://pre-commit.ci/

### DIFF
--- a/torchvision/io/_video_opt.py
+++ b/torchvision/io/_video_opt.py
@@ -73,7 +73,11 @@ def _validate_pts(pts_range):
     if pts_range[1] > 0:
         assert (
             pts_range[0] <= pts_range[1]
-        ), f"Start pts should not be smaller than end pts, got start pts: {pts_range[0]:d} and end pts: {pts_range[1]:d}"
+        ), """Start pts should not be smaller than end pts, got
+            start pts: {0:d} and end pts: {1:d}""".format(
+            pts_range[0],
+            pts_range[1],
+        )
 
 
 def _fill_info(vtimebase, vfps, vduration, atimebase, asample_rate, aduration):


Borda:
> One thing I'm not sure about is the commit hook, considering that some of these changes are making torchscript unhappy (see comment below), we might not be able to enforce this with a hook.

I may have related thought as each statement (code line) can be written in many ways, is the actual optimal, personally, I would keep any of these hook/bots and try to rewrite some small number of lines to complain of hook and also to be fine with you and `torchscript`... what do you think?

Borda:
> We do enforce black-formatting now, since last week.

do you have it also in pre-commit hooks?
I see this one https://github.com/pytorch/vision/blob/221dd5f76c206d2d9c93dd910c71007bd6791082/.pre-commit-config.yaml#L5
but I don't think that it is the optimal one, rather use
```yaml
  - repo: https://github.com/psf/black
    rev: 21.9b0
    hooks:
      - id: black
        name: Format code
```
I can add it in separate PR

NicolasHug:
`ufmt` is a combination of `black` and `usort`, we have to rely on this (and not on the individual formatters) because of compatibility reasons with FB's internal linters.

Borda:
> Reference scripts are not tested by unit tests, did you verify that everything runs fine?

seems they require multi-node/GPU machines which I do not have in my disposal :(

Borda:
seems to be failing on Segmentation fault :/

datumbox:
@Borda It's not your PR, our CI is toasted... It should be resolved tomorrow using the latest nightly of PyTorch. Apologies for the inconvenience.

PR #4590, opened 2021-10-11
Fixing bug on SSD backbone freezing

### LINKED ISSUE #4548: `trainable_backbone_layers` in `ssd300_vgg16` seems to work wrong
### 🐛 Describe the bug

I called `torchvision.models.detection.ssd300_vgg16` with different values for `trainable_backbone_layers` and it seems to freeze more layers than expected.

### To Reproduce
I took a part of the script from this #3905 issue:
``` python
import torch
import torchvision

n_trainable_params = []
for trainable_layers in range(6):
    model = torchvision.models.detection.ssd300_vgg16(pretrained=False,
            pretrained_backbone=True,
            trainable_backbone_layers=trainable_layers)

    n_trainable_params.append(len([p for p in model.parameters() if p.requires_grad]))

print(f"SSD number of trainable parameters: {n_trainable_params}")

print(f"SSD total number of parameters: {len([p for p in model.parameters()])}")
```

Output:
```
SSD number of trainable parameters: [45, 45, 51, 57, 63, 67]
SSD total number of parameters: 71
```
So we have the same number of trainable parameters when setting `trainable_backbone_layers` to 0 and 1.
And even when we set `trainable_backbone_layers` to 5, the number of trainable parameters is less than their total number.

Looks like it works correctly when setting `trainable_backbone_layers` to 0, but in the other cases it freezes one layer more than needed.
 
I think the problem is with these two lines: https://github.com/pytorch/vision/blob/4b6fc6b8e8b9505a1c6c8d89937123ed34ec83c4/torchvision/models/detection/ssd.py#L535 https://github.com/pytorch/vision/blob/4b6fc6b8e8b9505a1c6c8d89937123ed34ec83c4/torchvision/models/detection/ssd.py#L540

### Expected behavior
The number of trainable parameters should be greater when `trainable_backbone_layers` is set to a value greater than 0.

### Additional notes
In function header `trainable_backbone_layers` is set to `None` by default:
https://github.com/pytorch/vision/blob/4b6fc6b8e8b9505a1c6c8d89937123ed34ec83c4/torchvision/models/detection/ssd.py#L549-L556
But a few lines below, it defaults to 5 implicitly: https://github.com/pytorch/vision/blob/4b6fc6b8e8b9505a1c6c8d89937123ed34ec83c4/torchvision/models/detection/ssd.py#L604-L606 
This confused me somewhat when I tried to train this model on my own dataset, because I thought (for no reason) the entire backbone was frozen by default. Maybe, for more transparency, it would be better to explicitly set the `trainable_backbone_layers` to 5 by default in the header?

### Versions

Google Colab environment:
```
PyTorch version: 1.9.0+cu102
Is debug build: False

OS: Ubuntu 18.04.5 LTS (x86_64)
Clang version: 6.0.0-1ubuntu2 (tags/RELEASE_600/final)
CMake version: version 3.12.0
Libc version: glibc-2.26

Python version: 3.7.12 (default, Sep 10 2021, 00:21:48)  [GCC 7.5.0] (64-bit runtime)
Python platform: Linux-5.4.104+-x86_64-with-Ubuntu-18.04-bionic
Is CUDA available: False
CUDA runtime version: 11.1.105
GPU models and configuration: Could not collect
Nvidia driver version: Could not collect
cuDNN version: Probably one of the following:
/usr/lib/x86_64-linux-gnu/libcudnn.so.7.6.5
/usr/lib/x86_64-linux-gnu/libcudnn.so.8.0.5
/usr/lib/x86_64-linux-gnu/libcudnn_adv_infer.so.8.0.5
/usr/lib/x86_64-linux-gnu/libcudnn_adv_train.so.8.0.5
/usr/lib/x86_64-linux-gnu/libcudnn_cnn_infer.so.8.0.5
/usr/lib/x86_64-linux-gnu/libcudnn_cnn_train.so.8.0.5
/usr/lib/x86_64-linux-gnu/libcudnn_ops_infer.so.8.0.5
/usr/lib/x86_64-linux-gnu/libcudnn_ops_train.so.8.0.5
HIP runtime version: N/A
MIOpen runtime version: N/A

[pip3] numpy==1.19.5
[pip3] torch==1.9.0+cu102
[pip3] torchsummary==1.5.1
[pip3] torchtext==0.10.0
[pip3] torchvision==0.10.0+cu102
[conda] Could not collect
```

cc @datumbox

datumbox:
Fixes #4548

There was a bug on the way we counted the frozen layers. This mean that the first "block" (layers 0-3 inclusive) of VGG was frozen. Here is how to confirm this is the case on the current main branch:
```python
from torchvision.models.detection import *

freeze_before = 4
trained = ssd300_vgg16(pretrained=True)
nottrained = ssd300_vgg16(pretrained=False, pretrained_backbone=True)
sum_trained = sum(x.sum() for x in trained.backbone.features[:freeze_before].parameters())
sum_nottrained = sum(x.sum() for x in nottrained.backbone.features[:freeze_before].parameters())

assert sum_trained == sum_nottrained
```

This is how we can confirm that we freeze the right number of params:
```python
import torchvision

n_trainable_params = []
for trainable_layers in range(6):
    model = torchvision.models.detection.ssd300_vgg16(pretrained=False,
            pretrained_backbone=True,
            trainable_backbone_layers=trainable_layers)

    n_trainable_params.append(len([p for p in model.parameters() if p.requires_grad]))

print("SSD:")
print(f"Number of trainable parameters: {n_trainable_params}")
print(f"Total number of parameters: {len([p for p in model.parameters()])}")
```

Results:
```
SSD:
Number of trainable parameters: [45, 51, 57, 63, 67, 71]
Total number of parameters: 71
```

cc @datumbox

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -41,9 +41,10 @@ architectures, and common image transformations for computer vision.
 
 .. toctree::
    :maxdepth: 1
-   :caption: Examples
+   :caption: Examples and training references
 
    auto_examples/index
+   training_references
 
 .. automodule:: torchvision
    :members:
--- a/docs/source/models.rst
+++ b/docs/source/models.rst
@@ -1,3 +1,5 @@
+.. _models:
+
 torchvision.models
 ##################
 
--- a/docs/source/training_references.rst
+++ b/docs/source/training_references.rst
@@ -1,0 +1,29 @@
+Training references
+===================
+
+On top of the many models, datasets, and image transforms, Torchvision also
+provides training reference scripts. These are the scripts that we use to train
+the :ref:`models <models>` which are then available with pre-trained weights.
+
+These scripts are not part of the core package and are instead available `on
+GitHub <https://github.com/pytorch/vision/tree/main/references>`_. We currently
+provide references for
+`classification <https://github.com/pytorch/vision/tree/main/references/classification>`_,
+`detection <https://github.com/pytorch/vision/tree/main/references/detection>`_,
+`segmentation <https://github.com/pytorch/vision/tree/main/references/segmentation>`_,
+`similarity learning <https://github.com/pytorch/vision/tree/main/references/similarity>`_,
+and `video classification <https://github.com/pytorch/vision/tree/main/references/video_classification>`_.
+
+While these scripts are largely stable, they do not offer backward compatibility
+guarantees.
+
+In general, these scripts rely on the latest (not yet released) pytorch version
+or the latest torchvision version. This means that to use them, **you might need
+to install the latest pytorch and torchvision versions**, with e.g.::
+
+    conda install pytorch torchvision -c pytorch-nightly
+
+If you need to rely on an older stable version of pytorch or torchvision, e.g.
+torchvision 0.10, then it's safer to use the scripts from that corresponding
+release on GitHub, namely
+https://github.com/pytorch/vision/tree/v0.10.0/references.
--- a/references/classification/README.md
+++ b/references/classification/README.md
@@ -110,13 +110,13 @@ torchrun --nproc_per_node=8 train.py\
 Here `$MODEL` is one of `regnet_x_32gf`, `regnet_y_16gf` and `regnet_y_32gf`.
 
 ## Mixed precision training
-Automatic Mixed Precision (AMP) training on GPU for Pytorch can be enabled with the [NVIDIA Apex extension](https://github.com/NVIDIA/apex).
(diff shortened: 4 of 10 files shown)


datumbox:
@NicolasHug Thanks for the stamps. 

I definitely agree we should write tests for the `trainable_backbone_layers` parameter of all detection models. The challenge is that re-initializing the models is quite slow and likely to impact significantly our execution speed. I'll create a separate issue to investigate how we could do this.

Issue #4593, opened 2021-10-12
Adding Vision Transformer to torchvision/models


sallysyw:
### 🚀 The feature

1. Adding ViT architecture from this paper: "[An Image is Worth 16x16 Words: Transformers for Image Recognition at Scale](https://arxiv.org/abs/2010.11929)"
2. Adding DeiT architecture from this paper: "[Training data-efficient image transformers & distillation through attention](https://arxiv.org/abs/2012.12877)"

@fmassa @datumbox @mannatsingh @kazhang 

### Motivation, pitch

Vision Transformer models should exist in torchvision repo because they are good models :)

I'm currently working on this project.

### Additional context

We can also consider adding some techniques from the following papers  ^^
For example, adding Conv stem for ViT, see details in "[Early Convolutions Help Transformers See Better](https://arxiv.org/abs/2106.14881)" 

References:
https://github.com/google-research/vision_transformer
https://github.com/facebookresearch/deit
https://github.com/facebookresearch/ClassyVision/blob/main/classy_vision/models/vision_transformer.py

cc @datumbox

sallysyw:
> Note that DeiT has the same architecture as ViT - it's the same model, only the training setup is different!

Interesting! I was looking at the implementation here
https://github.com/facebookresearch/deit/blob/main/models.py#L20
and found that DeiT model actually inherits the ViT model class. 

Does it make sense if I also do this in torchvision?

take2rohit:
I believe that having ViT/DeiT in torchvision library would be really useful!
So is anyone else working with the implementation or should I go ahead and create a PR?

datumbox:
@take2rohit Thanks. @sallysyw is working on it at PR #4594.

PR #4594, opened 2021-10-12
Adding ViT to torchvision/models

sallysyw:
The first part of #4593 :)


sallysyw:
I've finished the first iteration of training the `vit_b_32` and `vit_b_16` models from scratch using the ImageNet dataset, here are the numbers I got:
| Model    	| Epochs 	| Nodes 	| BatchSizePerGPU 	| GlobalBatchSize 	| Original Acc@1 	| ClassyVision Acc@1 	| Acc@1  	| Acc@5  	|
|----------	|--------	|-------	|-----------------	|-----------------	|----------------	|--------------------	|--------	|--------	|
| vit_b_32 	| 300    	| 2     	| 256             	| 4096            	| 73.38          	| 73.3               	| 71.876 	| 89.784 	|
| vit_b_16 	| 300    	| 8     	| 64              	| 4096            	| 77.91          	| 78.98              	| 76.854 	| 92.959 	|

I plan to further tune the parameters and I'll upload the pre-trained weights in a following PR once I got the accuracy numbers matching the previous results.
I have also checked the number of params in these models and they are exactly the same as the ClassyVision models.

Let me know if there's other concerns before I can merge this PR.
@datumbox @mannatsingh

sallysyw:
> > @sallysyw I haven't checked the validity of the ViT implementation itself comparing to the paper. I know you are porting this from trusted sources, but if you want me to check it let me know.
> 
> This is a great point! If you train from scratch and the axes are messed up, you still get reasonable results sometimes (speaking from experience lol). We should try and maybe repro a result from here - https://github.com/facebookresearch/ClassyVision/tree/main/examples/vit
> 
> Or if it's easier, we can use a pretrained model from any source (like Classy), and evaluate it with this implementation and verify that the accuracy matches!

Do you know where can I find some pre-trained classy-vision checkpoints?

datumbox:
@sallysyw Thanks for the update.

As far as I can see the unit-tests are [failing](https://app.circleci.com/pipelines/github/pytorch/vision/12125/workflows/9e096b2a-57da-463f-b7f2-fa120708705f/jobs/951770) and it seems related. From what I can see the FX feature extractor seg faults on vit_b_16. This needs to be fixed before we merge:
```
test/test_backbone_utils.py::TestFxFeatureExtraction::test_jit_forward_backward[vit_b_16] Windows fatal exception: access violation
```

> Do you know where can I find some pre-trained classy-vision checkpoints?

This is a good idea to do. I don't know where you could find one (worth checking with internal teams to see if they have checkpoints you can use) but it's definitely a necessary step prior merging. I would also recommend, unless you have other important reasons, to fully reproduce the achieved accuracies before merging. In the past, there were cases were the cause of the drop in accuracy was identified in the architecture itself. Though you can mitigate that risk by loading pre-trained checkpoints and reproducing their accuracies to the digit, this doesn't cover for bugs for when the model is on training mode.

> Let me know if there's other concerns before I can merge this PR.

Some of the above comments, for example Francisco's https://github.com/pytorch/vision/pull/4594#discussion_r734350806 and https://github.com/pytorch/vision/pull/4594#discussion_r734385444, affect the architecture so if you merge now you might have to make radical changes later. Given that the next release is months from now, we got time to fix them. Still given that there are outstanding comments, it would be good to confirm with @mannatsingh and @fmassa that an early merge is OK in this case. Another option is to merge this on `prototype` where the criteria are more relaxed. Thoughts?

torchvision/models/vision_transformer.py
```diff
@@ -0,0 +1,356 @@
...
+        x = self.encoder(x)
+
+        if self.classifier == "token":
+            # Classifier as used by standard language architectures
+            x = x[0, :, :]
+        elif self.classifier == "gap":
+            # Classifier as used by standard vision architectures
+            x = x.mean(dim=0)
+        else:
+            raise ValueError(f"Invalid classifier={self.classifier}")
```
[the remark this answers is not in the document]
fmassa: > but for this PR let me keep things simple and I'll just follow what has been done in the original github repo & classy implementation. wdyt?

I'm ok following it this way if we keep this implementation as prototype (and thus move it to the `prototype` folder.
I think it really pays off to think twice about the APIs before making it public, as breaking public APIs is generally not great.

If I were to go with one approach, I would start with only the `token` mode, remove `classifier` altogether and then revisit this in the future.

torchvision/models/vision_transformer.py
```diff
@@ -0,0 +1,353 @@
...
+        # (n, c, h, w) -> (n, hidden_dim, n_h, n_w)
+        x = self.conv_proj(x)
+        # (n, hidden_dim, n_h, n_w) -> (n, hidden_dim, (n_h * n_w))
+        x = x.reshape(n, self.hidden_dim, n_h * n_w)
+
+        # (n, hidden_dim, (n_h * n_w)) -> ((n_h * n_w), n, hidden_dim)
+        # The self attention layer expects inputs in the format (S, N, E)
+        # where S is the source sequence length, N is the batch size, E is the
+        # embedding dimension
+        x = x.permute(2, 0, 1)
```
fmassa, line 214: Did you try using the `batch_first=True` option in `MHA`? This would save us from this transposition, which could be potentially a bit faster (but needs to be benchmarked).

Also, a few other things would need to change if we do `batch_first=True`, like the dimension where we query the `token` etc

torchvision/models/vision_transformer.py
```diff
@@ -0,0 +1,353 @@
...
+        self.ln_1 = norm_layer(hidden_dim)
+        self.self_attention = nn.MultiheadAttention(hidden_dim, num_heads, dropout=attention_dropout)
+        self.dropout = nn.Dropout(dropout)
+
+        # MLP block
+        self.ln_2 = norm_layer(hidden_dim)
+        self.mlp = MLPBlock(hidden_dim, mlp_dim, dropout)
+
+    def forward(self, input: Tensor):
+        torch._assert(input.dim() == 3, f"Expected (seq_length, batch_size, hidden_dim) got {input.shape}")
```
fmassa, line 68: The failing test is probably due to the feature extractor picking the output of `input.dim() == 3`.

One easy way to fix this is to add `operator.gt` in the exclude list in https://github.com/pytorch/vision/blob/eb48a1d81c879089df78b77be5891d079dd643b1/test/test_backbone_utils.py#L92

sallysyw: Yes, thanks for helping me spotting that!

sallysyw:
Looks like the gpu test of `vit_h_14` is falling because of `CUDA out of memory` error: 
https://app.circleci.com/pipelines/github/pytorch/vision/12532/workflows/85f58ea0-b60b-416f-b593-1d725dd93fb6/jobs/991105

However, `test/test_models.py::test_classification_model[cpu-vit_h_14]` works just fine (I tested it locally).

Should I somehow bump up the memory allocation for the gpu test? any suggestions?
@datumbox @fmassa

datumbox:
@sallysyw 

> However, test/test_models.py::test_classification_model[cpu-vit_h_14] works just fine (I tested it locally).

From what I see Windows CPU is still [failing](https://app.circleci.com/pipelines/github/pytorch/vision/12532/workflows/85f58ea0-b60b-416f-b593-1d725dd93fb6/jobs/991159) with `Fatal Python error: Aborted` nevertheless it might be a memory issue. I can see that the linux equivalent test is [passing](https://app.circleci.com/pipelines/github/pytorch/vision/12532/workflows/85f58ea0-b60b-416f-b593-1d725dd93fb6/jobs/991138). The available memory for the Linux instance is smaller than the one on Windows, but it could be because Windows needs more memory.

> Any suggestions?

You could try reducing the memory needs of the test. This can be done by adding a record with the name of the model `vit_h_14` here:
https://github.com/pytorch/vision/blob/f093d0821194ed55f80f4764cf56e744f9217533/test/test_models.py#L257-L258

The parameters that you could overwrite in this model are:
https://github.com/pytorch/vision/blob/f093d0821194ed55f80f4764cf56e744f9217533/test/test_models.py#L512-L513

Probably significantly reduce the size of input and set the number of classes to 5. If you do that, you will have to also update your expected files. 

If the above recommendation doesn't fix the situation, we will need to bump the size of the CI instance type, which is obviously not ideal. A quick way to confirm that this fixes the issue is by bumping the types here:
https://github.com/pytorch/vision/blob/49ec677c970256a377293beb5902d6a238a3b90e/.circleci/config.yml#L9-L20
https://github.com/pytorch/vision/blob/49ec677c970256a377293beb5902d6a238a3b90e/.circleci/config.yml#L677-L681

The available executor types on CircleCI can be found on [their documentation](https://circleci.com/docs/2.0/configuration-reference/).

The above is just a temporary mitigation to show-case that there is indeed a problem with the resources. We will need to find a better solution than universally bumping the instances types across all tests prior merging.

Let me know if I can help unblock this. It would be great if we could address the remaining issues/comments and merge this PR soon.

sallysyw:
Hm, So I reduced the `input_size` and `num_classes` of `vit_h_14` in testing, however I'm still seeing a python frame error on all the windows tests:
https://app.circleci.com/pipelines/github/pytorch/vision/12577/workflows/4d307d90-98ee-4faa-afd7-739b7c22bb38

The error message is vague, something like 
```
test/test_models.py::test_classification_model[cuda-vit_h_14] Fatal Python error: Aborted

Thread 0x0000036c (most recent call first):
<no Python frame>
```

Does this still look like memory issue? 

@datumbox

sallysyw:
Hm, So I reduced the `input_size` and `num_classes` of `vit_h_14` in testing, however I'm still seeing a python frame error on all the windows tests:
https://app.circleci.com/pipelines/github/pytorch/vision/12577/workflows/4d307d90-98ee-4faa-afd7-739b7c22bb38

The error message is vague, something like 
```
test/test_models.py::test_classification_model[cuda-vit_h_14] Fatal Python error: Aborted

Thread 0x0000036c (most recent call first):
<no Python frame>
```

Does this still look like memory issue? 

@datumbox

datumbox:
@sallysyw The error is thrown during the JIT  check, specifically when the model is serialized and saved on a memory buffer:
https://github.com/pytorch/vision/blob/f093d0821194ed55f80f4764cf56e744f9217533/test/test_models.py#L125

This could still be a memory issue but the error message doesn't help. As it's on windows only, I don't have the hardware to test it locally. I see that the linux memory issue disappeared after changing the size of the unit-tests. Have you tried increasing the CI instance types on windows to see if more memory helps?

sallysyw:
> Have you tried increasing the CI instance types on windows to see if more memory helps?

Well I'm trying to avoid bumping up memories, but I'll give it a try. I can set the windows-cpu executor to be `2xlarge` but I cannot do anything with the windows-gpu executor because `windows.gpu.nvidia.medium` is already the largest instance I can found.

datumbox:
@sallysyw I see you [tried](https://github.com/pytorch/vision/pull/4594/commits/8e71e4b7c473bf54d3d89ba04d6d1b3dfd8de0bb) the very large instances on Windows CPU. That should have done the trick if it was a memory issue.

I think this will require additional investigation. I would start by avoid serialising the model in memory and dump it to a temporary file to see if that raises the same issue:
https://github.com/pytorch/vision/blob/f093d0821194ed55f80f4764cf56e744f9217533/test/test_models.py#L124-L128

@fmassa @prabhat00155 any good ideas for getting a more informative error message on Windows?

sallysyw:
> Did you try using the batch_first=True option in MHA? This would save us from this transposition, which could be potentially a bit faster (but needs to be benchmarked).

Yes, I did some benchmarking and found that `batch_first=True` is actually a little bit faster, see this table 
| job id | model      | batchfirst | epochs | nodes | Total Training Time |
| ------ | ---------- | ---------- | ------ | ----- | ------------------- |
| 5718   | vit\_b\_32 | F          | 50     | 2     | 8:45:32             |
| 5719   | vit\_b\_32 | T          | 50     | 2     | 8:41:25             |
| 6010   | vit\_b\_16 | F          | 50     | 4     | 5:27:44             |
| 5721   | vit\_b\_16 | T          | 50     | 4     | 5:08:21             |

I will change the default of `batch_first=True` and make changes accrodingly :)

>  I would start with only the token mode, remove classifier altogether and then revisit this in the future.

Sure thing, will update the API in my next few commits.

@fmassa

sallysyw:
I trained vit_b_32 from scratch to verify `batch_first=True` works as expected, here is the result I got:
| Training job id | Model      | Augumentation Level | Epochs | Nodes | batch\_size per GPU | Global batch\_size | Original Acc@1 | ACC@1 | ACC@5  | 
| --------------- | ---------- | ------------------- | ------ | ----- | ------------------- | ------------------ | -------------- | ------------ | ----- | 
| 6080            | vit\_b\_16 | 2+                  | 300    | 8     | 64                  | 4096               | 77.91          | 77.45 | 92.717 |

datumbox:
@sallysyw Awesome work!

I think by completing the refactoring of the classifier and by reproducing the paper accuracy we are almost there for merging. We just need to address two things before that:
1. The failing tests. See my comment above on how to get started: https://github.com/pytorch/vision/pull/4594#issuecomment-970634680
2. We need to move this on prototype. This is super easy, let's do it once the tests are fixed because on prototype they don't run automatically.

fmassa:
Hi,

I would skip some of those tests for the huge (and maybe even large) models. They are too big, and I'm not even sure we will be providing pre-trained weights for them given their size and accuracies (they huge model is not better than the base model if trained only on ImageNet without further data augmentation tricks)

datumbox:
@fmassa Sounds good to me. 

@sallysyw remove the model builder and the expected file of `vit_h_14` and let's wrap it up. :)

After removing the above and seeing the tests pass, we should move this to the prototype area. The model builder methods need to be updated a bit to use the new API but that's a 10-minute job. [Here is a reference](https://github.com/pytorch/vision/blob/main/torchvision/prototype/models/mnasnet.py) of how to update the builders (check out how to handle models without weights in the example). If you want we can even upload the temporary weights you got above (the ones with `77.45` Acc@1) and iterate on them using the new recipe. Happy to catch up with you offline to discuss the details.

sallysyw:
> If you want we can even upload the temporary weights you got above (the ones with 77.45 Acc@1) and iterate on them using the new recipe. Happy to catch up with you offline to discuss the details.

This is just the results for pre-training. To upload the weights, I think we will also need to perform fine-tuning. I plan to upload weights in another PR when I got the fine-tuning results :)  For this one let me just leave the pre-trained weights empty.
@datumbox

datumbox:
@sallysyw Sounds good to me! Feel free to upload them when you are done.

All tests pass which is great! We can now safely port this on prototype area. Let me know if you need any additional pointers on how to use the new model builders. Other than that, we should be good to merge when you are happy with the weights etc. :)

torchvision/prototype/models/vision_transformer.py
```diff
@@ -0,0 +1,354 @@
...
+
+    model = VisionTransformer(image_size=image_size, **kwargs)
+
+    if weights:
+        model.load_state_dict(weights.get_state_dict(progress=progress))
+
+    return model
+
+
+def vit_b_16(weights: bool = False, progress: bool = True, **kwargs: Any) -> VisionTransformer:
```
datumbox, line 265:
```suggestion
def vit_b_16(weights: Optional[VisionTransformer_B_16Weights] = None, progress: bool = True, **kwargs: Any) -> VisionTransformer:
```

torchvision/prototype/models/vision_transformer.py
```diff
@@ -0,0 +1,354 @@
...
+        patch_size=16,
+        num_layers=12,
+        num_heads=12,
+        hidden_dim=768,
+        mlp_dim=3072,
+        **kwargs,
+    )
+
+
+def vit_b_32(weights: bool = False, progress: bool = True, **kwargs: Any) -> VisionTransformer:
```
datumbox, line 288:
```suggestion
def vit_b_32(weights: Optional[VisionTransformer_B_32Weights] = None, progress: bool = True, **kwargs: Any) -> VisionTransformer:
```

torchvision/prototype/models/vision_transformer.py
```diff
@@ -0,0 +1,354 @@
...
+        patch_size=32,
+        num_layers=12,
+        num_heads=12,
+        hidden_dim=768,
+        mlp_dim=3072,
+        **kwargs,
+    )
+
+
+def vit_l_16(weights: bool = False, progress: bool = True, **kwargs: Any) -> VisionTransformer:
```
datumbox, line 311:
```suggestion
def vit_l_16(weights: Optional[VisionTransformer_L_16Weights] = None, progress: bool = True, **kwargs: Any) -> VisionTransformer:
```

torchvision/prototype/models/vision_transformer.py
```diff
@@ -0,0 +1,354 @@
...
+def vit_b_16(weights: bool = False, progress: bool = True, **kwargs: Any) -> VisionTransformer:
+    """
+    Constructs a ViT_b_16 architecture from
+    `"An Image is Worth 16x16 Words: Transformers for Image Recognition at Scale" <https://arxiv.org/abs/2010.11929>`_.
+
+    Args:
+        weights (bool, optional): If True, returns a model pre-trained on ImageNet. Default: False.
+        progress (bool, optional): If True, displays a progress bar of the download to stderr. Default: True.
+    """
+    weights = VisionTransformer_B_16Weights.verify(weights)
```
datumbox, line 274: You need some additional boiler plate code similar to: 
https://github.com/pytorch/vision/blob/1deb2ec22796350801aea44a000e4669cbf2fe1c/torchvision/prototype/models/mnasnet.py#L91-L94

This is necessary in all the builders. Don't worry about the duplicate code. I'm addressing this on #4948. Once we merge yours, I'll clean it up.

torchvision/prototype/models/vision_transformer.py
sallysyw: Good catch!

test/test_backbone_utils.py
```diff
@@ -89,7 +89,17 @@ def _create_feature_extractor(self, *args, **kwargs):
...
+        exclude_nodes_filter = [
+            "getitem",
+            "floordiv",
+            "size",
+            "chunk",
+            "_assert",
+            "eq",
+            "dim",
+            "getattr",
+            "self_attention",
```
datumbox, line 101: I think we missed this line in earlier reviews. I was under the impression that things like `getattr`, `eq` and `dim` were handled already and don't need to be filtered explicitly. Another concern is the addition of `self_attention` here. I assume you add it because it returns a tuple but this probably hints we should handle it differently. 

@alexander-soare any thoughts on whether there is a better way to handle this?

@sallysyw Note that because the model is now moved out of the main area, the tests will no longer run for it on CI automatically. We will need to check any modifications manually.

torchvision/prototype/models/vision_transformer.py
```diff
@@ -0,0 +1,383 @@
...
+
+class VisionTransformer_L_32Weights(Weights):
+    # If a default model is added here the corresponding changes need to be done in vit_l_32
+    pass
+
+
+def _vision_transformer(weights: Optional[Weights], progress: bool, **kwargs: Any) -> VisionTransformer:
+    image_size = kwargs.get("image_size", 224)
+    if "image_size" in kwargs:
+        kwargs.pop("image_size")
```
datumbox, line 256: Optional nit: You can replace all the 3 above lines with:
```python
    image_size = kwargs.pop("image_size", 224)
```

torchvision/prototype/models/vision_transformer.py
```diff
@@ -0,0 +1,383 @@
...
+
+    model = VisionTransformer(image_size=image_size, **kwargs)
```
datumbox, line 258: Optional nit: Given the parameters on the constructor are value arguments and not named ones, you could consider passing their values explicitly instead via the `**kwargs`. This is what we usually do for other models, for example:
https://github.com/pytorch/vision/blob/9841a9076e711f73482cb84e106fe5ec40707023/torchvision/prototype/models/mobilenetv3.py#L22-L32

There are a few discussions about updating our signatures on the future (see https://github.com/pytorch/vision/pull/4937#discussion_r749549044) so this could future-proof the implementation.

datumbox (commented):
All look great to me. I left one last comment for something missed in earlier reviews and 2 optional nits. Hopefully this is the last round before merging,

test/test_backbone_utils.py
alexander-soare: @datumbox making sure we are on the same page: I [originally](https://github.com/pytorch/vision/blob/a7a5818027b95b97f4c04c53e1e10759925a4877/test/test_backbone_utils.py#L58-L70) did it this way because of the way that we **randomly** select any 10 nodes as the return nodes. It's only to facilitate the backward pass test, and doesn't reflect the real usage of the API. The plan was to add filtered node names as and when needed. And I guess it was a bit careless to assume that the same seed wouldn't maybe behave differently on another setup, thereby returning another type of invalid node that hasn't been already accounted for in `exclude_nodes_filter`.

I couldn't think of any way to know the output types of nodes in advance. It could be neater, but a bit cumbersome to explicitly specify a set of node names to choose for each model in turn.

But otherwise I'd just forge ahead with this setup. Addressing the specific lines you raised:

For `self_attention` and generally for any tuple of tensors we can do a type check around [here](https://github.com/pytorch/vision/blob/1d4e2aaebeef6cf9d65b8920c17708a4f9c2e42d/test/test_backbone_utils.py#L157). Maybe replace that line with something like
```python
out_agg = 0
for node_out in out.values():
    if isinstance(node_out, Sequence):
        out_agg += sum(o.mean() for o in node_out)
    # handle dict while we are at it?
    elif isinstance(node_out, Dict):
        out_agg += sum(o.mean() for o in node_out.values())
    else:
        # and the assumption here is that the only other alternative at this point could be a Tensor
        out_agg += node_out.mean()
out_agg.backward()
```

For everything else except `getattr` I was able to verify that they do indeed come through from the feature extraction module:

```python 
import torch


    def __init__(self):
        super().__init__()
        self.param = torch.nn.Parameter(torch.rand(3, 4))

    def forward(self, x):
        x[0]
        x // 2
        x.size()
        torch._assert(x == x, '')
        x.dim()
        self.param


model = MyModule()

train_nodes, eval_nodes = get_graph_node_names(model)
```
Output is:
```
['x', 'getitem', 'floordiv', 'size', 'eq', '_assert', 'dim', 'param']
```

So I don't know why `get_attr` should be there. That's actually the op associated with the FX node, not the name.

I still feel a little icky about the above though. Thoughts?

datumbox: @alexander-soare Thanks for the detailed comment. I think adopting your proposal makes sense here. We should try to end up as close to the expected output as you have on your comment.

@sallysyw We will probably need to restore the legacy model builders in `torchvision.models` momentarily (perhaps just importing the classes from prototype might work without causing circular dependencies) in order to be able to run the unit-test on the CI and confirm everything still works...

torchvision/prototype/models/vision_transformer.py
sallysyw: What do you mean by `passing their values explicitly instead via the **kwargs`? In the above example, `num_classes` is passed via `**kwargs`, right?

test/test_backbone_utils.py
sallysyw: The weird thing is that if I comment out `self_attention`, one of the nodes' output would be 
image

If I comment out `getattr`, this node would be selected:
image

alexander-soare: @sallysyw Hmm looks like assignment to a tuple like `n, c, h, w = x.shape` results in a gettattr node. So yes,, it should be added to the exclusion as well.

```python
import torch


    def forward(self, x):
        s, h = x.shape


model = MyModule()

train_nodes, eval_nodes = get_graph_node_names(model)
print(train_nodes)
```
output is: `['x', 'getattr', 'getitem', 'getitem_1']`.

Re your first example with the `None` in the tuple can you print the name of the node? Not sure I'll be able to get the same seed as you to check myself. In any case I suppose you can always add more logic to dodge it like:

```python
out_agg = 0
for node_out in out.values():
    if isinstance(node_out, Sequence):
        out_agg += sum(o.mean() for o in node_out if o is not None)
    # handle dict while we are at it?
    elif isinstance(node_out, Dict):
        out_agg += sum(o.mean() for o in node_out.values() if o is not None)
    else:
        # and the assumption here is that the only other alternative at this point could be a Tensor
        out_agg += node_out.mean()
out_agg.backward()
```

It will always be possible that we need to add yet more logic for another corner case, but short of explicitly naming nodes for each model, we're just going to play this game.

torchvision/prototype/models/vision_transformer.py
datumbox: Sorry, I should have been clearer. 

In the MobileNetV3 example above, I'm referring to the model configuration value parameters passed by the model builder. That is the `inverted_residual_setting` and `last_channel`. 

In your case, you could explicitly pass the parameters `patch_size`, `num_layers`, `num_heads`, `hidden_dim` and `mlp_dim`. All of them (including `image_size`) are not named and are always passed. The `kwargs` in our setup stores optional named parameters for the constructor which have default values.

I should note, this is really a nitpick and definitely not a blocking one.

sallysyw: Got it! Let me make changes accordingly :)

test/test_backbone_utils.py
sallysyw: > Re your first example with the None in the tuple can you print the name of the node?

```
train_return_nodes = ['encoder.layers.encoder_layer_16.add_1', 'encoder.layers.encoder_layer_18.mlp.act', 'encoder.layers.encoder_layer_0.mlp.dropout_1', 'encoder.layers.encoder_layer_11.ln', 'encoder.layers.encoder_layer_22.mlp.linear_1', 'encoder.layers.encoder_layer_21.mlp.act', 'encoder.layers.encoder_layer_17.mlp.dropout_1', 'encoder.layers.encoder_layer_13.self_attention', 'encoder.layers.encoder_layer_21.dropout', 'encoder.layers.encoder_layer_15.mlp.dropout']
eval_return_nodes = ['encoder.layers.encoder_layer_9.self_attention', 'encoder.layers.encoder_layer_22.mlp.linear', 'encoder.layers.encoder_layer_5.mlp.linear', 'encoder.layers.encoder_layer_12.self_attention', 'encoder.layers.encoder_layer_3.ln_1', 'encoder.layers.encoder_layer_10.mlp.dropout', 'encoder.layers.encoder_layer_23.mlp.linear_1', 'encoder.layers.encoder_layer_5.mlp.dropout_1', 'encoder.layers.encoder_layer_13.ln_1', 'encoder.layers.encoder_layer_3.mlp.act']
```

@alexander-soare

sallysyw:
whoops - trying to import `prototype` in torchvision.models causes this error on circleCI:
```
ModuleNotFoundError: `torchvision.prototype.datasets` depends on PyTorch's `torchdata` (https://github.com/pytorch/data). You can install it with `pip install git+https://github.com/pytorch/data.git`. Note that you cannot install it with `pip install torchdata`, since this is another package.
Tests failed for torchvision-0.12.0.dev20211124-py36_cpu.tar.bz2 - moving package to /opt/conda/conda-bld/broken
WARNING:conda_build.build:Tests failed for torchvision-0.12.0.dev20211124-py36_cpu.tar.bz2 - moving package to /opt/conda/conda-bld/broken
WARNING conda_build.build:tests_failed(2970): Tests failed for torchvision-0.12.0.dev20211124-py36_cpu.tar.bz2 - moving package to /opt/conda/conda-bld/broken
TESTS FAILED: torchvision-0.12.0.dev20211124-py36_cpu.tar.bz2
```

But I confirmed locally that after the changes the tests passed on linux cpu.

cc @datumbox

datumbox (commented):
@sallysyw I tested them locally and the tests fail:
```
[W ir_emitter.cpp:4340] Warning: Dict values consist of heterogeneous types, which means that the dict has been typed as containing Dict[str, Union[Tensor, Tuple[Tensor, Optional[Tensor]]]]. To use any of the values in this Dict, it will be necessary to add an `assert isinstance` statement before first use to trigger type refinement.
  File "<eval_with_key>.76", line 196
    eq_14 = dim_12 == 3;  dim_12 = None
    _assert_14 = torch._assert(eq_14, 'Expected (seq_length, batch_size, hidden_dim) got Proxy(getattr_14)');  eq_14 = None
    return {'encoder.dropout': encoder_dropout, 'encoder.layers.encoder_layer_5.ln': encoder_layers_encoder_layer_5_ln_1, 'encoder.layers.encoder_layer_6.ln': encoder_layers_encoder_layer_6_ln_1, 'encoder.layers.encoder_layer_7.add': add_15, 'encoder.layers.encoder_layer_7.add_1': add_16, 'encoder.layers.encoder_layer_8.ln_1': encoder_layers_encoder_layer_8_ln_2, 'encoder.layers.encoder_layer_8.mlp.linear_1': encoder_layers_encoder_layer_8_mlp_linear_2, 'encoder.layers.encoder_layer_10.self_attention': encoder_layers_encoder_layer_10_self_attention, 'encoder.layers.encoder_layer_10.add': add_21, 'encoder.layers.encoder_layer_10.mlp.dropout_1': encoder_layers_encoder_layer_10_mlp_dropout_2}
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ <--- HERE
 (function operator())

test/test_backbone_utils.py:187 (TestFxFeatureExtraction.test_jit_forward_backward[vit_b_16])
  File "./vision/test/test_backbone_utils.py", line 198, in test_jit_forward_backward
    sum(o.mean() for o in fgn_out.values()).backward()
  File "./vision/test/test_backbone_utils.py", line 198, in <genexpr>
    sum(o.mean() for o in fgn_out.values()).backward()
AttributeError: 'tuple' object has no attribute 'mean'
```

I think we moved the code to prototype to early (that's on me) before we confirm that all issues are fixed. Since the tests on prototypes are not executed on CI, we now risk not detecting issues with the implementation.

I propose undoing the move to prototype, running the tests on the CI and ensure everything works prior moving everything to the prototype again. I did this job on a separate no-merge PR at #4984 and as you can see the [tests fail](https://app.circleci.com/pipelines/github/pytorch/vision/12826/workflows/eb0ae338-2e22-44f5-ad0d-92dbc7716e62/jobs/1019329). You are welcome to merge my changes into your current PR to investigate.

sallysyw:
> @sallysyw I tested them locally and the tests fail:
> 
> ```
> [W ir_emitter.cpp:4340] Warning: Dict values consist of heterogeneous types, which means that the dict has been typed as containing Dict[str, Union[Tensor, Tuple[Tensor, Optional[Tensor]]]]. To use any of the values in this Dict, it will be necessary to add an `assert isinstance` statement before first use to trigger type refinement.
>   File "<eval_with_key>.76", line 196
>     eq_14 = dim_12 == 3;  dim_12 = None
>     _assert_14 = torch._assert(eq_14, 'Expected (seq_length, batch_size, hidden_dim) got Proxy(getattr_14)');  eq_14 = None
>     return {'encoder.dropout': encoder_dropout, 'encoder.layers.encoder_layer_5.ln': encoder_layers_encoder_layer_5_ln_1, 'encoder.layers.encoder_layer_6.ln': encoder_layers_encoder_layer_6_ln_1, 'encoder.layers.encoder_layer_7.add': add_15, 'encoder.layers.encoder_layer_7.add_1': add_16, 'encoder.layers.encoder_layer_8.ln_1': encoder_layers_encoder_layer_8_ln_2, 'encoder.layers.encoder_layer_8.mlp.linear_1': encoder_layers_encoder_layer_8_mlp_linear_2, 'encoder.layers.encoder_layer_10.self_attention': encoder_layers_encoder_layer_10_self_attention, 'encoder.layers.encoder_layer_10.add': add_21, 'encoder.layers.encoder_layer_10.mlp.dropout_1': encoder_layers_encoder_layer_10_mlp_dropout_2}
>            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ <--- HERE
>  (function operator())
> 
> test/test_backbone_utils.py:187 (TestFxFeatureExtraction.test_jit_forward_backward[vit_b_16])
> Traceback (most recent call last):
>   File "./vision/test/test_backbone_utils.py", line 198, in test_jit_forward_backward
>     sum(o.mean() for o in fgn_out.values()).backward()
>   File "./vision/test/test_backbone_utils.py", line 198, in <genexpr>
>     sum(o.mean() for o in fgn_out.values()).backward()
> AttributeError: 'tuple' object has no attribute 'mean'
> ```
> 
> I think we moved the code to prototype to early (that's on me) before we confirm that all issues are fixed. Since the tests on prototypes are not executed on CI, we now risk not detecting issues with the implementation.
> 
> I propose undoing the move to prototype, running the tests on the CI and ensure everything works prior moving everything to the prototype again. I did this job on a separate no-merge PR at #4984 and as you can see the [tests fail](https://app.circleci.com/pipelines/github/pytorch/vision/12826/workflows/eb0ae338-2e22-44f5-ad0d-92dbc7716e62/jobs/1019329). You are welcome to merge my changes into your current PR to investigate.

hmm... I think the seed generated on your machine is different from mine and that's why I didn't catch this failure previously.
Can you please print out the nodes in your test run and I'll investigate (or filter) it?
@datumbox

datumbox (approved):
@sallysyw All worked like a charm. I pushed the removal of the extra classed from your branch and merged main. We should be good to merge whenever you want. Thanks for the great contribution, looking forward to the weights.

alexander-soare:
@datumbox I think I remember you asked for my feedback at some point here but can't see it anymore? Yes, the things I commented on look like they've been addressed, thank you. @sallysyw nice work!

PR #4598, opened 2021-10-12
Migrate mnist dataset from np.frombuffer

### LINKED ISSUE #4552: Use `torch.frombuffer` instead of `np.frombuffer`
PyTorch recently introduced a `torch.frombuffer` method https://github.com/pytorch/pytorch/pull/59077 
We should use it to simplify our code whenever possible, instead of relying on alternatives like `torch.from_numpy(np.frombuffer))`.

lc0:
Accordingly to pytorch/pytorch#59077 this PR migrates from np.frombuffer to pytorch.frombuffer and closes #4552

This PR is closing the last piece with mnist dataset. 

So far I tested all mnist tests:

image

cc @pmeier @datumbox

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -4,6 +4,23 @@ files = torchvision
 show_error_codes = True
 pretty = True
 allow_redefinition = True
+warn_redundant_casts = True
+
+[mypy-torchvision.prototype.datasets.*]
+
+; untyped definitions and calls
+disallow_untyped_defs = True
+
+; None and Optional handling
+no_implicit_optional = True
+
+; warnings
+warn_unused_ignores = True
+warn_return_any = True
+warn_unreachable = True
+
+; miscellaneous strictness flags
+allow_redefinition = True
 
 [mypy-torchvision.io._video_opt.*]
 
--- a/test/test_models.py
+++ b/test/test_models.py
@@ -223,7 +223,7 @@ autocast_flaky_numerics = (
 # The tests for the following quantized models are flaky possibly due to inconsistent
 # rounding errors in different platforms. For this reason the input/output consistency
 # tests under test_quantized_classification_model will be skipped for the following models.
-quantized_flaky_models = ("inception_v3",)
+quantized_flaky_models = ("inception_v3", "resnet50")
 
 
 # The following contains configuration parameters for all models which are used by
--- a/torchvision/datasets/usps.py
+++ b/torchvision/datasets/usps.py
@@ -1,5 +1,5 @@
 import os
-from typing import Any, Callable, cast, Optional, Tuple
+from typing import Any, Callable, Optional, Tuple
 
 import numpy as np
 from PIL import Image
@@ -63,7 +63,7 @@ class USPS(VisionDataset):
             raw_data = [line.decode().split() for line in fp.readlines()]
             tmp_list = [[x.split(":")[-1] for x in data[1:]] for data in raw_data]
             imgs = np.asarray(tmp_list, dtype=np.float32).reshape((-1, 16, 16))
-            imgs = ((cast(np.ndarray, imgs) + 1) / 2 * 255).astype(dtype=np.uint8)
+            imgs = ((imgs + 1) / 2 * 255).astype(dtype=np.uint8)
             targets = [int(d[0]) - 1 for d in raw_data]
 
         self.data = imgs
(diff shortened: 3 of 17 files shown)


lc0:
> Thanks for the PR @lc0!
> 
> This looks like a good improvement to me but I'll let @pmeier to make the last call as he knows the datasets domain much better than me. Note that the [linter](https://app.circleci.com/pipelines/github/pytorch/vision/11236/workflows/eb82b9ed-11bd-4dc5-894b-ca62f945ac12/jobs/867866) is currently failing, could you please fix?

good point, somehow on the first format run - I've got a lot of changes. Seems like after rebasing it's all good now. So sorry for the inconvenience 🙈

pmeier:
Could you test this against the original data? IIRC, there was something about the byte order, which is why I implemented the same in the new prototype like this:

https://github.com/pytorch/vision/blob/c790216ad47fb379b2a158fe8e3602865b26e09f/torchvision/prototype/datasets/_builtin/mnist.py#L79

If this works with `torch.frombuffer` I'm all in favor of removing the ` numpy` dependency. I'm on PTO right now, so it will take me some time to test this properly.

datumbox:
@pmeier We are happy to wait until you are back from PTO to allow for more thorough tests. Would you mind marking this with "Request changes" so that it's not accidentally merged?

pmeier:
It seems to work just fine, but if we use it, we are getting an "old" warning back:

```
UserWarning: The given buffer is not writable, and PyTorch does not support non-writable tensors. This means you can write to the underlying (supposedly non-writable) buffer using the tensor. You may want to copy the buffer to protect its data or make it writable before converting it to a tensor. This type of warning will be suppressed for the rest of this program. (Triggered internally at  ../torch/csrc/autograd/python_torch_functions_manual.cpp:489.)
  parsed = torch.frombuffer(data, dtype=torch_type, offset=(4 * (nd + 1)))
```

This was only removed only recently in #4184. Maybe we can send a patch upstream to fix it, but the warning seems justified to me.

NicolasHug:
According to the docs of `torch.frombuffer`:

> The returned tensor and buffer share the same memory. Modifications to
the tensor will be reflected in the buffer and vice versa. The returned
tensor is not resizable

so I'm not sure we can avoid the warning if we switch to `torch.frombuffer`, unless we copy the buffer beforehand.

I would suggest to either:

- if we really want to use `torch.frombuffer`, just silence the warning by wrapping the call in a context manager. For this specific buffer, I really doubt the "writeable tensor on a read-only buffer" will ever cause a problem (FWIW, we used to ignore the warning before we started making a copy in  #4184)
- not care about all this and keep the code as-is (i.e. not merge this PR). The code-quality improvement is pretty marginal anyway.

datumbox:
@pmeier Apologies, I misread the previous comment. TBH I would prefer to merge the PR because it removes the extra code of juggling between numpy and pytorch. Concerning fixing the warning, we could either call `clone()` or suppress it. Copying doesn't sound too bad given the dataset is tiny and can avoid issues with in-place modifications. It will also align the behaviour with #4578 but no strong opinions.

pmeier:
> Concerning fixing the warning, we could either call `clone()` or suppress it.

Sorry to be pedantic here, but we need to suppress **and** clone. `torch.frombuffer` emits the warning unconditionally. After we loaded the buffer we can then clone to avoid problems with inplace modifications. For example:

```python
with suppress_read_only_warning():
    data = torch.from_buffer(buffer, dtype=dtype).clone()
```

This issues did not arise before, because `numpy` can either deal with read-only buffers or simply doesn't warn you about it.

datumbox:
I might be missing something here but why can't we copy the buffer and make it mutable?

Something like the following should work because the `bytearray`:
```python
parsed = torch.frombuffer(bytearray(data), dtype=torch_type, offset=(4 * (nd + 1)))
```

We do copy but we don't have to suppress warnings etc.

pmeier:
I like `torch.from_buffer(bytearray(buffer), dtype=dtype)` best, because it is quite concise and we avoid to write extra functionality to suppress a warning that we can't control directly. Thanks @datumbox for the suggestion.

lc0:
Hi @pmeier it's a good point. It feels that we are maintaining two implementations of dataset API. Does it make sense somehow abstract it away and do not implement/maintain it twice? :) 

Also, if this is broken currently, I guess it's subtle enough to be useful enough to cover with a test. What do you think?

Issue #4582, opened 2021-10-10
Add typing annotations to models/detection


oke-aditya:
### 🚀 The feature

This is an issue to track down all the files in detection module and work up the chain.

- [ ] _utils.py
- [ ] anchor_utils.py 
- [ ] backbone_utils.py
- [ ] image_list.py 
- [ ] transform.py
- [ ] rpn.py 
- [ ] roi_heads.py 
- [ ] generalized_rcnn.py 
- [ ] faster_rcnn.py
- [ ] mask_rcnn.py
- [ ] keypoint_rcnn.py  
- [ ] retinanet.py
- [ ] ssd.py
- [ ] ssdlite.py

### Motivation, pitch

@khushi-411 and I will sync and work on these files.

Solving This issue also helps #2025 

As first step I'm closing #4220 and splitting it up into multiple PRs preferably one file each.

Hopefully we should complete this soon.

### Alternatives

Note that there are cases where we have make JIT happy over mypy.

Most of these cases occur as JIT does not Support Tuple fully. In most cases we should prefer List.

### Additional context

cc @datumbox @pmeier @NicolasHug

oke-aditya:
Go ahead :smile: Just remember that after #4583 is merged you would need to remove that file from `mypy.ini` configuration.

PR #4599, opened 2021-10-12
Add typing to anchor utils

oke-aditya:
Helps #4582


oke-aditya:
Hmm so Tuple[Tuple[int, ...]] worked fine. I guess this should be good :) 

cc @datumbox I think Phillip is on PTO.

PR #4603, opened 2021-10-12
Add typing annotations to detection/backbone_utils

oke-aditya:
Helps #4582

A headstart for @khushi-411 :smiley: 

Initially I was bit hesitant to start, but soon felt that this file isn't very simple for new contributor.

I have just added a couple of types. I felt this would help you.

cc @datumbox


datumbox:
FYI, there is another file in prototypes that should be types along with this:
https://github.com/pytorch/vision/blob/main/torchvision/prototype/models/detection/backbone_utils.py#L5-L12

PR #4606, opened 2021-10-12
Refactor unnecessary `else` / `elif` when `if` block has a `return` statement

Borda:
Co-authored-by: deepsource-autofix[bot] <62050782+deepsource-autofix[bot]@users.noreply.github.com>

### DIFF
--- a/.circleci/regenerate.py
+++ b/.circleci/regenerate.py
@@ -134,10 +134,10 @@ manylinux_images = {
 def get_manylinux_image(cu_version):
     if cu_version == "cpu":
         return "pytorch/manylinux-cuda102"
-    elif cu_version.startswith("cu"):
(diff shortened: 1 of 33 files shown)


NicolasHug:
If you're referring to the changes you made in https://github.com/pytorch/vision/pull/4585 (i.e. https://github.com/pytorch/vision/pull/4585/commits/d9069101fa1ddae532f7cb251403d5c0f6964c91) , it's just a few lines, so while they're unrelated with the original goal of the PR, it's probably not worth debating over

prabhat00155:
Well `elif` after `return` doesn't make sense to me and depending on what linter you use, it may throw a warning as well(example `pylint`). Anyways, as most people are in favour of leaving it as it is, I am not going to push for the change. Thanks!

PR #4612, opened 2021-10-13
Add typing annotations to detection/roi_heads

oke-aditya:
Helps #4582

Current status:
~~Opened PR to check if CI passes after adding types.~~
~~Fixing mypy errors~~


oke-aditya:
We can turn off the places where it is messier :) Can you point out such places, since it is slightly subjective where they are messier.

My 2 cents

Probably there will be very few cases < 10 and this file is probably largest ~800 lines.
So I think instead of ignoring the file, it will be better to just ignore the issues at those places.

PR #4614, opened 2021-10-14
fix check integrity check with download==false

edgarriba:
hi ! I had to make this work around while using WiderFace dataset from downloaded directories.

Besides, in my computer seems that downloading doesn't work as expected.


NicolasHug:
CCing @pmeier on this one

PR #4618, opened 2021-10-14
Update PR labeling requirements

NicolasHug:
I just took care of categorizing all the PRs of the upcoming release. Overall the current labeling in place is tremendously useful (thanks @pmeier for the bot!!).

Our [release notes](https://github.com/pytorch/vision/releases) consists of mainly these sections:

- Backward-incompatible changes
- Deprecations
- New features
- Improvements
- Bug fixes
- Code quality

Since each of these has a 1-1 mapping to one of our tags, **I'd like to make such tags mandatory for labeling a PR.** Often, these tags are missing, and this forces the one who's categorizing the release notes to manually check every single one of these. If you take a look at the last cell (11) of [this notebook](http://localhost:8888/notebooks/release_notes_new.ipynb) that I used, there were about 150 such PRs for this release. That's a lot!! It would save a lot of time if we would have these tags properly set right from the beginning.

Since it's not always incredibly obvious what a PR should be labeled with, we can have a new "other" tag to use as an escape and which would be manually treated, but hopefully we wouldn't use this one too much.

cc @seemethere


NicolasHug:
Thanks for the review!

I think we'll still need to be a bit mindful because if we completely forget to tag something and then tag it with e.g. just "modules: dataset", the bot won't be activated again to tell us that we're still missing a primary label.

But hopefully these cases should be rare. @fmassa @datumbox @prabhat00155 @pmeier let's just try to remember to always include one of these labels when merging a PR:

    "new feature",
    "bug",
    "code quality",
    "enhancement",
    "bc-breaking",
    "deprecation",
    "other"


Thanks!

PR #4619, opened 2021-10-14
Add typing annotations to detection/rpn

oke-aditya:
Helps #4582

Sorry for the flood of PRs, I luckily have couple of days holiday :sweat_smile: 

cc @datumbox


datumbox:
Thanks for the PR @oke-aditya. All changes look good to me. We just need to take care of the issue you marked here https://github.com/pytorch/vision/pull/4619#discussion_r729241121 and we should be good to go :)

PR #4629, opened 2021-10-15
Replace all `no_grad()` instances with `inference_mode()` in reference scripts

### LINKED ISSUE #4624: Replace instances of `torch.no_grad()` with `torch.inference_mode()` in evaluation reference scripts
Since last release of PyTorch, it now provides a new context manager / function decorator that enables faster execution than `torch.no_grad()` https://pytorch.org/docs/stable/generated/torch.inference_mode.html?highlight=inference_mode#torch.inference_mode

We should replace instances of `torch.no_grad()` in our reference scripts whenever `torch.inference_mode()` is the correct candidate (for example, in evaluation code as in https://github.com/pytorch/vision/blob/3d7244b5280301792a6959c59154d4809ad1209a/references/classification/train.py#L60

cc @datumbox

XoMute:
Closes #4624 
All instances of `no_grad()` seemed ok to replace with `inference_mode()` since all of them were related to evaluation

cc @datumbox

### DIFF
--- a/references/classification/train.py
+++ b/references/classification/train.py
@@ -57,7 +57,7 @@ def evaluate(model, criterion, data_loader, device, print_freq=100, log_suffix="
     header = f"Test: {log_suffix}"
 
     num_processed_samples = 0
-    with torch.no_grad():
+    with torch.inference_mode():
         for image, target in metric_logger.log_every(data_loader, print_freq, header):
             image = image.to(device, non_blocking=True)
             target = target.to(device, non_blocking=True)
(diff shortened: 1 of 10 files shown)


XoMute:
It seems like some CI tests failed due to network issues.
Can someone please rerun the workflow to make sure that everything will pass?

oke-aditya:
I wanted to test the performance difference between these two. No grad vs inference mode. 

I am not planning to open PR and revert this :)

PR #4631, opened 2021-10-15
Add typing annotations to detection/generalized_rcnn 

oke-aditya:
Helps #4582

This is easier to tackle. I think this should be good to go with less friction.


oke-aditya:
A Friendly ping to @NicolasHug and @datumbox  for a review :)

torchvision/models/detection/generalized_rcnn.py
```diff
@@ -97,7 +104,7 @@ def forward(self, images, targets=None):
             features = OrderedDict([("0", features)])
         proposals, proposal_losses = self.rpn(images, features, targets)
         detections, detector_losses = self.roi_heads(features, proposals, images.image_sizes, targets)
-        detections = self.transform.postprocess(detections, images.image_sizes, original_image_sizes)
+        detections = self.transform.postprocess(detections, images.image_sizes, original_image_sizes)  # type: ignore[operator]
```
datumbox, line 107: Out of curiosity, what's up with that warning?

datumbox (approved):
LGTM, thanks!

torchvision/models/detection/generalized_rcnn.py
oke-aditya: Here is the error.

```
You should consider upgrading via the '/usr/local/bin/python -m pip install --upgrade pip' command.
torchvision/models/detection/generalized_rcnn.py:106: error: "Tensor" not
callable  [operator]
            detections = self.transform.postprocess(detections, images.ima...
                         ^
Found 1 error in 1 file (checked 143 source files)

Exited with code exit status 1
```
We pass transform as `nn.Module`, the transform `GeneralizedRCNNTransform` has a method called `postprocess` but `GeneralizedRCNN` does not know that. So it things it's a Tensor and not a function, hence the error.

I don't know easy way to fix this, maybe we should type `transform` as `GeneralizedRCNNTransform` ?

torchvision/models/detection/generalized_rcnn.py
```diff
@@ -36,19 +36,26 @@ def __init__(self, backbone, rpn, roi_heads, transform):
...
 
         return detections
 
-    def forward(self, images, targets=None):
-        # type: (List[Tensor], Optional[List[Dict[str, Tensor]]]) -> Tuple[Dict[str, Tensor], List[Dict[str, Tensor]]]
+    def forward(
+        self,
+        images: List[Tensor],
+        targets: Optional[List[Dict[str, Tensor]]] = None,
+    ) -> Union[Tuple[Dict[str, Tensor], List[Dict[str, Tensor]]], Dict[str, Tensor], List[Dict[str, Tensor]]]:
```
prabhat00155, line 54: @datumbox This change causes failure in internal tests, check D32216673.

oke-aditya: FYI
This change was done as Union is supported by Torchscript. There isn't a CI failure on main branch. (Probably you are referring to fbcode)

prabhat00155: Yeah.
```
RuntimeError: 
Union[Tuple[Dict[str, Tensor], List[Dict[str, Tensor]]], List[Dict[str, Tensor]], Dict[str, Tensor]] cannot be used as a tuple:
  File "/data/sandcastle/boxes/eden-trunk-hg-fbcode-fbsource/fbcode/buck-out/dev/gen/smart/smart_catalog_filter/inference/handler_test#binary,link-tree/smart_catalog_filters/embedding.py", line 144
    def forward(self, images: List[Tensor]):
        _, results = self.rcnn(images)
                     ~~~~~~~~~~~~~~~~ <--- HERE
        return [
            (self.get_emb(result["features"]), result["boxes"]) for result in results


----------------------------------------------------------------------
```

datumbox: @prabhat00155 Thanks for flagging, I see it. This is something we should keep in mind for future Annotations and PRs. 

Union of a tuple with a non-tuple is likely to break code because of expressions like:
`a, b = result` 

My proposal is to amend the diff on FBcode to either remove the annotation or remove the union (whichever passes all tests) and then sync the change that you will do on FBcode to GH on a cherrypick OR on a new PR (Which ever easier). 

Let me know your thoughts on the above and if I can do something to help.

prabhat00155: Thanks, I'll amend the diff.

PR #4636, opened 2021-10-18
Add typing in detection module (faster rcnn).

khushi-411:
Follows https://github.com/pytorch/vision/issues/4582.


oke-aditya:
Eventhough the CI looks green, let's not hurry :smiley: Since inherited classes can often mismatch later.

PR #4639, opened 2021-10-18
check if installed torch with cuda

mszhanyi:
similar to https://github.com/pytorch/audio/pull/1867


mszhanyi:
@NicolasHug All tests passed.

mszhanyi:
It'd better to merge my PR soon, else we would see some wizard errors tomorrow again.

PR #4650, opened 2021-10-19
improve COCO prototype

pmeier:
This adds a few improvements:
1. Add the ability to select which annotations to load, e.g. `instances` (default) or `captions`. `None` is a special value that will only load the images which is much faster if the annotations are not needed. This is for example useful for unsupervised training.
2. Provide categories for each bounding box.
3. Add tests.
4. Add support for segmentation masks. Now you can do something like this:

    ```python
    from torchvision.utils import draw_segmentation_masks
    
    for sample in datasets.load("coco"):
        draw_segmentation_masks(sample["image"], sample["segmentations"])
        break
    ```
    foo

cc @pmeier @bjuncek


pmeier:
The new implementation is working, but a major perf regression. Cold start now takes ~20 minutes on my system. With minmal fake data warm up is minimal, so I'm guessing one or more things I used in the implementation does not scale to more inputs.

torchvision/prototype/datasets/_builtin/coco.py
```diff
@@ -133,25 +177,121 @@ def _make_datapipe(
...
+                year=config.year,
+                ann_type_idcs=dict(zip(ann_types, range(len(ann_types)))),
+            ),
+            drop_none=True,
             buffer_size=INFINITE_BUFFER_SIZE,
         )
+        partial_anns_dps = [self._make_partial_anns_dp(dp) for dp in partial_anns_dps]
+
+        anns_dp = Concater(*partial_anns_dps)
+        anns_dp = Grouper(anns_dp, group_key_fn=getitem(1, "id"), buffer_size=INFINITE_BUFFER_SIZE)
```
ejguan, line 220: We should use `IterKeyZipper` here as we know the size of each group. If each datapipe in `partial_anns_dps` has a similar order, the `Grouper` would have to fill out most of buffer before yields any data.

torchvision/prototype/datasets/_builtin/coco.py
```diff
@@ -133,25 +177,121 @@ def _make_datapipe(
         anns_meta_dp = Mapper(anns_meta_dp, getitem(1))
         anns_meta_dp = UnBatcher(anns_meta_dp)
 
-        anns_dp = Grouper(anns_meta_dp, group_key_fn=getitem("image_id"), buffer_size=INFINITE_BUFFER_SIZE)
-        # drop images without annotations
-        anns_dp = Filter(anns_dp, bool)
-        anns_dp = Shuffler(anns_dp, buffer_size=INFINITE_BUFFER_SIZE)
-        anns_dp = IterKeyZipper(
-            anns_dp,
+        partial_anns_dp = Grouper(anns_meta_dp, group_key_fn=getitem("image_id"), buffer_size=INFINITE_BUFFER_SIZE)
```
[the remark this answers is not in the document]
pmeier: Nope. For every single type it can 0 or any positive integer (usually < 100, but I haven't checked that)

torchvision/prototype/datasets/_builtin/coco.py
```diff
@@ -133,25 +177,121 @@ def _make_datapipe(
...
             ref_key_fn=getitem("id"),
             buffer_size=INFINITE_BUFFER_SIZE,
         )
 
-        images_dp = ZipArchiveReader(images_dp)
+    def _precollate_anns(
+        self, data: List[Tuple[List[Dict[str, Any]], Dict[str, Any]]], *, types: List[str]
+    ) -> Tuple[Dict[str, Any], Dict[str, Dict[str, Any]]]:
+        ann_data, (image_meta, *_) = zip(*data)
+        anns = dict(zip(types, ann_data))
```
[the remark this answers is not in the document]
pmeier: In `_make_anns_dp` we use `ann_types` to split the archive with the `Demultiplexer`.

torchvision/prototype/datasets/_builtin/coco.py
pmeier: > We should use `IterKeyZipper` here as we know the size of each group. 

We don't. The grouper groups on the image id so basically this groups all annotations of one image together. So we neither know which annotations (instances, captions, ...) are present at all nor how many of them.

> If each datapipe in `partial_anns_dps` has a similar order, the `Grouper` would have to fill out most of buffer before yields any data.

Since we ride with an infinite buffer size, won't it fill up the entire buffer regardless of the order? In any case, I'm not sure why this is such bottleneck. I mean we are talking about roughly 500 MB of annotations and the key function is only item access. Surely that should be faster than 20 minutes, right? Reading the whole file into memory with `json.load` takes less than a second.

torchvision/prototype/datasets/_builtin/coco.py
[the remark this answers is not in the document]
pmeier: You were right, it didn't work the way I expected it to work. I've added a new custom datapipe in the latest commit that handles this and replaces the `Grouper` you think is the bottleneck.

torchvision/prototype/datasets/_builtin/coco.py
pmeier: It did some more benchmarking and it turns out that `Grouper` is indeed the bottleneck, but not at the place where you suspected it: the bottleneck happens when grouping all annotations of one type together.

I've benchmarked with [this script](https://gist.github.com/pmeier/6bf44e22f29f0bc840f8b9ba6a7f3830), which does the following:

1. Load one annotations json, i.e. one annotation type like instances or captions, from a zip archive.
2. Group all annotations of the same image id together.
3. Match the grouped annotations with the corresponding meta information of the image.

These steps are performed once without datapipes and once with. This is the result:

```
without datapipes 9.711 seconds
with datapipes 520.147 seconds
```

The `Grouper` seems to be hella inefficient.

ejguan: > We don't. The grouper groups on the image id so basically this groups all annotations of one image together. So we neither know which annotations (instances, captions, ...) are present at all nor how many of them.

IIUC, we have fixed number of `partial_anns_dps`. In each `partial_anns_dp`, the annotations have been aggregated into a list by the `Grouper` per image. Then, we know there are a certain amount of types (same as the number of `partial_anns_dp`). 

I will run the script and try to figure out if there is any work around.

ejguan: Thank to your script, I found the perf regression is caused by `Grouper` here:
https://github.com/pytorch/pytorch/blob/4eb772fde619c52f5877b917419beb8220160588/torch/utils/data/datapipes/iter/grouping.py#L318-L321

We are yielding group of contents with the order from bigger group to smaller group. Then, each iteration would take O(N) to find the biggest group. In total, it would be O(N^2) for whole epoch.

I will open an issue for it.

pmeier: I'm not sure what the initial motivation for returning ordered groups was in the first place, but that seems like a hidden feature. The docstring doesn't mention it and I wouldn't expect it either. IMO, this should be completely removed from the `Grouper` and if someone needs this sorting by group size, it can be easily added by a successive datapipe.

pmeier: Running with this patch

```diff
diff --git a/torch/utils/data/datapipes/iter/grouping.py b/torch/utils/data/datapipes/iter/grouping.py
index 37dcc6238b..ffc038232e 100644
--- a/torch/utils/data/datapipes/iter/grouping.py
+++ b/torch/utils/data/datapipes/iter/grouping.py
@@ -315,7 +315,5 @@ class GrouperIterDataPipe(IterDataPipe[DataChunk]):
                 if result_to_yield is not None:
                     yield self.wrapper_class(result_to_yield)
 
-        while buffer_size:
-            (result_to_yield, buffer_size) = self._remove_biggest_key(buffer_elements, buffer_size)
-            if result_to_yield is not None:
-                yield self.wrapper_class(result_to_yield)
+        for key in tuple(buffer_elements.keys()):
+            yield self.wrapper_class(buffer_elements.pop(key))
```

now gives

```
without datapipes 9.781 seconds
with datapipes 10.549 seconds
```

which sounds reasonable.

torchvision/prototype/datasets/_builtin/coco.py
```diff
@@ -26,24 +28,144 @@
...
+            yield ref_image_meta, anns
 
 
 class Coco(Dataset):
+    def _decode_instances_anns(self, anns: List[Dict[str, Any]], image_meta: Dict[str, Any]) -> Dict[str, Any]:
+        labels = [ann["category_id"] for ann in anns]
+        categories = [self.info.categories[label] for label in labels]
+        return dict(
+            areas=torch.tensor([ann["area"] for ann in anns]),
+            iscrowd=torch.tensor([ann["iscrowd"] for ann in anns], dtype=torch.bool),
```
fmassa, line 118: FYI COCO also contains segmentation masks for each instance (and sometimes even keypoints depending on the instantiation).

We will need to decide at that point what will be the format representation for the segmentation masks (they are stored as RLE polygons sometimes).

torchvision/prototype/datasets/_builtin/coco.py
```diff
@@ -26,24 +28,144 @@
...
+            bounding_boxes=BoundingBox(
+                [ann["bbox"] for ann in anns], format="xywh", image_size=(image_meta["height"], image_meta["width"])
+            )
+            if anns
+            else [],
```
fmassa, line 123: For non-keypoints datasets, do we have by default images without annotations in COCO? I thought we didn't.

torchvision/prototype/datasets/_builtin/coco.py
```diff
@@ -26,24 +28,144 @@
...
+    def _parse_meta_data(
+        cls,
+        category: Optional[str] = DEFAULT,  # type: ignore[assignment]
+        super_category: Optional[str] = DEFAULT,  # type: ignore[assignment]
+    ) -> Dict[str, Tuple[Any, Any]]:
+        return dict(category=(category, None), super_category=(super_category, None))
+
+
+# TODO: this is in desperate need of comments. They will be added as soon as we are sure this is the way to go
+class CocoAnnotationsCollater(IterDataPipe[Tuple[Dict[str, Any], Dict[str, Dict[str, Any]]]]):
```
fmassa, line 54: Is this re-implementing what `pycocotools` do internally, to avoid a dependency?

torchvision/prototype/datasets/_builtin/coco.py
ejguan: Yeah. This is used for the case that buffer is smaller than the dataset, we would yield the biggest group. Even though it definitely can be optimized to O(log(n)) IMO.
Anyway, I will send a patch today.

pmeier:
I've added COCO to the benchmarks. Running 

```bash
python -m torchvision.prototype.datasets.benchmark -n1 --no-start coco
```

with the patch described in https://github.com/pytorch/vision/pull/4650#discussion_r752022496 gives:

```
coco (instances)
legacy iteration 185.052 it/s
new iteration 152.523 it/s
coco (captions)
legacy iteration 240.053 it/s
new iteration 198.828 it/s

```

torchvision/prototype/datasets/_builtin/coco.py
pmeier: Yes, I'm aware of that, and I'll support for that in a follow-up PR starting with segmentations. This improvement makes it really easy to load additional annotations. For example `datasets.load("coco", instances=True, captions=True)` will load both and `keypoints` will be added to that list.

torchvision/prototype/datasets/_builtin/coco.py
pmeier: Kind of, but not to avoid a dependency. I'm not sure if it works well with datapipes. Will have a closer look. Unless we want to avoid the dependency?

torchvision/prototype/datasets/_builtin/coco.py
pmeier: This is just the default value if a user deselects the instances, e.g. `datasets.load("coco", instances=False, captions=True)`. Or would you prefer to not have a `"bounding_box` entry in the sample in that case?

torchvision/prototype/datasets/_builtin/coco.py
pmeier: We maybe can use it, but it depends how far we want to go into bad practices. The `COCO` object loads the file with `open` which fails for file handles coming from a zip file. Patching this would be fairly easy using something like `unittest.mock`, but that is kind of fragile, because we can't treat `pycocotools` as black box anymore. That being said, I'm fairly certain that newer versions of `pycocotools` won't break our patch, because it is already really stable and this is something which is most likely not going to be touched in newer versions.

fmassa: There are basically 3 things that `pycocotools` do for us which are helpful:
- load the data (what you are doing here)
- convert polygons to segmentation maps (with C++ extensions for that)
- evaluation code

I would like the evaluation code to also be part of torchvision sometime in the future, as it would enable us to avoid having to hold the COCO json in memory (and we would be able to perform progressive evaluation in a more streamlined fashion, in a better way than what we do in the `references/` for now).

This would leave us with just the polygons conversion, which is a very handy tool but we could implement it in torchvision as well.

So ultimately I think that we might want to drop a dependency on `pycocotools`, but it will require further work

torchvision/prototype/datasets/_builtin/coco.py
fmassa: Fine with me as is then

torchvision/prototype/datasets/_builtin/coco.py
```diff
@@ -73,6 +95,62 @@ def resources(self, config: DatasetConfig) -> List[OnlineResource]:
...
+            # TODO: create a segmentation feature
+            segmentations=torch.stack(
+                [
+                    self._segmentation_to_mask(ann["segmentation"], is_crowd=ann["iscrowd"], image_size=image_size)
+                    for ann in anns
+                ]
+            ),
+            areas=torch.tensor([ann["area"] for ann in anns]),
+            crowds=torch.tensor([ann["iscrowd"] for ann in anns], dtype=torch.bool),
+            bounding_boxes=BoundingBox(
```
fmassa, line 122: For a future PR, I think it might be preferable to rename the class to `BoundingBoxes` as we hold more than one box now.

torchvision/prototype/datasets/_builtin/coco.py
```diff
@@ -73,6 +95,62 @@ def resources(self, config: DatasetConfig) -> List[OnlineResource]:
...
+        labels = [ann["category_id"] for ann in anns]
+        categories = [self.info.categories[label] for label in labels]
+        return dict(
+            # TODO: create a segmentation feature
+            segmentations=torch.stack(
+                [
+                    self._segmentation_to_mask(ann["segmentation"], is_crowd=ann["iscrowd"], image_size=image_size)
+                    for ann in anns
+                ]
+            ),
```
fmassa, line 119: Yes, we can even have a custom class that holds the raw polygons, and knows how to convert itself if needed. But can be discussed at a separate stage.

fmassa (approved):
LGTM, thanks!

fmassa:
ONNX-test error seems suspicious, but probably unrelated. I'm merging this but let's keep an eye on it

Issue #4611, opened 2021-10-13
Extend TorchVision's existing model builders to support multiple weights


datumbox:
### 🚀 The feature

_Note: To track the progress of the project check out [this board](https://github.com/pytorch/vision/projects/2)._

The existing model builders support only single pre-trained weights. 

To specify if the pretrained weights should be loaded, we pass a boolean `pretrained` parameter. Example:
https://github.com/pytorch/vision/blob/5f0edb97b46e5bff71dc19dedef05c5396eeaea2/torchvision/models/resnet.py#L321

Builders that require model composition (for example in Object Detection), receive an additional extra boolean parameter called `pretrained_backbone` which indicates if a pre-trained backbone should be loaded:
https://github.com/pytorch/vision/blob/5f0edb97b46e5bff71dc19dedef05c5396eeaea2/torchvision/models/detection/ssd.py#L551-L558

The above approach is limiting and doesn't allow us to support multiple weights for the same model builder.

### Motivation, pitch

The [Batteries Included project](https://github.com/pytorch/vision/issues/3911), enabled us to train far more accurate models than before. In order to make the new weights available to the users, we should extend the current model builders to support multiple weights. This can be done by converting the boolean parameters to an object that contains information about the weights.

### Alternatives

An alternative approach would be to modify the weight URLs in place (replace the old with new ones). Unfortunately this would break the BC guarantees of vision.

### Additional context

There was an original RFC with far bigger scope that discussed [Model Versioning](https://github.com/datumbox/dapi-model-versioning) and included a solution for multi-pretrained weights support. We've decided to separate this vision specific extension from the original RFC to cover our needs and give additional time until the model versioning problem is thoroughly investigated.

PR #4656, opened 2021-10-19
Refactor the backbone builders of detection

### LINKED ISSUE #4676: Refactor Models builders to make them reusable on Prototype
### 🚀 The feature

The current model builders are hard to reuse from the prototype area, leading to large amounts of copy-pasting. It would be good to refactor the space in a BC manner so that we can reuse as many parts as we can:
- [x] Segmentation models:
  - Update them to follow the standard convention of defining the builders on the same file as their class.
  - Make private model builders reusable from prototype.
  - Clean up the implementations and expose publicly variables that are currently pushed to kwargs.
- [x] Detection models:
  - Make private model builders reusable from prototype.
  - Refactor the VGG16 Features weights mechanism on SSD to work well with the proposed weights prototype. 

### Motivation, pitch

-

### Alternatives

_No response_

### Additional context

_No response_

datumbox:
Fixes #4676

Implements some of the same ideas on the detection space. More specifically, we move the CNN backbone initialization on the model builder and pass it directly to the method that implements the feature extractor for the specific task.

This PR will enable the work on #4611. Changes are BC, though we are forced to update an expected file because of the different order of initialization. See https://github.com/pytorch/vision/pull/4656#issuecomment-946955438 for details.

Accuracy remains the same for all pre-trained models.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -254,6 +254,27 @@ jobs:
             pip install --user pytest
             python test/test_onnx.py
 
+  prototype_test:
+    docker:
+      - image: circleci/python:3.7
+    steps:
+      - run:
+          name: Install torch
+          command: pip install --user --progress-bar=off --pre torch -f https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html
+      - run:
+          name: Install prototype dependencies
+          command: pip install --user --progress-bar=off git+https://github.com/pytorch/data.git
+      - checkout
+      - run:
+          name: Install torchvision
+          command: pip install --user --progress-bar off --no-build-isolation .
+      - run:
+          name: Install test utilities
+          command: pip install --user --progress-bar=off pytest
+      - run:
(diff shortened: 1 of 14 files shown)


datumbox:
@NicolasHug Thanks for the help!

> Do you think this is something worth mentioning in the future release notes, or nah?

Hm... I don't think we have done this before. This is probably because PyTorch doesn't guarantee getting the same random behaviour across versions. @fmassa any thoughts on this based on history?

We have 3 options here:
1. Say nothing.
2. Just mention it on the notes.
3. Ensure that the behaviour is identical with before. For that I just need to initialize the `LastLevelP6P7(256, 256)` before the backbone and we are good to go.

I don't mind going for any of the above solutions. I'll merge now but happy to follow up with option 3 if needed.

datumbox:
I've tested again all models and their metrics look fine. I'm merging this.

Issue #4528, opened 2021-10-02
Generate colors from random color palette in draw_bounding_boxes


oke-aditya:
### 🚀 The feature

The current default white color looks bad. Maybe we can generate colors from palette as we do in drawing masks?

### Motivation, pitch

It should be simple tweak to `_generate_color_palette` to pass num_boxes and get a List of RGB values. Instead of List[Tensor]. The fix is simple and that can be passed to get RGB colors.

### Alternatives

Users can anyway use `color="white"` to get white boxes.

### Additional context

It's simple issue and good-first. We can probably wait for some new contributor :smiley:  to join.

gessha:
I can take this if that's okay.

oke-aditya:
I'm glad that you want to take it. 
Just need a green signal from the maintainers @NicolasHug @datumbox

NicolasHug:
Thanks for the proposal @oke-aditya and @gessha for volunteering.

That sounds good to me, we should just make sure that the same color is generated for all boxes with the same label.

oke-aditya:
Well I'm not sure how we would ensure that same color is generated for all boxes with same label.
My thoughts were to generate color palette equal to number of boxes. And then we simply pick ith color from the palette.

We don't keep consistency of randomly generating same color for same label in drawing segmentation masks too.

NicolasHug:
> Well I'm not sure how we would ensure that same color is generated for all boxes with same label.

We can generate num_labels colors and then assign one color per label. Obviously we don't need to do that if `labels` isn't specified

> We don't keep consistency of randomly generating same color for same label in drawing segmentation masks too.

True but `draw_segmentation_masks` doesn't accept a label anyway, as it probably doesn't need one.

oke-aditya:
Great, I think it should be feasible with some refactor. As of now we plot each bbox and then the labels separately.

@gessha feel free to send a PR!

ABD-01:
Yes, I can if no one's working on it.

I guess the main change involved will be here
https://github.com/pytorch/vision/blob/a75dc89abbbb233469d1dbe2ee00ee012dd70522/torchvision/utils.py#L203-L204
Just to make the variable `colors` be a list returned from [`_generate_color_palette`](https://github.com/pytorch/vision/blob/a75dc89abbbb233469d1dbe2ee00ee012dd70522/torchvision/utils.py#L303), right?

oke-aditya:
Yes and use the colors the generated from the palette in draw_bounding_boxes. 

[...]

PR #4658, opened 2021-10-19
Random colors for draw_bounding_boxes

ABD-01:
Hi!
This pr is with reference to https://github.com/pytorch/vision/issues/4528

I have made the changes for generating random colors from the palette for bounding boxes.

But I am not getting why the tests are failing for the following cases even before the changes are done.
i.e. it is not accepting `colors="red"` or `colors="#FF00FF"` as valid inputs
```
======================================================== short test summary info =========================================================
FAILED test/test_utils.py::test_draw_boxes_colors[red] - ValueError: unknown color specifier: 'r'
FAILED test/test_utils.py::test_draw_boxes_colors[#FF00FF] - ValueError: unknown color specifier: '#'
FAILED test/test_utils.py::test_draw_boxes_colors[colors4] - IndexError: tuple index out of range
FAILED test/test_utils.py::test_draw_boxes_grayscale - TypeError: Cannot handle this data type: (1, 1, 1), |u1
FAILED test/test_utils.py::test_draw_invalid_boxes - Failed: DID NOT RAISE <class 'ValueError'>
```
 
Thanks to @oke-aditya for pseudocode.


oke-aditya:
Also 
```
if fill:
    if color is None:
fill_color = (255, 255, 255, 100)
```

Do you think we will ever hit color is None here in fill? we can just remove these lines I guess?

torchvision/utils.py
```diff
@@ -199,10 +199,16 @@ def draw_bounding_boxes(
 
     txt_font = ImageFont.load_default() if font is None else ImageFont.truetype(font=font, size=font_size)
 
+    if colors is None:
+        if labels is None:
+            colors = _generate_color_palette(len(img_boxes))
+        else:
+            assert len(labels) == len(img_boxes)
```
oke-aditya, line 206: But why do we want len(labels) to be same as len(boxes)?

I thought we should assume len(labels) < len(boxes).
As we do the same in drawing masks

```
if colors is not None and num_masks > len(colors):
    raise ValueError(f"There are more masks ({num_masks}) than colors ({len(colors)})")
```

I thought if a box is unlabelled, we will just draw a random color.

In case a box is not labelled 
```
if ! label in label_color_map.keys():
   colors.append(None) (Or generate a random color)
```

NicolasHug: > Is it really necessary for every box to be labelled?

I would say so. Another way to phrase the question is: Why would someone label the first N boxes, but not the rest?

I can't think of a good reason to do so, and I don't think this is something we should be bothered to support anyway. Users can always pass empty strings as the labels if they reaaaaally have a use-case where only the first few boxes have labels.

ABD-01: I agree with Nicolas on this

oke-aditya: I suggest to replace plain assert by raising ValueError and suggesting user what Nicolas said

torchvision/utils.py
```diff
@@ -199,10 +199,16 @@ def draw_bounding_boxes(
...
+            label_color_map = dict(zip(labels, _generate_color_palette(len(labels))))
```
oke-aditya, line 207: Since now we agree that len(labels) == len(img_boxes) Line 207 and 204 can be simplified.

datumbox:
Closing in favour of #5127

PR #4721, opened 2021-10-22
add prototype features

pmeier:
For a seamless interplay between the new prototype datasets and upcoming transforms, the datasets need to forward information about what each item of a sample is in some form or way. In pmeier/torchvision-datasets-rework#1it was decided that the best way to do this is by subclassing `torch.Tensor` and wrap the items in custom classes for each type.

An additional benefit is that these custom `Feature`'s can carry extra meta data. For example in this PR I've added a `color_space` attribute for images and a `category` attribute for labels`:

```python

for sample in datasets.load("fashionmnist"):
    image, label = sample["image"], sample["label"]
    print(image.shape, image.color_space)
    print(label, label.category)
    break
```

```
torch.Size([28, 28]) ColorSpace.GRAYSCALE
tensor(6) Shirt
```

cc @pmeier @vfdev-5 @datumbox @bjuncek


torchvision/prototype/features.py
```diff
@@ -0,0 +1,104 @@
...
+        types: Tuple[Type[torch.Tensor], ...],
+        args: Sequence[Any] = (),
+        kwargs: Optional[Mapping[str, Any]] = None,
+    ) -> torch.Tensor:
+        with DisableTorchFunction():
+            output = func(*args, **(kwargs or dict()))
+        if func is not torch.Tensor.clone:
+            return output
+
+        return cls(output, like=args[0])
```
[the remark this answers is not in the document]
vfdev-5: I'd like to chime into this discussion. As far as I understand the purpose of these tensor subclasses: Feature, Image, Label is that we could correctly dispatch torchvision transformations on image, bboxes, masks, labels etc.

- lifetime of Feature, Image, Label subclasses. IMO they should somehow persist through all transforms.

For example, for a detection task with very basic "hflip"-like transform we have the following transformations to apply on the image and the set of bboxes:
```python
transforms = T.Compose([
    T.RandomHorizontalFlip(p=0.5),
    T.ConvertImageDtype(torch.float),
])
```
When we apply `torch.flip` from `RandomHorizontalFlip` we may still want to know that `ConvertImageDtype` is only applicable to images and not bboxes.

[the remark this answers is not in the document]
pmeier: True, but inside `HorizontalFlip` we can simply manually do 

```python
return BoundingBox(torch.flip(input), like=input)
```

We can also go a step further and wrap each feature transform for more convenience:

```python
def convert_output_type(feature_transform):
    def wrapper(input, **kwargs):
        input_type = type(input)
        output = feature_transform(input, kwargs)
        if type(output) is not input_type:
            output = input_type(output, like=input)
        return output

    return wrapper


@convert_output_type
def horizontal_flip(input: BoundingBox):
    return input.flip()
```

test/test_prototype_features.py
```diff
@@ -0,0 +1,107 @@
+import functools
+
+import pytest
+import torch
+from torch.testing import make_tensor as _make_tensor, assert_close
+from torchvision.prototype import features
+from torchvision.prototype.datasets.utils._internal import FrozenMapping
```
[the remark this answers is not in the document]
pmeier: I don't recall the exact reason, but I remember I was hitting an issue with serialization / jit / ... with normal dicts. I'll try to reproduce and update here. If I can't reproduce, I agree, a plain dict is probably better.

test/test_prototype_features.py
```diff
@@ -0,0 +1,170 @@
...
+        expected = features.BoundingBox([6, 1, 10, 3], format="xyxy", image_size=image_size)
+
+        actual_eager = compose(input, size)
+        assert_close(actual_eager, expected)
+
+        sample_inputs = (features.BoundingBox(torch.zeros((4,)), image_size=(10, 10)), torch.tensor((20, 5)))
+        actual_jit = torch.jit.trace(compose, sample_inputs, check_trace=False)(input, size)
+        print(actual_jit)
+        print(expected)
+        assert_close(actual_jit, expected)
```
pmeier, line 170: This test fails, but I was unable to get it to pass. The eager variant works as expected, but the jitted one gives incorrect output. I'm not sure if the test setup is wrong somehow or of JIT simply doesn't work with tensor subclasses that carry additional meta data. This needs to be reviewed by someone with more JIT experience than I have. cc @fmassa @datumbox

fmassa: Pinging @Chillee here, but my gut feeling is that attributes of the Tensor subclass might be embedded in the graph as constants.

fmassa: You can check the computation graph of the traced model (via `print(traced_function.code)`) to see if the `image_size` was embedded as a constant or not.

Another thing would be to make the `image_size` for `sample_inputs` to be the same of for the `input`, which should make tests pass (but it is still a bug)

pmeier: You are probably right:

```python
def compose(input: Tensor,
    size: Tensor) -> Tensor:
  new_height, new_width, = torch.unbind(size)
  height_scale = torch.div(new_height, CONSTANTS.c0)
  width_scale = torch.div(new_width, CONSTANTS.c0)
  data = torch.clone(input)
  data0 = torch.to(data, torch.device("cpu"), 6)
  old_x1, old_y1, old_x2, old_y2, = torch.unbind(torch.detach(data0), -1)
  a = torch.mul(old_x1, width_scale)
  b = torch.mul(old_y1, height_scale)
  c = torch.mul(old_x2, width_scale)
  d = torch.mul(old_y2, height_scale)
  data1 = torch.stack([a, b, c, d], -1)
  data2 = torch.to(data1, torch.device("cpu"), 6)
  x, y, x2, y2, = torch.unbind(torch.detach(data2), -1)
  c0 = torch.sub(x2, x)
  d0 = torch.sub(y2, y)
  data3 = torch.stack([x, y, c0, d0], -1)
  data4 = torch.to(data3, torch.device("cpu"), 6)
  x0, b0, w, d1, = torch.unbind(torch.detach(data4), -1)
  a0 = torch.rsub(torch.add(x0, w), 5)
  data5 = torch.stack([a0, b0, w, d1], -1)
  data6 = torch.to(data5, torch.device("cpu"), 6)
  x1, y1, w0, h, = torch.unbind(torch.detach(data6), -1)
  c1 = torch.add(x1, w0)
  d2 = torch.add(y1, h)
  data7 = torch.stack([x1, y1, c1, d2], -1)
  data8 = torch.to(data7, torch.device("cpu"), 6)
  return torch.detach(data8)
```

The important lines are

```python
new_height, new_width, = torch.unbind(size)
height_scale = torch.div(new_height, CONSTANTS.c0)
width_scale = torch.div(new_width, CONSTANTS.c0)
```

which come from 

```python
old_height, old_width = input.image_size
new_height, new_width = size

height_scale = new_height / old_height
width_scale = new_width / old_width
```

So `old_height` and `old_width`, which come directly from the `BoundingBox.image_size` meta data, are stored as constants.

fmassa: Here is what I propose: let's forget torchscript for now and continue with the prototype

PR #4767, opened 2021-10-27
Change to stable sort in nms implementations

jdsgomes:
relates to https://github.com/pytorch/vision/pull/4491
triggered by https://github.com/pytorch/vision/pull/4766

I am trying to replicate the errors and follow up on the work done in the initial PR in order to fix them and introduce the stable sort (given that is acceptably fast) 

This is an investigation PR and pretty much work in progree


jdsgomes:
Benchmark results are reassuring, the difference between using stable sort and not using seems negligible.

Screenshot 2021-11-19 at 11 46 45
Screenshot 2021-11-19 at 11 46 28

Code used for benchmark:

```python
import torch
from time import time


def _create_tensors_with_iou(N, iou_thresh):
    # force last box to have a pre-defined iou with the first box
    # let b0 be [x0, y0, x1, y1], and b1 be [x0, y0, x1 + d, y1],
    # then, in order to satisfy ops.iou(b0, b1) == iou_thresh,
    # we need to have d = (x1 - x0) * (1 - iou_thresh) / iou_thresh
    # Adjust the threshold upward a bit with the intent of creating
    # at least one box that exceeds (barely) the threshold and so
    # should be suppressed.
    boxes = torch.rand(N, 4) * 100
    boxes[:, 2:] += boxes[:, :2]
    boxes[-1, :] = boxes[0, :]
    x0, y0, x1, y1 = boxes[-1].tolist()
    iou_thresh += 1e-5
    boxes[-1, 2] += (x1 - x0) * (1 - iou_thresh) / iou_thresh
    scores = torch.rand(N)
    return boxes, scores

def _create_random_boxes(N):
    boxes = torch.rand(N, 4) * 100
    scores = torch.rand(N)
    return boxes, scores

run_on="cuda"
for n in range(100, 10001, 100):
    boxes, scores = _create_random_boxes(n)
    boxes = boxes.to(device=run_on)
    scores = scores.to(device=run_on)
    start = time()
    keep = ops.nms(boxes, scores, 0.5)
    end = time()
    time_for_random = end-start
    boxes, scores = _create_tensors_with_iou(n, 0.5)
    boxes = boxes.to(device=run_on)
    scores = scores.to(device=run_on)
    start = time()
    keep = ops.nms(boxes, scores, 0.5)
    end = time()
    time_for_twiou = end - start
    print(f'{n}\t{time_for_random}\t{time_for_twiou}')
```

datumbox (approved):
Thanks @jdsgomes, this looks great to me! 

This might resolve a lot of the flakiness we face on the model tests. Could you create a new issue to investigate if upon merging this, we can now remove hacks/workaround from:
https://github.com/pytorch/vision/blob/f093d0821194ed55f80f4764cf56e744f9217533/test/test_models.py#L658-L672

Note that it's likely that we will need to recreate the expected values of the tests. This can be part of our test improvement efforts work for next half. cc @NicolasHug @pmeier

NicolasHug:
I'll add the BC-breaking tag just so that we remember to add a short note in the release notes. This isn't BC-breaking strictly speaking as the order of ties isn't guaranteed, but it's be nice to add a short note to let users know that they might get slightly different results in some rare cases.


Regarding the benchmark: for GPU benchmarks we should call `torch.cuda.synchronize()` before calling `time()`. This is because CUDA kernels run asynchronously, so you can reach `time()` before the GPU operation has actually finished.
Optionally you can also use pytorch benchmarks utils: https://pytorch.org/tutorials/recipes/recipes/benchmark.html

PR #4783, opened 2021-10-28
Big doc revamp - simplify and improve the subpackage pages

NicolasHug:
This PR adds a minor update to the side TOC by changing the order and the title of the sections, instead of the non-informative module names (see screenshot) (EDIT: order was upated since screenshot).

This PR also adds a major update which consist in putting each function and class into their own page. This adds a few benefits:

- it make the pages much shorter, and easier to navigate
- it makes classes and functions more discoverable, as there is much more that fits into one screen
- it makes the structure of each page easier to parse and understand, which will greatly help better organizing the docs in the future (see note below about models)
- it will provide much better google analytics feedback

Note: the model page still needs a significant update, more PRs will come later. 

Here are the rendered docs: https://945987-73328905-gh.circle-artifacts.com/0/docs/index.html vs the current docs: https://pytorch.org/vision/stable/index.html

Screenshot 2021-10-29 at 23 54 26


datumbox (approved):
I understand that more work is needed to iron out some of the issues but I think this is already a major improvement in terms of readability. I would be in favour of merging it and continue changes on a follow up PR.

docs/source/ops.rst
```diff
@@ -1,7 +1,7 @@
 .. _ops:
 
-torchvision.ops
-===============
+Operators
```
oke-aditya, line 3: I wanted to suggest this from long time, current operators page look a bit bloated.
We should probably divide in sections.
E.g. operators on boxes, operators on masks, detection operators, etc.

docs/source/utils.rst
```diff
@@ -1,14 +1,18 @@
...
-.. autofunction:: save_image
+.. currentmodule:: torchvision.utils
 
-.. autofunction:: draw_bounding_boxes
+.. autosummary::
+    :toctree: generated/
+    :template: function.rst
 
-.. autofunction:: draw_segmentation_masks
+    draw_bounding_boxes
```
oke-aditya, line 15: And specific reason to change the order?
It's kinda nice to push `make_grid` down as we don't actively maintain it.

docs/source/index.rst
```diff
@@ -31,12 +31,12 @@ architectures, and common image transformations for computer vision.
    :maxdepth: 2
    :caption: Package Reference
 
+   models
```
oke-aditya, line 34: Somehow the brain of users is accustomed to see docs in the previous order. 
It might be unpleasant to change to order. I love the previous order. And kind off in presentations related to torchvision, E.g. Francisco's at PyTorch day, he presents by the previous order,

Would love the old order :)

docs/source/models.rst
```diff
@@ -1,10 +1,10 @@
 .. _models:
 
-torchvision.models
-##################
+Models and pre-trained weights
```
oke-aditya, line 3: Probably write in Title Case. As that's how PyTorch docs on left are

image

docs/source/index.rst
```diff
@@ -31,12 +31,12 @@ architectures, and common image transformations for computer vision.
...
+   transforms
    datasets
+   ops
    io
-   models
    feature_extraction
```
oke-aditya, line 39: I think We should club this with models, just after it.

oke-aditya (commented):
This is so awesome :heart: @NicolasHug 

I have few thoughts, hope you don't mind :sweat_smile:

docs/source/utils.rst
NicolasHug: >  It's kinda nice to push make_grid down as we don't actively maintain it.

That'st the reason :)

docs/source/models.rst
NicolasHug: I'm not sure they're using Title Case - see e.g. "Features or large-scale deployments" or "Autograd mechanics".

They're suing "Automatic Mixed Prececision" and "Distributed Data Parallel" and "Frequently Asked Questions" because these are commonly known and shortened AMP,  DDP, FAQ, etc.

docs/source/index.rst
NicolasHug: > We should club this with models

I agree, I think eventually we will move it into the "Models and pre-trained weights" section. It is `torchvision.models.feature_extraction` after all. But since the models section needs a much bigger revamp, I wanted to leave this out for now

docs/source/index.rst
NicolasHug: The old order is alphabetical, and might not reflect the order of what users may be looking for.

I think that users are more interested in models, transforms and ops than say i.o. operations, so I changed the order to reflect the importance of each section. Obviously it's not set in stone and I'm happy to hear other's thoughts, but I don't think we should worry too much about current readers' habits. It's still very obvious which section corresponds to what, so users will get used to this one fast :)

docs/source/ops.rst
NicolasHug: Definitely, one positive thing about the changes in this PR is that it surfaces our current doc structure weaknesses and it makes them more obvious, which will allow us to fix them.

To keep things manageable though, I would leave this out for future improvements, otherwise there will be too many things to review at once

docs/source/index.rst
oke-aditya: So let's arrange according to usage order. Simple LRU caching here :smile: 

Datasets like MNIST, ImageNet are universal, so probably everyone needs it.
Transforms to convert ToTensor etc Are almost in every code.
Models are not used all places.
Ops are then used a bit more scarce.
Utils, are least used I think,

So the order datasets, transforms, models, ops, utils :)

NicolasHug: Thanks for the input @oke-aditya , I will update accordingly

fmassa (approved):
This looks great, thanks!

Issue #4805, opened 2021-10-31
AutoAugment augmentation space goes over uint8 range.


jopo666:
### 🐛 Describe the bug

Simple error with a simple fix.

https://github.com/pytorch/vision/blob/48ebc0bbdb12089eca557d52a8db7b0674b23486/torchvision/transforms/autoaugment.py#L226

256 is cannot be represented as an `uint8` and the range should start at 255.

## How to reproduce.
```python
import torch
import torchvision.transforms as T

aug_space = torch.linspace(256.0, 0.0, 10)
image = torch.randint(0, 225, (3, 512, 512), dtype=torch.uint8)

T.functional.solarize(image, aug_space[0])
```
```error
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
my_awesome_network.py in <module>
      1 image = torch.randint(0, 225, (3, 512, 512), dtype=torch.uint8)
      2 
----> 3 T.functional.solarize(image, aug_space[0])

path/to/torchvision/transforms/functional.py in solarize(img, threshold)
   1318         return F_pil.solarize(img, threshold)
   1319 
-> 1320     return F_t.solarize(img, threshold)
   1321 
   1322 

path/to/torchvision/torchvision/transforms/functional_tensor.py in solarize(img, threshold)
    872 
    873     inverted_img = invert(img)
--> 874     return torch.where(img >= threshold, inverted_img, img)
    875 
    876 

RuntimeError: value cannot be converted to type uint8_t without overflow: 256
```

### Versions

Collecting environment information...
PyTorch version: 1.10.0+cu113
Is debug build: False
CUDA used to build PyTorch: 11.3
ROCM used to build PyTorch: N/A

OS: Ubuntu 20.04.3 LTS (x86_64)
GCC version: (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0
Clang version: Could not collect
CMake version: Could not collect
Libc version: glibc-2.33

Python version: 3.8.12 (default, Oct 12 2021, 13:49:34)  [GCC 7.5.0] (64-bit runtime)
Python platform: Linux-5.11.0-38-generic-x86_64-with-glibc2.17
Is CUDA available: True
CUDA runtime version: Could not collect
GPU models and configuration: GPU 0: Tesla V100-PCIE-16GB
Nvidia driver version: 470.63.01
cuDNN version: Could not collect
HIP runtime version: N/A
MIOpen runtime version: N/A

Versions of relevant libraries:
[...]

PR #4806, opened 2021-10-31
Fix augmentation space to be uint8 compatible

jopo666:
A simple fix that addresses #4805.

cc @vfdev-5 @datumbox


jopo666:
Maybe unit tests should also be extended to cover these types of bugs?

datumbox (approved):
LGTM, thanks for fixing it.

prabhat00155 (approved):
Thanks @jopo666! Could you also perhaps add a unit test to cover this case?

datumbox:
@prabhat00155 What are your thoughts on testing this?

On the AutoAugmentation module we currently don't have any similar tests to check for misconfigurations of the methods and it's unclear how such a test would look like. One place we could potentially add it is on the implementation of `solarize`. For example we could add an assertion to ensure that the threshold is within bounds and then create a test to check the assertion is triggered:

https://github.com/pytorch/vision/blob/d367a01a18a3ae6bee13d8be3b63fd6a581ea46f/torchvision/transforms/functional_tensor.py#L885-L886

Probably a similar assertion is worth to be added at `posterize` for the number of bits.

Finally I wonder if it's worth merging this so that we cover the bug and add the tests on a follow up PR. No strong opinions, just want to make sure this bug fix will be merged. :)

prabhat00155:
> @prabhat00155 What are your thoughts on testing this?
> 
> On the AutoAugmentation module we currently don't have any similar tests to check for misconfigurations of the methods and it's unclear how such a test would look like. One place we could potentially add it is on the implementation of `solarize`. For example we could add an assertion to ensure that the threshold is within bounds and then create a test to check the assertion is triggered:
> 
> https://github.com/pytorch/vision/blob/d367a01a18a3ae6bee13d8be3b63fd6a581ea46f/torchvision/transforms/functional_tensor.py#L885-L886
> 
> Probably a similar assertion is worth to be added at `posterize` for the number of bits.
> 
> Finally I wonder if it's worth merging this so that we cover the bug and add the tests on a follow up PR. No strong opinions, just want to make sure this bug fix will be merged. :)

Thanks @datumbox for the explanation. Makes sense to merge this PR then. We could look at adding tests in a follow-up PR.

datumbox:
@prabhat00155 Sounds great, thanks for the flexibility. I've created a ticket for this so that we don't forget: #4818 

Feel free to merge on green CI.

PR #4807, opened 2021-10-31
add val split and year 2017 to prototype coco

pmeier:
cc @pmeier @bjuncek

### DIFF
--- a/torchvision/prototype/datasets/_builtin/coco.py
+++ b/torchvision/prototype/datasets/_builtin/coco.py
@@ -41,29 +41,36 @@ class Coco(Dataset):
             type=DatasetType.IMAGE,
             homepage="https://cocodataset.org/",
             valid_options=dict(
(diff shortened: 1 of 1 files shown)


torchvision/prototype/datasets/_builtin/coco.py
```diff
@@ -41,29 +41,36 @@ def _make_info(self) -> DatasetInfo:
             type=DatasetType.IMAGE,
             homepage="https://cocodataset.org/",
             valid_options=dict(
-                split=("train",),
-                year=("2014",),
+                split=("train", "val"),
+                year=("2014", "2017"),
```
fmassa, line 45: Should we make `2017` the default year? It's much more widely used now

fmassa (approved):
Looks great, thanks a lot!

Only have one minor question, but it can be addressed in a follow-up PR

PR #4810, opened 2021-11-01
Feature extraction default arguments -  ops

### LINKED ISSUE #4777: The symbolic tracer of `feature_extraction` should consider all `torchvision.ops` as leaf nodes
### 🚀 The feature

The `create_feature_extractor` method should by default pass as `leaf_modules` all `torchvision.ops` layers.

https://github.com/pytorch/vision/blob/d367a01a18a3ae6bee13d8be3b63fd6a581ea46f/torchvision/models/feature_extraction.py#L394-L396

### Motivation, pitch

Currently the FX tracers consider all `torch.nn` layers leaf nodes. By doing the same for torchvision we will align the two behaviours.

cc @datumbox @alexander-soare

jdsgomes:
Fixes #4777

Should only be merged after https://github.com/pytorch/vision/pull/4951

cc @datumbox @alexander-soare


jdsgomes:
There is one test failing. I think this test fails because the tracing methods expect `x: Tensor` as argument in the forward pass, but in [`MultiScaleRoIAlign`](https://github.com/pytorch/vision/blob/fcd7603bab639d32dbbe8f3c261855474a8d639d/torchvision/ops/poolers.py#L181) we have `x: Dict[str, Tensor]`. 
Should this test be removed, or should we change the tracing functions instead in order to accommodate for this use case? 

cc @fmassa

alexander-soare:
@jdsgomes just noticed this and decided to check it out. I'm getting a Proxy related error when trying to enumerate the dict you mentioned. So it seems `MultiScaleRoiAlign` is not being wrapped.

I think I misled you with the `autowrap_modules` idea... I could swear when I tested it that it included all the class definitions within the module, but now I realised that it doesn't... It only includes function definitions. If you explicitly add `MultiScaleRoiAlign` as a leaf module in the tracer kwargs it ends up working.

To illustrate what I'm saying, make a folder called `some_module` with `thing.py`:

```python
import torch


def untraceable(x):
    int(x)


class UnTraceable(torch.nn.Module):
    def forward(self, x):
        int(x)
```

and the `__init__.py`

```python
from .thing import untraceable, UnTraceable
```

And then make a `test.py`

```python
import torch
import math
import some_module
from some_module import untraceable, UnTraceable


    def __init__(self):
        super().__init__()

    def forward(self, x, b, o):
        untraceable(x)


model = MyModule()

train_nodes, eval_nodes = get_graph_node_names(
    model,
    tracer_kwargs={'autowrap_modules': (math, some_module.thing)})
```

You will find that `model=MyModule()` works but `model=UnTraceable()` fails. But in the later case, if you update `tracer_kwargs` with ` {'leaf_modules': (UnTraceable,)}` it works.

jdsgomes:
@alexander-soare, sorry I missed your last comment and was working on this today so I am was about to propose a different solution (please see the latest commit). 

Regarding your comment, if we can make it work I believe is much better than the solution I propose, but I I couldn't make it work by just tried the following locally:

```
from torchvision import ops

def make_MultiScaleRoIAlign_obj(fmap_names=["0"], output_size=(7, 7), sampling_ratio=2, **kwargs):
        return ops.poolers.MultiScaleRoIAlign(fmap_names, output_size, sampling_ratio).to(device='cpu')
```

and when I run 

```
get_graph_node_names(
            make_MultiScaleRoIAlign_obj(),
            tracer_kwargs={'leaf_modules': (ops.poolers.MultiScaleRoIAlign,)}
    
        )
```

I get the original error

```
TraceError                                Traceback (most recent call last)
<ipython-input-15-64ccbcc41b7d> in <module>
      1 from torchvision.ops.poolers import MultiScaleRoIAlign
      2 
----> 3 get_graph_node_names(
      4             make_MultiScaleRoIAlign_obj(),
      5             tracer_kwargs={'leaf_modules': (MultiScaleRoIAlign,)}

~/vision/torchvision/models/feature_extraction.py in get_graph_node_names(model, tracer_kwargs, suppress_diff_warning)
    216     is_training = model.training
    217     train_tracer = NodePathTracer(**tracer_kwargs)
--> 218     train_tracer.trace(model.train())
    219     eval_tracer = NodePathTracer(**tracer_kwargs)
    220     eval_tracer.trace(model.eval())

~/homebrew/anaconda3/lib/python3.8/site-packages/torch/fx/_symbolic_trace.py in trace(self, root, concrete_args)
    616                 for module in self._autowrap_search:
    617                     _autowrap_check(patcher, module.__dict__, self._autowrap_function_ids)
--> 618                 self.create_node('output', 'output', (self.create_arg(fn(*args)),), {},
    619                                  type_expr=fn.__annotations__.get('return', None))
    620 

~/vision/torchvision/ops/poolers.py in forward(self, x, boxes, image_shapes)
    213         """
    214         x_filtered = []
--> 215         for k, v in x.items():
    216             if k in self.featmap_names:
    217                 x_filtered.append(v)

~/homebrew/anaconda3/lib/python3.8/site-packages/torch/fx/proxy.py in __iter__(self)
    246             return (self[i] for i in range(inst.argval))  # type: ignore[index]
    247 
--> 248         return self.tracer.iter(self)
    249 
    250     def __bool__(self) -> bool:

~/homebrew/anaconda3/lib/python3.8/site-packages/torch/fx/proxy.py in iter(self, obj)
    159         information to the graph node using create_node and can choose to return an iterator.
    160         """
--> 161         raise TraceError('Proxy object cannot be iterated. This can be '
    162                          'attempted when the Proxy is used in a loop or'
    163                          ' as a *args or **kwargs function argument. '

TraceError: Proxy object cannot be iterated. This can be attempted when the Proxy is used in a loop or as a *args or **kwargs function argument. See the torch.fx docs on pytorch.org for a more detailed explanation of what types of control flow can be traced, and check out the Proxy docstring for help troubleshooting Proxy iteration errors
```

When looking at https://github.com/pytorch/pytorch/blob/master/torch/fx/proxy.py#L202 I decided to extract the logic from the forward function to a separate one and wrap it.

Let me know your thoughts.

alexander-soare:
@jdsgomes thanks for the summary. I originally wouldn't have considered making a change to the TorchVision code itself. Also, your changes only fix that error, but we still aren't ensuring all modules from `torchvision.ops` are treated as leaves (and I understand the latter is the original intent of https://github.com/pytorch/vision/issues/4777).

So if that is indeed the case then we should catch all like this:

```python
tracer_kwargs: Dict = {
        "autowrap_modules": (
            math,
            torchvision.ops,
        ),
        "leaf_modules": tuple(
            obj for _, obj in inspect.getmembers(torchvision.ops) if inspect.isclass(obj) and issubclass(obj, torch.nn.Module)
        )
    }
```

In any case, it gets kind of ugly when a user wants to actually specify `tracer_kwargs` because they then need to add in their own kwargs plus all the above. This is why I previously suggested burying these kwargs into `LeafModuleAwareTracer`. 

Another different point is that the above still won't pass your tests. That's because the root module can't be a leaf module (it would make no sense). So you need to bury the module in a:

```python
class TestModule(nn.Module):
    def __init__(self):
        self.layer = Class()

    def forward(self, *args):
        self.layer(*args)
```

then pass TestModule to `create_feature_extractor`.

On yet another note, if you stick with keeping the kwargs in `create_feature_extractor` I might consider also putting them in `get_graph_node_names`. In your tests and code it looks like you did it for one but not the other, but I can't see why anyone would want them to be treated differently.

alexander-soare:
Also, to perhaps inspire yet another way of doing things, here's how I do it in timm: https://github.com/rwightman/pytorch-image-models/blob/65d827c7a6739b20dcd4c57216f20adc521a6b2a/timm/models/fx_features.py

I explicitly import the ones from timm.models.layers that I want to treat as leaf modules, and then I use a registration decorator for everything else.

Not saying that's how you should do it here though.

jdsgomes:
@alexander-soare thank you again for the comments, very helpful!

I decided to go for the the original solution (thanks for pointing out the missing `leaf_modules`) because I think it is more explicit to the caller.

torchvision/ops/poolers.py
```diff
@@ -9,6 +10,158 @@
 from .roi_align import roi_align
 
 
+def _convert_to_roi_format(boxes: List[Tensor]) -> Tensor:
```
jdsgomes, line 13: copy-pasted from  from MultiScaleRoIAlign

torchvision/ops/poolers.py
```diff
@@ -9,6 +10,158 @@
...
+    device, dtype = concat_boxes.device, concat_boxes.dtype
+    ids = torch.cat(
+        [torch.full_like(b[:, :1], i, dtype=dtype, layout=torch.strided, device=device) for i, b in enumerate(boxes)],
+        dim=0,
+    )
+    rois = torch.cat([ids, concat_boxes], dim=1)
+    return rois
+
+
+def _infer_scale(feature: Tensor, original_size: List[int]) -> float:
```
jdsgomes, line 24: copy-pasted from from MultiScaleRoIAlign

torchvision/ops/poolers.py
```diff
@@ -9,6 +10,158 @@
...
+    possible_scales: List[float] = []
+    for s1, s2 in zip(size, original_size):
+        approx_scale = float(s1) / float(s2)
+        scale = 2 ** float(torch.tensor(approx_scale).log2().round())
+        possible_scales.append(scale)
+    assert possible_scales[0] == possible_scales[1]
+    return possible_scales[0]
+
+
+def _setup_scales(
```
jdsgomes, line 36: extracted from MultiScaleRoIAlign. please review carefully has this is not just a copy paste, as the instance variables referenced originally became arguments or return values

torchvision/ops/poolers.py
```diff
@@ -9,6 +10,158 @@
...
+        int(lvl_min),
+        int(lvl_max),
+        canonical_scale=canonical_scale,
+        canonical_level=canonical_level,
+    )
+    return scales, map_levels
+
+
+@torch.fx.wrap
+def multiscale_roi_align(
```
jdsgomes, line 63: forward method from MultiScaleRoIAlign. References and usage of instance variables were removed so please review this changes in detail to make sure no unintended changes in the logic were made

test/test_ops.py
```diff
@@ -422,12 +458,15 @@ def test_boxes_shape(self):
 
 
 class TestMultiScaleRoIAlign:
+    def make_obj(fmap_names=["0"], output_size=(7, 7), sampling_ratio=2, **kwargs):
```
datumbox, line 461: Is this meant to be static? Or you missed the `self`?

test/test_ops.py
```diff
@@ -422,12 +458,15 @@ def test_boxes_shape(self):
...
+        return ops.poolers.MultiScaleRoIAlign(fmap_names, output_size, sampling_ratio)
+
     def test_msroialign_repr(self):
         fmap_names = ["0"]
         output_size = (7, 7)
         sampling_ratio = 2
         # Pass mock feature map names
-        t = ops.poolers.MultiScaleRoIAlign(fmap_names, output_size, sampling_ratio)
+        t = TestMultiScaleRoIAlign.make_obj(fmap_names, output_size, sampling_ratio)
```
datumbox, line 469: Was this supposed to be as follows?
```suggestion
        t = self.make_obj(fmap_names, output_size, sampling_ratio)
```

test/test_ops.py
```diff
@@ -436,6 +475,12 @@ def test_msroialign_repr(self):
         )
         assert repr(t) == expected_string
 
+    @pytest.mark.parametrize("device", cpu_and_gpu())
+    def test_feature_extractor(self, device):
+        op_obj = TestMultiScaleRoIAlign.make_obj().to(device=device)
```
datumbox, line 480: same here, unclear why we do static calls if the base-class defines `make_obj()` as a non-static method.

test/test_ops.py
```diff
@@ -693,6 +738,30 @@ def get_fn_args(self, device, contiguous, batch_sz, dtype):
 
         return x, weight, offset, mask, bias, stride, pad, dilation
 
+    def make_obj(
+        self, device, contiguous, batch_sz, dtype, in_channels=6, out_channels=2, kernel_size=(3, 2), groups=2
+    ):
+        _, _, _, _, _, stride, padding, dilation = self.get_fn_args(device, contiguous, batch_sz, dtype)
```
datumbox, line 744: The `get_fn_args()` can be quite slow. Since you only care about the stride, padding and dilation which are hardcoded in the method, you can just set their values.

torchvision/models/feature_extraction.py
```diff
@@ -172,8 +175,25 @@ def _warn_graph_differences(train_tracer: NodePathTracer, eval_tracer: NodePathT
     warnings.warn(msg + suggestion_msg)
 
 
+def _get_leaf_modules_for_ops():
+    members = inspect.getmembers(torchvision.ops)
+    result = []
+    for _, obj in members:
+        if inspect.isclass(obj) and issubclass(obj, torch.nn.Module):
+            result.append(obj)
```
datumbox, line 183: Future Nit: Perhaps a registration mechanism would be useful here. We got one at the new Datasets and one planned for Models. Perhaps operators could benefit from it (TBD).

torchvision/models/feature_extraction.py
```diff
@@ -172,8 +175,25 @@ def _warn_graph_differences(train_tracer: NodePathTracer, eval_tracer: NodePathT
...
 def get_graph_node_names(
-    model: nn.Module, tracer_kwargs: Dict = {}, suppress_diff_warning: bool = False
+    model: nn.Module,
+    tracer_kwargs: Dict = {
+        "autowrap_modules": (
+            math,
+            torchvision.ops,
+        ),
+        "leaf_modules": _get_leaf_modules_for_ops(),
+    },
```
datumbox, line 195: If you define the default value like that, it's going to be mutated for all future calls. Moreover the method calls to get the values won't happen on the first call but on the method declaration.

It's best to instantiate with something non-mutable, like `None` for instance and then if the value is still None, set the default within the method. This idiom is very common in TorchVision and is mainly done for such cases.

torchvision/models/feature_extraction.py
```diff
@@ -294,7 +314,13 @@ def create_feature_extractor(
...
     train_return_nodes: Optional[Union[List[str], Dict[str, str]]] = None,
     eval_return_nodes: Optional[Union[List[str], Dict[str, str]]] = None,
-    tracer_kwargs: Dict = {},
+    tracer_kwargs: Dict = {
+        "autowrap_modules": (
+            math,
+            torchvision.ops,
+        ),
+        "leaf_modules": _get_leaf_modules_for_ops(),
+    },
```
datumbox, line 323: same here

torchvision/ops/poolers.py
```diff
@@ -9,40 +10,6 @@
...
-
-# TODO: (eellison) T54974082 https://github.com/pytorch/pytorch/issues/26744/pytorch/issues/26744
-def initLevelMapper(
-    k_min: int,
-    k_max: int,
-    canonical_scale: int = 224,
-    canonical_level: int = 4,
-    eps: float = 1e-6,
-):
-    return LevelMapper(k_min, k_max, canonical_scale, canonical_level, eps)
```
datumbox, line 43: I understand you moved both methods at the bottom without modifications, correct? If yes then I agree that it's best to define them after the method.

Could you separate the refactoring of this file on a separate PR, so that we make the reviewing easier?

torchvision/ops/poolers.py
datumbox: I understand you just move this outside of the operator to share code, right? If that's the case, I would include this on the refactoring PR.

torchvision/ops/poolers.py
datumbox: same here

torchvision/ops/poolers.py
datumbox: Leave this change in this PR, happy to review it in detail afterwards.

datumbox (commented):
Thanks for the PR, I've added a few comments. Let me know what you think:

test/test_ops.py
```diff
@@ -422,12 +458,15 @@ def test_boxes_shape(self):
 
 
 class TestMultiScaleRoIAlign:
+    def make_obj(self, fmap_names=["0"], output_size=(7, 7), sampling_ratio=2, **kwargs):
```
datumbox, line 461: We need to avoid any mutable argument in methods. Here is why:

```python
>>> def x(v, d=[]):
...     d.append(v)
...     print(d)
... 
>>> 
>>> x(1, [])  # that works as expected because we pass a new list
[1]
>>> x(1, [])  # same here
[1]
>>> x(1)  # This addition happen on the single default list
[1]
>>> x(1) # oops!
[1, 1]
>>> x(1)
[1, 1, 1]
```

torchvision/models/feature_extraction.py
```diff
@@ -294,7 +317,7 @@ def create_feature_extractor(
     return_nodes: Optional[Union[List[str], Dict[str, str]]] = None,
     train_return_nodes: Optional[Union[List[str], Dict[str, str]]] = None,
     eval_return_nodes: Optional[Union[List[str], Dict[str, str]]] = None,
-    tracer_kwargs: Dict = {},
+    tracer_kwargs: Optional[Dict] = None,
```
datumbox, line 320:
```suggestion
    tracer_kwargs: Optional[Dict[str, Any]] = None,
```

torchvision/ops/poolers.py
datumbox: Shall we make this private?

datumbox:
@alexander-soare could we check your thoughts?

torchvision/models/feature_extraction.py
```diff
@@ -172,8 +175,19 @@ def _warn_graph_differences(train_tracer: NodePathTracer, eval_tracer: NodePathT
     warnings.warn(msg + suggestion_msg)
 
 
+def _get_leaf_modules_for_ops():
+    members = inspect.getmembers(torchvision.ops)
+    result = []
+    for _, obj in members:
+        if inspect.isclass(obj) and issubclass(obj, torch.nn.Module):
+            result.append(obj)
+    return tuple(result)
```
datumbox, line 184: Non blocking: Due to the variable size of the result, it would be more natural to declare this a list. The mutability won't be an issue as you don't use it as default arg any more.

torchvision/ops/poolers.py
```diff
@@ -165,32 +295,6 @@ def __init__(
         self.canonical_scale = canonical_scale
         self.canonical_level = canonical_level
 
-    def setup_scales(
```
datumbox, line 168: This in theory is a BC breaking change because the method was public. I don't think anyone should be calling it directly but it's worth checking on Github and FBcode to see if there are any direct calls. A safer approach could also be to expose it publicly, call the private equivalent and throw a warning.

cc @NicolasHug @fmassa thoughts?

torchvision/ops/poolers.py
```diff
@@ -106,6 +107,135 @@ def _infer_scale(feature: Tensor, original_size: List[int]) -> float:
...
+
+    x_filtered = []
+    for k, v in x.items():
+        if k in featmap_names:
+            x_filtered.append(v)
+    num_levels = len(x_filtered)
+    rois = _convert_to_roi_format(boxes)
+
+    if scales is None:
+        scales, mapper = _setup_scales(x_filtered, image_shapes, canonical_scale, canonical_level)
```
datumbox, line 175: Neither `scales` nor `mapper` are passed by reference, so as far as I understand you end up calling `_setup_scales()` every time. I believe this needs to be refactored to return every value we end up modifying and assigning them to the original `self` parameters.

datumbox (commented):
I had a closer look and added few more comments on the main methods. Let me know what you think.

datumbox:
Thanks a lot @alexander-soare.

I agree that https://github.com/pytorch/vision/pull/4810#issuecomment-969359521 is worth resolving prior merging. Concerning the nit, I also agree but don't have either a better recommendation. @jdsgomes any thoughts on your side?

jdsgomes:
Thank you @datumbox and @alexander-soare for your comments.

> @datumbox sure! Looking really good @jdsgomes :)
> 
> Just my final outstanding thoughts (and I may be repeating myself on the above, but wanted to make sure to get your feedback on these):
> 
> * As I mentioned [above](https://github.com/pytorch/vision/pull/4810#issuecomment-969359521), I still think we should be wrapping the [obj](https://github.com/pytorch/vision/blob/0944b3259d5384722a5429f5a7a447a4683cc41b/test/test_ops.py#L461-L464) in the test harness in a TestModule. If the obj is not traceable the test will fail whether you make it a leaf module or not. @jdsgomes avoided that by making MultiScaleROIAlign traceable (by extracting the non-traceable part as a function), but if someone introduces another module that's not traceable, the tests will fail again.

This really depends on what is the desired behaviour. If I understood the problem correctly the only way that the current tests would fail is if someone changes an existing Module and makes it not traceable. In that case I think it might be ok to have the tests fail to signal that, and then developer can choose to make it traceable or actively change the test and wrap the obj. 

> * nit: If someone wants to use the `tracer_kwargs` argument, and they also happen to use a non-traceable function/module from `torchvision.ops` they will get a trace error. Then they will probably know how to manually add the culprit as either an `autowarp_function` or a `leaf_module`. If they are doing this programmatically, they will probably need to look into the [code](https://github.com/pytorch/vision/blob/0944b3259d5384722a5429f5a7a447a4683cc41b/torchvision/models/feature_extraction.py#L424-L431), and replicate it in their code. I can't think of any nicer alternative. Just alternatives like appending the defaults to the user input rather than letting the user override them unknowingly, or using an extra kwarg like `wrap_torchvision_ops=True`. I don't like any of them though haha. So this is just food for thought.

I think the option of appending is acceptable since we are already setting a default value in the case tracer_kwargs is None, and therefore we keep the behaviour consistent. Just wondering if there will be a case where the user would want to overwrite the default values?

torchvision/ops/poolers.py
jdsgomes: You are right they are set to None every time in the `MultiScaleRoIAlign` constructor so this would be always called. But in that case I suggest to change this to 
```python
scales, mapper = _setup_scales(x_filtered, image_shapes, canonical_scale, canonical_level)
```
instead of 
```python
if scales is None:
        scales, mapper = _setup_scales(x_filtered, image_shapes, canonical_scale, canonical_level)
else:
        mapper = None
```

and also remove `scales` from the arguments list.

But it is not clear to me why we would need to return the values and assigning them to the original `self` paramters.

torchvision/ops/poolers.py
NicolasHug: >  A safer approach could also be to expose it publicly, call the private equivalent and throw a warning.

I would go with that, it's safer and it's also the simplest option (we don't need to check fbcode or github). We can mark those with a ` # TODO: deprecate eventually` comment.

Similar thing could be done for https://github.com/pytorch/vision/pull/4951 as we just discussed.

datumbox: Agreed. As Nicolas pointed out in an offline, the methods `convert_to_roi_format` and `infer_scale` which were modified at #4951 need the same treatment. Aka keep a public method for the class with a call to the private one, a warning and a `TODO` comment.

alexander-soare:
@jdsgomes 

> If I understood the problem correctly the only way that the current tests would fail is if someone changes an existing Module and makes it not traceable. In that case I think it might be ok to have the tests fail to signal that, and then developer can choose to make it traceable or actively change the test and wrap the obj.

I think it should be acceptable that someone adds a module who's forward method is not traceable, because this PR is meant to guarantee that we **don't trace** through anything in `torchvision.ops`. So let's say someone makes `NewModule` whose forward method is **not** traceable. Then the tests should be designed such that they pass anyway (because the API should be designed such that `NewModule` is treated as a leaf). But right now the tests are actually designed to check that `NewModule` **is** traceable, which is not a requirement of #4777.

In saying that though, if we all agree that we **do want to make all modules in torchvision.ops** traceable then your tests reflect that requirement. Afaik that's not the case though right?

--- 

> Just wondering if there will be a case where the user would want to overwrite the default values?

Agreed, which is why I don't totally like my proposal :smile:.

torchvision/ops/poolers.py
jdsgomes: Please ignore this, because of https://github.com/pytorch/vision/pull/4810/files#r754140003 this comment no longer applies.

jdsgomes:
@alexander-soare thank you for your patience. I didn't get the point fully in the first place.
I think you are right, enforcing all ops to be traceable is not the point, so I have changed the tests with your suggestion to reflect that.

Regarding the nit, I am leaning to leave it as is, so if someone passes them as argument they wont be modified unknowingly, despite the problems you raised seems to be preferable to me.

torchvision/models/feature_extraction.py
```diff
@@ -172,8 +175,19 @@ def _warn_graph_differences(train_tracer: NodePathTracer, eval_tracer: NodePathT
     warnings.warn(msg + suggestion_msg)
 
 
+def _get_leaf_modules_for_ops():
```
datumbox, line 178: Shall we add the typing info here? Something like:
```suggestion
def _get_leaf_modules_for_ops() -> List[type]:
```

datumbox (approved):
LGTM, thanks a lot @jdsgomes. I just left an optional nit, other than that it's all great.

@alexander-soare Would you be able to confirm on your side that all the parts you wanted to see covered are in place? Thanks a lot for the help!

alexander-soare:
@datumbox you're welcome, it's a pleasure to give a little back! Looks good to me. Nice work @jdsgomes !

test/test_ops.py
```diff
@@ -7,12 +7,54 @@
...
 from common_utils import needs_cuda, cpu_and_gpu, assert_equal
 from PIL import Image
 from torch import nn, Tensor
 from torch.autograd import gradcheck
 from torch.nn.modules.utils import _pair
 from torchvision import models, ops
+from torchvision.models.feature_extraction import get_graph_node_names
+
+
+class RoIOpTesterModuleWrapper(nn.Module):
```
jdsgomes, line 20: @pmeier could I get your input here please. the reason I define all these  `*Wrapper` classes is that we need wrappers with variable numbers of `forward` arguments. Initially we considered using `*args` in the `forward` method but this is not traceable. Could you think of any other option for this case, or it looks ok as it it?

datumbox: @pmeier The bat sign is out. We need your wisdom :)

pmeier: I've talked with @jdsgomes offline and I don't think there is way to make this pythonic *and* appease torchscript. We probably need to go with the current design.

jdsgomes: After discussing offline with @pmeier we agreed that the best action is to merge the PR as there is not obvious scriptable alternative to this one.

Issue #3995, opened 2021-06-07
Improve the accuracy of Classification models by using SOTA recipes and primitives


datumbox:
## 🚀 Feature
Update the weights of all pre-trained models to improve their accuracy.

## Motivation

<h4 id="new-recipe-with-fixres">New Recipe + FixRes mitigations</h4>

```
torchrun --nproc_per_node=8 train.py --model $MODEL_NAME --batch-size 128 --lr 0.5 \
--lr-scheduler cosineannealinglr --lr-warmup-epochs 5 --lr-warmup-method linear \
--auto-augment ta_wide --epochs 600 --random-erase 0.1 --weight-decay 0.00002 \
--norm-weight-decay 0.0 --label-smoothing 0.1 --mixup-alpha 0.2 --cutmix-alpha 1.0 \
--train-crop-size 176 --model-ema --val-resize-size 232
```

Using a recipe which includes Warmup, Cosine Annealing, Label Smoothing, Mixup, Cutmix, Random Erasing, TrivialAugment, No BN weight decay, EMA and long training cycles and optional FixRes mitigations we are able to improve the `resnet50` accuracy by over 4.5 points. For more information on the training recipe, check [here](https://pytorch.org/blog/how-to-train-state-of-the-art-models-using-torchvision-latest-primitives/):
```
Old ResNet50:
Acc@1 76.130 Acc@5 92.862

New ResNet50:
Acc@1 80.674 Acc@5 95.166
```

Running other models through the same recipe, achieves the following improved accuracies:
```
ResNet101:
Acc@1 81.728 Acc@5 95.670

ResNet152:
Acc@1 82.042 Acc@5 95.926

ResNeXt50_32x4d:
Acc@1 81.116 Acc@5 95.478

ResNeXt101_32x8d:
Acc@1 82.834 Acc@5 96.228

MobileNetV3 Large:
Acc@1 74.938 Acc@5 92.496

Wide ResNet50 2:
Acc@1 81.602 Acc@5 95.758 (@prabhat00155)

Wide ResNet101 2:
Acc@1 82.492 Acc@5 96.110 (@prabhat00155)

regnet_x_400mf:
Acc@1 74.864 Acc@5 92.322 (@kazhang)

regnet_x_800mf:
Acc@1 77.522 Acc@5 93.826 (@kazhang)

regnet_x_1_6gf:
Acc@1 79.668 Acc@5 94.922 (@kazhang)
```

[...]

PR #4811, opened 2021-11-01
Add new weights for ResNet101 and ResNet151

datumbox:
Fixes partially #3995

```
torchrun --nproc_per_node=1 train.py --test-only --weights ImageNet1K_RefV2 --model resnet101 -b 1
Acc@1 81.728 Acc@5 95.670

torchrun --nproc_per_node=1 train.py --test-only --weights ImageNet1K_RefV2 --model resnet152 -b 1
Acc@1 82.042 Acc@5 95.926
```

cc @datumbox


torchvision/prototype/models/resnet.py
```diff
@@ -127,6 +137,16 @@ class ResNet152Weights(Weights):
             "acc@5": 94.046,
         },
     )
+    ImageNet1K_RefV2 = WeightEntry(
+        url="https://download.pytorch.org/models/resnet152-089c0848.pth",
+        transforms=partial(ImageNetEval, crop_size=224, resize_size=232),
```
jdsgomes, line 142: the resize_size was not used in the original training?

jdsgomes (approved):

torchvision/prototype/models/resnet.py
datumbox: Nope, that's a new optimization only available on the new weights. :-)

PR #4812, opened 2021-11-01
Expose misc ops at package level

jdsgomes:
cc @datumbox  @NicolasHug


datumbox (approved):
LGTM, thanks!

NicolasHug (commented):
Thanks for the PR @jdsgomes ,

If we're going to expose these in the `torchvision.ops` namespace, I think we should also expose them in the docs in `ops.rst`. These ops seem to lack docstring though, so maybe we should write some

oke-aditya:
We were going to expose using `torchvision.nn` ? #4333 Or is it decided with `ops.misc.py` ?

datumbox:
@oke-aditya The reason we are not doing it at the moment is because there is no consensus for renaming/moving stuff to `nn`. This PR is just addressing the fact that the classes are public but not exposed directly via `ops`. Though the RFC you send is still valid, I would recommend not to link the two so that we don't block fixing outstanding issues.

oke-aditya:
Can understand it :) thanks for the clarification.

jdsgomes:
> Thanks for the PR @jdsgomes ,
> 
> If we're going to expose these in the `torchvision.ops` namespace, I think we should also expose them in the docs in `ops.rst`. These ops seem to lack docstring though, so maybe we should write some

Do you suggest adding the docstrings in this PR as well or that is out of scope and we should create a ticket for it?

torchvision/ops/misc.py
```diff
@@ -99,12 +102,29 @@ def __repr__(self) -> str:
...
+
+    """
+
     def __init__(
         self,
         in_channels: int,
         out_channels: int,
         kernel_size: int = 3,
-        stride: int = 1,
+        stride: _size_2_t = 1,
```
datumbox, line 128: If you define stride as `_size_2_t` then probably have kernel_size, padding and dilation like [here](https://github.com/pytorch/pytorch/blob/master/torch/nn/modules/conv.py#L419-L422). If you decide to switch back to integers, we will need to also modify the documentation above.

datumbox (commented):
Overall looks good. Just one comment on the stride size type. Let me know what you think.

Issue #4814, opened 2021-11-01
Source code not found on torchvision documentation page.


nateanl:
### 📚 The doc issue

When I click the ``[Source]`` link of the [ImageNet](https://pytorch.org/vision/stable/datasets.html#imagenet) dataset, the webpage shows ``404 File not found``.

![image](https://user-images.githubusercontent.com/8653221/139680634-5f92db6d-e2ad-4aa9-b7e8-22edaca2246f.png)
I checked the link for other methods, such as ``torchvision.io``. It seems all links are not valid.

cc @NicolasHug @prabhat00155 @fmassa

NicolasHug:
Thanks for the report @nateanl 

 Comparing https://github.com/pytorch/vision/tree/gh-pages/0.11/_modules and https://github.com/pytorch/vision/tree/gh-pages/0.10/_modules, it looks like an entire folder is missing

@mattip Do you know what could have gone wrong here?

vadimkantorov:
Related: https://github.com/pytorch/pytorch/issues/67237

mattip:
Not sure. It is only in the `0.11` directory, in `main` (run nightly) the subdirectory exists. I will make a PR to restore them, although I really don't understand how this happened.

mattip:
The problem is [this `torchvision/` directive](https://github.com/pytorch/vision/blob/f1e783e9a29a27507735a723a06ee205c1cf5c32/.gitignore#L7) in the .gitignore file on the gh-pages branch.

mattip:
When I try to rebuild the 0.11.0 docs, I get this sphinx error and docs are not created. Is that known?
```
Warning, treated as error:
/home/matti/miniconda3/envs/throwaway/lib/python3.7/site-packages/torchvision/ops/deform_conv.py 
    :docstring of torchvision.ops.deform_conv.DeformConv2d.forward:7:
    Field list ends without a blank line; unexpected unindent.
make: *** [Makefile:41: html] Error 2

```

mattip:
> When I try to rebuild the 0.11.0 docs, ...

Nvrmind, I had cruft from a levtover build in my `docs/source` directory

nateanl:
Resolved by #4825. Thanks @mattip for working on this!

PR #4820, opened 2021-11-01
use toplevel directories

mattip:
Fixes the cause for #4814

The internal torchvision directory was being ignored, and not committed to the branch.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,8 +3,8 @@
 *.vscode
 
 # Additional Ignore to exclude torchvision code and docs.
-docs/
-torchvision/
+/docs/
+/torchvision/
 
 # Byte-compiled / optimized / DLL files
 __pycache__/


NicolasHug:
Thanks a lot @mattip

NicolasHug (approved):

.gitignore
```diff
@@ -3,8 +3,8 @@
 *.vscode
 
 # Additional Ignore to exclude torchvision code and docs.
-docs/
-torchvision/
+/docs/
+/torchvision/
```
oke-aditya, line 7: Oooh my bad. I think I had added this in a PR.
Apologies :pray:

oke-aditya: This occurred when I was migrating docs from master to main branch. I thought the generated folder were unnecessary. But it doesn't seem to be the case. Sorry once again :(

NicolasHug: No problem @oke-aditya , if anything, it's our responsibility as reviewer to make sure nothing breaks, not yours.

This is the main reason why we are quite defensive when it comes new features, or seemingly harmless changes: even the smallest change can boomerang at us in ways we can't really imagine at the time of the PR. So sometimes, we push back on including something - it can be a bit frustrating for the contributor, but it can help avoiding a lot of headache in the long run.

oke-aditya: I will keep this in mind.

PR #4822, opened 2021-11-01
DOC: fix intersphinx links

mattip:
This is a tiny cleanup. The config file had old links. The old links have forwards to the new ones, so nothing was broken.


prabhat00155 (approved):
Thanks @mattip for the PR.

PR #4824, opened 2021-11-02
[references/classification] Adding gradient clipping

sallysyw:
Gradient Clipping can be useful in training - it is a technique to prevent exploding gradients in very deep networks.
Pytorch has a related API [`torch.nn.utils.clip_grad_norm_`](https://pytorch.org/docs/stable/generated/torch.nn.utils.clip_grad_norm_.html) which can be utilized here.

By applying gradient clipping to training, we significantly improved the training accuracy of our modes. For example:

| | Job ID | Model    	| Epochs 	| Nodes 	| Batch Size Per GPU 	| Global Batch Size 	| Acc@1  	| Acc@5  	|
|---------| --------- |----------	|--------	|-------	|-----------------		|--------------------	|--------	|--------	|
|before gradient clipping| 2420 | vit_b_16 	| 300    	| 2     	| 256             	| 4096                         	|68.148 	| 87.648 	|
|after gradient clipping| 2512 | vit_b_16 	| 300    	| 2     	| 256              	| 4096                       	| 71.876 	| 89.784 	|

cc @kazhang @datumbox

### DIFF
--- a/references/classification/utils.py
+++ b/references/classification/utils.py
@@ -411,20 +411,9 @@ def reduce_across_processes(val):
     return t
 
 
-try:
-    import apex
-
-    apex_available = True
-except ImportError:
-    apex_available = False
-
-
 def master_params(optimizer):
     """Generator to iterate over all parameters in the optimizer param_groups."""
 
-    if apex_available:
-        yield from apex.amp.master_params(optimizer)
-    else:
-        for group in optimizer.param_groups:
-            for p in group["params"]:
-                yield p
+    for group in optimizer.param_groups:
+        for p in group["params"]:
+            yield p
--- a/references/classification/utils.py
+++ b/references/classification/utils.py
@@ -411,7 +411,7 @@ def reduce_across_processes(val):
     return t
 
 
-def master_params(optimizer):
+def get_optimizer_params(optimizer):
     """Generator to iterate over all parameters in the optimizer param_groups."""
 
     for group in optimizer.param_groups:
--- a/references/classification/utils.py
+++ b/references/classification/utils.py
@@ -409,3 +409,21 @@ def reduce_across_processes(val):
     dist.barrier()
     dist.all_reduce(t)
     return t
+
+
+try:
+    import apex
+
+    apex_available = True
+except ImportError:
+    apex_available = False
+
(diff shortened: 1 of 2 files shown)


references/classification/utils.py
```diff
@@ -409,3 +409,21 @@ def reduce_across_processes(val):
     dist.barrier()
     dist.all_reduce(t)
     return t
+
+
+try:
```
sallysyw, line 414: This section is borrowed from ClassyVision:
https://github.com/facebookresearch/ClassyVision/blob/main/classy_vision/generic/util.py#:~:text=parameters%20in%20the-,optimizer,-param_groups.

references/classification/utils.py
```diff
@@ -409,3 +409,22 @@ def reduce_across_processes(val):
...
+    apex_available = True
+except ImportError:
+    apex_available = False
+
+
+def master_params(optimizer):
+    """Generator to iterate over all parameters in the optimizer param_groups."""
+
+    if apex_available:
+        yield from apex.amp.master_params(optimizer)
```
kazhang, line 426: The implementation is exactly the same as below
https://github.com/NVIDIA/apex/blob/master/apex/amp/_amp_state.py#L67-L69.
In PyTorch domain lib, let's try to use PyTorch native as much as possible.

references/classification/utils.py
datumbox: Indeed we switched to `torch.cuda.amp` recently. See #4547

references/classification/utils.py
sallysyw: Good catch! will remove this part

references/classification/utils.py
```diff
@@ -409,3 +409,11 @@ def reduce_across_processes(val):
     dist.barrier()
     dist.all_reduce(t)
     return t
+
+
+def master_params(optimizer):
```
kazhang, line 414: naming nit: maybe `optimizer_params` or `get_optimizer_params`? I don't know why there is a "master"

kazhang (approved):
LGTM, thanks!

datumbox (approved):
LGTM, thanks!

PR #4830, opened 2021-11-02
Update training references from legacy models

datumbox:
Adding references where possible on all current weights.


fmassa (approved):
Thanks!

Issue #4679, opened 2021-10-20
Move new model builders from Prototype to main TorchVision 


datumbox:
### 🚀 The feature

Once all model builders are updated to support multiple pretrained weights, we should replace the old builders with the new ones. This involves the following steps:
1. Replace the code on `torchvision.models` with the updates of `torchvision.prototype.models`.
2. Delete `torchvision.prototype.models` and their prototype tests.
3. Update the existing model tests (`pretrained` => `weights`) and ensure all work without modifying the expected files.
4. Update the reference scripts to work with the new API.
5. Run all models using the reference scripts and validate that the weights can still be loaded and that the accuracies remain unaffected.

### Motivation, pitch

-

Issue #4652, opened 2021-10-19
Multi pretrained weights: Cleanups and Refactoring


jdsgomes:
Candidates:
- [x] Reduce code duplication on model builders by refactoring. For example potentially push the weigh loading mechanism in the private builder methods currently inherited from `torchvision.models`. Also we could provide a `pretrained` deprecation method.
- [x] Review recipe links and ensure they are accurate for all Community Contributed models. Check if possible for some of them to remap their training process to our current ref scripts.
- [x] Decide how quantized models will link to their unquantized versions.
- [x] Naming conventions:
  - The name of the `Weights` base `Enum` class and the naming convention of the inheriting classes. Also make sure that all classes are aligned with the model builder names. How are we going to handle `_` in names (for example on inception_v3 and the resnext models)?
  - The name of the `WeightEntry` data class.
  - The name of `Weights.verify()` method (because it does more than just verification).
  - The names of the enum values; also their styles (all capital?). Example: `ImageNet1K_RefV1`
  - Currently the `_RefV1` of one model doesn't have to be `_RefV1` of another model. For example [ResNet](https://github.com/pytorch/vision/blob/18cf5ab1c1307e9d0e1b1e4d26ce9c45e04f8376/torchvision/prototype/models/resnet.py#L86-L91) vs [RegNets](https://github.com/pytorch/vision/blob/18cf5ab1c1307e9d0e1b1e4d26ce9c45e04f8376/torchvision/prototype/models/regnet.py#L110-L115) use different recipes with the same enum value. In other cases, we skip `_RefV1` and go directly to `_RefV2`, for example the [WideResNet50 weights](https://github.com/pytorch/vision/blob/18cf5ab1c1307e9d0e1b1e4d26ce9c45e04f8376/torchvision/prototype/models/resnet.py#L211-L222). Is this a problem?
- [x] Decide the signature of `models._api.get_weight()` and make it public. Ensure it's torchhub-friendly.

cc @datumbox @pmeier @bjuncek @NicolasHug

jdsgomes:
Just opening the discussion here regarding the naming for `Weights` and `WeightEntry`. As This classes store metadata about weights and how they were generated (recipe + transforms) rather than the actual weights so I think we could try to find a better name.

Having said that I'm struggling to find a much better name, but want to leave some suggestions for discussion:
- PretrainedModelEnum and PretrainedModel
- PretrainedModel and PretrainedModelEntry
- Weights and WeightsInfo

[...]

PR #4837, opened 2021-11-02
Update Reference scripts to support the prototype models

### LINKED ISSUE #4671: Multi-pretrained weights: Add initial API and basic implementation
### 🚀 The feature

Part of #4611

Add the initial API mechanism for the Multi-pretrained weights: 
- The `WeightEntry` and `Weights` classes.
- The necessary preset transforms.
- The meta information needed for the models.

Moreover provide basic implementations for the following models:
- ResNet50
- FasterRCNN + ResNet50 
- Quantized ResNet50

datumbox:
Fixes #4671

This PR adds a similar mechanism as in `classification` for `segmentation` and `video`. The target is to enable us to test the new model weights API (+ presets) and confirm it returns the same results as the old one. The co-existence of `--pretrained` and `--weights` is temporary and allows us to test that all models we introduce produce the expected results.

The approach is not perfect as it exposes the `prototype` stuff in the example reference scripts but the alternative would be to duplicate the reference scripts or keep a separate branch with their modifications which makes the work cumbersome. These will be cleaned up prior to adopting the new API, see #4652 and #4679.  

cc @datumbox @bjuncek

### DIFF
--- a/references/segmentation/train.py
+++ b/references/segmentation/train.py
@@ -11,6 +11,12 @@ from coco_utils import get_coco
 from torch import nn
 
 
+try:
+    from torchvision.prototype import models as PM
+except ImportError:
+    PM = None
+
+
 def get_dataset(dir_path, name, image_set, transform):
     def sbd(*args, **kwargs):
         return torchvision.datasets.SBDataset(*args, mode="segmentation", **kwargs)
@@ -26,11 +32,15 @@ def get_dataset(dir_path, name, image_set, transform):
     return ds, num_classes
 
 
-def get_transform(train):
-    base_size = 520
-    crop_size = 480
-
-    return presets.SegmentationPresetTrain(base_size, crop_size) if train else presets.SegmentationPresetEval(base_size)
+def get_transform(train, args):
+    if train:
+        return presets.SegmentationPresetTrain(base_size=520, crop_size=480)
+    elif not args.weights:
+        return presets.SegmentationPresetEval(base_size=520)
+    else:
+        fn = PM.segmentation.__dict__[args.model]
+        weights = PM._api.get_weight(fn, args.weights)
+        return weights.transforms()
 
 
 def criterion(inputs, target):
@@ -90,8 +100,8 @@ def main(args):
 
     device = torch.device(args.device)
 
-    dataset, num_classes = get_dataset(args.data_path, args.dataset, "train", get_transform(train=True))
-    dataset_test, _ = get_dataset(args.data_path, args.dataset, "val", get_transform(train=False))
+    dataset, num_classes = get_dataset(args.data_path, args.dataset, "train", get_transform(True, args))
+    dataset_test, _ = get_dataset(args.data_path, args.dataset, "val", get_transform(False, args))
 
     if args.distributed:
         train_sampler = torch.utils.data.distributed.DistributedSampler(dataset)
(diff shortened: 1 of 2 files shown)


references/segmentation/train.py
```diff
@@ -26,11 +32,15 @@ def sbd(*args, **kwargs):
...
-    return presets.SegmentationPresetTrain(base_size, crop_size) if train else presets.SegmentationPresetEval(base_size)
+def get_transform(train, args):
+    if train:
+        return presets.SegmentationPresetTrain(base_size=520, crop_size=480)
+    elif not args.weights:
+        return presets.SegmentationPresetEval(base_size=520)
+    else:
+        fn = PM.segmentation.__dict__[args.model]
+        weights = PM._api.get_weight(fn, args.weights)
+        return weights.transforms()
```
datumbox, line 43: If we are in train mode, we always initialize the SegmentationPresetTrain. For validation if the weights are not defined (aka not a prototype model) then use the old preprocessing method for evaluation. Else use the one attached to the weights.

references/segmentation/train.py
```diff
@@ -113,9 +123,18 @@ def main(args):
         dataset_test, batch_size=1, sampler=test_sampler, num_workers=args.workers, collate_fn=utils.collate_fn
     )
 
-    model = torchvision.models.segmentation.__dict__[args.model](
-        num_classes=num_classes, aux_loss=args.aux_loss, pretrained=args.pretrained
-    )
+    if not args.weights:
```
datumbox, line 126: If the weights are not defined, we use the standard way. Else it's a prototype run which means we will use the prototype model mechanism.

datumbox (commented):
Some clarifications below:

references/video_classification/train.py
```diff
@@ -149,7 +155,12 @@ def main(args):
     print("Loading validation data")
     cache_path = _get_cache_path(valdir)
 
-    transform_test = presets.VideoClassificationPresetEval((128, 171), (112, 112))
+    if not args.weights:
+        transform_test = presets.VideoClassificationPresetEval((128, 171), (112, 112))
+    else:
+        fn = PM.video.__dict__[args.model]
+        weights = PM._api.get_weight(fn, args.weights)
```
fmassa, line 162: nit: using a private API here. We probably don't want to advertise private APIs in the references

references/video_classification/train.py
```diff
@@ -149,7 +155,12 @@ def main(args):
     print("Loading validation data")
     cache_path = _get_cache_path(valdir)
 
-    transform_test = presets.VideoClassificationPresetEval((128, 171), (112, 112))
+    if not args.weights:
```
fmassa, line 158: nit: `pretrained` and `weights` are overlapping and can be confusing. This ideally should be cleaned up in the future

references/video_classification/train.py
```diff
@@ -149,7 +155,12 @@ def main(args):
...
+        transform_test = presets.VideoClassificationPresetEval((128, 171), (112, 112))
+    else:
+        fn = PM.video.__dict__[args.model]
```
fmassa, line 161: are we providing some sort of registration API to get the models without having to resort to `__dict__` manipulations?

fmassa (approved):
Approving to unblock

references/video_classification/train.py
datumbox: This is temporary and behind a feature switch (`--weights`). It's already recorded to clean up at #4652

references/video_classification/train.py
datumbox: That's exactly the plan. `--pretrained` will go away and `--weights` is going to be the right parameter. Right now we support both temporarily so that we can switch between the two completely different APIs. The `--weights` acts as a feature switch here.

references/video_classification/train.py
datumbox: Yeap, that's the plan. There will be a proper registration mechanism, possibly something similar to what was discussed [here](https://github.com/datumbox/dapi-model-versioning). There are still pending discussions with other domains, so I didn't want to adopt something before those discussions take place.

PR #4838, opened 2021-11-02
[WIP] UCF101 prototype with utilities for video loading

bjuncek:
A simple `pyav` based set of utilities with a POC implementation for UCF101 dataset


torchvision/prototype/datasets/_builtin/ucf101.py
```diff
@@ -0,0 +1,94 @@
...
+        self,
+        data: Tuple[np.ndarray, int],
+        *,
+        decoder: Optional[Callable[[io.IOBase], Dict[str, Any]]],
+    ) -> Dict[str, Any]:
+        annotations_d, file_d = data
+
+        label = annotations_d[1]
+        _path, file_handle = file_d
+        return {"path": _path, "file": file_handle, "target": label}
```
pmeier, line 65: We also need to handle the decoder here. By default, `decoder` will be `None`, but the user is free to pass something.

torchvision/prototype/datasets/_builtin/ucf101.py
```diff
@@ -0,0 +1,94 @@
...
+        decoder: Optional[Callable[[io.IOBase], torch.Tensor]],
+    ) -> IterDataPipe[Dict[str, Any]]:
+
+        annotations = resource_dps[0]
+        files = resource_dps[1]
+
+        annotations_dp = ZipArchiveReader(annotations)
+        annotations_dp = Filter(annotations_dp,
+                                self._filtername,
+                                fn_kwargs=dict(tgt=f"{config.split}list0{config.fold}.txt"))
```
pmeier, line 87:
```suggestion
        annotations_dp = Filter(annotations_dp, path_comparator("name", f"{config.split}list0{config.fold}.txt"))
```

You can import `path_comparator` from `torchvision.prototype.datasets.utils._internal`

torchvision/prototype/datasets/_builtin/ucf101.py
```diff
@@ -0,0 +1,94 @@
...
+        annotations_dp = CSVParser(annotations_dp, delimiter=" ")
+        # COMMENT OUT FOR TESTING
+        annotations_dp = Shuffler(annotations_dp, buffer_size=INFINITE_BUFFER_SIZE)
+
+        files_dp = RarArchiveReader(files)
+        dp = KeyZipper(annotations_dp, files_dp, self._getname, self._getname)
```
pmeier, line 93: If you only pass a single key function to `KeyZipper`, it will be used for both datapipes. 

```suggestion
        dp = KeyZipper(annotations_dp, files_dp, path_accessor("name"))
```

You can import `path_accessor` from `torchvision.prototype.datasets.utils._internal`.

torchvision/prototype/datasets/_builtin/ucf101.py
```diff
@@ -0,0 +1,94 @@
...
+            valid_options={'split': ["train", "test"], 'fold': ["1", "2", "3"]},
+            # categories=HERE / "ucf101.categories",
+            homepage="https://www.crcv.ucf.edu/data/UCF101.php",
+        )
+
+    def resources(self, config: DatasetConfig) -> List[OnlineResource]:
+        return [
+            HttpResource(
+                "https://www.crcv.ucf.edu/data/UCF101/UCF101TrainTestSplits-RecognitionTask.zip",
+                sha256="",
```
pmeier, line 47:
```suggestion
                sha256="5c0d1a53b8ed364a2ac830a73f405e51bece7d98ce1254fd19ed4a36b224bd27",
```

torchvision/prototype/datasets/_builtin/ucf101.py
```diff
@@ -0,0 +1,94 @@
...
+            ),
+            HttpResource(
+                "https://www.crcv.ucf.edu/data/UCF101/UCF101.rar",
+                sha256="",
```
pmeier, line 51:
```suggestion
                sha256="ca8dfadb4c891cb11316f94d52b6b0ac2a11994e67a0cae227180cd160bd8e55",
```

torchvision/prototype/datasets/_builtin/ucf101.py
```diff
@@ -0,0 +1,94 @@
...
+            )
+        ]
+
+    def _collate_and_decode(
+        self,
+        data: Tuple[np.ndarray, int],
```
pmeier, line 57: This doesn't seem right. Is that an artifact from copy pasting from another dataset?

torchvision/prototype/datasets/_builtin/ucf101.py
```diff
@@ -0,0 +1,94 @@
...
+
+        annotations = resource_dps[0]
+        files = resource_dps[1]
+
+        annotations_dp = ZipArchiveReader(annotations)
+        annotations_dp = Filter(annotations_dp,
+                                self._filtername,
+                                fn_kwargs=dict(tgt=f"{config.split}list0{config.fold}.txt"))
+        annotations_dp = CSVParser(annotations_dp, delimiter=" ")
+        # COMMENT OUT FOR TESTING
```
pmeier, line 89: True, but should be removed before merge.

torchvision/prototype/datasets/utils/_dataset.py
```diff
@@ -28,6 +28,7 @@
 class DatasetType(enum.Enum):
     RAW = enum.auto()
     IMAGE = enum.auto()
+    VIDEO = enum.auto()
```
pmeier, line 31: We also need to update 

https://github.com/pytorch/vision/blob/0817f7f5983b05fdc34e06476ca270e2168c57b8/torchvision/prototype/datasets/_api.py#L54-L57

torchvision/prototype/datasets/decoder.py
```diff
@@ -13,3 +16,17 @@ def raw(buffer: io.IOBase) -> torch.Tensor:
 
 def pil(buffer: io.IOBase, mode: str = "RGB") -> torch.Tensor:
     return pil_to_tensor(PIL.Image.open(buffer).convert(mode.upper()))
+
+
+def av_kf(buffer: io.IOBase, **read_video_kwargs: Any) -> Dict[str, Any]:
+    with unittest.mock.patch("torchvision.io.video.os.path.exists", return_value=True):
```
pmeier, line 22: I don't think we need that anymore. I only added it in #4541 because I used `torchvision.io.video.read_video` and it enforces that the input is a file on disk although it can very well work with file handles.

torchvision/prototype/datasets/decoder.py
```diff
@@ -13,3 +16,17 @@ def raw(buffer: io.IOBase) -> torch.Tensor:
 
 def pil(buffer: io.IOBase, mode: str = "RGB") -> torch.Tensor:
     return pil_to_tensor(PIL.Image.open(buffer).convert(mode.upper()))
+
+
+def av_kf(buffer: io.IOBase, **read_video_kwargs: Any) -> Dict[str, Any]:
```
pmeier, line 21: I feel like a decoder in this file should follow the "one file handle, one decoded output" paradigm. That means, I would prefer to have an `av` function here that decodes the whole video. Given that you also don't use this function in the video utilities that you wrote, I think that should not be a problem.

torchvision/prototype/datasets/video_utils.py
```diff
@@ -0,0 +1,192 @@
...
+    def __init__(self, video_dp: IterDataPipe[Dict[str, Any]], transform=None) -> None:
+        """TorchData Iterdatapype that takes in video datapipe
+        and yields all keyframes in a video
+
+        Args:
+            video_dp (IterDataPipe[Dict[str, Any]]): Video dataset IterDataPipe
+            transform ([type], optional): Transform to be applied to every frame. Defaults to None.
+        """
+        self.datapipe = video_dp
+        self.transform = transform
```
pmeier, line 19: That is an anti-pattern we want to avoid. Transformations can applied in a separate `.map` call.

torchvision/prototype/datasets/video_utils.py
```diff
@@ -0,0 +1,192 @@
...
+
+    def __iter__(self) -> Iterator[Dict[str, Any]]:
+        for video_d in self.datapipe:
+            buffer = video_d["file"]
```
pmeier, line 23: The key should probably be a parameter to the datapipe, because we probably won't enforce a fixed key so we can hardcode it here.
```suggestion
            buffer = video_d.pop("file")
```

torchvision/prototype/datasets/video_utils.py
```diff
@@ -0,0 +1,192 @@
...
+            with av.open(buffer, metadata_errors="ignore") as container:
+                stream = container.streams.video[0]
+                stream.codec_context.skip_frame = 'NONKEY'
+                for frame in container.decode(stream):
+                    img = frame.to_image()
+                    if self.transform is not None:
+                        img = self.transform(img)
+                    yield {
```
pmeier, line 31: We should make sure that we include all the information that we had previously and only replace `"file"` with new data. The easiest way to do that is to `pop` the key and then use the old data as base for the new dictionary.

```suggestion
                    yield dict(video_d, frame=img, ...)
```

pmeier (commented):
Thanks a lot @bjuncek. I have some comments inline about the general infrastructure. I can't really comment on the validity of the video utility datapipes that you added, because I have to little experience with videos. I'll leave that up to other reviewers.

main.py
```diff
@@ -0,0 +1,32 @@
+from torchvision.prototype import datasets
```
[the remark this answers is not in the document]
bjuncek: Yes, exactly - this is to play around with and modify different params to mock the behaviour

torchvision/prototype/datasets/decoder.py
bjuncek: actually, after thinking about it, all decoding is atm happening in the `video_utils` file so I'll revert this file since we don't use it

torchvision/prototype/datasets/video_utils.py
bjuncek: will remove it - force of habit I suppose :)

torchvision/prototype/datasets/decoder.py
pmeier: Fine by me. We can always add it later if the need arises.

torchvision/prototype/datasets/_builtin/ucf101.py
fmassa: I would really want our datasets to be deterministic outside of a DataLoader though. Making it stochastic will make it much harder to debug.

Maybe what we should do instead is have a new `DataLoaderOnlyShuffler` which is a no-op in general, and inside the dataloader it can be activated if `shuffle` is true.

Thoughts?

cc @VitalyFedyunin @ejguan

torchvision/prototype/datasets/_builtin/ucf101.py
bjuncek: ah, yes; mea culpa - should be fixed

torchvision/prototype/datasets/_builtin/ucf101.py
```diff
@@ -0,0 +1,89 @@
...
+    Dataset,
+    DatasetConfig,
+    DatasetInfo,
+    HttpResource,
+    OnlineResource,
+    DatasetType,
+)
+
+
+class ucf101(Dataset):
```
pmeier, line 27:
```suggestion
class UCF101(Dataset):
```

torchvision/prototype/datasets/_builtin/ucf101.py
```diff
@@ -0,0 +1,89 @@
...
+    """This is a base datapipe that returns a file handler of the video.
+    What we want to do is implement either several decoder options or additional
+    datapipe extensions to make this work.
+    """
```
pmeier, line 31: True, but probably shouldn't be here.

torchvision/prototype/datasets/_builtin/ucf101.py
pmeier: See pytorch/pytorch#65756.

torchvision/prototype/datasets/utils/_internal.py
```diff
@@ -176,6 +177,38 @@ def __iter__(self) -> Iterator[Tuple[str, io.IOBase]]:
             yield path, decompressor(file)
 
 
+class RarArchiveReader(IterDataPipe[Tuple[str, io.BufferedIOBase]]):
```
pmeier, line 180: This should be landed in `torchdata`. See pytorch/data#41.

pmeier (approved):
Some minor comments inline. I didn't review the video datapipes and leaving that to someone with domain knowledge. Otherwise LGTM when CI is green! Thanks Bruno!

pmeier (changes requested):
Just changing review status so that we don't forget to remove the `main.py` file before merge.

torchvision/prototype/datasets/utils/_internal.py
bjuncek: it's not for now. should I wait until it is?

bjuncek:
Ok, so I've tried doing a pass on this, trying to fix the decoder inconsistency we've been talking about offline. 

I don't understand datapipes well enough to understand why pop from a dict would fail or why I'd need to annotate variables in a datapipe. Everything since `8f57ee6` has nothing to do with the functionality of the UCF101 dataset and can be reverted if you want to do a clean pass and then merge into this.

torchvision/prototype/datasets/utils/_internal.py
pmeier: Nope, let's move forward here. If the PR in `torchdata` is merged, we can remove it here.

torchvision/prototype/datasets/video_utils.py
```diff
@@ -0,0 +1,178 @@
...
+                        "clip": vframes,
+                        "pts": clip_pts,
+                        "range": (start_pts, end_pts),
+                        "video_meta": {
+                            "time_base": float(stream.time_base),
+                            "guessed_fps": float(stream.guessed_rate),
+                        },
+                        "path": video_d["path"],
+                        "target": video_d["target"],
+                    }
```
fmassa, line 178: Why not just do the following:
- sample `m` start positions
- for every start position, read `k` frames
- yield the `k` frames at once, `m` times

torchvision/prototype/datasets/video_utils.py
```diff
@@ -0,0 +1,178 @@
+import random
+from typing import Any, Dict, Iterator
+
+import av
+import numpy as np
+import torch
+from torchdata.datapipes.iter import IterDataPipe
+from torchvision.io import video, _video_opt
```
fmassa, line 8: I'm not sure if I would use `_video_opt` in here.

pmeier:
@bjuncek I've cleaned up the PR with recent changes and also removed all the decoder changes. We will handle this separate from this PR following #5075.

torchvision/prototype/datasets/video_utils.py
bjuncek: Unless I'm missing something, this is exactly what I do:
1. sample starting positions (line 132)
2. for every start position (line 134) read `k` frames (line 140)
3. yield the frames as a sample (line 168)

Are you suggesting to take the yield outside of the loop? If so, is there any benefit to this?

torchvision/prototype/datasets/video_utils.py
bjuncek: Sure.
Any particular reason why not?

Issue #3608, opened 2021-03-25
COCO dataset to also return image_id


vadimkantorov:
So that copies like https://github.com/facebookresearch/detr/blob/master/datasets/coco.py are not needed

In general, I found it useful to return origina data ids, in case the user wants to load some extra information. Part of the problem is the Tuple return type which means any changes are not BC.

Maybe `return_dict` constructor arg could be introduced which would be more extensible, just like DETR did

PR #4845, opened 2021-11-03
Add Kitti and Sintel datasets for optical flow

NicolasHug:
This PR Adds the Kitti and Sintel datasets for optical flow.

Here are a few design decisions, following our initial discussion (in this PR):

- Unlike Sintel, Kitti has a built-in ``valid`` mask indicating which flow values are returned. As a resut Sintel returns ``img1, img2, flow`` while Kitti returns ``img1, img2, flow, valid``.

- For both these datasets, the targets aren't known if split="test" so we return ``img1, img2, None`` and  ``img1, img2, None, None`` in these cases.

- For both, ``flow`` is a numpy array of shape (2, H, W) and valid is a boolean numpy array of shape (H, W). The images are PIL images.

- The transforms expect to receive ``img1, img2, flow, valid``, even for Sintel, and even in test mode.

- [X] add tests
- [X] write docs


torchvision/datasets/_optical_flow.py
```diff
@@ -0,0 +1,149 @@
...
+
+    @abstractmethod
+    def _read_flow(self, file_name):
+        # Return the flow or a tuple (flow, valid) for datasets where the valid mask is built-in
+        pass
+
+    def __getitem__(self, index):
+        # Some datasets like Kitti have a built-in valid mask, indicating which flow values are valid
+        # For those we return (img1, img2, flow, valid), and for the rest we return (img1, img2, flow),
+        # and it's up to whatever consumes the dataset to decide what `valid` should be.
```
NicolasHug, line 40: What do we think of this? Essentially this means that one will do

```py
for e in KittiFlow():
    img1, img2, flow, valid = e

for e in Sintel():
    img1, img2, flow = e
    valid = 2_norm_of_flow_here < SOME_BIG_THRESHOLD  # Typically 1000

# See https://github.com/NicolasHug/vision/blob/6bd34ed5dfbf2710a4865e981db0764ba535007c/references/raft/train.py#L208-L212
```

I'm a bit torn here. One one hand it's not convenient that different datasets have a different return size, but on the other hand we don't want to decide what `valid` should be on behalf of the user when `valid` isn't built in the dataset.

Maybe one option would be to always return `img1, img2, flow, valid` and to set `valid` to None. But it would be a bit awkward for Sintel (and for other future datasets like FlyingChairs and FlyingThings3D) to always have a return value that's always None.

torchvision/datasets/_optical_flow.py
```diff
@@ -0,0 +1,149 @@
...
+        img2 = self._read_img(self._image_list[index][1])
+        flow = self._read_flow(self._flow_list[index]) if self._flow_list else None
+
+        if isinstance(flow, tuple):
+            flow, valid = flow
+        else:
+            valid = None
+
+        if self.transforms is not None:
+            img1, img2, flow, valid = self.transforms(img1, img2, flow, valid)
```
NicolasHug, line 52: We're forcing all transforms to accept a `flow` and a `valid` parameter, even if those aren't used. Typically for Sintel where `valid` is always None, `valid` would always be a passthough. An example of such transform would be e.g.

```py
class RandomHorizontalFlip(T.RandomHorizontalFlip):
    def forward(self, img1, img2, flow, valid):
        if torch.rand(1) > self.p:
            return img1, img2, flow, valid

        img1 = F.hflip(img1)
        img2 = F.hflip(img2)
        flow = F.hflip(flow) * torch.tensor([-1, 1])[:, None, None]
        if valid is not None:
            valid = F.hflip(valid)
        return img1, img2, flow, valid
```

You can find more at https://github.com/NicolasHug/vision/blob/raft/references/raft/transforms.py

torchvision/datasets/_optical_flow.py
```diff
@@ -0,0 +1,149 @@
...
+        pass
+
+    def __getitem__(self, index):
+        # Some datasets like Kitti have a built-in valid mask, indicating which flow values are valid
+        # For those we return (img1, img2, flow, valid), and for the rest we return (img1, img2, flow),
+        # and it's up to whatever consumes the dataset to decide what `valid` should be.
+
+        img1 = self._read_img(self._image_list[index][0])
+        img2 = self._read_img(self._image_list[index][1])
+        flow = self._read_flow(self._flow_list[index]) if self._flow_list else None
```
NicolasHug, line 44: Some datasets have a built-in "test" part but this part doesn't contain any target flow value, so we set the flow to `None` in this case.

torchvision/datasets/_optical_flow.py
```diff
@@ -0,0 +1,149 @@
...
+        else:
+            valid = None
+
+        if self.transforms is not None:
+            img1, img2, flow, valid = self.transforms(img1, img2, flow, valid)
+
+        if valid is None:
+            return img1, img2, flow
+        else:
+            return img1, img2, flow, valid
```
NicolasHug, line 57: leaving the `valid` mask aside (see other discussion above for that) and focusing on the `flow` value:

- For Sintel and other datasets (FlyingThings, FlyingChairs), the flow is currently a numpy array with shape (H, W, 2).
- For Kitti, the flow is a **tensor** with shape (2, H, W)

Should I change the output of Kitti to (H, W, 2) for it to be similar to what the other datasets return? Should I change the shape *and* the type (i.e. convert the tensor to a numpy array), or just the shape?

What the model consumes eventually is a tensor with shape (2, H, W), i.e. the format that Kitti outputs natively. I feel like it might be worth it for consistency sake tho.

torchvision/datasets/__init__.py
```diff
@@ -1,3 +1,4 @@
+from ._optical_flow import KittiFlow, Sintel
```
NicolasHug, line 1: Note the underscore: we may or may not make the `FlowDataset` base class public eventually, but for now it's safer to not expose anything

torchvision/datasets/_optical_flow.py
datumbox: Concerning `valid` being a passthrough, this is already happening in other cases so I don't think it's a problem. See:
https://github.com/pytorch/vision/blob/d367a01a18a3ae6bee13d8be3b63fd6a581ea46f/references/detection/transforms.py#L48-L54

Concerning the `self.transforms()` call, it assumes a specific API on the provided transforms. Might be worth checking with @pmeier and @fmassa that this is compatible with the new transforms RFC.

torchvision/datasets/_optical_flow.py
datumbox: Good question. Intuitively I'm weighting towards providing a consistent result (doesn't matter the option), but I think I would love to hear from @pmeier on whether this dilemma popped up in the past and how it was handled.

datumbox (commented):
Just a few thoughts:

torchvision/datasets/_optical_flow.py
datumbox: This is similar to `ImageNet`. Worth checking that the policy is aligned.

torchvision/datasets/_optical_flow.py
pmeier: > I'm a bit torn here. One one hand it's not convenient that different datasets have a different return size

This will be resolved by the new style datasets, because they always return a dictionary, with more or less information with respect to the dataset. So for now I would go with the different return sizes and equalize them in the reference script:

```python
for img1, img2, flow, *other_features in dataset:
    if not other_features:
        valid = flow.abs() < SOME_BIG_THRESHOLD
    else:
        valid = other_features[0]
```

We will probably always need some kind of equalization, because the datasets authors rarely follow a pattern.

torchvision/datasets/_optical_flow.py
pmeier: `ImageNet` does not provide a test split. I can't recall which dataset, but I think we agreed that if an piece of information is not available for a specific configuration, we simply return it as `None`.

torchvision/datasets/_optical_flow.py
pmeier: The new API will transform everything it knows to transform. So for example if you wrap `flow` in a (not-yet-existing) `features.Flow` the transforms will ignore it unless they know how to deal with it.

torchvision/datasets/_optical_flow.py
pmeier: > Intuitively I'm weighting towards providing a consistent result (doesn't matter the option)

Yes, that is the way to go. If there is a format that is used by the later models, go for that and convert everything to it. This is similar to images in conjunction with `transforms.ToTensor()`: the output will always be `CxHxW` regardless of how it is stored in the dataset.

fmassa: I agree. I would return consistent shapes and types, with a preference for `CxHxW` as this will work out of the box with our current transforms. We can always permute it back when using it with `grid_sample`

torchvision/datasets/_optical_flow.py
fmassa: This is fine, and the new API can handle the `flow` + `valid` internally (the `valid` could be an attribute of `flow` as well)

torchvision/datasets/_optical_flow.py
datumbox: @pmeier ImageNet has a test split but it doesn't have labels. I can't link to it cause you need to register to get it but it's 100k images in a flat folder structure. Here are the 10 first images:
```
$ ls -lh ./imagenet_full_size/061417/test/ | head
total 15G
-rw-r--r-- 1 root root   34K Sep 22  2014 ILSVRC2012_test_00000001.JPEG
-rw-r--r-- 1 root root  120K Sep 22  2014 ILSVRC2012_test_00000002.JPEG
-rw-r--r-- 1 root root   27K Sep 22  2014 ILSVRC2012_test_00000003.JPEG
-rw-r--r-- 1 root root  122K Sep 22  2014 ILSVRC2012_test_00000004.JPEG
-rw-r--r-- 1 root root   97K Sep 22  2014 ILSVRC2012_test_00000005.JPEG
-rw-r--r-- 1 root root  207K Sep 22  2014 ILSVRC2012_test_00000006.JPEG
-rw-r--r-- 1 root root  215K Sep 22  2014 ILSVRC2012_test_00000007.JPEG
-rw-r--r-- 1 root root  178K Sep 22  2014 ILSVRC2012_test_00000008.JPEG
-rw-r--r-- 1 root root   11K Sep 22  2014 ILSVRC2012_test_00000009.JPEG
```

pmeier: Sorry, let me rephrase: our builtin `ImageNet` dataset does not provide a test split:

https://github.com/pytorch/vision/blob/3300692c6e7c2023d2f2356a69ec22ca91e38790/torchvision/datasets/imagenet.py#L22-L28

NicolasHug:
@datumbox @pmeier @fmassa thanks a lot for your very useful feedback.

I updated the PR accordingly (see original PR message above) and added tests and docs. This is now ready for a proper review.

NicolasHug:
Rendered docs: https://963455-73328905-gh.circle-artifacts.com/0/docs/datasets.html

fmassa (approved):
Thanks!

vadimkantorov:
There were problems with older datasets that returning tuple doesn't allow for easy adding extra fields in the future, e.g. https://github.com/pytorch/vision/issues/3608. I propose to return img, target_dict tuple. And there is more unity when iterating between different related datasets.

pmeier:
@vadimkantorov We are aware of that and the new prototype dataset API will always return a single dictionary per sample. There is no documentation yet, but if you want to you can have a look into [`torchvision/prototype/datasets`](https://github.com/pytorch/vision/tree/main/torchvision/prototype/datasets) to see how things will look like in the future. Plus, if you have feedback, that is very welcome!

vadimkantorov:
If you're committed to using only tuples, please also return some internal example id, image file name or something of this form. Then at least people can hack around and fetch additional information

NicolasHug:
Thanks for your input @vadimkantorov 

I don't think any of the datasets return that kind of info so far (sadly), so returning a dict with extra info would make KittiFlow and Sintel outliers w.r.t. what we currently have. Returning the ids etc. is definitely useful, but such new design is better suited for a complete re-work of the datasets, which we are currently doing as @pmeier pointed out above.

datumbox (approved):
LGTM, thanks!

Definitely interesting to see how the ideas mentioned above can be incorporated on the new API but I think it's worth merging this PR to unblock your work on the flow models.

pmeier (approved):
LGTM, thanks @NicolasHug!

Issue #4675, opened 2021-10-20
Multi-pretrained weights: Add model builders for Detection


datumbox:
### 🚀 The feature

Update the model builders for all Detection models:
- [x] FasterRCNN
- [x] KeypointRCNN
- [x] MaskRCNN
- [x] RetinaNet
- [x] SSD
- [x] SSDlite

PR #4847, opened 2021-11-03
Adding multiweight support to FasterRCNN

datumbox:
Fixes partially #4675

Verified that both commands return the same:
```
torchrun --nproc_per_node=1 train.py --test-only --pretrained --model fasterrcnn_mobilenet_v3_large_320_fpn

torchrun --nproc_per_node=1 train.py --test-only --weights Coco_RefV1 --model fasterrcnn_mobilenet_v3_large_320_fpn
```

cc @datumbox @bjuncek


references/detection/train.py
```diff
@@ -33,6 +33,12 @@
 from group_by_aspect_ratio import GroupedBatchSampler, create_aspect_ratio_groups
 
 
+try:
+    from torchvision.prototype import models as PM
```
datumbox, line 37: Temporarily adding support of prototype on the detection reference script similar to all other areas, so we can test the results.

test/test_prototype_models.py
```diff
@@ -95,7 +106,10 @@ def test_old_vs_new_factory(model_fn, module_name, dev):
     model_name = model_fn.__name__
     kwargs = {"pretrained": True, **defaults[module_name], **TM._model_params.get(model_name, {})}
     input_shape = kwargs.pop("input_shape")
+    kwargs.pop("num_classes", None)  # ignore this as it's an incompatible speed optimization for pre-trained models
     x = torch.rand(input_shape).to(device=dev)
+    if module_name == "detection":
+        x = [x]
```
datumbox, line 112: Detection models receive a list of images.

torchvision/models/shufflenetv2.py
```diff
@@ -162,7 +162,7 @@ def _shufflenetv2(arch: str, pretrained: bool, progress: bool, *args: Any, **kwa
     if pretrained:
         model_url = model_urls[arch]
         if model_url is None:
-            raise NotImplementedError(f"pretrained {arch} is not supported as of now")
+            raise ValueError(f"No checkpoint is available for model type {arch}")
```
datumbox, line 165: Just an unrelated correction, to align the exception type and value with the entire TorchVision so that we can capture it easier.

torchvision/prototype/models/detection/faster_rcnn.py
```diff
@@ -64,3 +102,109 @@ def fasterrcnn_resnet50_fpn(
...
+    **kwargs: Any,
+) -> FasterRCNN:
+    if "pretrained" in kwargs:
+        warnings.warn("The argument pretrained is deprecated, please use weights instead.")
+        weights = FasterRCNNMobileNetV3LargeFPNWeights.Coco_RefV1 if kwargs.pop("pretrained") else None
+    weights = FasterRCNNMobileNetV3LargeFPNWeights.verify(weights)
+    if "pretrained_backbone" in kwargs:
+        warnings.warn("The argument pretrained_backbone is deprecated, please use weights_backbone instead.")
+        weights_backbone = MobileNetV3LargeWeights.ImageNet1K_RefV1 if kwargs.pop("pretrained_backbone") else None
+    weights_backbone = MobileNetV3LargeWeights.verify(weights_backbone)
```
datumbox, line 160: Will eventually be refactored to reduce duplicate code but that's part of the clean up.

datumbox (commented):
Adding few comments:

references/detection/train.py
```diff
@@ -41,8 +47,15 @@ def get_dataset(name, image_set, transform, data_path):
...
 
-def get_transform(train, data_augmentation):
-    return presets.DetectionPresetTrain(data_augmentation) if train else presets.DetectionPresetEval()
+def get_transform(train, args):
+    if train:
+        return presets.DetectionPresetTrain(args.data_augmentation)
+    elif not args.weights:
+        return presets.DetectionPresetEval()
+    else:
+        fn = PM.detection.__dict__[args.model]
```
jdsgomes, line 56: should we also check for PM here and raise exception?

references/detection/train.py
```diff
@@ -41,8 +47,15 @@ def get_dataset(name, image_set, transform, data_path):
     return ds, num_classes
 
 
-def get_transform(train, data_augmentation):
-    return presets.DetectionPresetTrain(data_augmentation) if train else presets.DetectionPresetEval()
```
jdsgomes, line 45: nit: would it be clearer to have weights as explicit argument

jdsgomes (commented):
Just some small comments, let me know your thoughts. Otherwise looks good.

references/detection/train.py
datumbox: Yeah we could/should. 

Note that this is temporary for as long as we check the models, so I don't want to have a massive amount of error handling spread all over the place. Perhaps a middle ground is to move the existing exception somewhere where it will be triggered if the PM is None once. I would prefer to follow up on a different PR because this change needs to happen across all scripts. Thoughts?

references/detection/train.py
jdsgomes: yes that makes sense.

references/detection/train.py
jdsgomes: I am ok with that. approving the PR in that case

jdsgomes (approved):
Thanks!

references/detection/train.py
datumbox: Thanks I'll send a new PR in a bit.

Issue #4674, opened 2021-10-20
Multi-pretrained weights: Add model builders for Quantization


datumbox:
### 🚀 The feature

Update the model builders for all Quantized Classification models:
- [x] GoogleNet
- [x] InceptionV3
- [x] MobileNetV2
- [x] MobileNetV3
- [x] ResNet
- [x] ShuffleNetV2

PR #678, opened 2018-12-08
Add GoogLeNet (Inception v1)

TheCodez:
This adds the GoogLeNet (Inception v1) model, including the auxiliary classifiers found in the paper. Related to Issue #537

I have updated the example project on my [branch](https://github.com/TheCodez/examples/tree/inception) to add support for training the _GoogLeNet_. Sadly I don't have the computing resources available to train the model on ImageNet.

PR #4848, opened 2021-11-03
Adding multiweight support to Quantized GoogLeNet

datumbox:
Fixes partially #4674

Verified with:
```
python -u train_quantization.py --test-only --weights ImageNet1K_FBGEMM_TFV1 --model googlenet --device cpu
```

cc @bjuncek


torchvision/prototype/models/googlenet.py
```diff
@@ -14,14 +14,13 @@
 
 
 class GoogLeNetWeights(Weights):
-    ImageNet1K_Community = WeightEntry(
+    ImageNet1K_TFV1 = WeightEntry(
```
datumbox, line 17: This was previously incorrectly tagged as community contribution while in reality the weights were ported from TF. Proof: https://github.com/pytorch/vision/pull/678#issuecomment-470318912

torchvision/prototype/models/quantization/googlenet.py
```diff
@@ -0,0 +1,88 @@
...
+            weights = QuantizedGoogLeNetWeights.ImageNet1K_FBGEMM_TFV1 if quantize else GoogLeNetWeights.ImageNet1K_TFV1
+        else:
+            weights = None
+
+    if quantize:
+        weights = QuantizedGoogLeNetWeights.verify(weights)
+    else:
+        weights = GoogLeNetWeights.verify(weights)
+
+    original_aux_logits = kwargs.get("aux_logits", False)
```
datumbox, line 61: We simplify similarly to the unquantized builder.

torchvision/prototype/models/quantization/googlenet.py
```diff
@@ -0,0 +1,88 @@
...
+        if original_aux_logits:
+            warnings.warn(
+                "auxiliary heads in the pretrained googlenet model are NOT pretrained, so make sure to train them"
+            )
+        kwargs["aux_logits"] = True
+        kwargs["init_weights"] = False
+        kwargs["num_classes"] = len(weights.meta["categories"])
+        if "backend" in weights.meta:
+            kwargs["backend"] = weights.meta["backend"]
+    backend = kwargs.pop("backend", "fbgemm")
```
datumbox, line 74: The rest is similar to the quantized implementation for resnet.

datumbox (commented):
Few comments to assist review:

torchvision/prototype/models/quantization/resnet.py
```diff
@@ -54,7 +56,9 @@ def _resnet(
 _common_meta = {
     "size": (224, 224),
     "categories": _IMAGENET_CATEGORIES,
+    "interpolation": InterpolationMode.BILINEAR,
     "backend": "fbgemm",
+    "quantization": "ptq",
```
datumbox, line 61: Adding some additional meta-data such as interpolation, type of quantization and a reference to the unquantized enum of the the weights to all previous models.

torchvision/prototype/models/googlenet.py
```diff
@@ -14,14 +14,13 @@
...
         url="https://download.pytorch.org/models/googlenet-1378be20.pth",
         transforms=partial(ImageNetEval, crop_size=224),
         meta={
             "size": (224, 224),
             "categories": _IMAGENET_CATEGORIES,
             "interpolation": InterpolationMode.BILINEAR,
-            "recipe": "https://github.com/TheCodez/examples/blob/inception/imagenet/README.md#googlenet",
```
jdsgomes, line 24: I Think in other PRs we did `"recipe": ""`, should we align this in the end as well?

jdsgomes (approved):
LGTM!

torchvision/prototype/models/googlenet.py
datumbox: Good spot. So there are no longer `"recipe": ""` because in an earlier commit I actually added all of them. Only one of them doesn't support it, so I've set it to None:
https://github.com/pytorch/vision/blob/6a60b9bccc4faf48b9738f0f3906d1b8108fd03a/torchvision/prototype/models/densenet.py#L70

But your comment made me realize that for a few TF models that don't have a recipe, I haven't set the key to `None`. Should I omit if not available or set it to `None`? Let me know and I'll align the behaviour everywhere in this PR.

PR #4850, opened 2021-11-03
Adding multiweight support to Quantized InceptionV3

datumbox:
Fixes partially #4674

Verified with:
```
python -u train_quantization.py --test-only --weights ImageNet1K_FBGEMM_TFV1 --model inception_v3 --device cpu
```

cc @bjuncek

### DIFF
--- a/torchvision/models/quantization/inception.py
+++ b/torchvision/models/quantization/inception.py
@@ -24,72 +24,6 @@ quant_model_urls = {
 }
 
 
-def inception_v3(
-    pretrained: bool = False,
-    progress: bool = True,
-    quantize: bool = False,
-    **kwargs: Any,
-) -> "QuantizableInception3":
-
-    r"""Inception v3 model architecture from
-    `"Rethinking the Inception Architecture for Computer Vision" <http://arxiv.org/abs/1512.00567>`_.
-
-    .. note::
-        **Important**: In contrast to the other models the inception_v3 expects tensors with a size of
-        N x 3 x 299 x 299, so ensure your images are sized accordingly.
-
-    Note that quantize = True returns a quantized model with 8 bit
-    weights. Quantized models only support inference and run on CPUs.
-    GPU inference is not yet supported
-
-    Args:
-        pretrained (bool): If True, returns a model pre-trained on ImageNet
-        progress (bool): If True, displays a progress bar of the download to stderr
-        quantize (bool): If True, return a quantized version of the model
-        aux_logits (bool): If True, add an auxiliary branch that can improve training.
-            Default: *True*
-        transform_input (bool): If True, preprocesses the input according to the method with which it
-            was trained on ImageNet. Default: *False*
-    """
-    if pretrained:
-        if "transform_input" not in kwargs:
-            kwargs["transform_input"] = True
-        if "aux_logits" in kwargs:
-            original_aux_logits = kwargs["aux_logits"]
-            kwargs["aux_logits"] = True
-        else:
-            original_aux_logits = False
-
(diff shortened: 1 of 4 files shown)


torchvision/models/quantization/inception.py
```diff
@@ -24,72 +24,6 @@
 }
 
 
-def inception_v3(
-    pretrained: bool = False,
-    progress: bool = True,
-    quantize: bool = False,
-    **kwargs: Any,
-) -> "QuantizableInception3":
```
datumbox, line 32: Moving the builder on the bottom of the page to use proper typing.

torchvision/models/quantization/inception.py
```diff
@@ -237,3 +171,68 @@ def fuse_model(self) -> None:
...
             if type(m) is QuantizableBasicConv2d:
                 m.fuse_model()
+
+
+def inception_v3(
+    pretrained: bool = False,
+    progress: bool = True,
+    quantize: bool = False,
+    **kwargs: Any,
+) -> QuantizableInception3:
```
datumbox, line 181: No review, just copy-paste.

torchvision/prototype/models/inception.py
```diff
@@ -28,11 +28,11 @@ class Inception3Weights(Weights):
     )
 
 
-def inception_v3(weights: Optional[Inception3Weights] = None, progress: bool = True, **kwargs: Any) -> Inception3:
+def inception_v3(weights: Optional[InceptionV3Weights] = None, progress: bool = True, **kwargs: Any) -> Inception3:
```
datumbox, line 31: The naming is inconsistent. The model class is called `Inception3`, while the mode builder inception_**v**3. It's unclear how we should name the weights. Following the convention from ResNets this should have been `Inception_V3` which is ugly. I propose not to spend more time on this now and tackle it at #4652. I've already added a reference for this issue.

torchvision/prototype/models/quantization/inception.py
```diff
@@ -0,0 +1,86 @@
...
+        weights = QuantizedInceptionV3Weights.verify(weights)
+    else:
+        weights = InceptionV3Weights.verify(weights)
+
+    original_aux_logits = kwargs.get("aux_logits", False)
+    if weights is not None:
+        if "transform_input" not in kwargs:
+            kwargs["transform_input"] = True
+        kwargs["aux_logits"] = True
+        kwargs["init_weights"] = False
```
datumbox, line 66: The original implementation is missing this. I believe that should be added to avoid the unnecessary initialization. For context, this was added in the past because the initialization was slow and nowadays it's fast so this is less useful.

torchvision/prototype/models/quantization/inception.py
```diff
@@ -0,0 +1,86 @@
...
+            kwargs["backend"] = weights.meta["backend"]
+    backend = kwargs.pop("backend", "fbgemm")
+
+    model = QuantizableInception3(**kwargs)
+    _replace_relu(model)
+    if quantize:
+        quantize_model(model, backend)
+
+    if weights is not None:
+        if quantize and not original_aux_logits:
```
datumbox, line 78: Special handling needed here. See original implementation.

datumbox (commented):
Few comments to assist review:

torchvision/prototype/models/quantization/inception.py
datumbox: Actually for this to work, we would also need to expose it on the `QuantizableInception3` where we currently don't. Not sure if it's worth expanding it but this should be tackled on a separate PR.

jdsgomes (approved):
LGTM!

Issue #4818, opened 2021-11-01
Verify the threshold input for `solarize()` and add a test


datumbox:
### 🚀 The feature

The `solarize()` method receives a `threshold` which is currently not asserted based on the image type:
https://github.com/pytorch/vision/blob/d367a01a18a3ae6bee13d8be3b63fd6a581ea46f/torchvision/transforms/functional_tensor.py#L876-L886

Ideally we should assert it. It's upper bound depends on the image type (1.0 for float and 255 for uint8) similar to:
https://github.com/pytorch/vision/blob/d367a01a18a3ae6bee13d8be3b63fd6a581ea46f/torchvision/transforms/functional_tensor.py#L858

Along with the assertion we should add a unit-test to ensure that if the threshold is over the permitted value it will fail.

### Motivation, pitch

Due to the missing assertion, we didn't spot a bug at the configuration of AutoAugment methods. See #4805.

cc @vfdev-5 @datumbox @pmeier

puhuk:
Can I take this

pmeier:
@puhuk Sure, go ahead! Ping me or @datumbox on the PR for review.

PR #4852, opened 2021-11-04
Update functional_tensor.py

puhuk:
To resolve issue #4818
Add assert function and logic after checking bound of image.


fmassa (commented):
Thanks!

Failures seems related, can you have a look?

puhuk:
Sure, let me see and send again :)
Thanks for review.

torchvision/transforms/functional_tensor.py
```diff
@@ -15,6 +15,10 @@ def _assert_image_tensor(img: Tensor) -> None:
     if not _is_tensor_a_torch_image(img):
         raise TypeError("Tensor is not a torch image.")
 
+def _assert_threshold(img: Tensor, threshold:int) -> None:
+    bound = 1 if img.is_floating_point() else 255
+    if threshold >= bound:
```
datumbox, line 20: Are you sure we need equality here? Is the upper bound a permitted value?

torchvision/transforms/functional_tensor.py
```diff
@@ -15,6 +15,10 @@ def _assert_image_tensor(img: Tensor) -> None:
     if not _is_tensor_a_torch_image(img):
         raise TypeError("Tensor is not a torch image.")
 
+def _assert_threshold(img: Tensor, threshold:int) -> None:
```
datumbox, line 18:
```suggestion
def _assert_threshold(img: Tensor, threshold: float) -> None:
```

datumbox (commented):
Thanks for the PR. I left a couple of comments, let me know what you think.

torchvision/transforms/functional_tensor.py
puhuk: I add equality because threshold should be less than upper bound.
It was bit confusing to me when I though what happens if threshold is 255 which is same as upper bound (255).

torchvision/transforms/functional_tensor.py
puhuk: Thanks for the review :) It should be `float`.

torchvision/transforms/functional_tensor.py
datumbox: I think the upper bound should be a permitted value. See:
https://github.com/pytorch/vision/blob/40c3ba24bfea6de02e6e1b0bd10b09d978e70506/torchvision/transforms/autoaugment.py#L226

puhuk: Thanks! Let me check and resend the PR :)

datumbox (commented):
@puhuk Thanks it looks great. 

Before merging could you also add tests to cover for:
1. Corner cases such values right on the threshold
2. Ensure that the code fails when an incorrect value is passed

puhuk:
@datumbox, sure let me add test.
But may I ask where can I add test and is there any code I can refer to?

datumbox:
@puhuk You can find similar tests [here](https://github.com/pytorch/vision/blob/main/test/test_transforms_tensor.py).

I think we need two tests:
- One that receives an all white (255) 4x4 image and checks that the method works as expected for threshold 255.
- One that passes the wrong threshold and checks that it fails. Here is an example of a test that [checks for failures](https://github.com/pytorch/vision/blob/main/test/test_transforms_tensor.py#L749-L754).

puhuk:
@datumbox Thanks for sharing! Let me check and send PR

datumbox:
@puhuk Some of your tests are [failing](https://app.circleci.com/pipelines/github/pytorch/vision/12540/workflows/c5b69f05-8bd9-48c6-9079-4d60e0f237ef/jobs/991970) and they are related to your changes.

puhuk:
@datumbox, I'm trying to fix this. Could you give me some time.

datumbox:
@puhuk Sure thing. Sometimes contributors don't give access to the CircleCi and cant see the errors, so we give a nudge to ensure they are aware. Since you know it and actively work on it, there is no rush. :)

test/test_functional_tensor.py
```diff
@@ -795,6 +795,32 @@ def test_solarize2(device, dtype, config, channels):
     )
 
 
+@pytest.mark.parametrize("config", [{"threshold": threshold} for threshold in [0.0, 0.25, 0.5, 0.75, 1.0,]])
+def test_solarize_threshold1_bound(config):
```
datumbox, line 799: Why not simplify as follows?
```python
@pytest.mark.parametrize("threshold", [0.0, 0.25, 0.5, 0.75, 1.0,])
def test_solarize_threshold1_bound(threshold):
    # rest goes here
```

test/test_functional_tensor.py
```diff
@@ -795,6 +795,32 @@ def test_solarize2(device, dtype, config, channels):
...
+    img = torch.rand((3, 12, 23)).to("cpu")
```
datumbox, line 800: It doesn't run on GPU by default:
```suggestion
    img = torch.rand((3, 12, 23))
```

datumbox (commented):
Thanks @puhuk, I think we can simplify the unit-tests a bit. Let me know what you think.

test/test_functional_tensor.py
puhuk: Thanks for the review! Because I'm not accustomed to pytest, I followed some previous code :)
Let me simplify and send PR again.

test/test_functional_tensor.py
puhuk: Thanks! Let me correct it to run on GPU as well.

datumbox: I should have been clearer here. Your test doesn't benefit greatly from running both on CPU and GPU because we just check if the test passes/fails due to the provided threshold. So here I was just suggesting to remove the `.to("cpu")` and keep everything else as-is. But given that all other tests run in both CPU/GPU and your new tests are super fast, there is no harm in running them in both. :)

PR #4854, opened 2021-11-04
Cleanup Quantized ShuffleNet

datumbox:
This PR:
- Removes the quantized model builders for variants `shufflenet_v2_x1_5` and `shufflenet_v2_x2_0`. These two were added by mistake and they are unusable because we have neither quantized weights nor normal weights.
- Adds quantized weights for `shufflenet_v2_x0_5` variant, so that the equivalent model builder makes sense.

Verified by:
```
python train_quantization.py --device='cpu' --model='shufflenet_v2_x0_5' --test-only
Test:  Acc@1 57.972 Acc@5 79.780
```

The above builders caused problems to users and quite often on FBcode you can find code that tries to skip/ignore it. See this for example: https://github.com/pytorch/pytorch/pull/67836

With the removal of the problematic methods, we can now clean up the skips non the next fbsync.

### DIFF
--- a/docs/source/models.rst
+++ b/docs/source/models.rst
@@ -479,7 +479,8 @@ Model                             Acc@1          Acc@5
 ================================  =============  =============
 MobileNet V2                      71.658         90.150
 MobileNet V3 Large                73.004         90.858
-ShuffleNet V2                     68.360         87.582
+ShuffleNet V2 x1.0                68.360         87.582
+ShuffleNet V2 x0.5                57.972         79.780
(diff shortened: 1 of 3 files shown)


fmassa (approved):
Thanks!

datumbox:
The single [failed](https://app.circleci.com/pipelines/github/pytorch/vision/12251/workflows/b95900da-c611-43af-a2e8-213bffe37421/jobs/963484/tests) tests is unrelated and caused by network flakiness. Merging.

PR #4860, opened 2021-11-04
Add FlyingChairs dataset for optical flow

NicolasHug:
This PRs adds the FlyingChairs dataset for optical flow

It also adds a proper way of testing the .flo reader, instead of relying on a mock as previouly done

cc @pmeier


test/test_datasets.py
```diff
@@ -1920,7 +1917,8 @@ def test_flow(self):
         with self.create_dataset(split="train") as (dataset, _):
             assert dataset._flow_list and len(dataset._flow_list) == len(dataset._image_list)
             for _, _, flow in dataset:
-                assert flow == self._FAKE_FLOW
+                assert flow.shape == (2, self.FLOW_H, self.FLOW_W)
```
pmeier, line 1920: Instead of passing this information through class variables, `inject_fake_data` can also return a dictionary. If you return `dict(num_examples=..., flow_shape=...)` there, you can access it here with

```python
with self.create_dataset(...) as (dataset, info):
    ...
    assert flow.shape == (2, *info["flow_shape"])
```

torchvision/datasets/_optical_flow.py
```diff
@@ -208,6 +209,72 @@ def _read_flow(self, file_name):
...
+            ``img1, img2, flow, valid`` and returns a transformed version.
+            ``valid`` is expected for consistency with other datasets which
+            return a built-in valid mask, such as :class:`~torchvision.datasets.KittiFlow`.
+    """
+
+    def __init__(self, root, split="train", transforms=None):
+        super().__init__(root=root, transforms=transforms)
+
+        if split not in ("train", "val"):
+            raise ValueError("split must be either 'train' or 'val'")
```
pmeier, line 242: Nit: use 

https://github.com/pytorch/vision/blob/eb48a1d81c879089df78b77be5891d079dd643b1/torchvision/datasets/utils.py#L442-L447

```suggestion
        verify_str_arg("split", split, ("train", "val"))
```

torchvision/datasets/_optical_flow.py
```diff
@@ -208,6 +209,72 @@ def _read_flow(self, file_name):
         return _read_16bits_png_with_flow_and_valid_mask(file_name)
 
 
+class FlyingChairs(FlowDataset):
+    """`FlyingChairs <https://lmb.informatik.uni-freiburg.de/resources/datasets/FlyingChairs.en.html#flyingchairs>`_ Dataset for optical flow.
+
+    You will also need to download the FlyingChairs_train_val.txt file from the dataset page.
```
pmeier, line 215: Is not providing download functionality for these datasets intentional?

pmeier (approved):
A few minor comments inline. Apart from them, you also need to add the dataset to the documentation. Otherwise LGTM, thanks @NicolasHug!

torchvision/datasets/_optical_flow.py
NicolasHug: For this one specifically and for Sintel, I think we could provide download functionalities. I'm just leaving this for another potential PR

For the others it's not as easy because e.g. FlyingThings3D can only be downloaded via bittorrent, and for Kitti one need to register on the website first (in theory...).

PR #4861, opened 2021-11-04
add initial chunk of prototype transforms

pmeier:
This ports the initial chunk of transforms from https://github.com/pmeier/torchvision-datasets-rework/tree/rfc/transforms. You can find a discussion of this in pmeier/torchvision-datasets-rework#1.

Here is a quick demonstration:

```python
import os.path

from torchvision import io, utils
from torchvision.prototype import transforms, features


def show(image, bounding_box):
    annotated_image = utils.draw_bounding_boxes(image, bounding_box.convert("xyxy").unsqueeze(0))
    to_pil_image(annotated_image).show()


image = features.Image(io.read_image(os.path.join("test", "assets", "fakedata", "logos", "rgb_pytorch.jpg")))
bounding_box = features.BoundingBox((60, 30, 12, 12), format="cxcywh", image_size=image.shape[-2:])

show(image, bounding_box)

    transforms.Resize(256),
    transforms.HorizontalFlip(),
)

# this would normally come from a dataset, i.e. `for sample in datasets.load("coco")`
sample = dict(foo=image, bar=bounding_box)
transformed_sample = transform(sample)

show(transformed_sample["foo"], transformed_sample["bar"])
```
before
after

Note to reviewers: there are two changes that make the diff look a lot larger than it is:

1. I've renamed `torchvision/prototype/transforms/presets.py` to `_presets.py` in order to make it consistent with the other files there. All the changes in `torchvision/prototype/models` are just fixing the imports.
2. I've moved some functionality from `torchvision/prototype/datasets/utils/_internal.py` to `torchvision/prototype/utils/_internal.py`. This also results in a few import changes in other files.


torchvision/prototype/features/_image.py
```diff
@@ -18,6 +18,15 @@ class Image(Feature):
     color_spaces = ColorSpace
     color_space: ColorSpace
 
+    @classmethod
+    def _to_tensor(cls, data, *, dtype, device):
+        tensor = torch.as_tensor(data, dtype=dtype, device=device)
+        if tensor.ndim == 2:
+            tensor = tensor.unsqueeze(0)
+        elif tensor.ndim != 3:
+            raise ValueError()
```
fmassa, line 27: So batch of images should be handled separately?

torchvision/prototype/transforms/_geometry.py
```diff
@@ -0,0 +1,130 @@
...
+class HorizontalFlip(Transform):
+    @staticmethod
+    def image(input: Image) -> Image:
+        return Image(input.flip((-1,)), like=input)
+
+    @staticmethod
+    def bounding_box(input: BoundingBox) -> BoundingBox:
+        x, y, w, h = input.convert("xywh").to_parts()
+        x = input.image_size[1] - (x + w)
+        return BoundingBox.from_parts(x, y, w, h, like=input, format="xywh")
```
fmassa, line 19: I'm not 100% sure about what `from_parts` is doing, but we should make sure that the format returned by `HorizontalFlip` is consistent with the format from `input` -- I assume `like` will take care of this here, and `format` will just take the format given to itself as `a, b, c, d` ?

torchvision/prototype/transforms/_geometry.py
```diff
@@ -0,0 +1,130 @@
...
+        super().__init__()
+        self.size = (size, size) if isinstance(size, int) else size
+        self.interpolation_mode = interpolation_mode
+
+    def get_params(self, sample: Any) -> Dict[str, Any]:
+        return dict(size=self.size, interpolation_mode=self.interpolation_mode)
+
+    @staticmethod
+    def image(input: Image, *, size: Tuple[int, int], interpolation_mode: str = "nearest") -> Image:
+        return Image(interpolate(input.unsqueeze(0), size, mode=interpolation_mode).squeeze(0), like=input)
```
fmassa, line 38: Again, we are assuming here that `image` is just an individual image, and we will have separate classes for video and batches of images?

torchvision/prototype/transforms/_geometry.py
```diff
@@ -0,0 +1,130 @@
...
+        self.crop_box = crop_box.convert("xyxy")
+
+    def get_params(self, sample: Any) -> Dict[str, Any]:
+        return dict(crop_box=self.crop_box)
+
+    @staticmethod
+    def image(input: Image, *, crop_box: BoundingBox) -> Image:
+        # FIXME: pad input in case it is smaller than crop_box
+        x1, y1, x2, y2 = crop_box.convert("xyxy").to_parts()
+        return Image(input[..., y1 : y2 + 1, x1 : x2 + 1], like=input)  # type: ignore[misc]
```
fmassa, line 94: At first I thought we weren't implementing it for `BoundingBox`. Might be good to maybe add a comment or something to indicate that `bounding_box` is handled by the default implementation.

Or maybe even better and be explicit about it and just implement a `bounding_box` method that just returns the identity?

torchvision/prototype/transforms/_misc.py
```diff
@@ -0,0 +1,48 @@
...
+            self.register_feature_transform(feature_type, lambda input, **params: input)
+
+
+class Lambda(Transform):
+    def __new__(cls, lambd: Callable) -> Transform:  # type: ignore[misc]
+        warnings.warn("transforms.Lambda(...) is deprecated. Use transforms.Transform.from_callable(...) instead.")
+        # We need to generate a new class everytime a Lambda transform is created, since the feature transforms are
+        # registered on the class rather than on the instance. If we didn't, registering a feature transform will
+        # overwrite it on **all** Lambda transform instances.
+        return Transform.from_callable(lambd, name="Lambda")
```
fmassa, line 25: Should we even implement it? I think it might be better to just ask users to implement their own transform, no?

torchvision/prototype/transforms/_container.py
```diff
@@ -0,0 +1,92 @@
...
+
+class Compose(MultiTransform):
+    def forward(self, *inputs: Any, strict: bool = False) -> Any:
+        sample = inputs if len(inputs) > 1 else inputs[0]
+        for transform in self._transforms:
+            sample = transform(sample, strict=strict)
+        return sample
+
+
+class RandomApply(WrapperTransform):
```
fmassa, line 63: Is that fair to assume that the pattern we will want to follow for now on is to instead of implementing `RandomHorizontalFlip` to instead just to `RandomApply(HorizontalFlip)` ?

fmassa (approved):
LGTM, thanks a lot!

Would be good to add tests in a follow-up PR. Also, it would be instructive to implement the basic presets to see how this combines together (without AutoAugment for now for simplicity).

torchvision/prototype/features/_image.py
pmeier: Isn't that what we agreed to? Right now there is no plan to add a special feature for `BatchImage` or the like.

torchvision/prototype/transforms/_geometry.py
pmeier: > I'm not 100% sure about what `from_parts` is doing

It is a convenience method, that creates a new `BoundingBox` from the individual parts. It is equivalent to do `BoundingBox(torch.stack((x, y, w, h), dim=-1), ...)`.

> but we should make sure that the format returned by `HorizontalFlip` is consistent with the format from `input`

It this a general requirement? My idea was to keep the format variable an each transform converts to the format that is most convenient for it. Given that the user will need to perform some "special" collation for bounding boxes anyway, IMO it is ok to ask them to convert it to whatever format they want to use then. For example

```python
dataset = datasets.load("foo").map(transform).map(my_special_collate(bbox_format="cxcywh"))
image_batch, bbox_batch = next(iter(dataset))
```

>  I assume `like` will take care of this here, and `format` will just take the format given to itself as `a, b, c, d` ?

Nope. By passing `format` explicitly, this takes precedence over `like`. To achieve what you want we need to add a `.convert(input.format)` after the creation. Note that we cannot drop the explicit `format`, because the data is in `xywh` format and we need create the `BoundingBox` with this format first before we can convert to any other format.

torchvision/prototype/transforms/_geometry.py
pmeier: `Video` yes, batches of images no-ish. From our preliminary talk I got that we do not want to support batched images.

torchvision/prototype/transforms/_geometry.py
pmeier: > At first I thought we weren't implementing it for `BoundingBox`. 

Your first impression is right: currently `Crop` is a no-op for `BoundingBox`'es. This is of course not "right" and needs to be implemented, but I haven't done that yet. What behavior to we want for this? Return `None` in case the area that the bounding box would cover after the crop is less than a threshold?

>  Might be good to maybe add a comment or something to indicate that `bounding_box` is handled by the default implementation.

What "default implementation" do you mean here? The default behavior is to ignore unknown `Feature`'s.

> Or maybe even better and be explicit about it and just implement a bounding_box method that just returns the identity?

Not sure the verbosity is worth it. Note for consistency we would need to do that for all unsupported `Feature`'s such as `Label` and so on. That means if we add a new `Feature` we need to touch every single transform to add a no-op to it.

IMO clearly communicating that "unlisted" feature transforms will be no-ops is easy enough to understand and reduces the boilerplate by quite a bit.

torchvision/prototype/transforms/_misc.py
pmeier: IIRC, we wanted to keep BC for the transforms. If that still holds, we need it. Otherwise I'm happy to remove it.

torchvision/prototype/transforms/_container.py
pmeier: Yes, that is the pattern that I have in mind. That way we would only need to implement the "base" version of each transform without loosing any functionality. Plus, the user now has the option to use `HorizontalFlip` without a random component.

Note that there will be some `RandomX` transforms, namely ones where the parameter sampling has a random component. Since they otherwise behave like the "base" transform, it is sufficient to only implement the sampling and wrap  the "base" version. See `Resize` and `RandomResize` from this PR for an example.

torchvision/prototype/features/_image.py
fmassa: I think we might be confusing things. Probably better to discuss on VC

torchvision/prototype/transforms/_geometry.py
fmassa: > It this a general requirement?

This is generally a good thing to have, as it will avoid surprises to users. I would strongly recommend returning the same format as the input, even if internally we convert to a different format for performing the operation.

Imagine if you pass bounding boxes as `xyxy`, apply a `flip` operation, get a bounding box as `xywh`, and then cast it to Tensor to pass to the model? You won't even see that the bounding box was in `xywh` because it's implicit, and your model will be ingesting the wrong data.

torchvision/prototype/transforms/_geometry.py
fmassa: Let's chat to make sure we are aligned there

datumbox (changes requested):
@pmeier Some of the tests are failing and they are related (see [1](https://app.circleci.com/pipelines/github/pytorch/vision/12354/workflows/4f0064b1-addb-4b5d-8736-6a36ab7cfe82/jobs/973759), [2](https://app.circleci.com/pipelines/github/pytorch/vision/12354/workflows/ad7a2a62-9456-4551-9718-30b69f284ec3/jobs/973701))

Marking with "Request changes" to avoid accidental merges. We can merge after fixing the tests.

torchvision/prototype/transforms/_geometry.py
fmassa: I think we should either return the right thing, or raise a `NotImplemented` error.

This is one of the downsides of having a no-op by default -- no error is raised, and user might think that things are just magically working, which isn't the case.

torchvision/prototype/transforms/_misc.py
fmassa: We can keep the old `Lambda` transform as it was originally implemented, and say that it doesn't work with the new style of transforms. This is basically how I would envision a deprecation of a transform on a new style transforms

torchvision/prototype/features/_image.py
pmeier: As discussed offline, we will not a `BatchedImages` feature for now.

torchvision/prototype/transforms/_geometry.py
pmeier: As discussed offline, I'll add a `Video` feature as soon as we land the first video dataset.

torchvision/prototype/transforms/_geometry.py
pmeier: As discussed offline, we will go with explicit no-ops to avoid subtle bugs.

datumbox (approved):
Thanks for fixing the import issue. LGTM.

PR #4863, opened 2021-11-04
use better default decoder for prototype dataset tests

pmeier:
Without it, we can't know if a file could be read or not. Test failures are valid and will be resolved by pytorch/data#97.

cc @pmeier @bjuncek


fmassa (approved):
Good to merge once torchdata is updated

PR #4864, opened 2021-11-04
return features instead of vanilla tensors from prototype datasets

pmeier:
This PR is not exhaustive. I'll send follow-up PRs for the missing datasets as soon as we have tests for them.

cc @pmeier @bjuncek


torchvision/prototype/datasets/_builtin/caltech.py
```diff
@@ -95,8 +96,8 @@ def _collate_and_decode_sample(
         image = decoder(image_buffer) if decoder else image_buffer
 
         ann = read_mat(ann_buffer)
-        bbox = torch.as_tensor(ann["box_coord"].astype(np.int64))
-        contour = torch.as_tensor(ann["obj_contour"])
+        bbox = BoundingBox(ann["box_coord"].astype(np.int64).squeeze()[[2, 0, 3, 1]], format="xyxy")
```
pmeier, line 99: Let's not block this here, but I was wondering what we set as `image_size` if the user specifies `decoder=None`? If not set, the `BoundingBox` will default to an `image_size` that is barely able to contain it.

fmassa: It is possible to get the image sizes without fully decoding the image. We haven't implemented it in torchvision yet but I think we should. So we could always return the correct image properties here

torchvision/prototype/features/_bounding_box.py
```diff
@@ -115,7 +115,10 @@ def guess_image_size(cls, data: torch.Tensor, *, format: BoundingBoxFormat) -> T
                 data = cls._TO_XYXY_MAP[format](data)
             data = cls._FROM_XYXY_MAP[BoundingBoxFormat.XYWH](data)
         *_, w, h = to_parts(data)
-        return int(h.ceil()), int(w.ceil())
+        if data.dtype.is_floating_point:
+            w.ceil_()
+            h.ceil_()
```
fmassa, line 120: Is this change needed? Also, does it change `data` inplace?

torchvision/prototype/features/_feature.py
```diff
@@ -16,14 +16,16 @@ class Feature(torch.Tensor):
...
+        meta_attrs = {attr for attr in cls.__annotations__.keys() - cls.__dict__.keys() if not attr.startswith("_")}
+        for super_cls in cls.__mro__[1:]:
+            if super_cls is Feature:
+                break
+
+            meta_attrs.update(super_cls._META_ATTRS)
+
+        cls._META_ATTRS = meta_attrs
+        for attr in meta_attrs:
+            setattr(cls, attr, property(lambda self, attr=attr: self._meta_data[attr]))
```
fmassa, line 28: Can you add some comments explaining what is happening here?

torchvision/prototype/features/_pixels.py
```diff
@@ -0,0 +1,5 @@
+from ._feature import Feature
+
+
+class Pixels(Feature):
+    pass
```
fmassa, line 5: What are the semantics that we envision for `Pixels`? What is its format structure? Can it be tied to `Keypoint`s?

fmassa (commented):
Generally LGTM, although I have some questions regarding `Pixels`

torchvision/prototype/features/_bounding_box.py
pmeier: > Is this change needed?

Yes:

```python
>>> torch.tensor(0, dtype=torch.uint8).ceil()
RuntimeError: "ceil" "_vml_cpu" not implemented for 'Byte'
```

If we decide that images always have a floating point dtype, we can remove it once again.

> Also, does it change `data` inplace?

Yes, good catch. I was under the impression that `.unbind` returned a copy.

torchvision/prototype/features/_pixels.py
pmeier: > What are the semantics that we envision for `Pixels`?

I don't have anything specific in mind yet, besides that `data.shape == (N, 2)`.

> What is its format structure?

Could you elaborate what you mean?

> Can it be tied to `Keypoint`s?

Yes. `KeyPoints` would be a good subclass of `Pixels`.

pmeier: Failing tests are related to this. I'll fix them after we decide what to do here. I'm also fine with returning this as a vanilla tensor for now until we have a `Contour` feature.

fmassa: I think I would encourage returning a vanilla tensor for now, until we have a good semantic for what it should be.

One thing I would like to avoid is proliferation of custom types without clear semantics, so I would like to treat each type with utmost care and consideration

test/test_prototype_builtin_datasets.py
```diff
@@ -76,6 +77,15 @@ def test_decoding(self, dataset, mock_info):
                 f"{sequence_to_str(sorted(undecoded_features), separate_last='and ')} were not decoded."
             )
 
+    @builtin_datasets
+    def test_no_vanilla_tensors(self, dataset, mock_info):
```
fmassa, line 81: I would maybe remove this test, not sure if we want to enforce this blindly

pmeier: I think I'll relax it to test if the dataset returns at least one feature. Otherwise the dataset is rather useless for our transforms.

fmassa (approved):
Thanks!

PR #4866, opened 2021-11-05
add test split for imagenet

### LINKED ISSUE #4855: Add support of the `test` split on ImageNet
### 🚀 The feature

The new Dataset API allows returning meta-data about the images. This means that the datasets can also be used for producing predictions. As a result now it makes sense for the `ImageNet` dataset to also support the `test` split which doesn't contain labels but instead is a flat list of files.

### Motivation, pitch

See this discussion for more info: https://github.com/pytorch/vision/pull/4845#discussion_r742055625

cc @pmeier @bjuncek

pmeier:
Closes #4855.

cc @pmeier @bjuncek

### DIFF
--- a/torchvision/prototype/datasets/_builtin/imagenet.py
+++ b/torchvision/prototype/datasets/_builtin/imagenet.py
@@ -34,11 +34,17 @@ class ImageNet(Dataset):
             type=DatasetType.IMAGE,
             categories=categories,
             homepage="https://www.image-net.org/",
-            valid_options=dict(split=("train", "val")),
+            valid_options=dict(split=("train", "val", "test")),
             extra=dict(
                 wnid_to_category=FrozenMapping(zip(wnids, categories)),
                 category_to_wnid=FrozenMapping(zip(categories, wnids)),
-                sizes=FrozenMapping([(DatasetConfig(split="train"), 1281167), (DatasetConfig(split="val"), 50000)]),
+                sizes=FrozenMapping(
+                    [
+                        (DatasetConfig(split="train"), 1_281_167),
+                        (DatasetConfig(split="val"), 50_000),
+                        (DatasetConfig(split="test"), 100_000),
+                    ]
+                ),
             ),
         )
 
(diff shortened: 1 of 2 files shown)


torchvision/prototype/datasets/_builtin/imagenet.py
```diff
@@ -141,10 +148,13 @@ def _make_datapipe(
                 devkit_dp,
                 images_dp,
                 key_fn=getitem(0),
-                ref_key_fn=self._val_image_key,
+                ref_key_fn=self._val_test_image_key,
                 buffer_size=INFINITE_BUFFER_SIZE,
             )
             dp = Mapper(dp, self._collate_val_data)
+        else:  # config.split == "test"
+            dp = Shuffler(images_dp)
```
datumbox, line 156: No `buffer_size=INFINITE_BUFFER_SIZE` needed here?

torchvision/prototype/datasets/_builtin/imagenet.py
```diff
@@ -108,7 +115,7 @@ def _collate_and_decode_sample(
         return dict(
             path=path,
             image=decoder(buffer) if decoder else buffer,
-            label=torch.tensor(label),
```
datumbox, line 111: This is probably an unrelated change for the target of the PR. Could you explain why we keep it as integer instead of a tensor?

datumbox (approved):
@pmeier Looks good to me. Just 2 questions, see below:

torchvision/prototype/datasets/_builtin/imagenet.py
pmeier: No, that is needed here. Good catch!

torchvision/prototype/datasets/_builtin/imagenet.py
pmeier: For the test split `label is None` and so `torch.tensor(label)` would fail. This will properly resolved by #4864 after which we return `Optional[Label]`.

PR #4867, opened 2021-11-05
disable weight download and state dict loading for model tests

### LINKED ISSUE #4660: The `test_detection_model_trainable_backbone_layers` test shouldn't download the pretrained_backbone weights
### Feature Improvement

The `test_detection_model_trainable_backbone_layers` test currently downloads the weights of the backbone:
https://github.com/pytorch/vision/blob/6530546080148cd5d9533b50a8258f12cb050c82/test/test_models.py#L786

Setting the value `pretrained_backbone=True` is necessary because the number of trainable layers depends on this value. Unfortunately downloading pre-trained weights can lead to flakiness and slow tests and should be avoided. Until we setup a cache to store the weights locally on the CI, we should find a way to skip the actual downloading of weights during the test execution.

cc @datumbox @pmeier

pmeier:
Fixes #4660.

cc @datumbox @pmeier

### DIFF
--- a/test/test_models.py
+++ b/test/test_models.py
@@ -790,7 +790,7 @@ def test_detection_model_trainable_backbone_layers(model_fn):
         for trainable_layers in range(0, max_trainable + 1):
             model = model_fn(pretrained=False, pretrained_backbone=True, trainable_backbone_layers=trainable_layers)
 
-        n_trainable_params.append(len([p for p in model.parameters() if p.requires_grad]))
+            n_trainable_params.append(len([p for p in model.parameters() if p.requires_grad]))
     assert n_trainable_params == _model_tests_values[model_name]["n_trn_params_per_layer"]
 
 
--- a/test/test_models.py
+++ b/test/test_models.py
@@ -28,30 +28,29 @@ def get_models_from_module(module):
 @pytest.fixture
 def disable_weight_loading(mocker):
     starting_point = models
-    python_module_names = {
-        info.name for info in pkgutil.walk_packages(starting_point.__path__, f"{starting_point.__name__}.")
-    }
+    module_names = {info.name for info in pkgutil.walk_packages(starting_point.__path__, f"{starting_point.__name__}.")}
 
     function_name = "load_state_dict_from_url"
     method_name = "load_state_dict"
     targets = {f"torchvision._internally_replaced_utils.{function_name}", f"torch.nn.Module.{method_name}"}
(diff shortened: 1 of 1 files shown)


datumbox (changes requested):
As discussed offline, it seems it might be still downloading the weights. Worth confirming though! Marking as "request changes" to give time for more investigations.

pmeier:
5t2xc9

test/test_models.py
```diff
@@ -23,6 +25,37 @@ def get_models_from_module(module):
...
+        targets.update(
+            {
+                f"{python_module.__name__}.{obj.__name__}.{method_name}"
+                for obj in python_module.__dict__.values()
+                if isinstance(obj, type) and issubclass(obj, nn.Module) and method_name in obj.__dict__
+            }
+        )
+
+    for target in targets:
+        with contextlib.suppress(AttributeError):
```
pmeier, line 55: There is (atleast) one situation where this patching doesn't work. We have the module `torchvision.models.alexnet` that defines the function `alexnet`, which is then imported into the `torchvision.models` namespace. Thus, after the import is complete, we can no longer access the module through `torchvision.models.alexnet`.

This is an anti-pattern and should be avoided. Given that everything we expose everything under `torchvision.models`, I suggest making all the other modules private. This avoids this naming issue and gives us the freedom to change the underlying structure however we want without needing to keep BC.

datumbox: Yeap, it is what it is. Not sure many other models have the same issue but I don't think it's worth breaking BC over it. Alexnet is in for historical reasons and unlikely to be used as a backbone for any modern architecture.

datumbox (approved):
LGTM, thanks a lot @pmeier 

I've tested this and it works locally.

test/test_models.py
```diff
@@ -23,6 +25,36 @@ def get_models_from_module(module):
     return [v for k, v in module.__dict__.items() if callable(v) and k[0].lower() == k[0] and k[0] != "_"]
 
 
+@pytest.fixture
+def disable_weight_loading(mocker):
```
NicolasHug, line 29: Thanks @pmeier 

Considering this is a fairly complex logic, it might be useful to future us to add a few comments about what this fixture is doing, and most importantly *why* it's doing it the way it's doing it.

NicolasHug: Thank you, great docs :)

PR #4868, opened 2021-11-05
Support big endian for .flo file decoding

NicolasHug:
From https://vision.middlebury.edu/flow/code/flow-code/README.txt,
.flo files contain a magic number which is to 'PIEH' in ascii.

Our previous code was relying on the expectation that this is equivalent to `202021.25` on little endian archs. This PR updates the magic-number checks to also support  big-endian archs.

cc @pmeier


fmassa (approved):
Thanks!

PR #4870, opened 2021-11-05
Fix .flo file decoding

NicolasHug:
I previously made a mistake in the reshaping which led to wrong values - the tests was wrong as well, so it wasn't caught. Both are now fixed.

cc @pmeier

### DIFF
--- a/test/test_datasets.py
+++ b/test/test_datasets.py
@@ -1914,11 +1914,13 @@ class SintelTestCase(datasets_utils.ImageDatasetTestCase):
 
     def test_flow(self):
         # Make sure flow exists for train split, and make sure there are as many flow values as (pairs of) images
+        h, w = self.FLOW_H, self.FLOW_W
+        expected_flow = np.arange(2 * h * w).reshape(h, w, 2).transpose(2, 0, 1)
         with self.create_dataset(split="train") as (dataset, _):
             assert dataset._flow_list and len(dataset._flow_list) == len(dataset._image_list)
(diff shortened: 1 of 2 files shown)


test/test_datasets.py
```diff
@@ -2095,11 +2100,16 @@ def inject_fake_data(self, tmpdir, config):
 
     @datasets_utils.test_all_configs
     def test_flow(self, config):
+        h, w = self.FLOW_H, self.FLOW_W
+        expected_flow = np.arange(3 * h * w).reshape(h, w, 3).transpose(2, 0, 1)
+        expected_flow = np.flip(expected_flow, axis=1)
+        expected_flow = expected_flow[:2, :, :]
```
NicolasHug, line 2106: This is unrelated to the fix, just a slightly more robust way to check the .pfm file decoding. I decided to disagree with myself and my previous comment

> We don't check the values because the reshaping and flipping makes it hard to figure out

vadimkantorov:
Is there flo visualization added to pytorch utils as well?

NicolasHug:
There isn't at the moment, but it could be interesting to add something that converts flow files into a pil/tensor image. Do you know how one usually maps a displacement value into a color map?

vadimkantorov:
I think one first maps angle and magnitude to HSV coordinates and then converts to RGB. These utils should be included in many public optical flow repos, like RAFT

jdsgomes (approved):
stamping

PR #4876, opened 2021-11-05
Adding interpolation in meta for all models and cleaning up unused vars

datumbox:
Fix partially #4652

Minor clean ups on all existing models:
- Adding interpolation info in all missing meta.
- Removing unused `arch` var in VGG.

cc @datumbox @bjuncek


kazhang (approved):
LGTM, thanks!

PR #4879, opened 2021-11-05
Remove unused import os in datasets

kazhang:
os module is not actually used in these datasets

cc @pmeier


NicolasHug (approved):
Thanks @kazhang

pmeier:
I'm wondering why `flake8` hasn't picked up on that. Probably because we also import `os.path`?

PR #4882, opened 2021-11-08
add prototype utilities to read arbitrary numeric binary files

pmeier:
This is not needed right now, but helps @krshrimali who is porting the Sintel "legacy" dataset to prototypes. This has quite a few parallels to what we do for MNIST:

https://github.com/pytorch/vision/blob/6d9a42c322cb815516d5ea556b751d0c7e767c7f/torchvision/prototype/datasets/_builtin/mnist.py#L42

We can probably merge these two in the future.

cc @pmeier @bjuncek


torchvision/prototype/datasets/__init__.py
```diff
@@ -7,7 +7,7 @@
         "Note that you cannot install it with `pip install torchdata`, since this is another package."
     ) from error
 
-from . import decoder, utils
+from . import decoder, utils, datapipes
```
pmeier, line 10: I've added a new module for this, because this datapipe probably shouldn't be an internal one since it is a useful building block if users want to construct their own dataset.

torchvision/prototype/datasets/datapipes.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class FloReader(IterDataPipe[torch.Tensor]):
+    def __init__(self, datapipe: IterDataPipe[Tuple[str, IO]]) -> None:
+        self.datapipe = datapipe
+
+    def _read_data(self, file: IO, *, dtype: torch.dtype, count: int) -> torch.Tensor:
+        num_bytes_per_value = (torch.finfo if dtype.is_floating_point else torch.iinfo)(dtype).bits // 8
+        chunk_size = count * num_bytes_per_value
+        return torch.frombuffer(bytearray(file.read(chunk_size)), dtype=dtype)
```
NicolasHug, line 16: The value are encoded as little endian in the file - will it be properly decoded on big endian architectures
?

torchvision/prototype/datasets/datapipes.py
```diff
@@ -0,0 +1,29 @@
...
+
+    def _read_scalar(self, file: IO, *, dtype: torch.dtype) -> Union[int, float]:
+        return self._read_data(file, dtype=dtype, count=1).item()
+
+    def __iter__(self) -> Iterator[torch.Tensor]:
+        for _, file in self.datapipe:
+            if self._read_scalar(file, dtype=torch.float32) != 202021.25:
```
NicolasHug, line 23: Might be worth keeping this consistent with the other part https://github.com/pytorch/vision/pull/4868

torchvision/prototype/datasets/datapipes.py
```diff
@@ -0,0 +1,29 @@
...
+from torchdata.datapipes.iter import IterDataPipe
+
+__all__ = ["FloReader"]
+
+
+class FloReader(IterDataPipe[torch.Tensor]):
+    def __init__(self, datapipe: IterDataPipe[Tuple[str, IO]]) -> None:
+        self.datapipe = datapipe
+
+    def _read_data(self, file: IO, *, dtype: torch.dtype, count: int) -> torch.Tensor:
```
NicolasHug, line 13: IMHO this helper is not really necessary, same for `_read_scalar`. The original code is fairly short and having 1 or 2 levels of indirection when reading it makes it harder to follow

torchvision/prototype/datasets/datapipes.py
pmeier: No, I need to handle that. Thanks for the heads-up!

torchvision/prototype/datasets/datapipes.py
```diff
@@ -0,0 +1,29 @@
...
+
+    def __iter__(self) -> Iterator[torch.Tensor]:
+        for _, file in self.datapipe:
+            if self._read_scalar(file, dtype=torch.float32) != 202021.25:
+                raise ValueError("Magic number incorrect. Invalid .flo file")
+
+            width = cast(int, self._read_scalar(file, dtype=torch.int32))
+            height = cast(int, self._read_scalar(file, dtype=torch.int32))
+
+            yield self._read_data(file, dtype=torch.float32, count=2 * height * width).reshape((2, height, width))
```
NicolasHug, line 29: This is incorrect (my fault), it should be

`reshape(height, weight, 2).transpose(2, 0, 1)`

I have a PR to fix the original code: https://github.com/pytorch/vision/pull/4870/files

torchvision/prototype/datasets/datapipes.py
```diff
@@ -0,0 +1,29 @@
...
+
+    def _read_scalar(self, file: IO, *, dtype: torch.dtype) -> Union[int, float]:
+        return self._read_data(file, dtype=dtype, count=1).item()
+
+    def __iter__(self) -> Iterator[torch.Tensor]:
+        for _, file in self.datapipe:
+            if self._read_scalar(file, dtype=torch.float32) != 202021.25:
+                raise ValueError("Magic number incorrect. Invalid .flo file")
+
+            width = cast(int, self._read_scalar(file, dtype=torch.int32))
```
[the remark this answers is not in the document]
pmeier: Yes, that would also work. `cast(int, ...)` is a no-op at runtime, but probably so `int(...)` if the argument is already an `int`.

torchvision/prototype/datasets/datapipes.py
pmeier: I can get onboard with the `_read_scalar`, but I feel like `_read_data` is quite useful. It isn't a one-liner and especially with the endian handling will get even more logic.

torchvision/prototype/datasets/datapipes.py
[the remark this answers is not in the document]
NicolasHug: We probably don't need to worry too much about the computation cost of calling `int` twice on a ML script :p

torchvision/prototype/datasets/datapipes.py
NicolasHug: BTW IDK if this is intended but we decided in the original PR that these would be numpy array
I assume eveything is supposed to be a tensor in datapipes?

torchvision/prototype/datasets/datapipes.py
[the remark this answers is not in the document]
NicolasHug: OK.
Sorry, these things really make me sad. Half of the code is just annotations or extra bits to appease mypy. Things are getting hard to follow for me.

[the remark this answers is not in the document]
pmeier: You definitively have a point. Maybe I'm to "used" to stuff like that now, but if you find it harder to follow, than simply go with your suggestion. Readability should be value higher than `mypy` compatibility where ever possible. We have discussed this at length in the past and there are situations where `mypy` complaining is just the symptom of spaghetti code. In those situations, we should refactor instead of silencing it. But this is not one of these situations.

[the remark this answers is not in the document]
NicolasHug: TBH I would prefer your original code and use `cast` instead of `int` for the reasons you mentioned above.

It also makes it clear that it's `mypy` being a PITA again.

I'm just frustrated to see all these code obfuscations for benefits that have never been clear to me.

torchvision/prototype/datasets/utils/_internal.py
```diff
@@ -335,3 +340,45 @@ def _make_sharded_datapipe(root: str, dataset_size: int) -> IterDataPipe:
...
+        self._reverse = byte_order != sys.byteorder
+
+    def _compute_params(self, dtype: torch.dtype, shape: Sequence[int]) -> Tuple[int, bool]:
+        num_bytes_per_value = (torch.finfo if dtype.is_floating_point else torch.iinfo)(dtype).bits // 8
+        num_values = prod(shape) if shape else 1
+        chunk_size = num_bytes_per_value * num_values
+        reverse = num_bytes_per_value > 1 and self._reverse
+        return chunk_size, reverse
+
+    def read(self, dtype: torch.dtype, *, shape: Sequence[int] = ()) -> torch.Tensor:
```
NicolasHug, line 360: Do we ever need to pass a tuple as the shape?
That's a bit unexpected IMO.

Even if one asks for a shape of (N, M), there's not guarantee that the underlying layout is adequate.

pmeier: > Do we ever need to pass a tuple as the shape?

No, but it is quite convenient. Otherwise we need to manually compute the number of bytes to be read and reshape the results afterwards. This way, this can happen automatically. 

> Even if one asks for a shape of (N, M), there's not guarantee that the underlying layout is adequate.

True, you have use the "right" format for the data at hand. But you have to do it anyway at some point:

https://github.com/pytorch/vision/blob/0c68a6d0bf34ff3e526207b3e083e2384394ea0d/torchvision/datasets/_optical_flow.py#L378-L379

vs

```python
return reader.read(torch.float32, shape=(height, width, 2)).permute((2, 0, 1))
```

torchvision/prototype/datasets/utils/_internal.py
```diff
@@ -335,3 +340,45 @@ def _make_sharded_datapipe(root: str, dataset_size: int) -> IterDataPipe:
...
+        chunk_size, reverse = self._compute_params(dtype, shape)
+        # As is, the chunk we read is not writeable, because it is read from a file and not from memory. Thus, we copy
+        # here to a bytearray in order to avoid the warning that torch.frombuffer would emit otherwise. This also
+        # enables inplace operations on the contents, which would otherwise fail.
+        chunk = bytearray(self._file.read(chunk_size))
```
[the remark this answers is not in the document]
pmeier: > Does this induce an extra copy for all .flo images, compared to the current implementation?

Yes.

> If yes, this isn't negligible. It's OK for the MNIST dataset because we only read the file once IIRC, but it's not the same for flo target images

Why do we need to read a `.flo` file multiple times?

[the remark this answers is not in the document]
NicolasHug: Flow files are read as we need them so thousands files are read per training stage

torchvision/prototype/datasets/utils/_internal.py
```diff
@@ -335,3 +340,47 @@ def _make_sharded_datapipe(root: str, dataset_size: int) -> IterDataPipe:
...
+
+prod = functools.partial(functools.reduce, operator.mul)
+
+
+class NumericBinaryReader:
+    def __init__(self, file: IO, *, byte_order: str = sys.byteorder) -> None:
+        self._file = file
+        # torch.frombuffer interprets the bytes in the same byte order as the system. Thus, if the data is stored in
+        # the opposite byte order, we need to reverse the bytes before feeding them to torch.frombuffer().
+        self._reverse = byte_order != sys.byteorder
```
NicolasHug, line 353: Is there a way to let pytorch do that logic for us instead of having to manually take care of endianness? In numpy one can just add `<` or `>` to the dtype specification to indicate what is the expected encoding and everything gets taken care of

pmeier: Nope. See #4598 and #4651 for discussions about this.

NicolasHug: Then could we just rely on numpy for the reading part, since it's much more convenient (and possibly faster if one needs to account for endianness)?

pmeier:
After some more offline discussion with @NicolasHug, I've refactored the PR to use `numpy` for reading the data since it natively is able to handle the endianness. 

Note that we **cannot** use `numpy.fromfile`, because it [doesn't support reading from compressed files](https://github.com/numpy/numpy/issues/1103) like the ones the MNIST dataset uses:

```python
import gzip
import tempfile

import numpy as np

data = np.array([0.0])

path = tempfile.mktemp()
with open(path, "wb") as file:
    file.write(gzip.compress(data.tobytes()))


with open(path, "rb") as compressed_file:
    with gzip.open(compressed_file, "rb") as file:
        print(np.fromfile(file))

        file.seek(0)
        print(np.frombuffer(file.read()))
```
```
[7.1653857e+161 5.9808143e+197]
[0.]
```

If we go through with this design, we should revisit #4598. It probably than should be reverted for consistency. cc @datumbox

torchvision/prototype/datasets/utils/_internal.py
pmeier: See https://github.com/pytorch/vision/pull/4882#issuecomment-963523041

NicolasHug:
My main concern wasn't just about using numpy vs pytorch it was mostly about code complexity. The original flow decoding function is extremely simple, seld-contained, and easy to understand: 

https://github.com/pytorch/vision/blob/bbd9ff8fb936846aa0412996abab19b563677e5b/torchvision/datasets/_optical_flow.py#L365-L379

I personally see no strong reason to create a new class or a new helper that will add maintenance overhead - we don't need a helper for anything. I would just copy paste it and be done with it.

> If we go through with this design, we should revisit #4598. It probably than should be reverted for consistency

Not necessarily, I don't think we need to worry about consistency between the use of `np.frombuffer` vs `torch.frombuffer`. We should just use which tool gets the best job done, while accounting for factors like code simplicity (i.e. maintainability) / robustness.

pmeier:
> My main concern wasn't just about using numpy vs pytorch it was mostly about code complexity. The original flow decoding function is extremely simple, seld-contained, and easy to understand

That is true with two caveats:

1. We can't use PyTorch dtypes to indicate the datatypes. Before diving into this, I had no idea what `"<f4"` meant. IMO, using `dtype=torch.float32, byte_order="little"` is a lot more expressive and does not require you to look up documentation.
2. Your implementation assumes that everything stays as `numpy.ndarray` and thus does not need to care how the data is stored in memory. As soon as you want to convert the result into a `torch.Tensor` your system also needs to be little endian. In other words: if I'm on a big-endian system, your implementation is broken without manually fixing the byte order:
    
    ```python
    >>> torch.from_numpy(np.array([0.0], dtype=np.dtype(">f4")))
    ValueError: given numpy array has byte order different from the native byte order. Conversion between byte orders is currently not supported.
    ```
    
    (This example uses the inverted logic, i.e. having big endian data on a little endian system, but it will be the same the other way around)

As for simplicity, my implementation is even shorter and hides all of the complexity above from the user. Plus it is as self-contained as your version in the sense that both implementations let another function do the heavy lifting for them.

> Not necessarily, I don't think we need to worry about consistency between the use of np.frombuffer vs torch.frombuffer. We should just use which tool gets the best job done, while accounting for factors like code simplicity (i.e. maintainability) / robustness.

#4598 was specifically about using `torch.frombuffer` over `np.frombuffer` in a situation where endianness matters. Thus, the new implementation is more complex in favor of using a PyTorch native approach. If we go by your sentiment, we should revert the PR.

NicolasHug:
I guess I should have  started with this: why are we implementing a new version of the MNIST thing and `_read_flow` in the prototype folder?
Both current implementations work fine (module a minor bug in _read_flo as we previouly discussed). If we merge this, we'll have 2 implementations for both functions - surely we want to avoid that kind of duplication.

Shouldn't we have this refactoring discussion outside of the prototype concerns?

pmeier:
> I guess I should have started with this: why are we implementing a new version of the MNIST thing and `_read_flow` in the prototype folder?

For MNIST the functionality we had always reads the complete file whereas now we only read one image / label at the time. This is why we need a custom datapipe.

This is not necessary for `read_flo` and thus it is just a function. The difference between the current implementation and this PR is that we also need to be able to pass a file handle directly instead of a path. I'll push a commit shortly that shows the changes to the current `_read_flo` that we need to make in order for it to also work with the prototype datasets.

> Shouldn't we have this refactoring discussion outside of the prototype concerns?

I'm also fine with that. Note however for an implementation that is also used by the prototype datasets, the functionality need to be able to handle open files instead of just paths and should return a tensor (we could also have a thin wrapper for that).

torchvision/prototype/datasets/utils/_internal.py
```diff
@@ -250,3 +255,46 @@ def _make_sharded_datapipe(root: str, dataset_size: int) -> IterDataPipe:
...
+    """
+    byteorder = "<" if byte_order == "little" else ">"
+    char = "f" if dtype.is_floating_point else ("i" if dtype.is_signed else "u")
+    itemsize = (torch.finfo if dtype.is_floating_point else torch.iinfo)(dtype).bits // 8
+    np_dtype = byteorder + char + str(itemsize)
+
+    buffer = file.read(-1 if count == -1 else count * itemsize)
+    # torch.frombuffer can only deal with with the native byte order,
+    # so we use numpy for the I/O and convert to a tensor.
+    return torch.from_numpy(np.frombuffer(buffer, dtype=np_dtype).astype(np_dtype[1:]))
```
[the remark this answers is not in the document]
NicolasHug: Also, don't we want to pass `copy=False`?

NicolasHug (commented):
Thanks @pmeier , I made some comments below.

Before merging I think it would be nice to have some tests to ensure that the new implementation of `read_flow` is consistent with the current one.

I would also be curious if you could run some quick benchmark to see if there's a significant time difference between the 2.

torchvision/prototype/datasets/utils/_internal.py
```diff
@@ -250,3 +255,46 @@ def _make_sharded_datapipe(root: str, dataset_size: int) -> IterDataPipe:
...
+
+        This function is similar to :func:`numpy.fromfile` with two notable differences:
+
+        1. This function only accepts an open binary file, but not a path to it.
+        2. This function has an additional ``byte_order`` parameter, since PyTorch's ``dtype``'s do not support that
+            concept.
+
+    Args:
+        file (IO): Open binary file.
+        dtype (torch.dtype): Data type of the returned tensor.
```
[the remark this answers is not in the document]
pmeier: > isn't this the dtype of the underlying data as well?

Yes. I can adapt the description to reflect that.

> Also what happens if we happen to need to read uint16?

Can't be done with this function. But since PyTorch doesn't support that, we can't have a tensor with that `dtype` anyway.

torchvision/prototype/datasets/utils/_internal.py
```diff
@@ -250,3 +255,46 @@ def _make_sharded_datapipe(root: str, dataset_size: int) -> IterDataPipe:
...
+        byte_order (str): Byte order of the data. Can be ``"little"`` or ``"big"`` endian.
+        count (int): Number of values of the returned tensor. If ``-1`` (default), will read the complete file.
+    """
+    byteorder = "<" if byte_order == "little" else ">"
```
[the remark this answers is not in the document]
pmeier:
```python
import numpy as np

dtype = np.dtype(">f4")

assert dtype.byteorder == ">"
assert dtype.char == "f"
assert dtype.itemsize == 4
```

torchvision/prototype/datasets/utils/_internal.py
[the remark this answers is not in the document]
pmeier: > Could you explain what the `astype(np_dtype[1:])` call does, and why it's needed?

`np.ndarray.astype()` converts the array to the given type. `np_dtype` is a `str` so `np_dtype[1:] == char + str(itemsize)`. The complete call converts the array from an arbitrary endianness to native one, so that it can be read by `torch.frombuffer`. 

> Also, don't we want to pass `copy=False`?

In theory, yes, but in the current implementation, no. `torch.frombuffer` will scream at you if the underlying memory is read-only and without a copy, it is. I'm looking into ways to read from the file and store the data directly in mutable memory.

torchvision/prototype/datasets/utils/_internal.py
```diff
@@ -250,3 +255,46 @@ def _make_sharded_datapipe(root: str, dataset_size: int) -> IterDataPipe:
...
+    # torch.frombuffer can only deal with with the native byte order,
+    # so we use numpy for the I/O and convert to a tensor.
+    return torch.from_numpy(np.frombuffer(buffer, dtype=np_dtype).astype(np_dtype[1:]))
+
+
+def read_flo(file: BinaryIO) -> torch.Tensor:
+    if file.read(4) != b"PIEH":
+        raise ValueError("Magic number incorrect. Invalid .flo file")
+
+    width, height = fromfile(file, dtype=torch.int32, byte_order="little", count=2)
```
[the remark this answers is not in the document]
pmeier: AFAIK, byte order is foreign to PyTorch. Thus, there is no reference to go on. In general though, PyTorch's policy is to side with Python over `numpy` in case they disagree. In Python [`sys.byteorder`](https://docs.python.org/3/library/sys.html#sys.byteorder) is either `"little"` or `"big"`. Plus, although more verbose, I find that more expressive than using `">"` and `"<"`.

torchvision/prototype/datasets/utils/_internal.py
```diff
@@ -250,3 +255,46 @@ def _make_sharded_datapipe(root: str, dataset_size: int) -> IterDataPipe:
...
+        This function is similar to :func:`numpy.fromfile` with two notable differences:
+
+        1. This function only accepts an open binary file, but not a path to it.
+        2. This function has an additional ``byte_order`` parameter, since PyTorch's ``dtype``'s do not support that
+            concept.
+
+    Args:
+        file (IO): Open binary file.
+        dtype (torch.dtype): Data type of the returned tensor.
+        byte_order (str): Byte order of the data. Can be ``"little"`` or ``"big"`` endian.
```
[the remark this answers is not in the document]
pmeier: Agreed, they make the raw docstring harder to read. But for the rendered documentation, they are not redundant. Double backticks will render as inline code. Without, the possible options will be rendered as normal text.

Given that this (at least for now) is not part of the documentation, I'm ok with removing them.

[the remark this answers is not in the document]
NicolasHug: Even for rendered doc I believe they are unnecessary and bloat the text. We use backticks to separate text from code-like things. When the code-like think is in quotes, the separation is already obvious. Compare e.g.

image

with

image

Although it's not necessarily the case here, sometimes when there are a lot of back-quoted strings, the text becomes hard to read.

torchvision/prototype/datasets/utils/_internal.py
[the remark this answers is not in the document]
NicolasHug: Surely we can still declare `byte_order` and `item_size` variables in our code?

torchvision/prototype/datasets/utils/_internal.py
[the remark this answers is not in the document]
NicolasHug: > The complete call converts the array from an arbitrary endianness to native one, so that it can be read by torch.frombuffer.

Thanks, it'd be great to document this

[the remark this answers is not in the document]
pmeier: A solution to avoid the memcopy is to use a [`mmap`]():

```python
import mmap

import torch

data = torch.rand(5)
path = "data.bin"

data.numpy().tofile(path)

with open(path, "r+b") as file:
    buffer = mmap.mmap(file.fileno(), 0)
    restored_data = torch.frombuffer(buffer, dtype=data.dtype)

assert torch.all(restored_data == data)
```

The only problem with that is the file has to be opened in read / write mode (`"r+"`) and their is currently no interface to do that with datapipes. I'll benchmark first and see if this something that could be helpful.

pmeier:
> I would also be curious if you could run some quick benchmark to see if there's a significant time difference between the 2.

```python
import contextlib
import pathlib
import time

import torch
from torchvision.datasets._optical_flow import _read_flo as read_flo_baseline
from torchvision.prototype.datasets.utils._internal import read_flo as read_flo_new


files = [
    file
    for file in (pathlib.Path(datasets.home()) / "sintel" / "training" / "flow").glob("**/*")
    if file.suffix == ".flo"
]


@contextlib.contextmanager
def timeit(label):
    start = time.perf_counter()
    yield
    stop = time.perf_counter()
    print(label, f"{(stop - start) / len(files) * 1e3:.1f} milliseconds per file")


with timeit("baseline"):
    for file in files:
        torch.from_numpy(read_flo_baseline(file).astype("f4", copy=False))

with timeit("baseline+copy"):
    for file in files:
        torch.from_numpy(read_flo_baseline(file).astype("f4"))

with timeit("new"):
    for file in files:
        with open(file, "r+b") as f:
            read_flo_new(f)
```

```
baseline 0.3 milliseconds per file
baseline+copy 2.2 milliseconds per file
new 2.3 milliseconds per file
```

So roughly 8x slower than before, but in absolute terms probably insignificant within the image training loop. The memcopy is the offender. If we add this to the old implementation, the difference is negligible.

torchvision/prototype/datasets/utils/_internal.py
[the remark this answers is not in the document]
pmeier: See https://github.com/pytorch/vision/pull/4882#issuecomment-971674863.

pmeier:
I've added support for dropping the memcopy if the file is opened in update mode (`r+b` or `w+b`). This is even faster than the original solution. In addition, calling `bytearray()` on the read bytes if we are not in update mode is faster than letting `numpy` do the copy. With the latest commit and this addition

```python
with timeit("new+update"):
    for file in files:
        with open(file, "r+b") as f:
            read_flo_new(f)
```

the benchmark script now gives:

```
baseline 0.3 milliseconds per file
baseline+copy 2.1 milliseconds per file
new 1.3 milliseconds per file
new+update 0.1 milliseconds per file
```

torchvision/prototype/datasets/utils/_internal.py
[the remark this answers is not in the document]
pmeier: > Even for rendered doc I believe they are unnecessary and bloat the text.

Not sure about your definition of "bloat", but I don't see how they bloat anything. Character count is the same.

> We use backticks to separate text from code-like things.

Exactly. Possible options for a parameter are code-like things. Quotation marks serve a very different purpose in text than in code. The quotation marks in the sentence in question make little sense in the text domain. I don't see a reason to treat strings here any other than numbers or `None`.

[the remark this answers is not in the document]
NicolasHug: > I don't see a reason to treat strings here any other than numbers or None

Because, again: strings are already delimited with the quote delimiter, and adding 2 delimiters makes it less smooth to read.

Yes, you're right, going by the book would tell us to surround strings in double backticks as well. But that's not necessary here and from my experience makes it harder to read the docstring - both in the code and in the rendered html. Anyway, I won't insist too much on this. I'll probably just end up fixing it in another PR eventually when I stumble on the docs

torchvision/prototype/datasets/utils/_internal.py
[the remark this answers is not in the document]
NicolasHug: Sorry, unresolving: please avoid changing `byte_order` into `byteorder`

torchvision/prototype/datasets/utils/_internal.py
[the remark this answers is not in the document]
pmeier: It is already removed.

test/test_prototype_datasets_utils.py
```diff
@@ -0,0 +1,19 @@
...
+import torch
+from datasets_utils import make_fake_flo_file
+from torchvision.datasets._optical_flow import _read_flo as read_flo_ref
+from torchvision.prototype.datasets.utils._internal import read_flo
+
+
+@pytest.mark.parametrize("mode", ("rb", "r+b"))
+def test_read_flo(tmpdir, mode):
+    path = tmpdir / "test.flo"
+    height, width = torch.randint(3, 10, (2,))
```
NicolasHug, line 11: Thanks for the test.
Could we just hardcode h and w?
We should try to avoid RNG without setting the seed in general, but here it seems a bit unnecessary at first sight.

torchvision/prototype/datasets/utils/_internal.py
```diff
@@ -250,3 +256,57 @@ def _make_sharded_datapipe(root: str, dataset_size: int) -> IterDataPipe:
...
+
+        This function is similar to :func:`numpy.fromfile` with two notable differences:
+
+        1. This function only accepts an open binary file, but not a path to it.
+        2. This function has an additional ``byte_order`` parameter, since PyTorch's ``dtype``'s do not support that
+            concept.
+
+    Args:
+        file (IO): Open binary file.
+        dtype (torch.dtype): Data type of the underlying data.
```
NicolasHug, line 280:
```suggestion
        dtype (torch.dtype): Data type of the underlying data. This will also be the dtype of the returned tensor.
```

torchvision/prototype/datasets/utils/_internal.py
```diff
@@ -250,3 +256,57 @@ def _make_sharded_datapipe(root: str, dataset_size: int) -> IterDataPipe:
...
+    """
+    byteorder = "<" if byte_order == "little" else ">"
+    char = "f" if dtype.is_floating_point else ("i" if dtype.is_signed else "u")
+    itemsize = (torch.finfo if dtype.is_floating_point else torch.iinfo)(dtype).bits // 8
+    np_dtype = byteorder + char + str(itemsize)
+
+    # PyTorch does not support tensors with underlying read-only memory. If the file was opened for updating, i.e.
+    # 'r+b' or 'w+b', the memory is already writable. Otherwise we need to copy it to a mutable location after reading.
+    buffer: Union[memoryview, bytearray]
+    try:
```
NicolasHug, line 292: Should we use an if/else instead of a try/except? I understand EAFP but with an `if`  it would be more obvious and more explicit that both branches depend on whether the file was opened with `+` or not.

torchvision/prototype/datasets/utils/_internal.py
```diff
@@ -250,3 +256,57 @@ def _make_sharded_datapipe(root: str, dataset_size: int) -> IterDataPipe:
     # dp = dp.cycle(2)
     dp = TakerDataPipe(dp, dataset_size)
     return dp
+
+
+def fromfile(
```
NicolasHug, line 261: It would be great to test this util a bit, e.g. against various equivalent parametrization of `numpy.fromfile`.

One important thing to test seems to be that we never get the "non-writable buffer" warning from pytorch.

torchvision/prototype/datasets/utils/_internal.py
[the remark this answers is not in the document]
pmeier: Oops, I thought I did that.

torchvision/prototype/datasets/utils/_internal.py
```diff
@@ -250,3 +256,57 @@ def _make_sharded_datapipe(root: str, dataset_size: int) -> IterDataPipe:
...
+    .. note::
+
+        This function is similar to :func:`numpy.fromfile` with two notable differences:
+
+        1. This function only accepts an open binary file, but not a path to it.
+        2. This function has an additional ``byte_order`` parameter, since PyTorch's ``dtype``'s do not support that
+            concept.
+
+    Args:
+        file (IO): Open binary file.
```
NicolasHug, line 279: IIUC opening the file with `+` is much faster than not? Maybe this could be documented here

test/test_prototype_datasets_utils.py
```diff
@@ -0,0 +1,46 @@
...
+        ("<f8", torch.float64, "little"),
+        ("<i4", torch.int32, "little"),
+        (">i8", torch.int64, "big"),
+        ("|u1", torch.uint8, sys.byteorder),
+    ],
+)
+def test_fromfile(tmpdir, np_dtype, torch_dtype, byte_order, mode):
+    path = tmpdir / "data.bin"
+    count = 5
+    np.random.randn(count).astype(np_dtype).tofile(path)
```
NicolasHug, line 26: we should avoid leaking numpy RNG as well: in other places we use

```py
rng = np.random.RandomState(0)
rng.randn(count)...
```

test/test_prototype_datasets_utils.py
```diff
@@ -0,0 +1,46 @@
...
+
+import numpy as np
+import pytest
+import torch
+from datasets_utils import make_fake_flo_file
+from torchvision.datasets._optical_flow import _read_flo as read_flo_ref
+from torchvision.prototype.datasets.utils._internal import read_flo, fromfile
+
+
+@pytest.mark.filterwarnings("error:The given NumPy array is not writeable:UserWarning")
```
NicolasHug, line 11: I think this will filter out the warnings so that they don't appear when we run the tests, right?

My suggestion earlier was to make sure that the call to `fromfile` does *not* raise such a warning.

test/test_prototype_datasets_utils.py
```diff
@@ -0,0 +1,46 @@
...
+@pytest.mark.parametrize("mode", ("rb", "r+b"))
+@pytest.mark.parametrize(
+    ("np_dtype", "torch_dtype", "byte_order"),
```
NicolasHug, line 14: maybe just build byte_order from `np_dtype[0]` to simplify the parametrization?

test/test_prototype_datasets_utils.py
```diff
@@ -0,0 +1,46 @@
...
+        (">i8", torch.int64, "big"),
+        ("|u1", torch.uint8, sys.byteorder),
+    ],
+)
+def test_fromfile(tmpdir, np_dtype, torch_dtype, byte_order, mode):
+    path = tmpdir / "data.bin"
+    count = 5
+    np.random.randn(count).astype(np_dtype).tofile(path)
+
+    for count_ in (-1, count // 2):
```
NicolasHug, line 28: parametrize instead?

NicolasHug (commented):
Thanks @pmeier , some last comments for the tests, looks good otherwise

test/test_prototype_datasets_utils.py
pmeier: > I think this will filter out the warnings so that they don't appear when we run the tests, right?

Nope, that turns `UserWarning`'s that start with the text `"The given NumPy array is not writeable"` into errors. To silence them, the action would have to be `ignore` instead of `error`. See https://docs.python.org/3/library/warnings.html?highlight=warnings#the-warnings-filter for details.

test/test_prototype_datasets_utils.py
pmeier: Hm, I feel like either we don't build anything dynamically or everything, i.e. keep it as is or also build `torch_dtype`. Thoughts?

NicolasHug: let's leave as is , it's just a nit

NicolasHug (approved):
Thanks a lot for your work and patience @pmeier , approving now since the remaining comments are a bit trivial / not very important.

PR #4883, opened 2021-11-08
[FBcode->GH] Removed type annotations from rcnn

prabhat00155:
These type annotations were causing test failures in Phabricator. Hence, these were removed in D32216673.

cc @datumbox

### DIFF
--- a/torchvision/models/detection/generalized_rcnn.py
+++ b/torchvision/models/detection/generalized_rcnn.py
@@ -36,22 +36,15 @@ class GeneralizedRCNN(nn.Module):
         self._has_warned = False
 
     @torch.jit.unused
-    def eager_outputs(
-        self,
-        losses: Dict[str, Tensor],
-        detections: List[Dict[str, Tensor]],
-    ) -> Union[Dict[str, Tensor], List[Dict[str, Tensor]]]:
-
+    def eager_outputs(self, losses, detections):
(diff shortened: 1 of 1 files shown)


NicolasHug (approved):
Stamping

Do you remember what the issue was? Where the new annotations incorrect or is it something else?

prabhat00155:
> Stamping
> 
> Do you remember what the issue was? Where the new annotations incorrect or is it something else?

Error details: https://github.com/pytorch/vision/pull/4631/files#r744161323

prabhat00155:
@NicolasHug Detailed error can be viewed here: https://www.internalfb.com/diff/D32216673?dst_version_fbid=621645568869442, under `diff signals`.

oke-aditya:
Just a small doubt @prabhat00155 can we use the Python annotations which are commented ? Or is that too causing failure?

prabhat00155:
> Just a small doubt @prabhat00155 can we use the Python annotations which are commented ? Or is that too causing failure?

The type annotations in arguments in `eager_outputs()` and `forward()` was causing some pyre errors in a third-party file.
```
Unused ignore [0]: The `pyre-ignore` or `pyre-fixme` comment is not suppressing type errors, please remove it. 
```
Also, the return type was causing a test failure as discussed before.

datumbox:
@prabhat00155 I don't think this change is considered bc breaking.

PR #4888, opened 2021-11-09
Add pass_name='both' for Sintel dataset

NicolasHug:
I ended up needing this, also makes things consistent with FlyingThings3D

cc @pmeier


pmeier (approved):
LGTM, thanks! cc @krshrimali who is working on the prototype implementation of the dataset.

PR #4892, opened 2021-11-09
Add gallery example for drawing keypoints

oke-aditya:
Follow up of #4216 

~~Work in progress should be done in few hours :)~~

Open for review. Feedbacks are welcome!


datumbox:
@oke-aditya I'm not sure from where you sourced the image you are using but I recommend finding an alternative picture from the Coco keypoint dataset used to train the keypoint model.

oke-aditya:
This image was sourced from COCO dataset. Is it a concern since it displays a human image?

datumbox:
@oke-aditya It's good you got it from COCO. My concern is that it seems to depict a minor so let's be extra careful and replace it. Do you think that we can reuse the standard Grace Hopper picture we already have? This might be interesting because on your example you might need to filter out keypoints that don't appear on the picture (and thus it's going to be a more realistic example of visualization). Thoughts?

oke-aditya:
Oh! So sorry for depicting a minor! I completely agree we should avoid this image. 
Very very thankful to you @datumbox for pointing this out.

My apologies to the minor person depicted (who would have grown up by today :smiley: ) who might see this PR.

That's Good idea to add an example where we need to filter out keypoints. Both based on visibility and detection threshold. 
But I feel that joining a complete skeleton is particular useful as for pose-detection / other tasks. So I will keep two examples. 
First one just joining skeletons and next one helping to filter keypoints well

datumbox:
@oke-aditya Sounds good. Let's select  an alternative picture from COCO and do 2 examples as you proposed. Thanks!

torchvision/utils.py
```diff
@@ -304,7 +304,7 @@ def draw_segmentation_masks(
 def draw_keypoints(
     image: torch.Tensor,
     keypoints: torch.Tensor,
-    connectivity: Optional[Tuple[Tuple[int, int]]] = None,
+    connectivity: Optional[List[Tuple[int, int]]] = None,
```
oke-aditya, line 307: This was a typing bug, we want connectivity to have multiple tuples in it, so making it a list.

oke-aditya, line 307: This was a typing bug, we want connectivity to have multiple tuples in it, so making it a list.

test/test_utils.py
```diff
@@ -256,7 +256,14 @@ def test_draw_keypoints_vanilla():
 
     img = torch.full((3, 100, 100), 0, dtype=torch.uint8)
     img_cp = img.clone()
-    result = utils.draw_keypoints(img, keypoints, colors="red", connectivity=((0, 1),))
+    result = utils.draw_keypoints(
```
oke-aditya, line 259: updated the test too, uFmt has changed the stuff.

gallery/plot_visualization_utils.py
```diff
@@ -366,3 +366,89 @@ def show(imgs):
...
+#
+kpts = outputs[0]['keypoints']
+scores = outputs[0]['scores']
+
+#####################################
+# Some desc
+#
+#
+
+detect_threshold = 0.75
```
oke-aditya, line 405: Thresholding on keypoint scores.

oke-aditya (commented):
Fixing a small bug.
I'm still unable to find an image where visibility of keypoint isn't 1.000.

oke-aditya:
Absolutely baffled still that every run with `keypoint_rcnn_resnet50` with both pretrained=True and False, the visibility was always 1.00 for all the keypoints and instances. 

Tried over 15 images from coco, of not just persons, dogs, cats, surfboards. :( 

Is there really a case where visibility won't be 1 ? Can someone provide an image for such case.

datumbox:
@oke-aditya I had the same exact discussion with Francisco couple of days ago and he clarified that this is always 1. So I suggest to just ignore this column and move on.

@fmassa could you please confirm that this is the case?

gallery/plot_visualization_utils.py
```diff
@@ -366,3 +366,110 @@ def show(imgs):
...
+from torchvision.utils import draw_keypoints
+
+res = draw_keypoints(person_int, keypoints, colors="blue", radius=3)
+show(res)
+
+#####################################
+# As we see the keypoints appear as colored circles over the image.
+# The coco keypoints for a person are ordered and represent the following list.\
+
+coco_keypoints = [
```
oke-aditya, line 437: I think this will be great when people can use `metadata` from the new keypoint rcnn weights.
Right now we have to explicitly remember this.

oke-aditya: Although it isn't used in the code. I made this as a code block so that it is copy-pastable.

oke-aditya:
I have added one single example showing plotting keypoints and joining them. 
I felt that one detailed and simple example is necessary as keypoint rcnn involves little bit more processing and understanding to connect keypoints.

Let me know if an additional example such as batch of images or plotting multi instance images is needed. 
I think it will be nice to have.

gallery/plot_visualization_utils.py
datumbox: Yes agreed. We've already added it on the new API:
https://github.com/pytorch/vision/blob/9a7dc1a127255343551903b8b64c25dc2b5df2b1/torchvision/prototype/models/detection/keypoint_rcnn.py#L24

https://github.com/pytorch/vision/blob/9a7dc1a127255343551903b8b64c25dc2b5df2b1/torchvision/prototype/models/_meta.py#L1106-L1109

gallery/plot_visualization_utils.py
```diff
@@ -366,3 +366,110 @@ def show(imgs):
 # The two 'people' masks in the first image where not selected because they have
 # a lower score than the score threshold. Similarly in the second image, the
 # instance with class 15 (which corresponds to 'bench') was not selected.
+
+#####################################
+# Visualizing keypoints
+# ------------------------------
+# The :func:`~torchvision.utils.draw_keypoints` function can be used to
+# draw keypoints on images. We will see how to use it with
+# torchvision's KeypointRCNN loaded with :func:`~torchvision.models.detection.keypointrcnn_resnet50_fpn`
```
datumbox, line 375:
```suggestion
# torchvision's KeypointRCNN loaded with :func:`~torchvision.models.detection.keypointrcnn_resnet50_fpn`.
```

gallery/plot_visualization_utils.py
```diff
@@ -366,3 +366,110 @@ def show(imgs):
...
+person_float = convert_image_dtype(person_int, dtype=torch.float)
+
+model = keypointrcnn_resnet50_fpn(pretrained=True, progress=False)
+model = model.eval()
+
+outputs = model([person_float])
+print(outputs)
+
+#####################################
+# As we see the output contains a list of dictionary.
```
datumbox, line 394:
```suggestion
# As we see the output contains a list of dictionaries.
```

datumbox (commented):
Overall it looks good to me, thanks @oke-aditya.

Just a couple minor comments. Let me know what you think:

gallery/plot_visualization_utils.py
oke-aditya: I guess once the new prototype models and transforms are moved to stable, we would need to revisit the gallery examples.

NicolasHug:
Thanks @oke-aditya , I just gave a brief look to the rendered example and it looks great. The only comment I have is about the first sentence in the example which is

> This example illustrates some of the utilities that torchvision offers for visualizing images, bounding boxes, and segmentation masks.

and that could now be updated to 

> This example illustrates some of the utilities that torchvision offers for visualizing images, bounding boxes, segmentation masks, and keypoints.

Nice work! :)

oke-aditya:
> This example illustrates some of the utilities that torchvision offers for visualizing images, bounding boxes, and segmentation masks.

Nice, catch. I will also update the visualization thumbnail to include this image too.

> Nice work! :)

Thanks Nicolas :) once we finish supporting multiple colors, line colors, etc. 
I will add one small example which shows multiple instances. The utility looks cool :sunglasses:  on multiple images.

oke-aditya:
Here is the final gallery. Feel free to suggest changes :) 

https://980059-73328905-gh.circle-artifacts.com/0/docs/auto_examples/index.html

datumbox:
@oke-aditya Great job as always Oke. Thanks for contributing this and for your patience for the reviews on the original PR.

Issue #4787, opened 2021-10-29
Automated testing discovered that `image::read_png` sometimes can not be initialized on Windows


malfet:
### 🐛 Describe the bug

See https://github.com/pytorch/pytorch.github.io/runs/4042987825?check_suite_focus=true
Several Windows configuration tests of torchvision failed with `No such operator image::decode_png`

### Versions

0.11.1

cc @peterjc123 @nbcsm @guyang3532 @maxluk @gunandrose4u @mszhanyi

malfet:
I've already checked that libjpeg and libpng dlls are included both in variant of the package that succeeded and failed the test:
```
$ python -c "import torchvision;import os;import glob;print(glob.glob(f'{os.path.dirname(torchvision.__file__)}/*.dll'))"
['C:\\tools\\miniconda3\\lib\\site-packages\\torchvision\\libjpeg.dll', 'C:\\tools\\miniconda3\\lib\\site-packages\\torchvision\\libpng16.dll', 'C:\\tools\\miniconda3\\lib\\site-packages\\torchvision\\zlib.dll']
```

fmassa:
@andfoy can you try
```
import torch

torch.ops.load_library('path/to/image.so')
```

and see the error message?

andfoy:
I tried to reproduce it locally on my Windows machine, but without success. I installed everything from the pytorch-stable url, just like the CIs did. The only guess is that the path collection used by `LoadLibrary` is being modified somehow that those DLLs are not being picked up? 

![imagen](https://user-images.githubusercontent.com/1878982/139707196-34457542-f7be-4865-b26c-d9fce4b6ce1f.png)

mszhanyi:
Interesting!
Py3.8 + cu113 works. I verified it locally
https://github.com/pytorch/pytorch.github.io/runs/4042987908?check_suite_focus=true

mszhanyi:
> I tried to reproduce it locally on my Windows machine, but without success. I installed everything from the pytorch-stable url, just like the CIs did. The only guess is that the path collection used by `LoadLibrary` is being modified somehow that those DLLs are not being picked up?
> 
> ![imagen](https://user-images.githubusercontent.com/1878982/139707196-34457542-f7be-4865-b26c-d9fce4b6ce1f.png)

@andfoy it could be reproduced in windows CPU machine with cudatoolkit installed.
If your  windows machine has GPUs, you might set CUDA_VISIBLE_DEVICES="" or disable it in device manager.

[...]

PR #4893, opened 2021-11-10
[Windows] Workaround for loading bundled DLLs

malfet:
Python-3.8+ adds `add_dll_directory` call, see https://docs.python.org/3/whatsnew/3.8.html#ctypes
Simulate this behaviour on older versions of Python runtime by calling
`LoadLibraryExW` with the appropriate flags

Fixes https://github.com/pytorch/vision/issues/4787

### DIFF
--- a/torchvision/_internally_replaced_utils.py
+++ b/torchvision/_internally_replaced_utils.py
@@ -22,7 +22,7 @@ except ImportError:
     from torch.utils.model_zoo import load_url as load_state_dict_from_url  # noqa: 401
 
 
-def _get_extension_path(lib_name):
+def _get_extension_path(lib_name) -> str:
 
     lib_dir = os.path.dirname(__file__)
     if os.name == "nt":
--- a/torchvision/_internally_replaced_utils.py
+++ b/torchvision/_internally_replaced_utils.py
@@ -56,21 +56,3 @@ def _get_extension_path(lib_name: str) -> str:
         raise ImportError
 
     return ext_specs.origin
-
-
-def _load_library(lib_name: str) -> None:
-    lib_path = _get_extension_path(lib_name)
-    # On Windows Python-3.8+ has `os.add_dll_directory` call,
-    # which is called from _get_extension_path to configure dll search path
-    # Condition below adds a workaround for older versions by
-    # explicitly calling `LoadLibraryExW` with the following flags:
-    #  - LOAD_LIBRARY_SEARCH_DEFAULT_DIRS (0x1000)
-    #  - LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR (0x100)
-    if os.name == "nt" and sys.version_info < (3, 8):
-
-        _kernel32 = ctypes.WinDLL("kernel32.dll", use_last_error=True)
-        if hasattr(_kernel32, "LoadLibraryExW"):
-            _image_lib_handle = _kernel32.LoadLibraryExW(lib_path, None, 0x00001100)
-        else:
-            warn("LoadLibraryExW is missing in kernel32.dll")
-    torch.ops.load_library(lib_path)
--- a/torchvision/_internally_replaced_utils.py
+++ b/torchvision/_internally_replaced_utils.py
@@ -22,7 +22,7 @@ except ImportError:
     from torch.utils.model_zoo import load_url as load_state_dict_from_url  # noqa: 401
 
 
-def _get_extension_path(lib_name) -> str:
+def _get_extension_path(lib_name: str) -> str:
 
     lib_dir = os.path.dirname(__file__)
     if os.name == "nt":
@@ -56,3 +56,21 @@ def _get_extension_path(lib_name) -> str:
         raise ImportError
 
     return ext_specs.origin
+
+
+def _load_library(lib_name: str) -> None:
+    lib_path = _get_extension_path(lib_name)
+    # On Windows Python-3.8+ has `os.add_dll_directory` call,
+    # which is called from _get_extension_path to configure dll search path
+    # Condition below adds a workaround for older versions by
+    # explicitly calling `LoadLibraryExW` with the following flags:
+    #  - LOAD_LIBRARY_SEARCH_DEFAULT_DIRS (0x1000)
+    #  - LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR (0x100)
+    if os.name == "nt" and sys.version_info < (3, 8):
+
+        _kernel32 = ctypes.WinDLL("kernel32.dll", use_last_error=True)
+        if hasattr(_kernel32, "LoadLibraryExW"):
+            _image_lib_handle = _kernel32.LoadLibraryExW(lib_path, None, 0x00001100)
+        else:
+            warn("LoadLibraryExW is missing in kernel32.dll")
+    torch.ops.load_library(lib_path)
--- a/torchvision/_internally_replaced_utils.py
+++ b/torchvision/_internally_replaced_utils.py
@@ -22,7 +22,7 @@ except ImportError:
     from torch.utils.model_zoo import load_url as load_state_dict_from_url  # noqa: 401
 
 
-def _get_extension_path(lib_name: str) -> str:
+def _get_extension_path(lib_name):
 
     lib_dir = os.path.dirname(__file__)
     if os.name == "nt":
--- a/torchvision/io/image.py
+++ b/torchvision/io/image.py
@@ -1,4 +1,8 @@
+import ctypes
+import os
+import sys
 from enum import Enum
+from warnings import warn
 
 import torch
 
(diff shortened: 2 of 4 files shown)


torchvision/io/image.py
```diff
@@ -7,8 +7,27 @@
...
+                                                         0x00001100)
+        else:
+            import warnings
+            warnings.warn("LoadLibraryExW is missing in kernel32.dll")
     torch.ops.load_library(lib_path)
-except (ImportError, OSError):
+except (ImportError, OSError) as e:
+    import warnings
+    warnings.warn(f"Failed to load {global().get('lib_path','image.pyd')}: {e}")
     pass
```
pmeier, line 31:
```suggestion
```

torchvision/io/image.py
```diff
@@ -7,8 +7,27 @@
...
 try:
     lib_path = _get_extension_path("image")
+    # On Windows Python-3.8+ has `os.add_dll_directory` call,
+    # which is called from _get_extension_path to configure dll search path
+    # Condition below adds a workaround for older versions by
+    # explicitly calling `LoadLibraryExW` with the following flags:
+    #  - LOAD_LIBRARY_SEARCH_DEFAULT_DIRS (0x1000)
+    #  - LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR (0x100)
+    import os as _os, _sys as _sys
+    if _os.name == "nt" and sys.version_info < (3, 8):
```
pmeier, line 17: Why do we lazily and "privately" import this?

torchvision/io/image.py
```diff
@@ -7,8 +7,27 @@
...
+
+        from ctypes.windll import kernel32 as _kernel32
+        if hasattr(_kernel32, "LoadLibraryExW"):
+            _image_lib_handle = _kernel32.LoadLibraryExW(lib_path,
+                                                         None,
+                                                         0x00001100)
+        else:
+            import warnings
```
pmeier, line 25: Same as above.

torchvision/io/image.py
```diff
@@ -7,8 +7,27 @@
...
+            warnings.warn("LoadLibraryExW is missing in kernel32.dll")
     torch.ops.load_library(lib_path)
-except (ImportError, OSError):
+except (ImportError, OSError) as e:
+    import warnings
+    warnings.warn(f"Failed to load {global().get('lib_path','image.pyd')}: {e}")
```
pmeier, line 30:
```suggestion
    warnings.warn(f"Failed to load {globals().get('lib_path','image.pyd')}: {e}")
```

pmeier (commented):
I can't comment on the actual content. One minor syntax error and a few questions inline.

torchvision/io/image.py
```diff
@@ -1,15 +1,33 @@
...
+    #  - LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR (0x100)
+    if os.name == "nt" and sys.version_info < (3, 8):
+
+        from ctypes.windll import kernel32 as _kernel32
+        if hasattr(_kernel32, "LoadLibraryExW"):
+            _image_lib_handle = _kernel32.LoadLibraryExW(lib_path,
+                                                         None,
+                                                         0x00001100)
+        else:
+            warn("LoadLibraryExW is missing in kernel32.dll")
```
fmassa, line 27: Wouldn't it be better to move this to https://github.com/pytorch/vision/blob/57e6e30203d24d13f361da2ec98fbd8e87a2b4b3/torchvision/_internally_replaced_utils.py#L25 , so that other invocations (such as the video loading one) benefits from this improvement as well?

andfoy:
It seems that these changes were also introduced back in PR https://github.com/pytorch/vision/pull/2811, do we know why they were removed?

torchvision/io/image.py
malfet: Good point, but I've better add some automated PR first

malfet:
> It seems that these changes were also introduced back in PR #2811, do we know why they were removed?

Hmm, that's a very good point. Looks like it was accidentally deleted by #4080
cc: @NicolasHug

NicolasHug:
> > It seems that these changes were also introduced back in PR #2811, do we know why they were removed?
> 
> Hmm, that's a very good point. Looks like it was accidentally deleted by #4080 cc: @NicolasHug

Thanks for the ping, and sorry for the trouble. Prior to #4080 we had 3 different registration mechanisms - one for extensions.py, one for image, and one for video. I unified them into a single one - it wasn't clear why each of them were different (no comments, no link to related issues), so I picked the simplest one that seemed to work: the one from `extension.py`.

torchvision/io/image.py
```diff
@@ -1,15 +1,18 @@
...
 
 
 try:
-    lib_path = _get_extension_path("image")
-    torch.ops.load_library(lib_path)
-except (ImportError, OSError):
-    pass
+    _load_library("image")
+except (ImportError, OSError) as e:
+    warn(f"Failed to load {globals().get('lib_path','image.pyd')}: {e}")
```
fmassa, line 15: nit: does `pyd` file exists for all platforms?

torchvision/_internally_replaced_utils.py
```diff
@@ -56,3 +56,21 @@ def _get_extension_path(lib_name):
...
+    #  - LOAD_LIBRARY_SEARCH_DEFAULT_DIRS (0x1000)
+    #  - LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR (0x100)
+    if os.name == "nt" and sys.version_info < (3, 8):
+
+        _kernel32 = ctypes.WinDLL("kernel32.dll", use_last_error=True)
+        if hasattr(_kernel32, "LoadLibraryExW"):
+            _image_lib_handle = _kernel32.LoadLibraryExW(lib_path, None, 0x00001100)
+        else:
+            warn("LoadLibraryExW is missing in kernel32.dll")
+    torch.ops.load_library(lib_path)
```
fmassa, line 76: This is great as is, but do you think that we should move this to PyTorch?

torchvision/_internally_replaced_utils.py
```diff
@@ -56,3 +56,21 @@ def _get_extension_path(lib_name):
         raise ImportError
 
     return ext_specs.origin
+
+
+def _load_library(lib_name: str) -> None:
```
fmassa, line 61: @malfet this is ok here, but given that this function is generic wrt fbcode and OSS, we could also move it to a different location (like https://github.com/pytorch/vision/blob/main/torchvision/extension.py for example) so that we don't need to re-implement it in fbcode (this file is entirely replaced in there)

otherwise, @NicolasHug we will need to implement this function in fbcode as well.

fmassa (approved):
Thanks a lot Nikita, and sorry for the delay in getting back on this.

I have one high-level comment, otherwise the PR is good to merge.

torchvision/_internally_replaced_utils.py
malfet: Not really, as dependencies should usually be self-contained, but TorchVision decided to ship dynamic dependencies on Windows

malfet:
Moved `_load_library` to `torchvision.extension`

torchvision/io/image.py
```diff
@@ -1,15 +1,18 @@
+import ctypes
+import os
+import sys
```
fmassa, line 3: nit: you probably need to move those imports to `extension.py` now

torchvision/_internally_replaced_utils.py
fmassa, line 76: also, looks like this function hasn't been deleted from this file after the move?

torchvision/_internally_replaced_utils.py
malfet: Sorry, I've intended to remove it here

torchvision/io/image.py
malfet: Thank you for pointing this out, this is an artifact from a refactor, I guess simple warning it sufficient there

torchvision/_internally_replaced_utils.py
malfet: Done

torchvision/io/image.py
```diff
@@ -1,15 +1,15 @@
...
 
 
 try:
-    lib_path = _get_extension_path("image")
-    torch.ops.load_library(lib_path)
-except (ImportError, OSError):
-    pass
+    _load_library("image")
+except (ImportError, OSError) as e:
+    warn(f"Failed to load image Python extension: {e}")
```
adamjstewart, line 12: Is there a reason that this was changed from `pass` to `warn`? When running `pytest` on my project, I see warnings like:
```
../.spack/.spack-env/view/lib/python3.9/site-packages/torchvision/io/image.py:11
  /Users/Adam/.spack/.spack-env/view/lib/python3.9/site-packages/torchvision/io/image.py:11: UserWarning: Failed to load image Python extension: 
    warn(f"Failed to load image Python extension: {e}")
```
It's unclear if this is a serious bug in my installation that I should fix, or something I can safely ignore. FWIW, I'm using the pillow-SIMD backend.

PR #4482, opened 2021-09-24
[CircleCI] s/gpu.small/gpu.nvidia.small/

malfet:
Old resource class is getting deprecated

PR #4895, opened 2021-11-10
Add Sintel Dataset to the dataset prototype API

krshrimali:
This PR attempts to add Sintel Dataset to the prototype API.

There are a few TODOs:

* [ ] Add comments wherever required
* [ ] Add sha256 value

Though these TODOs should not block the review process, and I'm looking forward to the first round of review for this PR. In case there are things that I missed, please feel free to point out.

Special thanks to @pmeier on being very helpful with the bug fixing, and helping with the PR.

cc: @pmeier @NicolasHug


torchvision/prototype/datasets/_builtin/sintel.py
```diff
@@ -0,0 +1,167 @@
...
+
+class SINTEL(Dataset):
+    def _make_info(self) -> DatasetInfo:
+        return DatasetInfo(
+            "sintel",
+            type=DatasetType.IMAGE,
+            homepage="",
+            valid_options=dict(
+                split=("train", "test"),
+                pass_=("clean", "final"),
```
krshrimali, line 63: TODO: we should have `both` as an option here. And this should be renamed to `pass`, however that's a Python keyword and I'll prefer to call it something else.

pmeier: In the other implementation we call it `pass_name`:

https://github.com/pytorch/vision/blob/031e129b5590ec9619962ba44254405fd9ec90c8/torchvision/datasets/_optical_flow.py#L106-L107

```suggestion
                pass_name=("clean", "final"),
```

torchvision/prototype/datasets/_builtin/sintel.py
```diff
@@ -0,0 +1,167 @@
...
+            ),
+        )
+
+    def resources(self, config: DatasetConfig) -> List[OnlineResource]:
+        archive = HttpResource(
+            "http://sintel.cs.washington.edu/MPI-Sintel-complete.zip",
+            sha256="",
```
pmeier, line 70: If you have the file locally, an you run 

```shell
$ sha256sum MPI-Sintel-complete.zip
```

and add the value here?

torchvision/prototype/datasets/_builtin/sintel.py
```diff
@@ -0,0 +1,167 @@
...
+
+    def _read_flo(self, file: IterDataPipe[Tuple[Any, io.IOBase]]) -> Any:
+        magic = file.read(4)
+        if magic != b"PIEH":
+            raise ValueError("Magic number incorrect. Invalid .flo file")
+        w = int.from_bytes(file.read(4), "little")
+        h = int.from_bytes(file.read(4), "little")
+        data = file.read(2 * w * h * 4)
+        data = np.frombuffer(data, dtype=np.float32)
+        return data.reshape(h, w, 2).transpose(2, 0, 1)
```
pmeier, line 95: There is still discussion about that in #4482. Could you for now `from torchvision.datasets._optical_flow import _read_flo` and use that?

pmeier: Plus, we don't want to return `np.ndarray`'s anymore, so please convert it to a tensor with `torch.from_numpy`.

torchvision/prototype/datasets/_builtin/sintel.py
```diff
@@ -0,0 +1,167 @@
...
+    def _classify_archive(self, data: Tuple[str, Any], *, config: DatasetConfig) -> Union[int, None]:
+        path = pathlib.Path(data[0])
+        if config.pass_ == path.parent.parent.name and path.suffix == ".png":
+            return 0
+        elif path.suffix == ".flo":
+            return 1
+        else:
+            return None
+
+    def _read_flo(self, file: IterDataPipe[Tuple[Any, io.IOBase]]) -> Any:
```
pmeier, line 87:
```suggestion
    def _read_flo(self, file: io.IOBase) -> torch.Tensor:
```

The annotation for the return value is only valid after we actually convert to a tensor.

torchvision/prototype/datasets/_builtin/sintel.py
```diff
@@ -0,0 +1,167 @@
...
+        config: DatasetConfig,
+    ) -> Dict[str, Any]:
+        if config.split == "train":
+            flo, images = data
+            img1, img2 = images
+            flow_arr = self._read_flo(flo[1])
+            del images, flo
+        else:
+            # When split is `test`
+            img1, img2 = data
```
pmeier, line 120: Having different input types / sizes based on the split is an anti-pattern. Instead we should have the same structure for both splits and substitute `None` for the flo file handle in the test split. That way we can simply do

```python
flow = self._read_flo(flo_buffer) if flow_buffer else None
```

torchvision/prototype/datasets/_builtin/sintel.py
```diff
@@ -0,0 +1,167 @@
...
+        data: Tuple[Tuple[str, io.IOBase], Any],
+        *,
+        decoder: Optional[Callable[[io.IOBase], torch.Tensor]],
+        config: DatasetConfig,
+    ) -> Dict[str, Any]:
+        if config.split == "train":
+            flo, images = data
+            img1, img2 = images
+            flow_arr = self._read_flo(flo[1])
+            del images, flo
```
pmeier, line 117: Not sure why you are `del`eting stuff here?

torchvision/prototype/datasets/_builtin/sintel.py
```diff
@@ -0,0 +1,167 @@
...
+        else:
+            # When split is `test`
+            img1, img2 = data
+
+        del data
```
pmeier, line 122: Same as above.

torchvision/prototype/datasets/_builtin/sintel.py
```diff
@@ -0,0 +1,167 @@
...
+
+        path1, buffer1 = img1
+        path2, buffer2 = img2
+
+        return dict(
+            image1=decoder(buffer1) if decoder else buffer1,
+            image2=decoder(buffer2) if decoder else buffer2,
+            flow=flow_arr if config.split == "train" else None,
+        )
```
pmeier, line 131: There is some information missing from this. We should put all the paths (image1, image2, and flow) in here. Is there something like a label? Even if not, we should probably put the scene in too.

torchvision/prototype/datasets/_builtin/sintel.py
```diff
@@ -0,0 +1,167 @@
...
+    def resources(self, config: DatasetConfig) -> List[OnlineResource]:
+        archive = HttpResource(
+            "http://sintel.cs.washington.edu/MPI-Sintel-complete.zip",
+            sha256="",
+        )
+        return [archive]
+
+    def _classify_train_test(self, data: Tuple[str, Any], *, config: DatasetConfig) -> bool:
+        path = pathlib.Path(data[0])
+        return config.split in str(path.parent)
```
pmeier, line 76:
```suggestion
        return config.split in path.parent.name
```

pmeier: Plus, please leave a comment on why you don't check for equality here.

torchvision/prototype/datasets/_builtin/sintel.py
```diff
@@ -0,0 +1,167 @@
...
+        self,
+        resource_dps: List[IterDataPipe],
+        *,
+        config: DatasetConfig,
+        decoder: Optional[Callable[[io.IOBase], torch.Tensor]],
+    ) -> IterDataPipe[Tuple[Dict[str, Any], Dict[str, Any]]]:
+        dp = resource_dps[0]
+        archive_dp = ZipArchiveReader(dp)
+
+        curr_split = Filter(archive_dp, partial(self._classify_train_test, config=config))
```
pmeier, line 143: No need for `partial` here

```suggestion
        curr_split = Filter(archive_dp, self._classify_train_test, fn_kwargs=dict(split=split))
```

torchvision/prototype/datasets/_builtin/sintel.py
```diff
@@ -0,0 +1,167 @@
...
+        )
+
+    def resources(self, config: DatasetConfig) -> List[OnlineResource]:
+        archive = HttpResource(
+            "http://sintel.cs.washington.edu/MPI-Sintel-complete.zip",
+            sha256="",
+        )
+        return [archive]
+
+    def _classify_train_test(self, data: Tuple[str, Any], *, config: DatasetConfig) -> bool:
```
pmeier, line 74: IMO, "classify" should not be used here, because it is usually used for the `Demultiplexer` datapipe, which returns `Optional[int]` instead of `bool`.

```suggestion
    def _filter_split(self, data: Tuple[str, Any], *, config: DatasetConfig) -> bool:
```

pmeier: Plus, there is no need to take the whole config.

```suggestion
    def _classify_train_test(self, data: Tuple[str, Any], *, split: str) -> bool:
```

torchvision/prototype/datasets/_builtin/sintel.py
```diff
@@ -0,0 +1,167 @@
...
+        path = pathlib.Path(data[0])
+        return config.split in str(path.parent)
+
+    def _classify_archive(self, data: Tuple[str, Any], *, config: DatasetConfig) -> Union[int, None]:
+        path = pathlib.Path(data[0])
+        if config.pass_ == path.parent.parent.name and path.suffix == ".png":
```
pmeier, line 80: Instead of checking for the `pass_name` here, couldn't we have a second `Filter` datapipe before?

torchvision/prototype/datasets/_builtin/sintel.py
```diff
@@ -0,0 +1,167 @@
...
+    DatasetType,
+)
+from torchvision.prototype.datasets.utils._internal import INFINITE_BUFFER_SIZE
+
+T = TypeVar("T")
+
+FILE_NAME_PATTERN = re.compile(r"(frame|image)_(?P<idx>\d+)[.](flo|png)")
+
+try:
+    from itertools import pairwise  # type: ignore
```
pmeier, line 33: Can you make the directive more concrete? You should put the error code `mypy` is showing you in brackets, like `type: ignore[foo]`.

torchvision/prototype/datasets/_builtin/sintel.py
```diff
@@ -0,0 +1,167 @@
...
+    DatasetInfo,
+    HttpResource,
+    OnlineResource,
+    DatasetType,
+)
+from torchvision.prototype.datasets.utils._internal import INFINITE_BUFFER_SIZE
+
+T = TypeVar("T")
+
+FILE_NAME_PATTERN = re.compile(r"(frame|image)_(?P<idx>\d+)[.](flo|png)")
```
pmeier, line 30: This should probably be a class attribute of the dataset. Doing so please also make it private.

torchvision/prototype/datasets/_builtin/sintel.py
```diff
@@ -0,0 +1,167 @@
...
+except ImportError:
+    from itertools import tee
+
+    def pairwise(iterable: Iterable[T]) -> Iterable[Tuple[T, T]]:
+        a, b = tee(iterable)
+        next(b, None)
+        return zip(a, b)
+
+
+class IntCategoryGrouper(IterDataPipe[Tuple[Tuple[str, T], Tuple[str, T]]]):
```
pmeier, line 43:
```suggestion
class InCategoryGrouper(IterDataPipe[Tuple[Tuple[str, T], Tuple[str, T]]]):
```

torchvision/prototype/datasets/_builtin/sintel.py
```diff
@@ -0,0 +1,167 @@
...
+            "http://sintel.cs.washington.edu/MPI-Sintel-complete.zip",
+            sha256="",
+        )
+        return [archive]
+
+    def _classify_train_test(self, data: Tuple[str, Any], *, config: DatasetConfig) -> bool:
+        path = pathlib.Path(data[0])
+        return config.split in str(path.parent)
+
+    def _classify_archive(self, data: Tuple[str, Any], *, config: DatasetConfig) -> Union[int, None]:
```
pmeier, line 78: Both are equivalent, but `Optional` is the preferred way to go.

```suggestion
    def _classify_archive(self, data: Tuple[str, Any], *, config: DatasetConfig) -> Optional[int]:
```

torchvision/prototype/datasets/_builtin/sintel.py
```diff
@@ -0,0 +1,167 @@
...
+            flow=flow_arr if config.split == "train" else None,
+        )
+
+    def _make_datapipe(
+        self,
+        resource_dps: List[IterDataPipe],
+        *,
+        config: DatasetConfig,
+        decoder: Optional[Callable[[io.IOBase], torch.Tensor]],
+    ) -> IterDataPipe[Tuple[Dict[str, Any], Dict[str, Any]]]:
```
pmeier, line 139:
```suggestion
    ) -> IterDataPipe[Dict[str, Any]]:
```

torchvision/prototype/datasets/_builtin/sintel.py
```diff
@@ -0,0 +1,167 @@
...
+                pass_images_dp,
+                key_fn=self._flows_key,
+                ref_key_fn=self._images_key,
+            )
+            return Mapper(zipped_dp, self._collate_and_decode_sample, fn_kwargs=dict(decoder=decoder, config=config))
+        else:
+            # When split is `test`, flo_dp will be empty and thus should not be zipped with pass_images_dp
+            return Mapper(
+                pass_images_dp, self._collate_and_decode_sample, fn_kwargs=dict(decoder=decoder, config=config)
+            )
```
pmeier, line 167: I would restructure this as follows: We can branch (`if config.split == "train"`) directly after the filtering. In the train branch, we are going to split and merge the datapipe in order to get image / flow pairs. In the test branch we don't need to split anything, but just filter out the images. Note that they also need to be shuffled.

pmeier (changes requested):
Hey @krshrimali and thanks a lot for the PR! I left some comments inline. Overall this looks good.

torchvision/prototype/datasets/_builtin/sintel.py
```diff
@@ -0,0 +1,167 @@
...
+
+            yield item1, item2
+
+
+class SINTEL(Dataset):
+    def _make_info(self) -> DatasetInfo:
+        return DatasetInfo(
+            "sintel",
+            type=DatasetType.IMAGE,
+            homepage="",
```
pmeier, line 60:
```suggestion
            homepage="http://sintel.is.tue.mpg.de/",
```

torchvision/prototype/datasets/_builtin/sintel.py
```diff
@@ -0,0 +1,167 @@
...
+            valid_options=dict(
+                split=("train", "test"),
+                pass_=("clean", "final"),
+            ),
+        )
+
+    def resources(self, config: DatasetConfig) -> List[OnlineResource]:
+        archive = HttpResource(
+            "http://sintel.cs.washington.edu/MPI-Sintel-complete.zip",
```
pmeier, line 69: Kind of a pickle here. The authors of the dataset provide a mirror in the US (which is used here) as well as a mirror in the EU. Do we have any data of the geo distribution of our user base? If not, I would use the EU one, because it is the "original" mirror.

```suggestion
            "http://files.is.tue.mpg.de/sintel/MPI-Sintel-complete.zip",
```

torchvision/prototype/datasets/_builtin/sintel.py
pmeier: I ended up downloading it anyway to get a better feel for the structure.

```suggestion
            sha256="bdc80abbe6ae13f96f6aa02e04d98a251c017c025408066a00204cd2c7104c5f",
```

torchvision/prototype/datasets/_builtin/sintel.py
NicolasHug: Yes, there is still time to change the original implementation (it hasn't been released yet), so feel free to suggest new names. But otherwise we should try to keep both the current and new version as close as possible, unless there's a really good reason not to

torchvision/prototype/datasets/_builtin/sintel.py
```diff
@@ -0,0 +1,182 @@
...
+        magic = file.read(4)
+        if magic != b"PIEH":
+            raise ValueError("Magic number incorrect. Invalid .flo file")
+        w = int.from_bytes(file.read(4), "little")
+        h = int.from_bytes(file.read(4), "little")
+        data = file.read(2 * w * h * 4)
+        data_arr = np.frombuffer(data, dtype=np.float32)
+        # Creating a copy of the underlying array, to avoid UserWarning: "The given NumPy array
+        #     is not writeable, and PyTorch does not support non-writeable tensors."
+        return torch.from_numpy(np.copy(data_arr.reshape(h, w, 2).transpose(2, 0, 1)))
```
krshrimali, line 106: The reason this function exists, and is not used from the existing `read_flo` function (in the [other API](https://github.com/pytorch/vision/blob/main/torchvision/datasets/_optical_flow.py#L366)):

`zipfile` objects don't work well with `np.fromfile` and there is an `io.unsupportedoperation: fileno` error that is raised. Please note that earlier while testing with @pmeier, we didn't get an error because a generator was returned which doesn't give an error unless and until you use it. This error was raised when I replaced `yield` with `return`.

I also verified the results from the function used [here](https://github.com/pytorch/vision/blob/main/torchvision/datasets/_optical_flow.py#L366) with this function, and the results are same for a single file (didn't check for all of them).

Suggestions are, as always, welcome. :)

torchvision/prototype/datasets/_builtin/sintel.py
```diff
@@ -0,0 +1,182 @@
...
+                pass_images_dp,
+                key_fn=self._flows_key,
+                ref_key_fn=self._images_key,
+            )
+            return Mapper(zipped_dp, self._collate_and_decode_sample, fn_kwargs=dict(decoder=decoder, config=config))
+        else:
+            # When config.split is "test"
+            # There are no flow files for test split
+            # TODO: How to zip an empty IterDataPipe and pass to the Mapper?
+            # flo_dp = IterDataPipe()
```
krshrimali, line 176: I'll need some help on this, @pmeier. I'm not sure how to create a `IterKeyZipper` from `flo_dp` (which will be `None`/empty here) and `pass_images_dp` (not `None`). Can you please take a look if you have the time?

torchvision/prototype/datasets/_builtin/sintel.py
```diff
@@ -0,0 +1,182 @@
...
+    _FILE_NAME_PATTERN = re.compile(r"(frame|image)_(?P<idx>\d+)[.](flo|png)")
+
+    def _make_info(self) -> DatasetInfo:
+        return DatasetInfo(
+            "sintel",
+            type=DatasetType.IMAGE,
+            homepage="http://sintel.is.tue.mpg.de/",
+            valid_options=dict(
+                split=("train", "test"),
+                pass_name=("clean", "final"),
```
krshrimali, line 64: TODO: Add `both` as an option here, will do this by tonight. (apologies, a little caught up with other PRs to handle).

krshrimali (commented):
Hi, @pmeier and @NicolasHug 

Thank you for your comments and reviews. I'm sorry that the threads could not be continued from the review (I committed the changes, and then the reviews got outdated). Can you guys please review the PR again whenever you find the time?

Looking forward to your comments, and thoughts.

Note: I'll be able to address them in a day, a little caught with other stuff. Thanks!

torchvision/prototype/datasets/_builtin/sintel.py
```diff
@@ -0,0 +1,182 @@
...
+            # When split is `test`
+            img1, img2 = data
+
+        path1, buffer1 = img1
+        path2, buffer2 = img2
+
+        return dict(
+            image1=(path1, decoder(buffer1)) if decoder else (path1, buffer1),
+            image2=(path2, decoder(buffer2)) if decoder else (path2, buffer2),
+            flow=(flo[0], flow_arr) if config.split == "train" else ("", None),
```
krshrimali, line 141: Question: What do we want to return here? `image1` should be a `tuple` of path and buffer? Or should it be `image1_path` and `image1? (same question for `image2` and `flow`)

torchvision/prototype/datasets/_builtin/sintel.py
```diff
@@ -0,0 +1,182 @@
...
+    Dataset,
+    DatasetConfig,
+    DatasetInfo,
+    HttpResource,
+    OnlineResource,
+    DatasetType,
+)
+from torchvision.prototype.datasets.utils._internal import INFINITE_BUFFER_SIZE
+
+T = TypeVar("T")
```
pmeier, line 28: Nit: could you move this below the `pairwise` import? This way we have the imports in one block.

torchvision/prototype/datasets/_builtin/sintel.py
```diff
@@ -0,0 +1,182 @@
...
+            sha256="bdc80abbe6ae13f96f6aa02e04d98a251c017c025408066a00204cd2c7104c5f",
+        )
+        return [archive]
+
+    def _filter_split(self, data: Tuple[str, Any], *, split: str) -> bool:
+        path = pathlib.Path(data[0])
+        # The dataset contains has the folder "training", while allowed options for `split` are
+        # "train" and "test", we don't check for equality here ("train" != "training") and instead
+        # check if split is in the folder name
+        return split in path.parent.parent.parent.name
```
pmeier, line 80: You can avoid chained `.parent` by using `.parents`

```suggestion
        return split in path.parents[2].name
```

torchvision/prototype/datasets/_builtin/sintel.py
```diff
@@ -0,0 +1,182 @@
...
+        self,
+        resource_dps: List[IterDataPipe],
+        *,
+        config: DatasetConfig,
+        decoder: Optional[Callable[[io.IOBase], torch.Tensor]],
+    ) -> IterDataPipe[Dict[str, Any]]:
+        dp = resource_dps[0]
+        archive_dp = ZipArchiveReader(dp)
+
+        curr_split = Filter(archive_dp, self._filter_split, fn_kwargs=dict(split=config.split))
```
pmeier, line 154: Can't we have a second `Filter` below this one that filters for the `pass_name`? That way we don't need to handle it twice in `_classify_archive` and in `_filter_images`.

torchvision/prototype/datasets/_builtin/sintel.py
```diff
@@ -0,0 +1,182 @@
...
+            )
+            flo_dp = Shuffler(flo_dp, buffer_size=INFINITE_BUFFER_SIZE)
+            pass_images_dp: IterDataPipe[Tuple[str, Any], Tuple[stry, Any]] = InSceneGrouper(pass_images_dp)
+            zipped_dp = IterKeyZipper(
+                flo_dp,
+                pass_images_dp,
+                key_fn=self._flows_key,
+                ref_key_fn=self._images_key,
+            )
+            return Mapper(zipped_dp, self._collate_and_decode_sample, fn_kwargs=dict(decoder=decoder, config=config))
```
pmeier, line 171: Since this call is the same for both splits, I would move it out of `if / else` branch so both get the same call.

torchvision/prototype/datasets/_builtin/sintel.py
pmeier: You don't need an `IterKeyZipper` here, because, like you said, `flo_dp` is empty or better non-existent. To add fake flow data, you can simply do

```python
pass_images_dp = Mapper(pass_images_dp, lambda data: ((None, None), data))
```

Now `pass_images_dp` and `zipped_dp` from the train split should have the same format and can easily be processed by `_collate_and_decode_sample`.

torchvision/prototype/datasets/_builtin/sintel.py
pmeier: The latter. The dictionary should contain the image, which is either a tensor or a buffer depending on `decoder`, as well the image path.

torchvision/prototype/datasets/_builtin/sintel.py
```diff
@@ -0,0 +1,182 @@
...
+        # flo, images = data
+        # img1, img2 = images
+
+        if config.split == "train":
+            flo, images = data
+            img1, img2 = images
+            flow_arr = self._read_flo(flo[1])
+        else:
+            # When split is `test`
+            img1, img2 = data
```
pmeier, line 133: This needs to go. We do not want different input structures based on the config. You can add fake flow data as described in a comment above.

torchvision/prototype/datasets/_builtin/sintel.py
pmeier: Oh well. @NicolasHug that is a strong argument in favor of #4882 since reading from archives will be the norm for the new datasets.

pmeier (changes requested):
I did a second pass. Let me know if anything is not clear or you need help.

torchvision/prototype/datasets/_builtin/sintel.py
krshrimali: I gave it a thought before, but I feel the current implementation saves an extra pass.

Either `_classify_archive` or `_filter_images` function is called depending on `config.split`, but if we have an extra filter below the `curr_split = Filter(..., self._filter_split, ...)` - then that's one pass, and then there's another in the `Demultiplexer` (which uses `_classify_archive`) or in the `Filter` (which uses `_filter_images`). So in total, there are 2 passes through the path name. Not a big difference, I agree, but IMO - the current implementation is okay, as either of them is called depending on the split.

If you feel otherwise, please let me know. Thanks!

torchvision/prototype/datasets/_builtin/sintel.py
krshrimali: Awesome, thanks Philip! This should be fixed now. :)

torchvision/prototype/datasets/_builtin/sintel.py
krshrimali: Done! :)

torchvision/prototype/datasets/_builtin/sintel.py
krshrimali: Done! :) Should the label be `flow` here?

torchvision/prototype/datasets/_builtin/sintel.py
krshrimali: Sure, fixed. :)

torchvision/prototype/datasets/_builtin/sintel.py
krshrimali: Sure, thanks. Fixed now.

torchvision/prototype/datasets/_builtin/sintel.py
krshrimali: Thanks, it's gone now. :)

torchvision/prototype/datasets/_builtin/sintel.py
pmeier: Not sure I understand. If we have `Filter` on the `pass_name` before we get to the `Demultiplexer`, we don't need to do it inside `_classify_archive`, right? If that is the case, the total number of comparisons stays the same, since `_classify_archive` would only split on the suffix.

torchvision/prototype/datasets/_builtin/sintel.py
```diff
@@ -0,0 +1,180 @@
...
+    DatasetType,
+)
+from torchvision.prototype.datasets.utils._internal import INFINITE_BUFFER_SIZE
+
+try:
+    from itertools import pairwise  # type: ignore[attr-defined]
+except ImportError:
+    from itertools import tee
+
+    T = TypeVar("T")
```
pmeier, line 33: I was wrong, sorry. Since we need it for `pairwise` as well as the `InSceneGrouper`, it needs to be before the `try`.

torchvision/prototype/datasets/_builtin/sintel.py
```diff
@@ -0,0 +1,180 @@
...
+            if pathlib.Path(item1[0]).parent != pathlib.Path(item2[0]).parent:
+                continue
+
+            yield item1, item2
+
+
+class SINTEL(Dataset):
+
+    _FILE_NAME_PATTERN = re.compile(r"(frame|image)_(?P<idx>\d+)[.](flo|png)")
+    _PASS_NAME_BOTH = re.compile(r"clean|final")
```
pmeier, line 56: No need to use a regex here since there is nothing "regular" about the two possibilities.

torchvision/prototype/datasets/_builtin/sintel.py
```diff
@@ -0,0 +1,180 @@
...
+        path1, buffer1 = img1
+        path2, buffer2 = img2
+
+        return dict(
+            image1=decoder(buffer1) if decoder else buffer1,
+            image1_path=path1,
+            image2=decoder(buffer2) if decoder else buffer2,
+            image2_path=path2,
+            flow=flow_arr,
+            flow_path=flo[0] or "",
```
pmeier, line 142: Returning `None` is fine

```suggestion
            flow_path=flo[0],
```

torchvision/prototype/datasets/_builtin/sintel.py
```diff
@@ -0,0 +1,180 @@
...
+                key_fn=self._flows_key,
+                ref_key_fn=self._images_key,
+            )
+        else:
+            pass_images_dp = Filter(curr_split, self._filter_images, fn_kwargs=dict(pass_name=config.pass_name))
+            pass_images_dp = Shuffler(pass_images_dp, buffer_size=INFINITE_BUFFER_SIZE)
+            pass_images_dp = InSceneGrouper(pass_images_dp)
+            # When config.split is "test"
+            # There are no flow files for test split, hence we create a fake flow data
+            zipped_dp = Mapper(pass_images_dp, lambda data: ((None, None), data))
```
pmeier, line 178: Although I suggested a `lambda` in my previous review, `torchdata` screams at you if you use one. So we should create dedicated function to do that. If we give it a descriptive name like `_add_fake_flow_data` we can also get rid of the comment.

pmeier (commented):
A few minor things left. I think when all open comments are addressed that is good to go from my side. After that Nicolas will take look in case I missed something, since he is more familiar with the dataset.

torchvision/prototype/datasets/_builtin/sintel.py
```diff
@@ -0,0 +1,180 @@
...
+        else:
+            return None
+
+    def _read_flo(self, file: io.IOBase) -> torch.Tensor:
+        magic = file.read(4)
+        if magic != b"PIEH":
+            raise ValueError("Magic number incorrect. Invalid .flo file")
+        w = int.from_bytes(file.read(4), "little")
+        h = int.from_bytes(file.read(4), "little")
+        data = file.read(2 * w * h * 4)
```
NicolasHug, line 107: these are expected to be encoded as little endian as well

torchvision/prototype/datasets/_builtin/sintel.py
pmeier: > Before merging anything we should try to run more robust tests to avoid surprises in the future :)

We should not add this functionality here, but rather go for #4882 and depend on that here. Of course that would involve writing tests for a `read_flo` or any other wrapper.

torchvision/prototype/datasets/_builtin/sintel.py
krshrimali: You are right. Should be fixed now. Sorry for the misunderstanding, and thanks for the suggestion, Philip! :)

torchvision/prototype/datasets/_builtin/sintel.py
krshrimali: Done. Thanks!

torchvision/prototype/datasets/_builtin/sintel.py
krshrimali: Sure. :) Fixed now.

torchvision/prototype/datasets/_builtin/sintel.py
krshrimali: This is a TODO left from the current comments. Working on it now. Thanks @NicolasHug for pointing this out.

krshrimali: I changed the `dtype` from `np.float32` to `<f4` for `data_arr`, but just wanted to quickly check, is this the correct fix? cc: @pmeier

NicolasHug: Yes `<f4` would be correct, it's a little endian float that fits in 4 bytes (float32).

But it might be preferable to always call the same function instead of mixing `frombuffer` and `from_bytes`.

krshrimali: Right, that makes sense. Thanks, @NicolasHug! I've fixed this (preferred using `frombuffer` in both instances).

torchvision/prototype/datasets/_builtin/sintel.py
```diff
@@ -0,0 +1,182 @@
...
+        elif suffix == ".png":
+            return 1
+        else:
+            return None
+
+    def _read_flo(self, file: io.IOBase) -> torch.Tensor:
+        magic = file.read(4)
+        if magic != b"PIEH":
+            raise ValueError("Magic number incorrect. Invalid .flo file")
+        w = np.frombufer(file.read(4), dtype="<i4")
```
pmeier, line 104:
```suggestion
        w = np.frombuffer(file.read(4), dtype="<i4")
```

torchvision/prototype/datasets/_builtin/sintel.py
```diff
@@ -0,0 +1,182 @@
...
+        if config.split == "train":
+            flo_dp, pass_images_dp = Demultiplexer(
+                filtered_curr_split,
+                2,
+                partial(self._classify_archive, pass_name=config.pass_name),
+                drop_none=True,
+                buffer_size=INFINITE_BUFFER_SIZE,
+            )
+            flo_dp = Shuffler(flo_dp, buffer_size=INFINITE_BUFFER_SIZE)
+            pass_images_dp: IterDataPipe[Tuple[str, Any], Tuple[stry, Any]] = InSceneGrouper(pass_images_dp)
```
pmeier, line 170:
```suggestion
            pass_images_dp: IterDataPipe[Tuple[str, Any], Tuple[str, Any]] = InSceneGrouper(pass_images_dp)
```

pmeier (changes requested):
Two nits now, but something is off:

```python
>>> len(tuple(datasets.load("sintel", decoder=None, split="train")))
0
```

Could you please add some mock data, so this is tested by our test suite? You need to add a `sintel` function https://github.com/pytorch/vision/blob/main/test/builtin_dataset_mocks.py and add `"sintel"` to https://github.com/pytorch/vision/blob/4b20ac52d868fcb8c9522df07e18f998f59bae6d/test/test_prototype_builtin_datasets.py#L14-L26

With the exception of archive generation, you can probably copy a lot from 

https://github.com/pytorch/vision/blob/4b20ac52d868fcb8c9522df07e18f998f59bae6d/test/test_datasets.py#L1874

krshrimali:
Update:

The code had a logical bug in the filter function and has been fixed now. I remember @pmeier mentioning in the review to have the scene returned in the `dict` as `label`, and that has been added in the recent commit.

I'm a little caught up today but will try working on the test as suggested by @pmeier above. Thanks!

test/builtin_dataset_mocks.py
```diff
@@ -504,3 +504,41 @@ def imagenet(info, root, config):
     make_tar(root, f"{devkit_root}.tar.gz", devkit_root, compression="gz")
 
     return num_samples
+
+
+@dataset_mocks.register_mock_data_fn
+def sintel(info, root, config):
```
krshrimali, line 510: @pmeier - I wasn't very sure what you meant by "archive generation" in your previous comment, but I tried replicating the test. Could you please take a look at this function and suggest if this does the job?

I ran the test using `python3 test/test_prototype_builtin_datasets.py -k sintel` and the test seems to pass (no failures). But I just want to double check.

pmeier: The test has to be run by `pytest` otherwise no test is run. If you do 

```bash
pytest test/test_prototype_builtin_datasets.py -k sintel
```

you will see [these failures](https://app.circleci.com/pipelines/github/pytorch/vision/12637/workflows/2819215a-ad45-4248-b753-67feed8123f5/jobs/1001187) popping up.

pmeier (changes requested):
Don't shoot the messenger here, but I think we need to postpone this PR for a bit. While debugging I realized in order to be able to read from the same image handle, it has to be seekable. Unfortunately, file handles inside a zip file are not for Python <=3.6:

```python
import zipfile

content = "foo"
archive = "foo.zip"

open(content, "w").close()
with zipfile.ZipFile(archive, "w") as file:
    file.write(content)

with zipfile.ZipFile(archive) as file:
    for info in file.infolist():
        with file.open(info) as internal_file:
            # This will fail on Python <= 3.6
            assert internal_file.seekable()
```
Since PyTorch's minimum Python requirement is 3.6, we cannot depend on that. I currently three ways out ordered by ascending preference:
1. Wait it out. EOL for Python 3.6 is next month, but I don't know when PyTorch will stop supporting it officially. I'm not aware for any plans for the next release, which will we be the one we plan on going public with the new datasets.
2. Every time we read and image for the first time, store the data in a buffer and retrieve it from there whenever we request it the second time. This has some major memory implications, since in the worst case we need to store half of the dataset in memory. When storing only the raw bytes, i.e. postpone decoding after the buffer, that is about 1GB.
3. Extract the zip archive and work with extracted files. They are seekable and we would circumvent the problem in its entirety. I'm currently working on adding support for that.

Thoughts?

NicolasHug:
> EOL for Python 3.6 is next month, but I don't know when PyTorch will stop supporting it officially

According to https://github.com/pytorch/pytorch/issues/66462 the next release should be >= 3.7, but I'll ask for confirmation

pmeier:
@malfet confirmed that the tentative plan is to drop 3.6 with the next release. 

Leaving that aside, I disregarded one fact in https://github.com/pytorch/vision/pull/4895#pullrequestreview-811255311 that makes the solution straight forward: the images are ordered. Thus, we need don't need to keep an arbitrary sized buffer, but only need to store the bytes of a single image. I've pushed a commit adding this feature and now we should be able to finalize this PR.

Issue #4880, opened 2021-11-06
Bug with transforms.Resize when used with transforms.ConvertImageDtype


numpee:
### 🐛 Describe the bug

Recent releases of Torchvision and the documentations that support it seem to  suggest that we can use `io.read_image + transforms.ConvertImageDtype` instead of the traditional `PIL.Image.read_fn + transforms.ToTensor`. However, I have found that there are two issues:

1. `io.read_image + transforms.ConvertImageDtype` do not actually return the same tensor values as `PIL + transforms.ToTensor`, even though they are supposed to provide the same functionality.
2. While `io.read_image + transforms.ConvertImageDtype` itself is significantly faster than using PIL, combining it with the `transforms.Resize` operation - specifically when upsampling - makes the operation much slower than the PIL alternative.

To add onto point 2, the two sets of functions I mention return the same type of tensor: `torch.float`. However, applying `transforms.Resize` on the tensor generated by `io.read_image + transforms.ConvertImageDtype` is much slower than applying the same resize operation on the output of `PIL read + transforms.ToTensor`. I can't really understand why this happens, since both calls to `Resize` are on tensors of type `torch.FloatTensor`. Also, this only occurs when upsampling.

Please refer to my post on the Pytorch Forum [here](https://discuss.pytorch.org/t/no-speedup-with-torchvision-io-and-weird-behavior-with-convertimagedtype/135943) for the full analysis.

### Versions

Collecting environment information...
PyTorch version: 1.10.0+cu113
Is debug build: False
CUDA used to build PyTorch: 11.3
ROCM used to build PyTorch: N/A

OS: Ubuntu 16.04.7 LTS (x86_64)
GCC version: (Ubuntu 5.4.0-6ubuntu1~16.04.12) 5.4.0 20160609
Clang version: Could not collect
CMake version: Could not collect
Libc version: glibc-2.23

Python version: 3.9.4 (default, Apr  9 2021, 01:15:05)  [GCC 5.4.0 20160609] (64-bit runtime)
Python platform: Linux-4.15.0-142-generic-x86_64-with-glibc2.23
Is CUDA available: True
CUDA runtime version: 10.0.130
GPU models and configuration:
GPU 0: NVIDIA GeForce RTX 2080 Ti
GPU 1: NVIDIA GeForce RTX 2080 Ti
GPU 2: NVIDIA GeForce RTX 2080 Ti
GPU 3: NVIDIA GeForce RTX 2080 Ti

[...]

PR #4898, opened 2021-11-10
Applied contiguous in `decode_*` ops

vfdev-5:
Description:
- Applied `contiguous` on decoded output tensor in decode_jpeg and decode_png ops
- Updated tests and docs

Related to #4880

Performance improvement on following transforms, see https://github.com/pytorch/vision/issues/4880#issuecomment-964981658

### DIFF
--- a/test/test_image.py
+++ b/test/test_image.py
@@ -94,6 +94,8 @@ def test_decode_jpeg(img_path, pil_mode, mode):
     data = read_file(img_path)
     img_ljpeg = decode_image(data, mode=mode)
 
+    assert img_ljpeg.is_contiguous()
+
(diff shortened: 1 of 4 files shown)


datumbox (approved):
LGTM, thanks @vfdev-5. Nice performance boost.

datumbox:
@NicolasHug Do you think it's worth adding `perf` in the list of high-level labels?

NicolasHug:
I would just add enhancement + perf

fmassa:
This was here on purpose for speed improvements on the reading (avoids copy).

I would revert this PR, and work on improving the performance on the transforms instead so that they better handle channel_last format

PR #4907, opened 2021-11-11
support label_smoothing in MixUp and CutMix

xiaohu2015:


datumbox (changes requested):
@xiaohu2015 Thanks for the PR.

Unfortunately we want to avoid coupling CutMix and Mixup with the loss. This was discussed back when the transforms were introduced, see https://github.com/pytorch/vision/pull/4379#discussion_r705281765. Note that label smoothing is already supported in our training scripts because PyTorch Core's CrossEntropyLoss can operate with smooth labels and accepts a `label_smoothing` parameter.

Also please note, even thought it's not strictly mandatory to raise an issue prior to sending a PR, it does help to discuss the details of more complicated features. This avoids throwaway work, gives context to the team during code review and maximizes the chance of merging.

xiaohu2015:
@datumbox Ok. thanks.

Issue #4899, opened 2021-11-10
RuntimeError: offset output dims: (160, 160) - computed output dims: (80, 80)


wduo:
### 🐛 Describe the bug

I run source build pytorch(torch==1.10.0) and torchvision(torchvision==0.12.0). 
train code snippet:
```
                from torchvision.ops import DeformConv2d as ModulatedDeformConv
                conv_op = ModulatedDeformConv
                offset_channels = 27
                self.conv2_offset = nn.Conv2d(
                    planes, deformable_groups * offset_channels,
                    kernel_size=3,
                    padding=1)
                self.conv2 = conv_op(
                    planes, planes, kernel_size=3, padding=1, stride=stride,
                    groups=deformable_groups, bias=False)

                offset_mask = self.conv2_offset(out)
                offset = offset_mask[:, :18, :, :]
                mask = offset_mask[:, -9:, :, :].sigmoid()
                out = self.conv2(out, offset, mask)
```
tensor shape below:
```
(Pdb) offset_mask.shape                                                                                                                                                                  
    torch.Size([16, 27, 160, 160])                                                                                                                                                           
(Pdb) offset.shape                                                                                                                                                                       
    torch.Size([16, 18, 160, 160])                                                                                                                                                           
(Pdb) mask.shape                                                                                                                                                                         
    torch.Size([16, 9, 160, 160])                                                                                                                                                            
(Pdb) self.conv2.__doc__                                                                                                                                                                 
    '\n    See :func:`deform_conv2d`.\n    '                                                                                                                                                 
(Pdb) out.shape                                                                                                                                                                          
    torch.Size([16, 128, 160, 160])                                                                                                                                                          
```

When training(run `out = self.conv2(out, offset, mask)` this line), issue the error below, pls help me, tks. 
```
Traceback (most recent call last):                                                                                                                                                       
  File "/DB/train.py", line 89, in <module>                                                                                                                                              
    main()                                                                                                                                                                               
  File "/DB/train.py", line 86, in main                                                                                                                                                  
    trainer.train()                                                                                                                                                                      
  File "/DB/trainer.py", line 115, in train                                                                                                                                              
    epoch=epoch, step=self.steps)                                                                                                                                                        
  File "/DB/trainer.py", line 138, in train_step                                                                                                                                         
    results = model.forward(batch, training=True)                                                                                                                                        
  File "/DB/structure/model.py", line 56, in forward                                                                                                                                     
[...]

PR #4908, opened 2021-11-11
Update the error message on DeformConv2d

datumbox:
The error message says offset while it should have been mask. This affects only the C++ version of the operator and makes it hard to debug what is failing at https://github.com/pytorch/vision/issues/4899#issuecomment-966166101.


NicolasHug (approved):

prabhat00155 (approved):
Thanks @datumbox!

Issue #2707, opened 2020-09-24
Are new models planned to be added?


talcs:
## 🚀 Feature
Adding new models to the models section.

## Motivation

Many new models have been proposed in the recent years and do not exist in the models module.
For example, the EfficientNets appear to provide with 8 models of different complexities that outperform everything else that exists at each complexity level.

## Pitch

Adding the EfficientNets, MobileNet-v3, SE-ResNet and SE-ResNeXt.

oke-aditya:
This request has come often. Just linking all those for reference. #2655. #1697 (this model is coming) 
#2689 (hopefully comes in). Also, this had a discussion about Xception.
Efficient Net request #980 
Some more backbones #2700.

fmassa:
Hi,

To complement @oke-aditya great answer, we will be adding more models to torchvision, including Efficient Nets and MobileNetV3.

The current limitation is that we would like to ensure that we can reproduce the pretrained model using the training scripts from `references/classification, but those models require a different training recipe than then one present in [`references/classification`](https://github.com/pytorch/vision/tree/master/references/classification), so we will need to update those recipes before uploading those new models.

songyuc:
I hope to add Mish activation function.

digantamisra98:
@songyuc There is a closed feature request on PyTorch for adding Mish. You can comment over there for increased visibility so that Mish can be considered to be added in the future. Link to the issue - https://github.com/pytorch/pytorch/issues/25584

WZMIAOMIAO:
first, thanks for your great works.
I hope to add Swish activation and NFNets(High-Performance Large-Scale Image Recognition Without Normalization) https://arxiv.org/abs/2102.06171.
In addition, I would like to ask when eficientnet can be added. I found that it was mentioned in 2019, but now it's 2021. I refer to the [mobilenetV3 model](https://github.com/pytorch/vision/blob/master/torchvision/models/mobilenetv3.py) in torchvision, then I built efficientnet models [Test9_efficientNet](https://github.com/WZMIAOMIAO/deep-learning-for-image-processing/blob/master/pytorch_classification/Test9_efficientNet/model.py), but I don't have a GPU to train with.

oke-aditya:
Hi @WZMIAOMIAO Swish Activation function is added in to PyTorch (not torchvision) as `nn.Silu`. 
Mobilenetv3 would be hopefully available in next release.

WZMIAOMIAO:
[...]

PR #4910, opened 2021-11-11
add efficientnetv2

xiaohu2015:
I want to add efficientnetv2. https://github.com/pytorch/vision/issues/2707

### DIFF
--- a/torchvision/models/efficientnetv2.py
+++ b/torchvision/models/efficientnetv2.py
@@ -1,0 +1,324 @@
+import copy
+import math
+from functools import partial
+from typing import Any, Callable, Optional, List, Sequence
+
+import torch
+from torch import nn, Tensor
+
+from ._utils import _make_divisible
+from .._internally_replaced_utils import load_state_dict_from_url
+from ..ops import StochasticDepth
+from .efficientnet import MBConv, ConvNormActivation
+
+
+__all__ = [
+    "EfficientNetV2", 
+    "efficientnet_v2_s", # 384
+    "efficientnet_v2_m", # 480 
+    "efficientnet_v2_l", # 480 
+]
+
+
+model_urls = {
+    # Weights ported from https://github.com/rwightman/pytorch-image-models/
+    "efficientnet_v2_s": "",
+    'efficientnet_v2_m': "",
+    'efficientnet_v2_l': ""
+}
+
+
+class MBConvConfig:
+    # Stores information listed at Table 1 of the EfficientNet paper
+    def __init__(
+        self,
+        block_type: str,
+        expand_ratio: float,
+        kernel: int,
+        stride: int,
+        input_channels: int,
+        out_channels: int,
+        num_layers: int,
+    ) -> None:
+        self.block_type = block_type
+        self.expand_ratio = expand_ratio
+        self.kernel = kernel
+        self.stride = stride
+        self.input_channels = input_channels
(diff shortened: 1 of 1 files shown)


xiaohu2015:
I implement EfficientNetV2 just like the style of EfficientNetV1 in torchvision models. I also converted the timm pretrained weights, but I found the acc dropped a lot (for example, EfficientNetV2-S is about 82.9 rather 83.9) because of the different padding for 3x3 conv of stride2.

datumbox:
@xiaohu2015 First of all let me confirm that the EfficientNetV2 is within the list of models we would like to include on TorchVision. So the contribution is definitely useful and thank you for it. Moreover your code is of high quality and follows the conventions/style of TorchVision.

Unfortunately accepting model contributions from the community has been challenging in the past, primarily due to training issues. You can read more about this at https://github.com/pytorch/vision/issues/2707#issuecomment-937581682. The TLDR for it is that we don't accept contributions without trained weights which fully reproduce the accuracy of the paper. On the one hand, getting the weights is quite a lot to ask from open-source contributors because training is quite expensive. On the other, a contribution without weights means that an internal maintainer needs to do a significant amount of work to reproduce the paper and that might not be possible due to different priorities. Taking weights from elsewhere is something done in the past but we usually try to avoid for various reasons (lack of reproducibility, minor implementation differences can lead to large accuracy gaps as the ones you report above, adds little value to the community etc).

This is why especially for very large features such as this, it's imperative that we discuss the details prior starting the work. Here are a few things we want to typically clarify:
1. Confirm that the feature is needed 
2. Find one maintainer from our team with the capacity to review/verify/guide the effort
3. Discuss the details of the implementation with the maintainer and agree on the direction

Doing the above help us identify the risks, plan features based on our capacity and ensure we don't throwaway work. I've been an Open-source contributor myself in the past and there is nothing worst than spending couple of weeks to write a feature only to be rejected or to remain unmerged for months due to lack of priority... This is the reason we want to clarify all the above prior starting writing the PR.

Concerning the contents of this PR, here are a few concerns:
1. Though I see you try to inherit components from EfficientNet, there is a lot of code duplication. Ideally we would like to extend V1 in a BC way to accommodate most of the changes of V2. Where we plan to extend V1 and where we introduce a new class for V2 is non-trivial and requires investigation.
2. The PR needs to be accompanied with pre-trained weights, trained from scratch using our reference scripts and fully reproduce the performances of the paper. That is only possible if you have access on hardware that will allow you to train the model.

I think to make this PR mergeable will require a lot of work, so I would like to understand the situation on your side. :)

xiaohu2015:
@datumbox thanks, I am looking foward to your implemention. I will close this pr.

datumbox:
@xiaohu2015 Your PRs are definitely welcome. If you have any other features you would like to work on let's chat about it on their linked issues to map out the details. :)

xiaohu2015:
@datumbox Ok, thanks again.

bratao:
Thank you @xiaohu2015 for this great implementation!

xiaohu2015:
@bratao you can get the original implementation here: https://github.com/xiaohu2015/nncls/blob/main/models/efficientnetv2.py

datumbox:
@xiaohu2015 We should consider adapting the existing efficientnet.py implementation of TorchVision to support both V1 and V2 and port it in. This can wait for after your other PR #4961 about FCOS is merged. :)

PR #4911, opened 2021-11-11
Set seed on `test_nms_ref` to reduce flakiness

datumbox:
The `test_nms_ref` seems to be [flaky](https://app.circleci.com/pipelines/github/pytorch/vision/12449/workflows/ef9169de-d537-4b02-b3d2-1ae9808c6410/jobs/983040/tests). This is related to the unstable sort used by NMS. To resolve this we could fix the seeds. A better solution currently investigated at #4767 is to make the sort stale within NMS.

cc @pmeier


NicolasHug (approved):
Thanks @datumbox

PR #4913, opened 2021-11-11
Default value docs for models using trainable_backbone_layers

### LINKED ISSUE #4905: Clarify documentation/defaults for trainable_backbone_layers
### 📚 The doc issue

The defaults for `trainable_backbone_layers` are currently set to `None` (used in `faster_rcnn`, `keypoint_rcnn`, `mask_rcnn`, and `retinanet`) and the description in `Args:` is a version of:

```
trainable_backbone_layers (int): number of trainable (not frozen) resnet layers starting from final block.
   Valid values are between 0 and 5, with 5 meaning all backbone layers are trainable.
```

This makes it unclear what value is used in the default case. One reasonable reading of this is that none of the backbone layers are trained. The actual value used if this optional argument isn't change is `3`, which is set by `_validate_trainable_layers`, and can only be determined by reading the source.

Either the defaults should be changed to 3 to clearly indicate the default behavior or the docstrings in `faster_rcnn`, `keypoint_rcnn`, `mask_rcnn`, and `retinanet` should be updated to something like:

```
trainable_backbone_layers (int): number of trainable (not frozen) resnet layers starting from final block.
   Valid values are between 0 and 5, with 5 meaning all backbone layers are trainable. 3 is used if this argument is not set.
```

I'd be happy to submit a PR with either of these approaches, but wanted to check if this is useful and which approach is preferred before proceeding.

ethanwhite:
Fixes #4905

Default values for models using `trainable_backbone_layers=None` can be confusing since it is not clear in the top level function documentation what value is used when not setting the optional argument (see #4905).

This PR provides that information for each model using `trainable_backbone_layers=None` in the associated docstring.

### DIFF
--- a/torchvision/models/detection/faster_rcnn.py
+++ b/torchvision/models/detection/faster_rcnn.py
@@ -379,7 +379,8 @@ def fasterrcnn_resnet50_fpn(
         num_classes (int): number of output classes of the model (including the background)
         pretrained_backbone (bool): If True, returns a model with backbone pre-trained on Imagenet
         trainable_backbone_layers (int): number of trainable (not frozen) resnet layers starting from final block.
-            Valid values are between 0 and 5, with 5 meaning all backbone layers are trainable.
+            Valid values are between 0 and 5, with 5 meaning all backbone layers are trainable. If ``None`` is
+            passed (the default) this value is set to 3.
     """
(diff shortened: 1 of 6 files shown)


datumbox (approved):
LGMT, thanks!

PR #4920, opened 2021-11-11
Move cmake_[windows|linux]_gpu to CUDA-11.3

malfet:
Also, made a small change to CUDA install script on windows:
`echo %errorlevel%` would always return 0, and save cuda install log into the home folder rather than in current directory

cc @seemethere

### DIFF
--- a/.circleci/regenerate.py
+++ b/.circleci/regenerate.py
@@ -268,9 +268,9 @@ def cmake_workflows(indentation=6):
         for device in device_types:
             job = {"name": f"cmake_{os_type}_{device}", "python_version": python_version}
 
-            job["cu_version"] = "cu102" if device == "gpu" else "cpu"
+            job["cu_version"] = "cu113" if device == "gpu" else "cpu"
             if device == "gpu" and os_type == "linux":
-                job["wheel_docker_image"] = "pytorch/manylinux-cuda102"
+                job["wheel_docker_image"] = "pytorch/manylinux-cuda113"
             jobs.append({f"cmake_{os_type}_{device}": job})
     return indent(indentation, jobs)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,6 +10,11 @@ if(WITH_CUDA)
   add_definitions(-D__CUDA_NO_HALF_OPERATORS__)
   add_definitions(-DWITH_CUDA)
   set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} --expt-relaxed-constexpr")
+  # CUDA-11.x can not be compiled using C++14 standard on Windows
+  string(REGEX MATCH "^[0-9]+" CUDA_MAJOR ${CMAKE_CUDA_COMPILER_VERSION})
+  if(${CUDA_MAJOR} GREATER 10 AND MSVC)
+    set(CMAKE_CXX_STANDARD 17)
+  endif()
 endif()
 
 find_package(Python3 COMPONENTS Development)
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1644,16 +1644,16 @@ workflows:
           name: cmake_linux_cpu
           python_version: '3.8'
       - cmake_linux_gpu:
-          cu_version: cu102
+          cu_version: cu113
           name: cmake_linux_gpu
           python_version: '3.8'
-          wheel_docker_image: pytorch/manylinux-cuda102
+          wheel_docker_image: pytorch/manylinux-cuda113
       - cmake_windows_cpu:
           cu_version: cpu
           name: cmake_windows_cpu
           python_version: '3.8'
       - cmake_windows_gpu:
-          cu_version: cu102
+          cu_version: cu113
           name: cmake_windows_gpu
           python_version: '3.8'
       - cmake_macos_cpu:
--- a/packaging/pkg_helpers.bash
+++ b/packaging/pkg_helpers.bash
@@ -52,7 +52,6 @@ setup_cuda() {
       else
         export CUDA_HOME=/usr/local/cuda-11.3/
       fi
-      export FORCE_CUDA=1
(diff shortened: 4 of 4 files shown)


datumbox:
@malfet I don't think that works and neither passing `CU_VERSION`. I gave it a try at #4921 and it seems the fix will be more elaborate as it requires installing missing libs, changing the image type etc. I closed it for now and plan to revisit it. If you have pointers let me know. Thanks!

packaging/windows/internal/cuda_install.bat
```diff
@@ -222,9 +224,9 @@ set "NVTOOLSEXT_PATH=%ProgramFiles%\NVIDIA Corporation\NvToolsExt\bin\x64"
 if not exist "%ProgramFiles%\NVIDIA GPU Computing Toolkit\CUDA\v%CUDA_VERSION_STR%\bin\nvcc.exe" (
```
mszhanyi, line 224: Shall we move this section to line 195

mszhanyi:
> Also, made a small change to CUDA install script on windows: `echo %errorlevel%` would always return 0, and save cuda install log into the home folder rather than in current directory

@malfet , errorlevel would be non-zero if setup.exe failed.
https://app.circleci.com/pipelines/github/pytorch/pytorch/372302/workflows/8a242e28-ad44-4af6-a53a-a031cb9e92d2/jobs/15706071

```
C:\w\b\windows\temp_build\cuda>start /wait setup.exe -s thrust_11.3 nvcc_11.3 cuobjdump_11.3 nvprune_11.3 nvprof_11.3 cupti_11.3 cublas_11.3 cublas_dev_11.3 cudart_11.3 cufft_11.3 cufft_dev_11.3 curand_11.3 curand_dev_11.3 cusolver_11.3 cusolver_dev_11.3 cusparse_11.3 cusparse_dev_11.3 npp_11.3 npp_dev_11.3 nvrtc_11.3 nvrtc_dev_11.3 nvml_dev_11.3 -loglevel:6 -log:"C:\w\b\windows\temp_build\cuda/cuda_install_logs" 

C:\w\b\windows\temp_build\cuda>echo -469237760 
-469237760
```
in https://circleci.com/api/v1.1/project/github/pytorch/pytorch/15706071/output/104/0?file=true&allocation-id=6129e01c942d271d4e1ef010-0-build%2F7371EF63

packaging/windows/internal/cuda_install.bat
malfet: Sounds like a good idea. Let me do it as part of separate PR

malfet:
> @malfet , errorlevel would be non-zero if setup.exe failed. 
Yes, but `errorlevel` is not the same as `%errorlevel%`, and as far as I understand, general advise for batch scripts is to use `if errorlevel`, rather than `if %errorlevel%`, see https://ss64.com/nt/errorlevel.html

CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.12)
 project(torchvision)
-set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD 17)
```
datumbox, line 3: Is this necessary? My understanding is that TorchVision aligns with PyTorch
s approach and sets `CMAKE_CXX_STANDARD` to 14.

packaging/build_cmake.sh
```diff
@@ -42,7 +42,14 @@ else
   PYTORCH_MUTEX_CONSTRAINT=''
 fi
 
-conda install -yq \pytorch=$PYTORCH_VERSION $CONDA_CUDATOOLKIT_CONSTRAINT $PYTORCH_MUTEX_CONSTRAINT $MKL_CONSTRAINT -c "pytorch-${UPLOAD_CHANNEL}"
+if [[ "$CU_VERSION" == "cu111" ]]; then
+  # cuda-toolkit=11.1 is only available on conda-forge
+  CONDA_EXTRA_FLAGS="-c conda-forge"
```
datumbox, line 47: @fmassa I remember there was a discussion prior my leave on June about using forge, but not sure what was decided?

.circleci/regenerate.py
```diff
@@ -268,9 +268,9 @@ def cmake_workflows(indentation=6):
         for device in device_types:
             job = {"name": f"cmake_{os_type}_{device}", "python_version": python_version}
 
-            job["cu_version"] = "cu102" if device == "gpu" else "cpu"
+            job["cu_version"] = "cu111" if device == "gpu" else "cpu"
```
datumbox, line 271: Do we have to align with PyTorch and use CUDA 11.3 for which we get nightly binaries OR it's OK to stick with 11.1? Are there any incompatibilities we should be aware of?

datumbox (commented):
@malfet Thanks for your help in fixing this. I've added a couple of questions, let me know your thoughts.

CMakeLists.txt
malfet: This is necessary to fix Windows compilation errors, where CUDA-11.x (which supports C++17 standard) could not be compiled using C++14 standard.

packaging/build_cmake.sh
malfet: conda-forge is still needed if one wants to install CUDA-11.1, but I will probably move the config to CUDA-11.3 and be done with it.

CMakeLists.txt
datumbox: Thanks for the clarification. Makes sense. So I guess PyTorch Core will be moving as well to C++17, right?

.circleci/regenerate.py
malfet: Yes, moving it to 11.3, don't want to mess us with 11.1 here

malfet:
Ok, splitting this PR into 3: 
https://github.com/pytorch/vision/pull/4944 <- cherry-picks changes from audio back to vision that fixes CMake builds on Linux
https://github.com/pytorch/vision/pull/4945 <- Fix CUDA-11 Windows CMake builds

datumbox (approved):
LGTM, thanks!

Issue #4509, opened 2021-09-30
Replace uses of `apex.amp` with PyTorch's amp in references


fmassa:
Some reference scripts have added support for AMP via APEX a couple of years ago, see https://github.com/pytorch/vision/blob/4bf608633f9299aeda61b08ae126961acaadec22/references/classification/train.py#L17 for one example

PyTorch [now natively supports AMP in the `torch.cuda.amp` namespace](https://pytorch.org/docs/stable/amp.html?highlight=amp#module-torch.cuda.amp), so we should use PyTorch's AMP instead of apex for simplicity.

The replacement needs to happen on the following scripts:
- [x] Classification
- [x] Detection #4933
- [x] Segmentation #4994
- [ ] [Video Classification](https://github.com/pytorch/vision/blob/main/references/video_classification/train.py)

cc @datumbox

datumbox:
Do you know if manually installing apex from the repo will still be needed? Cause if this step is removed it would be great.

fmassa:
We won't need to install apex anymore, and everything will be handled by PyTorch now. We only used apex for amp in the reference scripts

prabhat00155:
The name of the argument is [apex](https://github.com/pytorch/vision/blob/5f0edb97b46e5bff71dc19dedef05c5396eeaea2/references/classification/train.py#L388). Now that we are going to use Pytorch's amp for mixed precision training, should we update this or keep it for BC?

datumbox:
@prabhat00155 I don't think it's a problem to change the name. We don't have any models that actually use it for training and BC on the references is not our primary focus anyway. My view is that it's fine to update it.

fmassa:
I second @datumbox comment, let's just change the argument name -- note that the reference scripts are not part of our API

kbozas:
I am interested in the Video Classification bit.

datumbox:
Sounds good @kbozas. Feel free to add me as a Reviewer on your upcoming PR.

PR #4923, opened 2021-11-12
save grad_scaler if use amp for better resume

xiaohu2015:
The pr is about https://github.com/pytorch/vision/issues/4509
when use amp, it is better to save/resuem scaler state_dict https://pytorch.org/tutorials/recipes/recipes/amp_recipe.html#saving-resuming

cc @datumbox


datumbox (approved):
LGTM, thanks!

PR #4926, opened 2021-11-12
Android app switch to pytorch_android stable

datumbox:
This PR does 2 things:
- Fixes a dependency problem our Android demo app
- Switches the version to stable to align with the iOS demo app


jdsgomes (approved):
stamping

PR #4927, opened 2021-11-12
[WIP] Align versioning logic across Android and iOS demo apps

datumbox:
In this PR:
- We align the versioning names across Android and iOS.
- We switch Android to use the latest snapshot dependencies.


android/gradle.properties
```diff
@@ -1,6 +1,6 @@
 ABI_FILTERS=armeabi-v7a,arm64-v8a,x86,x86_64
 
-VERSION_NAME=0.10.0-SNAPSHOT
+VERSION_NAME=0.12.0-SNAPSHOT
```
datumbox, line 3: FYI this version hasn't been bumped for several TorchVision releases...

ios/LibTorchvision.podspec
```diff
@@ -1,8 +1,8 @@
-pytorch_version = '1.10.0'
+pytorch_version = '1.11.0-SNAPSHOT'
```
datumbox, line 1: The iOS setup job actually installs the nightly versions, so I guess marking it like that makes sense? On the other hand, `-SNAPSHOT` is Java lingo. Perhaps replace with `-nightly`?

datumbox:
I spoke with @IvanKobzarev and told me that there is a bug on the current pipeline which means that they don't publish on the -SNAPSHOT versions the right files. I'll close this for now.

datumbox:
The _lite reported issue from above is fixed at https://github.com/pytorch/pytorch/pull/68651

PR #4931, opened 2021-11-14
Fix clang_format CI failure

datumbox:
Fixes recent CI [failures](https://app.circleci.com/pipelines/github/pytorch/vision/12538/workflows/ba382961-a261-40b4-938a-c20558aaecda/jobs/991776) on the clang_format job caused by a missing lib.

cc @seemethere

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -225,6 +225,8 @@ jobs:
       - checkout
       - run:
           command: |
+            sudo apt-get update -y
+            sudo apt install -y libtinfo5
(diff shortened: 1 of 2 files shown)


NicolasHug (approved):
Thanks @datumbox

jdsgomes (approved):
LGTM. Just would like to understand if it was working in the past and what was the change that broke the process?

datumbox:
@jdsgomes It was working in the past and broke at one point last week. Could be due to a change on the OS level. The dependency is no longer installed and it's now missing.

PR #4937, opened 2021-11-15
NOMRG Comments on prototype weights and model builders

NicolasHug:
Here are a bunch of comments on the new weights prototype. I'm opening this as a PR (not to be merged) instead of an issue as I think it will better help localize each discussion.

Overall I don't have any major comment or concern, except maybe for the somewhat BC breaking changes of some model building functions. I think we can work around BC for some of them, but not all.

### DIFF
--- a/torchvision/prototype/models/_api.py
+++ b/torchvision/prototype/models/_api.py
@@ -51,6 +51,19 @@ class Weights(Enum):
             if type(obj) is str:
                 obj = cls.from_str(obj)
             elif not isinstance(obj, cls) and not isinstance(obj, WeightEntry):
+                # Allowing WeightEntry to pass-through is unexpected IMHO.
+                # Since verify() is a factory method, we would expect it to
+                # return only Weights instances, not WeightsEntry instances.
+                # This could lead to bugs if the caller is unaware of that
+                # subtly hidden detail.
+                # From what I understand the motivation for allowing WeightEntry
+                # is for user convenience when calling model builder functions
+                # like resnet18(). IMO for a first version, it's perfectly
+                # reasonable to expect users to write a small wrapper class that
+                # inherits from Weights instead. Maybe this could be re-visited
+                # in the future, but for a first version, I think we should try
+                # to be conservative: if there's a decent way of achieving a
+                # somewhat rare use-case, it's not worth worrying about for now.
                 raise TypeError(
                     f"Invalid Weight class provided; expected {cls.__name__} but received {obj.__class__.__name__}."
                 )
@@ -63,6 +76,11 @@ class Weights(Enum):
                 return v
         raise ValueError(f"Invalid value {value} for enum {cls.__name__}.")
 
+    # Nit: I would expect 'state_dict' to be an attribute, not a method.
+    # I understand we want to avoid
+    # model.load_state_dict(weights.load_state_dict(progress=progress))
+    # which reads strangely.
+    # Maybe `get_state_dict()` would better convey the idea that this is a procedure?
     def state_dict(self, progress: bool) -> OrderedDict:
         return load_state_dict_from_url(self.url, progress=progress)
 
--- a/torchvision/prototype/models/densenet.py
+++ b/torchvision/prototype/models/densenet.py
@@ -26,6 +26,8 @@ __all__ = [
 
 
 def _load_state_dict(model: nn.Module, weights: Weights, progress: bool) -> None:
+    # Could you explain the comments below a bit more? Why are dots no longer allowed?
+
     # '.'s are no longer allowed in module names, but previous _DenseLayer
     # has keys 'norm.1', 'relu.1', 'conv.1', 'norm.2', 'relu.2', 'conv.2'.
     # They are also in the checkpoints in model_urls. This pattern is used
@@ -67,6 +69,9 @@ _common_meta = {
     "size": (224, 224),
     "categories": _IMAGENET_CATEGORIES,
     "interpolation": InterpolationMode.BILINEAR,
+    # Is this worth adding literally this comment as the value? e.g.
+    # "recipe": "weights ported from LuaTorch",
+    # ?
     "recipe": None,  # weights ported from LuaTorch
 }
 
--- a/torchvision/prototype/models/detection/faster_rcnn.py
+++ b/torchvision/prototype/models/detection/faster_rcnn.py
@@ -207,7 +207,7 @@ def fasterrcnn_mobilenet_v3_large_320_fpn(
         # I don't have a better proposal TBH.
         # I think the complexity of those weight names can be greatly reduced by:
         # - good docs (which I know is part of the future work)
-        # - good defaults like "pretrained" or "latest" or whatever is easy to remember for the user, as suggested in another comment
+        # - good "preset values" like "pretrained" or "latest" or whatever is easy to remember for the user, as suggested in another comment
         weights = FasterRCNNMobileNetV3Large320FPNWeights.Coco_RefV1 if kwargs.pop("pretrained") else None
     weights = FasterRCNNMobileNetV3Large320FPNWeights.verify(weights)
     if "pretrained_backbone" in kwargs:
--- a/torchvision/prototype/models/detection/faster_rcnn.py
+++ b/torchvision/prototype/models/detection/faster_rcnn.py
@@ -147,11 +147,25 @@ def _fasterrcnn_mobilenet_v3_large_fpn(
     return model
 
 
+# Unfortunately, the addition and removal of parameters here is BC breaking for
+# users that are using positional arguments.
+#
+# It might be possible to preserve BC by allowing:
+# weights to be a bool - it maps to pretrained
+# weights_backbone to be a bool - it maps to progress
+# progress to be an int - it maps to num_classes
+# num_classes to be a bool: it maps to pretrained_backbones
+# And to raise a warning in all these cases.
+#
+# That will complicate the code though, we should only do this if we are 100%
+# sure to remove these deprecated behaviours in 1 or 2 versions.
+# Also, this only works because the types of the new vs old parameters order
+# don't overlap.
 def fasterrcnn_mobilenet_v3_large_fpn(
     weights: Optional[FasterRCNNMobileNetV3LargeFPNWeights] = None,
     weights_backbone: Optional[MobileNetV3LargeWeights] = None,
     progress: bool = True,
-    num_classes: int = 91,
+    num_classes: int = 91,  # just wondering why num_classes is a parameter here instead of being part of kwargs like for other models?
     trainable_backbone_layers: Optional[int] = None,
     **kwargs: Any,
 ) -> FasterRCNN:
--- a/torchvision/prototype/models/detection/faster_rcnn.py
+++ b/torchvision/prototype/models/detection/faster_rcnn.py
@@ -203,6 +203,11 @@ def fasterrcnn_mobilenet_v3_large_320_fpn(
 ) -> FasterRCNN:
     if "pretrained" in kwargs:
         warnings.warn("The argument pretrained is deprecated, please use weights instead.")
+        # These long weight names are a bit of a mouthful lol
+        # I don't have a better proposal TBH.
+        # I think the complexity of those weight names can be greatly reduced by:
+        # - good docs (which I know is part of the future work)
+        # - good defaults like "pretrained" or "latest" or whatever is easy to remember for the user, as suggested in another comment
         weights = FasterRCNNMobileNetV3Large320FPNWeights.Coco_RefV1 if kwargs.pop("pretrained") else None
     weights = FasterRCNNMobileNetV3Large320FPNWeights.verify(weights)
     if "pretrained_backbone" in kwargs:
--- a/torchvision/prototype/models/detection/keypoint_rcnn.py
+++ b/torchvision/prototype/models/detection/keypoint_rcnn.py
@@ -48,6 +48,14 @@ class KeypointRCNNResNet50FPNWeights(Weights):
     )
 
 
+# This is BC breaking too, and unfortunately I'm not sure we can raise warnings
+# like in fasterrcnn_mobilenet_v3_large_fpn because we wouldn't be able
+# to differentiate num_keypoints from num_classes.
+#
+# I'm not sure what the best course of action could be here
+#
+# Not a solution, but related: these issues are a strong argument in favour of
+# forcing keyword-only parameters in new APIs.
 def keypointrcnn_resnet50_fpn(
     weights: Optional[KeypointRCNNResNet50FPNWeights] = None,
     weights_backbone: Optional[ResNet50Weights] = None,
--- a/torchvision/prototype/models/googlenet.py
+++ b/torchvision/prototype/models/googlenet.py
@@ -37,8 +37,10 @@ def googlenet(weights: Optional[GoogLeNetWeights] = None, progress: bool = True,
     original_aux_logits = kwargs.get("aux_logits", False)
     if weights is not None:
         if "transform_input" not in kwargs:
+            # Similar to the comment about num_classes, I think we should raise and error if the user manually specified transform_input=False
             kwargs["transform_input"] = True
         if original_aux_logits:
+            # Nit: maybe move this as an `else` clause below
             warnings.warn(
                 "auxiliary heads in the pretrained googlenet model are NOT pretrained, so make sure to train them"
             )
@@ -50,6 +52,7 @@ def googlenet(weights: Optional[GoogLeNetWeights] = None, progress: bool = True,
 
     if weights is not None:
         model.load_state_dict(weights.state_dict(progress=progress))
+        # I understand this is present in the current code, just curious why this is needed?
         if not original_aux_logits:
             model.aux_logits = False
             model.aux1 = None  # type: ignore[assignment]
--- a/torchvision/prototype/models/resnet.py
+++ b/torchvision/prototype/models/resnet.py
@@ -41,6 +41,16 @@ def _resnet(
     **kwargs: Any,
 ) -> ResNet:
     if weights is not None:
+        # I remember Francisco raising concern about overriding num_class.
+        # IMHO this is fine, I would consider this to be a case of
+        # "we have param A and param B, param B's default depends on the value of A."
+        # which is a very common pattern.
+        # However, in order to prevent users to do wrong things, I think we could raise
+        # an error if the user specified an incorrect 'num_classes':
+        if "num_classes" in kwargs and kwargs["num_classes"] != len(weights.meta["categories"]):
+            raise ValueError(
+                f"Oops, you specified num_classes={blah} but this is incompatible with the pre-trained weights which support {blop} classes."
+            )
         kwargs["num_classes"] = len(weights.meta["categories"])
 
     model = ResNet(block, layers, **kwargs)
@@ -51,6 +61,7 @@ def _resnet(
     return model
 
 
+# Nit: upper-case for global vars?
 _common_meta = {"size": (224, 224), "categories": _IMAGENET_CATEGORIES, "interpolation": InterpolationMode.BILINEAR}
 
 
@@ -231,10 +242,27 @@ class WideResNet101_2Weights(Weights):
     )
 
 
+# Maybe we should rename kwargs into model_kwargs to be more explicit? This is not a BC-breaking change I believe.
 def resnet18(weights: Optional[ResNet18Weights] = None, progress: bool = True, **kwargs: Any) -> ResNet:
+    # Considering we're showing this warning literally everywhere, it might be worth writing a helper for it
+    # This would make sure the warning message is consistent, avoid code duplication, etc.
     if "pretrained" in kwargs:
-        warnings.warn("The argument pretrained is deprecated, please use weights instead.")
+        # Nit: "argument" should probably be "parameter", since the term
+        # "argument" refers to the _value_ of the parameter at call time.
+        warnings.warn("The 'pretrained' parameter is deprecated, please use the 'weights' parameter instead.")
         weights = ResNet18Weights.ImageNet1K_RefV1 if kwargs.pop("pretrained") else None
+
+    # To preserve full BC with the current resnet18, weights coudl also accept
+    # a boolean and this would issue deprecation warning as well. This is for
+    # cases where resnet18 is called with resnet18(True) or resnet18(False)
+    # something like this:
+    if isinstance(weights, bool):
+        warnings.warn("The 'pretrained' parameter is deprecated, please use the 'weights' parameter instead.")
+        weights = ResNet18Weights.ImageNet1K_RefV1 if weights else None
+
+    # I remember we discussed this before: before releasing this new API, we
+    # mght want to try to allow users to pass simple stuff like e.g.
+    # resnet18(weights='pretrained') or resnet18(weights='latest')
 
     weights = ResNet18Weights.verify(weights)
 
--- a/torchvision/prototype/models/resnet.py
+++ b/torchvision/prototype/models/resnet.py
@@ -248,7 +248,7 @@ def resnet18(weights: Optional[ResNet18Weights] = None, progress: bool = True, *
     # This would make sure the warning message is consistent, avoid code duplication, etc.
     if "pretrained" in kwargs:
         # Nit: "argument" should probably be "parameter", since the term
-        # "argument" refers to the _value_ of the parameter at call time.
+        # "argument" refers to the *value* of the parameter at call time.
         warnings.warn("The 'pretrained' parameter is deprecated, please use the 'weights' parameter instead.")
         weights = ResNet18Weights.ImageNet1K_RefV1 if kwargs.pop("pretrained") else None
 


torchvision/prototype/models/resnet.py
```diff
@@ -41,6 +41,16 @@ def _resnet(
     **kwargs: Any,
 ) -> ResNet:
     if weights is not None:
+        # I remember Francisco raising concern about overriding num_class.
+        # IMHO this is fine, I would consider this to be a case of
+        # "we have param A and param B, param B's default depends on the value of A."
+        # which is a very common pattern.
+        # However, in order to prevent users to do wrong things, I think we could raise
+        # an error if the user specified an incorrect 'num_classes':
```
datumbox, line 49: In the prototype there are 3 different overwrites similar to this. (all other overwrites are pre-existing)

The 1st is for the `num_classes` key of `kwargs` as in this case. This applies to all classification models.

The 2nd is for the `backend` key of `kwargs` (the case is identical to previous). This applies to all the quantized models like this:
https://github.com/pytorch/vision/blob/272e080c46ebef2b74f89c9cc94d84a9dc541bde/torchvision/prototype/models/quantization/mobilenetv3.py#L38

The 3rd is for the `num_classes` parameter of detection/segmentation models:
https://github.com/pytorch/vision/blob/272e080c46ebef2b74f89c9cc94d84a9dc541bde/torchvision/prototype/models/detection/faster_rcnn.py#L94

The latter case is a bit of an oddity as highlighted earlier by Nicolas because that's a parameter defined for Detection and Segmentation but not Classification. As you understand, when it's a parameter and not a `kwargs` value, we can't tell if it was the user who passed this or if they relied on the default value of the method and they forgot to pass the right value. So it's unclear how to implement the proposal below in all cases.

Perhaps given we already do these silent overwrites in a bunch of places (see GoogleNet, InceptionV3, ResNeXt etc), we don't have to throw a warning/error. Thoughts @NicolasHug @fmassa ?

torchvision/prototype/models/googlenet.py
```diff
@@ -37,8 +37,10 @@ def googlenet(weights: Optional[GoogLeNetWeights] = None, progress: bool = True,
     original_aux_logits = kwargs.get("aux_logits", False)
     if weights is not None:
         if "transform_input" not in kwargs:
+            # Similar to the comment about num_classes, I think we should raise and error if the user manually specified transform_input=False
```
datumbox, line 40: This is actually part of the original implementation, though I tried to simplify its unnecessarily complex logic a bit:
https://github.com/pytorch/vision/blob/72054ca05870174b18a1421479e76a3da5e39782/torchvision/models/googlenet.py#L293-L303

So parameter overwriting is something that used to happen in the past and changing the behaviour now to throw errors will be BC-breaking. cc @fmassa

torchvision/prototype/models/resnet.py
```diff
@@ -231,11 +242,28 @@ class WideResNet101_2Weights(Weights):
     )
 
 
+# Maybe we should rename kwargs into model_kwargs to be more explicit? This is not a BC-breaking change I believe.
```
datumbox, line 245: It shouldn't be BC breaking you are right. But I think `kwargs` is more common idiom and renaming it might be confusing to users? Not sure.

torchvision/prototype/models/resnet.py
```diff
@@ -231,11 +242,28 @@ class WideResNet101_2Weights(Weights):
...
 def resnet18(weights: Optional[ResNet18Weights] = None, progress: bool = True, **kwargs: Any) -> ResNet:
+    # Considering we're showing this warning literally everywhere, it might be worth writing a helper for it
+    # This would make sure the warning message is consistent, avoid code duplication, etc.
```
datumbox, line 248: Yes indeed that's the plan. Just wanted to leave it for last moment until I have a clearer idea of how the implementation works for Quantization, Detection etc. I'll make this change soon.

torchvision/prototype/models/resnet.py
```diff
@@ -231,11 +242,28 @@ class WideResNet101_2Weights(Weights):
...
     if "pretrained" in kwargs:
-        warnings.warn("The argument pretrained is deprecated, please use weights instead.")
+        # Nit: "argument" should probably be "parameter", since the term
+        # "argument" refers to the *value* of the parameter at call time.
```
datumbox, line 251: Sounds good.

torchvision/prototype/models/resnet.py
```diff
@@ -231,11 +242,28 @@ class WideResNet101_2Weights(Weights):
...
+        warnings.warn("The 'pretrained' parameter is deprecated, please use the 'weights' parameter instead.")
         weights = ResNet18Weights.ImageNet1K_RefV1 if kwargs.pop("pretrained") else None
 
+    # To preserve full BC with the current resnet18, weights coudl also accept
+    # a boolean and this would issue deprecation warning as well. This is for
+    # cases where resnet18 is called with resnet18(True) or resnet18(False)
+    # something like this:
+    if isinstance(weights, bool):
```
datumbox, line 259: Agreed. I bet you wrote this comment first and I'm reading it last. Happy to make this change if we decide for full-BC.

torchvision/prototype/models/resnet.py
```diff
@@ -231,11 +242,28 @@ class WideResNet101_2Weights(Weights):
...
+        warnings.warn("The 'pretrained' parameter is deprecated, please use the 'weights' parameter instead.")
+        weights = ResNet18Weights.ImageNet1K_RefV1 if weights else None
+
+    # I remember we discussed this before: before releasing this new API, we
+    # mght want to try to allow users to pass simple stuff like e.g.
+    # resnet18(weights='pretrained') or resnet18(weights='latest')
```
datumbox, line 265: Concerning `weights='latest'`, yes we will eventually support it. We had an [early version](https://github.com/datumbox/dapi-model-versioning/blob/main/dapi_lib/models/_api.py#L157-L159) of this on an old prototype. This will be handled once we have the registration mechanism in place. I'm delaying including this to see if we can achieve some cross-domain agreement. Could you clarify if you think this is a prerequisite for releasing or if this can be implemented later? I assume your concern is about TorchHub compatibilities.

Could you clarify if you literally meant that we should support `weights='pretrained'`? If yes, what does this do? Note that currently we support `weights="ImageNet1K_RefV1"`. Would that be good enough?

datumbox (commented):
Thanks for the comments @NicolasHug.

I'm going to go ahead and implement many of the proposals and send a PR (I'll add you as reviewer). 

I also added more comments/context to those that I require more clarifications. Let me know your thoughts.

torchvision/prototype/models/densenet.py
```diff
@@ -67,6 +69,9 @@ def _densenet(
     "size": (224, 224),
     "categories": _IMAGENET_CATEGORIES,
     "interpolation": InterpolationMode.BILINEAR,
+    # Is this worth adding literally this comment as the value? e.g.
+    # "recipe": "weights ported from LuaTorch",
+    # ?
```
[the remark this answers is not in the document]
NicolasHug: OK, let's not worry about this for now and we'll just keep this in mind for when we work on the doc revamp.

torchvision/prototype/models/googlenet.py
NicolasHug: I wouldn't consider this a BC-breaking change, since we're preventing the user from doing something they shouldn't do. I think it's fair to consider this a bug-fix :)

torchvision/prototype/models/googlenet.py
```diff
@@ -37,8 +37,10 @@ def googlenet(weights: Optional[GoogLeNetWeights] = None, progress: bool = True,
...
             kwargs["transform_input"] = True
         if original_aux_logits:
+            # Nit: maybe move this as an `else` clause below
```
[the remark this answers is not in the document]
NicolasHug: Further down (current line 56) we have `if not original_aux_logits:` so I think we could just move that part down there as an `else` block. Just a nit though :)

torchvision/prototype/models/resnet.py
NicolasHug: > when it's a parameter and not a kwargs value, we can't tell if it was the user who passed this or if they relied on the default value of the method and they forgot to pass the right value

That's a good point.

My suggestion then would be to change the default of `num_classes=91` (I don't remember the exact number) to `num_classes='auto'`, which would:
- defaults to 91 if weights is None - this preserves BC
- defaults to `len(weights.meta["categories"])` if  weights is passed

I think this is backward compatible, and will allow us to safely raise an error when the user manually overwrites `num_classes` to a wrong value.

torchvision/prototype/models/resnet.py
NicolasHug: > Could you clarify if you think this is a prerequisite for releasing or if this can be implemented later?

I wouldn't make this a requirement for merging the prototype into `main`, but I think we should support this for the release. Currently users can get pretrained weights with a super simple api, just passing `pretrained=True`. Now we offer a more powerful API but we should still try to support these very simple workflows IMO. If we're telling them "don't use this, use that instead", using "that" should be as painless as possible. Ideally we would just tell them in the deprecation warning something like

> The 'pretrained' parameter is deprecated, please use the 'weights' parameter instead. 'pretrained=True' is now equivalent to 'weights="latest"' (or 'pretrained' or 'default' or whatever we want to decide)





> Could you clarify if you literally meant that we should support weights='pretrained'?

Maybe 'pretrained' isn't a good name. But I think that ideally we would allow users to get both a) the latest weights and b) the "default" weights (which would never change): the latest weights might change across versions and for some users the changes might not be desirable.

torchvision/prototype/models/googlenet.py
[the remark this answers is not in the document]
datumbox: Oh yeah, totally. I'll make the change on a PR.

torchvision/prototype/models/resnet.py
datumbox: Yes, this idea is possible and BC. 

Instead of `'auto'` we could use `None` as this is a common idiom used in TorchVision. Unfortunately, this idiom has caused complaints (see #4905) among users. Some extra handling will be necessary to achieve full BC, for example if num_classes=None and weights=None then it should become `91` but all these are possible.

I'll wait for @fmassa to comment on the pending issues before tackling this as the change is big and in multiple places.

torchvision/prototype/models/resnet.py
datumbox: > But I think that ideally we would allow users to get both a) the latest weights and b) the "default" weights (which would never change)

No concerns for a), it's going to be supported soon.

Concerning b), it was a conscious choice to set `weights=None` for default instead of `weights=SomeValue` in the method signature. The idea was that the default weights might become obsolete and then we would be stuck keeping them in the signature due to the BC considerations of TorchVision. By forcing the users to explicitly define the weights we avoid that. I think asking the user to pass as a string the weights that they want, is a reasonable compromise between easy of use and keeping the API fresh?

torchvision/prototype/models/_api.py
```diff
@@ -51,6 +51,19 @@ def verify(cls, obj: Any) -> Any:
...
+                # This could lead to bugs if the caller is unaware of that
+                # subtly hidden detail.
+                # From what I understand the motivation for allowing WeightEntry
+                # is for user convenience when calling model builder functions
+                # like resnet18(). IMO for a first version, it's perfectly
+                # reasonable to expect users to write a small wrapper class that
+                # inherits from Weights instead. Maybe this could be re-visited
+                # in the future, but for a first version, I think we should try
+                # to be conservative: if there's a decent way of achieving a
+                # somewhat rare use-case, it's not worth worrying about for now.
```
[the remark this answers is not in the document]
datumbox: Addressed at commit 21569339074756d877bf527718e54b14eaa4496e of PR #4940

torchvision/prototype/models/_api.py
```diff
@@ -63,6 +76,11 @@ def from_str(cls, value: str) -> "Weights":
                 return v
         raise ValueError(f"Invalid value {value} for enum {cls.__name__}.")
 
+    # Nit: I would expect 'state_dict' to be an attribute, not a method.
+    # I understand we want to avoid
+    # model.load_state_dict(weights.load_state_dict(progress=progress))
+    # which reads strangely.
+    # Maybe `get_state_dict()` would better convey the idea that this is a procedure?
```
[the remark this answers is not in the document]
datumbox: Addressed at commit bfd8a21ec6fd42aab5462078aa0235de96dde934 of PR #4940

torchvision/prototype/models/densenet.py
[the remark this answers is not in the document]
datumbox: Addressed at commit 3cc107c396975eb197faf02072769c70bc0818db of PR #4940

torchvision/prototype/models/resnet.py
NicolasHug: I'm not suggesting to change the default of `weights`, it's fine to leave it as None.

What I'm saying is that we might want to support an equivalent behaviour as the current `pretrained=True`, i.e. allow the user to pass `weights=<something>` (where `<something>` is to be determined) that will always return the same pre-trained weights.

torchvision/prototype/models/googlenet.py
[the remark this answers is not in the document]
datumbox: Addressed at commit 9ff5f6e44b108452214550b81ecac4e1ce6f32ed of PR #4940

torchvision/prototype/models/resnet.py
```diff
@@ -51,6 +61,7 @@ def _resnet(
     return model
 
 
+# Nit: upper-case for global vars?
```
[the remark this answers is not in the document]
datumbox: Addressed at commit 9a6691147ee67452bb8d8dce40c281304ec1f706 of PR #4940

torchvision/prototype/models/googlenet.py
datumbox: So you are saying that if we change our policy and start throwing exceptions, instead of silently overwriting variables, you would be OK and consider it not BC-breaking but a bug fix?

torchvision/prototype/models/detection/faster_rcnn.py
```diff
@@ -189,6 +203,11 @@ def fasterrcnn_mobilenet_v3_large_320_fpn(
 ) -> FasterRCNN:
     if "pretrained" in kwargs:
         warnings.warn("The argument pretrained is deprecated, please use weights instead.")
+        # These long weight names are a bit of a mouthful lol
+        # I don't have a better proposal TBH.
+        # I think the complexity of those weight names can be greatly reduced by:
+        # - good docs (which I know is part of the future work)
+        # - good "preset values" like "pretrained" or "latest" or whatever is easy to remember for the user, as suggested in another comment
```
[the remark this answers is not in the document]
NicolasHug: > We could reduce slightly their size by:

I agree these changes might bring a bit of confusion, so for now I would keep the names as they are. Long name are better than confusing ones :)

torchvision/prototype/models/resnet.py
datumbox: > I'm not suggesting to change the default of weights, it's fine to leave it as None.

I understood that. Sorry I should have been more explicit on how I made the link between `weights='pretrained'` and the default values of `weights` in the signature.

The reason I linked the two is because IMO having a non-None value as default has exactly the same effect as supporting something like `weights='pretrained'`. They anchor us to a specific old and potentially superseded set of weights. That's why I'm in favour of supporting still a `weights=<something>` but that something shouldn't be constant across all models but rather explicitly define which set of weights to load. This is currently already supported because the weights can be a string with the name of the Enum.

In other words, I'm hoping to avoid introducing anything on the API that will anchor the user to legacy weights. I'm happy to support a `weights='latest'` thought, for those who are happy to always access the best available weights for a specific model architecture.

torchvision/prototype/models/googlenet.py
NicolasHug: yes :)

torchvision/prototype/models/resnet.py
NicolasHug: Thanks for clarifying @datumbox , that's fair.

My main concern here is to be as nice as possible to users and help them transitioning from the old to the new API. Without introducing a `weights='pretrained'` option, perhaps it would be enough to expand the deprecation message a bit to e.g.:

> The 'pretrained' parameter is deprecated, please use the 'weights' parameter instead. 'pretrained=True' is now equivalent to 'weights="TheWeightsNameWithIsSpecificToEachModelBuilder"'; you can also use 'pretrained="latest"' to get the most up-to-date weights for this model. For more details see \<link to doc note where we explain the changes introduced with this API\>.

(instead of suggesting a string like "TheWeightsNameWithIsSpecificToEachModelBuilder" we can also suggest the enum field, not strong opinion).

NicolasHug: > I wouldn't make this a requirement for merging the prototype into main, but I think we should support this for the release

BTW, if we go with a fully detailed deprecation message as above, I would say that it's OK to release without the weights='lastest' option (although still preferable)

torchvision/prototype/models/detection/faster_rcnn.py
```diff
@@ -147,11 +147,25 @@ def _fasterrcnn_mobilenet_v3_large_fpn(
...
+# weights to be a bool - it maps to pretrained
+# weights_backbone to be a bool - it maps to progress
+# progress to be an int - it maps to num_classes
+# num_classes to be a bool: it maps to pretrained_backbones
+# And to raise a warning in all these cases.
+#
+# That will complicate the code though, we should only do this if we are 100%
+# sure to remove these deprecated behaviours in 1 or 2 versions.
+# Also, this only works because the types of the new vs old parameters order
+# don't overlap.
```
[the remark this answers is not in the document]
NicolasHug: It's a tough decision...
I feel like BC is probably *the* most important contract that we have with our users. I would assume that if there are occurrences of positional arguments in fbcode, we can expect that there are a whole lot of them out in the world.

I wouldn't want to delay the new API either though. My preferred course of action would be to preserve BC. For cases like `resnet18` it should be fairly easy to keep BC and still do the clean up (as per the suggestion below). For cases like here or `keypointrcnn_resnet50_fpn`, that probably means we can't "clean up" the API - but it's a fair price to pay I think.

----

Then, irrespective of this new API, and for the upcoming release, I would suggest to start allowing the the clean ups by e.g. forcing users to use keyword-only arguments (this can be done smoothly with stuff like https://scikit-learn-enhancement-proposals.readthedocs.io/en/latest/slep009/proposal.html#deprecation-path). I think our future-selves will thank us in 2-3 versions.

torchvision/prototype/models/resnet.py
datumbox: I agree with the approach of the fully detailed deprecation message. I think it's a reasonable compromise.

For the `weights='latest'`, I don't really have concerns and I would actually like to give the option to users to access the best/greatest available weights. Not all people in the industry are interested in BC guarantees on the weights side. Sometimes you just want to get the best one and continue your work on transfer learning.

torchvision/prototype/models/_api.py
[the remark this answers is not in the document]
fmassa: Indeed, after discussing with Kai having the ability to pass the path to the pre-trained weights seems to be something they have been traditionally doing (as it also simplifies doing things through configs)

My suggestion was to have a helper function that basically did
```python
def load_state_dict_in_model(model, path):
    state_dict = torch.load(path)
    model.load_state_dict(state_dict)
    return model
```

which I believe covers all use-cases. It adds extra complexity (as it requires one more function call in the config), but I think it might be the best trade-off for explicitly saying that we won't be doing any magic in there.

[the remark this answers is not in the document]
NicolasHug: @fmassa are you suggesting to add such a helper in torchvision, or instead to suggest users to implement this themselves?

torchvision/prototype/models/densenet.py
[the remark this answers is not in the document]
fmassa: I would maybe put the link to the original repo in Lua-Torch, or at least the PR that added this model, so that it gives more information to the user

torchvision/prototype/models/_api.py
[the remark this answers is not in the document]
datumbox: I would like more clarifications on @fmassa's proposal. I wonder if the same thing can be currently achieved by patching the `Weights.get_state_dict()` method, rather than trying to pass callables.

We should note that there are a few places where we need to modify the weights prior loading:
https://github.com/pytorch/vision/blob/d367a01a18a3ae6bee13d8be3b63fd6a581ea46f/torchvision/models/densenet.py#L232-L239

torchvision/prototype/models/detection/faster_rcnn.py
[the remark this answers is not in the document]
fmassa: I think the most important one to keep is just `resnet50(True)` (i.e., for the `pretrained` argument being the first one), the others I think we don't need to worry about positional arguments.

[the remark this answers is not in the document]
datumbox: Unless I'm missing something, it's possible to do fully-BC across all builders by moving the parameters on the original position and applying workaround like the one proposed by @NicolasHug here:
https://github.com/pytorch/vision/blob/500f826e97b20848863574523227644b40a6bf69/torchvision/prototype/models/resnet.py#L255-L261

We can do the same for `weights_backbone`. 

I think from both your feedback it's clear that maintaining compatibility for things like `resnet50(True)` is important and perhaps we should aim for full-BC until we decide to clean up the API on a separate coordinated initiative.

torchvision/prototype/models/detection/faster_rcnn.py
```diff
@@ -147,11 +147,25 @@ def _fasterrcnn_mobilenet_v3_large_fpn(
...
 def fasterrcnn_mobilenet_v3_large_fpn(
     weights: Optional[FasterRCNNMobileNetV3LargeFPNWeights] = None,
     weights_backbone: Optional[MobileNetV3LargeWeights] = None,
     progress: bool = True,
-    num_classes: int = 91,
+    num_classes: int = 91,  # just wondering why num_classes is a parameter here instead of being part of kwargs like for other models?
```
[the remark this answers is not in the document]
fmassa: I think we can start using more keyword-only arguments in torchvision now. We didn't do it before because we supported Python 2 as well, but now that our minimum Python version is 3.6 this can be added.

torchvision/prototype/models/detection/keypoint_rcnn.py
```diff
@@ -48,6 +48,14 @@ class KeypointRCNNResNet50FPNWeights(Weights):
...
 
 
+# This is BC breaking too, and unfortunately I'm not sure we can raise warnings
+# like in fasterrcnn_mobilenet_v3_large_fpn because we wouldn't be able
+# to differentiate num_keypoints from num_classes.
+#
+# I'm not sure what the best course of action could be here
+#
+# Not a solution, but related: these issues are a strong argument in favour of
+# forcing keyword-only parameters in new APIs.
```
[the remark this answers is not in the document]
fmassa: We will only be releasing the new dataset APIs once torchdata becomes beta / stable

torchvision/prototype/models/googlenet.py
fmassa: The parameter overwriting is legacy due to the fact of how the model was implemented / ported.
If I could start from scratch I would have just removed `transform_input` altogether and changed the pre-trained weights.

torchvision/prototype/models/densenet.py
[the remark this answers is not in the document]
datumbox: Addressed at commit b124ad40b58e959d8124d18e5a5f1c8ed4707d64 of PR #4948

torchvision/prototype/models/googlenet.py
datumbox: @fmassa Understood, we had to do what we had to do.

The question here is whether we will adopt everywhere the policy of throwing exceptions instead of silently overwriting  input parameters with weight specific config. If you recall you brought up this topic in an earlier PR, suggesting that instead of overwriting the `num_classes` we should raise an error if the numbers don't match with the pretrained config. So the question here is, should we do the same for other parameters such as `transform_input`, `aux_logits`, `init_weights` etc.

@NicolasHug proposed to treat this as a bug and thus fix it. Hence we wouldn't consider this BC-breaking. It wouldn't be the first time to do this, so the idea is not crazy.

I'm a bit concerned that this might have downstream effects to users. An alternative approach could be to throw warnings instead of errors. Thoughts?

torchvision/prototype/models/resnet.py
datumbox: @NicolasHug We could do a "poor man's implementation" of the `latest` feature by literally setting the value to the selected weight within the builder. This is an implementation detail invisible to the API and can be replaced with a fancy technique once we have the registration mechanism. I'm tempted to implement this.

NicolasHug: Yes, that sounds reasonable

torchvision/prototype/models/detection/faster_rcnn.py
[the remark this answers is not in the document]
datumbox: The default value of `num_classes` changed to `None` in a BC way on at commit f253f4c104bde286891dacdd0bbd7c87b9e74535 of PR #4948

torchvision/prototype/models/resnet.py
datumbox: The default value of the `num_classes` changed to `None`. So this is partially addressed at commit f253f4c104bde286891dacdd0bbd7c87b9e74535 of PR #4948

fmassa: Ok with me to let the default value of `num_classes` to be `None`

torchvision/prototype/models/googlenet.py
fmassa: I think silent overrides can be bad and we should avoid it.
```python
resnet50(pretrained=True, num_classes=10)
```
should raise a loud error (which is currently the case in the old code-path).

torchvision/prototype/models/resnet.py
datumbox: Addressed at PR #4948 by commit 1172e1fefce5b3975ab30a78ad3da1673f89942b

torchvision/prototype/models/googlenet.py
datumbox: Addressed at PR #4948 commit 9ff811ba6ae4af939aa8fb6f44b0e7b5eaef818a

torchvision/prototype/models/resnet.py
datumbox: The silent overwrites of `num_classes` and other similar config params is addressed at PR #4948 commit 9ff811ba6ae4af939aa8fb6f44b0e7b5eaef818a

torchvision/prototype/models/resnet.py
datumbox: Addressed at PR #4948 commit 6030338780decdf24e36b5045b960d05463e6766

torchvision/prototype/models/detection/faster_rcnn.py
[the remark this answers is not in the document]
datumbox: Addressed at PR #4948 commit 6030338780decdf24e36b5045b960d05463e6766

torchvision/prototype/models/resnet.py
datumbox: @NicolasHug @fmassa I'm just throwing this out there: 

An alternative way to implement this `latest weights` feature could be to maintain `pretrained=True` as a shortcut to get always the most updated weights. This USED to be the case since all models had only a single weight. There were even cases where old weights moved to `pretrained="legacy"` (see [here](https://github.com/pytorch/vision/blob/main/torchvision/models/detection/keypoint_rcnn.py#L379-L380)). So in a sense one could argue that the default behaviour of `pretrained=True` is currently to return the best available weights, hence it's not BC-breaking. Note that this approach is followed by other libraries where they overwrite directly the weights in-place.

Doing this will get us automatic adoption to the best/greatest weights while at the same time it will allow users who want to always access specific weights to be able to do so. Thoughts?

datumbox: The above comment will be discussed separately.

For the original request discussed here, I implemented a basic mechanism that allows us to define `weights="default"` at PR #4948 commit 8aa1c233dbf71790235840e8448a34ad43217d5c. I preferred `default` instead of `latest` because it describes better what it does and gives us the freedom to update the default weights for reasons other than recency. This is particularly useful if we have multiple weights on different datasets (for example Voc vs COCO) but we can have only 1 default.

datumbox:
@NicolasHug I believe we addressed all the comments on this PR and we can close it. If there are more, happy to make the needed changes. Let me know

NicolasHug:
Thanks for all the discussions @datumbox @fmassa , we can safely close this one now. Latest discussions were tracked in https://github.com/pytorch/vision/issues/4652#issuecomment-981587337

PR #1448, opened 2019-10-11
Added elastic transform in torchvision.transforms.

gokriznastic:
I've added elastic transformations support as requested in #888.

Issue #888, opened 2019-05-05
Ask for elastic transform support


Usongkam:
Elastic transform do well on MNIST and some other image datasets. 

However, it is not supported in torchvision.transforms , which make it difficult for me to compose an ideal transformation.

Could u please add this to torchvision.transforms in next verision ?

ekagra-ranjan:
@fmassa If this is to be added then can I take this up?

fmassa:
@ekagra-ranjan yes, this is on the plans to be added, a PR adding it would be awesome!

JLrumberger:
A PR adding elastic transform got merged today, so this issue and the other PR can be closed I guess.

PR #4938, opened 2021-11-15
Added elastic transform in torchvision.transforms

JLrumberger:
Hi all!

I've added elastic transformations as requested in #888 for torch tensor images, without adding new dependencies as in PR #1448 . I didn't add a PIL version, since it would also require a dependency to `scipy.ndimage.interpolation.map_coordinates` as a substitute for `torch.nn.functional.grid_sample` which seems to be not desired. However, I could add a PIL version that uses `F.to_tensor` internally, but that's something that users could also do when composing the augmentation pipeline. What do you think?

I'll be happy about your comments on this PR. It's my first PR, so if I did something wrong, please don't feel offended. 

Best
Lorenz


vfdev-5:
Hi @JLrumberger , thanks for the PR ! I wonder if you have any references in addition to https://github.com/funkey/augment explaining this elastic transform implementation.
As there is no single way to implement that, for example, 
- https://github.com/pytorch/vision/pull/1448 is based on https://gist.github.com/ernestum/601cdf56d2b424757de5
- [albumentation](https://albumentations.ai/docs/api_reference/augmentations/geometric/transforms/#albumentations.augmentations.geometric.transforms.ElasticTransform) is also based on the same gist 
- yet another implementation: https://github.com/IssamLaradji/elastic_transform/blob/master/elastic_transform.py
- monai elastic transform: https://docs.monai.io/en/latest/transforms.html#monai.transforms.Rand2DElastic

All these implementations have a bit different APIs and maybe even different behaviour. I'd like to find out the most practical API before moving forward on implementation details.

Thanks!

JLrumberger:
Hey vfdev-5,
thanks for the feedback. Yes, they seem to have different APIs and behaviors.

The following four implementations require `alpha` values that control the strength of the displacements and `sigma` values for gaussian smoothing of displacements. These implementations get random displacements in x and y direction for every pixel in the image, then smooth the displacements via a gaussian filter parameterized by `sigma` and upscale the strength of the random displacements by `alpha`.
- #1448
- [ernestum](https://gist.github.com/ernestum/601cdf56d2b424757de5)
-  [albumentations](https://albumentations.ai/docs/api_reference/augmentations/geometric/transforms/#albumentations.augmentations.geometric.transforms.ElasticTransform)
- [imgaug](https://imgaug.readthedocs.io/en/latest/_modules/imgaug/augmenters/geometric.html#ElasticTransformation)

This PR, the [MONAI](https://docs.monai.io/en/latest/transforms.html#monai.transforms.Rand2DElastic) and [gunpowder](https://github.com/funkey/gunpowder/blob/master/gunpowder/nodes/elastic_augment.py) (based on Jan Funkes implementation as well) versions of elastic augment set regular spaced `control_points` onto the image, displace them randomly by sampling from a zero-centered Gaussian parameterized by `sigma` and interpolate the displacement vectors for all other pixels between the control_points. 

The first option seems to be used more in libraries that do exclusively 2D augmentations natural images whereas the latter caters more towards 3D volumes and microscopy or medical images. The results look quite different, [here](https://imgur.com/a/GPan4Wa) is an image transformed with the code in this PR, [here ](https://imgur.com/a/YwKtrQB) is the same image transformed with ernestums code.

vfdev-5:
Thanks for details @JLrumberger ! 


> I'm not sure where to go from here: Shall we implement both options and add to torchvision or decide if one of the two options is superior?

Good question. Maybe we can produce both. I assume the option 1 (albu, imgaug, etc) is probably the most used vs option 2 (also taking into account projects popularity). So, we can implement option 1 as ElasticTransform and in another PR add option2 with another name that clearly states the difference between two... (e.g. MedElasticTransform or something else with a better name).
If we could find some paper or a bit more scientific reference on option 2 it would be helpful for the second PR.

JLrumberger:
Hi @vfdev-5,
I like your idea and I'll be looking for reference for the MedElasticTransform. In the meantime, I implemented an imgaug-like version of elastic_deformation.

torchvision/transforms/functional_tensor.py
```diff
@@ -985,3 +985,53 @@ def equalize(img: Tensor) -> Tensor:
...
+
+    if isinstance(alpha, float) or isinstance(alpha, int):
+        alpha = [
+            alpha,
+        ] * 2
+
+    if isinstance(sigma, float) or isinstance(sigma, int):
+        sigma = [
+            sigma,
+        ] * 2
```
vfdev-5, line 1009: I think this trick maybe wont pass with torch script... Let's write some tests.

vfdev-5 (commented):
Thanks for the update @JLrumberger !
I'll check the details a bit later, but some comment right now:
- we have to write some tests, especially to ensure that the op can be torch scripted and gives the same output as non-scripted version.
- support PIL input could be done similarly to GaussianBlur where we do PIL -> tensor -> op -> PIL

torchvision/transforms/transforms.py
```diff
@@ -2014,3 +2014,51 @@ def forward(self, img):
...
+    and sigma controls the smoothness of the displacements.
+    The displacements are added to an identity grid and the resulting grid is
+    used to grid_sample from the image.
+    Applications:
+        Randomly transforms the morphology of objects in images and produces a
+        see-through-water-like effect.
+    Args:
+        alpha: List [alpha H, alpha W] or (float), defines the magnitude of displacements.
+        sigma: List [sigma H, sigma W] or (float) defining the smoothness of displacements.
+        interpolation: str saying either NEAREST, BILINEAR or BICUBIC. Defining the interpolation used in grid_sample.
```
vfdev-5, line 2033: interpolation arg should be similar to this one https://github.com/pytorch/vision/blob/fc52e86fa34413d35713ce264255a9f00902b2e2/torchvision/transforms/transforms.py#L748-L751

torchvision/transforms/transforms.py
```diff
@@ -2014,3 +2014,51 @@ def forward(self, img):
...
+
+    def forward(self, tensor: Tensor) -> Tensor:
+        """
+        Args:
+            tensor (Tensor): Tensor image to be transformed.
+
+        Returns:
+            Tensor: Transformed image.
+        """
+        return elastic_deformation(tensor, self.control_point_spacing, self.sigma, self.interpolation, self.fill)
```
vfdev-5, line 2058: Let's make the code same as other transforms:
```python
def forward(self, img):
    return F.elastic_deformation(img, ...)
```

For example, https://github.com/pytorch/vision/blob/fc52e86fa34413d35713ce264255a9f00902b2e2/torchvision/transforms/transforms.py#L1492-L1510

torchvision/transforms/functional_tensor.py
```diff
@@ -985,3 +985,53 @@ def equalize(img: Tensor) -> Tensor:
...
+    )
+    dy = (
+        gaussian_blur(
+            torch.rand(*shape).unsqueeze(0).unsqueeze(0).to(img.device) * 2 - 1,
+            [4 * int(sigma[0]) + 1, 4 * int(sigma[1]) + 1],
+            sigma,
+        )
+        * alpha[1]
+        / shape[1]
+    )
```
vfdev-5, line 1030: Let's also handle sigma = 0 case.

vfdev-5 (commented):
@JLrumberger I left few other comments.

I also quickly checked on a synth data the result vs imgaug and albu and obtained 3 different results. Let me think if we need to align with an implementation...

vfdev-5:
@JLrumberger any updates on this PR ?

JLrumberger:
I'm currently busy with other stuff and will work on it by the end of this week.

JLrumberger:
I added some minor changes and still need to write the test cases.

torchvision/transforms/transforms.py
```diff
@@ -6,6 +6,7 @@
 from typing import Tuple, List, Optional
 
 import torch
+from numpy import not_equal
```
vfdev-5, line 9: Do we really need to use numpy. It would be better to avoid adding new dependency

torchvision/transforms/transforms.py
```diff
@@ -2014,3 +2015,79 @@ def forward(self, img):
 
     def __repr__(self):
         return self.__class__.__name__ + f"(p={self.p})"
+
+
+class ElasticTransform(torch.nn.Module):
```
vfdev-5, line 2020: Let's also add `ElasticTransform` to `__all__`.

torchvision/transforms/transforms.py
```diff
@@ -2014,3 +2015,79 @@ def forward(self, img):
...
+    """Transform a tensor image with elastic transformations.
+    Given alpha and sigma, it will generate displacement
+    vectors for all pixels based on random offsets. Alpha controls the strength
+    and sigma controls the smoothness of the displacements.
+    The displacements are added to an identity grid and the resulting grid is
+    used to grid_sample from the image.
+    Applications:
```
vfdev-5, line 2027:
```suggestion

    Applications:
```

torchvision/transforms/transforms.py
```diff
@@ -2014,3 +2015,79 @@ def forward(self, img):
...
+        Randomly transforms the morphology of objects in images and produces a
+        see-through-water-like effect.
+    Args:
```
vfdev-5, line 2030:
```suggestion

    Args:
```

torchvision/transforms/transforms.py
```diff
@@ -2014,3 +2015,79 @@ def forward(self, img):
...
+        alpha: List [alpha H, alpha W] or (float), defines the magnitude of displacements.
```
vfdev-5, line 2031:
```suggestion
        alpha (float or sequence of floats): Magnitude of displacements.
```

torchvision/transforms/transforms.py
```diff
@@ -2014,3 +2015,79 @@ def forward(self, img):
...
+        sigma: List [sigma H, sigma W] or (float) defining the smoothness of displacements.
```
vfdev-5, line 2032:
```suggestion
        sigma (float or sequence of floats): Smoothness of displacements.
```

vfdev-5 (commented):
Thanks for the updates @JLrumberger 
I'll push to your branch some code updates to make better tests etc.

torchvision/transforms/transforms.py
JLrumberger: I think we don't need it. I do not use it in my code and I'm not sure why this is in here.

torchvision/transforms/functional.py
```diff
@@ -1390,3 +1390,70 @@ def equalize(img: Tensor) -> Tensor:
...
     return F_t.equalize(img)
+
+
+def elastic_transform(
+    img: Tensor,
+    alpha: List[float],
+    sigma: List[float],
+    interpolation: InterpolationMode = InterpolationMode.BILINEAR,
+    fill: Optional[List[float]] = None,
+) -> Tensor:
```
vfdev-5, line 1401: @JLrumberger I wonder if functional op shouldn't be derandomized, it can accept either
1) displacement as [1, H, W, 2] tensor
or
2) same signature as here and we can pass random generator as arg as well (similarly to https://github.com/albumentations-team/albumentations/blob/37e714fd2e326f6f88778e425f98c2de8c8d5372/albumentations/augmentations/geometric/transforms.py#L178)

JLrumberger: @vfdev-5 this would make it possible to better test the behavior of the op and slightly extend it's functionality. Sounds reasonable to me.

JLrumberger: I added the requested changes.

vfdev-5: Thanks for the update, I'll check in details asap. 
Looking at failing tests we may have a limitation from torch jit script side when passing random state. Maybe it is impossible to pass it and thus we have to do something else...
https://app.circleci.com/pipelines/github/pytorch/vision/13819/workflows/452c2c97-199b-47c8-9294-577f77c14583/jobs/1112516

torchvision/transforms/functional.py
```diff
@@ -1461,3 +1461,83 @@ def equalize(img: Tensor) -> Tensor:
...
     return F_t.equalize(img)
+
+
+def elastic_transform(
+    img: Tensor,
+    alpha: List[float],
+    sigma: List[float],
+    interpolation: InterpolationMode = InterpolationMode.BILINEAR,
+    fill: Optional[List[float]] = None,
+    displacement: Optional[Tensor] = None,
```
vfdev-5, line 1472: In terms of API, there are two things I'm a bit unhappy:
- if user provides `displacement` and `alpha` and `sigma` then only `displacement` is taken into account =>
either we have make both optional and warn/raise an error that only one is possible.
- if we are using PIL image as input, how user should define a displacement. Mixing PIL image as input and `displacement` as tensor does not look as a proper solution...

JLrumberger: 1) sounds reasonable to do it that way.
2) okay, so we also accept np.array as the displacement type and wrap it with torch.tensor internally?

vfdev-5: > okay, so we also accept np.array as the displacement type and wrap it with torch.tensor internally?

Yes, that's can be an option

JLrumberger: Okay, I implemented both changes and pushed it.

torchvision/transforms/functional.py
```diff
@@ -1528,7 +1528,7 @@ def elastic_transform(
 
         t_img = to_tensor(img)
     if not isinstance(displacement, torch.Tensor):
-        displacement = torch.from_numpy(displacement)
+        displacement = torch.Tensor(displacement)
```
vfdev-5, line 1531: @JLrumberger this change breaks the CI and we shouldn't use `torch.Tensor` to create tensors.
Why you replaced `torch.from_numpy` ?

JLrumberger: @vfdev-5 I thought that `torch.from_numpy`  breaks CI and therefore I replaced it by `torch.Tensor` but I was wrong. Sorry.

torchvision/transforms/functional.py
vfdev-5: > warn/raise an error that only one is possible.

@JLrumberger we need explicitly check that user does not provide both alph/sigma and displacement. if displacement is given, both alpha and sigma should be None, otherwise raise an error.


As for displacement, JIT can not work with numpy input => if input is tensor, then displacement should be a tensor, if input is PIL.Image, displacement should be a ndarray.

torchvision/transforms/functional_tensor.py
```diff
@@ -993,3 +993,52 @@ def equalize(img: Tensor) -> Tensor:
...
+        torch.manual_seed(random_state)
+
+    if displacement is not None:
+        displacement = displacement.to(img.device)
+    elif alpha is not None and sigma is not None:
+        dx = torch.rand([1, 1] + size, device=img.device) * 2 - 1
+        if sigma[0] > 0.0:
+            dx = gaussian_blur(
+                dx,
+                [4 * int(sigma[0]) + 1, 4 * int(sigma[1]) + 1],
```
vfdev-5, line 1022: @JLrumberger do you have any refs for this choice of kernel size ?
If we take https://gist.github.com/ernestum/601cdf56d2b424757de5#file-elastic_transform-py-L16-L17 as ref, then
it would be something like `~int(8 * sigma + 1)`:
```
sd = float(sigma)
lw = int(truncate * sd + 0.5)
x = numpy.arange(-radius, radius+1)  # radius = lw
```
- https://github.com/scipy/scipy/blob/47bb6febaa10658c72962b9615d5d5aa2513fa3a/scipy/ndimage/filters.py#L258

- https://github.com/scipy/scipy/blob/47bb6febaa10658c72962b9615d5d5aa2513fa3a/scipy/ndimage/filters.py#L187

JLrumberger: @vfdev-5 no, I don't have a reference for this, but I have a few comments:
1. I think ernestums implementation and ours are not really comparable when it comes to the parameters, since ernestum uses `scipy.ndimage.interpolation.map_coordinates`  which expects as an identity sampling grid this `np.meshgrid(np.arange(shape[0]), np.arange(shape[1]), np.arange(shape[2]))`, so values lie between 0 and H or W.  In our implementation we use `vision.torchvision.transforms.functional_tensor._apply_grid_transform` which expects as an identity grid this `[torch.linspace((-s + 1) / s, (s - 1) / s, s) for s in size]`, so values lie between -1:1 for any given image independant of the shape. Anyways, I think the smoothing shouldn't be affected by this
2. I looked into imgaugs Elastic Transformation and they calculate the kernel_size based on the sigma like this:
`def _compute_gaussian_blur_ksize(sigma):
    if sigma < 3.0:
        ksize = 3.3 * sigma  # 99% of weight
    elif sigma < 5.0:
        ksize = 2.9 * sigma  # 97% of weight
    else:
        ksize = 2.6 * sigma  # 95% of weight

    # we use 5x5 here as the minimum size as that simplifies
    # comparisons with gaussian_filter() in the tests
    # TODO reduce this to 3x3
    ksize = int(max(ksize, 5))
    return ksize` 
I don't think that this is necessary for our implementation.
3. I ran our implementation as is `elastic_transform(img,alpha=[40,40],sigma=[4,4], random_state=0)` and with the kernel size changed to `int(8 * sigma + 1)` and stored the results in the red channel of the as is implementation in the blue channel and the one with the bigger kernel size in the red channel and the results look identical:
red_8_blue_4
 
 So all in all, I think we can change it to `int(8 * sigma + 1)` for consistency with ernestums implementation. Sorry for the late reply and thanks for all your work :)

vfdev-5: Thanks for updates @JLrumberger !
Major difference between `int(8 * sigma + 1)` vs `4 * int(sigma) + 1` is when using sigma < 1. For all sigmas < 1 we'll have kernel size as 1 in the second case. 
For sigma = 4, `int(8 * sigma + 1)=33` vs `4 * int(sigma) + 1 = 17`.

JLrumberger:
@vfdev-5 I added the changes you requested some months ago. Could you once again look through it and give me roadmap, mentioning what needs to be done in order to get this finally merged?

datumbox:
@Lrumberger Your last commits seem to have broken the PR. Github currently reports the modification of over 150 files, which I'm sure is not intentional. I recommend revert this commit and force pushing, then merging master to it to get it to a clean state.

vfdev-5:
> @vfdev-5 I added the changes you requested some months ago. Could you once again look through it and give me roadmap, mentioning what needs to be done in order to get this finally merged?

@JLrumberger thanks for pinging and moving this forward. Sorry for delay on this PR, I somehow switched to other things without keeping this one in my list.
I'm reviewing it again and think about `random_state` arg in functional part which is rather unusual. So, maybe we have to restructure the API once again a bit and it could landed later

vfdev-5:
@JLrumberger I updated your PR by removing random_state from functional part and fixing tests. Right now we have to ensure that the transformation does what we expect.
Can you check from your side about default parametrization (alpha, sigma) we can put into the docs such that transformed image is meaningful.
I'm checking it as following does not give much visual differences:
```python
from torchvision.io import read_image

t = read_image("sample_dog.jpg")
# t.shape -> torch.Size([3, 323, 500])

displacement = get_params([5.0, 5.0], [0.12, 0.12], [323, 500])
tt = elastic_transform(t, displacement, interpolation="nearest")

plt.figure(figsize=(15, 7))
plt.subplot(121)
plt.title("Original")
plt.imshow(t.permute(1, 2, 0).numpy())
plt.subplot(122)
plt.title("Elastic")
plt.imshow(tt.permute(1, 2, 0).numpy())
```

image

vfdev-5:
@JLrumberger could you please update our default value to those and we can be good to go with this PR

gallery/plot_video_api.py
```diff
@@ -325,7 +325,7 @@ def __iter__(self):
 # ----------------------------------
 # Example of visualized video
 
-import matplotlib.pylab as plt
+import matplotlib.pyplot as plt
```
vfdev-5, line 328: pylab is deprecated : https://matplotlib.org/stable/api/index.html?highlight=pylab#module-pylab

torchvision/transforms/functional.py
```diff
@@ -1484,3 +1484,65 @@ def equalize(img: Tensor) -> Tensor:
         return F_pil.equalize(img)
 
     return F_t.equalize(img)
+
+
+def elastic_transform(
```
datumbox, line 1489: Should this be added to the `__init__.py` or not?

datumbox (commented):
@JLrumberger Thanks a lot for the work. We work with Victor to get this merged ASAP.

@vfdev-5 Thanks for the changes. I've added few minor comments below. Let me know your thoughts.

torchvision/transforms/functional.py
vfdev-5: No, it is ok like that :)

datumbox (approved):
LGTM, thanks!

@vfdev-5 feel free to merge on green CI.

PR #4939, opened 2021-11-16
Updates PIL version specifier to prevent issues with padding when image mode = P

frgfm:
As discussed in #4934, this PR updates the version specifier of PIL to avoid incompatibility of the padding function for specific versions.

Closes #4934 

Any feedback is welcome!

cc @oke-aditya

Issue #4934, opened 2021-11-14
[transforms] Troubles with padding for PIL image in mode P


frgfm:
### 🐛 Describe the bug

Hello there :wave: 

I was running some training for semantic segmentation using the implementation of `RandomCrop` provided in the [references](https://github.com/pytorch/vision/blob/main/references/segmentation/transforms.py#L9) folder. And apparently there is an issue with the padding function for PIL images. Please note that this was not happening a few months back.

The following snippet
```python
from PIL import Image
from torchvision.transforms.functional import pad

img = Image.open('/path/to/any/img.jpg').convert('P')
padded = pad(img, (0, 0, 5, 10), fill=255)
```
produces
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-14-e151dc3a245c> in <module>
----> 1 padded = pad(img, (0, 0, 5, 10), fill=255)

~/miniconda3/lib/python3.8/site-packages/torchvision/transforms/functional.py in pad(img, padding, fill, padding_mode)
    470     """
    471     if not isinstance(img, torch.Tensor):
--> 472         return F_pil.pad(img, padding=padding, fill=fill, padding_mode=padding_mode)
    473 
    474     return F_t.pad(img, padding=padding, fill=fill, padding_mode=padding_mode)

~/miniconda3/lib/python3.8/site-packages/torchvision/transforms/functional_pil.py in pad(img, padding, fill, padding_mode)
    162         if img.mode == "P":
    163             palette = img.getpalette()
--> 164             image = ImageOps.expand(img, border=padding, **opts)
    165             image.putpalette(palette)
    166             return image

~/miniconda3/lib/python3.8/site-packages/PIL/ImageOps.py in expand(image, border, fill)
    401 
    402         draw = ImageDraw.Draw(out)
--> 403         draw.rectangle((0, 0, width - 1, height - 1), outline=color, width=border)
    404     else:
    405         out = Image.new(image.mode, (width, height), color)

~/miniconda3/lib/python3.8/site-packages/PIL/ImageDraw.py in rectangle(self, xy, fill, outline, width)
    257             self.draw.draw_rectangle(xy, fill, 1)
    258         if ink is not None and ink != fill and width != 0:
--> 259             self.draw.draw_rectangle(xy, ink, 0, width)
    260 
    261     def rounded_rectangle(self, xy, radius=0, fill=None, outline=None, width=1):

TypeError: an integer is required (got type tuple)
```

[...]

PR #4946, opened 2021-11-16
chore: Updated all pillow version specifiers

frgfm:
Following up on #4939, this PR updates the remaining pillow version specifiers. I used a repo-wide search for occurrences of pillow to avoid missing any other.

As suggested in https://github.com/pytorch/vision/pull/4939#issuecomment-970486674, I suggest we implement a dependency specifier consistency check.

Any feedback is welcome!
cf #4934

cc @datumbox @pmeier @prabhat00155


pmeier (approved):
`grep`'ing didn't turn up any misses. Thanks a lot @frgfm!

prabhat00155 (approved):
Thanks @frgfm!

PR #4950, opened 2021-11-17
make prototype datasets traversable

pmeier:
Turns out, apart from `lambda`'s, local functions are also not supported by datapipes. @ejguan Maybe also detect local functions and warn about them together with the `lambda` warning?

cc @pmeier @bjuncek


test/test_prototype_builtin_datasets.py
```diff
@@ -76,6 +77,10 @@ def test_decoding(self, dataset, mock_info):
                 f"{sequence_to_str(sorted(undecoded_features), separate_last='and ')} were not decoded."
             )
 
+    @builtin_datasets
+    def test_traversable(self, dataset, mock_info):
+        traverse(dataset)
```
ejguan, line 82: For traversing without need of serialization, should we at least use `exclude_primitive` in DataLoader for now?
cc: @VitalyFedyunin

pmeier: This test should just make sure that we are later able to set the `Shuffler` from the data loader. And for that we need to travese the graph, right? If there is a better way to test that, I'm open for suggestions.

fmassa (approved):

PR #4952, opened 2021-11-17
Remove non-functional Transforms from presets

datumbox:
Fixes partially #4652

cc @datumbox @vfdev-5 @bjuncek

### DIFF
--- a/torchvision/prototype/models/video/resnet.py
+++ b/torchvision/prototype/models/video/resnet.py
@@ -62,7 +62,7 @@ _COMMON_META = {
 class R3D_18Weights(Weights):
     Kinetics400_RefV1 = WeightEntry(
         url="https://download.pytorch.org/models/r3d_18-b3b3357e.pth",
-        transforms=partial(Kinect400Eval, resize_size=(128, 171), crop_size=(112, 112)),
+        transforms=partial(Kinect400Eval, crop_size=(112, 112), resize_size=(128, 171)),
         meta={
             **_COMMON_META,
             "acc@1": 52.75,
@@ -74,7 +74,7 @@ class R3D_18Weights(Weights):
 class MC3_18Weights(Weights):
     Kinetics400_RefV1 = WeightEntry(
         url="https://download.pytorch.org/models/mc3_18-a90a0ba3.pth",
-        transforms=partial(Kinect400Eval, resize_size=(128, 171), crop_size=(112, 112)),
+        transforms=partial(Kinect400Eval, crop_size=(112, 112), resize_size=(128, 171)),
         meta={
             **_COMMON_META,
             "acc@1": 53.90,
@@ -86,7 +86,7 @@ class MC3_18Weights(Weights):
 class R2Plus1D_18Weights(Weights):
     Kinetics400_RefV1 = WeightEntry(
         url="https://download.pytorch.org/models/r2plus1d_18-91a641e6.pth",
-        transforms=partial(Kinect400Eval, resize_size=(128, 171), crop_size=(112, 112)),
+        transforms=partial(Kinect400Eval, crop_size=(112, 112), resize_size=(128, 171)),
         meta={
             **_COMMON_META,
             "acc@1": 57.50,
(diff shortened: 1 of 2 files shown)


torchvision/prototype/models/video/resnet.py
```diff
@@ -62,7 +62,7 @@ def _video_resnet(
 class R3D_18Weights(Weights):
     Kinetics400_RefV1 = WeightEntry(
         url="https://download.pytorch.org/models/r3d_18-b3b3357e.pth",
-        transforms=partial(Kinect400Eval, resize_size=(128, 171), crop_size=(112, 112)),
+        transforms=partial(Kinect400Eval, crop_size=(112, 112), resize_size=(128, 171)),
```
datumbox, line 65: Changing order to match the other presets.

torchvision/prototype/transforms/_presets.py
```diff
@@ -26,42 +25,47 @@ def __init__(
...
     def forward(self, img: Tensor) -> Tensor:
-        img = self._crop(self._resize(img))
+        img = F.resize(img, self._size, interpolation=self._interpolation)
+        img = F.center_crop(img, self._crop_size)
         if not isinstance(img, Tensor):
             img = F.pil_to_tensor(img)
         img = F.convert_image_dtype(img, torch.float)
-        return self._normalize(img)
+        img = F.normalize(img, mean=self._mean, std=self._std)
+        return img
```
datumbox, line 44: replacing with the functional equivalents.

torchvision/prototype/transforms/_presets.py
```diff
@@ -26,42 +25,47 @@ def __init__(
...
     def forward(self, vid: Tensor) -> Tensor:
         vid = vid.permute(0, 3, 1, 2)  # (T, H, W, C) => (T, C, H, W)
-        vid = self._convert(vid)
-        vid = self._resize(vid)
-        vid = self._normalize(vid)
-        vid = self._crop(vid)
+        vid = F.resize(vid, self._size, interpolation=self._interpolation)
+        vid = F.center_crop(vid, self._crop_size)
+        vid = F.convert_image_dtype(vid, torch.float)
+        vid = F.normalize(vid, mean=self._mean, std=self._std)
```
datumbox, line 68: I reordered the operations. This is more "usual/canonical" order of ops. I'm running tests to confirm the accuracy remains the same.

datumbox (commented):
Few clarifications to assist review:

torchvision/prototype/transforms/_presets.py
datumbox: The accuracy before and after the change remains the same. The above change is safe.

torchvision/prototype/models/video/resnet.py
pmeier: This is just for appearance, right?

torchvision/prototype/transforms/_presets.py
```diff
@@ -26,42 +25,47 @@ def __init__(
...
+        img = F.normalize(img, mean=self._mean, std=self._std)
+        return img
 
 
 class Kinect400Eval(nn.Module):
     def __init__(
         self,
-        resize_size: Tuple[int, int],
         crop_size: Tuple[int, int],
+        resize_size: Tuple[int, int],
```
pmeier, line 51: Unrelated to this PR, but why do we use `Tuple[int, int]` here for the sizes, but plain `int`'s for `ImageNetEval`?

pmeier (approved):
Two questions inline, otherwise LGTM. Thanks Vasilis!

torchvision/prototype/transforms/_presets.py
datumbox: It's equivalent. Confirmed by checking the accuracy of models before and after.

torchvision/prototype/models/video/resnet.py
datumbox: yes, just for styling. :)

torchvision/prototype/transforms/_presets.py
datumbox: This is a good point. This is how the original recipes have been implemented, so I need to do this here too. If you see the resize operates differently if you specify both dimensions vs only 1. Prior merging we should consider adding unions etc here and cleaning up further.

PR #4960, opened 2021-11-18
Improve prototype dataset benchmarks

pmeier:
This adds two feature sets:

1. Instead of testing the full iteration multiple times, we now test for a number of samples. If the number of samples is "large enough" say `10_000` that should also give an accurate reading. The old behavior is not suitable to test large datasets like ImageNet.
2. All parts of the benchmarks are now skippable through flags to avoid running everything all the time.

cc @pmeier @bjuncek


fmassa (approved):
LGTM, although I would maybe add the `.take` to the dataset. But can be done in a follow-up PR if needed

Issue #4917, opened 2021-11-11
torchvision conda package missing requests dependency


malfet:
### 🐛 Describe the bug

torchvision depends on requests, but it is not specified as conda package dependency, which results in the following failures:
```
% python
Python 3.8.11 (default, Aug  6 2021, 08:56:27) 
[Clang 10.0.0 ] :: Anaconda, Inc. on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from torchvision import datasets
>>> calldetch_ds = datasets.Caltech101("datasets", download=True)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/nshulga/miniconda3/envs/py_38/lib/python3.8/site-packages/torchvision/datasets/caltech.py", line 52, in __init__
    self.download()
  File "/Users/nshulga/miniconda3/envs/py_38/lib/python3.8/site-packages/torchvision/datasets/caltech.py", line 124, in download
    download_and_extract_archive(
  File "/Users/nshulga/miniconda3/envs/py_38/lib/python3.8/site-packages/torchvision/datasets/utils.py", line 427, in download_and_extract_archive
    download_url(url, download_root, filename, md5)
  File "/Users/nshulga/miniconda3/envs/py_38/lib/python3.8/site-packages/torchvision/datasets/utils.py", line 135, in download_url
    return download_file_from_google_drive(file_id, root, filename, md5)
  File "/Users/nshulga/miniconda3/envs/py_38/lib/python3.8/site-packages/torchvision/datasets/utils.py", line 204, in download_file_from_google_drive
    import requests
ModuleNotFoundError: No module named 'requests'
```

### Versions

0.11.1 on MacOS

datumbox:
I believe the dependency is on purpose not in the mandatory list. It's considered optional (see old issue at #2618).

The only use of the `requests` is at:
https://github.com/pytorch/vision/blob/d367a01a18a3ae6bee13d8be3b63fd6a581ea46f/torchvision/datasets/utils.py#L203-L217

malfet:
@datumbox in that case it should be add to extra_deps, same as scipy, shouldn't it?

datumbox:
@malfet I've had a quick look to see what other optional dependencies we are using. Here is a list: 
- IO:
  - `accimage`
  - `av`
- Datasets:
  - `torchdata` (Prototype-only)
  - `scipy`
  - `pycocotools`
  - `lmdb`
  - `requests`

[...]

PR #4962, opened 2021-11-19
add third party dependencies to prototype datasets

pmeier:
Addresses one aspect of the discussion in #4917 for the prototype datasets. @datumbox With this patch, we can easily extract all third party dependencies with

```python
>>> from torchvision.prototype import datasets
>>> sorted({dependency for name in datasets.list() for dependency in datasets.info(name).dependecies})
['scipy']
```

If we go for the `"datasets"` extra requirements proposed in https://github.com/pytorch/vision/issues/4917#issuecomment-970174681, it is easy to write a test that ensures parity between the datasets and the entry in `setup.py`.

cc @pmeier @bjuncek


datumbox (approved):
LGTM, thanks @pmeier.

PR #4963, opened 2021-11-19
don't close file handles in protoype dataset tests

pmeier:
I thought that would be a good idea when I added it in #4863. Turns out, this only works under the assumption that every file is present in exactly one sample. Optical flow datasets like Sintel violate that assumption, which leads to [test failures](https://app.circleci.com/pipelines/github/pytorch/vision/12723/workflows/553b8f97-51fd-4534-b1ee-33f8df0af2e8/jobs/1009461) although the dataset works fine.

`torchdata` has some internal functionality to close files when a datapipe is exhausted. Better to just rely on that to avoid issues like this in the future.

cc @pmeier @bjuncek

### DIFF
--- a/test/test_prototype_builtin_datasets.py
+++ b/test/test_prototype_builtin_datasets.py
@@ -10,10 +10,7 @@ from torchvision.prototype.utils._internal import sequence_to_str
 
 
 def to_bytes(file):
-    try:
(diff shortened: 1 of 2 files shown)


fmassa (approved):
LGTM!

PR #4967, opened 2021-11-20
Add multi-type support on `get_weight()`

datumbox:
Extend the `get_weight()` to handle properly model builders with multiple acceptable weight enums. This is useful for quantization.

cc @datumbox @bjuncek

### DIFF
--- a/references/classification/train.py
+++ b/references/classification/train.py
@@ -30,12 +30,12 @@ def train_one_epoch(model, criterion, optimizer, data_loader, device, epoch, arg
     for i, (image, target) in enumerate(metric_logger.log_every(data_loader, args.print_freq, header)):
         start_time = time.time()
         image, target = image.to(device), target.to(device)
-        with torch.cuda.amp.autocast(enabled=args.amp):
+        with torch.cuda.amp.autocast(enabled=scaler is not None):
             output = model(image)
             loss = criterion(output, target)
 
         optimizer.zero_grad()
-        if args.amp:
+        if scaler is not None:
             scaler.scale(loss).backward()
             if args.clip_grad_norm is not None:
                 # we should unscale the gradients of optimizer's assigned params if do gradient clipping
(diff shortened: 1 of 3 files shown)


references/classification/train.py
```diff
@@ -30,12 +30,12 @@ def train_one_epoch(model, criterion, optimizer, data_loader, device, epoch, arg
     for i, (image, target) in enumerate(metric_logger.log_every(data_loader, args.print_freq, header)):
         start_time = time.time()
         image, target = image.to(device), target.to(device)
-        with torch.cuda.amp.autocast(enabled=args.amp):
+        with torch.cuda.amp.autocast(enabled=scaler is not None):
```
prabhat00155, line 33: Why this change? scaler is not None in case args.amp is true, but having args.amp here makes the code more readable.

torchvision/prototype/models/_api.py
```diff
@@ -101,6 +101,11 @@ def get_weight(fn: Callable, weight_name: str) -> Weights:
         # TODO: Replace ann.__args__ with typing.get_args(ann) after python >= 3.8
         for t in ann.__args__:  # type: ignore[union-attr]
             if isinstance(t, type) and issubclass(t, Weights):
+                # ensure the name exists. handles builders with multiple types of weights like in quantization
+                try:
+                    t.from_str(weight_name)
+                except ValueError:
+                    continue
```
prabhat00155, line 108: Shouldn't we at least print a warning related to the exception rather than just ignoring it?

prabhat00155 (commented):
Thanks @datumbox! I have a couple of comments.

references/classification/train.py
datumbox: This is only because in the quantized training script the method `train_one_epoch` is reused. Since there the `amp` param is not supported, accessing `args.amp` breaks the code. I think a more thorough refactoring is needed here but at least this fixes the broken code.

torchvision/prototype/models/_api.py
datumbox: Ignoring is probably the right thing to do here. We try to match the provided string of the user with any known weight enums and if all fail, on the next couple of lines we throw an exception:
https://github.com/pytorch/vision/blob/015ee04a54908fc3ba9df65587d682f3d56268cf/torchvision/prototype/models/_api.py#L112-L115

Throwing a warning for every enum trying to match will only cause confusion.

prabhat00155 (approved):

PR #4971, opened 2021-11-22
remove custom code for model output comparison

pmeier:
After pytorch/pytorch#67795 was merged, we can now also safely check `None` vs. `None`. @NicolasHug IIRC, this was the only place this workaround was needed, correct?

cc @datumbox @pmeier


NicolasHug (approved):
Thanks @pmeier , LGTM when green!

> @NicolasHug IIRC, this was the only place this workaround was needed, correct?

I think so, yes

PR #4973, opened 2021-11-22
replace requests with urllib

### LINKED ISSUE #4941: Replace `requests` by `urllib`
### 🚀 The feature

See title.

### Motivation, pitch

The `requests` library is currently an optional dependency. It is used only for downloading from GDrive:

https://github.com/pytorch/vision/blob/30f4d108319b0cd28ae5662947e300aad98c32e9/torchvision/datasets/utils.py#L193-L203

It would be nice to refactor this to use `urllib` and drop a "hidden" dependency. From skimming the the linked StackOverflow issue, it seems our implementation is almost a 1-to-1 copy. Thus, I'd wager a guess that `requests` was simply chosen, because the solution was already there and not because the functionality is not available in `urllib`.

### Alternatives

_No response_

### Additional context

See #4917.

cc @pmeier

pmeier:
Closes #4941. With this we lose the ability to send the confirm token, since `urllib` has no support for it. I've checked all of our GDrive resources in the datasets and none required it. TBH, I don't think we ever had a case where it was needed.

In case the confirm token is indeed needed in the future, I suggest we hard depend on `requests`, since `urllib` is only meant for very simple uses cases.

cc @pmeier

### DIFF
--- a/torchvision/datasets/utils.py
+++ b/torchvision/datasets/utils.py
@@ -183,7 +183,7 @@ def list_files(root: str, suffix: str, prefix: bool = False) -> List[str]:
     return files
 
 
-def _quota_exceeded(first_chunk: bytes) -> bool:  # type: ignore[name-defined]
+def _quota_exceeded(first_chunk: bytes) -> bool:
     try:
         return "Google Drive - Quota exceeded" in first_chunk.decode()
     except UnicodeDecodeError:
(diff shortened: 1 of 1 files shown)


pmeier:
While working on #5035, I realized we unfortunately sometimes need the confirm token. This happens if GDrive is unable to perform a virus scan and is asking the user for confirmation to proceed. This regularly happens for large files. Thus, I would follow my suggestion in the top comment to revert this PR and afterwards make `requests` a hard dependency.

datumbox:
@pmeier There were a couple of PRs that removed the functionality (including cleaning up a private method), so clicking the revert button here wont do the trick. You might need to do a bit more to undo the change. Concerning adding `requests` as a hard dependency, I'm OK with it.

PR #4976, opened 2021-11-22
use decorator to log API usage

### LINKED ISSUE #4957: What's purpose of logging?
### 🐛 Describe the bug

What is the purpose of logging? Is it going to auto-upload usage statistics somewhere?

Regarding implementation, what is the reason for not just doing it in a separate file or at least with a decorator? Currently it pollutes the library code during code browsing, especially for simple small functions. It also hurts copy-pasteability, as one needs to delete all this logging unnecessary for small copied snippets.

### Versions

It would also be nice to make this field not required. Often, there is no specific version. Core PyTorch also doesn't force entering version.

kazhang:
fix #4957

### DIFF
--- a/torchvision/utils.py
+++ b/torchvision/utils.py
@@ -1,7 +1,8 @@
 import math
 import pathlib
 import warnings
-from typing import Union, Optional, List, Tuple, BinaryIO, no_type_check
+from functools import wraps
+from typing import Union, Optional, List, Tuple, BinaryIO
 
 import numpy as np
 import torch
@@ -375,13 +376,16 @@ def _generate_color_palette(num_masks: int):
     return [tuple((i * palette) % 255) for i in range(num_masks)]
 
 
-@no_type_check
-def _log_api_usage_once(obj: str) -> None:  # type: ignore
-    if torch.jit.is_scripting() or torch.jit.is_tracing():
-        return
-    # NOTE: obj can be an object as well, but mocking it here to be
-    # only a string to appease torchscript
-    if isinstance(obj, str):
-        torch._C._log_api_usage_once(obj)
-    else:
-        torch._C._log_api_usage_once(f"{obj.__module__}.{obj.__class__.__name__}")
+def _log_api_usage_once(f):
+    @wraps(f)
+    def wrapper(*args, **kwargs):
+        event = f.__module__
+        if f.__name__.endswith("__init__"):
+            # inside module instantiation
+            event += args[0].__class__.__name__
+        else:
+            event += f.__name__
+        torch._C._log_api_usage_once(event)
+        return f(*args, **kwargs)
+
+    return wrapper
--- a/torchvision/datasets/vision.py
+++ b/torchvision/datasets/vision.py
@@ -28,6 +28,7 @@ class VisionDataset(data.Dataset):
 
     _repr_indent = 4
 
+    @_log_api_usage_once
     def __init__(
         self,
         root: str,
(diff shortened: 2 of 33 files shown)


kazhang:
Unfortunately, torchscript doesn't like decorator. When using the decorator on boxes functions, JIT looks for helper functions in `utils.py`
@fmassa @datumbox do you have any workarounds?

torchvision/utils.py
```diff
@@ -375,13 +376,16 @@ def _generate_color_palette(num_masks: int):
...
-        torch._C._log_api_usage_once(obj)
-    else:
-        torch._C._log_api_usage_once(f"{obj.__module__}.{obj.__class__.__name__}")
+def _log_api_usage_once(f):
+    @wraps(f)
+    def wrapper(*args, **kwargs):
+        event = f.__module__
+        if f.__name__.endswith("__init__"):
+            # inside module instantiation
+            event += args[0].__class__.__name__
```
NicolasHug, line 385: I think we'll need a dot in the middle to preserve the previous naming (same below)

```suggestion
            event += "." + args[0].__class__.__name__
```

kazhang: Good catch, I addressed in local, but haven't pushed here yet.

kazhang:
> @kazhang I tried to reproduce the problem locally but I can't. Aka pulling your branch and running the tests doesn't cause issues to me. I'm using pytorch nightly `1.11.0.dev20211117`. Have you tried it locally and fails? It could be good to investigate and eliminate that it is caused by an upstream issue that was recently introduced.

@datumbox yes, I tested locally with Python
```
from torchvision.ops.boxes import box_area,box_iou
import torch
s = torch.jit.script(box_iou)
```
And saw the same failure as in unit test.

datumbox:
I can see now why the tests passed locally. We need to pass the `PYTORCH_TEST_WITH_SLOW=1` var to tests to be executed with JIT support. 

I don't believe decorators will be supported by JIT. I believe what happens is the `functools.wraps(f)` modifies information about the method and confuses JIT. If you remove it you will see it failing with another error:
```python
from torchvision.ops.boxes import box_area,box_iou
... import torch
... s = torch.jit.script(box_iou)
  File "<input>", line 3, in <module>
  File "./torch/jit/_script.py", line 1308, in script
    ast = get_jit_def(obj, obj.__name__)
  File "./torch/jit/frontend.py", line 264, in get_jit_def
    return build_def(parsed_def.ctx, fn_def, type_line, def_name, self_name=self_name, pdt_arg_types=pdt_arg_types)
  File "./torch/jit/frontend.py", line 302, in build_def
    param_list = build_param_list(ctx, py_def.args, self_name, pdt_arg_types)
  File "./torch/jit/frontend.py", line 326, in build_param_list
    raise NotSupportedError(ctx_range, _vararg_kwarg_err)
torch.jit.frontend.NotSupportedError: Compiled functions can't take variable number of arguments or use keyword-only arguments with defaults:
  File "./vision/torchvision/utils.py", line 380
    def wrapper(*args, **kwargs):
                        ~~~~~~~ <--- HERE
        event = f.__module__
        if f.__name__.endswith("__init__"):
```

This is a known limitation for JIT, so I don't think there is any way around it.

vadimkantorov:
Maybe JIT could provide its own version of functools.wraps? Or its own decorator-maker? Worth opening an issue about this in PyTorch core?

Various decorators like logging or forced autocasts are useful in many contexts

PR #4980, opened 2021-11-23
fix `fromfile` on windows

pmeier:
In some situations, simply calling `mmap.mmap()` on a file that doesn't support it on windows, corrupts the internal state and no data can be read anymore. See https://github.com/pytorch/data/runs/4406380310?check_suite_focus=true for an example.

Thus, we skip the speedup on Windows entirely.

cc @peterjc123 @nbcsm @guyang3532 @maxluk @gunandrose4u @mszhanyi @pmeier @bjuncek


torchvision/prototype/datasets/_builtin/mnist.py
```diff
@@ -75,6 +75,8 @@ def __iter__(self) -> Iterator[torch.Tensor]:
                 file.seek(num_bytes_per_value * count * start, 1)
 
             for _ in range(stop - start):
+                print("#" * 80)
+                print(f"{shape}: {_}")
```
NicolasHug, line 79: can probably be removed :p

NicolasHug (approved):
Thanks @pmeier , LGTM when green!

torchvision/prototype/datasets/_builtin/mnist.py
pmeier: :facepalm: Thanks

PR #4982, opened 2021-11-23
Fix Torchvision iOS nightly build

husthyc:
1. Change the PyTorch binary dependency from libtorch_ios_nightly_build.zip to libtorch_lite_ios_nightly_build.zip
2. Upgrade the testapp from using LibTorch to LibTorch-Lite
3. Fix some build errors for the testapp project


husthyc:
The CI failure is due to the LibTorch-Lite binary on 11/24 has not been generated yet. Since there is a guarantee that the Torchvision nightly build will happen after the LibTorch-Lite nightly, these two jobs will be successful when the nightly builds are triggered.

datumbox:
@husthyc Is this something that will make our CI appear broken often? If this will be a source of flakiness, we should consider removing the date from the file name and access the latest available nightly. Thoughts?

husthyc:
PR updated:
1. Change the PyTorch binary dependency from libtorch_ios_nightly_build.zip to libtorch_lite_ios_nightly_build.zip
2. Upgrade the testapp from using LibTorch to LibTorch-Lite
3.. Fix some build errors for the testapp project

husthyc:
Close this one because unrelated changes involved. Will create a new one just for VisionTestApp setups.

PR #4986, opened 2021-11-24
Fix CircleCI yaml

bigfootjon:
This is breaking Dr. CI in a subtle way. We can't promote the viable/strict branch due to this.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1496,7 +1496,7 @@ workflows:
           filters:
             branches:
               only:
-              - nightly
+                - nightly
             tags:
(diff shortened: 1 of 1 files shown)


prabhat00155 (commented):
Thanks @bigfootjon! I think you need to [sync with congif.yml.in](https://app.circleci.com/pipelines/github/pytorch/vision/12847/workflows/b54184a3-6c64-449f-84be-859a1804a98e/jobs/1021292).

bigfootjon:
This doesn't seem possible to fix, and I think it's actually FB's parser that is broken here. Closing and I'll try and fix internally

Issue #4985, opened 2021-11-24
`cmake_windows_gpu` and `cmake_windows_cpu` failures


NicolasHug:
These have been failing on PRs recently

- https://app.circleci.com/pipelines/github/pytorch/vision/12828/workflows/7495fce3-b084-4e7b-83a8-32cd5dc888a2/jobs/1019628
- https://app.circleci.com/pipelines/github/pytorch/vision/12828/workflows/7495fce3-b084-4e7b-83a8-32cd5dc888a2/jobs/1019604

Both with 

```
Searching for numpy
Reading https://pypi.org/simple/numpy/
Downloading https://files.pythonhosted.org/packages/7a/7f/5384d4c85a2349bd89ff1c0253bff77c7b617e48af201f2d823fc619189a/numpy-1.22.0rc1.zip#sha256=bc991b3f8ea7c0f6703df2bc23c098cfe6f1a3a5e8a3a901eb6a5619275d53ff
Best match: numpy 1.22.0rc1
Processing numpy-1.22.0rc1.zip
Writing C:\Users\circleci\AppData\Local\Temp\easy_install-wlc558u4\numpy-1.22.0rc1\setup.cfg
Running numpy-1.22.0rc1\setup.py -q bdist_egg --dist-dir C:\Users\circleci\AppData\Local\Temp\easy_install-wlc558u4\numpy-1.22.0rc1\egg-dist-tmp-9va09qdc
C:\tools\miniconda3\lib\site-packages\torch\utils\cpp_extension.py:317: UserWarning: Error checking compiler version for cl: [WinError 2] The system cannot find the file specified
  warnings.warn(f'Error checking compiler version for {compiler}: {error}')
Traceback (most recent call last):
  File "C:\tools\miniconda3\lib\site-packages\setuptools\sandbox.py", line 154, in save_modules
    yield saved
  File "C:\tools\miniconda3\lib\site-packages\setuptools\sandbox.py", line 195, in setup_context
    yield
  File "C:\tools\miniconda3\lib\site-packages\setuptools\sandbox.py", line 250, in run_setup
    _execfile(setup_script, ns)
  File "C:\tools\miniconda3\lib\site-packages\setuptools\sandbox.py", line 45, in _execfile
    exec(code, globals, locals)
  File "C:\Users\circleci\AppData\Local\Temp\easy_install-wlc558u4\numpy-1.22.0rc1\setup.py", line 34, in <module>
    
RuntimeError: Python version >= 3.8 required.

```

[...]

PR #4987, opened 2021-11-24
Add numpy as explicit dependency to build_cmake.sh

malfet:
Otherwise, it `setuptools.py`  will try to install latest, which is not compatible with Python runtime older than 3.8

Fixes https://github.com/pytorch/vision/issues/4985

cc @seemethere

### DIFF
--- a/packaging/build_cmake.sh
+++ b/packaging/build_cmake.sh
@@ -42,7 +42,7 @@ else
   PYTORCH_MUTEX_CONSTRAINT=''
 fi
 
-conda install -yq \pytorch=$PYTORCH_VERSION $CONDA_CUDATOOLKIT_CONSTRAINT $PYTORCH_MUTEX_CONSTRAINT $MKL_CONSTRAINT -c "pytorch-${UPLOAD_CHANNEL}"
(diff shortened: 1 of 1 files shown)


atalman (approved):
LGTM

packaging/build_cmake.sh
```diff
@@ -42,7 +42,7 @@ else
   PYTORCH_MUTEX_CONSTRAINT=''
 fi
 
-conda install -yq \pytorch=$PYTORCH_VERSION $CONDA_CUDATOOLKIT_CONSTRAINT $PYTORCH_MUTEX_CONSTRAINT $MKL_CONSTRAINT -c "pytorch-${UPLOAD_CHANNEL}"
+conda install -yq \pytorch=$PYTORCH_VERSION $CONDA_CUDATOOLKIT_CONSTRAINT $PYTORCH_MUTEX_CONSTRAINT $MKL_CONSTRAINT numpy -c "pytorch-${UPLOAD_CHANNEL}"
```
seemethere, line 45: Should we pin it to something?

malfet: No need I guess, as it should be compatible with all the tests in question

PR #4989, opened 2021-11-24
[iOS] Fix VisionTestApp setups

husthyc:
The project file change includes:
1. Removed the "install" folder from the navigation area, the user needs to manually drag this folder into Xcode, because we cannot assume that the file structure of the folder is a fixed one, it may change along with the updates of LibTorch.
2. Added `Accelerate.framework`, `MetalPerformanceShaders.framework` and `CoreML.framework` for linking. Because LibTorch now introduces GPU and CoreML as backends. Otherwise, it may have "symbol not found" errors.

If we want to get rid of the manual work to set up the project, we can improve the setup script in the future. Now, this change just focus on resolving the build errors.


datumbox (approved):
The CI for the iOS builds looks good. Thanks @husthyc.

Issue #1311, opened 2019-09-08
Space/time complexity of batched_nms grows quadratically with batch size


ppwwyyxx:
`batched_nms` now performs a regular NMS for all boxes combined together. Since NMS takes O(n^2) space and time, the complexity of `batch_nms` grows quadratically with batch size.

Ideally the complexity should grow linearly with batch size.

fmassa:
Indeed, we should have a custom implementation which is linear with the batch size.

I believe a naive for-loop implementation in C++ for CPU would be enough, but for CUDA it will probably not be very easy to write.

NicolasHug:
Here's a benchmark comparing the current `batched_nms` implementation with a manual for loop, as done in [detectron2](https://github.com/facebookresearch/detectron2/blob/d6e8c1309b93068c97f89f28bba09190ccc4da6a/detectron2/layers/nms.py#L32)


<details>
<summary>Benchmark code</summary>

```py
from time import time
from collections import defaultdict

from typing import List
import torch
from torchvision.ops import nms, batched_nms
import pandas as pd


N_BOXES_LIST  = [100, 1_000, 2_000, 3_000, 4_000, 5_000, 6_000, 10_000, 50_000, 100_000]
AVERAGE_OVER = 5


def batched_nms_for_loop(boxes, scores, idxs, iou_threshold):

    result_mask = scores.new_zeros(scores.size(), dtype=torch.bool)
    for id in torch.jit.annotate(List[int], torch.unique(idxs).cpu().tolist()):
        mask = (idxs == id).nonzero(as_tuple=False).view(-1)
        keep = nms(boxes[mask], scores[mask], iou_threshold)
        result_mask[mask[keep]] = True
    keep = result_mask.nonzero(as_tuple=False).view(-1)
    keep = keep[scores[keep].argsort(descending=True)]
    return keep

def bench(func, device):

    iou_threshold = .9

    times = []
    for n_boxes in N_BOXES_LIST:

        boxes = torch.cat((torch.rand(n_boxes, 2), torch.rand(n_boxes, 2) + 10), dim=1).to(device)
        assert max(boxes[:, 0]) < min(boxes[:, 2])  # x1 < x2
        assert max(boxes[:, 1]) < min(boxes[:, 3])  # y1 < y2

        scores = torch.rand(n_boxes).to(device)
        idxs = torch.randint(0, 4, size=(n_boxes,)).to(device)

        tic = time()
        for _ in range(AVERAGE_OVER):
[...]

PR #4990, opened 2021-11-24
Change batched NMS threshold for-loop version

ppwwyyxx:
According to the benchmark link https://github.com/pytorch/vision/issues/1311#issuecomment-781329339, for GPU the threshold should be higher (which is why detectron2 used 40k).
This PR changes the threshold to be device dependent.


NicolasHug (commented):
Thanks for the PR @ppwwyyxx 

We decided to use the same threshold for CPU and GPU to keep things as simple as possible and avoid potential surprises.

Do you have a use-case where there are more than 4k boxes on a GPU?

ppwwyyxx:
LVIS has 1200 classes. So in the last NMS of a faster r-cnn trained on LVIS it's common to have 10k-30k boxes.

NicolasHug (approved):
Thanks for the details @ppwwyyxx . The changes LGTM but I'm wondering, did you observe a significant time improvement with the new threshold?

ppwwyyxx:
IIRC we added thresholding in detectron2 because batched_nms OOM for a specific model with too many boxes. Then we did some benchmarks and determined 40k threshold for GPU. This is similar to the conclusion in #1311. 
D17157130 has some discussions at the time.

fmassa (approved):
Thanks for the PR @ppwwyyxx !

I believe this is a net improvement. I think the previous threshold was indeed not ideal for GPU, and https://github.com/pytorch/benchmark/pull/547#issuecomment-970487925 seems to indicate that.

PR #4995, opened 2021-11-26
Make `transforms.functional_tensor` functions differential w.r.t. their parameters

ain-soph:
Linked Issue: #5000 

Make operations in `torchvision.transforms.functional_tensor` differential w.r.t. hyper-parameters, which is helpful for Faster AutoAugment search (hyper-parameters are learnable parameters via backward). 

Some operations are not differential (e.g., Posterize), which might require users to write their own implementations.

### Todo List:
- [ ] JIT support
- [x] Affine has no backward gradient w.r.t. matrix.
- [x] Brightness, Contrast, Hue, Saturation, Gamma, Sharpness, Solarize
- [x] Affine, Rotate (require some modification in `torchvision.transforms.functional`)
- [x] corresponding annotation of methods in `torchvision.transforms.functional`
- [x] (solved in #4999 )`autocontrast` seems to be incorrect. Need further check.

https://github.com/pytorch/vision/blob/18cf5ab1c1307e9d0e1b1e4d26ce9c45e04f8376/torchvision/transforms/functional_tensor.py#L935-L945

The input Image could be either `(N, C, H, W)` or `(C, H, W)`, which makes L940 to rely on different dimensions. This is inconsistent behavior for these 2 cases.

Other implementations as reference:  
https://github.com/moskomule/dda/blob/master/dda/functional.py#L203  
https://pillow.readthedocs.io/en/stable/_modules/PIL/ImageOps.html#autocontrast

-------------------------------------------------------------------
If any maintainer thinks this worth doing, I'll continue to work on it.


datumbox:
@ain-soph Thanks for opening this.

What you propose is interesting but it might be worth discussing the API prior writing the PR. This is to avoid a situation where you spend lots of time working on the feature and not being able to merge due to some unforeseen limitation (JIT might give us headaches for example). Since you already opened a PR, we can discuss this here as well if you prefer.

Concerning the autoaugment behaviour, I would like to understand more why you say the behaviour is inconsistent. 

> The input Image could be either (N, C, H, W) or (C, H, W), which makes L940 to rely on different dimensions. This is inconsistent behavior for these 2 cases. 

My understanding is that the lines at question always look for the min/max across H and W no matter the input. The next 2 lines ensure that we won't divide by 0. Could you provide an example that show-cases the issue?

ain-soph:
@datumbox 

```python3
import torch


a=torch.rand(1,3,32,32)
a=a/2+0.3

(F.autocontrast(a)[0]==F.autocontrast(a[0])).all()   # True

a[0,2]=0.7

(F.autocontrast(a)[0]==F.autocontrast(a[0])).all()   # False
```

We should expect it to be True. Because `(1,C,H,W)` should be same as `(C,H,W)`. But the `[0]` in L940 makes it rely on the first dimension of the image, which might be `N` or `C`.

datumbox:
@ain-soph Thanks a lot for reporting and for submitting a snippet that reproduces the issue. I submitted an alternative solution for this at #4999. Let me know your thoughts.

ain-soph:
@datumbox 

Almost done (except JIT, which I'm totally unfamiliar). But there is one issue about the `Affine`.  
The current implementation seems to have no gradient w.r.t. `matrix` (all zero).

https://github.com/pytorch/vision/blob/47281bbf9b8cd7be952ec23c69c0f62e4733c2a2/torchvision/transforms/functional_tensor.py#L694-L704

I have debugged and find `grid.sum().backward()` will have gradient w.r.t. `matrix`, but L704 loses the gradient information, which means `_apply_grid_transform(img, grid, interpolation, fill=fill)` has no gradient w.r.t. `grid`.

However, it relies on `torch.nn.functional.grid_sample`, which seems to be a C++ implementation. Any further thoughts on how to make it differential?

I tried `kornia.geometry.transform.affine` ([link](https://kornia.readthedocs.io/en/latest/geometry.transform.html?highlight=affine#kornia.geometry.transform.affine)) and it works well.

vfdev-5:
@ain-soph I agree with @datumbox that prior to the PR it would be better to open a feature request issue and give motivations about the requested feature.

According to the code in the PR, your idea is to propagate grad through the transformations and make sure that arguments could be of type `nn.Parameters` or tensors with grads. I think if we could make it without any BC breaking changes, it would be nice otherwise we have to argue why it is more important to keep gradients and make args learnable vs previously working API.

As for grid_sample op and grad propagation for grid, this is done by pytorch. For example:
```python
import torch
from torch.nn.functional import grid_sample

img = torch.rand(2, 3, 4, 4, requires_grad=True)
grid = torch.rand(2, 6, 6, 2, requires_grad=True)

out = grid_sample(img, grid, align_corners=False)
assert out.requires_grad
out.mean().backward()
img.grad is not None, grid.grad is not None
> True, True
```

ain-soph:
@vfdev-5 Thanks for your response!

You are right, `grid_sample` has backward support. I've tested more cases and find the reason why it doesn't work here is that `mode="nearest"`. If I change it to `bilinear` everything works.

Don't know if it is the expected behavior to be all zero gradient when `mode="nearest"`.


I'll later propose a feature request issue linked to this PR.

ain-soph:
I think the `rotate` function seems to be unnecessary since it's already covered by `affine`.  
I think it might be a better option to deprecate `rotate` and arrange default values for arguments (e.g., `angle=0.0, shear=[0.0, 0.0], transpose=[0.0, 0.0]`) in `affine`, so that we only need to call `affine(tensor, shear=shear)` without setting other arguments.

vfdev-5:
> I think the rotate function seems to be unnecessary since it's already covered by affine. It might be a better option to deprecate rotate to make API more unified. But it's totally fine to keep current structure.

@ain-soph due to historical reasons it was first implemented rotate and then affine ops. Rotate API is not completely redundant compared to affine op, as rotate has flag `expand` which is not present for affine op.
I do not think that deprecating this op in favor of affine is a good idea now. Replacing the implementation of rotate and use affine behind can be checked, especially in terms of performances on PIL and tensors (it may appear that rotate can be a bit faster then affine with angle only)

> It would be helpful if there are default values for arguments (e.g., angle=0.0, shear=[0.0, 0.0], transpose=[0.0, 0.0]) in affine, so that we only need to call affine(tensor, shear=shear) without setting other arguments.

Yes, I think default values could be helpful in this case.

vfdev-5:
> Not sure if it is the expected behavior to be all zero gradient when mode="nearest". It seems correct to me that nearest sampling makes the gradient irrelevant to grid.

yes, it is expected (https://github.com/pytorch/pytorch/blob/ca945d989ac56e0070f82fbe657d078610043ea3/aten/src/ATen/native/cpu/GridSamplerKernel.cpp#L806). Also, take a look at section 3.3 "Differentiable Image Sampling" of [this paper](https://arxiv.org/pdf/1506.02025.pdf) where they derive forward pass formulas for nearest and bilinear modes and how to compute grid gradients for bilinear case.

> Maybe we should modify the default interpolation mode to be bilinear?

I think we should not change default value to cover particular case of non-zero grid grads. For you use-case you can specify interpolation mode and thus obtain will obtain non-zero grads.

ain-soph:
@vfdev-5 So you preference is that the default argument value of `affine` is still `interpolation='nearest'`?  
If so, I'll modify the codes.

And any idea on how to fix this issue?

```python3
        center_f[0] = 1.0 * (center[0] - img_size[0] * 0.5)
        center_f[1] = 1.0 * (center[1] - img_size[1] * 0.5)
```

Error Log:

```
  File "/root/project/test/test_functional_tensor.py", line 62, in <module>
    class TestRotate:
  File "/root/project/test/test_functional_tensor.py", line 65, in TestRotate
    scripted_rotate = torch.jit.script(F.rotate)
  File "/root/project/env/lib/python3.7/site-packages/torch/jit/_script.py", line 1312, in script
    qualified_name, ast, _rcb, get_default_args(obj)
RuntimeError: 
'Union[Tensor, List[int]]' object is not subscriptable:
  File "/root/project/torchvision/transforms/functional.py", line 1081
        img_size = get_image_size(img)
        # Center values should be in pixel coordinates but translated such that (0, 0) corresponds to image center.
        center_f[0] = 1.0 * (center[0] - img_size[0] * 0.5)
                             ~~~~~~~~~ <--- HERE
        center_f[1] = 1.0 * (center[1] - img_size[1] * 0.5)
```

The `center_f` is `Union[Tensor, List[int]]` right now.

vfdev-5:
@ain-soph checkout https://github.com/vfdev-5/vision/commit/1214e6f1256f3bb403c0a1b29464e2c6e7d82859
It still does not pass JIT checks for center as tuple of ints but could be tweaked later...

ain-soph:
> @ain-soph checkout [vfdev-5@1214e6f](https://github.com/vfdev-5/vision/commit/1214e6f1256f3bb403c0a1b29464e2c6e7d82859) It still does not pass JIT checks for center as tuple of ints but could be tweaked later...
> 
> > So your preference is that the default argument value of affine is still interpolation='nearest'?
> 
> yes, let's make things less BC-breaking possible: no public API breaks, but only private methods changes

I believe it is a typo in the main branch and `center` type should be `List[float]` rather than `List[int]`.  
I'm also working on the JIT issue. Hope I can fix it tonight.

vfdev-5:
@ain-soph let's make the following to avoid debugging on the CI, I'll finish up the code for `rotate` that will pass JIT etc in another PR and you could then make other color-related transforms and affine. What do you think ?

ain-soph:
@vfdev-5 Oops, sorry if it causes any trouble when I debug...

I try to debug locally, but it seems I have to do some extra work from source code to final binary version.

I'll appreciate if you can help me with JIT issues. I have no experience with JIT previously. I'm still facing a ton of errors, even after I have fixed a bunch of them.

ain-soph:
@vfdev-5 I've get my local environment ready for debugging, and it seems all tests have passed except the new differentiable test that you added.  

Here are the only 2 issues remaining: 
- (personal guess) the current type of `angle` declared in `rotate` is `float`, which makes the `Tensor` converted to `float` in JIT script function. Therefore, the gradient information is lost.
- the current type of `center` declared in `rotate` is `List[float]`, but the test example passes `Tensor`

ain-soph:
And another CI error that I can't reproduce at local environment:

```
test_random_rotate[85-InterpolationMode.NEAREST-degrees3-True-center0-cuda] - test.test_transforms_tensor


  File "C:\Users\circleci\project\test\test_transforms_tensor.py", line 555, in test_random_rotate
    _test_transform_vs_scripted(transform, s_transform, tensor)
  File "C:\Users\circleci\project\test\test_transforms_tensor.py", line 29, in _test_transform_vs_scripted
    assert_equal(out1, out2, msg=msg)
  File "C:\Users\circleci\project\env\lib\site-packages\torch\testing\_comparison.py", line 1217, in assert_close
    assert_equal(
  File "C:\Users\circleci\project\env\lib\site-packages\torch\testing\_comparison.py", line 997, in assert_equal
    raise error_metas[0].to_error()

Mismatched elements: 3 / 9333 (0.0%)
Greatest absolute difference: 43 at index (0, 3, 2) (up to 1e-06 allowed)
Greatest relative difference: 0.5058823823928833 at index (0, 3, 2) (up to 0 allowed)
```

ain-soph:
close in favor of #5110

PR #4999, opened 2021-11-28
Fix bug on autocontrast when `min==max`

datumbox:
Fixes issue reported by @ain-soph at https://github.com/pytorch/vision/pull/4995#issuecomment-980352057

cc @vfdev-5 @datumbox

### DIFF
--- a/test/test_functional_tensor.py
+++ b/test/test_functional_tensor.py
@@ -821,6 +821,18 @@ def test_autocontrast(device, dtype, channels):
 
 
 @pytest.mark.parametrize("device", cpu_and_gpu())
+@pytest.mark.parametrize("dtype", (None, torch.float32, torch.float64))
+@pytest.mark.parametrize("channels", [1, 3])
+def test_autocontrast_equal_minmax(device, dtype, channels):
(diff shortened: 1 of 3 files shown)


ain-soph:
Thanks for your response, but this fix doesn't seem to be correct?

You are replacing NAN with 1.0, which is not expected.

Constant across channel should still keep the value after autocontrast, rather than replacing them to 1.0.

Am I understanding autocontrast correctly?

ain-soph:
Current fix seems to implement **(channel-wise)** normalization as I illustrated in https://github.com/pytorch/vision/pull/4995#issuecomment-980352057. Is this expected? Or shall we use **(image-wise)**?

datumbox:
@ain-soph I believe the correct is to do channel-wise. 

I've added one more test that shows/confirms this. Below we are comparing the output of PIL vs TorchVision implementation for an image that has constant values across channels:
https://github.com/pytorch/vision/blob/35de04cd87d4b6eac361548093faed0644b6fece/test/test_transforms.py#L576-L582

ain-soph:
That sounds nice.

fmassa (approved):
LGTM, thanks!

PR #5001, opened 2021-11-29
simplify model builders

pmeier:
Addresses #4948.

This PR adds two decorators that help reducing verbose and duplicated code in the prototype model builders:

1. `@handle_legacy_interface` This decorator does two things:
    1. Turn positional into keyword arguments and warn that positional is deprecated.
    2. Perform the conversion between `pretrained` and `weights` and warn that `pretrained` is deprecated.
2. `@handle_num_categories_mismatch` This decorator will raise an error if `num_classes` is passed, but doesn't match the number of categories of the pretrained weights.

Examples:

```python
from torchvision.prototype import models

models.alexnet(True, False)
```

```
UserWarning: Using 'pretrained', and 'progress' as positional parameter(s) is deprecated. Please use them as keyword parameter(s) instead.
UserWarning: The parameter 'pretrained' is deprecated, please use 'weights' instead.
```

```python
models.detection.fasterrcnn_resnet50_fpn(True, False, 0, True)
```
```
UserWarning: Using 'pretrained', 'progress', 'num_classes', and 'pretrained_backbone' as positional parameter(s) is deprecated. Please use them as keyword parameter(s) instead.
UserWarning: The parameter 'pretrained' is deprecated, please use 'weights' instead.
UserWarning: The parameter 'pretrained_backbone' is deprecated, please use 'weights_backbone' instead.
ValueError: The number of categories of the weights does not match the `num_classes` argument: 91 != 0.
```

cc @datumbox @bjuncek


torchvision/prototype/models/_utils.py
```diff
@@ -1,13 +1,123 @@
...
+            kwargs = handle_pretrained_to_weights(builder, kwargs, default_weights=default_weights)
+
+            return builder(*args, **kwargs)
+
+        return inner_wrapper
+
+    return outer_wrapper
+
+
+def handle_num_categories_mismatch(*, parameter: str = "num_classes") -> Callable[[Callable[..., M]], Callable[..., M]]:
```
pmeier, line 104: Is this always called `num_classes`? If yes, we can hardcode it.

torchvision/prototype/models/quantization/shufflenetv2.py
```diff
@@ -24,18 +24,20 @@
...
     **kwargs: Any,
 ) -> QuantizableShuffleNetV2:
-    if weights is not None:
-        _ovewrite_named_param(kwargs, "num_classes", len(weights.meta["categories"]))
-        if "backend" in weights.meta:
-            _ovewrite_named_param(kwargs, "backend", weights.meta["backend"])
+    # FIXME
+    # if weights is not None:
+    # if "backend" in weights.meta:
+    #     _ovewrite_named_param(kwargs, "backend", weights.meta["backend"])
```
pmeier, line 40: Is this a one-time-thing or should we have another decorator for that?

NicolasHug:
> Turn positional into keyword arguments and warn that positional is deprecated.

I think this is great and this is something we have discussed with @datumbox already, but I would prefer waiting a bit before doing it here: I'd like this to be a more pervasive change throughout torchvision, not just something specific to the models. Also, I'd like us to have a clearer deprecation policy, i.e. decide when we actually remove stuff. I want to get this done before then next release anyway, likely in January (release is planned for late Feb I think)


Also, this is only tangentially related to this PR, so feel free not to address it here: it'd be great if we could advertise the new API in the deprecation message, and also help users with the transition. So ideally, the message could be something like

> The pretrained parameter is deprecated, please use the weights parameter instead. The current behaviour is equivalent to passing weights=ResnetWeights.Blahblah. You can also set weights="default" to get the most up-to-date weights.

pmeier:
> I'd like this to be a more pervasive change throughout torchvision, not just something specific to the models.

I can turn the positional to keyword only handling into a general decorator if you want. That way we can move forward with the model builders by using said decorator and handle all other cases later by simply using it. 

> Also, I'd like us to have a clearer deprecation policy, i.e. decide when we actually remove stuff.

In PyTorch core the "regular" policy is to deprecate for one release and unless there is major pushback remove it the next release.

> I want to get this done before then next release anyway, likely in January (release is planned for late Feb I think)

Tentative release data is 04.03.2022.

> Also, this is only tangentially related to this PR, so feel free not to address it here: it'd be great if we could advertise the new API in the deprecation message, and also help users with the transition. So ideally, the message could be something like
> 
> > The pretrained parameter is deprecated, please use the weights parameter instead. The current behaviour is equivalent to passing weights=ResnetWeights.Blahblah. You can also set weights="default" to get the most up-to-date weights.

@datumbox What would a good message look like? Please use `$FOO` or whatever to indicate which parts need to be parametrized.

torchvision/prototype/models/quantization/shufflenetv2.py
datumbox: This is again part of the building mechanism logic. I wouldn't move this to a decorator. Also it's not a one-time thing, it's quite common for all quantization models.

datumbox (commented):
@pmeier I love the direction of this. I've added a few comments below, please have a look. Note that there is a final refactoring PR that is going to affect you at #5006. Hopefully we will merge it later today and avoid conflicts.

> I would prefer waiting a bit before doing it here: I'd like this to be a more pervasive change throughout torchvision, not just something specific to the models.

I agree we should roll this out in the main area in a coordinated manner. Concerning the prototype area, I would be open adopting it early to avoid introducing changes over and over on the API. The Datasets API already uses it, so it makes sense to have it here as well. Thoughts?

> I can turn the positional to keyword only handling into a general decorator if you want. That way we can move forward with the model builders by using said decorator and handle all other cases later by simply using it.

Sounds like a good idea. I'll love to see the details.

> What would a good message look like?

I think the message provided by @NicolasHug looks great. The words `pretrained`, `weights` and `ResnetWeights.Blahblah` are the parameterized parts, which you already handle.

torchvision/prototype/models/alexnet.py
```diff
@@ -29,16 +32,10 @@ class AlexNetWeights(Weights):
...
-        _deprecated_positional(kwargs, "pretrained", "weights", True)
-    if "pretrained" in kwargs:
-        weights = _deprecated_param(kwargs, "pretrained", "weights", AlexNetWeights.ImageNet1K_RefV1)
+@handle_legacy_interface(AlexNetWeights.ImageNet1K_RefV1)
+@handle_num_categories_mismatch()
+def alexnet(*, weights: Optional[AlexNetWeights] = None, progress: bool = True, **kwargs: Any) -> AlexNet:
     weights = AlexNetWeights.verify(weights)
-
-    if weights is not None:
-        _ovewrite_named_param(kwargs, "num_classes", len(weights.meta["categories"]))
```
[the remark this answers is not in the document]
pmeier: Could you explain your thinking behind this? IIUC, this is input validation and if multiple functions need the same, a decorator fits well. That way the body of the builder becomes more readable because it can focus on on the actual functionality.

NicolasHug:
> I agree we should roll this out in the main area in a coordinated manner. Concerning the prototype area, I would be open adopting it early to avoid introducing changes over and over on the API. The Datasets API already uses it, so it makes sense to have it here as well. Thoughts?


The *new* datasets make use of the positional argument, which is perfectly fine since they're new classes. But here we're deprecating exising functions and existing usage - I would prefer to do that in a few weeks when we have a good idea of what we want the deprecation process to look like.

torchvision/prototype/models/alexnet.py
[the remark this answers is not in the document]
datumbox: Sure. This is not input validation but affects the model config which is key part of the method builder. This is also where the default value of the parameter lives (because the signature defines it as `None`). The idiom of modifying the model configuration prior calling the model class is very common across TorchVision and belongs in the method builder. You can see similar examples on the InceptionV3 builder and on the `_validate_trainable_layers()` calls of all detection models. Similarly `num_classes`, `num_keypoints` etc are building parameters of the model classes and changing their configuration should remain inside the model builder.

torchvision/prototype/models/_utils.py
```diff
@@ -1,13 +1,123 @@
...
+from torchvision.prototype.utils._internal import sequence_to_str
 
+from ._api import Weights
 
 W = TypeVar("W", bound=Weights)
+M = TypeVar("M", bound=nn.Module)
 V = TypeVar("V")
 
 
+def handle_positional_to_keyword_only(
```
NicolasHug, line 17: I would suggest to use https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/utils/validation.py#L36 instead. It's been extensively used in scikit-learn and it's just a decorator, no need to pass anything

NicolasHug:
We had a chat with @datumbox .

I'm fine with introducing the keyword-only deprecation here. That means that we need to get to an agreement regarding the keyword deprecations for the rest of the code base **before** we move the new models from prototypes to main. If we can't reach that point (unlikely, but possible), we would have to remove the deprecations.

torchvision/prototype/models/_utils.py
pmeier: Yes, this is what I meant in https://github.com/pytorch/vision/pull/5001#issuecomment-982412063. I've added a general decorator in the latest commit.

torchvision/prototype/utils/_internal.py
```diff
@@ -126,3 +144,70 @@ def __delattr__(self, item: Any) -> NoReturn:
 
     def __repr__(self) -> str:
         return make_repr(type(self).__name__, self.items())
+
+
+def kwonly_to_pos_or_kw(
+    *,
+    param_map: Optional[Dict[str, str]] = None,
+    warn: bool = False,
```
NicolasHug, line 152: Are we ever going to call this with warn=False?

torchvision/prototype/utils/_internal.py
```diff
@@ -126,3 +144,70 @@ def __delattr__(self, item: Any) -> NoReturn:
 
     def __repr__(self) -> str:
         return make_repr(type(self).__name__, self.items())
+
+
+def kwonly_to_pos_or_kw(
+    *,
+    param_map: Optional[Dict[str, str]] = None,
```
NicolasHug, line 151: It's not clear to me what this does, could you detail?

datumbox:
> If we can't reach that point before the next release (unlikely, but possible), we would have to remove the deprecations.

Agreed. I've captured this point step 1 of the release tasks at #4679

torchvision/prototype/utils/_internal.py
pmeier: As explained in the docstring, this is used to map the parameter names. Let's say you had a function

```python
def foo(bar):
    pass
```

and the new signature looks like 

```python
def foo(*, baz):
   pass
```

If we don't have a map for `"baz": "bar"`, the warning would read something like "The parameter 'baz' is deprecated" although the user only ever used `bar`. I think it is easier to understand for the user this way, but no strong opinion.

torchvision/prototype/utils/_internal.py
pmeier: I don't know. Let's resolve https://github.com/pytorch/vision/pull/5001/files#r759167502 first. If we remove `param_map`, the case for removing this is stronger, since it could simplify the whole decorator.

torchvision/prototype/utils/_internal.py
NicolasHug: Do we have such a scenario where we plainly remove a parameter?

I believe we only replace `pretrained` with `weights`. In this case, users should get a deprecation warning telling them "Use weights, not pretrained". Once they address this, I'm not sure the mapping is relevant?

torchvision/prototype/utils/_internal.py
NicolasHug: My stance here is that if we're introducing keyword-only parameters, we should deprecate the use of positional arguments, and thus warn.

Otherwise this whole decorator seems like a no-op?

torchvision/prototype/utils/_internal.py
pmeier: The problem is that we do both, convert to keyword only and rename the parameter. For example, removing the map will result in the following warnings:

```python
from torchvision.prototype import models

models.alexnet(True)
```

```
UserWarning: Using 'weights' as positional parameter(s) is deprecated. Please use it/them as keyword parameter(s) instead.
UserWarning: The parameter 'pretrained' is deprecated, please use 'weights' instead.
```

which I think is quite confusing. Without manually telling the decorator that there was a `pretrained` parameter before, it can't generate a fitting warning.

NicolasHug: Thanks for the explanation @pmeier , I agree the 2 warnings might be confusing

datumbox: I think this mapping of parameters is something also proposed by you Nicolas in our early discussions when I had changed the order of parameters. Could you clarify if your concern is around the way this is implemented or over the need to inform people to stop using positional arguments?

torchvision/prototype/utils/_internal.py
pmeier: It won't be a no-op since the conversion of the parameters would still happen, but no warning would be emitted. I'm ok with either behavior.

torchvision/prototype/utils/_internal.py
pmeier: Let me see if I can refactor the warning generation so both warnings together are not confusing.

NicolasHug: Sorry, I meant I agree with @pmeier 's statement 

> which I think is quite confusing

and I thus agree that this `param_map` is useful

torchvision/prototype/utils/_internal.py
datumbox: My understanding is that this might be useful if we don't manage to finalize the deprecation of positional arguments across the entire torchvision before the release. One way to use it would be to maintain the decorator but inhibit the warning. This will make the conversion of the parameters but won't inform users about the deprecation until we are ready. This is an alternative course of action to removing the mandatory keyword args from the method signature. I don't have a strong opinion over which way to follow.

torchvision/prototype/utils/_internal.py
pmeier: I have a local commit that emits the following warnings in the four use cases without using the manual map:

- `alexnet(True)`
    ```
    UserWarning: Using 'weights' as positional parameter(s) is deprecated. Please use it/them as keyword parameter(s) instead.
    UserWarning: The current behavior is equivalent to passing `weights=AlexNet_Weights.ImageNet1K_V1`. You can also use `weights='default'` to get the most up-to-date weights.
    ```
- `alexnet(False)`
    ```
    UserWarning: Using 'weights' as positional parameter(s) is deprecated. Please use it/them as keyword parameter(s) instead.
    UserWarning: The current behavior is equivalent to passing `weights=None`. You can also use `weights='default'` to get the most up-to-date weights.
    ```
- `alexnet(pretrained=True)`
    ```
    UserWarning: The parameter 'pretrained' is deprecated, please use 'weights' instead. The current behavior is equivalent to passing `weights=AlexNet_Weights.ImageNet1K_V1`. You can also use `weights='default'` to get the most up-to-date weights.
    ```
- `alexnet(pretrained=False)`
    ```
    UserWarning: The parameter 'pretrained' is deprecated, please use 'weights' instead. The current behavior is equivalent to passing `weights=None`. You can also use `weights='default'` to get the most up-to-date weights.
    ```

Happy with them?

torchvision/prototype/utils/_internal.py
pmeier: I've decided to remove it since it makes the implementation easier now that https://github.com/pytorch/vision/pull/5001#discussion_r759167502 is resolved.

pmeier:
The warnings now look as follows:

- `alexnet(True)`
    ```
    UserWarning: Using 'weights' as positional parameter(s) is deprecated. Please use keyword parameter(s) instead.
    UserWarning: The current behavior is equivalent to passing `weights=AlexNet_Weights.ImageNet1K_V1`. You can also use `weights='default'` to get the most up-to-date weights.
    ```
- `alexnet(False)`
    ```
    UserWarning: Using 'weights' as positional parameter(s) is deprecated. Please use keyword parameter(s) instead.
    UserWarning: The current behavior is equivalent to passing `weights=None`.
    ```
- `alexnet(pretrained=True)`
    ```
    UserWarning: The parameter 'pretrained' is deprecated, please use 'weights' instead. The current behavior is equivalent to passing `weights=AlexNet_Weights.ImageNet1K_V1`. You can also use `weights='default'` to get the most up-to-date weights.
    ```
- `alexnet(pretrained=False)`
    ```
    UserWarning: The parameter 'pretrained' is deprecated, please use 'weights' instead. The current behavior is equivalent to passing `weights=None`.
    ```

Happy with them?

torchvision/prototype/models/_utils.py
```diff
@@ -1,13 +1,55 @@
...
+                    continue
+
+                pretrained_positional = pretrained_param not in kwargs
+                pretrained_arg = weights_arg if pretrained_positional else kwargs.pop(pretrained_param)
+                if pretrained_arg:
+                    default_arg = default(kwargs) if callable(default) else default
+                    if default_arg is None:
+                        raise ValueError(f"No checkpoint is available for model {builder.__name__}")
+                else:
+                    default_arg = None
```
datumbox, line 33: IMO the use of `default_arg` in lines 29-30 is not the same as the variable at lines 33-43. One has to do with the default value passed by the decorator and the other has to do with the actual value that is going to be used.

The semantics are tricky because Nones are not allowed on the first (we should throw an error as you do) but can be allowed on the latter. So I would slightly refactor to use different name for the variable at line 33 onwards (let's say `default_value`) and then set after line 31 `default_value = default_arg`.

torchvision/prototype/models/_utils.py
```diff
@@ -1,13 +1,55 @@
...
+
+                msg = f"The current behavior is equivalent to passing `{weights_param}={default_arg}`."
+                if not pretrained_positional:
+                    msg = (
+                        f"The parameter '{pretrained_param}' is deprecated, please use '{weights_param}' instead. {msg}"
+                    )
+                if default_arg is not None:
+                    msg = f"{msg} You can also use `{weights_param}='default'` to get the most up-to-date weights."
```
datumbox, line 41: Thought strings are supported, they are not the preferred way to do it. Instead of `'default'` suggest to use `Some_Weights.default`

torchvision/prototype/models/_utils.py
```diff
@@ -1,13 +1,55 @@
...
 
+def handle_legacy_interface(
+    **pretrained_weights: Tuple[str, Union[Optional[W], Callable[[Dict[str, Any]], Optional[W]]]]
+):
+    def outer_wrapper(builder: Callable[..., M]) -> Callable[..., M]:
+        @functools.wraps(builder)
+        def inner_wrapper(**kwargs: Any) -> M:
+            for weights_param, (pretrained_param, default) in pretrained_weights.items():  # type: ignore[union-attr]
+                weights_arg = kwargs.get(weights_param)
+                if weights_param in kwargs and not isinstance(weights_arg, bool):
```
datumbox, line 23: This correct way of handling it. You need to check again that `weights_param in kwargs` instead of doing `weights_param is not None` to handle properly all cases. But this can be unexpected by readers of the code so please add comments to explain why you need to handle it the way you do. My concern is that someone might refactor it without understanding the implications.

torchvision/prototype/models/quantization/shufflenetv2.py
```diff
@@ -87,47 +88,43 @@ class ShuffleNet_V2_X1_0_QuantizedWeights(WeightsEnum):
     default = ImageNet1K_FBGEMM_V1
 
 
+@handle_legacy_interface(
+    weights=(
+        "pretrained",
+        lambda kwargs: ShuffleNet_V2_X0_5_QuantizedWeights.ImageNet1K_FBGEMM_V1
+        if kwargs.get("quantize", False)
+        else ShuffleNet_V2_X0_5_Weights.ImageNet1K_V1,
```
datumbox, line 96: Perfect. This mechanism should also support the `KeypointRCNN_ResNet50_FPN_Weights.Coco_Legacy` case, but check it does. I would also check all other special cases I marked on an earlier PR before starting replacing it form everywhere to ensure your solution works fine. I think it does.

torchvision/prototype/models/_utils.py
```diff
@@ -1,13 +1,55 @@
...
+                        f"The parameter '{pretrained_param}' is deprecated, please use '{weights_param}' instead. {msg}"
+                    )
+                if default_arg is not None:
+                    msg = f"{msg} You can also use `{weights_param}='default'` to get the most up-to-date weights."
+                warnings.warn(msg)
+                kwargs[weights_param] = default_arg
+
+            return builder(**kwargs)
+
+        return kwonly_to_pos_or_kw(inner_wrapper)
```
datumbox, line 47: Can you put `kwonly_to_pos_or_kw` as a decorator to the `inner_wrapper`?

datumbox (commented):
@pmeier The solution looks good to me. As discussed offline, there are a few minor comments that you could address before start moving your solution in all model builders. Once you are done with resolving the comments, you can start replacing your approach in all model builders. I would recommend to keep your commits clean and "stacked" so that we can review easier. In other words keep separate the commits that require thorough review from those that contain mechanical replaces. 

Thanks so much for writing this, the API will look much nicer when we use your solution.

torchvision/prototype/models/_utils.py
```diff
@@ -1,13 +1,79 @@
...
+                # infer the names of positionally passed arguments and thus has no knowledge that there used to be a
+                # pretrained parameter.
+                pretrained_positional = weights_param in kwargs
+                pretrained_arg = kwargs.pop(weights_param if pretrained_positional else pretrained_param)
+                if pretrained_arg:
+                    default_weights_arg = default(pretrained_arg, kwargs) if callable(default) else default
+                    if not isinstance(default_weights_arg, WeightsEnum):
+                        raise ValueError(f"No weights available for model {builder.__name__}")
+                else:
+                    default_weights_arg = None
```
datumbox, line 54: I thinks this https://github.com/pytorch/vision/pull/5001#discussion_r760984921 remains unaddressed. You have renamed the parameter everywhere. See if you agree with my point on the renaming and happy to discuss this if you dont.

torchvision/prototype/models/_utils.py
```diff
@@ -1,13 +1,79 @@
...
+                    continue
+
+                # If the pretrained parameter was passed as positional argument, it is now mapped to
+                # `kwargs[weights_param]`. This happens because the @kwonly_to_pos_or_kw use the current signature to
+                # infer the names of positionally passed arguments and thus has no knowledge that there used to be a
+                # pretrained parameter.
+                pretrained_positional = weights_param in kwargs
+                pretrained_arg = kwargs.pop(weights_param if pretrained_positional else pretrained_param)
+                if pretrained_arg:
+                    default_weights_arg = default(pretrained_arg, kwargs) if callable(default) else default
```
datumbox, line 50: I noticed that you changed the lambda to include an extra parameter `pretrained_arg` which you don't currently use. Could you elaborate on which corner case requires that?

torchvision/prototype/models/_utils.py
pmeier: I agree in the old implementation this was confusing, but I've refactored a little different. Instead of checking if `default_weights_arg is None`, we now simply enforce that it should be a `Weights` instance. Through this we now have `default_weights_arg: Optional[Weights]`, which hopefully is not confusing anymore. Thoughts?

torchvision/prototype/models/_utils.py
pmeier: The `kwargs` argument passed to the callable did not contain the `pretrain_arg`, because it was already removed at that point. Thus, we pass it explicitly to be used in cases where decision for the default weights need to be made with respect to it. See https://github.com/pytorch/vision/pull/5001#discussion_r760989162 for an example.

torchvision/prototype/models/_utils.py
datumbox: OK you sold it. Sounds good to me :)

torchvision/prototype/models/_utils.py
datumbox: Hmmmm... This makes for a confusing signature. Should we refactor to defer the poping of the param until later to avoid requiring both arguments?

pmeier: I thought about that too, but then we need to decide under what key we put it. Furthermore, if the parameter was passed as positional we need to put it in first to only remove it again after the call.

Finally, what is confusing about

```python
def default(
    pretrained_arg: Any,  # usually bool, but might also be "legacy"
    other_kwargs: Dict[str, Any],
) -> Optional[Weights]:
    ...
```

datumbox: I don't mind adding it and popping it. In a sense you try to decouple all complex scenarios that can occur by collapsing them to 1-2 possible cases. If its positional, you just plug it on the right keyword argument and pop it later no matter its origin (positional or keyword). The only reason I'm discussing this so extensively is because the signature is quite visible from the model builder methods. I wouldn't mind it if that was a technical detail buried somewhere in the decorators. Let me know what you think.

pmeier: Fair point. The internals of the decorator are a little (I promise :stuck_out_tongue: ) more complicated now, but since they are not "visible" that should be fine.

datumbox:
I reviewed the proposed solution and looks great to me. I think we can implement the proposal on all model builders.

torchvision/prototype/models/segmentation/deeplabv3.py
```diff
@@ -72,25 +72,19 @@ class DeepLabV3_MobileNet_V3_Large_Weights(WeightsEnum):
...
+    weights_backbone=("pretrained_backbone", ResNet50_Weights.ImageNet1K_V1),
+)
 def deeplabv3_resnet50(
+    *,
     weights: Optional[DeepLabV3_ResNet50_Weights] = None,
     progress: bool = True,
     num_classes: Optional[int] = None,
     aux_loss: Optional[bool] = None,
     weights_backbone: Optional[ResNet50_Weights] = None,
-    **kwargs: Any,
```
datumbox, line 81: This changes the signature of the method. Though a corner case, it means that now if someone passed an extra parameter as argument it won't be ignored but instead they will get an exception. Note that `kwargs` is currently supported by all model builders so I suggest leaving it in place.

torchvision/prototype/models/segmentation/deeplabv3.py
```diff
@@ -109,25 +103,19 @@ def deeplabv3_resnet50(
...
+    weights_backbone=("pretrained_backbone", ResNet101_Weights.ImageNet1K_V1),
+)
 def deeplabv3_resnet101(
+    *,
     weights: Optional[DeepLabV3_ResNet101_Weights] = None,
     progress: bool = True,
     num_classes: Optional[int] = None,
     aux_loss: Optional[bool] = None,
     weights_backbone: Optional[ResNet101_Weights] = None,
-    **kwargs: Any,
```
datumbox, line 118: Same

torchvision/prototype/models/segmentation/deeplabv3.py
```diff
@@ -146,27 +134,19 @@ def deeplabv3_resnet101(
...
+    weights_backbone=("pretrained_backbone", MobileNet_V3_Large_Weights.ImageNet1K_V1),
+)
 def deeplabv3_mobilenet_v3_large(
+    *,
     weights: Optional[DeepLabV3_MobileNet_V3_Large_Weights] = None,
     progress: bool = True,
     num_classes: Optional[int] = None,
     aux_loss: Optional[bool] = None,
     weights_backbone: Optional[MobileNet_V3_Large_Weights] = None,
-    **kwargs: Any,
```
datumbox, line 155: Same

torchvision/prototype/models/segmentation/fcn.py
```diff
@@ -48,25 +48,19 @@ class FCN_ResNet101_Weights(WeightsEnum):
...
+    weights_backbone=("pretrained_backbone", ResNet50_Weights.ImageNet1K_V1),
+)
 def fcn_resnet50(
+    *,
     weights: Optional[FCN_ResNet50_Weights] = None,
     progress: bool = True,
     num_classes: Optional[int] = None,
     aux_loss: Optional[bool] = None,
     weights_backbone: Optional[ResNet50_Weights] = None,
-    **kwargs: Any,
```
datumbox, line 57: Same

torchvision/prototype/models/segmentation/fcn.py
```diff
@@ -85,25 +79,19 @@ def fcn_resnet50(
...
+    weights_backbone=("pretrained_backbone", ResNet101_Weights.ImageNet1K_V1),
+)
 def fcn_resnet101(
+    *,
     weights: Optional[FCN_ResNet101_Weights] = None,
     progress: bool = True,
     num_classes: Optional[int] = None,
     aux_loss: Optional[bool] = None,
     weights_backbone: Optional[ResNet101_Weights] = None,
-    **kwargs: Any,
```
datumbox, line 94: Same

datumbox (approved):
@pmeier LGTM, thanks for the major readability improvement. The code looks much much better now.

I've pushed 3 commits, 1 for a missing verify call and 2 for kwargs. If you are OK with it, feel free to merge on green CI. Else revert and let's chat.

torchvision/prototype/models/segmentation/deeplabv3.py
pmeier: > Though a corner case, it means that now if someone passed an extra parameter as argument it won't be ignored but instead they will get an exception.

That is kind of the point. The builder **silently** ignores wrong keyword arguments. Imagine you made a typo

```python
deeplabv3_resnet50(weight=DeepLabV3_ResNet50_Weights.CocoWithVocLabels_V1)
```

and instead of loading some weights, you get a randomly initialized model instead. I don't think this is expected neither a good thing to have. At the very least, we should emit a warning that there were unused keyword arguments.

> Note that `kwargs` is currently supported by all model builders so I suggest leaving it in place.

There are places where this is useful. For example the `**kwargs` of VGG variants like

https://github.com/pytorch/vision/blob/93b26da3623c9134d64f026a234d97caab0675f2/torchvision/prototype/models/vgg.py#L172-L179

will ultimately be resolved in 

https://github.com/pytorch/vision/blob/93b26da3623c9134d64f026a234d97caab0675f2/torchvision/models/vgg.py#L35-L38

So if there is any unknown keyword argument, the class instantiation would fail.

datumbox: @pmeier I understand and I agree with many points you are raising here. My concern is that this is not the right PR to fix this. Fixing it here only resolves the problem for 2 models, leaving the rest affected. Shall we take this on a separate issue?

pmeier: Fair enough :+1:

PR #5003, opened 2021-11-29
Cleanup namings of Multi-weights classes and enums

### LINKED ISSUE #4652: Multi pretrained weights: Cleanups and Refactoring

datumbox:
Fixes partially #4652

- [x] Rename classes `Weights => WeightsEnum` and `WeightEntry => Weights`
- [x] Make enum values follow the naming convention `_V1`, `_V2` etc
- [x] Cleanup the Enum class naming conventions. We settled on the following convention: We will use proper casing so that the class names are more readable but their lowercased names should be `capitalized(model_builder_name) + "_Weights"` or `capitalized(model_builder_name) + "_QuantizedWeights"`. 

cc @bjuncek

### DIFF
--- a/test/test_prototype_models.py
+++ b/test/test_prototype_models.py
@@ -18,6 +18,12 @@ def _get_original_model(model_fn):
     return module.__dict__[model_fn.__name__]
 
 
+def _get_parent_module(model_fn):
+    parent_module_name = ".".join(model_fn.__module__.split(".")[:-1])
+    module = importlib.import_module(parent_module_name)
+    return module
+
+
 def _build_model(fn, **kwargs):
     try:
         model = fn(**kwargs)
@@ -27,11 +33,6 @@ def _build_model(fn, **kwargs):
             pytest.skip(msg)
         raise e
     return model.eval()
-
-
-def get_models_with_module_names(module):
-    module_name = module.__name__.split(".")[-1]
-    return [(fn, module_name) for fn in TM.get_models_from_module(module)]
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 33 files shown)


NicolasHug (approved):
LGTM when green, thanks

NicolasHug (approved):
Looked at each commit independently, still LGTM :)

PR #5004, opened 2021-11-29
Some updates to optical flow datasets

NicolasHug:
This PR does a few different things related to the optical flow datasets:

- always return an RGB image, like for the rest of the datasets (there are some images that come in RGBA or Grayscale but the transforms don't support these, so we need to convert)
- rename `valid` in to `valid_flow_mask` which is clearer
- allow the transforms to generate the `valid_flow_mask` from the `flow` values. This helps keeping the reference code clean.
- A few doc clarifications
- Implement `__rmul__` which helps duplicating a dataset like `100 * sintel`. Might remove eventually, it will depend on the training reference eventually


datumbox (approved):
Stamp

PR #5006, opened 2021-11-29
Refactor the `get_weights` API

### LINKED ISSUE #4652: Multi pretrained weights: Cleanups and Refactoring

datumbox:
Fixes #4652

- [x] Change the `default` weights mechanism to sue Enum aliases.
- [x] Change `get_weights` to work with full Enum names and make it public.
- [x] Test that all reference scripts continue to run.

cc @datumbox @bjuncek

### DIFF
--- a/test/test_prototype_models.py
+++ b/test/test_prototype_models.py
@@ -49,24 +49,22 @@ def _build_model(fn, **kwargs):
 
 
 @pytest.mark.parametrize(
-    "model_fn, name, weight",
+    "name, weight",
     [
-        (models.resnet50, "ImageNet1K_V1", models.ResNet50_Weights.ImageNet1K_V1),
-        (models.resnet50, "default", models.ResNet50_Weights.ImageNet1K_V2),
+        ("ResNet50_Weights.ImageNet1K_V1", models.ResNet50_Weights.ImageNet1K_V1),
+        ("ResNet50_Weights.default", models.ResNet50_Weights.ImageNet1K_V2),
         (
-            models.quantization.resnet50,
-            "default",
+            "ResNet50_QuantizedWeights.default",
             models.quantization.ResNet50_QuantizedWeights.ImageNet1K_FBGEMM_V2,
         ),
         (
-            models.quantization.resnet50,
-            "ImageNet1K_FBGEMM_V1",
+            "ResNet50_QuantizedWeights.ImageNet1K_FBGEMM_V1",
             models.quantization.ResNet50_QuantizedWeights.ImageNet1K_FBGEMM_V1,
         ),
     ],
 )
-def test_get_weight(model_fn, name, weight):
-    assert models._api.get_weight(model_fn, name) == weight
+def test_get_weight(name, weight):
+    assert models.get_weight(name) == weight
 
 
 @pytest.mark.parametrize(
--- a/test/test_prototype_models.py
+++ b/test/test_prototype_models.py
@@ -24,6 +24,19 @@ def _get_parent_module(model_fn):
     return module
 
 
+def _get_model_weights(model_fn):
+    module = _get_parent_module(model_fn)
+    weights_name = "_QuantizedWeights" if module.__name__.split(".")[-1] == "quantization" else "_Weights"
+    try:
+        return next(
+            v
+            for k, v in module.__dict__.items()
+            if k.endswith(weights_name) and k.replace(weights_name, "").lower() == model_fn.__name__
+        )
+    except StopIteration:
+        return None
+
+
 def _build_model(fn, **kwargs):
     try:
         model = fn(**kwargs)
@@ -65,10 +78,9 @@ def test_get_weight(model_fn, name, weight):
     + TM.get_models_from_module(models.video),
 )
 def test_naming_conventions(model_fn):
-    model_name = model_fn.__name__
-    module = _get_parent_module(model_fn)
-    weights_name = "_QuantizedWeights" if module.__name__.split(".")[-1] == "quantization" else "_Weights"
-    assert model_name in set(x.replace(weights_name, "").lower() for x in module.__dict__ if x.endswith(weights_name))
+    weights_enum = _get_model_weights(model_fn)
+    assert weights_enum is not None
+    assert len(weights_enum) == 0 or hasattr(weights_enum, "default")
 
 
 @pytest.mark.parametrize("model_fn", TM.get_models_from_module(models))
--- a/torchvision/prototype/models/_api.py
+++ b/torchvision/prototype/models/_api.py
@@ -30,7 +30,6 @@ class Weights:
     url: str
     transforms: Callable
     meta: Dict[str, Any]
-    default: bool
 
 
 class WeightsEnum(Enum):
@@ -59,9 +58,14 @@ class WeightsEnum(Enum):
 
     @classmethod
     def from_str(cls, value: str) -> "WeightsEnum":
-        for v in cls:
-            if v._name_ == value or (value == "default" and v.default):
-                return v
+        if value == "default":
+            default_value = getattr(cls, "default", None)
+            if default_value is not None:
+                return default_value
+        else:
+            for v in cls:
+                if v._name_ == value:
+                    return v
         raise ValueError(f"Invalid value {value} for enum {cls.__name__}.")
 
     def get_state_dict(self, progress: bool) -> OrderedDict:
--- a/torchvision/prototype/models/_api.py
+++ b/torchvision/prototype/models/_api.py
@@ -60,14 +60,9 @@ class WeightsEnum(Enum):
 
     @classmethod
     def from_str(cls, value: str) -> "WeightsEnum":
-        if value == "default":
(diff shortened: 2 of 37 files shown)


datumbox:
The references were validated by running the following commands. All work as expected:
```
torchrun --nproc_per_node=1 train.py --test-only --weights ResNet50_Weights.ImageNet1K_V2 --model resnet50

python train_quantization.py --device cpu --test-only --weights ResNet50_QuantizedWeights.ImageNet1K_FBGEMM_V2 --model resnet50

gpurun torchrun --nproc_per_node=1 train.py --test-only --weights SSDLite320_MobileNet_V3_Large_Weights.Coco_V1 --model ssdlite320_mobilenet_v3_large

gpurun torchrun --nproc_per_node=1 train.py  --dataset coco --model lraspp_mobilenet_v3_large --test-only --weights LRASPP_MobileNet_V3_Large_Weights.CocoWithVocLabels_V1

torchrun --nproc_per_node=1 train.py --data-path /datasets01/kinetics/070618/ --train-dir=val_avi-480p --val-dir=val_avi-480p --batch-size=64 --sync-bn --test-only --weights R2Plus1D_18_Weights.Kinetics400_V1 --cache-dataset
```

test/test_prototype_models.py
```diff
@@ -65,10 +76,9 @@ def test_get_weight(model_fn, name, weight):
     + TM.get_models_from_module(models.video),
 )
 def test_naming_conventions(model_fn):
-    model_name = model_fn.__name__
-    module = _get_parent_module(model_fn)
-    weights_name = "_QuantizedWeights" if module.__name__.split(".")[-1] == "quantization" else "_Weights"
-    assert model_name in set(x.replace(weights_name, "").lower() for x in module.__dict__ if x.endswith(weights_name))
+    weights_enum = _get_model_weights(model_fn)
+    assert weights_enum is not None
+    assert len(weights_enum) == 0 or hasattr(weights_enum, "default")
```
NicolasHug, line 81: It's not obvious at first sight what this line is testing?

torchvision/prototype/models/_api.py
```diff
@@ -59,9 +60,14 @@ def verify(cls, obj: Any) -> Any:
...
-            if v._name_ == value or (value == "default" and v.default):
-                return v
+        if value == "default":
+            default_value = getattr(cls, "default", None)
+            if default_value is not None:
+                return default_value
+        else:
+            for v in cls:
+                if v._name_ == value:
+                    return v
```
NicolasHug, line 70: I think we can simplify this with something like

```suggestion
        for v in cls.__members__:  # __members__ will include the "default" field which is an alias
            if field == value:
                return v
```

torchvision/prototype/models/_api.py
```diff
@@ -78,41 +84,37 @@ def __getattr__(self, name):
         return super().__getattr__(name)
 
 
-def get_weight(fn: Callable, weight_name: str) -> WeightsEnum:
+def get_weight(name: str) -> WeightsEnum:
     """
-    Gets the weight enum of a specific model builder method and weight name combination.
+    Gets the weight enum value by its full name. Example: ``ResNet50_Weights.ImageNet1K_V1``
```
NicolasHug, line 89:
```suggestion
    Gets the weight enum value by its full name. Example: ``"ResNet50_Weights.ImageNet1K_V1"``
```

(Although I would remove the backticks since we already have the string delimiters)

torchvision/prototype/models/_api.py
```diff
@@ -78,41 +84,37 @@ def __getattr__(self, name):
...
-    sig = signature(fn)
-    if "weights" not in sig.parameters:
-        raise ValueError("The method is missing the 'weights' parameter.")
+    try:
+        enum_name, value_name = name.split(".")
+    except ValueError:
+        raise ValueError(f"Invalid weight name provided: '{name}'.")
+
+    base_module_name = ".".join(sys.modules[__name__].__name__.split(".")[:-1])
+    base_module = importlib.import_module(base_module_name)
```
NicolasHug, line 103: Not a fan of importing stuff programmatically (it can have strange side effects), but we agreed that the underlying implementation could eventually change once we have a proper registration mechanism, so that's fine for now

torchvision/prototype/models/_api.py
```diff
@@ -78,41 +84,37 @@ def __getattr__(self, name):
...
-                break
+    for m in model_modules:
+        potential_class = m.__dict__.get(enum_name, None)
+        if potential_class is not None and issubclass(potential_class, WeightsEnum):
+            weights_enum = potential_class
+            break
 
     if weights_enum is None:
         raise ValueError(
             "The weight class for the specific method couldn't be retrieved. Make sure the typing info is correct."
```
NicolasHug, line 117: I think the message needs to be changed since we don't rely on typing info anymore

torchvision/prototype/models/_api.py
```diff
@@ -78,41 +84,37 @@ def __getattr__(self, name):
         return super().__getattr__(name)
 
 
-def get_weight(fn: Callable, weight_name: str) -> WeightsEnum:
+def get_weight(name: str) -> WeightsEnum:
```
NicolasHug, line 87: I'm wondering: shoudn't this return a `Weight` instance, instead of `WeightsEnum`?
Same for `from_str` (and in the return section below).

NicolasHug (approved):
Thanks @datumbox , I made some minor comments and a question, but this LGTM

test/test_prototype_models.py
datumbox: The first thing I test here is that every model builder has a WeightsEnum named like `capitalized(model_builder) + "_Weights"`. The capitalization used on the class is not the same as on the builder, so I check on the lower-case string. Effectively I am ensuring that all of the prototypes use the same naming conventions. 

The second test checks that all WeightsEnums have defined a default weight. The only ones allowed not to define any is those enums that don't have any pretrained weights (we got a couple of cases).

The fact that you have questions here is a sign that the unit-test is not readable and easily understood (key rule for unit-tests). As I don't plan to move these tests in the main area, these were checks to simplify my development so if you are OK, I don't what to spent time improving them.

torchvision/prototype/models/_api.py
datumbox: I agree it can be simplified with:
```python
        for k, v in cls.__members__.items():
            if k == value:
                return v
```

test/test_prototype_models.py
NicolasHug: > The only ones allowed not to define any is those enums that don't have any pretrained weights (we got a couple of cases).

This was the bit that I was missing, the rest is clear - no need to change anything

> The first thing I test here is that every model builder has a WeightsEnum named like capitalized(model_builder) + "_Weights". The capitalization used on the class is not the same as on the builder, so I check on the lower-case string. Effectively I am ensuring that all of the prototypes use the same naming conventions

That's a good test to have IMO, but I think it was removed lol

torchvision/prototype/models/_api.py
datumbox: I think we need to return a `WeightsEnum` value. That is the value of an Enum which maintains information about the class it comes from (`ResNet50_Weights`). Returning a `Weights` loses the information necessary to validate that the right type of weights were passed to the method.

Issue #4955, opened 2021-11-17
Add `_log_api_usage_once()` calls on remaining endpoints


datumbox:
### 🚀 The feature

Add `_log_api_usage_once()` calls to:
- [x] The constructors of all Segmentation model constructors. The implementation should be identical to what was done for the other models. See #4735
- [ ] All the Transforms. For those that have [functional equivalent](https://github.com/pytorch/vision/blob/main/torchvision/transforms/functional.py) the call should be placed directly in their method. For those that don't have a functional equivalent (for example [Compose](https://github.com/pytorch/vision/blob/main/torchvision/transforms/transforms.py#L61)), the call should be placed in the class constructor. The approach is identical to what was done for the Operators. See #4799
- [ ] To the public python IO methods at `torchvision.io.image` and `torchvision.io.video` and to the `VideoReader` class constructor. On the python side these are the following endpoints: [1](https://github.com/pytorch/vision/blob/29e0f66a23549ce41e59e326d953cb46da31f922/torchvision/io/__init__.py#L18-L27), [2](https://github.com/pytorch/vision/blob/29e0f66a23549ce41e59e326d953cb46da31f922/torchvision/io/__init__.py#L30-L32), [3](https://github.com/pytorch/vision/blob/29e0f66a23549ce41e59e326d953cb46da31f922/torchvision/io/__init__.py#L108). It might be  also worth covering C++ endpoints that might be called directly (without Python) such as [4](https://github.com/pytorch/vision/blob/4ccef06c66020817fe4966cbd4360946cf2f736e/torchvision/csrc/io/image/image.cpp#L18-L25), [5](https://github.com/pytorch/vision/blob/7947fc8fb38b1d3a2aca03f22a2e6a3caa63f2a0/torchvision/csrc/io/video_reader/video_reader.cpp#L665-L668), [6](https://github.com/pytorch/vision/blob/4ccef06c66020817fe4966cbd4360946cf2f736e/torchvision/csrc/io/video/video.cpp#L159).

cc @vfdev-5 @pmeier

PR #5007, opened 2021-11-29
Add api usage log to transforms

kazhang:
* Call API usage log for every functional transforms.
* Call API usage log for all random transforms (e.g. `RandomCrop`), even though crop is logged in functional transforms, so that we could distinguish whether the `crop` is called from a random transform or used directly;
* transforms that are strictly mapped to a functional transforms are skipped. For example: `ToTensor`
partially-close: #5052
partially-close: #4955

cc @vfdev-5 @datumbox


kazhang:
Currently we can't log `Compose` because it breaks this test
[`test/test_transforms_tensor.py - test_compose](https://app.circleci.com/pipelines/github/pytorch/vision/12991/workflows/2fe32265-9535-405c-b216-aefe8fb1fa5f/jobs/1034989), which asserts `Compose` is not scriptable, however, the torchscript type check error suppresses the origin runtime error.

IMO we could skip `Compose` because the wrapped transforms will be logged anyway. If we do want to log it, we could split `_log_api_usage_once` into `_log_api_usage_once(api: str)` and `_log_instantiation_usage_once(obj: object)` to make torchscript happy.

datumbox:
@kazhang @fmassa The policy on whether the call should be added on a constructor VS functional part changes from area to area. On the operators, I believe Francisco opted for adding the call only once, carefully adding it mainly on functional and secondary on constructors (only for those Ops with no functional component). On the other hand Kai proposes to add it to all constructor and functions but excludes those that are too simple (like `ToTensor`). I think that this change of policy on different parts of the code can be confusing. Shall we align it everywhere? Personally I would favour a simple approach were we add the calls in every constructor and method call we want to log without any exceptions. Thoughts?

kazhang:
> Personally I would favour a simple approach were we add the calls in every constructor and method call we want to log without any exceptions. Thoughts?

Agree with this simple approach, log API is only called once and pretty lightweight. We could further dedup in the downstream if needed.
@fmassa WDYT?

torchvision/transforms/functional.py
```diff
@@ -68,7 +68,7 @@ def get_image_size(img: Tensor) -> List[int]:
     Returns:
         List[int]: The image size.
     """
-    _log_api_usage_once("torchvision.transforms.functional.get_image_size")
+    _log_api_usage_once("transforms", "get_image_size")
```
datumbox, line 71: @kazhang If I understand correctly, this will record the method as `torchvision.transforms.get_image_size`. Unfortunately this method doesn't actually exist because the `torchvision.transforms.functional` methods are not imported directly to `torchvision.transforms`.

Your previous modifications made sense because the recorded methods actually existed. For example, the call of `torchvision.ops.nms` is valid because that's an alternative path to access the method despite being defined under `torchvision.ops.boxes.nms`.

I'm afraid this can lead to confusion. @NicolasHug any thoughts on this?

NicolasHug: yes, if we can, it'd be preferable to follow the exact path

torchvision/transforms/transforms.py
```diff
@@ -87,6 +88,8 @@ class Compose:
     """
 
     def __init__(self, transforms):
+        if not torch.jit.is_scripting() and not torch.jit.is_tracing():
```
kazhang, line 91: The `__init__` of `Compose` is compiled by JIT in some cases.

torchvision/transforms/transforms.py
```diff
@@ -152,6 +156,7 @@ def __call__(self, pic):
         Returns:
             Tensor: Converted image.
         """
+        _log_api_usage_once(self)
```
datumbox, line 159: Can you create a constructor and move this there? It's not supposed to be on the __call__ or forward methods.

torchvision/transforms/transforms.py
```diff
@@ -128,6 +131,7 @@ def __call__(self, pic):
...
```
datumbox, line 134: Same here.

datumbox (commented):
Thanks @kazhang, overall looks good. Just one comment, let me know what you think.

datumbox (approved):
LGTM, thanks @kazhang

PR #5008, opened 2021-11-29
add api usage log for io

kazhang:
Log IO api usages.
Partially-close: #4955


datumbox (commented):
Thanks for the PR @kazhang.

The added logging looks good. Some of the calls (for example `read_image()`) will generate additional logging as they call other public methods but I don't think this is an issue or something we can avoid.

There is at least one more logging needed at the constructor of `VideoReader`  defined [here](https://github.com/pytorch/vision/blob/main/torchvision/io/__init__.py#L108). 

Finally, does it make sense to consider at this point adding the logging calls on the C++ side so that we can include cases where the Python interface is not used? I think that has been discussed offline at one point.

torchvision/csrc/io/image/cpu/decode_jpeg.cpp
```diff
@@ -70,6 +70,7 @@ static void torch_jpeg_set_source_mgr(
 } // namespace
 
 torch::Tensor decode_jpeg(const torch::Tensor& data, ImageReadMode mode) {
+  C10_LOG_API_USAGE_ONCE("torchvision.io.decode_jpeg");
```
datumbox, line 73: Do we want to use the same log message as Python's `decode_jpeg` here? Or we want to have a separate distinct message?

torchvision/csrc/io/image/cpu/decode_png.cpp
```diff
@@ -23,6 +23,7 @@ torch::Tensor decode_png(
     const torch::Tensor& data,
     ImageReadMode mode,
     bool allow_16_bits) {
+  C10_LOG_API_USAGE_ONCE("torchvision.io.decode_png");
```
datumbox, line 26: Same here and in all other places where the string on the C++ side matches the one on Python.

torchvision/csrc/io/video_reader/video_reader.cpp
```diff
@@ -653,10 +655,12 @@ torch::List<torch::Tensor> read_video_from_file(
 }
 
 torch::List<torch::Tensor> probe_video_from_memory(torch::Tensor input_video) {
+  C10_LOG_API_USAGE_ONCE("torchvision.io.probe_video_from_memory");
```
datumbox, line 658: @prabhat00155 Do we want to log this? Is this considered part of public API?

datumbox (approved):
I have a few questions up for discussion, other than that the changes look good to me.

torchvision/csrc/io/video_reader/video_reader.cpp
prabhat00155: Yeah, see this: https://github.com/pytorch/vision/blob/01ffb3ae9f9cc3a445bcaabf787251a05e67c7a7/torchvision/csrc/io/video_reader/video_reader.cpp#L667

prabhat00155 (approved):
Thanks @kazhang!

torchvision/csrc/io/image/cpu/decode_jpeg.cpp
kazhang: It depends on whether we're particular interested in events from C++ API calls. Our logging system currently couldn't tell whether it's logged in C++ or Python.

I can add "_cpp" suffix here (e.g. `torchvision.io.decode_jpeg_cpp`) for now, we could always further process in log pipeline.

datumbox (changes requested):
@kazhang It seems that something is wrong with the PR: 78 modified files and 3.1k additions/1.6k deletions.

I haven't checked which commit caused this but perhaps its worth force pushing to undo or starting a new one. Marking as "request changes" to avoid accidental merges while the PR is in this state.

kazhang:
@datumbox I created a new PR: #5038

PR #5012, opened 2021-11-30
Add bias parameter to ConvNormActivation

NicolasHug:
This PR adds a `bias` parameter to `ConvNormActivation`. The current behaviour in `main` is to add a bias if there's no normalization layer, but in some cases like in RAFT, we need the biases even though there *is* a norm layer. For details, see:  https://github.com/NicolasHug/vision/blob/5f58ec164f6cce45c155815ab5f845d4773fad94/torchvision/models/video/_raft/raft.py#L15

### DIFF
--- a/torchvision/ops/misc.py
+++ b/torchvision/ops/misc.py
@@ -116,7 +116,7 @@ class ConvNormActivation(torch.nn.Sequential):
         activation_layer (Callable[..., torch.nn.Module], optinal): Activation function which will be stacked on top of the normalization layer (if not None), otherwise on top of the conv layer. If ``None`` this layer wont be used. Default: ``torch.nn.ReLU``
         dilation (int): Spacing between kernel elements. Default: 1
         inplace (bool): Parameter for the activation layer, which can optionally do the operation in-place. Default ``True``
-        bias (bool): Whether to use bias in the ``norm_layer``. By default, biases are included if ``norm_layer is None``.
+        bias (bool, optional): Whether to use bias in the convolution layer. By default, biases are included if ``norm_layer is None``.
 
     """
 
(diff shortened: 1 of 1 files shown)


datumbox (approved):
LGTM, just a minor typo and a missing `optional` on the documentation.

torchvision/ops/misc.py
```diff
@@ -131,9 +132,12 @@ def __init__(
         activation_layer: Optional[Callable[..., torch.nn.Module]] = torch.nn.ReLU,
         dilation: int = 1,
         inplace: bool = True,
+        bias: Optional[bool] = None,
```
prabhat00155, line 135: bias is a bool, shouldn't the default be False?

torchvision/ops/misc.py
```diff
@@ -131,9 +132,12 @@ def __init__(
...
     ) -> None:
         if padding is None:
             padding = (kernel_size - 1) // 2 * dilation
+        if bias is None:
```
prabhat00155, line 139: These two statements could be removed by adding
```
bias=bias or norm_layer is None
```
on line 150(given bias is bool and defaults to False).

prabhat00155 (commented):
Thanks @NicolasHug, I have a couple of comments.

torchvision/ops/misc.py
NicolasHug: We want to keep BC so we can't have None be the default. Previously, bias was set to `norm_layer is not None`.

torchvision/ops/misc.py
NicolasHug: I don't think this is equivalent: This would override bias whenever bias is False. We only want to override bias when it's None.

prabhat00155: I see, my understanding was wrong. I thought if bias is False we default back to `norm_layer is None`

PR #5013, opened 2021-11-30
[FBcode->GH] Add import to fix linter error

prabhat00155:
https://github.com/pytorch/vision/pull/4973 introduced a linter error internally(D32694312):
```
Linter (FLAKE8)raised a F821 error online 236-
undefined name 'requests'
```

### DIFF
--- a/torchvision/datasets/utils.py
+++ b/torchvision/datasets/utils.py
@@ -15,6 +15,7 @@ import zipfile
 from typing import Any, Callable, List, Iterable, Optional, TypeVar, Dict, IO, Tuple, Iterator
 from urllib.parse import urlparse
 
+import requests
 import torch
 from torch.utils.model_zoo import tqdm
 
--- a/torchvision/datasets/utils.py
+++ b/torchvision/datasets/utils.py
@@ -15,7 +15,6 @@ import zipfile
 from typing import Any, Callable, List, Iterable, Optional, TypeVar, Dict, IO, Tuple, Iterator
 from urllib.parse import urlparse
 
-import requests
 import torch
 from torch.utils.model_zoo import tqdm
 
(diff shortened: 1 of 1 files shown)


prabhat00155:
The import fails in our tests. Should I remove the type annotation?

datumbox (changes requested):
I know this a FBcode=>GH sync and that eventually we will have to merge, just blocking to ensure we discuss it a bit.

I think the right way to address this would be to delete `_get_confirm_token()` method which seems not being used. The `requests` package was recently removed and I bet we missed clearing this method.

@pmeier could you please confirm?

If that's the case then we can merge the PR to sync internal/external and the follow up with one that removes the import and the private method completely.

prabhat00155:
> I think the right way to address this would be to delete `_get_confirm_token()` method which seems not being used. The `requests` package was recently removed and I bet we missed clearing this method.

Yeah, I don't see _get_confirm_token() getting used anywhere; could well be removed.

pmeier:
> I think the right way to address this would be to delete `_get_confirm_token()` method which seems not being used. The `requests` package was recently removed and I bet we missed clearing this method.

Exactly. I'll send a PR.

prabhat00155:
> > I think the right way to address this would be to delete `_get_confirm_token()` method which seems not being used. The `requests` package was recently removed and I bet we missed clearing this method.
> 
> Exactly. I'll send a PR.

@pmeier Don't send a PR. I'll update this PR after making the changes internally.

datumbox:
@prabhat00155 Updating this PR means we will need to export it via the script to sync it again. Because this is tagged as FBcode->GH it will be automatically excluded. So usually what we do (because it's the easiest way) is merge the original PR as-is and then fix the issue on a follow up PR which can easily be synced to FBcode. Would you be OK doing this?

prabhat00155:
> @prabhat00155 Updating this PR means we will need to export it via the script to sync it again. Because this is tagged as FBcode->GH it will be automatically excluded. So usually what we do (because it's the easiest way) is merge the original PR as-is and then fix the issue on a follow up PR which can easily be synced to FBcode. Would you be OK doing this?

I have already made this change in the diff to fix the error. So, we won't need to sync this change again, hence the `FBcode->GH` tag.

PR #5014, opened 2021-11-30
[FBcode->GH] remove unused requests functionality

pmeier:
Addresses https://github.com/pytorch/vision/pull/5013#pullrequestreview-819374936 and supersedes #5013.

cc @pmeier

### DIFF
--- a/torchvision/datasets/utils.py
+++ b/torchvision/datasets/utils.py
@@ -233,14 +233,6 @@ def download_file_from_google_drive(file_id: str, root: str, filename: Optional[
         _save_response_content(itertools.chain((first_chunk,), content), fpath)
 
 
-def _get_confirm_token(response: "requests.models.Response") -> Optional[str]:  # type: ignore[name-defined]
-    for key, value in response.cookies.items():
-        if key.startswith("download_warning"):
-            return value
-
-    return None
-
-
 def _save_response_content(
     response_gen: Iterator[bytes],
     destination: str,  # type: ignore[name-defined]


pmeier:
It seems, we make the change in #5013. @prabhat00155 feel free to close or merge this what ever is easier for you.

datumbox (commented):
I would prefer to merge this (see https://github.com/pytorch/vision/pull/5013#issuecomment-982994318) but I'm not going to approach to avoid conflicts and accidental merges. I'll leave @prabhat00155 to decide the course of action.

prabhat00155 (approved):
I have made these changes internally to fix the linter error. So, we should not sync these changes again. I am happy to merge this PR(after adding `FBcode->GH` tag).

datumbox:
@prabhat00155 I had a look on the internal diffs to understand what you were talking about and why you were saying it was OK to sync the other PR. What I currently understand happened was:
1. During your weekly sync one of the diffs had issues on FBcode and you had to make changes in-place to make it in.
2. These changes where exported to Github by creating a new PR. 

Your point was that either merging #5013 or #5014, it would be the same. And you are right.

What I understood was different. I thought someone else internally submitted a change on FBcode and you were trying to sync it on Github. Since we identified an issue with the fix, the usual course of action would be to merge as-is #5013 and open a new PR (say #5014) with the change which in turn would have to be synced back to FBcode.

I'm sorry for the misunderstanding, I confirm that the course of action you took is the correct and that this PR shouldn't be synced back to FBcode.

PR #5018, opened 2021-12-01
remove vanilla tensors from prototype datasets samples

pmeier:
As discussed offline with @fmassa and @NicolasHug, for now we opt to always return a plain `Feature` instead of a vanilla `torch.Tensor`. This let's us keep BC with the transforms that interpret vanilla tensors as images. This just fixes the return types for COCO. I'll send a more comprehensive follow-up PR.

cc @pmeier @bjuncek


NicolasHug (approved):
Stamping to unlock!

Issue #2439, opened 2020-07-09
[RFC] Hardware-accelerated video decoding


bryandeng:
## 🚀 Feature
Hardware-accelerated video decoding

## Motivation
Now that torchscriptable transforms natively supporting GPU have landed, hardware-accelerated video decoding may further help relieve the IO bottleneck commonly seen in large-scale video deep learning tasks.

## Pitch
This functionality is likely to be built upon FFmpeg's hardware acceleration APIs, since FFmpeg is already in use and it's easier to support multiple hardware platforms and platform APIs in this way.

## Alternatives
[Decord](https://github.com/dmlc/decord) and [NVIDIA VPF](https://github.com/NVIDIA/VideoProcessingFramework) are both PyTorch-friendly video IO libraries which support (NVIDIA only) hardware-accelerated video decoding to some extent.
NVIDIA VPF is built upon NVIDIA Video Codec SDK directly without FFmpeg.

## Additional context
https://trac.ffmpeg.org/wiki/HWAccelIntro
https://developer.nvidia.com/video-encode-and-decode-gpu-support-matrix-new

fmassa:
Hi,

Thanks for opening the issue!

I agree that video decoding on the GPU would be a nice functionality to have, although it's not in the near-term plan for now (maybe in 6 months?).

About seeking on videos, we are preparing a revamp of the video reading abstractions which will be more generic and allow for more flexibility while performing video decoding. cc @bjuncek 

For video transformations on the GPU, note that we are making all transforms in torchvision to be torchscriptable and work with Tensors, so that they will natively support GPU if needed.

bryandeng:
Updated description after the release of torchscriptable transforms.

bryandeng:
We may first
- [ ] discuss which hardware platforms to support
- [ ] find out the constraints on input video formats and output pixel formats, etc.

JuanFMontesinos:
Hi,
I wrote this small library which wraps ffmeg and does exactly that.
https://github.com/JuanFMontesinos/PyNVIdeoReader/tree/main
For those who need it, it can work in the meanwhile.

dwrodri:
Pinging this thread to check for interest. Nvidia has been working on writing their own bridge between their GPU's encoders/decoders and popular machine learning libraries ([link here](https://github.com/nvidia/DALI)), but it's still in beta. Is there any interest in making this a feature in PyTorch? 

[...]

PR #5019, opened 2021-12-01
Add video GPU decoder

prabhat00155:
This PR adds support for GPU decoding in torchvision’s video reading API.
Resolves https://github.com/pytorch/vision/issues/2439 and partly addresses https://github.com/pytorch/vision/issues/4392.

This is the initial version of GPU video decoder. This can be called like this:
```
reader = torchvision.io.VideoReader(file_name, device="cuda:0")
for frame in reader:
  print(frames['data'])
```
The result after performing GPU decoding can be returned in the form of a CUDA tensor(when using use_device_frame=True) or a CPU tensor(use_device_frame=False). When `use_device_frame=True`, `nv12` is the only supported output format, when using `use_device_frame=False`, `nv12` and `yuv420` are the supported output formats.

Work items to extend this further can be found [here](https://github.com/pytorch/vision/issues/4392).


torchvision/io/__init__.py
```diff
@@ -32,6 +32,10 @@
     write_video,
 )
 
+from .gpu_decoder import (
```
bjuncek, line 35: Shouldn't this be somehow guarded (if it exists, if there is a cuda device, etc...)

torchvision/io/gpu_decoder.py
```diff
@@ -0,0 +1,22 @@
+import torch
+
+
+class GPUDecoder:
+    def __init__(self, src_file: str):
+        torch.ops.load_library('build/lib.linux-x86_64-3.8/torchvision/Decoder.so')
```
bjuncek, line 6: is it possible to register it through tochbind?

prabhat00155: What do you mean? Here, the library that was built needs to be loaded before they could be used. Similar to https://github.com/pytorch/vision/blob/ba299e8f52737448f1cbb9fffd25bf3d6e7ceef7/torchvision/io/image.py#L10.

torchvision/io/__init__.py
prabhat00155: That's a good point. We could perhaps do that in the GPUDecoder class.

torchvision/io/_video_opt.py
```diff
@@ -14,6 +14,13 @@
 except (ImportError, OSError):
     _HAS_VIDEO_OPT = False
 
+try:
+    #_load_library("Decoder")
```
prabhat00155, line 18: This doesn't seem to work, hence hardcoded the path below.

fmassa: Discussed offline, we just need to `cd ..` to avoid picking `torchvision` from the current folder (which doesn't contain the `.so` files).

In general, I would recommend installing locally with `python setup.py develop`, so that the `.so` files are available and you won't face those problems in the future

torchvision/csrc/io/decoder/gpu/demuxer.h
```diff
@@ -0,0 +1,205 @@
+extern "C" {
+#include <libavcodec/avcodec.h>
+#include <libavcodec/bsf.h>
+#include <libavformat/avformat.h>
+#include <libavformat/avio.h>
+}
+
+inline bool check(int ret, int line) {
+  if (ret < 0) {
+    printf("Error %d at line %d in demuxer.h.\n", ret, line);
```
prabhat00155, line 10: Will replace this with runtime_error().

torchvision/csrc/io/decoder/gpu/demuxer.h
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    if (bMp4H264 || bMp4HEVC) {
+      if (pktFiltered.data) {
+        av_packet_unref(&pktFiltered);
+      }
+      check_for_errors(av_bsf_send_packet(bsfCtx, &pkt));
+      check_for_errors(av_bsf_receive_packet(bsfCtx, &pktFiltered));
+      *video = pktFiltered.data;
+      *videoBytes = pktFiltered.size;
```
fmassa, line 150: Let's double-check that we don't have leaks in here. I think it should be fine, but just checking

fmassa (commented):
Thanks for the PR @prabhat00155 !

I've done an initial pass but didn't have the chance to finish everything.

I think there might be some memory leaks in the implementation, and I have a few suggestions.

Let me know what you think

torchvision/csrc/io/decoder/gpu/demuxer.h
```diff
@@ -0,0 +1,205 @@
...
+    } else {
+      if (bMp4MPEG4 && (frameCount == 0)) {
+        int extraDataSize =
+            fmtCtx->streams[iVideoStream]->codecpar->extradata_size;
+
+        if (extraDataSize > 0) {
+          dataWithHeader = (uint8_t*)av_malloc(
+              extraDataSize + pkt.size - 3 * sizeof(uint8_t));
```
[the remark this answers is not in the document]
prabhat00155: This is getting freed in the destructor.

torchvision/csrc/io/decoder/gpu/demuxer.h
prabhat00155: I ran cuda-memcheck and didn't get any memory leaks.

torchvision/csrc/io/decoder/gpu/decoder.cpp
```diff
@@ -0,0 +1,428 @@
...
+  if (!data || size == 0) {
+    pkt.flags |= CUVID_PKT_ENDOFSTREAM;
+  }
+  check_for_cuda_errors(cuvidParseVideoData(parser, &pkt), __LINE__);
+  cuvidStream = 0;
+}
+
+/* Fetch a decoded frame and remove it from the queue.
+ */
+uint8_t* Decoder::fetch_frame() {
```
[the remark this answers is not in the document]
prabhat00155: Well this function could return nullptr in some cases, when decodedFrames queue is empty(at the start of decoding). This avoids the need to creating unnecessary tensors. Also, I have kept it this way to avoid changing these functions when we support RGB or extend it to other output types.

torchvision/csrc/io/decoder/gpu/demuxer.h
fmassa: This code runs on the CPU. I might be wrong but I though that cuda memcheck was for checking CUDA memory?

torchvision/csrc/io/decoder/gpu/demuxer.h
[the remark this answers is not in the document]
fmassa: Note that the `malloc` happens at every `demux` call, while the destructor is only called once. So I still believe there might be a memory leak here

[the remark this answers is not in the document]
prabhat00155: If you look at the `if` condition around this block, this statement would at max only get called once at the beginning when `frameCount == 0`, hence needing to be freed only once.

torchvision/csrc/io/decoder/gpu/demuxer.h
prabhat00155: You are right, I'll run valgrind to make sure there is no CPU memory leak.

setup.py
```diff
@@ -427,6 +427,46 @@ def get_extensions():
             )
         )
 
+    # Locating video codec
+    # Should be included in CUDA_HOME for CUDA >= 10.1, which is the minimum version we have in the CI
+    video_codec_found = (
+        extension is CUDAExtension
+        and CUDA_HOME is not None
+        and os.path.exists("/usr/local/include/cuviddec.h")
```
prabhat00155, line 435: Change to default to TORCHVISION_INCLUDE.

setup.py
```diff
@@ -427,6 +427,46 @@ def get_extensions():
...
+        and os.path.exists("/usr/local/include/nvcuvid.h")
+    )
```
fmassa, line 437: Can you instead use the environment variable `TORCHVISION_INCLUDE` (and `TORCHVISION_LIBRARY`) here instead?

setup.py
```diff
@@ -427,6 +427,49 @@ def get_extensions():
...
+        gpu_decoder_path = os.path.join(extensions_dir, "io", "decoder", "gpu")
+        gpu_decoder_src = glob.glob(os.path.join(gpu_decoder_path, "*.cpp"))
+        cuda_libs = os.path.join(CUDA_HOME, "lib64")
+        cuda_inc = os.path.join(CUDA_HOME, "include")
+
+        ext_modules.append(
+            extension(
+                "torchvision.Decoder",
+                gpu_decoder_src,
+                include_dirs=include_dirs + [gpu_decoder_path] + [cuda_inc],
```
fmassa, line 454: Can you add `ffmpeg_include_dir` here as well?

torchvision/csrc/io/decoder/gpu/README.rst
```diff
@@ -0,0 +1,20 @@
...
+* Download the latest `nvidia-video-codec-sdk <https://developer.nvidia.com/nvidia-video-codec-sdk/download>`_
+* Extract the zipped file and copy the header files and libraries.
+
+.. code:: bash
+
+sudo cp Interface/* /usr/local/include/
+sudo cp Lib/linux/stubs/x86_64/libnv* /usr/local/lib/
+
+* Install ffmpeg and make sure ffmpeg headers and libraries are present under /usr/local/include and /usr/local/lib respectively.
+* Set CUDA_HOME environment variable to the cuda root directory.
```
fmassa, line 15: You'll need to update those instructions to take `TORCHVISION_INCLUDE` and `TORCHVISION_LIBRARY` into account

torchvision/csrc/io/decoder/gpu/README.rst
```diff
@@ -0,0 +1,20 @@
+GPU Decoder
+===========
+
+GPU decoder depends on ffmpeg for demuxing, uses NVDECODE APIs from the nvidia-video-codec sdk and uses cuda for processing on gpu. In order to use this, please follow the following steps:
```
fmassa, line 4: On the ffmpeg side, you'll probably need to make sure the the `bsf.h` file is available before trying to compile.

It turns out that `ffmpeg` installed through the default channel (anaconda), which is 4.2.2, for some reason doesn't ship with `bsf.h`.

torchvision/csrc/io/decoder/gpu/demuxer.h
```diff
@@ -0,0 +1,241 @@
...
+  AVCodecID eVideoCodec;
+  uint8_t* dataWithHeader = NULL;
+  bool bMp4H264, bMp4HEVC, bMp4MPEG4;
+  unsigned int frameCount = 0;
+  int iVideoStream;
+  int64_t userTimeScale = 0;
+  double timeBase = 0.0;
+
+ public:
+  Demuxer(const char* filePath, int64_t timeScale = 1000 /*Hz*/) {
```
fmassa, line 22: For the (near) future: we will need to add support from reading from memory.

Also, is this `timeScale` used anywhere? Or do we plan to use it anywhere?

torchvision/csrc/io/decoder/gpu/gpu_decoder.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+/* Fetch a decoded frame tensor after demuxing and decoding.
+ */
+torch::Tensor GPUDecoder::decode() {
+  torch::Tensor frameTensor;
+  unsigned long videoBytes = 0;
+  uint8_t* video = nullptr;
+  at::cuda::CUDAGuard device_guard(device);
+  auto options = torch::TensorOptions().dtype(torch::kU8).device(torch::kCUDA);
+  torch::Tensor frame = torch::zeros({0}, options);
```
fmassa, line 32: nit: I believe you can just do
```c
torch::Tensor frame;
```
now that `decoder.fetch_frame()` returns a tensor of the right dtype and device

torchvision/io/__init__.py
```diff
@@ -6,6 +6,7 @@
 from ._video_opt import (
     Timebase,
     VideoMetaData,
+    _HAS_VIDEO_DECODER,
```
fmassa, line 9: nit: this is not a great name for this attribute.

I would maybe name it as `_HAS_GPU_VIDEO_DECODER` instead.

But this can be changed in a follow-up PR.

torchvision/io/__init__.py
```diff
@@ -104,17 +105,31 @@ class VideoReader:
...
+    def __init__(self, path: str, stream: str = "video", num_threads: int = 0, device: str = "cpu") -> None:
         _log_api_usage_once(self)
+        self.is_cuda = False
+        device = torch.device(device)
+        if device.type == "cuda":
+            if not _HAS_VIDEO_DECODER:
+                raise RuntimeError("Not compiled with GPU decoder support.")
+            self.is_cuda = True
+            if device.index is None:
+                raise RuntimeError("Invalid cuda device!")
```
fmassa, line 122: I think this is a bit too restrictive, as it's very common to have a `device` which is just `'cuda'` (and not `'cuda:0'`).
If you check the other functions in torchvision (including image GPU decoding), we do support `cuda` just as well as `cuda:0`.

What I would do is to let `GPUDecoder` accept a `Device` object directly, so that just doing
```c
  at::cuda::CUDAGuard device_guard(device);
```
at the beginning of the constructor would be enough, and we would be able to retrieve the `device_guard.current_device().index()` (or maybe just simply `device.index()`)

torchvision/io/_video_opt.py
```diff
@@ -14,6 +14,12 @@
 except (ImportError, OSError):
     _HAS_VIDEO_OPT = False
 
+try:
+    _load_library("Decoder")
+    _HAS_VIDEO_DECODER = True
+except (ImportError, OSError):
+    _HAS_VIDEO_DECODER = False
+
```
fmassa, line 22: nit: I would probably have moved this to the file where `GPUDecoder` is actually used.

I would actually in the future move all the code related to `VideoReader` to a separate file, and not keep it in `__init__.py`.

fmassa (approved):
Thanks for all your work Prabhat!

I've left some more comments, but I think all of them can be addressed in a follow-up comment.

As we discussed in our call earlier, I've also gave a try at implementing the NV12->RGB conversion on the GPU and it avoids the 2 memcopies that we are currently doing (so a single kernel does the reading + conversion).

I'll post it in a branch soon

fmassa:
We need to check if the errors are related or not (maybe not).

Also, in a follow-up PR we will need to add CI tests for those functions. This might be a bit involved, but looks like the NVIDIA-docker has nvdec available in the container

fmassa:
Looks like it's only on Python 3.6, which reached its EOL last week https://endoflife.date/python

So we should probably just remove support for Python 3.6 in torchvision main branch (as long as PyTorch also drops support for it)

torchvision/csrc/io/decoder/gpu/decoder.cpp
```diff
@@ -0,0 +1,417 @@
...
+ */
+int Decoder::handle_picture_decode(CUVIDPICPARAMS* pic_params) {
+  if (!decoder) {
+    TORCH_CHECK(false, "Uninitialised decoder");
+  }
+  pic_num_in_decode_order[pic_params->CurrPicIdx] = decode_pic_count++;
+  check_for_cuda_errors(cuCtxPushCurrent(cu_context), __LINE__, __FILE__);
+  check_for_cuda_errors(
+      cuvidDecodePicture(decoder, pic_params), __LINE__, __FILE__);
+  check_for_cuda_errors(cuCtxPopCurrent(NULL), __LINE__, __FILE__);
```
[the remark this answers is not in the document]
prabhat00155: https://github.com/pytorch/vision/issues/5144

torchvision/csrc/io/decoder/gpu/decoder.cpp
```diff
@@ -0,0 +1,417 @@
...
+      cuvidGetDecodeStatus(decoder, disp_info->picture_index, &decode_status);
+  if (result == CUDA_SUCCESS &&
+      (decode_status.decodeStatus == cuvidDecodeStatus_Error ||
+       decode_status.decodeStatus == cuvidDecodeStatus_Error_Concealed)) {
+    VLOG(1) << "Decode Error occurred for picture "
+            << pic_num_in_decode_order[disp_info->picture_index];
+  }
+
+  auto options = torch::TensorOptions().dtype(torch::kU8).device(torch::kCUDA);
+  torch::Tensor decoded_frame = torch::empty({get_frame_size()}, options);
```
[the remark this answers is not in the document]
prabhat00155: https://github.com/pytorch/vision/issues/5145

torchvision/io/__init__.py
prabhat00155: https://github.com/pytorch/vision/issues/5146

torchvision/csrc/io/decoder/gpu/decoder.cpp
```diff
@@ -0,0 +1,417 @@
...
+      video_output_format = cudaVideoSurfaceFormat_P016;
+    } else if (
+        decode_caps.nOutputFormatMask & (1 << cudaVideoSurfaceFormat_YUV444)) {
+      video_output_format = cudaVideoSurfaceFormat_YUV444;
+    } else if (
+        decode_caps.nOutputFormatMask &
+        (1 << cudaVideoSurfaceFormat_YUV444_16Bit)) {
+      video_output_format = cudaVideoSurfaceFormat_YUV444_16Bit;
+    } else {
+      TORCH_CHECK(false, "No supported output format found");
```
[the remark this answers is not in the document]
prabhat00155: https://github.com/pytorch/vision/issues/5148

torchvision/csrc/io/decoder/gpu/decoder.cpp
```diff
@@ -0,0 +1,417 @@
...
+          ? cudaVideoSurfaceFormat_P016
+          : cudaVideoSurfaceFormat_NV12;
+      break;
+    case cudaVideoChromaFormat_444:
+      video_output_format = video_format->bit_depth_luma_minus8
+          ? cudaVideoSurfaceFormat_YUV444_16Bit
+          : cudaVideoSurfaceFormat_YUV444;
+      break;
+    case cudaVideoChromaFormat_422:
+      video_output_format = cudaVideoSurfaceFormat_NV12;
```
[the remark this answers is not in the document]
prabhat00155: https://github.com/pytorch/vision/issues/5148

torchvision/csrc/io/decoder/gpu/decoder.h
```diff
@@ -0,0 +1,99 @@
+#include <cuda.h>
+#include <cuda_runtime_api.h>
+#include <cuviddec.h>
+#include <nvcuvid.h>
+#include <torch/torch.h>
+#include <cstdint>
+#include <queue>
+
+static auto check_for_cuda_errors =
+    [](CUresult result, int line_num, std::string file_name) {
```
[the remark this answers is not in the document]
prabhat00155: https://github.com/pytorch/vision/issues/5148

torchvision/csrc/io/decoder/gpu/demuxer.h
prabhat00155: https://github.com/pytorch/vision/issues/5148

torchvision/csrc/io/decoder/gpu/gpu_decoder.cpp
prabhat00155: https://github.com/pytorch/vision/issues/5148

torchvision/io/__init__.py
prabhat00155: https://github.com/pytorch/vision/issues/5148

torchvision/io/_video_opt.py
prabhat00155: https://github.com/pytorch/vision/issues/5148

Issue #4644, opened 2021-10-18
RAFT model and training reference


NicolasHug:
I'm opening this issue to keep track of the progress made on the [RAFT](https://github.com/princeton-vl/RAFT) implementation. Implementing RAFT is part of our H2 goals: https://github.com/pytorch/vision/issues/4187

- [x] model architecture (normal and "small" one) https://github.com/pytorch/vision/pull/5022
- [x] new datasets 
  - [x] Flying Chairs https://github.com/pytorch/vision/pull/4860
  - [x] FlyingThings3D https://github.com/pytorch/vision/pull/4858
  - [X] Sintel https://github.com/pytorch/vision/pull/4845
  - [X] KittiFlow https://github.com/pytorch/vision/pull/4845
    - [X] support for 16bits pngs for KittiFlow  https://github.com/pytorch/vision/pull/4657
  - [X] HD1K (maybe) https://github.com/pytorch/vision/pull/4890

- [x] new transforms and augmentations https://github.com/pytorch/vision/pull/5026
- [x] training and evaluation reference for optical flow models https://github.com/pytorch/vision/pull/5027
- [X] pretrained weights (our own weights from training recipe + porting of original weights from paper):
  - C+T: https://github.com/pytorch/vision/pull/5060
  - C+T (small) https://github.com/pytorch/vision/pull/5070 https://github.com/pytorch/vision/pull/5076
  - C+T + Sintel fine-tuning https://github.com/pytorch/vision/pull/5079
  - C+T + Sintel fine-tuning + Kitti fine-tuning https://github.com/pytorch/vision/pull/5081
- [X] Bonus: implement new prototype API https://github.com/pytorch/vision/pull/5043

The support for Kitti would require support for 16bits png, ~which we don't have at the moment.~ EDIT: we do now

cc @pmeier @datumbox

PR #5022, opened 2021-12-02
Add RAFT model for optical flow

NicolasHug:

This PR adds the RAFT model, along with its basic building blocks (feature encoder, correlation block, update block, etc.) and two model builder function: `raft_large()` and `raft_small()`.

The architecture (not the code!) is exactly the same as the original implementation from https://github.com/princeton-vl/RAFT, which will allow us support the original paper's weights if we want to. This architecture differs slightly from what is described in the paper. I have annotated the paper with these differences, hoping this can help the review:
[RAFT- Recurrent All-Pairs Field Transforms forOptical Flow (1).pdf](https://github.com/pytorch/vision/files/7641423/RAFT-.Recurrent.All-Pairs.Field.Transforms.forOptical.Flow.1.pdf)

A summary is this:

image

----

### API

The `RAFT` class accepts **torch.nn.Module instances** as input and offers a low-level API.

```py
RAFT(
        feature_encoder=feature_encoder,
        context_encoder=context_encoder,
        corr_block=corr_block,
        update_block=update_block,
        mask_predictor=mask_predictor,
    )
```
The model builder functions `raft_large()` and `raft_small()` are higher-level and do not require any parameter. They can however take as input the same parameters as the `RAFT` class, so as to override their defaults. E.g.:

```py
model = raft_large(mask_predictor=MyOwnMaskPredictor())  # all the other building blocks will use the default architecture
```

The building blocks like `FeatureEncoder`, `ResidualBlock`,  `UpdateBlock`, etc. are (sort of publicly) available in `torchvision.models.optical_flow.raft`, but are not exposed in `__all__`.

----

Still left TODO, here or in follow-up PRs:

- [x] Add an expect test for both `raft()` and `raft_small()`
- [x] Add better docs
- [ ] Add type annotations (🥲)

In follow up PRs I will also submit the training reference and associated transforms.

I will write review comments below to hightlight important bits, or things where I'm not to sure what the best way is.

cc @datumbox


torchvision/models/optical_flow/__init__.py
```diff
@@ -0,0 +1 @@
+from ._raft.raft import RAFT, raft, raft_small
```
NicolasHug, line 1: We should expose the building blocks as well, stuff like `ResidualBlock`, `FeatureEncoder`, etc.

Should we expose these in a `models.optical_flow.raft` namespace?

I remember @datumbox mentioning a few issues when we have both a module and a function with the same name. Ideally, I'd like to keep the `raft()` name for the function builder, but I'm happy to get your thoughts on this

torchvision/models/optical_flow/_raft/raft.py
```diff
@@ -0,0 +1,538 @@
...
+import torch.nn.functional as F
+from torch.nn.modules.batchnorm import BatchNorm2d
+from torch.nn.modules.instancenorm import InstanceNorm2d
+from torchvision.ops import ConvNormActivation
+
+from .utils import grid_sample, make_coords_grid, upsample_flow
+
+
+class ResidualBlock(nn.Module):
+    # This is pretty similar to resnet.BasicBlock except for one call to relu, and the bias terms
```
NicolasHug, line 12: We can probably merge this with the resnet implementation (same for BottleneckBlock below)... but this might make the API awkward and bloated. For now I'd say it's simpler and safer to keep these implementations separate

torchvision/models/optical_flow/_raft/raft.py
```diff
@@ -0,0 +1,538 @@
...
+        assert len(corr_layers) in (1, 2)
+
+        self.convcorr1 = ConvNormActivation(in_channels_corr, corr_layers[0], norm_layer=None, kernel_size=1)
+        if len(corr_layers) == 2:
+            self.convcorr2 = ConvNormActivation(corr_layers[0], corr_layers[1], norm_layer=None, kernel_size=3)
+        else:
+            self.convcorr2 = None
+
+        self.convflow1 = ConvNormActivation(2, flow_layers[0], norm_layer=None, kernel_size=7)
+        self.convflow2 = ConvNormActivation(flow_layers[0], flow_layers[1], norm_layer=None, kernel_size=3)
```
NicolasHug, line 151: Does it make sense to use a `ConvNormActivation` while passing `norm_layer=None`? No strong opinion from me

torchvision/models/optical_flow/_raft/raft.py
```diff
@@ -0,0 +1,538 @@
...
+    motion_encoder_out_channels,
+    # Recurrent block
+    recurrent_block_hidden_state_size,
+    recurrent_block_kernel_size,
+    recurrent_block_padding,
+    # Flow Head
+    flow_head_hidden_size,
+    # Mask predictor
+    use_mask_predictor,
+    **kwargs,
```
NicolasHug, line 424: kwargs are here to override the `RAFT` class parameters, e.g. to override the entire `feature_encoder`, or the whole `update_block`.

torchvision/models/optical_flow/_raft/raft.py
```diff
@@ -0,0 +1,538 @@
...
+        y = self.convnormrelu3(y)
+
+        if self.downsample is not None:
+            x = self.downsample(x)
+
+        return self.relu(x + y)
+
+
+class FeatureEncoder(nn.Module):
+    def __init__(self, *, block=ResidualBlock, layers=(64, 64, 96, 128, 256), norm_layer=nn.BatchNorm2d):
```
NicolasHug, line 97: I named this `layers` to be sort of consistent with the `ResNet` class. I'm happy to consider other names though

torchvision/models/optical_flow/_raft/raft.py
```diff
@@ -0,0 +1,538 @@
...
+        hx = torch.cat([h, x], dim=1)
+        z = torch.sigmoid(self.convz(hx))
+        r = torch.sigmoid(self.convr(hx))
+        q = torch.tanh(self.convq(torch.cat([r * h, x], dim=1)))
+        h = (1 - z) * h + z * q
+        return h
+
+
+class RecurrentBlock(nn.Module):
+    def __init__(self, *, input_size, hidden_size, kernel_size=((1, 5), (5, 1)), padding=((0, 2), (2, 0))):
```
NicolasHug, line 191: Where possible I tried to write defaults that will correspond to the normal RAFT model (kernel_size, padding). I refrained from doing that for the input and output shapes though. Thoughts?

torchvision/models/optical_flow/_raft/raft.py
```diff
@@ -0,0 +1,538 @@
...
+        motion_features = self.motion_encoder(flow, corr_features)
+        x = torch.cat([context, motion_features], dim=1)
+
+        hidden_state = self.recurrent_block(hidden_state, x)
+        delta_flow = self.flow_head(hidden_state)
+        return hidden_state, delta_flow
+
+
+class MaskPredictor(nn.Module):
+    def __init__(self, *, in_channels, hidden_size, multiplier=0.25):
```
NicolasHug, line 246: I'm keeping the 0.25 default for consistency with the original code, but I'm tempted to set it to 1 to encourage users not to use it

torchvision/models/optical_flow/_raft/raft.py
```diff
@@ -0,0 +1,538 @@
...
+        torch._assert(fmap1.shape[-2:] == (h / 8, w / 8), "The feature encoder should downsample H and W by 8")
+
+        self.corr_block.build_pyramid(fmap1, fmap2)
+
+        context_out = self.context_encoder(image1)
+        torch._assert(context_out.shape[-2:] == (h / 8, w / 8), "The context encoder should downsample H and W by 8")
+
+        # As in the original paper, the actual output of the context encoder is split in 2 parts:
+        # - one part is used to initialize the hidden state of the reccurent units of the update block
+        # - the rest is the "actual" context.
```
NicolasHug, line 367: This is a bit that I did not see mentioned in the paper. I wish we could separate the initialization of the hidden state from the context encoder, but this would likely prevent us from exactly reproduce the original implementation.

torchvision/models/optical_flow/_raft/raft.py
```diff
@@ -0,0 +1,538 @@
...
+        self.mask_predictor = mask_predictor
+
+        if not hasattr(self.update_block, "hidden_state_size"):
+            raise ValueError("The update_block parameter should expose a 'hidden_state_size' attribute.")
+
+    def forward(self, image1, image2, *, num_flow_updates=12):
+
+        batch_size, _, h, w = image1.shape
+        torch._assert((h, w) == image2.shape[-2:], "input images should have the same shape")
+        torch._assert((h % 8 == 0) and (w % 8 == 0), "input image H and W should be divisible by 8")
```
NicolasHug, line 354: This "8" downscaling factor is hard-coded in different places. We should be able to generalize it in future versions.

NicolasHug (commented):
Here are my initial comments @fmassa @datumbox @haooooooqi !

torchvision/models/optical_flow/_raft/raft.py
datumbox: FYI: Agreed. One more reason why we shouldn't expose this on the `__all__`.

torchvision/models/optical_flow/_raft/raft.py
```diff
@@ -0,0 +1,538 @@
...
+        # See note in ResidualBlock for the reason behind bias=True
+        self.convnormrelu = ConvNormActivation(3, layers[0], norm_layer=norm_layer, kernel_size=7, stride=2, bias=True)
+
+        self.layer1 = self._make_2_blocks(block, layers[0], layers[1], norm_layer=norm_layer, first_stride=1)
+        self.layer2 = self._make_2_blocks(block, layers[1], layers[2], norm_layer=norm_layer, first_stride=2)
+        self.layer3 = self._make_2_blocks(block, layers[2], layers[3], norm_layer=norm_layer, first_stride=2)
+
+        self.conv = nn.Conv2d(layers[3], layers[4], kernel_size=1)
+
+        for m in self.modules():
```
datumbox, line 111: Nit: RegNets and ViT started adding these in a private method (`_reset_parameters` or `_init_weights`).

Should we start a convention and align the two implementations? cc @kazhang @sallysyw

datumbox (commented):
@NicolasHug Great work.

I'm reviewing code practices and API, not ML validity as this is something you checked already with @haooooooqi.

I clearly marked my comments as FYI (which are just for discussion), Questions (where I just want more info), NIT (which are non-blocking and could optionally be addressed on follow up PRs) and Blocking (which I think need to be done here).

My Blocking comments are minimal as you can see below and can be addressed easily. After addressing them, ping me to approve but I would wait for the review of @fmassa prior merging.

torchvision/models/optical_flow/_raft/raft.py
fmassa: Let's keep this here as is

torchvision/models/optical_flow/_raft/raft.py
fmassa: I'm ok with this.

torchvision/models/optical_flow/_raft/raft.py
fmassa: We can make everything consistent, but this can probably be tackled as a separate task that addresses this globally in torchvision

torchvision/models/optical_flow/_raft/raft.py
```diff
@@ -0,0 +1,538 @@
...
+        super().__init__()
+        self.convz = nn.Conv2d(hidden_size + input_size, hidden_size, kernel_size=kernel_size, padding=padding)
+        self.convr = nn.Conv2d(hidden_size + input_size, hidden_size, kernel_size=kernel_size, padding=padding)
+        self.convq = nn.Conv2d(hidden_size + input_size, hidden_size, kernel_size=kernel_size, padding=padding)
+
+    def forward(self, h, x):
+        hx = torch.cat([h, x], dim=1)
+        z = torch.sigmoid(self.convz(hx))
+        r = torch.sigmoid(self.convr(hx))
+        q = torch.tanh(self.convq(torch.cat([r * h, x], dim=1)))
```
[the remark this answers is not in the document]
fmassa: We already have `torch.cat` in the implementation, and having to have 2 different `nn.Sigmoid` + a `nn.Tanh` will be annoying.

I believe the new FX-based quantization doesn't have those limitations, and as we are not planning on quantizing this model as of now, I would say to keep this as is. IMO it reads better, and using the functional API for operations that don't hold state was the "recommended" way at some point.

torchvision/models/optical_flow/_raft/raft.py
```diff
@@ -0,0 +1,538 @@
...
+        hidden_state_size = self.update_block.hidden_state_size
+        out_channels_context = context_out.shape[1] - hidden_state_size
+        torch._assert(
+            out_channels_context > 0,
+            f"The context encoder outputs {context_out.shape[1]} channels, but it should have at least "
+            f"hidden_state={hidden_state_size} channels",
+        )
+        hidden_state, context = torch.split(context_out, [hidden_state_size, out_channels_context], dim=1)
+        hidden_state = torch.tanh(hidden_state)
+        context = torch.relu(context)
```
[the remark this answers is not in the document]
fmassa: I still hold to my previous comment, although I have to say that `torch.relu` being visible to the user is probably a mistake (an artifact from how operators were registered in PyTorch). I think we should use instead `torch.nn.functional.relu`.

torchvision/models/optical_flow/_raft/raft.py
```diff
@@ -0,0 +1,538 @@
...
+        feature_encoder=feature_encoder,
+        context_encoder=context_encoder,
+        corr_block=corr_block,
+        update_block=update_block,
+        mask_predictor=mask_predictor,
+        **kwargs,  # not really needed, all params should be consumed by now
+    )
+
+
+def raft(*, weights=None, progress=True, **kwargs):
```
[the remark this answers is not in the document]
NicolasHug: Sounds good, I'll change to pretrained to avoid any risk.

As long as we graduate the new model builders from prototype to main, we should be able to change back `pretrained` into `weights`, right?

torchvision/models/optical_flow/_raft/raft.py
NicolasHug: That makes sense, I created a `_init_weights()` method

torchvision/models/optical_flow/_raft/raft.py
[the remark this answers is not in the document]
datumbox: That is correct. There are also a bunch of stuff we will need to create for this model to introduce it to the prototype builders (such as the presets used for inference) but this can be addressed separately on another PR. Can help you with that.

torchvision/models/optical_flow/__init__.py
NicolasHug: Thanks for the feedback 

> Can you talk about the choice to create a second private submodule _raft? Why is that?

I wrote it like this out of habit but I had no intention of keeping it this way

Instead of renaming `raft()` to `raft_large()`, would it be OK instead to rename the `.py` file to something like `raft_core.py` ? Or `raft_implem.py`?

We would have `raft()`, `raft_small` and `RAFT` available from `torchvision.models.optical_flow` and then we would be able to access the other building blocks like `ResidualBlock` in `torchvision.models.optical_flow.raft_core`  -- I would make sure to exclude them from `__all__`.

Would that be OK?

torchvision/models/optical_flow/_raft/utils.py
```diff
@@ -0,0 +1,42 @@
+import torch
+import torch.nn.functional as F
+
+
+def grid_sample(img, absolute_grid, *args, **kwargs):
+    """Same as torch's grid_sample, with absolute pixel coordinates instead of normalized coordinates."""
```
fmassa, line 6: Food for thought: would it be a significant limitation if we were to return the flows in normalized coordinates, and perform all computations in normalized coordinates?

Or is it standard that flow images are in absolute coordinates?

Some of the benefits of keeping it in relative coordinates is that you don't need to multiply by the scaling factor when upsampling an image.

fmassa (approved):
Thanks a lot @NicolasHug !
Approving to unblock, but check @datumbox blocking comments as well before merging.

torchvision/models/optical_flow/__init__.py
datumbox: Though calling it `raft_large` would make it consistent with other small/large models, I can see that on the paper they don't actually call it large. They do use `Raft-s` or "small" for the small version, so that's canonical at least. So I understand why you want to avoid naming it like that.

Renaming the file should be OK but @pmeier should confirm. Though I'm not sure what the actual name should be. We typically name the files after their algorithm and we let the model builder have extra info on the variant. 

@fmassa thoughts on this?

NicolasHug: OK, I'll use `raft.py` and `raft_large()` for now to move forward. We can revisit later if needed, thanks for the input

torchvision/models/optical_flow/_raft/raft.py
```diff
@@ -0,0 +1,538 @@
...
+        self.multiplier = multiplier
+
+    def forward(self, x):
+        x = self.conv1(x)
+        x = self.relu(x)
+        x = self.conv2(x)
+        return self.multiplier * x
+
+
+class CorrBlock:
```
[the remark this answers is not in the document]
NicolasHug: I'll see if it helps w.r.t. JIT, but otherwise I feel like we don't really need to make this a `nn.Module`. As you said the semantics don't match right. The original code uses `__call__` instead of `index_pyramid`, but I deliberately changed the name because I thought it would be clearer what the function actually does.

torchvision/models/optical_flow/raft.py
```diff
@@ -0,0 +1,543 @@
...
+        update_block=update_block,
+        mask_predictor=mask_predictor,
+        **kwargs,  # not really needed, all params should be consumed by now
+    )
+
+
+def raft_large(*, pretrained=False, progress=True, **kwargs):
+
+    if pretrained:
+        raise NotImplementedError("Pretrained weights aren't available yet")
```
datumbox, line 483: Perhaps using `NotImplementedError` is indeed the right exception here but I would recommend for now to throw:
```suggestion
        raise ValueError(f"No checkpoint is available for model.")
```

https://github.com/pytorch/vision/blob/3d8723d5f7782881fc19d802819130857ed68ca4/test/test_prototype_models.py#L45-L46



This is because some tests check for this specific exception:

test/test_models.py
```diff
@@ -818,5 +818,31 @@ def test_detection_model_trainable_backbone_layers(model_fn, disable_weight_load
     assert n_trainable_params == _model_tests_values[model_name]["n_trn_params_per_layer"]
 
 
+@needs_cuda
+@pytest.mark.parametrize("model_builder", (models.optical_flow.raft_large, models.optical_flow.raft_small))
+@pytest.mark.parametrize("scripted", (False, True))
```
NicolasHug, line 823: I'm parametrizing over this because testing with `_check_jit_scriptable` unfortunately fails on very few entries, e.g.:

```
Mismatched elements: 153 / 11520 (1.3%)
Greatest absolute difference: 0.0002608299255371094 at index (0, 0, 79, 45) (up to 0.0001 allowed)
Greatest relative difference: 0.021354377198448304 at index (0, 1, 53, 68) (up to 0.0001 allowed)
```

I could add `tol` parameters to the check, but I feel like this current test is just as fine

datumbox (approved):

NicolasHug:
Test failures are unrelated, merging. Thanks a lot for the reviews!!

PR #5025, opened 2021-12-03
[ViT] Support fine-tuning with different image resolution

sallysyw:
As discussed in #4594, we should be able to interpolate embeddings from one resolution to a different one when training ViT models. 
This PR adds the support for it. 

References: [ClassyVision Implementation](https://github.com/facebookresearch/ClassyVision/blob/main/classy_vision/models/vision_transformer.py#L325)

Experiments:
* Launching Command:
```
PYTHONPATH=$PYTHONPATH:`pwd` python -u ~/workspace/scripts/run_with_submitit.py --timeout 3000 --ngpus 8 --nodes 4 --partition train --model vit_b_16 --batch-size 16 --epochs 8 --opt sgd --lr 0.01 --wd 0 --lr-scheduler cosineannealinglr --amp --mixup-alpha 0.2 --auto-augment ra --data-path /datasets01_ontap/imagenet_full_size/061417/ --clip-grad-norm 1 --cutmix-alpha 1.0 --resume /checkpoints/sallysyw/experiments/8022/model_299.pth --train-crop-size 384 --val-crop-size 384
```


torchvision/prototype/models/vision_transformer.py
```diff
@@ -230,6 +231,69 @@ def forward(self, x: torch.Tensor):
...
+            if reset_heads:
+                model_state_copy = {}
+                for k, v in model_state.items():
+                    if not k.startswith("heads"):
+                        model_state_copy[k] = v
+                model_state = model_state_copy
+
+            self.load_state_dict(model_state, strict=strict)
+
+            logging.info("Chekpoint loaded.")
```
datumbox, line 295: We are not currently using `logging` anywhere else. I would suggest removing the debug messages.

torchvision/prototype/models/vision_transformer.py
```diff
@@ -230,6 +231,69 @@ def forward(self, x: torch.Tensor):
...
+            torch._assert(seq_length_1d * seq_length_1d == seq_length, "seq_length is not a perfect square!")
+
+            logging.info(
+                "Interpolating the position embeddings from image "
+                f"{seq_length_1d * self.patch_size} to size {self.image_size}"
+            )
+
+            # (1, hidden_dim, seq_length) -> (1, hidden_dim, seq_l_1d, seq_l_1d)
+            pos_embedding_img = pos_embedding_img.reshape(1, hidden_dim, seq_length_1d, seq_length_1d)
+            new_seq_length_1d = self.image_size // self.patch_size
```
datumbox, line 266: Is it possible that one might be interested in providing another value for `new_seq_length_1d` which is estimated differently? I wonder if it would make sense to make this value an argument to the method and eliminate the need for accessing `self.image_size` and `self.patch_size` here. This will also allow you to remove this method from the nn.Module and instead offer it as a standalone utility method.

One way you could potentially do this is by offering the following signature:
```python
def interpolate_embeddings(model_state: Dict, new_seq_length: int, new_seq_length_1d: int, reset_heads=False) -> Dict
```

If the `new_seq_length_1d` is always calculated as above, then an alternative signature could be:
```python
def interpolate_embeddings(model_state: Dict, new_seq_length: int, image_size: int, patch_size: int, reset_heads=False) -> Dict
```

Thoughts?

datumbox (commented):
Thanks for the PR @sallysyw, let me know your thoughts.

torchvision/prototype/models/vision_transformer.py
sallysyw: Thanks for your comment.

Actually, if we decide to change the API like this, then we assume the users know exactly how to compute  `seq_length_1d` and `seq_length`.
i.e. they should know `seq_length_1d = image_size // patch_size` & `seq_length = (image_size // patch_size) ** 2 + 1`

Do you feel this is a reasonable assumption?
cc @mannatsingh @datumbox

datumbox: If the above is the only reasonable way to set the `seq_length*` variables then we can pass `image_size` and `patch_size` and make the estimation internally. If we anticipate that there will be more reasonable ways to set the vars, then we could pass `seq_length` and `new_seq_length`.

I don't have a strong opinion over how to parameterize this. I just think this would be better served in a utility method than in a method which is attached to the model and explicitly does the loading of the dictionary.

sallysyw: Sure - we can do that. Where shall we park this utility function? In `vision_transformer.py` or in `prototype/utils`?

sallysyw: > If we anticipate that there will be more reasonable ways to set the vars, then we could pass seq_length and new_seq_length.

I don't know whether there are use cases like this, probably @fmassa @mannatsingh  know better here?

datumbox: > Sure - we can do that. Where shall we park this utility function? In vision_transformer.py or in prototype/utils?

I think we should leave it on vision_transformer.py :)

mannatsingh: I think the only way the seq length is calculated is based on image size and patch size, so there shouldn't be a need to take `seq_length` as a parameter directly. In fact, it will probably cause a bit of confusion.

datumbox: Sounds good. Let's make `image_size` and `patch_size` parameters of the method and estimate the `seq_length_1d` and `seq_length` within the utility as @sallysyw suggested. :)

torchvision/prototype/models/vision_transformer.py
```diff
@@ -443,3 +380,74 @@ def vit_l_32(*, weights: Optional[ViT_L_32_Weights] = None, progress: bool = Tru
         progress=progress,
         **kwargs,
     )
+
+
+def interpolate_embeddings(image_size: int, patch_size: int, model_state: Mapping[str, torch.Tensor], reset_heads=False):
```
datumbox: The `model_state` should be of type `'OrderedDict[str, Tensor]'` as you can see [here](https://github.com/malfet/pytorch/blob/master/torch/nn/modules/module.py#L1376-L1377).

Also, could you add the types for other missing vars (`reset_heads`, the return type etc).

torchvision/prototype/models/vision_transformer.py
```diff
@@ -443,3 +380,74 @@ def vit_l_32(*, weights: Optional[ViT_L_32_Weights] = None, progress: bool = Tru
...
+    especially when you want to apply a pre-trained model on images with different resolution.
+
+    Returns a model state dict which can be loaded into the new model.
+
+    Args:
+        image_size (int): Image size of the new model.
+        patch_size (int): Patch size of the new model.
+        model_state (Mapping[str, torch.Tensor]): State dict of the pre-trained model.
+        reset_heads (bool): If true, not copying the state of heads. Default: False.
+    """
```
datumbox, line 399: Also the return on the documentation.

torchvision/prototype/models/vision_transformer.py
```diff
@@ -443,3 +380,74 @@ def vit_l_32(*, weights: Optional[ViT_L_32_Weights] = None, progress: bool = Tru
...
+        model_state["encoder.pos_embedding"] = new_pos_embedding
+
+        if reset_heads:
+            model_state_copy = {}
+            for k, v in model_state.items():
+                if not k.startswith("heads"):
+                    model_state_copy[k] = v
+            model_state = model_state_copy
+
+        return model_state
```
datumbox: I think there is a bug here due to the indentation. What should the method return if `new_seq_length == seq_length`?

torchvision/prototype/models/vision_transformer.py
sallysyw: Hey would you let to check the latest version? I've addressed these issues you mentioned in the last few commits :)

torchvision/prototype/models/vision_transformer.py
```diff
@@ -379,3 +379,73 @@ def vit_l_32(*, weights: Optional[ViT_L_32_Weights] = None, progress: bool = Tru
...
+        # (1, hidden_dim, seq_length) -> (1, hidden_dim, seq_l_1d, seq_l_1d)
+        pos_embedding_img = pos_embedding_img.reshape(1, hidden_dim, seq_length_1d, seq_length_1d)
+        new_seq_length_1d = image_size // patch_size
+
+        # Use bicubic interpolation.
+        # (1, hidden_dim, seq_l_1d, seq_l_1d) -> (1, hidden_dim, new_seq_l_1d, new_seq_l_1d)
+        new_pos_embedding_img = nn.functional.interpolate(
+            pos_embedding_img,
+            size=new_seq_length_1d,
+            mode="bicubic",
```
datumbox, line 431: Should `mode` be a parameter specified by the user?

datumbox (approved):
@sallysyw I pushed to your branch a change to fix the typing issues. The problem here is that the `OrderedDict` is not subscriptable. Using quotes will do the trick.

I'm approving to unblock your work, but it's important to follow up with another PR that adds some tests to cover the method.

torchvision/prototype/models/vision_transformer.py
sallysyw: Yeah sure, let me make it an argument and default to `bicubic` (which is the one gives best performance during fine-tuning).

PR #5026, opened 2021-12-03
Add transforms and presets for optical flow models

NicolasHug:

This PR adds transforms and train/eval presets needed for the optical flow training reference. Ideally https://github.com/pytorch/vision/pull/5004 should be merged first.

A lot of these transforms are just wrappers around existing tranforms, modified so that they accept all four parameters `img1, img2, flow, valid_flow_mask`. For the rest of the transforms I added comments in the code.

CC @fmassa @datumbox @haooooooqi


references/optical_flow/presets.py
```diff
@@ -0,0 +1,64 @@
...
+                crop_size=crop_size, min_scale=min_scale, max_scale=max_scale, stretch_prob=stretch_prob
+            ),
+        ]
+
+        if do_flip:
+            transforms += [T.RandomHorizontalFlip(p=0.5), T.RandomVerticalFlip(p=0.1)]
+
+        transforms += [
+            T.ConvertImageDtype(torch.float32),
+            T.Normalize(mean=0.5, std=0.5),  # map [0, 1] into [-1, 1]
```
NicolasHug, line 57: All transforms are dtype-agnostic, except for the current `RandomErase()` which calls `round()`.

I'm keeping the conversion at the end for now, but in the future it would be worth trying moving it at the beginning of the transforms pipeline.

references/optical_flow/transforms.py
```diff
@@ -0,0 +1,267 @@
...
+        else:
+            # symmetric: same transform for img1 and img2
+            batch = torch.stack([img1, img2])
+            batch = super().forward(batch)
+            img1, img2 = batch[0], batch[1]
+
+        return img1, img2, flow, valid_flow_mask
+
+
+class RandomErase(torch.nn.Module):
```
NicolasHug, line 104: This is similar to our existing `RandomErase`, but replaces with the mean of the pixels, instead of replacing with a given (known-in-advance) value

datumbox: I understand that this class is supposed to be used BEFORE converting the image from int to floats. This is unlike TorchVision's `RandomErasing` class which is applied AFTER converting the image to float, scaling it to 0-1 and normalizing it. I would definitely prefer having in TorchVision following this approach rather than the current one and that's something worth changing on the future, but I think offering another transform right now that operates slightly differently could be confusing. Thoughts?

Concerning using mean VS a fixed value, note that TorchVision's RandomErasing implementation uses a fixed zero value because the images are expected to be de-meaned. I wonder if it would be possible to reuse for now TorchVision's transform if you passed a normalized image. Perhaps that's worth doing even if the outcome is not 100% the same, just to maintain parity between transforms defined in references and those in legacy transforms (this will simplify porting to the new API).

references/optical_flow/transforms.py
```diff
@@ -0,0 +1,267 @@
...
+        img2 = F.vflip(img2)
+        flow = F.vflip(flow) * torch.tensor([1, -1])[:, None, None]
+        if valid_flow_mask is not None:
+            valid_flow_mask = F.vflip(valid_flow_mask)
+        return img1, img2, flow, valid_flow_mask
+
+
+class MaybeResizeAndCrop(torch.nn.Module):
+    # This transform will resize the input with a given proba, and then crop it.
+    # These are the reversed operations of the built-in RandomResizedCrop,
```
datumbox, line 148: I understand that this transform does `resize + crop` while TorchVision's `RandomResizedCrop` does `crop + resize`. Perhaps then a fitting name would be `RandomResizeAndCrop`?

references/optical_flow/transforms.py
```diff
@@ -0,0 +1,267 @@
...
+    # although the order of the operations doesn't matter too much.
+    # The reason we don't rely on RandomResizedCrop is because of a significant
+    # difference in the parametrization of both transforms.
+    #
+    # There *is* a mapping between the inputs of MaybeResizeAndCrop and those of
```
datumbox, line 153: I am confused by this comment. Does this imply that there is a way to map the parameters between the two transforms (just the arguments and NOT the behaviours)? Or does this imply that there is a way to take the inputs of one transform and convert them to the input of the other so that they both behave the same (aka re-parameterization)? 

I guess you mean the former. Also why is your argument here more about different distributions rather than the fact that the order of operations is different?

references/optical_flow/transforms.py
```diff
@@ -0,0 +1,267 @@
...
+
+        # Note: For sparse datasets (Kitti), the original code uses a "margin"
+        # See e.g. https://github.com/princeton-vl/RAFT/blob/master/core/utils/augmentor.py#L220:L220
+        # We don't, not sure it matters much
+        y0 = torch.randint(0, img1.shape[1] - self.crop_size[0], size=(1,)).item()
+        x0 = torch.randint(0, img1.shape[2] - self.crop_size[1], size=(1,)).item()
+
+        img1 = img1[:, y0 : y0 + self.crop_size[0], x0 : x0 + self.crop_size[1]]
+        img2 = img2[:, y0 : y0 + self.crop_size[0], x0 : x0 + self.crop_size[1]]
+        flow = flow[:, y0 : y0 + self.crop_size[0], x0 : x0 + self.crop_size[1]]
```
datumbox, line 209: Why not use `F.crop()` here?

references/optical_flow/presets.py
```diff
@@ -0,0 +1,64 @@
...
+        do_flip=True,
+    ):
+        super().__init__()
+
+        transforms = [
+            T.PILToTensor(),
+            T.AsymmetricColorJitter(
+                brightness=brightness, contrast=contrast, saturation=saturation, hue=hue, p=asymmetric_jitter_prob
+            ),
+            T.RandomApply([T.RandomErase()], p=0.5),
```
datumbox, line 46: If we do end up supporting a different `RandomErase` than TochVision's `RandomErasing`, we should consider adding the `p` in the constructor of `RandomErase` to make them more similar.

references/optical_flow/presets.py
datumbox: As mentioned earlier, I think it makes more sense to eventually support a `RandomErase` that doesn't take as input a normalized image. This will make the transform similar to all other transforms and enable users to visualize its effect easier (now one needs to undo the scaling). But for now, as mentioned above, it might be confusing to users that some Random Erasing transforms are applied prior normalization while other after.

datumbox (commented):
Thanks for the PR @NicolasHug, I've added a few comments for discussion. Might be worth discussing this offline and then summarize our decision to speed this up.

references/optical_flow/transforms.py
```diff
@@ -0,0 +1,267 @@
...
+        self.p = p
+
+    def forward(self, img1, img2, flow, valid_flow_mask):
+
+        if torch.rand(1) < self.p:
+            # asymmetric: different transform for img1 and img2
+            img1 = super().forward(img1)
+            img2 = super().forward(img2)
+        else:
+            # symmetric: same transform for img1 and img2
```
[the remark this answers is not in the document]
pmeier: As it stands, this would not be supported. A transform always treats a sample as atomic unit and so multiple images in the same sample would be transformed with the same parameters.

references/optical_flow/transforms.py
NicolasHug: It's the former, I'll clarify.

> Also why is your argument here more about different distributions rather than the fact that the order of operations is different?

The order of operation doesn't matter much in practice: resizing a crop or cropping a resized image would give the same result, up to some artifacts at the borders of the output (due to interpolation).

Here the problem we have is really about the distribution of the randomly-sampled parameters. Here, scale_x and scale_y are randomly sampled and their density looks like this

image


Now if I map these to the current parametrization of our RandomResizedCrop, this would be equivalent to the following distribution for the `scale` parameter

image


and the following for the `ratio` parameter

image

Unfortunately, `RandomResizedCrop` is unable to sample according to these 2 distributions, no matter what we pass as values.

references/optical_flow/transforms.py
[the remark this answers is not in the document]
NicolasHug: OK, I'll clarify.

Ultimately this is a special case of `RandomApply(t1, t2, [p, p - 1])`, so there's nothing too fancy here

references/optical_flow/presets.py
NicolasHug: OK,  thanks for the input. I'll try to move this at the beginning of the transforms, and change `RandomErase` to replace with 0 instead of the mean (which should now the same).

I'll start some runs on Chairs + Things today and we should get the results tomorrow. If epe is still good I'll keep the changes, and otherwise I'll keep things as-is and will mark a TODO + open an issue to keep track of this progress. Does that sound good?

datumbox: Just to confirm that we mean the same thing, you will:
1. Move the `RandomErase` after `Normalize` similar to how `RandomErasing` works in TorchVision (all other transforms remain where they are as normal):
https://github.com/pytorch/vision/blob/d367a01a18a3ae6bee13d8be3b63fd6a581ea46f/torchvision/transforms/transforms.py#L1619-L1626

2. Use `mean=0` as now it should be equivalent. Perhaps also check if your implementation can reuse `F.erase()` or other parts of `RandomErasing` class.
3. Run `Chairs + Things` to confirm that the `epe` is OK else rollback.

If that's what you mean, then sounds like a plan!

references/optical_flow/transforms.py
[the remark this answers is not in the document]
datumbox: @pmeier No worries, this is why we give the option for someone to write custom transforms without he magic of the new API. For weird cases like this. Could you now confirm that this is indeed a workaround we can apply?

references/optical_flow/presets.py
NicolasHug: Yes, thanks for the correction.

references/optical_flow/transforms.py
```diff
@@ -0,0 +1,267 @@
+import torch
+import torchvision.transforms as T
+import torchvision.transforms.functional as F
+
+
+class ValidateModelInput(torch.nn.Module):
```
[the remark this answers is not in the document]
NicolasHug: For some of these (e.g. `valid_flow_mask`) we wouldn't be able to make the assertion in the model, because the model doesn't use it as input - it's only used for the loss computation.

Similarly we assert that the dtype is float32 but we wouldn't want to leave this in the model. I know this is somewhat exotic w.r.t. the rest of the references, but would it be OK to leave it (possibly commenting it out from the Presets? I feel like this adds value at least as documentation

references/optical_flow/transforms.py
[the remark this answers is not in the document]
pmeier: @NicolasHug 

> Sequential(take_care_of_img1_only, take_care_of_img2_only)

I'm guessing `take_care_of_img1_only` and `take_care_of_img2_only` are transforms here, correct? If yes, how would you tell the transform to only handle one or the other image if both receive the full sample?

I think this is one of the cases @datumbox mentioned where we need to circumvent the automatic dispatch a little. In case we want to transform both samples separately, we could split the sample and and perform the transformation once for the sample minus image 2 and once for image2. The problem I see with this, is that it can't be automated without assumptions about how the sample is structured. So we either need to use the same structure for every dataset (for example flat dictionary with `image1` and `image2` keys) or provide a way to parametrize the transform.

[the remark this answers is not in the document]
NicolasHug: > I'm guessing take_care_of_img1_only and take_care_of_img2_only are transforms here, correct? If yes, how would you tell the transform to only handle one or the other image if both receive the full sample?

Each transform would receive the entire input (which IIRC is a dict) and operate on a subset of that dict.

Are you suggesting that img1 and img2 would be concatenated?

references/optical_flow/transforms.py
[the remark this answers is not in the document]
datumbox: Yeah that's fine. It's not a big issue.

datumbox (approved):
LGTM thanks @NicolasHug.

references/optical_flow/transforms.py
```diff
@@ -0,0 +1,255 @@
...
+    def __init__(self, p=0.5, scale=(0.02, 0.33), ratio=(0.3, 3.3), value=0, inplace=False, max_erase=1):
+        super().__init__()
+        self.max_erase = max_erase
+        assert self.max_erase > 0
+
+    def forward(self, img1, img2, flow, valid_flow_mask):
+        if torch.rand(1) > self.p:
+            return img1, img2, flow, valid_flow_mask
+
+        for _ in range(torch.randint(self.max_erase, size=(1,)).item()):
```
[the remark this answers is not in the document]
fmassa: a comment here would be good.

references/optical_flow/transforms.py
```diff
@@ -0,0 +1,255 @@
...
+        self.max_stretch = 0.2
+
+    def forward(self, img1, img2, flow, valid_flow_mask):
+        # randomly sample scale
+        h, w = img1.shape[-2:]
+        # Note: in original code, they use + 1 instead of + 8 for sparse datasets (e.g. Kitti)
+        # It shouldn't matter much
+        min_scale = max((self.crop_size[0] + 8) / h, (self.crop_size[1] + 8) / w)
+
+        scale = 2 ** torch.FloatTensor(1).uniform_(self.min_scale, self.max_scale).item()
```
fmassa, line 174: Using `torch.FloatTensor` is deprecated API, and the recommended solution is to use `torch.empty(..., dtype=torch.float32)` instead.

references/optical_flow/transforms.py
```diff
@@ -0,0 +1,255 @@
...
+        valid_flow_new[0] *= scale_x
+        valid_flow_new[1] *= scale_y
+
+        flow_new[:, ii_valid_new, jj_valid_new] = valid_flow_new
+        valid_new[ii_valid_new, jj_valid_new] = 1
+
+        return flow_new, valid_new
+
+
+class Compose:
```
NicolasHug, line 248: note for self before merging: should be torch.nn.Module

fmassa (approved):
Thanks for the PR!

I've only a couple of minor comments, otherwise LGTM!

PR #5027, opened 2021-12-03
Add training reference for optical flow models

NicolasHug:

This PR adds a training reference for optical flow models (mostly just RAFT at the moment), as well utilities for evaluating the model on Sintel (epe, 1px, 3px, 5px) or Kitti (per-image-epe, F1).

Right now, the training script assumes that CUDA is available and that DDP is available as well. It must be run with `torchrun`, e.g.

```
torchrun --nproc_per_node 8 --nnodes 1 references/optical_flow/train.py --batch-size 10 --train-dataset chairs --val-dataset kitti sintel
```

Our custom `run_with_submitit.py` script is also partially supported (but it's not as useful anyway, because the training procedure involves training on more than one dataset).

I marked a few TODO comments, leaving them as future potential improvements.

CC @fmassa @datumbox @haooooooqi

cc @datumbox


references/optical_flow/utils.py
```diff
@@ -0,0 +1,390 @@
...
+    return t
+
+
+def freeze_batch_norm(model):
+    for m in model.modules():
+        if isinstance(m, torch.nn.BatchNorm2d):
+            m.eval()
+
+
+def map_orig_to_ours(orig, mine=None):
```
NicolasHug, line 272: I will obviously remove this monstruosity before merging.
Sadly I couldn't just `zip` the dict weights into one-another, because the original RAFT weights have some extra duplicated weights, which offset the whole thing. See e.g. `self.norm3` https://github.com/princeton-vl/RAFT/blob/aac9dd54726caf2cf81d8661b07663e220c5586d/core/extractor.py#L20 which will be present twice

references/optical_flow/utils.py
```diff
@@ -0,0 +1,390 @@
...
+                    )
+            i += 1
+            end = time.time()
+        total_time = time.time() - start_time
+        total_time_str = str(datetime.timedelta(seconds=int(total_time)))
+        print(f"{header} Total time: {total_time_str}")
+
+
+def sequence_loss(flow_preds, flow_gt, valid_flow_mask, gamma=0.8, max_flow=400):
+    """Loss function defined over sequence of flow predictions"""
```
NicolasHug, line 156: This loss function is very RAFT-specific, because it assumes the model outputs a series of predictions, instead of a single predicted flow.

references/optical_flow/utils.py
```diff
@@ -0,0 +1,390 @@
...
+import time
+from collections import defaultdict
+from collections import deque
+
+import torch
+import torch.distributed as dist
+import torch.nn.functional as F
+
+
+class SmoothedValue:
```
NicolasHug, line 12: This, and the `MetricLogger` below have been copy/pasted from the classification references. I only made some very minor changes like setting some defaults, probably not worth reviewing.

references/optical_flow/train.py
```diff
@@ -0,0 +1,331 @@
...
+
+    parser.add_argument("--gamma", type=float, default=0.8, help="exponential weighting for loss. Must be < 1.")
+
+    parser.add_argument("--dist-url", default="env://", help="URL used to set up distributed training")
+
+    # TODO: remove
+    parser.add_argument("--map-orig-to-ours", action="store_true")
+
+    # TODO: remove the default
+    _DATASET_ROOT = "/data/home/nicolashug/cluster/work/downloads"
```
NicolasHug, line 318: will remove these before merging

references/optical_flow/train.py
```diff
@@ -0,0 +1,331 @@
...
+        epe = ((flow_pred - flow_gt) ** 2).sum(dim=1).sqrt()
+
+        logger.meters["epe"].update(epe.mean().item(), n=epe.numel())
+        for distance in (1, 3, 5):
+            logger.meters[f"{distance}px"].update((epe < distance).float().mean().item(), n=epe.numel())
+
+        relative_epe = epe / (flow_gt ** 2).sum(dim=1).sqrt()
+        if valid_flow_mask is not None:
+            epe, relative_epe = epe[valid_flow_mask], relative_epe[valid_flow_mask]
+        bad_predictions = ((epe > 3) & (relative_epe > 0.05)).float()
```
fmassa, line 102: nit: this part of the evaluation could probably be shared with the training loss, which also computes a very similar thing I believe.

```python
    last_pred = flow_preds[-1]
    epe = ((last_pred - flow_gt) ** 2).sum(dim=1).sqrt()
    epe = epe[valid_flow_mask]

    metrics = {
        "flow_loss": flow_loss,
        "epe": epe.mean().item(),
        "1px": (epe < 1).float().mean().item(),
        "3px": (epe < 3).float().mean().item(),
        "5px": (epe < 5).float().mean().item(),
    }
```

fmassa (approved):
Thanks for the PR!

I've made a few comments, all of which can be addressed in follow-up PRs.

In particular, I think we can split a bit more the loss so that there are parts that can be re-used across train / val.

references/optical_flow/train.py
```diff
@@ -0,0 +1,331 @@
...
+                break
+
+        # Note: we don't sync the SmoothedValues across processes, so the printed metrics are just those of rank 0
+        print(f"Epoch {current_epoch} done. ", logger)
+
+        current_epoch += 1
+
+        if args.rank == 0:
+            torch.save(model.state_dict(), Path(args.output_dir) / f"{args.name}_{current_epoch}.pth")
+            torch.save(model.state_dict(), Path(args.output_dir) / f"{args.name}.pth")
```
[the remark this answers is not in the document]
datumbox: We should definitely do this. I think it's worth refactoring the script to have same functionality and structure as other reference scripts. Moreover we will need to link the ref scripts with the model prototype and add the `--weights` feature switch.

@NicolasHug do you mind creating an issue for all the above so that we dont forget?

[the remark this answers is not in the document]
NicolasHug: Opened https://github.com/pytorch/vision/issues/5056

Issue #5017, opened 2021-11-30
unittest_onnx test failure


prabhat00155:
https://app.circleci.com/pipelines/github/pytorch/vision/12995/workflows/31211a7c-01cc-4675-99b1-4d1bd0c0d5f0/jobs/1035420

Issue seen in https://github.com/pytorch/vision/pull/4983 and perhaps introduced by https://github.com/pytorch/vision/pull/4692.

cc @pmeier @neginraoof

PR #5029, opened 2021-12-04
Skip test_roi_align_aligned

prabhat00155:
Resolves https://github.com/pytorch/vision/issues/5017.
ROIAlign with aligned=True is not supported in ONNX, see https://github.com/pytorch/vision/pull/4692/. 

cc @pmeier @neginraoof


NicolasHug (approved):
Thanks @prabhat00155

PR #5033, opened 2021-12-06
Add new weights for efficientnet_b1 model

prabhat00155:
```
gpurun torchrun --nproc_per_node=1 train.py --test-only --weights EfficientNet_B1_Weights.ImageNet1K_V2 --model efficientnet_b1 -b 1

Test:  Acc@1 79.838 Acc@5 94.934       
```


torchvision/prototype/models/efficientnet.py
```diff
@@ -94,7 +94,17 @@ class EfficientNet_B1_Weights(WeightsEnum):
             "acc@5": 94.186,
         },
     )
-    default = ImageNet1K_V1
+    ImageNet1K_V2 = Weights(
+        url="https://download.pytorch.org/models/efficientnet_b1-c2e40cbf.pth",
+        transforms=partial(ImageNetEval, crop_size=224, resize_size=232, interpolation=InterpolationMode.BILINEAR),
```
datumbox, line 99: Blocking: As discussed offline the network was trained with an inference resolution 224x224 in mind instead of the 240x240 required by B1. This means that you could push this model to even higher accuracies! So first of all change your inference configuration to `--val-crop-size 240 --val-resize-size 240`.

Check a few more `--train-crop-size` values for the FixRes mitigations and keep the one that gives you the best trained model at 240 inference. Once you have the final configuration, do 2 more runs and then select the best possible model and tune `--val-resize-size`.

torchvision/prototype/models/efficientnet.py
```diff
@@ -94,7 +94,17 @@ class EfficientNet_B1_Weights(WeightsEnum):
...
+        meta={
+            **_COMMON_META,
```
datumbox, line 101: You should have also defined `"size": (240, 240),` in the meta. This indicates the size of the image that goes in the network.

You also must overwrite the `interpolation` value on the meta as unlike all other models, you use bilinear.

torchvision/prototype/models/efficientnet.py
```diff
@@ -94,7 +94,17 @@ class EfficientNet_B1_Weights(WeightsEnum):
...
+            "recipe": "https://github.com/pytorch/vision/issues/3995#new-recipe-with-lr-wd-crop-tuning",
```
datumbox, line 102: Don't forget to update the stats and recipe once you are done here.

datumbox (changes requested):
@prabhat00155 as discussed offline we can push this model even further, which is great news. I'm marking as blocked to ensure this is not accidentally merged.

torchvision/prototype/models/efficientnet.py
datumbox: @prabhat00155 This still needs to be updated. Could you please add the new command and accuracy on the linked issue?

torchvision/prototype/models/efficientnet.py
```diff
@@ -95,17 +95,17 @@ class EfficientNet_B1_Weights(WeightsEnum):
         },
     )
     ImageNet1K_V2 = Weights(
-        url="https://download.pytorch.org/models/efficientnet_b1-c2e40cbf.pth",
-        transforms=partial(ImageNetEval, crop_size=224, resize_size=232, interpolation=InterpolationMode.BILINEAR),
+        url="https://download.pytorch.org/models/efficientnet_b1-c27df63c.pth",
```
datumbox, line 98: I assume this is also on manifold?

datumbox (approved):
Very nice! Massive bump on accuracy :)

I've added 1 comment and 1 question but approving to unblock.

torchvision/prototype/models/efficientnet.py
prabhat00155: Yup.

PR #5038, opened 2021-12-07
Log io usage

kazhang:
Log IO api usages.
Partially-close: #4955
Partially-close: #5052


torchvision/csrc/io/image/cpu/decode_jpeg.cpp
```diff
@@ -70,6 +70,7 @@ static void torch_jpeg_set_source_mgr(
 } // namespace
 
 torch::Tensor decode_jpeg(const torch::Tensor& data, ImageReadMode mode) {
+  C10_LOG_API_USAGE_ONCE("torchvision.io.decode_jpeg_cpp");
```
datumbox, line 73: For context, @kazhang added `_cpp` on the name of each C++ logger to distinguish it from the Python method calls. Alternatives:
1. Use the above `_cpp` postfix
2. Leave the same name between C++ and Python the same
3. Use `torchvision.csrc.io.decode_jpeg`
4. Something else?

cc @NicolasHug @fmassa @prabhat00155

datumbox (dismissed):

torchvision/csrc/io/image/cpu/decode_jpeg.cpp
NicolasHug: The way it's currently done (option 1) is fine with me.

Just curious, do we expect some users to directly call the cpp function? I think we don't provide BC guarantees on these, do we?

fmassa: Leaving the same name is not ideal as each instance will be logged twice (with the same name).

If we want to start tracking down adoption of the raw C++ API (e.g. through torchscript) vs python, I think I would add a namespace like option 3, so that it simplifies grouping together the different function calls

prabhat00155: Yeah, we shouldn't log twice. C++ video APIs are used directly by a number of internal teams. Wouldn't just logging in the C++ function for python APIs with a C++ backend solve this?

NicolasHug: If we just log in the C++ function, I think we would be losing the ability to discriminate between C++ and Python calls.

prabhat00155: Is that important to track?

NicolasHug: I would say so, it could for example give us some insight on potential breakage when we change something on the C++ side

prabhat00155: If that's the case, we could potentially have an optional argument passed when making Python to C++ call which could be used when logging, indicating whether the user call is from Python or directly a C++ call.

datumbox: A few thoughts about this:
1. Discriminating between C++ and Python is important to understand how the code is used by internal FB projects.
2. Tracking both C++ and Python will allow us account for situations such as when a team uses directly the C++ backend OR when the Python API along with avpy is used.
3. Using a pseudo-submodule such as `torchvision.csrc.` for all C++ log messages aligns with the structure of our code. 
4. I would favour a simple logging approach over a complex strategy. Hence I would avoid introducing extra arguments just to avoid duplicate logging calls because this complicates the API. 
5. I would also favour the adoption of a simple policy (such as put the log call in all constructors and all public methods), even if this leads to 2 separate calls. Having a simple policy means easier enforcement on PR reviews.

kazhang:
I opened an issue(#5052) to get consensus around logging policy and event format.
cc @datumbox @fmassa  @NicolasHug @prabhat00155

datumbox (approved):
LGTM, thanks!

PR #5042, opened 2021-12-07
Removing code coverage and cleanup tox

oke-aditya:
I don't think so these files or stuff are used let's see.


.circleci/unittest/linux/scripts/run_test.sh
```diff
@@ -7,4 +7,4 @@ conda activate ./env
 
 export PYTORCH_TEST_WITH_SLOW='1'
 python -m torch.utils.collect_env
-pytest --cov=torchvision --junitxml=test-results/junit.xml -v --durations 20
+pytest -v --durations 20
```
NicolasHug, line 10: we shouldn't remove the `--junitxml=test-results/junit.xml` part, this is what CircleCI uses to provide the "extended" test failure summary https://circleci.com/docs/2.0/collect-test-data/

oke-aditya: Hmm, I didn't know that. I thought we dump the coverage file as xml. (Something like [this](https://github.com/oke-aditya/quickvision/blob/master/.github/workflows/ci-testing.yml#L41))

I think you are right. Let me rollback this
https://circleci.com/docs/2.0/collect-test-data/#pytest

oke-aditya:
```suggestion
pytest --junitxml=test-results/junit.xml -v --durations 20
```

.circleci/unittest/windows/scripts/run_test.sh
```diff
@@ -10,4 +10,4 @@ source "$this_dir/set_cuda_envs.sh"
 
 export PYTORCH_TEST_WITH_SLOW='1'
 python -m torch.utils.collect_env
-pytest --cov=torchvision --junitxml=test-results/junit.xml -v --durations 20
+pytest -v --durations 20
```
oke-aditya, line 13:
```suggestion
pytest --junitxml=test-results/junit.xml -v --durations 20
```

oke-aditya (commented):
Applying changes on the fly

oke-aditya:
I think this should be good. cc @NicolasHug @pmeier

oke-aditya:
Additionally we can think about cleaning up `pytest.ini`, `setup.cfg` and `pyproject.toml` to a single file. Probably just keep `pyproject.ml`. (Not sure if mypy too can be combined)

NicolasHug (commented):
Thanks @oke-aditya .

I don't mind removing the `--cov` part as we don't use it (I think).

NicolasHug:
> Additionally we can think about cleaning up pytest.ini, setup.cfg and pyproject.toml to a single file. Probably just keep pyproject.ml. (Not sure if mypy too can be combined)


I would rather keep things as is.
Any change we make about these will need to be carefully checked on the fbcode side, which has its own requirements and constraints. We wouldn't gain a tremendous value by merging all these into one file anyway, and the low benefit doesn't outweight the extra-work IMHO.

tox.ini
```diff
@@ -1,4 +0,0 @@
-[flake8]
-max-line-length = 120
-ignore = F401,E402,F403,W503,W504
-exclude = docs/src
```
oke-aditya, line 4: Since it was there in this file I thought let's keep it

setup.cfg
```diff
@@ -17,7 +17,7 @@ per-file-ignores =
     torchvision/models/mobilenet.py: F401, F403
     torchvision/models/quantization/mobilenet.py: F401, F403
     test/smoke_test.py: F401
-exclude = venv
+exclude = venv, docs/src
```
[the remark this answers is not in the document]
oke-aditya: See my comment below.

tox.ini
NicolasHug: This file is very outdated I think, we shouldn't trust whatever is in it. Let's please keep setup.cfg as it was

setup.cfg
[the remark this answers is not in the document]
oke-aditya:
```suggestion
exclude = venv
```

tox.ini
oke-aditya: Reverted.

oke-aditya:
I'm little bit scared :fearful: Since last time I had broken the docs while modifying such files

pmeier (approved):
LGTM, thanks @oke-aditya!

oke-aditya:
CI seems all green. Hopefully this doesn't break stuff on fbcode side.

oke-aditya:
I guess it's not breaking anything. (But take no risks, I can close it otherwise) 
 
cc @datumbox @NicolasHug

datumbox (approved):
LGTM, thanks!

oke-aditya:
Aah finally, Is everything fine on fbcode side? I don't want to break something again :stuck_out_tongue:

PR #5043, opened 2021-12-07
Add raft builders and presets in prototypes

NicolasHug:

This PR adds the `raft_large()` and `raft_small()` builders, following the new model prototype API.

cc @datumbox @bjuncek

### DIFF
--- a/test/test_prototype_models.py
+++ b/test/test_prototype_models.py
@@ -3,7 +3,7 @@ import importlib
 import pytest
 import test_models as TM
 import torch
-from common_utils import cpu_and_gpu, run_on_env_var
+from common_utils import cpu_and_gpu, run_on_env_var, needs_cuda
 from torchvision.prototype import models
 from torchvision.prototype.models._api import WeightsEnum, Weights
 from torchvision.prototype.models._utils import handle_legacy_interface
@@ -75,10 +75,12 @@ def test_get_weight(name, weight):
     + TM.get_models_from_module(models.detection)
     + TM.get_models_from_module(models.quantization)
     + TM.get_models_from_module(models.segmentation)
-    + TM.get_models_from_module(models.video),
+    + TM.get_models_from_module(models.video)
+    + TM.get_models_from_module(models.optical_flow),
 )
 def test_naming_conventions(model_fn):
     weights_enum = _get_model_weights(model_fn)
+    print(weights_enum)
     assert weights_enum is not None
     assert len(weights_enum) == 0 or hasattr(weights_enum, "default")
 
@@ -117,13 +119,22 @@ def test_video_model(model_fn, dev):
     TM.test_video_model(model_fn, dev)
 
 
+@needs_cuda
+@pytest.mark.parametrize("model_builder", (models.optical_flow.raft_large, models.optical_flow.raft_small))
+@pytest.mark.parametrize("scripted", (False, True))
+@run_if_test_with_prototype
+def test_raft(model_builder, scripted):
+    TM.test_raft(model_builder, scripted)
+
+
 @pytest.mark.parametrize(
     "model_fn",
     TM.get_models_from_module(models)
     + TM.get_models_from_module(models.detection)
     + TM.get_models_from_module(models.quantization)
     + TM.get_models_from_module(models.segmentation)
-    + TM.get_models_from_module(models.video),
+    + TM.get_models_from_module(models.video)
+    + TM.get_models_from_module(models.optical_flow),
 )
 @pytest.mark.parametrize("dev", cpu_and_gpu())
 @run_if_test_with_prototype
@@ -144,6 +155,9 @@ def test_old_vs_new_factory(model_fn, dev):
         },
         "video": {
             "input_shape": (1, 3, 4, 112, 112),
+        },
+        "optical_flow": {
+            "input_shape": (1, 3, 128, 128),
         },
     }
     model_name = model_fn.__name__
--- a/test/test_prototype_models.py
+++ b/test/test_prototype_models.py
@@ -120,7 +120,7 @@ def test_video_model(model_fn, dev):
 
 
 @needs_cuda
-@pytest.mark.parametrize("model_builder", (models.optical_flow.raft_large, models.optical_flow.raft_small))
+@pytest.mark.parametrize("model_builder", TM.get_models_from_module(models.optical_flow))
 @pytest.mark.parametrize("scripted", (False, True))
 @run_if_test_with_prototype
 def test_raft(model_builder, scripted):
--- a/torchvision/prototype/models/optical_flow/raft.py
+++ b/torchvision/prototype/models/optical_flow/raft.py
@@ -7,6 +7,7 @@ from torchvision.models.optical_flow.raft import _raft, BottleneckBlock, Residua
 from torchvision.prototype.transforms import RaftEval
 
 from .._api import WeightsEnum, Weights
+from .._utils import handle_legacy_interface
 
 
 __all__ = (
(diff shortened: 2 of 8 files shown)


torchvision/prototype/transforms/_presets.py
```diff
@@ -97,3 +98,36 @@ def forward(self, img: Tensor, target: Optional[Tensor] = None) -> Tuple[Tensor,
...
+        img1 = F.normalize(img1, mean=0.5, std=0.5)
+        img2 = F.normalize(img2, mean=0.5, std=0.5)
+
+        img1 = img1.contiguous()
+        img2 = img2.contiguous()
+
+        return img1, img2, flow, valid_flow_mask
+
+    def _pil_or_numpy_to_tensor(self, img1, img2, flow, valid_flow_mask):
+        if not isinstance(img1, Tensor):
```
NicolasHug, line 123: I find it a bit confusing that we type the input as `Tensor` and actually handle the case where it's not a tensor. I saw that on the other presets so I did the same. I assume that this is only temporary to check these presets on the current transforms (which return PIL images), and that we will remove the conversions eventually?

torchvision/prototype/models/optical_flow/raft.py
```diff
@@ -0,0 +1,162 @@
...
+
+class Raft_Large_Weights(WeightsEnum):
+    C_T = Weights(
+        # Chairs + Things
+        url="",
+        transforms=RaftEval,
+        meta={
+            "recipe": "",
+            "epe": -1234,
+        },
```
NicolasHug, line 27: I'll fill these up ASAP when I get the results and when https://github.com/pytorch/vision/pull/5027 and https://github.com/pytorch/vision/pull/5026 are merged.

@datumbox can the `meta` field support epe for one than one dataset?
I assume I can just have different keys like "sintel_train_epe", "sintel_test_epe", "kitti_train_epe", etc?

torchvision/prototype/models/optical_flow/raft.py
```diff
@@ -0,0 +1,162 @@
...
+        recurrent_block_padding=((0, 2), (2, 0)),
+        # Flow head
+        flow_head_hidden_size=256,
+        # Mask predictor
+        use_mask_predictor=True,
+        **kwargs,
+    )
+
+
+def raft_small(weights: Optional[Raft_Small_Weights] = None, progress=True, **kwargs):
```
datumbox, line 117: Same remarks as above for the entire method here.

torchvision/prototype/transforms/_presets.py
datumbox: Good point. It's also because the user is supposed to use these transforms during inference. At that point, you don't know if they chose to read the image with PIL or with TV's io. So here we support both. 

This is also done because the reference scripts for other tasks only support PIL. BTW now that you added a prototype section for your model, you should add a support for it in your reference scripts on the other PR.

torchvision/prototype/transforms/_presets.py
```diff
@@ -97,3 +98,36 @@ def forward(self, img: Tensor, target: Optional[Tensor] = None) -> Tuple[Tensor,
...
+        # map [0, 1] into [-1, 1]
+        img1 = F.normalize(img1, mean=0.5, std=0.5)
+        img2 = F.normalize(img2, mean=0.5, std=0.5)
+
+        img1 = img1.contiguous()
+        img2 = img2.contiguous()
+
+        return img1, img2, flow, valid_flow_mask
+
+    def _pil_or_numpy_to_tensor(self, img1, img2, flow, valid_flow_mask):
```
datumbox, line 122: Typing for JIT script:
```suggestion
    def _pil_or_numpy_to_tensor(self, img1: Tensor, img2: Tensor, flow: Optional[Tensor], valid_flow_mask: Optional[Tensor])-> Tuple[Tensor, Tensor, Optional[Tensor], Optional[Tensor]]:
```

Or you can also decide to move this private method in the main body of forward since you use it once, to avoid defining typing multiple times. Up to you.

torchvision/prototype/transforms/_presets.py
```diff
@@ -97,3 +98,36 @@ def forward(self, img: Tensor, target: Optional[Tensor] = None) -> Tuple[Tensor,
...
+        if not isinstance(img1, Tensor):
+            img1 = F.pil_to_tensor(img1)
+        if not isinstance(img2, Tensor):
+            img2 = F.pil_to_tensor(img2)
+
+        if flow is not None and not isinstance(flow, np.ndarray):
```
datumbox, line 128: Typo?
```suggestion
        if flow is not None and not isinstance(flow, torch.Tensor):
```

This also will help you drop the numpy dependency which won't play well with JIT.

torchvision/prototype/transforms/_presets.py
```diff
@@ -97,3 +98,36 @@ def forward(self, img: Tensor, target: Optional[Tensor] = None) -> Tuple[Tensor,
...
+            flow = torch.from_numpy(flow)
+        if valid_flow_mask is not None and not isinstance(flow, np.ndarray):
```
datumbox, line 130: Typo?
```suggestion
        if valid_flow_mask is not None and not isinstance(valid_flow_mask, torch.Tensor):
```

datumbox (commented):
Thanks for the PR @NicolasHug. I've added a few comments to discuss with you, happy to chat offline if that's easier and summarize afterwards.

torchvision/prototype/models/optical_flow/raft.py
```diff
@@ -0,0 +1,162 @@
...
+        transforms=RaftEval,
+        meta={
+            "recipe": "",
+            "epe": -1234,
+        },
+    )
+    default = C_T
+
+
+def raft_large(weights: Optional[Raft_Large_Weights] = None, progress=True, **kwargs):
```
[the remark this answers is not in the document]
NicolasHug: I'll add the `*`, but regarding the deprecation, I'm not sure we need it.

Since we're planning to move the model prototype into `main` before the next release, users won't ever be using the `pretrained` parameter, so there's no need to handle this.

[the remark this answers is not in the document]
datumbox: You will need this the moment you put pretrained weights. Some of the tests in prototype still use `pretrained=True` to confirm that the builders between the two areas operate the same using real data. So I would probably recommend adding it here to simplify the situation. You will also plan to for the worse case scenario in which case the prototype builders might not be released in time. Overall I think that RAFT is more likely to be released prior to model builders.

torchvision/prototype/models/optical_flow/raft.py
```diff
@@ -0,0 +1,168 @@
...
+        progress (bool): If True, displays a progress bar of the download to stderr
+        kwargs (dict): Parameters that will be passed to the :class:`~torchvision.models.optical_flow.RAFT` class
+            to override any default.
+
+    Returns:
+        nn.Module: The model.
+    """
+
+    if weights is not None:
+        raise ValueError("No checkpoint is available for raft_large")
```
datumbox, line 90: These two lines can go, it's handled by the decorator.

torchvision/prototype/models/optical_flow/raft.py
```diff
@@ -0,0 +1,168 @@
...
+        kwargs (dict): Parameters that will be passed to the :class:`~torchvision.models.optical_flow.RAFT` class
+            to override any default.
+
+    Returns:
+        nn.Module: The model.
+
+    """
+
+    if weights is not None:
+        raise ValueError("No checkpoint is available for raft_small")
```
datumbox, line 139: Same here.

torchvision/prototype/transforms/_presets.py
```diff
@@ -97,3 +97,38 @@ def forward(self, img: Tensor, target: Optional[Tensor] = None) -> Tuple[Tensor,
                 target = F.pil_to_tensor(target)
             target = target.squeeze(0).to(torch.int64)
         return img, target
+
+
+class RaftEval(nn.Module):
```
datumbox, line 102: Github won't let me comment on the right line, but add RaftEval in `__all__`

datumbox (approved):
LGTM, only couple of minor nits and you are ready to merge on green CI.

datumbox:
@NicolasHug I just realized there are a few other things that we need to add in this PR, just because you introduce a new family of models (optical flow).

1. Add `models.optical_flow` in all places where we do:
https://github.com/pytorch/vision/blob/588e9b5ed7e2327f9bf5b2136b8c5fd0a85a62d8/test/test_prototype_models.py#L72-L79

2. Add a `test_raft` method similar to the detection (not sure why we call it [test_raft](https://github.com/pytorch/vision/blob/01ffb3ae9f9cc3a445bcaabf787251a05e67c7a7/test/test_models.py#L825) on the test_models.py file instead of test_optical_flow_model but the name should match between prototypes and non-prototypes) https://github.com/pytorch/vision/blob/588e9b5ed7e2327f9bf5b2136b8c5fd0a85a62d8/test/test_prototype_models.py#L106-L110

3. Add default params for `"optical_flow"` models on the following test (not sure what parameter is right for you here, but the test expects you to have it defined the moment you put weights):
https://github.com/pytorch/vision/blob/588e9b5ed7e2327f9bf5b2136b8c5fd0a85a62d8/test/test_prototype_models.py#L130-L137

test/test_prototype_models.py
```diff
@@ -117,13 +119,22 @@ def test_video_model(model_fn, dev):
...
+
 @pytest.mark.parametrize(
     "model_fn",
     TM.get_models_from_module(models)
     + TM.get_models_from_module(models.detection)
     + TM.get_models_from_module(models.quantization)
     + TM.get_models_from_module(models.segmentation)
-    + TM.get_models_from_module(models.video),
+    + TM.get_models_from_module(models.video)
+    + TM.get_models_from_module(models.optical_flow),
```
NicolasHug, line 137: Once weights are supported, I'll need to edit the code below a bit because the raft model needs 2 inputs instead of 1.

datumbox: Hmm... You are right.

If we can't modify the method yet to make it ready, then there is no point adding the `optical_flow` module yet. Up to you on whether you want to revert or update the method.

test/test_prototype_models.py
```diff
@@ -117,13 +119,22 @@ def test_video_model(model_fn, dev):
     TM.test_video_model(model_fn, dev)
 
 
+@needs_cuda
+@pytest.mark.parametrize("model_builder", (models.optical_flow.raft_large, models.optical_flow.raft_small))
```
datumbox, line 123: Why not the following?
```suggestion
@pytest.mark.parametrize("model_builder", TM.get_models_from_module(models.optical_flow))
```

test/test_prototype_models.py
NicolasHug: I think there's value in keeping it: currently the test is skipped because there are no weights, but once I add the weights the test will fail. This is a good thing since it will prevent us from forgetting to add this test again?

datumbox: Sounds good to me.

NicolasHug:
Thanks a lot for the reviews! I'll merge now, and will update these bits as we publish the pretrained weights - PR coming later today.

PR #5044, opened 2021-12-07
Adding logging calls for raft and vit

datumbox:
Unfortunately we missed adding the logger on the constructor of the newest models. Not sure if we can automate the check but until then might be worth building a memory muscle to check in the PR reviews prior merging.

cc @datumbox @sallysyw @kazhang @prabhat00155 @NicolasHug @pmeier @fmassa


NicolasHug (approved):

prabhat00155 (approved):
Thanks @datumbox!

PR #5049, opened 2021-12-07
Adding schema validation for model meta-data

datumbox:
Adding a crude model meta-data validation. Models can define more fields but we define a minimum number of fields per model family.

There are definitely better ways to do this (create classes for meta-data rather than using dictionaries), but until it's clear what we need to store, I propose keeping a lightweight solution.

In this PR I'm also adding a missing `interpolation` meta-field from keypoint. Without this fix, the new test will fail with:
```
  File "./vision/test/test_prototype_models.py", line 115, in test_schema_meta_validation
    assert len(problematic_weights) == 0, str(problematic_weights)
AssertionError: {KeypointRCNN_ResNet50_FPN_Weights.Coco_Legacy: {'interpolation'}, KeypointRCNN_ResNet50_FPN_Weights.Coco_V1: {'interpolation'}}
assert 2 == 0
```

cc @datumbox @bjuncek

### DIFF
--- a/test/test_prototype_models.py
+++ b/test/test_prototype_models.py
@@ -112,7 +112,7 @@ def test_schema_meta_validation(model_fn):
         if missing_fields:
             problematic_weights[w] = missing_fields
 
-    assert len(problematic_weights) == 0, str(problematic_weights)
+    assert not problematic_weights
 
 
 @pytest.mark.parametrize("model_fn", TM.get_models_from_module(models))
--- a/test/test_prototype_models.py
+++ b/test/test_prototype_models.py
@@ -96,7 +96,7 @@ def test_schema_meta_validation(model_fn):
     defaults = {
         "all": ["interpolation", "recipe"],
         "models": classification_fields,
-        "detection": ["categories"],
+        "detection": ["categories", "map"],
         "quantization": classification_fields + ["backend", "quantization", "unquantized"],
         "segmentation": ["categories", "mIoU", "acc"],
         "video": classification_fields,
--- a/test/test_prototype_models.py
+++ b/test/test_prototype_models.py
@@ -83,6 +83,38 @@ def test_naming_conventions(model_fn):
     assert len(weights_enum) == 0 or hasattr(weights_enum, "default")
 
 
+@pytest.mark.parametrize(
+    "model_fn",
+    TM.get_models_from_module(models)
+    + TM.get_models_from_module(models.detection)
+    + TM.get_models_from_module(models.quantization)
+    + TM.get_models_from_module(models.segmentation)
+    + TM.get_models_from_module(models.video),
+)
+def test_schema_meta_validation(model_fn):
+    classification_fields = ["size", "categories", "acc@1", "acc@5"]
+    defaults = {
+        "all": ["interpolation", "recipe"],
+        "models": classification_fields,
+        "detection": ["categories"],
+        "quantization": classification_fields + ["backend", "quantization", "unquantized"],
+        "segmentation": ["categories", "mIoU", "acc"],
+        "video": classification_fields,
+    }
+    module_name = model_fn.__module__.split(".")[-2]
+    fields = set(defaults["all"] + defaults[module_name])
+
+    weights_enum = _get_model_weights(model_fn)
+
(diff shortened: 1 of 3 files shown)


test/test_prototype_models.py
```diff
@@ -83,6 +83,38 @@ def test_naming_conventions(model_fn):
...
+    TM.get_models_from_module(models)
+    + TM.get_models_from_module(models.detection)
+    + TM.get_models_from_module(models.quantization)
+    + TM.get_models_from_module(models.segmentation)
+    + TM.get_models_from_module(models.video),
+)
+def test_schema_meta_validation(model_fn):
+    classification_fields = ["size", "categories", "acc@1", "acc@5"]
+    defaults = {
+        "all": ["interpolation", "recipe"],
```
datumbox, line 97: Though currently the `categories` field is used in all existing models, I haven't put it in the `all` key because of the upcoming addition of optical flow models.

test/test_prototype_models.py
```diff
@@ -83,6 +83,38 @@ def test_naming_conventions(model_fn):
...
+        "models": classification_fields,
+        "detection": ["categories"],
```
datumbox, line 99: `map` not included here because keypoint and maskrcnn have several differently named mAPs. One option is to rename their `box_map` to `map`.

test/test_prototype_models.py
```diff
@@ -83,6 +83,38 @@ def test_naming_conventions(model_fn):
...
+    }
+    module_name = model_fn.__module__.split(".")[-2]
+    fields = set(defaults["all"] + defaults[module_name])
+
+    weights_enum = _get_model_weights(model_fn)
+
+    problematic_weights = {}
+    for w in weights_enum:
+        missing_fields = fields - set(w.meta.keys())
+        if missing_fields:
```
datumbox, line 112: The model is allowed to have more fields than the necessary but it has to have all the required fields.

test/test_prototype_models.py
```diff
@@ -83,6 +83,38 @@ def test_naming_conventions(model_fn):
...
+            problematic_weights[w] = missing_fields
+
+    assert len(problematic_weights) == 0, str(problematic_weights)
```
datumbox, line 115: This is asserted at the end so that we print the full list of problematic weights for each model in the error message.

torchvision/prototype/models/detection/keypoint_rcnn.py
```diff
@@ -22,7 +23,11 @@
 ]
 
 
-_COMMON_META = {"categories": _COCO_PERSON_CATEGORIES, "keypoint_names": _COCO_PERSON_KEYPOINT_NAMES}
+_COMMON_META = {
+    "categories": _COCO_PERSON_CATEGORIES,
+    "keypoint_names": _COCO_PERSON_KEYPOINT_NAMES,
+    "interpolation": InterpolationMode.BILINEAR,
```
datumbox, line 29: Adding missing interpolation field.

datumbox (commented):
Few clarification comments to assist review:

test/test_prototype_models.py
NicolasHug: I think that if you do

```py
assert not problematic_weights
```

pytest would automatically print it in case of an error

test/test_prototype_models.py
NicolasHug: Q: Is the interpolation field related to the resize transform?

NicolasHug (approved):
Thanks @datumbox , LGTM

test/test_prototype_models.py
datumbox: That's correct. It affects resize and affine transforms. You could lose significant accuracy if you use a different interpolation on inference than the one used on training. That's why we put it on the meta-data.

test/test_prototype_models.py
datumbox: You are right, adopted.

test/test_prototype_models.py
datumbox: I ended up renaming `box_map` => `map` and added the field as mandatory for detection models.

PR #5051, opened 2021-12-07
Adding Repeated Augment Sampler 

sallysyw:
This is a SUPER useful technique used in DeiT training, which helps bumping the ViT training accuracy for about 4-5%.
Therefore, we should also add it into pytorch vision reference script.

* Code Reference: [DeiT RA Sampler](https://github.com/facebookresearch/deit/blob/main/samplers.py#L8)

* Results reported from the original [paper](https://arxiv.org/abs/2012.12877):
image

### DIFF
--- a/references/classification/ra_sampler.py
+++ b/references/classification/ra_sampler.py
@@ -27,7 +27,6 @@ class RASampler(torch.utils.data.Sampler):
         self.epoch = 0
         self.num_samples = int(math.ceil(len(self.dataset) * 3.0 / self.num_replicas))
         self.total_size = self.num_samples * self.num_replicas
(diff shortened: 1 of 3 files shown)


datumbox (approved):
LGTM, thanks!

As far as I can see, this is taken from DeIT's [codebase](https://github.com/facebookresearch/deit/blob/main/samplers.py). Since @fmassa is the author of that file and it's from a Meta repo, not sure if we need to put a reference. Francisco let me know if that's not the case.

sallysyw:
I just left a comment inside the file saying this function is borrowed from DeiT repo, please let me know if there's some license issue I need to resolve furthermore @fmassa

PR #4997, opened 2021-11-26
fix mobilenetv3 quantization state dict loading

pmeier:
The failure is currently not visible in CI, since we do not test with loaded weights. See [this CI run](https://app.circleci.com/pipelines/github/pytorch/vision/12903/workflows/ab47f567-64d9-4074-aca5-b7817bb27aff/jobs/1026438) from #4772 for the failure.

PR #5053, opened 2021-12-08
support random seed for RA sampler

xiaohu2015:
This pr is about https://github.com/pytorch/vision/pull/5051
Two modification:
- support random seed for RA sampler
- fix import bug for RA sampler

cc @datumbox

### DIFF
--- a/references/classification/sampler.py
+++ b/references/classification/sampler.py
@@ -15,7 +15,7 @@ class RASampler(torch.utils.data.Sampler):
     https://github.com/facebookresearch/deit/blob/main/samplers.py
     """
 
-    def __init__(self, dataset, num_replicas=None, rank=None, shuffle=True):
+    def __init__(self, dataset, num_replicas=None, rank=None, shuffle=True, seed=0):
         if num_replicas is None:
             if not dist.is_available():
                 raise RuntimeError("Requires distributed package to be available!")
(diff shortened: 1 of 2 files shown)


references/classification/train.py
```diff
@@ -9,7 +9,7 @@
 import torchvision
 import transforms
 import utils
-from references.classification.sampler import RASampler
+from sampler import RASampler
```
datumbox, line 12: Good catch.

references/classification/sampler.py
```diff
@@ -15,7 +15,7 @@ class RASampler(torch.utils.data.Sampler):
     https://github.com/facebookresearch/deit/blob/main/samplers.py
     """
 
-    def __init__(self, dataset, num_replicas=None, rank=None, shuffle=True):
+    def __init__(self, dataset, num_replicas=None, rank=None, shuffle=True, seed=0):
```
datumbox, line 18: I agree this might be useful but it's also a paradigm we don't do on TorchVision. Don't get me wrong, this approach is popular with other libraries it's just that setting seeds on constructors is something we don't typically do.

@fmassa thoughts?

xiaohu2015: @datumbox The RA sampler is modified from torch.utils.data.DistributedSampler which also has a parameter `seed`. The default value of `seed` is 0, so it does not affect the behavior of RA samper if users don't explicitly set a seed.

references/classification/train.py
datumbox: I will need to patch this urgently as it is interfering with an investigation. See https://github.com/pytorch/vision/pull/4997#discussion_r765129140

Thanks for reporting @xiaohu2015

references/classification/sampler.py
datumbox: Fair enough. Since DistributedSampler supports it we can also have it.

datumbox (approved):

PR #5054, opened 2021-12-08
rename TestWeights to appease pytest

### LINKED ISSUE #5050: Pytest can't collect TestWeights
```
test/test_prototype_models.py:175
  /fsx/users/nicolashug/work/vision/test/test_prototype_models.py:175: PytestCollectionWarning: cannot collect test class 'TestWeights' because it has a __init__ constructor (from: test/test_prototype_models.py::TestHandleLegacyInterface)
    class TestWeights(WeightsEnum):
```

I think we just need to rename TestWeights into something that doesn't start with `Test` so that pytest doesn't mistake it for a test class

CC @pmeier 

cc @pmeier

pmeier:
Fixes #5050.

cc @pmeier

### DIFF
--- a/test/test_prototype_models.py
+++ b/test/test_prototype_models.py
@@ -171,7 +171,7 @@ def test_smoke():
 # With this filter, every unexpected warning will be turned into an error
 @pytest.mark.filterwarnings("error")
 class TestHandleLegacyInterface:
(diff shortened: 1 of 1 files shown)


NicolasHug (approved):
thanks @pmeier

PR #5055, opened 2021-12-08
remove extraced archive if flag was set

pmeier:
Without this patch, the `remove_finished` flag is not honored unless we are only decompressing a file rather than extracting it.

cc @pmeier


datumbox (approved):
LGTM, thanks!

Issue #5058, opened 2021-12-08
handle_legacy_interface raises warning when passing weights=SomeString


NicolasHug:
`handle_legacy_interface` raises a warning when passing weights as strings. This leads to warnings in the training references which all rely on strings rather than weight instances.

```py
from torchvision.prototype.models import resnet18, ResNet18_Weights
resnet18(weights="ResNet18_Weights.ImageNet1K_V1")  # warning
resnet18(weights=ResNet18_Weights.ImageNet1K_V1)  # no warning, all good
```

The warning is 

```
 other than a weight enum or `None` for 'weights' are deprecated. The current behavior is equivalent to passing `weights=ResNet18_Weights.ImageNet1K_V1`. You can also use `weights=ResNet18_Weights.default` to get the most up-to-date weights.
```

CC @pmeier @datumbox

PR #5063, opened 2021-12-08
Fix `handle_legacy_interface` when passing strings

NicolasHug:
Hopefully fixes https://github.com/pytorch/vision/issues/5058. This does remove the warning but @pmeier please double check that this doesn't have undesired side effect!

### DIFF
--- a/torchvision/prototype/models/_utils.py
+++ b/torchvision/prototype/models/_utils.py
@@ -43,7 +43,7 @@ def handle_legacy_interface(**weights: Tuple[str, Union[Optional[W], Callable[[D
                 if (
                     (weights_param not in kwargs and pretrained_param not in kwargs)
                     or isinstance(weights_arg, WeightsEnum)
-                    or isinstance(weights_arg, str) and weights_arg != "legacy"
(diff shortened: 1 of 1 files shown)


pmeier (commented):
That might conflict with

https://github.com/pytorch/vision/blob/65438e9eba26951206cbfaafeac1d5b1ac805193/torchvision/prototype/models/detection/keypoint_rcnn.py#L57-L66

Could you check if 

```python
models.detection.keypointrcnn_resnet50_fpn("legacy")
```

and 

```python
models.detection.keypointrcnn_resnet50_fpn(pretrained="legacy")
```

works as expected?

NicolasHug:
Good catch, thanks. Since this seems to be the only case where `pretrained` can be a string, I hardcoded it in the check.

Now both of these raise a warning (instead of failing before the last commit)

```py
keypointrcnn_resnet50_fpn("legacy")
keypointrcnn_resnet50_fpn(pretrained="legacy")
```

pmeier (approved):
LGTM, thanks @NicolasHug!

Issue #5052, opened 2021-12-08
API usage logging within TorchVision


kazhang:
# Goal

To understand TorchVision usage within an organization(e.g. Meta).

The events give insights into torchvision usage with regards to individual callsites, workflows etc. The organization could also learn the trending APIs, which could be used to guide component development/deprecation etc.

# Policy
* Usage should be recorded only once for the same API within a process;
* We should record events as broadly as possible, duplicated events(e.g. module and function log the same thing) is OK and can be dedup in downstream pipelines. 
* For modules, API usage should be recorded at the beginning of constructor of the main class. For example `__init__` of `RegNet`, but not on ones of submodules(e.g. `ResBottleneckBlock`) 
* For functions, API usage should be recorded at the beginning of the method;
* For `torchvision.io`, the logging must be added both on the Python and the C++ (using the csrc submodule as mentioned) side.
* On `torchvision.ops`, the calls should be added both on the main class of the operator (eg StochasticDepth) and on its functional equivalent (eg stochastic_depth) if available. 
* On `torchvision.transforms`, the calls should be placed on the constructors of the Transform classes, the Auto-Augment classes and the functional methods.
* On `torchvision.datasets`, the calls are placed once on the constructor of VisionDataset so we don't need to add them individually on each dataset.
* On `torchvision.utils`, call should be added to the top of each public method.

# Event Format
Full qualified name of the component is logged as the event. For example: `torchvision.models.resnet.ResNet`
Note: for events from C++ APIs, “.csrc” should be added after torchvision, for example: `torchvision.csrc.ops.nms.nms`
# Usage Log API

* C++: `C10_LOG_API_USAGE_ONCE()`
* Python: 
```
from ..utils import _log_api_usage_once
# for class
_log_api_usage_once(self)
# for method
_log_api_usage_once(nms) 
```

Above APIs are lightweight. By default, they are just [no-op](https://github.com/pytorch/pytorch/blob/c236247826bbf49d4f491d97bac6df1da6c1abe8/c10/util/Logging.cpp#L102-L106). It’s guaranteed that the same event is only recorded once within a process. Please note that 8 GPUs will still lead to 8 events, so the events should be dedup by a unique identifier like workflow job_id.

# Also considered
* use decorator
  * doesn’t work with TorchScript since decorator needs to use kwargs, which is not supported in TorchScript;
* use flat namespace
  * there might be name collisions;
* use object or function as param
  * doesn’t work with TorchScript;
* use function’s `__module__`:
[...]

PR #5072, opened 2021-12-08
revamp log api usage method

kazhang:
Change the API to
```
def _log_api_usage_once(module: str, name: str) -> None:
```
to be consistent between modules and functions.
partially-close: #5052

### DIFF
--- a/torchvision/utils.py
+++ b/torchvision/utils.py
@@ -376,12 +376,7 @@ def _generate_color_palette(num_masks: int):
 
 
 @no_type_check
-def _log_api_usage_once(obj: str) -> None:  # type: ignore
+def _log_api_usage_once(module: str, name: str) -> None:
     if torch.jit.is_scripting() or torch.jit.is_tracing():
         return
-    # NOTE: obj can be an object as well, but mocking it here to be
-    # only a string to appease torchscript
-    if isinstance(obj, str):
-        torch._C._log_api_usage_once(obj)
-    else:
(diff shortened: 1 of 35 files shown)


torchvision/utils.py
```diff
@@ -375,13 +375,7 @@ def _generate_color_palette(num_masks: int):
     return [tuple((i * palette) % 255) for i in range(num_masks)]
 
 
-@no_type_check
-def _log_api_usage_once(obj: str) -> None:  # type: ignore
+def _log_api_usage_once(module: str, name: str) -> None:
     if torch.jit.is_scripting() or torch.jit.is_tracing():
```
datumbox, line 379: I wonder if this is still necessary for JIT or if the if statement can be removed?

datumbox (approved):
LGTM, thanks!

@NicolasHug @fmassa Any concerns with the change?

NicolasHug (approved):
Thanks @kazhang

torchvision/utils.py
kazhang: It's still necessary, torch.jit.script complains
> Python builtin <built-in method _log_api_usage_once of PyCapsule object at 0x7f6b4d09a780> is currently not supported in Torchscript:
    torch._C._log_api_usage_once("torchvision.hi")
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~ <--- HERE
    return x

PR #5084, opened 2021-12-10
Allow variable number of repetitions for RA

tbennun:
The number of repetitions in the RA sampler was set to 3. With this PR it can be changed through CLI flags.

Related to #5051.

### DIFF
--- a/references/classification/train.py
+++ b/references/classification/train.py
@@ -174,7 +174,7 @@ def load_data(traindir, valdir, args):
     print("Creating data loaders")
     if args.distributed:
         if args.ra_sampler:
-            train_sampler = RASampler(dataset, shuffle=True)
+            train_sampler = RASampler(dataset, shuffle=True, repetitions=args.ra_reps)
         else:
             train_sampler = torch.utils.data.distributed.DistributedSampler(dataset)
         test_sampler = torch.utils.data.distributed.DistributedSampler(dataset_test, shuffle=False)
@@ -485,7 +485,10 @@ def get_args_parser(add_help=True):
         "--train-crop-size", default=224, type=int, help="the random crop size used for training (default: 224)"
     )
     parser.add_argument("--clip-grad-norm", default=None, type=float, help="the maximum gradient norm (default None)")
-    parser.add_argument("--ra-sampler", action="store_true", help="whether to use ra_sampler in training")
+    parser.add_argument("--ra-sampler", action="store_true", help="whether to use Repeated Augmentation in training")
+    parser.add_argument(
+        "--ra-reps", default=3, type=int, help="number of repetitions for Repeated Augmentation (default: 3)"
+    )
 
     # Prototype models only
     parser.add_argument("--weights", default=None, type=str, help="the weights enum name to load")
--- a/references/classification/sampler.py
+++ b/references/classification/sampler.py
@@ -15,7 +15,7 @@ class RASampler(torch.utils.data.Sampler):
     https://github.com/facebookresearch/deit/blob/main/samplers.py
     """
 
-    def __init__(self, dataset, num_replicas=None, rank=None, shuffle=True, seed=0):
+    def __init__(self, dataset, num_replicas=None, rank=None, shuffle=True, seed=0, repetitions=3):
         if num_replicas is None:
             if not dist.is_available():
                 raise RuntimeError("Requires distributed package to be available!")
@@ -28,11 +28,12 @@ class RASampler(torch.utils.data.Sampler):
         self.num_replicas = num_replicas
         self.rank = rank
         self.epoch = 0
-        self.num_samples = int(math.ceil(len(self.dataset) * 3.0 / self.num_replicas))
+        self.num_samples = int(math.ceil(len(self.dataset) * float(repetitions) / self.num_replicas))
         self.total_size = self.num_samples * self.num_replicas
         self.num_selected_samples = int(math.floor(len(self.dataset) // 256 * 256 / self.num_replicas))
         self.shuffle = shuffle
         self.seed = seed
+        self.repetitions = repetitions
 
     def __iter__(self):
         # Deterministically shuffle based on epoch
@@ -44,7 +45,7 @@ class RASampler(torch.utils.data.Sampler):
             indices = list(range(len(self.dataset)))
 
         # Add extra samples to make it evenly divisible
-        indices = [ele for ele in indices for i in range(3)]
+        indices = [ele for ele in indices for i in range(self.repetitions)]
         indices += indices[: (self.total_size - len(indices))]
         assert len(indices) == self.total_size
 


references/classification/train.py
```diff
@@ -485,7 +485,10 @@ def get_args_parser(add_help=True):
         "--train-crop-size", default=224, type=int, help="the random crop size used for training (default: 224)"
     )
     parser.add_argument("--clip-grad-norm", default=None, type=float, help="the maximum gradient norm (default None)")
-    parser.add_argument("--ra-sampler", action="store_true", help="whether to use ra_sampler in training")
+    parser.add_argument("--ra-sampler", action="store_true", help="whether to use Repeated Augmentation in training")
+    parser.add_argument(
+        "--ra-reps", default=3, type=int, help="number of repetitions for Repeated Augmentation (default: 3)"
```
datumbox, line 490: Thoughts on whether we should have both a boolean and an integer param VS combining the two a single? Other parameters (such as `--clip-grad-norm`, `--auto-augment`) take the latter approach by defining  `default=None` or `default=0` to turn off the feature by default. Though having an extra boolean option is useful when you have many config parameters for the feature, it also creates noise when reading the training log to determine which inputs were actually active for the training.

datumbox (commented):
Thanks @tbennun. Overall it looks a reasonable addition. I've added a question below, let me know your thoughts.

references/classification/train.py
tbennun: I completely agree, though I followed the style of the EMA flag(s). Should I switch to a more general `--repeated-augmentations` with default 0?

datumbox: Thanks for the input. :) When we introduced EMA, we offered 2 parameters `--model-ema-steps` and `--model-ema-decay` that's why we didnt follow this approach. 

@sallysyw Any thoughts on which approach we should choose here?

tbennun: Makes sense. Two things to consider here:
1. Users may not be aware of what is a "good value" for the number of repetitions. This is solvable though: it could be given as an example in the `help` of the argument.
2. In my fork, I'm adding another boolean flag to keep the same epoch length (rather than cutting it down by the number of repetitions) and replicate the samples within the same minibatch (and GPU), following the original Batch Augmentation paper. If it improves generalization more than the DeiT approach, would you be interested in this as part of another PR?

datumbox: Concerning point 1, I agree with your arguments and I don't have strong opinions. I think your PR is mergeable as-is but I'll let @sallysyw who contributed the class to decide whether we should go with 2 parameters or 1.

Concerning point 2, unless the gap in accuracy is massive, I would keep things simple and leave it as is.

tbennun:
@sallysyw @datumbox any news about this PR? I can confirm that with this PR (ra_reps=4) and the latest ResNet-50/ImageNet recipe, there is an improvement in top1 accuracy to 80.802%

datumbox:
@tbennun Sounds good let's merge it. Do you have the log file and checkpoint for the ResNet50 run?

datumbox (approved):

tbennun:
> @tbennun Sounds good let's merge it. Do you have the log file and checkpoint for the ResNet50 run?

@datumbox 

Checkpoints:
* Last (epoch 600): https://polybox.ethz.ch/index.php/s/1y4nRKxNMYRjcRf
* Best (epoch 566, 80.808% Top1): https://polybox.ethz.ch/index.php/s/rlMIvIhpP7nbnt6

The script didn't store a log file, so I only have the last few printouts (would be nice if it did by default :) ). However, it did store all the models from every epoch. I attached above the relevant two (last, best), and here is the log of running with `--test-only` on the last 50 models:  [test.log](https://github.com/pytorch/vision/files/7745364/test.log)


The command line I used for training is as follows (as opposed to the blog, I trained on four GPUs rather than eight, but with the same overall batch size):
```
torchrun --nproc_per_node=4 train.py --model resnet50 --batch-size 256 --lr 0.5 --lr-scheduler cosineannealinglr --lr-warmup-epochs 5 --lr-warmup-method linear --auto-augment ta_wide --epochs 600 --random-erase 0.1 --weight-decay 0.00002 --norm-weight-decay 0.0 --label-smoothing 0.1 --mixup-alpha 0.2 --cutmix-alpha 1.0 --train-crop-size 176 --model-ema --val-resize-size 232 --data-path /path/to/imagenet --cache-dataset --ra-sampler --ra-reps=4
```

I am now testing the mode where a full epoch is used on the same system (training will see 4x samples).

datumbox:
@tbennun Thanks for sharing the checkpoints and the info. This is great!

It's a bummer you don't got the log. We aim for full reproducibility and we normally store the logs and all checkpoints for every model we release. I'm going to fire a new set of training with `--ra-sampler --ra-reps=4` on our side to see if we can reproduce and if we do, I'll be happy to update the checkpoint. We'll work out a way on the PR to credit you as the author (we still try to find out what's the best way to receive such contributions by the community so, bear with me).

datumbox:
@tbennun Happy new year! I was able to verify your findings and get a run which achieves:
```
Acc@1 80.858 Acc@5 95.434
```

Would you be interested sending a PR so that you can be credited for the contribution? Given that you don't have access to deploy the weights, I can do that for you. I propose the following:
1. Write a comment on the issue #3995, briefly describing the modification on top of the [main recipe](https://github.com/pytorch/vision/issues/3995#new-recipe-with-fixres). Make is self sufficient, so that someone who reads it understands what you did to achieve the improvement. Feel free to include papers/references that inspired the change.
2. Send a new PR which modifies the [accuracies](https://github.com/pytorch/vision/blob/588e9b5ed7e2327f9bf5b2136b8c5fd0a85a62d8/torchvision/prototype/models/resnet.py#L102-L103) of the V2 weights to the values I mention above. Moreover change the [recipe link](https://github.com/pytorch/vision/blob/588e9b5ed7e2327f9bf5b2136b8c5fd0a85a62d8/torchvision/prototype/models/resnet.py#L101) to point to your previous comment.
3. Make sure you have the "Allow edits from maintainers" checkbox on the PR checked, so that I can update the URL of the weights once I deploy them on S3.

tbennun:
@datumbox that's great! I'll take care of the comments and the PR. I'm currently running the other version I mentioned before (saving all logs this time :) ), is it ok if we wait with merging the new PR until training completes? This way we can upload a modified version if the accuracy improvement is substantial.

datumbox:
@tbennun Yes, no problem at all. If the accuracy diff is substantial, we can deploy yours. Though in that case, we would need to find a way to send us all the checkpoints and log files associated with the model. This is several hundreds of GB, so it might be a bit tricky but I'm sure we will work it out.

references/classification/train.py
sallysyw: Sorry I missed the discussion before. I think for this specific usage case, having 2 parameters for RA are better than 1 because, as @tbennun mentioned, users might don't have a good sense on which `ra_reps` to use. So it's important to have a non-zero default (also for the backward compatibility of replicating ViT training).

datumbox:
@tbennun Just checking to see if you have any update concerning the training. Thanks!

tbennun:
> @tbennun Just checking to see if you have any update concerning the training. Thanks!

@datumbox yes. It's almost complete (epoch 552 right now) and max top1 accuracy is 81.05% (attained at epoch 513, top5 95.52%), but it's not getting any higher at this point. This is the exact same recipe but with the full BA. I kept the log and all checkpoints as well, but you can judge if this is something worth uploading. What do you think?

datumbox:
@tbennun Sounds good, let's review the full details once the training is completed. Please upload somewhere the best epoch checkpoint once the training is completed along with the log so that we can check. If we do end up using your run, it's going to be a challenge fetching all the logs and checkpoints. Bear with us as we try to work these details out; you are the first person who wants to contribute a half TB run, so that comes with its own challenges. 😄

datumbox:
@tbennun Thanks for the update.

I assume that by nodes you meant the number of available GPUs. Judging from the patch you provided, I understand that each GPU actually runs the full size of the dataset, meaning that for `ra_reps=4` it takes 4x longer to train the model. This is why according to your log, the total number of iters per epoch is 5005 instead of 1252 which is the expected.

Unfortunately I don't think we would like to make such a modification to the RASampler, because we try to keep our implementations canonical. Moreover, I suspect that one could achieve a similar result by running the training for 4x more epochs. On the other hand, your [original proposal](https://github.com/pytorch/vision/pull/5084#issuecomment-997898808) of using the `RASampler` with 4 repetitions is a quite elegant as it keeps the training budget the same while still increasing the accuracy by 0.2 points. We can also use the checkpoint with accuracy `80.858` which will save us from having to exchange 1 TB worth of checkpoints.

If you send a PR with the changes described [here](https://github.com/pytorch/vision/pull/5084#issuecomment-1003565087), I can help you deploy the pre-trained model on PyTorch's S3 bucket and release it. Let me know what you think.

tbennun:
@datumbox agreed on all counts. If that's alright with you, I will create the PR over this weekend. Thanks again for your patience!

tbennun:
@datumbox Done:
Recipe and reference: https://github.com/pytorch/vision/issues/3995#issuecomment-1013906621
PR: #5201

PR #5089, opened 2021-12-13
Fix googlenet and inception docstring for transform_input parameter

NicolasHug:
Same as https://github.com/pytorch/vision/pull/4137 but without merge conflict


pmeier (approved):
Thanks!

Issue #5073, opened 2021-12-09
Update API usage log calls in ops


kazhang:
As we discuss in API usage logging policy, `torchvision.ops` needs to be updated to include calls in all class constructors. Currently there are calls on all methods.
For example, logging is triggered for [roi_align](https://github.com/pytorch/vision/blob/4cacf5a19f68f6b6483c758e3ac95d1dd9b6194c/torchvision/ops/roi_align.py#L53), but we also need log call for [RoIAlign](https://github.com/pytorch/vision/blob/4cacf5a19f68f6b6483c758e3ac95d1dd9b6194c/torchvision/ops/roi_align.py#L78).

datumbox:
@kazhang Thanks for raising issues. If you add more info here to make the ticket self-explanatory without requiring reading other issues and potentially provide an example of how to fix RoiAlign in code, we will be able to tag it as "good first issue" and ask for the help of the community. Usually it takes a bit more effort to write detailed descriptions so that new contributors with no context on Vision can help.

Also tagging a bunch of regulars in case they have the bandwidth to help: @frgfm @oke-aditya @zhiqwang @xiaohu2015 :)

frgfm:
Happy to help :+1: 
But I would need some clarification if possible:
- in the example of `RoIAlign`, considering it calls `roi_align` (cf. https://github.com/pytorch/vision/blob/4cacf5a19f68f6b6483c758e3ac95d1dd9b6194c/torchvision/ops/roi_align.py#L84), do we want to only add an additional log call when `RoIAlign.forward` is used?
- or we want to disable the underlying call from the functional API, and only have a log call for the module API?

xiaohu2015:
@frgfm I think he want to include logging calls in all class constructors for `torchvision.ops`.

datumbox:
Thanks a lot for your support :)

That is correct. The original idea was why include a call on `RoIAlign` if a call exists on `roi_align`? This is a fair point but it leads to a complicated policy where one needs to check if a functional method exists before adding the call on the constructor. So we propose to simplify this by adding always calls in both the Class and the functional ops. Note that not all ops have both Class and functional parts so this way we cover everything.

Note before picking this up, note that @kazhang has an unmerged PR at #5072 which slightly changes the log call. So maybe wait for a few hours until this is merged before you open PRs. He is on a different timezone, so he hasn't seen yet the messages.

xiaohu2015:
[...]

PR #5095, opened 2021-12-13
Simplified usage log API

kazhang:
As we discussed in #5052, we want to:
1. log fully qualified name;
2. use a simple API that takes either class self or function as input;

### DIFF
--- a/torchvision/ops/deform_conv.py
+++ b/torchvision/ops/deform_conv.py
@@ -60,8 +60,8 @@ def deform_conv2d(
         >>> # returns
         >>>  torch.Size([4, 5, 8, 8])
     """
-
-    _log_api_usage_once(deform_conv2d)
+    if not torch.jit.is_scripting() and not torch.jit.is_tracing():
+        _log_api_usage_once(deform_conv2d)
     _assert_has_ops()
     out_channels = weight.shape[0]
 
--- a/torchvision/ops/deform_conv.py
+++ b/torchvision/ops/deform_conv.py
@@ -61,7 +61,7 @@ def deform_conv2d(
         >>>  torch.Size([4, 5, 8, 8])
     """
 
-    _log_api_usage_once("ops", "deform_conv2d")
+    _log_api_usage_once(deform_conv2d)
     _assert_has_ops()
     out_channels = weight.shape[0]
 
--- a/torchvision/ops/deform_conv.py
+++ b/torchvision/ops/deform_conv.py
@@ -124,6 +124,7 @@ class DeformConv2d(nn.Module):
         bias: bool = True,
     ):
         super().__init__()
+        _log_api_usage_once(self)
 
         if in_channels % groups != 0:
             raise ValueError("in_channels must be divisible by groups")
--- a/torchvision/ops/ps_roi_align.py
+++ b/torchvision/ops/ps_roi_align.py
@@ -69,6 +69,7 @@ class PSRoIAlign(nn.Module):
         sampling_ratio: int,
     ):
         super().__init__()
+        _log_api_usage_once(self)
         self.output_size = output_size
         self.spatial_scale = spatial_scale
         self.sampling_ratio = sampling_ratio
--- a/torchvision/ops/ps_roi_align.py
+++ b/torchvision/ops/ps_roi_align.py
@@ -43,7 +43,7 @@ def ps_roi_align(
     Returns:
         Tensor[K, C / (output_size[0] * output_size[1]), output_size[0], output_size[1]]: The pooled RoIs
     """
-    _log_api_usage_once("ops", "ps_roi_align")
+    _log_api_usage_once(ps_roi_align)
     _assert_has_ops()
     check_roi_boxes_shape(boxes)
     rois = boxes
--- a/torchvision/ops/ps_roi_align.py
+++ b/torchvision/ops/ps_roi_align.py
@@ -43,7 +43,8 @@ def ps_roi_align(
     Returns:
         Tensor[K, C / (output_size[0] * output_size[1]), output_size[0], output_size[1]]: The pooled RoIs
     """
-    _log_api_usage_once(ps_roi_align)
+    if not torch.jit.is_scripting() and not torch.jit.is_tracing():
+        _log_api_usage_once(ps_roi_align)
     _assert_has_ops()
     check_roi_boxes_shape(boxes)
     rois = boxes
--- a/torchvision/ops/ps_roi_pool.py
+++ b/torchvision/ops/ps_roi_pool.py
@@ -37,7 +37,8 @@ def ps_roi_pool(
     Returns:
         Tensor[K, C / (output_size[0] * output_size[1]), output_size[0], output_size[1]]: The pooled RoIs.
     """
-    _log_api_usage_once(ps_roi_pool)
+    if not torch.jit.is_scripting() and not torch.jit.is_tracing():
+        _log_api_usage_once(ps_roi_pool)
     _assert_has_ops()
     check_roi_boxes_shape(boxes)
     rois = boxes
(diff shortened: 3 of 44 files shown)


torchvision/ops/deform_conv.py
```diff
@@ -60,8 +60,8 @@ def deform_conv2d(
         >>> # returns
         >>>  torch.Size([4, 5, 8, 8])
     """
-
-    _log_api_usage_once("ops", "deform_conv2d")
+    if not torch.jit.is_scripting() and not torch.jit.is_tracing():
+        _log_api_usage_once(deform_conv2d)
```
datumbox, line 64: This looks good but we are missing the logging call on the `DeformConv2d` class constructor. Can you add it?

This is because when originally logging was added, it wasn't put both on the constructors and on the functional methods. My understanding is that we agreed to put it in both places for simplicity.

Also, should we add the call on the C++ side of the operator as well?

torchvision/ops/ps_roi_align.py
```diff
@@ -43,7 +43,8 @@ def ps_roi_align(
     Returns:
         Tensor[K, C / (output_size[0] * output_size[1]), output_size[0], output_size[1]]: The pooled RoIs
     """
-    _log_api_usage_once("ops", "ps_roi_align")
+    if not torch.jit.is_scripting() and not torch.jit.is_tracing():
+        _log_api_usage_once(ps_roi_align)
```
datumbox, line 47: Same, missing from the class.

torchvision/ops/ps_roi_pool.py
```diff
@@ -37,7 +37,8 @@ def ps_roi_pool(
     Returns:
         Tensor[K, C / (output_size[0] * output_size[1]), output_size[0], output_size[1]]: The pooled RoIs.
     """
-    _log_api_usage_once("ops", "ps_roi_pool")
+    if not torch.jit.is_scripting() and not torch.jit.is_tracing():
+        _log_api_usage_once(ps_roi_pool)
```
datumbox, line 41: Same, missing from the class.

torchvision/ops/roi_align.py
```diff
@@ -50,7 +50,8 @@ def roi_align(
     Returns:
         Tensor[K, C, output_size[0], output_size[1]]: The pooled RoIs.
     """
-    _log_api_usage_once("ops", "roi_align")
+    if not torch.jit.is_scripting() and not torch.jit.is_tracing():
+        _log_api_usage_once(roi_align)
```
datumbox, line 54: Same, missing from the class.

torchvision/ops/roi_pool.py
```diff
@@ -39,7 +39,8 @@ def roi_pool(
     Returns:
         Tensor[K, C, output_size[0], output_size[1]]: The pooled RoIs.
     """
-    _log_api_usage_once("ops", "roi_pool")
+    if not torch.jit.is_scripting() and not torch.jit.is_tracing():
+        _log_api_usage_once(roi_pool)
```
datumbox, line 43: Same, missing from the class.

torchvision/ops/stochastic_depth.py
```diff
@@ -23,7 +23,8 @@ def stochastic_depth(input: Tensor, p: float, mode: str, training: bool = True)
     Returns:
         Tensor[N, ...]: The randomly zeroed tensor.
     """
-    _log_api_usage_once("ops", "stochastic_depth")
+    if not torch.jit.is_scripting() and not torch.jit.is_tracing():
+        _log_api_usage_once(stochastic_depth)
```
datumbox, line 27: Same, missing from the class.

torchvision/utils.py
```diff
@@ -380,7 +385,11 @@ def _generate_color_palette(num_masks: int):
     return [tuple((i * palette) % 255) for i in range(num_masks)]
 
 
-def _log_api_usage_once(module: str, name: str) -> None:
-    if torch.jit.is_scripting() or torch.jit.is_tracing():
+@no_type_check
+def _log_api_usage_once(obj) -> None:  # type: ignore
```
datumbox, line 389: Perhaps?
```suggestion
def _log_api_usage_once(obj: Any) -> None:
```

OR
```suggestion
def _log_api_usage_once(obj):
```

datumbox (commented):
Thanks @kazhang. I think the approach works.

I noted a couple of log calls that were missed that I think are worth adding in this PR to wrap it up.
Also I noticed that a couple of tests are [failing](https://app.circleci.com/pipelines/github/pytorch/vision/13437/workflows/7be0ba67-b306-435f-bf84-4f443c539060/jobs/1076234).

kazhang:
Once this PR is merged, I will update #5073 and #5007.

torchvision/utils.py
```diff
@@ -88,6 +89,7 @@ def norm_range(t, value_range):
         else:
             norm_range(tensor, value_range)
 
+    assert isinstance(tensor, torch.Tensor)
```
kazhang, line 92: Type check complaints about
> Item "List[Tensor]" of "Union[Tensor, List[Tensor]]" has no attribute "size"

Since list of tensors should have been stacked together on line 57, I think this is a fair assertion.

datumbox: Thanks @kazhang, clearly this error is not related to your changes.

@pmeier I tagged you because of the git-blame info. The `tensor: Union[torch.Tensor, List[torch.Tensor]]` is indeed defined as list. My understanding is that this won't work, can you confirm?

datumbox (approved):
Thanks a lot for the work @kazhang. I think this looks much better.

@NicolasHug could you also have a quick look just to be sure?

@vfdev-5 Does it make sense to add log calls to the `_interpolate_bilinear2d_aa` and `_interpolate_bicubic2d_aa` methods on a separate PR?

vfdev-5:
@datumbox I think we can skip logging these methods (_interpolate_bilinear2d_aa, _interpolate_bicubic2d_aa)

torchvision/utils.py
```diff
@@ -380,7 +386,10 @@ def _generate_color_palette(num_masks: int):
...
 
-def _log_api_usage_once(module: str, name: str) -> None:
-    if torch.jit.is_scripting() or torch.jit.is_tracing():
+def _log_api_usage_once(obj: Any) -> None:
+    if not obj.__module__.startswith("torchvision"):
         return
-    torch._C._log_api_usage_once(f"torchvision.{module}.{name}")
+    name = obj.__class__.__name__
+    if name == "function":
+        name = obj.__name__
```
NicolasHug, line 394: Nit: it might be a bit cleaner to explicitly check for functions with either `inspect.isfunction()` or `isinstance(obj, types.FunctionType)`.

NicolasHug (approved):
Thanks @kazhang , one nit (feel free not to address it) but LGTM anyway!

torchvision/utils.py
datumbox: +1 on the `isinstance` approach which avoids importing `inspect`.

torchvision/utils.py
pmeier: Nope, this works as expected: 

https://github.com/pytorch/vision/blob/f5dca445a2a22945a3a34fbf9abe409e23f83fc5/torchvision/utils.py#L54-L56

I look into it an see why `mypy` doesn't pick up on this.

datumbox: Sorry I looked it quickly and missed this. Sounds good.

pmeier: Keeping this open for now. I'm fine with the fix for this PR, but something weird is going on. I can reproduce the issue, but this PR shouldn't do anything to trigger it. I'll investigate more thoroughly.

kazhang: @pmeier just to clarify, should I remove this assertion and ignore the type check error for now? since you'll investigate more thoroughly

pmeier: Keep it in for now. I'll try to figure out how to make `mypy` happy without the `assert` in.

PR #5099, opened 2021-12-14
(temporary) add `pyyaml` to unittest requirements

pmeier:
pytorch/pytorch#68772 added `pyyaml` as dependency and with pytorch/pytorch#68773 it is now a hard dependency when using `torch.jit.load`. Since it is not listed in `torch`'s dependency, we need to install it manually. 

I propose to merge this PR now and revert it as soon as this is resolved upstream.

cc @seemethere @tugsbayasgalan @gmagogsfm

### DIFF
--- a/.circleci/unittest/linux/scripts/environment.yml
+++ b/.circleci/unittest/linux/scripts/environment.yml
@@ -9,6 +9,7 @@ dependencies:
   - libpng
   - jpeg
   - ca-certificates
+  - pyyaml
   - pip:
     - future
     - pillow >=5.3.0, !=8.3.*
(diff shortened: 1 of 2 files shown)


fmassa (approved):
Ok to unblock, but let's not forget to remove it once PyTorch fixes it.

Can you add a comment on each line so that we remember to remove this?

fmassa:
Issue to track this https://github.com/pytorch/pytorch/issues/69905

malfet (approved):

malfet:
Hmm, @pmeier looks like you`ve added `dependency issue` label which is mentioned in `process_commit.py`, why does it still complain about missing labels?

cccclai:
https://github.com/pytorch/pytorch/issues/69905 is resolved. `pyyaml` can be safely removed.

PR #5101, opened 2021-12-15
Fix `draw_bounding_boxes` for tensors on GPU

sumanthratna:
# Background
without this PR, this happens when `image` is on a CUDA device: 
```
  File ".../my_code.py", line 186, in build_graph
    wsi_with_patches = draw_bounding_boxes(wsi, boxes)  # `wsi` is on CUDA device 0
  File ".../venv/lib/python3.8/site-packages/torch/autograd/grad_mode.py", line 28, in decorate_context
    return func(*args, **kwargs)
  File ".../venv/lib/python3.8/site-packages/torchvision/utils.py", line 188, in draw_bounding_boxes
    ndarr = image.permute(1, 2, 0).numpy()
TypeError: can't convert cuda:0 device type tensor to numpy. Use Tensor.cpu() to copy the tensor to host memory first.
```

# Notes
- not detaching the tensor because the function is decorated with `torch.no_grad`

### DIFF
--- a/torchvision/utils.py
+++ b/torchvision/utils.py
@@ -189,7 +189,7 @@ def draw_bounding_boxes(
     if image.size(0) == 1:
         image = torch.tile(image, (3, 1, 1))
 
-    ndarr = image.permute(1, 2, 0).numpy()
+    ndarr = image.permute(1, 2, 0).cpu().numpy()
     img_to_draw = Image.fromarray(ndarr)
 
     img_boxes = boxes.to(torch.int64).tolist()


prabhat00155 (approved):
Thanks @sumanthratna for the fix. I guess this may have only been tested for CPU tensors only. I see a potential for a similar issue in [draw_keypoints()](https://github.com/pytorch/vision/blob/237e0b4bc165674c6fba2d9144cd4d45be455a69/torchvision/utils.py#L349). Do you want to fix that as well?

sumanthratna:
> Thanks @sumanthratna for the fix. I guess this may have only been tested for CPU tensors only. I see a potential for a similar issue in [draw_keypoints()](https://github.com/pytorch/vision/blob/237e0b4bc165674c6fba2d9144cd4d45be455a69/torchvision/utils.py#L349). Do you want to fix that as well?

sure, I can fix that issue—should I do it in a separate PR or add it to this one?

prabhat00155:
> sure, I can fix that issue—should I do it in a separate PR or add it to this one?

Feel free to create a separate PR, thanks!

oke-aditya:
Sorry, but I guess the label for this should be bug fix and not other.

PR #5102, opened 2021-12-16
Fix `draw_keypoints` for tensors on GPU

sumanthratna:
also see https://github.com/pytorch/vision/pull/5101

CC @prabhat00155


prabhat00155 (approved):
Thanks @sumanthratna for fixing this.

oke-aditya:
I think this isn't classified as bug. As this feature isn't released yet. Previously I remember for `masks_to_boxes` we had same issue.

cc @datumbox  and @prabhat00155

NicolasHug:
For such cases I tend to use the "other" label so that we can then bundle such PR with the original one in the release notes

Issue #5075, opened 2021-12-09
[RFC] How should datasets handle decoding of files?


pmeier:
It is a common feature request (for example #4991) to be able to disable the decoding when loading a dataset. To solve this we added a `decoder` keyword argument to the load mechanism (`torchvision.prototype.datasets.load(..., decoder=...)`). It takes an `Optional[Callable]` with the following signature:

```python
def my_decoder(buffer: BinaryIO) -> torch.Tensor: ...
```

If it is a callable, it will be passed a buffer from the dataset and the result will be integrated into the sample dictionary. If the decoder is `None` instead, the buffer will be integrated in the sample dictionary instead, leaving it to the user to decode.

https://github.com/pytorch/vision/blob/4cacf5a19f68f6b6483c758e3ac95d1dd9b6194c/torchvision/prototype/datasets/_builtin/imagenet.py#L132-L134

This works well for images, but already breaks down for videos as discovered in #4838. The issue is that decoding a video results in more information than a single tensor. The tentative plan in #4838 was to change the signature to

```python
def my_decoder(buffer: BinaryIO) -> Dict[str, Any]: ...
```

With this, a decoder can now return arbitrary information, which can be integrated in the top level of the sample dictionary.

Unfortunately, looking ahead, I don't think even this architecture will be sufficient. Two issues came to mind:

1. The current signature assumes that there is only one type of payload to decode in a dataset, i.e. images or videos. Other types, for example annotation files stored as `.mat`, `.xml`, or `.flo`, will always be decoded. Thus, the user can't completely deactivate the decoding after all. Furthermore, they can also not use any custom decoding for these types if need be.
2. The current signature assumes that all payloads of a single type can be decoded by the same decoder. Counter examples to this are the HD1K optical flow datasets that uses 16bit `.png` images as annotations which have sub-par support by `Pillow`.

To overcome this, I propose a new architecture that is similar to the [`RoutedDecoder` datapipe](https://github.com/pytorch/pytorch/blob/d44d59aa70c71091d878e4b0ced963e3074727cd/torch/utils/data/datapipes/iter/routeddecoder.py#L13-L14). We should have a `Decoder` class that has a sequence of `Handler`'s (name up for discussion):

```python
class Decoder:
    def __init__(
        self,
        *handlers: Callable[[str, BinaryIO], Optional[str, Any]],
        must_decode: bool = True,
    ):
        self.handlers = handlers
        self.must_decode = must_decode

    def __call__(
        self,
        path: str,
[...]

PR #5105, opened 2021-12-16
[PoC] separate decoding from datasets

pmeier:
Addresses https://github.com/pytorch/vision/issues/5075#issuecomment-994756308.

With this patch samples will be undecoded by default, but can easily decoded:

```python

encoded_sample = next(iter(datasets.load("caltech101")))
for key, value in sorted(encoded_sample.items()):
    print(f"{key}: {type(value)}")

print("#" * 80)

decoded_sample = datasets.utils.decode_sample(encoded_sample)
for key, value in sorted(decoded_sample.items()):
    print(f"{key}: {type(value)}")
```

```
ann: <class 'torchvision.prototype.datasets.utils._decoder.DecodeableStreamWrapper'>
ann_path: <class 'str'>
image: <class 'torchvision.prototype.datasets.utils._decoder.DecodeableImageStreamWrapper'>
image_path: <class 'str'>
label: <class 'torchvision.prototype.features._label.Label'>
ann_path: <class 'str'>
bounding_box: <class 'torchvision.prototype.features._bounding_box.BoundingBox'>
contour: <class 'torchvision.prototype.features._feature.Feature'>
image: <class 'torchvision.prototype.features._image.Image'>
image_path: <class 'str'>
label: <class 'torchvision.prototype.features._label.Label'>
```

Of course, `decode_sample` can be applied through `.map`

```python

dataset = datasets.load(...).map(datasets.utils.decode_sample)
```

For even more convenience, this also adds a `SampleDecoder` datapipe, that is a thin wrapper around `Mapper` applying `decode_sample`. Although, I generally favor using the class interface, I think this is a case where the functional interface comes in handy, because most users will want to use the default decoders:

```python

dataset = datasets.load(...).decode_samples()
```

@ reviewers: Don't worry about the large diff. I already touched all datasets to see if I missed an edge case in my proposal. That was not the case, so it is sufficient to have a look at `torchvision/prototype/datasets/utils/_decoder.py` and one implementation for example `torchvision/prototype/datasets/_builtin/caltech.py`. I did not yet fix the tests, so it is expected that they are failing. I'll only do that if you are otherwise happy with the proposal.

cc @pmeier @bjuncek


torchvision/prototype/datasets/utils/_decoder.py
```diff
@@ -0,0 +1,47 @@
...
+            decoded_sample.update(obj.decode())
+        else:
+            decoded_sample[name] = obj
+    return decoded_sample
+
+
+@functional_datapipe("decode_samples")
+class SampleDecoder(Mapper[Dict[str, Any]]):
+    def __init__(self, datapipe: IterDataPipe[Dict[str, Any]]) -> None:
+        super().__init__(datapipe, decode_sample)
```
ejguan, line 47: So, how can user provide a custom decoder? Then, do we want users add their own `StreamWrapper` at the end of  each pipeline.

pmeier: The idea of the design is that each `DecodableStreamWrapper` dispatches all attributes to wrapped file object (this is not yet implemented). With that there is no need to provide a mechanism for custom decoders, because they can just use the file directly. Imagine I want to decode an image with my custom decoder, while using the default one for the annotations:

```python

def custom_image_decoder(buffer: BinaryIO) -> Dict[str, Any]:
    ...

def custom_decode_sample(sample):
    sample.update(custom_image_decoder(sample.pop("image")))
    return datasets.utils.decode_sample(sample)


dataset = datasets.load(...).map(custom_decode_sample)
```

ejguan: But, then where this `decode_samples` DataPipe API can be used with custom decoder?

pmeier:
After some more discussion, we want away from supplying custom file handles to the user, but rather will return the encoded data as uint8 tensors. See https://github.com/pytorch/vision/issues/5075#issuecomment-999664152 for details.

With the current implementation, loading data from a dataset now looks like this:

```python

for sample in datasets.load("caltech101").map(datasets.utils.decode_images):
    ...
```

`decode_images` is only a thin wrapper around the workhorse `decode_sample` that sets a default decoder for images. There will be `decode_videos` in the future, but it will probably need additional parameters compared to `decode_images` and thus we cannot unify the two.

In the future we can also provide a transform that handles the decoding, so it can simply be used as first transform in an `Compose`

```python

    transforms.DecodeImages(),
    transforms.Resize(...),
    ...
)

for sample in datasets.load(...).map(transform):
    ...
```

torchvision/prototype/datasets/utils/_decoder.py
pmeier: See https://github.com/pytorch/vision/pull/5105#issuecomment-999673849

pmeier:
Superseded by #5287

PR #5106, opened 2021-12-16
upgrade usort to `1.0.2` and black to 22.3.0

pmeier:
Compared to `usort==0.6.4` which we are currently using, there are two major improvments in `1.0.0`:

1. Multiple `import` statements that import from the same module will be merged into one.
2. Imports inside an `import` statement are sorted lexiographically.

For example

```python
from foo import baz
from foo import spam, ham
from foo import bar
```

will be turned into 

```python
from foo import bar, baz, ham, spam
```

That means we need to pay even less attention to `import`'s while reviewing and would prevent stuff like 

https://github.com/pytorch/vision/blob/582d12a19706b7ddb80a63866727f2b3376a9d9a/torchvision/prototype/datasets/_folder.py#L5-L6

Note that `usort==1.0.0` is not yet released, but it is scheduled for the next weeks. I ran locally with the current preview `1.0.0rc1` and it works as expected. Since we will only touch `import`'s, the noise on `git blame` is minimal.


.pre-commit-config.yaml
```diff
@@ -10,20 +10,13 @@ repos:
         args: [--fix=lf]
       - id: end-of-file-fixer
 
-  # - repo: https://github.com/asottile/pyupgrade
-  #   rev: v2.29.0
-  #   hooks:
-  #     - id: pyupgrade
-  #       args: [--py36-plus]
-  #       name: Upgrade code
-
```
pmeier, line 19: It seems that it was decided in #4585 to not use this hook, but it was kept here as a comment. Was this simply forgotten / overlooked and is safe to be removed?

NicolasHug: we can remove it

NicolasHug:
Thanks @pmeier for the initiative.
For ref for fb employees, this is the corresponding internal post: https://fb.workplace.com/groups/pyfmt/permalink/925830131387837/

I'll have to make sure updating to 1.0 doesn't is compliant with the internal tooling. Since I'll be on PTO in 2 days, let's do all this next year :) ?

NicolasHug (approved):
After a lot of back and forth I think I finally managed to make this consistent with internal linting. There was a minor formatting bug on the internal side that was fixed as well.

The repo is quiet today so it's a good time to merge this. I'll approve and merge, if anything goes wrong by EOD I'll have time to revert. Current CI failure are unrelated

vfdev-5:
> The repo is quiet today so it's a good time to merge this.

I have my branch on proto transforms fully conflicting now )))
No worries, I'll be able to fix it

NicolasHug:
Your best bet is probably to format with the new versions before merging with `main`

PR #5109, opened 2021-12-17
remove md5 for devkit

lepoeme20:
ImageNet no more server md5 information for Development kit.
Also Development kit has not .tar.gz but .tar extension.
url: https://image-net.org/challenges/LSVRC/2012/2012-downloads.php

cc @pmeier

### DIFF
--- a/torchvision/datasets/imagenet.py
+++ b/torchvision/datasets/imagenet.py
@@ -13,7 +13,7 @@ from .utils import check_integrity, extract_archive, verify_str_arg
 ARCHIVE_META = {
     "train": ("ILSVRC2012_img_train.tar", "1d675b47d978889d74fa0da5fadfb00e"),
     "val": ("ILSVRC2012_img_val.tar", "29b22e2961454d5413ddabcf34fc5622"),
-    "devkit": ("ILSVRC2012_devkit_t12.tar.gz", "fa75699e90414af021442c21a62c3abf"),
+    "devkit": ("ILSVRC2012_devkit_t12.tar", None),
 }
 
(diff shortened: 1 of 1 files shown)


pmeier:
Hey @lepoeme20, I can still access the [devkit](https://image-net.org/data/ILSVRC/2012/ILSVRC2012_devkit_t12.tar.gz) with the `.tar.gz` suffixes.

lepoeme20:
@pmeier Thanks to reply, I think it might be OS problem. I tested once again on my Mac, the devkit still has `.tar` suffixes. Even if download the devkit from your link, `.tar.gz` is changed to `.tar` after download is completed (I can see `.tar.gz` suffixes on download status bar). How about md5?? Your devkit has md5?

pmeier:
> How about md5?? Your devkit has md5?

Not sure what you mean by that. Every file has an MD5 since that is only a checksum. I have no access to a Mac, but you should be able to compute it with

```shell
$ md5 ILSVRC2012_devkit_t12.tar.gz
```

Of course the MD5 you compute will be different to what is stored in the dataset, since your devkit seems to be decompressed.

If you have trouble downloading the files through your browser, you can also do it through Python

```python
from torchvision.datasets.utils import download_url
from torchvision.datasets.imagenet import ARCHIVE_META

url = "https://image-net.org/data/ILSVRC/2012/ILSVRC2012_devkit_t12.tar.gz"
md5 = ARCHIVE_META["devkit"][1]

download_url(url, root=".", md5=md5)
```

---

Since this is not a bug on our side, I'm closing the PR.

PR #5110, opened 2021-12-17
Make F.rotate accept learnable params

vfdev-5:
Following https://github.com/pytorch/vision/pull/4995#issuecomment-989676100 and that PR, here is `F.rotate` that should be able to accept differentiable angle and center and also be jit scriptable.


ain-soph:
Thanks a lot! I'll follow your style to work on other methods.

torchvision/transforms/functional.py
```diff
@@ -948,12 +949,48 @@ def _get_inverse_affine_matrix(
...
+
+    i_m_center = torch.eye(3, 3)
+    i_m_center[:2, 2] = -center
+
+    i_m_translate = torch.eye(3, 3)
+    i_m_translate[:2, 2] = -translate
+
+    # RSS without scaling
+    sx, sy = shear_rad[0], shear_rad[1]
+    a = torch.cos(rot - sy) / torch.cos(sy)
```
vadimkantorov, line 971: just wondering, could here be division by zero?

torchvision/transforms/functional.py
```diff
@@ -948,12 +949,48 @@ def _get_inverse_affine_matrix(
...
+    d = -torch.sin(rot - sy) * torch.tan(sx) / torch.cos(sy) + torch.cos(rot)
+
+    output[0, 0] = d
+    output[0, 1] = b
+    output[1, 0] = c
+    output[1, 1] = a
+    output = output / scale
+    output[2, 2] = 1.0
+
+    output = torch.chain_matmul(m_center, output, i_m_center, i_m_translate)
```
vadimkantorov, line 983: I think this `chain_matmul` is going to get deprecated soon in core PyTorch in favor of some similar method in `torch.linalg`

torchvision/transforms/functional.py
vfdev-5: Yes, sy = pi/2 and same for sx=pi/2 with tan. IMO shear of ~90 degrees is too much distortion

vadimkantorov: Maybe best to just protect against division by exact zero, or point it out in docs

torchvision/transforms/functional.py
vfdev-5: Thanks for pointing out that ! We have to use `torch.linalg.multi_dot` instead: https://pytorch.org/docs/stable/generated/torch.chain_matmul.html

torchvision/transforms/functional.py
vfdev-5: Docs however say that values between -180 and 180 are valid. I'm not sure how can we protect from zero devision if user requires 90 degree, like how to evaluate tan(pi/2) without getting infinity. Any ideas ?

ain-soph: I don't think it should be solved in this issue since original non-tensor codes also have this problem. 
Maybe we can open a new issue about this. The potential solution might be just to modify the docs or clip within a range. 90 degree shearing seems meaningless.

vadimkantorov: Yes, if it's not guarded against, the docs or a separate issue would be great.

vfdev-5: @vadimkantorov please create a new issue for this discussion. Thanks

test/test_transforms_tensor.py
```diff
@@ -464,6 +464,7 @@ def test_resized_crop_save(self, tmpdir):
 
 
 def _test_random_affine_helper(device, **kwargs):
+    torch.manual_seed(12)
```
vfdev-5, line 467: I had to add this line and fix shear value due to some random jit vs non-jit results mismatch in one pixel

ain-soph:
Hi, @vfdev-5, sorry for being busy during the past semester. I'm currently available in the following month.
I want to know where we are right now and make contribution to this PR.

Is there any problem to merge the current PR? If not, I'll draft the PR for other operators.

ain-soph:
@datumbox (Maybe a question not that relevant) I'm currently doing internship at Meta (local0state@fb.com), will that help for making contribution and communicate ideas for the issue?

The basic idea for this PR is to wrap parameters with `torch.Tensor` instead of plain `float` to allow gradient computation w.r.t. those parameters, so I personally don't see anything that breaks backward compatibility yet.

datumbox:
@ain-soph Thanks for the feedback! I've pinged you via the internal chat but for transparency and to keep the community up-to-date I'll summarize why I proposed to wait a little bit.

The current prototype implementation merged on main branch doesn't have the `F` layer. This high-level layer of functional transforms, which is also the public interface for the users, connects the low-level kernels (PIL, Tensor). In the original new proposal, this layer was removed in favour of exposing directly the low-level kernels. This was done because the new solution uses Tensor subclassing that is NOT JIT-scriptable. So merging the PR means we are doing changes on an API that could be deprecated. Nevertheless, we are currently examining introducing it back to resolve the verbosity of the proposed API on the Transforms. @vfdev-5 has actually an open PR for that at #6219. So waiting a little bit longer to help us finalize the API will allow us to make better decisions. This doesn't mean that we ignore your feature request, just postpone the merge for a bit until the API proposal is finalized. Let me know if you have more concerns, happy to schedule a chat.

vfdev-5:
@datumbox at first we can apply this PR's logic to proto low-level ops, e.g. `F.rotate` -> `F.rotate_image_tensor`. Next, we can try to apply that to proto's `F.rotate`. Here i tried to apply that for rotate op: https://github.com/pytorch/vision/compare/main...vfdev-5:vision:proto-learnable-params
I still have some test issues, but I think this can be fixed.

ain-soph:
@vfdev-5 @datumbox  Would we reconsider this PR since Transfroms V2 API is out?

Issue #5108, opened 2021-12-17
New classification datasets support for FLAVA


NicolasHug:
To support our colleagues' work on the [FLAVA](https://arxiv.org/abs/2112.04482) paper, and to foster collaborations in the multi-modal space, we would like to implement a few new datasets. Almost all of them are classification datasets but some also support other tasks like segmentation.

- [x] [Food 101](https://data.vision.ee.ethz.ch/cvl/datasets_extra/food-101/) @jdsgomes #5119
- [x] [Stanford Cars](http://ai.stanford.edu/~jkrause/cars/car_dataset.html) @abhi-glitchhg #5166
- [x] [FGVC Aircraft](https://www.robots.ox.ac.uk/~vgg/data/fgvc-aircraft/) @sallysyw #5178
- [x] [DTD](https://www.robots.ox.ac.uk/~vgg/data/dtd/). A good starting point is [this PR](https://github.com/pytorch/vision/pull/743) from @pmeier #5115 
- [x] [Oxford Pets](https://www.robots.ox.ac.uk/~vgg/data/pets/). This one also comes with ROIs and segmentation masks, which would be nice to support. We could do something similar to [CelebA](https://github.com/pytorch/vision/blob/main/torchvision/datasets/celeba.py) with a `target_type` parameter. @pmeier #5116 
- [x] [Flowers-102](https://www.robots.ox.ac.uk/~vgg/data/flowers/102/index.html). @zhiqwang #5177 
- [x] [EuroSAT](https://github.com/phelber/eurosat) @frgfm #5114
- [x] [GSTRB](https://paperswithcode.com/dataset/gtsrb). The homepage is timing out for me, but download links can be found [here](https://github.com/facebookresearch/vissl/blob/484cdecd1a71cb457d8ea74942603b907a23d39d/extra_scripts/datasets/create_gtsrb_data_files.py#L43-L53) @sumukhaithal6 #5117
- [X] [PCAM](https://github.com/basveeling/pcam) @NicolasHug https://github.com/pytorch/vision/pull/5203
- [x] [Clevr Counts](https://cs.stanford.edu/people/jcjohns/clevr/). See also [here](https://github.com/facebookresearch/vissl/blob/484cdecd1a71cb457d8ea74942603b907a23d39d/extra_scripts/datasets/create_clevr_count_data_files.py#L53-L54) for what we exactly need @pmeier #5130
[...]

PR #5114, opened 2021-12-19
Adds EuroSAT to the list of supported datasets

frgfm:
Following up on #5108, this PR introduces the following modifications:
- adds the EuroSAT dataset to the list of supported datasets (please note that there is no train/test split)
- adds a corresponding unittest
- adds the corresponding entry in the documentation

Any feedback is welcome!

### DIFF
--- a/test/test_datasets.py
+++ b/test/test_datasets.py
@@ -2168,5 +2168,23 @@ class HD1KTestCase(KittiFlowTestCase):
         return num_sequences * (num_examples_per_sequence - 1)
 
 
+class EuroSATTestCase(datasets_utils.ImageDatasetTestCase):
+    DATASET_CLASS = datasets.EuroSAT
+
+    def inject_fake_data(self, tmpdir, config):
+        tmpdir = pathlib.Path(tmpdir)
+
+        wnid = "AnnualCrop"
+        num_examples = 3
+        datasets_utils.create_image_folder(
+            root=tmpdir,
+            name=tmpdir / wnid / wnid,
+            file_name_fn=lambda image_idx: f"{wnid}_{image_idx}.JPEG",
+            num_examples=num_examples,
+        )
+
+        return num_examples
+
+
 if __name__ == "__main__":
     unittest.main()
--- a/torchvision/datasets/eurosat.py
+++ b/torchvision/datasets/eurosat.py
@@ -1,0 +1,76 @@
+import os
+from typing import Any
+
+from .folder import ImageFolder
+from .utils import download_and_extract_archive, check_integrity
+
+
+class EuroSAT(ImageFolder):
+    """`EuroSAT <https://arxiv.org/pdf/1709.00029.pdf>`_ Dataset.
+
+    Args:
+        root (string): Root directory of dataset where ``EuroSAT.zip`` exists.
+        download (bool, optional): If True, downloads the dataset from the internet and
+            puts it in root directory. If dataset is already downloaded, it is not
+            downloaded again.
+        transform (callable, optional): A function/transform that  takes in an PIL image
+            and returns a transformed version. E.g, ``transforms.RandomCrop``
+        target_transform (callable, optional): A function/transform that takes in the
+            target and transforms it.
+    """
+
+    url = "https://madm.dfki.de/files/sentinel/EuroSAT.zip"
+    md5 = "c8fa014336c82ac7804f0398fcb19387"
+    filename = "EuroSAT.zip"
+
+    _classes = [
+        "Annual Crop",
+        "Forest",
+        "Herbaceous Vegetation",
+        "Highway",
+        "Industrial Buildings",
+        "Pasture",
+        "Permanent Crop",
+        "Residential Buildings",
+        "River",
+        "Sea & Lake",
+    ]
+
+    def __init__(
+        self,
+        root: str,
+        download: bool = False,
+        **kwargs: Any,
+    ) -> None:
+        self.root = os.path.expanduser(root)
+
+        # Download
+        if download:
+            self.download()
+        if not self._check_exists():
+            raise RuntimeError("Dataset not found. You can use download=True to download it")
+
+        # ImageFolder
+        super().__init__(os.path.join(self.data_folder, "2750"), **kwargs)
+        self.classes = self._classes
+        self.class_to_idx = {_class: i for i, _class in enumerate(self.classes)}
+
+    def __len__(self) -> int:
+        return len(self.data)
+
+    @property
+    def data_folder(self) -> str:
+        return os.path.join(self.root, self.__class__.__name__)
+
+    def _check_exists(self) -> bool:
+        return check_integrity(os.path.join(self.data_folder, self.filename))
+
+    def download(self) -> None:
+        """Download the EuroSAT data if it doesn't exist already."""
+
+        if self._check_exists():
+            return
+
+        os.makedirs(self.data_folder, exist_ok=True)
+        print(f"Downloading {self.url}")
+        download_and_extract_archive(self.url, download_root=self.data_folder, filename=self.filename, md5=self.md5)
--- a/torchvision/datasets/eurosat.py
+++ b/torchvision/datasets/eurosat.py
@@ -6,7 +6,7 @@ from .utils import download_and_extract_archive, check_integrity
 
 
 class EuroSAT(ImageFolder):
-    """`EuroSAT <https://arxiv.org/pdf/1709.00029.pdf>`_ Dataset.
+    """RGB version of the `EuroSAT <https://arxiv.org/pdf/1709.00029.pdf>`_ Dataset.
 
     Args:
         root (string): Root directory of dataset where ``EuroSAT.zip`` exists.
@@ -44,9 +44,10 @@ class EuroSAT(ImageFolder):
     ) -> None:
         self.root = os.path.expanduser(root)
 
-        # Download
+        # Download the dataset
         if download:
             self.download()
+
         if not self._check_exists():
             raise RuntimeError("Dataset not found. You can use download=True to download it")
 
--- a/docs/source/datasets.rst
+++ b/docs/source/datasets.rst
@@ -39,6 +39,7 @@ You can also create your own datasets using the provided :ref:`base classes <bas
     CocoCaptions
     CocoDetection
     EMNIST
+    EuroSAT
     FakeData
     FashionMNIST
     Flickr8k
--- a/torchvision/datasets/__init__.py
+++ b/torchvision/datasets/__init__.py
@@ -4,6 +4,7 @@ from .celeba import CelebA
 from .cifar import CIFAR10, CIFAR100
 from .cityscapes import Cityscapes
 from .coco import CocoCaptions, CocoDetection
+from .eurosat import EuroSAT
 from .fakedata import FakeData
 from .flickr import Flickr8k, Flickr30k
 from .folder import ImageFolder, DatasetFolder
@@ -77,4 +78,5 @@ __all__ = (
     "FlyingChairs",
     "FlyingThings3D",
     "HD1K",
+    "EuroSAT",
 )


test/test_datasets.py
```diff
@@ -2168,5 +2168,23 @@ def inject_fake_data(self, tmpdir, config):
         return num_sequences * (num_examples_per_sequence - 1)
 
 
+class EuroSATTestCase(datasets_utils.ImageDatasetTestCase):
+    DATASET_CLASS = datasets.EuroSAT
+
+    def inject_fake_data(self, tmpdir, config):
+        tmpdir = pathlib.Path(tmpdir)
+
+        wnid = "AnnualCrop"
```
pmeier, line 2177: "wnid" is specific to `ImageNet` meaning "WordNet ID". I think you mean "category", right?

test/test_datasets.py
```diff
@@ -2168,5 +2168,23 @@ def inject_fake_data(self, tmpdir, config):
...
+        num_examples = 3
+        datasets_utils.create_image_folder(
```
pmeier, line 2179: The implementation hard codes a `"2750"` folder. Don't we need that here?

torchvision/datasets/eurosat.py
```diff
@@ -0,0 +1,77 @@
...
+        self,
+        root: str,
+        download: bool = False,
+        **kwargs: Any,
+    ) -> None:
+        self.root = os.path.expanduser(root)
+
+        # Download the dataset
+        if download:
+            self.download()
```
pmeier, line 49: That should be self explanatory

```suggestion
        if download:
            self.download()
```

torchvision/datasets/eurosat.py
```diff
@@ -0,0 +1,77 @@
...
+
+        if not self._check_exists():
+            raise RuntimeError("Dataset not found. You can use download=True to download it")
+
+        # ImageFolder
```
pmeier, line 54:
```suggestion
```

torchvision/datasets/eurosat.py
```diff
@@ -0,0 +1,77 @@
...
+        super().__init__(os.path.join(self.data_folder, "2750"), **kwargs)
+        self.classes = self._classes
+        self.class_to_idx = {_class: i for i, _class in enumerate(self.classes)}
```
pmeier, line 57: Why are you setting them manually? Isn't `ImageFolder` doing that automatically for us?

torchvision/datasets/eurosat.py
```diff
@@ -0,0 +1,77 @@
...
+
+        # Download the dataset
+        if download:
+            self.download()
+
+        if not self._check_exists():
+            raise RuntimeError("Dataset not found. You can use download=True to download it")
+
+        # ImageFolder
+        super().__init__(os.path.join(self.data_folder, "2750"), **kwargs)
```
pmeier, line 55: Maybe we should overwrite `root` again after the `ImageFolder` constructor, because otherwise

```python
root = ...
dataset = EuroSAT(root)
assert dataset.root != root
```

torchvision/datasets/eurosat.py
```diff
@@ -0,0 +1,77 @@
...
+        self.classes = self._classes
+        self.class_to_idx = {_class: i for i, _class in enumerate(self.classes)}
+
+    def __len__(self) -> int:
+        return len(self.data)
+
+    @property
+    def data_folder(self) -> str:
+        return os.path.join(self.root, self.__class__.__name__)
```
pmeier, line 64: Nit: the prototype datasets use all lower names for the folders. Using the same here makes it easier to port later

```suggestion
        return os.path.join(self.root, self.__class__.__name__.lower())
```

torchvision/datasets/eurosat.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def _check_exists(self) -> bool:
+        return check_integrity(os.path.join(self.data_folder, self.filename))
+
+    def download(self) -> None:
+        """Download the EuroSAT data if it doesn't exist already."""
+
```
pmeier, line 71: We are not going to expose this to the documentation and thus we don't need a docstring here.

```suggestion
```

pmeier (commented):
Thanks a lot @frgfm for following up so quickly. The PR looks good. I only have some minor suggestions and a few questions inline.

test/test_datasets.py
frgfm: indeed

test/test_datasets.py
frgfm: good catch (I wasn't sure how to use the inherited test case)

torchvision/datasets/eurosat.py
frgfm: It's nitpicking but the class names are not complete with the folder names, so I override them to have full class names ("Industrial" --> "Industrial Buildings", etc.)

torchvision/datasets/eurosat.py
frgfm: sure!

torchvision/datasets/eurosat.py
pmeier: So "Industrial Building" is the identifier used in the paper, but the folder is only named "Industrial"? In that case, I think it is ok to overwrite them.  propose you simply turn `_classes` into a dictionary and do `self.classes = [self._class_map[cls] for cls in self.classes]`. That way the intention is more clear and we also don't need to overwrite the indices.

frgfm: Correct, I'll make the changes :ok_hand:

torchvision/datasets/eurosat.py
```diff
@@ -0,0 +1,74 @@
...
+            downloaded again.
+        transform (callable, optional): A function/transform that  takes in an PIL image
+            and returns a transformed version. E.g, ``transforms.RandomCrop``
+        target_transform (callable, optional): A function/transform that takes in the
+            target and transforms it.
+    """
+
+    url = "https://madm.dfki.de/files/sentinel/EuroSAT.zip"
+    md5 = "c8fa014336c82ac7804f0398fcb19387"
+    filename = "EuroSAT.zip"
```
pmeier, line 24: Given that the file name matches the end of the url, we don't need it. The download functionality is able to do that automatically.
```suggestion
```

torchvision/datasets/eurosat.py
```diff
@@ -0,0 +1,74 @@
...
+        return check_integrity(os.path.join(self.data_folder, self.filename))
+
+    def download(self) -> None:
+
+        if self._check_exists():
+            return
+
+        os.makedirs(self.data_folder, exist_ok=True)
+        print(f"Downloading {self.url}")
+        download_and_extract_archive(self.url, download_root=self.data_folder, filename=self.filename, md5=self.md5)
```
pmeier, line 74: See above.

```suggestion
        download_and_extract_archive(self.url, download_root=self.data_folder, md5=self.md5)
```

pmeier (commented):
Looking good. A small nits inline. Other than that, I propose we use at least a somewhat standardized naming scheme for the folders. I would use `base_folder` as the root folder plus the dataset name and `data_folder` as the folder that gets extracted from the archive. For an example, have a look at #5115.

torchvision/datasets/eurosat.py
```diff
@@ -0,0 +1,74 @@
...
+
+    url = "https://madm.dfki.de/files/sentinel/EuroSAT.zip"
+    md5 = "c8fa014336c82ac7804f0398fcb19387"
+    filename = "EuroSAT.zip"
+
+    _class_map = {
+        "AnnualCrop": "Annual Crop",
+        "Forest": "Forest",
+        "HerbaceousVegetation": "Herbaceous Vegetation",
+        "Highway": "Highway",
```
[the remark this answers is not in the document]
frgfm: I hesitated to do so, but I thought that if in the future someone renames one folder by mistake, this will catch it automatically. I'll apply your suggestions

pmeier (approved):
LGTM, thanks a lot @frgfm!

torchvision/datasets/eurosat.py
```diff
@@ -0,0 +1,72 @@
+import os
+from typing import Any
+
+from .folder import ImageFolder
+from .utils import download_and_extract_archive
+
+
+class EuroSAT(ImageFolder):
+    """RGB version of the `EuroSAT <https://arxiv.org/pdf/1709.00029.pdf>`_ Dataset.
```
NicolasHug, line 9: I think it's best to link to the GitHub repo instead, which seems to be the "official" landing point for the dataset. The paper is linked from there.

```suggestion
    """RGB version of the `EuroSAT <https://github.com/phelber/eurosat>`_ Dataset.
```

torchvision/datasets/eurosat.py
```diff
@@ -0,0 +1,72 @@
...
+
+    Args:
+        root (string): Root directory of dataset where ``EuroSAT.zip`` exists.
```
NicolasHug, line 12:
```suggestion
        root (string): Root directory of dataset where ``root/eurosat`` exists.
```

torchvision/datasets/eurosat.py
```diff
@@ -0,0 +1,72 @@
...
+        download (bool, optional): If True, downloads the dataset from the internet and
+            puts it in root directory. If dataset is already downloaded, it is not
+            downloaded again.
```
NicolasHug, line 15:
```suggestion
            downloaded again. Default is False.
```

torchvision/datasets/eurosat.py
```diff
@@ -0,0 +1,72 @@
...
+
+    @property
+    def _base_folder(self) -> str:
+        return os.path.join(self.root, self.__class__.__name__.lower())
+
+    @property
+    def _data_folder(self) -> str:
+        return os.path.join(self._base_folder, "2750")
+
+    def _check_exists(self) -> bool:
```
NicolasHug, line 63: Since this is only used in one place, let's just inline it in `download()`

torchvision/datasets/eurosat.py
```diff
@@ -0,0 +1,72 @@
...
+    def __len__(self) -> int:
+        return len(self.samples)
+
+    @property
+    def _base_folder(self) -> str:
+        return os.path.join(self.root, self.__class__.__name__.lower())
+
+    @property
+    def _data_folder(self) -> str:
+        return os.path.join(self._base_folder, "2750")
```
NicolasHug, line 61: I think we could just have these 2 `@property`s as attributes declared in `__init__()` instead.

This would be more consistent with the rest of the code-base, and less verbose

torchvision/datasets/eurosat.py
```diff
@@ -0,0 +1,72 @@
...
+        super().__init__(self._data_folder, **kwargs)
+        self.classes = [self._class_map.get(cls, cls) for cls in self.classes]
+        self.root = os.path.expanduser(root)
+
+    def __len__(self) -> int:
+        return len(self.samples)
+
+    @property
+    def _base_folder(self) -> str:
+        return os.path.join(self.root, self.__class__.__name__.lower())
```
NicolasHug, line 57:
```suggestion
        return os.path.join(self.root, "eurosat")
```

torchvision/datasets/eurosat.py
```diff
@@ -0,0 +1,72 @@
...
+    md5 = "c8fa014336c82ac7804f0398fcb19387"
+
+    _class_map = {
+        "AnnualCrop": "Annual Crop",
+        "HerbaceousVegetation": "Herbaceous Vegetation",
+        "Industrial": "Industrial Buildings",
+        "PermanentCrop": "Permanent Crop",
+        "Residential": "Residential Buildings",
+        "SeaLake": "Sea & Lake",
+    }
```
NicolasHug, line 32: Sorry, this has possibly been discussed already, but what's the reason for changing the class names? At first sight I would say we should stick to the initial names?

NicolasHug (commented):
Thanks a lot @frgfm for the PR and @pmeier for the review.

I have a few comments / questions below. I understand it's been 2 weeks since the last approval already, so I'm happy to address these myself if you'd like. Let me know!

torchvision/datasets/eurosat.py
pmeier: It is also used in `__init__` to have a check if `download=False` is passed.

torchvision/datasets/eurosat.py
pmeier: See https://github.com/pytorch/vision/pull/5114#discussion_r771967671.

NicolasHug: Thanks for the ref.

I would prefer sticking to the author's implementation. There seem to be little value in changing the names, and any deviation from the the original folder names is a potential source of confusion for users.

torchvision/datasets/eurosat.py
frgfm: Not sure I get that one? Is it to ensure compatibility if we change the name of the class later on?
Because this string is exactly was it yields

torchvision/datasets/eurosat.py
frgfm: I would argue that the author's will is expressed in more detail in the paper, and though the missing spaces are not an issue for understanding, the truncated class names could be. Rather than a deviation, I'd say it's more an addition to stick closer to the paper, which is more comprehensive :ok_hand: 

But if you prefer to change that, let me know!

torchvision/datasets/eurosat.py
pmeier: I think Nicolas' rationale is that it is harder parse than just the explicit version.

torchvision/datasets/eurosat.py
pmeier: I agree with @frgfm here. Ideally, a user should be able to use our interface to the data without ever having seen the raw data. Thus, I would prefer to stick closer to the paper rather than the data. 

In general, the classes should be human readable. Of course `SeaLake` is still readable, but `Sea & Lake` conveys the meaning better. In most cases these changes only are made since the classes are also folder names and spaces or symbols shouldn't be part of them.

NicolasHug: @pmeier are there other instances where we have changed the class names for similar reasons?

In COCO, we have kept names such as `pottedplant`, `tvmonitor`, or `diningtable`.

Personally I don't think the current names are confusing, nor ambiguous: AnnualCrop  Forest  HerbaceousVegetation  Highway  Industrial  Pasture	PermanentCrop  Residential  River  SeaLake

There are cases where users need the path of the files they use. This is not supported in torchvision but this has been asked various times and this is a typical use-case for e.g. submitting to datasets leaderboards. So often, users will hack around the class to return the path and do what they want to do. Any discrepancy between the class names and the path is a potential source of confusion for the user here.

I understand it may sound overcautious (and maybe it is), but every decision that we make introduces risks. I see no risk in keeping the current class names, but I believe there are potential pitfalls with changing them, even if we haven't clearly identified them yet.

pmeier: > are there other instances where we have changed the class names for similar reasons?

I don't recall an instance of the top of my head.

> Personally I don't think the current names are confusing, nor ambiguous

If it is just a space or a symbol that is missing, I agree. Note that there are 2 categories that drop the word "Building":

- "Industrial Building" -> "Industrial"
- "Residental Building" -> "Residental"

I'm not too familiar with the data, but I'd wager a guess this could also be confusing.

> Returning paths is possibly something that we'll support in the future in the new datasets, so I want to avoid setting a precedent of deviating from that here.

The prototype datasets already do that. There are two distinct features usually called "category" and "path" that may or may not carry the same information (of course the will always correspond to the same information). 

---

Anyway, as @frgfm already stated in https://github.com/pytorch/vision/pull/5114#discussion_r771972043, this is probably is nitpicking. If you want to have your way, I'm not going to oppose it.

pmeier:
@frgfm the test failures are related. Could you have a look?

frgfm:
> @frgfm the test failures are related. Could you have a look?

Oops, I overlooked that one. This should be fixed now :+1:

torchvision/datasets/eurosat.py
pmeier: @NicolasHug and I discussed more offline. We should go with the safest option here and solve this with the prototype datasets. @frgfm could you remove the name mapping? After that (and the merge conflicts), the PR should be good to go.

frgfm: Oops my bad, I think I check the notification right before the last reply of Philip!
I just updated the PR accordingly :+1:

pmeier (approved):
Thanks a lot for sticking to it @frgfm! LGTM if CI is green.

NicolasHug (approved):
Thanks a lot @frgfm and @pmeier !

PR #743, opened 2019-02-13
Added the DTD dataset

pmeier:
This adds the [DTD dataset](https://www.robots.ox.ac.uk/~vgg/data/dtd/). At its core it is a `ImageFolder` dataset. Next to a split into `train`, `val` and `test` sets, it contains 10 predefined `fold`s of each partial set.

I've sub-classed `ImageFolder` class and added support to download it from scratch. I can verify the downloaded archive, but I couldn't think of a way to do this for the extracted content. Thus, the `download` flag defaults to `False`.

The index of each image within `self.imgs` depends on `split` and `fold`. I've implemented a converter, which converts the indices between the complete and partial sets. Since the images are only loaded at runtime, `split` and `fold` can also be changed at runtime, which triggers a re-initialization of the converter.

PR #5115, opened 2021-12-19
Add DTD dataset

pmeier:
Addresses #5108 and supersedes #743. I've also added the corresponding prototype dataset.

cc @pmeier @bjuncek


torchvision/datasets/dtd.py
```diff
@@ -0,0 +1,86 @@
...
+        self._image_files = []
+        classes = []
+        with open(self._meta_folder / f"{self._split}{self._fold}.txt") as file:
+            for line in file:
+                cls, name = line.strip().split("/")
+                self._image_files.append(self._images_folder.joinpath(cls, name))
+                classes.append(cls)
+
+        self.classes = sorted(set(classes))
+        self.class_to_idx = dict(zip(self.classes, range(len(self.classes))))
```
NicolasHug, line 59: Do you know why some of our datasets expose `self.classes` while others expose `self.labels`?

Also, what's the reasons for making these public? Should we document them?

torchvision/datasets/dtd.py
```diff
@@ -0,0 +1,86 @@
...
+    """
+
+    _URL = "https://www.robots.ox.ac.uk/~vgg/data/dtd/download/dtd-r1.0.1.tar.gz"
+    _MD5 = "fff73e5086ae6bdbea199a49dfb8a4c1"
+
+    def __init__(
+        self,
+        root: str,
+        split: str = "train",
+        fold: Union[str, int] = 1,
```
NicolasHug, line 30: The 10 splits seem to be rather specific to the paper. It's great to have this parameter for reproducibility, but I wonder if we should allow an "all" option here, and perhaps it could be the default?

torchvision/prototype/datasets/_builtin/dtd.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def _classify_archive(self, data: Tuple[str, Any]) -> Optional[int]:
+        path = pathlib.Path(data[0])
+        if path.parent.name == "labels":
+            if path.name == "labels_joint_anno.txt":
+                return 1
+
+            return 0
+        elif path.parents[1].name == "images":
+            return 2
```
NicolasHug, line 61: instead of hardcoded 0 1 2, would a small private enum be overkill here? or maybe just hardcoded named constants? No strong opinion but this might help readability

NicolasHug (commented):
Thanks @pmeier , this looks great. I made some minor comment, the only one I'm still not sure about is whether we should allow `fold="all"`. LMK what you think

test/test_datasets.py
```diff
@@ -2168,5 +2168,42 @@ def inject_fake_data(self, tmpdir, config):
...
+    FEATURE_TYPES = (PIL.Image.Image, int)
+
+    ADDITIONAL_CONFIGS = datasets_utils.combinations_grid(
+        split=("train", "test", "val"),
+        # There is no need to test the whole matrix here, since each fold is treated exactly the same
+        fold=(5,),
+    )
+
+    def inject_fake_data(self, tmpdir: str, config):
+        data_folder = os.path.join(tmpdir, "dtd", "dtd")
```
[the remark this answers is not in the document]
pmeier: No idea TBH :D

torchvision/datasets/dtd.py
pmeier: IIUC, each fold contains all images across the splits. So if we want all images, we should adapt the `split` parameter.

torchvision/datasets/dtd.py
pmeier: > Do you know why some of our datasets expose `self.classes` while others expose `self.labels`?

They are two different things: `self.classes` is a list of all class names whereas `self.labels` is a list of integers for every sample. Almost every dataset should have something similar to `self.labels`, while `self.classes` is not always available depending on the nature of the dataset.

> Also, what's the reasons for making these public?

Force of habit I guess. Most datasets that have these attributes also have them public. 

> Should we document them?

Given that these are the last additions to a (hopefully) soon to be legacy API, I wouldn't sweat it. For most other datasets they are also undocumented.

torchvision/prototype/datasets/_builtin/dtd.py
pmeier: Yes, that could work. If that is ok with you, I'll put it in my backlog, because we should do that for all datasets that use a `Demultiplexer`.

PR #5118, opened 2021-12-21
Fix missing Torch includes

peterbell10:
The `torchvision` build currently relies on transitive includes through unrelated `torch` headers which is causing build failures in pytorch/pytorch#68693.

### DIFF
--- a/torchvision/csrc/io/image/image.cpp
+++ b/torchvision/csrc/io/image/image.cpp
@@ -1,7 +1,7 @@
 #include "image.h"
 
+#include <ATen/core/op_registration/op_registration.h>
 #include <Python.h>
-#include <torch/library.h>
 
 // If we are in a Windows environment, we need to define
 // initialization functions for the _custom_ops extension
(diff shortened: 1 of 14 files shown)


datumbox:
@peterbell10 Thanks for the PR. 

We are intentionally trying to be more selective on our includes. Can you adapt the PR to explicitly import those relevant missed includes rather than just importing `torch/library.h`?

torchvision/csrc/io/image/image.cpp
```diff
@@ -1,6 +1,7 @@
 #include "image.h"
 
 #include <Python.h>
+#include <torch/library.h>
```
datumbox, line 4: I assume this is for `RegisterOperators`. Shouldn't we explicitly import the following instead?
```suggestion
#include <ATen/core/op_registration/op_registration.h>
```

datumbox (commented):
Fair enough, for `TORCH_LIBRARY_IMPL`. I got one more comment, let me know what you think.

torchvision/csrc/io/image/image.cpp
peterbell10: Ah good point. Changed.

datumbox (approved):
LGTM, thanks!

PR #5119, opened 2021-12-21
Food 101 dataset

jdsgomes:
Addresses #5108.

cc @pmeier @NicolasHug


torchvision/datasets/food101.py
```diff
@@ -0,0 +1,106 @@
...
+            and self._meta_folder.exists()
+            and self._meta_folder.is_dir()
+            and self._images_folder.exists()
+            and self._images_folder.is_dir()
+        )
+
+    def _download(self) -> None:
+        if self._check_exists():
+            return
+        download_and_extract_archive(self._URL, download_root=str(self._base_folder), md5=self._MD5)
```
pmeier, line 106: As is, this doesn't work. `self._base_folder is `food-101`, but the archive extracts itself also into a `food-101` folder. Two possible remedies:

1. `self._meta_folder` should be `self._base_folder / "food-101" / "meta"` and the same for `self._image_folder`
2. The archive should be placed in `self.root`

```suggestion
        download_and_extract_archive(self._URL, download_root=self.root, md5=self._MD5)
```

No strong opinion on my side for either of them.

torchvision/datasets/food101.py
```diff
@@ -0,0 +1,106 @@
...
+
+        if not self._check_exists():
+            raise RuntimeError("Dataset not found. You can use download=True to download it")
+
+        self._labels = []
+        self._image_files = []
+        with open(self._meta_folder / f"{split}.json", "r") as f:
+            metadata = json.loads(f.read())
+
+        self.classes = sorted(set(metadata.keys()))
```
pmeier, line 56: Nit: Dictionary keys are unique by design and can be sorted directly

```suggestion
        self.classes = sorted(metadata.keys())
```

torchvision/datasets/food101.py
```diff
@@ -0,0 +1,106 @@
...
+        self.class_to_idx = dict(zip(self.classes, range(len(self.classes))))
+
+        for class_label, im_paths in metadata.items():
+            self._labels += [self.class_to_idx[class_label]] * len(im_paths)
+            self._image_files += im_paths
```
pmeier, line 61: This is currently not working. The dataset does not store the actual file path, but rather its ID. Thus, we need to do three things to turn this into actual paths:

1. Add the `.jpg` suffix
2. Convert the `/` in the ID into the platform specific path component separator
3. Prefix `self._images_folder`

```suggestion
        for class_label, im_ids in metadata.items():
            self._labels += [self.class_to_idx[class_label]] * len(im_ids)
            self._image_files += [self._images_folder.joinpath(*f"{im_id}.jpg".split(os.sep)) for im_id in im_ids]
```

The tests currently pass, because you put the actual path in the meta data rather than the IDs.

pmeier (changes requested):
Thanks a lot @jdsgomes for the PR! I've left a few comments inline.

torchvision/datasets/food101.py
jdsgomes: After looking closely while checking why the tests are failing I don't think we need to add the extension as it is already in the im_id (relative image path). Please look at my new commit and let me know if it makes sense for you

test/test_datasets.py
```diff
@@ -2168,5 +2168,39 @@ def inject_fake_data(self, tmpdir, config):
...
+        n_samples_per_class = 3 if config["split"] == "train" else 2
+        sampled_classes = ("apple_pie", "crab_cakes", "gyoza")
+        for cls in sampled_classes:
+            im_fnames = datasets_utils.create_image_folder(
+                image_folder,
+                cls,
+                file_name_fn=lambda idx: f"{idx}.jpg",
+                num_examples=num_images_per_class,
+            )
+            metadata[cls] = [str(fname) for fname in random.choices(im_fnames, k=n_samples_per_class)]
```
pmeier: This needs the inverse treatment

```suggestion
            metadata[cls] = [
                "/".join(fname.relative_to(image_folder).with_suffix("").parts)
                for fname in random.choices(im_fnames, k=n_samples_per_class)
            ]
```

pmeier (approved):
A suggestion for the failing tests and some small nits left. Thanks a lot @jdsgomes!

torchvision/datasets/food101.py
```diff
@@ -0,0 +1,91 @@
...
+            metadata = json.loads(f.read())
+
+        self.classes = sorted(metadata.keys())
+        self.class_to_idx = dict(zip(self.classes, range(len(self.classes))))
+
+        for class_label, im_rel_paths in metadata.items():
+            self._labels += [self.class_to_idx[class_label]] * len(im_rel_paths)
+            self._image_files += [
+                self._images_folder.joinpath(*f"{im_rel_path}".split("/")) for im_rel_path in im_rel_paths
+            ]
```
pmeier, line 65: The string that you call `im_rel_path` does **not** include the `.jpg` suffix. The changes you send in 23f685a3e53d5e8f10572cc4e8b82543317216e5 align the fake data in the test with the implementation of the dataset, but this will not work on the actual data. To test this you can simply do

```python

datasets.Food101(root, download=True)[0]
```

I suggest you revert the commit and fix the test as suggested in https://github.com/pytorch/vision/pull/5119#discussion_r773884639.

pmeier (changes requested):
After 23f685a3e53d5e8f10572cc4e8b82543317216e5 the functionality is broken again. My former review was based on the commits before that.

torchvision/datasets/food101.py
pmeier: > as it is already in the im_id

That is not true:

```python
import json

with open(".../food-101/meta/train.json") as file:
    meta = json.load(file)

print(meta[next(iter(meta))][0])
```

```
churros/1004234
```

torchvision/datasets/food101.py
jdsgomes: Thanks for the comment, I am not sure how i thought it did include the extension :/

torchvision/datasets/food101.py
jdsgomes: sorry, I was convinced that it did, but you are 100% right

torchvision/datasets/food101.py
pmeier: No worries. Believe me when I say: this happened to me so many times while writing or reviewing datasets, that I now know what to look out for :grin:

pmeier (approved):
LGTM (again :stuck_out_tongue:) if CI is green. Thanks @jdsgomes!

torchvision/datasets/food101.py
```diff
@@ -0,0 +1,90 @@
...
+        self,
+        root: str,
+        split: str = "train",
+        download: bool = True,
+        transform: Optional[Callable] = None,
+        target_transform: Optional[Callable] = None,
+    ) -> None:
+        super().__init__(root, transform=transform, target_transform=target_transform)
+        self._split = verify_str_arg(split, "split", ("train", "test"))
+        self._base_folder = Path(self._root_path) / "food-101"
```
pmeier, line 42:
```suggestion
        self._base_folder = Path(self.root) / "food-101"
```

PR #5120, opened 2021-12-21
FER2013 dataset

pmeier:
Addresses #5108.

cc @pmeier @bjuncek


torchvision/datasets/fer2013.py
```diff
@@ -0,0 +1,78 @@
...
+        if self.transform is not None:
+            image = self.transform(image)
+
+        if self.target_transform is not None:
+            target = self.target_transform(target)
+
+        return image, target
+
+    def _verify_integrity(self):
+        base_folder = os.path.join(self.root, type(self).__name__.lower())
```
NicolasHug, line 66: This makes it more obvious to relate to the docs.

```suggestion
        base_folder = os.path.join(self.root, "fer2013")
```

torchvision/datasets/fer2013.py
```diff
@@ -0,0 +1,78 @@
...
+from .vision import VisionDataset
+
+
+class FER2013(VisionDataset):
+    """`FER2013
+    <https://www.kaggle.com/c/challenges-in-representation-learning-facial-expression-recognition-challenge>`_ Dataset.
+
+    Args:
+        root (string): Root directory of dataset where directory
+            ``caltech101`` exists or will be saved to if download is set to True.
```
NicolasHug, line 19:
```suggestion
            ``root/fer2013`` exists.
```

torchvision/datasets/fer2013.py
```diff
@@ -0,0 +1,78 @@
...
+        self,
+        root: str,
+        split: str = "train",
+        transform: Optional[Callable] = None,
+        target_transform: Optional[Callable] = None,
+    ) -> None:
+        self._split = verify_str_arg(split, "split", self._RESOURCES.keys())
+        super().__init__(root, transform=transform, target_transform=target_transform)
+
+        with open(self._verify_integrity(), "r", newline="") as file:
```
NicolasHug, line 41: It feels a bit convoluted that it's the job of `_verify_integrity()` to return the current file to read.

I would instead suggest to define the file name in this function (`__init__`) and to pass that file name to `_verify_integrity()`, and only then to read the file in this CM.

NicolasHug (approved):
Thanks a lot @pmeier . I made some minor comments, but LGTM anyway.

PR #5125, opened 2021-12-25
Update code to Python 3.7 compliance

oke-aditya:
Simply runs [pyupgrade](https://github.com/asottile/pyupgrade) with the option --py37-plus

Since Python 3.6 has officially reached end of life from this year. It might be nice to keep our code a bit up to date :) 

P.S. My health has recovered. And extremely sorry for all the delays this month.


oke-aditya:
This is the solution to run recursively on all python files. I am sure I will forget this command and won't find it easily again. Hence pasting it here.

```
pyupgrade `find . -name "*.py" -type f` --py37-plus
```

pmeier (approved):
Thanks @oke-aditya, LGTM! Although I'm wondering why this only pops up for Python >= 3.7. All of the changes also work for Python 3.6 which is evident by the green CI.

oke-aditya:
Then I guess our code was already compliant to python 3.7 and we aren't using any special Features introduced in python 3.7.

pmeier:
Although it looks like we could automate this, https://github.com/pytorch/vision/pull/4585#pullrequestreview-775918877 implies that we can't have it as pre-commit hook. This mean, we probably should run `pyupgrade` regularly to fix stuff like this.

oke-aditya:
Python 3.7 end of life is in 2023 so probably we don't need to run this soon to upgrade python to 3.8. 

But we would need to run it remove code which carries old python syntax. 
I guess we can run some cronjob once in 2 months ? Which just reports it such old code exist? Or just leave it as it doesn't affect much.

PR #3330, opened 2021-01-30
Added utility to draw segmentation masks

oke-aditya:
Closes #3272.

Initial Implementation

- [x] Adds Code
- [x] Adds Tests
- [x] Adds Docs

PR #3296, opened 2021-01-26
Adds fill paramter to utils.draw_bounding_boxes

oke-aditya:
Closes #3280 .

Currently, I have kept fill to `Tuple[int, int, int, int]` or `str`.
Let me know if changes are needed.

Also, let me know how we can test this.

PR #2785, opened 2020-10-10
Add utility to draw bounding boxes

oke-aditya:
Closes #2556  Supersedes #2631 

As per the new API discussion, I will make this compatible with output of detection models.
This will be compatible with only VOC format boxes as this is our default for Input and Output in torchvision.
Users can convert the boxes using the new `box_convert` function and pass. (We can fix this internally too but let's leave it for now)

Will try to get this in before October release 😃 

- [x] Adds code
- [x] Adds docs
- [x] Adds tests

PR #5127, opened 2021-12-25
Random colors for drawing boxes

### LINKED ISSUE #4528: Generate colors from random color palette in draw_bounding_boxes

oke-aditya:
Closes #4528 Supersedes #4658

~~Work in progress will detail it soon.~~
Done! awaiting review.

:) Feels so good to be healthy and code again.

cc @pmeier (I guess Nicolas will return next week)


torchvision/utils.py
```diff
@@ -191,34 +193,45 @@ def draw_bounding_boxes(
     elif image.size(0) not in {1, 3}:
         raise ValueError("Only grayscale and RGB images are supported")
 
+    if labels:
```
oke-aditya, line 196: We need to check these. As agreed. If labels are specified, all the boxes should have labels.

torchvision/utils.py
```diff
@@ -191,34 +193,45 @@ def draw_bounding_boxes(
...
+        if len(labels) != boxes.size(0):
+            raise ValueError("Number of boxes and labels mismatch. Please specify labels for each box.")
+
+    # List of colors should have minimum colors.
+    if colors:
+        if isinstance(colors, list):
+            if len(colors) < boxes.size(0):
+                raise ValueError("Number of colors should be greater or equal to the number of boxes.")
```
oke-aditya, line 204: This is used to silently throw list index out of range error, if number of colors were less than boxes.
I guess it's better to throw an error.

oke-aditya (commented):
Hi! This does the job! I verified the outputs over test cases. Will post the same here.

torchvision/utils.py
```diff
@@ -387,6 +400,30 @@ def _generate_color_palette(num_masks: int):
...
 
 
+def _generate_random_color() -> Tuple[int, int, int]:
+    """
+    Generates a random RGB Color.
+
+    Returns:
+        color Tuple(int, int, int): A Tuple containing Random RGB Color.
+    """
+    return tuple(torch.randperm(256)[:3].tolist())
```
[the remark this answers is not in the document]
oke-aditya: Also do you need tests for generating random color and color palette?
I can add those as well?

Is it worth considering making these two utilities (`generate_random_color` and `generate_random_color_palette`) as part of public API?

oke-aditya:
Script to test the same. Choice of boxes can be improved. :) 

```
dog1_int = read_image('dog1.jpg')
dog2_int = read_image('dog2.jpg')

boxes = torch.tensor(
    [
        [50, 50, 100, 200], 
        [210, 150, 350, 430], 
        [150, 30, 200, 150],
        [20, 30, 70, 80],
        [80, 90, 100, 100],
        [1, 5, 3, 8],
    ]
    , dtype=torch.float
)

print(boxes.size(0))

# # No labels and no colors
result = draw_bounding_boxes(dog1_int, boxes, width=3, font_size=10, fill=True)
show(result)

# # No labels and colors. len colors == no of boxes.
colors = ["blue", "yellow", "red", "orange", "pink", "brown", "pink", "black"]
result = draw_bounding_boxes(dog1_int, boxes, colors=colors, width=3, font_size=10, fill=True)
show(result)

# Labels and no colors
labels = ["l1", "l3", "l3", "l2", "l1", "l3"]
# labels_wrong = ["l3", "l1"]
result = draw_bounding_boxes(dog1_int, boxes, labels=labels, width=3, font_size=10, fill=True)
show(result)


# Labels and colors
labels = ["l1", "l3", "l3", "l2", "l1", "l3"]
colors = ["blue", "yellow", "red", "orange", "pink", "brown"]
result = draw_bounding_boxes(dog1_int, boxes, colors=colors, labels=labels, width=3, font_size=10, fill=True)
show(result)
```

Case 1: - 
No labels and no colors.
Random colors are generated.

Figure_1


Case 2: -
No labels but colors.
Just allocate the pre-defined colors.

Figure_2


Case 3: -
Labels but no colors.
Boxes with same labels have same colors.

Figure_3

Case 4: -
Labels and colors
Just put the defined labels and colors.

Figure_4

torchvision/utils.py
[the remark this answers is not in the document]
oke-aditya: A nice improvement can be to extend this utility and allow to generate random RGBA color as well? 
But I don't see any great use case.

pmeier (commented):
Thanks @oke-aditya for the PR. Given that this was originally discussed with @NicolasHug, we should wait for his review. 

In general though, I'm not sure why we want to have a random color palette by default. I fully agree it is better than using only white for everything, but I'm not sure why randomizing the colors is better than fixed ones. Apart from the non-determinism, by generating the colors at random we also risk using very similar colors for different classes, which can be confusing. It is actually not easy to have multiple colors that are easily distinguishable by humans. I suggest we hardcode color maps such as [this one](https://colorbrewer2.org/#type=qualitative&scheme=Paired&n=12). It "only" defines colors for 12 classes. If we need more, we can always supplement with randomly generated afterwards.

oke-aditya:
The color palette issue is always the concern. I don't exactly recall the conclusion from #2785 #2556 #3296.

I agree with your points. A good default palette > random palette > white.

It is possible to have one (or maybe more than one) good default color palettes, say supporting upto 12 classes. 
After these we could add randomly generated colors to the palette? This can be handled easily in the code.

The problem of colors actually arises when there are many objects of different classes in the same image. I am not sure if there are cases where we have > 12 classes in same image. So I guess it should be okay if two colors are little bit close to each other. The chance of colors being close and having both the classes same is probably a bit low.

Colors is kinda hard and I don't have any great solution.

torchvision/utils.py
```diff
@@ -169,6 +169,8 @@ def draw_bounding_boxes(
         colors (color or list of colors, optional): List containing the colors
             of the boxes or single color for all boxes. The color can be represented as
             PIL strings e.g. "red" or "#FF00FF", or as RGB tuples e.g. ``(240, 10, 157)``.
+            By default, random colors are generated for boxes.
+            If labels are provided, boxes with same labels have same color.
```
datumbox, line 173: I'm concerned with this API change because the behaviour changes depending on whether the labels are defined. Effectively the new API tries to do things for the user instead of letting them decide what to do. I think it would be best to stick to generating a random palette for as many masks as we detected. This will be similar to what we do on the segmentation drawing tool and will enable us to reuse methods such as `_generate_color_palette()`.

NicolasHug: @datumbox would you mind detailing what you mean by "enable us to reuse methods such as _generate_color_palette()" ?

Having one box color per label is a default behaviour that makes sense to me, and @oke-aditya followed my earlier suggestion of https://github.com/pytorch/vision/issues/4528#issuecomment-933232255. It avoids users some extra code.

`draw_segmentation_masks()` doesn't accept a `label` parameter, so it's not unexpected that its behaviour differs from `draw_bounding_boxes()` in that respect.

datumbox: @NicolasHug I was thinking that `utils._generate_color_palette(num_masks)` can be used to generate a colour per mask (not per label) as currently happens on the segmentation case. 

Concerning the lack of support for labels on segmentation, this is indeed the situation with our utility but doesn't have to be the case going forwards. Other libraries such as [detectron2](https://github.com/facebookresearch/detectron2) offer visualizations for segmentation and masks that display the labels.

TBH I haven't followed the original discussions (I try to catch up coming back from PTO). My thoughts on this is that if we want to keep the utility simple, we could let the users define one colour per mask. If they need to apply a more complex logic (such as keeping the same colour per label), this should be done outside of the utility using their own custom logic and provide the list of colours to our util. This would align the approach across segmentation and detection. Let me know your thoughts.

NicolasHug: > Concerning the lack of support for labels on segmentation, this is indeed the situation with our utility but doesn't have to be the case going forwards. Other libraries such as detectron2 offer visualizations for segmentation and masks that display the labels.

I agree. And if we were to support labels in `draw_segmentation_masks()`, I think it would also make sense to have one color per label by default :)

I also agree that complexity is a concern, both regarding the code and regarding what the user would expect from that function. As far as I can tell the only extra code that handling labels adds are these few lines below:


```py
        if labels is not None:
            label_color_map = dict(zip(labels, colors))
            colors = [label_color_map[label] for label in labels]
```

which I feel is minimal.

oke-aditya: Users still can draw colors for their labels. If both labels and colors are passed. The `ith` box would get the `ith` color, irrespective of the labels.
I guess the current improvement just improves how the the default colors are handled. The need to pass a color palette reduces.

torchvision/utils.py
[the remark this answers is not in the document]
oke-aditya: Yes you are right Philip. I didn't know we could do this.

test/test_utils.py
```diff
@@ -317,5 +325,18 @@ def test_draw_keypoints_errors():
         utils.draw_keypoints(image=img, keypoints=invalid_keypoints)
 
 
+def test_random_colors():
```
oke-aditya, line 328: Some tests so that we are OK with the implementation. I guess these aren't very necessary but anyways good to have and don't hurt much.

test/test_utils.py
```diff
@@ -317,5 +325,18 @@ def test_draw_keypoints_errors():
...
+    color_t = utils._generate_random_color()
+    assert isinstance(color_t, tuple)
+    assert len(color_t) == 3
+    assert 256 not in color_t
```
oke-aditya, line 332: Checks the overflow so that we don't go beyond 256.

oke-aditya:
How should we proceed @datumbox @NicolasHug @pmeier ?

I'm bit more inclined with Philip's thoughts. While this would make visualizing boxes with same labels a bit easier. A flip side is the color palette is still messy.

> In general though, I'm not sure why we want to have a random color palette by default. I fully agree it is better than using only white for everything, but I'm not sure why randomizing the colors is better than fixed ones. Apart from the non-determinism, by generating the colors at random we also risk using very similar colors for different classes, which can be confusing. It is actually not easy to have multiple colors that are easily distinguishable by humans. I suggest we hardcode color maps such as this one. It "only" defines colors for 12 classes. If we need more, we can always supplement with randomly generated afterwards.

Maybe we should have a good default color palette for say around 12 or 15 colors and then opt for random color generation?

datumbox:
@oke-aditya As I mentioned earlier, I recommend not handling the per-label colouring within the method but instead accepting a list of colours for each box. Having this extra assumption in the code is not worth it as there are many ways to handle colouring for "stuff vs things". For reference I provide how Detectron2 does it, which shows that a) there are many ways to achieve this and b) you need extra information to handle it effectively:

- stuff handling (per label): https://github.com/facebookresearch/detectron2/blob/main/detectron2/utils/visualizer.py#L456
- things handling (per instance derived as random jittering of label colours): https://github.com/facebookresearch/detectron2/blob/main/detectron2/utils/visualizer.py#L528

oke-aditya:
I understand your point @datumbox . We make an implicit assumption of colors and labels. Let's not think about `per label 
coloring`.   Although I still feel that assuming same color for every label is something expected. Let's keep a ticket open for this and seek user feedback. (Let's not focus on this for now and move on)

This PR also tackles another issue. The default white color palette for all boxes (if no color / color palette is passed) looks bad.
Do you think it's worth replacing the default white with set of random colors?
If yes, then if labels are passed and colors aren't what should be the solution? We still fill random colors?

I believe if we don't keep the label coloring consistency, we should still consider not using plain white as default for every box.

datumbox:
@oke-aditya Though strictly speaking BC breaking, I don't think it would be a big problem to replace the whiles with a reasonable randomly generated palette and use the same colour both for boxes and for labels. 

@NicolasHug any concerns with that?

NicolasHug:
I agree we don't need to be too strict with BC regarding the coloring strategy.

I'm not sure I fully understand the things vs stuff discussion. It seems to me that assigning one color per label could cover a lot of useful use-cases (more so than assigning random colors), but I don't want to further delay this discussion.

pmeier:
> reasonable randomly generated palette

That is actually quite hard. You would need to make sure that the generated colors have a reasonable "distance" between them. Thus, I would suggest to hardcode a color palette and fallback to best effort random generation, i.e. only enforcing the uniqueness of each color, if the input exceeds the number of hardcoded colors.

datumbox:
Indeed, it's a hard problem but we already something like that on `_generate_color_palette`. It produces a reasonable different colours for large number of input colours. We could use this.

oke-aditya:
The problem is kind of NP.

1. Color spaces and colors are hard (considering all 3 should be happy. human eyes, computers and RGB)
2. Torchvision is not a visualization library.
Although libraries like detectron2, etc seem to focus significantly more than torchvision on visualizations.

Also provide a few color palette options like `utils.get_color_palette("bright")` `utils.get_color_palette("dark")`. 
These would give a good default color palette of say 10 (fixed) colors? Which users can plug into colors.  And of course we have `.get_random_color_palette()` to get random palette (of any number of colors).

We can adopt a few palettes from [seaborn](https://github.com/mwaskom/seaborn/blob/master/seaborn/palettes.py). Which seem to be popular.

I'm not sure of re-using `_generate_color_palette`. It returns `Tuple[Tensor(), Tensor(), Tensor()]` and that's neessary for masks. So I won't touch that.

I agree to Phillip and this is something feasible.
By default our utils first try to use colors from one of these palettes (a fixed one) and in case there are boxes > no of colors in palette. It will continue to use random color (no choice). 

(Actually here is where the assumption of one color per label could make the stuff easier. It's less likely that there are variety of objects and we would need lot of colors. More often an object will be either repeated or have num boxes < 10. But let's leave it for later discussion.)

cc @datumbox @NicolasHug @pmeier

datumbox:
> I'm not sure of re-using _generate_color_palette. It returns Tuple[Tensor(), Tensor(), Tensor()] and that's neessary for masks. So I won't touch that.

@oke-aditya I might be missing something but aren't you the one who added the first version of `_generate_color_palette` on this commit https://github.com/pytorch/vision/pull/3330/commits/9bb4191c5bcacb405200b775e4ff247a9bd51ad5? The snippet was refactored on a separate method but code remains mostly the same. You just return as many colours as requested, which is pretty much what we want here. Aka if the user hasn't provided labels, we can generate as many as we need for the provided number of boxes. I recall that we tested the method and it provides a reasonable palette for a sensible small number of boxes back when it was first introduced.

> Torchvision is not a visualization library. Although libraries like detectron2, etc seem to focus significantly more than torchvision on visualizations.

Indeed. We need a small non-complicated utility for people to use to quickly check the model results. There are far better libraries out there for complex visualizations, with detectron2 being one of them. Instead of introducing another palette making method, I would suggest to reuse what we have. Feel free to refactor if necessary. Thankfully it's a private method so there shouldn't be any BC concerns.

> one color per label could make the stuff easier.

Please have a look on the references I provided on https://github.com/pytorch/vision/pull/5127#issuecomment-1008708167 to see how the colours of "things" and "stuff" are handled by libraries like detectron2. As you will see, they require a lot more meta-data about the labels in order to determine if they should keep the same colour for the same label (stuff) or they should produce a slightly similar colour for each instance (things). Adding this logic in our method, without having access to these meta-data is limiting. Hence my preference not to try to do complex things and keep the behaviour simple.

NicolasHug:
> Please have a look on the references I provided on #5127 (comment) to see how the colours of "things" and "stuff" are handled by libraries like detectron2. As you will see, they require a lot more meta-data about the labels in order to determine if they should keep the same colour for the same label (stuff) or they should produce a slightly similar colour for each instance (things). Adding this logic in our method, without having access to these meta-data is limiting. Hence my preference not to try to do complex things and keep the behaviour simple.


Thanks for providing more details on this @datumbox . It seems like detectron2 is using a much more advanced strategy than what we were planning here. Personally, I don't feel like this should prevent us from implementing our initial "one color per label" strategy though. It sounds a bit like we're not moving forward with V1 because we can't yet address V2. I think I understand the idea and in a lot of cases I would agree, but in this specific instance, we all agreed that V1 was really easy to implement (only a few lines https://github.com/pytorch/vision/pull/5127#discussion_r778099309), already quite useful, and we won't have to worry about BC the day we implement V2 (should that day ever come). So that seems like a low risk feature, with solid benefits.

Anyway, I'm OK to move on with this PR and not to implement the proposed "one color per label" strat, but hopefully this is something we can revisit in the future. Perhaps if we open a new PR exclusively with these changes, it will be easier to assess the exact complexity.

oke-aditya:
Hi Sorry for the delay. I will try to minimize the changes in this PR so that we can move forward.

Although I have slightly wider proposal for utils, colors and colormaps in general. 
In a new draft, I will try to implement my thoughts.
The new proposal will look to unify the random color generation in all our viz methods, while also keeping good defaults. This is slightly advanced strategy not as good as detectron2, but I believe it might help. I will comment on it once I finish it. Although I don't fully understand the stuff vs label yet. (I will look into it, but still struggling to get the point. Sorry for it)

oke-aditya:
I guess this should do the minimal change cc @datumbox @NicolasHug

torchvision/utils.py
```diff
@@ -395,33 +391,15 @@ def draw_keypoints(
...
-        colors(List[Tuple[int, int, int]): A List of Tuples each containing random RGB color.
-            Each color is Tuple containing RGB values.
-    """
-    return [_generate_random_color() for i in range(num_colors)]
+    clrs = []
+    for i in range(num_masks):
+        if return_tensor:
+            clrs.append((tuple((palette * i) % 255)))
+        else:
+            clrs.append(tuple(((palette * i) % 255).tolist()))
```
datumbox, line 401: Can you refactor the above 4 lines to avoid code duplication across the 2 cases? Try assigning to a temp var and doing `tolist` on it when necessary prior appending.

oke-aditya: Yeah can do :)

oke-aditya:
I guess we can go ahead with this :) Sorry for the delays.

torchvision/utils.py
```diff
@@ -382,9 +391,16 @@ def draw_keypoints(
...
+def _generate_color_palette(num_masks: int, return_tensor: bool = True):
     palette = torch.tensor([2 ** 25 - 1, 2 ** 15 - 1, 2 ** 21 - 1])
-    return [tuple((i * palette) % 255) for i in range(num_masks)]
+    clrs = []
+    for i in range(num_masks):
+        clr = (palette * i) % 255
+        if return_tensor:
+            clrs.append(tuple(clr))
+        else:
+            clrs.append(tuple((clr.tolist())))
```
datumbox, line 402: Nit: this can still be simplified up as discussed above.

datumbox (approved):
LGTM, there is one more clean up we could do but it's not blocking for me. I'll leave @NicolasHug to review once more and merge if he is OK.

torchvision/utils.py
```diff
@@ -191,34 +192,42 @@ def draw_bounding_boxes(
     elif image.size(0) not in {1, 3}:
         raise ValueError("Only grayscale and RGB images are supported")
 
+    if labels:
+        if len(labels) != boxes.size(0):
+            raise ValueError("Number of boxes and labels mismatch. Please specify labels for each box.")
```
NicolasHug, line 197: Nit: we could indicate the number of boxes and labels to give a more actionable error message:

```suggestion
            raise ValueError(f"Number of boxes ({boxes.size(0)}) and labels ({len(labels)}) mismatch. Please specify labels for each box.")
```

torchvision/utils.py
```diff
@@ -191,34 +192,42 @@ def draw_bounding_boxes(
...
+
+    # List of colors should have minimum colors.
+    if colors:
+        if isinstance(colors, list):
+            if len(colors) < boxes.size(0):
+                raise ValueError("Number of colors should be greater or equal to the number of boxes.")
```
NicolasHug, line 203: Nit: same here we could indicate the exact values.

Also, should we raise an error if num_colors != num_boxes in general?

torchvision/utils.py
```diff
@@ -191,34 +192,42 @@ def draw_bounding_boxes(
...
     txt_font = ImageFont.load_default() if font is None else ImageFont.truetype(font=font, size=font_size)
 
+    if colors is None:
+        colors = _generate_color_palette(len(img_boxes))
+
     for i, bbox in enumerate(img_boxes):
-        if colors is None:
-            color = None
-        elif isinstance(colors, list):
+        if isinstance(colors, list):
```
NicolasHug, line 224: we seem to do a lot of embedded `isinstance()` checks for colors.

Could we instead make sure the `colors` have a consistent format (no matter what the input is) before entering  this loop so that we can simplify the code? E.g. always make sure that colors is a list of tensors of size num_images, or something similar. It would make the code easier to follow and easier to maintain

torchvision/utils.py
NicolasHug: I don't see this `return_tensor` being used anywhere: do we really need it?

NicolasHug (commented):
Thanks @oke-aditya , I gave it a quick look. It looks good, I think we can try to simplify the code a bit.

torchvision/utils.py
oke-aditya: yeah it can be :( Did this in a hurry. 
I guess best way is
```
if not return_tensor:
    clr.tolist()
clrs.append(tuple(clr))
return clrs
```

torchvision/utils.py
oke-aditya: Nope, I would disagree. Providing excess colors makes it easy to define a larger color palette. Like say of 12 colors and utility would dynamically choose colors as needed. This is essentially a feature and intended :smiley:

oke-aditya:
I had a brief chat with @datumbox over slack.
He clarified about Stuff vs Things and why allocating one color per class is not most optimal.

1. Consider the case of Instance segmentation. Here we need to specifically distinguish each person in the image. So say we have 4 people, we can't draw same color masks. It would be tougher to visualize.
2. The current API is flexible enough and if one color per class is needed. Users can write this mapping logic and get it done. We would lose this flexibility once we enforce it.
3. We can consider adding this once we have sufficient metadata (maybe in near future with metadata of models and datasets). Then it would be good and interpretable as why the colors are allocated in specific fashion (something which detectron2 does I believe)

I highly agree to the thoughts and felt that it isn't very necessary for one color per class. Although most use-cases might hold it valid, considering **all** the possibilities it isn't good.

cc @NicolasHug @pmeier 

Apart from that let's move ahead with this :smiley:

torchvision/utils.py
```diff
@@ -198,45 +199,50 @@ def draw_bounding_boxes(
     elif image.size(0) not in {1, 3}:
         raise ValueError("Only grayscale and RGB images are supported")
 
+    num_boxes = boxes.shape[0]
+
+    if labels is None:
+        labels = [None] * num_boxes
```
NicolasHug, line 205: I'm going to need @pmeier 's enlightment on the mypy issue here https://app.circleci.com/pipelines/github/pytorch/vision/14562/workflows/da348bb7-e7cb-46f5-9d6e-0ea7248d59f1/jobs/1171625

What would you suggest?

I tried

```py
        labels: Union[List[str], List[None]] = [None] * num_boxes
```

But `mypy` complains with 

```
torchvision/utils.py:205: error: Name "labels" already defined on line 154  [no-redef]
            labels: Union[List[str], List[None]] = [None] * num_boxes
            ^
```


despite the `allow_redefinition = True` option.

NicolasHug: Very naturally:

```py
        labels: Union[List[str], List[None]] = [None] * num_boxes  # type: ignore[no-redef]
```

won't work either because mypy will instead use the first (outdated) definition instead of this new one.

pmeier: > despite the `allow_redefinition = True` option.

This happens because `allow_redefinition` only applies to cases where the value was used at least once. In your case, we overwrite `labels` without usage and thus this option has no effect. 

> What would you suggest?

If `labels = [None] * ...` is a valid value, I would simply change the input annotation to reflect that: `Optional[List[Optional[str]]]`. This also gives the user the option to exclude certain labels (`labels=["label1", None, "label2"]`) and should be compliant with the rest of the code.

NicolasHug: Thanks @pmeier for the info.

I'm not sure I understand what the "nesting rule" is meant to enforce. For example this makes mypy happy and yet it's strictly equivalent to our code:

```py
    labels: Union[List[str], List[None]] = [None] * num_boxes if labels is None else labels
```

I'd prefer not allowing None values in the `labels` list for users, as I'm not sure there's a use-case where users would have labels for some boxes, but not all (worst cast scenario they can always pass empty  strings). In fact, before this PR, passing a `None`  value for a label would fail as `draw.text` doesn't support it.


So I'll just silence mypy here (twice), because this error  doesn't make much sense.

NicolasHug:
Merging, thanks a lot @oke-aditya for the PR!

oke-aditya:
Sometimes I feel a bit odd. Maintainers say thanks a lot. But really my contribution < 30%. 

Thanks to you guys

NicolasHug:
@oke-aditya you suggested this improvement on your own in https://github.com/pytorch/vision/issues/4528 and took the time to discuss alternatives / implementations strategies, and got to a good working version here... while doing all of this on your free time. This is very valuable to us and to the project, and definitely accounts for a lot more than 30% :)

pmeier:
LoC is a very bad metric to judge a contribution. Keep up the good work!

ElHouas:
Hi @datumbox & @oke-aditya,

I am writing to ask about the colouring of bboxes, as it doesn't colour the bboxes per label. Instead, we are colouring per num_boxes instead per num_labels.
Is possible to colour the bboxes per label with the current API or I am missing something?
Thanks for your time

datumbox:
@ElHouas The API was chosen to give to the user maximum freedom to colour things how they want without making assumptions on our side. You can achieve what you need by something along the lines:
```python
unique = set(labels)
distinct_labels = dict(zip(unique, pallette[:len(unique)]))
colors = [distinct_labels[c] for c in labels]
```

ElHouas:
@datumbox  It makes sense from your side, I appreciate the quick response.
I ll try your suggestion :)

PR #5132, opened 2021-12-27
Add SUN397 Dataset

saswatpp:
Adds [SUN397](https://vision.princeton.edu/projects/2010/SUN/) dataset to address: #5108.  

The size of the dataset was 39GB and a download option was added. Also, 10 official training and testing partitions can be loaded using the `partition` argument (valid values : `int` from 1-10 or `None` for the entire data). Is it a helpful ? @pmeier  [Official Partitions](https://vision.princeton.edu/projects/2010/SUN/download/Partitions.zip)

Thank you.


pmeier (changes requested):
Hey @saswatpp and thanks a lot for the PR! I have some comments inline that make the implementation (hopefully) a little cleaner. On top of that, you need to run our auto-formatters on your code to pass CI. Have a look at our [contributing guide](https://github.com/pytorch/vision/blob/main/CONTRIBUTING.md). In short, you can do the following:

```bash
pip install pre-commit
pre-commit install  # activates the auto-formatter for each subsequent commit
pre-commit run --all-files  # only needed once, since you already commited
git commit -am "fix code format"
```

> Also, I couldn't understand what exactly the tests for datasets does just by looking at the code. Do we have some resources for that ?

The test for the datasets are located in `test/test_datasets.py`. From the contributor perspective you need to add a new test case that configures some basics and provides fake data. 

```python
class SUN397TestCase(datasets_utils.ImageDatasetTestCase):
    DATASET_CLASS = datasets.Food101

    ADDITIONAL_CONFIGS = datasets_utils.combinations_grid(
        split=("train", "test"),
        # There is no need to test all individual partitions, since they all behave the same
        partition=(1, 10, "all"),
    )

    def inject_fake_data(tmp_dir, config):
        ...
```

`inject_fake_data` should prepare `tmp_dir` (in your case the same as `root` in the dataset) in a way that the structure of the files is equal to what would be there if I had instantiated the dataset with `download=True`. It needs to return the number of samples in the dataset. If you want to know more, you can read the [documentation of the underlying test case](https://github.com/pytorch/vision/blob/922db3086e654871c35cd80c2c01eabb65d78475/test/datasets_utils.py#L185).

torchvision/datasets/sun397.py
```diff
@@ -0,0 +1,122 @@
...
+            
+        else:
+            if self.partition is not "all":
+                raise RuntimeError("Enter a valid integer partition from 1 to 10 or \"all\" ")
+            else:
+                for path, _, files in os.walk(self.data_dir):
+                    for file in files:
+                        if(file[:3]=="sun"):
+                            self._image_files.append(Path(path)/file)
+                            self._labels.append(Path(path).relative_to(self.data_dir).as_posix()[2:])
```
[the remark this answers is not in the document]
saswatpp: The integer partition from 1-10 have files associated with them but, "all" doesn't have a file, so I manually browse the class directories and prepare a list out of it. 
That's why I think they cannot be merged. is it correct ?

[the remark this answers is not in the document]
pmeier: Assuming we have either an integer `partition` or `"all"`, you could do something like this:

```python
if partition == "all":
    partitions = list(range(1, 11))
else:
    partitions = [partition]

for partition in partitions:
    with open(self.data_dir / f"{self.split.title()}ing_{partition:02d}.txt", "r") as f:
        ...
```

This would also allow us to accept a collection of integers (or strings) for the `partition` parameter.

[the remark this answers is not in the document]
saswatpp: Sorry I didn't mention it clearly before, but, the "partitions" are just a random subset of entire dataset consisting of 50 images per class(they aren't disjoint, all are random). So adding all of them will lead to repeated images.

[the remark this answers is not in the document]
pmeier: Oh, so `partition="all"` does not mean "all partitions", but rather "all images"? In that case, I suggest we change the value. What about `partition=None` to mean "don't give me any partitions, but rather all images"?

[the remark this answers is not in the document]
saswatpp: Yes seems logical. I will make the change.

saswatpp:
@pmeier Can you tell me how to run tests for specific dataset ? thank you

pmeier:
The easiest would be to run

```bash
$ pytest test/test_datasets.py -k sun397 
```

if the test case you wrote contains `sun397` (case-insenstive) in the name.

saswatpp:
@pmeier, i had mentioned FEATURE_TYPES = (PIL.Image.Image, int) in the test class but the unittest still fails. Do you know what is going on ?

torchvision/datasets/sun397.py
```diff
@@ -0,0 +1,108 @@
...
+                        for posix_rel_path in (line.strip()[1:] for line in f)
+                    )
+                )
+
+        else:
+            for path, _, files in os.walk(self.data_dir):
+                for file in files:
+                    if file[:3] == "sun":
+                        self._image_files.append(Path(path) / file)
+                        self._labels.append(Path(path).relative_to(self.data_dir).as_posix()[2:])
```
pmeier, line 81: We can at least oust the label generation

```suggestion
                if self.partition is not None:
            with open(self.data_dir / f"{self.split.title()}ing_{self.partition:02d}.txt", "r") as f:
                self._image_files = [self.data_dir.joinpath(*line.strip()[1:].split("/")) for line in f]
        else:
            self._image_files = list(self.data_dir.rglob("sun_*.jpg"))
        self._labels = [
            self.class_to_idx["/".join(path.relative_to(self.data_dir).parts[1:-1])] for path in self._image_files
        ]
```

test/test_datasets.py
```diff
@@ -2205,5 +2205,45 @@ def inject_fake_data(self, tmpdir: str, config):
         return len(sampled_classes * n_samples_per_class)
 
 
+class SUN397TestCase(datasets_utils.ImageDatasetTestCase):
+    DATASET_CLASS = datasets.SUN397
+    FEATURE_TYPES = (PIL.Image.Image, int)
```
pmeier, line 2210: This is the default for `ImageDatasetTestCase`'s.

```suggestion
```

pmeier (commented):
Hey @saswatpp, I've suggested one change to the label generation. With that `mypy` is happy and I can't reproduce the test error locally. Let's see if CI is happy to.

torchvision/datasets/sun397.py
```diff
@@ -0,0 +1,108 @@
...
+    """`The SUN397 Data Set <https://vision.princeton.edu/projects/2010/SUN/>`_.
+    The SUN397 is a dataset for scene recognition consisting of 397 categories with 108'754 images.
+    The dataset also provides 10 paritions for training and testing, with each partition
+    consisting of 50 images per class.
+
+    Args:
+        root (string): Root directory of the dataset.
+        split (string, optional): The dataset split, supports ``"train"`` (default) and ``"test"``.
+        parition (string, integer, optional): A valid partition can be an integer from 1 to 10 or ``"all"``
+            for the entire dataset.
```
pmeier, line 21: This needs to be adapted for `None`

torchvision/datasets/sun397.py
```diff
@@ -0,0 +1,108 @@
...
+        transform (callable, optional): A function/transform that  takes in an PIL image and returns a transformed
+            version. E.g, ``transforms.RandomCrop``.
+        target_transform (callable, optional): A function/transform that takes in the target and transforms it.
+    """
+
+    _URL = "http://vision.princeton.edu/projects/2010/SUN/SUN397.tar.gz"
+    _FILENAME = "SUN397.tar.gz"
+    _MD5 = "8ca2778205c41d23104230ba66911c7a"
+    _PARTITIONS_URL = "https://vision.princeton.edu/projects/2010/SUN/download/Partitions.zip"
+    _PARTITIONS_FILENAME = "Partitions.zip"
```
pmeier, line 31: 1. I think we don't need them anymore, right?
2. Shouldn't we also have a checksum for the partitions file?

```suggestion
    _URL = "http://vision.princeton.edu/projects/2010/SUN/SUN397.tar.gz"
    _MD5 = "8ca2778205c41d23104230ba66911c7a"
    _PARTITIONS_URL = "https://vision.princeton.edu/projects/2010/SUN/download/Partitions.zip"
```

torchvision/datasets/sun397.py
```diff
@@ -0,0 +1,108 @@
...
+        return self.data_dir.exists() and self.data_dir.is_dir()
+
+    def extra_repr(self) -> str:
+        return "Split: {split}".format(**self.__dict__)
+
+    def _download(self) -> None:
+        if self._check_exists:
+            return
+        download_and_extract_archive(self._URL, download_root=self.root, md5=self._MD5)
+        download_and_extract_archive(self._PARTITIONS_URL, download_root=self.data_dir)
```
pmeier, line 108: `download_and_extract_archive` works with `Path`'s, but is only annotated for `str`'s.

```suggestion
        download_and_extract_archive(self._PARTITIONS_URL, download_root=str(self.data_dir))
```

test/test_datasets.py
```diff
@@ -2205,5 +2205,45 @@ def inject_fake_data(self, tmpdir: str, config):
...
+
+    ADDITIONAL_CONFIGS = datasets_utils.combinations_grid(
+        split=("train", "test"),
+        partition=(1, 10, None),
+    )
+
+    def inject_fake_data(self, tmpdir: str, config):
+        data_dir = pathlib.Path(tmpdir) / "SUN397"
+
+        data_dir.mkdir(parents=True)
```
pmeier, line 2220: `tmpdir` already exists, so there is no need to also create the `parents`.

```suggestion
        data_dir.mkdir()
```

test/test_datasets.py
```diff
@@ -2205,5 +2205,45 @@ def inject_fake_data(self, tmpdir: str, config):
...
+                datasets_utils.create_image_folder(
+                    image_folder,
+                    image_folder / cls,
+                    file_name_fn=lambda idx: f"sun_{idx}.jpg",
+                    num_examples=num_images_per_class,
+                )
+            )
+
+            with open(data_dir / "ClassName.txt", "a") as file:
+                file.write("/" + cls[0] + "/" + cls + "\n")
```
pmeier, line 2237: Nit: there is no need to write to the file multiple times by appending to it:

```suggestion
        with open(data_dir / "ClassName.txt", "w") as file:
            file.writelines("\n".join(f"/{cls[0]}/{cls}" for cls in sampled_classes))
```

test/test_datasets.py
```diff
@@ -2205,5 +2205,45 @@ def inject_fake_data(self, tmpdir: str, config):
...
+        data_dir = pathlib.Path(tmpdir) / "SUN397"
+
+        data_dir.mkdir(parents=True)
+
+        num_images_per_class = 5
+        sampled_classes = ("abbey", "airplane_cabin", "airport_terminal")
+        im_paths = []
+        for cls in sampled_classes:
+            image_folder = data_dir / cls[0]
+            im_paths.append(
```
pmeier, line 2227: By `extend`'ing here, you can avoid the list of lists structure.

```suggestion
            im_paths.extend(
```

test/test_datasets.py
```diff
@@ -2205,5 +2205,45 @@ def inject_fake_data(self, tmpdir: str, config):
...
+            with open(data_dir / "ClassName.txt", "a") as file:
+                file.write("/" + cls[0] + "/" + cls + "\n")
+
+        if config["partition"] is not None:
+            with open(data_dir / f"{config['split'].title()}ing_{config['partition']:02d}.txt", "w") as file:
+                for f_pathlist in im_paths:
+                    for f_path in f_pathlist:
+                        file.write("/" + str(pathlib.Path(f_path).relative_to(data_dir).as_posix()) + "\n")
+
+        return len(sampled_classes * num_images_per_class)
```
pmeier, line 2245: I think it would be a good idea to introduce some differences between the partitions here. The suggestion adds a variable number of elements based on the split and also selects random images instead.

```suggestion
                if config["partition"] is not None:
            num_samples = max(len(im_paths) // (2 if config["split"] == "train" else 3), 1)
            with open(data_dir / f"{config['split'].title()}ing_{config['partition']:02d}.txt", "w") as file:
                file.writelines(
                    "\n".join(
                        f"/{f_path.relative_to(data_dir).as_posix()}"
                        for f_path in random.choices(im_paths, k=num_samples)
                    )
                )
        else:
            num_samples = len(im_paths)

        return num_samples
```

torchvision/datasets/sun397.py
```diff
@@ -0,0 +1,108 @@
...
+            if self.partition < 0 or self.partition > 10:
+                raise RuntimeError("Enter a valid integer partition from 1 to 10 or None, for entire dataset")
+
+        if download:
+            self._download()
+
+        if not self._check_exists():
+            raise RuntimeError("Dataset not found. You can use download=True to download it")
+
+        with open(self.data_dir / "ClassName.txt", "r") as f:
```
pmeier, line 57: Nit: opening in reading mode is the default.

```suggestion
        with open(self.data_dir / "ClassName.txt") as f:
```

torchvision/datasets/sun397.py
```diff
@@ -0,0 +1,95 @@
...
+    """`The SUN397 Data Set <https://vision.princeton.edu/projects/2010/SUN/>`_.
+
+    The SUN397 or Scene UNderstanding (SUN) is a dataset for scene recognition consisting of
+    397 categories with 108'754 images. The dataset also provides 10 partitions for training
+    and testing, with each partition consisting of 50 images per class.
+
+    Args:
+        root (string): Root directory of the dataset.
+        split (string, optional): The dataset split, supports ``"train"`` (default) and ``"test"``.
+        parition (integer, optional): A valid partition can be an integer from 1 to 10 or None,
```
pmeier, line 20:
```suggestion
        partition (integer, optional): A valid partition can be an integer from 1 (default) to 10 or None,
```

pmeier (approved):
Thanks a lot @saswatpp for the PR. LGTM!

torchvision/datasets/sun397.py
```diff
@@ -0,0 +1,95 @@
...
+
+    The SUN397 or Scene UNderstanding (SUN) is a dataset for scene recognition consisting of
+    397 categories with 108'754 images. The dataset also provides 10 partitions for training
+    and testing, with each partition consisting of 50 images per class.
+
+    Args:
+        root (string): Root directory of the dataset.
+        split (string, optional): The dataset split, supports ``"train"`` (default) and ``"test"``.
+        partition (integer, optional): A valid partition can be an integer from 1 (default) to 10 or None,
+            for the entire dataset.
```
NicolasHug, line 21: It looks like the download parameter is missing an entry in the docstring

torchvision/datasets/sun397.py
```diff
@@ -0,0 +1,95 @@
...
+        return image, label
+
+    def _check_exists(self) -> bool:
+        return self.data_dir.exists() and self.data_dir.is_dir()
+
+    def extra_repr(self) -> str:
+        return "Split: {split}".format(**self.__dict__)
+
+    def _download(self) -> None:
+        if self._check_exists:
```
NicolasHug, line 92:
```suggestion
        if self._check_exists():
```

torchvision/datasets/sun397.py
```diff
@@ -0,0 +1,95 @@
...
+            return
+        download_and_extract_archive(self._DATASET_URL, download_root=self.root, md5=self._MD5)
```
NicolasHug, line 94:
```suggestion
        download_and_extract_archive(self._DATASET_URL, download_root=self.root, md5=self._DATASET_MD5)
```

torchvision/datasets/sun397.py
```diff
@@ -0,0 +1,95 @@
...
+        split: str = "train",
+        partition: Optional[int] = 1,
+        download: bool = True,
+        transform: Optional[Callable] = None,
+        target_transform: Optional[Callable] = None,
+    ) -> None:
+        super().__init__(root, transform=transform, target_transform=target_transform)
+        self.split = verify_str_arg(split, "split", ("train", "test"))
+        self.partition = partition
+        self.data_dir = Path(self.root) / "SUN397"
```
NicolasHug, line 44: Let's keep this private

```suggestion
        self._data_dir = Path(self.root) / "SUN397"
```

torchvision/datasets/sun397.py
```diff
@@ -0,0 +1,95 @@
...
+
+        if self.partition is not None:
+            if self.partition < 0 or self.partition > 10:
+                raise RuntimeError("Enter a valid integer partition from 1 to 10 or None, for entire dataset")
```
NicolasHug, line 48: Perhaps this would be a bit more actionable:

```suggestion
                raise RuntimeError(f"The partition parameter should be an int in [1, 10] or None, got {partition}.")
```

torchvision/datasets/sun397.py
NicolasHug, line 21: @pmeier it looks like this `partition` parameter has a different meaning from that of DTD, where a given "partition" covers the entire dataset, not a subset of it. If that's the case, we should think of a different name (either for this one or for that of DTD).

NicolasHug (commented):
Thanks a lot for the PR @saswatpp and @pmeier for the review. I made a few comments below, the `download()` function needs some minor fixes but other than that it looks great.  LMK what you think.

torchvision/datasets/sun397.py
pmeier: Good catch!

torchvision/datasets/sun397.py
saswatpp: yes in the case of SUN397, partition is a random subset of whole dataset with 50 images per class. Let me know if changes needs to be made to the naming, I will do this in this PR itself. @pmeier

NicolasHug: We've discussed this a bit with @pmeier , let's keep it as it is here. We might change that of DTD but we'll do that later. Thanks for checking @saswatpp

pmeier: We should stick with `partition` here and change the name in `DTD`.

saswatpp: Ok Sure <3

pmeier:
@saswatpp Could you fix the merge conflicts? Afterwards this should be good to go.

NicolasHug:
> Afterwards this should be good to go

I'm wondering whether @saswatpp pushed their changes yet? I see all comments as resolved but I don't see any new changes in the diff 😅

saswatpp:
oh @NicolasHug I have made the changes in local repo gotta commit them only 😓

NicolasHug:
No worries at all @saswatpp !

saswatpp:
git was behaving weird, so force pushed

torchvision/datasets/sun397.py
```diff
@@ -0,0 +1,98 @@
...
+    """`The SUN397 Data Set <https://vision.princeton.edu/projects/2010/SUN/>`_.
+
+    The SUN397 or Scene UNderstanding (SUN) is a dataset for scene recognition consisting of
+    397 categories with 108'754 images. The dataset also provides 10 partitions for training
+    and testing, with each partition consisting of 50 images per class.
+
+    Args:
+        root (string): Root directory of the dataset.
+        split (string, optional): The dataset split, supports ``"train"`` (default) and ``"test"``.
+        parition (integer, optional): A valid partition can be an integer from 1 to 10 or None,
```
NicolasHug, line 20:
```suggestion
        partition (int, optional): A valid partition can be an integer from 1 to 10 or None,
```

NicolasHug (approved):
Thanks a lot @saswatpp !

Running the tests locally, I'm seeing a lot of warnings about unclosed files:

```
test/test_datasets.py::LSUNTestCase::test_transforms
  /Users/nicolashug/dev/vision/torchvision/datasets/lsun.py:31: ResourceWarning: unclosed file <_io.BufferedWriter name='_cache_varfolderssyvkyzpyhrlqqcgnTtmpclcnsatowervallmdb'>
    pickle.dump(self.keys, open(cache_file, "wb"))

```


This is simlar to what happened in https://github.com/pytorch/vision/pull/5116#pullrequestreview-844492353, perhaps @pmeier can advise on how he fixed it over there?


Other than that, LGTM!

pmeier:
@NicolasHug As the error message implies, this is about files of the LSUN dataset and thus unrelated to this PR. The same warnings are present on the `main` branch. I'll look into it.

PR #5134, opened 2021-12-28
Adds utlity to convert optical flow to an image

### LINKED ISSUE #5091: Add `flow_to_image()` visualization util
Now that RAFT is done, it'd be nice to have a conversion util from a 2xHxW `flow` to a 3xHxW RGB image. I think the implementation of the original RAFT paper follows a standard color map, so we could do something similar (trimmed down to the minimal parts, e.g. no BRG etc.) https://github.com/princeton-vl/RAFT/blob/master/core/utils/flow_viz.py

Note: this is a visualization util meant to land in `torchvision.utils`; it's not meant to be a `transform` nor an `ops`.

oke-aditya:
Closes #5091 

Basically adds the utility that does the job. 
1. It ports the suggested code and replaces the use of numpy with torch.

2. Supports flow of shape (2, H, W) and returns an image of shape (3, H, W).
Slight difference from the original code. To support the outputs of raft and avoid permuting.

3. Does not support conversion to BGR. I am not sure if that is needed to be added.

4. Uses the same colorwheel.
I am not sure of any other alternative. Also I think the colorwheel is very standard.

~~Tests are pending.~~

cc @NicolasHug


torchvision/utils.py
```diff
@@ -382,6 +389,132 @@ def draw_keypoints(
...
+        clip_flow (float, optional): Clip maximum of flow values. Defaults to None.
+
+    Returns:
+        img (Tensor[3, H, W]): Image Tensor with flow drawn.
+    """
+
+    if flow_uv.ndim != 3:
+        raise ValueError("Input flow should have 3 dimensions [2, H, W]")
+    if flow_uv.size(0) != 2:
+        raise ValueError("Input flow should have shape [2, H, W]")
```
oke-aditya, line 413: Plain asserts were replaced with ValueErrors.

torchvision/utils.py
```diff
@@ -382,6 +389,132 @@ def draw_keypoints(
...
+
+    if clip_flow is not None:
```
oke-aditya, line 415: I guess clip flow is useful so it's good to keep it.

torchvision/utils.py
```diff
@@ -382,6 +389,132 @@ def draw_keypoints(
...
+        v (torch.Tensor): Input vertical flow of shape [H, W]
+
+    Returns:
+       img (Tensor[3, H, W]): Flow visualization image.
+    """
+
+    flow_image = torch.zeros((3, u.shape[0], u.shape[1]), dtype=torch.uint8)
+    colorwheel = _make_colorwheel()
+    # shape [55x3]
+    ncols = colorwheel.shape[0]
```
oke-aditya, line 448: This is fixed to 255. Should we hardcode this?

torchvision/utils.py
```diff
@@ -382,6 +389,132 @@ def draw_keypoints(
...
+    rad = torch.sqrt(torch.square(u) + torch.square(v))
+
+    a = torch.atan2(-v, -u) / torch.pi
+    fk = (a + 1) / 2 * (ncols - 1)
+    k0 = torch.floor(fk).to(torch.long)
+    k1 = k0 + 1
+    k1[k1 == ncols] = 0
+    f = fk - k0
+
+    for i in range(colorwheel.shape[1]):
```
oke-aditya, line 458: Again this is fixed to 3. Should we hardcode it?

torchvision/utils.py
```diff
@@ -382,6 +389,132 @@ def draw_keypoints(
...
+        tmp = colorwheel[:, i]
+        col0 = tmp[k0] / 255.0
+        col1 = tmp[k1] / 255.0
+        col = (1 - f) * col0 + f * col1
+        idx = rad <= 1
+        col[idx] = 1 - rad[idx] * (1 - col[idx])
+        col[~idx] = col[~idx] * 0.75  # out of range
+        ch_idx = i
+        flow_image[ch_idx, :, :] = torch.floor(255 * col)
```
oke-aditya, line 467: Note that the `ch_idx` becomes the first dimension as our flow_image is CHW unlike HWC in original implementation.
Also, I avoided `convert_to_bgr` to trim down the implementation.

test/test_utils.py
```diff
@@ -317,5 +317,15 @@ def test_draw_keypoints_errors():
         utils.draw_keypoints(image=img, keypoints=invalid_keypoints)
 
 
+def test_flow_to_image_errors():
```
oke-aditya, line 320: Not sure what and how to test. Will need further guidance here.

torchvision/utils.py
```diff
@@ -382,6 +389,132 @@ def draw_keypoints(
...
+
+    flow_image = torch.zeros((3, u.shape[0], u.shape[1]), dtype=torch.uint8)
+    colorwheel = _make_colorwheel()
+    # shape [55x3]
+    ncols = colorwheel.shape[0]
+    rad = torch.sqrt(torch.square(u) + torch.square(v))
+
+    a = torch.atan2(-v, -u) / torch.pi
+    fk = (a + 1) / 2 * (ncols - 1)
+    k0 = torch.floor(fk).to(torch.long)
```
oke-aditya, line 453: Need to differ here and use torch.long instead of torch.int32. This is due to indexing error of int dtype. (you can scroll through  https://github.com/pytorch/vision/pull/5134/commits/036c35ef9186832a80f771ed4f46875cc8590b04 )

oke-aditya (commented):
Few points to clarify while implementing.
Also I guess we shouldn't make `_make_colorwheel` and `_flow_uv_to_colors` public.

oke-aditya:
Here is the new demo script. I rewrote it a bit so that it doesn't use argparse and can directly be used as example

```
import cv2
import numpy as np
import torch

from torchvision.models.optical_flow import raft_large
import flow_viz_new


def frame_preprocess(frame, device):
    frame = torch.from_numpy(frame).permute(2, 0, 1).float()
    frame = frame.unsqueeze(0)
    frame = frame.to(device)
    return frame


def vizualize_flow(img, flo, save, counter):
    # With the new utility we shouldn't need to permute and change to numpy
    # Since the utility will not work on batches. We need first image and flow
    img = img[0]
    flo = flo[0]

    # map flow to rgb image
    flo = flow_viz_new.flow_to_image(flo)

    # Yeah but we need to reverse the stuff.
    # After We draw the flow for Opencv.
    flo = flo.permute(1, 2, 0).cpu().numpy()
    img = img.permute(1, 2, 0).cpu().numpy()
    flo = cv2.cvtColor(flo, cv2.COLOR_RGB2BGR)

    # concatenate, save and show images
    img_flo = np.concatenate([img, flo], axis=0)

    if save:
        cv2.imwrite(f"demo_frames/frame_{str(counter)}.jpg", img_flo)
    cv2.imshow("Optical Flow", img_flo / 255.0)
 
    k = cv2.waitKey(25) & 0xFF
    if k == 27:
        return False
    return True


if __name__ == "__main__":
    model = raft_large(pretrained=True)
    device = "cpu"
    model = model.eval()
    video_path = "./crowd.mp4"

    cap = cv2.VideoCapture(video_path)
    cap.set(cv2.CAP_PROP_POS_FRAMES, int(3 * 30))
    ret, frame_1 = cap.read()
    frame_1 = frame_preprocess(frame_1, device)

    counter = 0
    with torch.no_grad():
        while True:
            # read the next frame
            ret, frame_2 = cap.read()
            if not ret:
                break

            frame_2 = frame_preprocess(frame_2, device)
            flow_preds = model(2 * frame_1 / 255 - 1, 2 * frame_2 / 255 - 1)
            flow_up = flow_preds[-1]

            ret = vizualize_flow(frame_1, flow_up, save=False, counter=counter)
            if not ret:
                break
            frame_1 = frame_2
            counter += 1

```

oke-aditya:
Few outputs.

Optical Flow_screenshot_0

Optical Flow_screenshot_1

Optical Flow_screenshot_2

torchvision/utils.py
```diff
@@ -382,6 +389,132 @@ def draw_keypoints(
...
+    # BM
+    colorwheel[col : col + BM, 2] = 255
+    colorwheel[col : col + BM, 0] = torch.floor(255 * torch.arange(0, BM) / BM)
+    col = col + BM
+    # MR
+    colorwheel[col : col + MR, 2] = 255 - torch.floor(255 * torch.arange(MR) / MR)
+    colorwheel[col : col + MR, 0] = 255
+    return colorwheel
+
+
```
oke-aditya, line 517: I believe this file is getting a bit big. I don't know what's the convention and limits for number of lines in a file.
Maybe we should consider making `utils` a folder a expose accordingly.

This will allow to define colormaps like these in separate file.

torchvision/utils.py
NicolasHug, line 448: nit: our naming convention would be `num_cols`

torchvision/utils.py
```diff
@@ -382,6 +389,132 @@ def draw_keypoints(
...
+    epslion = 1e-5
+    u = u / (rad_max + epslion)
+    v = v / (rad_max + epslion)
+    return _flow_uv_to_colors(u, v)
+
+
+@torch.no_grad()
+def _flow_uv_to_colors(
+    u: torch.Tensor,
+    v: torch.Tensor,
```
NicolasHug, line 431: instead of acception `u` and `v` separately I think we could just accept a single tensor `normalized_flow`

torchvision/utils.py
```diff
@@ -382,6 +389,132 @@ def draw_keypoints(
...
+
+
+def _make_colorwheel() -> torch.Tensor:
+    """
+    Generates a color wheel for optical flow visualization as presented in:
+    Baker et al. "A Database and Evaluation Methodology for Optical Flow" (ICCV, 2007)
+    URL: http://vision.middlebury.edu/flow/flowEval-iccv07.pdf.
+
+    Returns:
+        colorwheel (Tensor[255, 3]): Colorwheel Tensor.
```
NicolasHug, line 478: are you sure that this is 255? According to the code below this should be 55

torchvision/utils.py
```diff
@@ -382,6 +389,132 @@ def draw_keypoints(
...
+        u (torch.Tensor): Input horizontal flow of shape [H, W]
+        v (torch.Tensor): Input vertical flow of shape [H, W]
+
+    Returns:
+       img (Tensor[3, H, W]): Flow visualization image.
+    """
+
+    flow_image = torch.zeros((3, u.shape[0], u.shape[1]), dtype=torch.uint8)
+    colorwheel = _make_colorwheel()
+    # shape [55x3]
```
NicolasHug, line 447: I would remove this comment, or move it next to the `_make_colorwheel()` call

NicolasHug (commented):
Thanks for working on this @oke-aditya! I made a few comments and suggestions, let me know what you think!

torchvision/utils.py
oke-aditya: It's 55 my bad

torchvision/utils.py
```diff
@@ -382,6 +389,130 @@ def draw_keypoints(
...
+    rad = torch.sqrt(torch.square(u) + torch.square(v))
+    rad_max = torch.max(rad)
+    epslion = 1e-5
+    u = u / (rad_max + epslion)
+    v = v / (rad_max + epslion)
+    return _flow_to_colors(u, v)
+
+
+@torch.no_grad()
+def _flow_to_colors(
```
NicolasHug, line 428: maybe this would be a more descriptive name?

```suggestion
def _normalized_flow_to_image(
```

oke-aditya: I guess you mean the other function. But I get it.

torchvision/utils.py
```diff
@@ -382,6 +389,132 @@ def draw_keypoints(
...
+        raise ValueError("Input flow should have 3 dimensions [2, H, W]")
+    if flow_uv.size(0) != 2:
+        raise ValueError("Input flow should have shape [2, H, W]")
+
+    if clip_flow is not None:
+        flow_uv = torch.clip(flow_uv, 0, clip_flow)
+
+    u = flow_uv[0, :, :]
+    v = flow_uv[1, :, :]
+    rad = torch.sqrt(torch.square(u) + torch.square(v))
```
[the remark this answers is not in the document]
oke-aditya: .finfo will need to flow to by torch.float dtype. Do we assume it and check for it? running .finfo on int tensor will lead to error.
Should I check for this and raise a value error to pass flow of float dtype? or implicitly cast flow to float?

torchvision/utils.py
oke-aditya: I guess this is internal function and it might be safe to call it colors, as it can confuse the notion of colors and image. 
So will rename to `normalized_flow_to_colors`

torchvision/utils.py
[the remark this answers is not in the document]
oke-aditya: I believe there is small error it should be dim=0 as we normalize across channels? I could not get good outputs or match the results with dim=1. But dim=0 worked.

torchvision/utils.py
oke-aditya: I tried playing with clip flow. 0 leads to white image, (nothing remains from flow).
I think it doesn't help much and too can't find any significant visualization difference. I too think let's remove it and let users ask for use case.

torchvision/utils.py
```diff
@@ -382,6 +389,130 @@ def draw_keypoints(
     return torch.from_numpy(np.array(img_to_draw)).permute(2, 0, 1).to(dtype=torch.uint8)
 
 
+# Flow visualization code adapted from https://github.com/tomrunia/OpticalFlow_Visualization
+@torch.no_grad()
+def flow_to_image(
```
oke-aditya, line 394: Maybe we should name this as `draw_optical_flow` ?

oke-aditya:
Would it be nice to have an example of this utlity + RAFT + using the torchvision video reader API (if possible) as gallery example?

torchvision/utils.py
```diff
@@ -382,6 +389,116 @@ def draw_keypoints(
...
+
+    Returns:
+        img (Tensor(3, H, W)): Image Tensor where each color corresponds to a given flow direction.
+    """
+
+    if flow.ndim != 3 or flow.size(0) != 2:
+        raise ValueError(f"Input flow should have shape (2, H, W), got {flow.shape}.")
+
+    max_norm = torch.sum(flow ** 2, dim=0).sqrt().max()
+    epsilon = torch.finfo((flow).dtype).eps
```
NicolasHug, line 412: As you correctly noted earlier this will fail for non-float dtype, so let's add a check above that ensures that `flow` is of dtype `torch.float`.

torchvision/utils.py
```diff
@@ -382,6 +389,116 @@ def draw_keypoints(
...
+    normalized_flow = flow / (max_norm + epsilon)
+    return _normalized_flow_to_colors(normalized_flow)
+
+
+@torch.no_grad()
+def _normalized_flow_to_colors(normalized_flow: torch.Tensor) -> torch.Tensor:
```
NicolasHug, line 418: It's a private function so I won't insist more, but I think it still makes sense to name this `_normalized_flow_to_image` because:

- this accurately describes what this function returns: it returns a tensor image, not "colors"
- we already have`flow_to_image()` which returns the same thing as this function.

torchvision/utils.py
```diff
@@ -382,6 +389,116 @@ def draw_keypoints(
...
+
+    """
+    Applies the flow color wheel to (possibly clipped) flow components u and v.
```
NicolasHug, line 421:
```suggestion
    Converts a normalized flow to an RGB image
```

torchvision/utils.py
```diff
@@ -382,6 +389,116 @@ def draw_keypoints(
...
+
+    Args:
+        normalized_flow (torch.Tensor): Normalized flow tensor of shape (2, H, W)
+    Returns:
+       img (Tensor(3, H, W)): Flow visualization image of dtype uint8.
+    """
+
+    u = normalized_flow[0, :, :]
+    v = normalized_flow[1, :, :]
```
NicolasHug, line 430: Would it be possible to drop the use of `u` and `v` here as well?

torchvision/utils.py
```diff
@@ -382,6 +389,116 @@ def draw_keypoints(
...
+    k1 = k0 + 1
+    k1[k1 == num_cols] = 0
+    f = fk - k0
+
+    for i in range(colorwheel.shape[1]):
+        tmp = colorwheel[:, i]
+        col0 = tmp[k0] / 255.0
+        col1 = tmp[k1] / 255.0
+        col = (1 - f) * col0 + f * col1
+        idx = rad <= 1
```
NicolasHug, line 447: Because this function takes the normalized flow as input, I believe that `rad` (actually the 2-norm) is  guaranteed to always be `<= 1`. So I feel like we might be able to drop this `idx` logic altogether?

torchvision/utils.py
```diff
@@ -382,6 +389,116 @@ def draw_keypoints(
...
+@torch.no_grad()
+def flow_to_image(
+    flow: torch.Tensor,
+) -> torch.Tensor:
+
+    """
+    Converts the two-dimensional flow to an RGB Image.
+
+    Args:
+        flow (Tensor): Flow of shape (2, H, W) and dtype torch.float.
```
NicolasHug, line 402: On top of single flow inputs of shape `(2, H, W)`, I think it would be very useful to also support batches of flows of shape `(N, 2, H, W)`.

But I'm happy to leave that for a follow-up PR

NicolasHug (commented):
Thanks a lot @oke-aditya , I made a few more comments but we're getting close :) !

> Would it be nice to have an example of this utlity + RAFT + using the torchvision video reader API (if possible) as gallery example?
> Can do that in a new PR 😄

Having an example in the gallery would be fantastic (in fact it's the main reason why I wanted to have such `flow_to_image()` utility in the first place). Thanks for proposing, we'll be happy to review your PR :)

torchvision/utils.py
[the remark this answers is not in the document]
NicolasHug: Good catch, I copy/pasted it from another place and `1`  only works when the input is batched. Here it should be 0 indeed.

torchvision/utils.py
oke-aditya: Can do that, but this code seems a bit more easy to read.

NicolasHug: I'm not sure I agree here :)

For example this would make it more obvious that we're computing the norm in the same way as above in the other function, and perhaps it would make https://github.com/pytorch/vision/pull/5134/files#r786121504 more obvious.

Also this is not "idiomatic" in the sense that this is not something that we've been using so far, so sticking with the full tensor would be more consistent in terms of style.

torchvision/utils.py
```diff
@@ -382,6 +389,116 @@ def draw_keypoints(
...
+       img (Tensor(3, H, W)): Flow visualization image of dtype uint8.
+    """
+
+    u = normalized_flow[0, :, :]
+    v = normalized_flow[1, :, :]
+    flow_image = torch.zeros((3, u.shape[0], u.shape[1]), dtype=torch.uint8)
+    colorwheel = _make_colorwheel()  # shape [55x3]
+    num_cols = colorwheel.shape[0]
+    rad = torch.sqrt(torch.square(u) + torch.square(v))
+    a = torch.atan2(-v, -u) / torch.pi
```
NicolasHug, line 435:
```suggestion
    _, H, W = normalized_flow.shape
    flow_image = torch.zeros((3, H, W), dtype=torch.uint8)
    colorwheel = _make_colorwheel()  # shape [55x3]
    num_cols = colorwheel.shape[0]
    norm = torch.sum(flow ** 2, dim=0).sqrt()
    a = torch.atan2(-normalized_flow[1], -normalized_flow[0]) / torch.pi
```

torchvision/utils.py
oke-aditya: Simplified this entire logic.

torchvision/utils.py
oke-aditya: Other utilities don't support batched input. I would like to know why we need to support batches of flow in this case? (Probably since we are operating on videos?)
Will follow-up in another PR.

oke-aditya:
Guess we are left with tests, what should it be? Should I do something like we did for drawing boxes? Compare two images or is there way to check the specified pixels have got assigned the specific color for given image?

torchvision/utils.py
```diff
@@ -382,6 +389,115 @@ def draw_keypoints(
...
+    k0 = torch.floor(fk).to(torch.long)
+    k1 = k0 + 1
+    k1[k1 == num_cols] = 0
+    f = fk - k0
+
+    for i in range(colorwheel.shape[1]):
+        tmp = colorwheel[:, i]
+        col0 = tmp[k0] / 255.0
+        col1 = tmp[k1] / 255.0
+        col = 1 - norm * (1 - col0 + f * (col0 - col1))
```
oke-aditya, line 448: Working out on pen and paper. Removing idx results in this simple expression. I verified the outputs as well they look same.

torchvision/utils.py
NicolasHug: The model outputs batched input. Imagine you have a video for which you want to get all predicted flows as images to re-construct a gif of predicted flow afterwards.

You'll first get batched predicted flow, then if `flow_to_image` does not support batches, you'll have to manually loop over it. This is both inconvenient and also probably slower. It'd be nice to just be able to do

```py
predicted_flows = model(...)[-1]   # N 2 H W
images = flow_to_image(predicted_flows)  # N 3 H W
```

It's the most common use-case and this is what we'll need when we write a gallery example as well

torchvision/utils.py
NicolasHug: `flow_to_image` sounds good to me: it's in line with our current utilities like "pil_to_tensor()", etc.

NicolasHug (commented):
Thanks a lot @oke-aditya , this looks great!

For the test, I think a good test would be to generate an image that creates all possible flow values. Something like this:

```py
import torch
from torchvision.utils import flow_to_image
from torchvision.io import write_jpeg
h, w = 100, 100

flow = torch.meshgrid(torch.arange(h), torch.arange(w), indexing="ij")
flow = torch.stack(flow[::-1], dim=0).float()
flow[0] -= h/2
flow[1] -= w/2

img = flow_to_image(flow)
torch.save(img, "expected_flow.pt")
write_jpeg(img, "lol.jpg")  # if you want to look at it
```
lol


Then the test would compare the output of `img_to_tensor()` with the `expected_flow.pt` tensor that we would save to disk (and upload to the repo). It's only 32K so it's quite light, and I don't expect it to ever change.

NicolasHug (approved):
Thanks a lot @oke-aditya !!

I made a minor fix to the tests in  eeff9b5 . LGTM, I'll merge when green :) !

Would you be interested in submitting a follow-up PR to add support for batches of flows?

oke-aditya:
Yes will work on it soon! Thanks for fixes

NicolasHug:
CC @vadimkantorov, IIRC a flow -> image conversion util was something you had suggested in the past

PR #5135, opened 2021-12-29
only set random generator if shuffle=true

xiaohu2015:
This pr is about https://github.com/pytorch/vision/pull/5051

set random_generator only when shuffle=true, just like https://pytorch.org/docs/stable/_modules/torch/utils/data/distributed.html#DistributedSampler

cc @datumbox


datumbox (approved):
LGTM, thanks @xiaohu2015.

PR #5136, opened 2021-12-29
refactor prototype datasets tests

pmeier:
This refactors the prototype datasets in two ways:

1. Lazily generate the fake data. Before, all test data was generated at collection time, which is very time consuming if you only want to run some tests selective.
2. Only generate each file once. Before, the caching of the generated data was only half-baked since the files were cached on a per-config basis. If multiple configs were using the same file it was regenerated every time. This is especially problematic if the common file contains random elements that could vary between configs.

### DIFF
--- a/test/builtin_dataset_mocks.py
+++ b/test/builtin_dataset_mocks.py
@@ -1,35 +1,33 @@
+import contextlib
 import functools
 import gzip
+import itertools
 import json
 import lzma
 import pathlib
 import pickle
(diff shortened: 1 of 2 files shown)


NicolasHug (approved):
Thanks @pmeier , I haven't looked in details so I'll stamp.
I'll make sure to provide in-depth feedback on this once I implement some of the FLAVA datasets in the prototype area

PR #5138, opened 2021-12-30
Country dataset

puhuk:
cc @pmeier


torchvision/datasets/country211.py
```diff
@@ -0,0 +1,73 @@
...
+    """`The Country211 Data Set <https://github.com/openai/CLIP/blob/main/data/country211.md>`_.
+
+    filtered the YFCC100m dataset that have GPS coordinate corresponding to a ISO-3166 country code
+    and created a balanced dataset by sampling 150 train images, 50 validation images,
+    and 100 test images images for each country.
+
+
+    Args:
+        root (string): Root directory of the dataset.
+        split (string, optional): The dataset split, supports ``"train"`` (default) and ``"test"``.
```
pmeier, line 21: This is missing the `"valid"` split.

torchvision/datasets/country211.py
```diff
@@ -0,0 +1,73 @@
...
+        extensions: Tuple[str, ...] = ("jpg", "png"),
+    ) -> None:
+        super().__init__(root, transform=transform, target_transform=target_transform)
+        self._split = verify_str_arg(split, "split", ("train", "valid", "test"))
+        self._base_folder = Path(self.root) / "country211"
+
+        if download:
+            self._download()
+
+        self.split_folder = self._base_folder / self._split
```
pmeier, line 45: We should have an extra `self._check_exists()` here and raise an error, if it comes back negative. Otherwise the user will probably get an uninformative error while setting `download=False` and the data is not present.

torchvision/datasets/country211.py
```diff
@@ -0,0 +1,73 @@
...
+
+    Args:
+        root (string): Root directory of the dataset.
+        split (string, optional): The dataset split, supports ``"train"`` (default) and ``"test"``.
+        transform (callable, optional): A function/transform that  takes in an PIL image and returns a transformed
+            version. E.g, ``transforms.RandomCrop``.
+        target_transform (callable, optional): A function/transform that takes in the target and transforms it.
+    """
+
+    _URL = "https://openaipublic.azureedge.net/clip/data/country211.tgz"
```
pmeier, line 27:
```suggestion
    _URL = "https://openaipublic.azureedge.net/clip/data/country211.tgz"
    _MD5 = "84988d7644798601126c29e9877aab6a"
```

torchvision/datasets/country211.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def _check_exists(self) -> bool:
+        return all(
+            folder.exists() and folder.is_dir() for folder in (Path(self.root), self._base_folder, self._images_folder)
+        )
+
+    def _download(self) -> None:
+        if self._check_exists():
+            return
+        download_and_extract_archive(self._URL, download_root=self.root, md5=None)
```
pmeier, line 73:
```suggestion
        download_and_extract_archive(self._URL, download_root=self.root, md5=self._MD5)
```

pmeier (commented):
Hey @puhuk, thanks a lot for the PR. I've added a few comments inline. There are three more issues:

1. You used `find_classes` and `make_dataset` to create the dataset. This is far better than doing it yourself, but it would be even better to simply subclass `ImageFolderDataset` so you don't have to worry about that at all. Basically, you only need to handle the input parsing and download and simply pass `os.path.join(root, "country", self._split)` to `super().__init__()` (along with the transforms of course). If you do that, you don't need to define `__len__` and `__getitem__` at all.
2. Could you add the dataset to our documentation? For that, simply add the class name [to the list in `docs/source/datasets.rst`](https://raw.githubusercontent.com/pytorch/vision/main/docs/source/datasets.rst).
3. Could you add tests for the dataset? You can create a `Country211TestCase` in [this file](https://github.com/pytorch/vision/blob/e65a857b5487a8493bc8a80a95d64d9f049de347/test/test_datasets.py#L2206). You basically only need to provide a minimal configuration as well as provide fake data. You can read [this docstring](https://github.com/pytorch/vision/blob/e65a857b5487a8493bc8a80a95d64d9f049de347/test/datasets_utils.py#L186) or look at the other test cases for examples. Due to the very easy structure, all the fake data can be created by `datasets_utils.create_image_folder`.

torchvision/datasets/country211.py
puhuk: @pmeier Thank you for your kind review. Let me check and correct it soon :)

puhuk:
@pmeier Thank you for your kind review. Let me check and correct it soon :)

puhuk:
@pmeier  I resend PR with your suggestion and test code.
But there are errors, could you advise me about how to resolve this.

ClassName.txt
```diff
@@ -0,0 +1,3 @@
+/a/abbey
```
pmeier, line 1: This file shouldn't be here, please remove it.

docs/source/datasets.rst
```diff
@@ -73,6 +73,7 @@ You can also create your own datasets using the provided :ref:`base classes <bas
     VOCSegmentation
     VOCDetection
     WIDERFace
+    Country211
```
pmeier, line 76: This list should alphabetically sorted. Please put it in the right place.

test/test_datasets.py
```diff
@@ -2205,5 +2205,49 @@ def inject_fake_data(self, tmpdir: str, config):
         return len(sampled_classes * n_samples_per_class)
 
 
+class Country211TestCase(datasets_utils.ImageDatasetTestCase):
+    DATASET_CLASS = datasets.Country211
+    FEATURE_TYPES = (PIL.Image.Image, int)
```
pmeier, line 2210: This is the default for `ImageDatasetTestCase` so we don't need to set it at all.

test/test_datasets.py
```diff
@@ -2205,5 +2205,49 @@ def inject_fake_data(self, tmpdir: str, config):
...
+        for cls in sampled_classes:
+            image_folder = root_folder / cls
+            im_paths.extend(
+                datasets_utils.create_image_folder(
+                    image_folder,
+                    name=cls,
+                    file_name_fn=lambda idx: f"{cls}_{idx:05d}.jpg",
+                    num_examples=num_images_per_class,
+                )
+            )
```
pmeier, line 2231: 1. As is, this will create images with the following path `{tmpdir}/country211/{cls}/{cls}`. So the class name occurs twice, while the split does not occur at all. 
2. The file name does not reflect the actual data. If you want it to be accurate it should be something like `f"{:7d}_{:.6f}_{:.6f}"` with all three randomly generated. On the other hand, the file name is irrelevant for the dataset. Thus, I suggest to go with the simplest choice and only append `.jpg` to the index.

```suggestion
        split_folder = pathlib.Path(tmpdir) / "country211" / config["split"]
        split_folder.mkdir(parents=True, exist_ok=True)

        return sum(
            [
                len(
                    datasets_utils.create_image_folder(
                        split_folder,
                        name=cls,
                        file_name_fn=lambda idx: f"{idx}.jpg",
                        num_examples=5,
                    )
                )
                for cls in ("AD", "BS", "GR")
            ]
        )
```

test/test_datasets.py
```diff
@@ -2205,5 +2205,49 @@ def inject_fake_data(self, tmpdir: str, config):
...
+                file.writelines(
+                    "\n".join(
+                        f"/{f_path.relative_to(root_folder).as_posix()}"
+                        for f_path in random.choices(im_paths, k=num_samples)
+                    )
+                )
+        else:
+            num_samples = len(im_paths)
+
+        return num_samples
```
pmeier, line 2249: This seems copy-pasted from another dataset. These files do not exist for the Country211 dataset.

```suggestion
```

torchvision/datasets/country211.py
```diff
@@ -0,0 +1,77 @@
...
+
+        if self.target_transform:
+            label = self.target_transform(label)
+
+        return image, label
+
+    def _check_exists(self) -> bool:
+        return all(
+            folder.exists() and folder.is_dir() for folder in (Path(self.root), self._base_folder, self._images_folder)
+        )
```
pmeier, line 72: 1. There is no `self._images_folder`
2. It is sufficient to only check `self._base_folder` since it is inside `self.root`, i.e. if it exists the parent also have to exist

```suggestion
        return self._base_folder.exists() and self._base_folder.is_dir()
```

torchvision/datasets/country211.py
```diff
@@ -0,0 +1,77 @@
...
+        if not self._check_exists():
+            raise RuntimeError("Dataset not found. You can use download=True to download it")
+
+        self.split_folder = self._base_folder / self._split
+
+        self.classes, class_to_idx = find_classes(str(self.split_folder))
+        self.samples = make_dataset(str(self.split_folder), class_to_idx, extensions, is_valid_file=None)
+
+    def __len__(self) -> int:
+        return len(self._image_files)
```
pmeier, line 55: `self._image_files` does not exist

```suggestion
        return len(self.samples)
```

pmeier (changes requested):
@puhuk I suggested some changes inline. After that, my comment for inheriting from `ImageFolderDataset` (https://github.com/pytorch/vision/pull/5138#pullrequestreview-842564390) is still open.

Don't worry about the merge conflicts yet, we can fix them before merge. Plus, there are some unrelated failing test that you also don't need to worry about.

test/test_datasets.py
puhuk: I mistakenly forgot to delete those lines.
Thanks for the review 👍

puhuk:
@pmeier I resend the PR with your review.
But I don't exactly understand how to implement by inheriting ``ImageFolderDataset``.
May I ask for your help with this.

test/test_datasets.py
```diff
@@ -2463,5 +2463,29 @@ def _meta_to_split_and_classification_ann(self, meta, idx):
         return (image_id, class_id, species, breed_id)
 
 
+class Country211TestCase(datasets_utils.ImageDatasetTestCase):
+    DATASET_CLASS = datasets.Country211
+    FEATURE_TYPES = (PIL.Image.Image, int)
```
pmeier, line 2468:
```suggestion
```

torchvision/datasets/country211.py
```diff
@@ -0,0 +1,56 @@
...
+from .folder import ImageFolder
+from .utils import verify_str_arg, download_and_extract_archive
+
+
+class Country211(ImageFolder):
+    """`The Country211 Data Set <https://github.com/openai/CLIP/blob/main/data/country211.md>`_.
+
+    filtered the YFCC100m dataset that have GPS coordinate corresponding to a ISO-3166 country code
+    and created a balanced dataset by sampling 150 train images, 50 validation images,
+    and 100 test images images for each country.
```
pmeier, line 13: This description is copy-pasted from another dataset. Please fix it.

torchvision/datasets/country211.py
```diff
@@ -0,0 +1,56 @@
...
+    _URL = "https://openaipublic.azureedge.net/clip/data/country211.tgz"
+    _MD5 = "84988d7644798601126c29e9877aab6a"
+
+    def __init__(
+        self,
+        root: str,
+        split: str = "train",
+        transform: Optional[Callable] = None,
+        target_transform: Optional[Callable] = None,
+        download: bool = True,
```
pmeier, line 33: There is no mention of the download flag in the docstring. Please add it.

pmeier (commented):
@puhuk I pushed the changes to inherit from `ImageFolder`. There are two things left to fix in the docstring. Afterwards this is good to go.

torchvision/datasets/country211.py
pmeier: My bad, I was confused by the amount of dataset PRs that I review ATM.

NicolasHug (approved):
Thanks a lot @puhuk for the PR and @pmeier for the review!

In 9eb4f11 I pushed some minor changes to address @pmeier 's last comments, and to make the tests slightly more robust. LGTM, I'll merge when green. Thanks again!

puhuk:
@pmeier Really thanks to your review and re-creation of my humble code.
Hope to contribute with better code in next time.

PR #5151, opened 2021-12-31
resolve redirection for HTTP resources

pmeier:
This is far from perfect, since the `mirrors` do not get resolved. So far we never hit an issue where this would be a problem, but still.

In the future, the plan is to switch the resources to [`iopath`](https://github.com/facebookresearch/iopath). If that is available, resolving will no longer generate a new resource, but rather change the underlying path in place.

cc @pmeier @bjuncek


torchvision/prototype/datasets/utils/_resource.py
```diff
@@ -141,8 +143,40 @@ def __init__(
...
+                "file_name",
+                "sha256",
+                "_preprocess",
+                "_loader",
+            )
+        }
+
+        gdrive_id = _get_google_drive_file_id(redirect_url)
+        if gdrive_id:
+            return GDriveResource(gdrive_id, **meta)
```
NicolasHug, line 170: Do we have current cases where an HTTP resource redirects to a Google drive?

torchvision/prototype/datasets/utils/_resource.py
```diff
@@ -141,8 +143,40 @@ def __init__(
...
+
+        http_resource = HttpResource(redirect_url, **meta)
+        http_resource._resolved = True
+        return http_resource
 
     def _download(self, root: pathlib.Path) -> None:
+        if not self.resolved:
+            return cast(OnlineResource, self.resolve())._download(root)
```
NicolasHug, line 178: Q: I'm seeing a few instances like this where we have to call `cast()` because the function we call returns an `Optional[...]`

Is there a less intrusive way of telling mypy that we *know* the output won't be `None`? Peherpas this is a configurable option?

In this very specific case though, it looks like the function never returns `None`, perhaps we could remove the `Optional` annotation?

torchvision/prototype/datasets/utils/_resource.py
```diff
@@ -141,8 +143,40 @@ def __init__(
...
+            )
+        }
+
+        gdrive_id = _get_google_drive_file_id(redirect_url)
+        if gdrive_id:
+            return GDriveResource(gdrive_id, **meta)
+
+        http_resource = HttpResource(redirect_url, **meta)
+        http_resource._resolved = True
+        return http_resource
```
NicolasHug, line 174: Q: Do we consider that this new `http_resource` is itself resolved? (the code doesn't, but should we?)

torchvision/prototype/datasets/utils/_resource.py
```diff
@@ -141,8 +143,40 @@ def __init__(
...
 
     def _download(self, root: pathlib.Path) -> None:
+        if not self.resolved:
+            return cast(OnlineResource, self.resolve())._download(root)
+
         for url in itertools.chain((self.url,), self.mirrors or ()):
             try:
                 download_url(url, str(root), filename=self.file_name, md5=None)
```
NicolasHug, line 182: In your first comment you wrote

> This is far from perfect, since the mirrors do not get resolved

Could we instead do the redirection here and handle the mirrors?

NicolasHug (approved):
Thanks @pmeier ,

I mostly have questions but I'll approve to unlock. For my own understanding, did we face cases in the past where not handling redirections was a problem?

torchvision/prototype/datasets/utils/_resource.py
pmeier: Yes. It is one of the datasets not yet ported to prototype. I've hit this while doing #5154.

torchvision/prototype/datasets/utils/_resource.py
pmeier: Yes and the code does this. In L173 we set the `_resolved` attribute, which marks this as resolved.

NicolasHug: oops, I missed that thanks. What about the GDriveResource above?

torchvision/prototype/datasets/utils/_resource.py
pmeier: > Is there a less intrusive way of telling mypy that we _know_ the output won't be `None`? Peherpas this is a configurable option?

I don't think so as would go against the idea of type checking. If something is annotated as `Optional[Foo]` you should not use it as if it was an instance of `Foo`.

> In this very specific case though, it looks like the function never returns `None`, perhaps we could remove the `Optional` annotation?

Yes, that was a mistake on my side.

torchvision/prototype/datasets/utils/_resource.py
pmeier: We could, but that would make stuff more complicated. Given that the plan is to refactor this anyway with `iopatch` and we currently only have a single datasets that has official mirrors, I wouldn't sweat it at the moment.

Issue #1654, opened 2019-12-10
[proposal] CUB 200-2011 dataset


vadimkantorov:
This is a very widely used dataset in metric learning research and fine-grained recognition ("Birds dataset"): http://www.vision.caltech.edu/visipedia/CUB-200-2011.html

At few years ago I abused ImageFolder and CIFAR10 multiple inheritance classes to add this dataset:

```python
from torchvision.datasets import ImageFolder
from torchvision.datasets.utils import check_integrity, download_and_extract_archive

class CUB2011(ImageFolder):
	base_folder = 'CUB_200_2011/images'
	url = 'http://www.vision.caltech.edu/visipedia-data/CUB-200-2011/CUB_200_2011.tgz'
	filename = 'CUB_200_2011.tgz'
	tgz_md5 = '97eceeb196236b17998738112f37df78'

	train_list = [
		['001.Black_footed_Albatross/Black_Footed_Albatross_0001_796111.jpg', '4c84da568f89519f84640c54b7fba7c2'],
		['002.Laysan_Albatross/Laysan_Albatross_0001_545.jpg', 'e7db63424d0e384dba02aacaf298cdc0'],
	]
	test_list = [
		['198.Rock_Wren/Rock_Wren_0001_189289.jpg', '487d082f1fbd58faa7b08aa5ede3cc00'],
		['200.Common_Yellowthroat/Common_Yellowthroat_0003_190521.jpg', '96fd60ce4b4805e64368efc32bf5c6fe']
	]

	def _check_integrity(self):
		root = self.root
		for fentry in (self.train_list + self.test_list):
			filename, md5 = fentry[0], fentry[1]
			fpath = os.path.join(root, self.base_folder, filename)
			if not check_integrity(fpath, md5):
				return False
		return True
	
	def download(self):
		if self._check_integrity():
			print('Files already downloaded and verified')
			return
		download_and_extract_archive(self.url, self.root, filename=self.filename, md5=self.tgz_md5)

	def __init__(self, root, transform=None, target_transform=None, download=False, **kwargs):
		self.root = root
		if download:
			self.download()

		if not self._check_integrity():
			raise RuntimeError('Dataset not found or corrupted.' +
							   ' You can use download=True to download it')
		super().__init__(self, os.path.join(root, self.base_folder),
			transform=transform, target_transform=target_transform, **kwargs)
```

I extended this class for a metric learning task like this:
```python
class CUB2011MetricLearning(CUB2011):
	num_training_classes = 100

[...]

PR #5154, opened 2022-01-03
add CUB200 prototype datasets

pmeier:
This is the only popular vision dataset (according to [this list](https://paperswithcode.com/datasets?)) that we are currently missing.

cc @pmeier @bjuncek


vadimkantorov:
will also fix this: https://github.com/pytorch/vision/issues/1654

vadimkantorov:
Another frequent dataset for metric learning is Stanford Online Products...

torchvision/prototype/datasets/_builtin/cub.py
```diff
@@ -0,0 +1,233 @@
...
+        elif path.name == "train_test_split.txt":
+            return 1
+        elif path.name == "images.txt":
+            return 2
+        elif path.name == "bounding_boxes.txt":
+            return 3
+        else:
+            return None
+
+    _2011_SPLIT_MAP = {
```
NicolasHug, line 97: Nit: maybe inline this below since we just use it once?

torchvision/prototype/datasets/_builtin/cub.py
```diff
@@ -0,0 +1,233 @@
...
+            getitem(0),
+            buffer_size=INFINITE_BUFFER_SIZE,
+        )
+        return Mapper(dp, functools.partial(self._collate_and_decode_sample, year=config.year, decoder=decoder))
+
+    def _generate_categories(self, root: pathlib.Path) -> List[str]:
+        config = self.info.make_config(year="2011")
+        resources = self.resources(config)
+
+        dp = resources[0].load(pathlib.Path(root) / self.name)
```
NicolasHug, line 229: Similarly to the changes in https://github.com/pytorch/vision/pull/5188, should this just be

```suggestion
        dp = resources[0].load(root)
```
?

torchvision/prototype/datasets/_builtin/cub.py
```diff
@@ -0,0 +1,233 @@
...
+        self, data: Tuple[str, Tuple[str, io.IOBase]], *, decoder: Optional[Callable[[io.IOBase], torch.Tensor]]
+    ) -> Dict[str, Any]:
+        _, (path, buffer) = data
+        content = read_mat(buffer)
+        return dict(
+            ann_path=path,
+            bounding_box=BoundingBox(
+                [int(content["bbox"][coord]) for coord in ("left", "bottom", "right", "top")], format="xyxy"
+            ),
+            segmentation=Feature(content["seg"]),
```
NicolasHug, line 141: Could you help me understand why we use the `decoder` for the `Feature(...)` in `_2011_decode_ann`, but not in this function?

NicolasHug (approved):
Thanks @pmeier , some minor comments / questions but LGTM

torchvision/prototype/datasets/_builtin/cub.py
pmeier: Yes, good catch.

torchvision/prototype/datasets/_builtin/cub.py
pmeier: The `decoder` is used to turn an open file handle into a the pixel values as tensor. In this case the raw pixel values are already present as a `numpy.ndarray` so there is nothing left to decode.

pmeier:
Test failure is unrelated.

vadimkantorov:
I guess it would be nice to include some okay metric learning tutorial / reference impl using CUB/Stanford Online Products, this could be a useful example of new-style dataset usage

PR #5155, opened 2022-01-03
add SVHN prototype dataset

pmeier:


NicolasHug (approved):
Thanks @pmeier ! I only took a quick look but LGTM, we'll make sure to add tests and potential fixed in follow up PRs

pmeier:
Let's wait for #5187 to land and add tests here, since this PR is not urgent.

PR #5085, opened 2021-12-10
Adding pretrained ViT weights

sallysyw:
In #4594, we added ViT models' architecture to torchvision prototype. 
In this PR, we are going to add pretrained weights for ViT models :D

| Model      | Training Recipe | Training Command | Training Job ID | Epochs  | Nodes | Num of GPUs per node | Batch Size per GPU | Global batch\_size | Image Size | Representation Size | Original Paper Acc@1 | Classy Vision Acc@1 | ACC@1  | ACC@5  | Testing Job ID | Testing Command                                                                                                                                                                                                                                                                                 |
| ---------- | ---------- | ------ | --------------- | ------- | ----- | -------------------- | ------------------ | ------------------ | ---------- | ------------------- | -------------- | ------------------ | ------ | ------ | -------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| vit\_b\_16   | Close to DeiT | PYTHONPATH=$PYTHONPATH:`pwd` python -u run_with_submitit.py --timeout 3000 --ngpus 8 --nodes 8 --partition train --model vit_b_16 --batch-size 64 --epochs 300 --opt adamw --lr 0.003 --wd 0.3 --lr-scheduler cosineannealinglr --lr-warmup-method linear --lr-warmup-epochs 30 --lr-warmup-decay 0.033 --amp --label-smoothing 0.11 --mixup-alpha 0.2 --auto-augment ra --clip-grad-norm 1 --ra-sampler --cutmix-alpha 1.0 --model-ema |10266    | 268/300 | 8     | 8                    | 64                 | 4096               | 224        | None                | 77.91          | 78.98              | 81.072 | 95.318 | 13799  | PYTHONPATH=$PYTHONPATH:\`pwd\` python -u run\_with\_submitit.py --timeout 3000 --ngpus 1 --nodes 1 --partition train --model vit\_b\_16 --batch-size 1 --test-only --weights ViT\_B\_16\_Weights.ImageNet1K\_V1 |
[...]

PR #5156, opened 2022-01-04
Update the best training commands for ViT models

sallysyw:
As per #5085.

cc @datumbox

### DIFF
--- a/references/classification/README.md
+++ b/references/classification/README.md
@@ -144,26 +144,41 @@ torchrun --nproc_per_node=8 train.py\
 Here `$MODEL` is one of `regnet_x_32gf`, `regnet_y_16gf` and `regnet_y_32gf`.
 
 ### Vision Transformer
-
-#### Base models
+#### vit_b_16
 ```
 torchrun --nproc_per_node=8 train.py\
-    --model $MODEL --epochs 300 --batch-size 64 --opt adamw --lr 0.003 --wd 0.3\
+    --model vit_b_16 --epochs 300 --batch-size 64 --opt adamw --lr 0.003 --wd 0.3\
     --lr-scheduler cosineannealinglr --lr-warmup-method linear --lr-warmup-epochs 30\
     --lr-warmup-decay 0.033 --amp --label-smoothing 0.11 --mixup-alpha 0.2 --auto-augment ra\
     --clip-grad-norm 1 --ra-sampler --cutmix-alpha 1.0 --model-ema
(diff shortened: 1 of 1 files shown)


references/classification/README.md
```diff
@@ -144,26 +144,41 @@ torchrun --nproc_per_node=8 train.py\
...
+torchrun --nproc_per_node=8 train.py\
+    --model vit_b_32 --epochs 300 --batch-size 64 --opt adamw --lr 0.003 --wd 0.3\
+    --lr-scheduler cosineannealinglr --lr-warmup-method linear --lr-warmup-epochs 30\
+    --lr-warmup-decay 0.033 --amp --label-smoothing 0.11 --mixup-alpha 0.2 --auto-augment imagenet\
+    --clip-grad-norm 1 --ra-sampler --cutmix-alpha 1.0 --model-ema
+```
+#### vit_l_16
+```
+torchrun --nproc_per_node=8 train.py\
+    --model vit_l_16 --epochs 600 --batch-size 64 --lr 0.5 --lr-scheduler cosineannealinglr\
```
datumbox, line 167: I believe that the table at https://github.com/pytorch/vision/pull/5085#issuecomment-1003247327 has the wrong information concerning the number of nodes, the batch-sizes etc. For example the record for `vit_l_16` is definitely wrong as I used 2 nodes instead of 8 that you mention. 

Moreover the commands as is they don't specify how many nodes we used. For some we use 8 nodes, for others 2 which means that if someone tries to reproduce our results it won't work. Typically the commands provided in this file are for 8 gpus (1 node), unless specified otherwise. I recommend to either "translate" the commands to their 8 GPU equivalents (by adjusting the batch-size) or specify the number of nodes per model here.

datumbox (commented):
@sallysyw I spotted a few inconsistencies on the commands, see below.

Would you mind bring these changes on the main PR-#5085? Note that typically documentation changes related to the [references](https://github.com/pytorch/vision/blob/main/references/classification/README.md) and [model accuracies](https://github.com/pytorch/vision/blob/main/docs/source/models.rst) of pre-trained weights come on the same PR that introduces them. We do this to facilitate reviewing. It also ensures that everything lands together keeping the main branch consistent.

references/classification/README.md
sallysyw: Ok, let me close this PR and include the changes you suggested into #5085 together with the weights.

PR #5133, opened 2021-12-28
add contribution instructions for prototype datasets

pmeier:
This is a preliminary guide to foster early contributions to the prototype datasets outside of the core team. You can find the rendered version [here](https://github.com/pmeier/vision/blob/datasets/readme/torchvision/prototype/datasets/_builtin/README.md). In the future this should probably moved to `CONTRIBUTING.md` or the into the documentation.

@ reviewers: All the changes apart from `README.md` are minor convenience changes that I found while writing the guide.

cc @pmeier @bjuncek

PR #5158, opened 2022-01-04
add prettier as non-code auto formatter

pmeier:
Follow-up to https://github.com/pytorch/vision/pull/5133#pullrequestreview-843429034

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,19 +3,23 @@ repos:
     rev: v4.0.1
     hooks:
       - id: check-docstring-first
-      - id: check-toml
-      - id: check-yaml
-        exclude: packaging/.*
       - id: mixed-line-ending
         args: [--fix=lf]
       - id: end-of-file-fixer
 
-  # - repo: https://github.com/asottile/pyupgrade
-  #   rev: v2.29.0
-  #   hooks:
-  #     - id: pyupgrade
-  #       args: [--py36-plus]
-  #       name: Upgrade code
+  - repo: https://github.com/pre-commit/mirrors-prettier
+    rev: v2.5.1
+    hooks:
+      - id: prettier
+        types_or:
+          - markdown
+          - toml
+          - yaml
+        exclude: |
+          (?x)^(
+              .circleci/config.yml|
+              packaging/.*
+          )$
 
   - repo: https://github.com/omnilib/ufmt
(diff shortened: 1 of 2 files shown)


NicolasHug:
Thanks @pmeier ,

if we add this as a pre-commit hook we would also need to add guidelines on how to run it manually. I feel like this would make our  contribution process a bit heavy, especially for new contributors. It would also make our contributing guidelines a bit longer, scaring away potential contributors.

Considering the small benefit (IMHO) that this check would bring, I would personally prefer not introducing the checks, and just collectively remember to limit line lengths in files.

pmeier:
> if we add this as a pre-commit hook we would also need to add guidelines on how to run it manually. I feel like this would make our contribution process a bit heavy, especially for new contributors. It would also make our contributing guidelines a bit longer, scaring away potential contributors.

We only explain `ufmt` and `flake8` in the contributing guide lines. For example 

https://github.com/pytorch/vision/blob/bbeb32035c0df752a0fe0cd5921de537e7f68d72/.pre-commit-config.yaml#L37

will be enforced, but is not mentioned. Worse, 

https://github.com/pytorch/vision/blob/bbeb32035c0df752a0fe0cd5921de537e7f68d72/.pre-commit-config.yaml#L5-L11

are not even executable without `pre-commit`.

`prettier` would be in between. It can be executed as an executable, but since it is not a Python package, the contributor would need to setup a JS environment.

Can't we make `pre-commit` the default way? IIRC, the reason to not do that were slow hooks. We could shorten the contribution guide lines, by explaining how to run with `pre-commit` and additionally how to skip hooks if need be.

> Considering the small benefit (IMHO) that this check would bring

Note that this is not only line-wrapping in markdown files. This also automatically formats yaml and toml files. Previously, we only checked if there were syntactically correct, but didn't enforce an format. With this PR, [20 files](https://app.circleci.com/pipelines/github/pytorch/vision/13686/workflows/e6ed3a2d-e7b2-43af-bff5-1d8f61d4f96d/jobs/1101577) would get fixed.

NicolasHug:
We don't explain how to use the other checks used in the hooks because those very rarely fail, or are very rarely triggered by most non core-devs.
Describing them in the contributing guidelines would mainly make our guide longer for very little benefit.

Personally I don't feel like formatting yaml or even .md files is such an important priority.

PR #5086, opened 2021-12-10
Update readme.md with ViT training command

sallysyw:
As titled.

PR #5159, opened 2022-01-04
Remove incorrect ViT recipe commands.

datumbox:
Removes the training commands of ViT which were introduced prematurely at #5086. 

The commands are not up-to-date and since the weights are not yet merged (see #5085) they don't reflect what we have on main branch.

cc @sallysyw


prabhat00155 (approved):
Thanks @datumbox!

PR #5164, opened 2022-01-05
Cleanups for FLAVA datasets

NicolasHug:
~All datasets but 2 have `download=False` as the default, so this PR sets the default to False as well for `Food101` and `DTD` for consistency. It also documents the `download` parameter for `Food101` which was missing from the Docstring.~

See https://github.com/pytorch/vision/pull/5164#issuecomment-1005813311 for complete set of changes

cc @pmeier


pmeier (approved):

NicolasHug:
Thanks for the review @pmeier . Following up on your https://github.com/pytorch/vision/pull/5130#discussion_r778910110, let's use this PR to 

- make sure all datasets have `download=False`
- make sure all datasets have `download` *after* the `transforms` parameter.

I'll mark it as draft and we can come back to this once the rest of the PRs are merged.

pmeier:
@NicolasHug You are only talking about the "FLAVA" datasets here, right? Because for other datasets that would be BC breaking and I want to avoid that, since we probably don't have time for a deprecation cycle before the API is deprecated in general.

NicolasHug:
Fully agreed @pmeier , sorry for not being clearer in my comment above

pmeier:
https://github.com/pytorch/vision/blob/49468279d9070a5631b6e0198ee562c00ecedb10/torchvision/datasets/pcam.py#L78

is an anti-pattern. Since `h5py` has no annotations, it is better to ignore it globally rather than locally like

https://github.com/pytorch/vision/blob/49468279d9070a5631b6e0198ee562c00ecedb10/mypy.ini#L117-L119

torchvision/datasets/gtsrb.py
```diff
@@ -91,16 +76,26 @@ def __getitem__(self, index: int) -> Tuple[Any, Any]:
...
         if self._check_exists():
             return
 
-        download_and_extract_archive(self._urls[self.train], download_root=self.root, md5=self._md5s[self.train])
-
-        if not self.train:
-            # Download Ground Truth for the test set
+        if self._split == "train":
+            download_and_extract_archive(
+                "https://sid.erda.dk/public/archives/daaeac0d7ce1152aea9b61d9f1e19370/GTSRB-Training_fixed.zip",
```
pmeier, line 87: Could we split off the URL root here to make this more concise?

pmeier (approved):
One minor comment inline. Plus let's also resolve https://github.com/pytorch/vision/pull/5220#discussion_r787841892. Otherwise, LGTM if CI is green! Thanks @NicolasHug

torchvision/datasets/rendered_sst2.py
```diff
@@ -53,18 +54,13 @@ def __init__(
         if not self._check_exists():
             raise RuntimeError("Dataset not found. You can use download=True to download it")
 
-        self._labels = []
-        self._image_files = []
-
-        for p in (self._base_folder / self._split_to_folder[self._split]).glob("**/*.png"):
-            self._labels.append(self.class_to_idx[p.parent.name])
-            self._image_files.append(p)
+        self._samples = make_dataset(str(self._base_folder / self._split_to_folder[self._split]), extensions="png")
```
pmeier, line 57:
```suggestion
        self._samples = make_dataset(str(self._base_folder / self._split_to_folder[self._split]), extensions=(".png",))
```

NicolasHug: I believe it ultimately comes down to `endswith` which accepts tuples but also just plain strings.
I think the type annotations are incorrect here.

pmeier: Let me send a PR to fix that.

NicolasHug:
Failure is unrelated, I'll merge. Thanks for the review!

Issue #5162, opened 2022-01-04
fix JPEG reference tests


pmeier:
We are currently skipping JPEG reference tests on Windows, since the Windows wheels of `Pillow` are built against `libjpeg-turbo` while we build `torchvision` against `libjpeg`

https://github.com/pytorch/vision/blob/bbeb32035c0df752a0fe0cd5921de537e7f68d72/test/test_image.py#L481

As of [`Pillow==9.0.0`](https://pillow.readthedocs.io/en/stable/releasenotes/9.0.0.html#switched-to-libjpeg-turbo-in-macos-and-linux-wheels), all wheels are now built against `libjpeg-turbo`, which leads to the [CI failures](https://app.circleci.com/pipelines/github/pytorch/vision/13688/workflows/8d5ddae1-d149-4114-888d-8713aa668626/jobs/1101779) that are currently happening. Thus, these tests will now fail for every OS or at least only succeed by chance. 

There are multiple ways out in descending order of preference by me:

1. Install `Pillow` from conda-forge rather from PyPI to enforce the same JPEG library for `torchvision` and `Pillow`.
2. Pin `Pillow<9` for unittest CI. This basically preserves the behavior we had before.
3. Remove these tests completely.

The only downside of 1. that I see is that `Pillow` on conda-forge is released later than on PyPI. IIUC, the conda-forge package is community managed and they only start building when it is released on PyPI. If there are build failures it might take some time until the versions converge again. In the meantime, our tests would not pick up on any incompatibilities with `Pillow` in general not just for JPEG I/O. Thus, we would need to rely on user feedback that something is not working.

Given that this is an dependency issue, I would be fine with waiting for user reports in the time of `Pillow` version divergence between conda-forge and PyPI. If we don't want to accept that, we could also hide all JPEG compatibility tests behind a runtime flag and add another CI job that only runs these tests while enforcing 1. from above.

cc @pmeier @NicolasHug @datumbox

NicolasHug:
Thanks a lot for the summary and for opening this issue @pmeier 

In #5169 I temporarily skipped the offending tests unconditionally, for the CI to go back to green.

Paraphrasing what you mentioned about option 1: there's the risk that PIL releases a new version, and that we also release torchvision shortly after, *before* the conda-forge PIL channel gets updated with PIL's latest. In this case, we might miss bugs (related to jpeg, or to other parts where torchvision interacts with PIL, like the transforms).

[...]

PR #5169, opened 2022-01-06
Skip jpeg comparison tests with PIL

NicolasHug:
Linux and MacOS CIs are failing because the latest PIL 9 version now relies on libjpeg-turbo (it used to be the case only for Windows). Torchvision relies on libjpeg, not libjpeg-turbo, hence the comparison test failures.

We skip them in this PR to put CI back to green. This is only temporary until we find a better solution, tracked in https://github.com/pytorch/vision/issues/5162.


pmeier (approved):
Thanks!

Issue #5088, opened 2021-12-11
[FEEDBACK] Multi-weight support prototype API


datumbox:
### 🚀 The feature

-

### Motivation, pitch

-

cceyda:
I was looking at the example in the article and the one thing that jumped out at me as a potential source of confusion (especially for newcomers) was `weights.transforms()`. Although `weights` is just the variable name here, and I understand the need to associate the specific model('s weights) and its transform.
I think maybe what I don't like is using the word "weights" for the whole "model" abstraction.
One potential way to avoid confusion would be to be more specific.
so "PM.ResNet50.ImageNet1K_V1" would be the models interface(?) and 'weights','transforms','meta' would be accessed from there like;

```python
resnet = PM.ResNet50.ImageNet1K_V1
weights = resnet.weights
preprocess = resnet.transforms()
class_ids = resnet.meta["categories"]
```
And the example would turn into

```python
from torchvision.prototype import models as PM

# Step 1: Initialize model
resnet = PM.ResNet50.ImageNet1K_V1 #Less verbose than ResNet50_Weights also stops the ResNet50_weights confusion for those not using autocomplete IDEs
model = PM.resnet50(weights=resnet.weights) # or maybe just resnet
model.eval()

# Step 2: Initialize the inference transforms
preprocess = resnet.transforms()

# Step 3: Apply inference preprocessing transforms
batch = preprocess(img).unsqueeze(0)
prediction = model(batch).squeeze(0).softmax(0)

# Step 4: Use the model and print the predicted category
class_id = prediction.argmax().item()
score = prediction[class_id].item()
category_name = resnet.meta["categories"][class_id] # I like meta attributes
print(f"{category_name}: {100 * score}*%*")
```

Of course all this is a matter of style/preference, and there might be reasons it was implemented that way. I haven't looked at the code.

Overall I really like this feature, it will surely make life easier ❤️ no more copy pasting transforms, hurray! 🥳

datumbox:
@cceyda Thanks for taking the time to provide detailed feedback. 😄 

[...]

Issue #4996, opened 2021-11-26
Adding a list of model's param size( MB ) and layernum of models in torchvision with a standard input shape like 3,224,224


Force1ess:
### 📚 The doc issue

Torchvision is lacking a table of models' size and layernum

### Suggest a potential alternative/fix

Recently, I'm working on choosing some models from torchvision.
And as we all know, a model's size and layers are critical to many aspects.
Especially for beginners who don't familiar with these models, this table can help a lot.
If you think this is necessary, I'm glad to help collect these models' size and layernum and pull a request

datumbox:
Thanks for the proposal. I agree.

@NicolasHug We should consider adding the model size in the meta-data of the new model API and expose it via the documentation.

PR #5170, opened 2022-01-06
Improved meta-data for models

datumbox:
Enhance the meta-data information stored for each model by including the following info:
- Type of task
- Network architecture
- Year of publication
- Number of parameters

cc @datumbox @bjuncek


torchvision/prototype/models/detection/mask_rcnn.py
```diff
@@ -28,6 +28,9 @@ class MaskRCNN_ResNet50_FPN_Weights(WeightsEnum):
         url="https://download.pytorch.org/models/maskrcnn_resnet50_fpn_coco-bf2d0c1e.pth",
         transforms=CocoEval,
         meta={
+            "task": "image_object_detection",
+            "architecture": "FasterRCNN",
+            "publication_year": 2015,
```
fmassa, line 33: Mask R-CNN was published in a different paper than Faster R-CNN (2017)

torchvision/prototype/models/detection/mask_rcnn.py
```diff
@@ -28,6 +28,9 @@ class MaskRCNN_ResNet50_FPN_Weights(WeightsEnum):
         url="https://download.pytorch.org/models/maskrcnn_resnet50_fpn_coco-bf2d0c1e.pth",
         transforms=CocoEval,
         meta={
+            "task": "image_object_detection",
+            "architecture": "FasterRCNN",
```
fmassa, line 32: I would say the `architecture` for this one should probably be `MaskRCNN`, as it's an extension of Faster R-CNN which has its own paper https://paperswithcode.com/paper/mask-r-cnn

torchvision/prototype/models/detection/mask_rcnn.py
```diff
@@ -28,6 +28,9 @@ class MaskRCNN_ResNet50_FPN_Weights(WeightsEnum):
         url="https://download.pytorch.org/models/maskrcnn_resnet50_fpn_coco-bf2d0c1e.pth",
         transforms=CocoEval,
         meta={
+            "task": "image_object_detection",
```
fmassa, line 31: I'm unclear on what to put here.
Indeed, this model could also be categorized as `image_instance_segmentation`. Maybe having `task` be a list would be of help here?

Ensuring that this is adapted with paperswithcode categorization would be good as well https://paperswithcode.com/paper/mask-r-cnn

torchvision/prototype/models/detection/keypoint_rcnn.py
```diff
@@ -24,6 +24,9 @@
 
 
 _COMMON_META = {
+    "task": "image_object_detection",
+    "architecture": "FasterRCNN",
+    "publication_year": 2015,
```
fmassa, line 29: Same comments as for MaskRCNN -- KeypointRCNN was introduced in the Mask R-CNN paper. In this case you could potentially use the same `architecture` though, although I could also see it being its own architecture

torchvision/prototype/models/detection/mask_rcnn.py
datumbox: Yeah same here. I'll leave as is for now and review in more detail on the future.

fmassa (approved):
LGTM

torchvision/prototype/models/detection/keypoint_rcnn.py
```diff
@@ -24,6 +24,9 @@
 
 
 _COMMON_META = {
+    "task": "image_object_detection",
```
oke-aditya, line 27: image keypoint detection. I guess?

oke-aditya:
Could we add size of the model in MBs #4996 ? Anything about FLOPS of computation / Runtime statistics too would be cool.

(Although it's hard to say as runtime highly various from system to system and configs)

datumbox:
@oke-aditya Sounds like good suggestions. Would you like to send a PR that adds model sizes in? Ideally with a unit-test to compare the declared meta-data size with the actual size (you are guaranteed that they are available locally but you might need to be creative on how to do this; I got some similar checks here in this PR for model size in params).

FLOPS would be really cool, but it's not always straightforward to automatically estimate this for all models (for example detection models). I'm open for any ideas. BTW we've got a feedback issue at #5088 were we keep proposals about the new API centralized. Feel free to post more there if you have additional ideas. :)

oke-aditya:
I would need some time for this as it's been hectic week at office :( (It's literally odd to code in Java in morning and python at night)

PR #5174, opened 2022-01-07
Expose raft_large and raft_small to torchhub

NicolasHug:
Minor PR to add `raft_large()` and `raft_small()` to our `hubconf.py` file so that they can be exposed through torchhub.


datumbox (approved):
LGTM.

PR #5175, opened 2022-01-07
Add RegNet model variant regnet_y_128gf

kazhang:
I'm curious to know if regnet_y_128gf can fly pass GPU tests.


datumbox:
@kazhang The tests didn't pass for a different reason than model size:
```
TypeError: _regnet() missing 1 required positional argument: 'progress'
```

I would rerun by changing to the following:
```python
def regnet_y_128gf(progress: bool = True, **kwargs: Any) -> RegNet:
```

PR #5176, opened 2022-01-07
add regnet_y_128gf factory function

kazhang:
Add a factory function for RegNetY 128GF.

cc @datumbox


sallysyw:
Yeah...same for `vit_h_14`. It's always the tests on windows that fail... I tried bumping up the memory allocation (by upgrading CI instance type) but it still didn't help with my case :)
Probably you can give it a try:
https://github.com/pytorch/vision/blob/49ec677c970256a377293beb5902d6a238a3b90e/.circleci/config.yml#L9-L20
The available executor types on CircleCI can be found on [this documentation](https://circleci.com/docs/2.0/configuration-reference/).

sallysyw:
Yeah...same for `vit_h_14`. It's always the tests on windows that fail... I tried bumping up the memory allocation (by upgrading CI instance type) but it still didn't help with my case :)
Probably you can give it a try:
https://github.com/pytorch/vision/blob/49ec677c970256a377293beb5902d6a238a3b90e/.circleci/config.yml#L9-L20
The available executor types on CircleCI can be found on [this documentation](https://circleci.com/docs/2.0/configuration-reference/).

kazhang:
The tests on Windows failed when saving model torchscript. I tried to use `tempfile.NamedTemporaryFile` instead of bytes io, hoping to avoid OOM, but the temporary file can't be opened to write on windows.

datumbox:
@kazhang It seems to work now. The failing test is not related.

@prabhat00155 The test.test_videoapi.TestVideoApi test has been [failing](https://app.circleci.com/pipelines/github/pytorch/vision/13889/workflows/62bf350f-c0a9-4506-a09f-9dbc56b7cf1b/jobs/1117960) periodically for a while now. Can we adjust the thresholds or modify to reduce flakiness?

prabhat00155:
> @prabhat00155 The test.test_videoapi.TestVideoApi test has been [failing](https://app.circleci.com/pipelines/github/pytorch/vision/13889/workflows/62bf350f-c0a9-4506-a09f-9dbc56b7cf1b/jobs/1117960) periodically for a while now. Can we adjust the thresholds or modify to reduce flakiness?

@bjuncek is looking into this.

kazhang:
The reason why `NamedTemporaryFile` is because
> Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms (it can be so used on Unix; **it cannot on Windows NT or later**)

T.I.L
`TemporaryDirectory` works.

datumbox:
@kazhang The [failures](https://app.circleci.com/pipelines/github/pytorch/vision/13906/workflows/44c3825a-a691-4502-bc10-a3f04ab79e0c/jobs/1119587) on Windows are related. It's a memory problem, might be hard to fix.

kazhang:
hmmm, saving and loading from temporary file caused failure in **detection** models, sounds weird to me.

datumbox:
@kazhang Previously we faced similar issues with ViT and I noticed that the biggest Windows instance available to us is "GPU Windows Medium". I'm not sure if there are larger instances available that we could use but you could try checking with CircleCI. My hypothesis for why Linux doesn't fail is that we either use a different type of instance or it uses the GPU memory more efficiently due to the lack of GUI.

test/test_models.py
```diff
@@ -126,16 +126,16 @@ def assert_export_import_module(m, args):
 
         def get_export_import_copy(m):
             """Save and load a TorchScript model"""
-            buffer = io.BytesIO()
-            torch.jit.save(m, buffer)
```
kazhang, line 130: Crash when saving large model to buffer.

test/test_models.py
```diff
@@ -126,16 +126,16 @@ def assert_export_import_module(m, args):
...
-            buffer.seek(0)
-            imported = torch.jit.load(buffer)
+            with TemporaryDirectory() as dir:
+                path = os.path.join(dir, "script.pt")
```
kazhang, line 130: `NamedTemporaryFile` doesn't work on windows since it can't be open twice.

datumbox (approved):
LGTM, thanks @kazhang!

PR #5177, opened 2022-01-08
Add Flowers102 dataset

zhiqwang:
Addresses Flowers102 dataset in #5108.

cc @pmeier @NicolasHug

### DIFF
--- a/test/test_datasets.py
+++ b/test/test_datasets.py
@@ -2463,5 +2463,44 @@ class OxfordIIITPetTestCase(datasets_utils.ImageDatasetTestCase):
         return (image_id, class_id, species, breed_id)
 
 
+class Flowers102TestCase(datasets_utils.ImageDatasetTestCase):
+    DATASET_CLASS = datasets.Flowers102
+    FEATURE_TYPES = (PIL.Image.Image, int)
+
+    ADDITIONAL_CONFIGS = datasets_utils.combinations_grid(split=("train", "valid", "test"))
+    REQUIRED_PACKAGES = ("scipy",)
+
+    def inject_fake_data(self, tmpdir: str, config):
+        base_folder = pathlib.Path(tmpdir) / "flowers-102"
+
+        num_classes = 3
+        num_images_per_split = dict(train=3, valid=3, test=4)
+        num_images_total = sum(num_images_per_split.values())
+        datasets_utils.create_image_folder(
+            base_folder,
+            "jpg",
+            file_name_fn=lambda idx: f"image_{idx + 1:05d}.jpg",
+            num_examples=num_images_total,
+        )
+
+        meta_folder = base_folder / "labels"
+        meta_folder.mkdir()
+
+        label_dict = dict(
+            labels=np.random.randint(1, num_classes + 1, size=(1, num_images_total), dtype=np.uint8),
+        )
+        datasets_utils.lazy_importer.scipy.io.savemat(str(meta_folder / "imagelabels.mat"), label_dict)
+
+        setid_mat = np.arange(1, num_images_total + 1, dtype=np.uint16)
+        setid_dict = dict(
+            trnid=setid_mat[: num_images_per_split["train"]].reshape(1, -1),
+            valid=setid_mat[num_images_per_split["train"] : -num_images_per_split["test"]].reshape(1, -1),
+            tstid=setid_mat[-num_images_per_split["test"] :].reshape(1, -1),
+        )
+        datasets_utils.lazy_importer.scipy.io.savemat(str(meta_folder / "setid.mat"), setid_dict)
+
+        return num_images_per_split[config["split"]]
+
+
 if __name__ == "__main__":
     unittest.main()
--- a/test/test_datasets.py
+++ b/test/test_datasets.py
@@ -2474,7 +2474,7 @@ class Flowers102TestCase(datasets_utils.ImageDatasetTestCase):
         base_folder = pathlib.Path(tmpdir) / "flowers-102"
 
         num_classes = 3
-        num_images_per_split = dict(train=3, valid=3, test=4)
+        num_images_per_split = dict(train=5, valid=4, test=3)
         num_images_total = sum(num_images_per_split.values())
         datasets_utils.create_image_folder(
             base_folder,
--- a/test/test_datasets.py
+++ b/test/test_datasets.py
@@ -2494,14 +2494,14 @@ class Flowers102TestCase(datasets_utils.ImageDatasetTestCase):
     DATASET_CLASS = datasets.Flowers102
     FEATURE_TYPES = (PIL.Image.Image, int)
 
-    ADDITIONAL_CONFIGS = datasets_utils.combinations_grid(split=("train", "valid", "test"))
+    ADDITIONAL_CONFIGS = datasets_utils.combinations_grid(split=("train", "val", "test"))
     REQUIRED_PACKAGES = ("scipy",)
 
     def inject_fake_data(self, tmpdir: str, config):
         base_folder = pathlib.Path(tmpdir) / "flowers-102"
 
         num_classes = 3
-        num_images_per_split = dict(train=5, valid=4, test=3)
+        num_images_per_split = dict(train=5, val=4, test=3)
         num_images_total = sum(num_images_per_split.values())
         datasets_utils.create_image_folder(
             base_folder,
--- a/test/test_datasets.py
+++ b/test/test_datasets.py
@@ -2483,13 +2483,10 @@ class Flowers102TestCase(datasets_utils.ImageDatasetTestCase):
             num_examples=num_images_total,
         )
 
-        meta_folder = base_folder / "labels"
-        meta_folder.mkdir()
-
         label_dict = dict(
             labels=np.random.randint(1, num_classes + 1, size=(1, num_images_total), dtype=np.uint8),
         )
-        datasets_utils.lazy_importer.scipy.io.savemat(str(meta_folder / "imagelabels.mat"), label_dict)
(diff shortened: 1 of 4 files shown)


torchvision/datasets/flowers102.py
```diff
@@ -0,0 +1,116 @@
...
+        transform (callable, optional): A function/transform that takes in an PIL image and returns a
+            transformed version. E.g, ``transforms.RandomCrop``.
+        target_transform (callable, optional): A function/transform that takes in the target and transforms it.
+    """
+
+    def __init__(
+        self,
+        root: str,
+        split: str = "train",
+        download: bool = True,
```
pmeier, line 37: The download flag is missing from the docstring.

torchvision/datasets/flowers102.py
```diff
@@ -0,0 +1,116 @@
...
+        transform: Optional[Callable] = None,
+        target_transform: Optional[Callable] = None,
+    ) -> None:
+        super().__init__(root, transform=transform, target_transform=target_transform)
+        self._split = verify_str_arg(split, "split", ("train", "valid", "test"))
+        self._base_folder = Path(self.root) / "flowers-102"
+        self._meta_folder = self._base_folder / "labels"
```
pmeier, line 44: Do we actually need this? The images come in their separate folder anyway, so I feel like we can place the two other files in `self.root`.

torchvision/datasets/flowers102.py
```diff
@@ -0,0 +1,116 @@
...
+        self._images_folder = self._base_folder / "jpg"
+
+        if download:
+            self._download()
+
+        if not self._check_exists():
+            raise RuntimeError("Dataset not found. You can use download=True to download it")
+
+        self._labels = []
+        self._image_files = []
```
pmeier, line 54: There is no need to create them up here. Please move them down below to where they are actually used.

torchvision/datasets/flowers102.py
```diff
@@ -0,0 +1,116 @@
...
+
+        from scipy.io import loadmat
+
+        # Read the label ids
+        label_mat = loadmat(self._meta_folder / "imagelabels.mat")
+        labels = label_mat["labels"][0]
+
+        self.classes = np.unique(labels).tolist()
+        self.class_to_idx = dict(zip(self.classes, range(len(self.classes))))
```
pmeier, line 63: `self.classes` is traditionally reserved for human-readable names. What you define here would be more fitting for `self.labels`.

If I'm not mistaken, there is no way to get the human readable classes from the original data. So we could either use some external mapping such as [this](https://gist.github.com/JosephKJ/94c7728ed1a8e0cd87fe6a029769cde1) or [this](https://github.com/bdevnani3/oxfordflowers102-label-name-mapping/blob/main/mapping.json), but then we would also need to make sure that this is correct and from a quick glance, both variants are not the same. The other option is to simply drop the human-readable classes outright, since the authors didn't provide them either. I would go for the latter.

test/test_datasets.py
```diff
@@ -2463,5 +2463,45 @@ def _meta_to_split_and_classification_ann(self, meta, idx):
...
+    FEATURE_TYPES = (PIL.Image.Image, int)
+
+    ADDITIONAL_CONFIGS = datasets_utils.combinations_grid(split=("train", "valid", "test"))
+    REQUIRED_PACKAGES = ("scipy",)
+
+    def inject_fake_data(self, tmpdir: str, config):
+        base_folder = pathlib.Path(tmpdir) / "flowers-102"
+
+        num_classes = 3
+        num_images_per_split = dict(train=3, valid=3, test=4)
```
pmeier, line 2477: Let's make the number of samples different for all splits

```suggestion
        num_images_per_split = dict(train=5, valid=4, test=3)
```

pmeier (commented):
Hey @zhiqwang, thanks a lot for the PR. I have some minor comments inline.

torchvision/datasets/flowers102.py
zhiqwang: Hi @pmeier , The reason for creating this directory is to make it easier to determine if the two files were downloaded correctly.

torchvision/datasets/flowers102.py
zhiqwang: Added now.

torchvision/datasets/flowers102.py
pmeier: Given that we are talking about two files vs. one folder, I think we can simply check for the existence of both files? [`torchvision.datasets.utils.check_integrity`](https://github.com/pytorch/vision/blob/8265469b2526e8d520e2e5ede71454ac01e357aa/torchvision/datasets/utils.py#L66) was made for that.

torchvision/datasets/flowers102.py
zhiqwang: Hi @pmeier , I rename `self.classes` to `self.labels` here.

torchvision/datasets/flowers102.py
zhiqwang: Hi @pmeier , I've updated these with `check_integrity`.

torchvision/datasets/flowers102.py
```diff
@@ -0,0 +1,119 @@
...
+    def extra_repr(self) -> str:
+        return f"split={self._split}"
+
+    def _check_integrity(self):
+        st1 = check_integrity(str(self._base_folder / self.file_dict["image"][0]), self.file_dict["image"][1])
+        st2 = check_integrity(str(self._base_folder / self.file_dict["label"][0]), self.file_dict["label"][1])
+        st3 = check_integrity(str(self._base_folder / self.file_dict["setid"][0]), self.file_dict["setid"][1])
+        if not (st1 and st2 and st3):
+            return False
+        return True
```
pmeier, line 107: 1. We should not check the image archive, since the user can delete it after we have extracted it and the dataset should still work.
2. Using a loop makes this more concise without reducing readability:

```suggestion
        for id in ["label", "setid"]:
            filename, md5 = self.file_dict[id]
            if not check_integrity(str(self._base_folder / filename), md5):
                return False
        return True
```

torchvision/datasets/flowers102.py
```diff
@@ -0,0 +1,119 @@
...
+        if self._check_integrity():
+            print("Files already downloaded and verified")
+            return
+        download_and_extract_archive(
+            f"{self.download_url_prefix}{self.file_dict['image'][0]}",
+            str(self._base_folder),
+            md5=self.file_dict["image"][1],
+        )
+        download_url(f"{self.download_url_prefix}{self.file_dict['label'][0]}", str(self._base_folder))
+        download_url(f"{self.download_url_prefix}{self.file_dict['setid'][0]}", str(self._base_folder))
```
pmeier, line 119: Same as above

```suggestion
        for id in ["label", "setid"]:
            filename, md5 = self.file_dict[id]
            download_url(self.download_url_prefix + filename, str(self._base_folder), md5=md5)
```

torchvision/datasets/flowers102.py
pmeier: I think you misunderstood me. As is, `self.labels` is a list of all available labels. That is not useful. Either you provide human readable strings for the classes or simply remove `self.labels` and `self.label_to_index`. Since there is no official mapping between labels and class names, I would go for the latter and remove the attributes.

pmeier (approved):
Thanks a lot @zhiqwang! I've pushed a minor simplification to not prolong this. LGTM from my side!

torchvision/datasets/flowers102.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def _check_integrity(self):
+        for id in ["label", "setid"]:
+            filename, md5 = self.file_dict[id]
+            if not check_integrity(str(self._base_folder / filename), md5):
+                return False
+        return True
+
+    def download(self):
+        if not self._check_exists():
```
zhiqwang, line 105: Hi @pmeier , I move the folder existence check here. Otherwise if the `labels` or `setids` is missing, we will download the image archive.

NicolasHug: Could we perhaps simplify the logic here, and just download all the files if `_check_exists()` returns False?

I'm not sure that allowing the image archive to be there without the labels or setids is really useful?

NicolasHug (commented):
Thanks a lot @zhiqwang for the PR, and @pmeier for the review.

I made some very minor changes above, the only relevant one is that I changed `split="valid"` into `split="val"` for more consistency with other datasets.

This LGTM but I just have one question below regarding the download logic, LMK what you think

torchvision/datasets/flowers102.py
pmeier: IIRC the discrepancy in `download` was introduced after my review. I agree with Nicolas here. We should have one method that checks the existence / integrity of all files and one download method that downloads everything if the former method returns `False.`

zhiqwang: Hi @pmeier and @NicolasHug , I've reverted the change back.

torchvision/datasets/flowers102.py
```diff
@@ -0,0 +1,111 @@
...
+        if self.target_transform:
+            label = self.target_transform(label)
+
+        return image, label
+
+    def extra_repr(self) -> str:
+        return f"split={self._split}"
+
+    def _check_integrity(self):
+        for id in ["label", "setid"]:
```
pmeier, line 95: We still need to check if the image folder exists

```suggestion
        if not (self._images_folder.exists() and self._images_folder.isdir()):
            return False

        for id in ["label", "setid"]:
```

NicolasHug (approved):
Thanks a lot @zhiqwang !! Will merge when green

PR #5178, opened 2022-01-09
Adding fvgc_aircraft dataset

sallysyw:
Per #5108

image

cc @pmeier


pmeier (commented):
Hey @sallysyw, thanks a lot for the PR. I've one question before I actually dive into the PR: why do we want restructure the data into an image folder hierarchy? IIUC, we could also work with the given format, which I would prefer.

torchvision/datasets/fvgc_aircraft.py
```diff
@@ -0,0 +1,119 @@
...
+import shutil
+from typing import Any, Callable, List, Optional, Tuple
+
+import PIL.Image
+
+from .utils import download_and_extract_archive, verify_str_arg
+from .vision import VisionDataset
+
+
+class FVGCAircraft(VisionDataset):
```
NicolasHug, line 11: Minor typo :)
I think we'll also need to update other parts like the test class, the docstring below, etc.


```suggestion
class FGVCAircraft(VisionDataset):
```

torchvision/datasets/fvgc_aircraft.py
```diff
@@ -0,0 +1,119 @@
...
+    """
+
+    _URL = "https://www.robots.ox.ac.uk/~vgg/data/fgvc-aircraft/archives/"
+    _URL_FILE = "fgvc-aircraft-2013b.tar.gz"
+
+    def __init__(
+        self,
+        root: str,
+        split: str = "trainval",
+        download: Optional[str] = None,
```
NicolasHug, line 37: I think we usually set the default to `False` instead of `None`

```suggestion
        download: bool = False,
```

torchvision/datasets/fvgc_aircraft.py
```diff
@@ -0,0 +1,119 @@
...
+        transform: Optional[Callable] = None,
+        target_transform: Optional[Callable] = None,
+        **kwargs: Any,
```
NicolasHug, line 40: Looks like we're not using this parameter, we can probably remove it?

test/test_datasets.py
```diff
@@ -2206,6 +2206,52 @@ def inject_fake_data(self, tmpdir: str, config):
...
+    ADDITIONAL_CONFIGS = datasets_utils.combinations_grid(split=("train", "val", "trainval", "test"))
+
+    def inject_fake_data(self, tmpdir: str, config):
+        split = config["split"]
+        root_folder = pathlib.Path(tmpdir) / "fgvc-aircraft-2013b"
+        data_folder = root_folder / "data"
+
+        num_images_per_class = 5
+        variants = ["707-320", "Hawk T1", "Tornado"]
+        n_samples_per_class = 4 if split == "trainval" else 2
```
NicolasHug, line 2222: Nit, for our naming style consistency (although I personally prefer `n_` as well):

```suggestion
        num_samples_per_class = 4 if split == "trainval" else 2
```

test/test_datasets.py
```diff
@@ -2206,6 +2206,52 @@ def inject_fake_data(self, tmpdir: str, config):
...
+            images_variants.extend(
+                [
+                    f"{idx} {variant}"
+                    for idx in random.sample(
+                        range(i * num_images_per_class, (i + 1) * num_images_per_class), n_samples_per_class
+                    )
+                ]
+            )
+
+        varients_file = root_folder / "data" / "variants.txt"
```
NicolasHug, line 2243: Nit, minor typo

```suggestion
        variants_file = root_folder / "data" / "variants.txt"
```

NicolasHug (commented):
Thanks a lot for the PR Yiwen!

This looks great, I made a few minor comments below. As Philip noted above, we tend not to change the dataset underlying structure, partly so that users can also just manually download the files, should they want to. Was there a specific reason to re-organise the dataset structure?

Thanks!

sallysyw:
> Hey @sallysyw, thanks a lot for the PR. I've one question before I actually dive into the PR: why do we want restructure the data into an image folder hierarchy? IIUC, we could also work with the given format, which I would prefer.

Good point! I did that for no specific reason - just follow what VISSL did. But I agree that we'd better not copy these images to another folder after downloading to avoid additional overhead. Let me send a patch to address this issue. Thanks!  

@pmeier @NicolasHug

torchvision/datasets/fgvc_aircraft.py
```diff
@@ -0,0 +1,106 @@
...
+        labels_path = os.path.join(input_path, "data", f"images_variant_{self._split}.txt")
+
+        with open(labels_path, "r") as labels_file:
+            lines = [line.strip() for line in labels_file]
+            for line in lines:
+                line_list = line.split(" ")
+                image_name = line_list[0]
+                label_name = self._parse_aircraft_name(" ".join(line_list[1:]))
+                self._labels.append(self._label_name_to_idx[label_name])
+                self._image_files.append(os.path.join(image_data_folder, image_name + str(".jpg")))
```
pmeier, line 98: Given that this method again does nothing complicated, is quite short, and is only used once, I think we can inline it in `__init__`. This will also resolve the linting errors the CI is currently complaining about.

test/test_datasets.py
```diff
@@ -2206,6 +2206,52 @@ def inject_fake_data(self, tmpdir: str, config):
         return len(sampled_classes * n_samples_per_class)
 
 
+class FGVCAircraftTestCase(datasets_utils.ImageDatasetTestCase):
+    DATASET_CLASS = datasets.FGVCAircraft
+    FEATURE_TYPES = (PIL.Image.Image, int)
```
pmeier, line 2211: This is the default for `ImageDatasetTestCase` and thus not needed

```suggestion
```

pmeier (commented):
Hey @sallysyw, I have some minor comments inline.

Apart from that I'm wondering if it makes sense to add a parameter to select the different "levels" of the dataset. 

> - **Model**, e.g. Boeing 737-76J. Since certain models are nearly visually indistinguishable, this level is not used in the evaluation.
> - **Variant**, e.g. Boeing 737-700. A variant collapses all the models that are visually indistinguishable into one class. The dataset comprises 102 different variants.
> - **Family**, e.g. Boeing 737. The dataset comprises 70 different families.
> - **Manufacturer**, e.g. Boeing. The dataset comprises 41 different manufacturers.

We could have a `level="variant"` that also allows `"family"` and `"manufacturer"`. From what I see, we would only need to change the files we read from based on this parameter.

NicolasHug:
> Apart from that I'm wondering if it makes sense to add a parameter to select the different "levels" of the dataset.
> ...
> We could have a level="variant" that also allows "family" and "manufacturer". From what I see, we would only need to change the files we read from based on this parameter.

That's a great point. Ultimately our goal with this dataset is to support our colleagues which are implementing FLAVA, and the "reference" in this case is https://github.com/facebookresearch/vissl/blob/main/extra_scripts/datasets/create_fgvc_aircraft_data_files.py. From what I understand, they would only need to use the "variant" categories. Perhaps it would be OK to just implement the "variant" category in this PR to keep it simple, and to implement the rest if users request more categories?

sallysyw:
Thanks both of you for the review. For the categories, I think for completeness I will provide a parameter called "level" or "annotation_level" and set the default to be "variant", this way it covers the existing functionalities in FLAVA and still giving users flexibility to use other levels' annotations of this dataset.

@pmeier @NicolasHug

torchvision/datasets/fgvc_aircraft.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def _get_label_names(self, input_path: str) -> List[str]:
+        variants_file = os.path.join(input_path, "data", "variants.txt")
+        with open(variants_file, "r") as f:
+            return [self._parse_aircraft_name(line.strip()) for line in f]
+
+    def _parse_aircraft_name(self, name: str) -> str:
+        return name.replace("/", "-").replace(" ", "-")
```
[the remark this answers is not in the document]
sallysyw: Thanks for the suggestion, I think we can drop the "/ -> -" part but still need to replace " " with "-", because inside "images_variants_{split}.txt", the format for each line is like `f"{id} {name}"`, if `name` itself contains space we will need to deal with it after we split the line with space.

torchvision/datasets/fgvc_aircraft.py
```diff
@@ -0,0 +1,121 @@
...
+class FGVCAircraft(VisionDataset):
+    """`FGVC Aircraft <https://www.robots.ox.ac.uk/~vgg/data/fgvc-aircraft/>`_ Dataset.
+
+    The dataset contains 10,200 images of aircraft, with 100 images for each of 102
+    different aircraft model variants, most of which are airplanes.
+    Aircraft models are organized in a four-levels hierarchy. The four levels, from
+    finer to coarser, are:
+
+    - Model, e.g. Boeing 737-76J. Since certain models are nearly visually indistinguishable,
+        this level is not used in the evaluation.
```
pmeier, line 27: If we don't expose `annotation_level="model"`, we shouldn't include it in the docstring.

torchvision/datasets/fgvc_aircraft.py
```diff
@@ -0,0 +1,121 @@
...
+        annotation_level (str, optional): The annotation level, supports ``variant``,
+            ``family`` and ``manufacturer``.
+        transform (callable, optional): A function/transform that  takes in an PIL image
+            and returns a transformed version. E.g, ``transforms.RandomCrop``
+        target_transform (callable, optional): A function/transform that takes in the
+            target and transforms it.
+    """
+
+    _URL = "https://www.robots.ox.ac.uk/~vgg/data/fgvc-aircraft/archives/"
+    _URL_FILE = "fgvc-aircraft-2013b.tar.gz"
```
pmeier, line 49: It seems these two are never used separately. We should merge them.

torchvision/datasets/fgvc_aircraft.py
```diff
@@ -0,0 +1,121 @@
...
+import PIL.Image
+
+from .utils import download_and_extract_archive, verify_str_arg
+from .vision import VisionDataset
+
+annotation_level_to_file = {
+    "variant": "variants.txt",
+    "family": "families.txt",
+    "manufacturer": "manufacturers.txt",
+}
```
pmeier, line 15: Given that we only talk about a dictionary with three items that is used in a single location, maybe we can inline it?

```python
file_name = {
    "variant": "variants.txt",
    "family": "families.txt",
    "manufacturer": "manufacturers.txt",
}[self._annotation_level]
annotation_file = os.path.join(input_path, "data", file_name)
```

torchvision/datasets/fgvc_aircraft.py
```diff
@@ -0,0 +1,121 @@
...
+            self._download()
+
+        if not self._check_exists():
+            raise RuntimeError("Dataset not found. You can use download=True to download it")
+
+        self.classes = self._get_classes(self._data_path)
+
+        # Parse the downloaded files
+        self._image_folder = os.path.join(self.root, self._split)
+        self.class_to_idx = dict(zip(self.classes, range(len(self.classes))))
```
pmeier, line 77: This seems to be unused.

```suggestion
        self.classes = self._get_classes(self._data_path)
        self.class_to_idx = dict(zip(self.classes, range(len(self.classes))))
```

torchvision/datasets/fgvc_aircraft.py
```diff
@@ -0,0 +1,121 @@
...
+
+        if self.transform:
+            image = self.transform(image)
+
+        if self.target_transform:
+            label = self.target_transform(label)
+
+        return image, label
+
+    def _download(self):
```
pmeier, line 107:
```suggestion
    def _download(self) -> None:
```

test/test_datasets.py
```diff
@@ -2206,6 +2206,58 @@ def inject_fake_data(self, tmpdir: str, config):
...
+                        range(i * num_images_per_class, (i + 1) * num_images_per_class), num_samples_per_class
+                    )
+                ]
+            )
+
+        annotation_file = root_folder / "data" / annotation_level_to_file[annotation_level]
+        images_annotation_file = root_folder / "data" / f"images_{annotation_level}_{split}.txt"
+
+        with open(annotation_file, "w") as file:
+            file.write("\n".join(classes))
```
pmeier, line 2253: Can we move this further to the top given that this is somewhat static and does not depend on the other data generated in the loop above?

torchvision/datasets/fgvc_aircraft.py
```diff
@@ -0,0 +1,121 @@
...
+            return
+        download_and_extract_archive(self._URL + self._URL_FILE, self.root)
+
+    def _check_exists(self) -> bool:
+        return os.path.exists(self._data_path) and os.path.isdir(self._data_path)
+
+    def _get_classes(self, input_path: str) -> List[str]:
+        annotation_file = os.path.join(input_path, "data", annotation_level_to_file[self._annotation_level])
+        with open(annotation_file, "r") as f:
+            return [line.strip() for line in f]
```
pmeier, line 121: Since you "resolved" my last comment about inlining this, what is your reason to keep this as a method?

pmeier (approved):
Small nits and one question inline. Otherwise LGTM. Thanks a lot @sallysyw!

torchvision/datasets/fgvc_aircraft.py
sallysyw: whoops... i forgot, let me inline it.

test/test_datasets.py
sallysyw: Actually, it has to happen after the `create_image_folder ` part because otherwise it would complain `root_folder / "data"` is not an existing path.

torchvision/datasets/fgvc_aircraft.py
```diff
@@ -0,0 +1,115 @@
...
+        self.class_to_idx = dict(zip(self.classes, range(len(self.classes))))
+
+        image_data_folder = os.path.join(self._data_path, "data", "images")
+        labels_path = os.path.join(self._data_path, "data", f"images_{self._annotation_level}_{self._split}.txt")
+
+        self._image_files = []
+        self._labels = []
+
+        with open(labels_path, "r") as labels_file:
+            lines = [line.strip() for line in labels_file]
```
pmeier, line 85: You don't need this line anymore, since we strip in L87,
```suggestion
```

torchvision/datasets/fgvc_aircraft.py
```diff
@@ -0,0 +1,115 @@
+from __future__ import annotations
+
+import os
+from typing import Any, Callable, List, Optional, Tuple
```
pmeier, line 4: This was unused.

```suggestion
from typing import Any, Callable, Optional, Tuple
```

pmeier (approved):

PR #5182, opened 2022-01-09
Fix torchvision.utils.make_grid's pad_value variable type hint: int -> float

ProGamerGov:
The `pad_value` variable for `torchvision.utils.make_grid` was incorrectly type hinted as an `int` instead of `float`.


prabhat00155 (approved):
Thanks @ProGamerGov for fixing this.

PR #5186, opened 2022-01-10
relax split requirement for prototype datasets

### LINKED ISSUE #5185: Relax `split="train"` requirement for prototype datasets
Given that `split` is by far the most common option on the datasets, we opted to enforce it for every prototype dataset:

https://github.com/pytorch/vision/blob/ffd0d2375d02cb55e746414aa8db2b4252cd6c42/torchvision/prototype/datasets/utils/_dataset.py#L63-L72

This lead to the separation of `split` and `**options`

https://github.com/pytorch/vision/blob/ffd0d2375d02cb55e746414aa8db2b4252cd6c42/torchvision/prototype/datasets/_api.py#L65-L66

Datasets that do not define any split, still work fine with the default, although it is somewhat questionable to instantiate them with `split="train"`:

https://github.com/pytorch/vision/blob/ffd0d2375d02cb55e746414aa8db2b4252cd6c42/torchvision/prototype/datasets/_builtin/caltech.py#L149-L153

The problem arises if we have a dataset that does define splits, but `"train"` isn't one of them:

https://github.com/pytorch/vision/blob/ffd0d2375d02cb55e746414aa8db2b4252cd6c42/torchvision/prototype/datasets/_builtin/oxford_iiit_pet.py#L29-L37

Now we would have to add a `"train"` option to satisfy our previous rule. Given that our test suite currently instantiates a test for every available config, these datasets are currently no testable.

I suggest we merge `split` into `options` and simply enforce the rule to actually call the split option `split` through code review. 

cc @pmeier @bjuncek

pmeier:
Closes #5185.

cc @pmeier @bjuncek


torchvision/prototype/datasets/_builtin/cifar.py
```diff
@@ -45,14 +45,32 @@ def __iter__(self) -> Iterator[Tuple[np.ndarray, int]]:
...
+            valid_options=dict(split=("train", "test")),
+        )
+
+    def resources(self, config: DatasetConfig) -> List[OnlineResource]:
+        return [
+            HttpResource(
+                f"https://www.cs.toronto.edu/~kriz/{self._FILE_NAME}",
+                sha256=self._SHA256,
+            )
+        ]
```
NicolasHug, line 72: Do I understand correctly that the changes to `resources()` are mostly unrelated to the rest of this PR?
It's fine leaving them here, no need to revert, I just want to make sure I'm not missing something important

torchvision/prototype/datasets/_builtin/cifar.py
```diff
@@ -45,14 +45,32 @@ def __iter__(self) -> Iterator[Tuple[np.ndarray, int]]:
...
-    def _is_data_file(self, data: Tuple[str, io.IOBase], *, config: DatasetConfig) -> Optional[int]:
+    def _is_data_file(self, data: Tuple[str, io.IOBase], *, split: str) -> Optional[int]:
         pass
 
+    def _make_info(self) -> DatasetInfo:
+        return DatasetInfo(
+            type(self).__name__.lower(),
+            type=DatasetType.RAW,
+            homepage="https://www.cs.toronto.edu/~kriz/cifar.html",
+            valid_options=dict(split=("train", "test")),
```
NicolasHug, line 63: Prior to these changes, I believe that `Cifar10`  did not specify the `valid_option`. Is this a fix?

NicolasHug (approved):
Thanks @pmeier , LGTM.

I just have minor questions. It makes sense to me not to make `split` a mandatory parameter, and to enforce our convention through consistent reviews in this case

test/test_prototype_datasets_api.py
```diff
@@ -208,7 +193,7 @@ def test_default_config(self):
         ("config", "kwarg"),
         [
             pytest.param(*(datasets.utils.DatasetConfig(split="test"),) * 2, id="specific"),
-            pytest.param(make_minimal_dataset_info().default_config, None, id="default"),
+            pytest.param(datasets.utils.DatasetConfig(split="train"), None, id="default"),
```
NicolasHug, line 196: just wondering if the `id="default"` is still accurate?

pmeier: I wondered the same thing re-reading. Let me check.

pmeier: The test implicitly relied on the setup of the mock dataset. This is fixed now.

PR #5187, opened 2022-01-10
Expand tests for prototype datasets

pmeier:
This PR adds tests and fixes error of all missing prototype datasets.

Blocked by #5186.

cc @pmeier @bjuncek


test/test_prototype_builtin_datasets.py
```diff
@@ -6,10 +6,20 @@
...
-from torchvision.prototype import transforms
+from torchvision.prototype import transforms, datasets
 from torchvision.prototype.utils._internal import sequence_to_str
 
 
+def test_coverage():
+    untested_datasets = set(datasets.list()) - set(dataset_mock.name for dataset_mock in DATASET_MOCKS)
+    if untested_datasets:
+        raise AssertionError(
+            f"The dataset(s) {sequence_to_str(sorted(untested_datasets), separate_last='and ')} "
```
NicolasHug, line 17: Not related to this PR, but perhaps we should add tests for `sequence_to_str`, so as to ensure that our error message are properly formatted ?

Also nit but if we're going to have such a helper, we could handle the case `a, and b` -> `a and b` (i.e. remove the comma if len == 2

test/builtin_dataset_mocks.py
```diff
@@ -129,10 +133,31 @@ def load(self, config, *, decoder=DEFAULT_DECODER):
         return datapipe, mock_info
 
 
+def config_id(name, config):
```
NicolasHug, line 136: Would the default pytest id generation be enough here?

pmeier: Unfortunately not. You would get something like

```
test/test_prototype_builtin_datasets.py::TestCommon::test_smoke[dataset_mock0-config0]
test/test_prototype_builtin_datasets.py::TestCommon::test_smoke[dataset_mock1-config1]
test/test_prototype_builtin_datasets.py::TestCommon::test_smoke[dataset_mock2-config2]
...
```

The problem is that by default, `pytest` uses the name of the parameter as well as a running index as id. So if we want to have somewhat expressive test names, we need to set the id ourselves. The same example as above but with the custom id from this PR looks like

```
test/test_prototype_builtin_datasets.py::TestCommon::test_smoke[mnist-train]
test/test_prototype_builtin_datasets.py::TestCommon::test_smoke[mnist-test]
test/test_prototype_builtin_datasets.py::TestCommon::test_smoke[fashionmnist-train]
...
```

pmeier (commented):
@ejguan The [test failure](https://app.circleci.com/pipelines/github/pytorch/vision/13999/workflows/92f5cb2e-818d-4046-b239-35e47bd95cba/jobs/1126896/tests) most likely stems from this:

https://github.com/pytorch/vision/blob/f923aeb2cda27e9f87b2011f0202f7159c494623/torchvision/prototype/datasets/_builtin/cub200.py#L171-L178

I need this mapping twice and so I thought I can simply create it once. IIUC, the problem comes from the fact that this starts iteration of the `Demultiplexer` while creating the datapipe and thus it has some file handles in buffer when `traverse` tries to pickle it.

What is recommended pattern here?

pmeier:
@NicolasHug I added the ability add marks like `pytest.mark.xfail` to datasets through the parametrization decorator. With this in place, we can xfail `"cub200"` on the failing tests until https://github.com/pytorch/vision/pull/5187#issuecomment-1015479165 is resolved on the `torchdata` side.

PR #5188, opened 2022-01-11
fix category file generation

pmeier:
#5133 changed some semantics of how the category files are generated, but this was not included in the recent dataset additions.

cc @pmeier @bjuncek


NicolasHug (approved):
Thanks @pmeier !

Issue #3965, opened 2021-06-04
Remove need of include `Python.h` in `vision.cpp`


fmassa:
## 🚀 Feature

Related to https://github.com/pytorch/vision/issues/2349

We currently need to `#include <Python.h>` in https://github.com/pytorch/vision/blob/96ad7f0c80538e06323ab2a4d2d4fb513022f59a/torchvision/csrc/vision.cpp#L4
as well as we need to define `PyInit__C` for Windows https://github.com/pytorch/vision/blob/96ad7f0c80538e06323ab2a4d2d4fb513022f59a/torchvision/csrc/vision.cpp#L15-L22

This is ideally not needed. Windows builds currently require it due to an artifact on how we build our C++ extensions I believe (with setuptools).

We should try to remove this. This might involve using CMake to compiling the extensions (instead of setuptools), so might involve significant rework.

Further discussion can be found in https://github.com/pytorch/vision/pull/3087#discussion_r539154713

For reference, this [is how TF extensions does it](https://github.com/tensorflow/addons/blob/3bae3813f86a748d8d3833377368c98e5289da5d/tensorflow_addons/utils/resource_loader.py#L52) for loading extensions, vs our custom code in https://github.com/pytorch/vision/blob/96ad7f0c80538e06323ab2a4d2d4fb513022f59a/torchvision/extension.py#L8-L47 (and for image and video as well).

Given that our extensions are now completely independent from Python, we just need to know the location where they live and [directly call into `torch.ops.load_library`](https://github.com/pytorch/vision/blob/96ad7f0c80538e06323ab2a4d2d4fb513022f59a/torchvision/extension.py#L47).

Note that we will also need to make the solution compatible with fbcode

fmassa:
@prabhat00155 this might be related to the task I've assigned you. At least the part of the extension loading (hopefully we will not need to change the build, but it's not 100% sure)

PR #5190, opened 2022-01-11
Conditionally include Python in the C++ builds.

### LINKED ISSUE #2692: Using torchvision C++ API without installing python
## ❓ Questions and Help
I have followed the document and built the libtorchvsion.dylib but when I ran `otool -L  libtorchvsion.dylib` it showed
```
libtorchvision.dylib:
	@rpath/libtorchvision.dylib (compatibility version 0.0.0, current version 0.0.0)
	@rpath/libc10.dylib (compatibility version 0.0.0, current version 0.0.0)
	@rpath/libpython3.8.dylib (compatibility version 3.8.0, current version 3.8.0)
	@rpath/libtorch.dylib (compatibility version 0.0.0, current version 0.0.0)
	@rpath/libtorch_cpu.dylib (compatibility version 0.0.0, current version 0.0.0)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 902.1.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1281.100.1)
```
Is there a way to get rid of `libpython3.8.dylib`, I think it is required for using python custom operators and we will only use C++.

dfalbel:
@fmassa Thanks for your suggestions and links in https://github.com/pytorch/vision/issues/2692#issuecoment-1009783544

This PR adds an option `USE_PYTHON` that allows users to decide to link to Python3 or not.
We made it `OFF` by default as suggest in https://github.com/pytorch/vision/issues/2692#issuecoment-1009783544.
~~We might still need to add the definitions for `USE_PYTHON` so it gets linked in the Windows build mentioned in https://github.com/pytorch/vision/issues/3965#issue-911500276~~ (that should be done now)

This should fix #2692

Let me know if this makes sense. Thanks again!


CMakeLists.txt
```diff
@@ -4,6 +4,7 @@ set(CMAKE_CXX_STANDARD 14)
 file(STRINGS version.txt TORCHVISION_VERSION)
 
 option(WITH_CUDA "Enable CUDA support" OFF)
+option(USE_PYTHON "Link to Python when building" OFF)
```
datumbox, line 7: Might be worth adding a reference of that on the README similar to what we do for `WITH_CUDA` to explain when and under which conditions this is necessary to be turned on.

datumbox (commented):
Overall looks good. Thanks @dfalbel, I've added only one comment for putting some documentation around the flag.

Before merging this, is worth checking on FBcode side to see if it breaks anything or if changes will be needed on BUCK. @prabhat00155 I know you already did the syncing this week, but would you have the bandwidth to cherrypick and test this PR internally to confirm everything works as expected?

CMakeLists.txt
dfalbel: Thanks! Just added a paragraph in https://github.com/pytorch/vision/pull/5190/commits/a2594f58c4e9d39b71d3d9eba2be31a06053ccab

dfalbel:
Thanks @datumbox ! If something fails I think we can still decide to make `USE_PYTHON=on` by default so as to keep the current behavior while allowing one to disable Python.

CMakeLists.txt
prabhat00155, line 7: Should we keep the default value of `USE_PYTHON` ON to maintain the current behaviour? Thoughts?

dfalbel: @fmassa suggested in https://github.com/pytorch/vision/issues/2692#issuecomment-1009783544 that the default behavior could be changed. I think it's fine to go safer and keep the old behavior if that's preferable.

prabhat00155: I see. I'll run the internal tests on this PR and if that goes without any errors, we can merge this.

prabhat00155: Internal tests: D33582385

prabhat00155 (approved):
Thanks @dfalbel!

fmassa (approved):
Changes LGTM, thanks!

For a future PR (which will need FB access), it would be good to remove the `MOBILE` var, as it can now be used through `USE_PYTHON`. It will need changes to fbcode and xplat I believe.

prabhat00155:
> Hey @prabhat00155!
> 
> You merged this PR, but no labels were added. The list of valid labels is available at https://github.com/pytorch/vision/blob/main/.github/process_commit.py

@NicolasHug Should we add "build" to [SECONDARY LABELS](https://github.com/pytorch/vision/blob/5e56575e688a85a3bc9dc3c97934dd864b65ce47/.github/process_commit.py#L27)?

NicolasHug:
sure, thanks for flagging

Issue #5141, opened 2021-12-30
Output decoded frames in RGB format


prabhat00155:
Currently, GPU decoder returns frames in NV12 format. We need to support RGB to be consistent with CPU decoder.

Issue #5145, opened 2021-12-30
Return frames with appropriate shape


prabhat00155:
Currently, GPU video decoder returns 1D frames, this needs to be modified to return 3D frames.

PR #5191, opened 2022-01-11
Return RGB frames as output of GPU decoder

prabhat00155:
GPU decoder currently outputs NV12 flattened frames. This PR changes changes the output type to RGB, also fixing the output shape.
Most of this work was done by @fmassa in his draft [PR](https://github.com/pytorch/vision/pull/5149), I made some adjustments when calculating pyav output.
Resolves https://github.com/pytorch/vision/issues/5141 and https://github.com/pytorch/vision/issues/5145.


torchvision/csrc/io/decoder/gpu/decoder.cpp
```diff
@@ -138,38 +139,25 @@ int Decoder::handle_picture_display(CUVIDPARSERDISPINFO* disp_info) {
   }
 
   auto options = torch::TensorOptions().dtype(torch::kU8).device(torch::kCUDA);
-  torch::Tensor decoded_frame = torch::empty({get_frame_size()}, options);
+  torch::Tensor decoded_frame =
+      torch::empty({get_height(), get_width(), 3}, options);
```
fmassa, line 143: Are you sure we need the rounding by multiples of 2 from https://github.com/pytorch/vision/blob/5e56575e688a85a3bc9dc3c97934dd864b65ce47/torchvision/csrc/io/decoder/gpu/decoder.h#L46 here?

This will give a wrong output shape for the videos which have odd size, so this looks wrong to me. In my prototype I used [width directly](https://github.com/pytorch/vision/pull/5149/files#diff-482fc8a8c571ceb755bd7f447b3803637e06d646a954397f6cf8b0614e2bbe42R143)

torchvision/csrc/io/decoder/gpu/decoder.cpp
```diff
@@ -138,38 +139,25 @@ int Decoder::handle_picture_display(CUVIDPARSERDISPINFO* disp_info) {
...
   uint8_t* frame_ptr = decoded_frame.data_ptr<uint8_t>();
+  const uint8_t* const source_arr[] = {
+      (const uint8_t* const)source_frame,
+      (const uint8_t* const)(source_frame + source_pitch * ((surface_height + 1) & ~1))};
```
fmassa, line 147: I would like us to double-check this `(surface_height + 1) & ~1)` condition. Can you create some videos with odd dimensions to validate what is actually needed?

test/test_video_gpu_decoder.py
```diff
@@ -31,10 +31,10 @@ def test_frame_reading(self):
             decoder = VideoReader(full_path, device="cuda:0")
             with av.open(full_path) as container:
                 for av_frame in container.decode(container.streams.video[0]):
-                    av_frames = torch.tensor(av_frame.to_ndarray().flatten())
+                    av_frames = torch.tensor(av_frame.to_rgb(src_colorspace="ITU709").to_ndarray())
                     vision_frames = next(decoder)["data"]
-                    mean_delta = torch.mean(torch.abs(av_frames.float() - decoder._reformat(vision_frames).float()))
-                    assert mean_delta < 0.1
+                    mean_delta = torch.mean(torch.abs(av_frames.float() - vision_frames.cpu().float()))
+                    assert mean_delta < 0.75
```
fmassa, line 37: I generally dislike performing the comparisons as the mean over the whole tensor because it can completely allow for having a whole section of the image to be all black / white / random (which is btw an existing issue with the current CPU video reader cc @bjuncek )

I would add a TODO here to try to use a different function for comparing equality here.

fmassa (commented):
Thanks for the PR Prabhat!

I made some comments, let me know what you think

fmassa:
Also, I think a bunch of things related to `output_format` like in https://github.com/pytorch/vision/blob/5e56575e688a85a3bc9dc3c97934dd864b65ce47/torchvision/csrc/io/decoder/gpu/decoder.h#L44-L45 can be cleaned now, right? As we will enforce / assume that the decoder returns NV12, and then we convert it to RGB.

torchvision/csrc/io/decoder/gpu/decoder.cpp
prabhat00155: I think you are right, this should be width as the output format is not NV12.

test/test_video_gpu_decoder.py
prabhat00155: Do you have a recommendation of what to use instead of `torch.mean()`? I see mean() being used at other places as well, for example: https://github.com/pytorch/vision/blob/5e56575e688a85a3bc9dc3c97934dd864b65ce47/test/test_image.py#L366
I know numpy has [assert_array_almost_equal](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_array_almost_equal.html), which I have used in the past to compare arrays. I guess `torch.allclose()` is an option.

prabhat00155:
> Also, I think a bunch of things related to `output_format` like in
> 
> https://github.com/pytorch/vision/blob/5e56575e688a85a3bc9dc3c97934dd864b65ce47/torchvision/csrc/io/decoder/gpu/decoder.h#L44-L45
> 
> can be cleaned now, right? As we will enforce / assume that the decoder returns NV12, and then we convert it to RGB.

Yeah, I'll clean these up in a separate PR. There are a few functions in there that may not be needed anymore.

torchvision/csrc/io/decoder/gpu/decoder.cpp
prabhat00155: [surface_height](https://github.com/pytorch/vision/blob/bf073e785528970e6a1605e411e4fc382d686dc7/torchvision/csrc/io/decoder/gpu/decoder.cpp#L323) is different from [luma_height](https://github.com/pytorch/vision/blob/bf073e785528970e6a1605e411e4fc382d686dc7/torchvision/csrc/io/decoder/gpu/decoder.cpp#L316) which is directly related to the video dimensions.
I can revisit this when doing code refactoring.

fmassa: My point is the `+1) & ~1` condition. I'm not sure if it's actually necessary

test/test_video_gpu_decoder.py
fmassa: So, my preference would be to use `(a - b).abs().max() < threshold`. But looks like this doesn't work already in the CPU video decoder because of the black bands that appear sometimes (cc @bjuncek ), but ideally this is the type of things I would consider.

But this can be done in the future, just leave a TODO

fmassa (approved):
I'm approving to unblock, but I think we need to add tests for odd-sized videos to validate that our implementation handles things correctly

torchvision/csrc/io/decoder/gpu/decoder.cpp
prabhat00155: luma height is aligned by 2, so the chroma offset should not be odd(chroma base address can't be odd memory location), hence the alignment.

prabhat00155:
> I'm approving to unblock, but I think we need to add tests for odd-sized videos to validate that our implementation handles things correctly

I tested it locally using odd-sized videos. We'll could perhaps add some odd-sized videos in torchvision for testing.

test/test_video_gpu_decoder.py
prabhat00155: https://github.com/pytorch/vision/issues/5216

Issue #4947, opened 2021-11-16
Flaky video test


prabhat00155:
https://app.circleci.com/pipelines/github/pytorch/vision/12610/workflows/9ab703cc-7277-4f05-9079-d2ace1dc94b4/jobs/998680

https://app.circleci.com/pipelines/github/pytorch/vision/13889/workflows/62bf350f-c0a9-4506-a09f-9dbc56b7cf1b/jobs/1117960

```
Traceback (most recent call last):
  File "/root/project/test/test_videoapi.py", line 75, in test_frame_reading
    assert mean_delta.item() < 2.5
AssertionError: assert 2.513227701187134 < 2.5
 +  where 2.513227701187134 = <built-in method item of Tensor object at 0x7f3941851ea8>()
 +    where <built-in method item of Tensor object at 0x7f3941851ea8> = tensor(2.5132).item
```

cc @pmeier @seemethere

prabhat00155:
@bjuncek Do you have time to take a look at this?

PR #5192, opened 2022-01-11
flakey videoAPI test (#4947)

bjuncek:
Finding the cause of the flaky test as cannot repro locally with either of current settings:
python 3.7/8/9
ffmpeg 4.2.2

- [x] repro on ci?
_(everything seems to work - I'll re trigger test few more times just to be sure)_
- [x] remove IO bottleneck by creating a resource pool
- [x] profit?


test/test_videoapi.py
```diff
@@ -57,34 +57,66 @@ def test_frame_reading(self):
...
+                    av_frames.append(torch.tensor(av_frame.to_rgb().to_ndarray()).permute(2, 0, 1))
+                    av_pts.append(av_frame.pts * av_frame.time_base)
+                av_reader.close()
 
-                    av_array = torch.tensor(av_frame.to_rgb().to_ndarray()).permute(2, 0, 1)
-                    vr_array = vr_frame["data"]
-                    mean_delta = torch.mean(torch.abs(av_array.float() - vr_array.float()))
+                # get vr frames
+                video_reader = VideoReader(full_path, "video")
+                for vr_frame in video_reader:
```
prabhat00155, line 73: Is there any advantage of have two separate for loops? We could just use one like here: https://github.com/pytorch/vision/blob/e3767f8aa51921460c89e1d396ccce097cebc9d5/test/test_video_gpu_decoder.py#L33 and do something like line 84 on a frame by frame basis and not have to store a list of frames.

test/test_videoapi.py
```diff
@@ -57,34 +57,66 @@ def test_frame_reading(self):
...
+                # get av frames
+                for av_frame in av_reader.decode(av_reader.streams.audio[0]):
+                    av_frames.append(torch.tensor(av_frame.to_ndarray()).permute(1, 0))
+                    av_pts.append(av_frame.pts * av_frame.time_base)
+                av_reader.close()
 
-                    max_delta = torch.max(torch.abs(av_array.float() - vr_array.float()))
+                # get vr frames
+                video_reader = VideoReader(full_path, "audio")
+                for vr_frame in video_reader:
```
prabhat00155, line 108: Same as above.

test/test_videoapi.py
```diff
@@ -57,34 +57,66 @@ def test_frame_reading(self):
...
+                av_frames, vr_frames = [], []
+                av_pts, vr_pts = [], []
+                # get av frames
                 for av_frame in av_reader.decode(av_reader.streams.video[0]):
-                    vr_frame = next(video_reader)
-
-                    assert float(av_frame.pts * av_frame.time_base) == approx(vr_frame["pts"], abs=0.1)
+                    av_frames.append(torch.tensor(av_frame.to_rgb().to_ndarray()).permute(2, 0, 1))
+                    av_pts.append(av_frame.pts * av_frame.time_base)
+                av_reader.close()
```
prabhat00155, line 69: Minor comment: If you use `with`, you won't need to worry about this.

prabhat00155 (commented):
Thanks @bjuncek! I have a couple of comments. Let me know what you think.

test/test_videoapi.py
bjuncek: fair, addressed

test/test_videoapi.py
bjuncek: Theoretically, having separate loops, and closing the file pointers between the readers should increase the stability of the tests as there is no instance in which there are multiple pointers to the same file.\TBH figured it's one of those things that can't hurt, but might increase stability

prabhat00155: Do you suspect that being the reason for the tests being flaky?

bjuncek: Not 100% sure, but I was told it was a good practice, especially when dealing with C++ memcpys.

PR #5195, opened 2022-01-14
Change order of sections in toctree

NicolasHug:
This PR changes the order of our main sections in our docs (transforms, models, datasets...). The new proposed order follows the measured popularity of each page as measured by Google analytics.


prabhat00155 (approved):
Thanks @NicolasHug! Just out of curiosity, do we plan on changing this every time there is change in popularity order?

NicolasHug:
Thanks for the review @prabhat00155 .

I wouldn't think so, as I doubt the relative page usage would drastically change anytime soon. What prompted me to update the order is that there is a significant difference between the dataset usage vs models and transforms (I can't share numbers publicly but you have them in our workplace chat :) ), so I thought it would make sense *not* to have the datasets as the first section.

For a few % difference I wouldn't bother though.

docs/source/index.rst
```diff
@@ -31,13 +31,13 @@ architectures, and common image transformations for computer vision.
...
-   datasets
    transforms
    models
-   feature_extraction
+   datasets
+   utils
    ops
    io
-   utils
+   feature_extraction
```
oke-aditya, line 40: Not clubbing with models? Potentially this is very new feature (and hence low popularity), but the docs are best read after reading models.

oke-aditya:
The order is very surprising. Surprised by the fact that transforms and models are used more than datasets and utils more than io or ops!
Btw out of curiosity how did you know the popularity %s ? Is this through the torch logging or something internal fb?

docs/source/index.rst
NicolasHug: We'll likely merge it into the models page once we get to the more general revamp of the models page

NicolasHug:
@oke-aditya I updated my comment above: we measure the popularity of each page with Google Analytics.

This doesn't necessarily reflect which classes/functions are the most used in practice in users code (this is **not** something we track), but it tells us exactly which pages are more popular than others, and so it gives us some insight on what doc readers are looking for. In this case, the datasets page is significantly less popular than the transforms and models pages.

Issue #3331, opened 2021-01-31
[proposal] Use self.flatten instead of torch.flatten and when becomes possible derive ResNet from nn.Sequential (scripting+quantization is blocker), would simplify model surgery in the most frequent cases


vadimkantorov:
Currently In https://github.com/pytorch/vision/blob/master/torchvision/models/resnet.py#L243: 
```python
        x = self.avgpool(x)
        x = torch.flatten(x, 1)
        x = self.fc(x)
```

If it instead used `x = self.flatten(x)`, then it would simplify model surgery: `del model.avgpool, model.flatten, model.fc`. Also in this case the class can just derive from Sequential and use OrderedDict to pass submodules (like in https://discuss.pytorch.org/t/ux-mix-of-nn-sequential-and-nn-moduledict/104724/2?u=vadimkantorov), this would preserve checkpoint compat as well. The method `forward` could then be removed

datumbox:
Thanks for the proposal. Replacing torch.flatten with nn.Flatten seems it won't break anything but could you describe a bit more your use-case and how it will help you achieve what you want?

Concerning doing a major refactoring and inheriting from Sequential, this might have side-effects on all the other models that depend on resnet (segmentation, object detection etc), so I'm not sure if it can be done in a backward-compatible manner. Note also that replacing the forward is not something we can do due to how Quantization works:
https://github.com/pytorch/vision/blob/6116812d51c6e5524ef48f8b14c0293ed674ed55/torchvision/models/quantization/resnet.py#L93-L100

@fmassa Let me know if you see any problems about replacing flatten.

fmassa:
I would be fine replacing `torch.flatten` with `nn.Flatten`, although it would only simplify model surgery if we were to make further modifications to the ResNet model as you suggested (making it inherit from `nn.Sequential`). Without this, only `del model.fc` etc wouldn't be enough.

So from this perspective, there is limited value in replacing `torch.flatten` with `nn.Flatten`, it being only syntactic sugar from one another.

vadimkantorov:
Oh, I didn't realize:
```
# Ensure scriptability 
# super(QuantizableResNet,self).forward(x) 
# is not scriptable 
```

If this becomes fixed, I guess inheriting from nn.Sequential should be fine.

fmassa:
[...]

PR #5197, opened 2022-01-14
Adding ConvNeXt architecture in prototype

datumbox:
Experimental implementation of ConvNeXt from https://arxiv.org/pdf/2201.03545.pdf

We managed to fully reproduce the paper using the original recipe described on the paper (`Acc@1 82.064 Acc@5 95.858`). Nevertheless, using [TorchVision's new recipe](https://pytorch.org/blog/how-to-train-state-of-the-art-models-using-torchvision-latest-primitives/) along with a switch from SGD=>AdamW (`--opt adamw --lr 1e-3 --weight-decay 0.05`) we are able to improve the accuracy by 0.45 points:
```
PYTHONPATH=$PYTHONPATH:`pwd` python -u run_with_submitit.py --ngpus 8 --nodes 2 \ 
--model convnext_tiny --batch-size 64 --opt adamw --lr 1e-3 --lr-scheduler cosineannealinglr \ 
--lr-warmup-epochs 5 --lr-warmup-method linear --auto-augment ta_wide --epochs 600 --random-erase 0.1 \ 
--label-smoothing 0.1 --mixup-alpha 0.2 --cutmix-alpha 1.0 --weight-decay 0.05 --norm-weight-decay 0.0 \
--train-crop-size 176 --model-ema --val-resize-size 236 --ra-sampler --ra-reps 4
```

Model Validation:
```
torchrun --nproc_per_node=1 train.py --test-only --weights ConvNeXt_Tiny_Weights.ImageNet1K_V1 --model convnext_tiny
Acc@1 82.520 Acc@5 96.146
```

cc @datumbox @vfdev-5 @bjuncek


torchvision/models/convnext.py
```diff
@@ -0,0 +1,201 @@
...
+        super().__init__()
+        self.block = nn.Sequential(
+            ConvNormActivation(
+                dim,
+                dim,
+                kernel_size=7,
+                groups=dim,
+                norm_layer=norm_layer,
+                activation_layer=None,
+                bias=True,  # TODO: check
```
datumbox, line 48: Can you talk about your decision using Bias on the first convolution of the ConvNeXt block? Have you experimented and found it necessary?

torchvision/models/convnext.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+class LayerNorm2d(nn.LayerNorm):
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        self.channels_last = kwargs.pop("channels_last", False)
+        super().__init__(*args, **kwargs)
+
+    def forward(self, x):
+        if not self.channels_last:
+            x = x.permute(0, 2, 3, 1)
```
datumbox, line 30: At TorchVision, we try to reuse as much as possible standard PyTorch components. I understand that in your original implementation, you provide a custom implementation for [channels first](https://github.com/facebookresearch/ConvNeXt/blob/main/models/convnext.py#L138-L143). Could you talk about the performance degradation you've experienced and how much this was that lead you to reimplementing it?

datumbox (commented):
@HannaMao @s9xie @liuzhuang13 Thanks for your work on [ConvNeXt](https://arxiv.org/pdf/2201.03545.pdf). 

Last week we've added an experimental TorchVision-compatible implementation of your architecture based on your [code](https://github.com/facebookresearch/ConvNeXt). This is done to facilitate experiments on how [TorchVision's new recipe](https://pytorch.org/blog/how-to-train-state-of-the-art-models-using-torchvision-latest-primitives/) works on new architectures.

Below I've left a couple of comments for you, could you please check and let me know your thoughts?

Also we managed to reproduce the reported accuracies using TorchVision's components for the tiny version (the other variants require more time to complete). While experimenting with the recipe, I noticed that it's very difficult to train the architecture with standard SGD. AdamW certainly does the trick, but I just wanted to know if that's something you observed as well doing your research.

torchvision/models/convnext.py
s9xie: Thanks for noting this. We have not tested (I'm going to launch a run with T model soon) but it should be totally unnecessary due to the LayerNorm after this.

torchvision/models/convnext.py
s9xie: The thing is we don't have a standard PyTorch component for channels_first LN. Here are some discussions with Ross Wightman (https://twitter.com/wightmanr/status/1481383509142818817?s=20). In Ross's timm implementation, he has a presumably better LN implementation for channel_first. See here: https://github.com/rwightman/pytorch-image-models/blob/b669f4a5881d17fe3875656ec40138c1ef50c4c9/timm/models/convnext.py#L109

s9xie:
> Also we managed to reproduce the reported accuracies using TorchVision's components for the tiny version (the other variants require more time to complete). While experimenting with the recipe, I noticed that it's very difficult to train the architecture with standard SGD. AdamW certainly does the trick, but I just wanted to know if that's something you observed as well doing your research.

Regarding SGD vs AdamW, this is something we are particularly interested and plan to explore ourselves. Thanks for the data point, we also find ConvNeXt behaves very similarly to ViT/Swin Transformer. There are a couple of hypothesis - 1) the hyper-parameter choices needs to be revisited with SGD. 2) Based on the ConvStem paper (https://arxiv.org/abs/2106.14881), one source of instability w.r.t SGD might be the initial patchify stem. We plan to try ConvStem with SGD and hopefully this will address the optimization problem (and the patchify layer did not bring a substantial gain in accuracy anyways).

torchvision/models/convnext.py
datumbox: Thanks for confirming. Indeed in 99% of cases, the bias prior the Normalization layer is unnecessary. We've seen previously one [niche case in optical flow](https://github.com/pytorch/vision/blob/922db3086e654871c35cd80c2c01eabb65d78475/torchvision/models/optical_flow/raft.py#L35-L40) were a combination of transfer learning and freezing the BN made sense to have biases, so I just wanted to confirm there isn't anything like that here. Let's compare notes after you run the model, to confirm.

torchvision/models/convnext.py
datumbox: Thanks for the references. I wasn't aware of the concurrent discussions. I'll have a look to measure on our side. Ideally I would like to reuse the existing kernels as much as possible, unless there is a big gap in performance to justify a custom implementation.

torchvision/models/convnext.py
s9xie: The 99% of cases you have seen are BN, where the biases will become spurious parameters since they will be canceled out anyways. ConvNeXt uses LN, which means the Conv2D biases here will have an effect. I still believe it won't matter much for the final results but it is not equivalent. Only showing results with Tiny model is also not enough, we need to run all the model sizes to make sure the conclusion holds true when scaling up. My suggestion is to keep it True here.

datumbox:
> Regarding SGD vs AdamW, this is something we are particularly interested and plan to explore ourselves.

@s9xie Sounds great, please do let me know about any of your findings. In the meantime, I changed optimizer on our recipe and we got the following accuracies for the tiny variant: `Acc@1 82.508 Acc@5 96.196`

The above is achieved by doing the minimum possible modifications on our recipe to change SGD with AdamW, so here are the overwrites only related to the optimizer change: `--opt adamw --lr 1e-3 --weight-decay 0.05`. The values are adopted straight from the paper without any grid-search, so it's very likely that we can do better.

xiaohu2015:
@datumbox Good! Do you plan to add another CNN: https://github.com/facebookresearch/deit/blob/main/patchconvnet_models.py

datumbox:
@xiaohu2015 It's not in the immediate plans but you can add a proposal at #2707 and we can discuss it. :)

torchvision/models/convnext.py
liuzhuang13: Hi @datumbox,
In this post I try to explain the motivation for our specific choices, in case it's still helpful: https://github.com/facebookresearch/ConvNeXt/issues/18#issuecomment-1014390947 
Thanks for your work on incorporating ConvNeXt!

datumbox: @liuzhuang13 Thanks for the reply. Is it fair to say that the approach that we follow here is expected to be 0-5% slower than the optimum? I haven't had the chance to run benchmarks but that's what I understand from your note.

FYI, the issue is that TorchVision has a common API across all models to accept the `norm_layer` as a parameter which is going to be tricky to support if I switch the 1x1 convs to linear. Where or not we will do this, depends on the speed impact.

liuzhuang13: Yes, I just tried switching to conv layers in residual blocks, and using permutation + LN in all cases. I found combined together they cause 20-30% slowdown in inference @ 224 resolution for ConvNeXt-T/S. However, in ConvNeXt-B at 224 or any model at 384 resolution, it seems as fast as our original impl. This is on V100s, I cannot say much on other platforms though. It is indeed a bit strange to me.

datumbox: Thanks a lot for looking into it @liuzhuang13. 20-30% slow sounds very large and we would probably want to make it faster. I'll run benchmarks later and share with you any findings.

vadimkantorov:
It may be useful to derive ConvNexT from nn.Sequential (https://github.com/pytorch/vision/issues/3331), it should be very simple by just calling `super().__init__(collections.OrderedDict(features = features, avgpool = avgpool, classifier = classifier))` and then probably the custom forward is not needed.

torchvision/models/convnext.py
```diff
@@ -0,0 +1,201 @@
...
+
+        lastblock = block_setting[-1]
+        lastconv_output_channels = (
+            lastblock.out_channels if lastblock.out_channels is not None else lastblock.input_channels
+        )
+        self.classifier = nn.Sequential(
+            norm_layer(lastconv_output_channels), nn.Flatten(1), nn.Linear(lastconv_output_channels, num_classes)
+        )
+
+        for m in self.modules():
```
vadimkantorov, line 163: Also, if further along some standardization of reset_parameters is envisioned (https://github.com/pytorch/vision/issues/3410), at least moving initialization to a separate function could be useful

datumbox: Thanks for the proposal. Unfortunately the approach we use for initializing the layers does not play nice with FX as the information of classes are lost and we can't use `isinstance()`. So while this check works in the constructor, after the model gets traced it won't be possible to be reset.

This is something raised previously at https://github.com/pytorch/pytorch/issues/66335 if you want to read more info. Until this use-case gets better support, we can't use both FX and making public a `reset_parameters` method.

datumbox: As for the proposal of using Sequential, I'm also a very big fan of implementing the models as Sequential (This is why the `features` of this implementation uses it). Unfortunately, there are some limitations imposed by quantization and JIT that require defining explicitly a forward and even using a `_forward_impl()` (you can see an example of the idiom [here](https://github.com/pytorch/vision/blob/main/torchvision/models/quantization/mobilenetv3.py#L98-L102)). So many of these not-so-elegant approaches are done to circumvent known limitations but I'm hopeful that we will be able to get around them soon.

vadimkantorov: My 

> making public a `reset_parameters` method.

Well, even if it's a non-public randomly named `_reset_parameters`, `_init_parameters` etc, this will make further uniformization easier. And maybe if jitted model calls to it are not supported yet, having it named as in nn.Linear helps readability etc.

vadimkantorov: > Unfortunately the approach we use for initializing the layers does not play nice with FX as the information of classes are lost and we can't use `isinstance()`

The fact of FX blocking subclassing in such ways is a pretty serious degradation...

vadimkantorov: > As for the proposal of using Sequential, I'm also a very big fan of implementing the models as Sequential (This is why the `features` of this implementation uses it). Unfortunately, there are some limitations imposed by quantization and JIT that require defining explicitly a forward and even using a `_forward_impl()` (you can see an example of the idiom [here](https://github.com/pytorch/vision/blob/main/torchvision/models/quantization/mobilenetv3.py#L98-L102)). So many of these not-so-elegant approaches are done to circumvent known limitations but I'm hopeful that we will be able to get around them soon.

In that old issue, I proposed of just inserting quant/dequant stubs in the sequential itself, then there should be no problem with forward_impl, right?

datumbox: Unfortunately not everything can be implemented neatly as a Sequential ([example](https://github.com/pytorch/vision/blob/main/torchvision/models/quantization/mobilenetv3.py#L28-L29)). Though I do like to simplify the code, I wouldn't want to introduce things if they are not fully supported because it can bite downstream users. 

For example, introducing a `_reset_parameters` (which we've done previously in [RegNets](https://github.com/pytorch/vision/blob/main/torchvision/models/regnet.py#L370) before we learn about the FX limitation and we should remove) will give the ability to users to call it whenever they want, even if it's private. But then if they combine this with the FX-base feature extractor, it won't work. Keeping it in the constructor where it's inaccessible makes it a responsibility of the end user to write a method to re-init the model in cases where this is allowed.

This discussion is important and I'm glad you raised it. Let's continue it on the respective tickets you opened so that that this doesn't get buried in this unrelated PR.

vadimkantorov: > I wouldn't want to introduce things if they are not fully supported because it can bite downstream users.

Well, currently FX feature extractor isn't fully supported wrt hooks (which is an old core feature), it hasn't blocked its introduction and publicizing. FX is a nice feature, but it's not seamless, e.g. it blocks any existing module-based transformations, as it erases moduels.

> will give the ability to users to call it whenever they want, even if it's private. B

This also has never been a problem in core pytorch, plenty of private underscore-prefixed methods exist in all sort of classes. And them being private is sufficient for not worrying people would complain of them problems. Anyway, `reset_parameters`-kind of methods already exist in almost all models.

vadimkantorov: > Unfortunately not everything can be implemented neatly as a Sequential ([example](https://github.com/pytorch/vision/blob/main/torchvision/models/quantization/mobilenetv3.py#L28-L29)). Though I do like to simplify the code, I wouldn't want to introduce things if they are not fully supported because it can bite downstream users.

Of course, not everything is sequential (I would say, this is a straw argument), but backbones more often than not are and can be quantized just with inserting stubs without more complex forward_impl idiom, right? unless I'm missing some other problems of quantization

torchvision/ops/misc.py
```diff
@@ -131,7 +131,7 @@ def __init__(
         norm_layer: Optional[Callable[..., torch.nn.Module]] = torch.nn.BatchNorm2d,
         activation_layer: Optional[Callable[..., torch.nn.Module]] = torch.nn.ReLU,
         dilation: int = 1,
-        inplace: bool = True,
+        inplace: Optional[bool] = True,
```
datumbox, line 134: Not all activation layers support inplace. So this allows us to support those that dont.

torchvision/prototype/models/convnext.py
```diff
@@ -0,0 +1,225 @@
...
+__all__ = ["ConvNeXt", "ConvNeXt_Tiny_Weights", "convnext_tiny"]
+
+
+class LayerNorm2d(nn.LayerNorm):
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        self.channels_last = kwargs.pop("channels_last", False)
+        super().__init__(*args, **kwargs)
+
+    def forward(self, x):
+        #  TODO: Benchmark this against the approach described at https://github.com/pytorch/vision/pull/5197#discussion_r786251298
```
datumbox, line 27: Benchmarking necessary and potential rewrite to move out of prototype.

torchvision/prototype/models/convnext.py
```diff
@@ -0,0 +1,225 @@
...
+        input_channels: int,
+        out_channels: Optional[int],
+        num_layers: int,
+    ) -> None:
+        self.input_channels = input_channels
+        self.out_channels = out_channels
+        self.num_layers = num_layers
+
+    def __repr__(self) -> str:
+        s = self.__class__.__name__ + "("
```
jdsgomes, line 81: missing f-string indication `f`

nit: if you want to remove multiple assignments you can write something like
```python
s = (
    self.__class__.__name__ +
    f"(input_channels={input_channels}, out_channels={out_channels}, num_layers={num_layers})"
)
```

or if you rename input_channels to in_channels you can get everything in one line

jdsgomes (commented):
LGTM, just made some minor comments

torchvision/prototype/models/convnext.py
datumbox: This is a quite common pattern in TorchVision that I'm repeating here. See [this](https://github.com/pytorch/vision/blob/eac3dc7bab436725b0ba65e556d3a6ffd43c24e1/torchvision/ops/deform_conv.py#L192). We could change in all instances perhaps on a separate issue?

Also good call for the `input_channels` vs `in_channels`. Here I maintain it for consistency with other models such a shufflenets, mobilenetv3, efficientnets, vit etc.

jdsgomes (approved):

PR #5203, opened 2022-01-17
Add support for PCAM dataset

NicolasHug:
Towards https://github.com/pytorch/vision/issues/5108

This PR adds support for the PCAM dataset.

cc @pmeier


torchvision/datasets/oxford_iiit_pet.py
```diff
@@ -27,8 +27,8 @@ class OxfordIIITPet(VisionDataset):
         transform (callable, optional): A function/transform that  takes in a PIL image and returns a transformed
             version. E.g, ``transforms.RandomCrop``.
         target_transform (callable, optional): A function/transform that takes in the target and transforms it.
-        download (bool, optional): If True, downloads the dataset from the internet and puts it into ``root/dtd``. If
-            dataset is already downloaded, it is not downloaded again.
+        download (bool, optional): If True, downloads the dataset from the internet and puts it into
+            ``root/oxford-iiit-pet``. If dataset is already downloaded, it is not downloaded again.
```
pmeier, line 31: Thanks!

torchvision/datasets/pcam.py
```diff
@@ -0,0 +1,136 @@
...
+from PIL import Image
+
+from .utils import download_file_from_google_drive, extract_archive, verify_str_arg
+from .vision import VisionDataset
+
+
+class PCAM(VisionDataset):
+    """`PCAM Dataset   <https://github.com/basveeling/pcam>`_.
+
+    The PatchCamelyon dataset is a binray classification dataset with 327,680
```
pmeier, line 15:
```suggestion
    The PatchCamelyon dataset is a binary classification dataset with 327,680
```

torchvision/datasets/pcam.py
```diff
@@ -0,0 +1,136 @@
...
+    color images (96 x 96px), extracted from histopathologic scans of lymph node
```
pmeier, line 16: Not sure if there is a canonical way to write image sizes

```suggestion
    color images (96px x 96px), extracted from histopathologic scans of lymph node
```

torchvision/datasets/pcam.py
```diff
@@ -0,0 +1,136 @@
...
+    sections. Each image is annoted with a binary label indicating presence of
```
pmeier, line 17:
```suggestion
    sections. Each image is annotated with a binary label indicating presence of
```

torchvision/datasets/pcam.py
```diff
@@ -0,0 +1,136 @@
...
+    metastatic tissue.
+
+    This dataset requires the ``h5py`` package which you can install with ``pip install h5py``.
+
+     Args:
+         root (string): Root directory of the dataset.
+         split (string, optional): The dataset split, supports ``"trai"`` (default), ``"test"`` or ``"val"``.
```
pmeier, line 24:
```suggestion
         split (string, optional): The dataset split, supports ``"train"`` (default), ``"test"`` or ``"val"``.
```

Plus, is there a reason to order `"val"` after `"test"`?

torchvision/datasets/pcam.py
```diff
@@ -0,0 +1,136 @@
...
+
+        if not self._check_exists():
+            raise RuntimeError("Dataset not found. You can use download=True to download it")
+
+        self.classes = ["neg", "pos"]
+        self.classes_to_idx = {"neg": 0, "pos": 1}
+
+    def __len__(self) -> int:
+        images_file = self._FILES[self._split]["images"][0]
+        with self.h5py.File(self._base_folder / images_file) as images_data:
```
NicolasHug, line 104: Note for here and below: opening a `File` does not load its data into memory, so the operation is very cheap and fast.

Similarly below accessing a single row in the file will not load the entire file, just a specific section of it.

I guess we could open the files and keep the handles in `__init__`, but I'm not sure it would be any faster, and we might not be able to ever close the handles properly.

torchvision/datasets/pcam.py
```diff
@@ -0,0 +1,136 @@
...
+        self._base_folder = pathlib.Path(self.root) / "pcam"
+
+        if download:
+            self._download()
+
+        if not self._check_exists():
+            raise RuntimeError("Dataset not found. You can use download=True to download it")
+
+        self.classes = ["neg", "pos"]
+        self.classes_to_idx = {"neg": 0, "pos": 1}
```
pmeier, line 100: Do we really need this for binary classification datasets? I feel like 0 / 1 is kinda self-explanatory here.

torchvision/datasets/pcam.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def __len__(self) -> int:
+        images_file = self._FILES[self._split]["images"][0]
+        with self.h5py.File(self._base_folder / images_file) as images_data:
+            print(images_data.keys())
```
pmeier, line 105:
```suggestion
```

torchvision/datasets/pcam.py
```diff
@@ -0,0 +1,136 @@
...
+             dataset is already downloaded, it is not downloaded again.
+    """
+
+    _FILES = {
+        "train": {
+            "images": (
+                "camelyonpatch_level_2_split_train_x.h5",  # Data file name
+                "1Ka0XfEMiwgCYPdTI-vv6eUElOBnKFKQ2",  # Google Drive ID
+                "1571f514728f59376b705fc836ff4b63",  # md5 hash
+            ),
```
NicolasHug, line 38: I'm not ecstatic about this big dict, but I needed everything in the same place to support a per-split download logic (i.e. only download the test data if we don't need train nor val).

torchvision/datasets/pcam.py
```diff
@@ -0,0 +1,136 @@
...
+    def __len__(self) -> int:
+        images_file = self._FILES[self._split]["images"][0]
+        with self.h5py.File(self._base_folder / images_file) as images_data:
+            print(images_data.keys())
+            return images_data["x"].shape[0]
+
+    def __getitem__(self, idx: int) -> Tuple[Any, Any]:
+        images_file = self._FILES[self._split]["images"][0]
+        with self.h5py.File(self._base_folder / images_file) as images_data:
+            image = Image.fromarray(images_data["x"][idx]).convert("RGB")
```
pmeier, line 111: Is this a cheap operation or should we somehow cache stuff to not open the file over and over again?

torchvision/datasets/pcam.py
```diff
@@ -0,0 +1,136 @@
...
+            image = self.transform(image)
+        if self.target_transform:
+            target = self.target_transform(target)
+
+        return image, target
+
+    def _check_exists(self) -> bool:
+        images_file = self._FILES[self._split]["images"][0]
+        targets_file = self._FILES[self._split]["targets"][0]
+        return all((os.path.exists(str(self._base_folder / h5_file)) for h5_file in (images_file, targets_file)))
```
pmeier, line 127:
```suggestion
        return all(self._base_folder.joinpath(h5_file).exists() for h5_file in (images_file, targets_file)))
```

torchvision/datasets/pcam.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def _download(self) -> None:
+        if self._check_exists():
+            return
+
+        for file_name, file_id, md5 in self._FILES[self._split].values():
+            archive_name = file_name + ".gz"
+            download_file_from_google_drive(file_id, str(self._base_folder), filename=archive_name, md5=md5)
+            extract_archive(str(self._base_folder / archive_name))
```
pmeier, line 136: It is not really an archive, but rather a file that happens to be compressed with gzip, correct? Internally, `extract_archive` defers to `_decompress`

https://github.com/pytorch/vision/blob/5e56575e688a85a3bc9dc3c97934dd864b65ce47/torchvision/datasets/utils.py#L398-L404

Maybe we can use it directly?

torchvision/datasets/pcam.py
pmeier: See https://github.com/pytorch/vision/pull/5203#discussion_r786049239

PR #5207, opened 2022-01-18
Remove split parameter from datasets.load()

NicolasHug:
This PR removes the `split` parameter from `datasets.load()` and also removes it from the call to `make_config()`. This is possibly something we forgot to do in https://github.com/pytorch/vision/pull/5186 ?

Without this, running e.g. `datasets.load("oxford-iiit-pet")` would result in the following error:

```
ValueError: Invalid argument 'train' for option 'split' of dataset oxford-iiit-pet. Did you mean 'trainval'?
```

cc @pmeier @bjuncek

### DIFF
--- a/test/builtin_dataset_mocks.py
+++ b/test/builtin_dataset_mocks.py
@@ -1212,7 +1212,7 @@ class CUB2002011MockData(_CUB200MockData):
                 size=[1, *make_tensor((2,), low=3, dtype=torch.int).tolist()],
             )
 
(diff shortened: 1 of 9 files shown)


pmeier (approved):
You are right, this was an oversight. I took the liberty to add a test for that. It wouldn't have triggered for `"oxford-iiit-pet"` since we have no tests for it yet, but both `"caltech101"` and `"caltech256"` would have failed.

pmeier:
After more investigation, it is probably better to always use the public API to load the datasets. I'll revert my changes and send a separate PR for that.

Issue #5204, opened 2022-01-17
There is a Difference between Torchvsion Shear and PIL Shear


SamuelGabriel:
### 🐛 Describe the bug

As @agaldran pointed out in the TrivialAugment repo (https://github.com/automl/trivialaugment/issues/6), there seems to be difference between the behavior of a shear implemented with the PIL and the TorchVision affine transform.

This is an issue for the autoaugment algorithms, as this might yield to different results compared to other implementations of these algorithms. It might be an issue for other applications as well. Both the AutoAugment (https://github.com/tensorflow/models/blob/fd34f711f319d8c6fe85110d9df6e1784cc5a6ca/research/autoaugment) and the TrivialAugment (https://github.com/automl/trivialaugment) reference implementations use PIL, while RandAugment has no reference implementation.

```
from PIL import Image
import math
import torchvision # >= 0.11

from torchvision.transforms import functional as F
interpolation = torchvision.transforms.InterpolationMode.NEAREST
fill = None

img = Image.new('RGB', (32,32), (255,255,0))
magnitude = .7

# shear_x as seen in torchvision https://github.com/pytorch/vision/blob/b5aa0915fe16e82ee4c24919032b4e7afae3ae1b/torchvision/transforms/autoaugment.py#L17
im_torch = F.affine(img, angle=0.0, translate=[0, 0], scale=1.0, shear=[math.degrees(magnitude), 0.0],
                    interpolation=interpolation, fill=fill)

# shear_x as seen in https://github.com/automl/trivialaugment/blob/3bfd06552336244b23b357b2c973859500328fbb/aug_lib.py#L156 and https://github.com/tensorflow/models/blob/fd34f711f319d8c6fe85110d9df6e1784cc5a6ca/research/autoaugment/augmentation_transforms.py#L290
im_pil = img.transform(img.size, Image.AFFINE, (1, magnitude, 0, 0, 1, 0))

im_torch.show()
im_pil.show()
```

This script will yield the following images:

![download](https://user-images.githubusercontent.com/9828297/149813813-b1dfbebd-7310-49d2-b2e1-438d2e6ad6e8.png)
![download](https://user-images.githubusercontent.com/9828297/149813838-f719c77d-917c-4fb1-9e98-72359ab345d0.png)

Which should be the same.

It looks like torchvision has a fixed center, while PIL uses a fixed top. I did not dig into the code much, yet, though. Is there someone here who implemented `affine` maybe and can give a reasoning for the different shearing. 

[...]

PR #5208, opened 2022-01-18
Added center arg to F.affine and RandomAffine ops

vfdev-5:
Related to #5204

Description:
- Added `center` arg to F.affine and RandomAffine ops
- tests

Example from https://github.com/pytorch/vision/issues/5204#issue-1106104198
```python
from PIL import Image
import math
import torchvision # >= 0.11

from torchvision.transforms import functional as F
interpolation = torchvision.transforms.InterpolationMode.NEAREST
fill = None

img = Image.new('RGB', (32,32), (255,255,0))
magnitude = .7

im_torch = F.affine(
    img, angle=0.0, translate=[0, 0], scale=1.0, shear=[math.degrees(magnitude), 0.0],
    interpolation=interpolation, fill=fill, center=[0, 0])
```

image


torchvision/transforms/functional.py
```diff
@@ -945,26 +945,42 @@ def adjust_gamma(img: Tensor, gamma: float, gain: float = 1) -> Tensor:
...
+    #         [ 0       ,  0       , 1 ]
+    #
     # where R is a rotation matrix, S is a scaling matrix, and SHx and SHy are the shears:
     # SHx(s) = [1, -tan(s)] and SHy(s) = [1      , 0]
     #          [0, 1      ]              [-tan(s), 1]
-    #
-    # Thus, the inverse is M^-1 = C * RSS^-1 * C^-1 * T^-1
+
+    # TODO: implement the option
+    assert centered_shear
```
datumbox, line 983: I would recommend to remove this and open an issue where we provide the details of why we should implement this option, why it's necessary and references to papers or other implementations that follow this approach.

datumbox (commented):
@vfdev-5 Thanks for the PR. The changes look good, I only added a comment for an unimplemented feature left in.

I think it would be useful also to document a bit the approach of this PR by adding references to specific papers or implementations that follow your approach for centering.

datumbox (approved):
LGTM, thanks @vfdev-5.

PR #5210, opened 2022-01-19
Adding vit_h_14 architecture

sallysyw:
See whether the unittest can pass after changes in #5176 

cc @datumbox


kazhang (approved):
LGTM, thanks @sallysyw !

datumbox:
This PR [breaks our CI](https://app.circleci.com/pipelines/github/pytorch/vision/14224/workflows/8be00cf1-926d-45e4-949d-6bb3f3d10e71/jobs/1144749) and causes memory issue not only for ViT but also for other models. Given that we don't offer weights for it, we should revert and investigate ways to add it back without breaking the CI.

PR #5212, opened 2022-01-19
Use public API for loading in prototype datasets tests

pmeier:
Follow-up to https://github.com/pytorch/vision/pull/5207#issuecomment-1016270132

cc @pmeier @bjuncek


test/builtin_dataset_mocks.py
```diff
@@ -61,77 +63,71 @@ def info(self):
     def name(self):
         return self.info.name
 
-    def _parse_mock_data(self, mock_data_fn):
-        def wrapper(info, root, config):
-            mock_infos = mock_data_fn(info, root, config)
+    def _parse_mock_data(self, config, mock_infos):
```
pmeier, line 66: This is just GitHub not picking up on the identation change. Before this was a decorator, but I changed it into a regular method. The actual body has not changed.

test/builtin_dataset_mocks.py
```diff
@@ -1000,7 +996,7 @@ def dtd(info, root, _):
 def fer2013(info, root, config):
     num_samples = 5 if config.split == "train" else 3
 
-    path = root / f"{config.split}.txt"
+    path = root / f"{config.split}.csv"
```
pmeier, line 999: This change and everything below in this file are actual bugs in our mock data generation that were hidden by our custom loading logic.

test/datasets_utils.py
```diff
@@ -868,9 +868,9 @@ def _split_files_or_dirs(root, *files_or_dirs):
 def _make_archive(root, name, *files_or_dirs, opener, adder, remove=True):
     archive = pathlib.Path(root) / name
     if not files_or_dirs:
-        dir = archive.with_suffix("")
-        if dir.exists() and dir.is_dir():
-            files_or_dirs = (dir,)
+        file_or_dir = archive.with_name(archive.name.rsplit("".join(archive.suffixes), 1)[0])
```
pmeier, line 871: This didn't work before for archives like `foo.tar.gz` since `.with_suffix` only replaces the last suffix and thus turns it into `foo.tar`.

test/builtin_dataset_mocks.py
NicolasHug: Do you know why the tests were passing before despite the files had the wrong name?

test/datasets_utils.py
NicolasHug: It's hard to figure out what this line does. Could we perhaps add a comment, or find a simpler solution?

NicolasHug (approved):

test/builtin_dataset_mocks.py
pmeier: Before the resources where not collected by the regular logic, but rather "hand-fed" to `_make_datapipe`. So as long as the data in `mock_resources[0]` corresponded to `Dataset.resources(...)[0]` the test suite didn't notice.

test/datasets_utils.py
pmeier: Yeah, I really hate that `pathlib.Path` has no `with_suffixes` method :( Basically, this gives me the "name" of the file without any suffixes. Let me see if I can find a more readable solution.

test/datasets_utils.py
```diff
@@ -868,9 +868,12 @@ def _split_files_or_dirs(root, *files_or_dirs):
 def _make_archive(root, name, *files_or_dirs, opener, adder, remove=True):
     archive = pathlib.Path(root) / name
     if not files_or_dirs:
-        dir = archive.with_suffix("")
-        if dir.exists() and dir.is_dir():
-            files_or_dirs = (dir,)
+        # We cannot use Path.with_suffix("") here, since that only applies to the last suffix if multiple suffixes are
+        # present. For example, `pathlib.Path("foo.tar.gz").with_suffix("")` results in `foo.tar`. Since we want to
+        # extract the name without any suffixes here, we replace them manually.
+        file_or_dir = archive.with_name(archive.name.replace("".join(archive.suffixes), ""))
```
pmeier, line 874: Alternative as discussed offline

```suggestion
        file_or_dir = archive
        for _ in range(len(path.suffixes)):
            file_or_dir = file_or_dir.with_suffix("")
```

I'm fine with either.

test/builtin_dataset_mocks.py
```diff
@@ -61,77 +63,71 @@ def info(self):
...
+
+            self._cache[config_] = mock_info
 
         return self._cache[config]
 
-    def load(self, config, *, decoder=DEFAULT_DECODER):
-        try:
-            self.info.check_dependencies()
-        except ModuleNotFoundError as error:
-            pytest.skip(str(error))
```
pmeier, line 126: This should have not been removed. The tests should be runnable even if no or not all third party dependencies are installed.

PR #5213, opened 2022-01-19
fix and add test for sequence_to_str

### LINKED ISSUE #5202: Add tests for prototype utilities
Perhaps we should add tests for `sequence_to_str`, so as to ensure that our error message are properly formatted ?

Also nit but if we're going to have such a helper, we could handle the case `a, and b` -> `a and b` (i.e. remove the comma if len == 2

_Originally posted by @NicolasHug in https://github.com/pytorch/vision/pull/5187#discussion_r784852697_

cc @bjuncek

pmeier:
Closes #5202.

cc @pmeier @bjuncek

### DIFF
--- a/test/test_prototype_utils.py
+++ b/test/test_prototype_utils.py
@@ -1,0 +1,17 @@
+import pytest
+from torchvision.prototype.utils._internal import sequence_to_str
+
+
+@pytest.mark.parametrize(
+    ("seq", "separate_last", "expected"),
+    [
+        pytest.param([], "", "", id="empty"),
+        pytest.param(["foo"], "", "'foo'", id="single"),
+        pytest.param(["foo", "bar"], "", "'foo', 'bar'", id="double"),
+        pytest.param(["foo", "bar"], "and ", "'foo' and 'bar'", id="double-separate_last"),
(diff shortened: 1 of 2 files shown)


test/test_prototype_utils.py
```diff
@@ -0,0 +1,17 @@
...
+
+@pytest.mark.parametrize(
+    ("seq", "separate_last", "expected"),
+    [
+        pytest.param([], "", "", id="empty"),
+        pytest.param(["foo"], "", "'foo'", id="single"),
+        pytest.param(["foo", "bar"], "", "'foo', 'bar'", id="double"),
+        pytest.param(["foo", "bar"], "and ", "'foo' and 'bar'", id="double-separate_last"),
+        pytest.param(["foo", "bar", "baz"], "", "'foo', 'bar', 'baz'", id="multi"),
+        pytest.param(["foo", "bar", "baz"], "and ", "'foo', 'bar', and 'baz'", id="multi-separate_last"),
```
NicolasHug, line 13: Unless we have a strong reason for keeping them I would prefer avoiding the use of `pytest.param`. They add a lot of extra characters, and make the parametrization harder to read.

For such tests that are super fast to run, do we really need the specify the `id`s?

NicolasHug (approved):
Thanks @pmeier

PR #5214, opened 2022-01-19
Add GTSRB dataset to prototypes

NicolasHug:
This PR adds the prototype version of the GTSRB dataset

cc @pmeier @bjuncek


torchvision/prototype/datasets/_builtin/gtsrb.py
```diff
@@ -0,0 +1,122 @@
...
+            "gtsrb",
+            type=DatasetType.IMAGE,
+            homepage="https://benchmark.ini.rub.de",
+            valid_options=dict(split=("train", "test")),
+        )
+
+    _URLS = {
+        "train": "https://sid.erda.dk/public/archives/daaeac0d7ce1152aea9b61d9f1e19370/GTSRB-Training_fixed.zip",
+        "test": "https://sid.erda.dk/public/archives/daaeac0d7ce1152aea9b61d9f1e19370/GTSRB_Final_Test_Images.zip",
+        "test_gt": "https://sid.erda.dk/public/archives/daaeac0d7ce1152aea9b61d9f1e19370/GTSRB_Final_Test_GT.zip",
```
pmeier, line 38: Can we maybe split off the URL root here?

torchvision/prototype/datasets/_builtin/gtsrb.py
```diff
@@ -0,0 +1,122 @@
...
+        self,
+        resource_dps: List[IterDataPipe],
+        *,
+        config: DatasetConfig,
+        decoder: Optional[Callable[[io.IOBase], torch.Tensor]],
+    ) -> IterDataPipe[Dict[str, Any]]:
+
+        if config["split"] == "train":
+            images_dp = resource_dps[0]
+            images_dp = Filter(images_dp, self._filter_images)
```
pmeier, line 92: This is equivalent to

```suggestion
            images_dp = Filter(images_dp, path_comparator("suffix", ".ppm"))
```

torchvision/prototype/datasets/_builtin/gtsrb.py
```diff
@@ -0,0 +1,122 @@
...
+                images_dp,
+                gt_dp,
+                key_fn=path_accessor("name"),
+                ref_key_fn=getitem("Filename"),
+                buffer_size=INFINITE_BUFFER_SIZE,
+                merge_fn=self._append_label_test,
+            )  # path, handle, label
+
+        dp = hint_sharding(dp)
+        dp = hint_shuffling(dp)
```
pmeier, line 110: Given that we have all samples in `images_dp`, can we do this before the `IterKeyZipper`?

torchvision/prototype/datasets/_builtin/gtsrb.py
```diff
@@ -0,0 +1,122 @@
...
+        decoder: Optional[Callable[[io.IOBase], torch.Tensor]],
+    ) -> IterDataPipe[Dict[str, Any]]:
+
+        if config["split"] == "train":
+            images_dp = resource_dps[0]
+            images_dp = Filter(images_dp, self._filter_images)
+            dp = Mapper(images_dp, self._append_label_train)  # path, handle, label
+        else:
+            images_dp, gt_dp = resource_dps
+            dp = Filter(images_dp, self._filter_images)
```
pmeier, line 96: It seems this is the same for both branches. This is another reason to oust `images_dp` outside of the `if`.

torchvision/prototype/datasets/_builtin/gtsrb.py
```diff
@@ -0,0 +1,122 @@
...
+            )
+
+        return rsrcs
+
+    def _filter_images(self, data: Tuple[str, Any]) -> bool:
+        return pathlib.Path(data[0]).suffix == ".ppm"
+
+    def _append_label_train(self, path_and_handle: Tuple[str, Any]) -> Tuple[str, Any, int]:
+        path, handle = path_and_handle
+        label = int(pathlib.Path(path).parent.stem)
```
pmeier, line 64: Given that `pathlib.Path(path).parent` is a directory, `.name` and `.stem` should be the same, correct? If yes, I would prefer `.name`.

torchvision/prototype/datasets/_builtin/gtsrb.py
```diff
@@ -0,0 +1,122 @@
...
+        dp = Mapper(dp, partial(self._collate, decoder=decoder))
+        return dp
+
+    def _generate_categories(self, root: pathlib.Path) -> List[str]:
+        config = self.default_config
+
+        images_dp = self.resources(config)[0].load(root)
+        images_dp = Filter(images_dp, self._filter_images)
+
+        labels = sorted(set(pathlib.Path(path).parent.stem for path, _ in images_dp))
```
pmeier, line 121: Same as above.

torchvision/prototype/datasets/_builtin/gtsrb.py
```diff
@@ -0,0 +1,122 @@
...
+            )  # path, handle, label
+
+        dp = hint_sharding(dp)
+        dp = hint_shuffling(dp)
+
+        dp = Mapper(dp, partial(self._collate, decoder=decoder))
+        return dp
+
+    def _generate_categories(self, root: pathlib.Path) -> List[str]:
+        config = self.default_config
```
pmeier, line 116: The categories are only available in the `"train"` split, correct? Although I don't believe we ever make `split="test"` the default, we can be more specific here

```suggestion
        config = self.info.make_config(split="train")
```

torchvision/prototype/datasets/_builtin/gtsrb.categories
```diff
@@ -0,0 +1,43 @@
+00000
```
pmeier, line 1: Are there no human-readable categories available? If not, you can replace this file (as well as the method that generates it), by passing `categories=[f"{label:05d}" for label in range(43)]` to the `DatasetInfo`.

test/builtin_dataset_mocks.py
```diff
@@ -1021,6 +1021,57 @@ def fer2013(info, root, config):
...
+                    f"{image_idx:05d}.ppm",
+                    torch.randint(1, 100, size=()).item(),
+                    torch.randint(1, 100, size=()).item(),
+                    torch.randint(1, 100, size=()).item(),
+                    torch.randint(1, 100, size=()).item(),
+                    torch.randint(1, 100, size=()).item(),
+                    torch.randint(1, 100, size=()).item(),
+                    torch.randint(1, len(classes) + 1, size=()).item(),
+                ]
+                csv_file.write(";".join(map(str, row)) + "\n")
```
pmeier, line 1069: Nit: Maybe `csv.DictWriter` could increase readability here?

test/builtin_dataset_mocks.py
```diff
@@ -1021,6 +1021,57 @@ def fer2013(info, root, config):
...
+            csv_file.write("Filename;Width;Height;Roi.X1;Roi.Y1;Roi.X2;Roi.Y2;ClassId\n")
+            for image_idx in range(num_examples):
+                row = [
+                    f"{image_idx:05d}.ppm",
+                    torch.randint(1, 100, size=()).item(),
+                    torch.randint(1, 100, size=()).item(),
+                    torch.randint(1, 100, size=()).item(),
+                    torch.randint(1, 100, size=()).item(),
+                    torch.randint(1, 100, size=()).item(),
+                    torch.randint(1, 100, size=()).item(),
```
pmeier, line 1066: Nit2: If you don't go with the suggestion above, maybe merge these together?

```suggestion
                    *torch.randint(1, 100, size=(6,)).tolist(),
```

pmeier (commented):
Thanks a lot @NicolasHug! I've only reviewed the prototype part, let me know if I also need to look at the other changes.

test/datasets_utils.py
```diff
@@ -877,7 +877,7 @@ def _make_archive(root, name, *files_or_dirs, opener, adder, remove=True):
     files, dirs = _split_files_or_dirs(root, *files_or_dirs)
 
     with opener(archive) as fh:
-        for file in files:
+        for file in sorted(files):
```
NicolasHug, line 880: @pmeier LMK what you think of this.

Down below I set `buffer_size=1` in the Zipper, because in the original .zip files, both the `image_dp` and the `gt_dp` are fully aligned: they both contain images 00001, 00002, etc. in this order. So I'm assuming that `buffer_size=1` is better than `buffer_size=UNLIMITED`?

Without this call to `sorted()`, the tests would fail: the .zip archive created by `make_zip` would contain the files in a shuffled order (because `files` is a set), and so `image_dp` and the `gt_dp` would not be aligned anymore, leading to a failure to match keys in the Zipper. (Note: this is only a problem in the tests; the code works fine otherwise on my custom script iterating over the dataset).

I hope this won't make other tests fails. This might not be a problem that we have right now, but perhaps something to keep in mind for the future: we might need the test archives to *exactly* match the order of the "original" archives.

torchvision/prototype/datasets/_builtin/gtsrb.py
```diff
@@ -0,0 +1,113 @@
...
+
+            fieldnames = ["Filename", "Width", "Height", "Roi.X1", "Roi.Y1", "Roi.X2", "Roi.Y2", "ClassId"]
+            gt_dp = CSVDictParser(gt_dp, fieldnames=fieldnames, delimiter=";")
+
+            dp = IterKeyZipper(
+                images_dp,
+                gt_dp,
+                key_fn=path_accessor("name"),
+                ref_key_fn=getitem("Filename"),
+                buffer_size=1,
```
NicolasHug, line 105: Is there a simpler alternative to `IterKeyZipper` here?

In our case we know that image_dp and gt_dp are fully aligned, and all we need is a pure `zip`, we don't need any key matching.

torchvision/prototype/datasets/_builtin/gtsrb.py
NicolasHug: I just set `buffer_size=1` above (with a rabbit hole of questions). If we shuffle before the `IterKeyZipper`, would this force us to increase the `buffer_size`?

torchvision/prototype/datasets/_builtin/gtsrb.py
pmeier: There is [`Zipper`](https://github.com/pytorch/pytorch/blob/214f4bf2ff4a7a880d335bcb795582d9d7935ab5/torch/utils/data/datapipes/iter/combining.py#L365) that behaves just like `zip`

test/datasets_utils.py
pmeier: Good point, I didn't think of that. Do you know if the returned paths of `pathlib.Path.glob()` is stable? If yes, we could simply replace the sets in `_split_files_or_dirs` with lists.

torchvision/prototype/datasets/_builtin/gtsrb.py
pmeier: Just to make sure we are not going into dangerous waters here: @ejguan `images_dp` comes from `ZipArchiveLoader`. Is the returned order guaranteed to be stable?

ejguan: I believe so. ref: https://docs.python.org/3/library/zipfile.html#zipfile.ZipFile.infolist

test/datasets_utils.py
NicolasHug: I'm not sure about `Path.glob()`. I know that `glob.glob` has no guaranteed order, but I don't think `Path.glob()` relies on it. Maybe the safest is to not assume a specific order.

BTW, slightly related, what was the reason to use sets instead of lists?

pmeier: > what was the reason to use sets instead of lists?

It think the reason was to avoid duplicates, but I don't remember if there was a case where I hit something like that.

torchvision/prototype/datasets/_builtin/gtsrb.py
pmeier: Yes. Not just increase, but rather make "infinite".

NicolasHug: Since we're now using Zipper instead of IterKeyZipper, it looks like we're unable to move these now?

test/datasets_utils.py
pmeier: Have you tried using lists rather than sorting afterwards? If CI is not complaining for other datasets, I feel like that would be the better approach.

torchvision/prototype/datasets/_builtin/README.md
```diff
@@ -72,35 +107,81 @@ def sha256sum(path, chunk_size=1024 * 1024):
...
-There are two special datapipes that are not used through their class, but through the functions `hint_sharding` and `hint_shuffling`. As the name implies they only hint part in the datapipe graph where sharding and shuffling should take place, but are no-ops by default. They can be imported from `torchvision.prototype.datasets.utils._internal` and are required in each dataset.
+Get the skeleton of your dataset class ready with all 3 methods. For
+`_make_datapipe()`, you can just do `return resources_dp[0]` to get started.
+Then import the dataset class in
+`torchvision/prototype/datasets/_builtin/__init__.py`: this will automatically
+register the dataset and it will be instantiable via
+`datasets.load("mydataset")`. On a separate script, try something like
 
-Finally, each item in the final datapipe should be a dictionary with `str` keys. There is no standardization of the names (yet!).
+```py
```
pmeier, line 166: Oh, does this work as intended? My current understanding is that you need 

```suggestion
```python
```

torchvision/prototype/datasets/_builtin/README.md
```diff
@@ -72,35 +107,81 @@ def sha256sum(path, chunk_size=1024 * 1024):
...
+d = datasets.load("mydataset")
+for e in d:
+    print(e)  # this is the content of the datapipe returned by _make_datapipe()
```
pmeier, line 169: 1. Especially for an example, let's be more explicit here
2. I usually add a `break` to avoid loading the whole datapipe every time and terminating the execution manually. No strong opinion on that.

```suggestion
from torchvision.prototype import datasets

dataset = datasets.load("mydataset")
for sample in dataset:
    # this is the content of an item in datapipe returned by _make_datapipe()
    print(sample)
    break
```

torchvision/prototype/datasets/_builtin/README.md
```diff
@@ -72,35 +107,81 @@ def sha256sum(path, chunk_size=1024 * 1024):
...
+# Or you can also inspect e in a debugger
+```
 
-## FAQ
+This will give you an idea of what the first datapipe in `resources_dp`
+contains. You can also do that with `resources_dp[1]` or `resources_dp[2]`
+(etc.) if they exist. Then follow the instructions above to manipulate these
+datapipes and return the appropriate dict format.
```
pmeier, line 176:
```suggestion
datapipes and return the appropriate dictionary format.
```

torchvision/prototype/datasets/_builtin/gtsrb.py
```diff
@@ -0,0 +1,105 @@
...
+        *,
+        config: DatasetConfig,
+        decoder: Optional[Callable[[io.IOBase], torch.Tensor]],
+    ) -> IterDataPipe[Dict[str, Any]]:
+
+        images_dp = resource_dps[0]
+        images_dp = Filter(images_dp, path_comparator("suffix", ".ppm"))
+
+        if config["split"] == "train":
+            dp = Mapper(images_dp, self._append_label_train)  # path, handle, label
```
pmeier, line 91: I don't think we need that comment. The same information is available in `self._append_label_train` as well as in `self._collate` without anything in between where it would be useful to have it.

```suggestion
            dp = Mapper(images_dp, self._append_label_train)
```

torchvision/prototype/datasets/_builtin/gtsrb.py
```diff
@@ -0,0 +1,105 @@
...
+        else:
+            gt_dp = resource_dps[1]
+
+            fieldnames = ["Filename", "Width", "Height", "Roi.X1", "Roi.Y1", "Roi.X2", "Roi.Y2", "ClassId"]
+            gt_dp = CSVDictParser(gt_dp, fieldnames=fieldnames, delimiter=";", skip_lines=1)
+
+            dp = Zipper(images_dp, gt_dp)
+            dp = Mapper(dp, self._append_label_test)  # path, handle, label
```
pmeier, line 99: Same as above.

torchvision/prototype/datasets/_builtin/gtsrb.py
```diff
@@ -0,0 +1,105 @@
...
+        decoder: Optional[Callable[[io.IOBase], torch.Tensor]],
+    ) -> IterDataPipe[Dict[str, Any]]:
+
+        images_dp = resource_dps[0]
+        images_dp = Filter(images_dp, path_comparator("suffix", ".ppm"))
+
+        if config["split"] == "train":
+            dp = Mapper(images_dp, self._append_label_train)  # path, handle, label
+        else:
+            gt_dp = resource_dps[1]
```
pmeier, line 93: It was not instantly obvious what "gt" means. Thus, maybe we can be more explicit here?

```suggestion
            ground_truth_dp = resource_dps[1]
```

or

```suggestion
            label_dp = resource_dps[1]
```

torchvision/prototype/datasets/_builtin/gtsrb.py
```diff
@@ -0,0 +1,105 @@
...
+
+            fieldnames = ["Filename", "Width", "Height", "Roi.X1", "Roi.Y1", "Roi.X2", "Roi.Y2", "ClassId"]
```
pmeier, line 95: 1. It seems that we only use `"FileName"` and `"ClassId"` from these. Thus, there is no need to specify the others. `csv.DictReader` will automatically place them as a list under the `"None"` key.
```suggestion
            field_names = ["Filename", "ClassId"]
```

pmeier (approved):
Only minor stuff inline. Thanks a lot @NicolasHug!

Besides that one other question: Given that the dataset also provides a bounding box for each image, shouldn't we also return it? For the test split we already load the data. For the training data, each image folder contains a `GT-{label:05d}.csv` file in the same format as the testing annotations. Basically instead of `Filter`ing the images, you could use a `Demultiplexer` to get a images and a annotations datapipe.

test/datasets_utils.py
NicolasHug: I just saw this call to `remove()` which might the reason for using sets:

https://github.com/pytorch/vision/blob/afda28accbc79035384952c0359f0e4de8454cb3/test/datasets_utils.py#L862-L863

I can still switch to lists if you'd like, I guess I  would have to write something like

```py
dirs = [dir in dirs if dir != root]
```

LMK which one you prefer

torchvision/prototype/datasets/_builtin/README.md
NicolasHug: I think it works, if we try to edit a .md file interactively from GitHub we can see that it gets formatted properly

image

image

torchvision/prototype/datasets/_builtin/gtsrb.py
NicolasHug: That was my assumption as well, but it doesn't seem to be the case.
Tests fail if I use `field_names = ["Filename", "ClassId"]` with

```
    "label": Label(label, category=self.categories[label]),
IndexError: tuple index out of range
```


and judging by the values, I believe it's becausethe csv reader now considers that the ClassId column is the second one (i.e. actually the Width).

Perhaps I'm not using this correctly? I was surprised myself that I even had to specificy all fieldnames, and that I had to add `skip_lines=1` to skip the header (which could just be read automatically)

torchvision/prototype/datasets/_builtin/gtsrb.py
```diff
@@ -61,21 +62,32 @@ def _append_label_train(self, path_and_handle: Tuple[str, Any]) -> Tuple[str, An
...
+        (image_path, image_buffer), csv_info = data
         label = int(csv_info["ClassId"])
-        return path, handle, label
 
-    def _collate(
-        self, data: Tuple[str, Any, int], decoder: Optional[Callable[[io.IOBase], torch.Tensor]]
-    ) -> Dict[str, Any]:
-        image_path, image_buffer, label = data
+        bbox = BoundingBox(
+            torch.tensor([int(csv_info[k]) for k in ("Roi.X1", "Roi.Y1", "Roi.X2", "Roi.Y2")]), format="xyxy"
```
pmeier, line 72:
```suggestion
            torch.tensor([int(csv_info[k]) for k in ("Roi.X1", "Roi.Y1", "Roi.X2", "Roi.Y2")]), format="xyxy", image_size=(int(csv_info["Height"]), int(csv_info["Width"]))
```

test/test_prototype_builtin_datasets.py
```diff
@@ -143,3 +144,19 @@ def test_extra_label(self, dataset_mock, config):
...
             assert key in sample and isinstance(sample[key], type)
+
+
+@parametrize_dataset_mocks(DATASET_MOCKS["gtsrb"])
+class TestGTSRB:
+    def test_label_matches_path(self, dataset_mock, config):
+        # We read the labels from the csv files instead. But for the trainset, the labels are also part of the path.
+        # This test makes sure that they're both the same
+        if config["split"] != "train":
+            return
```
NicolasHug, line 155: @pmeier that test is only relevant for the training data. is there a way to only call this test with a specific config?

pmeier: Not yet unfortunately. I would say let's keep it they way you have right now and revisit this later when we extend the test suite in general. Still, mini nit: `config` is a bunch

```suggestion
        if config.split != "train":
            return
```

PR #5215, opened 2022-01-19
Add seek in GPU decoder

prabhat00155:
Resolves https://github.com/pytorch/vision/issues/5139.


test/test_video_gpu_decoder.py
```diff
@@ -36,6 +36,33 @@ def test_frame_reading(self):
...
+        decoder = VideoReader(full_path, device="cuda:0")
+        time = duration / 2
+        decoder.seek(time, keyframes_only=keyframes)
+        with av.open(full_path) as container:
+            container.seek(int(time * 1000000), any_frame=not keyframes, backward=False)
+            for av_frame in container.decode(container.streams.video[0]):
+                av_frames = torch.tensor(av_frame.to_rgb(src_colorspace="ITU709").to_ndarray())
+                vision_frames = next(decoder)["data"]
+                mean_delta = torch.mean(torch.abs(av_frames.float() - vision_frames.cpu().float()))
+                assert mean_delta < 0.75
```
jdsgomes, line 64: Why 0.75 ?

jdsgomes (commented):
Just a minor clarification, otherwise looks great to me.

test/test_video_gpu_decoder.py
prabhat00155: There are some differences in pixel values that we get compared to pyav(not entirely sure what's the reason; could be rounding related),  I found that the mean of the differences is under 0.75 for the test videos. See https://github.com/pytorch/vision/blob/44ae1e51341faed0873c0b83096a6e64b72080ba/test/test_video_gpu_decoder.py#L37

jdsgomes (approved):
LGTM, thanks!

bjuncek:
Looks good to me; on the test/pyav side, have we manually confirmed that "seek" does the same thing (i.e. that it doesn't go to keyframes)?

PR #5219, opened 2022-01-19
add lazily filled dict for prototype datasets

pmeier:
This is my proposed solution to https://github.com/pytorch/vision/pull/5187#issuecomment-1015479165. Since we only need the mapping during iteration, we can also delay its instantiation until then. Thoughts?

cc @pmeier @bjuncek

### DIFF
--- a/torchvision/prototype/datasets/_builtin/cub200.py
+++ b/torchvision/prototype/datasets/_builtin/cub200.py
@@ -31,6 +31,7 @@ from torchvision.prototype.datasets.utils._internal import (
     getitem,
     path_comparator,
     path_accessor,
+    LazyDict,
 )
 from torchvision.prototype.features import Label, BoundingBox, Feature
 
@@ -93,6 +94,9 @@ class CUB200(Dataset):
             return 3
         else:
             return None
+
+    def _2011_image_key(self, rel_posix_path: str) -> str:
+        return rel_posix_path.rsplit("/", 1)[1]
 
     def _2011_filter_split(self, row: List[str], *, split: str) -> bool:
         _, split_id = row
@@ -173,9 +177,8 @@ class CUB200(Dataset):
             )
 
             image_files_dp = CSVParser(image_files_dp, dialect="cub200")
-            image_files_map = dict(
-                (image_id, rel_posix_path.rsplit("/", maxsplit=1)[1]) for image_id, rel_posix_path in image_files_dp
-            )
+            image_files_dp = Mapper(image_files_dp, self._2011_image_key, input_col=1)
+            image_files_map = LazyDict(image_files_dp)
 
             split_dp = CSVParser(split_dp, dialect="cub200")
             split_dp = Filter(split_dp, functools.partial(self._2011_filter_split, split=config.split))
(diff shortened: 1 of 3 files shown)


torchvision/prototype/datasets/_builtin/cub200.py
```diff
@@ -173,9 +177,8 @@ def _make_datapipe(
             )
 
             image_files_dp = CSVParser(image_files_dp, dialect="cub200")
-            image_files_map = dict(
-                (image_id, rel_posix_path.rsplit("/", maxsplit=1)[1]) for image_id, rel_posix_path in image_files_dp
-            )
+            image_files_dp = Mapper(image_files_dp, self._2011_image_key, input_col=1)
```
ejguan, line 180: What we can do here is a `in_memory_cache` over `image_files_dp`. Then, we could add an API to convert a `IterDataPipe` to a lazy loaded `MapDataPipe` to represent the `LazyDict`.
If we use `LazyDict` here, I have concern that `image_files_map` would be missing from the DataLoader graph.
cc: @VitalyFedyunin

ejguan: Besides, I think the DataLoader would complain this datapipe graph in the second epoch because `image_files_dp` is never used after the first epoch then `Demux` would also be non-serializable same as the comment I made in the other PR.

So, a fix from `Demux` is not avoidable.

torchvision/prototype/datasets/_builtin/cub200.py
```diff
@@ -173,9 +177,8 @@ def _make_datapipe(
             )
```
ejguan, line 178: A second thought. Could we simply filter `image_files_dp` from `archive_dp` here and create the `image_files_map` dictionary?

Then, we can do demux over `archive_dp` again and drop data in `image_files_dp`.

pmeier: So basically splitting of `image_files_dp` from the graph?

torchvision/prototype/datasets/_builtin/cub200.py
pmeier: This seems like a good thing to test in general. What should a test look like. Is something like

```python
for _ in dataset.cycle(2):
    pass
```

enough? If yes, my proposal passes this test.

ejguan: I mean if we put the dataset (datapipes) into DataLoader, the second epoch of DataLoader would break.

torchvision/prototype/datasets/_builtin/cub200.py
ejguan: Yeah. Then, we can materialize the data from it like a meta-datapipe.

torchvision/prototype/datasets/_builtin/cub200.py
pmeier: So something like

```python
data_loader = DataLoader2(dataset)

for epoch in range(2):
    for sample in data_loader:
        pass
```

?

ejguan: Yeah.

pmeier: My proposal still works. I've pushed the test I'm running against. There are multiple failures for other datasets, but `cub200` is not one of them.

pmeier:
The agreed upon idiom is that we can iterate before the full pipeline is built, but we have to exhaust it completely. See #6065 for fixes to `ImageNet` and `CUB200`.

PR #5220, opened 2022-01-19
Add Rendered sst2 dataset

jdsgomes:
Addresses #5108


torchvision/datasets/rendered_sst2.py
```diff
@@ -0,0 +1,91 @@
...
+
+        if not self._check_exists():
+            raise RuntimeError("Dataset not found. You can use download=True to download it")
+
+        self._labels = []
+        self._image_files = []
+
+        for p in (self._base_folder / self._split).glob("**/*.png"):
+            self._labels.append(self.class_to_idx[p.parent.name])
+            self._image_files.append(p)
```
NicolasHug, line 57: I think this is something that [`make_dataset()`](https://github.com/pytorch/vision/blob/main/torchvision/datasets/folder.py#L48:L48) could be used for. But the code here is very simple so IMHO it's fine to keep as-is (perhaps @pmeier can share his thoughts).

torchvision/datasets/rendered_sst2.py
```diff
@@ -0,0 +1,91 @@
...
+        return image, label
+
+    def extra_repr(self) -> str:
+        return f"split={self._split}"
+
+    def _check_exists(self) -> bool:
+        for class_label in set(self.classes):
+            if not (
+                (self._base_folder / self._split / class_label).exists()
+                and (self._base_folder / self._split / class_label).is_dir()
```
NicolasHug, line 83: Nit: I think that `is_dir()` properly returns False when the directory does not exist, so perhaps we can avoid using `exists()`:

```py
In [1]: from pathlib import Path

In [2]: Path("alfjnaljefeajlfbaeljnaljen").is_dir()
Out[2]: False
```

test/test_datasets.py
```diff
@@ -2665,5 +2665,27 @@ def inject_fake_data(self, tmpdir: str, config):
...
+class RenderedSST2TestCase(datasets_utils.ImageDatasetTestCase):
+    DATASET_CLASS = datasets.RenderedSST2
+    FEATURE_TYPES = (PIL.Image.Image, int)
+    ADDITIONAL_CONFIGS = datasets_utils.combinations_grid(split=("train", "valid", "test"))
+
+    def inject_fake_data(self, tmpdir: str, config):
+        root_folder = pathlib.Path(tmpdir) / "rendered-sst2"
+        image_folder = root_folder / config["split"]
+
+        num_images_per_class = 5
```
NicolasHug, line 2677: To slightly increase robustness:

```suggestion
        num_images_per_class = {"train": 5, "test": 6, "val": 7}
```

test/test_datasets.py
```diff
@@ -2665,5 +2665,27 @@ def inject_fake_data(self, tmpdir: str, config):
         return num_images
 
 
+class RenderedSST2TestCase(datasets_utils.ImageDatasetTestCase):
+    DATASET_CLASS = datasets.RenderedSST2
+    FEATURE_TYPES = (PIL.Image.Image, int)
```
NicolasHug, line 2670: Nit: we don't need this line as it's the default of the `datasets_utils.ImageDatasetTestCase` class

torchvision/datasets/rendered_sst2.py
```diff
@@ -0,0 +1,91 @@
...
+    This dataset contains two classes (positive and negative) and is divided in three splits: a  train
+    split containing 6920 images (3610 positive and 3310 negative), a validation split containing 872 images
+    (444 positive and 428 negative), and a test split containing 1821 images (909 positive and 912 negative).
+
+    Args:
+        root (string): Root directory of the dataset.
+        split (string, optional): The dataset split, supports ``"train"`` (default), `"valid"` and ``"test"``.
+        transform (callable, optional): A function/transform that  takes in an PIL image and returns a transformed
+            version. E.g, ``transforms.RandomCrop``.
+        target_transform (callable, optional): A function/transform that takes in the target and transforms it.
```
NicolasHug, line 26: Nit: Download docstring is missing but I can also add it later in my follow-up PR https://github.com/pytorch/vision/pull/5164

NicolasHug (approved):
Thanks @jdsgomes , this looks great! I only have very minor comments below. I'll approve now, perhaps @pmeier can give this a quick look too?

NicolasHug:
Thanks a lot @jdsgomes !

torchvision/datasets/rendered_sst2.py
pmeier: Yeah, `make_dataset` could make this even shorter:

Either

```python
self._image_files, self._labels = zip(*make_dataset(str(self._base_folder / self._split)))
```

or

```python
self._samples = make_dataset(str(self._base_folder / self._split))
```

and do

```python
image_file, label = self._samples[idx]
```

in `__getitem__`.

No strong opinion, but if we go for `make_dataset`, I would prefer the latter option.

NicolasHug: I took care of that in https://github.com/pytorch/vision/pull/5164 !

PR #5224, opened 2022-01-19
improve error message for prototype datasets without options

pmeier:
```python

datasets.load("semeion", foo="foo")
```

Before:

```
ValueError: Unknown option 'foo' of dataset semeion. Can be .
```

After:

```
ValueError: Dataset semeion does not take any options, but got 'foo'.
```

cc @pmeier @bjuncek


torchvision/prototype/datasets/utils/_dataset.py
```diff
@@ -78,6 +78,15 @@ def read_categories_file(path: pathlib.Path) -> List[List[str]]:
...
 
     def make_config(self, **options: Any) -> DatasetConfig:
+        if not self._valid_options:
+            if options:
+                raise ValueError(
+                    f"Dataset {self.name} does not take any options, "
+                    f"but got {sequence_to_str(list(options), separate_last=' and')}."
+                )
+
+            return DatasetConfig()
```
NicolasHug, line 88: Would `return DatasetConfig(self.default_config, **options)` work here?


Because if it does we could simplify to just

```py
        if not self._valid_options and options:
                raise ValueError(
                    f"Dataset {self.name} does not take any options, "
                    f"but got {sequence_to_str(list(options), separate_last=' and')}."
                )
```

and return only at the end

pmeier: Yes it would. Both `self.default_config` and `options` would be empty dicts here. I thought an early exit if no options are available and none are specified would be preferable, but I don't mind your suggestion either.

PR #5226, opened 2022-01-20
[ViT] Adding conv_stem support

sallysyw:
In this PR, we support the feature introduced in this paper: [Early Convolutions Help Transformers See Better](https://arxiv.org/abs/2106.14881)

Code reference: [Classy Vision Implementation](https://github.com/facebookresearch/ClassyVision/blob/main/classy_vision/models/vision_transformer.py#L219-L252)

Testing: successfully converted classyvision's vitc model to be loadable by torchvision. 
Meta-only: D33675072

cc @datumbox


torchvision/models/vision_transformer.py
```diff
@@ -122,6 +130,8 @@ def forward(self, input: torch.Tensor):
 class VisionTransformer(nn.Module):
     """Vision Transformer as per https://arxiv.org/abs/2010.11929."""
 
+    conv_proj: nn.Module
```
sallysyw, line 133: otherwise lint would complain

torchvision/models/vision_transformer.py
```diff
@@ -148,12 +159,32 @@ def __init__(
...
 
-        # The conv_proj is a more efficient version of reshaping, permuting
-        # and projecting the input
-        self.conv_proj = nn.Conv2d(input_channels, hidden_dim, kernel_size=patch_size, stride=patch_size)
+        if self.conv_stem_layers is not None:
+            # As per https://arxiv.org/abs/2106.14881
+            self.conv_proj = nn.Sequential()
+            prev_channels = input_channels
+            for i, conv_stem_layer in enumerate(self.conv_stem_layers):
+                if isinstance(conv_stem_layer, Dict):
```
sallysyw, line 171: Line 171-174 looks redundant but otherwise lint complains

torchvision/models/vision_transformer.py
```diff
@@ -26,6 +26,14 @@
 }
 
 
+class ConvStemLayer(NamedTuple):
+    out_channels: int
+    kernel_size: int
+    stride: int
+    padding: Union[str, int]
```
datumbox, line 33: Nowhere else in TorchVision we accept padding as string. If you want to estimate it automatically, we encode this by passing None.
```suggestion
    padding: int
```

torchvision/models/vision_transformer.py
```diff
@@ -135,6 +145,7 @@ def __init__(
         num_classes: int = 1000,
         representation_size: Optional[int] = None,
         norm_layer: Callable[..., torch.nn.Module] = partial(nn.LayerNorm, eps=1e-6),
+        conv_stem_layers: Union[List[ConvStemLayer], List[Dict], None] = None,
```
datumbox, line 148: I think we should choose to either accept the named tuple or the dictionary, not both. This will lead to a simpler API. I think the idea of using a config object is a good approach comparing to a Dictionary because it clearly defines the supported supported options:
```suggestion
        conv_stem_layers: Optional[List[ConvStemLayer]] = None,
```

torchvision/models/vision_transformer.py
```diff
@@ -148,12 +159,32 @@ def __init__(
         self.num_classes = num_classes
         self.representation_size = representation_size
         self.norm_layer = norm_layer
+        self.conv_stem_layers = conv_stem_layers
```
datumbox, line 162: Do we need to store this within the model and have it serialized? I don't think we need to use its info anywhere else.

torchvision/models/vision_transformer.py
```diff
@@ -26,6 +26,14 @@
 }
 
 
+class ConvStemLayer(NamedTuple):
```
datumbox, line 29: This is not a layer but rather the configuration object of the layer. We got [similar objects](https://github.com/pytorch/vision/blob/eac3dc7bab436725b0ba65e556d3a6ffd43c24e1/torchvision/models/mobilenetv3.py#L38) in vision, so I recommend renaming:
```suggestion
class ConvStemConfig(NamedTuple):
```

torchvision/models/vision_transformer.py
```diff
@@ -148,12 +159,32 @@ def __init__(
...
+                    conv_stem_layer_params = conv_stem_layer._asdict()
+                self.conv_proj.add_module(
+                    f"conv_{i}",
+                    nn.Conv2d(
+                        in_channels=prev_channels,
+                        **conv_stem_layer_params,
+                    ),
+                )
+                self.conv_proj.add_module(f"bn_{i}", nn.BatchNorm2d(conv_stem_layer_params["out_channels"]))
+                self.conv_proj.add_module(f"relu_{i}", nn.ReLU())
```
datumbox, line 183: We should try to use the common blocks of Vision as much as we can. For Conv-Norm-Activation we have `ConvNormActivation`. See examples on [RegNets](https://github.com/pytorch/vision/blob/e3767f8aa51921460c89e1d396ccce097cebc9d5/torchvision/models/regnet.py#L91).

torchvision/models/vision_transformer.py
```diff
@@ -185,9 +216,10 @@ def __init__(
         self._init_weights()
 
     def _init_weights(self):
-        fan_in = self.conv_proj.in_channels * self.conv_proj.kernel_size[0] * self.conv_proj.kernel_size[1]
-        nn.init.trunc_normal_(self.conv_proj.weight, std=math.sqrt(1 / fan_in))
-        nn.init.zeros_(self.conv_proj.bias)
+        if self.conv_stem_layers is None:
```
datumbox, line 219: To avoid needing to store this, you can check that `isinstance(self.conv_proj, nn.Conv2d)`.

torchvision/models/vision_transformer.py
```diff
@@ -148,12 +159,32 @@ def __init__(
         self.num_classes = num_classes
         self.representation_size = representation_size
         self.norm_layer = norm_layer
+        self.conv_stem_layers = conv_stem_layers
 
         input_channels = 3
```
datumbox, line 164: Looks like this could be moved inside the else statement below or hardcoded directly in line 187.

torchvision/models/vision_transformer.py
```diff
@@ -185,9 +216,10 @@ def __init__(
         self._init_weights()
 
     def _init_weights(self):
-        fan_in = self.conv_proj.in_channels * self.conv_proj.kernel_size[0] * self.conv_proj.kernel_size[1]
-        nn.init.trunc_normal_(self.conv_proj.weight, std=math.sqrt(1 / fan_in))
-        nn.init.zeros_(self.conv_proj.bias)
+        if self.conv_stem_layers is None:
+            fan_in = self.conv_proj.in_channels * self.conv_proj.kernel_size[0] * self.conv_proj.kernel_size[1]
+            nn.init.trunc_normal_(self.conv_proj.weight, std=math.sqrt(1 / fan_in))
+            nn.init.zeros_(self.conv_proj.bias)
```
datumbox, line 222: What about the initialization scheme for the stem layers case? Shouldn't we be initializing them as well or the default will do? cc @mannatsingh

torchvision/models/vision_transformer.py
```diff
@@ -148,12 +159,32 @@ def __init__(
...
-        self.conv_proj = nn.Conv2d(input_channels, hidden_dim, kernel_size=patch_size, stride=patch_size)
+        if self.conv_stem_layers is not None:
+            # As per https://arxiv.org/abs/2106.14881
+            self.conv_proj = nn.Sequential()
+            prev_channels = input_channels
+            for i, conv_stem_layer in enumerate(self.conv_stem_layers):
+                if isinstance(conv_stem_layer, Dict):
+                    conv_stem_layer_params = ConvStemLayer(**conv_stem_layer)._asdict()
+                elif isinstance(conv_stem_layer, ConvStemLayer):
+                    conv_stem_layer_params = conv_stem_layer._asdict()
```
datumbox, line 174: The CI is [complaining](https://app.circleci.com/pipelines/github/pytorch/vision/14096/workflows/c10e5394-a469-46f8-b99a-6dda938e974a/jobs/1134575) about this. It would be best to avoid calling private methods and instead explicitly access the parameters that we want. Using the `ConvNormActivation` will help you do this easier anyway as you will only have to pass the parameters once.

datumbox (commented):
@sallysyw Thanks for the update.

I left a few comments related to making the implementation closer to the conventions used by TorchVision. Note that there are a few tests failing and these are related, so please have a look to fix.

Finally what is the strategy for testing this and confirming that the changes are like you expected? Do you plan to load weights from other models, test their accuracy and confirm?

torchvision/models/vision_transformer.py
mannatsingh: All the conv layers use default PyTorch init, except for the final 1x1 conv in the conv stem - https://github.com/facebookresearch/pycls/blob/main/pycls/models/vit.py#L275-L281.

torchvision/models/vision_transformer.py
sallysyw: But the `_init_weights` method needs to know whether this parameter is None or not, do we pass it explicitly?

datumbox (commented):
Thanks for the changes @sallysyw. I think we are close to merging. I've cced Mannat on 3 comments below to get his input and see if more simplifications are possible.

@sallysyw The tests are failing due to JIT and it's related to your changes. It's not clear what JIT doesn't like but worth checking it out.

torchvision/models/vision_transformer.py
```diff
@@ -158,32 +159,33 @@ def __init__(
...
-                    nn.Conv2d(
+                    f"conv_bn_relu_{i}",
+                    ConvNormActivation(
                         in_channels=prev_channels,
-                        **conv_stem_layer_params,
+                        out_channels=conv_stem_layer.out_channels,
+                        kernel_size=conv_stem_layer.kernel_size,
+                        stride=conv_stem_layer.stride,
+                        padding=conv_stem_layer.padding,
+                        norm_layer=nn.BatchNorm2d,
```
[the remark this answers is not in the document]
sallysyw: I think in the original repo they always use BN, see
https://github.com/facebookresearch/pycls/blob/main/pycls/models/vit.py#L144
https://github.com/facebookresearch/pycls/blob/8c79a8e2adfffa7cae3a88aace28ef45e52aa7e5/pycls/models/blocks.py#L32-L34

sallysyw:
> Finally what is the strategy for testing this and confirming that the changes are like you expected? Do you plan to load weights from other models, test their accuracy and confirm?

I converted the classy-vision's vitc checkpoints into torchvision's checkpoints and made sure that for a random input, the torchvision model and classyvision model produce the same output.

torchvision/models/vision_transformer.py
```diff
@@ -148,11 +158,32 @@ def __init__(
...
+                        activation_layer=nn.ReLU,
+                        bias=conv_stem_layer.bias,
+                    ),
+                )
+                prev_channels = conv_stem_layer.out_channels
+            self.conv_proj.add_module(
+                "conv_last", nn.Conv2d(in_channels=prev_channels, out_channels=hidden_dim, kernel_size=1)
+            )
+        else:
+            self.conv_proj = nn.Conv2d(  # type: ignore
```
datumbox, line 184: Ignoring is usually a last ditch attempt. We should fix properly.

Try the following: Define use a local variable `conv_proj` inside the if/else statement. Then right after the end of the if do:
`self.conv_proj: nn.Module = conv_proj`

torchvision/models/vision_transformer.py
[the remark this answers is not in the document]
datumbox: Yes, I understand that, but this feature is there to allow uses to change the standard BN with SyncBN easily. As-is, the API is limiting because a) the norm_layer provided by the user is not used and b) there is no way to overwrite the BN. I think we need to answer to 2 questions:
1. Are there any scenarios that make it replacing the BN necessary?
2. If it is necessary, it's clear that the default `norm_layer=LayerNorm` can't be used as a drop-in replacement for the norm layers of the ConvStem. Shall we make the layer overwriteable by adding it inside the `ConvStemConfig`?

@fmassa If you have any thoughts on this, please let us know.

torchvision/models/vision_transformer.py
datumbox, line 184: Ignoring is usually a last ditch attempt. We should fix properly.

Try the following: Define use a local variable conv_proj inside the if/else statement. Then right after the end of the if do:
self.conv_proj: nn.Module = conv_proj

datumbox (commented):
@sallysyw 
> I converted the classy-vision's vitc checkpoints into torchvision's checkpoints and made sure that for a random input, the torchvision model and classyvision model produce the same output.

This is great, thanks for checking. There is a significant amount of code that is not covered by any test. Since this is a feature of substantial size, we will need to cover it. Because we don't have a dedicated model builder method that uses the feature, it won't be tested by our standard `test_classification_model`. A quick workaround is to add a new test in [test_models.py](https://github.com/pytorch/vision/blob/main/test/test_models.py), define a nested simplified model builder within it and pass it as param to the `test_classification_model` method. The rest remains the same from there (you will need an expected file etc).

torchvision/models/vision_transformer.py
```diff
@@ -134,6 +143,7 @@ def __init__(
         num_classes: int = 1000,
         representation_size: Optional[int] = None,
         norm_layer: Callable[..., torch.nn.Module] = partial(nn.LayerNorm, eps=1e-6),
+        conv_stem_layers: Optional[List[ConvStemConfig]] = None,
```
kazhang, line 146: naming nit: `conv_stem_layers` looks like the layer instance while this is actually config, how about `conv_stem_config` or `conv_stem_params`?

torchvision/models/vision_transformer.py
sallysyw: I tried this in the latest version... looks like it's still complaining for the same reason as before :|

sallysyw: One solution is to use ignore for now and switch to Any when it's supported by pytorch:
https://pytorch.org/docs/stable/jit_language_reference.html#unsupported-typing-constructs

sallysyw:
Ah, this can be annoying... since that [`nn.Module`](https://pytorch.org/docs/stable/_modules/torch/nn/modules/module.html#Module) has this line:
```
T = TypeVar('T', bound='Module')
```

It forces the subclass to not return the looser type of `Module`, which causes mypy to complain because it thinks the type annotation `nn.Module` is incorrect: [example error](https://app.circleci.com/pipelines/github/pytorch/vision/14374/workflows/1f184498-f4f3-415d-96e0-d054cd565590/jobs/1156557)

which is documented [here](https://mypy.readthedocs.io/en/latest/generics.html#generic-methods-and-generic-self), and it suggests that:
image

However, using `Any` causes torchscript test failure because it hasn't support `Any` as a valid type yet: [reference](https://pytorch.org/docs/stable/jit_language_reference.html#unsupported-typing-constructs)


I'm afraid there's no workaround except ignoring the mypy error x_x

test/test_models.py
```diff
@@ -29,6 +30,47 @@ def get_models_from_module(module):
...
+        conv_stem_configs=test_vit_conv_stem_configs,
+        **kwargs,
+    )
+
+
+def get_models_from_builder():
+    # This function returns model builders created in this file
+    return [vitc_b_16, vitc_l_16]
+
+
```
datumbox, line 60: Nit:

Can you create a new test called `test_vitc_model(dev)` and move those definitions as nested methods within the test? You can then loop through the available methods `[vitc_b_16, vitc_l_16]` and call directly `test_classification_model(model_fn, dev)`. This way you will reuse the infra but keep things contained.

test/test_models.py
```diff
@@ -29,6 +30,47 @@ def get_models_from_module(module):
...
+        hidden_dim=1024,
+        mlp_dim=4096,
+        conv_stem_configs=test_vit_conv_stem_configs,
+        **kwargs,
+    )
+
+
+def get_models_from_builder():
+    # This function returns model builders created in this file
+    return [vitc_b_16, vitc_l_16]
```
datumbox: If you do this above this method won't be needed.

test/test_models.py
```diff
@@ -514,7 +556,7 @@ def test_generalizedrcnn_transform_repr():
     assert t.__repr__() == expected_string
 
 
-@pytest.mark.parametrize("model_fn", get_models_from_module(models))
+@pytest.mark.parametrize("model_fn", get_models_from_module(models) + get_models_from_builder())
```
datumbox, line 546: Similarly, this could be removed.

torchvision/models/vision_transformer.py
```diff
@@ -25,6 +26,16 @@
 }
 
 
+class ConvStemConfig(NamedTuple):
+    out_channels: int
+    kernel_size: int
+    stride: int
+    padding: int
+    norm_layer: nn.Module = nn.BatchNorm2d
```
datumbox, line 34: When you get issues with mypy, it's useful for look for similar patterns in the codebase. This type is wrong here, it's a callable that will return a Module. Aka:
```suggestion
    norm_layer: Callable[..., nn.Module] = nn.BatchNorm2d
```

torchvision/models/vision_transformer.py
```diff
@@ -25,6 +26,16 @@
...
+    activation_layer: nn.Module = nn.ReLU
```
datumbox, line 35: Same here:
```suggestion
    activation_layer: Callable[..., nn.Module] = nn.ReLU
```

torchvision/models/vision_transformer.py
```diff
@@ -148,11 +160,31 @@ def __init__(
...
+                        padding=conv_stem_layer_config.padding,
+                        norm_layer=conv_stem_layer_config.norm_layer,
+                        activation_layer=conv_stem_layer_config.activation_layer,
+                        bias=conv_stem_layer_config.bias,
+                    ),
+                )
+                prev_channels = conv_stem_layer_config.out_channels
+            conv_proj.add_module(
+                "conv_last", nn.Conv2d(in_channels=prev_channels, out_channels=hidden_dim, kernel_size=1)
+            )
```
datumbox, line 184: [I'm writing the comment here in relation to the `nn.Sequential` [mypy](https://app.circleci.com/pipelines/github/pytorch/vision/14374/workflows/1f184498-f4f3-415d-96e0-d054cd565590/jobs/1156557) issue you face. Github doesn't let me reply to the original thread because the code is outdated.]

Try renaming `conv_proj` to `seq_proj` and after the final add_module, do:
`conv_proj: nn.Module = seq_proj`

The above should work but another alternative (which I'm not 100% will work) is doing after the final add_module a cast:
`conv_proj = typing.cast(nn.Module, conv_proj)`

datumbox (commented):
I think we are getting closer to merging this. I've added some tips for mypy but couldn't put my comments to the previous thread for some reason...

We got 2 outstanding comments that I would love to get @mannatsingh's input: https://github.com/pytorch/vision/pull/5226#discussion_r791541395 and https://github.com/pytorch/vision/pull/5226#discussion_r791544245

torchvision/models/vision_transformer.py
```diff
@@ -158,32 +159,33 @@ def __init__(
...
-                    f"conv_{i}",
-                    nn.Conv2d(
+                    f"conv_bn_relu_{i}",
+                    ConvNormActivation(
                         in_channels=prev_channels,
-                        **conv_stem_layer_params,
+                        out_channels=conv_stem_layer.out_channels,
+                        kernel_size=conv_stem_layer.kernel_size,
+                        stride=conv_stem_layer.stride,
+                        padding=conv_stem_layer.padding,
```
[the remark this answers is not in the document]
mannatsingh: Agreed! Same padding is what we use!

torchvision/models/vision_transformer.py
```diff
@@ -158,32 +159,33 @@ def __init__(
...
+                        norm_layer=nn.BatchNorm2d,
+                        activation_layer=nn.ReLU,
+                        bias=conv_stem_layer.bias,
```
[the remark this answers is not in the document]
mannatsingh: This also sounds reasonable and covers the ConvStem use case!

mannatsingh (approved):
Added my comments about the 2 open questions. Please let me know if I missed anything!

torchvision/models/vision_transformer.py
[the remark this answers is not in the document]
datumbox: @mannatsingh Thanks for the clarification.

@sallysyw Sounds like we can drop bias from the `ConvNormActivation` definitions and from the `ConvStemConfig`.

torchvision/models/vision_transformer.py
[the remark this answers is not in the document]
datumbox: @mannatsingh Thanks again.

@sallysyw Same thing as above with padding. :)

test/test_models.py
```diff
@@ -514,6 +515,35 @@ def test_generalizedrcnn_transform_repr():
     assert t.__repr__() == expected_string
 
 
+test_vit_conv_stem_configs = [
```
datumbox, line 518: Nit: No need for defining this on module level. Could move in `vitc_b_16` as it's used only once.

test/test_models.py
```diff
@@ -514,6 +515,35 @@ def test_generalizedrcnn_transform_repr():
...
+    models.vision_transformer.ConvStemConfig(kernel_size=3, stride=2, out_channels=64),
+    models.vision_transformer.ConvStemConfig(kernel_size=3, stride=2, out_channels=128),
+    models.vision_transformer.ConvStemConfig(kernel_size=3, stride=1, out_channels=128),
+    models.vision_transformer.ConvStemConfig(kernel_size=3, stride=2, out_channels=256),
+    models.vision_transformer.ConvStemConfig(kernel_size=3, stride=1, out_channels=256),
+    models.vision_transformer.ConvStemConfig(kernel_size=3, stride=2, out_channels=512),
+]
+
+
+def vitc_b_16(**kwargs: Any):
```
datumbox, line 528: Nit: This could be moved straight in `test_vitc_models` with a minor refactoring.

datumbox (approved):
LGTM, thanks @sallysyw. The nits are non-blocking and can be addressed on separate cleanup PRs.

test/test_models.py
sallysyw: Will clean this up (with several other changes) in a seperate diff!

PR #5227, opened 2022-01-20
Fix DTD prototype dataset on Windows

pmeier:
This is the second time we are alerted by the torchdata folks that something broke, but without our CI is picking up on it. Both times it was related to Windows and we didn't notice, because we only run the prototype tests on a Linux machine. 

I'm wondering if we should switch to a Windows machine for prototype tests in CI given that AFAIK no one of us is developing on Windows. Thoughts @datumbox @NicolasHug ?

cc @pmeier @bjuncek


pmeier:
Turns out, this is actually an issue in `torchdata` with properly stripping the newline characters: pytorch/data#173.

pmeier:
With my fix in pytorch/data#174, we do not need to care about the line endings while generating or reading the files. Thus, this PR only fixes the mock data generation of DTD.

NicolasHug (approved):
Thanks @pmeier

PR #5229, opened 2022-01-20
allow single extension as str in make_dataset

pmeier:
Fixes https://github.com/pytorch/vision/pull/5164#discussion_r788643592.


test/test_datasets_utils.py
```diff
@@ -214,5 +216,30 @@ def test_verify_str_arg(self):
...
+    @pytest.mark.parametrize(
+        ("kwargs", "expected_error_msg"),
+        [
+            (dict(is_valid_file=lambda path: pathlib.Path(path).suffix in {".png", ".jpeg"}), "classes c"),
+            (dict(extensions=".png"), r"classes b, c.*?[.]png"),
+            (dict(extensions=[".png", ".jpeg"]), "c.*?[.]png, [.]jpeg"),
+        ],
+    )
+    def test_make_dataset_no_valid(self, tmpdir, kwargs, expected_error_msg):
+        tmpdir = pathlib.Path(tmpdir)
```
pmeier, line 229: TIL, that `tmpdir` is not a `pathlib.Path`, but rather a custom type which acts somewhat like a `pathlib.Path`. Still, methods like `.touch` do not exists, so we cast here.

test/test_datasets_utils.py
```diff
@@ -214,5 +216,30 @@ def test_verify_str_arg(self):
         pytest.raises(ValueError, utils.verify_str_arg, "b", ("a",), "arg")
 
 
+class TestFolderUtils:
+    @pytest.mark.parametrize(
+        ("kwargs", "expected_error_msg"),
+        [
+            (dict(is_valid_file=lambda path: pathlib.Path(path).suffix in {".png", ".jpeg"}), "classes c"),
+            (dict(extensions=".png"), r"classes b, c.*?[.]png"),
+            (dict(extensions=[".png", ".jpeg"]), "c.*?[.]png, [.]jpeg"),
```
pmeier, line 225:
```suggestion
            (dict(extensions=[".png", ".jpeg"]), "classes c.*?[.]png, [.]jpeg"),
```

test/test_datasets_utils.py
```diff
@@ -214,5 +216,30 @@ def test_verify_str_arg(self):
         pytest.raises(ValueError, utils.verify_str_arg, "b", ("a",), "arg")
 
 
+class TestFolderUtils:
```
NicolasHug, line 219: There's only one method, maybe we could just have a single function instead of a class?

torchvision/datasets/folder.py
```diff
@@ -17,7 +18,7 @@ def has_file_allowed_extension(filename: str, extensions: Tuple[str, ...]) -> bo
     Returns:
         bool: True if the filename ends with one of given extensions
     """
-    return filename.lower().endswith(extensions)
+    return filename.lower().endswith(extensions if isinstance(extensions, str) else tuple(extensions))
```
NicolasHug, line 21: Do we need this line? `endswith` is supposed to properly support strings and tuples of strings already

torchvision/datasets/folder.py
```diff
@@ -72,8 +73,11 @@ def make_dataset(
 
     if extensions is not None:
 
+        if isinstance(extensions, str):
+            extensions = (extensions,)
+
         def is_valid_file(x: str) -> bool:
-            return has_file_allowed_extension(x, cast(Tuple[str, ...], extensions))
+            return has_file_allowed_extension(x, cast(Collection[str], extensions))
```
NicolasHug, line 80: why do we need to cast here? Aren't the type annotations the same?

test/test_datasets_utils.py
NicolasHug, line 225: It's not really clear what the actual expected message is

NicolasHug (approved):
Thanks @pmeier , I'm going to approve to unlock but since we'll likely not use that utility anymore, I'm not convinced this is strongly needed

torchvision/datasets/folder.py
pmeier: If you look at the annotations, I annotated `extensions` with `Collection[str]`. This means one could also reasonable call with lists, sets or whatever adheres to the collection protocol. `str.endswith()` however only takes a tuple:

```python
>>> "abc".endswith(["a", "c"])
TypeError: endswith first arg must be str or a tuple of str, not list
```

Thus, we can either limit our input types or simply create a tuple here.

torchvision/datasets/folder.py
pmeier: `mypy` has a long standing issue that it doesn't pick up on outer type restriction in local functions. Thus, although logic excludes `None` and `str` here, inside `is_valid_file` `mypy` still thinks `extensions` has type `Optional[Union[str, Collection[str]]]`. Thus, we need to cast to tell it otherwise.

torchvision/datasets/folder.py
NicolasHug: Thanks for the details, it's fine to restrict to tuples if it makes the code easier to follow IMO. But I don't want to block either

test/test_datasets_utils.py
pmeier: `.*?` basically is a non-greedy expression that matches everything until we get a match from a later expression. So without regex syntax, this matches `classes b, c[...].png`. The only other alternative would be to produce the actual part that lies inbetween, for example: `class b, c. Supported extensions are: .png`

torchvision/datasets/folder.py
NicolasHug: Of course....


🥲

test/test_datasets_utils.py
NicolasHug: IMO there's value in making "the parts that lies inbetween" (and before, and after) explicit instead of hiding them in a regex. The documenting value of test is otherwise diminished.

torchvision/datasets/folder.py
pmeier: I've opted to change the `cast` to a `# type: ignore` to indicate that there is nothing "wrong" on our side, but this is actually a `mypy` issue.

PR #5233, opened 2022-01-20
add test if dataset samples can be collated

pmeier:


ejguan:
QQ: If there is no target in test set, why not emit target and only keep the `img`?

NicolasHug:
It's because we want to keep the returned types / length consistent across splits. Otherwise users need to unpack the returned values and check whether there is a target or not, which can be a cumbersome

ejguan:
> It's because we want to keep the returned types / length consistent across splits. Otherwise users need to unpack the returned values and check whether there is a target or not, which can be a cumbersome

Makes sense. And, what would we expect in the batch before collate if target is `None`?
- [(img0, None), (img1, None) ...]
- [img0, img1, ...]
And, we need a custom map (collate) function for either case.
Case 1:
```py
def collate_fn(batch):
    imgs, targets = tuple(zip(*batch))
    return default_collate(imgs), None
```
Case 2:
```py
def collate_fn(batch):
    return default_collate(batch), None
```
I'm using `None` to represent collated target because the collated images is a single instance of Tensor.

pmeier:
> And, what would we expect in the batch before collate if target is None?

That is the root of the problem: we cannot say past `[dict(...), dict(...), ...]`. It might be `dict(label=None)`, `dict(labels=[None, None])` or a even deeper nested container. This is why I said that `default_collate` is to inflexible. Let me propose a an alternative solution that makes solving our use case trivial.

pmeier:
I've summarized my proposal [here](https://gist.github.com/pmeier/da684e84283113ed6e427d6519d21efe). Please have a look @NicolasHug, @ejguan, @NivekT, @VitalyFedyunin

PR #5240, opened 2022-01-20
ci: Pin av to 8.0.3

seemethere:
The newest release does not provide support for Python 3.6 so let's just
pin to a version that does support that


atalman (approved):

seemethere:
cmake tests are expected to fail for this, merging

PR #5244, opened 2022-01-21
Replace Coco to COCO in enum fields

zhiqwang:
This modification is just to follow the suggestion in https://github.com/pytorch/vision/issues/5088#issuecomment-1016785307 .

cc @datumbox

### DIFF
--- a/torchvision/prototype/models/detection/keypoint_rcnn.py
+++ b/torchvision/prototype/models/detection/keypoint_rcnn.py
@@ -45,7 +45,7 @@ class KeypointRCNN_ResNet50_FPN_Weights(WeightsEnum):
             "map_kp": 61.1,
         },
     )
-    Coco_V1 = Weights(
+    COCO_V1 = Weights(
         url="https://download.pytorch.org/models/keypointrcnn_resnet50_fpn_coco-fc266e95.pth",
         transforms=CocoEval,
         meta={
@@ -56,7 +56,7 @@ class KeypointRCNN_ResNet50_FPN_Weights(WeightsEnum):
             "map_kp": 65.0,
         },
     )
-    default = Coco_V1
+    default = COCO_V1
 
 
 @handle_legacy_interface(
@@ -64,7 +64,7 @@ class KeypointRCNN_ResNet50_FPN_Weights(WeightsEnum):
         "pretrained",
         lambda kwargs: KeypointRCNN_ResNet50_FPN_Weights.Coco_Legacy
         if kwargs["pretrained"] == "legacy"
-        else KeypointRCNN_ResNet50_FPN_Weights.Coco_V1,
+        else KeypointRCNN_ResNet50_FPN_Weights.COCO_V1,
     ),
     weights_backbone=("pretrained_backbone", ResNet50_Weights.ImageNet1K_V1),
 )
@@ -101,7 +101,7 @@ def keypointrcnn_resnet50_fpn(
 
     if weights is not None:
         model.load_state_dict(weights.get_state_dict(progress=progress))
-        if weights == KeypointRCNN_ResNet50_FPN_Weights.Coco_V1:
+        if weights == KeypointRCNN_ResNet50_FPN_Weights.COCO_V1:
             overwrite_eps(model, 0.0)
 
     return model
--- a/torchvision/prototype/models/detection/ssd.py
+++ b/torchvision/prototype/models/detection/ssd.py
@@ -23,7 +23,7 @@ __all__ = [
 
 
 class SSD300_VGG16_Weights(WeightsEnum):
-    Coco_V1 = Weights(
+    COCO_V1 = Weights(
         url="https://download.pytorch.org/models/ssd300_vgg16_coco-b556d3b4.pth",
         transforms=CocoEval,
         meta={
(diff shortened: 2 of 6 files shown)


torchvision/prototype/models/detection/keypoint_rcnn.py
```diff
@@ -56,15 +56,15 @@ class KeypointRCNN_ResNet50_FPN_Weights(WeightsEnum):
...
         },
     )
-    default = Coco_V1
+    default = COCO_V1
 
 
 @handle_legacy_interface(
     weights=(
         "pretrained",
         lambda kwargs: KeypointRCNN_ResNet50_FPN_Weights.Coco_Legacy
```
datumbox, line 65: `Coco_Legacy` was missed

torchvision/prototype/models/detection/ssd.py
```diff
@@ -38,11 +38,11 @@ class SSD300_VGG16_Weights(WeightsEnum):
...
         },
     )
-    default = Coco_V1
+    default = COCO_V1
 
 
 @handle_legacy_interface(
-    weights=("pretrained", SSD300_VGG16_Weights.Coco_V1),
+    weights=("pretrained", SSD300_VGG16_Weights.COCO_V1),
     weights_backbone=("pretrained_backbone", VGG16_Weights.ImageNet1K_Features),
```
datumbox, line 46: `ImageNet1K_Features` was missed

datumbox (commented):
Thanks @zhiqwang. I'm not sure if @yoshitomo-matsubara was also looking into it renaming.

As you can see below I've noted that we missed a couple of names (except your intention was indeed to replace only coco as you say in your PR title). At any case, my recommendation would be to do all replacements in a single PR, so that we can review the proposal as a whole (assess readability). 

Hopefully the change is very straightforward and done mostly with IDEs (the reason why we opted for Enums instead of strings). If you prefer to split the work with @yoshitomo-matsubara, and do it piece by piece we can also create a feature branch send the individual PRs against it and finally at the end send one single PR for the entire change assessing the proposal as a whole. 

We are flexible so let me know your thoughts. Thanks!

torchvision/prototype/models/detection/keypoint_rcnn.py
zhiqwang: I'm not sure what it should be modified to, maybe `COCO_Legacy` ?

torchvision/prototype/models/detection/ssd.py
zhiqwang: I'm also not sure what it should be modified to, maybe `ImageNet1K_FEATURES`?

zhiqwang (commented):
> If you prefer to split the work with @yoshitomo-matsubara, and do it piece by piece we can also create a feature branch send the individual PRs against it and finally at the end send one single PR for the entire change assessing the proposal as a whole.

Hi @datumbox . Sure, I can help @yoshitomo-matsubara with this part of the work.

torchvision/prototype/models/detection/keypoint_rcnn.py
datumbox: Probably the latter. All caps. This is part of the reason why ideally we should have these changes on a single PR, so that we can see all these corner cases as a whole and decide a common approach.

torchvision/prototype/models/detection/ssd.py
datumbox: So in the discussion with @yoshitomo-matsubara at #5088, the proposal is to have them all capitalized. My concern is that this might hurt readability in some cases and in these we should consider slight renaming. 

Another approach could be to gather all enum values in a single issue and agree upon the renaming and then split the work in multiple PRs. Again I'm flexible on how you would like to do this as long as we get an at-a-glance view of all the modifications to ensure consistency. :)

yoshitomo-matsubara:
Hi @datumbox
Yes, I am still working on it locally. To avoid conflict, I'd like to leave `detection/` to @zhiqwang and will focus on other modules.

Hi @zhiqwang 
Is that ok?

datumbox:
@yoshitomo-matsubara Thanks for confirming you are working on it.

> To avoid conflict, I'd like to leave detection/ to @zhiqwang and will focus on other modules.

I'm not sure that's possible. The detection area contains references of the classification backbones, so your PRs will end up touching the same files at the same time.

yoshitomo-matsubara:
@datumbox 

> I'm not sure that's possible. The detection area contains references of the classification backbones, so your PRs will end up touching the same files at the same time.

Then, should I wait until this PR is merged and then start working on the rest?

datumbox:
@zhiqwang would it be OK to let @yoshitomo-matsubara send the full PR, given he has already started it? We can definitely get your help on the review. Of course there are alternatives we can do, such as careful merging and resolving conflicts but we might as well do the simplest thing. Let me know if that's OK with you, thanks!

zhiqwang:
Hi @datumbox and @yoshitomo-matsubara ,

> would it be OK to let @yoshitomo-matsubara send the full PR, given he has already started it?

Yep, I think this way would be better. As such I'm closing this PR, and thank you both for reviewing this.

yoshitomo-matsubara:
Thank you @zhiqwang !

I just made local commits and sent a PR #5257

PR #5246, opened 2022-01-21
[PoC] add test to check available feature types of prototype datasets

pmeier:


test/test_prototype_builtin_datasets.py
```diff
@@ -125,6 +140,20 @@ def scan(graph):
...
+            raise pytest.skip("No feature types are specified.")
+
+        with dataset_mock.prepare(config):
+            dataset = datasets.load(dataset_mock.name, **config)
+
+        sample_structure = dataset_mock.sample_structure
+        if callable(sample_structure):
+            sample_structure = sample_structure(config)
+
+        assert_equal(next(iter(dataset)), sample_structure, pair_types=[SampleStructurePair])
```
NicolasHug, line 155: Looks like we're comparing dicts of types, e.g.

```py
            dict(
                image=features.Image,
                label=features.Label,
            ),
```

Do we need to use `assert_equal` and rely on a Pair class here, or could we just use a simple `==` check?

pmeier: If we stick with flat dicts, I agree plain asserts in a single loop is preferable. If we want to go this way, I'll check if we have more complicated sample structures. In any way, we can also decide to flatten them.

pmeier:
Another idea came to mind: would it be worth having this meta data on the dataset info? I'm planning to auto-generate documentation from the info and it might be helpful to have the sample structure there.

PR #5248, opened 2022-01-21
Adding prototype flag on reference scripts

datumbox:
Previously we've added in the reference scripts a `--weights` argument which operates as a feature switch for the prototype area (`None` vs fully weights name). When this was first introduced, we just wanted to confirm that the new model builders yielded the same results as the old ones for pre-trained models. Thus we only focused on inference.

Unfortunately the old approach of checking if `weights` was `None` won't work if we have models that haven't been graduated from the prototype area and need training. In this PR, we add the explicit feature switch `--prototype` which allows us to access the prototype model builders.


references/video_classification/train.py
```diff
@@ -149,11 +150,14 @@ def main(args):
...
+    if not args.prototype:
+        transform_test = presets.VideoClassificationPresetEval(resize_size=(128, 171), crop_size=(112, 112))
     else:
-        weights = PM.get_weight(args.weights)
-        transform_test = weights.transforms()
+        if args.weights:
+            weights = PM.get_weight(args.weights)
+            transform_test = weights.transforms()
+        else:
+            transform_test = PT.Kinect400Eval(crop_size=(112, 112), resize_size=(128, 171))
```
datumbox, line 160: I confirm that the arguments in the two classes take the parameters in different order. We are going to remove the old presets anyway and stick with the second one.

references/classification/train.py
```diff
@@ -16,9 +16,9 @@
 
 
 try:
-    from torchvision.prototype import models as PM
```
sallysyw, line 19: I can foresee that in the future if we are going to import prototype datasets as well, we'll introduce another annotation named `PD`... so my question is, why not import the full prototype here? 
I assume the logic behind the way you raise `ImportError` is, if we're having import error with one of them, then we are setting all of them to None.
So I'm thinking about simplify the import to
```
try:
      import torchvision.prototype as prototype
except ImportError:
      prototype = None
```

datumbox: I was going for the minimum number of modifications on this PR. I don't mind refactoring if you think that's needed.

references/video_classification/train.py
```diff
@@ -360,6 +364,12 @@ def parse_args():
     parser.add_argument("--dist-url", default="env://", type=str, help="url used to set up distributed training")
 
     # Prototype models only
+    parser.add_argument(
+        "--prototype",
```
sallysyw, line 368: nit: about the naming, what if we want to use prototype datasets for training in the future? shall we restrict the name a little bit so that it only implies that we are loading prototype models?
But I'm ok with leaving it as it is right now and revisiting it in the future. It's up to you :D

datumbox: Let's address it separately. One of the original ideas was to switch references to the prototype APIs of Transforms and Datasets to facilitate testing. There are no BC guarantees on the reference scripts (as long as all previous models are reproducible), so we can make changes as needed.

sallysyw:
Overall LGTM, can you launch 1 to 2 jobs to verify it loads prototype models properly?

datumbox:
@sallysyw I've checked, see below. Please stamp to unblock.

```
torchrun --nproc_per_node=1 train.py --test-only --weights ConvNeXt_Tiny_Weights.ImageNet1K_V1 --model convnext_tiny 
ValueError: The weights parameter works only in prototype mode. Please pass the --prototype argument.
```

```
torchrun --nproc_per_node=1 train.py --test-only --prototype --weights ConvNeXt_Tiny_Weights.ImageNet1K_V1 --model convnext_tiny 
Test:  Acc@1 82.522 Acc@5 96.146
```

```
torchrun --nproc_per_node=1 train.py --test-only --prototype --model convnext_tiny 
Test:  Acc@1 0.092 Acc@5 0.444
```

```
torchrun --nproc_per_node=2 train.py --prototype --model convnext_tiny --batch-size 128 --opt adamw --lr 1e-3 --lr-scheduler cosineannealinglr --lr-warmup-epochs 5 --lr-warmup-method linear --auto-augment ta_wide --epochs 600 --random-erase 0.1 --label-smoothing 0.1 --mixup-alpha 0.2 --cutmix-alpha 1.0 --weight-decay 0.05 --norm-weight-decay 0.0 --train-crop-size 176 --model-ema --val-resize-size 232 --ra-sampler --ra-reps 4

Works as expected, output omitted.
```

references/detection/train.py
```diff
@@ -150,8 +159,10 @@ def get_args_parser(add_help=True):
 
 
 def main(args):
-    if args.weights and PM is None:
+    if args.prototype and prototype.models is None:
```
sallysyw, line 162: nit:
```
if args.prototype and prototype is None:
```

sallysyw (approved):

references/detection/train.py
datumbox: That's a bug not a nit! God spot!

datumbox:
@sallysyw Thanks for the careful reviews, checked again everything and looks OK.

PR #5253, opened 2022-01-21
Adding more ConvNeXt variants + Speed optimizations

datumbox:
Adding `convnext_small`, `convnext_base` and `convnext_large`.

### Small
```
torchrun --nproc_per_node=1 train.py --test-only --prototype --weights ConvNeXt_Small_Weights.IMAGENET1K_V1 --model convnext_small -b 1
Acc@1 83.616 Acc@5 96.650
```

### Base
```
torchrun --nproc_per_node=1 train.py --test-only --prototype --weights ConvNeXt_Base_Weights.IMAGENET1K_V1 --model convnext_base -b 1
Acc@1 84.062 Acc@5 96.870
```

### Large
```
torchrun --nproc_per_node=1 train.py --test-only --prototype --weights ConvNeXt_Large_Weights.IMAGENET1K_V1 --model convnext_large -b 1
Acc@1 84.414 Acc@5 96.976
```

Also switch from channels first + Conv1x1 to channels last + Linear layers as it gets a 20% speed boost.


datumbox:
image

@s9xie @liuzhuang13 as discussed at https://github.com/pytorch/vision/pull/5197#discussion_r786327747 here are some benchmarks on our side. This is the speed of TorchVision's current implementation VS after applying this patch https://github.com/pytorch/vision/pull/5253/commits/290440b0560f8bea65f8824f9e3a1016df88463e. I understand that your numbers were different but I acknowledge that my benchmarks focus only on a specific batch-size and hardware combination. The difference is big but I was hoping that we could fix the speed difference by optimizing the underlying kernel and avoid needing workarounds on TorchVision. Have you spoken with Core about it? Overall what are your thoughts?

liuzhuang13:
@datumbox Thanks! It seems after your patch our implementations are really similar, both using permuting and linear layers. If I understand your results correctly, yours is 15~20% faster than ours somehow. Any idea why? We will benchmark your current patched version against ours too.

datumbox:
@liuzhuang13 I've assumed it's because I don't use a custom LayerNorm. I would love to know if you can confirm the above with your benchmarks. I've also notified the PyTorch performance team to see if that's something known and work on or something worth improving on the future (I've tagged you and Saining to the post for FYI).

NicolasHug (approved):
Thanks @datumbox

datumbox:
Thanks for the review @NicolasHug. I'm going to move now the class to main TorchVision so that we can include it in the upcoming release.

PR #5257, opened 2022-01-22
Rename prototype weight names to comply with PEP8

yoshitomo-matsubara:
@datumbox @zhiqwang @pmeier 
As discussed in #5088, https://github.com/pytorch/vision/issues/5088#issuecomment-1016256185 and #5244, I renamed all the weights names, following [PEP8](https://www.python.org/dev/peps/pep-0008/#constants).


torchvision/prototype/models/detection/keypoint_rcnn.py
```diff
@@ -56,17 +56,17 @@ class KeypointRCNN_ResNet50_FPN_Weights(WeightsEnum):
             "map_kp": 65.0,
         },
     )
-    default = Coco_V1
+    default = COCO_V1
```
datumbox, line 59: I think all the `default` aliases in all enums need to become `DEFAULT` as well. Thoughts?

If we do this, there are a couple of places the default is referenced as a string in the code that we also need to change. Example [1](https://github.com/pytorch/vision/blob/main/torchvision/prototype/models/_utils.py#L81) and [2](https://github.com/pytorch/vision/blob/main/test/test_prototype_models.py#L58).

datumbox (commented):
The changes look great @yoshitomo-matsubara. Thanks for your help. I've left only 1 comment that needs to be addressed. Please have a look and let me know.

Finally, could you please add a test on `test/test_prototype_models.py` to confirm that all available weights have enums that are capitalized? This will help us catch cases were we missed to change them and also enforce this style on the future. You can do this on the `test_schema_meta_validation` where we already loop throw all the `w` values of the `weights_enum`. Thanks in advance.

torchvision/prototype/models/detection/keypoint_rcnn.py
yoshitomo-matsubara: @datumbox Yes, I agree with replace `default` with `DEFAULT`

yoshitomo-matsubara: done

datumbox (commented):
@yoshitomo-matsubara All look great. Only one remaining comment on the test and we are good to merge.

You can ignore the 2 failing tests as they are not caused by your PR. One is caused by an unrelated PR that we will revert and one by an upstream bug on Core which is being investigated.

yoshitomo-matsubara:
@datumbox Thank you for the feedback. I think all the unit tests were passed.

datumbox:
@yoshitomo-matsubara Somehow the test was dropped, so I've just added it back. I'll merge once everything passes.

datumbox (approved):
LGTM, thanks for your contribution @yoshitomo-matsubara. 

We'll need to update the [article](https://pytorch.org/blog/introducing-torchvision-new-multi-weight-support-api/) to reflect upon the changes. I'll ping the team to do it on my side. :)

Edit: The article was updated https://github.com/pytorch/pytorch.github.io/pull/915

yoshitomo-matsubara:
@datumbox 
Thank you for the reviews!

Issue #5228, opened 2022-01-20
Replace plain asserts with ValueErros in Resnet


oke-aditya:
As noted in https://github.com/pytorch/vision/issues/5209#issuecomment-1017281755

It would be great if these can be replaced by ValueErrors.

https://github.com/pytorch/vision/blob/main/torchvision/models/detection/backbone_utils.py#L114
https://github.com/pytorch/vision/blob/main/torchvision/models/detection/backbone_utils.py#L127

Maybe these too?
https://github.com/pytorch/vision/blob/main/torchvision/models/detection/backbone_utils.py#L155
https://github.com/pytorch/vision/blob/main/torchvision/models/detection/backbone_utils.py#L186

Once the assert is replaced  by valueError. We should add test to check for it. 
Probably here
https://github.com/pytorch/vision/blob/main/test/test_backbone_utils.py

cc @abhi-glitchhg

oke-aditya:
Hi! I see there are 2 PRs. @piyush01123 and @abhi-glitchhg can you guys sync a bit so that we can avoid duplication? Maybe work together and come up with tests?

abhi-glitchhg:
yes Sure, @piyush01123 what is your thought?
I have added tests for resnet_backbone, and am yet to add for the mobilenet_backbone; if you want to do that part let me know; we can merge our work in a single pr later.

piyush01123:
Sure @abhi-glitchhg .  I'll add the test for mobilenet backbone in `test_backbone_utils` and `test_models_detection_utils`. 
Also @oke-aditya, `mobilenet_backbone` function is missing docstring similar to this in`resnet_fpn_backbone` https://github.com/pytorch/vision/blob/59c723cb45d0f8ab897cc7836d408e9fdde4b552/torchvision/models/detection/backbone_utils.py#L67-L101 
Should I add it?

PR #5262, opened 2022-01-23
changed assert to ValueError in backbone_utils

piyush01123:
Solves #5228. I did not add tests because I could not think of any testing logic since these are mostly internal methods which do not get directly called in the testing script. Please let me know if this needs any changes.  
I am trying this again from a fresh pull of `main` because the earlier one failed CI tests.


piyush01123:
closing as we are working on this in #5275

PR #5264, opened 2022-01-24
Replace Assert statement with ValueError message.

abhi-glitchhg:
Yet to add tests

cc @oke-aditya 

#5228


torchvision/models/detection/backbone_utils.py
```diff
@@ -111,7 +111,8 @@ def _resnet_fpn_extractor(
 ) -> BackboneWithFPN:
 
     # select layers that wont be frozen
-    assert 0 <= trainable_layers <= 5
+    if trainable_layers not in range(0, 6):
```
oke-aditya, line 114: range() is a bit slow and this of course takes O(n) time to check. I suggest a simpler logic 
```suggestion
    if trainable_layers < 0 or trainable_layers > 5:
```

torchvision/models/detection/backbone_utils.py
```diff
@@ -124,7 +125,8 @@ def _resnet_fpn_extractor(
 
     if returned_layers is None:
         returned_layers = [1, 2, 3, 4]
-    assert min(returned_layers) > 0 and max(returned_layers) < 5
+    if min(returned_layers) <= 0 or max(returned_layers) >= 5:
+        raise ValueError(f" `returned_layers` object should contain integers between [1,4], got {returned_layers} ")
```
oke-aditya, line 129: Maybe more idiomatic is 
```suggestion
        raise ValueError(f"Each returned layer must be in the range [1,4]. Got {returned_layers}")
```

oke-aditya, line 129: Maybe more idiomatic is 
```suggestion
        raise ValueError(f"Each returned layer must be in the range [1,4]. Got {returned_layers}")
```

torchvision/models/detection/backbone_utils.py
```diff
@@ -152,7 +154,10 @@ def _validate_trainable_layers(
     # by default freeze first blocks
     if trainable_backbone_layers is None:
         trainable_backbone_layers = default_value
-    assert 0 <= trainable_backbone_layers <= max_value
+        if trainable_backbone_layers not in range(0, max_value + 1):
```
oke-aditya, line 157: Again range() would be bit slower. This is faster.

```suggestion
        if trainable_backbone_layers < 0 or trainable_backbone_layers > max_value:
```

torchvision/models/detection/backbone_utils.py
```diff
@@ -172,7 +177,7 @@ def mobilenet_backbone(
 def _mobilenet_extractor(
     backbone: Union[mobilenet.MobileNetV2, mobilenet.MobileNetV3],
     fpn: bool,
-    trainable_layers,
+    trainable_layers: int,
```
oke-aditya, line 180: Nice catch

torchvision/models/detection/backbone_utils.py
```diff
@@ -183,7 +188,8 @@ def _mobilenet_extractor(
     num_stages = len(stage_indices)
 
     # find the index of the layer from which we wont freeze
-    assert 0 <= trainable_layers <= num_stages
+    if trainable_layers not in range(0, num_stages + 1):
```
oke-aditya, line 191:
```suggestion
    if trainable_layers < 0 or trainable_layers > num_stages:
```

oke-aditya (commented):
Few thoughts.

torchvision/models/detection/backbone_utils.py
abhi-glitchhg: okay! Understood

torchvision/models/detection/backbone_utils.py
abhi-glitchhg: yes, replacing range generator with simple logic

test/test_backbone_utils.py
```diff
@@ -23,6 +23,13 @@ def test_resnet_fpn_backbone(backbone_name):
     y = resnet_fpn_backbone(backbone_name=backbone_name, pretrained=False)(x)
     assert list(y.keys()) == ["0", "1", "2", "3", "pool"]
 
+    with pytest.raises(ValueError):
```
oke-aditya, line 26: We should also check for error message, kindly append that too.
E.g. https://github.com/pytorch/vision/blob/main/test/test_utils.py#L306

abhi-glitchhg: oh, okay! Will do it

abhi-glitchhg:
#5275 contains commits from this PR; hence closed

Issue #5266, opened 2022-01-24
FCOS empty box images


barschiiii:
### 🐛 Describe the bug

I am playing around with new FCOS models (thanks for that) and am encountering issues when providing images without box annotations. This is a common use case in object detection, and also works for other detector models in torchvision.

A simple example to replicate:

```
model = fcos_resnet50_fpn(pretrained=True)
model(torch.zeros((1,3,512,512)), targets=[{"boxes": torch.empty(0,4), "labels": torch.empty(0,1).to(torch.int64)}])
```

An indexing error happens in `FCOSHead` when running `compute_loss` in this part:

```
all_gt_classes_targets = []
all_gt_boxes_targets = []
for targets_per_image, matched_idxs_per_image in zip(targets, matched_idxs):
    gt_classes_targets = targets_per_image["labels"][matched_idxs_per_image.clip(min=0)]
    gt_classes_targets[matched_idxs_per_image < 0] = -1  # backgroud
    gt_boxes_targets = targets_per_image["boxes"][matched_idxs_per_image.clip(min=0)]
    all_gt_classes_targets.append(gt_classes_targets)
    all_gt_boxes_targets.append(gt_boxes_targets)
```

A workaround seems to be necessary, when having empty targets. Happy for any guidance, maybe there is also a different way necessary for me to train on empty images.

@jdsgomes @xiaohu2015 @zhiqwang 

### Versions

Torchvision @ master

xiaohu2015:
@barschiiii I also found this issue, we plan to add another PR to handle this.

barschiiii:
@xiaohu2015 that's great to hear, hope I will be able to try it soon :)

xiaohu2015:
@hi, we fix this bug in https://github.com/pytorch/vision/pull/5267

datumbox:
> Using AMP mixed precision and Adam.

Thanks for providing this info, it helps the debugging. Could you try @xiaohu2015 patch without them and let us know if it's fixed? Thanks!

barschiiii:
@datumbox this is after applying the patch - do you mean running it without AMP?

barschiiii:
SGD with AMP, and Adam without AMP seem to both run fine. Adam with AMP and gradient clipping runs into instability (NaN) issues, also with different learning rates.

barschiiii:
It might be related to the default initialization of anchor boxes:

```
if anchor_generator is None:
[...]

PR #4961, opened 2021-11-18
add FCOS

xiaohu2015:
The pr is about https://github.com/pytorch/vision/issues/2707
We (@xiaohu2015 and @zhiqwang ) implemented torchvision-style FCOS by referencing to [detectron2](https://github.com/facebookresearch/detectron2) and [AdelaiDet](https://github.com/aim-uofa/AdelaiDet) repo.

if train with the retinanet training command:

```
torchrun --nproc_per_node=8 train.py\
    --dataset coco --model fcos_resnet50_fpn --epochs 26 \
    --lr-steps 16 22 --aspect-ratio-group-factor 3 --lr 0.01 --amp
```

you can get:

```
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.390
 Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=100 ] = 0.585
 Average Precision  (AP) @[ IoU=0.75      | area=   all | maxDets=100 ] = 0.415
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.221
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.419
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.511
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=  1 ] = 0.325
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets= 10 ] = 0.526
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.562
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.359
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.603
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.717
```
the AP result is better to the paper (38.6).

if train with training command (1x, 12 epoch, the dafult setting in the FCOS paper):

```
torchrun --nproc_per_node=8 train.py\
    --dataset coco --model fcos_resnet50_fpn --epochs 12 \
    --lr-steps 8 11 --aspect-ratio-group-factor 3 --lr 0.01 --amp
```

you can get:

```
[...]

PR #2784, opened 2020-10-10
RetinaNet object detection (take 2)

fmassa:
This is entirely based on top of the great work from @hgaiser in https://github.com/pytorch/vision/pull/1697

I'm creating a new PR because there are some minor things that could be fixed in that PR but I don't have rights to push to the PR, so in order to move faster I'm creating a new PR but all the history is kept here.

Here is the mAP for the uploaded model:
```
IoU metric: bbox
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.364
 Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=100 ] = 0.558
 Average Precision  (AP) @[ IoU=0.75      | area=   all | maxDets=100 ] = 0.383
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.193
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.400
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.490
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=  1 ] = 0.315
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets= 10 ] = 0.506
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.558
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.386
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.595
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.699
```

PR #5267, opened 2022-01-24
fix bug when the target is empty in FCOS

xiaohu2015:
Last week, we released the detetion model: FCOS at https://github.com/pytorch/vision/pull/4961#

some users have found there is a problem when the inputs has empty targets.
https://github.com/pytorch/vision/issues/5266

same issus can also be found in detectron2: https://github.com/facebookresearch/detectron2/pull/3851, https://github.com/facebookresearch/detectron2/pull/3910

we fix this bug in this pr, you can test:

```
model = fcos_resnet50_fpn(pretrained=False)
out = model(torch.zeros((1,3,512,512)), targets=[{"boxes": torch.empty(0,4), "labels": torch.empty(0,1).to(torch.int64)}])

## 
{'classification': tensor(1.5403, grad_fn=<DivBackward0>), 'bbox_regression': tensor(0., grad_fn=<DivBackward0>), 'bbox_ctrness': tensor(0., grad_fn=<DivBackward0>)}
```


torchvision/models/detection/fcos.py
```diff
@@ -59,9 +59,13 @@ def compute_loss(
         all_gt_classes_targets = []
         all_gt_boxes_targets = []
         for targets_per_image, matched_idxs_per_image in zip(targets, matched_idxs):
-            gt_classes_targets = targets_per_image["labels"][matched_idxs_per_image.clip(min=0)]
+            if len(targets_per_image["labels"]) == 0:
```
datumbox, line 62: We might need to confirm that this will works as expected and doesn't produce errors when `find_unused_params=False` (see discussion at https://github.com/pytorch/vision/pull/2784#discussion_r503818118)

The runtime error to be looking for is something like:
```
RuntimeError: Expected to have finished reduction in the prior iteration before starting a new one. This error indicates that your module has parameters that were not used in producing loss. You can enable unused parameter detection by (1) passing the keyword argument `find_unused_parameters=True` to `torch.nn.parallel.DistributedDataParallel`; (2) making sure all `forward` function outputs participate in calculating loss. If you already have done the above two steps, then the distributed data parallel module wasn't able to locate the output tensors in the return value of your module's `forward` function. Please include the loss function and the structure of the return value of `forward` of your module when reporting this issue (e.g. list, dict, iterable).
```

xiaohu2015: @datumbox thanks. I have not considered this problem. By the way, I want to know that how retinanet can handle this.

datumbox: We had this issue previously with a couple of models. Retina was one of them. The way we avoided it was by rewriting the loss estimations in a way that the vectors can cope with empty indeces. Let's check first if it is an issue before starting rewriting.

xiaohu2015: @datumbox I think the current implementation will not have this trouble, as you see, the regression loss and centerness loss is just in the way that the vectors can cope with empty indeces. but for safe, we had better check once. but I don't have the empty datasets, so can you help to check this?

datumbox: That's our understanding as well, that you should be OK. A single run on the scripts for 1 epoch should be enough to confirm if it's a problem (at least that was the case previously). I'll sync with Joao to confirm. :)

datumbox: Looks like the other user confirmed the patch works, see https://github.com/pytorch/vision/issues/5266#issuecomment-1020286871.

datumbox (approved):
The changes look good to me and it seems we have confirmation that the patch works from https://github.com/pytorch/vision/issues/5266#issuecomment-1020286871. I'm approving but I'll leave @jdsgomes merge once he reviews and gets the results of the 1 epoch run.

Thanks for the update @xiaohu2015!

torchvision/models/detection/fcos.py
jdsgomes: I can confirm that the patch works and also that after training for an epoch no runtime errors are observed.

jdsgomes (approved):
Looks good to me, thank you for the fix!

PR #5268, opened 2022-01-24
Mino simplifications to prototype dataset testings

NicolasHug:
This PR contains a few simplifications to the prototype dataset test suite as discussed offline with @pmeier


pmeier (approved):
LGTM, thanks a lot @NicolasHug!

PR #5270, opened 2022-01-24
Fix giou naming bug

jdsgomes:
This PR should fix the existing bug caused by the method name being the same as the module name.

before
```python
import torch
from torchvision.ops import generalized_box_iou_loss
generalized_box_iou_loss(torch.zeros((0,4)), torch.zeros((0,4)), reduction="sum")
  File "./IPython/core/interactiveshell.py", line 3457, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-23-efc1ac3887dc>", line 1, in <module>
    generalized_box_iou_loss(torch.zeros((0,4)), torch.zeros((0,4)), reduction="sum")
TypeError: 'module' object is not callable
```

After PR

```python
import torch
from torchvision.ops import generalized_box_iou_loss
generalized_box_iou_loss(torch.zeros((0,4)), torch.zeros((0,4)), reduction="sum")
tensor(0.)
```

cc @datumbox


datumbox (approved):
LGTM, thanks!

Issue #5269, opened 2022-01-24
Potential simplifications to the prototype dataset testing


NicolasHug:
As discussed offline with @pmeier, there are a few potential simplification / improvements we could make to the prototype datasets tests. 

- The tests currently create a tmp dir, but this tmp dir isn't cleaned up after the tests are run.
- The tests are currently designed such that the mocked dataset root is the same across all tests. Is this worth enforcing? We might be able to simplify the code if we don't.
- We currently rely on a cache, which isn't really needed in terms of speed (tests are fast enough for now). The cache leads to slightly more complicated code, in particular regarding what each dataset mock can return (a dict with datasetConfig as keys). Removing the cache might simplify the code without sacrificing too much in testing time. If the test time becomes an issue this is something we can revisit.

- qmnist test 50k is quite slow as it needs to generate a lot of images. Is it worth keeping the test? Is it worth even keeping the dataset at all? Should we think of a faster way to test it?

cc @pmeier @bjuncek

pmeier:
> qmnist test 50k is quite slow as it needs to generate a lot of images. [...] Is it worth even keeping the dataset at all?

Yes, I see no point in removing it. What is debatable is if we want to keep the `split="test10k"` and `split="test50k"` variants. The latter is the offender for the slow tests since it needs to generate at least 50k + 1 samples.

QMNIST was generated from the same source as MNIST with the goal of having more test samples while trying to match the processing of the original MNIST data. In the [words of the authors](https://github.com/facebookresearch/qmnist):

> The QMNIST10K test uses the first 10000 QMNIST testing digits, that is, the reconstruction of the ordinary MNIST testing set. The QMNIST50K test uses the remaining 50000 QMNIST testing digits, that is, the reconstruction of the lost MNIST testing digits.

IIUC, these two splits only have some merit when you want to compare QMNIST to MNIST. If you actually train on QMNIST, I don't think someone will actually use them since the whole idea of the dataset is to provide more test samples.

Thus, I'm ok with removing these splits and thus simplify the implementation and speed up the tests.

PR #5271, opened 2022-01-24
Use fixture for dataset root in tests and remove caching

NicolasHug:
Towards https://github.com/pytorch/vision/issues/5269, ~but this doesn't work yet.~ EDIT it does not thanks to @pmeier 's changes

@pmeier with this small POC, it looks like the resources still get downloaded for some reason, and the tests then fail due to a sha256 mismatch. Do you know what's going on?


test/test_prototype_builtin_datasets.py
```diff
@@ -3,14 +3,25 @@
...
+from builtin_dataset_mocks import parametrize_dataset_mocks, DATASET_MOCKS, TEST_HOME  # noqa: F401
 from torch.utils.data.datapipes.iter.grouping import ShardingFilterIterDataPipe as ShardingFilter
 from torch.utils.data.graph import traverse
 from torchdata.datapipes.iter import IterDataPipe, Shuffler
 from torchvision.prototype import transforms, datasets
 from torchvision.prototype.utils._internal import sequence_to_str
 
 
+@pytest.fixture(scope="session", autouse=True)
+def test_home(session_mocker, tmp_path_factory):
```
pmeier, line 15: This is a version as discussed yesterday where we the test does not need to take a `home` fixture to only pass it on to `dataset_mock.prepare`. 

A few things make the implementation a little complicated.

1. `pytest` only collects fixtures in `conftest.py` files or in the actual test module. Since the dataset mock module doesn't fall in either category, we need to keep the fixture here.
2. Since we now need to import `TEST_HOME` from another module, we need to make sure it is a pointer rather than a value. This means, we cannot `TEST_HOME = None` since after the import both values are decoupled. I've used a list here, but we can also have other objects.

pmeier:
If we remove the caching I get these timings on my local machine:

```
=============================================== slowest 10 durations ================================================
1.76s call     test/test_prototype_builtin_datasets.py::TestCommon::test_has_annotations[ShufflerIterDataPipe-imagenet-train]
1.75s call     test/test_prototype_builtin_datasets.py::TestCommon::test_decoding[imagenet-train]
1.45s call     test/test_prototype_builtin_datasets.py::TestCommon::test_num_samples[qmnist-test10k]
1.44s call     test/test_prototype_builtin_datasets.py::TestCommon::test_num_samples[qmnist-test]
1.19s call     test/test_prototype_builtin_datasets.py::TestCommon::test_num_samples[imagenet-train]
0.76s call     test/test_prototype_builtin_datasets.py::TestCommon::test_has_annotations[ShardingFilterIterDataPipe-imagenet-train]
0.76s call     test/test_prototype_builtin_datasets.py::TestCommon::test_sample[imagenet-train]
0.76s call     test/test_prototype_builtin_datasets.py::TestCommon::test_smoke[imagenet-train]
0.75s call     test/test_prototype_builtin_datasets.py::TestCommon::test_no_vanilla_tensors[imagenet-train]
0.74s call     test/test_prototype_builtin_datasets.py::TestCommon::test_traversable[imagenet-train]
[...]
========================================= 1292 passed, 12 xpassed in 42.61s =========================================
```

On the `main` branch, I get these timings:

```
=============================================== slowest 10 durations ================================================
1.02s call     test/test_prototype_builtin_datasets.py::TestCommon::test_num_samples[qmnist-test10k]
1.02s call     test/test_prototype_builtin_datasets.py::TestCommon::test_num_samples[qmnist-test]
0.71s call     test/test_prototype_builtin_datasets.py::TestCommon::test_smoke[imagenet-train]
0.42s call     test/test_prototype_builtin_datasets.py::TestCommon::test_num_samples[imagenet-train]
0.39s call     test/test_prototype_builtin_datasets.py::TestCommon::test_smoke[qmnist-test]
0.19s call     test/test_prototype_builtin_datasets.py::TestCommon::test_smoke[caltech101]
0.19s call     test/test_prototype_builtin_datasets.py::TestCommon::test_smoke[caltech256]
0.12s call     test/test_prototype_builtin_datasets.py::TestCommon::test_num_samples[caltech256]
0.09s call     test/test_prototype_builtin_datasets.py::TestCommon::test_num_samples[caltech101]
0.04s call     test/test_prototype_builtin_datasets.py::TestCommon::test_smoke[fer2013-train]
[...]
========================================= 1292 passed, 12 xpassed in 7.64s ==========================================
```

As is we are talking about a 5-6x increase in test time. Excluding imagenet (I'll look into the slow test in a separate PR) and qmnist (#5276), the quotient of the timings is in the same range. 

Given that we have quite a few more datasets to add and probably also more tests, I think the caching is already viable. Thoughts @NicolasHug?

NicolasHug:
Thanks a lot for taking a look and for the benchmark!

I agree a 6X increase times **sounds** bad... but in absolute terms the tests still run in under 1 minute, so it's nothing compared to the rest of the tests. Especially considering that we run these tests in a separate CI job, it will never become a bottleneck.

Since removing the cache allows for a much simpler code, I'd tend towards removing it now. The cache also violates a basic assumption that all unit tests should be unitary without any side effect (as you explained offline, the current cache side-effect where the reason my first version didn't work).

There's a good chance that the cache will be useful in the future though, if/when we run these tests in the same CI jobs as the rest. But I feel like this is something we should consider later: as you said we'll likely add more tests, and if we have the cache now, we'll have to resort to workarouds that we otherwise wouldn't (an example is the somewhat complex dict of dict returns of the mocks that we wouldn't need without the cache).

NicolasHug (commented):
I'm happy to merge as-is @pmeier, this is a nice simplification.
but I can't self-approve this PR because I'm technically still the author :p . LMK what you think regarding the cache above

pmeier (approved):
I'm kinda irresolute on this. I'll approve for now, but let's revisit this when when / if we are merging theses tests into our main ones.

PR #5273, opened 2022-01-25
add docstring for _log_api_usage_once

kazhang:
closes: #4957

### DIFF
--- a/torchvision/utils.py
+++ b/torchvision/utils.py
@@ -388,6 +388,21 @@ def _generate_color_palette(num_masks: int):
 
 
 def _log_api_usage_once(obj: Any) -> None:
+
+    """
+    Logs API usage(module and name) within an organization.
+    In a large ecosystem, it's often useful to track the PyTorch and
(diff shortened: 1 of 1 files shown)


datumbox (commented):
@kazhang Previously the community raised concerns over the nature of logging. Though the proposed documentation is factually correct, I feel it doesn't adequately explain what this is and how it can be used.

I would recommend to  cover the following points within the docstring:
- The function provides similar functionality as the `logging` module from the Python stdlib. 
- It can be used for debugging purposes to log which methods are used and by default it is inactive (no-op), unless the user manually subscribes a logger via the [SetAPIUsageLogger](https://github.com/pytorch/pytorch/blob/eb3b9fe719b21fae13c7a7cf3253f970290a573e/c10/util/Logging.cpp#L102-L117) method.
- It doesn't collect any data from open-source users and by default is no-op.

datumbox (approved):
LGTM, thanks @kazhang!

PR #3403, opened 2021-02-15
Add SSD architecture with VGG16 backbone

datumbox:
Resolves #440 and resolves #1422

This PR implements SSD with VGG16 backbone as described in the [original paper](https://arxiv.org/pdf/1512.02325.pdf). 

Trained using the code committed at https://github.com/pytorch/vision/pull/3403/commits/a167edc8faeeae1e0f765c725b70c73d9d160d5a. The current [best](https://download.pytorch.org/models/ssd300_vgg16_coco-b556d3b4.pth) pre-trained model was trained with (using latest git hash):

```
python -m torch.distributed.launch --nproc_per_node=8 --use_env train.py\
    --dataset coco --model ssd300_vgg16 --epochs 120\
    --lr-steps 80 110 --aspect-ratio-group-factor 3 --lr 0.002 --batch-size 4\
    --weight-decay 0.0005 --data-augmentation ssd
```

Submitted batch job 40773612

Accuracy metrics:
```
Epoch 119:
0: IoU metric: bbox
0:  Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.251
0:  Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=100 ] = 0.415
0:  Average Precision  (AP) @[ IoU=0.75      | area=   all | maxDets=100 ] = 0.262
0:  Average Precision  (AP) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.055
0:  Average Precision  (AP) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.268
0:  Average Precision  (AP) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.435
0:  Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=  1 ] = 0.239
0:  Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets= 10 ] = 0.344
0:  Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.365
0:  Average Recall     (AR) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.088
0:  Average Recall     (AR) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.406
0:  Average Recall     (AR) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.602
```

Validated with:
```
[...]

PR #5278, opened 2022-01-25
Fix a bug and document RandomZoomOut

datumbox:
While investigating a user question in relation to the RandomZoomOut transform at https://github.com/pytorch/vision/pull/3403#issuecomment-982640253, I noticed that the `torch.rand(1)` check is flipped the other way around. It had no effects on the training of SSD because we use `p=0.5` but the bug needs to be fixed.

Moreover I've added a comment to explain why an extra bit is necessary on the transform. I think the comment is necessary since I also have forgotten why the extra code is necessary and almost removed it.

### DIFF
--- a/references/detection/transforms.py
+++ b/references/detection/transforms.py
@@ -189,7 +189,7 @@ class RandomZoomOut(nn.Module):
             elif image.ndimension() == 2:
                 image = image.unsqueeze(0)
 
-        if torch.rand(1) < self.p:
+        if torch.rand(1) >= self.p:
             return image, target
 
         orig_w, orig_h = F.get_image_size(image)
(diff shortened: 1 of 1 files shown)


NicolasHug (approved):
Thanks @datumbox

jdsgomes (approved):
LGTM, thanks

datumbox:
@xiaohu2015 Sorry for getting ages to answer your question at https://github.com/pytorch/vision/pull/3403#issuecomment-982640253. I had to catch up with other things and forgot to check. This extra bit is indeed needed; I've added it to get around the limitation of PyTorch pad, which only accept integers for the `fill` value and thus the extra code ensures we end up with the background colour we asked. This is only needed when the image is Tensor as PIL already supports multi-colour fill.

PR #5282, opened 2022-01-26
Minor changes to prototype datasets

NicolasHug:
Some minor changes we discussed earlier with @pmeier


torchvision/prototype/datasets/_api.py
```diff
@@ -23,8 +23,7 @@ def register(dataset: Dataset) -> None:
         register(obj())
 
 
-# This is exposed as 'list', but we avoid that here to not shadow the built-in 'list'
-def _list() -> List[str]:
+def list_names() -> List[str]:
```
NicolasHug, line 26: I went for `list_names` since it corresponds to the `name` parameter of the `load()` function. Alternatives I could think of were more verbose (e.g. `list_available_datasets()`) but I'm happy to hear others thoughts

pmeier: To recap: `list` is a fitting name and this works fine if you do something like

```py
datasets.list()
```

but you cannot (or better should not) do

```py
from torchvision.prototype.datasets import list
```

Since we don't want to enforce a specific style on the user, we should use a name that works in both cases.

pmeier: TensorFlow datasets uses [`list_builders`](https://www.tensorflow.org/datasets/api_docs/python/tfds/list_builders) probably for the same reason. That would correspond to `lists_datasets` for us, which I like a little better than `list_names`.

pmeier (approved):
Minor comments inline. Thanks a lot @NicolasHug!

torchvision/prototype/datasets/utils/_resource.py
```diff
@@ -32,23 +32,16 @@ def __init__(
...
-        loader: Optional[Callable[[pathlib.Path], IterDataPipe[Tuple[str, IO]]]] = None,
     ) -> None:
         self.file_name = file_name
         self.sha256 = sha256
 
-        if preprocess and (decompress or extract):
-            warnings.warn("The parameters 'decompress' and 'extract' are ignored when 'preprocess' is passed.")
-        elif extract:
-            preprocess = self._extract
+        if extract:
```
pmeier, line 39:
```suggestion
        self._preprocess: Optional[Callable[[pathlib.Path], pathlib.Path]]
        if extract:
```

Issue #5045, opened 2021-12-07
[Discussion] How do we want to handle `torchvision.prototype.features.Feature`'s?


pmeier:
This issue should spark a discussion about how we want to handle `Feature`'s in the future. There are a lot of open questions I'm trying to summarize. I'll give my opinion to each of them. You can find the current implementation under `torchvision.prototype.features`.

## What are `Feature`'s?

`Feature`'s are subclasses of `torch.Tensor` and their purpose is threefold:

1. With their type, e.g. `Image`, they information about the data they carry. The prototype transformations (`torchvision.prototype.transforms`) use this information to automatically dispatch an input to the correct kernel.
2. They can optionally carry additional meta data that might be needed for transforming the feature. For example, most geometric transformations can only be performed on bounding boxes if the size of the corresponding image is known.
3. They provide a convenient interface for feature specific functionality, for example transforming the format of a bounding box.

There are currently three `Feature`'s implemented

- `Image`,
- `BoundingBox`, and
- `Label`,

but in the future we should add at least three more:

- `SemanticSegmentationMask`,
- `InstanceSegementationMask`, and
- `Video`.

## What is the policy of adding new `Feature`'s?

We could allow subclassing of `Feature`'s. On the one hand, this would make it easier for datasets to conveniently bundle meta data. For example, the COCO dataset could return a `CocoLabel`, which in addition to the default `Label.category` could also have the `super_category` field. On the other hand, this would also mean that the transforms need to handle subclasses of features well, for example a `CocoLabel` could be treated the same as a `Label`.

I see two downsides with that:

1. What if a transform needs the additional meta data carried by a feature subclass? Imagine I've added a special transformation that needs `CocoLabel.super_category`. Although from the surface this now supports plain `Label`'s this will fail at runtime.
2. Documentation custom features is more complicated than documenting a separate field in the sample dictionary of a dataset.

Thus, I'm leaning towards only having a few base classes.

## From what data should a `Feature` be instantiable?

Some of the features like `Image` or `Video` have non-tensor objects that carry the data. Should these features know how to handle them? For example should something like `Image(PIL.Image.open(...))` work?

My vote is out for yes. IMO this is very convenient and also not an unexpected semantic compared to passing the data directly, e.g. `Image(torch.rand(3, 256, 256))`

## Should `Feature`'s have a fixed shape?

Consider the following table:

| `Feature`                   | `.shape`                      |
|-----------------------------|-------------------------------|
[...]

PR #5283, opened 2022-01-26
revamp prototype features

pmeier:
This PR was extracted from a prototype branch that changed multiple things. I tried to separate as much as possible, but there are still multiple changes here. I'll explain the large changes below and leave more comments inline for other changes.

Note, that we are not merging into `main` but into a temporary branch. All lint errors will be fixed when merging the other branch into `main`.

---

## Streamline `Feature` implementation

The old implementation was quite convoluted. This patch simplifies this by a lot without hurting flexibility. This is achieved by dropping the `like` keyword from the constructor and have a dedicated `.new_like()` method for that.

## Rework `Label` 

Currently, `Label` took a `category` parameter to store the human readable name. As explained in #5045 this has a major downside: when stacking multiple labels like in a batch the category does not need to be the same. This PR changes this to have a `categories` parameter. which is constant for all labels of the same dataset. In addition, this PR adds a `.to_categories()` method that turns a label into the human readable form. 

## Add features for encoded data

This was discussed in https://github.com/pytorch/vision/issues/5075#issuecomment-999664152.


torchvision/prototype/features/_feature.py
```diff
@@ -38,63 +34,28 @@ def __init_subclass__(cls):
...
-    def _to_tensor(cls, data, *, dtype, device):
+    def _to_tensor(self, data: Any, *, dtype, device) -> torch.Tensor:
         return torch.as_tensor(data, dtype=dtype, device=device)
 
     @classmethod
-    def _parse_meta_data(cls) -> Dict[str, Tuple[Any, Any]]:
-        return dict()
-
-    @classmethod
-    def __torch_function__(
```
pmeier, line 85: This will be added back with the transforms rework.

torchvision/prototype/utils/_internal.py
```diff
@@ -186,3 +213,114 @@ def wrapper(*args: Any, **kwargs: Any) -> D:
...
+
+
+def apply_recursively(fn: Callable, obj: Any) -> Any:
+    # We explicitly exclude str's here since they are self-referential and would cause an infinite recursion loop:
+    # "a" == "a"[0][0]...
+    if isinstance(obj, collections.abc.Sequence) and not isinstance(obj, str):
+        sequence: List[Any] = []
+        for item in obj:
+            result = apply_recursively(fn, item)
+            if isinstance(result, collections.abc.Sequence) and hasattr(result, "__inline__"):
```
pmeier, line 311: This makes no sense yet, but it will on a later PR.

NicolasHug (approved):
Thanks @pmeier , I only took a very brief look. As discussed offline I'll approve to unlock and will give it a more in-depth look later

PR #5286, opened 2022-01-26
Add PCAM dataset to prototype area

NicolasHug:


torchvision/prototype/datasets/_builtin/pcam.py
```diff
@@ -0,0 +1,122 @@
...
+            with h5py.File(handle) as data:
+                if self.key is not None:
+                    data = data[self.key]
+                yield from data
+
+class PCAM(Dataset):
+    def _make_info(self) -> DatasetInfo:
+        return DatasetInfo(
+            "pcam",
+            type=DatasetType.RAW,
```
NicolasHug, line 43: The .h5 files store decoded numpy arrays, so I assume `RAW` is the right one here?

torchvision/prototype/datasets/_builtin/pcam.py
```diff
@@ -0,0 +1,122 @@
...
+        datapipe: IterDataPipe[Tuple[str, io.IOBase]],
+        key: Optional[str] = None,  # Note: this key thing might be very specific to the PCAM dataset
+    ) -> None:
+        self.datapipe = datapipe
+        self.key = key
+
+    def __iter__(self) -> Iterator[Tuple[str, io.IOBase]]:
+        import h5py  # noqa
+        for _, handle in self.datapipe:
+            with h5py.File(handle) as data:
```
NicolasHug, line 34: Note: using the file_path (`_`) works just as well here. @pmeier are there any subtleties that would make us choose one over the other?

test/builtin_dataset_mocks.py
```diff
@@ -1312,3 +1312,32 @@ def svhn(info, root, config):
...
+        f["y"] = np.random.randint(0, 2, size=(num_images, 1, 1, 1), dtype=np.uint8)
+
+    # Create .gz compressed files
+    for file_name in (images_file, targets_file):
+        with open(file_name, "rb") as file:
+            compressed_data = gzip.compress(file.read())
+        compressed_file_name = file_name.with_suffix(".h5.gz")
+        with open(compressed_file_name, "wb") as compressed_file:
+            compressed_file.write(compressed_data)
+        file_name.unlink()
```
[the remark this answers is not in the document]
jdsgomes: I think if you want to do everything in memory this pattern would work:

```python
images_io = io.BytesIO()
with h5py.File(images_io, 'w') as f:
    f["x"] = np.random.randint(0, 256, size=(num_images, 10, 10, 3), dtype=np.uint8)
with open(compressed_file_name, "wb") as compressed_file:
    compressed_file.write(gzip.compress(images_io.read()))
```

torchvision/prototype/datasets/_builtin/pcam.py
```diff
@@ -0,0 +1,122 @@
...
+    hint_shuffling,
+)
+from torchvision.prototype.features import Label
+
+
+class H5Reader(IterDataPipe[Tuple[str, io.IOBase]]):
+    def __init__(
+        self,
+        datapipe: IterDataPipe[Tuple[str, io.IOBase]],
+        key: Optional[str] = None,  # Note: this key thing might be very specific to the PCAM dataset
```
pmeier, line 26: Nit: I would reserve line comments for linter directives.

```suggestion
        # Note: this key thing might be very specific to the PCAM dataset
        key: Optional[str] = None,
```

This applies to all occurrences of this.

torchvision/prototype/datasets/_builtin/pcam.py
```diff
@@ -0,0 +1,122 @@
...
+    GDriveResource,
+)
+from torchvision.prototype.datasets.utils._internal import (
+    hint_sharding,
+    hint_shuffling,
+)
+from torchvision.prototype.features import Label
+
+
+class H5Reader(IterDataPipe[Tuple[str, io.IOBase]]):
```
pmeier, line 22: Maybe we can call it 

```suggestion
class PCAMH5Reader(IterDataPipe[Tuple[str, io.IOBase]]):
```

and only make it more general if another dataset also needs this?

torchvision/prototype/datasets/_builtin/pcam.py
```diff
@@ -0,0 +1,122 @@
...
+    def __init__(
+        self,
+        datapipe: IterDataPipe[Tuple[str, io.IOBase]],
+        key: Optional[str] = None,  # Note: this key thing might be very specific to the PCAM dataset
+    ) -> None:
+        self.datapipe = datapipe
+        self.key = key
+
+    def __iter__(self) -> Iterator[Tuple[str, io.IOBase]]:
+        import h5py  # noqa
```
pmeier, line 32: What is `flake8` complaining here? In any case can we make this more specific like `# noqa: F401`.

torchvision/prototype/datasets/_builtin/pcam.py
pmeier: 1. The file is already open so there is no need to open it again. I'm guessing this works on Linux and Mac, but could maybe bite us on Windows.
2. For this dataset the files are present on disk directly. If they live inside an archive, the path is just an indicator and not a real path, like `/foo/bar.zip/baz.h5`. In this case we can only use the handle.

For these reasons I would always go with the handle directly unless there is a reason not to.

torchvision/prototype/datasets/_builtin/pcam.py
```diff
@@ -0,0 +1,122 @@
...
+        return DatasetInfo(
+            "pcam",
+            type=DatasetType.RAW,
+            homepage="https://github.com/basveeling/pcam",
+            categories=2,
+            valid_options=dict(split=("train", "test", "val")),
+            dependencies=["h5py"]
+        )
+
+    _RESOURCES= {
```
pmeier, line 50: This is kinda hard to parse. Maybe we can have small dataclass or namedtuple to carry the information?

torchvision/prototype/datasets/_builtin/pcam.py
```diff
@@ -0,0 +1,122 @@
...
+            for file_name, gdrive_id, sha256 in self._RESOURCES[config.split]
+        ]
+
+    def _collate_and_decode(
+        self, data: Tuple[Any, Any]) -> Dict[str, Any]:
+        image, target = data  # They're both numpy arrays at this point
+
+        return {
+            "image": features.Image(torch.tensor(image)),
+            "label": Label(target.item()),
```
pmeier, line 101: Internally, the custom `Feature`'s call 

https://github.com/pytorch/vision/blob/11d903ec6b8401d072c48d42ab0c3a6e3c8f62c0/torchvision/prototype/features/_feature.py#L76-L78

on the input. Thus, we don't need to convert manually here

```suggestion
            "image": features.Image(image),
            "label": Label(target),
```

torchvision/prototype/datasets/_builtin/pcam.py
```diff
@@ -0,0 +1,122 @@
...
+
+        images_dp = H5Reader(images_dp, key="x")
+        targets_dp = H5Reader(targets_dp, key="y")
+
+        dp = Zipper(images_dp, targets_dp)
+        dp = hint_sharding(dp)
+        dp = hint_shuffling(dp)
+        dp = Mapper(dp, self._collate_and_decode)
+
+        return dp
```
pmeier, line 122: Nit: saw this a couple of times in your PRs. Any advantage of returning directly?

```suggestion
        return Mapper(dp, self._collate_and_decode)
```

pmeier (approved):
Only minor comments inline. Thanks a lot @NicolasHug! LGTM when CI is green.

test/builtin_dataset_mocks.py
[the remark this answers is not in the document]
NicolasHug: Thanks @jdsgomes !
For some reason I had to use `getbuffer()` instead of `read()`.

torchvision/prototype/datasets/_builtin/pcam.py
NicolasHug: It's just more consistent with the rest of the code which uses

`dp = ...`

a lot and it makes it obvious that what it returned is still a "dp" for non-expert readers. But I don't feel strongly about this

torchvision/prototype/datasets/_builtin/pcam.py
NicolasHug: Do you feel strongly about this? I like line comments because it's very clear what they apply to.
In some occasions, I have a hard time figuring out whether a comment applies to a single line below, or to a bigger block

jdsgomes (approved):
LGTM, there are just some errors related to imports, once those are sorted the PR is ready to be merged

torchvision/prototype/datasets/_builtin/pcam.py
pmeier: No strong opinion.

PR #5287, opened 2022-01-26
remove decoding from prototype datasets

pmeier:
Same deal as #5283. Supersedes #5105.

---

- Main change is to use the new encoded data feature types in the prototype datasets. This completely removes the need to pass a `decoder` to `datasets.load`. The decoding will be performed by a transform that will be added later on.
- Doing that the `decoding` part would need to be dropped from the canonical `_collate_and_decode_sample` method. Since "collate" is also used to prepare a batch of data for model consumption, I've opted to change the name to `_prepare_sample`.
- This also fixes the ImageNet validation split. We use a different order of the categories than the dataset for BC reasons. This was not considered in the current implementation. cc @datumbox 
- Add categories for the VOC dataset


test/builtin_dataset_mocks.py
```diff
@@ -728,32 +730,33 @@ def _make_detection_anns_folder(cls, root, name, *, file_name_fn, num_examples):
...
         def add_name(obj, name="dog"):
             add_child(obj, "name", name)
-            return name
 
-        def add_bndbox(obj, bndbox=None):
-            if bndbox is None:
-                bndbox = {"xmin": "1", "xmax": "2", "ymin": "3", "ymax": "4"}
+        def add_size(obj):
+            obj = add_child(obj, "size")
+            size = {"width": 0, "height": 0, "depth": 3}
```
pmeier, line 741: VOC provides the image size together with the annotations. Since the reworked `BoundingBox` requires the image size, we need to add it to the mock data.

test/test_prototype_transforms.py
```diff
@@ -1,61 +0,0 @@
-import pytest
```
pmeier, line 1: Same deal as before. The tests are only partially compatible with the new features in the datasets. Thus, we remove them here and can re-add them when the transforms API is more stable.

torchvision/prototype/datasets/_builtin/coco.py
```diff
@@ -150,41 +147,38 @@ def _classify_meta(self, data: Tuple[str, Any]) -> Optional[int]:
...
         *,
-        annotations: Optional[str],
-        decoder: Optional[Callable[[io.IOBase], torch.Tensor]],
+        annotations: str,
     ) -> Dict[str, Any]:
         ann_data, image_data = data
         anns, image_meta = ann_data
 
-        sample = self._collate_and_decode_image(image_data, decoder=decoder)
-        if annotations:
```
pmeier, line 170: We will never get to this point if `annotations is None`

NicolasHug (approved):
Stamping! Happy to discuss and review in more details down the road

torchvision/prototype/datasets/_builtin/sbd.py
```diff
@@ -31,20 +27,17 @@
...
 class SBD(Dataset):
     def _make_info(self) -> DatasetInfo:
         return DatasetInfo(
             "sbd",
-            type=DatasetType.IMAGE,
             dependencies=("scipy",),
             homepage="http://home.bharathh.info/pubs/codes/SBD/download.html",
             valid_options=dict(
                 split=("train", "val", "train_noval"),
-                boundaries=(True, False),
```
pmeier, line 46: If we already load the mat file that stores the annotations, there is no need to only return only part of the data.

torchvision/prototype/datasets/_builtin/sbd.py
```diff
@@ -138,10 +101,10 @@ def _make_datapipe(
             buffer_size=INFINITE_BUFFER_SIZE,
             drop_none=True,
         )
-
         if config.split == "train_noval":
             split_dp = extra_split_dp
-        split_dp = Filter(split_dp, path_comparator("stem", config.split))
+
+        split_dp = Filter(split_dp, path_comparator("name", f"{config.split}.txt"))
```
pmeier, line 107: Using the name is probably more readable than using the stem.

torchvision/prototype/datasets/utils/_internal.py
```diff
@@ -70,14 +66,6 @@ def read_mat(buffer: io.IOBase, **kwargs: Any) -> Any:
     return sio.loadmat(buffer, **kwargs)
 
 
-def image_buffer_from_array(array: np.ndarray, *, format: str = "png") -> io.BytesIO:
```
pmeier, line 73: This is no longer needed as it always was a crutch to enable custom decoders for datasets that didn't contain encoded images in the first place.

Issue #5245, opened 2022-01-21
Change CPU decoder output frames to use ITU709 colour space


prabhat00155:
GPU decoder outputs frame with ITU709 colour space, whereas CPU decoder uses the default colour space. We need to change the CPU decoder frames to use ITU709 colour space to match the GPU decoder output. 
For verification, we could use pyav like this:
```
av_frames = torch.tensor(av_frame.to_rgb(src_colorspace="ITU709").to_ndarray())
```
and compare the CPU decoded frames against `av_frames`.

PR #5291, opened 2022-01-26
Default colorspace for CPU decoder is now ITU709

bjuncek:
Addressing #5245

### DIFF
--- a/torchvision/csrc/io/decoder/video_sampler.cpp
+++ b/torchvision/csrc/io/decoder/video_sampler.cpp
@@ -3,7 +3,6 @@
 #include "util.h"
 
 // www.ffmpeg.org/doxygen/0.5/swscale-example_8c-source.html
-
 namespace ffmpeg {
 
 namespace {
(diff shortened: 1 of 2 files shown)


torchvision/csrc/io/decoder/video_sampler.cpp
```diff
@@ -65,6 +64,48 @@ int transformImage(
...
+             dst_range,
+             brightness,
+             contrast,
+             saturation)) >= 0) {
+      LOG(INFO) << "Successfuly set new colourspace info";
+    } else {
+      LOG(ERROR) << "Failed to set new colourspace info";
+    }
+  } else {
+    LOG(ERROR) << "Failed to get new colourspace info";
```
prabhat00155, line 106: In case of failure, we continue with the default colour space right? If that's the case, should we also explicitly add that message?

torchvision/csrc/io/decoder/video_sampler.cpp
```diff
@@ -65,6 +64,48 @@ int transformImage(
...
+  if ((ret_get = sws_getColorspaceDetails(
+           context,
+           (int**)&inv_tbl,
+           &src_range,
+           (int**)&tbl,
+           &dst_range,
+           &brightness,
+           &contrast,
+           &saturation)) >= 0) {
+    LOG(INFO) << "Successfuly got the information about colourspace \n "
```
prabhat00155, line 87: Is there a "DEBUG" log level? Otherwise, this could end up printing a lot of unnecessary messages.

torchvision/csrc/io/decoder/video_sampler.cpp
```diff
@@ -65,6 +64,48 @@ int transformImage(
   if ((result = preparePlanes(outFormat, out, planes, lines)) < 0) {
     return result;
   }
+
+  // Here we're changing colourspace in order to maintain consistency with the
+  // GPU decoder Since SWSContext cannot be edited directly, we're borrowing
+  // pyAV approach from:
+  // https://github.com/PyAV-Org/PyAV/blob/9ac05d9ac902d71ecb2fe80f04dcae454008378c/av/video/reformatter.pyx#L141
+  // very sory for this, found no cleaner way of doing it
```
prabhat00155, line 72: I think this is just an FYI, perhaps we don't need the last statement to be checked-in.

prabhat00155 (commented):
Thanks @bjuncek, this looks good to me. I have a few minor comments for you to consider.

prabhat00155:
@bjuncek Also, could you fix the linter error please?

prabhat00155:
Internal tests: D33830672

prabhat00155 (changes requested):
Some internal tests are failing, blocking merge for now.

bjuncek (commented):
I think I've addressed most of the concerns, and have added additional coverage for `video_reader` tests

bjuncek:
@prabhat00155 @datumbox any updates on this?

datumbox:
@bjuncek I don't have any context about this PR. Not sure if @prabhat00155 or @fmassa can help here.

BTW there is a conflict in one of the test files, I assume you will take care of it after you get the OK on principle for the direction of the PR?

prabhat00155:
@datumbox This PR would need to be imported into Phabricator to make sure all the internals tests are fine, before merging here, as we are changing the colour space for CPU decoder, which means the output values would be different. Last time, I saw some internal test failures on importing this(D33830672).

PR #5295, opened 2022-01-27
readd functional transforms API to prototype

pmeier:
This re-adds the functional API to the prototype transforms. They are not yet compatible with the transforms. That will come in a follow-up PR.

Design principles:

- Every kernel has to be `jit.scriptable`
- `Pillow` is no longer supported. Note this only applies to the low level kernels. There will be another layer on top, i.e. a higher level functional API, that will preserve BC.
- Every kernel needs to work on vanilla tensors
- Kernels can make assumptions on the tensor data, i.e. requiring bounding boxes in XYXY format
- Kernels that implement the same transform for different feature types should use the transform name suffixed by the feature type in snake case, e.g. `horizontal_flip_image` and `horizontal_flip_bounding_box`.
- Kernels that implement the same transform for different feature types should use the same parameters as much as possible.

In the current state, this PR only adds a few non-image kernels and imports most image kernels from the stable implementation.


torchvision/prototype/transforms/functional/utils.py
```diff
@@ -0,0 +1,13 @@
+import types
+
+
+def _from_legacy_kernel(legacy_kernel, new_name=None):
```
pmeier, line 4: This is basically copying a function and changing its name. This is `jit.script`'able.

NicolasHug: Could you explain why we need this helper?

For example, apart from ending up with different objects, what's the difference between

```py
center_crop_image = _from_legacy_kernel(_F.center_crop)
```
and
```py
center_crop_image = _F.center_crop
```
?

test/test_prototype_transforms_functional.py
```diff
@@ -0,0 +1,197 @@
...
+    @pytest.mark.parametrize("kernel_info", KERNEL_INFOS, ids=lambda kernel_info: kernel_info.name)
+    def test_scriptable(self, kernel_info):
+        jit.script(kernel_info.kernel)
+
+    @pytest.mark.parametrize(
+        ("kernel_info", "sample_input"),
+        [
+            pytest.param(kernel_info, sample_input, id=f"{kernel_info.name}-{idx}")
+            for kernel_info in KERNEL_INFOS
+            for idx, sample_input in enumerate(kernel_info.sample_inputs())
```
NicolasHug, line 190: instead of the 2 manual `for` loops could we simply use 2 `@parametrize` statements?

Also, is there stark difference between passing `idx` instead of relying on pytest's default ids?

test/test_prototype_transforms_functional.py
```diff
@@ -0,0 +1,197 @@
...
+    extra_dims=((4,), (2, 3)),
+):
+    for format, image_size, dtype in itertools.product(formats, image_sizes, dtypes):
+        yield make_bounding_box(format=format, image_size=image_size, dtype=dtype)
+
+    for format, extra_dims_ in itertools.product(formats, extra_dims):
+        yield make_bounding_box(format=format, extra_dims=extra_dims_)
+
+
+class SampleInput:
```
NicolasHug, line 109: Could you explain why we need this class?

NicolasHug (commented):
Thanks @pmeier , only took a very brief look

torchvision/prototype/features/_bounding_box.py
```diff
@@ -36,3 +36,14 @@ def __new__(
         bounding_box._metadata.update(dict(format=format, image_size=image_size))
 
         return bounding_box
+
+    def to_format(self, format: Union[str, BoundingBoxFormat]) -> "BoundingBox":
+        # import at runtime to avoid cyclic imports
+        from torchvision.prototype.transforms.functional import convert_bounding_box_format
```
vfdev-5, line 42: Maybe, this is a sign to redesign the structure and put this method here or in `utils` module ?

torchvision/prototype/transforms/functional/utils.py
pmeier: The difference is that in the former `center_crop_image.__name__ == "center_crop_image"`, while in the latter `center_crop_image.__name__ == "center_crop`. 

I've checked, I don't use the `.__name__` attribute for anything anymore, so this is safe to remove.

test/test_prototype_transforms_functional.py
pmeier: > instead of the 2 manual `for` loops could we simply use 2 `@parametrize` statements?

We can't use two separate parametrizations, since the inner loop depends on the outer.

> Also, is there stark difference between passing idx instead of relying on pytest's default ids?

Let's look into this more after https://github.com/pytorch/vision/pull/5295#discussion_r793614843.

test/test_prototype_transforms_functional.py
pmeier: The other option would be for a simple inputs function to return `((...), dict(...))` to bundle `args` and `kwargs`. IMO this is not as convenient as having a single structure to hold everything. To me

```py
yield SampleInput(bounding_box, old_image_size=bounding_box.image_size, new_image_size=new_image_size)
```

is more readable than

```py
yield (bounding_box,), dict(old_image_size=bounding_box.image_size, new_image_size=new_image_size)
```

given that it resembles the actual call signature.

torchvision/prototype/features/_bounding_box.py
pmeier: It was a design choice to have _all_ transforming functions in `torchvision.prototype.transforms.functional`. This is not implemented yet, but for the automatic dispatch to work, we need to depend on `torchvision.prototype.features`. Thus, if we want to honor the design choice, we can't get around the cyclic imports if we want to provide these convenience conversion methods.

If we relax the design choice to all transforming functions need to be present in `torchvision.prototype.transforms.functional`, but can be imported from somewhere else, I fully agree, and this should be refactored.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -0,0 +1,88 @@
...
+    ).unbind(-1)
+    x = image_size[1] - (x + w)
+    return convert_bounding_box_format(
+        torch.stack((x, y, w, h), dim=-1),
+        old_format=BoundingBoxFormat.XYWH,
+        new_format=BoundingBoxFormat.XYXY,
+    )
+
+
+_resize_image = _from_legacy_kernel(_FT.resize)
```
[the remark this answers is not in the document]
pmeier: Remnant from an older version of the implementation. I'll remove.

test/test_prototype_transforms_functional.py
pmeier: Plus, I was looking into lazily loading the samples. This is not implemented yet, so we are generating all samples at test collection time. This can become an issue real quick if we go along this automated tests direction.

NicolasHug (approved):
Thanks @pmeier

PR #5296, opened 2022-01-27
Switch torch.quantization to torch.ao.quantization

datumbox:
Related to #4554

Adding `ao` to MobileNetV3 prototype which was not originally changed.


NicolasHug (approved):
Thanks @datumbox

PR #5298, opened 2022-01-27
Temporarily use IntermediateLayerGetter on segmentation

datumbox:
I propose to revert #4563 and temporarily switch to IntermediateLayerGetter prior the upcoming feature cut. This will give us a bit more time to test PyTorch FX and ensure that both Detection and Segmentation use the same solution. There is already a ticket for the latter at #4540, so we can keep it open to keep track of the work.


fmassa (approved):
Fine with me!

PR #5301, opened 2022-01-27
fix annotations for Python >= 3.8

pmeier:


setup.py
```diff
@@ -58,6 +58,7 @@ def write_version_file():
     pytorch_dep += "==" + os.getenv("PYTORCH_VERSION")
 
 requirements = [
+    "typing_extensions",
```
pmeier, line 61: PyTorch core already hard-depends on it: https://github.com/pytorch/pytorch/blob/be2dc8f2940d3c95941516a811be8c504910d1ea/setup.py#L408-L411

pmeier (commented):
These changes are necessary when linting against `numpy>=1.22`, which is the latest version for Python >= 3.8. This was not caught by our CI, because we run 3.7 there.

torchvision/transforms/functional_pil.py
```diff
@@ -130,7 +131,7 @@ def pad(
     img: Image.Image,
     padding: Union[int, List[int], Tuple[int, ...]],
     fill: Optional[Union[float, List[float], Tuple[float, ...]]] = 0,
-    padding_mode: str = "constant",
+    padding_mode: Literal["constant", "edge", "reflect", "symmetric"] = "constant",
```
datumbox, line 134: That's a new one for me, `Literally`.

datumbox (approved):
LGTM!

torchvision/datasets/stl10.py
```diff
@@ -65,10 +65,12 @@ def __init__(
         self.labels: Optional[np.ndarray]
         if self.split == "train":
             self.data, self.labels = self.__loadfile(self.train_list[0][0], self.train_list[1][0])
+            self.labels = cast(np.ndarray, self.labels)
```
NicolasHug, line 68: We already have `self.labels: Optional[np.ndarray]` above, should we remove one of these?

pmeier: I missed that. Let me have a look.

PR #5302, opened 2022-01-27
Expose generalized_box_iou to docs

oke-aditya:
Btw where are the tests for this loss function and also for sigmoid focal loss?

Also do you think it is worth to have a canonical `box_iou_loss` function?


datumbox (approved):
LGTM, thanks!

PR #5308, opened 2022-01-28
Add support for flow batches in flow_to_image

NicolasHug:
Follow up to https://github.com/pytorch/vision/pull/5134 and to https://github.com/pytorch/vision/issues/5091, this PR adds support for batches of flows to the `flow_to_image()` util.

CC @oke-aditya

### DIFF
--- a/torchvision/utils.py
+++ b/torchvision/utils.py
@@ -397,42 +397,51 @@ def flow_to_image(flow: torch.Tensor) -> torch.Tensor:
     Converts a flow to an RGB image.
 
     Args:
-        flow (Tensor): Flow of shape (2, H, W) and dtype torch.float.
+        flow (Tensor): Flow of shape (N, 2, H, W) or (2, H, W) and dtype torch.float.
 
     Returns:
-        img (Tensor(3, H, W)): Image Tensor of dtype uint8 where each color corresponds to a given flow direction.
+        img (Tensor): Image Tensor of dtype uint8 where each color corresponds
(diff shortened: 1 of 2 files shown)


datumbox (approved):
LGTM, thanks!

torchvision/utils.py
```diff
@@ -397,42 +397,51 @@ def flow_to_image(flow: torch.Tensor) -> torch.Tensor:
...
     """
 
     if flow.dtype != torch.float:
         raise ValueError(f"Flow should be of dtype torch.float, got {flow.dtype}.")
 
-    if flow.ndim != 3 or flow.size(0) != 2:
-        raise ValueError(f"Input flow should have shape (2, H, W), got {flow.shape}.")
+    orig_shape = flow.shape
+    if flow.ndim == 3:
+        flow = flow[None]  # Add batch dim
```
oke-aditya, line 412: One small doubt.

Does this modify flow inplace?

NicolasHug: No, the 2 objects may be a view on the same underlying memory, but they're different objects


```py
In [1]: import torch
torch.tensor
In [2]: a = torch.tensor([1, 2])

In [3]: b = a[None]

In [4]: b is a
Out[4]: False

In [5]: a, b
Out[5]: (tensor([1, 2]), tensor([[1, 2]]))
```

oke-aditya: Thanks for clarification. Will note this.

PR #5310, opened 2022-01-28
Add topk min function for trace and onnx

shubhambhokare1:
ONNX spec requires the k-value to be less than or equal to the number of inputs along provided dim. Certain models use the number of elements along a particular axis instead of K if K exceeds the number of elements along that axis. Previously, python's min() function was used to determine whether to use the provided k-value or the specified dim axis value.

However in cases where the model is being exported in tracing mode, python min() is static causing the model to be traced incorrectly and eventually fail at the topk node. In order to avoid this situation, in tracing mode, torch.min() is used instead.

### DIFF
--- a/torchvision/models/detection/_utils.py
+++ b/torchvision/models/detection/_utils.py
@@ -1,6 +1,6 @@
 import math
 from collections import OrderedDict
-from typing import List, Tuple, cast
+from typing import List, Tuple
 
 import torch
 from torch import Tensor, nn
@@ -491,6 +491,4 @@ def _topk_min(input: Tensor, orig_kval: int, axis: int) -> int:
     """
     axis_dim_val = torch._shape_as_tensor(input)[axis].unsqueeze(0)
     min_kval = torch.min(torch.cat((torch.tensor([orig_kval], dtype=axis_dim_val.dtype), axis_dim_val), 0))
-
-    # for mypy we cast at runtime
-    return cast(int, min_kval)
+    return min_kval  # type: ignore[arg-type]
--- a/torchvision/models/detection/_utils.py
+++ b/torchvision/models/detection/_utils.py
@@ -470,7 +470,7 @@ def retrieve_out_channels(model: nn.Module, size: Tuple[int, int]) -> List[int]:
     return out_channels
 
 
-def _topk_min(input: Tensor, orig_kval: int, axis: int) -> int:
+def _topk_min(input: Tensor, orig_kval: int, axis: int) -> Tensor:
     """
     ONNX spec requires the k-value to be less than or equal to the number of inputs along
     provided dim. Certain models use the number of elements along a particular axis instead of K
@@ -487,7 +487,7 @@ def _topk_min(input: Tensor, orig_kval: int, axis: int) -> int:
         axis(int): Axis along which we retreive the input size.
 
     Returns:
-        min_kval (int): Appropriately selected k-value.
+        min_kval (Tensor): Appropriately selected k-value.
     """
     axis_dim_val = torch._shape_as_tensor(input)[axis].unsqueeze(0)
     min_kval = torch.min(torch.cat((torch.tensor([orig_kval], dtype=axis_dim_val.dtype), axis_dim_val), 0))
--- a/torchvision/models/detection/_utils.py
+++ b/torchvision/models/detection/_utils.py
@@ -470,12 +470,12 @@ def retrieve_out_channels(model: nn.Module, size: Tuple[int, int]) -> List[int]:
     return out_channels
 
 
-@torch.jit.unused
-def _onnx_tracing_topk_min(input: Tensor, orig_kval: int, axis: int) -> Tuple[int, int]:
+def _topk_min(input: Tensor, orig_kval: int, axis: int) -> int:
     """
     ONNX spec requires the k-value to be less than or equal to the number of inputs along
-    provided dim. Certain models uses python's min() function to determine whether to use
-    the provided k-value or the specified dim axis value.
+    provided dim. Certain models use the number of elements along a particular axis instead of K
+    if K exceeds the number of elements along that axis. Previously, python's min() function was
+    used to determine whether to use the provided k-value or the specified dim axis value.
 
     However in cases where the model is being exported in tracing mode, python min() is
     static causing the model to be traced incorrectly and eventually fail at the topk node.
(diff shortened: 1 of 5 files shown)


torchvision/models/detection/fcos.py
```diff
@@ -501,7 +502,10 @@ def postprocess_detections(
                 topk_idxs = torch.where(keep_idxs)[0]
 
                 # keep only topk scoring predictions
-                num_topk = min(self.topk_candidates, topk_idxs.size(0))
+                if torchvision._is_tracing():
+                    _, num_topk = det_utils._onnx_tracing_topk_min(topk_idxs, self.topk_candidates, 0)
```
garymm, line 506: Does this provide the wrong answer for non-tracing mode?
If not, why do we need the if-statement?

datumbox:
@shubhambhokare1 Thanks for the contribution. TorchVision's detection models do not properly support tracing. Instead we recommend users to JIT-script them. 

@fmassa @prabhat00155 Could you weight in on this one as you are more knowledgeable on ONNX? I also find it strange that despite not properly supporting tracing, our RPN [code](https://github.com/pytorch/vision/blob/d367a01a18a3ae6bee13d8be3b63fd6a581ea46f/torchvision/models/detection/rpn.py#L17-L24) has some kind of mitigation about it. Any thoughts on why this is the case?

torchvision/models/detection/_utils.py
```diff
@@ -468,3 +468,31 @@ def retrieve_out_channels(model: nn.Module, size: Tuple[int, int]) -> List[int]:
...
 
     return out_channels
+
+
+@torch.jit.unused
+def _onnx_tracing_topk_min(input: Tensor, orig_kval: int, axis: int) -> Tuple[int, int]:
+    """
+    ONNX spec requires the k-value to be less than or equal to the number of inputs along
+    provided dim. Certain models uses python's min() function to determine whether to use
+    the provided k-value or the specified dim axis value.
```
prabhat00155, line 478: > "Certain models uses python's min() function to determine whether to use the provided k-value or the specified dim axis value."?

This statement is not very clear. I think you are trying to say that certain models use the number of elements along a particular `axis` instead of `K` if `K` exceeds the number of elements along that axis?

torchvision/models/detection/_utils.py
```diff
@@ -468,3 +468,31 @@ def retrieve_out_channels(model: nn.Module, size: Tuple[int, int]) -> List[int]:
...
+    Returns:
+        (Tuple[int, int]): Input size along axis, appropriately selected k-value.
+    """
+    from torch.onnx import operators
+
+    axis_dim_val = operators.shape_as_tensor(input)[axis].unsqueeze(0)
+    min_kval = torch.min(torch.cat((torch.tensor([orig_kval], dtype=axis_dim_val.dtype), axis_dim_val), 0))
+
+    # for mypy we cast at runtime
+    return cast(int, axis_dim_val), cast(int, min_kval)
```
prabhat00155, line 498: Casting a tensor to an int doesn't look very clean to me. Can't you use [item](https://pytorch.org/docs/stable/generated/torch.Tensor.item.html)?

prabhat00155:
Thanks @shubhambhokare1 for the PR, I have a couple of comments.

torchvision/models/detection/_utils.py
zhiqwang: Hi @prabhat00155 , 

Seems that the `item()` will cause something wrong in the tracing mode, and there is a similar workaround in the detection transformation:

https://github.com/pytorch/vision/blob/44e93beb0833ff51ffe75d994fe660434335b53d/torchvision/models/detection/transform.py#L47-L50

shubhambhokare1: Yup, as @zhiqwang mentioned, tensor.item() is replaced with a constant during tracing.
This is a snippet from https://pytorch.org/docs/stable/onnx.html that addresses this limitation.
```And rather than using [torch.Tensor.item()] (which converts a Tensor to a Python built-in number):
# Bad! y.item() will be replaced with a constant during tracing.
```

torchvision/models/detection/_utils.py
shubhambhokare1: Yup, you are right. Will change the wording to reflect this.

torchvision/models/detection/fcos.py
shubhambhokare1: I don't see why this change shouldn't work for both modes. 
@prabhat00155 Could you point me to the appropriate tests for detection models in tracing and scripting mode to test the above suggestion?

zhiqwang: Hi @shubhambhokare1 , Seems that torchvion now only uses tracing mode when exporting ONNX, I guess the following tests might be suitable for reference

https://github.com/pytorch/vision/blob/97eddc5d6a83a9bf620070075ef1e1864c9a68ac/test/test_onnx.py#L430-L444
https://github.com/pytorch/vision/blob/97eddc5d6a83a9bf620070075ef1e1864c9a68ac/test/test_onnx.py#L372-L386

datumbox (changes requested):
The PR is toasted, possibly due to a bad synch. Worth force pushing to an earlier version or closing and opening an new one.

shubhambhokare1:
> The PR is toasted, possibly due to a bad synch. Worth force pushing to an earlier version or closing and opening an new one.

I think it should be fixed now

shubhambhokare1:
> The PR is toasted, possibly due to a bad synch. Worth force pushing to an earlier version or closing and opening an new one.

Think it should be fixed now

torchvision/models/detection/_utils.py
```diff
@@ -468,3 +468,29 @@ def retrieve_out_channels(model: nn.Module, size: Tuple[int, int]) -> List[int]:
         model.train()
 
     return out_channels
+
+
+def _topk_min(input: Tensor, orig_kval: int, axis: int) -> Tuple[int, int]:
```
garymm, line 473:
```suggestion
def _topk_min(input: Tensor, orig_kval: int, axis: int) -> int:
```

garymm (changes requested):

torchvision/models/detection/fcos.py
shubhambhokare1: Seems like the call
```
cast(int, min_kval)
```
leads to the error
```
fn = torch._C._jit_script_compile(
RuntimeError: 
__int__ cannot be used as a value:
```
This might be a torchscript limitation, reverting back to different branches of scripting and tracing for now

prabhat00155 (approved):
Thanks @shubhambhokare1!

torchvision/models/detection/_utils.py
```diff
@@ -468,3 +468,27 @@ def retrieve_out_channels(model: nn.Module, size: Tuple[int, int]) -> List[int]:
...
+        input (Tensor): The orignal input tensor.
+        orig_kval (int): The provided k-value.
+        axis(int): Axis along which we retreive the input size.
+
+    Returns:
+        min_kval (Tensor): Appropriately selected k-value.
+    """
+    axis_dim_val = torch._shape_as_tensor(input)[axis].unsqueeze(0)
+    min_kval = torch.min(torch.cat((torch.tensor([orig_kval], dtype=axis_dim_val.dtype), axis_dim_val), 0))
+    return min_kval  # type: ignore[arg-type]
```
datumbox, line 494: @prabhat00155 @shubhambhokare1  There might be performance implications here that we need to benchmark thoroughly. If we have such benchmarks we should post them in this PR to make them easily accessible.

If we don't I propose either to revert the PR until we measure it. Another alternative is to do the onnx friendly approach only when we are tracing and do the previous simpler estimation in all other cases.

PR #5315, opened 2022-01-29
Model contrib guidelines

jdsgomes:
New model contribution guidelines that aim to clarify the process for adding new model architectures or improved model weights.

Feedback welcome, as always, before it gets merged.

cc: @datumbox  @xiaohu2015 @zhiqwang


jdsgomes:
@xiaohu2015 @zhiqwang since you have been through this process recently your feedback would be very valuable :)

zhiqwang (commented):
Hi @jdsgomes , This guidelines seem looks very useful to me. I think one thing needs to be supplemented is that the added model needs to provide at least torchscript export support (especially the `torch.jit.script` mode as torchvision is currently using).

CONTRIBUTING.md
```diff
@@ -186,10 +186,9 @@ You can also choose to only build a subset of the examples by using the
 example ``EXAMPLES_PATTERN="transforms" make html`` will only build the examples
 with "transforms" in their name.
 
-### New model
+### New or improved model
```
datumbox, line 189: Perhaps be explicit on the improvement we mean here?
```suggestion
### New architecture or improved model weights
```

CONTRIBUTING_MODELS.md
```diff
@@ -0,0 +1,120 @@
...
+### 4. Submit a PR
+
+Submit a PR and tag the assigned maintainer. This PR should:
+
+- Link the original ticket
+- Provide a link for the original paper and the original repository if available
+- Highlight the important test metrics and how they compare to the original paper
+- Highlight any design choices that deviate from the original paper/implementation and rationale for these choices
+
+## New Model Architectures - Implementation Details
```
datumbox, line 56: We've discussed with Joao the possibility of moving this section out of the guideline doc (perhaps to an issue or a wiki article?). This is because it's quite low-level and potentially subjective to regular change. I would like to hear the thoughts of the community on this.

zhiqwang: I vote for issue because it allows for further discussion.

jdsgomes: agree, moved to a new issue for now, if we decide to move it back later that's ok: https://github.com/pytorch/vision/issues/5319

datumbox (approved):
LGTM, though I wouldn't merge before the end of the week to see if there is more feedback. 

Might be worth posting this link on previous issues, such as #2707 to let people know we work on a new contribution guideline.

CONTRIBUTING_MODELS.md
datumbox: Looks good. Worth adding on the new issue an explicit note about JIT-scriptability as @zhiqwang said. Going forwards we also expect all models to have scriptable preprocessing transforms as part of the multi-weight API, though I would probably omit this until the API graduates from prototype.

PR #5316, opened 2022-01-29
Add example for optical flow visualizaition and RAFT

### LINKED ISSUE #5309: Sphinx-gallery example for RAFT  model
Now that we have a `flow_to_image()` util https://github.com/pytorch/vision/issues/5091, we can write a sphinx-gallery example to illustrate how to use the RAFT model.

Ideally, the example would look like something like this:

- Briefly explain what optical flow is: predicting movement from 2 consecutive frames
- load the model with pre-trained weights
- load a video (from our `assets` folder), or download it. This is TBD. If we host the video on our GitHub repo we need to make sure it's free of use **and** super lightweight, so it's unlikely. One potential option might be to host the video on our S3 bucket where we host the pre-trained weighst?
- read frames from this video with `read_video()`. Ideally we would be using `VideoReader()`, but it is only available to users who build torchvision from source, so it may not be usable by most. Note: we don't want to rely on external dependencies to read the video
- pass consecutive frames to the model, get predictions, and convert the predicted flows to images via `flow_to_image()`
- plot a few of the predictions for illustration
- save the predicted images to a folder and tell the user they can create a video with e.g. `ffmpeg -f image2 -framerate 30 -i frame_%d.jpg -loop -1 my_cool_gif.gif`

A good starting point is https://github.com/pytorch/vision/issues/5091#issuecomment-992677348

Some caveats worth noting:
- we need the example to be reasonably fast to run, so that means the image resolution should not be too big. Ideally it should take < 1 min on the CI (which does not have a GPU!)
- If loading a video is impossible (or hard), we can fallback to just using pre-selected pairs of images. The example won't be as fancy, but still interesting, and we can expand later.
- loading the weights from `torchvision.models.optical_flow` will give pre-trained weights from Chairs + Things. These may not be the most accurate weights, depending on the type of video that we apply the model to. The most accurate weights would probably be e.g. `Raft_Large_Weights.C_T_SKHT_V2` or `C_T_SKHT_K_V2`, but these are only available through the prototype API in `torchvision.prototype.models.optical_flow`. Whether we can already use this API in a sphinx-gallery example is TBD: perhaps @datumbox can share his thoughts here? In any case it's safe to start writing the eaxmple with the basic pre-trained weights from `torchvision.models.optical_flow`.

As discussed offline @oke-aditya is interested in this so I'll assign it to you :)

oke-aditya:
Closes #5309

Thanks a lot to @NicolasHug for taking it over.


gallery/plot_optical_flow.py
```diff
@@ -0,0 +1,105 @@
...
+        axs[0, i].imshow(np.asarray(img))
+        axs[0, i].set(xticklabels=[], yticklabels=[], xticks=[], yticks=[])
+
+
+####################################
+# Reading Videos Using Torchvision
+# --------------------------------
+# We will first read a video using torchvision's read_video.
+# Alternatively once can use the new VideoReader API (if torchvision is built from source) or
+# open-cv VideoCapture method.
```
NicolasHug, line 41:
```suggestion
# Alternatively one can use the new VideoReader API (if torchvision is built from source).
```

gallery/plot_optical_flow.py
```diff
@@ -0,0 +1,105 @@
...
+video_path = img_path = os.path.join(ASSETS_DIRECTORY, "./basketball.mp4")
+
+#########################
+# Read video returns the video_frames, audio_frames and the metadata
+# We will focus only on the video frames.
+# Note that v_frames tensor is of shape ``(num_frames, height, width, num_channels)``.
+# We will visualize the optical flow between the first two frames of the video.
+
+
+v_frames, a_frames, meta_data = read_video(video_path)
```
NicolasHug, line 54: Since we don't use audio nor metadata:

```suggestion
frames, _, _ = read_video(video_path)
```

gallery/plot_optical_flow.py
```diff
@@ -0,0 +1,105 @@
...
+# Estimating Optical flow using RAFT
+# ----------------------------------
+# Some text
+#
+
+from torchvision.models.optical_flow import raft_large
+
+model = raft_large(pretrained=True, progress=False)
+model = model.eval()
+flow_preds = model(2 * frame_1 / 255 - 1, 2 * frame_2 / 255 - 1)
```
NicolasHug, line 91: Let's include these transformations as part of the `frame_preprocess()` function instead. Also I'd suggest using the same transforms that we use in our reference scripts:

https://github.com/pytorch/vision/blob/4ae20e53a016b5d48ed8b4619499be3beb4b44f3/references/optical_flow/presets.py#L12-L13

gallery/plot_optical_flow.py
```diff
@@ -0,0 +1,105 @@
...
+# Visualizing optical flow
+# -------------------------
+# Torchvision provides ``flow_to_image`` utlity to visualize optical flow.
+# It can be used to convert single images as well as batches to flow.
+
+
+from torchvision.utils import flow_to_image
+img = flow_to_image(flow)
+
+img = img.squeeze(0)
```
NicolasHug, line 104: It'd be nice to show the flow image and frame_1 side by side :)

gallery/plot_optical_flow.py
```diff
@@ -0,0 +1,105 @@
...
+==============================
+
+Optical flow is a task consisting of estimating per pixel motion between two consecutive frames of a video.
+We aim to find the displacement of all image pixels and calculate their motion vectors.
+The following example illustrates how torchvision can be used in predicting as well as
+visualizing optical flow.
+
+"""
+
+# sphinx_gallery_thumbnail_path = "../../gallery/assets/optical_flow_thumbnail.png"
```
NicolasHug, line 13: We might need to remove the thumbnail, I'm not sure we have the rights of the `optical_flow_thumbnail.png ` image that was uploaded. But I'm sure we can use on of the images that this example will generate!

NicolasHug (commented):
Thanks @oke-aditya this looks great so far! Made a quick review as discussed offline

gallery/plot_optical_flow.py
oke-aditya: Exactly wanted to do that :( Faced some errors, will try again.

gallery/plot_optical_flow.py
oke-aditya: Yes it's pending. Haven't worked much on this yet, will try to complete over this week.

oke-aditya:
Not much well over the weekend :cry: will try to finish this in few days

NicolasHug:
@oke-aditya thank you so much for your work so far.

I took care of polishing the example a bit, here are the latest rendered docs: https://1177465-73328905-gh.circle-artifacts.com/0/docs/auto_examples/plot_optical_flow.html#sphx-glr-auto-examples-plot-optical-flow-py

LMK what you think of it!

gallery/plot_optical_flow.py
```diff
@@ -0,0 +1,197 @@
...
+# <https://www.pexels.com/video/a-man-playing-a-game-of-basketball-5192157/>`_,
+# credits go to `Pavel Danilyuk <https://www.pexels.com/@pavel-danilyuk>`_.
+
+
+import tempfile
+from pathlib import Path
+from urllib.request import urlretrieve
+
+
+video_url = "https://download.pytorch.org/tutorial/pixelscom_pavel_danilyuk_basketball_hd.mp4"
```
datumbox, line 62: Shall I correct the typo here? Let me know to do it on S3
```suggestion
video_url = "https://download.pytorch.org/tutorial/pexelscom_pavel_danilyuk_basketball_hd.mp4"
```

gallery/plot_optical_flow.py
```diff
@@ -0,0 +1,197 @@
...
+from torchvision.utils import flow_to_image
+
+flow_imgs = flow_to_image(predicted_flows)
+
+# The images have been mapped into [-1, 1] but for plotting we want them in [0, 1]
+img1_batch = [(img1 + 1) / 2 for img1 in img1_batch]
+
+grid = [[img1, flow_img] for (img1, flow_img) in zip(img1_batch, flow_imgs)]
+plot(grid)
+
```
oke-aditya, line 168: It Will be nice to describe the outputs a bit, so that new readers understand "what" we are visualizing.
E.g. 
In the two consecutive frames we see that the basketball net and the trees haven't moved. Whereas the player is jumping and hence we see more vibrant colors which reflect motion or optical flow in the frames.

P.S. please frame in better words. My point is beginners don't understand what is being visualized or interpret the visual meaning of the output.

gallery/plot_optical_flow.py
datumbox: Done one S3.

gallery/plot_optical_flow.py
```diff
@@ -0,0 +1,197 @@
...
+####################################
+# Bonus: creating GIFs of predicted flows
```
oke-aditya, line 170:
```suggestion
# Bonus: Creating GIFs of predicted flows
```

gallery/plot_optical_flow.py
```diff
@@ -0,0 +1,197 @@
...
+img2_batch = torch.stack([frames[101], frames[151]])
+
+plot(img1_batch)
+
+#########################
+# The RAFT model that we will use accepts RGB float images with pixel values in
+# [-1, 1]. The frames we got from :func:`~torchvision.io.read_video` are int
+# images with values in [0, 255], so we will have to pre-process them. We also
+# reduce the image sizes for the example to run faster. Image dimension must be
+# divisible by 8.
```
oke-aditya, line 89: Why do the image dimensions need to be divisble by 8?

oke-aditya (commented):
Looks fantastic, awesome. Few comments :heart:

gallery/plot_optical_flow.py
```diff
@@ -0,0 +1,197 @@
...
+
+import numpy as np
+import torch
+import matplotlib.pyplot as plt
+import torchvision.transforms.functional as F
+import torchvision.transforms as T
+
+
+plt.rcParams["savefig.bbox"] = "tight"
+# sphinx_gallery_thumbnail_number = 2
```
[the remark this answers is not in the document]
NicolasHug: Yes this tells sphinx-gallery to use the second image as the thumbnail instead of the first one: https://1177465-73328905-gh.circle-artifacts.com/0/docs/generated/torchvision.models.optical_flow.raft_large.html#torchvision.models.optical_flow.raft_large

gallery/plot_optical_flow.py
NicolasHug: It's a hardcoded constraint within the model, the feature extractor downsamples the images by 8

gallery/plot_optical_flow.py
```diff
@@ -0,0 +1,197 @@
...
+print(f"dtype = {predicted_flows.dtype}")
+print(f"shape = {predicted_flows.shape} = (N, 2, H, W)")
+print(f"min = {predicted_flows.min()}, max = {predicted_flows.max()}")
+
+
+####################################
+# Visualizing predicted flows
+# ---------------------------
+# Torchvision provides the :func:`~torchvision.utils.flow_to_image` utlity to
+# convert a flow into an RGB image. It also supports batches of flows.
```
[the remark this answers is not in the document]
NicolasHug: All flows that uses manipulate are non-normalized as we explain above, so IMO it's not necessary to repeat that. It could actually lead to confusion because it may suggest that other utils work with normalized flows.

gallery/plot_optical_flow.py
NicolasHug: I'll write something above

gallery/plot_optical_flow.py
```diff
@@ -0,0 +1,197 @@
...
+#     flow_img = flow_to_image(predicted_flow).to("cpu")
+#     output_folder = "/tmp/"  # Update this to the folder of your choice
+#     write_jpeg(flow_img, output_folder + f"predicted_flow_{i}.jpg")
+
+####################################
+# Once the .jpg flow images are saved, you can convert them into a video or a
+# GIF using ffmpeg with e.g.:
+#
+# `
+# ffmpeg -f image2 -framerate 30 -i predicted_flow_%d.jpg -loop -1 flow.gif
```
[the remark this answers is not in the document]
NicolasHug: I'm not sure sphinx can. It's a good idea though, I'll do it once I find a way to reduce the size of the gif to a decent size (the ones i have right now are 300MB...)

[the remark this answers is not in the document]
oke-aditya: Looks like we can. 
https://gifs-as-documentation.readthedocs.io/en/latest/

We need to compress the GIF though :( Maybe possible by running in very small video size say 120x60? For 3 seconds?

gallery/plot_optical_flow.py
oke-aditya: My doubt is it hard necessicity that image sizes should be divisible by 8. Or its adjusted by the model. 
The text description above says. Image sizes must be divisible by 8. Meaning that the model does not adjust.

NicolasHug: It's really a hardcoded constraint and the model cannot accept images that aren't divisible by 8 (even if it wanted to):

https://github.com/pytorch/vision/blob/74a1efcd06d9a72846f1805ce68b52cb5f01b2ea/torchvision/models/optical_flow/raft.py#L456

It has to be exactly divisible by 8 because we first downsample the inputs by 8, predict a downsampled flow, and then upsample the predicted flow by a factor of 8:

https://github.com/pytorch/vision/blob/74a1efcd06d9a72846f1805ce68b52cb5f01b2ea/torchvision/models/optical_flow/_utils.py#L26-L32

If the image wasn't a multiple of 8 to begin with, we wouldn't be able to upsample the flow to the right dimensions.


The fact that it's `8`  is somewhat arbitrary (we could downsample by 4 and upsample by 4) but we would still require images to be divisible by an integer N in general.

NicolasHug (approved):
Thanks a lot for your help @oke-aditya , LGTM, I'll merge when the docs are built!

NicolasHug:
Looks good lessgooooo https://1182173-73328905-gh.circle-artifacts.com/0/docs/auto_examples/plot_optical_flow.html#sphx-glr-auto-examples-plot-optical-flow-py

oke-aditya:
Lessssss goooooooooooooooo. Yay

PR #5317, opened 2022-01-31
use clang-format in pre-commit

pmeier:
Follow-up to https://github.com/pytorch/text/pull/1546#pullrequestreview-866729015.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -289,7 +289,9 @@ jobs:
             sudo mv clang-format /opt/clang-format
       - run:
           name: Lint C code
-          command: ./.circleci/unittest/linux/scripts/run-clang-format.py -r torchvision/csrc --clang-format-executable /opt/clang-format
+          command: |
+            /opt/clang-format --version
+            ./.circleci/unittest/linux/scripts/run-clang-format.py -r torchvision/csrc --clang-format-executable /opt/clang-format
       - run:
           name: Required lint modifications
           when: on_fail
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -247,55 +247,34 @@ jobs:
     steps:
       - checkout
       - pip_install:
+          descr: Install check utilities
           args: jinja2 pyyaml
       - run:
           name: Check CircleCI config consistency
-          command: |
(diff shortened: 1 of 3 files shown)


datumbox:
A couple of notes:
- I believe that PyTorch uses a specific version of clang and to keep it aligned with Core we have the specific binaries.
- I have concerns about the use of `pre-commit` if it forces a specific workflow to contributors. I am OK using it in the CI as long as we don't force users to use the specific paradigm but instead offer them to ability to run linters manually via commands or via IDE integrations.

pmeier:
> I believe that PyTorch uses a specific version of clang and to keep it aligned with Core we have the specific binaries.

Not sure I understand. There is difference in that we currently download the binary from https://oss-clang-format.s3.us-east-2.amazonaws.com/linux64/clang-format-linux64 whereas the hook downloads it from the official LLVM repo. I couldn't find any mention of this link outside of PyTorch. As you can see from the failing CI job, there seem to be differences. If that is intended, i.e. the binary we currently use is actually doing something different from the official one, we cannot use the hook anyway.

> I have concerns about the use of `pre-commit` if it forces a specific workflow to contributors. I am OK using it in the CI as long as we don't force users to use the specific paradigm but instead offer them to ability to run linters manually via commands or via IDE integrations.

It doesn't force a specific workflow. You can still run everything manually if you like. In this specific case, It should actually help the user. Our contribution guide doesn't mention `clang-format` at all. If a contributor currently hits an issue with it, we don't have any instructions how to run it at all. With this hook, this will be done automatically.

datumbox:
> I couldn't find any mention of this link outside of PyTorch.

See references at #2872. It might no longer be the case, if the tests pass without modifications on the C++ files you should be OK. Worth checking.

> You can still run everything manually if you like.

Could you explain how one runs the clang now without any use of pre-commit? For example, right now my IDE is setup to run `./.circleci/unittest/linux/scripts/run-clang-format.py -r torchvision/csrc`.

pmeier:
> if the tests pass without modifications on the C++ files you should be OK. 

That is not the case, thus we probably can't use the hook.

PR #5161, opened 2022-01-04
Removing/replacing Py3.6 references.

datumbox:
Removing remaining references of python 3.6 from CI and replacing where necessary with later python versions.

cc @seemethere

PR #5320, opened 2022-01-31
Added python 3.10 to CI

vfdev-5:
Description:
- Added python 3.10 to CI by updating `.circleci/regenerate.py` and running it to update `.circleci/config.yml`
  - skipped macosx as no 3.10 binaries yet.

### DIFF
--- a/.circleci/regenerate.py
+++ b/.circleci/regenerate.py
@@ -42,6 +42,9 @@ def build_workflows(prefix="", filter_branch=None, upload=False, indentation=6, 
                     # ROCm conda packages not yet supported
                     if cu_version.startswith("rocm") and btype == "conda":
                         continue
+                    # Skip MacOSX and python 3.10
+                    if python_version == "3.10" and os_type == "macos":
+                        continue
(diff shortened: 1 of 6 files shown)


pmeier:
Could you also check #5161 and add 3.10 where ever we removed 3.6? Plus, we also need to adapt the compatibility table in the README.

.circleci/regenerate.py
```diff
@@ -42,6 +42,9 @@ def build_workflows(prefix="", filter_branch=None, upload=False, indentation=6,
                     # ROCm conda packages not yet supported
                     if cu_version.startswith("rocm") and btype == "conda":
                         continue
+                    # Skip MacOSX and python 3.10
+                    if python_version == "3.10" and os_type == "macos":
```
datumbox, line 46: Are we missing dependencies? Can you link to the issue/error so that the release engineering team have visibility?

datumbox: We could remove the check and unblock the PR once they have fixed the macosx binaries.

PR #5323, opened 2022-01-31
add automatic feature type dispatch to functional transforms

pmeier:
This PR adds support for dispatching different `Feature`'s based on their type to the functional API. This effectively separates the API into three parts:
- low level: this section includes the kernels that only take vanilla tensors and the user is responsible to pass input of the correct type. This was explained and added in #5295.
- mid level: These functions take a single `Feature` and dispatch to the respective low level kernel. They also handle converting the meta into the required format as well as passing additional meta to the kernel. This part of the API is not user facing and should not be called manually. 
- high level: These functions take all `Feature`'s and dispatch them to the mid level API.

I'll illustrate this through an example with inline comments.

---

Implementing the dispatch logic, i.e. the mid and high level API explained above, involves a lot of boilerplate code. Thus, I've opted to autogenerate it from the kernels and a minimal accompanying configuration file. I'm well aware that autogenerated code is harder on the user so I tried to strike a balance. I wanted to avoid having only a configuration file for the user to see, because that makes it a lot harder to implement transformations outside of torchvision. At the same time I wanted to avoid manually written, error-prone code and documentation that we need to maintain.

My solution to this, is to track the generated file in version control. This way, the full implementation is out in the open for everyone to see, but we don't have to manually write it. Let me know what you think.


torchvision/prototype/transforms/functional/_dispatch.py
```diff
@@ -0,0 +1,448 @@
+# THIS FILE IS AUTOGENERATED
```
pmeier, line 1: As the line states, this is the autogenerated file. Note that if we don't autogenerate this. we would still have exactly the same implementation, but written manually.

torchvision/prototype/transforms/functional/_dispatch.py
```diff
@@ -0,0 +1,448 @@
...
+
+    output = F.horizontal_flip_bounding_box(converted_input, image_size=input.image_size)
+    output = F.convert_bounding_box_format(output, old_format=intermediate_format, new_format=input.format)
+
+    return features.BoundingBox.new_like(input, output)
+
+
+@Dispatcher
+def resize(
+    input: T,
```
pmeier, line 75: This is the high level API. It dispatches based on the type of `input`. The actual body is never called. The actual dispatch happens in `Dispatcher.__call__`.

torchvision/prototype/transforms/functional/_dispatch.py
```diff
@@ -0,0 +1,448 @@
...
+    interpolation: InterpolationMode = FEATURE_SPECIFIC_DEFAULT,  # type: ignore[assignment]
+    max_size: Optional[int] = None,
+    antialias: Optional[bool] = None,
+) -> T:
+    """ADDME"""
+    pass
+
+
+@resize.implements(features.Image, pil_kernel=_F.resize)
+def _resize_image(
```
pmeier, line 87: This is the mid level API. It should never be called directly and thus is not exposed.

torchvision/prototype/transforms/functional/utils.py
```diff
@@ -0,0 +1,72 @@
...
+                missing = [
+                    param
+                    for param in feature_specific_params
+                    if kwargs.get(param, self.FEATURE_SPECIFIC_PARAM) is self.FEATURE_SPECIFIC_PARAM
+                ]
+                if missing:
+                    raise TypeError(
+                        f"{implement_fn.__name__}() missing {len(missing)} required keyword-only arguments: "
+                        f"{sequence_to_str(missing, separate_last='and ')}"
+                    )
```
pmeier, line 48: There are cases where only one kernel needs extra information. For example, `cutmix_image` needs the `box` argument, while `cutmix_one_hot_label` does not. In order for the dispatcher to also work for labels if `box` is not supplied, we use a sentinel as default argument and emulate the error message here.

torchvision/prototype/transforms/functional/_dispatch.py
```diff
@@ -0,0 +1,448 @@
...
+    return features.Image.new_like(input, output)
+
+
+@Dispatcher
+def erase(input: T, *, i: int, j: int, h: int, w: int, v: torch.Tensor, inplace: bool = False) -> T:
+    """ADDME"""
+    pass
+
+
+@erase.implements(features.Image, pil_kernel=_F.erase)
```
vfdev-5, line 194: nit: erase does not work on PIL images
https://github.com/pytorch/vision/blob/8aad0e0a64c65b4e16855764288f0bc2ad2e14c0/torchvision/transforms/functional.py#L1276-L1277

pmeier: Good catch, thanks!

torchvision/prototype/transforms/functional/_dispatch.py
```diff
@@ -0,0 +1,448 @@
...
+def _resize_segmentation_mask(
+    input: features.SegmentationMask,
+    *,
+    size: List[int],
+    interpolation: InterpolationMode = InterpolationMode.NEAREST,
+    max_size: Optional[int] = None,
+    antialias: Optional[bool] = None,
+) -> features.SegmentationMask:
+    output = F.resize_segmentation_mask(
+        input, size=size, interpolation=interpolation, max_size=max_size, antialias=antialias
```
vfdev-5, line 121: We typically use nearest interp mode for segmenation masks resize and there is no antialias option for nearest mode. I think we can remove `antialias` argument.

vfdev-5:
Dispatcher approach looks interesting and rather complicated to me.
Due to a meta-language introduced in yml + a script generating final python script, I wonder how quickly we can solve future issues. Basically, the question is how developer-friendly `scripts/regenerate_transform_dispatch.py`? 

I also wonder why we could not use class inheritence ? Looks like majority of the ops are 
```python
class GenericTransform:

def _op_image(self, input: features.Image, **kwargs) -> features.Image:
    output = self.op_image(input, **kwargs)

def _op_bounding_box(self, input: features.BoundingBox, **kwargs) -> features.BoundingBox:
    intermediate_format = BoundingBoxFormat.XYXY
    converted_input = F.convert_bounding_box_format(input, old_format=input.format, new_format=intermediate_format)

    output = self.op_bbox(converted_input, image_size=input.image_size, **kwargs)
    output = F.convert_bounding_box_format(output, old_format=intermediate_format, new_format=input.format)

    return features.BoundingBox.new_like(input, output)

def _op_segm_mask(self, input, **kwargs):
    output = self.op_segm_mask(input, **kwargs)

    return features.SegmentationMask.new_like(input, output)    
```

Using that we may reduce the size of the codebase. For specific ops, we can introduce specific classes. There could be more transparency, IMO.
What do you think ?

torchvision/prototype/transforms/functional/_dispatch.py
pmeier: In that case, we should remove it from the kernel:

https://github.com/pytorch/vision/blob/7cffef67e251ab96663c3970056ad471db2aa6ed/torchvision/prototype/transforms/functional/_geometry.py#L53-L59

But that would mean one could set the interpolation mode to something else manually and would have no option to set antialias. Should we also hardcode the interpolation mode?

pmeier:
@vfdev 

> Due to a meta-language introduced in yml + a script generating final python script, I wonder how quickly we can solve future issues. Basically, the question is how developer-friendly `scripts/regenerate_transform_dispatch.py`?

Let's split "developers" into three roles: downstream library maintainers, contributors to torchvision and torchvision maintainers.

- downstream library maintainers: since the generated file is tracked in git, they don't need to know about the auto generation. If they want to add a new transform of their own, they can simply use the contents of `_dispatch.py` as templates and alter them to fit their needs.
- contributors to torchvision: contributing a new kernel is almost the same as before. The only extra step that one has to do now, is to add an entry to `dispatch.yaml` file. I feel like this is on par if not easier than to adapt the dispatcher manually.
- torchvision maintainers: IMO this is the only "controversial" bit. The code is dense and is currently lacking comments. Even when the documentation is sufficient, the implementation is still complex and it will take time to digest. 

For me this is still an advantage for two main reasons:

1. Given that we probably will add quite a few kernels in the near future to support bounding boxes, segmentation masks, and so on, auto generating the dispatch will pay off fast in terms of time invested.
2. The documentation of the dispatchers currently is just a dummy, but in the future it should link all kernels and all extra processing that happens before the kernel is called. If we autogenerate the dispatch, this information is already available and we could also autogenerate the docstrings. This would keep the documentation consistent across the board.

> I also wonder why we could not use class inheritence ?

The idea was to have the dispatch functionality available at the functional level. For example, `F.horizontal_flip(input)` should work regardless if `input` is an `Image` or a `BoundingBox`. With your approach we would lose that.

> There could be more transparency, IMO.

Transparency with respect to what? How the dispatch works?

torchvision/prototype/transforms/functional/_dispatch.py
vfdev-5: Yes, I'd say there is less sense to use linear interpolation mode for masks as it would produce artificial target classes on resizing.

torchvision/prototype/transforms/functional/utils.py
```diff
@@ -0,0 +1,82 @@
...
+        if issubclass(feature_type, PIL.Image.Image):
+            if self._pil_kernel is None:
+                raise TypeError("No PIL kernel")
+
+            return self._pil_kernel(input, *args, **kwargs)
+        elif not isinstance(input, torch.Tensor):
+            raise TypeError("No tensor")
+
+        if not (issubclass(type(input), features.Feature)):
+            input = features.Image(input)
```
vfdev-5, line 77: I wonder why `isinstance` does not work for you here ?

pmeier: This seems like artifacts of an older implementation. Let me clean this up.

pmeier:
@vfdev-5 Thanks for the clarification, I misunderstood what you were referring to. That is indeed a possibility that would achieve feature parity with the upside of eliminating the `__torch_function__` complexity. Let me implement a version of that in this PR.

datumbox (commented):
I think we should avoid autogenerating code because it adds a lot of extra complexity. Inheritance sounds an idea worth investigating.

Also perhaps if we can make all kernels to follow the same pattern:
```
@MY_METHOD.implements(features.Image, pil_kernel=_F.MY_METHOD)
def _MY_METHOD_image(input: features.Image, *, ...) -> features.Image:
    output = F.MY_METHOD(input, ...)

```

Then we could eliminate also their method bodies:

```
@MY_METHOD.implements(features.Image, pil_kernel=_F.MY_METHOD, tensor_kernel= F.MY_METHOD)
def _MY_METHOD_image(input: features.Image, *, ...) -> features.Image: pass
```

If that's true, then perhaps we don't even need those intermediate per type kernels and all we need is to define these directly as arguments to the `@Dispatcher` annotation. I don't know if that's possible, let me know your thoughts.

torchvision/prototype/transforms/functional/_dispatch.py
```diff
@@ -0,0 +1,448 @@
...
+) -> features.Image:
+    output = F.resize_image(input, size=size, interpolation=interpolation, max_size=max_size, antialias=antialias)
+
+    return features.Image.new_like(input, output)
+
+
+@resize.implements(features.BoundingBox)
+def _resize_bounding_box(input: features.BoundingBox, *, size: List[int], **_: Any) -> features.BoundingBox:
+    intermediate_format = BoundingBoxFormat.XYXY
+    converted_input = F.convert_bounding_box_format(input, old_format=input.format, new_format=intermediate_format)
```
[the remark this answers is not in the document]
pmeier: `resize_bounding_box` makes the assumption that the input is always in the XYXY format since a vanilla tensor doesn't carry this information. Since we have a `BoundingBox` feature here, we know what format the input has, we can convert it to the correct one. 

We can reasonably argue that the conversion could also happen at kernel level. But keep in mind that this means that a user has to pass `format=...` to every bounding box kernel that needs this conversion. Given that users will most likely only call the kernels together with JIT scripting or other low-level applications, I'm ok with both options.

torchvision/prototype/transforms/functional/_dispatch.py
```diff
@@ -0,0 +1,448 @@
...
+def resize(
+    input: T,
+    *,
+    size: List[int],
+    interpolation: InterpolationMode = FEATURE_SPECIFIC_DEFAULT,  # type: ignore[assignment]
+    max_size: Optional[int] = None,
+    antialias: Optional[bool] = None,
+) -> T:
+    """ADDME"""
+    pass
```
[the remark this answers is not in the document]
pmeier: This is only true for [`.pyi` stub files](https://mypy.readthedocs.io/en/stable/stubs.html#stub-file-syntax). In regular Python files, [`...` is just the literal for Ellipsis ](https://docs.python.org/dev/library/constants.html#Ellipsis). By convention, `...` should only be used for slicing syntax and annotations.

torchvision/prototype/transforms/functional/_dispatch.py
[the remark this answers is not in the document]
datumbox: The benefit of moving this to the kernel level is that kernels are completely self-contained with the entire logic they need. Also adding the code there means that JIT users will be able to access the entire logic of the kernel which is great. Another benefit is that we keep the dispatch code VERY minimal without any custom logic. This could allow us to completely eliminate method bodies as discussed above.

The only drawback of pushing the entire code in the kernels is that for images, where we will have to support PIL, we would need to have that extra code on the kernel (the verification mostly). But from what I can see so far, that's a reasonable compromise for keeping this layer as thin as possible.

If you see other drawbacks it's worth discussing them. Otherwise I would be OK moving this to the kernel.

torchvision/prototype/transforms/functional/_dispatch.py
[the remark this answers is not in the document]
datumbox: Ellipsis can be used also as a synonymous of `pass` and is no-op. See it's use [here](https://www.python.org/dev/peps/pep-0484/) as an example. The 2 are equivalent so at the end of the day you can use whatever you want. I just thought that `...` is more common due to their use on stub files. No strong opinion TBH.

pmeier (commented):
I've completely removed the auto-generation and moved the dispatchers to the respective kernels. Please have another look.

torchvision/prototype/transforms/functional/_dispatch.py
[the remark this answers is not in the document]
pmeier: See https://github.com/pytorch/vision/pull/5323#pullrequestreview-874795726.

torchvision/prototype/features/_feature.py
```diff
@@ -57,5 +57,24 @@ def new_like(cls, other, data, *, dtype=None, device=None, **metadata):
...
         return cls(data, dtype=dtype or other.dtype, device=device or other.device, **metadata)
 
+    @classmethod
+    def __torch_function__(
+        cls,
+        func: Callable[..., torch.Tensor],
+        types: Tuple[Type[torch.Tensor], ...],
+        args: Sequence[Any] = (),
+        kwargs: Optional[Mapping[str, Any]] = None,
+    ) -> torch.Tensor:
```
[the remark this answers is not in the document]
pmeier: Technically, return type should be `Union[torch.Tensor, F]` but this has no practical use. Since the return type depends in the input argument `func`, there is no way for `mypy` to figure out which type it is going to get. So if we would be calling this explicitly, we needed to `cast` anyway if we know we get a `Feature` type back. Good thing for us, we never call this directly. Thus, we can get away with the `Tensor` annotation, since all `Feature`'s are also `Tensor`'s.

torchvision/prototype/transforms/functional/__init__.py
```diff
@@ -1,27 +1,54 @@
-from ._augment import erase_image, mixup_image, mixup_one_hot_label, cutmix_image, cutmix_one_hot_label
+from . import utils  # usort: skip
+
+from torchvision.transforms import InterpolationMode
+
+from ._augment import (
+    erase_image,
```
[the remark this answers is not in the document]
pmeier: Exactly. Due to the dynamic nature of the dispatch as well as the extra meta data on the `Feature`'s, we can't script the dispatchers. The low level kernels enable users to have this functionality at the cost of inconvenience.

torchvision/prototype/transforms/functional/utils.py
```diff
@@ -0,0 +1,93 @@
...
+
+import PIL.Image
+import torch
+import torch.overrides
+from torchvision.prototype import features
+from torchvision.prototype.utils._internal import sequence_to_str
+
+
+def is_supported(obj: Any, *types: Type) -> bool:
+    return (obj if isinstance(obj, type) else type(obj)) in types
```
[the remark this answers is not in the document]
pmeier: It will be used by some transforms later on. See for example https://github.com/pytorch/vision/blob/16a4473519b2105154b33f34083c62de9c00396c/torchvision/prototype/transforms/_misc.py#L29. I would keep it as function or otherwise I'll make it a function again for the next PR.

torchvision/prototype/transforms/functional/_augment.py
```diff
@@ -1,10 +1,29 @@
...
 from torchvision.transforms import functional as _F
 
+from .utils import dispatch
+
+T = TypeVar("T", bound=features.Feature)
+
+
+@dispatch
+def erase(input: T, *, i: int, j: int, h: int, w: int, v: torch.Tensor, inplace: bool = False) -> T:
+    """ADDME"""
```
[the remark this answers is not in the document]
pmeier: Whatever we want to show up in the documentation for `resize`. Given that these dispatchers don't have any functionality, I would keep the docstring minimal and for example just have a table that details what feature type gets dispatched to which kernel. Let me add one example to help reviewing.

torchvision/prototype/transforms/functional/utils.py
```diff
@@ -0,0 +1,93 @@
...
+        if issubclass(feature_type, PIL.Image.Image):
+            if self._pil_fn is None:
+                raise TypeError("No PIL kernel")
+
+            return self._pil_fn(input, *args, **kwargs)
+        elif not issubclass(feature_type, torch.Tensor):
+            raise TypeError("No tensor")
+
+        if not issubclass(feature_type, features.Feature):
+            input = features.Image(input)
```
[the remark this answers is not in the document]
pmeier: For BC. Right now, users can use `F.resize` with a vanilla tensor and it will treat it as an image since we don't have other kernels yet. Without this, the dispatcher no longer accepts vanilla tensors.

We can phase out this behavior later if the API is stable by first deprecating calling the dispatchers with vanilla tensors and later on bailing out, which will happen automatically when removing the lines above.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -1,21 +1,51 @@
...
+def resize(
+    input: T,
+    *,
+    size: List[int],
+    interpolation: InterpolationMode = dispatch.FEATURE_SPECIFIC_DEFAULT,  # type: ignore[assignment]
+    max_size: Optional[int] = None,
+    antialias: Optional[bool] = None,
+) -> T:
+    """ADDME"""
+    pass
```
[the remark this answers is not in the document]
pmeier: > Do we really need to delay the registration into multiple commands vs RAII (resize is created with provided resize_image, resize_bbox etc at construction time) ?

*We* don't need to, no. But we would have to keep this mechanism if someone outside of `torchvision` wants to register a custom `Feature` kernel for a transform. Imagine I have `MyAwesomeFeature` and a kernel `resize_my_awesome_feature` how would I add it to the default `F.resize` dispatch if we don't have an option to register at runtime?

> What would happen if signature of `resize` is different from `resize_image`, `resize_segmentation_mask` etc ? and to what corresponds the signature of `resize` ?

Just like the body of the dispatcher function, the signature is only used for the documentation. If it diverges from the kernel signatures, we should have explanations in the docstring what is happening.

torchvision/prototype/features/_feature.py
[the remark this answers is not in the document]
datumbox: Thanks for the input. I understand now. Questions:
- Should we remove the F type definition on top then?
- Or perhaps it is still worth it to return F and do an explicit cast, just to reinforce to the reader of the code that this is supposed to return Features? (the argument here is that code is read more times than written)

torchvision/prototype/transforms/functional/__init__.py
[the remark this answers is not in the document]
datumbox: Personally I would keep them public but wouldn't expose these from `__init__`. We want to push people to use the high-level kernels. Those who want low-level could import directly. Unfortunately that is not possible because all the subpackages are private (for example `_augment`). We can keep as-is and examine potentially exposing these from somewhere else.

torchvision/prototype/transforms/functional/utils.py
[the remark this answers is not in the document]
datumbox: I suggest to add a warning explaining that this will not be supported on the future. No need to specify version as we don't know when we will release this prototype.

torchvision/prototype/transforms/functional/_geometry.py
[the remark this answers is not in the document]
vfdev-5: > We don't need to, no. But we would have to keep this mechanism if someone outside of torchvision wants to register a custom Feature kernel for a transform. Imagine I have MyAwesomeFeature and a kernel resize_my_awesome_feature how would I add it to the default F.resize dispatch if we don't have an option to register at runtime?

If we think to expose this as public API, we have to test it better, write docs etc. 
Yes, feature like extending resize to make it work on other types would be nice but then we have to ensure which scope we provide. Can user override `resize_image` for example ? Can we register on other non-tensor types, e.g. `str`:
```python
@resize.implements(str)
def resize_str(image, *args, **kwargs):
    print("Call resize_str")
    return image[1:-1]
```
(Today, this does not error but raises an exception only on call)
How about parity then ? If we provide publicly add operation (`dispatch.register`), then we may want to provide removal operation, e.g `dispatch.deregister`. I do not see any application of that at the moment...

Being careful today, I'd make `register` as private method and do not expose right now for the users.

torchvision/prototype/transforms/functional/__init__.py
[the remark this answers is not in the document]
pmeier: > We want to push people to use the high-level kernels.

Is that the case? I was under the impression that scriptable kernels is an integral part of some workflows. Why would make it harder for these users?

torchvision/prototype/transforms/functional/_geometry.py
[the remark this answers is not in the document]
pmeier: @vfdev-5 

> If we think to expose this as public API, we have to test it better, write docs etc.

Of course, but that has time until we actually expose this outside of `torchvision.prototype`.

> Yes, feature like extending resize to make it work on other types would be nice but then we have to ensure which scope we provide.

I don't think this is nice-to-have, but rather integral. Our transforms should be flexible enough to cater to the needs of 90% of our user base without them using hacky workarounds. To give an example, Franciscos new project already needs this if they want to adopt our new transforms.

> Can we register on other non-tensor types, e.g. `str`

Nope, not for now. The output wrapping into a `Feature` already implied this, but I will make it explicit with an error. This way, we have instant feedback.

> How about parity then ? If we provide publicly add operation (`dispatch.register`), then we may want to provide removal operation, e.g `dispatch.deregister`. I do not see any application of that at the moment...

I think that is fair game to wait for someone actually requesting this. In the current setup if you call `register` again with an already registered feature type, the old entry is simply overwritten. Thus, if the user wants to have a custom kernel this is already supported. 

I don't see a use case for de-registering feature types. If we had such a functionality, calling a dispatcher with a de-registered feature type would result in an error. Thus, it should be much simpler to not call the dispatcher in the first place with whatever feature type you want to de-register.

---

@datumbox 

> This is a potential expansion feature but it shouldn't be our target for the first version.

See my answer to @vfdev-5 above.

> I agree that we should provide these at construction time as arguments to the dispatch annotation to simplify code and make it less verbose.

I don't think it would simplify the implementation or make it less verbose. Compare

```py
@dispatch
def resize(...):
    ...


resize.register(resize_image, features.Image, pil_kernel=_F.resize)
resize.register(resize_segmentation_mask, features.SegmentationMask)
```

with

```py
@dispatch(
    {
        features.Image: resize_image,
        features.SegmentationMask: resize_segmentation_mask,
    },
    pil_kernel=_F.resize,
)
def resize(...):
    ...
```

Plus, we would need to have extra indication if the callable we pass handles the output wrapping or not. With the `register` approach we can simply set `wrap_output=False` which automatically only applies to the current callable.

torchvision/prototype/transforms/functional/__init__.py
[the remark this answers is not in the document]
pmeier: I'll propose something in a follow-up PR.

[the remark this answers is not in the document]
datumbox: Instead of creating PRs directly, create an issue outlining the problem. This requires discussion as there are alternatives. I just want to avoid biasing towards one direction without discussion.

torchvision/prototype/features/_feature.py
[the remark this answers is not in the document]
pmeier: > Should we remove the F type definition on top then?

I've added annotations to the other methods. Now `F` is used.

> Or perhaps it is still worth it to return F and do an explicit cast

This method is never called by us directly, so there no place where we could perform an explicit cast. `__torch_function__` is for example triggered if you do something like `Image() + 1` or any other PyTorch operation. I've added an explanation as docstring. Let me know if you feel it is not sufficient.

pmeier (commented):
@vfdev-5 @datumbox I've refactored and now `dispatch` is a regular decorator that is configured with all kernels upfront. Please do another round.

torchvision/prototype/transforms/functional/_geometry.py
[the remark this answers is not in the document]
pmeier: Marking this thread as resolved since the latest commits (hopefully) bring the `@dispatch` decorator more in the direction of your suggestions.

torchvision/prototype/transforms/functional/_color.py
```diff
@@ -1,20 +1,125 @@
...
+    {
+        features.Image: adjust_sharpness_image,
+    }
+)
+def adjust_sharpness(input: T, *, sharpness_factor: float) -> T:
+    """ADDME"""
+    pass
+
+
 posterize_image = _F.posterize
```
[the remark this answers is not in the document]
pmeier: As discussed offline, I've added a new commit that moves the kernels into `transforms.kernels` and only returns the dispatchers in `transforms.functional`.

torchvision/prototype/features/_feature.py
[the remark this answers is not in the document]
pmeier: > I still don't see it being used on this file. Am I missing something?

You are probably only looking at the changes in this PR. I've added the annotations to the base branch, for example https://github.com/pytorch/vision/blob/ab11f617e09733768b30a80a552e8fae2e5de983/torchvision/prototype/features/_feature.py#L40-L46

torchvision/prototype/transforms/functional/_augment.py
```diff
@@ -1,52 +1,68 @@
...
 
-    return _mixup(image_batch, -4, lam, inplace)
+@dispatch(
+    {
+        features.Image: K.cutmix_image,
+        features.OneHotLabel: K.cutmix_one_hot_label,
+    },
+)
+def cutmix(input: T, *, box: Tuple[int, int, int, int], lam_adjusted: float, inplace: bool) -> T:
+    """Perform the CutMix operation as introduced in the paper
```
pmeier, line 41: Rendered documentation looks like this

Screenshot from 2022-02-09 22-57-43

I'm not super happy with the design yet, but it has all the information that I think is necessary on the dispatcher. Of course the kernels documentation is clickable.

torchvision/prototype/transforms/functional/_augment.py
```diff
@@ -1,52 +1,68 @@
...
 
 
-def _mixup(input: torch.Tensor, batch_dim: int, lam: float, inplace: bool) -> torch.Tensor:
-    if not inplace:
-        input = input.clone()
+@dispatch(
+    {
+        features.Image: K.erase_image,
+    },
+)
```
vfdev-5, line 17: Matter of taste question, can we remove trailing commas to make this inline ? 4 lines to create a dispatcher for images looks a bit ugly.
Otherwise, we can precode the functions inside dispatch constructor: (op_image, op_bbox, op_segm_mask, etc).

torchvision/prototype/transforms/functional/_utils.py
```diff
@@ -0,0 +1,91 @@
...
+        TypeError: If any key in ``kernels`` is not a strict subclass of
+            :class:`~torchvision.prototype.features.Feature`.
+        TypeError: If ``pil_kernel`` is specified, but no kernel for :class:`~torchvision.prototype.features.Image` is
+            available.
+        TypeError: If the decorated function is called with neither a ``PIL`` image nor a :class:`~torch.Tensor`.
+        TypeError: If the decorated function is called with an input that cannot be dispatched.
+    """
+    for feature_type in kernels:
+        if not (issubclass(feature_type, features.Feature) and feature_type is not features.Feature):
+            raise TypeError("XXX")
```
vfdev-5, line 54:
```suggestion
    for feature_type in kernels:
        if not issubclass(feature_type, features.Feature):
            raise TypeError("Unsupported feature type")
```

```
class A: pass

class B(A): pass

issubclass(A, A), issubclass(B, A)  
> True, True
```

vfdev-5: Shouldn't we also check if value is callable ?

torchvision/prototype/transforms/functional/_utils.py
```diff
@@ -0,0 +1,91 @@
...
+    if pil_kernel and features.Image not in kernels:
+        raise TypeError("PIL kernel can only be registered for images")
+
+    def outer_wrapper(dispatch_fn: Callable[..., F]) -> Callable[..., F]:
+        @functools.wraps(dispatch_fn)
+        def inner_wrapper(input: F, *args: Any, **kwargs: Any) -> F:
+            feature_type = type(input)
+
+            if issubclass(feature_type, PIL.Image.Image):
```
vfdev-5, line 63: Matter of taste, but i'd prefer to do `isinstance(input, PIL.Image.Image)` and same below

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -1,71 +1,121 @@
...
-    *,
-    old_image_size: List[int],
-    new_image_size: List[int],
-) -> torch.Tensor:
-    old_height, old_width = old_image_size
-    new_height, new_width = new_image_size
-    ratios = torch.tensor((new_width / old_width, new_height / old_height))
-    return bounding_box.view(-1, 2, 2).mul(ratios).view(bounding_box.shape)
 
+def _resize_bounding_box(input: features.BoundingBox, *, size: List[int]) -> features.BoundingBox:
```
vfdev-5, line 29: It is still ambiguous signatures for decorated function and implementations. As I understand dispatch decorates empty function and does completely another thing. Unfortunately, I do not understand the point of providing `resize` complete signature which can be unrelated to the implementations signatures. I'd instead code it with *args, **kwargs and in the docs would detail that *args, **kwargs are specific to the implementations:
```python
@dispatch(
    {
        features.Image: K.resize_image,
        features.SegmentationMask: K.resize_segmentation_mask,
        features.BoundingBox: _resize_bounding_box,
    },
    pil_kernel=_F.resize,
)
def resize(input: T, *args: Any, **kwargs: Any) -> T:
    """ADDME"""
    ...
```

vfdev-5 (commented):
@pmeier thanks for the updates, looks good to me.
Left few nits and comments.

torchvision/prototype/transforms/functional/_augment.py
pmeier: I agree for dispatchers with a single kernel, but I would like to keep the format for more kernels.

torchvision/prototype/transforms/functional/_utils.py
pmeier: To continue  you example, we want to explicitly exclude `A` here since it is the generic super class.

torchvision/prototype/transforms/functional/_utils.py
pmeier: In general, I do too. But here we already need the `feature_type` variable to select the kernels so the intention was to re-use it. But no strong opinion. Maybe @datumbox can be the tie breaker? :innocent:

torchvision/prototype/transforms/functional/_augment.py
```diff
@@ -15,7 +15,7 @@
         features.Image: K.erase_image,
     },
 )
-def erase(input: T, *, i: int, j: int, h: int, w: int, v: torch.Tensor, inplace: bool = False) -> T:
+def erase(input: T, *, i: int, j: int, h: int, w: int, v: torch.Tensor, inplace: bool) -> T:
```
datumbox: Unrelated to your API (more related to the previous implementation) but I recommend not to support in-place here since we don't support it in 90% of the transforms. I would do a scan in all places where inplace is support it and straight remove the functionality.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -1,71 +1,121 @@
...
+        features.SegmentationMask: K.resize_segmentation_mask,
+        features.BoundingBox: _resize_bounding_box,
+    },
+    pil_kernel=_F.resize,
+)
+def resize(
+    input: T,
+    *,
+    size: List[int],
+    interpolation: InterpolationMode,
```
datumbox: What happened to the default values?

torchvision/prototype/transforms/kernels/_augment.py
```diff
@@ -0,0 +1,52 @@
...
+
+def _mixup(input: torch.Tensor, batch_dim: int, lam: float, inplace: bool) -> torch.Tensor:
+    if not inplace:
+        input = input.clone()
+
+    input_rolled = input.roll(1, batch_dim)
+    return input.mul_(lam).add_(input_rolled.mul_(1 - lam))
+
+
+def mixup_image(image_batch: torch.Tensor, *, lam: float, inplace: bool = False) -> torch.Tensor:
```
datumbox, line 18: I'm a bit confused over the use of default values. What happened to them and why this method is allowed to have one?

datumbox (commented):
I love the new proposal. Prior merging we need to close 3 things:
1. Clarify what happened to the default values in all high-level kernels and why the low-level kernels retain them? I would expect the opposite. 
2. I propose to remove `inplace` from all kernels to align with the general policy of not modifying input. This is not related to your API and admittedly is something I wan't convinced about previously (see https://github.com/pytorch/vision/pull/4379#discussion_r704322791; @fmassa I hope you feel vindicated).
- Resolve the conflict with main branch.

torchvision/prototype/transforms/kernels/_augment.py
```diff
@@ -0,0 +1,52 @@
...
+import torch
+from torchvision.transforms import functional as _F
+
+
+erase_image = _F.erase
+
+
+def _mixup(input: torch.Tensor, batch_dim: int, lam: float, inplace: bool) -> torch.Tensor:
+    if not inplace:
+        input = input.clone()
```
fmassa, line 12: FYI this is effectively useless, as `Tensor.roll` returns a copy.

torchvision/prototype/transforms/kernels/_geometry.py
```diff
@@ -0,0 +1,76 @@
...
+def horizontal_flip_bounding_box(
+    bounding_box: torch.Tensor, *, format: features.BoundingBoxFormat, image_size: Tuple[int, int]
+) -> torch.Tensor:
+    shape = bounding_box.shape
+
+    bounding_box = convert_bounding_box_format(
+        bounding_box, old_format=format, new_format=features.BoundingBoxFormat.XYXY
+    ).view(-1, 4)
+
+    bounding_box[:, [0, 2]] = image_size[1] - bounding_box[:, [0, 2]]
```
fmassa, line 25: You probably forgot to flip `[2, 0]` somewhere in here?

torchvision/prototype/transforms/kernels/_meta_conversion.py
```diff
@@ -34,13 +34,10 @@ def _xyxy_to_cxcywh(xyxy: torch.Tensor) -> torch.Tensor:
...
 def convert_bounding_box_format(
-    bounding_box: torch.Tensor, *, old_format: BoundingBoxFormat, new_format: BoundingBoxFormat, copy: bool = False
+    bounding_box: torch.Tensor, *, old_format: BoundingBoxFormat, new_format: BoundingBoxFormat
 ) -> torch.Tensor:
     if new_format == old_format:
-        if copy:
-            return bounding_box.clone()
-        else:
-            return bounding_box
+        return bounding_box
```
datumbox, line 40: You missed `clone()`. Without it, you make in place modifications at `horizontal_flip_bounding_box`

torchvision/prototype/transforms/kernels/_geometry.py
```diff
@@ -16,11 +16,17 @@
...
+
     bounding_box = convert_bounding_box_format(
-        bounding_box, old_format=format, new_format=features.BoundingBoxFormat.XYXY, copy=True
-    )
-    bounding_box[..., (0, 2)] = image_size[1] - bounding_box[..., (2, 0)]
-    return convert_bounding_box_format(bounding_box, old_format=features.BoundingBoxFormat.XYXY, new_format=format)
+        bounding_box, old_format=format, new_format=features.BoundingBoxFormat.XYXY
+    ).view(-1, 4)
+
+    bounding_box[:, [0, 2]] = image_size[1] - bounding_box[:, [0, 2]]
```
datumbox, line 25: Whenever you do in-place modifications like that, make sure you have cloned the tensor prior. The above method seems not to clone in specific conditions. Check the code-base to confirm we are good for the rest.

We will also later on require a test to check explicitly for inplace modifications.

torchvision/prototype/transforms/kernels/_geometry.py
```diff
@@ -0,0 +1,76 @@
...
+
+def resize_segmentation_mask(
+    segmentation_mask: torch.Tensor,
+    size: List[int],
+    interpolation: InterpolationMode = InterpolationMode.NEAREST,
+    max_size: Optional[int] = None,
+    antialias: Optional[bool] = None,
+) -> torch.Tensor:
+    return resize_image(
+        segmentation_mask, size=size, interpolation=interpolation, max_size=max_size, antialias=antialias
```
fmassa, line 59: shouldn't we disable the `interpolation` altogether here?

Otherwise if the user passes `interpolation=BICUBIC` in the main `Resize()` transform, this will be using bicubic interpolation unwillingly?

torchvision/prototype/transforms/__init__.py
```diff
@@ -1,4 +1,3 @@
-from . import functional
-from .functional import InterpolationMode  # usort: skip
-
+from . import kernels  # usort: skip
+from . import functional  # usort: skip
```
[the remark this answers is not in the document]
pmeier: We need to keep the order here since `functional` depends on `kernels`. Otherwise you cannot do `from torchvision.prototype.transforms import kernels as K` in `torchvision.prototype.transforms.functional`.

torchvision/prototype/transforms/kernels/_geometry.py
```diff
@@ -0,0 +1,76 @@
...
+    )
+
+
+# TODO: handle max_size
+def resize_bounding_box(
+    bounding_box: torch.Tensor, *, old_image_size: List[int], new_image_size: List[int]
+) -> torch.Tensor:
+    old_height, old_width = old_image_size
+    new_height, new_width = new_image_size
+    ratios = torch.tensor((new_width / old_width, new_height / old_height))
```
fmassa, line 69: nit: you probably want to keep the device of the original boxes in here, otherwise the code below won't work with bounding boxes on the GPU

torchvision/prototype/transforms/functional/_augment.py
datumbox: How does the documentation look like now that you switched to `*args: Any, **kwargs: Any`. I'm concerned that as a user I don't know what parameters I need to pass. Also if I need to go to the low-level kernel, am I guaranteed that the two will match exactly in terms of parameters? I think there were exceptions, right?

datumbox (commented):
More comments, since you pushed changes right before I submit my previous.

torchvision/prototype/transforms/kernels/_geometry.py
datumbox: @fmassa Ah good catch thanks for flagging. 
@pmeier The mask is basically a mapping of label values. We can't linearly interpolate between ids. Only Nearest makes sense here.

torchvision/prototype/transforms/functional/_utils.py
pmeier: I'm going to keep it for now

pmeier:
After some offline discussion with @datumbox we are not sure how to handle the documentation of the dispatchers. In the case regular case it is fine to just link the low-level kernels. But there are dispatchers that do not call the low-level kernels directly but rather an intermediate layer. 

This intermediate layer is very thin and only performs some meta data access and maybe parameter mapping. For example, the low level kernel `horizontal_flip_bounding_box` needs `format` and `image_size` since they are not available on vanilla tensors. The dispatcher however has access to a `BoundingBox` feature which stores them. Thus, we have this intermediate function:

```py
def _horizontal_flip_bounding_box(input: features.BoundingBox) -> torch.Tensor:
    return kernels.horizontal_flip_bounding_box(input, format=input.format, image_size=input.image_size)
```

This is fine from a functionality standpoint, but we need to document this somehow. Otherwise, if we just link `kernels.horizontal_flip_bounding_box`, the user might (rightfully) assume that they need to pass `format` and `image_size` explicitly.

We came up with two solutions:

1. Make the intermediate layer public, document its behavior and link it from the dispatcher.
2. Put the documentation of the intermediate layer into the documentation of the dispatcher. 

Thoughts?

datumbox:
I'm in favour of option 1 because it provides a clean policy and reuses standard documentation tools in python. The policy is simple:
- The dispatcher documentation always links to the subscribed kernels for info on arguments.
- When the intermediate kernel IS the subscribed kernel then we add the documentation of its parameters there. We also describe it uses the low-level kernel and provide a link.
- When the low-level kernel is subscribed, then we juts link to it as above.
- No special cases, no need for extra handling.

Option 2 has its pros (it doesn't make public yet another kernel to the users) but I feel that making it private you hide information from the users and they have to figure out what is going on and why the low-level kernel isn't used directly. To answer that you need to bring more documentation to the high-level kernel, describe parameters etc. This can bring disconnection between the documentation on the high-level and the params of the intermediate.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -1,72 +1,68 @@
...
-horizontal_flip_image = _F.hflip
+T = TypeVar("T", bound=features.Feature)
 
 
-def horizontal_flip_bounding_box(bounding_box: torch.Tensor, *, image_size: Tuple[int, int]) -> torch.Tensor:
-    shape = bounding_box.shape
-    bounding_box = bounding_box.view(-1, 4)
-    bounding_box[:, [0, 2]] = image_size[1] - bounding_box[:, [0, 2]]
-    return bounding_box.view(shape)
+def _horizontal_flip_bounding_box(input: features.BoundingBox) -> torch.Tensor:
```
vfdev-5, line 13: @pmeier why this one returns torch.Tensor and in `_resize_bounding_box` output is features.BoundingBox and wrapping is explicit.

datumbox: I think that might be a bug actually. IT should return a BoundingBox not a Tensor.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -1,72 +1,101 @@
...
+        features.BoundingBox: None,
+    },
 )
+def horizontal_flip(input: T, *args: Any, **kwargs: Any) -> T:
+    """ADDME"""
+    if isinstance(input, features.BoundingBox):
+        output = K.horizontal_flip_bounding_box(input, format=input.format, image_size=input.image_size)
+        return cast(T, features.BoundingBox.new_like(input, output))
+
+    raise RuntimeError(
```
datumbox, line 28: I would just throw a simple RuntimeError() here without text (to avoid needing to maintain all these messages. These should be unreachable code if the dispatcher works right.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -1,72 +1,101 @@
...
+    }
+)
+def resize(input: T, *args: Any, **kwargs: Any) -> T:
+    """ADDME"""
+    if isinstance(input, features.BoundingBox):
+        size = kwargs.pop("size")
+        output = K.resize_bounding_box(input, old_image_size=list(input.image_size), new_image_size=size)
+        return cast(T, features.BoundingBox.new_like(input, output, image_size=size))
+
+    raise RuntimeError(
```
datumbox, line 50: same

torchvision/prototype/transforms/functional/_utils.py
```diff
@@ -0,0 +1,108 @@
...
+
+    Raises:
+        TypeError: If any key in ``kernels`` is not a strict subclass of
+            :class:`~torchvision.prototype.features.Feature`.
+        TypeError: If any value in ``kernels`` is not callable with ``kernel(input, *args, **kwargs)``.
+        TypeError: If ``pil_kernel`` is specified, but no kernel for :class:`~torchvision.prototype.features.Image` is
+            available.
+        TypeError: If the decorated function is called with neither a ``PIL`` image nor a :class:`~torch.Tensor`.
+        TypeError: If the decorated function is called with an input that cannot be dispatched.
+    """
```
datumbox, line 54: This needs updating. I know it's on your radar, just marking.

torchvision/prototype/transforms/functional/_utils.py
```diff
@@ -0,0 +1,108 @@
...
+                        if isinstance(input, feature_type)
+                    )
+                except StopIteration:
+                    raise TypeError(f"No support for {type(input).__name__}") from None
+
+            if kernel is None:
+                output = dispatch_fn(input, *args, **kwargs)
+                if output is None:
+                    raise RuntimeError(
+                        f"dispatch_fn() did not handle inputs of type {type(input).__name__} "
```
datumbox, line 95: perhaps `dispatch_fn.__name__` will give you the right name?

datumbox (approved):
The code looks good to me, only 2 nits and we could consider it as final candidate for solution. I would recommend merging it to the feature branch (avoid making this PR longer) and follow up with alternatives on top of this on new PRs.

torchvision/prototype/transforms/functional/_augment.py
```diff
@@ -1,52 +1,57 @@
...
+def erase(input: T, *args: Any, **kwargs: Any) -> T:
+    """ADDME"""
+    ...
 
-    input_rolled = input.roll(1, batch_dim)
-    return input.mul_(lam).add_(input_rolled.mul_(1 - lam))
 
+@dispatch(
+    {
+        features.Image: K.mixup_image,
```
datumbox, line 26: I believe we are missing the dispatch for `torch.Tensor`. Worth checking where else this is the case and add it.

datumbox (commented):
@pmeier I noticed a few more improvements we should bring on a new PR. Please check below and let me know what you think.

torchvision/prototype/features/_feature.py
```diff
@@ -76,5 +76,45 @@ def new_like(
...
+        with DisableTorchFunction():
+            output = func(*args, **kwargs)
+
+        if func is torch.Tensor.clone:
+            return cls.new_like(args[0], output)
+        elif func is torch.Tensor.to:
+            return cls.new_like(args[0], output, dtype=output.dtype, device=output.device)
+        else:
+            return output
+
```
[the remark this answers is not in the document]
pmeier: Although this does not need be supported in the first version, `Feature` should not be an internal class. We want users to be able to create their own custom features if it is useful for their use case.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -1,72 +1,95 @@
...
+        PIL.Image.Image: _F.resize,
+        features.Image: K.resize_image,
+        features.SegmentationMask: K.resize_segmentation_mask,
+        features.BoundingBox: None,
+    }
+)
+def resize(input: T, *args: Any, **kwargs: Any) -> T:
+    """ADDME"""
+    if isinstance(input, features.BoundingBox):
+        size = kwargs.pop("size")
```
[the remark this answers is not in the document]
pmeier: Taking `resize_bounding_box` as an example, it would then take `size` and `old_image_size`. In that case we should probably also drop the "old" to make it `size` and `image_size`.

torchvision/prototype/features/_feature.py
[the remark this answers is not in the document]
datumbox: It should be private for now.

PR #5326, opened 2022-02-01
add keypoints annotations to COCO prototype dataset

pmeier:
We use them in our detection references.

### DIFF
--- a/torchvision/prototype/datasets/_builtin/coco.py
+++ b/torchvision/prototype/datasets/_builtin/coco.py
@@ -120,6 +120,12 @@ class Coco(Dataset):
             ann_ids=[ann["id"] for ann in anns],
         )
 
+    def _decode_person_keypoints_anns(self, anns: List[Dict[str, Any]], image_meta: Dict[str, Any]) -> Dict[str, Any]:
+        return dict(
+            self._decode_instances_anns(anns, image_meta),
+            keypoints=Feature(torch.tensor([ann["keypoints"] for ann in anns]).view(len(anns), -1, 3)),
+        )
+
     def _decode_captions_ann(self, anns: List[Dict[str, Any]], image_meta: Dict[str, Any]) -> Dict[str, Any]:
         return dict(
             captions=[ann["caption"] for ann in anns],
@@ -129,6 +135,7 @@ class Coco(Dataset):
     _ANN_DECODERS = OrderedDict(
         [
             ("instances", _decode_instances_anns),
+            ("person_keypoints", _decode_person_keypoints_anns),
             ("captions", _decode_captions_ann),
         ]
     )
--- a/test/builtin_dataset_mocks.py
+++ b/test/builtin_dataset_mocks.py
@@ -543,6 +543,16 @@ class CocoMockData:
         )
 
     @staticmethod
+    def _make_person_keypoints_data(ann_id, image_meta):
+        data = CocoMockData._make_instances_data(ann_id, image_meta)
+        keypoints = [
+            (*make_tensor((2,), dtype=torch.int, low=0).tolist(), visibility) if visibility else (0, 0, 0)
+            for visibility in torch.randint(3, (17,)).tolist()
+        ]
+        data["keypoints"] = list(itertools.chain.from_iterable(keypoints))
+        return data
+
+    @staticmethod
     def _make_captions_data(ann_id, image_meta):
         return dict(caption=f"Caption {ann_id} describing image {image_meta['id']}.")
 
@@ -551,6 +561,7 @@ class CocoMockData:
         num_anns_per_image = torch.zeros((len(images_meta),), dtype=torch.int64)
         for annotations, fn in (
             ("instances", cls._make_instances_data),
+            ("person_keypoints", cls._make_person_keypoints_data),
             ("captions", cls._make_captions_data),
         ):
             num_anns_per_image += cls._make_annotations_json(


torchvision/prototype/transforms/_geometry.py
```diff
@@ -3,12 +3,12 @@
...
 from torch.nn.functional import interpolate
 from torchvision.prototype.datasets.utils import SampleQuery
-from torchvision.prototype.features import BoundingBox, Image, Label
+from torchvision.prototype.features import BoundingBox, Image, Label, KeyPoint
 from torchvision.prototype.transforms import Transform
 
 
 class HorizontalFlip(Transform):
-    NO_OP_FEATURE_TYPES = {Label}
+    NO_OP_FEATURE_TYPES = {Label, KeyPoint}
```
fmassa, line 11: Keypoint should be aware of how to flip itself.

This is our current implementation for COCO: https://github.com/pytorch/vision/blob/d5a22a852f178b181a2077e01a14eae9e25a8ff7/references/detection/transforms.py#L10-L17

basically, `left` becomes `right` and vice-versa, so this needs to be taken into account as well (apart from just the coordinate flip).

This hints to having an extra meta-data associated with keypoints that can be applicable in those use-cases.
This is what I came up with back in [`maskrcnn-benchmark`](https://github.com/facebookresearch/maskrcnn-benchmark/blob/57eec25b75144d9fb1a6857f32553e1574177daf/maskrcnn_benchmark/structures/keypoint.py#L97-L130), but which can probably be improved upon / generalized

torchvision/prototype/transforms/_geometry.py
```diff
@@ -22,7 +22,7 @@ def bounding_box(input: BoundingBox) -> BoundingBox:
 
 
 class Resize(Transform):
-    NO_OP_FEATURE_TYPES = {Label}
+    NO_OP_FEATURE_TYPES = {Label, KeyPoint}
```
fmassa, line 25: Same thing in here

torchvision/prototype/transforms/_geometry.py
```diff
@@ -86,7 +86,7 @@ def extra_repr(self) -> str:
 
 
 class Crop(Transform):
-    NO_OP_FEATURE_TYPES = {BoundingBox, Label}
+    NO_OP_FEATURE_TYPES = {BoundingBox, Label, KeyPoint}
```
fmassa, line 89: and in here

torchvision/prototype/features/_keypoint.py
```diff
@@ -0,0 +1,23 @@
+from typing import Dict, Any, Tuple, Optional, Sequence
+
+import torch
+
+from ._feature import Feature, DEFAULT
+
+
+class KeyPoint(Feature):
+    descriptions: Optional[Sequence[str]]
```
fmassa, line 9: I believe `Keypoint` should also know about the image sizes on which they are mapped to. Otherwise we won't be able to perform operations such as `flip`.

torchvision/prototype/datasets/_builtin/coco.py
```diff
@@ -129,6 +159,7 @@ def _decode_captions_ann(self, anns: List[Dict[str, Any]], image_meta: Dict[str,
     _ANN_DECODERS = OrderedDict(
         [
             ("instances", _decode_instances_anns),
+            ("person_keypoints", _decode_person_keypoints_anns),
```
fmassa, line 162: This is unrelated to this PR, but I would maybe revisit if we want to have dataset to have a varying return type depending on arguments passed to it.
It might be better to have different flavors of the dataset to be different dataset classes.

fmassa (commented):
Thanks for the PR!

I've made some comments, let me know what you think

torchvision/prototype/datasets/_builtin/coco.py
pmeier: Yeah, that is a design choice that we should address. cc @NicolasHug

torchvision/prototype/features/_keypoint.py
pmeier: Good point, will add that.

torchvision/prototype/transforms/_geometry.py
pmeier: We are in the process of revamping the prototype transforms and thus I'm not going to add transforms for the `KeyPoint`'s in this PR. Still, you got a point about the descriptions.

To achieve these automatic conversion, we would need to store some kind of symmetry information about the individual keypoints. In the example of persons or faces we only have horizontal symmetry, but I don't think it is far fetched to also consider other symmetries such as vertical. Thus, I'm hesitant to add meta data to the feature that is only valid for a single transformation. 

Is the order of the keypoints important for further processing? If not, maybe we can take the easy way out and simply invalidate the descriptions in transformations like `horizontal_flip`?

fmassa: Yes, the order of the keypoints is super important. Indeed, we train models to predict left-right elements, and their order in the vector indicates if they are left-right.

If we don't carry this information properly, it will be impossible to apply those transformations to the object, or they will be wrong.

And I strongly believe we should take transformations into account when defining what should be in the Feature. This is one of the main reasons why we have Feature to begin with.

I believe we should add enough metadata that geometric transformations are well-defined. `image_size` is an easy case, and we do take this information into account as well in the `(x0, y0, x1, y1)` bounding-box world (we know that 3rd element is x1, and this changes how we perform flip).

pmeier:
In response to I've added `symmetries` meta data to the `KeyPoint` feature. With this each dataset specify what symmetries exists between the keypoints.

Here is an example with `HorizontalFlip`:

```py
from torchvision.utils import draw_keypoints

def draw_non_right_keypoints(sample, file):
    image = sample["image"]
    keypoints = sample["keypoints"]
    non_right_keypoints = keypoints[[not descr.startswith("right") for descr in keypoints.descriptions], :]

    annotated_image = draw_keypoints(
        image,
        non_right_keypoints.unsqueeze(0),
        radius=int(2e-2 * min(sample["image"].shape[-2:])),
        colors="red",
    )
    to_pil_image(annotated_image).save(file)


dataset = datasets.load("celeba")
sample = next(iter(dataset))

draw_non_right_keypoints(sample, "before.jpg")

transform = transforms.HorizontalFlip()
transformed_sample = transform(sample)

draw_non_right_keypoints(transformed_sample, "after.jpg")


```

before after

torchvision/prototype/transforms/_geometry.py
pmeier: I've pushed a new commit that handles this.

torchvision/prototype/transforms/_geometry.py
```diff
@@ -20,9 +20,25 @@ def bounding_box(input: BoundingBox) -> BoundingBox:
...
+                continue
+
+            idcs[first], idcs[second] = second, first
+
+        return Keypoint(coordinates[..., idcs, :], like=input)
+
 
 class Resize(Transform):
-    NO_OP_FEATURE_TYPES = {Label}
+    NO_OP_FEATURE_TYPES = {Label, Keypoint}
```
fmassa, line 41: Ok if we leave this as a TODO, but we should change the coordinates of the keypoints following the rescaling factor

torchvision/prototype/datasets/_builtin/celeba.py
```diff
@@ -128,23 +130,41 @@ def _collate_and_decode_sample(
...
         _, ann = ann_data
 
-        image = decoder(buffer) if decoder else buffer
+        if decoder:
+            image = decoder(buffer)
+            image_size = image.shape[-2:]
+        else:
+            with PIL.Image.open(buffer) as image:
+                image_size = image.height, image.width
+            buffer.seek(0)
```
fmassa, line 139: unrelated: why do we need this? Shouldn't this be the job of whatever is the `decoder`?

fmassa (commented):
The proposal for the symmetry LGTM, thanks!

I only have one more comment regarding the resize transform, otherwise good to merge

torchvision/prototype/transforms/_geometry.py
pmeier: Yes, in general all geometric transforms need to support `Keypoint`'s. We can probably share a lot of functionality with the bounding box kernels.

torchvision/prototype/datasets/_builtin/celeba.py
pmeier: We have a feature branch where the decoding is completely removed from datasets so this is just a crutch for the current state to have working examples. The point here is to get the image size without actually decoding the image.

torchvision/prototype/datasets/_builtin/coco.py
```diff
@@ -120,6 +120,45 @@ def _decode_instances_anns(self, anns: List[Dict[str, Any]], image_meta: Dict[st
...
+            "left_knee",
+            "right_knee",
+            "left_ankle",
+            "right_ankle",
+        )
+        symmetries = [
+            ("vertical", description, description.replace("left", "right"))
+            for description in descriptions
+            if description.startswith("left")
+        ]
```
datumbox, line 154: Could you clarify the type of the resulting `symmetries` variable?

Below they `Keypoint` expects `Sequence[Tuple[KeypointSymmetry, Tuple[int, int]]]`. It's non obvious to me how the current code has the specific type.

torchvision/prototype/datasets/_builtin/celeba.py
datumbox: I think the large number of concurrent and experimental PRs make reviewing hard and hinter clarity of direction. It might be worth dedicating some time to clean up (similar to #5377) and merging existing PRs prior sending more.

torchvision/prototype/features/_keypoint.py
```diff
@@ -0,0 +1,67 @@
+from typing import Dict, Any, Tuple, Optional, Sequence, Union, Collection
+
+import torch
+from torchvision.prototype.utils._internal import StrEnum
+
+from ._feature import Feature, DEFAULT
```
datumbox, line 6: What is `DEFAULT` here? From what I remember, this is currently missing from your latest proposal for the API at the [feature branch](https://github.com/pytorch/vision/tree/revamp-prototype-features-transforms). Coming into this work to review without much context, I"m very confused on what the end proposal is because we have simply too many open PRs with conflicting features. As mentioned above, we should dedicate time to clean up, close/merge PRs before opening new ones, prioritizing those that describe the skeleton of the API and deprioritizing those that add more features.

torchvision/prototype/features/_keypoint.py
```diff
@@ -0,0 +1,67 @@
...
+        if tensor.shape[-1] != 2:
+            raise ValueError
+        if tensor.ndim == 1:
+            tensor = tensor.view(1, -1)
+        return tensor
+
+    @classmethod
+    def _parse_meta_data(
+        cls,
+        image_size: Tuple[int, int] = DEFAULT,  # type: ignore[assignment]
```
datumbox, line 31: This is not a common idiom in TorchVision, as `None` is favoured. Do you plan to remove this?

datumbox (commented):
Thanks, I got some questions below

PR #5328, opened 2022-02-01
Updated gpu decoder test to use parametrize

prabhat00155:
Use parametrize instead of looping over the list. This would help point out the exact scenario in case of failures.


test/test_video_gpu_decoder.py
```diff
@@ -12,30 +12,33 @@
...
-            with av.open(full_path) as container:
-                for av_frame in container.decode(container.streams.video[0]):
-                    av_frames = torch.tensor(av_frame.to_rgb(src_colorspace="ITU709").to_ndarray())
-                    vision_frames = next(decoder)["data"]
-                    mean_delta = torch.mean(torch.abs(av_frames.float() - vision_frames.cpu().float()))
-                    assert mean_delta < 0.75
+    @pytest.mark.parametrize(
+        "full_path",
+        [
+            os.path.join(VIDEO_DIR, x)
```
NicolasHug, line 22: Nit: We could call that within the test function instead, and avoid a `for` loop / list comprehension within the parametrization

NicolasHug (approved):
Thanks @prabhat00155 , LGTM.

PR #5330, opened 2022-02-01
Graduate ConvNeXt to main TorchVision area

datumbox:
Moving ConvNeXt's key classes from prototype to main area. The prototype area remains only for the Multi-weights support API.


torchvision/models/convnext.py
```diff
@@ -0,0 +1,271 @@
...
+
+model_urls: Dict[str, Optional[str]] = {
+    "convnext_tiny": "https://download.pytorch.org/models/convnext_tiny-983f1562.pth",
+    "convnext_small": "https://download.pytorch.org/models/convnext_small-0c510722.pth",
+    "convnext_base": "https://download.pytorch.org/models/convnext_base-6075fbad.pth",
+    "convnext_large": "https://download.pytorch.org/models/convnext_large-ea097f82.pth",
+}
+
+
+class LayerNorm2d(nn.LayerNorm):
```
datumbox, line 31: Starting copy-pasted code.

torchvision/models/convnext.py
```diff
@@ -0,0 +1,271 @@
...
+                    nn.init.zeros_(m.bias)
+
+    def _forward_impl(self, x: Tensor) -> Tensor:
+        x = self.features(x)
+        x = self.avgpool(x)
+        x = self.classifier(x)
+        return x
+
+    def forward(self, x: Tensor) -> Tensor:
+        return self._forward_impl(x)
```
datumbox, line 186: End of copy-pasted code.

torchvision/models/convnext.py
```diff
@@ -0,0 +1,271 @@
...
+__all__ = [
+    "ConvNeXt",
+    "convnext_tiny",
+    "convnext_small",
+    "convnext_base",
+    "convnext_large",
+]
+
+
+model_urls: Dict[str, Optional[str]] = {
```
NicolasHug, line 23: nit
```suggestion
_MODELS_URLS: Dict[str, Optional[str]] = {
```

NicolasHug (approved):
Thanks @datumbox , I only have minor comments, LGTM

torchvision/models/convnext.py
```diff
@@ -0,0 +1,271 @@
...
+        model.load_state_dict(state_dict)
+    return model
+
+
+def convnext_tiny(pretrained: bool = False, progress: bool = True, **kwargs: Any) -> ConvNeXt:
+    r"""ConvNeXt Tiny model architecture from the
+    `"A ConvNet for the 2020s" <https://arxiv.org/abs/2201.03545>`_ paper.
+    Args:
+        pretrained (bool): If True, returns a model pre-trained on ImageNet
+        progress (bool): If True, displays a progress bar of the download to stderr
```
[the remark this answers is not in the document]
datumbox: Hmm... Nowhere else on the `models` module we use document `kwargs`. Perhaps that's something worth considering on the new documentation? I think it might be worth chatting about how we will use it going forwards (previously we pushed parameters that are specific to the model and that kept the model builder signatures uniform, though I'm not sure why this is good).

torchvision/models/convnext.py
```diff
@@ -0,0 +1,271 @@
...
+    model = ConvNeXt(block_setting, stochastic_depth_prob=stochastic_depth_prob, **kwargs)
+    if pretrained:
+        if arch not in model_urls:
+            raise ValueError(f"No checkpoint is available for model type {arch}")
+        state_dict = load_state_dict_from_url(model_urls[arch], progress=progress)
+        model.load_state_dict(state_dict)
+    return model
+
+
+def convnext_tiny(pretrained: bool = False, progress: bool = True, **kwargs: Any) -> ConvNeXt:
```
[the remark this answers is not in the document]
datumbox: SGTM. We could do the same for ALL new model builders we introduced on this version to reinforce that this is coming on the next versions of TorchVision. If you want that, we can create a ticket with all new builders and do the change prior the release.

PR #5331, opened 2022-02-01
Move all weight initializations from private methods to constructors

datumbox:
The main idiom for weight initialisation in TorchVision is to do it within constructors. Nevertheless, there is limited use of another idiom of placing the initialization in a separate private method and calling it from constructor. Though the two can be considered equivalent (the method is private after all), the use of the latter idiom can lead to gotchas.

It's worth noting that TorchVision users have the appetite for being able to re-initialize the weights (see #3410) and thus some might be tempted to call the private method directly. Unfortunately having such a method currently stumbles upon fundamental [FX limitations](https://github.com/pytorch/pytorch/issues/66335) and can lead to [inheritance issues](https://github.com/pytorch/pytorch/issues/71404#issuecomment-1026011644). Thus there are cases (for example after using TorchVision's [Feature Extraction util](https://github.com/pytorch/vision/blob/main/torchvision/models/feature_extraction.py)) where if one calls the private method to reinitialize the model, the code won't work or even break. By moving the initialization within the constructor, we make it impossible for someone to shoot themselves in the foot. Until proper support is added for this feature in PyTorch (see https://github.com/pytorch/pytorch/issues/71404), TorchVision should use idioms that minimize the chance of gotchas.

### DIFF
--- a/torchvision/models/googlenet.py
+++ b/torchvision/models/googlenet.py
@@ -90,15 +90,12 @@ class GoogLeNet(nn.Module):
         self.fc = nn.Linear(1024, num_classes)
 
         if init_weights:
-            self._initialize_weights()
-
-    def _initialize_weights(self) -> None:
-        for m in self.modules():
-            if isinstance(m, nn.Conv2d) or isinstance(m, nn.Linear):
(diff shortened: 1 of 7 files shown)


fmassa (approved):
LGTM!

PR #5335, opened 2022-02-02
add HMDB51 and UCF101 datasets as well as prototype for new style video decoding

pmeier:
This adds HMDB51 as first video dataset (see #4541). It is used for prototyping video decoder datapipes initially implemented by @bjuncek in #4838. You can test with the following snippet

```py

dataset = datasets.load("hmdb51")
dataset = datasets.utils.KeyframeDecoder(dataset)

for sample in dataset:
    print(sample)
    break
```

Apart from that the biggest change is the added `meta` dictionary attribute to the `EncodedData` feature. The rationale here is that each dataset might provide very different meta data for each file and this is hard to standardize. We might be able to have some common attributes like `path`, but I would still leave the option open for arbitrary meta data.


pmeier:
@bjuncek I've also ported UCF101 from #4848 so you can close it and take over this PR.

torchvision/prototype/datasets/_builtin/hmdb51.py
```diff
@@ -0,0 +1,116 @@
...
+
+        splits = HttpResource(
+            f"{url_root}/test_train_splits.rar",
+            sha256="229c94f845720d01eb3946d39f39292ea962d50a18136484aa47c1eba251d2b7",
+        )
+        videos = HttpResource(
+            f"{url_root}/hmdb51_org.rar",
+            sha256="9e714a0d8b76104d76e932764a7ca636f929fff66279cda3f2e326fa912a328e",
+        )
+        videos._preprocess = self._extract_videos_archive
```
pmeier, line 54: @NicolasHug The archive is a rar of rars so using a single `extract=True` won't cut it. We need the full extraction since reading from rar archives is rather slow and with this we get a significant performance increase.

Another option would be to use this "recursive extraction" by default when setting `extract=True`.

bjuncek:
@pmeier I see that you closed this; should I switch to the new PR?

pmeier:
For whatever reason, GitHub does not let me re-open this. I've send the PR again in #5422.

datumbox:
@pmeier It's because the revamp branch is deleted.

pmeier:
Yeah, I figured that, but it doesn't let me select a new merge target like `main` either. This is possible for open PRs.

Issue #5281, opened 2022-01-26
Potential improvements to the prototype dataset implementation #5269 


NicolasHug:
Keeping track of a meeting we just had with @pmeier , here are a few things we can try:

- Instead of having a separate `DatasetInfo` class, we could try having all these fields be part of the `Dataset` class itself. Then `_make_info()` would become `Dataset.__init__()`, and we could expose an `info()` method / property which would expose the relevant fields to the user (currently, this is the `info` attribute).
- The current logic for defining valid parameters  and their defaults is a bit unconventional (though the `DatasetInfo` class), and potentially limiting. This issue might resolve itself by tackling the one above
- The content of the `extra` field could maybe be exposed directly at the `DatasetInfo` level (or if we pursue with the first item above, exposed at the `Dataset` level).
- We might want to allow a `root=None` parameter to the `datasets.load()` so users can set a custom `home`, without having to go through the `home()` global function
- Dataset registration logic: currently this is done somewhat magically by adding the dataset class to `__init__.py` file. Perhaps we could do that with a more explicit registration mechanism (e.g. a decorator on the class or something)
- The decoder parameter will eventually be removed from `load()`, and from the `_make_datapipe()` method
- The current default of `skip_integrity_check` is False, but it doesn't exactly reflect the actual behaviour: we still skip the integrity checks if the data is already downloaded (because the check can take a while on big archives). Perhaps we can try to find an API that accurately describes this behaviour.
- When we're ready to "graduate" the prototypes, we should review all of the private / public decisions that were made so far
- Similarly we should review, document and test the utilities in `folder.py`

cc @pmeier @bjuncek

PR #5369, opened 2022-02-03
merge DatasetInfo into Dataset

pmeier:
Addresses Points 1. to 3. from #5281.


test/builtin_dataset_mocks.py
```diff
@@ -31,12 +32,20 @@
...
+        self.name = self.dataset.name
 
         self.mock_data_fn = mock_data_fn
-        self.configs = self.info._configs
+
+        self.configs = []
+        for combination in itertools.product(*[option.valid for option in self.dataset.options]):
+            options = dict(zip([option.name for option in self.dataset.options], combination))
+            with contextlib.suppress(Exception):
+                self.configs.append(self.dataset.make_config(**options))
```
pmeier, line 48: This is the only part of the test suite changes that needs reviewing. All other changes are just hacks to temporary only test VOC and ImageNet.

torchvision/prototype/datasets/utils/_dataset.py
```diff
@@ -62,50 +87,31 @@ def __init__(
...
-            if arg not in valid_args:
-                raise ValueError(
-                    add_suggestion(
-                        f"Invalid argument '{arg}' for option '{name}' of dataset {self.name}.",
-                        word=arg,
-                        possibilities=valid_args,
-                    )
-                )
+            if arg not in option.valid:
+                raise ValueError(f"Invalid argument '{arg}' for option '{name}' of dataset '{self.name}'. {option.doc}")
```
pmeier, line 114: Maybe we could also have this logic on the `DatasetOption` class.

pmeier (commented):
I only implemented the new design for VOC and ImageNet. If we agree on the changes, I'll address the remaining datasets.

torchvision/prototype/datasets/utils/_dataset.py
```diff
@@ -23,29 +21,56 @@ class DatasetType(enum.Enum):
...
     # This needs to be Frozen because we often pass configs as partial(func, config=config)
     # and partial() requires the parameters to be hashable.
     pass
 
 
-class DatasetInfo:
+SENTINEL = object()
+
+
+class DatasetOption:
```
NicolasHug, line 38: What is the reason for having a `DatasetOption` class instead of relying on standard Python parameter passing + docstrings?

I'm not sure we should try to automate input validation or docstring generation at all cost TBH. If it forces us to deviate from our usual Python development practices, it may not be worth it?

torchvision/prototype/datasets/utils/_dataset.py
```diff
@@ -62,50 +87,31 @@ def __init__(
...
-    def default_config(self) -> DatasetConfig:
-        return self._configs[0]
-
-    @staticmethod
-    def read_categories_file(path: pathlib.Path) -> List[List[str]]:
+    def read_categories_file(self, path: pathlib.Path) -> List[List[str]]:
         with open(path, newline="") as file:
             return [row for row in csv.reader(file)]
 
     def make_config(self, **options: Any) -> DatasetConfig:
```
NicolasHug, line 94: What was the original reason for doing the config creation / validation separately from `__init__()`?

torchvision/prototype/datasets/utils/_dataset.py
pmeier: Right now, we instantiate all `Dataset` objects at import time

https://github.com/pytorch/vision/blob/44e93beb0833ff51ffe75d994fe660434335b53d/torchvision/prototype/datasets/_api.py#L21-L23

At this stage no specific configuration is known. So something like

```py
class MyDataset(Dataset):
    def __init__(self, foo="foo", bar=1):
        if not isinstance(bar, int):
            raise TypeError
```

is not possible. 

Now, we can of course change the API to only instantiate the objects at runtime. But then we also need a different way to access the static data such as `categories` or the like, since they should be accessible without a specific configuration.

torchvision/prototype/datasets/utils/_dataset.py
pmeier: I've explained in https://github.com/pytorch/vision/pull/5369#discussion_r798679279. Let's keep the discussion there.

torchvision/prototype/datasets/utils/_dataset.py
NicolasHug: > But then we also need a different way to access the static data such as categories or the like, since they should be accessible without a specific configuration.

Would it be reasonable to expect all parameters to have a default? This would allow us to instanciate the dataset objects whenever we need their static data.

Also, about the need for accessing meta-data without a config: perhaps we don't have that yet, but I could see a dataset that has different classes (or different "extra" field) depending on a parameter. Are we supporting this use-case?

pmeier: > Would it be reasonable to expect all parameters to have a default?

IMO, yes. We shouldn't have a dataset that is not loadable with `datasets.load("foo")`. Looking through our old datasets, I think there two different scenarios where this violated:

1. Datasets that cannot be automatically downloaded and need multiple files as input

    https://github.com/pytorch/vision/blob/44e93beb0833ff51ffe75d994fe660434335b53d/torchvision/datasets/coco.py#L28

2. Datasets with arguments that could very well have a default but don't for an probably unknown reason

    https://github.com/pytorch/vision/blob/44e93beb0833ff51ffe75d994fe660434335b53d/torchvision/datasets/mnist.py#L293

Since the download is handled differently and we no longer need to pass paths to the dataset, I currently don't see a reason to not assume it. If there ever is a case where we actually have no good default value, we can probably hack around it.

> Also, about the need for accessing meta-data without a config: perhaps we don't have that yet, but I could see a dataset that has different classes (or different "extra" field) depending on a parameter. Are we supporting this use-case?

Currently this is not supported. But I'm not sure what would be a good way to add this. With the design discussed in this thread, `datasets.info()` would give you the `Dataset` object instantiated with default parameters. Thus, we would always get the same meta data back. One way would be to accept `**options` there.

NicolasHug: Perhaps it would help to see how we would be documenting these new datasets, before we can make a decision on this.

I'm sceptical on general about not using normal parameter passing, defaults, and docs. Especially if we ever allow users to create their own classes, in we don't want them to learn yet another framework (but it's relevant for developers too). If it makes the docs much easier to write them we might have a reason, but for now it does not appear as obvious to me

pmeier: Let me write minimalistic docs to showcase.

pmeier: See https://1173984-73328905-gh.circle-artifacts.com/0/docs/datasets_prototype/index.html. Of course if we had more meta information filled in such as the citation, license, description, and so on, the pages would look not as empty.

NicolasHug: Thanks a lot @pmeier .

I have a few questions:

- how would we document the type of the parameters?
- how would we document extra methods on the datasets (if relevant)? What about extra attributes (again not sure if relevant)?
- how would we document additional info that is specific to each dataset? Stuff like "This dataset contains X images of cars and Y images of trucks".

pmeier: > how would we document the type of the parameters?

I would put another column in the table for the type. `DatasetOption` could have an `annotation` field for that. Let me add it to the examples.

> how would we document extra methods on the datasets (if relevant)?

IMO, we shouldn't have extra methods on the dataset objects. Currently, I don't think we have datasets that need this. Did you have anything in mind what other datasets might provide?

> What about extra attributes (again not sure if relevant)?

In our current stable API we sometimes add them to the docstring in the `Attributes:` section. We could simply have another table similar to `Options` that holds these. Let me add them to my examples.

> how would we document additional info that is specific to each dataset? Stuff like "This dataset contains X images of cars and Y images of trucks".

Given that this is very individual, I would simply have `description` field which takes a string input and puts it at the top of the generated documentation. Let me add an example.

pmeier: See https://1181665-73328905-gh.circle-artifacts.com/0/docs/datasets_prototype/index.html. I've taken the examples from https://paperswithcode.com/ for demo purposes.

pmeier:
Superseded by the development on the https://github.com/pytorch/vision/tree/prototype-datasets-inheritance branch.

PR #5375, opened 2022-02-04
[NOMERGE] Feedback Transforms API implementation

datumbox:

### DIFF
--- a/torchvision/prototype/transforms/functional/_geometry.py
+++ b/torchvision/prototype/transforms/functional/_geometry.py
@@ -3,7 +3,7 @@ from typing import Tuple, List, Optional
 import torch
 from torchvision.prototype.features import BoundingBoxFormat
 from torchvision.transforms import (  # noqa: F401
-    functional as _F,
+    functional as _F,  # Shouldn't we importing `functional_tensor`?
     InterpolationMode,
 )
 
@@ -11,7 +11,7 @@ from ._meta_conversion import convert_bounding_box_format
 
 
 def horizontal_flip_image(image: torch.Tensor) -> torch.Tensor:
-    return image.flip((-1,))
+    return image.flip((-1,))  # Why not use the _F.hflip()?
 
 
 def horizontal_flip_bounding_box(bounding_box: torch.Tensor, *, image_size: Tuple[int, int]) -> torch.Tensor:
@@ -20,7 +20,7 @@ def horizontal_flip_bounding_box(bounding_box: torch.Tensor, *, image_size: Tupl
         old_format=BoundingBoxFormat.XYXY,
         new_format=BoundingBoxFormat.XYWH,
     ).unbind(-1)
-    x = image_size[1] - (x + w)
+    x = image_size[1] - (x + w)  # Why not avoid formatting and do straight `boxes[:, [0, 2]] = width - boxes[:, [2, 0]]`?
     return convert_bounding_box_format(
         torch.stack((x, y, w, h), dim=-1),
         old_format=BoundingBoxFormat.XYWH,
@@ -28,7 +28,7 @@ def horizontal_flip_bounding_box(bounding_box: torch.Tensor, *, image_size: Tupl
     )
 
 
-_resize_image = _F.resize
+_resize_image = _F.resize  # How about videos? The low level transfroms supports it
 
 
 def resize_image(
@@ -40,7 +40,7 @@ def resize_image(
 ) -> torch.Tensor:
     new_height, new_width = size
     num_channels, old_height, old_width = image.shape[-3:]
-    batch_shape = image.shape[:-3]
+    batch_shape = image.shape[:-3]  # In some places you use `image` to denote batches and in others `image_batch`. Should we align the names?
     return _resize_image(
         image.reshape((-1, num_channels, old_height, old_width)),
         size=size,


torchvision/prototype/transforms/_geometry.py
```diff
@@ -7,6 +7,10 @@
 from torchvision.prototype.transforms import Transform
 
 
+# I recommend deleting all classes from this branch that are old to assist code reviews.
```
pmeier, line 10: Will do.

torchvision/prototype/transforms/functional/_augment.py
```diff
@@ -24,6 +24,7 @@ def mixup_one_hot_label(one_hot_label_batch: torch.Tensor, *, lam: float, inplac
 
 
 def cutmix_image(image: torch.Tensor, *, box: Tuple[int, int, int, int], inplace: bool = False) -> torch.Tensor:
+    # Shouldn't be we using `image_batch` similar to `image_batch`, to indicate that the input must be a batch?
```
pmeier: Yes, we can rename this. Note, that I also left out any input validation here for simplicity. Let me add a minimalistic one to help reviewers.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -3,15 +3,15 @@
...
+    functional as _F,  # Shouldn't we importing `functional_tensor`?
     InterpolationMode,
 )
 
 from ._meta_conversion import convert_bounding_box_format
 
 
 def horizontal_flip_image(image: torch.Tensor) -> torch.Tensor:
-    return image.flip((-1,))
+    return image.flip((-1,))  # Why not use the _F.hflip()?
```
pmeier, line 14: Just a remnant of an old implementation. Will fix.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -20,15 +20,15 @@ def horizontal_flip_bounding_box(bounding_box: torch.Tensor, *, image_size: Tupl
         old_format=BoundingBoxFormat.XYXY,
         new_format=BoundingBoxFormat.XYWH,
     ).unbind(-1)
-    x = image_size[1] - (x + w)
+    x = image_size[1] - (x + w)  # Why not avoid formatting and do straight `boxes[:, [0, 2]] = width - boxes[:, [2, 0]]`?
```
pmeier, line 23: That would modify inplace so we need to do `boxes = boxes.clone()` first. Otherwise I have no preference over one or the other.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -20,15 +20,15 @@ def horizontal_flip_bounding_box(bounding_box: torch.Tensor, *, image_size: Tupl
...
     return convert_bounding_box_format(
         torch.stack((x, y, w, h), dim=-1),
         old_format=BoundingBoxFormat.XYWH,
         new_format=BoundingBoxFormat.XYXY,
     )
 
 
-_resize_image = _F.resize
+_resize_image = _F.resize  # How about videos? The low level transfroms supports it
```
pmeier, line 31: Right now there is custom feature type for videos and the tentatively plan is to treat videos as `Image`'s with an extra batch dimensions cc @bjuncek. Since the kernel should be agnostic to batch dimensions, they are automatically supported by all image transforms.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -40,7 +40,7 @@ def resize_image(
 ) -> torch.Tensor:
     new_height, new_width = size
     num_channels, old_height, old_width = image.shape[-3:]
-    batch_shape = image.shape[:-3]
+    batch_shape = image.shape[:-3]  # In some places you use `image` to denote batches and in others `image_batch`. Should we align the names?
```
pmeier, line 43: The difference is that `image` is batch size agnostic whereas `image_batch` requires at least one batch dimension. Note that `batch_shape` here can very well be `()` if it is a single image without batch dimensions.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -71,7 +71,7 @@ def resize_bounding_box(
 ) -> torch.Tensor:
     old_height, old_width = old_image_size
     new_height, new_width = new_image_size
-    return (
+    return (  # Shouldn't we be using a low-level kernel instead, similar to above?
```
pmeier, line 74: Not sure I understand. This is the low level kernel. I'm not aware of a resize kernel for bounding boxes that I could import from anywhere.

torchvision/prototype/transforms/functional/_meta_conversion.py
```diff
@@ -4,7 +4,7 @@
 
 
 def _xywh_to_xyxy(xywh: torch.Tensor) -> torch.Tensor:
-    xyxy = xywh.clone()
+    xyxy = xywh.clone()  # Nice. Do we have tests to ensure that no transform does an in-place modification of input?
```
pmeier, line 7: We currently have very minimal test coverage for the transforms. I'll expand them when the design is more stable.

torchvision/prototype/transforms/functional/_augment.py
datumbox: Makes sense.

torchvision/prototype/transforms/functional/_geometry.py
datumbox: Yeah, I'm not advocating inplace modifications of the input. Just a simplification of the transform.

torchvision/prototype/transforms/functional/_geometry.py
datumbox: We should finalize this plan soon. We got outstanding work with Videos so we need to be able to have an idea of how this works for them. Once the plan is finalized, we will be able to parallelize the work of porting all transforms concurrently.

I agree about the underlying kernels being agnostic, sounds good.

pmeier: For reference, the initial support for videos happens in #5335. @bjuncek is checking with other researchers working on videos if the API is sufficient or if we are missing something important.

torchvision/prototype/transforms/functional/_meta_conversion.py
datumbox: This is not for now. It's just a note that we should eventually have a test that checks that no Transform modifies the original input that we pass. Agreed to keep test-coverage minimal at the moment until we have a stable design.

datumbox:
Spoke offline with @pmeier and I'm going to close this. The comments will be addressed in a follow up PR.

torchvision/prototype/transforms/functional/_geometry.py
datumbox: There is an existing method:
https://github.com/pytorch/vision/blob/dad6e6a44d63272459b54175c92ed22955eb4207/torchvision/models/detection/transform.py#L289-L302

But I agree, probably we want to rewrite it here.

torchvision/prototype/transforms/functional/_geometry.py
bjuncek: Annoyingly, I still haven't gotten a reply from any of them.
@datumbox , maybe you could help me offline in order to make this discussion happen?

> tentatively plan is to treat videos as Image's with an extra batch dimensions

Is there a downside to creating a `video` feature class that for the time being implement just this, but would be possible to extend with stuff like audio or cc-data or encoded bytes if support for this will be needed?

datumbox: @bjuncek ping me offline with the names of the people you try to talk to to see if I can help.

When we say "extra branch dimensions", could you clarify the proposal?

pmeier: I don't think there is real downside other than being more verbose in our transforms implementations. For example, we would need to have `resize_video` that simply defers to `resize_image`. This has to happen for every single image / video kernel.

PR #5377, opened 2022-02-04
fix prototype features and functional transforms after review

pmeier:
Addresses #5375 and partially #5379.


torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -1,31 +1,18 @@
...
     functional as _F,
     InterpolationMode,
 )
 
-from ._meta_conversion import convert_bounding_box_format
-
-
-def horizontal_flip_image(image: torch.Tensor) -> torch.Tensor:
-    return image.flip((-1,))
+horizontal_flip_image = _F.hflip
```
datumbox, line 9: Should we be importing _F from functional or from functional_tensor?

NicolasHug:
image

image

datumbox (approved):
LGTM, thanks for cleaning up the stale code.

Only 1 question below:

torchvision/prototype/transforms/functional/_geometry.py
pmeier: Somehow my comment got swallowed. @vfdev-5 asked me to do the exact opposite in https://github.com/pytorch/vision/pull/5295#discussion_r793622248. I don't have any preference over one or the other. Is there a difference if we ever only use the them with tensors?

datumbox: Lol. That's what happens we you get too many cooks in one kitchen 😄 Let's not block this, feel free to merge and address later.

@vfdev-5 The new API doesn't support PIL and thus the Tensor API in the kernel we want (we also avoid redirections). Perhaps I'm missing something?

pmeier: As discussed offline, we can use `functional_tensor` here to avoid the extra redirection.

pmeier: After looking into this, I think we should stick with `transforms.functional`:

- The signature is not necessarily the same. For example `functional.resize` takes 
     https://github.com/pytorch/vision/blob/dad6e6a44d63272459b54175c92ed22955eb4207/torchvision/transforms/functional.py#L370

    while `functional_tensor.resize` takes

    https://github.com/pytorch/vision/blob/dad6e6a44d63272459b54175c92ed22955eb4207/torchvision/transforms/functional_tensor.py#L489
- Composite transforms such as `functional.resized_crop` have no equivalence in `functional_tensor`, since the primitives in `functional` already handle the dispatch

    https://github.com/pytorch/vision/blob/dad6e6a44d63272459b54175c92ed22955eb4207/torchvision/transforms/functional.py#L594-L598
- If the transform has only a tensor implementation, for example `normalize` and `erase`, the implementation lives in `functional` rather than `functional_tensor`

    https://github.com/pytorch/vision/blob/dad6e6a44d63272459b54175c92ed22955eb4207/torchvision/transforms/functional.py#L1258
    https://github.com/pytorch/vision/blob/dad6e6a44d63272459b54175c92ed22955eb4207/torchvision/transforms/functional.py#L320

datumbox: Sounds good. As discussed offline, we should later fixe all these issues listed above, so that we can remove this intermediate layer.

Issue #2583, opened 2020-08-13
Make transforms work on video tensor inputs or batch of images


vfdev-5:
## 🚀 Feature

Following https://github.com/pytorch/vision/issues/2292#issuecomment-671325017 and discussion with @fmassa and @bjuncek , this feature request is to improve the transforms module to support video inputs as `torch.Tensor` of shape `(C, T, H, W)`, where C - number of image channels (e.g. 3 for RGB), T - number of frames, H, W - image dimensions. 

Points to discuss:
- Convention for geometric transforms: 2 last dimensions are H, W ?
- Convention for color transforms: 1 dimension is color, like above `(C, T, H, W)` ?
  - should we also support `(T, C, H, W)` ?

joakimjohnander:
The representation `(T, C, H, W)` is convenient when processing one frame at a time, such as in object tracking.

fmassa:
@vfdev-5 I believe this can be closed now?

vfdev-5:
@fmassa I thought to close it once we updated video classification ref example. But as you wish, otherwise I can create another one for that update.

fmassa:
Ok, let's wait until we update the reference examples then before closing this one

vfdev-5:
Close the issue as #2935 has been merged.

PR #5379, opened 2022-02-04
[NOMERGE] Review Features prototype

datumbox:
Adding comments to the features classes.

### DIFF
--- a/torchvision/prototype/features/__init__.py
+++ b/torchvision/prototype/features/__init__.py
@@ -4,3 +4,5 @@ from ._feature import Feature
 from ._image import ColorSpace, Image
 from ._label import Label, OneHotLabel
 from ._segmentation_mask import SegmentationMask
+
+# We put lots of effort on Video this half. We will need to figure out video tensors as well in this prototype
--- a/torchvision/prototype/features/_bounding_box.py
+++ b/torchvision/prototype/features/_bounding_box.py
@@ -15,7 +15,7 @@ class BoundingBoxFormat(StrEnum):
 
 
 class BoundingBox(Feature):
-    formats = BoundingBoxFormat
+    formats = BoundingBoxFormat  # Couldn't find a use of this in code. Is there a reason why we don't just let people access the enums directly?
     format: BoundingBoxFormat
     image_size: Tuple[int, int]
 
@@ -40,6 +40,9 @@ class BoundingBox(Feature):
     def to_format(self, format: Union[str, BoundingBoxFormat]) -> "BoundingBox":
         # import at runtime to avoid cyclic imports
         from torchvision.prototype.transforms.functional import convert_bounding_box_format
+        # I think we can avoid this by not having a `to_format` method but instead require users to explicitly call the
+        # convert method. As far as I see, the specific method is used only once on the code, so it is something we
+        # could avoid all together.
 
         if isinstance(format, str):
             format = BoundingBoxFormat[format]
--- a/torchvision/prototype/features/_image.py
+++ b/torchvision/prototype/features/_image.py
@@ -13,6 +13,7 @@ from ._feature import Feature
 
 class ColorSpace(StrEnum):
     # this is just for test purposes
+    # How about the transparency spaces supported by ImageReadMode?
     _SENTINEL = -1
     OTHER = 0
     GRAYSCALE = 1
@@ -77,7 +78,9 @@ class Image(Feature):
             return ColorSpace.OTHER
 
     def show(self) -> None:
+        # This is a nice to have, but not a necessary method, for this early in the prototype
         to_pil_image(make_grid(self.view(-1, *self.shape[-3:]))).show()
 
     def draw_bounding_box(self, bounding_box: BoundingBox, **kwargs: Any) -> "Image":
+        # Same as above and nothing that this is the only method that requires to_format().
         return Image.new_like(self, draw_bounding_boxes(self, bounding_box.to_format("xyxy").view(-1, 4), **kwargs))
--- a/torchvision/prototype/features/_label.py
+++ b/torchvision/prototype/features/_label.py
@@ -15,7 +15,7 @@ class Label(Feature):
         *,
         dtype: Optional[torch.dtype] = None,
         device: Optional[torch.device] = None,
-        like: Optional["Label"] = None,
+        like: Optional["Label"] = None,  # Since are at Py3.7, perhaps we could do `from __future__ import annotations` now.
         categories: Optional[Sequence[str]] = None,
     ):
         label = super().__new__(cls, data, dtype=dtype, device=device)
@@ -26,7 +26,7 @@ class Label(Feature):
 
     @classmethod
     def from_category(cls, category: str, *, categories: Sequence[str]):
-        categories = list(categories)
+        categories = list(categories)  # why shallow copy here? If this method is in a loop, we run the risk of creating many shallow-copies
         return cls(categories.index(category), categories=categories)
 
     def to_categories(self):
--- a/torchvision/prototype/features/_encoded.py
+++ b/torchvision/prototype/features/_encoded.py
@@ -41,6 +41,7 @@ class EncodedImage(EncodedData):
     def decode(self) -> Image:
         # import at runtime to avoid cyclic imports
         from torchvision.prototype.transforms.functional import decode_image_with_pil
+        # Same commens as on the BoundingBox.to_format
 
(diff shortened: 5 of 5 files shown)


torchvision/prototype/features/_bounding_box.py
```diff
@@ -15,7 +15,7 @@ class BoundingBoxFormat(StrEnum):
 
 
 class BoundingBox(Feature):
-    formats = BoundingBoxFormat
+    formats = BoundingBoxFormat  # Couldn't find a use of this in code. Is there a reason why we don't just let people access the enums directly?
```
pmeier, line 18: My thought was just convenience. I thought it would be more convenient to have something like

```py
from torchvision.prototype.features import BoundingBox

box = BoundingBox(..., format=BoundingBox.formats.XYXY)
```

rather than

```py
from torchvision.prototype.features import BoundingBox, BoundingBoxFormat

box = BoundingBox(..., format=BoundingBoxFormat.XYXY)
```

But since I never even used it myself and we also accept strings, e.g. `format="xyxy"`, we can probably remove it.

torchvision/prototype/features/_bounding_box.py
```diff
@@ -40,6 +40,9 @@ def __new__(
     def to_format(self, format: Union[str, BoundingBoxFormat]) -> "BoundingBox":
         # import at runtime to avoid cyclic imports
         from torchvision.prototype.transforms.functional import convert_bounding_box_format
+        # I think we can avoid this by not having a `to_format` method but instead require users to explicitly call the
```
pmeier, line 43: This is for convenience and I expect the most usages in application code that we don't provide. This is similar to `Tensor.to()` to change the device or dtype. It would be rather inconvenient to do something like `torch.to(Tensor())` all the time.

Still, I agree with the avoidance of lazy imports here since they point to a design problem. This was also observed in https://github.com/pytorch/vision/pull/5295#discussion_r793619498. My solution would be to move the conversion implementation here and only expose it through this method as well through a transform.

torchvision/prototype/features/_image.py
```diff
@@ -13,6 +13,7 @@
 
 class ColorSpace(StrEnum):
     # this is just for test purposes
+    # How about the transparency spaces supported by ImageReadMode?
```
pmeier, line 16: This enum are not complete yet and should be expanded as we mature the prototype.

torchvision/prototype/features/_image.py
```diff
@@ -77,7 +78,9 @@ def guess_color_space(data: torch.Tensor) -> ColorSpace:
             return ColorSpace.OTHER
 
     def show(self) -> None:
+        # This is a nice to have, but not a necessary method, for this early in the prototype
```
pmeier, line 81: Necessary no, but really helpful for debugging the dataset prototypes. Given that this is just one line, I would keep it in instead of adding it later unless you are strongly against it.

torchvision/prototype/features/_image.py
```diff
@@ -77,7 +78,9 @@ def guess_color_space(data: torch.Tensor) -> ColorSpace:
...
         to_pil_image(make_grid(self.view(-1, *self.shape[-3:]))).show()
 
     def draw_bounding_box(self, bounding_box: BoundingBox, **kwargs: Any) -> "Image":
+        # Same as above and nothing that this is the only method that requires to_format().
```
pmeier, line 85: Even without the method call, this is really helpful while debugging a dataset implementation to see at a glance if the bounding boxes are properly set up.

torchvision/prototype/features/_label.py
```diff
@@ -15,7 +15,7 @@ def __new__(
         *,
         dtype: Optional[torch.dtype] = None,
         device: Optional[torch.device] = None,
-        like: Optional["Label"] = None,
+        like: Optional["Label"] = None,  # Since are at Py3.7, perhaps we could do `from __future__ import annotations` now.
```
pmeier, line 18: :100: Thanks, I wasn't aware this was a possibility. Will change.

torchvision/prototype/features/_label.py
```diff
@@ -26,7 +26,7 @@ def __new__(
 
     @classmethod
     def from_category(cls, category: str, *, categories: Sequence[str]):
-        categories = list(categories)
+        categories = list(categories)  # why shallow copy here? If this method is in a loop, we run the risk of creating many shallow-copies
```
pmeier, line 29: When writing that, I wasn't sure all `Sequence`'s supported the `.index()` method, but I forgot to check. Since [that seems to be the case](https://docs.python.org/3/library/collections.abc.html?highlight=sequence#collections.abc.Sequence), there is no need for the conversion to a list.

torchvision/prototype/features/__init__.py
```diff
@@ -4,3 +4,5 @@
 from ._image import ColorSpace, Image
 from ._label import Label, OneHotLabel
 from ._segmentation_mask import SegmentationMask
+
+# We put lots of effort on Video this half. We will need to figure out video tensors as well in this prototype
```
pmeier, line 8: The question that we need to answer is if there are video transforms that work with the temporal information. This would require the video feature that always have at least 4 dimensions like `(time, channels, height, width)`. If this is not required, we can simply treat a video as a batch of images and thus use the `Image` feature for it. cc @bjuncek

torchvision/prototype/features/_bounding_box.py
datumbox: Let's omit for now any convenience methods if they cause such issues and we can revise once the API is complete.

torchvision/prototype/features/_image.py
datumbox: If it assists your development it's fine to keep. See my above comment for all other cases of nice-to-haves that cause issues.

torchvision/prototype/features/__init__.py
datumbox: It depends. Some video-specific transforms can potentially use the temporal dimension but for now those that we support don't. So the API should limit us from supporting such transforms but the low-level kernels must support Videos (I believe that's the case now for the vast majority of them; worth confirming). 

Note that Videos will have 5 dimensions if you include the batch. This is why many tensor low-level kernels currently implement the transforms using negative indexes (-1, -2, -3) for C, H, W. It is on the current roadmap of TorchVision to improve video support, so that's something we need to factor in.

cc @vfdev-5 for visibility on which transforms need to be adjusted to support videos.

vfdev-5: I can link this issue https://github.com/pytorch/vision/issues/2583 where it was an attempt to make transforms to support video data.

bjuncek: > This would require the video feature that always have at least 4 dimensions

In principle, it's not _outlandish_ to expect video tensors to have (at least) 4 dimensions, with fixed expectation for C,H,W being the last three in order to avoid having to re-implement things with negative indexes.


> It depends. Some video-specific transforms can potentially use the temporal dimension but for now those that we support don't. So the API should limit us from supporting such transforms but the low-level kernels must support Videos

I'm not sure I follow. There are video transforms (temporal jittering, subsampling, ...) that we should _be able_ to support. In principle, I wouldn't expect torchvision to support every possible scenario, but it would be helpful if basic transforms for video worked on an arbitrary dimensional (i.e. `[..., C, H, W]` ) tensor. In that way, if user wants to implement some obscure temporal transform, they can implement something operating on the T dimension (so, -4) in a way that follows common API for a video feature (or batch image feature, whatever will be easiest). Would the current implementation make that troublesome?

datumbox: @bjuncek I think we agree on the approach. Having the Videos stored as `[..., T, C, H, W]` allows us to implement most of the basic transforms with Video support. We should investigate which of the current transforms support this and which don't. Also how about the batch dimension aka `[B, T, C, H, W]`; any thoughts for handling that on the new API?

> I'm not sure I follow. 

I just wanted to bring up the fact that it's not a given that ALL transforms will operate independently on the temporal dimension. So on the new API, there might be the need to support eventually a transform that operates on (T, C, H, W) jointly. Just mentioning this so that we don't limit in on the API.

pmeier: Having a `Video` feature with `[..., T, C, H, W]` will give us the most flexibility. All transforms that don't need the temporal information can simply use the corresponding image kernel that expects `[..., C, H, W]`. At the same time, if we later have transforms that need the temporal information, they are automatically supported.

PR #5381, opened 2022-02-07
Change default weights of RAFT model builders

NicolasHug:
This changes the default of the RAFT model builders to rely on the more accurate ``C_T_SKHT_V2`` weights, + some doc updates

### DIFF
--- a/torchvision/prototype/models/optical_flow/raft.py
+++ b/torchvision/prototype/models/optical_flow/raft.py
@@ -151,7 +151,7 @@ class Raft_Small_Weights(WeightsEnum):
     DEFAULT = C_T_V2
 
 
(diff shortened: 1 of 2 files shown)


torchvision/prototype/models/optical_flow/raft.py
```diff
@@ -115,7 +115,7 @@ class Raft_Large_Weights(WeightsEnum):
         },
     )
 
-    DEFAULT = C_T_V2
+    DEFAULT = C_T_SKHT_K_V2
```
datumbox, line 118: I can't comment on the right spot of the code but you need to change this also below at the:
```
@handle_legacy_interface(weights=("pretrained", Raft_Large_Weights.C_T_SKHT_K_V2))
```

datumbox (approved):
Approving to unblock but you need one more change, see below.

Issue #5347, opened 2022-02-03
Places365


pmeier:

Amapocho:
Can I take this up?

pmeier:
Sure, go ahead.

Amapocho:
The Places365-**Challenge** dataset is way too big for me to download and find the sha256 code, is there any other way to find it?

Amapocho:
I'm sorry but I just realized I will need to download the entire dataset to build the data pipe and I cannot do that for such a huge dataset.  Please do unassign this issue, I will look into the other datasets that I will be able to download and try to go forward with them.

Regret any inconvenience caused.

Should I make a PR with resources that I have updated or should I just let it be?

Amapocho:
I'm sorry but I just realized I will need to download the entire dataset to build the data pipe and I cannot do that for such a huge dataset. Please do unassign this issue, I will look into the other datasets that I will be able to download and try to go forward with them.

Regret any inconvenience caused.

Should I make a PR with resources that I have updated or should I just let it be?

pmeier:
> I'm sorry but I just realized I will need to download the entire dataset to build the data pipe and I cannot do that for such a huge dataset. Please do unassign this issue, I will look into the other datasets that I will be able to download and try to go forward with them.
> 
> Regret any inconvenience caused.

No worries, I could have told you before you got started and this is on me. Still, before I unassign you, would you be willing to implement this for the small variant of the dataset. This limits the download size for all files to about 30GB. IIRC, they have exactly the same structure as their larger "sisters". So the only thing I would have to do after your PR is checking if the implementation still works with the other files.

> Should I make a PR with resources that I have updated or should I just let it be?

Depends. Do you have more than just the skeleton? If yes, go ahead and send the PR. Otherwise I don't think it will be of much help for someone taking over.

Amapocho:
I only have 25GB on the drive which has PyTorch so I won't be able to run it on my device sadly. I have added all the links for the resource download, image type, and all the other links so I'll add a PR as it'll help start off whoever picks up the issue next.

In the meanwhile, I have looked into the RenderedSST2 and I would be able to take that up. Is there any quirk to that dataset I should be wary of before commenting on the issue?

pmeier:
> In the meanwhile, I have looked into the RenderedSST2 and I would be able to take that up. Is there any quirk to that dataset I should be wary of before commenting on the issue?

No, I don't think so. Go ahead, I can assign you there.

[...]

PR #5382, opened 2022-02-07
Places365 dataset #5347

Amapocho:
Added necessary links and dependencies as a starting point for places365


Amapocho:
> Hey @Amapocho, would you please allow me to push some updates to this branch? I have a local cleanup commit that could help others start with this dataset.

Yeah, should I add you as a collaborator or should I just accept a pull request from your side?

Amapocho:
> The easiest is if you simply check this box
> 
> that is located at the bottom of the right column of a PR. With that you grant us push access to just this branch.

I already have ticked the box so shouldn't be an issue

Amapocho:
>  In general, you should do your development on a feature branch. As a rule of thumb, unless you actually want to diverge from the upstream repository, you should not touch main of your fork.

Thanks a lot, I did not know about it beforehand. I have made the necessary changes

PR #5384, opened 2022-02-07
Added reference tests for shear X/Y in autoaugment ops

vfdev-5:
Description:

Added a test to check torchvision's implementation vs ref implementations using PIL, e.g.
- https://github.com/google-research/augmix/blob/master/augmentations.py#L81-L96
- https://github.com/rwightman/pytorch-image-models/blob/master/timm/data/auto_augment.py#L56-L63

Pillow affine vs Torch affine transformations are not always identical:
- for nearest mode small shear values produces almost identical results
- for larger magnitudes we can have certain discrepancies between some lines
- bilinear modes produce pixel diffs ~10 for majority of pixels


test/test_transforms_tensor.py
```diff
@@ -725,6 +725,35 @@ def test_autoaugment_save(augmentation, tmpdir):
     s_transform.save(os.path.join(tmpdir, "t_autoaugment.pt"))
 
 
+def test_autoaugment__op_apply_shear():
```
datumbox, line 728: LGTM, but shouldn't we add this test to the `test_transforms.py` file instead since this is where the AutoAugment tests live anyway?

vfdev-5: autoaugment tests live here and there. In `test_transforms.py` we generally test on PIL images and here on tensors. Do you mean that we have to make a similar test on PIL images ? I agree that this can be good as well.

test/test_transforms_tensor.py
```diff
@@ -762,12 +763,19 @@ def shear(pil_img, level, mode, resample):
...
-            t_img, op_name=f"Shear{mode}", magnitude=arc_level, interpolation=interpolation, fill=0
-        )
-        _assert_approx_equal_tensor_to_pil(out, expected_out)
+    if interpolation == F.InterpolationMode.BILINEAR:
+        # We skip bilinear mode for tensors as
+        # affine transformation results are not exactly the same
+        # between tensors and pil images
+        # MAE as around 1.40
+        # Max Abs error can be 163 or 170
+        return
```
datumbox, line 771: Shall we remove it from the `parametrize` then?

datumbox (approved):
LGTM, with a minor nit:

test/test_transforms_tensor.py
vfdev-5: We can't remove it from parametrize as still checking that for torchvision out PIL vs expected PIL, just above, L758 (`assert out == expected_out`)

PR #5386, opened 2022-02-07
Remove APIs that were deprecated before 0.8

NicolasHug:
Pytorch core is moving to a [2-version deprecation cycle](https://github.com/pytorch/pytorch/wiki/%5BDraft%5D-PyTorch's-Python-Frontend-Backward-and-Forward-Compatibility-Policy), and so are we.

This PR removes APIs that were removed before or on 0.8, released 1.5 years ago. We're going to release 0.12 so technically we could remove stuff up to 0.10, but we're just being extra nice.


NicolasHug:
D34048252 is green internally, this should be good to merge

datumbox (approved):
LGTM, thanks!

PR #5387, opened 2022-02-07
properly deprecate legacy implementation

pmeier:
Follow-up to #5386.

### DIFF
--- a/test/test_functional_tensor.py
+++ b/test/test_functional_tensor.py
@@ -144,7 +144,7 @@ class TestRotate:
         with pytest.warns(
             UserWarning,
             match=re.escape(
-                "Parameter 'resample' is deprecated since 0.12.0 and will be removed 0.14.0. "
+                "The parameter 'resample' is deprecated since 0.12 and will be removed 0.14. "
                 "Please use 'interpolation' instead."
             ),
         ):
--- a/test/test_functional_tensor.py
+++ b/test/test_functional_tensor.py
@@ -371,7 +371,13 @@ class TestAffine:
         tensor, pil_img = _create_data(26, 26, device=device)
 
         # assert deprecation warning and non-BC
-        with pytest.warns(UserWarning, match=r"Argument resample is deprecated and will be removed"):
+        with pytest.warns(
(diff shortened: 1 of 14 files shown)


torchvision/datasets/kinetics.py
```diff
@@ -247,6 +248,9 @@ class Kinetics400(Kinetics):
     `Kinetics-400 <https://deepmind.com/research/open-source/open-source-datasets/kinetics/>`_
     dataset.
 
+    .. deprecated:: 0.12.0
+        Please use ``Kinetics(..., num_classes='400')`` instead. Will be removed in ``0.14.0``.
+
```
NicolasHug, line 253: I don't know why but unfortunately this isn't rendered properly by our sphinx team. I guess we'll have to switch to warnings for it to be visible. We'll also have to indicate the 0.12 version in the message

```suggestion
    .. warning::
        This class was deprecated in 0.12 and will be removed in 0.14. Please use ``Kinetics(..., num_classes='400')`` instead.

```

test/test_functional_tensor.py
```diff
@@ -141,7 +141,13 @@ def test_rotate_batch(self, device, dt):
     def test_rotate_deprecation_resample(self):
         tensor, _ = _create_data(26, 26)
         # assert deprecation warning and non-BC
-        with pytest.warns(UserWarning, match=r"Argument resample is deprecated and will be removed"):
+        with pytest.warns(
+            UserWarning,
+            match=re.escape(
+                "Parameter 'resample' is deprecated since 0.12.0 and will be removed 0.14.0. "
```
NicolasHug, line 147: Nit here and in every other message: we don't need the `.0` part, we can just write 0.12 and 0.14 which is simpler to read. But it's a nit

NicolasHug, line 147: also tests are failing
```suggestion
                "The parameter 'resample' is deprecated since 0.12.0 and will be removed 0.14.0. "
```

NicolasHug (approved):
Thanks @pmeier , minor comment but LGTM when addressed and green

PR #5389, opened 2022-02-08
fix `make html-noplot` docs build command

### LINKED ISSUE #5388: Building docs with `make html-noplot` ignores sphinx options
The command currently looks like this:

https://github.com/pytorch/vision/blob/97eddc5d6a83a9bf620070075ef1e1864c9a68ac/docs/Makefile#L26-L27

Not that `ASPHINXOPTS` is non-existent and thus we build without any options. Replacing this with `SPHINXOPTS`

https://github.com/pytorch/vision/blob/97eddc5d6a83a9bf620070075ef1e1864c9a68ac/docs/Makefile#L9

leads to a warning of sphinx gallery (see sphinx-gallery/sphinx-gallery#913) which in turn stops the build in its tracks through our `-W` usage. 

If we don't want these options for `html-noplot`, I suggest we remove the `ASPHINXOPTS` substitution. Otherwise, we should probably wait for them to fix this.

pmeier:
Fixes #5388.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -24,7 +24,7 @@ docset: html
 	convert $(SPHINXPROJ).docset/icon@2x.png -resize 16x16 $(SPHINXPROJ).docset/icon.png
 
 html-noplot:  # Avoids running the gallery examples, which may take time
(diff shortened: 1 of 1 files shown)


NicolasHug (approved):
Thanks @pmeier

PR #5391, opened 2022-02-08
properly deprecate legacy implementation -- bis

NicolasHug:
Some more deprecations that I forgot in https://github.com/pytorch/vision/pull/5387


pmeier (approved):
Thanks @NicolasHug! LGTM when CI is green.

torchvision/models/mobilenetv3.py
```diff
@@ -29,7 +29,7 @@ def __init__(self, input_channels: int, squeeze_factor: int = 4):
         self.relu = self.activation
         delattr(self, "activation")
         warnings.warn(
-            "This SqueezeExcitation class is deprecated and will be removed in future versions. "
+            "This SqueezeExcitation class is deprecated since 0.12 and will be removed in 0.14. "
```
oke-aditya, line 32: Coz we have alternative Squeeze Excite in ops.  Can we suggest to use it instead. 
Like

Mobilenet Squeeze Excite is deprecated since 0.12 and will be removed in 0.14. Use ops.SqueezeExcite instead.

torchvision/datasets/voc.py
```diff
@@ -80,7 +80,8 @@ def __init__(
         if year == "2007-test":
             if image_set == "test":
                 warnings.warn(
-                    "Acessing the test image set of the year 2007 with year='2007-test' is deprecated. "
+                    "Acessing the test image set of the year 2007 with year='2007-test' is deprecated "
```
prabhat00155, line 83: Typo here: `Accessing`

prabhat00155 (approved):
Thanks @NicolasHug!

PR #5392, opened 2022-02-08
Consolidate   repr  

jdsgomes:
Attempt to make the `__repr__` uniform across the repo by:
- adding return type
- using almost uniquely f-strings
- removing the pattern `s.format(**self.__dict__)` and making the variables we want to use explicit in the f-string


torchvision/transforms/transforms.py
```diff
@@ -1130,11 +1128,13 @@ def forward(self, tensor: Tensor) -> Tensor:
         tensor = transformed_tensor.view(shape)
         return tensor
 
-    def __repr__(self):
-        format_string = self.__class__.__name__ + "(transformation_matrix="
-        format_string += str(self.transformation_matrix.tolist()) + ")"
-        format_string += ", (mean_vector=" + str(self.mean_vector.tolist()) + ")"
-        return format_string
+    def __repr__(self) -> str:
```
jdsgomes, line 1131: This is not equivalent, but it seems to me that the original implementation was wrong. I will ask the reviewer to double check this, as my interpretation could be wrong.

For example if the original output would is

`class_name(transformation_matrix=[1, 2, 3, 4]), (mean_vector=[2, 3, 4, 5])`

with the modifications we should expect


`class_name(transformation_matrix=[1, 2, 3, 4], mean_vector=[2, 3, 4, 5])`

torchvision/transforms/transforms.py
```diff
@@ -1529,24 +1531,18 @@ def forward(self, img):
...
-        return s.format(name=self.__class__.__name__, **d)
+    def __repr__(self) -> str:
+        s = (
+            f"{self.__class__.__name__}(degrees={self.degrees}"
+            f"{f', translate={self.translate}' if self.translate is not None else ''}"
+            f"{f', translate={self.scale}' if self.scale is not None  else ''}"
+            f"{f', translate={self.shear}' if self.shear is not None  else ''}"
+            f"{f', translate={self.interpolation.value}' if self.interpolation != InterpolationMode.NEAREST else ''}"
+            f"{f', translate={self.fill}' if self.fill != 0 else ''}"
+            f"{f', translate={self.center}' if self.center is not None  else ''}"
```
datumbox, line 1542: Nit: Ouch! I don't think this makes the code more readable.

torchvision/ops/deform_conv.py
```diff
@@ -179,14 +179,16 @@ def forward(self, input: Tensor, offset: Tensor, mask: Optional[Tensor] = None)
...
-        return s.format(**self.__dict__)
+        s = (
+            f"{self.__class__.__name__}("
+            f"{self.in_channels}"
+            f", {self.out_channels}"
+            f", kernel_size={self.kernel_size}"
+            f", stride={self.stride}"
+            f"{f', padding={self.padding}' if self.padding != (0, 0) else ''}"
+            f"{f', dilation={self.dilation}' if self.dilation != (1, 1) else ''}"
+            f"{f', groups={self.groups}'  if self.groups != 1 else ''}"
```
datumbox, line 190: Nit: Same here.

torchvision/ops/deform_conv.py
```diff
@@ -179,14 +179,16 @@ def forward(self, input: Tensor, offset: Tensor, mask: Optional[Tensor] = None)
...
+            f"{', bias=False' if self.bias is None else ''}"
```
datumbox, line 191: Did you miss a nested `f`?
```suggestion
            f"{f', bias=False' if self.bias is None else ''}"
```

torchvision/transforms/transforms.py
datumbox: Agreed, the old is wrong.

datumbox (commented):
Thanks, definitely an improvement overall. There is a minor bug and 2 nit comments.

torchvision/ops/deform_conv.py
jdsgomes: in this case there is no need because 'bias=False is just a string without placeholders

NicolasHug:
Thanks for the PR @jdsgomes ,

No strong opinion but we might be able to avoid the use of the use of double-f strings like 

```py
            f"{f', translate={self.interpolation.value}' if self.interpolation != InterpolationMode.NEAREST else ''}"
```  

by using `join()`?

```py
s = "".join([
    f"x={x} if x is not None" else "",
    f"y={y} if y is not None" else "",
....
])
```

torchvision/ops/deform_conv.py
datumbox: Oh yes, you are right. Let's go with @NicolasHug's proposal below to avoid nesting and make the code easier to understand.

jdsgomes:
> Thanks for the PR @jdsgomes ,
> 
> No strong opinion but we might be able to avoid the use of the use of double-f strings like
> 
> ```python
>             f"{f', translate={self.interpolation.value}' if self.interpolation != InterpolationMode.NEAREST else ''}"
> ```
> 
> by using `join()`?
> 
> ```python
> s = "".join([
>     f"x={x} if x is not None" else "",
>     f"y={y} if y is not None" else "",
> ....
> ])
> ```

Thanks for the comment. I'm reverting this.

torchvision/ops/deform_conv.py
jdsgomes: missed this, but I already removed the nested fstrings (different from @NicolasHug  proposal). Despite his proposal is more in line with the spirit of this PR I think we can merge as it is if you agree.

datumbox (approved):
LGTM!

PR #5401, opened 2022-02-09
Keep prototype.datasets._api.find() private

NicolasHug:
After discussing with @fmassa there is no obvious reason to expose it publicly


datumbox (approved):
LGTM.

PR #5402, opened 2022-02-09
Refactor Augmentation Space calls to speed up

datumbox:
Minor speed optimization to avoid calling repeatedly to get the augmentation space within the same forward call.


prabhat00155 (approved):
Thanks @datumbox!

PR #5404, opened 2022-02-10
[WIP] Mask RCNN implementation to support PyTorch / XLA

miladm:
Run Mask RCNN on PyTorch/XLA.


miladm:
The current implementation successfully runs on PyTorch/XLA.

Some performance optimizations to hand dynamic shape remain to be supported after this feature is available on PyTorch/XLA.

PR #5411, opened 2022-02-11
Adding AugMix implementation

datumbox:
Adding the [AugMix](https://arxiv.org/abs/1912.02781) data augmentation method. Inspired from the work on the [official repo](https://github.com/google-research/augmix/blob/9b9824c7c19bf7e72df2d085d97b99b3bfb00ba4/imagenet.py#L209).


torchvision/transforms/autoaugment.py
```diff
@@ -458,3 +458,140 @@ def __repr__(self) -> str:
...
+            img = orig_img
+            if isinstance(fill, (int, float)):
+                fill = [float(fill)] * F.get_image_num_channels(img)
+            elif fill is not None:
+                fill = [float(f) for f in fill]
+        else:
+            img = self._pil_to_tensor(orig_img)
+
+        mixing_weights = torch._sample_dirichlet(torch.tensor([self.alpha] * self.mixture_width, device=img.device))
+        m = torch._sample_dirichlet(torch.tensor([self.alpha, self.alpha], device=img.device))[0]
```
datumbox, line 556: We use the same `m` value for the entire batch.

torchvision/transforms/autoaugment.py
```diff
@@ -458,3 +458,140 @@ def __repr__(self) -> str:
...
+        op_meta = self._augmentation_space(self._PARAMETER_MAX, F.get_image_size(img))
+
+        orig_dims = list(img.shape)
+        batch = img.view([1] * max(4 - img.ndim, 0) + orig_dims)
+        mix = torch.zeros_like(batch, dtype=torch.float)
+        for i in range(self.mixture_width):
+            aug = batch.clone()
+            depth = self.chain_depth if self.chain_depth > 0 else int(torch.randint(low=1, high=4, size=(1,)).item())
```
datumbox, line 564: If negative (-1) we randomly sample a depth between `[1,3]`

datumbox:
@hendrycks Thanks for the input! 

In the current implementation, we support it and make it optional to use the extra 4 transforms by setting `all_ops=True` on the constructor. Check this https://github.com/pytorch/vision/pull/5411#discussion_r804798735

Let me know if you have any other feedback on the implementation. The main changes are on the `autoaugment.py` file of this PR.

datumbox:
@normster @hendrycks We intend to merge soon. If you have any thoughts on the comments above let us know, thanks!

torchvision/transforms/autoaugment.py
```diff
@@ -458,3 +458,140 @@ def __repr__(self) -> str:
...
+            image. If given a number, the value is used for all bands respectively.
+    """
+
+    def __init__(
+        self,
+        severity: int = 1,
+        mixture_width: int = 3,
+        chain_depth: int = -1,
+        alpha: float = 1.0,
+        all_ops: bool = True,
```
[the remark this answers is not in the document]
normster: These look good

torchvision/transforms/autoaugment.py
```diff
@@ -458,3 +458,140 @@ def __repr__(self) -> str:
...
+        if isinstance(orig_img, Tensor):
+            img = orig_img
+            if isinstance(fill, (int, float)):
+                fill = [float(fill)] * F.get_image_num_channels(img)
+            elif fill is not None:
+                fill = [float(f) for f in fill]
+        else:
+            img = self._pil_to_tensor(orig_img)
+
+        mixing_weights = torch._sample_dirichlet(torch.tensor([self.alpha] * self.mixture_width, device=img.device))
```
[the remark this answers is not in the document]
normster: We sample different mixing/m values per image in the original implementation. Since this is a torchvision transform, isn't it usually applied per image?

normster (commented):
This looks great! I had a small question about per-image/per-batch processing that I noted inline but what you have should work fine in practice.

datumbox:
@normster Thanks a lot for the feedback. In relation to the sampling of weights, I will benchmark your proposal. Out of curiosity, did you do any experiments on the sampling per image vs batch?

torchvision/transforms/autoaugment.py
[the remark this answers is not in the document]
datumbox: Thanks for checking. In this implementation it depends on whether the user calls it by providing an image vs batch. We can certainly force sampling per image in all cases, but it's going to be slightly slower. I wonder what the accuracy implications of this choice is. If you have done experiments on this please let me know.

torchvision/transforms/autoaugment.py
```diff
@@ -458,3 +458,150 @@ def __repr__(self) -> str:
...
+        """
+        fill = self.fill
+        if isinstance(orig_img, Tensor):
+            img = orig_img
+            if isinstance(fill, (int, float)):
+                fill = [float(fill)] * F.get_image_num_channels(img)
+            elif fill is not None:
+                fill = [float(f) for f in fill]
+        else:
+            img = self._pil_to_tensor(orig_img)
```
vfdev-5, line 557: Later we may want to refactor this part of code that could be applicable to all other aug strategies...

datumbox: Certainly. We would need to consider how Videos can be handled as well here too.

vfdev-5 (approved):
LGTM, thanks @datumbox !

Issue #5414, opened 2022-02-13
[RFC] New Ops in TorchVision


datumbox:
### 🚀 The feature

Consider adding the following operators in TorchVision:

## Layers
- [ ] [DropBlock Layer](https://arxiv.org/abs/1810.12890)
- [ ] [AutoDropout Layer](https://arxiv.org/abs/2101.01761)
- [ ] [DropConnect Layer](http://yann.lecun.com/exdb/publis/pdf/wan-icml-13.pdf)
- [ ] [ShakeDrop Layer](https://arxiv.org/abs/1802.02375)

## Operators
- [ ] [SoftNMS](https://arxiv.org/pdf/1704.04503.pdf)

## Schedulers
- [ ] [Polynomial LR scheduler](https://github.com/pytorch/vision/issues/4438) 
- [ ] [Random Noise LR Scheduler](https://arxiv.org/abs/1810.01322)

### Motivation, pitch

-

datumbox:
@xiaohu2015 wow, I literally just slacked you to see if you are interested 😄 

Great! If you want to send a PR for `DropBlock` it would be awesome. Let me create a ticket for it so that people know you are working on it already.

lezwon:
hey @datumbox, can I take up SoftNMS implementation?

datumbox:
@lezwon Thanks for offering help!

The SoftNMS would have to be implemented in [C++](https://github.com/pytorch/vision/blob/main/torchvision/csrc/ops/cpu/nms_kernel.cpp) and [CUDA](https://github.com/pytorch/vision/blob/main/torchvision/csrc/ops/cuda/nms_kernel.cu) because this is where we implement the standard NMS. Some additional discussion would be required to see exactly how this will be implemented and what its API would look like. As you understand this is quite a lot of work and it's not guaranteed that the feature will be merged. If you are up for it, we can discuss more. Just wanted to give you a heads up that this is a more risky feature to work on.

If the above doesn't sound too appealing, there are features listed at #5410 you might find fun to work on. Have a look and let me know if anything interests you. :)

lezwon:
@datumbox Sure thing :)
I'll pick up something from #5410

oke-aditya:
I want to try [DropConnect Layer](http://yann.lecun.com/exdb/publis/pdf/wan-icml-13.pdf). Any other info / implementation I could look to will be great :smiley:

datumbox:
[...]

PR #5416, opened 2022-02-13
New Feature: add DropBlock layer

xiaohu2015:
As discussed in https://github.com/pytorch/vision/issues/5414, I am working to add DropBlock layer to Torchvision.

we add DropBlock2d and DropBlock3d module.

### Reference
- https://github.com/tensorflow/tpu/blob/master/models/official/resnet/resnet_model.py#L276
- https://github.com/rwightman/pytorch-image-models/blob/master/timm/models/layers/drop.py
- https://github.com/open-mmlab/mmdetection/blob/master/mmdet/models/plugins/dropblock.py


torchvision/ops/dropblock.py
```diff
@@ -1 +1,64 @@
...
+    <https://arxiv.org/abs/1810.12890>`.
+
+    Args:
+        p (float): Probability of an element to be dropped.
+        block_size (int): Size of the block to drop.
+        inplace (bool): If set to ``True``, will do this operation in-place. Default: ``False``
+    """
+
+    def __init__(self, p: float, block_size: int, inplace: bool = False) -> None:
+        super(DropBlock2d, self).__init__()
```
datumbox, line 20: Nit:
```suggestion
        super().__init__()
```

torchvision/ops/dropblock.py
```diff
@@ -1 +1,64 @@
...
+
+    def forward(self, input: Tensor) -> Tensor:
+        """
+        Args:
+            input (Tensor): Input feature map on which some areas will be randomly
+                dropped.
+        Returns:
+            Tensor: The tensor after DropBlock layer.
+        """
+        if not self.training:
```
datumbox, line 37: Skip early if p is zero:
```suggestion
        if not self.training or self.p == 0.0:
```

torchvision/ops/dropblock.py
```diff
@@ -1 +1,64 @@
...
+        mask = F.pad(mask, [self.block_size // 2] * 4, value=0)
+        mask = F.max_pool2d(
+            input=mask,
+            stride=(1, 1),
+            kernel_size=(self.block_size, self.block_size),
+            padding=self.block_size // 2)
+        mask = 1 - mask
+        normalize_scale = mask.numel() / (1e-6 + mask.sum())
+        if self.inplace:
+            input.mul_(mask * normalize_scale)
```
datumbox, line 53: Perhaps the following will be faster?
```suggestion
            input.mul_(mask).mul_(normalize_scale)
```

torchvision/ops/dropblock.py
```diff
@@ -1 +1,64 @@
...
+            return input
+        
+        N, C, H, W = input.size()
+        # compute the gamma of Bernoulli distribution
+        gamma = (self.p * H * W) / ((self.block_size ** 2) * ((H - self.block_size + 1) * \
+            (W - self.block_size + 1)))
+        mask_shape = (N, C, H - self.block_size + 1, W - self.block_size + 1)
+        mask = torch.bernoulli(torch.full(mask_shape, gamma, device=input.device))
+
+        mask = F.pad(mask, [self.block_size // 2] * 4, value=0)
```
datumbox, line 46: Could you provide a reference for the choice of `mask_shape` and the subsequent padding with zeros that you do? I need to check the paper in more detail but my expectation was that we would sample from bernoulli with shape equal the input. I'll need to have a closer look on the paper to remember the details.

torchvision/ops/dropblock.py
```diff
@@ -1 +1,64 @@
...
+        mask = F.max_pool2d(
+            input=mask,
+            stride=(1, 1),
+            kernel_size=(self.block_size, self.block_size),
+            padding=self.block_size // 2)
+        mask = 1 - mask
+        normalize_scale = mask.numel() / (1e-6 + mask.sum())
```
datumbox, line 51: Shall we make `eps=1e-6` configurable on the constructor of the class?

torchvision/ops/dropblock.py
```diff
@@ -1 +1,64 @@
...
+        super(DropBlock2d, self).__init__()
+        _log_api_usage_once(self)
+
+        if p < 0.0 or p > 1.0:
+            raise ValueError(f"drop probability has to be between 0 and 1, but got {p}")
+        self.p = p
+        self.block_size = block_size
+        self.inplace = inplace
+
+    def forward(self, input: Tensor) -> Tensor:
```
datumbox, line 29: I recommend moving the entire code of `forward` to a separate public method so that it's available on a functional API. See the `stochastic_depth` method for [reference](https://github.com/pytorch/vision/blob/main/torchvision/ops/stochastic_depth.py#L8).

torchvision/ops/dropblock.py
```diff
@@ -1 +1,64 @@
...
+        """
+        if not self.training:
+            return input
+        
+        N, C, H, W = input.size()
+        # compute the gamma of Bernoulli distribution
+        gamma = (self.p * H * W) / ((self.block_size ** 2) * ((H - self.block_size + 1) * \
+            (W - self.block_size + 1)))
+        mask_shape = (N, C, H - self.block_size + 1, W - self.block_size + 1)
+        mask = torch.bernoulli(torch.full(mask_shape, gamma, device=input.device))
```
datumbox, line 44: Worth refactoring this to use the following idiom:
https://github.com/pytorch/vision/blob/93c85bbcc31f8d5a052daf06f2f91f39697af1a4/torchvision/ops/stochastic_depth.py#L40-L41

torchvision/ops/dropblock.py
```diff
@@ -1 +1,64 @@
 import torch
+import torch.nn.functional as F
+from torch import nn, Tensor
+
+from ..utils import _log_api_usage_once
+
+
+class DropBlock2d(nn.Module):
```
datumbox, line 8: Creating new tensors is not supported by FX, so we will need to declare it:
https://github.com/pytorch/vision/blob/93c85bbcc31f8d5a052daf06f2f91f39697af1a4/torchvision/ops/stochastic_depth.py#L47

torchvision/ops/dropblock.py
```diff
@@ -1 +1,64 @@
...
+        if self.inplace:
+            input.mul_(mask * normalize_scale)
+        else:
+            input = input * mask * normalize_scale
+        return input
+
+    def __repr__(self) -> str:
+        s = f"{self.__class__.__name__}(p={self.p}, block_size={self.block_size}, " + \
+            f"inplace={self.inplace})"
+        return s
```
datumbox, line 60: There are a few boilerplate things that need to be done when adding an operator such as adding it to docs, including it on init files, adding tests etc etc. Here is an equivalent PR that you can follow as an example: #4301

datumbox (commented):
@xiaohu2015 Awesome work as always. Thanks a lot!

The implementation looks very nice overall. I've added a couple of comments concerning nits and boilerplate code needed for ops in TorchVision. My only real question is on the way you do the sampling on bernoulli, see below.

Finally, would you be interested in implementing the 3d version on the same PR? We want to start offering better video support and this contribution will help us offer better models. If you prefer, we can also do it on a follow up.

torchvision/ops/dropblock.py
xiaohu2015: @datumbox the implementation is consistent with the original paper, there is 3 steps:
1. generate the center mask, which means the block center, so the center mask shape : `mask_shape = (N, C, H - self.block_size + 1, W - self.block_size + 1)`; we also should convert the `drop_prob` to `gamma` of the bernoulli.
2. convert the center mask to block mask: `pad` and use a `maxpool` to get block mask.
3. use the block mask to mask the input.

datumbox: I've checked the paper in detail and you are right. Page 3, figure 2a shows that the center mask should be drawn from the green area, hence your approach is correct. Also this implementation aligns closely with the [original TF implementation](https://github.com/tensorflow/tpu/blob/a6005716b311f96190590414479f0261726270e7/models/official/resnet/resnet_model.py#L315-L316).

datumbox:
@xiaohu2015 I've checked the paper and the original TF implementation and I believe your approach looks great. Thanks a lot for the contribution.

The only pendings are:
- Fix tests (explicitly `import torch.fx` to solve some of the [failing tests] + fix the [linter](https://app.circleci.com/pipelines/github/pytorch/vision/14924/workflows/86e431cb-a503-412a-8d13-0424bb7ea59b/jobs/1201333))
- Add the new ops on `docs/source/ops.rst`
- Add a test on `test/test_ops.py` (see #4301)

xiaohu2015:
> @xiaohu2015 I've checked the paper and the original TF implementation and I believe your approach looks great. Thanks a lot for the contribution.
> 
> The only pendings are:
> 
> * Fix tests (explicitly `import torch.fx` to solve some of the [failing tests](https://app.circleci.com/pipelines/github/pytorch/vision/14924/workflows/c4da8b86-f1ae-4809-8991-01ea8ad4274e/jobs/1201344) + fix the [linter](https://app.circleci.com/pipelines/github/pytorch/vision/14924/workflows/86e431cb-a503-412a-8d13-0424bb7ea59b/jobs/1201333))
> * Add the new ops on `docs/source/ops.rst`
> * Add a test on `test/test_ops.py` (see [Add StochasticDepth implementation #4301](https://github.com/pytorch/vision/pull/4301))

OK

torchvision/ops/drop_block.py
```diff
@@ -103,7 +101,11 @@ class DropBlock2d(nn.Module):
 
     def __init__(self, p: float, block_size: int, inplace: bool = False, eps: float = 1e-06) -> None:
         super().__init__()
-
+        
+        if p < 0.0 or p > 1.0:
+            raise ValueError(f"drop probability has to be between 0 and 1, but got {p}")
+        if block_size % 2 == 0:
+            raise ValueError(f"block size has to be an odd number, but got {block_size}")
```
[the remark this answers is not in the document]
xiaohu2015: oh, I remove this.

test/test_ops.py
```diff
@@ -1355,6 +1366,56 @@ def test_split_normalization_params(self, norm_layer):
...
+            layer = ops.DropBlock3d(p=p, block_size=block_size, inplace=inplace)
+            feature_size = depth * height * width
+        layer.__repr__()
+
+        out = layer(x)
+        if p == 0:
+            assert out.equal(x)
+        if block_size == height:
+            for b, c in product(range(batch_size), range(channels)):
+                assert out[b, c].count_nonzero() in (0, feature_size)
```
datumbox, line 1399: Is this because it's either going to have selected 1 center and thus the whole block will turn to zeros OR no center in which case no zeros?

test/test_ops.py
```diff
@@ -1355,6 +1366,56 @@ def test_split_normalization_params(self, norm_layer):
...
         assert len(params[1]) == 82
+        
+        
+class TestDropBlock:
+    @pytest.mark.parametrize("seed", range(10))
+    @pytest.mark.parametrize("dim", (2, 3))
+    @pytest.mark.parametrize("p", (0, 0.2, 0.5, 0.8))
+    @pytest.mark.parametrize("block_size", [5, 7, 9, 11])
+    @pytest.mark.parametrize("inplace", [True, False])
+    def test_drop_block(self, seed, dim, p, block_size, inplace):
```
datumbox, line 1377: The testing strategy here is not great as we only validate the input for a fraction of the inputs. An alternative approach would be to try to count the zeroes based on the sampling probability. As noted on the paper, the number of zeroes will be approximate due to overlaps but perhaps we can create a not so flaky test for this.

If you prefer, we can do this on a follow up PR.

datumbox (commented):
@xiaohu2015 The changes look good to me. I've pushed to your branch to fix one remaining linter issue. Let's see if the CI passes now.

test/test_ops.py
xiaohu2015: yes

test/test_ops.py
xiaohu2015: I have read the test function of  https://github.com/pytorch/vision/pull/4301, is it necessary check this? I found there is no such check for `nn.Dropout` in PyTorch.

datumbox: Yes that's what I had in mind. The test of `StochasticDepth` confirms that the number of drops are within the expended confidence intervals. These intervals should be larger at DropBlock because of the overlapping blocks.

Concerning whether this test is necessary, the answer is I wouldn't block the PR for this but it will protect us from someone breaking the op accidentally on the future. I'm surprised you say that `Dropout` is not properly checked at PyTorch Core. Perhaps the test is not on an obvious location (I know dropout has tests in a few different places)?

xiaohu2015: @datumbox oh, maybe I missed some codes.

datumbox:
@xiaohu2015 All tests pass. 🚀 🚀 

On the positive side, your implementation seems correct to me, fully aligns with the paper. Though some of the implementation details differ from the references you provided, after doing a deep dive I believe that your approach is equivalent. Some limited testing also shows that the implementation works as intended.

On the negative side, the behaviour of the method is not fully tested (see points at  https://github.com/pytorch/vision/pull/5416#discussion_r811818770). Moreover, we haven't trained a model to show-case that using the layer leads to the expected increase.

What is your bandwidth at this point? Do you have the capacity to do the above 2 checks prior merging? Do you think it's something you would do on a follow up PR after merge? It's a bit hard to provide support on this one due to limited bandwidth on our side because of the upcoming release and other work on the pipeline. Another approach could be to ping a couple of other regular contributors to see if they can help us test it. Let me know your preference/thoughts. Thanks!

test/test_ops.py
xiaohu2015: @datumbox I add the random test check. But, just as you said, the actual sampling probability is slightly lower than the setting value because of overlaps in dropblock.

test/test_ops.py
```diff
@@ -1357,5 +1368,87 @@ def test_split_normalization_params(self, norm_layer):
         assert len(params[1]) == 82
 
 
+class TestDropBlock:
+    @pytest.mark.parametrize("seed", range(10))
+    @pytest.mark.parametrize("dim", [2, 3])
+    @pytest.mark.parametrize("p", [0, 0.2, 0.5, 0.8])
+    @pytest.mark.parametrize("block_size", [5, 7, 9, 11])
```
datumbox, line 1375: Do we need all the combinations of `p` and `block_sizes` given we only assert to a handful of experiments? 

Instead of doing a cartesian product of p x block_sizes, we might want to test just a couple of specific of them to speed the tests.

datumbox (approved):
LGTM, @xiaohu2015.

I've added one last optional test improvement for your consideration below. The testing strategy looks good to me. Let me know when the training checks is completed to merge. 

Thanks a lot as always for the awesome contribution!

test/test_ops.py
xiaohu2015: OK, I remove some.

xiaohu2015:
> LGTM, @xiaohu2015.
> 
> I've added one last optional test improvement for your consideration below. The testing strategy looks good to me. Let me know when the training checks is completed to merge.
> 
> Thanks a lot as always for the awesome contribution!

I just run 90-epoch Resnet50, maybe I can get result next day.

datumbox:
@xiaohu2015 Awesome. The paper uses 90 epochs for the baseline and 250 epochs for those with Dropblock (see page 4, section 4.1 on [paper](https://arxiv.org/pdf/1810.12890.pdf)). If you don't have enough compute to run this, ping me on slack and we can work something out.

FYI the last update broke the [linter](https://app.circleci.com/pipelines/github/pytorch/vision/14999/workflows/985b2195-ebf7-41ef-b3c1-b85af9677da3/jobs/1208564).

datumbox:
Here is my branch for testing the ResNet50 with DropBlock: #5483 

We are not going to used scheduled DropBlock. I'll let you know about the results once I got them.

datumbox:
@xiaohu2015 I got good news. The addition of your DropBlock seems to improve the accuracy. 

Here is the baseline ResNet50 accuracy for a 90-epoch training without DropBlock:
`Test:  Acc@1 76.130 Acc@5 92.862`

Here is the one if we add DropBlock on Groups 3&4 wo/ scheduling (implementation details [here](https://github.com/datumbox/vision/blob/4e76a423bc0ef8c7065125f60745f4745da44885/torchvision/models/resnet.py)):
`Test:  Acc@1 77.064 Acc@5 93.436`

Both the baseline and DropBlock values are different from the ones reported on the paper (baseline 76.51 and DropBlock ~77.5), but I think that's the result of other minor differences that we didn't replicate (hyper-params, small architectural improvements etc). I think we have sufficient evidence that your implementation work as expected and we can merge the PR. Thanks so much for the awesome contribution and work!

PR #5407, opened 2022-02-11
Revamp prototype features and transforms

pmeier:
This PR is a large overhaul of what we currently have for prototype features and transforms. For larger changes, individual PRs were send to this branch.

PR #5417, opened 2022-02-14
Cleanup prototype kernel dispatchers

pmeier:
This PR is a follow-up to #5407 and does two things:

1. It ports all the remaining stable kernels to the prototype kernels
2. It clarifies the docstring of the dispatchers, since `ADDME` could also mean that an implementation needs to be added, which is not the case.


datumbox (approved):
LGTM, thanks @pmeier.

PR #5418, opened 2022-02-14
add prototype transforms that use the prototype dispatchers

pmeier:
This PR re-adds the transformations removed in #5407 to use the dispatchers that were introduced in that PR. Since we have the high-level functional API (dispatchers), most transforms only need to subclass from the `Transform` base class and set the `_DISPATCHER` attribute. In case the transformation takes additional parameters, `__init__` and `get_params` has to be overwritten. In case the parameters are constant, one can also subclass from `ConstantParamTransform` to make this even more concise.


torchvision/prototype/transforms/_transform.py
```diff
@@ -0,0 +1,53 @@
+from typing import Any, Dict, Optional
+
+from torch import nn
+from torchvision.prototype.utils._internal import apply_recursively
+
+from .functional._utils import Dispatcher
+
+
+class Transform(nn.Module):
```
datumbox, line 9: You need a constructor that calls the [logger](https://github.com/pytorch/vision/blob/main/torchvision/transforms/transforms.py#L89-L90). We should also consider on a separate PR to handle the logger on the functional/kernel level. Could be done on the decorator but that won't capture JIT.

torchvision/prototype/transforms/_container.py
```diff
@@ -0,0 +1,59 @@
+from typing import Any, Optional, Dict
+
+import torch
+
+from ._transform import Transform
+
+
+class _WrapperTransform(Transform):
```
datumbox, line 8: We should consider the option of simplifying the inheritance patterns. If not necessary `_WrapperTransform` and `_MultiTransform` can be eliminated.

datumbox (commented):
Some early thoughts:

torchvision/prototype/transforms/_transform.py
```diff
@@ -0,0 +1,53 @@
...
+    def _dispatch(self, input: Any, params: Dict[str, Any]) -> Any:
+        if not self._DISPATCHER:
+            raise NotImplementedError()
+
+        return self._DISPATCHER(input, **params)
+
+    def _transform_recursively(self, sample: Any, params: Dict[str, Any]) -> Any:
+        def transform(input: Any) -> Any:
+            if not self._supports(input):
+                return sample
```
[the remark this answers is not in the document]
pmeier: Nope, that is one key difference between the dispatchers and the transforms. Since the transforms get complete samples from the dataset, most of the transforms will not support all inputs. For example, we don't want `transforms.Normalize()` to fail if anything besides an image is passed.

torchvision/prototype/transforms/_transform.py
```diff
@@ -0,0 +1,53 @@
...
+            return self._dispatch(input, params)
+
+        return apply_recursively(transform, sample)
+
+    def forward(self, *inputs: Any, params: Optional[Dict[str, Any]] = None) -> Any:
+        sample = inputs if len(inputs) > 1 else inputs[0]
+        return self._transform_recursively(sample, params or self.get_params(sample))
+
+    def _extra_repr_from_attrs(self, *names: str) -> str:
+        return ", ".join(f"{name}={getattr(self, name)}" for name in names)
```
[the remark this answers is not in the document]
pmeier: Given that this is just a helper that one still has to invoke manually from `extra_repr()`, I don't see a point.

torchvision/prototype/transforms/_transform.py
```diff
@@ -0,0 +1,53 @@
...
+        def transform(input: Any) -> Any:
+            if not self._supports(input):
+                return sample
+
+            return self._dispatch(input, params)
+
+        return apply_recursively(transform, sample)
+
+    def forward(self, *inputs: Any, params: Optional[Dict[str, Any]] = None) -> Any:
+        sample = inputs if len(inputs) > 1 else inputs[0]
```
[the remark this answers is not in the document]
pmeier: Due to the `*inputs`, everything gets wrapped into a tuple. If we didn't have this line and you would get something like

```py
input = ...
transform = tranforms.Identity()

output = transform(input)

assert (input,) == output
```

We only have to perform this if we detect single inputs, because the behavior is already what you expect if you pass multiple inputs:

```py
image = ...
bounding_box = ...

transform = tranforms.Identity()

image2, bounding_box2 = transform(image, bounding_box)
```

torchvision/prototype/transforms/_transform.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class Transform(nn.Module):
+    _DISPATCHER: Optional[Dispatcher] = None
+
+    def __init__(self) -> None:
+        super().__init__()
+        _log_api_usage_once(self)
+
+    def get_params(self, sample: Any) -> Dict[str, Any]:
```
datumbox, line 18: Why is this public?

torchvision/prototype/transforms/_meta_conversion.py
```diff
@@ -0,0 +1,50 @@
...
+        if isinstance(color_space, str):
+            color_space = features.ColorSpace[color_space]
+        self.color_space = color_space
+
+    def _transform(self, input: Any, params: Dict[str, Any]) -> Any:
+        if not isinstance(input, features.Image):
+            return input
+
+        output = K.convert_color_space(input, old_color_space=input.color_space, new_color_space=self.color_space)
+        return features.Image.new_like(input, output, color_space=self.color_space)
```
datumbox, line 50: This should have been part of the high-level kernel I think. Most of them are supposed to return the right type with the right meta-data. Could you check that this is the case with all high-level kernels (convert_image_dtype, convert_bounding_box_format, etc)? Again I would recommend bringing this on a separate PR.

datumbox (commented):
Few more comments:

torchvision/prototype/transforms/functional/_utils.py
```diff
@@ -9,10 +8,10 @@
 F = TypeVar("F", bound=features._Feature)
 
 
-def dispatch(kernels: Dict[Any, Optional[Callable]]) -> Callable[[Callable[..., F]], Callable[..., F]]:
-    """Decorates a function to automatically dispatch to registered kernels based on the call arguments.
+class Dispatcher:
```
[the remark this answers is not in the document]
pmeier: Nope, this is not only refactoring. Before this PR, we had a regular decorator that defines the wrapper callables as local functions. This is fine as long as you don't need access to these functions. Since we do need this here, for example to find the supported types, I extracted the inner functionality into a class. Recall, that this was the case in #5407 until we stripped away the functionality since it already was aligned with the transforms.

torchvision/prototype/transforms/_transform.py
pmeier: No particular reason other than this is also public on our current stable transforms. I'm ok with making it private.

torchvision/prototype/transforms/_meta_conversion.py
pmeier: Not sure what you are saying here:

- "Why do we have a `ConvertColorSpace` transform?": Because it might be useful to have a transform to convert to a specific color space even if all low-level kernels know how to handle the color space.
- "Why is there no high-level kernel for `convert_color_space`": I intentionally didn't implement any high-level kernels, i.e. dispatchers, for the type and meta conversions, since they by definition apply to only a single feature type. Thus, there wrapping them in a dispatcher is fruitless.

torchvision/prototype/transforms/_geometry.py
```diff
@@ -0,0 +1,75 @@
...
+        self,
+        size: Union[int, Sequence[int]],
+        interpolation: InterpolationMode = InterpolationMode.BILINEAR,
+    ) -> None:
+        super().__init__()
+        self.size = size
+        self.interpolation = interpolation
+
+    def get_params(self, sample: Any) -> Dict[str, Any]:
+        return dict(size=self.size, interpolation=self.interpolation)
```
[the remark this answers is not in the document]
pmeier: > On the previous API, the `get_params` method used to return only the randomly generated parameters. Here it seems that we use it for all of them. Could you confirm that this is the intention? (we should document this prior merging this)

That is intentional, yes. The issue is that due to the automatic dispatch, we have no other "hook" to define which parameters get passed to the dispatchers.

> Also do you think it's good to keep the same name for the function if we change its behaviour? For example to avoid confusion we could name it `_get_args` or something similar. The motivation here is to not confuse this new method with the old static one.

Good point. If we make it private as suggested in https://github.com/pytorch/vision/pull/5418#discussion_r805937596, we should be good.

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -0,0 +1,296 @@
...
+from torchvision.prototype.utils._internal import apply_recursively
+from torchvision.transforms import AutoAugment as _AutoAugment, AutoAugmentPolicy
+
+from . import functional as F
+from .functional._utils import Dispatcher
+from .utils import Query
+
+
+@dataclasses.dataclass
+class AutoAugmentDispatcher:
```
datumbox, line 16: Ad discussed offline we should consider removing this method and use lambdas on the `_DISPATCHER_MAP` or restore the old `_apply_op` method.

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -0,0 +1,296 @@
...
+                if input not in dispatcher:
+                    return input
+
+                return dispatcher(input, magnitude=magnitude, **params)  # type: ignore[arg-type]
+
+            sample = apply_recursively(transform, sample)
+
+        return sample
+
+    def _randbool(self, p: float = 0.5) -> bool:
```
datumbox, line 120: This doesn't actually save us any lines of code, just adds one more method on the base class. Consider just doing `torch.rand() <= p` where necessary.

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -0,0 +1,296 @@
...
+        """Randomly returns either ``True`` or ``False``.
+
+        Args:
+            p: Probability to return ``True``. Defaults to ``0.5``.
+        """
+        return float(torch.rand(())) <= p
+
+
+@dataclasses.dataclass
+class AugmentationMeta:
```
datumbox, line 130: Nit: We will need to review later the names of all these.

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -0,0 +1,296 @@
...
+        elif fill is not None:
+            fill = [float(f) for f in fill]
+
+        return dict(interpolation=self.interpolation, fill=fill)
+
+    def forward(self, *inputs: Any, params: Optional[Dict[str, Any]] = None) -> Any:
+        sample = inputs if len(inputs) > 1 else inputs[0]
+        params = params or self.get_params(sample)
+
+        for transform_id, magnitude in self.get_transforms_meta(Query(sample).image_size()):
```
datumbox, line 107: AutoAugmentation methods should be structured exactly like any other transform. In that sense, can we refactor this have a `_transform` where we do all the operators and remove `get_transforms_meta`?

torchvision/prototype/transforms/_meta_conversion.py
datumbox: As discuss offline, there PIL images can also be grayscaled so you have not a single feature type. But even in the case where there is only 1, the dispatcher will handle for you the type convention (`new_like`) and also the overall solution can benefit from the simplicity of always having a high-level kernel for all low-level kernels.

torchvision/prototype/transforms/_auto_augment.py
pmeier: We cannot. `Transform._transform` already gets passed the non-container elements of the inputs. For the auto augment transforms, we need to apply multiple transforms to the whole input. To illustrate, this PR proposes

```py
for transform in transforms:
    for non_container_element in inputs:
        transform(non_container_element)
```

while your proposal would mean

```py
for non_container_element in inputs:
    for transform in transforms:
        transform(non_container_element)
```

We reserved the option to override the `forward` of a transform for specifically cases like this, when the common structure does not fit the use case.

datumbox: OK, we should try to simplify the `forward` even if this means implementing it separately for each transform. To me, the `get_transforms_meta` is not clear what is doing and I'm afraid it's not generic enough to cover all cases (see AugMix for instance #5411 which is a substantial departure from the existing methods). The less abstraction the better.

torchvision/prototype/transforms/_meta_conversion.py
pmeier: We don't have any of the benefits of the automatic dispatch here, since all types need manual handling

```py
@dispatch(
    {
        torch.Tensor: None,
        PIL.Image.Image: None,
        features.Image: None,
    }
)
def convert_color_space(input: T, *args: Any, **kwargs: Any) -> T:
    color_space = kwargs["color_space"]
    if type(input) is torch.Tensor or isinstance(input, PIL.Image.Image):
        if color_space != "grayscale":
            raise ValueError("For vanilla tensors and PIL images only RGB to grayscale is supported")
        return _F.rgb_to_grayscale(input)
    elif isinstance(input, features.Image):
        output = K.convert_color_space(input, old_color_space=input.color_space, new_color_space=color_space)
        return features.Image.new_like(input, output, color_space=color_space)

    raise RuntimeError
```

torchvision/prototype/transforms/_auto_augment.py
pmeier: > OK, we should try to simplify the `forward` even if this means implementing it separately for each transform.

I'm against that. Most transforms need the default behavior. In fact, the auto augment base class is the only transform that deviates from this scheme. I wouldn't make it more complicated for all other cases just to not have a single (or maybe very few) exceptions in the future. 

> To me, the `get_transforms_meta` is not clear what is doing

I have a hard time to figure out what you want me to document on these PRs and what not. Maybe this is a naming thing (as I said, I'm not really deep into the literature here so maybe my names are wrong), but this simply yields a tuple with the name of the op to apply and the magnitude. It is only used in `forward` to achieve the loop I detailed in my comment above. 

> The less abstraction the better.

I somewhat disagree. I we can simplify our implementation significantly by having a thin abstraction (in this case a single method), I don't think we should strive for no abstraction at all. I'm aware that researchers do crazy things, but our code is also not set in stone. If we later find that some parts are not general enough, we can revise them. It is impossible to foresee the future, so I don't want to maintain a lot of boilerplate code for a maybe upcoming scenario.

> I'm afraid it's not generic enough to cover all cases (see AugMix for instance #5411 which is a substantial departure from the existing methods)

That being said, it seems the abstraction I built is already no general enough for a very tangible scenario and thus I'm ok to remove it. My suggestion is to do this when #5411 is merged.

datumbox: @pmeier This is not a theoretical debate; I've provided you a clear example where the proposed abstraction won't work. I'm just waiting for the authors of the paper to confirm the default hyper-params other than this the AugMix PR is ready to merge. Thus there is no point adding this abstraction now just to remove it later. So here we either need to make the abstraction clear and generic enough to be useful or eliminate it.

torchvision/prototype/transforms/_meta_conversion.py
datumbox: Where do you plan to put the logger if you don't have a high-level kernel? Note that the proposed solution should handle all of the needs outlined. Logging is currently one of them and maintaining a parity between low-level and high-level kernels can allow you to deal with this. If you have alternatives, I'm happy to discuss as long as we don't have to introduce a brand new mechanism.

torchvision/prototype/utils/_internal.py
```diff
@@ -305,22 +305,22 @@ def apply_recursively(fn: Callable, obj: Any) -> Any:
...
-            else:
-                mapping[name] = result
-        return mapping
+        return {key: apply_recursively(fn, item) for key, item in obj.items()}
     else:
         return fn(obj)
+
+
+def query_recursively(fn: Callable[[Any], Optional[D]], obj: Any) -> Iterator[D]:
+    # We explicitly exclude str's here since they are self-referential and would cause an infinite recursion loop:
```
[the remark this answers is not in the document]
pmeier: As discussed offline, we can't avoid this since we need to be able to recurse through an unknown structure.

test/test_prototype_transforms.py
```diff
@@ -0,0 +1,184 @@
+import itertools
```
pmeier, line 1: These test are just a quick and dirty implementation to see if the transforms don't error when called with the supported inputs.

torchvision/prototype/transforms/_augment.py
```diff
@@ -0,0 +1,149 @@
...
+            raise TypeError("Scale should be a sequence")
+        if not isinstance(ratio, (tuple, list)):
+            raise TypeError("Ratio should be a sequence")
+        if (scale[0] > scale[1]) or (ratio[0] > ratio[1]):
+            warnings.warn("Scale and ratio should be of kind (min, max)")
+        if scale[0] < 0 or scale[1] > 1:
+            raise ValueError("Scale should be between 0 and 1")
+        if p < 0 or p > 1:
+            raise ValueError("Random erasing probability should be between 0 and 1")
+        # TODO: deprecate p in favor of wrapping the transform in a RandomApply
```
datumbox, line 38: This TODO is probably left from earlier when you used more abstraction. Is this still relevant given the changes?

torchvision/prototype/transforms/_augment.py
```diff
@@ -0,0 +1,149 @@
...
+        self.p = p
+        self.scale = scale
+        self.ratio = ratio
+        self.value = value
+
+    def _get_params(self, sample: Any) -> Dict[str, Any]:
+        image: Any = next(
+            query_recursively(
+                lambda input: input if input in self._DISPATCHER else None, sample  # type: ignore[no-any-return]
```
datumbox, line 47: Nit: Though the naming of the variable to `image` was helpful to understand that indeed you were fetching an image, this snippet was not immediately obvious to me how it worked. It works because the `F.erase` is only defined at the moment only for `torch.Tensor` and `features.Image`. How do you feel above requesting the first available image more explicitly? Also you can consider turning this into a utility method given it's being used a few times in the code-base.

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -0,0 +1,417 @@
+import dataclasses
+import math
+from typing import Any, Dict, Tuple, Optional, Callable, List, cast, Iterator
+
+import PIL.Image
+import torch
+from torchvision.prototype import features
+from torchvision.prototype.transforms import Transform, InterpolationMode, AutoAugmentPolicy, functional as F
+from torchvision.prototype.utils._internal import apply_recursively
+from torchvision.prototype.utils._internal import query_recursively
```
datumbox, line 10: Nit: import both on the same line.

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -0,0 +1,417 @@
...
+            .int(),
+            False,
+        ),
+        AugmentationMeta("Solarize", lambda num_bins, image_size: torch.linspace(255.0, 0.0, num_bins), False),
+        AugmentationMeta("AutoContrast", lambda num_bins, image_size: None, False),
+        AugmentationMeta("Equalize", lambda num_bins, image_size: None, False),
+        AugmentationMeta("Invert", lambda num_bins, image_size: None, False),
+    )
+    _AUGMENTATION_SPACE = {
+        augmentation_meta.dispatcher_id: augmentation_meta for augmentation_meta in _AUGMENTATION_SPACE
```
datumbox, line 137: Nit: Why don't we explicitly define the map similar to what we do at `_DISPATCHER_MAP`?

Note that the 2 share the exact same keys. In this case, you could remove the `id` from `AugmentationMeta` and rename it Magnitudes, though TBH I'm not sure why a simple tuple son't do in this case aka:
str: (lambda, bool)

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -0,0 +1,417 @@
...
+        params = params or self._get_params(sample)
+
+        image: Any = next(
+            query_recursively(
+                lambda input: input if self._is_supported(input) else None, sample  # type: ignore[no-any-return]
+            )
+        )
+        image_size = F.get_image_size(image)
+
+        for transform_id, magnitude in self._get_transforms_meta(image_size):
```
datumbox, line 265: I don't think we need to move the implementation to a separate method `_get_transforms_meta` that uses yield. Move everything on forward so that the iterative algo is one place.

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -0,0 +1,417 @@
...
+            dispatcher = self._DISPATCHER_MAP[transform_id]
+
+            def transform(input: Any) -> Any:
+                if not self._is_supported(input):
+                    return input
+
+                return dispatcher(input, magnitude, params["interpolation"], params["fill"])  # type: ignore[index]
+
+            sample = apply_recursively(transform, sample)
```
datumbox, line 274: Nit: 
In earlier versions of your code, you had been sharing `forward` but the overall abstraction didn't work for edge-cases like AugMix. Now, after implementing forward everywhere you end up having some duplicate code (lines 266-274). This is worth abstracting to a separate method called `_apply_transform(id, mangitude, params)` and moving on the base class because the application of transforms is the same across all methods (see previous implementation).

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -0,0 +1,417 @@
...
+    def __init__(self, *, num_ops: int = 2, magnitude: int = 9, num_magnitude_bins: int = 31, **kwargs: Any) -> None:
+        super().__init__(**kwargs)
+        self.num_ops = num_ops
+        self.magnitude = magnitude
+        self.num_magnitude_bins = num_magnitude_bins
+
+    def _get_transforms_meta(self, image_size: Tuple[int, int]) -> Iterator[Tuple[str, float]]:
+        for _ in range(self.num_ops):
+            augmentation_meta = self._AUGMENTATION_SPACE[int(torch.randint(len(self._AUGMENTATION_SPACE), ()))]
+            if augmentation_meta.dispatcher_id == "Identity":
```
datumbox, line 320: Nit: Add an "Identity" transform at `_DISPATCHER_MAP` of Base to avoid these if statements. Having Identity is very common on space and it will keep the keys between `_DISPATCHER_MAP` and `_AUGMENTATION_SPACE` synced.

torchvision/prototype/transforms/functional/_augment.py
```diff
@@ -7,16 +7,14 @@
...
-
 
 @dispatch(
     {
         torch.Tensor: _F.erase,
         features.Image: K.erase_image,
     }
 )
-def erase(input: T, *args: Any, **kwargs: Any) -> T:
+def erase(input: Any, *args: Any, **kwargs: Any) -> Any:
```
datumbox, line 17: Is this because often the outcome in these kernels can be `torch.Tensor` and `PIL.Image`?

datumbox (commented):
@pmeier Thanks for the patience. We are getting there, mostly questions and nits:

torchvision/prototype/transforms/_augment.py
pmeier: Yes, this is still relevant. The idea is to remove all probability arguments from the transforms in favor of wrapping them in a `RandomApply` transform. In this example this means

```py
transforms.RandomErasing(p=..., ...)
```

would change to

```py
transforms.RandomApply(
    transforms.RandomErasing(...),
    p=...,
)
```

Another example is to no longer have a `RandomHorizontalFlip`. The term "Random" in the name of a transform should mean "random parameters sampled at runtime".

torchvision/prototype/transforms/functional/_augment.py
pmeier: Yes.

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -0,0 +1,417 @@
...
+                (("ShearX", 0.7, 2), ("Invert", 0.1, None)),
+            ]
+        else:
+            raise ValueError(f"The provided policy {policy} is not recognized.")
+
+    def _get_transforms_meta(self, image_size: Tuple[int, int]) -> Iterator[Tuple[str, float]]:
+        policy = self._policies[int(torch.randint(len(self._policies), ()))]
+
+        for dispatcher_id, probability, magnitude_idx in policy:
+            if not float(torch.rand(())) <= probability:
```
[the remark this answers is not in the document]
pmeier: Being more explicit. I'll remove.

torchvision/prototype/transforms/_utils.py
```diff
@@ -0,0 +1,19 @@
+from typing import Any, Union, Optional
+
+import PIL.Image
+import torch
+from torchvision.prototype import features
+from torchvision.prototype.utils._internal import query_recursively
+
+
+def query_image(sample: Any) -> Union[PIL.Image.Image, torch.Tensor, features.Image]:
```
datumbox, line 9: Nit: Shall we put it next to `query_recursively`?

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -0,0 +1,356 @@
...
+        keys = tuple(dct.keys())
+        key = keys[int(torch.randint(len(keys), ()))]
+        return key, dct[key]
+
+    def _apply_transform(self, sample: Any, params: Dict[str, Any], transform_id: str, magnitude: float) -> Any:
+        dispatcher = self._DISPATCHER_MAP[transform_id]
+
+        def transform(input: Any) -> Any:
+            if not self._is_supported(input):
+                return input
```
datumbox, line 107: Transformations such as Rotate, Shear and Translate don't currently know how to transform masks and bounding boxes so if someone tries to the AutoAugment methods for Detection/Segmentation it will mess up the target data. 

What is the recommended way for detecting that a specific operator doesn't support the input? This is not a problem on the AutoAugment algorithms, this can be an issue for any Transform that can't handle a specific type.

datumbox (approved):
I think the proposed solution is a good candidate. I left one nit and one comment which I would address on a separate PR (to reduce the complexity of this one). 

Also Have you checked that all transforms work the way we expect them to work? If not, we should create an issue to review them and verify individually but I wouldn't block this merge as it focuses on the API rather than the functionality.

torchvision/prototype/transforms/_utils.py
pmeier: That would mean that `torchvision.prototype.utils` depends on `torchvision.prototype.features`. Just checking if we want that, because right now `torchvision.utils` is independent of the other submodules. `torchvision.prototype.{datasets, transforms}` need to depend on `torchvision.prototype.features` anyway, so we don't need to take this into consideration there.

datumbox: Sounds good. Let's not block and decide later.

torchvision/prototype/transforms/_auto_augment.py
datumbox: I agree that this is something that should be handled on the Transform Class.

I'm OK with option 1 for the current implementation. Could you please do it on a separate PR?

Option 2, could be more appropriate for the alternative approach that removes completely the high-level kernels and moves the whole dispatch to the Transform Classes. I strongly believe we shouldn't try to go for support of universal types here but instead limit ourselves to know types (_Feature subclasses, PIL.Image, Tensor) + primitives (int, string, float etc). This will address your concern for paths but at the same time won't open us up to unpredicted behaviours.

PR #5421, opened 2022-02-14
Transforms without dispatcher

pmeier:
This is the alternative proposal to #5418. The main difference is that this PR eliminates the high-level kernels aka dispatchers introduced in #5407 by merging it into the transforms. That means in contrast to #5418, in this PR transforms are responsible for the dispatch.

### Pros

- One reason why #5407 had such a lengthy discussion is that it is really hard to properly document the function of a dispatcher. The approach in this PR completely circumvents this issue since the expectation of calling a transform is very different from calling a high-level kernel. This means, we can simply link the low-level kernels, but don't need to specify how they are called exactly.
- By removing the high-level kernels we shrink our public API surface that we have to maintain / test / document / ... Although the functionally is equivalent equivalent, this PR removes ~500 LoC.
- The dispatch mechanism from #5407 involves a lot of magic to enable reasonable flexibility. Still, in some cases (I intentionally avoid "edge cases" here, because I don't think they are this uncommon), we need even more complicated logic to handle everything. Out of the 68 registered kernels, 12 need manual handling (~17%).
- As noted in https://github.com/pytorch/vision/pull/5418#discussion_r805943560, the other design requires `get_params` to return *all* parameters that are needed, whereas in this PR we can again only generate the dynamic ones there and simply access the constant ones.

### Cons

- By removing the magic of the dispatchers, we will need to write more boilerplate code ourselves. Note that this does not necessarily translates to more LoC. Compare

    ```py
    @dispatch(
        {
            torch.Tensor: _F.resize,
            PIL.Image.Image: _F.resize,
            features.Image: K.resize_image,
            features.SegmentationMask: K.resize_segmentation_mask,
            features.BoundingBox: None,
        }
    )
    def resize(input: Any, *args: Any, **kwargs: Any) -> Any:
        """TODO: add docstring"""
        if isinstance(input, features.BoundingBox):
            size = kwargs.pop("size")
            output = K.resize_bounding_box(input, size=size, image_size=input.image_size)
            return features.BoundingBox.new_like(input, output, image_size=size)
    
        raise RuntimeError
    
    class Resize(Transform):
        _DISPATCHER = F.resize
    ```
    
    to
    
    ```py
    class Resize(Transform):
            if type(input) is torch.Tensor or isinstance(input, PIL.Image.Image):
                return _F.resize(input, size=self.size, interpolation=self.interpolation)
            elif type(input) is features.Image:
                output = K.resize_image(input, size=self.size, interpolation=self.interpolation)
            elif type(input) is features.SegmentationMask:
                output = K.resize_segmentation_mask(input, size=self.size)
                return features.SegmentationMask.new_like(input, output)
            elif type(input) is features.BoundingBox:
                output = K.resize_bounding_box(input, size=self.size, image_size=input.image_size)
                return features.BoundingBox.new_like(input, output, image_size=self.size)
            else:
                return input
    ```

- Logging API calls is easier with dispatchers since the call can be placed inside the decorator. Since this layer of the API is removed here, but we still want to log the calls to the kernels, we need to perform the logging there. #5424 investigates if this is possible through a decorator, which would weaken this con.
- Writing custom composite transforms is harder. With the dispatchers we can do something like 

    ```py
        input = F.foo(input)
        input = F.bar(input)
        input = F.baz(input)
        return input
    ```
    
    Without dispatchers we could either call the low level kernels directly
    
    ```py
        if isinstance(input, features.Image):
            input = K.foo_image(input)
            input = K.bar_image(input)
            input = K.baz_image(input)
        elif isinstance(input, features.BoundingBox):
            input = K.foo_bounding_box(input)
            input = K.bar_bounding_box(input)
            input = K.baz_bounding_box(input)
    
        return input
    ```
    
    or rely on the transforms objects
    
    ```py
        input = transforms.Foo()(input)
        input = transforms.Bar()(input)
        input = transforms.Baz()(input)
        return input
    ```
    
    Out of the two options, I prefer the latter, since we have less duplicated code especially if the dispatch is not as straight forward as in the example.
- (This only applies if we go with option 2 form the point above) If we use other transforms to write composite ops, we need to have a primitive for every transforms that does not have any random functionality. For example, we would need to have a `transforms.Affine` as well as a `tranforms.RandomAffine`. From an implementation perspective this is fairly easy to achieve with `RandomAffine` subclassing `Affine` and overwriting the `_get_params()` method with random sampling. This also ties directly into https://github.com/pytorch/vision/pull/5418#discussion_r807267334


pmeier:
After some offline discussion with @vfdev-5, it became clear that always having a primitive transform will significantly increase the number of transforms that we provide. Thus, although this PR removes the high-level kernels and thus effectively turns 3 API layers into 2, the surface reduction is not as significant as originally thought. There will still be some reduction, since some transforms are already primitives, e.g. `Resize`, and thus no extra transform is needed there.

torchvision/prototype/transforms/_container.py
```diff
@@ -25,38 +25,38 @@ def __init__(self, transform: Transform, *, p: float = 0.5) -> None:
         self.transform = transform
         self.p = p
 
-    def forward(self, *inputs: Any, params: Optional[Dict[str, Any]] = None) -> Any:
+    def forward(self, *inputs: Any) -> Any:
```
datumbox, line 28: Question: I assume this is removed for the same reasons `Compose` doesn't support it. Aka it's hard to pass parameters when you have many transforms. Right?

torchvision/prototype/transforms/_misc.py
```diff
@@ -32,15 +34,17 @@ def extra_repr(self) -> str:
...
         self.std = std
 
-    def _get_params(self, sample: Any) -> Dict[str, Any]:
-        return dict(mean=self.mean, std=self.std)
+    def _transform(self, input: Any, params: Dict[str, Any]) -> Any:
+        if type(input) is torch.Tensor:
+            return _F.normalize(input, mean=self.mean, std=self.std)
+        if type(input) is features.Image:
+            output = K.normalize_image(input, mean=self.mean, std=self.std)
+            return features.Image.new_like(input, output)
```
datumbox, line 47: Hm... Should it be an image after Normalization? The colour scheme will not be valid anymore (RGB 0-255 assumption). I think we should be returning a Tensor here, not a `features.Image`.

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -1,16 +1,36 @@
...
 from torchvision.prototype.utils._internal import apply_recursively
+from torchvision.transforms import functional as _F
 
-from ._utils import query_image
+from ._utils import query_image, get_image_size, get_image_num_channels
 
-K = TypeVar("K")
-V = TypeVar("V")
+
+@dataclasses.dataclass
```
datumbox, line 15: It seems like this version of AutoAugment is not the one we have on main. Many of the previously requested changes are undone here. I'm not going to further review this file and chat offline for its status.

datumbox (commented):
Overall I like this approach a lot. I've added a few comments, let's sync offline for next steps.

torchvision/prototype/transforms/_auto_augment.py
datumbox: We will also need to include AugMix: #5411

torchvision/prototype/transforms/_container.py
pmeier: That is one reason. Plus, the previous design handled the complete parameter passing through the `get_params` method. Since that is now private and only handles the dynamic part of the parameters, accepting a `params` argument for the call opens up more problems than it solves.

torchvision/prototype/transforms/_misc.py
pmeier: That is one possibility. Another would be to use 

https://github.com/pytorch/vision/blob/e88a55493eadaab15d1df432fe248e70e04d29e4/torchvision/prototype/features/_image.py#L16-L17

Anyway, I agree, the output should not be an RGB image.

datumbox: There are some PyTorch features (such as the new FLOP estimation) that we will not be able to use if we subclass tensor. Let's remove the `Image` wrapping for now.

torchvision/prototype/transforms/_container.py
datumbox: I spoke with @NicolasHug about `get_params()` being private and he mentioned a couple of valid use cases on why we might want to keep it public. We can review this later but I agree that removing it from here is probably for the best.

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -89,24 +32,111 @@ def _get_params(self, sample: Any) -> Dict[str, Any]:
...
         return key, dct[key]
 
-    def _apply_transform(self, sample: Any, params: Dict[str, Any], transform_id: str, magnitude: float) -> Any:
-        dispatcher = self._DISPATCHER_MAP[transform_id]
-
-        def transform(input: Any) -> Any:
-            if type(input) in {features.Image, torch.Tensor} or isinstance(input, PIL.Image.Image):
-                return dispatcher(input, magnitude, params["interpolation"], params["fill"])
+    def _apply_transform(self, sample: Any, transform_id: str, magnitude: float) -> Any:
+        def dispatch(image_kernel: Callable, legacy_kernel: Callable, input: Any, *args: Any, **kwargs: Any) -> Any:
```
datumbox, line 43: Any chance we can simplify this?

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -115,10 +145,10 @@ class AutoAugment(_AutoAugmentBase):
         "ShearX": (lambda num_bins, image_size: torch.linspace(0.0, 0.3, num_bins), True),
         "ShearY": (lambda num_bins, image_size: torch.linspace(0.0, 0.3, num_bins), True),
         "TranslateX": (lambda num_bins, image_size: torch.linspace(0.0, 150.0 / 331.0 * image_size[1], num_bins), True),
-        "TranslateY": (lambda num_bins, image_size: torch.linspace(0.0, 150.0 / 331.0 * image_size[1], num_bins), True),
+        "TranslateY": (lambda num_bins, image_size: torch.linspace(0.0, 150.0 / 331.0 * image_size[0], num_bins), True),
```
datumbox, line 148: Oh snap. How did we end up with the wrong value here? I guess we will need to do a very thorough check of every line of code.

Also if you check the main area, the `image_size[0]` is used for TranslateX. Did you flip the size tuple?

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -115,10 +145,10 @@ class AutoAugment(_AutoAugmentBase):
...
         "Rotate": (lambda num_bins, image_size: torch.linspace(0.0, 30.0, num_bins), True),
         "Brightness": (lambda num_bins, image_size: torch.linspace(0.0, 0.9, num_bins), True),
-        "Color": (lambda num_bins, image_size: torch.linspace(0.0, 0.9, num_bins), True),
+        "Saturation": (lambda num_bins, image_size: torch.linspace(0.0, 0.9, num_bins), True),
```
datumbox, line 151: Though your proposed names describe more accurately what the ops do, we should revert all the name changes to their original values. This is to align with all the papers that use the same terminology. This originally came from the [AutoAugment](https://arxiv.org/pdf/1805.09501.pdf) paper at page 13.

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -222,15 +252,14 @@ def _get_policies(
...
-                (("ShearY", 0.8, 5), ("AutoContrast", 0.7, None)),
+                (("ShearY", 0.8, 5), ("Autocontrast", 0.7, None)),
                 (("ShearX", 0.7, 2), ("Invert", 0.1, None)),
             ]
         else:
             raise ValueError(f"The provided policy {policy} is not recognized.")
 
     def forward(self, *inputs: Any, params: Optional[Dict[str, Any]] = None) -> Any:
         sample = inputs if len(inputs) > 1 else inputs[0]
-        params = params or self._get_params(sample)
```
datumbox, line 233: Do we need the params input on the method at all?

pmeier: Nope. I already removed it for all other transforms, but forgot about it here.

torchvision/prototype/transforms/_auto_augment.py
pmeier: > How did we end up with the wrong value here?

Most likely copy-pasting stuff around.

> I guess we will need to do a very thorough check of every line of code.

I'll do another pass.

> Also if you check the main area, the `image_size[0]` is used for TranslateX. Did you flip the size tuple?

Yes. `image_size` is always `(height, width)`. In our old implementation we are flip-flopping around with the convention and for auto augment we are using `(width, height)`. Thus, the index swap is correct here.

torchvision/prototype/transforms/_auto_augment.py
pmeier: Could you elaborate what you want to be simplified? I don't think there is room for significant simplification without introducing more abstractions.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -1,165 +1,73 @@
...
+
+def resize_image(
+    image: torch.Tensor,
+    size: List[int],
+    interpolation: InterpolationMode = InterpolationMode.BILINEAR,
+    max_size: Optional[int] = None,
+    antialias: Optional[bool] = None,
+) -> torch.Tensor:
+    new_height, new_width = size
+    num_channels, old_height, old_width = image.shape[-3:]
```
datumbox, line 37: Above the `horizontal_flip_image` kernel uses the old `_F.hflip` which handles both Images and PIL. Nevertheless, here the `resize_image` doesn't handle PIL. This is somehow inconsistent. We shouldn't have half of the methods supporting PIL and half not doing so.

Moreover I think we need to phase out the idiom `num_channels, old_height, old_width = image.shape[-3:]` from the code in favour of calling the correct kernels such as `get_image_size` and `get_image_num_channels`.

torchvision/prototype/transforms/_auto_augment.py
datumbox: As discussed there is a bug on the way we handle sizes. Even our methods sometimes return h,w and others w,h. Example:
https://github.com/pytorch/vision/blob/25e2ec0dac86a37d2e0083a3432ca3594f8185e4/torchvision/prototype/transforms/functional/_utils.py#L11

I think we should avoid estimating the image size from the tensor or the image attributes and instead rely to the get_image_size kernels everywhere so that at least we are consistent.

torchvision/prototype/transforms/_auto_augment.py
datumbox: Let's have a chat offline to see pros/cons.

torchvision/prototype/transforms/_meta_conversion.py
```diff
@@ -28,22 +29,49 @@ def __init__(self, dtype: torch.dtype = torch.float32) -> None:
...
+
     def _transform(self, input: Any, params: Dict[str, Any]) -> Any:
-        if type(input) is features.Image:
-            output = F.convert_color_space(input, old_color_space=input.color_space, new_color_space=self.color_space)
+        if isinstance(input, features.Image):
+            output = F.convert_image_color_space_tensor(
+                input, old_color_space=input.color_space, new_color_space=self.color_space
+            )
             return features.Image.new_like(input, output, color_space=self.color_space)
+        if isinstance(input, torch.Tensor):
```
datumbox, line 60: Nit: Use elifs here similar to everywhere else on the code for consistency.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -26,17 +29,17 @@ def horizontal_flip_bounding_box(
     ).view(shape)
 
 
-def resize_image(
+def resize_image_tensor(
     image: torch.Tensor,
     size: List[int],
-    interpolation: InterpolationMode = InterpolationMode.BILINEAR,
+    interpolation: str = "nearest",
```
datumbox, line 35: First of all we should use bilinear as the default, not nearest. This is already the default on functional_tensor.resize.

Second I believe we should move towards Enums here. I think one reason we don't support them here is for JIT scriptability but now enums are scriptable. I would change it here and perhaps on a separate PR, I would update the tensor functional to use enums (it is currently considered private so we can do it). Thoughts?

torchvision/prototype/transforms/functional/_geometry.py
datumbox: Reminder to change the image size calculations to always call the above kernels.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -45,29 +48,124 @@ def resize_image(
...
+
+
 def resize_segmentation_mask(
-    segmentation_mask: torch.Tensor,
-    size: List[int],
-    max_size: Optional[int] = None,
+    segmentation_mask: torch.Tensor, size: List[int], max_size: Optional[int] = None
 ) -> torch.Tensor:
-    return resize_image(segmentation_mask, size=size, interpolation=InterpolationMode.NEAREST, max_size=max_size)
+    return resize_image_tensor(segmentation_mask, size=size, interpolation="nearest", max_size=max_size)
```
datumbox, line 57: same here.

torchvision/prototype/transforms/functional/_utils.py
```diff
@@ -1,14 +1,17 @@
...
 from torchvision.prototype import features
 from torchvision.transforms import functional as _F
 
+# FIXME
+
 
 def get_image_size(image: Union[PIL.Image.Image, torch.Tensor, features.Image]) -> Tuple[int, int]:
     if type(image) is torch.Tensor or isinstance(image, PIL.Image.Image):
-        return cast(Tuple[int, int], tuple(_F.get_image_size(image)))
+        width, height = _F.get_image_size(image)
```
datumbox, line 13: We should not be calling the high-level kernel here `_F` but instead dispatch separately like everywhere else for consistency reasons. (also because this API you introduce will replace `F`)

torchvision/prototype/transforms/functional/_utils.py
```diff
@@ -1,14 +1,17 @@
...
+        return height, width
```
datumbox, line 14: This is a major BC breaking change. Keeping the same method name and flipping the order is guaranteed to cause confusion.

My recommendation is to either undo OR offer a completely new method that returns the image_dims including channels in the order that you like.

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -22,37 +22,47 @@ def __init__(
...
+            image_pil_kernel: Callable,
+            input: Any,
+            *args: Any,
+            **kwargs: Any,
+        ) -> Any:
+            if isinstance(input, (features.BoundingBox, features.SegmentationMask)):
+                raise TypeError(f"{type(input).__name__}'s are not supported by {type(self).__name__}()")
+            elif isinstance(input, features.Image):
+                if "interpolation" in kwargs:
+                    kwargs["interpolation"] = kwargs["interpolation"].value
```
datumbox, line 42: Another str vs enum instance here

torchvision/prototype/datasets/_builtin/caltech.py
```diff
@@ -95,7 +95,7 @@ def _prepare_sample(
             bounding_box=BoundingBox(
                 ann["box_coord"].astype(np.int64).squeeze()[[2, 0, 3, 1]], format="xyxy", image_size=image.image_size
             ),
-            contour=_Feature(ann["obj_contour"].T),
```
datumbox, line 98: What is this change?

datumbox (commented):
I think it's looking very good. Few more comments:

torchvision/prototype/datasets/_builtin/caltech.py
pmeier: It seems my IDE was overly ambitious when renaming variables. I'll revert this.

torchvision/prototype/transforms/functional/_geometry.py
pmeier: We have warnings like 

https://github.com/pytorch/vision/blob/761d09f1c3dd94ffe489f2c08a72a719f996a93d/torchvision/transforms/transforms.py#L255-L261

since `0.9`. If we strictly stick to the new [BC policy](https://github.com/pytorch/pytorch/wiki/PyTorch's-Python-Frontend-Backward-and-Forward-Compatibility-Policy), we can only remove if the warning includes the version this is going to be removed. This is why we merged #5387 and others before this release to be able to remove with `0.14`. If we also use this scheme here, we could only remove in `0.15`.

However, the spirit of the BC policy is to keep the behavior for two releases or 180 days in the nightlies. Since we are way past this, I think we can go for the removal here given that it would simplify the implementation quite a bit. cc @NicolasHug @mruberry to make sure.

datumbox: @pmeier Dont take what I say as disagreement on the deprecation policy stuff. It's unrelated.

What I say here is that the `functional_tensor.py` file was considered private. This was done several releases back. So due to this, you can do all changes as you want without BC concerns.

mruberry: I would also consider all BC-breaking changes made before the latest policy as "grandfathered into" whatever old policies (or lack of old policies) existed. That said, when those changes are public they could be documented following the new policy's suggestions.

torchvision/prototype/transforms/functional/_meta_conversion.py
```diff
@@ -1,50 +1,91 @@
...
+    if new_format == BoundingBoxFormat.XYWH:
+        bounding_box = _xyxy_to_xywh(bounding_box)
+    elif new_format == BoundingBoxFormat.CXCYWH:
+        bounding_box = _xyxy_to_cxcywh(bounding_box)
+
+    return bounding_box
+
+
+def _grayscale_to_rgb_tensor(grayscale: torch.Tensor) -> torch.Tensor:
+    return grayscale.expand(3, 1, 1)
```
datumbox, line 57: Non-blocking TODO: This kernel assumes the tensor is a single image and not a batch. We shouldn't make this assumption. We should fix on a separate PR.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -1,165 +1,318 @@
...
+
+
+def rotate_image_tensor(
+    img: torch.Tensor,
+    angle: float,
+    interpolation: InterpolationMode = InterpolationMode.NEAREST,
+    expand: bool = False,
+    fill: Optional[List[float]] = None,
+    center: Optional[List[float]] = None,
+) -> torch.Tensor:
```
datumbox, line 186: @vfdev-5 FYI some transforms have been copypasted here to make refactoring easier. So any changes on main area need to be made also here.

datumbox (commented):
Nits and notes:

datumbox (approved):
I think this is a major improvement over the previous proposal. 

We should merge. I'll follow up with a more thorough review on the code-base that we have on the prototype transforms.

torchvision/prototype/transforms/functional/_geometry.py
datumbox: @pmeier I was wrong here.

The new get_image_size/get_image_num_channels that live in utilities are not jit-scriptable (due to the fact they know about PIL and `Image`). So here you will either use the `_FT` versions (which is what you did) or do `num_channels, old_height, old_width = image.shape[-3:]` as you proposed originally.

I think the latter makes more sense once and I'll restore it on a follow up PR.

PR #5424, opened 2022-02-15
try api call logging with decorator

pmeier:
We have a lot of these sprinkled in throughout our code base:

https://github.com/pytorch/vision/blob/6fcf0a27e38dced48f297175d8bdb7b2a5d2f2ea/torchvision/transforms/functional.py#L419-L420

This is almost a text book case for using a decorator. I was always told that decorators are not supported by `torch.jit.script` and thus we have to go the "long way". But it seems decorators are supported

```py
import functools

import torch


def logger(fn):
    @functools.wraps(fn)
    def wrapper(*args, **kwargs):
        print(f"Logging call to {fn.__name__}")
        return fn(*args, **kwargs)

    return wrapper


@logger
def foo(input: torch.Tensor) -> torch.Tensor:
    print("inside foo()")
    return input.abs()


input = torch.tensor(-1)
print("eager")
print(foo(input))
print("scripted")
print(torch.jit.script(foo)(input))
```

```
eager
Logging call to foo
inside foo()
tensor(1)
scripted
inside foo()
tensor(1)
```

Since we get no print from the logger during scripted mode, maybe it picks up on the `functools.wraps` decorator and only executes the underlying function.

### DIFF
--- a/torchvision/prototype/transforms/_presets.py
+++ b/torchvision/prototype/transforms/_presets.py
@@ -1,4 +1,4 @@
-from typing import Dict, Optional, Tuple, cast
+from typing import Dict, Optional, Tuple
 
 import torch
 from torch import Tensor, nn
@@ -41,7 +41,7 @@ class ImageNetEval(nn.Module):
             img = F.pil_to_tensor(img)
         img = F.convert_image_dtype(img, torch.float)
         img = F.normalize(img, mean=self._mean, std=self._std)
-        return cast(Tensor, img)
+        return img
 
 
 class Kinect400Eval(nn.Module):
@@ -65,7 +65,7 @@ class Kinect400Eval(nn.Module):
         vid = F.resize(vid, self._size, interpolation=self._interpolation)
         vid = F.center_crop(vid, self._crop_size)
         vid = F.convert_image_dtype(vid, torch.float)
-        vid = cast(Tensor, F.normalize(vid, mean=self._mean, std=self._std))
+        vid = F.normalize(vid, mean=self._mean, std=self._std)
         return vid.permute(1, 0, 2, 3)  # (T, C, H, W) => (C, T, H, W)
 
 
--- a/torchvision/prototype/transforms/_presets.py
+++ b/torchvision/prototype/transforms/_presets.py
@@ -1,4 +1,4 @@
-from typing import Dict, Optional, Tuple
+from typing import Dict, Optional, Tuple, cast
 
 import torch
 from torch import Tensor, nn
@@ -41,7 +41,7 @@ class ImageNetEval(nn.Module):
             img = F.pil_to_tensor(img)
         img = F.convert_image_dtype(img, torch.float)
         img = F.normalize(img, mean=self._mean, std=self._std)
-        return img
+        return cast(Tensor, img)
 
 
 class Kinect400Eval(nn.Module):
@@ -65,7 +65,7 @@ class Kinect400Eval(nn.Module):
         vid = F.resize(vid, self._size, interpolation=self._interpolation)
         vid = F.center_crop(vid, self._crop_size)
         vid = F.convert_image_dtype(vid, torch.float)
-        vid = F.normalize(vid, mean=self._mean, std=self._std)
+        vid = cast(Tensor, F.normalize(vid, mean=self._mean, std=self._std))
         return vid.permute(1, 0, 2, 3)  # (T, C, H, W) => (C, T, H, W)
 
 
--- a/torchvision/transforms/functional.py
+++ b/torchvision/transforms/functional.py
@@ -21,9 +21,9 @@ from . import functional_pil as F_pil
 from . import functional_tensor as F_t
 
 
-def log_api_usage_once(fn: Callable) -> Callable:
+def log_api_usage_once(fn: Callable[..., Tensor]) -> Callable[..., Tensor]:
     @functools.wraps(fn)
-    def wrapper(*args, **kwargs):
+    def wrapper(*args: Any, **kwargs: Any) -> Tensor:
         _log_api_usage_once(fn)
         return fn(*args, **kwargs)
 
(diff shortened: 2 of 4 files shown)


pmeier:
@kazhang in #5052 you stated that one cannot

> * use decorator
>   * For example: `@log_api_usage` in #4976
>   * doesn’t work with TorchScript since decorator needs to use kwargs, which is not supported in TorchScript;

to log the API calls. Either there was support added or I'm missing something here, since it seems to work out fine. I've added the `_log_api_usage_once` call to a decorator and applied it to all our functional transforms. JIT scriptability is tested by our test suite, but there no errors.

kazhang:
Hi @pmeier , thanks for the PR! looks like decorator (more specifically `kwargs`) is supported now. Initially I tried it out in #4976 but got [this error from JIT](https://github.com/pytorch/vision/pull/4976#issuecomment-977057707). I'm glad that it works now!

torchvision/transforms/functional.py
```diff
@@ -19,6 +21,15 @@
 from . import functional_tensor as F_t
 
 
+def log_api_usage_once(fn: Callable) -> Callable:
```
kazhang, line 24: shall we move this decorator to utils? or turn `_log_api_usage_once` into a decorator if it works for class as well?

torchvision/prototype/transforms/_presets.py
```diff
@@ -41,7 +41,7 @@ def forward(self, img: Tensor) -> Tensor:
             img = F.pil_to_tensor(img)
         img = F.convert_image_dtype(img, torch.float)
         img = F.normalize(img, mean=self._mean, std=self._std)
-        return img
+        return cast(Tensor, img)
```
datumbox, line 44: Why is this necessary? The return type of the functional API is `Tensor` so this casting seems unnecessary. Here is how normalize looks for example:
```python
def normalize(tensor: Tensor, mean: List[float], std: List[float], inplace: bool = False) -> Tensor:
```

torchvision/prototype/transforms/_presets.py
```diff
@@ -65,7 +65,7 @@ def forward(self, vid: Tensor) -> Tensor:
         vid = F.resize(vid, self._size, interpolation=self._interpolation)
         vid = F.center_crop(vid, self._crop_size)
         vid = F.convert_image_dtype(vid, torch.float)
-        vid = F.normalize(vid, mean=self._mean, std=self._std)
+        vid = cast(Tensor, F.normalize(vid, mean=self._mean, std=self._std))
```
datumbox, line 68: Same here.

torchvision/transforms/functional.py
datumbox: @pmeier I believe if you add this on the functional side, it will break JIT-scriptability. At least that's what I recall from Kai's earlier experiments (see #4976 and #5052). But then the CI tests are not currently failing which is very strange.

@kazhang any thoughts on whether the approach followed by Philip actually works OR if we have a gap on our testing strategy?

datumbox (changes requested):
I'm going to mark as blocked to avoid any accidental merges because it's possible that this PR breaks JIT-scriptability without the tests catching it. If that's not the case, then that's great and we can unblock. We just need to do a bit more due diligence to ensure this is not due to a gap in our testing strategy.

torchvision/prototype/transforms/_presets.py
pmeier: The problem is that the decorator is annotated with

```py
def log_api_usage_once(fn: Callable) -> Callable:
```

This means `mypy` assumes `Callable` as full annotation for the decorated function. This is roughly equivalent to

```py
def foo(*args: Any, **kwargs: Any) -> Any
```

If we restrict the decorator for the functional transforms for now, we can fix this by annotating with `Callable[..., torch.Tensor]`.

datumbox: Thanks @pmeier, makes sense. No need to change then.

torchvision/transforms/functional.py
```diff
@@ -18,6 +19,17 @@
 from . import functional_pil as F_pil
 from . import functional_tensor as F_t
 
+F = TypeVar("F", bound=Callable[..., Any])
```
pmeier, line 22: This is the [official way](https://mypy.readthedocs.io/en/stable/generics.html#declaring-decorators) to declare decorators that preserve the signature. Let's see if the test suite is still happy.

torchvision/transforms/functional.py
kazhang: Now I remember why decorator didn't work for me in #4976
See [my comment](https://github.com/pytorch/vision/pull/4976#issuecomment-976039784):
> Unfortunately, torchscript doesn't like decorator. When using the decorator on boxes functions, JIT looks for helper functions definition(e.g. _upcast) in utils.py


In this PR, we define the decorator in `functional.py` instead of `utils.py`, so JIT is still able to find helper functions used by the function to be scripted. So basically we'd need to define this decorator everywhere, not too bad compared with current solution....

kazhang: In #5436 I tried to move the decorator to utils.py, and reproduced the error: https://app.circleci.com/pipelines/github/pytorch/vision/14851/workflows/ad9beb44-ba9e-472e-a71b-8775e909f8b5/jobs/1195207

datumbox:
I spoke with @gmagogsfm and he suspects that TorchScript completely ignored the logger decorator. He said that this is likely an unintended use case and suggested it is not safe to depend on it.

torchvision/transforms/functional.py
pmeier: Looking through the issue tracker, I found the long standing issue pytorch/pytorch#38964 that should have been fixed in pytorch/pytorch#63248. Maybe @qihqi can have a look if the fix was incomplete or our setup is different?

vadimkantorov:
related:https://github.com/pytorch/vision/issues/4957

PR #5426, opened 2022-02-15
add feat_dim attribute for ResNet

towzeur:
Added ResNet.feat_dim which is the features dimension just before the `avgpool` layer

```python
resnets_func = [
    resnet18,
    resnet34,
    resnet50,
    resnet101,
    resnet152,
    resnext50_32x4d,
    resnext101_32x8d,
    wide_resnet50_2,
    wide_resnet101_2,
]
for resnet_func in resnets_func:
    resnet = resnet_func(pretrained=False, progress=False)
    print(resnet_func.__name__, resnet.feat_dim)
```

```python
resnet18 512
resnet34 512
resnet50 2048
resnet101 2048
resnet152 2048
resnext50_32x4d 2048
resnext101_32x8d 2048
wide_resnet50_2 2048
wide_resnet101_2 2048
```

### DIFF
--- a/torchvision/models/resnet.py
+++ b/torchvision/models/resnet.py
@@ -75,6 +75,7 @@ class BasicBlock(nn.Module):
             raise ValueError("BasicBlock only supports groups=1 and base_width=64")
         if dilation > 1:
             raise NotImplementedError("Dilation > 1 not supported in BasicBlock")
+        self.dim = planes * self.expansion
         # Both self.conv1 and self.downsample layers downsample the input when stride != 1
         self.conv1 = conv3x3(inplanes, planes, stride)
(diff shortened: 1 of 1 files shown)


datumbox (changes requested):
@towzeur Thanks for the contribution and welcome to TorchVision.

Typically code-changes that modify the API, [should be discussed](https://github.com/pytorch/vision/blob/main/CONTRIBUTING.md#tldr) on a separate issue prior opening the PR. Unfortunately we can't merge this because:
- The attribute name used here is not consistent with the equivalents of other models
- The PR modifies the public interface of the Blocks and requires them to have a `dim` attribute so that the `feat_dim` can be defined. This is BC breaking for those who use custom block classes.

PR #5435, opened 2022-02-16
Adding Scale Jitter transform for detection

datumbox:
Implemented as described at the [CopyPaste](https://arxiv.org/pdf/2012.07177.pdf) paper. Supports Large Scale Jitter and Standard Scale Jitter.


references/detection/transforms.py
```diff
@@ -282,3 +282,47 @@ def forward(
...
+    Args:
+        scale_range (tuple of ints): scaling factor interval, e.g (a, b), then scale is randomly sampled from the
+            range a <= scale <= b.
+        interpolation (InterpolationMode): Desired interpolation enum defined by
+            :class:`torchvision.transforms.InterpolationMode`. Default is ``InterpolationMode.BILINEAR``.
+    """
+
+    def __init__(
+        self,
+        scale_range: Tuple[float, float] = (0.1, 2.0),
```
fmassa, line 301: This is fine as is, but this means user will need to apply a `Resize(target_size)` before applying `ScaleJitter`.

This is fine and probably more composable than having this function resize internally as well, although it could be slightly inefficient as we will be performing two interpolations of the image.

references/detection/transforms.py
```diff
@@ -282,3 +282,47 @@ def forward(
...
+        old_width, old_height = F.get_image_size(image)
+
+        r = self.scale_range[0] + torch.rand(1) * (self.scale_range[1] - self.scale_range[0])
+        new_width = int(old_width * r)
+        new_height = int(old_height * r)
+
+        image = F.resize(image, [new_height, new_width], interpolation=self.interpolation)
+
+        if target is not None:
+            target["boxes"] *= r
```
fmassa, line 326: Might be good to ensure that the user don't pass this transform to segmentation problems, otherwise they will have wrong results?

fmassa (approved):
LGTM, thanks!

references/detection/transforms.py
datumbox: Do you mean "apply resize AFTER"?

references/detection/transforms.py
datumbox: Hmmm... This is a temporary transform to unblock some of the training until the new Transform API is ready. Nevertheless I understand that here you refer to the semantic segmentation masks for mask rcnn, right?

references/detection/transforms.py
datumbox: I spoke offline with @fmassa and we decided to refactor this and add an extra `target_size` parameter in this transform to make it stand-alone.

references/detection/transforms.py
datumbox: Worth fixing.

references/detection/transforms.py
datumbox: More handling of the bbox case was needed to support the `target_size`. This is fixed at #5446

Issue #5430, opened 2022-02-16
ConvNormActivation block for 3D convolutions


jdsgomes:
### 🚀 The feature

I suggest we add a `Conv3d` equivalent for the  `ConvNormActivation` block

### Motivation, pitch

This will be used in the upcoming ResNext3d implementation. 

### Alternatives

I see two possible alternatives:

1. We can re-purpose the existing method in a BC way  by adding an extra named parameter `conv layer` with default value of `Conv2D`. 

2. We can explicitly add a new class `Conv3DNormActivation`, which will be very similar to the existing one.

oke-aditya:
In my opinion 2nd one is better. 

I can give a shot at this :)

jdsgomes:
Thanks for the feedback @oke-aditya!

I was thinking and discussing with @datumbox offline, and I'm inclined to the following solution:

1. Create an Alias for the existing `ConvNormActivation` and call it `Conv2dNormActivation`. (this will be done in a separate PR. I would also add a deprecation warning for the ConvNormActivation in this case.
2. Create a new method for 3D convolutions named `Conv3dNormActivation`, which seems to be your preferred method too. 

The advantages of this approach is that we have a clear naming convention in line with the `nn.Conv2d` and `nn.Conv3d`  from pytorch, and we wouldn't have to validate invalid combinations of `norm_layer` and  the new `conv_layer` paramter in `ConvNormActivation`.

The major disadvantage is that the current `ConvNormActivation` is already used in [9 models](https://github.com/pytorch/vision/search?q=ConvNormActivation) inside TorchVision and [numerous places](https://github.com/search?p=3&q=ConvNormActivation&type=Code) externally but I think this can be mitigated with the deprecation warning and fazing out in future releases.

jdsgomes:
Also, thank you for offering to give it a shot, let's just wait for some more feedback and then would be great if you wish to work on it!

jdsgomes:
@oke-aditya I kicked off by renaming the `ConvNormActivation` as we all seem to be aligned. Looking forward to your contribution for step 2.

oke-aditya:
Nice will do this over the weekend :smile:

PR #5440, opened 2022-02-18
Rename ConvNormActivation to Conv2dNormActivation

jdsgomes:
Addresses https://github.com/pytorch/vision/issues/5430

Renames `ConvNormActivation` to `Conv2dNormActivation` and adds deprecation warning in `ConvNormActivation`

cc @datumbox @oke-aditya

### DIFF
--- a/torchvision/ops/misc.py
+++ b/torchvision/ops/misc.py
@@ -1,5 +1,5 @@
 from typing import Callable, List, Optional
-
+import warnings
 import torch
 from torch import Tensor
 
@@ -65,7 +65,7 @@ class FrozenBatchNorm2d(torch.nn.Module):
         return f"{self.__class__.__name__}({self.weight.shape[0]}, eps={self.eps})"
 
 
-class ConvNormActivation(torch.nn.Sequential):
+class Conv2dNormActivation(torch.nn.Sequential):
     """
     Configurable block used for Convolution-Normalzation-Activation blocks.
 
@@ -123,6 +123,15 @@ class ConvNormActivation(torch.nn.Sequential):
         _log_api_usage_once(self)
         self.out_channels = out_channels
 
+
+class ConvNormActivation(Conv2dNormActivation):
+    def __init__(self, *args, **kwargs):
+        warnings.warn(
+            "The ConvNormActivation class are deprecated since 0.13 and will be removed in 0.15. "
+            "Use torchvision.ops.misc.ConvNormActivation instead.",
+            FutureWarning,
+        )
+        super().__init__(*args, **kwargs)
 
 class SqueezeExcitation(torch.nn.Module):
     """
--- a/docs/source/ops.rst
+++ b/docs/source/ops.rst
@@ -45,5 +45,5 @@ Operators
     FeaturePyramidNetwork
     StochasticDepth
     FrozenBatchNorm2d
-    ConvNormActivation
+    Conv2dNormActivation
     SqueezeExcitation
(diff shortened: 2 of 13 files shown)


torchvision/ops/misc.py
```diff
@@ -124,6 +124,15 @@ def __init__(
...
 
+class ConvNormActivation(Conv2dNormActivation):
+    def __init__(self, *args, **kwargs):
+        warnings.warn(
+            "The ConvNormActivation class are deprecated since 0.13 and will be removed in 0.15. "
+            "Use torchvision.ops.misc.ConvNormActivation instead.",
+            FutureWarning,
+        )
+        super().__init__(*args, **kwargs)
+
```
datumbox, line 135: 1 more new line needed.

oke-aditya (commented):
Can we use layers on [line 105 ](https://github.com/pytorch/vision/pull/5440/files#diff-c5d99dd50d7b031f14fb1ff63e467861267f5677ceed3ce45ee5dbd154fecd19R105)as [nn.ModuleList](https://pytorch.org/docs/stable/generated/torch.nn.ModuleList.html) ? I think that's what pytorch recommends for storing list of layers.

oke-aditya:
Also smallest nit. There is a typo in convolution spelling on Line 79. I

datumbox:
@oke-aditya Thanks for the comments, I wish Github could allow comments on unmodified lines. It would assist reviews so much.

> Can we use layers on line 105 as nn.ModuleList ? 

I think it's fine to extend directly from Sequential here. It's an idiom we use often at TorchVision. 

> Also smallest nit. There is a typo in convolution spelling on Line 79.

Great spot! `torchvision/ops/misc.py` line 79 reads `convolutiuon`

> By the way where do we test this function?

We test it as part of multiple models that use it. There is a point to be made that each op could be tested separately, but this is quite common with many layers (for example `BackboneWithFPN`).

jdsgomes:
Thank you @datumbox and @oke-aditya for the comments. I will address those. 

But before that, and taking into account the discussion in https://github.com/pytorch/vision/pull/5445, I am thinking if I should instead reformulate this as making `ConvNormActivation` parametrisable with the type of convolution.

Then both `Conv2dNormActivation` and `Conv3dNormActivation` would inherit from the parent `ConvNormActivation`.

In this version we would have the disadvantage of making `ConvNormActivation` more complicated as discussed earlier but we wouldn't have to deprecate it, and the inheritance would be more natural. Thoughts?

datumbox:
@jdsgomes Sounds good to me. I wouldn't worry too much about trying to detect all the possible incorrect combos that the user might pass to the base class in that case. I think `ConvNormActivation` can be extended in a BC way and then use the 2d and 3d versions as shortcuts.

Let's see how the implementation looks like and chat again.

oke-aditya:
Yeah we can do this. Also just name the base class as `_convrNormActivation`. As I think that is not supposed to be used. And we can then deprecate `ConvNormActivation`.

oke-aditya:
Looks like me and @datumbox just jinxed at same time to comment. :eyes: I can try doing this, but will take a couple of days time.

jdsgomes:
@oke-aditya the timeline works, so fell free to give it a shot.

I will cancel this PR and suggest a new PRs:
- turn  `convNormActivation` into `_convNormActivation` and make it generic, plus create an alias with a deprecation warning.
- create the 2d and 3d versions of _convNormActivation 

Finally, if your priorities change just let me know and I can work on it.

oke-aditya:
I will give this a go in couple of days. :smile:

Issue #5336, opened 2022-02-03
port datasets from the old to the new API


pmeier:
The new dataset API is now stable enough to start porting more datasets from the old API. For the `0.13.0` release planned for 2022H2 we want to achieve at least feature parity for the new API. If you want to help out, please comment on the respective issue so we can assign it to you.

The process of adding a dataset to the new API is described [here](https://github.com/pytorch/vision/blob/main/torchvision/prototype/datasets/_builtin/README.md). In addition, we already ported some datasets that you could use as reference. In any case, if you are blocked by something feel free to send a partial PR and ping me there so I can help.

The following datasets need to be ported:

## Image classification

Image classification datasets are good starting point if you are not familiar with the dataset or the new API since they these datsets tend to be the easiest.

- [x] #5337
- [x] #5338
- [x] #5339 [^1]
- [x] #5340 [^1]
- [x] #5341
- [x] #5342
- [x] #5343
- [x] #5344 [^3]
- [x] #5345
- [x] #5346
- [x] #5347
- [x] #5348
- [x] #5349 [^1]
- [x] #5350
- [x] #5351
- [x] #5352

## Image detection or segmentation

Image detection or segmentation datasets tend to be a little harder since one needs to merge more infomation into one sample compared to classification. My suggestion is to only pick one of these if you are either familiar with the dataset or the new API so you don't have two manage two things at once.

- [ ] #5353 (@vfdev-5 )
- [x] #5354 [^2]
- [x] #5355 [^2] [^3]
- [x] #5356 [^2]
- [x] #5367

## Image pairs

We are still designing how exactly image pair datasets should be implemented. I list them here for completeness, but I suggest not picking up any of them until the design is finished.

- [x] #5357
- [x] #5358

## Video classification

We are still designing how exactly video datasets should be implemented. I list them here for completeness, but I suggest not picking up any of them until the design is finished.

- [x] #5359
- [x] #5360
- [x] #5361

## Optical flow

We are still designing how exactly optical flow datasets should be implemented. I list them here for completeness, but I suggest not picking up any of them until the design is finished.

- [x] #5362 [^3]
- [x] #5363
- [x] #5364
- [x] #5365
- [x] #5366

[^1]: These datasets do not provide public download links for the data so they might be harder to work on.
[^2]: These datasets are implemented as classification datasets in the old API, but provide extra annotations for detection or segmentation.
[...]

PR #5441, opened 2022-02-18
Implementation for cityscapes in proto datasets

### LINKED ISSUE #5353: Cityscapes

vfdev-5:
Fixes #5353

Related to https://github.com/pytorch/vision/issues/5336

Description:
- Implementation for cityscapes in proto datasets

### DIFF
--- a/docs/source/ops.rst
+++ b/docs/source/ops.rst
@@ -20,6 +20,8 @@ Operators
     box_iou
     clip_boxes_to_image
     deform_conv2d
+    drop_block2d
+    drop_block3d
     generalized_box_iou
     generalized_box_iou_loss
     masks_to_boxes
@@ -48,3 +50,5 @@ Operators
     Conv2dNormActivation
     Conv3dNormActivation
     SqueezeExcitation
+    DropBlock2d
+    DropBlock3d
--- a/references/classification/utils.py
+++ b/references/classification/utils.py
@@ -166,17 +166,7 @@ class ExponentialMovingAverage(torch.optim.swa_utils.AveragedModel):
         def ema_avg(avg_model_param, model_param, num_averaged):
             return decay * avg_model_param + (1 - decay) * model_param
 
-        super().__init__(model, device, ema_avg)
-
-    def update_parameters(self, model):
-        for p_swa, p_model in zip(self.module.state_dict().values(), model.state_dict().values()):
-            device = p_swa.device
-            p_model_ = p_model.detach().to(device)
-            if self.n_averaged == 0:
-                p_swa.detach().copy_(p_model_)
-            else:
-                p_swa.detach().copy_(self.avg_fn(p_swa.detach(), p_model_, self.n_averaged.to(device)))
-        self.n_averaged += 1
+        super().__init__(model, device, ema_avg, use_buffers=True)
 
 
 def accuracy(output, target, topk=(1,)):
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -2,6 +2,7 @@ import math
 import os
 from abc import ABC, abstractmethod
 from functools import lru_cache
+from itertools import product
 from typing import Callable, List, Tuple
 
 import numpy as np
@@ -48,6 +49,16 @@ class DeformConvModuleWrapper(nn.Module):
 
 
 class StochasticDepthWrapper(nn.Module):
+    def __init__(self, obj):
+        super().__init__()
+        self.layer = obj
+        self.n_inputs = 1
+
+    def forward(self, a):
+        self.layer(a)
+
+
+class DropBlockWrapper(nn.Module):
     def __init__(self, obj):
         super().__init__()
         self.layer = obj
@@ -1355,6 +1366,88 @@ class TestUtils:
 
         assert len(params[0]) == 92
         assert len(params[1]) == 82
+
+
+class TestDropBlock:
+    @pytest.mark.parametrize("seed", range(10))
+    @pytest.mark.parametrize("dim", [2, 3])
+    @pytest.mark.parametrize("p", [0, 0.5])
+    @pytest.mark.parametrize("block_size", [5, 11])
+    @pytest.mark.parametrize("inplace", [True, False])
+    def test_drop_block(self, seed, dim, p, block_size, inplace):
+        torch.manual_seed(seed)
+        batch_size = 5
+        channels = 3
+        height = 11
+        width = height
+        depth = height
+        if dim == 2:
+            x = torch.ones(size=(batch_size, channels, height, width))
+            layer = ops.DropBlock2d(p=p, block_size=block_size, inplace=inplace)
+            feature_size = height * width
+        elif dim == 3:
+            x = torch.ones(size=(batch_size, channels, depth, height, width))
+            layer = ops.DropBlock3d(p=p, block_size=block_size, inplace=inplace)
+            feature_size = depth * height * width
+        layer.__repr__()
+
+        out = layer(x)
+        if p == 0:
+            assert out.equal(x)
+        if block_size == height:
+            for b, c in product(range(batch_size), range(channels)):
+                assert out[b, c].count_nonzero() in (0, feature_size)
+
+    @pytest.mark.parametrize("seed", range(10))
+    @pytest.mark.parametrize("dim", [2, 3])
+    @pytest.mark.parametrize("p", [0.1, 0.2])
+    @pytest.mark.parametrize("block_size", [3])
+    @pytest.mark.parametrize("inplace", [False])
+    def test_drop_block_random(self, seed, dim, p, block_size, inplace):
+        torch.manual_seed(seed)
+        batch_size = 5
+        channels = 3
+        height = 11
+        width = height
+        depth = height
+        if dim == 2:
+            x = torch.ones(size=(batch_size, channels, height, width))
+            layer = ops.DropBlock2d(p=p, block_size=block_size, inplace=inplace)
+        elif dim == 3:
+            x = torch.ones(size=(batch_size, channels, depth, height, width))
+            layer = ops.DropBlock3d(p=p, block_size=block_size, inplace=inplace)
+
+        trials = 250
+        num_samples = 0
+        counts = 0
+        cell_numel = torch.tensor(x.shape).prod()
+        for _ in range(trials):
+            with torch.no_grad():
+                out = layer(x)
+            non_zero_count = out.nonzero().size(0)
+            counts += cell_numel - non_zero_count
+            num_samples += cell_numel
+
+        assert abs(p - counts / num_samples) / p < 0.15
+
+    def make_obj(self, dim, p, block_size, inplace, wrap=False):
+        if dim == 2:
+            obj = ops.DropBlock2d(p, block_size, inplace)
+        elif dim == 3:
+            obj = ops.DropBlock3d(p, block_size, inplace)
+        return DropBlockWrapper(obj) if wrap else obj
+
+    @pytest.mark.parametrize("dim", (2, 3))
+    @pytest.mark.parametrize("p", [0, 1])
+    @pytest.mark.parametrize("block_size", [5, 7])
+    @pytest.mark.parametrize("inplace", [True, False])
+    def test_is_leaf_node(self, dim, p, block_size, inplace):
+        op_obj = self.make_obj(dim, p, block_size, inplace, wrap=True)
+        graph_node_names = get_graph_node_names(op_obj)
+
+        assert len(graph_node_names) == 2
+        assert len(graph_node_names[0]) == len(graph_node_names[1])
+        assert len(graph_node_names[0]) == 1 + op_obj.n_inputs
 
 
 if __name__ == "__main__":
--- a/test/test_video_gpu_decoder.py
+++ b/test/test_video_gpu_decoder.py
@@ -3,7 +3,7 @@ import os
 
 import pytest
 import torch
-from torchvision.io import _HAS_VIDEO_DECODER, VideoReader
+from torchvision.io import _HAS_GPU_VIDEO_DECODER, VideoReader
 
 try:
     import av
@@ -13,7 +13,7 @@ except ImportError:
 VIDEO_DIR = os.path.join(os.path.dirname(os.path.abspath(__file__)), "assets", "videos")
 
 
-@pytest.mark.skipif(_HAS_VIDEO_DECODER is False, reason="Didn't compile with support for gpu decoder")
+@pytest.mark.skipif(_HAS_GPU_VIDEO_DECODER is False, reason="Didn't compile with support for gpu decoder")
 class TestVideoGPUDecoder:
     @pytest.mark.skipif(av is None, reason="PyAV unavailable")
     @pytest.mark.parametrize(
--- a/torchvision/io/__init__.py
+++ b/torchvision/io/__init__.py
@@ -5,9 +5,9 @@ import torch
 from ..utils import _log_api_usage_once
 
 try:
-    from ._load_gpu_decoder import _HAS_VIDEO_DECODER
+    from ._load_gpu_decoder import _HAS_GPU_VIDEO_DECODER
 except ModuleNotFoundError:
-    _HAS_VIDEO_DECODER = False
+    _HAS_GPU_VIDEO_DECODER = False
 from ._video_opt import (
     Timebase,
     VideoMetaData,
@@ -119,7 +119,7 @@ class VideoReader:
         self.is_cuda = False
         device = torch.device(device)
         if device.type == "cuda":
-            if not _HAS_VIDEO_DECODER:
+            if not _HAS_GPU_VIDEO_DECODER:
                 raise RuntimeError("Not compiled with GPU decoder support.")
             self.is_cuda = True
             if device.index is None:
@@ -218,7 +218,7 @@ __all__ = [
     "_read_video_timestamps_from_memory",
     "_probe_video_from_memory",
     "_HAS_VIDEO_OPT",
-    "_HAS_VIDEO_DECODER",
+    "_HAS_GPU_VIDEO_DECODER",
     "_read_video_clip_from_memory",
     "_read_video_meta_data",
     "VideoMetaData",
--- a/torchvision/io/_load_gpu_decoder.py
+++ b/torchvision/io/_load_gpu_decoder.py
@@ -3,6 +3,6 @@ from ..extension import _load_library
 
 try:
     _load_library("Decoder")
-    _HAS_VIDEO_DECODER = True
+    _HAS_GPU_VIDEO_DECODER = True
 except (ImportError, OSError):
-    _HAS_VIDEO_DECODER = False
+    _HAS_GPU_VIDEO_DECODER = False
--- a/torchvision/ops/__init__.py
+++ b/torchvision/ops/__init__.py
@@ -11,6 +11,7 @@ from .boxes import (
 )
 from .boxes import box_convert
 from .deform_conv import deform_conv2d, DeformConv2d
+from .drop_block import drop_block2d, DropBlock2d, drop_block3d, DropBlock3d
 from .feature_pyramid_network import FeaturePyramidNetwork
 from .focal_loss import sigmoid_focal_loss
 from .giou_loss import generalized_box_iou_loss
@@ -55,4 +56,8 @@ __all__ = [
     "Conv3dNormActivation",
     "SqueezeExcitation",
     "generalized_box_iou_loss",
+    "drop_block2d",
+    "DropBlock2d",
+    "drop_block3d",
+    "DropBlock3d",
 ]
--- a/torchvision/ops/drop_block.py
+++ b/torchvision/ops/drop_block.py
@@ -1,0 +1,155 @@
+import torch
+import torch.fx
+import torch.nn.functional as F
+from torch import nn, Tensor
+
+from ..utils import _log_api_usage_once
+
+
+def drop_block2d(
+    input: Tensor, p: float, block_size: int, inplace: bool = False, eps: float = 1e-06, training: bool = True
+) -> Tensor:
+    """
+    Implements DropBlock2d from `"DropBlock: A regularization method for convolutional networks"
+    <https://arxiv.org/abs/1810.12890>`.
+
+    Args:
+        input (Tensor[N, C, H, W]): The input tensor or 4-dimensions with the first one
+                    being its batch i.e. a batch with ``N`` rows.
+        p (float): Probability of an element to be dropped.
+        block_size (int): Size of the block to drop.
+        inplace (bool): If set to ``True``, will do this operation in-place. Default: ``False``.
+        eps (float): A value added to the denominator for numerical stability. Default: 1e-6.
+        training (bool): apply dropblock if is ``True``. Default: ``True``.
+
+    Returns:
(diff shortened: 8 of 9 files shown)


torchvision/prototype/datasets/_builtin/cityscapes.py
```diff
@@ -0,0 +1,211 @@
...
+    "CityscapesClass",
+    ["name", "id", "train_id", "category", "category_id", "has_instances", "ignore_in_eval", "color"],
+)
+
+
+class Cityscapes(Dataset):
+
+    categories_to_details: FrozenMapping = FrozenMapping(
+        {
+            "unlabeled": CityscapesClass("unlabeled", 0, 255, "void", 0, False, True, (0, 0, 0)),
```
pmeier, line 59: It seems, all of this information is exposed to the user, but never used internally. Is this just static information or would it make sense to include it in the samples?

pmeier (commented):
Thanks @vfdev-5. I have a few questions and mostly simplifications inline.

torchvision/prototype/datasets/_builtin/cityscapes.py
vfdev-5: It is used in `extra=dict(classname_to_details=self.categories_to_details),` and is exposed to the user as in the previously implementation

torchvision/prototype/datasets/_builtin/cityscapes.py
```diff
@@ -0,0 +1,211 @@
...
+        "gtCoarse.zip": "3555e09349ed49127053d940eaa66a87a79a175662b329c1a26a58d47e602b5b",
+        "gtFine_trainvaltest.zip": "40461a50097844f400fef147ecaf58b18fd99e14e4917fb7c3bf9c0d87d95884",
+        "leftImg8bit_trainextra.zip": "e41cc14c0c06aad051d52042465d9b8c22bacf6e4c93bb98de273ed7177b7133",
+        "leftImg8bit_trainvaltest.zip": "3ccff9ac1fa1d80a6a064407e589d747ed0657aac7dc495a4403ae1235a37525",
+    }
+
+    def resources(self, config: DatasetConfig) -> List[OnlineResource]:
+        resources: List[OnlineResource] = []
+        if config.mode == "fine":
+            resources += [
```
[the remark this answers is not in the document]
vfdev-5: I think mypy does not like `[CityscapesResource(...), CityscapesResource(...)]` without explicit hint of `List[OnlineResource]`.

torchvision/prototype/datasets/_builtin/cityscapes.py
```diff
@@ -0,0 +1,211 @@
...
+                    sha256=self._FILES_CHECKSUMS[f"leftImg8bit_{split_label}.zip"],
+                ),
+                CityscapesResource(file_name="gtCoarse.zip", sha256=self._FILES_CHECKSUMS["gtCoarse.zip"]),
+            ]
+        return resources
+
+    def _filter_split_images(self, data: Tuple[str, Any], *, req_split: str) -> bool:
+        path = Path(data[0])
+        split = path.parent.parts[-2]
+        return split == req_split and ".png" == path.suffix
```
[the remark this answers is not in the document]
vfdev-5: Yes, I had that in mind but then thought about performance overhead we bring by that granularity, I'll measure the diff in times later.

torchvision/prototype/datasets/_builtin/cityscapes.py
```diff
@@ -0,0 +1,211 @@
...
+
+    def _filter_group_targets(self, data: Tuple[str, Any]) -> str:
+        path = Path(data[0])
+        split = path.parent.parts[-2]
+        stem = Path(path).stem
+        i = stem.rfind("_gt")
+        stem = stem[:i]
+        return f"{split}_{stem}"
```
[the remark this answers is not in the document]
vfdev-5: Yes, this can work here.

torchvision/prototype/datasets/_builtin/cityscapes.py
```diff
@@ -0,0 +1,211 @@
...
+            stem = Path(path).stem
+            for target_type in ["instance", "label", "polygon", "color"]:
+                if target_type in stem:
+                    if target_type == "polygon":
+                        enc_data = json.loads(data.read())
+                    else:
+                        enc_data = EncodedImage.from_file(data)
+                    output[target_type] = enc_data
+                    output[f"{target_type}_path"] = path
+                    break
```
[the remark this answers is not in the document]
vfdev-5: I left a comment on L158 :)

torchvision/prototype/datasets/_builtin/cityscapes.py
```diff
@@ -0,0 +1,211 @@
...
+        self,
+        resource_dps: List[IterDataPipe],
+        *,
+        config: DatasetConfig,
+    ) -> IterDataPipe[Dict[str, Any]]:
+        archive_images, archive_targets = resource_dps
+
+        images_dp = Filter(archive_images, filter_fn=partial(self._filter_split_images, req_split=config.split))
+
+        targets_dp = Grouper(
```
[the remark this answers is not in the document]
vfdev-5: `_filter_group_targets` provides the split, so thus we can avoid additional filtering. Even with your suggestion as city names are all different between train/test/val we do not need to filter I think

torchvision/prototype/datasets/_builtin/cityscapes.py
```diff
@@ -0,0 +1,211 @@
...
+
+        samples = IterKeyZipper(
+            images_dp,
+            targets_dp,
+            key_fn=img_key_fn,
+            ref_key_fn=target_key_fn,
+            buffer_size=INFINITE_BUFFER_SIZE,
+        )
+        samples = hint_sharding(samples)
+        samples = hint_shuffling(samples)
```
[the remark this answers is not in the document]
vfdev-5: For my understanding, why do you think it would be better there ?

vfdev-5:
@pmeier thanks for the review

torchvision/prototype/datasets/_builtin/cityscapes.py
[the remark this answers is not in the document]
pmeier: `hint_shuffling` uses an infinite buffer and thus fewer input data is better. Plus, AFAIK, we should shard as soon as possible to help the data loader. cc @VitalyFedyunin for details.

torchvision/prototype/datasets/_builtin/cityscapes.py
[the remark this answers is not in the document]
pmeier: In that case either cast or ignore it.

torchvision/prototype/datasets/_builtin/cityscapes.py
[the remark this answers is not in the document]
pmeier: Better yet, can't we also use the suggestion from https://github.com/pytorch/vision/pull/5441#discussion_r811398825 here?

torchvision/prototype/datasets/_builtin/cityscapes.py
```diff
@@ -0,0 +1,211 @@
...
+            buffer_size=INFINITE_BUFFER_SIZE,
+            group_size=4,
+        )
+
+        def img_key_fn(data: Tuple[str, Any]) -> str:
+            stem = Path(data[0]).stem
+            stem = stem[: -len("_leftImg8bit")]
+            return stem
+
+        def target_key_fn(data: Tuple[Any, Any]) -> str:
```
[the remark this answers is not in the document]
pmeier: > Your suggestion to return a str should be `"-".join(path.name.split("_", 3)[:3])`.

Is it necessary to use a string as key or can we get away with just using list if 3 strings? If the key needs to be hashable, I suggest we simply convert it into a tuple.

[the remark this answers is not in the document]
pmeier: > Well, "complicated" is rather subjective opinion.

True. "Complicated for what they are doing" is probably a better fit. From the two functions I couldn't see that they actually do the same thing.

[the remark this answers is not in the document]
pmeier: > it is ~25% faster.

What time constants are we talking here? I'm guessing speed on these operations is hardly a factor in the overall pipeline.

torchvision/prototype/datasets/_builtin/cityscapes.py
pmeier: My point is: what does the user do with this information? I reckon `color` is useful to parse a segmentation mask. The `category` looks like what we called `super_category` in COCO. So maybe we can just provide two dictionaries in the extra information that map the category to its color and super category respectively?

torchvision/prototype/datasets/_builtin/cityscapes.py
```diff
@@ -0,0 +1,201 @@
...
+        if config.mode == "fine":
+            filenames = ("leftImg8bit_trainvaltest.zip", "gtFine_trainvaltest.zip")
+        else:
+            split_label = "trainextra" if config.split == "train_extra" else "trainvaltest"
+            filenames = (f"leftImg8bit_{split_label}.zip", "gtCoarse.zip")
+
+        resources: List[OnlineResource] = [
+            CityscapesResource(file_name=file_name, sha256=self._FILES_CHECKSUMS[file_name]) for file_name in filenames
+        ]
+        return resources
```
pmeier, line 129: Nit: I would prefer to avoid having an extra variable

```suggestion
        return [
            CityscapesResource(file_name=file_name, sha256=self._FILES_CHECKSUMS[file_name]) for file_name in filenames
        ]
```

If `mypy` complains we can cast

```suggestion
        return cast(List[OnlineResource], [
            CityscapesResource(file_name=file_name, sha256=self._FILES_CHECKSUMS[file_name]) for file_name in filenames
        ])
```

No strong opinion though.

torchvision/prototype/datasets/_builtin/cityscapes.py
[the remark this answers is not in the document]
pmeier: I would prefer the granularity given that the performance overhead from this hardly will make any difference. cc @NicolasHug for an opinion.

torchvision/prototype/datasets/_builtin/cityscapes.py
```diff
@@ -0,0 +1,201 @@
...
+        output = dict(image_path=img_path, image=EncodedImage.from_file(img_data))
+        # reorder inside group of targets and setup output dictionary:
+        for path, data in target_data:
+            stem = Path(path).stem
+            for target_type in ["instance", "label", "polygon", "color"]:
+                if target_type in stem:
+                    if target_type == "polygon":
+                        enc_data = json.loads(data.read())
+                    else:
+                        enc_data = EncodedImage.from_file(data)
```
pmeier, line 155: Reminder for self: we need an `EncodedSegmentationMask` feature now that we also have a separate `SegementationMask`.

torchvision/prototype/datasets/_builtin/cityscapes.py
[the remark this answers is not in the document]
pmeier: > city names are all different between train/test/val we do not need to filter I think

True, but that is far from obvious. So we need to have at least a comment explaining that or better yet a `Filter(..., path_comparator("parent.parent.name"), config.split)` to make it explicit.

torchvision/prototype/datasets/_builtin/cityscapes.py
```diff
@@ -0,0 +1,201 @@
...
+            data0 = data[0]
+            if isinstance(data0, tuple):
+                data0 = data0[0]
+            path = Path(data0)
+            # The pathes for images and targets are
+            # - leftImg8bit/val/frankfurt/frankfurt_000001_066574_leftImg8bit.png
+            # - gtFine/val/frankfurt/frankfurt_000001_066574_gtFine_polygons.json
+            # - gtFine/val/frankfurt/frankfurt_000001_066574_gtFine_labelIds.png
+            # we transform them into "frankfurt_000001_066574"
+            return "_".join(path.name.split("_", 3)[:3])
```
pmeier, line 192: Can't we use this also for the `Grouper`? This way we only have a single function that generates keys.

torchvision/prototype/datasets/_builtin/cityscapes.py
```diff
@@ -0,0 +1,201 @@
...
+
+        samples = IterKeyZipper(
+            images_dp,
+            targets_dp,
+            key_fn=key_fn,
+            ref_key_fn=key_fn,
```
pmeier, line 198: If both datapipes use the same function, you only need to provide `key_fn`.

test/builtin_dataset_mocks.py
```diff
@@ -1344,3 +1344,96 @@ def pcam(info, root, config):
...
+            def make_image(name, size=10):
+                create_image_folder(
+                    root=gt_dir / split,
+                    name=city,
+                    file_name_fn=lambda idx: name.format(idx=idx),
+                    size=size,
+                    num_examples=num_samples,
+                )
+
+            make_image(f"{city}_000000_00000" + "{idx}" + f"_gt{mode}_instanceIds.png")
```
pmeier, line 1414: Maybe use a lambda here to avoid piecing the string together?

```suggestion
            make_image(lambda idx: f"{city}_000000_{idx:06d}_gt{mode}_instanceIds.png")
```

test/builtin_dataset_mocks.py
```diff
@@ -1344,3 +1344,96 @@ def pcam(info, root, config):
...
+            "objects": [
+                {
+                    "label": "sky",
+                    "polygon": [
+                        [1241, 0],
+                        [1234, 156],
+                        [1478, 197],
+                        [1611, 172],
+                        [1606, 0],
+                    ],
```
pmeier, line 1385: We are not doing anything with the polygons in the dataset, right? Maybe we can randomize the values here.

test/builtin_dataset_mocks.py
```diff
@@ -1344,3 +1344,96 @@ def pcam(info, root, config):
...
+        if split in ["train", "train_extra"]:
+            cities = ["bochum", "bremen"]
+            num_samples = 3
+        else:
+            cities = ["bochum"]
+            num_samples = 2
+
+        polygon_target = {
+            "imgHeight": 1024,
+            "imgWidth": 2048,
```
pmeier, line 1375: Same here. If they are not aligned with the actual size of the images, we can randomize them.

test/builtin_dataset_mocks.py
```diff
@@ -1344,3 +1344,96 @@ def pcam(info, root, config):
...
+    def generate(cls, root, config):
+
+        mode = config.mode.capitalize()
+        split = config.split
+
+        if split in ["train", "train_extra"]:
+            cities = ["bochum", "bremen"]
+            num_samples = 3
+        else:
+            cities = ["bochum"]
```
pmeier, line 1370: You stated that the names of the cities are different between the splits. This seems to violate this assumption given that `"bochum"` is present in every split.

test/builtin_dataset_mocks.py
```diff
@@ -1344,3 +1344,96 @@ def pcam(info, root, config):
...
+                    ],
+                },
+            ],
+        }
+
+        gt_dir = root / f"gt{mode}"
+
+        for city in cities:
+
+            def make_image(name, size=10):
```
pmeier, line 1405: Nit

```suggestion
            def make_images(name, size=10):
```

test/builtin_dataset_mocks.py
```diff
@@ -1344,3 +1344,96 @@ def pcam(info, root, config):
...
+                    size=size,
+                    num_examples=num_samples,
+                )
+
+            make_image(f"{city}_000000_00000" + "{idx}" + f"_gt{mode}_instanceIds.png")
+            make_image(f"{city}_000000_00000" + "{idx}" + f"_gt{mode}_labelIds.png")
+            make_image(f"{city}_000000_00000" + "{idx}" + f"_gt{mode}_color.png", size=(4, 10, 10))
+
+            for idx in range(num_samples):
+                polygon_target_name = gt_dir / split / city / f"{city}_000000_00000{idx}_gt{mode}_polygons.json"
```
pmeier, line 1419:
```suggestion
                polygon_target_name = gt_dir / split / city / f"{city}_000000_{idx:06d}_gt{mode}_polygons.json"
```

test/builtin_dataset_mocks.py
```diff
@@ -1344,3 +1344,96 @@ def pcam(info, root, config):
...
+                with open(polygon_target_name, "w") as outfile:
+                    json.dump(polygon_target, outfile)
+
+        # Create leftImg8bit folder
+        for city in cities:
+            create_image_folder(
+                root=root / "leftImg8bit" / split,
+                name=city,
+                file_name_fn=lambda idx: f"{city}_000000_00000{idx}_leftImg8bit.png",
```
pmeier, line 1428:
```suggestion
                file_name_fn=lambda idx: f"{city}_000000_{idx:06d}_leftImg8bit.png",
```

test/builtin_dataset_mocks.py
```diff
@@ -1344,3 +1344,96 @@ def pcam(info, root, config):
...
+            make_image(f"{city}_000000_00000" + "{idx}" + f"_gt{mode}_labelIds.png")
+            make_image(f"{city}_000000_00000" + "{idx}" + f"_gt{mode}_color.png", size=(4, 10, 10))
+
+            for idx in range(num_samples):
+                polygon_target_name = gt_dir / split / city / f"{city}_000000_00000{idx}_gt{mode}_polygons.json"
+                with open(polygon_target_name, "w") as outfile:
+                    json.dump(polygon_target, outfile)
+
+        # Create leftImg8bit folder
+        for city in cities:
```
pmeier, line 1424: We should also loop over the splits here precisely to find issues like not filtering on the split in the dataset implementation.

torchvision/prototype/datasets/_builtin/cityscapes.py
vfdev-5: CityscapesClass provides more info that just color, category name, super category. It may happen that users could use other fields for visualization purposes and thus new dataset would miss that info.

torchvision/prototype/datasets/_builtin/cityscapes.py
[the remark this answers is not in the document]
vfdev-5: To make it transparent I can leave a comment in the code. IMO, unfortunately your suggestion `Filter(..., path_comparator("parent.parent.name"), config.split)` does not make it more obvious while adding another step in the code... We can also leave a comment to explain why this filter is added...
I think this additional step is unnecessary

torchvision/prototype/datasets/_builtin/cityscapes.py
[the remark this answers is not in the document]
NicolasHug: I agree there should be no perf difference.

But is the granularity needed here? Unless the 2 new proposed filters are being used somewhere else in the code, there's probably no need to split this 4-liner into 2 separate functions.

test/builtin_dataset_mocks.py
vfdev-5: Actually, I'm a bit confused about what `generate` function generates as image given that config is provided. Should it everytime generate complete dataset regardless the config or it should generate only a split ? 
From testing perspective I'd say it should generate once everything so in this case why do we have `config` as argument ?

test/builtin_dataset_mocks.py
vfdev-5: I'm not quite sure about the purpose of randomizing all this data. I think it can remain static.

test/builtin_dataset_mocks.py
pmeier: `generate` is just a helper construct and is not needed at all. The important part is

```py
@register_mock
def cityscapes(info, root, config):
```

This function should return all files that are needed for given `config`. In particular, every file you generate should be complete, i.e. contain everything the normal file also contains, but of course with less data. So if the regular archive has `train` / `val` / `test` folders, the generated data should have them as well, but with only a few dummy images.

> From testing perspective I'd say it should generate once everything, but in this case why do we have config as argument ?

@NicolasHug and I discussed this at length and we ended up removing the behavior for now in #5271. If you generate everything only once, you need to somehow change the return value of  the function (usually only the number of samples, but might also be more complex) for all possible configs. That made the logic more complex with (for now) little benefit. If runtime of the test ever becomes and issue, we can revisit.

test/builtin_dataset_mocks.py
pmeier: I'm also ok with that. But in that case, please put a comment there to indicate that these are not some special values. Because that was my first thought when seeing this.

torchvision/prototype/datasets/_builtin/cityscapes.py
pmeier: In the stable API this is exposed through the *undocumented* `torchvision.datasets.Cityscapes(...).classes` attribute. Given that the new API is BC breaking anyway, IMO we should aim to reduce the baggage from the old implementation. I indeed see use cases for returning a color to category mapping (maybe we should have this on the `SegmentationMask` feature) as well as the category to super category mapping. For the rest I would remove them for now and only put them back if someone requests it. cc @NicolasHug for an opinion.

torchvision/prototype/datasets/_builtin/cityscapes.py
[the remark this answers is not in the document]
pmeier: If we go for granularity, we don't need any extra function. The current

```py
dp = Filter(dp, functools.partial(self._filter_split_images, req_split=config.split))
```

is equivalent to

```py
dp = Filter(dp, path_comparator("parent.parent.name", config.split))
dp = Filter(dp, path_comparator("suffix", ".png"))
```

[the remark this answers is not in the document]
pmeier: Looking through the archive again, I think we don't even need the second `Filter` here. The only non-png files are located in the parent folder of the split folders. Thus, if we drop every file that is not contained in the split folder we are after, we should only get png images.

torchvision/prototype/datasets/_builtin/cityscapes.py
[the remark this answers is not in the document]
pmeier: Let's keep the discussion in https://github.com/pytorch/vision/pull/5441#discussion_r811393342.

torchvision/prototype/datasets/_builtin/cityscapes.py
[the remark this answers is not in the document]
vfdev-5: Yes, that's work also without filtering on file extension. Archive contains files like : 
- `/cityscapes/leftImg8bit_trainvaltest.zip/README`
- `/cityscapes/leftImg8bit_trainvaltest.zip/license.txt`
for which computed split is `cityscapes` and thus filtered as does not match `train`, `val` etc
Such implicit filtering works but less evident and either we have to leave a comment about that or keep filtering by the extension.

torchvision/prototype/datasets/_builtin/cityscapes.py
vfdev-5: It is dataset meta-data, probably coming from https://github.com/mcordts/cityscapesScripts and is used for visualization (this can be seen by searching through the GitHub and exploring repository by repository, https://github.com/search?p=1&q=CityscapesClass&type=Code). Even if this wasn't explicitly documented previously, I do not think that we have to remove that in the new API.

PR #5442, opened 2022-02-18
[WIP] Stanford Cars prototype dataset

abhi-glitchhg:
Currently a draft.

### DIFF
--- a/torchvision/prototype/datasets/_builtin/__init__.py
+++ b/torchvision/prototype/datasets/_builtin/__init__.py
@@ -1,3 +1,4 @@
+from torch import imag
 from .caltech import Caltech101, Caltech256
 from .celeba import CelebA
 from .cifar import Cifar10, Cifar100
@@ -15,3 +16,4 @@ from .sbd import SBD
 from .semeion import SEMEION
 from .svhn import SVHN
 from .voc import VOC
+from .stanford_cars import StanfordCars
--- a/torchvision/prototype/datasets/_builtin/stanford_cars.py
+++ b/torchvision/prototype/datasets/_builtin/stanford_cars.py
@@ -1,0 +1,111 @@
+
+from cProfile import label
+from hashlib import sha256
+import io
+from collections import namedtuple
+from pickletools import read_string1
+from typing import Any, Callable, Dict, List, Optional, Tuple, Iterator
+import pathlib
+from torchdata.datapipes.iter import (
+    IterDataPipe,
+    Mapper,
+    Filter,
+    )
+
+from torchvision.prototype.datasets.utils._internal import read_mat,hint_sharding, hint_shuffling, path_comparator
+import torch
+from torchdata.datapipes.iter import IterDataPipe, Mapper, Zipper
+from torchvision.prototype import features
+from torchvision.prototype.features import Label, BoundingBox, EncodedImage
+from torchvision.prototype.datasets.utils import (
+    Dataset,
+    DatasetConfig,
+    DatasetInfo,
+    OnlineResource,
+    DatasetType,
+    HttpResource
+)
+
+
+from torchvision.prototype.datasets.utils._internal import (
+    hint_sharding,
+    hint_shuffling,
+)
+from torchvision.prototype.features import Label
+
+
+class StanfordCars(Dataset):
+    def _make_info(self) -> DatasetInfo:
+        return DatasetInfo(
+            name="stanford_cars",
+            type=DatasetType.IMAGE,
+            homepage="https://ai.stanford.edu/~jkrause/cars/car_dataset.html",
+            dependencies=("scipy",),
+            valid_options=dict(split=("test","train"),
+                               ),
+        )
+
+    _URL_ROOT = "https://ai.stanford.edu/~jkrause/"
+    _URLS = {
+        "train":f"{_URL_ROOT}car196/cars_train.tgz",
+        "test":f"{_URL_ROOT}car196/cars_test.tgz",
+        "test_ground_truth":f"{_URL_ROOT}car196/cars_test_annos_withlabels.mat",
+        "devkit":f"{_URL_ROOT}cars/car_devkit.tgz"
+    }
+
+    _CHECKSUM = {
+        "train": "512b227b30e2f0a8aab9e09485786ab4479582073a144998da74d64b801fd288",
+        "test": "bffea656d6f425cba3c91c6d83336e4c5f86c6cffd8975b0f375d3a10da8e243",
+        "test_ground_truth": "790f75be8ea34eeded134cc559332baf23e30e91367e9ddca97d26ed9b895f05",
+        "devkit": "512b227b30e2f0a8aab9e09485786ab4479582073a144998da74d64b801fd288",
+    }
+
+    def resources(self, config: DatasetConfig) -> List[OnlineResource]:
+        resources = [HttpResource(self._URLS[config.split] , sha256=self._CHECKSUM[config.split])]
+        print(len(resources))
+        if config.split == "test":
+            resources.append(HttpResource(self._URLS["test_ground_truth"],sha256=self._CHECKSUM["test_ground_truth"]))
+        else:
+            resources.append(HttpResource(url=self._URLS["devkit"],sha256=self._CHECKSUM["devkit"]))
+        return resources
+
+    def _make_datapipe(
+        self,
+        resource_dps: List[IterDataPipe],
+        *,
+        config: DatasetConfig,
+        decoder: Optional[Callable[[io.IOBase], torch.Tensor]],
+    ) -> IterDataPipe[Dict[str, Any]] :
+
+        images_dp, targets_dp = resource_dps
+
+        if config.split == "train":
+            targets_dp = Filter(targets_dp, path_comparator("name", "cars_train_annos.mat"))
+
+
+        #right now our both dps are of .tgz and .mat format; for all configs(train and test.)
+        targets_dp = Mapper(targets_dp,self._read_labels)
+        dp =Zipper(images_dp,targets_dp)
+        dp=hint_sharding(dp)
+        dp=hint_shuffling(dp)
+
+
+        return Mapper(dp,self._collate_and_decode) 
+
+    def _collate_and_decode(self, data: Tuple[Any, Any]) -> Dict[str, Any]:
+        image, target = data  # They're both numpy arrays at this point
+
+        image_path, image_buffer = image
+
+        image = EncodedImage.from_file(image_buffer)
+
+        label = Label(target.item())
+        return {"image": image, "label":label}
+
+    def _read_labels(self,data):
+        labels = data[0]
+        labels = read_mat(labels,squeeze_me=True)
+        labels=labels["annotations"]
+        bboxs_and_class = zip(labels["bbox_x1"],labels["bbox_y1"],labels["bbox_x2"],labels["bbox_y2"],labels["class"])
+
+        return  bboxs_and_class
--- a/torchvision/prototype/datasets/_builtin/stanford_cars.categories
+++ b/torchvision/prototype/datasets/_builtin/stanford_cars.categories
@@ -1,0 +1,196 @@
+AM General Hummer SUV 2000
+Acura RL Sedan 2012
+Acura TL Sedan 2012
+Acura TL Type-S 2008
+Acura TSX Sedan 2012
+Acura Integra Type R 2001
+Acura ZDX Hatchback 2012
+Aston Martin V8 Vantage Convertible 2012
+Aston Martin V8 Vantage Coupe 2012
+Aston Martin Virage Convertible 2012
+Aston Martin Virage Coupe 2012
+Audi RS 4 Convertible 2008
+Audi A5 Coupe 2012
+Audi TTS Coupe 2012
+Audi R8 Coupe 2012
+Audi V8 Sedan 1994
+Audi 100 Sedan 1994
+Audi 100 Wagon 1994
+Audi TT Hatchback 2011
+Audi S6 Sedan 2011
+Audi S5 Convertible 2012
+Audi S5 Coupe 2012
+Audi S4 Sedan 2012
+Audi S4 Sedan 2007
+Audi TT RS Coupe 2012
+BMW ActiveHybrid 5 Sedan 2012
+BMW 1 Series Convertible 2012
+BMW 1 Series Coupe 2012
+BMW 3 Series Sedan 2012
+BMW 3 Series Wagon 2012
+BMW 6 Series Convertible 2007
+BMW X5 SUV 2007
+BMW X6 SUV 2012
+BMW M3 Coupe 2012
+BMW M5 Sedan 2010
+BMW M6 Convertible 2010
+BMW X3 SUV 2012
+BMW Z4 Convertible 2012
+Bentley Continental Supersports Conv. Convertible 2012
+Bentley Arnage Sedan 2009
+Bentley Mulsanne Sedan 2011
+Bentley Continental GT Coupe 2012
+Bentley Continental GT Coupe 2007
+Bentley Continental Flying Spur Sedan 2007
+Bugatti Veyron 16.4 Convertible 2009
+Bugatti Veyron 16.4 Coupe 2009
+Buick Regal GS 2012
+Buick Rainier SUV 2007
+Buick Verano Sedan 2012
+Buick Enclave SUV 2012
+Cadillac CTS-V Sedan 2012
+Cadillac SRX SUV 2012
+Cadillac Escalade EXT Crew Cab 2007
+Chevrolet Silverado 1500 Hybrid Crew Cab 2012
+Chevrolet Corvette Convertible 2012
+Chevrolet Corvette ZR1 2012
+Chevrolet Corvette Ron Fellows Edition Z06 2007
+Chevrolet Traverse SUV 2012
+Chevrolet Camaro Convertible 2012
+Chevrolet HHR SS 2010
+Chevrolet Impala Sedan 2007
+Chevrolet Tahoe Hybrid SUV 2012
+Chevrolet Sonic Sedan 2012
+Chevrolet Express Cargo Van 2007
+Chevrolet Avalanche Crew Cab 2012
+Chevrolet Cobalt SS 2010
+Chevrolet Malibu Hybrid Sedan 2010
+Chevrolet TrailBlazer SS 2009
+Chevrolet Silverado 2500HD Regular Cab 2012
+Chevrolet Silverado 1500 Classic Extended Cab 2007
+Chevrolet Express Van 2007
+Chevrolet Monte Carlo Coupe 2007
+Chevrolet Malibu Sedan 2007
+Chevrolet Silverado 1500 Extended Cab 2012
+Chevrolet Silverado 1500 Regular Cab 2012
+Chrysler Aspen SUV 2009
+Chrysler Sebring Convertible 2010
+Chrysler Town and Country Minivan 2012
+Chrysler 300 SRT-8 2010
+Chrysler Crossfire Convertible 2008
+Chrysler PT Cruiser Convertible 2008
+Daewoo Nubira Wagon 2002
+Dodge Caliber Wagon 2012
+Dodge Caliber Wagon 2007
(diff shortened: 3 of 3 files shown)


abhi-glitchhg:
I have got a problem and I am not able to fix it. 
the `_make_datapipe` method recieves a list of `IterPipe` objects. There will be two objects becuase `resources` method return a list with two `HttpResource` objects. 

So as per my understanding first object will related to images and second object will be related to labels and bounding boxes. 
And this is correct ; 

`
images dp:  <torchdata.datapipes.iter.util.tararchivereader.TarArchiveReaderIterDataPipe object at 0x7f43f5ed4910> 
 targets dp:  <torch.utils.data.datapipes.iter.fileopener.FileOpenerIterDataPipe object at 0x7f43f5ed4340>`

but sometimes both the dps are related to the labels and bounding boxes;

`images dp:  <torch.utils.data.datapipes.iter.fileopener.FileOpenerIterDataPipe object at 0x7fd7ef58edc0> 
targets dp:  <torch.utils.data.datapipes.iter.fileopener.FileOpenerIterDataPipe object at 0x7fd7ef5a7160>`


I do not understand why this is happening?
Can anyone help me out?
Thanks

cc @pmeier

torchvision/prototype/datasets/_builtin/__init__.py
```diff
@@ -8,10 +8,11 @@
 from .fer2013 import FER2013
 from .gtsrb import GTSRB
 from .imagenet import ImageNet
-from .mnist import MNIST, FashionMNIST, KMNIST, EMNIST, QMNIST
+from .mnist import EMNIST, FashionMNIST, KMNIST, MNIST, QMNIST
```
pmeier, line 11: Could you revert this? I'm guessing you are using `usort>=1`. We are currently bound to `usort==0.6.4` until #5106 is merged.

torchvision/prototype/datasets/_builtin/stanford_cars.py
```diff
@@ -0,0 +1,85 @@
...
+from torchdata.datapipes.iter import Filter, IterDataPipe, Mapper, Zipper
+from torchvision.prototype.datasets.utils import Dataset, DatasetConfig, DatasetInfo, HttpResource, OnlineResource
+from torchvision.prototype.datasets.utils._internal import hint_sharding, hint_shuffling, path_comparator, read_mat
+from torchvision.prototype.features import BoundingBox, EncodedImage
+
+
+class StanfordCars(Dataset):
+    def _make_info(self) -> DatasetInfo:
+        return DatasetInfo(
+            name="stanford_cars",
```
pmeier, line 12: For now the convention is to use a hyphen to separate words in the dataset name.
```suggestion
            name="stanford-cars",
```

Note that you will also have to rename the `.categories` file and should rename the data folder `~/.cache/torch/datasets/vision/stanford_cars` to avoid re-downloading everything.

torchvision/prototype/datasets/_builtin/stanford_cars.py
```diff
@@ -0,0 +1,85 @@
...
+        "test_ground_truth": f"{_URL_ROOT}car196/cars_test_annos_withlabels.mat",
+        "devkit": f"{_URL_ROOT}cars/car_devkit.tgz",
+    }
+
+    _CHECKSUM = {
+        "train": "512b227b30e2f0a8aab9e09485786ab4479582073a144998da74d64b801fd288",
+        "test": "bffea656d6f425cba3c91c6d83336e4c5f86c6cffd8975b0f375d3a10da8e243",
+        "test_ground_truth": "790f75be8ea34eeded134cc559332baf23e30e91367e9ddca97d26ed9b895f05",
+        "devkit": "512b227b30e2f0a8aab9e09485786ab4479582073a144998da74d64b801fd288",
+    }
```
pmeier, line 33: I would prefer to have a mapping of the actual file names to the checksums.

torchvision/prototype/datasets/_builtin/stanford_cars.py
```diff
@@ -0,0 +1,85 @@
...
+        resource_dps: List[IterDataPipe],
+        *,
+        config: DatasetConfig,
+    ) -> IterDataPipe[Dict[str, Any]]:
+
+        images_dp, targets_dp = resource_dps
+        print(config.split)
+        if config.split == "train":
+            targets_dp = Filter(targets_dp, path_comparator("name", "cars_train_annos.mat"))
+        dp = Zipper(images_dp, targets_dp)
```
pmeier, line 54: This is dangerous. Did you make sure that both datapipes are perfectly aligned? Otherwise we should use an `IterKeyZipper`.

torchvision/prototype/datasets/_builtin/stanford_cars.py
```diff
@@ -0,0 +1,85 @@
...
+        dp = hint_sharding(dp)
+        dp = hint_shuffling(dp)
+        return Mapper(dp, self._read_images_and_labels)
+
+    def _read_images_and_labels(self, data):
```
pmeier, line 59: Nit: this method is canonically called `_prepare_sample` and is placed before the `_make_datapipe` method.

pmeier (commented):
Thanks @abhi-glitchhg for reporting this. I've send a patch in #5447 that should fix this behavior. After that is merged, you can update your branch and the issue should be gone.

I had an superficial look and so far this looks really good! I'll do another round after you mark the PR as ready for review.

torchvision/prototype/datasets/_builtin/stanford_cars.py
abhi-glitchhg: I wasn't aware of this! I will look into it. Thanks!

abhi-glitchhg:
Sorry for the delay! I will try to finish this by the weekend!

torchvision/prototype/datasets/_builtin/stanford_cars.py
abhi-glitchhg: I think that zipping these two pipelines is not a good idea; Because `targets_dp` contains only one file and images_dp contains around 8000 images;  So with the current implementation, we can access only one image.. 

I was thinking about decoding the `targets_dp` and saving the target_dp data into a variable (something like `self.labels` ) and then accessing it whenever needed...

Could you give me some feedback regarding this? Like should i go with this? Or is there any other way to tackle this..

abhi-glitchhg: They are aligned :)

torchvision/prototype/datasets/_builtin/stanford_cars.py
```diff
@@ -0,0 +1,97 @@
+from typing import Any, Dict, List, Tuple, Iterator
+
+import numpy as np
+from torchdata.datapipes.iter import Filter, IterDataPipe, Mapper, Zipper
+from torchvision.prototype.datasets.utils import Dataset, DatasetConfig, DatasetInfo, HttpResource
+from torchvision.prototype.datasets.utils._internal import hint_sharding, hint_shuffling, path_comparator, read_mat
+from torchvision.prototype.features import BoundingBox, EncodedImage
+
+
+class _StanfordCarsLabelReader(IterDataPipe[Tuple[np.ndarray, int]]):
```
pmeier, line 10: 1. You can make this "public" since this part of `torchvision` is private as a whole and we only explicitly expose stuff.
2. The generic type annotation has to match the one from `__iter__`. See the comment below for an explanation
3. 
```suggestion
class StanfordCarsLabelReader(IterDataPipe[Tuple[int, int, int, int, int, str]]):
```

torchvision/prototype/datasets/_builtin/stanford_cars.py
```diff
@@ -0,0 +1,97 @@
...
+    def __init__(self, datapipe: IterDataPipe[Dict[str, Any]]) -> None:
+        self.datapipe = datapipe
+
+    def __iter__(self) -> Iterator[Tuple[str]]:
+        for _, file in self.datapipe:
+            file = iter(read_mat(file, squeeze_me=True)["annotations"])
+            for line in file:
+                yield line
```
pmeier, line 18: This was hard to parse for me until I run this code myself. Maybe something like this?

```suggestion
    def __iter__(self) -> Iterator[Tuple[int, int, int, int, int, str]]:
        for _, file in self.datapipe:
            data = read_mat(file, squeeze_me=True)
            for ann in data["annotations"]:
                yield tuple(ann)
```

torchvision/prototype/datasets/_builtin/stanford_cars.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def _make_datapipe(
+        self,
+        resource_dps: List[IterDataPipe],
+        *,
+        config: DatasetConfig,
+    ) -> IterDataPipe[Dict[str, Any]]:
+
+        images_dp, targets_dp = resource_dps
+        if config.split == "train":
```
pmeier, line 91: Can we align this with the check in `self.resources()`? There you check for `config.split == "test"`. No opinion on what check we should use.

torchvision/prototype/datasets/_builtin/stanford_cars.py
```diff
@@ -0,0 +1,97 @@
...
+
+    _CHECKSUM = {
+        "train": "b97deb463af7d58b6bfaa18b2a4de9829f0f79e8ce663dfa9261bf7810e9accd",
+        "test": "bffea656d6f425cba3c91c6d83336e4c5f86c6cffd8975b0f375d3a10da8e243",
+        "cars_test_annos_withlabels": "790f75be8ea34eeded134cc559332baf23e30e91367e9ddca97d26ed9b895f05",
+        "car_devkit": "512b227b30e2f0a8aab9e09485786ab4479582073a144998da74d64b801fd288",
+    }
+
+    def resources(self, config: DatasetConfig) -> List[HttpResource]:
+        resources = [HttpResource(self._URLS[config.split], sha256=self._CHECKSUM[config.split])]
```
pmeier, line 48:
```suggestion
    def resources(self, config: DatasetConfig) -> List[OnlineResource]:
        resources: List[OnlineResource] = [HttpResource(self._URLS[config.split], sha256=self._CHECKSUM[config.split])]
```

torchvision/prototype/datasets/_builtin/stanford_cars.py
```diff
@@ -0,0 +1,97 @@
...
+                [
+                    target[0],
+                    target[1],
+                    target[2],
+                    target[3],
+                ],
+                format="xyxy",
+                image_size=image.image_size,
+            ),
+        )
```
pmeier, line 81: In general, this method does what it should do. I've cleaned it up a bit:

1. Not sure why you want to return the index of the image? Is there a concrete use case where this is necessary? If not, the information is available for the user to extract from the path and thus there is no need to return it explicitly.
2. Since the `image_path` is the only path, there is no need to namespace it.
3. We have a special feature type for labels: `Label`. Among the numeric label, it can also store the human readable categories. Plus, canonically, we store the label under the `"label"` key.
4. If the coordinates of the bounding box are already in the right order, you can simply use slice syntax.

```suggestion
    def _prepare_sample(self, data: Tuple[Tuple[str, BinaryIO], Tuple[int, int, int, int, int, str]]) -> Dict[str, Any]:
        image, target = data
        path, buffer = image
        image = EncodedImage.from_file(buffer)

        return dict(
            path=path,
            image=image,
            label=Label(target[4] - 1, categories=self.categories),
            bounding_box=BoundingBox(target[:4], format="xyxy", image_size=image.image_size),
        )
```

torchvision/prototype/datasets/_builtin/stanford-cars.categories
```diff
@@ -0,0 +1,196 @@
+AM General Hummer SUV 2000
```
pmeier, line 1: How did you generate this file? Unless you need to compile it manually, the best way is to put a `_generate_categories` method on the dataset and use the `torchvision/prototype/datasets/generate_categories_file.py` script. This way we have a record of the code that was used to generate the file.

test/builtin_dataset_mocks.py
```diff
@@ -1390,3 +1390,45 @@ def pcam(info, root, config):
...
+
+
+@register_mock
+def stanford_cars(info, root, config):
+    import scipy.io as io
+    from numpy.core.records import fromarrays
+
+    num_examples = {"train": 5, "test": 7}[config["split"]]
+    num_classes = 3
+    base_folder = pathlib.Path(root)
```
pmeier, line 1402: `root` is already a `pathlib.Path` so you can remove `base_folder` from this function and use `root` instead.

pmeier (commented):
Thanks a lot @abhi-glitchhg! I have some minor comments inline, but otherwise this looks really good!

torchvision/prototype/datasets/_builtin/stanford_cars.py
abhi-glitchhg: I was returning the index of the image for testing purposes; I was checking whether the images_dp and target_dp are aligned or not. 
Agree with all the changes that you have suggested!

test/builtin_dataset_mocks.py
abhi-glitchhg: Okay! thanks

torchvision/prototype/datasets/_builtin/stanford_cars.py
abhi-glitchhg: Oh sorry! My bad! I should have made it more readable.

torchvision/prototype/datasets/_builtin/stanford-cars.categories
abhi-glitchhg: I did not add the `generate_categories` method in the dataset because we will have to download `car_devkit.tgz` irrespective of the split(train/test.) 
That's why I did not include it. 
But if you insist, I will :)

pmeier: The method is not executed at runtime. So you only need this file locally if you want to generate the file. And this is a compromise I'm willing to make to have a reproducible generation. Make sure you use something like

https://github.com/pytorch/vision/blob/50608fbc263da386ad7abf7b32bd32ed7f691170/torchvision/prototype/datasets/_builtin/imagenet.py#L193-L195

to actually get the resources that include the file you need.

torchvision/prototype/datasets/_builtin/stanford_cars.py
```diff
@@ -0,0 +1,85 @@
...
+
+class StanfordCarsLabelReader(IterDataPipe[Tuple[int, int, int, int, int, str]]):
+    def __init__(self, datapipe: IterDataPipe[Dict[str, Any]]) -> None:
+        self.datapipe = datapipe
+
+    def __iter__(self) -> Iterator[Tuple[int, int, int, int, int, str]]:
+        for _, file in self.datapipe:
+            data = read_mat(file, squeeze_me=True)
+            for ann in data["annotations"]:
+                yield tuple(ann)
```
pmeier, line 17: We can just ignore `mypy` here

```suggestion
                yield tuple(ann)  # type: ignore[misc]
```

torchvision/prototype/datasets/_builtin/stanford_cars.py
```diff
@@ -0,0 +1,100 @@
...
+        return Mapper(dp, self._prepare_sample)
+
+    def _generate_categories(self, root: pathlib.Path) -> List[str]:
+        config = self.info.make_config(split="train")
+        resources = self.resources(config)
+
+        devkit_dp = resources[1].load(root)
+        meta_dp = Filter(devkit_dp, path_comparator("name", "cars_meta.mat"))
+        print(meta_dp)
+        return Mapper(meta_dp, self.prepare_categories)
```
pmeier, line 100: In this method you don't need to mess around with datapipes unless it is helpful. In this case it makes sense to stay in "datapipe land" until after the filter. Starting from that point you only have a single file that you need to parse.

```suggestion
        _, meta_file = next(iter(meta_dp))

        return list(read_mat(meta_file, squeeze_me=True)["class_names"])
```

> Here i am returning a list of strings which contains the class names

You returned a datapipe that contained a list of strings. This is treated the same way as a list of lists and this is why the generation put everything on one line.

torchvision/prototype/datasets/_builtin/stanford_cars.py
```diff
@@ -0,0 +1,100 @@
...
+            )
+        else:
+            resources.append(HttpResource(url=self._URLS["car_devkit"], sha256=self._CHECKSUM["car_devkit"]))
+
+        return resources
+
+    def prepare_categories(self, dp: IterDataPipe) -> List[str]:
+        _, file = dp
+        classes = list(read_mat(file, squeeze_me=True)["class_names"])
+        return classes
```
pmeier, line 63: This is no longer needed.

```suggestion
```

torchvision/prototype/datasets/_builtin/stanford_cars.py
```diff
@@ -0,0 +1,102 @@
...
```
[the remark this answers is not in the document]
pmeier: See https://github.com/pytorch/vision/pull/5442#discussion_r821839019.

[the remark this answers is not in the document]
abhi-glitchhg: Ohhhhh understood.

forgive me for those mistakes😞

test/builtin_dataset_mocks.py
```diff
@@ -1390,3 +1390,44 @@ def pcam(info, root, config):
...
 
     return num_images
+
+
+@register_mock
+def stanford_cars(info, root, config):
+    import scipy.io as io
+    from numpy.core.records import fromarrays
+
+    num_examples = {"train": 5, "test": 7}[config["split"]]
```
pmeier, line 1400: Can we call this 

```suggestion
    num_samples = {"train": 5, "test": 7}[config["split"]]
```

to align it with the terminology of the other datasets?

test/builtin_dataset_mocks.py
```diff
@@ -1390,3 +1390,44 @@ def pcam(info, root, config):
...
+    num_classes = 3
```
pmeier, line 1401: Can we call this

```suggestion
    num_categories = 3
```

to align it with the terminology of the other datasets?

test/builtin_dataset_mocks.py
```diff
@@ -1390,3 +1390,44 @@ def pcam(info, root, config):
...
+    classes = np.random.randint(1, num_classes + 1, num_examples, dtype=np.uint8)
+    fnames = [f"{i:5d}.jpg" for i in range(num_examples)]
+    rec_array = fromarrays(
+        [bbox, bbox, bbox, bbox, classes, fnames],
+        names=["bbox_x1", "bbox_y1", "bbox_x2", "bbox_y2", "class", "fname"],
+    )
+
+    io.savemat(annotations_mat_path, {"annotations": rec_array})
+    if config.split == "train":
+        make_tar(root, "car_devkit.tgz", devkit)
```
pmeier, line 1431:
```suggestion
        make_tar(root, "car_devkit.tgz", devkit, compression="gz")
```

pmeier (commented):
Thanks @abhi-glitchhg, we are almost done! Just a few minor comments including one from a previous review left. After that we are good to merge!

test/builtin_dataset_mocks.py
```diff
@@ -370,11 +370,11 @@ def create_ann_file(root, name):
 
         scipy.io.savemat(str(pathlib.Path(root) / name), dict(box_coord=box_coord, obj_contour=obj_contour))
 
-    def create_ann_folder(root, name, file_name_fn, num_examples):
```
abhi-glitchhg, line 373: Ooops!

pmeier (approved):
Awesome, thanks a lot for the patience @abhi-glitchhg!

abhi-glitchhg:
@pmeier  thanks for the guidance! <3

pmeier:
Hey @abhi-glitchhg, we appreciate your responsiveness, but you don't need to update the branch for every new commit on the `main` branch. We will do that before we merge. Otherwise, we would be wasting a lot of CI resources for very little gain. 

One exception is if GitHub auto-detects some conflicts. In that case, feel free to address them before the final approval.

NicolasHug (approved):
Stamping, thanks @abhi-glitchhg

PR #5443, opened 2022-02-18
Use frozen BN only if pre-trained backbone

datumbox:
Currently, the majority of our Detection models replace the `BatchNorm2d` layers with `FrozenBatchNorm2d`. This is a reasonable mitigation that improves the stability of training for small batch-sizes. Unfortunately, our current implementation freezes the BNs even when they are completely randomly initialized. Since `FrozenBatchNorm2d` freezes both the running stats and the affine parameters, its parameters get initialized and fixed to values to:
https://github.com/pytorch/vision/blob/0c2373d0bba3499e95776e7936e207d8a1676e65/torchvision/ops/misc.py#L30-L33

Consequently, the BN layers are effectively completely disabled for those who try to train the models from scratch. 

This PR fixes the issue by replacing the BNs with FrozenBNs when at least some pre-trained weights are loaded.


torchvision/models/detection/faster_rcnn.py
```diff
@@ -383,15 +383,15 @@ def fasterrcnn_resnet50_fpn(
             Valid values are between 0 and 5, with 5 meaning all backbone layers are trainable. If ``None`` is
             passed (the default) this value is set to 3.
     """
-    trainable_backbone_layers = _validate_trainable_layers(
-        pretrained or pretrained_backbone, trainable_backbone_layers, 5, 3
-    )
+    is_trained = pretrained or pretrained_backbone
```
fmassa, line 386: nit: if the model was trained for detection with large batch sizes from scratch, and then we finetune it afterwards (still with large batch sizes) then in this case we would be using `FrozenBatchNorm`.

This is an ok heuristic, but hints that we might want to make this an explicit parameter from the constructor in the future

fmassa (approved):
Thanks!

Issue #5325, opened 2022-02-01
Pass losses as callables when building detection models


Quintulius:
### 🚀 The feature

Some models currently accept normalization strategies as callables (```mobilenet_backbone``` accepts a ```norm_layer``` argument for example) but loss functions are currently hardcoded (```F.cross_entropy``` for ```fastercnn.roi_heads``` for example).

Following what has been done on normalization strategies loss function could be passed as callables in the modules constructor. This shouldn't break any backward compatibility. Reduction strategies still need to be properly handled.

### Motivation, pitch

Currently, trying different loss functions requires to use some dirty model patches. Accepting the losses in the model constructors would provide a much cleaner way to hack around the models.

If any interest I can propose a first PR modifying the Faster-RNN models.

Quintulius:
OK ! Modifying the detection models without breaking too much the backward compatibility seems indeed quite a challenge. I'm looking forward to seeing how these models evolve and become more flexible !

Issue #5307, opened 2022-01-28
Improve the accuracy of Detection models by using SOTA recipes and primitives


datumbox:
### 🚀 The feature

Similar to #3995 but focus on Object Detection.

Here is the accuracy of `fasterrcnn_resnet50_fpn` on latest main:
```
0:  Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.369
0:  Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=100 ] = 0.585
0:  Average Precision  (AP) @[ IoU=0.75      | area=   all | maxDets=100 ] = 0.396
0:  Average Precision  (AP) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.212
0:  Average Precision  (AP) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.403
0:  Average Precision  (AP) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.482
0:  Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=  1 ] = 0.307
0:  Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets= 10 ] = 0.485
0:  Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.509
0:  Average Recall     (AR) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.317
0:  Average Recall     (AR) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.544
0:  Average Recall     (AR) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.649
```

Some of the recent ML techniques added for the SSD work can immediately improve the accuracy of the FasterRCNN models without any additional changes. Obviously much bigger improvements can be achieved by carefully tuning our training recipes to the needs of each model. Using the SSD data augmentation on `fasterrcnn_resnet50_fpn` we can improve the accuracy by 2.1 mAP:
```
0:  Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.390
0:  Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=100 ] = 0.601
0:  Average Precision  (AP) @[ IoU=0.75      | area=   all | maxDets=100 ] = 0.422
[...]

Issue #4932, opened 2021-11-14
add GN and GIoU loss  for retinanet


xiaohu2015:
### 🚀 The feature

- support GroupNorm on retianet head;
- support GIoU loss for regression loss;

### Motivation, pitch

As we know, retinanet should get better mAP if using GN on retianet head and GIoU loss

cc @datumbox

oke-aditya:
Hi @xiaohu2015 Thanks a lot for the proposal.

Can you please clarify what GN is ?

gIoU is currently supported as `torchvision.ops.generalized_box_iou` (this name was kept for DeTR compatibility) 
and I guess creating loss function from the op is straightforward?
I think it is  `1 - ops.generalized_box_iou` ? 

For loss functions there is an RFC #2980 and I guess gIoU loss is proposed there.

xiaohu2015:
@oke-aditya GroupNorm

xiaohu2015:
@oke-aditya I think the pr is more about performance enhancement for retiannet model.

oke-aditya:
Yes I think it would be great to improve the model. Do you have some statistics of the new mAP possible with GroupNorm and gIoU? I'm asking to so that I could give a read 😅

xiaohu2015:
@oke-aditya I think the mAP can increase about 1~2 point if use GN and GIoU loss.

xiaohu2015:
@oke-aditya I want to add a new model retinanet_gn_resnet50_fpn which uses GN on head.

datumbox:
@xiaohu2015 Thanks for the proposal.

What other changes do you propose to make to the standard retinanet+resnet50 model builder other than passing GroupNorm? If the only thing needed is to pass GN, I'm not sure it's worth adding another model. This is something that users of TorchVision can do easily on their own. Let me know what you think, thanks!

xiaohu2015:
I want to the follow changes:
- add a argument `norm_layer` for  `RetinaNetClassificationHead` and `RetinaNetRegressionHead`
- add a argument `box_reg_loss_type` for `RetinaNet` to support GIoU loss for compute regresion loss.

do you think it valuable？

datumbox:
@xiaohu2015 Thanks for the clarifications and for checking in before making changes.

> add a argument norm_layer for RetinaNetClassificationHead and RetinaNetRegressionHead

[...]

Issue #2263, opened 2020-05-27
Making `transformation` an optional parameter in FasterRCNN


Sentient07:
Hello,

## 🚀 Feature
I think it would be more generic to have `transform`(https://github.com/pytorch/vision/blob/3d65fc6723f1e0709916f24d819d6e17a925b394/torchvision/models/detection/faster_rcnn.py#L231) as a function that can be modified by users rather than a default one.

## Motivation

I am applying transformations separately as a part of data augmentation, which includes cropping and resizing. Hence I would prefer to not do the twice while retraining `FasterRCNN`. 

## Pitch

I would like to have a fixed size input to be fed into the network for variable-sized images. At present, I do this by resizing the images separately as a part of DataLoader and adjust the parameters of `GeneralizedRCNNTransform` accordingly. 

## Alternatives

My present way of using `FasterRCNN` is an alternative. Since my set of transformations are pre-defined, I have to apply hacks such as setting mean to 0., std to 1. and altering min and max sizes to my default value(this would mean that `scale=1` and interpolation would return the same image.

## Additional context

While the input to the network is fixed size, I apply many other augmentations such as mirror, random cropping etc, inspired by SSD based networks. Hence I would prefer to do all augmentation in a separate place once instead of twice.

Thank you!

Sentient07:
Furthermore, this transformation class does not handle cases when targets are empty, leading to this error.

```
loss_dict = model(images, targets)
  File "/home/envs/tf13/lib/python3.7/site-packages/torch/nn/modules/module.py", line 547, in __call__
    result = self.forward(*input, **kwargs)
  File "/home/envs/tf13/lib/python3.7/site-packages/torchvision/models/detection/generalized_rcnn.py", line 47, in forward
    images, targets = self.transform(images, targets)
  File "/home/envs/tf13/lib/python3.7/site-packages/torch/nn/modules/module.py", line 547, in __call__
    result = self.forward(*input, **kwargs)
  File "/home/envs/tf13/lib/python3.7/site-packages/torchvision/models/detection/transform.py", line 41, in forward
    image, target = self.resize(image, target)
  File "/home/envs/tf13/lib/python3.7/site-packages/torchvision/models/detection/transform.py", line 76, in resize
    bbox = resize_boxes(bbox, (h, w), image.shape[-2:])
  File "/home/envs/tf13/lib/python3.7/site-packages/torchvision/models/detection/transform.py", line 137, in resize_boxes
[...]

PR #5444, opened 2022-02-19
Post-paper Detection Optimizations

datumbox:
Related to #5307 and #2263. Touches upon requests recorded at #4932 and #5325.

Our target is to improve the existing RetinaNet, FasterRCNN and MaskRCNN architectures using post-paper optimizations:
- Adds a new RetinaNet variant with post-paper optimizations proposed at "[Bridging the Gap Between Anchor-based and Anchor-free Detection via Adaptive Training Sample Selection](https://arxiv.org/pdf/1912.02424.pdf)".
- Adds a new FasterRCNN and MaskRCNN variants with post-paper optimizations proposed at "[Benchmarking Detection Transfer Learning with Vision Transformers](https://arxiv.org/pdf/2111.11429.pdf)".
- Adds support of elements from @vaibhava0's [recipe](https://github.com/facebookresearch/detectron2/blob/b70a86f913b6ed0b701e5b2734a6342cb3d5c4e3/configs/new_baselines/mask_rcnn_R_50_FPN_100ep_LSJ.py), @xiaohu2015's [recipe](https://github.com/pytorch/vision/issues/5307#issuecomment-1050572394) and @fmassa's work on [DETR](https://arxiv.org/pdf/2005.12872.pdf). Other recipes influencing this work is the ones of [Swin Transformers](https://arxiv.org/pdf/2103.14030.pdf) and [XCiT](https://arxiv.org/pdf/2106.09681.pdf).

This PR contains commits which were later split on separate PRs to assist the reviews. Currently it contains the following changes on top of the main branch:
- Adds support of `norm_layer` in all Detection heads.
- Adds a new extendible `FastRCNNConvFCHead` which follows a similar strategy as the existing `MaskRCNNHeads`.
- Replaces Conv-Norm-BNs with `Conv2dNormActivation` where possible; we use  PyTorch core's `_load_from_state_dict()` approach to maintain BC.
- Extends `RPNHead` to support heavier heads.
- Adds an experimental private `_box_loss` utility which allows for training models with different box losses. No plans to make this public; the entire API is kept private so that we can review our Detection API as a whole and examine the best way to support different Transforms and Losses in Detection models.
- Adds new 3 variants for RetinaNet and *RCNN based on follow up papers.

Best currently available models achieved using this PR (pre-trained weights to be added on separate PR):
- RetinaNet ResNet50 FPN:
  - Old `retinanet_resnet50_fpn`: 36.4 mAP
  - New `retinanet_resnet50_fpn_v2`: 41.5 mAP (+5.1)
- MaskRCNN ResNet50 FPN:
  - Old `maskrcnn_resnet50_fpn`: 37.9 box mAP / 34.6 mask mAP
  - New `maskrcnn_resnet50_fpn_v2`: 47.4 box mAP / 41.8 mask mAP (+9.5/+7.2)
- FasterRCNN ResNet50 FPN:
  - Old `fasterrcnn_resnet50_fpn`: 37.0 mAP
  - New `fasterrcnn_resnet50_fpn_v2`: 46.7 mAP (+9.7)


references/detection/transforms.py
```diff
@@ -326,3 +326,78 @@ def forward(
...
+        elif len(padding) == 1:
+            pad_left = pad_right = pad_top = pad_bottom = padding[0]
+        elif len(padding) == 2:
+            pad_left = pad_right = padding[0]
+            pad_top = pad_bottom = padding[1]
+        else:
+            pad_left = padding[0]
+            pad_top = padding[1]
+            pad_right = padding[2]
+            pad_bottom = padding[3]
```
vfdev-5, line 353: Nit: in D2 they pad only in right/bottom direction:
- https://github.com/facebookresearch/detectron2/blob/9258799e4e72786edd67940872e0ed2c4387aac5/detectron2/data/transforms/augmentation_impl.py#L334-L345
- https://github.com/facebookresearch/fvcore/blob/279b7d093aa078847fe13fefcc2968280abe5297/fvcore/transforms/transform.py#L747-L767

references/detection/transforms.py
```diff
@@ -326,3 +326,78 @@ def forward(
...
+
+        padding = [pad_left, pad_top, pad_right, pad_bottom]
+        img = F.pad(img, padding, self.fill, self.padding_mode)
+        if target is not None:
+            target["boxes"][:, 0::2] += pad_left
+            target["boxes"][:, 1::2] += pad_top
```
vfdev-5, line 359: I think this should be done explicitly for each pad_left, pad_top, pad_right, pad_bottom

references/detection/transforms.py
```diff
@@ -326,3 +326,78 @@ def forward(
...
         return image, target
+
+
+class FixedSizeCrop(nn.Module):
+    def __init__(self, size, fill=0, padding_mode="constant"):
+        super().__init__()
+        size = tuple(T._setup_size(size, error_msg="Please provide only two dimensions (h, w) for size."))
+        self.crop_height = size[0]
+        self.crop_width = size[1]
+        self.fill = tuple(int(x) for x in fill)  # TODO: Fill is currently respected only on PIL. Apply tensor patch.
```
vfdev-5, line 337: fill by default is int and `for x in fill` will fail with default value

references/detection/transforms.py
datumbox: I don't understand, that's what I also do too.

This snippet as I mention above it taken from the `pad()` of `functional_tensor.py`. When this eventually gets refactored and ported to the new API we should consider making it a utility and share its functionality.

references/detection/transforms.py
datumbox: This quick and dirty fix was necessary for PIL images because there is a bug on `pad()`. As we discussed on #5568 these patches needed to be fixed on a new PR. I'm currently busy but I'll patch it soon.

references/detection/transforms.py
vfdev-5: We can ignore my comment as below you specify the padding as `[0, 0, p_right, p_bottom]`.

references/detection/transforms.py
vfdev-5: The comment is about : 
```
t = FixedSizeCrop(size=(36, 14))

> self.fill = tuple(int(x) for x in fill)  # TODO: Fill is currently respected only on PIL. Apply tensor patch.
> 
> TypeError: 'int' object is not iterable
```

datumbox: Oops I see what you say. Let me patch now on main then and avoid this. Thanks for flagging it.

references/detection/transforms.py
```diff
@@ -401,3 +401,39 @@ def forward(self, img, target=None):
...
+    def forward(
+        self, image: Tensor, target: Optional[Dict[str, Tensor]] = None
+    ) -> Tuple[Tensor, Optional[Dict[str, Tensor]]]:
+        _, orig_height, orig_width = F.get_dimensions(image)
+
+        min_size = self.min_size[torch.randint(len(self.min_size), (1,)).item()]
+        r = min(min_size / min(orig_height, orig_width), self.max_size / max(orig_height, orig_width))
+
+        new_width = int(orig_width * r)
+        new_height = int(orig_height * r)
```
vfdev-5, line 427: D2 also does +0.5 here: 
https://github.com/facebookresearch/detectron2/blob/9258799e4e72786edd67940872e0ed2c4387aac5/detectron2/data/transforms/augmentation_impl.py#L193-L194

datumbox: Thanks for the comment. You are right to say that D2 is doing round in this way (it's a common idiom in their codebase) but here I'm porting TorchVision's [GeneralizedRCNNTransform.resize()](https://github.com/pytorch/vision/blob/main/torchvision/models/detection/transform.py#L163) which works very similarly but is not identical. My target is to export this transform from within the model and move it outside so that we can use other transformations such as LSJ.

torchvision/models/detection/faster_rcnn.py
```diff
@@ -298,6 +304,40 @@ def forward(self, x):
         return x
 
 
+class FastRCNNHeads(nn.Sequential):
```
fmassa, line 307: nit: Naming is hard. The first version of Fast R-CNN didn't use FPN, and the head [was actually the last stage of a ResNet](https://github.com/facebookresearch/maskrcnn-benchmark/blob/57eec25b75144d9fb1a6857f32553e1574177daf/maskrcnn_benchmark/modeling/roi_heads/box_head/roi_box_feature_extractors.py#L14). When FPN was introduced, they made the head lighter and thus the name TwoMLP.

I would be ok leaving the name as `FastRNNHeads` in here, but maybe a naming a [bit more specific like the one used in Detectron2](https://github.com/facebookresearch/detectron2/blob/6886f85baee349556749680ae8c85cdba1782d8e/detectron2/modeling/roi_heads/box_head.py#L26) might be preferable?

On an unrelated note, they actually fused both options together by allowing convs followed by linear layers

torchvision/models/detection/_utils.py
```diff
@@ -507,3 +508,26 @@ def _topk_min(input: Tensor, orig_kval: int, axis: int) -> int:
...
     return _fake_cast_onnx(min_kval)
+
+
+def _box_loss(
+    type: str,
+    box_coder: BoxCoder,
+    anchors_per_image: Tensor,
+    matched_gt_boxes_per_image: Tensor,
+    bbox_regression_per_image: Tensor,
+    cnf: Optional[Dict[str, float]] = None,
```
fmassa, line 519: nit: I would use a `**kwargs` here instead, but it's personal taste

datumbox: Yeap you are absolutely right. That's what I would love to use but JIT doesn't like it. :(

torchvision/models/detection/faster_rcnn.py
datumbox: Happy to make the naming change to `FastRCNNConvFCHead`. I used the specific one to keep it aligned with the equivalent `MaskRCNNHeads` class.

> On an unrelated note, they actually fused both options together by allowing convs followed by linear layers

Good idea. I can extend it with `conv_layers` and `fc_layers` params on the head constructor to support both. Would what work?

fmassa: Yes, that would work and could generalize TwoMLPHead, so that it could just inherit from it instead (for BC)

torchvision/models/detection/_utils.py
fmassa: Understood. Maybe sometime in the near future we won't need to bother about torchscript anymore. But we are not yet there

torchvision/models/detection/faster_rcnn.py
datumbox: We discussed offline with @fmassa the possibility of making `TwoMLPHead` inherit from `FastRCNNConvFCHead`. We decided not to do this because in order to keep things BC we would have to add another lengthy `_load_from_state_dict` implementation which adds more code than the inheritance saves. Thus we decided to keep `TwoMLPHead` as-is.

d4l3k:
@datumbox I'm getting errors when trying to load a model using FPN as a frozen layer.

```
  File "/mnt/ext/openape/apedepth/train.py", line 85, in <module>
    model.load_state_dict(state_dict, strict=False)
  File "/home/rice/venvs/openape/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1566, in load_state_dict
    load(self)
  File "/home/rice/venvs/openape/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1564, in load
    load(child, prefix + name + '.')
  File "/home/rice/venvs/openape/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1564, in load
    load(child, prefix + name + '.')
  File "/home/rice/venvs/openape/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1564, in load
    load(child, prefix + name + '.')
  File "/home/rice/venvs/openape/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1560, in load
    module._load_from_state_dict(
  File "/home/rice/venvs/openape/lib/python3.10/site-packages/torchvision/ops/feature_pyramid_network.py", line 131, in _load_from_state_dict
    state_dict[new_key] = state_dict.pop(old_key)
KeyError: 'semantic.backbone.fpn.inner_blocks.0.weight'
```

This seems to be a regression since PT 1.11 stable and I can no longer save/load my model. Excluding the semantic weights when calling torch.load doesn't help either.

My model has

```
        self.semantic = freeze(models.detection.fasterrcnn_mobilenet_v3_large_fpn(pretrained=True))
```

d4l3k:
Adding a `if old_key in state_dict` in a few spots seems to fix it.

```patch
        version = local_metadata.get("version", None)

        if version is None or version < 2:
            for type in ["weight", "bias"]:
                old_key = f"{prefix}conv.{type}"
                new_key = f"{prefix}conv.0.0.{type}"
+                if old_key in state_dict:
                    state_dict[new_key] = state_dict.pop(old_key)
```

datumbox:
@d4l3k Thanks for the heads up.

> This seems to be a regression since PT 1.11 stable and I can no longer save/load my model.

If this is true, it's probably not related to this PR. To reduce any confusion, could you please open a separate issue and provide a way to reproduce the problem?

> Adding a if old_key in state_dict in a few spots seems to fix it.

This might indicate that the old structure is modified and not compatible with v1. The proposed patch is not a solution but rather more likely to mask the issue. If you provide a way to reproduce the issue, we can help you invstigate.

PR #5445, opened 2022-02-19
Add Conv2dNormActivation and Conv3dNormActivation Blocks

### LINKED ISSUE #5430: ConvNormActivation block for 3D convolutions

oke-aditya:
Closes #5430 

Simple code to try this. It worked fine.

```
class SimpleModel(torch.nn.Module):
    def __init__(self) -> None:
        super().__init__()
        self.layer1 = Conv3dNormActivation(3, 15, stride=2, padding=2,
                                           activation_layer=torch.nn.LeakyReLU,
                                           inplace=True,
                                           bias=True)
    
    def forward(self, x: torch.Tensor) -> torch.Tensor:
        x = self.layer1(x)
        return x

class EasyModel(torch.nn.Module):
    def __init__(self) -> None:
        super().__init__()
        self.layer1 = Conv2dNormActivation(3, 15, stride=2, padding=2,
                                           activation_layer=torch.nn.LeakyReLU,
                                           inplace=True,
                                           bias=True)
    
    def forward(self, x: torch.Tensor) -> torch.Tensor:
        x = self.layer1(x)
        return x



if __name__ == "__main__":

    image_tensor = torch.rand((5, 3, 224, 224))
    model = EasyModel()
    out = model(image_tensor)
    print(out.size())  # torch.Size([5, 15, 113, 113])

    video_tensor = torch.rand((5, 3, 10, 224, 224))
    model = SimpleModel()
    out = model(video_tensor)
    print(out.size())  # torch.Size([5, 15, 6, 113, 113]

```

### DIFF
--- a/torchvision/ops/__init__.py
+++ b/torchvision/ops/__init__.py
@@ -52,7 +52,8 @@ __all__ = [
     "StochasticDepth",
     "FrozenBatchNorm2d",
     "ConvNormActivation",
-    "Conv2dNormActivation" "Conv3dNormActivation",
+    "Conv2dNormActivation",
+    "Conv3dNormActivation",
     "SqueezeExcitation",
     "generalized_box_iou_loss",
 ]
--- a/torchvision/ops/__init__.py
+++ b/torchvision/ops/__init__.py
@@ -14,7 +14,7 @@ from .deform_conv import deform_conv2d, DeformConv2d
 from .feature_pyramid_network import FeaturePyramidNetwork
 from .focal_loss import sigmoid_focal_loss
 from .giou_loss import generalized_box_iou_loss
-from .misc import FrozenBatchNorm2d, ConvNormActivation, SqueezeExcitation
+from .misc import FrozenBatchNorm2d, ConvNormActivation, Conv3dNormActivation, SqueezeExcitation
 from .poolers import MultiScaleRoIAlign
 from .ps_roi_align import ps_roi_align, PSRoIAlign
 from .ps_roi_pool import ps_roi_pool, PSRoIPool
@@ -52,6 +52,7 @@ __all__ = [
     "StochasticDepth",
     "FrozenBatchNorm2d",
     "ConvNormActivation",
+    "Conv3dNormActivation",
     "SqueezeExcitation",
     "generalized_box_iou_loss",
 ]
--- a/torchvision/ops/__init__.py
+++ b/torchvision/ops/__init__.py
@@ -14,7 +14,7 @@ from .deform_conv import deform_conv2d, DeformConv2d
 from .feature_pyramid_network import FeaturePyramidNetwork
 from .focal_loss import sigmoid_focal_loss
 from .giou_loss import generalized_box_iou_loss
-from .misc import FrozenBatchNorm2d, ConvNormActivation, Conv3dNormActivation, SqueezeExcitation
+from .misc import FrozenBatchNorm2d, ConvNormActivation, Conv2dNormActivation, Conv3dNormActivation, SqueezeExcitation
 from .poolers import MultiScaleRoIAlign
 from .ps_roi_align import ps_roi_align, PSRoIAlign
 from .ps_roi_pool import ps_roi_pool, PSRoIPool
@@ -52,7 +52,7 @@ __all__ = [
     "StochasticDepth",
     "FrozenBatchNorm2d",
     "ConvNormActivation",
-    "Conv3dNormActivation",
+    "Conv2dNormActivation" "Conv3dNormActivation",
     "SqueezeExcitation",
     "generalized_box_iou_loss",
 ]
--- a/torchvision/ops/misc.py
+++ b/torchvision/ops/misc.py
@@ -1,3 +1,4 @@
+import warnings
 from typing import Callable, List, Optional
 
 import torch
@@ -67,7 +68,6 @@ class FrozenBatchNorm2d(torch.nn.Module):
 
 class ConvNormActivation(torch.nn.Sequential):
     """
-    DEPRECATED
     Use Conv2dNormActivation instead.
     """
 
@@ -114,6 +114,9 @@ class ConvNormActivation(torch.nn.Sequential):
         super().__init__(*layers)
         _log_api_usage_once(self)
         self.out_channels = out_channels
+
+        if self.__class__ == ConvNormActivation:
+            warnings.warn("Don't use ConvNormActivation directly. Use Conv2dNormActivation instead.")
 
 
 class Conv2dNormActivation(ConvNormActivation):
--- a/torchvision/ops/misc.py
+++ b/torchvision/ops/misc.py
@@ -67,7 +67,6 @@ class FrozenBatchNorm2d(torch.nn.Module):
 
 
 class ConvNormActivation(torch.nn.Sequential):
-
     def __init__(
         self,
         in_channels: int,
--- a/torchvision/ops/misc.py
+++ b/torchvision/ops/misc.py
@@ -146,11 +146,10 @@ class Conv2dNormActivation(_ConvNormActivation):
         bias: Optional[bool] = None,
     ) -> None:
 
-        conv_layer = torch.nn.Conv2d
         super().__init__(
             in_channels,
             out_channels,
-            conv_layer,
+            torch.nn.Conv2d,
             kernel_size,
             stride,
             padding,
@@ -196,11 +195,10 @@ class Conv3dNormActivation(_ConvNormActivation):
         bias: Optional[bool] = None,
     ) -> None:
 
-        conv_layer = torch.nn.Conv3d
         super().__init__(
             in_channels,
             out_channels,
-            conv_layer,
+            torch.nn.Conv3d,
             kernel_size,
             stride,
             padding,
--- a/torchvision/ops/misc.py
+++ b/torchvision/ops/misc.py
@@ -67,9 +67,6 @@ class FrozenBatchNorm2d(torch.nn.Module):
 
 
 class ConvNormActivation(torch.nn.Sequential):
-    """
-    Use Conv2dNormActivation instead.
-    """
 
     def __init__(
         self,
--- a/torchvision/ops/misc.py
+++ b/torchvision/ops/misc.py
@@ -1,5 +1,4 @@
-import warnings
-from typing import Callable, List, Optional, Any
+from typing import Callable, List, Optional
 
 import torch
 from torch import Tensor
@@ -66,21 +65,26 @@ class FrozenBatchNorm2d(torch.nn.Module):
         return f"{self.__class__.__name__}({self.weight.shape[0]}, eps={self.eps})"
 
 
-class _ConvNormActivation(torch.nn.Sequential):
+class ConvNormActivation(torch.nn.Sequential):
+    """
+    DEPRECATED
+    Use Conv2dNormActivation instead.
+    """
+
     def __init__(
         self,
         in_channels: int,
         out_channels: int,
-        conv_layer: Callable[..., torch.nn.Module],
         kernel_size: int = 3,
         stride: int = 1,
         padding: Optional[int] = None,
         groups: int = 1,
-        norm_layer: Optional[Callable[..., torch.nn.Module]] = None,
+        norm_layer: Optional[Callable[..., torch.nn.Module]] = torch.nn.BatchNorm2d,
         activation_layer: Optional[Callable[..., torch.nn.Module]] = torch.nn.ReLU,
         dilation: int = 1,
         inplace: Optional[bool] = True,
         bias: Optional[bool] = None,
+        conv_layer: Callable[..., torch.nn.Module] = torch.nn.Conv2d,
     ) -> None:
 
         if padding is None:
@@ -112,7 +116,7 @@ class _ConvNormActivation(torch.nn.Sequential):
         self.out_channels = out_channels
 
 
-class Conv2dNormActivation(_ConvNormActivation):
+class Conv2dNormActivation(ConvNormActivation):
     """
     Configurable block used for Convolution2d-Normalzation-Activation blocks.
 
@@ -149,7 +153,6 @@ class Conv2dNormActivation(_ConvNormActivation):
         super().__init__(
             in_channels,
             out_channels,
-            torch.nn.Conv2d,
             kernel_size,
             stride,
             padding,
@@ -159,10 +162,11 @@ class Conv2dNormActivation(_ConvNormActivation):
             dilation,
             inplace,
             bias,
+            torch.nn.Conv2d,
         )
 
 
-class Conv3dNormActivation(_ConvNormActivation):
+class Conv3dNormActivation(ConvNormActivation):
     """
     Configurable block used for Convolution3d-Normalzation-Activation blocks.
 
@@ -198,7 +202,6 @@ class Conv3dNormActivation(_ConvNormActivation):
         super().__init__(
             in_channels,
             out_channels,
-            torch.nn.Conv3d,
             kernel_size,
             stride,
             padding,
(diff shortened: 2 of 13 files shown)


torchvision/ops/misc.py
```diff
@@ -161,3 +161,62 @@ def _scale(self, input: Tensor) -> Tensor:
...
+        padding: Optional[int] = None,
+        groups: int = 1,
+        norm_layer: Optional[Callable[..., torch.nn.Module]] = torch.nn.BatchNorm3d,
+        activation_layer: Optional[Callable[..., torch.nn.Module]] = torch.nn.ReLU,
+        dilation: int = 1,
+        inplace: Optional[bool] = True,
+        bias: Optional[bool] = None,
+    ) -> None:
+
+        if padding is None:
```
datumbox, line 199: There is lots of duplicate code with 2d. Should we have a base class to avoid it and both implementation call the same base? PyTorch dos something similar with [BatchNorm](https://github.com/pytorch/pytorch/blob/master/torch/nn/modules/batchnorm.py#L338).

oke-aditya: I thought the same. There is duplication. We can probably just create a base class convbatchNorm and then extend that for 2d and 3d ?

jdsgomes: I agree, please see my proposal here: https://github.com/pytorch/vision/pull/5440

torchvision/ops/__init__.py
```diff
@@ -52,7 +52,7 @@
     "StochasticDepth",
     "FrozenBatchNorm2d",
     "ConvNormActivation",
-    "Conv3dNormActivation",
+    "Conv2dNormActivation" "Conv3dNormActivation",
```
datumbox, line 55:
```suggestion
    "Conv2dNormActivation", "Conv3dNormActivation",
```

torchvision/ops/misc.py
```diff
@@ -114,56 +100,125 @@ def __init__(
...
+        stride (int, optional): Stride of the convolution. Default: 1
+        padding (int, tuple or str, optional): Padding added to all four sides of the input. Default: None, in wich case it will calculated as ``padding = (kernel_size - 1) // 2 * dilation``
+        groups (int, optional): Number of blocked connections from input channels to output channels. Default: 1
+        norm_layer (Callable[..., torch.nn.Module], optional): Norm layer that will be stacked on top of the convolution layer. If ``None`` this layer wont be used. Default: ``torch.nn.BatchNorm2d``
+        activation_layer (Callable[..., torch.nn.Module], optinal): Activation function which will be stacked on top of the normalization layer (if not None), otherwise on top of the conv layer. If ``None`` this layer wont be used. Default: ``torch.nn.ReLU``
+        dilation (int): Spacing between kernel elements. Default: 1
+        inplace (bool): Parameter for the activation layer, which can optionally do the operation in-place. Default ``True``
+        bias (bool, optional): Whether to use bias in the convolution layer. By default, biases are included if ``norm_layer is None``.
+
```
datumbox, line 131: If we deprecate, we probably can remove the documentation similar to what we did below:
https://github.com/pytorch/vision/blob/0663377dda32036d3f479ee3e54e7ffcc01e513c/torchvision/models/mobilenetv3.py#L23-L24

torchvision/ops/misc.py
```diff
@@ -114,56 +100,125 @@ def __init__(
...
+            kernel_size,
+            stride,
+            padding,
+            groups,
+            norm_layer,
+            activation_layer,
+            dilation,
+            inplace,
+            bias,
+        )
```
datumbox, line 167: To avoid redefining every single parameter we could do something like:
https://github.com/pytorch/vision/blob/8a16e12f3a7f10d124b26aeb7975cd2bf4a81695/torchvision/models/quantization/inception.py#L28-L29

Then throw the warning afterwards.

Also it seems that this class shouldn't be inheriting from `_ConvNormActivation` but straights from `Conv2dNormActivation` which does exactly the same thing.

torchvision/ops/misc.py
```diff
@@ -196,27 +251,57 @@ def __init__(
...
+
+
+class SqueezeExcitation(torch.nn.Module):
```
datumbox, line 271: I assume this code is completely unmodified and Github fails to create a proper diff. Could you confirm?

datumbox (commented):
@oke-aditya I've added a few suggestions that aim to simplify the code duplication. Let me know what you think. 

All of the proposals assume we will deprecate the old class. There is an alterantive but let's discuss it here after we clean up this proposal.

torchvision/ops/misc.py
oke-aditya: Yes, I haven't touched SqueezeExcitation

torchvision/ops/misc.py
```diff
@@ -114,16 +100,80 @@ def __init__(
...
         ]
+
         if norm_layer is not None:
             layers.append(norm_layer(out_channels))
+
         if activation_layer is not None:
             params = {} if inplace is None else {"inplace": inplace}
             layers.append(activation_layer(**params))
         super().__init__(*layers)
-        _log_api_usage_once(self)
```
oke-aditya, line 123: Where do I log? In the parent class or in each child class individually?

torchvision/ops/misc.py
```diff
@@ -114,16 +100,80 @@ def __init__(
...
+
         self.out_channels = out_channels
 
 
+class Conv2dNormActivation(_ConvNormActivation):
+    """
+    Configurable block used for Convolution-Normalzation-Activation blocks.
+
+    Args:
+        in_channels (int): Number of channels in the input image
```
oke-aditya, line 120: We need to document these both times, as we are not exposing the base class.

torchvision/ops/misc.py
```diff
@@ -114,16 +100,80 @@ def __init__(
...
+        groups (int, optional): Number of blocked connections from input channels to output channels. Default: 1
+        norm_layer (Callable[..., torch.nn.Module], optional): Norm layer that will be stacked on top of the convolution layer. If ``None`` this layer wont be used. Default: ``torch.nn.BatchNorm2d``
+        activation_layer (Callable[..., torch.nn.Module], optinal): Activation function which will be stacked on top of the normalization layer (if not None), otherwise on top of the conv layer. If ``None`` this layer wont be used. Default: ``torch.nn.ReLU``
+        dilation (int): Spacing between kernel elements. Default: 1
+        inplace (bool): Parameter for the activation layer, which can optionally do the operation in-place. Default ``True``
+        bias (bool, optional): Whether to use bias in the convolution layer. By default, biases are included if ``norm_layer is None``.
+
+    """
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
```
datumbox, line 134: Sorry, perhaps I should have been clearer. My earlier comment for using `*args: Any, **kwargs: Any` referred ONLY to the deprecated class. To the normal classes should have proper declarations as you had them. This will allow for better static analysis and IDE integration.

Note that this is also necessary for having different default values on `norm_layers` between 2d and 3d. In this implementation the default should be `torch.nn.BatchNorm2d`, not None as above.

torchvision/ops/misc.py
datumbox: This comment was referring ONLY to the deprecated method... The rest should be kept as normal. Sorry I wasn't clear.

torchvision/ops/misc.py
```diff
@@ -114,16 +100,80 @@ def __init__(
...
+        layer = torch.nn.Conv2d
+        super().__init__(layer=layer, *args, **kwargs)
+
+
+class Conv3dNormActivation(_ConvNormActivation):
+    """
+    Configurable block used for Convolution-Normalzation-Activation blocks.
```
datumbox, line 141:
```suggestion
    Configurable block used for Convolution3d-Normalzation-Activation blocks.
```

torchvision/ops/misc.py
```diff
@@ -124,6 +112,122 @@ def __init__(
...
+        norm_layer: Optional[Callable[..., torch.nn.Module]] = torch.nn.BatchNorm2d,
+        activation_layer: Optional[Callable[..., torch.nn.Module]] = torch.nn.ReLU,
+        dilation: int = 1,
+        inplace: Optional[bool] = True,
+        bias: Optional[bool] = None,
+    ) -> None:
+
+        conv_layer = torch.nn.Conv2d
+        super().__init__(
+            in_channels,
```
oke-aditya, line 151: I guess I can do something like 
`super().__init__(conv_layer=conv_layer, *args, **kwargs)`

datumbox: No really. I think you are good as you are.

torchvision/ops/misc.py
```diff
@@ -154,9 +181,36 @@ class Conv3dNormActivation(_ConvNormActivation):
...
+        padding: Optional[int] = None,
+        groups: int = 1,
+        norm_layer: Optional[Callable[..., torch.nn.Module]] = torch.nn.BatchNorm3d,
+        activation_layer: Optional[Callable[..., torch.nn.Module]] = torch.nn.ReLU,
+        dilation: int = 1,
+        inplace: Optional[bool] = True,
+        bias: Optional[bool] = None,
+    ) -> None:
+
+        conv_layer = torch.nn.Conv3d
```
datumbox, line 199: Same here

datumbox (commented):
Last couple of nits:

jdsgomes:
I agree with the comments from @datumbox. I suggested otherwise initially , but the fact is that there are already 39 occurrences of  `ConvNormActivation` internally and 81 occurences of `ops.misc import ConvNormActivation` in Github (and 374 of `ConvNormActivation` which might contain similarly named blocks from other implementations). This on its own wont stop from us deprecating it, but together with the fact that is was introduced recently it does’t make sense to deprecated it. 

So @oke-aditya agrees, I would suggest to make the changes proposed by @datumbox and we can get it merged!

jdsgomes:
@oke-aditya raised a good point offline related to this changes. This was recently exposed in [ops/__init__.py](https://github.com/pytorch/vision/blob/release/0.12/torchvision/ops/__init__.py) and we will be reverting this on a separate PR in the spirit not promoting the usage of `ConvNormActivation` directly. 

So I think I can work on this separately and create a PR that will
- remove ConvNormActivation `ops/__init__.py`
- change all imports of `ops.convNormActivation` to `ops.misc.ConvNormActivation`
- remove all mentions of `ConvNormActivation` in the docs

oke-aditya:
So in a brief offline discussion with @jdsgomes 

We are exposing `ConvNormActivatoin` in current release branch
https://github.com/pytorch/vision/blob/release/0.12/torchvision/ops/__init__.py
@jdsgomes  will patch this :smile: 

Also, fully agreeing that PyTorch is highly stable and we need to keep our codebase backward compatible and avoid such sudden changes. (Having the power to deprecate doesn't mean we should keep using it often :smile: ) 

I will refactor the implementation to be compatible with the old one as @datumbox suggested.

torchvision/ops/misc.py
```diff
@@ -67,21 +67,8 @@ def __repr__(self) -> str:
...
-        stride (int, optional): Stride of the convolution. Default: 1
-        padding (int, tuple or str, optional): Padding added to all four sides of the input. Default: None, in wich case it will calculated as ``padding = (kernel_size - 1) // 2 * dilation``
-        groups (int, optional): Number of blocked connections from input channels to output channels. Default: 1
-        norm_layer (Callable[..., torch.nn.Module], optional): Norm layer that will be stacked on top of the convolutiuon layer. If ``None`` this layer wont be used. Default: ``torch.nn.BatchNorm2d``
-        activation_layer (Callable[..., torch.nn.Module], optinal): Activation function which will be stacked on top of the normalization layer (if not None), otherwise on top of the conv layer. If ``None`` this layer wont be used. Default: ``torch.nn.ReLU``
-        dilation (int): Spacing between kernel elements. Default: 1
-        inplace (bool): Parameter for the activation layer, which can optionally do the operation in-place. Default ``True``
-        bias (bool, optional): Whether to use bias in the convolution layer. By default, biases are included if ``norm_layer is None``.
-
+    DEPRECATED
```
datumbox, line 70: It's not deprecated. It's just a schrodinger class (public and private at the same time). 🐈‍⬛ 

Just removed the documentation and we are good.

torchvision/ops/misc.py
```diff
@@ -124,6 +116,105 @@ def __init__(
         self.out_channels = out_channels
```
oke-aditya, line 117: If I add a warning here. 
```
        warnings.warn(
            "Do not use ConvNormActivation class directly. "
            "Instead use torchvision.ops.Conv2dNormActivation instead.",
            FutureWarning,
        )
```


This will also get raised when user uses `Conv2dNormActivation` or `Conv3dNormActivation` as we are calling super() of this class.
I don't know about workaround to avoid raising such inherited warning.

torchvision/ops/misc.py
```diff
@@ -97,13 +84,16 @@ def __init__(
         dilation: int = 1,
         inplace: Optional[bool] = True,
         bias: Optional[bool] = None,
+        conv_layer: Callable[..., torch.nn.Module] = torch.nn.Conv2d,
     ) -> None:
+
```
datumbox, line 89: I believe we can detect it if someone instantiated `ConvNormActivation` directly and throw a warning.

The way to do this is by something like:
```python
if self.__class__ == ConvNormActivation:
    warn("Please dont call me directly ....")
```

torchvision/ops/misc.py
datumbox: I was reviewing at the same time again... Github should add a feature to let you know someone is writing comments.

See my suggestion here: https://github.com/pytorch/vision/pull/5445#discussion_r814141497

torchvision/ops/misc.py
oke-aditya: Yeah schrodinger cats are everywhere! :) Fixed

torchvision/ops/misc.py
oke-aditya: Nice trick! I didn't know this. Works like charm. Fixed it!

jdsgomes (commented):
LGTM, thank you for working on this @oke-aditya !

torchvision/ops/misc.py
```diff
@@ -114,15 +101,119 @@ def __init__(
...
         if activation_layer is not None:
             params = {} if inplace is None else {"inplace": inplace}
             layers.append(activation_layer(**params))
         super().__init__(*layers)
         _log_api_usage_once(self)
         self.out_channels = out_channels
 
+        if self.__class__ == ConvNormActivation:
+            warnings.warn("Don't use ConvNormActivation directly. Use Conv2dNormActivation instead.")
```
datumbox, line 116:
```suggestion
            warnings.warn("Don't use ConvNormActivation directly, please use Conv2dNormActivation and Conv3dNormActivation instead.")
```

datumbox (commented):
@oke-aditya  Looks good! 

The work on the block itself is complete but now we need to replace all the usages of `ConvNormActivation` from the code-base with `Conv2dNormActivation`. This update needs to happen on this PR to avoid generating unwanted warnings to our users.

oke-aditya:
Yes I'm on it!!

oke-aditya:
Please compare https://github.com/pytorch/vision/pull/5445/commits/9a5ab98c2566486cb608e1c41cde02c654d199eb 
and https://github.com/pytorch/vision/pull/5440/files 

I think I have replaced all the places!

datumbox (approved):
LGTM, thanks @oke-aditya!

@jdsgomes What you thing of cherrypicking this on the release so that we don't advertise using ConvNormActication on the mobilenets? Did the final RC was produced by the release engineering team?

jdsgomes:
After discussing offline with @datumbox we decided not to cherrypick this one as we missed the official cut off date and this is quite a bit change.

jdsgomes (approved):
LGTM!

PR #5446, opened 2022-02-20
Fix bbox scaling estimation for Large Scale Jitter

datumbox:
Fixing the bbox scaling miscalculation for when the the target_size is not proportional to the original image.


datumbox:
Merging fix to be able to continue the training

PR #5422, opened 2022-02-15
add HMDB51 and UCF101 datasets as well as prototype for new style video decoding

pmeier:
Redo of #5335 since I can't reopen it after it was automatically closed.

PR #5447, opened 2022-02-21
fix prototype resource loading

pmeier:
The current logic checks for any files with the same stem regardless of the ending

https://github.com/pytorch/vision/blob/e88a55493eadaab15d1df432fe248e70e04d29e4/torchvision/prototype/datasets/utils/_resource.py#L93

It should be checking for the same stem and a different suffix which starts with a `.`. Without this patch, a file like `foo_bar.zip` is a valid path candidate for `foo.zip`.

This was independently discovered in

- #5442 (cc @abhi-glitchhg ): `cars_test.tgz` and `cars_test_annos_withlabels.mat`
- #5422: `UCF101.rar` and `UCF101TrainTestSplits-RecognitionTask.zip`

### DIFF
--- a/torchvision/prototype/datasets/utils/_resource.py
+++ b/torchvision/prototype/datasets/utils/_resource.py
@@ -22,7 +22,6 @@ from torchvision.datasets.utils import (
     download_file_from_google_drive,
     _get_redirect_url,
     _get_google_drive_file_id,
-    tqdm,
 )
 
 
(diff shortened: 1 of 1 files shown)


NicolasHug (approved):
Stamping to unlock.

If the logic starts to become complex and with various edge-cases that we're not seeing yet (like the one that just occured), it might be worth considering a different approach for this eventually

PR #5448, opened 2022-02-21
refactor prototype datasets to inherit from IterDataPipe

pmeier:
This adds initial support for simplifying our prototype dataset, by letting them inherit directly from `IterDataPipe`.


torchvision/prototype/datasets/utils/_dataset.py
```diff
@@ -181,3 +181,27 @@ def load(
 
     def _generate_categories(self, root: pathlib.Path) -> Sequence[Union[str, Sequence[str]]]:
         raise NotImplementedError
+
+
+class Dataset2(IterDataPipe[Dict[str, Any]], abc.ABC):
+    def __init__(self, root: Union[str, pathlib.Path], *, skip_integrity_check: bool = False) -> None:
```
NicolasHug, line 187: just wondering if there exist a builtin type that is equivalent to `Union[str, pathlib.Path]` ?

torchvision/prototype/datasets/_builtin/imagenet.py
```diff
@@ -201,3 +205,47 @@ def _generate_categories(self, root: pathlib.Path) -> List[Tuple[str, ...]]:
...
+        categories, wnids = info["categories"], info["wnids"]
+
+        self._old_style_dataset = ImageNet()
+        self._old_style_config = self._old_style_dataset.info.make_config(split=self._split)
+
+        self.categories = categories
+        self.info = SimpleNamespace(
+            wnid_to_category=zip(wnids, categories),
+            category_to_wnid=zip(categories, wnids),
+        )
```
NicolasHug, line 236: Do I understand correctly that we're only setting `self.info` for `_datapipe` to work, and that we'll remove it soon?

torchvision/prototype/datasets/_builtin/imagenet.py
```diff
@@ -201,3 +205,47 @@ def _generate_categories(self, root: pathlib.Path) -> List[Tuple[str, ...]]:
...
+
+        super().__init__(root)
+
+    def _resources(self) -> List[OnlineResource]:
+        return self._old_style_dataset.resources(self._old_style_config)
+
+    def _datapipe(self, resource_dps: List[IterDataPipe]) -> IterDataPipe[Dict[str, Any]]:
+        return self._old_style_dataset._make_datapipe(resource_dps, config=self._old_style_config)
```
NicolasHug, line 244: Should we start removing calls to the old design already?

NicolasHug (commented):
Quick pass, thanks @pmeier

pmeier:
Blocked by pytorch/data#237.

pmeier:
As discussed offline with @NicolasHug , we are going to keep the changes in a feature branch for now and send PRs against that. This way we can still go one dataset at a time without worrying about the intermediate phase were some datasets are ported and some are not.

test/builtin_dataset_mocks.py
```diff
@@ -61,27 +58,30 @@ def _parse_mock_info(self, mock_info):
...
 
-        available_file_names = {path.name for path in root.glob("*")}
-        required_file_names = {resource.file_name for resource in self.dataset.resources(config)}
-        missing_file_names = required_file_names - available_file_names
-        if missing_file_names:
-            raise pytest.UsageError(
-                f"Dataset '{self.name}' requires the files {sequence_to_str(sorted(missing_file_names))} "
-                f"for {config}, but they were not created by the mock data function."
-            )
+        # # FIXME: We need to handle missing files here
```
pmeier, line 67: Before we could access the resources without creating the datapipe and thus without loading the resources. This is no longer possible in the new design. We need to think about a new way to check for missing files. Note that this only makes the testing logic more complex, but has no effect on the user. 

One solution would be to mock out `Dataset2.__init__()`. I'll see if that is possible.

test/builtin_dataset_mocks.py
```diff
@@ -95,6 +95,8 @@ def parametrize_dataset_mocks(*dataset_mocks, marks=None):
     for mock in dataset_mocks:
         if isinstance(mock, DatasetMock):
             mocks[mock.name] = mock
+        elif isinstance(mock, collections.abc.Sequence):
```
pmeier, line 98: @NicolasHug we already agreed that the way we are parametrizing needs to change. If you don't oppose I wouldn't do that in this PR though.

test/builtin_dataset_mocks.py
```diff
@@ -111,21 +113,20 @@ def parametrize_dataset_mocks(*dataset_mocks, marks=None):
...
 
 
-DATASET_MOCKS = {}
+DATASET_MOCKS = []
 
 
 def register_mock(fn):
-    name = fn.__name__.replace("_", "-")
-    DATASET_MOCKS[name] = DatasetMock(name, fn)
+    # TODO: remove this decorator after all datasets have been migrated
```
pmeier, line 129: The decorator is a no-op with this change and thus allows us to go one by one for each dataset without touching everything in the first PR.

test/builtin_dataset_mocks.py
```diff
@@ -217,7 +218,7 @@ def mnist(info, root, config):
     )
 
 
-DATASET_MOCKS.update({name: DatasetMock(name, mnist) for name in ["fashionmnist", "kmnist"]})
+# DATASET_MOCKS.update({name: DatasetMock(name, mnist) for name in ["fashionmnist", "kmnist"]})
```
pmeier, line 221: This is the only instance where mocks are registered without using the decorator. Thus, we disable them manually.

test/builtin_dataset_mocks.py
```diff
@@ -30,13 +29,11 @@
 
 
 class DatasetMock:
-    def __init__(self, name, mock_data_fn):
-        self.dataset = find(name)
-        self.info = self.dataset.info
-        self.name = self.info.name
-
+    def __init__(self, name, *, mock_data_fn, configs):
```
pmeier, line 32: We need to manually specify the `configs` now since we can no longer access them from the datasets. Note that this has nothing to do with the (soon to be) removed `DatasetConfig` class, but is just a list of option dictionaries, e.g. `[dict(split="train"), dict(split="test")]`.

test/builtin_dataset_mocks.py
```diff
@@ -111,21 +113,20 @@ def parametrize_dataset_mocks(*dataset_mocks, marks=None):
...
+            pytest.param(dataset_mock, options, id=config_id(name, options), marks=marks.get(name, ()))
             for name, dataset_mock in dataset_mocks.items()
-            for config in dataset_mock.configs
+            for options in dataset_mock.configs
         ],
     )
 
 
-DATASET_MOCKS = {}
+DATASET_MOCKS = []
```
pmeier, line 125: Since we now need to manually add stuff, a list is simpler than a dictionary.

test/test_prototype_builtin_datasets.py
```diff
@@ -62,35 +71,18 @@ def test_sample(self, test_home, dataset_mock, config):
...
-        dataset = datasets.load(dataset_mock.name, **config)
-
-        num_samples = 0
-        for _ in dataset:
-            num_samples += 1
-
-        assert num_samples == mock_info["num_samples"]
-
-    @parametrize_dataset_mocks(DATASET_MOCKS)
-    def test_decoding(self, test_home, dataset_mock, config):
```
pmeier, line 77: I've removed this test since it is obsolete after #5287.

test/builtin_dataset_mocks.py
NicolasHug: > access the resources without creating the datapipe and thus without loading the resources


If the resources are already mocked, is this actually a problem?

pmeier: You are right in that the test would also fail if we did nothing here. The issue is that the error message wouldn't be expressive so it is harder for a contributor to fix the problem. 

However, this is already resolved. The solution I mentioned above worked out great and so we need very little changes to the old checking logic.

test/builtin_dataset_mocks.py
```diff
@@ -61,27 +60,30 @@ def _parse_mock_info(self, mock_info):
 
         return mock_info
 
-    def prepare(self, home, config):
+    def prepare(self, home, **options):
```
NicolasHug, line 63: We're using the name `options` here but IIUC just above we call it `config`? Perhaps worth using a single name? (I wouldn't mind keeping config to minimize the changes)

test/builtin_dataset_mocks.py
NicolasHug: Do you think there's an opportunity for simplification if we relied on classes for the mocks and expose a `CONFIG` attribute, much like the `ADDITIONAL_CONFIGS` attribute of the non-prototype datasets?

test/builtin_dataset_mocks.py
```diff
@@ -482,6 +482,17 @@ def imagenet(info, root, config):
     return num_samples
 
 
+DATASET_MOCKS.append(
+    DatasetMock(
+        "imagenet",
+        mock_data_fn=imagenet_mock_data_fn,
+        configs=combinations_grid(
+            split=("train", "val", "test"),
+        ),
```
NicolasHug, line 491: Kind of related to my comment above about `ADDITIONAL_CONFIGS`, possibly as an alternative:

Would it be reasonably feasible  to have a `@register_mock` decorator to which we would pass the `configs`? Something like

```py
@register_mock(configs=combinations_grid(...))
def imagenet_mock_data_fn(...):
     ....
```


```

File: test/builtin_dataset_mocks.py [hunk 13]
```diff
@@ -430,26 +433,23 @@ def caltech256(info, root, config):
     return num_images_per_category * len(info.categories)
 
 
-@register_mock
-def imagenet(info, root, config):
+def imagenet_mock_data_fn(info, root, **options):
```
NicolasHug, line 436: Do we need to pass the `info`s, or could we just access/build the info dict from the within the mock?

NicolasHug (approved):
Thanks a lot @pmeier , I have some minor comments and questions but nothing blocking. Happy to move forward with this and potentially go back to these later. LGTM!

test/builtin_dataset_mocks.py
pmeier: You mean like have a `class MNISTDatasetMock(DatasetMock)` that overrides a `mock_data_fn` method?

torchvision/prototype/datasets/utils/_dataset.py
```diff
@@ -181,3 +182,41 @@ def load(
 
     def _generate_categories(self, root: pathlib.Path) -> Sequence[Union[str, Sequence[str]]]:
         raise NotImplementedError
+
+
+class Dataset2(IterDataPipe[Dict[str, Any]], abc.ABC):
+    @staticmethod
+    def _verify_str_arg(
```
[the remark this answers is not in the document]
pmeier: Unless we have a use case for `verify_str_arg` outside of our datasets, I would simply move the functionality into this method later. I just didn't do it here to avoid copy-pasting the code.

Even if we move `verify_str_arg` to say `torch.utils._internal`, I still think it is ok to have it as a method on the dataset class. We already have quite a few imports at the top of each dataset implementation. We probably don't need even more.

torchvision/prototype/datasets/_builtin/imagenet.py
```diff
@@ -176,27 +168,28 @@ def _make_datapipe(
...
+            "val": 50_000,
+            "test": 100_000,
+        }[self._split]
 
-    def _generate_categories(self, root: pathlib.Path) -> List[Tuple[str, ...]]:
-        config = self.info.make_config(split="val")
-        resources = self.resources(config)
+    @classmethod
+    def _generate_categories(cls, root: pathlib.Path) -> List[Tuple[str, ...]]:
+        dataset = cls(root, split="val")
```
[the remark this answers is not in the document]
pmeier: Ah yes, I forgot to refactor. This was just my quick and dirty fix to be compliant with the old behavior. I'll refactor

test/builtin_dataset_mocks.py
pmeier: That works here, yes. If the `mock_data_fn` is the only function that we need for a dataset mock, I think this is preferable to overwriting `DatasetMock`.

test/builtin_dataset_mocks.py
pmeier: Like doing `info = datasets.info("imagenet")` here? I'm fine with that. I mostly left it in for compatibility with the old approach.

torchvision/prototype/datasets/utils/_dataset.py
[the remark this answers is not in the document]
NicolasHug: Moving it as a method and removing it from utils sounds good.

I wouldn't worry about the number of imports personally

test/builtin_dataset_mocks.py
NicolasHug: OK sounds good.
TBH it's also perfectly fine to keep what we currently have `DATASET_MOCKS.append(...)` as a first version. I might be early-optimizing here.

test/builtin_dataset_mocks.py
NicolasHug: I think the plural is because it refers to argumentS (`kwargs`), but it doesn't shock me to pass `**some_dict` without a plural. I'd err towards consistency in the naming here - no pref between **options or **config

test/builtin_dataset_mocks.py
NicolasHug: yes, something like that. But this is the kind of optimization that we can do later

PR #5450, opened 2022-02-21
Adding EfficientNetV2 architecture

datumbox:
Related to #2707

Adds EfficientNetV2 implementation on the existing EfficientNet class. 

This PR is influenced by earlier work done by @xiaohu2015 at #4910


torchvision/models/efficientnet.py
```diff
@@ -178,12 +256,19 @@ def __init__(
...
-        if block is None:
-            block = MBConv
+        if "block" in kwargs:
+            warnings.warn(
+                "The parameter 'block' is deprecated since 0.13 and will be removed 0.15. "
+                "Please pass this information on 'MBConvConfig.block' instead."
+            )
+            if kwargs["block"] is not None:
+                for s in inverted_residual_setting:
+                    if isinstance(s, MBConvConfig):
```
jdsgomes, line 270: Nit: according to the signature this should always be True, unless the caller passes the wrong type, in which case I think it is ok to break.

jdsgomes (commented):
Looks good to me so far. Just left a small nit

torchvision/models/efficientnet.py
datumbox: Good catch. The signature was wrong. It should have been `Union[MBConvConfig, FusedMBConvConfig]`. I'll correct now.

xiaohu2015:
> Related to #2707
> 
> Adds EfficientNetV2 implementation on the existing EfficientNet class.
> 
> This PR is influenced by earlier work done by @xiaohu2015 at #4910

Do you have checked the accuray of the converted TF weights. In TF, they used a different BN parameter:
`
norm_layer = partial(nn.BatchNorm2d, eps=1e-03)
` 
this difference affect the accuracy, another gap maybe lay in the padding mode.

datumbox:
@xiaohu2015 Yes. You get a hit of about 0.3 on the Small model (`Acc@1 83.602 Acc@5 96.556`).

The reason why I haven't overwritten the BN configuration is because we plan to train it from scratch rather than using the TF weights. There are a few techniques used on the paper that are not supported by our reference scripts (like progressive learning) so depending on how close the delta from the training will be we might try to close this gap.

torchvision/models/efficientnet.py
```diff
@@ -149,27 +172,88 @@ def forward(self, input: Tensor) -> Tensor:
...
 
         Args:
-            inverted_residual_setting (List[MBConvConfig]): Network structure
+            inverted_residual_setting (Sequence[Union[MBConvConfig, FusedMBConvConfig]]): Network structure
             dropout (float): The droupout probability
             stochastic_depth_prob (float): The stochastic depth probability
             num_classes (int): Number of classes
-            block (Optional[Callable[..., nn.Module]]): Module specifying inverted residual building block for mobilenet
             norm_layer (Optional[Callable[..., nn.Module]]): Module specifying the normalization layer to use
+            last_channel (int): The number of channels on the penultimate layer
```
jdsgomes, line 256: NIT: perhaps `lastconv_output_channels` would be a better name

jdsgomes: If you decide to make this change there are many occurrences in the codebase

jdsgomes (commented):
LGTM, just have a small nit

xiaohu2015:
> @xiaohu2015 Yes. You get a hit of about 0.3 on the Small model (`Acc@1 83.602 Acc@5 96.556`).
> 
> The reason why I haven't overwritten the BN configuration is because we plan to train it from scratch rather than using the TF weights. There are a few techniques used on the paper that are not supported by our reference scripts (like progressive learning) so depending on how close the delta from the training will be we might try to close this gap by making the proposed patch.

I have a question: you got 83.1 top-1 acc using TF weights, is this result tested on `nn.BatchNorm2d, eps=1e-03`?

torchvision/models/efficientnet.py
datumbox: That's what I originally had but then I noticed that the `last_channel` is used in other places such as the MobileNetV3 so I thought to make it aligned. I don't think we can change it at MobileNetV3 because that would be BC-breaking and not sure it's worth it.

datumbox:
@xiaohu2015 Yes correct. The accuracy you see on the source code (83.1) is with the TF weights and without the BN patch. After applying the BN patch, we can reach 83.6. The reason why the BN patch is not on the source code is because I'm currently training the model from scratch and I want to see if that's necessary. Note that the ported TF weights are just added for my convenience to be able to run some of the tests and for checking that nothing is fundamentally broken with the implementation.

xiaohu2015:
> @xiaohu2015 Yes correct. The accuracy you see on the source code (83.1) is with the TF weights and without the BN patch. After applying the BN patch, we can reach 83.6. The reason why the BN patch is not on the source code is because I'm currently training the model from scratch and I want to see if that's necessary. Note that the ported TF weights are just added for my convenience to be able to run some of the tests and for checking that nothing is fundamentally broken with the implementation.

thanks. because I used the timm weights (converted from TF), but I cannot get such accuacy even with the BN patch (about 83.0%), maybe I missed some things.

datumbox:
@xiaohu2015 I've just replaced the weights for the Small variant with some trained from scratch using TorchVision's recipe. We can do better than the paper by ~0.3 points:
`Acc@1 84.228 Acc@5 96.878`

The above means that we don't have to implement TF specific tricks to reproduce the paper, which massively simplifies our code. I kicked off the training for the Medium and Large variants, so once we have them I'll update the PR and release.

jdsgomes (approved):
LGTM thank you!

PR #5452, opened 2022-02-21
add EuroSAT prototype dataset

### LINKED ISSUE #5338: EuroSAT
cc @pmeier @bjuncek

Dbhasin1:
fixes #5338


.pre-commit-config.yaml
```diff
@@ -3,12 +3,17 @@ repos:
     rev: v4.0.1
     hooks:
       - id: check-docstring-first
+        language_version: python3
```
pmeier, line 6: Could you revert these? Not sure where this is coming from.

torchvision/prototype/datasets/_builtin/__init__.py
```diff
@@ -5,10 +5,11 @@
 from .coco import Coco
 from .cub200 import CUB200
 from .dtd import DTD
+from .eurosat import EuroSAT
 from .fer2013 import FER2013
 from .gtsrb import GTSRB
 from .imagenet import ImageNet
-from .mnist import MNIST, FashionMNIST, KMNIST, EMNIST, QMNIST
+from .mnist import EMNIST, FashionMNIST, KMNIST, MNIST, QMNIST
```
pmeier, line 12: Could you revert that? It looks like you are using `usort>=1.*`. Since I saw this on another PR as well, could you explain how you installed it?

torchvision/prototype/datasets/_builtin/eurosat.py
```diff
@@ -0,0 +1,51 @@
...
+from torchdata.datapipes.iter import IterDataPipe, Mapper
+from torchvision.prototype.datasets.utils import Dataset, DatasetConfig, DatasetInfo, HttpResource, OnlineResource
+from torchvision.prototype.datasets.utils._internal import hint_sharding, hint_shuffling
+from torchvision.prototype.features import EncodedImage, Label
+
+
+class EuroSAT(Dataset):
+    def _make_info(self) -> DatasetInfo:
+        return DatasetInfo(
+            "EuroSAT",
```
pmeier, line 12: For now, the canonical dataset names are all lower case

```suggestion
            "eurosat",
```

torchvision/prototype/datasets/_builtin/eurosat.py
```diff
@@ -0,0 +1,51 @@
...
+            ),
+        )
+
+    def resources(self, config: DatasetConfig) -> List[OnlineResource]:
+        url_root = "https://madm.dfki.de/files/sentinel"
+        data = HttpResource(
+            f"{url_root}/EuroSAT.zip",
+            sha256="8ebea626349354c5328b142b96d0430e647051f26efc2dc974c843f25ecf70bd",
+        )
+        return [data]
```
pmeier, line 33: Using an URL root makes the use case of multiple files from the same server more concise. With only one URL, it is not needed

```suggestion
        return [
            HttpResource(
                f"https://madm.dfki.de/files/sentinel/EuroSAT.zip",
                sha256="8ebea626349354c5328b142b96d0430e647051f26efc2dc974c843f25ecf70bd",
            )
        ]
```

torchvision/prototype/datasets/_builtin/eurosat.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def _prepare_sample(self, data: Tuple[str, Any]) -> Dict[str, Any]:
+        image_path = data[0]
+        category = image_path.split("/")[-2]
+        buffer = data[1]
+        return dict(
+            label=Label.from_category(category, categories=self.categories),
+            path=image_path,
+            image=EncodedImage.from_file(buffer),
+        )
```
pmeier, line 43: [`pathlib`](https://docs.python.org/3/library/pathlib.html) can make this more readable

```suggestion
        path, buffer = data
        category = pathlib.Path(path).parent.name
        return dict(
            label=Label.from_category(category, categories=self.categories),
            path=path,
            image=EncodedImage.from_file(buffer),
        )
```

torchvision/prototype/datasets/_builtin/eurosat.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def _make_datapipe(
+        self, resource_dps: List[IterDataPipe], *, config: DatasetConfig
+    ) -> IterDataPipe[Dict[str, Any]]:
+        images_dp = resource_dps[0]
+        images_dp = hint_sharding(images_dp)
+        images_dp = hint_shuffling(images_dp)
+        return Mapper(images_dp, self._prepare_sample)
```
pmeier, line 51: Nit: although you are correct that this is the images datapipe, it is also the only datapipe. Thus, there is no need to namespace it.

```suggestion
        dp = resource_dps[0]
        dp = hint_sharding(dp)
        dp = hint_shuffling(dp)
        return Mapper(dp, self._prepare_sample)
```

pmeier (commented):
Thanks a lot for the PR @Dbhasin1! I only have some style nit inline. Apart from that, we should also add a test for the dataset. To do that you add a

```py
@register_mock
def eurosat(info, root, config):
    ...
```

to the file `test/builtin_dataset_mocks.py`. The body of this function should generate mock data in the same structure of the original dataset, but with only a few samples. Luckily, we already have this functionality for the old dataset in

https://github.com/pytorch/vision/blob/c530b623a91238513a539da3791f6437bf3ce8fa/test/test_datasets.py#L2176-L2190

Thus, you only need to port this as well to the new architecture. Let me know if you face any issues with that.

.pre-commit-config.yaml
Dbhasin1: These were added by me to avoid getting errors as the one attached below whenever I made a commit
`An unexpected error has occurred: CalledProcessError: command: ('/home/dbhasin1/.conda/envs/vision-cpu/bin/python3.1', '-mvirtualenv', '/home/dbhasin1/.cache/pre-commit/repox5vbpbp2/py_env-python3.1', '-p', 'python3.1')
return code: 1
expected return code: 0
stdout:
    RuntimeError: failed to find interpreter for Builtin discover of python_spec='python3.1'
    
stderr: (none)`

torchvision/prototype/datasets/_builtin/__init__.py
Dbhasin1: I had some auto formatter enabled in vscode, that's how these changes were made. `save without formatting` seemed to work to get rid of these

test/builtin_dataset_mocks.py
```diff
@@ -10,13 +10,13 @@
 import pickle
 import random
 import xml.etree.ElementTree as ET
-from collections import defaultdict, Counter
```
pmeier, line 13: Please revert this as it has nothing to do with the PR. In general, you probably want to disable auto formatters that are not compatible with our code format.

test/builtin_dataset_mocks.py
```diff
@@ -1299,6 +1299,25 @@ def cub200(info, root, config):
     return num_samples_map[config.split]
 
 
+@register_mock
+def eurosat(info, root, config):
+    print("info", info)
```
pmeier, line 1304:
```suggestion
```

pmeier (commented):
Two minor things left. Otherwise LGTM when CI is green. Thanks a lot @Dbhasin1!

test/builtin_dataset_mocks.py
Dbhasin1: This change was actually made by the `ufmt` code formatter, which I used as specified in the contributing guide (see: https://github.com/pytorch/vision/blob/main/CONTRIBUTING.md#formatting). Should I still revert it?

pmeier: Yes, please. We need to fix the contribution guide to pin the right versions for the auto formatters. cc @NicolasHug 

@Dbhasin1 This change is made by `usort>=1.*`, but we are currently enforcing

https://github.com/pytorch/vision/blob/efb073682101e1f4f43e1f96a755f1e8ae6958ea/.pre-commit-config.yaml#L19

Would you be up to send a PR that changes the instruction to read `pip install ufmt==1.3.2 black==21.9b0 usort==0.6.4`?

Dbhasin1: sure @pmeier, I can do that

CONTRIBUTING.md
```diff
@@ -83,7 +83,7 @@ Instead of relying directly on `black` however, we rely on
 [ufmt](https://github.com/omnilib/ufmt), for compatibility reasons with Facebook
 internal infrastructure.
 
-To format your code, install `ufmt` with `pip install ufmt` and use e.g.:
+To format your code, install `ufmt` with `pip install ufmt==1.3.2 black==21.9b0 usort==0.6.4` and use e.g.:
```
pmeier, line 86: Could you send this patch in a separate PR given that it has nothing to do with the EuroSAT dataset?

pmeier (approved):
One minor thing left. Otherwise LGTM! Thanks a ton @Dbhasin1!

NicolasHug (approved):
stamping, thanks @Dbhasin1

PR #5454, opened 2022-02-22
disable ufmt pre-commit hook

pmeier:
Until omnilib/ufmt#56 is resolved, the `ufmt` hook cannot be installed and thus breaks the whole lint CI workflow (see for example https://app.circleci.com/pipelines/github/pytorch/vision/14964/workflows/2a6aab2b-9629-44ae-bf50-2b9c42dc85ca/jobs/1205317). Thus, we should disable it for now and wait until this is fixed upstream.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -260,14 +260,16 @@ jobs:
     steps:
       - checkout
       - pip_install:
-          args: pre-commit
+          args: pre-commit ufmt==1.3.0 black==21.9b0 usort==0.6.4
           descr: Install lint utilities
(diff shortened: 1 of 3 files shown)


NicolasHug:
I wonder if we should still keep the lint workflow failing until they fix it on ufmt's side, rather than risk merging un-formatted PRs?

NicolasHug:
Or, if this is a `pre-commit` related issue, still do the checks without using `pre-commit` ?

datumbox (approved):
LGTM!

PR #5455, opened 2022-02-22
call tensor-only kernels from prototype kernels if possible

pmeier:
Addresses the first part of https://github.com/pytorch/vision/pull/5421#discussion_r811798298. The idea behind this change is to make it more clear that our low-level kernels are tensor only and thus cannot use the same route as PIL images in the dispatch. There are two caveats:

1. Some old kernels are only implemented for tensors and thus the kernel is implemented in `torchvision.transforms.functional` directly:
    - `erase`
    - `normalize`
2. Some kernels are composite and thus don't have a PIL or tensor implementation and thus are also only implemented in `torchvision.transforms.functional`
    - `center_crop`
    - `resized_crop`
    - `ten_crop`
    - `five_crop`

### DIFF
--- a/torchvision/prototype/transforms/kernels/_color.py
+++ b/torchvision/prototype/transforms/kernels/_color.py
@@ -1,14 +1,14 @@
-from torchvision.transforms import functional as _F
+from torchvision.transforms import functional_tensor as _FT
 
 
-adjust_brightness_image = _F.adjust_brightness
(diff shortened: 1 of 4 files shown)


datumbox:
As discussed offline with @pmeier we will close this PR and move it to the changes directly to #5421 so that everything is in one place.

PR #5451, opened 2022-02-21
Update hardcoded WIN-CUDA from 11.1 with 11.3

malfet:
Also, remove reference to conda-forge, all CUDA toolchain should be
available from NVIDIA channel

Install h5py from pip on Windows and skip failing gaussian_blur tests if Win+CUDA11.3 is used

PR #5467, opened 2022-02-24
Update hardcoded WIN-CUDA from 11.1 with 11.3 (#5451)

jdsgomes:
Also, remove reference to conda-forge, all CUDA toolchain should be
available from NVIDIA channel

Install h5py from pip on Windows and skip failing gaussian_blur tests if Win+CUDA11.3 is used


datumbox (approved):
Thanks!

PR #5468, opened 2022-02-24
Use onnx function only in tracing mode

prabhat00155:
https://github.com/pytorch/vision/pull/5310 introduced extra operations that may affect performance. Adding conditional checks to only use onnx functions when in tracing mode.

### DIFF
--- a/torchvision/models/detection/_utils.py
+++ b/torchvision/models/detection/_utils.py
@@ -489,6 +489,8 @@ def _topk_min(input: Tensor, orig_kval: int, axis: int) -> Tensor:
     Returns:
         min_kval (Tensor): Appropriately selected k-value.
     """
+    if not torch.jit.is_tracing():
+        return min(orig_kval, input.size(axis))
     axis_dim_val = torch._shape_as_tensor(input)[axis].unsqueeze(0)
     min_kval = torch.min(torch.cat((torch.tensor([orig_kval], dtype=axis_dim_val.dtype), axis_dim_val), 0))
     return min_kval  # type: ignore[arg-type]
--- a/torchvision/models/detection/_utils.py
+++ b/torchvision/models/detection/_utils.py
@@ -470,11 +470,6 @@ def retrieve_out_channels(model: nn.Module, size: Tuple[int, int]) -> List[int]:
     return out_channels
 
 
-@torch.jit.unused
-def _fake_cast_onnx(v: Tensor) -> int:
-    return v
-
-
 def _topk_min(input: Tensor, orig_kval: int, axis: int) -> int:
     """
     ONNX spec requires the k-value to be less than or equal to the number of inputs along
@@ -498,4 +493,4 @@ def _topk_min(input: Tensor, orig_kval: int, axis: int) -> int:
         return min(orig_kval, input.size(axis))
     axis_dim_val = torch._shape_as_tensor(input)[axis].unsqueeze(0)
     min_kval = torch.min(torch.cat((torch.tensor([orig_kval], dtype=axis_dim_val.dtype), axis_dim_val), 0))
-    return _fake_cast_onnx(min_kval)
+    return min_kval  # type: ignore[arg-type]
--- a/torchvision/models/detection/_utils.py
+++ b/torchvision/models/detection/_utils.py
@@ -487,7 +487,7 @@ def _topk_min(input: Tensor, orig_kval: int, axis: int) -> int:
         axis(int): Axis along which we retreive the input size.
 
     Returns:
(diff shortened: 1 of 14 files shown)


torchvision/models/detection/fcos.py
```diff
@@ -501,7 +501,11 @@ def postprocess_detections(
                 topk_idxs = torch.where(keep_idxs)[0]
 
                 # keep only topk scoring predictions
-                num_topk = det_utils._topk_min(topk_idxs, self.topk_candidates, 0)
+                num_topk = (
+                    det_utils._topk_min(topk_idxs, self.topk_candidates, 0)
+                    if torchvision._is_tracing()
```
datumbox, line 506: Should be `torch.jit.is_tracing()`

torchvision/models/detection/fcos.py
```diff
@@ -501,7 +501,11 @@ def postprocess_detections(
                 topk_idxs = torch.where(keep_idxs)[0]
 
                 # keep only topk scoring predictions
-                num_topk = det_utils._topk_min(topk_idxs, self.topk_candidates, 0)
+                num_topk = (
+                    det_utils._topk_min(topk_idxs, self.topk_candidates, 0)
```
datumbox, line 505: Nit: might be worth moving this if statement inside the _topk_min method to avoid having to patch all methods again. Thoughts?

prabhat00155: Good point.

torchvision/models/detection/_utils.py
```diff
@@ -489,6 +489,8 @@ def _topk_min(input: Tensor, orig_kval: int, axis: int) -> Tensor:
     Returns:
         min_kval (Tensor): Appropriately selected k-value.
     """
+    if not torch.jit.is_tracing():
+        return min(orig_kval, input.size(axis))
```
datumbox, line 493: You probably need a `# type: ignore[arg-type]` here and potentially you don't need one on line 496? worth a try 🤷‍♂️

prabhat00155: Or I could use `Union[int, Tensor]` as return type.

datumbox: Nah JIT won't let you do cool things like this. Try lying to it.

zhiqwang:
Hi @datumbox and @prabhat00155 , do you think that we should also add a unit-test for SSD and fcos when exporting to ONNX, as I know, the downstream [TVM](https://github.com/apache/tvm/) also use the JIT tracing mode.

zhiqwang:
Hi @datumbox 

> If you have the capacity to do this on a follow up PR that would be lovely.

I can try to add it! Recently I also found that TensorRT 8.2+ has very good support for ONNX. It will become very easy for one stage models (SSD / RetineNet / Fcos) to be deployed on TensorRT through ONNX.

torchvision/models/detection/_utils.py
```diff
@@ -470,7 +470,12 @@ def retrieve_out_channels(model: nn.Module, size: Tuple[int, int]) -> List[int]:
     return out_channels
 
 
-def _topk_min(input: Tensor, orig_kval: int, axis: int) -> Tensor:
+@torch.jit.unused
+def _fake_cast_onnx(v: Tensor) -> int:
+    return v
```
prabhat00155, line 475: mypy error:
```
torchvision/models/detection/_utils.py:475: error: Incompatible return value
type (got "Tensor", expected "int")  [return-value]
        return v
```

datumbox: you need to turn off mypy. This proposed patch is taken from the following file:
https://github.com/pytorch/vision/blob/93c85bbcc31f8d5a052daf06f2f91f39697af1a4/torchvision/models/detection/transform.py#L19-L22

And it is not being scanned by mypy:
https://github.com/pytorch/vision/blob/93c85bbcc31f8d5a052daf06f2f91f39697af1a4/mypy.ini#L41-L43

If you turn off the warnings, it should work.

torchvision/models/detection/_utils.py
```diff
@@ -489,6 +489,8 @@ def _topk_min(input: Tensor, orig_kval: int, axis: int) -> Tensor:
     Returns:
         min_kval (Tensor): Appropriately selected k-value.
```
shubhambhokare1, line 490: nit: modify function description to reflect the signature
```suggestion
        min_kval (int): Appropriately selected k-value.
```

shubhambhokare1 (changes requested):

datumbox (approved):

PR #5470, opened 2022-02-24
Removing Win CU11.1 support

datumbox:
The cu111 windows binary has an issue. We are missing cuda-toolkit which was not published by nvidia on the official channel (it only exists in conda-forge). After speaking with @atalman offline he recommended to disable temporarily the job and bring this issue to Nvidia on the next meeting with them.


atalman (approved):
LGTM

PR #5472, opened 2022-02-24
Revert exposing convNormActivation in ops

jdsgomes:
And remove references in documentation

This change will be cherrypicked for the v12.0 release

### DIFF
--- a/torchvision/models/optical_flow/raft.py
+++ b/torchvision/models/optical_flow/raft.py
@@ -6,7 +6,7 @@ import torch.nn.functional as F
 from torch import Tensor
 from torch.nn.modules.batchnorm import BatchNorm2d
 from torch.nn.modules.instancenorm import InstanceNorm2d
-from torchvision.ops import ConvNormActivation
+from torchvision.ops.misc import ConvNormActivation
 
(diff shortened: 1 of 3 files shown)


torchvision/models/optical_flow/raft.py
```diff
@@ -6,7 +6,7 @@
 from torch import Tensor
 from torch.nn.modules.batchnorm import BatchNorm2d
 from torch.nn.modules.instancenorm import InstanceNorm2d
-from torchvision.ops import ConvNormActivation
+from torchvision.ops.misc import ConvNormActivation
```
jdsgomes, line 9: @NicolasHug I added you are reviewer as this change touches the RAFT model and is meant to be cherrypicked in the v12.0 release, so I want to double check with you if you spot any problems with this?

For context on this discussion you can have a look at https://github.com/pytorch/vision/pull/5445

datumbox: I think this should be fine and we will literally undo this change immediately after introducing the Conv2dNormActivation

datumbox (approved):
LGTM!

PR #5473, opened 2022-02-24
migrate coco prototype

pmeier:


torchvision/prototype/datasets/_builtin/imagenet.py
```diff
@@ -54,7 +54,11 @@ class ImageNetDemux(enum.IntEnum):
 
 @register_dataset(NAME)
 class ImageNet(Dataset2):
-    def __init__(self, root: Union[str, pathlib.Path], *, split: str = "train") -> None:
+    """
```
pmeier, line 57: Small fix to ImageNet that I found while working on COCO.

torchvision/prototype/datasets/utils/_dataset.py
```diff
@@ -196,7 +196,18 @@ def _verify_str_arg(
...
+    def __init__(
+        self, root: Union[str, pathlib.Path], *, skip_integrity_check: bool = False, dependencies: Collection[str] = ()
+    ) -> None:
+        for dependency in dependencies:
+            try:
+                importlib.import_module(dependency)
+            except ModuleNotFoundError as error:
+                raise ModuleNotFoundError(
+                    f"{type(self).__name__}() depends on the third-party package '{dependency}'. "
+                    f"Please install it, for example with `pip install {dependency}`."
```
NicolasHug, line 208: Are we sure all packages are available on PyPI? And are we sure these will be compatible with the user's setup? I'd rather not give explicit instrutions here

torchvision/prototype/datasets/utils/_dataset.py
```diff
@@ -196,7 +196,18 @@ def _verify_str_arg(
...
+                ) from error
```
NicolasHug, line 209: from None or from error?

torchvision/prototype/datasets/_builtin/coco.py
```diff
@@ -32,27 +31,49 @@
...
+@register_info(NAME)
+def _info() -> Dict[str, Any]:
+    categories, super_categories = zip(*DatasetInfo.read_categories_file(BUILTIN_DIR / f"{NAME}.categories"))
+    return dict(categories=categories, super_categories=super_categories)
+
+
+@register_dataset(NAME)
+class Coco(Dataset2):
+    """
+    - **homepage**: https://cocodataset.org/
```
NicolasHug, line 50: I would suggest to add the dependencies in the docstring as well

NicolasHug (approved):
Minor comments but LGTM, thanks @pmeier

torchvision/prototype/datasets/utils/_dataset.py
pmeier: > Are we sure all packages are available on PyPI?

I think so yes. I don't think we can depend on packages that are not hosted on PyPI for our public datasets.

> And are we sure these will be compatible with the user's setup? I'd rather not give explicit instrutions here

No, we can't be sure. Given that wheels on PyPI are self-contained, I'm guessing we should be fine. The intention here is to help new users to get started. If we can expect them to handle the installation themselves, I'm ok to remove the instructions.

torchvision/prototype/datasets/utils/_dataset.py
pmeier: `from None` is better given that the original error adds no information.

torchvision/prototype/datasets/utils/_dataset.py
NicolasHug: Ok let's leave it as is, we can always revisit

NicolasHug:
Tests are green locally, CI is toasted but as discussed offline we'll still merge these migrating PR against the `prototype-datasets-inheritance` branch in order to move faster with this. We'll make sure everything is green on the CI before merging on `main`.

Thanks @pmeier

PR #5475, opened 2022-02-24
Add barriers after init_process_group() for all recipes

datumbox:
Currently only the detection recipe supports this:
https://github.com/pytorch/vision/blob/d367a01a18a3ae6bee13d8be3b63fd6a581ea46f/references/detection/utils.py#L278-L281

Adding barriers to all other recipes.


mrshenli (approved):
LGTM

PR #5480, opened 2022-02-25
migrate mnist prototype datasets

pmeier:


test/builtin_dataset_mocks.py
```diff
@@ -214,70 +214,76 @@ def generate(
...
     )
 
 
-# DATASET_MOCKS.update({name: DatasetMock(name, mnist) for name in ["fashionmnist", "kmnist"]})
+DATASET_MOCKS.update(
+    {
+        name: DatasetMock(name, mock_data_fn=mnist, configs=combinations_grid(split=("train", "test")))
+        for name in ["mnist", "fashionmnist", "kmnist"]
+    }
+)
```
NicolasHug, line 232: IIUC the only thing that changes is the `name` here. Is this going to have an effect on the actual tests that are being run, or are we running the same test 3 times? If the latter, perhaps we could register the callable with the regular `regsiter_mock` decorator, and just add as a comment that this also applies to fashionmnist and kmnist?

test/builtin_dataset_mocks.py
```diff
@@ -214,70 +214,76 @@ def generate(
...
+def emnist(root, config):
     num_samples_map = {}
     file_names = set()
-    for config_ in info._configs:
-        prefix = f"emnist-{config_.image_set.replace('_', '').lower()}-{config_.split}"
+    for split, image_set in itertools.product(
+        ("train", "test"),
+        ("Balanced", "By_Merge", "By_Class", "Letters", "Digits", "MNIST"),
+    ):
```
NicolasHug, line 247: No strong opinion but what was the reason for changing this?

torchvision/prototype/datasets/_builtin/mnist.py
```diff
@@ -340,3 +404,12 @@ def _prepare_sample(self, data: Tuple[torch.Tensor, torch.Tensor], *, config: Da
...
         sample.update(dict(zip(("duplicate", "unused"), [bool(value) for value in extra_anns[-2:]])))
         return sample
+
+    def __len__(self) -> int:
+        return {
+            "train": 60_000,
+            "test": 60_000,
+            "test10k": 10_000,
+            "test50k": 50_000,
+            "nist": 402953,
```
NicolasHug, line 414: Nit

```suggestion
            "nist": 402_953,
```

NicolasHug (approved):
Thanks @pmeier , some minor Q but LGTM

test/builtin_dataset_mocks.py
pmeier: > IIUC the only thing that changes is the `name` here.

Correct.

> Is this going to have an effect on the actual tests that are being run, or are we running the same test 3 times?

We are running tests for all three datasets with the "same" mock data. If we didn't, our tests wouldn't cover the other datasets. Admittedly, the current implementation is just FashionMNIST and `KMNIST` subclassing `MNIST` and minimal parametrization, but I don't think we take much of a duration hit to include these. If that is a concern, there are other datasets that we should optimize first.

test/builtin_dataset_mocks.py
pmeier: We no longer have access to a `DatasetInfo` that holds all the `._configs`. Thus, we need to create them manually.

PR #5481, opened 2022-02-25
merge prototype dataset register decorators

pmeier:
Given that the info is static anyway, there is no reason to have them separate other than having access to it from within the dataset class. We can achieve the same by letting the register decorator set a private class variable.

### DIFF
--- a/test/test_prototype_builtin_datasets.py
+++ b/test/test_prototype_builtin_datasets.py
@@ -34,16 +34,6 @@ def test_coverage():
 
 
 class TestCommon:
-    @pytest.mark.parametrize("name", datasets.list_datasets())
-    def test_info(self, name):
-        try:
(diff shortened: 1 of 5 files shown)


NicolasHug:
Thanks @pmeier 

Is this solving a problem we're currently having? Otherwise I'm happy with the dumb simple registration we already have

pmeier:
> Is this solving a problem we're currently having?

It does not solve anything, it simplifies stuff. With the current approach we have two separate decorators that always need to be used together with the same parameters. We also have a test that makes sure that this happening. This PR removes all of that while only being minimally more complex.

NicolasHug:
The main difference is that it's making _INFO purely static and closes the door for having parameter-dependent infos. I'd rather leave that door open until the very end where we'll have a clearer picture.

PR #5488, opened 2022-02-26
Fix append of audio_pts

satojkovic:
Modified to append to audio_pts for audio stream processing block.


prabhat00155 (approved):
Thanks @satojkovic for fixing this.

Issue #1884, opened 2020-02-13
VideoClips Assertion Error


mjunyent:
Hello,

I'm trying to load a big video. Following https://github.com/pytorch/vision/issues/1446 I used a VideoClips object, but it's crashing when trying to get clips with certain ids with this error: 

> AssertionError                            Traceback (most recent call last)
> <ipython-input-9-6e97949ad7f5> in <module>()
> ----> 1 x = video_clips.get_clip(1)
> 
> /usr/local/lib/python3.6/dist-packages/torchvision/datasets/video_utils.py in get_clip(self, idx)
>     324             video = video[resampling_idx]
>     325             info["video_fps"] = self.frame_rate
> --> 326         assert len(video) == self.num_frames, "{} x {}".format(video.shape, self.num_frames)
>     327         return video, audio, info, video_idx
> 
> AssertionError: torch.Size([0, 1, 1, 3]) x 32

The code I use is just this:
```
from torchvision.datasets.video_utils import VideoClips
video_clips = VideoClips(["test_video.mp4"], clip_length_in_frames=32, frames_between_clips=32)
for i in range(video_clips.num_clips()):
    x = video_clips.get_clip(i)
```

`video_clips.num_clips()` is much bigger than the ids that are failing. Changing the clipt_length or frames_between doesn't help.

Checking the code I see [0,1,1,3] is returned by `read_video` when no vframes are read:
https://github.com/pytorch/vision/blob/85b8fbfd31e9324e64e24ca25410284ef238bcb3/torchvision/io/video.py#L251-L254
But, for some clip ids and clip_lengths it's just that the sizes don't match, as the assertion error is something like this `AssertionError: torch.Size([19, 360, 640, 3]) x 128`

I followed the issue to `_read_from_stream` and checked no AV exceptions where raised. And running this part of the function:
https://github.com/pytorch/vision/blob/85b8fbfd31e9324e64e24ca25410284ef238bcb3/torchvision/io/video.py#L144-L150
I saw that for an `start_pts=32032`, `end_pts=63063` it returned just one frame on `frames` with `pts=237237`. Which is later discarted as it's a lot bigger than `end_pts`.

Also, the `stream.time_base` is `Fraction(1, 24000)` which doesn't match the start and end pts provided by VideoClips.

So it seems there is a problem with the seeking on my video. But it has a standard h264 encoding and I have no problem reading it sequentially with pyav.
[...]

PR #5489, opened 2022-02-26
Fix shape mismatch error

prabhat00155:
Resolves https://github.com/pytorch/vision/issues/1884.


bjuncek (approved):

bjuncek:
@datumbox LGTM, this partially resolves kinetics issue

datumbox:
@prabhat00155 @bjuncek Thanks! Have we tested the patch and confirmed it works for training? 

What does partially means here?

prabhat00155:
@bjuncek Were you able to test this on your setup?

bjuncek:
So on my system that resolves dataset generation issue, but probably does not resolve accuracy issue (I can't test this due to only having one gpu).

Issue #5410, opened 2022-02-11
[RFC] Batteries Included - Phase 2


datumbox:
### 🚀 The feature

*Note: To track the progress of the project check out [this board](https://github.com/pytorch/vision/projects/4).*

This is the 2nd phase of TorchVision's modernization project (see [phase 1](https://github.com/pytorch/vision/issues/3911)). We aim to keep TorchVision relevant by ensuring it provides off-the-shelf all the necessary primitives, model architectures and recipe utilities to produce SOTA results for the supported  Computer Vision tasks. 

<h2 id="new-primitives">1. New Primitives</h2>

To enable our users to reproduce the latest state-of-the-art research we will enhance TorchVision with the following data augmentations, layers, losses and other operators:

### Data Augmentations

- [x] [Augmix](https://arxiv.org/abs/1912.02781) - #5411
- [x] [Large Scale Jitter](https://arxiv.org/pdf/2012.07177.pdf) - #5435 #5446 #5559
- [x] Fixed Size Crop - #5607
- [x] Random Shortest Size - #5610
- [x] [Simple CopyPaste](https://arxiv.org/pdf/2012.07177v2.pdf) - #5825

### Layers

- [x] [DropBlock](https://arxiv.org/abs/1810.12890) - #5416 
- [x] [Conv3DNormActivation](https://github.com/pytorch/vision/issues/5430) - #5445
- [x] [MLP](https://github.com/pytorch/vision/issues/4333) - #6053
- [x] Permute - #6055

### Losses
- [x] [Generalized-IoU loss](https://arxiv.org/pdf/1902.09630v2.pdf) - #4961
- [x] [Distance-IoU & Complete-IoU loss](https://arxiv.org/abs/1911.08287) - #5776 #5786 #5984

### Operators added in PyTorch Core

- [x] Better [EMA](https://github.com/pytorch/pytorch/issues/66686) support in `AveragedModel` - https://github.com/pytorch/pytorch/pull/71763
- [x] [Add support of empty output in SyncBatchNorm](https://github.com/pytorch/pytorch/issues/36530) - https://github.com/pytorch/pytorch/pull/74944

<h2 id="new-models">2. New Architectures & Model Iterations</h2>

To ensure that our users have access to the most popular SOTA models, we will add the following architectures along with pre-trained weights. Moreover we will improve existing architectures with commonly adopted optimizations introduced in follow up research:

### Image Classification

[...]

PR #5491, opened 2022-02-27
Adding Swin Transformer architecture

xiaohu2015:
This work related to https://github.com/pytorch/vision/issues/2707 and https://github.com/pytorch/vision/issues/5410: add swin transformer to torchvision model_zoo.

- -[x] Implement swin_transformer.
- -[x] Check model accuracy from converted weights [swin_tiny_weights](https://drive.google.com/file/d/1933Vroue-VyJ7KU6ZLnfUpxpjgRj7i0R/view?usp=sharing).

```
torchrun --nproc_per_node=8 train.py --model swin_tiny --interpolation bicubic --test-only --pretrained
```
- -[x] Refactor code.
I made some modifications compared to official code: 
- **remove `absolute position embedding`**:  as we can see from table 4 in the paper, the swin model with `relative position bias` get  best results, so the default swin model does not use `absolute position embedding`. Another trouble with `absolute position embedding` is that we have to set input_size to the model to initialize the pos_embedding.
- **remove the `input_resolution` parameter**: so input with arbitrary shape can be handled by the swint model, which is necessary for some tasks eg. segmentaion and object detection. Compared to offical code, we keep tensor with shape [B, H, W, C] instead of [B, N, C], so we can get width and height without `input_resolution`. But after do that, one must dynamically compare the window size and input size in the shifted window attention, for example, if the input size is lower than window size (when the image size is 224, the feature size of last stage is 7x7), you need't do shift operation. but the dynamic behavior is not well supported in torch.fx, so I create `shifted_window_attention` function and warp it. Note: this modification can add run time as we have to generate `attention_mask` dynamically, but the cost time is insignificant.

- -[x] Validate the training.

```
torchrun --nproc_per_node=8 train.py\
    --model swin_t --epochs 300 --batch-size 128 --opt adamw --lr 0.001 --weight-decay 0.05 --norm-weight-decay 0.0\
    --bias-weight-decay 0.0 --transformer-embedding-decay 0.0 --lr-scheduler cosineannealinglr --lr-min 0.00001 --lr-warmup-method linear\
    --lr-warmup-epochs 20 --lr-warmup-decay 0.01 --amp --label-smoothing 0.1 --mixup-alpha 0.8\
    --clip-grad-norm 5.0 --cutmix-alpha 1.0 --random-erase 0.25 --interpolation bicubic --auto-augment ra
```
which can give result: `Acc@1 81.222 Acc@5 95.332` [train logs](https://drive.google.com/file/d/1IMdSEVBY3vXt4PSOk6XOlk4j3KH3-X7B/view?usp=sharing)

I also modified the reference code, https://github.com/xiaohu2015/vision/blob/main/references/classification/utils.py#L406. as the current code only supports no weight decay for norm layers.

### references
- [Swin Transformer: Hierarchical Vision Transformer using Shifted Windows](https://arxiv.org/abs/2103.14030)
- https://github.com/microsoft/Swin-Transformer
- https://github.com/rwightman/pytorch-image-models
- https://github.com/open-mmlab/mmclassification/tree/master/configs/swin_transformer


datumbox:
@xiaohu2015 Concerning the [failing tests](https://app.circleci.com/pipelines/github/pytorch/vision/15504/workflows/4596012b-603d-4f84-a8ab-fb2e96ed1556/jobs/1252880), you can update the FX test to use sums instead of means OR you can just cast the outcome to float. It shouldn't matter too much.

Also I see that you haven't pushed the expected files for the models in your PR. You generate them by running the tests once by having an env var `EXPECTTEST_ACCEPT=1`

jdsgomes:
Hi [xiaohu2015](https://github.com/xiaohu2015), thank you for the submission! I am happy to assist with the review once the PR is ready.

xiaohu2015:
> Hi [xiaohu2015](https://github.com/xiaohu2015), thank you for the submission! I am happy to assist with the review once the PR is ready.

OK, thank you.

xiaohu2015:
> @xiaohu2015 Concerning the [failing tests](https://app.circleci.com/pipelines/github/pytorch/vision/15504/workflows/4596012b-603d-4f84-a8ab-fb2e96ed1556/jobs/1252880), you can update the FX test to use sums instead of means OR you can just cast the outcome to float. It shouldn't matter too much.
> 
> Also I see that you haven't pushed the expected files for the models in your PR. You generate them by running the tests once by having an env var `EXPECTTEST_ACCEPT=1`.
> 
> Finally, I see on the PR description that you still want to do a refactoring. Shall we mark the PR as draft until you are ready to kick off the review? This way we won't miss it when you are ready for feedback.

thanks, the only error is https://app.circleci.com/pipelines/github/pytorch/vision/15559/workflows/adceaf96-b697-4dce-8d61-327477cbc516/jobs/1257523.

torchvision/models/swin_transformer.py
```diff
@@ -372,14 +372,16 @@ def __init__(
...
-            if isinstance(m, (nn.Conv2d, nn.Linear)):
-                nn.init.trunc_normal_(m.weight, std=0.02)
-                if m.bias is not None:
-                    nn.init.constant_(m.bias, 0)
-            elif isinstance(m, nn.LayerNorm):
-                nn.init.constant_(m.bias, 0)
-                nn.init.constant_(m.weight, 1.0)
+        self.apply(self._init_weights)
+
+    def _init_weights(self, m):
```
datumbox, line 377: I recommend not to use this idiom as it can lead to unexpected results if the model gets FXed. The model initialization should happen on the constructor until there is proper support for model reinitialization from the nn.Module.

jdsgomes:
Thank you for updating the PR. As per our offline discussions I understand you have retrained the model and I can see that the gap has been closed. I will prioritise this review so you should expect to get feedback this week. 
Once again thanks for the contribution, it is looking great!

references/classification/utils.py
```diff
@@ -400,3 +401,57 @@ def reduce_across_processes(val):
...
+
+
+def set_weight_decay(
+    model: torch.nn.Module,
+    weight_decay: float,
+    norm_weight_decay: Optional[float] = None,
+    bias_weight_decay: Optional[float] = None,
+    custom_keys_weight_decay: Optional[Dict[str, float]] = None,
+):
+    norm_classes = (torch.nn.modules.batchnorm._BatchNorm, torch.nn.LayerNorm, torch.nn.GroupNorm)
```
datumbox, line 413: Might be worth adding `nn.modules.instancenorm._InstanceNorm` and `nn.LocalResponseNorm`. I've got the same update on my detection recipe [PR](https://github.com/pytorch/vision/pull/5444/files#diff-250a5139d0d5b18ff3513442f13dcb3760c9dfb1a8eeea2ecc3bf0f129dd220cR51).

references/classification/utils.py
```diff
@@ -400,3 +401,57 @@ def reduce_across_processes(val):
...
     return t
+
+
+def set_weight_decay(
+    model: torch.nn.Module,
+    weight_decay: float,
+    norm_weight_decay: Optional[float] = None,
+    bias_weight_decay: Optional[float] = None,
+    custom_keys_weight_decay: Optional[Dict[str, float]] = None,
+):
```
datumbox, line 412: We should add documentation on this and provide examples for using it on `custom_keys_weight_decay`. From our offline chats I understand that this could be useful if you want to exclude arbitrary things (like positional embeddings) but we should clarify how/why here.

references/classification/utils.py
```diff
@@ -400,3 +401,57 @@ def reduce_across_processes(val):
...
+            custom_params[key] = []
+
+    for module in model.modules():
+        if next(module.children(), None):
+            for name, p in module.named_parameters(recurse=False):
+                if not p.requires_grad:
+                    continue
+                is_custom_key = False
+                for key in custom_params:
+                    if key in name:
```
datumbox, line 430: I think we would need to discuss pros/cons of what the API expects the custom_keys to be. I understand you are looking for substring here. Why not require the user to directly define the full names of the param they want to exclude? Looks clearner; thoughts? 

Also it's worth checking if we can simplify the logic here.

references/classification/utils.py
```diff
@@ -400,3 +401,57 @@ def reduce_across_processes(val):
...
+                        custom_params[key].append(p)
+                        is_custom_key = True
+                if not is_custom_key:
+                    other_params.append(p)
+        elif isinstance(module, norm_classes):
+            if norm_weight_decay is not None:
+                norm_params.extend(p for p in module.parameters() if p.requires_grad)
+            else:
+                other_params.extend(p for p in module.parameters() if p.requires_grad)
```
datumbox, line 439: Consider refactoring to avoid duplication of code.

references/classification/utils.py
```diff
@@ -400,3 +401,57 @@ def reduce_across_processes(val):
...
+
+    param_groups = []
+    if norm_weight_decay is not None:
+        param_groups.append({"params": norm_params, "weight_decay": norm_weight_decay})
+    if bias_weight_decay is not None:
+        param_groups.append({"params": bias_params, "weight_decay": bias_weight_decay})
+    for key in custom_params:
+        param_groups.append({"params": custom_params[key], "weight_decay": custom_keys_weight_decay[key]})
+    param_groups.append({"params": other_params, "weight_decay": weight_decay})
+    return param_groups
```
datumbox, line 457: Overall, I think this could be a useful utility which can replace the `split_normalization_params` method in ops. I think we should consider refactoring/simplifying and moving it out on a separate PR. It would be a useful addition in the ops though I would still keep it private until we are certain that the API is generic enough.

torchvision/models/swin_transformer.py
```diff
@@ -0,0 +1,503 @@
...
+    )
+    if pretrained:
+        if arch not in _MODELS_URLS:
+            raise ValueError(f"No checkpoint is available for model type {arch}")
+        state_dict = load_state_dict_from_url(_MODELS_URLS[arch], progress=progress)
+        model.load_state_dict(state_dict)
+    return model
+
+
+def swin_tiny(pretrained: bool = False, progress: bool = True, **kwargs: Any) -> SwinTransformer:
```
datumbox, line 422: The multiweight support API landed yesterday on main area (see #5618). We should remove these old-style model builders for the new ones.

torchvision/models/swin_transformer.py
```diff
@@ -0,0 +1,503 @@
...
+        num_heads=[3, 6, 12, 24],
+        window_size=7,
+        stochastic_depth_prob=0.2,
+        pretrained=pretrained,
+        progress=progress,
+        **kwargs,
+    )
+
+
+def swin_samll(pretrained: bool = False, progress: bool = True, **kwargs: Any) -> SwinTransformer:
```
datumbox, line 443:
```suggestion
def swin_small(pretrained: bool = False, progress: bool = True, **kwargs: Any) -> SwinTransformer:
```

datumbox (commented):
Top notch job as always @xiaohu2015. Thanks for the contribution.

I think some of the optimizations you do on this PR are worth moving on separate PRs. I've added a few comments below. I didn't review the Swin architecture as this is something that @jdsgomes will do.

torchvision/models/swin_transformer.py
```diff
@@ -0,0 +1,503 @@
...
+class SwinTransformer(nn.Module):
+    """
+    Implements Swin Transformer from the `"Swin Transformer: Hierarchical Vision Transformer using
+    Shifted Windows" <https://arxiv.org/pdf/2103.14030>`_ paper.
+
+    Args:
+        patch_size (int): Patch size. Default: 4.
+        num_classes (int): Number of classes for classification head. Default: 1000.
+        embed_dim (int): Patch embedding dimension. Default: 96.
+        depths (List(int)): Depth of each Swin Transformer layer.
```
jdsgomes, line 286: We could add depths and numheads default values here

torchvision/models/swin_transformer.py
```diff
@@ -0,0 +1,503 @@
...
+        num_heads (List(int)): Number of attention heads in different layers.
+        window_size (int): Window size. Default: 7.
+        mlp_ratio (float): Ratio of mlp hidden dim to embedding dim. Default: 4.
+        dropout (float): Dropout rate. Default: 0.
+        attention_drop_rate (float): Attention dropout rate. Default: 0.
+        drop_path_rate (float): Stochastic depth rate. Default: 0.1.
```
jdsgomes, line 292: parameter bane and default value do not match the constructor

torchvision/models/swin_transformer.py
```diff
@@ -0,0 +1,503 @@
...
+        block (nn.Module): SwinTransformer Block.
```
jdsgomes, line 293: please mention these two parameters are optional

torchvision/models/swin_transformer.py
```diff
@@ -0,0 +1,503 @@
...
+
+class SwinTransformerBlock(nn.Module):
+    """
+    Swin Transformer Block.
+
+    Args:
+        dim (int): Number of input channels.
+        num_heads (int): Number of attention heads.
+        window_size (int): Window size.
+        shift_size (int): Shift size for SW-MSA.
```
jdsgomes, line 234: please document the missing default values here

torchvision/models/swin_transformer.py
```diff
@@ -0,0 +1,503 @@
...
+        qkv_weight (Tensor[in_dim, out_dim]): The weight tensor of query, key, value.
+        proj_weight (Tensor[out_dim, out_dim]): The weight tensor of projection.
+        relative_position_bias (Tensor): The learned relative position bias added to attention.
+        window_size (int)): The size of the window.
+        shift_size (int): Shift size for SW-MSA.
+        num_heads (int): Number of attention heads.
+        attention_dropout (float): Dropout ratio of attention weight. Default: 0.0.
+        dropout (float): Dropout ratio of output. Default: 0.0.
+        qkv_weight (Tensor[out_dim]): The bias tensor of query, key, value. Default: None.
+        proj_weight (Tensor[out_dim]): The bias tensor of projection. Default: None.
```
jdsgomes, line 84:
```suggestion
        proj_bias (Tensor[out_dim], optional): The bias tensor of projection. Default: None.
```

references/classification/utils.py
jdsgomes: I agree with this, it could be dangerous to just require a substring, but I might be missing something.

references/classification/utils.py
xiaohu2015: @datumbox I think that we should improve the setting of `weight decay` to support more. what do you think?

datumbox: I think we agree but to clarify what I meant: I see lots of value in making a more generic util that allows us setting custom weigh decay for biases, norm weights and arbitrary params. I think your implementation is a great starting point for this and that's why I've added some comments to refactor it and potentially move it potentially on a separate PR that is going to be merged first on main. Is your understanding the same?

references/classification/utils.py
xiaohu2015: you are right, in fact we can use `==` instead of  `in`, because the `name` here means the child name (not include the parent prefix).

references/classification/utils.py
xiaohu2015: @datumbox yes, I argee that we should open anthor PR to add a more generic  to replace `split_normalization_params`.

datumbox: Agreed. Just remove the method from this PR, move it to another one and incorporate the comments we added already here. We will then review the other PR and merge it first before we merge this one. :)

torchvision/models/swin_transformer.py
```diff
@@ -0,0 +1,503 @@
...
+        relative_coords[:, :, 0] += self.window_size - 1  # shift to start from 0
+        relative_coords[:, :, 1] += self.window_size - 1
+        relative_coords[:, :, 0] *= 2 * self.window_size - 1
+        relative_position_index = relative_coords.sum(-1).view(-1)  # Wh*Ww*Wh*Ww
+        self.register_buffer("relative_position_index", relative_position_index)
+
+        nn.init.trunc_normal_(self.relative_position_bias_table, std=0.02)
+
+    def forward(self, x: Tensor):
+        relative_position_bias = self.relative_position_bias_table[self.relative_position_index].view(
```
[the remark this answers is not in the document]
xiaohu2015: yes, but  here we define relative_position_bias as a parameter with shape (2M−1)×(2M−1) rather than M^2xM^2 to reduce reduce the number of parameters as discussed in paper page5, so we need do the conversion in the `forward` function.

jdsgomes:
I will wait until https://github.com/pytorch/vision/pull/5671 gets merged and the changes applied in this PR to replicate the results and train the other variants to make sure we train but also test the final version of the PR.

datumbox:
This PR also requires the model builder update noted at https://github.com/pytorch/vision/pull/5491#discussion_r833046233.

Other than that, @xiaohu2015 what is your preference here? Would you like to try to find a generic solution for #5671 or opt for adding this util in references (as per your original plan) to unblock this PR? There is no rush on our side, but we should to it in a way that works/makes it easier for you.

xiaohu2015:
Ok, I will finish this by this sunday

torchvision/models/swin_transformer.py
```diff
@@ -411,30 +395,58 @@ def _swin_transformer(
...
     return model
 
 
-def swin_tiny(pretrained: bool = False, progress: bool = True, **kwargs: Any) -> SwinTransformer:
+_COMMON_META = {
+    "task": "image_classification",
+    "architecture": "SwinTransformer",
+    "publication_year": 2021,
+    "categories": _IMAGENET_CATEGORIES,
+    "interpolation": InterpolationMode.BICUBIC,
```
datumbox, line 409: Did we indeed train it with Bicubic? IS this the pretrained model from the paper?

torchvision/models/swin_transformer.py
```diff
@@ -411,30 +395,58 @@ def _swin_transformer(
...
+            **_COMMON_META,
+            "num_params": 86567656,
+            "size": (224, 224),
+            "min_size": (224, 224),
+            "recipe": "https://github.com/pytorch/vision/tree/main/references/classification#swin_tiny",
+            "acc@1": 81.072,
+            "acc@5": 95.318,
+        },
+    )
+    DEFAULT = IMAGENET1K_V1
```
datumbox, line 427: These are dummy weights from ViT. To avoid issues I recommend removing this completely and just having pass here. There is an example on how to handle models with no weights [here](https://github.com/pytorch/vision/blob/main/torchvision/models/shufflenetv2.py#L230)

torchvision/models/swin_transformer.py
```diff
@@ -411,30 +395,58 @@ def _swin_transformer(
...
+
+
+@handle_legacy_interface(weights=("pretrained", Swin_Tiny_Weights.IMAGENET1K_V1))
```
datumbox, line 430: I think we can completely drop this. There is no reason to offer legacy support, this is a brand new model:
```suggestion
```

torchvision/models/swin_transformer.py
```diff
@@ -411,30 +395,58 @@ def _swin_transformer(
...
+def swin_tiny(*, weights: Optional[Swin_Tiny_Weights] = None, progress: bool = True, **kwargs: Any) -> SwinTransformer:
     """
     Constructs a swin_tiny architecture from
     `"Swin Transformer: Hierarchical Vision Transformer using Shifted Windows" <https://arxiv.org/pdf/2103.14030>`_.
     Args:
         pretrained (bool): If True, returns a model pre-trained on ImageNet
```
datumbox:
```suggestion
        weights (Swin_Tiny_Weights, optional): The pretrained weights for the model
```

torchvision/models/swin_transformer.py
```diff
@@ -411,30 +395,58 @@ def _swin_transformer(
         stochastic_depth_prob=stochastic_depth_prob,
         **kwargs,
     )
-    if pretrained:
-        if arch not in _MODELS_URLS:
-            raise ValueError(f"No checkpoint is available for model type {arch}")
-        state_dict = load_state_dict_from_url(_MODELS_URLS[arch], progress=progress)
-        model.load_state_dict(state_dict)
+
+    if weights:
```
datumbox, line 398: Optional nit for consistency with other places:
```suggestion
    if weights is not None:
```

torchvision/models/swin_transformer.py
xiaohu2015: yes, I trained with Bicubic to keep consistent with offical code.

torchvision/models/swin_transformer.py
```diff
@@ -18,6 +18,9 @@
 __all__ = [
     "SwinTransformer",
     "swin_tiny",
+    "swin_small",
+    "swin_base",
+    "swin_large",
```
datumbox, line 23: You need also to add the enums such as `Swin_Small_Weights` here.

torchvision/models/swin_transformer.py
```diff
@@ -0,0 +1,521 @@
...
+    num_windows = (pad_H // window_size) * (pad_W // window_size)
+    x = x.view(B, pad_H // window_size, window_size, pad_W // window_size, window_size, C)
+    x = x.permute(0, 1, 3, 2, 4, 5).reshape(B * num_windows, window_size * window_size, C)  # B*nW, Ws*Ws, C
+
+    # multi-head attention
+    qkv = F.linear(x, qkv_weight, qkv_bias)
+    qkv = qkv.reshape(x.size(0), x.size(1), 3, num_heads, C // num_heads).permute(2, 0, 3, 1, 4)
+    q, k, v = qkv[0], qkv[1], qkv[2]
+    q = q * (C // num_heads) ** -0.5
+    attn = q @ k.transpose(-2, -1)
```
datumbox, line 117: Let's switch to a `matmul` call instead. We don't use the `@` in TorchVision due to the fact it's hard to search on Google and integrate with IDEs.

torchvision/models/swin_transformer.py
```diff
@@ -0,0 +1,521 @@
...
+        attn_mask = attn_mask.unsqueeze(1) - attn_mask.unsqueeze(2)
+        attn_mask = attn_mask.masked_fill(attn_mask != 0, float(-100.0)).masked_fill(attn_mask == 0, float(0.0))
+        attn = attn.view(x.size(0) // num_windows, num_windows, num_heads, x.size(1), x.size(1))
+        attn = attn + attn_mask.unsqueeze(1).unsqueeze(0)
+        attn = attn.view(-1, num_heads, x.size(1), x.size(1))
+
+    attn = F.softmax(attn, dim=-1)
+    attn = F.dropout(attn, p=attention_dropout)
+
+    x = (attn @ v).transpose(1, 2).reshape(x.size(0), x.size(1), C)
```
datumbox, line 141: Same here about matmul.

datumbox (commented):
Overall it looks good thanks @xiaohu2015! I've added only minor comments. Note that to fix the [failing tests](https://app.circleci.com/pipelines/github/pytorch/vision/16324/workflows/891cbbcc-7f38-4ce9-842c-1251c1f4a47e/jobs/1323316), you just need to create expected files for all variants.

references/classification/train.py
```diff
@@ -424,6 +424,9 @@ def get_args_parser(add_help=True):
     parser.add_argument("--lr-warmup-decay", default=0.01, type=float, help="the decay for lr")
     parser.add_argument("--lr-step-size", default=30, type=int, help="decrease lr every step-size epochs")
     parser.add_argument("--lr-gamma", default=0.1, type=float, help="decrease lr by a factor of lr-gamma")
+    parser.add_argument(
+        "--lr-cos-min", default=0.0, type=float, help="minimum lr of cosine annealing schedule (default: 0.0)"
+    )
```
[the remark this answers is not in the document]
xiaohu2015: In fact, I checked the Exponential LR, it does't have a `lr_min` augment. But I agree with your advice.

xiaohu2015:
@datumbox yes, the estimation does't depend on the input, but if we move this to constructor, it will work well in inference mode, I think it maybe has a problem in computing the gradients for `relative_position_bias_table` in training mode.

datumbox:
@xiaohu2015 Hmmm... Did you try moving it on the constructor and got issues during training? Or that's something you didn't try and you are worried it might affect us?

xiaohu2015:
@datumbox since all checks have passed, can you start do some check on swin_s model: check the accuarcy of the converted weights, and validate the training recipe.


> This work related to #2707 and #5410: add swin transformer to torchvision model_zoo.
> 
> * [x]  Implement swin_transformer.
> * [x]  Check model accuracy from converted weights [swin_tiny_weights](https://drive.google.com/file/d/1933Vroue-VyJ7KU6ZLnfUpxpjgRj7i0R/view?usp=sharing).
> 
> ```
> torchrun --nproc_per_node=8 train.py --model swin_tiny --interpolation bicubic --test-only --pretrained
> ```
> 
> * [x]  Refactor code.
>   I made some modifications compared to official code:
> * **remove `absolute position embedding`**:  as we can see from table 4 in the paper, the swin model with `relative position bias` get  best results, so the default swin model does not use `absolute position embedding`. Another trouble with `absolute position embedding` is that we have to set input_size to the model to initialize the pos_embedding.
> * **remove the `input_resolution` parameter**: so input with arbitrary shape can be handled by the swint model, which is necessary for some tasks eg. segmentaion and object detection. Compared to offical code, we keep tensor with shape [B, H, W, C] instead of [B, N, C], so we can get width and height without `input_resolution`. But after do that, one must dynamically compare the window size and input size in the shifted window attention, for example, if the input size is lower than window size (when the image size is 224, the feature size of last stage is 7x7), you need't do shift operation. but the dynamic behavior is not well supported in torch.fx, so I create `shifted_window_attention` function and warp it. Note: this modification can add run time as we have to generate `attention_mask` dynamically, but the cost time is insignificant.
> * [x]  Validate the training.
> 
> ```
> torchrun --nproc_per_node=8 train.py\
>     --model swin_tiny --epochs 300 --batch-size 128 --opt adamw --lr 0.001 --weight-decay 0.05 --norm-weight-decay 0.0\
>     --bias-weight-decay 0.0 --transformer-embedding-decay 0.0 --lr-scheduler cosineannealinglr --lr-min 0.00001 --lr-warmup-method linear\
>     --lr-warmup-epochs 20 --lr-warmup-decay 0.01 --amp --label-smoothing 0.1 --mixup-alpha 0.8\
>     --clip-grad-norm 5.0 --cutmix-alpha 1.0 --random-erase 0.25 --interpolation bicubic
> ```
> 
> which can give result: `Acc@1 81.222 Acc@5 95.332` [train logs](https://drive.google.com/file/d/1IMdSEVBY3vXt4PSOk6XOlk4j3KH3-X7B/view?usp=sharing)
> 
> I also modified the reference code, https://github.com/xiaohu2015/vision/blob/main/references/classification/utils.py#L406. as the current code only supports no weight decay for norm layers.
> 
> ### references
> * [Swin Transformer: Hierarchical Vision Transformer using Shifted Windows](https://arxiv.org/abs/2103.14030)
> * https://github.com/microsoft/Swin-Transformer
> * https://github.com/rwightman/pytorch-image-models
> * https://github.com/open-mmlab/mmclassification/tree/master/configs/swin_transformer

xiaohu2015:
> @xiaohu2015 Hmmm... Did you try moving it on the constructor and got issues during training? Or that's something you didn't try and you are worried it might affect us?

I do a toy experiment, I found it works. But I am not sure if it will affect performance, maybe you can do end2end training to test it.

datumbox:
@xiaohu2015 I've checked the implementation and as far as I can see it's perfect. Now it's on us to make the necessary checks and train the remaining variants. As always thanks so much for the awesome work :)

@jdsgomes Might be worth checking your proposal of moving `relative_position_bias_table` to the constructor while retraining the tiny variant. I think it should be fine but worth confirming.

jdsgomes:
@xiaohu2015 thanks for the changes. I am now training now the tiny variant with and without the change for the `relative_position_bias_table`.

xiaohu2015:
> 

we can use the same training command to train `swin_s` and `swin_b`, but for `swin_l`, it is trained on imagenet21k

datumbox:
@xiaohu2015 Yes. Let's do the s/b variants and see later whether we will train L. It might be quite expensive to do so.

jdsgomes:
@datumbox please have a look at my latest commit. It is regarding moving the `relative_position_bias_table` to the constructor.

jdsgomes:
The training of `swin_t` before moving `relative_position_bias_table` is ongoing and so far the results seem to align nicely with the logs provided by @xiaohu2015.  Namely, after epoch 66 we get: 

`Test:  Acc@1 72.278 Acc@5 91.274 (own run id 36469)`

vs

`Test:  Acc@1 72.406 Acc@5 91.526 (@xiaohu2015 submission)`

The training of the [modified version ](https://github.com/pytorch/vision/pull/5491/commits/f0615440bf18617dc0e5dc4839bd5ed27e5ed010)(with `relative_position_bias_table` is the constructor) is also ongoing and so far (epoch 7) the results are slightly lower:

`Test:  Acc@1 39.926 Acc@5 65.688 (own run id 36960)`

vs

`Test:  Acc@1 40.964 Acc@5 66.462 (@xiaohu2015 submission)`

I have also training jobs for the `swin_s` and `swin_b` running in parallel. I will be away until 20/04/2022 so by then I hope all variants have been trained and we can confirm the final performance is in line with the submission and with the paper.

cc @xiaohu2015 @datumbox

datumbox:
> @datumbox please have a look at my latest commit. It is regarding moving the relative_position_bias_table to the constructor.

@jdsgomes Your changes look good to me. 


The tiny differences noted in accuracy on the trained models are more likely due to the seed than to the change you made. We normally do 3 runs per model anyway so you should be able to confirm this. For the other model that is still on training, epoch 7 as you know is too early, so it doesn't really worry me. We can keep an eye on it and confirm once you are back. Have a good one!

xiaohu2015:
@datumbox the drawbacks of the modified version：it will increase the parameters. 
`here we define relative_position_bias as a parameter with shape (2M−1)×(2M−1) rather than M^2xM^2 to reduce reduce the number of parameters as discussed in paper page5, so we need do the conversion in the forward function.`

torchvision/models/swin_transformer.py
```diff
@@ -0,0 +1,516 @@
...
+        relative_position_bias_table = torch.zeros(
+            (2 * window_size - 1) * (2 * window_size - 1), num_heads
+        )  # 2*Wh-1 * 2*Ww-1, nH
+        nn.init.trunc_normal_(relative_position_bias_table, std=0.02)
+
+        relative_position_bias = relative_position_bias_table[relative_position_index]  # type: ignore[index]
+        relative_position_bias = relative_position_bias.view(
+            self.window_size * self.window_size, self.window_size * self.window_size, -1
+        )
+        self.relative_position_bias = nn.Parameter(relative_position_bias.permute(2, 0, 1).contiguous().unsqueeze(0))
```
datumbox, line 207: Unfortunately as @xiaohu2015 highlighted this increased the number of parameters. Proof:
```python
from torchvision import models as M
model = M.swin_t()
sum(p.numel() for p in model.parameters())
```

Joao's commit 41faba232668f7ac4273a0cf632c0d0130c7ce9c: 28596370
Hu's commit 2500ff3763146d57d959f358dad1a667c4db0db8: 28288354

@jdsgomes I think here we both missed the fact that `relative_position_index` contains repeated indexes. To verify checkout @xiaohu2015's code and check:
```python

>>> model.features[1][0].attn.relative_position_index
tensor([84, 83, 82,  ..., 86, 85, 84])
```

I think we should revert the change.

torchvision/models/swin_transformer.py
```diff
@@ -0,0 +1,516 @@
...
+        norm_layer (nn.Module, optional): Normalization layer. Default: None.
+    """
+
+    def __init__(
+        self,
+        patch_size: int = 4,
+        num_classes: int = 1000,
+        embed_dim: int = 96,
+        depths: List[int] = [2, 2, 6, 2],
+        num_heads: List[int] = [3, 6, 12, 24],
```
datumbox, line 299: We shouldn't use mutable values as default values. If `depths` or `num_heads` get modified, due to the way that Python works, their default values will change for all new instances of the class. Effectively the default values are shared for all instances of the class and if one instance modifies it, it affects all others. This is why we often opt for the `some_value = None` idiom in this codebase.

I recommend the following:
1. Most of the parameters that change across variants such as depths, num_heads and embed_dim (excluding stochastic_depth_prob) should be mandatory and thus have no default values.
2. I would shuffle things around to make them required and I would rearrange the order of the args to make them aligned with ViT and other implementations (for example num_classes is typically added prior the bock/norm_layer

xiaohu2015: thanks a lot, I totally agree with your advice.

torchvision/models/swin_transformer.py
xiaohu2015: > > @xiaohu2015 Hmmm... Did you try moving it on the constructor and got issues during training? Or that's something you didn't try and you are worried it might affect us?
> 
> I do a toy experiment, I found it works. But I am not sure if it will affect performance, maybe you can do end2end training to test it.

@datumbox @jdsgomes the toy experiment is not for swin_t, so I am sorry that I maybe give some incorrect information.

xiaohu2015: the problem is that some compute can't be included in the `forward`

datumbox: No worries for the initial confusion with the toy example. I think you did awesome job flagging that we both missed your point. We can revert the change once Joao is back or feel free to revert at your end.

xiaohu2015:
> The training of `swin_t` before moving `relative_position_bias_table` is ongoing and so far the results seem to align nicely with the logs provided by @xiaohu2015. Namely, after epoch 66 we get:
> 
> `Test: Acc@1 72.278 Acc@5 91.274 (own run id 36469)`
> 
> vs
> 
> `Test: Acc@1 72.406 Acc@5 91.526 (@xiaohu2015 submission)`
> 
> The training of the [modified version ](https://github.com/pytorch/vision/pull/5491/commits/f0615440bf18617dc0e5dc4839bd5ed27e5ed010)(with `relative_position_bias_table` is the constructor) is also ongoing and so far (epoch 7) the results are slightly lower:
> 
> `Test: Acc@1 39.926 Acc@5 65.688 (own run id 36960)`
> 
> vs
> 
> `Test: Acc@1 40.964 Acc@5 66.462 (@xiaohu2015 submission)`
> 
> I have also training jobs for the `swin_s` and `swin_b` running in parallel. I will be away until 20/04/2022 so by then I hope all variants have been trained and we can confirm the final performance is in line with the submission and with the paper.
> 
> cc @xiaohu2015 @datumbox

any update?

datumbox:
@xiaohu2015 Apologies for the delays. There have been a few bank holidays in the UK and some members of our team are still away. I believe all pendings on your side are completed, we just need some additional time to revert the change proposed above and complete the training. So nothing to do on your side, just give us a bit more time and we should be good to go :)

jdsgomes:
> @xiaohu2015 Apologies for the delays. There have been a few bank holidays in the UK and some members of our team are still away. I believe all pendings on your side are completed, we just need some additional time to revert the change proposed above and complete the training. So nothing to do on your side, just give us a bit more time and we should be good to go :)

Hi, I was away until today but I am checking now how the results went. Reading through the messages I will also have to retrain the models, thanks for flagging it!

torchvision/models/swin_transformer.py
jdsgomes: I will revert this, thanks for pointing it out @xiaohu2015, and for reverting @datumbox

jdsgomes:
So the training in our end timed out but after roughly half of the epochs we can confirm that the results can be replicated nicely for the `swin_t` and for the other variants the numbers seem to make sense too.


  | acc @1 | acc @5
-- | -- | --
swin_t @xiaohu2015 (epoch 128) | 75.562  | 93.180
swin_t @jdsgomes (epoch 128) | 74.796 | 92.588
swin_s @jdsgomes (epoch 128) | 77.590 | 94.028
swin_b @jdsgomes (epoch 128) | 77.652 | 94.014

xiaohu2015:
> So the training in our end timed out but after roughly half of the epochs we can confirm that the results can be replicated nicely for the `swin_t` and for the other variants the numbers seem to make sense too.
> 
>  	acc @1	acc @5
> swin_t @xiaohu2015 (epoch 128)	75.562	93.180
> swin_t @jdsgomes (epoch 128)	74.796	92.588
> swin_s @jdsgomes (epoch 128)	77.590	94.028
> swin_b @jdsgomes (epoch 128)	77.652	94.014

it seems the result make sense, I checked the offical log, it got Acc@1 74.890 Acc@5 92.806 at 128 epoch. https://github.com/SwinTransformer/storage/files/7745562/log_swin_tiny_patch4_window7_224.txt

torchvision/models/swin_transformer.py
```diff
@@ -0,0 +1,524 @@
...
+    if weights is not None:
+        model.load_state_dict(weights.get_state_dict(progress=progress))
+
+    return model
+
+
+_COMMON_META = {
+    "task": "image_classification",
+    "architecture": "SwinTransformer",
+    "publication_year": 2021,
```
datumbox, line 413: The fields `task`, `architecture`, `publication_year` and `interpolation` are moved from meta-data. See #5848 and #5852. Let's remove them from here.

xiaohu2015: OK

jdsgomes:
Hi @xiaohu2015 

Finally I get the final results:
- `swin_t`: `80.912` (batch 1 testing) vs  `81.204` (results reported after training) vs `81.3` (paper)
- `swin_s`:  `82.426` (batch 1 testing) vs  `82.598` (results reported after training) vs `83` (paper)
- `swin_b`: `82.75` (batch 1 testing) vs  `82.940` (results reported after training) vs `83.5` (paper) 

As you see there are some small discrepancies in the final testing vs the results reported in the paper. I will reach out offline to discuss what could be the causes of the discrepancies

xiaohu2015:
> Hi @xiaohu2015
> 
> Finally I get the final results:
> 
> * `swin_t`: `81.204` (batch 1 testing) vs  `81.204` (results reported after training) vs `81.3` (paper)
> * `swin_s`:  `82.602` (batch 1 testing) vs  `82.598` (results reported after training) vs `83` (paper)
> * `swin_b`: `82.750` (batch 1 testing) vs  `82.940` (results reported after training) vs `83.5` (paper)
> 
> As you see there are some small discrepancies in the final testing vs the results reported in the paper. I will reach out offline to discuss what could be the causes of the discrepancies
> 
> ps: corrected the results after @xiaohu2015 spotted the missing pre-processing transforms in the Weights.

thanks, it seems that the swin_t can match the offical result, but for swin_s and swin_b, it drops about 0.5 point. I checked the training logs, I found the result can match the result with https://github.com/open-mmlab/mmclassification/tree/master/configs/swin_transformer before 200 epoch (for swint_b) and 250 epoch (for swint_s). I think maybe there is some difference about the lr scheduler. Torchvision use epoch-based lr scheduler, but timm and mmcls use iteration-based lr scheduler. @datumbox @jdsgomes what your thoughts?

xiaohu2015:
I can't understand the difference when use different batch_size, what can cause this? @datumbox

xiaohu2015:
> @xiaohu2015 Can you please share the logs with @jdsgomes and all the information that will allow us to reproduce your experiment (for example the git commit hashcodes you used etc). It's also unclear to me whether you used TorchVision's reference scripts or something else in your experiments. Could you please clarify?

for swin_t, I have shared the training logs with @jdsgomes, I got `Acc@1 81.222 Acc@5 95.332`, @jdsgomes can reproduce the training (81.204), the result can match the offical result. But for swin_s and swin_b, it drops about 0.5 point compared to the offical result, in fact, I only trained swin_t.

jdsgomes:
> > @xiaohu2015 Can you please share the logs with @jdsgomes and all the information that will allow us to reproduce your experiment (for example the git commit hashcodes you used etc). It's also unclear to me whether you used TorchVision's reference scripts or something else in your experiments. Could you please clarify?
> 
> for swin_t, I have shared the training logs (just use TorchVision's reference script ) with @jdsgomes, I got `Acc@1 81.222 Acc@5 95.332`, @jdsgomes can reproduce the training (81.204), the result can match the offical result. But for swin_s and swin_b, it drops about 0.5 point compared to the offical result, in fact, I only trained swin_t. I plan to check these two models. thanks to the training logs from @jdsgomes, I found that these two models seem to get behind in the last few epochs.
> 
> @jdsgomes yes, iteration-based lr scheduler and epoch-based lr scheduler should be equivalent. I just suspect that might be the reason, because some minor difference can make a result difference.

After discussing offline with @datumbox I think we should proceed to merge the PR with the swin_t only since it is clear that we can reproduce the result, so great work @xiaohu2015 ! 

After that we can continue investigations to close the gap and aim to merge the other variants in a different PR.

I will do the final cleanups between today and tomorrow.

datumbox:
Just wanted to echo what Joao said. Big massive thank you @xiaohu2015 for your awesome contribution. Top notch code and excellent research reproduction skills. Also apologies for taking us long to review and reproduce the PR; it's something we want to improve upon.

Looking forward seeing this merged!

xiaohu2015:
@datumbox @jdsgomes Thanks very much！

docs/source/models.rst
```diff
@@ -97,6 +98,9 @@ You can construct a model with random weights by calling its constructor:
     convnext_small = models.convnext_small()
     convnext_base = models.convnext_base()
     convnext_large = models.convnext_large()
+    swin_t = models.swin_t()
+    swin_s = models.swin_s()
+    swin_b = models.swin_b()
```
datumbox, line 103: This is the legacy way we document models and we should definitely add info here. In addition to this, we need to implement the new way of documenting models described at #5833

torchvision/models/swin_transformer.py
```diff
@@ -0,0 +1,529 @@
...
+
+class Swin_T_Weights(WeightsEnum):
+    IMAGENET1K_V1 = Weights(
+        url="https://download.pytorch.org/models/swin_t-81486767.pth",
+        transforms=partial(
+            ImageClassification, crop_size=224, resize_size=238, interpolation=InterpolationMode.BICUBIC
+        ),
+        meta={
+            **_COMMON_META,
+        },
```
datumbox, line 423: You need to add meta-data such as "metrics.Acc@1/5", "num_params", "recipe", "min_size" etc. These are all mandatory fields. See [here](https://github.com/pytorch/vision/blob/main/test/test_extended_models.py#L95-L97) for full list.

datumbox (commented):
Thanks @jdsgomes. I know you are looking into all these, just added a few comments for changes done on the documentation side last week so that we don't forget to include it.

docs/source/models.rst
jdsgomes: yes I was woking on this in a separate commit - should be added now.

xiaohu2015:
for other models, I can convert the offical weight to torchvision version just like efficientnet.

jdsgomes:
> to

@xiaohu2015 I understand that would be useful to include pre-trained weights from the initial implementation, but we would prefer to include the other variants once we can replicate the results fully. I am running a few experiments now, and hopefully we can get good results, but for now I will remove even the constructors  so this PR can be merged.

datumbox (approved):
LGTM, thanks again @xiaohu2015 for the awesome contribution.

@jdsgomes thanks as well for your support and guidance.

I think we are good to merge. Just make sure we remove the unnecessary expect files `ModelTester.test_swin_*_expect.pk` for variants s/b/l that were removed.

jdsgomes (approved):
Thanks @xiaohu2015 for the great contribution and @datumbox for the feedback

Issue #5148, opened 2021-12-30
GPU decoder refactoring


prabhat00155:
https://github.com/pytorch/vision/pull/5019#discussion_r776814346
https://github.com/pytorch/vision/pull/5019#discussion_r776807823
https://github.com/pytorch/vision/pull/5019#discussion_r776807411
https://github.com/pytorch/vision/pull/5019#discussion_r776804991
https://github.com/pytorch/vision/pull/5019#discussion_r776802688
https://github.com/pytorch/vision/pull/5019#discussion_r776800733
https://github.com/pytorch/vision/pull/5019#discussion_r776800512

PR #5495, opened 2022-02-28
Move VideoReader out of __init__

prabhat00155:
Create a separate file for VideoReader class and clean up `__init__.py`.
More information: https://github.com/pytorch/vision/pull/5019#discussion_r776814346
Partly addresses: https://github.com/pytorch/vision/issues/5148


datumbox (approved):
@prabhat00155 I understand that this is just a refactoring/moving out code to a separate file. If that's the case, I'm stumping to unblock.

PR #5496, opened 2022-02-28
Expand prototype functional kernel scriptability tests

pmeier:
While eager vs scripted tests require a case by case treatment due to varying inputs, just checking if something is scriptable can be achieved outside of that "framework".


test/test_prototype_transforms_functional.py
```diff
@@ -199,21 +199,38 @@ def resize_bounding_box():
...
+        and callable(kernel)
+        and any(feature_type in name for feature_type in {"image", "segmentation_mask", "bounding_box", "label"})
+        and "pil" not in name
+        and (
+            name
+            not in {
+                "get_image_size",
+                "get_image_num_channels",
+            }
+        )
```
pmeier, line 217: This is only temporary. If we have full coverage for eager vs scripted tests we can simply reuse this instead of this abomination. Still, I've checked manually and for now this covers all kernels that we want to test and should be somewhat robust for future cases.

torchvision/prototype/transforms/functional/_type_conversion.py
```diff
@@ -22,4 +22,4 @@ def decode_video_with_av(encoded_video: torch.Tensor) -> Tuple[torch.Tensor, tor
 
 
 def label_to_one_hot(label: torch.Tensor, *, num_categories: int) -> torch.Tensor:
-    return cast(torch.Tensor, one_hot(label, num_classes=num_categories))
+    return one_hot(label, num_classes=num_categories)  # type: ignore[no-any-return]
```
datumbox, line 25: Question: Is mypy complaining because it can't tell `LongTensor` is `Tensor`?

test/test_prototype_transforms_functional.py
```diff
@@ -199,21 +199,38 @@ def resize_bounding_box():
...
-
-    @pytest.mark.parametrize(
-        ("functional_info", "sample_input"),
-        [
-            pytest.param(functional_info, sample_input, id=f"{functional_info.name}-{idx}")
-            for functional_info in FUNCTIONAL_INFOS
-            for idx, sample_input in enumerate(functional_info.sample_inputs())
-        ],
-    )
-    def test_eager_vs_scripted(self, functional_info, sample_input):
```
datumbox, line 215: Sorry, I probably miss the context but why can't we keep the `test_eager_vs_scripted` for the image cases?

torchvision/prototype/transforms/functional/_type_conversion.py
pmeier: `mypy` infers `Any` as return type of `one_hot`.

test/test_prototype_transforms_functional.py
datumbox: Damn I missed this. Sorry.

datumbox (approved):
LGTM, thanks!

torchvision/prototype/transforms/functional/_type_conversion.py
datumbox: Yeah I can see why mypy can't read the type. [Here](https://github.com/pytorch/pytorch/blob/28339ddc252325a1fc93e3302e9c4f93cdc0034b/torch/nn/functional.py#L4477-L4480) is the code on PyTorch.

PR #5497, opened 2022-02-28
Improve test_videoapi

prabhat00155:
- Use parametrize instead of looping over the list. This would help point out the exact scenario in case of failures.
- Remove unused variables
- Simplify if conditions


datumbox (approved):
LGTM, thanks!

Issue #5398, opened 2022-02-09
Allow for arbitrary sequences in torchvision/transforms/functional


Parskatt:
### 🚀 The feature

Replace all checks for (list,tuple) in functional with Iterable. 

### Motivation, pitch

There are about 8 checks in functional right now which requires the argument to be either list or tuple, e.g.
https://github.com/pytorch/vision/blob/main/torchvision/transforms/functional.py#L1059
Instead of performing these checks, simply checking for Iterable allows to use e.g. numpy arrays.

In my usecase i generate parameters using numpy, and I don't want to convert them.

The documentation indicates that general sequences should work, so the current checks are counter-intuitive.

datumbox:
Sounds reasonable but note we are limited due to JIT compatibility. We wound't be able to add for example numpy arrays in the check because that won't be JIT-scritable. We need to review what exactly is supported by JIT and what's the common denominator/type for iterable objects. 

@pmeier @vfdev-5 Good feedback for the new API. Perhaps we could push all these checks on the non-JIT-scriptable high-level kernels and allow the low-level ones just work on input with minimal validation. Thoughts?

pmeier:
Given that we do not plan to make the transforms (objects not kernels) JIT scriptable, I think we should be able to fix the annotations on the transforms.

PR #3001, opened 2020-11-13
[BC-breaking] ColorJitter gets its random params by calling get_params()

datumbox:
As per @fmassa's [comment](https://github.com/pytorch/vision/pull/2298#discussion_r467093034), I'm rewriting ColorJitter so that it receives its random parameters from the `get_params()` method. The ColorJitter remains torch-scriptable. 

Affects #2672 and #2669.

PR #5500, opened 2022-02-28
[NOMERGE] Review new Transforms API

datumbox:
Adding comments for the proposed new Transforms API as a whole. Part of the comments need to be addressed now, others are FYI and others can be tackled by opening issues and to track the work. 

There are a few low-hanging fruit improvements that can be done immediately in separate PRs without much discussion. Others definitely require more thought and discussion so for these I propose having a chat offline and the summarizing here.

cc @pmeier @vfdev-5

### DIFF
--- a/torchvision/prototype/features/_bounding_box.py
+++ b/torchvision/prototype/features/_bounding_box.py
@@ -39,6 +39,7 @@ class BoundingBox(_Feature):
     def to_format(self, format: Union[str, BoundingBoxFormat]) -> BoundingBox:
         # TODO: this is useful for developing and debugging but we should remove or at least revisit this before we
         #  promote this out of the prototype state
+        # Do we still need this?
 
         # import at runtime to avoid cyclic imports
         from torchvision.prototype.transforms.functional import convert_bounding_box_format
--- a/torchvision/prototype/features/_image.py
+++ b/torchvision/prototype/features/_image.py
@@ -17,6 +17,9 @@ class ColorSpace(StrEnum):
     OTHER = 0
     GRAYSCALE = 1
     RGB = 3
+    # On io, we currently support Alpha transparency for Grayscale and RGB. We also support palette for PNG.
+    # Our enum must know about these. We should also allow for users to strip out the alpha transparency which is
+    # currently not supported by any colour transform.
 
 
 class Image(_Feature):
@@ -78,9 +81,11 @@ class Image(_Feature):
     def show(self) -> None:
         # TODO: this is useful for developing and debugging but we should remove or at least revisit this before we
         #  promote this out of the prototype state
+        # Do we still need this?
         to_pil_image(make_grid(self.view(-1, *self.shape[-3:]))).show()
 
     def draw_bounding_box(self, bounding_box: BoundingBox, **kwargs: Any) -> Image:
         # TODO: this is useful for developing and debugging but we should remove or at least revisit this before we
         #  promote this out of the prototype state
+        # Do we still need this?
         return Image.new_like(self, draw_bounding_boxes(self, bounding_box.to_format("xyxy").view(-1, 4), **kwargs))
--- a/torchvision/prototype/transforms/__init__.py
+++ b/torchvision/prototype/transforms/__init__.py
@@ -12,3 +12,14 @@ from ._meta import ConvertBoundingBoxFormat, ConvertImageDtype, ConvertImageColo
 from ._misc import Identity, Normalize, ToDtype, Lambda
 from ._presets import CocoEval, ImageNetEval, VocEval, Kinect400Eval, RaftEval
 from ._type_conversion import DecodeImage, LabelToOneHot
+
+# What are the migration plans for Classes without new API equivalents? There are two categories:
+# 1. Deprecated methods which have equivalents on the new API (_legacy.py?):
+# - Grayscale, RandomGrayscale: use ConvertImageColorSpace
+# 2. Those without equivalents on the new API:
+# - Pad, RandomCrop, RandomHorizontalFlip, RandomVerticalFlip, RandomPerspective, FiveCrop, TenCrop, ColorJitter,
+# RandomRotation, RandomAffine, GaussianBlur, RandomInvert, RandomPosterize, RandomSolarize, RandomAdjustSharpness,
+# RandomAutocontrast, RandomEqualize, LinearTransformation (must be added)
+# - PILToTensor, ToPILImage (_legacy.py?)
+# - ToTensor (deprecate vfdev-5?)
+# We need a plan for both categories implemented on the new API.
--- a/torchvision/prototype/transforms/_augment.py
+++ b/torchvision/prototype/transforms/_augment.py
@@ -34,6 +34,10 @@ class RandomErasing(Transform):
         if p < 0 or p > 1:
             raise ValueError("Random erasing probability should be between 0 and 1")
         # TODO: deprecate p in favor of wrapping the transform in a RandomApply
+        # The above approach is very composable but will lead to verbose code. Instead, we can create a base class
+        # that inherits from Transform (say RandomTransform) that receives the `p` on constructor and by default
+        # implements the `p` random check on forward. This is likely to happen on the final clean ups, so perhaps
+        # update the comment to indicate accordingly OR create an issue to track this discussion.
         self.p = p
         self.scale = scale
         self.ratio = ratio
@@ -84,6 +88,8 @@ class RandomErasing(Transform):
             break
         else:
             i, j, h, w, v = 0, 0, img_h, img_w, image
+            # FYI: Now taht we are not JIT-scriptable, I probably can avoid copying-pasting the image to itself in this
+            # scenario. Perhaps a simple clone would do.
 
         return dict(zip("ijhwv", (i, j, h, w, v)))
 
@@ -95,6 +101,7 @@ class RandomErasing(Transform):
             return features.Image.new_like(input, output)
         elif isinstance(input, torch.Tensor):
             return F.erase_image_tensor(input, **params)
+        # FYI: we plan to add support for PIL, as part of Batteries Included
         else:
             return input
 
@@ -112,6 +119,8 @@ class RandomMixup(Transform):
         self._dist = torch.distributions.Beta(torch.tensor([alpha]), torch.tensor([alpha]))
 
     def _get_params(self, sample: Any) -> Dict[str, Any]:
+        # Question: Shall we enforce here the existence of Labels in the sample? If yes, this method of validating
+        # input won't work if get_params() gets public and the user sis able to provide params in forward.
         return dict(lam=float(self._dist.sample(())))
 
     def _transform(self, input: Any, params: Dict[str, Any]) -> Any:
@@ -134,6 +143,7 @@ class RandomCutmix(Transform):
         self._dist = torch.distributions.Beta(torch.tensor([alpha]), torch.tensor([alpha]))
 
     def _get_params(self, sample: Any) -> Dict[str, Any]:
+        # Question: Same as above for Labels.
         lam = float(self._dist.sample(()))
 
         image = query_image(sample)
--- a/torchvision/prototype/transforms/_auto_augment.py
+++ b/torchvision/prototype/transforms/_auto_augment.py
@@ -35,6 +35,9 @@ class _AutoAugmentBase(Transform):
         self.fill = fill
 
     def _get_random_item(self, dct: Dict[K, V]) -> Tuple[K, V]:
+        # Question: This looks like a utility method that doesn't depend on self and could move out of this
+        # class to simplify its structure (for someone who tries to understand its API to implement a new method).
+        # Thoughts?
         keys = tuple(dct.keys())
         key = keys[int(torch.randint(len(keys), ()))]
         return key, dct[key]
@@ -46,6 +49,9 @@ class _AutoAugmentBase(Transform):
     def _extract_image(
         self, sample: Any
     ) -> Tuple[Tuple[Any, ...], Union[PIL.Image.Image, torch.Tensor, features.Image]]:
+        # How about providing the unsupported types as parameter to this method to avoid hardcoding it? This can
+        # allow us to remove the separate _check_unsupported method. Also this method could be removed from the
+        # class to simplify its structure.
         def fn(
             id: Tuple[Any, ...], input: Any
         ) -> Optional[Tuple[Tuple[Any, ...], Union[PIL.Image.Image, torch.Tensor, features.Image]]]:
@@ -67,6 +73,7 @@ class _AutoAugmentBase(Transform):
     def _parse_fill(
         self, image: Union[PIL.Image.Image, torch.Tensor, features.Image], num_channels: int
     ) -> Optional[List[float]]:
+        # Question: How do you feel about turning this also a util?
         fill = self.fill
 
         if isinstance(image, PIL.Image.Image) or fill is None:
@@ -204,6 +211,8 @@ class AutoAugment(_AutoAugmentBase):
         "ShearY": (lambda num_bins, image_size: torch.linspace(0.0, 0.3, num_bins), True),
         "TranslateX": (lambda num_bins, image_size: torch.linspace(0.0, 150.0 / 331.0 * image_size[1], num_bins), True),
         "TranslateY": (lambda num_bins, image_size: torch.linspace(0.0, 150.0 / 331.0 * image_size[0], num_bins), True),
+        # How do you feel about explicitly passing height, width here instead of image_size? I did it in an earlier PR
+        # but removed it due to the verbosity. Thoughts?
         "Rotate": (lambda num_bins, image_size: torch.linspace(0.0, 30.0, num_bins), True),
         "Brightness": (lambda num_bins, image_size: torch.linspace(0.0, 0.9, num_bins), True),
         "Color": (lambda num_bins, image_size: torch.linspace(0.0, 0.9, num_bins), True),
@@ -222,6 +231,8 @@ class AutoAugment(_AutoAugmentBase):
     }
 
     def __init__(self, policy: AutoAugmentPolicy = AutoAugmentPolicy.IMAGENET, **kwargs: Any) -> None:
+        # I think using kwargs on the constructor makes it hard to users to understand which parameters to provide.
+        # I recommend passing explicitly interpolation and fill in all AA constructors.
         super().__init__(**kwargs)
         self.policy = policy
         self._policies = self._get_policies(policy)
--- a/torchvision/prototype/transforms/_container.py
+++ b/torchvision/prototype/transforms/_container.py
@@ -38,6 +38,7 @@ class RandomApply(Transform):
 
 class RandomChoice(Transform):
     def __init__(self, *transforms: Transform) -> None:
+        # This method should receive optionally a list of probabilities and sample transforms proportionally.
         super().__init__()
         self.transforms = transforms
         for idx, transform in enumerate(transforms):
--- a/torchvision/prototype/transforms/_meta.py
+++ b/torchvision/prototype/transforms/_meta.py
@@ -4,7 +4,7 @@ import PIL.Image
 import torch
 from torchvision.prototype import features
 from torchvision.prototype.transforms import Transform, functional as F
-from torchvision.transforms.functional import convert_image_dtype
+from torchvision.transforms.functional import convert_image_dtype  # We should have our an alias for this on the new API
 
 
 class ConvertBoundingBoxFormat(Transform):
@@ -23,6 +23,10 @@ class ConvertBoundingBoxFormat(Transform):
 
 
 class ConvertImageDtype(Transform):
+    # Question: Why do we have both this and a ToDtype Transform? Is this due to BC? Ideally we could move people off
+    # from methods that did an implicit normalization of the values (like this one, or to_tensor). cc @vfdev-5
+    # If that's the case, we should move to _legacy and add deprecation warnings from day one to push people to use
+    # the new methods.
     def __init__(self, dtype: torch.dtype = torch.float32) -> None:
         super().__init__()
         self.dtype = dtype
@@ -59,7 +63,7 @@ class ConvertImageColorSpace(Transform):
             return features.Image.new_like(input, output, color_space=self.color_space)
         elif isinstance(input, torch.Tensor):
             if self.old_color_space is None:
-                raise RuntimeError("")
+                raise RuntimeError("")  # Add better exception message
 
             return F.convert_image_color_space_tensor(
                 input, old_color_space=self.old_color_space, new_color_space=self.color_space
--- a/torchvision/prototype/transforms/_transform.py
+++ b/torchvision/prototype/transforms/_transform.py
@@ -12,6 +12,10 @@ class Transform(nn.Module):
         super().__init__()
         _log_api_usage_once(self)
 
+    # FYI: I spoke with @NicolasHug offline and provided a valid use-case where it would be useful to make this public,
+    # as you originally intended. We don't have to do this now but if we do end up exposing this publicly, we should
+    # rename it to avoid conflicts with the previous static get_params() which worked completely differently.
+    # We dn't have to do anything now but we should discuss this again after the feedback.
     def _get_params(self, sample: Any) -> Dict[str, Any]:
         return dict()
 
--- a/torchvision/prototype/transforms/functional/__init__.py
+++ b/torchvision/prototype/transforms/functional/__init__.py
@@ -1,3 +1,9 @@
+# He should create an issue that lists the steps that need to be performed for rolling out the API to main TorchVision.
+# I got a similar for the models, see here: https://github.com/pytorch/vision/issues/4679
+# One of the key things we would need to take care of is that all the kernels below will need logging. This is because
+# there will be no high-level kernel (like `F` on main) and we would instead need to put tracking directly in each
+# low-level kernels which will be now public (now functional_pil|tensor are private).
+
 from torchvision.transforms import InterpolationMode  # usort: skip
 from ._meta import (
     convert_bounding_box_format,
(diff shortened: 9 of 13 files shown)


torchvision/prototype/transforms/functional/_misc.py
```diff
@@ -39,4 +39,5 @@ def gaussian_blur_image_tensor(
 
 
 def gaussian_blur_image_pil(img: PIL.Image, kernel_size: List[int], sigma: Optional[List[float]] = None) -> PIL.Image:
+    # It would really help to remove to_tensor from here vfdev-5
```
vfdev-5, line 42: I'd really like to help with that but there is no inverse for `pil_to_tensor` without rescaling the range...

torchvision/prototype/features/_bounding_box.py
```diff
@@ -39,6 +39,7 @@ def __new__(
     def to_format(self, format: Union[str, BoundingBoxFormat]) -> BoundingBox:
         # TODO: this is useful for developing and debugging but we should remove or at least revisit this before we
         #  promote this out of the prototype state
+        # Do we still need this?
```
pmeier, line 42: Need, no. Want, yes. I actually want to keep them after this is promoted out of prototype. It is a convenient way to work with the features outside of a training pipeline. Compare for example

```py

transform = transforms.ConvertBoundingBoxFormat(format="xyxy")

bounding_box = features.BoundingBox(...)
xyxy_bounding_box = transform(bounding_box)
```

or

```py

bounding_box = features.BoundingBox(...)
xyxy_bounding_box = features.BoundingBox.new_like(
    bounding_box, 
    convert_bounding_box_format(bounding_box, old_format=bounding_box.format, new_format="xyxy"),
    format="xyxy",
)
```

to

```py

bounding_box = features.BoundingBox(...)
xyxy_bounding_box = bounding_box.to_format("xyxy")
```

Still, I agree, the lazy import needs to go. My proposal is to move the actual conversion code from `transforms.functional` to a private function like `_convert_bounding_box_format` in this module. We could use it here in `to_format` and simply expose it in `transforms.functional`.

---

This comment applies to all other comments that ask the same.

torchvision/prototype/features/_image.py
```diff
@@ -17,6 +17,9 @@ class ColorSpace(StrEnum):
     OTHER = 0
     GRAYSCALE = 1
     RGB = 3
+    # On io, we currently support Alpha transparency for Grayscale and RGB. We also support palette for PNG.
+    # Our enum must know about these. We should also allow for users to strip out the alpha transparency which is
+    # currently not supported by any colour transform.
```
pmeier, line 22: > We also support palette for PNG.

Could you add a link for that? Otherwise, we are talking about this enum, correct?

https://github.com/pytorch/vision/blob/95d418970e6dbf2e4d928a204c4e620da7bccdc0/torchvision/io/image.py#L16-L31

I'll send a PR to include the `RGBA` and `GRAYSCALE_ALPHA` into the enum and also provide conversion functions for them.

torchvision/prototype/transforms/__init__.py
```diff
@@ -12,3 +12,14 @@
 from ._misc import Identity, Normalize, ToDtype, Lambda
 from ._presets import CocoEval, ImageNetEval, VocEval, Kinect400Eval, RaftEval
 from ._type_conversion import DecodeImage, LabelToOneHot
+
+# What are the migration plans for Classes without new API equivalents? There are two categories:
+# 1. Deprecated methods which have equivalents on the new API (_legacy.py?):
+# - Grayscale, RandomGrayscale: use ConvertImageColorSpace
```
pmeier, line 18: :+1:

torchvision/prototype/transforms/__init__.py
```diff
@@ -12,3 +12,14 @@
...
+# 2. Those without equivalents on the new API:
+# - Pad, RandomCrop, RandomHorizontalFlip, RandomVerticalFlip, RandomPerspective, FiveCrop, TenCrop, ColorJitter,
+# RandomRotation, RandomAffine, GaussianBlur, RandomInvert, RandomPosterize, RandomSolarize, RandomAdjustSharpness,
+# RandomAutocontrast, RandomEqualize, LinearTransformation (must be added)
```
pmeier, line 22: :+1: For everything but `FiveCrop` and `TenCrop`. They are outliers in the sense that they get one input and produce multiple outputs. This kind of structure would be suited a lot better to a datapipe rather than a transform. 

This is also reinforced by the fact that these transforms can only be put at the end of any pipeline with the old API given that the output type is not compatible with all the other transforms. Even then, using such a transform would most likely need a custom collate function.

I couldn't find any usage in our reference scripts. Does someone know how exactly they are used so I can propose something more concrete here?

torchvision/prototype/transforms/__init__.py
```diff
@@ -12,3 +12,14 @@
...
+# - PILToTensor, ToPILImage (_legacy.py?)
```
pmeier, line 23: Shouldn't we also have a more general `ConvertImageType` transform here that takes one of `"pil"`, `"vanilla_tensor"`, `"feature"` as input? That would follow the same approach that we have for the other conversions.

torchvision/prototype/transforms/_augment.py
```diff
@@ -34,6 +34,10 @@ def __init__(
         if p < 0 or p > 1:
             raise ValueError("Random erasing probability should be between 0 and 1")
         # TODO: deprecate p in favor of wrapping the transform in a RandomApply
+        # The above approach is very composable but will lead to verbose code. Instead, we can create a base class
+        # that inherits from Transform (say RandomTransform) that receives the `p` on constructor and by default
+        # implements the `p` random check on forward. This is likely to happen on the final clean ups, so perhaps
+        # update the comment to indicate accordingly OR create an issue to track this discussion.
```
pmeier, line 40: :+1:

torchvision/prototype/transforms/_augment.py
```diff
@@ -84,6 +88,8 @@ def _get_params(self, sample: Any) -> Dict[str, Any]:
             break
         else:
             i, j, h, w, v = 0, 0, img_h, img_w, image
+            # FYI: Now taht we are not JIT-scriptable, I probably can avoid copying-pasting the image to itself in this
+            # scenario. Perhaps a simple clone would do.
```
pmeier, line 92: Yeah, there is major cleanup coming up. A lot of the annotations on the transforms are either wrong or to strict to appease torchscript. We should revisit before we promote out of prototype. There is also a user request for this: #5398

torchvision/prototype/transforms/_augment.py
```diff
@@ -112,6 +119,8 @@ def __init__(self, *, alpha: float) -> None:
         self._dist = torch.distributions.Beta(torch.tensor([alpha]), torch.tensor([alpha]))
 
     def _get_params(self, sample: Any) -> Dict[str, Any]:
+        # Question: Shall we enforce here the existence of Labels in the sample? If yes, this method of validating
+        # input won't work if get_params() gets public and the user sis able to provide params in forward.
```
pmeier, line 123: I agree we should be loud about this, but I don't know about an error. Maybe a warning is sufficient? It is safer to upgrade to an error later than downgrading an error to a warning.

--- 

For input validation like this, we should simply do

```py
def forward(self, *inputs):
    validate_inputs(inputs)
```

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -35,6 +35,9 @@ def __init__(
         self.fill = fill
 
     def _get_random_item(self, dct: Dict[K, V]) -> Tuple[K, V]:
+        # Question: This looks like a utility method that doesn't depend on self and could move out of this
+        # class to simplify its structure (for someone who tries to understand its API to implement a new method).
+        # Thoughts?
```
pmeier, line 40: I think this is a matter of style. In general, I tend to put utility methods that only are used by the class and its derivatives on the base class. There is no harm to move them to standalone functions. The only downside is that if someone outside of `torchvision` inherits from `_AutoAugmentBase` they don't have access to the functionality, since we don't expose it. No strong opinion here.

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -46,6 +49,9 @@ def _check_unsupported(self, input: Any) -> None:
     def _extract_image(
         self, sample: Any
     ) -> Tuple[Tuple[Any, ...], Union[PIL.Image.Image, torch.Tensor, features.Image]]:
+        # How about providing the unsupported types as parameter to this method to avoid hardcoding it? This can
+        # allow us to remove the separate _check_unsupported method. Also this method could be removed from the
+        # class to simplify its structure.
```
pmeier, line 54: :+1: about the parametrization.

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -204,6 +211,8 @@ class AutoAugment(_AutoAugmentBase):
         "ShearY": (lambda num_bins, image_size: torch.linspace(0.0, 0.3, num_bins), True),
         "TranslateX": (lambda num_bins, image_size: torch.linspace(0.0, 150.0 / 331.0 * image_size[1], num_bins), True),
         "TranslateY": (lambda num_bins, image_size: torch.linspace(0.0, 150.0 / 331.0 * image_size[0], num_bins), True),
+        # How do you feel about explicitly passing height, width here instead of image_size? I did it in an earlier PR
+        # but removed it due to the verbosity. Thoughts?
```
pmeier, line 215: Originally I stated I also prefer the non-verbose way that we currently have. Thinking about it more, the extra verbosity wouldn't hurt here since this is not user facing and would make the instances where we actually need the height and width more clear. Thus, I'm now in slight favor of passing height and width separately.

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -222,6 +231,8 @@ class AutoAugment(_AutoAugmentBase):
     }
 
     def __init__(self, policy: AutoAugmentPolicy = AutoAugmentPolicy.IMAGENET, **kwargs: Any) -> None:
+        # I think using kwargs on the constructor makes it hard to users to understand which parameters to provide.
+        # I recommend passing explicitly interpolation and fill in all AA constructors.
```
pmeier, line 235: :+1:

torchvision/prototype/transforms/_container.py
```diff
@@ -38,6 +38,7 @@ def extra_repr(self) -> str:
 
 class RandomChoice(Transform):
     def __init__(self, *transforms: Transform) -> None:
+        # This method should receive optionally a list of probabilities and sample transforms proportionally.
```
pmeier, line 41: To be able to fine tune which transform is chosen? Like 30% transform1 and 70% transform2?

torchvision/prototype/transforms/_meta.py
```diff
@@ -23,6 +23,10 @@ def _transform(self, input: Any, params: Dict[str, Any]) -> Any:
 
 
 class ConvertImageDtype(Transform):
+    # Question: Why do we have both this and a ToDtype Transform? Is this due to BC? Ideally we could move people off
+    # from methods that did an implicit normalization of the values (like this one, or to_tensor). cc @vfdev-5
+    # If that's the case, we should move to _legacy and add deprecation warnings from day one to push people to use
+    # the new methods.
```
pmeier, line 29: One idea that @datumbox and I discussed a while back is abolish the close correspondence of the dtype and the value range for images. Right now we assume the range `[0, 1]` for floating point images and `[0, torch.iinfo(dtype).max]` for integral images. A possible solution for this would be have a `value_range` meta data on the `features.Image` class. It can be set with the defaults from above so for a regular use case the user won't feel any difference. 

For example, let's say we have a uint8 image

```py
import torch

image = features.Image(torch.randint(0, 256, (3, 128, 128)))
```

By default `print(image.value_range)` would print `(0, 255)`. If we now convert the dtype like `image = image.to(torch.float32)`, printing the value range again would give `(0.0, 255.0)` rather than `(0.0, 1.0)` what happens through `ConvertImageDtype`.

This means, all transformations that implicitly rely on the value range like `Normalize` need to be adapted to use the explicit value instead, but we could get rid of `ConvertImageDtype` all together.

torchvision/prototype/transforms/_meta.py
```diff
@@ -59,7 +63,7 @@ def _transform(self, input: Any, params: Dict[str, Any]) -> Any:
             return features.Image.new_like(input, output, color_space=self.color_space)
         elif isinstance(input, torch.Tensor):
             if self.old_color_space is None:
-                raise RuntimeError("")
+                raise RuntimeError("")  # Add better exception message
```
pmeier, line 66: :+1:

torchvision/prototype/transforms/_transform.py
```diff
@@ -12,6 +12,10 @@ def __init__(self) -> None:
         super().__init__()
         _log_api_usage_once(self)
 
+    # FYI: I spoke with @NicolasHug offline and provided a valid use-case where it would be useful to make this public,
+    # as you originally intended. We don't have to do this now but if we do end up exposing this publicly, we should
+    # rename it to avoid conflicts with the previous static get_params() which worked completely differently.
+    # We dn't have to do anything now but we should discuss this again after the feedback.
```
pmeier, line 18: Could you share this use case with us? cc @NicolasHug

torchvision/prototype/transforms/functional/_misc.py
pmeier: Related: https://github.com/pytorch/vision/pull/5500#discussion_r816525315

datumbox: @vfdev-5 I would like to avoid offering this method as part of the new API. Based on the implementation of `to_pil_image` I see that the x255 multiplication happens ONLY if the input is floating point. Hence I think we could avoid this by ensuring the image is of uint8 type prior using `to_pil_image`. Thoughts?

torchvision/prototype/features/_bounding_box.py
datumbox: I agree that the verbosity of option 2 is an indication that we need a utility. It doesn't necessarily means that the method should be part of BoundingBox, so let's explore options. Note that we should avoid the use of strings in favour of their enums in our examples and internal code to get better static analysis. 

> This comment applies to all other comments that ask the same.

The 2 visualization methods `show()` and `draw_bounding_box()` on `Image` are not the same thing. We should remove them as we don't want to necessarily push PIL on the new API.

torchvision/prototype/features/_image.py
pmeier: Looking into this, conversion from RGBA to RGB is more complicated. For a proper conversion we need to know the background color. We could assume white which will probably work for most cases, but we probably should discuss this first.

datumbox: Palettes are [supported](https://github.com/pytorch/vision/blob/main/torchvision/csrc/io/image/cpu/decode_png.cpp#L107-L109) if the user reads with `IMAGE_READ_MODE_UNCHANGED`. The outcome will be an image with a single channel that has integer IDs for every colour. This is ideal for reading masks that have integer IDs for the classes.

torchvision/prototype/transforms/__init__.py
datumbox: The problem is if we want to roll out the API with BC support, we must to include implementations for all previous methods + offer alternatives for those that would be placed in deprecated state. Thought I understand the challenge, excluding these methods is not OK and we should flag if others like these exist.

The specific transforms are used to produce multiple crops for inference in classification. It is a common trick (especially in the early days of ImageNet competition) to do boost the accuracy of the model by averaging the responses of multiple crops. You can see an example of how this is used [here](https://github.com/pytorch/vision/blob/main/torchvision/transforms/transforms.py#L981-L989).

torchvision/prototype/transforms/__init__.py
datumbox: Here is what I mean. Let's say we are close to releasing the prototype and we have achieved a significant level of BC. On day 0, we will need a PILToTensor to exist. These are likely be in a deprecated state with warnings informing the user that this functionality is removed and they should use X. Your proposal seems like a valid X alternative.

So here I think we need to figure out 2 things:
- What do we do with these legacy classes + provide their implementations on the new API (or aliases). Make sure that the classes continue to work along with the other non-deprecated classes in existing pipelines.
- Provide an alternative class that offers the new functionality in the new API.

torchvision/prototype/transforms/_augment.py
datumbox: Sounds good. There is a non-trivial number of clean ups we need to do prior landing this API to the main area. Please create an issue that describes these clean ups (see this [example](https://github.com/pytorch/vision/issues/4679)). I would also recommend creating a Project to keep track of all the related tickets (see this [example](https://github.com/pytorch/vision/projects/2)).

torchvision/prototype/transforms/_augment.py
datumbox: Because 99% of the classes don't need to validate the output, I would probably avoid creating a mechanism that would need to be overwritten etc (like a proper method). The solution you propose of placing this on top of forward sounds good to me. I would only put those validations straight in `forward`. I would like to hear other opinions thought.

We should figure out these matters prior starting adding the remaining classes.

torchvision/prototype/transforms/_auto_augment.py
datumbox: I agree, your usage if valid. No strong opinion on my side either, the implementation can stay as-is.

@vfdev-5 any thoughts?

torchvision/prototype/transforms/_auto_augment.py
datumbox: I've literally introduced this and took it back on subsequent commits (see [here](https://github.com/pytorch/vision/pull/5487/commits/ead34b3ea23ae207f3667569fe3ce22c42b3c841)). I'm OK if we be verbose.

torchvision/prototype/transforms/_transform.py
NicolasHug: I think the use-case is this: https://github.com/pytorch/vision/pull/3001#issuecomment-814849595

I.e. the ability to "freeze" a transform that has randomly-sampled parameters. This isn't too exotic of a need, I has the same use-case in the RAFT training reference, and ended having to workaround it by batching images (but in some scenarios batching isn't possible if all images aren't available at the same time)

https://github.com/pytorch/vision/blob/95d418970e6dbf2e4d928a204c4e620da7bccdc0/references/optical_flow/transforms.py#L94-L97

torchvision/prototype/transforms/_meta.py
datumbox: I absolutely love this and 100% agree we should do it. I've previously asked @vfdev-5 to create an issue and see which methods of the low-level kernels should be expanded to receive a max_value. We can write this in a BC manner on the low-level kernels; if the `max_value` is not passed we do the previous behaviour. `Tensors` will continue making this assumption, but `Images` shouldn't.

I feel this is a critical part of the proposal that needs to be investigated prior implementing all other transforms. This is probably one of the few remained uninvestigated bits of the proposal.

torchvision/prototype/features/_bounding_box.py
pmeier: >  Note that we should avoid the use of strings in favour of their enums in our examples and internal code to get better static analysis.

Internally I agree 100%, but do we actually want to discourage users from using them in application code? `"xyxy"` is more concise than `BoundingBoxFormat.XYXY` and potentially needs one less import.

torchvision/prototype/transforms/_augment.py
datumbox: We need to decide here if we want BC or not. You can't adopt this proposal without offering a new API that achieves a minimum BC-level. Personally I don't favour the verbosity that this brings and how trivial some of the transforms become. For example what's the use of `HorizontalFlip` if it always flips? Why not use a lambda + the underlying kernel? You can argue that this would be even more composable and have even less classes to maintain.

Personally I think to get the maximum chances of rolling out the transforms successfully and getting adoption is to offer straight replacements of the existing classes and limited-time support for newly deprecated classes (with clear messaging on how to move out of them). Think of it as having a limited budget of breakages you can make. Would you spend it on RandomHorizontalFlip or on fixing the ToTensor and eliminating the use of ConvertImageDtype? I would definitely choose the latter.

torchvision/prototype/transforms/_auto_augment.py
vfdev-5: Same for me, no strong option, the implementation can stay as-is

torchvision/prototype/features/_bounding_box.py
datumbox: We have previously made the decision to move with enums (you can see this on multiple APIs including io, multi-weights etc). For consistency we should continue this here.

Where we agree is that people should be allowed to pass strings if they want to. I want to discourage them from doing so by offering all the examples an internal code using enums but if they choose to pass strings so be it. I just don't want to promote this usage. Note that this is true for virtually all places where we have enums.

torchvision/prototype/transforms/_auto_augment.py
datumbox: Let's keep it then. Thanks both for the input.

torchvision/prototype/transforms/_transform.py
pmeier: IIUC, the use case is to transform multiple images with the same randomly sampled parameters, correct? If yes, this already works:

https://github.com/pytorch/vision/blob/95d418970e6dbf2e4d928a204c4e620da7bccdc0/torchvision/prototype/transforms/_transform.py#L22-L23

We sample the parameters once for the complete input and afterwards apply the transformation with these sampled parameters to all inputs individually. For example, let's say you want to randomly rotate two images the same way:

```py
transforms = RandomRotate(...)

image1, image2 = transform(image1, image2)
```

This of course also means we need no extra handling for image pair datasets:

```py
transform = RandomRotate(...)
dataset = datasets.load("sintel").map(transform)

for sample in dataset:
    ...
```

NicolasHug: Is it still supported when images are not available at the same time? i.e.

```
t = SomeRandomTransform(...)
t(img1)
t(img2)
...
t(imgn)
```

?

torchvision/prototype/transforms/functional/_misc.py
vfdev-5: @datumbox Let's say there are two cases we would like to cover (for instance without implementing `val_range`).
Input is PIL and its mode can be `"RGB"` or `"F"`.

```python
t_img = to_tensor(img)
# If RGB (uint8) -> [0-1], f32 tensor (rescaled by 255.0)
# If F (float32) -> f32 tensor (no rescale)

# t_img is float32 tensor
output = F_t.gaussian_blur(t_img, kernel_size, sigma)
# output is float32 tensor

output_pil = to_pil_image(output)
# No matter the input type output_pil is
# RGB uint8 Pillow image
# There is a bug here as we should pass origin mode
# It should be output_pil = to_pil_image(output, mode=img.mode)
```

> Hence I think we could avoid this by ensuring the image is of uint8 type prior using to_pil_image

If input is float32, conversion to uint8 may lead to precision reduction, I think.

Concerning the implementation, we can either make `to_tensor/to_pil_image` private and thus do not expose in the new API. Another solution could be to code PIL->Tensor->PIL explicitly:

```
import torch

def foo(img):
    t_img = torch.from_numpy(np.array(img))
    do_squeeze = False
    if t_img.ndim == 2:
        do_squeeze = True
        t_img = t_img.unsqueeze(-1)
    t_img = t_img.permute(2, 0, 1).contiguous()
    t_img = t_img.to(torch.float32)

    t_out = F.gaussian_blur(t_img, kernel_size=(3, 3), sigma=(0.5, 0.5))

    np_out = t_out.cpu().numpy()
    np_out = np.transpose(np_out, (1, 2, 0))
    if do_squeeze:
        np_out = np_out[:, :, 0]
    return PIL.Image.fromarray(np_out, mode=img.mode)
```

datumbox: @vfdev-5 What I had in mind was to do:
```python
t_img = F.pil_to_tensor(img)
output = F.gaussian_blur(t_img, (3, 3), [1.0, 1.0])
output_pil = to_pil_image(output)
```

An alternative if we are afraid of loss of precision could be:
```python
t_img = F.pil_to_tensor(img).div(255.0)
output = F.gaussian_blur(t_img, (3, 3), [1.0, 1.0])
output_pil = to_pil_image(output)
```

I just don't want to use `to_tensor` on the new API. Ideally we should deprecate and throw warnings if someone uses it.

torchvision/prototype/transforms/_meta.py
datumbox: FYI #5502

torchvision/prototype/transforms/_transform.py
datumbox: @NicolasHug Hmmm... No it's not. Can you describe more of why this is the case?

torchvision/prototype/features/_bounding_box.py
pmeier: As discussed offline, quietly supporting strings is still a good idea, but everything else like documentation or examples we should use the enums.

torchvision/prototype/transforms/_transform.py
NicolasHug: In general, *some* users may need to call the exact same transform (with the same randomly-sampled parameters) on all the batches during a single iteration. And all the batches aren't available at the same time.

However as we just discussed offline, I don't have a clear, immediate use-case for these, apart from the link to the PR above. In practice the one use-case that I encountered (the RAFT thing above) is properly covered by the design as @pmeier indicated https://github.com/pytorch/vision/pull/5500#discussion_r816615926 .

Since our current transforms don't support that anyway, we're not introducing any regression. And since most of the details are kept private, we should be able to support it in some way in the future, if we ever find a stronger need for it.

But for now I'm happy to keep things as they are.

torchvision/prototype/transforms/_augment.py
pmeier: As discussed offline, even if we want this change, it has a low priority. I'm going to remove the comments.

torchvision/prototype/features/_image.py
vfdev-5: Paletted images can be also used to store segmentation masks e.g. VOC

torchvision/prototype/transforms/functional/_misc.py
pmeier: Addressed in #5503.

torchvision/prototype/transforms/functional/__init__.py
```diff
@@ -63,3 +69,13 @@
 )
 from ._misc import normalize_image_tensor, gaussian_blur_image_tensor
 from ._type_conversion import decode_image_with_pil, decode_video_with_av, label_to_one_hot
+
+# What are the migration plans for public methods without new API equivalents? There are two categories:
```
datumbox, line 73: @pmeier Reminder for migration plan on these as well. Might be already on your radar as part of the Transform Classes.

PR #5503, opened 2022-03-01
Recoded gaussian blur using pil_to_tensor instead of to_tensor

vfdev-5:
Description:
- Recoded gaussian blur using pil_to_tensor instead of to_tensor
- Additionally, fixed docs for resize/antialias option

### DIFF
--- a/torchvision/prototype/transforms/functional/_misc.py
+++ b/torchvision/prototype/transforms/functional/_misc.py
@@ -3,7 +3,7 @@ from typing import Optional, List
 import PIL.Image
 import torch
 from torchvision.transforms import functional_tensor as _FT
-from torchvision.transforms.functional import to_tensor, to_pil_image
+from torchvision.transforms.functional import pil_to_tensor, to_pil_image
 
 
 normalize_image_tensor = _FT.normalize
@@ -39,4 +39,6 @@ def gaussian_blur_image_tensor(
 
 
 def gaussian_blur_image_pil(img: PIL.Image, kernel_size: List[int], sigma: Optional[List[float]] = None) -> PIL.Image:
-    return to_pil_image(gaussian_blur_image_tensor(to_tensor(img), kernel_size=kernel_size, sigma=sigma))
+    t_img = pil_to_tensor(img)
+    output = gaussian_blur_image_tensor(t_img, kernel_size=kernel_size, sigma=sigma)
+    return to_pil_image(output, mode=img.mode)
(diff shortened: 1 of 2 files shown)


torchvision/transforms/functional.py
```diff
@@ -126,6 +126,9 @@ def to_tensor(pic):
 
     See :class:`~torchvision.transforms.ToTensor` for more details.
 
+    .. warning::
+        This method is deprecated, please use :meth:`~torchvision.transforms.functional.pil_to_tensor`.
+
```
datumbox, line 131: Are we truly ready to deprecate this or should we wait until the new Transforms API is finalize? Note that we have been discouraging people from using it by eliminating its use internally and removing it from the references in favour of `pil_to_tensor`. Unfortunately though that is not a drop-in replacement and you need 2 calls. One for `pil_to_tensor` and for `convert_image_dtype` and even then those are not directly equivalents.

When we do deprecate it, we should throw warnings. For now I would just remove this.

torchvision/transforms/transforms.py
```diff
@@ -114,6 +114,9 @@ class ToTensor:
 
     In the other cases, tensors are returned without scaling.
 
+    .. warning::
+        This method is deprecated, please use :meth:`~torchvision.transforms.functional.pil_to_tensor`.
+
```
datumbox, line 119: Do you probably meant `PILToTensor`? But again I would just hold off to it.

torchvision/prototype/transforms/functional/_misc.py
```diff
@@ -39,4 +39,6 @@ def gaussian_blur_image_tensor(
 
 
 def gaussian_blur_image_pil(img: PIL.Image, kernel_size: List[int], sigma: Optional[List[float]] = None) -> PIL.Image:
-    return to_pil_image(gaussian_blur_image_tensor(to_tensor(img), kernel_size=kernel_size, sigma=sigma))
+    t_img = pil_to_tensor(img)
+    output = gaussian_blur_image_tensor(t_img, kernel_size=kernel_size, sigma=sigma)
+    return to_pil_image(output, mode=img.mode)
```
datumbox, line 44: This looks good but shouldn't we also do this on the functional on main area?
https://github.com/pytorch/vision/blob/095437aad6cfd25b01fa01961e5774904d738347/torchvision/transforms/functional.py#L1341-L1343

datumbox (commented):
@vfdev-5 thanks for the PR. Just a couple of comments, let me know what you think!

vfdev-5:
@datumbox thanks for the review ! Sounds good to me to remove the warnings and just scope for gaussian blur update.

datumbox (approved):
LGTM, thanks!

Issue #5146, opened 2021-12-30
Explore the possibility of using cuda device instead of device index


prabhat00155:
https://github.com/pytorch/vision/pull/5019#discussion_r776813367

PR #5505, opened 2022-03-01
Allow cuda device to be passed without the index for GPU decoding

prabhat00155:
Resolves https://github.com/pytorch/vision/issues/5146.


datumbox (approved):
LGTM, thanks!

PR #5512, opened 2022-03-02
improve StrEnum

pmeier:
This adds a `from_str` class method to `StrEnum` to clean up the interface. This seems to be fully scriptable

```py
import torch
from torchvision.prototype.utils._internal import StrEnum


class Foo(StrEnum):
    BAR = StrEnum.auto()
    BAZ = 42


@torch.jit.script
def check_str_enum(input: torch.Tensor, foo: Foo) -> torch.Tensor:
    if foo == Foo.BAR:
        return input.abs()
    elif foo == Foo.BAZ:
        return input + 1
    else:
        raise ValueError("This should not happen!")


input = torch.tensor(-1)

print("should be 1: ", check_str_enum(input, Foo.BAR))

print("should be 0: ", check_str_enum(input, Foo.from_str("BAZ")))

check_str_enum(input, Foo.from_str("CAR"))
```

```
should be 1:  tensor(1)
should be 0:  tensor(0)
ValueError: Unknown value 'CAR' for Foo. Did you mean 'BAR'?
```


torchvision/prototype/utils/_internal.py
```diff
@@ -48,8 +48,17 @@
 class StrEnumMeta(enum.EnumMeta):
     auto = enum.auto
 
-    def __getitem__(self, item):
-        return super().__getitem__(item.upper() if isinstance(item, str) else item)
+    def from_str(self, member: str):
```
pmeier, line 51: Meta classes are fairly advanced Python. This is what you get when using `type(self)` of an instance of a class that is using this as meta class. This means, `self` is a subclass of `StrEnum` here. Thus, we don't use the `@classmethod` decorator here, since this is already a class method.

torchvision/prototype/models/_api.py
```diff
@@ -34,7 +35,7 @@ class Weights:
     meta: Dict[str, Any]
 
 
-class WeightsEnum(Enum):
+class WeightsEnum(StrEnum):
```
datumbox, line 38: I'm about to move the Models prototype to the main area. We can't depend on an `StrEnum` that is on the prototype area here.

torchvision/prototype/utils/_internal.py
datumbox: I was about to comment on that. So we will be able to use it as class instance, right? See:
https://github.com/pytorch/vision/blob/93c85bbcc31f8d5a052daf06f2f91f39697af1a4/torchvision/prototype/models/_api.py#L50-L54

torchvision/prototype/utils/_internal.py
```diff
@@ -48,8 +48,17 @@
...
+        try:
+            return self[member]
```
datumbox, line 53: Does this work with aliases like the `DEFAULT` records of Weights?

datumbox (changes requested):
Looks good overall but it's blocked by the dependency on the internal of prototype. See comment below. Thoughts?

torchvision/prototype/models/_api.py
pmeier: True. Any reservations of using `StrEnum` also to the main area? I would put it into a private utils module like `torchvision._utils`.

torchvision/prototype/utils/_internal.py
pmeier: Yes. Nothing changes for the subclass.

torchvision/prototype/utils/_internal.py
pmeier: Confirmed locally and CI seems to be happy too.

datumbox (approved):

PR #5513, opened 2022-03-02
port FiveCrop and TenCrop to prototype API

pmeier:
Both transforms are outliers in the current stable API as discussed in https://github.com/pytorch/vision/pull/5500#discussion_r816507761. Even if we want to deprecate them in favor of a different approach eventually, we need to support them at day 0 of the roll out. 

The issues with these transforms arise, because they always need to be followed by another to batch them together:

https://github.com/pytorch/vision/blob/71d2bb0bc67044f55d38bfddf04e05be0343deab/torchvision/transforms/transforms.py#L980-L984

We cannot re-use this lambda approach since crops might be buried inside a nested container. Thus, we need a custom transformation that picks up on the result of such a multi crop transform and batches it. To communicate which elements need to be batched between the two transformed, I've opted to wrap the return values of the multi crop kernels in named tuples, because they are full JIT scriptable. The old `FiveCrop` returned a regular tuple, so this is 100% compatible. `TenCrop` returned a list, but this is still fine unless someone is using list specific methods like append or the like.

Example usage:

```py
from torchvision.prototype import datasets, transforms, features

    transforms.DecodeImage(),
    transforms.TenCrop(16),
    transforms.BatchMultiCrop(),
)

for sample in datasets.load("imagenet").map(transform):
    print(sample["image"].shape)
    break
```

```
torch.Size([10, 3, 16, 16])
```

To test with PIL images, insert `transforms.Lambda(to_pil_image, features.Image)` after the decoding.


torchvision/prototype/utils/_internal.py
```diff
@@ -301,13 +302,42 @@ def read(self, size: int = -1) -> bytes:
...
+    obj: Any,
+    *,
+    include_sequence_types: Collection[Type] = (collections.abc.Sequence,),
     # We explicitly exclude str's here since they are self-referential and would cause an infinite recursion loop:
     # "a" == "a"[0][0]...
-    if isinstance(obj, collections.abc.Sequence) and not isinstance(obj, str):
-        return [apply_recursively(fn, item) for item in obj]
-    elif isinstance(obj, collections.abc.Mapping):
-        return {key: apply_recursively(fn, item) for key, item in obj.items()}
+    exclude_sequence_types: Collection[Type] = (str,),
```
pmeier, line 312: We need this addition to be able to exclude named tuples as sequences in the `BatchMultiCrop` transform. My gut says we are going to need this fine grained control again for other transforms that are not yet ported / implemented. If that turns out not to be true, I'm happy to simply implement a custom solution only in the `BatchMultiCrop` transform given that we probably deprecate it anyway.

torchvision/prototype/transforms/_geometry.py
```diff
@@ -168,3 +171,77 @@ def forward(self, *inputs: Any) -> Any:
...
+            return batch
+        else:
+            return input
+
+    def forward(self, *inputs: Any) -> Any:
+        sample = inputs if len(inputs) > 1 else inputs[0]
+        return apply_recursively(
+            functools.partial(self._transform, params=self._get_params(sample)),
+            sample,
+            exclude_sequence_types=(str, *self._MULTI_CROP_TYPES),
```
pmeier, line 246: We need this exclude here, because named tuples by default would be recognized as sequence and thus we would only get the individual elements rather than everything at once.

torchvision/prototype/transforms/_geometry.py
```diff
@@ -168,3 +171,77 @@ def forward(self, *inputs: Any) -> Any:
...
+        self.vertical_flip = vertical_flip
+
+    def _transform(self, input: Any, params: Dict[str, Any]) -> Any:
+        if isinstance(input, features.Image):
+            output = F.ten_crop_image_tensor(input, self.size, vertical_flip=self.vertical_flip)
+            return F._TenCropResult(*[features.Image.new_like(input, o) for o in output])
+        elif type(input) is torch.Tensor:
+            return F.ten_crop_image_tensor(input, self.size)
+        elif isinstance(input, PIL.Image.Image):
+            return F.five_crop_image_pil(input, self.size)
```
pmeier, line 212:
```suggestion
            return F.ten_crop_image_pil(input, self.size)
```

torchvision/prototype/transforms/_geometry.py
```diff
@@ -168,3 +171,77 @@ def forward(self, *inputs: Any) -> Any:
...
+class FiveCrop(Transform):
+    def __init__(self, size: Union[int, Sequence[int]]) -> None:
+        super().__init__()
+        self.size = _setup_size(size, error_msg="Please provide only two dimensions (h, w) for size.")
+
+    def _transform(self, input: Any, params: Dict[str, Any]) -> Any:
+        if isinstance(input, features.Image):
+            output = F.five_crop_image_tensor(input, self.size)
+            return F._FiveCropResult(*[features.Image.new_like(input, o) for o in output])
+        elif type(input) is torch.Tensor:
```
datumbox, line 185: This would require updating it according to #5518 once merged.

torchvision/prototype/utils/_internal.py
datumbox: We should start from the assumption that this is not needed and add it later. Starting with a simple solution first is a good prior. Please simplify as much as possible.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -314,3 +314,96 @@ def resized_crop_image_pil(
 ) -> PIL.Image.Image:
     img = crop_image_pil(img, top, left, height, width)
     return resize_image_pil(img, size, interpolation=interpolation)
+
+
+class _FiveCropResult(NamedTuple):
```
datumbox, line 319: Why introduce a new private named duple? I think we can keep the simple tuple approach currently used on main:
https://github.com/pytorch/vision/blob/b8de0b84c2382ce237f801e141d954c1deaa090c/torchvision/transforms/functional.py#L751

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -314,3 +314,96 @@ def resized_crop_image_pil(
...
+    tl = crop_image_pil(img, 0, 0, crop_height, crop_width)
+    tr = crop_image_pil(img, 0, image_width - crop_width, crop_height, crop_width)
+    bl = crop_image_pil(img, image_height - crop_height, 0, crop_height, crop_width)
+    br = crop_image_pil(img, image_height - crop_height, image_width - crop_width, crop_height, crop_width)
+    center = center_crop_image_pil(img, [crop_height, crop_width])
+
+    return _FiveCropResult(tl, tr, bl, br, center)
+
+
+class _TenCropResult(NamedTuple):
```
datumbox, line 373: Same here.

torchvision/prototype/transforms/_geometry.py
datumbox: I think that's one more reason not to use named tuples.

torchvision/prototype/transforms/_geometry.py
```diff
@@ -168,3 +171,77 @@ def forward(self, *inputs: Any) -> Any:
...
+            return input
+
+    def forward(self, *inputs: Any) -> Any:
+        sample = inputs if len(inputs) > 1 else inputs[0]
+        if has_any(sample, features.BoundingBox, features.SegmentationMask):
+            raise TypeError(f"BoundingBox'es and SegmentationMask's are not supported by {type(self).__name__}()")
+        return super().forward(sample)
+
+
+class BatchMultiCrop(Transform):
```
datumbox, line 223: I think creating such new transform should be beyond the scope of this PR. Let's remove and introduce only if necessary on the future.

datumbox (commented):
@pmeier As we discussed previously offline, the specific transforms are a bit problematic as they don't fit nicely on the new API. 
The only reason why we provide support for them is for BC, so I would avoid introducing any new features. Let's offer the simplest possible solution just to ensure we won't break anyone's code.

pmeier:
After some offline discussion @datumbox and I agreed that we indeed need the `BatchMultiCrop` transform. Nevertheless, with the new commits the implementation is a lot simply, by moving the "result wrapper" onto the transformations and being able to use only a single subtype of list rather than two namedtuples that would have been necessary to satisfy torchscript. Plus, instead of making `apply_recursively` more general, `BatchMultiCrop` now implements a custom version of it to keep it simple.

datumbox (approved):
LGTM.

I think we should also review on the near future whether the very detailed nested `sample` is something we need to support or if that's something we don't need on the Datasets. If it turns out it's not needed it will allow us to massively simplify the transforms.

cc @NicolasHug

PR #5518, opened 2022-03-03
fix vanilla tensor image detection

pmeier:
We missed that in #5421. Although we often do `isinstance(input, torch.Tensor)` in the stable implementation, we can no longer do that for the prototypes. Since `features._Feature` is also a `torch.Tensor`, everything not explicitly handled before the `isinstance(input, torch.Tensor)` condition would be treated as image. For example

```py

sample = dict(label=features.Label(0))
transform = transforms.Resize((32, 32))

transform(sample)
```

```
TypeError: Tensor is not a torch image.
```


torchvision/prototype/transforms/_utils.py
```diff
@@ -46,3 +46,7 @@ def has_any(sample: Any, *types: Type) -> bool:
 
 def has_all(sample: Any, *types: Type) -> bool:
     return not bool(set(types) - set(_extract_types(sample)))
+
+
+def is_tensor_image(input: Any) -> bool:
```
datumbox, line 51: I find the naming a bit misleading. `features.Image` is a `Tensor`. We also use the word `*_image_tensor` in the names of the kernels. Perhaps we should avoid naming it like that.

How about `is_pure_tensor` or `is_simple_tensor` or something that doesn't use the words `image` and `tensor` together.

pmeier: Agreed, name needs to change. I've used "vanilla tensor" before. Are you ok with that? Otherwise I would prefer "simple tensor". 

Not that all names are not capturing the essence well. This checks for any tensor that is not a feature. Thus, other tensor subclasses are still allowed. If a user has something like `class MyTensor(torch.Tensor)` that is not a pure / simple / vanilla tensor, but will still be accepted here. The same is true for our current stable API.

datumbox: Let's do simple tensor. The method is private anyway so we will be able to refactor it once we review the names as a whole.

datumbox (approved):

PR #5539, opened 2022-03-03
simplify Feature implementation

pmeier:
This removes the  `__init_subclass__()`  and `_to_tensor()` methods of `_Feature`. `__init_subclass__()` was inspecting the MRO and dynamically setting attribute of class instances through the `_META_ATTRS` and `_metadata` attributes which are now removed. Instead, we now add `new_like()` class method to all _Feature  subclasses: slightly more verbose, but much simpler.

### DIFF
--- a/torchvision/prototype/features/_bounding_box.py
+++ b/torchvision/prototype/features/_bounding_box.py
@@ -22,19 +22,39 @@ class BoundingBox(_Feature):
         cls,
         data: Any,
         *,
-        dtype: Optional[torch.dtype] = None,
-        device: Optional[torch.device] = None,
         format: Union[BoundingBoxFormat, str],
         image_size: Tuple[int, int],
+        dtype: Optional[torch.dtype] = None,
+        device: Optional[Union[torch.device, str, int]] = None,
+        requires_grad: bool = False,
     ) -> BoundingBox:
-        bounding_box = super().__new__(cls, data, dtype=dtype, device=device)
+        bounding_box = super().__new__(cls, data, dtype=dtype, device=device, requires_grad=requires_grad)
 
         if isinstance(format, str):
             format = BoundingBoxFormat.from_str(format.upper())
+        bounding_box.format = format
 
-        bounding_box._metadata.update(dict(format=format, image_size=image_size))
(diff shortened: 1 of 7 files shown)


torchvision/prototype/features/_bounding_box.py
```diff
@@ -22,20 +22,40 @@ def __new__(
...
 
+    @classmethod
+    def new_like(
+        cls,
+        other: BoundingBox,
+        data: Any,
+        *,
+        format: Optional[Union[BoundingBoxFormat, str]] = None,
+        image_size: Optional[Tuple[int, int]] = None,
+        **kwargs: Any,
```
pmeier, line 49: Do we want to be explicit here, i.e. taking `dtype`, `device`, and `requires_grad` explicitly?

torchvision/prototype/features/_feature.py
```diff
@@ -8,59 +8,22 @@
...
-            device = torch.device(device)
-        feature = cast(
+        return cast(
             F,
             torch.Tensor._make_subclass(
                 cast(_TensorBase, cls),
-                cls._to_tensor(data, dtype=dtype, device=device),
-                # requires_grad
-                False,
+                torch.as_tensor(data, dtype=dtype, device=device),  # type: ignore[arg-type]
```
pmeier, line 23: As discussed offline, the annotations of `as_tensor` are wrong. `str` and `int` are valid types for the `device`.

pmeier (commented):
Apart from two previously undetected typing errors that were surfaced by this patch, all changes only happen in the `prototype.features` module. Given that all tests are passing and `mypy` is also happy, I'm guessing that this change is indeed functional equivalent.

NicolasHug (approved):
Thanks @pmeier , LGTM. Would you mind adding a description to the PR, something like (feel free to edit)

> This removes the  `__init_subclass__()`  and `_to_tensor()` methods of `_Feature`. `__init_subclass__()` was inspecting the mro and dynamically setting attribute of class instances through the `_META_ATTRS` and `_metadata` attributes which are now removed. Instead, we now add `new_like()` class method to all _Feature  subclasses: slightly more verbose, but much simpler.

PR #5540, opened 2022-03-03
Turn warnings in prototype datasets tests into errors

pmeier:
I noticed [a couple of warnings](https://app.circleci.com/pipelines/github/pytorch/vision/15279/workflows/ec045870-0fba-4834-be23-199f20ba074e/jobs/1233674) in the prototype datasets test suite:

```
test/test_prototype_builtin_datasets.py: 341 warnings
  /home/circleci/.local/lib/python3.7/site-packages/torch/utils/data/datapipes/utils/common.py:110: DeprecationWarning: _DataPipeMeta and its functional API are deprecated and will be removed from the package `torch`. Please use ZipArchiveLoader instead.
    new_name_statement, DeprecationWarning)

test/test_prototype_builtin_datasets.py: 860 warnings
  /home/circleci/.local/lib/python3.7/site-packages/torch/utils/data/datapipes/utils/common.py:110: DeprecationWarning: _DataPipeMeta and its functional API are deprecated and will be removed from the package `torch`. Please use TarArchiveLoader instead.
    new_name_statement, DeprecationWarning)

test/test_prototype_builtin_datasets.py: 10 warnings
  /home/circleci/.local/lib/python3.7/site-packages/scipy/io/matlab/mio5.py:493: VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray.
    narr = np.asanyarray(source)

test/test_prototype_builtin_datasets.py: 18 warnings
  /home/circleci/project/torchvision/prototype/features/_image.py:38: UserWarning: Unable to guess a specific color space. Consider passing it explicitly.
    warnings.warn("Unable to guess a specific color space. Consider passing it explicitly.")
```

From top to bottom this PR does the following to fix them:

- Replace the deprecated `(Zip|Tar)ArchiveReader` with `(Zip|Tar)ArchiveLoader`. This is related to the first two warnings although it is not immediately clear. See pytorch/data#322 for details.
- Suppress a warning emitted by internals of `scipy.io.savemat` that we have no control over
- Fix an error PCAM dataset that was missed during review that leads to a warning in the test suite.

Additionally, this PR enforces that in the future datasets test are not allowed to raise any warnings.


datumbox (approved):
LGTM, thanks.

test/builtin_dataset_mocks.py
```diff
@@ -470,7 +471,10 @@ def imagenet(info, root, config):
         ]
         num_children = 1
         synsets.extend((0, "", "", "", num_children, [], 0, 0) for _ in range(5))
-        savemat(data_root / "meta.mat", dict(synsets=synsets))
+        with warnings.catch_warnings():
```
NicolasHug, line 474: I'm not sure we should do that, if there is ever an actual deprecation warning for savemat we would miss it

test/test_prototype_builtin_datasets.py
```diff
@@ -35,6 +35,7 @@ def test_coverage():
         )
 
 
+@pytest.mark.filterwarnings("error")
```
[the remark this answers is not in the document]
datumbox: On the other hand, playing devil's advocate here, right now we don't have a good way to detect warnings and fix them. I'm not sure anyone is monitoring them either. So I'll be OK to test this, see how often it produces actionable valid breakage and how often it drives us nuts.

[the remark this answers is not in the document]
pmeier: I'll look into that. Still, we need to at least include `torchdata`, since otherwise the first two warnings would have been allowed although they are clearly for us to act on.

test/builtin_dataset_mocks.py
NicolasHug: > Thus, unless scipy.io.savemat emits a np.VisibleDeprecationWarning it will just be raised as it was before.

Yes, this is what I was concerned about. scipy does raise these warnings on their own utils sometimes (perhaps only on some very specific cases, IDK).  https://github.com/scipy/scipy/search?q=np.VisibleDeprecationWarning

NicolasHug: But I did not realize that this was a test, sorry, I thought this was dataset code.

What tripped me up is that we usually use `@pytest.mark.filterwarnings` to ignore specific warnings in some specific tests, instead of using the `catch_warning` context manager. I assume that because of our custom dataset testing framework, that kind of decorator-based filtering would be tricky to do on a per-dataset basis?

NicolasHug (approved):
Thanks @pmeier ,

pmeier:
If this comes back to bite us (see https://github.com/pytorch/vision/pull/5540#discussion_r835362547): `@pytest.mark.filterwarnings("error:::torch(vision|data)?.*")` is the decorator that will only fail the tests in case the warning is coming from `torch`, `torchvision`, or `torchdata`.

PR #5544, opened 2022-03-03
Disabling smoke test for nightly builds

atalman:
This PR disables the smoke tests. These tests have been broken for a while now and we don't check these tests. 
Here are the results of these tests failing:
https://app.circleci.com/pipelines/github/pytorch/vision?branch=nightly&filter=all

Please note: I simply commented these tests out for now, so we can actually revisit them later when we have velocity to do so. Please comment on the PR if you think we should clean them completely.


datumbox (approved):
LGTM, thanks!

Issue #5541, opened 2022-03-03
accimage not installed in CI skipping a number of tests


prabhat00155:
Context: Internal tests are failing on `test_transforms.py::TestAccImage::test_accimage_crop` with the following error:
```
AttributeError: 'accimage.Image' object has no attribute 'getbands'
```
It points to the following function:
```
def get_dimensions(img: Any) -> List[int]:
        if _is_pil_image(img):
>           channels = len(img.getbands())
E           AttributeError: 'accimage.Image' object has no attribute 'getbands'
```
from `functional_pil.py` which was added in this https://github.com/pytorch/vision/pull/5487.

cc @pmeier

pmeier:
Could you link a CI run so I can see where we need to install it?

prabhat00155:
@pmeier Added some context to this issue. This error was not caught by our CI. Also, looks like `conda install -c conda-forge accimage` from https://github.com/pytorch/accimage#installation doesn't work, so not sure how we would add this to CI.

pmeier:
I'll have a look.

NicolasHug:
This [comment](https://github.com/pytorch/vision/issues/268#issuecomment-775158609) from @fmassa seems to suggest that we actually don't support accimage anymore

> We don't have Travis (nor accimage) tests anymore, closing this.

That being said, we still claim on our README that accimage is part of the supported backends. Let's wait for @fmassa 's insights before we decide to fully remove it, or actually tests against it.

PR #5545, opened 2022-03-03
[FBcode->GH] Fix accimage tests

NicolasHug:
Should fix https://github.com/pytorch/vision/issues/5541


torchvision/transforms/functional_pil.py
```diff
@@ -39,7 +42,12 @@ def get_image_size(img: Any) -> List[int]:
 @torch.jit.unused
 def get_image_num_channels(img: Any) -> int:
     if _is_pil_image(img):
-        return len(img.getbands())
+        if hasattr(img, "getbands"):
+            channels = len(img.getbands())
+        else:
+            return img.channels  # accimage support
```
prabhat00155, line 48: This gives `AttributeError: channels`.

prabhat00155 (changes requested):
Thanks @NicolasHug, this doesn't seem to work. Blocking accidental merge.

torchvision/transforms/functional_pil.py
prabhat00155: I am running the command mentioned in the diff on my devserver.

datumbox:
@NicolasHug @prabhat00155 Thanks for investigating.

I agree with your comments on the issues that we should consider deprecating `accimage`. Until then we can circumvent the problem with 2 ways:
- Restore the previous inaccurate `channels = 1 if img.mode == "L" else 3`. This is what we had [before](https://github.com/pytorch/vision/pull/5487/files#diff-efb615f437f0b00ef2ed8abf2e0eda487cf609899a115bf3fa4691100db581afL33) but it's obviously wrong.
- Replace it with the inefficient `np.array(img).shape[2]`. Given Nicolas' patch it should only execute when `getbands()` is not available.

I've pushed the changes directly to the @NicolasHug branch, I hope you don't mind. I'll leave it to you to merge this or revert the change. Stamping to unblock.

datumbox (approved):

torchvision/transforms/functional_pil.py
NicolasHug: Using `conda install -c conda-forge accimage` from the installation instructions of accimage: https://github.com/pytorch/accimage#installation

NicolasHug:
>  Until then we can circumvent the problem with 2 ways:
> I've pushed the changes directly to the @NicolasHug branch,

Thanks @datumbox , did you check that the proposed solution properly fixes the internal tests? Otherwise we shouldn't merge it

datumbox:
I've checked the internal tests and to me `.channels` works fine. So I restored your approach as it's better:


```
buck test @mode/dev //pytorch/vision/test:torchvision_transforms -- --exact 'pytorch/vision/test:torchvision_transforms - test_transforms.py::TestAccImage::test_accimage_crop'
 ....
    ✓ ListingSuccess: pytorch/vision/test:torchvision_transforms : 1310 tests discovered (22.521)
    ✓ Pass: pytorch/vision/test:torchvision_transforms - test_transforms.py::TestAccImage::test_accimage_crop (0.266)
    ✓ Pass: pytorch/vision/test:torchvision_transforms - main (19.158)
Summary
  Pass: 2
  ListingSuccess: 1
```

prabhat00155 (approved):

PR #5546, opened 2022-03-03
[Feature] Add efficientnet_fpn_backbone

talregev:
Add efficientnet_fpn_backbone


datumbox (commented):
@talregev Thanks for the contribution, welcome to our community.

Though your code on the PR looks good, I'm skeptical merging this for a few reasons:
1. TorchVision focuses on providing canonical implementations. The specific FPN Backbone that you propose is aligned with the ResNet approach but it's not something that is described on a paper.
2. The method is not used by any pre-trained model or model builder. 
3. It's rather a utility that fits better in downstream applications. Even the decision over which backbones should use a FrozenBN is somehow arbitrary and possibly application specific.

Note that writing an issue prior sending a PR is not mandatory, but it can help you avoid doing throwaway work. Have a look for more information [here](https://github.com/pytorch/vision/blob/main/CONTRIBUTING.md#tldr).

If you have concerns about anything I'm mentioning above, please let me know. Thanks!

talregev:
> 1. The specific FPN Backbone that you propose is aligned with the ResNet approach

This is "maskrcnn" to build easily backbond for other networks. 
> 2. The method is not used by any pre-trained model or model builder.

The method used pre-train efficientnet backbond that come with pytorch. you must missed that. 

>3. It's rather a utility that fits better in downstream applications.

It the same as resnet and mobilenet function that help you to build backbond for maskrcnn and other network for detection.

datumbox:
@talregev I'm aware the we offer an Efficientnet backbone; I'm the one who added it. :)

My point is that, even though you follow a "maskrcnn" approach, this doesn't mean it's a canonical implementation of a model described on a paper. Unless you can send us a reference of a specific paper you are implementing and also decide to offer pre-trained weights for that model, then this PR can't be merged. 

I don't doubt that this utility method can be useful for users (this is why you could consider creating a small project for it at Github). But consider what would happen if we try to create methods for virtually every architecture/backbone that TorchVision supports. We would have to add tens of methods, duplicating massive amount of code. It would be nice at one point to offer a utility that creates an FPN out of arbitrary backbones but the main problem is that there are too many ways to build an FPN and each paper tends to do something different. That's the reason we currently don't offer such a general purpose utility.

talregev:
>But consider what would happen if we try to create methods for virtually every architecture/backbone that TorchVision supports

There is only 17 files inside models, with ~15 backbond models. most of them is not used for detection. 
This is a detection utils file. it tend to help to build backbond for detections networks, as the name of the file imply.
We should add as many backbond support for detections, and if we see duplications, we can start refactoring and generalize. 
until it happen, we can another useful support and easy access like mobilnet and resnt that found in the same file.

> Unless you can send us a reference of a specific paper you are implementing 

Not sure I understand this policy. This is raise a difficult questions for the community. 
Are you accepting only novel things that come from papers? 
There is no new things that come from engineer and try new stuff? 

Also there is a paper (not my paper) that use cascade mask-rcnn with efficientnet backbond with NAS-FPN:
https://arxiv.org/pdf/2012.07177v2.pdf

datumbox:
@talregev Please review our [general contribution](https://github.com/pytorch/vision/blob/main/CONTRIBUTING.md) and the [model contribution](https://github.com/pytorch/vision/blob/main/CONTRIBUTING_MODELS.md) guides for details on what kind of improvements we usually accept and what is the recommended approach for maximizing the chances of getting them merged.

As I explained earlier, this utility can't be added in TorchVision without having a specific canonical architecture using it. Moreover, prior implementing new architectures you would need to open an issue and discuss it with the maintainers as per the guide above. I see that you already separated some of the test improvements you made to a different PR at #5552. I'm happy to help you review and merge the other PR but unfortunately this one is not something we can accept. I don't argue that some people will find it useful; I'm saying that this is not the kind of utility we want to add in the library.

PR #5547, opened 2022-03-04
Fix libtorchvision.so not able to encode images by adding *_FOUND macros to CMakeLists.txt

zeruniverse:
When building `libtorchvision.so` for C++ following README

```bash
mkdir build
cd build
# Add -DWITH_CUDA=on support for the CUDA if needed
cmake ..
make
make install
```

I found the built library is not linked to libpng / libjpeg. After some debug, I found `#if *_FOUND` presents in related source files but corresponding macros is not set in CMakeLists.txt.

I have tested that with this PR, the built `libtorchvision.so` successfully links to libpng and libjpeg


NicolasHug (approved):
Thanks @zeruniverse , LGTM

PR #5549, opened 2022-03-04
[DEBUG] eurosat prototype on windows and macos

pmeier:


test/builtin_dataset_mocks.py
```diff
@@ -1329,20 +1329,20 @@ def cub200(info, root, config):
...
+    for category in categories:
         create_image_folder(
             root=data_folder,
-            name=cls,
-            file_name_fn=lambda idx: f"{cls}_{idx}.jpg",
+            name=category,
+            file_name_fn=lambda idx: f"{category}_{idx + 1}.jpg",
             num_examples=num_examples_per_class,
         )
     make_zip(root, "EuroSAT.zip", data_folder)
```
pmeier, line 1344: By using `data_folder` here, it is assumed that this folder and all of its contents should be included in the archive. By default, `make_zip` removes all files and folders it includes. Since the extra `eurosat` folder is not included, it is left as is. Thus, there are now an empty `eurosat` folder as well as the `EuroSAT.zip` file in the root directory.

Although not correct, that wouldn't be an issue if weren't for a quirk on Windows and macOS: `pathlib` treats file and folder names case-insensitive :exploding_head: This means, 

https://github.com/pytorch/vision/blob/e836b3d8baf8cbfdd18fcc872631962345a65689/torchvision/prototype/datasets/utils/_resource.py#L92-L99

is picking up the empty `eurosat` folder as a candidate for `EuroSAT.zip` although it should only look for an `EuroSAT` folder. Since we prefer extracted files over archives, it now loads the empty folder and thus all tests fail.

NicolasHug: > Windows and macOS: pathlib treats file and folder names case-insensitive

Is this a bug, or a documented "feature" ? Are we likely to hit more bugs in the future because of this?

test/builtin_dataset_mocks.py
```diff
@@ -1329,20 +1329,20 @@ def cub200(info, root, config):
 
 @register_mock
 def eurosat(info, root, config):
-    data_folder = pathlib.Path(root, "eurosat", "2750")
+    data_folder = root / "2750"
```
NicolasHug, line 1332: How come the linux tests were passing before with the extra `eurosat` folder?
I would assume that if we change the structure in the , this would have to be reflected somehow in the dataset code?

test/test_prototype_builtin_datasets.py
```diff
@@ -52,6 +52,8 @@ def test_sample(self, test_home, dataset_mock, config):
 
         try:
             sample = next(iter(dataset))
+        except StopIteration:
+            raise AssertionError("Unable to draw any sample.") from None
```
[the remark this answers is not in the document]
pmeier: It makes this case more expressive. Before you got an error message like `Drawing a sample resulted in the error above` while "above" is a plain `StopIteration`. If you are familiar with datapipes, you should be able to interpret this, but new contributors might not be. I've fixed this here, because we were hitting this exact case.

test/builtin_dataset_mocks.py
```diff
@@ -1329,20 +1329,20 @@ def cub200(info, root, config):
...
-    classes = ("AnnualCrop", "Forest")
-    for cls in classes:
+    categories = ["AnnualCrop", "Forest"]
+    for category in categories:
         create_image_folder(
             root=data_folder,
-            name=cls,
-            file_name_fn=lambda idx: f"{cls}_{idx}.jpg",
+            name=category,
+            file_name_fn=lambda idx: f"{category}_{idx + 1}.jpg",
```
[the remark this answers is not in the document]
pmeier: It doesn't, but it aligns the mock data more with real data. I just found this while looking into the real data to debug this issue and didn't want to have an extra PR for it.

test/builtin_dataset_mocks.py
pmeier: Since the prototype datasets never look at the absolute path, but only relative starting from the right

https://github.com/pytorch/vision/blob/27745e5be086ecdf2a8e609d65c13969cb5201f7/torchvision/prototype/datasets/_builtin/eurosat.py#L38

adding more folders on the left side makes no difference.

NicolasHug (approved):
Thanks @pmeier

Issue #5542, opened 2022-03-03
Migrate detection reference transforms to `torchvision.prototype.transforms`


pmeier:
In our detection references we have a couple of transformations that can handle images and bounding boxes simultaneously:

- [x] [`RandomPhotometricDistort`](https://github.com/pytorch/vision/blob/e836b3d8baf8cbfdd18fcc872631962345a65689/references/detection/transforms.py#L227) #5663
- [x] [`RandomZoomOut`](https://github.com/pytorch/vision/blob/e836b3d8baf8cbfdd18fcc872631962345a65689/references/detection/transforms.py#L164) #5551
- [x] [`RandomIoUCrop`](https://github.com/pytorch/vision/blob/e836b3d8baf8cbfdd18fcc872631962345a65689/references/detection/transforms.py#L77) @vfdev-5 https://github.com/pytorch/vision/pull/6401
- [x] [`ScaleJitter`](https://github.com/pytorch/vision/blob/e836b3d8baf8cbfdd18fcc872631962345a65689/references/detection/transforms.py#L287) @pmeier #6411
- [x] [`FixedSizeCrop`](https://github.com/pytorch/vision/blob/90dace8dbce9d59a4216a91c1d1ac6302ce3080f/references/detection/transforms.py#L331) @pmeier #6417
- [x] [`RandomShortestSize`](https://github.com/pytorch/vision/blob/90dace8dbce9d59a4216a91c1d1ac6302ce3080f/references/detection/transforms.py#L406) #6418
- [x] [`SimpleCopyPaste`](https://github.com/pytorch/vision/blob/6ca9c76adb6daf2695d603ad623a9cf1c4f4806f/references/detection/transforms.py#L543) @vfdev-5 https://github.com/pytorch/vision/pull/6451

They never made it into `torchvision.transforms` since have a different call signature than all other transforms there. Now that `torchvision.prototype.transforms` support simultaneous transforming of multiple types, we can migrate them.

cc @vfdev-5 @datumbox @bjuncek

vfdev-5:
Closing this tracker as done

PR #5551, opened 2022-03-04
port RandomZoomOut from detection references to prototype transforms

pmeier:
Addresses #5542.

Example script to be run in `references/detection`:

```py
import torch

from torchvision.prototype import features, transforms as prototype_transforms
from torchvision.transforms.functional import pil_to_tensor, to_pil_image
from torchvision.utils import draw_bounding_boxes

import transforms as reference_transforms
from train import get_dataset

seed = 1234


def draw(sample):
    image, targets = sample
    annotated_image = draw_bounding_boxes(image, targets["boxes"], width=4)
    to_pil_image(annotated_image).show()


dataset, _ = get_dataset(
    name="coco",
    image_set="train",
    transform=None,
    data_path=...,
)

fill = [123.0, 117.0, 104.0]

reference_sample = (
    pil_to_tensor(dataset[0][0]),
    dict(boxes=dataset[0][1]["boxes"]),
)
prototype_sample = (
    features.Image(reference_sample[0].clone()),
    dict(
        boxes=features.BoundingBox(
            reference_sample[1]["boxes"].clone(),
            format="xyxy",
            image_size=reference_sample[0].shape[-2:],
        )
    ),
)


reference_transform = reference_transforms.RandomZoomOut(fill=fill, p=1.0)
torch.manual_seed(seed)
reference_transformed_sample = reference_transform(*reference_sample)
draw(reference_transformed_sample)

prototype_transform = prototype_transforms.RandomZoomOut(fill=fill, p=1.0)
torch.manual_seed(seed)
prototype_transformed_sample = prototype_transform(prototype_sample)
draw(prototype_transformed_sample)

torch.testing.assert_close(prototype_transformed_sample, reference_transformed_sample, atol=0, rtol=0)
```

example


torchvision/prototype/features/_bounding_box.py
```diff
@@ -64,7 +64,7 @@ def to_format(self, format: Union[str, BoundingBoxFormat]) -> BoundingBox:
         from torchvision.prototype.transforms.functional import convert_bounding_box_format
 
         if isinstance(format, str):
-            format = BoundingBoxFormat[format]
+            format = BoundingBoxFormat.from_str(format.upper())
```
pmeier, line 67: I missed that in #5512.

torchvision/prototype/transforms/functional/_meta.py
```diff
@@ -39,10 +39,13 @@ def _xyxy_to_cxcywh(xyxy: torch.Tensor) -> torch.Tensor:
 
 
 def convert_bounding_box_format(
-    bounding_box: torch.Tensor, *, old_format: BoundingBoxFormat, new_format: BoundingBoxFormat
+    bounding_box: torch.Tensor, old_format: BoundingBoxFormat, new_format: BoundingBoxFormat, copy: bool = False
```
pmeier, line 42: Implementing padding for bounding boxes, I realized we should not always copy here. Given that we don't perform an inplace operation, pass through can avoid unnecessary copies. A (future) common idiom is to convert a bounding box into a specific format, perform some operations on it and convert it back. If we set `copy=True` on the first conversion call, we can potentially avoid another copy on the reverse conversion.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -210,6 +210,47 @@ def rotate_image_pil(
 pad_image_tensor = _FT.pad
 pad_image_pil = _FP.pad
 
+
+# TODO: this was copy-pasted from _FT.pad. Use this if _FT.pad is actually defined here
+def _parse_pad_padding(padding: List[int]) -> List[int]:
```
datumbox, line 215: Nit: Why not refactor at `_FT` and import here?

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -314,3 +355,46 @@ def resized_crop_image_pil(
 ) -> PIL.Image.Image:
     img = crop_image_pil(img, top, left, height, width)
     return resize_image_pil(img, size, interpolation=interpolation)
+
+
+def zoom_out_image_tensor(
```
datumbox, line 360: Hmmm.. Does it make sense to create a function equivalent for every possible transform? One can argue that the functional should be kept simple. Perhaps that's something we can leave in the Transform. Thoughts?

torchvision/prototype/transforms/functional/_meta.py
datumbox: Sounds good but I would prefer to be on the safe side and have the default `copy=True`. Then we should turn it off explicitly when "it's OK, I know what I'm doing" situations. This is similar to how `inplace` works in PyTorch. We should also consider renaming this to `inplace` to be consistent with the names.

datumbox (commented):
@pmeier Thanks! Overall looks good, but I got some comments. Let me know your thoughts

torchvision/prototype/transforms/functional/_geometry.py
pmeier: You are probably right. These three kernels are basically thin wrappers around the padding kernels. In fact, everything here should probably be supported by the padding kernels directly. If we don't want to go there right now, I can also move this back onto the transform. This will make the implementation there more elaborate than just a plain if / else, but I would be ok with that for now.

torchvision/prototype/transforms/functional/_meta.py
pmeier: > Sounds good but I would prefer to be on the safe side and have the default `copy=True`. Then we should turn it off explicitly when "it's OK, I know what I'm doing" situations.

:+1:

> This is similar to how `inplace` works in PyTorch. We should also consider renaming this to `inplace` to be consistent with the names.

Note that in the current state this is not a kernel that should have an `inplace` parameter, because it can be a no-op. So a comparison to something like `torch.Tensor.to` is better, which also just returns the input in case nothing needs to be done. 

```py
In [4]: t = torch.empty((), dtype=torch.float)
In [5]: t.to(torch.float) is t
Out[5]: True
In [6]: t.to(torch.int) is t
Out[6]: False
In [7]: t.to(torch.float, copy=True) is t
Out[7]: False
```

torchvision/prototype/transforms/functional/_geometry.py
datumbox: Yeah I agree with that approach. We probably need a definition of what qualifies to be added in the transforms and what should be its own kernel. I suspect that there are other kernels that shouldn't be in functional and instead should move in the Transforms (perhaps Mixup/cutmix). We should review later and somehow try to formalize in the design doc.

torchvision/prototype/transforms/functional/_meta.py
pmeier: We had a quick offline discussion about this. The difference between these two is as follows:
- If an op has a `copy` parameter it might be an actual no-op if `copy=False`. With `copy=True` or if any operation is performed, the user gets a new tensor. The old tensor is never changed inplace.
- If an op has an `inplace` parameter, it will always perform an operation. Whether this operation is performed inplace or on a new tensor is controlled by the flag.

Since the `convert_*` kernels fall in the first category, we are keeping the `copy=True` parameter.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -314,3 +355,46 @@ def resized_crop_image_pil(
...
+    output = pad_image_tensor(image, [left, top, right, bottom], fill=0, padding_mode="constant")
+
+    if not isinstance(fill, (list, tuple)):
+        fill = [fill] * num_channels
+    fill = torch.tensor(fill).to().view(-1, 1, 1)
+
+    output[..., :top, :] = fill
+    output[..., :, :left] = fill
+    output[..., (top + height) :, :] = fill
+    output[..., :, (left + width) :] = fill
```
datumbox, line 380: Shall we move this fix into main `pad_image_tensor` (aka the `_FT.pad`)? This makes sense as multiple other transforms will need padding (for example the `RandomCrop`). If the underlying pad method respects fill, then this workaround won't be needed in all transforms.

To avoid affecting performance, we should apply this workaround only when `any(x != 0 for x in fill)`

pmeier: Prototype for this is in #5568. I'll hold further development on this PR until that one is done.

datumbox: I think we should resume and put the patch on the transforms. Adding it on the functional JIT-scriptable kernel will lead to very brittle code :(

torchvision/prototype/transforms/_geometry.py
```diff
@@ -256,3 +256,84 @@ def apply_recursively(obj: Any) -> Any:
...
+            # PyTorch's pad supports only integers on fill. So we need to overwrite the colour
+            output = F.pad_image_tensor(input, params["padding"], fill=0, padding_mode="constant")
+
+            left, top, right, bottom = params["padding"]
+            fill = torch.tensor(params["fill"], dtype=input.dtype, device=input.device).to().view(-1, 1, 1)
+
+            output[..., :top, :] = fill
+            output[..., :, :left] = fill
+            output[..., -bottom:, :] = fill
+            output[..., :, -right:] = fill
```
datumbox, line 309: What happens here if bottom==0? I recommend porting the transforms faithfully from the references to avoid introducing bugs. Once we have proper tests for them, we can do optimizations such as this one. I believe what you should be doing here is applying the patch ONLY if top/left/bottom/right are positive values but again, it's safer not to do this now.

torchvision/prototype/transforms/_geometry.py
```diff
@@ -256,3 +256,84 @@ def apply_recursively(obj: Any) -> Any:
...
+            )
+        elif isinstance(input, features.BoundingBox):
+            output = F.pad_bounding_box(input, params["padding"], format=input.format)
+
+            left, top, right, bottom = params["padding"]
+            height, width = input.image_size
+            height += top + bottom
+            width += left + right
+
+            return features.BoundingBox.new_like(input, output, image_size=(height, width))
```
datumbox, line 330: This one is also not a faithful reproduction of the original code but I agree with the approach of creating a kernel.

torchvision/prototype/transforms/functional/_meta.py
```diff
@@ -40,10 +40,13 @@ def _xyxy_to_cxcywh(xyxy: torch.Tensor) -> torch.Tensor:
 
 
 def convert_bounding_box_format(
-    bounding_box: torch.Tensor, *, old_format: BoundingBoxFormat, new_format: BoundingBoxFormat
```
datumbox, line 43: Why did you remove the `*`?

torchvision/transforms/functional_tensor.py
```diff
@@ -369,23 +389,7 @@ def pad(img: Tensor, padding: List[int], fill: int = 0, padding_mode: str = "con
...
-        pad_left = pad_right = padding[0]
-        pad_top = pad_bottom = padding[1]
-    else:
-        pad_left = padding[0]
-        pad_top = padding[1]
-        pad_right = padding[2]
-        pad_bottom = padding[3]
-
-    p = [pad_left, pad_right, pad_top, pad_bottom]
+    p = _parse_pad_padding(padding)
```
datumbox, line 392: Nice! I need this as well for one of my transforms. I'll import instead of copy-pasting.

datumbox (commented):
A minor comment for a potential bug. Other than that it looks good.

torchvision/prototype/transforms/functional/_meta.py
pmeier:
```py
import torch


@torch.jit.script
def foo(bar: torch.Tensor, *, baz: bool = True) -> torch.Tensor:
    if baz:
        bar = bar.clone()

    return bar
```

```
torch.jit.frontend.NotSupportedError: Compiled functions can't take variable number of arguments or use keyword-only arguments with defaults
```

:disappointed:

torchvision/prototype/transforms/_geometry.py
pmeier: You are right. I keep forgetting that negative zero is not a thing for indexing. Plus, I agree that if we add checks for `> 0`, we should be able to use negative indexing.

torchvision/prototype/transforms/functional/_meta.py
datumbox: oh nooooo.... :(

datumbox (approved):
LGTM, thanks!

PR #5552, opened 2022-03-04
Improve test of backbone utils

talregev:
Improve test of backbone utils.

### DIFF
--- a/test/test_backbone_utils.py
+++ b/test/test_backbone_utils.py
@@ -7,7 +7,7 @@ import torch
 from common_utils import set_rng_seed
 from torchvision import models
 from torchvision.models._utils import IntermediateLayerGetter
-from torchvision.models.detection.backbone_utils import mobilenet_backbone, resnet_fpn_backbone
+from torchvision.models.detection.backbone_utils import BackboneWithFPN, mobilenet_backbone, resnet_fpn_backbone
 from torchvision.models.feature_extraction import create_feature_extractor, get_graph_node_names
 
 
@@ -19,7 +19,9 @@ def get_available_models():
 @pytest.mark.parametrize("backbone_name", ("resnet18", "resnet50"))
 def test_resnet_fpn_backbone(backbone_name):
     x = torch.rand(1, 3, 300, 300, dtype=torch.float32, device="cpu")
-    y = resnet_fpn_backbone(backbone_name=backbone_name, pretrained=False)(x)
+    model = resnet_fpn_backbone(backbone_name=backbone_name, pretrained=False)
+    assert isinstance(model, BackboneWithFPN)
+    y = model(x)
     assert list(y.keys()) == ["0", "1", "2", "3", "pool"]
 
     with pytest.raises(ValueError, match=r"Trainable layers should be in the range"):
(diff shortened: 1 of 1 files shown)


test/test_backbone_utils.py
```diff
@@ -16,7 +17,7 @@ def get_available_models():
     return [k for k, v in models.__dict__.items() if callable(v) and k[0].lower() == k[0] and k[0] != "_"]
 
 
-@pytest.mark.parametrize("backbone_name", ("resnet18", "resnet50"))
+@pytest.mark.parametrize("backbone_name", resnet.__all__[1:])
```
datumbox, line 20: Adding all possible variants of ResNet is increases the execution time of the text by 12x. All of the variants are constructed using the same class, which minimizes the chance of one of them being incompatible. Please also note that the `resnet_fpn_backbone` method aim to support a handful of ResNet variants which are used in our detection models (not ResNeXt, WideResNets and other architectures that you test now). 

I recommend reverting this change as what we had previously is a good tradeoff between cost/speed and coverage.

test/test_backbone_utils.py
```diff
@@ -28,16 +29,22 @@ def test_resnet_fpn_backbone(backbone_name):
         resnet_fpn_backbone(backbone_name, False, returned_layers=[0, 1, 2, 3])
     with pytest.raises(ValueError, match=r"Each returned layer should be in the range"):
         resnet_fpn_backbone(backbone_name, False, returned_layers=[2, 3, 4, 5])
+    model = resnet_fpn_backbone(backbone_name, False)
+    assert isinstance(model, BackboneWithFPN)
```
datumbox, line 33: I think this assertion is a good addition. To avoid reinstantiating the same variant, you can refactor line 23 above to return the `model` and have access to do the assertion prior the prediction.

test/test_backbone_utils.py
```diff
@@ -28,16 +29,22 @@ def test_resnet_fpn_backbone(backbone_name):
...
 
 
-@pytest.mark.parametrize("backbone_name", ("mobilenet_v2", "mobilenet_v3_large", "mobilenet_v3_small"))
+@pytest.mark.parametrize("backbone_name", mobilenet.mv2_all[1:] + mobilenet.mv3_all[1:])
```
datumbox, line 36: Here you actually get the same models but for consistency and for the sake of defining explicitly what we test, I recommend reverting the change.

datumbox (commented):
Thanks for the contribution @talregev. A few minor comments below but the rest looks good.

test/test_backbone_utils.py
talregev: good catch. I will refactor.

test/test_backbone_utils.py
talregev: I thoughts that tests should be check everting, and not maintain a good ratio of speed and coverage.  
Also when someone will add model to the resnet family, it will check it.

datumbox: I understand. Unfortunately making this test 12x slower is not an option due to the cost. The backbones that are currently passed to this test are those that use the utility so it's a reasonable tradeoff given the constraints. If you revert the two changes I highlights, I think we are in position to merge the PR. Thanks!

talregev: We should add at least resnet101.

talregev: @datumbox
reasons to add resnet101:
- resnet18, resnet 50, and resnet101, is the most common backbond for detection.
- It have better ration of cost/speed and coverage.
- we check 3 classes on mobilenet, and for consistency we check here also 3 classes. 
You cannot beat it with your logic.

talregev: @datumbox Please comment.

datumbox: @talregev As explained earlier, increasing the execution times on CI without a very good reason is not an option. Our team is working hard to reduce the duration of our tests as we must reduce the costs (see #5574 for related efforts). This PR is expected to increase the execution times as it contains extra model initializations. Therefor this PR can't be merged until these extra model initializations are decreased.

talregev: @datumbox I give you 3 reason why, you just Ignore them.
Also, improve test also increasing the execution times. It the fact that you test something, it cost time. 
I remove the resenet101. But I think it should added.

datumbox (approved):
Thanks!

PR #5553, opened 2022-03-04
Remove `to_tensor()` and `ToTensor()` usages

datumbox:
Replace all references of `to_tensor` with `pil_to_tensor()` and `ToTensor()` with `PILToTensor()`.

Also removing an unused `test/sanity_checks.ipynb` notebook.


test/test_transforms.py
```diff
@@ -1893,13 +1890,13 @@ def test_randomperspective():
...
+        tr_img2 = F.pil_to_tensor(F.perspective(tr_img, endpoints, startpoints))
+        tr_img = F.pil_to_tensor(tr_img)
         assert img.size[0] == width
         assert img.size[1] == height
-        assert torch.nn.functional.mse_loss(tr_img, F.to_tensor(img)) + 0.3 > torch.nn.functional.mse_loss(
-            tr_img2, F.to_tensor(img)
-        )
+        assert torch.nn.functional.mse_loss(
+            tr_img.float(), F.pil_to_tensor(img).float()
+        ) + 0.3 > torch.nn.functional.mse_loss(tr_img2.float(), F.pil_to_tensor(img).float())
```
NicolasHug, line 1899: Considering `pil_to_tensor` doesn't scale to [0, 1], is this comparison equivalent to the one we had before?

test/test_transforms.py
```diff
@@ -457,26 +455,25 @@ class TestPad:
...
         assert result.size(2) == width + 2 * padding
         # check that all elements in the padded region correspond
         # to the pad value
-        fill_v = fill / 255
         eps = 1e-5
         h_padded = result[:, :padding, :]
         w_padded = result[:, :, :padding]
-        torch.testing.assert_close(h_padded, torch.full_like(h_padded, fill_value=fill_v), rtol=0.0, atol=eps)
-        torch.testing.assert_close(w_padded, torch.full_like(w_padded, fill_value=fill_v), rtol=0.0, atol=eps)
+        torch.testing.assert_close(h_padded, torch.full_like(h_padded, fill_value=fill), rtol=0.0, atol=eps)
```
NicolasHug, line 475: Does assert_close + eps make sense  here? Looks like the image is now uint8?

NicolasHug (approved):
Thanks @datumbox , I have some question / comment but I'll approve to unlock.

test/test_transforms.py
datumbox: Good call changed to 0.

Issue #5522, opened 2022-03-03
Port `transforms.RandomCrop` to `prototype.transforms`


pmeier:
cc @vfdev-5 @datumbox @bjuncek

lezwon:
hi @pmeier, I'd like to work on this issue.

pmeier:
Sure, go ahead!

vfdev-5:
Done in https://github.com/pytorch/vision/pull/6305

Issue #5521, opened 2022-03-03
Port `transforms.Pad` to `prototype.transforms`


pmeier:
cc @vfdev-5 @datumbox @bjuncek

lezwon:
@pmeier I'd like to take up this issue.

datumbox:
@lezwon Give us a bit more time to iron out some final changes on the API? This is mainly to avoid requiring too many change from you on the PR. Happy to ping you once we are done, what you think?

@pmeier Could you please mark the tracked issues as blocked until we finalize the aforementioned changes? For any PR that is outstanding, we should try and merge. Happy to chat offline too to see what's the best course of action.

PR #5555, opened 2022-03-07
[WIP] 5522 random crop port

lezwon:
This PR ports the RandomCrop transform to prototype.transforms.
Linked issue: https://github.com/pytorch/vision/issues/5522


torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -314,3 +314,52 @@ def resized_crop_image_pil(
 ) -> PIL.Image.Image:
     img = crop_image_pil(img, top, left, height, width)
     return resize_image_pil(img, size, interpolation=interpolation)
+
+
+def random_pad_image_tensor(
```
pmeier, line 319: Why do we need this? Shouldn't `pad_image_tensor` be able to handle this? In general, we don't have kernels for random functions. All randomness should be handled in the transform.

pmeier (commented):
Thanks @lezwon for this PR! I realize you marked this as `[WIP]`, but I have one question that we should answer before you move forward.

torchvision/prototype/transforms/functional/_geometry.py
lezwon: Hi @pmeier, sorry I am a bit new to this repo. The reason I created a new function is that:
1. `random_pad_image_tensor` does not have the same logic as this function. This function takes into account the output shape required and the current image shape. I wasn't sure if this logic should be in `forward`.
2. Seeing the current code, I felt that transform-related code should be in `_geometry` and that there should be separate functions for tensor and PIL.
3. `_get_params` in `RandomCrop` requires the output from this function. So adding this logic within `_transform` wouldn't work as the `params` in `_transform` would not be valid.

Please do provide me with any other approach you have in mind. I could incorporate those changes.

pmeier: Sorry, I misjudged the situation. I was not aware that the forward actually modified the image:

https://github.com/pytorch/vision/blob/b4cb352c586ee6104a79b2d367d019ca480759b3/torchvision/transforms/transforms.py#L663-L674

This makes things more complicated. cc @datumbox for awareness.

I would move this code into `_transform`. Although the structure is the same for all possible types, we still need to call different `pad` kernels. That would be a lot easier if we had the `Pad` transform from #5521 first. This way we could simply substitute `pad_image_*(...)` with `pad(...)` where `pad` is

```py
pad = functools.partial(
    lambda image, padding: Pad(
        padding,
        fill=self.fill,
        padding_mode=self.padding_mode,
    )(image)
)
``` 

and not worry about the dispatch. Thoughts?

pmeier (commented):
> sorry I am a bit new to this repo

Don't be. We are super happy that you contribute and we don't expect that you do a deep dive before starting. 

I elaborated on my earlier comment inline.

torchvision/prototype/transforms/_geometry.py
```diff
@@ -168,3 +168,93 @@ def forward(self, *inputs: Any) -> Any:
...
+        elif is_simple_tensor(sample):
+            sample = F.random_pad_image_tensor(
+                sample,
+                output_size=self.size,
+                image_size=get_image_dimensions(sample),
+                padding=self.padding,
+                pad_if_needed=self.pad_if_needed,
+                fill=self.fill,
+                padding_mode=self.padding_mode,
+            )
```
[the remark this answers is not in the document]
lezwon: Noted. Will try to incorporate this.

torchvision/prototype/transforms/functional/_geometry.py
lezwon: hi @pmeier , I can keep this PR on hold and work on #5521 first, if it helps.

PR #5556, opened 2022-03-07
fail test suite on unexpected passes

pmeier:
I didn't know before that just using adding `@pytest.mark.xfail` to a test will *not* fail the test suite if the test passes unexpectedly. In such a case `pytest` will only report the test as `xpassed`, but I'm guessing no one looks at the actual report if the CI indicator is green.

This PR adds a configuration option to fail by default in such a case. If for some reason we want the other behavior, we can still override it locally with `@pytest.mark.xfail(..., strict=False)`.

### DIFF
--- a/test/test_prototype_builtin_datasets.py
+++ b/test/test_prototype_builtin_datasets.py
@@ -107,14 +107,7 @@ class TestCommon:
 
         next(iter(dataset.map(transforms.Identity())))
 
-    @parametrize_dataset_mocks(
-        DATASET_MOCKS,
-        marks={
-            "cub200": pytest.mark.xfail(
-                reason="See https://github.com/pytorch/vision/pull/5187#issuecomment-1015479165"
-            )
-        },
-    )
+    @parametrize_dataset_mocks(DATASET_MOCKS)
     def test_traversable(self, test_home, dataset_mock, config):
         dataset_mock.prepare(test_home, config)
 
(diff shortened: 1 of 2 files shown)


test/test_prototype_builtin_datasets.py
```diff
@@ -107,29 +107,15 @@ def test_transformable(self, test_home, dataset_mock, config):
 
         next(iter(dataset.map(transforms.Identity())))
 
-    @parametrize_dataset_mocks(
-        DATASET_MOCKS,
-        marks={
-            "cub200": pytest.mark.xfail(
-                reason="See https://github.com/pytorch/vision/pull/5187#issuecomment-1015479165"
-            )
```
pmeier, line 115: I'm guessing this was fixed by pytorch/pytorch#73118. Can you confirm @NivekT?

NicolasHug (approved):
Thanks @pmeier

test/test_prototype_builtin_datasets.py
NivekT: I believe by enabling the serialization of `fork`, it should fix one of the tests but I'm not sure if it is that one specifically. Does any step if that test use `fork`?

NivekT: Also worth mentioning that this other PR also fixed serialization related problems https://github.com/pytorch/pytorch/pull/72896

pmeier: This failed before because the `Demultiplexer` internally used a generator and thus was unable to be pickled after the iteration started.

PR #5559, opened 2022-03-07
Make ScaleJitter proportional

datumbox:
Make the ScaleJitter transform to maintain the ratios of the image during resize


NicolasHug (approved):
Stamping

PR #5562, opened 2022-03-07
Refactor preset transforms

datumbox:
This PR is part of #4679.

Changes the naming of preset transforms from "dataset-specific" to "task-specific".
  - `ImageNetEval` => `ImageClassificationEval`
  - `CocoEval` => `ObjectDetectionEval`
  - `RaftEval` => `OpticalFlowEval`
  - `VocEval` => `SemanticSegmentationEval`
  - `Kinect400Eval` => `VideoClassificationEval`

### DIFF
--- a/torchvision/prototype/transforms/__init__.py
+++ b/torchvision/prototype/transforms/__init__.py
@@ -10,5 +10,4 @@ from ._container import Compose, RandomApply, RandomChoice, RandomOrder
 from ._geometry import HorizontalFlip, Resize, CenterCrop, RandomResizedCrop, FiveCrop, TenCrop, BatchMultiCrop
 from ._meta import ConvertBoundingBoxFormat, ConvertImageDtype, ConvertImageColorSpace
 from ._misc import Identity, Normalize, ToDtype, Lambda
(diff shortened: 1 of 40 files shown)


torchvision/prototype/transforms/__init__.py
```diff
@@ -10,5 +10,4 @@
 from ._geometry import HorizontalFlip, Resize, CenterCrop, RandomResizedCrop, FiveCrop, TenCrop, BatchMultiCrop
 from ._meta import ConvertBoundingBoxFormat, ConvertImageDtype, ConvertImageColorSpace
 from ._misc import Identity, Normalize, ToDtype, Lambda
-from ._presets import CocoEval, ImageNetEval, VocEval, Kinect400Eval, RaftEval
```
datumbox, line 13: Removing them from the `transforms` namespace. These classes are considered private and users shouldn't be initiatlizing them directly. TorchVision models return them as callables and the users can use them but we reserve their right to change them on the future as long as the preset transforms of the pre-trained models remain correct.

jdsgomes (approved):
LGTM thank you

Issue #5523, opened 2022-03-03
Port `transforms.RandomHorizontalFlip` to `prototype.transforms`


pmeier:
cc @vfdev-5 @datumbox @bjuncek

federicopozzi33:
Hi @pmeier,

I'd like to work on this issue.

PR #5563, opened 2022-03-07
port RandomHorizontalFlip to prototype API

federicopozzi33:
Closes: #5523


federicopozzi33:
I'm not sure about testing.
I saw only smoke tests for other transformations and no particular tests for random transformations in `test_prototype_transforms.py`

How are expected to be tested the random transforms?

I would mock the `torch.rand` to control randomness. Another solution could be to add two tests, with p=0 and p=1, and check the expected result.

datumbox:
@federicopozzi33 Very good points. I think we should definitely look for a better testing strategy for the prototype area. I also agree that `SegmentationMask` should be easily supported. Perhaps we should incorporate your proposal in this PR.

@pmeier We shouldn't have both a `HorizontalFlip` and a `RandomHorizontalFlip`. I believe this was previously discussed but missed during the ticket creation (worth checking the existing tickets to avoid other issues)? I think the right approach here is to merge @federicopozzi33's `RandomHorizontalFlip` implementation in the existing `HorizontalFlip` and just rename the class.

pmeier:
To add to Vasilis comments:

> I would mock/patch the `torch.rand` to control randomness. Another solution could be to add two tests, with `p=0` and `p=1`, and check the expected result.

Mocking randomness can become really weird really fast. Since our transformations support the `p` parameter, I would go for the tests you proposed. For good measure we can throw in some smoke test for `p not in {0, 1}`. Not sure if it is worth it though.

> The `_transform` method is not handling inputs of type `features.SegmentationMask`.
> IMO, the horizontal flip is performed the same way for `features.Image` and `features.SegmentationMask`.

I agree with the sentiment that we should support segmentation masks here. They haven't got any love beyond a few examples yet, so there will probably be more transforms were we could add this support.

That being said, I think we need to decide if we want to have a separate kernel for them or not. Right now we have

https://github.com/pytorch/vision/blob/d8654bb0d84fd2ba8b42cd58d881523821a6214c/torchvision/prototype/transforms/functional/_geometry.py#L62-L65

which is a very thin wrapper. Given that on the "tensor level" images and segmentation masks are basically the same and the distinction only comes on the "feature level", I think we should remove the kernel here and handle this on the transform as proposed by @federicopozzi33.

federicopozzi33:
> @federicopozzi33 Very good points. I think we should definitely look for a better testing strategy for the prototype area. I also agree that `SegmentationMask` should be easily supported. Perhaps we should incorporate your proposal in this PR.
> 
> @pmeier We shouldn't have both a `HorizontalFlip` and a `RandomHorizontalFlip`. I believe this was previously discussed but missed during the ticket creation (worth checking the existing tickets to avoid other issues)? I think the right approach here is to merge @federicopozzi33's `RandomHorizontalFlip` implementation in the existing `HorizontalFlip` and just rename the class.

I agree with you. I did as you proposed.

> To add to Vasilis comments:
> 
> > I would mock/patch the `torch.rand` to control randomness. Another solution could be to add two tests, with `p=0` and `p=1`, and check the expected result.
> 
> Mocking randomness can become really weird really fast. Since our transformations support the `p` parameter, I would go for the tests you proposed. For good measure we can throw in some smoke test for `p not in {0, 1}`. Not sure if it is worth it though.

I added some unit tests to test all cases when the transform is performed (p=1.0).
IMO, I would create separate files for each transform, or at least a test class as I did.
I chose to create some helper methods to generate simple data tensors: this should make tests more readable.

One test case is missing, the fallback case, cause it sounds a little strange to me the following behavior:

```
actual = transforms.RandomHorizontalFlip(p=1.0)(1.0)
assert actual == 1.0
```

Anyway, before adding quite similar tests for the missing case (p=0), I would appreciate some feedback.



> > The `_transform` method is not handling inputs of type `features.SegmentationMask`.
> > IMO, the horizontal flip is performed the same way for `features.Image` and `features.SegmentationMask`.
> 
> I agree with the sentiment that we should support segmentation masks here. They haven't got any love beyond a few examples yet, so there will probably be more transforms were we could add this support.
> 
> That being said, I think we need to decide if we want to have a separate kernel for them or not. Right now we have
> 
> https://github.com/pytorch/vision/blob/d8654bb0d84fd2ba8b42cd58d881523821a6214c/torchvision/prototype/transforms/functional/_geometry.py#L62-L65
> 
> which is a very thin wrapper. Given that on the "tensor level" images and segmentation masks are basically the same and the distinction only comes on the "feature level", I think we should remove the kernel here and handle this on the transform as proposed by @federicopozzi33.

Yeah, I agree with you.

test/test_prototype_transforms.py
```diff
@@ -66,10 +66,10 @@ def parametrize_from_transforms(*transforms):
 class TestSmoke:
     @parametrize_from_transforms(
         transforms.RandomErasing(p=1.0),
-        transforms.HorizontalFlip(),
         transforms.Resize([16, 16]),
         transforms.CenterCrop([16, 16]),
         transforms.ConvertImageDtype(),
+        transforms.RandomHorizontalFlip(),
```
datumbox, line 72: Why this change? We should support the idiom of doing the image transforms prior converting the image to floats.

datumbox (commented):
The code looks good to me. I'll leave it to @pmeier to do the final review and approve.

Just one quick nit below:

test/test_prototype_transforms.py
pmeier: This is not a composed transform, but rather a list of transforms to test.

datumbox: lol, I need another coffee

federicopozzi33:
@datumbox, @pmeier 

Let me add tests for p=0, then you can approve it :)

test/test_prototype_transforms.py
```diff
@@ -152,3 +152,49 @@ def test_normalize(self, transform, input):
...
+
+    def expected_tensor(self, dtype: torch.dtype = torch.float32) -> torch.Tensor:
+        return torch.tensor([[[1, 0], [1, 0]], [[0, 1], [0, 1]]], dtype=dtype)
+
+    def test_simple_tensor_p1(self):
+        input = self.input_tensor()
+
+        actual = transforms.RandomHorizontalFlip(p=1.0)(input)
+
+        assert torch.equal(self.expected_tensor(), actual)
```
pmeier, line 169: `torch.equal` will be deprecated by PyTorch core soon. For comparisons, one should use [`torch.testing.assert_close`](https://pytorch.org/docs/master/testing.html#torch.testing.assert_close). By default, it is an equality check for integer inputs and a closeness check for floating point inputs. If you want to enforce equality here (which is correct), you can either use `torch.testing.assert_close(..., atol=0, rtol=0)` or import `from common_utils import assert_equal` that does this for you.

test/test_prototype_transforms.py
```diff
@@ -152,3 +152,49 @@ def test_normalize(self, transform, input):
...
+        assert torch.equal(features.SegmentationMask(self.expected_tensor()), actual)
+
+    def test_features_bounding_box_p1(self):
+        bbox_format = features.BoundingBoxFormat.XYXY
+        image_size = (10, 10)
+        input = features.BoundingBox(torch.tensor([0, 0, 5, 5]), format=bbox_format, image_size=image_size)
+
+        actual = transforms.RandomHorizontalFlip(p=1.0)(input)
+
+        expected = features.BoundingBox(torch.tensor([5, 0, 10, 5]), format=bbox_format, image_size=image_size)
```
pmeier, line 199: You can use the `.new_like()` class method on features to copy the meta data for you.

```suggestion
        expected = features.BoundingBox.new_like(torch.tensor([5, 0, 10, 5]))
```

test/test_prototype_transforms.py
```diff
@@ -152,3 +152,49 @@ def test_normalize(self, transform, input):
...
     def test_random_resized_crop(self, transform, input):
         transform(input)
+
+
+class TestRandomHorizontalFlip:
+    def input_tensor(self, dtype: torch.dtype = torch.float32) -> torch.Tensor:
+        return torch.tensor([[[0, 1], [0, 1]], [[1, 0], [1, 0]]], dtype=dtype)
+
+    def expected_tensor(self, dtype: torch.dtype = torch.float32) -> torch.Tensor:
+        return torch.tensor([[[1, 0], [1, 0]], [[0, 1], [0, 1]]], dtype=dtype)
```
pmeier, line 162: I'm not sure we need specific values here. I think we should be good to have random image inputs and use [`torch.flip`](https://pytorch.org/docs/master/generated/torch.flip.html). I'm aware that this is the only call in the kernel

https://github.com/pytorch/vision/blob/a8bde78130fd8c956780d85693d0f51912013732/torchvision/transforms/functional_tensor.py#L126-L129

but this is also what I would use to produce a expected output if I didn't know the internals of the kernel. cc @NicolasHug for an opinion

pmeier (commented):
Hey @federicopozzi33, I added some more comments.

test/test_prototype_transforms.py
```diff
@@ -152,3 +152,49 @@ def test_normalize(self, transform, input):
...
+        input = features.SegmentationMask(self.input_tensor())
+
+        actual = transforms.RandomHorizontalFlip(p=1.0)(input)
+
+        assert torch.equal(features.SegmentationMask(self.expected_tensor()), actual)
+
+    def test_features_bounding_box_p1(self):
+        bbox_format = features.BoundingBoxFormat.XYXY
+        image_size = (10, 10)
+        input = features.BoundingBox(torch.tensor([0, 0, 5, 5]), format=bbox_format, image_size=image_size)
```
pmeier, line 195: Internally, the input data to the prototype features is fed into [`torch.as_tensor`](https://pytorch.org/docs/master/generated/torch.as_tensor.html)

https://github.com/pytorch/vision/blob/a8bde78130fd8c956780d85693d0f51912013732/torchvision/prototype/features/_feature.py#L23

Thus, you don't need to manually create a tensor here:

```suggestion
        input = features.BoundingBox([0, 0, 5, 5], format=bbox_format, image_size=image_size)
```

torchvision/prototype/transforms/_geometry.py
```diff
@@ -13,11 +13,25 @@
...
+            return sample
+
+        return super().forward(sample)
+
     def _transform(self, input: Any, params: Dict[str, Any]) -> Any:
         if isinstance(input, features.Image):
             output = F.horizontal_flip_image_tensor(input)
             return features.Image.new_like(input, output)
+        elif isinstance(input, features.SegmentationMask):
+            output = F.horizontal_flip_image_tensor(input)
```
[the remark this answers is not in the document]
federicopozzi33: I created the wrapper, which has the following signature:

```python
def horizontal_flip_segmentation_mask(segmentation_mask: torch.Tensor) -> torch.Tensor:
    pass
```

Not sure about `segmenation_mask` type. I put `torch.Tensor` to be consistent with the other existing wrappers, but ... shouldn't be of type `features.SegmentationMask`?

test/test_prototype_transforms.py
federicopozzi33: I used specific values to create the expecting one. I preferred to not use the `torch.flip` to not "mock" the internals of the transformation.

test/test_prototype_transforms.py
```diff
@@ -152,3 +153,56 @@ def test_normalize(self, transform, input):
     )
     def test_random_resized_crop(self, transform, input):
         transform(input)
+
+
+class TestRandomHorizontalFlip:
+    def input_tensor(self, dtype: torch.dtype = torch.float32) -> torch.Tensor:
```
NicolasHug, line 159: @pmeier are we type-checking the tests?

If not, let's please remove all annotations.

test/test_prototype_transforms.py
NicolasHug: > cc @NicolasHug for an opinion

No strong opinion on my side for this specific transform. In some cases it's valuable to have a simple implementation of the transform that we're testing - it helps understanding what's going on with simpler code, and we can call it on arbitrary input. In the case of `flip` the transform is very elementary and doesn't need much explaining anyway.

If we're confident that this hard-coded input covers all of what we might want to test against, then that's fine.

test/test_prototype_transforms.py
pmeier: We are not. That would be madness even for me as `mypy` advocate. Still, do you want to enforce "no annotations" in the test suite? I agree they are not needed here, but for might be helpful for utility functions.

NicolasHug: I would prefer encouraging us not to use them. Type annotations hurt readability.
And when they're not being checked against, they can be wrong, which leads to confusion.

datumbox: I don't know if it hurts readability but it definitely hurts my laziness :D I agree with Nicolas that not having to deal with mypy on tests would be for the best. I would also be in favour of not putting any typing in the test suite.

federicopozzi33: I agree with you that if tests are not type-checked, it's better to avoid type annotation.

So I removed it.

test/test_prototype_transforms.py
```diff
@@ -152,3 +153,56 @@ def test_normalize(self, transform, input):
...
+
+
+class TestRandomHorizontalFlip:
+    def input_tensor(self, dtype=torch.float32):
+        return torch.tensor([[[0, 1], [0, 1]], [[1, 0], [1, 0]]], dtype=dtype)
+
+    def expected_tensor(self, dtype=torch.float32):
+        return torch.tensor([[[1, 0], [1, 0]], [[0, 1], [0, 1]]], dtype=dtype)
+
+    @pytest.mark.parametrize("p", [0.0, 1.0], ids=["p=0", "p=1"])
```
pmeier, line 165: 1. For builtin types like `int`, `float`, `str`, you don't need to put a manual id. `pytest` handles it for you.

    ```suggestion
        @pytest.mark.parametrize("p", [0.0, 1.0])
    ```

2. IIUC, you want this parametrization on every test in this class. If that is the case, you can simply decorate the class rather than every test individually.

test/test_prototype_transforms.py
```diff
@@ -152,3 +153,56 @@ def test_normalize(self, transform, input):
...
+        assert_equal(features.SegmentationMask(expected), actual)
+
+    @pytest.mark.parametrize("p", [0.0, 1.0], ids=["p=0", "p=1"])
+    def test_features_bounding_box(self, p):
+        input = features.BoundingBox([0, 0, 5, 5], format=features.BoundingBoxFormat.XYXY, image_size=(10, 10))
+
+        actual = transforms.RandomHorizontalFlip(p=p)(input)
+
+        expected = torch.tensor([5, 0, 10, 5]) if p == 1.0 else input
+        assert_equal(features.BoundingBox.new_like(input, expected), actual)
```
pmeier, line 208: Could you also check that the meta data of the feature stays the same?

```suggestion
        assert_equal(features.BoundingBox.new_like(input, expected), actual)
        assert actual.format == expected.format
        assert actual.image_size == expected.image_size
```

? 

I'm currently working on adding such functionality to assert equal, so we don't have to do it manually. But for now we should make sure here.

test/test_prototype_transforms.py
```diff
@@ -152,3 +153,56 @@ def test_normalize(self, transform, input):
...
     def test_random_resized_crop(self, transform, input):
         transform(input)
+
+
+class TestRandomHorizontalFlip:
+    def input_tensor(self, dtype=torch.float32):
+        return torch.tensor([[[0, 1], [0, 1]], [[1, 0], [1, 0]]], dtype=dtype)
+
+    def expected_tensor(self, dtype=torch.float32):
+        return torch.tensor([[[1, 0], [1, 0]], [[0, 1], [0, 1]]], dtype=dtype)
```
pmeier, line 163: 1. The names are somewhat misleading. These should be `{input|expected}_image_tensor`.
2. Is there a reason to keep both methods separate? IIUC, they are always requested together so it might makes sense to merge them into one method to have everything in a single place.
3. You could also accept a `p` parameter here and switch the expected tensor based on it. This way the test only needs to handle the type rather than the values.

test/test_prototype_transforms.py
```diff
@@ -152,3 +153,56 @@ def test_normalize(self, transform, input):
...
+
+    @pytest.mark.parametrize("p", [0.0, 1.0], ids=["p=0", "p=1"])
+    def test_simple_tensor(self, p):
+        input = self.input_tensor()
+
+        actual = transforms.RandomHorizontalFlip(p=p)(input)
```
pmeier, line 169: Nit: although this is correct, I personally feel creating objects and directly calling them is somewhat hard to read. Maybe split this into two lines?

```suggestion
        transform = transforms.RandomHorizontalFlip(p=p)
        actual = transform(input)
```

No strong opinion though. Feel free to leave this open.

pmeier (commented):
Thanks @federicopozzi33 for the continued work on this. I have some more comments on the tests, but otherwise we are good to go!

torchvision/prototype/transforms/_geometry.py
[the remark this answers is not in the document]
pmeier: > Not sure about `segmenation_mask` type. I put `torch.Tensor` to be consistent with the other existing wrappers, but ... shouldn't be of type `features.SegmentationMask`?

Nope, the way you implemented it is correct. The low level kernels need to be fully JIT scriptable. Unfortunately, the custom tensors we use like `features.SegmentationMask` are not. Thus, we opted to only have "simple" tensors there and perform all the wrapping on the transforms. 

This of course means that the low level kernels might be more verbose, because one needs to pass in meta data manually, which otherwise wouldn't be necessary:

https://github.com/pytorch/vision/blob/27745e5be086ecdf2a8e609d65c13969cb5201f7/torchvision/prototype/transforms/functional/_geometry.py#L18-L20

https://github.com/pytorch/vision/blob/27745e5be086ecdf2a8e609d65c13969cb5201f7/torchvision/prototype/transforms/_geometry.py#L21-L23

datumbox (approved):
The code on the transforms LGTM. Once @pmeier's comments on the tests are addressed we are good to merge.

test/test_prototype_transforms.py
federicopozzi33: Just a doubt about 1.
Do you refer to the two methods or to the variable names in tests?

For example:
```
input, expected = self.input_expected_image_tensor(p, dtype=torch.uint8)
transform = transforms.RandomHorizontalFlip(p=p)

actual = transform(to_pil_image(input))

assert_equal(expected, pil_to_tensor(actual))
```

`input` and `expected` are decorated in this case, so maybe it might be more correct using `input_image_tensor` and `expected_image_tensor`. On the other hand, they are very long and verbose variable names, and the method name is quite explicit now.

test/test_prototype_transforms.py
```diff
@@ -188,3 +189,58 @@ def test_random_resized_crop(self, transform, input):
...
+
+@pytest.mark.parametrize("p", [0.0, 1.0])
+class TestRandomHorizontalFlip:
+    def input_expected_image_tensor(self, p, dtype=torch.float32):
+        input = torch.tensor([[[0, 1], [0, 1]], [[1, 0], [1, 0]]], dtype=dtype)
+        expected = torch.tensor([[[1, 0], [1, 0]], [[0, 1], [0, 1]]], dtype=dtype)
+
+        if p == 1.0:
+            return input, expected
+        return input, input
```
pmeier, line 202:
```suggestion
        return input, expected if p == 1 else input
```

pmeier (approved):
One minor suggestion. Otherwise LGTM! Thanks a lot for the patience @federicopozzi33!

test/test_prototype_transforms.py
federicopozzi33: No problem! I saw you already fixed :)

Thank you for the reviews!

Issue #5040, opened 2021-12-07
Legacy interface for prototype datasets?


pmeier:
The prototype datasets change the interface in two ways:

1. The input parameters are a little different in some cases. For example in the current API the MNIST dataset would be instantiated with `datasets.MNIST(..., train=True)` whereas now it looks like `datasets.load("mnist", split="train")`. 
2. The output is completely different. Before we returned a tuple (sometimes of varying length) whereas now we always return a dictionary. Furthermore, before we used `PIL` images and `numpy` arrays as return types, whereas now we always use tensor subclasses. 

To lower the burden to move to the new style datasets a little, we could have a `legacy: bool = False` keyword argument on `datasets.load()`. For all datasets that have a legacy variant, we could simply implement two functions that map the input and output. If a dataset doesn't support a legacy variant, we could simply error out.

pmeier:
> I assume you refer to the `torchvision.prototype.datasets.load()` method. Is this where you propose to add the `legacy` parameter?

Correct. For the legacy example `datasets` comes from `from torchvision import datasets`, whereas for the prototype example it comes from `from torchvision.prototype import datasets`.

> Can you talk more about how supporting both legacy and non-legacy datasets via the new registration API will help/simplify adoption? It's unclear to me what the user would have to do if we don't implement this proposal versus implementing it. 

There is a difference between supporting legacy datasets and supporting the interface of the legacy datasets. The former will not be possible, while the second is the subject of this discussion. 

Let's say the current test loop looks somewhat like this:

```python
from torchvision import datasets

dataset = datasets.MNIST(..., train=False)

for input, target in dataset:
    ....
```

With the prototype datasets the same loop looks like this:

```python
from torchvision.prototype import datasets

dataset = datasets.load("mnist", split="test")

for sample in dataset:
    input, target = sample["image], sample["label"]
```

There are three differences here:

1. Instead of instantiating a dataset class, the user now has to call `datasets.load()`.
2. Instead of using `train=False`, the user now has to pass `split="test"`.
3. Instead of directly getting a tuple of the input and target, the user now has to grab them from the sample dictionary.

Point 1. is a deliberate choice and there is no way around that. If the user wants to use the new datasets, they need to change the instantiation.

Point 2. and 3. however are only changes we made to unify the datasets. Meaning, if the user wants the exact same behavior as before, they need to put in some extra work with no benefit. 

[...]

PR #5565, opened 2022-03-08
implement imagenet prototype dataset as function

pmeier:
First dataset that implements the "function design" we discussed in our latest sync. I didn't find any blockers, but there are still two issues I currently don't have a solution for.

### DIFF
--- a/test/builtin_dataset_mocks.py
+++ b/test/builtin_dataset_mocks.py
@@ -9,7 +9,6 @@ import lzma
 import pathlib
 import pickle
 import random
-import unittest.mock
 import xml.etree.ElementTree as ET
 from collections import defaultdict, Counter
 
@@ -21,7 +20,6 @@ from datasets_utils import make_zip, make_tar, create_image_folder, create_image
 from torch.nn.functional import one_hot
 from torch.testing import make_tensor as _make_tensor
 from torchvision.prototype import datasets
-from torchvision.prototype.utils._internal import sequence_to_str
 
 make_tensor = functools.partial(_make_tensor, device="cpu")
 make_scalar = functools.partial(make_tensor, ())
@@ -66,17 +64,17 @@ class DatasetMock:
 
         mock_info = self._parse_mock_info(self.mock_data_fn(root, config))
 
-        with unittest.mock.patch.object(datasets.utils.Dataset2, "__init__"):
-            required_file_names = {
-                resource.file_name for resource in datasets.load(self.name, root=root, **config)._resources()
-            }
-        available_file_names = {path.name for path in root.glob("*")}
-        missing_file_names = required_file_names - available_file_names
-        if missing_file_names:
-            raise pytest.UsageError(
-                f"Dataset '{self.name}' requires the files {sequence_to_str(sorted(missing_file_names))} "
-                f"for {config}, but they were not created by the mock data function."
-            )
+        # with unittest.mock.patch.object(datasets.utils.Dataset2, "__init__"):
+        #     required_file_names = {
+        #         resource.file_name for resource in datasets.load(self.name, root=root, **config)._resources()
+        #     }
+        # available_file_names = {path.name for path in root.glob("*")}
+        # missing_file_names = required_file_names - available_file_names
+        # if missing_file_names:
+        #     raise pytest.UsageError(
+        #         f"Dataset '{self.name}' requires the files {sequence_to_str(sorted(missing_file_names))} "
+        #         f"for {config}, but they were not created by the mock data function."
+        #     )
 
         return mock_info
 
--- a/test/test_prototype_builtin_datasets.py
+++ b/test/test_prototype_builtin_datasets.py
@@ -10,6 +10,7 @@ from torch.testing._comparison import assert_equal, TensorLikePair, ObjectPair
 from torch.utils.data.graph import traverse
 from torchdata.datapipes.iter import Shuffler, ShardingFilter
 from torchvision.prototype import transforms, datasets
+from torchvision.prototype.datasets.utils._internal import TakerDataPipe
 from torchvision.prototype.utils._internal import sequence_to_str
 
 
@@ -51,8 +52,10 @@ class TestCommon:
 
         dataset = datasets.load(dataset_mock.name, **config)
 
-        if not isinstance(dataset, datasets.utils.Dataset2):
-            raise AssertionError(f"Loading the dataset should return an Dataset, but got {type(dataset)} instead.")
+        if not isinstance(dataset, TakerDataPipe):
+            raise AssertionError(
+                f"Loading the dataset should return an TakerDataPipe, but got {type(dataset)} instead."
+            )
 
     @parametrize_dataset_mocks(DATASET_MOCKS)
     def test_sample(self, test_home, dataset_mock, config):
@@ -100,7 +103,6 @@ class TestCommon:
 
         next(iter(dataset.map(transforms.Identity())))
 
-    @pytest.mark.xfail(reason="See https://github.com/pytorch/data/issues/237")
     @parametrize_dataset_mocks(DATASET_MOCKS)
     def test_serializable(self, test_home, dataset_mock, config):
         dataset_mock.prepare(test_home, config)
@@ -109,7 +111,6 @@ class TestCommon:
 
         pickle.dumps(dataset)
 
-    @pytest.mark.xfail(reason="See https://github.com/pytorch/data/issues/237")
     @parametrize_dataset_mocks(DATASET_MOCKS)
     @pytest.mark.parametrize("annotation_dp_type", (Shuffler, ShardingFilter))
     def test_has_annotations(self, test_home, dataset_mock, config, annotation_dp_type):
--- a/test/test_prototype_builtin_datasets.py
+++ b/test/test_prototype_builtin_datasets.py
@@ -21,7 +21,7 @@ assert_samples_equal = functools.partial(
 
 @pytest.fixture
 def test_home(mocker, tmp_path):
-    mocker.patch("torchvision.prototype.datasets._api.home", return_value=str(tmp_path))
+    mocker.patch("torchvision.prototype.datasets.utils._internal.home", return_value=str(tmp_path))
     yield tmp_path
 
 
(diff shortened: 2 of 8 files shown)


test/builtin_dataset_mocks.py
```diff
@@ -66,17 +64,17 @@ def prepare(self, home, config):
...
+        #     required_file_names = {
+        #         resource.file_name for resource in datasets.load(self.name, root=root, **config)._resources()
+        #     }
+        # available_file_names = {path.name for path in root.glob("*")}
+        # missing_file_names = required_file_names - available_file_names
+        # if missing_file_names:
+        #     raise pytest.UsageError(
+        #         f"Dataset '{self.name}' requires the files {sequence_to_str(sorted(missing_file_names))} "
+        #         f"for {config}, but they were not created by the mock data function."
+        #     )
```
pmeier, line 77: Since the resources are now fully internal inside the dataset function, I don't see a good way to check if the mock data is set up correctly. One thing we could try is to patch 

https://github.com/pytorch/vision/blob/d8654bb0d84fd2ba8b42cd58d881523821a6214c/torchvision/prototype/datasets/utils/_resource.py#L117-L119

on all subclasses to raise the error I commented out above.

test/test_prototype_builtin_datasets.py
```diff
@@ -10,6 +10,7 @@
 from torch.utils.data.graph import traverse
 from torchdata.datapipes.iter import Shuffler, ShardingFilter
 from torchvision.prototype import transforms, datasets
+from torchvision.prototype.datasets.utils._internal import TakerDataPipe
```
pmeier, line 13: @NivekT If I recall correctly, we wanted to upstream this to `torchdata`. Any progress on that?

torchvision/prototype/datasets/_builtin/imagenet.py
```diff
@@ -20,176 +19,176 @@
...
             "test": 100_000,
-        }[self._split]
+        }[split],
+    )
+
 
-    def _generate_categories(self) -> List[Tuple[str, ...]]:
-        self._split = "val"
-        resources = self._resources()
+def generate_categories(root: Union[str, pathlib.Path], **kwargs: Any) -> List[Tuple[str, ...]]:
```
pmeier, line 186: This is currently not working. Before, we checked for this method and called it through the already exposed dataset class. Now, we either need to expose this explicitly or find another way to call it. 

In one of the very first iterations I simply added a `if __name__ == "__main__": ...` to the module and performed everything there. This way we wouldn't have an option to generate multiple category files at once, but that is probably not a significant issue.

torchvision/prototype/datasets/_builtin/__init__.py
```diff
@@ -1,3 +1,4 @@
+from . import imagenet
```
pmeier, line 1: Since we register through the decorator now, we don't need to explicitly import anything anymore. We just need to make sure the decorator is executed.

NicolasHug: We also don't even need to import anything here, right? Since the functions would become public we just have to expose them in `datasets.__init__.py`

torchvision/prototype/datasets/_builtin/imagenet.py
```diff
@@ -20,176 +19,176 @@
...
-        "val": "c7e06a6c0baccf06d8dbeb6577d71efff84673a5dbdd50633ab44f8ea0456ae0",
-        "test_v10102019": "9cf7f8249639510f17d3d8a0deb47cd22a435886ba8e29e2b3223e65a4079eb4",
-    }
-
-    def _resources(self) -> List[OnlineResource]:
-        name = "test_v10102019" if self._split == "test" else self._split
-        images = ImageNetResource(
-            file_name=f"ILSVRC2012_img_{name}.tar",
-            sha256=self._IMAGES_CHECKSUMS[name],
+def imagenet(root: Union[str, pathlib.Path], *, split: str = "train", **kwargs: Any) -> TakerDataPipe:
```
NicolasHug, line 136: I don't know yet whether the function should have the same name as the one available in `load()` (i.e. imagenet), or if we should keep the same names as our current dataset to minimize friction (i.e. ImageNet). We can decide that later.

torchvision/prototype/datasets/_builtin/__init__.py
pmeier: Should the functions be exposed at `torchvision.prototype.datasets`?

torchvision/prototype/datasets/_builtin/imagenet.py
NicolasHug: The way we'll handle the deprecation / transition is related indeed. That's also something we'll need to decide on. But there is value in conserving the same name, so let's keep an open mind on this.

pmeier: > But there is value in conserving the same name, so let's keep an open mind on this.

The new datasets are BC breaking in so many ways and the consensus from #5040 was that having a compatibility layer is not worth it. Thus, I don't think that the name is the problem here given that they need to do some more work on the entire pipeline.

NicolasHug: I'm agree that we don't want a compatibility layer.
Still, it's always good to preserve what we can preserve if we can do it for cheap. Again, let's please keep an open mind here.

pmeier: > Still, it's always good to preserve what we can preserve if we can do it for cheap.

I only partially agree. Although it is cheap it brings 2 problems:

1. The old and the new API cannot coexist. Thus, we would have a hard BC break even if users are fine with sticking to the old API for now.
2. Having the same name with different functionality is IMO more confusing than having two separate names one of which will be deprecated in favor of the other.

> Again, let's please keep an open mind here.

:+1:

test/test_prototype_builtin_datasets.py
NivekT: I will have a look at the implementation sometime today

test/builtin_dataset_mocks.py
ejguan: Could `load_images_dp` be registered and mocked?

pmeier: That would require registering functions like that just for testing purposes. Not sure if we should go that way.

torchvision/prototype/datasets/_builtin/imagenet.py
```diff
@@ -20,176 +27,181 @@
...
-    ) -> Dict[str, Any]:
-        label_data, (path, buffer) = data
-
-        return dict(
-            dict(zip(("label", "wnid"), label_data if label_data else (None, None))),
-            path=path,
-            image=EncodedImage.from_file(buffer),
+
+        meta_dp = Mapper(meta_dp, extract_categories_and_wnids)
+        _, wnids = zip(*next(iter(meta_dp)))
```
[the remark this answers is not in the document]
pmeier: Good catch! Should have been used below like

https://github.com/pytorch/vision/blob/d8654bb0d84fd2ba8b42cd58d881523821a6214c/torchvision/prototype/datasets/_builtin/imagenet.py#L170

torchvision/prototype/datasets/_builtin/imagenet.py
```diff
@@ -20,176 +27,181 @@
...
-        meta_dp = Mapper(meta_dp, self._extract_categories_and_wnids)
+    meta_dp = Filter(devkit_dp, path_comparator("name", "meta.mat"))
+    meta_dp = Mapper(meta_dp, extract_categories_and_wnids)
 
-        categories_and_wnids = cast(List[Tuple[str, ...]], next(iter(meta_dp)))
-        categories_and_wnids.sort(key=lambda category_and_wnid: category_and_wnid[1])
-        return categories_and_wnids
+    categories_and_wnids = cast(List[Tuple[str, ...]], next(iter(meta_dp)))
+    categories_and_wnids.sort(key=lambda category_and_wnid: category_and_wnid[1])
+    return categories_and_wnids
```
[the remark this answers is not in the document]
pmeier: > Noob question: why we want to generate these categories before loading the dataset?

1. We want to have access to the categories without loading the dataset, Particular, they are returned by `datasets.info("imagenet")`.
2. The file we can extract the categories from is only needed for the `"val"` split. If we didn't store the categories somewhere, we would need this file regardless of the selected split.

> And, based on the previous implementation, this function is attached to `Dataset2` class. Does that mean this function was invoked once a dataset instance?

Nope, it was only called from this script

https://github.com/pytorch/vision/blob/d8654bb0d84fd2ba8b42cd58d881523821a6214c/torchvision/prototype/datasets/generate_category_files.py#L21-L25

The method is just a way to document how we extracted the categories in the first place. Users have no touching point with it whatsoever. Even developers will only need to work on it when they are adding a new dataset or found a bug on a existing dataset affecting the categories. See also https://github.com/pytorch/vision/pull/5565#discussion_r821477962.

test/test_prototype_builtin_datasets.py
pmeier: Please tag me in the PR so I can make the changes here after it is landed.

torchvision/prototype/datasets/_builtin/imagenet.py
```diff
@@ -2,13 +2,20 @@
...
+)
+from torchvision.datasets.utils import verify_str_arg
+from torchvision.prototype.datasets import home
 from torchvision.prototype.datasets.utils import (
     DatasetInfo,
-    OnlineResource,
     ManualDownloadResource,
-    Dataset2,
 )
 from torchvision.prototype.datasets.utils._internal import (
```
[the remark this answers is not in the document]
pmeier: See #5570.

torchvision/prototype/datasets/_builtin/imagenet.py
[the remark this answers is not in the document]
ejguan: I see. SGTM

test/builtin_dataset_mocks.py
ejguan: Then, how about making `info` contains another dictionary of `Resource`s? And, `load_images_dp` could load resource from `info`.

pmeier: That would be a possibility, yes. cc @NicolasHug

pmeier:
Superseded by #5778.

Issue #5502, opened 2022-03-01
Update low-level functional transforms with value_range


vfdev-5:
Following https://github.com/pytorch/vision/pull/5500#discussion_r816599553 we may want to update low-level functional transforms with `value_range` argument to avoid implicit hard-coded max range definition:
- 255 for uint8
- 1.0 for float

Today this is done for 
- adjust_hue
https://github.com/pytorch/vision/blob/95d418970e6dbf2e4d928a204c4e620da7bccdc0/torchvision/transforms/functional_tensor.py#L210-L211

- all ops using `_blend`:
https://github.com/pytorch/vision/blob/95d418970e6dbf2e4d928a204c4e620da7bccdc0/torchvision/transforms/functional_tensor.py#L259-L262

We can introduce new argument `value_range` and use it explicitly for these ops.
In general we can think of `value_range` as a tuple (min, max) which would cover majority of imagery where channels ranges are similar. There could be however other type of images (e.g. think of non-RGB color spaces or particular imagery) where value ranges could vary per channel, thus we may need to represent `value_range` as a list of 2-tuples: [(min_1, max_1), (min_2, max_2), ...] 

cc @bjuncek @datumbox

datumbox:
I agree with the idea of passing `value_range` on our existing kernels. When `None` we can estimate the `bound` using our existing approach.

Concerning using single values vs single MinMax vs Multi MinMax, I think all boils down to what is our target. Do we see TorchVision offering native support for Transforms for more exotic non-RGB/Grayscale spaces? Perhaps that could be useful for medical or astronomical applications but we should be careful not to overcomplicate the API with features we won't use. Are transforms that use `_blend` even likely to be used in such domains?

I would like to get the advise of you and @pmeier on this one for how we proceed.

NicolasHug:
Apart from making the implicit ranges more explicit, what are other problems that having `value_range`  would address?

I agree that the way we currently handle ranges isn't nice (read: not documented enough). But from my non-informed POV, it looks like adding `value_range` would add a lot of extra complexity for something that might be solved with better docs / user awareness?

datumbox:
@NicolasHug Thanks for the feedback, I think we have a description/context gap on this issue. Instead of jumping directly to the problem, we should provide info on what we try to solve and why.

[...]

PR #5566, opened 2022-03-08
Recoded _max_value method using a dictionary

vfdev-5:
Description:
- Removed _max_value method and added a dictionary

Related to https://github.com/pytorch/vision/issues/5502

cc @datumbox


torchvision/transforms/functional_tensor.py
```diff
@@ -72,6 +54,14 @@ def convert_image_dtype(image: torch.Tensor, dtype: torch.dtype = torch.float) -
     if image.dtype == dtype:
         return image
 
+    _max_values: Dict[torch.dtype, float] = {
+        torch.uint8: float(255),
+        torch.int8: float(127),
+        torch.int16: float(32767),
+        torch.int32: float(2147483647),
+        torch.int64: float(9223372036854775807),
```
datumbox, line 62: Should we have the unsigned versions for uint16, uint32 etc?

torchvision/transforms/functional_tensor.py
```diff
@@ -91,19 +81,28 @@ def convert_image_dtype(image: torch.Tensor, dtype: torch.dtype = torch.float) -
         # `max + 1 - epsilon` provides more evenly distributed mapping of
         # ranges of floats to ints.
         eps = 1e-3
-        max_val = _max_value(dtype)
+        if dtype not in _max_values:
+            msg = f"Internal error when casting from {image.dtype} to {dtype}."
+            raise RuntimeError(msg)
```
datumbox, line 86: Nit: why the variable assignment?

datumbox (commented):
Thanks @vfdev-5. A few comments + there are related [failing](https://app.circleci.com/pipelines/github/pytorch/vision/15451/workflows/3aa467c0-ad3a-4fc5-9871-46794b889d9b/jobs/1248435) tests.

torchvision/transforms/functional_tensor.py
vfdev-5: I think there is no support for these types: https://pytorch.org/docs/stable/tensors.html#data-types

torchvision/transforms/functional_tensor.py
vfdev-5: no particular reason, just as above.

torchvision/transforms/functional_tensor.py
```diff
@@ -44,22 +44,14 @@ def get_image_num_channels(img: Tensor) -> int:
...
-            break
-    return max_value.item()
+def _max_value(dtype: torch.dtype) -> int:
+    return {
+        torch.uint8: int(2 ** 8) - 1,
+        torch.int8: int(2 ** 7) - 1,
+        torch.int16: int(2 ** 15) - 1,
+        torch.int32: int(2 ** 31) - 1,
+        torch.int64: int(2 ** 63) - 1,
+    }.get(dtype, 1)
```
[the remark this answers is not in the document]
pmeier: > Given that this method is being called often, does it make sense to move the dictionary definition on global scope to avoid reinitialisation?

```py
import torch
from typing import Dict

_MAX_VALUES: Dict[torch.dtype, int] = {
    torch.uint8: int(2 ** 8) - 1,
    torch.int8: int(2 ** 7) - 1,
    torch.int16: int(2 ** 15) - 1,
    torch.int32: int(2 ** 31) - 1,
    torch.int64: int(2 ** 63) - 1,
}


@torch.jit.script
def max_value_global(dtype: torch.dtype) -> int:
    return _MAX_VALUES.get(dtype, 1)
```

fails with

```
RuntimeError: 
attribute lookup is not defined on python value of type 'dict'
```

I also recall Victor saying that globals are not supported.

> If that causes JIT issues, we could also do ifs.

Although the timings are hardly significant here, it seems that the dictionary approach is slightly faster:

```py
import torch
import itertools


@torch.jit.script
def max_value_dict(dtype: torch.dtype) -> int:
    return {
        torch.uint8: int(2 ** 8) - 1,
        torch.int8: int(2 ** 7) - 1,
        torch.int16: int(2 ** 15) - 1,
        torch.int32: int(2 ** 31) - 1,
        torch.int64: int(2 ** 63) - 1,
    }.get(dtype, 1)


@torch.jit.script
def max_value_if(dtype: torch.dtype) -> int:
    if dtype == torch.uint8:
        return int(2 ** 8) - 1
    elif dtype == torch.int8:
        return int(2 ** 7) - 1
    elif dtype == torch.int16:
        return int(2 ** 15) - 1
    elif dtype == torch.int32:
        return int(2 ** 31) - 1
    elif dtype == torch.int64:
        return int(2 ** 63) - 1
    else:
        return 1


num = 1_000_000

dict_timings = []
if_timings = []

for dtype in itertools.islice(
    itertools.cycle(torch.testing._legacy.get_all_dtypes(include_bool=False, include_complex=False)), num
):
    tic = perf_counter_ns()
    max_value_dict(dtype)
    tac = perf_counter_ns()
    dict_timings.append(tac - tic)

    tic = perf_counter_ns()
    max_value_if(dtype)
    tac = perf_counter_ns()
    if_timings.append(tac - tic)

print(f"{torch.tensor(dict_timings, dtype=torch.float64).mean() * 1e-3:.3f} µs per call using dicts")
print(f"{torch.tensor(if_timings, dtype=torch.float64).mean() * 1e-3:.3f} µs per call using ifs")
```

```
0.990 µs per call using dicts
1.100 µs per call using ifs
```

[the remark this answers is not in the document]
pmeier: In eager mode ifs are indeed faster

```
0.565 µs per call using dicts
0.267 µs per call using ifs
```

but here the timings are becoming even more insignificant.

pmeier:
We could probably also add an xfail test like 

```py
import torch
import pytest


@pytest.mark.xfail(
    reason="torch.iinfo() is not supported by torchscript. See https://github.com/pytorch/pytorch/issues/41492."
)
def test_max_value_iinfo():
    @torch.jit.script
    def max_value(image: torch.Tensor) -> int:
        return 1 if image.is_floating_point() else torch.iinfo(image.dtype).max
```

to nudge us to switch as soon as support is added.

torchvision/transforms/functional_tensor.py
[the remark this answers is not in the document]
vfdev-5: :) I think it is overthinking but OK i'll change it to if/else implementation if it is better

datumbox:
@pmeier Good call. If they add support our test will fail and we will certainly know about it :)

@vfdev-5 Feel free to merge on green CI.

torchvision/transforms/functional_tensor.py
```diff
@@ -44,22 +44,19 @@ def get_image_num_channels(img: Tensor) -> int:
...
+    elif dtype == torch.int8:
+        return int(2 ** 7) - 1
+    elif dtype == torch.int16:
+        return int(2 ** 15) - 1
+    elif dtype == torch.int32:
+        return int(2 ** 31) - 1
+    elif dtype == torch.int64:
+        return int(2 ** 63) - 1
+    else:
+        return 1
```
datumbox, line 59: JIT on internal tests REALLY doesn't like this and fails with:
```
runtime error: 9.22337e+18 is outside the range of representable values of type 'long'
```

Replacing this with the following works:
```python
def _max_value(dtype: torch.dtype) -> int:
    if dtype == torch.uint8:
        return 255
    elif dtype == torch.int8:
        return 127
    elif dtype == torch.int16:
        return 32767
    elif dtype == torch.int32:
        return 2147483647
    elif dtype == torch.int64:
        return 9223372036854775807
    else:
        return 1
```

I'll cherrypick it after it's merged internally.

PR #5567, opened 2022-03-08
support grayscale / RGB alpha conversions

### LINKED ISSUE #5510: [RFC] How do we want to deal with images that include alpha channels?
This discussion started in https://github.com/pytorch/vision/pull/5500#discussion_r816503203 and @vfdev-5 and I continued offline.

PIL as well as our image reading functions support RGBA images

https://github.com/pytorch/vision/blob/95d418970e6dbf2e4d928a204c4e620da7bccdc0/torchvision/io/image.py#L16-L31

but our color transformations currently only support RGB images ignoring an extra alpha channel. This leads to wrong results. One thing that we agreed upon is that these transforms should fail if anything but 3 channels is detected.

Still, some datasets include non-RGB images so we need to deal with this for a smooth UX. Previously we implicitly converted every image to RGB before returning it from a dataset

https://github.com/pytorch/vision/blob/f9fbc104c02f277f9485d9f8727f3d99a1cf5f0b/torchvision/datasets/folder.py#L245-L249

Since we no longer decode images in the datasets, we need to provide a solution for the users here. I currently see two possible options:

1. We could deal with this on a per-image basis within the dataset. For example, the train split of ImageNet contains a single RGBA image. We could simply perform an appropriate conversion for irregular image modes in the dataset so this issue is abstracted away from the user. `tensorflow-datasets` uses this approach: https://github.com/tensorflow/datasets/blob/a1caff379ed3164849fdefd147473f72a22d3fa7/tensorflow_datasets/image_classification/imagenet.py#L105-L131
2. The most common non-RGB image in datasets are grayscale images. For example, the train split of ImageNet contains 19970 grayscale images. Thus, the users will need a `transforms.ConvertImageColorSpace("rgb")` in most cases anyway. If that would support RGBA to RGB conversions the problem would also be solved. The conversion happens with this formula:

    ```
    pixel_new = (1 - alpha) * background + alpha * pixel_old
    ```
    
    where `pixel_{old|new}` is a single value from a color channel. Since we don't know `background` we need to either make assumptions or require the user to provide a value for it. I'd wager a guess that in 99% of the cases the background is white. i.e. `background == 1`, but we can't be sure about that.
    
    Another issue with this is that the user has no option to set the background on a per-image basis in the transforms pipeline if that is needed.

    In special case for `alpha == 1` everywhere, the equation above simplifies to

    ```
    pixel_new = pixel_old
    ```

    which is equivalent to stripping the alpha channel. We could check for that and only perform the RGBA to RGB transform if the condition holds or the user supplies a background color.

cc @pmeier @vfdev-5 @datumbox @bjuncek

pmeier:
Closes #5510.

### DIFF
--- a/test/test_prototype_transforms.py
+++ b/test/test_prototype_transforms.py
@@ -102,7 +102,14 @@ class TestSmoke:
             (
                 transform,
                 itertools.chain.from_iterable(
-                    fn(dtypes=[torch.uint8], extra_dims=[(4,)])
+                    fn(
+                        color_spaces=[
+                            features.ColorSpace.GRAYSCALE,
+                            features.ColorSpace.RGB,
+                        ],
+                        dtypes=[torch.uint8],
+                        extra_dims=[(4,)],
+                    )
                     for fn in [
                         make_images,
                         make_vanilla_tensor_images,
@@ -151,4 +158,33 @@ class TestSmoke:
         ]
     )
     def test_random_resized_crop(self, transform, input):
+        transform(input)
+
+    @parametrize(
+        [
+            (
+                transforms.ConvertImageColorSpace(color_space=new_color_space, old_color_space=old_color_space),
+                itertools.chain.from_iterable(
+                    [
+                        fn(color_spaces=[old_color_space])
+                        for fn in (
+                            make_images,
+                            make_vanilla_tensor_images,
+                            make_pil_images,
+                        )
+                    ]
+                ),
+            )
+            for old_color_space, new_color_space in itertools.product(
+                [
+                    features.ColorSpace.GRAYSCALE,
+                    features.ColorSpace.GRAYSCALE_ALPHA,
+                    features.ColorSpace.RGB,
+                    features.ColorSpace.RGBA,
+                ],
+                repeat=2,
+            )
+        ]
+    )
+    def test_convert_image_color_space(self, transform, input):
         transform(input)
--- a/test/test_prototype_transforms.py
+++ b/test/test_prototype_transforms.py
@@ -180,7 +180,7 @@ class TestSmoke:
                     features.ColorSpace.GRAYSCALE,
                     features.ColorSpace.GRAYSCALE_ALPHA,
                     features.ColorSpace.RGB,
-                    features.ColorSpace.RGBA,
+                    features.ColorSpace.RGB_ALPHA,
                 ],
                 repeat=2,
             )
--- a/test/test_prototype_transforms.py
+++ b/test/test_prototype_transforms.py
@@ -104,7 +104,7 @@ class TestSmoke:
                 itertools.chain.from_iterable(
                     fn(
                         color_spaces=[
-                            features.ColorSpace.GRAYSCALE,
+                            features.ColorSpace.GRAY,
                             features.ColorSpace.RGB,
                         ],
                         dtypes=[torch.uint8],
@@ -177,8 +177,8 @@ class TestSmoke:
             )
             for old_color_space, new_color_space in itertools.product(
                 [
-                    features.ColorSpace.GRAYSCALE,
-                    features.ColorSpace.GRAYSCALE_ALPHA,
+                    features.ColorSpace.GRAY,
+                    features.ColorSpace.GRAY_ALPHA,
                     features.ColorSpace.RGB,
                     features.ColorSpace.RGB_ALPHA,
                 ],
--- a/test/test_prototype_transforms_functional.py
+++ b/test/test_prototype_transforms_functional.py
@@ -7,6 +7,7 @@ import torchvision.prototype.transforms.functional as F
 from torch import jit
 from torch.nn.functional import one_hot
 from torchvision.prototype import features
+from torchvision.transforms.functional_tensor import _max_value as get_max_value
 
 make_tensor = functools.partial(torch.testing.make_tensor, device="cpu")
 
@@ -25,7 +26,7 @@ def make_image(size=None, *, color_space, extra_dims=(), dtype=torch.float32, co
         raise pytest.UsageError() from error
 
     shape = (*extra_dims, num_channels, *size)
-    max_value = 1 if dtype.is_floating_point else torch.iinfo(dtype).max
+    max_value = get_max_value(dtype)
     data = make_tensor(shape, low=0, high=max_value, dtype=dtype)
     if color_space in {features.ColorSpace.GRAYSCALE_ALPHA, features.ColorSpace.RGBA} and constant_alpha:
         alpha = data[..., -1:, :, :]
--- a/test/test_prototype_transforms_functional.py
+++ b/test/test_prototype_transforms_functional.py
@@ -29,8 +29,7 @@ def make_image(size=None, *, color_space, extra_dims=(), dtype=torch.float32, co
     max_value = get_max_value(dtype)
     data = make_tensor(shape, low=0, high=max_value, dtype=dtype)
     if color_space in {features.ColorSpace.GRAYSCALE_ALPHA, features.ColorSpace.RGBA} and constant_alpha:
-        alpha = data[..., -1:, :, :]
-        data = torch.cat((data[..., :-1, :, :], torch.full_like(alpha, max_value)), dim=-3)
+        data[..., -1, :, :] = max_value
     return features.Image(data, color_space=color_space)
 
 
--- a/test/test_prototype_transforms_functional.py
+++ b/test/test_prototype_transforms_functional.py
@@ -17,9 +17,9 @@ def make_image(size=None, *, color_space, extra_dims=(), dtype=torch.float32, co
     try:
         num_channels = {
             features.ColorSpace.GRAYSCALE: 1,
-            features.ColorSpace.GRAYSCALE_ALPHA: 1,
+            features.ColorSpace.GRAYSCALE_ALPHA: 2,
             features.ColorSpace.RGB: 3,
-            features.ColorSpace.RGBA: 3,
+            features.ColorSpace.RGBA: 4,
         }[color_space]
     except KeyError as error:
         raise pytest.UsageError() from error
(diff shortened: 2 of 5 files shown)


test/test_prototype_transforms_functional.py
```diff
@@ -48,15 +65,12 @@ def randint_with_tensor_bounds(arg1, arg2=None, **kwargs):
...
-    try:
-        return torch.stack(
-            [
-                torch.randint(low_scalar, high_scalar, (), **kwargs)
-                for low_scalar, high_scalar in zip(low.flatten().tolist(), high.flatten().tolist())
-            ]
-        ).reshape(low.shape)
-    except RuntimeError as error:
-        raise error
+    return torch.stack(
```
pmeier, line 68: This seems like I left some debug statements.

torchvision/prototype/transforms/_meta.py
```diff
@@ -48,11 +48,11 @@ def __init__(
         super().__init__()
 
         if isinstance(color_space, str):
-            color_space = features.ColorSpace[color_space]
+            color_space = features.ColorSpace.from_str(color_space)
```
pmeier, line 51: Forgotten in the PR that refactored `StrEnum`.

test/test_prototype_transforms.py
```diff
@@ -152,3 +159,32 @@ def test_normalize(self, transform, input):
...
+                [
+                    features.ColorSpace.GRAYSCALE,
+                    features.ColorSpace.GRAYSCALE_ALPHA,
+                    features.ColorSpace.RGB,
+                    features.ColorSpace.RGBA,
+                ],
+                repeat=2,
+            )
+        ]
+    )
```
vfdev-5, line 188: Nit: It is not only related to this one but for previous as well. IMO these parametrizations are really hard to read and understand...  If we could simplify that somehow, it would be nice. Major difficulty is get what is passed to `input` for example...

test/test_prototype_transforms_functional.py
```diff
@@ -11,19 +11,31 @@
...
-        features.ColorSpace.GRAYSCALE: 1,
-        features.ColorSpace.RGB: 3,
-    }[color_space]
+    try:
+        num_channels = {
+            features.ColorSpace.GRAYSCALE: 1,
+            features.ColorSpace.GRAYSCALE_ALPHA: 1,
+            features.ColorSpace.RGB: 3,
+            features.ColorSpace.RGBA: 3,
+        }[color_space]
```
datumbox, line 23: Reading this first time, I found it hard to understand why the alpha values are the way they are. We should refactor on a follow up PR to make this more obvious. One approach could be to generate the right number of dimentions for each channel (gray-alpha=2, rbga=4) but then overwrite the final channel with white. That approach will be easier to understand than initializing an RGBA with 3 channels and concatenating an alpha.

test/test_prototype_transforms_functional.py
```diff
@@ -11,19 +11,31 @@
...
+    except KeyError as error:
+        raise pytest.UsageError() from error
 
     shape = (*extra_dims, num_channels, *size)
-    if dtype.is_floating_point:
-        data = torch.rand(shape, dtype=dtype)
-    else:
-        data = torch.randint(0, torch.iinfo(dtype).max, shape, dtype=dtype)
+    max_value = 1 if dtype.is_floating_point else torch.iinfo(dtype).max
```
datumbox, line 28: I assume this is where Victor's work at #5566 will come in once merged?

torchvision/prototype/features/_image.py
```diff
@@ -73,13 +84,12 @@ def guess_color_space(data: torch.Tensor) -> ColorSpace:
...
-        elif num_channels == 3:
-            return ColorSpace.RGB
-        else:
-            return ColorSpace.OTHER
+        return {
+            1: ColorSpace.GRAYSCALE,
+            2: ColorSpace.GRAYSCALE_ALPHA,
+            3: ColorSpace.RGB,
+            4: ColorSpace.RGBA,
+        }.get(data.shape[-3], ColorSpace.OTHER)
```
datumbox, line 92: I would just switch to ifs. There are 4 values.

torchvision/prototype/transforms/functional/_meta.py
```diff
@@ -57,41 +57,107 @@ def convert_bounding_box_format(
...
+    elif old_color_space == ColorSpace.RGBA:
+        image = _rgba_to_rgb_tensor(image)
 
     if new_color_space == ColorSpace.GRAYSCALE:
-        image = _FT.rgb_to_grayscale(image)
+        image = _rgb_to_grayscale_tensor(image)
+    if new_color_space == ColorSpace.GRAYSCALE_ALPHA:
+        image = _grayscale_to_grayscale_alpha_tensor(_rgb_to_grayscale_tensor(image))
+    elif new_color_space == ColorSpace.RGBA:
+        image = _rgb_to_rgba_tensor(image)
```
datumbox, line 140: I'm OK with this logic for first iteration but we should later come and optimize this to avoid unnecessary conventions. In fact, it might be good to keep a single issue with speed optimizations we consider doing on the future to keep track of them,

datumbox (commented):
@pmeier Thanks! Overall looks good. Only couple of comments for your attention.

test/test_prototype_transforms_functional.py
pmeier: Here and as well as in the kernel implementation.

test/test_prototype_transforms.py
pmeier: It is indeed hard to read. Refactoring the tests is at the top of my todo list.

torchvision/prototype/features/_image.py
```diff
@@ -16,7 +16,22 @@
...
+
+    @classmethod
+    def from_pil_mode(cls, mode: str) -> ColorSpace:
+        if mode == "L":
+            return cls.GRAYSCALE
+        elif mode == "LA":
+            return cls.GRAYSCALE_ALPHA
+        elif mode == "LA":
+            return cls.RGB
+        elif mode == "LA":
```
datumbox, line 31: Wrong strings in all `elif`

torchvision/prototype/features/_image.py
```diff
@@ -76,8 +91,12 @@ def guess_color_space(data: torch.Tensor) -> ColorSpace:
         num_channels = data.shape[-3]
         if num_channels == 1:
             return ColorSpace.GRAYSCALE
-        elif num_channels == 3:
+        elif num_channels == 2:
+            return ColorSpace.GRAYSCALE_ALPHA
+        elif num_channels == 2:
             return ColorSpace.RGB
+        elif num_channels == 2:
```
datumbox, line 98: Fix number of channels

torchvision/prototype/transforms/functional/_meta.py
```diff
@@ -57,41 +58,86 @@ def convert_bounding_box_format(
...
+        )
+    return image
+
+
+def _add_alpha(image: torch.Tensor, alpha: Optional[torch.Tensor] = None) -> torch.Tensor:
+    if alpha is None:
+        shape = list(image.shape)
+        shape[-3] = 1
+        alpha = torch.full(shape, _FT._max_value(image.dtype), dtype=image.dtype, device=image.device)
+    return torch.cat((image, alpha), dim=-3)
```
datumbox, line 79: From what I can tell this should work for batches too. Have you confirmed?

torchvision/prototype/features/_image.py
```diff
@@ -16,7 +16,22 @@
 class ColorSpace(StrEnum):
     OTHER = StrEnum.auto()
     GRAYSCALE = StrEnum.auto()
+    GRAYSCALE_ALPHA = StrEnum.auto()
     RGB = StrEnum.auto()
+    RGBA = StrEnum.auto()
```
datumbox, line 21: This naming is bothering me a bit. I was thinking that maybe `RGB_ALPHA` makes sense because it aligns with `GRAYSCALE_ALPHA`. On the other hand `RGBA` is a thing in PIL. So I'm on the fence, thoughts?

torchvision/prototype/transforms/functional/_meta.py
```diff
@@ -57,41 +58,86 @@ def convert_bounding_box_format(
...
+        return _grayscale_to_rgb_tensor(image)
+    elif old_color_space == ColorSpace.GRAYSCALE and new_color_space == ColorSpace.RGBA:
+        return _add_alpha(_grayscale_to_rgb_tensor(image))
+    elif old_color_space == ColorSpace.GRAYSCALE_ALPHA and new_color_space == ColorSpace.GRAYSCALE:
+        return _strip_alpha(image)
+    elif old_color_space == ColorSpace.GRAYSCALE_ALPHA and new_color_space == ColorSpace.RGB:
+        return _grayscale_to_rgb_tensor(_strip_alpha(image))
+    elif old_color_space == ColorSpace.GRAYSCALE_ALPHA and new_color_space == ColorSpace.RGBA:
+        image, alpha = _split_alpha(image)
+        return _add_alpha(_grayscale_to_rgb_tensor(image), alpha)
```
datumbox, line 112: Nice, you avoid losing info.

torchvision/prototype/transforms/functional/_meta.py
```diff
@@ -57,41 +58,86 @@ def convert_bounding_box_format(
...
+    elif old_color_space == ColorSpace.RGB and new_color_space == ColorSpace.GRAYSCALE_ALPHA:
+        return _add_alpha(_rgb_to_grayscale_tensor(image))
+    elif old_color_space == ColorSpace.RGB and new_color_space == ColorSpace.RGBA:
+        return _add_alpha(image)
+    elif old_color_space == ColorSpace.RGBA and new_color_space == ColorSpace.GRAYSCALE:
+        return _rgb_to_grayscale_tensor(_strip_alpha(image))
+    elif old_color_space == ColorSpace.RGBA and new_color_space == ColorSpace.GRAYSCALE_ALPHA:
+        image, alpha = _split_alpha(image)
+        return _add_alpha(_rgb_to_grayscale_tensor(image), alpha)
+    else:  # old_color_space == ColorSpace.RGBA and new_color_space == ColorSpace.RGB:
```
datumbox, line 124: Nit: Perhaps leave it as an `elif` and throw an error if anything else. This way it will continue behaving if we add something new on the future.

torchvision/prototype/transforms/functional/_meta.py
pmeier:
```py
>>> image = torch.rand(5, 4, 3, 16, 16)
>>> alpha = torch.rand(5, 4, 1, 16, 16)
>>> torch.cat((image, alpha), dim=-3)
torch.Size([5, 4, 4, 16, 16])
```

datumbox (approved):
LGTM, thanks!

PR #5568, opened 2022-03-08
[EXPERIMENTAL] Extending padding to support non-constant fill

datumbox:
PyTorch's [pad](https://pytorch.org/docs/stable/generated/torch.nn.functional.pad.html) supports only constant values. Unfortunately this can be a problem for data augmentation techniques that require padding with a specific fill colour. For some of them we have previously employed the following trick:
https://github.com/pytorch/vision/blob/79892d376c5faeaa55f6ccb656b1a09c0781815a/references/detection/transforms.py#L204-L209

This PR adapts the approach and moves it to `F.pad()`. The `fill` can be either a `float` or a `List[float]`. Unfortunately JIT doesn't allow us to include also `int` and `List[int]`. The PR modifies the default values of some of the methods.

### DIFF
--- a/test/test_functional_tensor.py
+++ b/test/test_functional_tensor.py
@@ -946,9 +946,9 @@ def test_adjust_gamma(device, dtype, config, channels):
 @pytest.mark.parametrize(
     "config",
     [
-        {"padding_mode": "constant", "fill": 0},
-        {"padding_mode": "constant", "fill": 10},
-        {"padding_mode": "constant", "fill": 20},
+        {"padding_mode": "constant", "fill": 10.0},
+        {"padding_mode": "constant", "fill": [10.0, 10.0, 10.0]},
+        {"padding_mode": "constant", "fill": [10.0, 0.0, 10.0]},
         {"padding_mode": "edge"},
         {"padding_mode": "reflect"},
         {"padding_mode": "symmetric"},
--- a/torchvision/transforms/functional.py
+++ b/torchvision/transforms/functional.py
@@ -431,7 +431,9 @@ def resize(
     return F_t.resize(img, size=size, interpolation=interpolation.value, max_size=max_size, antialias=antialias)
 
 
-def pad(img: Tensor, padding: List[int], fill: Union[List[float], float] = 0.0, padding_mode: str = "constant") -> Tensor:
+def pad(
+    img: Tensor, padding: List[int], fill: Union[List[float], float] = 0.0, padding_mode: str = "constant"
+) -> Tensor:
     r"""Pad the given image on all sides with the given "pad" value.
     If the image is torch Tensor, it is expected
     to have [..., H, W] shape, where ... means at most 2 leading dimensions for mode reflect and symmetric,
--- a/torchvision/transforms/functional.py
+++ b/torchvision/transforms/functional.py
@@ -2,7 +2,7 @@ import math
 import numbers
 import warnings
 from enum import Enum
-from typing import List, Tuple, Any, Optional
+from typing import List, Tuple, Any, Optional, Union
 
 import numpy as np
 import torch
@@ -431,7 +431,7 @@ def resize(
     return F_t.resize(img, size=size, interpolation=interpolation.value, max_size=max_size, antialias=antialias)
 
 
-def pad(img: Tensor, padding: List[int], fill: int = 0, padding_mode: str = "constant") -> Tensor:
+def pad(img: Tensor, padding: List[int], fill: Union[List[float], float] = 0.0, padding_mode: str = "constant") -> Tensor:
     r"""Pad the given image on all sides with the given "pad" value.
     If the image is torch Tensor, it is expected
     to have [..., H, W] shape, where ... means at most 2 leading dimensions for mode reflect and symmetric,
@@ -451,7 +451,7 @@ def pad(img: Tensor, padding: List[int], fill: int = 0, padding_mode: str = "con
         fill (number or str or tuple): Pixel fill value for constant fill. Default is 0.
             If a tuple of length 3, it is used to fill R, G, B channels respectively.
             This value is only used when the padding_mode is constant.
-            Only number is supported for torch Tensor.
+            Only number or tuple is supported for torch Tensor.
             Only int or str or tuple value is supported for PIL Image.
         padding_mode (str): Type of padding. Should be: constant, edge, reflect or symmetric.
             Default is constant.
--- a/torchvision/transforms/functional.py
+++ b/torchvision/transforms/functional.py
@@ -536,7 +536,7 @@ def center_crop(img: Tensor, output_size: List[int]) -> Tensor:
             (crop_width - image_width + 1) // 2 if crop_width > image_width else 0,
             (crop_height - image_height + 1) // 2 if crop_height > image_height else 0,
         ]
-        img = pad(img, padding_ltrb, fill=0)  # PIL uses fill value 0
+        img = pad(img, padding_ltrb, fill=0.0)  # PIL uses fill value 0
         _, image_height, image_width = get_dimensions(img)
         if crop_width == image_width and crop_height == image_height:
             return img
--- a/torchvision/transforms/functional_pil.py
+++ b/torchvision/transforms/functional_pil.py
@@ -302,10 +302,10 @@ def _parse_fill(
         fill = tuple(fill)
 
     if img.mode != "F":
-        if isinstance(fill, float):
-            fill = int(fill)
-        else:
+        if isinstance(fill, (list, tuple)):
             fill = tuple(int(x) for x in fill)
+        else:
+            fill = int(fill)
 
     return {name: fill}
 
--- a/torchvision/transforms/functional_pil.py
+++ b/torchvision/transforms/functional_pil.py
@@ -154,7 +154,7 @@ def pad(
 
     if not isinstance(padding, (numbers.Number, tuple, list)):
         raise TypeError("Got inappropriate padding arg")
-    if not isinstance(fill, (numbers.Number, str, tuple)):
+    if not isinstance(fill, (numbers.Number, str, list, tuple)):
         raise TypeError("Got inappropriate fill arg")
     if not isinstance(padding_mode, str):
         raise TypeError("Got inappropriate padding_mode arg")
@@ -291,7 +291,7 @@ def _parse_fill(
     # Process fill color for affine transforms
     num_bands = len(img.getbands())
     if fill is None:
-        fill = 0
+        fill = 0.0
     if isinstance(fill, (int, float)) and num_bands > 1:
         fill = tuple([fill] * num_bands)
     if isinstance(fill, (list, tuple)):
@@ -300,6 +300,12 @@ def _parse_fill(
             raise ValueError(msg.format(len(fill), num_bands))
 
         fill = tuple(fill)
+
+    if img.mode != "F":
+        if isinstance(fill, float):
+            fill = int(fill)
+        else:
+            fill = tuple(int(x) for x in fill)
 
     return {name: fill}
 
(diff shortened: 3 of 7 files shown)


test/test_functional_tensor.py
```diff
@@ -946,9 +946,9 @@ def test_adjust_gamma(device, dtype, config, channels):
 @pytest.mark.parametrize(
     "config",
     [
-        {"padding_mode": "constant", "fill": 0},
-        {"padding_mode": "constant", "fill": 10},
-        {"padding_mode": "constant", "fill": 20},
+        {"padding_mode": "constant", "fill": 10.0},
+        {"padding_mode": "constant", "fill": [10.0, 10.0, 10.0]},
+        {"padding_mode": "constant", "fill": [10.0, 0.0, 10.0]},
```
datumbox, line 951: The test won't pass if we provide integers. That's because the test conducts JIT-script checks as well.

Here we check for single values, lists with the same value and lists with different values.

torchvision/transforms/functional.py
```diff
@@ -431,7 +431,9 @@ def resize(
     return F_t.resize(img, size=size, interpolation=interpolation.value, max_size=max_size, antialias=antialias)
 
 
-def pad(img: Tensor, padding: List[int], fill: int = 0, padding_mode: str = "constant") -> Tensor:
+def pad(
+    img: Tensor, padding: List[int], fill: Union[List[float], float] = 0.0, padding_mode: str = "constant"
```
datumbox, line 435: The default value changed from int to float. JIT will fail if we pass integers.

torchvision/transforms/functional.py
```diff
@@ -536,7 +538,7 @@ def center_crop(img: Tensor, output_size: List[int]) -> Tensor:
             (crop_width - image_width + 1) // 2 if crop_width > image_width else 0,
             (crop_height - image_height + 1) // 2 if crop_height > image_height else 0,
         ]
-        img = pad(img, padding_ltrb, fill=0)  # PIL uses fill value 0
+        img = pad(img, padding_ltrb, fill=0.0)  # PIL uses fill value 0
```
datumbox, line 541: Again we need floats to appease JIT.

torchvision/transforms/functional_pil.py
```diff
@@ -154,7 +154,7 @@ def pad(
 
     if not isinstance(padding, (numbers.Number, tuple, list)):
         raise TypeError("Got inappropriate padding arg")
-    if not isinstance(fill, (numbers.Number, str, tuple)):
+    if not isinstance(fill, (numbers.Number, str, list, tuple)):
```
datumbox, line 157: Unrelated bug fix on the original code.

torchvision/transforms/functional_pil.py
```diff
@@ -301,6 +301,12 @@ def _parse_fill(
 
         fill = tuple(fill)
 
+    if img.mode != "F":
+        if isinstance(fill, (list, tuple)):
+            fill = tuple(int(x) for x in fill)
+        else:
+            fill = int(fill)
```
datumbox, line 308: Unrelated bug fix on the original code. This method doesn't work if floats are provided for PIL images, despite the method having floats in the signature.

torchvision/transforms/functional_pil.py
```diff
@@ -291,7 +291,7 @@ def _parse_fill(
     # Process fill color for affine transforms
     num_bands = len(img.getbands())
     if fill is None:
-        fill = 0
+        fill = 0.0
```
datumbox, line 294: Floats to please JIT

torchvision/transforms/functional_tensor.py
```diff
@@ -141,7 +141,7 @@ def crop(img: Tensor, top: int, left: int, height: int, width: int) -> Tensor:
 
     if left < 0 or top < 0 or right > w or bottom > h:
         padding_ltrb = [max(-left, 0), max(-top, 0), max(right - w, 0), max(bottom - h, 0)]
-        return pad(img[..., max(top, 0) : bottom, max(left, 0) : right], padding_ltrb, fill=0)
+        return pad(img[..., max(top, 0) : bottom, max(left, 0) : right], padding_ltrb, fill=0.0)
```
datumbox, line 144: Floats to please JIT

torchvision/transforms/functional_tensor.py
```diff
@@ -411,7 +413,28 @@ def pad(img: Tensor, padding: List[int], fill: int = 0, padding_mode: str = "con
...
 
-    img = torch_pad(img, p, mode=padding_mode, value=float(fill))
+    patch_padding = False
+    if isinstance(fill, (int, float)):
+        v = float(fill)
+    elif any([x != fill[0] for x in fill]):
+        patch_padding = True
+        v = 0.0
+    else:
+        v = fill[0]
```
datumbox, line 423: We only apply the patch if necessary. That is when `fill` is a list without constant values.

torchvision/transforms/functional_tensor.py
```diff
@@ -411,7 +413,28 @@ def pad(img: Tensor, padding: List[int], fill: int = 0, padding_mode: str = "con
...
+    img = torch_pad(img, p, mode=padding_mode, value=v)
+
+    if padding_mode == "constant" and patch_padding and fill is not None:
+        if not isinstance(fill, (tuple, list)):
+            fill = [fill]
+        fill_img = torch.tensor(fill, dtype=img.dtype, device=img.device).view(1, -1, 1, 1)
+        if pad_top > 0:
```
datumbox, line 430: Handling negative padding values.

torchvision/transforms/functional_tensor.py
```diff
@@ -522,17 +560,8 @@ def _assert_grid_transform_inputs(
...
-
-    # Check fill
     num_channels = get_dimensions(img)[0]
-    if isinstance(fill, (tuple, list)) and (len(fill) > 1 and len(fill) != num_channels):
-        msg = (
-            "The number of elements in 'fill' cannot broadcast to match the number of "
-            "channels of the image ({} != {})"
-        )
-        raise ValueError(msg.format(len(fill), num_channels))
+    _assert_fill(fill, num_channels)
```
datumbox, line 564: Just move out the code to reuse it above.

torchvision/transforms/transforms.py
```diff
@@ -428,7 +428,7 @@ def __init__(self, padding, fill=0, padding_mode="constant"):
         if not isinstance(padding, (numbers.Number, tuple, list)):
             raise TypeError("Got inappropriate padding arg")
 
-        if not isinstance(fill, (numbers.Number, str, tuple)):
+        if not isinstance(fill, (numbers.Number, str, tuple, list)):
```
datumbox, line 431: Unrelated bug fix on the original code.

torchvision/transforms/transforms.py
```diff
@@ -641,7 +641,7 @@ def get_params(img: Tensor, output_size: Tuple[int, int]) -> Tuple[int, int, int
         j = torch.randint(0, w - tw + 1, size=(1,)).item()
         return i, j, th, tw
 
-    def __init__(self, size, padding=None, pad_if_needed=False, fill=0, padding_mode="constant"):
+    def __init__(self, size, padding=None, pad_if_needed=False, fill=0.0, padding_mode="constant"):
```
datumbox, line 644: Floats to please JIT

datumbox (commented):
Providing comments to assist review

datumbox:
The previous solution [failed](https://app.circleci.com/pipelines/github/pytorch/vision/15461/workflows/b49d0bc1-07a0-4238-b0d1-2d4eeb6afbdc/jobs/1249301). Seems we might have a gap on our JIT scripts cause it was caught by the doc scripts.

torchvision/transforms/functional_pil.py
vfdev-5: If there is a bug fix here and we do not expect this to land, maybe better to split this into a separate PR ?

torchvision/transforms/functional_pil.py
vfdev-5: Same here ?

torchvision/transforms/functional_tensor.py
```diff
@@ -411,7 +413,22 @@ def pad(img: Tensor, padding: List[int], fill: int = 0, padding_mode: str = "con
         need_cast = True
         img = img.to(torch.float32)
 
-    img = torch_pad(img, p, mode=padding_mode, value=float(fill))
+    img = torch_pad(img, p, mode=padding_mode)
+
+    if padding_mode == "constant":
+        # The following if/else can't be simplified due to JIT
+        if isinstance(fill, (tuple, list)):
+            fill_img = torch.tensor(fill).to(dtype=img.dtype, device=img.device).view(1, -1, 1, 1)
```
vfdev-5, line 421: nit: can't we create it directly as 
```python
fill_img = torch.tensor(fill, dtype=img.dtype, device=img.device)
```
?

torchvision/transforms/functional_tensor.py
```diff
@@ -411,7 +413,22 @@ def pad(img: Tensor, padding: List[int], fill: int = 0, padding_mode: str = "con
         need_cast = True
         img = img.to(torch.float32)
 
-    img = torch_pad(img, p, mode=padding_mode, value=float(fill))
+    img = torch_pad(img, p, mode=padding_mode)
```
vfdev-5, line 416: if `fill` is a scalar now we still transform it to a tensor and apply to the image at most 4 times below (`img[..., :, :pad_left] = fill_img`). Maybe, for performance reasons we could do if/else here and keep previous behaviour with a single torch_pad call for scalars and for list/tuple do what you coded ?

torchvision/transforms/functional.py
datumbox: The `int` used previously is very misleading. We mainly use floats because our tensors get rescaled as you know. Unfortunately adding both `List[int]` and `List[float]` in the union doesn't work due to JIT issues. See https://github.com/pytorch/pytorch/issues/69434

torchvision/transforms/functional_pil.py
datumbox: Yeah we can cherrypick afterwards if we don't land this.

torchvision/transforms/functional_tensor.py
datumbox: No. :( JIT requires it to be behind an if statement. I believe this is because it invokes a different C++ method (The one that receives a list VS a scalar).

torchvision/transforms/functional_tensor.py
datumbox: I had that, see earlier versions of the commit. Unfortunately I couldn't find a way to write it in a JIT-friendly way. See [here](https://github.com/pytorch/vision/pull/5568/commits/ccef4e4c58d4a344f4fa86178472806dccf72d9c#diff-497b983ae7c82237c4a0722bc6f6637525e0cdf8369dbd78ee19cf986bffe258R414) for more details. If you have ideas on how to have the optimization and be JIT-scriptable I'm happy to use them :)

torchvision/transforms/functional_tensor.py
datumbox: Sorry I missed that. That's also needed for the scalar case. I believe some of the tests were failing due to `fill` being float and dtype being integer. Casting solves this.

BTW you are welcome to push to the branch if you want to experiment.

datumbox:
This solution is not good enough. Though the JIT tests pass, there are issues:
- This PR leaves the code in a worse state. It contains a few "voodoo" parts which are non-obvious and exist only to please JIT.
- The end result is very brittle. Making small changes can instantly break the code.
- It's not efficient. JIT doesn't let us apply the patch only when necessary.

For the above reasons, I will close the PR. Perhaps we can revisit this on the future when issue https://github.com/pytorch/pytorch/issues/69434 is addressed. For now I think that applying the right fill for padding can be addressed on the side of new Class Transforms where things don't have to be JIT-scriptable.

torchvision/transforms/functional_tensor.py
datumbox: @ansley FYI this is the kind of weird code one must write to make things JIT-scriptable. Without the explicit if statement, JIT doesn't know how to handle `fill` when scalar vs when list. I believe this has to do with the fact that the C++ implementation ends up calling a different method.

Issue #5046, opened 2021-12-07
Type annotations of normalize() are incorrect


NicolasHug:
I'm seeing the following typing error:

```
torchvision/prototype/transforms/_presets.py:114: error: Argument "mean" to
"normalize" has incompatible type "float"; expected "List[float]"  [arg-type]
            img1 = F.normalize(img1, mean=0.5, std=0.5)
```

This is incorrect, passing single numbers is explicitly supported by `noramlize()`

https://github.com/pytorch/vision/blob/47bd962069ba03f753e7ba711cb825317be0b00a/torchvision/transforms/functional.py#L340-L345

datumbox:
Probably a limitation from the days that JIT didn't support Union? Worth checking if a `Union[float, List[float]]` fixes the situation.

PR #5569, opened 2022-03-09
Allow F.normalize function to use float and list of float as mean and…

YosuaMichael:
Resolve issue https://github.com/pytorch/vision/issues/5046


torchvision/transforms/functional.py
```diff
@@ -334,7 +334,7 @@ def to_pil_image(pic, mode=None):
     return Image.fromarray(npimg, mode=mode)
 
 
-def normalize(tensor: Tensor, mean: List[float], std: List[float], inplace: bool = False) -> Tensor:
+def normalize(tensor: Tensor, mean: Union(float, List[float]), std: Union(float, List[float]), inplace: bool = False) -> Tensor:
```
NicolasHug, line 337: If you look at one of the CI jobs below 
image

you will see that they're failing with the following error:

```
torchvision/transforms/transforms.py:17: in <module>
    from . import functional as F
torchvision/transforms/functional.py:337: in <module>
    def normalize(tensor: Tensor, mean: Union(float, List[float]), std: Union(float, List[float]), inplace: bool = False) -> Tensor:
E   NameError: name 'Union' is not defined
```

So it looks like this is missing an import

test/test_transforms.py
```diff
@@ -1994,6 +1994,17 @@ def test_normalize_3d_tensor():
...
 
+def test_normalize_with_single_value_mean_std():
+    torch.manual_seed(42)
+    n_channels = 3
+    img_size = 10
+    mean = 0.5
+    std = 0.5
+    img = torch.rand(n_channels, img_size, img_size)
+    # Check that this work
+    target = F.normalize(img, mean, std)
```
NicolasHug, line 2005: Nit: Since we don't re-use these variable names, we don't really need to declare them. We can just shorten it to

```py
    img = torch.rand(3, 10, 10)
    target = F.normalize(img, mean=0.5, std=0.5)
```

But in reality, we don't actually need this test at all, as it wouldn't give us any error anyway, because the tests aren't checked by mypy. What we should try to do is to use `mean=some_float` or mean=`std` somewhere in our code instead. If there are no occurrence of such a thing, maybe we can just verify manually.

Note that we have a CI job for type annotations as well, called `type_check_python` https://app.circleci.com/pipelines/github/pytorch/vision/15467/workflows/20203b6e-c9ca-47a3-9d77-22148ff68086/jobs/1249720

NicolasHug (commented):
Thanks for the PR @YosuaMichael ! Following up on your questions on workchat, you'll need to sign the CLA as individual. I made a few comments below, LMK what you think

YosuaMichael:
Hi @datumbox @NicolasHug thanks for the comment! Yeah, I have just read the CI error and indeed I forgot to import Union... 

@NicolasHug you are right about the test! I think we dont need the unit test.
@datumbox will check on https://github.com/pytorch/pytorch/issues/69434 and see if it can be somehow resolved, thanks!

Usually what is the best way to detect this import issue offline? 
I have tried doing `pytest test -vvv -k test_trasforms.py` and `flake8 torchvision/transforms/functional.py` however both did not show the Union import related error. 
After knowing this error, I tried to use pylint and it can detect this error, however it is too noisy as it shows quite a lot of other warnings and errors (is there any lint config used as standard?)

NicolasHug:
@datumbox @YosuaMichael I'm not sure we need to worry about https://github.com/pytorch/pytorch/issues/69434, at least for now. If mypy is happy with `Union[float, List[float]]`, then I believe we're good. If mypy is not happy, then it's fine to close the issue as a no-fix.

> pytest test -vvv -k test_trasforms.py

There's a typo, it should be test_transforms.py. If you ran exactly `pytest test -vvv -k test_trasforms.py`, the no tests actually got run.

Also, flake8 won't check types, only mypy does. So to make sure that mypy is happy with the changes, the best way is to locally write a basic file that calls `F.normalize(img1, mean=0.5, std=0.5)`, and to check that mypy is happy with it

datumbox:
> After knowing this error, I tried to use pylint and it can detect this error, however it is too noisy as it shows quite a lot of other warnings and errors (is there any lint config used as standard?)

Hmmmm. Maybe you got old dependencies or something? Numpy has previously throw a bunch of warnings for me and it was because I had a very old version.

Mypy should also catch this problem. You could run it from project root with `mypy --config-file mypy.ini` (see [guide](https://github.com/pytorch/vision/blob/main/CONTRIBUTING.md#type-annotations)).

torchvision/transforms/functional.py
```diff
@@ -334,7 +335,9 @@ def to_pil_image(pic, mode=None):
     return Image.fromarray(npimg, mode=mode)
 
 
-def normalize(tensor: Tensor, mean: List[float], std: List[float], inplace: bool = False) -> Tensor:
+def normalize(
+    tensor: Tensor, mean: Union(float, List[float]), std: Union(float, List[float]), inplace: bool = False
```
datumbox, line 339:
```suggestion
    tensor: Tensor, mean: Union[float, List[float]], std: Union[float, List[float]], inplace: bool = False
```

datumbox (commented):
You also got a typo on the annotation (brackets vs parenthesis).

datumbox:
@YosuaMichael You need to update the method on `functional_tensor.py` as well. This is why the test is [failing](https://app.circleci.com/pipelines/github/pytorch/vision/15489/workflows/46627cb1-f931-4e53-9553-5fb93a549e3b/jobs/1251646).

YosuaMichael:
Hi @datumbox , thanks for your quick response!
Previously I am not sure how to debug on my local laptop and now thanks to @NicolasHug I am able to debug it locally.

After changing functional_tensor, it seems to still give error when it call `torch.as_tensor`.
As of now my fix is to use `isinstance` and make sure the mean and std is typed as `List[float]` before calling `torch.as_tensor`. It is working in my laptop, but I will wait the CI check first if there is anything that failed.

YosuaMichael:
Noted the failed test on Tuple type, and will work to fix it

YosuaMichael:
Take note that using `Union[float, List[float]]` as this PR suggest may have unexpected behaviour when called on `torch.jit.script`, for instance it will not accept tuple type as input. Here are example script to illustrate this:
```
import torch

class TestClass_List(torch.nn.Module):
    def forward(self, img):
        return F.normalize(img, [0.5,0.5,0.5], [0.5,0.5,0.5])

class TestClass_Float(torch.nn.Module):
    def forward(self, img):
        return F.normalize(img, 0.5, 0.5)

class TestClass_Tuple(torch.nn.Module):
    def forward(self, img):
        return F.normalize(img, (0.5,0.5,0.5), (0.5,0.5,0.5))

class TestClass_Int(torch.nn.Module):
    def forward(self, img):
        return F.normalize(img, 1, 1)
  
img = torch.rand(3, 2, 2)
tc_list = TestClass_List()
tc_float = TestClass_Float()
tc_tuple = TestClass_Tuple()
tc_int = TestClass_Int()

# NOTE: Call to F.normalize without jit will work on float, int, list, and tuple
F.normalize(img, 0.5, 0.5)
F.normalize(img, 1, 1)
F.normalize(img, [0.5, 0.5, 0.5], [0.5, 0.5, 0.5])
F.normalize(img, (0.5, 0.5, 0.5), (0.5, 0.5, 0.5))

# Call with jit will work on both list and float
torch.jit.script(tc_list)
torch.jit.script(tc_float)

# Call with jit trigger error for tuple and float
try:
    torch.jit.script(tc_tuple)
except Exception as e:
    print("Tuple ERROR: {}".format(e))
try:
    torch.jit.script(tc_int)
except Exception as e:
    print("Int ERROR: {}".format(e))
```

As for the tuple, I think it should be resolved by using typing.Sequence instead of typing.List. However unfortunately, it is not supported as of now: https://pytorch.org/docs/stable/jit_language_reference.html#unsupported-typing-constructs.

**Note: if we use `List[float]` instead of `Union[float, List[float]]` as the type of mean in F.normalize, then it will accept tuple even when called with torch.jit.script. So I think the `Union` somehow remove the autocast from tuple to list in torch.jit.script.

For the int, I think it is reasonable enough not to support this because usually the value of mean and std should be floating point rather than int.

datumbox:
@YosuaMichael Thanks for the detailed analysis. The reasons you mentioned above are why I'm skeptical merging this. I feel the same for my own PR at #5568. It feels like we fix something but break something else. What are your thoughts on this?

BTW have you raised a ticket on PyTorch for the specific problem? It might be worth raising it to get it fixed as this way it will make merging this PR a no-brainer.

YosuaMichael:
> @YosuaMichael Thanks for the detailed analysis. The reasons you mentioned above are why I'm skeptical merging this. I feel the same for my own PR at #5568. It feels like we fix something but break something else. What are your thoughts on this?
> 
> BTW have you raised a ticket on PyTorch for the specific problem? It might be worth raising it to get it fixed as this way it will make merging this PR a no-brainer.

@datumbox ,I also feel like it should not be merged because of this since it is very likely people use Tuple on mean / std.

I will raise an issue on the pytorch github soon, will update here again.

datumbox:
@YosuaMichael Thanks for being an awesome contributor. Please stick around and send more PRs. :)

For exactly the same reasons you mention, I decided to also close the other PR. See https://github.com/pytorch/vision/pull/5568#issuecomment-1063981861

Let's close this PR, raise the issue to PyTorch core and reassess on the near future once the issue is patched.

YosuaMichael:
@datumbox Raised this issue on Pytorch: https://github.com/pytorch/pytorch/issues/74030

Yeah, will close this PR for now.

PR #5572, opened 2022-03-09
Standardize errors for when pre-trained weights are not available

datumbox:
Our current implementation throws a `ValueError` if pre-trained weights are requested for an architecture that we don't offer checkpoints. In this PR we switch to `NotImplementedError` exceptions and we standardize the error messages.

### DIFF
--- a/torchvision/models/convnext.py
+++ b/torchvision/models/convnext.py
@@ -197,7 +197,7 @@ def _convnext(
     model = ConvNeXt(block_setting, stochastic_depth_prob=stochastic_depth_prob, **kwargs)
     if pretrained:
         if arch not in _MODELS_URLS:
-            raise NotImplemented(f"No checkpoint is available for model type {arch}")
+            raise NotImplementedError(f"No checkpoint is available for model type {arch}")
         state_dict = load_state_dict_from_url(_MODELS_URLS[arch], progress=progress)
         model.load_state_dict(state_dict)
     return model
--- a/torchvision/models/convnext.py
+++ b/torchvision/models/convnext.py
@@ -197,7 +197,7 @@ def _convnext(
     model = ConvNeXt(block_setting, stochastic_depth_prob=stochastic_depth_prob, **kwargs)
     if pretrained:
         if arch not in _MODELS_URLS:
-            raise NotImplementedError(f"No checkpoint is available for model type {arch}")
+            raise NotImplementedError(f"No pre-trained weights are available for model type {arch}")
         state_dict = load_state_dict_from_url(_MODELS_URLS[arch], progress=progress)
         model.load_state_dict(state_dict)
     return model
--- a/torchvision/models/convnext.py
+++ b/torchvision/models/convnext.py
@@ -197,7 +197,7 @@ def _convnext(
     model = ConvNeXt(block_setting, stochastic_depth_prob=stochastic_depth_prob, **kwargs)
     if pretrained:
         if arch not in _MODELS_URLS:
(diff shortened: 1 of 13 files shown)


NicolasHug (commented):
Thanks for the PR @datumbox . What is the logic for using a NotImplementedError here? At first sight it seems that ValueError may be more relevant (docs of the [NotImplementedError](https://docs.python.org/3/library/exceptions.html#NotImplementedError) exception, for ref)

torchvision/models/detection/faster_rcnn.py
```diff
@@ -402,7 +402,7 @@ def fasterrcnn_resnet50_fpn(
 
 
 def _fasterrcnn_mobilenet_v3_large_fpn(
-    weights_name,
+    arch,
```
jdsgomes, line 405: Weights name or key might be more accurate in my opinion. The reason for that is that this key represents a architecture plus dataset

jdsgomes (commented):
I left a bit but overall I'm aligned with Nicolas point that NotImplemented might not be the best candidate here

torchvision/models/convnext.py
```diff
@@ -197,7 +197,7 @@ def _convnext(
     model = ConvNeXt(block_setting, stochastic_depth_prob=stochastic_depth_prob, **kwargs)
     if pretrained:
         if arch not in _MODELS_URLS:
-            raise ValueError(f"No checkpoint is available for model type {arch}")
+            raise NotImplementedError(f"No checkpoint is available for model type {arch}")
```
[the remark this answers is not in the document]
datumbox: I was reusing the names we currently had. Happy to change.

torchvision/models/detection/faster_rcnn.py
datumbox: I agree with your comment for this specific case. The problem is that arch is actually used in many places as:
https://github.com/pytorch/vision/blob/f15ba56fec946e97a1f675e16e0c640a6284d337/torchvision/models/mobilenetv3.py#L312-L314

Here it does mean architecture. What I'm doing in this PR, is just aligning the name of the variable everywhere for the shake of consistency.

I can undo the changes here but I propose to not sweat about it because all the `weight_names` will be gone anyway with the Multi-weight API in the next couple of weeks. Thoughts?

datumbox:
@NicolasHug I think you changed your opinion on this because this change was first proposed by you here: https://github.com/pytorch/vision/pull/5022#discussion_r761123868

Personally I think throwing `ValueError` is not the right option for this. The models can throw multiple ValueError exceptions if not parameterized correctly. Throwing `NotImplementedError` for something that is not supported is in the spirit of the docs.

NicolasHug:
> @NicolasHug I think you changed your opinion on this because this change was first proposed by you here: https://github.com/pytorch/vision/pull/5022#discussion_r761123868

For the models modified in this PR, are we absolutely sure that we will implement the pretrained weights in the future (as it was the case in https://github.com/pytorch/vision/pull/5022#discussion_r761123868) ? If not, the Python docs explicitly recommend against using NotImplementedError.

datumbox:
@NicolasHug Yes, there is nothing stopping us from providing these models on the future and nor I can think of a case were we offer a model builder but we absolutely don't want to offer weights. Also note that the models missing are on the list to be added and can be contributed by the community #2707

NicolasHug:
I believe we rely on ValueError much more often than on  NotImplementedError in the current code-base for such scenarios. I won't fight against it but sorry, I prefer not to approve it.

PR #5573, opened 2022-03-09
Update the tests to exploit NMS' stable sort

### LINKED ISSUE #4970: Remove workarounds in tests that are no longer flaky due to stable sort
In the #4767 PR we enabled stable sort in the NMS op kernel. This can potentially resolve some of  the flakiness seen in the tests can be resolved, for instance:

https://github.com/pytorch/vision/blob/f093d0821194ed55f80f4764cf56e744f9217533/test/test_models.py#L658-L672

We should investigate if this is the case, and potentially adjust the the expected values of the tests if required. 

cc @NicolasHug @pmeier @datumbox

datumbox:
Fixes #4970

- Removes the workarounds for the unstable sort from the tests
- Switches `topk` calls for stable sort from all postprocessing methods
- Fixes a couple of unrelated bugs on the tests

Needs benchmarking if the tests pass


test/test_models.py
```diff
@@ -81,7 +81,7 @@ def _get_expected_file(name=None):
 
     # Note: for legacy reasons, the reference file names all had "ModelTest.test_" in their names
     # We hardcode it here to avoid having to re-generate the reference files
-    expected_file = expected_file = os.path.join(expected_file_base, "ModelTester.test_" + name)
+    expected_file = os.path.join(expected_file_base, "ModelTester.test_" + name)
```
datumbox, line 84: Unrelated fix that might need to be cherrypicked

test/test_models.py
```diff
@@ -665,6 +665,7 @@ def check_out(out):
         assert len(out) == 1
 
         def compact(tensor):
+            tensor = tensor.cpu()
```
datumbox, line 668: Unrelated fix that might need to be cherrypicked

torchvision/models/detection/fcos.py
```diff
@@ -502,7 +502,8 @@ def postprocess_detections(
 
                 # keep only topk scoring predictions
                 num_topk = det_utils._topk_min(topk_idxs, self.topk_candidates, 0)
-                scores_per_level, idxs = scores_per_level.topk(num_topk)
+                scores_per_level, idxs = scores_per_level.sort(dim=-1, descending=True, stable=True)
+                scores_per_level, idxs = scores_per_level[:num_topk], idxs[:num_topk]
```
datumbox, line 506: topk sorts by default so this should not be slower than what we currently do but needs to be benchmarked

datumbox:
Unfortunately the stable sort is not enough to fix the [flakiness](https://app.circleci.com/pipelines/github/pytorch/vision/15484/workflows/8bb7debd-e5a5-4181-a47f-1a368ee77c0c/jobs/1251191). I think the problem will be resolved once we change the testing strategy to load pretrained weights and use actual data in the tests. I think this investigation can close #4970 as nofix.

PR #5575, opened 2022-03-09
NOMRG Initial draft of models docs revamp

NicolasHug:
This is an initial draft of a proposal for revamping the models docs. The basic structure would be:

- a [main Models page](https://1251474-73328905-gh.circle-artifacts.com/0/docs/prototype_models.html) with a similar structure to what we have now (classification, detection, optical flow, etc.). This page contains links to:
    - [One page per model](https://1251474-73328905-gh.circle-artifacts.com/0/docs/prototype_models/resnet.html) - i.e. one page for Resnet, one for VGG, etc. Each of these pages contain links to
        - [one page per model builder](https://1251474-73328905-gh.circle-artifacts.com/0/docs/prototype_models/generated/torchvision.prototype.models.resnet18.html#torchvision.prototype.models.resnet18). The model builders would document the possible Weights that can be passed as parameter. The way this would be done is TBD.

### DIFF
--- a/.circleci/unittest/linux/scripts/install.sh
+++ b/.circleci/unittest/linux/scripts/install.sh
@@ -38,3 +38,4 @@ fi
 
 printf "* Installing torchvision\n"
 python setup.py develop
+pip install git+https://github.com/pytorch/data.git
--- a/docs/source/prototype_models.rst
+++ b/docs/source/prototype_models.rst
@@ -1,0 +1,23 @@
+.. _prototype_models:
+
+Prototype Models
+################
+
+Classification
+==============
+
+The following classification models are available, with or without pre-trained
+weights:
+
+.. toctree::
+   :maxdepth: 1
+
+   prototype_models/resnet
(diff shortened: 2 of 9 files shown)


docs/source/prototype_models.rst
```diff
@@ -0,0 +1,23 @@
+.. _prototype_models:
+
+Prototype Models
+################
```
NicolasHug, line 4: When the porting from prototype to main is done, this will just become "Models" and the `prototype_models.rst` file will just become (and overwrite) `models.rst`

.circleci/unittest/linux/scripts/install.sh
```diff
@@ -38,3 +38,4 @@ fi
 
 printf "* Installing torchvision\n"
 python setup.py develop
+pip install git+https://github.com/pytorch/data.git
```
NicolasHug, line 41: just a quick hack, will remove eventually

PR #5576, opened 2022-03-09
Fix minor issues with model tests.

datumbox:
Cherrypicking minor fixes from #5573

- Removing unnecessary variable
- Moving output from GPU to CPU to make comparison on Detection tests


NicolasHug (approved):
Thanks @datumbox LGTM when green

PR #5584, opened 2022-03-10
Food101 new dataset api

yassineAlouini:
In this PR, I migrate the `Food101` dataset to the new datasets API. For the code, I have got some inspiration from the `dtd.py` new dataset particulary for the `Demultiplexer` usage.

The pre-commits and food101 tests seem to work locally, hope that's the case within the CI/CD.

Any reviews are welcome, many thanks in advance. :smiley_cat:


torchvision/datasets/food101.py
```diff
@@ -7,6 +7,8 @@
 from .utils import verify_str_arg, download_and_extract_archive
 from .vision import VisionDataset
 
+# TODO: Update with new API.
+
```
pmeier, line 11: You don't need to do anything here.

torchvision/prototype/datasets/_builtin/food101.py
```diff
@@ -0,0 +1,111 @@
...
+
+    The Food-101 is a challenging data set of 101 food categories, with 101'000 images.
+    For each class, 250 manually reviewed test images are provided as well as 750 training images.
+    On purpose, the training images were not cleaned, and thus still contain some amount of noise.
+    This comes mostly in the form of intense colors and sometimes wrong labels. All images were
+    rescaled to have a maximum side length of 512 pixels.
+
+    TODO: Add more details?
+
+    """
```
pmeier, line 35: The docstring of this class is not visible in the documentation. All information needs to be supplied to the `DatasetInfo` in `_make_info`. You already put the homepage there and currently we don't have a field for an extra description.

```suggestion
```

Correct me if I'm wrong, but the description is copy-pasted from the old dataset, correct? Thus, we lose no work / information if we delete it here for now and re-add it later if we have support for descriptions.

torchvision/prototype/datasets/_builtin/food101.py
```diff
@@ -0,0 +1,111 @@
...
+            homepage="https://data.vision.ee.ethz.ch/cvl/datasets_extra/food-101",
+            # Only train and test splits, no validation.
+            valid_options=dict(split=("train", "test")),
+        )
+
+    def resources(self, config: DatasetConfig) -> List[OnlineResource]:
+        return [
+            HttpResource(
+                url=self._URL,
+                sha256=self._SHA256,
```
pmeier, line 63: Given that they seem to be only used here, there is no need to have separate class variables that hold this information

```suggestion
                url="http://data.vision.ee.ethz.ch/cvl/food-101.tar.gz",
                sha256="d97d15e438b7f4498f96086a4f7e2fa42a32f2712e87d3295441b2b6314053a4",
```

torchvision/prototype/datasets/_builtin/food101.py
```diff
@@ -0,0 +1,111 @@
...
+            return False
+
+    def _generate_categories(self, root: Path) -> List[str]:
+        resources = self.resources(self.default_config)
+        dp = resources[0].load(root)
+        dp = Filter(dp, path_comparator("name", self._CATEGORIES_FILE_NAME))
+        # TODO: Probably not the best way to parse, but it works for now. Improve later?
+        dp = CSVParser(dp, delimiter=" ")
+        categories = [sample[0] for sample in dp]
+        return cast(List[str], categories)
```
pmeier, line 93: Could you move this below `_make_datapipe` since this is not used during regular operation?

torchvision/prototype/datasets/_builtin/food101.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def _classify_archive(self, data: Tuple[str, Any]) -> Optional[int]:
+        path = Path(data[0])
+        if path.parents[1].name == "images":
+            return Food101Demux.IMAGES
+        elif path.parents[0].name == "meta":
+            return Food101Demux.METADATA
+        else:
+            # README and licence_agreement files.
+            return Food101Demux.EXTRA
```
pmeier, line 49: Could you move this below `resources()` to have all methods that are used for that datapipe in one place?

torchvision/prototype/datasets/_builtin/food101.py
```diff
@@ -0,0 +1,111 @@
...
+    def _make_datapipe(
+        self,
+        resource_dps: List[IterDataPipe],
+        *,
+        config: DatasetConfig,
+    ) -> IterDataPipe[Dict[str, Any]]:
+        archive_dp = resource_dps[0]
+        metadata_dp, images_dp, _ = Demultiplexer(
+            archive_dp, 3, self._classify_archive, drop_none=True, buffer_size=INFINITE_BUFFER_SIZE
+        )
```
pmeier, line 104: It seems the third datapipe is not used. In that case, you can capitalize on `drop_none=True` here:

```suggestion
        metadata_dp, images_dp = Demultiplexer(
            archive_dp, 2, self._classify_archive, drop_none=True, buffer_size=INFINITE_BUFFER_SIZE
        )
```

torchvision/prototype/datasets/_builtin/food101.py
pmeier, line 49: See above

```suggestion
            return None
```

torchvision/prototype/datasets/_builtin/food101.py
```diff
@@ -0,0 +1,111 @@
...
+    getitem,
+    INFINITE_BUFFER_SIZE,
+)
+from torchvision.prototype.features import Label, EncodedImage
+
+
+class Food101Demux(enum.IntEnum):
+    METADATA = 0
+    IMAGES = 1
+    EXTRA = 2
```
pmeier, line 21: Do we actually need this enum? IIUC, we only use it once to classify the archive. If that is the case, this is unneeded overhead. Just hardcode `0` and `1` there. Enums are useful if we need to have access to hardcoded values multiple times.

pmeier (commented):
Hey @yassineAlouini thanks a lot for the PR. I did an initial pass and have some comments inline. In general this looks already quite good and we only need to iron out some wrinkles.

torchvision/prototype/datasets/_builtin/food101.py
yassineAlouini: Yes, that's correct, no need to keep it, we can remove it.

torchvision/prototype/datasets/_builtin/food101.py
yassineAlouini: That's a great way to do it, thanks for the tip.

yassineAlouini:
@pmeier CLA is uploaded, let me know if I need to finish anything else. Thanks for your help and guidance.

torchvision/prototype/datasets/_builtin/food101.py
```diff
@@ -0,0 +1,86 @@
...
+)
+from torchvision.prototype.features import Label, EncodedImage
+
+
+class Food101(Dataset):
+    def _make_info(self) -> DatasetInfo:
+        return DatasetInfo(
+            "food101",
+            homepage="https://data.vision.ee.ethz.ch/cvl/datasets_extra/food-101",
+            # Only train and test splits, no validation.
```
pmeier, line 22: That should be obvious from only specifying the two. Datasets are weird. No one is going to bat an eye for "non-standard" option :wink:

torchvision/prototype/datasets/_builtin/food101.py
```diff
@@ -0,0 +1,86 @@
...
+            valid_options=dict(split=("train", "test")),
+        )
+
+    def resources(self, config: DatasetConfig) -> List[OnlineResource]:
+        return [
+            HttpResource(
+                url="http://data.vision.ee.ethz.ch/cvl/food-101.tar.gz",
+                sha256="d97d15e438b7f4498f96086a4f7e2fa42a32f2712e87d3295441b2b6314053a4",
+                decompress=True,
```
pmeier, line 31: Did you verify that the compression is actually a bottleneck here? If not could you run a small benchmark like

```py
import time

tic = time.perf_counter()
dataset = datasets.load("food101")
next(iter(dataset))
tac = time.perf_counter()
for sample in dataset.map(transforms.DecodeImage()):
    pass
toc = time.perf_counter()
print("warmup", toc - tic)
print("iter", toc - tac)
```

Note that in order to test the uncompressed variant, you temporarily need to remove the file from the data folder (`~/.cache/torch/datasets/vision/food101`) and remove `decompress=True`.

torchvision/prototype/datasets/_builtin/food101.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def _img_in_metadata(self, data: Tuple[str, Any], *, metadata: Dict[str, List[str]]) -> bool:
+        # Check whether or not the file is contained in the list of the splitted metdata.
+        path = Path(data[0])
+        category = Path(path).parent.name
+        if category in metadata:
+            img_ids = {e.split("/")[-1] for e in metadata[category]}
+            return path.stem in img_ids
+        else:
+            return False
```
pmeier, line 60: This helper is replaced by a new one for the suggestion I made above.

```suggestion
    def _image_key(self, data: Tuple[str, Any]) -> str:
        path = Path(data[0])
        return path.relative_to(path.parents[1]).with_suffix("").as_posix()
```

torchvision/prototype/datasets/_builtin/food101.py
```diff
@@ -0,0 +1,86 @@
...
+        else:
+            return None
+
+    def _prepare_sample(self, data: Tuple[str, BinaryIO]) -> Dict[str, Any]:
+        image_path, image_buffer = data
+        category = Path(image_path).parent.name
+        return dict(
+            image=EncodedImage.from_file(image_buffer),
+            label=Label.from_category(category, categories=self.categories),
+        )
```
pmeier, line 50: Small changes for the suggestion above.

```suggestion
    def _prepare_sample(self, data: Tuple[str, Tuple[str, BinaryIO]]) -> Dict[str, Any]:
        id, (path, buffer) = data
        return dict(
            label=Label.from_category(id.split("/", 1)[0], categories=self.categories),
            path=path,
            image=EncodedImage.from_file(buffer),
        )
```

pmeier (commented):
Hey @yassineAlouini, I have some suggestions inline. I have yet to review the mock data for the tests. Will do that in the next review round.

torchvision/prototype/datasets/_builtin/food101.py
yassineAlouini: haha, indeed, will remove it. :+1:

torchvision/prototype/datasets/_builtin/food101.py
```diff
@@ -0,0 +1,86 @@
...
+        images_dp = hint_shuffling(images_dp)
+        return Mapper(images_dp, self._prepare_sample)
+
+    def _generate_categories(self, root: Path) -> List[str]:
+        resources = self.resources(self.default_config)
+        dp = resources[0].load(root)
+        dp = Filter(dp, path_comparator("name", "classes.txt"))
+        dp = CSVParser(dp, delimiter=" ")
+        categories = [sample[0] for sample in dp]
+        return cast(List[str], categories)
```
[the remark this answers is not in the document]
yassineAlouini: That's what I was looking for, thanks for the suggestion.

torchvision/prototype/datasets/_builtin/food101.py
```diff
@@ -0,0 +1,86 @@
...
+        images_dp, metadata_dp = Demultiplexer(
+            archive_dp, 2, self._classify_archive, drop_none=True, buffer_size=INFINITE_BUFFER_SIZE
+        )
+        metadata_dp = Filter(metadata_dp, path_comparator("name", f"{config.split}.json"))
+        metadata_dp = JsonParser(metadata_dp)
+        metadata = next(iter(Mapper(metadata_dp, getitem(1))))
+        images_dp = Filter(images_dp, functools.partial(self._img_in_metadata, metadata=metadata))
+        images_dp = hint_sharding(images_dp)
+        images_dp = hint_shuffling(images_dp)
+        return Mapper(images_dp, self._prepare_sample)
```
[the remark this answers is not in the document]
yassineAlouini: This suggestion makes sense. I was aware that `metadata = next(iter(Mapper(metadata_dp, getitem(1))))` isn't the best way to do it but was missing some of the existing tools so thanks for making this clear.
Most of the building blocks should be clear enough now. :ok_hand:

torchvision/prototype/datasets/_builtin/food101.py
yassineAlouini: That's more clean, thanks.

torchvision/prototype/datasets/_builtin/food101.py
yassineAlouini: Questions: is returning the `path` something recommended? Also, shouldn't the `id` variable be given another name (since id is an internal Python variable) or is that fine here given the context? Thanks.

torchvision/prototype/datasets/_builtin/food101.py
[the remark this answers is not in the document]
yassineAlouini: @pmeier Something I am not sure to understand is how come some CI tests fail but some PRs are merged nonetheless? Are some checks optional? Thanks for the clarification.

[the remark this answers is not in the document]
yassineAlouini: Thanks for the clarification. The CI system here is quite impressive, hope to learn more about it in the upcoming contributions. :ok_hand:

torchvision/prototype/datasets/_builtin/food101.py
pmeier: > is returning the `path` something recommended?

Yes. We had a couple of user requests asking us to give them the ability to "perform surgery" on failing examples. For example, if your pipeline fails on a specific example, you might want to look at the file to see what is going on. Without the path, this is impossible.

> Also, shouldn't the `id` variable be given another name (since id is an internal Python variable) or is that fine here given the context?

We can, but this is nothing major. The overwrite is constrained to 5 lines and TBH I've never used `id()` anyway. I'm open to suggestions though.

yassineAlouini: I see. Makes sense to suggest having access to the path variable for debug. id should be fine indeed, I have almost never used `id()` as well. :smile_cat:

torchvision/prototype/datasets/_builtin/food101.py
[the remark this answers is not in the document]
yassineAlouini: One last question: I have this weird behavior, where the `HttpResource` mentions a `preprocess` and `loader`
parameters in the `kwargs`, so I do this below:


```
class HttpResource(OnlineResource):
    def __init__(
        self, url: str, *, file_name: Optional[str] = None, mirrors: Sequence[str] = (), **kwargs: Any
    ) -> None:
        # Had to add this...
        if "preprocess" in kwargs:
            del kwargs["preprocess"]
        if "loader" in kwargs:
            del kwargs["loader"]
        super().__init__(file_name=file_name or pathlib.Path(urlparse(url).path).name, **kwargs)
        self.url = url
        self.mirrors = mirrors
        self._resolved = False
```

This happens when downloading the dataset. Once it is in the `.cache` folder, there is no more issue.
Any idea why this happens? @pmeier

torchvision/prototype/datasets/_builtin/food101.py
yassineAlouini: Without the decompress=True, it is taking a very long time (I was expecting few seconds of run time but it is still running even after few minutes) and I am having this warning:


```
/home/yalouini/miniconda3/envs/torchvision/lib/python3.9/site-packages/torch/utils/data/datapipes/iter/combining.py:203: UserWarning: Some child DataPipes are not exhausted when __iter__ is called. We are resetting the buffer and each child DataPipe will read from the start again.
  warnings.warn("Some child DataPipes are not exhausted when __iter__ is called. We are resetting "
```

I will keep the benchmark that you have suggested running and see later how long it took but something looks weird here for sure.

pmeier: The warning is expected, since we start the iteration twice. Here is a tweaked benchmark code:

```py
import time

tic = time.perf_counter()

dataset = iter(datasets.load("food101").map(transforms.DecodeImage()))
next(dataset)

tac = time.perf_counter()

num_samples = 1
for sample in dataset:
    num_samples += 1
    
toc = time.perf_counter()

print(f"warmup: {tac - tic:.2f} seconds")
print(f"iter: {num_samples / (toc - tac):.1f} samples per second")
```

torchvision/prototype/datasets/_builtin/food101.py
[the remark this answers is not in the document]
pmeier: Those are old keyword arguments that were removed quite some time ago in #5282. My guess is that something is wrong with your development environment. Have you installed `torchvision` in development mode, i.e. `python setup.py develop`? One thing you could try is to `pip uninstall torchvision` multiple times until pip complains and afterwards install it again.

pmeier (commented):
@yassineAlouini Could you please resolve review comments that you have addressed. That makes it much easier for me to see at a glance what is still work in progress and where I might need to help out.

torchvision/prototype/datasets/_builtin/food101.py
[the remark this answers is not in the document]
yassineAlouini: I will give this a look, thanks for the suggestion. 
So far, the installation I have done is using `setup.py` and it did work well at the start.

torchvision/prototype/datasets/_builtin/food101.py
yassineAlouini: I will have some dedicated time this afternoon at work, so will check it, thanks.
I am also resolving comments that have been fixed (first time working on a PR on github, used to gitlab and bitbucket). :ok_hand:

pmeier: No worries :slightly_smiling_face: Thanks for the continued work on this!

yassineAlouini: So, managed to run the warmup check and it is quite long (I guess it should be few seconds at most) =>

`warmup: 98.69 seconds`

What remains (the sample loop) is still running as expected given the warmup time. 

Could it be something related to my computer? I will give it a try on my personal desktop later this weekend and check how it performs. 

Here are some of the specs of the computer running the benchmark:

computer_specs

pmeier: > So, managed to run the warmup check and it is quite long (I guess it should be few seconds at most) =>

Yes, everything above let's say 5 seconds is worth investigating / fixing.

> warmup: 98.69 seconds

Is this with or without compression? What timing do you get for the other variant?

yassineAlouini: So I did a CIFAR benchmark to have a comparaison point and it seems it is doing fine, so probably not an issue with my computer:

cifar_benchmark

I will check how the dataset is coded and report back.

pmeier: Before you go any deeper into this, could you please answer if you are seeing a similar warmup time for compressed and decompressed files? Because this very much looks like a case where decompressing the files is warranted.

yassineAlouini: So, managed to run the warmup check and it is quite long (I guess it should be few seconds at most) =>

Yes, everything above let's say 5 seconds is worth investigating / fixing.

warmup: 98.69 seconds

Is this with or without compression? What timing do you get for the other variant?

That's the time without any flag turned to True. I will set decompress=True and report back.

yassineAlouini: @pmeier Yes, it looks like the dataset is much bigger than the CIFAR10 one (as I was expecting it). The check is in progress, will report as soon as it is done.

yassineAlouini: Alright, it looks much better with the` decompress=True` =>

```
warmup: 12.22 seconds
iter: 673.5 samples per second
```
Will try with `extract=True` enabled as well.

yassineAlouini: With extract=True as well =>

```
warmup: 12.90 seconds
iter: 560.0 samples per second
```
Warmup is about the same and iteration are slightly better. I guess we could push this if we want as much gains as possible.

pmeier: > iteration are slightly better

Nope. More samples per second is better. Thus, `decompress=True` gives the best performance here so let's go with that. The high warm-up time is bad, but probably can't be avoided.

pmeier: Heads-up: after #5669, the flag should be `preprocess="decompress"`.

yassineAlouini: That's great, I will merge the latest main and update the flag. Thanks @pmeier.

pmeier: I'll have a look why we have that much warmup and will get back to you.

pmeier:
@yassineAlouini I see you are updating the PR from the current main branch. Please let me know if the PR is ready for another round of review.

torchvision/prototype/datasets/_builtin/food101.py
```diff
@@ -0,0 +1,97 @@
...
+    ) -> IterDataPipe[Dict[str, Any]]:
+        archive_dp = resource_dps[0]
+        images_dp, split_dp = Demultiplexer(
+            archive_dp, 2, self._classify_archive, drop_none=True, buffer_size=INFINITE_BUFFER_SIZE
+        )
+        split_dp = Filter(split_dp, path_comparator("name", f"{config.split}.json"))
+        split_dp = JsonParser(split_dp)
+        split_dp = Mapper(split_dp, getitem(1))
+        split_dp: IterDataPipe[List[str]] = MappingIterator(split_dp, drop_key=True)
+        split_dp = UnBatcher(split_dp)
```
pmeier, line 78: While looking into the raw data again, I saw that there are `{train, test}.txt` files that contain all images for the given split. That seems a lot easier than what we are doing here. Was there a reason to use the JSON files in the first place?

pmeier (commented):
One comment about the meta data files. Mock data generation looks good in general. I'll have another look after the comment is resolved, since it affects the generation.

pmeier:
Also, don't worry about CI at the moment. There is a larger outage.

torchvision/prototype/datasets/_builtin/food101.py
yassineAlouini: No particular reason except using the same JSON files as in the previous dataset. I guess using `LineReader` is what you have in mind?

pmeier: Yeah, if I'm not mistaken you can do

```suggestion
        split_dp = Filter(split_dp, path_comparator("name", f"{config.split}.json"))
        split_dp = LineReader(split_dp, decode=True, return_path=False)
```

which is a good deal simpler without losing any information.

yassineAlouini: That's even better, thanks for spotting this simplification. :+1:

yassineAlouini: I will update the mock accordingly.

test/builtin_dataset_mocks.py
```diff
@@ -911,6 +911,61 @@ def country211(info, root, config):
     return num_examples * len(classes)
 
 
+@register_mock
+def food101(info, root, config):
+    data_folder = root / "food101"
```
pmeier, line 916: Please re-check the original data, but I think we need this dash here.

```suggestion
    data_folder = root / "food-101"
```

test/builtin_dataset_mocks.py
```diff
@@ -911,6 +911,61 @@ def country211(info, root, config):
...
+
+    num_images_per_class = 3
+    image_folder = data_folder / "images"
+    categories = ["apple_pie", "baby_back_ribs", "baklava"]
```
pmeier, line 920: It is always good practice to include the first and last category to avoid problems for these "extreme" cases.

```suggestion
    categories = ["apple_pie", "baby_back_ribs", "waffles"]
```

test/builtin_dataset_mocks.py
```diff
@@ -911,6 +911,61 @@ def country211(info, root, config):
...
+                category,
+                file_name_fn=lambda idx: f"{idx:04d}.jpg",
+                num_examples=num_images_per_class,
+            )
+        ]
+        for category in categories
+    }
+
+    # Remove the extension from the list of images per category.
+    image_ids_per_category = {k: [e.replace(".jpg", "") for e in v] for k, v in image_ids_per_category.items()}
```
pmeier, line 935: We actually don't need the extra split into categories here. Thus, it is easier to keep all ids in one flat list.

```suggestion
    image_ids = []
    for category in categories:
        image_files = create_image_folder(
            image_folder,
            category,
            file_name_fn=lambda idx: f"{idx:04d}.jpg",
            num_examples=num_images_per_class,
        )
        image_ids.extend(path.relative_to(path.parents[1]).with_suffix("").as_posix() for path in image_files)
```

test/builtin_dataset_mocks.py
```diff
@@ -911,6 +911,61 @@ def country211(info, root, config):
...
+    with open(meta_folder / "test.txt", "w") as file:
+        for category in test_categories:
+            for img_id in image_ids_per_category[category]:
+                # Notice that the img_id is of the form category/img_id here.
+                file.write(f"{img_id}\n")
+
+    num_samples_map = {
+        "train": len(train_categories) * num_images_per_class,
+        "test": len(test_categories) * num_images_per_class,
+    }
```
pmeier, line 962: 1. It is somewhat dangerous to split along the categories. Given that this is a fairly simple dataset, there should be no issue, but other datasets have different behavior based on the split. Thus, we should make sure that each split sees every category at least once. This is of course that this is also the case in the original data.
2. We can simplify this by using a loop over the splits.

```suggestion
    splits = ["train", "test"]
    num_samples_map = {}
    for offset, split in enumerate(splits):
        image_ids_in_split = image_ids[offset :: len(splits)]
        num_samples_map[split] = len(image_ids_in_split)

        with open(meta_folder / f"{split}.txt", "w") as file:
            for image_id in image_ids_in_split:
                file.write(f"{image_id}\n")
```

test/builtin_dataset_mocks.py
```diff
@@ -911,6 +911,61 @@ def country211(info, root, config):
...
+
+    make_tar(root, "food-101.tar.gz", data_folder, compression="gz")
```
pmeier, line 964: With the suggestion above, we can use a shortcut here: if a folder with the same name as the archive exists in `root`, it is automatically picked up

```suggestion
    make_tar(root, f"{data_folder.name}.tar.gz", compression="gz")
```

It is also good practice to use the original folder instead of manually putting the name here to indicate this special case is used.

pmeier (commented):
Hey @yassineAlouini. I've added some suggestion and simplifications inline. Note that this doesn't mean that your version didn't work (it did!). Thanks a lot for the continued interest.

test/builtin_dataset_mocks.py
yassineAlouini: At first, I have also used the "food-101" name but then it was a hassle for naming the files and doing the imports in the code. Should I keep this name everywhere I deal with the imports with the "-" on them or is it only needed in the tests?

yassineAlouini: Got it. I thought this had an overall impact. Thanks for the clarification.

yassineAlouini:
@pmeier Mock data should be good now. Ready for a final check I guess. :ok_hand:

pmeier (approved):
LGTM, thanks a ton @yassineAlouini!

NicolasHug (approved):
Stamping

PR #5585, opened 2022-03-10
fix RandomGrayscale for grayscale inputs

### LINKED ISSUE #5581: RandomGrayscale fails for grayscale tensor inputs
#1505 added `rgb_to_grayscale` to tensor images and #2586 aligned it with the PIL op. It was missed that the PIL op is a no-op if the input is already grayscale, whereas the tensor op fails:

```py
>>> image_tensor = torch.rand(1, 16, 16)
>>> F.rgb_to_grayscale(image_tensor)
>>> image_pil = F.to_pil_image(image_tensor)
>>> F.rgb_to_grayscale(image_pil)
<PIL.Image.Image image mode=L size=16x16 at 0x7F8181384A10>
```

As the name as well as the docstring

https://github.com/pytorch/vision/blob/d0dede0e09d6d72253080cb366742a270f0ea8cd/torchvision/transforms/functional.py#L1231-L1232

imply, this transform should only handle RGB inputs so that shouldn't be an issue.

The problem is that the `RandomGrayscale` transform relies on the no-op behavior:

https://github.com/pytorch/vision/blob/d0dede0e09d6d72253080cb366742a270f0ea8cd/torchvision/transforms/transforms.py#L1613-L1616

```py
>>> transform = transforms.RandomGrayscale(p=1.0)
>>> transform(image_tensor)
>>> transform(image_pil)
<PIL.Image.Image image mode=L size=16x16 at 0x7F23E5B4E850>
```

Note the docstring is conflicting here:

https://github.com/pytorch/vision/blob/d0dede0e09d6d72253080cb366742a270f0ea8cd/torchvision/transforms/transforms.py#L1586-L1587

https://github.com/pytorch/vision/blob/d0dede0e09d6d72253080cb366742a270f0ea8cd/torchvision/transforms/transforms.py#L1595-L1596

BC compatible fix would be to patch `RandomGrayscale` to be an explicit no-op for grayscale inputs. 

cc @vfdev-5 @datumbox

pmeier:
Fixes #5581. With this fix we don't need to touch `F.rgb_to_grayscale`. The mismatch between tensor and PIL kernel is still open, but somewhat irrelevant for this. The documentation as well as the name of `F.rgb_to_grayscale` clearly state that this is converting RGB inputs. Thus, having mismatching but not wrong behavior for grayscale inputs is not a major problem, since the user is already misusing the kernel. Plus, this whole issue will go away for the prototype transforms, since we will handle color space conversions in a general way there.

`RandomGrayscale` relied on the undocumented fact that `F.rgb_to_grayscale` is a no-op for grayscale PIL. As shown in #5581, this assumption  does not hold for tensor images. This patch removes this assumption and handles grayscale inputs explicitly.

### DIFF
--- a/test/test_transforms.py
+++ b/test/test_transforms.py
@@ -2237,5 +2237,17 @@ def test_random_affine():
         assert t.interpolation == transforms.InterpolationMode.BILINEAR
 
 
+def test_random_grayscale_with_grayscale_input():
+    transform = transforms.RandomGrayscale(p=1.0)
+
+    image_tensor = torch.randint(0, 256, (1, 16, 16), dtype=torch.uint8)
+    output_tensor = transform(image_tensor)
+    torch.testing.assert_close(output_tensor, image_tensor)
+
+    image_pil = F.to_pil_image(image_tensor)
+    output_pil = transform(image_pil)
+    torch.testing.assert_close(F.pil_to_tensor(output_pil), image_tensor)
+
+
 if __name__ == "__main__":
     pytest.main([__file__])
--- a/torchvision/transforms/transforms.py
+++ b/torchvision/transforms/transforms.py
@@ -1584,7 +1584,7 @@ class Grayscale(torch.nn.Module):
 class RandomGrayscale(torch.nn.Module):
     """Randomly convert image to grayscale with a probability of p (default 0.1).
     If the image is torch Tensor, it is expected
-    to have [..., 3, H, W] shape, where ... means an arbitrary number of leading dimensions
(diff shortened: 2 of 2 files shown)


datumbox:
Thanks Philip. I confirm that this is a viable candidate solution.

As discussed offline, this approach solves the situation for the Class Transform but maintains the difference in behaviour between PIL and Tensor backends for `F.rgb_to_grayscale`. 

I think there are other candidate solutions with other pros/cons worth discussing. As this is not a priority right now and due to lack of time, let's postpone the discussion for later. To avoid having an open unreviewed PR, I'm going to close it. If after our discussion this becomes the selected approach, we will simply reopen the PR and merge.

PR #4480, opened 2021-09-24
Added gray image support to `adjust_saturation` function

VinhLoiIT:
This resolves #4466 

- [x] Update code for `adjust_contrast`
- [x] Update code for `adjust_saturation`
- [x] Update document for `adjust_contrast`
- [x] Update document for `adjust_saturation`

cc @vfdev-5 @datumbox

PR #5586, opened 2022-03-10
revert changes to assert_equal

pmeier:
#4480 added a default tolerance to `assert_equal` although this function is specifically designed to be an equality check. This change hides the fact that 

https://github.com/pytorch/vision/blob/a8bde78130fd8c956780d85693d0f51912013732/test/test_functional_tensor.py#L836

fails the batch vs single test

https://github.com/pytorch/vision/blob/a8bde78130fd8c956780d85693d0f51912013732/test/common_utils.py#L194-L198


NicolasHug (approved):
LGTM when green, thanks @pmeier

Issue #5494, opened 2022-02-28
Replace asserts with exceptions


jdsgomes:
### 🚀 The feature

I propose to replace most asserts that we find in TorchVision with an if-statement -> exception clause.

There are two patterns that I think could be modified:

### Checking input types
```python
assert isinstance(x, MyType)
``` 
can be replaced by

```python
if  not isinstance(x, MyType):
    raise TypeError(f"This function/method expects x of type MyType, instead  got {type(x}")
```

### Checking for input values

```python
assert x >= 0
```
can be replaced by 

```python
if x < 0:
    rase ValueError("x should be greater or equal to 0")
```

These changes should be done in cases where it is important to stop the normal flow and perhaps handle the exception in a particular manner. Otherwise, in my opinion, we should just use the type hints as we currently do and follow the general duck typing principle in python (so no assert or exceptions needed).

If this is implemented the test asserts should stay, the documentation ([development documentation](https://github.com/pytorch/vision/blob/main/CONTRIBUTING.md#development-process)) should be updated with a note about best practices, and other kinds of asserts and corresponding replacements should be considered.

### Motivation, pitch

Asserts do not run when we execute the python in the optimised mode, so when they perform important checks we should throw exceptions instead. If the checks are not important we can just follow the duck typing principle in python.

datumbox:
I'm only commenting because you marked with "needs discussion". I think it's a good idea as it will make the error messages more actionable by providing context on what went wrong.

Moreover `assert` is not supported by Torch-FX, so your proposal will effectively makes the entire code-base of TorchVision FX-friendly.

PR #5587, opened 2022-03-10
Replace asserts with exceptions

jdsgomes:
Changes most asserts with exceptions 
adresses https://github.com/pytorch/vision/issues/5494


references/classification/transforms.py
```diff
@@ -21,8 +21,11 @@ class RandomMixup(torch.nn.Module):
 
     def __init__(self, num_classes: int, p: float = 0.5, alpha: float = 1.0, inplace: bool = False) -> None:
         super().__init__()
-        assert num_classes > 0, "Please provide a valid positive value for the num_classes."
-        assert alpha > 0, "Alpha param can't be zero."
+        if not num_classes > 0:
+            raise ValueError("Please provide a valid positive value for the num_classes.")
+
+        if not alpha > 0:
+            raise ValueError("Alpha param can't be zero.")
```
NicolasHug, line 28: Nit
```suggestion
            raise ValueError("alpha param can't be zero.")
```

references/classification/transforms.py
```diff
@@ -21,8 +21,11 @@ class RandomMixup(torch.nn.Module):
 
     def __init__(self, num_classes: int, p: float = 0.5, alpha: float = 1.0, inplace: bool = False) -> None:
         super().__init__()
-        assert num_classes > 0, "Please provide a valid positive value for the num_classes."
-        assert alpha > 0, "Alpha param can't be zero."
+        if not num_classes > 0:
+            raise ValueError("Please provide a valid positive value for the num_classes.")
```
NicolasHug, line 25: Nit In general it's helpful for the error message to indicate which value they got, e.g. below. This isn't the original goal of the PR though, so maybe we can leave that as future improvement.

```suggestion
            raise ValueError(f"Please provide a valid positive value for the num_classes. Got num_classes={num_classes}")
```

references/optical_flow/transforms.py
```diff
@@ -7,16 +7,21 @@ class ValidateModelInput(torch.nn.Module):
...
+        if not all(arg.dtype == torch.float32 for arg in (img1, img2, flow) if arg is not None):
+            raise TypeError("This method expects the tensors img1, img2 and flow of be of dtype torch.float32.")
 
-        assert img1.shape == img2.shape
+        if not img1.shape == img2.shape:
+            raise ValueError("img1 and img2 should have the same shape.")
         h, w = img1.shape[-2:]
-        if flow is not None:
-            assert flow.shape == (2, h, w)
+        if flow is not None and not flow.shape == (2, h, w):
```
NicolasHug, line 18: I think the conversion from `assert` to `if not` can feel a bit unnatural in some places. E.g. here it would be more natural to check for inequality rather than have `if not a == b`. There are other places with similar quirks.

```suggestion
        if flow is not None and flow.shape != (2, h, w):
```

references/optical_flow/transforms.py
```diff
@@ -7,16 +7,21 @@ class ValidateModelInput(torch.nn.Module):
...
+            raise ValueError(f"flow.shape should be (2, {h}, {w}) instead of {flow.shape}")
         if valid_flow_mask is not None:
-            assert valid_flow_mask.shape == (h, w)
-            assert valid_flow_mask.dtype == torch.bool
+            if not valid_flow_mask.shape == (h, w):
```
NicolasHug, line 21:
```suggestion
            if valid_flow_mask.shape != (h, w):
```

references/optical_flow/transforms.py
```diff
@@ -7,16 +7,21 @@ class ValidateModelInput(torch.nn.Module):
...
+                raise ValueError(f"valid_flow_mask.shape should be ({h}, {w}) instead of {valid_flow_mask.shape}")
+            if not valid_flow_mask.dtype == torch.bool:
```
NicolasHug, line 23:
```suggestion
            if valid_flow_mask.dtype != torch.bool:
```

references/optical_flow/transforms.py
```diff
@@ -109,7 +114,8 @@ class RandomErasing(T.RandomErasing):
     def __init__(self, p=0.5, scale=(0.02, 0.33), ratio=(0.3, 3.3), value=0, inplace=False, max_erase=1):
         super().__init__(p=p, scale=scale, ratio=ratio, value=value, inplace=inplace)
         self.max_erase = max_erase
-        assert self.max_erase > 0
+        if not self.max_erase > 0:
```
NicolasHug, line 117: I guess normally we would write it this way?

```suggestion
        if self.max_erase <= 0:
```

NicolasHug (commented):
Thanks @jdsgomes , I only took a brief look at the first few changes. I feel like the translation of `assert X` to `if not X: raise...` can feel a bit unnatural in some places, because there is often a more Pythonic way to write `not X`. I made some comments below. I won't push too hard for this so if you feel like it's not worth it, that's fine. LMK what you think

torchvision/models/feature_extraction.py
```diff
@@ -431,17 +433,21 @@ def create_feature_extractor(
...
-        train_return_nodes is None
-    ), "If `train_return_nodes` and `eval_return_nodes` are specified, then both should be specified"
+    if (train_return_nodes is None) ^ (eval_return_nodes is None):
+        raise RuntimeError(
+            "If any of `train_return_nodes` and `eval_return_nodes` are specified, then both should be specified"
+        )
+
+    if not ((return_nodes is None) ^ (train_return_nodes is None)):
+        raise RuntimeError(
+            "If `train_return_nodes` and `eval_return_nodes` are specified, then both should be specified"
```
NicolasHug, line 449: This isn't from this PR, but I'm having a hard time understanding this error message. It reads as

> If A and B are specified, then both should be specified

Also it's metnionning `eval_return_nodes` but it should probably be `return_nodes`

torchvision/models/feature_extraction.py
```diff
@@ -497,7 +504,7 @@ def to_strdict(n) -> Dict[str, str]:
         for n in reversed(graph_module.graph.nodes):
             if n.op == "output":
                 orig_output_nodes.append(n)
-        assert len(orig_output_nodes)
+
```
NicolasHug, line 507: are we sure we want to remove this?

torchvision/models/optical_flow/raft.py
```diff
@@ -169,8 +170,10 @@ class MotionEncoder(nn.Module):
     def __init__(self, *, in_channels_corr, corr_layers=(256, 192), flow_layers=(128, 64), out_channels=128):
         super().__init__()
 
-        assert len(flow_layers) == 2
-        assert len(corr_layers) in (1, 2)
+        if len(flow_layers) != 2:
+            raise ValueError(f"The expected number of flow_layers is 2, instead got {len(flow_layers)}")
+        if not len(corr_layers) in (1, 2):
```
NicolasHug, line 175:
```suggestion
        if len(corr_layers) not in (1, 2):
```

torchvision/models/shufflenetv2.py
```diff
@@ -42,7 +42,6 @@ def __init__(self, inp: int, oup: int, stride: int) -> None:
         self.stride = stride
 
         branch_features = oup // 2
-        assert (self.stride != 1) or (inp == branch_features << 1)
```
NicolasHug, line 45: was this removed on purpose?

torchvision/ops/boxes.py
```diff
@@ -300,8 +300,10 @@ def generalized_box_iou(boxes1: Tensor, boxes2: Tensor) -> Tensor:
         _log_api_usage_once(generalized_box_iou)
     # degenerate boxes gives inf / nan results
     # so do an early check
-    assert (boxes1[:, 2:] >= boxes1[:, :2]).all()
-    assert (boxes2[:, 2:] >= boxes2[:, :2]).all()
+    if not (boxes1[:, 2:] >= boxes1[:, :2]).all():
```
NicolasHug, line 303: Nit, Same below

```suggestion
    if (boxes1[:, 2:] < boxes1[:, :2]).any():
```

torchvision/ops/poolers.py
```diff
@@ -104,15 +104,15 @@ def _infer_scale(feature: Tensor, original_size: List[int]) -> float:
...
-    assert possible_scales[0] == possible_scales[1]
     return possible_scales[0]
 
 
 @torch.fx.wrap
 def _setup_scales(
     features: List[Tensor], image_shapes: List[Tuple[int, int]], canonical_scale: int, canonical_level: int
 ) -> Tuple[List[float], LevelMapper]:
-    assert len(image_shapes) != 0
+    if len(image_shapes) == 0:
```
NicolasHug, line 114:
```suggestion
    if not image_shapes:
```

torchvision/ops/poolers.py
```diff
@@ -166,66 +166,66 @@ def _multiscale_roi_align(
...
-        return roi_align(
-            x_filtered[0],
-            rois,
-            output_size=output_size,
-            spatial_scale=scales[0],
-            sampling_ratio=sampling_ratio,
-        )
+    if scales is None or mapper is None:
+        raise ValueError("scales and mapper should not be None")
+    else:
```
NicolasHug, line 171: same about else

torchvision/utils.py
```diff
@@ -103,7 +101,8 @@ def norm_range(t, value_range):
         else:
             norm_range(tensor, value_range)
 
-    assert isinstance(tensor, torch.Tensor)
+    if not isinstance(tensor, torch.Tensor):
+        raise ValueError("tensor should be of type torch.Tensor")
```
NicolasHug, line 105:
```suggestion
        raise TypeError("tensor should be of type torch.Tensor")
```

NicolasHug (commented):
Thanks @jdsgomes , I checked all `if ...` and they look good. I made some comments regarding the use of `RuntimeError` vs `ValueError`. Looks like we're a bit inconsistent throughout the code base, but I think most of these should be ValueError. I also made some comments about some `else:` blocks that can potentially be remove to simplify the PR, LMK what you think

references/classification/transforms.py
```diff
@@ -21,8 +21,14 @@ class RandomMixup(torch.nn.Module):
 
     def __init__(self, num_classes: int, p: float = 0.5, alpha: float = 1.0, inplace: bool = False) -> None:
         super().__init__()
-        assert num_classes > 0, "Please provide a valid positive value for the num_classes."
-        assert alpha > 0, "Alpha param can't be zero."
+
+        if num_classes <= 1:
```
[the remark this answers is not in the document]
jdsgomes: This was an error, I don't think it is a bug so I will change it

torchvision/datasets/kinetics.py
```diff
@@ -118,7 +118,8 @@ def __init__(
             print("Using legacy structure")
             self.split_folder = root
             self.split = "unknown"
-            assert not download, "Cannot download the videos using legacy_structure."
+            if download:
+                raise RuntimeError("Cannot download the videos using legacy_structure.")
```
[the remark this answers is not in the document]
jdsgomes: agree - thanks for spotting these inconsistencies

test/test_backbone_utils.py
```diff
@@ -144,16 +144,16 @@ def test_build_fx_feature_extractor(self, model_name):
             model, train_return_nodes=train_return_nodes, eval_return_nodes=eval_return_nodes
         )
         # Check must specify return nodes
-        with pytest.raises(AssertionError):
+        with pytest.raises(RuntimeError):
```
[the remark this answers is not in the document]
jdsgomes: I was in doubt in this one. I was using a strict definition of ValueError: "Raised when an operation or function receives an argument that has the right type but an inappropriate value". And in this case each argument might have an appropriate value, but the combination of them is not appropriate. But I am ok with your interpretation as these are user inputs so still fall under ValueError.

As for the expected error messages I would leave it for another PR as it is outside the scope of this one.

torchvision/models/shufflenetv2.py
jdsgomes: no good catch.

torchvision/models/detection/roi_heads.py
```diff
@@ -615,11 +618,15 @@ def add_gt_proposals(self, proposals, gt_boxes):
...
-        assert all(["labels" in t for t in targets])
+        if targets is None:
+            raise ValueError("targets should not be None")
+        if not all(["boxes" in t for t in targets]):
+            raise ValueError("Every element of targets should have a boxes key")
+        if not all(["labels" in t for t in targets]):
+            raise ValueError("Every element of targets should have a labels key")
         if self.has_mask():
-            assert all(["masks" in t for t in targets])
+            if not all(["masks" in t for t in targets]):
```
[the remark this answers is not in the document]
jdsgomes: I would prefer to leave it in line with the original asserts

torchvision/models/feature_extraction.py
jdsgomes: I wanted to replace it with exception , but missed it.

torchvision/models/detection/ssdlite.py
```diff
@@ -244,7 +247,10 @@ def ssdlite320_mobilenet_v3_large(
     anchor_generator = DefaultBoxGenerator([[2, 3] for _ in range(6)], min_ratio=0.2, max_ratio=0.95)
     out_channels = det_utils.retrieve_out_channels(backbone, size)
     num_anchors = anchor_generator.num_anchors_per_location()
-    assert len(out_channels) == len(anchor_generator.aspect_ratios)
+    if len(out_channels) != len(anchor_generator.aspect_ratios):
+        raise RuntimeError(
+            f"The length of the output channels from the backbone {len(out_channels)} do not match the length of the anchor generator aspect ratios {len(anchor_generator.aspect_ratios)}"
+        )
```
datumbox, line 253: That's going to conflict with the weight porting.

torchvision/models/quantization/inception.py
```diff
@@ -216,7 +216,8 @@ def inception_v3(
         backend = "fbgemm"
         quantize_model(model, backend)
     else:
-        assert pretrained in [True, False]
+        if pretrained not in [True, False]:
+            raise ValueError(f"For non quantized models, pretrained should be a bollean value instead of {pretrained}")
```
datumbox, line 220: This one too.

torchvision/models/quantization/mobilenetv2.py
```diff
@@ -88,7 +88,8 @@ def mobilenet_v2(
         backend = "qnnpack"
         quantize_model(model, backend)
     else:
-        assert pretrained in [True, False]
+        if pretrained not in [True, False]:
+            raise ValueError(f"For non quantized models, pretrained should be a bollean value instead of {pretrained}")
```
datumbox, line 92: Same.

torchvision/models/quantization/resnet.py
```diff
@@ -125,7 +125,8 @@ def _resnet(
         backend = "fbgemm"
         quantize_model(model, backend)
     else:
-        assert pretrained in [True, False]
+        if pretrained not in [True, False]:
+            raise ValueError(f"For non quantized models, pretrained should be a bollean value instead of {pretrained}")
```
datumbox, line 129: Same.

torchvision/models/quantization/shufflenetv2.py
```diff
@@ -89,7 +89,8 @@ def _shufflenetv2(
         backend = "fbgemm"
         quantize_model(model, backend)
     else:
-        assert pretrained in [True, False]
+        if pretrained not in [True, False]:
+            raise ValueError(f"For non quantized models, pretrained should be a bollean value instead of {pretrained}")
```
datumbox, line 93: Same.

datumbox (commented):
There are a few potential conflicts with multi-weights porting noted below. This is FYI. 

Edit:
@jdsgomes With the exception of the first comment below, everything else will be removed by the porting. Shall I cherry pick the first in my branch and then undo the rest to avoid conflicts? No strong opinion if you want to keep them as you are likely to merge faster than I will.

jdsgomes:
> 

I will revert the others and keep the first one then. You can cherrypick it or resolve later if this lands first.

torchvision/models/detection/ssdlite.py
jdsgomes: please se my comment about this conflicts. I reverted all others but will leave this one let me know if you agree with this approach

jdsgomes:
thank you @NicolasHug  for all the comments! I think i've addressed everything but let me know if it looks good!

torchvision/models/feature_extraction.py
```diff
@@ -431,17 +433,19 @@ def create_feature_extractor(
...
+    if (train_return_nodes is None) ^ (eval_return_nodes is None):
+        raise ValueError(
+            "If any of `train_return_nodes` and `eval_return_nodes` are specified, then both should be specified"
+        )
 
-    assert (return_nodes is None) ^ (
-        train_return_nodes is None
-    ), "If `train_return_nodes` and `eval_return_nodes` are specified, then both should be specified"
+    if not ((return_nodes is None) ^ (train_return_nodes is None)):
+        raise ValueError("If `train_return_nodes` and `eval_return_nodes` are specified, then both should be specified")
```
NicolasHug, line 448: Should we  fix that now, or do that in another PR? 
Previous comment was https://github.com/pytorch/vision/pull/5587#discussion_r826727738

NicolasHug (approved):
Thanks @jdsgomes, LGTM

I still see a few `RuntimeError` that could be ValueError, from https://github.com/pytorch/vision/pull/5587#discussion_r826827146 I'm not sure whether you were planning to change them here in this PR, or later, or at all. Either way we are not consistent regarding this in the code-base so I will approve regardless. Thanks for the many changes!

torchvision/models/feature_extraction.py
jdsgomes: Sorry you are right, this should be done in this one. Done now. I will wait for the tests to pass and merge. Thanks for the heads up

Issue #5056, opened 2021-12-08
Follow-up improvements to RAFT training reference


NicolasHug:
This issue is to keep track of the progress on some minor improvements related to the RAFT training reference, as a follow up to https://github.com/pytorch/vision/issues/4644

- [x] support `--weights` param to support the prototype API, once pre-trained weights are available
- [x] support `--model` instead of the `--small` flag
- [x] support `--epochs` param instead of `--num-steps`, for consistency with the rest of the training references.
- [ ] save the scheduler and the optimizer, not just the model.
- [ ] try to keep function names as close as possible to existing references. For example `validate()` -> `evaluate()`
- [ ] Support `--device` (?? - not sure how useful it would be)
- [ ] Better support of weight serialization (move on CPU, avoid having the DDP model weights etc)

cc @datumbox

NicolasHug:
I beleieve all the remaining points have been addressed by @YosuaMichael in https://github.com/pytorch/vision/pull/5590. Thanks Yosua!

PR #5590, opened 2022-03-11
RAFT training reference Improvement

YosuaMichael:
Do some of the task on: https://github.com/pytorch/vision/issues/5056

- Change function name from `validate` to `evaluate`
- Support --device and enable training on non-distributed mode
- Include optimizer and scheduler in the checkpoint

Sample script to run on non-distributed mode and on cpu:
```
 python train.py \
    --dataset-root $dataset_root \
    --name $name_chairs \
    --model raft_small \
    --train-dataset chairs \
    --batch-size 2 \
    --lr 0.0004 \
    --weight-decay 0.0001 \
    --epochs 2 \
    --output-dir $out_chairs \
    --device cpu
```
To test on CPU, I run on a mock dataset by replacing https://github.com/pytorch/vision/blob/main/torchvision/datasets/_optical_flow.py with https://gist.github.com/YosuaMichael/9c49729243ff9d467ece06ab8641680d.

Note that as of now, if we run on distributed mode using torchrun, then it must use `--device cuda`.


references/optical_flow/train.py
```diff
@@ -60,16 +60,21 @@ def get_train_dataset(stage, dataset_root):
...
     batch_size = batch_size or args.batch_size
+    device = torch.device(args.device)
 
     model.eval()
 
-    sampler = torch.utils.data.distributed.DistributedSampler(val_dataset, shuffle=False, drop_last=True)
+    if args.distributed:
+        sampler = torch.utils.data.distributed.DistributedSampler(val_dataset, shuffle=False, drop_last=True)
+    else:
+        sampler = torch.utils.data.SequentialSampler(val_dataset, drop_last=True)
```
NicolasHug, line 76: This will raise an error because  `SequentialSampler` doesn't accepts a `drop_last` parameter.

```suggestion
        sampler = torch.utils.data.SequentialSampler(val_dataset)
```

references/optical_flow/train.py
```diff
@@ -115,21 +120,22 @@ def inner_loop(blob):
...
         f"Batch-processed {num_processed_samples} / {len(val_dataset)} samples. "
         "Going to process the remaining samples individually, if any."
     )
 
-    if args.rank == 0:  # we only need to process the rest on a single worker
+    if not args.distributed or args.rank == 0:  # we only need to process the rest on a single worker
         for i in range(num_processed_samples, len(val_dataset)):
             inner_loop(val_dataset[i])
 
     logger.synchronize_between_processes()
```
NicolasHug, line 134: Lines 124 to 134 (included) should actually be included in the `if args.distributed:` block. We only need to do these extra bits of work in distributed mode, because when in normal mode, the SequentialSampler will be processing all the samples properly. You should be able to remove the changes made to line 130 as well.

NicolasHug (commented):
Thank you for the PR @YosuaMichael. There are 2 minor issues (see below), but otherwise this looks great!

references/optical_flow/train.py
YosuaMichael: Ah yes, somehow I can complete the training without getting error before... But I will fix this, thanks!

YosuaMichael:
Update:

Support saving of the optimizer and scheduler on the checkpoint.

YosuaMichael:
Hi @NicolasHug , I decided to put the commit for saving optimizer and scheduler in this PR as well: https://github.com/pytorch/vision/pull/5590/commits/09d78d102088d73215dea4bcf89bf9d68ed40e53 
Could you also help to review this? Thanks!

references/optical_flow/train.py
```diff
@@ -115,21 +120,21 @@ def inner_loop(blob):
...
+        num_processed_samples = utils.reduce_across_processes(num_processed_samples)
+        print(
+            f"Batch-processed {num_processed_samples} / {len(val_dataset)} samples. "
+            "Going to process the remaining samples individually, if any."
+        )
+        if args.rank == 0:  # we only need to process the rest on a single worker
+            for i in range(num_processed_samples, len(val_dataset)):
+                inner_loop(val_dataset[i])
 
     logger.synchronize_between_processes()
```
NicolasHug, line 133: This should be part of the `if args.distributed:` block

```suggestion
        logger.synchronize_between_processes()
```

references/optical_flow/train.py
```diff
@@ -200,36 +206,64 @@ def main(args):
...
+        epochs=args.epochs,
+        steps_per_epoch=ceil(len(train_dataset) / (args.world_size * args.batch_size)),
+        pct_start=0.05,
+        cycle_momentum=False,
+        anneal_strategy="linear",
+    )
+
+    if args.resume is not None:
+        checkpoint = torch.load(args.resume, map_location="cpu")
+        model.load_state_dict(checkpoint["model"])
```
NicolasHug, line 249: This part should happen before the ``    if args.train_dataset is None:`` block (currently on line 224). Otherwise, the model won't be loaded before being evaluated and we'll just get random weights.

references/optical_flow/train.py
```diff
@@ -269,13 +293,19 @@ def main(args):
         # Note: we don't sync the SmoothedValues across processes, so the printed metrics are just those of rank 0
         print(f"Epoch {current_epoch} done. ", logger)
 
-        if args.rank == 0:
-            # TODO: Also save the optimizer and scheduler
-            torch.save(model.state_dict(), Path(args.output_dir) / f"{args.name}_{current_epoch}.pth")
-            torch.save(model.state_dict(), Path(args.output_dir) / f"{args.name}.pth")
+        if not args.distributed or args.rank == 0:
+            checkpoint = {
+                "model": model.state_dict(),
```
NicolasHug, line 298: Let's load and save the `model_without_ddp` instead of the model. This way, we would be able to seemlessly load a model that was trained with DDP in a non-DDP script, and vice-versa. Take a look at how we do that in the classification training reference - it should only take ~3 lines of code to be added. Happy to chat more about it if you need clarifications.

references/optical_flow/train.py
```diff
@@ -200,36 +206,64 @@ def main(args):
...
+    device = torch.device(args.device)
+
     if args.prototype:
         model = prototype.models.optical_flow.__dict__[args.model](weights=args.weights)
     else:
         model = torchvision.models.optical_flow.__dict__[args.model](pretrained=args.pretrained)
 
-    model = model.to(args.local_rank)
-    model = torch.nn.parallel.DistributedDataParallel(model, device_ids=[args.local_rank])
+    model.to(device)
```
NicolasHug, line 218: Nit: Instead of having that here we could instead add it just below in an else clause:

```py
if args.distributed:
    # unchanged code here
else:
    model = model.to(device)

NicolasHug (commented):
Thanks @YosuaMichael , we're almost there :) . I made a few comments bleow

PR #5596, opened 2022-03-11
Better handling for Pad's fill argument

datumbox:
Cherrypicks from #5568 which was not merged:
- Enables the use of lists in Pad
- Fixes the error that appears when we pass float pad values in PIL images
- Modifies the tests to check for the changes


vfdev-5 (approved):
Looks good to me, thanks @datumbox

PR #5602, opened 2022-03-13
add python linking dependency when USE_PYTHON=on

nihui:


de-vri-es:
Without this, if the project is built without `-DUSE_PYTHON=ON`, anyone doing `find_package(TorchVision)` will get this error:

```
-- Configuring done
CMake Error at /usr/share/cmake/TorchVision/TorchVisionConfig.cmake:61 (set_target_properties):
  The link interface of target "TorchVision::TorchVision" contains:
    Python3::Python
  but the target was not found.  Possible reasons include:
    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
```

YosuaMichael:
This seems to be related to https://github.com/pytorch/vision/issues/5863
cc @datumbox, Do you know who can review this PR?

datumbox:
Right now our CI is still broken due to PyTorch upstream issues, so it's not a good time to test it. Let's wait until they are resolved (tracked at #5881) and then check what the CI things about them.

@mthrok Would you be OK to provide input on this as you have good knowledge of cmake?

de-vri-es:
I was disappointed to see 0.13 released without this fix :(

Is there anything needed to get this moving?

PR #5603, opened 2022-03-14
[NOMERGE] Multi-weight support for Classification

datumbox:
Related to #4679


torchvision/models/_utils.py
```diff
@@ -1,7 +1,12 @@
+import functools
+import warnings
 from collections import OrderedDict
-from typing import Dict, Optional
+from typing import Any, Dict, Optional, TypeVar, Callable, Tuple, Union
 
 from torch import nn
+from torchvision.prototype.utils._internal import kwonly_to_pos_or_kw
```
pmeier, line 7: This is not going to work. We should never import prototype functionality into the main area. `torchvision.prototype` has some runtime dependencies that are not tracked regularly yet. Thus, something like that leads to failures as can be seen in [this CI run](https://app.circleci.com/pipelines/github/pytorch/vision/15651/workflows/7f5fa032-178a-4eb0-902a-e330ffc92adb/jobs/1264997).

The way to go is to move this functionality to `torchvision._utils` and import from there.

pmeier (changes requested):
I haven't checked the whole PR. If there are more cases like the one I highlighted below, we need to handle all of them.

torchvision/models/_utils.py
datumbox: Thank for helping me debug the CI problem. I didn't realize I was still importing from prototype. Still doing changes so this PR is not ready yet, let me do the change and try again.

PR #5604, opened 2022-03-14
Moving `sequence_to_str` to `torchvision._utils`

datumbox:
Related to #4679

The method `sequence_to_str` is used both by prototype models and datasets. The work on #5603 requires moving it on the `torchvision._utils` private package but during this move we modify `prototype.datasets`. As this will almost certainly lead to conflicts, I'm bringing this move on a separate PR. The method is added on a private package so future changes can be made without BC worries.


test/test_prototype_utils.py
```diff
@@ -1,5 +1,5 @@
 import pytest
-from torchvision.prototype.utils._internal import sequence_to_str
```
pmeier, line 2: We should rename this whole file to `test_internal_utils` or the like. Otherwise, the test will only be run by the prototype tests.

pmeier (approved):
Thanks! LGTM when CI is green.

PR #5610, opened 2022-03-14
Adding RandomShortestSize transform

datumbox:
Cherrypicking from #5444 the `RandomShortestSize`.

The specific transform is used to improve the accuracy of the `RetinaNet` model. It is a porting of the existing `GeneralizedRCNNTransform.resize()` method, to make it available outside of the model.


vfdev-5 (approved):
Thanks @datumbox

Issue #4839, opened 2021-11-03
Assertion error during kinetics400 validation


datumbox:
### 🐛 Describe the bug

Running on main:
```
torchrun --nproc_per_node=8 train.py --data-path /datasets01/kinetics/070618/400/ --train-dir=val --val-dir=val --batch-size=16 --sync-bn --test-only --pretrained --cache-dataset
```
throws the following error:
```
Test:  [2200/3008]  eta: 0:11:24  loss: 2.6703 (2.1475)  acc1: 43.7500 (57.4938)  acc5: 68.7500 (77.8623)  time: 0.9043  data: 0.6405  max mem: 5888
Traceback (most recent call last):
  File "train.py", line 392, in <module>
    main(args)
  File "train.py", line 273, in main
    evaluate(model, criterion, data_loader_test, device=device)
  File "train.py", line 62, in evaluate
    for video, target in metric_logger.log_every(data_loader, 100, header):
  File "/private/home/vvryniotis/vision/references/video_classification/utils.py", line 128, in log_every
    for obj in iterable:
  File "/private/home/vvryniotis/.conda/envs/datumbox/lib/python3.8/site-packages/torch/utils/data/dataloader.py", line 521, in __next__
    data = self._next_data()
  File "/private/home/vvryniotis/.conda/envs/datumbox/lib/python3.8/site-packages/torch/utils/data/dataloader.py", line 1183, in _next_data
    return self._process_data(data)
  File "/private/home/vvryniotis/.conda/envs/datumbox/lib/python3.8/site-packages/torch/utils/data/dataloader.py", line 1229, in _process_data
    data.reraise()
  File "/private/home/vvryniotis/.conda/envs/datumbox/lib/python3.8/site-packages/torch/_utils.py", line 438, in reraise
    raise exception
AssertionError: Caught AssertionError in DataLoader worker process 3.
Original Traceback (most recent call last):
  File "/private/home/vvryniotis/.conda/envs/datumbox/lib/python3.8/site-packages/torch/utils/data/_utils/worker.py", line 287, in _worker_loop
    data = fetcher.fetch(index)
  File "/private/home/vvryniotis/.conda/envs/datumbox/lib/python3.8/site-packages/torch/utils/data/_utils/fetch.py", line 49, in fetch
    data = [self.dataset[idx] for idx in possibly_batched_index]
[...]

PR #5618, opened 2022-03-15
Port Multi-weight support from prototype to main

### LINKED ISSUE #4679: Move new model builders from Prototype to main TorchVision

datumbox:
Fixes #4679

- [x] Ports all models from prototype to main area and adds documentation to the builders
- [x] Updates the docs, examples, tutorials and galleries
- [x] Updates the `test_models.py` to use `weights` and move the `test_prototype_models.py` tests
- [x] Updates the reference scripts and me `_presets.*` private
- [x] Fix BC breakings on `weights_backbone`
- [x] Test that all pretrained models are not affected and their builders work well

## Checks using existing tests

To confirm that all model builders work as expected:
- We run the tests of all models on the same expected files as on master. All tests pass for untrained models.
- We modify the `test_models.py` on the main branch to estimate new expected files with pretrained=True (see #5655). Then we run the tests on this branch against the new expected files. All tests pass for trained models.

## Checks using reference scripts

To confirm that all the reference scripts still work as expected we ran them all and compared their results with existing model numbers:
```
torchrun --nproc_per_node=2 train.py --model efficientnet_b0 --test-only --weights EfficientNet_B0_Weights.IMAGENET1K_V1
Test:  Acc@1 77.684 Acc@5 93.536

torchrun --nproc_per_node=2 train.py --model retinanet_resnet50_fpn --test-only --weights RetinaNet_ResNet50_FPN_Weights.COCO_V1
IoU metric: bbox
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.363
 Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=100 ] = 0.557
 Average Precision  (AP) @[ IoU=0.75      | area=   all | maxDets=100 ] = 0.382
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.193
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.400
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.490
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=  1 ] = 0.314
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets= 10 ] = 0.500
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.540
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.340
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.581
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.696

torchrun --nproc_per_node=2 train.py --model lraspp_mobilenet_v3_large --test-only --weights LRASPP_MobileNet_V3_Large_Weights.COCO_WITH_VOC_LABELS_V1
global correct: 91.2
average row correct: ['94.5', '84.3', '69.5', '72.8', '57.7', '42.0', '77.0', '56.9', '90.4', '36.1', '76.0', '60.8', '81.4', '78.9', '81.0', '87.6', '51.3', '83.9', '62.2', '84.2', '56.1']
IoU: ['90.2', '69.2', '57.7', '58.5', '47.8', '35.7', '69.5', '47.1', '79.1', '29.6', '62.6', '34.2', '65.5', '63.4', '70.0', '76.8', '30.1', '61.9', '46.8', '70.6', '49.1']
mean IoU: 57.9

torchrun --nproc_per_node 1 --nnodes 1 train.py --val-dataset sintel --batch-size 1 --dataset-root $dataset_root --model raft_large --weights Raft_Large_Weights.C_T_SKHT_V2
Batch-processed 1041 / 1041 samples. Going to process the remaining samples individually, if any.
Sintel val clean epe: 0.6499	1px: 0.9255	3px: 0.9701	5px: 0.9805	per_image_epe: 0.6499	f1: 2.6805
Batch-processed 1041 / 1041 samples. Going to process the remaining samples individually, if any.
Sintel val final epe: 1.0197	1px: 0.8898	3px: 0.9493	5px: 0.9660	per_image_epe: 1.0197	f1: 4.5752

torchrun --nproc_per_node=8 train.py --data-path /datasets01/kinetics/070618/400/ --train-dir=val --val-dir=val --batch-size=16 --sync-bn --test-only --weights R2Plus1D_18_Weights.KINETICS400_V1 --cache-dataset
 * Clip Acc@1 56.488 Clip Acc@5 77.745
```

The difference on the video accuracy comparing to the docs is not related to this PR. It pre-existed on main and was first reported at #4839


datumbox:
The commit 5a96c9a completes the porting. As we see, all the tests pass while using the BC compatibility layer of the solution. Aka all tests, documentation etc use of `pretrained` instead of `weights`.

.circleci/config.yml
```diff
@@ -360,19 +360,28 @@ jobs:
...
+    resource_class: xlarge
+    steps:
+      - checkout
+      - download_model_weights
+      - install_torchvision
+      - run:
+          name: Enable extended tests
+          command: echo 'export PYTORCH_TEST_WITH_EXTENDED=1' >> $BASH_ENV
+      - run_tests_selective:
+          file_or_dir: test/test_extended_*.py
```
datumbox, line 383: Moving some of the prototype tests that verify the meta-data and require model initialization to a separate test that only runs once. This will help us avoid testing them over and over in all CI jobs. We don't need to check these tests across all platforms as they don't depend on platform specific configuration.

gallery/plot_optical_flow.py
```diff
@@ -88,24 +87,19 @@ def plot(imgs, **imshow_kwargs):
...
-def preprocess(batch):
-    transforms = T.Compose(
-        [
-            T.ConvertImageDtype(torch.float32),
-            T.Normalize(mean=0.5, std=0.5),  # map [0, 1] into [-1, 1]
-            T.Resize(size=(520, 960)),
-        ]
-    )
-    batch = transforms(batch)
-    return batch
```
datumbox, line 101: I've replaced in all galleries the previous preprocessing with the new idioms. I changed only the code. It's very likely that the test in these galleries doesn't make sense any more but this can change on a separate PR.

references/classification/train.py
```diff
@@ -154,18 +148,13 @@ def load_data(traindir, valdir, args):
         print(f"Loading dataset_test from {cache_path}")
         dataset_test, _ = torch.load(cache_path)
     else:
-        if not args.prototype:
+        if args.weights and args.test_only:
+            weights = torchvision.models.get_weight(args.weights)
+            preprocessing = weights.transforms()
```
datumbox, line 153: We are going to use the bundled transforms only when we load weights and validate a model. When training happens (either full training or fine-tuning) we always want to use the presets defined in our reference scripts.

references/video_classification/presets.py
```diff
@@ -27,7 +27,7 @@ def __call__(self, x):
 
 
 class VideoClassificationPresetEval:
-    def __init__(self, resize_size, crop_size, mean=(0.43216, 0.394666, 0.37645), std=(0.22803, 0.22145, 0.216989)):
+    def __init__(self, crop_size, resize_size, mean=(0.43216, 0.394666, 0.37645), std=(0.22803, 0.22145, 0.216989)):
```
datumbox, line 30: Flipping their order to match the one of the presets in TorchVision. No BC concerns in references so that should be OK.

test/test_extended_models.py
```diff
@@ -205,43 +150,36 @@ def test_old_vs_new_factory(model_fn, dev):
...
     module_name = model_fn.__module__.split(".")[-2]
-    kwargs = {"pretrained": True, **defaults[module_name], **TM._model_params.get(model_name, {})}
-    input_shape = kwargs.pop("input_shape")
-    kwargs.pop("num_classes", None)  # ignore this as it's an incompatible speed optimization for pre-trained models
-    x = torch.rand(input_shape).to(device=dev)
-    if module_name == "detection":
-        x = [x]
 
+    kwargs = {**defaults[module_name], **TM._model_params.get(model_name, {})}
+    input_shape = kwargs.pop("input_shape")
```
datumbox, line 167: This logic could be simplified but I'm reusing the `TM._model_params` from the other tests. I would prefer to leave this as-is for now and tackle the testing revamping of models as a whole later this half.

test/test_extended_models.py
```diff
@@ -142,48 +125,6 @@ def test_schema_meta_validation(model_fn):
...
-
-
-@needs_cuda
-@pytest.mark.parametrize("model_builder", TM.get_models_from_module(models.optical_flow))
-@pytest.mark.parametrize("scripted", (False, True))
-@run_if_test_with_prototype
-def test_raft(model_builder, scripted):
-    TM.test_raft(model_builder, scripted)
-
-
```
datumbox, line 186: No more need for these tests. They are covered by the existing tests of TorchVision.

torchvision/models/detection/backbone_utils.py
```diff
@@ -55,9 +56,13 @@ def forward(self, x: Tensor) -> Dict[str, Tensor]:
         return x
 
 
+@handle_legacy_interface(
+    weights=("pretrained", True),  # type: ignore[arg-type]
```
datumbox, line 60: I'm trying to reuse the existing decorator. Passing `True` here is hacky but we can rely on the model builder invoked few lines below to throw the right warning to the user.

torchvision/transforms/_presets.py
```diff
@@ -67,53 +69,59 @@ def __init__(
...
     def forward(self, vid: Tensor) -> Tensor:
-        vid = vid.permute(0, 3, 1, 2)  # (T, H, W, C) => (T, C, H, W)
+        need_squeeze = False
+        if vid.ndim < 5:
+            vid = vid.unsqueeze(dim=0)
+            need_squeeze = True
+
+        vid = vid.permute(0, 1, 4, 2, 3)  # (N, T, H, W, C) => (N, T, C, H, W)
+        N, T, C, H, W = vid.shape
+        vid = vid.view(-1, C, H, W)
```
datumbox, line 79: I'm extending the class to handle batches not only single videos.

datumbox (commented):
Adding some clarifications:

torchvision/models/detection/keypoint_rcnn.py
```diff
@@ -296,22 +305,61 @@ def forward(self, x):
...
+            "map_kp": 61.1,
+        },
+    )
+    COCO_V1 = Weights(
+        url="https://download.pytorch.org/models/keypointrcnn_resnet50_fpn_coco-fc266e95.pth",
+        transforms=ObjectDetection,
+        meta={
+            **_COMMON_META,
+            "num_params": 59137258,
+            "recipe": "https://github.com/pytorch/vision/tree/main/references/detection#keypoint-r-cnn",
```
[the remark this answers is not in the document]
datumbox: This is a valid point.

Up to now, we make sure that all the commands either remain BC or are updated accordingly to ensure we can fully reproduce the result. As a matter of fact, in this PR I'm updating some of them to ensure we load the right backbone weights. Going forwards I think we will favour keeping track of these on the documentation. @NicolasHug is currently looking into this, so any value we put here will be temporary.

torchvision/models/detection/retinanet.py
```diff
@@ -578,14 +585,37 @@ def forward(self, images, targets=None):
...
-    "retinanet_resnet50_fpn_coco": "https://download.pytorch.org/models/retinanet_resnet50_fpn_coco-eeacb38b.pth",
-}
+class RetinaNet_ResNet50_FPN_Weights(WeightsEnum):
+    COCO_V1 = Weights(
+        url="https://download.pytorch.org/models/retinanet_resnet50_fpn_coco-eeacb38b.pth",
+        transforms=ObjectDetection,
+        meta={
+            "task": "image_object_detection",
+            "architecture": "RetinaNet",
+            "publication_year": 2017,
```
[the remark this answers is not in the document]
datumbox: You are right to say that `(1, 1)` is the right value but I felt that this is somehow misleading. It doesn't really matter what input size you provide to the model because the `GeneralizedRCNNTransform` will internally resize the image to 800 on the shortest edge and 1333 on the max. That's the reason I opted not to define it but I don't mind adding it; technically it will accept any size because internally it resizes. Thoughts?

torchvision/models/detection/ssd.py
```diff
@@ -6,27 +6,42 @@
...
+class SSD300_VGG16_Weights(WeightsEnum):
+    COCO_V1 = Weights(
+        url="https://download.pytorch.org/models/ssd300_vgg16_coco-b556d3b4.pth",
+        transforms=ObjectDetection,
+        meta={
+            "task": "image_object_detection",
+            "architecture": "SSD",
+            "publication_year": 2015,
+            "num_params": 35641826,
+            "size": (300, 300),
```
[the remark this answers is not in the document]
datumbox: Correct. This is fixed size. It doesn't matter what you provide, the `GeneralizedRCNNTransform` will set it to 300x300.

torchvision/models/detection/fcos.py
```diff
@@ -636,19 +643,37 @@ def forward(
...
+    COCO_V1 = Weights(
+        url="https://download.pytorch.org/models/fcos_resnet50_fpn_coco-99b0c9b7.pth",
+        transforms=ObjectDetection,
+        meta={
+            "task": "image_object_detection",
+            "architecture": "FCOS",
+            "publication_year": 2019,
+            "num_params": 32269600,
+            "categories": _COCO_CATEGORIES,
+            "interpolation": InterpolationMode.BILINEAR,
```
zhiqwang, line 656: Hi @datumbox ,

I just have one question here.

Seems that `GeneralizedRCNNTranform` also plays some roles of the pre-processing (pre-transforms) for an object detection task, could we set aside some attributes into the `WeightsEnum` class for further modifications or adjustments?

For example, we may also have different implementations in `GeneralizedRCNNTranform`, depending on the scaling and padding modes. And the current implementation of `GeneralizedRCNNTranform` is limited to only one padding mode (we put the image on the left-top plate). So could we make this Class more powerful and more general ?

datumbox: Thanks, great point. 

Ideally, we should explore if we can make the Object detection models to work similar to Classification, Segmentation, Optical Flow etc by removing the specific transforms from within the model. If you see PR #5444 I'm exploring adding the ability to turn `GeneralizedRCNNTranform.resize()` off to allow for more advanced augmentation strategies such as Large Scale Jitter etc. There are pros/cons for that and this is a broader investigation that I would actually like to get your input.

Concerning adding more fields to the meta, I think we should not **yet** add more until we actually decide what to do. Once this is clear to us, we will just add them. You can debate that interpolation is in the meta while not directly configurable, but the only reason it's there is for consistency across other models. Not sure if that makes sense to you. Please let me know your thoughts.

torchvision/models/detection/retinanet.py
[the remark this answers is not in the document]
oke-aditya: Then technically like SSD we should document this with size = (800, 1333) ?

I'm not sure. But somehow the aim is to not hide the fact what generalized RCNN does and what size the model expects / takes. Either the minimum size / like classification the size model takes.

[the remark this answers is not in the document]
datumbox: It's not going to be `(800, 1333)`. The `GeneralizedRCNNTransform` maintains the ratio of the image by setting the minimum edge to 800. Then if the maximum edge exceeds 1333 it resizes the image further. So the input size in the network is variable.

If we put `(1, 1)` as the `min_size`, that's technically accurate but it doesn't have the same interpretation as for the classification case. In classification this size depicts the actual size of the smallest image that can go through the network. Here the value `(1, 1)` doesn't tell us anything about the network; it just tells us that we can throw any size in because we will resize it eventually to at least 800.

More thoughts based on the above clarifications?

torchvision/models/vgg.py
```diff
@@ -95,107 +95,276 @@ def make_layers(cfg: List[Union[str, int]], batch_norm: bool = False) -> nn.Sequ
 }
 
 
-def _vgg(arch: str, cfg: str, batch_norm: bool, pretrained: bool, progress: bool, **kwargs: Any) -> VGG:
-    if pretrained:
+def _vgg(cfg: str, batch_norm: bool, weights: Optional[WeightsEnum], progress: bool, **kwargs: Any) -> VGG:
+    if weights is not None:
         kwargs["init_weights"] = False
```
datumbox, line 100: Proof that the testing strategy works. Prior commit 903a3d5, this optimization was missing from the PR and thus the VGG was being initialized prior loading the weights on the feature branch. Due to this, the set seed was being used for the initialization of the network and thus the input on the pretrained model was different. The issue was caught by the tests that use pretrained expected files and corrected.

torchvision/models/detection/retinanet.py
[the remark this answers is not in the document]
oke-aditya: I get it. I guess the best solution is what you are trying in #5444 . If we are able to separate out `GeneralizedRCNNTransform`. We can write clearer docs as what it does.

A small doubt though what does `size=(224,224)` in classification models then stands for? 
That's the size of image passed to them when the models were trained I think? Since even they are flexible due `AdaptiveAvgPool2d`?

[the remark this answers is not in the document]
datumbox: The `size` field gives you information about the inference size of the model. This is the size of the crop that the model expects. Whether this is fixed size or variable depends on the model; for instance ViT is not flexible.

Again I'm happy to make changes by adding, removing, amending the meta-data info to make them more readable/understandable. Just trying to provide clarifications of what each field means because currently we are lacking documentation (TBD). Let me know thoughts/ideas.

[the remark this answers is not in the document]
oke-aditya: Sounds good. We would need documentation on meta-data info. I will too think over it :)

references/classification/train.py
NicolasHug: Documenting this, as we just discussed offline:

- we want users to be able to use the builtin presets (`weights.transforms()`) for inference
- yet we still want users to be able to modify the presets they're using on the references for their own experiments. So it makes sense to have somewhat duplicated presets, the builtin ones and the ones on the references.
- we don't have a great weight to test the builtin presets right now, since we don't test model training (but we're planning to work on this). So relying on the builtin presets here acts somewhat as a functional test for them. It's also relevant to rely on them in `test_only` mode, because these are the actual presets that we except to be used in production settings.

NicolasHug (commented):
Thanks a lot @datumbox . I made some minor comments but LGTM anyway.

NicolasHug (approved):

torchvision/models/detection/backbone_utils.py
datumbox: I've updated the code to avoid the hack and ensure the warnings are of better quality.

jdsgomes (approved):
LGTM! thank you

datumbox:
@NicolasHug @jdsgomes Thank you both for your heroic reviews and support on this work. I'm running few last checks on the references and if all pass, I'll merge the PR soon.

I really hope that going forwards we will find better ways to land prototype APIs into main without such massive PRs.

oke-aditya:
Vamos!! Multi Weight support. 

( I really wish if GitHub has GIF search like Giphy lol)

PR #5631, opened 2022-03-16
Remove unnecessary PYTORCH_TEST_WITH_SLOW env

datumbox:
The JIT tests always execute on the CI. Making them optional doesn't make sense as JIT-scriptability is an important feature of our models. Having the test off by default also means that new contributors don't check for JIT-scriptability, which can lead to issues.

Hence I propose to remove the flag completely.


jdsgomes (approved):
LGTM

PR #5636, opened 2022-03-17
replace remaining asserts

jdsgomes:
Replace torch._asserts as a follow up to this PR https://github.com/pytorch/vision/pull/5587

Thanks @datumbox for raising it in https://github.com/pytorch/vision/issues/5494

### DIFF
--- a/torchvision/models/vision_transformer.py
+++ b/torchvision/models/vision_transformer.py
@@ -241,10 +241,8 @@ class VisionTransformer(nn.Module):
     def _process_input(self, x: torch.Tensor) -> torch.Tensor:
         n, c, h, w = x.shape
         p = self.patch_size
-        if h != self.image_size:
-            raise ValueError(f"Wrong image height! Expected {self.image_size} but got {h}")
-        if w != self.image_size:
-            raise ValueError(f"Wrong image width! Expected {self.image_size} but got {w}")
+        torch._assert(h == self.image_size, "Wrong image height!")
+        torch._assert(w == self.image_size, "Wrong image width!")
         n_h = h // p
         n_w = w // p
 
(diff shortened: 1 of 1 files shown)


NicolasHug (approved):
Thanks @jdsgomes

jdsgomes:
tests are failing because of this change: https://github.com/pytorch/vision/pull/5636/files#diff-678e419692ae76586d9f58ce62d2841be82d607637b7d507ead76eca412fcdd1R244

I am thinking to revert and leave this particular `torch._assert` because its traceable. 
@NicolasHug do you know if I have other options here?

NicolasHug:
Sorry, I thought I had replied to this comment but it looks like I didn't :man_facepalming: 

I don't know if there is a good workaround for this. But I'm surprised that we did not encouter the same error for the changes we did in https://github.com/pytorch/vision/pull/5587. Do you know why that might be the case?

datumbox:
I had a very quick look on https://github.com/pytorch/vision/pull/5587 and I think it works because all of our Exceptions are either on utils/methods that are not part of the models or on model constructors. I agree with jdgomes that we should probably close the PR and leave the assertions as is.

NicolasHug:
Looks like even in https://github.com/pytorch/vision/pull/5587 there were a bunch of changes made to the forward pass of e.g. [RAFT](https://github.com/pytorch/vision/pull/5587/files#diff-695b16043168ca4717ef0ac674e2c78c3055e752616b0366a0b2608a40375db8R397), but it looks like we don't test RAFT against fx

jdsgomes:
Abandoning this PR and will follow up with another PR reverting some of the recent exceptions added after the offline discussion with @datumbox

PR #5637, opened 2022-03-17
Default tracer args

jdsgomes:
Always sets default tracer kwargs

adresses https://github.com/pytorch/vision/issues/5508

cc @alexander-soare @datumbox


torchvision/models/feature_extraction.py
```diff
@@ -184,6 +184,23 @@ def _get_leaf_modules_for_ops() -> List[type]:
     return result
 
 
+def set_default_tracer_kwargs(original_tr_kwargs: Optional[Dict[str, Any]]) -> Dict[str, Any]:
```
datumbox, line 187: Is this on purpose public? If it is it should have a documentation. If not then we should prepend an `_` to its name.

torchvision/models/feature_extraction.py
```diff
@@ -184,6 +184,23 @@ def _get_leaf_modules_for_ops() -> List[type]:
...
+    result_tracer_kwargs["autowrap_modules"] = (
+        tuple(set(result_tracer_kwargs["autowrap_modules"] + default_autowrap_modules))
+        if "autowrap_modules" in result_tracer_kwargs
+        else default_autowrap_modules
+    )
+    result_tracer_kwargs["leaf_modules"] = (
+        list(set(result_tracer_kwargs["leaf_modules"] + default_leaf_modules))
+        if "leaf_modules" in result_tracer_kwargs
+        else default_leaf_modules
+    )
```
datumbox, line 200: Nit: You could simplify this by starting with the initial values and then appending the user modules if available. No strong opinions.

datumbox (commented):
Thanks @jdsgomes, it looks good overall. Just one comment concerning the visibility of the method:

torchvision/models/feature_extraction.py
jdsgomes: it should be private, thanks for pointing out

torchvision/models/feature_extraction.py
jdsgomes: I started that way but somehow I found this way better in the end.

datumbox (approved):
LGTM, thanks!

PR #5640, opened 2022-03-18
port image type conversion transforms to prototype API

pmeier:
This is a 1-to-1 port of the old transformations. This should serve as discussion basis.

### DIFF
--- a/torchvision/prototype/transforms/_type_conversion.py
+++ b/torchvision/prototype/transforms/_type_conversion.py
@@ -1,7 +1,11 @@
-from typing import Any, Dict
+from typing import Any, Dict, Optional
 
+import numpy as np
+import PIL.Image
 from torchvision.prototype import features
 from torchvision.prototype.transforms import Transform, functional as F
+
+from ._utils import is_simple_tensor
 
 
 class DecodeImage(Transform):
@@ -33,3 +37,31 @@ class LabelToOneHot(Transform):
             return ""
 
         return f"num_categories={self.num_categories}"
+
+
+class ToTensor(Transform):
+    def _transform(self, input: Any, params: Dict[str, Any]) -> Any:
+        if isinstance(input, (PIL.Image.Image, np.ndarray)):
+            return F.to_tensor(input)
+        else:
+            return input
+
+
+class PILToTensor(Transform):
+    def _transform(self, input: Any, params: Dict[str, Any]) -> Any:
+        if isinstance(input, PIL.Image.Image):
+            return F.pil_to_tensor(input)
+        else:
+            return input
+
+
+class ImageToPIL(Transform):
+    def __init__(self, mode: Optional[str] = None) -> None:
+        super().__init__()
+        self.mode = mode
+
+    def _transform(self, input: Any, params: Dict[str, Any]) -> Any:
+        if is_simple_tensor(input) or isinstance(input, (features.Image, np.ndarray)):
+            return F.image_to_pil(input, mode=self.mode)
+        else:
+            return input
--- a/torchvision/prototype/transforms/functional/_type_conversion.py
+++ b/torchvision/prototype/transforms/functional/_type_conversion.py
@@ -7,6 +7,7 @@ import torch
 from torch.nn.functional import one_hot
 from torchvision.io.video import read_video
 from torchvision.prototype.utils._internal import ReadOnlyTensorBuffer
+from torchvision.transforms import functional as F
 
 
 def decode_image_with_pil(encoded_image: torch.Tensor) -> torch.Tensor:
@@ -23,3 +24,8 @@ def decode_video_with_av(encoded_video: torch.Tensor) -> Tuple[torch.Tensor, tor
 
 def label_to_one_hot(label: torch.Tensor, *, num_categories: int) -> torch.Tensor:
     return one_hot(label, num_classes=num_categories)  # type: ignore[no-any-return]
+
+
+to_tensor = F.to_tensor
+pil_to_tensor = F.pil_to_tensor
+image_to_pil = F.to_pil_image
--- a/torchvision/prototype/transforms/__init__.py
+++ b/torchvision/prototype/transforms/__init__.py
@@ -28,4 +28,4 @@ from ._presets import (
     VideoClassificationEval,
     OpticalFlowEval,
 )
-from ._type_conversion import DecodeImage, LabelToOneHot
+from ._type_conversion import DecodeImage, LabelToOneHot, ToTensor, ImageToPIL, PILToTensor
--- a/torchvision/prototype/transforms/functional/__init__.py
+++ b/torchvision/prototype/transforms/functional/__init__.py
@@ -71,4 +71,11 @@ from ._geometry import (
     ten_crop_image_pil,
 )
 from ._misc import normalize_image_tensor, gaussian_blur_image_tensor
-from ._type_conversion import decode_image_with_pil, decode_video_with_av, label_to_one_hot
+from ._type_conversion import (
+    decode_image_with_pil,
+    decode_video_with_av,
+    label_to_one_hot,
+    to_tensor,
+    pil_to_tensor,
+    image_to_pil,
+)


torchvision/prototype/transforms/_type_conversion.py
```diff
@@ -33,3 +37,31 @@ def extra_repr(self) -> str:
             return ""
 
         return f"num_categories={self.num_categories}"
+
+
+class ToTensor(Transform):
```
pmeier, line 42: The only benefit of using `ToTensor` over `PILToTensor` is that the former also handles `numpy` images, i.e. arrays of shape `H x W x C`. So far, we haven't talked about such images. Do we want to keep supporting them?

torchvision/prototype/transforms/_type_conversion.py
```diff
@@ -33,3 +37,31 @@ def extra_repr(self) -> str:
...
+        if isinstance(input, (PIL.Image.Image, np.ndarray)):
+            return F.to_tensor(input)
+        else:
+            return input
+
+
+class PILToTensor(Transform):
+    def _transform(self, input: Any, params: Dict[str, Any]) -> Any:
+        if isinstance(input, PIL.Image.Image):
+            return F.pil_to_tensor(input)
```
pmeier, line 53: Should we convert into a `features.Image` here to "promote" the new API?

```suggestion
            output = F.pil_to_tensor(input)
            return features.Image(output, color_space=features.ColorSpace.from_pil_mode(input.mode))
```

For the user nothing changes, since a `features.Image` behaves just like a regular tensor.

pmeier (commented):
The overarching question is: do we want to have a `ConvertImageType` transform that takes all kind of image types and converts them to a fixed format? That could unify all three of the transforms below. As output formats, I would only allow tensor and PIL images for now.

torchvision/prototype/transforms/functional/_type_conversion.py
```diff
@@ -7,6 +7,7 @@
 from torch.nn.functional import one_hot
 from torchvision.io.video import read_video
 from torchvision.prototype.utils._internal import ReadOnlyTensorBuffer
+from torchvision.transforms import functional as F
```
pmeier, line 10:
```suggestion
from torchvision.transforms import functional as _F
```

torchvision/prototype/transforms/_type_conversion.py
vfdev-5: ToTensor is going to be deprecated

torchvision/prototype/transforms/_type_conversion.py
```diff
@@ -33,3 +37,31 @@ def extra_repr(self) -> str:
...
+
+class ToTensor(Transform):
+    def _transform(self, input: Any, params: Dict[str, Any]) -> Any:
+        if isinstance(input, (PIL.Image.Image, np.ndarray)):
+            return F.to_tensor(input)
+        else:
+            return input
+
+
+class PILToTensor(Transform):
```
vfdev-5, line 50: `PILToTensor`, `ImageToPIL` names are unclear.

`PILToTensor` returns `features.Image` and `Image` in `ImageToPIL` make think of `features.Image` but apparently it is more like `ToPIL`

torchvision/prototype/transforms/_type_conversion.py
pmeier: > ToTensor is going to be deprecated.

Is that the case? So far I'm only aware of discussions that we might want to do this, but not definitive plan. If we go for a deprecation, doing it in the course of adapting the new API is good time though.

> About transforming numpy to tensor, we may create `NupmyToTensor` ?

Do we really want to have a separate transform for each input type? IMO the largest appeal of `ToTensor` for users is that they don't have to care about the input type and can use it in every scenario. My proposal would be to have only two transforms:

- `ToImageTensor`
- `ToImagePIL`

Both transforms take "arbitrary" inputs and convert them to either a `features.Image` or a `PIL.Image.Image`. The naming is closely aligned with our new kernels and they give the same convenience as before.

torchvision/prototype/transforms/_type_conversion.py
pmeier: Not sure what you are saying here. Do you want to change this in the future? Because the current implementation in this PR is a 1-to-1 port of what we have right now in the stable API. This is not what I'm suggesting.

vfdev-5: I'm saying that current names are unclear. `ToImageTensor` and `ToImagePIL` sounds much better.

torchvision/prototype/transforms/_deprecated.py
```diff
@@ -0,0 +1,40 @@
...
+import numpy as np
+import PIL.Image
+from torchvision.prototype import features
+from torchvision.prototype.transforms import Transform
+from torchvision.transforms import functional as _F
+
+from ._utils import is_simple_tensor
+
+
+# TODO: add deprecation warning
```
datumbox, line 12: Is this going to happen later?

torchvision/prototype/transforms/functional/_type_conversion.py
```diff
@@ -23,3 +24,23 @@ def decode_video_with_av(encoded_video: torch.Tensor) -> Tuple[torch.Tensor, tor
...
+
+
+def to_image_pil(image: Union[torch.Tensor, PIL.Image.Image, np.ndarray], copy: bool = False) -> PIL.Image.Image:
+    if isinstance(image, PIL.Image.Image):
+        if copy:
+            return image.copy()
+        else:
+            return image
+
+    return _F.to_pil_image(to_image_tensor(image, copy=False))
```
datumbox, line 46: I am uncertain if it makes sense to do:
```suggestion
    return _F.to_pil_image(to_image_tensor(image, copy=copy))
```

Thoughts?

torchvision/prototype/transforms/_type_conversion.py
```diff
@@ -33,3 +37,28 @@ def extra_repr(self) -> str:
...
+
+class ToImageTensor(Transform):
+    def __init__(self, *, copy: bool = False) -> None:
+        super().__init__()
+        self.copy = copy
+
+    def _transform(self, input: Any, params: Dict[str, Any]) -> Any:
+        if isinstance(input, (features.Image, PIL.Image.Image, np.ndarray)) or is_simple_tensor(input):
+            output = F.to_image_tensor(input, copy=self.copy)
+            return features.Image(output)
```
datumbox, line 50: When I first read `to_image_tensor()` I expected you would be returning `features.Image` not `Tensor` but then remembered that all other `*_image_tensor()` methods return Tensors for JIT-scriptability...

But then isn't it weird that `ToImageTensor` return a `features.Image`? I'm not actually asking you to change it, just trying to figure out how we want this to work. 

Could you elaborate when people would use this transform? The way I understand it is that the datasets will return I `feature.Image` entries, so the use of this transform will be limited to only when someone has read a PIL image manually and want to convert them to Image. Right?

datumbox (commented):
Overall looks good. Just a few questions:

torchvision/prototype/transforms/_type_conversion.py
pmeier: > But then isn't it weird that `ToImageTensor` return a `features.Image`? I'm not actually asking you to change it, just trying to figure out how we want this to work.

Maybe a little unexpected, but I wouldn't say weird. Given that `features.Image` is a superset of what `torch.Tensor` provides, users should not feel a difference. No strong opinion though since all other transforms accept both types. Maybe it is a nice way to "promote" the new API.

> Could you elaborate when people would use this transform? The way I understand it is that the datasets will return I `feature.Image` entries, so the use of this transform will be limited to only when someone has read a PIL image manually and want to convert them to Image. Right?

Correct. Neither `ToImageTensor` nor `ToImagePIL` will be needed if users stay in "`torchvision` land". Depending on their own code, they might use these though.

torchvision/prototype/transforms/_deprecated.py
datumbox: Let's add the warnings to avoid having unfinished TODOs and we can always revise. This is a proposal after all, not necessarily the end result.

torchvision/prototype/transforms/_type_conversion.py
datumbox: Sounds good. I'm OK adopting this approach for now. If you fix the conflicts, breakages and remaining TODOs we should be able to merge.

datumbox (approved):
LGTM, thanks!

PR #5641, opened 2022-03-18
expand legacy model builder wrapper for default weights_backbone behavior

pmeier:
Diff looks larger as it is. I refactored the loop into a local function and call it twice explicitly instead of looping over generic inputs. We should probably update the warnings to include a specific deprecation timeline.


datumbox:
@pmeier Thanks for the PR. I've discussed with the team and adopting this would probably require more discussions across the domain libraries. To avoid delays we will revert the previous default values and go fully-BC and we can re-examine this approach on the future.

PR #5643, opened 2022-03-18
fix mobilnet norm layer test

### LINKED ISSUE #5642: `test/test_models::test_mobile_net_norm_layer` is broken
https://github.com/pytorch/vision/blob/297e2b8746d32b46d20b0e6a6f867fde65e50e8d/test/test_models.py#L403-L413

[fails on `main`](https://app.circleci.com/pipelines/github/pytorch/vision/15869/workflows/f81e239f-f520-45b5-b663-8d71a1aa90d3/jobs/1283688) due to pytorch/pytorch#74293.

cc @datumbox @pmeier

pmeier:
Fixes #5642. IIUC, the test only check if one can pass a custom `norm_layer` and it will actually embedded into the model. Thus, the parametrization of the layer should have no effect.

### DIFF
--- a/test/test_models.py
+++ b/test/test_models.py
@@ -400,14 +400,13 @@ def test_mobilenet_v2_residual_setting():
     assert out.shape[-1] == 1000
 
 
-@pytest.mark.xfail(reason="See https://github.com/pytorch/vision/issues/5642")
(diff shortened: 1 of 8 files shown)


datumbox (approved):
LGTM, thanks!

Issue #5616, opened 2022-03-14
Torchvision doesn't compile against ffmpeg-5


xuzhao9:
### 🐛 Describe the bug

Steps to reproduce:
1. Install torchvision nightly from conda: `conda install -y -c pytorch-nightly torchvision`. This will also install [ffmpeg-5](https://anaconda.org/conda-forge/ffmpeg) from conda
2. Compile torchvision from source: `python setup.py install`
3. The compile fails with the following error:
```
cc1plus: warning: command line option ‘-Wstrict-prototypes’ is valid for C/ObjC but not for C++
/home/ec2-user/nvme/srcs/vision/torchvision/csrc/io/decoder/decoder.cpp:21:36: error: use of enum ‘AVLockOp’ without previous declaration
 int ffmpeg_lock(void** mutex, enum AVLockOp op) {
                                    ^~~~~~~~
/home/ec2-user/nvme/srcs/vision/torchvision/csrc/io/decoder/decoder.cpp: In function ‘int ffmpeg::{anonymous}::ffmpeg_lock(void**, int)’:
/home/ec2-user/nvme/srcs/vision/torchvision/csrc/io/decoder/decoder.cpp:24:10: error: ‘AV_LOCK_CREATE’ was not declared in this scope
     case AV_LOCK_CREATE:
          ^~~~~~~~~~~~~~
/home/ec2-user/nvme/srcs/vision/torchvision/csrc/io/decoder/decoder.cpp:24:10: note: suggested alternative: ‘AV_LOG_TRACE’
     case AV_LOCK_CREATE:
          ^~~~~~~~~~~~~~
          AV_LOG_TRACE
/home/ec2-user/nvme/srcs/vision/torchvision/csrc/io/decoder/decoder.cpp:27:10: error: ‘AV_LOCK_OBTAIN’ was not declared in this scope
     case AV_LOCK_OBTAIN:
          ^~~~~~~~~~~~~~
/home/ec2-user/nvme/srcs/vision/torchvision/csrc/io/decoder/decoder.cpp:27:10: note: suggested alternative: ‘CLOCK_TAI’
     case AV_LOCK_OBTAIN:
          ^~~~~~~~~~~~~~
          CLOCK_TAI
/home/ec2-user/nvme/srcs/vision/torchvision/csrc/io/decoder/decoder.cpp:30:10: error: ‘AV_LOCK_RELEASE’ was not declared in this scope
     case AV_LOCK_RELEASE:
[...]

PR #5644, opened 2022-03-18
Attempt to fix FFMPEG 5.0 compatibility

bjuncek:
Attempting to address #5616 via simple API update.
Should be properly tested in FB internals.  

cc @datumbox


torchvision/csrc/io/decoder/subtitle_stream.cpp
```diff
@@ -44,20 +44,26 @@ int SubtitleStream::analyzePacket(const AVPacket* packet, bool* gotFrame) {
...
-  avPacket.data = nullptr;
-  avPacket.size = 0;
-  auto pkt = packet ? *packet : avPacket;
+  AVPacket* avPacket;
+  avPacket = av_packet_alloc();
+  if (avPacket == NULL) {
+    LOG(ERROR) << "decoder as not able to allocate the packet.";
+  }
+  avPacket->data = nullptr;
+  avPacket->size = 0;
```
datumbox, line 53: Nit: This code is repeated on the decoder.cpp file. Would it make sense to move the initialisation to a separate internal utility method?

datumbox (commented):
@bjuncek Thanks for fixing this. The changes look good to me. I've only added one comment for your attention.

Our CI is currently broken, so I propose not to merge this until #5648 is merged and it's properly tested on FB infra as you suggested. We'll get back to you on the latter.

torchvision/csrc/io/decoder/decoder.cpp
```diff
@@ -505,10 +484,14 @@ int Decoder::getFrame(size_t workingTimeInMs) {
   // once decode() method gets called and grab some bytes
   // run this method again
   // init package
-  AVPacket avPacket;
-  av_init_packet(&avPacket);
-  avPacket.data = nullptr;
-  avPacket.size = 0;
+  // update 03/22: moving memory management to ffmpeg
+  AVPacket* avPacket;
+  avPacket = av_packet_alloc();
```
jdsgomes, line 489: according to LeakSanitizer this call will result in a memory leak.

jdsgomes (commented):
I have imported the PR and pointed out a few things to be considered.

torchvision/csrc/io/decoder/decoder.cpp
datumbox: We should call `av_packet_free(&avPacket);` when we are done with it. We need to be careful and add this to all the places where we return early too.

jdsgomes:
@bjuncek I made a few changes in an attempt to fix the memory leak and the static code analysis warnings, but there are a few things that came up that would need discussion when you are back as you will probably have more context:

- if you agree with the approach [here](t[orchvision/csrc/io/decoder/subtitle_stream.cpp](https://github.com/pytorch/vision/pull/5644/files#diff-7733b1c80a95250516032ddd798d930100e5348ecdb6db558cc789acf11ad773R591)) and [[torchvision/csrc/io/decoder/subtitle_stream.cpp](https://github.com/pytorch/vision/pull/5644/files#diff-7733b1c80a95250516032ddd798d930100e5348ecdb6db558cc789acf11ad773R493)](url), we probably need to follow the same in `torchvision/csrc/io/decoder/subtitle_stream.cpp`
- it is not 100% clear to me what is the difference between `av_packet_unref` and `av_packet_free`. From what I can gather  in the documentation, the former unreferences the inner buffer and resets the default values and the later frees the actual avPacket object but please confirm that is the case and that we need both in the way it has been used

torchvision/csrc/io/decoder/decoder.cpp
bjuncek: Yeah, I was thinking about this, but my thought was that `av_packet_unref` would have freed whatever was allocated (which seems to have been an incorrect assumption). 

I'll fix this ASAP

bjuncek:
>> if you agree with the approach [here](https://github.com/pytorch/vision/pull/5644/files#diff-7733b1c80a95250516032ddd798d930100e5348ecdb6db558cc789acf11ad773R493) and [here](https://github.com/pytorch/vision/pull/5644/files#diff-7733b1c80a95250516032ddd798d930100e5348ecdb6db558cc789acf11ad773R502), we probably need to follow the same in torchvision/csrc/io/decoder/subtitle_stream.cpp

Yup, have added the changes; this is something you might need checking in FB infra as subtitle stream was (yet) not used in TV, but might have been used internally (I've been long gone). 

>> it is not 100% clear to me what is the difference between av_packet_unref and av_packet_free. From what I can gather in the documentation, the former unreferences the inner buffer and resets the default values and the later frees the actual avPacket object but please confirm that is the case and that we need both in the way it has been used

My understanding in this case is the same as yours, and I do thing we'd need both; do these changes pass the static memory leak checks?

torchvision/csrc/io/decoder/decoder.cpp
```diff
@@ -585,10 +572,10 @@ int Decoder::getFrame(size_t workingTimeInMs) {
 
     result = 0;
 
-    av_packet_unref(&avPacket);
+    av_packet_unref(avPacket);
   }
 
-  av_packet_unref(&avPacket);
+  av_packet_unref(avPacket);
```
datumbox, line 578: Two points on this:
1. This second unref was originally here because the loop might terminate via a break. The code is messy and we should refactor it to make it easier to read and have better flow. That's a future NIT.
2. Given that now we call `av_packet_free` call few lines below, should we drop the unref call?

torchvision/csrc/io/decoder/subtitle_stream.cpp
```diff
@@ -44,20 +44,28 @@ int SubtitleStream::analyzePacket(const AVPacket* packet, bool* gotFrame) {
...
   // check flush packet
-  AVPacket avPacket;
-  av_init_packet(&avPacket);
-  avPacket.data = nullptr;
-  avPacket.size = 0;
-  auto pkt = packet ? *packet : avPacket;
+  AVPacket* avPacket;
+  avPacket = av_packet_alloc();
+  if (avPacket == NULL) {
+    LOG(ERROR) << "decoder as not able to allocate the packet.";
```
datumbox, line 50: See Joao's comment on the `nullptr`. Also, should we be returning an error if allocating a packet fails? Finally it's unclear to me why we allocate a packet even when `packet` is not null. This can be improved on a future PR.

torchvision/csrc/io/decoder/decoder.cpp
bjuncek: 1. I agree. IMO the entire API needs a refactor, ideally to be merged with the GPU decoding; This approach  is overly complicated (for historical reasons) for our purposes.
2. Yes if the documentation is to be believed; there are two versions of free packet: `av_packet_free` (which does un-referencing for us), and `av_free_packet` (that doesn't)

torchvision/csrc/io/decoder/subtitle_stream.cpp
bjuncek: small enough change that I can push it now I take it, and revert if it breaks something profoundly

torchvision/csrc/io/decoder/subtitle_stream.cpp
```diff
@@ -43,21 +43,35 @@ int SubtitleStream::initFormat() {
...
+
+  auto pkt = packet;
+  if (pkt == nullptr) {
+    // check flush packet
+    AVPacket* avPacket;
+    avPacket = av_packet_alloc();
+    if (avPacket == nullptr) {
+      LOG(ERROR)
+          << "decoder as not able to allocate the subtitle-specific packet.";
+      return ENOMEM;
```
datumbox, line 55: Are we sure that this error code will work as return type for the method? Should we be returning one of the error codes of avcodec instead?

torchvision/csrc/io/decoder/subtitle_stream.cpp
```diff
@@ -43,21 +43,35 @@ int SubtitleStream::initFormat() {
...
-  int result = avcodec_decode_subtitle2(codecCtx_, &sub_, &gotFramePtr, &pkt);
+  // is these a better way than cast from const?
+  int result =
+      avcodec_decode_subtitle2(codecCtx_, &sub_, &gotFramePtr, (AVPacket*)pkt);
 
   if (result < 0) {
     LOG(ERROR) << "avcodec_decode_subtitle2 failed, err: "
                << Util::generateErrorDesc(result);
+    // free packet
+    av_packet_free(&avPacket);
```
datumbox, line 71: Why is this variable in scope?

torchvision/csrc/io/decoder/subtitle_stream.cpp
bjuncek: I've updated them to the most appropriate [AVERROR](https://ffmpeg.org/doxygen/trunk/group__lavu__error.html)


Having done that, I'd say ENOMEM is more appropriate than AV alternative that I see used (`AVERROR_BUFFER_TOO_SMALL`).

torchvision/csrc/io/decoder/subtitle_stream.cpp
```diff
@@ -66,7 +79,7 @@ int SubtitleStream::analyzePacket(const AVPacket* packet, bool* gotFrame) {
   // set proper pts in us
   if (gotFramePtr) {
     sub_.pts = av_rescale_q(
-        pkt.pts, inputCtx_->streams[format_.stream]->time_base, timeBaseQ);
+        pkt->pts, inputCtx_->streams[format_.stream]->time_base, timeBaseQ);
   }
 
   return result;
```
datumbox, line 85: We need to clean up `av_packet_free(&avPacket);` prior returning here.

torchvision/csrc/io/decoder/subtitle_stream.cpp
```diff
@@ -43,21 +43,34 @@ int SubtitleStream::initFormat() {
...
   releaseSubtitle();
+
+  // FIXME: should this even be created?
+  AVPacket* avPacket;
+  avPacket = av_packet_alloc();
+  if (avPacket == nullptr) {
+    LOG(ERROR)
+        << "decoder as not able to allocate the subtitle-specific packet.";
+    // alternative to ENOMEM
+    return AVERROR_BUFFER_TOO_SMALL;
```
datumbox, line 54: This is close enough for error code I think.

torchvision/csrc/io/decoder/decoder.cpp
```diff
@@ -505,23 +485,29 @@ int Decoder::getFrame(size_t workingTimeInMs) {
...
-  AVPacket avPacket;
-  av_init_packet(&avPacket);
-  avPacket.data = nullptr;
-  avPacket.size = 0;
+  // update 03/22: moving memory management to ffmpeg
+  AVPacket* avPacket;
+  avPacket = av_packet_alloc();
+  if (avPacket == nullptr) {
+    LOG(ERROR) << "decoder as not able to allocate the packet.";
+    return AVERROR_BUFFER_TOO_SMALL;
```
datumbox, line 493: In this method, this is not a supported error code (at least according to its doc). We should either update the doc to document this error code or adopt a more fitting enum for this.

torchvision/csrc/io/decoder/decoder.cpp
```diff
@@ -505,23 +485,29 @@ int Decoder::getFrame(size_t workingTimeInMs) {
...
+  }
+  avPacket->data = nullptr;
+  avPacket->size = 0;
 
   auto end = std::chrono::steady_clock::now() +
       std::chrono::milliseconds(workingTimeInMs);
   // return true if elapsed time less than timeout
   auto watcher = [end]() -> bool {
+    av_packet_free(&avPacket);
```
datumbox, line 502: Are you sure we should be freeing the memory here? I think this callable is used in the loop below, it doesn't actually return for the whole method.

torchvision/csrc/io/decoder/decoder.cpp
bjuncek: is this better?

datumbox (commented):
Thanks for your work @bjuncek. I think it looks good.

@jdsgomes Could you check if this works OK internally?

torchvision/csrc/io/decoder/subtitle_stream.cpp
```diff
@@ -43,21 +43,34 @@ int SubtitleStream::initFormat() {
 int SubtitleStream::analyzePacket(const AVPacket* packet, bool* gotFrame) {
   // clean-up
   releaseSubtitle();
+
+  // FIXME: should this even be created?
```
jdsgomes, line 47: agree - I am also not sure if this should be created. Is the plan to follow up on this?

torchvision/csrc/io/decoder/subtitle_stream.cpp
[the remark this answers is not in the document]
datumbox: @jdsgomes The two methods return different enums. Originally Bruno had them synced; no strong opinions on my side.

torchvision/csrc/io/decoder/subtitle_stream.cpp
bjuncek: Yes, but I think this would require a bit of coordination on Meta's side, as this is not used nor tested in TV.

torchvision/csrc/io/decoder/subtitle_stream.cpp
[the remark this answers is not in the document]
bjuncek: @jdsgomes would you like me to change it to match?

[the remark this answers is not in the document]
jdsgomes: I would prefer for consistency, unless there is a reason not to. Other than that the PR looks good and the internal tests are passing. Thank you for these changes

jdsgomes:
For reference the internal tests are passing: https://www.internalfb.com/diff/D35115411

bjuncek:
@datumbox I'll leave it up to you to decide if it's ready for merge.

jdsgomes (approved):

jdsgomes:
I have a nit regarding he return codes but definitely non blocking.

datumbox (approved):
@bjuncek LGTM, thanks!

@jdsgomes I agree that a good API should align the return error codes across functions. I'm OK if you want to merge this now and follow up in a new PR. I'm also OK if you want to push the change directly on this branch. I leave it to you how we should proceed. At any case, we should merge this soon. We are already a week with broken CI and the issue is masking other potential problems that are not reported to the release engineering team.

torchvision/csrc/io/decoder/decoder.cpp
```diff
@@ -505,10 +485,15 @@ int Decoder::getFrame(size_t workingTimeInMs) {
...
-  AVPacket avPacket;
-  av_init_packet(&avPacket);
-  avPacket.data = nullptr;
-  avPacket.size = 0;
+  // update 03/22: moving memory management to ffmpeg
+  AVPacket* avPacket;
+  avPacket = av_packet_alloc();
+  if (avPacket == nullptr) {
+    LOG(ERROR) << "decoder as not able to allocate the packet.";
+    return AVERROR(ENOMEM);
```
jdsgomes, line 493:
```suggestion
    return AVERROR_BUFFER_TOO_SMALL;
```

torchvision/csrc/io/decoder/decoder.cpp
```diff
@@ -1,6 +1,7 @@
 #include "decoder.h"
 #include <c10/util/Logging.h>
 #include <libavutil/avutil.h>
+#include <cerrno>
```
jdsgomes, line 4:
```suggestion
```

PR #5647, opened 2022-03-19
USPS dataset

### LINKED ISSUE #5352: USPS
cc @pmeier @bjuncek

lezwon:
Add USPS dataset to torchvision. Fixes #5352


torchvision/prototype/datasets/_builtin/usps.py
```diff
@@ -0,0 +1,72 @@
...
+        ),
+        "test": HttpResource(
+            f"{_URL}/usps.t.bz2", sha256="a9c0164e797d60142a50604917f0baa604f326e9a689698763793fa5d12ffc4e"
+        ),
+    }
+
+    def resources(self, config: DatasetConfig) -> List[OnlineResource]:
+        return [USPS._RESOURCES[config.split]]
+
+    def _prepare_sample(self, data: Tuple[torch.Tensor, torch.Tensor], *, config: DatasetConfig) -> Dict[str, Any]:
```
pmeier, line 56: I don't think we need the `config` here, correct?

```suggestion
    def _prepare_sample(self, data: Tuple[torch.Tensor, torch.Tensor]) -> Dict[str, Any]:
```

torchvision/prototype/datasets/_builtin/usps.py
```diff
@@ -0,0 +1,72 @@
...
+    def _make_datapipe(
+        self,
+        resource_dps: List[IterDataPipe],
+        *,
+        config: DatasetConfig,
+    ) -> IterDataPipe[Dict[str, Any]]:
+        dp = USPSFileReader(resource_dps[0])
+        dp = hint_sharding(dp)
+        dp = hint_shuffling(dp)
+        return Mapper(dp, functools.partial(self._prepare_sample, config=config))
```
pmeier, line 72:
```suggestion
        return Mapper(dp, self._prepare_sample)
```

torchvision/prototype/datasets/_builtin/usps.py
```diff
@@ -0,0 +1,72 @@
...
+
+import numpy as np
+import torch
+from torchdata.datapipes.iter import IterDataPipe, IterableWrapper, LineReader, Mapper
+from torchvision.prototype.datasets.utils import Dataset, DatasetInfo, DatasetConfig, OnlineResource, HttpResource
+from torchvision.prototype.datasets.utils._internal import hint_sharding, hint_shuffling
+from torchvision.prototype.features import Image, Label
+
+
+class USPSFileReader(IterDataPipe[torch.Tensor]):
```
pmeier, line 13: Could you refactor this now that your PR (thank you!) pytorch/data#312 was merged? My guess is that all the processing you do to the `line` should go into `_prepare_sample`.

pmeier (changes requested):
Thanks a lot for the PR @lezwon! I have some comments inline. When they are addressed, we also need to add tests. I'll add another section to the README to guide you.

torchvision/prototype/datasets/_builtin/usps.py
```diff
@@ -0,0 +1,61 @@
...
+        )
+
+    def _make_datapipe(
+        self,
+        resource_dps: List[IterDataPipe],
+        *,
+        config: DatasetConfig,
+    ) -> IterDataPipe[Dict[str, Any]]:
+        dp = Decompressor(resource_dps[0])
+        dp = LineReader(dp, decode=True)
```
pmeier, line 58: If we don't need the path, we can simply disable it.

```suggestion
        dp = LineReader(dp, decode=True, return_path=False)
```

torchvision/prototype/datasets/_builtin/usps.py
```diff
@@ -0,0 +1,61 @@
...
+        "test": HttpResource(
+            f"{_URL}/usps.t.bz2", sha256="a9c0164e797d60142a50604917f0baa604f326e9a689698763793fa5d12ffc4e"
+        ),
+    }
+
+    def resources(self, config: DatasetConfig) -> List[OnlineResource]:
+        return [USPS._RESOURCES[config.split]]
+
+    def _prepare_sample(self, data: Tuple[torch.Tensor, torch.Tensor]) -> Dict[str, Any]:
+        _filename, line = data
```
pmeier, line 37: With the suggestion above, this simplifies to

```suggestion
    def _prepare_sample(self, line: str) -> Dict[str, Any]:
```

torchvision/prototype/datasets/_builtin/usps.py
```diff
@@ -0,0 +1,61 @@
...
+        tmp_list = [x.split(":")[-1] for x in raw_data[1:]]
+        img = np.asarray(tmp_list, dtype=np.float32).reshape((-1, 16, 16))
+        img = ((img + 1) / 2 * 255).astype(dtype=np.uint8)
+        img = torch.from_numpy(img)
+        target = int(raw_data[0]) - 1
+
+        return dict(
+            image=Image(img),
+            label=Label(target, dtype=torch.int64, categories=self.categories),
+        )
```
pmeier, line 49: This is doing the right thing, but is a little convoluted:

1. Why do we construct a `numpy` array first only to later convert it into a tensor?
2. Why do we convert the image to be `uint8`?
3. Why do we explicitly set the `dtype` for the label? Is the default `dtype` not sufficient?

```suggestion
        label, *values = line.strip().split(" ")
        values = [float(value.split(":")[1]) for value in values]
        pixels = torch.tensor(values).add_(1).div_(2)
        return dict(
            image=Image(pixels.reshape(16, 16)),
            label=Label(int(label) - 1, categories=self.categories),
        )
```

pmeier (commented):
Thanks @lezwon this is looking pretty good. I've added some suggestions inline. Apart from that, we just merged #5666 that added instructions how to add tests. Please follow them and let me know if something is unclear.

lezwon:
@pmeier I've added the tests. Do let me know if any changes are required. :)

test/test_prototype_builtin_datasets.py
```diff
@@ -180,3 +180,27 @@ def test_label_matches_path(self, test_home, dataset_mock, config):
...
+            assert "image" in sample
+            assert "label" in sample
+
+            # check if correct instance type
+            assert isinstance(sample["image"], Image)
+            assert isinstance(sample["label"], Label)
+
+            # check is data type is correct
+            assert isinstance(sample["image"].data, torch.FloatTensor)
+            assert isinstance(sample["label"].data, torch.LongTensor)
```
pmeier, line 203: `torch.FloatTensor` is only the valid class for [tensors on the CPU](https://pytorch.org/docs/master/tensors.html#data-types). Although this is not the case here, this test would fail if we move the data to GPU. What you actually want to check here is

```suggestion
            # check is data type is correct
            assert sample["image"].dtype is torch.float
            assert sample["label"].dtype is torch.long
```

That being said, I'm not sure why you want to enforce the dtype here? Let's say at some point we decide that all images returned by the datasets should be uint8. This test would fail, but I don't see a reason why the dataset would be invalid with that choice.

test/test_prototype_builtin_datasets.py
```diff
@@ -180,3 +180,27 @@ def test_label_matches_path(self, test_home, dataset_mock, config):
...
+
+@parametrize_dataset_mocks(DATASET_MOCKS["usps"])
+class TestUSPS:
+    def test_label_matches_path(self, test_home, dataset_mock, config):
+        dataset_mock.prepare(test_home, config)
+
+        dataset = datasets.load(dataset_mock.name, **config)
+
+        for sample in dataset:
+            # check if correct keys exist
```
pmeier, line 193: Could you remove these and the following checks? They are just stating what the, objectively simple, code does and thus are redundant. As a rule of thumb, a code comment should state the intent rather than the content.

test/builtin_dataset_mocks.py
```diff
@@ -1431,3 +1432,79 @@ def stanford_cars(info, root, config):
...
+        return " ".join(data)
+
+
+@register_mock
+def usps(info, root, config):
+    num_samples = {"train": 15, "test": 7}[config["split"]]
+
+    train = config.split == "train"
+    data_file = f"usps.{'t.' if not train else ''}bz2"
+    return USPSMockData.generate(root, num_categories=info.categories, num_samples=num_samples, data_file=data_file)
```
pmeier, line 1510: It seems you have copied the data generation from MNIST. In general this is totally fine, but in this case this if kind of an overkill. MNIST mock data generation is quite complex since it handles not only MNIST but also all of its variants. 

I'm guessing no one in the team besides me would have recognized the pattern and thus would be confused on why the mock data is set up this way. I refactored the generation. Let me know if you agree.

```suggestion
@register_mock
def usps(info, root, config):
    num_samples = {"train": 15, "test": 7}[config.split]

    with bz2.open(root / f"usps{'.t' if not config.split == 'train' else ''}.bz2", "wb") as fh:
        lines = []
        for _ in range(num_samples):
            label = make_tensor(1, low=1, high=11, dtype=torch.int)
            values = make_tensor(256, low=-1, high=1, dtype=torch.float)
            lines.append(
                " ".join([f"{int(label)}", *(f"{idx}:{float(value):.6f}" for idx, value in enumerate(values, 1))])
            )

        fh.write("\n".join(lines).encode())

    return num_samples
```

pmeier (commented):
Almost there!

test/test_prototype_builtin_datasets.py
lezwon: Noted. I just assumed that the datatype of data would always be fixed. Will remove this test.

test/builtin_dataset_mocks.py
lezwon: I did copy and make changes from the MNIST generation class. As the dataset is similar I assumed this would be the right way to do it. The shorter version makes sense though :)

pmeier (approved):
LGTM, thanks a lot @lezwon for the continued work on this!

test/test_prototype_builtin_datasets.py
```diff
@@ -180,3 +180,20 @@ def test_label_matches_path(self, test_home, dataset_mock, config):
         for sample in dataset:
             label_from_path = int(Path(sample["path"]).parent.name)
             assert sample["label"] == label_from_path
+
+
+@parametrize_dataset_mocks(DATASET_MOCKS["usps"])
+class TestUSPS:
+    def test_label_matches_path(self, test_home, dataset_mock, config):
```
NicolasHug, line 187: @pmeier what is the goal of this test?
Is `test_label_matches_path` an accurate name? I don't see an obvious path check

pmeier: My bad, I missed the incorrect name. Should be something like

```suggestion
    def test_sample_content(self, test_home, dataset_mock, config):
```

The test checks what #5246 tried to add in a general way. Although USPS would be the only test right now that checks this, IMO it doesn't hurt. If we go through with #5246, we can remove this test.

NicolasHug (approved):
Thanks a lot @lezwon and @pmeier , LGTM . (Ideally we'd change the test name to avoid confusion before merging https://github.com/pytorch/vision/pull/5647/files#r835158538)

Issue #5635, opened 2022-03-17
ImportError: libcupti.so.10.2: cannot open shared object file: No such file or directory


datumbox:
### 🐛 Describe the bug

Multiple `binary_linux_conda_*_cu*` jobs are currently [failing](https://app.circleci.com/pipelines/github/pytorch/vision/15843/workflows/88805119-e3fc-4009-9ff2-5344feb6d36c/jobs/1281404) on the latest main with the following error:
```
Traceback (most recent call last):
  File "setup.py", line 9, in <module>
    import torch
  File "/opt/conda/conda-bld/torchvision_1647515141078/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_p/lib/python3.7/site-packages/torch/__init__.py", line 199, in <module>
    from torch._C import *  # noqa: F403
ImportError: libcupti.so.10.2: cannot open shared object file: No such file or directory
```

Seems like a missing dependency problem. Same applies for `libcupti.so.11.3` [here](https://app.circleci.com/pipelines/github/pytorch/vision/15840/workflows/9a49502e-3f15-454d-b302-cdf86f7d3a4e/jobs/1281138).

### Versions

Latest main branch 39772ec

malfet:
Thank you for reporting. Likely caused by the change that makes cupti a dynamic dependency. 
cc: @ezyang

ezyang:
We can switch back to statically linking cupti as well, I just need to do what malfet suggested when I changed the defaults anyway!

malfet:
@ezyang imo we should just bundle cupti into the conda package

isaak-willett:
Hej @malfet I'm facing the same issue with my build system, error below
```bash
/home/ubuntu/.cache/bazel/_bazel_ubuntu/8de0a1069de8d166c668173ca21c04ae/execroot/com_lyft_avsoftware/bazel-out/k8-fastbuild/bin/src/chronos/tasks/feed_task_cuda_test.runfiles/com_lyft_avsoftware/src/chronos/tasks/feed_task_cuda_test: 

error while loading shared libraries: libcupti.so.11.0: cannot open shared object file: No such file or directory
```
[...]

PR #5648, opened 2022-03-20
Prefer nvidia channel for conda builds

malfet:
Fixes https://github.com/pytorch/vision/issues/5635

### DIFF
--- a/.circleci/unittest/linux/scripts/install.sh
+++ b/.circleci/unittest/linux/scripts/install.sh
@@ -21,7 +21,7 @@ else
     fi
     echo "Using CUDA $CUDA_VERSION as determined by CU_VERSION"
     version="$(python -c "print('.'.join(\"${CUDA_VERSION}\".split('.')[:2]))")"
-    cudatoolkit="cudatoolkit=${version}"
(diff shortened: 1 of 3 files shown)


datumbox (approved):
Thanks @malfet.

Is this a temporary update to fix the issue or does it mean we will change our recommended channel on the download page at pytorch.org?

datumbox:
@malfet The issue persists on the [cmake_linux_gpu](https://app.circleci.com/pipelines/github/pytorch/vision/15891/workflows/2f82b23b-d4f2-4522-9e3d-7d5bea578ea4/jobs/1285532) job:
`ImportError: libcupti.so.11.3: cannot open shared object file: No such file or directory`

The [binary_linux_conda_*_cu115](https://app.circleci.com/pipelines/github/pytorch/vision/15891/workflows/c57eabde-9692-46be-bceb-8bd3634b7ed0/jobs/1285507) jobs are failing because they use ffmpeg-5.0.0 which deprecates some functionality that we currently use. There is another PR that attempts to fix this at #5644.

I think we should merge this after the cmake job is also fixed and then resolve the rest of the jobs on the other PR.

PR #5649, opened 2022-03-20
Adding `detect_flow` and `detect_flows` functions

sugatoray:
Added the following two functions to allow users easily create optical flow and save the data (both as images and as `.pt` data files).

- `detect_flow`: detects optical-flow for a single image-pair
- `detect_flows`: detects optical-flow for a series of image-pairs

The `detect_flow()` function makes sure of managing memory (grabage collection + emptying cuda cache), while evaluating and saving optical flow  information on a pair of images. When I tried running the [RAFT demo](https://pytorch.org/blog/pytorch-1.11-new-library-releases/#2-optical-flow-support-and-raft-model) to create a series of optical-flow images, I found the colab notebook running out of memory. These functions take care of the memory aspect and make it easier for the user to apply optical flow on image-pairs.

### DIFF
--- a/torchvision/models/optical_flow/detector.py
+++ b/torchvision/models/optical_flow/detector.py
@@ -1,0 +1,465 @@
+import os
+import gc
+from typing import Optional, List, Tuple, Union, Callable
+from shutil import copytree, copyfile
+from textwrap import dedent
+
+import numpy as np
+import matplotlib.pyplot as plt
+from tqdm.notebook import tqdm as tqdm_notebook
+from tqdm import tqdm
+
+import torch
+import torchvision
+import torch.nn as nn
+import torchvision.transforms.functional as F
+import torchvision.transforms as T
+from torch import Tensor
+from torchvision.utils import flow_to_image
+from torchvision.io import write_jpeg, write_png
+
+
+def _release_memory(with_no_grad: bool=True):
+    """Releases memory. 
+    
+    Steps:
+        - runs garbage collection: ``gc.collect()``
+        - empties the gpu cache: ``torch.cuda.empty_cache()``
+    """
+    gc.collect()
+    if torch.cuda.is_available():
+        if with_no_grad:
+            with torch.no_grad():
+                torch.cuda.empty_cache()
+        else:
+            torch.cuda.empty_cache()
+
+
+@torch.no_grad()
+def raft_preprocess(batch: Tensor) -> Tensor:
+    """Returns RAFT model preprocessing transformation."""
+    transforms = T.Compose(
+        [
+            T.ConvertImageDtype(torch.float32),
+            T.Normalize(mean=0.5, std=0.5),  # map [0, 1] into [-1, 1]
+            T.Resize(size=(520, 960)),
+        ]
+    )
+    return transforms(batch)
+
+
+@torch.no_grad()
+def _prepare_frame(frame: Union[Tensor, List[Tensor]], preprocess: Optional[Callable]=None) -> Tensor:
+    """Prepares a frame and returns the preprocessed frame as a Tensor of shape 
+    (N, 3, H, W).
+
+    Args:
+        frame (Union[Tensor, List[Tensor]]): Input frame(s).
+        .. warning::
+            Input ``frame`` should be of shape ``(N, 3, H, W)``, ``(N, 1, H, W)``, 
+            ``(3, H, W)``, ``(1, H, W)`` or ``(H, W)``.
+        preprocess (Callable, optional): Preprocessing function for the model. 
+            By default, this uses the preprocessing function for ``RAFT`` model.
+
+    Returns:
+        frame (Tensor): Preprocessed frame of shape (N, 3, H, W).
+        
+    """
+    frame_shape = frame.shape
+    if isinstance(frame, list):
+        device = frame[0].device
+        frame = torch.stack(frame).to(device)
+    else:
+        device = frame.device
+    # Ensure the frame has the shape: (N, C, H, W); C = 3
+    if frame.dim() == 3:
+        if frame.shape[0] == 1:
+            frame = torch.tile(frame, (3, 1, 1))
+        # Set shape: (1, 3, H, W); N = 1, C = 3
+        frame = frame.unsqueeze(0)
+    elif frame.dim() == 2:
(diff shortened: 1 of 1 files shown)


sugatoray:
@datumbox I am not sure if I should add this code where I did. I am not sure if this should go under `torchvision.utils`. I have not added any tests yet. Once it is determined where this code should be added to, I will work on adding the tests as well.

Also, I will post a correction/update to the RAFT demo with these functions.

current path: [torchvision/models/optical_flow/detector.py](https://github.com/pytorch/vision/pull/5649/files#diff-91f60a346c35c5725da3541938f167a5baa0206c8d39961f49f3a3f12fe40a4f)

datumbox:
@sugatoray First of all thanks for the contribution. 

It's not clear if this should be handled in the references of TorchVision or if this should be a separate utility added on your own repo. That's because we try to keep references as simple as possible and avoid multiple entry points. I will let @NicolasHug who has led the effort of supporting optical flow to provide input about whether specific optimisations of this PR are worth adopting on the current scripts.

A couple of notes that will help you maximise the chances of getting your contributions merged: If a PR touches more than 20 lines of code, it's usually recommended to first open an issue and discuss the changes. Discussing the reported issue with the rest of the community can help us decide if the reported is indeed an problem and how it should be addressed. For more info on the contribution process please check [this](https://github.com/pytorch/vision/blob/main/CONTRIBUTING.md)

sugatoray:
@datumbox Thank you. I will keep the "20 lines" in mind for future. Out of the two functions, `detect_flow` helps detect a single flow-pair. If this function is available, that would work as a helpful utility for people to use it directly, alike other functions in `torchvision.utils`. 

Also, for the demo notebook, I could push a separate PR, to include similar changes, but without the specific function in question (`detect_flow`).

torchvision/models/optical_flow/detector.py
```diff
@@ -0,0 +1,465 @@
...
+            f"""Input should be of shape (N, 3, H, W), (N, 1, H, W), 
+            (3, H, W), (1, H, W) or (H, W); got {frame_shape} 
+            and adjusted to {frame.shape}.
+            """))
+    if preprocess is None:
+        preprocess = raft_preprocess
+    return preprocess(frame).to(device)
+
+
+def detect_flow(frameA: Union[Tensor, List[Tensor]],
```
oke-aditya, line 95: Hi! @NicolasHug will have a deeper look.

I think that we do lot of implicit assumptions.

NicolasHug:
Thanks a lot for the PR @sugatoray . As @datumbox already noted, this kind of util is usually not included in the `models/` namespace; it's more likely to fit in the `utils/`. But here, it seems that this function is somewhat redundant with our [RAFT example](https://pytorch.org/vision/0.12/auto_examples/plot_optical_flow.html#sphx-glr-auto-examples-plot-optical-flow-py). 

>When I tried running the [RAFT demo](https://pytorch.org/blog/pytorch-1.11-new-library-releases/#2-optical-flow-support-and-raft-model) to create a series of optical-flow images, I found the colab notebook running out of memory

Could you please provide us with some details here? If the example runs out of memory, perhaps there's an issue that we can fix directly there :)

sugatoray:
@NicolasHug Yes, it ran out of memory.

sugatoray:
@datumbox So, will it be okay if I move the functions `detect_flow` and `detect_flows` under `utils/`?

torchvision/models/optical_flow/detector.py
sugatoray: @oke-aditya About the the assumptions on folder name, saving etc., it was necessary to save the files to disk. Now you may argue (*as I debated with myself*) that it should be perfectly fine returning just the `flow_img` and `predicted_flow` and let the user handle saving the files to disk. However, given the memory constraints, it might be useful, if you could just conveniently specify whether you want to save the raw and processed flow data to disk, or you want them returned as well.

Given, you have this function, `detect_flow` it all boils down to a single function call for generating optical flows. On the other hand, `detect_flows` handles the same for multiple images, which you can now create a video/gif file from.

So, yes, I had envisioned them as parts of `utils/`. But, just to begin with, pushed the initial commit to `torchvision/models/optical_flow/`.

sugatoray:
@NicolasHug The memory limit was struck when I tried running the last section to create multiple frames in a loop. Also the for loop had some obvious typos.

NicolasHug:
Thanks for the details @sugatoray . We'll be happy to review a PR that fixes these issues in the example. As noted there, this snippet that we commented out is only a starting point, not a full fledge solution.

sugatoray:
@NicolasHug I will send a PR with the fixes to the snippet section.

datumbox:
@sugatoray Shall we close this PR? I understand we are going ahead with a different approach, right?

sugatoray:
@datumbox Would it be possible to leave this PR to remain open for the time being, while I work on the other PR and submit it?

But, I was wondering if these functions could be added to `utils`. Please let me know your thoughts on that. Also, please let me know if you believe some changes to the functions could be a better option.

Thank you.

datumbox:
@sugatoray Happy to leave the PR open if you prefer. Please consider closing it once you are happy with the other PR to keep things tidy.

From @NicolasHug response, I understand that this is not something we probably want on the `utils` of references (this is something that people can't import from TorchVision anyway because references are not part of the binary). Nicolas, is my understanding wrong?

NicolasHug:
I'm not sure yet whether this would make sense as a util in the references either. It will depend on what are the needed changes to the snippet from the example, and why the changes are needed. It's important to keep in mind that the snippet is really a *starting point*, it doesn't pretend to be anything else, and there might be blind spots. The typos mentioned by @sugatoray are definitely worth fixing though.

sugatoray:
@datumbox and @NicolasHug Let me send the PR to introduce minimal changes to the RAFT example. Once it is working and the example is updated, perhaps you can take a more objective call on whether the functions in this PR are appropriate for `torchvision.utils` or not.

- `utils`: `torchvision/utils.py`

Does that sound reasonable to you?

datumbox:
Sounds good but this is definitely something we can't put on `torchvision.utils`. This area is not meant for model inference utilities. Let's get clarity over what the changes are and decide if there is any need to modify the examples.

Issue #4731, opened 2021-10-25
16 bits png support and compatibility with current transforms


NicolasHug:
We added support for decoding 16 bit pngs in https://github.com/pytorch/vision/pull/4657.

Pytorch doesn't support `uint16` dtype, so we have to return a `int32` tenor instead. However as @nairbv pointed out, this can lead to unexpected behaviour for the [`convert_image_dtype()`](https://github.com/pytorch/vision/blob/main/torchvision/transforms/functional_tensor.py#L58:L58), because this function relies on the max value of a given dtype to carry the conversions.

While the actual max value of a 16 bits image is `2**16 - 1`,  `convert_image_dtype()` will assume that it's `(2**31 - 1)` because the image is in a `int32` tensor. This leads to e.g. a tensor full of zeros when converting to uint8.

Potential solutions are:

- native support for uint16 in pytorch
- add a new `input_range=None` parameter to `convert_image_dtype()` (not a fan of that)

For now, I'll just make this a private API (e.g. `_decode_16bits_png()` or something). It will still allow to continue the work on RAFT, which was the main reason I needed 16bits decoding in the first place.

h-vetinari:
Trying to package torchvision 0.12 for conda-forge, all the test failures I see are:
```
        if "16" in img_path:
            # 16 bits image decoding is supported, but only as a private API
            # FIXME: see https://github.com/pytorch/vision/issues/4731 for potential solutions to making it public
>           with pytest.raises(RuntimeError, match="At most 8-bit PNG images are supported"):
E           Failed: DID NOT RAISE <class 'RuntimeError'>
```

Is there like a secret switch to allow these to pass - because (from what I can tell) it seems the respective images are there in the sources...

h-vetinari:
For context, in the so-called feedstock for conda-forge (latest [PR](https://github.com/conda-forge/tokenizers-feedstock/pull/40)), we're running the full torchvision test suite based on the github sources. So AFAIU, running into these "private" tests will be unavoidable, unless we skip/delete them, or there's a secret switch (I presume that's how the CI here does it, but haven't investigated yet...).

NicolasHug:
Thanks for the report @h-vetinari . The secret switch is just 

https://github.com/pytorch/vision/blob/fc63f82890651bb19f27e33eefb52465aeb1c25d/torchvision/csrc/io/image/cpu/decode_png.h#L12:L12

and the offending test calls `decode_image`

[...]

PR #5652, opened 2022-03-21
More robust check in tests for 16 bits images

NicolasHug:
Addresses https://github.com/pytorch/vision/issues/4731#issuecomment-1073787964

To determine whether an image is a 16 bits png, our tests just do `if "16" in img_path: ...` but that can break in some CIs where `16` is part of the path; this PR introduces a more robust check.

~It might be worth including in the bugfix release for package managers. @jdsgomes Do you know if/when the bugfix release is planned?~

### DIFF
--- a/test/test_image.py
+++ b/test/test_image.py
@@ -158,7 +158,7 @@ def test_decode_png(img_path, pil_mode, mode):
 
     img_pil = normalize_dimensions(img_pil)
 
-    if "16" in img_path:
+    if img_path.endswith("16.png"):
         # 16 bits image decoding is supported, but only as a private API
         # FIXME: see https://github.com/pytorch/vision/issues/4731 for potential solutions to making it public
         with pytest.raises(RuntimeError, match="At most 8-bit PNG images are supported"):


h-vetinari:
Thanks a lot for the quick handling!

Once ready (CI seems unhappy), we can backport this for conda-forge, so for our side it's not _critical_ that it's backported (though of course it would be nice 🙃).

The thing is that the way conda is set up, the test environment uses an int corresponding to a unix epoch in seconds (e.g. `torchvision-split_1647862604097`), and so we're _guaranteed_ to hit it, as it's in the two most significant bits for the... um... foreseeable future. 😅 
```
>>> pd.Series([pd.Timestamp.now()]).astype(np.int64)[0] // 10 ** 6
1647903504531
```

h-vetinari:
> Once ready (CI seems unhappy), we can backport this for conda-forge

When I wrote this I hadn't checked how simple the patch was. I tested this in https://github.com/conda-forge/torchvision-feedstock/pull/45 and it works for us - thanks! 🥳

NicolasHug:
Thanks for checking @h-vetinari ! According to your feedback, it might not be necessary to include it in the bugfix release.

Other than that I think this should be good to merge since the CI failure aren't related.

PR #5654, opened 2022-03-21
Minor updates to optical flow ref for consistency

NicolasHug:
Following up on https://github.com/pytorch/vision/pull/5590, some minor unifications

num_workers -> workers
use `epoch` instead of `current_epoch`
use `"."` instead of `"checkpoints"` for default output-dir


datumbox (approved):
LGTM, thanks!

Issue #5646, opened 2022-03-18
Some of torchvision transforms have very limiting argument checks


FeryET:
### 🐛 Describe the bug

I am trying to write a pytorch script that is configurable by Hydra. Hydra uses OmegaConf config objects, such as `ListConfig` and `DictConfig`. One of the benefits of using these types of configurations is that I can easily make templates and share them, alongside them being very readable.

```python 
# main.py
from omegaconf import DictConfig, OmegaConf
import hydra
from hydra.utils import get_class, instantiate


@hydra.main(config_path="conf/", config_name="config")
def train(cfg: DictConfig):
    OmegaConf.register_new_resolver(name="get_cls", resolver=lambda cls: get_class(cls))
    instantiate(cfg)


if __name__ == "__main__":
    train()
```
```YAML
# conf/config.yaml
transforms
    _target_: torchvision.tranforms.ColorJitter
    brightness: [0.5, 1.5]
```

By running `python -m main` given the configuration, this error will be thrown:
```
TypeError: Error instantiating 'torchvision.transforms.transforms.ColorJitter' : brightness should be a single number or a list/tuple with length 2.
```

When digging deeper, I came to  this line:
https://github.com/pytorch/vision/blob/297e2b8746d32b46d20b0e6a6f867fde65e50e8d/torchvision/transforms/transforms.py#L1177

The problem with OmegaConf.ListConfig's instance in this case is that it creates a sequence, but it is not of either of `List` or `Tuple`, but a`MutableSequence`. Given that `Hydra` is also being rapidly developed by a Facebook team, I think it is indeed in the great interest of PyTorch to facilitate such uses with simple fixes.

A simple fix that can make like easier can be using types such as `collections.Sequence` for such type checking, that is also what is usually recommended by Python devs.

### Versions

```
Collecting environment information...
PyTorch version: 1.11.0
Is debug build: False
CUDA used to build PyTorch: None
ROCM used to build PyTorch: N/A

OS: Ubuntu 20.04.4 LTS (x86_64)
GCC version: (Ubuntu 9.4.0-1ubuntu1~20.04) 9.4.0
Clang version: Could not collect
CMake version: Could not collect
Libc version: glibc-2.17

Python version: 3.7.11 (default, Jul 27 2021, 14:32:16)  [GCC 7.5.0] (64-bit runtime)
Python platform: Linux-5.13.0-27-generic-x86_64-with-debian-bullseye-sid
Is CUDA available: False
[...]

PR #5656, opened 2022-03-21
port ColorJitter to prototype transforms

### LINKED ISSUE #5528: Port `transforms.ColorJitter` to `prototype.transforms`

pmeier:
Closes #5528.


torchvision/prototype/transforms/color.py
```diff
@@ -0,0 +1,116 @@
...
+from ._utils import is_simple_tensor
+
+
+class ColorJitter(Transform):
+    def __init__(
+        self,
+        brightness: Optional[Union[float, Sequence[float]]] = None,
+        contrast: Optional[Union[float, Sequence[float]]] = None,
+        saturation: Optional[Union[float, Sequence[float]]] = None,
+        hue: Optional[Union[float, Sequence[float]]] = None,
```
pmeier, line 21: @FeryET I'm aware you wanted to work on this, but I needed `ColorJitter` for something else. Thus, I went ahead and implemented it. Could you check if the signature is now sufficient for what you were asking in #5646? I also adapted the `_check_input` method to reflect the changes.

torchvision/prototype/transforms/color.py
```diff
@@ -0,0 +1,116 @@
...
+            value = [center - value, center + value]
+            if clip_first_on_zero:
+                value[0] = max(value[0], 0.0)
+        elif isinstance(value, collections.abc.Sequence) and len(value) == 2:
+            if not bound[0] <= value[0] <= value[1] <= bound[1]:
+                raise ValueError(f"{name} values should be between {bound}")
+        else:
+            raise TypeError(f"{name} should be a single number or a sequence with length 2.")
+
+        return float(value[0]), float(value[1])
```
datumbox: You are missing:
https://github.com/pytorch/vision/blob/39772ece7c87ab38b9c2b9df8e7c85e967a739de/torchvision/transforms/transforms.py#L1183-L1186

I understand that you choose to configure `value = None` on the constructor for the same thing, but what you have here is not BC for the case where (0, 0) or (1, 1) etc.

torchvision/prototype/transforms/_color.py
```diff
@@ -0,0 +1,116 @@
...
+            if not bound[0] <= value[0] <= value[1] <= bound[1]:
+                raise ValueError(f"{name} values should be between {bound}")
+        else:
+            raise TypeError(f"{name} should be a single number or a sequence with length 2.")
+
+        return float(value[0]), float(value[1])
+
+    # TODO: remove this helper if primitive transforms for F.adjust_{brightness, contrast, saturation, hue} are added
+    class _ColorJitterHelper(Transform):
+        def __init__(self, kernel_tensor: Callable, kernel_pil: Callable, *args: Any, **kwargs: Any) -> None:
```
datumbox, line 56: Can you propose an alternative approach to the nested private class? Here you document quite a few reasons not to use it (logging, instantiation in every call, non-pythonic etc). Should this be a private method for instance?

Also all these weird idioms we are forced to use are papercuts for not having a high-level dispatcher. There are alternatives we could explore that avoid this.

datumbox (commented):
Thanks @pmeier, just a couple of comments

torchvision/prototype/transforms/color.py
pmeier: Good catch!

pmeier: To be fair, the behavior was not BC breaking. The difference is that the transformations would have been performed with a no-op parametrization. So if they don't have a shortcut for that, we get overhead, but not different results.

torchvision/prototype/transforms/_color.py
pmeier: > Can you propose an alternative approach to the nested private class?

I've added a commit that moves the functionality into a method.

> Also all these weird idioms we are forced to use are papercuts for not having a high-level dispatcher.

Yup.

> There are alternatives we could explore that avoid this.

Do you have anything specific in mind? I've been thinking on this for quite some time, and I haven't found a solution besides having primitive transforms like I mentioned in the code comment or to reinstate the "high level dispatcher kernels".

PR #5659, opened 2022-03-22
reverting some recently introduced exceptions

jdsgomes:
Removing exceptions and asserts from the forward pass in order to enable FX traceability. Instead reverting to torch._assert which is FX traceable


torchvision/models/detection/_utils.py
```diff
@@ -159,10 +159,8 @@ def encode_single(self, reference_boxes: Tensor, proposals: Tensor) -> Tensor:
         return targets
 
     def decode(self, rel_codes: Tensor, boxes: List[Tensor]) -> Tensor:
-        if not isinstance(boxes, (list, tuple)):
-            raise TypeError(f"This function expects boxes of type list or tuple, instead  got {type(boxes)}")
-        if not isinstance(rel_codes, torch.Tensor):
-            raise TypeError(f"This function expects rel_codes of type torch.Tensor, instead  got {type(rel_codes)}")
+        assert isinstance(boxes, (list, tuple))
+        assert isinstance(rel_codes, torch.Tensor)
```
datumbox, line 163: Should we be using `torch._assert()`?

torchvision/models/detection/anchor_utils.py
```diff
@@ -164,8 +162,8 @@ def __init__(
         clip: bool = True,
     ):
         super().__init__()
-        if steps is not None and len(aspect_ratios) != len(steps):
-            raise ValueError("aspect_ratios and steps should have the same length")
+        if steps is not None:
+            assert len(aspect_ratios) == len(steps)
```
datumbox, line 166: Why we need to undo this? It's part of the constructor where usually we don't care to be JIT-scriptable or FX-traceable. Do we initialize this directly in a forward?

torchvision/models/detection/retinanet.py
```diff
@@ -487,8 +487,8 @@ def forward(self, images, targets=None):
                 like `scores`, `labels` and `mask` (for Mask R-CNN models).
 
         """
-        if self.training and targets is None:
-            raise ValueError("In training mode, targets should be passed")
+        if self.training:
+            assert targets is not None
```
datumbox, line 491: The if statement 2 lines below does the exact same check. Shall we drop this completely? Also what about the ValueErrors in the rest of this method?

torchvision/ops/poolers.py
```diff
@@ -111,8 +111,7 @@ def _infer_scale(feature: Tensor, original_size: List[int]) -> float:
 def _setup_scales(
     features: List[Tensor], image_shapes: List[Tuple[int, int]], canonical_scale: int, canonical_level: int
 ) -> Tuple[List[float], LevelMapper]:
-    if not image_shapes:
-        raise ValueError("images list should not be empty")
+    assert len(image_shapes) != 0
```
datumbox, line 114: How come not the following?
```suggestion
    assert image_shapes
```

datumbox (commented):
Thanks @jdsgomes, just a few comments. Let me know your thoughts.

torchvision/ops/poolers.py
jdsgomes: I think I reverted the original asserts in most cases but yes you are right

torchvision/models/detection/_utils.py
jdsgomes: you mean everywhere or here in particular? 
I just reverted the changes to what we had previously. https://github.com/pytorch/vision/pull/5587/files#diff-5288f1a12d4cbc77afac42d57b75796da3c8734c0fa4cda80e5c07ac5d16212dL162

torchvision/models/detection/anchor_utils.py
jdsgomes: you are right, I think I mess this one up

torchvision/models/detection/retinanet.py
jdsgomes: You are right, I missed those because I went though all asserts that were replaced with exceptions. Those were there already as exceptions, but in the spirit of this PR should I change those too?

torchvision/models/detection/_utils.py
datumbox: I meant everywhere in this PR. I understand you are reverting the changes, what's unclear is what is the scope for them. If it's to enable fx-traceability then reverting alone is not enough and you would need to replace asserts with torch._asserts. There is a chance I misunderstand the target of the PR, so might be worth updating the description to explain better the what/why.

jdsgomes: Yes I was just reverting but you are right as per our offline discussion the entire point is to enable fx-traceability so made the changes accordingly

torchvision/models/detection/generalized_rcnn.py
```diff
@@ -58,8 +58,7 @@ def forward(self, images, targets=None):
 
         """
         if self.training:
-            if targets is None:
-                raise ValueError("In training mode, targets should be passed")
+            torch._assert(targets is not None, "targets should not be none when in training mode")
 
             for target in targets:
```
jdsgomes, line 63: the `if` above made the type checker happy: https://app.circleci.com/pipelines/github/pytorch/vision/16031/workflows/6e480a6f-2735-44f3-972f-dc44e8202d4c/jobs/1297766?invite=true#step-109-23

The problem is that without it the type checker assumes that target can be None in this part and None is not iterable.

I have seen cases that we remove the optional type hint but I would like to see if that's the standard approach here or we have other options other than keeping the exception. 

cc @datumbox

torchvision/models/detection/fcos.py
```diff
@@ -567,7 +567,7 @@ def forward(
         if self.training:
 
             if targets is None:
-                torch._assert(False, "targets should not be none when in training mode")
+                torch._assert(True, "targets should not be none when in training mode")
```
datumbox: I think you want this to be 
```suggestion
                torch._assert(False, "targets should not be none when in training mode")
```

Effectively you want ti to fail directly. If you pass True then the error message will never trigger.

datumbox:
@jdsgomes The problem [seems](https://app.circleci.com/pipelines/github/pytorch/vision/16055/workflows/872cc029-96a7-4130-850a-29d1442d8bb4/jobs/1299833) to be on the assertion message (type/tostring operator) of the `test_detection_model` tests. Have you tried changing the message to see if the issue gets resolved? There are also some other exceptions that are being triggered at `test_boxes_shape`. There might be a few more of them that might be handling.

Do you mind converting the PR to draft and ping us when you want a review?

torchvision/models/detection/_utils.py
```diff
@@ -375,7 +378,8 @@ def __call__(self, match_quality_matrix: Tensor) -> Tensor:
 
         if self.allow_low_quality_matches:
             if all_matches is None:
-                raise ValueError("all_matches should not be None")
+                torch._assert(False, "all_matches should not be None")
+                return torch.empty((0), dtype=torch.int64)  # not reachable - added to make type checker happy
```
datumbox, line 382: So clearly this is one of the things we need to do to please JIT/FX, so it's going to be ugly. Between adding a fake return and using an `if/else` idiom (when possible) I would much prefer the latter. Is the fake return necessary in this case or you can just use an else clause for the method below?

torchvision/models/detection/fcos.py
```diff
@@ -619,10 +621,10 @@ def forward(
         losses = {}
         detections: List[Dict[str, Tensor]] = []
         if self.training:
-            # compute the losses
             if targets is None:
-                raise ValueError("targets should not be none when in training mode")
-
+                torch._assert(False, "targets should not be none when in training mode")
+                return ({}, [{}])  # not reachable - added to make type checker happy
+            # compute the losses
```
datumbox, line 627: Same here, have we tried an if/else?

torchvision/models/detection/generalized_rcnn.py
```diff
@@ -59,23 +59,25 @@ def forward(self, images, targets=None):
         """
         if self.training:
             if targets is None:
-                raise ValueError("In training mode, targets should be passed")
-
+                torch._assert(False, "targets should not be none when in training mode")
+                return ({}, [{}])  # not reachable - added to make type checker happy
```
datumbox, line 63: same

torchvision/models/detection/retinanet.py
```diff
@@ -554,7 +553,8 @@ def forward(self, images, targets=None):
         detections: List[Dict[str, Tensor]] = []
         if self.training:
             if targets is None:
-                raise ValueError("In training mode, targets should be passed")
+                torch._assert(False, "targets should not be none when in training mode")
+                return ({}, [{}])  # not reachable - added to make type checker happy
```
datumbox, line 557: Same

torchvision/models/detection/ssd.py
```diff
@@ -379,8 +380,8 @@ def forward(
         detections: List[Dict[str, Tensor]] = []
         if self.training:
             if targets is None:
-                raise ValueError("targets should not be None when in training mode")
-
+                torch._assert(False, "targets should not be none when in training mode")
+                return ({}, [{}])  # not reachable - added to make type checker happy
```
datumbox, line 384: Same

datumbox (commented):
Thanks @jdsgomes. I understand that's far from a perfect situation, I left one comment below and marked all places where it appears. Let me know your thoughts.

torchvision/models/detection/_utils.py
jdsgomes: Else works I think (i made so many changes I am not sure anymore but I can test again). 
But I think despite both are ugly , the else option would have to ident the entire body which seems to be more distracting than the fake return.

datumbox: Indeed, we are between 2 problematic solutions here. 

Note that this won't be the only place where we have "fake code" to handle JIT (we have fake [casts](https://github.com/pytorch/vision/blob/a0867fac8dd13e436a4f756d3de8b9d6768c391e/torchvision/models/detection/transform.py#L19-L22) and [typing](https://github.com/pytorch/vision/blob/a0867fac8dd13e436a4f756d3de8b9d6768c391e/torchvision/transforms/functional.py#L364)) but we typically use them as last resort. Here a valid idiom exists (using `else`) which is why I think it should be preferred.

datumbox (approved):
Thanks for all the updates. Feel free to merge on green CI.

PR #5663, opened 2022-03-23
port RandomPhotoMetricDistort to prototype transforms

pmeier:
Addresses #5542

### DIFF
--- a/torchvision/prototype/transforms/_color.py
+++ b/torchvision/prototype/transforms/_color.py
@@ -6,8 +6,9 @@ import PIL.Image
 import torch
 from torchvision.prototype import features
 from torchvision.prototype.transforms import Transform, functional as F
+from torchvision.transforms import functional as _F
 
-from ._utils import is_simple_tensor
+from ._utils import is_simple_tensor, get_image_dimensions, query_image
 
 T = TypeVar("T", features.Image, torch.Tensor, PIL.Image.Image)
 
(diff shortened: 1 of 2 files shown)


torchvision/prototype/transforms/_color.py
```diff
@@ -120,5 +121,70 @@ def _transform(self, input: Any, params: Dict[str, Any]) -> Any:
 
         for transform in params["image_transforms"]:
             input = transform(input)
+        return input
+
+
+class _RandomChannelShuffle(Transform):
```
pmeier, line 127: I made this private for now, but we could also make it public. In that case, we should have a `image_permute_channels` kernel.

torchvision/prototype/transforms/_color.py
```diff
@@ -120,5 +121,70 @@ def _transform(self, input: Any, params: Dict[str, Any]) -> Any:
...
+    def _get_params(self, sample: Any) -> Dict[str, Any]:
+        image = query_image(sample)
+        num_channels, _, _ = get_image_dimensions(image)
+        return dict(permutation=torch.randperm(num_channels))
+
+    def _transform(self, input: Any, params: Dict[str, Any]) -> Any:
+        if not (isinstance(input, (features.Image, PIL.Image.Image)) or is_simple_tensor(input)):
```
pmeier, line 134: The structure is different from the usual if / elif / else here, because we use the same "kernel" for all supported types. If we have dedicated kernels for this as proposed above, we can go back to the regular structure.

datumbox (approved):
The implementation looks good. The problem is that this approach is bound to be quite slow as there will be lots of dispatching involved. When later we try to make the transforms as fast as possible, what are we going todo? We would need to remove the composability that you have and call the low-level kernels.

I'm OK to merge now as we shouldn't try to optimize the speed prematurely but that's something to keep in mind if we keep using composition as an idiom from many other transforms.

PR #5667, opened 2022-03-23
Add sun397 prototype datapipe

YosuaMichael:
Related issues: https://github.com/pytorch/vision/issues/5351

Adding SUN397 dataset into the builtin prototype datapipe.


torchvision/prototype/datasets/utils/_resource.py
```diff
@@ -162,10 +167,12 @@ def resolve(self) -> OnlineResource:
             for attr in (
                 "file_name",
                 "sha256",
-                "_preprocess",
-                "_loader",
```
YosuaMichael, line 166: Remove _loader because it cause error and as of now we can't see why it needed here (as self._loader is a function defined in the parent class)

test/builtin_dataset_mocks.py
```diff
@@ -1431,3 +1431,19 @@ def stanford_cars(info, root, config):
...
+
+
+@register_mock
+def sun397(info, root, config):
+    num_samples = 2
+    categories = ["abbey", "bakery/shop", "bar"]
+    for category in categories:
+        create_image_folder(
+            root=root,
+            name="{}/{}".format(category[0], category),
```
pmeier, line 1443: This will break on Windows. If the `/` actually means subdirectory, we need to handle that.

torchvision/prototype/datasets/_builtin/sun397.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def _filter_image_files(self, data: Tuple[str, Any]) -> bool:
+        match = self._IMAGE_FILENAME_PATTERN.match(pathlib.Path(data[0]).name)
+        return bool(match)
+
+    def _make_datapipe(
+        self, resource_dps: List[IterDataPipe], *, config: DatasetConfig
+    ) -> IterDataPipe[Dict[str, Any]]:
+        dp = resource_dps[0]
+        dp = Filter(dp, self._filter_image_files)
```
pmeier, line 55: Are there any images in the archive that end in `.jpg` but should be filtered out by not having starting with `sun_`. If not this can be simplified to

```suggestion
        dp = Filter(dp, path_comparator("suffix", ".jpg"))
```

You can import it like

```py
from torchvision.prototype.datasets.utils._internal import path_comparator
```

test/builtin_dataset_mocks.py
```diff
@@ -1431,3 +1431,19 @@ def stanford_cars(info, root, config):
...
+            file_name_fn=lambda idx: f"sun{idx:5d}.jpg",
+            num_examples=num_samples,
+        )
+
+    make_tar(root, "SUN397.tar.gz")
```
pmeier, line 1448:
```suggestion
    make_tar(root, "SUN397.tar.gz", compression="gz")
```

pmeier (commented):
Hey @YosuaMichael and welcome to `torchvision` team from me :hugs: The PR looks good in general. I've added a few comments inline. 

One major thing that we need to address is the last point in https://github.com/pytorch/vision/pull/5164#issuecomment-1005813311. cc @NicolasHug I'm sorry, I forgot to mention this on the porting issue. The point is that the dataset defines a train and test split as well as 10 different folds. We postponed handling this for the legacy dataset, but should handle this now.

torchvision/prototype/datasets/utils/_resource.py
```diff
@@ -162,10 +167,12 @@ def resolve(self) -> OnlineResource:
...
             )
         }
+        # Handle the _preprocess attr
```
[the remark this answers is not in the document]
yassineAlouini: @pmeier Yes exactly, it seems to match what I was encountering. :ok_hand:

test/builtin_dataset_mocks.py
```diff
@@ -1431,3 +1431,19 @@ def stanford_cars(info, root, config):
...
     return num_samples
+
+
+@register_mock
+def sun397(info, root, config):
+    num_samples = 2
+    categories = ["abbey", "bakery/shop", "bar"]
+    for category in categories:
+        create_image_folder(
+            root=root,
```
pmeier, line 1442:
```suggestion
            root=root / "SUN397",
```

torchvision/prototype/datasets/_builtin/sun397.py
YosuaMichael: Ah yeah, just checked and every `.jpg` have prefix `sun`. So I can change to your suggestion, thanks!

test/builtin_dataset_mocks.py
YosuaMichael: Got it, I think I can fix this with `pathlib`. Thanks!

YosuaMichael:
Hey @pmeier thanks a lot for your review!

Regarding your comment:

> One major thing that we need to address is the last point in https://github.com/pytorch/vision/pull/5164#issuecomment-1005813311. cc @NicolasHug I'm sorry, I forgot to mention this on the porting issue. The point is that the dataset defines a train and test split as well as 10 different folds. We postponed handling this for the legacy dataset, but should handle this now.

I understand from the dataset homepage: https://vision.princeton.edu/projects/2010/SUN/ they explain about the train and test (they have 10 different pair of train and test). Currently my idea is to have a split (train, test) and fold (1-10). Will try to do this.

torchvision/prototype/datasets/utils/_resource.py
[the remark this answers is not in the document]
pmeier: I've just merged #5669, so this workaround should no longer be needed. Could you revert this and update your branch?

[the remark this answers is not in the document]
YosuaMichael: @pmeier Sure, I have merged my branch to the main branch and I used your change for _resource.py

test/builtin_dataset_mocks.py
```diff
@@ -1431,3 +1431,48 @@ def stanford_cars(info, root, config):
         make_tar(root, "car_devkit.tgz", devkit, compression="gz")
 
     return num_samples
+
+
+@register_mock
+def sun397(info, root, config):
+    images_root = root / "SUN397"
+    images_root.mkdir()
```
pmeier, line 1439: No need for `os.makedirs()` here. In general, `pathlib` supports everything you need with paths. `.mkdir()` also takes the `exist_ok: bool` and `parents: bool` flags for more complex situations.

test/builtin_dataset_mocks.py
```diff
@@ -1431,3 +1431,48 @@ def stanford_cars(info, root, config):
...
+
+    categories = ["abbey", "airplane_cabin", "wrestling_ring/indoor"]
```
pmeier, line 1441: For the category selection it is good practice to select ones with different properties. I've replaced some of the old ones with ones with underscores. Plus, I usually take categories from the start and the end of the list of categories.

test/builtin_dataset_mocks.py
```diff
@@ -1431,3 +1431,48 @@ def stanford_cars(info, root, config):
...
+    for key in category_keys:
+        parts = key.split("/")
+        image_files = create_image_folder(
+            root=images_root.joinpath(*parts[1:-1]),
+            name=parts[-1],
+            file_name_fn=lambda idx: f"sun_{idx:05d}.jpg",
+            num_examples=5,
+        )
+
+        keys.extend([f"/{image_file.relative_to(images_root).as_posix()}" for image_file in image_files])
```
pmeier, line 1454: After the image generation, we no longer care for the actual image path, but only its key.

test/datasets_utils.py
```diff
@@ -952,3 +952,24 @@ def make_fake_flo_file(h, w, file_name):
     )
     with open(file_name, "wb") as f:
         f.write(content)
+
+
+def random_group(collection, groups):
```
pmeier, line 957: I wrote this initially for #5651, but it is very useful here too.

test/builtin_dataset_mocks.py
```diff
@@ -1431,3 +1431,48 @@ def stanford_cars(info, root, config):
...
+
+    partitions_root = root / "Partitions"
+    partitions_root.mkdir()
+
+    for fold in range(1, 11):
+        random.shuffle(keys)
+
+        for split, keys_in_split in random_group(keys, ["train", "test"]).items():
```
pmeier, line 1462: This randomly splits all our keys into two groups while making sure that each group has at least one element. Thus, we get a random number of samples for each config, which makes the common tests more robust.

test/builtin_dataset_mocks.py
```diff
@@ -1431,3 +1431,48 @@ def stanford_cars(info, root, config):
...
+            if split == config.split and str(fold) == config.fold:
+                num_samples = len(keys_in_split)
+
+            with open(partitions_root / f"{split.capitalize()}ing_{fold:02d}.txt", "w") as fh:
+                fh.write("\n".join(sorted(keys_in_split)))
+
+    # Both archives contain this file. Although it shouldn't be used at runtime, it serves as sentinel to test the
+    # filtering of the correct files
```
pmeier, line 1470: See https://github.com/pytorch/vision/blob/main/torchvision/prototype/datasets/_builtin/README.md#tests for details.

pmeier (commented):
@YosuaMichael The overarching change that simplified the mock data generation is similar to what we discussed yesterday offline: if the dataset provides "keys" for the data, it is preferable to use them. See detailed comments inline.

test/datasets_utils.py
```diff
@@ -952,3 +952,24 @@ def make_fake_flo_file(h, w, file_name):
...
+    """Randomly put items into groups without overlap.
+    Args:
+        collection: Collection of items to be grouped.
+        groups: Collection of group keys.
+    Returns:
+        Dictionary with ``groups`` as keys. Each value is a list of random items from ``collection`` without overlap
+            to the other values. Each list has at least length ``1``.
```
NicolasHug, line 964: Perhaps a docstring example would be helpful to understand what this really does?

test/builtin_dataset_mocks.py
NicolasHug: What's the reason for having random subsets here? In general, we need our tests to be deterministic

test/datasets_utils.py
pmeier: I refactored to have a `random_subsets` function rather than `random_groups`. The docstring now also includes a usage example. LMK if it is sufficient.

torchvision/prototype/datasets/_builtin/sun397.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+class SUN397(Dataset):
+    def _make_info(self) -> DatasetInfo:
+        return DatasetInfo(
+            "sun397",
+            homepage="https://vision.princeton.edu/projects/2010/SUN/",
+            valid_options=dict(
+                split=("train", "test"),
+                fold=tuple(str(fold) for fold in range(1, 11)),
```
NicolasHug, line 23: @pmeier 

In the original implementation of this dataset we decided not to include the fold parameter and also removed the split param https://github.com/pytorch/vision/pull/5164#issuecomment-1005813311. Should we keep things consistent here?

pmeier: That would ignore a lot of information the dataset provides. IIRC, the only reason we were ok to remove it, was that this was one of the last datasets in the old API and FLAVA didn't need the functionality. Given that this no longer applies, IMO we should handle this now.

NicolasHug: From what I recall, we decided not to include a `fold` parameter because:

- it's not clear what it would be used for for most users - FLAVA is our only known use-case ATM, and they don't need it
- we can add it in a BC way if users show a clear need for it
- The use of the `fold` parameter would set a precedent, and might conflict with other datasets like DTD which also partition the dataset, but in a different way. So by not supporting this we're also minimizing the chances of inconsitencies in our API.

I think all of these still apply today, so unless we have a strong and obvious use-case for including it, I would prefer to go with the status quo here, which would give us the maximal flexibility for future us.

YosuaMichael: Hey @NicolasHug @pmeier , I changes the valid_options to the following:
It only have `split` options which value can be:

- "all" (default) -> this will give same datasets as old API, and since this is the default it will BC with the old API
- "train-{fold}", sample: "train-5" -> this will give dataset of train at fold=5 according to the paper
- "test-{fold}", sampple: "test-5" -> this will give dataset of test at fold=5 according to the paper

What are your thought on this?

Issue #5520, opened 2022-03-03
port transforms from the old to the new API


pmeier:
We are currently in the process of revamping the `transforms` module, but there is still a lot of porting left. The design is now stable enough that the porting process should be manageable for someone not intimately familiar with the new design. Thus, we are actively looking for contributions helping us finishing this. 

Here is the list of transformations that still need to be ported to achieve feature parity (minus some special transformations) between the new and old API:

- [x] #5521
- [x] #5522
- [x] #5523
- [x] #5524
- [x] #5525
- [x] #5526
- [x] #5527
- [x] #5528
- [x] #5529
- [x] #5530
- [x] #5531
- [x] #5532
- [x] #5533
- [x] #5534
- [x] #5535
- [x] #5536
- [x] #5537
- [x] #5538
- [x] `RandomHorizontalFlip` #5421
- [x] `Resize` #5421
- [x] `CenterCrop` #5421
- [x] `RandomResizedCrop` #5421
- [x] `Normalize` #5421
- [x] `RandomErasing` #5421
- AutoAugment transforms
  - [x] `RandAugment` #5421
  - [x] `TrivialAugmentWide` #5421
  - [x] `AutoAugment` #5421
  - [x] #5492

There is an issue for each transformation. Please comment on that if you want to take up a task so we can assign it to you.

---

Here is a recipe on how the porting process looks like:

1. Port the kernels from `transforms.functional_tensor` and `transforms.functional_pil` to `prototype.transforms.functional`. In most cases this means just binding them to a new name

    https://github.com/pytorch/vision/blob/97385df0f675b805d974994389a6ea2f0de51a19/torchvision/prototype/transforms/functional/_geometry.py#L14-L15
    
    The naming scheme is `{kernel_name}_{feature_type}_{feature_subtype}`. To use the example above, the kernel name is `horizontal_flip`, the feature type is `image` and the subtypes are `tensor` and `pil`.
    
[...]

PR #5668, opened 2022-03-23
cleanup prototype transforms functional test

pmeier:
The goal of this PR is to facilitate best practices for our tests. In a normal scenario, these changes would be considered nitpicks. In the light of #5520 and new contributors usually (and rightfully) copy-paste code from similar tests, I think it would be beneficial to have proper usages early on. 

There are two patterns I found:

1. The constructor of the new features accepts a `device` kwarg. Thus, something like `features.Image(...).to(device)` is an anti-pattern and should be replaced by `features.Image(..., device=device)`.
2. `torch.testing.assert_close` is able to handle nested sequence comparison. For example, in the bounding box reference tests, we want to test the output tensor of a kernel against a two level list of values computed with another library. Instead of writing custom comparison logic for this scenario, we can simply do `torch.tensor(tensor.tolist(), list_of_lists)`.

cc @vfdev-5

### DIFF
--- a/test/test_prototype_transforms_functional.py
+++ b/test/test_prototype_transforms_functional.py
@@ -345,12 +345,15 @@ def test_correctness_affine_bounding_box(angle, translate, scale, shear, center)
             np.max(transformed_points[:, 1]),
         ]
         out_bbox = features.BoundingBox(
-            out_bbox, format=features.BoundingBoxFormat.XYXY, image_size=bbox.image_size, dtype=torch.float32
+            out_bbox,
+            format=features.BoundingBoxFormat.XYXY,
+            image_size=bbox.image_size,
+            dtype=torch.float32,
+            device=bbox.device,
         )
-        out_bbox = convert_bounding_box_format(
+        return convert_bounding_box_format(
             out_bbox, old_format=features.BoundingBoxFormat.XYXY, new_format=bbox.format, copy=False
         )
-        return out_bbox.to(bbox.device)
 
     image_size = (32, 38)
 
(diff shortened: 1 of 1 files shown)


vfdev-5:
@pmeier `torch.testing` is in beta and [may change without warning in future PyTorch releases](https://pytorch.org/docs/stable/testing.html), how problematic this could be to maintain ?

pmeier:
As of pytorch/pytorch#73348 roughly a month ago, `torch.testing` is considered stable.

test/test_prototype_transforms_functional.py
```diff
@@ -583,6 +584,4 @@ def test_correctness_rotate_bounding_box_on_fixed_input(device, expand):
         center=center,
     )
 
-    assert len(output_boxes) == len(expected_bboxes)
-    for a_out_box, out_box in zip(expected_bboxes, output_boxes.cpu()):
-        np.testing.assert_allclose(out_box.cpu().numpy(), a_out_box)
+    torch.testing.assert_close(output_boxes.tolist(), expected_bboxes)
```
NicolasHug, line 587: `assert_close` refuses to compare a list to a tensor?
IIRC numpy testing utils allow to compare array-likes with array-likes right?

NicolasHug (approved):
Thanks @pmeier , I agree it's worth encouraging good test practices and it's best done by example. Just a random question but LGTM

test/test_prototype_transforms_functional.py
pmeier: Yes and yes. `numpy` testing functions coerce everything into an `np.ndarray` first and than compare. We consciously opted against that for mainly two reasons:

1. If you have tests parametrized over operators like PyTorch core does, it is impossible to see if the returned type is correct. For example

```py
>>> expected = 1.0
>>> np.testing.assert_allclose(np.array(expected), expected)
>>> torch.testing.assert_close(torch.tensor(expected), expected)
TypeError: No comparison pair was able to handle inputs of type <class 'torch.Tensor'> and <class 'float'>.
```

2. It allows us to compare sequences of tensors with different shapes elementwise

    ```py
    >>> shapes = [(2, 3), (2, 4)]
    >>> np.testing.assert_allclose(*([np.ones(shape) for shape in shapes],) * 2)
    ValueError: could not broadcast input array from shape (2,3) into shape (2,)
    >>> torch.testing.assert_close(*([torch.ones(shape) for shape in shapes],) * 2)
    ```

     This aligns nicely with the ability of `torch.testing.assert_close` to compare mappings elementwise, which is not possible with the `numpy` functions.

PR #5669, opened 2022-03-23
fix HttpResource.resolve() with preprocessing

pmeier:
#5282 removed the `preprocess` flag from `OnlineResource`. In #5584 and #5667 it was discovered that this broke the use case if either `decompress` or `extract` is set on a `HttpResource` and the URL redirects. @YosuaMichael provided a solution in #5667, but I think it will be much better to have a more general solution.

This PR removes the `decompress` and `extract` parameters and re-adds `preprocess`. It can be any `Callable[[pathlib.Path], pathlib.Path]`. To keep the old convenience, one can also pass `preprocess="decompress"` or `preprocess="extract"`.

### DIFF
--- a/test/test_prototype_datasets_utils.py
+++ b/test/test_prototype_datasets_utils.py
@@ -5,6 +5,7 @@ import pytest
 import torch
 from datasets_utils import make_fake_flo_file
 from torchvision.datasets._optical_flow import _read_flo as read_flo_ref
+from torchvision.prototype.datasets.utils import HttpResource, GDriveResource
 from torchvision.prototype.datasets.utils._internal import read_flo, fromfile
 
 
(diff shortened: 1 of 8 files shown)


NicolasHug (commented):
Thanks for the PR @pmeier 

If i remember correctly, the way redirections are done is fairly hacky. Would it be worth thinking of a roboust way of doing these?

NicolasHug (approved):
Approving to unblock but if we're going to revisit everything anyway, perhaps there are more minimal changes that would work too (and that would be easier to review)? Not sure how tied the proposed changes are to the actual fix here.

pmeier:
> more minimal changes that would work too (and that would be easier to review)?

IMO, the changes are fairly minimal. As stated in my top comment, have a look at #5667 for an alternative solution. The gist from there is that we internally track whether `decompress` or `extract` was set to be able to pass the right values when instantiating a new resource after resolving.

PR #5671, opened 2022-03-24
add set_weight_decay to support custom weight decay setting

xiaohu2015:
In this PR, we want to add a more generic method to replace `split_normalization_params` in torchvision.ops. The `split_normalization_params` method only supports no weight decay for the parameters of the normalization layers. But for some recent models (eg. vit, swint), we need custom weight decay for other parameters (eg. class token, positision_embedding). This PR is related to https://github.com/pytorch/vision/pull/5491.

For current implementation, the priority is: **custom_keys > norm layers > others**.

two test case:

```python
    model = models.mobilenet_v3_large()
    params_1 = ops._utils.split_normalization_params(model)
    params_2 = set_weight_decay(model, 0.2, norm_weight_decay=0.1)
    params_2.sort(key=lambda x: x["weight_decay"])
    params_3 = set_weight_decay(model, 0.2)
    
    assert len(params_1[0]) == len(params_2[0]["params"]) == 92
    assert len(params_1[1]) == len(params_2[1]["params"]) == 82
    assert len(params_3) == 1, len(params_3[0]["params"]) == 174
    assert params_2[0]["weight_decay"] == 0.1, params_2[1]["weight_decay"] == 0.2
    assert params_3[0]["weight_decay"] == 0.2
```

```python
    @pytest.mark.parametrize("norm_weight_decay", [None, 0.2])
    @pytest.mark.parametrize("norm_layer", [None, nn.LayerNorm])
    @pytest.mark.parametrize("custom_keys_weight_decay", [None, [("class_token", 0.3), ("pos_embedding", 0.4)]])
    def test_set_weight_decay(self, norm_weight_decay, norm_layer, custom_keys_weight_decay):
        model = models.VisionTransformer(
            image_size=224,
            patch_size=16,
            num_layers=1,
            num_heads=2,
            hidden_dim=8,
            mlp_dim=4,
        )
        param_groups = ops._utils.set_weight_decay(
            model,
            0.1,
            norm_weight_decay=norm_weight_decay,
            norm_classes=None if norm_layer is None else [norm_layer],
            custom_keys_weight_decay=custom_keys_weight_decay,
        )

        if norm_weight_decay is None and custom_keys_weight_decay is None:
            assert len(param_groups) == 1
            assert len(param_groups[0]["params"]) == 20

        if norm_weight_decay is not None and custom_keys_weight_decay is None:
            assert len(param_groups) == 2
            param_groups.sort(key=lambda x: x["weight_decay"])
            assert len(param_groups[0]["params"]) == 14
            assert len(param_groups[1]["params"]) == 6

        if norm_weight_decay is not None and custom_keys_weight_decay is not None:
            assert len(param_groups) == 4
            param_groups.sort(key=lambda x: x["weight_decay"])
            assert len(param_groups[0]["params"]) == 12
            assert len(param_groups[1]["params"]) == 6
            assert len(param_groups[2]["params"]) == 1
            assert len(param_groups[3]["params"]) == 1
```

### references
- https://github.com/rwightman/pytorch-image-models/blob/master/timm/optim/optim_factory.py
- https://github.com/open-mmlab/mmcv/blob/master/mmcv/runner/optimizer/default_constructor.py


references/classification/train.py
```diff
@@ -229,12 +229,16 @@ def main(args):
...
 
-    if args.norm_weight_decay is None:
-        parameters = model.parameters()
-    else:
-        param_groups = torchvision.ops._utils.split_normalization_params(model)
-        wd_groups = [args.norm_weight_decay, args.weight_decay]
-        parameters = [{"params": p, "weight_decay": w} for p, w in zip(param_groups, wd_groups) if p]
+    custom_keys_weight_decay = None
+    if hasattr(model, "no_weight_decay_keys"):
+        custom_keys_weight_decay = {k: 0.0 for k in model.no_weight_decay_keys()}
```
datumbox, line 234: I don't think we should introduce this type of API in the models. Whenever we attach custom properties in models we risk compatibility with solutions like JIT/FX etc. Is it an option to pass these as arguments?

torchvision/ops/_utils.py
```diff
@@ -61,3 +61,64 @@ def split_normalization_params(
...
+            custom_params[key] = []
+
+    for module in model.modules():
+        if next(module.children(), None):
+            for name, p in module.named_parameters(recurse=False):
+                if not p.requires_grad:
+                    continue
+                is_custom_key = False
+                for key in custom_params:
+                    if key == name:
```
datumbox, line 99: Can line 97-103 be simplified further? For example something like:
```python
if name in custom_params:
    custom_params[key].append(p)
else:
    other_params.append(p)
```

torchvision/ops/_utils.py
```diff
@@ -61,3 +61,64 @@ def split_normalization_params(
...
+
+def set_weight_decay(
+    model: torch.nn.Module,
+    weight_decay: float,
+    norm_weight_decay: Optional[float] = None,
+    norm_classes: Optional[List[type]] = None,
+    bias_weight_decay: Optional[float] = None,
+    custom_keys_weight_decay: Optional[Dict[str, float]] = None,
+):
+    if not norm_classes:
```
datumbox, line 74: Given that you now provide a custom_keys solution, I don't think we need to support also custom `norm_classes`. We can just hardcode them and ask users to use the custom_keys feature if they need something different. Thoughts?

torchvision/ops/_utils.py
```diff
@@ -61,3 +61,64 @@ def split_normalization_params(
...
+                        is_custom_key = True
+                if not is_custom_key:
+                    other_params.append(p)
+        elif isinstance(module, norm_classes) and norm_weight_decay is not None:
+            norm_params.extend(p for p in module.parameters() if p.requires_grad)
+        else:
+            for name, p in module.named_parameters():
+                if not p.requires_grad:
+                    continue
+                if name == "bias" and bias_weight_decay is not None:
```
datumbox, line 110: Is it possible to move the handling of bias on the first if? If so, I think this might make the code easier to follow.

torchvision/ops/_utils.py
```diff
@@ -61,3 +61,64 @@ def split_normalization_params(
...
+
+
+def set_weight_decay(
+    model: torch.nn.Module,
+    weight_decay: float,
+    norm_weight_decay: Optional[float] = None,
+    norm_classes: Optional[List[type]] = None,
+    bias_weight_decay: Optional[float] = None,
+    custom_keys_weight_decay: Optional[Dict[str, float]] = None,
+):
```
datumbox, line 73: This method is a straight extension of `split_normalization_params`. We shouldn't keep both. I think what we could do is "prove" in this PR that the new method works the same as `split_normalization_params` for the right configuration and then remove it. It's currently private so this won't be a problem.

datumbox (commented):
@xiaohu2015 Thanks for the contribution.

I've added a few comments, let me know your thoughts.

references/classification/train.py
xiaohu2015: @datumbox in fact, the method `no_weight_decay_keys` is not included in the `forward` function, so I think it has no risk compatibility with solutions like JIT/FX etc. If we pass the `keys` as args, I think we should be careful to pass the argument for each model, maybe some users can forget it?

torchvision/ops/_utils.py
xiaohu2015: yes, but the custom keys in this function are only applied in the user-defined `Parameters`, the predefined modules in torch (eg. nn.LayerNorm) are not considered.

xiaohu2015: I think I must give some explanation：
- here we only user-defined `Parameters` for custom_keys
- the norm layer is firstly considered, if we set norm_weight_decay, the `bias parameter` in the norm layer use the norm_weight_decay even we set bias_norm_weight_decay
- we only consider the bias parameter in the predefined modules in torch.

torchvision/ops/_utils.py
xiaohu2015: yes, but for sake I don't remove this function.

torchvision/ops/_utils.py
datumbox: Sorry, I should have been clearer. No that's not what I meant.

I mean to move the handling of the bias on the first `if` statement section (aka add an other `elif` at current line 99).

torchvision/ops/_utils.py
datumbox: Thanks a lot for the clarifications, this was useful! 

The side-effect of the bias on Norm layers wan't apparent to me while reading the code. Could we update the code (refactoring + simplification + documenting) so that these corner cases are more apparent?

What I originally had in mind is that the functionality of Normalization doesn't work exactly as someone wants, they can easily overwrite its behaviour using the custom  keys. I think the custom keys should overwrite all other options. Let me know your thoughts. Happy to have a chat offline if you think that helps clarify the points easier.

references/classification/train.py
datumbox: Historically we have been avoiding properties/attributes on the nn.Module and try to keep it barebone. We could discuss pros/cons for this on a separate RFC but I think we shouldn't link this PR to the resolution of it because it will cause delays. I think passing `no_weight_decay_keys` is fine as long as we document them as part of the training command but this depends on how many arguments you have. Could you provide an example? 

An alternative approach could be to add this specifically for Swin/ViT in a utility method in references as a temporary solution, until the RFC is discussed.

torchvision/ops/_utils.py
datumbox: Sorry, I don't understand what you mean. Do you intent to remove it later before the PR is merged?

xiaohu2015: yes

torchvision/ops/_utils.py
```diff
@@ -61,3 +61,61 @@ def split_normalization_params(
...
+    if not norm_classes:
+        norm_classes = [
+            nn.modules.batchnorm._BatchNorm,
+            nn.LayerNorm,
+            nn.GroupNorm,
+            nn.modules.instancenorm._InstanceNorm,
+            nn.LocalResponseNorm,
+        ]
+    norm_classes = tuple(norm_classes)
```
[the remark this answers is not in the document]
xiaohu2015: I copy this code from `split_normalization_params`

torchvision/ops/_utils.py
```diff
@@ -61,3 +61,66 @@ def split_normalization_params(
...
+        if norm_weight_decay is not None and isinstance(module, norm_classes):
+            params["norm"].extend(p for p in module.parameters() if p.requires_grad)
+        else:
+            for name, p in module.named_parameters(recurse=False):
+                if not p.requires_grad:
+                    continue
+                is_custom_key = False
+                for key in custom_keys:
+                    full_name = f"{prefix}.{name}" if prefix != "" else name
+                    if key in full_name:
```
datumbox, line 109: This `in` statement could lead to issues in some corner-case. For example say that a module has multiple convolutions layers named `conv`, `conv2`, ... OR `conv1`, ..., `conv10`. This can lead to the `if` activating in more cases than we want.

references/classification/train.py
datumbox: Reminder that this still need to be removed.

references/classification/train.py
```diff
@@ -229,12 +229,16 @@ def main(args):
...
+    parameters = torchvision.ops._utils.set_weight_decay(
+        model,
+        args.weight_decay,
+        norm_weight_decay=args.norm_weight_decay,
+        bias_weight_decay=args.bias_weight_decay,
```
datumbox, line 239: Bias was removed if I understand correctly, so this won't work.

torchvision/ops/_utils.py
```diff
@@ -61,3 +61,66 @@ def split_normalization_params(
...
             other_params.extend(p for p in module.parameters() if p.requires_grad)
     return norm_params, other_params
+
+
+def set_weight_decay(
+    model: torch.nn.Module,
+    weight_decay: float,
+    norm_weight_decay: Optional[float] = None,
+    norm_classes: Optional[List[type]] = None,
+    custom_keys_weight_decay: Optional[List[Tuple[str, float]]] = None,
```
datumbox, line 71: Any specific reason why we use List of tuples here instead of a `Dict[str, float]`?

references/classification/train.py
xiaohu2015: yes, bias can be handled in the  `custom_keys`, eg: 
`
custom_keys_weight_decay = [("bias", 0.0)]
`

torchvision/ops/_utils.py
xiaohu2015: we want the order of List to define the priority

torchvision/ops/_utils.py
xiaohu2015: yes, it maybe a trouble, maybe we should define the type of custom keys (type 0: base_name, type 1: full_name, type2: other case (use `in`)

xiaohu2015: if fact, we can limit the user only use type0 and  type1, so we have the trouble to use `in`

datumbox: How about the following? We should always use equalities instead of `in`. When `key` contains `.` it should be matched against the `full_name` else it needs to be matched against the `name`. We can support only 2 types.

xiaohu2015: it is good. we don't have to define the type.

datumbox:
@xiaohu2015  LGTM, thanks for the awesome contributions as always!

As discussed offline, this approach is a great step towards a generic solution. There are still some cons such as the fact that the custom_key logic is applied always prior the norm_layers but we can investigate alternative more flexible solutions on separate PRs. Since more investigation will be done, we can move this back to references after all the tests on this PR pass.

datumbox:
LGTM and the test pass. We can move it to references and merge. :)

references/classification/train.py
```diff
@@ -393,6 +399,18 @@ def get_args_parser(add_help=True):
...
         help="weight decay for Normalization layers (default: None, same value as --wd)",
     )
+    parser.add_argument(
+        "--bias-weight-decay",
+        default=None,
+        type=float,
+        help="weight decay for bias parameters of all layers (default: None, same value as --wd)",
+    )
+    parser.add_argument(
+        "--transformer-weight-decay",
```
datumbox, line 409: @jdsgomes Any thoughts on this name?

This option controls the weight decay for parameters related to position encodings and embeddings that are usually found in Transformer models. We will eventually revise this utility, so this is temporary to unblock the work on Swin but I was hoping if you could offer a better alternative. Line 236 on this PR will also offer clarifications over what this does.

datumbox (approved):
@xiaohu2015 Thanks! LGTM!

FYI I fixed a conflict with master. I'll wait for the input from Joao but I'm happy to make for you inplace any name changes to avoid back and forth.

references/classification/train.py
jdsgomes: The only alternative I can think of is to introduce a generic weight decay argument with pairs of ([keys], weight decay) and then deprecate bias_weight_decay.

datumbox: Yeap, true. Another idea that @xiaohu2015 suggested was to pass directly the custom_keys_weight_decay from argparse. The reason that for now we avoided this was because it exposes too much info on the internals of the utility. An alternative approach for the util is to accept lambdas that allow users to select any module they want on arbitrary conditions. We discussed offline to explore similar approaches and see if we can create something generic to be added in `ops` eventually. But in the meantime, I didn't want to unblock your work on Swin and that's why we went with this workaround.

I wonder if there is at least a better naming convention for what this flag does. Do you think it would be accurate to name it `--positional-param-decay` or something else?

jdsgomes: Because of the class_token I would not  call it `positional-param-decay` so I would keep the original one for lack of better alternative.

datumbox: After an online chat with @jdsgomes, we decided to rename to `--transformer-embedding-decay`. This will do for now to unblock Swin and we can review soon.

PR #5675, opened 2022-03-25
port Grayccale and RandomGrayscale transforms to prototype API

pmeier:
For these two transformations, the new `ConvertColorSpace` transformation also needs the ability to return grayscale images with three channels.

### DIFF
--- a/torchvision/prototype/transforms/_legacy.py
+++ b/torchvision/prototype/transforms/_legacy.py
@@ -13,29 +13,52 @@ from ._transform import _RandomApplyTransform
 
 class Grayscale(Transform):
     def __init__(self, num_output_channels: Literal[1, 3] = 1) -> None:
-        warnings.warn(
-            "The transform `Grayscale(num_output_channels=...)` is deprecated and will be removed in a future release. "
-            "Please use "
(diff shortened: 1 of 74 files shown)


torchvision/prototype/transforms/_legacy.py
```diff
@@ -0,0 +1,41 @@
...
+        self.num_output_channels = num_output_channels
+        self._rgb_to_gray = ConvertImageColorSpace(
+            color_space=ColorSpace.GRAY, old_color_space=ColorSpace.RGB, gray_output_channels=num_output_channels
+        )
+
+    def _transform(self, input: Any, params: Dict[str, Any]) -> Any:
+        return self._rgb_to_gray(input)
+
+
+class RandomGrayscale(_RandomApplyTransform):
```
pmeier, line 33: Not sure if we want to deprecate here, since the randomness is not covered by `ConvertImageColorSpace`. The replacement would be

```py
transforms.RandomApply(
    transforms.ConvertImageColorSpace(
        color_space=ColorSpace.GRAY,
        old_color_space=ColorSpace.RGB,
        gray_output_channels=3,
    ),
    p=0.1,
)
```

datumbox (approved):
LGTM, thanks!

PR #5677, opened 2022-03-25
Fixed rotate with expand inconsistency

vfdev-5:
Description:
- Fixed rotate with expand inconsistency between torch vs PIL on odd-sized images
- Updated tests
- Rewritten bmm -> matmul

This PR fixes output size inconsistency of odd-sized images.

Before:
```python
import torch
import torchvision
from torchvision.transforms.functional import rotate

print(torchvision.__version__)

img = torch.ones(3, 7, 33, dtype=torch.uint8)

out = rotate(img, angle=67, expand=True)
print(out.shape[1:][::-1])

from PIL import Image

pil_img = Image.new("RGB", (33, 7))
pil_out = rotate(pil_img, angle=67, expand=True)
print(pil_out.size)
> 
0.13.0a0+1af20e8
torch.Size([20, 34])
(21, 35)
```

Now:
```python
> 
0.13.0a0+050a5c9
torch.Size([21, 35])
(21, 35)
```

### DIFF
--- a/test/test_functional_tensor.py
+++ b/test/test_functional_tensor.py
@@ -67,7 +67,7 @@ class TestRotate:
     IMG_W = 26
 
     @pytest.mark.parametrize("device", cpu_and_gpu())
-    @pytest.mark.parametrize("height, width", [(26, IMG_W), (32, IMG_W)])
+    @pytest.mark.parametrize("height, width", [(7, 33), (26, IMG_W), (32, IMG_W)])
     @pytest.mark.parametrize(
         "center",
         [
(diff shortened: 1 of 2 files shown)


datumbox:
@vfdev-5 could you elaborate on what inconsistency are you fixing? It's worth adding some details on the PR. Should this be considered a bug fix?

Also why do we change the tests? Is it to provide better corner cases? Or the previous test inputs would fail if you keep them the same on this PR?

vfdev-5:
@datumbox thanks for asking for clarifications ! I updated PR's description.

The issue is with output size which is different for torch image and PIL image if input has odd size (e.g. 7, 33). Current tests are not using odd sizes thus CI does not complain.
Context: in the prototype tests, @pmeier is using odd sizes for tests and segmentation mask tests were failing on this case and it appears that the roots are here.

datumbox (approved):
LGTM, thanks for the work and the clarifications.

Please tag the PR accordingly before merging.

PR #5679, opened 2022-03-25
Add places365 datapipe

YosuaMichael:
Related issues: https://github.com/pytorch/vision/issues/5347

Note: I have tried with the validation and test data, also with a mock train data. However I have not tried with the real train data (currently runnning and it takes time)


torchvision/prototype/datasets/_builtin/places365.py
```diff
@@ -0,0 +1,139 @@
...
+
+    _BASE_URL = "http://data.csail.mit.edu/places/places365/"
+
+    def _make_info(self) -> DatasetInfo:
+        return DatasetInfo(
+            name="places365",
+            homepage="http://places2.csail.mit.edu/index.html",
+            valid_options=dict(
+                variant=("standard", "challenge"),
+                image_res=("large", "256"),
```
pmeier, line 69: Can we call this "small"?

```suggestion
                image_res=("large", "small"),
```

torchvision/prototype/datasets/_builtin/places365.py
```diff
@@ -0,0 +1,139 @@
...
+
+        # Join images_dp with labels_dp
+        images_labels_dp = IterKeyZipper(
+            images_dp,
+            labels_dp,
+            key_fn=functools.partial(self._image_key, config),
+            ref_key_fn=getitem(0),
+            buffer_size=INFINITE_BUFFER_SIZE,
+        )
+        return Mapper(images_labels_dp, self._prepare_sample)
```
pmeier, line 130: 1. We canonically shuffle and shard the label datapipe and also use it as reference for zipping. This way, we don't need to touch the `images_dp` at all.
2. The final datapipe is canonically called `dp`

```suggestion
        labels_dp = CSVParser(labels_dp, delimiter=" ")
        labels_dp = hint_sharding(labels_dp)
        labels_dp = hint_shuffling(labels_dp)

        dp = IterKeyZipper(
            labels_dp,
            images_dp,
            key_fn=getitem(0),
            ref_key_fn=functools.partial(self._image_key, config),
            buffer_size=INFINITE_BUFFER_SIZE,
        )
        return Mapper(dp, self._prepare_sample)
```

torchvision/prototype/datasets/_builtin/places365.py
pmeier: Better yet, maybe make it a bool flag?

```suggestion
                large=(True, False),
```

torchvision/prototype/datasets/_builtin/places365.py
```diff
@@ -0,0 +1,139 @@
...
+
+        variant = config.variant
+        if config.split != "train":
+            variant = "_"
+        img_filename, img_sha256 = self._IMAGES_MAP[(variant, config.image_res, config.split)]
+        images = HttpResource(urljoin(self._BASE_URL, img_filename), sha256=img_sha256)
+
+        return [meta, images]
+
+    def _image_key(self, config: DatasetConfig, data: Tuple[str, Any]) -> str:
```
pmeier, line 86: 1. We prefer to make anything not coming from the datapipe as keyword only argument to imply that this has to specified through `functools.partial`.
2. Given that you only need `image_res` here, I would prefer to only pass this to keep the dependencies low.

```suggestion
    def _image_key(self, data: Tuple[str, Any], *, image_res: str) -> str:
```

torchvision/prototype/datasets/_builtin/places365.py
```diff
@@ -0,0 +1,139 @@
...
+            root_dir = f"data_{config.image_res}"
+            idx = list(reversed(path.parts)).index(root_dir) - 1
+            return f"/{path.relative_to(path.parents[idx]).as_posix()}"
+        else:
+            return path.name
+
+    def _prepare_sample(self, data: Tuple[Tuple[str, Any], List[str]]) -> Dict[str, Any]:
+        (path, buffer), key_label = data
+        if len(key_label) == 1:
+            label = None
```
pmeier, line 98: Could you add a comment that this only happens for the test split?

torchvision/prototype/datasets/_builtin/places365.py
```diff
@@ -0,0 +1,139 @@
...
+        else:
+            label = Label(int(key_label[1]), categories=self.categories)
```
pmeier, line 100: Have you checked if the labels provided by the dataset are zero-indexed? Otherwise we need to subtract one here, since `self.categories` is zero-indexed.

test/builtin_dataset_mocks.py
```diff
@@ -1438,6 +1438,97 @@ def stanford_cars(info, root, config):
     return num_samples
 
 
+@register_mock
+def places365(info, root, config):
+    num_samples = {"train": 5, "test": 4, "val": 3}[config.split]
+    categories = ["alley", "coffee_shop", "field/wild", "subway_station/platform"]
```
pmeier, line 1444: Could we use the last category here to catch potential issues with zero / one based indexing of the labels?

```suggestion
    categories = ["alley", "field/wild", "subway_station/platform", "zen_garden"]
```

torchvision/prototype/datasets/_builtin/places365.py
YosuaMichael: sure, will do

torchvision/prototype/datasets/_builtin/places365.py
YosuaMichael: regarding this, if I remember correctly we want the `valid_options` to have `str` as values? In this case, perhaps using `large` and `small` might be better compared to bool flag?

pmeier: True, I said that. Yeah, I'm not as consistent as I would like to be. Thanks for pointing that out. No strong opinion then.

torchvision/prototype/datasets/_builtin/places365.py
```diff
@@ -0,0 +1,143 @@
...
+    _BASE_URL = "http://data.csail.mit.edu/places/places365/"
+
+    def _make_info(self) -> DatasetInfo:
+        return DatasetInfo(
+            name="places365",
+            homepage="http://places2.csail.mit.edu/index.html",
+            valid_options=dict(
+                variant=("standard", "challenge"),
+                image_res=("large", "small"),
+                split=("train", "val", "test"),
```
pmeier, line 70: Given that there is only a `challenge` variant for `split=train`, should we just make it a special split?

```suggestion
                image_res=("large", "small"),
                split=("train", "train-challenge", "val", "test"),
```

That should also simplify the resources.

torchvision/prototype/datasets/_builtin/places365.py
YosuaMichael: Hi @pmeier I have tried this but found a surprising decrease in performance. I measure the performance with tqdm comparing the **old** method (images_dp as key joined with labels_dp as reference) and the **new** method you proposed (labels_dp as key joined with images_dp as reference).  Here is the measurement:

### old method:
10000it [01:37, 102.37it/s]
10000it [01:35, 104.66it/s]
100000it [01:56, 860.18it/s]
100000it [01:58, 843.96it/s]

I assume that the processing time follow the formula: total_time = startup_time + num_iteration * processing_speed
In this way, I found out what is the startup_time (preprocessing_time) and the processing_speed:

-  processing_speed = 21s / (100000 - 10000) = 21s / 90000 ~ 2.33 * 10^-4 s/it (4285 it / s)
-  startup_time = 96s - 10000 * processing_speed_1 = 93.66 s


### new method:
10000it [13:07, 12.70it/s]
10000it [13:06, 12.71it/s]
100000it [14:34, 114.33it/s]
100000it [14:15, 116.84it/s]

- processing_speed = 77s / 90000 = 8.56 * 10^-4 s/it (1168 it / s)
- startup_time = 787.5s - 10000 * processing_speed_2 ~ 778.94 s ~ 12m 59s

### Analysis:
I try to find out why this happen, and I think it is because `IterKeyZipper` will put the reference into memory ( as explained [here](https://pytorch.org/data/main/generated/torchdata.datapipes.iter.IterKeyZipper.html) ) hence in the new method case it will put the `images_dp` into memory and this process takes more time. And this also tell that not just the speed, but the memory usage for the new method also significantly higher.

That being said, I am not too sure whether the `sharding` and `shuffling` will take the `images_dp` into memory or not if we actually implement it. If yes, then the new method should be okay I think. Do you have any thought on this?

torchvision/prototype/datasets/_builtin/places365.py
YosuaMichael: The standard also have `split=train` according to their website: http://places2.csail.mit.edu/download.html

torchvision/prototype/datasets/_builtin/places365.py
pmeier: Could you rerun your tests with

```py
import time
import itertools

tic = time.perf_counter()

dataset = itertools.islice(datasets.load("places365").map(transforms.DecodeImage()), 1_000)
next(dataset)

tac = time.perf_counter()

num_samples = 1
for sample in dataset:
    num_samples += 1

toc = time.perf_counter()

print(f"warmup: {tac - tic:.2f} seconds")
print(f"iter: {num_samples / (toc - tac):.1f} samples per second")
```

That should give a clearer picture. If you no longer see a significant difference, remove the `.map(transforms.DecodeImage())` call above and retry.

> That being said, I am not too sure whether the `sharding` and `shuffling` will take the `images_dp` into memory or not if we actually implement it. If yes, then the new method should be okay I think. Do you have any thought on this?

Yes, shuffling will load anything into memory anyways. Still, I think it is time well spent to investigate the performance difference, since I didn't expected it at all. Maybe something is off that also affects the other datasets.

torchvision/prototype/datasets/_builtin/places365.py
YosuaMichael: ah okay, got what you mean now! Indeed the test and val for both `standard` and `challenge` are the same.
Ok, will change this to your suggestion then, it will simplify the `valid_options`, thanks!

torchvision/prototype/datasets/_builtin/places365.py
YosuaMichael: I have tried to run your code (with DecodeImage), here are the result:
```
old method:
>>> print(f"warmup: {tac - tic:.2f} seconds")
warmup: 10.91 seconds
>>> print(f"iter: {num_samples / (toc - tac):.1f} samples per second")
iter: 334.0 samples per second

new method:
>>> print(f"warmup: {tac - tic:.2f} seconds")
warmup: 100.11 seconds
>>> print(f"iter: {num_samples / (toc - tac):.1f} samples per second")
iter: 88.4 samples per second
```

YosuaMichael: I checked on train, challenge, and large because this is the biggest dataset (and have the biggest differences).

I think the main differences is because the IterKeyZipper will iterate through the main dp while putting the references into memory, hence putting the images_dp will take significantly more memory (and lookup time) compared to having the labels_dp on memory.

pmeier: I can confirm that my proposal is slower. The slow down stems from the fact that the files inside the image tar are not ordered whereas the the ids in the split file in ascending order. For example for `split="test"` we need to pull ~74% of files from the archive before we can yield the first sample. That is a lot slower than drawing an image and loading a single file into memory for lookup.

Drawing more samples than the `1_000` used above will gradually remove the difference since at some point all images are in the buffer and thus the approaches have the same speed. Still, the significantly increased warm-up time is a good reason to go against the informal way of doing things.

YosuaMichael: Agree that having more samples will eventually make the warmup time insignificant differences. Will implement on the standard way.

pmeier (approved):
Thanks a lot @YosuaMichael!

PR #5680, opened 2022-03-25
enforce shuffling before sharding

pmeier:
As discussed in today's sync, this PR places the shuffling before the sharding. Plus, this adds a test that enforces this in the future.

### DIFF
--- a/test/test_prototype_builtin_datasets.py
+++ b/test/test_prototype_builtin_datasets.py
@@ -120,8 +120,7 @@ class TestCommon:
         pickle.dumps(dataset)
 
     @parametrize_dataset_mocks(DATASET_MOCKS)
-    @pytest.mark.parametrize("annotation_dp_type", (Shuffler, ShardingFilter))
-    def test_has_annotations(self, test_home, dataset_mock, config, annotation_dp_type):
+    def test_shuffling_and_sharding(self, test_home, dataset_mock, config):
         def scan(graph):
             for node, sub_graph in graph.items():
                 yield node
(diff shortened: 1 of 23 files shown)


test/test_prototype_builtin_datasets.py
```diff
@@ -131,8 +130,31 @@ def scan(graph):
...
+            )
+
+        for node in nodes:
+            if isinstance(node, Shuffler):
+                break
+        else:
+            raise AssertionError(
+                "Found no `Shuffler` in the graph. "
+                "Please place a `dp = hint_shuffling(dp)` call in the dataset implementation. "
+            )
```
pmeier, line 157: @NivekT @ejguan @VitalyFedyunin This or similar logic can probably reused for the warning we potentially wanted to emit in the data loader.

torchvision/prototype/datasets/_builtin/dtd.py
```diff
@@ -98,7 +90,7 @@ def _make_datapipe(
 
         splits_dp = Filter(splits_dp, path_comparator("name", f"{config.split}{config.fold}.txt"))
         splits_dp = LineReader(splits_dp, decode=True, return_path=False)
-        splits_dp = Shuffler(splits_dp, buffer_size=INFINITE_BUFFER_SIZE)
+        splits_dp = hint_shuffling(splits_dp)
```
pmeier, line 93: This was unintentionally already correct. This was only detected, because the test did not fail without me fixing the dataset. Things like this will be prevented in the future.

NicolasHug (commented):
Thanks for the PR @pmeier 

I'm happy with the changes to the dataset code but regarding the test, I'd prefer waiting for a solution on the torchdata side before we implement our own

pmeier:
I've reverted the test and left a TODO comment.

NicolasHug (approved):
Thanks @pmeier

PR #5683, opened 2022-03-26
[WIP] Port SBU dataset

### LINKED ISSUE #5349: SBU 1

lezwon:
fixes #5349


lezwon:
@pmeier mind checking this PR and letting me know if I'm headed in the right direction?

torchvision/prototype/datasets/_builtin/sbu.py
```diff
@@ -0,0 +1,77 @@
...
+        photo_urls_dp = HttpReader(photo_urls_dp)
+
+        photo_captions_dp = LineReader(photo_captions_dp, decode=True, return_path=False)
+
+        dp = Zipper(photo_urls_dp, photo_captions_dp)
+        dp = Filter(dp, lambda data: self._filter_by_content_type(data[0][1]))
+        dp = hint_shuffling(dp)
+        dp = hint_sharding(dp)
+
+        return Mapper(dp, self._prepare_sample)
```
pmeier, line 61: This just adapts your implementation to downloaded images. Since there might be missing images after download, we can no longer zip and filter like you did. My idea to solve this is to an `IterKeyZipper` (conditionally zip two datapipes) with the images as reference. As zip key we can use the name of the image which is encoded in that path as well as the URL.

```suggestion
        images_dp, urls_dp, captions_dp = Demultiplexer(
            resource_dps[0], 3, self._classify_files, drop_none=True, buffer_size=INFINITE_BUFFER_SIZE
        )

        images_dp = hint_shuffling(images_dp)
        images_dp = hint_sharding(images_dp)

        urls_dp = LineReader(urls_dp, decode=True, return_path=False)
        captions_dp = LineReader(captions_dp, decode=True, return_path=False)
        anns_dp = Zipper(urls_dp, captions_dp)

        dp = IterKeyZipper(images_dp, anns_dp, path_accessor("name"), buffer_size=INFINITE_BUFFER_SIZE)
        return Mapper(dp, self._prepare_sample)
```

torchvision/prototype/datasets/_builtin/sbu.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def _make_datapipe(
+        self,
+        resource_dps: List[IterDataPipe],
+        *,
+        config: DatasetConfig,
+    ) -> IterDataPipe[Dict[str, Any]]:
+
+        photo_urls_dp, photo_captions_dp = Demultiplexer(
+            resource_dps[0], 2, self._classify_files, drop_none=True, buffer_size=-1
```
pmeier, line 48: We prefer to use `from torchvision.prototype.datasets.utils._internal import INIFINITE_BUFFER_SIZE`. Unfortunately, for now some datapipes use `-1` for an infinite buffer, others `None` and some don't even support it at all. Thus, we just use a really large number until this is fixed on `torchdata` side.

torchvision/prototype/datasets/_builtin/sbu.py
```diff
@@ -0,0 +1,77 @@
...
+        )
+
+        photo_urls_dp = LineReader(photo_urls_dp, decode=True, return_path=False)
+        photo_urls_dp = HttpReader(photo_urls_dp)
+
+        photo_captions_dp = LineReader(photo_captions_dp, decode=True, return_path=False)
+
+        dp = Zipper(photo_urls_dp, photo_captions_dp)
+        dp = Filter(dp, lambda data: self._filter_by_content_type(data[0][1]))
```
pmeier, line 57: `lambda`'s and local functions are currently not allowed. The data loader that performs some optimizations on the datapipe graph tries to `pickle` the datapipe and this fails for the aforementioned types.

torchvision/prototype/datasets/_builtin/sbu.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def _prepare_sample(self, sample: Dict[str, Any]) -> Dict[str, Any]:
+        image, caption = sample
+        _, image = image
+
+        # TODO: check for method to convert to tensor without PIL and Numpy
+        image = pil_image.open(io.BytesIO(image.data))
+        image = np.array(image)
+
+        return {"image": features.Image(image), "caption": caption}
```
pmeier, line 77: I think you are looking for `torchvision.prototype.features.EncodedImage` with its `.from_file` method.

```suggestion
    def _prepare_sample(self, data: Tuple[Tuple[str, BinaryIO], Tuple[str, str]]) -> Dict[str, Any]:
        (path, buffer), (_, caption) = data
        return dict(
            path=path,
            image=EncodedImage.from_file(buffer),
            caption=caption.strip(),
        )
```

pmeier (commented):
Argh, I knew there was a reason I marked SBU in the tracker issue. When I gave you the go I just saw that there is something to download and assumed I was mistaken before. That is on me, my bad.

In general your solution looks really good and works, but has one major downside: We need to re-download every image for every iteration. While we plan to support streaming datasets from the internet, we are not there yet. Thus, we need to download everything.

That could be achieved with a [`OnDiskCacheHolder`](https://github.com/pytorch/data/blob/2f66c15deeadebfd10d4893553e9ba8d8de52190/torchdata/datapipes/iter/util/cacheholder.py#L114), but that would mean we would only download at runtime. All current datasets download everything upfront and I would keep it that way for now.

My solution is to put a custom `preprocess` method onto the resource and download everything there.

torchvision/prototype/datasets/_builtin/sbu.py
```diff
@@ -0,0 +1,77 @@
...
+            name="sbu",
+            homepage="http://www.cs.virginia.edu/~vicente/sbucaptions/",
+        )
+
+    def resources(self, config: DatasetConfig) -> List[OnlineResource]:
+        return [
+            HttpResource(
+                "http://www.cs.virginia.edu/~vicente/sbucaptions/SBUCaptionedPhotoDataset.tar.gz",
+                sha256="2bf37d5e1c9e1c6eae7d5103030d58a7f2117fc5e8c6aa9620f0df165acebf09",
+            )
```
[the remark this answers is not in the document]
lezwon: @pmeier Can I use `_urlretrieve` directly instead of `HttpResource.download` to download the files? That way I can skip the prints.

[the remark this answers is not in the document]
pmeier: You could, but this brings other problems. `HttpResource.download()` handles redirects whereas `_urlretrieve` doesn't. This is not a problem you need to handle though. Maybe we can leave points 3. and 4. above as `TODO` comment in the code? 1. and 2. should be doable.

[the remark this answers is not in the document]
lezwon: That sounds good. I'll make the changes. :)

test/builtin_dataset_mocks.py
```diff
@@ -1454,3 +1454,24 @@ def usps(info, root, config):
...
+
+    dataset_folder = pathlib.Path(root).joinpath("dataset")
+    dataset_folder.mkdir(parents=True)
+
+    photo_urls_file = "SBU_captioned_photo_dataset_urls.txt"
+    photo_captions_file = "SBU_captioned_photo_dataset_captions.txt"
+    with open(dataset_folder.joinpath(photo_urls_file), "w") as url_file, open(
+        dataset_folder.joinpath(photo_captions_file), "w"
+    ) as caption_file:
+        urls = [f"https://via.placeholder.com/{random.randint(100, 1000)}.jpg" for _ in range(num_samples)]
```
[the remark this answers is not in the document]
lezwon: @pmeier I'll wait for the PR to get merged right? I can make the necessary changes after it.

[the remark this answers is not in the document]
pmeier: Yes, sorry for the delay. I'll try to get it merged soon.

torchvision/prototype/datasets/_builtin/sbu.py
```diff
@@ -0,0 +1,102 @@
...
+    hint_shuffling,
+    INFINITE_BUFFER_SIZE,
+    path_accessor,
+)
+from torchvision.prototype.features import EncodedImage
+
+
+class SBU(Dataset):
+    PHOTO_URLS = 0
+    PHOTO_CAPTIONS = 1
```
NicolasHug, line 19: Are these used?

pmeier: Nope, the should be removed:

```suggestion
```

lezwon: oops..missed this out.

PR #5684, opened 2022-03-26
add positional exception for giou loss

developer0hye:
https://github.com/pytorch/vision/blob/289fce29b3e2392114aadbe7a419df0f2e3ac1be/torchvision/ops/boxes.py#L301-L306


datumbox:
@developer0hye Thanks for the contribution. Unfortunately adding these assertions means a significant slowdown. If you check git blame you will actually see that originally we had assertions and we removed them to improve speed. If you need them then we advice to wrap the loss on your own method and add the assertions there.

developer0hye:
@datumbox 
Thanks for your review!

I also think that it makes **this function** slow. However, I am still curious about that adding this condition makes **whole** training process slow.

PR #5685, opened 2022-03-26
call _upcast to consider overflow for computing giou loss

developer0hye:
giou loss is weak at overflow problem because it computes area of box.

https://github.com/pytorch/vision/blob/main/torchvision/ops/boxes.py#L226-L242

### DIFF
--- a/torchvision/ops/giou_loss.py
+++ b/torchvision/ops/giou_loss.py
@@ -4,7 +4,7 @@ from torch import Tensor
 
 def _upcast(t: Tensor) -> Tensor:
     # Protects from numerical overflows in multiplications by upcasting to the equivalent higher type
-    if not t.dtype in (torch.float32, torch.float64):
+    if t.dtype not in (torch.float32, torch.float64):
         return t.float()
     return t
 
--- a/torchvision/ops/giou_loss.py
+++ b/torchvision/ops/giou_loss.py
@@ -4,10 +4,9 @@ from torch import Tensor
 
 def _upcast(t: Tensor) -> Tensor:
     # Protects from numerical overflows in multiplications by upcasting to the equivalent higher type
-    if t.is_floating_point():
-        return t if t.dtype in (torch.float32, torch.float64) else t.float()
-    else:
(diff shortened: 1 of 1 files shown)


torchvision/ops/giou_loss.py
```diff
@@ -1,4 +1,13 @@
 import torch
+from torch import Tensor
+
+
+def _upcast(t: Tensor) -> Tensor:
+    # Protects from numerical overflows in multiplications by upcasting to the equivalent higher type
+    if t.is_floating_point():
+        return t if t.dtype in (torch.float32, torch.float64) else t.float()
+    else:
+        return t if t.dtype in (torch.int32, torch.int64) else t.int()
```
datumbox, line 10: This method already exists at [boxes.py](https://github.com/pytorch/vision/blob/289fce29b3e2392114aadbe7a419df0f2e3ac1be/torchvision/ops/boxes.py#L218-L223) but I'm not sure we should use it here. This is because it keeps boxes in integer float and that would break giou's `mean()` reduction. Instead we should cast to float if not already float32/float64.

datumbox (commented):
@developer0hye Thanks for the PR. Looks like something we need to address. I've added a comment below for your attention, let me know your thoughts.

torchvision/ops/giou_loss.py
developer0hye: @datumbox 

Thanks for your review!

Should we newly define upcast method for this function like the below code?

```python
if t.dtype not in (torch.float32, torch.float64):
 return t.float()
```

datumbox: A private method that does that would be fine. No strong opinions. :)

developer0hye: I modified it!

https://github.com/pytorch/vision/blob/dc2d754410e25557edc233c54b99b9f3247684fb/torchvision/ops/giou_loss.py#L5-L9

datumbox (approved):
LGTM, thanks!

developer0hye:
@datumbox Thanks!

Issue #5686, opened 2022-03-28
Question on segmentation code


kcs6568:
### 🚀 The feature

Hello.
I want to ask you a simple question.
I'm not sure if it's right to post a question in this 'Feature request' category.

In train.py code in the reference/segmentation, the get_dataset function is set the coco dataset classes 21.
Why the number of classes is 21?
Is it wrong to set the number of classes to 91 which is the number of classes in the coco dataset?

Here is the reference code.
```python
def get_dataset(dir_path, name, image_set, transform):
    def sbd(*args, **kwargs):
        return torchvision.datasets.SBDataset(*args, mode="segmentation", **kwargs)

    paths = {
        "voc": (dir_path, torchvision.datasets.VOCSegmentation, 21),
        "voc_aug": (dir_path, sbd, 21),
        "coco": (dir_path, get_coco, 21),
    }
    p, ds_fn, num_classes = paths[name]

    ds = ds_fn(p, image_set=image_set, transforms=transform)
    return ds, num_classes

cc @vfdev-5 @datumbox @YosuaMichael

**Comment by pmeier** [comment]
> Why the number of classes is 21?

To align it with the VOC categories. The COCO dataset uses a subset that has a 1-to-1 correspondence with VOC categories:

https://github.com/pytorch/vision/blob/da03f51a0c94917b6a905d58974243a2b458e075/references/segmentation/coco_utils.py#L95

```py
from torchvision.prototype import datasets

coco_cats = datasets.info("coco").categories
coco_seg_cats = [coco_cats[idx] for idx in [5, 2, 16, 9, 44, 6, 3, 17, 62, 21, 67, 18, 19, 4, 1, 64, 20, 63, 7, 72]]

voc_cats = datasets.info("voc").categories

for coco_cat, voc_cat in zip(coco_seg_cats, voc_cats):
    print(coco_cat, voc_cat)
```

```
airplane aeroplane
bicycle bicycle
bird bird
boat boat
bottle bottle
bus bus
car car
cat cat
chair chair
cow cow
dining table diningtable
dog dog
horse horse
motorcycle motorbike
person person
potted plant pottedplant
sheep sheep
couch sofa
train train
tv tvmonitor
```

> Is it wrong to set the number of classes to 91 which is the number of classes in the coco dataset?

No it is not wrong, it just makes harder to compare the performance to models trained on VOC. If you don't need that, you can also train with all 91 categories. Note that you will also need to adapt / remove the transforms that currently create the subset of categories.

PR #5687, opened 2022-03-28
add background category to prototype VOC dataset

pmeier:
While checking #5686, I realized that the prototype VOC dataset is missing background category that is needed for segmentation. While fixing that, I also fixed the broken category file generation as well as using an enum for the demultiplexing.


NicolasHug (approved):
Thanks @pmeier , approving to unblock, although I think this might require a few merge conflicts resolution because of the recent refactoring of the dataset code

PR #5689, opened 2022-03-28
use enum class method for integer interpolation modes

pmeier:
https://github.com/pytorch/vision/blob/96aecd2d65c892a2c17ab2be2f1deae774380606/torchvision/transforms/functional.py#L36-L37

Seems to work now:

```py
import torch


from torchvision.transforms import functional as F, InterpolationMode


eager = F.resize
scripted = torch.jit.script(eager)

args = (torch.rand(3, 256, 256), [512, 128])
interpolation = 2

torch.testing.assert_close(
    eager(*args, interpolation=interpolation),
    scripted(*args, interpolation=InterpolationMode.from_legacy_int(interpolation)),
)
```

Note that we already cannot call the scripted function with an `int`, since we annotated `interpolation: InterpolationMode`:

```py
@torch.jit.script
def foo(bar: InterpolationMode):
    if isinstance(bar, int):
        baz = F._interpolation_modes_from_int(bar)
    else:
        baz = bar
    return baz


foo(1)
```

```
RuntimeError: foo() Expected a value of type 'Enum<__torch__.torchvision.transforms.functional.InterpolationMode>' for argument 'bar' but instead found type 'int'.
```

### DIFF
--- a/torchvision/transforms/functional.py
+++ b/torchvision/transforms/functional.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import math
 import numbers
 import warnings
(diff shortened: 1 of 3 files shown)


torchvision/transforms/functional.py
```diff
@@ -32,19 +34,16 @@ class InterpolationMode(Enum):
...
-        0: InterpolationMode.NEAREST,
-        2: InterpolationMode.BILINEAR,
-        3: InterpolationMode.BICUBIC,
-        4: InterpolationMode.BOX,
-        5: InterpolationMode.HAMMING,
-        1: InterpolationMode.LANCZOS,
-    }
-    return inverse_modes_mapping[i]
+    @classmethod
+    def from_legacy_int(cls, legacy_int: int) -> InterpolationMode:
```
NicolasHug, line 38: If the int behaviour is legacy, should this be private?

NicolasHug (approved):
Thanks @pmeier , This LGTM (modulo green CI) but I have a concern about whether this should be private?

I'll approve to unblock anyway

NicolasHug (changes requested):
(Flagging with a request for changes as you asked offline)

pmeier:
It seems this is not supported:

```py
@torch.jit.script
def rotate(
    resample: Optional[int] = None, interpolation: InterpolationMode = InterpolationMode.BILINEAR
) -> InterpolationMode:
    if resample is not None:
        interpolation = InterpolationMode.from_legacy_int(resample)
    if isinstance(interpolation, int):
        interpolation = InterpolationMode.from_legacy_int(interpolation)
    return interpolation
```

```
RuntimeError: 
Enum<__torch__.torchvision.transforms.functional.InterpolationMode>' has no attribute 'from_legacy_int':
```

I'm not sure why this was not triggered in the snippet in my top comment:

```py
@torch.jit.script
def foo(bar: InterpolationMode):
    if isinstance(bar, int):
        baz = F._interpolation_modes_from_int(bar)
    else:
        baz = bar
    return baz
```

My best guess is that since `bar` is annotated as `InterpolationMode`, torchscript completely skips the `if isinstance(bar, int)` branch and thus never hits the failure.

Issue #5514, opened 2022-03-02
[RFC] Implement transforms primitives for Bounding Boxes


vfdev-5:
Related tracker issue: https://github.com/pytorch/vision/issues/5754

## Functional transforms for bboxes

The idea is to implement the following low-level functional transforms for bounding boxes:
https://github.com/pytorch/vision/blob/main/torchvision/prototype/transforms/functional/_geometry.py
- [x] affine_bounding_box (@vfdev-5) #5597
- [x] rotate_bounding_box (@vfdev-5) #5638
- [x] perspective_bounding_box (@vfdev-5) #5888
- [x] center_crop_bounding_box (@vfdev-5) #5972
- [x] crop_bounding_box (@vfdev-5) #5781
- [x] resized_crop_bounding_box (@vfdev-5) #5853
- [x] pad_bounding_box (@pmeier) #5551, #6038
- [x] vertical_flip_bounding_box #5633

## How to test the implementation ?

Put tests into `test/test_prototype_transforms_functional.py`:
- correctness test
- eager vs scripted
- scriptable
- check on cpu and cuda

cc @vfdev-5 @datumbox @bjuncek

vfdev-5:
@pmeier can you provide tests for `pad_bounding_box` such that we could close this tracker ? Thanks

pmeier:
Thanks for reminding me!

vfdev-5:
I'm closing this tracker as done

PR #5692, opened 2022-03-28
[proto] Added functional `rotate_segmentation_mask` op

vfdev-5:
Related to #5514

## Description:
- Added functional `rotate_segmentation_mask` op
- Added tests

### Results on synthetic images/bboxes/segm mask:

<details>
<summary>
Code
</summary>

```python
import numpy as np

import torch
import torchvision
from torchvision.prototype.transforms.functional import rotate_bounding_box, rotate_image_tensor, rotate_segmentation_mask

size = (64, 72)
# xyxy format
in_boxes = [
    [10, 15, 25, 35],
    [50, 5, 70, 22],
    [45, 46, 56, 62],
]

for in_box in in_boxes:
    im1[in_box[1]:in_box[3], in_box[0]:in_box[2], :] = (127, 127, 127)
    
)
    
angle = 22
center = None
expand = True

out_boxes = rotate_bounding_box(
    in_boxes, 
    in_boxes.format,
    in_boxes.image_size,
    angle,
    expand=expand,
    center=center
)
print(out_boxes)

out_mask = rotate_segmentation_mask(
    in_mask, 
    angle,
    expand=expand,
    center=center    
)

t_im2 = rotate_image_tensor(t_im1, angle, expand=expand, center=center)


plt.figure(figsize=(14, 10))

plt.subplot(2,3,1)
plt.title("Input image + bboxes")
for in_box in in_boxes:    
plt.imshow(r1)


plt.subplot(2,3,2)
plt.title("Input segm mask")
plt.imshow(in_mask[0, :, :].cpu().numpy())


plt.subplot(2,3,3)
plt.title("Input image + bboxes + segm mask")
plt.imshow(r1, alpha=0.5)
plt.imshow(in_mask[0, :, :].cpu().numpy(), alpha=0.75)


plt.subplot(2,3,4)
plt.title("Output image + bboxes")
for out_box in out_boxes:
plt.imshow(r2)


plt.subplot(2,3,5)
plt.title("Output segm mask")
plt.imshow(out_mask[0, :, :].cpu().numpy())

plt.subplot(2,3,6)
plt.title("Output image + bboxes + segm mask")
plt.imshow(r2, alpha=0.5)
plt.imshow(out_mask[0, :, :].cpu().numpy(), alpha=0.75)
```
</details>

image


test/test_prototype_transforms_functional.py
```diff
@@ -411,8 +435,9 @@ def _compute_expected_bbox(bbox, angle_, translate_, scale_, shear_, center_):
             center=center,
         )
 
-        if center is None:
-            center = [s // 2 for s in bboxes_image_size[::-1]]
+        center_ = center
```
vfdev-5, line 438: Fixing a bug with in previous tests

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -361,6 +361,10 @@ def rotate_bounding_box(
     expand: bool = False,
     center: Optional[List[float]] = None,
 ) -> torch.Tensor:
+    if center is not None and expand:
+        warnings.warn("The provided center argument is ignored if expand is True")
+        center = None
```
pmeier, line 366: Given that both are not the default values, should we maybe fail if someone actually does this?

vfdev-5: It is not an error, that is why I raise a warning...

pmeier: It is not an error because we chose to ignore user input. If someone actually does `rotate_bounding_box(..., expand=True, center=[...])` I think we should just fail. Is there any use case for allowing both to be set?

vfdev-5: We can always imagine a use-case like data aug policy search where `expand` and `center` are parameters. Neither PIL nor D2 does not error in such case. I'm not sure why we should be that strict.

pmeier: > We can always imagine a use-case like data aug policy search where `expand` and `center` are parameters.

We should definitely not emit warnings from code where the user has no influence on. Imagine doing

```py
transform = transforms.NewAutoAug()
transform(image)
```

and getting a warning that the parameter `center` is ignored. Thus, internally `NewAutoAug` needs to do something like `center=center if not expand else None` anyway. 

> Neither PIL nor D2 does not error in such case. I'm not sure why we should be that strict.

1. What is D2?
2. Do they warn in this case?

vfdev-5: D2 = detectron2. 

> Do they warn in this case?

No, both are silent. Only docs states the behaviour.

pmeier: I just tried with PIL and it seems `center` has an effect on the output with `expand=True`

```py
image = PIL.Image.open("test/assets/encode_jpeg/grace_hopper_517x606.jpg")

image.rotate(30.0, center=(100, 100), expand=True).show()
image.rotate(30.0, center=(200, 200), expand=True).show()
```

vfdev-5: Yes, I should have mentioned that PIL modifies image but it is incorrect behaviour, IMO.

According to their docs:

> :param expand: Optional expansion flag.  If true, expands the output
>           image to make it large enough to hold the entire rotated image.
>           If false or omitted, make the output image the same size as the
>           input image.  Note that the expand flag assumes rotation around
>           the center and no translation.

https://github.com/python-pillow/Pillow/blob/a6efaa1ae1d0bc83bf1cbbe7170f7734a7f4203d/src/PIL/Image.py#L2122-L2123

datumbox: TorchVision has historically had a mixed approach on that. Sometimes we throw an error, sometimes we overwrite the value and emit a warning. Lately on models I started throwing errors when a parameter is incompatible. Might be worth aligning all places in TorchVision on follow up PRs but I'll leave it to you to decide how to handle this in this case; no strong opinions.

datumbox (approved):
LGTM, thanks!

torchvision/prototype/transforms/functional/_geometry.py
pmeier: Maybe we can rephrase the warning to make it more clear what is happening? The passed `center` is not ignored, but rather has no effect on the result if `expand=True`. This makes it more clear that we are not ignoring user input, but rather warn the user that the input combination makes little sense since it doesn't have an effect.

PR #5701, opened 2022-03-29
keep imagenet prototype as class but factor out methods

pmeier:
In pytorch/data#237 we found that it shouldn't be a fundamental issue with the design that would keep us from using methods as inputs for datapipes defined on the same object. Still, this PR factors out these methods into standalone functions to see if there are more blockers down the road.


pmeier:
Superseded by #5778.

Issue #4615, opened 2021-10-14
WIDERFace download fails with zip error


NicolasHug:
Originally from https://github.com/pytorch/vision/pull/4614#issuecomment-943468223

```py
In [4]: torchvision.datasets.WIDERFace(root='/tmp/lol', split='train', download=True)
```

```
91473it [00:00, 9168293.30it/s]
---------------------------------------------------------------------------
BadZipFile                                Traceback (most recent call last)
<ipython-input-4-61c4acdeef4e> in <module>
----> 1 torchvision.datasets.WIDERFace(root='/tmp/lol', split='train', download=True)

~/dev/vision/torchvision/datasets/widerface.py in __init__(self, root, split, transform, target_transform, download)
     70
     71         if download:
---> 72             self.download()
     73
     74         if not self._check_integrity():

~/dev/vision/torchvision/datasets/widerface.py in download(self)
    191             download_file_from_google_drive(file_id, self.root, filename, md5)
    192             filepath = os.path.join(self.root, filename)
--> 193             extract_archive(filepath)
    194
    195         # download and extract annotation files

~/dev/vision/torchvision/datasets/utils.py in extract_archive(from_path, to_path, remove_finished)
    407     extractor = _ARCHIVE_EXTRACTORS[archive_type]
    408
--> 409     extractor(from_path, to_path, compression)
    410
    411     return to_path

~/dev/vision/torchvision/datasets/utils.py in _extract_zip(from_path, to_path, compression)
    281
    282 def _extract_zip(from_path: str, to_path: str, compression: Optional[str]) -> None:
--> 283     with zipfile.ZipFile(
    284         from_path, "r", compression=_ZIP_COMPRESSION_MAP[compression] if compression else zipfile.ZIP_STORED
    285     ) as zip:

~/opt/miniconda3/envs/pt/lib/python3.8/zipfile.py in __init__(self, file, mode, compression, allowZip64, compresslevel, strict_timestamps)
   1267         try:
   1268             if mode == 'r':
-> 1269                 self._RealGetContents()
   1270             elif mode in ('w', 'x'):
   1271                 # set the modified flag so central directory gets written

~/opt/miniconda3/envs/pt/lib/python3.8/zipfile.py in _RealGetContents(self)
   1334             raise BadZipFile("File is not a zip file")
   1335         if not endrec:
-> 1336             raise BadZipFile("File is not a zip file")
   1337         if self.debug > 1:
   1338             print(endrec)

[...]

Issue #3708, opened 2021-04-22
Unable to load CelebA dataset: "File is not zip file" error.


lucaslingle:
## 🐛 Bug

I am unable to download the CelebFace Attributes (CelebA) dataset using TorchVision. The error I get when attempting to download is "File is not a zip file." 

I have encountered this error every time I've tried to download the file, over the course of the last 24 hours. 

## To Reproduce

1. Import torchvision
2. Call torchvision.datasets.CelebA(root='somewhere', split='train', download=True)

A code sample is provided below:

```
import torch as tc
import torchvision as tv

def get_dataloaders(batch_size):
    transform = tv.transforms.Compose([
        tv.transforms.CenterCrop(108),
        tv.transforms.Resize(64),
        tv.transforms.ToTensor(),
        tv.transforms.Normalize((0.5, 0.5, 0.5), (0.5, 0.5, 0.5))  # scales pixels to [-1, 1].
    ])
    train_data = tv.datasets.CelebA(root='data', split='train', download=True, transform=transform)
    test_data = tv.datasets.CelebA(root='data', split='test', download=True, transform=transform)

    train_dataloader = tc.utils.data.DataLoader(train_data, batch_size=batch_size, shuffle=True)
    test_dataloader = tc.utils.data.DataLoader(test_data, batch_size=batch_size, shuffle=True)

    return train_dataloader, test_dataloader
```

Here, my stacktrace is provided as well:

```
/Users/lucaslingle/opt/miniconda3/envs/pytorch181/bin/python /Users/lucaslingle/PycharmProjects/pytorch_dcgan/main.py
3112it [00:00, 2727831.57it/s]
26721026it [00:01, 23197510.91it/s]
3424458it [00:00, 12672950.45it/s]
6082035it [00:00, 17442168.31it/s]
12156055it [00:00, 21823850.99it/s]
2836386it [00:00, 11141828.55it/s]
Traceback (most recent call last):
  File "/Users/lucaslingle/PycharmProjects/pytorch_dcgan/main.py", line 35, in <module>
    dataloader, _ = get_dataloaders(batch_size=batch_size)
  File "/Users/lucaslingle/PycharmProjects/pytorch_dcgan/dataloaders.py", line 11, in get_dataloaders
    train_data = tv.datasets.CelebA(root='data', split='train', download=True, transform=transform)
[...]

Issue #2756, opened 2020-10-05
Bad zipfile error while downloading CelebA dataset


square-1111:
## 🐛 Bug

I am working with CelebA dataset with `download = True`. I am getting an error 
```BadZipFile: File is not a zip file```

## To Reproduce  
```
import torchvision.datasets as dset
data = dset.celeba.CelebA(root='.', split='train', target_type='attr', transform=None, download=True)
```

## Environment
I am working on Google Colab
```
PyTorch version: 1.6.0+cu101
Is debug build: False
CUDA used to build PyTorch: 10.1
ROCM used to build PyTorch: N/A

OS: Ubuntu 18.04.5 LTS (x86_64)
GCC version: (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0
Clang version: 6.0.0-1ubuntu2 (tags/RELEASE_600/final)
CMake version: version 3.12.0

Python version: 3.6 (64-bit runtime)
Is CUDA available: True
CUDA runtime version: 10.1.243
GPU models and configuration: GPU 0: Tesla K80
Nvidia driver version: 418.67
cuDNN version: /usr/lib/x86_64-linux-gnu/libcudnn.so.7.6.5
HIP runtime version: N/A
MIOpen runtime version: N/A  

Versions of relevant libraries:
[pip3] numpy==1.18.5 
[pip3] torch==1.6.0+cu101
[pip3] torchsummary==1.5.1
[pip3] torchtext==0.3.1
[pip3] torchvision==0.7.0+cu101
[conda] Could not collect```

cc @pmeier

fmassa:
This is a duplicate of https://github.com/pytorch/vision/issues/2387 and https://github.com/pytorch/vision/issues/2317, and a "fix" is done in https://github.com/pytorch/vision/pull/2321

The issue is that CelebA is stored in Google Drive, and it has exceeded the download quota for the day.
Try again in a different moment and it should work

square-1111:
Thank You for the leads.

square-1111:
`AttributeError: 'tqdm' object has no attribute 'close'` -- I am getting this error on local machine. Is this error associated with exceeding download quota or something else?

vfdev-5:
@square-1111 could you please provide a minimal code snippet and list of installed packages to reproduce the issue. Thanks

fmassa:
This error is different -- can you check if you have tqdm installed? If not, might be that the helper implementation of tqdm that we provide doesn't have the close method

[...]

PR #5704, opened 2022-03-30
improve error handling for GDrive downloads

pmeier:
We have plenty of reports that the download of datasets hosted on GDrive does not work as expected:

- #1920
- #2262
- #2317
- #2756
- #3708
- #4615
- #5615
- #5716
- #5800

Although most of them are closed, they see another comment from time to time since a user stumbled upon "the same issue"

The errors non-descriptive in most cases. The problem is that we don't check the MD5 sum after the download and naively write the response from GDrive to disk. In contrast, on `download_url` we perform such a check

https://github.com/pytorch/vision/blob/aa21197462591a89f527ded3938ffe62f7d8cb8f/torchvision/datasets/utils.py#L150-L152

The most common failure case is GDrive returning an unknown API response as HTML. This PR adds a MD5 check after the download with an additional HTML check to make the error message more descriptive.


torchvision/datasets/utils.py
```diff
@@ -257,6 +257,27 @@ def download_file_from_google_drive(file_id: str, root: str, filename: Optional[
...
+        msg = f"The MD5 checksum of the download file {fpath} does not match the one on record."
+        remediation_msg = (
+            "Please delete the file and try again. "
+            "If the issue persists, please report this to torchvision at https://github.com/pytorch/vision/issues."
+        )
+        # GDrive API responses should be smaller than 10kB
+        if os.stat(fpath).st_size < 10 * 1024:
+            with open(fpath) as fh:
+                content = fh.read()
+                # Regular expression to detect HTML. Copied from https://stackoverflow.com/a/70585604
```
NicolasHug, line 270: thank you for writing this comment

NicolasHug (commented):
Thanks @pmeier . 

> The most common failure case is GDrive returning an unknown API response as HTML

Shouldn't we check the HTML code of the download / reply from GDrive instead of downloading then?

pmeier:
> Shouldn't we check the HTML code of the download / reply from GDrive instead of downloading then?

Unfortunately, GDrive mostly just returns `200` and thus there is very little we can do without actually inspecting the response. Maybe we can improve this by always analyzing the first chunk of data rather than only if we detect an MD5 mismatch. Thoughts?

NicolasHug:
> analyzing the first chunk of data rather than only if we detect an MD5 mismatch. Thoughts

If you mean checking against the HTML regex right after the files are downloaded, then yeah I feel like this might be where the check needs to be

NicolasHug (approved):
Thanks Philip

PR #5707, opened 2022-03-30
add test to enforce infinite buffer size for all applicable datapipes

pmeier:
Given that is not acceptable for our datasets to drop any sample, we need to use an infinite buffer size in all datapipes that have a variable sized buffer. This PR adds a test that enforces this.

Currently we use 

https://github.com/pytorch/vision/blob/1db8795733b91cd6dd62a0baa7ecbae6790542bc/torchvision/prototype/datasets/utils/_internal.py#L42-L43

but that hopefully changes after pytorch/data#335 is resolved.

### DIFF
--- a/test/test_prototype_builtin_datasets.py
+++ b/test/test_prototype_builtin_datasets.py
@@ -8,6 +8,7 @@ import torch
 from builtin_dataset_mocks import parametrize_dataset_mocks, DATASET_MOCKS
 from torch.testing._comparison import assert_equal, TensorLikePair, ObjectPair
 from torch.utils.data.graph import traverse
+from torch.utils.data.graph_settings import get_all_graph_pipes
 from torchdata.datapipes.iter import (
     IterDataPipe,
     Shuffler,
@@ -29,6 +30,10 @@ assert_samples_equal = functools.partial(
 )
 
 
+def extract_datapipes(dp):
+    return get_all_graph_pipes(traverse(dp, only_datapipe=True))
+
+
 @pytest.fixture
 def test_home(mocker, tmp_path):
     mocker.patch("torchvision.prototype.datasets._api.home", return_value=str(tmp_path))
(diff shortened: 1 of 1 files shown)


test/test_prototype_builtin_datasets.py
```diff
@@ -148,6 +163,28 @@ def test_save_load(self, test_home, dataset_mock, config):
             buffer.seek(0)
             assert_samples_equal(torch.load(buffer), sample)
 
+    @parametrize_dataset_mocks(DATASET_MOCKS)
+    def test_infinite_buffer_size(self, test_home, dataset_mock, config):
+        dataset_mock.prepare(test_home, config)
+        dataset = datasets.load(dataset_mock.name, **config)
+
+        for dp in extract_datapipes(dataset):
+            if isinstance(
```
NicolasHug, line 172: Should we just check the existence of the `buffer_size` attribute instead? Otherwise we'll have to update this list for all classes that have a buffer

test/test_prototype_builtin_datasets.py
```diff
@@ -35,6 +45,15 @@ def test_coverage():
         )
 
 
+def extract_datapipes(dp):
```
NicolasHug, line 48: Is this the same as https://github.com/pytorch/pytorch/blob/40bb880009e86831d5186f619b33ad390d702b16/torch/utils/data/graph_settings.py#L5 ?

NicolasHug (approved):
Thanks @pmeier , approving to unblock but I have minor comments

test/test_prototype_builtin_datasets.py
```diff
@@ -169,21 +155,14 @@ def test_infinite_buffer_size(self, test_home, dataset_mock, config):
...
-                    IterKeyZipper,
-                ),
-            ):
-                # TODO: replace this with the proper sentinel as soon as https://github.com/pytorch/data/issues/335 is
-                #  resolved
-                assert dp.buffer_size == INFINITE_BUFFER_SIZE
+            try:
+                buffer_size = getattr(dp, "buffer_size")
+            except AttributeError:
+                continue
```
NicolasHug, line 161: why not `if hasattr`?

pmeier: > The first snippet you wrote is also an early return right?

Yeah, but I prefer the other style.

> I won't fight for this but the code we have right now is literally re-implementing the builtin [hasattr](https://docs.python.org/3/library/functions.html#hasattr)

I agree, my solution was an overkill. This is why I already fixed it in 7e36fd9.

PR #5712, opened 2022-03-31
upgrade to next-gen CircleCI convenience images

pmeier:
All workflows that currently run on a `circleci/*` docker image issue this warning:

circleci_warning

[Example workflow](https://app.circleci.com/pipelines/github/pytorch/vision/16209/workflows/3cb5da6c-40b6-43bb-8d81-1aed43597927/jobs/1313281).

The warning points to [this thread](https://discuss.circleci.com/t/legacy-convenience-image-deprecation/41034/3). TL;DR: we should to migrate to `cimg/*` images as the `circleci/*` images will no longer receive support starting from 2023. Otherwise nothing should change :crossed_fingers:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -266,7 +266,7 @@ smoke_test_common: &smoke_test_common
 jobs:
   circleci_consistency:
     docker:
-      - image: circleci/python:3.7
+      - image: cimg/python:3.7
     steps:
       - checkout
       - pip_install:
@@ -279,7 +279,7 @@ jobs:
 
   lint_python_and_config:
     docker:
-      - image: circleci/python:3.7
+      - image: cimg/python:3.7
     steps:
       - checkout
       - pip_install:
@@ -298,7 +298,7 @@ jobs:
 
   lint_c:
     docker:
-      - image: circleci/python:3.7
+      - image: cimg/python:3.7
     steps:
       - apt_install:
           args: libtinfo5
@@ -320,7 +320,7 @@ jobs:
 
   type_check_python:
     docker:
-      - image: circleci/python:3.7
+      - image: cimg/python:3.7
     steps:
       - apt_install:
           args: libturbojpeg-dev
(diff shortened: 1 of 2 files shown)


pmeier:
It seems the tagging scheme for Android images has changed from the API version, e.g. `circleci/android:api-29-ndk` to a time-based one, e.g. `cimg/android:2022.03-ndk`. I couldn't find an official "conversion table" or something similar. 

We are currently using API version 29, which was [published in September 2019](https://en.wikipedia.org/wiki/Android_version_history#Android_10) and got its [latest update in July 2020](https://developer.android.com/studio/releases/platforms#10). The [earliest next-gen image](https://hub.docker.com/layers/android/cimg/android/2021.08-ndk/images/sha256-3f4d74183e996ed6dc84103b0be8bc7c62ac988ffe8793acb3b32b451e2466ec) is from August 2021. 

If I'm not mistaken, the next-gen image installs the [SDK for all API versions from 23 to 31](https://github.com/CircleCI-Public/cimg-android/blob/7af5c262821e4bf68c19cdb0ed34283165684e11/2021.08/Dockerfile#L51-L59) whereas the legacy only installs for [API version 29](https://github.com/CircleCI-Public/circleci-dockerfiles/blob/eda702fe560b258138d3a895c5d397f4a2cdfcd0/android/images/api-29-ndk/Dockerfile#L202). 

Someone with Android experience should have a look to see if this will affect our workflows. In the meantime, I'll try the latest next-gen convenience image to see if CI is happy or not.

datumbox (approved):
LGTM, thanks!

pmeier:
The support [clarified](https://discuss.circleci.com/t/legacy-convenience-image-deprecation/41034#android-7):

> The tagging scheme has changed from API level based to date based. This now means that multiple API versions are available at the same time. Any one Android image tag will have at least the 5 most recent API levels.

IIUC, we should really specify the API version we are compiling against. As explained in https://github.com/pytorch/vision/pull/5712#issuecomment-1084241643, the new image also has API version 30 and 31 installed. If we don't specify, probably the latest version is used, which means that we are now building against 31 instead of 29. The `2022.3` image includes two later versions and it seems to fail without the pin.

Issue #5708, opened 2022-03-30
Add the SWAG pre-trained weights in TorchVision


datumbox:
### 🚀 The feature

It would be great to add support of the pre-trained weights from the Supervised Weakly from hashtAGs ([SWAG](https://github.com/facebookresearch/SWAG)) to TorchVision. There are two sets of weights to add for each model variant:
1. The original frozen trunk SWAG weights with a linear classifier learnt on ImageNet1K
2. The end-to-end fine-tuned weights on ImageNet1K

We will focus on the variants that are currently supported by TorchVision (`regnet_y_16gf`, `regnet_y_32gf`, `vit_b_16` and `vit_l_16`). We should also investigate if the larger variants can be added (`regnet_y_128gf` and `vit_h_14`) or if they cause issues on our CI (memory, increased execution times etc).

This task includes the following subtasks:
- Convert the weights to be compatible with TorchVision's implementation
- Add the [weight entries](https://github.com/pytorch/vision/blob/main/torchvision/models/regnet.py#L584-L594) with the right transform configuration and meta-data
- Add the necessary licensing info (name, URL etc) in the meta-data; update the README to clarify they are offered under [CC-BY-NC 4.0](https://github.com/facebookresearch/SWAG/blob/main/LICENSE) 
- Verify that the accuracies reported by our reference scripts match the ones reports on the SWAG repo
- Confirm that our CI works well and the additions don't bring significant slowdowns or breakages. If there are such effects, take actions to mitigate

PR #5714, opened 2022-03-31
Add SWAG Vision Transformer Weight

YosuaMichael:
subtask of https://github.com/pytorch/vision/issues/5708

Adding weight of ViT_B_16 and ViT_L_16 model from https://github.com/facebookresearch/SWAG

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -185,3 +185,10 @@ Disclaimer on Datasets
 This is a utility library that downloads and prepares public datasets. We do not host or distribute these datasets, vouch for their quality or fairness, or claim that you have license to use the dataset. It is your responsibility to determine whether you have permission to use the dataset under the dataset's license.
 
 If you're a dataset owner and wish to update any part of it (description, citation, etc.), or do not want your dataset to be included in this library, please get in touch through a GitHub issue. Thanks for your contribution to the ML community!
+
+Disclaimer on Models
+====================
+
+Pretrained models provided in this library may have their own license or term and condition to use. It is your responsibility to determine whether you have permission to use the model for your use case.
+
+More specifically, SWAG models are released under the CC-BY-NC 4.0 license. See `SWAG LICENSE <https://github.com/facebookresearch/SWAG/blob/main/LICENSE>`_ for additional details.
--- a/README.rst
+++ b/README.rst
@@ -186,9 +186,9 @@ This is a utility library that downloads and prepares public datasets. We do not
 
 If you're a dataset owner and wish to update any part of it (description, citation, etc.), or do not want your dataset to be included in this library, please get in touch through a GitHub issue. Thanks for your contribution to the ML community!
 
-Disclaimer on Models
-====================
+Pre-trained Model License
+=========================
 
-Pretrained models provided in this library may have their own license or term and condition to use. It is your responsibility to determine whether you have permission to use the model for your use case.
+Pre-trained models provided in this library may have their own license or terms and conditions to use. It is your responsibility to determine whether you have permission to use the model for your use case.
 
 More specifically, SWAG models are released under the CC-BY-NC 4.0 license. See `SWAG LICENSE <https://github.com/facebookresearch/SWAG/blob/main/LICENSE>`_ for additional details.
--- a/README.rst
+++ b/README.rst
@@ -189,6 +189,6 @@ If you're a dataset owner and wish to update any part of it (description, citati
 Pre-trained Model License
 =========================
 
-Pre-trained models provided in this library may have their own license or terms and conditions to use. It is your responsibility to determine whether you have permission to use the model for your use case.
+The pre-trained models provided in this library may have their own licenses or terms and conditions derived from the dataset used for training. It is your responsibility to determine whether you have permission to use the models for your use case.
 
 More specifically, SWAG models are released under the CC-BY-NC 4.0 license. See `SWAG LICENSE <https://github.com/facebookresearch/SWAG/blob/main/LICENSE>`_ for additional details.
--- a/torchvision/models/vision_transformer.py
+++ b/torchvision/models/vision_transformer.py
@@ -325,11 +325,8 @@ _COMMON_META = {
 }
 
 _COMMON_SWAG_META = {
-    "task": "image_classification",
-    "architecture": "ViT",
-    "publication_year": 2022,
+    **COMMON_META,
     "recipe": "https://github.com/facebookresearch/SWAG",
-    "categories": _IMAGENET_CATEGORIES,
     "interpolation": InterpolationMode.BICUBIC,
 }
 
@@ -358,8 +355,8 @@ class ViT_B_16_Weights(WeightsEnum):
             "num_params": 86859496,
             "size": (384, 384),
             "min_size": (384, 384),
-            "acc@1": 85.29,
-            "acc@5": 97.65,
+            "acc@1": 85.304,
+            "acc@5": 97.650,
         },
     )
     DEFAULT = IMAGENET1K_V1
--- a/torchvision/models/vision_transformer.py
+++ b/torchvision/models/vision_transformer.py
@@ -328,6 +328,7 @@ _COMMON_SWAG_META = {
     **_COMMON_META,
     "publication_year": 2022,
     "recipe": "https://github.com/facebookresearch/SWAG",
+    "license": "https://github.com/facebookresearch/SWAG/blob/main/LICENSE",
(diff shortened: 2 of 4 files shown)


torchvision/models/vision_transformer.py
```diff
@@ -284,7 +285,15 @@ def _vision_transformer(
...
-    image_size = kwargs.pop("image_size", 224)
+
+    image_size = None
+    if "image_size" in kwargs:
+        image_size = kwargs.pop("image_size", None)
+    if image_size is None and weights is not None and "size" in weights.meta:
+        image_size = weights.meta["size"]
+        if isinstance(image_size, abc.Sequence):
+            image_size = image_size[0]
+    image_size = image_size or 224
```
YosuaMichael, line 296: The weights for SWAG vit_b_16 tied to `image_size=384`. I edited this part so the users can just call:
```
m = vit_b_16(ViT_B_16_Weights.IMAGENET1K_SWAG_V1)
```
instead of
```
m = vit_b_16(ViT_B_16_Weights.IMAGENET1K_SWAG_V1, image_size=384)
```
However, there might be a better way to handle this. Any suggestions are welcome :)

datumbox: That's a good approach. I would only refactor slightly to simplify (use if/elif/else). Other than that, your approach looks good.

torchvision/models/vision_transformer.py
```diff
@@ -313,6 +322,15 @@ def _vision_transformer(
     "interpolation": InterpolationMode.BILINEAR,
 }
 
+_COMMON_SWAG_META = {
+    "task": "image_classification",
+    "architecture": "ViT",
+    "publication_year": 2022,
+    "recipe": "https://github.com/facebookresearch/SWAG",
+    "categories": _IMAGENET_CATEGORIES,
+    "interpolation": InterpolationMode.BICUBIC,
```
datumbox, line 331: Nit: Should we share some of the parameters from common and overwrite them here?

Something like:
```python
_COMMON_SWAG_META = {
    **_COMMON_META,
    "publication_year": 2022,
    "recipe": "https://github.com/facebookresearch/SWAG",
 }
```

We also need to confirm what the interpolation that they used is.

torchvision/models/vision_transformer.py
```diff
@@ -328,6 +346,20 @@ class ViT_B_16_Weights(WeightsEnum):
             "acc@5": 95.318,
         },
     )
+    IMAGENET1K_SWAG_V1 = Weights(
+        url="https://download.pytorch.org/models/vit_b_16_swag-9ac1b537.pth",
+        transforms=partial(
+            ImageClassification, resize_size=384, interpolation=InterpolationMode.BICUBIC, crop_size=384
```
datumbox, line 352: We need to confirm this config with the team.

torchvision/models/vision_transformer.py
```diff
@@ -328,6 +346,20 @@ class ViT_B_16_Weights(WeightsEnum):
...
+        ),
+        meta={
+            **_COMMON_SWAG_META,
+            "num_params": 86859496,
+            "size": (384, 384),
+            "min_size": (384, 384),
+            "acc@1": 85.29,
+            "acc@5": 97.65,
```
datumbox, line 360: We should put here the stats from a `total_batch_size=1` run. We usually do 3 digit accuracy.

torchvision/models/vision_transformer.py
```diff
@@ -328,6 +346,20 @@ class ViT_B_16_Weights(WeightsEnum):
...
         },
     )
+    IMAGENET1K_SWAG_V1 = Weights(
+        url="https://download.pytorch.org/models/vit_b_16_swag-9ac1b537.pth",
+        transforms=partial(
+            ImageClassification, resize_size=384, interpolation=InterpolationMode.BICUBIC, crop_size=384
+        ),
+        meta={
+            **_COMMON_SWAG_META,
+            "num_params": 86859496,
```
datumbox, line 356: I assume this is different from the one at `IMAGENET1K_V1` because of the size of the input. Our tests currently make the assumption that all weights have the same exactly number of parameters, this is true for 99% of the models (actually this one is the first that has this issue). So we will need to update the test logic to account for deliberate differences like this. What do you propose?

datumbox (commented):
Thanks @YosuaMichael. I've added a couple of comments but overall I like your approach:

torchvision/models/vision_transformer.py
YosuaMichael: oh, didn't know we can overwrite like that! Good point!

For the interpolation, I got it from their eval method: https://github.com/facebookresearch/SWAG/blob/main/imagenet_1k_eval.py#L198

and to me it seems like they are using `BICUBIC`.

torchvision/models/vision_transformer.py
YosuaMichael: I think we can do as you said in the chat:

> The specific line on the test is written this way to avoid slowdowns related to downloading and loading the model. I think it's a good trick for 99% of the models but when it fails we should check explicitly if the number of parameters match the meta-data. Aka do the same as line: https://github.com/pytorch/vision/blob/main/test/test_extended_models.py#L115

So if the fast check fails, a more thorough check is done prior marking it as incorrect. What you think?

will add a commit for this

torchvision/models/vision_transformer.py
```diff
@@ -284,10 +285,20 @@ def _vision_transformer(
...
 ) -> VisionTransformer:
-    image_size = kwargs.pop("image_size", 224)
-
     if weights is not None:
         _ovewrite_named_param(kwargs, "num_classes", len(weights.meta["categories"]))
+        if "size" in weights.meta:
+            if isinstance(weights.meta["size"], int):
+                _ovewrite_named_param(kwargs, "image_size", weights.meta["size"])
+            elif isinstance(weights.meta["size"], abc.Sequence):
+                torch._assert(
```
datumbox, line 294: FYI: Here you can use Exceptions because it executed during the construction of the model. The issue is when Exceptions appear in the forward. Up to you if you want to keep it as is.

torchvision/models/vision_transformer.py
```diff
@@ -313,6 +324,12 @@ def _vision_transformer(
     "interpolation": InterpolationMode.BILINEAR,
 }
 
+_COMMON_SWAG_META = {
+    **COMMON_META,
+    "recipe": "https://github.com/facebookresearch/SWAG",
```
datumbox, line 329: "publication_year: 2022,

datumbox (commented):
Looking good. Could you provide also the commands that you used to verify the model along with their output? Something similar to https://github.com/pytorch/vision/pull/5450#issuecomment-1053560434 (note that the commands will be slightly different because the Multi-weight support API is out of prototype).

README.rst
```diff
@@ -186,9 +186,9 @@ This is a utility library that downloads and prepares public datasets. We do not
 
 If you're a dataset owner and wish to update any part of it (description, citation, etc.), or do not want your dataset to be included in this library, please get in touch through a GitHub issue. Thanks for your contribution to the ML community!
 
-Disclaimer on Models
-====================
+Pre-trained Model License
+=========================
 
-Pretrained models provided in this library may have their own license or term and condition to use. It is your responsibility to determine whether you have permission to use the model for your use case.
+Pre-trained models provided in this library may have their own license or terms and conditions to use. It is your responsibility to determine whether you have permission to use the model for your use case.
```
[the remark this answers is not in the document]
YosuaMichael: sure, will add the command here

YosuaMichael:
@datumbox here is the command I used for testing the `vit_b_16_swag` model:
```
python -u ~/script/run_with_submitit.py --timeout 3000 --ngpus 1 --nodes 1 --partition train --model vit_l_16 --data-path="/datasets01_ontap/imagenet_full_size/061417" --test-only --batch-size=1 --weights="ViT_L_16_Weights.IMAGENET1K_SWAG_V1"
```

datumbox (approved):
Thanks a lot of the changes @YosuaMichael, everything looks good to me.

@mannatsingh Could you have also a look to let us know if all look good to you?

mannatsingh (approved):
Everything looks perfect, thanks @YosuaMichael !

datumbox:
The [failures](https://app.circleci.com/pipelines/github/pytorch/vision/16280/workflows/9a63e50e-845b-4c9a-b6c4-29df878ed591/jobs/1319486) are unrelated. cc @pmeier 

@YosuaMichael you are good to merge.

PR #5715, opened 2022-03-31
Detection recipe enhancements

datumbox:
Related to #5307 

This PR cherrypicks some of the improvements from #5444 to simplify review.

Improve the existing detection recipes:
- Add support of LSJ and multiscale augmentation strategies
- Add `--opt` and `--norm-weight-decay` support in Detection
- Add support of `InstanceNorm` and `LocalResponseNorm` layers in `split_normalization_params`
- Other minor nit fixes across references

### DIFF
--- a/references/classification/train.py
+++ b/references/classification/train.py
@@ -230,7 +230,7 @@ def main(args):
     criterion = nn.CrossEntropyLoss(label_smoothing=args.label_smoothing)
 
     if args.norm_weight_decay is None:
-        parameters = model.parameters()
+        parameters = [p for p in model.parameters() if p.requires_grad]
     else:
         param_groups = torchvision.ops._utils.split_normalization_params(model)
         wd_groups = [args.norm_weight_decay, args.weight_decay]
--- a/references/detection/presets.py
+++ b/references/detection/presets.py
@@ -3,10 +3,31 @@ import transforms as T
 
 
 class DetectionPresetTrain:
-    def __init__(self, data_augmentation, hflip_prob=0.5, mean=(123.0, 117.0, 104.0)):
+    def __init__(self, *, data_augmentation, hflip_prob=0.5, mean=(123.0, 117.0, 104.0)):
         if data_augmentation == "hflip":
             self.transforms = T.Compose(
                 [
+                    T.RandomHorizontalFlip(p=hflip_prob),
+                    T.PILToTensor(),
+                    T.ConvertImageDtype(torch.float),
+                ]
+            )
+        elif data_augmentation == "lsj":
+            self.transforms = T.Compose(
+                [
+                    T.ScaleJitter(target_size=(1024, 1024)),
+                    T.FixedSizeCrop(size=(1024, 1024), fill=mean),
+                    T.RandomHorizontalFlip(p=hflip_prob),
+                    T.PILToTensor(),
+                    T.ConvertImageDtype(torch.float),
+                ]
+            )
+        elif data_augmentation == "multiscale":
+            self.transforms = T.Compose(
+                [
+                    T.RandomShortestSize(
+                        min_size=(480, 512, 544, 576, 608, 640, 672, 704, 736, 768, 800), max_size=1333
+                    ),
                     T.RandomHorizontalFlip(p=hflip_prob),
                     T.PILToTensor(),
                     T.ConvertImageDtype(torch.float),
--- a/references/detection/train.py
+++ b/references/detection/train.py
@@ -209,7 +209,13 @@ def main(args):
 
     opt_name = args.opt.lower()
     if opt_name.startswith("sgd"):
-        optimizer = torch.optim.SGD(parameters, lr=args.lr, momentum=args.momentum, weight_decay=args.weight_decay)
+        optimizer = torch.optim.SGD(
+            parameters,
+            lr=args.lr,
+            momentum=args.momentum,
+            weight_decay=args.weight_decay,
+            nesterov="nesterov" in opt_name,
+        )
     elif opt_name == "adamw":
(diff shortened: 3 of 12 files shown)


references/classification/train.py
```diff
@@ -230,7 +230,7 @@ def main(args):
     criterion = nn.CrossEntropyLoss(label_smoothing=args.label_smoothing)
 
     if args.norm_weight_decay is None:
-        parameters = model.parameters()
+        parameters = [p for p in model.parameters() if p.requires_grad]
```
datumbox, line 233: Nit fix to align reference scripts. This is not strictly needed in Classification since we don't freeze weights but that's more generic.

references/detection/presets.py
```diff
@@ -3,7 +3,7 @@
 
 
 class DetectionPresetTrain:
-    def __init__(self, data_augmentation, hflip_prob=0.5, mean=(123.0, 117.0, 104.0)):
+    def __init__(self, *, data_augmentation, hflip_prob=0.5, mean=(123.0, 117.0, 104.0)):
```
datumbox, line 6: Keyword args are enforced in all presets but this was missed during Multi-weight support porting.

references/detection/presets.py
```diff
@@ -12,6 +12,27 @@ def __init__(self, data_augmentation, hflip_prob=0.5, mean=(123.0, 117.0, 104.0)
...
+            self.transforms = T.Compose(
+                [
+                    T.RandomShortestSize(
+                        min_size=(480, 512, 544, 576, 608, 640, 672, 704, 736, 768, 800), max_size=1333
+                    ),
+                    T.RandomHorizontalFlip(p=hflip_prob),
+                    T.PILToTensor(),
+                    T.ConvertImageDtype(torch.float),
+                ]
+            )
```
datumbox, line 35: Two new augmentation strategies. The transforms were already added in previous PRs.

references/detection/train.py
```diff
@@ -84,6 +85,12 @@ def get_args_parser(add_help=True):
         help="weight decay (default: 1e-4)",
         dest="weight_decay",
     )
+    parser.add_argument(
+        "--norm-weight-decay",
+        default=None,
+        type=float,
+        help="weight decay for Normalization layers (default: None, same value as --wd)",
+    )
```
datumbox, line 93: Adding standard params that we support on Classification.

references/detection/train.py
```diff
@@ -176,6 +183,8 @@ def main(args):
 
     print("Creating model")
     kwargs = {"trainable_backbone_layers": args.trainable_backbone_layers}
+    if args.data_augmentation in ["multiscale", "lsj"]:
+        kwargs["_skip_resize"] = True
```
datumbox, line 187: Because Multi-scale and LSJ require to do the resizing outside the model, this is necessary. For now I don't expose the property publicly (we got other params like this), but this is something we should consider removing from the models. See #2263 for a similar request.

torchvision/ops/_utils.py
```diff
@@ -43,7 +43,13 @@ def split_normalization_params(
...
     # Adapted from https://github.com/facebookresearch/ClassyVision/blob/659d7f78/classy_vision/generic/util.py#L501
     if not norm_classes:
-        norm_classes = [nn.modules.batchnorm._BatchNorm, nn.LayerNorm, nn.GroupNorm]
+        norm_classes = [
+            nn.modules.batchnorm._BatchNorm,
+            nn.LayerNorm,
+            nn.GroupNorm,
+            nn.modules.instancenorm._InstanceNorm,
+            nn.LocalResponseNorm,
+        ]
```
datumbox, line 52: Adding a couple of other Norm_layers that exist in PyTorch but were originally missed. We don't really use them in any of the models.

datumbox (commented):
Adding a few comments to clarify what is what.

references/detection/train.py
```diff
@@ -191,8 +200,20 @@ def main(args):
...
+        wd_groups = [args.norm_weight_decay, args.weight_decay]
+        parameters = [{"params": p, "weight_decay": w} for p, w in zip(param_groups, wd_groups) if p]
+
+    opt_name = args.opt.lower()
+    if opt_name.startswith("sgd"):
+        optimizer = torch.optim.SGD(parameters, lr=args.lr, momentum=args.momentum, weight_decay=args.weight_decay)
+    elif opt_name == "adamw":
+        optimizer = torch.optim.AdamW(parameters, lr=args.lr, weight_decay=args.weight_decay)
+    else:
+        raise RuntimeError(f"Invalid optimizer {args.opt}. Only SGD and AdamW are supported.")
```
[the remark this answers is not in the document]
jdsgomes: since is copy-pasted I think it can stay as is and if we want to change we can do in a different PR, but still wonder if there is a reson  why for sgd we use `starts_with` and for adamw we use ==?

test/test_extended_models.py
```diff
@@ -64,7 +64,6 @@ def test_get_weight(name, weight):
 )
 def test_naming_conventions(model_fn):
     weights_enum = _get_model_weights(model_fn)
-    print(weights_enum)
```
jdsgomes, line 67: profit!

jdsgomes (commented):
LGTM!

jdsgomes (approved):

references/detection/train.py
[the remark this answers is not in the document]
datumbox: You are right! I've indeed copy-pasted it from classification but replaced the previous sgd optimizer line. The problem is that the existing recipe didn't contain the nesterov momentum update. I've just updated the file to support it; it's not something I used so far but it's a simple update.

PR #4102, opened 2021-06-23
Add support for more hashing algorithms

adamjstewart:
MD5 is cryptographically insecure. It's likely okay for most datasets as we aren't executing anything, we're just loading images. However, some datasets include scripts that need to be run. For example:

* [LandCover.ai](https://landcover.ai/) includes a `split.py` script to compute the train-test splits

In order to avoid arbitrary code injection in these cases, it's best to use a cryptographically secure hashing algorithm. We may want to update existing datasets to use SHA-* instead of MD5 because why not.

@pmeier @calebrob6

PR #5717, opened 2022-04-01
indicate md5 checksum is not used for security

pmeier:
Fixes https://github.com/pytorch/vision/pull/4102#issuecomment-1085108412. 

This does not change the functionality in any way, but allows the use of our datasets on [FIPS compliant systems](https://en.wikipedia.org/wiki/FIPS_140-2) that use Python >= 3.9.

### DIFF
--- a/torchvision/datasets/utils.py
+++ b/torchvision/datasets/utils.py
@@ -53,6 +53,9 @@ def gen_bar_updater() -> Callable[[int, int, int], None]:
 
 
 def calculate_md5(fpath: str, chunk_size: int = 1024 * 1024) -> str:
+    # Setting the `usedforsecurity` flag does not change anything about the functionality, but indicates that we are
+    # not using the MD5 checksum for cryptography. This enables its usage in restricted environments like FIPS. Without
+    # it torchvision.datasets is unusable in these environments since we perform a MD5 check everywhere.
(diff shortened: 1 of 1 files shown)


NicolasHug (approved):
I'll approve, but I don't undestand how this `usedforsecurity` parameter will prevent any sort of security issue. If our code *was* indeed used for security, there is absolutely nothing that prevents us from setting `usedforsecurity=False`, and downstream FIPS system will happily run it?  I'm not sure what the goal is, but I hope I'm completely missing the point.

As discussed offline, it'd be useful to add a comment with a link to this PR, with more context.

pmeier:
> I'm not sure what the goal is, but I hope I'm completely missing the point.

I don't think you are, but this shouldn't be our concern. I agree, the design of the flag is dubious at best. But we are a good faith actor here. The annotation is correct: we are not using the MD5 checksum for security (read cryptography). Thus, it is on the user to evaluate if they trust the libraries to set this flag correctly. Worst case, the environment flat out forbids MD5 checksum and we are back at the situation we had before. If however the environment accepts the annotation (like FIPS seems to do), we enable users to use `torchvision.datasets` which was not possible for them before.

NicolasHug:
> this shouldn't be our concern

I'm sorry, it is our concern. You and I have spent time on this thing. We're adding maintenance burden with this PR

Issue #5254, opened 2022-01-21
Fix flaky `test_classification_model[cuda-vit_h_14]` test


datumbox:
### 🐛 Describe the bug

After the addition of `vit_h_14` (#5210) we started receiving memory issues in some of our tests. Unfortunately the don't fail consistently but rather occasionally. Examples [1](https://app.circleci.com/pipelines/github/pytorch/vision/14208/workflows/be78ecf7-48b2-4c9a-be99-505930f86e35/jobs/1143437) and [2](https://app.circleci.com/pipelines/github/pytorch/vision/14208/workflows/be78ecf7-48b2-4c9a-be99-505930f86e35/jobs/1143479). 

Sometimes the memory issues are limited to that specific test, while other (see example 2) it takes down with it all other tests. We need to address this and restore our CI to green.

### Versions

Latest main branch - 4bf6c6e4b4cf2dea9c7f9952aa30d5820ab8ae33

PR #5721, opened 2022-04-01
Adding the huge vision transformer from SWAG

YosuaMichael:
Part of https://github.com/pytorch/vision/issues/5708

However we separate the PR from https://github.com/pytorch/vision/pull/5714 because the huge models might break the CI.

The purpose of this PR is just to check whether or not the CI is okay with the huge model.


torchvision/models/vision_transformer.py
```diff
@@ -531,6 +554,30 @@ def vit_l_32(*, weights: Optional[ViT_L_32_Weights] = None, progress: bool = Tru
     )
 
 
+@handle_legacy_interface(weights=("pretrained", ViT_H_14_Weights.IMAGENET1K_SWAG_V1))
```
datumbox, line 557: We don't need the legacy compatibility as this is a new model:
```suggestion
```

datumbox:
As discussed offline, let's measure how much extra time this adds on our CI. We should also check if after adding it, our CI remains stable across runs. We've previously tried to add this version (see #5210) but then was reverted (see #5259) due to breakages at #5254. So if the tests pass now, we should be sure to understand that they are stable and won't break our CI again.

YosuaMichael:
> As discussed offline, let's measure how much extra time this adds on our CI. We should also check if after adding it, our CI remains stable across runs. We've previously tried to add this version (see #5210) but then was reverted (see #5259) due to breakages at #5254. So if the tests pass now, we should be sure to understand that they are stable and won't break our CI again.

@datumbox I have rerun the test 5x and all are successful,  so this indicates that it seems stable now.
I also take note of the additional time the model vit_h_14 introduced:

- test_windows_cpu: add 70.95s
- test_windows_gpu: add 27.27s
- test_extended: add 10.86s

In my opinion this addition is still reasonable and in fact it is quite similar with the existing model regnet_y_128gf in term of the test time.
What do you think about this?

YosuaMichael:
This is the validation script and final output:
```
python -u ~/script/run_with_submitit.py --timeout 3000 --ngpus 1 --nodes 1 --partition train --model vit_h_14 --data-path="/datasets01_ontap/imagenet_full_size/061417" --test-only --batch-size=1 --weights="ViT_H_14_Weights.IMAGENET1K_SWAG_V1"
# Acc@1 88.552 Acc@5 98.694
```

datumbox:
@YosuaMichael Thanks for the analysis. Having tried adding reducing the size of the input to speed up the execution?

We do this for multiple models and even have a list that get their input reduced (you will need to update the expected file):
https://github.com/pytorch/vision/blob/61f82669acb71b45c55b83b025f54062f24445f6/test/test_models.py#L285

Nevertheless, because ViT must be initialized for a specific size, you might need to add a record to the `_model_params` directly. Could you check it out and let us know how much you can reduce the time by passing smaller input size?

test/test_models.py
```diff
@@ -280,6 +280,10 @@ def _check_input_backprop(model, inputs):
         "rpn_pre_nms_top_n_test": 1000,
         "rpn_post_nms_top_n_test": 1000,
     },
+    "vit_h_14": {
+        "image_size": 56,
+        "input_shape": (1, 3, 56, 56),
+    },
```
YosuaMichael, line 286: @datumbox this according to your suggestions on changing the input image_size for the test on vit_h_14 models to speed up. The image_size need to be a multiple of patch_size which is 14, hence we use image_size of 56.

datumbox: Very interesting. Does this mean that the majority of the time is spent of the model initialization or on the JIT-script parsing?

datumbox (approved):
LGTM, thanks!

test/test_models.py
YosuaMichael: @datumbox I did a bit profiling locally and here are the results :
```
[2022-04-04 16:48:41.768663] Before building the model
[2022-04-04 16:48:45.000341] After building model
[2022-04-04 16:48:45.002153] After model.eval().to(device=dev)
[2022-04-04 16:48:45.207033] After doing model(x)
[2022-04-04 16:48:45.208375] After assert expected
[2022-04-04 16:48:45.208385] After assert shape num_classes
[2022-04-04 16:48:50.452526] After check_jit_scripttable
[2022-04-04 16:48:50.667378] After check_fx_compatible
[2022-04-04 16:48:51.256744] Finish
```
Seems like around 40% of the time is on building the model, and another 45% of the time is for `check_jit_scriptable`.

PR #5723, opened 2022-04-01
Add DDP support test for prototype datasets

NicolasHug:
Kinda works, kinda doesn't. Still very WIP.


test/test_prototype_builtin_datasets.py
```diff
@@ -30,6 +32,23 @@ def test_home(mocker, tmp_path):
     yield tmp_path
 
 
+@pytest.fixture
+def ddp_fixture():
+    # Note: we only test DDP with world_size=1, but it should be enough for our purpose.
+    # If we ever need to go full DDP, we'll need to implement a much more complex logic, similar to
+    # MultiProcessTestCase from torch core.
```
vfdev-5, line 39: There is also a way to do that using pytext-xdist and using `--dist=each` option to launch same test multiple times

PR #5731, opened 2022-04-04
[DEBUG] possible pyav DLL conflict

pmeier:
Follow-up to https://github.com/PyAV-Org/PyAV/issues/952#issuecomment-1086588594.


pmeier:
Failure is not on our side as shown in https://github.com/PyAV-Org/PyAV/issues/952#issuecomment-1087184636.

pmeier:
The new binaries are sufficient :tada:

PR #5736, opened 2022-04-05
add resize legacy dispatcher to prototype transforms API

pmeier:
This is the first PR to address the last missing piece of the prototype transforms API: the legacy functional dispatchers. 

There is still some discussion if we maybe want to reinstate the functional dispatchers. Thus, this PR can serve as a basis to see how the API would look like if we didn't do that. 

Todo:

- [ ] proper tests
- [ ] deprecation warning

### DIFF
--- a/.circleci/unittest/linux/scripts/install.sh
+++ b/.circleci/unittest/linux/scripts/install.sh
@@ -31,9 +31,9 @@ esac
 
 printf "Installing PyTorch with %s\n" "${cudatoolkit}"
 if [ "${os}" == "MacOSX" ]; then
-    conda install -y -c "pytorch-${UPLOAD_CHANNEL}" "pytorch-${UPLOAD_CHANNEL}"::pytorch "${cudatoolkit}" pytest
+    conda install -y -c "pytorch-${UPLOAD_CHANNEL}" "pytorch-${UPLOAD_CHANNEL}"::pytorch "${cudatoolkit}" pytest torchdata
 else
-    conda install -y -c "pytorch-${UPLOAD_CHANNEL}" -c nvidia "pytorch-${UPLOAD_CHANNEL}"::pytorch[build="*${version}*"] "${cudatoolkit}" pytest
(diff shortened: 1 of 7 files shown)


.circleci/unittest/linux/scripts/install.sh
```diff
@@ -31,9 +31,9 @@ esac
 
 printf "Installing PyTorch with %s\n" "${cudatoolkit}"
 if [ "${os}" == "MacOSX" ]; then
-    conda install -y -c "pytorch-${UPLOAD_CHANNEL}" "pytorch-${UPLOAD_CHANNEL}"::pytorch "${cudatoolkit}" pytest
```
pmeier, line 34: This is only for demo purposes and needs to be removed before merge

pmeier:
> I'm missing some context on what this PR is aiming at. Is this added to show-case how the current `F` high-level API will be handled when the new API is rolled out?

Correct. As explained above, this is what the API will look like in case we decide to not reinstate the high level dispatchers. If we reinstate them, this PR can be closed. 

> I assume the various FIXME/workarounds are there to force the current tests to validate that the new approach works as expected.

Correct. Since the old tests mostly do not pararmetrize over the functionals, we need to copy-paste almost everything. To avoid this for the discussion above, I opted to monkey patch the old tests to test the new dispatcher instead. This needs to be handled properly before we move on.

datumbox:
@pmeier The PR direction looks good to me.

Does this mean that we don't intend to merge this PR but rather leave it open until the decision about the future of the `F` is determined? If that's the case, I would instead recommend to cherrypick any refactoring (if any) you think is worth keeping now and close the PR. The reason why I'm not in favour of keeping it open is because it already has a bunch of conflicts and that's going to get worse over time. Thoughts?

Let me know if I misunderstood the situation/intention of this.

pmeier:
> The reason why I'm not in favour of keeping it open is because it already has a bunch of conflicts and that's going to get worse over time. Thoughts?

The conflicts are in files that we need to revert anyway, so I'm not really concerned there. In general though I agree with your sentiment. Should we have the discussion about the high level dispatchers soon? If not, is there a "milestone" that we need to reach before we can have it? It is one of the two large open problems with the new API that we need to address eventually.

PR #5738, opened 2022-04-05
fix missing torchdata error message

pmeier:
I missed that in #5672.

### DIFF
--- a/torchvision/prototype/datasets/__init__.py
+++ b/torchvision/prototype/datasets/__init__.py
@@ -1,11 +1,10 @@
 try:
     import torchdata
-except (ModuleNotFoundError, TypeError) as error:
+except ModuleNotFoundError:
     raise ModuleNotFoundError(
         "`torchvision.prototype.datasets` depends on PyTorch's `torchdata` (https://github.com/pytorch/data). "
-        "You can install it with `pip install git+https://github.com/pytorch/data.git`. "
-        "Note that you cannot install it with `pip install torchdata`, since this is another package."
-    ) from error
+        "You can install it with `pip install --pre torchdata --extra-index-url https://download.pytorch.org/whl/nightly/cpu"
+    ) from None
 
 from . import utils
 from ._home import home


torchvision/prototype/datasets/__init__.py
```diff
@@ -1,11 +1,10 @@
 try:
     import torchdata
-except (ModuleNotFoundError, TypeError) as error:
+except ModuleNotFoundError:
```
pmeier, line 3: We can drop the `TypeError` here since that was the error type of the legacy `torchdata` package hosted on PyPI. Since we took over the name, `pip install torchdata` will give you the correct package.

torchvision/prototype/datasets/__init__.py
```diff
@@ -1,11 +1,10 @@
...
     raise ModuleNotFoundError(
         "`torchvision.prototype.datasets` depends on PyTorch's `torchdata` (https://github.com/pytorch/data). "
-        "You can install it with `pip install git+https://github.com/pytorch/data.git`. "
-        "Note that you cannot install it with `pip install torchdata`, since this is another package."
-    ) from error
+        "You can install it with `pip install --pre torchdata --extra-index-url https://download.pytorch.org/whl/nightly/cpu"
+    ) from None
```
NicolasHug, line 7: What's the reason not to `raise from error` like before?

NicolasHug (approved):

torchvision/prototype/datasets/__init__.py
pmeier: After the removal of the `TypeError`, showing the excepted error adds no information:

```py
try:
    import foo
except ImportError as error:
    raise ModuleNotFoundError("foo is not available") from error
```

```
  File "main.py", line 2, in <module>
    import foo
ModuleNotFoundError: No module named 'foo'

The above exception was the direct cause of the following exception:

  File "main.py", line 4, in <module>
    raise ModuleNotFoundError("foo is not available") from error
ModuleNotFoundError: foo is not available
```

vs.

```py
try:
    import foo
except ImportError as error:
    raise ModuleNotFoundError("foo is not available")
```

```
  File "main.py", line 2, in <module>
    import foo
ModuleNotFoundError: No module named 'foo'

  File "main.py", line 4, in <module>
    raise ModuleNotFoundError("foo is not available")
ModuleNotFoundError: foo is not available
```

vs.

```py
try:
    import foo
except ImportError as error:
    raise ModuleNotFoundError("foo is not available") from None
```

```
  File "main.py", line 4, in <module>
    raise ModuleNotFoundError("foo is not available") from None
ModuleNotFoundError: foo is not available
```

PR #5577, opened 2022-03-09
NOMRG Models docs revamp -- v2

NicolasHug:
Follow up to https://github.com/pytorch/vision/pull/5575, adding documentation for the Weight enums.

The Weight docs is embedded in the same page as their corresponding model builders: https://1256058-73328905-gh.circle-artifacts.com/0/docs/prototype_models/generated/torchvision.prototype.models.resnet50.html#torchvision.prototype.models.resnet50
image

PR #5741, opened 2022-04-05
WIP Add table summarizing classification weights accuracies in docs

NicolasHug:
Rendered docs: https://output.circle-artifacts.com/output/job/e411a4b2-f155-422e-bd08-e03f854aecc6/artifacts/0/docs/models.html

Table looks like this

image

I'll make sure to add links from the weight names to their documentation page as in https://github.com/pytorch/vision/pull/5577
Ideally the rows should be more narrow but I don't know how to do that yet. Not too big of a deal anyway.


datumbox:
@NicolasHug It looks pretty good. I assume we can fix the parameters to not use scientific format, right?

NicolasHug:
I did it on purpose to give an idea of the order of magnitude directly, but I don't have a strong opinion, I'll revert

docs/source/conf.py
```diff
@@ -292,5 +292,36 @@ def inject_minigalleries(app, what, name, obj, options, lines):
...
 
 
+def generate_table():
+
+    import torchvision.models as M
+    from tabulate import tabulate
+    import textwrap
+
+    # TODO: this is ugly af and incorrect. We'll need an automatic way to
+    # retrieve weight enums for each section, or manually list them.
```
NicolasHug, line 302: Any thought on that @datumbox ?

datumbox: Yes you are right. We need a registration mechanism for the models. The new Datasets API has one, so part of the reason I didn't want to invest time creating one is to potentially adopt/extend the one on Datasets. Thoughts?

NicolasHug: The registration of the dataset is very very basic, it's just a decorator that adds the callable/object to a private dict. It would probably make sense to use something similar for the models / weights. Whether we should be relying on the same utils though is up for discussion - as a first version I'd suggest not to merge things and for the models to have a separate implementation. The code is really basic anyway.

datumbox: I'm happy with what you propose and on the technical details you mentioned. I had a similar simple approach on the [original proposal](https://github.com/datumbox/dapi-model-versioning/blob/main/dapi_lib/models/_api.py#L162) of the Multiweights support but I didn't port it to adopt some solution in common with Datasets. The code doesn't have to be the same but I think the interface can be basic and similar. As discuss offline, the only thing different for models is the fact that there is a hierarchy (Detection, Optical Flow, Classification etc) and this needs to be taken into account because names across modules conflict (for example `resnet50` exists both in Classification and Quantizaztion submodules).

PR #5742, opened 2022-04-05
add test to enforce __len__ is working on prototype datasets

pmeier:
Note that this test only checks if the `__len__` method returns a plausible result. It cannot check the actual value, since the mock data only contains a few examples. Thus, `len(dataset)` diverges from the actual number of samples. The latter is tested in  

https://github.com/pytorch/vision/blob/b5149557f0a7113860836faf71514c4238c7a2cc/test/test_prototype_builtin_datasets.py#L85


NicolasHug (approved):
Thanks @pmeier

PR #5743, opened 2022-04-05
migrate VOC prototype dataset

pmeier:


torchvision/prototype/datasets/_builtin/voc.py
```diff
@@ -67,31 +71,35 @@ def _make_info(self) -> DatasetInfo:
...
-        detection="Main",
-        segmentation="Segmentation",
-    )
+    @property
+    def _anns_folder(self) -> str:
+        return "Annotations" if self._task == "detection" else "SegmentationClass"
+
+    @property
+    def _split_folder(self) -> str:
+        return "Main" if self._task == "detection" else "Segmentation"
```
NicolasHug, line 85: Should these be set in `__init__()`?

torchvision/prototype/datasets/_builtin/voc.py
```diff
@@ -12,48 +13,51 @@
...
+        **kwargs: Any,
+    ) -> None:
+        self._year = self._verify_str_arg(year, "year", ("2007", "2008", "2009", "2010", "2011", "2012"))
+        if split == "test" and year != "2007":
+            raise ValueError("`split='test'` is only available for `year='2007'`")
+        else:
+            self._split = self._verify_str_arg(split, "split", ("train", "val", "trainval", "test"))
+        self._task = self._verify_str_arg(task, "task", ("detection", "segmentation"))
+
+        self._categories: List[str] = _info()["categories"]
```
NicolasHug, line 58: Are we going to need to set types for these every single time?

NicolasHug (approved):
Thanks @pmeier . I didn't look at the test since you suggested offline that this will be reverted. The rest LGTM with a minor caveat regarding the use of `@property`.

torchvision/prototype/datasets/_builtin/voc.py
pmeier: Nope. I've copy-pasted them from the imagenet implementation, but it seems we also don't need them there anymore. I'll remove.

PR #5745, opened 2022-04-05
Migrate PCAM prototype dataset

NicolasHug:

### DIFF
--- a/torchvision/prototype/datasets/_builtin/pcam.py
+++ b/torchvision/prototype/datasets/_builtin/pcam.py
@@ -130,5 +130,4 @@ class PCAM(Dataset2):
         return Mapper(dp, self._prepare_sample)
 
     def __len__(self):
-        # TODO imlement len
-        return 3
+        return 262_144 if self._split == "train" else 32_768
--- a/torchvision/prototype/datasets/_builtin/pcam.py
+++ b/torchvision/prototype/datasets/_builtin/pcam.py
@@ -2,6 +2,7 @@ import io
 import pathlib
 from collections import namedtuple
 from typing import Any, Dict, List, Optional, Tuple, Iterator, Union
+from unicodedata import category
 
 from torchdata.datapipes.iter import IterDataPipe, Mapper, Zipper
 from torchvision.prototype import features
@@ -46,10 +47,7 @@ _Resource = namedtuple("_Resource", ("file_name", "gdrive_id", "sha256"))
 
 @register_info(NAME)
 def _info() -> Dict[str, Any]:
-    return dict(
-        categories=[0, 1],
-        dependencies=["h5py"],
-    )
+    return dict(categories=["0", "1"])
 
 
 @register_dataset(NAME)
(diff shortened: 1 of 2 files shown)


torchvision/prototype/datasets/_builtin/pcam.py
```diff
@@ -40,15 +44,25 @@ def __iter__(self) -> Iterator[Tuple[str, io.IOBase]]:
...
-            homepage="https://github.com/basveeling/pcam",
-            categories=2,
-            valid_options=dict(split=("train", "test", "val")),
-            dependencies=["h5py"],
-        )
+@register_info(NAME)
+def _info() -> Dict[str, Any]:
+    return dict(
+        categories=[0, 1],
+        dependencies=["h5py"],
```
pmeier, line 51: Do we need this in the static info? Or better, can someone actually do something with this information at runtime? In #5473 I added a [`dependencies` parameter to `Dataset2.__init__`](https://github.com/pytorch/vision/pull/5473/files#diff-74f5dec81cf4038124e53c3db44c91281f1c60d6f036574a07c3dc2ce30d65ed) similar to what the current `DatasetInfo` is doing:

https://github.com/pytorch/vision/blob/63576c9f46d9c6f471db789d41149cdb6fea435b/torchvision/prototype/datasets/utils/_dataset.py#L29

Maybe we can wait for the other PR to be merged and adopt this change here?

torchvision/prototype/datasets/_builtin/pcam.py
```diff
@@ -40,15 +44,25 @@ def __iter__(self) -> Iterator[Tuple[str, io.IOBase]]:
...
-            "pcam",
-            homepage="https://github.com/basveeling/pcam",
-            categories=2,
-            valid_options=dict(split=("train", "test", "val")),
-            dependencies=["h5py"],
-        )
+@register_info(NAME)
+def _info() -> Dict[str, Any]:
+    return dict(
+        categories=[0, 1],
```
pmeier, line 50: Categories should be strings

```suggestion
        categories=["0", "1"],
```

torchvision/prototype/datasets/_builtin/pcam.py
```diff
@@ -116,3 +128,7 @@ def _make_datapipe(
         dp = hint_shuffling(dp)
         dp = hint_sharding(dp)
         return Mapper(dp, self._prepare_sample)
+
+    def __len__(self):
+        # TODO imlement len
+        return 3
```
pmeier, line 134: Just marking so we don't forget. Do you have the files locally? Otherwise I can provide you the number of samples.

torchvision/prototype/datasets/_builtin/pcam.py
NicolasHug: Why? That sounds overkill for non-labeled classes which are usually treated as ints in other popular frameworks

pmeier: The the naming scheme a "label" is the integer representation while "category" is the human-readable string representation. That is the same what `categories=2` did before:

https://github.com/pytorch/vision/blob/08cc9a7fb5baa9655fa5d2dd44a6fcfcbe48a1c7/torchvision/prototype/datasets/utils/_dataset.py#L44-L45

So we either should stick to this scheme and make the categories strings or drop the categories. I agree, the latter might be the better option for datasets that don't have human readable categories.

NicolasHug: Let's just mark this as a TODO for later. For now we can assume that all datasets have a `categories` info. I guess it's still useful so as to know the number of classes

torchvision/prototype/datasets/_builtin/pcam.py
pmeier, line 134:
```suggestion
        return 262_144 if self._split == "train" else 32_768
```

pmeier (approved):
Thanks @NicolasHug, LGTM! We need to wait or #5473 before this can pass CI due to the missing `dependencies` parameter in the base class.

PR #5749, opened 2022-04-06
migrate caltech prototype datasets

pmeier:


test/builtin_dataset_mocks.py
```diff
@@ -370,8 +370,8 @@ def cifar100(info, root, config):
     return len(train_files if config.split == "train" else test_files)
 
 
-# @register_mock
-def caltech101(info, root, config):
+@register_mock(configs=[dict()])
```
pmeier, line 373: We could also make this the default value, but I think being explicit here is fine,

test/builtin_dataset_mocks.py
```diff
@@ -390,15 +390,17 @@ def create_ann_folder(root, name, file_name_fn, num_examples):
     images_root = root / "101_ObjectCategories"
     anns_root = root / "Annotations"
 
-    ann_category_map = {
-        "Faces_2": "Faces",
-        "Faces_3": "Faces_easy",
-        "Motorbikes_16": "Motorbikes",
-        "Airplanes_Side_2": "airplanes",
+    image_category_map = {
```
pmeier, line 393: The map was inversed before and did nothing.

NicolasHug: So the annotation folder had the same name as the image folder.  I'm wondering why the test didn't fail then. Is it because the dataset code doesn't actually care about the annotation folder names?

torchvision/prototype/datasets/_builtin/caltech.py
```diff
@@ -9,26 +9,48 @@
...
+        self,
+        root: Union[str, pathlib.Path],
+        skip_integrity_check: bool = False,
+    ) -> None:
+        self._categories = _caltech101_info()["categories"]
+
+        super().__init__(
+            root,
+            # TODO: this will only be available after https://github.com/pytorch/vision/pull/5473
+            # dependencies=("scipy",),
```
NicolasHug, line 49: just making sure we don't forget before merging

NicolasHug (approved):
Minor Q but LGTM, thanks @pmeier

test/builtin_dataset_mocks.py
pmeier: To give some context: Caltech101 uses folders to separate the categories. The folder names match between the images and annotations except for four exceptions. The implementation matches the annotations folder names with this map

https://github.com/pytorch/vision/blob/08cc9a7fb5baa9655fa5d2dd44a6fcfcbe48a1c7/torchvision/prototype/datasets/_builtin/caltech.py#L45-L50

and this logic

https://github.com/pytorch/vision/blob/08cc9a7fb5baa9655fa5d2dd44a6fcfcbe48a1c7/torchvision/prototype/datasets/_builtin/caltech.py#L72-L73

The old mock data used the already matched names for the annotations so the dataset didn't need to do any matching. This means that the test were passing fine, although the implementation might be broken (it wasn't), since the matching was never tested.

PR #5750, opened 2022-04-06
migrate CelebA prototype dataset

pmeier:


torchvision/prototype/datasets/_builtin/celeba.py
```diff
@@ -60,15 +59,37 @@ def __iter__(self) -> Iterator[Tuple[str, Dict[str, str]]]:
...
+        *,
+        split: str = "train",
+        skip_integrity_check: bool = False,
+    ) -> None:
+        self._split = self._verify_str_arg(split, "split", ("train", "val", "test"))
+        self._split_id = {
+            "train": "0",
+            "val": "1",
+            "test": "2",
+        }[split]
```
NicolasHug, line 88: Should we just inline that in `_filter_split` where it's used? Also, any reason to use strings here?

NicolasHug (approved):
Minor comment but LGTM, thanks

torchvision/prototype/datasets/_builtin/celeba.py
pmeier: > Should we just inline that in `_filter_split` where it's used?

:+1:

> Also, any reason to use strings here?

The actual id is read from a csv and thus is a string.

PR #5757, opened 2022-04-06
Migrate DTD prototype dataset

NicolasHug:


test/builtin_dataset_mocks.py
```diff
@@ -965,8 +965,8 @@ def food101(info, root, config):
     return num_samples_map[config.split]
 
 
-# @register_mock
-def dtd(info, root, config):
+@register_mock(configs=combinations_grid(split=("train", "val", "test"), fold=(1, 4)))
```
pmeier, line 968: I agree we probably don't need to test all folds, but maybe we can do the first, the last and one from the middle?

```suggestion
@register_mock(configs=combinations_grid(split=("train", "val", "test"), fold=(1, 4, 10)))
```

torchvision/prototype/datasets/_builtin/dtd.py
```diff
@@ -14,30 +13,55 @@
...
+        self,
+        root: Union[str, pathlib.Path],
+        *,
+        split: str = "train",
+        fold: int = 1,
+        skip_validation_check: bool = False,
+    ) -> None:
+        self._split = self._verify_str_arg(split, "split", {"train", "val", "test"})
+
+        if not 1 <= fold <= 10:
```
pmeier, line 56: I always forget the precedence for this two-way comparison. If you don't mind, I would feel better with extra parentheses

```suggestion
        if not (1 <= fold <= 10):
```

pmeier (approved):
Two minor comments. Otherwise LGTM. Thanks Nicolas!

PR #5758, opened 2022-04-06
Migrate Food101 prototype dataset

NicolasHug:

### DIFF
--- a/torchvision/prototype/datasets/_builtin/food101.py
+++ b/torchvision/prototype/datasets/_builtin/food101.py
@@ -1,5 +1,5 @@
 from pathlib import Path
-from typing import Any, Tuple, List, Dict, Optional, BinaryIO
+from typing import Any, Tuple, List, Dict, Optional, BinaryIO, Union
 
 from torchdata.datapipes.iter import (
     IterDataPipe,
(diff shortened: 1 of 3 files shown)


torchvision/prototype/datasets/_builtin/food101.py
```diff
@@ -83,9 +95,15 @@ def _make_datapipe(
...
+        dp = resources[0].load(self._root)
         dp = Filter(dp, path_comparator("name", "classes.txt"))
         dp = LineReader(dp, decode=True, return_path=False)
         return list(dp)
+
+    def __len__(self) -> int:
+        return {
+            "train": 75_750,
+            "test": 25_250,
+        }[self._split]
```
pmeier, line 109: Nit: do we always want to go with dicts here? For two values, I went for a simple `if` / `else`

```suggestion
        return 75_750 if self._split == "train" else 25_250
```

No strong opinion though.

pmeier (approved):
One nit, otherwise LGTM. Thanks Nicolas!

pmeier:
cc @yassineAlouini This is the PR that moves the Food101 dataset to the new style.

yassineAlouini:
> cc @yassineAlouini This is the PR that moves the Food101 dataset to the new style.

Thanks @pmeier for the link. This looks neat, well done to the team.

PR #5761, opened 2022-04-06
Migrate Semeion prototype dataset

NicolasHug:

### DIFF
--- a/torchvision/prototype/datasets/_builtin/semeion.py
+++ b/torchvision/prototype/datasets/_builtin/semeion.py
@@ -1,31 +1,43 @@
-from typing import Any, Dict, List, Tuple
+import pathlib
+from typing import Any, Dict, List, Tuple, Union
 
 import torch
+from pytest import skip
 from torchdata.datapipes.iter import (
     IterDataPipe,
     Mapper,
     CSVParser,
 )
 from torchvision.prototype.datasets.utils import (
-    Dataset,
-    DatasetConfig,
-    DatasetInfo,
+    Dataset2,
     HttpResource,
     OnlineResource,
 )
 from torchvision.prototype.datasets.utils._internal import hint_sharding, hint_shuffling
 from torchvision.prototype.features import Image, OneHotLabel
 
+from .._api import register_dataset, register_info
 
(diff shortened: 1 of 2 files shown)


torchvision/prototype/datasets/_builtin/semeion.py
```diff
@@ -36,18 +48,16 @@ def _prepare_sample(self, data: Tuple[str, ...]) -> Dict[str, Any]:
         image_data, label_data = data[:256], data[256:-1]
 
         return dict(
-            image=Image(torch.tensor([float(pixel) for pixel in image_data], dtype=torch.uint8).reshape(16, 16)),
-            label=OneHotLabel([int(label) for label in label_data], categories=self.categories),
+            image=Image(torch.tensor([float(pixel) for pixel in image_data], dtype=torch.float).reshape(16, 16)),
```
NicolasHug, line 51: @pmeier  the original PR used `uint8` here, but these are float values in [0, 1]. Do you know if the use of `dtype=uint8` was inteded for something specific?


Also, the manual conversion of every single value is fairly inefficient for all of the 16 * 16 pixels. Do you know if there's a more efficient conversion that could be applied at the CSVParser level?

pmeier: > the original PR used `uint8` here, but these are float values in [0, 1]. Do you know if the use of `dtype=uint8` was inteded for something specific?

Nope, that seems like a bug. Good catch!

> Also, the manual conversion of every single value is fairly inefficient for all of the 16 * 16 pixels. Do you know if there's a more efficient conversion that could be applied at the CSVParser level?

I'm not aware of anything. Without a user request for a better performance, I wouldn't sweat it though. I'm guessing this dataset is not in widespread use and the code will probably get more complicated. If you have a simple solution, go for it.

pmeier: Just saw that this dataset has only 1.5k samples. If someone wants more speed, the best way out is probably to just load the whole dataset into memory.

pmeier (approved):
LGTM, thanks Nicolas!

PR #5765, opened 2022-04-06
migrate cub200 prototype dataset

pmeier:


torchvision/prototype/datasets/_builtin/cub200.py
```diff
@@ -28,26 +27,53 @@
...
+    def __init__(
+        self,
+        root: Union[str, pathlib.Path],
+        *,
+        split: str = "train",
+        year: str = "2011",
+        skip_integrity_check: bool = False,
+    ) -> None:
+        self._split = self._verify_str_arg(split, "split", ("train", "test"))
+        self._year = self._verify_str_arg(year, "year", ("2010", "2011"))
```
NicolasHug, line 64: Just a random thought, I wonder if we should allow year to be an int as well?

torchvision/prototype/datasets/_builtin/cub200.py
```diff
@@ -144,22 +170,17 @@ def _prepare_sample(
...
 
-        image = EncodedImage.from_file(buffer)
+        # image = EncodedImage.from_file(buffer)
 
         return dict(
-            prepare_ann_fn(anns_data, image.image_size),
-            image=image,
-            label=Label(int(pathlib.Path(path).parent.name.rsplit(".", 1)[0]), categories=self.categories),
+            # prepare_ann_fn(anns_data, image.image_size),
+            # image=image,
```
NicolasHug, line 177: is this expected?

torchvision/prototype/transforms/__init__.py
```diff
@@ -20,6 +20,6 @@
 )
 from ._meta import ConvertBoundingBoxFormat, ConvertImageDtype, ConvertImageColorSpace
 from ._misc import Identity, Normalize, ToDtype, Lambda
-from ._type_conversion import DecodeImage, LabelToOneHot
+from ._type_conversion import DecodeImage, LabelToOneHot, ToImagePIL, ToImageTensor
```
NicolasHug, line 23: Is this related?

NicolasHug (approved):
Thanks @pmeier , minor comments but LGTM

torchvision/prototype/transforms/__init__.py
pmeier: Nope, sorry.

torchvision/prototype/datasets/_builtin/cub200.py
pmeier: Probably, yes. At least it would be unexpected that `year=2011` does not work. VOC currently uses only strings as well. Let's put this in our backlog for the "final" API review before we merge into `main`.

PR #5770, opened 2022-04-06
fix category file generation

pmeier:
We missed that in the other reviews. I'll re-run  the generation for all migrated datasets again before we merge into `main`.


NicolasHug (approved):
Thanks Philip

PR #5771, opened 2022-04-06
Migrate USPS prototype dataset

NicolasHug:


pmeier (approved):
LGTM, thanks Nicolas! cc @lezwon This is the PR that migrates the USPS dataset to the new style I talked about.

PR #5772, opened 2022-04-06
migrate SBD prototype dataset

pmeier:


torchvision/prototype/datasets/_builtin/sbd.py
```diff
@@ -26,22 +20,44 @@
...
+    def __init__(
+        self,
+        root: Union[str, pathlib.Path],
+        *,
+        split: str = "train",
+        skip_integrity_check: bool = False,
+    ) -> None:
+        self._split = self._verify_str_arg(split, "split", ("train", "val", "train_noval"))
+
+        self._categories = _info()["categories"]
```
NicolasHug, line 56: Since you created `_CATEGORIES`, should we just use it here?

NicolasHug (approved):
Minor nit but LGTM, thanks @pmeier

PR #5774, opened 2022-04-06
Cleanup prototype dataset implementation

NicolasHug:
I tried to list out the steps in the commits below


torchvision/prototype/datasets/utils/_dataset.py
```diff
@@ -1,191 +1,21 @@
...
-from .._home import use_sharded_dataset
-from ._internal import BUILTIN_DIR, _make_sharded_datapipe
 from ._resource import OnlineResource
 
 
-class DatasetConfig(FrozenBunch):
-    # This needs to be Frozen because we often pass configs as partial(func, config=config)
-    # and partial() requires the parameters to be hashable.
-    pass
+def read_categories_file(path: pathlib.Path) -> List[List[str]]:
```
pmeier, line 13: Maybe we can move this into `_internal` and provide the extra convenience to load the file by only supplying the dataset name? I don't think we need to expose this, given that without sugar, the value of this function is quite minimal.

pmeier:
The category file reading was broken for the datasets that I implemented. I used this idiom

```py
CALTECH101_CATEGORIES, *_ = zip(*DatasetInfo.read_categories_file(BUILTIN_DIR / "caltech101.categories"))
```

which does not work if you already unpack single item lists inside the function. I pushed a commit to fix that.

pmeier (approved):
LGTM if CI is green. Thanks Nicolas!

Issue #2980, opened 2020-11-10
[RFC] Loss Functions in Torchvision


oke-aditya:
## 🚀 Feature

A loss functions API in torchvision.

## Motivation

The request is simple, we have loss functions available in torchvision
E.g. `sigmoid_focal_loss` , `l1_loss`. But these are quite scattered and we have to use `torchvision.ops.sigmoid_focal_loss` etc.

In future we might need to include further loss functions. E.g. `dice_loss`

Since loss functions are differentiable we can put them under `nn`.
We can have

`torchvision.nn.losses.sigmoid_focal_loss` and so on.

This keeps scope of `nn` open for other differentiable functions such as layers, etc.

## Pitch

These losses are very specific and pertain to vision domain. These are really useful and in general not tied to any specific model.
Though the loss functions that we keep are usually in `torch`. If we keep under `nn` namespace, future migration stays simple.

instead of `torchvision.nn.sigmoid_focal_loss` it would be `torch.nn.sigmoid_focal_loss`.

This Pitch comes from the above issues. 
[Dice Loss](https://github.com/pytorch/pytorch/issues/1249)
[More Loss Functions](https://github.com/pytorch/pytorch/issues/41142)
Recently Added [Focal Loss](https://github.com/pytorch/pytorch/issues/35882)

## Alternatives

Alternatively this should go in torch. But if we keep the above idea, we can support them in torchvision and later deprecate and move to torch (when needed).

Currently we include them under `ops` but it is actually not an `operation` it is a differentiable loss functions. 

Where as other ops are not differentiable and perform transformations / some manipulation over boxes / layers.

## Additional context

I am unsure of what loss functions are most popular / should be included. We could make a list of them and think.

mthrok:
cc. @dongreenberg @cpuhrsch I think this is very reasonable, there is a similar work going for torchaudio too. We should resume the library naming convention discussion and wrap it up to provide a comprehensive solution for loss/metrics.

mthrok:
@oke-aditya 
The domain team had a brief discussion on this;

* we agree that domain specific loss functions are coming up.
* But we would like to hold off on creating a dedicated module untill we actually have a good number of functions that fall into the category. It's easy to add such a module, but once added we can not remove it.
* meanwhile we can update the documentation and add a new category so that the existing loss is easy to find.

What do you think?

cc @fmassa

oke-aditya:
I agree with your thoughts @mthrok . It can be too early to call for such API.
[...]

PR #5691, opened 2022-03-28
Remove input checks for generalized_box_iou

datumbox:
Similar to #5322 and as discussed at https://github.com/pytorch/vision/pull/5684#issuecomment-1079746782, we shouldn't put input checks in the box utility ops to avoid slow-downs due to GPU/CPU synchronisation.

Issue #5322, opened 2022-01-31
generalized_box_iou_loss input checks raise on nan values


nlgranger:
### 🐛 Describe the bug

The assert:
https://github.com/pytorch/vision/blob/abdae5a1ea8ad8ce0b155a8bda090683c34d3551/torchvision/ops/giou_loss.py#L40-L41
crashes on nan inputs instead of just forwarding the nan values to the output.

It can be resolved by using this test instead:

```py
    assert not (x2 < x1).any(), "bad box: x1 larger than x2"
    assert not (y2 < y1).any(), "bad box: y1 larger than y2"
```

Not sure there should be any tests at all though. It costs a GPU->CPU synchronization.

### Versions

PyTorch version: 1.11.0.dev20220131
Is debug build: False
CUDA used to build PyTorch: 11.3
ROCM used to build PyTorch: N/A

OS: Arch Linux (x86_64)
GCC version: (GCC) 11.1.0
Clang version: 13.0.0
CMake version: version 3.22.2
Libc version: glibc-2.33

Python version: 3.9.7 (default, Sep 16 2021, 13:09:58)  [GCC 7.5.0] (64-bit runtime)
Python platform: Linux-5.15.17-1-lts-x86_64-with-glibc2.33
Is CUDA available: True
CUDA runtime version: 11.3.109
GPU models and configuration: GPU 0: Quadro T1000
Nvidia driver version: 495.46
cuDNN version: Probably one of the following:
/opt/cuda-11.3/targets/x86_64-linux/lib/libcudnn.so.8.2.0
/opt/cuda-11.3/targets/x86_64-linux/lib/libcudnn_adv_infer.so.8.2.0
/opt/cuda-11.3/targets/x86_64-linux/lib/libcudnn_adv_train.so.8.2.0
/opt/cuda-11.3/targets/x86_64-linux/lib/libcudnn_cnn_infer.so.8.2.0
/opt/cuda-11.3/targets/x86_64-linux/lib/libcudnn_cnn_train.so.8.2.0
/opt/cuda-11.3/targets/x86_64-linux/lib/libcudnn_ops_infer.so.8.2.0
/opt/cuda-11.3/targets/x86_64-linux/lib/libcudnn_ops_train.so.8.2.0
HIP runtime version: N/A
MIOpen runtime version: N/A

Versions of relevant libraries:
[pip3] numpy==1.21.2
[pip3] torch==1.11.0.dev20220131
[...]

PR #5776, opened 2022-04-06
Added CIOU loss function

abhi-glitchhg:
#2980 

[Reference](https://github.com/facebookresearch/detectron2/blob/221448e4dcfbebf215b8d21ae7e4b1dfbf422d29/detectron2/layers/losses.py#L66) 

cc: @oke-aditya  @datumbox


oke-aditya:
Hi. It's late in India rn. Will review it in daytime.

torchvision/ops/ciou_loss.py
```diff
@@ -0,0 +1,84 @@
...
+
+    """
+
+    x1, y1, x2, y2 = boxes1.unbind(dim=-1)
+    x1g, y1g, x2g, y2g = boxes2.unbind(dim=-1)
+
+    if (x2 < x1).all():
+        raise ValueError("x1 is larger than x2")
+    if (y2 < y1).all():
+        raise ValueError("y1 is larger than y2")
```
datumbox, line 40: We shouldn't put an assertion here to avoid slowdowns due to GPU<=>CPU syncs.

datumbox (commented):
@abhi-glitchhg Thanks for the contribution!

I've added a few comments for your attention.

torchvision/ops/ciou_loss.py
```diff
@@ -0,0 +1,84 @@
...
+    boxes1: torch.Tensor,
+    boxes2: torch.Tensor,
+    reduction: str = "none",
+    eps: float = 1e-7,
+) -> torch.Tensor:
+
+    """
+    Original Implementation from
+    https://github.com/facebookresearch/detectron2/blob/main/detectron2/layers/losses.py
+
```
oke-aditya, line 16:
```suggestion
    Some description about CIoU
    This loss is symmetric, so the boxes1 and boxes2 arguments are interchangeable.
     Both sets of boxes are expected to be in ``(x1, y1, x2, y2)`` format with
    ``0 <= x1 < x2`` and ``0 <= y1 < y2``, and The two boxes should have the
    same dimensions.
```

oke-aditya (commented):
Nice a few comments though

torchvision/ops/ciou_loss.py
```diff
@@ -0,0 +1,84 @@
+import math
+
+import torch
+
+
+def ciou_loss(
```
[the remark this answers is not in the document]
datumbox: Good spot. The name of the file and the method can't be the same. Else it will create ambiguities on the imports from `torchvision.ops`. Following the example of giou, the file could be called `ciou_loss.py` and the method `complete_box_iou_loss`; I'm not terribly happy with the latter if I'm honest. Thoughts?

torchvision/ops/ciou_loss.py
abhi-glitchhg: Okay I will do that.

abhi-glitchhg: A doubt - 

So the user has to manually check if there are no bad inputs to this function, right?  should we have some warning message in that scenario?

torchvision/ops/ciou_loss.py
```diff
@@ -0,0 +1,84 @@
...
+            ``'sum'``: The output will be summed. Default: ``'none'``
+        eps : (float, optional): small number to prevent division by zero. Default: 1e-7
+
+    Reference:
+
+    Complete Intersection over Union Loss (Zhaohui Zheng et. al)
+    https://arxiv.org/abs/1911.08287
+
+    """
+
```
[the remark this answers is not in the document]
abhi-glitchhg: Thanks, Aditya for the review.  I love your feedback. :)
 I am not much familiar with jit logging. 
 
 Is there any material/documentation that I can go through to understand this? 
 Thanks.

torchvision/ops/ciou_loss.py
abhi-glitchhg: Understood thanks!

torchvision/ops/ciou_loss.py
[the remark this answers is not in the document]
oke-aditya: * PyTorch note: https://pytorch.org/docs/stable/notes/large_scale_deployments.html#api-usage-logging;
    * Logging policy: https://github.com/pytorch/vision/issues/5052;

also see https://github.com/pytorch/vision/blob/main/torchvision/utils.py#L525

torchvision/ops/ciou_loss.py
datumbox: The problem is you can't have a warning unless you check and if you check you slowdown the execution significantly. The user needs to make sure that they don't provide bad inputs. If they want the warnings the recommendation is to wrap this method on their end and perform those checks.

You can see similar discussions at #5322, #5691 and #5684.

datumbox: Damn, Github **really** needs to implement a feature to info you that others are writing comments at the same time. It happens all the time that we are posting concurrently the same things. lol :D

yassineAlouini:
Working on dIoU, I have noticed that many things can be refactored between gIoU, cIoU, and dIoU.
I will push my contribution anyway but we can discuss what can be done next to improve things. :ok_hand:

oke-aditya:
yeah :) There is a neat relation between these 3 and code can be re-used. :smile:

yassineAlouini:
> yeah :) There is a neat relation between these 3 and code can be re-used. smile

Yes, I guess we can repeat some of the code for now since we have two branches at the same time + the existing gIoU. Next, we can start a new PR to refactor things a bit.

test/test_ops.py
```diff
@@ -1450,5 +1451,25 @@ def test_is_leaf_node(self, dim, p, block_size, inplace):
         assert len(graph_node_names[0]) == 1 + op_obj.n_inputs
 
 
+class TestCIOULoss:
+    box1 = ([-1, -1, 1, 1],[-1, -1, 1, 1],[-1, -1, 1, 1],[-1, -1, 1, 1])
```
oke-aditya, line 1455: hmm, to be on safer side can you add `dtype=torch.float`. It might just take these as ints
also 
```suggestion
    box1 = ([-1., -1., 1., 1.],[-1., -1., 1., 1.],[-1., -1., 1., 1.],[-1., -1., 1., 1.])
```

torchvision/ops/ciou_loss.py
```diff
@@ -0,0 +1,97 @@
+import math
+
+import torch
+from torch import Tensor
+
+from ..utils import _log_api_usage_once
+
+
+def _upcast(t: Tensor) -> Tensor:
```
oke-aditya, line 9: (but why duplicate?)

torchvision/ops/ciou_loss.py
```diff
@@ -0,0 +1,97 @@
...
+#[0.0000, 0.8125, 1.1923, 1.2500]
+
+def complete_box_iou_loss(
+    boxes1: torch.Tensor,
+    boxes2: torch.Tensor,
+    reduction: str = "none",
+    eps: float = 1e-7,
+) -> torch.Tensor:
+
+    """
```
oke-aditya, line 23:
```suggestion
    """
    Need one line describing CIoU loss.
```

test/test_ops.py
abhi-glitchhg: oh yeah, will add tests with different datatypes.

oke-aditya:
Please add docs in docs/ops.rst

test/test_ops.py
```diff
@@ -79,7 +79,7 @@ def test_forward(self, device, contiguous, x_dtype=None, rois_dtype=None, **kwar
         rois_dtype = self.dtype if rois_dtype is None else rois_dtype
         pool_size = 5
         # n_channels % (pool_size ** 2) == 0 required for PS opeartions.
-        n_channels = 2 * (pool_size ** 2)
+        n_channels = 2 * (pool_size**2)
```
oke-aditya, line 82: I think this is unrelated or due to formatter. But I think the original one is right.

test/test_ops.py
```diff
@@ -115,7 +115,7 @@ def test_is_leaf_node(self, device):
     def test_backward(self, seed, device, contiguous):
         torch.random.manual_seed(seed)
         pool_size = 2
-        x = torch.rand(1, 2 * (pool_size ** 2), 5, 5, dtype=self.dtype, device=device, requires_grad=True)
+        x = torch.rand(1, 2 * (pool_size**2), 5, 5, dtype=self.dtype, device=device, requires_grad=True)
```
oke-aditya, line 118: Same here

test/test_ops.py
```diff
@@ -1569,5 +1569,42 @@ def test_jit(self, alpha, gamma, reduction, device, dtype, seed) -> None:
...
+        )
+        box2s = torch.stack(
+            [box3, box4],
+            dim=0,
+        )
+
+        def assert_ciou_loss(box1, box2, expected_output, reduction="none"):
+
+            output = ops.complete_box_iou_loss(box1, box2, reduction=reduction)
+            expected_output = torch.tensor(expected_output, device=device, dtype=dtype)
```
oke-aditya, line 1593:
```suggestion
            expected_output = torch.tensor(expected_output, device=device)
```

This will automatically take the dtype. You don't need to parameterize here.

test/test_ops.py
```diff
@@ -1569,5 +1569,42 @@ def test_jit(self, alpha, gamma, reduction, device, dtype, seed) -> None:
...
+            tol = 1e-5 if dtype != torch.half else 1e-3
+            torch.testing.assert_close(output, expected_output, rtol=tol, atol=tol)
+
+        assert_ciou_loss(box1, box1, 0.0)
```
oke-aditya, line 1597: 0.0 is treated with dtype as float and will fail when you pass boxes as torch.half. That's what CI us complaining.

torchvision/ops/ciou_loss.py
```diff
@@ -0,0 +1,94 @@
...
+    x_g = (x1g + x2g) / 2
+    y_g = (y1g + y2g) / 2
+    distance = ((x_p - x_g) ** 2) + ((y_p - y_g) ** 2)
+
+    # width and height of boxes
+    w_pred = x2 - x1
+    h_pred = y2 - y1
+    w_gt = x2g - x1g
+    h_gt = y2g - y1g
+    v = (4 / (math.pi ** 2)) * torch.pow((torch.atan(w_gt / h_gt) - torch.atan(w_pred / h_pred)), 2)
```
oke-aditya, line 84:
```suggestion
    v = (4 / (torch.pi ** 2)) * torch.pow((torch.atan(w_gt / h_gt) - torch.atan(w_pred / h_pred)), 2)
```

oke-aditya: We do have torch.pi. Very debated but finally included 

```
>>> torch.pi
3.141592653589793
```

the reason is general compatibility of array API (Broad topic, I have 0 knowledge) 
see
https://github.com/pytorch/pytorch/blob/main/torch/__init__.py#L587

torchvision/ops/ciou_loss.py
```diff
@@ -0,0 +1,94 @@
+import math
```
oke-aditya, line 1: safely drop this import after you remove math.pi

```suggestion

```

test/test_ops.py
abhi-glitchhg: Ohh! 
Thanks for the help!

test/test_ops.py
abhi-glitchhg: Tests were showing ufmt errors for `test/test_ops.py` file, so I used ufmt format for this file. 
Will revert these changes

oke-aditya: I'm not sure, of formatting. Since, I use pre-commit and it automatically (auto magically?) corrects stuff and then safely commits. You might as well try it :)

oke-aditya:
Once CI goes green (which probably should) please UNmark this as draft. @abhi-glitchhg

torchvision/ops/ciou_loss.py
abhi-glitchhg: There exists functions which depend on `math` module. 

https://github.com/pytorch/vision/blob/467b841ac2650c962952bbee5a542edb835c98bc/test/test_ops.py#L277-L278

But i see no problem in changing these  from `math.floor()` to `torch.floor()`. 
@oke-aditya

oke-aditya: I'm pointing to different file I guess. 
But good idea to replace there as well.

datumbox:
@abhi-glitchhg @oke-aditya Thank you very much for your contributions and reviews so far.

Apologies for the delayed response, but I was OOO. I might be missing something but what is the reason we have duplicate implementations for cIoU in the loss and the box version? Is there a reason we don't share some of the code?

yassineAlouini:
> @abhi-glitchhg @oke-aditya Thank you very much for your contributions and reviews so far.
> 
> Apologies for the delayed response, but I was OOO. I might be missing something but what is the reason we have duplicate implementations for cIoU in the loss and the box version? Is there a reason we don't share some of the code?

As far as my exploration goes, it seems it is the case for gIoU as well and I haven't found any reason why this should be the case but maybe I am missing something. I have also duplicated some code in dIoU. Perhaps it is best to have a PR to refactor some of the shared code between gIoU, cIoU, and dIoU?

datumbox:
@yassineAlouini I think sending a separate PR to refactor gIoU makes a lot of sense. For new code (cIoU and dIoU) I would fix in the current open PRs.

yassineAlouini:
@datumbox Noted, I will work on this (when I have allocated time) and update the code. :ok_hand:

torchvision/ops/boxes.py
```diff
@@ -311,6 +311,54 @@ def generalized_box_iou(boxes1: Tensor, boxes2: Tensor) -> Tensor:
...
+    boxes1 = _upcast(boxes1)
+    boxes2 = _upcast(boxes2)
+
+    inter, union = _box_inter_union(boxes1, boxes2)
+    iou = inter / union
+
+    lti = torch.min(boxes1[:, None, :2], boxes2[:, None, :2])
+    rbi = torch.max(boxes1[:, None, 2:], boxes2[:, None, 2:])
+
+    whi = _upcast(rbi - lti).clamp(min=0)  # [N,M,2]
```
datumbox, line 339: No need to upcast now. You've done it once above:
```suggestion
    whi = (rbi - lti).clamp(min=0)  # [N,M,2]
```

torchvision/ops/boxes.py
```diff
@@ -311,6 +311,54 @@ def generalized_box_iou(boxes1: Tensor, boxes2: Tensor) -> Tensor:
...
+    diagonal_distance_squared = (whi[:, :, 0] ** 2) + (whi[:, :, 1] ** 2) + eps
+
+    # centers of boxes
+    x_p = (boxes1[:, 0] + boxes1[:, 2]) / 2
+    y_p = (boxes1[:, 1] + boxes1[:, 3]) / 2
+    x_g = (boxes2[:, 0] + boxes2[:, 2]) / 2
+    y_g = (boxes2[:, 1] + boxes2[:, 3]) / 2
+    # The distance between boxes' centers squared.
+    centers_distance_squared = (_upcast(x_p - x_g) ** 2) + (_upcast(y_p - y_g) ** 2)
```
datumbox, line 348:
```suggestion
    centers_distance_squared = (x_p - x_g) ** 2 + (y_p - y_g) ** 2
```

datumbox (commented):
@abhi-glitchhg Just a minor comment below. When you are ready for us to check the PR, remove the draft flag. :)

torchvision/ops/ciou_loss.py
```diff
@@ -0,0 +1,92 @@
+import torch
+
+from ..utils import _log_api_usage_once
+from .giou_loss import _upcast
```
datumbox, line 4: Nit: This _upcast needs to move in `._utils`

datumbox (commented):
@abhi-glitchhg  Thanks for the contribution. 

I'm looping @yassineAlouini who is looking at dIoU as many of the comments below are applicable.

I've verified the implementation both by reviewing and by comparing the two methods as follows:
```python
import torch

from torchvision.ops.boxes import complete_box_iou
from torchvision.ops.ciou_loss import complete_box_iou_loss

def random_box(canvas_size):
    x1y1 = torch.rand((1, 2)) * canvas_size
    wh = torch.rand((1, 2)) * canvas_size
    x2y2 = (x1y1 + wh).clamp(0, canvas_size)
    return torch.cat((x1y1, x2y2), axis=1)


canvas_size = 1000


for _ in range(10000):
    box1 = random_box(canvas_size)
    box2 = random_box(canvas_size)

    v1 = 1 - complete_box_iou(box1, box2)
    v2 = complete_box_iou_loss(box1, box2)

    torch.testing.assert_close(v1[0], v2, rtol=0, atol=1e-6)
```

This confirms that both implementations are equivalent. Though I think the PR is correct, I feel there is need for deep cleaning both for dIoU and cIoU (and gIoU). In particular we should consider refactoring the code across losses and boxes to share some of the internal implementations. Not everything is shareable and we should be mindful to maintain the speed but we shouldn't have duplicate implementations. Moreover, @pmeier suggested a few improvements on the test side at https://github.com/pytorch/vision/pull/5786#discussion_r857325564, so these improvements should also be completed.

Kindly let me know what direction you would like us to take now. We could bring both PRs on a similar state as this one and merge and then 1 of you can pick up the cleaning bit. Alternatively we can keep iterating on the current PRs until the nits are taken care of. It's up to you and I'm happy to take your preferred approach, I just want to make sure the implementations will be cleaned up prior the next release.

abhi-glitchhg:
> Kindly let me know what direction you would like us to take now. We could bring both PRs on a similar state as this one and merge and then 1 of you can pick up the cleaning bit. Alternatively we can keep iterating on the current PRs until the nits are taken care of. It's up to you and I'm happy to take your preferred approach, I just want to make sure the implementations will be cleaned up prior the next release.

I would prefer the first approach(bring both PRs to similar state, merge, and then do refactor) but I'm also open to the second approach.

datumbox (approved):
@abhi-glitchhg Sounds good to me. Approving to unblock. We should do the clean ups once dIoU is merged.

@pmeier Any concerns/blocking changes on your side for tests?

pmeier (commented):
Minor style nit, but otherwise testing looks good to me minus the things we discussed for the follow up PR.

abhi-glitchhg:
Hey, @datumbox , what's the expected date for next release?

pmeier:
Tentative date for branch cut, i.e. "feature freeze", is on 19th of May. After that it will be roughly one month until release, but no official date is set yet.

PR #5777, opened 2022-04-07
update prototype dataset README

pmeier:
[Rendered version](https://github.com/pmeier/vision/blob/update-readme/torchvision/prototype/datasets/_builtin/README.md).

PR #5778, opened 2022-04-07
Refactor prototype datasets

pmeier:
This merges a feature branch into `main`. All changes up to #5777/ cd36d06a9db23909fcc0d0fae57a8597ffcd2641 were reviewed on individual PRs.

---

The `Dataset` class now inherits from `IterDataPipe` rather than being a custom class from which a datapipe is created. This has several upsides:

- The object can have a regular constructor with standard input checking. Before the inputs were checked automatically, but we needed hacky workarounds if not all combinations of parameters were valid. Compare 

    https://github.com/pytorch/vision/blob/5f74f0317e21f53a7ee5f61210e406c27f06a667/torchvision/prototype/datasets/_builtin/voc.py#L33-L43

    with

    https://github.com/pytorch/vision/blob/3be12c78105260a4949e35ccabbd538eaf9aa2c0/torchvision/prototype/datasets/_builtin/voc.py#L54-L55
- Instead of passing a `config` object around, the dataset can now simply use instance attributes set in the constructor. Compare

    https://github.com/pytorch/vision/blob/5f74f0317e21f53a7ee5f61210e406c27f06a667/torchvision/prototype/datasets/_builtin/cub200.py#L93-L98
    https://github.com/pytorch/vision/blob/5f74f0317e21f53a7ee5f61210e406c27f06a667/torchvision/prototype/datasets/_builtin/cub200.py#L174

    with

    https://github.com/pytorch/vision/blob/4c73a5efe03af5e18463c96a9a605a38951ddfbe/torchvision/prototype/datasets/_builtin/cub200.py#L116-L121
    https://github.com/pytorch/vision/blob/4c73a5efe03af5e18463c96a9a605a38951ddfbe/torchvision/prototype/datasets/_builtin/cub200.py#L192
- Although not yet added, we can now document the dataset with a regular docstring. This was not possible before, because the old `Dataset` object would have not been exposed and in turn would not show up in the documentation.
- We can use the native `__len__` method to return the number of samples in the dataset. In the old implementation we would have needed to attach a custom datapipe that does nothing but adds this annotation.


NicolasHug (approved):
Thanks a lot @pmeier 

The prototype job is green so it's safe to approve now, but to be extra careful let's wait until all jobs are green before merging.

The summary above is very useful, thank you. On top of what you already said, it's worth noting that even though this isn't reflected in this PR, the number of line of code has significantly reduced in our testing + base data structures, by ~500 (this PR reflects this more accurately: https://github.com/pytorch/vision/pull/5774). Thanks again for your help Philip

NicolasHug:
Also CCint @ejguan @NivekT @VitalyFedyunin ; Thanks to you guys' work on fixing https://github.com/pytorch/data/issues/237, we were able to greatly simplify our implementations, as described above. As you know we still need a full-fledged fix that also handles environments where `dill` is installed, but for now we're able to move forward with our preferred design. Thank you!

PR #5781, opened 2022-04-07
[proto] Added functional `crop_bounding_box` op

vfdev-5:
Related to #5514

## Description:
- Added functional `crop_bounding_box` op
- Added tests

Results match albumentations' [crop_bbox_by_coords](https://github.com/albumentations-team/albumentations/blob/89a675cbfb2b76f6be90e7049cd5211cb08169a5/albumentations/augmentations/crops/functional.py#L57)

### Results on synthetic images/bboxes:

<details>
<summary>
Code
</summary>

```python
import numpy as np

import torch
import torchvision
from torchvision.prototype.transforms.functional import crop_bounding_box, crop_image_tensor

size = (64, 76)
# xyxy format
in_boxes = [
    [10, 15, 25, 35],
    [50, 5, 70, 22],
    [45, 46, 56, 62],
]

for i, in_box in enumerate(in_boxes):
    im1[in_box[1]:in_box[3], in_box[0]:in_box[2], :] = (i * 40, 200 - i * 40, 127 + i * 40)

)
    
t, l, h, w = -8, 12, 80, 40


out_boxes = crop_bounding_box(
    in_boxes, 
    in_boxes.format,
    in_boxes.image_size,
    t,
    l,
    h,
    w,
)

t_im2 = crop_image_tensor(t_im1, t, l, h, w).clone()

import cv2
%matplotlib inline


plt.figure(figsize=(14, 8))

ax = plt.subplot(1,2,1)
for in_box in in_boxes:    
plt.imshow(r1)


ax = plt.subplot(1,2,2)
for out_box in out_boxes:
plt.imshow(r2)
```
</details>

image


torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -419,6 +419,32 @@ def pad_bounding_box(
...
+def crop_bounding_box(
+    bounding_box: torch.Tensor,
+    format: features.BoundingBoxFormat,
+    image_size: Tuple[int, int],
+    top: int,
+    left: int,
+    height: int,
+    width: int,
+) -> torch.Tensor:
+    pass
```
datumbox, line 431:
```suggestion
```

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -419,6 +419,32 @@ def pad_bounding_box(
...
+
+    shape = bounding_box.shape
+
+    bounding_box = convert_bounding_box_format(
+        bounding_box, old_format=format, new_format=features.BoundingBoxFormat.XYXY
+    ).view(-1, 4)
+
+    # Crop and optionally pad:
```
datumbox, line 439: What you mean optionally pad?

datumbox (commented):
Thanks @vfdev-5, just one question.

@pmeier Thoughts on the testing strategy?

torchvision/prototype/transforms/functional/_geometry.py
vfdev-5: Crop transform on images does explicitly padding if params suggest that. In case of cropping bboxes padding is implicit and we do not need to call `pad_bounding_box` for that. For example, if left is negative (=padding) then it will be implicitly done with the same operation as for cropping. Hope to be a bit more clear.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -419,6 +419,30 @@ def pad_bounding_box(
...
 
 
+def crop_bounding_box(
+    bounding_box: torch.Tensor,
+    format: features.BoundingBoxFormat,
+    image_size: Tuple[int, int],
+    top: int,
+    left: int,
+    height: int,
+    width: int,
```
pmeier, line 429: As an discussion opener: since these are unused, should we have them on the kernel at all?

pmeier (commented):
One comment inline. Testing looks ok.

torchvision/prototype/transforms/functional/_geometry.py
datumbox: I somehow missed this comment. This is a fair remark though I'm split on what we should do.

On one hand, passing unnecessary parameters creates noisy method signatures. On the other, the current proposal aligns the image kernel with the bbox in terms of parameters which is nice.

Assume we actually adopt the idea of putting kernel dispatcher on the object level (Image, Bbox, Mask). Then all these methods need to receive the same arguments (excluding meta-data that are provided directly from the class). But even in this case, the low-level kernels don't have to receive these parameters. I would be in favour of removing them from here but I want to hear both your thoughts.

vfdev-5: Yes, makes sense. Let's remove unused args

pmeier (approved):
Thanks Victor!

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -419,6 +419,27 @@ def pad_bounding_box(
...
+    left: int,
+) -> torch.Tensor:
+    shape = bounding_box.shape
+
+    bounding_box = convert_bounding_box_format(
+        bounding_box, old_format=format, new_format=features.BoundingBoxFormat.XYXY
+    ).view(-1, 4)
+
+    # Crop or implicit pad if left and/or top have negative values:
+    bounding_box[:, 0::2] -= left
```
vadimkantorov, line 435: For batch support it may be better to instead use `bounding_box[..., 0::2]`...

pmeier: Good catch. I was under the impression `...` was not supported by `torch.jit.script` in general, but this seems fine:

```py
@torch.jit.script
    y = x.clone()
    y[..., 0::2] += 1
    return y
```

It seems, only the explicit indices are not supported:

```py
@torch.jit.script
def bar(x: torch.Tensor) -> torch.Tensor:
    y = x.clone()
    y[..., [0, 2]] += 1
    return y
```

```
RuntimeError: 
[...]
def bar(x: torch.Tensor) -> torch.Tensor:
    y = x.clone()
    y[..., [0, 2]] += 1
    ~~~~~~~~~~~~~ <--- HERE
    return y
```

Issue #5688, opened 2022-03-28
Add tests for `sigmoid_focal_loss` and `generalized_box_iou_loss`


datumbox:
### 🚀 The feature

The `sigmoid_focal_loss` and `generalized_box_iou_loss` ops are not currently tested. We should create the necessary unit-tests that check the validity for multiple input types.

oke-aditya:
This will also help folks working on Ciou ad Diou loss. @YosuaMichael I can work on this if you like :) 

Reference tests fvcore
https://github.com/facebookresearch/fvcore/blob/main/tests/test_giou_loss.py

yassineAlouini:
Thanks @oke-aditya for the link.

YosuaMichael:
Hi @oke-aditya thanks for the offer! Is there anything urgent that need this test?

Currently I actually do this task for the sake of learning (since I am new and it would be good to touch as many part of the code as possible), hence I would like to do this issues myself if possible, however it would take more times.

YosuaMichael:
This issue is done with #5783 and #5792

PR #5783, opened 2022-04-07
Add test for sigmoid_focal_loss

YosuaMichael:
Sub-task for https://github.com/pytorch/vision/issues/5688

adding test for sigmoid_focal_loss function

### DIFF
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -9,6 +9,7 @@ import numpy as np
 import pytest
 import torch
 import torch.fx
+import torch.nn.functional as F
 from common_utils import assert_equal, cpu_and_gpu, needs_cuda
 from PIL import Image
 from torch import nn, Tensor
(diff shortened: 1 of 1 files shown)


test/test_ops.py
```diff
@@ -1450,5 +1451,140 @@ def test_is_leaf_node(self, dim, p, block_size, inplace):
...
+        return x
+
+    result = [
+        (device, dtype)
+        for device in ["cpu", "cuda"]
+        for dtype in [torch.float32, torch.half]
+        if (device, dtype) != ("cpu", torch.half)
+    ]
+    result = [augment_need_cuda_param(x) for x in result]
+    return result
```
pmeier, line 1470: Not sure about this. This adds a lot more complexity than

```py
@cpu_and_gpu
@pytest.mark.parametrize("dtype", [torch.float32, torch.half])
def test_foo(device, dtype):
    if device.type.startswith("cuda"):
        return
    ...
```

Was there a reason to not go this way?

test/test_ops.py
```diff
@@ -1450,5 +1451,140 @@ def test_is_leaf_node(self, dim, p, block_size, inplace):
...
+    def test_correct_ratio(self, alpha, gamma, device_dtype, seed) -> None:
+        # For testing the ratio with manual calculation, we require the reduction to be "none"
+        reduction = "none"
+        device, dtype = device_dtype
+        torch.random.manual_seed(seed)
+        inputs, targets = self._generate_diverse_input_target_pair(dtype=dtype, device=device)
+        focal_loss = ops.sigmoid_focal_loss(inputs, targets, gamma=gamma, alpha=alpha, reduction=reduction)
+        ce_loss = F.binary_cross_entropy_with_logits(inputs, targets, reduction=reduction)
+
+        assert torch.all(focal_loss <= ce_loss)
```
pmeier, line 1533: Should we have a proper error message here given that `pytest`'s introspection does not work well on tensors?

YosuaMichael (commented):
Thansk @pmeier , I have made some changes and respond on your review

test/test_ops.py
```diff
@@ -1450,5 +1451,140 @@ def test_is_leaf_node(self, dim, p, block_size, inplace):
...
+    @pytest.mark.parametrize("device_dtype", focal_loss_device_and_dtype())
+    @pytest.mark.parametrize("seed", [2, 3])
+    def test_equal_ce_loss(self, reduction, device_dtype, seed) -> None:
+        # focal loss should be equal ce_loss if alpha=-1 and gamma=0
+        alpha = -1
+        gamma = 0
+        device, dtype = device_dtype
+        torch.random.manual_seed(seed)
+        inputs, targets = self._generate_diverse_input_target_pair(dtype=dtype, device=device)
+        inputs_fl = inputs.clone().requires_grad_()
```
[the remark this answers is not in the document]
pmeier: I'm ok with both. Take your pick :slightly_smiling_face:

test/test_ops.py
```diff
@@ -1450,5 +1451,117 @@ def test_is_leaf_node(self, dim, p, block_size, inplace):
         assert len(graph_node_names[0]) == 1 + op_obj.n_inputs
 
 
+class TestFocalLoss:
+    def _logit(self, p: Tensor) -> Tensor:
+        return torch.log(p / (1 - p))
+
+    def _generate_tensor_with_range_type(self, shape, range_type, **kwargs):
```
pmeier, line 1458: Another small nit: given that these two methods are only used from within `_generate_diverse_input_target_pair` can we maybe inline them there?

YosuaMichael: I see, I guess having them inline make it more explicit to the reader that it is only used on `generate_diverse_input_target_pair`. Will do this.

test/test_ops.py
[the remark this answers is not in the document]
YosuaMichael: In that case I will just leave it as it is then

pmeier (approved):
Thanks Yosua, LGTM if CI is green!

test/test_ops.py
```diff
@@ -1557,7 +1557,10 @@ def test_jit(self, alpha, gamma, reduction, device, dtype, seed) -> None:
         torch.random.manual_seed(seed)
         inputs, targets = self._generate_diverse_input_target_pair(dtype=dtype, device=device)
         focal_loss = ops.sigmoid_focal_loss(inputs, targets, gamma=gamma, alpha=alpha, reduction=reduction)
-        scripted_focal_loss = script_fn(inputs, targets, gamma=gamma, alpha=alpha, reduction=reduction)
+        with torch.jit.fuser("fuser2"):
+            # Use fuser2 to prevent the bug from fuser that will be triggered in following condition:
+            # dtype=torch.half, device="cuda"
```
datumbox: We should probably put the issue with the bug here in the comments. If there is no such issue, we should raise one. Let's see what the JIT team responds to us and take the necessary action. :)

It's also worth noting that JIT tests are not necessary for losses as we aim to support scriptability for inference. If this test doesn't produce a major slow-down, we can keep. No strong opinions on my side.

datumbox (approved):

PR #5785, opened 2022-04-07
add warning if font is not set

### LINKED ISSUE #5725: torchvision.utils.draw_bounding_boxes set font_size
### 🚀 The feature

font_ The size setting must be specified at the same time as the font to be valid, otherwise it is invalid.
I think utils.py in torchvision Line 239 can be improved.Make font settings and font size settings separate.
utils.py Line 239:
"txt_font = ImageFont.load_default() if **font is None** else ImageFont.truetype(font=font, size=font_size)"

### Motivation, pitch

Improve the experience of torchvision.draw_bounding_boxes

### Alternatives

_No response_

### Additional context

_No response_

lezwon:
Fixes #5725


torchvision/utils.py
```diff
@@ -223,6 +223,9 @@ def draw_bounding_boxes(
 
     colors = [(ImageColor.getrgb(color) if isinstance(color, str) else color) for color in colors]
 
+    if font is None and font_size != 10:
```
lezwon, line 226: I have a minor confusion here. The default argument for `font_size` is set to 10. How do I check if it's the user set or the default value?

cc @oke-aditya @pmeier

pmeier: I would change the default to `None`. With that you can do something like

```py
if font is None and font_size is not None:
    warnings.warn(...)
    font_size = None
elif font is not None and font_size is None:
    font_size = 10
```

test/test_utils.py
```diff
@@ -168,6 +168,15 @@ def test_draw_invalid_boxes():
         utils.draw_bounding_boxes(img_correct, boxes, colors=colors_wrong)
 
 
+def test_draw_boxes_warning():
+    img = torch.full((3, 100, 100), 255, dtype=torch.uint8)
+    labels = ["a", "b", "c", "d"]
+    colors = ["green", "#FF00FF", (0, 255, 0), "red"]
```
pmeier, line 174: Neither of those is required by the function or required to trigger the warning, right? If so, we can simplify the test.

torchvision/utils.py
```diff
@@ -223,6 +223,12 @@ def draw_bounding_boxes(
 
     colors = [(ImageColor.getrgb(color) if isinstance(color, str) else color) for color in colors]
 
+    if font is None and font_size is not None:
+        warnings.warn("Argument 'font_size' will be unused as 'font' is not set.")
```
pmeier, line 227: Nit

```suggestion
        warnings.warn("Argument 'font_size' will be ignored since 'font' is not set.")
```

pmeier (commented):
Two minor comments. Otherwise LGTM.

test/test_utils.py
```diff
@@ -168,6 +168,13 @@ def test_draw_invalid_boxes():
         utils.draw_bounding_boxes(img_correct, boxes, colors=colors_wrong)
 
 
+def test_draw_boxes_warning():
+    img = torch.full((3, 100, 100), 255, dtype=torch.uint8)
+
+    with pytest.warns(UserWarning, match="Argument 'font_size' will be unused as 'font' is not set."):
```
pmeier, line 174: The string passed to `match` is actually a regex pattern. Thus, if you have special regex characters like `.` in it, you should wrap it in `re.escape` to mark them as "regular". For example

```py
>>> import re
>>> match = "'font' is not set."
>>> re.escape(match)
"'font'\\ is\\ not\\ set\\."
```

```suggestion
    with pytest.warns(UserWarning, match=re.escape("Argument 'font_size' will be ignored since 'font' is not set.")):
```

Note that you will have to `import re` at the top of the module.

lezwon: Noted. Will add the change. 👍

test/test_utils.py
```diff
@@ -168,6 +169,13 @@ def test_draw_invalid_boxes():
         utils.draw_bounding_boxes(img_correct, boxes, colors=colors_wrong)
 
 
+def test_draw_boxes_warning():
+    img = torch.full((3, 100, 100), 255, dtype=torch.uint8)
+
+    with pytest.warns(UserWarning, match=re.escape("Argument 'font_size' will be unused as 'font' is not set.")):
+        utils.draw_bounding_boxes(img, boxes, fill=True, font_size=11)
```
pmeier, line 176: Maybe I'm missing something here, but I don't see `boxes` defined anywhere. Is this test actually passing for you locally? You can run it with `pytest test/test_utils.py::test_draw_boxes_warning`.

pmeier: Plus, I don't think we need the `fill` argument here:

```suggestion
        utils.draw_bounding_boxes(img, boxes, font_size=11)
```

lezwon: Yep. The test passes locally. `boxes` is defined in the global scope at the top of the file.

test/test_utils.py
```diff
@@ -168,6 +169,13 @@ def test_draw_invalid_boxes():
         utils.draw_bounding_boxes(img_correct, boxes, colors=colors_wrong)
 
 
+def test_draw_boxes_warning():
+    img = torch.full((3, 100, 100), 255, dtype=torch.uint8)
+
+    with pytest.warns(UserWarning, match=re.escape("Argument 'font_size' will be unused as 'font' is not set.")):
```
NicolasHug, line 175: I don't think we need `re.escape`?

pmeier: Resolving this after offline discussion. We technically don't need it here, since the only special regex character here is the `.` at the end. Since it matches anything, the test would also pass without escaping the string.

IMO, it is still useful here to communicate the intent. Using it makes it clear that we want an exact match. Whether or not we should want this in the first place is a different discussion.

test/test_utils.py
pmeier: @lezwon You forgot to update the warning message to match the one that is emitted from the function.

lezwon: Sorry. Missed noticing this somehow. Have updated it now.

torchvision/utils.py
```diff
@@ -223,6 +223,12 @@ def draw_bounding_boxes(
 
     colors = [(ImageColor.getrgb(color) if isinstance(color, str) else color) for color in colors]
 
+    if font is None and font_size is not None:
+        warnings.warn("Argument 'font_size' will be ignored since 'font' is not set.")
+        font_size = None
+    elif font is not None and font_size is None:
+        font_size = 10
+
```
oke-aditya, line 231: Will it be nice to club this code along with line 245? So that we handle font and font size in single place?

No strong opinion. 
What do you both think?
cc @pmeier

torchvision/utils.py
```diff
@@ -223,6 +223,12 @@ def draw_bounding_boxes(
 
     colors = [(ImageColor.getrgb(color) if isinstance(color, str) else color) for color in colors]
 
+    if font is None and font_size is not None:
+        warnings.warn("Argument 'font_size' will be ignored since 'font' is not set.")
+        font_size = None
+    elif font is not None and font_size is None:
+        font_size = 10
```
pmeier, line 230: Following @oke-aditya's suggestion in https://github.com/pytorch/vision/pull/5785/files#r849794680. Please also remove L245.

```suggestion
    if font is None:
        if font_size is not None:
            warnings.warn("Argument 'font_size' will be ignored since 'font' is not set.")
        txt_font = ImageFont.load_default()
    else:
        txt_font = ImageFont.truetype(font=font, size=font_size or 10)
```

torchvision/utils.py
pmeier: Yes, that is a really good idea. Added a suggestion for this in https://github.com/pytorch/vision/pull/5785/files#r850126056.

torchvision/utils.py
oke-aditya: I had same thing in my mind :smile: Makes the code more readable

lezwon: Done. Have updated the code accordingly.

torchvision/utils.py
```diff
@@ -223,6 +223,13 @@ def draw_bounding_boxes(
 
     colors = [(ImageColor.getrgb(color) if isinstance(color, str) else color) for color in colors]
 
+    if font is None:
+        if font_size is not None:
+            warnings.warn("Argument 'font_size' will be ignored since 'font' is not set.")
+        ImageFont.load_default()
+    else:
+        ImageFont.truetype(font=font, size=font_size or 10)
```
pmeier, line 231:
```suggestion
        txt_font = ImageFont.load_default()
    else:
        txt_font = ImageFont.truetype(font=font, size=font_size or 10)
```

pmeier (approved):
Thanks @lezwon, LGTM!

datumbox (approved):
LGTM, thanks!

PR #5786, opened 2022-04-07
Distance IoU 

yassineAlouini:
In this PR, I implement the distance IoU and associated loss as described from this paper: https://arxiv.org/abs/1911.08287.
The implementation is inspired and adapted from here     https://github.com/facebookresearch/detectron2/blob/dfe8d368c8b7cc2be42c5c3faf9bdcc3c08257b1/detectron2/layers/losses.py#L5 and from the implemented giou.


yassineAlouini:
I have to improve the tests a bit. I have also some refactoring ideas in mind but it is better to leave those for a new PR.

yassineAlouini:
> @yassineAlouini Thanks for the contribution. I know your PR is still draft but it's worth looking at the comments at #5776 since many of them are relevant to you too.

Indeed. I am already using some of the first feedbacks and I will look for new ones. :ok_hand:

torchvision/ops/__init__.py
```diff
@@ -7,6 +7,7 @@
     box_area,
     box_iou,
     generalized_box_iou,
+    distance_box_iou,
```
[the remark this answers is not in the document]
yassineAlouini: I am missing something maybe but it is imported below, isn't it? :thinking:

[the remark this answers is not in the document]
oke-aditya: hmm no, I don't think so. I think you have imported it in docs/ops.rst but not in init here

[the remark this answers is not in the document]
yassineAlouini: It is imported below:

`from .diou_loss import distance_box_iou_loss`

It isn't available in the boxes file. Sorry if I am misunderstanding. :grimacing:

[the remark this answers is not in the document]
oke-aditya: oh yes! I see. My bad. Somehow I got confused :)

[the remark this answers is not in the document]
yassineAlouini: No worries @oke-aditya.

torchvision/ops/giou_loss.py
```diff
@@ -36,7 +36,7 @@ def generalized_box_iou_loss(
             ``'none'`` | ``'mean'`` | ``'sum'``. ``'none'``: No reduction will be
             applied to the output. ``'mean'``: The output will be averaged.
             ``'sum'``: The output will be summed. Default: ``'none'``
-        eps (float, optional): small number to prevent division by zero. Default: 1e-7
+        eps (float): small number to prevent division by zero. Default: 1e-7
```
abhi-glitchhg, line 39: Similar to `eps`,  doc-string of `reduction`  parameter contains unnecessary  `optional`. I think you should remove that as well.

yassineAlouini: Noticed, thanks for the pointer. However, I didn't write the `giou_loss.py` file. :smile:

abhi-glitchhg: yes, I know 😄. 

BTW  I think it's okay if we don't remove the `optional`  from the above docstrings.  Refer - https://github.com/pytorch/vision/pull/5799#discussion_r847518711

yassineAlouini: I see, thanks for the link @abhi-glitchhg. Indeed, here the `eps` argument has a default value so the optional makes some sense (even though it is better to use it for the None case).

yassineAlouini:
@oke-aditya @abhi-glitchhg I have added more tests highly inspired from cIoU and fixed the ones I have added already.
I think it is a good time to give this PR reviews if you can and I will then take into account your suggestions. Thanks in advance. :+1:

test/test_ops.py
```diff
@@ -1127,6 +1127,8 @@ def _perform_box_operation(self, box: Tensor, run_as_script: bool = False) -> Te
     def _run_test(self, test_input: List, dtypes: List[torch.dtype], tolerance: float, expected: List) -> None:
         def assert_close(box: Tensor, expected: Tensor, tolerance):
             out = self._perform_box_operation(box)
+            print("The computed box is: ", out)
```
oke-aditya, line 1130: can we remove printing here? We don't usually print in tests.

yassineAlouini: Oh yes, my bad, I used it to debug. Will remove, thanks for catching this.

test/test_ops.py
```diff
@@ -1257,6 +1259,84 @@ def test_gen_iou(self, test_input: List, dtypes: List[torch.dtype], tolerance: f
...
+class TestDistanceIoULoss:
+    # Inspired and adapted from:
+    # https://github.com/pytorch/vision/pull/5776/files#diff-d183f2afc51d6a59bc70094e8f476d2468c45e415500f6eb60abad955e065156
+
+    @staticmethod
+    def assert_distance_iou_loss(box1, box2, expected_output, dtype, reduction="none"):
+        output = ops.distance_box_iou_loss(box1, box2, reduction=reduction)
+        expected_output = torch.tensor(expected_output, dtype=dtype)
+        tol = 1e-5 if dtype != torch.half else 1e-3
+        torch.testing.assert_close(output, expected_output, rtol=tol, atol=tol)
```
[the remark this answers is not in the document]
yassineAlouini: Yes, that's a good idea. I see that it has been added in the cIoU tests as well. :ok_hand:

test/test_ops.py
```diff
@@ -1258,6 +1258,85 @@ def test_giou_jit(self) -> None:
...
+
+
+class TestDistanceIoULoss:
+    # Inspired and adapted from:
+    # https://github.com/pytorch/vision/pull/5776/files#diff-d183f2afc51d6a59bc70094e8f476d2468c45e415500f6eb60abad955e065156
+
+    @staticmethod
+    def assert_distance_iou_loss(box1, box2, expected_output, dtype, device, reduction="none"):
+        output = ops.distance_box_iou_loss(box1, box2, reduction=reduction)
+        expected_output = torch.tensor(expected_output, dtype=dtype, device=device)
```
oke-aditya, line 1306: I think this will automatically take the dtype, by passing dtype as function arg. It will also get parameterized,

```suggestion
        expected_output = torch.tensor(expected_output, device=device)
```


Notice that

https://github.com/pytorch/vision/pull/5792/files#diff-d183f2afc51d6a59bc70094e8f476d2468c45e415500f6eb60abad955e065156R1587

this works!

test/test_ops.py
```diff
@@ -1258,6 +1258,85 @@ def test_giou_jit(self) -> None:
         self._run_jit_test([[0, 0, 100, 100], [0, 0, 50, 50], [200, 200, 300, 300]])
 
 
+class TestDistanceBoxIoU(BoxTestBase):
+    def _target_fn(self) -> Tuple[bool, Callable]:
```
oke-aditya, line 1262: A general question to torchvision maintainers
Do we type hint in tests? 
It's not bad to type hint. But it isn't something which we follow either 😅


cc @pmeier @datumbox

pmeier: No, we don't. See https://github.com/pytorch/vision/pull/5563#discussion_r824758253. The reason is that without also checking the tests with `mypy` they might go out of date and that is usually more harmful than not having them at all.

test/test_ops.py
yassineAlouini: Not exactly the same. I made the nested function into a staticmethod so it doesn't have access to the same scope.

test/test_ops.py
yassineAlouini: I am removing the type hints then, thanks for the link @pmeier. :+1:

yassineAlouini:
@oke-aditya @abhi-glitchhg @pmeier I have a question regarding float16 input to the distance_box_iou_loss function:
is it required to also have the output as a float16 or is float32 enough? 

If float16 is required, what is the best way to achieve this? Is disabling the autocast a good thing to do or is there a better way. Thanks for any help.

pmeier:
> is it required to also have the output as a `float16` or is `float32` enough?

In general, the loss needs to preserve the input dtype.

> If `float16` is required, what is the best way to achieve this? Is disabling the autocast a good thing to do or is there a better way.

Could you be more specific where the "autocast" happens? By default, most PyTorch ops preserve the `dtype`.

yassineAlouini:
> Could you be more specific where the "autocast" happens? By default, most PyTorch ops preserve the `dtype`.

I will give this a look later and let you know (might take some time since I am new to autocasting but will be a useful exercise to learn more).) :ok_hand:

datumbox:
@yassineAlouini Thanks for the contribution! 

I've added a comment on the cIoU PR which might be applicable here as well: https://github.com/pytorch/vision/pull/5776#issuecomment-1102355838

yassineAlouini:
@datumbox @abhi-glitchhg @oke-aditya Ready for a new round of code review if you are available, thanks in advance. :)

test/test_ops.py
```diff
@@ -1258,6 +1258,83 @@ def test_giou_jit(self) -> None:
         self._run_jit_test([[0, 0, 100, 100], [0, 0, 50, 50], [200, 200, 300, 300]])
 
 
+class TestDistanceBoxIoU(BoxTestBase):
+    def _target_fn(self):
+        return (True, ops.distance_box_iou)
+
+    def _generate_int_input():
```
pmeier, line 1265: This has nothing to do with this PR, but I want make this visible. Although this works, there are two issues with this structure introduced in #5380 cc @datumbox:

1. We shouldn't have regular methods without `self` as first parameter. Given that we don't need it here, we should use the `@staticmethod` decorator.
2. It makes little sense to separate the input and expected values into two methods given that they are always used together. This just makes it harder to read.

test/test_ops.py
```diff
@@ -1258,6 +1258,83 @@ def test_giou_jit(self) -> None:
...
+    def test_distance_iou(self, test_input, dtypes, tolerance, expected):
+        self._run_test(test_input, dtypes, tolerance, expected)
+
+    def test_distance_iou_jit(self):
+        self._run_jit_test([[0, 0, 100, 100], [0, 0, 50, 50], [200, 200, 300, 300]])
+
+
+class TestDistanceIoULoss:
+    # Inspired and adapted from:
+    # https://github.com/pytorch/vision/pull/5776/files#diff-d183f2afc51d6a59bc70094e8f476d2468c45e415500f6eb60abad955e065156
```
pmeier, line 1300: This link no longer works or better no longer points to a comment.

torchvision/ops/boxes.py
```diff
@@ -311,6 +311,50 @@ def generalized_box_iou(boxes1: Tensor, boxes2: Tensor) -> Tensor:
     return iou - (areai - union) / areai
 
 
+# Implementation inspired from the generalized_box_iou one.
+# TODO: Some refactoring and homogenization could be done with
+# the loss function in diou_loss.
```
pmeier, line 316: According to https://github.com/pytorch/vision/pull/5776#issuecomment-1102613291 we should do that here, correct?

torchvision/ops/diou_loss.py
```diff
@@ -0,0 +1,51 @@
...
+        diou = distance_box_iou(batch_boxes1, batch_boxes2, eps)[0, 0]
+    else:
+        diou = distance_box_iou(boxes1, boxes2, eps)[0]
+    loss = 1 - diou
+    if reduction == "mean":
+        loss = loss.mean() if loss.numel() > 0 else 0.0 * loss.sum()
+    elif reduction == "sum":
+        loss = loss.sum()
+    # Cast the loss to the same dtype as the input boxes
+    loss = loss.to(boxes1.dtype)
```
pmeier, line 50: Why is this needed? Shouldn't all operations preserve the dtype anyway?

pmeier, line 50: Why is this needed? Shouldn't all operations preserve the dtype anyway?

datumbox: Agreed this shouldn't be needed. 

I believe the casting here happens to return the loss on the same dtype as the original boxes (due to upcastings) but I don't think we should do this.

datumbox (commented):
Thanks for the contribution @yassineAlouini. Looks on the right direction. Just a couple of comments below.

torchvision/ops/diou_loss.py
yassineAlouini: I agree that it shouldn't be necessary but I had to do it to preserve the "half" dtype. Not sure what casting happens so that it is turned into float. I will give this another look and let you know if I need help.

torchvision/ops/boxes.py
```diff
@@ -311,6 +311,50 @@ def generalized_box_iou(boxes1: Tensor, boxes2: Tensor) -> Tensor:
...
+    iou = inter / union
+
+    lti = torch.min(boxes1[:, None, :2], boxes2[:, None, :2])
+    rbi = torch.max(boxes1[:, None, 2:], boxes2[:, None, 2:])
+
+    whi = _upcast(rbi - lti).clamp(min=0)  # [N,M,2]
+    diagonal_distance_squared = (whi[:, :, 0] ** 2) + (whi[:, :, 1] ** 2) + eps
+
+    # centers of boxes
+    x_p = boxes1[:, None, :2].sum() / 2
```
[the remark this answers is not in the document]
yassineAlouini: That's a great suggestion, will add the `_upcast` operartion. :ok_hand:

test/test_ops.py
```diff
@@ -1258,6 +1258,83 @@ def test_giou_jit(self) -> None:
...
+        ],
+    )
+    def test_distance_iou(self, test_input, dtypes, tolerance, expected):
+        self._run_test(test_input, dtypes, tolerance, expected)
+
+    def test_distance_iou_jit(self):
+        self._run_jit_test([[0, 0, 100, 100], [0, 0, 50, 50], [200, 200, 300, 300]])
+
+
+class TestDistanceIoULoss:
```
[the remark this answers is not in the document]
yassineAlouini: That's a good point. :ok_hand:

torchvision/ops/boxes.py
yassineAlouini: Yes, will be removed. :ok_hand:

torchvision/ops/diou_loss.py
```diff
@@ -0,0 +1,51 @@
...
+    Reference:
+        Zhaohui Zheng et. al: Distance Intersection over Union Loss:
+        https://arxiv.org/abs/1911.08287
+    """
+    if boxes1.dim() == 1 and boxes2.dim() == 1:
+        batch_boxes1 = boxes1.unsqueeze(0)
+        batch_boxes2 = boxes2.unsqueeze(0)
+        diou = distance_box_iou(batch_boxes1, batch_boxes2, eps)[0, 0]
+    else:
+        diou = distance_box_iou(boxes1, boxes2, eps)[0]
```
datumbox, line 43: @yassineAlouini I'm not sure this approach is equivalent to what we had earlier. Please correct me if I'm wrong but I understand that `distance_box_iou()` does NxM pairwise comparisons while here we just want the Nx1 comparisons. This can be given by using the diagonal but this should be extremely slow as we would be throwing away the majority of operations. 

What I had in mind is try to refactor the code at `ops.boxes` so that we can share some of the estimations. Thoughts?

cc @abhi-glitchhg because you follow a similar approach on the other PR.

datumbox: BTW if you both prefer to revert to your earlier versions of the code (which didn't reuse ops.boxes) and tackle this on separate future PRs, I'm happy to go down that route. The PRs for cIoU and dIoU has been dragging for a while and I appreciate that this can become frustrating at one point. Let me know your preference so that we make this a more enjoyable experience for both of you. Thanks a bunch for your work so far. :)

abhi-glitchhg: W@datumbox , I agree with your concerns. This is not computationally efficient.

abhi-glitchhg: > BTW if you both prefer to revert to your earlier versions of the code (which didn't reuse ops.boxes) and tackle this on separate future PRs,

Yeah, sounds good. :)

yassineAlouini: > @yassineAlouini I'm not sure this approach is equivalent to what we had earlier. Please correct me if I'm wrong but I understand that `distance_box_iou()` does NxM pairwise comparisons while here we just want the Nx1 comparisons. This can be given by using the diagonal but this should be extremely slow as we would be throwing away the majority of operations.
> 
> What I had in mind is try to refactor the code at `ops.boxes` so that we can share some of the estimations. Thoughts?
> 
> cc @abhi-glitchhg because you follow a similar approach on the other PR.

That's a very good point and I might have introduced a bug by going quickly on my refactor, so thanks for pointing this out.

I can revert to previous code or keep working on this here (in this PR), both work for me. :ok_hand:

datumbox: Thanks for the flexibility! Let's revert and use the previously vetted code on the loss. We can investigate refactoring all losses to share code with `ops.boxes` on separate PRs as you originally suggested.

test/test_ops.py
abhi-glitchhg: +1 @yassineAlouini , Even I was thinking about this.

torchvision/ops/diou_loss.py
datumbox: @yassineAlouini From what I see you've reverted the code that reused estimations from `ops.boxes` and now use the a modified version of the loss at [Detectron2](https://github.com/facebookresearch/detectron2/blob/dfe8d368c8b7cc2be42c5c3faf9bdcc3c08257b1/detectron2/layers/losses.py#L5). BTW we should add a reference in the source code similar to [this](https://github.com/pytorch/vision/blob/main/torchvision/ops/focal_loss.py#L34), to indicate the original source.

Concerning the casting question, note that in `ops.boxes` the [_upcast method](https://github.com/pytorch/vision/blob/main/torchvision/ops/boxes.py#L218-L223) is primarily there to handle overflows of low precision values that might overflow. The main concern is integers (because those methods need to support them). In the case of losses, the [_upcast method](https://github.com/pytorch/vision/blob/main/torchvision/ops/giou_loss.py#L7-L11) works a bit different as it converts everything to floats (we don't handle integers in losses). This is crucial because some of the area estimation operations will overflow in `torch.float16`.

```
>>> torch.finfo(torch.float16)
finfo(resolution=0.001, min=-65504, max=65504, eps=0.000976562, tiny=6.10352e-05, dtype=float16)
```

So I think the safe thing to do here is to follow the same approach as in gIoU and upcast. I think we should move the method `_upcast` to `_utils` so it can be shared. This is going to be useful also for the cIoU PR that @abhi-glitchhg is working on.

cc @fmassa for visibility in case I stated something incorrect here.

torchvision/ops/boxes.py
```diff
@@ -311,6 +311,50 @@ def generalized_box_iou(boxes1: Tensor, boxes2: Tensor) -> Tensor:
...
+    rbi = torch.max(boxes1[:, None, 2:], boxes2[:, 2:])
+
+    whi = _upcast(rbi - lti).clamp(min=0)  # [N,M,2]
+    diagonal_distance_squared = (whi[:, :, 0] ** 2) + (whi[:, :, 1] ** 2) + eps
+
+    # centers of boxes
+    x_p = boxes1[:, None, :2].sum() / 2
+    y_p = boxes1[:, None, 2:].sum() / 2
+    x_g = boxes2[:, :2].sum() / 2
+    y_g = boxes2[:, 2:].sum() / 2
```
abhi-glitchhg, line 349: Hey @yassineAlouini , I think there is a problem with this implementation.  The calculation of centre of boxes is not correct acc to me. We should be adding up only `x1` `x2` and `y1` `y2`, [ref](https://github.com/facebookresearch/detectron2/blob/dfe8d368c8b7cc2be42c5c3faf9bdcc3c08257b1/detectron2/layers/losses.py#L49-L53) . 
But in  current implementation, we are adding `x1`,`y1` and `x2`, `y2`. (BBox shape is in form [x1,y1,x2,y2]) .

This can also be checked by calculating `distance_box_iou_loss` and `distance_box_iou` on a sample tensors. 


    import torch
    from torchvision.ops import distance_box_iou, distance_box_iou_loss

    box1 = torch.tensor([[-1, -1, 1, 1]], )
    box2 = torch.tensor([[0, 0, 1, 1]],)

    1-distance_box_iou(box1, box2)[0] == distance_box_iou_loss(box1, box2)


Last statement returns `False`. Ideally it should return `True`.

I suggest you to do following changes. 

```suggestion
    # centers of boxes
    x_p = (boxes1[:, 0] + boxes1[:, 2]) / 2
    y_p = (boxes1[:, 1] + boxes1[:, 3]) / 2
    x_g = (boxes2[:, 0] + boxes2[:, 2]) / 2
    y_g = (boxes2[:, 1] + boxes2[:, 3]) / 2
```

@datumbox,
Please correct me if I'm wrong.
Thanks.

test/test_ops.py
pmeier: > I guess we should keep this for another PR?

:+1:

torchvision/ops/boxes.py
datumbox: Good catch. I haven't yet reviewed the correctness of the implementation as we still discuss the structure/API. 

I think that's probably a typo and @yassineAlouini intended to write something like:
```
x_p = boxes1[:, 0::2].sum() / 2
y_p = boxes1[:, 1::2].sum() / 2
...
```

test/test_ops.py
datumbox: Is the suggested plan to fix this on a separate PR across all tests? Just trying to figure out if this is blocking of this PR and whether we need to take similar action at #5776. @pmeier it's your call.

pmeier: No this is not blocking for this PR. I would appreciate if either @yassineAlouini or  @abhi-glitchhg could send a follow-up PR fixing this. Whoever wants to pick this up, please comment here to avoid duplicate work and resolve this comment afterwards.

abhi-glitchhg: Hey @pmeier, sorry :(  
I have uni exams starting next week, so I will not be able to devote much time to this.

yassineAlouini:
@datumbox I see that the cIoU metric has been merged. I will take inspiration to fix remaining issues today hopefully. :ok_hand:

torchvision/ops/boxes.py
yassineAlouini: Yes indeed, I think I went too quickly over this and thought that the bounding box was in the `x1x2y1y2` format. Thanks for pointing this out and your suggestions. :+1:

test/test_ops.py
pmeier: No worries @abhi-glitchhg. Good luck with that. @yassineAlouini would you be able to pick this up? Otherwise, I'll open an issue so we don't forget and we can do this later.

yassineAlouini: @pmeier Yes, I have time this afternoon so will take this on the next PR. :ok_hand:

yassineAlouini:
@datumbox @pmeier I made the code iso with the `cIoU` one. However, I have noticed two  things:

* some imports are missing in the `__all__` for `cIoU` and have fixed them.
* the torch.half nightmarish test :grin: is still not working. I thought that the `cIoU` code found a clever fix but it seems that it is missing the `dtype` setup (I have put a TODO so that we don't forget it) in the appropriate test and so once I add it, it fails as well. I think it is best to fix this in the next PR (where I will refactor some of the tests).

test/test_ops.py
yassineAlouini: And good luck @abhi-glitchhg in your exams. :+1:

datumbox (approved):
@yassineAlouini Apologies for the delay. I was OOO last week.

I've reviewed the code of the implementation and it looks good to me! There was a code attribution missing but I pushed directly into your branch to avoid the back and forth.

In addition to the review, I verified that the two implementations for boxes and losses return the same result by running:
```python
import torch

from torchvision.ops.boxes import distance_box_iou
from torchvision.ops.diou_loss import distance_box_iou_loss

def random_box(canvas_size):
    x1y1 = torch.rand((1, 2)) * canvas_size
    wh = torch.rand((1, 2)) * canvas_size
    x2y2 = (x1y1 + wh).clamp(0, canvas_size)
    return torch.cat((x1y1, x2y2), axis=1)


canvas_size = 1000


for _ in range(10000):
    box1 = random_box(canvas_size)
    box2 = random_box(canvas_size)

    v1 = 1 - distance_box_iou(box1, box2)
    v2 = distance_box_iou_loss(box1, box2)

    torch.testing.assert_close(v1[0], v2, rtol=0, atol=1e-6)
```

@pmeier Any blocking changes required on the testing side? If not I recommend merging and do a deep cleaning between the g/c/d-iou. 

@yassineAlouini @abhi-glitchhg @oke-aditya Anyone interested in doing this cleanup?

test/test_ops.py
```diff
@@ -1676,6 +1767,7 @@ def test_ciou_loss(self, dtype, device):
         def assert_ciou_loss(box1, box2, expected_output, reduction="none"):
 
             output = ops.complete_box_iou_loss(box1, box2, reduction=reduction)
+            # TODO: When passing the dtype, the torch.half test doesn't pass...
```
oke-aditya, line 1770: Is it still valid?

torchvision/ops/diou_loss.py
```diff
@@ -0,0 +1,86 @@
+import torch
+
+from ..utils import _log_api_usage_once
+from .boxes import _upcast
+
+
```
oke-aditya, line 6: Commenting above the file might be better. As comments in the function call will be a small execution of commented code everytime we call code? (Is there a subtle performance difference? Not sure but always had this doubt)

torchvision/ops/diou_loss.py
```diff
@@ -0,0 +1,86 @@
...
+    Returns:
+        Tensor: Loss tensor with the reduction option applied.
+
+    Reference:
+        Zhaohui Zheng et. al: Distance Intersection over Union Loss:
+        https://arxiv.org/abs/1911.08287
+    """
+
+    # Original Implementation : https://github.com/facebookresearch/detectron2/blob/main/detectron2/layers/losses.py
```
oke-aditya, line 40: I meant this comment

test/test_ops.py
datumbox: Can we provide a bit more info on what doesn't pass here and what's exactly the issue?

pmeier (approved):
Testing LGTM, thanks @yassineAlouini!

oke-aditya:
@yassineAlouini  Feel free to review #5979

test/test_ops.py
yassineAlouini: @datumbox I read the cIoU code since it was passing the torch.half tests and I found out that the dtype wasn't passed so the test wasn't correct for torch.half. For now, I have removed the dtype to have the same code as cIoU but I think we should investigate this further (or maybe we can't do anything since we use the _upcast function? 🤔). Let me know if this clear enough. I can provide more details.

yassineAlouini:
> @yassineAlouini Feel free to review #5979

I will, thanks for the suggestion @oke-aditya. :ok_hand:

PR #5789, opened 2022-04-08
fix caltech dataset downloads

### LINKED ISSUE #5737: Caltech download is broken
Oh no, something went wrong in the scheduled workflow tests/download. 
Please look into it:

https://github.com/pytorch/vision/actions/runs/2095101012

Feel free to close this if this was just a one-off error.

cc @pmeier @YosuaMichael

pmeier:
Fixes #5737. Their [website](http://www.vision.caltech.edu/datasets/) features this message:

caltech

The files are now hosted on [GDrive](https://drive.google.com/drive/folders/1cnQHqa8BkVx90-6-UojHnbMB0WhksSRc) linked under "Some of our datasets are available here.". The links at the bottom for Caltech 101 / 256 seem to feature the same data, but packaged in a different way. Thus, I've opted to just switch to GDrive as source.

I've confirmed that the download now works again and that the checksums indeed match the ones we have on record.

### DIFF
--- a/torchvision/datasets/caltech.py
+++ b/torchvision/datasets/caltech.py
@@ -129,13 +129,15 @@ class Caltech101(VisionDataset):
             return
 
         download_and_extract_archive(
-            "http://www.vision.caltech.edu/Image_Datasets/Caltech101/101_ObjectCategories.tar.gz",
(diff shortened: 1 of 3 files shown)


NicolasHug (approved):
Thanks @pmeier

Issue #4559, opened 2021-10-07
Evaluation code of references is slightly off


NicolasHug:
There is a subtle known bug in the evaluation code of the classification references (and other references as well, but not all):

https://github.com/pytorch/vision/blob/261cbf7e939a253ea316b74bec8dbba58155ab4e/references/classification/train.py#L65-L66

It deserves some attention, because it's easy to miss and yet can **impact our reported results, and those of research papers.**

As the comment above describes, when computing the accuracy of the model on a validation set in a distributed setting, some images will be counted more than once if `len(dataset)` isn't divisible by `batch_size * world_size` [^bignote].

On top of that, since the `test_sampler` uses `shuffle=True` by default, the duplicated images aren't even the same across executions, which means that evaluating the same model on the same dataset can lead to different results every time.

Should we try to fix this, or should we just leave it and wait for the new lightning recipes to handle it? And as a follow-up question, is there a builtin way in lightning to mitigate this at all? (I'm not familiar with lightning, so this one may not make sense.)

[^bignote]: For example if we have 10 images and 2 workers with a batch_size of 3, we will have something like:

    ```
    worker1: img1, img2, img3
    worker2: img4, img5, img6
    worker1: img7, img8, img9
    worker2: img10, **img1, img2** 
                      ^^^^^^^^^
     "padding": duplicated images which will affect the validation accuracy
     ```

cc @datumbox

datumbox:
Thanks for raising. 

Here are some thoughts:
1. Reporting slightly biased val stats during the training process is less concerning. I agree that this is concerning when we report final results.
2. We could change the batch_size to 1 similar to what we do on detection. If that slows down the epochs too much we can try doing it only when `--test-only` is passed. This way we will be able to report correct stats for our models.
3. Setting `shuffle=False` for testing is IMO something worth doing anyway to get more determinism.
4. We could also try setting `drop_last=True` on the DataLoader during training. Not sure it improves things by much though.

I think lightning is likely to face similar issues but for now I think it's worth patching batch_size=1 when `--test-only` and settings `shuffle=False` on the validation sampler. This can be a good bootcamp task BTW. Let me know what you think.

NicolasHug:
OK, things are a bit clearer to me now.

As @fmassa's suggested (thanks!) the variance might come from the non-deterministic algorithms that are in use. I set the following:

```py
    torch.backends.cudnn.benchmark = False
    torch.use_deterministic_algorithms(True)
[...]

PR #3323, opened 2021-01-29
Add Quantizable MobilenetV3 architecture for Classification

datumbox:
The pre-trained model was trained:

```
python -m torch.distributed.launch --nproc_per_node=8 --use_env train_quantization.py\
--model='mobilenet_v3_large' --wd 0.00001 --lr 0.001
```
Submitted batch job 35496554 

Validated with:
```
python train_quantization.py --device='cpu' --model='mobilenet_v3_large' --test-only
```

Accuracy metrics (Epoch 89):
`Acc@1 73.004 Acc@5 90.858`

PR #5797, opened 2022-04-10
[Quant] Add FX support in quantization examples

andrewor14:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #5797

Summary: Previously, the quantization examples use only eager
mode quantization. This commit adds support for FX mode
quantization as well.

Test Plan:

```
# ==================== PTQ ====================
# MODEL is one of googlenet, inception_v3, resnet18, resnet50, resnext101_32x8d,
# shufflenet_v2_x0_5, shufflenet_v2_x1_0

# eager
python train_quantization.py --device="cpu" --post-training-quantize --backend="fbgemm"\
  --model="$MODEL" --weights="IMAGENET1K_V1" --quantization-workflow-type="eager_mode_quantization"

# fx
python train_quantization.py --device="cpu" --post-training-quantize --backend="fbgemm"\
  --model="$MODEL" --weights="IMAGENET1K_V1" --quantization-workflow-type="eager_mode_quantization"

# ==================== QAT ====================
# mobilenet_v2 eager
python train_quantization.py --device="cuda" --backend="qnnpack" --model="mobilenet_v2"\
  --epochs=10 --workers=64 --weights="IMAGENET1K_V1" --lr=0.0001 --weight-decay=0.0001\
  --quantization-workflow-type="eager_mode_quantization"

# mobilenet_v2 fx
python train_quantization.py --device="cuda" --backend="qnnpack" --model="mobilenet_v2"\
  --epochs=10 --workers=64 --weights="IMAGENET1K_V1" --lr=0.0001 --weight-decay=0.0001\
  --quantization-workflow-type="fx_graph_mode_quantization"

# mobilenet_v3_large eager
python train_quantization.py --device="cuda" --backend="qnnpack" --model="mobilenet_v3_large"\
  --epochs=10 --workers=64 --weights="IMAGENET1K_V1" --lr=0.001 --weight-decay=0.00001\
  --quantization-workflow-type="eager_mode_quantization"

# mobilenet_v3_large fx
python train_quantization.py --device="cuda" --backend="qnnpack" --model="mobilenet_v3_large"\
  --epochs=10 --workers=64 --weights="IMAGENET1K_V1" --lr=0.001 --weight-decay=0.00001\
  --quantization-workflow-type="fx_graph_mode_quantization"
```

Screen Shot 2022-04-26 at 12 35 30 PM

Reviewers: jerryzh168, vkuzo

Subscribers: jerryzh168, vkuzo


torchvision/models/quantization/mobilenetv3.py
```diff
@@ -135,20 +136,30 @@ def _mobilenet_v3_model(
...
-        model.qconfig = torch.ao.quantization.get_default_qat_qconfig(backend)
-        torch.ao.quantization.prepare_qat(model, inplace=True)
+        qconfig = torch.ao.quantization.get_default_qat_qconfig(backend)
+        if fx_mode:
+            qconfig_dict = {"": qconfig}
+            model = torch.ao.quantization.quantize_fx.fuse_fx(model)
+            model = torch.ao.quantization.quantize_fx.prepare_qat_fx(model, qconfig_dict)
+        else:
+            model.qconfig = qconfig
+            model.fuse_model(is_qat=True)
```
datumbox, line 152: You've switched the order of `qconfig` assignment and `fuse_model()` call. Is this expected to have an effect?

references/classification/train_quantization.py
```diff
@@ -46,13 +47,19 @@ def main(args):
 
     print("Creating model", args.model)
     # when training quantized models, we always start from a pre-trained fp32 reference model
-    model = torchvision.models.quantization.__dict__[args.model](weights=args.weights, quantize=args.test_only)
+    model = torchvision.models.quantization.__dict__[args.model](weights=args.weights, quantize=args.test_only, fx_mode=args.fx)
```
datumbox, line 50: Though this is the most BC approach for adding support of fx quantization, this API exposes technical details about quantization to the user. Should we be keeping this an internal implementation detail?

Another question I have is related to the use of the various custom nn.Modules (QuantizableMobileNetV2, QuantizableInvertedResidual etc) required by the previous legacy quantization API. My understanding is FX quantization no longer needs this but yet the way we implement fx quantization here still uses those custom modules. Is this necessary? Should we instead create a new set of model builders at `torchvision.prototype.models.quantization` that use the new FX api and don't make use of the custom modules?

datumbox (commented):
Thanks for the work @andrewor14. Some initial questions concerning the new API below.

references/classification/train_quantization.py
vkuzo: `FX` is a generic framework which is not specific to quantization.  I think a better name for this flag would be something like `quantization_workflow_type`, which defaults to Eager mode and has FX graph mode quantization as an option.

torchvision/models/quantization/mobilenetv3.py
```diff
@@ -135,20 +136,30 @@ def _mobilenet_v3_model(
         if "backend" in weights.meta:
             _ovewrite_named_param(kwargs, "backend", weights.meta["backend"])
     backend = kwargs.pop("backend", "qnnpack")
+    fx_mode = kwargs.pop("fx_mode", False)
```
vkuzo, line 139: same here, can this flag be named in a way that's specific to quantization?  The official name of the workflow is "FX graph mode quantization".

torchvision/models/quantization/utils.py
```diff
@@ -18,28 +19,36 @@ def _replace_relu(module: nn.Module) -> None:
...
             activation=torch.ao.quantization.default_observer, weight=torch.ao.quantization.default_weight_observer
         )
 
-    # TODO https://github.com/pytorch/vision/pull/4232#pullrequestreview-730461659
-    model.fuse_model()  # type: ignore[operator]
-    torch.ao.quantization.prepare(model, inplace=True)
-    model(_dummy_input_data)
-    torch.ao.quantization.convert(model, inplace=True)
+    if fx_mode:
+        qconfig_dict = {"": qconfig}
```
vkuzo, line 40: would it make sense to use `get_default_qconfig_dict` and `get_default_qat_qconfig_dict` here?

torchvision/models/quantization/utils.py
```diff
@@ -18,28 +19,36 @@ def _replace_relu(module: nn.Module) -> None:
...
+        model = torch.ao.quantization.quantize_fx.fuse_fx(model)
```
vkuzo, line 41: I thought the fusion happens inside of prepare, do we need to do it again here?

references/classification/train_quantization.py
```diff
@@ -46,13 +47,19 @@ def main(args):
...
     model.to(device)
 
     if not (args.test_only or args.post_training_quantize):
-        model.fuse_model(is_qat=True)
-        model.qconfig = torch.ao.quantization.get_default_qat_qconfig(args.backend)
-        torch.ao.quantization.prepare_qat(model, inplace=True)
+        qconfig = torch.ao.quantization.get_default_qat_qconfig(args.backend)
+        if args.fx:
+            qconfig_dict = {"": qconfig}
+            model = torch.ao.quantization.quantize_fx.fuse_fx(model)
```
vkuzo, line 57: do we need to fuse explicitly here?

torchvision/models/quantization/utils.py
jerryzh168: yeah we don't need to call fuse_fx explicitly I think

torchvision/models/quantization/mobilenetv3.py
andrewor14: I checked with @jerryzh168. It should not change any behavior

andrewor14: Right. I just switched it to make the code look more consistent with the FX branch. I don't think it should have any change in behavior.

torchvision/models/quantization/utils.py
andrewor14: I think so. I just wanted to preserve the eager mode code path as much as possible so I didn't change the qconfig definitions in L30 and L35, but I think `get_default_qconfig_dict` should be the same here

references/classification/train_quantization.py
jerryzh168: yeah I feel exposing fx as an option to user sounds OK, otherwise, user still need to call a different script to use fx graph mode quantization, this information is exposed anyways

for the custom module for eager mode quantization, just chat with Andrew offline about this, he will change the code to use original floating point model directly in fx mode

andrewor14: @datumbox Yes, it's actually not necessary. I think I will modify this to use the original model (e.g. `MobileNetV2`) instead of the one modified for eager mode quantization (e.g. `QuantizableMobileNetV2`) in FX mode. Also, I believe "custom modules" refers to a different, unrelated quantization API that we're not using here.

torchvision/models/quantization/mobilenetv3.py
andrewor14: Ok, I think I will rename both flags to the same thing `quantization_workflow_type` and just check if it's equal to `"fx"`

torchvision/models/quantization/utils.py
andrewor14: Ok, will remove

torchvision/models/quantization/mobilenetv3.py
vkuzo: nit: it's still not ideal to use "fx" as a string value, the workflow name is "FX graph mode quantization", it would be good to be consistent to minimize confusion.  There are also other FX passes people might be using.

references/classification/train_quantization.py
vkuzo: > My understanding is FX quantization no longer needs this but yet the way we implement fx quantization here still uses those custom modules. Is this necessary? Should we instead create a new set of model builders at torchvision.prototype.models.quantization that use the new FX api and don't make use of the custom modules?

maybe we can throw a warning if a user tries to use the eager mode quantizeable modules with the FX graph mode API?

torchvision/models/quantization/utils.py
andrewor14: (Actually I'm reverting all these changes since FX mode quantization should not touch the models under `torchvision.models.quantization`)

jerryzh168 (approved):
looks great! can you include the accuracy numbers in the summary and update the docs that shows accuracy as well

references/classification/train_quantization.py
andrewor14: @vkuzo Since the latest code doesn't pass the FX flag to the quantized models anymore, I think it'll be a bit complicated to print the warning. Maybe we can do this in a future PR.

torchvision/models/quantization/mobilenetv3.py
andrewor14: (This change was reverted since we don't need to explicitly fuse in FX mode)

andrewor14:
I updated the description with numbers from `mobilenet_v2` and `mobilenet_v3_large`. The results are good: the eager mode accuracies match FX accuracies across all models. This is ready from my side. Any other comments @vkuzo @datumbox?

andrewor14:
There was a problem with the way I set up the FX experiments. After fixing this problem I uncovered a new blocking issue that I summarized here https://github.com/pytorch/pytorch/issues/75825. I will rerun all the experiments after we fix that issue.

references/classification/train_quantization.py
```diff
@@ -233,6 +268,12 @@ def get_args_parser(add_help=True):
         help="Post training quantize the model",
         action="store_true",
     )
+    parser.add_argument(
+        "--quantization-workflow-type",
```
datumbox, line 272: Why not have a `--use-fx-quantization` boolean flag instead? Do we expect more types of quantization on the future?

references/classification/train_quantization.py
```diff
@@ -22,6 +29,17 @@ def main(args):
...
         raise RuntimeError("Post training quantization example should not be performed on distributed mode")
 
+    # Validate quantization workflow type
+    quantization_workflow_type = args.quantization_workflow_type.upper()
+    if quantization_workflow_type not in QuantizationWorkflowType.__members__:
+        raise RuntimeError(
+            "Unknown workflow type '%s', please choose from: %s"
+            % (args.quantization_workflow_type, str(tuple([t.lower() for t in QuantizationWorkflowType.__members__])))
+        )
+    use_fx_graph_mode_quantization = (
```
datumbox, line 39: This will be simplified drastically if we change to a boolean flag as mentioned above. The Enum class can be removed and so does this additional code with:
```
use_fx_graph_mode_quantization = args.use_fx_quantization
```

references/classification/train_quantization.py
```diff
@@ -84,13 +109,20 @@ def main(args):
...
+            model.qconfig = torch.ao.quantization.get_default_qconfig(args.backend)
+            torch.ao.quantization.prepare(model, inplace=True)
         # Calibrate first
         print("Calibrating")
         evaluate(model, criterion, data_loader_calibration, device=device, print_freq=1)
-        torch.ao.quantization.convert(model, inplace=True)
+        if use_fx_graph_mode_quantization:
+            model = torch.ao.quantization.quantize_fx.convert_fx(model)
+        else:
+            torch.ao.quantization.convert(model, inplace=True)
```
datumbox, line 125: Since the updated version of this PR, no longer touches the `torchvision.models.quantization` methods, it's unclear how one would use models trained with the FX-based approach. Aka:
- How do we construct quantized versions of the models?
- How do we load the weights?

From what I understand in the Post-Training-Quantization case, the user would just do the following:
```python
# 1. Construct pre-trained unquantized model
model = torchvision.models.resnet50(weights=ResNet50_Weights.IMAGENET1K_V1)
model.eval()

# 2. Quantize it 
qconfig_dict = torch.ao.quantization.get_default_qconfig_dict("qnnpack")
model = torch.ao.quantization.quantize_fx.prepare_fx(model, qconfig_dict)
model = torch.ao.quantization.quantize_fx.convert_fx(model)

# 3. Use it as normal
output = model(input)
```

Because step 2 is identical to all models, it can be a utility and thus there is no need for a custom quantized model builder (aka `torchvision.models.quantized.resnet50()`). Correct?

For QAT, I believe we need a quantized model builder because one needs to do:
```python
# 1. Construct an non-initialized model 
model = torchvision.models.resnet50(weights=None)
model.eval()

# 2. Convert it to FX
model = torch.ao.quantization.quantize_fx.convert_fx(model)

# 3. Load the QAT weights
weights = ResNet50_Weights.IMAGENET1K_QNNPACK_V1
model.load_state_dict(weights.get_state_dict(progress=True))

# 4. Use it as normal
output = model(input)
```

For QAT, the loading of the weights (step 3) must be performed after FX conversion (step 2) which means that now a custom quantized model builder is necessary to make it convenient to the users. Right?

references/classification/train_quantization.py
```diff
@@ -46,13 +64,20 @@ def main(args):
 
     print("Creating model", args.model)
     # when training quantized models, we always start from a pre-trained fp32 reference model
-    model = torchvision.models.quantization.__dict__[args.model](weights=args.weights, quantize=args.test_only)
+    if use_fx_graph_mode_quantization:
+        model = torchvision.models.__dict__[args.model](weights=args.weights)
```
datumbox, line 68: I suspect this change doesn't work for all cases. What if you want to perform inference (with `--test-only`) on a model trained with QAT? The way we do it here, we will end up evaluating the unquantized model.

datumbox (commented):
Thanks for the update. I've added some additional comments and questions. Let me know your thoughts.

docs/source/models.rst
```diff
@@ -486,13 +486,13 @@ Model                             Acc@1          Acc@5
 ================================  =============  =============
 MobileNet V2                      71.658         90.150
 MobileNet V3 Large                73.004         90.858
-ShuffleNet V2 x1.0                68.360         87.582
-ShuffleNet V2 x0.5                57.972         79.780
-ResNet 18                         69.494         88.882
-ResNet 50                         75.920         92.814
-ResNext 101 32x8d                 78.986         94.480
+ShuffleNet V2 x1.0                67.886         87.332
```
datumbox, line 489: These are pretty big accuracy gaps. Diffs smaller than 0.2 could be attributed to randomness but anything bigger needs to be investigated. Any thoughts on why we get differences of 0.5% here? Note that if this is indeed randomness, doing 2-3 more runs should give you similar good results.

I also noticed that you didn't update the values for MobileNet V2/V3 and Inception. What are the updated numbers for these models?

datumbox (changes requested):
Just marking as "Request changes" to avoid accidental merges while we clarify some remaining questions.

references/classification/train_quantization.py
vkuzo: note: the product name is "FX graph mode quantization".  There may be other quantization workflows that use FX in the future.

references/classification/train_quantization.py
andrewor14: Hi @datumbox, actually both post-training quantization and QAT code paths are the same for FX and eager modes. In both cases, the workflows are as follows (pseudocode):
```
# PTQ
model = prepare(model) # or prepare_fx
calibrate(model, data)
model = convert(model) # or convert_fx

# QAT
model = prepare_qat(model) # or prepare_qat_fx
train_one_epoch(model, data)
evaluate(model, data)
model = convert(model) # or convert_fx
```
The main difference between FX and eager modes is which model is used. In eager mode, the user needs to manually at `QuantStub`s and `DeQuantStub`s, so they must use the `torchvision.models.quantization` version. In FX mode, however, the user can simply quantize the model as is (assuming it is symbolically traceable). This is why I used the unmodified `torchvision.models` version for FX mode in this PR.

> Because step 2 is identical to all models, it can be a utility and thus there is no need for a custom quantized model builder (aka torchvision.models.quantized.resnet50()). Correct?

Correct. This is why we use the original `torchvision.models` for FX quantization. However, we still need the custom quantized model builders (`torchvision.models.quantization`) for eager mode quantization.

> For QAT, the loading of the weights (step 3) must be performed after FX conversion (step 2) which means that now a custom quantized model builder is necessary to make it convenient to the users. Right?

Actually if there is no training involved and the goal of the user is to simply load quantized weights then they can just use PTQ. For e.g., they would simply initialize the model with these [quantized weights](https://github.com/pytorch/vision/blob/main/torchvision/models/quantization/resnet.py#L178) instead.

For QAT, my understanding is the user would only load the weights in the beginning _before_ training (and `convert`) as part of loading from pretrained weights. For FX QAT, however, there is still no need for a custom quantized model builder (i.e. `torchvision.models.quantization`) because the user doesn't have to make any manual modifications to the models in order to quantize them.

Please let me know if there is anything unclear in my explanation or if I misunderstood any part of your questions.

references/classification/train_quantization.py
andrewor14: Yes, actually in the original changes we had a simple `--fx` flag. However, in the future we may add support for `define_by_run_quantization` in addition to the two existing workflow types.

references/classification/train_quantization.py
andrewor14: I think the Enum is still needed if we want to support other quantization modes in the future.

docs/source/models.rst
andrewor14: Ok, I will rerun them again to see if we get a smaller gap. Also, these numbers are now outdated and I will follow up with the latest set of accuracies. For MobileNet, they are taking a long time to run so I don't have the numbers yet. For inception, I was not able to reproduce the original accuracies in my initial attempt but I am still working on it.

references/classification/train_quantization.py
andrewor14: I think you're right. What we want here is to have something like the following:
```
model = torchvision.models.resnet50(ResNet50_QuantizedWeights.IMAGENET1K_FBGEMM_V1)
```
However, the way the unquantized models are set up today there is no way to pass in `ResNet50_QuantizedWeights` to `torchvision.models.resnet50`. I think one way forward is to make `ResNet50_QuantizedWeights` a subclass of `ResNet50_Weights`. I will look into this further.

references/classification/train_quantization.py
datumbox: Thanks for clarifying. I now understand that this was done to support other FX-based quantization approaches on the future. :)

Note that we try to keep our examples/reference scripts as simple as possible and that we don't offer strong BC guarantees on this code. So it would be OK in this case for you to just introduce a boolean argument if this greatly simplifies the code. This is just a nit, so definitely no blocking. I'll leave this up to you.

vkuzo: by the way, my comments were strictly about the name.  There were some long internal discussions in the past on what to name this product, and in the end "FX graph mode quantization" was the name that was chosen, I was just hoping we can stay consistent to it.  

We also considered "FX quantization" but did not choose that because we wanted to make sure that "graph mode" was in the name.  There is also an outdated "TorchScript graph mode quantization", and an upcoming "DBR quantization" which does not use FX directly but may interact with FX.

references/classification/train_quantization.py
datumbox: This is why I'm trying to figure out what are the steps for loading pre-trained quantized weights to a model. Can this be done directly to the original model without any conversions? Or some FX convert call is needed prior loading?

references/classification/train_quantization.py
datumbox: Thanks a lot for the reply.

The examples you provided are great. I think the training part is clear to me after reading the PR. What I would like to understand better is how PTQ and QAT weight are loaded in a model. I want to confirm that indeed there is no need to add new model builders at `torchvision.models.quantization` for the FX case and that I don't need to modify any of the existing model builders at `torchvision.models`.

So assume I'm using the original builders for `torchvision.models` without any modification. What steps if any do I have to take to load PTQ/QAT weights? Example:

```python
# 1. Construct an non-initialized model 
model = torchvision.models.resnet50(weights=None)
model.eval()

# 2a. Load PTQ weights
# TODO

# 2b. Load QAT weights
# TODO
```

Could you fill the TODOs for me?

docs/source/models.rst
datumbox: Sounds good. Let me know if I can help with any of it.

andrewor14:
Hi @datumbox, as of the latest set of changes, both FX and eager code paths go through the "custom model builders" (e.g. `torchvision.models.quantization.googlenet`). Note that FX still uses the original unmodified model (e.g. `GoogLeNet`), while eager mode still uses the version with manually inserted quant and dequant stubs (e.g. `QuantizableGoogLeNet`) as before.

The main reason why I decided to route both FX and eager mode through the "custom model builder" was to handle the `test_only` use case. For some context, `train_quantization.py` has three different modes: PTQ, QAT, and `test_only`. The way PTQ and QAT differ from `test_only` mode is as follows:

```
# PTQ / QAT eager
weights = get_pretrained_fp32_weights()
model = torchvision.models.quantization.QuantizableResNet50(...)
model.load_state_dict(weights.get_state_dict())
model = prepare(model) # or prepare_qat
calibrate(model, data) # or train
model = convert(model)
evaluate(model)

# test_only eager
weights = get_quantized_weights()
model = torchvision.models.quantization.QuantizableResNet50(...)
model = prepare(model)
calibrate(model, dummy_data)
model = convert(model)
model.load_state_dict(weights.get_state_dict())
evaluate(model)
```

For `test_only`, the weights are quantized and so they must be loaded after the call to convert. However, this ordering constraint was difficult to implement cleanly in FX if we used the original model builders (e.g. `torchvision.models.googlenet`), as it would require adding quantization logic to these FP32 builders. This is why FX now also goes through the quantized model builders (e.g. `torchvision.models.quantization.googlenet`) as in eager mode, but additionally performs a check to ensure the original unmodified model is used. This leads to the following behavior:

```
# PTQ / QAT FX
weights = get_pretrained_fp32_weights()
model = torchvision.models.ResNet50(...)
model.load_state_dict(weights.get_state_dict())
model = prepare_fx(model) # or prepare_qat_fx
calibrate(model, data) # or train
model = convert_fx(model)
evaluate(model)

# test_only FX
weights = get_quantized_weights()
model = torchvision.models.ResNet50(...)
model = prepare_fx(model)
calibrate(model, dummy_data)
model = convert_fx(model)
model.load_state_dict(weights.get_state_dict())
evaluate(model)
```

This approach has the additional advantage of allowing the user to use these quantized model builders with FX outside the context of `train_quantization.py` as follows:

```
googlenet = models.quantization.googlenet(
    weights=GoogLeNet_QuantizedWeights.IMAGENET1K_FBGEMM_V1,
    quantize=True,
    quantization_workflow_type=QuantizationWorkflowType.FX_GRAPH_MODE)
```

references/classification/train_quantization.py
andrewor14: Hi @datumbox, I'm continuing the discussion in the main thread. My comment there should clarify when the weights are loaded in different modes (PTQ/QAT/test_only).

docs/source/models.rst
andrewor14: Just reran and uploaded the new numbers as part of the PR description. There was a lot of variance across runs so I just selected the runs with the highest accuracies. Note that inception and mobilenet are still TODOs at this point.

references/classification/train_quantization.py
andrewor14: I was unable to make `ResNet50_QuantizedWeights` inherit from `ResNet50_Weights` in the end because they are both enums. Instead I made FX mode go through the quantized model builders (`torchvision.models.quantization.resnet`) which are more extensible. Let's move this discussion to the main thread, where I added a comment to clarify this new approach.

andrewor14: Quick answer to your question here is, yes, in `test_only` mode, we need to call `convert` before loading the weights, since they are quantized weights. This is true for both eager and FX. However, PTQ and QAT don't have this constraint.

andrewor14:
Hi @datumbox, continuing our discussion here:

> What steps if any do I have to take to load PTQ/QAT weights?

> I'm trying to figure out what are the steps for loading pre-trained quantized weights to a model. Can this be done directly to the original model without any conversions? Or some FX convert call is needed prior loading?

For the `test_only` mode, we use quantized weights so convert must happen before loading the weights. For PTQ and QAT, however, we use pretrained FP32 weights so convert must happen after loading the weights. This is the same for eager mode and FX graph mode. In some more detail, here is the behavior today (pseudocode):

```
# PTQ / QAT eager
weights = get_pretrained_fp32_weights()
model = torchvision.models.quantization.QuantizableResNet50(...)
model.load_state_dict(weights.get_state_dict())
model = prepare(model) # or prepare_qat in QAT mode
calibrate(model, data) # or train in QAT mode
model = convert(model)
evaluate(model)

# test_only eager
model = torchvision.models.quantization.QuantizableResNet50(...)
model = prepare(model)
calibrate(model, dummy_data)
model = convert(model)
weights = get_quantized_weights()
model.load_state_dict(weights.get_state_dict())
evaluate(model)
```

What we want is the equivalent for FX. The only differences here are (1) We use `prepare_fx` (and the like) instead of `prepare`, and (2) We use the unmodified FP32 model (e.g. `torchvision.models.ResNet50`) instead of the model with manually inserted quant and dequant stubs (e.g. `torchvision.models.quantization.QuantizableResNet50`).

docs/source/models.rst
andrewor14: Just reran and uploaded the new numbers as part of the PR description (see table). There was a lot of variance across runs so I just selected the runs with the highest accuracies. Everything should be within 0.2 or 0.3 of the reference accuracy now. As for mobilenet, I'm still waiting on the results.

references/classification/train_quantization.py
andrewor14: No problem, let's continuing the discussion in the main thread. My comment there should clarify when the weights are loaded in different modes (PTQ/QAT/test_only).

torchvision/models/quantization/mobilenetv3.py
```diff
@@ -135,20 +136,32 @@ def _mobilenet_v3_model(
...
+        model = MobileNetV3(inverted_residual_setting, last_channel, **kwargs)
+    else:
+        model = QuantizableMobileNetV3(inverted_residual_setting, last_channel, block=QuantizableInvertedResidual, **kwargs)
+        _replace_relu(model)
 
     if quantize:
-        model.fuse_model(is_qat=True)
-        model.qconfig = torch.ao.quantization.get_default_qat_qconfig(backend)
-        torch.ao.quantization.prepare_qat(model, inplace=True)
+        # TODO: This shouldn't be QAT?
```
datumbox, line 148: As you see MobiletNetV3 doesn't use `quantize_model` but instead does this very weird thing. This was done to avoid an old bug (see https://github.com/pytorch/vision/pull/3323/commits/274c6a1393384054876d701ffa1b54eb6750f1d8#r569264679) in which when we loaded back the results of the QAT and tried to use them with `--test-only` the accuracy was extremely low. But if we instead loaded the QAT unconverted weights, and did the conversion on the fly it worked.

docs/source/models.rst
```diff
@@ -486,13 +486,13 @@ Model                             Acc@1          Acc@5
...
-ResNext 101 32x8d                 78.986         94.480
-Inception V3                      77.176         93.354
-GoogleNet                         69.826         89.404
+ShuffleNet V2 x1.0                68.240         87.524
+ShuffleNet V2 x0.5                57.810         79.724
+ResNet 18                         69.500         88.958
+ResNet 50                         75.802         92.764
+ResNext 101 32x8d                 79.020         94.468
+Inception V3                      77.206         93.576
+GoogleNet                         69.702         89.388
```
datumbox, line 495: Worth noting that we need to estimate these values with batch-size=1 on 1 GPU to avoid variances introduced due to batch padding (see #4559).

datumbox:
@andrewor14 Thanks a lot for your work on this. Let's discuss the design doc you shared with me offline to decide on the approach and minimize throw away work. :)

torchvision/models/quantization/mobilenetv3.py
andrewor14: I see, that is very weird indeed. Maybe we should add a comment here and link to that issue to provide some context

andrewor14:
As discussed offline, I'm closing this PR for now since the FX graph mode quantization API is subject to change. I will reopen this PR again and rerun all the experiments once the new API is ready. Thank you everyone for your comments so far!

torchvision/models/quantization/mobilenetv3.py
datumbox: Good call. Added a comment at https://github.com/pytorch/vision/pull/5891 and documented the issue at https://github.com/pytorch/vision/issues/5890

PR #5799, opened 2022-04-10
Focal Loss Documentation enhanced

abhi-glitchhg:


torchvision/ops/focal_loss.py
```diff
@@ -16,19 +16,20 @@ def sigmoid_focal_loss(
     Loss used in RetinaNet for dense detection: https://arxiv.org/abs/1708.02002.
 
     Args:
-        inputs: A float tensor of arbitrary shape.
+        inputs: (Tensor) A float tensor of arbitrary shape.
```
datumbox, line 19: The type is put before the colon character:
```suggestion
        inputs (Tensor): A float tensor of arbitrary shape.
```

The same everywhere else.

datumbox (commented):
@abhi-glitchhg Thanks for the improvements. Just one comment on the styles.

torchvision/ops/focal_loss.py
abhi-glitchhg: Oh my bad . Sorry

torchvision/ops/focal_loss.py
```diff
@@ -16,19 +16,20 @@ def sigmoid_focal_loss(
     Loss used in RetinaNet for dense detection: https://arxiv.org/abs/1708.02002.
 
     Args:
-        inputs: A float tensor of arbitrary shape.
+        inputs (Tensor) : A float tensor of arbitrary shape.
```
datumbox, line 19: Sorry for nitpicking but we should keep the styles consistent across the code:
```suggestion
        inputs (Tensor): A float tensor of arbitrary shape.
```

Here is an example of the styles we use for reference:
https://github.com/pytorch/vision/blob/289fce29b3e2392114aadbe7a419df0f2e3ac1be/torchvision/utils.py#L32-L55

datumbox: You are right. I don't think we have been very consistent with the use of `optional` in docstrings. A common use in the codebase is to indicate that `None` is a valid value. Less commonly it's also used when the parameter can omitted because it has a default value (keyword argument). @NicolasHug do you have any thoughts on this?

torchvision/ops/focal_loss.py
```diff
@@ -10,25 +10,25 @@ def sigmoid_focal_loss(
     alpha: float = 0.25,
     gamma: float = 2,
     reduction: str = "none",
-):
+) -> torch.Tensor:
     """
     Original implementation from https://github.com/facebookresearch/fvcore/blob/master/fvcore/nn/focal_loss.py .
```
datumbox, line 15: Ad discussed at https://github.com/pytorch/vision/pull/5776#discussion_r844892807, long string URLs cause rendering issues on the documentation. Let's move this to a comment on source code.
```suggestion
```

torchvision/ops/focal_loss.py
```diff
@@ -10,25 +10,25 @@ def sigmoid_focal_loss(
...
+                positive vs negative examples or -1 for ignore. Default: ``0.25``.
+        gamma (float): Exponent of the modulating factor (1 - p_t) to
+                balance easy vs hard examples. Default: ``2``.
+        reduction (string): ``'none'`` | ``'mean'`` | ``'sum'``
+                ``'none'``: No reduction will be applied to the output.
+                ``'mean'``: The output will be averaged.
+                ``'sum'``: The output will be summed. Default: ``'none'``.
     Returns:
         Loss tensor with the reduction option applied.
     """
```
datumbox, line 34:
```suggestion
    """
    # Original implementation from https://github.com/facebookresearch/fvcore/blob/master/fvcore/nn/focal_loss.py .
```

PR #5801, opened 2022-04-11
upgrade prototype test CI from 3.7 to 3.10

pmeier:
@ejguan @NivekT starting for Python >= 3.8 we are seeing [a lot of I/O warnings](https://app.circleci.com/pipelines/github/pytorch/vision/16602/workflows/d6a57e68-d35b-464d-8742-b4a7550eb8a0/jobs/1345131) that are turned into errors by our test suite. I was under the impression that we don't need to handle file closing. Has this changed?

### DIFF
--- a/test/test_prototype_builtin_datasets.py
+++ b/test/test_prototype_builtin_datasets.py
@@ -41,7 +41,7 @@ def test_coverage():
         )
 
 
-@pytest.mark.filterwarnings("ignore::ResourceWarning")
+# @pytest.mark.filterwarnings("ignore::ResourceWarning")
 @pytest.mark.filterwarnings("error")
 class TestCommon:
     @pytest.mark.parametrize("name", datasets.list_datasets())
--- a/test/test_prototype_builtin_datasets.py
+++ b/test/test_prototype_builtin_datasets.py
@@ -83,12 +83,12 @@ class TestCommon:
         except Exception as error:
             raise AssertionError("Drawing a sample raised the error above.") from error
 
-        for dp in extract_datapipes(dataset):
-            if not isinstance(dp, _DemultiplexerIterDataPipe):
-                continue
-
-            for dq in dp.child_buffers:
-                dq.clear()
+        # for dp in extract_datapipes(dataset):
+        #     if not isinstance(dp, _DemultiplexerIterDataPipe):
+        #         continue
+        #
+        #     for dq in dp.child_buffers:
+        #         dq.clear()
 
         if not isinstance(sample, dict):
             raise AssertionError(f"Samples should be dictionaries, but got {type(sample)} instead.")
--- a/test/test_prototype_builtin_datasets.py
+++ b/test/test_prototype_builtin_datasets.py
@@ -9,7 +9,6 @@ import torch
 from builtin_dataset_mocks import parametrize_dataset_mocks, DATASET_MOCKS
 from torch.testing._comparison import assert_equal, TensorLikePair, ObjectPair
 from torch.utils.data import DataLoader
-from torch.utils.data.datapipes.iter.combining import _DemultiplexerIterDataPipe
 from torch.utils.data.graph import traverse
 from torch.utils.data.graph_settings import get_all_graph_pipes
 from torchdata.datapipes.iter import Shuffler, ShardingFilter
@@ -82,13 +81,6 @@ class TestCommon:
             raise AssertionError("Unable to draw any sample.") from None
         except Exception as error:
             raise AssertionError("Drawing a sample raised the error above.") from error
-
-        # for dp in extract_datapipes(dataset):
-        #     if not isinstance(dp, _DemultiplexerIterDataPipe):
-        #         continue
-        #
-        #     for dq in dp.child_buffers:
-        #         dq.clear()
 
         if not isinstance(sample, dict):
             raise AssertionError(f"Samples should be dictionaries, but got {type(sample)} instead.")
--- a/test/test_prototype_builtin_datasets.py
+++ b/test/test_prototype_builtin_datasets.py
@@ -1,3 +1,4 @@
+import contextlib
 import functools
 import gc
 import io
@@ -9,6 +10,7 @@ import torch
 from builtin_dataset_mocks import parametrize_dataset_mocks, DATASET_MOCKS
 from torch.testing._comparison import assert_equal, TensorLikePair, ObjectPair
 from torch.utils.data import DataLoader
+from torch.utils.data.datapipes.iter.combining import _DemultiplexerIterDataPipe
 from torch.utils.data.graph import traverse
 from torch.utils.data.graph_settings import get_all_graph_pipes
 from torchdata.datapipes.iter import Shuffler, ShardingFilter
@@ -81,6 +83,11 @@ class TestCommon:
             raise AssertionError("Unable to draw any sample.") from None
         except Exception as error:
             raise AssertionError("Drawing a sample raised the error above.") from error
+
+        with contextlib.suppress(StopIteration):
+            demux = next(dp for dp in extract_datapipes(dataset) if isinstance(dp, _DemultiplexerIterDataPipe))
+            if any(bool(buffer) for buffer in demux.child_buffers):
+                print("Found some data in the buffer")
 
         if not isinstance(sample, dict):
             raise AssertionError(f"Samples should be dictionaries, but got {type(sample)} instead.")
--- a/test/test_prototype_builtin_datasets.py
+++ b/test/test_prototype_builtin_datasets.py
@@ -1,7 +1,7 @@
-import contextlib
 import functools
 import gc
 import io
+import itertools
 import pickle
 from pathlib import Path
 
@@ -84,10 +84,12 @@ class TestCommon:
         except Exception as error:
             raise AssertionError("Drawing a sample raised the error above.") from error
 
-        with contextlib.suppress(StopIteration):
-            demux = next(dp for dp in extract_datapipes(dataset) if isinstance(dp, _DemultiplexerIterDataPipe))
-            if any(bool(buffer) for buffer in demux.child_buffers):
-                print("Found some data in the buffer")
+        for dp in extract_datapipes(dataset):
+            if not isinstance(dp, _DemultiplexerIterDataPipe):
+                continue
+
+            for _, buffer in itertools.chain.from_iterable(dp.child_buffers):
+                buffer.close()
 
         if not isinstance(sample, dict):
             raise AssertionError(f"Samples should be dictionaries, but got {type(sample)} instead.")
--- a/test/test_prototype_builtin_datasets.py
+++ b/test/test_prototype_builtin_datasets.py
@@ -48,7 +48,7 @@ def test_coverage():
         )
 
 
-@pytest.mark.filterwarnings("error")
+# @pytest.mark.filterwarnings("error")
 class TestCommon:
     @pytest.mark.parametrize("name", datasets.list_datasets())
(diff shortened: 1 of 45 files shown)


datumbox:
@pmeier Sorry for jumping on a draft PR. To clarify, is this PR an investigation or you eventually propose to merge? Typically we want to pin all tests on the earliest supported version, so if indeed the problems disappear we probably don't want to merge this but rather investigate and fix the underlying issue. Probably that's what you have in mind as well but just confirming. Thanks!

pmeier:
For now, this is just an investigation. @NicolasHug saw some failures locally that neither I nor CI could reproduce. Turns out, the errors (or better warnings turned into errors) only pop up for Python >= 3.8. This PR confirms that this is indeed the case. 

It should not be merged in the current state, but maybe will be the basis for further investigations or fixes. I'll either close it or mark it as "ready to be reviewed".

ejguan:
Checking it

ejguan:
I will spend more time on this issue.

ejguan:
I may find the culprit.
For example of cub200, when we construct data pipeline using `_datapipe(resources)`, we load a few data via 
https://github.com/pytorch/vision/blob/467b841ac2650c962952bbee5a542edb835c98bc/torchvision/prototype/datasets/_builtin/cub200.py#L188-L190

But, before creating the dict `image_files_map`, we have a `Demux` to hold file handlers. Before the iteration of data-loading pipeline, the `Demux` has been iterated and the buffer will be filled with unclosed file handlers. Then, for these file handlers, we can never close them as the reference count is always larger than 1.

I only narrowed down the root cause of such error. I haven't got enough time to verify whether or not we change `image_files_map` to `IterToMapConverter` to support lazy loading would solve the issue
https://github.com/pytorch/data/blob/fd942eec986db373f76f521528570bfef2f1d22f/torchdata/datapipes/map/util/utils.py#L20

cc: @NivekT

pmeier:
> But, I couldn't find the code to extract files from the tar file. Could you please provide the code pointer?

To complement Kevins comment in https://github.com/pytorch/vision/pull/5801#issuecomment-1096915679: the loading happens in 

https://github.com/pytorch/vision/blob/467b841ac2650c962952bbee5a542edb835c98bc/torchvision/prototype/datasets/utils/_resource.py#L62-L72

- Folder are loaded with `FileOpener(FileLister(...))`
- Files are loaded with `FileOpener(IterableWrapper(...))`
- Archives are additionally loaded with the respective loader datapipe to have the same behavior for folders and archives

> But, how about the `custom file`? What kind of files are you referring to?

We have a failing test for 

https://github.com/pytorch/vision/blob/467b841ac2650c962952bbee5a542edb835c98bc/torchvision/prototype/datasets/_builtin/semeion.py#L29

and that uses a `.data` file (csv in disguise) file and features no `Demultiplexer`. Upon closer inspection, it seems like the warning is raised at the wrong time:

```
failed on setup with "pytest.PytestUnraisableExceptionWarning: Exception ignored in: <_io.FileIO [closed]>
[...]
  File "/home/circleci/.pyenv/versions/3.10.4/lib/python3.10/unittest/mock.py", line 2153, in __init__
    self.name = name
ResourceWarning: unclosed file <_io.FileIO name='/tmp/pytest-of-circleci/pytest-0/test_num_samples_sbd_train_nov0/sbd/benchmark.tgz' mode='rb' closefd=True>
```

So actually SBD is the culprit here. I'll go through all of them again to find the actual datasets that are causing this.

pmeier:
Rechecking our and torchdata logs, it turns out these datasets are the culprits:

- [`clevr`](https://github.com/pytorch/vision/blob/467b841ac2650c962952bbee5a542edb835c98bc/torchvision/prototype/datasets/_builtin/clevr.py#L27)
- [`cub200`](https://github.com/pytorch/vision/blob/467b841ac2650c962952bbee5a542edb835c98bc/torchvision/prototype/datasets/_builtin/cub200.py#L43)
- [`dtd`]()
- [`food101`](https://github.com/pytorch/vision/blob/467b841ac2650c962952bbee5a542edb835c98bc/torchvision/prototype/datasets/_builtin/food101.py#L35)
- [`gtsrb`](https://github.com/pytorch/vision/blob/467b841ac2650c962952bbee5a542edb835c98bc/torchvision/prototype/datasets/_builtin/gtsrb.py#L31)
- [`sbd`](https://github.com/pytorch/vision/blob/467b841ac2650c962952bbee5a542edb835c98bc/torchvision/prototype/datasets/_builtin/sbd.py#L38)
- [`voc`](https://github.com/pytorch/vision/blob/467b841ac2650c962952bbee5a542edb835c98bc/torchvision/prototype/datasets/_builtin/voc.py#L39)

This eliminates point 1 from https://github.com/pytorch/vision/pull/5801#issuecomment-1096825604. My guess is that the warning is emitted during garbage collection and thus happens at a non-deterministic time.

pmeier:
As for the other similarities: all culprits feature a `Demultiplexer`, but `cub200` is the only dataset that starts the iteration during the construction of the datapipe. Another pointer is that we have other datasets that feature a `Demultiplexer`

- [`coco`](https://github.com/pytorch/vision/blob/467b841ac2650c962952bbee5a542edb835c98bc/torchvision/prototype/datasets/_builtin/coco.py#L47)
- [`imagenet`](https://github.com/pytorch/vision/blob/467b841ac2650c962952bbee5a542edb835c98bc/torchvision/prototype/datasets/_builtin/imagenet.py#L55)
- [`emnist`](https://github.com/pytorch/vision/blob/467b841ac2650c962952bbee5a542edb835c98bc/torchvision/prototype/datasets/_builtin/mnist.py#L230)

but don't seem to emit this warning.

ejguan:
> but `cub200` is the only dataset that starts the iteration during the construction of the datapipe

Yeah. It's the reason that only `cub200` fails for the case of `test_smoke`.

ejguan:
Adding this patch seems solve all the CI issue:
```
+    def __del__(self):
+        for q in self.child_buffers:
+            q.clear()
+
```
@NivekT Do you mind taking a lead to see why `StreamWrapper` does not close the inner file handler automatically? And how can we add a test to prevent such scenario?

NivekT:
I had a look and tried to reproduce the error locally with `torchvision`. I played around with `cub200` and `demux`.

This is the simplest test case that I have and it seems to work fine.
```python
dp = IterableWrapper([f'{i}.tar.gz' for i in [1, 2, 3]])
dp = FileOpener(dp, mode='rb').load_from_tar()
dp1, dp2 = dp.demux(2, lambda _: 0)
```
When the `demux` DataPipes is no longer needed, the file handles inside its buffer get closed correctly by `StreamWrapper`. 

`cub200` seems to be working locally for me as well. Let me know what code snippet you run to have a `ResourceWarning`.

I have been trying to figure out the following:
1. If `_DemultiplexerIterDataPipe` not cleaning up properly, what is the specific situation to cause the buffer to not clean up properly?
    - Are there references to those DataPipes inside the buffer elsewhere, such that they don't get deleted?
2. Is `StreamWrapper`'s `__del__` logic causing the issue here?

I am leaning towards the first thing is the issue here, therefore, I think @ejguan's suggestion to specifically clear out the buffer is reasonable and we should do that unless we can figure out what specifically cause the buffer not to clear.

ejguan:
You can reproduce the Error from `pytest test/test_prototype_builtin_datasets.py -v -k <pattern of test>` @NivekT

pmeier:
I'll try to compile a minimal reproduction outside of our test environment.

pmeier:
I was able to pinpoint it to cases where the buffer of the `Demultiplexer` was not empty. For the datasets in https://github.com/pytorch/vision/pull/5801#issuecomment-1097630775 the buffer is empty and thus we are not seeing a failure.

test/test_prototype_builtin_datasets.py
```diff
@@ -75,6 +83,13 @@ def test_sample(self, test_home, dataset_mock, config):
         except Exception as error:
             raise AssertionError("Drawing a sample raised the error above.") from error
 
+        for dp in extract_datapipes(dataset):
+            if not isinstance(dp, _DemultiplexerIterDataPipe):
+                continue
+
+            for dq in dp.child_buffers:
+                dq.clear()
```
pmeier, line 91: This is similar to what @ejguan proposed in https://github.com/pytorch/vision/pull/5801#issuecomment-1098128946, but we still have [one failure](https://app.circleci.com/pipelines/github/pytorch/vision/16777/workflows/cf05d413-2eea-4a70-bf5a-8f8a6ec748ad/jobs/1359645).

pmeier:
After devoting quite a significant time, I'm fairly confident it is not worth sinking any more time into this. My hunch is that this is `pytest` issue, but I'm not sure. 

Minimal steps to reproduce:

- Use Python >= 3.8
- Test a dataset that internally uses a `Demultiplexer`
- Have any data stored in the `_Demultiplexer().child_buffers`

Even with that, sometimes the error shows up, but sometimes it doesn't. If someone else wants to debug further:

- If you try to reproduce outside of the test suite, make sure `ResourceWarning`'s are visible. They are ignored by default by Python. `pytest` enables them.
- It seems that the error is pointing the original archive for being not closed rather the file handles extracted from it.

pmeier:
After some offline discussion with @NicolasHug, we think it is best to just ignore the warnings. I tested on

- [Python 3.8](https://app.circleci.com/pipelines/github/pytorch/vision/16797/workflows/127f0c83-748e-4a68-bb0c-73bb826b4857/jobs/1361263)
- [Python 3.9](https://app.circleci.com/pipelines/github/pytorch/vision/16798/workflows/68ae1780-5f99-424e-bb8c-e69d8107001f/jobs/1361352)
- [Python 3.10](https://app.circleci.com/pipelines/github/pytorch/vision/16799/workflows/8bb93830-4675-4e53-b99c-0c23e57b6a3c/jobs/1361448)

and CI was green.

ejguan:
After https://github.com/pytorch/pytorch/pull/76345 is landed into latest PyTorch nightly, the Error caused by `ResourceWarning` is eliminated in the TorchVision domain test for TorchData.
But, an Error pops up due to the lambda function from https://github.com/pytorch/vision/blob/1ac6e8b91b980b052324f77828a5ef4a6715dd66/test/test_prototype_builtin_datasets.py#L138

See: https://github.com/pytorch/data/runs/6195902638?check_suite_focus=true

pmeier:
> But, an Error pops up due to the lambda function from

I can't reproduce that in our own CI: https://app.circleci.com/pipelines/github/pytorch/vision/17117/workflows/612d3491-b1ec-4b2f-9dee-860a078c65c3/jobs/1387558

pmeier:
The failures in `torchdata` CI happen for Python 3.[89] on macOS and Python 3.[789] on Windows. The failure is

```
AttributeError: Can't pickle local object 'TestCommon.test_data_loader.<locals>.<lambda>'
```

which is correct, but this leads to more questions:

1. Why aren't we seeing this failure on Linux and even more peculiar not macOS when using Python 3.7?
2. Why aren't we getting a warning about the lambda usage?

ejguan:
> 1. Why aren't we seeing this failure on Linux and even more peculiar not macOS when using Python 3.7?

This is because `fork` is the default mp method for Linux and macOS (3.7). If you change starting method to spawn, the lambda function won't work.

> 2. Why aren't we getting a warning about the lambda usage?

`collate_fn` is an argument for DataLoader or for DataPipe. So, we didn't get chance to update it. We can add a warning to DataLoader when `collate_fn` is lambda and `num_workers>0`.

pmeier:
Superseded by #5914.

PR #5806, opened 2022-04-12
cleanup features._Feature

pmeier:
This started with an offline comment from @vfdev-5 why we are not calling `super().__torch_function__` in `features._Feature.__torch_function__` although this is [recommended by the documentation](https://github.com/pytorch/pytorch/blob/9a7bfaa92900e6e9c8e2979d4d588ded9d10a1f4/torch/_tensor.py#L1108-L1109). The problem is that there are no hooks to prevent [coercing the output into the input type](https://github.com/pytorch/pytorch/blob/9a7bfaa92900e6e9c8e2979d4d588ded9d10a1f4/torch/_tensor.py#L1192-L1196). Since we need this (see the docstring for the reason), we need to re-implement the functionality.

This PR cleans-up the functionality and adds a comment to explain the reasoning for not calling `super().__torch_function__`.


torchvision/prototype/features/_feature.py
```diff
@@ -70,12 +71,17 @@ def __torch_function__(
...
-        - :func:`torch.clone`
+        - :meth:`torch.Tensor.clone`
         - :meth:`torch.Tensor.to`
         """
-        kwargs = kwargs or dict()
+        # Since super().__torch_function__ has no hook to prevent the coercing of the output into the input type, we
+        # need to reimplement the functionality.
+
+        if not all(issubclass(cls, t) for t in types):
+            return NotImplemented
```
pmeier, line 81: That was missing in our implementation, but is present in [`super().__torch_function__`](https://github.com/pytorch/pytorch/blob/9a7bfaa92900e6e9c8e2979d4d588ded9d10a1f4/torch/_tensor.py#L1116-L1117).

torchvision/prototype/features/_feature.py
```diff
@@ -16,13 +16,14 @@ def __new__(
...
-                torch.as_tensor(data, dtype=dtype, device=device),  # type: ignore[arg-type]
-                requires_grad,
-            ),
+        return (
+            torch.as_tensor(  # type: ignore[return-value]
+                data,
+                dtype=dtype,
+                device=device,  # type: ignore[arg-type]
+            )
+            .as_subclass(cls)  # type: ignore[arg-type]
```
pmeier, line 25: @NicolasHug While looking into the dispatch, I found that this is recommended and documented way to create tensor subclasses. I recall, you mentioned we should look into finding such a solution when we went over the proposal.

torchvision/prototype/features/_feature.py
```diff
@@ -16,13 +16,14 @@ def __new__(
...
+            .requires_grad_(requires_grad)
```
pmeier, line 26: Given that we no longer use `torch.Tensor._make_subclass`, we could also drop the `required_grad` here. 

On one hand we don't expect users to set `requires_grad=True` given that the features are inputs rather than parameters. Plus, even if someone needs this, a simple `.requires_grad_(True)` call is sufficient.

On the other hand, we would diverging from the other tensor creation ops. For example, the ubiquitous [`torch.tensor`](https://pytorch.org/docs/master/generated/torch.tensor.html) takes `requires_grad`. Still, it also takes the `pin_memory` parameter, which we currently also don't support.

Thoughts?

datumbox: Features can require gradients in some ML problems, for example adversarial attacks.

pmeier: Let's keep it then.

torchvision/prototype/features/_feature.py
```diff
@@ -28,13 +29,14 @@ def __new__(
...
-                cast(_TensorBase, cls),
-                torch.as_tensor(data, dtype=dtype, device=device),  # type: ignore[arg-type]
-                requires_grad,
-            ),
+        return (
+            torch.as_tensor(  # type: ignore[return-value]
+                data,
+                dtype=dtype,  # type: ignore[arg-type]
+                device=device,  # type: ignore[arg-type]
+            )
```
datumbox, line 37: Do I understand correctly that here we use `as_tensor` only to ensure that the requested `dtype` and `device` are respected? Otherwise we could just be doing `data.as_subclass(cls)` because we don't have cases were `data` are numpy array. I wonder if that's faster than `data.to(dtype=dtype, device=device).as_subclass(cls)`?

vfdev-5: Data is not always a tensor as far as I understand

pmeier: Correct. Basically, the constructor of features currently accepts everything that can be turned into a tensor. This is on par with what `torch.Tensor` does, i.e

```pycon
>>> torch.Tensor([0, 1, 2])
tensor([0., 1., 2.])
```

datumbox: Doing `to()` seems faster.

I use this approach to measure:
```python
import timeit
import torch

data = torch.randint(0, 256, (32, 3, 128, 128))
n = 10000


print(f"Torch version: {torch.__version__}")
for device in [None, "cpu", "cuda"]:
    for dtype in [None, torch.uint8, torch.float16, torch.float32]:
        as_tensor_time = timeit.timeit(lambda: torch.as_tensor(data, device=device, dtype=dtype), number=n)
        to_time = timeit.timeit(lambda: data.to(device=device, dtype=dtype), number=n)
        winner = "as_tensor" if as_tensor_time < to_time else "to"
        print(f"Winner: {winner}(device={device}, dtype={dtype}) - as_tensor: {as_tensor_time}, to: {to_time}")
```

I'm having issues with our cluster and had to run it on an old PyTorch version but here is the output:
```
Torch version: 1.11.0.dev20211102
Winner: to(device=None, dtype=None) - as_tensor: 0.0055221691727638245, to: 0.003016095608472824
Winner: to(device=None, dtype=torch.uint8) - as_tensor: 0.9256818778812885, to: 0.9209173992276192
Winner: to(device=None, dtype=torch.float16) - as_tensor: 3.5527884736657143, to: 3.2720518335700035
Winner: to(device=None, dtype=torch.float32) - as_tensor: 0.6391827426850796, to: 0.6276023834943771
Winner: to(device=cpu, dtype=None) - as_tensor: 0.011921301484107971, to: 0.008555687963962555
Winner: to(device=cpu, dtype=torch.uint8) - as_tensor: 0.6593559384346008, to: 0.6507497057318687
Winner: to(device=cpu, dtype=torch.float16) - as_tensor: 2.140716001391411, to: 2.108065687119961
Winner: to(device=cpu, dtype=torch.float32) - as_tensor: 0.6577045619487762, to: 0.6373300738632679
Winner: to(device=cuda, dtype=None) - as_tensor: 13.870690949261189, to: 11.862402316182852
Winner: to(device=cuda, dtype=torch.uint8) - as_tensor: 4.089442141354084, to: 3.84384398534894
Winner: to(device=cuda, dtype=torch.float16) - as_tensor: 7.160290367901325, to: 7.144624650478363
Winner: to(device=cuda, dtype=torch.float32) - as_tensor: 10.704698234796524, to: 10.372353758662939
```

pmeier: As discussed offline, it is useful to keep the `torch.as_tensor` call to enable things like 

```py
label = features.Label(0)
```

vfdev-5 (approved):
Thanks @pmeier , lgtm

datumbox (approved):
LGTM, thanks!

PR #5811, opened 2022-04-12
Fix OxfordIIITPet dataset transformation

detkov:
Remove unused `transforms` which involves both image and label, change the `__get__` function so it transform image and label independently based on `transform` and `target_transform` variables. The same behavior can be found in all major datasets classes, like [CIFAR10](https://pytorch.org/vision/stable/_modules/torchvision/datasets/cifar.html#CIFAR10) and [CelebA](https://pytorch.org/vision/stable/_modules/torchvision/datasets/celeba.html#CelebA).

### DIFF
--- a/torchvision/datasets/oxford_iiit_pet.py
+++ b/torchvision/datasets/oxford_iiit_pet.py
@@ -42,7 +42,6 @@ class OxfordIIITPet(VisionDataset):
         root: str,
         split: str = "trainval",
         target_types: Union[Sequence[str], str] = "category",
-        transforms: Optional[Callable] = None,
         transform: Optional[Callable] = None,
         target_transform: Optional[Callable] = None,
         download: bool = False,
@@ -54,7 +53,7 @@ class OxfordIIITPet(VisionDataset):
             verify_str_arg(target_type, "target_types", self._VALID_TARGET_TYPES) for target_type in target_types
         ]
 
-        super().__init__(root, transforms=transforms, transform=transform, target_transform=target_transform)
+        super().__init__(root, transform=transform, target_transform=target_transform)
         self._base_folder = pathlib.Path(self.root) / "oxford-iiit-pet"
         self._images_folder = self._base_folder / "images"
         self._anns_folder = self._base_folder / "annotations"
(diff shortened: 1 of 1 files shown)


pmeier:
Hey @detkov and thanks for the PR. Unfortunately, your assessment is a little off:

> Remove unused `transforms` which involves both image and label, 

`transforms` is not unused. Since the dataset can return segmentation masks, you need the ability to transform the image as well as the target simultaneously. This is done through `transforms`.

> change the `__get__` function so it transform image and label independently based on `transform` and `target_transform` variables. 

Internally, if no `transforms` is passed, `transform` and `target_transform` are just merged into one `transforms`:

https://github.com/pytorch/vision/blob/467b841ac2650c962952bbee5a542edb835c98bc/torchvision/datasets/vision.py#L52-L53

https://github.com/pytorch/vision/blob/467b841ac2650c962952bbee5a542edb835c98bc/torchvision/datasets/vision.py#L88-L98

> The same behavior can be found in all major datasets classes, like [CIFAR10](https://pytorch.org/vision/stable/_modules/torchvision/datasets/cifar.html#CIFAR10) and [CelebA](https://pytorch.org/vision/stable/_modules/torchvision/datasets/celeba.html#CelebA).

As explained above, datasets that go beyond image classification, have a `transforms` attribute. For example:

- https://github.com/pytorch/vision/blob/467b841ac2650c962952bbee5a542edb835c98bc/torchvision/datasets/coco.py#L31
- https://github.com/pytorch/vision/blob/467b841ac2650c962952bbee5a542edb835c98bc/torchvision/datasets/voc.py#L77

---

Did you hit any problem that lead you to send this PR? Otherwise, I'm going to close it since it the implementation already does it what is supposed to be doing.

detkov:
Thanks for the explanation and sorry for the mess, I just didn't spend enough time to dive deep and not to open the PR.

Yes, I opened PR since I faced the issue of using this dataset along with [timm](https://github.com/rwightman/pytorch-image-models)'s [loader factory function](https://github.com/rwightman/pytorch-image-models/blob/01a0e25a67305b94ea767083f4113ff002e4435c/timm/data/loader.py#L202) that does not take into account `dataset.transforms` variable. Then looked at other datasets and was confused. But yes, it is not the problem of torchvision, thanks for your time.

Issue #5809, opened 2022-04-12
Number of classes in FGVCAircraft differs from documentation


detkov:
### 🐛 Describe the bug

In the [torchvision.datasets.FGVCAircraft](https://pytorch.org/vision/0.12/generated/torchvision.datasets.FGVCAircraft.html?highlight=fgvcaircraft#torchvision.datasets.FGVCAircraft) dataset there are 100 classes — not 102, as it said in the documentation and on the [official dataset's page](https://www.robots.ox.ac.uk/~vgg/data/fgvc-aircraft/).

How to reproduce:
```
from torchvision.datasets import FGVCAircraft 
ds = FGVCAircraft('./', split='train', download=True)
print(len(set(ds._labels)))
```

### Versions

```
PyTorch version: 1.11.0+cu102
Is debug build: False
CUDA used to build PyTorch: 10.2
ROCM used to build PyTorch: N/A

OS: Debian GNU/Linux 10 (buster) (x86_64)
GCC version: (Debian 8.3.0-6) 8.3.0
Clang version: Could not collect
CMake version: version 3.13.4
Libc version: glibc-2.10

Python version: 3.7.12 | packaged by conda-forge | (default, Oct 26 2021, 06:08:53)  [GCC 9.4.0] (64-bit runtime)
Python platform: Linux-4.19.0-18-cloud-amd64-x86_64-with-debian-10.12
Is CUDA available: True
CUDA runtime version: 11.0.221
GPU models and configuration: GPU 0: Tesla V100-SXM2-16GB
Nvidia driver version: 460.73.01
cuDNN version: Could not collect
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

Versions of relevant libraries:
[pip3] mypy-extensions==0.4.3
[pip3] numpy==1.19.5
[pip3] torch==1.11.0
[pip3] torchaudio==0.10.0+cu113
[pip3] torchvision==0.12.0
[conda] blas                      2.113                       mkl    conda-forge
[conda] blas-devel                3.9.0            13_linux64_mkl    conda-forge
[conda] cudatoolkit               11.1.1              h6406543_10    conda-forge
[conda] dlenv-pytorch-1-10-gpu    1.0.20220226     py37h0ee201a_0    file:///tmp/conda-pkgs
[conda] libblas                   3.9.0            13_linux64_mkl    conda-forge
[conda] libcblas                  3.9.0            13_linux64_mkl    conda-forge
[...]

PR #5814, opened 2022-04-13
Fix FGVCAircraft dataset documentation

detkov:
There is a mistake in FGVCAircraft dataset documentation. If you load dataset and sum up 'trainval' and 'test' splits, there are not 10 200, but 10 000 images. Also mistake is in the number of classes: if we choose `annotation_level` to be `variant`, then unique number of classes is going to be not 102, but 100, and with `manufacturer` level it is not 41, but 30.

Initially was explored in issue #5809.

### DIFF
--- a/torchvision/datasets/fgvc_aircraft.py
+++ b/torchvision/datasets/fgvc_aircraft.py
@@ -12,15 +12,15 @@ from .vision import VisionDataset
 class FGVCAircraft(VisionDataset):
     """`FGVC Aircraft <https://www.robots.ox.ac.uk/~vgg/data/fgvc-aircraft/>`_ Dataset.
 
-    The dataset contains 10,200 images of aircraft, with 100 images for each of 102
+    The dataset contains 10,000 images of aircraft, with 100 images for each of 100
     different aircraft model variants, most of which are airplanes.
     Aircraft models are organized in a three-levels hierarchy. The three levels, from
     finer to coarser, are:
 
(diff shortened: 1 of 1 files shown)


torchvision/datasets/fgvc_aircraft.py
```diff
@@ -12,15 +12,15 @@
 class FGVCAircraft(VisionDataset):
     """`FGVC Aircraft <https://www.robots.ox.ac.uk/~vgg/data/fgvc-aircraft/>`_ Dataset.
 
-    The dataset contains 10,200 images of aircraft, with 100 images for each of 102
+    The dataset contains 10,000 images of aircraft, with 100 images for each of 100
```
pmeier, line 15: Oh boy, they even got the number of samples wrong on their website? :roll_eyes: Thanks for checking and fixing!

pmeier (approved):
Thanks a ton @detkov. LGTM!

torchvision/datasets/fgvc_aircraft.py
NicolasHug: If the website is reporting the original values, are we doing something wrong in the code?

pmeier: I don't think so, no:

```shell
$ cat variants.txt | wc -l
100
$ cat families.txt | wc -l
70
$ cat manufacturers.txt | wc -l
30
$ ls images | wc -l
10000

NicolasHug (approved):
Thanks @detkov and @pmeier

PR #5820, opened 2022-04-14
the ` vision_transformer.py` assert msg problem

tcmyxc:
the assert msg should be same


tcmyxc:
the assert msg of  `forward` in `EncoderBlock`  shoule be the same as the `forward` in `Encoder` , the input dim is not change

datumbox (approved):
LGTM, thanks!

PR #5821, opened 2022-04-14
Model doc revamp - first PR

NicolasHug:
This PR introduces our new models docs, as discussed and designed in previous PRs (see below) . For now I just wrote docs for  VGG and ResNet models, and  generated the weights table for classification.

Rendered docs: https://output.circle-artifacts.com/output/job/aaddb6cd-567c-4552-90fa-296878ff5ab1/artifacts/0/docs/models.html

We now have:

- a [main Models page](https://output.circle-artifacts.com/output/job/aaddb6cd-567c-4552-90fa-296878ff5ab1/artifacts/0/docs/models.html) with a similar structure to what we have now (classification, detection, optical flow, etc.). This page contains links to:
    - [One page per model](https://output.circle-artifacts.com/output/job/aaddb6cd-567c-4552-90fa-296878ff5ab1/artifacts/0/docs/models/resnet.html) - i.e. one page for Resnet, one for VGG, etc. Each of these pages contain links to:
        - [one page per model builder](https://output.circle-artifacts.com/output/job/aaddb6cd-567c-4552-90fa-296878ff5ab1/artifacts/0/docs/models/generated/torchvision.models.resnet18.html#torchvision.models.resnet18). The model builders also document the possible Weights that can be passed as parameters.
    - [A table](https://output.circle-artifacts.com/output/job/aaddb6cd-567c-4552-90fa-296878ff5ab1/artifacts/0/docs/models.html#table-of-all-available-classificaiton-weights) summarizing all the available weights, and their accuracies. We'll do that for classification, detection, Optical flow, etc. 

----

TODO, in future PRs:

- Find a way to reduce the row size of the generated tables. They're too big right now. 
- Generate tables for detection, optical flow, etc.
- Document all the remaining models. Might involve external community on this, TBD.
- After that, write the "narrative" docs about how to use the new weights API, etc. Then go through the examples in the gallery and update them too.
- once everything is done: decide on a way to document the recipes. Right now we just link to a script or a PR or an issue, but there might be a more suitable way to do this.

Closes https://github.com/pytorch/vision/pull/5741
Closes https://github.com/pytorch/vision/pull/5577
Closes https://github.com/pytorch/vision/pull/5575


docs/Makefile
```diff
@@ -6,7 +6,7 @@ ifneq ($(EXAMPLES_PATTERN),)
 endif
 
 # You can set these variables from the command line.
-SPHINXOPTS    = -W -j auto $(EXAMPLES_PATTERN_OPTS)
+SPHINXOPTS    = -j auto $(EXAMPLES_PATTERN_OPTS)
```
NicolasHug, line 9: The -W flag will make the docs error on warnings. It's good to have to avoid having broken references. But since I removed almost all of the `models.rst` file, a few of the refs (e.g. from gallery examples) are broken. To avoid noise I'm removing the flag for now.

Will definitely put it back once everything is done.

datumbox: Should we add a TODO or create an issue to avoid forgetting?

docs/source/conf.py
```diff
@@ -292,5 +296,57 @@ def inject_minigalleries(app, what, name, obj, options, lines):
...
+                table.append((str(k), str(v)))
+            table = tabulate(table, tablefmt="rst")
+            lines += [".. table::", ""]
+            lines += textwrap.indent(table, " " * 4).split("\n")
+            lines.append("")
+
+
+def generate_classification_table():
+
+    weight_enums = [getattr(M, name) for name in dir(M) if name.endswith("Weights")]
```
datumbox, line 322:
```suggestion
    weight_enums = [getattr(M, name) for name in dir(M) if name.endswith("_Weights")]
```

docs/source/models.rst
```diff
@@ -22,819 +22,27 @@ keypoint detection, video classification, and optical flow.
...
+   models/resnet
+   models/vgg
 
-For details on how to plot the masks of such models, you may refer to :ref:`semantic_seg_output`.
 
-The pre-trained models have been trained on a subset of COCO train2017, on the 20 categories that are
-present in the Pascal VOC dataset. You can see more information on how the subset has been selected in
-``references/segmentation/coco_utils.py``. The classes that the pre-trained model outputs are the following,
-in order:
+Table of all available classificaiton weights
```
datumbox, line 37:
```suggestion
Table of all available classification weights
```

docs/source/models.rst
```diff
@@ -22,819 +22,27 @@ keypoint detection, video classification, and optical flow.
...
-Raft
-----
-
-.. autosummary::
-    :toctree: generated/
-    :template: function.rst
-
-    torchvision.models.optical_flow.raft_large
-    torchvision.models.optical_flow.raft_small
+TODO: Something similar to classification models: list of models + table of weights
```
datumbox, line 48: We need to consider Quantized Classification models like a separate submodule (similar to detection and classification) and add documentations, tables with accuracies etc.

docs/source/models.rst
```diff
@@ -22,819 +22,27 @@ keypoint detection, video classification, and optical flow.
...
-- `SSDlite <https://arxiv.org/abs/1801.04381>`_
-
-The pre-trained models for detection, instance segmentation and
-keypoint detection are initialized with the classification models
-in torchvision.
-
-The models expect a list of ``Tensor[C, H, W]``, in the range ``0-1``.
-The models internally resize the images but the behaviour varies depending
-on the model. Check the constructor of the models for more information. The
-output format of such models is illustrated in :ref:`instance_seg_output`.
```
datumbox, line 595: I understand that your plan is to add back this info on follow up PRs. My concern about removing this info now is that we will no longer have this info committed on our main branch making it harder to recover. This is likely to be more of a problem if we consider involving the community.

Have you considered removing only the bits that you actually tackle in this PR (for example the overall accuracy table for Classification) + ResNet/VGG model builders? This way the info of anything not completed will be maintained and cleaned up on follow up PRs.

An alternative way to maintain the info is to create a feature branch for the entire documentation work and merge it at the very end on main when you are ready (similar to what we did with Multi-weight support).

docs/source/conf.py
```diff
@@ -292,5 +296,57 @@ def inject_minigalleries(app, what, name, obj, options, lines):
...
+        lines.append("")
+        for field in obj:
+            lines += [f"**{str(field)}**:", ""]
+
+            table = []
+            for k, v in field.meta.items():
+                if k == "categories":
+                    continue
+                elif k == "recipe":
+                    v = f"`link <{v}>`__"
```
datumbox, line 312: I understand that this is in order to avoid printing the whole URL on the documentation and instead add a link. It doesn't seem to work [here](https://output.circle-artifacts.com/output/job/aaddb6cd-567c-4552-90fa-296878ff5ab1/artifacts/0/docs/models/generated/torchvision.models.resnet18.html#torchvision.models.resnet18) though. Is the specific rendering old?

datumbox (commented):
Thanks @NicolasHug.

Overall the proposal looks good. I've added a few comments below, let me know your thoughts.

docs/source/conf.py
```diff
@@ -292,5 +296,57 @@ def inject_minigalleries(app, what, name, obj, options, lines):
...
+
+def generate_classification_table():
+
+    weight_enums = [getattr(M, name) for name in dir(M) if name.endswith("Weights")]
+    weights = [w for weight_enum in weight_enums for w in weight_enum]
+
+    column_names = ("**Weight**", "**Acc@1**", "**Acc@5**", "**Params**", "**Recipe**")
+    content = [
+        (
+            f":class:`{w} <{type(w).__name__}>`",
```
datumbox, line 328: Shouldn't this make the enum values are not hyperlinks on the HTML page? They don't seem to be [here](https://output.circle-artifacts.com/output/job/aaddb6cd-567c-4552-90fa-296878ff5ab1/artifacts/0/docs/models.html#table-of-all-available-classificaiton-weights).

docs/source/conf.py
NicolasHug: Yeah, I added that as a commit later and didn't update the original message in the PR. Latest build is here https://output.circle-artifacts.com/output/job/8009776e-b101-426d-8766-98bd34ecb855/artifacts/0/docs/models/generated/torchvision.models.resnet18.html#torchvision.models.resnet18

docs/source/conf.py
NicolasHug: They are link, but only for those pages that exist. Ultimately this will be all of them, but right now this only works for Resnet and VGG weights. Try e.g. ResNet101_Weights.IMAGENET1K_V2

docs/source/models.rst
NicolasHug: I'm in favour of starting from a clean slate here. By keeping the previous text, it's likely that we will forget removing some obsolete info. Re-writing the narrative part of docs is part of the TODOs in my original comment above, so this is definitely something that we'll need to tackle eventually.

I'm not sure I get your point regarding this making this difficult if we involve the community. From my point of view, it will make things easier since we won't have to worry about contributors editing the text in various places. But perhaps I'm missing something?

Regarding the use of a new branch, are you concerned about the live docs at [https://pytorch.org/vision/main/](https://pytorch.org/vision/main/)  missing some info?

datumbox: Thanks for the additional info. I agree that the current docs are far from good and are noisy because they provide lots of information about the models in an unstructured way. For example:
1. They explain how to use the models during inference.
2. They talk about the training data and the normalization values.
3. They provide random info on where the pre-trained models are stored, the `train()/eval()` calls, the accuracies, etc

Some of this information might be obsolete (for example all the inference info should be re-written with the presets in mind) but some is still valid and just needs to be transferred to the right place (for example the category mapping between Pascal and COCO for segmentation, info related to where the pretrained weights are stored etc). My concern for removing the latter prematurely is that for a period the live docs will be missing critical information. Correct me if I'm wrong but I understand that after merging this, we won't even have a list of the available models linking from the documentation. This is why I said earlier that it could be hard for the community to restore some of this info (they would have to dig on the previous versions of the documentation and recover the useful parts which got deleted).

One approach to mitigate the above is to only remove the info you replace. We can follow the same approach in every PR sent by the community (aka that the transfer of info on the new format is correct and that things were cleaned up correctly). We can also ensure that a deep cleaning is done at the end by reviewing the documentation as a whole (similar to what we did with the final clean up steps at #4652).

As mentioned above, another approach could be to keep a feature branch (which will certainly allow you to start from a clean slate) but the problem with that is you don't want to leave this open for too long to avoid conflicts. Also the community won't be credited for their PRs because in the end only 1 PR will land on main. 

Let me know your thoughts. Happy to have an offline chat if you prefer to speed this up.

datumbox (approved):
Approving as in principle this fixes most of the major issues of the docs.

I recommend resolving https://github.com/pytorch/vision/pull/5821#discussion_r852919397 prior merging to avoid leaving the live docs in a bad state.

docs/source/models.rst
NicolasHug: Thanks for the details @datumbox ! As discussed offline, I moved the new models docs into a completely separate `models_new.rst` file. This keeps the current models docs untouched, and we'll be able to make PRs against `main`. Once all is done, all we'll have to do is to rename `models_new.rst` into `models.rst` (and fix all the broken references ;) ).

PR #5825, opened 2022-04-18
Add SimpleCopyPaste augmentation

lezwon:
This PR is related to #3817. It implements SimpleCopyPaste augmentation for segmenatation tasks.
https://arxiv.org/abs/2012.07177


datumbox:
@lezwon Thanks for kicking this off. Just a quick note, I would recommend moving this transform on detection instead of segmentation. MaskRCNN is actually run using the detection pipeline (I know it's confusing).

lezwon:
> @lezwon Thanks for kicking this off. Just a quick note, I would recommend moving this transform on detection instead of segmentation. MaskRCNN is actually run using the detection pipeline (I know it's confusing).

Sure will do that 👍

references/segmentation/transforms.py
```diff
@@ -98,3 +99,183 @@ def __init__(self, mean, std):
     def __call__(self, image, target):
         image = F.normalize(image, mean=self.mean, std=self.std)
         return image, target
+
+
+class ScaleJitter:
```
vadimkantorov, line 104: should this be a nn.Module like SimpleCopyPaste below?

lezwon:
hey @vadimkantorov @datumbox , I've moved the augmentation to the detection module. It is a basic functioning POC right now. It would be really nice to get some early feedback from ya'll :)  I have attached some samples for your reference.

image_0_a
image_0_mask
image_1_a
image_1_mask

references/detection/transforms.py
```diff
@@ -437,3 +439,78 @@ def forward(
                 )
 
         return image, target
+
+
+class SimpleCopyPaste(torch.nn.Module):
+    def __init__(self, jittering_type: str = "LSJ"):
```
datumbox, line 445: @lezwon feel free to ignore this for now but note that we shouldn't make SimpleCopyPaste use `ScaleJitter` and the other transforms internally. Instead those transforms can be assumed to have been applied before hand. This will allow you to decouple them.

Again feel free to ignore this for as long as you experiment. We can review the design once you are happy with the implementation. I just thought to give you an early heads up.

lezwon: @datumbox sure. will keep that in mind :) Thanks

lezwon:
@datumbox are tests/docs necessary for this PR?

references/detection/transforms.py
```diff
@@ -437,3 +439,67 @@ def forward(
...
+
+            # remove masks where no annotations are present (all values are 0)
+            mask_filter = source_data["masks"].sum((2, 1)).not_equal(0)
+            source_data["masks"] = source_data["masks"][mask_filter]
+            source_data["boxes"] = ops.masks_to_boxes(source_data["masks"])
+            source_data["labels"] = source_data["labels"][mask_filter]
+            source_data["area"] = source_data["area"][mask_filter]
+            source_data["iscrowd"] = source_data["iscrowd"][mask_filter]
+
+            # TODO: update area
```
lezwon, line 484: @datumbox I need some help calculating the updated area of the source masks. I'm not sure how to go about it.

datumbox: @vfdev-5 Do you have bandwidth to support this?

datumbox:
@lezwon Apologies for the delayed response. I was OOO and trying to catch up. Let me see if I can find someone who could support you on the remaining bit.

No need for doc/tests etc at this point. Especially since it's on the references, the implementation won't appear on main TorchVision (for now). Let's focus on completing the implementation, verifying it works as expected, reviewing the API and training a model with it.

references/detection/transforms.py
```diff
@@ -437,3 +439,67 @@ def forward(
...
+            # concatenate paste data with original data
+            source_data["masks"] = torch.cat((source_data["masks"], paste_data["masks"][random_selection]))
+            source_data["boxes"] = torch.cat((source_data["boxes"], paste_data["boxes"][random_selection]))
+            source_data["labels"] = torch.cat((source_data["labels"], paste_data["labels"][random_selection]))
+            source_data["area"] = torch.cat((source_data["area"], paste_data["area"][random_selection]))
+            source_data["iscrowd"] = torch.cat((source_data["iscrowd"], paste_data["iscrowd"][random_selection]))
+
+        # update the original images with paste images
+        paste_masks = torch.stack(paste_masks).to(torch.uint8)
+        paste_masks = T.GaussianBlur((5, 5), sigma=2)(paste_masks)  # Adds Gaussian Filter
```
vfdev-5, line 495: Can't we use functional op here instead of class and call ?

lezwon: @vfdev-5 Thanks for pointing that out. I have replaced it with the functional API.

vfdev-5:
@lezwon I was checking the paper, [official TF implementation](https://github.com/tensorflow/tpu/tree/master/models/official/detection/projects/copy_paste) and [unofficial pytorch one](https://github.com/conradry/copy-paste-aug).

In our case, let's proceed in the following way. Originally, copy paste transform works on dataset level and mixes current image with a paste image taken randomly from the dataset. In our case in order to avoid a mess we could have if dealing with datasets, we can start by copy/pasting data on batch level. For detection recipee, images and targets are tuple of tensors and tuple of dict (key->Tensor). CopyPaste transform should work on that level instead of single image/target pair. Thus, we have to add it on DataLoader when collating data:
```python
copypaste = CopyPaste()

def copypaste_collate_fn(batch):
    return copypaste(*utils_collate_fn(batch))

data_loader = torch.utils.data.DataLoader(
    dataset, batch_sampler=train_batch_sampler, num_workers=args.workers, collate_fn=copypaste_collate_fn
)
``` 

Here is an unfinished implemention that we could update in order to make this PR merged:
```
from torch import nn

def copy_paste(image, target, paste_image, paste_target):
    # implement copy/paste logic
    return paste_image, paste_target

class CopyPaste(nn.Module):
    
    def __init__(self, inplace=True):
        super().__init__()
        self.inplace = inplace
    
    def forward(self, images, targets=None):
         assert targets is not None
        
        # assert images is a tuple of Tensors
        # assert targets is a tuple of dict of key -> Tensor
        
        if not self.inplace:
            # clone images and targets
            pass
                
        # images = [t1, t2, ..., tN]
        # Let's define paste_images as shifted list of input images
        # in TF they mix data on the dataset level
        # paste_images = [t2, t3, ..., tN, t1]
        images = list(images)
        targets = list(targets)
        shift = 1
        le = len(images)        
        
        out_images = []
        out_targets = []
        for i in range(le):
            image, target = images[i], targets[i]
            print("image:", image.float().mean())
            paste_image, paste_target = images[(i + shift) % le], targets[(i + shift) % le]
            print("paste_image:", paste_image.float().mean())
            image, target = copy_paste(image, target, paste_image, paste_target)
            print("-> image:", image.float().mean())
            out_images.append(image)
            out_targets.append(target)
        
        return tuple(out_images), tuple(out_targets)
```


Let me know what do you think

datumbox:
@lezwon Let us know if what @vfdev-5 works out for you or if you need additional help. Thanks! 😃

lezwon:
@datumbox @vfdev-5 Really sorry for the delay. Been a bit occupied lately. I'll get back to this PR by Thursday is that okay :)

lezwon:
Hi @vfdev-5, I have a few doubts and questions:
1. My current implementation takes in a batch image tensor as `batch` and list of dicts as `target`.  I did the implementation based on `[RandomMixup](https://github.com/pytorch/vision/blob/289fce29b3e2392114aadbe7a419df0f2e3ac1be/references/classification/transforms.py#L9)` in classification transforms as I felt it would be much faster during processing.
2. Your implementation iterates through all the image-target pairs and passes them to `copy_paste`. The output is then saved to `out_images` and `out_targets`. Isn't running `copy_paste()` for each tensor separately slower?  Cant we update the entire batch in one operation?
3. In your implementation the output is returned by `copy_paste()`. I don't suppose the original images/target are being changed inplace here right? This is would nullify the purpose of `self.inplace`.
4. If the changes are being done in place within `copy_paste()`, we would have to clone `paste_image` and `paste_target` as we don't want to pick up any image which has been modified previously by `copy_paste()`.

My current implementation updates the targets individually first and then updates the entire image batch as I felt it would be more efficient. Please do let me know if this is incorrect, and needs improvement.

vfdev-5:
Hi @lezwon 

> My current implementation takes in a batch image tensor as batch and list of dicts as target. I did the implementation based on [RandomMixup](https://github.com/pytorch/vision/blob/289fce29b3e2392114aadbe7a419df0f2e3ac1be/references/classification/transforms.py#L9) in classification transforms as I felt it would be much faster during processing.

For detection recipee, images and targets are tuple of tensors and tuple of dict (key->Tensor). Input images can't be batched into a single tensor as they are of different sizes. Please check that with the script: https://github.com/pytorch/vision/blob/main/references/detection/train.py

> Your implementation iterates through all the image-target pairs and passes them to copy_paste. The output is then saved to out_images and out_targets. Isn't running copy_paste() for each tensor separately slower? Cant we update the entire batch in one operation?

I think, no due to above point.

> In your implementation the output is returned by copy_paste(). I don't suppose the original images/target are being changed inplace here right? This is would nullify the purpose of self.inplace.

We can either 1) add inplace arg to `copy_paste` and clone tensors inside or 2) inplement `copy_paste` as inplace op which returns the same refs as inputs: `image`, `target`

>  If the changes are being done in place within copy_paste(), we would have to clone paste_image and paste_target as we don't want to pick up any image which has been modified previously by copy_paste().

Yeah, good point, seems like we have to clone the list of images and targets in any case (inplace or not inplace)


What do you think ?

lezwon:
@vfdev-5 I have made the changes you mentioned. Mind having a look?

vfdev-5 (commented):
Sorry for delay, I forgot to submit my review from the last week. I'll playaround with your implementation to check it in details.

vfdev-5:
@lezwon here is how you can test your code (for example from jupyter notebook) if you have MSCoco dataset:
```python
# !wget https://raw.githubusercontent.com/pytorch/vision/main/references/detection/group_by_aspect_ratio.py
# !wget https://raw.githubusercontent.com/pytorch/vision/main/references/detection/transforms.py
# !wget https://raw.githubusercontent.com/pytorch/vision/main/references/detection/coco_utils.py

# import presets
import torch
import torch.utils.data
import torchvision

from coco_utils import get_coco, get_coco_kp
from group_by_aspect_ratio import GroupedBatchSampler, create_aspect_ratio_groups

def get_args_parser(add_help=True):
    import argparse

    parser = argparse.ArgumentParser(description="PyTorch Detection Training", add_help=add_help)

    parser.add_argument("--data-path", default="/datasets01/COCO/022719/", type=str, help="dataset path")
    parser.add_argument("--dataset", default="coco", type=str, help="dataset name")
    parser.add_argument("--model", default="maskrcnn_resnet50_fpn", type=str, help="model name")
    parser.add_argument("--device", default="cuda", type=str, help="device (Use cuda or cpu Default: cuda)")
    parser.add_argument(
        "-b", "--batch-size", default=2, type=int, help="images per gpu, the total batch size is $NGPU x batch_size"
    )
    parser.add_argument("--epochs", default=26, type=int, metavar="N", help="number of total epochs to run")
    parser.add_argument(
        "-j", "--workers", default=4, type=int, metavar="N", help="number of data loading workers (default: 4)"
    )
    parser.add_argument("--opt", default="sgd", type=str, help="optimizer")
    parser.add_argument(
        "--lr",
        default=0.02,
        type=float,
        help="initial learning rate, 0.02 is the default value for training on 8 gpus and 2 images_per_gpu",
    )
    parser.add_argument("--momentum", default=0.9, type=float, metavar="M", help="momentum")
    parser.add_argument(
        "--wd",
        "--weight-decay",
        default=1e-4,
        type=float,
        metavar="W",
        help="weight decay (default: 1e-4)",
        dest="weight_decay",
    )
    parser.add_argument(
        "--norm-weight-decay",
        default=None,
        type=float,
        help="weight decay for Normalization layers (default: None, same value as --wd)",
    )
    parser.add_argument(
        "--lr-scheduler", default="multisteplr", type=str, help="name of lr scheduler (default: multisteplr)"
    )
    parser.add_argument(
        "--lr-step-size", default=8, type=int, help="decrease lr every step-size epochs (multisteplr scheduler only)"
    )
    parser.add_argument(
        "--lr-steps",
        default=[16, 22],
        nargs="+",
        type=int,
        help="decrease lr every step-size epochs (multisteplr scheduler only)",
    )
    parser.add_argument(
        "--lr-gamma", default=0.1, type=float, help="decrease lr by a factor of lr-gamma (multisteplr scheduler only)"
    )
    parser.add_argument("--print-freq", default=20, type=int, help="print frequency")
    parser.add_argument("--output-dir", default=".", type=str, help="path to save outputs")
    parser.add_argument("--resume", default="", type=str, help="path of checkpoint")
    parser.add_argument("--start_epoch", default=0, type=int, help="start epoch")
    parser.add_argument("--aspect-ratio-group-factor", default=3, type=int)
    parser.add_argument("--rpn-score-thresh", default=None, type=float, help="rpn score threshold for faster-rcnn")
    parser.add_argument(
        "--trainable-backbone-layers", default=None, type=int, help="number of trainable layers of backbone"
    )
    parser.add_argument(
        "--data-augmentation", default="hflip", type=str, help="data augmentation policy (default: hflip)"
    )
    parser.add_argument(
        "--sync-bn",
        dest="sync_bn",
        help="Use sync batch norm",
        action="store_true",
    )
    parser.add_argument(
        "--test-only",
        dest="test_only",
        help="Only test the model",
        action="store_true",
    )

    parser.add_argument(
        "--use-deterministic-algorithms", action="store_true", help="Forces the use of deterministic algorithms only."
    )

    # distributed training parameters
    parser.add_argument("--world-size", default=1, type=int, help="number of distributed processes")
    parser.add_argument("--dist-url", default="env://", type=str, help="url used to set up distributed training")
    parser.add_argument("--weights", default=None, type=str, help="the weights enum name to load")
    parser.add_argument("--weights-backbone", default=None, type=str, help="the backbone weights enum name to load")

    # Mixed precision training parameters
    parser.add_argument("--amp", action="store_true", help="Use torch.cuda.amp for mixed precision training")

    return parser

args = get_args_parser().parse_args(["--data-path", "/data/MSCoco"])

def get_transform(train, args):
    assert train
    return T.Compose([
        T.RandomHorizontalFlip(p=0.5),        
        T.PILToTensor(),
        T.ConvertImageDtype(torch.float)
    ])

# Data loading code
dataset, num_classes = get_dataset(args.dataset, "train", get_transform(True, args), args.data_path)

train_sampler = torch.utils.data.RandomSampler(dataset)

group_ids = create_aspect_ratio_groups(dataset, k=args.aspect_ratio_group_factor)
train_batch_sampler = GroupedBatchSampler(train_sampler, group_ids, args.batch_size)

# ....
# CopyPaste Definition 
# ....

copypaste = CopyPaste()

def copypaste_collate_fn(batch):
    return copypaste(*utils_collate_fn(batch))

args.workers = 0
data_loader = torch.utils.data.DataLoader(
    dataset, batch_sampler=train_batch_sampler, num_workers=args.workers, collate_fn=copypaste_collate_fn
)
batch = next(iter(data_loader))
```

Right now, it fails with 
```
Input In [28], in CopyPaste.update_source_image(self, image, paste_image, paste_alpha_mask)
    106 """
    107 Copies the pixels from the paste_image to image using the paste_alpha_mask
    108 Args:
   (...)
    112 Returns:
    113 """
    114 paste_alpha_mask = F.gaussian_blur(paste_alpha_mask.unsqueeze(0), kernel_size=(5, 5), sigma=2.0)  # add blur
--> 115 image.mul_(torch.logical_not(paste_alpha_mask))  # Delete pixels from source image
    116 paste_pixels = paste_image * paste_alpha_mask  # Copy pixels from paste image
    117 image.add_(paste_pixels)

RuntimeError: The size of tensor a (480) must match the size of tensor b (382) at non-singleton dimension 2
```

lezwon:
@vfdev-5 I have gone through your review. Please allow me some time to make the changes. I shall get back soon :)

lezwon:
@vfdev-5 I have made the following chages:
- moved `copy_paste` into a function.
- Make only one copy of the batch and target
- make `area` and `iscrowd` optional for datasets.

One problem I faced was:
- In some cases, the length of the tuple of area/iscrowd is greater than the number of masks present. The code was failing here while trying to filter the iscrowd/area attributes. (Have commented the code and added as TODO)


I have used the transforms as follows:

```python
jittering_type = "SSJ"
if jittering_type == "LSJ":
    scale_range = (0.1, 2.0)
else:
    scale_range = (0.8, 1.25)

def get_transform(train, args):
    assert train
    return T.Compose([
        T.RandomHorizontalFlip(p=0.5),
        T.PILToTensor(),
        T.ScaleJitter(target_size=(1024, 1024), scale_range=scale_range),
        T.FixedSizeCrop(size=(1024, 1024), fill=105),
        T.ConvertImageDtype(torch.float)
    ])
```

Some results:
image_0_a
image_0_mask
image_2_a
image_2_mask

lezwon:
@vfdev-5 I'm not sure if the authors of the paper trained the model on fixed sizes. I see it mentioned `We train Mask R-CNN (ResNet-50 FPN) on 1024×1024 image size for varying numbers of epochs`. If this is the case, stacking the images as a tensor would be a better solution as it would provide speedup, right?

datumbox:
@lezwon I haven't read the paper recently, but I remember that they used fixed 1024x1024 images.

vfdev-5:
> @vfdev-5 I'm not sure if the authors of the paper trained the model on fixed sizes. I see it mentioned `We train Mask R-CNN (ResNet-50 FPN) on 1024×1024 image size for varying numbers of epochs`. If this is the case, stacking the images as a tensor would be a better solution as it would provide speedup, right?

@lezwon thanks for the updates and your work on this PR ! As for benchmarking various solutions, I think right now it is a bit premature. We can do some time measurements to have an idea of what it gives.
I'm currently trying to run your code with MSCoco (following my code snippet shared above) and I still have few issues:
```
Input In [11], in copy_paste(image, source_data, paste_image, paste_data, inplace)
     19 paste_alpha_mask = F.gaussian_blur(paste_alpha_mask.unsqueeze(0), kernel_size=(5, 5), sigma=2.0)  # add blur
     21 # Paste pixels from paste_image to image
---> 22 image.mul_(torch.logical_not(paste_alpha_mask))  # Delete pixels from source image
     23 paste_pixels = paste_image * paste_alpha_mask  # Copy pixels from paste image
     24 image.add_(paste_pixels)  # paste it on source image

RuntimeError: The size of tensor a (427) must match the size of tensor b (425) at non-singleton dimension 1
```

Anyway, if you do not mind I'll commit to this PR few fixes/improvements such that it could ready to be merged. We can discuss other improvements here and maybe apply them in follow-up PRs...

lezwon:
> ```
> Input In [11], in copy_paste(image, source_data, paste_image, paste_data, inplace)
>      19 paste_alpha_mask = F.gaussian_blur(paste_alpha_mask.unsqueeze(0), kernel_size=(5, 5), sigma=2.0)  # add blur
>      21 # Paste pixels from paste_image to image
> ---> 22 image.mul_(torch.logical_not(paste_alpha_mask))  # Delete pixels from source image
>      23 paste_pixels = paste_image * paste_alpha_mask  # Copy pixels from paste image
>      24 image.add_(paste_pixels)  # paste it on source image
> 
> RuntimeError: The size of tensor a (427) must match the size of tensor b (425) at non-singleton dimension 1
> ```

I noticed this issue occurred when I didn't apply the fixed crop transform `T.FixedSizeCrop(size=(1024, 1024), fill=105)`

> Anyway, if you do not mind I'll commit to this PR few fixes/improvements such that it could ready to be merged. We can discuss other improvements here and maybe apply them in follow-up PRs...

Sure thing. Do go ahead 👍 :)

references/detection/transforms.py
```diff
@@ -437,3 +439,99 @@ def forward(
...
+
+    paste_alpha_mask = F.gaussian_blur(paste_alpha_mask.unsqueeze(0), kernel_size=(5, 5), sigma=2.0)  # add blur
+
+    # Paste pixels from paste_image to image
+    image.mul_(torch.logical_not(paste_alpha_mask))  # Delete pixels from source image
+    paste_pixels = paste_image * paste_alpha_mask  # Copy pixels from paste image
+    image.add_(paste_pixels)  # paste it on source image
+
+    # update original masks
+    source_data["masks"] = source_data["masks"] ^ paste_alpha_mask & source_data["masks"]
```
vfdev-5, line 463: Any objections why this line to update original masks can't be a single op like:
```python
source_data["masks"] = source_data["masks"] * (~paste_alpha_mask)
```
?

references/detection/transforms.py
```diff
@@ -437,3 +439,99 @@ def forward(
...
         return image, target
+
+
+def copy_paste(image, source_data, paste_image, paste_data, inplace=True):
+
+    image = image.clone() if not inplace else image
+    source_data = copy.deepcopy(source_data) if not inplace else source_data
+
+    number_of_masks = len(paste_data["masks"])
+    random_selection = torch.randint(0, number_of_masks, (number_of_masks,)).unique()
```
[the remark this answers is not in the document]
lezwon: I added a random selection based on the paper. 
It mentions `we select a random subset of objects from one of the images and paste them onto the other image`

vfdev-5:
@lezwon if you would like to check this version of code, run on your examples and comment out if something is missing, it would be great. Thanks!

lezwon:
> @lezwon if you would like to check this version of code, run on your examples and comment out if something is missing, it would be great. Thanks!

@vfdev-5 just ran it on some sample images. It's looking great so far. Will review it in detail and point out any issues if present. :)

vfdev-5:
> Will review it in detail and point out any issues if present. :)

Thanks @lezwon !

Meanhwile, I changed previous cropping strategy to resize of the data to paste, if sizes are different. This way input image will have the same size after pasting data into it. Previously, common minimal size was found and it could have few problems with cropped targets and also input size can change its size.

references/detection/transforms.py
```diff
@@ -437,3 +439,121 @@ def forward(
...
+        paste_masks = F.resize(paste_masks, size1, interpolation=F.InterpolationMode.NEAREST)
+        # resize bboxes:
+        ratios = torch.tensor((size1[1] / size2[1], size1[0] / size2[0]), device=paste_boxes.device)
+        paste_boxes = paste_boxes.view(-1, 2, 2).mul(ratios).view(paste_boxes.shape)
+
+    paste_alpha_mask = paste_masks.sum(dim=0) > 0
+    paste_alpha_mask = F.gaussian_blur(paste_alpha_mask.unsqueeze(0), kernel_size=(5, 5), sigma=2.0)
+
+    # Copy-paste images:
+    image = (image * (~paste_alpha_mask)) + (paste_image * paste_alpha_mask)
```
lezwon, line 470: @vfdev-5 if `inplace` is `True`, I don't suppose this operation will make the changes on the original image right?

references/detection/transforms.py
```diff
@@ -437,3 +439,121 @@ def forward(
...
+        # Let's define paste_images as shifted list of input images
+        # paste_images = [t2, t3, ..., tN, t1]
+        # FYI: in TF they mix data on the dataset level
+        shift = 1
+        images_rolled = images[-shift:] + images[:-shift]
+        targets_rolled = targets[-shift:] + targets[:-shift]
+
+        if self.inplace:
+            images_rolled = copy.deepcopy(images_rolled)
+            targets_rolled = copy.deepcopy(targets_rolled)
```
lezwon, line 544: we could use shallow copy here too maybe?

references/detection/transforms.py
vfdev-5: Yes, you are right, we have to figure out if `inplace` arg can bring something helpful or we can simply drop it.

vfdev-5:
Right now, I encountered an issue in CopyPaste code with some batch during the training:
```
Original Traceback (most recent call last):                                                                                                                                       
  File "/usr/local/lib/python3.8/dist-packages/torch/utils/data/_utils/worker.py", line 302, in _worker_loop                                                                      
    data = fetcher.fetch(index)                                                                                                                                                   
  File "/usr/local/lib/python3.8/dist-packages/torch/utils/data/_utils/fetch.py", line 52, in fetch                                                                               
    return self.collate_fn(data)                                                                                                                                                  
  File "train.py", line 193, in copypaste_collate_fn                                                                                                                              
    return copypaste(*utils.collate_fn(batch))                                                                                                                                    
  File "/usr/local/lib/python3.8/dist-packages/torch/nn/modules/module.py", line 1130, in _call_impl                                                                              
  File "/vision/references/detection/transforms.py", line 551, in forward                                                                                                         
    output_image, output_data = _copy_paste(image, target, paste_image, paste_target, self.inplace)                                                                               
  File "/vision/references/detection/transforms.py", line 497, in _copy_paste                                                                                                     
    iscrowd = target["iscrowd"][non_all_zero_masks]                                                                                                                               
IndexError: The shape of the mask [14] at index 0 does not match the shape of the indexed tensor [15] at index 0 
```
Something is not aligned and rather complicated to reproduce due to randomness. Let's see if we could make it a bit more bulletproof

lezwon:
> Right now, I encountered an issue in CopyPaste code with some batch during the training:
> 
> ```
> Original Traceback (most recent call last):                                                                                                                                       
>   File "/usr/local/lib/python3.8/dist-packages/torch/utils/data/_utils/worker.py", line 302, in _worker_loop                                                                      
>     data = fetcher.fetch(index)                                                                                                                                                   
>   File "/usr/local/lib/python3.8/dist-packages/torch/utils/data/_utils/fetch.py", line 52, in fetch                                                                               
>     return self.collate_fn(data)                                                                                                                                                  
>   File "train.py", line 193, in copypaste_collate_fn                                                                                                                              
>     return copypaste(*utils.collate_fn(batch))                                                                                                                                    
>   File "/usr/local/lib/python3.8/dist-packages/torch/nn/modules/module.py", line 1130, in _call_impl                                                                              
>     return forward_call(*input, **kwargs)                                                                                                                                         
>   File "/vision/references/detection/transforms.py", line 551, in forward                                                                                                         
>     output_image, output_data = _copy_paste(image, target, paste_image, paste_target, self.inplace)                                                                               
>   File "/vision/references/detection/transforms.py", line 497, in _copy_paste                                                                                                     
>     iscrowd = target["iscrowd"][non_all_zero_masks]                                                                                                                               
> IndexError: The shape of the mask [14] at index 0 does not match the shape of the indexed tensor [15] at index 0 
> ```
> 
> Something is not aligned and rather complicated to reproduce due to randomness. Let's see if we could make it a bit more bulletproof

Ah yes, I faced this issue. The size of iscrowd and area does not match with masks, boxes for some images. Not sure about the reason for it in COCO dataset.

references/detection/transforms.py
vfdev-5: I decided to drop the arg

references/detection/transforms.py
```diff
@@ -437,3 +439,114 @@ def forward(
...
+    paste_boxes = paste_target["boxes"][random_selection]
+    paste_labels = paste_target["labels"][random_selection]
+
+    masks = target["masks"]
+    # If source and paste data have different sizes
+    # Let's resize paste data
+    size1 = image.shape[-2:]
+    size2 = paste_image.shape[-2:]
+    if size1 != size2:
+        paste_image = F.resize(paste_image, size1, interpolation=F.InterpolationMode.BICUBIC)
```
datumbox, line 460: This interpolation should probably be provided as argument to align with the interpolation mode used elsewhere.

references/detection/transforms.py
```diff
@@ -437,3 +439,114 @@ def forward(
...
+class SimpleCopyPaste(torch.nn.Module):
+    def forward(self, images, targets=None):
+        assert targets is not None
+        assert isinstance(images, tuple) and all([isinstance(v, torch.Tensor) for v in images])
+        assert isinstance(targets, tuple) and len(images) == len(targets)
+        for target in targets:
+            assert isinstance(target, dict)
+            for k in ["masks", "boxes", "labels"]:
+                assert k in target, f"Key {k} should be present in targets"
+                assert isinstance(target[k], torch.Tensor)
```
datumbox, line 529: Nit: We can keep this assertions for now but in general we've phased out their use from the library. We usually favour torch._assert which is FX-traceable and JIT-scriptable.

references/detection/transforms.py
```diff
@@ -437,3 +439,114 @@ def forward(
...
+
+        # images = [t1, t2, ..., tN]
+        # Let's define paste_images as shifted list of input images
+        # paste_images = [t2, t3, ..., tN, t1]
+        # FYI: in TF they mix data on the dataset level
+        shift = 1
+        images_rolled = images[-shift:] + images[:-shift]
+        targets_rolled = targets[-shift:] + targets[:-shift]
```
datumbox, line 537: Nit: Why not use `roll(1, 0)`?

datumbox (commented):
@vfdev-5 looks good. Just minor comments, let me know your thoughts.

references/detection/transforms.py
vfdev-5: images and targets are tuples and there is no `roll` method on them

references/detection/transforms.py
vfdev-5: OK, I can expose it as an argument. I put it hard-coded as resize and also gaussian_blur are just helping ops and not "essential" of the method.

references/detection/transforms.py
vfdev-5: Sounds good, let me replace that.

references/detection/transforms.py
vfdev-5: As seen offline, detection reference does not expose args for interpolation mode and has a bit particular way to resize image and mask inside the model. Let's keep it as it is for now.

references/detection/train.py
```diff
@@ -180,8 +189,27 @@ def main(args):
...
+        if args.data_augmentation in ["ssd", "ssdlite"]:
+            raise RuntimeError("SimpleCopyPaste algorithm does support 'ssd', 'ssdlite' data augmentation policies")
+
+        print("Use SimpleCopyPaste data aug")
+        if args.data_augmentation not in ["lsj"]:
+            print(
+                "INFO: SimpleCopyPaste is intended to work together with data-augmentation='lsj'. "
+                "Currently, the algorithm can work with any data-augmentation policy "
+                "but an additional resize is applied to the pasted data"
+            )
```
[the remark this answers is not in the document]
datumbox: @vfdev-5 what you had there was correct and perhaps I am being overcautious here. I'm considering merging this to be able to test it easier, so I take the safe approach of only exposing it on the LSJ augmentation as on the paper.

references/detection/transforms.py
```diff
@@ -437,3 +438,156 @@ def forward(
...
+    # Random paste targets selection:
+    num_masks = len(paste_target["masks"])
+
+    if num_masks < 1:
+        # Such degerante case with num_masks=0 can happen with LSJ
+        # Let's just return (image, target)
+        return image, target
+
+    # We have to please torch script by explicitly specifying dtype as torch.long
+    random_selection = torch.unique(torch.randint(0, num_masks, (num_masks,))).to(torch.long)
```
datumbox, line 461: Nit:
```suggestion
    random_selection = torch.unique(torch.randint(0, num_masks, (num_masks,), device=paste_image.device)).to(torch.long)
```

references/detection/transforms.py
```diff
@@ -437,3 +438,156 @@ def forward(
...
+                blending=self.blending,
+                resize_interpolation=self.resize_interpolation,
+            )
+            output_images.append(output_image)
+            output_targets.append(output_data)
+
+        return output_images, output_targets
+
+    def __repr__(self) -> str:
+        s = f"{self.__class__.__name__}()"
```
datumbox, line 592: Nit: expose `blending` and `resize_interpolation`

references/detection/train.py
```diff
@@ -145,6 +147,13 @@ def get_args_parser(add_help=True):
     # Mixed precision training parameters
     parser.add_argument("--amp", action="store_true", help="Use torch.cuda.amp for mixed precision training")
 
+    # Use CopyPaste augmentation training parameter
+    parser.add_argument(
+        "--use-copypaste",
+        action="store_true",
+        help="Use CopyPaste data augmentation. It is intended to work together with data-augmentation='lsj'.",
```
datumbox, line 154:
```suggestion
        help="Use CopyPaste data augmentation. Works only with data-augmentation='lsj'.",
```

datumbox (commented):
Thanks a lot! Last set of nits, promise.

references/detection/train.py
```diff
@@ -180,8 +189,27 @@ def main(args):
...
+            raise RuntimeError("SimpleCopyPaste algorithm does support 'ssd', 'ssdlite' data augmentation policies")
+
+        print("Use SimpleCopyPaste data aug")
+        if args.data_augmentation not in ["lsj"]:
+            print(
+                "INFO: SimpleCopyPaste is intended to work together with data-augmentation='lsj'. "
+                "Currently, the algorithm can work with any data-augmentation policy "
+                "but an additional resize is applied to the pasted data"
+            )
+        copypaste = SimpleCopyPaste(resize_interpolation=InterpolationMode.BILINEAR, blending=True)
```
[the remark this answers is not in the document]
vfdev-5: I left them explicit to remember once we exposed interpolation mode as cmd args. If you think we can hide them, let's do that.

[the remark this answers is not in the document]
datumbox: I'm OK to keep them if you prefer. :)

references/detection/transforms.py
```diff
@@ -437,3 +438,156 @@ def forward(
...
+    # This is something we introduced here as originally the algorithm works
+    # on equal-sized data (for example, coming from LSJ data augmentations)
+    size1 = image.shape[-2:]
+    size2 = paste_image.shape[-2:]
+    if size1 != size2:
+        paste_image = F.resize(paste_image, size1, interpolation=resize_interpolation)
+        paste_masks = F.resize(paste_masks, size1, interpolation=F.InterpolationMode.NEAREST)
+        # resize bboxes:
+        ratios = torch.tensor((size1[1] / size2[1], size1[0] / size2[0]), device=paste_boxes.device)
+        paste_boxes = paste_boxes.view(-1, 2, 2).mul(ratios).view(paste_boxes.shape)
```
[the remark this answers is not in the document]
vfdev-5: Right, with LSJ this branch is not executed

datumbox (approved):
@lezwon Thank you very very much for this piece of work. I'll merge and start immediately investigating if this strategy improves the accuracy of the FasterRCNN models. I'll get back to you as soon as I have results.

@vfdev-5 thanks a lot for the reviews and final touches on the PR.

datumbox:
BTW the failing tests are unrelated and they are tracked separately on a different issue. Merging.

datumbox:
Below I summarize the training results of using the specific augmentation.

- mini-batch-size=2:
    - FasterRCNN-v2: box mAP: 46.8 (internal job ids 16740 & 17380)
    - MaskRCNN-v2: box mAP: 47.1, mask mAP: 41.5 (internal job ids 16732 & 17040)
- mini-batch-size=4: Running

Due to the way the technique combines images within the batches, the mini-batch size is expected to affect the performance. I was expecting that a naive config of setting mini-batch-size=2 would have very detrimental effects but actually it doesn't hurt much. I'm currently running for larger batch-sizes where the method makes more sense and I'll update the results above as I get them.

PR #5834, opened 2022-04-19
Update and rename resnet.py to resnet_with_k.py

NotSure2732:
ability to vary k to change layer size (width) to obtain different sizes of ResNets

I am not trying to actually implement this on the torchvision github, just trying to use it myself.


datumbox:
@NotSure2732 Welcome to our community. 

From your comment I understand that your intention is not to merge this but rather to show-case it as a proof of concept. I confirm that this approach works. We don't have a use-case to add it in TorchVision, so I'm going to close the PR to avoid accidental merges. If would like to discuss making similar extensions, feel free to open an issue and let's discuss the proposed changes. Thanks! :)

Issue #5836, opened 2022-04-19
new model weights loading breaks backwards compatibility for existing models


d4l3k:
### 🐛 Describe the bug

The new model loading format breaks numerous models that depend on custom loading logic via `model_urls`. This is used a lot for various academic repos and will make many of them no longer run on PT 1.12.

```
import torchvision

torchvision.models.resnet.model_urls["resnet18"]
```

Examples from notable papers/partners:
* monodepth2 https://github.com/NianticLabs/monodepth2/blob/master/networks/resnet_encoder.py#L55
* packnet-sfm https://github.com/TRI-ML/packnet-sfm/blob/master/packnet_sfm/networks/layers/resnet/resnet_encoder.py#L54
* pytorch blog https://pytorch.org/blog/how-to-train-state-of-the-art-models-using-torchvision-latest-primitives/#using-the-legacy-api
* Microsoft ELL https://microsoft.github.io/ELL/tutorials/Importing-models/
* AWS Labs https://cs.github.com/awslabs/aws-cv-task2vec/blob/0f05c390d65335f9dee55fd5195c48c748cfb50f/models.py?q=resnet.model_urls%5B%27resnet#L67
* diffnet https://github.com/brandleyzhou/DIFFNet/blob/main/networks/resnet_encoder.py#L50
* manydepth https://github.com/nianticlabs/manydepth/blob/master/manydepth/networks/resnet_encoder.py#L61

Looking at all of github for just resnet there's hundreds of usages of `resnet.model_urls`: https://cs.github.com/?q=resnet.model_urls[

16,460 code results that use model_urls https://github.com/search?p=1&q=torchvision+model_urls&type=Code

### Versions

PyTorch version: 1.12.0.dev20220419+cu113
Is debug build: False
CUDA used to build PyTorch: 11.3
ROCM used to build PyTorch: N/A

OS: Arch Linux (x86_64)
GCC version: (GCC) 11.2.0
Clang version: Could not collect
CMake version: version 3.22.2
Libc version: glibc-2.35

Python version: 3.10.2 (main, Mar 15 2022, 12:43:00) [GCC 11.2.0] (64-bit runtime)
Python platform: Linux-5.16.11-arch1-1-x86_64-with-glibc2.35
Is CUDA available: False
CUDA runtime version: 11.6.55
GPU models and configuration: Could not collect
Nvidia driver version: Could not collect
cuDNN version: Probably one of the following:
/usr/lib/libcudnn.so.8.3.1
[...]

Issue #5768, opened 2022-04-06
Remove problematic fields from weight meta-data


datumbox:
### 🚀 The feature

Currently the meta-data of weights contain a few problematic fields that should be removed.

The first one is `publication_year` field which originally meant to store the year an architecture is published. Unfortunately the current approach, doesn't account for cases where a technique is a combination of 2 or more papers (which is especially the case for post-paper optimizations). Adding the paper on the documentation (as we currently do) should be enough.

Another problematic field is `interpolation` because it is stored twice (once on the transforms and once on the meta-data). There is no reason to have duplicate information so we should remove it.

cc @datumbox

PR #5842, opened 2022-04-20
Update meta-data documentation 

datumbox:
It's quite likely that on future versions of TorchVision, we will have to modify the meta-data stored along with the weights. I believe that those meta-data are documentation and not code and as a result we should be able to modify them (add, delete, edit fields) on the future.

This PR adds a note on the documentation to clarify that there are no BC guarantees on meta-data.

### DIFF
--- a/torchvision/models/_api.py
+++ b/torchvision/models/_api.py
@@ -28,7 +28,9 @@ class Weights:
         meta (Dict[str, Any]): Stores meta-data related to the weights of the model and its configuration. These can be
             informative attributes (for example the number of parameters/flops, recipe link/methods used in training
             etc), configuration parameters (for example the `num_classes`) needed to construct the model or important
-            meta-data (for example the `classes` of a classification model) needed to use the model.
+            meta-data (for example the `classes` of a classification model) needed to use the model. These meta-data
+            are considered part of the documentation and as a result there are no BC guarnatees (fields can be added,
+            removed and modified on future releases).
     """
 
     url: str


NicolasHug (approved):
LGTM, thanks @datumbox 

Should we rename the `meta` field into `_meta` as well?

NicolasHug:
My concern is that documenting that something is private is usually not enough for users not to use it. On top of that, I'm not sure these docs will actually get rendered anywhere (will they?).
Unless we add a leading underscore we can be 99% sure that users will depend on it; a different but relevant instance is the recent https://github.com/pytorch/vision/issues/5836.

I'm personally not too concerned about using `_meta`, but if we want to keep it clean perhaps we can just make `Weights` a regular class instead of a DataClass?

```py
class Weights:
    def __init__(self, url, transforms, meta):
        self.url = url
        self.transforms = transforms
        self._meta = meta
```

Unless there's some killer feature from dataclasses that I'm missing?

datumbox:
@NicolasHug I think the use of dataclasses is from the time when the enums stored the data directly in them. I don't think there is a specific feature we need. Let me try make the change here and see if all work well.

datumbox:
It's been a while since I've worked on this part of the code-base so I had to refresh my memory about it.

The meta-data as they are now store two types of info:
1. Meta fields with information crucial for the construction of the model. These can be guaranteed to be BC and can be accessed programmatically. Examples of these fields are: `categories`, `keypoint_names`, `backend` (for quantization) and `size`.
2. Documentation fields that just provide context / info about the weights. These are only there for documentation purposes and have no BC guarantees. Examples of these fields are: `recipe`, `num_params`, accuracy metrics such as `acc@1`/`acc@5 etc 

One clean option is to separate the two and provide different guarantees for each. A bit of an overkill but I don't see a better way if we want to be super consistent. Thoughts?

datumbox:
After sleeping over it, my preferred solution is not to generate a bunch of different doc fields but instead clean up the current ones (remove unnecessary fields see #5768) and restructure existing fields such as metrics. This way we can still offer a great level of BC, excluding things like URLs that are purely informational and can change / move on the future. Thoughts?

NicolasHug:
> One clean option is to separate the two and provide different guarantees for each. A bit of an overkill but I don't see a better way if we want to be super consistent. Thoughts?

Something like a `meta` BC field and a private `_doc` field would be fine, I think.

I'm not sure I fully get your last comment, happy to chat more :)

datumbox:
@NicolasHug Sorry for not being very clear.

I'm proposing cleaning up the meta-data fields to keep only the bare minimum for which we can offer BC compatibility. With the exclusion of recipe URLs which can change to reflect the most updated info, all other fields like categories, num_params etc can be considered frozen and BC.

datumbox:
I had an offline chat with @NicolasHug and we agreed to strip the `meta` field down to the bare minimum fields (PR #5848 already is going towards that direction) and maintain BC.

I'm closing this PR and I'll follow up with a new one that further cleans up the fields and restructures them.

PR #5844, opened 2022-04-21
fix INaturalist 2021_train checksum

### LINKED ISSUE #5817: INaturalist download is broken for version="2021_train"
*Originally reported in https://discuss.pytorch.org/t/inaturalist-download-crashes-towards-the-end/149006*

> *INaturalist* download proceeds fine up to the end, then crashes with this error:
> 
> ```py
> File "/scratch/user/myenvs/env/lib/python3.8/site-packages/torchvision/datasets/utils.py", line 152, in download_url
>     raise RuntimeError("File not found or corrupted.")
> ```
> Using torchvision automatic downloading of the `2021_train` dataset.

---

The error points to

https://github.com/pytorch/vision/blob/05eae32f9663bbecad10a8d367ccbec50130e2f5/torchvision/datasets/utils.py#L151-L152

So my guess is that the MD5 checksum that we have on record is wrong:

https://github.com/pytorch/vision/blob/05eae32f9663bbecad10a8d367ccbec50130e2f5/torchvision/datasets/inaturalist.py#L24

I'll download and check. Given that this archive is > 200GB, this will take a while. I've tried with `version="2021_valid"` and this works fine. Thus, I'm guessing this is not a systematic problem.

cc @pmeier @YosuaMichael

pmeier:
Fixes #5817.

### DIFF
--- a/torchvision/datasets/inaturalist.py
+++ b/torchvision/datasets/inaturalist.py
@@ -22,7 +22,7 @@ DATASET_MD5 = {
     "2017": "7c784ea5e424efaec655bd392f87301f",
     "2018": "b1c6952ce38f31868cc50ea72d066cc3",
     "2019": "c60a6e2962c9b8ccbd458d12c8582644",
-    "2021_train": "38a7bb733f7a09214d44293460ec0021",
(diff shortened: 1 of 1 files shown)


NicolasHug (approved):
Thanks @pmeier !

PR #5847, opened 2022-04-21
Document make html-noplot as the default doc building method

NicolasHug:
This PR documents `make html-noplot` *before* `make html` in our contributing guide for building the docs.

`make html-noplot` is enough in 99% of cases, and `make html` can throw a bunch of scary errors (looking at you ffmpeg example 👀 ).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -159,7 +159,7 @@ pip install -r requirements.txt
 
 ```bash
 cd docs
-make html
+make html-noplot
 ```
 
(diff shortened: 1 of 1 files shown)


NicolasHug:
Maybe we could have a way to bypass an example if there is some failure (the ffmpeg one isn't just a dependency problem but sometimes a dynamic loading issue as well). If someone wants to spend time on this, why not, but I think this will become less and less relevant as we add more examples. Eventually the doc building time will be dominated by the examples (if that's not already the case), and the need for skipping all example building will be even stronger than it is right now

Issue #5830, opened 2022-04-19
error: can't copy 'build\lib.win-amd64-3.7\pyd': doesn't exist or not a regular file


datumbox:
### 🐛 Describe the bug

The CI jobs [unittest_windows_gpu_py3.7](https://app.circleci.com/pipelines/github/pytorch/vision/16740/workflows/20601594-11bc-41b8-b409-a5e65518d60a/jobs/1356410) and [unittest_windows_cpu_py3.7](https://app.circleci.com/pipelines/github/pytorch/vision/16740/workflows/20601594-11bc-41b8-b409-a5e65518d60a/jobs/1356437) fail with:
```
error: can't copy 'build\lib.win-amd64-3.7\pyd': doesn't exist or not a regular file
```

### Versions

Main branch has been failing since 53c853282697819df2d756f5d30911fb9cf20cf3

cc @seemethere

seemethere:
This one doesn't look related to upstream changes since the file it's trying to copy comes from torchvision itself and not upstream torch

atalman:
Looks like the error happended in nightly between April 16 and 17. It happens during the linking:

Good linker string: 
/OUT:build\lib.win-amd64-3.7\torchvision\_C.pyd /IMPLIB:C:\Users\circleci\project\build\temp.win-amd64-3.7\Release\Users\circleci\project\torchvision\csrc\ops\autograd\_C.lib

Linker string with error:
/OUT:build\lib.win-amd64-3.7\torchvision\pyd /IMPLIB:C:\Users\circleci\project\build\temp.win-amd64-3.7\Release\Users\circleci\project\torchvision\csrc\ops\autograd\pyd.lib

Looks like some Global variable is not set correctly

malfet:
I suspect it's caused by https://github.com/pypa/setuptools/issues/3219
Which is easily verifiable by pinning `setuptools` to say 58.0.4, which I'm doing in https://github.com/pytorch/vision/pull/5868

atalman:
Yes looks like you are correct setuptools is the problem

PR #5848, opened 2022-04-21
Remove publication_year and interpolation meta

YosuaMichael:
resolving https://github.com/pytorch/vision/issues/5768

### DIFF
--- a/torchvision/models/vision_transformer.py
+++ b/torchvision/models/vision_transformer.py
@@ -321,17 +321,13 @@ def _vision_transformer(
 _COMMON_META = {
     "task": "image_classification",
     "architecture": "ViT",
-    "publication_year": 2020,
     "categories": _IMAGENET_CATEGORIES,
-    "interpolation": InterpolationMode.BILINEAR,
 }
 
 _COMMON_SWAG_META = {
     **_COMMON_META,
-    "publication_year": 2022,
     "recipe": "https://github.com/facebookresearch/SWAG",
     "license": "https://github.com/facebookresearch/SWAG/blob/main/LICENSE",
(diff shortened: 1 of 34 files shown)


torchvision/models/vision_transformer.py
```diff
@@ -318,20 +318,16 @@ def _vision_transformer(
...
 
 
-_COMMON_META = {
+_COMMON_META: Dict[str, Any] = {
     "task": "image_classification",
     "architecture": "ViT",
-    "publication_year": 2020,
     "categories": _IMAGENET_CATEGORIES,
-    "interpolation": InterpolationMode.BILINEAR,
 }
```
YosuaMichael, line 326: Add the type explicitly to avoid mypy triggering error

torchvision/models/vision_transformer.py
```diff
@@ -318,20 +318,16 @@ def _vision_transformer(
     return model
 
 
-_COMMON_META = {
+_COMMON_META: Dict[str, Any] = {
```
datumbox, line 321: I assume mypy is to blame for this, right?

datumbox (approved):
LGTM, thanks @YosuaMichael!

torchvision/models/vision_transformer.py
YosuaMichael: Yes, correct

datumbox:
The failures seem unrelated. I've raised the issues separately at #5850 and #5830

YosuaMichael:
> The failures seem unrelated. I've raised the issues separately at #5850 and #5830

I will wait until all the tests finished first, just to be safe. After that I will merge if no other test that break.

Issue #5837, opened 2022-04-20
distutils.spawn.find_executable not working to find ffmpeg


hubutui:
### 🐛 Describe the bug

distutils.spawn.find_executable not working to find ffmpeg

### Versions

https://github.com/pytorch/vision/blob/e8cb0bacd86c49e67a7e1a5f83c6da866bc451cf/setup.py#L343
for python 3.10,
```python
import distutils
distutils.spawn.find_executable("ffmpeg")
```
outputs:
```text
Traceback (most recent call last):
  File "<input>", line 1, in <module>
    distutils.spawn.find_executable("ffmpeg")
AttributeError: module 'distutils' has no attribute 'spawn'
```
so `ffmpeg_exe = distutils.spawn.find_executable("ffmpeg")` leads to `ffmpeg_exe = None`
use this instead:
```python
from distutils import spawn
spawn.find_executable('ffmpeg')
```
also,
```text
DeprecationWarning: The distutils package is deprecated an
d slated for removal in Python 3.12. Use setuptools or check PEP 632 for potent
ial alternatives

```

bjuncek:
@datumbox thanks for flagging, I'll take the liberty to replace this with `shutil.which` as it's recommended in [PEP632 migration advice](https://peps.python.org/pep-0632/#id11)

bjuncek:
Closed via #5849

PR #5849, opened 2022-04-21
Replace distutils.spawn with shutil.which per PEP632 in setup script

bjuncek:
Addressing #5837

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -102,7 +102,7 @@ def find_library(name, vision_include):
             conda_installed = library_found
         else:
             # Check if using Anaconda to produce wheels
-            conda = distutils.spawn.find_executable("conda")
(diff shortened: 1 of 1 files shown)


NicolasHug (approved):
Thanks @bjuncek , LGTM if green

bjuncek:
@NicolasHug the failing unit test seems unrelated to me but I might be missing something.

NicolasHug:
Yes it's safe to ignore as it's been failing for a while.

I'll just wait for @datumbox to take a look before merging. Since all jobs have run (and compiled torchvision) successfully I think this is safe to go on my side

datumbox (approved):
LGTM, thanks for checking. The specific job is failing for unrelated reasons, we should be OK.

PR #5852, opened 2022-04-21
Clean up purely informational fields from Weight Meta-data

datumbox:
Continuation of work from #5848 and #5842 to clean up the meta-data of the Multi-weight support API.

This PR:
- Removes purely informational fields such as `task` and `architecture` from the meta-data as these can be inferred from the Weights object.
- Removes the information field `size` which is already encoded in the transforms (similar to what we did for `interpolation`). The specific field has caused confusion over its meaning (see https://github.com/pytorch/vision/pull/5618#discussion_r831149575 and https://github.com/pytorch/vision/pull/5618#discussion_r831146488). 
- Adds `min_size` to all models and makes it a mandatory field.
- Removes the `quantization` informational field from the Quantized models. This is likely that we will need to restore a field like this on the near future (depending on the work on #5797) to differentiate between models quantized with Eager vs FX graph mode API etc, but let's postpone this until we know all the details.

With the above changes, the meta-data should contain fields that are intended to be manipulated programmatically hence we can guarantee BC. The only exception is the recipe URL which is there for informational purposes and can be changed to a different value on the future but this is need to to build an effective documentation. All purely informational fields, should be moved on the doc strings.


torchvision/models/vision_transformer.py
```diff
@@ -288,18 +288,11 @@ def _vision_transformer(
...
-        elif isinstance(weights.meta["size"], Sequence):
-            if len(weights.meta["size"]) != 2 or weights.meta["size"][0] != weights.meta["size"][1]:
-                raise ValueError(
-                    f'size: {weights.meta["size"]} is not valid! Currently we only support a 2-dimensional square and width = height'
-                )
-            _ovewrite_named_param(kwargs, "image_size", weights.meta["size"][0])
-        else:
+        if weights.meta["min_size"][0] != weights.meta["min_size"][1]:
             raise ValueError(
-                f'weights.meta["size"]: {weights.meta["size"]} is not valid, the type should be either an int or a Sequence[int]'
```
datumbox, line 301: Replace `size` with `min_size` which stores the same info in ViT. I also simplify the logic here because all the extra checks are not needed. We fully control which weights get deployed and putting the wrong info in will cause issues anyway on our existing tests.

datumbox (commented):
Some clarifications:

torchvision/models/detection/ssd.py
```diff
@@ -30,10 +30,7 @@ class SSD300_VGG16_Weights(WeightsEnum):
         url="https://download.pytorch.org/models/ssd300_vgg16_coco-b556d3b4.pth",
         transforms=ObjectDetection,
         meta={
-            "task": "image_object_detection",
-            "architecture": "SSD",
             "num_params": 35641826,
-            "size": (300, 300),
```
NicolasHug, line 36: Q: Do the detection models have a `min_size` too?

torchvision/models/vision_transformer.py
```diff
@@ -288,18 +288,11 @@ def _vision_transformer(
...
+                f'min_size: {weights.meta["min_size"]} is not valid! Currently we only support a 2-dimensional square and width = height'
```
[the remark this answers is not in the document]
NicolasHug: IIUC this check never ever gets covered anyway since we don't test against it directly?

NicolasHug (approved):
Thanks @datumbox , minor Qs but LGTM anyway

torchvision/models/detection/ssd.py
datumbox: Good question. No they don't because they do resize within the model. As a result even if you give them a (1, 1) size image, it will end up being somewhere between 800 and 1333 pixels. Depending on the ratios you can get with sizes in between, making it hard to define what the size actually is.

torchvision/models/vision_transformer.py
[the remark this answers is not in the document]
datumbox: pretty much. Any concerns replacing it with the following?
```
assert weights.meta["min_size"][0] == weights.meta["min_size"][1]
```

Issue #5833, opened 2022-04-19
Revamping our classification models docs


NicolasHug:
We're re-writing our models docs to make them clearer, simpler, and to properly document the [upcoming multi-weight API](https://pytorch.org/blog/introducing-torchvision-new-multi-weight-support-api/). This issue is about **adding docs for the classification models.**

Perhaps this is something @oke-aditya @YosuaMichael @lezwon @zhiqwang  would be interested in :) ?

Our latest new docs are currently [here](https://output.circle-artifacts.com/output/job/5578549b-3bfa-4059-abfa-a608ef52b36c/artifacts/0/docs/models_new.html) (this link is likely outdated by the time you look at it, but it doesn't matter; the skeleton is there). We created a separate section that will eventually be merged into the main one. We have documented a few models, but most of them are still missing. The list of models that still need docs is listed below. If you'd like to participate, please comment below with a message saying "I'm working on XYZ" where XYZ is a model, so that others don't pick the same as you do. To keep things simple, please submit one PR per model, but feel free to contribute more than one model.

### How to write docs for a model

Note: below are detailed instructions. This makes it look more complicated than it actually is. Don't be scared!

A great place to start is to look a the changes in [this PR that documents SqueezeNet](https://github.com/pytorch/vision/pull/5832/files). You'll need to do exactly the same for your model:

- Create a new `.rst` file in https://github.com/pytorch/vision/tree/main/docs/source/models. The file should [look like this](https://raw.githubusercontent.com/pytorch/vision/main/docs/source/models/squeezenet.rst), with a link to the original paper, and a list of the corresponding model builders. It should also mention the base model class and link to the `.py` file where it is defined.
- Update the list in https://github.com/pytorch/vision/blame/main/docs/source/models_new.rst to link to this new file (without the `.rst` suffix). Please keep the list alphabetically sorted
- Update the docstring of each new model builder, similarly to [this one](https://github.com/pytorch/vision/blob/e8cb0bacd86c49e67a7e1a5f83c6da866bc451cf/torchvision/models/squeezenet.py#L162-L182).
[...]

PR #5854, opened 2022-04-21
EfficientNet New Documentation

yassineAlouini:
In this PR, I add the EfficientNet documentation following the guidelines outlined [here](https://github.com/pytorch/vision/issues/5833).

By the way, I thought about being clever with the docstrings (since they  repeat between the different) but then refrained from being clever (better explicit than implicit etc). Here is the code I had in mind nonetheless:

`efficientnet_b0.__doc__ = EFFICIENTNET_DOCSTRING.format(submodel="bo")`


docs/source/models/efficientnet.rst
```diff
@@ -0,0 +1,31 @@
+EfficientNet
+============
+
+.. currentmodule:: torchvision.models
+
+The EfficientNet model is based on the `EfficientNet: Rethinking Model Scaling for Convolutional Neural Networks <https://arxiv.org/pdf/1905.11946.pdf>`__
```
NicolasHug, line 6: Let's link to the arxiv page instead of linking to the `.pdf` file, because this can trigger undesired downloads

torchvision/models/efficientnet.py
```diff
@@ -651,13 +651,23 @@ class EfficientNet_V2_L_Weights(WeightsEnum):
 def efficientnet_b0(
     *, weights: Optional[EfficientNet_B0_Weights] = None, progress: bool = True, **kwargs: Any
 ) -> EfficientNet:
-    """
-    Constructs a EfficientNet B0 architecture from
-    `"EfficientNet: Rethinking Model Scaling for Convolutional Neural Networks" <https://arxiv.org/abs/1905.11946>`_.
+    """EfficientNet B0 model architecture from the `"EfficientNet: Rethinking Model Scaling for Convolutional
+    Neural Networks" <https://arxiv.org/abs/1905.11946>`_ paper.
```
NicolasHug, line 655: here and in the rest of the docstrings, I think we can get rid of the quotes for a cleaner rendering

```suggestion
    """EfficientNet B0 model architecture from the `EfficientNet: Rethinking Model Scaling for Convolutional
    Neural Networks <https://arxiv.org/abs/1905.11946>`_ paper.
```

NicolasHug (commented):
Thanks @yassineAlouini , minor comment but this looks great!

> By the way, I thought about being clever with the docstrings

I started  with a similar idea as well, but because this makes us document the model builders in slightly unconventional ways (the docstring wouldn't be right below the thing it's documenting), we decided against it, at least for now

docs/source/models/efficientnet.rst
yassineAlouini: That's a good point indeed. :ok_hand:

yassineAlouini:
> Thanks @yassineAlouini , minor comment but this looks great!
> 
> > By the way, I thought about being clever with the docstrings
> 
> I started with a similar idea as well, but because this makes us document the model builders in slightly unconventional ways (the docstring wouldn't be right below the thing it's documenting), we decided against it, at least for now

Awesome, I will fix these small details this afternoon and start the `EfficientNet v2` one. :ok_hand:

docs/source/models/efficientnet.rst
```diff
@@ -0,0 +1,31 @@
+EfficientNet
+============
+
+.. currentmodule:: torchvision.models
+
+The EfficientNet model is based on the `EfficientNet: Rethinking Model Scaling for Convolutional Neural Networks <https://arxiv.org/pdf/1905.11946>`__
```
NicolasHug, line 6:
```suggestion
The EfficientNet model is based on the `EfficientNet: Rethinking Model Scaling for Convolutional Neural Networks <https://arxiv.org/abs/1905.11946>`__
```

NicolasHug (approved):
Thanks @yassineAlouini , LGTM!

yassineAlouini:
Thanks @NicolasHug.

PR #5859, opened 2022-04-21
Restructuring metrics meta-data 

datumbox:
This PR restructures the way we store metrics in the meta-data:
- Moves all model accuracy metrics in a `metrics` key on meta-data.
- Renames several metrics to improve naming consistency.
- Adds new tests to enforce schema fields and keep the number of fields low
- Fixes the documentation

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -347,8 +347,8 @@ def generate_classification_table():
     content = [
         (
             f":class:`{w} <{type(w).__name__}>`",
-            w.meta["acc@1"],
-            w.meta["acc@5"],
+            w.meta["metrics"]["acc@1"],
+            w.meta["metrics"]["acc@5"],
             f"{w.meta['num_params']/1e6:.1f}M",
             f"`link <{w.meta['recipe']}>`__",
         )
--- a/test/test_extended_models.py
+++ b/test/test_extended_models.py
@@ -79,20 +79,32 @@ def test_naming_conventions(model_fn):
 )
 @run_if_test_with_extended
 def test_schema_meta_validation(model_fn):
-    # TODO: add list of permitted fields
-    classification_fields = ["categories", "acc@1", "acc@5"]
+    # list of all possible supported high-level fields for weights meta-data
+    permitted_fields = {
+        "backend",
+        "categories",
+        "keypoint_names",
+        "license",
+        "metrics",
+        "min_size",
+        "num_params",
+        "recipe",
+        "unquantized",
+    }
+    # mandatory fields for each computer vision task
+    classification_fields = {"categories", ("metrics", "acc@1"), ("metrics", "acc@5")}
     defaults = {
-        "all": ["recipe", "num_params", "min_size"],
+        "all": {"metrics", "min_size", "num_params", "recipe"},
         "models": classification_fields,
-        "detection": ["categories", "map"],
-        "quantization": classification_fields + ["backend", "unquantized"],
-        "segmentation": ["categories", "mIoU", "acc"],
+        "detection": {"categories", ("metrics", "box_map")},
+        "quantization": classification_fields | {"backend", "unquantized"},
+        "segmentation": {"categories", ("metrics", "miou"), ("metrics", "pixel_acc")},
         "video": classification_fields,
-        "optical_flow": [],
+        "optical_flow": set(),
     }
     model_name = model_fn.__name__
     module_name = model_fn.__module__.split(".")[-2]
-    fields = set(defaults["all"] + defaults[module_name])
+    fields = defaults["all"] | defaults[module_name]
 
     weights_enum = _get_model_weights(model_fn)
     if len(weights_enum) == 0:
@@ -102,9 +114,10 @@ def test_schema_meta_validation(model_fn):
     incorrect_params = []
     bad_names = []
     for w in weights_enum:
-        missing_fields = fields - set(w.meta.keys())
-        if missing_fields:
-            problematic_weights[w] = missing_fields
+        missing_fields = fields - (set(w.meta.keys()) | set(("metrics", x) for x in w.meta.get("metrics", {}).keys()))
+        unsupported_fields = set(w.meta.keys()) - permitted_fields
+        if missing_fields or unsupported_fields:
+            problematic_weights[w] = {"missing": missing_fields, "unsupported": unsupported_fields}
         if w == weights_enum.DEFAULT:
             if module_name == "quantization":
                 # parameters() count doesn't work well with quantization, so we check against the non-quantized
--- a/torchvision/models/detection/faster_rcnn.py
+++ b/torchvision/models/detection/faster_rcnn.py
@@ -383,7 +383,9 @@ class FasterRCNN_ResNet50_FPN_Weights(WeightsEnum):
             **_COMMON_META,
             "num_params": 41755286,
             "recipe": "https://github.com/pytorch/vision/tree/main/references/detection#faster-r-cnn-resnet-50-fpn",
-            "map": 37.0,
+            "metrics": {
+                "box_map": 37.0,
+            },
         },
     )
     DEFAULT = COCO_V1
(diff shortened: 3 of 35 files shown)


torchvision/models/detection/faster_rcnn.py
```diff
@@ -383,7 +383,9 @@ class FasterRCNN_ResNet50_FPN_Weights(WeightsEnum):
             **_COMMON_META,
             "num_params": 41755286,
             "recipe": "https://github.com/pytorch/vision/tree/main/references/detection#faster-r-cnn-resnet-50-fpn",
-            "map": 37.0,
+            "metrics": {
+                "box_map": 37.0,
```
datumbox, line 387: Rename `map` to `box_map`

torchvision/models/detection/keypoint_rcnn.py
```diff
@@ -322,8 +322,10 @@ class KeypointRCNN_ResNet50_FPN_Weights(WeightsEnum):
             **_COMMON_META,
             "num_params": 59137258,
             "recipe": "https://github.com/pytorch/vision/issues/1606",
-            "map": 50.6,
-            "map_kp": 61.1,
+            "metrics": {
+                "box_map": 50.6,
+                "kp_map": 61.1,
```
datumbox, line 327: Rename `map_kp` to `kp_map`.

torchvision/models/detection/mask_rcnn.py
```diff
@@ -364,8 +364,10 @@ class MaskRCNN_ResNet50_FPN_Weights(WeightsEnum):
             **_COMMON_META,
             "num_params": 44401393,
             "recipe": "https://github.com/pytorch/vision/tree/main/references/detection#mask-r-cnn",
-            "map": 37.9,
-            "map_mask": 34.6,
+            "metrics": {
+                "box_map": 37.9,
+                "mask_map": 34.6,
```
datumbox, line 369: Rename `map_mask` to `mask_map`

torchvision/models/optical_flow/raft.py
```diff
@@ -525,10 +525,12 @@ class Raft_Large_Weights(WeightsEnum):
...
             "recipe": "https://github.com/princeton-vl/RAFT",
-            "sintel_train_cleanpass_epe": 1.4411,
-            "sintel_train_finalpass_epe": 2.7894,
-            "kitti_train_per_image_epe": 5.0172,
-            "kitti_train_f1-all": 17.4506,
+            "metrics": {
+                "sintel_train_cleanpass_epe": 1.4411,
+                "sintel_train_finalpass_epe": 2.7894,
+                "kitti_train_per_image_epe": 5.0172,
+                "kitti_train_fl_all": 17.4506,
```
datumbox, line 532: Renamed `*f1-all` to `fl_all`.

torchvision/models/optical_flow/raft.py
```diff
@@ -525,10 +525,12 @@ class Raft_Large_Weights(WeightsEnum):
             **_COMMON_META,
             "num_params": 5257536,
             "recipe": "https://github.com/princeton-vl/RAFT",
-            "sintel_train_cleanpass_epe": 1.4411,
-            "sintel_train_finalpass_epe": 2.7894,
-            "kitti_train_per_image_epe": 5.0172,
-            "kitti_train_f1-all": 17.4506,
+            "metrics": {
+                "sintel_train_cleanpass_epe": 1.4411,
```
datumbox, line 529: I'm not terribly happy with this approach but I left it as-is to get your input @NicolasHug. Happy to leave this on a follow up PR but here are some thoughts.

The need of providing accuracy metrics on multiple datasets is common and something we might face on other models (for example in Classification give metrics both for ImageNet1K V1/V2/Real datasets). In such cases, we might want to provide a map of accuracies one for each dataset. For example:
```python
"metrics": {
   # single metrics on the "best" available dataset (whichever this is for each weight) used for summarization
  "epe": 1.4411,
  "fl_all": 17.4506,
  # metrics across different datasets. Naming TBD
  "epe_breakdown": {
    "sintel_train_cleanpass": 1.4411,
    "sintel_train_finalpass": 2.7894,
    "kitti_train_per_image_epe": 5.0172,
  },
  "fl_all_breakdown": {
    "kitti_train": 17.4506,
   },
}
```

We need to figure out this detail before the release because making changes on the meta-data will be considered BC-breaking in our current definition.

torchvision/models/segmentation/deeplabv3.py
```diff
@@ -142,8 +142,10 @@ class DeepLabV3_ResNet50_Weights(WeightsEnum):
             **_COMMON_META,
             "num_params": 42004074,
             "recipe": "https://github.com/pytorch/vision/tree/main/references/segmentation#deeplabv3_resnet50",
-            "mIoU": 66.4,
-            "acc": 92.4,
+            "metrics": {
+                "miou": 66.4,
```
datumbox, line 146: Adopt all lower-case capitalization.

torchvision/models/segmentation/deeplabv3.py
```diff
@@ -142,8 +142,10 @@ class DeepLabV3_ResNet50_Weights(WeightsEnum):
...
+                "pixel_acc": 92.4,
```
datumbox, line 147: Rename `acc` to `pixel_acc` for all segmentation models.

test/test_extended_models.py
```diff
@@ -79,20 +79,32 @@ def test_naming_conventions(model_fn):
 )
 @run_if_test_with_extended
 def test_schema_meta_validation(model_fn):
-    # TODO: add list of permitted fields
-    classification_fields = ["categories", "acc@1", "acc@5"]
+    # list of all possible supported high-level fields for weights meta-data
```
datumbox, line 82: Here we store all permitted high-level fields on the meta-data. Passing anything other than this will cause failures. This aims to keep the number of permitted fields low but we are able to add more fields by editing this list.

test/test_extended_models.py
```diff
@@ -79,20 +79,32 @@ def test_naming_conventions(model_fn):
...
+        "categories",
+        "keypoint_names",
+        "license",
+        "metrics",
+        "min_size",
+        "num_params",
+        "recipe",
+        "unquantized",
+    }
+    # mandatory fields for each computer vision task
```
datumbox, line 94: Here we store mandatory fields for every computer vision task. A specific model can have more fields but those are the minimum supported ones. A good example for this is detection. All models must have `metrics.box_map` metric but some models might have more than 1.

test/test_extended_models.py
```diff
@@ -79,20 +79,32 @@ def test_naming_conventions(model_fn):
...
+    classification_fields = {"categories", ("metrics", "acc@1"), ("metrics", "acc@5")}
```
datumbox, line 95: Tuples encode nested fields in the maps.

docs/source/conf.py
```diff
@@ -347,8 +347,8 @@ def generate_classification_table():
     content = [
         (
             f":class:`{w} <{type(w).__name__}>`",
-            w.meta["acc@1"],
-            w.meta["acc@5"],
+            w.meta["metrics"]["acc@1"],
+            w.meta["metrics"]["acc@5"],
```
datumbox, line 351: Fixing documentation.

datumbox (commented):
Adding some highlights to assist review:

torchvision/models/optical_flow/raft.py
NicolasHug: I agree, this is very adhoc for now. Happy to discuss this in the near future.

NicolasHug (approved):
Thanks @datumbox , LGTM.

The rendering of the docs is a bit strange now:

image


But this is an easy fix that we can leave for later

datumbox:
@NicolasHug Thanks for the review. You are right it looks ugly; let's follow up on the next PR. Any pointers on where this should be fixed?

PR #5860, opened 2022-04-21
test: add functional vertical flip tests on segmentation mask

federicopozzi33:

### DIFF
--- a/test/test_prototype_transforms_functional.py
+++ b/test/test_prototype_transforms_functional.py
@@ -332,6 +332,12 @@ def crop_bounding_box():
         )
 
 
+@register_kernel_info_from_sample_inputs_fn
+def vertical_flip_segmentation_mask():
+    for mask in make_segmentation_masks(extra_dims=((), (4,))):
+        yield SampleInput(mask)
+
+
 @pytest.mark.parametrize(
     "kernel",
     [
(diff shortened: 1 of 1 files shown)


test/test_prototype_transforms_functional.py
```diff
@@ -332,6 +332,12 @@ def crop_bounding_box():
         )
 
 
+@register_kernel_info_from_sample_inputs_fn
+def vertical_flip_segmentation_mask():
+    for mask in make_segmentation_masks(extra_dims=((), (4,))):
```
vfdev-5, line 337: @federicopozzi33 can't we use default extra dims ?

test/test_prototype_transforms_functional.py
```diff
@@ -860,3 +866,26 @@ def test_correctness_crop_bounding_box(device, top, left, height, width, expecte
     )
 
     torch.testing.assert_close(output_boxes.tolist(), expected_bboxes)
+
+
+@pytest.mark.parametrize("device", cpu_and_gpu())
+def test_correctness_vertical_flip_segmentation_mask_on_fixed_input(device):
+    mask = torch.tensor(
```
vfdev-5, line 873: Personally I'd prefer to create them as torch zeros and assign specific values to one, if possible. Also reduce the number of lines and improve readability. What do you think @federicopozzi33

vfdev-5 (commented):
Thanks for the PR @federicopozzi33 
I left few comments

test/test_prototype_transforms_functional.py
federicopozzi33: Yep, done.

test/test_prototype_transforms_functional.py
federicopozzi33: I think it's a good idea.

I was creating a 5x5x3 just for having a nice indentation, and increasing readibility. But, IMO 3x3x3 is enough.

I also set the dtype, which was missing.

federicopozzi33:
@vfdev-5 I saw that in other previous PR, such as the one concerning the affine or rotate primitives, you add two tests for check the correctness: one with random segmentation masks and one with a fixed input.

I skipped the one with random segmentation masks for some reasons:
- I can't figure out how to build the expected mask without calling some "vertical flip" functions.
- I understand that it represents broader check, but then why should we have both? Maybe because the one with fixed input is more readable and clear to read/understand?

vfdev-5:
> I can't figure out how to build the expected mask without calling some "vertical flip" functions.

@federicopozzi33 I think it is enough as you did as vflip is not that complicated op.

> I understand that it represents broader check, but then why should we have both? Maybe because the one with fixed input is more readable and clear to read/understand?

Affine/rotate ops are configurable and the first correctness check is generating hundreds of subtests running on CPU only. On the other hand, fixed input correctness check is testing the op on a single test-case and on CUDA and CPU. This is a trade-off between CI runtime and exhaustive op checking.

vfdev-5 (approved):
@federicopozzi33 LGTM, thanks for the PR !

datumbox (approved):
LGTM, thanks!

datumbox:
@federicopozzi33 Ooops there are conflicts with main branch. Could you please resolve so that we can merge?

vfdev-5:
@datumbox I can fix them quickly right now

PR #5861, opened 2022-04-22
simplify boundingbox kernels with jit compatible ellipsis

pmeier:
Addresses https://github.com/pytorch/vision/pull/5781#discussion_r855411284. Cross posting my answer there:

*Good catch. I was under the impression `...` was not supported by `torch.jit.script` in general, but this seems fine:*

```py
@torch.jit.script
    y = x.clone()
    y[..., 0::2] += 1
    return y
```

*It seems, only the explicit indices are not supported:*

```py
@torch.jit.script
def bar(x: torch.Tensor) -> torch.Tensor:
    y = x.clone()
    y[..., [0, 2]] += 1
    return y
```

```
RuntimeError: 
[...]
def bar(x: torch.Tensor) -> torch.Tensor:
    y = x.clone()
    y[..., [0, 2]] += 1
    ~~~~~~~~~~~~~ <--- HERE
    return y
```

### DIFF
--- a/torchvision/prototype/transforms/functional/_geometry.py
+++ b/torchvision/prototype/transforms/functional/_geometry.py
@@ -401,18 +401,16 @@ def pad_bounding_box(
 ) -> torch.Tensor:
     left, _, top, _ = _FT._parse_pad_padding(padding)
 
-    shape = bounding_box.shape
-
     bounding_box = convert_bounding_box_format(
         bounding_box, old_format=format, new_format=features.BoundingBoxFormat.XYXY
-    ).view(-1, 4)
+    )
 
-    bounding_box[:, 0::2] += left
-    bounding_box[:, 1::2] += top
+    bounding_box[..., 0::2] += left
+    bounding_box[..., 1::2] += top
 
     return convert_bounding_box_format(
         bounding_box, old_format=features.BoundingBoxFormat.XYXY, new_format=format, copy=False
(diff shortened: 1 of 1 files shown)


datumbox (approved):
LGTM.

If you look into the codebase for `.view(-1, 4)` you will find multiple other occurrences (for example at `horizontal_flip_bounding_box`, `vertical_flip_bounding_box`, `affine_bounding_box` etc. Should we fix this as well?

@vfdev-5 thoughts?

pmeier:
The examples you mentioned are all using explicit indices and thus cannot simplified as explained above:

https://github.com/pytorch/vision/blob/a64c6749b53fbd423e73ff42bf82bdecf6e0a219/torchvision/prototype/transforms/functional/_geometry.py#L31

https://github.com/pytorch/vision/blob/a64c6749b53fbd423e73ff42bf82bdecf6e0a219/torchvision/prototype/transforms/functional/_geometry.py#L97

https://github.com/pytorch/vision/blob/a64c6749b53fbd423e73ff42bf82bdecf6e0a219/torchvision/prototype/transforms/functional/_geometry.py#L239

vfdev-5:
IMO, view should be very tiny op. Great that we could simplify the code for this case and I think it is not a big deal that we couldn't for other cases... Maybe later when JIT will support that.
I think we can merge this PR.

PR #5862, opened 2022-04-22
Removing cuda115 since we migrated to cuda116

atalman:
Removing cuda115 since we migrated to cuda116


datumbox (approved):
LGTM, thanks @atalman. Feel free to merge once the CI turns green.

Issue #5782, opened 2022-04-07
[RFC] Implement transforms primitives for Segmentation Masks 


vfdev-5:
Related tracker issue: #5754

## Functional transforms for segmentation masks

The idea is to implement the following low-level functional transforms for segmentation masks:
https://github.com/pytorch/vision/blob/main/torchvision/prototype/transforms/functional/_geometry.py

- [x] affine_segmentation_mask (@vfdev-5) #5613
- [x] rotate_segmentation_mask (@vfdev-5) #5692
- [x] perspective_segmentation_mask (@vfdev-5) #5888
- [x] center_crop_segmentation_mask (@federicopozzi33) #5961
- [x] crop_segmentation_mask (@vfdev-5) #5851
- [x] resized_crop_segmentation_mask (@vfdev-5) #5855
- [x] pad_segmentation_mask (@federicopozzi33) #5866, need more tests: https://github.com/pytorch/vision/pull/5866/files#r859158083 -> https://github.com/pytorch/vision/pull/6104
- [x] vertical_flip_segmentation_mask (@federicopozzi33) #5860, #5633

## How to test the implementation ?

Put tests into `test/test_prototype_transforms_functional.py`:
- correctness test
- eager vs scripted
- scriptable
- check on cpu and cuda

cc @vfdev-5 @datumbox @bjuncek

federicopozzi33:
Hi @vfdev-5,

I would like to work on _vertical_flip_segmentation_mask_ primitive.

vfdev-5:
@federicopozzi33 sure, go ahead!

vadimkantorov:
Related: https://github.com/pytorch/vision/issues/4415

federicopozzi33:
@vfdev-5,

I'd like to go on with _pad_segmentation_mask_ primitive.

vfdev-5:
Thanks @federicopozzi33 , sure go ahead !

federicopozzi33:
I'd like to complete this issue doing _center_crop_segmentation_mask_.

vfdev-5:
@pmeier in order to close this tracker as done we have to add more correctness tests for padding with other padding modes (https://github.com/pytorch/vision/pull/5866/files#r859158083)

`pad_segmentation_mask` uses `pad_image_tensor` and right now we does not have any tests for `pad_image_tensor`. I suggest to check correctness with one of suggested ways (from the link above). Later these tests could be reused for checking `pad_image_tensor`.

[...]

PR #5866, opened 2022-04-22
feat: add functional pad on segmentation mask

federicopozzi33:
Part of #5782

### DIFF
--- a/test/test_prototype_transforms_functional.py
+++ b/test/test_prototype_transforms_functional.py
@@ -369,6 +369,7 @@ def resized_crop_segmentation_mask():
     ):
         yield SampleInput(mask, top=top, left=left, height=height, width=width, size=size)
 
+
 @register_kernel_info_from_sample_inputs_fn
 def pad_segmentation_mask():
     for mask, padding, padding_mode in itertools.product(
@@ -376,19 +377,12 @@ def pad_segmentation_mask():
         [[1], [1, 1], [1, 1, 2, 2]],  # padding
         ["constant", "symmetric", "edge", "reflect"],  # padding mode,
     ):
-        if padding_mode == "symmetric" and mask.ndim not in [3, 4]:
+        if padding_mode == "symmetric" and mask.ndim not in [2, 3, 4]:
             continue
-        if (
-            padding_mode == "edge"
-            and len(padding) == 2
-            and mask.ndim not in [2, 3]
-            or len(padding) == 4
-            and mask.ndim not in [4, 3]
-            or len(padding) == 1
-        ):
+
+        if (padding_mode == "edge" or padding_mode == "reflect") and mask.ndim not in [2, 3, 4]:
             continue
-        if padding_mode == "edge" and mask.ndim not in [2, 3, 4, 5]:
-            continue
+
         yield SampleInput(mask, padding=padding, padding_mode=padding_mode)
 
 
@@ -1054,6 +1048,7 @@ def test_correctness_resized_crop_segmentation_mask(device, top, left, height, w
     output_mask = F.resized_crop_segmentation_mask(in_mask, top, left, height, width, size)
     torch.testing.assert_close(output_mask, expected_mask)
 
+
 def test_correctness_pad_segmentation_mask_on_fixed_input(device):
     mask = torch.ones((1, 3, 3), dtype=torch.long, device=device)
 
@@ -1064,24 +1059,22 @@ def test_correctness_pad_segmentation_mask_on_fixed_input(device):
     torch.testing.assert_close(out_mask, expected_mask)
 
 
-@pytest.mark.parametrize("padding", [[1, 2, 3, 4], [1], 1, 1.0, [1, 2]])
+@pytest.mark.parametrize("padding", [[1, 2, 3, 4], [1], 1, [1, 2]])
 def test_correctness_pad_segmentation_mask(padding):
-    def _parse_padding():
-        if isinstance(padding, int):
-            return [padding] * 4
-        if isinstance(padding, float):
-            return [int(padding)] * 4
-        if isinstance(padding, list):
-            if len(padding) == 1:
-                return padding * 4
-            if len(padding) == 2:
-                return padding * 2  # [left, up, right, down]
+    def _compute_expected_mask():
+        def parse_padding():
+            if isinstance(padding, int):
+                return [padding] * 4
+            if isinstance(padding, list):
+                if len(padding) == 1:
+                    return padding * 4
+                if len(padding) == 2:
+                    return padding * 2  # [left, up, right, down]
 
-        return padding
+            return padding
 
-    def _compute_expected_mask(padding):
         h, w = mask.shape[-2], mask.shape[-1]
-        pad_left, pad_up, pad_right, pad_down = padding
+        pad_left, pad_up, pad_right, pad_down = parse_padding()
 
         new_h = h + pad_up + pad_down
         new_w = w + pad_left + pad_right
@@ -1092,10 +1085,8 @@ def test_correctness_pad_segmentation_mask(padding):
 
         return expected_mask
 
-    padding = _parse_padding()
-
     for mask in make_segmentation_masks():
         out_mask = F.pad_segmentation_mask(mask, padding, "constant")
 
-        expected_mask = _compute_expected_mask(padding)
+        expected_mask = _compute_expected_mask()
         torch.testing.assert_close(out_mask, expected_mask)
--- a/test/test_prototype_transforms_functional.py
+++ b/test/test_prototype_transforms_functional.py
@@ -371,15 +371,12 @@ def resized_crop_segmentation_mask():
 
 @register_kernel_info_from_sample_inputs_fn
 def pad_segmentation_mask():
-    for mask, padding, fill, padding_mode in itertools.product(
+    for mask, padding, padding_mode in itertools.product(
         make_segmentation_masks(),
         [[1], [1, 1], [1, 1, 2, 2]],  # padding
-        [0, 1],  # fill
         ["constant", "symmetric", "edge"],  # padding mode,
     ):
         if padding_mode == "symmetric" and mask.ndim not in [3, 4]:
-            continue
-        if padding_mode == "edge" and fill != 0:
             continue
         if (
             padding_mode == "edge"
@@ -392,7 +389,7 @@ def pad_segmentation_mask():
             continue
         if padding_mode == "edge" and mask.ndim not in [2, 3, 4, 5]:
             continue
-        yield SampleInput(mask, padding=padding, fill=fill, padding_mode=padding_mode)
+        yield SampleInput(mask, padding=padding, padding_mode=padding_mode)
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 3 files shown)


federicopozzi33:
Hi @vfdev-5,

I have some questions about pad function signature.

Hi @vfdev-5,

I have some questions about this PR.

1. Delving a little bit into the `pad` function implementation, it seems that typing is not correct.

`def pad(img: Tensor, padding: List[int], fill: int = 0, padding_mode: str = "constant") -> Tensor:`

In particular,`padding` and `fill` should be `Union[int, List[int], Tuple[int]]` and `Union[int, float]` respectively. 

2. Implementing the `pad_segmentation_mask`, I asked myself if I should expose `fill` and `padding_mode` parameters. In particular, segmentation masks shouldn't be binary? If so, should I check the `fill` value to be in `[0, 1]`?

vfdev-5:
> Delving a little bit into the [pad](https://github.com/pytorch/vision/blob/184b136f205df268f9bcfec8b6a89425d08fd1e8/torchvision/transforms/functional_tensor.py#L373) function implementation, it seems that typing is not correct.

@federicopozzi33 this may be due to torch jit script limitation of using Union which started being supported after the time pad was coded for tensors.

> Implementing the pad_segmentation_mask, I asked myself if I should expose fill and padding_mode parameters. In particular, segmentation masks shouldn't be binary? If so, should I check the fill value to be in [0, 1]?

In case of `fill`, this can be a feature request later if user would like to have background class other then 0. For existing ops with `fill`, I think we do not expose it for masks. 
https://github.com/pytorch/vision/blob/184b136f205df268f9bcfec8b6a89425d08fd1e8/torchvision/prototype/transforms/functional/_geometry.py#L297-L304

As for `padding_mode` I think it should be supported. For example modes like "replicate" recreate usable image data on images and it would make sense to do that on masks.

> segmentation masks shouldn't be binary?

I'd say no as it could have in general N classes

federicopozzi33:
> > segmentation masks shouldn't be binary?
> 
> I'd say no as it could have in general N classes

Ok, I was just thinking of one mask per class, in which 0 is the "background" class, and 1 represents class pixels.

Anyway, I think you can start looking at the PR.

In the meanwhile, I'll try figure out how to test other padding modes other than _constant_, that seems not so straightforward.

datumbox:
FYI, pad is busted and took it with it TorchVision's CI (see #5873). So any failures you see on the CI are not necessarily related to this PR.

test/test_prototype_transforms_functional.py
```diff
@@ -352,6 +352,29 @@ def vertical_flip_segmentation_mask():
...
+            padding_mode == "edge"
+            and len(padding) == 2
+            and mask.ndim not in [2, 3]
+            or len(padding) == 4
+            and mask.ndim not in [4, 3]
+            or len(padding) == 1
+        ):
+            continue
+        if padding_mode == "edge" and mask.ndim not in [2, 3, 4, 5]:
+            continue
```
vfdev-5, line 374: @federicopozzi33 can you comment out why we skip so much test cases ?

I printed locally which cases it skips and now quite sure about why it is so. For example
```
print(padding_mode, padding, mask.ndim, mask.shape)
>
constant [1] 3 torch.Size([1, 16, 16])
```

test/test_prototype_transforms_functional.py
```diff
@@ -933,3 +956,51 @@ def test_correctness_vertical_flip_segmentation_mask_on_fixed_input(device):
...
+        new_h = h + pad_up + pad_down
+        new_w = w + pad_left + pad_right
+
+        new_shape = (*mask.shape[:-2], new_h, new_w) if len(mask.shape) > 2 else (new_h, new_w)
+        expected_mask = torch.zeros(new_shape, dtype=torch.long)
+        expected_mask[..., pad_up:-pad_down, pad_left:-pad_right] = mask
+
+        return expected_mask
+
+    padding = _parse_padding()
```
vfdev-5, line 1000: I'm a bit confused about this method and what it does. Why it "unroll" padding argument into a list of 4 values for `F.pad_segmentation_mask(mask, padding, "constant")` ?

vfdev-5 (commented):
Thanks for the PR @federicopozzi33 
I left few questions. Let me think about how to test this op for all other params

test/test_prototype_transforms_functional.py
federicopozzi33: The purpose of `_parse_padding` is just to "unroll" padding argument ... for `_compute_expected_mask` only.

I moved the function where it is used, now it should be correct and more clear.

test/test_prototype_transforms_functional.py
federicopozzi33: I fixed. I was skipping too much tests, you're right.

The cases to be skipped are the following.

**First case:**
```
if padding_mode == "symmetric" and mask.ndim not in [2, 3, 4]:
    print(f"skipped: mask_shape={mask.shape}, ndim={mask.ndim}, padding={padding}, padding_mode={padding_mode}")
    continue
```       
Reason: 
`RuntimeError: Symmetric padding of N-D tensors are not supported yet` in `torchvision.transforms.functional_tensor._pad_symmetric`

**Second case:**
```
if (padding_mode == "edge" or padding_mode == "reflect") and mask.ndim not in [2, 3, 4]:
    print(f"skipped: mask_shape={mask.shape}, ndim={mask.ndim}, padding={padding}, padding_mode={padding_mode}")
    continue
```
Reason:
`NotImplementedError: Only 2D, 3D, 4D, 5D padding with non-constant padding are supported for now` in `torch.nn.functional.pad`

> So why am I also skipping 5D? 

You have to keep in mind that `len(padding)` is always equal to 4 (since it's parsed before passing it to the torch function), and with that padding the only allowed dimensions are 3 and 4.

> So why am I not skipping 2D?

That because in `torchvision.transforms.functional_tensor.pad`, they are unsqueezed:

```
if img.ndim < 4:
    img = img.unsqueeze(dim=0)
```

vfdev-5: > mask.ndim not in [2, 3, 4]:

I think `mask` can't have ndim == 2 ? Minimally, it is `(1, H, W)`. cc @pmeier

For N-D cases `e.g. (N, M, K, L, 1, H, W)` we could reshape it into 4D before padding `(N * M * K * L, 1, H, W)` and reshape back inside padding op ?

test/test_prototype_transforms_functional.py
```diff
@@ -1031,3 +1047,46 @@ def _compute_expected(mask, top_, left_, height_, width_, size_):
...
+        expected_mask = torch.zeros(new_shape, dtype=torch.long)
+        expected_mask[..., pad_up:-pad_down, pad_left:-pad_right] = mask
+
+        return expected_mask
+
+    for mask in make_segmentation_masks():
+        out_mask = F.pad_segmentation_mask(mask, padding, "constant")
+
+        expected_mask = _compute_expected_mask()
+        torch.testing.assert_close(out_mask, expected_mask)
```
vfdev-5, line 1092: Thinking of how to test all other padding modes. Maybe we could just check output value from `out_mask` for two lines: one horizontal and one vertical instead of constructing full expected mask.
While checking the lines we still need to route the checks according to the padding mode.
What do you think ?

test/test_prototype_transforms_functional.py
federicopozzi33: Do you mean something like this?

(The code can be further refactored).
```python
def pad_segmentation_mask(
    segmentation_mask: torch.Tensor, padding: List[int], padding_mode: str = "constant"
) -> torch.Tensor:
  *extra_dims, c, h, w = segmentation_mask.shape 
      
    mask_4d = segmentation_mask.reshape(-1, b, h, w) if len(extra_dims) > 0 else segmentation_mask
    padded = pad_image_tensor(img=mask_4d, padding=padding, fill=0, padding_mode=padding_mode)
      
  *_, new_c, new_h, new_w = padded.shape
  return padded.reshape(*extra_dims, new_c, new_h, new_w) if len(extra_dims) > 0 else padded
```

Seems that `.shape` operation is not supported by `torch.jit`...

```python
cannot statically infer the expected size of a list in this context:
  File "/Users/fpozzi/Documents/OpenSource/my_repos/vision/torchvision/prototype/transforms/functional/_geometry.py", line 402
    segmentation_mask: torch.Tensor, padding: List[int], padding_mode: str = "constant"
) -> torch.Tensor:
    *extra_dims, b, h, w = segmentation_mask.shape 
                           ~~~~~~~~~~~~~~~~~~~~~~ <--- HERE
    
    mask_4d = segmentation_mask.reshape(-1, b, h, w) if len(extra_dims) > 0 else segmentation_mask
```

I tried also with .`size()` but I got the same expection.

vfdev-5: `.shape` is supported by torch script but `*extra_dims` is not.
You can do something like:
```python
mask = torch.rand(5, 4, 1, 32, 32)

def foo(mask):
    c, h, w = mask.shape[-3], mask.shape[-2], mask.shape[-1]
    m = mask.view(-1, c, h, w)
    return m.view(mask.shape)

torch.jit.script(foo)(mask).shape
```

pmeier: > I think `mask` can't have ndim == 2 ? Minimally, it is `(1, H, W)`.

Masks have gotten very little love so far so there is no final decision. But I agree, boolean masks should have shape `(N, H, W)` where `N` is the number of segmentation whether they are individual objects (instance segmentation) or a crowd (semantic segmentation). 

> Do you mean something like this?

Have a look how we are doing it for the bounding boxes:

https://github.com/pytorch/vision/blob/a8f563dbf8520020054aa01f5ae169999775fd19/torchvision/prototype/transforms/functional/_geometry.py#L22-L35

The only difference is that the "data dimensions" are not fixed. For masks this would look like

```py
@torch.jit.script
def pad_segmentation_mask(segmentation_mask: torch.Tensor) -> torch.Tensor:
    shape = segmentation_mask.shape

    num_masks, height, width = shape[-3:]
    segmentation_mask = segmentation_mask.view(-1, num_masks, height, width)
    
    # kernel
    
    return segmentation_mask.view(shape)
```

federicopozzi33: > > Do you mean something like this?
> 
> Have a look how we are doing it for the bounding boxes:
> 
> https://github.com/pytorch/vision/blob/a8f563dbf8520020054aa01f5ae169999775fd19/torchvision/prototype/transforms/functional/_geometry.py#L22-L35
> 
> The only difference is that the "data dimensions" are not fixed. For masks this would look like
> 
> ```python
> @torch.jit.script
> def pad_segmentation_mask(segmentation_mask: torch.Tensor) -> torch.Tensor:
>     shape = segmentation_mask.shape
> 
>     num_masks, height, width = shape[-3:]
>     segmentation_mask = segmentation_mask.view(-1, num_masks, height, width)
>     
>     # kernel
>     
>     return segmentation_mask.view(shape)
> ```

Yeah, the point is that the mask shape before padding is different from mask shape after. 
So I can't simply do:

```python
return segmentation_mask.view(shape)
```

Mask can have any num_dims > 2. So I need to collect dimensions before padding: in particular, I need the extra dimensions, num_masks, height and width.

```python
# Let's suppose that segmentation_mask have shape (3, 4, 1, 5, 5).
extra_dims, num_masks, h, w = segmentation_mask.shape 
print(extra_dims, num_masks, h, w)
> (3, 4), 1, 5, 5
```

Moreover, I haven't figured out yet how to use the `.view` method without the unpacking operator:
```python
# with unpacking operator
padded_mask.view(*extra_dims, num_masks, height, width)
```

test/test_prototype_transforms_functional.py
federicopozzi33: Seems pretty reasonable.

Anyway, the code needed for constructing the two lines IMO still "mimic" (or re-implement) the padding operation.

test/test_prototype_transforms_functional.py
pmeier: > Yeah, the point is that the mask shape before padding is different from mask shape after padding.

:facepalm: My bad, sorry. [`resize_image_tensor`](https://github.com/pytorch/vision/blob/e0467c64e337c0d1140a9f9a70a413b7268231f4/torchvision/prototype/transforms/functional/_geometry.py#L38-L54) is a better reference:

```py
@torch.jit.script
def pad_segmentation_mask(segmentation_mask: torch.Tensor) -> torch.Tensor:
    num_masks, old_height, old_width = segmentation_mask.shape[-3:]
    batch_shape = segmentation_mask.shape[:-3]
    segmentation_mask = segmentation_mask.view((-1, num_masks, old_height, old_width))

    # kernel
    new_height, new_width = old_height, old_width

    return segmentation_mask.view(batch_shape + (num_masks, new_height, new_width))
```

test/test_prototype_transforms_functional.py
pmeier: > Thinking of how to test all other padding modes. 

I don't think we need a correctness check for them. Internally, [`torch.nn.functional.pad`](https://pytorch.org/docs/master/generated/torch.nn.functional.pad.html) does the heavy lifting. Thus, if we rely on that giving us the correct behavior there is no need to check if the values of the padding are correct.

test/test_prototype_transforms_functional.py
federicopozzi33: > > Yeah, the point is that the mask shape before padding is different from mask shape after padding.
> 
> 🤦 My bad, sorry. [`resize_image_tensor`](https://github.com/pytorch/vision/blob/e0467c64e337c0d1140a9f9a70a413b7268231f4/torchvision/prototype/transforms/functional/_geometry.py#L38-L54) is a better reference:
> 
> ```python
> @torch.jit.script
> def pad_segmentation_mask(segmentation_mask: torch.Tensor) -> torch.Tensor:
>     num_masks, old_height, old_width = segmentation_mask.shape[-3:]
>     batch_shape = segmentation_mask.shape[:-3]
>     segmentation_mask = segmentation_mask.view((-1, num_masks, old_height, old_width))
> 
>     # kernel
>     new_height, new_width = old_height, old_width
> 
>     return segmentation_mask.view(batch_shape + (num_masks, new_height, new_width))
> ```

Thank you!

I was missing some things:
First one:

```
size = torch.ones(3, 3).shape[:]
> torch.Size([3, 3])

size_as_tuple = np.ones((3, 3)).shape[:]
> (3, 3)
```

Second one:

```
torch.Size([1, 2]) + (3, 4)
> torch.Size([1, 2, 3, 4])
```

pmeier (commented):
I don't see anything the I would object too. I'll leave it to @vfdev-5 to approve, since he shepherded the PR so far.

test/test_prototype_transforms_functional.py
vfdev-5: @pmeier if we do not check correctness on the output, another option is to mock `torch_pad` and ensure that it is called with correct configuration otherwise the code is not covered.

vfdev-5:
@federicopozzi33 can you please update the discription with few visualizations of how padding works on images, bbox and segm masks. You can inspire from https://github.com/pytorch/vision/pull/5613 or other similar PRs. Thanks !

federicopozzi33:
> @federicopozzi33 can you please update the discription with few visualizations of how padding works on images, bbox and segm masks. You can inspire from #5613 or other similar PRs. Thanks !

Done.

Issue #2783, opened 2020-10-10
warning: ignoring #pragma omp parallel


xkszltl:
## 🐛 Bug

I'm seeing a lot of `warning: ignoring #pragma omp parallel` like this:
<img width="1085" alt="image" src="https://user-images.githubusercontent.com/5203025/95647645-b5553600-0b03-11eb-94e1-4ad6caf7215d.png">

Maybe there should be a `-fopenmp`?

## Environment

 - PyTorch / torchvision Version (e.g., 1.0 / 0.4.0): master/master
 - OS (e.g., Linux): Ubuntu 18.04
 - How you installed PyTorch / torchvision (`conda`, `pip`, source): source
 - Build command you used (if compiling from source): cmake+ninja+gcc-8
 - Python version: 3.6
 - CUDA/cuDNN version: 11.0/8

fmassa:
Thanks for the report. We don't directly use omp parallel in nms_cpu, so this might be somewhere else.

We will have a look.

xkszltl:
Any news so far?

vfdev-5:
@xkszltl i'll try to take a look at this issue this week.

klvnptr:
@vfdev-5 any update? :) I'm having the same problem

vfdev-5:
@klvnptr Sorry, seems like I missed this issue in my task list. Let me check this on monday. Thanks.

klvnptr:
thanks for getting back @vfdev-5 
for me this solved the problem:
$ apt-get install libopenmpi-dev
$ CFLAGS="-fopenmp" python3 setup.py bdist_wheel

xkszltl:
@vfdev-5 
I don't think that's the right way.
- It should be done in CMake, not `setup.py`, or the issue is still there when calling cmake directly.
- When adding openmp, you should use `find_package(OpenMP)` and link to it.
- Blindly adding `-fopenmp` may cause inconsistent result with pytorch's openmp. There're many openmp implementations, not just the default one.
- This could be an issue in pytorch where it doesn't expose openmp flag in dependency properly. If so issue should be forwarded to them as well.

xkszltl:
Bring up issue in upstream as well: https://github.com/pytorch/pytorch/issues/48033

vfdev-5:
@xkszltl the warning `warning: ignoring #pragma omp parallel` is seen when we build torchvision with `python setup.py build` and more precisely when C++ extensions are built: 
[...]

PR #5870, opened 2022-04-24
Enable openmp parallelization for nms kernel

leslie-fang-intel:
In this PR:
- Add the OpenMP support to NMS CPU Kernel to enable the parallel calculation. 

- The MaskRCNN NMS kernel(https://github.com/facebookresearch/maskrcnn-benchmark/blob/main/maskrcnn_benchmark/csrc/cpu/nms_cpu.cpp#L5-L65) has offset of 1 when doing the calculation. Add the offset parameter(by default zero) to torch vision NMS kernel, then the MaskRCNN NMS kernel can share the implementation of Torchvision NMS Kernel.

### DIFF
--- a/torchvision/csrc/ops/autocast/nms_kernel.cpp
+++ b/torchvision/csrc/ops/autocast/nms_kernel.cpp
@@ -12,12 +12,14 @@ namespace {
 at::Tensor nms_autocast(
     const at::Tensor& dets,
     const at::Tensor& scores,
-    double iou_threshold) {
+    double iou_threshold,
+    double bias) {
   c10::impl::ExcludeDispatchKeyGuard no_autocast(c10::DispatchKey::Autocast);
   return nms(
       at::autocast::cached_cast(at::kFloat, dets),
       at::autocast::cached_cast(at::kFloat, scores),
-      iou_threshold);
+      iou_threshold,
+      bias);
 }
 
 } // namespace
--- a/torchvision/csrc/ops/cpu/nms_kernel.cpp
+++ b/torchvision/csrc/ops/cpu/nms_kernel.cpp
@@ -32,10 +32,8 @@ at::Tensor nms_kernel_impl(
 
   auto ndets = dets.size(0);
   at::Tensor suppressed_t = at::zeros({ndets}, dets.options().dtype(at::kByte));
-  at::Tensor keep_t = at::zeros({ndets}, dets.options().dtype(at::kLong));
 
   auto suppressed = suppressed_t.data_ptr<uint8_t>();
-  auto keep = keep_t.data_ptr<int64_t>();
   auto order = order_t.data_ptr<int64_t>();
   auto x1 = x1_t.data_ptr<scalar_t>();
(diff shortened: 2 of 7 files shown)


datumbox (changes requested):
@leslie-fang-intel Thanks a lot for your contribution.

I think it's worth splitting this PR to separate so that we don't mix the openmp parallelization with the introduction of bias. The parallelization work looks OK to me. I'm not sure that the change on the bias should be added though. My understanding is that it was not included intentionally as part of a simplification process. Also the introduction of the `bias` parameter on CPU only makes for a problematic API.

@fmassa do you have any thoughts on the above?

vfdev-5:
I do not if this issue is related https://github.com/pytorch/vision/issues/2783 but previously we had a problem to make openmp available for torchvision...

leslie-fang-intel:
Thanks for the suggestions. I will collect the benchmark number w and wo the parallelization.

leslie-fang-intel:
Hi @datumbox @fmassa Here is the test case which I plan to collect performance with.
I suppose we will do `nms` for each batch and each class. So I will change the parameter of `batchsize`, `num_boxes`, `class_number` for each set of test and get the performance number with each `OMP_NUM_THREADS`.

Does this test case looks reasonable to you?
```
import torch
import time

if __name__ == "__main__":
    batchsize=16
    num_boxes = 15130
    class_number = 81 # For coco dataset
    iou_threshold = 0.9

    boxes = torch.cat((torch.rand(num_boxes, 2), torch.rand(num_boxes, 2) + 10), dim=1) # 15130 * 4
    assert max(boxes[:, 0]) < min(boxes[:, 2])  # x1 < x2
    assert max(boxes[:, 1]) < min(boxes[:, 3])  # y1 < y2

    scores = torch.rand(num_boxes, class_number) # 15130 * 81

    steps = 10
    start = time.time()
    for _ in range(10):
        for batch_id in range(batchsize):
            for class_id in range(class_number):
                ops.boxes.nms(boxes[:, :], scores[:, class_id], iou_threshold)
    print("Toal time: {}".format(time.time() - start))

```

datumbox:
@leslie-fang-intel Here is a [test](https://github.com/pytorch/vision/blob/11bd2eaa6d6976129836b329b01d1300babddcc9/test/test_onnx.py#L93) that does something similar. I would also fix the seeds for all tests to avoid differences related to the data. For example what you could do is generate different data per iteration by setting the seed equal to iteration number so that you get different results per iteration but same across experiments. Finally I would recommend doing a few warm up iters prior start measuring to ensure you get stable and accurate metrics.

fmassa:
Hi @leslie-fang-intel 

I'd run the benchmark for a number of different boxes (not only 15130). Indeed, for a vanilla Faster R-CNN with FPN, the NMS kernel will see at most 2000 boxes, so it would be good to have a range like `[300, 1000, 2000, 5000, 10000]` for example

leslie-fang-intel:
Will create new PR with parallelization optimization only as first step.

PR #5874, opened 2022-04-25
Fix weight doc generation

NicolasHug:
Fixes https://github.com/pytorch/vision/pull/5859#pullrequestreview-950197056

Rendered docs: https://output.circle-artifacts.com/output/job/910114e1-8910-4ee4-a997-f50b5e4cbb5b/artifacts/0/docs/models/generated/torchvision.models.regnet_y_400mf.html#torchvision.models.regnet_y_400mf

The doc looks like it used to: individual metrics appear just like any other field in the `meta` dict: 

image

We could also have a separate table for metrics, for each weight. No strong opinion.


datumbox (approved):
LGTM, thanks for fixing this @NicolasHug

Issue #5873, opened 2022-04-25
CI jobs broken due to upstream changes on padding


datumbox:
### 🐛 Describe the bug

The latest main seems broken potentially due to upstream changes on PyTorch's core. I believe the problem was introduced due to the changes on pad at https://github.com/pytorch/pytorch/pull/73433.

Here are different [failures](https://app.circleci.com/pipelines/github/pytorch/vision/16939/workflows/22a1367b-29f9-494c-880a-ee2c9c3c2fd4/jobs/1373186) we get:
```
test_pad[1-edge-cpu]:
Traceback (most recent call last):
  File "/root/project/test/test_transforms_tensor.py", line 205, in test_pad
    _test_functional_op(F.pad, fn_kwargs={"padding": mul * 2, "fill": fill, "padding_mode": m}, device=device)
  File "/root/project/test/test_transforms_tensor.py", line 54, in _test_functional_op
    transformed_tensor = f(tensor, **fn_kwargs)
  File "/root/project/torchvision/transforms/functional.py", line 480, in pad
    return F_t.pad(img, padding=padding, fill=fill, padding_mode=padding_mode)
  File "/root/project/torchvision/transforms/functional_tensor.py", line 415, in pad
    img = torch_pad(img, p, mode=padding_mode, value=float(fill))
RuntimeError: Padding mode "replicate" doesn't take in value argument
```

```
test_pad[1-reflect-cpu]
Traceback (most recent call last):
  File "/root/project/test/test_transforms_tensor.py", line 205, in test_pad
    _test_functional_op(F.pad, fn_kwargs={"padding": mul * 2, "fill": fill, "padding_mode": m}, device=device)
  File "/root/project/test/test_transforms_tensor.py", line 54, in _test_functional_op
    transformed_tensor = f(tensor, **fn_kwargs)
  File "/root/project/torchvision/transforms/functional.py", line 480, in pad
    return F_t.pad(img, padding=padding, fill=fill, padding_mode=padding_mode)
  File "/root/project/torchvision/transforms/functional_tensor.py", line 415, in pad
    img = torch_pad(img, p, mode=padding_mode, value=float(fill))
RuntimeError: Padding mode "reflect" doesn't take in value argument
```

```
test_color_jitter_all[3-cpu-9]:
Traceback (most recent call last):
  File "/root/project/test/test_transforms_tensor.py", line 194, in test_color_jitter_all
    channels=channels,
[...]

PR #5875, opened 2022-04-25
Fixed issue with padding on CI

vfdev-5:
Related to #5873

- Fixing issue with padding
- Disabled failing tests with color-jitter


test/test_transforms_tensor.py
```diff
@@ -119,8 +118,12 @@ def test_random(func, method, device, channels, fn_kwargs, match_kwargs):
 
 
 @pytest.mark.parametrize("seed", range(10))
-@pytest.mark.parametrize("device", cpu_and_gpu())
-@pytest.mark.parametrize("channels", [1, 3])
+# Temporary disable CUDA device
+# https://github.com/pytorch/vision/issues/5873
```
datumbox, line 122: I'm not sure temporarily disabling the CUDA execution is the way to go here. If there is no other way we can fix it on our side, we should consider reverting https://github.com/pytorch/pytorch/pull/73433 which broke TorchVision.

cc @albanD @jbschlosser

vfdev-5: Just few more info on failures. Failures with ColorJitter are flaky are failing on batched tensor tests: https://github.com/pytorch/vision/blob/6d85d74be6ac72b2ac3057d85d8ae0004fa0de3f/test/test_transforms_tensor.py#L86
https://github.com/pytorch/vision/blob/6d85d74be6ac72b2ac3057d85d8ae0004fa0de3f/test/test_transforms_tensor.py#L46

Failures are flaky as tests are failing when run all of them and pass if run one by one...

torchvision/transforms/functional_tensor.py
```diff
@@ -412,7 +412,10 @@ def pad(img: Tensor, padding: List[int], fill: int = 0, padding_mode: str = "con
         need_cast = True
         img = img.to(torch.float32)
 
-    img = torch_pad(img, p, mode=padding_mode, value=float(fill))
+    if padding_mode in ("reflect", "replicate"):
+        img = torch_pad(img, p, mode=padding_mode)
+    else:
+        img = torch_pad(img, p, mode=padding_mode, value=float(fill))
```
datumbox, line 418: There is a PR that resolves the accidental BC break at https://github.com/pytorch/pytorch/pull/76307

It's not merged yet but we can go ahead and adopt this approach in TorchVision to avoid it regardless.

datumbox (commented):
@vfdev-5 I have undone your skipping of the ColourJitter tests. I believe the issue was fixed. by reverting the upstream PR in question.

I think this should fix the issue, let's wait for the CI results and if green we can merge.

datumbox:
I'm merging this to fix the breakages from the previous nightly. Thanks for the contribution @vfdev-5

datumbox (approved):

PR #5876, opened 2022-04-25
Start doc revamp for detection models

NicolasHug:
This PR adds table generation for the detection weights, and adds doc for RetinaNet.

Rendered docs: https://output.circle-artifacts.com/output/job/a31ba961-8635-4fbf-8dc0-af4039a5c788/artifacts/0/docs/models_new.html#object-detection-instance-segmentation-and-person-keypoint-detection


docs/source/conf.py
```diff
@@ -330,7 +331,7 @@ def inject_weight_metadata(app, what, name, obj, options, lines):
             # the `meta` dict contains another embedded `metrics` dict. To
             # simplify the table generation below, we create the
             # `meta_with_metrics` dict, where the metrics dict has been "flattened"
-            meta = field.meta
+            meta = copy(field.meta)
```
NicolasHug, line 334: This is a minor bugfix: we avoid modifying the meta inplace.

docs/source/conf.py
```diff
@@ -366,13 +367,14 @@ def generate_classification_table():
...
+    with open(generated_dir / f"{table_name}_table.rst", "w+") as table_file:
         table_file.write(".. table::\n")
-        table_file.write("    :widths: 100 10 10 20 10\n\n")
+        table_file.write(f"    :widths: 100 {'15 ' * len(metrics_names)} 20 10\n\n")
         table_file.write(f"{textwrap.indent(table, ' ' * 4)}\n\n")
 
 
-generate_classification_table()
+generate_weights_table(module=M, table_name="classification", metrics_names=["Acc@1", "Acc@5"])
+generate_weights_table(module=M.detection, table_name="detection", metrics_names=["box_map"])
```
datumbox, line 382: I think this is going to put the KeypointRCNN model along with the rest of the models. Unfortunately the specific model was trained with different data from COCO (keypoints), so its focus and validation numbers are the same as for other models. 

Have a look on the current documentation [here](https://github.com/pytorch/vision/blob/main/docs/source/models.rst#object-detection-instance-segmentation-and-person-keypoint-detection) to see how the models are grouped and how extra accuracy metrics are being shown. We don't have to do exactly the same thing but good to get some ideas.

datumbox (commented):
Thanks @NicolasHug, I've added a few comments below to get your thoughts. Happy to chat more.

docs/source/conf.py
NicolasHug: Is it OK with you if I create an issue to special-case KeypointRCNN? This way I'll still be able to ask for external contributors while I'm thinking of a decent solution for this.

docs/source/conf.py
```diff
@@ -346,17 +347,17 @@ def inject_weight_metadata(app, what, name, obj, options, lines):
...
 
-    column_names = ("**Weight**", "**Acc@1**", "**Acc@5**", "**Params**", "**Recipe**")
+    column_names = ["weight"] + list(metrics_names) + ["params", "recipe"]
+    column_names = [f"**{name}**" for name in column_names]
     content = [
         (
             f":class:`{w} <{type(w).__name__}>`",
-            w.meta["metrics"]["acc@1"],
-            w.meta["metrics"]["acc@5"],
+            *tuple(w.meta["metrics"][metric.lower()] for metric in metrics_names),
```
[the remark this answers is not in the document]
NicolasHug: Yes, one problem at a time :)

docs/source/models/retinanet.rst
```diff
@@ -0,0 +1,23 @@
+RetinaNet
+=========
+
+.. currentmodule:: torchvision.models.detection
+
+The RetinaNet model is based on the `Focal Loss for Dense Object Detection
+<https://arxiv.org/abs/1708.02002>`__ paper.
```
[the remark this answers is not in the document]
datumbox: Hmm... Yeah that's OK with me.

docs/source/conf.py
datumbox: Yeah I think that would be fine.

docs/source/conf.py
```diff
@@ -346,17 +347,22 @@ def inject_weight_metadata(app, what, name, obj, options, lines):
...
 
-def generate_classification_table():
+def generate_weights_table(module, table_name, metrics_names):
 
-    weight_enums = [getattr(M, name) for name in dir(M) if name.endswith("_Weights")]
+    weight_enums = [getattr(module, name) for name in dir(module) if name.endswith("_Weights")]
     weights = [w for weight_enum in weight_enums for w in weight_enum]
 
-    column_names = ("**Weight**", "**Acc@1**", "**Acc@5**", "**Params**", "**Recipe**")
+    def clean_name(name):
```
datumbox, line 355: Nit: Might be cleaner if you pass `metrics_names` as `Tuple[str, str]` where one part of the tuple is the key and the second the doc-friendly name of it.

docs/source/conf.py
datumbox: Yeah I think that would be fine, let's do this.

datumbox (approved):
Approving since on principle it looks good. Just one non-blocking nit below:

PR #5877, opened 2022-04-25
use generalised_box_iou function to calculate giou_loss

abhi-glitchhg:
We can use the `generalized_box_iou` function to calculate the `generalized_box_iou_loss`, here is my attempt for the same. 

cc: @datumbox @pmeier

### DIFF
--- a/torchvision/ops/giou_loss.py
+++ b/torchvision/ops/giou_loss.py
@@ -2,6 +2,7 @@ import torch
 from torch import Tensor
 
 from ..utils import _log_api_usage_once
+from .boxes import generalized_box_iou
 
 
 def _upcast(t: Tensor) -> Tensor:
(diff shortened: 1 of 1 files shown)


torchvision/ops/giou_loss.py
```diff
@@ -48,30 +49,14 @@ def generalized_box_iou_loss(
...
-    area_c = (xc2 - xc1) * (yc2 - yc1)
-    miouk = iouk - ((area_c - unionk) / (area_c + eps))
 
+    # here we can use diagonal matrix with no problem because the resultant tensor of `generalized_box_iou` will be having shape = N X N
+    if boxes1.shape == torch.Size([4]):
+        boxes1 = boxes1[None, :]
+        boxes2 = boxes2[None, :]
+        miouk = generalized_box_iou(boxes1, boxes2)[0][0]
+    else:
+        miouk = generalized_box_iou(boxes1, boxes2).diagonal()
```
datumbox, line 59: This is going to be quite slow as noted at https://github.com/pytorch/vision/pull/5786#discussion_r858587170

Let's continue our discussion on the other thread just to avoid splitting the comments on various places.

datumbox (changes requested):
@abhi-glitchhg thanks for the contribution! :)

Marking as "Request Changes" to avoid accidental merges. Let's chat about this on the other thread linked below to decide a way forwards.

PR #5879, opened 2022-04-26
refactor prototype transforms functional tests

pmeier:
This PR refactors our prototype transforms functional tests. They are currently located in `test/test_prototype_transforms_functional.py`. To ease the reviewing process I've added a new `test/test_prototype_transforms_kernels.py` module that contains the refactored tests from this PR. In the end that should replace most parts of the old file, but doing it in a separate module avoids GH diff hell.

## Status quo

Our current implementation was the first attempt to automate our tests. I took some inspiration from the `OpInfo` framework from PyTorch core. The basic idea is to define a `FunctionalInfo`'s for each functional that stores some metadata about it. 

The most important info (and for now the only metadata we store) is the `sample_inputs_fn`. It yields call arguments

With that we can write common tests that can be `@pytest.mark.parametrize`'d over the kernel-call-args combinations. For example, a test that checks the `torch.jit.script`'ed output against its eager counterpart looks like this

https://github.com/pytorch/vision/blob/f36f3514e78485373ff04bc7efaef5aef9f4379a/test/test_prototype_transforms_functional.py#L583-L595

## Pros / Cons

This architecture has two main benefits over manually writing these tests:

1. It is trivial to add a new common test for all functionals. 
2. Writing comprehensive tests for a new functional is reduced to defining call args.

Plus, and in contrast to the `OpInfo`'s from PyTorch core, the test are easier to debug, since we use `@pytest.mark.parametrize` over `for` loops in the test body to iterate over the call args. If one of our tests fails, we can reproduce the parametrization from the log, whereas in PyTorch core one needs manually find which call args are responsible for the failure.

However, there are also downsides:

1. Using `@pytest.mark.parametrize` instantiates everything upfront. Especially with tensor inputs that can quickly become a big chunk of memory. Right now `test/test_prototype_transforms_functional.py` includes ~23k tests that instantiate tensors during collection (`pytest --co test/test_prototype_transforms_functional.py::test_eager_vs_scripted`). They all come from a common test, namely eager vs scripted. If we add more tests or more ops, this number will grow fast. This is the reason why PyTorch core does *not* rely on parametrization for their sample inputs, but rather fall back to `for` loops inside the test.
2. Having only one kind of sample inputs. Although each test might have different needs with regard to the sample inputs, we only have one generating function. This means that we need to support all needs with it and in turn unnecessary blow up the time to run all tests. For example, to test whether the output of the scripted `affine_image_tensor` kernel matches its eager counterpart, we only need to test a single set of affine parameters (as long as there is no branching based on them). However, for reference testing, we should test multiple parameter sets to make sure the kernel actually behaves like its reference.

## Design goal

This PR sets out to solve the problems detailed above while retaining all the positive aspects of the current implementation.

1. Introduce the `TensorLoader` class: it wraps another callable that in the end will create the tensor, but it knows the `shape`, `dtype` and possible other feature metadata ahead of time. The `device` will only be passed at runtime to allow us to parametrize over different devices. With this we can continue to rely on the tensor attributes during sample input generation, e.g.

    https://github.com/pytorch/vision/blob/b5c961d400b9fb16892fc57a52974aa676773a93/test/test_prototype_transforms_functional.py#L248
    
    https://github.com/pytorch/vision/blob/b5c961d400b9fb16892fc57a52974aa676773a93/test/test_prototype_transforms_functional.py#L264

    without actually instantiating the tensors.

    At test time, the tensor can simply instantiated with `TensorLoader(...).load(device)`. For convenience, `ArgsKwargs` was made aware of `TensorLoader` and got a `.load(device)` method as well. With these, the common tests will look somewhat like this:

    ```py
    @pytest.mark.parametrize(
        ("info", "args_kwargs"), [
            (info, args_kwargs) 
            for info in KERNEL_INFOS 
            for args_kwargs in info.sample_inputs
        ]
    )
    @pytest.mark.parametrize("device", ["cpu", "cuda"])
    def test_smoke(info, args_kwargs, device):
        args, kwargs = args_kwargs.load(device)
    
        result = info.kernel(*args, **kwargs)
    ```

    This approach of "lazy loading" is similar to the concept of [lazy tensors](https://pytorch.org/blog/understanding-lazytensor-system-performance-with-pytorch-xla-on-cloud-tpu/) although stripping everything we don't need. To avoid confusion, I preferred to use the term "load" over "lazy" here.
2. Introduce a `reference_inputs_fn` alongside the `sample_inputs_fn`. As the name implies, the former will only be used for reference tests and should be comprehensive with respect to the tested values. In contrast, the `sample_inputs_fn` should only cover all valid code paths. This is on par with PyTorch core does with their `OpInfo` framework although they have even more diverse sample inputs functions, like the [`error_inputs_func`](https://github.com/pytorch/pytorch/blob/5ea2eb304ea6e9bab0c68fc57dbffbc068354ce7/torch/testing/_internal/opinfo/core.py#L641).

## Limitations

There are two things that are not included in the current design:

1. Reference test against fixed inputs. The reference tests explained above work by using one function that generates sample inputs and passing them to the kernel as well as a reference function. Thus, we only need to parametrize over one set of sample inputs and the reference outputs only get computed at runtime. To support fixed references, we would need a map of sample inputs to their fixed reference and thus defeating the whole part of the design to nto instantiate tensors at collection. Fortunately, we don't need to force everything into the framework proposed here and can simply have separate "manual" tests for this. Of course this also applies to testing error / warning inputs although in the future we could also integrate that into the framework.
2. Tests for high level functionals aka dispatchers. As the name implies, they only dispatch, so it would be quite a waste of resources to test them again with sample inputs if we already tested the kernels. I think we should implement something like "if I put in a PIL Image, the PIL kernel gets called and its output gets returned". However, this is out of scope for this PR.

## Todo

This PR mostly introduces the new framework while adding some kernels as examples. There are three ways to add the remaining ones:

1. Finish everything in this PR.
2. Merge this PR and add follow-up PRs for the remaining kernels. This is possible without ripping large gaps in our CI since I've added the new tests in a new module. The last PR of this series could be a clean up to remove all the then duplicated tests in the old module.
3. Use this PR as feature branch and add more PRs against this until we are finished.

My preference is 3. -> 2. -> 1. but I'll leave that up to the reviewers. Here is the list of what kernels are done or missing:

- `clamp_bounding_box`
- `convert_bounding_box_format`
- `convert_color_space`
  - [ ] `image`
- `adjust_brightness`
  - [ ] `image`
- `adjust_contrast`
  - [ ] `image`
- `adjust_gamma`
  - [ ] `image`
- `adjust_hue`
  - [ ] `image`
- `adjust_saturation`
  - [ ] `image`
- `adjust_sharpness`
  - [ ] `image`
- `autocontrast`
  - [ ] `image`
- `equalize`
  - [ ] `image`
- `invert`
  - [ ] `image`
- `posterize`
  - [ ] `image`
- `solarize`
  - [ ] `image`
- `affine`
  - [x] `bounding_box`
  - [x] `image`
  - [ ] `segmentation_mask`
- `center_crop`
  - [ ] `bounding_box`
  - [ ] `image`
  - [ ] `segmentation_mask`
- `crop`
  - [ ] `bounding_box`
  - [ ] `image`
  - [ ] `segmentation_mask`
- `elastic`
  - [ ] `bounding_box`
  - [ ] `image`
  - [ ] `segmentation_mask`
- `five_crop`
  - [ ] `image`
- `horizontal_flip`
  - [x] `bounding_box`
  - [x] `image`
  - [x] `segmentation_mask`
- `pad`
  - [ ] `bounding_box`
  - [ ] `image`
  - [ ] `segmentation_mask`
- `perspective`
  - [ ] `bounding_box`
  - [ ] `image`
  - [ ] `segmentation_mask`
- `resize`
  - [x] `bounding_box`
  - [x] `image`
  - [ ] `segmentation_mask`
- `resized_crop`
  - [ ] `bounding_box`
  - [ ] `image`
  - [ ] `segmentation_mask`
- `rotate`
  - [ ] `bounding_box`
  - [ ] `image`
  - [ ] `segmentation_mask`
- `ten_crop`
  - [ ] `bounding_box`
  - [ ] `image`
  - [ ] `segmentation_mask`
- `vertical_flip`
  - [ ] `bounding_box`
  - [ ] `image`
  - [ ] `segmentation_mask`


test/test_prototype_transforms_functional.py
```diff
@@ -128,82 +145,184 @@ def make_one_hot_label(*args, **kwargs):
...
-    return sample_inputs_fn
+def sample_inputs_horizontal_flip_image_tensor(device):
+    for image in make_images(sizes=[DEFAULT_IMAGE_SIZE], device=device, dtypes=[torch.float32]):
+        yield SampleInput(image)
 
 
-@register_kernel_info_from_sample_inputs_fn
-def horizontal_flip_image_tensor():
-    for image in make_images():
+def reference_inputs_horizontal_flip_image_tensor():
```
vfdev-5, line 273: I was expecting to see here as input args: input tensor and params (for other ops). Seems like to reference output is generated no matter the input...
I maybe missing something here.

If we need to generate sample inputs, can't we refactor this part between `def sample_inputs_horizontal_flip_image_tensor` and `reference_inputs_horizontal_flip_image_tensor` and while configuring `FunctionalInfo`, developer just specify reference function working on a single input ?

pmeier: > I was expecting to see here as input args: input tensor and params (for other ops). Seems like to reference output is generated no matter the input...
I maybe missing something here.

The `reference_inputs_fn` is not the same as the reference. The former just creates the inputs that are used for testing against the reference. It is different from the `sample_inputs_fn` in the sense that it can test a more comprehensive set of inputs with regard to their values. `sample_inputs_fn` basically only checks all valid code paths.

I'm marking this as resolved since there was quite a few refactoring steps since this comment. If this is not clear in the current state, feel free to ask again.

test/prototype_common_utils.py
```diff
@@ -1,28 +1,127 @@
...
+)
 from torchvision.prototype import features
-from torchvision.prototype.transforms.functional import to_image_tensor
+from torchvision.prototype.transforms.functional import convert_image_dtype, to_image_tensor
 from torchvision.transforms.functional_tensor import _max_value as get_max_value
 
+__all__ = ["assert_close"]
+
+
+class PILImagePair(TensorLikePair):
```
pmeier, line 28: This is a superset of what the old `ImagePair` did. It includes the options to only test the aggregated difference or check the percentage of differing pixels. That is on par with what we are currently doing in our stable functional tests:

https://github.com/pytorch/vision/blob/a67cc87a33a3f713aebf5299bdeb2672c98e0bc5/test/common_utils.py#L172-L174

test/prototype_common_utils.py
```diff
@@ -34,26 +133,104 @@ def __iter__(self):
...
+    def load(self, device="cpu"):
+        args = tuple(arg.load(device) if isinstance(arg, TensorLoader) else arg for arg in self.args)
+        kwargs = {
+            keyword: arg.load(device) if isinstance(arg, TensorLoader) else arg for keyword, arg in self.kwargs.items()
+        }
+        return args, kwargs
 
-            return f"{repr_[:max//2]}...{repr_[-(max//2-3):]}"
+    def __repr__(self):
+        def better_repr(obj):
```
pmeier, line 144: We no longer need to shorten the `repr` here, since that was mostly done for tensors and they are now replaced with `TensorLoader`'s.

test/prototype_common_utils.py
```diff
@@ -34,26 +133,104 @@ def __iter__(self):
...
+    return wrapper
+
+
+def from_loaders(loaders_fn):
+    def wrapper(*args, **kwargs):
+        loaders = loaders_fn(*args, **kwargs)
+        for loader in loaders:
+            yield loader.load(kwargs.get("device", "cpu"))
+
+    return wrapper
```
pmeier, line 180: These functions are mostly for "BC" with our current tests. With them we can turn `make_*_loader{s}` back into `make_*`. For example, `make_images = from_loaders(make_image_loaders)`. This makes the transition period easer, since we don't need to touch the old files. 

In the future, most tests should use the loader architecture. Those that don't could simply invoke `TensorLoader(...).load(device)` manually.

test/prototype_common_utils.py
```diff
@@ -34,26 +133,104 @@ def __iter__(self):
...
+class TensorLoader:
+    def __init__(self, fn, *, shape, dtype):
+        self.fn = fn
+        self.shape = shape
+        self.dtype = dtype
+
+    def unwrap(self):
+        return TensorLoader(
+            lambda shape, dtype, device: torch.Tensor(self.fn(shape, dtype, device)), shape=self.shape, dtype=self.dtype
+        )
```
pmeier, line 192: This is useful in case we have a loader for a subclass, e.g. `ImageLoader`, but want to test a kernel that needs to work on plain tensors. You will see a lot of these `image_loader.unwrap()` calls in the sample inputs functions in the kernel tests.

test/prototype_common_utils.py
```diff
@@ -34,26 +133,104 @@ def __iter__(self):
...
-def make_image(size=None, *, color_space, extra_dims=(), dtype=torch.float32, constant_alpha=True):
+    def __repr__(self):
+        extra = ", ".join(
+            [
+                str(tuple(self.shape)),
+                str(self.dtype).replace("torch.", ""),
+                *[str(extra) for extra in self._extra_repr()],
+            ]
+        )
+        return f"{self._TYPE_NAME}[{extra}]"
```
pmeier, line 210: This is purely to have more readable test names. I would prefer to keep it, but removing this doesn't affect the framework proposed in this PR.

test/prototype_common_utils.py
```diff
@@ -64,36 +241,60 @@ def make_image(size=None, *, color_space, extra_dims=(), dtype=torch.float32, co
...
         features.ColorSpace.RGB_ALPHA,
     ),
+    extra_dims=DEFAULT_EXTRA_DIMS,
     dtypes=(torch.float32, torch.uint8),
-    extra_dims=((), (0,), (4,), (2, 3), (5, 0), (0, 5)),
+    constant_alpha=True,
 ):
     for size, color_space, dtype in itertools.product(sizes, color_spaces, dtypes):
-        yield make_image(size, color_space=color_space, dtype=dtype)
+        yield make_image_loader(size, color_space=color_space, dtype=dtype, constant_alpha=constant_alpha)
```
pmeier, line 273: These functions don't need to yield since we don't load the tensors anyway. Still, I prefer this idiom over the more verbose

```py
loaders = []

for ...:
    loaders.append(...)

for ...:
    loaders.append(...)

return loaders
```

Since we only iterate over the returned values this makes no difference.

test/test_prototype_transforms_kernels.py
```diff
@@ -0,0 +1,394 @@
...
+        reference=None,
+        reference_inputs_fn=None,
+        **closeness_kwargs,
+    ):
+        self.kernel = kernel
+        # smoke test that should hit all valid code paths
+        self.sample_inputs_fn = sample_inputs_fn
+        self.reference = reference
+        self.reference_inputs_fn = reference_inputs_fn or sample_inputs_fn
+        self.closeness_kwargs = closeness_kwargs
```
pmeier, line 30: TODO: add / expand comments here

test/test_prototype_transforms_kernels.py
```diff
@@ -0,0 +1,394 @@
...
+
+    def __str__(self):
+        return self.kernel.__name__
+
+
+def pil_reference_wrapper(pil_kernel):
```
pmeier, line 36: The reference defined in the `KernelInfo` will be passed the same inputs as the kernel. Since we use the PIL kernel as reference for its tensor counterpart, this is simple wrapper to avoid defining the same kind of reference function over and over.

test/test_prototype_transforms_kernels.py
```diff
@@ -0,0 +1,394 @@
...
+        }.get(feature_type)
+        if data_ndim is None:
+            raise pytest.UsageError(
+                f"The number of data dimensions cannot be determined for input of type {feature_type.__name__}."
+            ) from None
+        elif batched_input.ndim <= data_ndim:
+            pytest.skip("Input is not batched.")
+        elif batched_input.ndim > data_ndim + 1:
+            # FIXME: We also need to test samples with more than one batch dimension
+            pytest.skip("Test currently only supports a single batch dimension")
```
pmeier, line 350: TODO

test/test_prototype_transforms_kernels.py
```diff
@@ -0,0 +1,394 @@
...
+    def test_cuda_vs_cpu(self, info, args_kwargs):
+        (input_cpu, *other_args), kwargs = args_kwargs.load("cpu")
+        input_cuda = input_cpu.to("cuda")
+
+        output_cpu = info.kernel(input_cpu, *other_args, **kwargs)
+        output_cuda = info.kernel(input_cuda, *other_args, **kwargs)
+
+        assert_close(output_cuda, output_cpu, check_device=False)
+
+    # FIXME: enforce this only runs on CPU machines
```
pmeier, line 378: TODO

test/prototype_common_utils.py
pmeier: I've removed this once again, since tests like `test_batched_vs_single` need this information before the kernel is called. If a kernel actually depends on the subclass, the `test_scripted_vs_eager` will catch it, since subclass won't get scripted.

test/test_prototype_transforms_kernels.py
pmeier: Fixed in 338523d4989e04a48e7ff1e63cb93a1d45001afe.

pmeier:
@vfdev-5 The failure comes from a test you wanted to fix:

```
=================================== FAILURES ===================================
_ test_correctness_elastic_image_or_mask_tensor[elastic_segmentation_mask-make_segmentation_masks-cpu] _
  File "/home/runner/work/vision/vision/test/test_prototype_transforms_functional.py", line 1643, in test_correctness_elastic_image_or_mask_tensor
    sample[..., in_box[3] - 1, in_box[0]] = torch.arange(20, 20 + c)
  File "/home/runner/work/vision/vision/torchvision/prototype/features/_feature.py", line 87, in __torch_function__
    output = func(*args, **kwargs)
IndexError: index 34 is out of bounds for dimension 1 with size 25
----------------------------- Captured stdout call -----------------------------
torch.Size([1, 64, 76])
torch.Size([4, 1, 64, 76])
torch.Size([0, 64, 76])
torch.Size([4, 0, 64, 76])
torch.Size([9, 64, 76])
torch.Size([4, 4, 64, 76])
torch.Size([1, 25, 18])
```

Given that this PR does not touch this test at all, my best guess is that it was flaky before and depended on a random seed. Plus, we seem to have missed a debug statement:

https://github.com/pytorch/vision/blob/cac4e228c9ca9e7564cb34406e7ebccfdd736976/test/test_prototype_transforms_functional.py#L1659

test/prototype_common_utils.py
pmeier: I've refactored this to avoid overly complicated test ids. They distracted more than being helpful.

vfdev-5:
> The failure comes from a test you wanted to fix:

@pmeier it was fixed. The issue is with input image size that is too small for given bounding boxes. Probably, I changed that again somewhere.

test/test_prototype_transforms_kernels.py
```diff
@@ -0,0 +1,435 @@
...
+        (batched_input, *other_args), kwargs = args_kwargs.load(device)
+
+        feature_type = features.Image if features.is_simple_tensor(batched_input) else type(batched_input)
+        # This dictionary contains the number of rightmost dimensions that contain the actual data.
+        # Everything to the left is considered a batch dimension.
+        data_dims = {
+            features.Image: 3,
+            features.BoundingBox: 1,
+            features.SegmentationMask: 3,
+        }.get(feature_type)
```
pmeier, line 378: To be able to test the transform with batched vs. single inputs, we need to know where the batch dimensions start. For masks this cannot be done, since detection masks use the last three dims for a single mask while segmentation masks use the last two.

pmeier:
@vfdev-5 You were right and I fixed that in a49f0dbea15ed0652241f67bf21402eb4d185dee. However, now we get this failure:

```
=================================== FAILURES ===================================
_ test_correctness_perspective_segmentation_mask[startpoints0-endpoints0-cpu] __
  File "/Users/runner/work/vision/vision/test/test_prototype_transforms_functional.py", line 1496, in test_correctness_perspective_segmentation_mask
    torch.testing.assert_close(output_mask, expected_masks)
  File "/Users/runner/hostedtoolcache/Python/3.7.13/x64/lib/python3.7/site-packages/torch/testing/_comparison.py", line 1359, in assert_close
    msg=msg,
  File "/Users/runner/hostedtoolcache/Python/3.7.13/x64/lib/python3.7/site-packages/torch/testing/_comparison.py", line 1093, in assert_equal

Mismatched elements: 1 / 558 (0.2%)
Greatest absolute difference: 1 at index (0, 2, 17)
Greatest relative difference: inf at index (0, 2, 17)
```

In CI this is only failing on macOS, but this also fails for me locally. Given that we have only a single mismatched element, the test is probably flaky.

vfdev-5:
I do not see from your commit where the size was fixed. More general question, how to find the code of failing test ? It is totally obfuscated to me :)

pmeier:
> I do not see from your commit where the size was fixed.

The issue was that while refactoring the `make_*s` functions I didn't made sure that the "data dimensions", i.e. everything that is not a batch dimension, are constant. For example, before the last commit, setting `size` parameter in `make_segmentation_mask` only applied to a part of the generated samples. This is why the failing test popped back up: you set the `size` and depended on it being at least that, but some samples were smaller.

> More general question, how to find the code of failing test ? It is totally obfuscated to me :)

Not sure what you mean. Could you elaborate? Right now there is no failing test in the new tests, so I'll construct one to show what it looks like. Imagine `horizontal_flip_bounding_box` is not `torch.jit.script`'able. The failing test will look like this

```
________________ TestCommon.test_scripted_vs_eager[cpu-horizontal_flip_bounding_box35] _________________
```

<details><summary>Traceback</summary>
<p>

```
  File "/home/philip/git/pytorch/torchvision/test/test_prototype_transforms_kernels.py", line 343, in test_scripted_vs_eager
    kernel_scripted = torch.jit.script(kernel_eager)
  File "/home/philip/.local/opt/mambaforge/envs/torchvision-dev/lib/python3.7/site-packages/torch/jit/_script.py", line 1344, in script
    qualified_name, ast, _rcb, get_default_args(obj)
RuntimeError: 
  File "/home/philip/git/pytorch/torchvision/torchvision/prototype/transforms/functional/_geometry.py", line 42
    )

    bounding_box[..., [0, 2]] = image_size[1] - bounding_box[..., [2, 0]]
                                                ~~~~~~~~~~~~~~~~~~~~~~~~ <--- HERE

The above exception was the direct cause of the following exception:

  File "/home/philip/git/pytorch/torchvision/test/test_prototype_transforms_kernels.py", line 345, in test_scripted_vs_eager
    raise AssertionError("Trying to `torch.jit.script` the kernel raised the error above.") from error
AssertionError: Trying to `torch.jit.script` the kernel raised the error above.
```

</p>
</details>

From the test name and parametrization you should find everything:
1. The test is defined in the class `TestCommon` and is named `test_scripted_vs_eager`
2. The test was run on the CPU
3. It checked the kernel `horizontal_flip_bounding_box` and specifically the 35th sample input. (In this example all sample inputs fail since the scriptability of the kernel is not dependent on the input).

To debug, you can run this exact test with 

```
pytest 'test/test_prototype_transforms_kernels.py::TestCommon::test_scripted_vs_eager[cpu-horizontal_flip_bounding_box35]'
```

vfdev-5:
I was talking about `test_correctness_perspective_segmentation_mask[startpoints0-endpoints0-cpu]` from failed macosx job.

pmeier:
Top-most error in the traceback:

```
File "/Users/runner/work/vision/vision/test/test_prototype_transforms_functional.py", line 1496, in test_correctness_perspective_segmentation_mask
    torch.testing.assert_close(output_mask, expected_masks)
```

To reproduce:

```
pytest 'test/test_prototype_transforms_functional.py::test_correctness_perspective_segmentation_mask[startpoints0-endpoints0-cpu]'
```

vfdev-5 (approved):
Good to me

test/prototype_common_utils.py
```diff
@@ -108,128 +298,217 @@ def randint_with_tensor_bounds(arg1, arg2=None, **kwargs):
     ).reshape(low.shape)
 
 
-def make_bounding_box(*, format, image_size=(32, 32), extra_dims=(), dtype=torch.int64):
+def make_bounding_box_loader(*, extra_dims=(), format, image_size=None, dtype=torch.float32):
```
vfdev-5, line 301: Why bounding box does not have `num_objects` arg ?

pmeier: Because we defined a bounding box as `(*, 4)`, i.e. `features.BoundingBox([0, 0, 10, 10], ...)` is valid although it only has a single dimension. Thus, if you want to have multiple boxes, set `extra_dims=(num_objects,)`

PR #5880, opened 2022-04-26
remove magic comma in some places

pmeier:
I discovered these while working on #5879. 

By default `black` treats a trailing comma in a sequence as ["magic"](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#the-magic-trailing-comma) and explodes it regardless of the number of items in it. This is usually useful if you want to communicate that a sequence might grow in the future and you than only need to add a new line rather than reformatting the sequence.

However, sometimes the original author did not intend this and in turn `black` turns legible code into a mess. This PR fixes two instances where the new format is especially bad. In general, we can detect all such cases by running `black -C` and looking at the differences. There are 105 files in the `torchvision` package that will be formatted. I haven't looked at the diffs to see how bad the situation is.


NicolasHug (approved):
> black turns legible code into a mess

image



(Couldn't resist trolling)


Thanks Philip, LGTM!

PR #3641, opened 2021-04-07
Bump PIL dependency to >=5.3.0 and run 5.3.0 on some CI runs

NicolasHug:
Following https://github.com/pytorch/vision/pull/3638#issuecomment-814274883, if we go this route

Also make 3.6 runs rely on 5.3.0 instead of the latest PIL version

Closes https://github.com/pytorch/vision/pull/3638

PR #5898, opened 2022-04-27
Update transforms for PIL deprecation

kylematoba:
Updates torchvision transformations to the call recommended at https://github.com/python-pillow/Pillow/blob/5d070222d21138d2ead002fd33fdf5adcb708941/src/PIL/Image.py#L93 (Pillow 9.1.0 release).


datumbox (commented):
Thanks for the contribution @kylematoba.

@pmeier @vfdev-5 Any thoughts on whether we should make the use of `.Transform` conditional on the PIL version? We probably need to investigate when this was introduced. Unless we just want to pump the minimum required PIL version.

pmeier (changes requested):
> It worked on Pillow 6.0.0 for me.

I doubt that:

```py
>>> import PIL.Image
>>> PIL.__version__
'9.0.1'
>>> PIL.Image.Transpose
AttributeError: module 'PIL.Image' has no attribute 'Transpose'
```

This was only introduced in python-pillow/Pillow#5954 which made it into the latest `9.1.0` release. I talked offline with @vfdev-5 about that some time ago. He wanted to pick it up, but I'm not aware of a PR regarding this yet.

> Any thoughts on whether we should make the use of `.Transform` conditional on the PIL version? [...] Unless we just want to pump the minimum required PIL version.

I'm not opposed to that. Right now we require

https://github.com/pytorch/vision/blob/f0148413640d3e9efa774797a84f021ba8c356ea/setup.py#L68

and this was introduced in #3641. Bumping from `>=5.3.0` to `>=9.1.0` is quite a difference (~ 2.5 years), but it might still be worth it to avoid cluttering our code with version checks again that were removed in #3641. cc @NicolasHug

kylematoba:
Oh yeah, you're right. My IDE was caching the original version of PIL. I thought it was weird :joy: .

I guess you can delay this until around the PIL hard change date of July 1, 2023:
```
DeprecationWarning: AFFINE is deprecated and will be removed in Pillow 10 (2023-07-01). Use Transform.AFFINE instead.
```

If you want, close this PR and I'll add a note to my calendar to look at this again in June 2023.

pmeier:
> I guess you can delay this until around the PIL hard change date of July 1, 2023

Having user facing warnings for something they didn't do is not desirable. So we should handle this before the next release. 

If you are willing, please keep this PR open and finish it as soon as we decided which way to go:

1. If we decide to require `Pillow>=9.1`, you'll have to update `setup.py` and can keep the changes you already have.
2. If we decide not to update our PIL requirement, we will need to have something like

```py
import PIL.Image

if tuple(int(part) for part in PIL.__version__.split(".")) >= (9, 1):
    FLIP_LEFT_RIGHT = PIL.Image.FLIP_LEFT_RIGHT
    ...
else:
    FLIP_LEFT_RIGHT = PIL.Image.Transpose.FLIP_LEFT_RIGHT
    ...

def kernel(img):
    ...
    img.transpose(FLIP_LEFT_RIGHT)
    ...
```

In case we need these constants in multiple modules, it probably makes sense to have a `_pil_constants.py` module and put the snippet above in there. With that you can do something like

```py
from . import _pil_constants

def kernel(img):
    ...
    img.transpose(_pil_constants.FLIP_LEFT_RIGHT)
    ...
```

datumbox:
@kylematoba Thanks for the additional info and for notifying us about this. :)

@pmeier Thanks for the analysis. Requiring users to be on the latest version of PIL is kind of a bummer. In principle I don't think it's a bad thing to bump the minimum version a bit but going from >=5.x to >=9.1 is quite a big jump. I don't know how many breaking changes and feature deprecations exist across 4 major versions and what effect this jump will have to users who make use of PIL directly on their software. Option 2 looks viable to me. Do we know how many constants we need to handle?

@NicolasHug I would love to get your thoughts on the above.

pmeier:
I wrote a small script to check [all deprecated constants](https://github.com/radarhere/Pillow/blob/5d7b6c9d88082b0a18a54df09b28f9e7cd8167a3/docs/deprecations.rst#constants) and this is the result:

```
--------------------------------------------------------------------------------
Image.FLIP_LEFT_RIGHT
--------------------------------------------------------------------------------
torchvision/transforms/functional_pil.py::57
test/test_transforms.py::1527
--------------------------------------------------------------------------------
Image.FLIP_TOP_BOTTOM
--------------------------------------------------------------------------------
torchvision/transforms/functional_pil.py::65
test/test_transforms.py::1524
--------------------------------------------------------------------------------
Image.BILINEAR
--------------------------------------------------------------------------------
torchvision/transforms/functional_pil.py::243
test/test_transforms_tensor.py::781
test/test_onnx.py::420
--------------------------------------------------------------------------------
Image.NEAREST
--------------------------------------------------------------------------------
torchvision/transforms/functional_pil.py::317
torchvision/transforms/functional_pil.py::333
torchvision/transforms/functional.py::395
torchvision/transforms/functional.py::575
torchvision/transforms/functional.py::655
torchvision/transforms/functional.py::1015
torchvision/transforms/functional.py::1108
torchvision/transforms/transforms.py::301
torchvision/transforms/transforms.py::758
torchvision/transforms/transforms.py::872
torchvision/transforms/transforms.py::1271
torchvision/transforms/transforms.py::1392
test/test_transforms_tensor.py::780
--------------------------------------------------------------------------------
Image.AFFINE
--------------------------------------------------------------------------------
torchvision/transforms/functional_pil.py::326
test/test_transforms_tensor.py::775
--------------------------------------------------------------------------------
Image.BICUBIC
--------------------------------------------------------------------------------
torchvision/transforms/functional_pil.py::350
--------------------------------------------------------------------------------
Image.PERSPECTIVE
--------------------------------------------------------------------------------
torchvision/transforms/functional_pil.py::359
--------------------------------------------------------------------------------
Image.LINEAR
--------------------------------------------------------------------------------
test/test_transforms.py::176
```

NicolasHug:
Re minimal supported version: PIL has a deprecation policy of 2 major versions, so if we want to be a good citizen, that means the minimum PIL version we should support is probably `current version - 1`?

Regardless, we'll have to support both ways (the deprecated one and the new one) until our minimum version is `>=` the version that removes the deprecated way. I don't see a better solution than the wrappers proposed above by Philip.

pmeier:
> minimum PIL version we should support is probably `current version - 1`?

Just to be clear: Given that the current version is `9.1.0` this means we can require `pillow>=8`, correct?

datumbox:
@pmeier I'm OK with that. BTW we should investigate what's the best way to keep all dependencies in one place. Bumping PIL right now involves updating 100 places across CI, building scripts and configs.

kylematoba:
Okay, I'll update the PR to include to make all instances of the deprecated constants contingent upon `PIL.__version__.split(".")) >= (9, 1)`. Give me a few days.

kylematoba:
I added `transforms/_pil_constants.py` which makes the constants contingent on the PIL version. 

I extended my earlier changes to the docstrings, type annotations, and tests listed at https://github.com/pytorch/vision/pull/5898#issuecomment-1111912399.

Note the slightly confusing rename from `LINEAR` to `BILINEAR`:
https://github.com/python-pillow/Pillow/blob/main/src/PIL/Image.py#L66
I kept the version-abstracted name as LINEAR (this is only used in a single test).

torchvision/transforms/functional.py
```diff
@@ -392,7 +392,7 @@ def resize(
             :class:`torchvision.transforms.InterpolationMode`.
             Default is ``InterpolationMode.BILINEAR``. If input is Tensor, only ``InterpolationMode.NEAREST``,
             ``InterpolationMode.BILINEAR`` and ``InterpolationMode.BICUBIC`` are supported.
-            For backward compatibility integer values (e.g. ``PIL.Image.NEAREST``) are still acceptable.
+            For backward compatibility integer values (e.g. ``PIL.Image(.Resampling).NEAREST``) are still acceptable.
```
datumbox, line 395: I think we should deprecate this.

cc @NicolasHug @pmeier

pmeier: @vfdev-5

torchvision/transforms/_pil_constants.py
```diff
@@ -0,0 +1,22 @@
+
+import PIL
+from PIL import Image
+
+if tuple(int(part) for part in PIL.__version__.split(".")) >= (9, 1, 0):
```
pmeier, line 5: Nit: we don't need the `0` in the version tuple here. Python is checking tuples elementwise. If we hit the case where one tuple has no elements left, the other tuple is greater by definition. For example

```py
>>> (9, 1, 0) >= (9, 1)
True
>>> (9, 1, -1) >= (9, 1)
True
>>> (9, 0, 1) >= (9, 1)
False
>>> (9, 2, 0) >= (9, 1)
True
```

```suggestion
if tuple(int(part) for part in PIL.__version__.split(".")) >= (9, 1):
```

torchvision/transforms/_pil_constants.py
```diff
@@ -0,0 +1,22 @@
...
+    FLIP_LEFT_RIGHT = Image.Transpose.FLIP_LEFT_RIGHT
+    FLIP_TOP_BOTTOM = Image.Transpose.FLIP_TOP_BOTTOM
+    BILINEAR = Image.Resampling.BILINEAR
+    NEAREST = Image.Resampling.NEAREST
+    AFFINE = Image.Transform.AFFINE
+    BICUBIC = Image.Resampling.BICUBIC
+    PERSPECTIVE = Image.Transform.PERSPECTIVE
+    LINEAR = Image.Resampling.BILINEAR
```
pmeier, line 13: Could you order them by module and then alphanumeric? 

```suggestion
    BICUBIC = Image.Resampling.BICUBIC
    BILINEAR = Image.Resampling.BILINEAR
    NEAREST = Image.Resampling.NEAREST
    LINEAR = Image.Resampling.BILINEAR
    
    AFFINE = Image.Transform.AFFINE
    FLIP_LEFT_RIGHT = Image.Transpose.FLIP_LEFT_RIGHT
    FLIP_TOP_BOTTOM = Image.Transpose.FLIP_TOP_BOTTOM
    PERSPECTIVE = Image.Transform.PERSPECTIVE
```

Same for the other branch please.

torchvision/transforms/functional.py
```diff
@@ -572,7 +572,7 @@ def resized_crop(
...
```
pmeier, line 575: I'm ok with this style in the docstring. Any objections?

pmeier (commented):
Linting CI job is [failing](https://app.circleci.com/pipelines/github/pytorch/vision/17189/workflows/0b94a916-3708-4557-b85c-5dc7f7fa2d9c/jobs/1392954). Could you follow our instructions [here](https://github.com/pytorch/vision/blob/main/CONTRIBUTING.md#formatting) to auto format the files you've edited? Otherwise LGTM. Thanks @kylematoba!

torchvision/transforms/functional.py
vfdev-5: I'm OK with deprecating this.

vmoens:
A quick comment on the warning from a user perspective. I don't have the full context of this PIL deprecation issue so forgive me if there's anything I missed.

Every run of torchrl scripts we make gives us as many warnings as the number of processes that are importing torchvision (3 warnings / process x 32 process = almost 200 lines of warnings). This gives us [long logs](https://github.com/facebookresearch/rl/issues/92#issue-1214680724) where it's hard to find what's going on.

Besides, correct me if I'm wrong, but it feels like it's a warning on which I have little agency -- it's not really that we're importing anything that will be deprecated: it's an issue in torchvision code, not ours. I think it's confusing for the users, as they're not fully informed of what's going on and who should fix what. 

Apart from hiding those specific warnings (my last option is to do a temporary patch in torchrl to hide specific deprecation warnings, but i'm a bit reluctant to make a PR like this), is there anything that could be done?

torchvision/transforms/functional.py
NicolasHug: `()` is fine, although as a nit, I think backets `[]` are more common (typically in `man` pages)

pmeier:
@vmoens 

> Apart from hiding those specific warnings, [...] is there anything that could be done?

The only other option is to force `PIL < 9.1`. After this PR there should be no warnings regardless of the `PIL` version. Unfortunately, until this patch will be included in a stable release will be 1-2 months. However, you could use the nightly build.

torchvision/transforms/functional.py
pmeier: @kylematoba Could you address this in all docstrings?

torchvision/transforms/functional.py
pmeier: @kylematoba Would you be up to send a follow-up PR that deprecates this?

pmeier (approved):
I've checked CI and couldn't find a warning from PIL anywhere. Seems like we patched all places.

One small thing to do, otherwise LGTM. Thanks a lot @kylematoba!

torchvision/transforms/functional.py
kylematoba: Sure, so replace instances of `()` with `[]` in my changes thusfar, and a second PR that just removes all of the lines about `For backward compatibility integer values`?

pmeier: > Sure, so replace instances of `()` with `[]` in my changes thusfar,

Yes.


> and a second PR that just removes all of the lines about `For backward compatibility integer values`?

No. The docstring need to be changed indicating that passing `int`'s is actually deprecated, e.g.

https://github.com/pytorch/vision/blob/62740807c18e68bb0acd85895dca527f9a655bd5/torchvision/transforms/functional.py#L1029-L1031

Furthermore, we need to change the warning 

https://github.com/pytorch/vision/blob/62740807c18e68bb0acd85895dca527f9a655bd5/torchvision/transforms/functional.py#L415-L419

to include the concrete deprecation date, e.g.

https://github.com/pytorch/vision/blob/62740807c18e68bb0acd85895dca527f9a655bd5/torchvision/transforms/functional.py#L1042-L1045

For this we will be deprecating in `0.13` and remove in `0.15`.

torchvision/transforms/_pil_constants.py
```diff
@@ -0,0 +1,23 @@
+import PIL
+from PIL import Image
+
```
NicolasHug, line 3:
```suggestion

# See https://pillow.readthedocs.io/en/stable/releasenotes/9.1.0.html#deprecations
# TODO: Remove this file once PIL minimal version is >= 9.1

```

NicolasHug (approved):
Thanks for the PR @kylematoba , minor comment below but LGTM

PR #5902, opened 2022-04-27
Add swin_t to slow list

jdsgomes:
cc @datumbox


datumbox:
@jdsgomes you will need to regenerate the expected file as the input of the test changed.

datumbox (approved):
LGTM, thanks!

PR #5903, opened 2022-04-27
Skip big model in test to reduce memory usage in CI

YosuaMichael:
Seems like the CI getting flaky caused by OOM when testing big model ([sample](https://app.circleci.com/pipelines/github/pytorch/vision/17065/workflows/85192520-7d29-4f6f-a850-8082ea4eb3f8/jobs/1382986)) 

This PR aim to enable skipping the test for chosen big models (for example vit_h_14).

cc @datumbox


test/test_models.py
```diff
@@ -19,13 +19,19 @@
...
+}
+
 
 def get_models_from_module(module):
     # TODO add a registration mechanism to torchvision.models
     return [
         v
         for k, v in module.__dict__.items()
-        if callable(v) and k[0].lower() == k[0] and k[0] != "_" and k != "get_weight"
+        if callable(v) and k[0].lower() == k[0] and k[0] != "_" and k != "get_weight" and k not in skipped_models
```
datumbox, line 34: The specific method is being used in a few places. I'm not sure it's a good idea to remove them from this method. Also note that this one is being used for multiple modules so hardcoding classification models here makes less sense.

Instead we should add this on this skip on the `test_classification_model` method (and potentially in all other methods so that it can be used across all tasks. We should do a proper pytest skip to indicate that the test was skipped and potentially give the option to the user to run it. One common idiom used in TorchVision for such cases is using an env variable (see [this](https://github.com/pytorch/vision/blob/main/test/test_models.py#L20) for example).

test/test_models.py
```diff
@@ -328,6 +329,18 @@ def _check_input_backprop(model, inputs):
...
 
 
+# skip big models to reduce memory usage on CI test
+skipped_big_models = {
+    "vit_h_14",
+    "regnet_y_128gf",
+}
+
+
+def do_skip_big_model(model_name):
```
datumbox, line 339: Nit: Since we don't use this more than once, we can lose the extra method and move the check directly on the test.

test/test_models.py
```diff
@@ -591,6 +604,8 @@ def test_classification_model(model_fn, dev):
         "input_shape": (1, 3, 224, 224),
     }
     model_name = model_fn.__name__
+    if dev == torch.device("cuda") and SKIP_BIG_MODEL:
```
datumbox, line 607: I believe `dev` [is a string](https://github.com/pytorch/vision/blob/main/test/common_utils.py#L113), hence `dev == "cuda"` might be required. You should confirm with a debugger.

datumbox (commented):
Thanks @YosuaMichael. It looks good, just 2 minor comments for your attention.

test/test_models.py
```diff
@@ -605,7 +620,7 @@ def test_classification_model(model_fn, dev):
     _check_jit_scriptable(model, (x,), unwrapper=script_model_unwrapper.get(model_name, None), eager_out=out)
     _check_fx_compatible(model, x, eager_out=out)
 
-    if dev == torch.device("cuda"):
+    if dev == "cuda":
```
YosuaMichael, line 623: When debugging, I found out that dev is actually a python str, hence the previous expression `dev == torch.device("cuda")` will always False.

@datumbox is it okay to fix this in this PR, or should I open new PR for this since it is not so related with the PR?

test/test_models.py
YosuaMichael: Ah yeah, it is indeed a string and I have confirmed it (already update the PR).

datumbox (approved):
The changes LTGM, we can merge if the CI is green.

PR #5912, opened 2022-04-28
fix test_data_loader on Windows and macOS

pmeier:
Addresses https://github.com/pytorch/vision/pull/5801#issuecomment-1111125791.

### DIFF
--- a/test/test_prototype_builtin_datasets.py
+++ b/test/test_prototype_builtin_datasets.py
@@ -125,7 +125,7 @@ class TestCommon:
 
         pickle.dumps(dataset)
 
-    # This has to be a proper method, since lambda's or local functions
+    # This has to be a proper function, since lambda's or local functions
     # cannot be pickled, but this is a requirement for the DataLoader with
     # multiprocessing, i.e. num_workers > 0
     def _collate_fn(self, batch):
--- a/test/test_prototype_builtin_datasets.py
+++ b/test/test_prototype_builtin_datasets.py
@@ -125,6 +125,12 @@ class TestCommon:
 
         pickle.dumps(dataset)
 
+    # This has to be a proper method, since lambda's or local functions
+    # cannot be pickled, but this is a requirement for the DataLoader with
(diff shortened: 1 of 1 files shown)


test/test_prototype_builtin_datasets.py
```diff
@@ -125,6 +125,12 @@ def test_serializable(self, test_home, dataset_mock, config):
 
         pickle.dumps(dataset)
 
+    # This has to be a proper method, since lambda's or local functions
```
NicolasHug, line 128:
```suggestion
    # This has to be a proper function, since lambda's or local functions
```

test/test_prototype_builtin_datasets.py
```diff
@@ -125,6 +125,12 @@ def test_serializable(self, test_home, dataset_mock, config):
...
+    # cannot be pickled, but this is a requirement for the DataLoader with
```
NicolasHug, line 129: closures work, but not lambdas? I always thought they had the same issues

NicolasHug (approved):
Thanks @pmeier

test/test_prototype_builtin_datasets.py
pmeier: AFAIK, a closure is a function that was parametrized and returned by another function. Thus, something like

```py
def foo(bar):
    def closure():
        return bar
    
    return closure

closure = foo("bar")
print(closure())
print(closure())
```

This would make a closure a special case of a general local function. Using a local function for the code at hand would look like

```
def test():
    def collate_fn(batch):
        return batch

    dl = DataLoader(..., collate_fn=collate_fn)
```

Going by the definition above `collate_fn` is a local function, but not a closure. This is why I wrote

> lambda's or local functions cannot be pickled

Happy to adapt if you have better definitions.

pmeier: In that case, no local functions are not supported and the comment already states that.

NicolasHug: 🤦

PR #5914, opened 2022-04-28
add Windows and macOS CI jobs for prototype tests

### LINKED ISSUE #5550: Run prototype datasets test also on Windows and macOS?
There are now three separate occasions of bugs in the prototype datasets that were not picked up by our CI.

- #4980
- #5227
- #5549

We didn't notice them, because they happen on Windows on macOS, but we only run the prototype tests on Linux. Given that the prototypes work with files and the path handling is different on all platforms, this can be problematic.

The torchdata team brought the failures to our attention, since they run our dataset tests on the full matrix. I propose we also run the prototype datasets tests on Windows and macOS to avoid these post mortems. To limit the extra needed CI resources I would only run the tests only on Python3.7.

That means we are looking at 2 extra CI runs per PR with roughly 10 minutes runtime.

cc @pmeier @seemethere @bjuncek

pmeier:
Closes #5550. 

Apart from adding Windows and macOS jobs, this also migrates the prototype tests from CircleCI to Github Actions since CircleCI has no convenient way to setup a Python environment on all three platforms. There is the [Python orb](https://circleci.com/developer/orbs/orb/circleci/python), but it is not supported on Windows. Thus, we would need a rather complicated workflow (think current regular unittest setup, but simplified a little) with very little gain. 

The prototype tests don't require a GPU or a beefy machine and thus the OSS version of Github Actions is sufficient. As an added benefit, Github Actions is free for OSS projects and thus helps keeping the CI costs down (see #5479).


.github/workflows/prototype-tests.yml
```diff
@@ -0,0 +1,44 @@
...
+        uses: actions/checkout@v3
+
+      - name: Install PyTorch nightly builds
+        run: pip install --progress-bar=off --pre torch torchdata --extra-index-url https://download.pytorch.org/whl/nightly/cpu/
+
+      - name: Install torchvision
+        run: pip install --progress-bar=off --no-build-isolation --editable .
+
+      - name: Install other prototype dependencies
+        run: pip install --progress-bar=off scipy pycocotools h5py iopath
```
datumbox, line 37: Non-blocking Nit: The fact that we have to add the dependency definition in yet another place is far from ideal. We should look into options to centralize them. Not sure how easy this is to achieve across pip and conda.

datumbox (approved):
LGTM. I briefly checked the jobs and seem to work.

The failing test is unrelated and resolved on main. I think we can merge this.

.github/workflows/prototype-tests.yml
pmeier: Yeah, `conda` vs. `pip` will be a problem. Otherwise I'll tackle this.

datumbox: Well having them in 2 places is still better than having them in 10. So even if we need 2 different approaches, it's still worth investigating.

.github/workflows/prototype-tests.yml
```diff
@@ -0,0 +1,44 @@
...
+
+      - name: Install test requirements
+        run: pip install --progress-bar=off pytest pytest-mock
+
+      - name: Run prototype tests
+        shell: bash
+        run: pytest --durations=20 test/test_prototype_*.py
```
vfdev-5, line 44: @pmeier let's add `-vvv` flag to see more info on failing tests ?

vfdev-5:
@pmeier can you add caching installed deps to speed up tests when re-run

.github/workflows/prototype-tests.yml
pmeier: AFAIK, the extra verbosity only applies to the output during the run. The information displayed by `pytest` for failing tests in teh report should be the same. Do you have a counter example?

vfdev-5: https://github.com/pytorch/vision/runs/6254487636?check_suite_focus=true#step:9:195

pmeier: The report is there:

```

================================== FAILURES ===================================
___ test_correctness_perspective_bounding_box[startpoints2-endpoints2-cpu] ____
  File "D:\a\vision\vision\test\test_prototype_transforms_functional.py", line 1210, in test_correctness_perspective_bounding_box
    torch.testing.assert_close(output_bboxes, expected_bboxes)
  File "c:\hostedtoolcache\windows\python\3.7.9\x64\lib\site-packages\torch\testing\_comparison.py", line 1321, in assert_close
    msg=msg,
  File "c:\hostedtoolcache\windows\python\3.7.9\x64\lib\site-packages\torch\testing\_comparison.py", line 1074, in assert_equal
    raise error_metas[0].to_error()

Mismatched elements: 1 / 16 (6.2%)
Greatest absolute difference: 1.4543533325195312e-05 at index (0, 2) (up to 1e-05 allowed)
Greatest relative difference: 1.1173174124760395e-05 at index (0, 2) (up to 1.3e-06 allowed)
```

Is this not what you are looking for?

pmeier:
Yup, can do. Still, since we can only cache third-party dependencies other than  PyTorch core, we will shave off very little. The most time is spent installing PyTorch, building torchvision and running the tests.

vfdev-5:
You can cache pytorch with a daily key and thus drop the cache every day and thus reruning tests during the same day will be faster

.github/workflows/prototype-tests.yml
NicolasHug: I don't remember the details but I think scikit-learn maintains its dependencies in a centralized place https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/_min_dependencies.py. We could take inspiration from that

pmeier: I'll open an issue with all our requirements to discuss what would be the best way to address this.

Issue #5558, opened 2022-03-07
repr of `_Feature` attribute will print the entire underlying tensor


NicolasHug:
The repr of `_Feature` attributes also prints the underlying tensor data, which causes very long outputs:

```py
[ins] In [21]: Image(torch.rand(3, 10, 10)).draw_bounding_box
Out[21]: 
<bound method Image.draw_bounding_box of Image([[[4.0873e-01, 7.5065e-01, 3.9609e-01, 8.0974e-01, 5.4882e-01,
          2.1480e-01, 1.9220e-01, 8.9547e-02, 3.0651e-01, 4.5459e-01],
         [9.9827e-01, 4.7832e-01, 7.1335e-01, 8.0415e-02, 7.2666e-02,
          9.9300e-01, 8.6541e-01, 4.3845e-01, 8.4263e-01, 4.7697e-01],
         [3.2785e-01, 7.8933e-01, 1.5514e-01, 3.8665e-01, 8.1245e-01,
          4.2014e-01, 2.2072e-01, 7.7170e-02, 5.5109e-01, 5.3057e-02],
         [5.4669e-01, 9.1501e-02, 9.2565e-01, 2.2452e-01, 5.6280e-02,
          7.8277e-01, 4.4243e-01, 7.7856e-01, 1.0936e-01, 8.9094e-01],
         [4.8039e-01, 3.1899e-01, 2.5593e-02, 4.7847e-01, 2.2275e-01,
          1.0460e-01, 1.2189e-01, 7.3094e-01, 4.5118e-02, 3.4907e-01],
         [7.4443e-01, 9.9318e-01, 2.2914e-01, 8.3202e-02, 5.8815e-01,
          8.0911e-01, 8.1423e-01, 8.2576e-01, 3.3891e-01, 9.2449e-01],
         [7.5176e-01, 5.0756e-01, 4.1137e-01, 6.0892e-01, 5.3783e-01,
          9.1445e-01, 6.3350e-01, 2.4363e-01, 7.2003e-01, 5.5256e-01],
         [7.5895e-01, 1.7527e-01, 9.1331e-01, 7.5056e-01, 1.8107e-01,
          5.7988e-01, 4.5102e-01, 8.0804e-01, 8.2611e-01, 8.0811e-01],
         [6.6590e-01, 1.6192e-01, 5.8999e-01, 2.6620e-01, 1.6387e-01,
[...]

PR #5916, opened 2022-04-28
remove custom __repr__ from prototype features

pmeier:
Amongst other things, pytorch/pytorch#73459 added support for automatically [include the name of the tensor subclass in the `__repr__`](https://github.com/pytorch/pytorch/blob/7422ccea8b564132b50227b35a9bfbf06730d71a/torch/_tensor_str.py#L307). Thus, we no longer need to do this ourselves.

The same PR also changed the signature of `Tensor.__repr__` which lead to [`mypy` errors](https://app.circleci.com/pipelines/github/pytorch/vision/17152/workflows/0580fe16-3f0c-4da5-9464-3e61d1c6c95d/jobs/1389825).

### DIFF
--- a/torchvision/prototype/features/_feature.py
+++ b/torchvision/prototype/features/_feature.py
@@ -83,6 +83,3 @@ class _Feature(torch.Tensor):
             return cls.new_like(args[0], output, dtype=output.dtype, device=output.device)
         else:
             return output
-
-    def __repr__(self) -> str:
-        return cast(str, torch.Tensor.__repr__(self)).replace("tensor", type(self).__name__)


datumbox (approved):
LGTM, thanks!

NicolasHug:
Do you know if this affects https://github.com/pytorch/vision/issues/5558 ?

pmeier:
> Do you know if this affects https://github.com/pytorch/vision/issues/5558 ?

Yeah I also checked, but unfortunately no. I think the issue there depends on how we add methods onto tensors.

PR #5920, opened 2022-04-28
updating mnasnet docs

jdsgomes:
Part of https://github.com/pytorch/vision/issues/5833

Adding revamp docs for MNASNET.

cc @NicolasHug

### DIFF
--- a/docs/source/models/mnasnet.rst
+++ b/docs/source/models/mnasnet.rst
@@ -27,4 +27,3 @@ more details about this class.
     mnasnet0_75
     mnasnet1_0
     mnasnet1_3
-
--- a/docs/source/models/mnasnet.rst
+++ b/docs/source/models/mnasnet.rst
@@ -27,3 +27,4 @@ more details about this class.
     mnasnet0_75
     mnasnet1_0
     mnasnet1_3
+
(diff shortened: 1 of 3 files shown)


NicolasHug:
Thanks @jdsgomes 

I noticed that the 1.3 and 0.75 weights are missing / non-existent. I added them to the docs and then removed them after realizing it made the doc generation fail. Do you know if we're planning to support pre-trained weights for these models before the next release? If we don't, we might nee to change our docs to indicate that no pre-trained weights exist (and possibly remove the existence of the parameter altogether)

jdsgomes:
> Thanks @jdsgomes
> 
> I noticed that the 1.3 and 0.75 weights are missing / non-existent. I added them to the docs and then removed them after realizing it made the doc generation fail. Do you know if we're planning to support pre-trained weights for these models before the next release? If we don't, we might nee to change our docs to indicate that no pre-trained weights exist (and possibly remove the existence of the parameter altogether)

I don't know if we have plans to add those before next release, lets follow up with this offline. Merging for now, thank you for the review!

NicolasHug:
@jdsgomes it might still be worth updating the current docs, which for now contain ambiguous / incorrect info:

> The following model builders can be used to instanciate an MNASNet model, with or
without pre-trained weights

The `weights` parameter is also documeted as 

> The pretrained weights to use

(with broken links). But if we don't support pre-trained weights, we should make sure to indicate it clearly. ATM, users will believe that we support pre-trained weights while we don't.

PR #5926, opened 2022-04-30
Add edit on github button

oke-aditya:
A simple button 

image

This is already on pytorch docs home page :smile:  just replicating.

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -1,3 +1,5 @@
+:github_url: https://github.com/pytorch/vision
+
 torchvision
 ===========
 This library is part of the `PyTorch


NicolasHug (commented):
Thanks for the PR @oke-aditya 

For now, it doesn't seem to render properly: 

image


I'm also not sure it's extremely useful, at least as-is: The link says "Edit on GitHub" but all this does is to add a link to the GitHub repo - it doesn't open an editor or something like that.

oke-aditya:
It seems to be working 

https://output.circle-artifacts.com/output/job/f4f1f520-a542-4bd9-a4a1-96525e3718e3/artifacts/0/docs/index.html

image


Nope sadly it doesn't add any editor etc. I just added this to be consistent with pytorch core.

NicolasHug:
> It seems to be working

Hm, hopefully it's only local to me then.


I'd be happy to add the link if we can chance the description to "GitHub repository" or something like this, but right now the link isn't descriptive enough to be included IMO.

oke-aditya:
I cannot actually find a way to change the description "Edit on Github". but this seems something standard and probably non customizable. Pytorch Lightning too uses the same, pytorch core too.

Although we have an option of changing the url. It can point somewhere else if you prefer :)

PR #5931, opened 2022-05-02
[WIP] Adding m1 cross compilation workflows

atalman:
Adding m1 cross compilation workflows


.circleci/regenerate.py
```diff
@@ -195,7 +196,7 @@ def generate_upload_workflow(base_workflow_name, os_type, btype, cu_version, *,
     }
 
     if btype == "wheel":
-        d["subfolder"] = "" if os_type == "macos" else cu_version + "/"
+        d["subfolder"] = "" if (os_type == "macos" or os_type == "macos_arm64") else cu_version + "/"
```
pmeier, line 199: Nit:

```suggestion
        d["subfolder"] = "" if os_type.startswith("macos") else cu_version + "/"
```

Also applies to the two other places below.

atalman:
will not need it

PR #5933, opened 2022-05-03
Document Keypoint RCNN separately

NicolasHug:
Closes https://github.com/pytorch/vision/issues/5882

This PR documents Keypoint RCNN separately, in its own section and with its own table.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -348,9 +348,14 @@ def inject_weight_metadata(app, what, name, obj, options, lines):
             lines.append("")
 
 
-def generate_weights_table(module, table_name, metrics):
+def generate_weights_table(module, table_name, metrics, include_pattern=None, exclude_pattern=None):
     weight_enums = [getattr(module, name) for name in dir(module) if name.endswith("_Weights")]
     weights = [w for weight_enum in weight_enums for w in weight_enum]
+
+    if include_pattern is not None:
+        weights = [w for w in weights if include_pattern in str(w)]
+    if exclude_pattern is not None:
+        weights = [w for w in weights if exclude_pattern not in str(w)]
 
     metrics_keys, metrics_names = zip(*metrics)
     column_names = ["Weight"] + list(metrics_names) + ["Params", "Recipe"]
@@ -377,7 +382,8 @@ def generate_weights_table(module, table_name, metrics):
 
 
 generate_weights_table(module=M, table_name="classification", metrics=[("acc@1", "Acc@1"), ("acc@5", "Acc@5")])
-generate_weights_table(module=M.detection, table_name="detection", metrics=[("box_map", "Box MAP")])
+generate_weights_table(module=M.detection, table_name="detection", metrics=[("box_map", "Box MAP")], exclude_pattern="Keypoint")
+generate_weights_table(module=M.detection, table_name="detection_keypoint", metrics=[("box_map", "Box MAP"), ("kp_map", "Keypoint MAP")], include_pattern="Keypoint")
 generate_weights_table(
     module=M.segmentation, table_name="segmentation", metrics=[("miou", "Mean IoU"), ("pixel_acc", "pixelwise Acc")]
 )
--- a/docs/source/models_new.rst
+++ b/docs/source/models_new.rst
@@ -111,6 +111,21 @@ Box MAPs are reported on COCO
 
 .. include:: generated/detection_table.rst
 
+Keypoint detection
+------------------
+
+The following keypoint detection models are available, with or without
+pre-trained weights:
+
+.. toctree::
+   :maxdepth: 1
+
+   models/keypoint_rcnn
+
+Box and Keypoint MAPs are reported on COCO:
+
+.. include:: generated/detection_keypoint_table.rst
+
 
 Video Classification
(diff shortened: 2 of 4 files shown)


docs/source/conf.py
```diff
@@ -348,10 +348,15 @@ def inject_weight_metadata(app, what, name, obj, options, lines):
...
 
-def generate_weights_table(module, table_name, metrics):
+def generate_weights_table(module, table_name, metrics, include_pattern=None, exclude_pattern=None):
     weight_enums = [getattr(module, name) for name in dir(module) if name.endswith("_Weights")]
     weights = [w for weight_enum in weight_enums for w in weight_enum]
 
+    if include_pattern is not None:
+        weights = [w for w in weights if include_pattern in str(w)]
+    if exclude_pattern is not None:
+        weights = [w for w in weights if exclude_pattern not in str(w)]
```
NicolasHug, line 358: Not a die-hard fan of this, but we have to special-case this generation function (or its inputs) in some way. Happy to consider other suggestions.

docs/source/models_new.rst
```diff
@@ -111,6 +111,21 @@ Box MAPs are reported on COCO
 
 .. include:: generated/detection_table.rst
 
+Keypoint detection
+------------------
```
NicolasHug, line 115: This has the same level of heading as the table above. Should it be instead at the same level as "classification", "detection", "segmentation" sections?

docs/source/conf.py
jdsgomes: The other alternative makes the code simpler here but moves complexity to the caller, i.e., they could provide a validation pattern as regex.

```python
def generate_weights_table(module, table_name, metrics, include_pattern=".*"):
    weight_enums = [getattr(module, name) for name in dir(module) if name.endswith("_Weights")]
    weights = [w for weight_enum in weight_enums for w in weight_enum if re.match(include_pattern, w)]
    ...
```
regex would include both include and exclude patterns

NicolasHug: I thought about using regex, but I could not find a `include_pattern` where

```
re.match(include_pattern, w)
```

would be Truthy for weights that *do not* contain `"Keypoint"`. Any pointer @jdsgomes ?

docs/source/models_new.rst
jdsgomes: I think we should split Object Detection, Instance Segmentation and Person Keypoint Detection. In that case yes, this Keypoint detection would be at the same level as "classification", "detection", "segmentation". 
- Classification
    - Table of all available Classification weights
 ...
- Object Detection
    - Table of all available detection weights
- Keypoint detection
    - Table of all available detection weights

Otherwise we can keep them in the same group but the relative hierarchy need to change:

- Object Detection, Instance Segmentation and Person Keypoint Detection
    - Object Detection
        - Table of all available detection weights
    - Keypoint detection
        - Table of all available detection weights

docs/source/conf.py
jdsgomes: I think this should do the trick:

```python
In [59]:  re.match("^((?!Keypoint).)*$", "no_keypoint_in_name")
Out[59]: <re.Match object; span=(0, 19), match='no_keypoint_in_name'>
```

vs 

```python 
In [58]: re.match("^((?!Keypoint).)*$", "xxxKeypoint")
Out[58]: None
```

docs/source/models_new.rst
NicolasHug: Thanks for the input @jdsgomes , that sounds good.

The currently title is `Object Detection, Instance Segmentation and Person Keypoint Detection`. Do we actually have Instance Segmentation models ATM?

docs/source/conf.py
NicolasHug: Thanks @jdsgomes , that seems to work indeed. I'm a bit on the fence with this. In general, I try to avoid regex like the plague. This one typically doesn't read easily to me and would require some extra comment IMHO. 

I'll yield to whichever you prefer. Do you have a preference here?

docs/source/models_new.rst
jdsgomes: I don't think so

docs/source/conf.py
jdsgomes: since we are including/excluding simple patterns I would agree that regex here might just complicate things and just save a few lines of code, so I would say to leave as is.

jdsgomes (approved):
LGTM, thanks for making the changes!

docs/source/models_new.rst
oke-aditya: Iisn't mask rcnn instance segmentation?

NicolasHug: Thanks @oke-aditya  you're right. As you also properly noted offline, mask-rcnn has a custom  mask_map metric, so it would make sense for it to have its own table. Would you like to submit a PR to fix this?

Thinking about our layout again, I guess it could look like

```
Object Detection, Instance Segmentation and Person Keypoint Detection 
  Object Detection
     table
  Instance segmentation
     table
  Keypoint detection
     table
```

oke-aditya: yep, will send a PR :)

Issue #3485, opened 2021-03-02
add the pretrained resnext101_64x4d model


tangjiuqi097:
## 🚀 Feature

add the pretrained resnext101_64x4d model

## Motivation

In object detection, the resnext101_64x4d is a widely used backbone.
But the torchvision model only provide the resnext101_32x8d pretrained models.
If we want to use the resnext101_64x4d in object detection, it cannot be smoothly to change.

oke-aditya:
Hello. @tangjiuqi097 . The discussion for adding new models is in #2707. Feel free to comment there.
I will add this to the list.

PR #5935, opened 2022-05-03
Adding resnext101 64x4d model

YosuaMichael:
Resolve https://github.com/pytorch/vision/issues/3485

# Training and Validation script

We train the model using the following script:
```
    --timeout 3000 --ngpus 8 --nodes 1 --batch-size=128 \
    --partition train --model resnext101_64x4d \
    --data-path="/datasets01_ontap/imagenet_full_size/061417" \
    --lr=0.5 --lr-scheduler=cosineannealinglr --lr-warmup-epochs=5 --lr-warmup-method=linear \
    --auto-augment=ta_wide --epochs=600 --random-erase=0.1 --weight-decay=0.00002 \
    --norm-weight-decay=0.0 --label-smoothing=0.1 --mixup-alpha=0.2 --cutmix-alpha=1.0 \
    --train-crop-size=176 --model-ema --val-resize-size=232 --ra-sampler --ra-reps=4
```
From the training, we found the best result is on epoch 455 with the EMA model.
We validate with the 1 gpu and batch_size==1 using the following script (with final output):
```
    --timeout 3000 --nodes 1 --ngpus 1 --batch-size=1 \
    --partition train --model "resnext101_64x4d" \
    --data-path="/datasets01_ontap/imagenet_full_size/061417" \
    --weights="ResNeXt101_64X4D_Weights.IMAGENET1K_V1" \
    --test-only
# Acc@1 83.246 Acc@5 96.454
```

## Quantized model

We do quantize the model by using the following script:
```
python train_quantization.py \
    --device='cpu' --post-training-quantize --backend='fbgemm' \
    --model=resnext101_64x4d --weights="ResNeXt101_64X4D_Weights.IMAGENET1K_V1" \
    --train-crop-size 176 --val-resize-size 232 --data-path /datasets01_ontap/imagenet_full_size/061417/
```
And from the result we validate it again using 1 gpu and batch_size=1 with the following script:
```
python train_quantization.py --device='cpu' --eval-batch-size=1 --test-only --backend='fbgemm' \
    --model='resnext101_64x4d'  --data-path="/datasets01_ontap/imagenet_full_size/061417" \
    --weights="ResNeXt101_64X4D_QuantizedWeights.IMAGENET1K_FBGEMM_V1"
# Acc@1 82.898 Acc@5 96.326
```


docs/source/models.rst
```diff
@@ -481,8 +483,11 @@ a model with random weights by calling its constructor:
     resnet18 = models.quantization.resnet18()
     resnet50 = models.quantization.resnet50()
     resnext101_32x8d = models.quantization.resnext101_32x8d()
+    resnext101_64x4d = models.quantization.resnext101_64x4d()
```
datumbox, line 486: We should have a similar line on the non-quantized section of the doc. Github won't let me add a comment directly but it should be around line 63.

datumbox (approved):
@YosuaMichael LGTM, thanks! Just one minor nit, see below.

docs/source/models.rst
YosuaMichael: Ah, seems like I missed it. I usually search the location of text `resnext101` and add the new model after, but seems like we miss the `resnext101_32x8d` on the non-quantized part. The new commit would add both.

YosuaMichael:
The failing test seems to not related to this PR, see https://github.com/pytorch/vision/issues/5971

PR #5938, opened 2022-05-04
[NOMRG] Install dill to test prototype datasets when dill is installed

NicolasHug:
This PR installs `dill` for the prototype CIs.
Having `dill`  installed in one's environement has shown to create (now fixed) issues in the past https://github.com/pytorch/pytorch/pull/74984. I'm just creating this PR to make sure there are no additional problems. We shouldn't merge this PR (probably ever), but it's worth keeping it up to date from time to time.

CC @pmeier


pmeier:
> (now fixed) 

Well, the tests are failing for now, but will hopefully pass after pytorch/pytorch#74984.

@ejguan @NivekT This is our hack to make sure just having `dill` in the environment does not break our datasets. I can't find the comment now, but IIRC you wanted to go for a manual serialization backend selection. If this is implemented, we could simply parametrize our tests instead. Is this tracked anywhere?

Issue #3913, opened 2021-05-25
Fix jpeg encoding tests on windows


NicolasHug:
In https://github.com/pytorch/vision/pull/3908 we fixed the test logic of the tests for `encode_jpeg`. Unfortunately, the new "correct" tests don't work on windows.

This issue is about a) making those new correct test work on windows and b) remove the old incorrect tests

For b), having a roundtrip test as proposed in #3912 might be enough to remove the old tests.

CC @fmassa @datumbox

NicolasHug:
Update on this: From https://github.com/pytorch/vision/pull/3968 it is clear that:

- On the linux CI (and very likely on macOS), both torchvision and PIL rely on libjpeg 9.0
- On the Windows CI, torchvision relies on libjpeg 9.0 but PIL relies on libjpegturbo 8.

This is consistent with https://github.com/python-pillow/Pillow/issues/3833#issuecomment-585211263

> As mentioned before, the Windows wheels are built with libjpeg-turbo, the Linux wheels are not

This can explain why the new correct tests fail on windows: the encoding strategy is likely different between libjpeg9 and libjpegturbo 8.

Also from https://github.com/python-pillow/Pillow/issues/3833#issuecomment-803702604

> If the Pillow wheel does not have the version of libjpeg that you require, then install it and try compiling Pillow from source. Compiling on Windows became easier after the opening of this issue, thanks to #4495.

There are no Windows wheels that don't come with libjpeg-turbo as far as I can tell. Sooooo... for a) I only see 2 options:

- Start supporting libjpeg turbo so that we can properly compare our implementation on windows
- build PIL from source in the Windows CI runs

datumbox:
@NicolasHug awesome investigation skills. I think we might be in a situation where every solution stinks... 

If we do option 1, aka use libjpeg-turbo for our windows build and libjpeg 9.0 for our linux one, don't we run the risk of having different results between platforms on our TorchVision methods? Also if we do change our dependency, should we consider it a breaking change? Another thing to consider is whether the to implementation have differences that can cause more headaches on the C++ code. We already know that libjpeg has differences across versions/platforms and that forced us to pin the version to 9b (see #3787).

Building from source sounds tedious... I wonder if it makes sense to raise an issue to PIL and discuss the possibility of offering an official version on their end.

NicolasHug:
[...]

PR #5941, opened 2022-05-04
Use libjpeg-turbo in CI instead of libjpeg

NicolasHug:
This PR makes our CI rely on libjpeg-turbo insteaf of libjpeg. The main benefit of libjpeg-turbo is decoding speed, but this will also allow us to clean up our tests and make them more robust.

**Note**: This PR only concerns the CI tests, not the packaging for conda or PyPI. I'll look into that in a separate PR.

We had numerous issues and headaches in the past because of differences between PIL and torchvision and their underlying implementation (libjpeg vs libjpeg-turbo), e.g. https://github.com/pytorch/vision/issues/3913, https://github.com/pytorch/vision/pull/5910 or https://github.com/pytorch/vision/issues/5162. 

PIL has been shipped on libjpeg-turbo for a while **on Windows** https://github.com/python-pillow/Pillow/issues/3833#issuecomment-585211263, and since PIL 9, the linux and MacOS wheels are [now also linked against turbo](https://pillow.readthedocs.io/en/stable/releasenotes/9.0.0.html#switched-to-libjpeg-turbo-in-macos-and-linux-wheels).

Shipping with libjpeg-turbo instead of libjpeg will allow us to be fully aligned with PIL, and greatly simplify / cleanup our tests, which had become a bit messy (see the amount of removed code in this PR). Note that [internally](https://www.internalfb.com/code/fbsource/[ddae37f0f1bdc93133485334e1beaa5ebab51899]/fbcode/pytorch/vision/TARGETS?lines=280), we already rely on turbo.

Closes https://github.com/pytorch/vision/pull/5184
Closes https://github.com/pytorch/vision/issues/5162
Closes https://github.com/pytorch/vision/issues/3913
Fixes https://github.com/pytorch/vision/pull/5910

### DIFF
--- a/.circleci/unittest/linux/scripts/environment.yml
+++ b/.circleci/unittest/linux/scripts/environment.yml
@@ -1,6 +1,7 @@
 channels:
   - pytorch
   - defaults
+  - conda-forge
 dependencies:
   - pytest
   - pytest-cov
--- a/.circleci/unittest/linux/scripts/environment.yml
+++ b/.circleci/unittest/linux/scripts/environment.yml
@@ -8,7 +8,7 @@ dependencies:
   - pytest-mock
   - pip
   - libpng
-  - jpeg-turbo
+  - libjpeg-turbo
   - ca-certificates
   - h5py
   - pip:
--- a/.circleci/unittest/linux/scripts/environment.yml
+++ b/.circleci/unittest/linux/scripts/environment.yml
@@ -7,7 +7,7 @@ dependencies:
   - pytest-mock
   - pip
   - libpng
-  - jpeg
+  - jpeg-turbo
   - ca-certificates
   - h5py
   - pip:
--- a/test/test_image.py
+++ b/test/test_image.py
@@ -409,32 +409,6 @@ def test_encode_jpeg_errors():
         encode_jpeg(torch.empty((100, 100), dtype=torch.uint8))
 
 
-@_collect_if(cond=False)
-@pytest.mark.parametrize(
-    "img_path",
-    [pytest.param(jpeg_path, id=_get_safe_image_name(jpeg_path)) for jpeg_path in get_images(ENCODE_JPEG, ".jpg")],
-)
-def test_write_jpeg_reference(img_path, tmpdir):
-    # FIXME: Remove this eventually, see test_encode_jpeg_reference
-    data = read_file(img_path)
-    img = decode_jpeg(data)
-
-    basedir = os.path.dirname(img_path)
-    filename, _ = os.path.splitext(os.path.basename(img_path))
-    torch_jpeg = os.path.join(tmpdir, f"{filename}_torch.jpg")
-    pil_jpeg = os.path.join(basedir, "jpeg_write", f"{filename}_pil.jpg")
-
-    write_jpeg(img, torch_jpeg, quality=75)
-
-    with open(torch_jpeg, "rb") as f:
-        torch_bytes = f.read()
-
-    with open(pil_jpeg, "rb") as f:
-        pil_bytes = f.read()
-
-    assert_equal(torch_bytes, pil_bytes)
-
-
 @pytest.mark.parametrize(
     "img_path",
     [pytest.param(jpeg_path, id=_get_safe_image_name(jpeg_path)) for jpeg_path in get_images(ENCODE_JPEG, ".jpg")],
--- a/test/test_image.py
+++ b/test/test_image.py
@@ -478,8 +478,6 @@ def test_write_jpeg_reference(img_path, tmpdir):
     assert_equal(torch_bytes, pil_bytes)
 
 
-# TODO: Remove the skip. See https://github.com/pytorch/vision/issues/5162.
-@pytest.mark.skip("this test fails because PIL uses libjpeg-turbo")
 @pytest.mark.parametrize(
     "img_path",
     [pytest.param(jpeg_path, id=_get_safe_image_name(jpeg_path)) for jpeg_path in get_images(ENCODE_JPEG, ".jpg")],
@@ -498,8 +496,6 @@ def test_encode_jpeg(img_path):
         assert_equal(encoded_jpeg_torch, encoded_jpeg_pil)
 
 
-# TODO: Remove the skip. See https://github.com/pytorch/vision/issues/5162.
-@pytest.mark.skip("this test fails because PIL uses libjpeg-turbo")
 @pytest.mark.parametrize(
     "img_path",
     [pytest.param(jpeg_path, id=_get_safe_image_name(jpeg_path)) for jpeg_path in get_images(ENCODE_JPEG, ".jpg")],
--- a/test/test_image.py
+++ b/test/test_image.py
@@ -94,10 +94,14 @@ def test_decode_jpeg(img_path, pil_mode, mode):
     data = read_file(img_path)
     img_ljpeg = decode_image(data, mode=mode)
 
-    # Permit a small variation on pixel values to account for differences
-    # between Pillow's libjpeg and ours.
-    abs_mean_diff = (img_ljpeg.type(torch.float32) - img_pil).abs().mean().item()
-    assert abs_mean_diff < 1
+    # In theory PIL and torchvision rely on the same libjpeg-turbo, so decoding
+    # should be exactly the same. There seem to be small differences with some
+    # GRAY mode images though. Might be worth investigating why.
+    if mode == ImageReadMode.GRAY:
+        abs_mean_diff = (img_ljpeg.type(torch.float32) - img_pil).abs().mean().item()
+        assert abs_mean_diff < 1
+    else:
+        assert_equal(img_ljpeg, img_pil)
 
 
 def test_decode_jpeg_errors():
--- a/test/test_image.py
+++ b/test/test_image.py
@@ -94,10 +94,10 @@ def test_decode_jpeg(img_path, pil_mode, mode):
     data = read_file(img_path)
     img_ljpeg = decode_image(data, mode=mode)
 
-    # Permit a small variation on pixel values to account for implementation
-    # differences between Pillow and LibJPEG.
+    # Permit a small variation on pixel values to account for differences
+    # between Pillow's libjpeg and ours.
     abs_mean_diff = (img_ljpeg.type(torch.float32) - img_pil).abs().mean().item()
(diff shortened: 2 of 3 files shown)


NicolasHug:
yup

NicolasHug:
CI looks happy so far. @andfoy @fmassa do you remember by any chance why we didn't rely on libjpeg-turbo from the start? Perhaps it's because at the time, PIL was not linked against libjpeg-turbo?

test/test_image.py
```diff
@@ -94,10 +94,10 @@ def test_decode_jpeg(img_path, pil_mode, mode):
     data = read_file(img_path)
     img_ljpeg = decode_image(data, mode=mode)
 
-    # Permit a small variation on pixel values to account for implementation
-    # differences between Pillow and LibJPEG.
+    # Permit a small variation on pixel values to account for differences
+    # between Pillow's libjpeg and ours.
```
pmeier, line 98: Should this be

```suggestion
    # between Pillow's libjpeg-tubro and ours.
```

?

pmeier (approved):
Since CI seems to be happy, I think this is good to go. Thanks Nicolas!

.circleci/unittest/linux/scripts/environment.yml
```diff
@@ -1,13 +1,14 @@
 channels:
   - pytorch
   - defaults
+  - conda-forge
```
datumbox, line 4: cc @malfet as previously you've tried to phase out conda-forge from TorchVision.

datumbox (approved):
LGTM, thanks!

test/test_image.py
```diff
@@ -94,10 +94,14 @@ def test_decode_jpeg(img_path, pil_mode, mode):
...
-    # Permit a small variation on pixel values to account for implementation
-    # differences between Pillow and LibJPEG.
-    abs_mean_diff = (img_ljpeg.type(torch.float32) - img_pil).abs().mean().item()
-    assert abs_mean_diff < 2
+    # In theory PIL and torchvision rely on the same libjpeg-turbo, so decoding
+    # should be exactly the same. There seem to be small differences with some
+    # GRAY mode images though. Might be worth investigating why.
+    if mode == ImageReadMode.GRAY:
+        abs_mean_diff = (img_ljpeg.type(torch.float32) - img_pil).abs().mean().item()
+        assert abs_mean_diff < 1
```
NicolasHug, line 102: Not too sure why this happens TBH. But this is *not* a regression, in fact we're now making this check much stricter than it previously was.

.circleci/unittest/linux/scripts/environment.yml
malfet: There are two issues I'm slightly concerned about:
  - Would it mean that we are testing something different then what we ship?
  - This might unintentionally pull in different dependencies, that we think exist in conda, but they are not.

If this is only needed for testing, we can build and host `libjpeg-turbo` for all the platforms we care about in `pytorch` or `pytorch-nightly` channels

NicolasHug: Thanks for your feedback Nikita

> Would it mean that we are testing something different then what we ship?

Yes, but only until https://github.com/pytorch/vision/pull/5951 is merged. https://github.com/pytorch/vision/pull/5951 's goal is to ship torchvision with libjpeg turbo.

It's also not as bad as it sounds: it's fair at this point to assume that libjpeg-turbo's `libjpeg.h` is compatible with that of libjpeg. In terms of decoding results, they're both jpeg-compliant as well.

> If this is only needed for testing, we can build and host libjpeg-turbo for all the platforms we care about in pytorch or pytorch-nightly channels

Ultimately we don't want this just for testing: we also want to ship torchvision with libjpeg in https://github.com/pytorch/vision/pull/5951 . But if we're happy to host libjpeg-turbo in the `pytorch` channel, then this might make all this much easier. WDYT @malfet

datumbox:
@NicolasHug I ran a few benchmark checks to see what would be the effect of this change. I do factor in the argument that PIL has changed the backend on 9.x, but I think we should measure carefully the effect on existing pre-trained models, the speed improvements that we get by switching and potential alternative approaches. This is one of these changes that are trivial to make on the code side but might have effects that need to be investigated.

## Accuracy Benchmarks

First some good news. The effect on the models is detectable but very small. Here is how the model accuracies change with JPEG and JPEG-turbo:
```
ResNet50_Weights.IMAGENET1K_V1:
JPEG: Acc@1 76.130 Acc@5 92.864
JPEG-turbo: Acc@1 76.148 Acc@5 92.876

ResNet50_Weights.IMAGENET1K_V2:
JPEG: Acc@1 80.854 Acc@5 95.434
JPEG-turbo: Acc@1 80.844 Acc@5 95.436

MobileNet_V3_Large_Weights.IMAGENET1K_V1:
JPEG: Acc@1 74.044 Acc@5 91.322
JPEG-turbo: Acc@1 74.056 Acc@5 91.318
```

The above were executed on a single gpu and batch=1 to minimize variations. From the tests that your PR is removing, we know that there are differences between the two implementations (noted also on PIL release notes you reference) but in practice it's mostly noise. It would be worth to check other model families such as Object Detection, Semantic Segmentation and Optical Flow to see if any of them is more sensitive to the change. If the differences continue being so small and the speed gains are significant, I think there is a compelling case to introduce this minor BC-breaking change. Perhaps another argument for doing so is the stability we will get across platforms on our unit-tests and the ability to compare easier the results of PIL vs TorchVision on the reading of images.

## Alternative approaches

There are alternative routes we can take, some of which are temporary. None of them are "free" and we should consider them only if there is massive effect on the accuracy of the existing models:

1. We can temporarily patch the [`pil_loader()`](https://github.com/pytorch/vision/blob/6ca9c76adb6daf2695d603ad623a9cf1c4f4806f/torchvision/datasets/folder.py#L244-L248) method on our Datasets (until Datasets V2 is stable), to remove PIL's jpeg-decoding from the equation and use TorchVision's `read_image()` which relies on LibJPEG. Here is a quick and dirty patch for this that I tested and works as expected, giving identical accuracy.
```patch
index 40d5e26d..982dd2ba 100644
--- a/torchvision/datasets/folder.py
+++ b/torchvision/datasets/folder.py
@@ -243,9 +243,15 @@ IMG_EXTENSIONS = (".jpg", ".jpeg", ".png", ".ppm", ".bmp", ".pgm", ".tif", ".tif
 
 def pil_loader(path: str) -> Image.Image:
     # open path as file to avoid ResourceWarning (https://github.com/python-pillow/Pillow/issues/835)
+    """
     with open(path, "rb") as f:
         img = Image.open(f)
         return img.convert("RGB")
+    """
+    from torchvision.io.image import read_image
+    from torchvision.transforms.functional import to_pil_image
+    img = read_image(path)
+    return to_pil_image(img).convert("RGB")
```
It's worth noting that this method is not used by most of the datasets right now (they opt for doing `Image.open(f).convert("RGB")` directly) but in this scenario we could easily change that without BC-breaking problems. This only partially mitigates the issue as users who read images directly with PIL (outside of our datasets) would still be affected. At least this gives us an option to ensure that TorchVision doesn't break BC on the components it controls.

2. Pin PIL temporarily to a 8.x version and reach out to them and discuss options (potentially offer a way to switch backends). We can investigate options to address dependency issues that will be caused by this.

3. Do nothing to mitigate the issue on PIL, flag the problem to our users and push them to use the Tensor Transforms backend while ensuring that the accuracy will be identical if they switch to it. This might mean that we would have to aggressively align Tensor transforms with PIL in any difference (for example antialiasing=True) which is also an issue of its own and needs to be discussed.


I hope that none of the above will be necessary and that the difference on accuracy will remain minor, allowing us to get away with this minor BC breakage. I would also be OK to merge a PR like this immediately after the upcoming release to give a lot of time to users to detect and flag potential issues. It's highly likely we would need to rerun inference jobs for all models to correct the meta-data and documentation, so this needs to be factored in on the amount of work that this switch entails. It might be worth starting a project doc or an RFC on Github to record all these (instead of having these discussions spread across PRs and issues) to keep track of what we need to do. Happy to chat more on this.

NicolasHug:
Thanks a lot for the benchmarks @datumbox . Just sharing initial thoughts below:

>  Here is how the model accuracies change with JPEG and JPEG-turbo

Could you share the exact setup you used to compare libjpeg and libjpeg-turbo? Did you rely on `decode_jpeg()`, or did you instead compare PIL8 vs PIL9? I'm wondering if PIL8 vs PIL9 may flag some differences that come e.g. from a difference in transforms results, on top of the different decoders. I guess the transforms should always give the same results across PIL versions, but who knows?

> We can temporarily patch the [pil_loader()](https://github.com/pytorch/vision/blob/6ca9c76adb6daf2695d603ad623a9cf1c4f4806f/torchvision/datasets/folder.py#L244-L248) method on our Datasets (until Datasets V2 is stable), to remove PIL's jpeg-decoding from the equation and use TorchVision's read_image() which relies on LibJPEG

Unfortunately, our `read_image()` (and particularly `decode_jpeg()`) isn't as complete as PIL's decoders. For example we currently don't support decoding CMYK -> RGB jpegs, which makes it impossible to decode some of the ImageNet samples https://github.com/pytorch/vision/issues/6538 with `read_image()`

datumbox:
@NicolasHug 

> Could you share the exact setup you used to compare libjpeg and libjpeg-turbo?

I tried to isolate the effects of PIL versions as follows:
1. I used the patch to completely circumvent PIL and linked against different JPEG backends. So differences on PIL versions shouldn't factor in here.
2. I compared models (MobileNetV3) for which I know the full training history and was done prior the PIL 9 changes. The accuracy reported on our release notes and meta-data matches what I get. So this way I know that the old PIL backend aligns extremely close with our IO. Worth noting that we had flaky tests to check this, so historically prior the backend change, we ensured we were aligned.

From the above, I'm fairly confident that our IO `read_image()` closely aligns with PIL v8 and that my benchmarks isolate only the effect of JPEG vs JPEG-turbo. If you spot anything weird in my logic let me know.

> don't support decoding CMYK -> RGB jpegs

That's a good call out. There are probably no CMYK images on the validation split and that's why I didn't get an error during inference. I think, CMYK is something we should look into doing eventually; I had put a TODO when I worked on the `read_image()` API with some references and it didn't look like too hard at the time but we didn't have the time to do it. This might mean that if we get a substantial accuracy difference in any of the models and we have to go with the 1st alternative option proposed above, we would have to do more work.

PR #5943, opened 2022-05-04
Remove jpeg dep for type-checking

NicolasHug:
Not sure why this was ever needed, let's see if CI goes red(er)


pmeier (approved):
My guess is that the very first version of this workflow mirrored a different one that installed this. Anyway, happy to remove if it is no longer needed. Thanks Nicolas!

Issue #5893, opened 2022-04-27
Mac M1 pro segmentation fault on import


axmav:
### 🐛 Describe the bug

Hello! Can not import torchvision on my mac. Get an error:
`libc++abi: terminating with uncaught exception of type std::length_error: vector`
as well as segmentation fault.

code:
```
import torch
from torch import nn
from torch.nn import functional as F
from torch.distributions.normal import Normal
from torch.distributions.kl import kl_divergence
from torchvision.utils import save_image
```

it also fails just with `from torchvision.utils import save_image`

Similar issue: https://github.com/pyg-team/pytorch_geometric/issues/4419
But not a single proposed solution works...

### Versions

PyTorch version: 1.10.2
Is debug build: False
CUDA used to build PyTorch: None
ROCM used to build PyTorch: N/A

OS: macOS 12.1 (arm64)
GCC version: Could not collect
Clang version: 13.1.6 (clang-1316.0.21.2.3)
CMake version: Could not collect
Libc version: N/A

Python version: 3.9.10 | packaged by conda-forge | (main, Feb  1 2022, 21:25:34)  [Clang 11.1.0 ] (64-bit runtime)
Python platform: macOS-12.1-arm64-arm-64bit
Is CUDA available: False
CUDA runtime version: No CUDA
GPU models and configuration: No CUDA
Nvidia driver version: No CUDA
cuDNN version: No CUDA
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

Versions of relevant libraries:
[pip3] numpy==1.21.5
[pip3] torch==1.10.2
[pip3] torchaudio==0.11.0
[pip3] torchvision==0.12.0
[conda] numpy                     1.21.5           py39h47fb9ce_0    conda-forge
[conda] torch                     1.10.2                   pypi_0    pypi
[conda] torchaudio                0.11.0                   pypi_0    pypi
[conda] torchvision               0.12.0                   pypi_0    pypi

datumbox:
TorchVision doesn't properly provide support for this at the moment. We have a todo to investigate at #5171 but we don't currently have the bandwidth to resolve.

We'll definitely need to fix once when we investigate. Also by briefly checking the linked issue but I'm not sure what exactly it's invoking the `libc++abi` when you simply import `from torchvision.utils import save_image`. I wonder if that's linked to the `@torch.no_grad()` decorator. Could you test this out?

axmav:
[...]

Issue #5413, opened 2022-02-12
image.so on Apple Silicon can't load libpng16.16.dylib and libjpeg.9.dylib


alanyoungCN:
### 🐛 Describe the bug

code:
```python
import torchvision
```
UserWarning:
```
/Users/alanyoung/Documents/Codes/Kaggle/ClassifyLeaves/venv/lib/python3.9/site-packages/torchvision/io/image.py:11: UserWarning: Failed to load image Python extension: dlopen(/Users/alanyoung/Documents/Codes/Kaggle/ClassifyLeaves/venv/lib/python3.9/site-packages/torchvision/image.so, 0x0006): Library not loaded: @rpath/libpng16.16.dylib
  Referenced from: /Users/alanyoung/Documents/Codes/Kaggle/ClassifyLeaves/venv/lib/python3.9/site-packages/torchvision/image.so
  Reason: tried: '/Users/malfet/miniforge3/envs/py_39_torch-1.10.2/lib/libpng16.16.dylib' (no such file), '/Users/malfet/miniforge3/envs/py_39_torch-1.10.2/lib/libpng16.16.dylib' (no such file), '/Users/malfet/miniforge3/envs/py_39_torch-1.10.2/lib/libpng16.16.dylib' (no such file), '/Users/malfet/miniforge3/envs/py_39_torch-1.10.2/lib/libpng16.16.dylib' (no such file), '/opt/homebrew/Caskroom/miniforge/base/lib/libpng16.16.dylib' (no such file), '/opt/homebrew/Caskroom/miniforge/base/lib/libpng16.16.dylib' (no such file), '/opt/homebrew/Caskroom/miniforge/base/lib/python3.9/lib-dynload/../../libpng16.16.dylib' (no such file), '/opt/homebrew/Caskroom/miniforge/base/lib/libpng16.16.dylib' (no such file), '/opt/homebrew/Caskroom/miniforge/base/bin/../lib/libpng16.16.dylib' (no such file), '/usr/local/lib/libpng16.16.dylib' (no such file), '/usr/lib/libpng16.16.dylib' (no such file)
  warn(f"Failed to load image Python extension: {e}")
```
on M1 Max.
In https://github.com/pytorch/vision/issues/5137#issuecomment-1023708994, it is said to be fixed but I don't think so.

Therefore you can't use torchvision.io.read_img().
```python
import torchvision
img = torchvision.io.read_img('./1.jpg')
```

```
RuntimeError: No such operator image::read_file
```
[...]

PR #5948, opened 2022-05-05
Add TorchVision wheel builds for M1

malfet:

### DIFF
--- a/.github/workflows/build-m1-binaries.yml
+++ b/.github/workflows/build-m1-binaries.yml
@@ -1,0 +1,48 @@
+name: Build on M1
+on:
+  pull_request:
+    paths:
+      - .github/workflows/build-m1-binaries.yml
+  push:
+    branches:
+      - nightly
+  workflow_dispatch:
+jobs:
+  build_wheels:
+    name: "Build TorchVision M1 wheels"
+    runs-on: macos-m1
+    strategy:
+      matrix:
+        py_vers: [ "3.8", "3.9", "3.10" ]
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Build TorchVision M1 wheel
+        shell: arch -arch arm64 bash {0}
+        env:
+          ENV_NAME: conda-env-${{ github.run_id }}
+          PY_VERS: ${{ matrix.py_vers }}
+        run: |
+          echo $PATH
+          . ~/miniconda3/etc/profile.d/conda.sh
+          set -ex
+          conda create -yp ${ENV_NAME} python=${PY_VERS} numpy libpng openjpeg wheel pkg-config
+          conda run -p ${ENV_NAME} python3 -mpip install torch --extra-index-url=https://download.pytorch.org/whl/nightly torch
+          conda run -p ${ENV_NAME} python3 -mpip install delocate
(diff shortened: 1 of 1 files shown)


datumbox:
Also the [linter](https://app.circleci.com/pipelines/github/pytorch/vision/17813/workflows/3d7c8f18-c961-4d71-975e-dfeb5b238dd4/jobs/1442508) is complaining for a trailing new line that is missing.

.github/workflows/build-m1-binaries.yml
```diff
@@ -0,0 +1,48 @@
...
+    paths:
+      - .github/workflows/build-m1-binaries.yml
+  push:
+    branches:
+      - nightly
+  workflow_dispatch:
+jobs:
+  build_wheels:
+    name: "Build TorchVision M1 wheels"
+    runs-on: macos-m1
```
atalman, line 13: We have only 1 macos-m1 worker currently. I will change the rest of our workers to adhere to the same macos-m1 label

NicolasHug:
@malfet @atalman 

as @datumbox noted above https://github.com/pytorch/vision/pull/5948#issuecomment-1134505355, we're not certain that torchvision properly works on M1. In order to avoid additional bug reports from users, are you planning to add an M1 CI so we're sure the binaries are working properly?

malfet:
> In order to avoid additional bug reports from users, are you planning to add an M1 CI so we're sure the binaries are working properly?

@NicolasHug we don't have a capacity to run PR CI on M1 right now (there are 4 machines in total for the entire org), but if you can pick a few smoketests we can try running them.

datumbox:
@malfet Our users are currently reporting that multiple parts of TorchVision are not working on M1: https://github.com/pytorch/vision/issues/6067, https://github.com/pytorch/vision/issues/5893, https://github.com/pytorch/vision/issues/5413

If that's true, it will take more than a couple of smoketests to debug, fix and ensure it works smoothly. If we currently don't have the resources for adding proper support for M1 and testing them across PyTorch, we should revert this PR and consider adding it back once we have the necessary capacity. Otherwise, after release we will have a huge volume of users, reporting that TorchVision is broken on M1 with maintainers unable to know what to do next and how to address this. We are happy to discuss a middle ground solution that allows us to run our tests in one configuration on M1. Please let me know.

malfet:
@datumbox thank you for sharing. All those sound like a basic packaging issues to me, so I'm not sure how CI testing (where system is configured to have all above mentioned dependencies will help) And even if PR CI would be enabled, those machines would not have rerun with SSH capability.

But we should definitely test for the conditions mentioned in above during packaging.

datumbox:
I had an offline discussion with @malfet concerning this. I'll summarize here what was discussed:

Our team doesn't have an M1 mac locally but even if we did, we still need to test each PR prior merging on at least one M1/Python configuration. Nikita agreed that since TorchVision is the first to be onboarded on native M1, we can do that now. Nevertheless because PyTorch has only 4 machines, it is not guaranteed that the job will be maintained. To avoid having the job removed, TorchVision should dedicate time to speed up CI execution for its tests.

I will therefore use the approach on this PR to add a unit-test job for a single python configuration. See #6111

PR #5949, opened 2022-05-05
Add mask rcnn to instance segmentation docs

oke-aditya:
Fixes as discussed offline and https://github.com/pytorch/vision/pull/5933#discussion_r865031350

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -348,14 +348,25 @@ def inject_weight_metadata(app, what, name, obj, options, lines):
             lines.append("")
 
 
-def generate_weights_table(module, table_name, metrics, include_pattern=None, exclude_pattern=None):
+def generate_weights_table(module, table_name, metrics, include_patterns=None, exclude_patterns=None):
+    """
+
+    :param module: corresponding torchvision module
+    :param table_name: Name of table generated.
+    :param metrics: Metrics from weights enum to be used.
+    :param include_patterns: List of patterns to include
+    :param exclude_patterns: List of patterns to exclude
+
+    Generates table file and saves it to generated folder.
+
+    """
     weight_enums = [getattr(module, name) for name in dir(module) if name.endswith("_Weights")]
     weights = [w for weight_enum in weight_enums for w in weight_enum]
 
-    if include_pattern is not None:
-        weights = [w for w in weights if include_pattern in str(w)]
-    if exclude_pattern is not None:
-        weights = [w for w in weights if exclude_pattern not in str(w)]
+    if include_patterns is not None:
+        weights = [w for w in weights if any(p in str(w) for p in include_patterns)]
+    if exclude_patterns is not None:
+        weights = [w for w in weights if all(p not in str(w) for p in exclude_patterns)]
 
     metrics_keys, metrics_names = zip(*metrics)
     column_names = ["Weight"] + list(metrics_names) + ["Params", "Recipe"]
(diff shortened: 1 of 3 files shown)


docs/source/conf.py
```diff
@@ -385,6 +385,12 @@ def generate_weights_table(module, table_name, metrics, include_pattern=None, ex
 generate_weights_table(
     module=M.detection, table_name="detection", metrics=[("box_map", "Box MAP")], exclude_pattern="Keypoint"
 )
+generate_weights_table(
+    module=M.detection,
+    table_name="instance_segmentation",
+    metrics=[("box_map", "Box MAP"), ("mask_map", "Mask MAP")],
```
oke-aditya, line 391: IMO mean average precision is short handed as mAP and not MAP?  That's what I think pycocotools or other report?

docs/source/conf.py
```diff
@@ -385,6 +385,12 @@ def generate_weights_table(module, table_name, metrics, include_pattern=None, ex
...
+    include_pattern="Mask",
```
oke-aditya, line 392: This seems hack way to add docs by `include_patterns` and `exclude_pattern`. Not sure if we could come up with better way?

docs/source/conf.py
```diff
@@ -385,6 +385,12 @@ def generate_weights_table(module, table_name, metrics, include_pattern=None, ex
 generate_weights_table(
     module=M.detection, table_name="detection", metrics=[("box_map", "Box MAP")], exclude_pattern="Keypoint"
```
oke-aditya, line 386: I now need to ignore 2 exclude_pattern "Mask" and "Keypoint". This will need additional refactor

docs/source/conf.py
NicolasHug: I've seen both used. No strong opinion from me

docs/source/conf.py
NicolasHug: TBF, this entire doc generation strategy is a big hack.

We discussed alternatives in https://github.com/pytorch/vision/pull/5933#discussion_r863857591. Regexs are an alternative but they're potentially more complex, especially now that we need to ignore more than one pattern.

Ultimately if we have a registration mechanism for all these models and weights, we might be able to remove this.

docs/source/conf.py
NicolasHug: yeah, l guess we should change `accept_pattern` and `exclude_pattern` to take lists instead:

```py
    if include_patterns is not None:
        weights = [w for w in weights if any(p in str(w) for p in include_patterns)]
    if exclude_patterns is not None:
        weights = [w for w in weights if all(p not in str(w) for p in exclude_patterns)]
```

(and add an `s` to their name)

NicolasHug (commented):
Thanks @oke-aditya ! Looks great so far

docs/source/conf.py
oke-aditya: I wish I know these python tricks. I wrote loops and logic to do same. (too much java spoils python skills)
Anyways if some day we revisit 
```
if exclude_pattern is not None:
        weights_l = []
        for w in weights:
            fg = 0
            for e in exclude_pattern:
                if e in str(w):
                    fg = 1
                    break
            if(fg == 0):
                weights_l.append(w)
        weights = weights_l
```

i will refactor as your comment :)

docs/source/conf.py
oke-aditya: Yeah I'm too not in support of regex. Its like virus.

docs/source/conf.py
NicolasHug: > too much java spoils python skills

😅

Years ago, this thread was quite useful to me to up my Python game: https://stackoverflow.com/questions/2573135/python-progression-path-from-apprentice-to-guru

I think it's still relevant today (some of them are not, typically `map`, `reduce`, `filter` are mostly replaced by list comprehension or other utilities like `sum` or `any`, but they're staples of functional programming that are worth knowing. Same with `xrange` which is irrelevant in Python3).

oke-aditya: Thank you for the link :) Will read and learn more :rocket:

oke-aditya:
Rendered docs are here
https://output.circle-artifacts.com/output/job/16feb05f-f12f-4bd6-90ee-126f469149a1/artifacts/0/docs/models_new.html#

I guess it's good to merge

docs/source/conf.py
```diff
@@ -348,14 +348,25 @@ def inject_weight_metadata(app, what, name, obj, options, lines):
...
+
+    module: corresponding torchvision module
+    table_name: Name of table generated.
+    metrics: Metrics from weights enum to be used.
+    include_patterns: List of patterns to include
+    exclude_patterns: List of patterns to exclude
+
+    Generates table file and saves it to generated folder.
+
+    """
```
NicolasHug, line 362: I would suggest to  remove this docstring, it doesn't add much additional info to the parameter names. A good description would require a much longer docstring, but I don't feel like it's worth doing this right now.

NicolasHug (commented):
Thanks a lot @oke-aditya !

Because there are only one model in the `Instance Seg` and `Keypoint` sections, I wonder if we shouldn't keep them as subsections of a more general "Object Detection, Instance Segmentation and Person Keypoint Detection " section as suggested in https://github.com/pytorch/vision/pull/5933#discussion_r865031350 ?

docs/source/conf.py
oke-aditya: Removed, and changed the subsection. I hope the new layout is what you asked :)

NicolasHug (approved):
Thanks @oke-aditya , I didn't do a great job at describing what I had in mind, so I pushed directly instead. Docs look like this now:

image

Nothing is set in stone, we can revisit if we feel like it:) 

Thanks a lot for the PR !

PR #5868, opened 2022-04-22
Pin setuptools to 58.0.4 on Widnows

malfet:
`setuptools` were update on anaconda to [61.2.0](https://anaconda.org/anaconda/setuptools/files?version=61.2.0) 10 days ago, but CircleCI environment caching occluded the issue for a while, but when update happened Win+Python-3.7 build started to fail, which suggests that it might have been caused by https://github.com/pypa/setuptools/issues/3219

Fixes https://github.com/pytorch/vision/issues/5830

PR #5950, opened 2022-05-05
remove explicit install of Pillow throughout CI

pmeier:
While looking into unifying our dependencies I wondered why we are installing PIL explicitly in multiple places throughout the CI, although installing `torchvision` should pull it in automatically. We aren't doing that with the other hard dependencies such as `numpy` or `requests`.

Turns out, we don't need to do that. After this PR, the only mentions of the `Pillow` version are

https://github.com/pytorch/vision/blob/b0dbbd7a6439f2248e194f96ddb58d427670ca56/setup.py#L67-L70

and

https://github.com/pytorch/vision/blob/b0dbbd7a6439f2248e194f96ddb58d427670ca56/packaging/torchvision/meta.yaml#L31

Given that these are "package blueprints" there is not much room of improvement. `setup.py` is executed, so we could parse the version from the conda recipe. But that feels a little to much IMO.


.circleci/unittest/windows/scripts/environment.yml
```diff
@@ -10,10 +10,9 @@ dependencies:
   - jpeg
   - ca-certificates
   - hdf5
-  - setuptools == 58.0.4
+  - setuptools
```
pmeier, line 13: This is a refinement as proposed in https://github.com/pytorch/vision/pull/5868#issuecomment-1107881373. See the next comment for details.

.circleci/unittest/windows/scripts/setup_env.sh
```diff
@@ -37,3 +37,9 @@ conda activate "${env_dir}"
 # 3. Install Conda dependencies
 printf "* Installing dependencies (except PyTorch)\n"
 conda env update --file "${this_dir}/environment.yml" --prune
+
+# 4. Downgrade setuptools on Python 3.7.
+#    See https://github.com/pytorch/vision/pull/5868
+if [[ "${PYTHON_VERSION}" == '3.7' ]]; then
+  pip install --upgrade setuptools==58.0.4
+fi
```
pmeier, line 45: Initially, `setuptools` and similar packages used [`cp3_10`](https://github.com/pypa/setuptools/blob/4fa006da4bbced755463b99914029c05c61523da/setuptools/_vendor/packaging/tags.py#L723-L729) as the interpreter tag for Python 3.10. After [PEP 641](https://peps.python.org/pep-0641/) was rejected, they changed to `cp310`. This patch was included in [`setuptools==59.1.0`](https://github.com/pypa/setuptools/blob/c5a4c42d38cd0ba5ecd34977e211ae4b6865afec/setuptools/_vendor/packaging/tags.py#L466-L467). With this tag set incorrectly, `python setup.py develop` ignores wheels on Python 3.10 and tries to use the source version. That is fine for pure python libraries, but means that it will try to compile Pillow from source.

Unfortunately, the `defaults` channel provides nothing between `setuptools==58.0.4` and `setuptools==61.2.0`. The former has the problems described above and the latter has problems on Python 3.7 (see #5868). Thus, I opted to use the latest release in general and special case Python 3.7 here. When we hard-depend on Python 3.8, we can simply remove this.

datumbox (approved):
LGTM and the tests also pass (kicked them again to resolve the mypy problem).

@NicolasHug Any concerns? I think keeping the pil versions in 2 places is suboptimal but better than the previous approach, so I think it's worth merging this.

PR #5951, opened 2022-05-05
[WIP] Package torchvision with libjpeg-turbo instead of libjpeg

NicolasHug:
Follow up to https://github.com/pytorch/vision/pull/5941


packaging/torchvision/meta.yaml
```diff
@@ -10,7 +10,7 @@ requirements:
   build:
     - {{ compiler('c') }} # [win]
     - libpng
-    - jpeg
+    - libjpeg-turbo
```
NicolasHug, line 13: @malfet @seemethere ,
I'm trying to package torchvision against libjpeg-turbo instead of libjpeg.

libjpeg-turbo is available on conda-forge [for all platforms](https://anaconda.org/conda-forge/libjpeg-turbo), but on the default `anaconda` channel it is [only available](https://anaconda.org/anaconda/libjpeg-turbo) on Win and OSX, not Linux.

This causes the `binary_linux_conda` jobs to [fail](https://app.circleci.com/pipelines/github/pytorch/vision/17335/workflows/e028f369-462a-4818-944b-acae268c70c2/jobs/1403140) (the rest are fine).

Would you know how we could rely on `conda-forge` here?

NicolasHug: I think I fixed that in f81da49

packaging/pkg_helpers.bash
```diff
@@ -157,7 +157,9 @@ setup_wheel_python() {
     conda create ${CONDA_CHANNEL_FLAGS} -yn "env$PYTHON_VERSION" python="$PYTHON_VERSION"
     conda activate "env$PYTHON_VERSION"
     # Install libpng from Anaconda (defaults)
-    conda install ${CONDA_CHANNEL_FLAGS} libpng "jpeg<=9b" -y
```
datumbox, line 160: Is this script used? The `jpeg<=9b` is definitely stale as we no longer pin the library.

NicolasHug:
https://github.com/pytorch/vision/pull/7820

Issue #5787, opened 2022-04-07
Using Kinetics400 from references raises a deprecation warning


datumbox:
### 🐛 Describe the bug

Using Kinetics400 from references leads to a deprecation warning. The fastest way to reproduce:
```
torchrun --nproc_per_node=8 train.py --data-path /datasets01/kinetics/070618/ --train-dir=val_avi-480p --val-dir=val_avi-480p --batch-size=64 --sync-bn --test-only --weights R2Plus1D_18_Weights.DEFAULT --cache-dataset

./vision/torchvision/io/video.py:160: UserWarning: The pts_unit 'pts' gives wrong results and will be removed in a follow-up version. Please use pts_unit 'sec'.
  warnings.warn(
```

It seems this is due to the following line:
https://github.com/pytorch/vision/blob/b7c59a086f4e45b8e5e799084e97cca775977a3b/torchvision/datasets/video_utils.py#L327

Which calls the method with the default parameter for `pts_unit="pts"` which is deprecated:
https://github.com/pytorch/vision/blob/b7c59a086f4e45b8e5e799084e97cca775977a3b/torchvision/io/video.py#L155-L163

If the specific option is deprecated, TorchVision shouldn't be using the feature.

### Versions

Latest main branch

pmeier:
Plus, The `Kinetics400` class is also deprecated:

https://github.com/pytorch/vision/blob/1ac6e8b91b980b052324f77828a5ef4a6715dd66/torchvision/datasets/kinetics.py#L308-L311

datumbox:
@pmeier Good point. It does. I just missed this because the other warning is spammed 100s of times at the beginning:
```
UserWarning: The Kinetics400 class is deprecated since 0.12 and will be removed in 0.14.Please use Kinetics(..., num_classes='400') instead.
```

So we should does this mean that we need to update the reference scripts not to use Kinetics400 then?

bjuncek:
So this entails of two things:

- [ ] fix dataset creation

- [ ] adapt videoclips object to avoid warning

Now, since we've had to revert the change because of which we introduced 2nd (see #5489), I'll simply remove the warning for now, and document this in #5720 for further action

PR #3680, opened 2021-04-16
Download and Kinetics 400/600/700 Datasets

bjuncek:
Addresses #3452.

PR #5952, opened 2022-05-05
Addressing  deprecation wranings in Kinetics (#5787)

bjuncek:
See #5787

### DIFF
--- a/references/video_classification/presets.py
+++ b/references/video_classification/presets.py
@@ -1,6 +1,6 @@
 import torch
 from torchvision.transforms import transforms
-from transforms import ConvertBHWCtoBCHW, ConvertBCHWtoCBHW
+from transforms import ConvertBCHWtoCBHW
 
 
 class VideoClassificationPresetTrain:
@@ -14,7 +14,6 @@ class VideoClassificationPresetTrain:
         hflip_prob=0.5,
     ):
         trans = [
-            ConvertBHWCtoBCHW(),
             transforms.ConvertImageDtype(torch.float32),
             transforms.Resize(resize_size),
         ]
@@ -31,7 +30,6 @@ class VideoClassificationPresetEval:
     def __init__(self, *, crop_size, resize_size, mean=(0.43216, 0.394666, 0.37645), std=(0.22803, 0.22145, 0.216989)):
         self.transforms = transforms.Compose(
             [
-                ConvertBHWCtoBCHW(),
                 transforms.ConvertImageDtype(torch.float32),
                 transforms.Resize(resize_size),
                 transforms.Normalize(mean=mean, std=std),
--- a/references/video_classification/train.py
+++ b/references/video_classification/train.py
@@ -145,9 +145,11 @@ def main(args):
     else:
         if args.distributed:
             print("It is recommended to pre-compute the dataset cache on a single-gpu first, as it will be faster")
-        dataset = torchvision.datasets.Kinetics400(
-            traindir,
+        dataset = torchvision.datasets.Kinetics(
+            args.data_path,
             frames_per_clip=args.clip_len,
+            num_classes="400",
+            split="val",
             step_between_clips=1,
             transform=transform_train,
             frame_rate=15,
@@ -179,9 +181,11 @@ def main(args):
     else:
         if args.distributed:
             print("It is recommended to pre-compute the dataset cache on a single-gpu first, as it will be faster")
-        dataset_test = torchvision.datasets.Kinetics400(
-            valdir,
+        dataset_test = torchvision.datasets.Kinetics(
+            args.data_path,
             frames_per_clip=args.clip_len,
+            num_classes="400",
+            split="val",
             step_between_clips=1,
             transform=transform_test,
             frame_rate=15,
--- a/references/video_classification/train.py
+++ b/references/video_classification/train.py
@@ -11,7 +11,7 @@ import torchvision.datasets.video_utils
 import utils
 from torch import nn
 from torch.utils.data.dataloader import default_collate
-from torchvision.datasets.samplers import DistributedSampler, UniformClipSampler, RandomClipSampler
+from torchvision.datasets.samplers import DistributedSampler, RandomClipSampler, UniformClipSampler
 
 
 def train_one_epoch(model, criterion, optimizer, lr_scheduler, data_loader, device, epoch, print_freq, scaler=None):
@@ -148,7 +148,7 @@ def main(args):
         dataset = torchvision.datasets.Kinetics(
             args.data_path,
             frames_per_clip=args.clip_len,
-            num_classes="400",
+            num_classes=args.kinetics_version,
             split="val",
             step_between_clips=1,
             transform=transform_train,
@@ -184,7 +184,7 @@ def main(args):
         dataset_test = torchvision.datasets.Kinetics(
             args.data_path,
             frames_per_clip=args.clip_len,
-            num_classes="400",
+            num_classes=args.kinetics_version,
             split="val",
             step_between_clips=1,
             transform=transform_test,
@@ -318,6 +318,9 @@ def parse_args():
     parser.add_argument("--data-path", default="/datasets01_101/kinetics/070618/", type=str, help="dataset path")
     parser.add_argument("--train-dir", default="train_avi-480p", type=str, help="name of train dir")
     parser.add_argument("--val-dir", default="val_avi-480p", type=str, help="name of val dir")
+    parser.add_argument(
+        "--kinetics-version", default="400", type=str, choices=["400", "600"], help="Select kinetics version"
+    )
     parser.add_argument("--model", default="r2plus1d_18", type=str, help="model name")
     parser.add_argument("--device", default="cuda", type=str, help="device (Use cuda or cpu Default: cuda)")
     parser.add_argument("--clip-len", default=16, type=int, metavar="N", help="number of frames per clip")
--- a/torchvision/io/video.py
+++ b/torchvision/io/video.py
@@ -153,14 +153,11 @@ def _read_from_stream(
         gc.collect()
 
     if pts_unit == "sec":
+        # TODO: we should change all of this from ground up to simply take
+        # sec and convert to MS in C++ 
         start_offset = int(math.floor(start_offset * (1 / stream.time_base)))
         if end_offset != float("inf"):
             end_offset = int(math.ceil(end_offset * (1 / stream.time_base)))
-    else:
-        warnings.warn(
-            "The pts_unit 'pts' gives wrong results and will be removed in a "
-            + "follow-up version. Please use pts_unit 'sec'."
-        )
 
     frames = {}
     should_buffer = True


references/video_classification/train.py
```diff
@@ -11,7 +11,7 @@
 import utils
 from torch import nn
 from torch.utils.data.dataloader import default_collate
-from torchvision.datasets.samplers import DistributedSampler, UniformClipSampler, RandomClipSampler
```
pmeier, line 14: This seems unrelated. Please revert.

references/video_classification/presets.py
```diff
@@ -31,7 +30,6 @@ class VideoClassificationPresetEval:
     def __init__(self, *, crop_size, resize_size, mean=(0.43216, 0.394666, 0.37645), std=(0.22803, 0.22145, 0.216989)):
         self.transforms = transforms.Compose(
             [
-                ConvertBHWCtoBCHW(),
```
pmeier, line 34: Why can we simply remove this? It seems something is off here. The old `Kinetics400` states

https://github.com/pytorch/vision/blob/970ba3555794d163daca0ab95240d21e3035c304/torchvision/datasets/kinetics.py#L292

whereas the new `Kinetics` states

https://github.com/pytorch/vision/blob/970ba3555794d163daca0ab95240d21e3035c304/torchvision/datasets/kinetics.py#L69

Was this intentional or did we botch this during the introduction of the new class? In any way, we need to adapt our deprecation warnings accordingly:

https://github.com/pytorch/vision/blob/970ba3555794d163daca0ab95240d21e3035c304/torchvision/datasets/kinetics.py#L252-L254

https://github.com/pytorch/vision/blob/970ba3555794d163daca0ab95240d21e3035c304/torchvision/datasets/kinetics.py#L308-L311

references/video_classification/train.py
```diff
@@ -145,9 +145,11 @@ def main(args):
     else:
         if args.distributed:
             print("It is recommended to pre-compute the dataset cache on a single-gpu first, as it will be faster")
-        dataset = torchvision.datasets.Kinetics400(
-            traindir,
+        dataset = torchvision.datasets.Kinetics(
+            args.data_path,
```
pmeier, line 149: I'm guessing this is done, because we can now select the split through the dataset class rather than using a different root directory, correct? If yes, can we remove the `traindir` and `valdir` variables above and from the CLI?

torchvision/io/video.py
```diff
@@ -153,14 +153,11 @@ def _read_from_stream(
...
+        # TODO: we should change all of this from ground up to simply take
+        # sec and convert to MS in C++ 
         start_offset = int(math.floor(start_offset * (1 / stream.time_base)))
         if end_offset != float("inf"):
             end_offset = int(math.ceil(end_offset * (1 / stream.time_base)))
-    else:
-        warnings.warn(
-            "The pts_unit 'pts' gives wrong results and will be removed in a "
-            + "follow-up version. Please use pts_unit 'sec'."
-        )
```
pmeier, line 163: Instead of this warning, which seems to do the right thing, can't we change the reference script to not use `"pts"`?

references/video_classification/train.py
bjuncek: Yup; and done

torchvision/io/video.py
bjuncek: Not really; or rather, it used to. The issue is that the backend (both pyav and VideoReader) consume data in `pts` which is unfortunate as they are not standardised per stream (e.g. audio and video have different timebases so pts=24 means different things). In a PR some time ago, we've tried to fix this by introducing `sec` as a default unit that would then be converted to appropriate stream. This caused an imprecision issue (bc thigs would be incorrectly rounded) which broke all datasets bc VideoClips creation requires fairly precise pts and such rounding erros cause major issues. We've initially fixed these by using `pts` as a default in VideoClips (which raises this warrning). This would be fine (i.e. I'd keep the warning and try to re-work the video clips) if all else was equal, but the PR mentioned above was reverted for internal issues, so this warrning is not really relevant anymore. 

Having said that, I have a major PR with a VideoReading re-work in plan that will make this redundant in a way.

bjuncek: @prabhat00155  most likely, as he was the one who originally implemented and then reverted the sync PR.
But we can also keep this open as I'm bound to finish the rework by the end of next week

datumbox:
@bjuncek @pmeier We need to speed up the work on this as we are about to cut the branch for the release. If more work is needed, I would just address the deprecation warning on a different PR.

Concerning the removal of `ConvertBHWCtoBCHW`, we need to be extra careful with this because it affects how people are doing inference and use their models. See:
https://github.com/pytorch/vision/blob/769ae132152367f73ec71a1c7ab226e3127ae624/torchvision/transforms/_presets.py#L96-L102

bjuncek:
Hi @datumbox ,
So what is your proposal for this?

In summary:
1. Deprication warning can be straight-up removed as it's pointless at the moment. The goal is either to make it less pointless (by re-writing a solid chunk of videoIO), or to have a status quo where users can use whatever they like.
2. Addressing deprecation of Kinetics, by extension means removing of the BHWC to BCHW transform. I feel as if adding deprecation warning should be plenty, and we can liaison with whomever is in charge of transforms (Victor?) to figure out the best way to handle this on the preset level. I don't understand enough about _presets to have an opinion on it.
3.

datumbox:
@bjuncek I'm OK to removing the deprecation warning until we have a proper video module to recommend.

Concerning the presets, the problem is that the removal of the permutation affects how our users use the existing models to perform inference. This is a big problem and we need to figure it out prior the release. Else we will be looking at a BC breaking change on the near future. Effectively we need to be clear, what is the order of dimensions that the preprocessing (preset) transforms should accept the input and stick with it.

cc @pmeier @vfdev-5

bjuncek:
Ok, that makes sense. 
So in the short term, should I just add the parameter for the output format to `Kinetics` class, and set it to legacy by default and add a warning that it's about to be changed?

datumbox:
My current understanding (could be wrong) is that the permutation should not be removed because `read_video()` returns the output as `(T, H, W, C)` so the permutation is necessary and should remain in place. Correct me if I'm wrong, but if `read_video` uses this format then this is no longer about changing the order inside the Kinetics but rather changing the format everywhere in the library. That's a very big discussion to be had so close to the release with major BC-breaking implications.

For more info on using `read_video` with the new presets see the example at the [docs](https://pytorch.org/vision/main/models.html#video-classification).

vfdev-5:
Geometric transforms, like Resize, should assume the input data of shape `[..., H, W]` and for Normalize it should be `(..., C, H, W)`. So we need to adapt input data format before feeding it into transformations and finally readapt it for the model's expected format.

bjuncek:
So, you are right, the `read_video` outputs `(t, h, w, c)`. 
In new `Kinetics` class, we introduce this permutation in order to save the users of going to the trouble of permuting stuff every time they get a sample (see [here](https://github.com/pytorch/vision/blob/37665a0b3fd66661cf8d89e7d2cb99abd10a0f64/torchvision/datasets/kinetics.py#L236-L238)), which is correctly documented in _the Kinetics class_ and has nothing to do with `read_video`. 

If we want to prevent BC, we can simply comment out the lines referenced above, and return the  transform to the reference.

pmeier:
@datumbox The "BC break", i.e. the changed dimension order, happened in #3680 and more specifically after https://github.com/pytorch/vision/pull/3680#discussion_r646403139 almost a year ago.

datumbox:
Thanks for the info.

I understand the reasons why we wanted to switch to `(t, c, h, w)` but we could have done better work when we introduced the BC breaking change by ensuring that all usages/references/documentation/examples in the codebase are updated. 

To fix things prior the release, I think we need to:
1. Update the references to adopt the new Kinetics (this PR)
2. Update all the code examples to do permute the output of `read_video` immediately after reading it.
3. Update the presets to remove the `(N, T, H, W, C) => (N, T, C, H, W)` conversion.

Please let me know if this makes sense to you.

references/video_classification/train.py
```diff
@@ -312,8 +316,9 @@ def parse_args():
     parser = argparse.ArgumentParser(description="PyTorch Video Classification Training")
 
     parser.add_argument("--data-path", default="/datasets01_101/kinetics/070618/", type=str, help="dataset path")
-    parser.add_argument("--train-dir", default="train_avi-480p", type=str, help="name of train dir")
-    parser.add_argument("--val-dir", default="val_avi-480p", type=str, help="name of val dir")
+    parser.add_argument(
+        "--kinetics-version", default="400", type=str, choices=["400", "600"], help="Select kinetics version"
```
datumbox, line 320: You need to update the commands shown at [README.md](https://github.com/pytorch/vision/blob/main/references/video_classification/README.md).

references/video_classification/train.py
datumbox: @pmeier @bjuncek I checked on the servers, I believe there is a reason why these parameter are there. The datasets contain the training/val data at different preprocessed sizes. Example:

```
/datasets01/kinetics/070618/400$ ls 
README  list  list_cvt  scripts  test  train  train_avi-160p  train_avi-288p  val  val_avi-160p  val_avi-288p
```

Can we please restore this?

torchvision/datasets/kinetics.py
```diff
@@ -308,6 +308,7 @@ def __init__(
         warnings.warn(
             "The Kinetics400 class is deprecated since 0.12 and will be removed in 0.14."
             "Please use Kinetics(..., num_classes='400') instead."
+            "Note that Kinetics(..., num_classes='400') returns video in a more logicalTensor[T, C, H, W] format."
```
NicolasHug, line 311:
```suggestion
            "Note that Kinetics(..., num_classes='400') returns video in a more logical Tensor[T, C, H, W] format."
```

references/video_classification/train.py
bjuncek: Done.

references/video_classification/train.py
NicolasHug: As discussed with @datumbox : 

the different pre-processed size don't seem to be part of the original downloadable dataset. Looks more like an internal pre-processing that was made by someone.

*IF* someone was relying on this originally, they should have gotten deprecation warning telling them to use Kinetics, and they should have noticed that this wasn't supported anymore, which they should have told us.

This never happened, so we can assume this support for the different sizes isn't actually needed. If it is, we'll know soon enough.

So we're OK to remove that.

torchvision/io/video.py
NicolasHug: As discussed with @datumbox 


We'll put this warning back. There should only be 2 reasons to remove this warning:

- fix the actual bug
- completely forbid "pts".

NicolasHug (commented):
Thanks all. I've had a sync with @datumbox and I'll proceed to a few changes (listed below), and merge the PR.

I'll also follow up with a PR that adds a new `format=...` parameter to relevant datasets and video readers.

PR #5959, opened 2022-05-06
[testing] Test with NVFuser enabled

davidberard98:
Stack from [ghstack](https://github.com/ezyang/ghstack):
* __->__ #5959


davidberard98:
@datumbox  thanks! We're tracking these issues here https://github.com/pytorch/pytorch/issues/76791 and there should be a fix landing today

davidberard98:
to summarize failures:
- many are the same failures as were seen in other PRs opened around the same time
- https://github.com/pytorch/pytorch/issues/76791 these were also observed, and a fix was landed.

PR #5961, opened 2022-05-07
feat: add functional center crop on mask

federicopozzi33:
Related to #5782

<details>
  <summary>Code</summary>

```python
import torch
import torchvision
from torchvision.prototype.transforms.functional import center_crop_bounding_box, center_crop_image_tensor, center_crop_segmentation_mask

size = (64, 76)
# xyxy format
in_boxes = [
    [10, 15, 25, 35],
    [50, 5, 70, 22],
    [45, 46, 56, 62],
    [size[1] // 2 - 3, size[0] // 2 - 3, size[1] // 2 + 3, size[0] // 2 + 3]
]
labels = [1, 2, 3, 4]

mask = np.zeros(size, dtype=np.int64)
for in_box, label in zip(in_boxes, labels):
    im1[in_box[1]:in_box[3], in_box[0]:in_box[2], :] = (127, 127, 127)
    mask[in_box[1]:in_box[3], in_box[0]:in_box[2]] = label
    
)
in_mask = features.SegmentationMask(torch.tensor(mask)).view(1, *size)
    
output_size = (45, 50)

out_boxes = center_crop_bounding_box(
    in_boxes, 
    in_boxes.format,
    output_size=output_size,
    image_size=in_boxes.image_size
)
print(out_boxes)

out_mask = center_crop_segmentation_mask(
    in_mask,  
    output_size 
)

t_im2 = center_crop_image_tensor(t_im1, output_size)


import cv2


plt.figure(figsize=(14, 10))

plt.subplot(2,3,1)
plt.title("Input image + bboxes")
for in_box in in_boxes:    
plt.imshow(r1)


plt.subplot(2,3,2)
plt.title("Input segm mask")
plt.imshow(in_mask[0, :, :].cpu().numpy())


plt.subplot(2,3,3)
plt.title("Input image + bboxes + segm mask")
plt.imshow(r1, alpha=0.5)
plt.imshow(in_mask[0, :, :].cpu().numpy(), alpha=0.75)


plt.subplot(2,3,4)
plt.title("Output image + bboxes")
for out_box in out_boxes:
plt.imshow(r2)


plt.subplot(2,3,5)
plt.title("Output segm mask")
plt.imshow(out_mask[0, :, :].cpu().numpy())

plt.subplot(2,3,6)
plt.title("Output image + bboxes + segm mask")
plt.imshow(r2, alpha=0.5)
plt.imshow(out_mask[0, :, :].cpu().numpy(), alpha=0.75)
```
  
</details>

pad_viz

### DIFF
--- a/test/test_prototype_transforms_functional.py
+++ b/test/test_prototype_transforms_functional.py
@@ -10,10 +10,10 @@ from common_utils import cpu_and_gpu
 from torch import jit
 from torch.nn.functional import one_hot
 from torchvision.prototype import features
+from torchvision.prototype.transforms.functional._geometry import _center_crop_compute_padding
 from torchvision.prototype.transforms.functional._meta import convert_bounding_box_format
 from torchvision.transforms.functional import _get_perspective_coeffs
 from torchvision.transforms.functional_tensor import _max_value as get_max_value
-
 
 make_tensor = functools.partial(torch.testing.make_tensor, device="cpu")
 
@@ -423,7 +423,7 @@ def center_crop_bounding_box():
 
 def center_crop_segmentation_mask():
     for mask, output_size in itertools.product(
-        make_segmentation_masks(),
+        make_segmentation_masks(image_sizes=((16, 16), (7, 33), (31, 9)), extra_dims=((), (4,), (2, 3))),
         [[4, 3], [42, 70], [4]],  # crop sizes < image sizes, crop_sizes > image sizes, single crop size
     ):
         yield SampleInput(mask, output_size)
@@ -1348,10 +1348,20 @@ def test_correctness_center_crop_bounding_box(device, output_size):
 
 
 @pytest.mark.parametrize("device", cpu_and_gpu())
-@pytest.mark.parametrize("output_size", [[4, 3], [4], [7, 7]])
+@pytest.mark.parametrize("output_size", [[4, 2], [4], [7, 6]])
 def test_correctness_center_crop_segmentation_mask(device, output_size):
     def _compute_expected_segmentation_mask(mask, output_size):
-        return F.center_crop_image_tensor(mask, output_size)
+        crop_height, crop_width = output_size if len(output_size) > 1 else [output_size[0], output_size[0]]
+
+        _, image_height, image_width = mask.shape
+        if crop_width > image_height or crop_height > image_width:
+            padding = _center_crop_compute_padding(crop_height, crop_width, image_height, image_width)
+            mask = F.pad_image_tensor(mask, padding, fill=0)
+
+        left = round((image_width - crop_width) * 0.5)
+        top = round((image_height - crop_height) * 0.5)
+
+        return mask[:, top : top + crop_height, left : left + crop_width]
 
     mask = torch.randint(0, 2, size=(1, 6, 6), dtype=torch.long, device=device)
     actual = F.center_crop_segmentation_mask(mask, output_size)
--- a/test/test_prototype_transforms_functional.py
+++ b/test/test_prototype_transforms_functional.py
@@ -1,7 +1,6 @@
 import functools
 import itertools
 import math
-from unittest.mock import patch, Mock
 
 import numpy as np
 import pytest
@@ -1348,40 +1347,14 @@ def test_correctness_center_crop_bounding_box(device, output_size):
         torch.testing.assert_close(output_boxes, expected_bboxes)
 
 
-def test_correctness_center_crop_segmentation_mask_on_fixed_input(device):
-    mask = torch.ones((1, 6, 6), dtype=torch.long, device=device)
-    mask[:, 1:5, 2:4] = 0
+@pytest.mark.parametrize("device", cpu_and_gpu())
+@pytest.mark.parametrize("output_size", [[4, 3], [4], [7, 7]])
+def test_correctness_center_crop_segmentation_mask(device, output_size):
+    def _compute_expected_segmentation_mask(mask, output_size):
+        return F.center_crop_image_tensor(mask, output_size)
 
-    out_mask = F.center_crop_segmentation_mask(mask, [2])
-    expected_mask = torch.zeros((1, 4, 2), dtype=torch.long, device=device)
-    torch.testing.assert_close(out_mask, expected_mask)
-
-
-@pytest.mark.parametrize("output_size", [[4, 3], [4]])
-def test_correctness_center_crop_segmentation_mask(output_size):
-    def _compute_expected_segmentation_mask():
-        _output_size = output_size if isinstance(output_size, tuple) else (output_size, output_size)
-
-        _, h, w = mask.shape
-        left = w - _output_size[0]
-        top = h - _output_size[1]
-
-        return mask[:, top : _output_size[1], left : _output_size[0]]
-
-    mask = torch.randint(0, 2, shape=(1, 6, 6))
+    mask = torch.randint(0, 2, size=(1, 6, 6), dtype=torch.long, device=device)
     actual = F.center_crop_segmentation_mask(mask, output_size)
 
-    expected = _compute_expected_segmentation_mask()
-    assert expected == actual
-
-
-@pytest.mark.parametrize("output_size", [[4, 3], [4], [7, 7]])
-@patch("torchvision.prototype.transforms.functional._geometry.center_crop_image_tensor")
-def test_correctness_center_crop_segmentation_mask_mock(center_crop_mock, output_size):
-    mask, expected = Mock(spec=torch.Tensor), Mock(spec=torch.Tensor)
-    center_crop_mock.return_value = expected
-
-    out_mask = F.center_crop_segmentation_mask(mask, output_size)
-
-    center_crop_mock.assert_called_once_with(img=mask, output_size=output_size)
-    assert expected is out_mask
+    expected = _compute_expected_segmentation_mask(mask, output_size)
+    torch.testing.assert_close(expected, actual)
--- a/test/test_prototype_transforms_functional.py
+++ b/test/test_prototype_transforms_functional.py
@@ -1,6 +1,7 @@
 import functools
 import itertools
 import math
+from unittest.mock import patch, Mock
 
 import numpy as np
 import pytest
@@ -419,6 +420,12 @@ def center_crop_bounding_box():
         yield SampleInput(
             bounding_box, format=bounding_box.format, output_size=output_size, image_size=bounding_box.image_size
         )
+def center_crop_segmentation_mask():
+    for mask, output_size in itertools.product(
+        make_segmentation_masks(),
+        [[4, 3], [42, 70], [4]],  # crop sizes < image sizes, crop_sizes > image sizes, single crop size
+    ):
+        yield SampleInput(mask, output_size)
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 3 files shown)


federicopozzi33:
@pmeier, @vfdev-5 thoughts about testing the correctness of the function by mocking the function that really does the work? It's a proposal related to what we were saying in #5866.

IMO, there is no logic in `center_crop_segmentation_mask`, I'm just calling the `center_crop_image_tensor` function. If the latter is properly tested, the current test should be enough.

@vfdev-5, for the visualization side, the function `center_crop_bounding_box` is still missing. Is it ok to update the description  as soon as it's implemented?

vfdev-5:
> IMO, there is no logic in center_crop_segmentation_mask, I'm just calling the center_crop_image_tensor function. If the latter is properly tested, the current test should be enough.

@federicopozzi33 I agree, I have a similar reflection about that while working on `center_crop_bounding_box`. Currently, I opted to write `compute_expected_bbox` function and check for results... There could be however few corner cases.

> @vfdev-5, for the visualization side, the function center_crop_bounding_box is still missing. Is it ok to update the description as soon as it's implemented?

You can make visualization for segm mask and then add bbox when available ?

federicopozzi33:
> @federicopozzi33 I agree, I have a similar reflection about that while working on `center_crop_bounding_box`. Currently, I opted to write `compute_expected_bbox` function and check for results... There could be however few corner cases.

I added a test with random input too, but IMO it's not needed here: I don't see any corner cases.

> We have to think about how to test composed ops in general. 1) Ensure that we call base ops in a right way or 2) check if the output is what we expect (expected result could be either hard-coded, computed using base ops or recomputed in a different way) or 3) something else ?

We can do both 1) and 2). I'd achieve the former by patching the base ops, and the latter with fixed input. 
Then, 3) specific test cases can be added for corner cases.

> You can make visualization for segm mask and then add bbox when available ?

I've updated the description.

vfdev-5:
@federicopozzi33 CI failures are real : https://github.com/pytorch/vision/runs/6376638110?check_suite_focus=true#step:9:127
please check

test/test_prototype_transforms_functional.py
```diff
@@ -1337,3 +1346,42 @@ def _compute_expected_bbox(bbox, output_size_):
...
+    mask = torch.ones((1, 6, 6), dtype=torch.long, device=device)
+    mask[:, 1:5, 2:4] = 0
+
+    out_mask = F.center_crop_segmentation_mask(mask, [2])
+    expected_mask = torch.zeros((1, 4, 2), dtype=torch.long, device=device)
+    torch.testing.assert_close(out_mask, expected_mask)
+
+
+@pytest.mark.parametrize("output_size", [[4, 3], [4]])
+def test_correctness_center_crop_segmentation_mask(output_size):
```
vfdev-5, line 1361: I think we could merge these two tests into one with parameters `device` and `output_size`.

vfdev-5:
@federicopozzi33 thanks for the update. Code and test looks good to me (I left a minor suggestion). Once you fixed the problem with CI (probably due to resolving conflicts) we can merge it.

@pmeier any thoughts from your side about testing, particularly mocking part ?

test/test_prototype_transforms_functional.py
```diff
@@ -1337,3 +1346,42 @@ def _compute_expected_bbox(bbox, output_size_):
...
+
+    expected = _compute_expected_segmentation_mask()
+    assert expected == actual
+
+
+@pytest.mark.parametrize("output_size", [[4, 3], [4], [7, 7]])
+@patch("torchvision.prototype.transforms.functional._geometry.center_crop_image_tensor")
+def test_correctness_center_crop_segmentation_mask_mock(center_crop_mock, output_size):
+    mask, expected = Mock(spec=torch.Tensor), Mock(spec=torch.Tensor)
+    center_crop_mock.return_value = expected
```
pmeier, line 1382: We can use the [`mocker`](https://pytest-mock.readthedocs.io/en/latest/usage.html) fixture here:

```suggestion
def test_correctness_center_crop_segmentation_mask_mock(mocker, center_crop_mock, output_size):
    mask = mocker.MagicMock(spec=torch.Tensor)
    expected = mocker.MagicMock(spec=torch.Tensor)
    mocker.patch("torchvision.prototype.transforms.functional._geometry.center_crop_image_tensor", return_value=expected)
```

Note that you have to install `pytest-mock` for it to be available. We are doing that in CI:

https://github.com/pytorch/vision/blob/3ec4b94998bd2b413f1a5c77bfed590ce684b9b9/.github/workflows/prototype-tests.yml#L40

test/test_prototype_transforms_functional.py
```diff
@@ -1,6 +1,7 @@
 import functools
 import itertools
 import math
+from unittest.mock import patch, Mock
```
pmeier, line 4: This can go if we use the `mocker` fixture.

```suggestion
```

pmeier (commented):
One minor nit about mocking in `pytest`. Otherwise testing looks good. Still, I want to hear @NicolasHug's opinion on mocking in general for this use case.

pmeier (changes requested):
I've discussed this with @NicolasHug offline and we both agree the mock test seems a little overkill here. The function is a one-liner without any additional logic. The test is orders of magnitude more complex. Thus, in the future it is more likely that the test breaks rather then the kernel.

Plus, we already have tests that check the numerical correctness. Thus, the test is not bringing any new information.

What was the reason to go for the complex mocking here?

vfdev-5:
> What was the reason to go for the complex mocking here?

It is the first time we opt to use mock. I agree that in case of mask it is one-liner and it does not bring any useful checks but it can serve as an example for other composite ops, IMO. 
Anyway, as you are reworking tests there will be other pass on tests, so no strong opinion from my side about removing or keeping this test.

test/test_prototype_transforms_functional.py
federicopozzi33: Ok. I thought that having two test (one with fixed input + device and one with random input) was a sort of scheme to follow. That's why I kept them separated.

pmeier:
@federicopozzi33 I agree with the general sentiment, but I would go a different route. If `center_crop_image_tensor` is already tested and thus implements a `_prepare_expected` function for all cases, can't we simply re-use it here?

federicopozzi33:
> mply re-use it here?

If the only thing that my function does is to call another tested function, IMO, mocking is the way. It's the same that calling the inner function, but without really doing it. Just two different ways to approach the same problem. You choose :)

So, should I add replace my naive `prepare_expected` to cover many more test cases?

pmeier:
> If the only thing that my function does is to call another tested function, IMO, mocking is the way.

That is a dangerous conclusion. Mocking something always means that we switch from black box testing to white box testing. Meaning, we now need to know the internals of the function we want to test. That isn't a bad thing in general, but brings other problems with it. For example, if the internals change but the interface stays the same, the test might break although it shouldn't. 

Imagine we rename `center_crop_image_tensor` to `center_crop_image`. The test for the segmentation mask would now fail since the mock target needs to change. Admittedly, the fix wouldn't be hard, but it is still a breakage that does not need to happen.

In this situation here, IMO it would be best to just use the expected result computation for images and segmentation and stick to black box testing.

federicopozzi33:
> > If the only thing that my function does is to call another tested function, IMO, mocking is the way.
> 
> That is a dangerous conclusion. Mocking something always means that we switch from black box testing to white box testing. Meaning, we now need to know the internals of the function we want to test. That isn't a bad thing in general, but brings other problems with it. For example, if the internals change but the interface stays the same, the test might break although it shouldn't.
> 
> Imagine we rename `center_crop_image_tensor` to `center_crop_image`. The test for the segmentation mask would now fail since the mock target needs to change. Admittedly, the fix wouldn't be hard, but it is still a breakage that does not need to happen.
> 
> In this situation here, IMO it would be best to just use the expected result computation for images and segmentation and stick to black box testing.

I agree with you. I'd just like to clarify one point: IMO, mocking a public function is very different from mocking a private one, especially in cases like ours, considering that modifying public functions means doing a "major change".

Anyway, I added some extra test cases (output size bigger than image and odd output dimensions) and I'm computing the expected with the `center_crop_image_tensor`, so please take a look.

pmeier:
> considering that modifying public functions means doing a "major change".

That is true in general, but not so much during the prototype phase.

test/test_prototype_transforms_functional.py
```diff
@@ -421,6 +421,14 @@ def center_crop_bounding_box():
         )
 
 
+def center_crop_segmentation_mask():
+    for mask, output_size in itertools.product(
+        make_segmentation_masks(),
+        [[4, 3], [42, 70], [4]],  # crop sizes < image sizes, crop_sizes > image sizes, single crop size
```
pmeier, line 427: Should we maybe not rely on the default values here to make sure we actually get the cases detailed in the comment?

test/test_prototype_transforms_functional.py
```diff
@@ -1337,3 +1345,16 @@ def _compute_expected_bbox(bbox, output_size_):
         else:
             expected_bboxes = expected_bboxes[0]
         torch.testing.assert_close(output_boxes, expected_bboxes)
+
+
+@pytest.mark.parametrize("device", cpu_and_gpu())
+@pytest.mark.parametrize("output_size", [[4, 3], [4], [7, 7]])
+def test_correctness_center_crop_segmentation_mask(device, output_size):
+    def _compute_expected_segmentation_mask(mask, output_size):
+        return F.center_crop_image_tensor(mask, output_size)
```
pmeier, line 1354: I'm not really happy about this given that we don't seem to have a reference test that makes sure `F.center_crop_image_tensor`. Thus, if that had a bug, we would propagate it to this test. 

Plus, we now use exactly the same implementation for the test and kernel and thus are unable to detect any bug with this test.

Maybe we can have something like

```py
class TestCorrectnessCenterCrop:
    def _compute_expected_image_or_segmentation_mask(self, input, output_size):
        # manual implementation that does not use our tensor kernels
        # IMO, it would be fine to use `center_crop_image_pil` here.
        pass
    
    def test_image(self):
        pass
    
    def test_segmentation_mask(self):
        pass
```

vfdev-5: I think the problem here is with the `_compute_expected_segmentation_mask` implementation which is exactly the same as `F.center_crop_segmentation_mask`. IMO, this way of testing does not bring any value. 

As for restructuring with `TestCorrectnessCenterCrop`, @pmeier it can be maybe done in your functional tests refactor PR, not here.

Let's move forward with this. If we abandoned somehow the idea of mocking `F.center_crop_image_tensor` op and just checking the args, now let's recode `_compute_expected_segmentation_mask` such that we compute in another way a center crop and check output vs expected results.

federicopozzi33:
Sorry, I'm pretty busy these days, I'll fix the PR in the next few days (probably during the weekend).

test/test_prototype_transforms_functional.py
federicopozzi33: I still use the default values, but I have made them explicit.

vfdev-5 (approved):
LGTM, thanks @federicopozzi33

test/test_prototype_transforms_functional.py
```diff
@@ -421,6 +421,14 @@ def center_crop_bounding_box():
         )
 
 
+def center_crop_segmentation_mask():
+    for mask, output_size in itertools.product(
+        make_segmentation_masks(image_sizes=((16, 16), (7, 33), (31, 9)), extra_dims=((), (4,), (2, 3))),
```
pmeier, line 426: Nit: it is a good idea to be explicit about the `image_sizes` here due to their relation to the crop size, but the same is not true for the `extra_dims`:

```suggestion
        make_segmentation_masks(image_sizes=((16, 16), (7, 33), (31, 9))),
```

On the other hand it also doesn't hurt to be explicit about them. Feel free to leave as is.

test/test_prototype_transforms_functional.py
```diff
@@ -1337,3 +1345,26 @@ def _compute_expected_bbox(bbox, output_size_):
...
+    def _compute_expected_segmentation_mask(mask, output_size):
+        crop_height, crop_width = output_size if len(output_size) > 1 else [output_size[0], output_size[0]]
+
+        _, image_height, image_width = mask.shape
+        if crop_width > image_height or crop_height > image_width:
+            padding = _center_crop_compute_padding(crop_height, crop_width, image_height, image_width)
+            mask = F.pad_image_tensor(mask, padding, fill=0)
+
+        left = round((image_width - crop_width) * 0.5)
+        top = round((image_height - crop_height) * 0.5)
```
pmeier, line 1362: The kernel has an additional `int` call:

https://github.com/pytorch/vision/blob/b969cca783cfe6c1009b7bf60609e28d83c973a6/torchvision/prototype/transforms/functional/_geometry.py#L585-L586

@vfdev-5 I recall we had issues with `round` before. Should we just switch to `int` in general?

pmeier (approved):
Thanks @federicopozzi33!

test/test_prototype_transforms_functional.py
vfdev-5: @pmeier I do not quite remember what was the issue with round (maybe jit behaves differently to eager mode). For me the code you mention is more like a definition of crop_top and crop_left. For example, for bboxes we could also keep these values as float but let's define that crop_top/left are rounded integers.

PR #5966, opened 2022-05-08
CleanUp DenseNet code

NicolasHug:
This PR removes unnecessary calls to `add_module()` and removes confusing type annotations.


datumbox (approved):
This makes sense and the changes LGTM.

Issue #5965, opened 2022-05-08
Allow weight tables to be sorted


NicolasHug:
It'd be nice if [weight tables](https://pytorch.org/vision/main/models_new.html#table-of-all-available-classification-weights) could be sorted by number of params or accuracies.

This post might be a good start: https://stackoverflow.com/questions/64443832/sorting-table-with-rst-and-python-sphinx-in-html-output

oke-aditya:
I will have a go at this. On a high level it seems feasible.

PR #5968, opened 2022-05-08
Sort Tables for docs

oke-aditya:
Tries a hand at #5965 

Currently It looks dirty. And the sorting doesn't work for params column (JS and string sorting is a programming meme in itself).

### DIFF
--- a/docs/source/_static/js/main.js
+++ b/docs/source/_static/js/main.js
@@ -1,0 +1,3 @@
+$(document).ready( function () {
+    $('table.datatable').DataTable();
+} );
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -147,8 +147,15 @@ html_static_path = ["_static"]
 
 # TODO: remove this once https://github.com/pytorch/pytorch_sphinx_theme/issues/125 is fixed
 html_css_files = [
+    "https://cdn.datatables.net/1.10.23/css/jquery.dataTables.min.css",
     "css/custom_torchvision.css",
 ]
+
+html_js_files = [
+    "https://cdn.datatables.net/1.10.23/js/jquery.dataTables.min.js",
+    "js/main.js",
+]
+
 
 # -- Options for HTMLHelp output ------------------------------------------
 
--- a/docs/source/dummy_table.rst
+++ b/docs/source/dummy_table.rst
@@ -1,0 +1,122 @@
+.. _dummy_table:
+
+Dummy Table
+###########
+
+Same old classification table let's see.
+
+.. rst-class:: table-weights
+.. table::
+    :class: datatable
+    :widths: 100 20  20 10
+
+    ==================================================================================  ===========  ===========  ============  =========================================================================================================
+    **Weight**                                                                            **Acc@1**    **Acc@5**  **Params**    **Recipe**
+    ==================================================================================  ===========  ===========  ============  =========================================================================================================
+    :class:`AlexNet_Weights.IMAGENET1K_V1 <AlexNet_Weights>`                                 56.522       79.066  61.1M         `link <https://github.com/pytorch/vision/tree/main/references/classification#alexnet-and-vgg>`__
+    :class:`ConvNeXt_Base_Weights.IMAGENET1K_V1 <ConvNeXt_Base_Weights>`                     84.062       96.87   88.6M         `link <https://github.com/pytorch/vision/tree/main/references/classification#convnext>`__
+    :class:`ConvNeXt_Large_Weights.IMAGENET1K_V1 <ConvNeXt_Large_Weights>`                   84.414       96.976  197.8M        `link <https://github.com/pytorch/vision/tree/main/references/classification#convnext>`__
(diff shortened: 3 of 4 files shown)


oke-aditya:
I believe this feature is very cool. As it's painful to find the best model from the list (and our library will keep adding models) + potentially other statistics such as say run-time time etc. Let's try to hack some way :)

oke-aditya:
Update:

So a few tweaks to the CSS in Jquery made it work :smiley: 
Although the sad thing is something is still broken with classification table. Also some tweak can probably make it work on parms / disable it on params.

Ref
https://datatables.net/examples/basic_init/filter_only.html
https://datatables.net/reference/option/
https://stackoverflow.com/questions/13290626/datatable-hide-the-show-entries-dropdown-but-keep-the-search-box

datumbox:
@oke-aditya Just confirming the intention of this PR. I understand that currently you try to show-case the sorting functionality. You just copy-paste the accuracy tables for convenience and plan later to adopt @NicolasHug's solution that builds these tables dynamically form the weight meta-data, right?

oke-aditya:
Yes, I will build upon @NicolasHug solution. I inentionally took the tables out for dry run and to check how the functionality works.

docs/source/dummy_table.rst
```diff
@@ -0,0 +1,143 @@
+.. _dummy_table:
+
+Dummy Table
+###########
+
+Same old classification table let's see.
+
+.. rst-class:: table-weights
+.. table::
+    :class: datatable
```
NicolasHug, line 10: The way to set the css class of our tables is to use `.. rst-class: ` as above. We set that here:

https://github.com/pytorch/vision/blob/3ec4b94998bd2b413f1a5c77bfed590ce684b9b9/docs/source/conf.py#L379:L379

I ak garbage at CSS and web dev in general. Do you know if stuff can have 2 different classes?

NicolasHug (commented):
Thanks @oke-aditya ! The table with tweaks looks nice, although it might require some adjustements on the colomn width. Do you mind cleaning up the PR to only rely on our table generation instead of hard-coding the table? It will be easier what can be done to address it. 
Thank you!

docs/source/dummy_table.rst
oke-aditya: Yeah why not stuff can have as many CSS classes as possible. There is no limit. Only thing is classes are overridden according to CSS class hierarchy. (inline css over style tag over DOM selector).

NicolasHug: OK, then I assume we can try updating the line to

```
table_file.write(".. rst-class:: table-weights datatable\n")
```

(not sure about the order)

oke-aditya: Yeah don't worry i will take care :smile:

docs/source/_static/js/main.js
```diff
@@ -0,0 +1,7 @@
+$(document).ready( function () {
```
oke-aditya, line 1: This is jQuery, a slightly big JS lib. Which I feel we can avoid with plain simple vanilla JS. Using DOM manipulation.

docs/source/conf.py
```diff
@@ -147,9 +147,16 @@
 
 # TODO: remove this once https://github.com/pytorch/pytorch_sphinx_theme/issues/125 is fixed
 html_css_files = [
+    "https://cdn.datatables.net/1.10.23/css/jquery.dataTables.min.css",
     "css/custom_torchvision.css",
 ]
 
+html_js_files = [
+    "https://cdn.datatables.net/1.10.23/js/jquery.dataTables.min.js",
```
oke-aditya, line 155: Is there a preferred CDN for pytorch? Like CDN js? Or something which Meta uses and relies on.

NicolasHug: Perhaps this is something @svekars would know?

oke-aditya: An alternative is to download this file and keep in our repo. This will avoid CDN.

docs/source/conf.py
```diff
@@ -377,6 +385,7 @@ def generate_weights_table(module, table_name, metrics, include_patterns=None, e
     with open(generated_dir / f"{table_name}_table.rst", "w+") as table_file:
         table_file.write(".. rst-class:: table-weights\n")  # Custom CSS class, see custom_torchvision.css
         table_file.write(".. table::\n")
+        table_file.write("    :class: datatable\n")
```
NicolasHug, line 388: Would it work to use the `.. rst-class` directive as above? If not, do you know why?

oke-aditya:
Note that sorting still doesn't work well in Params column.

docs/source/conf.py
oke-aditya: Works fine

NicolasHug:
> Note that sorting still doesn't work well in Params column.

Do you know if we can provide a custom sort function for that specific column?

oke-aditya:
It should be possible. But I will need to search and try.

oke-aditya:
Sorry I still can't find the fix for it :( Maybe I'm doing something wrong. But my skills with JQuery and JS aren't that great either.

NicolasHug:
No problem at all @oke-aditya , thanks a lot for trying!!
I'll try to see if someone can help internally with this

PR #5975, opened 2022-05-09
Add Docs for QuantizedGoogleNet

NicolasHug:
This PR creates a "Quantized Classification models" subsection and document the quantized version of GoogleNet.


docs/source/conf.py
```diff
@@ -349,7 +349,7 @@ def inject_weight_metadata(app, what, name, obj, options, lines):
 
 
 def generate_weights_table(module, table_name, metrics, include_patterns=None, exclude_patterns=None):
-    weight_enums = [getattr(module, name) for name in dir(module) if name.endswith("_Weights")]
+    weight_enums = [getattr(module, name) for name in dir(module) if name.endswith("Weights")]
```
NicolasHug, line 352: @datumbox I think you were the one who suggested to use `_Weights` instead of `Weights`?

IIRC the reason was precisely to avoid mixing quantized and non-quantized weights, but this isn't a problem: there can't be any collision since we specify the module, and the quantized weights are kept separate in the `models.quantization` module. LMK if I'm missing something

datumbox: Out side of quantization we use the following convention: 
```
str(enum).lower() == str(model_fn) + "_Weights".lower()
```

On quantization, because the model_fn name is the same as on non-quantized, we do
```
str(enum).lower() == str(model_fn) + "_QuantizedWeights".lower()
```

Currently the idiom used in the code-base for when we need to handle both looks like [this](https://github.com/pytorch/vision/blob/main/test/test_extended_models.py#L26). I think, one reason why we don't want to mix them is because the `model.quantization` space imports the normal weights. Though we don't include these on `__all__` (hence they are considered an internal detail) they can be accessed via `torchvision.models.quantization.resnet.ResNet18_Weights` so I think it's safer to be explicit. Thoughts?

torchvision/models/quantization/googlenet.py
```diff
@@ -141,18 +141,34 @@ def googlenet(
...
+            base class. Please refer to the `source code
+            <https://github.com/pytorch/vision/blob/main/torchvision/models/quantization.googlenet.py>`_
+            for more details about this class.
+
+    .. autoclass:: torchvision.models.quantization.GoogLeNet_QuantizedWeights
+        :members:
+
+    .. autoclass:: torchvision.models.GoogLeNet_Weights
+        :members:
+        :noindex:
```
datumbox, line 171: Question: Why noindex here?

docs/source/conf.py
datumbox: Above I tried to provide context on the decision on how/why this was selected. If for the use-case of docs you think it's safe, we can adopt the `.endswith("Weights")` proposal to make things easier. No strong opinions.

torchvision/models/quantization/googlenet.py
oke-aditya: if we see error logs when we build docs we actually need this in many places. Any way to inject these?

NicolasHug: I think this has to be done manually. But there should be very few cases where this is going to be needed. It's the only use-case we have for `:noindex:` so far.

(Note that we get tons of warnings in the docs right now but it's normal, they'll be all gone once we remove `models.rst`).

datumbox (approved):
LGTM, thanks @NicolasHug!

PR #5977, opened 2022-05-09
Document ResNet architecture tweak

### LINKED ISSUE #5964: Document ResNet architecture tweak
Since we're revamping our docs, we might want to point out that our ResNet implementation slightly differs from the original one:

https://github.com/pytorch/vision/blob/970ba3555794d163daca0ab95240d21e3035c304/torchvision/models/resnet.py#L107:L111

We have this as a comment  in the code but it's worth documenting in the docs as well as `.. note::` IMHO

puhuk:
Add note for resnet architecture

Closes #5964


torchvision/models/resnet.py
```diff
@@ -575,6 +575,13 @@ class Wide_ResNet101_2_Weights(WeightsEnum):
 def resnet18(*, weights: Optional[ResNet18_Weights] = None, progress: bool = True, **kwargs: Any) -> ResNet:
     """ResNet-18 from `Deep Residual Learning for Image Recognition <https://arxiv.org/pdf/1512.03385.pdf>`__.
 
+    .. note::
+       Bottleneck in torchvision places the stride for downsampling at 3x3 convolution(self.conv2)
+       while original implementation places the stride at the first 1x1 convolution(self.conv1)
+       according to "Deep residual learning for image recognition"https://arxiv.org/abs/1512.03385.
+       This variant is also known as ResNet V1.5 and improves accuracy according to
+       https://ngc.nvidia.com/catalog/model-scripts/nvidia:resnet_50_v1_5_for_pytorch.
```
datumbox, line 583: This is added on the wrong place. ResNet18 and 34 use the basic `BasicBlock` not `Bottleneck`. It's the 50, 101 and 152 versions of the model that use the `Bottleneck` class.

puhuk: Sorry, let me correct soon !

docs/source/models/resnet.rst
```diff
@@ -6,6 +6,12 @@ ResNet
 The ResNet model is based on the `Deep Residual Learning for Image Recognition
 <https://arxiv.org/abs/1512.03385>`_ paper.
 
+.. note::
+    Bottleneck in torchvision places the stride for downsampling at 3x3 convolution(self.conv2)
+    while original implementation places the stride at the first 1x1 convolution(self.conv1)
+    according to "Deep residual learning for image recognition"https://arxiv.org/abs/1512.03385.
```
datumbox, line 12: This hyperlink will not render correctly. Given we already have a link to the paper few lines above, perhaps we can just say:
```suggestion
    according to the paper.
```

docs/source/models/resnet.rst
```diff
@@ -6,6 +6,12 @@ ResNet
...
+    This variant is also known as ResNet V1.5 and improves accuracy according to
```
datumbox, line 13: I don't think it's useful to place long string URLs in the documentation. How about:
```suggestion
    This variant improves the accuracy and it's known as `ResNet V1.5 <https://ngc.nvidia.com/catalog/model-scripts/nvidia:resnet_50_v1_5_for_pytorch>`_.
```

puhuk:
@datumbox I'm having errors with unittest_onnx which I did not change. Could you give me some clue to resolve this error. Or is there any other reason that cause the error?

datumbox:
@puhuk that's not something you broke. The main branch is also failing. Feel free to ignore it.

datumbox:
The failures on the CI are due to the `https://ghcr.io/v2/homebrew/core/gmp/manifests/6.2.1_1` being down temporarily. I checked the docs and they render fine.

PR #5979, opened 2022-05-09
Cleaning up Ops Boxes and Losses :broom: 

### LINKED ISSUE #5976: Clean up Ops Box and Losses implementations
### 🚀 The feature

Currently losses such as cIoU, dIoU and gIoU reimplement functions that exist already on [boxes ops](https://github.com/pytorch/vision/blob/main/torchvision/ops/boxes.py). In addition they contain a lot of duplicate code that can be shared. Finally the testing strategy can be improved.

This ticket is about:
- Investigating ways to re-use code on Losses from boxes. This should happen without affecting the speed (See https://github.com/pytorch/vision/pull/5877#discussion_r858595615).
- Reduce the amount of code duplication between methods gIoU/dIoU/cIoU boxes and losses
- Align the testing strategy between the different similar methods and incorporate the comments at https://github.com/pytorch/vision/pull/5786#discussion_r857325564

### Motivation, pitch

-

### Alternatives

-

### Additional context

-

oke-aditya:
Fixes #5976
Work in Progress. Opened PR to see if I keep passing tests :)


torchvision/ops/ciou_loss.py
```diff
@@ -30,50 +34,25 @@ def complete_box_iou_loss(
...
 
-    # Original Implementation : https://github.com/facebookresearch/detectron2/blob/main/detectron2/layers/losses.py
-
     if not torch.jit.is_scripting() and not torch.jit.is_tracing():
         _log_api_usage_once(complete_box_iou_loss)
 
-    boxes1 = _upcast(boxes1)
-    boxes2 = _upcast(boxes2)
+    diou_loss = distance_box_iou_loss(boxes1, boxes2, reduction="none", eps=eps)
+    intsct, union = _loss_inter_union(boxes1, boxes2)
```
datumbox, line 50: I would try to minimize speed impact and re-estimation as much as possible. Potentially a private diou method could return intermediate results so that we can reuse intersect and union?

torchvision/ops/ciou_loss.py
```diff
@@ -12,6 +13,9 @@ def complete_box_iou_loss(
 ) -> torch.Tensor:
 
     """
+    # Original Implementation from
+    https://github.com/facebookresearch/detectron2/blob/main/detectron2/layers/losses.py
```
datumbox, line 17: I would leave this as a comment on the source. Long unrendered URLs are not particularly helpful for the documentation. We should make the same change on diou.

torchvision/ops/_utils.py
```diff
@@ -67,3 +67,33 @@ def split_normalization_params(
         else:
             other_params.extend(p for p in module.parameters() if p.requires_grad)
     return norm_params, other_params
+
+
+def _upcast(t: Tensor) -> Tensor:
```
datumbox, line 72: Note that previous code used different versions of upcast. More specifically boxes permitted integers while losses didn't.

datumbox (commented):
I know I shouldn't be snooping around on draft PRs. I've added couple of comments but feel free to ignore if you think it's too early. 😄

torchvision/ops/ciou_loss.py
oke-aditya: Correct, the current refactor re computes stuff and is hence slightly slower. Let me refactor as per your suggestion.

torchvision/ops/ciou_loss.py
oke-aditya: Sphinx agrees with you :smiley: 

image


Let me re-wrtie

datumbox: Yeap. That looks ugly and it's on multiple places. If you want bring a separate quick PR that moves attributions on the main part of the methods to avoid the issue.

NicolasHug: One option is to embed the link e.g.

> Original Implementation from [Detectron2](https://github.com/facebookresearch/detectron2/blob/main/detectron2/layers/losses.py)

But the docstring should at least start by describing what the object is, even if it's very obvious from its name already. So I would suggest to write something like

```py

"""Complete Box IoU Loss.

Implementation is adapted from `Detectron2 <https://github.com/facebookresearch/detectron2/blob/main/detectron2/layers/losses.py>`__.

...
```

oke-aditya: image


That's how it's currently over main branch. I would suggest adding 3 words in end implementation adapted from [Detectron2](https://github.com/facebookresearch/detectron2/blob/main/detectron2/layers/losses.py).

oke-aditya: Will be tackling this in seperate PR anyways to unify all stuff. I feel we need bit more revamp for Ops docs.

datumbox: We should definitely provide attribution to other projects when we use code from them. Let's keep the reference on the code as a comment on the main body of the method similar to what we do in most places instead of placing a link. We can review this on the future and make changes in a coordinated manner.

torchvision/ops/giou_loss.py
```diff
@@ -1,14 +1,9 @@
...
 
 from ..utils import _log_api_usage_once
+from ._utils import _upcast, _loss_inter_union
 
-
-def _upcast(t: Tensor) -> Tensor:
-    # Protects from numerical overflows in multiplications by upcasting to the equivalent higher type
-    if t.dtype not in (torch.float32, torch.float64):
-        return t.float()
-    return t
```
datumbox, line 11: This is not the same `_upcast` as `_utils`. It doesn't support integers and converts everything to floats. Could you please review all the places where the `giou_loss._upcast()` was used and ensure the output will be a float? Basically box ops are OK to maintain things as integers but not losses.

oke-aditya: Yes I do understand that I have not used the int only _upcast I have intentionally not taken it back. Since I would like to know why 

> box ops are OK to maintain things as integers but not losses.

datumbox: The `_upcast` on boxes was introduced for operators that estimate information related to a box. Things like the area for instance. If you put integer boxes and you ask for the area of a box, you kind of expecting you receive the value in integers (that's debatable but that's how the operator worked). The issue was that if you used too small of a precision, the area estimation would overflow. So this method upcasts the boxes to a space where it's safe to do multiplications without risking overflowing for most applications.

On the losses side, I'm not aware of any application that does things on integer space. Not only that but doing reduce=mean will break things. So we need to be careful to definitely not support ints in losses. Similar care might be needed for some box operators. Area might still make sense to return as integer but I'm not 100% sure if that's the case with all the IoU metrics we deal here.

So I think it's important prior merging this PR, to make an explicit decision of what has to support integers and what doesn't, handle it appropriately and add tests and xfails to ensure we are not breaking anything.

test/test_losses.py
```diff
@@ -0,0 +1,229 @@
+import pytest
+import torch
+import torch.nn.functional as F
+from common_utils import cpu_and_gpu
+from torchvision import ops
+
+
```
oke-aditya, line 7: Making a class, inheriting and the calling method is also possible. For now is this fine?

test/test_losses.py
```diff
@@ -0,0 +1,229 @@
...
+        assert_iou_loss(ops.distance_box_iou_loss, box1s, box2s, 1.2250, dtype=dtype, device=device, reduction="mean")
+        assert_iou_loss(ops.distance_box_iou_loss, box1s, box2s, 2.4500, dtype=dtype, device=device, reduction="sum")
+
+    @pytest.mark.parametrize("device", cpu_and_gpu())
+    @pytest.mark.parametrize("dtype", [torch.float32, torch.half])
+    def test_empty_distance_iou_inputs(self, dtype, device):
+        assert_empty_loss(ops.distance_box_iou_loss, dtype, device)
+
+
+class TestFocalLoss:
```
oke-aditya, line 110: Since we are testing losses here, I felt to add this here as we don't cause confusion between files.

test/test_losses.py
```diff
@@ -0,0 +1,229 @@
+import pytest
+import torch
+import torch.nn.functional as F
+from common_utils import cpu_and_gpu
+from torchvision import ops
+
+
+def get_boxes(dtype, device):
+    box1 = torch.tensor([-1, -1, 1, 1], dtype=dtype, device=device)
```
oke-aditya, line 9: Not super happy with this choice of box. Since this is actually invalid input

datumbox:
@oke-aditya Shall we convert to draft and ping us when you are happy to review again?

oke-aditya:
Yep. :)

test/test_losses.py
abhi-glitchhg: Agree with the concern.

Detectron2 used the same set of boxes. [see this](https://github.com/facebookresearch/detectron2/blob/42c44b622eb7b179576bdac218c84c107fff8eab/tests/layers/test_losses.py#L9). 

I think we should use valid input boxes.

That being said, should we also check if the input boxes have non-negative values? 
What do you think?

oke-aditya: We cannot assert here as it will lead to cuda call and cause trouble. 

I'm not sure if we can use `torch._assert_async` either.

abhi-glitchhg: hmm, ig this situation is similar to  https://github.com/pytorch/vision/pull/5776#discussion_r844893521

oke-aditya:
Not sure if I simplified the IoU tests but at least I reduced the code complexity from previous tests. I basically did not want to override a class that handles both the box_area and the IoUs. So I wrote separate test for box_area and then used a class to simplify IoU tests. i feel this keeps code bit readable and clear.

Still need to get rid of _generate().

One way is to use pytest.fixture or other is to use a few global variables.

oke-aditya:
@datumbox @NicolasHug  as discussed with phillip. I will split this PR into 2 Parts. The first part will cleanup the code. The second part will cleanup the tests :smiley: This will make merging easier.

Issue #5913, opened 2022-04-28
`F.pad` in `FixedSizeCrop` needs `int` not `float`


rvandeghen:
### 🐛 Describe the bug

The augment `FixedSizeCrop` in the detection recipe is filled with `mean=(123.0, 117.0, 104.0)` but when we call the padding function:
https://github.com/pytorch/vision/blob/292117405af46e1cdbf2f9ec6eb4752d276bbbb6/references/detection/transforms.py#L356
we should use `int` not `float` as it produces the following error:
```
~/anaconda3/envs/pt1.11-tv0.12-cu10.2/lib/python3.7/site-packages/torchvision/transforms/functional.py in pad(img, padding, fill, padding_mode)
    483         _log_api_usage_once(pad)
    484     if not isinstance(img, torch.Tensor):
--> 485         return F_pil.pad(img, padding=padding, fill=fill, padding_mode=padding_mode)
    486 
    487     return F_t.pad(img, padding=padding, fill=fill, padding_mode=padding_mode)

~/anaconda3/envs/pt1.11-tv0.12-cu10.2/lib/python3.7/site-packages/torchvision/transforms/functional_pil.py in pad(img, padding, fill, padding_mode)
    166             return image
    167 
--> 168         return ImageOps.expand(img, border=padding, **opts)
    169     else:
    170         if isinstance(padding, int):

~/anaconda3/envs/pt1.11-tv0.12-cu10.2/lib/python3.7/site-packages/PIL/ImageOps.py in expand(image, border, fill)
    402     else:
    403         palette = None
--> 404     out = Image.new(image.mode, (width, height), color)
    405     if palette:
    406         out.putpalette(palette.palette)

~/anaconda3/envs/pt1.11-tv0.12-cu10.2/lib/python3.7/site-packages/PIL/Image.py in new(mode, size, color)
   2668         im.palette = ImagePalette.ImagePalette()
   2669         color = im.palette.getcolor(color)
-> 2670     return im._new(core.fill(mode, size, color))
   2671 
   2672 

TypeError: integer argument expected, got float
```

We can either cast the `tuple` of `float` to a tuple of `int` with
```
self.fill = tuple(map(int, fill))
```
in the `__init__` or change the value in the `presets` script.

### Versions

```
[pip3] numpy==1.21.5
[pip3] torch==1.11.0
[pip3] torchvision==0.12.0
[conda] blas                      1.0                         mkl  
[...]

PR #5981, opened 2022-05-10
Change pad value to int

rvandeghen:
Solve the problem of (#5913)
@datumbox

### DIFF
--- a/references/detection/presets.py
+++ b/references/detection/presets.py
@@ -3,7 +3,7 @@ import transforms as T
 
 
 class DetectionPresetTrain:
-    def __init__(self, *, data_augmentation, hflip_prob=0.5, mean=(123.0, 117.0, 104.0)):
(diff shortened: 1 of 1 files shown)


datumbox:
@rvandeghen I don't believe this needs fixing. The original issue with PIL and floats has been already addressed on the latest main branch with #5596. 

I suspect you are facing issues because you are using reference scripts from main (v0.13.0-dev) using v0.12.0 which doesn't contain the above patch. Could you try the latest nightly and confirm?

rvandeghen:
@datumbox Yes my bad.
I will close the PR.
However, I am always building on the reference script for detection for my own research so I always try to improve the whole pipeline.
Is it something that could interest you ?

datumbox:
@rvandeghen This specific issue was addressed with the other PR by making pad for PIL work with float values. So converting the scripts to use ints is no longer needed on the latest version. Nevertheless, we are happy to receive other improvements if you spot further issues. Feel free to open a new issue and let's chat there. :)

PR #5985, opened 2022-05-10
Fix regression on Detection training script

datumbox:
Detection models use variable input sizes and thus shouldn't set `torch.backends.cudnn.benchmark=True`. Doing so causes a significant speed regression.


YosuaMichael (approved):
LGTM

PR #5990, opened 2022-05-11
simplify OnlineResource.load

pmeier:
This simplifies the implementation of `OnlineResource.load` by quite a bit. The only functional difference is that we now only execute the preprocessing after the download. For the admittedly niche situation that the user downloads the data manually, they now also need to preprocess them manually.

In addition, this PR adds a bunch of tests.


pmeier:
> The only functional difference is that we now only execute the preprocessing after the download.

I was mistaken and that also has some implications for us. The current datasets tests are set up in way that they inject a mock of the "raw" downloaded file into the root directory of the dataset. Previously, the loading logic of the resources detected this at runtime and performed the preprocessing. 

Without this we get the failures as seen in this [CI run](https://github.com/pytorch/vision/runs/6388560017). The PCAM dataset expects gzip decompressed resources

https://github.com/pytorch/vision/blob/40a0ab797945e20fda0b35030094fb6741301e07/torchvision/prototype/datasets/_builtin/pcam.py#L108

but with the implementation in this PR, it will get the compressed ones.

I see two ways out here:

1. Reinstate the functionality to always perform the preprocessing if available and we only detect the raw version of the file.
2. Change our test suite to provide the "downloaded" and preprocessed data rather than only the "downloaded" one.

pmeier:
Yes, we can mock 

https://github.com/pytorch/vision/blob/40a0ab797945e20fda0b35030094fb6741301e07/torchvision/prototype/datasets/utils/_resource.py#L125

Still, that will require more changes in the test suite. Currently we use the following idiom

https://github.com/pytorch/vision/blob/40a0ab797945e20fda0b35030094fb6741301e07/test/test_prototype_builtin_datasets.py#L57-L60

If we want to patch in `prepare` we have two options:

1. Use the `mocker` fixture and pass it to `prepare` together with the `test_home`. The patch will automatically be reset for every run of the test.
2. Use `unittest.mock`. The only way this automatically resets is to patch with a context manager. We could do something like 

    ```py
    with dataset_mock.prepare(test_home, config):
        dataset = datasets.load(dataset_mock.name, **config) 
    ```

Another option is to merge the two calls into one like

```py
dataset, mock_info = dataset_mock.load(test_home, config)
```

I've checked, we are always using the two calls together.

Thoughts?

test/builtin_dataset_mocks.py
```diff
@@ -62,27 +62,47 @@ def _parse_mock_info(self, mock_info):
...
+            mock_data_folder.rmdir()
+        else:
+            pass
+            # raise pytest.UsageError(
+            #     (
+            #         f"Dataset '{self.name}' created the following files for {config} in the mock data function, "
+            #         f"but they were not loaded:\n\n"
+            #     )
+            #     + "\n".join(str(file.relative_to(mock_data_folder)) for file in extra_files)
+            # )
```
pmeier, line 103: This was not enforced before and a few datasets are failing. I already fixed this in the open PR #5706. If this PoC is approved, I can migrate them here.

pmeier (commented):
@NicolasHug in d62747962f9ed6a7b0b80849e7c971efabb5d3da I added a PoC for merging the mock data preparation as proposed in https://github.com/pytorch/vision/pull/5990#issuecomment-1124281978.

test/builtin_dataset_mocks.py
```diff
@@ -62,27 +62,47 @@ def _parse_mock_info(self, mock_info):
 
         return mock_info
 
-    def prepare(self, config):
+    def load(self, config):
         # `datasets.home()` is patched to a temporary directory through the autouse fixture `test_home` in
         # test/test_prototype_builtin_datasets.py
         root = pathlib.Path(datasets.home()) / self.name
-        root.mkdir(exist_ok=True)
+        mock_data_folder = root / "__mock__"
```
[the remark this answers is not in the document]
NicolasHug: as we discussed offline, we could also use a more global `home() / cache /` folder where each sub-folder would contain the data for a specfic dataset and config. Looking at our config, look like they should be hashable fairly easily. But we should definitely leave this for another PR.

test/builtin_dataset_mocks.py
```diff
@@ -62,27 +62,47 @@ def _parse_mock_info(self, mock_info):
 
         return mock_info
 
-    def prepare(self, config):
+    def load(self, config):
```
NicolasHug, line 65: this LGTM, but I think it's worth documenting that we need the "temp folder logic" so that the files appear as missing such that the download **+ pre-process** logic gets triggered. Perhaps a permalink to the 3 new lines could be useful

test/test_prototype_datasets_utils.py
```diff
@@ -48,6 +51,197 @@ def test_read_flo(tmpdir):
     torch.testing.assert_close(actual, expected)
 
 
+class TestOnlineResource:
```
NicolasHug, line 54: Do you mind submitting this in another PR so we can merge this one faster?

torchvision/prototype/datasets/utils/_resource.py
```diff
@@ -91,31 +91,30 @@ def load(
...
-
-        # In a first step, we check for a folder with the same stem as the raw file. If it exists, we use it since
-        # extracted files give the best I/O performance. Note that OnlineResource._extract() makes sure that an archive
-        # is always extracted in a folder with the corresponding file name.
+        # Although it looks like we could glob for f"{stem}*" to find the file candidates as well as the folder
+        # candidate simultaneously, that would also pick up other files that share the same prefix. For example, the
+        # test split of the stanford-cars dataset uses the files
+        # - cars_test.tgz
+        # - cars_test_annos_withlabels.mat
+        # Globbing for `"cars_test*"` picks up both.
```
NicolasHug, line 103: Is this comment needed? It feels a bit like "in case you had a bad idea, here's why it's a bad idea"

Also, would our tests fail if we were to use `f"{stem}*"`? If yes I think it's definitely safe to remove the comment

torchvision/prototype/datasets/utils/_resource.py
```diff
@@ -91,31 +91,30 @@ def load(
...
-            file_candidates = {self.download(root, skip_integrity_check=skip_integrity_check)}
-        # If the only thing we find is the raw file, we use it and optionally perform some preprocessing steps.
-        if file_candidates == {path}:
+        if folder_candidate.exists():
+            candidates.add(folder_candidate)
+
+        if not candidates:
+            self.download(root, skip_integrity_check=skip_integrity_check)
             if self._preprocess is not None:
                 path = self._preprocess(path)
```
NicolasHug, line 112: I think this should just be

```py
            if self._preprocess is not None:
                self._preprocess(path)
```

Otherwise this suggests that the path being used isn't the same depending on whether a download happened or not.

torchvision/prototype/datasets/utils/_resource.py
```diff
@@ -91,31 +91,30 @@ def load(
...
-        # Otherwise, we use the path with the fewest suffixes. This gives us the decompressed > raw priority that we
-        # want for the best I/O performance.
         else:
-            path = min(file_candidates, key=lambda path: len(path.suffixes))
+            # We use the path with the fewest suffixes. This gives us the
+            # extracted > decompressed > raw
+            # priority that we want for the best I/O performance.
+            path = min(candidates, key=lambda path: len(path.suffixes))
```
NicolasHug, line 117: With my comment just above I think we can remove the `else` clause and indent this to the left

NicolasHug (commented):
Thanks Philip, I left a few comments, LMK what you think

test/builtin_dataset_mocks.py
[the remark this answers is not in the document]
pmeier: > as we discussed offline, we could also use a more global `home() / cache /` folder

We can, but it can't be within `home()`, since that changes with every call.

torchvision/prototype/datasets/utils/_resource.py
pmeier: The path might change. For example when decompressing `foo.txt.gz` -> `foo.txt`. Same for extraction `foo.tar` -> `foo`.

test/builtin_dataset_mocks.py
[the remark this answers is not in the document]
NicolasHug: I assume we can make the `test_home()` fixture global for the test run?

torchvision/prototype/datasets/utils/_resource.py
NicolasHug: I'm not sure I understand

Is this true?

> the path being used isn't the same depending on whether a download happened or not.

pmeier:
The changes needed to the test suited were factored out in #6010. That will need to be merged first before we continue here.

test/test_prototype_datasets_utils.py
```diff
@@ -48,6 +51,197 @@ def test_read_flo(tmpdir):
...
+
+        resource = self.DummyResource(file_name=file.name)
+
+        dp = resource.load(tmp_path)
+        path, buffer = next(iter(dp))
+
+        assert path == str(file)
+        assert buffer.read().decode() == "decompressed_sentinel"
+
+    def test_priority_extracted_gt_decopressed(self, tmp_path):
```
NicolasHug, line 141:
```suggestion
    def test_priority_extracted_gt_decompressed(self, tmp_path):
```

test/test_prototype_datasets_utils.py
```diff
@@ -48,6 +51,197 @@ def test_read_flo(tmpdir):
...
+        assert isinstance(dp, FileOpener)
+
+        actual = {path: buffer.read().decode() for path, buffer in dp}
+        expected = {
+            path.replace(resource.file_name, resource.file_name.split(".")[0]): content
+            for path, content in files.items()
+        }
+        assert actual == expected
+
+    def test_preprocess_fn(self, tmp_path):
```
NicolasHug, line 208: Could we just parametrize `test_preprocess_decompress` above over `"decompress"`,  and  over a custom `lambda` that does exactly the same as `"decompress"` ?

test/test_prototype_datasets_utils.py
```diff
@@ -48,6 +51,197 @@ def test_read_flo(tmpdir):
...
+    def test_priority_extracted_gt_decopressed(self, tmp_path):
+        archive, _ = self._make_tar(tmp_path, remove=False)
+
+        resource = self.DummyResource(file_name=archive.name)
+
+        dp = resource.load(tmp_path)
+        # If the archive had been selected, this would be a `TarArchiveReader`
+        assert isinstance(dp, FileOpener)
+
+    def test_download(self, tmp_path):
```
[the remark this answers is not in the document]
pmeier: Yup. Do you want me to mock it out then?

test/test_prototype_datasets_utils.py
pmeier: Good point!

test/test_prototype_datasets_utils.py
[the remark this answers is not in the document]
NicolasHug: Or just let `download_fn` have a side effect like updating a nonlocal `download_was_called = True`

test/test_prototype_datasets_utils.py
```diff
@@ -48,6 +51,197 @@ def test_read_flo(tmpdir):
...
+        dp = resource.load(tmp_path)
+        assert files is not None, "`download_fn()` was never called"
+        assert isinstance(dp, FileOpener)
+
+        actual = {path: buffer.read().decode() for path, buffer in dp}
+        expected = {
+            path.replace(resource.file_name, resource.file_name.split(".")[0]): content
+            for path, content in files.items()
+        }
+        assert actual == expected
```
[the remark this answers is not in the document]
pmeier: Of course there are other ways, but I'm not sure they are easier. We can't create the archive upfront, because in that case we would never trigger the preprocessing. One option would be to create the data upfront in a temporary directory and move it to `root` inside the download function similar to what we are doing with the actual resource loading in our dataset tests (#6010).

Given that you were ok with `nonlocal` in https://github.com/pytorch/vision/pull/5990#discussion_r873553403, I don't think it will be much simpler. You choose.

torchvision/prototype/datasets/utils/_resource.py
pmeier: I'll check after #6010 is merged. IIRC, the test only failed sometimes, since `pathlib.Path().glob()` is non-deterministic and sometimes returned the correct path first, but not always.

test/test_prototype_datasets_utils.py
```diff
@@ -48,6 +51,187 @@ def test_read_flo(tmpdir):
     torch.testing.assert_close(actual, expected)
 
 
+# This cannot be defined on the TestOnlineResource class, since it is used in a `@pytest.mark.parametrize` decorator on
+# a method on said class
+def _decompress_gz(path):
```
NicolasHug, line 56: We don't have a decompressing util already written somewhere that we can use for that?

torchvision/prototype/datasets/utils/_resource.py
NicolasHug: I had not realized this comment was also about the `folder_candidate` logic and not just about the line below. I agree it's worth keeping it now, thanks for writing it

torchvision/prototype/datasets/utils/_resource.py
```diff
@@ -32,7 +32,7 @@ def __init__(
         *,
         file_name: str,
         sha256: Optional[str] = None,
-        preprocess: Optional[Union[Literal["decompress", "extract"], Callable[[pathlib.Path], pathlib.Path]]] = None,
+        preprocess: Optional[Union[Literal["decompress", "extract"], Callable[[pathlib.Path], None]]] = None,
```
NicolasHug, line 35: For my own education, why did we need to specify `None` in the annotation here? I assume `Optional[]` would have been enough - is it just to be more explicit about it?

NicolasHug (approved):
Thanks @pmeier , LGTM

torchvision/prototype/datasets/utils/_resource.py
pmeier: It becomes clearer when we let `black` explode the annotation:

```suggestion
        preprocess: Optional[
            Union[
                Literal["decompress", "extract"],
                Callable[[pathlib.Path], None],
            ]
        ] = None,
```

I only changed the return type of the callable from `pathlib.Path` to `None` since we refactored the loading logic that the return is no longer needed.

Still, in general you are right. [`Optional[Foo]` is equivalent to `Union[None, Foo]`](https://github.com/python/cpython/blob/93fc14933b8605c8df23073574048408df61b538/Lib/typing.py#L690-L693). Plus, [`Optional[Union[Foo, Bar]]` can be flattened to `Union[None, Foo, Bar]`](https://github.com/python/cpython/blob/93fc14933b8605c8df23073574048408df61b538/Lib/typing.py#L657-L659). It is just my personal preference to be explicit about `Optional`.

test/test_prototype_datasets_utils.py
pmeier: We have, but it is the same we use for `preprocess="decompress"`:

https://github.com/pytorch/vision/blob/d9a69506109cb970c7568c27cbeda1f9ffb7ad70/torchvision/datasets/utils.py#L343

https://github.com/pytorch/vision/blob/d9a69506109cb970c7568c27cbeda1f9ffb7ad70/torchvision/prototype/datasets/utils/_resource.py#L58-L60

If we are ok in re-using `_decompress`, I can remove this function.

PR #5991, opened 2022-05-11
remove SampleQuery from prototype datasets

pmeier:
This is a remnant from an ancient implementation. The functionality was moved into the prototype transforms and can be found here:

https://github.com/pytorch/vision/blob/64b1e279d7963c923bd453de07589a25cb6e8d03/torchvision/prototype/utils/_internal.py#L153


datumbox (approved):
@pmeier Nice, clip clip clip ✂️

NicolasHug (approved):
Thanks Philip

PR #5993, opened 2022-05-11
Indicate which weights correspond to pretrained=True in docs

NicolasHug:
As per @datumbox 's offline discussion, `pretrained=True` is equivalent to the V1 weights. This PR adds this information in the docs on all V1 weights.

image


datumbox (approved):
LGTM, great idea.

NicolasHug:
> The easiest, none hacky way I can think of is to keep a mapping hidden somehwere until the deprecation takes place.

I agree. Not sure when this will be available, so I'll close this PR for now. But I think it would be valuable info to document, so let's just keep it in the back of our mind

PR #5994, opened 2022-05-11
Clarify TypeError message

Gouvernathor:
To have a more clearer message than `TypeError: tensor or list of tensors expected, got <class 'list'>`, when having (for example) a list of numpy arrays.
Even better would be to parse the list and return the type of the first non-tensor. If you don't mind the code difference, I'll program that.

### DIFF
--- a/torchvision/utils.py
+++ b/torchvision/utils.py
@@ -60,7 +60,7 @@ def make_grid(
         if isinstance(tensor, list):
             typ = "a list not containing only tensors"
         else:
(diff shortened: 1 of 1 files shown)


Gouvernathor:
Yeah, I need to wrap the type() in a repr() so it shuts up (even though the f-string formatting does it by itself).

Gouvernathor:
Please consider 5447 before this one, I find it better.

datumbox:
superseded by #5997

PR #5995, opened 2022-05-11
Remove comment which is rendered as css class

oke-aditya:
image

The comment has unintentional side effect of appearing in css class over generated tables in html pages. Although it doesn't break anything, it could hurt sometime later.

cc @NicolasHug


oke-aditya:
To verify you can visit 
https://pytorch.org/vision/main/models_new.html 
right click 
view page source and 
see line number 382

While do the same on the current build you can do the same through Circle CI artifacts

datumbox (approved):
Nice catch, thanks @oke-aditya

NicolasHug:
I had noticed this, but since it doesn't have any effect I kept it.  I still think the comment is useful to provide context to the reader. Without it, we will definitely forget what this file does eventually. If we remove it, the docs will look terrible and we might not notive before it's too late.

@oke-aditya can you please try to see if we can add this comment on the line above? If not, I'd prefer referting this PR.

oke-aditya:
It's bit bad both the ways. Keeping it means we are injecting CSS which unnecassary (slowing the website) + say torch theme creates a CSS class that uses this. It would mean we are broken as we will override that CSS. I know the comment is crucial but having comment that injects code un intentionally is equally bad.

NicolasHug:
I'd really like to get this comment back one way or another. Do you mind trying adding it back in the file?

oke-aditya:
Lemme try adding it above so that it will not go inside CSS.

NicolasHug:
If it doesn't work, we can put back the comment as 

`#_Necessary_for_the_table_generated_by_autosummary_to_look_decent`

This will never correspond to any css class and we'll be safe. Not amazing, but better than not comment at all.

datumbox:
I think it's possible to fix the HTML and keep the comment. See below:

> The configuration file consists of sections, lead by a "[section]" header and followed by "name: value" entries, with continuations in the style of [RFC 822](http://www.rfc-editor.org/rfc/rfc822.txt); "name=value" is also accepted. Note that leading whitespace is removed from values. ... Lines beginning with "#" or ";" are ignored and may be used to provide comments.

From [here](https://docutils.sourceforge.io/0.6/docs/user/config.html).

PR #5997, opened 2022-05-11
Even better TypeError message than #5994

Gouvernathor:


datumbox (approved):
LGTM, thanks @Gouvernathor.

torchvision/utils.py
```diff
@@ -56,8 +56,13 @@ def make_grid(
...
         _log_api_usage_once(make_grid)
-    if not (torch.is_tensor(tensor) or (isinstance(tensor, list) and all(torch.is_tensor(t) for t in tensor))):
-        raise TypeError(f"tensor or list of tensors expected, got {type(tensor)}")
+    if not torch.is_tensor(tensor):
+        if isinstance(tensor, list):
+            for t in tensor:
+                if not torch.is_tensor(t):
+                    raise TypeError(f"tensor or list of tensors expected, got a list containing {type(t)}")
+        else:
+            raise TypeError(f"tensor or list of tensors expected, got {type(tensor)}")
```
NicolasHug, line 65: Thanks for the PR @Gouvernathor , 
The changes look good but the condition is now more complex. Perhaps we can simplify it to

```py
    if not torch.is_tensor(tensor):
        if isinstance(tensor, list) and any(not torch.is_tensor(t) for t in tensor):
                raise TypeError(f"tensor or list of tensors expected, got a list containing {type(t)}")
        else:
            raise TypeError(f"tensor or list of tensors expected, got {type(tensor)}")
```

Would you mind submitting a PR to simplify this?

PR #5998, opened 2022-05-12
rely on patched datasets home rather than passing it around

pmeier:
While working on #5990, I realized we don't need to pass the `test_home` to our mock data generation, since we later already rely on the patched version. With this PR, the preparation as well as the loading now rely on the patched home.

Note that this does not conflict with anything in #5990.


test/test_prototype_builtin_datasets.py
```diff
@@ -25,9 +25,10 @@ def extract_datapipes(dp):
     return get_all_graph_pipes(traverse(dp, only_datapipe=True))
 
 
-@pytest.fixture
+@pytest.fixture(autouse=True)
```
pmeier, line 28: Although this patch is now applied by default, if required, a test can still request this fixture the normal way.

test/test_prototype_builtin_datasets.py
```diff
@@ -25,9 +25,10 @@ def extract_datapipes(dp):
...
 def test_home(mocker, tmp_path):
     mocker.patch("torchvision.prototype.datasets._api.home", return_value=str(tmp_path))
+    mocker.patch("torchvision.prototype.datasets.home", return_value=str(tmp_path))
```
pmeier, line 31: We need this extra patch since the mock data generation is calling the public interface `datasets.home()` whereas the loading does not:

https://github.com/pytorch/vision/blob/40a0ab797945e20fda0b35030094fb6741301e07/torchvision/prototype/datasets/_api.py#L4

test/builtin_dataset_mocks.py
```diff
@@ -62,8 +62,8 @@ def _parse_mock_info(self, mock_info):
 
         return mock_info
 
-    def prepare(self, home, config):
-        root = home / self.name
+    def prepare(self, config):
+        root = pathlib.Path(datasets.home()) / self.name
```
NicolasHug, line 66:
```suggestion
        root = pathlib.Path(datasets.home()) / self.name  # Note: home() is patched by the test_home() autouse fixture
```

test/test_prototype_builtin_datasets.py
NicolasHug: Q: I'm always incredibly confused by the behavior of `mocker.patch`.

Is it because it patches names (references to objects) instead of patching objects?

NicolasHug (approved):
Thanks Philip

test/test_prototype_builtin_datasets.py
pmeier: It patches attributes. Not sure if that is what you mean by "names". The above is roughly equivalent to

```py
torchvision.prototype.datasets._api.home = lambda *_, **_: str(tmp_path)
torchvision.prototype.datasets.home = lambda *_, **_: str(tmp_path)
```

Thus, it patches the attribute `home` on the two modules rather than the actual function.

PR #6003, opened 2022-05-12
Clean up model documentation

datumbox:
This PR:
- Removes the old "minimum input size" info from the docstring
- Removes "currently only XYZ weights available" from the main page of MNASNET. This information is encoded in the model builder pages and might become stale on the future if the text is left there.
- Fixes the text of `wide_resnet101_2` which still mentioned the `50` variant
- Makes the license URLs display as links (similar to recipes).
- Clarifies the order of dims of `min_size`.
- Removes lengthy `keypoint_names` from meta-table.

image

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -349,7 +349,7 @@ def inject_weight_metadata(app, what, name, obj, options, lines):
                 lines += [custom_docs, ""]
 
             for k, v in meta_with_metrics.items():
-                if k == "recipe":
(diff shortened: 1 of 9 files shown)


NicolasHug (approved):
Thanks!

EDIT oops didn't realize this was draft, will take another look later sorry

NicolasHug (approved):
Still LGTM! Thanks @datumbox

PR #6009, opened 2022-05-13
Add preprocessing information on the Weights documentation

datumbox:
Each weight comes with a human-readable description of how the preprocessing happens during inference:
image

### DIFF
--- a/torchvision/transforms/_presets.py
+++ b/torchvision/transforms/_presets.py
@@ -25,6 +25,9 @@ class ObjectDetection(nn.Module):
             img = F.pil_to_tensor(img)
         return F.convert_image_dtype(img, torch.float)
 
+    def __repr__(self) -> str:
+        return self.__class__.__name__ + "()"
+
 
 class ImageClassification(nn.Module):
     def __init__(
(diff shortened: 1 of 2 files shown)


torchvision/transforms/_presets.py
```diff
@@ -37,21 +43,38 @@ def __init__(
...
-        self._crop_size = [crop_size]
-        self._size = [resize_size]
-        self._mean = list(mean)
-        self._std = list(std)
-        self._interpolation = interpolation
+        self.crop_size = [crop_size]
+        self.resize_size = [resize_size]
+        self.mean = list(mean)
+        self.std = list(std)
+        self.interpolation = interpolation
```
datumbox, line 50: Just removing the `_` because the whole class is private and marking the individual features seems an overkill. Plus it messes up the `__repr__` strings.

torchvision/transforms/_presets.py
```diff
@@ -25,6 +25,12 @@ def forward(self, img: Tensor) -> Tensor:
             img = F.pil_to_tensor(img)
         return F.convert_image_dtype(img, torch.float)
 
+    def __repr__(self) -> str:
+        return self.__class__.__name__ + "()"
+
+    def describe(self) -> str:
```
datumbox, line 31: This is the "human-readable" version. Inspired from Pandas' `describe()` method which summarizes dataframes. It provides a high-level description of what the transform is actually doing.

torchvision/transforms/_presets.py
```diff
@@ -25,6 +25,12 @@ def forward(self, img: Tensor) -> Tensor:
             img = F.pil_to_tensor(img)
         return F.convert_image_dtype(img, torch.float)
 
+    def __repr__(self) -> str:
+        return self.__class__.__name__ + "()"
```
datumbox, line 29: Available for a more "object-serialization" description.

torchvision/transforms/_presets.py
```diff
@@ -79,18 +102,35 @@ def forward(self, vid: Tensor) -> Tensor:
...
+        format_string += "\n)"
+        return format_string
+
+    def describe(self) -> str:
+        return (
+            f"The video frames are resized to ``resize_size={self.resize_size}`` using ``interpolation={self.interpolation}``, "
+            f"followed by a central crop of ``crop_size={self.crop_size}``. Then the values are rescaled to "
+            f"``[0.0, 1.0]`` and normalized using ``mean={self.mean}`` and ``std={self.std}``."
+        )
+
```
datumbox, line 133: Some repetition here. I could move to a separate class and inherit but thought it's an overkill to introduce inheritance here. Happy to do it though

datumbox (commented):
Some comments:

torchvision/transforms/_presets.py
NicolasHug: I agree, no need to inherit

NicolasHug (approved):
Thanks @datumbox , I think it looks nice!

Maybe as a follow-up (very low pri) we could check if all the descriptions are the same for a given weight enum. And if they're all the same, then we would just need to document it once at the top, like 

> all weights below follow the following pre-processing: ...

Just a thought, definitely not worth bothering with that now.

torchvision/transforms/_presets.py
datumbox: Sounds good. We can review changing this on the entire codebase on a follow up.

PR #6010, opened 2022-05-13
Merge mock data preparation and dataset logic in prototype tests

pmeier:
This was extracted out of #5990.


test/builtin_dataset_mocks.py
```diff
@@ -62,27 +61,49 @@ def _parse_mock_info(self, mock_info):
...
+            src = tmp_mock_data_folder / resource.file_name
+            if not src.exists():
+                raise pytest.UsageError(
+                    f"Dataset '{self.name}' requires the file {resource.file_name} for {config}"
+                    f"but it was not created by the mock data function."
+                )
 
-        mock_info = self._parse_mock_info(self.mock_data_fn(root, config))
+            dst = root / resource.file_name
+            shutil.move(str(src), str(root))
```
NicolasHug, line 87: Should we delete the `__mock__` dir if it's empty, which should happen after all resources have been "downloaded" ?

I agree that `__mock__` should not correspond to any specific dataset folder, but what if a dataset parses all folders at its root and expects them to contain images?

test/builtin_dataset_mocks.py
```diff
@@ -799,8 +822,11 @@ def add_bndbox(obj):
     def generate(cls, root, *, year, trainval):
         archive_folder = root
         if year == "2011":
-            archive_folder /= "TrainVal"
-        data_folder = archive_folder / "VOCdevkit" / f"VOC{year}"
+            archive_folder = root / "TrainVal"
+            data_folder = archive_folder / "VOCdevkit"
+        else:
+            archive_folder = data_folder = root / "VOCdevkit"
+        data_folder = data_folder / f"VOC{year}"
```
NicolasHug, line 829: Is this related?

test/builtin_dataset_mocks.py
```diff
@@ -1091,8 +1117,10 @@ def _make_ann_file(path, num_examples, class_idx):
                     }
                 )
 
+    archive_folder = root / "GTSRB"
+
```
NicolasHug, line 1121: Is this related?

test/builtin_dataset_mocks.py
```diff
@@ -514,20 +535,11 @@ def imagenet(root, config):
...
-            images_meta.append(dict(file_name=path.name, id=int(path.stem), width=width, height=height))
-
-        make_zip(root, f"{name}.zip")
-
-        return images_meta
+    def _make_images_meta(cls, *, num_samples):
+        return [
+            dict(file_name=f"{idx:012d}.jpg", id=idx, width=width, height=height)
+            for idx, (height, width) in enumerate(torch.randint(3, 11, size=(num_samples, 2), dtype=torch.int).tolist())
+        ]
```
NicolasHug, line 542: This is only used once I think. Should we inline it?

test/builtin_dataset_mocks.py
```diff
@@ -596,16 +608,27 @@ def generate(
...
     ):
         annotations_dir = root / "annotations"
         annotations_dir.mkdir()
 
-        for split in ("train", "val"):
-            config_name = f"{split}{year}"
+        for split_ in ("train", "val"):
+            config_name = f"{split_}{year}"
+
+            images_meta = cls._make_images_meta(num_samples=num_samples)
```
NicolasHug, line 621: It's a bit strange to create the meta dict and its associated file for both splits while only create the image folder for one split.
I understand the meta dict is supposed to contain meta for both train and val, but right now the dict will contain meta for both splits while there will only be one image folder. Should we remove the `for loop` here?

NicolasHug (commented):
Thanks @pmeier , I made some comment below

test/builtin_dataset_mocks.py
pmeier: Yes this is related if we want to enforce that no extra files are generated. See https://github.com/pytorch/vision/pull/5990#pullrequestreview-970563847 for details.

test/builtin_dataset_mocks.py
pmeier: > I understand the meta dict is supposed to contain meta for both train and val, but right now the dict will contain meta for both splits while there will only be one image folder.

Unfortunately, that is how the dataset is structured:

```shell
$ ls
.rw-rw-r-- 253M root   31 May  2021 annotations_trainval2014.zip
.rw-rw-r-- 253M root   19 Oct  2021 annotations_trainval2017.zip
.rw-rw-r--  14G root   11 May  2021 train2014.zip
.rw-rw-r--  19G root   31 Oct  2021 train2017.zip
.rw-rw-r-- 6.6G root   31 Oct  2021 val2014.zip
.rw-rw-r-- 816M root   31 Oct  2021 val2017.zip
$ unzip -l annotations_trainval2014.zip
Archive:  annotations_trainval2014.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
332556225  2017-09-01 20:56   annotations/instances_train2014.json
160682675  2017-09-01 20:57   annotations/instances_val2014.json
170733465  2017-09-01 20:59   annotations/person_keypoints_train2014.json
 81637509  2017-09-01 20:59   annotations/person_keypoints_val2014.json
 66782097  2017-09-01 20:59   annotations/captions_train2014.json
 32421077  2017-09-01 20:59   annotations/captions_val2014.json
---------                     -------
844813048                     6 files
```

test/builtin_dataset_mocks.py
pmeier: Same as above.

test/builtin_dataset_mocks.py
pmeier: This is also why we only create the annotations archive after the loop, whereas the image archive is created inside the loop:

https://github.com/pytorch/vision/blob/f176fb0dfc302a4cc58c374e59d2af6898629a33/test/builtin_dataset_mocks.py#L605-L615

NicolasHug: OK, fair enough. Just for my own understanding: everythign would still work if we removed the for loop, right?

test/builtin_dataset_mocks.py
NicolasHug: I'm still not sure I understand how these changes are needed. If anything, the changes seem "looser" than what we previously had.

The main diff seems to be

```
make_zip(root, "GTSRB-Training_fixed.zip", root / "GTSRB")  # now
```

vs

```
make_zip(root, "GTSRB-Training_fixed.zip", root / "GTSRB" / "Training" )  # before
```

test/builtin_dataset_mocks.py
pmeier: Yes, the test will still pass, but as explained in our guidelines, it would still pass if we remove 

https://github.com/pytorch/vision/blob/f176fb0dfc302a4cc58c374e59d2af6898629a33/torchvision/prototype/datasets/_builtin/coco.py#L161

On the original data, removing this line would fail.

test/builtin_dataset_mocks.py
pmeier: We rely on the implicit behavior of `make_zip` to clean up for us:

https://github.com/pytorch/vision/blob/f176fb0dfc302a4cc58c374e59d2af6898629a33/test/datasets_utils.py#L887-L891

It will remove all files and folders that were added to the archive. By specifying `root / "GTSRB" / "Training"` as source for the files and folders to add, the folder `GTSRB` is not part of the removal. Thus, after the creation of the archive, an empty `GTSRB` folder is left in the root directory. This trips the new check.

By specifying `root / "GTSRB"` the folder is part of the removal schedule and we don't have to do it manually like

```py
make_zip(root, "GTSRB-Training_fixed.zip", root / "GTSRB" / "Training" )
(root / "GTSRB").rmdir()
```

test/builtin_dataset_mocks.py
```diff
@@ -62,27 +61,51 @@ def _parse_mock_info(self, mock_info):
...
-                f"for {config}, but they were not created by the mock data function."
+                (
+                    f"Dataset '{self.name}' created the following files for {config} in the mock data function, "
+                    f"but they were not loaded:\n\n"
+                )
+                + "\n".join(str(file.relative_to(tmp_mock_data_folder)) for file in extra_files)
             )
 
-        return mock_info
+        tmp_mock_data_folder.rmdir()
```
NicolasHug, line 106: Would make more sense to remove it in the mock instead? I.e. remove if there are no more files in `__mock__`. This way we're sure that it will not affect `datasets.load()` whatsoever.

NicolasHug (approved):
Thanks Philip, last comment but LGTM regardless

test/builtin_dataset_mocks.py
pmeier: You mean inside `patched_download`? Not sure what the point would be. We could only remove after the directory is empty. Thus, in case there are multiple resources to load, only the last one would be loaded without `__mock__` in `root`. In case the directory somehow interferes with the loading logic, it will do so anyway.

NicolasHug: Hm good point, thanks

PR #6013, opened 2022-05-13
[WIP] Add examples of Multi-weight support + model usage

datumbox:

### DIFF
--- a/docs/source/models_new.rst
+++ b/docs/source/models_new.rst
@@ -13,7 +13,7 @@ Models and pre-trained weights - New
 The ``torchvision.models`` subpackage contains definitions of models for addressing
 different tasks, including: image classification, pixelwise semantic
 segmentation, object detection, instance segmentation, person
-keypoint detection, video classification, and optical flow.
+keypoint detection, video classification and optical flow.
 
 .. note ::
     Backward compatibility is guaranteed for loading a serialized 
@@ -56,6 +56,35 @@ weights:
    models/vision_transformer
    models/wide_resnet
 
+|
+
+Here is an example of how to use the pre-trained image classification models:
+
+.. code:: python
+
+    from torchvision.io import read_image
+    from torchvision.models import resnet50, ResNet50_Weights
+
+    img = read_image("test/assets/encode_jpeg/grace_hopper_517x606.jpg")
+
+    # Step 1: Initialize model
+    weights = ResNet50_Weights.DEFAULT
+    model = resnet50(weights=weights)
+    model.eval()
+
+    # Step 2: Initialize the inference transforms
+    preprocess = weights.transforms()
+
+    # Step 3: Apply inference preprocessing transforms
+    batch = preprocess(img).unsqueeze(0)
+    prediction = model(batch).squeeze(0).softmax(0)
+
+    # Step 4: Use the model and print the predicted category
+    class_id = prediction.argmax().item()
+    score = prediction[class_id].item()
+    category_name = weights.meta["categories"][class_id]
+    print(f"{category_name}: {100 * score}%")
+
 
 Table of all available classification weights
 ---------------------------------------------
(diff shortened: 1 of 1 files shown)


docs/source/models_new.rst
```diff
@@ -191,6 +357,38 @@ pre-trained weights:
...
+
+Here is an example of how to use the pre-trained video classification models:
+
+.. code:: python
+
+
+    from torchvision.io.video import read_video
+    from torchvision.models.video import r3d_18, R3D_18_Weights
+
+    vid, _, _ = read_video("test/assets/videos/v_SoccerJuggling_g23_c01.avi")
```
datumbox, line 370: @bjuncek Apparently this emits a vague deprecation warning. Does it make sense to remove the warning given that we are yet uncertain what we will do with the specific API? Or perhaps I should modify my example to avoid the warnings?

docs/source/models_new.rst
```diff
@@ -24,6 +24,48 @@ keypoint detection, video classification, and optical flow.
...
+
+    from torchvision.models import resnet50, ResNet50_Weights
+
+    # Old weights with accuracy 76.130%
+    resnet50(weights=ResNet50_Weights.IMAGENET1K_V1)
+
+    # New weights with accuracy 80.858%
+    resnet50(weights=ResNet50_Weights.IMAGENET1K_V2)
+
+    # Best available weights (currently alias for IMAGENET1K_V2)
```
NicolasHug, line 41: Do we have a more obvious way of documenting that `DEFAULT` may be aliased to other weights in the future?


```suggestion
    # Best available weights (currently alias for IMAGENET1K_V2)
    # Note that these weights may change across versions
```

docs/source/models_new.rst
```diff
@@ -56,6 +98,35 @@ weights:
...
+    # Step 3: Apply inference preprocessing transforms
+    batch = preprocess(img).unsqueeze(0)
+
+    # Step 4: Use the model and print the predicted category
+    prediction = model(batch).squeeze(0).softmax(0)
+    class_id = prediction.argmax().item()
+    score = prediction[class_id].item()
+    category_name = weights.meta["categories"][class_id]
+    print(f"{category_name}: {100 * score}%")
+
```
NicolasHug, line 129: Super Nit:

```suggestion
    print(f"{category_name}: {100 * score:.1f}%")
    # prints: bow tie: 14.4%
```

Although thinking about it the classification result looks dodgy lol. Bow tie?!

docs/source/models_new.rst
```diff
@@ -78,6 +149,35 @@ pre-trained weights:
    models/googlenet_quant
    models/mobilenetv2_quant
 
+|
+
+Here is an example of how to use the pre-trained quantized image classification models:
```
NicolasHug, line 154: I wonder if we should just refer to the snippet above instead, and mention to pass `quantize=True`? This seems to be the only difference, apart from the imports.

docs/source/models_new.rst
```diff
@@ -130,6 +261,41 @@ weights:
...
+
+    # Step 4: Use the model and visualize the prediction
+    prediction = model(batch)[0]
+    labels = [weights.meta["categories"][i] for i in prediction["labels"]]
+    box = draw_bounding_boxes(img, boxes=prediction["boxes"],
+                              labels=labels,
+                              colors="red",
+                              width=4, font_size=30)
+    im = to_pil_image(box.detach())
+    im.show()
```
NicolasHug, line 297: I'm getting 

```
RuntimeError: "qnms_kernel" not implemented for 'Float'
```

But I haven't compiled recently, maybe that's why

NicolasHug (approved):
Thanks @datumbox I made some minor suggestions, LGTM regardless.

I think the snippets describing the different usages are useful, but I wonder if it would make more sense to move them into a gallery example (maybe exampleS)?

Definitely fine as-is though, we can improve on that later (potentially after the release) if we have more bandwidth.

docs/source/models_new.rst
```diff
@@ -13,7 +13,7 @@ Models and pre-trained weights - New
 The ``torchvision.models`` subpackage contains definitions of models for addressing
 different tasks, including: image classification, pixelwise semantic
 segmentation, object detection, instance segmentation, person
-keypoint detection, video classification, and optical flow.
+keypoint detection, video classification and optical flow.
```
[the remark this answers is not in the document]
datumbox: OK Mr Oxford. Reverted.

docs/source/models_new.rst
datumbox: it detects her bow tie lol

docs/source/models_new.rst
datumbox: I'm OK giving a second example given we have a different namespace. Hopefully on the future this will be deprecated and move all quantization of models on main model builders (with FX quant)

docs/source/models_new.rst
bjuncek: Yeah, feel free to remove it.
I'm in the process of re-writing the video-module to have a more sensical stucture

datumbox: See #6056

Issue #5987, opened 2022-05-11
Revamping our Quantized models docs


NicolasHug:
We're re-writing our models docs to make them clearer, simpler, and to properly document the [upcoming multi-weight API](https://pytorch.org/blog/introducing-torchvision-new-multi-weight-support-api/). This issue is about **adding docs for the Quantized classification models.**

Our latest new docs are currently [here](https://output.circle-artifacts.com/output/job/8ce072c6-4b16-4aa9-9055-b6f19635d4af/artifacts/0/docs/models_new.html#quantized-models) (this link is likely outdated by the time you look at it, but it doesn't matter; the skeleton is there). We created a separate section that will eventually be merged into the main one. We have documented a few models, but most of them are still missing. The list of models that still need docs is listed below. If you'd like to participate, please comment below with a message saying "I'm working on XYZ" where XYZ is a model, so that others don't pick the same as you do. To keep things simple, please submit one PR per model, but feel free to contribute more than one model.

### How to write docs for a model

Note: below are detailed instructions. This makes it look more complicated than it actually is. Don't be scared!

A great place to start is to look a the changes in [this PR that documents SqueezeNet](https://github.com/pytorch/vision/pull/5832/files). You'll need to do exactly the same for your model:

- Create a new `.rst` file in https://github.com/pytorch/vision/tree/main/docs/source/models. The file should [look like this](https://raw.githubusercontent.com/pytorch/vision/main/docs/source/models/squeezenet.rst), with a link to the original paper, and a list of the corresponding model builders. It should also mention the base model class and link to the `.py` file where it is defined.
- Update the list in https://github.com/pytorch/vision/blame/main/docs/source/models_new.rst to link to this new file (without the `.rst` suffix). Please keep the list alphabetically sorted
- Update the docstring of each new model builder, similarly to [this one](https://github.com/pytorch/vision/blob/64b1e279d7963c923bd453de07589a25cb6e8d03/torchvision/models/quantization/googlenet.py#L144-L172).
[...]

PR #6016, opened 2022-05-14
Revamp docs for Quantized MobileNetV3

abhi-glitchhg:
Related to #5987 

cc : @NicolasHug

### DIFF
--- a/torchvision/models/quantization/mobilenetv3.py
+++ b/torchvision/models/quantization/mobilenetv3.py
@@ -194,18 +194,33 @@ def mobilenet_v3_large(
     **kwargs: Any,
 ) -> QuantizableMobileNetV3:
     """
-    Constructs a MobileNetV3 Large architecture from
-    `"Searching for MobileNetV3" <https://arxiv.org/abs/1905.02244>`_.
+    MobileNetV3 model from
+    `Searching for MobileNetV3 <https://arxiv.org/abs/1905.02244>`_.
 
-    Note that quantize = True returns a quantized model with 8 bit
-    weights. Quantized models only support inference and run on CPUs.
-    GPU inference is not yet supported
+    .. note::
+        Note that `quantize = True` returns a quantized model with 8 bit
+        weights. Quantized models only support inference and run on CPUs.
+        GPU inference is not yet supported
 
     Args:
-        weights (MobileNet_V3_Large_QuantizedWeights or MobileNet_V3_Large_Weights, optional): The pretrained
-            weights for the model
-        progress (bool): If True, displays a progress bar of the download to stderr
-        quantize (bool): If True, returns a quantized model, else returns a float model
+        weights (:class:`~torchvision.models.quantization.MobileNet_V3_Large_QuantizedWeights` or :class:`~torchvision.models.MobileNet_V3_Large_Weights`, optional): The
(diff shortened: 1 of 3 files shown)


torchvision/models/quantization/mobilenetv3.py
```diff
@@ -194,18 +194,33 @@ def mobilenet_v3_large(
...
+    MobileNetV3 model from
+    `Searching for MobileNetV3 <https://arxiv.org/abs/1905.02244>`_.
 
-    Note that quantize = True returns a quantized model with 8 bit
-    weights. Quantized models only support inference and run on CPUs.
-    GPU inference is not yet supported
+    .. note::
+        Note that `quantize = True` returns a quantized model with 8 bit
+        weights. Quantized models only support inference and run on CPUs.
+        GPU inference is not yet supported
```
abhi-glitchhg, line 203: I think we should highlight this note in other models' docstrings too.

abhi-glitchhg:
[generated docs](https://output.circle-artifacts.com/output/job/42b189ba-c859-4117-8ec4-049de90f35f6/artifacts/0/docs/models/mobilenetv3_quant.html?highlight=quantized) for mobilenet v3

torchvision/models/quantization/mobilenetv3.py
```diff
@@ -194,18 +194,33 @@ def mobilenet_v3_large(
     **kwargs: Any,
 ) -> QuantizableMobileNetV3:
     """
-    Constructs a MobileNetV3 Large architecture from
-    `"Searching for MobileNetV3" <https://arxiv.org/abs/1905.02244>`_.
+    MobileNetV3 model from
```
datumbox, line 197: Why did you make this change? Clarifying that this is the variant ("Large" in this case) is something we do in all similar docs of models.

torchvision/models/quantization/mobilenetv3.py
```diff
@@ -194,18 +194,33 @@ def mobilenet_v3_large(
...
+    `Searching for MobileNetV3 <https://arxiv.org/abs/1905.02244>`_.
 
-    Note that quantize = True returns a quantized model with 8 bit
-    weights. Quantized models only support inference and run on CPUs.
-    GPU inference is not yet supported
+    .. note::
+        Note that ``quantize = True`` returns a quantized model with 8 bit
+        weights. Quantized models only support inference and run on CPUs.
+        GPU inference is not yet supported
```
NicolasHug, line 203: Thanks for making this a `note`, I agree this looks better this way. Would you like to open an issue to address the rest of the models? Or perhaps you'd like to submit a PR directly?

torchvision/models/quantization/mobilenetv3.py
NicolasHug, line 197:
```suggestion
    MobileNetV3 (Large) model from
```

NicolasHug (approved):
Looks great @abhi-glitchhg , thank you!

I adressed @datumbox 's comment above directly, so I'll merge now. <o/

torchvision/models/quantization/mobilenetv3.py
abhi-glitchhg: On my way to submit PR :)

PR #5009, opened 2021-11-30
NOMRG Current CI status

NicolasHug:
This is just a dummy PR to see which jobs are green and which aren't
(Useful to decide whether failures are related for newly-created PRs)

Feel free to merge with `main` every now and then to keep it up to date

Please avoid commenting about new failures below. If there are failures, it's best to open a new issue and reference this PR in the issue. Also please do not push fixes to this branch, open a new PR instead.

(same as https://github.com/pytorch/vision/pull/3808, but reopening because https://github.com/pytorch/vision/pull/3808#issuecomment-982468042)

PR #6017, opened 2022-05-14
[NOMERGE] CI test

datumbox:


pmeier:
@datumbox FYI: there is #5009, which Nicolas and I are using for this purpose for a while now. I'll close this so we don't have multiple PRs doing the same thing.

PR #6021, opened 2022-05-16
Highlighting the Note in the docs

abhi-glitchhg:
Addressing this  https://github.com/pytorch/vision/pull/6016#discussion_r873476012

Also, I added the note in Resnet, Resnext, and Shufflenet docstrings(it wasn't there before).


NicolasHug (approved):
Thanks a lot @abhi-glitchhg !

Issue #6020, opened 2022-05-16
Restructuring our tests


oke-aditya:
# Motivation

Our tests are getting a bit lengthy and considering we are adding new features, we are actually placing multiple tests in single file.
This leads to problems as the tests become  harder to refactor and harder to read, manage. Standalone files for tests or grouped according to purpose are easier to anaylse and debug as well.

Our longest test files

Copy paste-able command
```
find test -name "*.py" | xargs -I {} wc -l {} | sort -nr | head -n 10
```

```
oke@psycon:~/Aditya/PyTorch/vision$ find test -name "*.py" | xargs -I {} wc -l {} | sort -nr | head -n 10
2675 test/test_datasets.py
2242 test/test_transforms.py
1535 test/builtin_dataset_mocks.py
1528 test/test_ops.py
1343 test/test_functional_tensor.py
1339 test/test_prototype_transforms_functional.py
1254 test/test_video_reader.py
982 test/test_transforms_tensor.py
954 test/datasets_utils.py
942 test/test_models.py
```

Many splits are possible that can make the tests smaller and easier to manage

```
E.g. 

test_models.py
-----  test_classification_models.py
----- test_detection_models.py
----- test_video_models.py

test_ops.py
------ test_post_processors.py
------ test_losses.py
------ test_layers.py

```

Possibilities are many, we can have multiple files with better grouping.

# Pitch

We should think of grouping tests in logically similar way together. or consider creating test folder a package. See 
https://github.com/pytorch/vision/pull/4436 . Earlier we have faced problems doing the same.

From discussion with Phillip offline
Considering we don't Pin pytest. We can make use of  https://github.com/pytest-dev/pytest/pull/9134

With the python path param, we can probably make use of `common_utils.py` file as well. So that we can test files without hurting other scenarios which were mentioned in the comments. 

# Alternatives

This is quite tricky but very helpful as we don't want tests to hurt us later. 

# Reference
https://github.com/pytorch/vision/pull/4436#issuecomment-923131237

https://github.com/pytorch/vision/pull/4436#issuecomment-924005657

https://github.com/pytorch/vision/pull/4436#issuecomment-924250093

NicolasHug:
Thanks for opening this issue @oke-aditya 

There are a few more things to consider, as `loc`s don't give a full picture of a file's complexity.

[...]

PR #2710, opened 2020-09-27
Adds bounding boxes conversion

oke-aditya:
Closes #2687

- [x] Added code
- [x] Added documentation
- [x] Added tests

I have added as per the Issue, two utility function to convert boxes to pascal VOC format (x1 y1 x2 y2).

The tests convert boxes to other format and vice-versa. This ensures that both operations are identical and can be interchangeably used. Tests passed locally. 

This is ready for review. Do let me know!

cc @pmeier

PR #6027, opened 2022-05-16
Refactor tests for ops

### LINKED ISSUE #5976: Clean up Ops Box and Losses implementations

oke-aditya:
Part 2. Closes #5976 
Second part of refactor. This touches only the tests. 

Currently have created 2 separate files so that the diff is slightly easier to review. I can merge those files back to test_ops.py if needed. :smile: 

cc @pmeier


test/test_ious.py
```diff
@@ -0,0 +1,147 @@
...
+def _generate_float_input():
+    return [
+        [285.3538, 185.5758, 1193.5110, 851.4551],
+        [285.1472, 188.7374, 1192.4984, 851.0669],
+        [279.2440, 197.9812, 1189.4746, 849.2019],
+    ]
+
+
+class TestBoxIou(IouTestBase):
+    def _generate_int_expected():
```
oke-aditya, line 44: not pretty happy with this. But there are 3 ways to work around as far as I know
1. Either create the function outside.
2. Global variable
3. Fixture.

NicolasHug:
Thanks @oke-aditya ,

Before moving large section of tests into separate files, I think it might be worth reaching consensus in https://github.com/pytorch/vision/issues/6020 first. I'm a bit wary about such changes as they're really not as free as they might seem.
Do you think it'd be possible to make the relevant changes in this PR while still preserving the current file structure?

test/test_ops.py
```diff
@@ -1073,7 +1072,6 @@ def test_bbox_xywh_cxcywh(self):
             [[0, 0, 100, 100], [0, 0, 0, 0], [10, 15, 20, 20], [23, 35, 70, 60]], dtype=torch.float
         )
 
-        # This is wrong
```
oke-aditya, line 1076: Really? It isn't. I verified it. I don't know how this comment is here. (Or is it probably me who left it 2 years ago lol)

test/test_ops.py
```diff
@@ -1113,277 +1111,48 @@ def test_bbox_convert_jit(self):
...
-            actual_box = torch.tensor(test_input, dtype=dtype)
-            expected_box = torch.tensor(expected)
-            assert_close(actual_box, expected_box, tolerance)
-
-    def _run_jit_test(self, test_input: List) -> None:
-        box_tensor = torch.tensor(test_input, dtype=torch.float)
-        expected = self._perform_box_operation(box_tensor, True)
-        scripted_area = self._perform_box_operation(box_tensor, True)
-        torch.testing.assert_close(scripted_area, expected, rtol=0.0, atol=1e-3)
+class TestBoxArea:
```
oke-aditya, line 1119: This is a trade-off. We don't  have a compact test for TestBoxArea, but we sacrifice that for IoU to make code more readable. 

I really hat to put my head around understanding how 1 class does testing for so many methods (and two are far different, box_area vs box_iou)

test/test_ops.py
```diff
@@ -1113,277 +1111,48 @@ def test_bbox_convert_jit(self):
...
-    def test_complete_iou(self, test_input: List, dtypes: List[torch.dtype], tolerance: float, expected: List) -> None:
-        self._run_test(test_input, dtypes, tolerance, expected)
+            dtype=dtype,
+        )
+        expected = torch.tensor([604723.0806, 600965.4666, 592761.0085], dtype=torch.float64)
+        area_check(box_tensor, expected, tolerance=0.05)
 
-    def test_ciou_jit(self) -> None:
-        self._run_jit_test([[0, 0, 100, 100], [0, 0, 50, 50], [200, 200, 300, 300]])
+    def test_float16_box(self):
```
oke-aditya, line 1141: We can probably make stuff a bit compact with parametrize with pytest.params. Hmm but I was like it's ok this isn't so bad?

oke-aditya:
Yes @NicolasHug I get your point. The idea is

> Currently have created 2 separate files so that the diff is slightly easier to review. I can merge those files back to test_ops.py if needed. smile

Let's finish reviewing and adjusting the code, i will eventually move back code to same file. I took it out so that I could analyse it independently and made it slightly easier to run tests. We won't create this file structure at end of PR.

test/test_ious.py
```diff
@@ -0,0 +1,147 @@
...
+        return [[1.0, 0.25, 0.0], [0.25, 1.0, 0.0], [0.0, 0.0, 1.0]]
+
+    def _generate_float_input():
```
pmeier, line 47: This seems to be duplicated outside of the class.

test/test_ious.py
pmeier: > 3. Fixture.

That is probably not a good option since we want to iterate over multiple cases, but fixtures in general only support one value. Of course we could also loop over the inputs in the test, but that is less expressive as there will be no information what input failed.

> 1. Either create the function outside.
> 2. Global variable

Given that these functions don't take any inputs, I think we can use a variable instead. Plus, can't we define that as class attribute like

```py
class TestFoo:
    FOOS = ["foo", "bar", "baz"]

    @pytest.mark.parametrize("foo", FOOS)
    def test_foo(self, foo):
        pass
```

?

test/test_ious.py
```diff
@@ -0,0 +1,147 @@
...
+    @staticmethod
+    def _run_jit_test(target_fn: Callable, test_input: List):
+        box_tensor = torch.tensor(test_input, dtype=torch.float)
+        expected = target_fn(box_tensor, box_tensor)
+        scripted_fn = torch.jit.script(target_fn)
+        scripted_out = scripted_fn(box_tensor, box_tensor)
+        torch.testing.assert_close(scripted_out, expected, rtol=0.0, atol=1e-3)
+
+
+def _generate_int_input():
```
pmeier, line 31: Can we merge the `_generate_*_input` and `_generate_*_expected` functions? I see no point in keeping the values separate.

test/test_losses.py
```diff
@@ -0,0 +1,229 @@
...
+    torch.testing.assert_close(computed_loss, expected_loss, rtol=tol, atol=tol)
+
+
+def assert_empty_loss(iou_fn, dtype, device):
+    box1 = torch.randn([0, 4], dtype=dtype, device=device).requires_grad_()
+    box2 = torch.randn([0, 4], dtype=dtype, device=device).requires_grad_()
+    loss = iou_fn(box1, box2, reduction="mean")
+    loss.backward()
+    tol = 1e-3 if dtype is torch.half else 1e-5
+    torch.testing.assert_close(loss, torch.tensor(0.0, device=device), rtol=tol, atol=tol)
```
pmeier, line 33: Wait, we want the loss to be `0` if no boxes, i.e. boxes with a degenerate shape, were passed and we use `reduction="mean"`? But if we set `reduction="none"` the loss should be empty?

test/test_losses.py
```diff
@@ -0,0 +1,229 @@
...
+    box4 = torch.tensor([1, 1, 2, 2], dtype=dtype, device=device)
+
+    box1s = torch.stack([box2, box2], dim=0)
+    box2s = torch.stack([box3, box4], dim=0)
+
+    return box1, box2, box3, box4, box1s, box2s
+
+
+def assert_iou_loss(iou_fn, box1, box2, expected_loss, dtype, device, reduction="none"):
+    tol = 1e-3 if dtype is torch.half else 1e-5
```
pmeier, line 21: Is this still needed? [`torch.testing.assert_close`](https://pytorch.org/docs/master/testing.html#torch.testing.assert_close) uses the following default tolerances

| `dtype`   | `rtol`   | `atol`   |
|-----------|----------|----------|
| `float16` | `1e-3`   | `1e-5`   |
| `float32` | `1.3e-6` | `1e-5`   |
| `float64` | `1e-7`   | `1e-7`   |

They are a little stricter than what we have right now. Could you try if the tests also pass without specifying the tolerances?

This comment applies to all occurrences of setting the tolerances.

test/test_ops.py
pmeier: Yup, see #2710 :sweat_smile: If you verified that this ok, we can remove.

test/test_ops.py
```diff
@@ -1051,7 +1050,7 @@ def test_bbox_xyxy_xywh(self):
         assert_equal(box_xyxy, box_tensor)
 
     def test_bbox_xyxy_cxcywh(self):
-        # Simple test convert boxes to xywh and back. Make sure they are same.
+        # Simple test convert boxes to cxcywh and back. Make sure they are same.
```
pmeier, line 1053: Good catch!

test/test_ops.py
```diff
@@ -1113,277 +1111,48 @@ def test_bbox_convert_jit(self):
...
-
-    def _run_jit_test(self, test_input: List) -> None:
-        box_tensor = torch.tensor(test_input, dtype=torch.float)
-        expected = self._perform_box_operation(box_tensor, True)
-        scripted_area = self._perform_box_operation(box_tensor, True)
-        torch.testing.assert_close(scripted_area, expected, rtol=0.0, atol=1e-3)
+class TestBoxArea:
+    @pytest.mark.parametrize("dtype", [torch.int8, torch.int16, torch.int32, torch.int64])
+    def test_int_boxes(self, dtype):
+        # Check for int boxes
```
pmeier, line 1122: I don't think that comment is necessary as the the test name is expressive enough.

```suggestion
```

test/test_ops.py
```diff
@@ -1113,277 +1111,48 @@ def test_bbox_convert_jit(self):
...
-
-    def _generate_int_input() -> List[List[int]]:
-        return [[0, 0, 100, 100], [0, 0, 0, 0]]
-
-    def _generate_int_expected() -> List[int]:
-        return [10000, 0]
-
-    def _generate_float_input(index: int) -> List[List[float]]:
-        return [
+    # Check for float32 and float64 boxes
```
pmeier, line 1127: Same here.

```suggestion
```

test/test_ops.py
pmeier: Do you know why this check is special cases and not grouped with the other float dtypes?

oke-aditya:
I will get back to this on Tuesday. I'm shifting homes this week 🏡.

oke-aditya:
I think this weekend might give me some time. :(

datumbox:
@oke-aditya No worries. We thankfully managed to merge the PR that cleaned up the code prior the release. This is less urgent.

yassineAlouini:
@oke-aditya Do you need help finishing this? I have some dedicated time this afternoon (UTC +2 time) so I can review the code and/or help finishing it. Let me know. :ok_hand:

oke-aditya:
Hey I'm back on this. Was about to give time on this today. I have finally settled in new (or perhaps old?) home town :smile:

yassineAlouini:
> Hey I'm back on this. Was about to give time on this today. I have finally settled in new (or perhaps old?) home town smile
> 
> Feel free to help in reviewing as Philip is on PTO.

I will. Good luck with what remains. :+1:

test/test_ops.py
oke-aditya: Not sure probably to check overflow from float16 to float32?

test/test_ious.py
```diff
@@ -0,0 +1,147 @@
+from typing import List, Callable
+
+import pytest
+import torch
+import torch.fx
+from torch import Tensor
+from torchvision import ops
+
+
+class IouTestBase:
```
yassineAlouini, line 10: Minor detail: `class IoUTestBase` would better match the `IoU` way of writing it. Not sure if this is the recommended way but I would write it that way if I had to choose.

test/test_ious.py
```diff
@@ -0,0 +1,147 @@
...
+    @staticmethod
+    def _run_test(target_fn: Callable, test_input: List, dtypes: List[torch.dtype], tolerance: float, expected: List):
+        def assert_close(box: Tensor, expected: Tensor, tolerance):
+            out = target_fn(box, box)
+            torch.testing.assert_close(out, expected, rtol=0.0, check_dtype=False, atol=tolerance)
```
yassineAlouini, line 15: Why do you ignore checking the `dtype`? :thinking:

test/test_ious.py
yassineAlouini: Also, it would be better to have the same convention as the other test classes (that starts with `Test`), i.e. something like:

`TestIoUBase`

test/test_losses.py
```diff
@@ -0,0 +1,226 @@
...
+        assert_iou_loss(ops.generalized_box_iou_loss, box1s, box2s, 2.5, dtype=dtype, device=device, reduction="sum")
+        assert_iou_loss(ops.generalized_box_iou_loss, box1s, box2s, 1.25, dtype=dtype, device=device, reduction="mean")
+
+    @pytest.mark.parametrize("device", cpu_and_gpu())
+    @pytest.mark.parametrize("dtype", [torch.float32, torch.half])
+    def test_empty_inputs(self, dtype, device):
+        assert_empty_loss(ops.generalized_box_iou_loss, dtype, device)
+
+
+class TestCIOULoss:
```
yassineAlouini, line 70: Minor detail: here the class name follows a different convetion `IOU` vs `Iou`. Better have everything following the same, so for example `IoU`.

test/test_ops.py
yassineAlouini, line 1053: Nice catch. :+1:

test/test_ious.py
oke-aditya: Yeah Let me Rename all and keep things consistent.

oke-aditya: Our current convention is to name the test function as Test<op_name>  so I have refactored accordingly for consistency. Nice catch @yassineAlouini

test/test_ops.py
```diff
@@ -1113,277 +1111,45 @@ def test_bbox_convert_jit(self):
         torch.testing.assert_close(scripted_cxcywh, box_cxcywh, rtol=0.0, atol=TOLERANCE)
 
 
-class BoxTestBase(ABC):
-    @abstractmethod
-    def _target_fn(self) -> Tuple[bool, Callable]:
-        pass
+def area_check(box, expected, tolerance=1e-4):
+    out = ops.box_area(box)
+    torch.testing.assert_close(out, expected, rtol=0.0, check_dtype=False, atol=tolerance)
```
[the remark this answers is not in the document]
oke-aditya: Ok. So It is necessary to set `check_dtype = False` here. Since the boxes would be casted to higher dtype.
You can try running this by simply putting `check_dtype=True` It will throw error. 

As the calculated value is 10000 for box_area which is int32. (torch will automatically use int32 or you can set that)
But there is a case where we use int64 boxes.  This will lead to dtype mismatch.

Same applies for float too.

Safer assumption is to ignore dtypes when we are parametrizing the tests.

Hence keeping the ignore_dtype to True.

test/test_ious.py
oke-aditya: See my reasoning below. I think the same will apply for IoU as well.

test/test_ious.py
oke-aditya: Making global variable for these. I feel that's easier

test/test_ious.py
oke-aditya: Adapted to these per class as the expected calculation differs for each type of IoU

oke-aditya:
Hi @NicolasHug , @datumbox , @YosuaMichael could someone please review the PR as Philip is having a well deserved PTO :)

test/test_ious.py
```diff
@@ -0,0 +1,116 @@
...
+import torch
+import torch.fx
+from torch import Tensor
+from torchvision import ops
+
+
+class TestIouBase:
+    @staticmethod
+    def _run_test(target_fn: Callable, test_input: List, dtypes: List[torch.dtype], tolerance: float, expected: List):
+        def assert_close(box: Tensor, expected: Tensor, tolerance):
```
datumbox, line 13: Nit: This method is used once. Worth removing it to simplify the code.

test/test_ious.py
```diff
@@ -0,0 +1,116 @@
...
+            pytest.param(IOU_INT_BOXES, [torch.int16, torch.int32, torch.int64], 1e-4, generate_int_expected),
+            pytest.param(IOU_FLOAT_BOXES, [torch.float16], 0.002, float_expected),
+            pytest.param(IOU_FLOAT_BOXES, [torch.float32, torch.float64], 1e-3, float_expected),
+        ],
+    )
+    def test_iou(self, test_input, dtypes, tolerance, expected):
+        self._run_test(ops.box_iou, test_input, dtypes, tolerance, expected)
+
+    def test_iou_jit(self):
+        self._run_jit_test(ops.box_iou, [[0, 0, 100, 100], [0, 0, 50, 50], [200, 200, 300, 300]])
```
datumbox, line 55: Looks like this can be replaces with:
```suggestion
        self._run_jit_test(ops.box_iou, IOU_INT_BOXES)
```

Similarly on the other methods.

test/test_ops.py
```diff
@@ -1113,285 +1111,53 @@ def test_bbox_convert_jit(self):
...
-        box_tensor = torch.tensor(test_input, dtype=torch.float)
-        expected = self._perform_box_operation(box_tensor, True)
-        scripted_area = self._perform_box_operation(box_tensor, True)
-        torch.testing.assert_close(scripted_area, expected, rtol=0.0, atol=1e-3)
-
-
-class TestBoxArea(BoxTestBase):
-    def _target_fn(self) -> Tuple[bool, Callable]:
-        return (False, ops.box_area)
+def area_check(box, expected, tolerance=1e-4):
```
datumbox, line 1114: Nit: This method could optionally be placed within `TestBoxArea` to keep everything together.

datumbox (commented):
Thanks a lot @oke-aditya! I had a quick look and looks an improvement. I got only a few nits.

Also splitting the tests in separate files is definitely worth it IMO, but I would expect that the box op tests would all be together, the losses all together etc. Definitely not a blocking issue for me and we can review the structure on a future PR.

test/test_ops.py
```diff
@@ -1021,7 +1021,7 @@ def test_convert_boxes_to_roi_format(self, box_sequence):
             assert_equal(ref_tensor, ops._utils.convert_boxes_to_roi_format(box_sequence))
 
 
-class TestBox:
+class TestBoxConvert:
```
NicolasHug, line 1024: We have `TestBoxConversion` just above which is very similar. Should we merge them? Or find less conflicting names?

test/test_ops.py
```diff
@@ -1113,285 +1112,372 @@ def test_bbox_convert_jit(self):
         torch.testing.assert_close(scripted_cxcywh, box_cxcywh, rtol=0.0, atol=TOLERANCE)
 
 
-class BoxTestBase(ABC):
-    @abstractmethod
-    def _target_fn(self) -> Tuple[bool, Callable]:
-        pass
+class TestBoxArea:
+    def area_check(self, box, expected, tolerance=1e-4):
```
NicolasHug, line 1116:
```suggestion
    def area_check(self, box, expected, atol=1e-4):
```

test/test_ops.py
```diff
@@ -1113,285 +1112,372 @@ def test_bbox_convert_jit(self):
...
+    def test_float_boxes(self, dtype):
+        box_tensor = torch.tensor(
+            [
+                [285.3538, 185.5758, 1193.5110, 851.4551],
+                [285.1472, 188.7374, 1192.4984, 851.0669],
+                [279.2440, 197.9812, 1189.4746, 849.2019],
+            ],
+            dtype=dtype,
+        )
+        expected = torch.tensor([604723.0806, 600965.4666, 592761.0085], dtype=torch.float64)
```
NicolasHug, line 1136: Should this be `dtype=dtype`? The `box_area()` function will return the same dtype as the input I believe. Same  in other places

```suggestion
        expected = torch.tensor([604723.0806, 600965.4666, 592761.0085], dtype=dtype)
```

test/test_ops.py
NicolasHug: Could you please check if the test above works with float16 and potentially remove this one?

test/test_ops.py
```diff
@@ -1113,285 +1112,372 @@ def test_bbox_convert_jit(self):
...
+        )
+        expected = torch.tensor([605113.875, 600495.1875, 592247.25])
+        self.area_check(box_tensor, expected)
+
+    def test_box_area_jit(self):
+        box_tensor = torch.tensor([[0, 0, 100, 100], [0, 0, 0, 0]], dtype=torch.float)
+        expected = ops.box_area(box_tensor)
+        scripted_fn = torch.jit.script(ops.box_area)
+        scripted_area = scripted_fn(box_tensor)
+        torch.testing.assert_close(scripted_area, expected, rtol=0.0, atol=1e-3)
```
NicolasHug, line 1152: I am surprised that we need such a high tolerance here. Can we lower it?

test/test_ops.py
```diff
@@ -1113,285 +1112,372 @@ def test_bbox_convert_jit(self):
...
+class TestIouBase:
+    @staticmethod
+    def _run_test(target_fn: Callable, test_input: List, dtypes: List[torch.dtype], tolerance: float, expected: List):
         for dtype in dtypes:
             actual_box = torch.tensor(test_input, dtype=dtype)
             expected_box = torch.tensor(expected)
-            assert_close(actual_box, expected_box, tolerance)
+            out = target_fn(actual_box, actual_box)
+            torch.testing.assert_close(out, expected_box, rtol=0.0, check_dtype=False, atol=tolerance)
+            # assert_close(actual_box, expected_box, tolerance)
```
NicolasHug, line 1163: should this be removed?

test/test_ops.py
```diff
@@ -1113,285 +1112,372 @@ def test_bbox_convert_jit(self):
...
-                [285.1472, 188.7374, 1192.4984, 851.0669],
-                [279.2440, 197.9812, 1189.4746, 849.2019],
-            ],
-            [[285.25, 185.625, 1194.0, 851.5], [285.25, 188.75, 1192.0, 851.0], [279.25, 198.0, 1189.0, 849.0]],
-        ][index]
 
-    def _generate_float_expected(index: int) -> List[float]:
-        return [[604723.0806, 600965.4666, 592761.0085], [605113.875, 600495.1875, 592247.25]][index]
+class TestBoxIou(TestIouBase):
+    generate_int_expected = [[1.0, 0.25, 0.0], [0.25, 1.0, 0.0], [0.0, 0.0, 1.0]]
```
NicolasHug, line 1183: why is this called `generate_`?

test/test_ops.py
```diff
@@ -1113,285 +1112,372 @@ def test_bbox_convert_jit(self):
...
+        torch.testing.assert_close(scripted_area, expected, rtol=0.0, atol=1e-3)
 
-    def _run_test(self, test_input: List, dtypes: List[torch.dtype], tolerance: float, expected: List) -> None:
-        def assert_close(box: Tensor, expected: Tensor, tolerance):
-            out = self._perform_box_operation(box)
-            torch.testing.assert_close(out, expected, rtol=0.0, check_dtype=False, atol=tolerance)
 
+class TestIouBase:
+    @staticmethod
+    def _run_test(target_fn: Callable, test_input: List, dtypes: List[torch.dtype], tolerance: float, expected: List):
```
NicolasHug, line 1157:
```suggestion
    def _run_test(target_fn: Callable, test_input: List, dtypes: List[torch.dtype], atol: float, expected: List):
```

test/test_ops.py
```diff
@@ -1113,285 +1112,372 @@ def test_bbox_convert_jit(self):
...
+    box1s = torch.stack([box2, box2], dim=0)
+    box2s = torch.stack([box3, box4], dim=0)
 
-    assert_distance_iou_loss(box1, box1, 0.0)
+    return box1, box2, box3, box4, box1s, box2s
 
-    assert_distance_iou_loss(box1, box2, 0.8125)
 
-    assert_distance_iou_loss(box1, box3, 1.1923)
+def assert_iou_loss(iou_fn, box1, box2, expected_loss, dtype, device, reduction="none"):
```
NicolasHug, line 1270: dtype is never used

NicolasHug (commented):
Thanks @oke-aditya , made a first pass

test/test_ops.py
oke-aditya: We are using tolerance in all other similar places. So maybe good to keep for consistency>

test/test_ops.py
oke-aditya: Yes

test/test_ops.py
```diff
@@ -1113,285 +1112,372 @@ def test_bbox_convert_jit(self):
...
+        scripted_area = scripted_fn(box_tensor)
+        torch.testing.assert_close(scripted_area, expected, rtol=0.0, atol=1e-3)
 
-    def _run_test(self, test_input: List, dtypes: List[torch.dtype], tolerance: float, expected: List) -> None:
-        def assert_close(box: Tensor, expected: Tensor, tolerance):
-            out = self._perform_box_operation(box)
-            torch.testing.assert_close(out, expected, rtol=0.0, check_dtype=False, atol=tolerance)
 
+class TestIouBase:
+    @staticmethod
```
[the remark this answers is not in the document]
oke-aditya: Hmm, nothing special. But its not wrong either. As such it doesn't need `self` argument.

test/test_ops.py
oke-aditya: Agreed, After few more thoughts, I got the point. tolerance is ambigous and better to specify its atol.

test/test_ops.py
oke-aditya: Removed it :+1:

test/test_ops.py
oke-aditya: right

oke-aditya:
Need to check the area one. Probably there is something

oke-aditya:
Any other concerns apart from the area tests? I will have a look on Friday.

NicolasHug:
I think that's all on my side, thanks @oke-aditya

datumbox:
@oke-aditya I see you asked for another review. My understanding from your above comment is that you will still try to fix the situation on the area test in this PR, right? If that's the case, I recommend to mark it as draft and reopen once ready for reviews. If you want a review now, let me know and I'll make some time.

Thanks for the contribution as always! :)

test/test_ops.py
oke-aditya: Well the float16 test is a bit notorious. It needs bit of higher tolerance, to pass. And especially with smaller boxes, Using such big boxes will of course lead to float overflow. 

I cannot find a workaround to merge these two tests without adjusting the tolerance :( So as of now let it stay

test/test_ops.py
```diff
@@ -1102,296 +1101,382 @@ def test_bbox_convert_jit(self):
...
-        target_fn = self._target_fn()[1]
-        box_operation = torch.jit.script(target_fn) if run_as_script else target_fn
-        return box_operation(box, box) if is_binary_fn else box_operation(box)
+    def test_float16_box(self):
+        box_tensor = torch.tensor(
+            [[285.25, 185.625, 1194.0, 851.5], [285.25, 188.75, 1192.0, 851.0], [279.25, 198.0, 1189.0, 849.0]],
+            dtype=torch.float16,
+        )
+        expected = torch.tensor([605113.875, 600495.1875, 592247.25], dtype=torch.float32)
+        self.area_check(box_tensor, expected)
```
datumbox, line 1145: As @NicolasHug highlighted above, we are trying to check float16 values that are caped at 65k against expected values at around 600k. This shouldn't work. We should revise the testing values to avoid overflows.

test/test_ops.py
datumbox: I'm probably missing something but I don't see the float16 used in this test. At a glance my understanding is we are at float32. What am I missing?

oke-aditya: Aah I commented in wrong place :(

oke-aditya: This doesn't need such weak limits. We can use the default ones.

test/test_ops.py
oke-aditya: Yep. I understood now. Let me  revise the tests,

test/test_ops.py
```diff
@@ -1102,296 +1101,381 @@ def test_bbox_convert_jit(self):
         )
 
         scripted_fn = torch.jit.script(ops.box_convert)
-        TOLERANCE = 1e-3
 
         box_xywh = ops.box_convert(box_tensor, in_fmt="xyxy", out_fmt="xywh")
         scripted_xywh = scripted_fn(box_tensor, "xyxy", "xywh")
-        torch.testing.assert_close(scripted_xywh, box_xywh, rtol=0.0, atol=TOLERANCE)
+        torch.testing.assert_close(scripted_xywh, box_xywh)
```
oke-aditya, line 1107: We can tighten the tolerance here too.

test/test_ops.py
```diff
@@ -1102,296 +1101,381 @@ def test_bbox_convert_jit(self):
...
+        expected = torch.tensor([604723.0806, 600965.4666, 592761.0085], dtype=dtype)
+        self.area_check(box_tensor, expected)
+
+    def test_float16_box(self):
+        box_tensor = torch.tensor(
+            [[28.25, 18.625, 39.0, 48.5], [28.25, 48.75, 192.0, 51.0], [29.25, 18.0, 89.0, 49.0]],
+            dtype=torch.float16,
+        )
+        expected = torch.tensor([321.1562, 368.4375, 1852.2500], dtype=torch.float16)
+        self.area_check(box_tensor, expected, atol=0.3)
```
oke-aditya, line 1144: Well sadly for small boxes and keeping both as float16. We need to loosen atol. I can't find some other workaround.

test/test_ops.py
```diff
@@ -1102,296 +1101,381 @@ def test_bbox_convert_jit(self):
...
+        self.area_check(box_tensor, expected)
+
+    @pytest.mark.parametrize("dtype", [torch.float32, torch.float64])
+    def test_float_boxes(self, dtype):
+        box_tensor = torch.tensor(
+            [
+                [285.3538, 185.5758, 1193.5110, 851.4551],
+                [285.1472, 188.7374, 1192.4984, 851.0669],
+                [279.2440, 197.9812, 1189.4746, 849.2019],
+            ],
```
NicolasHug, line 1132: Should this be `IOU_FLOAT_BOXES`?

test/test_ops.py
NicolasHug: It's hard to decipher the diff. What test does this correspond to in `main`? It looks like input and generated values were changed. Is this on purpose?

oke-aditya: It is on purpose. So that box_area after multiplication too stays in float16 input format.

yassineAlouini:
@oke-aditya Any more help/review needed here? I have some time this Friday for OSS, so let me know. Thanks. :ok_hand:

oke-aditya:
Seems we are very close to finish this. Sadly I'm busy at office and sunsetting my old laptop to move onto new one. (it's gonna take some time etc) 

You can help me wrap up this PR by simply sending a PR to my fork of torchvision. On this branch. I can merge that. 

Then Just ping Nicolas / Vasilis if anything else is needed.

yassineAlouini:
> Seems we are very close to finish this. Sadly I'm busy at office and sunsetting my old laptop to move onto new one. (it's gonna take some time etc)
> 
> You can help me wrap up this PR by simply sending a PR to my fork of torchvision. On this branch. I can merge that.
> 
> Then Just ping Nicolas / Vasilis if anything else is needed.

Alright @oke-aditya, will do this on Firday and let you know. Good luck resetting your computer (I hope you have some of your dot files saved somewhere. :smile: )

oke-aditya:
Yeah the challenge is that the new laptop has GPU. And I never had GPU locally before. So setting up that with nvidia drivers, building torchvision from source etc is gonna be a cool challenge. 

(Yes till now I did all contributions or learn deep learning without GPU :) colab was my messiah)

datumbox:
@oke-aditya compiling from source with GPU should be seamless. If it's not, we will take your PR, thank you very much. 😄

oke-aditya:
Yeah hopefully :) Doing it first time. All should go well 😊

oke-aditya:
Major blockers are cleared!

I could successfully build pytorch from source. Which is sadly needed as torch with cuda 11.7 binaries aren't available in nightlies yet :(

I should wrap up the PR by Wednesday 🤠

oke-aditya:
@NicolasHug how do we go about this ? Is it good to go?

datumbox:
@oke-aditya seems there is a related issue with the [linter](https://app.circleci.com/pipelines/github/pytorch/vision/18980/workflows/c53a1c50-3b05-4d6d-baed-5f1dfbdc85d6/jobs/1535489):
```
diff --git a/test/test_ops.py b/test/test_ops.py
index 53637baf3..db2ba4776 100644
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -3,7 +3,7 @@ import os
 from abc import ABC, abstractmethod
 from functools import lru_cache
 from itertools import product
-from typing import Tuple, Callable, List
+from typing import Callable, List, Tuple
 
 import numpy as np
 import pytest
```

oke-aditya:
I can fix that 😃. Anything else apart from that?

test/test_ops.py
datumbox: Important: The `atol=0.3` is extremely large and might lead to not being able to detect problems. Is that the minimum you can do? If yes then it's worth considering reviewing the values of the boxes so you can use smaller values like `0.001` which I believe was used previously.

test/test_ops.py
```diff
@@ -1102,296 +1101,374 @@ def test_bbox_convert_jit(self):
...
+    def test_distance_iou_loss(self, dtype, device):
+        box1, box2, box3, box4, box1s, box2s = get_boxes(dtype, device)
 
-    assert_distance_iou_loss(box1, box4, 1.2500)
+        assert_iou_loss(ops.distance_box_iou_loss, box1, box1, 0.0, device=device)
+        assert_iou_loss(ops.distance_box_iou_loss, box1, box2, 0.8125, device=device)
+        assert_iou_loss(ops.distance_box_iou_loss, box1, box3, 1.1923, device=device)
+        assert_iou_loss(ops.distance_box_iou_loss, box1, box4, 1.2500, device=device)
+        assert_iou_loss(ops.distance_box_iou_loss, box1s, box2s, 1.2250, device=device, reduction="mean")
+        assert_iou_loss(ops.distance_box_iou_loss, box1s, box2s, 2.4500, device=device, reduction="sum")
```
datumbox, line 1341: Future improvement: I know that these values were also used originally but I think it's problematic that both cIoU and dIoU losses are being tested against examples that yield identical outputs. Aka if I send a PR that replaces cIoU with dIoU, how do we capture the issue?

test/test_ops.py
```diff
@@ -1102,296 +1101,374 @@ def test_bbox_convert_jit(self):
...
-    )
-    def test_complete_iou(self, test_input: List, dtypes: List[torch.dtype], tolerance: float, expected: List) -> None:
-        self._run_test(test_input, dtypes, tolerance, expected)
+    @pytest.mark.parametrize("alpha", [-1.0, 0.0, 0.58, 1.0])
+    @pytest.mark.parametrize("gamma", [0, 2])
+    @pytest.mark.parametrize("reduction", ["none", "mean", "sum"])
+    @pytest.mark.parametrize("device", cpu_and_gpu())
+    @pytest.mark.parametrize("dtype", [torch.float32, torch.half])
+    @pytest.mark.parametrize("seed", [4, 5])
+    def test_jit(self, alpha, gamma, reduction, device, dtype, seed):
```
datumbox, line 1447: Future improvement: It's kinda problematic that the only loss that checks for JIT scriptability is Focal. Why this discrepancy in the testing strategy? I would imagine that a refactoring of the test ops would resolve this.

test/test_ops.py
```diff
@@ -1102,296 +1101,374 @@ def test_bbox_convert_jit(self):
...
+    def test_empty_distance_iou_inputs(self, dtype, device):
+        assert_empty_loss(ops.distance_box_iou_loss, dtype, device)
 
 
-@pytest.mark.parametrize("device", cpu_and_gpu())
-@pytest.mark.parametrize("dtype", [torch.float32, torch.half])
-def test_empty_distance_iou_inputs(dtype, device) -> None:
-    box1 = torch.randn([0, 4], dtype=dtype, device=device).requires_grad_()
-    box2 = torch.randn([0, 4], dtype=dtype, device=device).requires_grad_()
+class TestFocalLoss:
```
datumbox, line 1349: `<rant>`I took the previous code vs new and diffed them. This indicated that you made only minor modifications to the class which Github fails to highlight. This is where splitting the PR into multiple comes handy and can speed up reviews.`</rant>`

datumbox (commented):
@oke-aditya I've left 4 comments below but you need to take action only for the tolerance value (+ the linter)

Overall I'm tempted to merge this because:
1. It's an improvement comparing to the previous tests which had various issues.
2. It removes 200 lines of code from the tests

Your PR remained open for so long because it appears to be very lengthy and touched multiple lines. In practice you only make minor changes on the main part of the classes but Github fails to see them. There are other improvements that I felt could have been made in this PR but this is already too long and can be handled on a follow up. I think to speed up reviews, it's worth splitting work in chunks and actively trying to structure the code in a way that leads to a more compact Github diff (but I know this is not always possible).

After you make the changes on the linter and the tolerance value, I intend to merge unless someone else has strong opinion against it. cc @NicolasHug @pmeier

test/test_ops.py
oke-aditya: Seems I should learn to use ghstack https://github.com/ezyang/ghstack This would make it simple.

oke-aditya:
So I have fixed the required issues. I see that even more improvements can be done. And especially we need to address #6317 
and add tests accordingly. I will follow up with a new PR.

datumbox (approved):
LGTM, thanks!

PR #6029, opened 2022-05-16
temporaily disable nvfuser

davidberard98:
Stack from [ghstack](https://github.com/ezyang/ghstack):
* __->__ #6029

Waiting on pytorch/pytorch#77471 to fix the breakage, this will stop
failures for now.

### DIFF
--- a/.circleci/unittest/linux/scripts/run_test.sh
+++ b/.circleci/unittest/linux/scripts/run_test.sh
@@ -6,4 +6,4 @@ eval "$(./conda/bin/conda shell.bash hook)"
 conda activate ./env
 
 python -m torch.utils.collect_env
-pytest --junitxml=test-results/junit.xml -v --durations 20
+PYTORCH_JIT_ENABLE_NVFUSER=0 pytest --junitxml=test-results/junit.xml -v --durations 20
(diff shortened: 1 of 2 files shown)


.circleci/unittest/linux/scripts/run_test.sh
```diff
@@ -6,4 +6,4 @@ eval "$(./conda/bin/conda shell.bash hook)"
 conda activate ./env
 
 python -m torch.utils.collect_env
-pytest --junitxml=test-results/junit.xml -v --durations 20
+PYTORCH_JIT_ENABLE_NVFUSER=0 pytest --junitxml=test-results/junit.xml -v --durations 20
```
datumbox, line 9: @davidberard98 If I understand correctly, PyTorch core recently turned on a new fuser and this command will turn it off for TorchVision's tests by default. Though this will make the tests Green, it will mask the problem and potentially stop us from discovering more issues on the future. Even though I understand your intention is to leave this only temporarily, the new fuser has repeatedly broke our tests and since the work is still in progress, I'm worried it will make it harder to detect more on-going problems.

I'm very reluctant to approve this because we are getting closer to the release. My preferred solution is to find and revert the offending PR and try to reland it after fixing it. I believe this is also the recommended practice on Core. Note that TorchVision contains a series of tests that use the new Fuser in real-world scenarios, so it might be very useful running its tests for any follow up PR.

datumbox:
Closing the PR because @davidberard98 reverted https://github.com/pytorch/pytorch/pull/77213 which caused the breakage. A fix by @jjsjann123 is underway at https://github.com/pytorch/pytorch/pull/77471 which could potentially address the specific breakage.

Issue #6014, opened 2022-05-13
Improve documentation for Multi-weight support


datumbox:
### 📚 The doc issue

Previously we have restructured the documentation for models (#5833, #5987, #5897 etc), so that we can add necessary information about how the models were trained and integrate with the new Multi-weight support API.

This ticket is a tracker for all remaining tasks related to model documentation + Multi-weight support API:
- [x] Clean up the documentation pages #6003 #6002
- [x] Include information about the preprocessing transforms of models #6009
- [x] Add examples on how to use the new Multi-weight model builders #6013 
- [x] Move from the old `models.srt` any useful info to the new one (TORCH_HOME + TorchHub info, train/eval modes etc) #6026 #6025
- [x] Update galleries to use new idioms (for example fetching [classes](https://github.com/pytorch/vision/blob/11bd2eaa6d6976129836b329b01d1300babddcc9/gallery/plot_visualization_utils.py#L149) from meta-data and the text concerning [preprocessing](https://github.com/pytorch/vision/blob/11bd2eaa6d6976129836b329b01d1300babddcc9/gallery/plot_visualization_utils.py#L122-L124)) #6030
- [ ] Add description to pre-trained weights that clarify their scope #6036 #6039

PR #6030, opened 2022-05-16
Update galleries to use Multi-weight idioms

datumbox:
Resolves partially #6014

This PR:
- Updates the preprocessing description for RAFT
- Fixes the incorrect examples for detection models (passing tensors vs List[tensor])
- Updates the preprocessing descriptions and the meta-data usage of viz-utils gallery

### DIFF
--- a/gallery/plot_optical_flow.py
+++ b/gallery/plot_optical_flow.py
@@ -81,11 +81,11 @@ img2_batch = torch.stack([frames[101], frames[151]])
 plot(img1_batch)
 
 #########################
-# The RAFT model that we will use accepts RGB float images with pixel values in
-# [-1, 1]. The frames we got from :func:`~torchvision.io.read_video` are int
-# images with values in [0, 255], so we will have to pre-process them. We also
-# reduce the image sizes for the example to run faster. Image dimension must be
-# divisible by 8.
+# The RAFT model accepts RGB images. We first get the frames from
+# :func:`~torchvision.io.read_video` and resize them to ensure their
+# dimensions are divisible by 8. Then we use the transforms bundled into the
+# weights in order to preprocess the input and rescale its values to the
+# required ``[-1, 1]`` interval.
 
 from torchvision.models.optical_flow import Raft_Large_Weights
 
--- a/gallery/plot_visualization_utils.py
+++ b/gallery/plot_visualization_utils.py
@@ -43,8 +43,9 @@ from pathlib import Path
 
 dog1_int = read_image(str(Path('assets') / 'dog1.jpg'))
 dog2_int = read_image(str(Path('assets') / 'dog2.jpg'))
+dog_list = [dog1_int, dog2_int]
 
-grid = make_grid([dog1_int, dog2_int, dog1_int, dog2_int])
+grid = make_grid(dog_list)
 show(grid)
 
 ####################################
@@ -67,26 +68,21 @@ show(result)
 # Naturally, we can also plot bounding boxes produced by torchvision detection
 # models.  Here is demo with a Faster R-CNN model loaded from
 # :func:`~torchvision.models.detection.fasterrcnn_resnet50_fpn`
-# model. You can also try using a RetinaNet with
-# :func:`~torchvision.models.detection.retinanet_resnet50_fpn`, an SSDlite with
-# :func:`~torchvision.models.detection.ssdlite320_mobilenet_v3_large` or an SSD with
-# :func:`~torchvision.models.detection.ssd300_vgg16`. For more details
-# on the output of such models, you may refer to :ref:`instance_seg_output`.
+# model. For more details on the output of such models, you may
+# refer to :ref:`instance_seg_output`.
 
 from torchvision.models.detection import fasterrcnn_resnet50_fpn, FasterRCNN_ResNet50_FPN_Weights
 
 
-batch_int = torch.stack([dog1_int, dog2_int])
-
 weights = FasterRCNN_ResNet50_FPN_Weights.DEFAULT
 transforms = weights.transforms()
 
-batch = transforms(batch_int)
+images = [transforms(d) for d in dog_list]
 
 model = fasterrcnn_resnet50_fpn(weights=weights, progress=False)
 model = model.eval()
 
-outputs = model(batch)
+outputs = model(images)
 print(outputs)
 
 #####################################
@@ -96,7 +92,7 @@ print(outputs)
 score_threshold = .8
 dogs_with_boxes = [
     draw_bounding_boxes(dog_int, boxes=output['boxes'][output['scores'] > score_threshold], width=4)
-    for dog_int, output in zip(batch_int, outputs)
+    for dog_int, output in zip(dog_list, outputs)
 ]
 show(dogs_with_boxes)
 
@@ -131,8 +127,8 @@ transforms = weights.transforms(resize_size=None)
 model = fcn_resnet50(weights=weights, progress=False)
 model = model.eval()
 
-normalized_batch = transforms(batch)
-output = model(normalized_batch)['out']
+batch = torch.stack([transforms(d) for d in dog_list])
+output = model(batch)['out']
 print(output.shape, output.min().item(), output.max().item())
 
 #####################################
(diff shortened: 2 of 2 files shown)


gallery/plot_optical_flow.py
```diff
@@ -81,11 +81,11 @@ def plot(imgs, **imshow_kwargs):
...
-# The RAFT model that we will use accepts RGB float images with pixel values in
-# [-1, 1]. The frames we got from :func:`~torchvision.io.read_video` are int
-# images with values in [0, 255], so we will have to pre-process them. We also
-# reduce the image sizes for the example to run faster. Image dimension must be
-# divisible by 8.
+# The RAFT model accepts RGB images. We first get the frames from
+# :func:`~torchvision.io.read_video` and resize them to ensure their
+# dimensions are divisible by 8. Then we use the transforms bundled into the
+# weights in order to preprocess the input and rescale its values to the
+# required ``[-1, 1]`` interval.
```
datumbox, line 88: I rewrote this to reflect what preprocessing is currently visible to the user while using the presets.

gallery/plot_visualization_utils.py
```diff
@@ -43,8 +43,9 @@ def show(imgs):
 
 dog1_int = read_image(str(Path('assets') / 'dog1.jpg'))
 dog2_int = read_image(str(Path('assets') / 'dog2.jpg'))
+dog_list = [dog1_int, dog2_int]
 
-grid = make_grid([dog1_int, dog2_int, dog1_int, dog2_int])
+grid = make_grid(dog_list)
```
datumbox, line 48: Not sure why we were showing twice the picture of each dog previously.

gallery/plot_visualization_utils.py
```diff
@@ -65,28 +66,23 @@ def show(imgs):
 
 #####################################
 # Naturally, we can also plot bounding boxes produced by torchvision detection
-# models.  Here is demo with a Faster R-CNN model loaded from
+# models.  Here is a demo with a Faster R-CNN model loaded from
 # :func:`~torchvision.models.detection.fasterrcnn_resnet50_fpn`
-# model. You can also try using a RetinaNet with
-# :func:`~torchvision.models.detection.retinanet_resnet50_fpn`, an SSDlite with
-# :func:`~torchvision.models.detection.ssdlite320_mobilenet_v3_large` or an SSD with
-# :func:`~torchvision.models.detection.ssd300_vgg16`. For more details
```
datumbox, line 73: I'm not sure why we kept adding more and more models here but I ended up removing them all. I think people get it that they can use other models too, without listing them all.

gallery/plot_visualization_utils.py
```diff
@@ -65,28 +66,23 @@ def show(imgs):
...
-# on the output of such models, you may refer to :ref:`instance_seg_output`.
+# model. For more details on the output of such models, you may
+# refer to :ref:`instance_seg_output`.
 
 from torchvision.models.detection import fasterrcnn_resnet50_fpn, FasterRCNN_ResNet50_FPN_Weights
 
 
-batch_int = torch.stack([dog1_int, dog2_int])
```
datumbox, line 79: This was the wrong way to use detection models. They receive a list of tensors rather than a batch, so I made the change everywhere below. I've also made sure not to reference `batch_int` in other places.

gallery/plot_visualization_utils.py
```diff
@@ -131,8 +121,8 @@ def show(imgs):
 model = fcn_resnet50(weights=weights, progress=False)
 model = model.eval()
 
-normalized_batch = transforms(batch)
```
datumbox, line 134: Previously we were doubly preprocessing the images. I've changed it to do it only once using the original data (dog_list).

gallery/plot_visualization_utils.py
```diff
@@ -145,18 +135,13 @@ def show(imgs):
 # Let's plot the masks that have been detected for the dog class and for the
 # boat class:
 
-sem_classes = [
-    '__background__', 'aeroplane', 'bicycle', 'bird', 'boat', 'bottle', 'bus',
-    'car', 'cat', 'chair', 'cow', 'diningtable', 'dog', 'horse', 'motorbike',
-    'person', 'pottedplant', 'sheep', 'sofa', 'train', 'tvmonitor'
-]
-sem_class_to_idx = {cls: idx for (idx, cls) in enumerate(sem_classes)}
+sem_class_to_idx = {cls: idx for (idx, cls) in enumerate(weights.meta["categories"])}
```
datumbox, line 138: Switch to meta-data usage.

gallery/plot_visualization_utils.py
```diff
@@ -252,7 +237,7 @@ def show(imgs):
 # Instance segmentation models
 # ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 #
-# Instance segmentation models have a significantly different output from the
+# Instance segmentation models have a slightly different output from the
```
datumbox, line 240: I'm not sure why we said that MaskRCNN has a significantly different output than FasterRCNN here. It just includes the same as FasterRCNN + the mask info.

datumbox (commented):
Few comments to highlight important changes:

gallery/plot_visualization_utils.py
NicolasHug: It was for the backlinks. See https://pytorch.org/vision/main/models/generated/torchvision.models.detection.retinanet_resnet50_fpn.html#torchvision.models.detection.retinanet_resnet50_fpn at the bottom, theres a "Examples using retinanet_resnet50_fpn:" section.

Not mentioning the model builders in the example removes that. I don't have a strong opinion, I agree it's not really scalable if we have to remember to link all models from the examples. But it's also nice to have the backlinks... IDK. Maybe we can do that more manually.

gallery/plot_visualization_utils.py
NicolasHug: I'd be in favour of removing this section about `make_grid` entirely. We shouldn't encourage our users to use it anymore, there are much better alternatives

gallery/plot_visualization_utils.py
NicolasHug: The masks are quite different: one is a non-normalized score while for the other it's probabilities directly. Also one can detect an arbiratry number of classes while for the other one it's fixed, right?

(I'm fine with the change regardless)

gallery/plot_visualization_utils.py
```diff
@@ -304,30 +289,15 @@ def show(imgs):
...
-    'bottle', 'N/A', 'wine glass', 'cup', 'fork', 'knife', 'spoon', 'bowl',
-    'banana', 'apple', 'sandwich', 'orange', 'broccoli', 'carrot', 'hot dog', 'pizza',
-    'donut', 'cake', 'chair', 'couch', 'potted plant', 'bed', 'N/A', 'dining table',
-    'N/A', 'N/A', 'toilet', 'N/A', 'tv', 'laptop', 'mouse', 'remote', 'keyboard', 'cell phone',
-    'microwave', 'oven', 'toaster', 'sink', 'refrigerator', 'N/A', 'book',
-    'clock', 'vase', 'scissors', 'teddy bear', 'hair drier', 'toothbrush'
-]
-
-inst_class_to_idx = {cls: idx for (idx, cls) in enumerate(inst_classes)}
+inst_class_to_idx = {cls: idx for (idx, cls) in enumerate(weights.meta["categories"])}
```
NicolasHug, line 297: should we print it?

gallery/plot_visualization_utils.py
datumbox: Let's add manually, maybe with refs from the documentation pages of the methods as you recommended previously offline. The specific list was not complete and missing several new models.

gallery/plot_visualization_utils.py
datumbox: happy to revert

gallery/plot_visualization_utils.py
datumbox: Sounds good to me, let's do this on a separate PR to explore what the better recommended alternative is. cc @oke-aditya

gallery/plot_visualization_utils.py
datumbox: Good spot. Actually I don't know why this was here. We never printed it. I think I'll just remove it rather than printing a big mapping table on the docs.

PR #6038, opened 2022-05-17
add tests for F.pad_bounding_box

pmeier:
Related to #5514 as it is the last missing item.


vfdev-5:
@pmeier a correctness test could be good to have as well

pmeier:
> a correctness test on cpu/cuda could be good to have as well

Agreed, but I'm covered right now. Can we add that after #5879? Otherwise, would you be able to contribute it?

vfdev-5:
@pmeier OK let me merge this PR and add a correctness test

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -415,16 +415,14 @@ def pad_bounding_box(
...
-
-    bounding_box[..., 0::2] += left
-    bounding_box[..., 1::2] += top
-
-    return convert_bounding_box_format(
-        bounding_box, old_format=features.BoundingBoxFormat.XYXY, new_format=format, copy=False
-    )
+    bounding_box = bounding_box.clone()
+
+    bounding_box[..., 0] += left
```
pmeier, line 420: Could you add a comment here that this works without conversion since padding only affects xy coordinates

test/test_prototype_transforms_functional.py
```diff
@@ -1103,22 +1112,67 @@ def test_correctness_pad_segmentation_mask_on_fixed_input(device):
...
+        bbox[1::2] += pad_up
+
+        bbox = convert_bounding_box_format(
+            bbox, old_format=features.BoundingBoxFormat.XYXY, new_format=bbox_format, copy=False
+        )
+        if bbox.dtype != bbox_dtype:
+            # Temporary cast to original dtype
+            # e.g. float32 -> int
+            bbox = bbox.to(bbox_dtype)
+        return bbox
```
pmeier, line 1147: 1. There is no need for this `if`: `.to` is a no-op if no conversion needs to be performed.
2. You don't need to use the dtype explicitly. You can pass a tensor to `.to` in which case it will convert device and dtype to the input.

```suggestion
        return convert_bounding_box_format(
            bbox, old_format=features.BoundingBoxFormat.XYXY, new_format=bbox_.format, copy=False
        ).to(bbox_)
```

test/test_prototype_transforms_functional.py
```diff
@@ -1103,22 +1112,67 @@ def test_correctness_pad_segmentation_mask_on_fixed_input(device):
...
+
+@pytest.mark.parametrize("device", cpu_and_gpu())
+@pytest.mark.parametrize("padding", [[1], [1, 1], [1, 1, 2, 2]])
+def test_correctness_pad_bounding_box(device, padding):
+    def _compute_expected_bbox(bbox, padding_):
+        pad_left, pad_up, _, _ = _parse_padding(padding_)
+
+        bbox_format = bbox.format
+        bbox_dtype = bbox.dtype
+        bbox = convert_bounding_box_format(bbox, old_format=bbox_format, new_format=features.BoundingBoxFormat.XYXY)
```
pmeier, line 1135:
```suggestion
    def _compute_expected_bbox(bbox_, padding_):
        pad_left, pad_up, _, _ = _parse_padding(padding_)

        bbox = convert_bounding_box_format(bbox_, old_format=bbox_.format, new_format=features.BoundingBoxFormat.XYXY)
```

pmeier (commented):
A few minor comments. Otherwise LGTM. Thanks Victor!

torchvision/prototype/transforms/functional/_geometry.py
vfdev-5: By the way a similar trick could be done for croping

vfdev-5 (approved):

test/test_prototype_transforms_functional.py
vfdev-5: @pmeier actually it does not work with `.to` between tensors and BoundingBox:
```python
>>> t = torch.tensor([[1.0, 2.0, 3.0, 4.0]])
>>> bbox = features.BoundingBox([[0, 1, 2, 3]], format=features.BoundingBoxFormat.XYXY, image_size=(32, 32))
>>> t.to(bbox)
  File "<stdin>", line 1, in <module>
  File "/vision/torchvision/prototype/features/_feature.py", line 83, in __torch_function__
    return cls.new_like(args[0], output, dtype=output.dtype, device=output.device)
  File "/vision/torchvision/prototype/features/_bounding_box.py", line 54, in new_like
    format=format if format is not None else other.format,
AttributeError: 'Tensor' object has no attribute 'format'
```

pmeier: I'll added #6094 to track this.

PR #6039, opened 2022-05-17
Document all remaining pre-trained weights

### LINKED ISSUE #6014: Improve documentation for Multi-weight support

datumbox:
Fixes #6014

Similar to #6036, we document the weights for all remaining weights.

### DIFF
--- a/test/test_extended_models.py
+++ b/test/test_extended_models.py
@@ -95,10 +95,10 @@ def test_schema_meta_validation(model_fn):
     # mandatory fields for each computer vision task
     classification_fields = {"categories", ("metrics", "acc@1"), ("metrics", "acc@5")}
     defaults = {
-        "all": {"metrics", "min_size", "num_params", "recipe"},
-        "models": classification_fields | {"_docs"},
+        "all": {"metrics", "min_size", "num_params", "recipe", "_docs"},
+        "models": classification_fields,
         "detection": {"categories", ("metrics", "box_map")},
-        "quantization": classification_fields | {"backend", "unquantized"} | {"_docs"},
+        "quantization": classification_fields | {"backend", "unquantized"},
         "segmentation": {"categories", ("metrics", "miou"), ("metrics", "pixel_acc")},
         "video": classification_fields,
         "optical_flow": set(),
--- a/test/test_extended_models.py
+++ b/test/test_extended_models.py
@@ -98,7 +98,7 @@ def test_schema_meta_validation(model_fn):
         "all": {"metrics", "min_size", "num_params", "recipe"},
         "models": classification_fields | {"_docs"},
         "detection": {"categories", ("metrics", "box_map")},
-        "quantization": classification_fields | {"backend", "unquantized"},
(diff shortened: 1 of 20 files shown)


test/test_extended_models.py
```diff
@@ -95,8 +95,8 @@ def test_schema_meta_validation(model_fn):
     # mandatory fields for each computer vision task
     classification_fields = {"categories", ("metrics", "acc@1"), ("metrics", "acc@5")}
     defaults = {
-        "all": {"metrics", "min_size", "num_params", "recipe"},
-        "models": classification_fields | {"_docs"},
+        "all": {"metrics", "min_size", "num_params", "recipe", "_docs"},
```
NicolasHug, line 98: Now that this is mandatory for all weights, we can change these lines

https://github.com/pytorch/vision/blob/ae1d70713a542aac905cf403844a21b38b2be593/docs/source/conf.py#L348-L350

into 

```py
 lines += [meta_with_metrics.pop("_docs")]   # Custom per-Weights docs 
```

torchvision/models/detection/faster_rcnn.py
```diff
@@ -454,7 +458,7 @@ def fasterrcnn_resnet50_fpn(
 ) -> FasterRCNN:
     """
     Faster R-CNN model with a ResNet-50-FPN backbone from the `Faster R-CNN: Towards Real-Time Object
-    Detection with Region Proposal Networks <https://arxiv.org/abs/1703.06870>`__
+    Detection with Region Proposal Networks <https://arxiv.org/abs/1506.01497>`__
```
NicolasHug, line 461: Nice catch!

NicolasHug (approved):
Thanks @datumbox , LGTM!

I haven't checked the rendering but I'm sure we'll catch major problems (if any) soon enough.

test/test_extended_models.py
datumbox: good spot done.

PR #6041, opened 2022-05-18
add support for BoundinBox and SegmentationMask to RandomResizeCrop

pmeier:
Kernels were already there, so this only adds them to the transform.

Test failures are real though. @datumbox we get bitten now but what I mentioned in https://github.com/pytorch/vision/pull/5487#issuecomment-1053967688: although the transformation is perfectly able to work with only bounding boxes or segmentation masks, our choice of only extracting the image size from images, forces the user to always include an image in the inputs.

### DIFF
--- a/torchvision/prototype/transforms/_utils.py
+++ b/torchvision/prototype/transforms/_utils.py
@@ -20,7 +20,7 @@ def query_image(sample: Any) -> Union[PIL.Image.Image, torch.Tensor, features.Im
     try:
         return next(query_recursively(fn, sample))[1]
     except StopIteration:
-        raise TypeError("No image was found in the sample")
+        raise TypeError("No image was found in the sample") from None
 
 
 def get_image_dimensions(image: Union[PIL.Image.Image, torch.Tensor, features.Image]) -> Tuple[int, int, int]:
--- a/test/test_prototype_transforms.py
+++ b/test/test_prototype_transforms.py
@@ -3,7 +3,12 @@ import itertools
 import pytest
 import torch
 from common_utils import assert_equal
-from test_prototype_transforms_functional import make_images, make_bounding_boxes, make_one_hot_labels
+from test_prototype_transforms_functional import (
+    make_images,
+    make_bounding_boxes,
+    make_one_hot_labels,
+    make_segmentation_masks,
+)
 from torchvision.prototype import transforms, features
 from torchvision.transforms.functional import to_pil_image, pil_to_tensor
 
@@ -153,6 +158,8 @@ class TestSmoke:
                 transforms.RandomResizedCrop([16, 16]),
                 itertools.chain(
                     make_images(extra_dims=[(4,)]),
+                    make_bounding_boxes(),
+                    make_segmentation_masks(),
                     make_vanilla_tensor_images(),
                     make_pil_images(),
                 ),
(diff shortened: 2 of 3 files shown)


datumbox:
@pmeier As discussed previously, the example you provide at https://github.com/pytorch/vision/pull/5487#issuecomment-1053967688 is not a valid real-world use-case. Aka resizing just the bbox without the image will lead to corrupted information. Hence requiring always to have the image in such transforms makes sense. If you manage to find another valid counter example, I'm happy to revise. But unless we have that, there is no need to create a solution that covers for a corner-case that is not needed.

Concerning the failure, I think it's a library is missing:
```
ImportError: libcrypto.so.10: cannot open shared object file: No such file or directory
```

A potentially easy fix is to install the dependencies that probably are missing. See [here](https://askubuntu.com/questions/339364/libssl-so-10-cannot-open-shared-object-file-no-such-file-or-directory).

datumbox:
> If Pad would not support bounding boxes individually, we couldn't use it the way we do now. I'm aware that this composite approach has other issues as well, but it is the best we got right now.

It's worth noting that you are using Pad like that to circumvent a design limitation of the API. You are not using it because Padding a bbox on its own makes sense but rather as a workaround. I think all these are indications that the proposal can be improved. My concern about the use of "primitive transforms" is that I don't know exactly what's their difference with the kernels rather than adding dispatching. Perhaps the whole thing can be solved by moving the dispatch on _Feature objects.

pmeier:
Re test failures: the missing dependency only happens on Linux. I've opened pytorch/data#418 to check if this is a mistake on their side. The real failures are visible on [Windows](https://github.com/pytorch/vision/runs/6484672273) and [macOS](https://github.com/pytorch/vision/runs/6484672358).

datumbox:
@pmeier Thanks, yes I've seen it. I don't remember all the details TBH but I think you could go around the issue by dispatching to the low-level kernel with our normal ifs dispatch approach, right?

torchvision/prototype/transforms/_utils.py
```diff
@@ -20,7 +20,7 @@ def fn(
     try:
         return next(query_recursively(fn, sample))[1]
     except StopIteration:
-        raise TypeError("No image was found in the sample")
+        raise TypeError("No image was found in the sample") from None
```
vfdev-5, line 23: Why do you "from None" here ?

pmeier: Sorry, forgot to add a comment. The `StopIteration` is an internal detail that does not need to be propagated outside. By adding `from None` the error will look like a normal exception

```py
try:
    next(iter([]))
except StopIteration:
    raise TypeError("Argh!") from None
```

```
TypeError: Argh!
```

Not doing that gives

```py
try:
    next(iter([]))
except StopIteration:
    raise TypeError("Argh!")
```

```
StopIteration

[...]
TypeError: Argh!
```

pmeier:
> you could go around the issue by dispatching to the low-level kernel with our normal ifs dispatch approach, right?

Not without duplicating all the other logic that happens in `Pad._transform`, which is quite a lot:

https://github.com/pytorch/vision/blob/edb7bbbdacb73d258c222805b10a7a1b9ba0cd3e/torchvision/prototype/transforms/_geometry.py#L313-L350

> It's worth noting that you are using Pad like that to circumvent a design limitation of the API. [...] I think all these are indications that the proposal can be improved. 

True. Let's postpone this discussion until we have decided on a general way to handle composite transforms. If we for example reinstate the high-level dispatchers, my argument is moot. Otherwise it might still apply.

PR #6042, opened 2022-05-18
Check `torchdata` import on Linux

pmeier:
Investigation for pytorch/data#418.


pmeier:
Without installing anything extra:

:x: [Linux](https://github.com/pytorch/vision/runs/6487171577)
:heavy_check_mark: [Windows](https://github.com/pytorch/vision/runs/6487171676)
:heavy_check_mark: [macOS](https://github.com/pytorch/vision/runs/6487171774)

pmeier:
Installing additional system libraries like `torchdata` does [here](https://github.com/pytorch/data/blob/f80d07b0670ca2c68f06e4dfe236a6dbe6dd1b28/.github/workflows/_build_test_upload.yml#L236) is [not sufficient](https://github.com/pytorch/vision/runs/6488772795).

PR #6049, opened 2022-05-18
Models doc revamp: final cleanup

NicolasHug:
("Final" until "Final for real" and "Final for real V2"  comes along, obviously).

This PR:

- replaces `models.rst` with `models_new.rst` (diff looks huge but it's really just that (+ 1 ref fix))
- puts back the "fail on warning" option when building the docs
- removes the conversion of `-` into `_` in metrics.


NicolasHug:
I also ended up putting the torchhub section where you originally had it @datumbox db7e206

datumbox (approved):
LGTM, thanks @NicolasHug! Let's merge this "final" PR before the next "final v2". ;)

Issue #5705, opened 2022-03-30
CelebA download is broken


pmeier:
The download of all CelebA files except `identity_CelebA.txt` is broken. For example, the URL to download `img_align_celeba.zip` resolves to https://drive.google.com/uc?id=0B7EVK8r0v71pZjFTYXZWM3FlRnM&export=download. This link is publicly accessible, but you have to be logged into Google. Otherwise you'll see a 404 page.

I'll have a look if it is possible to get a general download link from the ID.

cc @pmeier @YosuaMichael

pmeier:
It seems the author has intentionally restricted the visibility for these files for some reason. If that is true, I think there is no way for us to provide an automatic download.

pmeier:
It seems something on the GDrive side has changed. The link above now gives a 403 (forbidden) instead a 404 page when you are logged. When you are not logged in you will be prompted to do so and afterwards see the same 403 page.

Exporting the link manually gives https://drive.google.com/file/d/0B7EVK8r0v71pZjFTYXZWM3FlRnM/view. While logged in you get transferred to the download page. If you are not logged in, you get the login prompt but land on page stating that access need to be requested.

For other datasets, e.g. Caltech101, both link variants are equivalent:

- https://drive.google.com/uc?id=137RyRjvTBkBiIfeYBNZBtViDHQ6_Ewsp&export=download
- https://drive.google.com/file/d/137RyRjvTBkBiIfeYBNZBtViDHQ6_Ewsp/view

We prefer the first, because it has less redirects and checks.

In any case it seems that Celeba is no longer automatically downloadable unless you are logged in. Thus, I propose we disable the download functionality preferably before the upcoming release. @NicolasHug Thoughts?

NicolasHug:
Thanks for reminding me and for your investigations @pmeier !

What do you mean by "being logged in"? Is it logged in from a browser?

pmeier:
> What do you mean by "being logged in"? Is it logged in from a browser?

Yes. In particular, we need the session cookies. As possible way is to ask users to export the cookies to a file from the browser that we can read in. But this far from an "automated approach".

AFAIK, there is no way to login from the command line or through env vars or the like.

NicolasHug:
Thanks for the deets. I agree we should deactivate it.

pmeier:
Closed in #6052.

NicolasHug:
Should we keep it open? Ultimately we'll want to put back the download feature, if the Gdrive becomes available again?

[...]

PR #6052, opened 2022-05-19
Deactivate CelebA download

NicolasHug:
Deactivate CelebA download as decided in  https://github.com/pytorch/vision/issues/5705


torchvision/datasets/celeba.py
```diff
@@ -38,6 +38,12 @@ class CelebA(VisionDataset):
         download (bool, optional): If true, downloads the dataset from the internet and
             puts it in root directory. If dataset is already downloaded, it is not
             downloaded again.
+
+            .. warning::
```
pmeier, line 42: Shouldn't we replace the original description with the warning? With this path the original description is wrong, since we will error out earlier.

torchvision/datasets/celeba.py
```diff
@@ -146,10 +152,15 @@ def download(self) -> None:
...
+        raise ValueError(
+            "Downloading CelebA is not supported anymore as of 0.13. See "
+            "https://github.com/pytorch/vision/issues/5705 for more details."
+        )
+        # TODO maybe uncomment lines below to put back download if it ever gets fixed.
+        # for (file_id, md5, filename) in self.file_list:
+        #     download_file_from_google_drive(file_id, os.path.join(self.root, self.base_folder), filename, md5)
 
-        extract_archive(os.path.join(self.root, self.base_folder, "img_align_celeba.zip"))
+        # extract_archive(os.path.join(self.root, self.base_folder, "img_align_celeba.zip"))
```
pmeier, line 163: Should we keep this? Otherwise the user does need to download *and* extract.

pmeier, line 163: Should we keep this? Otherwise the user does need to download *and* extract.

torchvision/datasets/celeba.py
```diff
@@ -38,6 +38,12 @@ class CelebA(VisionDataset):
         download (bool, optional): If true, downloads the dataset from the internet and
             puts it in root directory. If dataset is already downloaded, it is not
             downloaded again.
+
+            .. warning::
+
+                Downloading CelebA is not supported anymore as of 0.13. See
```
pmeier, line 44: This should also include instructions where to put the files.

pmeier, line 44: This should also include instructions where to put the files.

torchvision/datasets/celeba.py
```diff
@@ -146,10 +152,15 @@ def download(self) -> None:
...
 
-        for (file_id, md5, filename) in self.file_list:
-            download_file_from_google_drive(file_id, os.path.join(self.root, self.base_folder), filename, md5)
+        raise ValueError(
+            "Downloading CelebA is not supported anymore as of 0.13. See "
+            "https://github.com/pytorch/vision/issues/5705 for more details."
+        )
+        # TODO maybe uncomment lines below to put back download if it ever gets fixed.
+        # for (file_id, md5, filename) in self.file_list:
+        #     download_file_from_google_drive(file_id, os.path.join(self.root, self.base_folder), filename, md5)
```
pmeier, line 161: Can't we simply remove and look at the `git` history later?

torchvision/datasets/celeba.py
pmeier: Not there, but we could adapt the integrity check to check if we find only the archive and extract it:

https://github.com/pytorch/vision/blob/77cad12779a237c6fb415f48243bc149e9ecc36a/torchvision/datasets/celeba.py#L141-L142

NicolasHug: Do we extract on behalf of the user on other datasets where download isn't available?
I won't have time to do that today unfortunately, considering we have to test all this manually

pmeier: IIRC, `ImageNet` is the only dataset where we deactivated the download after it was available at first. For it, we do extract for the users:

https://github.com/pytorch/vision/blob/77cad12779a237c6fb415f48243bc149e9ecc36a/torchvision/datasets/imagenet.py#L61-L65

Other datasets that don't have a download functionality at all assume that the users do it themselves

https://github.com/pytorch/vision/blob/d57f929da98f927365bdef303d3453712067e15b/torchvision/datasets/coco.py#L15-L16

(Don't mind the "download" in the description. That is wrong as we don't offer anything for COCO)

NicolasHug: Thanks for the details.
Since in the majority of cases we don't extract for on behalf of users, I'd prefer merging this and just add to the instructions that they need to extract themselves (I updated the message in a previous commit)

pmeier (approved):
Thanks Nicolas!

PR #6053, opened 2022-05-19
Adding multi-layer perceptron in ops

datumbox:
We should avoid using ViT's MLP block from Swin:
https://github.com/pytorch/vision/blob/9d9cfab21e8fd7e930b1ef2aae02fe87e08e1f71/torchvision/models/swin_transformer.py#L15

The specific layer is very common and has been previously requested at #4333

This PR:
- [x] Adds a generic MLP block to TorchVision. 
- [x] Handles BC on ViT.
- [x] Replaces `MLPBlock` with `MLP` on Swin and patches the unreleased weights (weights uploaded on S3 and manifold).

References:
- [TorchVision](https://github.com/pytorch/vision/blob/main/torchvision/models/vision_transformer.py#L40)
- [Multimodal](https://github.com/facebookresearch/multimodal/blob/5dec8a6b897a706cbf3cdadb96b1e464afae11cd/torchmultimodal/modules/layers/mlp.py#L13) 
- [DETR](https://github.com/facebookresearch/detr/blob/eb9f7e03ed8e2ed2cd55528989fe7df890bc3fc0/models/detr.py#L289)
- [TF](https://www.tensorflow.org/recommenders/api_docs/python/tfrs/layers/blocks/MLP)

Proof that the new API doesn't break Swin. The minor differences are expected due to the known non-deterministic behaviour of some kernels:
```
srun -p dev --cpus-per-task=96 -t 24:00:00 --gpus-per-node=1 torchrun --nproc_per_node=1 train.py --model swin_t --test-only -b 1 --weights="Swin_T_Weights.IMAGENET1K_V1"
Test:  Acc@1 81.476 Acc@5 95.780

srun -p dev --cpus-per-task=96 -t 24:00:00 --gpus-per-node=1 torchrun --nproc_per_node=1 train.py --model swin_s --test-only -b 1 --weights="Swin_S_Weights.IMAGENET1K_V1"
Test:  Acc@1 83.182 Acc@5 96.366

srun -p dev --cpus-per-task=96 -t 24:00:00 --gpus-per-node=1 torchrun --nproc_per_node=1 train.py --model swin_b --test-only -b 1 --weights="Swin_B_Weights.IMAGENET1K_V1"
Test:  Acc@1 83.584 Acc@5 96.636
```

cc @ankitade


torchvision/models/vision_transformer.py
```diff
@@ -37,21 +37,47 @@ class ConvStemConfig(NamedTuple):
...
+        prefix,
+        local_metadata,
+        strict,
+        missing_keys,
+        unexpected_keys,
+        error_msgs,
+    ):
+        version = local_metadata.get("version", None)
+
+        if version is None or version < 2:
```
NicolasHug, line 64: Maybe add a comment linking to this PR to explain where this comes from?

torchvision/ops/misc.py
```diff
@@ -253,3 +253,48 @@ def _scale(self, input: Tensor) -> Tensor:
     def forward(self, input: Tensor) -> Tensor:
         scale = self._scale(input)
         return scale * input
+
+
+class MLP(torch.nn.Sequential):
+    """
+    This block implements the multi-layer perceptron (MLP) module.
```
NicolasHug, line 260: Nit
```suggestion
    """This block implements the multi-layer perceptron (MLP) module.
```

torchvision/ops/misc.py
```diff
@@ -253,3 +253,48 @@ def _scale(self, input: Tensor) -> Tensor:
...
+
+    Args:
+        in_channels (int): Number of channels of the input
+        hidden_channels (List[int]): List of the hidden channel dimensions
+        out_channels (int): Number of channels of the output
+        norm_layer (Callable[..., torch.nn.Module], optional): Norm layer that will be stacked on top of the convolution layer. If ``None`` this layer wont be used. Default: ``None``
+        activation_layer (Callable[..., torch.nn.Module], optional): Activation function which will be stacked on top of the normalization layer (if not None), otherwise on top of the conv layer. If ``None`` this layer wont be used. Default: ``torch.nn.ReLU``
```
NicolasHug, line 267: Annotations in docstring make me sad :'(

torchvision/ops/misc.py
```diff
@@ -253,3 +253,48 @@ def _scale(self, input: Tensor) -> Tensor:
...
+        for hidden_dim in hidden_channels:
+            layers.append(torch.nn.Linear(in_dim, hidden_dim, bias=bias))
+            if norm_layer is not None:
+                layers.append(norm_layer(hidden_dim))
+            layers.append(activation_layer(**params))
+            layers.append(torch.nn.Dropout(dropout, **params))
+            in_dim = hidden_dim
+
+        layers.append(torch.nn.Linear(in_dim, out_channels, bias=bias))
+        layers.append(torch.nn.Dropout(dropout, **params))
```
NicolasHug, line 297: isn't this layer also considered "hidden", technically?

torchvision/ops/misc.py
datumbox: I know how you feel above this. :( I think all the callables all over TorchVision are added like that to provide info on what they are supposed to return.

torchvision/ops/misc.py
datumbox: Sorry I've missed your point. What change do you propose?

datumbox: Are you suggesting that hidden_dims should include output_channels? Possible, let me give this a try.

NicolasHug: That's my understanding of what a "hidden layer" is: everything that's not the input or the output. But I don't know if we're being consistent about that across the other ops / models TBH.

jdsgomes (approved):
LGTM, feel free to merge after the same changes are done in Swin. If you prefer I can have a second look once thats done

torchvision/models/swin_transformer.py
```diff
@@ -5,14 +5,14 @@
...
+from ..ops.misc import MLP
 from ..ops.stochastic_depth import StochasticDepth
 from ..transforms._presets import ImageClassification, InterpolationMode
 from ..utils import _log_api_usage_once
 from ._api import WeightsEnum, Weights
 from ._meta import _IMAGENET_CATEGORIES
 from ._utils import _ovewrite_named_param
-from .convnext import Permute
-from .vision_transformer import MLPBlock
+from .convnext import Permute  # TODO: move Permute on ops
```
datumbox, line 15: This is a straight move from convnext to ops (no weight patching needed) but to avoid doing everything on a single PR I plan to do it on a follow up.

torchvision/ops/misc.py
```diff
@@ -253,3 +253,47 @@ def _scale(self, input: Tensor) -> Tensor:
...
+
+    def __init__(
+        self,
+        in_channels: int,
+        hidden_channels: List[int],
+        norm_layer: Optional[Callable[..., torch.nn.Module]] = None,
+        activation_layer: Optional[Callable[..., torch.nn.Module]] = torch.nn.ReLU,
+        inplace: Optional[bool] = True,
+        bias: bool = True,
+        dropout: float = 0.0,
```
oke-aditya, line 279: I'm slightly confused over this choice. 

We should rather not append dropout layer if dropout is 0.0. 

Maybe 
```suggestion
        dropout: Optional[float] = None,
```

And check if dropout is not None and then add the layer?

datumbox: That was originally the plan but it would make it very hard to maintain BC for MLPBlock. See ViT's usage for more details. It was possible but would require a lot of ugly code. Leaving the dropout layer should have no performance issues.

oke-aditya: While Dropout is not going to cause any performance issues. But I'm afraid we aren't encouraging best practices. 

I thought we could just leave MLP in ViT as it as (not inherit from ops) And make the MLP layer a bit better as suggested. 

Again it's either way, both have pros and cons.

torchvision/ops/misc.py
```diff
@@ -253,3 +253,47 @@ def _scale(self, input: Tensor) -> Tensor:
...
+        for hidden_dim in hidden_channels[:-1]:
+            layers.append(torch.nn.Linear(in_dim, hidden_dim, bias=bias))
+            if norm_layer is not None:
+                layers.append(norm_layer(hidden_dim))
+            layers.append(activation_layer(**params))
+            layers.append(torch.nn.Dropout(dropout, **params))
+            in_dim = hidden_dim
+
+        layers.append(torch.nn.Linear(in_dim, hidden_channels[-1], bias=bias))
+        layers.append(torch.nn.Dropout(dropout, **params))
```
thomasbbrunner, line 296: It is not very clear for me why there's a Dropout layer after the last layer. I saw that it was present in the previous MLPBlock class, but no other implementation of MLP with dropout (that I could find) has a dropout layer on the output. Including the one in the [multimodal package](https://github.com/facebookresearch/multimodal/blob/cef848f3d0313d4a682f2b10ca7f8506da820451/torchmultimodal/modules/layers/mlp.py#L62).

Maybe this was something specific for the use of the MLPBlock? If so, this should not be in this class.

datumbox: You are right there are various implementations of MLP, some of which don't have at all dropout, some have at the middle but not at the end or some have everywhere. If you check the references, you will see that all patterns exist. Our implementation is like that because it replaces MLP layers used in existing models like ViT and Swin. We also try to  support more complex variations with more than 2 linears. You observation is correct though that if one wanted to avoid having dropout at the end, the current implementation wouldn't let them. Since that variant is also valid, perhaps it's worth making this update in a non-BC way with perhaps a new boolean that controls the appearance of Dropout at the end or not. WDYT?

thomasbbrunner: I think your suggestion sounds very nice. What would be the default value be for the boolean? I guess that setting it to True (with dropout) would cause no breaking changes. At the same time, I would say that not having a dropout in the last layer is more common (default) configuration? Also, I'd be intested in working on this, whichever option is chosen.

torchvision/ops/misc.py
thomasbbrunner: Would it be an option to implement this as it was done in the [multimodal package](https://github.com/facebookresearch/multimodal/blob/cef848f3d0313d4a682f2b10ca7f8506da820451/torchmultimodal/modules/layers/mlp.py#L59)? 
```
if dropout > 0:
    layers.append(nn.Dropout(dropout))
```

torchvision/ops/misc.py
datumbox: > I guess that setting it to True (with dropout) would cause no breaking changes.

Yes, you are right we will need to maintain BC. Note that using True is the "default" setup on TorchVision at the moment as literally all existing models require dropout everywhere.

> I'd be intested in working on this

Sounds great, let me recommend the following. Could you start an issue, summarizing what you said here and providing a few references of the usage of MLP with a middle dropout but without the final one? Providing a few examples from real-world vision architectures will help build a stronger case. Once we clarify the details on the issue, we can discuss a potential PR. 😃

thomasbbrunner: Ok! I am a bit short on time at the moment, but will have more time in the upcoming weeks. Nevertheless, I'm interested in this and will be working on it!

PR #6054, opened 2022-05-19
Minor Swin Transformer fixes

datumbox:
Fix a few minor omissions on Swin, such as adding it on the hubconfig and including the rest of the variants in the slow_models list.


jdsgomes (approved):
LGTM - thanks for the fix!

PR #6055, opened 2022-05-19
Move Permute layer to ops

datumbox:
Resolving https://github.com/pytorch/vision/pull/6053#discussion_r877333538

### DIFF
--- a/docs/source/ops.rst
+++ b/docs/source/ops.rst
@@ -87,8 +87,9 @@ TorchVision provides commonly used building blocks as layers:
     DeformConv2d
     DropBlock2d
     DropBlock3d
-    MLP
     FrozenBatchNorm2d
+    MLP
+    Permute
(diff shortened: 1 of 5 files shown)


NicolasHug (approved):
LGTM.

For my own understanding, do we need a `Permute` module, or could we just rely on `torch.permute`? Looks like tihs `Permute` is a very simple wrapper

datumbox:
@NicolasHug you are right that `Permute` is nothing but a wrapper for `torch.permute`. Unfortunately the latter can't be used as part of sequentials, which are very common in many of our models. I thought that instead of rewriting this over and over, it would be useful to have it as a shortcut on ops but I don't have strong opinions. Thoughts?

datumbox:
As discussed offline with @NicolasHug, it's OK to have this as an op.

PR #6056, opened 2022-05-20
Remove deprecation warning for pts on read_video

datumbox:
Addressing https://github.com/pytorch/vision/pull/6013#discussion_r877965119

I've opted to comment it out instead of removing it to leave some kind of notice about the potential issue of using pst. Happy to remove it.


bjuncek (approved):
thanks, LGTM

datumbox:
I think we should not merge this and instead finalize and merge #5952

PR #6058, opened 2022-05-20
Remove `(N, T, H, W, C) => (N, T, C, H, W)` from presets

datumbox:
Should be merged after #5952.

In this PR we:
- [x] Remove the `(N, T, H, W, C) => (N, T, C, H, W)` conversion from the presets
- [x] Update the example on the docs to explicitly permute the output of `read_video()`
- [x] Add `output_format` param in `Kinetics` dataset and `read_video()` usages.
- [x] Add better descriptions on presets to describe all accepted inputs and their dimensions.
- [ ] Verify that the accuracy of pre-trained models using the reference scripts

```
torchrun --nproc_per_node=8 train.py --data-path /datasets01/kinetics/070618/400/ --batch-size=16 --sync-bn --test-only --weights R2Plus1D_18_Weights.KINETICS400_V1 --cache-dataset
```

### DIFF
--- a/docs/source/models.rst
+++ b/docs/source/models.rst
@@ -471,8 +471,7 @@ Here is an example of how to use the pre-trained video classification models:
     from torchvision.io.video import read_video
     from torchvision.models.video import r3d_18, R3D_18_Weights
 
-    vid, _, _ = read_video("test/assets/videos/v_SoccerJuggling_g23_c01.avi")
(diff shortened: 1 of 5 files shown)


NicolasHug (approved):
Thanks @datumbox

datumbox:
The macOS and ONNX jobs are broken and unrelated to this PR. Merging as discussed with @NicolasHug offline.

PR #1457, opened 2019-10-14
Remove download for ImageNet

pmeier:
addresses #1453 

I introduced the function `parse_train_archive` and `parse_val_archive` and changed `parse_devkit` to take in an `str` pointing to an archive. Users that downloaded the dataset externally can use them to prepare the folders.

PR #6059, opened 2022-05-20
Indicate Celeba download parameter is deprecated and will be removed

NicolasHug:
Addresses https://github.com/pytorch/vision/issues/5705#issuecomment-1132837801

I'm keeping the ValueError instead of a deprecation warning: raising a warning would lead to an error later anyway since the files aren't available. Raising an error early is much cleaner


pmeier:
When we did that for ImageNet in #1457, we used the following idiom:

```py
def __init__(self, ..., download=None):
    if download is True:
        raise RuntimeError
    elif download is False:
        warnings.warn(...)
```

Meaning, if the user didn't pass `download` before nothing happens. If they explicitly set `donwload=False`, they will be warned that the parameter is deprecated. If they set `download=True`, they will get an error saying that download can no longer be performed.

PR #6060, opened 2022-05-20
Replace torchvision.datasets.utils with functionality from torchdata

pmeier:
As the title implies, we want to get rid of as much functionality as possible from `torchvision.datasets.utils` to avoid duplication between the libraries. This PR adds an initial draft what already can be achieved and what has to be ported to `torchdata` before we can depend on it.

### DIFF
--- a/torchvision/prototype/datasets/_builtin/__init__.py
+++ b/torchvision/prototype/datasets/_builtin/__init__.py
@@ -1,22 +1,26 @@
 from .caltech import Caltech101, Caltech256
-from .celeba import CelebA
+
+# from .celeba import CelebA
 from .cifar import Cifar10, Cifar100
-from .clevr import CLEVR
-from .coco import Coco
-from .country211 import Country211
-from .cub200 import CUB200
-from .dtd import DTD
-from .eurosat import EuroSAT
+
+# from .clevr import CLEVR
+# from .coco import Coco
+# from .country211 import Country211
+# from .cub200 import CUB200
+# from .dtd import DTD
+# from .eurosat import EuroSAT
 from .fer2013 import FER2013
-from .food101 import Food101
-from .gtsrb import GTSRB
+
+# from .food101 import Food101
+# from .gtsrb import GTSRB
 from .imagenet import ImageNet
-from .mnist import MNIST, FashionMNIST, KMNIST, EMNIST, QMNIST
-from .oxford_iiit_pet import OxfordIIITPet
-from .pcam import PCAM
-from .sbd import SBD
-from .semeion import SEMEION
-from .stanford_cars import StanfordCars
-from .svhn import SVHN
-from .usps import USPS
-from .voc import VOC
+
+# from .mnist import MNIST, FashionMNIST, KMNIST, EMNIST, QMNIST
+# from .oxford_iiit_pet import OxfordIIITPet
+# from .pcam import PCAM
+# from .sbd import SBD
+# from .semeion import SEMEION
+# from .stanford_cars import StanfordCars
+# from .svhn import SVHN
+# from .usps import USPS
+# from .voc import VOC
--- a/torchvision/prototype/datasets/utils/_dataset.py
+++ b/torchvision/prototype/datasets/utils/_dataset.py
@@ -42,7 +42,7 @@ class Dataset(IterDataPipe[Dict[str, Any]], abc.ABC):
         yield from self._dp
 
     @abc.abstractmethod
-    def _resources(self) -> List[OnlineResource]:
+    def _resources(self) -> Sequence[OnlineResource]:
         pass
 
     @abc.abstractmethod
--- a/torchvision/prototype/datasets/utils/_resource.py
+++ b/torchvision/prototype/datasets/utils/_resource.py
@@ -1,10 +1,11 @@
+from __future__ import annotations
+
 import abc
-import hashlib
-import itertools
 import pathlib
-from typing import Optional, Sequence, Tuple, Callable, IO, Any, Union, NoReturn, Set
+from typing import Optional, Tuple, Callable, BinaryIO, Any, Union, NoReturn, Set
 from urllib.parse import urlparse
 
+from torch.hub import tqdm
 from torchdata.datapipes.iter import (
     IterableWrapper,
     FileLister,
@@ -13,27 +14,23 @@ from torchdata.datapipes.iter import (
     ZipArchiveLoader,
     TarArchiveLoader,
     RarArchiveLoader,
+    OnlineReader,
+    HashChecker,
 )
-from torchvision.datasets.utils import (
-    download_url,
-    _detect_file_type,
-    extract_archive,
-    _decompress,
-    download_file_from_google_drive,
-    _get_redirect_url,
-    _get_google_drive_file_id,
-)
+from torchvision.datasets.utils import _detect_file_type, extract_archive, _decompress
 from typing_extensions import Literal
 
 
 class OnlineResource(abc.ABC):
     def __init__(
         self,
+        url: str,
         *,
         file_name: str,
         sha256: Optional[str] = None,
         preprocess: Optional[Union[Literal["decompress", "extract"], Callable[[pathlib.Path], None]]] = None,
     ) -> None:
+        self.url = url
         self.file_name = file_name
         self.sha256 = sha256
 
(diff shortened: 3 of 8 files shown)


torchvision/prototype/datasets/_builtin/__init__.py
```diff
@@ -1,22 +1,26 @@
 from .caltech import Caltech101, Caltech256
-from .celeba import CelebA
+
```
pmeier, line 2: I only made changes to four datasets to cover all cases, i.e.

- HTTP download (CIFAR)
- GDrive download (Caltech)
- Kaggle download (FER2013)
- Manual download (ImageNet)

The others are commented out to be able to import `torchvision`.

torchvision/prototype/datasets/utils/_dataset.py
```diff
@@ -42,7 +42,7 @@ def __iter__(self) -> Iterator[Dict[str, Any]]:
         yield from self._dp
 
     @abc.abstractmethod
-    def _resources(self) -> List[OnlineResource]:
+    def _resources(self) -> Sequence[OnlineResource]:
```
pmeier, line 45: This gets rid of the annoying `mypy` warning when returning `List[ManualDownloadResource]` Instead of `List[OnlineResource]`.

torchvision/prototype/datasets/utils/_resource.py
```diff
@@ -57,7 +54,42 @@ def _extract(file: pathlib.Path) -> None:
...
         _decompress(str(file), remove_finished=True)
 
-    def _loader(self, path: pathlib.Path) -> IterDataPipe[Tuple[str, IO]]:
+    @classmethod
+    def from_http(cls, url: str, *, file_name: Optional[str] = None, **kwargs: Any) -> OnlineResource:
+        return cls(url, file_name=file_name or pathlib.Path(urlparse(url).path).name, **kwargs)
+
+    @classmethod
+    def from_gdrive(cls, id: str, **kwargs: Any) -> OnlineResource:
+        return cls(f"https://drive.google.com/uc?export=download&id={id}", **kwargs)
```
pmeier, line 63: By using the `OnlineReader` datapipe, we no longer need different download functionality for HTTP or GDrive. Thus, I was able to remove the subclasses and replace them with these thin instantiation class methods.

If we switch to `iopath` in the future, we can probably remove them as well.

torchvision/prototype/datasets/utils/_resource.py
```diff
@@ -57,7 +54,42 @@ def _extract(file: pathlib.Path) -> None:
...
+            stream = list(dp)[0][1]
+
+            with open(file, "wb") as fh, tqdm() as progress_bar:
+                for chunk in iter(lambda: stream.read(1024 * 1024), b""):  # type: ignore[no-any-return]
+                    # filter out keep-alive new chunks
+                    if not chunk:
+                        continue
+
+                    fh.write(chunk)
+                    progress_bar.update(len(chunk))
```
pmeier, line 82: There a two reasons I didn't do this with datapipes:

1. We should probably have a progress bar for downloads, since they usually are quite large and a long running process without output is always suspicious. By importing `tqdm` from `torch.hub` we get vanilla `tqdm` if it is installed and a minimal port if it is not.
2. AFAIK, `torchdata` has no builtin option to read chunks from a stream. We need that, because we otherwise load the whole file into memory before saving it.

cc @ejguan @NivekT

torchvision/prototype/datasets/utils/_resource.py
```diff
@@ -57,7 +54,42 @@ def _extract(file: pathlib.Path) -> None:
...
+
+        if self.sha256 and not skip_integrity_check:
+            dp = IterableWrapper([str(file)])
+            dp = FileOpener(dp, mode="rb")
+            dp = HashChecker(dp, {str(file): self.sha256}, hash_type="sha256")
+            list(dp)
```
pmeier, line 88: Even if we could do the above in datapipes, we cannot attach the `HashChecker` directly. It will consume the stream and if it is not seekable, e.g. a HTTP response, there is no data left to save afterwards. This can probably be done with a `OnDiskCacheHolder`, but let's not worry about that if we are already breaking out of the graph above.

torchvision/prototype/datasets/utils/_resource.py
```diff
@@ -57,7 +54,42 @@ def _extract(file: pathlib.Path) -> None:
...
+
+        return file
```
pmeier, line 90: Since we opted to only perform the preprocessing after the download in #5990, we can probably also move it here.

torchvision/prototype/datasets/utils/_resource.py
```diff
@@ -57,7 +54,42 @@ def _extract(file: pathlib.Path) -> None:
...
+
+    def download(self, root: Union[str, pathlib.Path], *, skip_integrity_check: bool = False) -> pathlib.Path:
+        root = pathlib.Path(root).expanduser().resolve()
+        root.mkdir(parents=True, exist_ok=True)
+        file = root / self.file_name
+
+        if not file.exists():
+            dp = IterableWrapper([self.url])
+            dp = OnlineReader(dp)
+            stream = list(dp)[0][1]
```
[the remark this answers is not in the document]
NivekT: I think that can be a useful functionality. We probably want to add it as a new DataPipe rather than modifying the `HttpReader`.

torchvision/prototype/datasets/utils/_resource.py
NivekT: 1. I agree having a progress bar for download is a good idea. Maybe `torchdata` can have a dependency for `tqdm`.
2. I suppose adding an option to return `r.iter_content()` instead of `r.raw` would work?

NivekT (commented):
> `OnlineReader` has no support for redirects. 

Agree we can add that to that and `HttpReader`.

> `GDriveReader` is missing functionality that we support. For example, we can deal with Virus scan warnings that popped up recently.

We should add that.
> For that we need automatic detection of the archive type.

Is it possible to extend `Decompressor` to do more of that? Or will you need something different?

torchvision/prototype/datasets/utils/_resource.py
pmeier: 1. Apart from [`tqdm`](https://github.com/tqdm/tqdm), [`rich`](https://github.com/Textualize/rich) also added support for [progress bars](https://rich.readthedocs.io/en/stable/progress.html) recently. That is what newer versions of `pip` are using. Just leaving this here to asses the options before deciding to depend on `tqdm`.
2. I would avoid having a flag the changes the output types. Having something like a `ChunkReader` datapipe can be useful in general not just for HTTP requests, right?

    ```py
    from torchdata.datapipes.iter import IterDataPipe
    
    
    class ChunkReader(IterDataPipe):
        def __init__(self, datapipe, *, chunk_size=32 * 1024 * 1024):
            self.datapipe = datapipe
            self.chunk_size = chunk_size
    
        def __iter__(self):
            for path, stream in self.datapipe:
                for chunk in iter(lambda: stream.read(self.chunk_size), b""):
                    # filter out keep-alive new chunks from HTTP streams
                    if not chunk:
                        continue
    
                    yield path, chunk
    ```

    A utility like that is needed anyway for `Saver` anyway. AFAIK, there is currently no builtin functionality to read data from a stream, correct? I'm guessing that is why the [`Saver` example](https://github.com/pytorch/data/blob/12cfaf8899b1337981cd4edf9deef127f925f1bd/torchdata/datapipes/iter/util/saver.py#L34-L36) uses `bytes` as input.

If there is more to discuss, let's migrate this to a proper issue on the `torchdata` repository.

pmeier:
> > For that we need automatic detection of the archive type.
> 
> Is it possible to extend `Decompressor` to do more of that? Or will you need something different?

Nope. As the name implies `Decompressor` just decompresses files, but does not "extract" archives. We need something similar to what `OnlineReader` is to `HttpReader` and `GDriveReader`, i.e. an `ArchiveLoader`. Something along the lines:

```py
class ZipArchiveLoader(IterDataPipe):
    def __init__(self, datapipe):
        self.datapipe = datapipe

    @staticmethod
    def _load(path, stream):
        for info in zipfile.ZipFile(stream):
            extracted_path = ...
            extracted_stream = ...
            yield os.path.join(path, extracted_path), StreamWrapper(extracted_stream)

    def __iter__(self):
        for path, stream in self.datapipe:
            yield from self._load(path, stream)


class TarArchiveLoader(IterDataPipe):
    def __init__(self, datapipe):
        self.datapipe = datapipe

    @staticmethod
    def _load(path, stream):
        for info in tarfile.open(stream):
            extracted_path = ...
            extracted_stream = ...
            yield os.path.join(path, extracted_path), StreamWrapper(extracted_stream)

    def __iter__(self):
        for path, stream in self.datapipe:
            yield from self._load(path, stream)


class ArchiveLoader(IterDataPipe):
    def __init__(self, datapipe):
        self.datapipe = datapipe

    _LOADERS = {
        ".zip": ZipArchiveLoader._load,
        ".tar": TarArchiveLoader._load,
    }

    def __iter__(self):
        for path, stream in self.datapipe:
            load = self._LOADERS[os.path.splitext(path)[1]]
            yield from load(path, stream)
```

With this we can extract any archive with 

```py
dp = IterableWrapper([path_to_any_kind_of_archive])
dp = ArchiveLoader(dp)
dp = ChunkReader(dp)  # see comment above
dp = Saver(dp, ...)
list(dp)
```

Alternatively, I could also live with a `select_archive_loader` functionality that I can pass in the path of an arbitrary archive and get back the correct loader.

torchvision/prototype/datasets/utils/_resource.py
NivekT: I will open an issue in `torchdata` to get more thoughts from the community and see if there is anything that we may be missing.

IIUC, since `_get_response_from_http` returns `r.raw`, rather than say `r.iter_content()`, it loads the whole response into memory at once. Is that correct?

https://github.com/pytorch/data/blob/12cfaf8899b1337981cd4edf9deef127f925f1bd/torchdata/datapipes/iter/load/online.py#L46

pmeier: > I will open an issue in `torchdata` to get more thoughts from the community and see if there is anything that we may be missing.

See pytorch/data#439. It seems we missed `StreamReader`, which does exactly that. Still, could you open an issue regarding the progress bar?

> IIUC, since `_get_response_from_http` returns `r.raw`, rather than say `r.iter_content()`, it loads the whole response into memory at once. Is that correct?

`r.raw` just gives you the stream and is somewhat like `BinaryIO`, whereas `r.iter_content()` reads chunks from the stream and thus is a `Iterator[bytes]`. IMO returning `r.raw` is the right choice, since attaching a `StreamReader` to a `HTTPReader` will give us the `iter_content()` behavior.

torchvision/prototype/datasets/utils/_resource.py
pmeier: I was able to solve this with a `Forker` in 2d03f026d2c04c38b23540f95c71f2006e4fbbcb.

vadimkantorov:
Related on GDrive and other cloud downloads: https://github.com/pytorch/pytorch/issues/73466

PR #6061, opened 2022-05-20
Add output_format do video datasets and readers

NicolasHug:
Follow-up to https://github.com/pytorch/vision/pull/5952#pullrequestreview-980307726

This PR adds a new `output_format` to all video datasets, as well as `VideoClips()` and `read_video()`. The default is BC, so it corresponds to:

- `TCHW` for Kinetics
- `THWC` for everything else.

This is far from ideal, but this is the only way for users to have a consistent experience in torchvision. Right now, Kinetics is an outlier. For ref, the decision to have Kinetics follow the `TCHW` format comes from https://github.com/pytorch/vision/pull/3680#discussion_r646403139.

### DIFF
--- a/torchvision/datasets/kinetics.py
+++ b/torchvision/datasets/kinetics.py
@@ -62,11 +62,14 @@ class Kinetics(VisionDataset):
         download (bool): Download the official version of the dataset to root folder.
         num_workers (int): Use multiple workers for VideoClips creation
         num_download_workers (int): Use multiprocessing in order to speed up download.
+        output_format (str, optional): The format of the output video tensors (before transforms).
+            Can be either "THWC" or "TCHW" (default).
+            Note that in most other utils and datasets, the default is actually "THWC".
 
     Returns:
         tuple: A 3-tuple with the following entries:
 
-            - video (Tensor[T, C, H, W]): the `T` video frames in torch.uint8 tensor
+            - video (Tensor[T, C, H, W] or Tensor[T, H, W, C]): the `T` video frames in torch.uint8 tensor
             - audio(Tensor[K, L]): the audio frames, where `K` is the number of channels
               and `L` is the number of points in torch.float tensor
             - label (int): class of the video clip
@@ -106,6 +109,7 @@ class Kinetics(VisionDataset):
         _audio_samples: int = 0,
         _audio_channels: int = 0,
         _legacy: bool = False,
+        output_format: str = "TCHW",
     ) -> None:
 
         # TODO: support test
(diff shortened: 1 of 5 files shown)


torchvision/datasets/kinetics.py
```diff
@@ -233,9 +240,6 @@ def __len__(self) -> int:
 
     def __getitem__(self, idx: int) -> Tuple[Tensor, Tensor, int]:
         video, audio, info, video_idx = self.video_clips.get_clip(idx)
-        if not self._legacy:
-            # [T,H,W,C] --> [T,C,H,W]
-            video = video.permute(0, 3, 1, 2)
```
NicolasHug, line 238: This is now handled above when instantiating the `VideoClips()` object

torchvision/datasets/video_utils.py
```diff
@@ -366,6 +371,11 @@ def get_clip(self, idx: int) -> Tuple[torch.Tensor, torch.Tensor, Dict[str, Any]
             video = video[resampling_idx]
             info["video_fps"] = self.frame_rate
         assert len(video) == self.num_frames, f"{video.shape} x {self.num_frames}"
+
+        if self.output_format == "TCHW":
+            # [T,H,W,C] --> [T,C,H,W]
+            video = video.permute(0, 3, 1, 2)
```
datumbox, line 377: I think this might not work when `backend` != pyav. The `_read_video_from_file()` returns `Tensor[T, H, W, C]`, so we need to enforce the right format depending on the backend. This requires a bit more logic.

datumbox: I think the easiest way to do this is pass the parameter to `read_video()` at line 332 and potentially add a similar param to `_read_video_from_file()`.

NicolasHug: Are you sure? I'm converting the format at the very end, after both backend have done their magic.

> The _read_video_from_file() returns Tensor[T, H, W, C]

Agreed and this is what `read_video()` with `pyav` should return as well.

NicolasHug: The 2 backends being used is precisely why I didn't pass `read_video(output_format=output_format)` BTW

datumbox: You are right. Somehow I got confused and thought the two backends returned different order of dims. What you have will work. I'll check again on Monday with clearer head and merge.

Out of curiosity, what's the reason you opted for adding the parameter to `read_video()` but not to `_read_video_from_file()` and handle this in there uniformly?

NicolasHug: > adding the parameter to read_video() but not to _read_video_from_file() and handle this in there uniformly

I added this to all public-facing video readers (I think?) but I didn't want to bother with the private ones.

datumbox (approved):
LGTM, thanks @NicolasHug!

PR #6064, opened 2022-05-23
use torchdata datapipe linter for prototype dataset tests

pmeier:
Use the linter from pytorch/data#429 for our internal tests.


pmeier:
Oops, missed that, sorry.

pmeier:
@ejguan @NivekT The infinite recursion is back: https://github.com/pytorch/vision/runs/6551564325. It seems everything that traverses the graph like `_check_shuffle_before_sharding` should be tested against datapipes with circular references.

pmeier:
See pytorch/data#435.

PR #6065, opened 2022-05-23
expand prototype test matrix to different Python versions

pmeier:
This expands the prototype test matrix to test all Python versions on Linux. This should give us some extra safety to catch stuff like #5801 and https://github.com/pytorch/data/runs/6500848588#step:9:1977, which are not visible on Python 3.7.

Since GitHub actions is free for OSS projects, this does not add any costs.

### DIFF
--- a/.github/workflows/prototype-tests.yml
+++ b/.github/workflows/prototype-tests.yml
@@ -7,10 +7,18 @@ jobs:
   prototype:
     strategy:
       matrix:
+        python-version:
+          - "3.7"
+          - "3.8"
+          - "3.9"
+          - "3.10"
         os:
           - ubuntu-latest
-          - windows-latest
-          - macos-latest
+        include:
+          - python-version: "3.7"
+            os: windows-latest
+          - python-version: "3.7"
+            os: macos-latest
       fail-fast: false
 
     runs-on: ${{ matrix.os }}


pmeier:
After some offline discussion with @NicolasHug, we decided to only follow up on this after the release to avoid for the CI going red for something that does not need to be fixed before the release.

torchvision/prototype/datasets/_builtin/imagenet.py
```diff
@@ -46,9 +45,9 @@ def __init__(self, **kwargs: Any) -> None:
         super().__init__("Register on https://image-net.org/ and follow the instructions there.", **kwargs)
 
 
-class ImageNetDemux(enum.IntEnum):
-    META = 0
-    LABEL = 1
+# class ImageNetDemux(enum.IntEnum):
+#     META = 0
+#     LABEL = 1
```
pmeier, line 50:
```suggestion
```

pmeier:
@NicolasHug apart from the flaky open file handle warning, expanding the CI matrix also turned up warnings that #5219 tried to solve. I've added fixes for `ImageNet` and `CUB200` here and CI is now green. Do you want me to extract these changes out of this PR and send a separate one, or keep them here?

NicolasHug:
> Do you want me to extract these changes out of this PR and send a separate one

Yes please!

pmeier:
See #6076.

PR #6076, opened 2022-05-24
Fully exhaust datapipes that are needed to construct a dataset

pmeier:
Failures that will be fixed by this PR are only visible on Python >= 3.8. See #6065 for details.

---

Some datasets store a file inside an archive that needs to be read completely in order to construct the final datapipe. While it is tempting to do this in one `Demultiplexer`, this has the downside that (depending on how the archive is structured), the `Demultiplexer` now has items in the buffer before we actually start iterating on the final datapipe.

To avoid that we should always exhaust datapipes completely in case we need them while constructing another. This means, there are two changes to be made:

1. Remove the file we need to read from the `classify_fn` of the `Demultiplexer` and use a `Filter` to extract it separately.
2. Replace `next(iter(dp))` with `list(dp)` to make sure `dp` is exhausted.


pmeier:
After some offline discussion with @NicolasHug, it became clear that the situation is not as straight forward as I thought it was. For example, the child datapipes from the `Demultiplexer` seem to be [aware if an iteration is ongoing](https://github.com/pytorch/pytorch/blob/master/torch/utils/data/datapipes/iter/combining.py#L361-L362).

@ejguan @NivekT @VitalyFedyunin what is the correct way here?

ejguan:
IMHO, it depends on your Dataset:

1. If you want to pre-load labels or all meta data before iteration, I think it might be better to use `filter` for `label_dp` and `image_dp` and deplete `label_dp` before iteration starts:
```py
label_dp = resource_dp.filter(is_meta_file)
labels = list(label_dp)

image_dp = resource_dp.filter(is_image_file)
```
2. If meta data or labels can be loaded lazily, `IterToMap` might be a better solution here. The meta data is going to be loaded lazily. And, `zip` between `image_dp` and `label_dp`.
    - `to_map_datapipe`: https://github.com/pytorch/data/blob/13b574c80e8732744fee6ab9cb7e35b5afc34a3c/torchdata/datapipes/iter/util/converter.py#L21
    - `zip_with_map`: https://github.com/pytorch/data/blob/13b574c80e8732744fee6ab9cb7e35b5afc34a3c/torchdata/datapipes/iter/util/combining.py#L148-L149
```py
label_dp, image_dp = resource_dp.demux(classify_fn, 2)

label_dict = label_dp. to_map_datapipe()
image_dp.zip_with_map(label_dict)
...
```

NivekT:
> After some offline discussion with @NicolasHug, it became clear that the situation is not as straight forward as I thought it was. For example, the child datapipes from the `Demultiplexer` seem to be [aware if an iteration is ongoing](https://github.com/pytorch/pytorch/blob/master/torch/utils/data/datapipes/iter/combining.py#L361-L362).> 

That only matters if you have read one of your child DataPipes eagerly during the construction (it seems that is no longer the case after your code change in this PR since it no longer calls `demux` in cases where labels are eagerly needed).

NivekT (commented):
I do agree with @ejguan's points. Is it possible to use `MapDataPipe` in place of the `dict`?

pmeier:
So for my own understanding: the result of `IterToMapConverter` is something similar to what I proposed in #5219, right? Basically we get a dictionary that will only be filled at runtime, correct?

@ejguan's proposal was to use the map together with a `MapKeyZipper`. While playing with that I found that it is quite inconvenient since we cannot use a `lambda` or local function for the `merge_fn`. I preferred to use the map directly with

```py
dp = ...
map = IterToMapConverter(dp)

other_dp = ...
other_dp = Mapper(other_dp, map.__getitem__, input_col=...)
```

Do you see any downside with that?

Regardless of the approach chosen, I found that using `IterToMapConverter` introduces a performance degradation. Take the following snippet:

```py
from time import perf_counter

import torch

for name, config in [
    ("imagenet", dict(split="val")),
    ("cub200", dict(year="2011")),
]:
    warmup_times = []
    for _ in range(5):
        tic = perf_counter()
        for sample in datasets.load(name, **config):
            break
        tac = perf_counter()

        warmup_times.append(tac - tic)

    print(f"Warmup for {name} took on average {float(torch.tensor(warmup_times).mean()):.2f} seconds")
```

Running this on `main` prints

```
Warmup for imagenet took on average 2.53 seconds
Warmup for cub200 took on average 2.97 seconds
```

Running it on this PR in the current state prints

```
Warmup for imagenet took on average 2.77 seconds
Warmup for cub200 took on average 3.64 seconds
```

Any idea what could cause this? I'm aware that pytorch/data#454 proposes a speed-up, but the implementation on `main` currently also loads the full dictionary at once. Thus, this is not the cause here.

torchvision/prototype/datasets/_builtin/cub200.py
```diff
@@ -185,17 +189,16 @@ def _datapipe(self, resource_dps: List[IterDataPipe]) -> IterDataPipe[Dict[str,
...
 
             split_dp = CSVParser(split_dp, dialect="cub200")
             split_dp = Filter(split_dp, self._2011_filter_split)
             split_dp = Mapper(split_dp, getitem(0))
-            split_dp = Mapper(split_dp, image_files_map.get)
+            split_dp = Mapper(split_dp, image_files_map.__getitem__)
 
             bounding_boxes_dp = CSVParser(bounding_boxes_dp, dialect="cub200")
-            bounding_boxes_dp = Mapper(bounding_boxes_dp, image_files_map.get, input_col=0)
+            bounding_boxes_dp = Mapper(bounding_boxes_dp, image_files_map.__getitem__, input_col=0)
```
ejguan, line 201: This is a good trick for `MapDataPipe` to have the similar behavior as `zip_with_map`
cc: @NivekT

NicolasHug:
@pmeier you might be interested in https://github.com/pytorch/vision/pull/6128 which IIRC attemps to fix this issue as well

pmeier (commented):
This PR indeed solves the issue reported in #6515. The CI https://github.com/pytorch/vision/runs/8089441510 ran [before the fresh nightly](https://github.com/pytorch/vision/runs/8089441510?check_suite_focus=true#step:5:30) hit. There are still errors, but they are unrelated and reported in https://github.com/pytorch/pytorch/pull/80267#discussion_r958423241. 

Maybe we can have another go at this PR given that it has a much narrower scope than #6128? cc @NivekT @ejguan

pmeier:
> Is the issue related to `ResourceWarning: unclosed` (which only happens for [Python 3.9 Windows](https://github.com/pytorch/data/runs/8020678323?check_suite_focus=true)) caused by what you described above, or is that separate? I think #6128 is trying to fix that.

Yes, this is related. Although #6128 patches more things, I think this PR is sufficient to get rid of the warnings, given that they were always related to items left in a `Demultiplexer`.

torchvision/prototype/datasets/_builtin/imagenet.py
```diff
@@ -172,12 +179,15 @@ def _datapipe(self, resource_dps: List[IterDataPipe]) -> IterDataPipe[Dict[str,
...
+            meta_dp = CategoryAndWordNetIDExtractor(meta_dp)
+            wnid_dp = Mapper(meta_dp, getitem(1))
+            wnid_dp = Enumerator(wnid_dp, 1)
+            wnid_dp = Mapper(wnid_dp, str, input_col=0)
+            wnid_map = IterToMapConverter(wnid_dp)
 
             label_dp = LineReader(label_dp, decode=True, return_path=False)
-            # We cannot use self._wnids here, since we use a different order than the dataset
-            label_dp = Mapper(label_dp, functools.partial(self._imagenet_label_to_wnid, wnids=wnids))
+            label_dp = Mapper(label_dp, wnid_map.__getitem__)
```
ejguan, line 190: Nit comment: Performance wise, it might be slightly better to use integer as the key for `wnid_map`.
```
            wnid_dp = Enumerator(wnid_dp, 1)
            wnid_map = IterToMapConverter(wnid_dp)

            label_dp = LineReader(label_dp, decode=True, return_path=False)
            label_dp = Mapper(label_dp, int)
            label_dp = Mapper(label_dp, wnid_map.__getitem__)
```

torchvision/prototype/datasets/_builtin/imagenet.py
```diff
@@ -110,22 +133,6 @@ def _classifiy_devkit(self, data: Tuple[str, BinaryIO]) -> Optional[int]:
...
-    def _extract_categories_and_wnids(self, data: Tuple[str, BinaryIO]) -> List[Tuple[str, str]]:
-        synsets = read_mat(data[1], squeeze_me=True)["synsets"]
-        return [
-            (self._WNID_MAP.get(wnid, category.split(",", 1)[0]), wnid)
-            for _, wnid, category, _, num_children, *_ in synsets
-            # if num_children > 0, we are looking at a superclass that has no direct instance
-            if num_children == 0
-        ]
-
     def _imagenet_label_to_wnid(self, imagenet_label: str, *, wnids: Tuple[str, ...]) -> str:
```
ejguan, line 136: It seems this function can be removed as well.

ejguan (approved):
Overall LGTM!

torchvision/prototype/datasets/_builtin/imagenet.py
pmeier: Good catch!

PR #6077, opened 2022-05-24
Validate against expected files on videos

datumbox:
This PR adds similar tests to the video classification models as the ones we have for images.


jdsgomes (approved):
LGTM! thanks for the improvement. When this is green fell free to merge.

NicolasHug (approved):
LGTM. Out of curiosity do we know how long these new tests take?

datumbox:
@NicolasHug it should not increase the execution time. For some reason, we used to execute all the expensive parts of the tests such as producing predictions and jit-scripting but we didn't compare the output against expected files. Not sure what was the decision behind this but as I'm adding a new model I want to make sure I wont accidentally break the implementation.

Issue #6074, opened 2022-05-23
ResNet with InstanceNorm and zero_init_residual


FloCF:
### 🐛 Describe the bug

Since `nn.InstanceNorm2d` has `affine=False` as default, `ResNet` can not be used with `zero_init_residual=True` and `norm_layer = nn.InstanceNorm2d`.

```python
import torch.nn as nn

from torchvision.models import ResNet
from torchvision.models.resnet import BasicBlock


ResNet(BasicBlock, [2,2,2,2], norm_layer=nn.InstanceNorm2d, zero_init_residual=True)
```
Two possible solutions:
1. Throw out an error, as the combination of `norm_layer = nn.InstanceNorm2d` and `zero_init_residual=True` simply should not be used (not sure if that is the case).
2. Call `norm_layer` with `affine=True` (and then probably also `track_running_stats=True`) in the `BasicBlock` and `Bottleneck`.

### Versions

OS: Ubuntu 20.04.4 LTS (x86_64)
Python version: 3.9.5
torch version: 1.10.2
torchvision version: 0.11.3

cc @datumbox

YosuaMichael:
Thanks for the report @FloCF !
I have confirmed the problem, and I notice the problem is during weight initialization of the norm_layer: https://github.com/pytorch/vision/blob/main/torchvision/models/resnet.py#L223
```
nn.init.constant_(m.bn2.weight, 0)
```
In this case, if `m.bn2` is `nn.InstanceNorm2d` then it will not have any weight, or in other word `m.bn2.weight` is None and hence it cause error.

I tried to create a bugfix here: https://github.com/pytorch/vision/pull/6082 but I may not have a full context, so lets wait until it is reviewed.

PR #6082, opened 2022-05-24
[bugfix] Check if norm_layer has not-None weight before init

YosuaMichael:
I tried to resolve bug reported on https://github.com/pytorch/vision/issues/6074.

@datumbox I may not have a full context here, let me know what you think.

### DIFF
--- a/torchvision/models/resnet.py
+++ b/torchvision/models/resnet.py
@@ -217,9 +217,9 @@ class ResNet(nn.Module):
         # This improves the model by 0.2~0.3% according to https://arxiv.org/abs/1706.02677
         if zero_init_residual:
             for m in self.modules():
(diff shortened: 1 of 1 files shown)


FloCF:
@YosuaMichael : Your bugfix looks reasonable to me. Small remark: Maybe for the user it would be good to also throw out a warning (if `weight is None`) so that it is clear that zero_init_residual will be skipped since `norm_layer` has no weights.

datumbox (approved):
LGTM, thanks @YosuaMichael.

@FloCF Thanks for reporting the bug and for the recommendations. The implementation is generally expecting to receive a `norm_layer` compatible with the original architecture (BN, SyncBN, FrozenBN etc). I'm fine with fixing the breakage though.

PR #6088, opened 2022-05-25
[WIP] Refactor swin transfomer so later we can reuse component for 3d version

YosuaMichael:
Refactoring swin_transformer so we can reuse some components for video swin transformer.
Discussion are in: https://github.com/facebookresearch/multimodal/pull/43#discussion_r881387972

### DIFF
--- a/torchvision/models/swin_transformer.py
+++ b/torchvision/models/swin_transformer.py
@@ -59,9 +59,9 @@ def shifted_window_attention(
     qkv_weight: Tensor,
     proj_weight: Tensor,
     relative_position_bias: Tensor,
-    window_size: int,
+    window_size: List[int],
     num_heads: int,
-    shift_size: int = 0,
+    shift_size: List[int],
     attention_dropout: float = 0.0,
     dropout: float = 0.0,
     qkv_bias: Optional[Tensor] = None,
@@ -75,9 +75,9 @@ def shifted_window_attention(
         qkv_weight (Tensor[in_dim, out_dim]): The weight tensor of query, key, value.
         proj_weight (Tensor[out_dim, out_dim]): The weight tensor of projection.
         relative_position_bias (Tensor): The learned relative position bias added to attention.
-        window_size (int): Window size.
+        window_size (List[int]): Window size.
         num_heads (int): Number of attention heads.
-        shift_size (int): Shift size for shifted window attention. Default: 0.
+        shift_size (List[int]): Shift size for shifted window attention.
         attention_dropout (float): Dropout ratio of attention weight. Default: 0.0.
         dropout (float): Dropout ratio of output. Default: 0.0.
         qkv_bias (Tensor[out_dim], optional): The bias tensor of query, key, value. Default: None.
@@ -87,23 +87,25 @@ def shifted_window_attention(
     """
     B, H, W, C = input.shape
     # pad feature maps to multiples of window size
-    pad_r = (window_size - W % window_size) % window_size
-    pad_b = (window_size - H % window_size) % window_size
+    pad_r = (window_size[1] - W % window_size[1]) % window_size[1]
+    pad_b = (window_size[0] - H % window_size[0]) % window_size[0]
     x = F.pad(input, (0, 0, 0, pad_r, 0, pad_b))
     _, pad_H, pad_W, _ = x.shape
 
     # If window size is larger than feature size, there is no need to shift window.
-    if window_size == min(pad_H, pad_W):
-        shift_size = 0
+    if window_size[0] >= pad_H:
+        shift_size[0] = 0
+    if window_size[1] >= pad_W:
+        shift_size[1] = 0
 
     # cyclic shift
-    if shift_size > 0:
-        x = torch.roll(x, shifts=(-shift_size, -shift_size), dims=(1, 2))
+    if sum(shift_size) > 0:
+        x = torch.roll(x, shifts=(-shift_size[0], -shift_size[1]), dims=(1, 2))
 
     # partition windows
-    num_windows = (pad_H // window_size) * (pad_W // window_size)
-    x = x.view(B, pad_H // window_size, window_size, pad_W // window_size, window_size, C)
-    x = x.permute(0, 1, 3, 2, 4, 5).reshape(B * num_windows, window_size * window_size, C)  # B*nW, Ws*Ws, C
+    num_windows = (pad_H // window_size[0]) * (pad_W // window_size[1])
+    x = x.view(B, pad_H // window_size[0], window_size[0], pad_W // window_size[1], window_size[1], C)
+    x = x.permute(0, 1, 3, 2, 4, 5).reshape(B * num_windows, window_size[0] * window_size[1], C)  # B*nW, Ws*Ws, C
 
     # multi-head attention
     qkv = F.linear(x, qkv_weight, qkv_bias)
@@ -114,17 +116,17 @@ def shifted_window_attention(
     # add relative position bias
     attn = attn + relative_position_bias
 
-    if shift_size > 0:
+    if sum(shift_size) > 0:
         # generate attention mask
         attn_mask = x.new_zeros((pad_H, pad_W))
-        slices = ((0, -window_size), (-window_size, -shift_size), (-shift_size, None))
+        slices = [((0, -window_size[i]), (-window_size[i], -shift_size[i]), (-shift_size[i], None)) for i in range(2)]
         count = 0
-        for h in slices:
-            for w in slices:
+        for h in slices[0]:
+            for w in slices[1]:
                 attn_mask[h[0] : h[1], w[0] : w[1]] = count
                 count += 1
-        attn_mask = attn_mask.view(pad_H // window_size, window_size, pad_W // window_size, window_size)
-        attn_mask = attn_mask.permute(0, 2, 1, 3).reshape(num_windows, window_size * window_size)
+        attn_mask = attn_mask.view(pad_H // window_size[0], window_size[0], pad_W // window_size[1], window_size[1])
+        attn_mask = attn_mask.permute(0, 2, 1, 3).reshape(num_windows, window_size[0] * window_size[1])
         attn_mask = attn_mask.unsqueeze(1) - attn_mask.unsqueeze(2)
         attn_mask = attn_mask.masked_fill(attn_mask != 0, float(-100.0)).masked_fill(attn_mask == 0, float(0.0))
         attn = attn.view(x.size(0) // num_windows, num_windows, num_heads, x.size(1), x.size(1))
@@ -139,12 +141,12 @@ def shifted_window_attention(
     x = F.dropout(x, p=dropout)
 
     # reverse windows
-    x = x.view(B, pad_H // window_size, pad_W // window_size, window_size, window_size, C)
+    x = x.view(B, pad_H // window_size[0], pad_W // window_size[1], window_size[0], window_size[1], C)
     x = x.permute(0, 1, 3, 2, 4, 5).reshape(B, pad_H, pad_W, C)
 
     # reverse cyclic shift
-    if shift_size > 0:
-        x = torch.roll(x, shifts=(shift_size, shift_size), dims=(1, 2))
+    if sum(shift_size) > 0:
+        x = torch.roll(x, shifts=(shift_size[0], shift_size[1]), dims=(1, 2))
 
     # unpad features
     x = x[:, :H, :W, :].contiguous()
@@ -162,8 +164,8 @@ class ShiftedWindowAttention(nn.Module):
     def __init__(
         self,
         dim: int,
-        window_size: int,
-        shift_size: int,
+        window_size: List[int],
+        shift_size: List[int],
         num_heads: int,
         qkv_bias: bool = True,
         proj_bias: bool = True,
@@ -171,6 +173,8 @@ class ShiftedWindowAttention(nn.Module):
         dropout: float = 0.0,
     ):
         super().__init__()
+        # Assert for 2d shift window attention
+        assert len(window_size) == 2 and len(shift_size) == 2, "window_size and shift_size must be of length 2"
         self.window_size = window_size
         self.shift_size = shift_size
         self.num_heads = num_heads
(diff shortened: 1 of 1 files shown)


YosuaMichael:
I notice the failed test on jit / fx, will investigate on it

datumbox:
@xiaohu2015 I wonder if you have the bandwidth to check the proposed updates in this PR.

To provide some context, Yosua is making some small adjustments to the original implementation so that we can support easier the 3D version of the Swin model. Please let us know if you have any thoughts.

torchvision/models/swin_transformer.py
```diff
@@ -182,19 +189,19 @@ def __init__(
...
         coords = torch.stack(torch.meshgrid(coords_h, coords_w, indexing="ij"))  # 2, Wh, Ww
         coords_flatten = torch.flatten(coords, 1)  # 2, Wh*Ww
         relative_coords = coords_flatten[:, :, None] - coords_flatten[:, None, :]  # 2, Wh*Ww, Wh*Ww
         relative_coords = relative_coords.permute(1, 2, 0).contiguous()  # Wh*Ww, Wh*Ww, 2
-        relative_coords[:, :, 0] += self.window_size - 1  # shift to start from 0
-        relative_coords[:, :, 1] += self.window_size - 1
-        relative_coords[:, :, 0] *= 2 * self.window_size - 1
+        relative_coords[:, :, 0] += self.window_size[0] - 1  # shift to start from 0
+        relative_coords[:, :, 1] += self.window_size[1] - 1
+        relative_coords[:, :, 0] *= 2 * self.window_size[1] - 1
```
datumbox, line 204: Are you sure this is correct?

torchvision/models/swin_transformer.py
```diff
@@ -276,39 +303,89 @@ def forward(self, x: Tensor):
...
         depths: List[int],
         num_heads: List[int],
-        window_size: int = 7,
+        window_size: List[int],
         mlp_ratio: float = 4.0,
         dropout: float = 0.0,
         attention_dropout: float = 0.0,
         stochastic_depth_prob: float = 0.0,
-        num_classes: int = 1000,
+        num_classes: Optional[int] = 1000,
```
YosuaMichael, line 385: We make `num_classes` as optional, and it will not have any head if we set it None during instantiation.
This would be useful for use-cases that don't need head or need to switch head externally

torchvision/models/swin_transformer.py
```diff
@@ -228,31 +253,33 @@ class SwinTransformerBlock(nn.Module):
...
-        window_size (int): Window size. Default: 7.
-        shift_size (int): Shift size for shifted window attention. Default: 0.
+        window_size (List[int]): Window size.
+        shift_size (List[int]): Shift size for shifted window attention.
         mlp_ratio (float): Ratio of mlp hidden dim to embedding dim. Default: 4.0.
         dropout (float): Dropout rate. Default: 0.0.
         attention_dropout (float): Attention dropout rate. Default: 0.0.
         stochastic_depth_prob: (float): Stochastic depth rate. Default: 0.0.
         norm_layer (nn.Module): Normalization layer.  Default: nn.LayerNorm.
+        attn_layer (nn.Module): Attention layer. Default: ShiftedWindowAttention
```
YosuaMichael, line 263: We put `attn_layer` as constructor param so the SwinTransformerBlock can accomodate both 2d and 3d `ShiftedWindowAttention`

torchvision/models/swin_transformer.py
```diff
@@ -39,18 +39,23 @@ def __init__(self, dim: int, norm_layer: Callable[..., nn.Module] = nn.LayerNorm
...
 
     def forward(self, x: Tensor):
-        B, H, W, C = x.shape
-
-        x0 = x[:, 0::2, 0::2, :]  # B H/2 W/2 C
-        x1 = x[:, 1::2, 0::2, :]  # B H/2 W/2 C
-        x2 = x[:, 0::2, 1::2, :]  # B H/2 W/2 C
-        x3 = x[:, 1::2, 1::2, :]  # B H/2 W/2 C
-        x = torch.cat([x0, x1, x2, x3], -1)  # B H/2 W/2 4*C
-        x = x.view(B, -1, 4 * C)  # B H/2*W/2 4*C
```
YosuaMichael, line 49: I have confirmed that:
```
x = x.view(B, -1, 4 * C)
x = self.norm(x)
x = self.reduction(x)
x = x.view(B, H // 2, W // 2, 2 * C)
```
is equivalent with
```
 x = self.norm(x)
x = self.reduction(x)
```
using `nn.LayerNorm` as the `norm`

torchvision/models/swin_transformer.py
```diff
@@ -39,18 +39,23 @@ def __init__(self, dim: int, norm_layer: Callable[..., nn.Module] = nn.LayerNorm
...
+        """
+        Args:
+            x (Tensor): input tensor with expected layout of [..., H, W, C]
+        Returns:
+            Tensor with layout of [..., H/2, W/2, 2*C]
+        """
+        H, W, _ = x.shape[-3:]
+        x = F.pad(x, (0, 0, 0, W % 2, 0, H % 2))
+
+        x0 = x[..., 0::2, 0::2, :]  # ... H/2 W/2 C
```
YosuaMichael, line 51: by using `...` we can accept both 2d and 3d image as long as the layout ends with `H W C`.
In 2d case it would be: `B H W C` and in 3d case it would be `B T H W C`

xiaohu2015:
@YosuaMichael All your modifcation looks good for me. But I have one question about the `Update the method to handle edge cases where window_size is larger than input`
In the original code, we will pad the input and set shift_size=0 but not change the window size. But you choose to change the window size but not pad the input.

torchvision/models/swin_transformer.py
xiaohu2015: > @xiaohu2015 I wonder if you have the bandwidth to check the proposed updates in this PR.
> 
> To provide some context, Yosua is making some small adjustments to the original implementation so that we can support easier the 3D version of the Swin model. Please let us know if you have any thoughts.

I think the adjustments don't affect current model, but can @jdsgomes double check the model accuary?

YosuaMichael:
> @YosuaMichael All your modifcation looks good for me. But I have one question about the Update the method to handle edge cases where window_size is larger than input
In the original code, we will pad the input and set shift_size=0 but not change the window size. But you choose to change the window size but not pad the input.

Hi @xiaohu2015 , thanks for taking a look :)
I think the original code also modify the `window_size` when it is larger than `input_size`. Here is the part I refer to: https://github.com/microsoft/Swin-Transformer/blob/main/models/swin_transformer.py#L192-L195

xiaohu2015:
> > @YosuaMichael All your modifcation looks good for me. But I have one question about the Update the method to handle edge cases where window_size is larger than input
> > In the original code, we will pad the input and set shift_size=0 but not change the window size. But you choose to change the window size but not pad the input.
> 
> Hi @xiaohu2015 , thanks for taking a look :) I think the original code also modify the `window_size` when it is larger than `input_size`. Here is the part I refer to: https://github.com/microsoft/Swin-Transformer/blob/main/models/swin_transformer.py#L192-L195

You are right, I also think the second way is more reasonable.

YosuaMichael:
> @YosuaMichael I think you maybe want to implement the https://arxiv.org/pdf/2106.13230.pdf? It should use 3D windows? maybe I missed some thing?

Yes, you are right! I want to implement `Video Swin Transformer` , currently I added it on `torchmultimodal` repository first (see this [PR](https://github.com/facebookresearch/multimodal/pull/43)) but I plan to upstream it into `torchvision` as `video_classifier` after that.

torchvision/models/swin_transformer.py
datumbox: I don't think the variation is due to this. This is more likely due to the non-determinism. If you run it a few times, you will get different results.

torchvision/models/swin_transformer.py
```diff
@@ -182,39 +200,48 @@ def __init__(
...
+            x (Tensor): Tensor with layout of [B, H, W, C]
+        Returns:
+            Tensor with same layout as input, i.e. [B, H, W, C]
+        """
+        _, H, W, _ = x.shape
+        input_hw = [H, W]
+        window_size, shift_size = _fix_window_and_shift_size(input_hw, self.window_size, self.shift_size)
+
+        N = window_size[0] * window_size[1]
+        relative_position_bias = self.relative_position_bias_table[self.relative_position_index[:N, :N]]  # type: ignore[index]
```
datumbox, line 233: @YosuaMichael Why `self.relative_position_index[:N, :N]` instead of the original?

@xiaohu2015 thoughts on this?

torchvision/models/swin_transformer.py
```diff
@@ -362,7 +436,11 @@ def __init__(
         num_features = embed_dim * 2 ** (len(depths) - 1)
         self.norm = norm_layer(num_features)
         self.avgpool = nn.AdaptiveAvgPool2d(1)
-        self.head = nn.Linear(num_features, num_classes)
+        self.head: Optional[nn.Module] = None
+        if num_classes is not None:
+            self.head = nn.Linear(num_features, num_classes)
+        else:
+            self.head = None
```
datumbox, line 443: I'm not sure about this change. It's not an idiom we support at TorchVision, I would revert this one.

torchvision/models/swin_transformer.py
datumbox: I don't think this API proposal should be part of this PR. It introduces something not supported by any other model. I would recommend examining this separately across the library.

torchvision/models/swin_transformer.py
```diff
@@ -276,39 +305,89 @@ def forward(self, x: Tensor):
...
+        window_size: List[int],
         mlp_ratio: float = 4.0,
         dropout: float = 0.0,
         attention_dropout: float = 0.0,
         stochastic_depth_prob: float = 0.0,
-        num_classes: int = 1000,
+        num_classes: Optional[int] = 1000,
         norm_layer: Optional[Callable[..., nn.Module]] = None,
         block: Optional[Callable[..., nn.Module]] = None,
+        patch_embed: Optional[Callable[..., nn.Module]] = None,
```
datumbox, line 390: Do we really need this change? Do you need it for the 3d case?

torchvision/models/swin_transformer.py
```diff
@@ -428,7 +506,7 @@ class Swin_T_Weights(WeightsEnum):
             "recipe": "https://github.com/pytorch/vision/tree/main/references/classification#swintransformer",
             "_metrics": {
                 "ImageNet-1K": {
-                    "acc@1": 81.474,
+                    "acc@1": 81.470,
```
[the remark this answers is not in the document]
datumbox: I don't think you need to change this. Every time you run the analysis you might get a slightly different number on the 3rd decimal and there is not much you can do about it.

torchvision/models/swin_transformer.py
```diff
@@ -452,7 +532,7 @@ class Swin_S_Weights(WeightsEnum):
             "_metrics": {
                 "ImageNet-1K": {
                     "acc@1": 83.196,
-                    "acc@5": 96.360,
+                    "acc@5": 96.362,
```
datumbox, line 535: Same here. I would revert this.

datumbox (commented):
Thanks @YosuaMichael, just a few comments below

torchvision/models/swin_transformer.py
datumbox: The tests are [failing](https://app.circleci.com/pipelines/github/pytorch/vision/17913/workflows/d1d170ca-5003-4cb3-b45d-63c5ca5893fe/jobs/1450534) due to this. I recommend reverting this change

YosuaMichael: Yeah, currently investigating this. The error message indicate about the type (not integer), so my thinking now is to make the type explicit. If this doesn't work I will revert this change

torchvision/models/swin_transformer.py
YosuaMichael: Ok, agree that this is not part of this PR

torchvision/models/swin_transformer.py
YosuaMichael: Yes, I particularly need it for `Omnivore` case since they have a custom patch embedding layer.

torchvision/models/swin_transformer.py
YosuaMichael: Ok, noted

torchvision/models/swin_transformer.py
datumbox: But do you intent to reuse the `SwinTransformer` there? My understanding is that you wouldn't. Instead you would be introducing a `SwinTransformer3d` which would overwrite things and can change the patch definitions.

YosuaMichael: Yeah thats correct, I would introduce `SwinTransformer3d`

torchvision/models/swin_transformer.py
```diff
@@ -324,7 +342,9 @@ def __init__(
         # split image into non-overlapping patches
         layers.append(
             nn.Sequential(
-                nn.Conv2d(3, embed_dim, kernel_size=patch_size, stride=patch_size),
+                nn.Conv2d(
+                    3, embed_dim, kernel_size=(patch_size[0], patch_size[1]), stride=(patch_size[0], patch_size[1])
```
datumbox, line 346: Nit:
```suggestion
                    3, embed_dim, kernel_size=tuple(patch_size), stride=tuple(patch_size)
```

datumbox (approved):
LGTM, thanks @YosuaMichael.

Can you please provide proof for new inference runs to confirm the accuracy of the models remain unaffected?

We can merge on Green CI.

torchvision/models/swin_transformer.py
YosuaMichael: This is actually what I did first, but unfortunately mypy complaints :(
```
swin_transformer.py:346: error: Argument "kernel_size" to "Conv2d" has incompatible type "Tuple[int, ...]"; expected "Union[int, Tuple[int, int]]"  [arg-type]
                        3, embed_dim, kernel_size=tuple(patch_size), stride=(patch_size[0], patch_size[1])
```

YosuaMichael:
@datumbox I have rerun the validation script and it return the same exact result as when I ran before (the one in PR description).

xiaohu2015:
> > @YosuaMichael All your modifcation looks good for me. But I have one question about the Update the method to handle edge cases where window_size is larger than input
> > In the original code, we will pad the input and set shift_size=0 but not change the window size. But you choose to change the window size but not pad the input.
> 
> Hi @xiaohu2015 , thanks for taking a look :) I think the original code also modify the `window_size` when it is larger than `input_size`. Here is the part I refer to: https://github.com/microsoft/Swin-Transformer/blob/main/models/swin_transformer.py#L192-L195

why you revert this modifcation?

xiaohu2015:
> > > @YosuaMichael All your modifcation looks good for me. But I have one question about the Update the method to handle edge cases where window_size is larger than input
> > > In the original code, we will pad the input and set shift_size=0 but not change the window size. But you choose to change the window size but not pad the input.
> > 
> > 
> > Hi @xiaohu2015 , thanks for taking a look :) I think the original code also modify the `window_size` when it is larger than `input_size`. Here is the part I refer to: https://github.com/microsoft/Swin-Transformer/blob/main/models/swin_transformer.py#L192-L195
> 
> why you revert this modifcation?

It maybe important for swin_v2 model, since they use 192x192 input, the window_size in the last stage will be 6 rather than 7

datumbox:
@YosuaMichael could you pleas provide the details?

YosuaMichael:
@xiaohu2015 Sorry for the late reply and thanks for the feedback.

The reason why we revert the changes is because we want to cherrypick this changes in the new release of torchvision 0.13.0, at that time we decide to not change any previous behaviour but only do more "safe" changes that aim for reusing components to swin_transformer 3d.

I think we can put it back now with another PR, I will create a PR for this and tag you!

YosuaMichael:
@xiaohu2015 I create the changes on this PR: #6222

PR #6090, opened 2022-05-25
Fix test_backbone_utils by adding .float() before .mean()

YosuaMichael:
In [test_backbone_utils.TestFxFeatureExtraction](https://github.com/pytorch/vision/blob/main/test/test_backbone_utils.py#L101), there is edge case when the [sampled nodes](https://github.com/pytorch/vision/blob/main/test/test_backbone_utils.py#L136) are buffer and does not depend on input and has integer value ([example on swin transformer](https://github.com/pytorch/vision/blob/main/torchvision/models/swin_transformer.py#L199)).

In this case, the test will trigger error because `Tensor.mean` does not accept `integer`.
This PR provide a fix by adding `.float()` before `.mean()`.

### DIFF
--- a/test/test_backbone_utils.py
+++ b/test/test_backbone_utils.py
@@ -183,12 +183,12 @@ class TestFxFeatureExtraction:
         out_agg = 0
         for node_out in out.values():
             if isinstance(node_out, Sequence):
-                out_agg += sum(o.mean() for o in node_out if o is not None)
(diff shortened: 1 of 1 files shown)


datumbox (approved):
LGTM, thanks @YosuaMichael. Feel free to merge once the CI is green.

PR #6091, opened 2022-05-25
Fix edge cases on test_backbone_utils when the sampled node is buffer with integer value

YosuaMichael:
Cherry pick https://github.com/pytorch/vision/pull/6090 from main into `release/0.13`

### DIFF
--- a/test/test_backbone_utils.py
+++ b/test/test_backbone_utils.py
@@ -183,12 +183,12 @@ class TestFxFeatureExtraction:
         out_agg = 0
         for node_out in out.values():
             if isinstance(node_out, Sequence):
-                out_agg += sum(o.mean() for o in node_out if o is not None)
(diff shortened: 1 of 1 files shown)


YosuaMichael:
The type check failed, but seems like it is not related with this PR. It affect all the model loading. Will investigate...

datumbox (approved):
LGTM. Looks like the breakage is completely unrelated. Possibly due to a change on Core??? I recommend prior merging this PR to open a dummy PR against the release branch to confirm.

YosuaMichael:
Seems like the problem in typecheck is not because of this PR: https://github.com/pytorch/vision/pull/6092
I have created this dummy PR to check.

I am not too sure what cause this problem, but I have a fix for the type check, will create separate PR for it

datumbox:
The problem is not on your side. We should merge this and review the issue on a separate PR.

PR #6092, opened 2022-05-25
[DONT MERGE] Dummy PR to check Release/0.13 CI is green

YosuaMichael:
This PR is just to check whether there the CI is green or not for branch `release/0.13`

### DIFF
--- a/dummy.txt
+++ b/dummy.txt
@@ -1,0 +1,1 @@
+just to check CI
--- a/dummy.txt
+++ b/dummy.txt
@@ -1,1 +1,2 @@
 just to check CI
+0


pmeier:
We already have a PR for that: #5009. Closing this to avoid duplicates.

pmeier:
Re-opening, because I missed that it uses the release branch as base.

datumbox:
@pmeier I think we can be a bit more lenient to allow people create dummy PRs to check the CI, as long as they close it in due time.

pmeier:
I agree with the sentiment. If this PR was using `main` as base, I felt it was justified to close and simultaneously to inform that we already have something like that (no changes and just checking CI). Basically the same what I did on your PR in https://github.com/pytorch/vision/pull/6017#issuecomment-1127284180.

Of course I blundered and base is not `main` making the other PR irrelevant. The same were true if a test PR introduces any changes.

datumbox:
@pmeier I like the proactiveness to keep things clean. We should remain proactive and perhaps eliminate some of the hundreds of stale PRs from years ago.

But for the CI case, I don't think it adds any value closing the PRs like that. While on-call, we often need to indicate to the Release Engineering team that the main branch is broken and provide proof. Redirecting them to a PR with 10+ commits adds unnecessary complexity as they need to confirm that this is indeed a dummy commit on top of main. For all these reasons, I suggest to leave people close their own dummy PRs, provided they haven't left them there forever.

PR #6095, opened 2022-05-26
prevent feature wrapping if the feature is not the primary operand

### LINKED ISSUE #6094: `Tensor.to` does not work with feature as reference
`torch.Tensor.to` accepts a tensor reference object as input in which case it converts to the reference dtype and device:

```py
>>> a = torch.zeros((), dtype=torch.float)
>>> b = torch.ones((), dtype=torch.int)
>>> b
tensor(1, dtype=torch.int32)
>>> b.to(a)
tensor(1.)
```

This also works for our custom features:

```py
>>> data = torch.rand(3, 2, 2)
>>> image = features.Image(data.to(torch.float64))
>>> image
Image([[[0.9595, 0.0947],
        [0.9553, 0.5563]],
       [[0.5435, 0.2975],
        [0.3037, 0.0863]],
       [[0.6253, 0.3481],
        [0.3518, 0.4499]]], dtype=torch.float64)
>>> image.to(torch.float32)
Image([[[0.9595, 0.0947],
        [0.9553, 0.5563]],
       [[0.5435, 0.2975],
        [0.3037, 0.0863]],
       [[0.6253, 0.3481],
        [0.3518, 0.4499]]])
>>> image.to(data)
Image([[[0.9595, 0.0947],
        [0.9553, 0.5563]],
       [[0.5435, 0.2975],
        [0.3037, 0.0863]],
       [[0.6253, 0.3481],
        [0.3518, 0.4499]]])

```

However, it doesn't work if we want to convert a plain tensor and use a custom feature as reference:

```py
>>> data.to(image)
AttributeError: 'Tensor' object has no attribute 'color_space'
```

cc @bjuncek @pmeier

pmeier:
Fixes #6094. 

Apart from the fix, this PR also adds a few minimal tests that I copied from the design doc, to ensure I'm not breaking anything else.


test/test_prototype_features.py
```diff
@@ -0,0 +1,72 @@
...
+    label = features.Label(tensor, categories=["foo", "bar"])
+
+    label_to = label.to(torch.int32)
+
+    assert type(label_to) is features.Label
+    assert label_to.dtype is torch.int32
+    assert label_to.categories is label.categories
+
+
+def test_to_feature_reference():
```
pmeier, line 30: This is the test case that makes sure we actually fix #6094.

torchvision/prototype/features/_feature.py
```diff
@@ -77,6 +77,13 @@ def __torch_function__(
         with DisableTorchFunction():
             output = func(*args, **kwargs)
 
+        # The __torch_function__ protocol will invoke this method on all types involved in the computation by walking
+        # the MRO upwards. For example, `torch.Tensor(...).to(features.Image(...))` will invoke
+        # `features.Image.__torch_function__` first. The check below makes sure that we do not try to wrap in such a
+        # case.
+        if not isinstance(args[0], cls):
+            return output
```
pmeier, line 85: Although the behavior should be kept for all ops where we might wrap, right now this only applies to `torch.Tensor.to`. `torch.Tensor.clone` does not accept any inputs and thus this method will only be invoked in case the data to clone is actually a feature. All other ops will be handled by the `else` branch which performs no wrapping.

Thus, we could also merge this into the `torch.Tensor.to` branch for now, but will need to put it back where it is right now as soon as we special case another op that takes tensors as inputs. I would keep it were it is, but won't oppose putting it elsewhere for now.

test/test_prototype_features.py
```diff
@@ -0,0 +1,72 @@
...
+    assert label.data_ptr() == tensor.data_ptr()
+
+
+def test_to_wrapping():
+    tensor = torch.tensor([0, 1, 0], dtype=torch.int64)
+    label = features.Label(tensor, categories=["foo", "bar"])
+
+    label_to = label.to(torch.int32)
+
+    assert type(label_to) is features.Label
```
vfdev-5, line 25: why not using canonical `isinstance` ?

test/test_prototype_features.py
```diff
@@ -0,0 +1,72 @@
...
+    assert type(label_clone) is features.Label
+    assert label_clone.data_ptr() != label.data_ptr()
+    assert label_clone.categories is label.categories
+
+
+def test_other_op_no_wrapping():
+    tensor = torch.tensor([0, 1, 0], dtype=torch.int64)
+    label = features.Label(tensor, categories=["foo", "bar"])
+
+    # any operation besides .to() and .clone() will do here
```
vfdev-5, line 55: this comment is unclear. What any op will do here?

torchvision/prototype/features/_feature.py
```diff
@@ -77,6 +77,13 @@ def __torch_function__(
         with DisableTorchFunction():
             output = func(*args, **kwargs)
 
+        # The __torch_function__ protocol will invoke this method on all types involved in the computation by walking
+        # the MRO upwards. For example, `torch.Tensor(...).to(features.Image(...))` will invoke
+        # `features.Image.__torch_function__` first. The check below makes sure that we do not try to wrap in such a
+        # case.
+        if not isinstance(args[0], cls):
```
vfdev-5, line 84: By default args is empty tuple `()`. Shouldn't we make sure that args is not empty before calling `args[0]` ?

test/test_prototype_features.py
pmeier: Consistency with the other tests. No strong opinion though.

test/test_prototype_features.py
pmeier: Not being snarky: any operation. It doesn't matter as long as it is not `to` and `clone`. You can do `output = label + 0`, `output = label.sub(0)` or whatever.

torchvision/prototype/features/_feature.py
pmeier: Good catch. Should be:

```suggestion
        if args and not isinstance(args[0], cls):
```

pmeier: Actually, this is right as is. `args` is never empty but will always have `self` as the first element.

vfdev-5 (approved):
Let's finally merge this, thanks @pmeier !

PR #6097, opened 2022-05-26
Change weights return type to Mapping

datumbox:
As of recently, PyTorch Core's [switched](https://github.com/pytorch/pytorch/blob/master/torch/nn/modules/module.py#L1534) its types to the following signature:
```python
    def load_state_dict(self, state_dict: Mapping[str, Any],
                        strict: bool = True):
```

This PR aligns with it.


YosuaMichael (approved):
Looks good :)

PR #6101, opened 2022-05-26
Fix install command for torch cpu release 1.12

### LINKED ISSUE #6099: The Release 0.13 is not using the correct torch core version
### 🐛 Describe the bug

Currently the CI on branch `release/0.13` is failing (I create dummy PR https://github.com/pytorch/vision/pull/6092 to track this). 
We fail on type_check  and this does not happen on main branch.
After looking at the error: https://app.circleci.com/pipelines/github/pytorch/vision/17906/workflows/8f008f5a-13b2-4207-a378-51ae705b99b7/jobs/1450718
especially on the installation of pytorch, we get:
```
#!/bin/bash -eo pipefail
pip install  --user  --progress-bar=off --pre torch --extra-index-url https://download.pytorch.org/whl/test/cpu/torch_test.html

Looking in indexes: https://pypi.org/simple, https://download.pytorch.org/whl/test/cpu/torch_test.html
Collecting torch
  Downloading torch-1.11.0-cp37-cp37m-manylinux1_x86_64.whl (750.6 MB)
```

We should use the RC branch from pytorch core but instead we get version 1.11.

After digging a bit deeper, if we search "torch-" on https://download.pytorch.org/whl/test/cpu/torch_test.html currently I can only see 4 versions and the latest one is indeed 1.11. I think there is a problem here.

cc @malfet @seemethere 

### Versions

-

atalman:
Fix install command for torch cpu release 1.12
Fixes #6099

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -142,7 +142,7 @@ commands:
         default: true
     steps:
       - pip_install:
-          args: --pre torch --extra-index-url https://download.pytorch.org/whl/test/cpu/torch_test.html
+          args: --pre torch -f https://download.pytorch.org/whl/test/cpu/torch_test.html
           descr: Install PyTorch from nightly releases
       - pip_install:
           args: --no-build-isolation <<# parameters.editable >> --editable <</ parameters.editable >> .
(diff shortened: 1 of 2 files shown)


datumbox (approved):
Thanks! Feel free to merge on Green CI.

PR #6104, opened 2022-05-27
[proto] Added tests for other padding modes

vfdev-5:
Related to https://github.com/pytorch/vision/issues/5782

Description:
- Added missing tests for other padding modes: "edge", "reflect", "symmetric"

### DIFF
--- a/test/test_prototype_transforms_functional.py
+++ b/test/test_prototype_transforms_functional.py
@@ -1101,17 +1101,6 @@ def test_correctness_resized_crop_segmentation_mask(device, top, left, height, w
     torch.testing.assert_close(output_mask, expected_mask)
 
 
-@pytest.mark.parametrize("device", cpu_and_gpu())
-def test_correctness_pad_segmentation_mask_on_fixed_input(device):
-    mask = torch.ones((1, 3, 3), dtype=torch.long, device=device)
-
-    out_mask = F.pad_segmentation_mask(mask, padding=[1, 1, 1, 1])
-
-    expected_mask = torch.zeros((1, 5, 5), dtype=torch.long, device=device)
-    expected_mask[:, 1:-1, 1:-1] = 1
-    torch.testing.assert_close(out_mask, expected_mask)
-
-
 def _parse_padding(padding):
     if isinstance(padding, int):
         return [padding] * 4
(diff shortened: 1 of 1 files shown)


test/test_prototype_transforms_functional.py
```diff
@@ -1168,25 +1157,67 @@ def _compute_expected_bbox(bbox, padding_):
         torch.testing.assert_close(output_boxes, expected_bboxes)
 
 
+@pytest.mark.parametrize("device", cpu_and_gpu())
+def test_correctness_pad_segmentation_mask_on_fixed_input(device):
```
pmeier, line 1161: Does this test have any value with the other test added?

test/test_prototype_transforms_functional.py
```diff
@@ -1168,25 +1157,67 @@ def _compute_expected_bbox(bbox, padding_):
...
+            output[..., :pad_up, :pad_left] = mask[..., 0, 0].unsqueeze(-1).unsqueeze(-2)
+            output[..., pad_up:-pad_down, :pad_left] = mask[..., :, 0].unsqueeze(-1)
+            output[..., -pad_down:, :pad_left] = mask[..., -1, 0].unsqueeze(-1).unsqueeze(-2)
+
+            output[..., :pad_up, -pad_right:] = mask[..., 0, -1].unsqueeze(-1).unsqueeze(-2)
+            output[..., pad_up:-pad_down, -pad_right:] = mask[..., :, -1].unsqueeze(-1)
+            output[..., -pad_down:, -pad_right:] = mask[..., -1, -1].unsqueeze(-1).unsqueeze(-2)
+
+            output[..., :pad_up, pad_left:-pad_right] = mask[..., 0, :].unsqueeze(-2)
+            output[..., -pad_down:, pad_left:-pad_right] = mask[..., -1, :].unsqueeze(-2)
```
pmeier, line 1199: This is quite hard to parse, but I guess there is little we can do to improve. Two suggestions though:

1. If you specify a range as index, the dimensions will not be squeezed. For example: `mask[..., 0, 0].unsqueeze(-1).unsqueeze(-2) == mask[..., :1, :1]`
2. Could you add a comment above each block which side, i.e. top, bottom, left, or right, the padding applies to? Please also do so in the branch below.

test/test_prototype_transforms_functional.py
```diff
@@ -1168,25 +1157,67 @@ def _compute_expected_bbox(bbox, padding_):
...
-    def _compute_expected_mask(mask, padding_):
+@pytest.mark.parametrize("padding_mode", ["constant", "edge", "reflect", "symmetric"])
+def test_correctness_pad_segmentation_mask(padding, padding_mode):
+    def _compute_expected_mask(mask, padding_, padding_mode_):
         h, w = mask.shape[-2], mask.shape[-1]
         pad_left, pad_up, pad_right, pad_down = _parse_padding(padding_)
 
+        # expected output can be computed on positive pad values only
+        # but F.pad_* can also crop for negative values
+        assert pad_left > 0 and pad_up > 0 and pad_right > 0 and pad_down > 0
```
pmeier, line 1180: Let's make it clear that this is not an error of the op.

```suggestion
        if any(pad <= 0 for pad in [pad_left, pad_up, pad_right, pad_down])
            raise pytest.UsageError(
                "expected output can be computed on positive pad values only, "
                "but F.pad_* can also crop for negative values"
            )
```

pmeier (commented):
A few minor comments inline.

test/test_prototype_transforms_functional.py
vfdev-5: This test checks one option on cuda device instead of [cuda, cpu] x all_options from below test

test/test_prototype_transforms_functional.py
vfdev-5: Let's keep `unsqueeze`, it is much less obvious to parse with `mask[..., :1, :1]`.

datumbox (approved):
LGTM, thanks!

PR #6106, opened 2022-05-30
Fixing CI job "cmake_linux_gpu" issue with setup_env.sh in docker

vfdev-5:
New version of git checks folder owner and errors with 'fatal: unsafe repository' if owner is different.
Thus, `root_dir="$(git rev-parse --show-toplevel)"` is failing.

To fix the issue, just added all folders as safe directories


datumbox (approved):
LGTM, thanks!

PR #6107, opened 2022-05-30
[WIP] Add raft-stereo model to prototype/models

YosuaMichael:
Adding a prototype for depth stereo model, RAFT-Stereo.

### DIFF
--- a/torchvision/prototype/models/depth/stereo/raft_stereo.py
+++ b/torchvision/prototype/models/depth/stereo/raft_stereo.py
@@ -1,0 +1,692 @@
+from typing import List, Optional
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from torch import Tensor
+from torch.nn.modules.batchnorm import BatchNorm2d
+from torch.nn.modules.instancenorm import InstanceNorm2d
+from torchvision.ops import Conv2dNormActivation
+
+from torchvision.utils import _log_api_usage_once
+from torchvision.models._api import Weights, WeightsEnum
+from torchvision.models.optical_flow._utils import grid_sample, make_coords_grid, upsample_flow
+
+
+# Write helper function here temporarily
+
+# MODIFIED from torchvision.models.optical_flow._utils.upsample_flow
+# Make it more generic by adding factor as params and can handle different num_channels input
+def upsample_with_mask_and_factor(x, up_mask: Optional[Tensor] = None, factor=8):
+    """Upsample tensor x to have factor times the size
+
+    If up_mask is None we just interpolate.
+    If up_mask is specified, we upsample using a convex combination of its weights.
+    """
+    batch_size, num_channels, h, w = x.shape
+    new_h, new_w = h * factor, w * factor
+
+    if up_mask is None:
+        return factor * F.interpolate(x, size=(new_h, new_w), mode="bilinear", align_corners=True)
+
+    up_mask = up_mask.view(batch_size, 1, 9, factor, factor, h, w)
+    up_mask = torch.softmax(up_mask, dim=2)  # "convex" == weights sum to 1
+
+    upsampled_x = F.unfold(factor * x, kernel_size=3, padding=1).view(batch_size, num_channels, 9, 1, 1, h, w)
+    upsampled_x = torch.sum(up_mask * upsampled_x, dim=2)
+
+    return upsampled_x.permute(0, 1, 4, 2, 5, 3).reshape(batch_size, num_channels, new_h, new_w)
+
+
+# REUSING from torchvision.models.optical_flow.raft
+class ResidualBlock(nn.Module):
+    """Slightly modified Residual block with extra relu and biases."""
+
+    def __init__(self, in_channels, out_channels, *, norm_layer, stride=1):
+        super().__init__()
+
+        # Note regarding bias=True:
+        # Usually we can pass bias=False in conv layers followed by a norm layer.
+        # But in the RAFT training reference, the BatchNorm2d layers are only activated for the first dataset,
+        # and frozen for the rest of the training process (i.e. set as eval()). The bias term is thus still useful
+        # for the rest of the datasets. Technically, we could remove the bias for other norm layers like Instance norm
+        # because these aren't frozen, but we don't bother (also, we woudn't be able to load the original weights).
+        self.convnormrelu1 = Conv2dNormActivation(
+            in_channels, out_channels, norm_layer=norm_layer, kernel_size=3, stride=stride, bias=True
+        )
+        self.convnormrelu2 = Conv2dNormActivation(
+            out_channels, out_channels, norm_layer=norm_layer, kernel_size=3, bias=True
+        )
+
+        if stride == 1:
+            self.downsample = nn.Identity()
+        else:
+            self.downsample = Conv2dNormActivation(
+                in_channels,
+                out_channels,
+                norm_layer=norm_layer,
+                kernel_size=1,
+                stride=stride,
+                bias=True,
+                activation_layer=None,
+            )
+
+        self.relu = nn.ReLU(inplace=True)
+
+    def forward(self, x):
+        y = x
+        y = self.convnormrelu1(y)
+        y = self.convnormrelu2(y)
+
+        x = self.downsample(x)
+
+        return self.relu(x + y)
+
+
+# MODIFIED from torchvision.models.optical_flow.raft.FeatureEncoder by adding strides param
+class BaseEncoder(nn.Module):
+    """The feature encoder, used both as the actual feature encoder, and as the context encoder.
+
+    It must downsample its input by 8.
+    """
+
+    def __init__(self, *, block=ResidualBlock, layers=(64, 64, 96, 128), strides=(2, 1, 2, 2), norm_layer=nn.BatchNorm2d):
+        super().__init__()
+
+        if len(layers) != 4:
+            raise ValueError(f"The expected number of layers is 4, instead got {len(layers)}")
+
+        # See note in ResidualBlock for the reason behind bias=True
+        self.convnormrelu = Conv2dNormActivation(
+            3, layers[0], norm_layer=norm_layer, kernel_size=7, stride=strides[0], bias=True
+        )
+
+        self.layer1 = self._make_2_blocks(block, layers[0], layers[1], norm_layer=norm_layer, first_stride=strides[1])
+        self.layer2 = self._make_2_blocks(block, layers[1], layers[2], norm_layer=norm_layer, first_stride=strides[2])
+        self.layer3 = self._make_2_blocks(block, layers[2], layers[3], norm_layer=norm_layer, first_stride=strides[3])
+        self.output_dim = layers[3]
+
+        for m in self.modules():
+            if isinstance(m, nn.Conv2d):
+                nn.init.kaiming_normal_(m.weight, mode="fan_out", nonlinearity="relu")
+            elif isinstance(m, (nn.BatchNorm2d, nn.InstanceNorm2d)):
+                if m.weight is not None:
+                    nn.init.constant_(m.weight, 1)
+                if m.bias is not None:
+                    nn.init.constant_(m.bias, 0)
+
+        num_downsampling = sum([x-1 for x in strides])
+        self.downsampling_ratio = 2 ** (num_downsampling)
+
+    def _make_2_blocks(self, block, in_channels, out_channels, norm_layer, first_stride):
+        block1 = block(in_channels, out_channels, norm_layer=norm_layer, stride=first_stride)
+        block2 = block(out_channels, out_channels, norm_layer=norm_layer, stride=1)
+        return nn.Sequential(block1, block2)
+
+    def forward(self, x):
+        x = self.convnormrelu(x)
+
+        x = self.layer1(x)
+        x = self.layer2(x)
+        x = self.layer3(x)
+        return x
+
+
+class FeatureEncoder(nn.Module):
+    def __init__(self, base_encoder, output_dim=256, shared_base=False, block=ResidualBlock):
+        super().__init__()
+        self.base_encoder = base_encoder
+        self.base_downsampling_ratio = base_encoder.downsampling_ratio
+        base_dim = base_encoder.output_dim
+
+        # If we share base encoder weight for Feature and Context Encoder
+        # we need to add residual block with InstanceNorm2d
+        self.shared_base = shared_base
+        if shared_base:
+            self.residual_block = block(base_dim, base_dim, norm_layer=nn.InstanceNorm2d, stride=1)
+        self.conv = nn.Conv2d(base_dim, output_dim, kernel_size=1)
+
+    def forward(self, x):
+        x = self.base_encoder(x)
+        if self.shared_base:
+            x = self.residual_block(x)
+        x = self.conv(x)
+        return x
+
+
+class MultiLevelContextEncoder(nn.Module):
+    def __init__(self, base_encoder, output_dim=256, out_with_blocks=[1, 1, 0], block=ResidualBlock):
+        super().__init__()
+        self.num_level = len(out_with_blocks)
+        self.base_encoder = base_encoder
+        self.base_downsampling_ratio = base_encoder.downsampling_ratio
+        base_dim = base_encoder.output_dim
+        self.output_layers = nn.ModuleList([self._make_out_layer(base_dim, output_dim, with_block=with_block, block=block) for with_block in out_with_blocks])
+        self.downsamplers = nn.ModuleList([self._make_downsampler(block, base_dim, base_dim) for i in range(1, self.num_level)])
+
+    def _make_out_layer(self, in_channels, out_channels, with_block=1, block=ResidualBlock):
+        conv_layer = nn.Conv2d(in_channels, out_channels, kernel_size=3, padding=1)
+        if with_block:
+            block_layer = block(in_channels, in_channels, norm_layer=nn.BatchNorm2d, stride=1)
+            return nn.Sequential(block_layer, conv_layer)
+        else:
+            return conv_layer
+
+    def _make_downsampler(self, block, in_channels, out_channels):
+        block1 = block(in_channels, out_channels, norm_layer=nn.BatchNorm2d, stride=2)
+        block2 = block(out_channels, out_channels, norm_layer=nn.BatchNorm2d, stride=1)
+        return nn.Sequential(block1, block2)
+
+    def forward(self, x):
+        x = self.base_encoder(x)
+        outs = [self.output_layers[0](x)]
+        for i in range(0, self.num_level - 1):
+            x = self.downsamplers[i](x)
+            outs.append(self.output_layers[i + 1](x))
+        return outs
+
+
+# REUSE FROM torchvision.models.optical_flow.raft.MotionEncoder
+class MotionEncoder(nn.Module):
+    """The motion encoder, part of the update block.
+
+    Takes the current predicted flow and the correlation features as input and returns an encoded version of these.
+    """
+
+    def __init__(self, *, in_channels_corr, corr_layers=(256, 192), flow_layers=(128, 64), out_channels=128):
+        super().__init__()
+
+        if len(flow_layers) != 2:
+            raise ValueError(f"The expected number of flow_layers is 2, instead got {len(flow_layers)}")
+        if len(corr_layers) not in (1, 2):
+            raise ValueError(f"The number of corr_layers should be 1 or 2, instead got {len(corr_layers)}")
+
+        self.convcorr1 = Conv2dNormActivation(in_channels_corr, corr_layers[0], norm_layer=None, kernel_size=1)
+        if len(corr_layers) == 2:
+            self.convcorr2 = Conv2dNormActivation(corr_layers[0], corr_layers[1], norm_layer=None, kernel_size=3)
+        else:
+            self.convcorr2 = nn.Identity()
+
+        self.convflow1 = Conv2dNormActivation(2, flow_layers[0], norm_layer=None, kernel_size=7)
+        self.convflow2 = Conv2dNormActivation(flow_layers[0], flow_layers[1], norm_layer=None, kernel_size=3)
+
+        # out_channels - 2 because we cat the flow (2 channels) at the end
+        self.conv = Conv2dNormActivation(
+            corr_layers[-1] + flow_layers[-1], out_channels - 2, norm_layer=None, kernel_size=3
+        )
+
+        self.out_channels = out_channels
+
+    def forward(self, flow, corr_features):
+        corr = self.convcorr1(corr_features)
+        corr = self.convcorr2(corr)
+
+        flow_orig = flow
+        flow = self.convflow1(flow)
+        flow = self.convflow2(flow)
+
+        corr_flow = torch.cat([corr, flow], dim=1)
+        corr_flow = self.conv(corr_flow)
+        return torch.cat([corr_flow, flow_orig], dim=1)
+
+
+# REUSE torchvision.models.optical_flow.raft.ConvGRU
+class ConvGRU(nn.Module):
+    """Convolutional Gru unit."""
+
+    def __init__(self, *, input_size, hidden_size, kernel_size, padding):
+        super().__init__()
+        self.convz = nn.Conv2d(hidden_size + input_size, hidden_size, kernel_size=kernel_size, padding=padding)
+        self.convr = nn.Conv2d(hidden_size + input_size, hidden_size, kernel_size=kernel_size, padding=padding)
+        self.convq = nn.Conv2d(hidden_size + input_size, hidden_size, kernel_size=kernel_size, padding=padding)
+
+    def forward(self, h, x):
+        hx = torch.cat([h, x], dim=1)
+        z = torch.sigmoid(self.convz(hx))
+        r = torch.sigmoid(self.convr(hx))
+        q = torch.tanh(self.convq(torch.cat([r * h, x], dim=1)))
+        h = (1 - z) * h + z * q
+        return h
+
+
+# MODIFIED torchvision.models.optical_flow.raft.FlowHead
+class DepthHead(nn.Module):
+    """Depth head, part of the update block.
+
+    Takes the hidden state of the recurrent unit as input, and outputs the predicted "delta depth".
+    """
+
+    def __init__(self, *, in_channels, hidden_size, out_channels):
+        super().__init__()
+        self.conv1 = nn.Conv2d(in_channels, hidden_size, 3, padding=1)
+        self.conv2 = nn.Conv2d(hidden_size, out_channels, 3, padding=1)
+        self.relu = nn.ReLU(inplace=True)
+
+    def forward(self, x):
+        return self.conv2(self.relu(self.conv1(x)))
+
+
+class MultiLevelUpdateBlock(nn.Module):
+    """The update block which contains the motion encoder and grus
+
+    It must expose a ``hidden_dims`` attribute which is the hidden dimension size of its gru blocks
+    """
+
+    def __init__(self, *, motion_encoder, hidden_dims=[128, 128, 128]):
+        super().__init__()
+        self.motion_encoder = motion_encoder
+
+        # The GRU input size is the size of previous level hidden_dim plus next level hidden_dim
+        # if this is the first gru, then we replace previous level with motion_encoder output channels
+        # for the last GRU, we dont add the next level hidden_dim
+        gru_input_dims = []
+        for i in range(len(hidden_dims)):
+            input_dim = hidden_dims[i - 1] if i > 0 else motion_encoder.out_channels
+            if i < len(hidden_dims) - 1:
+                input_dim += hidden_dims[i + 1]
+            gru_input_dims.append(input_dim)
+
+        self.grus = nn.ModuleList([
+            ConvGRU(input_size=gru_input_dims[i], hidden_size=hidden_dims[i], kernel_size=3, padding=1)
+            for i in range(len(hidden_dims))
+        ])
+
+        self.hidden_dims = hidden_dims
+
+    def _downsample2x(self, x):
+        return F.avg_pool2d(x, 3, stride=2, padding=1)
+
+    def _upsample2x(self, x):
+        _, _, h, w = x.shape
+        new_h, new_w = h * 2, w * 2
+        return F.interpolate(x, size=(new_h, new_w), mode="bilinear", align_cornders=True)
+
+    def forward(self, hidden_states, contexts, corr_features, depth, num_processed=[1, 1, 1]):
+        for i in range(len(self.grus) - 1, -1, -1):
+            for it in num_processed:
+                # X is concatination of downsampled hidden_dim (or motion_features if no bigger dim) with
+                # upsampled hidden_dim (or nothing if not exist)
+                features = self._downsample2x(hidden_states[i - 1]) if i > 0 else self.motion_encoder(depth, corr_features)
+                if i < len(self.grus) - 1:
+                    features = torch.cat([features, self._upsample2x(hidden_states[i + 1])], dim=1)
+                x = torch.cat([contexts[i], features], dim=1)
+
+                hidden_states[i] = self.grus[i](hidden_states[i], x)
+
+                # NOTE: For slow-fast gru, we dont always want to calculate delta depth for every call on UpdateBlock
+                # Hence we move the delta depth calculation to the RAFT-Stereo main forward
+
+        return hidden_states
+
+
+# MODIFIED from torchvision.models.optical_flow.raft.MaskPredictor
+class MaskPredictor(nn.Module):
+    """Mask predictor to be used when upsampling the predicted depth.
+    """
+
+    def __init__(self, *, in_channels, hidden_size, out_channels, multiplier=0.25):
+        super().__init__()
+        self.convrelu = Conv2dNormActivation(in_channels, hidden_size, norm_layer=None, kernel_size=3)
+        self.conv = nn.Conv2d(hidden_size, out_channels, kernel_size=1, padding=0)
+
+        # In the original code, they use a factor of 0.25 to "downweight the gradients" of that branch.
+        # See e.g. https://github.com/princeton-vl/RAFT/issues/119#issuecomment-953950419
+        # or https://github.com/princeton-vl/RAFT/issues/24.
+        # It doesn't seem to affect epe significantly and can likely be set to 1.
+        self.multiplier = multiplier
+
+    def forward(self, x):
+        x = self.convrelu(x)
+        x = self.conv(x)
+        return self.multiplier * x
+
+
+# MODIFIED from torchvision.models.optical_flow.raft.CorrBlock to only consider 1d neighbour instead of 2d
+class CorrBlock(nn.Module):
+    """The correlation block.
+
+    Creates a correlation pyramid with ``num_levels`` levels from the outputs of the feature encoder,
+    and then indexes from this pyramid to create correlation features.
+    The "indexing" of a given centroid pixel x' is done by concatenating its surrounding row neighbours
+    within radius
+    """
+
+    def __init__(self, *, num_levels: int = 4, radius: int = 4):
+        super().__init__()
+        self.num_levels = num_levels
+        self.radius = radius
+
+        self.corr_pyramid: List[Tensor] = [torch.tensor(0)]  # useless, but torchscript is otherwise confused :')
+
+        self.out_channels = num_levels * (2 * radius + 1)
+
+    def build_pyramid(self, fmap1, fmap2):
+        """Build the correlation pyramid from two feature maps.
+
+        The correlation volume is first computed as the dot product of each pair (pixel_in_fmap1, pixel_in_fmap2)
+        The last 2 dimensions of the correlation volume are then pooled num_levels times at different resolutions
+        to build the correlation pyramid.
+        """
+
+        if fmap1.shape != fmap2.shape:
+            raise ValueError(
+                f"Input feature maps should have the same shape, instead got {fmap1.shape} (fmap1.shape) != {fmap2.shape} (fmap2.shape)"
+            )
+        corr_volume = self._compute_corr_volume(fmap1, fmap2)
+
+        batch_size, h, w, num_channels, _ = corr_volume.shape
+        corr_volume = corr_volume.reshape(batch_size * h * w, num_channels, 1, w)
+        self.corr_pyramid = [corr_volume]
+        for _ in range(self.num_levels - 1):
+            corr_volume = F.avg_pool2d(corr_volume, kernel_size=2, stride=2)
+            self.corr_pyramid.append(corr_volume)
+
+    def index_pyramid(self, centroids_coords):
+        """Return correlation features by indexing from the pyramid."""
+        neighborhood_side_len = 2 * self.radius + 1  # see note in __init__ about out_channels
+        di = torch.linspace(-self.radius, self.radius, neighborhood_side_len)
+        di = di.view(1, 1, neighborhood_side_len, 1).to(centroid_coords.device)
+
+        batch_size, _, h, w = centroids_coords.shape  # _ = 2
+        centroids_coords = centroids_coords.permute(0, 2, 3, 1).reshape(batch_size * h * w, 1, 1, 1)
+
+        indexed_pyramid = []
+        for corr_volume in self.corr_pyramid:
+            x0 = centroids_coords + di  # end shape is (batch_size * h * w, 1, side_len, 1)
+            y0 = torch.zeros_like(x0)
+            sampling_coords = torch.cat([x0, y0], dim=1)
+            indexed_corr_volume = grid_sample(corr_volume, sampling_coords, align_corners=True, mode="bilinear").view(
+                batch_size, h, w, -1
+            )
+            indexed_pyramid.append(indexed_corr_volume)
+            centroids_coords = centroids_coords / 2
+
+        corr_features = torch.cat(indexed_pyramid, dim=-1).permute(0, 3, 1, 2).contiguous()
+
+        expected_output_shape = (batch_size, self.out_channels, h, w)
+        if corr_features.shape != expected_output_shape:
+            raise ValueError(
+                f"Output shape of index pyramid is incorrect. Should be {expected_output_shape}, got {corr_features.shape}"
+            )
+
+        return corr_features
+
+    def _compute_corr_volume(self, fmap1, fmap2):
+        batch_size, num_channels, h, w = fmap1.shape
+        fmap1 = fmap1.view(batch_size, num_channels, h, w)
+        fmap2 = fmap2.view(batch_size, num_channels, h, w)
+
+        corr = torch.einsum("aijk,aijh->ajkh", fmap1, fmap2) 
+        corr = corr.view(batch_size, h, w, 1, w)
+        return corr / torch.sqrt(torch.tensor(num_channels))
+
+
+# MODIFIED from torchvision.models.optical_flow.raft.RAFT
+class RaftStereo(nn.Module):
+    def __init__(self, *, feature_encoder, context_encoder,
+            corr_block, update_block, depth_head, mask_predictor=None, slow_fast=False):
+        """RAFT-Stereo model from
+        `RAFT: Recurrent All Pairs Field Transforms for Optical Flow <https://arxiv.org/abs/2003.12039>`_.
+        `RAFT-Stereo: Multilevel Recurrent Field Transforms for Stereo Matching <https://arxiv.org/abs/2109.07547>`_.
+
+        args:
+            feature_encoder (nn.Module): The feature encoder. It must downsample the input by 8.
+                Its input is the concatenation of ``image1`` and ``image2``.
+            context_encoder (nn.Module): The context encoder. It must downsample the input by 8.
+                Its input is ``image1``. As in the original implementation, its output will be split into 2 parts:
+
+                - one part will be used as the actual "context", passed to the recurrent unit of the ``update_block``
+                - one part will be used to initialize the hidden state of the of the recurrent unit of
+                  the ``update_block``
+
+                These 2 parts are split according to the ``hidden_state_size`` of the ``update_block``, so the output
+                of the ``context_encoder`` must be strictly greater than ``hidden_state_size``.
+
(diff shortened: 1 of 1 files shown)


torchvision/prototype/models/depth/stereo/raft_stereo.py
```diff
@@ -0,0 +1,692 @@
...
+    up_mask = up_mask.view(batch_size, 1, 9, factor, factor, h, w)
+    up_mask = torch.softmax(up_mask, dim=2)  # "convex" == weights sum to 1
+
+    upsampled_x = F.unfold(factor * x, kernel_size=3, padding=1).view(batch_size, num_channels, 9, 1, 1, h, w)
+    upsampled_x = torch.sum(up_mask * upsampled_x, dim=2)
+
+    return upsampled_x.permute(0, 1, 4, 2, 5, 3).reshape(batch_size, num_channels, new_h, new_w)
+
+
+# REUSING from torchvision.models.optical_flow.raft
```
datumbox, line 41: If you straight reuse something at this stage, it might be best to import it from raft to avoid the degree of copy-pasting and assist reviewing. We won't leave it like that at the end, but it will really help us at the early stages of the work.

torchvision/prototype/models/depth/stereo/raft_stereo.py
```diff
@@ -0,0 +1,692 @@
...
+        y = x
+        y = self.convnormrelu1(y)
+        y = self.convnormrelu2(y)
+
+        x = self.downsample(x)
+
+        return self.relu(x + y)
+
+
+# MODIFIED from torchvision.models.optical_flow.raft.FeatureEncoder by adding strides param
```
datumbox, line 86: Similarly, would it make sense to use inheritance to overwrite part of the behaviour you want? Again the focus here is to simplify reviews and reduce the size of the code-base. We will later decide how to handle these on the clean ups. If not possible, feel free to leave as-is.

datumbox (commented):
Some ultra early thoughts. Feel free to ignore if you are not looking at all for input at this point.

YosuaMichael:
I have tested this implementation by loading the pretrained weight from original implementation https://github.com/princeton-vl/RAFT-Stereo and comparing the result of forward pass on both models (the original and this prototype torchvision model).

Here is the script that I used for the testing: https://gist.github.com/YosuaMichael/4b402f84ab6365eeac20c0dcd318f618

And here are the command I use for the script and the result:

Verify standard model:
```
$ python verify_load_pretrained_weight.py --model_type=standard
[RESULT] Have exactly same result!
[RESULT] Ori run in 6.985464096069336 seconds
[RESULT] TV run in 6.933291912078857 seconds
```
Verify fast model:
```
python verify_load_pretrained_weight.py --model_type=fast
[RESULT] Have exactly same result!
[RESULT] Ori run in 2.0360748767852783 seconds
[RESULT] TV run in 2.169339895248413 seconds
```

From the result, both model have exactly same output and their running time are similar.

YosuaMichael:
@datumbox  I have updated this PR and it should be kinda ready-ish to be reviewed if you have free time please help me to review this. I have tested this model class and it seems to agree with the original implementation (see https://github.com/pytorch/vision/pull/6107#issuecomment-1150908902).

One question that I have in mind now is about the helper functions. Do you think I should create a separate file `_utils.py` to put the helper functions?

torchvision/prototype/models/depth/stereo/raft_stereo.py
```diff
@@ -0,0 +1,746 @@
...
+        super().__init__()
+        self.num_level = len(out_with_blocks)
+        self.base_encoder = base_encoder
+        self.base_downsampling_ratio = base_encoder.downsampling_ratio
+        base_dim = base_encoder.output_dim
+
+        # Layer to output hidden_state and context separately (each produce output_dim//2 dims)
+        self.out_hidden_states = nn.ModuleList(
+            [
+                self._make_out_layer(base_dim, output_dim // 2, with_block=with_block, block=block)
```
datumbox, line 162: Optional Nit: We rarely do the construction of the layers in private methods any more as it hurts readability. Consider simplifying the logic of the method by always building sequential layers and moving the construction in the constructor.

torchvision/prototype/models/depth/stereo/raft_stereo.py
```diff
@@ -0,0 +1,746 @@
...
+        block1 = block(in_channels, out_channels, norm_layer=nn.BatchNorm2d, stride=2)
+        block2 = block(out_channels, out_channels, norm_layer=nn.BatchNorm2d, stride=1)
+        return nn.Sequential(block1, block2)
+
+    def forward(self, x: Tensor) -> List[Tensor]:
+        x = self.base_encoder(x)
+        outs = [torch.cat([self.out_hidden_states[0](x), self.out_contexts[0](x)], dim=1)]
+        for i in range(0, self.num_level - 1):
+            x = self.downsamplers[i](x)
+            outs.append(torch.cat([self.out_hidden_states[i + 1](x), self.out_contexts[i + 1](x)], dim=1))
```
datumbox, line 195: Nit: You can avoid having the `torch.cat` twice here by doing minor refactoring.

torchvision/prototype/models/depth/stereo/raft_stereo.py
```diff
@@ -0,0 +1,746 @@
...
+
+        self.hidden_dims = hidden_dims
+
+    def _downsample2x(self, x):
+        return F.avg_pool2d(x, 3, stride=2, padding=1)
+
+    def _upsample2x(self, x):
+        _, _, h, w = x.shape
+        new_h, new_w = h * 2, w * 2
+        return F.interpolate(x, size=(new_h, new_w), mode="bilinear", align_corners=True)
```
datumbox, line 249: Nit: we can do without these private methods as you use them only once.

torchvision/prototype/models/depth/stereo/raft_stereo.py
```diff
@@ -0,0 +1,746 @@
...
+        corr_volume = self._compute_corr_volume(fmap1, fmap2)
+
+        batch_size, h, w, num_channels, _ = corr_volume.shape
+        corr_volume = corr_volume.reshape(batch_size * h * w, num_channels, 1, w)
+        self.corr_pyramid = [corr_volume]
+        for _ in range(self.num_levels - 1):
+            corr_volume = F.avg_pool2d(corr_volume, kernel_size=(1, 2), stride=(1, 2))
+            self.corr_pyramid.append(corr_volume)
+
+    def index_pyramid(self, centroids_coords: Tensor) -> Tensor:
```
datumbox, line 327: This idiom is problematic. You are inheriting from nn.Module, yet this block doesn't have forward. Should this be the forward method?

torchvision/prototype/models/depth/stereo/raft_stereo.py
```diff
@@ -0,0 +1,746 @@
...
+    The "indexing" of a given centroid pixel x' is done by concatenating its surrounding row neighbours
+    within radius
+    """
+
+    def __init__(self, *, num_levels: int = 4, radius: int = 4):
+        super().__init__()
+        self.num_levels = num_levels
+        self.radius = radius
+
+        self.corr_pyramid: List[Tensor] = [torch.tensor(0)]  # useless, but torchscript is otherwise confused :')
```
datumbox, line 302: Show us the message you get if you oit the torch.tensor here. Let's see if we can fix JIT without the workaround.

torchvision/prototype/models/depth/stereo/raft_stereo.py
```diff
@@ -0,0 +1,746 @@
...
+                f"Input feature maps should have the same shape, instead got {fmap1.shape} (fmap1.shape) != {fmap2.shape} (fmap2.shape)"
+            )
+        corr_volume = self._compute_corr_volume(fmap1, fmap2)
+
+        batch_size, h, w, num_channels, _ = corr_volume.shape
+        corr_volume = corr_volume.reshape(batch_size * h * w, num_channels, 1, w)
+        self.corr_pyramid = [corr_volume]
+        for _ in range(self.num_levels - 1):
+            corr_volume = F.avg_pool2d(corr_volume, kernel_size=(1, 2), stride=(1, 2))
+            self.corr_pyramid.append(corr_volume)
```
datumbox, line 325: From checking how this is used, it seems you are modifying the nn.Module on each batch forward and leave side-effects on the method. Can you try refactoring this (perhaps splitting in multiple modules?) to avoid keeping this information within the class? I haven't checked everything in details to tell you how to do it but I can further have a more detailed view if it helps.

torchvision/prototype/models/depth/stereo/raft_stereo.py
```diff
@@ -0,0 +1,746 @@
...
+
+        expected_output_shape = (batch_size, self.out_channels, h, w)
+        if corr_features.shape != expected_output_shape:
+            raise ValueError(
+                f"Output shape of index pyramid is incorrect. Should be {expected_output_shape}, got {corr_features.shape}"
+            )
+
+        return corr_features
+
+    def _compute_corr_volume(self, fmap1: Tensor, fmap2: Tensor) -> Tensor:
```
datumbox, line 358: Nit: lose private methods that are called once

torchvision/prototype/models/depth/stereo/raft_stereo.py
```diff
@@ -0,0 +1,746 @@
...
+
+class Raft_Stereo_Fast_Weights(WeightsEnum):
+    pass
+
+
+class Raft_Stereo_Weights(WeightsEnum):
+    pass
+
+
+def raft_stereo_fast(*, weights: Optional[Raft_Stereo_Fast_Weights] = None, progress=True, **kwargs) -> RaftStereo:
```
datumbox, line 635: Is this the `Slow-fast` variant? If yes, we should name as such because the "slow-fast" approach is something recognizable across papers.

torchvision/prototype/models/depth/stereo/raft_stereo.py
```diff
@@ -0,0 +1,746 @@
...
+            x0 = centroids_coords + di  # end shape is (batch_size * h * w, 1, side_len, 1)
+            y0 = torch.zeros_like(x0)
+            sampling_coords = torch.cat([x0, y0], dim=-1)
+            indexed_corr_volume = grid_sample(corr_volume, sampling_coords, align_corners=True, mode="bilinear").view(
+                batch_size, h, w, -1
+            )
+            indexed_pyramid.append(indexed_corr_volume)
+            centroids_coords = centroids_coords / 2
+
+        corr_features = torch.cat(indexed_pyramid, dim=-1).permute(0, 3, 1, 2).contiguous()
```
datumbox, line 348: Out of curiosity, why the `contiguous()` call? Is it for performance reasons? Have you measured it personally or found it on the original implementation?

datumbox (commented):
Thanks @YosuaMichael. I like the approach you took to inherit from RAFT. I would crank it up a notch and reuse as much as possible at this point, including in cases where minor modifications are necessary to make its classes more generic. This will help us reuse as much code as possible and make easier decision on how to restructure the code. I've also added a few comments regarding some of the idioms you use. Let me know what you think.

Because the implementation is large and uses lots of new idioms, we might need to do a couple of rounds of reviews. While you work on the comments, I recommend marking the implementation as draft and unmarking it when you need fresh reviews.

torchvision/prototype/models/depth/stereo/raft_stereo.py
```diff
@@ -0,0 +1,746 @@
...
+        x = self.conv(x)
+        return x
+
+
+class MultiLevelContextEncoder(nn.Module):
+    def __init__(
+        self,
+        base_encoder: nn.Module,
+        output_dim: int = 256,
+        out_with_blocks: List[int] = [1, 1, 0],
```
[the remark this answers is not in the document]
YosuaMichael: Yes, it supposed to be a list of booleans. I will remove the default value for all the list parameters and for this I will use booleans instead.

torchvision/prototype/models/depth/stereo/raft_stereo.py
YosuaMichael: Thanks, will try to simplify this.

torchvision/prototype/models/depth/stereo/raft_stereo.py
YosuaMichael: In my opinion, using private method here actually help the readability. 
```
features = torch.cat([features, self._upsample2x(hidden_states[i + 1])], dim=1)
```
It is clear that we want to do upsample (from the function name) on hidden_states[i+1] (in 1 line). And reader who want to know more detail on how we do the upsampling can read the function.

Do you think it is okay to keep the private function?

torchvision/prototype/models/depth/stereo/raft_stereo.py
YosuaMichael: It use slow-fast gru, but not just that. I think the paper refer this model as "realtime model". It is specified on section 4.7 on RAFT-Stereo paper.

torchvision/prototype/models/depth/stereo/raft_stereo.py
YosuaMichael: This class is actually heavily modified from the raft: https://github.com/pytorch/vision/blob/main/torchvision/models/optical_flow/raft.py#L328

This part comes from there. Let me try it out.

torchvision/prototype/models/depth/stereo/raft_stereo.py
YosuaMichael: Ah yes, this can be the forward method, you are right.

torchvision/prototype/models/depth/stereo/raft_stereo.py
YosuaMichael: I follow the raft implementation for this: https://github.com/pytorch/vision/blob/main/torchvision/models/optical_flow/raft.py#L392

torchvision/prototype/models/depth/stereo/raft_stereo.py
YosuaMichael: Agree, will remove this private method.

torchvision/prototype/models/depth/stereo/raft_stereo.py
YosuaMichael: Yes, this is called on each batch forward once to update `self.corr_pyramid`. I think we can put this `self.corr_pyramid` on the RaftStereo class instead, do you think this is preferable?

torchvision/prototype/models/depth/stereo/raft_stereo.py
```diff
@@ -0,0 +1,731 @@
...
+
+    upsampled_x = F.unfold(factor * x, kernel_size=3, padding=1).view(batch_size, num_channels, 9, 1, 1, h, w)
+    upsampled_x = torch.sum(up_mask * upsampled_x, dim=2)
+
+    return upsampled_x.permute(0, 1, 4, 2, 5, 3).reshape(batch_size, num_channels, new_h, new_w)
+
+
+# MODIFIED from torchvision.models.optical_flow.raft.FeatureEncoder by adding strides param
+# and remove the last conv layer
+class BaseEncoder(nn.Module):
```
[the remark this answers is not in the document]
YosuaMichael: Ah, good suggestion with the Identity. Will try to do that.

torchvision/prototype/models/depth/stereo/raft_stereo.py
YosuaMichael: Actually thanks to this, I realized the model is not yet jit scriptable. After make change to make sure the model is jit scriptable, when I remove the `torch.tensor(0)` I will got the following error message:
```
RuntimeError: 
Tried to set nonexistent attribute: corr_pyramid. Did you forget to initialize it in __init__()?:
  File "/Users/yosuamichael/manual_build_lib/pytorch/vision/torchvision/prototype/models/depth/stereo/raft_stereo.py", line 251
        batch_size, h, w, num_channels, _ = corr_volume.shape
        corr_volume = corr_volume.reshape(batch_size * h * w, num_channels, 1, w)
        self.corr_pyramid = [corr_volume]
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ <--- HERE
        for _ in range(self.num_levels - 1):
            corr_volume = F.avg_pool2d(corr_volume, kernel_size=(1, 2), stride=(1, 2))
'CorrBlock1d.build_pyramid' is being compiled since it was called from 'RaftStereo.forward'
  File "/Users/yosuamichael/manual_build_lib/pytorch/vision/torchvision/prototype/models/depth/stereo/raft_stereo.py", line 383
            raise ValueError(f"The feature encoder should downsample H and W by {self.base_downsampling_ratio}")
    
        self.corr_block.build_pyramid(fmap1, fmap2)
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ <--- HERE

```

torchvision/prototype/models/depth/stereo/raft_stereo.py
YosuaMichael: Ok, I have separate the module into two separate modules (one for corr_pyramid, another one for building corr_features)

torchvision/prototype/models/depth/stereo/raft_stereo.py
YosuaMichael: I have tested removing the `.contiguous()` call, and seems like there is no error nor significant change in performance (there is only a very small change on the output result but it still within 1e-6).

Take note however that both the [torchvision's RAFT implementation](https://github.com/pytorch/vision/blob/main/torchvision/models/optical_flow/raft.py#L392) and the [original RAFT-Stereo implementation](https://github.com/princeton-vl/RAFT-Stereo/blob/main/core/corr.py#L146) have this contiguous call.

torchvision/prototype/models/depth/stereo/raft_stereo.py
YosuaMichael: After separating the corr_block into 2 modules, and putting the `corr_pyramid` in the `RaftStereo` forward, this issue is not relevant anymore.

torchvision/prototype/models/depth/stereo/raft_stereo.py
```diff
@@ -0,0 +1,696 @@
...
+        output_dim: int = 256,
+        block: Callable[..., nn.Module] = ResidualBlock,
+    ):
+        super().__init__()
+        self.num_level = len(out_with_blocks)
+        self.base_encoder = base_encoder
+        self.base_downsampling_ratio = base_encoder.downsampling_ratio
+        base_dim = base_encoder.output_dim
+
+        self.downsample_and_out_layers = nn.ModuleList(
```
YosuaMichael, line 94: We wrap the multi level output block and downsampler block into `ModuleList of ModuleDict` in order to make the model jit scriptable.
Previously we have 3 separate `ModuleList` and we need to iterate and access with index on one for loop, however jit script only allow to iterate module or access with literal integer.

torchvision/prototype/models/depth/stereo/raft_stereo.py
datumbox: I agreed that readability can be quite subjective. There is an argument to be made about keeping the styling of the code similar across the code-base. Having 1-liner private methods used once is probably something we try to avoid across the code-base. As far as I see, you already have comments to discuss what the 2 function calls are doing. Do the separate methods really add more?

torchvision/prototype/models/depth/stereo/raft_stereo.py
datumbox: Thank for the references. Page 8, just before section 4.7 reads:
> In table 6, the ”Slow-Fast” version of RAFT-Stereo updates the lowest, middle and highest resolution hidden states 30, 20 and 10 times, respectively, while ”Regular” updates each hidden state 32 times. 

If they refer as "slow-fast" to this variant, I recommend naming it slowfast. It's a recognizable name in the literature.

torchvision/prototype/models/depth/stereo/raft_stereo.py
datumbox: Let's leave and benchmark later.

torchvision/prototype/models/depth/stereo/raft_stereo.py
```diff
@@ -0,0 +1,696 @@
...
+
+    def __init__(self, *, num_levels: int = 4, radius: int = 4):
+        super().__init__()
+        self.radius = radius
+        self.out_channels = num_levels * (2 * radius + 1)
+
+    def forward(self, centroids_coords: Tensor, corr_pyramid: List[Tensor]) -> Tensor:
+        """Return correlation features by indexing from the pyramid."""
+        neighborhood_side_len = 2 * self.radius + 1  # see note in __init__ about out_channels
+        di = torch.linspace(-self.radius, self.radius, neighborhood_side_len)
```
datumbox, line 290: Similarly, let's put a device here and in all other places where you create new tensors.

torchvision/prototype/models/depth/stereo/raft_stereo.py
```diff
@@ -0,0 +1,696 @@
...
+                batch_size, h, w, -1
+            )
+            indexed_pyramid.append(indexed_corr_volume)
+            centroids_coords = centroids_coords / 2
+
+        corr_features = torch.cat(indexed_pyramid, dim=-1).permute(0, 3, 1, 2).contiguous()
+
+        expected_output_shape = (batch_size, self.out_channels, h, w)
+        if corr_features.shape != expected_output_shape:
+            raise ValueError(
```
datumbox, line 312: Another place where possibly you need torch._assert. Basically you might need to remove all ValueError exceptions from `forward()` calls to make it FX-traceable. You should put a test to assist your development and catch these early.

torchvision/prototype/models/depth/stereo/raft_stereo.py
```diff
@@ -0,0 +1,696 @@
...
+
+
+def _raft_stereo(
+    *,
+    weights: Optional[WeightsEnum] = None,
+    progress: bool = False,
+    shared_encoder_weight: bool = False,
+    # Feature encoder
+    feature_encoder_layers: Tuple[int, int, int, int, int],
+    feature_encoder_strides: Tuple[int, int, int, int],
```
datumbox, line 461: Are you sure these shouldn't be `List[int]`? Is it possible that someone might want to make the encoder deeper?

datumbox (commented):
Thanks @YosuaMichael, a few more comments. Let me know what you think below.

torchvision/prototype/models/depth/stereo/raft_stereo.py
```diff
@@ -0,0 +1,696 @@
...
+    def forward(self, x: Tensor) -> List[Tensor]:
+        x = self.base_encoder(x)
+        outs = []
+        for layer_dict in self.downsample_and_out_layers:
+            x = layer_dict["downsampler"](x)
+            outs.append(torch.cat([layer_dict["out_hidden_state"](x), layer_dict["out_context"](x)], dim=1))
+        return outs
+
+
+class ConvGRU(nn.Module):
```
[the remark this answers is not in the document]
YosuaMichael: Here is the mypy error if I specify the typing:
```
raft_stereo.py:137: error: Signature of "forward" incompatible with supertype "ConvGRU"  [override]
        def forward(self, h: Tensor, context: List[Tensor], x: Tensor):
        ^
raft_stereo.py:137: note:      Superclass:
raft_stereo.py:137: note:          def forward(self, h: Any, x: Any) -> Any
raft_stereo.py:137: note:      Subclass:
raft_stereo.py:137: note:          def forward(self, h: Tensor, context: List[Tensor], x: Tensor) -> Any
Found 1 error in 1 file (checked 1 source file)
```
I have tried to put the `context` after `x` in the argument list, but it will still got the same error.
Also, here is a relevant discussion on this error: https://stackoverflow.com/questions/54346721/mypy-argument-of-method-incompatible-with-supertype

torchvision/prototype/models/depth/stereo/raft_stereo.py
```diff
@@ -0,0 +1,696 @@
...
+    def forward(self, fmap1: Tensor, fmap2: Tensor) -> List[Tensor]:
+        """Build the correlation pyramid from two feature maps.
+
+        The correlation volume is first computed as the dot product of each pair (pixel_in_fmap1, pixel_in_fmap2) on the same row.
+        The last 2 dimensions of the correlation volume are then pooled num_levels times at different resolutions
+        to build the correlation pyramid.
+        """
+
+        if fmap1.shape != fmap2.shape:
+            raise ValueError(
```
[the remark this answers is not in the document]
YosuaMichael: I haven't really check for FX-traceable and thanks for pointing this out (on later comment as well). Will create a test for this.

torchvision/prototype/models/depth/stereo/raft_stereo.py
YosuaMichael: Yes, as of now the number of layer on the encoder is not variable size. On the raft implementation we put check on this: https://github.com/pytorch/vision/blob/main/torchvision/models/optical_flow/raft.py#L122

torchvision/prototype/models/depth/stereo/raft_stereo.py
YosuaMichael: I think the one they refer on the last part of section 4.6 is different model that they use for ablation study where they want to compare using `Slow-Fast` vs not using it (shown in table 6).
The model that I implement (and the one they provide pretrained weight on) has combination of techniques they explained on the ablation. Other than using slow-fast GRU, it use more downsampling, less GRU, shared backbone (kinda use all result of table 6).
In the original repo they call it `raftstereo-realtime.pth` (see last part of demos https://github.com/princeton-vl/RAFT-Stereo#demos ). 
Maybe I should call it `raft_stereo_realtime` instead?

torchvision/prototype/models/depth/stereo/raft_stereo.py
```diff
@@ -0,0 +1,710 @@
...
+
+def _zeroes_second_channel(delta_depth: Tensor):
+    delta_depth[:, 1] = 0.0
+
+
+torch.fx.wrap("_corr_channel_normalize")
+torch.fx.wrap("_zeroes_second_channel")
+torch.fx.wrap("make_coords_grid")
+torch.fx.wrap("grid_sample")
+
```
YosuaMichael, line 36: We wrap some global functions (including one that is imported from raft) with `torch.fx.wrap` to make sure the model is fx compatible.

YosuaMichael, line 36: We wrap some global functions (including one that is imported from raft) with torch.fx.wrap to make sure the model is fx compatible.

torchvision/prototype/models/depth/stereo/raft_stereo.py
```diff
@@ -0,0 +1,710 @@
...
+        hidden_dims = self.update_block.hidden_dims
+        # Follow the original implementation to do pre convolution on the context
+        # See: https://github.com/princeton-vl/RAFT-Stereo/blob/main/core/raft_stereo.py#L32
+        self.context_convs = nn.ModuleList(
+            [nn.Conv2d(hidden_dims[i], hidden_dims[i] * 3, kernel_size=3, padding=1) for i in range(self.num_level)]
+        )
+        self.slow_fast = slow_fast
+        self.num_iters = num_iters
+
+    def forward(self, image1: Tensor, image2: Tensor) -> List[Tensor]:
```
YosuaMichael, line 391: Previously we have `num_iters` as input to forward as this is also the case in the [original implementation](https://github.com/princeton-vl/RAFT-Stereo/blob/main/core/raft_stereo.py#L70). However, this violate the fx constraint, and we would need to wrap the big for loop (line 423 - 463) in order to make the model fx compatible.

Currently I decide to set the num_iters in the constructor instead, and users still can modify during inference by set it before calling the forward. However I am still not too sure this is a good idea, do you have any suggestion @datumbox ?

torchvision/prototype/models/depth/stereo/raft_stereo.py
```diff
@@ -0,0 +1,710 @@
...
+        num_iters = self.num_iters
+        batch_size, _, h, w = image1.shape
+        torch._assert(
+            (h, w) == image2.shape[-2:],
+            f"input images should have the same shape, instead got ({h}, {w}) != {image2.shape[-2:]}",
+        )
+
+        torch._assert(
+            # We use addition, "+", instead of "and" operator so it can pass fx symbolic trace
+            (h % self.base_downsampling_ratio + w % self.base_downsampling_ratio == 0),
```
YosuaMichael, line 403: Even inside `torch._assert` if we put the condition with `and` operator, the fx will still complain. Since in this case the behaviour of `+` and `and` should be exactly the same, therefore I use `+` instead.

torchvision/prototype/models/depth/stereo/raft_stereo.py
```diff
@@ -0,0 +1,710 @@
...
+            outs.append(torch.cat([layer_dict["out_hidden_state"](x), layer_dict["out_context"](x)], dim=1))
+        return outs
+
+
+class ConvGRU(raft.ConvGRU):
+    """Convolutional Gru unit."""
+
+    # Modified from raft.ConvGRU to accept pre-convolved contexts,
+    # see: https://github.com/princeton-vl/RAFT-Stereo/blob/main/core/update.py#L23
+    def forward(self, h: Tensor, context: List[Tensor], x: Tensor):
```
YosuaMichael, line 151: I understand that the current code will have mypy complained (about not compatible with supertype). We purposely revert this back to expose the error on the CI.

YosuaMichael (commented):
Thanks for your comments @datumbox , I have made some changes. The current changes focus on making sure the model is fx compatible. I also create a simple test file to check if the model jit scriptable and fx compatible.

torchvision/prototype/models/depth/stereo/raft_stereo.py
YosuaMichael: Here is the error on the CI: https://app.circleci.com/pipelines/github/pytorch/vision/18352/workflows/de86cd36-ddbc-4ea1-a726-30713617abbf/jobs/1484828

torchvision/prototype/models/depth/stereo/raft_stereo.py
```diff
@@ -0,0 +1,711 @@
...
+        hidden_dims = self.update_block.hidden_dims
+        # Follow the original implementation to do pre convolution on the context
+        # See: https://github.com/princeton-vl/RAFT-Stereo/blob/main/core/raft_stereo.py#L32
+        self.context_convs = nn.ModuleList(
+            [nn.Conv2d(hidden_dims[i], hidden_dims[i] * 3, kernel_size=3, padding=1) for i in range(self.num_level)]
+        )
+        self.slow_fast = slow_fast
+        self.num_iters = num_iters
+
+    def forward(self, image1: Tensor, image2: Tensor, num_iters: Optional[int] = None) -> List[Tensor]:
```
YosuaMichael, line 393: We decide to put back the `num_iters` as forward parameters (like the raft implementation). The consequences for this is that the model become not `fx compatible`.

torchvision/prototype/models/depth/stereo/raft_stereo.py
```diff
@@ -0,0 +1,711 @@
...
+            outs.append(torch.cat([layer_dict["out_hidden_state"](x), layer_dict["out_context"](x)], dim=1))
+        return outs
+
+
+class ConvGRU(raft.ConvGRU):
+    """Convolutional Gru unit."""
+
+    # Modified from raft.ConvGRU to accept pre-convolved contexts,
+    # see: https://github.com/princeton-vl/RAFT-Stereo/blob/main/core/update.py#L23
+    def forward(self, h: Tensor, x: Tensor, context: List[Tensor]) -> Tensor:  # type: ignore[override]
```
YosuaMichael, line 153: We add `#type: ignore[override]` to suppress mypy error.

YosuaMichael (commented):
On last changes, we do:
1) Suppress mypy error
2) Decide to prefer adding `num_iters` as forward parameter (similar with raft and original raft-stereo implementation) even if it make the model not `fx compatible`

torchvision/prototype/models/depth/stereo/raft_stereo.py
datumbox: Sounds good to me.

torchvision/prototype/models/depth/stereo/raft_stereo.py
```diff
@@ -0,0 +1,711 @@
...
+
+
+def _zeroes_second_channel(delta_depth: Tensor):
+    delta_depth[:, 1] = 0.0
+
+
+torch.fx.wrap("_corr_channel_normalize")
+torch.fx.wrap("_zeroes_second_channel")
+torch.fx.wrap("make_coords_grid")
+torch.fx.wrap("grid_sample")
```
datumbox, line 35: Nit: Now that we no longer aim for FX scriptability, I think we should remove these workarounds.

torchvision/prototype/models/depth/stereo/raft_stereo.py
```diff
@@ -0,0 +1,711 @@
...
+        )
+
+        self.hidden_dims = hidden_dims
+
+    def _downsample2x(self, x):
+        return F.avg_pool2d(x, 3, stride=2, padding=1)
+
+    def _upsample2x(self, x):
+        _, _, h, w = x.shape
+        return F.interpolate(x, size=(2 * h, 2 * w), mode="bilinear", align_corners=True)
```
datumbox, line 200: Nit: I still think these 2 methods need to be removed, if not for readability, for ensuring we follow the best practices of other parts of the code-base. Usually introducing private methods used once is discouraged.

torchvision/prototype/models/depth/stereo/raft_stereo.py
```diff
@@ -0,0 +1,711 @@
...
+        *,
+        feature_encoder: FeatureEncoder,
+        context_encoder: MultiLevelContextEncoder,
+        corr_pyramid: CorrPyramid1d,
+        corr_block: CorrBlock1d,
+        update_block: MultiLevelUpdateBlock,
+        depth_head: nn.Module,
+        mask_predictor: Optional[nn.Module] = None,
+        slow_fast: bool = False,
+        num_iters: int = 12,
```
datumbox, line 339: This needs to be removed as discussed yesterday, in favour of moving it to forward to make it possible to change during inference and align with RAFT.

torchvision/prototype/models/depth/stereo/raft_stereo.py
datumbox: We no longer worry about FX. We could perhaps use something more intuitive that + but no strong opinions.

torchvision/prototype/models/depth/stereo/raft_stereo.py
datumbox: Sounds good, let's use `raft_stereo_realtime`

torchvision/prototype/models/depth/stereo/raft_stereo.py
```diff
@@ -0,0 +1,711 @@
...
+        output_dim: int = 256,
+        shared_base: bool = False,
+        block: Callable[..., nn.Module] = ResidualBlock,
+    ):
+        super().__init__()
+        self.base_encoder = base_encoder
+        self.base_downsampling_ratio = base_encoder.downsampling_ratio
+        base_dim = base_encoder.output_dim
+
+        self.shared_base = shared_base
```
datumbox, line 77: No need to store it:
```suggestion
        self.shared_base = shared_base
```

torchvision/prototype/models/depth/stereo/raft_stereo.py
datumbox: Let's make it non-optional.

datumbox (commented):
@YosuaMichael I think we are getting very close to merging. I few more comments. I'll mark this draft, feel free to change when you are ready.

torchvision/prototype/models/depth/stereo/raft_stereo.py
```diff
@@ -269,7 +260,7 @@ def forward(self, fmap1: Tensor, fmap2: Tensor) -> List[Tensor]:
 
         corr = torch.einsum("aijk,aijh->ajkh", fmap1, fmap2)
         corr = corr.view(batch_size, h, w, 1, w)
-        corr_volume = _corr_channel_normalize(corr, num_channels)
+        corr_volume = corr / torch.sqrt(torch.tensor(num_channels).to(corr.device))
```
datumbox, line 263:
```suggestion
        corr_volume = corr / torch.sqrt(torch.tensor(num_channels, device=corr.device))
```

torchvision/prototype/models/depth/stereo/raft_stereo.py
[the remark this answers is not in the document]
YosuaMichael: This is solved by adding `# type: ignore[override]`

test/test_prototype_models.py
```diff
@@ -0,0 +1,28 @@
...
+    img2 = torch.rand(1, 3, 64, 64).to(dev)
+    num_iters = 3
+
+    preds = model(img1, img2, num_iters=num_iters)
+
+    assert len(preds) == num_iters, "Number of predictions should be the same as model.num_iters"
+
+    assert preds[0].shape == torch.Size(
+        [1, 1, 64, 64]
+    ), f"The output shape of preds[0] should be [1, 1, 64, 64] but instead it is {preds[0].shape}"
```
datumbox, line 28: We should be testing against an expected file, similar to all other model tests. This will help you catch issues and breakages from the original values.

datumbox (approved):
LGTM, in principle. I've added one comment concerning the test that I think it's worth addressing prior merging. The rest are optional or FYIs/Questions. You are adding things to prototype anyway so you will be able to do rapid experimentation.

torchvision/prototype/models/depth/stereo/raft_stereo.py
```diff
@@ -0,0 +1,701 @@
...
+
+class RaftStereo(nn.Module):
+    def __init__(
+        self,
+        *,
+        feature_encoder: FeatureEncoder,
+        context_encoder: MultiLevelContextEncoder,
+        corr_pyramid: CorrPyramid1d,
+        corr_block: CorrBlock1d,
+        update_block: MultiLevelUpdateBlock,
```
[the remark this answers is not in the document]
YosuaMichael: @datumbox at the beginning I use `nn.Module`, but after that mypy will complain if I access the variable of the object. For instance if I set feature_encoder as `nn.Module`, I will get the following mypy error:
```
raft_stereo.py:395: note: Right operand is of type "Union[Tensor, Module]"
raft_stereo.py:402: error: Unsupported operand types for // ("int" and Module)  [operator]
                fmap1.shape[-2:] == (h // self.base_downsampling_ratio, w // self.base_downsampling_ratio),
```
it is because the `base_downsampling_ratio` comes from feature encoder:
```
        self.base_downsampling_ratio = feature_encoder.base_downsampling_ratio
```
and mypy assumed any class variable of `nn.Module` as type `Union[Tensor, Module]`.

I think in this case we need to make decision whether to ignore these mypy error, or use specific class (that have the variable we wanted)

torchvision/prototype/models/depth/stereo/raft_stereo.py
```diff
@@ -0,0 +1,701 @@
...
+    # Update block
+    update_block_hidden_dims: List[int],
+    # Flow Head
+    flow_head_hidden_size: int,
+    # Mask predictor
+    mask_predictor_hidden_size: int,
+    use_mask_predictor: bool,
+    slow_fast: bool,
+    **kwargs,
+):
```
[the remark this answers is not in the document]
YosuaMichael: I think the only part where we dont really use is `use_mask_predictor`, since on raft stereo they always use one.
There some variable that have same value on `raft_stereo_basic` and `raft_stereo_realtime` , 

i.e.`feature_encoder_layers`, `corr_num_levels`, `corr_radius`, `motion_encoder_corr_layers`, `motion_encoder_flow_layers`, `motion_encoder_out_channels`, `flow_head_hidden_size`, `mask_predictor_hidden_size`
 
but I think they are still worth adding as parameter to `_raft_stereo`

torchvision/prototype/models/depth/stereo/raft_stereo.py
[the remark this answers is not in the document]
datumbox: Yeah fair enough. Let's keep it as is for now and see if we want to overwrite or workaround mypy when we clean up the implementation before graduating from prototype.

torchvision/prototype/models/depth/stereo/raft_stereo.py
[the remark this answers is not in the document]
datumbox: OK if there is anything you would like to clip, go ahead. Else mark as resolved.

torchvision/prototype/models/depth/stereo/raft_stereo.py
```diff
@@ -0,0 +1,704 @@
...
+        if (
+            feature_encoder_layers[:-1] != context_encoder_layers[:-1]
+            or feature_encoder_strides != context_encoder_strides
+        ):
+            raise ValueError(
+                "If shared_encoder_weight is True, then the feature_encoder_layers[:-1]"
+                + " and feature_encoder_strides must be the same with context_encoder_layers[:-1] and context_encoder_strides!"
+            )
+
+        base_encoder = kwargs.pop("base_encoder", None) or BaseEncoder(
```
YosuaMichael, line 529: We enable users to specify instantiated block directly in the model builder, this style follow `torchvision.models.optical_flow.raft` and used during test to modify the `num_levels` and `radius` of the corr_block to prevent nan output (see `test_prototype_models.py` for more detail

torchvision/prototype/models/depth/stereo/raft_stereo.py
```diff
@@ -0,0 +1,704 @@
...
+        flow_head_hidden_size=256,
+        # Mask predictor
+        mask_predictor_hidden_size=256,
+        use_mask_predictor=True,
+        slow_fast=True,
+        **kwargs,
+    )
+
+
+def raft_stereo_basic(*, weights: Optional[Raft_Stereo_Basic_Weights] = None, progress=True, **kwargs) -> RaftStereo:
```
YosuaMichael, line 679: We name the standard model builder as `raft_stereo_basic` instead of the previous name `raft_stereo`. The previous name might conflict with the file_name during import, therefore it is changed.

test/test_prototype_models.py
```diff
@@ -0,0 +1,38 @@
...
+@pytest.mark.parametrize("dev", cpu_and_gpu())
+def test_raft_stereo(model_builder, model_mode, dev):
+    # A simple test to make sure the model can do forward pass and jit scriptable
+    set_rng_seed(0)
+
+    # Use corr_pyramid and corr_block with smaller num_levels and radius to prevent nan output
+    # get the idea from test_models.test_raft
+    corr_pyramid = raft_stereo.CorrPyramid1d(num_levels=2)
+    corr_block = raft_stereo.CorrBlock1d(num_levels=2, radius=2)
+    model = model_builder(corr_pyramid=corr_pyramid, corr_block=corr_block).eval().to(dev)
```
YosuaMichael, line 19: We need to use `corr_pyramid` and `corr_block` with smaller `num_levels` and `radius` to prevent `nan` in the output.

torchvision/prototype/models/depth/stereo/raft_stereo.py
YosuaMichael: I realized this when play around with `get_models_from_module` function on the test and if we add `from .raft_stereo import *` in the `__init__.py` then it cause `torchvision.prototype.models.depth.stereo.raft_stereo` to become a function instead of module, and this is not what we want.

torchvision/prototype/models/depth/stereo/raft_stereo.py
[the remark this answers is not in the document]
YosuaMichael: I add default value so users who want to use this function builder directly need not to specify everything.

YosuaMichael (commented):
I update the test with adding the output comparison with expected file.

We also change the model_builder name from `raft_stereo` to `raft_stereo_basic` to prevent import name conflict with the file name `raft_stereo.py`.

torchvision/prototype/models/depth/stereo/raft_stereo.py
datumbox: FYI Base is more common than basic. It's commonly used in variants like Swin, Convnexts etc.

test/test_prototype_models.py
datumbox: hmmm. This means that you are not testing the original model. This is probably OK to do for now, but also it's the kind of thing we probably want to avoid on the future when we revise out testing strategy.

torchvision/prototype/models/depth/stereo/raft_stereo.py
```diff
@@ -462,32 +484,38 @@ def _raft_stereo(
...
     # Context encoder
-    context_encoder_layers: Tuple[int, int, int, int, int],
-    context_encoder_strides: Tuple[int, int, int, int],
+    context_encoder_layers: Tuple[int, int, int, int, int] = (64, 64, 96, 128, 256),
+    context_encoder_strides: Tuple[int, int, int, int] = (1, 1, 2, 2),
+    # if the `out_with_blocks` param of the context_encoder is True, then
+    # the particular output on that level position will have additional `context_encoder_block` layer
     context_encoder_out_with_blocks: List[bool],
-    context_encoder_block: Callable[..., nn.Module],
+    context_encoder_block: Callable[..., nn.Module] = ResidualBlock,
```
datumbox, line 496: Not sure how this works. You shouldn't be able to have default value for `context_encoder_block` but no default value for `context_encoder_out_with_blocks` which is before that.

Do we even need the default values here? Typically the internal methods don't have default values (see `_resnet` for example) and we place the weights and process at the end.

torchvision/prototype/__init__.py
```diff
@@ -2,4 +2,3 @@
 from . import features
 from . import transforms
 from . import utils
-from . import models
```
datumbox: What's the reason for removing the models from all the `__init__.py` files? This means you won't be able to grab the model builders using some of the utility methods on the tests.

YosuaMichael: the commit that remove this actually just revert, I add it before when testing the utilities to get all modules for instance. But you are right, I think its better to have this import on __init__.py, wil add it back.

YosuaMichael:
The failing test doesn't seems related, will merge.

PR #6108, opened 2022-05-30
[FBcode->GH] Add more logging information for decoder

YosuaMichael:
Differential Revision: D36671123

fbshipit-source-id: eb99f293978ed150bca60073115971586f0fa02a

Cherrypick from fbsync commit: dd445864d0873300daaf29be010c07ed158a24eb


NicolasHug (approved):
Thanks Yosua,
CI is toasted but linting is green so I think it should be OK to merge

PR #6080, opened 2022-05-24
Test 0.13 release branch against torch RC (test) instead of nightly

NicolasHug:
Similar to what was done in previous release: https://github.com/pytorch/vision/pull/4496. In theory this shouldn't conflict with https://github.com/pytorch/vision/pull/6078.

@malfet would you mind reviewing this one please?

PR #6116, opened 2022-06-01
Use -f instead of  --extra-index-url to install torch core in CI

NicolasHug:
Addresses https://github.com/pytorch/vision/issues/6099#issuecomment-1141867661 to avoid future confusion like https://github.com/pytorch/vision/pull/6080.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -146,7 +146,7 @@ commands:
         default: true
     steps:
       - pip_install:
-          args: --pre torch --extra-index-url https://download.pytorch.org/whl/nightly/cpu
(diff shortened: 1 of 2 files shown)


datumbox:
@NicolasHug After reading the [pip doc](https://pip.pypa.io/en/stable/cli/pip_install/), I agree with you that `-f` or `--find-links` is the right option to be used. I still see a few references of the previous idiom in the code-base ([here](https://github.com/pytorch/vision/blob/main/torchvision/prototype/datasets/__init__.py#L6) and [here](https://github.com/pytorch/vision/blob/main/.github/workflows/build-m1-binaries.yml#L31)). Could you update them?

NicolasHug:
Not sure this is still relevant now; closing

PR #6117, opened 2022-06-01
[BE] Unify version computation

malfet:
Instead of hardcoding dev version in various script, use one from
`version.txt` if `setup_build_version` is called without arguments

Also, pass `--pre` option to M1 build/test pip install commands to build
TorchVision against nightly PyTorch and pin torchvision version to a specific PyTorch version by setting `PYTORCH_VERSION` variable


NicolasHug:
@malfet should this be cherry-picked into the release branch?

PR #6122, opened 2022-06-02
[M1] Install "jpeg<=9b" rather than OpenJpeg

malfet:


albanD (approved):
SGTM

datumbox:
@malfet Thanks for the PR. Could you clarify why we require the version to be less than 9b? The old pinning issue was fixed at #4439.

PR #6128, opened 2022-06-03
Closing streams to avoid testing issues

VitalyFedyunin:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #6128

GC workaround is not working inside of test environments as they are keeping references to FH. So we have to close FH manually.


VitalyFedyunin:
Rebased, would be nice to land to clean torchdata's dependency tests.

test/test_prototype_datasets_builtin.py
```diff
@@ -78,23 +79,33 @@ def test_sample(self, dataset_mock, config):
...
         if vanilla_tensors:
             raise AssertionError(
                 f"The values of key(s) "
                 f"{sequence_to_str(sorted(vanilla_tensors), separate_last='and ')} contained vanilla tensors."
             )
 
+        list(iterator)  # Cleanups and closing streams in buffers
+
+        if len(StreamWrapper.session_streams) > 0:
+            raise Exception(StreamWrapper.session_streams)
```
pmeier, line 107: Could you explain what this does? Is `StreamWrapper.session_streams` just a counter for open streams? If yes, why are we only testing this here and not in the other tests? If this is something we should check in general, we can use a decorator like

```py
def check_unclosed_streams(test_fn):
    @functools.wraps(test_fn)
    def wrapper(*args, **kwargs):
        if len(StreamWrapper.session_streams) > 0:
            raise pytest.UsageError("Some previous test didn't clean up")
        
        test_fn(*args, **kwargs)
        
        if len(StreamWrapper.session_streams) > 0:
            raise Assertion("This test didn't clean up")
        
    return wrapper
```

pmeier (commented):
Thanks for the effort @VitalyFedyunin! I left some questions and suggestions inline.

pmeier:
> If we have reached consensus on everything, I can take over and implement it if you want me to.

@VitalyFedyunin In #6647 I redid this PR with all my suggested changes. We can take the discussion there if you want.

pmeier:
Superseded by #6647.

PR #6129, opened 2022-06-03
explain version tagging procedure

PieterBlomme:
I've had to build custom wheels of both torch and torchvision for Jetson.   Since we use the Poetry dependency resolver (rather than requirements.txt of setup.py) I had to ensure that torchvision is linked to the correct pre-compiled version of torch.  

After some digging, I found that there are environment variables to do this, but the process is not well-documented.  I would propose to add this to the README.  Initial proposal below.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -94,6 +94,26 @@ By default, GPU support is built if CUDA is found and ``torch.cuda.is_available(
 It's possible to force building GPU support by setting ``FORCE_CUDA=1`` environment variable,
 which is useful when building a docker image.
 
+Building custom wheels
+======================
+
+When running on custom hardware (eg. Jetson), you may want to build your own wheels.  If so,
+you should make sure that your torchvision wheel has a dependency on the correct custom torch
+version.  
+For eg. the `Poetry dependency resolver <https://python-poetry.org/>`_ this seems to be a 
+requirement to make torch installations work.  
+
+Proper version tagging can be achieved with the following procedure (example here for +cu113 tag)
+
+.. code:: bash
+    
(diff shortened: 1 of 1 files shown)


vfdev-5:
@PieterBlomme thanks for the PR. A question when you are building torchvision from source against a stable pytorch (e.g. 1.11.0) without specifying env vars, do checkout [v0.12.0 tag](https://github.com/pytorch/vision/tree/v0.12.0) ?

PieterBlomme:
Both for torch and torchvision we are checking out the branch, not a tag.  Is it recommended to use the tags?

eg for torch:
`git clone --recursive --branch v1.10.0 http://github.com/pytorch/pytorch`
and torchvision:
`git clone -b release/0.10 https://github.com/pytorch/vision /torchvision`

README.rst
```diff
@@ -94,6 +94,26 @@ By default, GPU support is built if CUDA is found and ``torch.cuda.is_available(
 It's possible to force building GPU support by setting ``FORCE_CUDA=1`` environment variable,
 which is useful when building a docker image.
 
+Building custom wheels
```
NicolasHug, line 97: The info below is very nice, and is only related to building wheels and packaging. So a more appropriate place would probably be `packaging/README.md`.

NicolasHug (commented):
Thanks for the PR @PieterBlomme,

This is documenting env vars that are being used in our packaging scripts. These are maintained by our releng team. @malfet @atalman @seemethere  could you please advise whether the instructions below are reasonable, and whether we should expect these variables to continue working in the future?

Issue #6126, opened 2022-06-02
Add title to the tables in the `Models and Pretrained Weights` section of torchvision documentation.


abhi-glitchhg:
### 🚀 The feature

Currently, there is no title to the tables in the `Models and Pretrained Weights` section.  

[click here](https://output.circle-artifacts.com/output/job/a621d2f7-44c0-4929-af91-e82b10c81671/artifacts/0/docs/generated/detection_table.html?highlight=fcos_resnet50) to open docs .

![image](https://user-images.githubusercontent.com/72816663/171637097-380e04fc-0845-498d-b039-31957b086b7c.png)

As you can see `Docs <no title>`, doesn't look nice. Also, adding the titles will also help users in searching the tables and other details effectively.

same goes for [classification models table](https://output.circle-artifacts.com/output/job/08b9abf5-3ad4-4093-802c-0d84a238dce2/artifacts/0/docs/generated/classification_table.html?highlight=regnet_x_400mf_weights%20imagenet1k_v2), [segmentation models table](https://output.circle-artifacts.com/output/job/08b9abf5-3ad4-4093-802c-0d84a238dce2/artifacts/0/docs/generated/segmentation_table.html?highlight=deeplabv3_resnet101_weights%20coco_with_voc_labels_v1), [quantized model table](https://output.circle-artifacts.com/output/job/08b9abf5-3ad4-4093-802c-0d84a238dce2/artifacts/0/docs/generated/classification_quant_table.html?highlight=googlenet_quantizedweights%20imagenet1k_fbgemm_v1) :)

### Motivation, pitch

Enhance the searching experience for the users?

PR #6131, opened 2022-06-06
Add names to weight tables

NicolasHug:
Attempt at https://github.com/pytorch/vision/issues/6126

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -400,7 +400,7 @@ def generate_weights_table(module, table_name, metrics, dataset, include_pattern
     generated_dir.mkdir(exist_ok=True)
     with open(generated_dir / f"{table_name}_table.rst", "w+") as table_file:
         table_file.write(".. rst-class:: table-weights\n")  # Custom CSS class, see custom_torchvision.css
-        table_file.write(".. table::\n")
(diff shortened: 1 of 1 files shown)


datumbox:
The problem [seems](https://output.circle-artifacts.com/output/job/7b6b4830-2038-441f-92ba-1c9f6be1e04c/artifacts/0/docs/generated/detection_table.html) to persist :(

oke-aditya:
Any idea @abhi-glitchhg ?

abhi-glitchhg:
Not an expert in this domain, but I would like to drop some thoughts; I might be wrong here.

I feel that we should add the subtitles according to the [rst file format](https://thomas-cokelaer.info/tutorials/sphinx/rest_syntax.html#headings). That might help in this case.

So that change should be like 

`               table_file.write(f".. table:: {table_name}\n")`
`                table_file.write(f".. table:: {len(table_name)*'#'}")`

And above lines should be written at the start of the rst file before writing the tables into the rst file.
What do you think? Does this approach make sense? @oke-aditya, @datumbox  lemme know your thoughts , :relaxed: 



I couldn't verify this on my local machine as weight tables are not generated with `make html-noplot` :(

PR #6134, opened 2022-06-07
Try making ops folder a package

oke-aditya:
Trying to solve #6020

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -3,7 +3,8 @@ import random
 import numpy as np
 import pytest
 import torch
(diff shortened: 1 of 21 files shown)


oke-aditya:
As per latest pytest docs https://docs.pytest.org/en/latest/explanation/pythonpath.html


importlib: new in pytest-6.0, this mode uses [importlib](https://docs.python.org/3/library/importlib.html#module-importlib) to import test modules. This gives full control over the import process, and doesn’t require changing [sys.path](https://docs.python.org/3/library/sys.html#sys.path).

For this reason this doesn’t require test module names to be unique.

One drawback however is that test modules are non-importable by each other. Also, utility modules in the tests directories are not automatically importable because the tests directory is no longer added to [sys.path](https://docs.python.org/3/library/sys.html#sys.path).

Initially we intended to make importlib the default in future releases, however it is clear now that it has its own set of drawbacks so the default will remain prepend for the foreseeable future.

PR #6135, opened 2022-06-08
GH: Add M1 conda build step

malfet:


.github/workflows/build-m1-binaries.yml
```diff
@@ -71,3 +70,52 @@ jobs:
...
+        run: |
+          . ~/miniconda3/etc/profile.d/conda.sh
+          set -ex
+          . packaging/pkg_helpers.bash
+          setup_build_version
+          setup_conda_pytorch_constraint
+          export SOURCE_ROOT_DIR=$(pwd)
+          conda build -c defaults $CONDA_CHANNEL_FLAGS --no-anaconda-upload --python "$PYTHON_VERSION" packaging/torchvision
+          mkdir -p dist
+          cp ~/miniconda3/conda-bld/osx-arm64/*.tar.bz2 dist/
```
datumbox, line 105: Can we test the conda package by running the smoke tests as we do above with the wheels?

malfet: @datumbox isn't it part of coda build process? I.e. it should execute commands defined in `test` section, i.e.:
https://github.com/pytorch/vision/blob/d4a03fc02d0566ec97341046de58160370a35bd2/packaging/torchvision/meta.yaml#L53-L63

malfet:
@datumbox I'm going to merge this one as is and then perhaps at testing when moving both x86 and m1 binary builds to GHA infrastructure.... (as neither has testing at the moment)

.github/workflows/build-m1-binaries.yml
NicolasHug: It will just try to import `torchvision`, `torchvision.datasets` and `torchvision.transforms`?
We should at least check that ops are properly present like we do for the wheels

```
          conda run --cwd /tmp -p ${ENV_NAME} python3 -c "import torchvision;print(torchvision.io.read_image('${PWD}/gallery/assets/dog1.jpg').shape)"
```

PR #6146, opened 2022-06-09
Fix `if` condition for s3/conda uploads 

malfet:
Replace `steps.extract_branch.outputs.branch` (which were probably taken from
    https://stackoverflow.com/questions/58033366/how-to-get-the-current-branch-within-github-actions
    ) with straightforward `github.event.ref`

Before this change nightly M1 builds were not actually uploaded as once can observe from https://github.com/pytorch/vision/runs/6811614195?check_suite_focus=true
Tested in  https://github.com/malfet/deleteme/runs/6822647720?check_suite_focus=true and https://github.com/malfet/deleteme/runs/6822691158?check_suite_focus=true


atalman (approved):
LGTM

PR #6150, opened 2022-06-10
Fix ViT and Resnext docs

NicolasHug:
Minor fixes. Some links weren't clickable. Should be cherry-picked into release


datumbox (approved):
LGTM, good catch!

Issue #6142, opened 2022-06-09
M1 unit-test fail with `not yet supported with named tensors`


datumbox:
### 🐛 Describe the bug

The M1 unit-tests [fail](https://github.com/pytorch/vision/runs/6806004576?check_suite_focus=true) with the following error:
```
self = <OpOverloadPacket(op='torchvision.roi_align')>
args = (tensor([[[[ 8.2470e+00,  1.3783e+01,  1.3401e+01,  ...,  2.4769e+00,
            2.6093e+00,  6.0664e+00],
          ...68.9164],
        [  0.0000, 175.5915, 200.8642, 382.0600, 212.2028]],
       dtype=torch.float64), 0.25, 7, 7, 2, ...)
kwargs = {}
    def __call__(self, *args, **kwargs):
        # overloading __call__ to ensure torch.ops.foo.bar()
        # is still callable from JIT
        # We save the function ptr as the `op` attribute on
        # OpOverloadPacket to access it here.
>       return self._op(*args, **kwargs or {})
E       RuntimeError: torchvision::roi_align is not yet supported with named tensors. Please drop names via `tensor = tensor.rename(None)`, call the op with an unnamed tensor, and set names on the result of the operation.
```

The last time the tests run successfully was on the 6th of June at commit d4a03fc02d0566ec97341046de58160370a35bd2. Unfortunately the days after that our CI broke for unrelated issues, so I guess it could be a commit commits from Core between the 6th and 9th of June.

### Versions

Latest main branch.

vfdev-5:
I bit more coherent info on failing/passing tests according to the latest job from #6145 : 

https://github.com/pytorch/vision/runs/6815037410?check_suite_focus=true

TestRoiPool::test_forward is passing until dev20220520 and since May 20 the error is
```python
NotImplementedError: Could not run 'torchvision::roi_pool' with arguments from the 'MkldnnCPU' backend. 
```
and since dev20220601 it started failing with 
```
RuntimeError: torchvision::roi_pool is not yet supported with named tensors
```

albanD:
@vfdev-5 could you set `TORCH_SHOW_CPP_STACKTRACES=1` for each of these to get the full c++ stacktrace?

malfet:
Looks like something is wrong with how we setup cross-compilation: if both PyTorch and TorchVision are build on MacOS-12.4 machine, they work fine together, but downloading cross-compiled nightly package and building vision locally reproduces the error

malfet:
[...]

PR #6152, opened 2022-06-10
Upgrade clang version for M1 test

albanD:
See if a more recent clang helps

EDIT: looks like it does.
Note that this is NOT changing the build job or any binary building job!

Seems to fix https://github.com/pytorch/vision/issues/6142

### DIFF
--- a/.github/workflows/test-m1.yml
+++ b/.github/workflows/test-m1.yml
@@ -30,6 +30,7 @@ jobs:
           export PATH=~/miniconda3/bin:$PATH
           set -ex
           conda create -yp ${ENV_NAME} python=${PY_VERS} numpy libpng jpeg scipy
+          conda install -y -c conda-forge clang==14.0.4
           conda run -p ${ENV_NAME} python3 -mpip install --pre torch --extra-index-url=https://download.pytorch.org/whl/nightly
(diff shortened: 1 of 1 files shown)


malfet (approved):

atalman (approved):

vfdev-5:
@albanD if we do not install `clang==14.0.4` what does it use by default to compile vision ? Maybe we should trace that in `setup.py` and ask users to install latest clang

albanD:
It uses 13.0.0 from what I was discussing with @malfet 

torch is built with `13.1.6`. And if vision is built with `13.1.6`, it also fixes the problem. Note that the clang version itself might not be the core of the issue and the "fix" might come from the fact that the toolchain that comes with the newer clang version avoids the issue.

I went with `14.0.4` here as conda-forge only has `13.0.X` versions. So I had to go to `14.0` to get `13.1+`.

albanD:
So the problem is not that "clang is too old" but some misterious bad interaction between 13.1 (used to build torch) and 13.0 (used to build vision)

albanD:
Canceling this as the conda-forge dependency is not great. @malfet will update the runners directly with 13.1.6+

PR #6154, opened 2022-06-10
Update _pil_constants.py

### LINKED ISSUE #6153: PIL version check for enum change appears to break SIMD versions
### 🐛 Describe the bug

This change appears to break current Pillow-SIMD version #5898 

```
    if tuple(int(part) for part in PIL.__version__.split(".")) >= (9, 1):
  File "/home/.../lib/python3.10/site-packages/torchvision/transforms/_pil_constants.py", line 7, in <genexpr>
    if tuple(int(part) for part in PIL.__version__.split(".")) >= (9, 1):
ValueError: invalid literal for int() with base 10: 'post1'
```

Amusingly enough, I warned against this approach in a users PR in `timm` https://github.com/rwightman/pytorch-image-models/pull/1256

Would be nice to have it fixed before 1.12 is finalized, I just hit this trying out the RC

### Versions

PT 1.12 RC, TV 0.13.0

vfdev-5:
Fixes #6153

### DIFF
--- a/torchvision/transforms/_pil_constants.py
+++ b/torchvision/transforms/_pil_constants.py
@@ -1,4 +1,3 @@
-import PIL
 from PIL import Image
 
 # See https://pillow.readthedocs.io/en/stable/releasenotes/9.1.0.html#deprecations
--- a/torchvision/transforms/_pil_constants.py
+++ b/torchvision/transforms/_pil_constants.py
@@ -4,7 +4,7 @@ from PIL import Image
 # See https://pillow.readthedocs.io/en/stable/releasenotes/9.1.0.html#deprecations
 # TODO: Remove this file once PIL minimal version is >= 9.1
 
-if tuple(int(part) if part.isdigit() else -1 for part in version.split(".")[:2]) >= (9, 1):
+if "Resampling" in dir(Image):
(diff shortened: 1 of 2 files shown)


torchvision/transforms/_pil_constants.py
```diff
@@ -1,10 +1,9 @@
-import PIL
 from PIL import Image
 
 # See https://pillow.readthedocs.io/en/stable/releasenotes/9.1.0.html#deprecations
 # TODO: Remove this file once PIL minimal version is >= 9.1
 
-if tuple(int(part) for part in PIL.__version__.split(".")) >= (9, 1):
+if "Resampling" in dir(Image):
```
datumbox, line 6: I think the following idiom is cleaner:
```suggestion
if hasattr(Image, "Resampling"):
```

datumbox (commented):
Looks good overall, just a minor nit.

cc @NicolasHug, @YosuaMichael this would need to be cherrypicked into the release branch.

torchvision/transforms/_pil_constants.py
vfdev-5: I agree, thanks !

PR #6112, opened 2022-05-31
[Cherrypick 0.13] Add M1 build and test jobs

YosuaMichael:
Cherrypick from main: #5948 and #6110

This change will add a file `.github/workflows/build-m1-binaries.yml` which build m1 on the CI.

EDIT: this more generally picks all M1-related stuff, including https://github.com/pytorch/vision/pull/6122  and https://github.com/pytorch/vision/pull/6132 (and a lot more stuff tracked in https://github.com/pytorch/vision/issues/6081#issuecomment-1141874556)

PR #6158, opened 2022-06-13
Fix typo in GHA nightly build condition

malfet:
s#ref/heads/#refs/heads/#

I should have noticed it while copy-n-pasting the condition. Unfortunately there are no way to test is other than in prod, but nightly builds are still not getting pushed, see https://github.com/pytorch/vision/runs/6860407007?check_suite_focus=true for example


datumbox (approved):
LGTM. 🤞 

Needs to be included at #6112

PR #6173, opened 2022-06-15
[WIP] Add remove-able warnings for Beta modules or functions

NicolasHug:
Some POC for https://github.com/pytorch/pytorch/issues/72317

```py
#######################################################
# For entire module like torchvision.models.detection #
#######################################################

# No warning - anything before torchvision.models.detection
import torchvision
import torchvision.models
from torchvision import models
from torchvision.models import resnet50
from torchvision.models import segmentation

# Any of these will warn:
from torchvision.models import detection
from torchvision.models.detection import ssd
from torchvision.models.detection.ssd import ssd300_vgg16

torchvision.models.enable_beta_detection()  # Now all warnings are disabled
from torchvision.models import detection
from torchvision.models.detection import ssd
from torchvision.models.detection.ssd import ssd300_vgg16


####################
# For single class #
####################

from contextlib import suppress
from torchvision.io import VideoReader
with suppress(RuntimeError):  # VideoReader raises a RuntimeError, but that's unrelated
    VideoReader(path="path")  # raises warning

with suppress(RuntimeError):
    torchvision.io.enable_beta_video_api()
    VideoReader(path="path")  # no warning
```

### DIFF
--- a/torchvision/models/__init__.py
+++ b/torchvision/models/__init__.py
@@ -1,3 +1,5 @@
+import importlib
+
 from .alexnet import *
 from .convnext import *
 from .densenet import *
@@ -18,6 +20,14 @@ from . import quantization
 from . import segmentation
 from . import video
 from ._api import get_weight
+
+
+def __getattr__(name):
+    if name == "detection":
+        return importlib.import_module("." + name, __name__)
+    else:
+        raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
+
 
 _BETA_DETECTION_IS_ENABLED = False
 
--- a/torchvision/models/__init__.py
+++ b/torchvision/models/__init__.py
@@ -13,9 +13,15 @@ from .squeezenet import *
 from .vgg import *
 from .vision_transformer import *
 from .swin_transformer import *
-from . import detection
 from . import optical_flow
 from . import quantization
 from . import segmentation
 from . import video
 from ._api import get_weight
+
+_BETA_DETECTION_IS_ENABLED = False
+
+
+def enable_beta_detection():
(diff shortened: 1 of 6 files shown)


torchvision/models/__init__.py
```diff
@@ -13,9 +15,23 @@
...
 from .swin_transformer import *
-from . import detection
 from . import optical_flow
 from . import quantization
 from . import segmentation
 from . import video
 from ._api import get_weight
+
+
+def __getattr__(name):
```
NicolasHug, line 25: This adds the `detection` attribute lazily. If we imported it like the others (e.g. `quantization` above), users would get the "beta detection warning" even when just importing `torchvision.models`. See more details about `__getattr__` at https://peps.python.org/pep-0562/

Note: the detection attribute is needed when doing e.g.
```py
import torchvision.models as M
M.detection  # <- directly accessing the attribute
```

but in a lot of patterns this gets bypassed, typically when doing

```py
from torchvision.models.detection import ssd
```
or even just

```py
import torchvision.models.detection
```

torchvision/models/detection/__init__.py
```diff
@@ -5,3 +5,15 @@
 from .retinanet import *
 from .ssd import *
 from .ssdlite import *
+
+import warnings
+
+from .. import _BETA_DETECTION_IS_ENABLED
+
+if not _BETA_DETECTION_IS_ENABLED:
```
NicolasHug, line 13: Here we mark the entire `torchvision.models.detection` as Beta and accessing anything within its namespace will throw a warning unless `torchvision.models.enable_beta_detection()` is called.

PR #6176, opened 2022-06-16
Fix all broken URLs in docs

NicolasHug:
`make linkcheck` allows to see which URLs are currently broken. There's a lot of false positives but these ones came up as legit.


datumbox (approved):
LGTM. Thanks for looking into it @NicolasHug.

datumbox:
BTW, this is definitely worth cherrypicking but hope it won't create issues to the RelEng team. Possibly not because it's on docs?

PR #6181, opened 2022-06-21
Fix copypaste collate pickle issues

datumbox:
Running the current reference script on cpu throws:
```
AttributeError: Can't pickle local object 'main.<locals>.copypaste_collate_fn'
```

This is because the method is nested within another. Moving it on the top-level fixes the issue.


NicolasHug (approved):
LGTM

vfdev-5 (commented):
LGTM as well :)

PR #6187, opened 2022-06-22
[proto] Added base tests for rotate_image_tensor

vfdev-5:
Description:
- Added base tests for `rotate_image_tensor`
- switched the order of kwargs in `rotate_image_tensor` signature


torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -321,8 +330,8 @@ def rotate_image_tensor(
     angle: float,
     interpolation: InterpolationMode = InterpolationMode.NEAREST,
     expand: bool = False,
-    fill: Optional[List[float]] = None,
     center: Optional[List[float]] = None,
+    fill: Optional[List[float]] = None,
```
datumbox, line 334: On the same file, fill is always before center. Why change it here?

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -42,10 +47,9 @@ def resize_image_tensor(
     max_size: Optional[int] = None,
     antialias: Optional[bool] = None,
 ) -> torch.Tensor:
-    # TODO: use _compute_output_size to enable max_size option
-    max_size  # ununsed right now
-    new_height, new_width = size
     num_channels, old_height, old_width = get_dimensions_image_tensor(image)
+    size = _compute_output_size((old_height, old_width), size=size, max_size=max_size)
+    new_height, new_width = size
```
datumbox, line 52: Nit: why not directly assign it to the vars?

vfdev-5: Basically, size is reused below in `_FT.resize(..., size=size, ...)` but yes, we can remove one line and use `[new_height, new_width]` for `_FT.resize`. Thanks!

torchvision/prototype/transforms/functional/_geometry.py
vfdev-5: I see, thanks for pointing out. I was trying to align to v1 `F.rotate` where it is `center` -> `fill`. But, looks like `F.affine` has `fill` -> `center`...

datumbox (approved):
LGTM, thanks!

PR #6191, opened 2022-06-22
Refactored and modified private api for resize functional op

vfdev-5:
Description:
- Refactored output size computation into a single method `_compute_output_size` and removed `max_size` arg from `F_t.resize` and `F_pil.resize`.
  - We can also reuse `_compute_output_size` in proto for bboxes (right now `max_size` is not taken into account)
- Minor changes and cosmetics in tests
- Temporarily I also have to modify prototype functional implementation `resize_image_*`. This will be fixed in https://github.com/pytorch/vision/pull/6187


test/test_transforms_tensor.py
```diff
@@ -412,19 +410,15 @@ def test_resize_scripted(self, dt, size, max_size, interpolation, device):
             # This is a trivial cast to float of uint8 data to test all cases
             tensor = tensor.to(dt)
         if max_size is not None and len(size) != 1:
-            pytest.xfail("with max_size, size must be a sequence with 2 elements")
+            pytest.skip("Size should be an int or a sequence of length 1 if max_size is specified")
```
datumbox, line 413: Question: Why skip instead of xfail?

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -42,14 +42,15 @@ def resize_image_tensor(
     max_size: Optional[int] = None,
     antialias: Optional[bool] = None,
 ) -> torch.Tensor:
+    # TODO: use _compute_output_size to enable max_size option
+    max_size  # ununsed right now
```
datumbox, line 46: It's unclear to me why we remove `max_size` from here. Shouldn't the TODO be address in this PR?

datumbox (commented):
Thanks for the PR @vfdev-5. I have a few questions for you below. Let me know what you think.

test/test_transforms_tensor.py
vfdev-5: That's a good question :)

According to https://docs.pytest.org/en/stable/how-to/skipping.html
We skip in case windows tests on non-windows platform and xfail if we expect a test to fail for some reason.

In our case, the configuration "max_size is not None and len(size) != 1" can not be tested as we explicitly raise and error. Proper solution to that is to catch the error and check the message text with `pytest.raises`.

I felt like skipping is more appropriate than xfail. But it is a matter of taste if you think it is better to revert to xfail I can do that and will only fix the message which is incorrect

torchvision/prototype/transforms/functional/_geometry.py
vfdev-5: I'll do that in another PR with appropriate tests. Just splitting into 2 PRs and make CI pass.

datumbox (approved):
LGTM, our offline discussion covers all my points. I'll let @vfdev-5 summarize what we discussed.

test/test_transforms_tensor.py
datumbox: I was expecting a pytest.raises where if I'm honest but no strong opinions.

torchvision/transforms/functional.py
```diff
@@ -423,13 +446,32 @@ def resize(
...
+            )
+
+    _, image_height, image_width = get_dimensions(img)
+    if isinstance(size, int):
+        size = [size]
+    output_size = _compute_output_size((image_height, image_width), size, max_size)
+
+    if (image_height, image_width) == output_size:
+        return img
+
```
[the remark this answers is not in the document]
vfdev-5: Sorry, description may not be obvious. The point is the following:
- we refactor duplicated code into a method `_compute_output_size`
- we remove `max_size` arg from private functional modules: `F_t` and `F_pil` where we can do what we want
- For v1 API no breakage as we keep public functional `resize` method without any change in the signature.
- For v2 API, we will implement things in the following way:
  - For image tensor, we reuse `_compute_output_size` and call as previously `F_t.resize`
```python
def resize_image_tensor(
    image: torch.Tensor,
    size: List[int],
    interpolation: InterpolationMode = InterpolationMode.BILINEAR,
    antialias: Optional[bool] = None,
) -> torch.Tensor:
    num_channels, old_height, old_width = get_dimensions_image_tensor(image)
    size = _compute_output_size((old_height, old_width), size=size, max_size=max_size)
    new_height, new_width = size
    batch_shape = image.shape[:-3]
    return _FT.resize(
        image.reshape((-1, num_channels, old_height, old_width)),
        size=size,
        interpolation=interpolation.value,
        antialias=antialias,
    ).reshape(batch_shape + (num_channels, new_height, new_width))
```
  
  - For pillow images, we do the same thing: we reuse `_compute_output_size` and call as previously `F_pil.resize`
  
```python
def resize_image_pil(
    img: PIL.Image.Image,
    size: Union[Sequence[int], int],
    interpolation: InterpolationMode = InterpolationMode.BILINEAR,
) -> PIL.Image.Image:
    if isinstance(size, int):
        size = [size, size]
    # Explicitly cast size to list otherwise mypy issue: incompatible type "Sequence[int]"; expected "List[int]"
    size: List[int] = list(size)
    size = _compute_output_size(img.size[::-1], size=size, max_size=max_size)
    return _FP.resize(img, size, interpolation=pil_modes_mapping[interpolation])
```

  - For bounding boxes we can now implement `max_size` option (which is not the case today):
```python
def resize_bounding_box(
    bounding_box: torch.Tensor, size: List[int], image_size: Tuple[int, int], max_size: Optional[int] = None
) -> torch.Tensor:
    old_height, old_width = image_size
    size = _compute_output_size(image_size, size=size, max_size=max_size)
    new_height, new_width = size
    ratios = torch.tensor((new_width / old_width, new_height / old_height), device=bounding_box.device)
    return bounding_box.view(-1, 2, 2).mul(ratios).view(bounding_box.shape)
```

Source: https://github.com/pytorch/vision/blob/f0c896ff1391dcac098539db79f14e2c50549d7a/torchvision/prototype/transforms/functional/_geometry.py

NicolasHug:
@vfdev-5 @datumbox 

@vfdev-5 pointed me to this PR from our discussions in https://github.com/pytorch/vision/pull/6209#discussion_r908294878

As discussed over there, `functional_pil` and `functional_tensor` are technically public, unfortunately. I understand some of these changes are useful for the prototype area (where breakages are OK), but would it be possible to preserve BC here? It looks like the removal of the parameters aren't strictly necessary.

vfdev-5:
> For example, perhaps it's possible to call _compute_output_size within functional_pil.resize and functional_tensor.resize instead of calling it in functional.resize?

It's possible if we introduce additional `_utils.py` as a common place for `functional_pil.py` and `functional_tensor.py`

NicolasHug:
I guess we could just define it in `functional_tensor` and have `functional_pil` rely on it?

datumbox:
@NicolasHug as far as I know those APIs are private and this change has been done many versions ago. I don't think we should worry about BC here.

datumbox:
Sorry I was referring to when the API became private not to this PR. Let's continue our discussion at https://github.com/pytorch/vision/pull/6209#discussion_r908474578 to avoid further confusion

Issue #6189, opened 2022-06-22
CI fails on windows: ci/circleci: unittest_windows_cpu_pyX.Y


vfdev-5:
Tests on windows are started failing:

- https://app.circleci.com/pipelines/github/pytorch/vision/18503/workflows/3c477207-67b9-45c8-8960-ab3efb40b8b1/jobs/1497031
- https://app.circleci.com/pipelines/github/pytorch/vision/18503/workflows/3c477207-67b9-45c8-8960-ab3efb40b8b1/jobs/1496973

```
test/test_models.py::test_classification_model[cpu-regnet_y_16gf] PASSED [ 79%]
test/test_models.py::test_classification_model[cpu-regnet_y_32gf] PASSED [ 79%]
test/test_models.py::test_classification_model[cpu-regnet_y_128gf] 

Exited with code exit status 127

CircleCI received exit code 127
```

It started appearing on PyTorch core nightly 20220622

cc @pmeier @seemethere

YosuaMichael:
cc @atalman

YosuaMichael:
Seems like it is green now https://github.com/pytorch/vision/pull/5009
should we close this issues?
@vfdev-5

vfdev-5:
Thanks for the update! Let's close this issue if everything is OK now

vfdev-5:
@YosuaMichael actually, tests are still failing on https://github.com/pytorch/vision/pull/5009. I reopen

YosuaMichael:
> @YosuaMichael actually, tests are still failing on #5009. I reopen

Ah yeah, previously I just rerun the test and it seems green. But it get the error after I update the branch. Sorry for the false negative @vfdev-5 !

datumbox:
@vfdev-5 the failure is suspicious because it's on a very large model. Can you try skipping the specific test to see if this is related to issues on CircleCI side rather than on core? Another thing we can do to confirm that the core is not the issue, is to fix the nightly the the one before and rerun the job. If it fails we will know it's the CircleCI.

YosuaMichael:
I have confirmed that skipping the big models indeed make the CI green again. Now with the same PR #6195  I use the older nightly version (torch==1.13.0.dev20220621) to check if this issue is caused by core or circle CI.

YosuaMichael:
[...]

PR #6197, opened 2022-06-23
Skip big models on cpu

YosuaMichael:
Addressing issue https://github.com/pytorch/vision/issues/6189

It seems like the CI is broken in windows machine cpu due to the big model.
I have sorted all weights base on the number of parameters and here is what I got:
image

In this PR we will use the existing list of `skipped_big_models` but we also skip in cpu instead of just cuda.

### DIFF
--- a/test/test_models.py
+++ b/test/test_models.py
@@ -603,7 +603,7 @@ def test_classification_model(model_fn, dev):
         "input_shape": (1, 3, 224, 224),
     }
     model_name = model_fn.__name__
-    if dev == "cuda" and SKIP_BIG_MODEL and model_name in skipped_big_models:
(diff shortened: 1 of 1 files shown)


YosuaMichael:
> Since the default for `SKIP_BIG_MODEL` is `1`, does that mean we'll never actually test these models anymore?
> 
> Also side note, probably not too important: the memory footprint of a model is related to the size of its weights but there are other factors like the size and number of the feature maps in conv layers. EDIT: maybe it's not really the case in `eval()` mode tho

Yeah, I think for now we disable the test for the big model until we can find a better way (I plan to revamp test CI on H2 so this will definitely a consideration on how to enable testing on big model)

PR #6199, opened 2022-06-23
Add flag to ignore ffmpeg in python build.

vcarpani:
This PR adds a flag to disable building the parts of the code that depend on ffmpeg in the python build

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -356,6 +356,8 @@ def get_extensions():
             print("Error fetching ffmpeg version, ignoring ffmpeg.")
             has_ffmpeg = False
 
+    use_ffmpeg = os.getenv("USE_FFMPEG", "1") == "1"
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -356,6 +356,8 @@ def get_extensions():
             print("Error fetching ffmpeg version, ignoring ffmpeg.")
             has_ffmpeg = False
 
+    use_ffmpeg = os.getenv("USE_FFMPEG", "1") == "1"
```
NicolasHug, line 359:
```suggestion
    use_ffmpeg = os.getenv("TORCHVISION_USE_FFMPEG", "1") == "1"
```

NicolasHug (commented):
Thanks @vcarpani , I think this is reasonable. Just a small suggestion below to avoid potential collisions with other projects

NicolasHug (approved):
Thanks @vcarpani

PR #6201, opened 2022-06-24
Refactor setup.py to build only wanted extensions.

vcarpani:
This PR makes it possible to include or exclude different extensions in the python setups.
It also adds a brief overview of what is enabled by command line that should look like this:
```
Compiling extensions with following flags:
WITH_CPP_MODELS_TEST: False
FORCE_CUDA: False
DEBUG: False
TORCHVISION_USE_FFMPEG: True
TORCHVISION_USE_PNG: True
TORCHVISION_USE_JPEG: True
TORCHVISION_USE_NVJPEG: True
TORCHVISION_USE_FFMPEG: True
TORCHVISION_USE_VIDEO_CODEC: True
NVCC_FLAGS: 
```
Do you believe it makes sense to add a similar report before building, to explain to the user which extensions are actually being built?


datumbox:
@vcarpani Thanks for the PR!

I think it's an improvement as it summarizes the variables for the users and improves the overall output. Perhaps we could review the messaging to ensure that it's clear to the user that `TORCHVISION_USE_PNG==True` doesn't mean that actually PNG is used. This is the only thing I feel can confuse people.

I saw you changed the names of the vars from `has_*` to `use_*`. I think that makes sense given the nature of the change.

@NicolasHug Any thoughts on the above?

setup.py
```diff
@@ -139,6 +139,29 @@ def get_extensions():
...
+    use_ffmpeg = os.getenv("TORCHVISION_USE_FFMPEG", "1") == "1"
+    print(f"TORCHVISION_USE_FFMPEG: {use_ffmpeg}")
+    use_png = os.getenv("TORCHVISION_USE_PNG", "1") == "1"
+    print(f"TORCHVISION_USE_PNG: {use_png}")
+    use_jpeg = os.getenv("TORCHVISION_USE_JPEG", "1") == "1"
+    print(f"TORCHVISION_USE_JPEG: {use_jpeg}")
+    use_nvjpeg = os.getenv("TORCHVISION_USE_NVJPEG", "1") == "1"
+    print(f"TORCHVISION_USE_NVJPEG: {use_nvjpeg}")
+    use_ffmpeg = os.getenv("TORCHVISION_USE_FFMPEG", "1") == "1"
+    print(f"TORCHVISION_USE_FFMPEG: {use_ffmpeg}")
```
NicolasHug, line 158: Looks like the ffmpeg bit this is duplicated

NicolasHug (commented):
Thanks for the PR @vcarpani 

> Perhaps we could review the messaging to ensure that it's clear to the user that TORCHVISION_USE_PNG==True doesn't mean that actually PNG is used

I agree - perhaps we should print all the png-related stuff at the same place (same for ffmpeg, nvjpeg, etc.). It could look like this:

```
<Flat specificed by user> = XYZ
Found <lib> = XYZ
Will [not] build with XYZ support
``` 

?

vcarpani:
Changed a bit the format, here is what it should look like:

png_disabled


png_enabled_found

vcarpani:
Ok, shall I keep the same structure without the colors?

setup.py
```diff
@@ -280,30 +286,35 @@ def get_extensions():
     libpng = shutil.which("libpng-config")
     pngfix = shutil.which("pngfix")
     png_found = libpng is not None or pngfix is not None
-    print(f"PNG found: {png_found}")
+    if not use_png and png_found:
+        print(f"{colors.WARNING}Found PNG library, but PNG extension is disabled by compile flag{colors.ENDC}")
+    if use_png and not png_found:
+        print(f"{colors.WARNING}Could not find PNG library{colors.ENDC}")
```
datumbox, line 292: Nit: The logic here is correct but complex. I would prefer to just put a message on the else of `use_png` check instead.

setup.py
```diff
@@ -280,30 +286,35 @@ def get_extensions():
...
                 _, png_include = png_include.split("-I")
-                print(f"libpng include path: {png_include}")
                 image_include += [png_include]
                 image_link_flags.append("png")
+                print(f"  libpng version: {png_version}")
+                print(f"  libpng include path: {png_include}")
             else:
-                print("libpng installed version is less than 1.6.0, disabling PNG support")
+                print(f"{colors.FAIL}Could not add PNG image support to torchvision:{colors.ENDC}")
+                print(f"{colors.FAIL}  libpng minimum version 1.6.0, found {png_version}{colors.ENDC}")
```
datumbox, line 317: Nit: Avoid hardcoding the min version?

datumbox (commented):
Overall it looks good to me if we remove the colours. Just 2 minor nits for your attention but nothing major from my side.

setup.py
```diff
@@ -340,6 +367,7 @@ def get_extensions():
             )
         )
 
+    # Locating ffmpeg
     ffmpeg_exe = shutil.which("ffmpeg")
     has_ffmpeg = ffmpeg_exe is not None
```
datumbox, line 372: Nit: ensure that the `ffmpeg_version` is initialized
```suggestion
    # Locating ffmpeg
    ffmpeg_exe = shutil.which("ffmpeg")
    has_ffmpeg = ffmpeg_exe is not None
    ffmpeg_version = None
```

datumbox (commented):
Just a few final nits on my side. I will possibly commit these changes to avoid back and worth.

datumbox:
@vcarpani I'm unable to push to the branch the last changes and also update the main branch because you haven't clicked the "[Allow edits from maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork)" option. Could you give access or alternatively you will need to make the last changes on your side and make sure your branch is up-to-date with main branch or else we won't be able to click the merge button on our side due to branch protection settings. Thanks!

vcarpani:
> @vcarpani I'm unable to push to the branch the last changes and also update the main branch because you haven't clicked the "[Allow edits from maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork)" option. Could you give access? Alternatively you will need to make the last changes on your side and make sure your branch is up-to-date with main branch. Without this we won't be able to click the merge button on our side due to branch protection settings. Thanks!

I unfortunately cannot: https://github.com/github-community/community/discussions/5634?sort=top

datumbox:
@vcarpani No worries. The changes look good. You need to update the PR and merge the latest main branch. Right now the CI doesn't let me merge.
image

vcarpani:
Hello, sorry to bother on this, but any news on merging this PR?

datumbox:
@vcarpani Literally Github doesn't let me merge it because the branch is out-of-date. Please merge one more time with main so we can do it. The problem is we need to synchronize your update because everytime someone merges something new, we need to remerge.

vcarpani:
Sorry, next time I will use a private fork ;)

vcarpani:
> LGTM, thanks @vcarpani!
> 
> Yes please stick around, we would love to improve the situation on our building scripts. It might be worth indeed to send the PR from a different account/setup so that we can merge things easier.
> 
> To unblock others, I'll merge now. All relevant tests passed. I'll keep an eye on the tests post-merge and if something broke I'll revert and ping you.

Yes, I am also looking at the CPP part now, since we are using torchvision extensively at the company where I work, but we do not need all the modules built

datumbox:
@vcarpani Sounds good. I strongly recommend opening an RFC before writing any code because messing with the building scripts to exclude C++ modules is something that requires discussion as it can affect internal systems at Meta.

vcarpani:
> @vcarpani Sounds good. I strongly recommend opening an RFC before writing any code because messing with the building scripts to exclude C++ modules is something that requires discussion as it can affect internal systems at Meta.

My plan is to only make retrocompatible changes, I will make a simple PR with just one of them as an example, to see if an RFC is needed

Issue #5247, opened 2022-01-21
Vectorize box decoding in FCOS model


jdsgomes:
### 🚀 The feature

The for loop in `torchvision/models/detection/fcos.py:80` can probably be vetorized.

```python
        pred_boxes = [
            self.box_coder.decode_single(bbox_regression_per_image, anchors_per_image)
            for anchors_per_image, bbox_regression_per_image in zip(anchors, bbox_regression)
        ]
```

### Motivation, pitch

Simplify/speedup the execution

abhi-glitchhg:
Can I try this?

datumbox:
@abhi-glitchhg absolutely! Thanks for offering to help.

Issue #3478, opened 2021-03-01
[feature request] Support batches for box_iou / generalized_box_iou


vadimkantorov:
Currently it supports only `(N4, M4) -> NM`. I propose to also support `(BN4, BM4) -> BNM`.

This is useful for computing a cost matrix between predicted boxes and ground truth boxes for a batch of frames. Probably it can be done by adjusting tensor indexing. Something like that:
```python
def _box_inter_union(boxes1: Tensor, boxes2: Tensor) -> Tuple[Tensor, Tensor]:
    area1 = box_area(boxes1)
    area2 = box_area(boxes2)

    lt = torch.max(boxes1[..., None, :2], boxes2[..., :2, None])
    rb = torch.min(boxes1[..., None, 2:], boxes2[..., 2:, None])

    wh = _upcast(rb - lt).clamp(min=0) 
    inter = wh[..., :, 0] * wh[..., :, 1]

    union = area1[..., None] + area2 - inter

    return inter, union
```

oke-aditya:
Hi. Sorry for the delay.

I think this can be done with `torch.vmap`. Have a look [here](https://pytorch.org/docs/master/generated/torch.vmap.html)

Have a look at [pytorch/issue/42368](https://github.com/pytorch/pytorch/issues/42368)

```
batched_iou = torch.vmap(box_iou) 

And now we can use boxes with batches

iou = batched_iou(batched_boxes1, batched_boxes2)

```
I don't have much experience with `vmap` but I think this would be extensible.
Can you give a run @vadimkantorov ?

vadimkantorov:
This already exists in detectron2: https://github.com/facebookresearch/detectron2/blob/cbbc1ce26473cb2a5cc8f58e8ada9ae14cb41052/detectron2/structures/boxes.py#L346

But given how frequent this is, I propose this (and maybe some related box ops) are included in core torchvision

PR #6203, opened 2022-06-25
[WIP] Vectorize box decoding in FCOS

abhi-glitchhg:
my attempt at #5247 

Currently WIP.


vadimkantorov:
Would it be possible to use `...` instead of `:, :`? This way is more flexible to number of batch dimensions.

Same question for batch encoding (which seems not even supported now :( )

In my own code I made these classes nn.Module and then do forward encoding in `forward()` and decoding in `inverse()`

abhi-glitchhg:
> Would it be possible to use `...` instead of `:, :`? This way is more flexible to number of batch dimensions.
> 

Ooo, Totally forgot about this method! Thanks @vadimkantorov

torchvision/models/detection/_utils.py
```diff
@@ -300,6 +300,32 @@ def decode_single(self, rel_codes: Tensor, boxes: Tensor) -> Tensor:
...
+        list_boxes = list_boxes.to(list_rel_codes.dtype)
+
+        list_ctr_x = 0.5 * (list_boxes[..., 0] + list_boxes[..., 2])
+        list_ctr_y = 0.5 * (list_boxes[..., 1] + list_boxes[..., 3])
+
+        if self.normalize_by_size:
+            list_boxes_w = list_boxes[..., 2] - list_boxes[..., 0]
+            list_boxes_h = list_boxes[..., 3] - list_boxes[..., 1]
+
+            list_box_size = torch.stack((list_boxes_w, list_boxes_h, list_boxes_w, list_boxes_h), dim=2)
```
datumbox, line 318: I believe the following is left-over from when you used `[:, :]` instead of `...`. To avoid assuming a specific number of dimensions, I think we can do:
```suggestion
            list_box_size = torch.stack((list_boxes_w, list_boxes_h, list_boxes_w, list_boxes_h), dim=-1)
```

datumbox (commented):
@abhi-glitchhg Thanks a lot for the PR. I've added a few comments as I'm uncertain about some of the changes made. Please let me know what you think.

@xiaohu2015 and @zhiqwang would you please be able to have a look and confirm that what we do here is correct? 

@vfdev-5 I would also appreciate if you could have a look and perhaps elaborate on the "equal box size" assumption you shared offline with me.

torchvision/models/detection/fcos.py
```diff
@@ -88,13 +88,12 @@ def compute_loss(
...
-            self.box_coder.decode_single(bbox_regression_per_image, anchors_per_image)
-            for anchors_per_image, bbox_regression_per_image in zip(anchors, bbox_regression)
-        ]
+
+        pred_boxes = self.box_coder.decode_all(bbox_regression, anchors)
+
         # amp issue: pred_boxes need to convert float
         loss_bbox_reg = generalized_box_iou_loss(
-            torch.stack(pred_boxes)[foregroud_mask].float(),
+            pred_boxes[foregroud_mask].float(),
```
vfdev-5, line 96: nit: I may be wrong but it seems to me that `.float` is a bit redundant as `pred_boxes` will be float due to the way centers are computed.

vfdev-5 (commented):
I have just few nits. Thanks for the PR @abhi-glitchhg

torchvision/models/detection/_utils.py
```diff
@@ -300,6 +300,32 @@ def decode_single(self, rel_codes: Tensor, boxes: Tensor) -> Tensor:
         pred_boxes = torch.stack((pred_boxes1, pred_boxes2, pred_boxes3, pred_boxes4), dim=1)
         return pred_boxes
 
+    def decode_all(self, list_rel_codes: Tensor, list_boxes: List[Tensor]) -> Tensor:
+        if len(list_rel_codes) != len(list_boxes):
+            min_len = min(len(list_rel_codes), len(list_boxes))
+            list_boxes, list_rel_codes = list_boxes[:min_len], list_rel_codes[:min_len]
```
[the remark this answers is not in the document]
abhi-glitchhg: Right! while coding, I was thinking about this synchronization issue. 
> I think we can assume here that the two lists are of the same size. 

With this assumption, I can remove this piece of code.  Will also add this assumption to the docstring.

torchvision/models/detection/_utils.py
abhi-glitchhg: sounds good! Thanks @datumbox !

torchvision/models/detection/_utils.py
```diff
@@ -300,6 +300,32 @@ def decode_single(self, rel_codes: Tensor, boxes: Tensor) -> Tensor:
...
+            list_boxes_w = list_boxes[..., 2] - list_boxes[..., 0]
+            list_boxes_h = list_boxes[..., 3] - list_boxes[..., 1]
+
+            list_box_size = torch.stack((list_boxes_w, list_boxes_h, list_boxes_w, list_boxes_h), dim=2)
+            list_rel_codes = list_rel_codes * list_box_size
+
+        pred_boxes1 = list_ctr_x - list_rel_codes[..., 0]
+        pred_boxes2 = list_ctr_y - list_rel_codes[..., 1]
+        pred_boxes3 = list_ctr_x - list_rel_codes[..., 2]
+        pred_boxes4 = list_ctr_y - list_rel_codes[..., 3]
```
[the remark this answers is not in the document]
abhi-glitchhg: You are right! Copy-Paste errors  :see_no_evil:

torchvision/models/detection/fcos.py
abhi-glitchhg: You are right!

abhi-glitchhg:
Thanks, @datumbox and @vfdev-5 for your valuable comments!. 

I checked the correctness of my code with the following script.


```python
from torchvision.models.detection._utils import BoxLinearCoder
import time 
import torch

box_coder = BoxLinearCoder(True)

for _ in range(10000):
    boxes = [torch.randn(3,4) for i in range(5)]
    rel_codes = torch.randn(5,3,4)
    

    original = [
            box_coder.decode_single(bbox_regression_per_image, anchors_per_image)
            for anchors_per_image, bbox_regression_per_image in zip(boxes, rel_codes)
        ]

    original  = torch.stack(original)
    new = box_coder.decode_all(rel_codes,boxes)

    torch.testing.assert_close(original, new, rtol=0, atol=1e-6)
```
This did not raise any error, so I assume the implementation is correct. [But please check this from your side too. :) ]

Assuming the above code is correct, I ran the following code to check if the implementation is faster or not.  
```python
import time
import torch 
from torchvision.models.detection._utils import BoxLinearCoder

box_coder = BoxLinearCoder(True)
start = time.process_time()
for _ in range(10000):
    boxes = [torch.randn(3,4) for i in range(5)]
    rel_codes = torch.randn(5,3,4)

    original = [
            box_coder.decode_single(bbox_regression_per_image, anchors_per_image)
            for anchors_per_image, bbox_regression_per_image in zip(boxes, rel_codes)
        ]
print(f"time taken for decode_single: {time.process_time() -start}")

start = time.process_time()
for _ in range(10000):
    boxes = [torch.randn(3,4) for i in range(5)]
    rel_codes = torch.randn(5,3,4)
    new = box_coder.decode_all(rel_codes, boxes)

print(f"time taken for decode_all: {time.process_time()-start}")

print('done ')
```
And results are as following: 
image


So we are getting 4x speed up on CPU..

vadimkantorov:
> > Would it be possible to use `...` instead of `:, :`? This way is more flexible to number of batch dimensions.
> 
> Ooo, I forgot about this method! Thanks @vadimkantorov
> 
> > In my own code I made these classes derived from nn.Module and then do forward encoding in forward() and decoding in inverse()
> 
> oh! I would love to have a look at the code! 😃

My related feature request on doing the same fix for other box-related functions: https://github.com/pytorch/vision/issues/3478

xiaohu2015:
@abhi-glitchhg good job! I think maybe we should consider merge the `decode_single` and `decode_all` because the share nearly same code.

torchvision/models/detection/fcos.py
```diff
@@ -88,13 +88,12 @@ def compute_loss(
 
         # regression loss: GIoU loss
         # TODO: vectorize this instead of using a for loop
```
zhiqwang, line 90: Maybe we could remove this `TODO` now 😄 others looks great to me!

datumbox, line 90:
```suggestion
```

datumbox (approved):
@abhi-glitchhg Thanks a lot for your work on this.


I've run similar checks on my side. On your first script that confirms that we produce the same results, I replaced the `assert_close` with `equal` to ensure it matches exactly and it does.

For benchmarks, I modified slightly your script to move the random input generation out of the loops and revert the stack that was part of the previous solution. Here is the updated script:
```python
import time
import torch
from torchvision.models.detection._utils import BoxLinearCoder

device = "cpu"
boxes = [torch.randn(3, 4).to(device=device) for i in range(5)]
rel_codes = torch.randn(5, 3, 4).to(device=device)

box_coder = BoxLinearCoder(True)

# Warmup

for _ in range(1000):
    original = [
            box_coder.decode_single(bbox_regression_per_image, anchors_per_image)
            for anchors_per_image, bbox_regression_per_image in zip(boxes, rel_codes)
        ]
    original = torch.stack(original)
    new = box_coder.decode_all(rel_codes, boxes)


start = time.process_time()
for _ in range(10000):
    original = [
            box_coder.decode_single(bbox_regression_per_image, anchors_per_image)
            for anchors_per_image, bbox_regression_per_image in zip(boxes, rel_codes)
        ]
    original = torch.stack(original)
print(f"time taken for decode_single: {time.process_time() -start}")

start = time.process_time()
for _ in range(10000):
    new = box_coder.decode_all(rel_codes, boxes)

print(f"time taken for decode_all: {time.process_time()-start}")

print('done ')
```

On CPU I get a 5.5x improvement and on GPU about 5x. So LGTM!

Thanks a lot to everyone for their reviews and suggestions. And +1 on cleaning up the single implementation as it's quite similar to this one. @abhi-glitchhg perhaps you could tackle this on a follow up PR?

Otherwise I think we are good to merge as this optimization seems quite good. :)

PR #6205, opened 2022-06-27
[proto] Draft for Transforms API v2

vfdev-5:
A draft for Transforms API v2


torchvision/prototype/transforms/_geometry.py
```diff
@@ -2,107 +2,104 @@
...
-        elif isinstance(input, PIL.Image.Image):
-            return F.horizontal_flip_image_pil(input)
-        elif is_simple_tensor(input):
-            return F.horizontal_flip_image_tensor(input)
+    def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
+        if isinstance(inpt, features._Feature):
+            return inpt.horizontal_flip()
+        elif isinstance(inpt, PIL.Image.Image):
+            return F.horizontal_flip_image_pil(inpt)
+        elif is_simple_tensor(inpt):
```
datumbox, line 30: Nit: we could avoid the `is_simple_tensor` and just check for isinstance(..., Tensor) now.

torchvision/prototype/features/_feature.py
```diff
@@ -7,7 +7,52 @@
 F = TypeVar("F", bound="_Feature")
 
 
-class _Feature(torch.Tensor):
+class _TransformsMixin:
```
datumbox, line 10: Question: Interesting. You don't plan to include primitive transforms related to colour? Is this because only images support them?

torchvision/prototype/transforms/_geometry.py
```diff
@@ -213,19 +202,19 @@ def __init__(self, size: Union[int, Sequence[int]]) -> None:
...
-        elif isinstance(input, PIL.Image.Image):
-            return MultiCropResult(F.five_crop_image_pil(input, self.size))
+    def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
+        if isinstance(inpt, features.Image):
+            output = F.five_crop_image_tensor(inpt, self.size)
+            return MultiCropResult(features.Image.new_like(inpt, o) for o in output)
+        elif is_simple_tensor(inpt):
+            return MultiCropResult(F.five_crop_image_tensor(inpt, self.size))
+        elif isinstance(inpt, PIL.Image.Image):
+            return MultiCropResult(F.five_crop_image_pil(inpt, self.size))
```
datumbox, line 212: Nit: Unless there is a reason for changing the order of ifs, I recommend to always keep the same order.

torchvision/prototype/features/_feature.py
```diff
@@ -7,7 +7,52 @@
...
+        # How dangerous to do this instead of raising an error ?
+        return self
+
+    def rotate(self, angle, *, interpolation, expand, fill, center):
+        # Just output itself
+        # How dangerous to do this instead of raising an error ?
+        return self
+
+
+class _Feature(_TransformsMixin, torch.Tensor):
```
datumbox, line 55: Question: Why an extra `_TransformsMixin` class and not putting everything directly in the `_Feature`? Also do you have a recommendation to rename `_Feature` (it's a placeholder name that causes confusion)?

datumbox (commented):
Looking cool so far. Just a few questions:

torchvision/prototype/features/_feature.py
vfdev-5: I haven't yet worked on that. But yes, color-related method will be there as well

torchvision/prototype/features/_feature.py
```diff
@@ -7,7 +7,52 @@
...
+
+        # To avoid circular dependency between features and transforms
+        from ..transforms import functional as F
+
+        self._F = F
+
+    def horizontal_flip(self):
+        # Just output itself
+        # How dangerous to do this instead of raising an error ?
+        return self
```
datumbox, line 22: I would raise a "not implemented" error here and implement no-op explicitly on the implementations. So if bbox can't blur, then it should explicitly implement to be no-op as opposed to leaving the default implementation which returns self. Could help us avoid issues but if you disagree I'm happy to leave as-is and discuss later.

vfdev-5: I agree that raising `NotImplemented` could be good. The only drawback I found that all targets features like `Label`, `OneHotLabel` should also implement explicitly these no-ops. From the first glance, `OneHotLabel.blur(self, *args, **kwargs)` as no-op made me think that it is a bit too much...

datumbox: Up to you. I've added this for your consideration. Feel free to ignore or postpone for later.

torchvision/prototype/features/_feature.py
datumbox: Yeap sorry, I read more into the absence of the methods than I should. 😅

vfdev-5:
Closed in favor of https://github.com/pytorch/vision/pull/6305

PR #3200, opened 2020-12-22
Transforms documentation clean-up

voldemortX:
A documentation clean-up that concerns` transform.py functional.py functional_pil.py functional_tensor.py` for issue #3071.
@datumbox After some initial checking, there are somethings I'm not certain:

1. Does the `Sequence` type represent only list and tuple in torchvision? I find that in Python other types (e.g. strings) can also count as a Sequence. Maybe we should clear it up in the docs, like maybe just say list or tuple.
2. Does the Posterize operation also only support uint8 images in PIL? I'm inclined to add "only support L or RGB" to all related transforms like I did with invert(). But maybe we can just let PIL throw the OSError.
3. Should we also add "This transform supports both PIL and Tensor, ..." in the documentation for AutoAugment as well?
4. Maybe we could delete the private function docs, as mentioned by @vfdev-5 [here](https://github.com/pytorch/vision/pull/3069#pullrequestreview-541808920).

And there are some other things I found after cleaning the docs up, maybe they can be addressed in the future:

1. In Pad (or RandomCrop, it also uses padding), it seems tensors only support single value fill. Also I'm not sure whether str is really needed here by PIL, what is the use case here?
2. Some type checking is also Sequence, maybe we should check explicitly for tuple and list. Or is this the intended implementation?

PR #3069, opened 2020-12-01
Check num of channels on adjust_* transformations

datumbox:
Some of the transformations can be applied only to RGB or Grayscale images, nevertheless currently we don't explicitly check to ensure that an image with the right number of channels is provided. In the past this was not a problem since `io.image` only supported RGB, nevertheless recent PRs #2984, #2988 and #3024 added support for additional image types. Though the tensor meta data are not enough to tell us the type of image and its color space, the number of channels can be used as a proxy.

In this PR, I explicitly check for the right number of channels in all the  adjust_* transforms. Note that those transforms don't handle yet correctly transparency. There are other transformations that could potentially provide incorrect results when transparency is available but I don't patch here because they might be used in other applications. Some of these transformations are `affine()`, `rotate()` and `perspective()`.

PR #6209, opened 2022-06-28
Minor fix antialias arg

vfdev-5:
Description:

Antialias was defined as None in the public API as we deal with Pillow and Tensor backend. Pillow applies unconditionally the anti-aliasing and there is no way to disable it. Thus for Pillow we keep `antialias=None`. For Tensors, we can do antialias=True or False, None->False. 

The purpose of this PR is to make `antialias` argument's default value and type hint more explcit in proto's `resize_image_tensor` signature and in private `resize` methods. Mainly, for tensors `antialias=None` does not make much sense.

### DIFF
--- a/torchvision/transforms/functional_tensor.py
+++ b/torchvision/transforms/functional_tensor.py
@@ -430,15 +430,12 @@ def resize(
     img: Tensor,
     size: List[int],
     interpolation: str = "bilinear",
-    antialias: Optional[bool] = None,
+    antialias: bool = False,
 ) -> Tensor:
     _assert_image_tensor(img)
 
     if isinstance(size, tuple):
         size = list(size)
-
-    if antialias is None:
-        antialias = False
 
     if antialias and interpolation not in ["bilinear", "bicubic"]:
         raise ValueError("Antialias option is supported for bilinear and bicubic interpolation modes only")
--- a/torchvision/prototype/transforms/functional/_geometry.py
+++ b/torchvision/prototype/transforms/functional/_geometry.py
@@ -45,7 +45,7 @@ def resize_image_tensor(
     size: List[int],
     interpolation: InterpolationMode = InterpolationMode.BILINEAR,
     max_size: Optional[int] = None,
-    antialias: Optional[bool] = None,
+    antialias: bool = False,
 ) -> torch.Tensor:
     num_channels, old_height, old_width = get_dimensions_image_tensor(image)
     new_height, new_width = _compute_output_size((old_height, old_width), size=size, max_size=max_size)
--- a/torchvision/transforms/functional.py
+++ b/torchvision/transforms/functional.py
@@ -471,6 +471,9 @@ def resize(
         pil_interpolation = pil_modes_mapping[interpolation]
         return F_pil.resize(img, size=output_size, interpolation=pil_interpolation)
 
+    if antialias is None:
+        antialias = False
+
     return F_t.resize(img, size=output_size, interpolation=interpolation.value, antialias=antialias)
 
 


YosuaMichael (approved):
Thanks @vfdev-5 , LGTM

torchvision/transforms/functional_tensor.py
```diff
@@ -430,16 +430,13 @@ def resize(
...
+    antialias: bool = False,
 ) -> Tensor:
     _assert_image_tensor(img)
 
     if isinstance(size, tuple):
         size = list(size)
 
-    if antialias is None:
-        antialias = False
-
```
NicolasHug, line 442: I think this is a BC-breaking change for users who passed `antialias=None`.

Considering the relatively low benefit of removing support for `None`, we might want to keep it to avoid any potential issues?

NicolasHug: This function is importable without a leading underscore as

```py
from torchvision.transforms.functional_tensor import resize
```

and as a result, it is likely that most users will consider it public, as this is a widely used convention throughout the pydata ecosystem.

I think it's fine to modify the prototype, but I'd rather keep the stable area untouched to avoid any potential breakage. Especially considering the low benefit, and the fact that these transforms will eventually be superseded by the prototype ones.

vfdev-5: Yeah, we should have added leading underscore to them. It very unfortunate if people today rely on this undocumented API. In this PR https://github.com/pytorch/vision/pull/3200 we removed "PRIVATE METHOD" in the docstring.
Previously I already refactored and removed an arg from `resize` method in `functional_pil` and `functional_tensor` https://github.com/pytorch/vision/pull/6191 . 
Should we restore the arg for BC and warn about new behaviour ?

YosuaMichael: Sorry just realized this is a public function, thanks @NicolasHug for noticing.
I agree that we shouldn't make BC changes in this case.

Looking at the update of @vfdev-5 comment, I think `resize_image_tensor` is under prototype which should be okay to change explicitly, but lets keep functional_tensor.resize unchanged.

YosuaMichael (changes requested):
As commented by @NicolasHug , lets keep `functional_tensor.resize` unchanged to prevent BC changes

torchvision/transforms/functional_tensor.py
NicolasHug: > Yeah, we should have added leading underscore to them. It very unfortunate if people today rely on this undocumented API

I agree. Unfortunately right now, there is literally nothing that distinguishes `torchvision.transforms.functional` from `torchvision.transforms.functional_tensor`. So we can't assume users know that one is private while the other one isn't.

> Previously I already refactored and removed an arg from resize method in functional_pil and functional_tensor https://github.com/pytorch/vision/pull/6191 .
Should we restore the arg for BC and warn about new behaviour ?


Yes, I think this is safer. What do you mean about "warning about new behaviour" though? We can't really do silent changes either in this case, we would have to go through deprecations.

vfdev-5: > Unfortunately right now, there is literally nothing that distinguishes torchvision.transforms.functional from torchvision.transforms.functional_tensor

The only diff is that public methods are documented and visible in the docs. Private methods do not have any docstring and invisible in the docs. 

> What do you mean about "warning about new behaviour" though? We can't really do silent changes either in this case, we would have to go through deprecations.

Yes, for public API is it the right way to proceed. I still have them in private space that's why was suggesting something in between.

I think we should decide and make things more clear with `functional_tensor` and `functional_pil`. If they are private, let's put them to `_functional_tensor.py` and `_functional_pil.py` with all necessary warnings/deprecations etc.

NicolasHug: > I still have them in private space

Sorry what do you mean? We should treat everything in `functional_tensor` and `functional_pil` as public

vfdev-5: > After that the decision was made to remove the documentation warning

Decision was very quick: https://github.com/pytorch/vision/pull/3069#pullrequestreview-541808920 and https://github.com/pytorch/vision/pull/3200

datumbox (approved):
So let me put some context for @vfdev-5 and for the people who read the thread to understand better @NicolasHug's concerns at https://github.com/pytorch/vision/pull/6209/files#r908517255.

First of all, we all probably agree that the way `functional_*` are structured are unfortunate. The APIs have been communicated to be private but then after v0.8 there is no clear indication. So to Nicolas' point, how are users supposed to know? There is definitely room for improvement on the deprecation policies we used previously and that's exactly what Nicolas is trying to do now. He is going through all the cases where this is not communicated properly, updates the docs and adds mechanisms in place to let users know.

Concerning the changes on this PR, I don't think the modifications are strictly necessary for the Transforms v2 API (perhaps they can be avoided?). Nevertheless I see a potential issue if we block this change. Either we agree or disagree with policies introduced on old version of TorchVision, we must respect them. The specific API was marked private using the deprecation process followed at that time (around version v0.8). I don't think we should go back and undo this now (aka consider it public using today's definition). This puts a very bad precedence for other parts of the API such as the C++ side of things where everything is considered private or the Beta APIs where there are no BC guarantees. So my concern here is that blocking this PR will open the gates for blocking other important work on private or Beta APIs on the future. My worry is that the next PR might require a modification on this private API and it will be blocked because are now reconsidering it to be public.

For the above reasons and because the PR is technically correct and on a private API, I will approve it. I will leave it to @vfdev-5 to decide if he wants to merge it or close because it's not critical to his work. Again, I agree with @NicolasHug's concerns here; I'm just equally concerned that if we adopt this very strict policy it's going to come back and bite us on the near future as it will slow down critical planned work. I think more discussions are necessary around these corner-cases once we are all back at the office. The good thing is, we just released TorchVision yesterday which means we got 3+ months until the next one so we can always revert the decision afterwards.

vfdev-5:
Closing as outdated

PR #6219, opened 2022-06-29
[proto] Added mid-level ops and feature-based ops

vfdev-5:
Description:

- Added functional mid-level ops
- Added Image/BoundingBoxes/SegmentationMask ops: `Image.resize` etc

Source: https://github.com/pytorch/vision/pull/6205


torchvision/prototype/features/_bounding_box.py
```diff
@@ -69,3 +70,140 @@ def to_format(self, format: Union[str, BoundingBoxFormat]) -> BoundingBox:
...
+        output = _F.perspective_bounding_box(self, self.format, perspective_coeffs)
+        return BoundingBox.new_like(self, output, dtype=output.dtype)
+
+    def erase(self, i: int, j: int, h: int, w: int, v: torch.Tensor) -> BoundingBox:
+        raise TypeError("Erase transformation does not support bounding boxes")
+
+    def mixup(self, lam: float) -> BoundingBox:
+        raise TypeError("Mixup transformation does not support bounding boxes")
+
+    def cutmix(self, box: Tuple[int, int, int, int], lam_adjusted: float) -> BoundingBox:
```
datumbox, line 208: I don't believe these operations should be kernels. Mixup, Cutmix etc are augmentation strategies. I would even be inclined not to add erase and let people to access the functional directly. The rational is that we should keep the kernels low. As we keep adding augmentations, we do this on the Transforms side. Thoughts?

torchvision/prototype/features/_feature.py
```diff
@@ -83,3 +90,159 @@ def __torch_function__(
...
+        # Just output itself
+        # How dangerous to do this instead of raising an error ?
+        return self
+
+    def vertical_flip(self) -> Any:
+        # Just output itself
+        # How dangerous to do this instead of raising an error ?
+        return self
+
+    def resize(  # type: ignore[override]
```
datumbox, line 104: Why `ignore[override]` is needed here?

torchvision/prototype/transforms/_augment.py
```diff
@@ -82,59 +83,45 @@ def _get_params(self, sample: Any) -> Dict[str, Any]:
...
-        if isinstance(input, features.Image):
-            output = F.erase_image_tensor(input, **params)
-            return features.Image.new_like(input, output)
-        elif is_simple_tensor(input):
-            return F.erase_image_tensor(input, **params)
+    def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
+        if isinstance(inpt, features._Feature):
+            return inpt.erase(**params)
+        elif isinstance(inpt, PIL.Image.Image):
+            # Shouldn't we implement a fallback to tensor ?
```
datumbox, line 92: Yes let's create a quick and dirty implementation that casts to tensor and back to PIL similar to what we do at `gaussian_blur`.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -474,7 +597,7 @@ def perspective_image_tensor(
 
 def perspective_image_pil(
     img: PIL.Image.Image,
-    perspective_coeffs: float,
+    perspective_coeffs: List[float],
```
datumbox, line 600: I assume that was a bug/typo previously?

torchvision/prototype/features/_feature.py
vfdev-5: there is torch.Tensor built-in deprecated op: Tensor.resize : https://github.com/pytorch/pytorch/blob/e8727994eb7cdb2ab642749d6549bc497563aa06/torch/_tensor.py#L588-L593

torchvision/prototype/transforms/functional/_geometry.py
vfdev-5: yes, exactly

torchvision/prototype/transforms/_augment.py
```diff
@@ -82,59 +83,45 @@ def _get_params(self, sample: Any) -> Dict[str, Any]:
...
+            raise RuntimeError("Not implemented")
+        elif isinstance(inpt, torch.Tensor):
```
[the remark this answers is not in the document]
vfdev-5: we can just repeat one line:
```
            inpt_tensor = some_pil_to_tensor_method(inpt)
            return F.erase_image_tensor(inpt_tensor, **params)
        elif isinstance(inpt, torch.Tensor):
            return F.erase_image_tensor(inpt, **params)
```

torchvision/prototype/features/_feature.py
datumbox: holly molly! that's not good. Does it make sense to rename the method? I can see that Tensor.resize_ exists still on the [docs](https://pytorch.org/docs/stable/generated/torch.Tensor.resize_.html?highlight=tensor%20resize#torch.Tensor.resize_) but can't find the resize.

vfdev-5: non-inplace Tensor.resize is deprecated -> no docs. IMO, we can keep both.

torchvision/prototype/transforms/_augment.py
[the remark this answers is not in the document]
datumbox: Sure but what if you have the kernel. I'm asking in general what idiom you will favour when you call the same F.

torchvision/prototype/features/_feature.py
datumbox: Add a big TODO to get feedback on this. cc @NicolasHug thoughts?

torchvision/prototype/features/_bounding_box.py
```diff
@@ -69,3 +70,142 @@ def to_format(self, format: Union[str, BoundingBoxFormat]) -> BoundingBox:
         return BoundingBox.new_like(
             self, convert_bounding_box_format(self, old_format=self.format, new_format=format), format=format
         )
+
+    def horizontal_flip(self) -> BoundingBox:
+        from torchvision.prototype.transforms import functional as _F
```
vadimkantorov, line 75: Btw if this happens on a hot path, this import (even if it's not the first run) may hurt perf (from my experience of a few years back)

vfdev-5: Yeah, this is not ideal solution. Previously I tried to add the submodule as an attribute and somehow dataloader with multiple processes hangs due to that...
If you have any better ideas, please share

datumbox: Yeah that's a fair point. This workaround is temporary for as long as we work on the API. We should seek a better solution prior finalising it. Potentially a refactoring and reorganization of the modules might be a solution here but that can be decided later.

torchvision/prototype/features/_feature.py
```diff
@@ -132,13 +117,11 @@ def resized_crop(
...
     ) -> Any:
-        # Just output itself
-        # How dangerous to do this instead of raising an error ?
         return self
 
-    def pad(self, padding: List[int], fill: int = 0, padding_mode: str = "constant") -> Any:
-        # Just output itself
-        # How dangerous to do this instead of raising an error ?
+    def pad(
+        self, padding: List[int], fill: Union[int, float, Sequence[float]] = 0, padding_mode: str = "constant"
```
datumbox, line 123: Why don't we support `Sequence[int]` as well? Did you face JIT-scriptability issues related to how it handles seq of ints and floats? If we add support, we need to put them in all places.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -503,10 +503,48 @@ def rotate(
...
+    padding: List[int],
+    fill: Sequence[float] = [0.0],
+    padding_mode: str = "constant",
+) -> torch.Tensor:
+    if padding_mode != "constant":
+        raise ValueError(f"Padding mode '{padding_mode}' is not supported if fill is not scalar")
+
+    output = pad_image_tensor(img, padding, fill=0, padding_mode="constant")
+    left, top, right, bottom = padding
+    fill = torch.tensor(fill, dtype=img.dtype, device=img.device).view(-1, 1, 1)
```
datumbox, line 535: Good trick. This will ensure we respect the image type, no matter the fill provided.

torchvision/prototype/features/_segmentation_mask.py
```diff
@@ -1,5 +1,124 @@
...
+    def vertical_flip(self) -> SegmentationMask:
+        from torchvision.prototype.transforms import functional as _F
+
+        output = _F.vertical_flip_segmentation_mask(self)
+        return SegmentationMask.new_like(self, output)
+
+    def resize(  # type: ignore[override]
+        self,
+        size: List[int],
+        interpolation: InterpolationMode = InterpolationMode.BILINEAR,
```
datumbox, line 27: I understand that this is not used but should we have as default value `BILINEAR` which is not really supported for masks? I think it's best if everything changes to NEAREST in this file, to at least indicate that the right value is used.

torchvision/prototype/transforms/_augment.py
datumbox: Change the comment to a TODO.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -40,12 +40,58 @@ def horizontal_flip_bounding_box(
     ).view(shape)
 
 
+def horizontal_flip(inpt: Any) -> Any:
```
datumbox, line 43: Why Any and not Union[Tensor, PIL.Image]?

torchvision/transforms/functional_pil.py
```diff
@@ -317,9 +317,9 @@ def rotate(
 @torch.jit.unused
 def perspective(
     img: Image.Image,
-    perspective_coeffs: float,
+    perspective_coeffs: List[float],
     interpolation: int = _pil_constants.BICUBIC,
-    fill: Optional[Union[float, List[float], Tuple[float, ...]]] = 0,
+    fill: Optional[Union[float, List[float], Tuple[float, ...]]] = None,
```
datumbox, line 322: Here is my understanding for why this is OK. The change on `fill` doesn't have any effect on the actual operation due to the call on `_parse_fill()`. Concerning the type of `perspective_coeffs` I believe that the change actually fixes a bug because PIL breaks if you pass a number:
```
TypeError: 'int' object is not subscriptable
```

datumbox (approved):
I understand that things started piling up and it will help if we do the changes incrementally. I'm approving as in principle, I believe the approach is correct. Below I have some comments and questions for your consideration.

torchvision/prototype/transforms/functional/_geometry.py
vfdev-5: If we replace Any by `union[tensor, pil-image]` should we keep`else` branch ? In this case, it could be 
```
def horizontal_flip(inpt: Union[torch.Tensor, PIL.Image.Image]) -> Union[torch.Tensor, PIL.Image.Image]:
    if isinstance(inpt, features._Feature):
        return inpt.horizontal_flip()
        return horizontal_flip_image_pil(inpt)
    # elif isinstance(inpt, torch.Tensor):
    #     return horizontal_flip_image_tensor(inpt)
    # else:
    #     return inpt
    else:
        return horizontal_flip_image_tensor(inpt)
```

torchvision/prototype/transforms/functional/_geometry.py
vfdev-5: I just copied from another file :)

torchvision/prototype/transforms/functional/_geometry.py
datumbox: Agreed. As discussed earlier, the middle-layer shouldn't return the input as-is on else. I assume this will be removed on a follow up PR? Also we might want to play around with the order of the ifs and the type info, to see if there is any way we can make the middle layer JIT-scriptable.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -643,6 +808,14 @@ def center_crop_segmentation_mask(segmentation_mask: torch.Tensor, output_size:
     return center_crop_image_tensor(img=segmentation_mask, output_size=output_size)
 
 
+def center_crop(inpt: DType, output_size: List[int]) -> DType:
+    if isinstance(inpt, features._Feature):
+        return inpt.center_crop(output_size)
+    elif isinstance(inpt, PIL.Image.Image):
```
datumbox, line 814: Just a minor nit, but we should decide if we will use the if/if/return idiom or if/elif/else or if/elif/return and keep the same everywhere. 

The changes still LGTM overall.

torchvision/prototype/features/_feature.py
vfdev-5: I think we can add `Sequence[int]` to the type hint (I didn't add it as it starts looking very bulky). JIT is not concerned as `pad` is not scriptable (in general torch script does not recognize `Sequence` and we should map it to `List`).
In following PR, we have to make same type hint for all `fill` usages.

datumbox: Sounds good. You wanna put a TODO on the code or you keep track of this elsewhere?

vfdev-5: I'll be dealing with this in Transforms PR (next one, not yet sent)

torchvision/prototype/features/_feature.py
vfdev-5: @NicolasHug I agree that the TODO note is unclear. I'll update in a follow-up PR. The point is to fix `# type: ignore[override]` and make mypy happy.

vfdev-5: This is a good question. Maybe it is impossible. I was thinking about mypy overload: https://mypy.readthedocs.io/en/stable/more_types.html#function-overloading if this could help...

torchvision/prototype/features/_bounding_box.py
datumbox: Here is an alternative approach for this at #6476

PR #6222, opened 2022-06-30
Swin transformer handle window size smaller than input size

YosuaMichael:
This PR introduce logic in swin transformer to update the window_size if it is bigger than the input size.
This behaviour follow the original implementation: https://github.com/microsoft/Swin-Transformer/blob/main/models/swin_transformer.py#L192-L195

We introduced this change on a commit in #6088, however since we want to cherrypick that PR and the main goal of #6088 is to adapt the components so it can be reused on the swin transformer 3d, we decide to revert back this change in order to lower the risk during the release. 

However since this change can be useful especially if we want to adapt the components to swin transformer v2 (as commented by @xiaohu2015 in https://github.com/pytorch/vision/pull/6088#issuecomment-1170819104), we introduce the change back in this PR.

We have tested this change by running the validation script:
```
    --timeout 3000 --nodes 1 --ngpus 4 --batch-size=1 \
    --partition train --model <model_variant> \
    --data-path="/datasets01_ontap/imagenet_full_size/061417" \
    --weights="<weight_variant>_Weights.IMAGENET1K_V1" \
    --test-only \
```
where model_variant can be `swin_t, swin_s, swin_b` and weight_variant can be `Swin_T, Swin_S, Swin_B`.
Here are the validation resutl:

- swin_t, Swin_T : Acc@1 81.472 Acc@5 95.780
- swin_s, Swin_S: Acc@1 83.186 Acc@5 96.366
- swin_b, Swin_B: Acc@1 83.584 Acc@5 96.636

### DIFF
--- a/torchvision/models/swin_transformer.py
+++ b/torchvision/models/swin_transformer.py
@@ -1,5 +1,5 @@
 from functools import partial
-from typing import Optional, Callable, List, Any
+from typing import Any, Callable, List, Optional, Tuple
 
 import torch
 import torch.nn.functional as F
@@ -9,7 +9,7 @@ from ..ops.misc import MLP, Permute
 from ..ops.stochastic_depth import StochasticDepth
 from ..transforms._presets import ImageClassification, InterpolationMode
 from ..utils import _log_api_usage_once
-from ._api import WeightsEnum, Weights
+from ._api import Weights, WeightsEnum
 from ._meta import _IMAGENET_CATEGORIES
 from ._utils import _ovewrite_named_param
 
@@ -374,7 +374,7 @@ class SwinTransformer(nn.Module):
         # build SwinTransformer blocks
         for i_stage in range(len(depths)):
             stage: List[nn.Module] = []
-            dim = embed_dim * 2 ** i_stage
+            dim = embed_dim * 2**i_stage
             for i_layer in range(depths[i_stage]):
                 # adjust stochastic depth probability based on the depth of the stage block
                 sd_prob = stochastic_depth_prob * float(stage_block_id) / (total_stage_blocks - 1)
--- a/torchvision/models/swin_transformer.py
+++ b/torchvision/models/swin_transformer.py
@@ -160,7 +160,20 @@ def shifted_window_attention(
     return x
 
 
+def _fix_window_and_shift_size(
+    input_hw: List[int], window_size: List[int], shift_size: List[int]
+) -> Tuple[List[int], List[int]]:
+    # Handle case where window_size is larger than input tensor
+    # Reference on the original implementation: https://github.com/microsoft/Swin-Transformer/blob/main/models/swin_transformer.py#L192-L195
+    for i in range(2):
+        if input_hw[i] <= window_size[i]:
+            window_size[i] = input_hw[i]
+            shift_size[i] = 0
+    return window_size, shift_size
+
+
 torch.fx.wrap("shifted_window_attention")
+torch.fx.wrap("_fix_window_and_shift_size")
 
 
 class ShiftedWindowAttention(nn.Module):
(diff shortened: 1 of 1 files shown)


torchvision/models/swin_transformer.py
```diff
@@ -218,8 +231,12 @@ def forward(self, x: Tensor):
         Returns:
             Tensor with same layout as input, i.e. [B, H, W, C]
         """
+        _, H, W, _ = x.shape
+        input_hw = [H, W]
+        # Handle case where the window_size is larger than the input
+        window_size, shift_size = _fix_window_and_shift_size(input_hw, self.window_size, self.shift_size)
```
YosuaMichael, line 237: The [original implementation](https://github.com/microsoft/Swin-Transformer/blob/main/models/swin_transformer.py#L192-L195) do this in the constructor. However in the original implementation they require to have `input_resolution` in the constructor which we dont have here. Therefore I do this in the forward so I can access the input size.

torchvision/models/swin_transformer.py
```diff
@@ -160,7 +160,20 @@ def shifted_window_attention(
     return x
 
 
+def _fix_window_and_shift_size(
+    input_hw: List[int], window_size: List[int], shift_size: List[int]
```
datumbox, line 164: Should `input_hw` be named `input_size` so that it is aligned with 3d versions?

torchvision/models/swin_transformer.py
```diff
@@ -160,7 +160,20 @@ def shifted_window_attention(
...
+) -> Tuple[List[int], List[int]]:
+    # Handle case where window_size is larger than input tensor
+    # Reference on the original implementation: https://github.com/microsoft/Swin-Transformer/blob/main/models/swin_transformer.py#L192-L195
+    for i in range(2):
```
datumbox, line 168: Does it make sense to not hard-code 2 but make it an argument? Do you think this is likely to be reused in 3d?

torchvision/models/swin_transformer.py
datumbox: The issue here is you modify the `self.window_size` and `self.shift_size` methods in-place, leaving side-effects. Is this intentional? If not we should copy them. If yes, it might be worth reviewing the logic, because as far as I understand once the values are modified you currently have no way to recover the original values.

torchvision/models/swin_transformer.py
```diff
@@ -357,7 +374,7 @@ def __init__(
         # build SwinTransformer blocks
         for i_stage in range(len(depths)):
             stage: List[nn.Module] = []
-            dim = embed_dim * 2 ** i_stage
+            dim = embed_dim * 2**i_stage
```
datumbox, line 377: Revert, that's likely to break the linter.

datumbox (commented):
Thanks @YosuaMichael. I left a couple of comments, let me know what you think.

Edit: also the [failing](https://app.circleci.com/pipelines/github/pytorch/vision/18635/workflows/eee26e08-5d9b-4b69-b491-95586e5b61a5/jobs/1507523) tests seem related.

torchvision/models/swin_transformer.py
YosuaMichael: Thanks for noticing this @datumbox ! Yes, I think we shouldn't modify `self.window_size` and this is currently not intentional. I will change the function `_fix_window_and_shift_size` so that it does not modify the class variable.

torchvision/models/swin_transformer.py
YosuaMichael: hmm... this is strange, somehow my local ufmt is the one changing this line. Let me fix my local ufmt and revert this.

YosuaMichael:
Investigating on the failing test, the changes affect the result when the input size is 112 or smaller.
And when I try the resize_size and crop_size on imagenet I got the following resize:
```
resize_size=256, crop_size=112
- [Original] Acc@1 66.318 Acc@5 85.934
- [This PR] Acc@1 64.236 Acc@5 84.412

resize_size=128, crop_size=112
- [Original] Acc@1 69.248 Acc@5 88.824
- [This PR] Acc@1 67.060 Acc@5 87.218
```
As we can see, this PR reduce the accuracy significantly.

When looking at the reason, it because the `relative_position_index` depend on the `window_size`. And I think this is the reason the original implementation put the `window_size update` on the constructor with a cost of having the `input_resolution` specified on the constructor as well (which mean the model can only accept a fixed input).

Looking at this, I think this change need more discussion and if we really want to implement like the original implementation its better to have a "better" reason like to implement Swin Transformer V2. Therefore for now I will close this PR.

xiaohu2015:
@YosuaMichael I think we can do some modification on the model: the `window_size` can be `List[int]` or `List[List[int]]`, if we use `List[int]`, then the window size of all block should be the same (just when the input size is 224 or greater), if we use `List[List[int]]`, we can choose use different window size for each block ([[7, 7], [7, 7], [7, 7], [6, 6]] for input size 192.

YosuaMichael:
Hi @xiaohu2015 , sorry for the late reply! I am trying to understand your suggestion on having window_size specified for each layer. In this case, is it correct to say that the user need to retrain the model in order to change the window_size?
For instance if we train our model in input size 224, then if the user want to have input of size 384 then he need to change window_size and hence the pretrained weight cannot load and we need to have separate weight for input size 384?

xiaohu2015:
@YosuaMichael yes.

PR #6230, opened 2022-07-01
Add build-system requirements

malfet:
To make project [PEP-517](https://peps.python.org/pep-0517/) compliant

Solves https://github.com/pytorch/pytorch/issues/80745

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,3 +12,7 @@ target-version = ["py37"]
 excludes = [
     "gallery",
 ]
+
+[build-system]
+
+requires = ["setuptools", "torch", "wheel"]


datumbox (approved):
LGTM,  thanks @malfet

pyproject.toml
```diff
@@ -12,3 +12,7 @@ target-version = ["py37"]
 excludes = [
     "gallery",
 ]
+
+[build-system]
+
+requires = ["setuptools", "torch", "wheel"]
```
pmeier, line 18: Adding `torch` as a requirement here does only somewhat solve this:

1. `pip` will create a separate virtual env for these requirements, which means it will pull `torch` from PyPI. That means you will get a CPU version for macOS and Windows and an outdated CUDA version (10.2 IIRC) for Linux.
2. We only have the stable binaries up on PyPI, but in general we depend on the current nightly.

Thus, even with this set, users could have problems building from source.

I don't have a solution for this. Just raising awareness.

PR #6241, opened 2022-07-06
Adding video accuracy for video_classification reference script

YosuaMichael:
During our recent work on improving TorchVision’s support for Video models, we’ve discovered a perceived gap in the accuracy compared to other libraries such as PyTorch Video and SlowFast. This post summarizes the findings of our investigation.

**TLDR:**

The differences were originally observed on the MViT and later on the ResNet-based Video models (R3D, MC3 and R2+1). We are happy to confirm that the accuracy difference was not due to a bug on TorchVision’s models, video APIs, transforms or datasets but instead was the result of using different metrics (clip accuracy vs video accuracy) and of the frame rate configuration. Our team made the necessary changes on our documentation and reference scripts to allow for the easier reproduction of our pre-trained models and confirmed that there are no accuracy gaps or bugs.

## Why do the accuracies differ?

Our initial analysis showed an accuracy gap of 5.3 Acc@1 points for MViT, 10.1 for R3D, 10.6 for MC3 and 10.0 for R2+1. These metrics were estimated by running the same models through the reference scripts of TorchVision and SlowFast. Since the difference is very large, we decided to investigate to find the root cause.

The libraries implement similar features but there are some differences that have small effect on the accuracies:

* Different Video decoding backends (PyAV vs TorchVision decoder). Not all videos can be decoded by the 2 backends so one needs to control for running the analysis on the same data.
* Different temporal sampling strategies on how clips are sampled from videos.
* Minor differences on the order of execution of inference transforms.
* TorchVision’s reference scripts support integer frame rates, while floats are supported by SlowFast.

The following differences were found to have significant effects on the accuracies:

* SlowFast supports [2 hyper-parameters](https://github.com/facebookresearch/SlowFast/blob/1aebd71a2efad823d52b827a3deaf15a56cf4932/slowfast/datasets/decoder.py#L307) to control frame sampling: the `sampling_rate` and `target_fps`. TorchVision supports a [single parameter](https://github.com/pytorch/vision/blob/4c66813985a93ed583c8483149c8dfef9c1f3597/torchvision/datasets/video_utils.py) called `frame_rate` and thus it needs to be configured when porting pre-trained weights from SlowFast.
* Both libraries produce multiple temporal clips from the same videos during inference. Nevertheless, the way the results are combined to produce the final values differ significantly. TorchVision estimates the overall accuracies on the [clip-level](https://github.com/pytorch/vision/blob/8f98aee5f48962ba7c73cbcc0e3df5c13bc6b5e8/references/video_classification/train.py#L67-L68). SlowFast uses a multi-view [ensemble](https://github.com/facebookresearch/SlowFast/blob/1aebd71a2efad823d52b827a3deaf15a56cf4932/slowfast/utils/meters.py#L247), which aggregates the results of clips on the video-level. This strategy has a material effect on the final reported accuracy.

## How to make them equal?

To much the accuracies between TorchVision and SlowFast we [took](https://github.com/pytorch/vision/pull/6241) the following steps:

* For the case of MViT, we adjust the hyper-parameters of the sampling configuration. SlowFast uses `target_fps=30` and `sampling_rate=4` while the equivalent value for TorchVision is `frame_rate=7.5` (rounded to 8 as we don't support fractional).
* We adopt in TorchVision’s reference scripts, the SlowFast’s ensemble strategy for aggregating accuracies on video level.
* We document the changes on our documentation and meta-data to ensure our users can reproduce the results easier.

All changes are on the documentation and reference scripts thus they don’t affect production users.

## Reproducing the Results

We run the validation on Kinetics-400 datasets with some video excluded because they can't be decoded in either pyav or torchvision backend. Here is the [list of the excluded video](https://gist.github.com/YosuaMichael/02533740966be21d0f0a608c9ca8fb87#file-removed_videos-json). Note that the estimates below are on 8-GPUs with batch-size > 1, meaning that a small variation is expected due to #4559.

Here are the commands and result in MViT:
```
    --timeout 3000 --ngpus 8 --nodes 1 \
    --data-path="/datasets/clean_kinetics_400/" \
    --batch-size=16 --test-only \
    --clip-len 16 --frame-rate 8 --clips-per-video 5 \
    --cache-dataset \
    --model mvit_v1_b --weights="MViT_V1_B_Weights.DEFAULT" 
    
# result:
# Test: Total time: 1:25:33
# * Clip Acc@1 70.351 Clip Acc@5 88.288
# * Video Acc@1 78.477 Video Acc@5 93.582
```

And here are commands and result for Video ResNet models:
```
    --timeout 3000 --ngpus 8 --nodes 1 \
    --batch-size=64 --test-only \
    --data-path="/datasets/clean_kinetics_400/" \
    --clip-len 16 --frame-rate 15 --clips-per-video 5 \
    --cache-dataset \
    --model mc3_18 --weights="MC3_18_Weights.DEFAULT" 
#    --model r3d_18 --weights="R3D_18_Weights.DEFAULT" 
#    --model r2plus1d_18 --weights="R2Plus1D_18_Weights.DEFAULT" 
```
Results:
```
MC3_18:
Test: Total time: 0:11:17
 * Clip Acc@1 52.813 Clip Acc@5 75.220
 * Video Acc@1 63.960 Video Acc@5 84.130

R3D_18:
Test: Total time: 0:11:24
 * Clip Acc@1 51.795 Clip Acc@5 74.365
 * Video Acc@1 63.200 Video Acc@5 83.479

R2Plus1D_18:
Test: Total time: 0:11:26
 * Clip Acc@1 56.517 Clip Acc@5 77.739
 * Video Acc@1 67.463 Video Acc@5 86.175
```


datumbox (commented):
Few early comments, I understand it's Draft so feel free to ignore.

references/video_classification/train.py
```diff
@@ -313,10 +332,10 @@ def main(args):
     print(f"Training time {total_time_str}")
 
 
-def parse_args():
+def get_args_parser(add_help=True):
```
YosuaMichael, line 335: We change this so it become more similar to classification script and it is more compatible with `submitit` script

datumbox: Good call.

references/video_classification/train.py
```diff
@@ -52,13 +53,25 @@ def evaluate(model, criterion, data_loader, device):
...
     with torch.inference_mode():
-        for video, target in metric_logger.log_every(data_loader, 100, header):
+        for video, target, video_idx in metric_logger.log_every(data_loader, 100, header):
             video = video.to(device, non_blocking=True)
             target = target.to(device, non_blocking=True)
             output = model(video)
             loss = criterion(output, target)
 
+            # Use softmax to convert output into prediction probability
+            preds = nn.Softmax(dim=1)(output)
```
datumbox, line 69: Let's avoid generating a class in a for loop and use the functional equivalent.
```suggestion
            preds = torch.softmax(output, dim=1)
```

references/video_classification/train.py
```diff
@@ -95,6 +108,11 @@ def evaluate(model, criterion, data_loader, device):
             top1=metric_logger.acc1, top5=metric_logger.acc5
         )
     )
+    # Reduce the agg_preds and agg_targets from all gpu and show result
+    agg_preds = utils.reduce_across_processes(agg_preds)
+    agg_targets = utils.reduce_across_processes(agg_targets, op=torch.distributed.ReduceOp.MAX)
+    agg_acc1, agg_acc5 = utils.accuracy(agg_preds, agg_targets, topk=(1, 5))
+    print(" * Video Acc@1 {acc1:.3f} Video Acc@5 {acc5:.3f}".format(acc1=agg_acc1, acc5=agg_acc5))
```
datumbox, line 115: Not the best place to put the print statement but we can keep it. Worth improving the situation on the future.

datumbox (commented):
@YosuaMichael amazing work. I have only 1 minor suggestion, see below.

Can you please provide proof of the new accuracies for all video models (MViT + Video ResNet)? We should update their meta-data stats.

references/video_classification/train.py
```diff
@@ -95,6 +108,11 @@ def evaluate(model, criterion, data_loader, device):
             top1=metric_logger.acc1, top5=metric_logger.acc5
```
YosuaMichael, line 108: @datumbox For the print I follow the format like this one. And the original print is saying `Clip Acc@1 ...` and for the new one I change `Clip` into `Video` below. However this might be quite confusing. Do you have good idea for the accuracy name here?

datumbox:
@YosuaMichael thanks for providing the stats. My recommendation is to provide the exact commands you used to verifying the models + the raw output of accuracies; this is going to highlight that specific models must be run with specific configurations (for example the frame rate on MViT). The process I describe here is identical with how we deploy models. I would also update the documentation + meta-data to show-case how the models should be verified (simple links will do).

This can happen of a follow up but given the impact of this update, I would advise having all on a single PR that explains the changes in one place.

references/video_classification/train.py
datumbox: I agree that printing 2 stats could be confusing. The thing is, even the "Video" one is clips; they are just aggregated all together instead on per video basis. One way around this would be to make your changes directly on the metric_logger but this would require additional checks. For now I think it's OK to show both; this can be reviewed on follow up PRs. Concerning the naming, the technique seems to be doing an ensemble of the clips, so you could name it as such. I would definitely not call it Clip vs Video because they are both based on clips.

torchvision/models/video/resnet.py
```diff
@@ -313,6 +313,9 @@ def _video_resnet(
     "categories": _KINETICS400_CATEGORIES,
     "recipe": "https://github.com/pytorch/vision/tree/main/references/video_classification",
     "_docs": """These weights reproduce closely the accuracy of the paper for 16-frame clip inputs.""",
+    "_frame_rate": 8,
+    "_clips_per_video": 5,
+    "_clip_len": 16,
```
datumbox, line 318: The extended tests are failing because those are not part of the schema. You can add them but then we need to review the names. First of all, I agree that all of them should be included here. I'm not sure about the names or structure and I don't have off the top of my head a good recommendation. Let's chat about this tomorrow and solicit feedback from others.

YosuaMichael: As discussed offline, for now we will put this param on the _docs, and we can have a follow up PR for adding this evaluation parameters as meta.

torchvision/models/video/mvit.py
```diff
@@ -445,12 +445,15 @@ class MViT_V1_B_Weights(WeightsEnum):
             "min_temporal_size": 16,
             "categories": _KINETICS400_CATEGORIES,
             "recipe": "https://github.com/facebookresearch/pytorchvideo/blob/main/docs/source/model_zoo.md",
-            "_docs": """These weights support 16-frame clip inputs and were ported from the paper.""",
+            "_docs": (
+                "These weights were ported from the paper and it is evaluated on video level accuracy "
+                "with parameters frame_rate=7.5, clips_per_video=5, and clip_len=16"
```
datumbox, line 450:
```suggestion
                "The weights were ported from the paper. The accuracies are estimated on video-level "
                "with parameters `frame_rate=7.5`, `clips_per_video=5` and `clip_len=16`."
```

torchvision/models/video/resnet.py
```diff
@@ -312,7 +312,10 @@ def _video_resnet(
     "min_size": (1, 1),
     "categories": _KINETICS400_CATEGORIES,
     "recipe": "https://github.com/pytorch/vision/tree/main/references/video_classification",
-    "_docs": """These weights reproduce closely the accuracy of the paper for 16-frame clip inputs.""",
+    "_docs": (
+        "These weights reproduce closely the video accuracy of the paper for 16-frame clip inputs. "
+        "To produce the metrics we use frame_rate=8, clips_per_video=5, and clip_len=16 as parameters."
```
datumbox, line 317:
```suggestion
        "These weights reproduce closely the accuracy of the paper. The accuracies are estimated on video-level "
        "with parameters `frame_rate=15`, `clips_per_video=5` and `clip_len=16`."
```

datumbox (approved):
@YosuaMichael LGTM, thanks! Just a few minor suggestions on the documentation (plus a correction of the hyper-param on ResNets) but other than this it looks good.

PR #6243, opened 2022-07-07
Update clips/sec for average value

puneeshkhanna:
Update clips/sec for average value over last 10 steps according to window size rather than instantaneous value of the last step.


datumbox (commented):
@puneeshkhanna Thanks for the contribution.

I'm not certain about the benefit of this change. This is the only metric that is updated to present the average and the change misaligns the output across other scripts.

puneeshkhanna:
> @puneeshkhanna Thanks for the contribution.
> 
> I'm not certain about the benefit of this change. This is the only metric that is updated to present the average and the change misaligns the output across other scripts.

@datumbox - why would we have window size of 10 if we just need to deque last value only. By default when fmt is None, all metrics return median and global avg. I really think that returning immediate value only is not a true estimate of performance. Maybe we should rethink this strategy across other scripts too.

PR #6245, opened 2022-07-07
[proto] Added some transformations and fixed type hints

vfdev-5:
Description:
- Updated and added Transforms
- removed cutmix/mixup from functional part and put the code into Transforms
- Changed `padding` arg type hint
- Changed `fill` arg type hint
- ~Changed `fill` arg type hint, default from None to 0~
- code formatting in tests


torchvision/prototype/transforms/_augment.py
```diff
@@ -110,15 +114,32 @@ def forward(self, *inpts: Any) -> Any:
...
 class RandomMixup(_BaseMixupCutmix):
     def _get_params(self, sample: Any) -> Dict[str, Any]:
         return dict(lam=float(self._dist.sample(())))
 
     def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
-        if isinstance(inpt, features._Feature):
-            return inpt.mixup(**params)
-        else:
-            return inpt
+        lam = params["lam"]
```
vfdev-5, line 130: @datumbox I removed mixup/cutmix from functional part and put it into Transforms. Is it that what you had in mind ?

torchvision/prototype/features/_image.py
```diff
@@ -183,11 +190,19 @@ def rotate(
...
+        fill: Union[int, float, Sequence[int], Sequence[float]] = 0,
         center: Optional[List[float]] = None,
     ) -> Image:
         from torchvision.prototype.transforms import functional as _F
 
+        # This cast does Sequence -> List[float] to please mypy and torch.jit.script
+        if not isinstance(fill, (int, float)):
+            fill = [float(v) for v in list(fill)]
+
+        if isinstance(fill, (int, float)):
```
datumbox, line 202: Nit: Use `else` here? Similar in other places.

torchvision/prototype/transforms/_augment.py
```diff
@@ -110,15 +114,32 @@ def forward(self, *inpts: Any) -> Any:
         return super().forward(sample)
 
 
+def _mixup_onehotlabel(inpt: features.OneHotLabel, lam: float) -> features.OneHotLabel:
```
datumbox, line 117: Nit: shouldn't this be part of the `_BaseMixupCutmix` class?

torchvision/prototype/transforms/_augment.py
datumbox: Yes, looks good.

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -66,11 +69,11 @@ def fn(
 
     def _parse_fill(
         self, image: Union[PIL.Image.Image, torch.Tensor, features.Image], num_channels: int
-    ) -> Optional[List[float]]:
+    ) -> Union[int, float, Sequence[int], Sequence[float]]:
         fill = self.fill
 
-        if isinstance(image, PIL.Image.Image) or fill is None:
-            return fill
+        # if isinstance(image, PIL.Image.Image) or fill is None:
```
datumbox, line 75: Nit: seems like we don't need `image` any more here.

torchvision/prototype/transforms/_color.py
```diff
@@ -55,74 +54,52 @@ def _check_input(
...
         **kwargs: Any,
     ) -> T:
-        if isinstance(input, features.Image):
-            output = kernel_tensor(input, **kwargs)
-            return features.Image.new_like(input, output)
-        elif is_simple_tensor(input):
-            return kernel_tensor(input, **kwargs)
-        elif isinstance(input, PIL.Image.Image):
-            return kernel_pil(input, **kwargs)  # type: ignore[no-any-return]
+        if isinstance(inpt, features.Image):
```
datumbox, line 63: Nit: this looks like a remnant method that is not longer used from the time we didn't have a middle layer. I think it should be removed and we need to do a quick review and clean up of any similar logic from the codebase after we merge this.

torchvision/prototype/transforms/_color.py
```diff
@@ -175,33 +152,25 @@ def _get_params(self, sample: Any) -> Dict[str, Any]:
...
-            input = self._saturation(input)
+            inpt = self._saturation(inpt)
         if params["contrast2"] and not params["contrast_before"]:
-            input = self._contrast(input)
+            inpt = self._contrast(inpt)
         if params["channel_shuffle"]:
-            input = self._channel_shuffle(input)
-        return input
+            inpt = self._channel_shuffle(inpt)
+        return inpt
```
datumbox, line 168: When possible for future PRs, do such changes on clean up PRs. This will reduce the LOC size for big PRs like this. No need to take action now.

torchvision/prototype/transforms/_geometry.py
```diff
@@ -125,20 +80,16 @@ def __init__(
...
         self.size = size
         self.scale = scale
         self.ratio = ratio
         self.interpolation = interpolation
+        self.antialias = antialias
 
     def _get_params(self, sample: Any) -> Dict[str, Any]:
+        # vfdev-5: techically, this op can work on bboxes/segm masks only inputs without image in samples
+        # What if we have multiple images/bboxes/masks of different sizes ?
+        # TODO: let's support bbox or mask in samples without image
```
datumbox, line 92: Correct. This logic was added before you land kernels for handling crops BBoxes and masks. Now we can review this transform. It's worth noting that passing bboxes or masks without image doesn't make a lot of sense in practice because this will cause desyncing between the image and the targets. This is something that was brought up many times as a potential feature (passing only bboxes) but so far we haven't found a compelling example on where this can be used in practice. Happy to review this if we do, but for now we can assume that an image (or video coming up) will be available to give us access to the width/height here.

torchvision/prototype/transforms/_geometry.py
```diff
@@ -309,61 +250,28 @@ def __init__(
...
-        if fill is None:
-            fill = 0.0
+        _check_fill_arg(fill)
         self.fill = fill
 
         self.side_range = side_range
         if side_range[0] < 1.0 or side_range[0] > side_range[1]:
             raise ValueError(f"Invalid canvas side range provided {side_range}.")
 
+        self._pad_op = Pad(0, padding_mode="constant")
```
datumbox, line 273: I would probably avoid doing this. It puts an idiom of using transforms within transforms which is probably unnecessary now that we have a middle layer. This was added originally by Philip to avoid duplicating massive amount of code from Pad. There was also the idea of having different primitive non-random Transform classes which we are currently moving away from.

torchvision/transforms/functional_tensor.py
```diff
@@ -383,8 +385,13 @@ def pad(img: Tensor, padding: List[int], fill: Union[int, float] = 0, padding_mo
     if isinstance(padding, tuple):
         padding = list(padding)
 
-    if isinstance(padding, list) and len(padding) not in [1, 2, 4]:
-        raise ValueError(f"Padding must be an int or a 1, 2, or 4 element tuple, not a {len(padding)} element tuple")
+    if isinstance(padding, list):
+        # TODO: Jit is failing on loading this op when scripted and saved
```
datumbox, line 389: What's the plan for this?

datumbox (commented):
Thanks @vfdev-5 for the PR. I've added a few comments, let me know your thoughts.

torchvision/prototype/features/_bounding_box.py
```diff
@@ -128,13 +128,20 @@ def resized_crop(
...
+        fill: Union[int, float, Sequence[int], Sequence[float]] = 0,
+        padding_mode: str = "constant",
     ) -> BoundingBox:
         from torchvision.prototype.transforms import functional as _F
 
         if padding_mode not in ["constant"]:
             raise ValueError(f"Padding mode '{padding_mode}' is not supported with bounding boxes")
 
+        # This cast does Sequence[int] -> List[int] and is required to make mypy happy
+        if not isinstance(padding, int):
```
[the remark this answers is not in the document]
vfdev-5: Padding can be int only

torchvision/prototype/features/_feature.py
```diff
@@ -120,7 +120,10 @@ def resized_crop(
         return self
 
     def pad(
-        self, padding: List[int], fill: Union[int, float, Sequence[float]] = 0, padding_mode: str = "constant"
+        self,
+        padding: Union[int, Sequence[int]],
+        fill: Union[int, float, Sequence[int], Sequence[float]] = 0,
+        padding_mode: str = "constant",
```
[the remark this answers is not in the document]
vfdev-5: Yes, we can do that in another PR. I'm OK with both, but I agree that it would be a bit more coherent with the codebase, e.g interpolation mode.

torchvision/prototype/features/_bounding_box.py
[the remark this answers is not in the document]
datumbox: Lol, sorry I misread the code; I thought it was fill. Sounds good.

torchvision/prototype/features/_feature.py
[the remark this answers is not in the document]
datumbox: Let's add a TODO and tackle it later.

torchvision/prototype/features/_bounding_box.py
```diff
@@ -128,13 +128,20 @@ def resized_crop(
         return BoundingBox.new_like(self, output, image_size=image_size, dtype=output.dtype)
 
     def pad(
-        self, padding: List[int], fill: Union[int, float, Sequence[float]] = 0, padding_mode: str = "constant"
+        self,
+        padding: Union[int, Sequence[int]],
+        fill: Union[int, float, Sequence[int], Sequence[float]] = 0,
```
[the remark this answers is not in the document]
vfdev-5: > Are there any BC/JIT considerations by changing the default value and type?

yes, I do not touch any stable public/private API with this except: https://github.com/pytorch/vision/pull/6245/files#diff-efb615f437f0b00ef2ed8abf2e0eda487cf609899a115bf3fa4691100db581afR307

Right now in the stable API, we have the following for fill default value:
```
-- Stable Transforms:

RandomAffine:
- fill=0

ElasticTransform:
- fill=0

RandomRotation:
- fill=0

RandomPerspective:
- fill=0

Pad:
- fill=0


-- Stable functional:

affine:
- fill: Optional[List[float]] = None,
-- pil: fill: Optional[Union[float, List[float], Tuple[float, ...]]] = 0,
-- t: fill: Optional[List[float]] = None

elastic_transform:
- fill: Optional[List[float]] = None,

rotate:
- fill: Optional[List[float]] = None,
-- pil: fill: Optional[Union[float, List[float], Tuple[float, ...]]] = 0,
-- t: fill: Optional[List[float]] = None,

perspective:
- fill: Optional[List[float]] = None,

pad:
- fill: Union[int, float] = 0
-- pil: fill: Optional[Union[float, List[float], Tuple[float, ...]]] = 0,
-- t: fill: Union[int, float] = 0,
```

We have two values: 0 and None. Supported fill types are numbers and sequences of numbers: affine, rotate started to support sequences about 1-2 years now and pad has a workaround introduced in prototype to support sequences.
Inspecting the code I do not see why keeping None is useful as fill is directly reassigned to 0.

> I understand that this layer is not JIT scriptable but the change on low-level kernels will expose differences to the users.

Yes, there is a change of default value from None -> 0 in low-level ops in proto which will become a replacement for current API. IMO, the change is minor and I think there is rare user who calls `pad(...., fill=None)`. If we agree that the change is necessary for code coherence etc I'm OK to add `Optional` and check for None and raise a warning about change. Otherwise, please also let me know if you want to revert default value change.

torchvision/prototype/transforms/_augment.py
vfdev-5: OK, let me include it

torchvision/prototype/transforms/_auto_augment.py
vfdev-5: Yes, I'll remove this code once we agreed on `fill=None` or `fill=0`. Probably, if we go deprecation cycle I had to get this code back

torchvision/transforms/functional_tensor.py
vfdev-5: Let's wait if this will be fixed on pytorch side and revert this into a single if condition

torchvision/prototype/transforms/_color.py
vfdev-5: Thanks for catching that !

torchvision/prototype/features/_feature.py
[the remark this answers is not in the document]
vfdev-5: I open an issue for that: https://github.com/pytorch/vision/issues/6250

torchvision/prototype/transforms/_geometry.py
vfdev-5: Not a practical use-case, but we can't test this transform on bboxes/masks without an image :)

datumbox: I'm curious to see if this creates issues such as having no access to specific meta-data you would need from the image. If you can do this without them, we can chat adding it. But if you have to add lots of workarounds to support it, we should find a real-world use-case. I think the test use-case can be achieved by passing an equivalent image.

torchvision/prototype/features/_bounding_box.py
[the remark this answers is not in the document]
datumbox: From your response, I understand that the only challenge is for the users of the public `F` interface who currently pass None. If we are able to handle that in a BC way (by replacing None with 0 within the method), then I don't see any blocker. As long as we don't break any BC on public APIs, I'm in favour of aligning the calls. Do I miss anything?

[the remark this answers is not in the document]
vfdev-5: Restored default fill value to None

torchvision/prototype/transforms/_augment.py
```diff
@@ -86,15 +86,19 @@ def _get_params(self, sample: Any) -> Dict[str, Any]:
...
+                return features.Image.new_like(inpt, output)
+            return output
         elif isinstance(inpt, PIL.Image.Image):
             # TODO: We should implement a fallback to tensor, like gaussian_blur etc
             raise RuntimeError("Not implemented")
         elif isinstance(inpt, torch.Tensor):
             return F.erase_image_tensor(inpt, **params)
-        else:
-            return inpt
+        raise TypeError(
```
datumbox, line 99: This means we no longer support meta-data such as integers, strings etc. Where else did we make this change? We probably need to undo it. Here we should check for bboxes and masks and throw an error. Labels should be permitted and returned as-is.

torchvision/prototype/transforms/_augment.py
```diff
@@ -109,16 +113,32 @@ def forward(self, *inpts: Any) -> Any:
...
+                raise ValueError("Need a batch of images")
+            output = inpt.clone()
+            output = output.roll(1, -4).mul_(1 - lam).add_(output.mul_(lam))
+            return features.Image.new_like(inpt, output)
+        if isinstance(inpt, features.OneHotLabel):
+            return self._mixup_onehotlabel(inpt, lam)
+
+        raise TypeError(
+            "RandomMixup transformation does not support bounding boxes, segmentation masks and plain labels"
+        )
```
datumbox, line 141: Here BBox and Masks shouldn't be supported, but everything else should be allowed (meta-data).

datumbox (approved):
LGTM, as discussed offline the `has_all()` and `has_any()` methods need to change to support some additional requirement but this can happen on a separate PR.

vfdev-5:
I'll merge this PR to main and then work with proto-transforms-api branch

PR #6246, opened 2022-07-07
Add SwinV2

### LINKED ISSUE #6242: Add SwinV2 in TorchVision
### 🚀 The feature

Swin Transforms were added to TorchVision on #5491 thanks to the work of @xiaohu2015 and @jdsgomes. At that point, the official code of [SwinV2](https://arxiv.org/abs/2111.09883) was not released so we decided to postpone its addition. Later @YosuaMichael has refactored the class to make it compatible with 3d use-cases at #6088. With the code of V2 now [available](https://github.com/microsoft/Swin-Transformer/blob/main/models/swin_transformer_v2.py), we should consider adding it to TorchVision.

Here is what needs to be done:
- Examine if it's possible to extend the existing [Swin Transformer](https://github.com/pytorch/vision/blob/main/torchvision/models/swin_transformer.py) class to support both V1 and V2 models. An example of such an extension was recently done for [EfficientNets](https://github.com/pytorch/vision/blob/main/torchvision/models/efficientnet.py). On the other hand, if the implementations are substantially different, we can follow the approach we took for MobileNets [v2](https://github.com/pytorch/vision/blob/main/torchvision/models/mobilenetv2.py) and [v3](https://github.com/pytorch/vision/blob/main/torchvision/models/mobilenetv3.py) and use separate classes.
- Provide an implementation that uses TorchVision building blocks and idioms. See #5319 for pointers.
- Provide proof in form of pre-trained weights for the smallest variant that reproduces the accuracy of the paper.

### Motivation, pitch

We should keep TorchVision fresh by offering the latest influential architectures to support the work of ML researchers and practitioners.

### Alternatives

_No response_

### Additional context

_No response_

cc @datumbox

ain-soph:
Fixes #6242


ain-soph:
And I see an issue from the Microsoft SwinTransformer repo: https://github.com/microsoft/Swin-Transformer/issues/194

It thinks it's not necessary to divide the mask into 9 parts because 4 parts are already enough.  
I kind of agree with that. Anyone has opinion on that?

ain-soph:
According to the unittest logs, the state_dict has certain `int` values inside? I currently have no clue about this actually.

https://app.circleci.com/pipelines/github/pytorch/vision/18731/workflows/91dedbbe-264b-4665-a146-91117a6c8f63/jobs/1515479?invite=true#step-108-2975

ain-soph:
It seems these mod operations are considered as nodes of the fx graph, and their output is `float` rather than `torch.Tensor` (actually why is it `float` rather than `int`?)

image

datumbox:
Concerning the linter, if you check the [CI](https://app.circleci.com/pipelines/github/pytorch/vision/18731/workflows/38887801-2a22-478d-9528-6fdf95b89d4a/jobs/1515423?invite=true#step-104-12) on tab `Required lint modifications` it will show you what's the problem:
image

Concerning your point the PatchMerging, I believe you are right. FX doesn't let you use the input of the tensor to control the flow of the program. This needs to move outside of the main function and be declared a non fx-traceable operator. I would patch this ASAP outside of this PR. cc @YosuaMichael

torchvision/models/swin_transformer.py
```diff
@@ -239,6 +245,221 @@ def forward(self, x: Tensor):
...
+        qkv_bias[length : 2 * length].zero_()
+    qkv = F.linear(x, qkv_weight, qkv_bias)
+    qkv = qkv.reshape(x.size(0), x.size(1), 3, num_heads, -1).permute(2, 0, 3, 1, 4)
+    q, k, v = qkv[0], qkv[1], qkv[2]
+    # v2 cosine attention
+    attn = F.normalize(q, dim=-1) @ F.normalize(k, dim=-1).transpose(-2, -1)
+    logit_scale = torch.clamp(logit_scale, max=torch.log(torch.tensor(1.0 / 0.01))).exp()
+    attn = attn * logit_scale
+    # add relative position bias
+    attn = attn + relative_position_bias
```
datumbox, line 315: Massive amount of code duplication in this method comparing to v1. Everything else is the same except:
image

I recommend refactoring the original method to avoid introducing so much duplicate code. We need to be mindful for the 3d case that we added at https://github.com/facebookresearch/multimodal/pull/43. cc @YosuaMichael

torchvision/models/swin_transformer.py
```diff
@@ -239,6 +245,221 @@ def forward(self, x: Tensor):
...
+
+    # unpad features
+    x = x[:, :H, :W, :].contiguous()
+    return x
+
+
+torch.fx.wrap("shifted_window_attention_v2")
+
+
+class ShiftedWindowAttentionV2(nn.Module):
```
datumbox, line 358: This class seems to be justified to be separate from its V1 counter part. I think it's still worth examining if we can share some of the code from the original.

@YosuaMichael any thoughts on this?

torchvision/models/swin_transformer.py
```diff
@@ -357,7 +596,7 @@ def __init__(
         # build SwinTransformer blocks
         for i_stage in range(len(depths)):
             stage: List[nn.Module] = []
-            dim = embed_dim * 2 ** i_stage
+            dim = embed_dim * 2**i_stage
```
datumbox, line 599: Nit: Undo this. This is why your linter breaks.

torchvision/models/swin_transformer.py
```diff
@@ -506,6 +748,75 @@ class Swin_B_Weights(WeightsEnum):
...
+            "_metrics": {
+                "ImageNet-1K": {
+                    "acc@1": 0.100,  # TODO
+                    "acc@5": 0.100,  # TODO
+                }
+            },
+            "_docs": """These weights reproduce closely the results of the paper using a similar training recipe.""",
+        },
+    )
+    DEFAULT = IMAGENET1K_V1
```
datumbox, line 771:
```suggestion
    pass
```

You can add empty weights for now.

torchvision/models/swin_transformer.py
```diff
@@ -506,6 +748,75 @@ class Swin_B_Weights(WeightsEnum):
...
```
datumbox, line 817: Keep only one variant for now. Tiny will do. This will reduce the LOC of this PR. Things can be added later.
```suggestion
```

torchvision/models/swin_transformer.py
```diff
@@ -608,6 +921,118 @@ def swin_b(*, weights: Optional[Swin_B_Weights] = None, progress: bool = True, *
...
+        window_size=[7, 7],
+        stochastic_depth_prob=0.2,
+        version=2,
+        weights=weights,
+        progress=progress,
+        **kwargs,
+    )
+
+
+def swin_v2_s(*, weights: Optional[Swin_V2_S_Weights] = None, progress: bool = True, **kwargs: Any) -> SwinTransformer:
```
datumbox, line 968: Same here, remove the `swin_v2_s` and `swin_v2_b` variants for now.

torchvision/models/swin_transformer.py
```diff
@@ -266,11 +488,20 @@ def __init__(
         attention_dropout: float = 0.0,
         stochastic_depth_prob: float = 0.0,
         norm_layer: Callable[..., nn.Module] = nn.LayerNorm,
-        attn_layer: Callable[..., nn.Module] = ShiftedWindowAttention,
+        attn_layer: Optional[Callable[..., nn.Module]] = None,
+        version: int = 1,
```
datumbox, line 492: Undo this change.
```suggestion
        attn_layer: Callable[..., nn.Module] = ShiftedWindowAttention,
```

We shouldn't pass version. Instead we can pass a boolean flag that allows doing pre_normalization (better name can be used here). This flag will be explosed to the model builders and they will be responsible for passing the right `att_layer` and bool value to construct the network for V1 and V2. This will align your code to the idioms we used at Vision.

datumbox (commented):
@ain-soph The PR looks in the right direction. I've added a few comments for your attention, please have a look.

Note that I'll be on PTO the next 2 weeks so @YosuaMichael offered to provide assistance. He is also a Meta maintainer so you can schedule a call if needed to speed this up.

Please note that my comments don't address the validity of the changes you do on the ML side but mostly on the idioms and coding practices we use at TorchVision. A more deep dive check would be necessary to confirm validity. The first step I think is to take the original pre-trained weights from the paper and confirm that we can load them in your implementation (by making some conversion) and then reproduce the reported accuracy using our reference scripts. Once we verify this, the next step is to train the tiny variant to show-case we can reproduce the accuracy. If you don't have access to a GPU cluster, we do so we can help.

Again many thanks for your contribution and help on this. 

cc @xiaohu2015 and @jdsgomes who worked on the original Swin implementation to see if they have any early feedback.

YosuaMichael:
Hi @ain-soph, thanks a lot for the PR! 
As of now, I am still reading SwinTransformerV2 paper and original code, and I will try to review afterwards.

Meanwhile, let me address some of the issue you raise:

1. On ufmt issue, can you make sure you install the following version: `pip install ufmt==1.3.2 black==21.9b0 usort==0.6.4` ([reference](https://github.com/pytorch/vision/blob/main/CONTRIBUTING.md#code-formatting-and-typing))
2. For the fx issue, I create a small patch: https://github.com/pytorch/vision/pull/6252

torchvision/models/swin_transformer.py
YosuaMichael: I agree that having separate class from `V1` is justified here. 
I think the code for generating `relative_position_index` can be shared (we can have a separate function for this)

torchvision/models/swin_transformer.py
```diff
@@ -239,6 +245,217 @@ def forward(self, x: Tensor):
...
+        coords_h = torch.arange(self.window_size[0])
+        coords_w = torch.arange(self.window_size[1])
+        coords = torch.stack(torch.meshgrid([coords_h, coords_w]))  # 2, Wh, Ww
+        coords_flatten = torch.flatten(coords, 1)  # 2, Wh*Ww
+        relative_coords = coords_flatten[:, :, None] - coords_flatten[:, None, :]  # 2, Wh*Ww, Wh*Ww
+        relative_coords = relative_coords.permute(1, 2, 0).contiguous()  # Wh*Ww, Wh*Ww, 2
+        relative_coords[:, :, 0] += self.window_size[0] - 1  # shift to start from 0
+        relative_coords[:, :, 1] += self.window_size[1] - 1
+        relative_coords[:, :, 0] *= 2 * self.window_size[1] - 1
+        relative_position_index = relative_coords.sum(-1)  # Wh*Ww, Wh*Ww
```
YosuaMichael, line 416: I think we can refactor and create a function for generating `relative_position_index` and this function can be shared among `V1` and `V2` (from line 406 to 416).

And in this specific line, I think it should be okay to use same one as V1 by having additional `.view(-1)` in the constructor and remove the `.view(-1)` on the `forward`.

torchvision/models/swin_transformer.py
```diff
@@ -239,6 +245,217 @@ def forward(self, x: Tensor):
...
+        self,
+        dim: int,
+        window_size: List[int],
+        shift_size: List[int],
+        num_heads: int,
+        qkv_bias: bool = True,
+        proj_bias: bool = True,
+        attention_dropout: float = 0.0,
+        dropout: float = 0.0,
+        pretrained_window_size: Tuple[int, int] = (0, 0),
```
YosuaMichael, line 373: Can we use `List[int]` as the type for `pretrained_window_size` similar to `window_size` and `shift_size`. We use `List` so it can have same api as the 3D version of swin transformer.

YosuaMichael (commented):
Hi @ain-soph , overall the PR looks good!
I agree with @datumbox comments that we may want to refactor some code so it can be reused in both `V1` and `V2` and I add some suggestion here.

torchvision/models/swin_transformer.py
ain-soph: Sure!

ain-soph: I previously use Tuple to avoid the in-place modification.

torchvision/models/swin_transformer.py
ain-soph: Would it be better to use `.flatten()` rather than `.view(-1)` ?

YosuaMichael: I think `.flatten()` and `.view(-1)` will do similar things here. I suggested using `.view(-1)` because it is what is used in `V1`. Also, what you use in [forward](https://github.com/pytorch/vision/blob/6e488d535a63d27af3ca8f063821c033b2e9521c/torchvision/models/swin_transformer.py#L437) function.

torchvision/models/swin_transformer.py
```diff
@@ -214,24 +235,127 @@ def __init__(
...
+        shift_size: List[int],
+        num_heads: int,
+        qkv_bias: bool = True,
+        proj_bias: bool = True,
+        attention_dropout: float = 0.0,
+        dropout: float = 0.0,
+        pretrained_window_size: List[int] = [0, 0],
+    ):
+        self.pretrained_window_size = pretrained_window_size  # TODO: unsafe, need copy?
+        super().__init__(
```
ain-soph, line 293: Shall we embed all those unmodified parameters in `**kwargs`? Or leave them as explicit parameters?

torchvision/models/swin_transformer.py
```diff
@@ -214,24 +235,127 @@ def __init__(
...
+        relative_coords = coords_flatten[:, :, None] - coords_flatten[:, None, :]  # 2, Wh*Ww, Wh*Ww
+        relative_coords = relative_coords.permute(1, 2, 0).contiguous()  # Wh*Ww, Wh*Ww, 2
+        relative_coords[:, :, 0] += self.window_size[0] - 1  # shift to start from 0
+        relative_coords[:, :, 1] += self.window_size[1] - 1
+        relative_coords[:, :, 0] *= 2 * self.window_size[1] - 1
+        relative_position_index = relative_coords.sum(-1).flatten()  # Wh*Ww*Wh*Ww
+        self.register_buffer("relative_position_index", relative_position_index)
+
+    def get_relative_position_bias(self) -> torch.Tensor:
+        relative_position_bias_table: torch.Tensor = self.cpb_mlp(self.relative_coords_table).view(-1, self.num_heads)
```
ain-soph, line 346: Shall we use `flatten(end_dim=-2)` here?

ain-soph:
I'm currently using `v2_logit_scale` and `use_v2` as 2 variables to check versions. Please comment if you have any better solution.

xiaohu2015:
@ain-soph Thanks for you good PR. I have two advice:
- we should fix the problem in  https://github.com/pytorch/vision/pull/6222
- I think we should also add the `pretrained_window_size` on the `SwinTransformer`, I think it should be `List[List[int]]`

ain-soph:
@xiaohu2015 I'll take a look at the problem you mentioned here, and add the `pretrained_window_size` as the argument (it's zero for all variants though). And I think it should be `List[int]`? (according to the original code in https://github.com/microsoft/Swin-Transformer/blob/b720b4191588c19222ccf129860e905fb02373a7/models/swin_transformer_v2.py#L526)

xiaohu2015:
> @xiaohu2015 I'll take a look at the problem you mentioned here, and add the `pretrained_window_size` as the argument (it's zero for all variants though). And I think it should be `List[int]`? (according to the original code in https://github.com/microsoft/Swin-Transformer/blob/b720b4191588c19222ccf129860e905fb02373a7/models/swin_transformer_v2.py#L526)

In torchvision, we define the window size as `List[int]`, it means the window size of each dimension. The `Tuple[int]` in offical code means the window size of each block.

torchvision/models/swin_transformer.py
ain-soph: And btw, for L292, it's not safe to set List as a default argument and make it an attribute because it's mutable.

ain-soph:
@xiaohu2015 I just read the `window_size` problem you mentioned and agree that the current implementation has some potential risk at some edge cases.

But I think it's a little out of scope and we should make another issue to get that solved.  
I'm following the current V1 architecture.  
If we plan to pass `image_size` as argument of `Transformer` model in the future to limit window size, we can definitely do that in a new PR.

ain-soph:
The evaluation log of Swin_V2_T using pretrained weights from official repo:

```
torchrun --nproc_per_node=4 train.py --model swin_v2_t --test-only --weights Swin_V2_T_Weights.IMAGENET1K_V1
WARNING:torch.distributed.run:
*****************************************
Setting OMP_NUM_THREADS environment variable for each process to be 1 in default, to avoid your system being overloaded, please further tune the variable for optimal performance in your application as needed.
*****************************************
| distributed init (rank 3): env://
| distributed init (rank 1): env://
| distributed init (rank 0): env://
| distributed init (rank 2): env://
Namespace(data_path='/data/user/data/image/imagenet', model='swin_v2_t', device='cuda', batch_size=32, epochs=90, workers=16, opt='sgd', lr=0.1, momentum=0.9, weight_decay=0.0001, norm_weight_decay=None, bias_weight_decay=None, transformer_embedding_decay=None, label_smoothing=0.0, mixup_alpha=0.0, cutmix_alpha=0.0, lr_scheduler='steplr', lr_warmup_epochs=0, lr_warmup_method='constant', lr_warmup_decay=0.01, lr_step_size=30, lr_gamma=0.1, lr_min=0.0, print_freq=10, output_dir='.', resume='', start_epoch=0, cache_dataset=False, sync_bn=False, test_only=True, auto_augment=None, random_erase=0.0, amp=False, world_size=4, dist_url='env://', model_ema=False, model_ema_steps=32, model_ema_decay=0.99998, use_deterministic_algorithms=False, interpolation='bilinear', val_resize_size=256, val_crop_size=224, train_crop_size=224, clip_grad_norm=None, ra_sampler=False, ra_reps=3, weights='Swin_V2_T_Weights.IMAGENET1K_V1', rank=0, gpu=0, distributed=True, dist_backend='nccl')
Loading data
Loading training data
Took 3.2938268184661865
Loading validation data
Creating data loaders
Creating model
/home/user/miniconda3/envs/dev/lib/python3.10/site-packages/torch/nn/parallel/scatter_gather.py:9: UserWarning: is_namedtuple is deprecated, please use the python checks instead
  warnings.warn("is_namedtuple is deprecated, please use the python checks instead")
/home/user/miniconda3/envs/dev/lib/python3.10/site-packages/torch/nn/parallel/scatter_gather.py:9: UserWarning: is_namedtuple is deprecated, please use the python checks instead
  warnings.warn("is_namedtuple is deprecated, please use the python checks instead")
/home/user/miniconda3/envs/dev/lib/python3.10/site-packages/torch/nn/parallel/scatter_gather.py:9: UserWarning: is_namedtuple is deprecated, please use the python checks instead
  warnings.warn("is_namedtuple is deprecated, please use the python checks instead")
/home/user/miniconda3/envs/dev/lib/python3.10/site-packages/torch/nn/parallel/scatter_gather.py:9: UserWarning: is_namedtuple is deprecated, please use the python checks instead
  warnings.warn("is_namedtuple is deprecated, please use the python checks instead")
Test:   [  0/391]  eta: 0:28:16  loss: 0.2255 (0.2255)  acc1: 96.8750 (96.8750)  acc5: 100.0000 (100.0000)  time: 4.3390  data: 2.0224  max mem: 817
Test:   [100/391]  eta: 0:01:28  loss: 0.6355 (0.6686)  acc1: 87.5000 (86.2314)  acc5: 96.8750 (97.4938)  time: 0.2590  data: 0.1852  max mem: 817
Test:   [200/391]  eta: 0:00:57  loss: 1.1148 (0.7348)  acc1: 71.8750 (84.2973)  acc5: 93.7500 (97.0771)  time: 0.2569  data: 0.1827  max mem: 817
Test:   [300/391]  eta: 0:00:27  loss: 0.9935 (0.8109)  acc1: 78.1250 (82.6100)  acc5: 93.7500 (96.2417)  time: 0.3584  data: 0.2839  max mem: 817
Test:  Total time: 0:02:00
Test:  Acc@1 81.644 Acc@5 95.858
```


Here is the transform:

```python3
        transforms=partial(
            ImageClassification, crop_size=256, resize_size=256, interpolation=InterpolationMode.BILINEAR
        ),
```

Here is the python script to transform official pretrained weights to the supported version of our implemented version:

```
import torch

_dict: OrderedDict[str, torch.Tensor] = torch.load("swinv2_tiny_patch4_window8_256.pth")["model"]
new_dict = OrderedDict()

for key, value in _dict.items():
    k = key
    v = value
    if 'attn_mask' in k:
        continue
    if 'relative_position_index' in k:
        v = v.flatten()
    if k.startswith('patch_embed'):
        k = k.replace('patch_embed.proj', 'features.0.0')
        k = k.replace('patch_embed.norm', 'features.0.2')
    elif k.startswith('layers'):
        k = k.removeprefix('layers.')
        if 'blocks' in k:
            k = f'features.{2*int(k[0]) + 1:d}.'+k[2:].removeprefix('blocks.')
            k = k.replace('logit_scale', 'v2_logit_scale')
            if k.endswith('q_bias'):
                v_bias = _dict[key[:-6]+'v_bias']
                k = k.replace('q_bias', 'qkv.bias')
                assert value.numel() == v_bias.numel()
                v = torch.cat([value, torch.zeros_like(value), v_bias])
            elif k.endswith('v_bias'):
                continue
            if '.mlp.' in k:
                k = k.replace('fc1', '0')
                k = k.replace('fc2', '3')
        else:
            assert 'downsample' in k, k
            k = f'features.{2*(int(k[0]) + 1):d}.'+k[2:].removeprefix('downsample.')
    new_dict[k] = v
torch.save(new_dict, './test.pth')
```

ain-soph:
I'm currently training a `swin_v2_t` using following command.

```
torchrun --nproc_per_node=4 train.py --model swin_v2_t --epochs 300 --batch-size 128 --opt adamw --lr 0.001 --weight-decay 0.05 --norm-weight-decay 0.0  --bias-weight-decay 0.0 --transformer-embedding-decay 0.0 --lr-scheduler cosineannealinglr --lr-min 0.00001 --lr-warmup-method linear  --lr-warmup-epochs 20 --lr-warmup-decay 0.01 --amp --label-smoothing 0.1 --mixup-alpha 0.8 --clip-grad-norm 5.0 --cutmix-alpha 1.0 --random-erase 0.25 --interpolation bicubic --auto-augment ta_wide --model-ema --ra-sampler --ra-reps 4  --val-resize-size 256 --val-crop-size 256 --train-crop-size 256
```

torchvision/models/swin_transformer.py
```diff
@@ -304,6 +457,57 @@ def forward(self, x: Tensor):
...
+        num_heads: int,
+        window_size: List[int],
+        shift_size: List[int],
+        mlp_ratio: float = 4.0,
+        dropout: float = 0.0,
+        attention_dropout: float = 0.0,
+        stochastic_depth_prob: float = 0.0,
+        norm_layer: Callable[..., nn.Module] = nn.LayerNorm,
+        attn_layer: Callable[..., nn.Module] = ShiftedWindowAttentionV2,
+        pretrained_window_size: int = 0,
```
YosuaMichael, line 489: Does this pretrained_window_size always has the same value for all the dimension? Just for context, we may expand similar API to Swin Transformer Video, and usually for video the window_size for time dimension is different with height and width, therefore we use `List[int]` at the end.

torchvision/models/swin_transformer.py
```diff
@@ -335,19 +540,26 @@ def __init__(
         attention_dropout: float = 0.0,
         stochastic_depth_prob: float = 0.0,
         num_classes: int = 1000,
-        norm_layer: Optional[Callable[..., nn.Module]] = None,
         block: Optional[Callable[..., nn.Module]] = None,
+        norm_layer: Optional[Callable[..., nn.Module]] = None,
+        use_v2: bool = False,
```
YosuaMichael, line 545: I think we can avoid using `use_v2` here.
Not sure if this is the best approach, but here is my idea:
1) We can let `block` by default `SwinTransformerBlock` as before. And for `v2` builder we should explicitly pass the `block=SwinTransformerBlockv2`.
2) For `v2_pretrained_window_sizes` I think we can just have the if - else logic based on whether the value is `None`
3) I think we should let `downsample_layer` as a parameters now where the default is `PatchMerging` but you can pass `PatchMergingV2` on the `swin_v2` builder

Let me know what are your thought on this :)

YosuaMichael (commented):
Hi @ain-soph, I think you have refactor the code nicely and now we have very minimum duplicated code!
I have a few comments on the parameters for v2, let me know what you think!

YosuaMichael:
@xiaohu2015 could you elaborate more on the bug detail that you mention:

> we should fix the problem in https://github.com/pytorch/vision/pull/6222 ( for input size=256 window size=8, it works well, but for input size=384 window size=12, it has a bug at the last block (it should be 6, but we do padding and use 8))

To be specific, can you explain for input_size=384 and window_size=12, what is the bug on the last block. And what do you mean by `it should be 6, but we do padding and use 8` (what this number refer for).

torchvision/models/swin_transformer.py
ain-soph: It's only used for some edge cases: That means The model is pretrained with `pretrained_window size`, but uses a different `window_size` at evaluation time.

You can view those cases in the yamls:
https://github.com/search?q=pretrained_window_sizes+repo%3Amicrosoft%2FSwin-Transformer+filename%3A*.yaml&type=Code&ref=advsearch&l=&l=

ain-soph:
@YosuaMichael and I are considering deprecating argument `pretrained_window_size` because it seems not used in standard variants. It's an argument about some fine-tuning related ablation studies in the original paper.

But this need some discussion. Any opinion on this? @xiaohu2015

torchvision/models/swin_transformer.py
```diff
@@ -92,26 +136,28 @@ def shifted_window_attention(
...
+        logit_scale (Tensor[out_dim], optional): Logit scale of cosine attention for Swin Transformer V2. Default: None.
     Returns:
         Tensor[N, H, W, C]: The output tensor after shifted window attention.
     """
     B, H, W, C = input.shape
+
+    # If window size is larger than feature size, there is no need to shift window
+    if window_size[0] >= H:
+        shift_size[0] = 0
+        window_size[0] = H
```
YosuaMichael, line 150: As discussed on chat, since this might introduce issue because relative_bias_index size is based on window_size, we should revert this and not having dynamic window_size for now.

YosuaMichael (commented):
Hi @ain-soph , I think overall the code looks really good now!
I just put the comments on things we have discussed on the calls so it is easier to recall.

YosuaMichael (commented):
Just few NITs and a question

torchvision/models/swin_transformer.py
```diff
@@ -333,18 +528,18 @@ def __init__(
         mlp_ratio: float = 4.0,
         dropout: float = 0.0,
         attention_dropout: float = 0.0,
-        stochastic_depth_prob: float = 0.0,
+        stochastic_depth_prob: float = 0.1,
```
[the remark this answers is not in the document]
ain-soph: To keep consistent with official repo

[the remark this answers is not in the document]
ain-soph: https://github.com/microsoft/Swin-Transformer/blob/main/models/swin_transformer_v2.py#L521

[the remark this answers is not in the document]
YosuaMichael: Got it 👍

jdsgomes:
Hi @ain-soph,  as discussed offline I am happy to take the review from here, and support you in the verification of the model weights that will be shipped.

ain-soph:
This is my training result of `swin_v2_t` using the command from `https://github.com/pytorch/vision/tree/main/references/classification#swintransformer`

```
Epoch: [299] Total time: 0:20:24
Test:   [ 0/98]  eta: 0:06:43  loss: 1.2362 (1.2362)  acc1: 93.7500 (93.7500)  acc5: 98.4375 (98.4375)  time: 4.1221  data: 3.8947  max mem: 14880
Test:  Total time: 0:00:26
Test:  Acc@1 81.560 Acc@5 95.892
Test: EMA  [ 0/98]  eta: 0:06:01  loss: 1.2378 (1.2378)  acc1: 93.7500 (93.7500)  acc5: 98.4375 (98.4375)  time: 3.6888  data: 3.4649  max mem: 14880
Test: EMA Total time: 0:00:25
Test: EMA Acc@1 81.614 Acc@5 95.930
Training time 4 days, 13:34:44
```

Compared with the official result [81.2%](https://github.com/pytorch/vision/tree/main/references/classification#swintransformer), I think our result (81.56%) is satisfying.

ain-soph:
@jdsgomes I've added the small and base variants in the previous commit. Please train the variants as well.  

Btw, do you think the current tiny variant performance is acceptable?

ain-soph:
Actually, maybe we can port the weights directly instead of training.

I've done such work and validate it at https://github.com/pytorch/vision/pull/6246#issuecomment-1179650923

jdsgomes:
I am going to launch the other training variants today and hopefully they will be ready by Monday. Regarding the difference in resolution size we can do a post training optimisation similar to what what we did in swin v1. There is no need for retraining we can just try to find the optimal val-resize-size

jdsgomes:
> I am going to launch the other training variants today and hopefully they will be ready by Monday. Regarding the difference in resolution size we can do a post training optimisation similar to what what we did in swin v1. There is no need for retraining we can just try to find the optimal val-resize-size

As discussed offline my train runs failed to converge so we will work together with @ain-soph  to find the source of the problem  and update the PR once we managed to train all variants successfully.

ain-soph:
I've launched another train based on the most up-to-date code for tiny architecture.  
Currently I don't observe any strange trend showing failure of convergence.
```
Epoch: [142]  [2500/2502]  eta: 0:00:00  lr: 0.0006042917608127198  img/s: 266.4080904318139  loss: 3.6911 (3.6364)  acc1: 48.4375 (48.2548)  acc5: 73.4375 (71.9340)  time: 0.4798  data:
 0.0001  max mem: 14880                                                                                                                                                                   
Epoch: [142] Total time: 0:20:10                                                                                                                                                          
Test:   [ 0/98]  eta: 0:06:25  loss: 1.5864 (1.5864)  acc1: 82.0312 (82.0312)  acc5: 96.0938 (96.0938)  time: 3.9314  data: 3.7073  max mem: 14880                                        Test:  Total time: 0:00:25                                                                                                                                                                
Test:  Acc@1 71.494 Acc@5 91.496                                                                                                                                                          Test: EMA  [ 0/98]  eta: 0:05:43  loss: 1.4705 (1.4705)  acc1: 88.2812 (88.2812)  acc5: 97.6562 (97.6562)  time: 3.5061  data: 3.2585  max mem: 14880                                     
Test: EMA Total time: 0:00:25                                                                                                                                                             Test: EMA Acc@1 76.420 Acc@5 93.660 
```

datumbox:
@ain-soph Could you please share the exact command you use to train it to ensure we are not missing anything from our side?

datumbox:
Ooops, there are also memory issues on GPU both on [linux](https://app.circleci.com/pipelines/github/pytorch/vision/19216/workflows/631388a0-8439-4f65-b907-4b41a9370cf8/jobs/1555278) and [windows](https://app.circleci.com/pipelines/github/pytorch/vision/19216/workflows/631388a0-8439-4f65-b907-4b41a9370cf8/jobs/1555296). I know that a different test is actually failing but this can happen. Sometimes adding a bing new model leads to issues on other models because of failing to clear the memory properly. Usually the way around this is to reduce the memory footprint of the test by passing thought smaller sizes for input or disabling the particular test on the GPU.

ain-soph:
```
torchrun --nproc_per_node=4 train.py --model swin_v2_t --epochs 300 --batch-size 128 --opt adamw --lr 0.001 --weight-decay 0.05 --norm-weight-decay 0.0  --bias-weight-decay 0.0 --transformer-embedding-decay 0.0 --lr-scheduler cosineannealinglr --lr-min 0.00001 --lr-warmup-method linear  --lr-warmup-epochs 20 --lr-warmup-decay 0.01 --amp --label-smoothing 0.1 --mixup-alpha 0.8 --clip-grad-norm 5.0 --cutmix-alpha 1.0 --random-erase 0.25 --interpolation bicubic --auto-augment ta_wide --model-ema --ra-sampler --ra-reps 4 --val-resize-size 256 --val-crop-size 256 --train-crop-size 256
```

```
Epoch: [221] Total time: 0:25:25
Test:   [ 0/98]  eta: 0:06:23  loss: 1.3831 (1.3831)  acc1: 91.4062 (91.4062)  acc5: 98.4375 (98.4375)  time: 3.9118 data: 3.6845  max mem: 14880
Test:  Total time: 0:00:25
Test:  Acc@1 77.738 Acc@5 94.460
Test: EMA  [ 0/98]  eta: 0:05:15  loss: 1.3172 (1.3172)  acc1: 90.6250 (90.6250)  acc5: 99.2188 (99.2188)  time: 3.2230 data: 2.9882  max mem: 14880
Test: EMA Total time: 0:00:25
Test: EMA Acc@1 80.012 Acc@5 95.286
```

jdsgomes:
> ```
> torchrun --nproc_per_node=4 train.py --model swin_v2_t --epochs 300 --batch-size 128 --opt adamw --lr 0.001 --weight-decay 0.05 --norm-weight-decay 0.0  --bias-weight-decay 0.0 --transformer-embedding-decay 0.0 --lr-scheduler cosineannealinglr --lr-min 0.00001 --lr-warmup-method linear  --lr-warmup-epochs 20 --lr-warmup-decay 0.01 --amp --label-smoothing 0.1 --mixup-alpha 0.8 --clip-grad-norm 5.0 --cutmix-alpha 1.0 --random-erase 0.25 --interpolation bicubic --auto-augment ta_wide --model-ema --ra-sampler --ra-reps 4 --val-resize-size 256 --val-crop-size 256 --train-crop-size 256
> ```
> 
> ```
> Epoch: [221] Total time: 0:25:25
> Test:   [ 0/98]  eta: 0:06:23  loss: 1.3831 (1.3831)  acc1: 91.4062 (91.4062)  acc5: 98.4375 (98.4375)  time: 3.9118 data: 3.6845  max mem: 14880
> Test:  Total time: 0:00:25
> Test:  Acc@1 77.738 Acc@5 94.460
> Test: EMA  [ 0/98]  eta: 0:05:15  loss: 1.3172 (1.3172)  acc1: 90.6250 (90.6250)  acc5: 99.2188 (99.2188)  time: 3.2230 data: 2.9882  max mem: 14880
> Test: EMA Total time: 0:00:25
> Test: EMA Acc@1 80.012 Acc@5 95.286
> ```

Thank you for sharing the commands. Although you previously mention I missed that in v2 they used a different resolution size. I launched new jobs to train all the variants with the correct resolution size and seems are looking better now.

jdsgomes:
@ain-soph  thank you for your patience and great work! I managed to train all three variants and the results look good.

My plan is to update this PR in the next couple of ours with the model weights.

jdsgomes:
Trainning commands:

```
# swin_v2_t
python -u run_with_submitit.py --timeout 3000 --ngpus 8 --nodes 1  --model swin_v2_t --epochs 300 \
--batch-size 128 --opt adamw --lr 0.001 --weight-decay 0.05 --norm-weight-decay 0.0  \
--bias-weight-decay 0.0 --transformer-embedding-decay 0.0 --lr-scheduler cosineannealinglr \
--lr-min 0.00001 --lr-warmup-method linear  --lr-warmup-epochs 20 --lr-warmup-decay 0.01 \
--amp --label-smoothing 0.1 --mixup-alpha 0.8 --clip-grad-norm 5.0 --cutmix-alpha 1.0 \
--random-erase 0.25 --interpolation bicubic --auto-augment ta_wide --model-ema --ra-sampler \
--ra-reps 4  --val-resize-size 256 --val-crop-size 256 --train-crop-size 256 

# swin_v2_s
python -u run_with_submitit.py --timeout 3000 --ngpus 8 --nodes 1  --model swin_v2_s --epochs 300 \
--batch-size 128 --opt adamw --lr 0.001 --weight-decay 0.05 --norm-weight-decay 0.0  \
--bias-weight-decay 0.0 --transformer-embedding-decay 0.0 --lr-scheduler cosineannealinglr \
--lr-min 0.00001 --lr-warmup-method linear  --lr-warmup-epochs 20 --lr-warmup-decay 0.01 --amp --label-smoothing 0.1 --mixup-alpha 0.8 --clip-grad-norm 5.0 --cutmix-alpha 1.0 \
--random-erase 0.25 --interpolation bicubic --auto-augment ta_wide --model-ema --ra-sampler \
--ra-reps 4  --val-resize-size 256 --val-crop-size 256 --train-crop-size 256

# swin_v2_b
python -u run_with_submitit.py --timeout 3000 --ngpus 8 --nodes 1  --model swin_v2_b --epochs 300 \
--batch-size 128 --opt adamw --lr 0.001 --weight-decay 0.05 --norm-weight-decay 0.0  \
--bias-weight-decay 0.0 --transformer-embedding-decay 0.0 --lr-scheduler cosineannealinglr \
--lr-min 0.00001 --lr-warmup-method linear  --lr-warmup-epochs 20 --lr-warmup-decay 0.01 \
--amp --label-smoothing 0.1 --mixup-alpha 0.8 --clip-grad-norm 5.0 --cutmix-alpha 1.0 \
--random-erase 0.25 --interpolation bicubic --auto-augment ta_wide --model-ema --ra-sampler \
--ra-reps 4  --val-resize-size 256 --val-crop-size 256 --train-crop-size 256 

```

Test commands and acuracies
```
# swin_v2_t
srun -p dev --cpus-per-task=96 -t 24:00:00 --gpus-per-node=1 torchrun --nproc_per_node=1 train.py     \
    --model swin_v2_t --test-only --resume $EXPERIMENTS_PATH/44757/model_299.pth --interpolation bicubic  \
    --val-resize-size 260 --val-crop-size 256
# Test:  Acc@1 82.072 Acc@5 96.132

# swin_v2_s
srun -p dev --cpus-per-task=96 -t 24:00:00 --gpus-per-node=1 torchrun --nproc_per_node=1 train.py     \
    --model swin_v2_s --test-only --resume $EXPERIMENTS_PATH/44758/model_299.pth --interpolation bicubic   \
    --val-resize-size $260--val-crop-size 256
# Test:  Acc@1 83.712 Acc@5 96.816

# swin_v2_b
srun -p dev --cpus-per-task=96 -t 24:00:00 --gpus-per-node=1 torchrun --nproc_per_node=1 train.py     \
    --model swin_v2_b --test-only --resume $EXPERIMENTS_PATH/44759/model_299.pth --interpolation bicubic   \
    --val-resize-size 272 --val-crop-size 256
# Test:  Acc@1 84.112 Acc@5 96.864
```

datumbox:
@ain-soph @jdsgomes Awesome work! Having SwinV2 in TorchVision is really awesome. Looking forward using them.

> Btw, should we provide porting model weights from official repo as an alternative?

I think given we were able to reproduce the accuracy of the paper, there is no point offering both. What would be interesting on the future is to offer higher accuracy weights by using newer training recipes.

jdsgomes (approved):
Based on Yosua previous review of the ML side and mine replication of the results approving. Thanks @ain-soph  for you patience and great work. 

As soon as the tests are green we we are good to merge

datumbox (approved):
I reviewed @jdsgomes 5 last commits and everything LGTM. I would be OK to merge on green CI (if the memory issue persists on the CI, we might need to turn off some of the tests).

PR #6248, opened 2022-07-08
[jit] Updated tests checking save and load transforms

vfdev-5:
Description:

If we change type hints, jit can fail only on loading scripted saved transform. Such error was only caught accidentally by a CI job building gallery example for the PR with type hint change: https://github.com/pytorch/vision/pull/6245.

In this PR let's add loading pass for a saved transform.

Additionally, I fixed code formatting to reduce the number of lines.

### DIFF
--- a/test/test_transforms_tensor.py
+++ b/test/test_transforms_tensor.py
@@ -93,6 +93,15 @@ def _test_op(func, method, device, channels=3, fn_kwargs=None, meth_kwargs=None,
     _test_class_op(method, device, channels, meth_kwargs, test_exact_match=test_exact_match, **match_kwargs)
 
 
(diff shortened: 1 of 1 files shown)


datumbox (approved):
LGTM, thanks!

NicolasHug:
Thanks for the PR @vfdev-5 

> Additionally, I fixed code formatting to reduce the number of lines.

Just a minor note, in the future it might be preferable to do that kind of formatting in a separate PR, so that we can add the entry to https://github.com/pytorch/vision/blob/main/.git-blame-ignore-revs in order to keep `git-blame` clean

vfdev-5:
@NicolasHug thanks, sounds good!

PR #6251, opened 2022-07-08
Generalize ConvNormActivation function to accept tuple for some parameters

YosuaMichael:
Currently the function `ConvNormActivation` and its variant `Conv2dNormActivation` / `Conv3dNormActivation` only accept integer for kernel_size, stride, padding, and dilation. However, some model require tuple kernel_size that have different value ([example in S3D model](https://github.com/kylemin/S3D/blob/master/model.py#L57)).

With this PR, we enable these 4 parameters to accept tuple as well as integer and we make sure it is BC.

### DIFF
--- a/torchvision/ops/misc.py
+++ b/torchvision/ops/misc.py
@@ -72,7 +72,7 @@ class ConvNormActivation(torch.nn.Sequential):
         out_channels: int,
         kernel_size: Union[int, Tuple[int, ...]] = 3,
         stride: Union[int, Tuple[int, ...]] = 1,
-        padding: Optional[Union[int, Tuple[int, ...]]] = None,
+        padding: Optional[Union[int, Tuple[int, ...], str]] = None,
         groups: int = 1,
         norm_layer: Optional[Callable[..., torch.nn.Module]] = torch.nn.BatchNorm2d,
         activation_layer: Optional[Callable[..., torch.nn.Module]] = torch.nn.ReLU,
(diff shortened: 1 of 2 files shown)


torchvision/ops/misc.py
```diff
@@ -70,20 +70,23 @@ def __init__(
...
         norm_layer: Optional[Callable[..., torch.nn.Module]] = torch.nn.BatchNorm2d,
         activation_layer: Optional[Callable[..., torch.nn.Module]] = torch.nn.ReLU,
-        dilation: int = 1,
+        dilation: Union[int, Tuple[int, ...]] = 1,
         inplace: Optional[bool] = True,
         bias: Optional[bool] = None,
         conv_layer: Callable[..., torch.nn.Module] = torch.nn.Conv2d,
     ) -> None:
 
+        _conv_dim = 2 if conv_layer == torch.nn.Conv2d else 3
```
datumbox, line 85: This is a bit flimsy way to capture the dimensionality. What if we have a different type of 2d [convolution](https://pytorch.org/docs/stable/nn.html#convolution-layers) like `LazyConv2d` or `ConvTranspose2d`?

I think one way to avoid this is by not requiring to find this number. We can rely on the user to provide us with either single integer and keep it that way or a tuple.

datumbox (commented):
Thanks @YosuaMichael. I left a few comments below, let me know what you think.

torchvision/ops/misc.py
```diff
@@ -70,20 +70,23 @@ def __init__(
         self,
         in_channels: int,
         out_channels: int,
-        kernel_size: int = 3,
-        stride: int = 1,
-        padding: Optional[int] = None,
+        kernel_size: Union[int, Tuple[int, ...]] = 3,
```
oke-aditya, line 73: I was scared that by adding Union are we JIT compatible? I had avoided Union for the same.

YosuaMichael: thanks for the input @oke-aditya , let me check if this is JIT compatible or not. Will update here

datumbox: There are no JIT-scriptability concerns here. Constructors can have whatever calls. It's on the forward calls that we have the restrictions. :)

torchvision/ops/misc.py
```diff
@@ -70,20 +70,29 @@ def __init__(
...
 
         if padding is None:
-            padding = (kernel_size - 1) // 2 * dilation
+            if isinstance(kernel_size, int) and isinstance(dilation, int):
+                padding = (kernel_size - 1) // 2 * dilation
+            else:
+                if isinstance(kernel_size, Tuple):
+                    _conv_dim = len(kernel_size)
+                elif isinstance(dilation, Tuple):
+                    _conv_dim = len(dilation)
```
YosuaMichael, line 92: @datumbox I have changed the implementation. I still need to know the dimension in case either `kernel_size` or `dilation` is tuple because I will need to generate padding as tuple as well.

datumbox, line 92: Nit: Your code is correct but the if/elif idiom can be replaced with:
```suggestion
                _conv_dim = len(kernel_size) if isinstance(kernel_size, Sequence) else len(dilation)
```

torchvision/ops/misc.py
YosuaMichael: I have also try to do 
```
import torchvision.models as models
import torch

m = models.efficientnet_b0()
torch.jit.script(m)
```
and it works as well (notes: efficientnet use Conv2dNormActivation)

torchvision/ops/misc.py
YosuaMichael: I have updated the code, to get convolution dimension from `kernel_size` or `dilation` if needed. What do you think on this?

oke-aditya:
If JIT is happy then we are all good :smile:

Issue #6257, opened 2022-07-11
Remove deprecated APIs for 0.14


NicolasHug:
In 0.12 we marked some APIs as deprecated, and to be removed in 0.14. Since 0.14 is the upcoming version, now is the time to remove them.

I'll send a PR soon.

vsuryamurthy:
Hello Nicolas, Is there a release date planned for 0.14 already?

PR #6258, opened 2022-07-11
Remove deprecated APIs for 0.14

NicolasHug:
Closes https://github.com/pytorch/vision/issues/6257

`git grep -n "0\.14" torchvision/` now gives zero output


torchvision/csrc/models/modelsimpl.h
```diff
@@ -36,8 +36,8 @@ inline bool double_compare(double a, double b) {
 
 inline void deprecation_warning() {
   TORCH_WARN_ONCE(
-      "The vision::models namespace is deprecated since 0.12 and will be "
-      "removed in 0.14. We recommend using Torch Script instead: "
+      "The vision::models namespace is not actively maintained, use at "
+      "your own discretion. We recommend using Torch Script instead: "
```
NicolasHug, line 40: Looks like we originally didn't intend to deprecate the C++ models: at least according to https://github.com/pytorch/vision/pull/4375#pullrequestreview-751524747, the original intention was to stop maintaininng them but to still keep them around.

So I put back the original warning, instead of removing the files. I'll still wait for @datumbox 's input on this tho.

datumbox: Thanks for checking @NicolasHug. The intention was to remove them. The majority of the new models can't be added on C++ because the layers are only available on Python. So I think we are good to drop them, which was the intention when we deprecated them.

NicolasHug:
@datumbox following up on https://github.com/pytorch/vision/pull/6258#discussion_r928926367 about the C++ models: if that works for you, I'll remove them in another PR. The only reason is that I fear that removing them might break some stuff internally. I'd rather isolate that in a separate PR, to avoid having to potentially revert other unrelated changes (like the rest of the deprecations).

jdsgomes:
@kazhang thanks for pointing out that `_transforms_video.py` is still referred internally so we need to discuss a plan for deprecating this particular file.

NicolasHug:
Thanks @jdsgomes . I opened D38393214 to see what tests might break if we removed these files.

datumbox:
@NicolasHug I don't have a super strong opinion about splitting the PRs if that simplifies your work. Curious why you think this can cause issues on FBcode though. As far as I see we don't even compile these sources on BUCK, so I don't see how people are using them. Have you saw anything different? If not, you can add the removal here; otherwise send a separate PR that removes the endpoints in question as you proposed.

Happy to provide the necessary stamps once you confirm things work on FBcode + the tests pass.

torchvision/transforms/_functional_video.py
```diff
@@ -1,114 +0,0 @@
-import warnings
-
-import torch
-
-
-warnings.warn(
-    "The 'torchvision.transforms._functional_video' module is deprecated since 0.12 and will be removed in 0.14. "
```
datumbox, line 7: I commented this on the internal [diff](https://www.internalfb.com/diff/D39726995) that @YosuaMichael opened to check the deprecation but for visibility to the community here is my comment here:
> we can not deprecate torchvision.transforms._transforms_video since we don't offer currently an alternative. I will be working on this next week. I recommend changing the deprecation message on OSS to notify that we would be removing this on a future version and then go from scratch through a deprecation process once we have a clear alternative to define.

YosuaMichael: @datumbox I add back this file and change the warnings message, could you double check if this message is good? We can change the message to give exact version when it will be removed once we have a clear alternative as you said.

datumbox: I was wondering if it's worth making this change on a separate PR on main because this one is quite massive. The message looks OK to me. There are a few more things we would need to revert such as the unit-tests for the specific module.

NicolasHug: I'm not sure we really need to delay the removal here: on top of being deprecated, these APIs are private. Users who are still relying on it have been ignoring strong signals. IMHO we should either:

- contact all downstream users who are still relying on this and tell them they have 10 days (or whatever) to update their code
- be nice to them and submit diffs so that they vendor `_functional_video.py` themselves.

YosuaMichael: For the unit-test, do you mean the `test_transforms_video.py`? I also reverted it in this PR so it is not deleted.
For changing the message, I think it is a good idea to have separate PR just to make sure it has more visibility for the change in deprecation plan. I will revert back the message here and create another PR for that then.

YosuaMichael: @NicolasHug in my understanding, currently there is no alternatives for the `_functional_video.py`. In this situation, it will be very difficult for the users to update their code by just changing API or import.

By `they vendor _functional_video.py themselves` do you mean to copy and paste the _function_video.py on users code?

datumbox: @NicolasHug Multiple internal users are still depending on this. Unless you want to dedicate your next 10 days porting their code away from these definitions and working with each individual team to ensure the results in production are identical, I don't think we can proceed with the removal. 

If you check Yosua's [Diff](https://www.internalfb.com/diff/D39726995) you will see multiple tests failing. In addition to the internal stuff, libraries such as [PyTorch Video](https://github.com/facebookresearch/pytorchvideo/blob/430768adbb8e279ca4192db125583a231f3d60c5/pytorchvideo/neural_engine/hook.py) (cc @haooooooqi @lyttonhao) and [ClassyVision](https://github.com/facebookresearch/ClassyVision/blob/35c8f33f2edf4d08d7660bfb61ce4b938aa7aa02/classy_vision/dataset/transforms/util_video.py) (cc @jdsgomes @nairbv) are also affected.

IMO the issue with the specific deprecation is that it doesn't provide an actionable alternative that works out of the box for users and that lead to them not moving away from it. Though I agree there are concerns about respecting our own policies, we probably want to avoid breaking all downstream projects. I recommend reverting this removal to unblock the rest of the deprecations that are more straightforward and bring this topic up for discussion on the next weekly. There we can decide if we should move forwards with deprecating the specific part or not, without delaying the rest of the work.

NicolasHug: > By they vendor _functional_video.py themselves do you mean to copy and paste the _function_video.py on users code?

Yes, I mean that each downstream library could simply copy this file in their code-base. [` _functional_video.py`](https://github.com/pytorch/vision/blob/main/torchvision/transforms/_functional_video.py) is a small self-contained file that only has `torch` as dependency, so this would be a very reasonable ask from us.

> I recommend reverting this removal to unblock the rest of the deprecations that are more straightforward and bring this topic up for discussion on the next weekly. There we can decide if we should move forwards with deprecating the specific part or not, without delaying the rest of the work

I'm not sure I understand what we mean by "move forwards with deprecating the specific part or not", but I agree we can discuss how to move forward with ` _functional_video.py` case separately (urgently, considering the approaching release).

jdsgomes: yes I confirm that _ClassyVision uses `_transforms_video.py`

YosuaMichael: If I understand correctly, we are agree to discuss how to handle `_functional_video.py` (probably in weekly). And if we decide to remove it, I will create separate PR for that.

YosuaMichael:
@NicolasHug @datumbox @jdsgomes Could you check if this PR is okay to merge?
I think regarding the removal of deprecated API, other than this PR only the following two things left:
- Removal of cpp models: https://github.com/pytorch/vision/pull/6632
- To be discussed whether or not we remove `transforms/_functional_video.py` and `transforms/_transforms_video.py` but we agree this will be on separate PR

I have checked and everything seems fine to me on this PR.

NicolasHug (commented):
GH doesn't let me approve as I'm the original author, but LGMT. Thanks @YosuaMichael

torchvision/models/mobilenetv2.py
```diff
@@ -17,24 +16,6 @@
...
-        )
-        if kwargs.get("norm_layer", None) is None:
-            kwargs["norm_layer"] = nn.BatchNorm2d
-        if kwargs.get("activation_layer", None) is None:
-            kwargs["activation_layer"] = nn.ReLU6
-        super().__init__(*args, **kwargs)
-
-
-ConvBNReLU = _DeprecatedConvBNAct
-ConvBNActivation = _DeprecatedConvBNAct
```
datumbox, line 35: @YosuaMichael Just being paranoid. Have you confirmed that this works on FBCode and that there are no breakages?

torchvision/models/mobilenetv3.py
```diff
@@ -22,21 +21,6 @@
 ]
 
 
-class SqueezeExcitation(SElayer):
```
datumbox, line 25: Same here.

datumbox (approved):
@YosuaMichael stamping to unblock. LGTM, with only 2 questions below.

> GH doesn't let me approve as I'm the original author,

@NicolasHug do you prefer to do this on a new PR so that the attribution is right?

NicolasHug:
> @NicolasHug do you prefer to do this on a new PR so that the attribution is right?

Just FYI, the changes in this PR come from my commits. 

But whether this PR gets merged or another makes no difference to me.

torchvision/models/mobilenetv2.py
YosuaMichael: Hi @datumbox, I have tested on this diffs: [D39730821](https://www.internalfb.com/diff/D39730821). It is a stack on top of the diffs that remove everything including `transforms/_functional_video.py` but I add this file back.

YosuaMichael:
Just to update, I found 2 breakages internally after trying this changes. I have created diffs to fix for each of them, and I will merge this PR once the fix is landed.

1) Break due to Kinetics400 removal: [D39832659](https://www.internalfb.com/diff/D39832659)
2) Break due removal of `resample` and `fillcolor` argument on `RandomAffine`: [D39843275](https://www.internalfb.com/diff/D39843275)

Issue #4902, opened 2021-11-10
JIT not working with `torchvision.transforms.CenterCrop`


JonathanWoollett-Light:
## Bug

[All files here (`model.pth`, `model.zip`, `jit-load-test.py` and `test.py`)](https://drive.google.com/drive/folders/1XyEGzQSQzEu5ED78adkS3rW9UsQUhBZe?usp=sharing)

I have a semantic segmentation model which utilizes a center crop on its output (`test.py`):
```python
class UNet(nn.Module):
    def __init__(
        # ...
    ):
        # ...
        self.crop = torchvision.transforms.CenterCrop(output_size)
    def forward(self,x):
        # ...
        crop = self.crop(x)
        # ...
```
Trying to save this as a JIT model results in the error:
```
Traceback (most recent call last):
  File "test.py", line 110, in <module>
    main()
  File "test.py", line 103, in main
    net_trace = torch.jit.trace(model,example_input) # Traces using this batch
  File "/home/ubuntu/.local/lib/python3.8/site-packages/torch/jit/_trace.py", line 741, in trace
    return trace_module(
  File "/home/ubuntu/.local/lib/python3.8/site-packages/torch/jit/_trace.py", line 958, in trace_module
    module._c._create_method_from_trace(
  File "/home/ubuntu/.local/lib/python3.8/site-packages/torch/nn/modules/module.py", line 1102, in _call_impl
    return forward_call(*input, **kwargs)
  File "/home/ubuntu/.local/lib/python3.8/site-packages/torch/nn/modules/module.py", line 1090, in _slow_forward
    result = self.forward(*input, **kwargs)
  File "test.py", line 84, in forward
    crop = torchvision.transforms.functional.center_crop(final,self.output_size)
  File "/home/ubuntu/.local/lib/python3.8/site-packages/torchvision/transforms/functional.py", line 534, in center_crop
    crop_top = int(round((image_height - crop_height) / 2.))
TypeError: type Tensor doesn't define __round__ method
```
If I replace `self.crop = torchvision.transforms.CenterCrop(output_size)` with `self.crop = torch.jit.script(torch.nn.Sequential(torchvision.transforms.CenterCrop(output_size)))` saving with JIT works.

However when trying to load this (`jit-load-test.py`):
```python
import torch
model = torch.jit.load("model.zip")
input = torch.ones([12,3,300,600])
output = model(input)
print(output)
```
Results in the error:
```
Traceback (most recent call last):
[...]

Issue #4607, opened 2021-10-13
torch.onnx.export failes due to missing __round__ function


mechaot:
### 🐛 Describe the bug

When exporting a trained CNN to onnx file format using torch.onnx.export(), an exception is thrown from torchvision/transforms/functional.py and the model is not saved.

## To Reproduce

1. Define a CNN-torch.Model instance similar to a VGG-Network, train it
2. Try to export the trained model to an *.onnx - file using the save_model() function below

``` python
from os import makedirs, join
import torch
import torch.onnx as onnx

def save_model(model, batch_size, image_size, save_dir, input_channels, tensor_type):
    makedirs(save_dir, exist_ok=True)

    torch.save(model.state_dict(), join(save_dir, f"model.pt"))

    x = torch.randn(batch_size, len(input_channels), *image_size, requires_grad=True)  # a bogus input tensor; needed for tracing onnx export
    x = x.type(tensor_type) # move to device

    onnx.export(model,                                          # model being run
                x,                                              # model input (or a tuple for multiple inputs)
                join(save_dir, f"model.onnx"),                  # where to save the model (can be a file or file-like object)
                export_params=True,                             # store the trained parameter weights inside the model file
                opset_version=10,                               # the ONNX version to export the model to
                do_constant_folding=True,                       # whether to execute constant folding for optimization
                input_names = ['input'],                        # the model's input names
                output_names = ['output'],                      # the model's output names
                dynamic_axes={'input' : {0 : 'batch_size'},     # variable length axes
                            'output' : {0 : 'batch_size'}})
    print("onnx model exported")  
```

``` bash
~/miniconda3/envs/playground/lib/python3.8/site-packages/torchvision/transforms/functional.py in center_crop(img, output_size)
    514             return img
    515 
--> 516     crop_top = int(round((image_height - crop_height) / 2.))
    517     crop_left = int(round((image_width - crop_width) / 2.))
    518     return crop(img, crop_top, crop_left, crop_height, crop_width)

TypeError: type Tensor doesn't define __round__ method

```

## Expected behavior

Model is saved to file and can be used with onnxruntime library later on.

## Workaround (or possible solution)

in `torchvision/transforms/functional.py` function `center_crop` change lines 516-517 to use the numpy.round() instead of built-in round():

``` python
# old code for reference:    
# crop_top = int(round((image_height - crop_height) / 2.))
# crop_left = int(round((image_width - crop_width) / 2.))
# working by using numpy.round():
[...]

Issue #5274, opened 2022-01-24
Use of Python round() inbuilt function inside Functional.CenterCrop() NonDeterministic


revorg7:
### 🐛 Describe the bug

Barring the padding process, CenterCrop is implemented with these lines:
crop_top = int(round((image_height - crop_height) / 2.0))
crop_left = int(round((image_width - crop_width) / 2.0))
return crop(img, crop_top, crop_left, crop_height, crop_width)

Note that round(11/2)=6 and round(9/2)=4, i.e., python round() selects floor() sometimes, and sometimes ceil(). This causes crop_top and crop_left coordinates to be non-deterministic (depending on input sizes). 

Remedy: Just remove the round() since you already have int() outside (change it to floor() to avoid further confusion)

### Versions

Collecting environment information...
PyTorch version: 1.9.1+cu111
Is debug build: False
CUDA used to build PyTorch: 11.1
ROCM used to build PyTorch: N/A

OS: Red Hat Enterprise Linux release 8.5 (Ootpa) (x86_64)
GCC version: Could not collect
Clang version: Could not collect
CMake version: version 3.20.2
Libc version: glibc-2.3.4

Python version: 3.6.8 (default, Sep  9 2021, 07:49:02)  [GCC 8.5.0 20210514 (Red Hat 8.5.0-3)] (64-bit runtime)
Python platform: Linux-4.18.0-348.2.1.el8_5.x86_64-x86_64-with-redhat-8.5-Ootpa
Is CUDA available: True
CUDA runtime version: Could not collect
GPU models and configuration: GPU 0: Quadro RTX 6000
Nvidia driver version: 470.57.02
cuDNN version: Could not collect
HIP runtime version: N/A
MIOpen runtime version: N/A

Versions of relevant libraries:
[pip3] numpy==1.19.5
[pip3] torch==1.9.1+cu111
[pip3] torchvision==0.10.1+cu111
[conda] Could not collect

cc @vfdev-5 @datumbox @fmassa @pmeier

vfdev-5:
@revorg7 python round is a deterministic operation as `round(value)` always gives the same output. 
Concerning your point about that round(5.5) -> 6 and round(4.5) -> 4 is related to how floating point works. See https://docs.python.org/3/library/functions.html#round

Your suggestion to remove `round` would introduce a backward incompatibility.

Can you please give more details on what is your exact issue with the center_crop ? Thanks

revorg7:
[...]

PR #6263, opened 2022-07-13
fix center_crop to support jit trace

dongkyuk:
Proposed fix for issue https://github.com/pytorch/vision/issues/4607, https://github.com/pytorch/pytorch/issues/76370, https://github.com/pytorch/vision/issues/4902

### DIFF
--- a/torchvision/transforms/functional.py
+++ b/torchvision/transforms/functional.py
@@ -584,8 +584,8 @@ def center_crop(img: Tensor, output_size: List[int]) -> Tensor:
         if crop_width == image_width and crop_height == image_height:
             return img
 
-    crop_top = int(round((image_height - crop_height) / 2.0))
-    crop_left = int(round((image_width - crop_width) / 2.0))
+    crop_top = round(int(image_height - crop_height) / 2.0)
+    crop_left = round(int(image_width - crop_width) / 2.0)
     return crop(img, crop_top, crop_left, crop_height, crop_width)
 
 


vfdev-5:
@dongkyuk thanks for the PR, however, as it was stated https://github.com/pytorch/vision/issues/4902#issuecomment-973054194, torchvision transforms supports only torch jit script and not tracing. 
Also, the change you propose will break backward-compatibility.

derrickgw:
This bug also effect jit scripting.  I experienced this when trying to export a pytorch model with onnx. Fortunately I was able to monkey patch this so that I could pytorch.

Can you please go into more detail about how casting scalars representing the number of pixels in an image to integers will break backwards compatibility?  Are we worried that the number of pixels in an image is a float? Generally backwards compatibility simply means that the code give the same answers for the same input.  Are there instances where the new code would give a different answer for the same inputs? What if we cast before the subtraction instead of after?  Would that be acceptable to not break backwards compatibility? 

It is unfortunate that this didn't get fixed in the v2 transforms, because then backwards compatibility would not have been an issue.

pac48:
@derrickgw 
Doesn't the proposed change in https://github.com/pytorch/vision/pull/6263  
```
crop_top = round(int(image_height - crop_height) / 2.0)
crop_left = round(int(image_width - crop_width) / 2.0)
```
yield a different result when `(image_height - crop_height) / 2.0`  are not integer numbers? I think there is more detail in this comment: https://github.com/pytorch/vision/issues/5274#issuecomment-1020999656

The desired behavior is:
```
11.0 / 2.0 = 5.5
round(11.0 / 2.0) = 6.0
9.0 / 2.0 = 4.5
round(9.0 / 2.0) = 4.0
```

I think the proper fix might look more like this:
```
if isinstance(image_height, torch.Tensor):
    crop_top = int(((image_height - crop_height) / 2.0).round())
    crop_left = int(((image_width - crop_width) / 2.0).round())
else:
    crop_top = int(round((image_height - crop_height) / 2.0))
    crop_left = int(round((image_width - crop_width) / 2.0))
``` 

@vfdev-5 Can you confirm this is the backward incompatibility that you are referring to?

Issue #6259, opened 2022-07-11
[RFC] Stereo Matching Datasets API


TeodorPoncu:
### 🚀 The feature

The proposed feature aims to **extend the current datasets API** with datasets that are geared towards `the task of Stereo Matching`. It's main use case is that of providing  **a unified way for consuming classic Stereo Matching datasets** such as: 

- [Middlebury2014](https://vision.middlebury.edu/stereo/data/scenes2014/)
- [Kitti](http://www.cvlibs.net/datasets/kitti/eval_scene_flow.php?benchmark=stereo)
- [SceneFlow](https://lmb.informatik.uni-freiburg.de/resources/datasets/SceneFlowDatasets.en.html)

Other considered dataset additions are: Sintel, FallingThings, InStereo2K, ETH3D, Holopix50k. A high level preview of the dataset interface would be:

```python3
class StereoMatchingDataset(Dataset):
    def __init__(self, ...):
        # constructor code / dataset specific code

    def __getitem__(self, index: int) -> Tuple[Tensor, Tensor, Tensor, Tensor]:
        # processing code
        # ...

        # imgs: Tuple[Tensor, Tensor] 
        # dispartieis: Tuple[Tensor, Tensor]
        # occlusion_masks: Tuple[Tensor, Tensor]

        if self.transforms is not None:
            imgs, disparities, occlusion_masks = self.transforms(imgs, disparities, occlusion_masks)

        img_left = imgs[0]
        img_right = imgs[1]
        disparity = disparities[0]
        occlusion_mask = occlusion_masks[0]

        return img_left, img_right, disparity, occlusion_mask
```

### Motivation, pitch

This API addition would cut down engineering time required for people that are looking into working on, experimenting, or evaluating Stereo Matching models or that want easy access to stereo image data.

Throughout the literature, recent methods ([1](https://openaccess.thecvf.com/content/CVPR2022/papers/Li_Practical_Stereo_Matching_via_Cascaded_Recurrent_Network_With_Adaptive_Correlation_CVPR_2022_paper.pdf), [2](https://arxiv.org/pdf/2109.07547.pdf)) make use of multiple datasets that all have different formatting or specifications. A unified dataset API would streamline interacting with different datasources at the same time.

### Alternatives

[...]

PR #6270, opened 2022-07-14
Moved pfm file reading into dataset utils

TeodorPoncu:
Addresses [#6259](https://github.com/pytorch/vision/pull/6269#discussion_r920971491) functionality redefinition both in _optical_flow.py and _stereo_matching datasets.

### DIFF
--- a/torchvision/datasets/utils.py
+++ b/torchvision/datasets/utils.py
@@ -18,6 +18,7 @@ import zipfile
 from typing import Any, Callable, List, Iterable, Optional, TypeVar, Dict, IO, Tuple, Iterator
 from urllib.parse import urlparse
 
+import numpy as np
 import requests
 import torch
 from torch.utils.model_zoo import tqdm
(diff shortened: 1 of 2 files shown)


torchvision/datasets/utils.py
```diff
@@ -483,3 +484,39 @@ def verify_str_arg(
         raise ValueError(msg)
 
     return value
+
+
+def read_pfm(file_name: str, slice_channels: int = 2) -> np.ndarray:
```
NicolasHug, line 489: Let's keep it private, we don't really need to expose it  to users

```suggestion
def _read_pfm(file_name: str, slice_channels: int = 2) -> np.ndarray:
```

torchvision/datasets/utils.py
```diff
@@ -483,3 +484,39 @@ def verify_str_arg(
...
+    """Read file in .pfm format. Might contain
```
NicolasHug, line 490: Looks like this sentence is unfinished?

torchvision/datasets/utils.py
```diff
@@ -483,3 +484,39 @@ def verify_str_arg(
...
+
+    Args:
+        file_name (str): Path to the file.
+        slice_channels (int): Number of channels to slice out of the file.
+        Useful for reading different data formats stored in .pfm files: Optical Flows, Stereo Disparity Maps, etc.
```
NicolasHug, line 495: Skipping a line in a parameter description requires to indent it to the right (otherwise sphinx doesn't render it properly):

```suggestion
            Useful for reading different data formats stored in .pfm files: Optical Flows, Stereo Disparity Maps, etc.
```

NicolasHug (commented):
Thanks @TeodorPoncu , LGTM. The only important comment I have is to keep it private as suggested below

NicolasHug (approved):
Thanks @TeodorPoncu ! LGTM when CI is happy

PR #6271, opened 2022-07-14
[proto] Added RandomCrop transform and tests

vfdev-5:
Description:

- Added RandomCrop transform and tests


torchvision/prototype/transforms/_utils.py
```diff
@@ -2,17 +2,28 @@
...
 def query_image(sample: Any) -> Union[PIL.Image.Image, torch.Tensor, features.Image]:
+    flat_sample, _ = tree_flatten(sample)
+    for i in flat_sample:
+        if type(i) == torch.Tensor or isinstance(i, (PIL.Image.Image, features.Image)):
+            return i
+
+    raise TypeError("No image was found in the sample")
+
+
+# vfdev-5: let's use tree_flatten instead of query_recursively and internal fn to make the code simplier
```
vfdev-5, line 21: Let's simplify the code using `tree_flatten` instead of home-made methods.

torchvision/prototype/transforms/_geometry.py
```diff
@@ -416,3 +421,75 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
...
+        return [F.crop(flat_input, **params) for flat_input in flat_inputs]
+
+    def forward(self, *inputs: Any) -> Any:
+        from torch.utils._pytree import tree_flatten, tree_unflatten
+
+        sample = inputs if len(inputs) > 1 else inputs[0]
+
+        flat_inputs, spec = tree_flatten(sample)
+        out_flat_inputs = self._forward(flat_inputs)
+        return tree_unflatten(out_flat_inputs, spec)
```
vfdev-5, line 495: Here, we can't just override `_transform` method as we need to access image data and generate params on transformed sample. 
Proposed solution is to 1) flatten sample structure into a list, 2) apply crop+pad logic, 3) unflatten output into input sample structure

PR #6282, opened 2022-07-17
add type-var ignore for the recently introduced mypy error

anilbey:
A new error introduced to mypy is detected in vision's repository. 

> vision (https://github.com/pytorch/vision)
> + torchvision/prototype/datasets/utils/_internal.py:89: error: A function returning TypeVar should receive at least one argument containing the same Typevar  [type-var]
> 

This PR adds a type ignore to that error.

Reference: https://github.com/python/mypy/pull/13166


hauntsaninja (commented):
This should probably be typed as returning Any.

Also it looks like `path_accessor` might need an overload to be typed fully correctly

PR #6284, opened 2022-07-18
[proto] Added RandomPerspective and tests

vfdev-5:
Description:
- Added RandomPerspective and tests
- replaced real image creation by mocks for other tests

### DIFF
--- a/torchvision/prototype/transforms/functional/_geometry.py
+++ b/torchvision/prototype/transforms/functional/_geometry.py
@@ -11,6 +11,7 @@ from torchvision.transforms.functional import (
     _get_inverse_affine_matrix,
     InterpolationMode,
     _compute_output_size,
+    _get_perspective_coeffs,
 )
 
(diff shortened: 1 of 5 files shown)


torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -765,10 +766,13 @@ def perspective_segmentation_mask(img: torch.Tensor, perspective_coeffs: List[fl
 
 def perspective(
     inpt: DType,
-    perspective_coeffs: List[float],
+    startpoints: List[List[int]],
```
vfdev-5, line 769: Here, I missed in previous work that `perspective` mid-level op have to match `F.perspective` from stable one. So, now mid-level `perspective` matches: 
https://github.com/pytorch/vision/blob/3db30442a52e250c569ba131485fc432c8300f16/torchvision/transforms/functional.py#L683-L689

PR #6290, opened 2022-07-20
ci: Update macos runner to self hosted

seemethere:
Updates macos runner to new self hosted runners from AWS


datumbox:
We got some weird [failings](https://github.com/pytorch/vision/runs/7448134823?check_suite_focus=true) which are probably unrelated:
```
2022-07-21T11:40:40.4916820Z         t = transforms.RandomCrop(48)
2022-07-21T11:40:40.4916950Z         img = torch.ones(3, 32, 32)
2022-07-21T11:40:40.4917140Z         with pytest.raises(ValueError, match=r"Required crop size .+ is larger then input image size .+"):
2022-07-21T11:40:40.4917310Z >           t(img)
2022-07-21T11:40:40.4917610Z E           AssertionError: Regex pattern 'Required crop size .+ is larger then input image size .+' does not match 'Required crop size (48, 48) is larger than input image size (32, 32)'.
```

Let me update the branch with latest main and restart tests.

datumbox (approved):
The changes LGTM. We should merge on Green CI.

@seemethere Given that now we have 45 runners for M1, we would like run these tests not after merging on main but also on the PRs. This will help us move towards beta support of M1. Given our resources are still constraint, we consider running the full unit-tests on a single python config. Any thoughts on this?

seemethere:
> The changes LGTM. We should merge on Green CI.
> 
> @seemethere Given that now we have 45 runners for M1, we would like run these tests not after merging on main but also on the PRs. This will help us move towards beta support of M1. Given our resources are still constraint, we consider running the full unit-tests on a single python config. Any thoughts on this?

Should be fine, but would still limit it to 1 python config for the time being

datumbox:
@seemethere I'm a bit confused on why this is [failing](https://app.circleci.com/pipelines/github/pytorch/vision/18973/workflows/4049546e-f45a-4161-adcb-29dbb3bc9a1e/jobs/1535719). The issue seems unrelated as the changes don't affect CircleCI. Nevertheless while running the same test on main branch (see dummy PR #5009), it [passes](https://app.circleci.com/pipelines/github/pytorch/vision/18982/workflows/ed683e80-6532-466e-b0c7-f7a683cb4c7d/jobs/1535677). The issue seems related to the dependency with PyTorch. Any idea?

seemethere:
> @seemethere I'm a bit confused on why this is [failing](https://app.circleci.com/pipelines/github/pytorch/vision/18973/workflows/4049546e-f45a-4161-adcb-29dbb3bc9a1e/jobs/1535719). The issue seems unrelated as the changes don't affect CircleCI. Nevertheless while running the same test on main branch (see dummy PR #5009), it [passes](https://app.circleci.com/pipelines/github/pytorch/vision/18982/workflows/ed683e80-6532-466e-b0c7-f7a683cb4c7d/jobs/1535677). The issue seems related to a dependency with PyTorch. Any idea?

I think I understand the issue here, the windows runners don't have python3 installed by default so it should default to `python` when `python3` isn't available

NicolasHug:
Is this something we need to cherry-pick for the bugfix release @seemethere ?

PR #6291, opened 2022-07-20
Fix grammatical error

craymichael:
Change "then" to "than" (minor fix)


NicolasHug (approved):
Thank you @craymichael

PR #6295, opened 2022-07-21
[proto] Added elastic transform and tests

vfdev-5:


torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -783,6 +787,90 @@ def perspective(
...
+    ).view(-1, 4)
+
+    # Question (vfdev-5): should we rely on good displacement shape and fetch image size from it
+    # Or add image_size arg and check displacement shape
+    image_size = displacement.shape[-3], displacement.shape[-2]
+
+    id_grid = _FT._create_identity_grid(list(image_size)).to(bounding_box.device)
+    # We construct an approximation of inverse grid as inv_grid = id_grid - displacement
+    # This is not an exact inverse of the grid
+    inv_grid = id_grid - displacement
```
vfdev-5, line 831: This is an approximation which should work on small displacements.
Either we keep that and make user aware of that. Either we say that bbox transformation is not supported.

PR #6218, opened 2022-06-29
Move from xcode v12 to v14

datumbox:
Fixes #6210

PR #6298, opened 2022-07-21
Set MACOSX_DEPLOYMENT_TARGET=10.9 for binary jobs

NicolasHug:
Similar to https://github.com/pytorch/text/pull/1835, thanks @mthrok for the heads up.

TL;DR: in https://github.com/pytorch/vision/pull/6218 we bumped our xcode version to v14, which led to create binaries that are only compatible with macosx_11_0 (see recent builds in https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html). For compatibility with the rest of the ecosystem, we need to keep compatibility with 10.9.

This will need to be cherry-picked in the bugfix release.

CC @atalman


atalman (approved):
LGTM

datumbox (approved):
LGTM, thanks for fixing this.

PR #6272, opened 2022-07-14
[proto] Updated center_crop tests

vfdev-5:
Description:
- Added missing decorator for `center_crop_segmentation_mask` tests
- Added missing `center_crop_image_tensor` test

PR #6305, opened 2022-07-22
[proto] Ported all transforms to the new API

vfdev-5:
## Description:
- Ported all transforms to the new API
- Added tests

~Please merge https://github.com/pytorch/vision/pull/6272 before this PR~

## What remains

- Polish sharp edges:
  - Search and fix for TODOs in `torchvision/prototype/transforms` and `test/test_prototype_*`. Mostly those to implement a missing functionality.
- Better testing:
  - check if we test all ops and transforms
  - functional ops have to be tested on result correctness, jit/eager, devices etc. Philip was proposing to provide a simplified way to test that.
  - transforms have to be tested on parameter generation and checking if functional counterpart is correctly called via mocking (this is already done for majority of transforms in test_prototype_transforms.py)
  - we need to test specific `Feature.op` for example `BoundingBox.rotate` as they change metadata like `image_size`.

CI Failures to take into account:
- FAILED test/test_prototype_transforms_functional.py::test_eager_vs_scripted[gaussian_blur_image_tensor-26]
```
____________ test_eager_vs_scripted[gaussian_blur_image_tensor-26] _____________
  File "/home/runner/work/vision/vision/test/test_prototype_transforms_functional.py", line 740, in test_eager_vs_scripted
    msg=msg,
```


torchvision/prototype/transforms/_geometry.py
```diff
@@ -225,6 +225,19 @@ def _check_fill_arg(fill: Union[int, float, Sequence[int], Sequence[float]]) ->
...
 
+def _check_padding_arg(padding: Union[int, Sequence[int]]) -> None:
+    if not isinstance(padding, (numbers.Number, tuple, list)):
+        raise TypeError("Got inappropriate padding arg")
+
+    if isinstance(padding, (tuple, list)) and len(padding) not in [1, 2, 4]:
+        raise ValueError(f"Padding must be an int or a 1, 2, or 4 element tuple, not a {len(padding)} element tuple")
+
+
+def _check_padding_mode_arg(padding_mode: Literal["constant", "edge", "reflect", "symmetric"]) -> None:
```
datumbox, line 236: Nit: We should consider the use of enums here; this a textbook case for them. Using `torchvision._utils.StrEnum` will enable us to have the best of both worlds (strings and enums): staying aligned with Core while at the same time giving the option to our users to use idioms friendly to static analysis.

torchvision/prototype/transforms/_geometry.py
```diff
@@ -414,3 +422,201 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
...
+            )
+
+        if width == output_width and height == output_height:
+            return dict(top=0, left=0, height=height, width=width)
+
+        top = torch.randint(0, height - output_height + 1, size=(1,)).item()
+        left = torch.randint(0, width - output_width + 1, size=(1,)).item()
+        return dict(top=top, left=left, height=output_height, width=output_width)
+
+    def _forward(self, flat_inputs: List[Any]) -> List[Any]:
```
datumbox, line 469: My understanding is that `_forward` receives the whole flattened input while `_transform` would require a single input. But why do we need another new idiom here? We should avoid introducing multiple idioms on how to structure the Transforms if we can.

Can we move some of the param estimations done here on the `_get_params()` and use the rewrite this to do all ops in a `_transform` individually?

torchvision/prototype/transforms/_geometry.py
```diff
@@ -414,3 +422,201 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
...
+
+        image = query_image(flat_inputs)
+        _, height, width = get_image_dimensions(image)
+
+        # pad the width if needed
+        if self.pad_if_needed and width < self.size[1]:
+            padding = [self.size[1] - width, 0]
+            flat_inputs = [F.pad(flat_input, padding, self.fill, self.padding_mode) for flat_input in flat_inputs]
+        # pad the height if needed
+        if self.pad_if_needed and height < self.size[0]:
```
datumbox, line 481: Nit: Use nested ifs to avoid `if self.pad_if_needed` twice?

torchvision/prototype/transforms/_geometry.py
```diff
@@ -414,3 +422,201 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
...
+        return [F.crop(flat_input, **params) for flat_input in flat_inputs]
+
+    def forward(self, *inputs: Any) -> Any:
+        from torch.utils._pytree import tree_flatten, tree_unflatten
+
+        sample = inputs if len(inputs) > 1 else inputs[0]
+
+        flat_inputs, spec = tree_flatten(sample)
+        out_flat_inputs = self._forward(flat_inputs)
+        return tree_unflatten(out_flat_inputs, spec)
```
datumbox, line 496: Good one. Can we replace our custom `apply_recursively()` method used on the default forward with this one?

torchvision/prototype/transforms/_utils.py
```diff
@@ -2,17 +2,28 @@
...
+    flat_sample, _ = tree_flatten(sample)
+    for i in flat_sample:
+        if type(i) == torch.Tensor or isinstance(i, (PIL.Image.Image, features.Image)):
+            return i
+
+    raise TypeError("No image was found in the sample")
+
+
+# vfdev-5: let's use tree_flatten instead of query_recursively and internal fn to make the code simplier
+def query_image_(sample: Any) -> Union[PIL.Image.Image, torch.Tensor, features.Image]:
```
datumbox, line 22: Are we still using this? Can we remove the old method?

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -779,6 +787,90 @@ def perspective(
...
+
+
+def elastic_image_pil(
+    img: PIL.Image.Image,
+    displacement: torch.Tensor,
+    interpolation: InterpolationMode = InterpolationMode.BILINEAR,
+    fill: Optional[Union[int, float, Sequence[int], Sequence[float]]] = None,
+) -> PIL.Image.Image:
+    t_img = pil_to_tensor(img)
+    fill = _convert_fill_arg(fill)
```
datumbox, line 806: It's a bit weird that `_convert_fill_arg()` calls exist within `*_pil` kernels but not for `_tensor` ones. I understand you do this because that's not JIT-scriptable for tensor and that we don't care about JIT-scriptability for PIL. What are the implications of phasing this out from all `*_pil` kernels and instead move them to the high-level kernels? This comment applies to all kernels.

datumbox (commented):
Thanks @vfdev-5, looks good overall. I've added a few comments, let me know what you think. I'm not going to check the tests because as you indicate more discussions will be required over the testing strategy.

torchvision/prototype/transforms/_color.py
```diff
@@ -151,8 +151,42 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
 
 
 class RandomEqualize(_RandomApplyTransform):
-    def __init__(self, p: float = 0.5):
+    def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
+        return F.equalize(inpt)
```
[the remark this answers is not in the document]
vfdev-5: Yes, you are right, we are not bypassing int, float, strings. Do you think we could do that in `Transform` class as following:
```python
from typing import Any, Dict

import torch
import torch.nn as nn
from torch.utils._pytree import tree_flatten, tree_unflatten


class Transform(nn.Module):
    def __init__(self) -> None:
        super().__init__()

        return dict()

        raise NotImplementedError

        params=self._get_params(sample)
        
        flat_inputs, spec = tree_flatten(sample)
        # pass-through types
        pass_types = (int, float, str)
        flat_outputs = [
            self._transform(inpt, params) if not isinstance(inpt, pass_types) else inpt
            for inpt in flat_inputs
        ]
        return tree_unflatten(flat_outputs, spec)
```
?

This way we do not need to make if/else in each derived transform.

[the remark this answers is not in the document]
datumbox: Yes, I'm OK with that for now.

torchvision/prototype/transforms/_geometry.py
vfdev-5: Let me do that in a follow-up PR

datumbox: Could you just add a TODO comment?

torchvision/prototype/transforms/functional/_geometry.py
vfdev-5: This is a particular case when we call `elastic_image_tensor` from `elastic_image_pil` as we do not implement natively elastic transformation on pillow. So, the sequence of 2 ops:
```
fill = _convert_fill_arg(fill)
output = elastic_image_tensor(t_img, displacement, interpolation=interpolation, fill=fill)
```
is the same as in a Image.elastic implementation: https://github.com/pytorch/vision/pull/6305/files#diff-f49325ccedde943495459399bf815a68db299bdd344f2e8309e87934a106b79cR246-R257

datumbox: Right, but why not move all these `_convert_fill_arg()` calls directly on the high-level kernels to avoid calling them on the low-level kernels over and over?

vfdev-5: OK, I can remove it from here and do that for this particular case. I can see that there should be no other similar implementation.

vfdev-5: As discussed offline, let's keep the actual code as we keep uniform method signatures

torchvision/prototype/transforms/_geometry.py
vfdev-5: There can be 2 options:
-    We acknowledge that this Transform is different from the rest and overwrite the forward(). What you proposed above looks OK to me though it hides information on what it does and it's harder to read.
-    We try to estimate the sizes on get_params and then you put the whole implementation in _transform() as per all other transforms. (we still need to make sure if it feasible)

The option 1 is almost done and shows a way how to do things in a mixed way: _transform + other things.

Option 2, will require to create mid-level and low-level ops doing padded crop op (same as resized_crop) but this way (I have to verify that though if feasible) we keep following our main paradigm (implementing only _transform) with RandomCrop transform. A minor drawback of option2 is that we implictly show by our code to the users that a composed ops should be done at mid/low-level.
While option 1 shows that Transforms could be easily overridden and "hacked" as user wants ...

torchvision/prototype/transforms/_utils.py
```diff
@@ -36,16 +30,14 @@ def get_image_dimensions(image: Union[PIL.Image.Image, torch.Tensor, features.Im
...
 def has_any(sample: Any, *types: Type) -> bool:
-    return any(issubclass(type, types) for type in _extract_types(sample))
+    flat_sample, _ = tree_flatten(sample)
+    return any(issubclass(type(obj), types) for obj in flat_sample)
 
 
 def has_all(sample: Any, *types: Type) -> bool:
-    return not bool(set(types) - set(_extract_types(sample)))
+    flat_sample, _ = tree_flatten(sample)
+    return not bool(set(types) - set([type(obj) for obj in flat_sample]))
```
datumbox, line 40: This will fail if we subclass one of the types (say we subclass Feature.Image) but since that's not the intention now, we are OK to keep it as-is.

torchvision/transforms/functional_pil.py
```diff
@@ -260,7 +260,7 @@ def _parse_fill(
 ) -> Dict[str, Optional[Union[float, List[float], Tuple[float, ...]]]]:
 
     # Process fill color for affine transforms
-    num_bands = len(img.getbands())
+    num_bands = get_image_num_channels(img)
```
datumbox, line 263: Nit: Instead of `num_bands` use `num_channels`?

datumbox (approved):
LGTM, I haven't checked the testing strategy but I know @pmeier's plan is to revise it thoroughly so I'm happy to go with whatever you prefer for now.

PR #6310, opened 2022-07-25
add crestereo implementation

TeodorPoncu:
This is a draft version for the [CREStereo](https://openaccess.thecvf.com/content/CVPR2022/papers/Li_Practical_Stereo_Matching_via_Cascaded_Recurrent_Network_With_Adaptive_Correlation_CVPR_2022_paper.pdf) model implementation. 

It's over-commented for easier readability for those that are unfamiliar with the paper.

It also makes use of ModuleDicts and Dicts to process / access feature pyramid level instead of ModuleLists and Lists for readability purposes.


jdsgomes (commented):
First pass, looks good in general so only left a few comments

torchvision/prototype/models/depth/stereo/crestereo.py
```diff
@@ -0,0 +1,1021 @@
...
+        search_dilate_2d=corr_search_dilate_2d,
+    )
+
+    if weights is not None:
+        model.load_state_dict(weights.get_state_dict(progress=progress))
+
+    return model
+
+
+def crestereo_base(*, weights: Optional[WeightsEnum] = None, progress=True, **kwargs) -> CREStereo:
```
jdsgomes, line 989: A couple of comments regarding this constructor:

1) the weights enum should have a more specific type, eg, `CREStereo_Weights`, you can create an empty class for now as a placeholder until we have the metadata to fill: 

```
class CREStereo(CREStereo_Weights):
    pass
```

2) If we are not planning to release other variants in terms of architecture, this model builder could be just called `crestereo`. If more variants are planned we could follow the same naming standard as in other models and just append b to the name (`crestereo_b` and `CREStereo_B_Weights`)

torchvision/prototype/models/depth/stereo/crestereo.py
```diff
@@ -0,0 +1,1021 @@
...
+        self.factors = factors
+
+    def forward(self, x: torch.Tensor) -> List[Tensor]:
+        results = [x]
+        for factor in self.factors:
+            results.append(F.avg_pool2d(x, kernel_size=factor, stride=factor))
+        return results
+
+
+class CREStereo(nn.Module):
```
jdsgomes, line 603: please add an args section in the docstring (e.g. https://github.com/pytorch/vision/blob/main/torchvision/models/swin_transformer.py#L502)

jdsgomes:
you also need to add `from .crestereo import *` to `torchvision/prototype/models/depth/stereo/__init__.py`

torchvision/prototype/models/depth/stereo/crestereo.py
```diff
@@ -0,0 +1,1021 @@
...
+import numpy as np
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+import torchvision.models.optical_flow.raft as raft
+from torch import Tensor
+from torchvision.models._api import WeightsEnum
+from torchvision.models.optical_flow._utils import make_coords_grid, grid_sample, upsample_flow
+from torchvision.ops import Conv2dNormActivation
+
```
jdsgomes, line 14: please add __all__ = (
    "CREStereo",
   "CREStereo_Weights",
    "crestereo_base",
)
 
in this section.

torchvision/prototype/models/depth/stereo/crestereo.py
```diff
@@ -0,0 +1,1021 @@
...
+        layers: Tuple[int, int, int, int, int] = (64, 64, 96, 128, 256),
+        strides: Tuple[int, int, int, int] = (2, 1, 2, 1),
+        norm_layer: Callable[..., nn.Module] = nn.InstanceNorm2d,
+    ):
+        super().__init__(block=block, layers=layers, strides=strides, norm_layer=norm_layer)
+        for s in strides:
+            if s not in [1, 2]:
+                raise ValueError(f"FeatureEncoder unsupported stride size {s}. Supported values are one of ``[1, 2]``.")
+
+        self.output_dim = layers[-1]
```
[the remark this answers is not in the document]
TeodorPoncu: That was actually used to dynamically construct to keys for the ModuleDicts. Although I do not expect someone to wish to change the default architecture, I just thought it would be nice to keep it in case someone might change the strides for the FeatureEncoder

torchvision/prototype/models/depth/stereo/crestereo.py
```diff
@@ -0,0 +1,1020 @@
...
+            # i.e ["1 / 16", "1 / 8", "1 / 4"]
+            # the correlation layer ModuleDict has layers for all the resolutions except the fine one
+            # i.e {"1 / 16": Module, "1 / 8": Module}
+            # for these resolution we perform only half of the number of refinement iterations
+            for idx, (resolution, correlation_layer) in enumerate(self.correlation_layers.items()):
+                # compute the scale difference between the first pyramid scale and the current pyramid scale
+                scale_to_base = l_pyramid[fine_grained_resolution].shape[2] // l_pyramid[resolution].shape[2]
+                for it in range(iterations // 2):
+                    # set wether or not we want to search on (X, Y) axes for correlation or just on X axis
+                    use_small_search_patch = (it % 2) == 1
```
YosuaMichael, line 788: NIT, I think we should change the variable name to be like `use_2d_search` instead of `use_small_search_patch`. 
The variable `use_small_search_patch` kinda emphasize on `small` which I think not the case here.

torchvision/prototype/models/depth/stereo/crestereo.py
```diff
@@ -0,0 +1,1020 @@
...
+                        flow_estimates[resolution], up_mask, factor=self.downsampling_factors[0]
+                    )
+                    # we then bilinear upsample to the final resolution
+                    # we use a factor that's equivalent to the difference between
+                    # the current downsample resolution and the base downsample resolution
+                    #
+                    # i.e. if a 1 / 16 flow is upsampled by 4 (base downsampling) we get a 1 / 4 flow.
+                    # therefore we have to further upscale it by the difference between
+                    # the current level 1 / 16 and the base level 1 / 4.
+                    flow_pred = -upsample_flow(flow_pred_prior, None, factor=scale_to_base)
```
YosuaMichael, line 821: Question, is this `-` sign because we use `-scale` before? I wonder if we can remove all the `-` sign...

torchvision/prototype/models/depth/stereo/crestereo.py
```diff
@@ -0,0 +1,1020 @@
...
+                    predictions.append(flow_pred)
+
+                # when constructing the next resolution prior, we resample w.r.t
+                # to the scale of the next level in the pyramid
+                next_resolution = coarse_resolutions[idx + 1]
+                scale_to_next = l_pyramid[next_resolution].shape[2] / flow_pred_prior.shape[2]
+                # we use the flow_up_prior because this is a more accurate estimation of the true flow
+                # due to the convex upsample, which resembles a learned super-resolution module.
+                # this is not necessarily an upsample, it can be a downsample, based on the provided configuration
+                flow_estimates[next_resolution] = -scale_to_next * F.interpolate(
```
YosuaMichael, line 831: Just another comment related on `-scale`

torchvision/prototype/models/depth/stereo/crestereo.py
```diff
@@ -0,0 +1,1020 @@
...
+            )
+
+            net_pyramid[max_res], delta_flow = self.update_block(
+                net_pyramid[max_res], ctx_pyramid[max_res], correlations, flow_estimates[max_res]
+            )
+
+            up_mask = self.mask_predictor(net_pyramid[max_res])
+            flow_estimates[max_res] = flow_estimates[max_res] + delta_flow
+            # at the final resolution we simply do a convex upsample using the base downsample rate
+            flow_pred = -upsample_flow(flow_estimates[max_res], up_mask, factor=self.downsampling_factors[0])
```
YosuaMichael, line 869: Just another comment related to `-scale`

torchvision/prototype/models/depth/stereo/crestereo.py
```diff
@@ -0,0 +1,1020 @@
...
+        model.load_state_dict(weights.get_state_dict(progress=progress))
+
+    return model
+
+
+class CREStereo_Weights(WeightsEnum):
+    pass
+
+
+class CREStereo_B_Weights(CREStereo_Weights):
```
YosuaMichael, line 983: We usually just inherit from `WeightsEnum`.
Another thing, do you expect to have variant other than `crestereo_b` ? If not, I think we can just name it `crestereo`.

YosuaMichael (commented):
Nice work @TeodorPoncu ! Overall it looks pretty good. I have several comments that mostly NIT.

torchvision/prototype/models/depth/stereo/crestereo.py
TeodorPoncu: @YosuaMichael Going through all the `-scale` I agree that it might be the case to try removing them. However looking at the base case which is this final for loop; assuming we receive a valid flow prior for this resolution (cascaded inference) I believe the - might be actually necessary since there is no sign inversion at this stage. Notice that the `flow_estimate[max_res] ` is updated solely with `delta_flow`. My intuition is that it somehow aids with the convex upsampling function.

torchvision/prototype/models/depth/stereo/crestereo.py
```diff
@@ -0,0 +1,1020 @@
...
+        )
+
+        self.multiplier = multiplier
+
+    def forward(self, x: Tensor) -> Tensor:
+        x = self.mask_head(x) * self.multiplier
+        return x
+
+
+class AdaptiveGroupCorrelationLayer(nn.Module):
```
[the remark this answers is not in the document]
TeodorPoncu: I'm not particularly sure what the best approach of breaking it down might be. Breaking it in independent modules will add more complexity in the forward loop. What I believe might be done to reduce the size of this module would be to move `get_correlation` outside the module in a simple function.

My reasoning for keeping everything together is such that if one was to look at the paper, each individual layers code would be functionally grouped together.

Splitting this into multiple layers that you need to keep track of for each individual pyramid level in the forward pass seems prone to risks in terms of Jit and readability since now we have to dynamically elect which layer we use based on two criteria:
1.  At what pyramid level we are (since only the lowest level contains an attention module)
2. At what recursive step level we are (which basically selects the window shape)
3. Which inputs are we passing going to pass into the layers (all pyramid levels but the last one have offsets)

torchvision/prototype/models/depth/stereo/crestereo.py
```diff
@@ -0,0 +1,1020 @@
...
+        in_channels: int,
+        hidden_size: int,
+        upsample_factor: int,
+        multiplier: float = 0.25,
+    ) -> None:
+
+        super().__init__()
+        self.mask_head = nn.Sequential(
+            Conv2dNormActivation(in_channels, hidden_size, norm_layer=None, kernel_size=3),
+            nn.Conv2d(hidden_size, upsample_factor**2 * 9, 1, padding=0),
```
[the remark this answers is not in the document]
TeodorPoncu: This boils down from the convex upsample formula presented in the Raft Appendix. Each high resolution pixel (the total number of pixels increase by a factor of `upsample x upsample`) is a weighted sum from the convex neighbours of a low resolution pixels (the convex neighbours of a pixels consists of a 3x3 window placed on top of said pixel, hence the 9).

torchvision/prototype/models/depth/stereo/crestereo.py
TeodorPoncu: I mostly expect to have variants that are pretrained at either different resolutions, different number of optimization steps, different number of datasets or different losses.

torchvision/prototype/models/depth/stereo/crestereo.py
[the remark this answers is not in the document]
YosuaMichael: Is this 9 depend on the patch size? If I remember correctly we wave patch size of (1, 9) or (3,3) which in total 9 pixels?
Or is it always assumed we use 3x3 windows for the convex upsample?

torchvision/prototype/models/depth/stereo/crestereo.py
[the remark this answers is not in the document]
YosuaMichael: One possible way I can think on top of my head now is to have:
- `IterativeCorrelation (nn.Module)`
- `AttentionOffsetCorrelation (nn.Module)`
- Then they are combined again in `AdaptiveGroupCorrelationLayer (nn.Module)`

But no strong opinion on this though, it is fine to leave as it is now.

torchvision/prototype/models/depth/stereo/crestereo.py
YosuaMichael: In that case, we can have `CREStereo_B_Weights (WeightsEnum)` and you can remove the class `CREStereo_Weights`

torchvision/prototype/models/depth/stereo/crestereo.py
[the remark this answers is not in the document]
TeodorPoncu: No, the 9 does not depend on the correlation window. In pixel space, one pixel has 8 neighbours, and the convex region of that pixel consists of the pixel itself and its neighbours.

Reference: https://arxiv.org/pdf/2003.12039.pdf page 20 section B

torchvision/prototype/models/depth/stereo/crestereo.py
YosuaMichael: I see, I think we should still have a capability to port the author's weight. In that case, lets keep the `-scale` in my opinion.

torchvision/prototype/models/depth/stereo/crestereo.py
```diff
@@ -0,0 +1,1100 @@
...
+            right_feature = right_feature.permute(0, 2, 3, 1).reshape(B, H * W, C)
+            # this can be either self attention or cross attention, hence the tupple return
+            left_feature, right_feature = self.attention_module(left_feature, right_feature)
+            left_feature = left_feature.reshape(B, H, W, C).permute(0, 3, 1, 2)
+            right_feature = right_feature.reshape(B, H, W, C).permute(0, 3, 1, 2)
+
+        left_groups = torch.chunk(left_feature, self.groups, dim=1)
+        right_groups = torch.chunk(right_feature, self.groups, dim=1)
+
+        num_search_candidates = self.search_window_2d[1] * self.search_window_2d[0]
```
YosuaMichael, line 251: Question, If the input `window_type` is `"1d"`, do we still use `self.search_window_2d` here?

torchvision/prototype/models/depth/stereo/crestereo.py
```diff
@@ -0,0 +1,1100 @@
...
+
+        self.iterative_correlation_layer = iterative_correlation_layer
+        self.attention_offset_correlation_layer = attention_offset_correlation_layer
+
+    def forward(
+        self,
+        left_features: Tensor,
+        right_features: Tensor,
+        flow: torch.Tensor,
+        extra_offset: Union[torch.Tensor, None],
```
YosuaMichael, line 327: NIT, lets use `Optional[Tensor]` instead of `Union[torch.Tensor, None]`. Also you might want to add default value `None` ?

Also NIT on the previous line, to be consistent lets use `Tensor` instead of `torch.Tensor`.

torchvision/prototype/models/depth/stereo/crestereo.py
```diff
@@ -0,0 +1,1100 @@
...
+
+        Returns:
+            x: [B, C, H, W]
+        """
+        torch._assert(
+            len(x.shape) == 4,
+            f"PositionalEncodingSine requires a 4-D dimensional input. Provided tensor is of shape {x.shape}",
+        )
+
+        self.pe: Tensor
```
YosuaMichael, line 481: Why do we need this line?

YosuaMichael (commented):
Hi @TeodorPoncu, overall looks good, I just have a few more comments

torchvision/prototype/models/depth/stereo/crestereo.py
TeodorPoncu: This line exists for mypy checking, since it does not correctly parse the `register_buffer` method and its effects.

torchvision/prototype/models/depth/stereo/crestereo.py
```diff
@@ -0,0 +1,1096 @@
...
+            "1d": [search_window_1d for _ in range(self.groups)],
+        }
+
+        self.dilate_sizes = {
+            "2d": [search_dilate_2d for _ in range(self.groups)],
+            "1d": [search_dilate_1d for _ in range(self.groups)],
+        }
+
+    def _make_coords(self, feature_map: Tensor) -> Tensor:
+        return make_coords_grid(feature_map.shape[0], feature_map.shape[2], feature_map.shape[3]).to(feature_map.device)
```
YosuaMichael, line 151: NIT, In this PR you introduce `device` param on `make_coords_grid`, I think in that case we should use it instead of using `.to(device)`. Also since this is only called once in this Class, maybe we dont need to have `_make_coords` and call the `make_coords_grid` directly

torchvision/prototype/models/depth/stereo/crestereo.py
```diff
@@ -0,0 +1,1096 @@
...
+
+        self.dilate_sizes = {
+            "2d": [search_dilate_2d for _ in range(self.groups)],
+            "1d": [search_dilate_1d for _ in range(self.groups)],
+        }
+
+        self.attention_module = attention_module
+
+    def _make_coords(self, feature_map: Tensor) -> Tensor:
+        return make_coords_grid(feature_map.shape[0], feature_map.shape[2], feature_map.shape[3]).to(feature_map.device)
```
YosuaMichael, line 217: Similar to previous comment in IterativeCorrelationLayer

torchvision/prototype/models/depth/stereo/crestereo.py
```diff
@@ -0,0 +1,1096 @@
...
+    def __init__(
+        self,
+        *,
+        dim_model: int,
+        num_heads: int,
+        attention_module: Callable[..., nn.Module] = LinearAttention,
+    ) -> None:
+        super().__init__()
+
+        attention_module = attention_module()
```
YosuaMichael, line 497: NIT, we can use `self.attention_op` directly instead of having temporary `attention_module`.

YosuaMichael (commented):
Hi @TeodorPoncu, I think this is very close to a merge. Just have few things left that you forgot to update and NITs.

YosuaMichael:
@TeodorPoncu Oh ya, also please resolve the conflict on the test files

YosuaMichael (approved):
Thanks for your hard work @TeodorPoncu !
The PR looks good to me!

@jdsgomes could you give a one more pass of review as well? Thanks!

jdsgomes (approved):
LGTM, thank you for addressing all the comments. I think the weights can be added in a separate PR.

torchvision/models/optical_flow/_utils.py
```diff
@@ -19,8 +19,8 @@ def grid_sample(img: Tensor, absolute_grid: Tensor, mode: str = "bilinear", alig
     return F.grid_sample(img, normalized_grid, mode=mode, align_corners=align_corners)
 
 
-def make_coords_grid(batch_size: int, h: int, w: int):
-    coords = torch.meshgrid(torch.arange(h), torch.arange(w), indexing="ij")
+def make_coords_grid(batch_size: int, h: int, w: int, device: torch.device = torch.device("cpu")):
```
YosuaMichael, line 22: @TeodorPoncu I notice on another PR @jdsgomes comment this: https://github.com/pytorch/vision/pull/6554#discussion_r969626378

We might want to have similar thing here.

PR #6329, opened 2022-07-28
Replaced python built-in input by inpt

vfdev-5:


datumbox (approved):
LGTM, thanks!

pmeier:
`input` is the standard name that `torch.nn` uses:

- https://github.com/pytorch/pytorch/blob/7191ae58a75e58e3b5bca67d4758303fcbbc050f/torch/nn/modules/module.py#L229
- https://github.com/pytorch/pytorch/blob/7191ae58a75e58e3b5bca67d4758303fcbbc050f/torch/nn/modules/module.py#L1160

I don't think we ever want to use the builtin `input` function and thus overwriting it should be fine. Now we have this disconnect between the `forward` and the `_transform`:

https://github.com/pytorch/vision/blob/cc0a83856c979f834132a5faf71a2eece0d32dc3/torchvision/prototype/transforms/_transform.py#L24-L27

PR #6330, opened 2022-07-28
[RFC] Registration mechanism for models 

### LINKED ISSUE #1143: Is there a plan to add a registration mechanism to torchvision.models?
Hi,
As mentioned in the test scripts:

https://github.com/pytorch/vision/blob/c187c2b12d86c3909e59a40dbe49555d85b98703/test/test_models.py#L9

I want to know whether the registration mechanism is adding to `./torchvision/models/resnet.py` or somewhere like this?

datumbox:
Fixes #1143

**Note:** This PR is meant to document the RFC with code examples and it's not mean to be merged. Some tests are failing because they are not compatible with the mechanism. This will be resolved on the follow up PRs that will actually introduce the mechanism to TorchVision.

## Objective

Create a model registration mechanism responsible for:
- Listing the names of all available models under a specific module
- Getting a model (initializing) from its name
- Getting the weight enum of the model from its name or model builder
- Registering models under a specific name

## Background

As discussed at #5088, a model registration mechanism is needed to be able to quickly list the available modules.

Currently we use hacky workarounds in order to list our models:
https://github.com/pytorch/vision/blob/6ca9c76adb6daf2695d603ad623a9cf1c4f4806f/test/test_models.py#L24-L30

The same applies for initializing models from their names:
https://github.com/pytorch/vision/blob/e288f6ca01a5009ab32f1db010b31f14756e0303/references/classification/train.py#L223-L225

https://github.com/pytorch/vision/blob/6ca9c76adb6daf2695d603ad623a9cf1c4f4806f/torchvision/models/detection/backbone_utils.py#L112

## Overview

Offer a mechanism that stores all models on a single private global dictionary along with public getter/listing methods located under `torchvision.models` that will allow users to interact with the mechanism. This proposal aligns closely with the registration mechanism proposed at the Datasets V2 API, offering a similar user experience.

### List models

Listing all available models:
```python
>>> torchvision.models.list_models()
['alexnet', 'mobilenet_v3_large', 'mobilenet_v3_small', 'quantized_mobilenet_v3_large', ...]
```

Listing available models from a specific module:
```python
>>> torchvision.models.list_models(module=torchvision.models)
['alexnet', 'mobilenet_v3_large', 'mobilenet_v3_small', ...]
>>> torchvision.models.list_models(module=torchvision.models.quantization)
['quantized_mobilenet_v3_large', ...]

```

### Get models

Getting an initialized model with pre-trained weights from its name:
```python
>>> torchvision.models.get_model("quantized_mobilenet_v3_large", weights="DEFAULT")
QuantizableMobileNetV3(
  (features): Sequential(
   ....
   )
)
```

### Get weights

Getting the Weight Enum class of a model from its name:
```python
>>> torchvision.models.get_model_weights("quantized_mobilenet_v3_large")
<enum 'MobileNet_V3_Large_QuantizedWeights'>
```

Getting the Weight Enum class of a model from its callable (model builder):
```python
>>> torchvision.models.get_model_weights(torchvision.models.quantization.mobilenet_v3_large)
<enum 'MobileNet_V3_Large_QuantizedWeights'>
```

### Register models

To register a model we use a special decorator and provide an arbitrary name:
```python
@register_model(name="mobilenet_v3_large")
def mobilenet_v3_large(
    *, weights: Optional[MobileNet_V3_Large_Weights] = None, progress: bool = True, **kwargs: Any
) -> MobileNetV3:
    pass

@register_model(name="quantized_mobilenet_v3_large")
def mobilenet_v3_large(
    *,
    weights: Optional[Union[MobileNet_V3_Large_QuantizedWeights, MobileNet_V3_Large_Weights]] = None,
    progress: bool = True,
    quantize: bool = False,
    **kwargs: Any,
) -> QuantizableMobileNetV3:
    pass
```

If you are registering a method under the same as its method name, you can omit passing the name. The following call will register the method under the name `"some_model"`:
```python
@register_model()
def some_model():
    pass
```

Registering a model under an existing name throws an error.

### Other details
- We will keep the registration mechanism private. Perhaps on the future we can consider allowing users to use it to register custom models but that's not part of this RFC.
- Unfortunately the names of the methods under `torchvision.models.quantization` conflict with those under `torchvision.models`. As a result we need to offer the ability to register them under arbitrary names that resolve the conflicts.
- The current registration mechanism will throw an error if one tries to register a model under an existing name. It's possible to extend the mechanism to allow overwrites (similar to ClassyVision) but given that the mechanism is private we don't have an immediate need of this feature.

## Alternatives Considered

We considered offering separate registration APIs per submodule (aka one for `torchvision.models`, one for `torchvision.models.detection` etc). This would resolve the issue of name conflicts across submodules but would lead to having separate public methods per package. Moreover this would lead to an awkward design since methods like `torchvision.models.get_weight()` already support all model subpackages.

### DIFF
--- a/torchvision/models/__init__.py
+++ b/torchvision/models/__init__.py
@@ -14,4 +14,4 @@ from .vgg import *
 from .vision_transformer import *
 from .swin_transformer import *
 from . import detection, optical_flow, quantization, segmentation, video
-from ._api import get_model, get_model_weight, get_weight, list_models
+from ._api import get_model, get_model_weights, get_weight, list_models
--- a/torchvision/models/__init__.py
+++ b/torchvision/models/__init__.py
@@ -14,4 +14,4 @@ from .vgg import *
 from .vision_transformer import *
 from .swin_transformer import *
 from . import detection, optical_flow, quantization, segmentation, video
-from ._api import get_weight
+from ._api import get_weight, list_models, load
--- a/torchvision/models/__init__.py
+++ b/torchvision/models/__init__.py
@@ -14,4 +14,4 @@ from .vgg import *
 from .vision_transformer import *
 from .swin_transformer import *
 from . import detection, optical_flow, quantization, segmentation, video
-from ._api import get_weight, list_models, load_model
+from ._api import get_weight, list_models, get_model, get_model_weight
--- a/torchvision/models/__init__.py
+++ b/torchvision/models/__init__.py
@@ -14,4 +14,4 @@ from .vgg import *
 from .vision_transformer import *
 from .swin_transformer import *
 from . import detection, optical_flow, quantization, segmentation, video
-from ._api import get_weight, list_models, get_model, get_model_weight
+from ._api import get_model, get_model_weight, get_weight, list_models
--- a/torchvision/models/__init__.py
+++ b/torchvision/models/__init__.py
@@ -14,4 +14,4 @@ from .vgg import *
 from .vision_transformer import *
 from .swin_transformer import *
 from . import detection, optical_flow, quantization, segmentation, video
-from ._api import get_weight, list_models, load
+from ._api import get_weight, list_models, load_model
--- a/torchvision/models/_api.py
+++ b/torchvision/models/_api.py
@@ -162,7 +162,7 @@ def register_model(name: str, overwrite: bool = False) -> Callable[[Callable[...
 
 def list_models(module: Optional[ModuleType] = None) -> List[str]:
     """
-    Returns a list with the names of registred models.
+    Returns a list with the names of registered models.
 
     Args:
         module (ModuleType, optional): The module from which we want to extract the available models.
--- a/torchvision/models/_api.py
+++ b/torchvision/models/_api.py
@@ -13,7 +13,7 @@ from torchvision._utils import StrEnum
 from .._internally_replaced_utils import load_state_dict_from_url
 
 
-__all__ = ["WeightsEnum", "Weights", "get_weight"]
+__all__ = ["WeightsEnum", "Weights", "get_weight", "list_models", "get_model", "get_model_weight"]
 
 
 @dataclass
@@ -110,10 +110,26 @@ def get_weight(name: str) -> WeightsEnum:
     return weights_enum.from_str(value_name)
 
 
-def _get_enum_from_fn(fn: Callable) -> WeightsEnum:
+W = TypeVar("W", bound=Type[WeightsEnum])
+
+
+def get_model_weight(name: str) -> W:
+    """
+    Retuns the Weights Enum from the model name.
+
+    Args:
+        name (str): The name under which the model is registered.
+
+    Returns:
+        weights_enum (W): The weights enum class associated with the model.
+    """
+    fn = _find_model(name)
+    return _get_enum_from_fn(fn)
+
+
+def _get_enum_from_fn(fn: Callable) -> W:
     """
     Internal method that gets the weight enum of a specific model builder method.
-    Might be removed after the handle_legacy_interface is removed.
 
     Args:
         fn (Callable): The builder method used to create the model.
@@ -142,7 +158,7 @@ def _get_enum_from_fn(fn: Callable) -> WeightsEnum:
             "The WeightsEnum class for the specific method couldn't be retrieved. Make sure the typing info is correct."
         )
 
-    return cast(WeightsEnum, weights_enum)
+    return cast(W, weights_enum)
 
 
 M = TypeVar("M", bound=Type[nn.Module])
@@ -177,7 +193,16 @@ def list_models(module: Optional[ModuleType] = None) -> List[str]:
     return sorted(models)
 
 
-def load_model(name: str, **config: Any) -> M:
+def _find_model(name: str) -> Callable[..., M]:
+    name = name.lower()
+    try:
+        fn = BUILTIN_MODELS[name]
+    except KeyError:
+        raise ValueError(f"Unknown model {name}")
+    return fn
+
+
+def get_model(name: str, **config: Any) -> M:
     """
     Gets the model name and configuration and returns an instantiated model.
 
@@ -188,9 +213,5 @@ def load_model(name: str, **config: Any) -> M:
     Returns:
         model (nn.Module): The initialized model.
     """
-    name = name.lower()
-    try:
-        fn = BUILTIN_MODELS[name]
-    except KeyError:
-        raise ValueError(f"Unknown model {name}")
+    fn = _find_model(name)
     return fn(**config)
--- a/torchvision/models/_api.py
+++ b/torchvision/models/_api.py
@@ -3,7 +3,9 @@ import inspect
 import sys
 from dataclasses import dataclass, fields
 from inspect import signature
-from typing import Any, Callable, cast, Dict, Mapping
+from torch import nn
+from types import ModuleType
+from typing import Any, Callable, cast, Dict, List, Mapping, Optional, Type, TypeVar
 
 from torchvision._utils import StrEnum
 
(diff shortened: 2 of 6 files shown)


torchvision/models/_api.py
```diff
@@ -140,3 +143,53 @@ def _get_enum_from_fn(fn: Callable) -> WeightsEnum:
...
+    Returns a list with the names of registered models.
+
+    Args:
+        module (ModuleType, optional): The module from which we want to extract the available models.
+
+    Returns:
+        models (list): A list with the names of available models.
+    """
+    models = [
+        k for k, v in BUILTIN_MODELS.items() if module is None or v.__module__.rsplit(".", 1)[0] == module.__name__
```
NicolasHug, line 175: Is this the same?

```suggestion
        k for k, v in BUILTIN_MODELS.items() if module is None or v.__module__ is module
```

torchvision/models/_api.py
```diff
@@ -140,3 +143,54 @@ def _get_enum_from_fn(fn: Callable) -> WeightsEnum:
...
 
     return cast(WeightsEnum, weights_enum)
+
+
+M = TypeVar("M", bound=Type[nn.Module])
+
+BUILTIN_MODELS = {}
+
+
+def register_model(name: Optional[str] = None, overwrite: bool = False) -> Callable[[Callable[..., M]], Callable[..., M]]:
```
NicolasHug, line 153: Q: what do we need the `overwrite` parameter for? IIUC the registration decorator is private, so this must come  from an internal need?

torchvision/models/__init__.py
```diff
@@ -14,4 +14,4 @@
 from .vision_transformer import *
 from .swin_transformer import *
 from . import detection, optical_flow, quantization, segmentation, video
-from ._api import get_weight
+from ._api import get_weight, list_models, load
```
NicolasHug, line 17: We might want to call this `load_model` (and rename `datasets.load` into `load_dataset` as well) to avoid conflicts?

NicolasHug (commented):
Thanks @datumbox , just question from me for now

Some wasys we wanted to use the registration mechanism were:

- define which weight correspond to `pretrained=True` so we can document this automatically
- get the available weights from a model name / builder

Do we intend to plan for this now, or later?

torchvision/models/_api.py
datumbox: No it's not. `__module__` is string while `module` is an actual module. We do the operations on string level to avoid extra method calls on `inspect.getmodule(v)` or `sys.modules[v.__module__]`. If you have lightweight alternatives happy to adopt.

torchvision/models/_api.py
datumbox: ClassyVision's registration mechanism had it as an option, so I thought to do the same here. You are right to say that internally, we can fully control this and won't need the overwrite feature. I suspect that eventually we might need it for fb internal bits but if you prefer it we can drop it for now.

torchvision/models/__init__.py
datumbox: Yeah sounds good to me. I'm not going to rename load for datasets on this PR (cause it's going to add too many changes) but I'll rename `model_load()`

datumbox:
@NicolasHug Thanks for the feedback.

> define which weight correspond to pretrained=True so we can document this automatically

I don't think this is something that the registration mechanism should handle because the `pretrained=True` will go away in a few versions. I think this is still useful but probably needs to be implemented on a different annotation than `register_model`.

> get the available weights from a model name / builder

Yes we can do that. We already have a private method that can achieve exactly this called `_get_enum_from_fn()` but we can make one public using the string names.

torchvision/models/_api.py
NicolasHug: My only concern is that this seems to be based on the assumption that the model builders are always defined in `THE_MODULE/some_file.py` (e.g. `models/resnet.py`) which may or may not always be the case in the future. But maybe this isn't too important

NicolasHug (approved):
> I don't think this is something that the registration mechanism should handle because the pretrained=True will go away in a few versions


In my mind, it doesn't really matter when `pretrained=True` will be fully removed. This kind of info will be relevant and useful even when it's completely removed. For example for people  looking at code (e.g. from past papers implementations) that still use this syntax, they would need to know which weight it corresponds to.

We can do it in another decorator / registration mechanism, but it looks like it might have some overlap with this one.

about the `"quantized_xyz"` name: it's a bit of a shame that the name doesn't match the model builder's name, but as discussed earlier I don't think there's a much better way that what you did here.

torchvision/models/_api.py
datumbox: True, doing string manipulation is not great. Maybe doing `sys.modules[v.__module__] is module` is lightweight enough. Let me give it a try.

torchvision/models/_api.py
```diff
@@ -107,10 +110,26 @@ def get_weight(name: str) -> WeightsEnum:
     return weights_enum.from_str(value_name)
 
 
-def _get_enum_from_fn(fn: Callable) -> WeightsEnum:
+W = TypeVar("W", bound=Type[WeightsEnum])
+
+
+def get_model_weight(name: str) -> W:
```
NicolasHug, line 116: Maybe this could support a callable as well?

torchvision/models/_api.py
datumbox: @NicolasHug here is what we can do:
```
k for k, v in BUILTIN_MODELS.items() if module is None or sys.modules[v.__module__].__package__ == module.__package__
```

Note that the comparison again drops on string level. Also I don't think this is actually avoid the problem that you highlighted. Perhaps we can get around it if we start calling `inspect.getmodule()` but I think that's going to be an overkill. :(

NicolasHug: Ah. I guess the question is: if we write a new model and it doesn't get registered properly, will we 100% catch it?

If the answer is yes we can sleep easy.

Regardless, I'd say let's just keep the implementation the way it is. It's still a rather unlikely that this won't work.

torchvision/models/_api.py
datumbox: I can make `_get_enum_from_fn` public or I can make this one accept both names and fns. I think having separate methods is cleaner. 

Unfortunately if I support both, we might have issues with the naming of the methods. I could perhaps name them like `get_model_weight_by_name()` and `get_model_weight_by_fn()`, but I'm not too happy with them. Any thoughts?

NicolasHug: Personally I'd go with a single function that accepts strings and callables (similarly to the `weights` parameters which accept strings and `Weight` objects)

datumbox: Done can you have one more look overall?

torchvision/models/_api.py
```diff
@@ -107,10 +110,27 @@ def get_weight(name: str) -> WeightsEnum:
     return weights_enum.from_str(value_name)
 
 
+W = TypeVar("W", bound=Type[WeightsEnum])
+
+
+def get_model_weight(model: Union[Callable, str]) -> W:
```
nairbv, line 116: It feels a bit unintuitive to me that this returns metadata about the weights instead of the actual model parameters. it's probably too late to change the enums, but I wonder if we should have named them WeightsMetadata, or if this be any clearer as something like `get_model_weight_enum`?

datumbox: Thanks @nairbv for the feedback.

I agree that the names can be confusing but thankfully these are just the base classes. The concrete user-facing enums look like `ResNet50_Weights`, so hopefully this doesn't lead to confusion. I can make the change to enum if you feel it's going to be clearer, just note that the word `enum` is not actually visible to the users. WDYT?

jdsgomes: +1 to the suggestion of renaming to `get_model_weights`

torchvision/models/_api.py
```diff
@@ -140,3 +160,59 @@ def _get_enum_from_fn(fn: Callable) -> WeightsEnum:
         )
 
     return cast(WeightsEnum, weights_enum)
+
+
+M = TypeVar("M", bound=Type[nn.Module])
+
+BUILTIN_MODELS = {}
```
jdsgomes, line 167: Nit: someone could use this mechanism to register their own models. Should we rename this to available_models or something similar?

torchvision/models/_api.py
datumbox: I had an offline chat with @nairbv and his main concern is that the current naming doesn't make it clear whether a single Weight entry is returned versus the whole class. I think thus renaming this to `get_model_weights_enum()` makes sense. I'll update the PR accordingly.

datumbox: @jdsgomes I missed your reply. Any thoughts on Brian's proposal? cc @NicolasHug

torchvision/models/_api.py
datumbox: Happy to make the change to `AVAILABLE_MODELS`. The reason why I used this naming is to align with the ones we use on Datasets:
https://github.com/pytorch/vision/blob/6ca9c76adb6daf2695d603ad623a9cf1c4f4806f/torchvision/prototype/datasets/_api.py#L12

https://github.com/pytorch/vision/blob/6ca9c76adb6daf2695d603ad623a9cf1c4f4806f/torchvision/prototype/datasets/_api.py#L23

Do you think we should make the change now or postpone for when/if we decide to open-up the registration mechanism?

torchvision/models/_api.py
NicolasHug: No strong opinion. The minor issue with `get_model_weights_enum` is that it exposes an implementation detail (the `enum` part).

Like @jdsgomes , I think `get_model_weights` makes more sense than `get_model_weight` since all of these objects end with `Weights` ("s" at the end).

datumbox: Thanks everyone for your input. I think I'll move forward with `get_model_weights()` for now. I'm not going to merge the PR but instead leave this be the live RFC doc. If there are strong concerns over the naming, we can address them on the actual PR that lands the code (coming up). If you have more concerns, please keep commenting on this doc.

torchvision/models/_api.py
jdsgomes: In that case we can keep it consistent and make the changes later in both should we decide to

datumbox:
Thanks everyone for their feedback and input. I will close this PR and send a new one with all the proposed changes. This way we will keep this live RFC doc easy to read.

PR #6331, opened 2022-07-28
[proto] Fixed Normalize._transform and added mid-level normalize

vfdev-5:
Description:
- Fixed Normalize._transform
- Changed return type hint for _Feature
- Added a test

### DIFF
--- a/torchvision/prototype/features/_feature.py
+++ b/torchvision/prototype/features/_feature.py
@@ -84,10 +84,10 @@ class _Feature(torch.Tensor):
         else:
             return output
 
-    def horizontal_flip(self) -> Any:
+    def horizontal_flip(self) -> F:
         return self
 
-    def vertical_flip(self) -> Any:
+    def vertical_flip(self) -> F:
         return self
 
     # TODO: We have to ignore override mypy error as there is torch.Tensor built-in deprecated op: Tensor.resize
@@ -98,13 +98,13 @@ class _Feature(torch.Tensor):
         interpolation: InterpolationMode = InterpolationMode.BILINEAR,
         max_size: Optional[int] = None,
         antialias: bool = False,
-    ) -> Any:
+    ) -> F:
         return self
 
-    def crop(self, top: int, left: int, height: int, width: int) -> Any:
+    def crop(self, top: int, left: int, height: int, width: int) -> F:
         return self
 
-    def center_crop(self, output_size: List[int]) -> Any:
+    def center_crop(self, output_size: List[int]) -> F:
         return self
 
     def resized_crop(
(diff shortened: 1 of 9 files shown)


torchvision/prototype/features/_image.py
```diff
@@ -327,3 +327,9 @@ def gaussian_blur(self, kernel_size: List[int], sigma: Optional[List[float]] = N
 
         output = _F.gaussian_blur_image_tensor(self, kernel_size=kernel_size, sigma=sigma)
         return Image.new_like(self, output)
+
+    def normalize(self, mean: List[float], std: List[float], inplace: bool = False) -> Image:
```
datumbox, line 331: I wouldn't put the normalize in the Image object. This is because the moment a picture is normalized, we lose information about its max value from within the object which is necessary for many transforms.

datumbox (commented):
Most of the changes look fine, just one comment below.

torchvision/prototype/features/_feature.py
```diff
@@ -84,10 +86,10 @@ def __torch_function__(
         else:
             return output
 
-    def horizontal_flip(self) -> Any:
+    def horizontal_flip(self) -> _Feature:
```
datumbox, line 89: Nit: How about returning `T` where `T = TypeVar("T", bound=_Feature)`?

torchvision/prototype/features/_feature.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
datumbox, line 1: What's the minimum Python version that supports this?

torchvision/prototype/transforms/_augment.py
```diff
@@ -86,7 +86,7 @@ def _get_params(self, sample: Any) -> Dict[str, Any]:
         return dict(i=i, j=j, h=h, w=w, v=v)
 
     def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
-        if isinstance(inpt, (features.Image, torch.Tensor)):
+        if is_simple_tensor(inpt) or isinstance(inpt, features.Image):
```
datumbox, line 89: FYI: Everywhere you currently do isinstance(..., Image) will need to change to support Videos. We don't have to act upon this now, but something to keep in mind.

datumbox (approved):
LGTM, thanks @vfdev-5. I've noted some optional nits and fyis, but nothing blocking if you want to tackle them on a follow up.

torchvision/prototype/features/_feature.py
vfdev-5: Probably, 3.7 (https://docs.python.org/3/whatsnew/3.7.html#pep-563-postponed-evaluation-of-annotations)
By the way, we are already doing that in `_image.py`, `_label.py` and others

torchvision/prototype/features/_feature.py
vfdev-5: I tried that https://github.com/pytorch/vision/pull/6331/commits/41a3f30fb24c264aa81de2d03e056c3ab79be2c2 with already defined `F` but mypy was failing...

PR #6333, opened 2022-07-29
Add registration mechanism for models

### LINKED ISSUE #1143: Is there a plan to add a registration mechanism to torchvision.models?

datumbox:
Fixes #1143

Implements the proposal from RFC #6330

### DIFF
--- a/test/test_extended_models.py
+++ b/test/test_extended_models.py
@@ -1,4 +1,3 @@
-import importlib
 import os
 
 import pytest
--- a/test/test_extended_models.py
+++ b/test/test_extended_models.py
@@ -52,15 +52,14 @@ def test_list_models(module):
         return [
             v.__name__
             for k, v in module.__dict__.items()
-            if callable(v) and k[0].lower() == k[0] and k[0] != "_" and k not in models._api.__all__
+            if callable(v) and k[0].islower() and k[0] != "_" and k not in models._api.__all__
         ]
 
     a = set(get_models_from_module(module))
     b = set(x.replace("quantized_", "") for x in models.list_models(module))
 
     assert len(b) > 0
-    assert len(a - b) == 0
-    assert len(b - a) == 0
+    assert a == b
 
 
 @pytest.mark.parametrize(
--- a/test/test_extended_models.py
+++ b/test/test_extended_models.py
@@ -5,7 +5,7 @@ import pytest
 import test_models as TM
 import torch
 from torchvision import models
-from torchvision.models._api import Weights, WeightsEnum
+from torchvision.models._api import get_model_weights, Weights, WeightsEnum
 from torchvision.models._utils import handle_legacy_interface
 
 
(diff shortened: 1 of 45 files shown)


test/test_extended_models.py
```diff
@@ -15,23 +14,53 @@
...
+def test_list_models(module):
+    def get_models_from_module(module):
+        return [
+            v.__name__
             for k, v in module.__dict__.items()
-            if k.endswith(weights_name) and k.replace(weights_name, "").lower() == model_fn.__name__
-        )
-    except StopIteration:
-        return None
+            if callable(v) and k[0].lower() == k[0] and k[0] != "_" and k not in models._api.__all__
```
NicolasHug, line 55:
```suggestion
            if callable(v) and k[0].islower() and k[0] != "_" and k not in models._api.__all__
```

test/test_extended_models.py
```diff
@@ -15,23 +14,53 @@
...
+        ]
+
+    a = set(get_models_from_module(module))
+    b = set(x.replace("quantized_", "") for x in models.list_models(module))
+
+    assert len(b) > 0
+    assert len(a - b) == 0
+    assert len(b - a) == 0
```
NicolasHug, line 63: is this the same as `assert a == b` ?

torchvision/models/_api.py
```diff
@@ -140,3 +164,63 @@ def _get_enum_from_fn(fn: Callable) -> WeightsEnum:
...
+    Returns:
+        models (list): A list with the names of available models.
+    """
+    models = [
+        k for k, v in BUILTIN_MODELS.items() if module is None or v.__module__.rsplit(".", 1)[0] == module.__name__
+    ]
+    return sorted(models)
+
+
+def find_model(name: str) -> Callable[..., M]:
```
NicolasHug, line 203: Reading the PR above, it's not obvious what the difference is between `find_model` and `get_model`. Maybe rename this to `get_model_fn` or `get_model_builder`?

Also might be worth prefixing with `_`, just to make clear that this is purely internal when reading the rest of the code.

torchvision/models/_api.py
```diff
@@ -75,7 +78,9 @@ def __getattr__(self, name):
 
 def get_weight(name: str) -> WeightsEnum:
     """
-    Gets the weight enum value by its full name. Example: "ResNet50_Weights.IMAGENET1K_V1"
+    Gets the weights enum value by its full name. Example: "ResNet50_Weights.IMAGENET1K_V1"
```
NicolasHug, line 81: Do we really want this and the rest to be Beta? I feel like we should reserve that for exceptional cases only. What is the plan to graduate these to stable?

torchvision/models/_api.py
```diff
@@ -75,7 +78,9 @@ def __getattr__(self, name):
...
+
+    .. betastatus:: method
```
NicolasHug, line 83: Technically this is a function

torchvision/models/_api.py
datumbox: The reasons it's named like this is to align with the naming on Datasets side. See here:
https://github.com/pytorch/vision/blob/6ca9c76adb6daf2695d603ad623a9cf1c4f4806f/torchvision/prototype/datasets/_api.py#L38-L41

But at any case the method is already internal as it's not exposed via the `__all__` and it's in a private module.

torchvision/models/_api.py
datumbox: I would rather release this as beta and depending on community feedback graduate before release than releasing to stable and then having to communicate it as beta on the last moment.

NicolasHug: Do we have a way to remember to graduate this before the release?

I'm a bit concerned that we're going to start merging everything as Beta as soon as we're not 100% sure of it. Maybe it's a sign we need to think more about the PR?  Eventually, this might lead to the Beta status having no meaningful definition.

NicolasHug: IMO we should use beta for things where we're clearly not sure, or where we foresee potentially mixed community feedback. Is this the case here?

Also, we should consider that in the future, the guideline might be that Beta APIs  must come with a warning (which can potentially be disabled): https://github.com/pytorch/pytorch/issues/72317, https://github.com/pytorch/vision/pull/6173, https://github.com/pytorch/pytorch/compare/master...albanD:feature_warning_system?expand=1.
I feel like this would be overkill here?

datumbox: > IMO we should use beta for things where we're clearly not sure, or where we foresee potentially mixed community feedback. Is this the case here?

Yes that's totally the case. As you can see from #6330, our team had different opinions on the naming of the methods. We should seek the input of the community.

> I feel like this would be overkill here?

Yes I think adding warnings is too extreme and it's something that needs to be discussed separately.

NicolasHug: If we're only concerned about the naming, then I don't think breaking user code for changing an imperfect name to another one is something reasonable to do.

torchvision/models/_api.py
```diff
@@ -140,3 +164,63 @@ def _get_enum_from_fn(fn: Callable) -> WeightsEnum:
...
+def find_model(name: str) -> Callable[..., M]:
+    name = name.lower()
+    try:
+        fn = BUILTIN_MODELS[name]
+    except KeyError:
+        raise ValueError(f"Unknown model {name}")
+    return fn
+
+
+def get_model(name: str, **config: Any) -> M:
```
jdsgomes, line 212: +1 to Nicolas naming suggestion. My preference would be `build_model` and `get_model_builder`. This can be done in a separate PR and align to datasets or together with any changes that might arise from community feedback. 
Shouldn't have missed it in the RFC.

jdsgomes (approved):

torchvision/models/_api.py
datumbox: Sounds good. Let's release is as beta to get input and we can promote to stable once we get enough positive feedback from the community. Concerning the details of this comment, the `get_model()` returns an initialized model it doesn't return the model builder. It's the `find_model()` that returns the model builder, but that's an internal method on the `_api` module and not exposed publicly. At any case, I agree it's worth reviewing the names on a follow up and ensure they are aligned with the mechanism proposed on the datasets.

torchvision/models/_api.py
datumbox: Other than the naming, there are a few more API considerations that might arise such as the way the `module` parameter of `list_models()` works and Aligning the two registration APIs across Datasets and Models. More issues might pop up after we communicate the mechanism (I'm conscious that we all have blindspots). Overall I would much rather get input from the community on this than later introduce breaking changes in order to act upon their feedback. I think what we can do is try to push this mechanism via social and blogposts and collect feedback prior the release. If all is good, we should be able to promote to stable pretty fast aka before the next release.

NicolasHug:
> there are a few more API considerations that might arise such as the way the module parameter of list_models() works and Aligning the two registration APIs across Datasets and Models

As discussed internally:

- any change we make to the `module` parameter can be BC
- aligning with the dataset API isn't constraining in any way since it's still prototype.

Merging this as Beta without clear potentially BC-breaking changes sets a precedent, and leaves the door open for marking any future APIs as Beta just for the sake of circumventing our deprecation policy. I'm sorry but I cannot support this change.

torchvision/models/_api.py
jdsgomes: I meant `get_model` -> `build_model` and `find_model` -> `get_model_builder`.

PR #6335, opened 2022-07-29
Simplify _check_integrity for cifar and stl10

kit1980:


oke-aditya (approved):
LGTM! Logically change is same! cc @pmeier 

Although I wonder in general we should be using better checksum algorithms like SHA-256? 
md5 isn't cryptographically secure or best.

NicolasHug (approved):
thanks @kit1980

PR #6338, opened 2022-07-31
Fix d/c IoU for different batch sizes

### LINKED ISSUE #6317: distance_box_iou() and complete_box_iou() don't work if both sets don't have the same number of boxes
### 🐛 Describe the bug

`*box_iou()` functions should return a matrix of results for every possible pair (box1, box2), where box1 is a box from `boxes1` and box2 is a box from `boxes2`. `box_iou()` and `generalized_box_iou()` work this way, i.e. if `boxes1` is an Nx4 matrix and `boxes2` is an Mx4 matrix, the result is an NxM matrix. The recently added `distance_box_iou()` and `complete_box_iou()` don't work if there aren't as many boxes in `boxes1` and `boxes2`.

```python
import torch
from torchvision.ops import box_iou, generalized_box_iou, distance_box_iou, complete_box_iou

N = 5
M = 6
boxes1 = torch.rand((N, 4))
boxes2 = torch.rand((M, 4))
print(box_iou(boxes1, boxes2).shape)  # torch.Size([5, 6])
print(generalized_box_iou(boxes1, boxes2).shape)  # torch.Size([5, 6])
print(distance_box_iou(boxes1, boxes2).shape)  # RuntimeError
print(complete_box_iou(boxes1, boxes2).shape)  # RuntimeError
```

When running the above code, `distance_box_iou()` and `complete_box_iou()` will fail with a `RuntimeError`. The output is below:

```
torch.Size([5, 6])
torch.Size([5, 6])
  File ".../test.py", line 10, in <module>
    print(distance_box_iou(boxes1, boxes2).shape)  # RuntimeError
  File ".../lib/python3.9/site-packages/torchvision/ops/boxes.py", line 361, in distance_box_iou
    diou, _ = _box_diou_iou(boxes1, boxes2)
  File ".../lib/python3.9/site-packages/torchvision/ops/boxes.py", line 378, in _box_diou_iou
    centers_distance_squared = (_upcast(x_p - x_g) ** 2) + (_upcast(y_p - y_g) ** 2)
RuntimeError: The size of tensor a (5) must match the size of tensor b (6) at non-singleton dimension 0
```

This is not caught by the unit tests, because there's no such case where there's a different number of boxes in the two sets.

The problem is in [`_box_diou_iou()`](https://github.com/pytorch/vision/blob/main/torchvision/ops/boxes.py#L378). It looks like `iou` and `diagonal_distance_squared` are calculated for every possible pair (by adding an empty dimension), but `centers_distance_squared` is not.

_As a side note, I personally feel it's confusing that these functions produce the output for every possible pair. By convention, PyTorch functions produce element-wise results. For example, `torch.add(boxes1, boxes2)` only works if `boxes1` and `boxes2` contain the same number of boxes. If you want a pair-wise addition, you can easily call `torch.add(boxes1[:, None, :], boxes2)`. The fact that `*box_iou()` functions produce pair-wise results makes the implementation complicated. And the only way to get element-wise results is calling `box_iou(boxes1, boxes2).diagonal()`, which is inefficient._

### Versions

PyTorch version: 1.12.0+cu113
Is debug build: False
CUDA used to build PyTorch: 11.3

OS: Ubuntu 20.04.4 LTS (x86_64)
GCC version: (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0
CMake version: version 3.16.3
Libc version: glibc-2.31

Python version: 3.9.12 (main, Apr  5 2022, 06:56:58)  [GCC 7.5.0] (64-bit runtime)
Python platform: Linux-5.10.102.1-microsoft-standard-WSL2-x86_64-with-glibc2.31
Is CUDA available: True
GPU models and configuration: GPU 0: NVIDIA GeForce GTX 1650 Ti with Max-Q Design
Nvidia driver version: 516.59
cuDNN version: Could not collect
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

[pip3] mypy==0.950
[pip3] mypy-extensions==0.4.3
[pip3] numpy==1.22.3
[pip3] pytorch-lightning==1.6.5
[pip3] pytorch-lightning-bolts==0.2.5
[pip3] pytorch-quantization==2.1.2
[pip3] torch==1.12.0+cu113
[pip3] torchmetrics==0.6.0
[pip3] torchtext==0.12.0
[pip3] torchvision==0.13.0+cu113
[conda] Could not collect

oke-aditya:
Closes #6317 

I hope I have done it correct.

Also I added tests to check both the cases M>N and N>M, this needed a small tests refactor. Let's be on safe side :smile: 

cc @datumbox @yassineAlouini

### DIFF
--- a/torchvision/ops/boxes.py
+++ b/torchvision/ops/boxes.py
@@ -331,7 +331,9 @@ def complete_box_iou(boxes1: Tensor, boxes2: Tensor, eps: float = 1e-7) -> Tenso
     w_gt = boxes2[:, 2] - boxes2[:, 0]
     h_gt = boxes2[:, 3] - boxes2[:, 1]
 
-    v = (4 / (torch.pi**2)) * torch.pow((torch.atan(w_gt / h_gt) - torch.atan(w_pred / h_pred)), 2)
+    aspect_gt = torch.atan(w_gt / h_gt)
+    aspect_pred = torch.atan(w_pred / h_pred)
+    v = (4 / (torch.pi**2)) * torch.pow((aspect_gt - aspect_pred[:, None]), 2)
     with torch.no_grad():
         alpha = v / (1 - iou + v + eps)
     return diou - alpha * v
@@ -364,6 +366,9 @@ def distance_box_iou(boxes1: Tensor, boxes2: Tensor, eps: float = 1e-7) -> Tenso
 
 def _box_diou_iou(boxes1: Tensor, boxes2: Tensor, eps: float = 1e-7) -> Tuple[Tensor, Tensor]:
 
+    # if boxes1.size(0) > boxes2.size(0):
+    #     boxes1, boxes2 = boxes2, boxes1
+
     iou = box_iou(boxes1, boxes2)
     lti = torch.min(boxes1[:, None, :2], boxes2[:, :2])
     rbi = torch.max(boxes1[:, None, 2:], boxes2[:, 2:])
@@ -375,10 +380,14 @@ def _box_diou_iou(boxes1: Tensor, boxes2: Tensor, eps: float = 1e-7) -> Tuple[Te
     x_g = (boxes2[:, 0] + boxes2[:, 2]) / 2
     y_g = (boxes2[:, 1] + boxes2[:, 3]) / 2
     # The distance between boxes' centers squared.
-    centers_distance_squared = (_upcast(x_p - x_g) ** 2) + (_upcast(y_p - y_g) ** 2)
+    centers_distance_squared = (_upcast((x_p - x_g[:, None]).diag()) ** 2) + (_upcast((y_p - y_g[:, None]).diag()) ** 2)
     # The distance IoU is the IoU penalized by a normalized
     # distance between boxes' centers squared.
-    return iou - (centers_distance_squared / diagonal_distance_squared), iou
+    if boxes1.size(0) > boxes2.size(0):
+        center_distance_ratio = centers_distance_squared[None, :] / diagonal_distance_squared
+    else:
+        center_distance_ratio = centers_distance_squared[:, None] / diagonal_distance_squared
+    return iou - center_distance_ratio, iou
 
 
 def masks_to_boxes(masks: torch.Tensor) -> torch.Tensor:
--- a/torchvision/ops/boxes.py
+++ b/torchvision/ops/boxes.py
@@ -328,10 +328,10 @@ def complete_box_iou(boxes1: Tensor, boxes2: Tensor, eps: float = 1e-7) -> Tenso
     w_pred = boxes1[:, None, 2] - boxes1[:, None, 0]
     h_pred = boxes1[:, None, 3] - boxes1[:, None, 1]
 
-    w_gt = boxes2[:, None, 2] - boxes2[:, None, 0]
-    h_gt = boxes2[:, None, 3] - boxes2[:, None, 1]
+    w_gt = boxes2[:, 2] - boxes2[:, 0]
+    h_gt = boxes2[:, 3] - boxes2[:, 1]
 
-    v = (4 / (torch.pi**2)) * torch.pow(torch.atan(w_pred / h_pred) - torch.atan(w_gt / h_gt).t(), 2)
+    v = (4 / (torch.pi**2)) * torch.pow(torch.atan(w_pred / h_pred) - torch.atan(w_gt / h_gt), 2)
     with torch.no_grad():
         alpha = v / (1 - iou + v + eps)
     return diou - alpha * v
--- a/torchvision/ops/boxes.py
+++ b/torchvision/ops/boxes.py
@@ -366,9 +366,6 @@ def distance_box_iou(boxes1: Tensor, boxes2: Tensor, eps: float = 1e-7) -> Tenso
 
 def _box_diou_iou(boxes1: Tensor, boxes2: Tensor, eps: float = 1e-7) -> Tuple[Tensor, Tensor]:
 
-    # if boxes1.size(0) > boxes2.size(0):
-    #     boxes1, boxes2 = boxes2, boxes1
-
     iou = box_iou(boxes1, boxes2)
     lti = torch.min(boxes1[:, None, :2], boxes2[:, :2])
     rbi = torch.max(boxes1[:, None, 2:], boxes2[:, 2:])
--- a/torchvision/ops/boxes.py
+++ b/torchvision/ops/boxes.py
@@ -360,7 +360,7 @@ def distance_box_iou(boxes1: Tensor, boxes2: Tensor, eps: float = 1e-7) -> Tenso
 
     boxes1 = _upcast(boxes1)
     boxes2 = _upcast(boxes2)
-    diou, _ = _box_diou_iou(boxes1, boxes2)
+    diou, _ = _box_diou_iou(boxes1, boxes2, eps)
     return diou
 
 
--- a/torchvision/ops/boxes.py
+++ b/torchvision/ops/boxes.py
@@ -325,15 +325,13 @@ def complete_box_iou(boxes1: Tensor, boxes2: Tensor, eps: float = 1e-7) -> Tenso
 
     diou, iou = _box_diou_iou(boxes1, boxes2, eps)
 
-    w_pred = boxes1[:, 2] - boxes1[:, 0]
-    h_pred = boxes1[:, 3] - boxes1[:, 1]
+    w_pred = boxes1[:, None, 2] - boxes1[:, None, 0]
+    h_pred = boxes1[:, None, 3] - boxes1[:, None, 1]
 
-    w_gt = boxes2[:, 2] - boxes2[:, 0]
-    h_gt = boxes2[:, 3] - boxes2[:, 1]
+    w_gt = boxes2[:, None, 2] - boxes2[:, None, 0]
+    h_gt = boxes2[:, None, 3] - boxes2[:, None, 1]
 
-    aspect_gt = torch.atan(w_gt / h_gt)
-    aspect_pred = torch.atan(w_pred / h_pred)
-    v = (4 / (torch.pi**2)) * torch.pow((aspect_gt - aspect_pred[:, None]), 2)
+    v = (4 / (torch.pi**2)) * torch.pow(torch.atan(w_pred / h_pred) - torch.atan(w_gt / h_gt).t(), 2)
     with torch.no_grad():
         alpha = v / (1 - iou + v + eps)
     return diou - alpha * v
@@ -360,7 +358,7 @@ def distance_box_iou(boxes1: Tensor, boxes2: Tensor, eps: float = 1e-7) -> Tenso
 
     boxes1 = _upcast(boxes1)
     boxes2 = _upcast(boxes2)
-    diou, _ = _box_diou_iou(boxes1, boxes2, eps)
+    diou, _ = _box_diou_iou(boxes1, boxes2, eps=eps)
     return diou
 
 
@@ -377,14 +375,12 @@ def _box_diou_iou(boxes1: Tensor, boxes2: Tensor, eps: float = 1e-7) -> Tuple[Te
     x_g = (boxes2[:, 0] + boxes2[:, 2]) / 2
     y_g = (boxes2[:, 1] + boxes2[:, 3]) / 2
     # The distance between boxes' centers squared.
-    centers_distance_squared = (_upcast((x_p - x_g[:, None]).diag()) ** 2) + (_upcast((y_p - y_g[:, None]).diag()) ** 2)
+    centers_distance_squared = (_upcast((x_p[:, None] - x_g[None, :])) ** 2) + (
+        _upcast((y_p[:, None] - y_g[None, :])) ** 2
+    )
     # The distance IoU is the IoU penalized by a normalized
     # distance between boxes' centers squared.
-    if boxes1.size(0) > boxes2.size(0):
-        center_distance_ratio = centers_distance_squared[None, :] / diagonal_distance_squared
-    else:
-        center_distance_ratio = centers_distance_squared[:, None] / diagonal_distance_squared
-    return iou - center_distance_ratio, iou
+    return iou - (centers_distance_squared / diagonal_distance_squared), iou
 
 
 def masks_to_boxes(masks: torch.Tensor) -> torch.Tensor:
--- a/torchvision/ops/ciou_loss.py
+++ b/torchvision/ops/ciou_loss.py
@@ -14,7 +14,7 @@ def complete_box_iou_loss(
 
     """
     Gradient-friendly IoU loss with an additional penalty that is non-zero when the
-    boxes do not overlap overlap area, This loss function considers important geometrical
+    boxes do not overlap area, This loss function considers important geometrical
     factors such as  overlap area, normalized central point distance and aspect ratio.
     This loss is symmetric, so the boxes1 and boxes2 arguments are interchangeable.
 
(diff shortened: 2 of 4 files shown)


torchvision/ops/boxes.py
```diff
@@ -375,10 +380,14 @@ def _box_diou_iou(boxes1: Tensor, boxes2: Tensor, eps: float = 1e-7) -> Tuple[Te
     x_g = (boxes2[:, 0] + boxes2[:, 2]) / 2
     y_g = (boxes2[:, 1] + boxes2[:, 3]) / 2
     # The distance between boxes' centers squared.
-    centers_distance_squared = (_upcast(x_p - x_g) ** 2) + (_upcast(y_p - y_g) ** 2)
+    centers_distance_squared = (_upcast((x_p - x_g[:, None]).diag()) ** 2) + (_upcast((y_p - y_g[:, None]).diag()) ** 2)
     # The distance IoU is the IoU penalized by a normalized
     # distance between boxes' centers squared.
-    return iou - (centers_distance_squared / diagonal_distance_squared), iou
+    if boxes1.size(0) > boxes2.size(0):
```
oke-aditya, line 386: I could not find a better way of handling it. Let me know any other solution.

torchvision/ops/boxes.py
```diff
@@ -375,10 +380,14 @@ def _box_diou_iou(boxes1: Tensor, boxes2: Tensor, eps: float = 1e-7) -> Tuple[Te
     x_g = (boxes2[:, 0] + boxes2[:, 2]) / 2
     y_g = (boxes2[:, 1] + boxes2[:, 3]) / 2
     # The distance between boxes' centers squared.
-    centers_distance_squared = (_upcast(x_p - x_g) ** 2) + (_upcast(y_p - y_g) ** 2)
+    centers_distance_squared = (_upcast((x_p - x_g[:, None]).diag()) ** 2) + (_upcast((y_p - y_g[:, None]).diag()) ** 2)
```
oke-aditya, line 383: `.diag()` is needed as per new semantics of broadcast 

https://pytorch.org/docs/stable/notes/broadcasting.html#backwards-compatibility

```
would previously produce a Tensor with size: torch.Size([4,1]), but now produces a Tensor with size: torch.Size([4,4]). In order to help identify cases in your code where backwards incompatibilities introduced by broadcasting may exist, you may set torch.utils.backcompat.broadcast_warning.enabled to True, which will generate a python warning in such cases.
```

oke-aditya (commented):
@datumbox Plz verify these changes, I think I have done it correctly.

torchvision/ops/boxes.py
```diff
@@ -358,7 +360,7 @@ def distance_box_iou(boxes1: Tensor, boxes2: Tensor, eps: float = 1e-7) -> Tenso
 
     boxes1 = _upcast(boxes1)
     boxes2 = _upcast(boxes2)
-    diou, _ = _box_diou_iou(boxes1, boxes2)
+    diou, _ = _box_diou_iou(boxes1, boxes2, eps)
```
oke-aditya, line 363: We weren't using the epsilon? I think this too went unnoticed :sweat:

torchvision/ops/ciou_loss.py
```diff
@@ -14,7 +14,7 @@ def complete_box_iou_loss(
 
     """
     Gradient-friendly IoU loss with an additional penalty that is non-zero when the
-    boxes do not overlap overlap area, This loss function considers important geometrical
+    boxes do not overlap area, This loss function considers important geometrical
     factors such as  overlap area, normalized central point distance and aspect ratio.
```
abhi-glitchhg, line 18:
```suggestion
    boxes do not overlap. This loss function considers important geometrical
    factors such as overlap area, normalized central point distance and aspect ratio.
```

torchvision/ops/boxes.py
datumbox: I am going to suggest an alternative overall approach. Here is the patch:
```patch
diff --git a/torchvision/ops/boxes.py b/torchvision/ops/boxes.py
index e42e7e04a..fd2627f12 100644
--- a/torchvision/ops/boxes.py
+++ b/torchvision/ops/boxes.py
@@ -325,13 +325,13 @@ def complete_box_iou(boxes1: Tensor, boxes2: Tensor, eps: float = 1e-7) -> Tenso
 
     diou, iou = _box_diou_iou(boxes1, boxes2, eps)
 
-    w_pred = boxes1[:, 2] - boxes1[:, 0]
-    h_pred = boxes1[:, 3] - boxes1[:, 1]
+    w_pred = boxes1[:, None, 2] - boxes1[:, None, 0]
+    h_pred = boxes1[:, None, 3] - boxes1[:, None, 1]
 
-    w_gt = boxes2[:, 2] - boxes2[:, 0]
-    h_gt = boxes2[:, 3] - boxes2[:, 1]
+    w_gt = boxes2[:, None, 2] - boxes2[:, None, 0]
+    h_gt = boxes2[:, None, 3] - boxes2[:, None, 1]
 
-    v = (4 / (torch.pi**2)) * torch.pow((torch.atan(w_gt / h_gt) - torch.atan(w_pred / h_pred)), 2)
+    v = (4 / (torch.pi**2)) * torch.pow(torch.atan(w_pred / h_pred) - torch.atan(w_gt / h_gt).t(), 2)
     with torch.no_grad():
         alpha = v / (1 - iou + v + eps)
     return diou - alpha * v
@@ -358,7 +358,7 @@ def distance_box_iou(boxes1: Tensor, boxes2: Tensor, eps: float = 1e-7) -> Tenso
 
     boxes1 = _upcast(boxes1)
     boxes2 = _upcast(boxes2)
-    diou, _ = _box_diou_iou(boxes1, boxes2)
+    diou, _ = _box_diou_iou(boxes1, boxes2, eps=eps)
     return diou
 
 
@@ -370,12 +370,12 @@ def _box_diou_iou(boxes1: Tensor, boxes2: Tensor, eps: float = 1e-7) -> Tuple[Te
     whi = _upcast(rbi - lti).clamp(min=0)  # [N,M,2]
     diagonal_distance_squared = (whi[:, :, 0] ** 2) + (whi[:, :, 1] ** 2) + eps
     # centers of boxes
-    x_p = (boxes1[:, 0] + boxes1[:, 2]) / 2
-    y_p = (boxes1[:, 1] + boxes1[:, 3]) / 2
-    x_g = (boxes2[:, 0] + boxes2[:, 2]) / 2
-    y_g = (boxes2[:, 1] + boxes2[:, 3]) / 2
+    x_p = (boxes1[:, None, 0] + boxes1[:, None, 2]) / 2
+    y_p = (boxes1[:, None, 1] + boxes1[:, None, 3]) / 2
+    x_g = (boxes2[:, None, 0] + boxes2[:, None, 2]) / 2
+    y_g = (boxes2[:, None, 1] + boxes2[:, None, 3]) / 2
     # The distance between boxes' centers squared.
-    centers_distance_squared = (_upcast(x_p - x_g) ** 2) + (_upcast(y_p - y_g) ** 2)
+    centers_distance_squared = (_upcast(x_p - x_g.t()) ** 2) + (_upcast(y_p - y_g.t()) ** 2)
     # The distance IoU is the IoU penalized by a normalized
     # distance between boxes' centers squared.
     return iou - (centers_distance_squared / diagonal_distance_squared), iou

```

I'll commit directly to the PR and ask for the support of @fmassa to review.

torchvision/ops/boxes.py
datumbox: I've proposed a different approach. See above. :)

torchvision/ops/boxes.py
oke-aditya: Is this a bit slower ? As we do a few more operations?

fmassa:
Hi,

I've implemented D-IoU a while back (when I was working on DETR), and it had the same API as `generalized_iou`. Here is the implementation:

```python

# modified from torchvision to also return the union
def box_iou(boxes1, boxes2):
    area1 = box_area(boxes1)
    area2 = box_area(boxes2)

    lt = torch.max(boxes1[:, None, :2], boxes2[:, :2])  # [N,M,2]
    rb = torch.min(boxes1[:, None, 2:], boxes2[:, 2:])  # [N,M,2]

    wh = (rb - lt).clamp(min=0)  # [N,M,2]
    inter = wh[:, :, 0] * wh[:, :, 1]  # [N,M]

    union = area1[:, None] + area2 - inter

    iou = inter / union
    return iou, union


def distance_iou(boxes1, boxes2):
    iou, union = box_iou(boxes1, boxes2)

    lt = torch.min(boxes1[:, None, :2], boxes2[:, :2])
    rb = torch.max(boxes1[:, None, 2:], boxes2[:, 2:])

    wh = (rb - lt).clamp(min=0)  # [N,M,2]
    c2 = wh[..., 0] ** 2 + wh[..., 1] ** 2

    b1 = (boxes1[:, 2:] + boxes1[:, :2]) / 2
    b2 = (boxes2[:, 2:] + boxes2[:, :2]) / 2
    # rho2 = torch.cdist(b1, b2, p=2) ** 2
    rho2 = ((b1[:, None, :] - b2[None, :, :]) ** 2).sum(-1)

    return 1 - iou + rho2 / c2
```

Note that I commented out the `torch.cdist` line in the code and replaced it with the line below because using `cdist` would cause instabilities. I don't exactly remember what type of instabilities was there (and maybe this was fixed with a newer version of PyTorch already).

torchvision/ops/boxes.py
fmassa: I made a comment just below with an implementation that I used in the past for distance IoU, hope it can be of help!

torchvision/ops/boxes.py
```diff
@@ -325,13 +325,13 @@ def complete_box_iou(boxes1: Tensor, boxes2: Tensor, eps: float = 1e-7) -> Tenso
...
-    w_pred = boxes1[:, 2] - boxes1[:, 0]
-    h_pred = boxes1[:, 3] - boxes1[:, 1]
+    w_pred = boxes1[:, None, 2] - boxes1[:, None, 0]
+    h_pred = boxes1[:, None, 3] - boxes1[:, None, 1]
 
     w_gt = boxes2[:, 2] - boxes2[:, 0]
     h_gt = boxes2[:, 3] - boxes2[:, 1]
 
-    v = (4 / (torch.pi**2)) * torch.pow((torch.atan(w_gt / h_gt) - torch.atan(w_pred / h_pred)), 2)
+    v = (4 / (torch.pi**2)) * torch.pow(torch.atan(w_pred / h_pred) - torch.atan(w_gt / h_gt), 2)
```
fmassa, line 334: Was this a bug before?

fmassa (approved):
Changes LGTM!

One thing that I might have missed is that it would be good to have a sanity test that compares that the result of a `[N, 4]` by `[M, 4]` tensor is actually equal to performing multiple calls to `[1, 4]` and `[1, 4]` tensors. This way we ensure that we don't have the matrix transposed in any way

torchvision/ops/boxes.py
fmassa: Nevermind, it's a squared function afterwards, doesn't matter :-)

datumbox:
There seems to be a bug on the original implementation other than what you try to fix. As a result the expected values were wrong. To prove this, take @fmassa's script from https://github.com/pytorch/vision/pull/6338#issuecomment-1201505934 and run the current test cases:
```python
def distance_box_iou(boxes1, boxes2):
    return 1 - distance_iou(boxes1, boxes2)

box1 = torch.tensor([[0, 0, 100, 100], [0, 0, 50, 50], [200, 200, 300, 300], [0, 0, 25, 25]])
box2 = torch.tensor([[0, 0, 100, 100], [0, 0, 50, 50], [200, 200, 300, 300]])
print(distance_box_iou(box1, box2))
```

The output is:
```
tensor([[ 1.0000,  0.1875, -0.4444],
        [ 0.1875,  1.0000, -0.5625],
        [-0.4444, -0.5625,  1.0000],
        [-0.0781,  0.1875, -0.6267]])
```

Which is different from the expected values on main:
```
[[1.0, 0.25, 0.0], [0.25, 1.0, 0.0], [0.0, 0.0, 1.0], [0.0625, 0.25, 0.0]]
```

So the follow up commits fix the problems and update the expected values. Here is a script that I used to the bbox methods operate as expected by comparing them to the equivalent loss methods:
```python
from functools import partial

import torch
from torchvision import ops


def loss_to_iou(a, b, fn):
    # Convert Loss to IoU
    return 1.0 - fn(a, b)


def gen_box(size):
    # Build proper bboxes
    xy1 = torch.rand((size, 2))
    xy2 = xy1 + torch.rand((size, 2))
    return torch.cat([xy1, xy2], axis=-1)


def cartesian_product_slow(boxes1, boxes2, fn):
    # Extemely slow method for testing cartesian product and use of the loss fn
    N = boxes1.size(0)
    M = boxes2.size(0)

    result = torch.zeros((N, M))
    for i in range(N):
        for j in range(M):
            result[i, j] = fn(boxes1[i].unsqueeze(0), boxes2[j].unsqueeze(0))

    return result


def cartesian_product_fast(boxes1, boxes2, fn):
    # Faster method for testing cartesian product and use of the loss fn
    N = boxes1.size(0)
    M = boxes2.size(0)

    bx_l = boxes1.repeat_interleave(M, dim=0)
    bx_r = boxes2.repeat((N, 1))

    result = fn(bx_l, bx_r)

    return result.view(N, M)


N = 5
M = 7

boxes1 = gen_box(N)
boxes2 = gen_box(M)

# Check that the slow and fast are equivalent
a = cartesian_product_fast(boxes1, boxes2, partial(loss_to_iou, fn=ops.distance_box_iou_loss))
b = cartesian_product_slow(boxes1, boxes2, partial(loss_to_iou, fn=ops.distance_box_iou_loss))
assert torch.allclose(a, b)

# Check the build-in Generalized IoU method against the cartesian of the equivalent loss
a = cartesian_product_fast(boxes1, boxes2, partial(loss_to_iou, fn=ops.generalized_box_iou_loss))
b = ops.generalized_box_iou(boxes1, boxes2)
assert torch.allclose(a, b)

# Check the build-in Distance IoU method against the cartesian of the equivalent loss
a = cartesian_product_fast(boxes1, boxes2, partial(loss_to_iou, fn=ops.distance_box_iou_loss))
b = ops.distance_box_iou(boxes1, boxes2)
assert torch.allclose(a, b)

# Check the build-in Complete IoU method against the cartesian of the equivalent loss
a = cartesian_product_fast(boxes1, boxes2, partial(loss_to_iou, fn=ops.complete_box_iou_loss))
b = ops.complete_box_iou(boxes1, boxes2)
assert torch.allclose(a, b)

print("OK")
```

torchvision/ops/boxes.py
datumbox: No that's just a styling/nit thingy. The actual bug was on [this](https://github.com/pytorch/vision/pull/6338/files#diff-c49f3b58b42045e50c2157a195bfb83f34cc0d1842dc92aa0e9aa402499dcd88R378-R380) line. Thanks for providing the reference implementation, it really helped me debug it. 😃

torchvision/ops/boxes.py
```diff
@@ -325,13 +325,13 @@ def complete_box_iou(boxes1: Tensor, boxes2: Tensor, eps: float = 1e-7) -> Tenso
 
     diou, iou = _box_diou_iou(boxes1, boxes2, eps)
 
-    w_pred = boxes1[:, 2] - boxes1[:, 0]
-    h_pred = boxes1[:, 3] - boxes1[:, 1]
+    w_pred = boxes1[:, None, 2] - boxes1[:, None, 0]
```
jdsgomes, line 328: why this is needed?

datumbox: This is necessary for the broadcasting to work properly. See also:
https://github.com/pytorch/vision/blob/ea0be26b88778b1033d4a176be68bcdd008ff934/torchvision/ops/boxes.py#L368-L371

Issue #6323, opened 2022-07-27
[RFC] Batteries Included - Phase 3


datumbox:
### 🚀 The feature

TBD

### Motivation, pitch

TBD

datumbox:
Tagging a few of the regular contributors in case they are interested in specific items: 
@abhi-glitchhg @federicopozzi33 @frgfm @lezwon @oke-aditya @xiaohu2015 @yassineAlouini @zhiqwang

Feel free to propose additional candidates.

oke-aditya:
I will be happy to take losses :) dice loss first.

frgfm:
Like Christmas in july hehe :grin:

I have a few questions though:
- for losses, I already have implemented the poly loss on my end. Do you want us to do the Python implementation or also the C++ / CUDA binding? (I saw that there are some in PyTorch core, so I'm not sure what's the target here)
- about optimizers, I also have implementations for LARS & LAMB, should we open PR directly on core? or do we need to contact them on a dedicated issue beforehand?
- about models, happy to go for the implementation but I don't have the gear to train it on Imagenet with the usual procedure. Do we need to train them as well?

Looking forward to help for the next release :)

federicopozzi33:
I'd like to take the Polynomial LR scheduler!

datumbox:
Thanks for offering to help! We are lucky to have you guys! :)

@frgfm excellent questions. Let me try to provide some more context here:
- Losses are going to be added for now in TorchVision with a Python only implementation. Ideally we should reuse as much as possible Core's existing methods that have C++/CUDA bindings. That's particularly true for the PolyLoss where we want to reuse Core's `cross_entropy` rather than rewriting it with pure tensor op. Unless I missed a development on Core (in which case please correct me and provide reference) neither [Poly](https://github.com/pytorch/pytorch/issues/76732) nor [Dice](https://github.com/pytorch/pytorch/issues/1249) are planned to be added.
- For the 2 optimizers and 1 scheduler, the PR should come directly on Core but we will provide help to maximize the changes of getting this landed. Due to the nature of the PR, it has higher risk of not getting merged but I've already spoke with some Core devs about it and I'm hopeful we can get it in.
- For the models the plan is to use the process of our [new model contribution guideline](https://github.com/pytorch/vision/blob/main/CONTRIBUTING_MODELS.md). The TLDR is that yes we want the PR to contain weights that prove at least a tiny variant of the model works as expected but then we will help you train them by running the e2e training on our internal infra.

[...]

Issue #6227, opened 2022-06-30
Swin in this repo + dynamic resolution


vadimkantorov:
### 📚 The doc issue

Does Swin impl in this repo support arbitrary dynamic execution-time-defined input resolution (same as other backbones)?

Initially Swin was trained to support only one resolution, but then hacks can be done to support arbitrary resolution. Two repos with such hacks:
* https://github.com/SwinTransformer/Swin-Transformer-Object-Detection/blob/master/mmdet/models/backbones/swin_transformer.py
* https://github.com/megvii-research/SOLQ/blob/main/models/swin_transformer.py

Related issues: https://github.com/microsoft/SimMIM/issues/13 https://github.com/microsoft/esvit/issues/17

cc @YosuaMichael

YosuaMichael:
Hi @vadimkantorov , from my understanding the answer is yes. 

The swin_transformer implementation in TorchVision dont restrict the input_resolution. However take note that the accuracy may changes depending on your input resolution (or transform).

For example, the original accuracy of swin_transformer using `resize_size=232` and `crop_size=224` achieve `Acc@1 81.472 Acc@5 95.780` in imagenet1k. I have tried to run the validation with `resize_size=128` and `crop_size=112` on the same dataset and I got: `Acc@1 69.248 Acc@5 88.824`.

vadimkantorov:
well, `resize_size` makes think that it's resizing the input instead and always gives a fixed output resolution

from what I understand, the two impl I mentioned above allow for scaling output resolution depending on input resolution

vadimkantorov:
at the very least, the actual behavior should be clearly explained in the docs...

YosuaMichael:
> well, resize_size makes think that it's resizing the input instead and always gives a fixed output resolution
> 
> from what I understand, the two impl I mentioned above allow for scaling output resolution depending on input resolution
> 

Hi @vadimkantorov , just to clarify, the `resize_size` that I mentioned is the `transform` before we do the classification task. It is not part of the `swin_transformer` parameters (sorry for the confusion!)).

For `output resolution` can you explain more what do you mean by it?
From what I understand now the swin transformer model on TorchVision is a meant for classification model, so it will definitely output fixed size output according to the num_classes parameter.

For the two links that you provide, I think their implementation are focus as a backbone and the model return the output of all the layers.

vadimkantorov:
[...]

PR #6342, opened 2022-08-01
MaxVit model 

TeodorPoncu:
This PR is w.r.t. [Batteries Phase 3](https://github.com/pytorch/vision/issues/6323) proposal to add the [MaxVit](https://arxiv.org/abs/2204.01697) architecture. It is still a work in progress as it has yet to be trained.

One caveat w.r.t. the way we would be exposing this model API to users is that the architecture is bound to the specific input size it was trained one (due to the usage of relative positional encodings)

Running the command: `torchrun --nproc_per_node=1 train.py --test-only --prototype --weights MaxVit_T_Weights.IMAGENET1K_V1 --model maxvit_t -b 1` yields the following results: 

`Test: Acc@1 83.700 Acc@5 96.722`


test/test_models.py
```diff
@@ -594,6 +594,14 @@ def test_vitc_models(model_fn, dev):
     test_classification_model(model_fn, dev)
 
 
+@pytest.mark.parametrize(
+    "model_fn", [models.max_vit_T_224, models.max_vit_S_224, models.max_vit_B_224, models.max_vit_L_224]
+)
+@pytest.mark.parametrize("dev", cpu_and_gpu())
+def test_max_vit(model_fn, dev):
+    test_classification_model(model_fn, dev)
```
datumbox, line 602: We shouldn't do this. If you register the model it with the registration mechanism your test will be automatically picked up and run on the `test_classification_model()`

torchvision/models/maxvit.py
```diff
@@ -0,0 +1,595 @@
...
+    coords_flat = torch.flatten(coords, 1)
+    relative_coords = coords_flat[:, :, None] - coords_flat[:, None, :]
+    relative_coords = relative_coords.permute(1, 2, 0).contiguous()
+    relative_coords[:, :, 0] += height - 1
+    relative_coords[:, :, 1] += width - 1
+    relative_coords[:, :, 0] *= 2 * width - 1
+    return relative_coords.sum(-1)
+
+
+class GeluWrapper(nn.Module):
```
datumbox, line 23: This wrapper isn't necessary. You can pass `inplace=None` to `ConvNormActivation2D`. See:
https://github.com/pytorch/vision/blob/0a919dbb825d3f33eefcaaaafd3a2bbf6ac91fb8/torchvision/models/swin_transformer.py#L294

torchvision/models/maxvit.py
```diff
@@ -0,0 +1,595 @@
...
+        self.positional_bias = nn.parameter.Parameter(
+            torch.zeros(((2 * self.size - 1) * (2 * self.size - 1), self.n_heads), dtype=torch.float32),
+        )
+
+        self.register_buffer("relative_position_index", get_relative_position_index(self.size, self.size))
+
+        # initialize with truncated normal the bias
+        self.positional_bias.data.normal_(mean=0, std=0.02)
+
+    def _get_relative_positional_bias(self) -> torch.Tensor:
```
datumbox, line 119: Nit: This is a private method used once. Do we really need it?

torchvision/models/maxvit.py
```diff
@@ -0,0 +1,595 @@
...
+        # split P * P dimension into 2 P tile dimensionsa
+        x = x.reshape(B, HP, WP, P, P, C)
+        # permute into B, C, HP, P, WP, P
+        x = x.permute(0, 5, 1, 3, 2, 4)
+        # reshape into B, C, H, W
+        x = x.reshape(B, C, HP * P, WP * P)
+        return x
+
+
+class MLP(nn.Module):
```
datumbox, line 207: There is an existing MLP block offered at TorchVision. See [here](https://github.com/pytorch/vision/blob/main/torchvision/ops/misc.py#L263).

torchvision/models/maxvit.py
```diff
@@ -0,0 +1,595 @@
...
+        partition_size: int,
+        input_grid_size: Tuple[int, int],
+        # number of layers
+        n_layers: int,
+        p_stochastic: List[float],
+    ) -> None:
+        super().__init__()
+        assert (
+            len(p_stochastic) == n_layers
+        ), f"p_stochastic must have length n_layers={n_layers}, got p_stochastic={p_stochastic}."
```
datumbox, line 391: Nit: We've recently replaced asserts with `ValueError` in the constructors.

torchvision/models/maxvit.py
```diff
@@ -0,0 +1,595 @@
...
+        activation_fn: Callable[..., nn.Module],
+        # transformer parameters
+        head_dim: int,
+        mlp_ratio: int,
+        mlp_dropout: float,
+        attn_dropout: float,
+        # partitioning parameters
+        partition_size: int,
+    ) -> None:
+        super().__init__()
```
datumbox, line 452: You need at least a logger here. Maybe in other blocks that you plan to expose on the API.
```suggestion
        super().__init__()
        _log_api_usage_once(self)
```

torchvision/models/maxvit.py
```diff
@@ -0,0 +1,595 @@
...
+
+    def forward(self, x: Tensor) -> Tensor:
+        x = self.stem(x)
+        for block in self.blocks:
+            x = block(x)
+        x = self.classifier(x)
+        return x
+
+
+def max_vit_T_224(num_classes: int) -> MaxVit:
```
datumbox, line 514: Add an Enum for the weights. You can have them empty like:

```python
class MaxVit_T_Weights(WeightsEnum):
    pass
```

Also I recommend renaming the `max_vit_T_224` to `maxvit_t` because:
- It will follow the standards of other classes
- we shouldn't hard code the input size in the method. (I'm aware the the model operates on fixed input size. See [ViT](https://github.com/pytorch/vision/blob/main/torchvision/models/vision_transformer.py) for similar approach)

Finally, make sure you register the model builder using the new API, so that it will appear on your tests. Here is how it's done: #6330. Feedback is very welcome concerning the mechanism as it's very new.

torchvision/models/maxvit.py
```diff
@@ -0,0 +1,595 @@
...
+        activation_fn=GeluWrapper,
+        head_dim=32,
+        mlp_ratio=2,
+        mlp_dropout=0.0,
+        attn_dropout=0.0,
+        partition_size=7,
+    )
+
+
+def max_vit_S_224(num_classes: int) -> MaxVit:
```
datumbox, line 535: I recommend dropping all other variants (and expected files) from here until we have fully trained the tiny version. It will reduce the amount of boilerplate code we have to review.

datumbox (commented):
@TeodorPoncu That was an unexpected surprise, thanks a lot for contributing this architecture.

I know that your PR is draft but I've added a few remarks mostly related to our coding styles and practices. I haven't verified the ML side of things. Feel free to ignore this is you thing my input is premature.

torchvision/models/maxvit.py
TeodorPoncu: Not necessarily, it can be directly included in the forward pass of the attention op. I've left as is for readability purposes on the user's side so it can be a bit more straightforward on how the relative bias interacts with the dot products.

test/test_models.py
TeodorPoncu: I've used that test as a way of solely testing the `maxvit` configurations. Is there any way to specify a subset of models on when running `test_classification_model`?

datumbox: The recommended way is to keep only 1 variant while you develop the main implementation. Not only the tests will be faster but also you reduce the LOC and boilerplate code of the PR, making it easier for others to review.

torchvision/models/maxvit.py
datumbox: Typically we avoid private methods that are used once but no strong opinions from my side. Up to you.

datumbox:
@TeodorPoncu It seems that in a recent commit, you accidentally updated all the expected files for all models. Could you please revert that?

TeodorPoncu:
@datumbox  Sorry about that, everything should be fine now.

> @TeodorPoncu It seems that in a recent commit, you accidentally updated all the expected files for all models. Could you please revert that?

vadimkantorov:
Related discussion on generalizing fixed resolution for Swin: https://github.com/pytorch/vision/issues/6227

torchvision/models/maxvit.py
```diff
@@ -0,0 +1,572 @@
...
+
+        self.merge = nn.Linear(self.head_dim * self.n_heads, feat_dim)
+        self.positional_bias = nn.parameter.Parameter(
+            torch.zeros(((2 * self.size - 1) * (2 * self.size - 1), self.n_heads), dtype=torch.float32),
+        )
+
+        self.register_buffer("relative_position_index", get_relative_position_index(self.size, self.size))
+
+        # initialize with truncated normal the bias
+        self.positional_bias.data.normal_(mean=0, std=0.02)
```
vadimkantorov, line 114: shouldn't https://pytorch.org/docs/stable/nn.init.html?highlight=nn%20init#torch.nn.init.normal_ be used here?

torchvision/models/maxvit.py
```diff
@@ -0,0 +1,572 @@
...
+        self.n_heads = feat_dim // head_dim
+        self.head_dim = head_dim
+        self.size = int(math.sqrt(max_seq_len))
+        self.max_seq_len = max_seq_len
+
+        self.to_qkv = nn.Linear(feat_dim, self.n_heads * self.head_dim * 3)
+        self.scale_factor = feat_dim**-0.5
+
+        self.merge = nn.Linear(self.head_dim * self.n_heads, feat_dim)
+        self.positional_bias = nn.parameter.Parameter(
```
vadimkantorov, line 107: if it's initizlized to normal just below, shouldn't torch.empty be used here?

torchvision/models/maxvit.py
```diff
@@ -0,0 +1,572 @@
...
+        )
+
+        self.register_buffer("relative_position_index", get_relative_position_index(self.size, self.size))
+
+        # initialize with truncated normal the bias
+        self.positional_bias.data.normal_(mean=0, std=0.02)
+
+    def get_relative_positional_bias(self) -> torch.Tensor:
+        bias_index = self.relative_position_index.view(-1)  # type: ignore
+        relative_bias = self.positional_bias[bias_index].view(self.max_seq_len, self.max_seq_len, -1)  # type: ignore
```
vadimkantorov, line 118: if it's just for flattening some end dimensions .flatten(start_dim = 2) can be used here

torchvision/models/maxvit.py
```diff
@@ -0,0 +1,572 @@
...
+
+
+class SwapAxes(nn.Module):
+    def __init__(self, a: int, b: int) -> None:
+        super().__init__()
+        self.a = a
+        self.b = b
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        res = torch.swapaxes(x, self.a, self.b)
```
vadimkantorov, line 153: swapaxes is NumPy-dialect alias for torch.transpose (https://pytorch.org/docs/stable/generated/torch.swapaxes.html#torch.swapaxes). If the rest is in Torch-lingo, shouldn't it be:
```python
class SwapAxes(nn.Module):
    def __init__(self, dim1: int, dim2: int) -> None:
        super().__init__()
        self.dim1 = dim1
        self.dim2 = dim2

    def forward(self, x: torch.Tensor) -> torch.Tensor:
        return x.transpose(self.dim1, self.dim2)
```

references/classification/run_with_submitit.py
```diff
@@ -0,0 +1,122 @@
+import argparse
```
datumbox: This shouldn't be submitted in the repo. :)

references/classification/train.py
```diff
@@ -182,7 +188,12 @@ def load_data(traindir, valdir, args):
 def main(args):
     if args.output_dir:
         utils.mkdir(args.output_dir)
-
+        
+    if args.seed is None:
+        # randomly choose a seed
+        args.seed = random.randint(0, 2 ** 32)
+    utils.set_seed(args.seed)
+
```
datumbox: What is the reason we need to set the seed? We don't do this anywhere else. Shall we remove it?

references/classification/utils.py
```diff
@@ -463,3 +465,13 @@ def _add_params(module, prefix=""):
...
+
+def set_seed(seed: int):
+    """
+    Function for setting all the RNGs to the same seed
+    """
+    torch.manual_seed(seed)
+    torch.cuda.manual_seed(seed)
+    torch.cuda.manual_seed_all(seed)
+    np.random.seed(seed)
+    random.seed(seed)
```
datumbox: I think we shouldn't offer this because PyTorch doesn't guarantee reproducibility for the random generator across versions. See [here](https://pytorch.org/docs/stable/notes/randomness.html).

references/classification/train.py
```diff
@@ -371,7 +391,7 @@ def get_args_parser(add_help=True):
 
     parser = argparse.ArgumentParser(description="PyTorch Classification Training", add_help=add_help)
 
-    parser.add_argument("--data-path", default="/datasets01/imagenet_full_size/061417/", type=str, help="dataset path")
+    parser.add_argument("--data-path", default="/datasets01_ontap/imagenet_full_size/061417/", type=str, help="dataset path")
```
datumbox: Revert. That's an implementation detail on our cluster.

references/classification/train.py
```diff
@@ -261,13 +272,21 @@ def main(args):
...
 
     scaler = torch.cuda.amp.GradScaler() if args.amp else None
+    
+    batches_per_epoch = len(data_loader)
+    warmup_iters = args.lr_warmup_epochs
+    total_iters = args.epochs
+    
+    if args.lr_step_every_batch:
+        warmup_iters *= batches_per_epoch
+        total_iters *= batches_per_epoch
```
datumbox, line 290: Hmmm... Though I agree that per iteration stepping gives better control, it's something that we should probably do once we migrate to a better trainer solution. The proposed change has some side-effects such as interactions with the EMA decay estimation and the overcomplication of the training loops. I think we should be able to reproduce the models without this change, even if we need to do a few more iterations to achieve it. WDYT?

datumbox (commented):
Thanks for the PR @TeodorPoncu. I've added some initial comments on the reference scripts updates. Happy to chat more.

references/classification/train.py
TeodorPoncu: I used it to perform a fixed ablation earlier on when experimenting with various configurations. I assume it's useful when there's no clear way of reproducing a specific experimentation and we want to remove seed randomness from the equation when quantifying effects.

references/classification/train.py
TeodorPoncu: This netted me about about 0.3% improvement. I can rerun without it in the final form to get quantifiable results.

references/classification/train.py
datumbox: I see, so some features are still from the moment you did the ablations. Now that this is completed, we should remove them. This simplification is a standard practice for every model that we port.

references/classification/train.py
datumbox: > This netted me about about 0.3% improvement.

I think that's acceptable. You can also try to run it a little bit longer to have more gradient updates and perhaps close the gap. 

> I can rerun without it in the final form to get quantifiable results.

Sounds good!

TeodorPoncu:
Running the deployed weights with the following command:
`torchrun --nproc_per_node=1 train.py --model maxvit_t --interpolation bicubic --batch-size 1 --test-only --weights MaxVit_T_Weights.IMAGENET1K_V1`

Yields the following results:
`Test:  Acc@1 83.700 Acc@5 96.722`

torchvision/models/maxvit.py
```diff
@@ -0,0 +1,854 @@
...
+        # attention layers, block -> grid
+        layers["window_attention"] = PartitionAttentionLayer(
+            in_channels=out_channels,
+            head_dim=head_dim,
+            partition_size=partition_size,
+            partition_type="window",
+            grid_size=grid_size,
+            mlp_ratio=mlp_ratio,
+            activation_fn=activation_fn,
+            normalization_fn=nn.LayerNorm,
```
datumbox, line 445: Just confirming that this is intentionally hardcoded and not `normalization_fn`, right?

torchvision/models/maxvit.py
```diff
@@ -0,0 +1,854 @@
...
+        # activation_fn is applied both to conv and transformer layers
+        normalization_fn: Callable[..., nn.Module],
+        activation_fn: Callable[..., nn.Module],
+        # transformer parameters
+        head_dim: int,
+        mlp_ratio: int,
+        mlp_dropout: float,
+        attn_dropout: float,
+        # partitioning parameters
+        partition_size: int,
```
datumbox, line 609: Quite a few parameters with several non-default configs. Consider adding some good default values for some of these configurations.

Also usually things like `squeeze_ratio`, `expansion_ratio` are constant across model variants and usually get hardcoded. These parameters are often configurable in research repos as a means to reproduce the ablation experiments. When we clean up the implementations, typically we reduce the number of parameters to those necessary to differentiate across canonical variants.

Perhaps it's worth looking into Swin Transformer or ViT to see which parameters are usually set to defaults and the naming/order convention across the parameters. For example `num_classes=1000` or `normalization_fn` is usually called `norm_layer`.

torchvision/models/maxvit.py
```diff
@@ -0,0 +1,854 @@
...
+    attn_dropout: float,
+    # partitioning parameters
+    partition_size: int,
+    # Weights API
+    weights: Optional[WeightsEnum],
+    progress: bool,
+    # task parameters
+    num_classes: int = 1000,
+    # kwargs,
+    **kwargs,
```
datumbox, line 745:
```suggestion
    **kwargs: Any,
```

torchvision/models/maxvit.py
```diff
@@ -0,0 +1,854 @@
...
+    mlp_ratio: int,
+    mlp_dropout: float,
+    attn_dropout: float,
+    # partitioning parameters
+    partition_size: int,
+    # Weights API
+    weights: Optional[WeightsEnum],
+    progress: bool,
+    # task parameters
+    num_classes: int = 1000,
```
datumbox, line 743: Minor Nit: Usually num_classes and other parameters that are default for the class are not defined here as mandatory parameters and we rely on the Model Constructor to choose the values. If the user chooses to overwrite them that happens using the `kwargs`

torchvision/models/maxvit.py
```diff
@@ -0,0 +1,854 @@
...
+        attn_dropout (float): Dropout probability for the attention layer.
+        partition_size (int): Size of the partitions.
+    """
+
+    def __init__(
+        self,
+        # input size parameters
+        input_size: Tuple[int, int],
+        # stem and task parameters
+        input_channels: int,
```
datumbox, line 588: No other model at the moment accepts a `input_channels` parameter. It has been a request from the community to introduce it to support single-channel (and other) images. Perhaps it's something we want to do, but I would see it as a coordinated effort to make it available to all models.

torchvision/models/maxvit.py
```diff
@@ -0,0 +1,854 @@
...
+
+    if weights is not None:
+        model.load_state_dict(weights.get_state_dict(progress=progress))
+
+    return model
+
+
+_COMMON_META = {
+    "categories": _IMAGENET_CATEGORIES,
+}
```
datumbox, line 783: Nit: You don't need to define one given you only have one set of weights. You could just move the key below.

datumbox (commented):
@TeodorPoncu pretty awesome work and top quality code.

I've left a few Nits here and there, just to make sure that the implementation code aligns with the idioms of the rest of TorchVision. The only comment worth considering a bit more is the one concerning the number of parameters you will pass on the constructor. This comment is mostly to align `MaxViT` with all other models and make it easier on the future to make changes to all of them. In some cases in particular, you expose parameters that are indeed quite useful (num of input channels), but such changes on the API would be best if we introduce them in all models, not just MaxViT.

Other than the above, I didn't validate much the architecture part of things but focused mainly on idioms and code styles. I know that you've already reproduced the accuracy of the paper, which is great but it's worth doing one scan with @jdsgomes prior merging to confirm that there are no deviations from the original implementation (for example on padding of input whose spatial dimensions are not divisible by `p`).

All CI failures seems unrelated. Other than these final validations and nits, I think the implementation is in awesome shape and we should be able to merge it soon.

references/classification/presets.py
```diff
@@ -13,14 +13,20 @@ def __init__(
         interpolation=InterpolationMode.BILINEAR,
         hflip_prob=0.5,
         auto_augment_policy=None,
+        policy_magnitude=9,
```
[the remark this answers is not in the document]
TeodorPoncu: From the research stand-point, it's nice to have as many options as possible. My suggestion would be to add the support for `AugMix` as well. I've personally had a use-case in Medical Imaging where the whole thing was extremely sensitive to the choice of severity in `AugMix`.

datumbox:
Two more requests:
- Could you please upload the weights on manifold (see internal guide)
- Could you update the PR description to show-case the output accuracy of the following command?
```
torchrun --nproc_per_node=1 train.py --test-only --prototype --weights MaxVit_T_Weights.IMAGENET1K_V1 --model maxvit_t -b 1
```

torchvision/models/maxvit.py
TeodorPoncu: Yes, this is intentionally hard-coded. `normalization_fn ` is destined to be used only in conjunction with the convolutional layers.

references/classification/presets.py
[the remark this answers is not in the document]
TeodorPoncu: Making it more generic comes with the downside that we might not be able to keep BC in parameter configurations. Since the default magnitude for `RandAugment` is 9 whilst the default severity for `AugMix` is 3.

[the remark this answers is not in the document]
datumbox: ah good call out. You could pass `None` as default and let the class choose its default if this happens but we might be overcomplicating things. Perhaps the cleanest approach is to just go with the `ra_magnitude` approach. We will review our references in H1 to clean them up and avoid the mess with so many parameters. I think for now, the target is to make your work reproducible for others so adding the extra parameter will do.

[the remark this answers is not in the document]
TeodorPoncu: I've just opted for the same `.getattr` we're using for `random_erase_prob`.

datumbox (commented):
Thanks for all the changes, few follow ups:

torchvision/models/maxvit.py
```diff
@@ -0,0 +1,829 @@
...
+        # stem
+        self.stem = nn.Sequential(
+            Conv2dNormActivation(
+                input_channels,
+                stem_channels,
+                3,
+                stride=2,
+                norm_layer=normalization_layer,
+                activation_layer=activation_layer,
+                bias=False,
```
jdsgomes, line 648: Should this be true? https://github.com/google-research/maxvit/blob/main/maxvit/models/maxvit.py#L1084

torchvision/models/maxvit.py
```diff
@@ -0,0 +1,829 @@
...
+                inplace=None,
+            ),
+            Conv2dNormActivation(
```
jdsgomes, line 651: can this be replaced by conv2d only?

torchvision/models/maxvit.py
TeodorPoncu: It gets followed by a norm-layer which basically renders the bias as a no-op since it gets normalised w.r.t. the mean. https://github.com/google-research/maxvit/blob/da76cf0d8a6ec668cc31b399c4126186da7da944/maxvit/models/maxvit.py#L1088

In the paper they also cite: https://arxiv.org/pdf/2106.14881.pdf which basically follows the Conv-Norm-Act paradigm.

torchvision/models/maxvit.py
TeodorPoncu: This could be changed, however I guess that would require us to translate the weight dictionary to accomodate the change

torchvision/models/maxvit.py
```diff
@@ -0,0 +1,829 @@
...
+        # normalization_layer is applied only to the conv layers
+        # activation_layer is applied both to conv and transformer layers
+        normalization_layer: Optional[Callable[..., nn.Module]] = None,
+        activation_layer: Callable[..., nn.Module] = nn.GELU,
+        # conv parameters
+        squeeze_ratio: float = 0.25,
+        expansion_ratio: float = 4,
+        # transformer parameters
+        mlp_ratio: int = 4,
+        mlp_dropout: float = 0.0,
```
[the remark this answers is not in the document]
TeodorPoncu: These are mainly left as is for `**kwargs` especially the dropouts for regularization and the ratios for parameter count tuning.

torchvision/models/maxvit.py
jdsgomes: I would keep it as is since its equivalent

torchvision/models/maxvit.py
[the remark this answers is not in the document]
jdsgomes: we do have for swin, but possibly was missed then. @datumbox do you think it is necessary to remove?

https://github.com/pytorch/vision/blob/main/torchvision/models/swin_transformer.py#L530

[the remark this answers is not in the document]
datumbox: ah good spot, I missed this. No it's fine by me.

test/test_architecture_ops.py
```diff
@@ -0,0 +1,46 @@
+import unittest
+
+import pytest
+import torch
+
+from torchvision.models.maxvit import SwapAxes, WindowDepartition, WindowPartition
+
+
+class MaxvitTester(unittest.TestCase):
```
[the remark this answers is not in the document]
YosuaMichael: Sorry for a pretty late response!
Currently I dont have any opinion how we should test specific layer of the model and I think this is okay. (Need more time to think and discuss whether we should do more of this kind of test or not)

references/classification/presets.py
```diff
@@ -13,18 +13,25 @@ def __init__(
...
+        ra_magnitude=9,
+        augmix_severity=3,
         random_erase_prob=0.0,
+        center_crop=False,
     ):
-        trans = [transforms.RandomResizedCrop(crop_size, interpolation=interpolation)]
+        trans = (
+            [transforms.RandomResizedCrop(crop_size, interpolation=interpolation)]
+            if center_crop
+            else [transforms.CenterCrop(crop_size)]
```
datumbox, line 24: @TeodorPoncu I think this is a bug. I believe you meant to write:
```python
            [transforms.CenterCrop(crop_size)]
            if center_crop
            else [transforms.RandomResizedCrop(crop_size, interpolation=interpolation)]
```

Could you please confirm?

TeodorPoncu: Yes, sorry for that. You've correctly guessed what I wanted to write. Thanks for catching it out. I think the `--train-center-crop` flag should be removed from the docs as well to reflect the way the weights were trained.

datumbox: @TeodorPoncu thanks for coming back to me. Does this mean that you didn't actually use the flag during training? Can we remove it?

TeodorPoncu: Yes, the flag can be removed.

datumbox: That's not what I see on the training log you used on the trained model. I see that `train_center_crop=True`. Do we have the right model available on the checkpoint area on AWS?

TeodorPoncu: Yes, but given the fix in order to replicate, one will have to run with `train_center_crop=True` in order to have the same preprocessing behavior as the AWS weights had.

TeodorPoncu: Yes, the bug was introduced and used when performing the training as in 1fddecc8d4e3cadafa33993dc82b0f82b93a8dd7

datumbox: Agreed. I checked all commits prior f561edfc8f012425cb30986439eaacee2695e3cf (date before training) and all of them use RandomCrop. I'll remove the flag.

PR #6343, opened 2022-08-02
Fix d/c IoU for different batch sizes

datumbox:
Cherrypick of #6338

### DIFF
--- a/torchvision/ops/boxes.py
+++ b/torchvision/ops/boxes.py
@@ -331,7 +331,7 @@ def complete_box_iou(boxes1: Tensor, boxes2: Tensor, eps: float = 1e-7) -> Tenso
     w_gt = boxes2[:, 2] - boxes2[:, 0]
     h_gt = boxes2[:, 3] - boxes2[:, 1]
 
-    v = (4 / (torch.pi**2)) * torch.pow(torch.atan(w_pred / h_pred) - torch.atan(w_gt / h_gt), 2)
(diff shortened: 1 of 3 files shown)


torchvision/ops/boxes.py
```diff
@@ -325,13 +325,13 @@ def complete_box_iou(boxes1: Tensor, boxes2: Tensor, eps: float = 1e-7) -> Tenso
 
     diou, iou = _box_diou_iou(boxes1, boxes2, eps)
 
-    w_pred = boxes1[:, 2] - boxes1[:, 0]
-    h_pred = boxes1[:, 3] - boxes1[:, 1]
+    w_pred = boxes1[:, None, 2] - boxes1[:, None, 0]
```
jdsgomes, line 328: why we need this None for the pred and not for the gt boxes? I might be missing something obvious..

jdsgomes (approved):
LGTM, thanks

torchvision/ops/boxes.py
datumbox: Good question! I had the same broadcast originally but @fmassa pointed out to me that it can be simplified to this. See commit: https://github.com/pytorch/vision/pull/6338/commits/7bef3d047e33572bb84343f507401560be1bdbe9

jdsgomes: Thank you for the answer here https://github.com/pytorch/vision/pull/6338#discussion_r935351475

PR #6345, opened 2022-08-02
Added SceneFLow variant datasets

TeodorPoncu:
This is a continuation of the PR split (#6311, #6269) which contains the SceneFlow dataset variants.

### DIFF
--- a/test/test_datasets.py
+++ b/test/test_datasets.py
@@ -2873,8 +2873,6 @@ class SceneFlowStereoTestCase(datasets_utils.ImageDatasetTestCase):
             "both": ["frames_cleanpass", "frames_finalpass"],
         }.get(config["pass_name"], [])
 
-        # pass_dir_name = pass_dir_map.get(config["pass_name"], None)
-
         for pass_dir_name in passes:
             # create pass directories
             pass_dir = variant_dir / pass_dir_name
--- a/test/test_datasets.py
+++ b/test/test_datasets.py
@@ -13,7 +13,7 @@ import string
 import unittest
 import xml.etree.ElementTree as ET
 import zipfile
-from typing import Union
+from typing import Callable, Tuple, Union
 
 import datasets_utils
 import numpy as np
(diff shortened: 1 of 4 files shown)


test/test_datasets.py
```diff
@@ -2841,5 +2841,79 @@ def test_train_splits(self):
...
+
+        split_dir = scene_flow_dir / config["split"]
+        os.makedirs(split_dir, exist_ok=True)
+
+        pass_dir_map = {
+            "clean": "frames_cleanpass",
+            "final": "frames_finalpass",
+        }
+
+        num_examples = 1
```
NicolasHug, line 2873:
```suggestion
```

torchvision/datasets/_stereo_matching.py
```diff
@@ -359,3 +359,105 @@ def __getitem__(self, index: int) -> Tuple:
...
+                            right
+                                img1.pfm
+                                img2.pfm
+                FlyingThings3D
+                    ...
+                    ...
+
+    Args:
+        root (string): Root directory where SceneFlow is located.
+        split (string): Which dataset variant to user, "FlyingThings3D" (default), "Monkaa" or "Driving".
```
NicolasHug, line 412: Following up on our discussoin in https://github.com/pytorch/vision/pull/6269#discussion_r921184790, it looks like we don't really need to expose this dataset. There's no obvious use-case yet, so I'm happy to drop it.

If really we want to allow users to control this, I think we should just call this parameter `variant` instead of `split`, as `split` is usually reserved for train / test / val. But perhaps the  dataset authors have a dedicated name already?

test/test_datasets.py
```diff
@@ -2841,5 +2841,79 @@ def test_train_splits(self):
                 datasets_utils.shape_test_for_stereo(left, right, disparity)
 
 
+class SceneFlowStereoTestCase(datasets_utils.ImageDatasetTestCase):
+    DATASET_CLASS = datasets.SceneFlowStereo
+    ADDITIONAL_CONFIGS = datasets_utils.combinations_grid(
+        split=("FlyingThings3D", "Driving", "Monkaa"), pass_name=("clean", "final")
```
NicolasHug, line 2847:
```suggestion
        split=("FlyingThings3D", "Driving", "Monkaa"), pass_name=("clean", "final", "both")
```

NicolasHug (commented):
Thanks @TeodorPoncu , just a few comments but it looks great already

torchvision/datasets/_stereo_matching.py
TeodorPoncu: In the CREStereo paper, they have a vague formulation from which one could deduce that they are using ALL mentioned datasets (about 6 synthetic datasets if I recall correctly).

I agree that variant might be a better naming scheme.

torchvision/datasets/_stereo_matching.py
```diff
@@ -359,3 +359,109 @@ def __getitem__(self, index: int) -> Tuple:
...
+        root: str,
+        variant: str = "FlyingThings3D",
+        pass_name: str = "clean",
+        transforms: Optional[Callable] = None,
+    ):
+        super().__init__(root, transforms)
+
+        root = Path(root) / "SceneFlow"
+
+        verify_str_arg(variant, "variant", valid_values=("FlyingThings3D", "Driving", "Monkaa"))
```
NicolasHug, line 429: I'm a bit confused: I thought we were relying on all variants for the training? Do we want to allow `"all"` here as well?

NicolasHug (commented):
Thanks @TeodorPoncu LGTM, just one final question regarding the `variant` parameter

torchvision/datasets/_stereo_matching.py
TeodorPoncu: Well, in the Optical Flow datasets we directly provide access to FlyingThings3D. Some authors use all the variants, or just a subset of them. An "all" split could work as well, however if a user requires some sort of variant combination I guess they could opt for `ConcatDataset`, without requiring us to provide  a way of handling variant permutations. 

If I recall correctly we settled for removing variants for the CREStereo Dataset since the only existing use-case is to use it with all its variants (#6351,  https://github.com/pytorch/vision/pull/6269#discussion_r921999054).

NicolasHug: > If I recall correctly we settled for removing variants for the CREStereo Dataset since the only existing use-case is to use it with all its variants (https://github.com/pytorch/vision/pull/6351, https://github.com/pytorch/vision/pull/6269#discussion_r921999054).

Is this the case here as well?

Honestly I'm fine with either at this point, I'm just surprised that we don't need `"all"` because this is what we needed for CREStereo

TeodorPoncu: We could add an "all" split as well. As a user, generally my preference is to have access to a "smaller" version of a dataset for quick experimentation / validation. Similarly to the ImageNet vs. [ImageNette](https://github.com/fastai/imagenette) scenario. 

As far as I am aware [RAFT-Stereo ](https://github.com/princeton-vl/RAFT-Stereo) uses all 3 variants as well, however I do not have an exhaustive list to guarantee that this is a universally used approach.

NicolasHug: OK, well let's just leave it as is and figure out whether not having it complicates things in the training reference. Thanks for the details

NicolasHug (approved):
LGTM, thanks @TeodorPoncu

torchvision/datasets/_stereo_matching.py
TeodorPoncu: For instance, this is a configuration example for a dataset chain schedule, similarly to how RAFT-Stereo is used, where the datasets are trained in sequential order, performing optimisation using samples from the dataset for the specified number of steps. Adding the "all" variant or removing the variant argument all-together would result in:

```yaml
train_dataset: ["instereo-2k", "sintel", "sceneflow"]
dataset_steps: [200_000, 150_000, 100_000]
```

Having the current version would yield:
```yaml
train_dataset: ["instereo-2k", "sintel", "flythings3d", "monkaa", "driving"]
dataset_steps: [200_000, 150_000, 33_000, 33_000, 33_000]
```

PR #6346, opened 2022-08-02
Add FallingThings dataset

TeodorPoncu:
This is a continuation of the PR split (https://github.com/pytorch/vision/pull/6311, https://github.com/pytorch/vision/pull/6269) which contains the FallingThings dataset variants.


torchvision/datasets/_stereo_matching.py
```diff
@@ -359,3 +360,101 @@ def __getitem__(self, index: int) -> Tuple:
...
+                    scene2
+                    ...
+
+    Args:
+        root (string): Root directory where FallingThings is located.
+        variant (string): Which variant to use. Either "single", "mixed", or "both".
+        transforms (callable, optional): A function/transform that takes in a sample and returns a transformed version.
+    """
+
+    def __init__(self, root: str, variant: str = "single", transforms: Optional[Callable] = None):
```
NicolasHug, line 409: Just trying to figure out whether there are relevant alternatives for "variant" here. Looking at the authors' README I could only find "scene type".

> At the root level, there are two folders representing the two types of scenes: 
> - `single` (single falling object), and 
> - `mixed` (2 to 10 falling objects).

Not sure it's better though...

torchvision/datasets/_stereo_matching.py
```diff
@@ -359,3 +360,101 @@ def __getitem__(self, index: int) -> Tuple:
...
+        # (H, W) image
+        depth = np.asarray(Image.open(file_path))
+        # as per https://research.nvidia.com/sites/default/files/pubs/2018-06_Falling-Things/readme_0.txt
+        # in order to extract disparity from depth maps
+        camera_settings_path = Path(file_path).parent / "_camera_settings.json"
+        with open(camera_settings_path, "r") as f:
+            # inverse of depth-from-disparity equation: depth = (baseline * focal) / (disparity * pixel_constatnt)
+            intrinsics = json.load(f)
+            focal = intrinsics["camera_settings"][0]["intrinsic_settings"]["fx"]
+            baseline, pixel_constant = 6.0, 100.0  # pixel constant is inverted
```
NicolasHug, line 441: Nit

```suggestion
            baseline, pixel_constant = 6, 100  # pixel constant is inverted
```

NicolasHug (approved):
Thanks @TeodorPoncu , minor comments from me but LGTM anyway

torchvision/datasets/_stereo_matching.py
TeodorPoncu: I've got no particular feelings about this, it's definitely not a "split" therefor I just opted for "variant" to keep consistent with other datasets.

PR #6349, opened 2022-08-02
Added ETH3D stereo dataset

TeodorPoncu:
This is a continuation of the PR split (https://github.com/pytorch/vision/pull/6311, https://github.com/pytorch/vision/pull/6269) which contains the ETH3D low-res two-view dataset.


torchvision/datasets/_stereo_matching.py
```diff
@@ -359,3 +359,103 @@ def __getitem__(self, index: int) -> Tuple:
...
+    def _read_disparity(self, file_path: str) -> Tuple:
+        # test split has no disparity maps
+        if file_path is None:
+            return None, None
+
+        disparity_map = _read_pfm_file(file_path)
+        disparity_map = np.abs(disparity_map)  # ensure that the disparity is positive
+        mask_path = Path(file_path).parent / "mask0nocc.png"
+        valid_mask = Image.open(mask_path)
+        valid_mask = np.asarray(valid_mask).astype(np.bool_)
```
NicolasHug, line 445: I was surprised by the us of `np.bool_` and looking at numpy's warning, I think this should just be the buildin `bool` https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated

test/test_datasets.py
```diff
@@ -2841,5 +2841,75 @@ def test_train_splits(self):
...
+            annot_dir = eth3d_dir / "two_view_training_gt"
+            self._create_annotation_folder(num_examples, annot_dir)
+
+        return num_examples
+
+    def test_training_splits(self):
+        with self.create_dataset(split="train") as (dataset, _):
+            assert dataset._images and len(dataset._images) == len(
+                dataset._disparities
+            ), "Training images do not match with training disparities"
```
NicolasHug, line 2897: Curious why we need this assertion. Isn't it taken care of in the `shape_test_...` call below?

NicolasHug (approved):
Thanks @TeodorPoncu , minor comments but LGTM anyway

test/test_datasets.py
TeodorPoncu: That assertion is unnecessary indeed! Thanks!

PR #6350, opened 2022-08-02
Added Middlebury2014  dataset

TeodorPoncu:
This is a continuation of the PR split (https://github.com/pytorch/vision/pull/6311, https://github.com/pytorch/vision/pull/6269) which contains the Middlebury 2014 version dataset.


torchvision/datasets/_stereo_matching.py
```diff
@@ -359,3 +362,255 @@ def __getitem__(self, index: int) -> Tuple:
...
+
+    def _read_img(self, file_path: str) -> Image.Image:
+        """
+        Function that reads either the original right image or an augmented view when ``use_ambient_views`` is True.
+        When ``use_ambient_views`` is True, the dataset will return at random one of ``[im1.png, im1E.png, im1L.png]``
+        as the right image.
+        """
+        ambient_file_paths: List[Union[str, Path]]  # make mypy happy
+
+        if os.path.basename(file_path) == "im1.png" and self.use_ambient_views:
```
jdsgomes, line 544: Nit: you can convert file_path to Path in the beginning of the method instead of the line bellow and use file_path.name here

torchvision/datasets/_stereo_matching.py
```diff
@@ -359,3 +362,255 @@ def __getitem__(self, index: int) -> Tuple:
...
+
+        if split_name != "test":
+            for split_scene in self.splits[split_name]:
+                split_root = root / split_name
+                for calibration in ["perfect", "imperfect"]:
+                    scene_name = f"{split_scene}-{calibration}"
+                    scene_url = f"{base_url}/{scene_name}.zip"
+                    print(f"Downloading {scene_url}")
+                    # download the scene only if it doesn't exist
+                    if not os.path.exists(split_root / scene_name):
```
jdsgomes, line 579: Another nit: you can just use exists from pathlib: (solit_root / scene_name).exists()

jdsgomes (approved):
LGTM, just left some nits

test/test_datasets.py
```diff
@@ -2841,5 +2841,98 @@ def test_train_splits(self):
...
+            "test": ["Plants", "Classroom2E", "Classroom2", "Australia"],
+        }
+
+        middlebury_dir = pathlib.Path(tmpdir, "Middlebury2014")
+        os.makedirs(middlebury_dir, exist_ok=True)
+
+        split_dir = middlebury_dir / config["split"]
+        os.makedirs(split_dir, exist_ok=True)
+
+        num_examples = {"train": 2, "additional": 3, "test": 4}.get(config["split"], 0)
```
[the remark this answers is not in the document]
TeodorPoncu: The issue with letting it fail is that from what I've experienced so far, when we are testing for the dataset behaviour in the case when a bad argument is passed to the constructor, when injecting fake data that method will crash resulting in the test failing. The alternative would be to do error handling when injecting fake data. https://github.com/pytorch/vision/blob/336524d30ff92d2acddb4b74a41755b245b512bd/test/test_datasets.py#L2913-L2919

PR #6353, opened 2022-08-02
Add missing None type hints to tests

ProGamerGov:
Tests should be type hinted with `-> None`.

### DIFF
--- a/test/datasets_utils.py
+++ b/test/datasets_utils.py
@@ -67,7 +67,7 @@ class LazyImporter:
         "h5py",
     )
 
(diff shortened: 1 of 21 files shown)


datumbox:
@ProGamerGov Thanks for the contribution.

I don't think we currently add typing on the tests. Maintaining types has benefits for main TorchVision but it's too much of a hassle on the testing side. I believe only a couple of cases have typing to stress the expected types but the vast majority of methods by design don't support them.

ProGamerGov:
@datumbox Ah, okay!

PR #6198, opened 2022-06-23
Add MViT architecture in TorchVision

datumbox:
This PR adds on TorchVision the Video variant of "Multiscale Vision Transformers". The model can also handle images by considering them single frame videos. This PR contains changes from #6086, #6105 and #6179.

The implementations is based on [MViTv1](https://arxiv.org/abs/2104.11227) but includes some extensions from [MViTv2](https://arxiv.org/abs/2112.01526) (the Attention supports a `residual_pool` option). The `mvit_v1_b` variant introduced is canonical and the weights are ported from the paper. This is based on the work of @haooooooqi, @feichtenhofer and @lyttonhao on [PyTorchVideo](https://github.com/facebookresearch/pytorchvideo).

# Verification process

## Comparing outputs

To confirm that the implementation is compatible with the original from PyTorch Video we create a weight converter, load the same weights for both implementations and compare them against the same input:
```python
import collections

import torch
from pytorchvideo.models.hub.vision_transformers import mvit_base_16x4
from pytorchvideo.models.vision_transformers import create_multiscale_vision_transformers
from torchvision.models.video import mvit as TorchVision


class PyTorchVideo:
    @staticmethod
    def mvit_v1_b(**kwargs):
        return create_multiscale_vision_transformers(
            spatial_size=(224, 224),
            temporal_size=16,
            depth=16,
            embed_dim_mul=[[1, 2.0], [3, 2.0], [14, 2.0]],
            atten_head_mul=[[1, 2.0], [3, 2.0], [14, 2.0]],
            pool_q_stride_size=[[1, 1, 2, 2], [3, 1, 2, 2], [14, 1, 2, 2]],
            droppath_rate_block=0.2,
            # additional params for PyTorch Video
            residual_pool=False,
            separate_qkv=False,
            pool_kv_stride_adaptive=(1, 8, 8),
            pool_kvq_kernel=(3, 3, 3),
            head_dropout_rate=0.0,
            **kwargs,
        )

    @staticmethod
    def mvit_base_16x4(**kwargs):
        # return mvit_base_16x4(pretrained=True)
        m = create_multiscale_vision_transformers(
            spatial_size=(224, 224),
            temporal_size=16,
            depth=16,
            embed_dim_mul=[[1, 2.0], [3, 2.0], [14, 2.0]],
            atten_head_mul=[[1, 2.0], [3, 2.0], [14, 2.0]],
[...]

PR #6373, opened 2022-08-05
Add support of MViTv2 video variants

datumbox:
This is the continuation of the work from #6198 to finalize the API of MViT class. The PR extends TorchVision's existing MViT architecture to support [v2](https://arxiv.org/abs/2112.01526) variants.

The `mvit_v2_s` variant introduced is canonical and the weights are ported from the paper. This is based on the work of @lyttonhao, @haooooooqi and @feichtenhofer on [SlowFast](https://github.com/facebookresearch/SlowFast).

# Verification process

## Comparing outputs

To confirm that the implementation is compatible with the original from SlowFast we create a weight converter, load the same weights for both implementations and compare them against the same input:
```python
import collections
import tempfile
import torch
from torchvision.models.video import mvit_v2_s
from slowfast.config.defaults import assert_and_infer_cfg, get_cfg
from slowfast.models.video_model_builder import MViT
from slowfast.utils.parser import load_config


def mvit_v2_s_slowfast():
    config = """
    DATA:
      NUM_FRAMES: 16
      SAMPLING_RATE: 4
      TRAIN_CROP_SIZE: 224
      TEST_CROP_SIZE: 224
      INPUT_CHANNEL_NUM: [3]
    MVIT:
      ZERO_DECAY_POS_CLS: False
      USE_ABS_POS: False
      REL_POS_SPATIAL: True
      REL_POS_TEMPORAL: True
      DEPTH: 16
      NUM_HEADS: 1
      EMBED_DIM: 96
      PATCH_KERNEL: (3, 7, 7)
      PATCH_STRIDE: (2, 4, 4)
      PATCH_PADDING: (1, 3, 3)
      MLP_RATIO: 4.0
      QKV_BIAS: True
      DROPPATH_RATE: 0.2
      NORM: "layernorm"
      MODE: "conv"
      CLS_EMBED_ON: True
      DIM_MUL: [[1, 2.0], [3, 2.0], [14, 2.0]]
      HEAD_MUL: [[1, 2.0], [3, 2.0], [14, 2.0]]
      POOL_KVQ_KERNEL: [3, 3, 3]
      POOL_KV_STRIDE_ADAPTIVE: [1, 8, 8]
      POOL_Q_STRIDE: [[0, 1, 1, 1], [1, 1, 2, 2], [2, 1, 1, 1], [3, 1, 2, 2], [4, 1, 1, 1], [5, 1, 1, 1], [6, 1, 1, 1], [7, 1, 1, 1], [8, 1, 1, 1], [9, 1, 1, 1], [10, 1, 1, 1], [11, 1, 1, 1], [12, 1, 1, 1], [13, 1, 1, 1], [14, 1, 2, 2], [15, 1, 1, 1]]
      DROPOUT_RATE: 0.0
      DIM_MUL_IN_ATT: True
      RESIDUAL_POOLING: True
    """
    temp = tempfile.NamedTemporaryFile(mode="w+", delete=False)
    try:
        temp.write(config)
    finally:
        temp.close()

    cfg = get_cfg()
    cfg.merge_from_file(temp.name)
    cfg = assert_and_infer_cfg(cfg)

    cfg.NUM_GPUS = 0
    model = MViT(cfg)
    model.head.act = torch.nn.Identity()
    return model


def slowfast_to_tv_weights(state_dict):
    d = dict(state_dict)

    # remapping keys
    mapping = collections.OrderedDict(
        [
            ("patch_embed.project.weight", "conv_proj.weight"),
            ("patch_embed.project.bias", "conv_proj.bias"),
            ("cls_token", "pos_encoding.class_token"),
            ("pos_embed_spatial", "pos_encoding.spatial_pos"),
            ("pos_embed_temporal", "pos_encoding.temporal_pos"),
            ("pos_embed_class", "pos_encoding.class_pos"),
            ("attn.proj.weight", "attn.project.0.weight"),
            ("attn.proj.bias", "attn.project.0.bias"),
            ("attn.pool_q.weight", "attn.pool_q.pool.weight"),
            ("attn.norm_q.weight", "attn.pool_q.norm_act.0.weight"),
            ("attn.norm_q.bias", "attn.pool_q.norm_act.0.bias"),
            ("attn.pool_k.weight", "attn.pool_k.pool.weight"),
            ("attn.norm_k.weight", "attn.pool_k.norm_act.0.weight"),
            ("attn.norm_k.bias", "attn.pool_k.norm_act.0.bias"),
            ("attn.pool_v.weight", "attn.pool_v.pool.weight"),
            ("attn.norm_v.weight", "attn.pool_v.norm_act.0.weight"),
            ("attn.norm_v.bias", "attn.pool_v.norm_act.0.bias"),
            ("mlp.fc1.weight", "mlp.0.weight"),
            ("mlp.fc1.bias", "mlp.0.bias"),
            ("mlp.fc2.weight", "mlp.3.weight"),
            ("mlp.fc2.bias", "mlp.3.bias"),
            ("head.projection.weight", "head.1.weight"),
            ("head.projection.bias", "head.1.bias"),
            ("patch_embed.proj.weight", "conv_proj.weight"),
            ("patch_embed.proj.bias", "conv_proj.bias"),
            ("norm.weight", "norm.weight"),
            ("norm.bias", "norm.bias"),
            ("proj.weight", "project.weight"),
            ("proj.bias", "project.bias"),
        ]
    )
    for k in list(d.keys()):
        for pattern, replacement in mapping.items():
            if pattern in k:
                new_key = k.replace(pattern, replacement)
                d[new_key] = d.pop(k)
                break

    # matching dimensions
    d["pos_encoding.class_token"] = d["pos_encoding.class_token"][0, 0, :]
    if "pos_encoding.spatial_pos" in d:
        d["pos_encoding.spatial_pos"] = d["pos_encoding.spatial_pos"][0, :]
        d["pos_encoding.temporal_pos"] = d["pos_encoding.temporal_pos"][0, :]
        d["pos_encoding.class_pos"] = d["pos_encoding.class_pos"][0, 0, :]

    return d


def compare_models(sf_model_fn, tv_model_fn, input_shape):
    print(tv_model_fn.__name__)
    x = torch.randn(input_shape)

    sf_m = sf_model_fn().eval()
    exp_result = sf_m([x]).sum()

    d = sf_m.state_dict()
    d = slowfast_to_tv_weights(d)

    tv_m = tv_model_fn().eval()
    tv_m.load_state_dict(d)
    result = tv_m(x).sum()

    torch.testing.assert_close(result, exp_result)
    print("OK")


compare_models(mvit_v2_s_slowfast, mvit_v2_s, (1, 3, 16, 224, 224))
```

## Benchmarks

To ensure that we don't introduce any speed regression we test the speed as follows:
```python
import time


def benchmark(model_fn, input_shape, device, put_in_list, n=5, warmup=0.1):
    torch.manual_seed(42)
    m = model_fn().to(device).eval()
    x = torch.randn(input_shape).to(device)
    if put_in_list:
        x = [x]

    s = []
    for i in range(n):
        start = time.time()
        m(x)
        t = time.time() - start
        if i > n * warmup:
            s.append(t)

    print(model_fn.__name__, torch.tensor(s).median())


device = "cuda"
batch_size = 4
n = 100

print(f"device={device}, batch_size={batch_size}, n={n}")
for name, fn, put_in_list in [("TorchVision", mvit_v2_s, False), ("SlowFast", mvit_v2_s_slowfast, True)]:
    print(name)
    benchmark(fn, (batch_size, 3, 16, 224, 224), device, put_in_list, n=n)
```

This was tested on an A100 and as we see below the implementation is 5% faster than the original:
```
device=cuda, batch_size=4, n=100
TorchVision
mvit_v2_s tensor(0.0492)
SlowFast
mvit_v2_s_slowfast tensor(0.0520)
```

## Accuracy

To verify the accuracy of the model we run the following:
```
torchrun --nproc_per_node=8 train.py --data-path="/datasets/clean_kinetics_400/" \ 
--batch-size=16 --test-only --cache-dataset \
--clip-len 16 --frame-rate 8 --clips-per-video 5 \
--model mvit_v2_s --weights="MViT_V2_S_Weights.DEFAULT"
 * Clip Acc@1 72.914 Clip Acc@5 89.507
 * Video Acc@1 80.757 Video Acc@5 94.665
```

Note that the reporting Acc@1 is a bit lower than the one of the paper but this is due to the version of the dataset that we use to assess the model (some corrupted videos are removed). To ensure that the accuracy of TorchVision's implementation is not lagging, we are testing the same data and weights using [Slowfast](https://github.com/facebookresearch/SlowFast) reference scripts:
```
INFO:slowfast.utils.logging:json_stats: {"split": "test_final", "top1_acc": "80.79", "top5_acc": "94.66"}
```

As we can see the accuracies are practically the same, with minor differences caused by differences on the VideoClip sampling mechanism.


torchvision/models/video/mvit.py
```diff
@@ -177,24 +261,53 @@ def __init__(
                 norm_layer(self.head_dim),
             )
 
+        self.rel_pos_h: Optional[nn.Parameter] = None
+        self.rel_pos_w: Optional[nn.Parameter] = None
+        self.rel_pos_t: Optional[nn.Parameter] = None
+        if rel_pos:
+            if input_size[1] != input_size[2]:
+                raise ValueError("Relative Positional Embeddings require square input shape.")
+            size = input_size[1]
```
datumbox, line 270: @haooooooqi @lyttonhao @feichtenhofer I would love to get your input as authors of the paper on how one could remove this limitation from the implementation. Ideally we would like to support any arbitrary rectangle. Do you have an advice here? Perhaps `size = max(input_size[1:])` could work?

lyttonhao: Yeah,  using `max(input_size[1:])` could work or calculating the size for h and w separately should also work.

datumbox: Thanks a lot for the prompt reply. I'll make the necessary change.

torchvision/models/video/mvit.py
```diff
@@ -106,28 +108,120 @@ def forward(self, x: torch.Tensor, thw: Tuple[int, int, int]) -> Tuple[torch.Ten
...
+
+    return attn
+
+
+def _add_shortcut(x: torch.Tensor, shortcut: torch.Tensor, skip_cls: bool):
+    if skip_cls:
+        x[:, :, 1:, :] += shortcut[:, :, 1:, :]
+    else:
+        x.add_(shortcut)
+    return x
```
datumbox, line 191: @lyttonhao @haooooooqi @feichtenhofer I got one more question for you concerning how the shortcut in the attention is added. I believe there is a difference between the two reference implementations of SlowFast and PyTorch Video:
- PTV seems to [straight add](https://github.com/facebookresearch/pytorchvideo/blob/718d0a46a8ec3cfd12b823a2063ab66b006281b2/pytorchvideo/layers/attention.py#L488) q in the outcome of attn @ v. 
- Slowfast seems to [condition](https://github.com/facebookresearch/SlowFast/blob/1aebd71a2efad823d52b827a3deaf15a56cf4932/slowfast/models/attention.py#L432-L435) how this addition will happen based on the value of `has_cls_embed`. 

Both implementations support `has_cls_embed`, but each handles it differently (note that in TorchVision this value is always True, so it's omitted). My questions to you are:
- Why is there a difference between the two implementations?
- Do you think the way I handle the difference here makes sense?

datumbox:
Script to port the weights from SlowFast and to confirm it yields the same results:
```python
import collections

import torch
from torchvision.models.video import mvit_v2_s


def slowfast_to_tv_weights(state_dict):
    d = dict(state_dict)

    # remapping keys
    mapping = collections.OrderedDict(
        [
            ("patch_embed.project.weight", "conv_proj.weight"),
            ("patch_embed.project.bias", "conv_proj.bias"),
            ("cls_token", "pos_encoding.class_token"),
            ("pos_embed_spatial", "pos_encoding.spatial_pos"),
            ("pos_embed_temporal", "pos_encoding.temporal_pos"),
            ("pos_embed_class", "pos_encoding.class_pos"),
            ("attn.proj.weight", "attn.project.0.weight"),
            ("attn.proj.bias", "attn.project.0.bias"),
            ("attn.pool_q.weight", "attn.pool_q.pool.weight"),
            ("attn.norm_q.weight", "attn.pool_q.norm_act.0.weight"),
            ("attn.norm_q.bias", "attn.pool_q.norm_act.0.bias"),
            ("attn.pool_k.weight", "attn.pool_k.pool.weight"),
            ("attn.norm_k.weight", "attn.pool_k.norm_act.0.weight"),
            ("attn.norm_k.bias", "attn.pool_k.norm_act.0.bias"),
            ("attn.pool_v.weight", "attn.pool_v.pool.weight"),
            ("attn.norm_v.weight", "attn.pool_v.norm_act.0.weight"),
            ("attn.norm_v.bias", "attn.pool_v.norm_act.0.bias"),
            ("mlp.fc1.weight", "mlp.0.weight"),
            ("mlp.fc1.bias", "mlp.0.bias"),
            ("mlp.fc2.weight", "mlp.3.weight"),
            ("mlp.fc2.bias", "mlp.3.bias"),
            ("head.projection.weight", "head.1.weight"),
            ("head.projection.bias", "head.1.bias"),
            ("patch_embed.proj.weight", "conv_proj.weight"),
            ("patch_embed.proj.bias", "conv_proj.bias"),
            ("norm.weight", "norm.weight"),
            ("norm.bias", "norm.bias"),
            ("proj.weight", "project.weight"),
            ("proj.bias", "project.bias"),
        ]
    )
    for k in list(d.keys()):
        for pattern, replacement in mapping.items():
            if pattern in k:
                new_key = k.replace(pattern, replacement)
                d[new_key] = d.pop(k)
                break

    # matching dimensions
    d["pos_encoding.class_token"] = d["pos_encoding.class_token"][0, 0, :]
    if "pos_encoding.spatial_pos" in d:
        d["pos_encoding.spatial_pos"] = d["pos_encoding.spatial_pos"][0, :]
        d["pos_encoding.temporal_pos"] = d["pos_encoding.temporal_pos"][0, :]
        d["pos_encoding.class_pos"] = d["pos_encoding.class_pos"][0, 0, :]

    return d


#path = "K400_MVIT_B_16x4_CONV.pyth"  # Needs patching on the model config with `kernel_kv[-2:]=(1, 1, 1)`
path = "MViTv2_S_16x4_k400_f302660347.pyth"
s = torch.load(path, map_location="cpu")
d = s["model_state"]

d = slowfast_to_tv_weights(d)

model = mvit_v2_s()
model.load_state_dict(d)
print("OK")

torch.manual_seed(42)
model.eval()
y = model(torch.randn((1, 3, 16, 224, 224)))
print(y)
print(y.sum())
```

torchvision/models/video/mvit.py
lyttonhao: I think they are just two implementations while the SlowFast one is our original implementation. Regarding the performance difference, our ablation shows the results are very close. So I think as long as the training/testing are consistent, this should not be an issue.

The `skip_cls` implementation looks good to me!

datumbox: Thanks for letting me know! 

Do you think it's OK to populating `skip_cls` with the value of `rel_pos` makes sense or we should introduce another parameter for the Attention layer? 
https://github.com/pytorch/vision/blob/be6d7e080a9b22de59089efd8268f99c19dfa5c9/torchvision/models/video/mvit.py#L318

> So I think as long as the training/testing are consistent, this should not be an issue.

Do you happen to have pre-trained weights that don't require this. If yes, I would rather use them instead and avoid the extra parameter.

datumbox: I had an offline discussion with @lyttonhao where he suggested that the `skip_cls` should be independent from the `rel_pos`. Unfortunately this difference on the implementations between SlowFast and PyTorchVideo means that we will need an extra config to control for it. Based on his feedback, I introduced an extra parameter called `residual_cls`. If True, the addition in the residual connection includes the class embeddings. If False, the class embeddings are excluded in the addition. This parameter is exposed in the constructor of MViT.

torchvision/models/video/mvit.py
```diff
@@ -177,24 +273,52 @@ def __init__(
                 norm_layer(self.head_dim),
             )
 
+        self.rel_pos_h: Optional[nn.Parameter] = None
```
jdsgomes, line 276: nit: I would suggest to move this whole block to a separate method to make the init less lengthy and more readable.

torchvision/models/video/mvit.py
```diff
@@ -303,12 +453,18 @@ def __init__(
             temporal_size (int): The temporal size ``T`` of the input.
             block_setting (sequence of MSBlockConfig): The Network structure.
             residual_pool (bool): If True, use MViTv2 pooling residual connection.
+            residual_cls (bool): If True, the addition on the residual connection will include the class embedding.
+            rel_pos (bool): If True, use MViTv2's relative positional embeddings.
```
jdsgomes, line 457: NIt: `rel_pos_embed` possibly more clear, also `residual_cls` could be renamed to something like `residual_with_cls_embed`.

torchvision/models/video/mvit.py
datumbox: Yes you are right this is quite messy. I'm happy to make changes to make it more readable. Concerning moving this on separate method, here are few concerns:
1. It can be invoked by users after FX and the initialization wont work any more. This is why I moved previously all inits in the constructors.
2. The idiom can in principle lead to weird inheritance issues as the object is not instantiated fully yet. This won't be the problem in this specific case, so it's more of an argument of why I try to avoid the idiom.

What do you think?

torchvision/models/video/mvit.py
datumbox: Sounds good. I'll make the change.

torchvision/models/video/mvit.py
jdsgomes: I agree that it does become problematic in some non standard usage patterns. This was just a nit, so  given your explanation the PR is good to go from my side.

jdsgomes (commented):
LGTM, thanks

torchvision/models/video/mvit.py
datumbox: Thanks a lot for the flexibility.

jdsgomes (approved):

PR #6376, opened 2022-08-06
Type fix in transformers.py

victorcwai:
Update `RandomPhotometricDistort` `__init__` argument to correct types.


datumbox (approved):
LGTM, thanks!

PR #6380, opened 2022-08-08
Make test precision stricter for Classification

datumbox:
Currently we are using extremely low precision thresholds on some of our classification tests (image, quantized and video). Especially on video, the value is so high that it doesn't capture breaking changes on the code. This PR tries to update the values to stricter thesholds without causing flakiness.

### DIFF
--- a/test/test_models.py
+++ b/test/test_models.py
@@ -884,7 +884,7 @@ def test_quantized_classification_model(model_fn):
     out = model(x)
 
     if model_name not in quantized_flaky_models:
-        _assert_expected(out, model_name + "_quantized", prec=1e-3)
+        _assert_expected(out, model_name + "_quantized", prec=2e-2)
         assert out.shape[-1] == 5
         _check_jit_scriptable(model, (x,), unwrapper=script_model_unwrapper.get(model_name, None), eager_out=out)
         _check_fx_compatible(model, x, eager_out=out)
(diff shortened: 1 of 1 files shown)


jdsgomes (approved):
LGTM, thanks

test/test_models.py
```diff
@@ -841,7 +841,7 @@ def test_video_model(model_fn, dev):
     # RNG always on CPU, to ensure x in cuda tests is bitwise identical to x in cpu tests
     x = torch.rand(input_shape).to(device=dev)
     out = model(x)
-    _assert_expected(out.cpu(), model_name, prec=0.1)
+    _assert_expected(out.cpu(), model_name, prec=1e-5)
```
NicolasHug, line 844: FYI @datumbox I am suspecting that this changing might cause flakyness in our internal tests

D38824237 shows that pytorch_vision_gpu-buck_2 is failing: https://www.internalfb.com/intern/testinfra/diagnostics/844425187765605.562950021904293.1660819094/

I don't see that on the other diffs (yet?) so I assume it's flaky.

datumbox: Thanks for the heads up. I would recommend turning off the internal test and relying on Github for verifying the models. Historically the `test_models.py` at Meta has been a source of flakyness, so feel free to turn off anything that breaks.

jdsgomes: Looking into it, could be that the expected files are wrong.

PR #6383, opened 2022-08-09
[DO NOT MERGE] LibTorchvision 0.13 binary push

jmdetloff:


datumbox:
@jmdetloff Thanks for the PR. Do we need to send this against the `release/0.13` branch instead of the `main`?

jmdetloff:
I could be wrong, but I don't think so. This PR is not intended to be merged into any branch, we're just using the CI workflow to upload the build product to our aws s3 bucket. This is the demo PR referenced from the pod release runbook https://github.com/pytorch/vision/pull/4717 and it's a pr on main as well.

.circleci/unittest/ios/scripts/binary_ios_build.sh
```diff
@@ -22,6 +22,8 @@ export CMAKE_PREFIX_PATH=${CONDA_PREFIX:-"$(dirname $(which conda))/../"}
 
 # sync submodules
 cd ${PROJ_ROOT_IOS}
+git fetch
+git checkout release/0.13
```
datumbox, line 26: I missed this. I think you are right, doesn't matter to which branch you will send the PR. Thanks for clarifying.

PR #6385, opened 2022-08-09
[proto] Fix for handling numpy arrays by Transform

vfdev-5:
Fixing the following TODO:
```
# TODO: Transforms allows to pass only (torch.Tensor, _Feature, PIL.Image.Image)
# so input as np.ndarray is not possible. We need to make it possible
```
for transforms: ToImagePIL, ToImageTensor and others

### DIFF
--- a/test/test_prototype_transforms.py
+++ b/test/test_prototype_transforms.py
@@ -1,5 +1,7 @@
 import itertools
 
+import numpy as np
+
 import PIL.Image
 
 import pytest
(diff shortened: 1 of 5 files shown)


torchvision/prototype/transforms/_transform.py
```diff
@@ -10,6 +10,10 @@
 
 
 class Transform(nn.Module):
+
+    # Class attribute defining transformed types. Other types are passed-through without any transformation
+    transformed_types: Tuple[Type, ...] = (torch.Tensor, _Feature, PIL.Image.Image)
```
vfdev-5, line 15: Does this approach works for you @datumbox ?

datumbox: It does. Can we make it private? I'm not sure we should expose it upfront. We can always revise it and make it public but right now it looks like an implementation detail.

vfdev-5: Yes, you are right, I'll update that

datumbox (approved):
LGTM, thanks!

test/test_prototype_transforms.py
```diff
@@ -991,3 +993,94 @@ def test__transform(self, p, inpt_type, mocker):
...
+        fn = mocker.patch(
+            "torchvision.prototype.transforms.functional.to_image_tensor",
+            return_value=torch.rand(1, 3, 8, 8),
+        )
+
+        inpt = mocker.MagicMock(spec=inpt_type)
+        transform = transforms.ToImageTensor()
+        transform(inpt)
+        if inpt_type in (features.BoundingBox, str, int):
+            fn.call_count == 0
```
NicolasHug, line 1031: @vfdev-5 @datumbox 
Our internal linter pointed that this line (and various similar other lines) is probably missing an `assert`

datumbox: Thanks for the heads up. @vfdev-5 could you please send a PR that fixes it?

vfdev-5: Hopefully computers check what we are doing here :)
Sure, I'll fix that asap

PR #6386, opened 2022-08-09
[proto] Fixed issue with `F.pad` from RandomZoomOut

vfdev-5:
- Fixed issue with `F.pad` if fill is a vector
- Fixed issue with default value for PIL branch. If call `F.pad(image_pil, padding)` without any `fill` arg, method complains about unsupported `None` value.

### DIFF
--- a/torchvision/prototype/transforms/functional/_geometry.py
+++ b/torchvision/prototype/transforms/functional/_geometry.py
@@ -608,9 +608,6 @@ def pad(
     if isinstance(inpt, features._Feature):
         return inpt.pad(padding, fill=fill, padding_mode=padding_mode)
     elif isinstance(inpt, PIL.Image.Image):
-        if fill is None:
-            fill = 0
-
         return pad_image_pil(inpt, padding, fill=fill, padding_mode=padding_mode)
     else:
         # This cast does Sequence[int] -> List[int] and is required to make mypy happy
--- a/torchvision/prototype/transforms/functional/_geometry.py
+++ b/torchvision/prototype/transforms/functional/_geometry.py
@@ -555,7 +555,7 @@ def _pad_with_vector_fill(
         raise ValueError(f"Padding mode '{padding_mode}' is not supported if fill is not scalar")
 
     output = pad_image_tensor(img, padding, fill=0, padding_mode="constant")
-    left, top, right, bottom = _FT._parse_pad_padding(padding)
+    left, right, top, bottom = _FT._parse_pad_padding(padding)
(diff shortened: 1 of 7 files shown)


torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -555,7 +555,7 @@ def _pad_with_vector_fill(
         raise ValueError(f"Padding mode '{padding_mode}' is not supported if fill is not scalar")
 
     output = pad_image_tensor(img, padding, fill=0, padding_mode="constant")
-    left, top, right, bottom = _FT._parse_pad_padding(padding)
+    left, right, top, bottom = _FT._parse_pad_padding(padding)
```
datumbox, line 558: Good catch!

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -608,6 +608,9 @@ def pad(
     if isinstance(inpt, features._Feature):
         return inpt.pad(padding, fill=fill, padding_mode=padding_mode)
     elif isinstance(inpt, PIL.Image.Image):
+        if fill is None:
+            fill = 0
+
```
datumbox, line 613: Is this needed? PIL's pad signature seems to suggest that is handles None:
https://github.com/pytorch/vision/blob/51f27d766846e8347d4b56b95434f02950fe9e2b/torchvision/transforms/functional_pil.py#L146-L149

datumbox (commented):
Thanks @vfdev-5. It looks good just one question.

Edit: One of your tests is [failing](https://github.com/pytorch/vision/runs/7748386206?check_suite_focus=true). Seems unrelated so I restarted it to see if it's flaky.

torchvision/prototype/transforms/functional/_geometry.py
vfdev-5: Actually, this type hint is wrong due to the internal type check: 
https://github.com/pytorch/vision/blob/2e70ee1af3bb3c5ba0f9a8bcf16cd1611831690f/torchvision/transforms/functional_pil.py#L158-L159
so, fill can't be `None`

datumbox: Do you think we should be handling it inside the pad method? Aka assign 0 to it.

vfdev-5: Yes, we can. Let me do that

datumbox (approved):
LGTM, thanks!

PR #6391, opened 2022-08-10
[proto] Compose transform keeps BC

vfdev-5:
- input transforms should be an iterable to keep BC
- Removed subclassing from `Transform` such that transforms could be any non-nn-Module functions

### DIFF
--- a/torchvision/prototype/transforms/_container.py
+++ b/torchvision/prototype/transforms/_container.py
@@ -2,19 +2,18 @@ from typing import Any, Callable, Dict, List, Optional, Sequence
 
 import torch
 from torchvision.prototype.transforms import Transform
-from torchvision.utils import _log_api_usage_once
 
 from ._transform import _RandomApplyTransform
 
 
-class Compose:
+class Compose(Transform):
     def __init__(self, transforms: Sequence[Callable]) -> None:
-        _log_api_usage_once(self)
+        super().__init__()
         if not isinstance(transforms, Sequence):
             raise TypeError("Argument transforms should be a sequence of callables")
         self.transforms = transforms
 
-    def __call__(self, *inputs: Any) -> Any:
+    def forward(self, *inputs: Any) -> Any:
         sample = inputs if len(inputs) > 1 else inputs[0]
         for transform in self.transforms:
             sample = transform(sample)
--- a/torchvision/prototype/transforms/_container.py
+++ b/torchvision/prototype/transforms/_container.py
@@ -1,19 +1,20 @@
-from typing import Any, Dict, List, Optional
+from typing import Any, Callable, Dict, List, Optional, Sequence
 
 import torch
 from torchvision.prototype.transforms import Transform
+from torchvision.utils import _log_api_usage_once
 
 from ._transform import _RandomApplyTransform
 
 
-class Compose(Transform):
-    def __init__(self, *transforms: Transform) -> None:
-        super().__init__()
+class Compose:
+    def __init__(self, transforms: Sequence[Callable]) -> None:
+        _log_api_usage_once(self)
+        if not isinstance(transforms, Sequence):
+            raise TypeError("Argument transforms should be a sequence of callables")
         self.transforms = transforms
-        for idx, transform in enumerate(transforms):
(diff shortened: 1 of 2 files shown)


torchvision/prototype/transforms/_container.py
```diff
@@ -1,19 +1,20 @@
...
 from torchvision.prototype.transforms import Transform
+from torchvision.utils import _log_api_usage_once
 
 from ._transform import _RandomApplyTransform
 
 
-class Compose(Transform):
-    def __init__(self, *transforms: Transform) -> None:
-        super().__init__()
+class Compose:
```
datumbox, line 10: Can we at least define this as an `nn.Module` and structure it with a forward?

vfdev-5: Yes, we can inherit from nn.Module without registering `transforms` as submodules otherwise passing just a callable wont work

datumbox: @pmeier Thoughts on this?

pmeier: `Transform` is already thin wrapper around `nn.Module` and if we overwrite the `forward` method without calling `super().forward` the only difference is the added `__repr__` as well as logging. Thus, if we can derive from `nn.Module`, we should be able to derive from `Transform`.

pmeier (commented):
I agree that the `Compose` in the current state is not fully BC. However, `lambda`'s or local functions will not supported by the dataloader:

```py
import torch
from torch import nn
from torchdata.datapipes.iter import IterableWrapper


class Compose:
    def __init__(self, *transforms):
        self.transforms = transforms

    def __call__(self, input):
        for transform in self.transforms:
            input = transform(input)
        return input


class Foo(nn.Module):
    def forward(self, input):
        return input + 1


dataset = IterableWrapper([torch.tensor(1)]).map(
    Compose(
        Foo(),
        lambda input: input * 2,
    )
)

traverse(dataset)
```

```
_pickle.PicklingError: Can't pickle <function <lambda> at 0x7fabc90d2b90>: attribute lookup <lambda> on __main__ failed
```

IMO, we can go full BC now, but should deprecate the non `nn.Module` usage.

torchvision/prototype/transforms/_container.py
datumbox: I also agree that deriving from `Transform` for all classes is preferable. The less exceptions the better IMO.

vfdev-5:
> However, lambda's or local functions will not supported by the dataloader

@pmeier local functions seem to be supported ?
```python
def foo(x):
    return x * 2
    

dataset = IterableWrapper([torch.tensor(1)]).map(
    Compose(
        Foo(),
        foo,
    )
)
traverse(dataset)
> {MapperIterDataPipe: {IterableWrapperIterDataPipe: {}}}
```

pmeier:
> local functions seem to be supported ?

Your example is not a local function. For example:

```py
import torch
from torchdata.datapipes.iter import IterableWrapper


def main():
    def foo(x):
        return x * 2

    dataset = IterableWrapper([torch.tensor(1)]).map(foo)

    traverse(dataset)

main()
```

```
AttributeError: Can't pickle local object 'main.<locals>.foo'
```

> As for going to `Compose` subclassing `nn.Module`, what would be the benefit of that ?

I agree with @datumbox in https://github.com/pytorch/vision/pull/6391#discussion_r943367773.

torchvision/prototype/transforms/_container.py
vfdev-5: What is the benefit of deriving `Transform` from `nn.Module` ?

datumbox: For me it's weird we propose a new API and `Compose` is this weird thing that doesn't follow it. IMO inheriting from `Transform` and overwriting `forward()` is something perfectly aligned with the new API. It's something we did discuss as possibility during designs.

From the comments in this PR I think it might be more beneficial to sync with Philip offline and align. I can join you later if that's needed and we can summarize the decision here afterwards.

datumbox (approved):
LGTM, thanks!

PR #6397, opened 2022-08-11
Fix typing jit issue on RoIPool and RoIAlign

### LINKED ISSUE #6395: Inconsistency between RoIPool's forward and roi_pool
### 🚀 The feature

Currently the functional `roi_pool` method supports both `Tensor` and `List[Tensor]` for boxes, but the module forward function of `RoIPool` expects boxes(rois) to be `Tensor` only. Can we change the function signature to support both? 

The inconsistency can cause issue during torchscripting, for example:
```
from typing import List
import torch
from torch import Tensor
from torchvision.ops import RoIPool

class Model(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.pool = RoIPool(output_size=[3,3], spatial_scale=1.0)

    def forward(self, imgs, boxes: List[Tensor]):
        return self.pool(imgs, boxes)

model = Model()
torch.jit.script(model)
```
Raise the following error:

> forward(__torch__.torchvision.ops.roi_pool.RoIPool self, Tensor input, Tensor rois) -> Tensor:
> Expected a value of type 'Tensor' for argument 'rois' but instead found type 'List[Tensor]'.
> Empty lists default to List[Tensor]. Add a variable annotation to the assignment to create an empty list of another type (torch.jit.annotate(List[T, []]) where T is the type of elements in the list for Python 2)
> :
>   File "<ipython-input-8-9536dd2f0cc1>", line 12
>     def forward(self, imgs, boxes: List[Tensor]):
>         return self.pool(imgs, boxes)
>                ~~~~~~~~~ <--- HERE

### Motivation, pitch

Make sure the model can be scriptable when using list of boxes as the input.

### Alternatives

Only support Tensor as the input, user are required to convert bbox of different images into one tensor. Potentially make `_utils.convert_boxes_to_roi_format` public so that users don't need to write the conversion function.

### Additional context

_No response_

datumbox:
Fixes #6395

### DIFF
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -69,6 +69,15 @@ class DropBlockWrapper(nn.Module):
         self.layer(a)
 
 
+class PoolWrapper(torch.nn.Module):
+    def __init__(self, pool: nn.Module):
+        super().__init__()
+        self.pool = pool
+
+    def forward(self, imgs: Tensor, boxes: List[Tensor]):
+        return self.pool(imgs, boxes)
+
+
 class RoIOpTester(ABC):
     dtype = torch.float64
 
@@ -209,6 +218,10 @@ class TestRoiPool(RoIOpTester):
 
     def test_boxes_shape(self):
         self._helper_boxes_shape(ops.roi_pool)
+
+    def test_jit_boxes_list(self):
+        model = PoolWrapper(ops.RoIPool(output_size=[3, 3], spatial_scale=1.0))
+        torch.jit.script(model)
 
 
 class TestPSRoIPool(RoIOpTester):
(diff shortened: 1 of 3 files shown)


test/test_ops.py
```diff
@@ -210,6 +219,10 @@ def get_slice(k, block):
     def test_boxes_shape(self):
         self._helper_boxes_shape(ops.roi_pool)
 
+    def test_jit_boxes_list(self):
+        model = PoolWrapper(ops.RoIPool(output_size=[3, 3], spatial_scale=1.0))
+        torch.jit.script(model)
```
vfdev-5, line 224: Maybe we could also check model execution on list of tensors ?

vfdev-5 (approved):
LGTM. 
I added a comment to discuss

test/test_ops.py
datumbox: Yes good call. I'll add it now.

vfdev-5 (approved):
Thanks !

PR #6398, opened 2022-08-11
extract kernels into separate module to avoid cyclic imports

pmeier:
This is my suggestion on how to solve the cyclic dependencies in the methods of the features:

https://github.com/pytorch/vision/blob/cc0a83856c979f834132a5faf71a2eece0d32dc3/torchvision/prototype/features/_image.py#L112-L116

The gist is that we factor out the kernels from `torchvision.prototype.transforms.functional` into `torchvision.prototype.kernels`.

### Cons

- The kernels and the dispatchers are now in separate modules. Given that our dispatchers currently are very thin wrappers

    https://github.com/pytorch/vision/blob/cc0a83856c979f834132a5faf71a2eece0d32dc3/torchvision/prototype/transforms/functional/_geometry.py#L50-L56

    this makes the structure of our source a little more complex.

### Pros

- The kernels and the dispatchers are now in separate modules. Given that kernels and dispatchers serving different purposes, a clean separation between the two can help reduce confusion. This was the functional module stays the same as before and fully BC and all new functionalities (or rather newly exposed functionality since the kernels were already there) will have a clean new namespace
- The cyclic dependencies are resolved and in turn the dynamic imports are no longer needed

For demonstration purposes I only ported the `horizontal_flip_*` kernels and everything that was needed for them to the proposed architecture.

### DIFF
--- a/torchvision/prototype/constants.py
+++ b/torchvision/prototype/constants.py
@@ -1,0 +1,34 @@
+# Maybe a good place to put all enums that are relevant to more than module?
+
+from __future__ import annotations
+
+from torchvision._utils import StrEnum
+
+__all__ = ["BoundingBoxFormat", "ColorSpace"]
+
+
+class BoundingBoxFormat(StrEnum):
+    XYXY = StrEnum.auto()
+    XYWH = StrEnum.auto()
+    CXCYWH = StrEnum.auto()
+
+
+class ColorSpace(StrEnum):
+    OTHER = StrEnum.auto()
+    GRAY = StrEnum.auto()
+    GRAY_ALPHA = StrEnum.auto()
+    RGB = StrEnum.auto()
+    RGB_ALPHA = StrEnum.auto()
+
+    @classmethod
+    def from_pil_mode(cls, mode: str) -> ColorSpace:
+        if mode == "L":
+            return cls.GRAY
+        elif mode == "LA":
+            return cls.GRAY_ALPHA
+        elif mode == "RGB":
+            return cls.RGB
+        elif mode == "RGBA":
+            return cls.RGB_ALPHA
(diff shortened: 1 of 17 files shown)


torchvision/prototype/constants.py
```diff
@@ -0,0 +1,34 @@
+# Maybe a good place to put all enums that are relevant to more than module?
```
pmeier, line 1: We need these enums in features and transforms. Thus, we can't put them in these modules if we want to avoid cyclic dependencies. They could go into kernels, but IMO they are misplaced there. This is why I created a separate module. We can also expose them under `torchvision.prototype` (or `torchvision`) later to make it more convenient for users.

torchvision/prototype/features/_bounding_box.py
```diff
@@ -62,21 +58,15 @@ def new_like(
     def to_format(self, format: Union[str, BoundingBoxFormat]) -> BoundingBox:
         # TODO: this is useful for developing and debugging but we should remove or at least revisit this before we
         #  promote this out of the prototype state
-
```
pmeier, line 65: This is the core of what this PR is trying to resolve. The same happens for the `BoundingBox` and `SegmentationMask` feature

vfdev-5:
@pmeier can we close this PR ?

pmeier:
Yes. This was "solved" by #6476. Still, https://github.com/pytorch/vision/pull/6398#discussion_r943251138 remains valid and we should address this in the future.

datumbox:
I agree about the validity of https://github.com/pytorch/vision/pull/6398#discussion_r943251138. The work at https://github.com/pytorch/vision/pull/6476 was mostly a way to get around it.

PR #6401, opened 2022-08-11
[proto] Ported RandomIoUCrop from detection refs

vfdev-5:
- [x] Ported RandomIoUCrop transform
- [x] Check visually
- [x] Tests to add


torchvision/prototype/transforms/_geometry.py
```diff
@@ -611,3 +613,94 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
...
+
+                return dict(top=top, left=left, height=new_h, width=new_w, is_within_crop_area=is_within_crop_area)
+
+    def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
+        if len(params) < 1:
+            return inpt
+
+        is_within_crop_area = params["is_within_crop_area"]
+        if isinstance(inpt, features.Label):
+            return features.Label.new_like(inpt, inpt[is_within_crop_area])
```
pmeier, line 697: This shows one of the issues I highlighted in the design doc: here we assume that every `Label` we have in the pytree is "linked" to the bounding boxes, i.e. has the same batch dimensions. What if have another label in there for the whole image or this assumption is otherwise violated, e.g. the labels are not a batched tensor but rather a list of labels?

In our detection references we can make this assumption, because we control the input. I have no idea how we want to handle this in a general way and this is why I haven't ported the remaining detection transforms yet.

datumbox: I think you can safely make the assumption that the labels are linked to the bboxes, for all the transforms under references/detection. Would that be enough to unblock you?

torchvision/prototype/transforms/_geometry.py
```diff
@@ -611,3 +613,102 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
...
+            output = features.BoundingBox.new_like(output, bboxes)
+
+        return output
+
+    def forward(self, *inputs: Any) -> Any:
+        sample = inputs if len(inputs) > 1 else inputs[0]
+        if not has_all(sample, features.Image, features.BoundingBox, features.Label):
+            raise TypeError(f"{type(self).__name__}() is only defined for Images, BoundingBoxes and Labels.")
+        if has_any(sample, features.OneHotLabel):
+            raise TypeError(f"{type(self).__name__}() does not support OneHotLabels.")
```
vfdev-5, line 713: @datumbox you were previously saying about filtering input data types. How about this one ?

datumbox: Not 100% sure why we can't support it or whether we should. Perhaps OneHotLabel should inherit from Label so that we don't have to check both of them in `has_all()` calls.

torchvision/prototype/transforms/_geometry.py
```diff
@@ -611,3 +613,102 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
...
+            bboxes = output[is_within_crop_area]
+            bboxes = F.clamp_bounding_box(bboxes, output.format, output.image_size)
+            output = features.BoundingBox.new_like(output, bboxes)
+
+        return output
+
+    def forward(self, *inputs: Any) -> Any:
+        sample = inputs if len(inputs) > 1 else inputs[0]
+        if not has_all(sample, features.Image, features.BoundingBox, features.Label):
+            raise TypeError(f"{type(self).__name__}() is only defined for Images, BoundingBoxes and Labels.")
```
pmeier, line 711: Do we want to enforce labels here? This transform works fine if we don't have labels, right?

torchvision/prototype/transforms/functional/_meta.py
```diff
@@ -61,6 +61,15 @@ def convert_bounding_box_format(
     return bounding_box
 
 
+def clamp_bounding_box(
```
pmeier, line 64: Can we maybe also make this a method on `features.BoundingBox`? I'm aware that `clamp` is already used and so is `clip`. Maybe we can call it `trim`?

torchvision/prototype/transforms/_geometry.py
datumbox: We do. Object Detection needs not only BBoxes but also Labels to work.

torchvision/prototype/transforms/functional/_meta.py
vfdev-5: We may try to override existing `Tensor.clamp` ?

datumbox: > We may try to override existing Tensor.clamp ?

I would avoid overwriting default tensor methods. Even deprecated ones might be a problem.

vfdev-5: Why do you think it could be harmful ?

datumbox: I know we do this more on the resize which is deprecated but as we discussed in that case it might surprise users. We should probably ask for feedback from the community on the topic. From one hand, subclassing Tensor means we can overwrite methods to make the more "relevant" to the specific type. On the other hand, it might confuse users who expect that they are accessing the parent method. To avoid blocking this PR, I think we can go ahead without a method on the BBox object and raise this as an open question on the API design.

torchvision/prototype/transforms/_geometry.py
```diff
@@ -620,6 +622,108 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
...
+                    continue
+
+                return dict(top=top, left=left, height=new_h, width=new_w, is_within_crop_area=is_within_crop_area)
+
+    def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
+        if len(params) < 1:
+            return inpt
+
+        is_within_crop_area = params["is_within_crop_area"]
+        if isinstance(inpt, features.Label):
```
datumbox, line 703: Should we be supporting here OneHotLabels too explicitly until #6420 is resolved?

vfdev-5: Yes, we should. I have to understand how it would work with OneHotLabels in this case...
I do not plan to continue on #6420 right now

torchvision/prototype/transforms/_geometry.py
```diff
@@ -620,6 +622,108 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
...
+            return features.Label.new_like(inpt, inpt[is_within_crop_area])
+
+        output = F.crop(inpt, top=params["top"], left=params["left"], height=params["height"], width=params["width"])
+
+        if isinstance(output, features.BoundingBox):
+            bboxes = output[is_within_crop_area]
+            bboxes = F.clamp_bounding_box(bboxes, output.format, output.image_size)
```
datumbox, line 710: It feels that this clamping operation should have been also done on all other Crop ops. Do you plan to add them in this PR or on separate one?

torchvision/prototype/transforms/_geometry.py
```diff
@@ -620,6 +622,113 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
...
+    def forward(self, *inputs: Any) -> Any:
+        sample = inputs if len(inputs) > 1 else inputs[0]
+        if not (
+            has_all(sample, features.BoundingBox)
+            and has_any(sample, PIL.Image.Image, features.Image)
+            and has_any(sample, features.Label, features.OneHotLabel)
+        ):
+            raise TypeError(
+                f"{type(self).__name__}() is only defined for Images, PIL Images, "
+                "BoundingBoxes and Labels or OneHotLabels."
```
datumbox, line 727: We should also add a quick support of Masks. The previous operator didn't support it on references but now we can do this. We just need to crop it exactly how we do the image and we should be OK.

vfdev-5: Masks are implicitly supported. I added a test case

datumbox: Can you update the error message?

datumbox (approved):
LGTM.

@pmeier any thoughts?

torchvision/prototype/transforms/_utils.py
```diff
@@ -17,6 +17,15 @@ def query_image(sample: Any) -> Union[PIL.Image.Image, torch.Tensor, features.Im
     raise TypeError("No image was found in the sample")
 
 
+def query_bboxes(sample: Any) -> features.BoundingBox:
```
pmeier, line 20: 1. We are not querying boxes, since we only return one.
2. Let's keep the naming scheme consistent with the kernels.

```suggestion
def query_bounding_box(sample: Any) -> features.BoundingBox:
```

torchvision/prototype/transforms/_utils.py
```diff
@@ -17,6 +17,15 @@ def query_image(sample: Any) -> Union[PIL.Image.Image, torch.Tensor, features.Im
...
+    flat_sample, _ = tree_flatten(sample)
+    for i in flat_sample:
+        if isinstance(i, features.BoundingBox):
+            return i
+
+    raise TypeError("No bounding boxes were found in the sample")
```
pmeier, line 26:
```suggestion
    raise TypeError("No bounding box was found in the sample")
```

pmeier (commented):
Small naming nit for consistency. Otherwise LGTM.

torchvision/prototype/transforms/_geometry.py
```diff
@@ -620,6 +622,115 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
...
+        output = F.crop(inpt, top=params["top"], left=params["left"], height=params["height"], width=params["width"])
+
+        if isinstance(output, features.BoundingBox):
+            bboxes = output[is_within_crop_area]
+            bboxes = F.clamp_bounding_box(bboxes, output.format, output.image_size)
+            output = features.BoundingBox.new_like(output, bboxes)
+        elif isinstance(output, features.SegmentationMask) and output.shape[-3] > 1:
+            # apply is_within_crop_area if mask is one-hot encoded
+            masks = output[is_within_crop_area]
+            output = features.SegmentationMask.new_like(output, masks)
```
vfdev-5, line 716: @datumbox here is a support for one-hot encoded masks meanwhile other solutions we could decide about

torchvision/prototype/transforms/_geometry.py
```diff
@@ -620,6 +622,115 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
...
+
+        return output
+
+    def forward(self, *inputs: Any) -> Any:
+        sample = inputs if len(inputs) > 1 else inputs[0]
+        if not (
+            has_all(sample, features.BoundingBox)
+            and has_any(sample, PIL.Image.Image, features.Image)
```
pmeier, line 724: What about plain tensors? For `CutMix` and `MixUp` we don't allow the "old" image types at all:

https://github.com/pytorch/vision/blob/162267ca90a8a7d03a8d9249602d6c72b497656a/torchvision/prototype/transforms/_augment.py#L108-L109

Given that have ported this from references there is no BC constraint to allow plain tensors and PIL images. Still, it feels unnecessary restrictive.

torchvision/prototype/transforms/_geometry.py
```diff
@@ -620,6 +622,115 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
...
+            and has_any(sample, features.Label, features.OneHotLabel)
+        ):
+            raise TypeError(
+                f"{type(self).__name__}() requires input sample to contain Images or PIL Images, "
+                "BoundingBoxes and Labels or OneHotLabels. Sample can also contain Segmentation Masks."
+            )
+        return super().forward(*inputs)
```
pmeier, line 731:
```suggestion
        return super().forward(sample)
```

torchvision/prototype/transforms/_geometry.py
vfdev-5: To understand your comment, you want to add support for images as Tensors and keep Image and PIL Image ?

pmeier: Either that or remove support for PIL here. We should either support all image types or only the "new one" like we do in `CutMix` and `MixUp` (not saying this is a good thing, but we should be consistent).

vfdev-5: Is the reason why CutMix, MixUp are not supporting PIL due to lack of implementation ?
Here in RandomIoUCrop we can support everything, so I would add image as torch.Tensor support and keep PIL.

datumbox: I think we should support all types. Part of the API supporting them and part not is weird and will hinter adoption.

PR #6407, opened 2022-08-12
Added padding_mode, fill and pad_both_sides options to F.crop

vfdev-5:
Description:
- In order to make prototype RandomCrop implementation simplier, let's enhance crop behaviour by supporting padding_mode, fill and pad_both_sides
- Tests needed

### DIFF
--- a/torchvision/transforms/functional.py
+++ b/torchvision/transforms/functional.py
@@ -522,7 +522,16 @@ def pad(img: Tensor, padding: List[int], fill: Union[int, float] = 0, padding_mo
     return F_t.pad(img, padding=padding, fill=fill, padding_mode=padding_mode)
 
 
-def crop(img: Tensor, top: int, left: int, height: int, width: int) -> Tensor:
+def crop(
+    img: Tensor,
+    top: int,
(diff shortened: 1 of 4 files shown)


pmeier:
We also need this for `FixedSizeCrop`:

https://github.com/pytorch/vision/blob/6e535db255cee3ce878dd7a54dda01d4ec8932c1/references/detection/transforms.py#L398-L401

The snippet above is just the padding that happens if the crop size is larger than the image size. However, `F.pad` is called with the additional arguments

https://github.com/pytorch/vision/blob/6e535db255cee3ce878dd7a54dda01d4ec8932c1/references/detection/transforms.py#L356

so we can't replace it with `F.crop` since we can't pass them on.

datumbox:
I don't think I would go with this solution. As discussed offline (just summarizing here for the community), the whole point is to write the Transform as a combination (chaining) of `F.pad()` and `F.crop()`. It's not to modify `crop()` to do more. 

I think the direction of #6410 is more aligned with what I had in mind and with the whole idea of the API.

PR #6408, opened 2022-08-12
Proto transform cleanup

pmeier:
This is cleanup for the prototype transforms. It includes the following changes that were put in individual commits for easier review:

1. Fix the `TenCrop` transform. The calls for PIL and simple tensor images were missing the `vertical_flip` flag.
2. Use the `F.adjust_{brightness, saturation, contrast, hue}` dispatchers in `RandomPhotometricDistort` rather than relying on the `ColorJitter` transform.
3. Add a `convert_color_space` dispatcher and use it in the `ConvertColorSpace` as well as in the deprecated `Grayscale` and `RandomGrayScale` transforms.

### DIFF
--- a/torchvision/prototype/features/_feature.py
+++ b/torchvision/prototype/features/_feature.py
@@ -1,10 +1,13 @@
 from __future__ import annotations
 
-from typing import Any, Callable, cast, List, Mapping, Optional, Sequence, Tuple, Type, TypeVar, Union
+from typing import Any, Callable, cast, List, Mapping, Optional, Sequence, Tuple, Type, TYPE_CHECKING, TypeVar, Union
 
 import torch
 from torch._C import _TensorBase, DisableTorchFunction
 from torchvision.transforms import InterpolationMode
+
+if TYPE_CHECKING:
+    from ._image import ColorSpace
 
 F = TypeVar("F", bound="_Feature")
 
@@ -201,4 +204,7 @@ class _Feature(torch.Tensor):
         return self
 
     def gaussian_blur(self, kernel_size: List[int], sigma: Optional[List[float]] = None) -> _Feature:
+        return self
+
+    def convert_color_space(self, color_space: "ColorSpace", *, copy: bool = True) -> _Feature:
         return self
--- a/torchvision/prototype/features/_feature.py
+++ b/torchvision/prototype/features/_feature.py
@@ -1,13 +1,10 @@
 from __future__ import annotations
 
-from typing import Any, Callable, cast, List, Mapping, Optional, Sequence, Tuple, Type, TYPE_CHECKING, TypeVar, Union
+from typing import Any, Callable, cast, List, Mapping, Optional, Sequence, Tuple, Type, TypeVar, Union
 
 import torch
 from torch._C import _TensorBase, DisableTorchFunction
 from torchvision.transforms import InterpolationMode
-
-if TYPE_CHECKING:
-    from ._image import ColorSpace
 
 F = TypeVar("F", bound="_Feature")
 
@@ -204,7 +201,4 @@ class _Feature(torch.Tensor):
         return self
 
     def gaussian_blur(self, kernel_size: List[int], sigma: Optional[List[float]] = None) -> _Feature:
-        return self
-
-    def convert_color_space(self, color_space: "ColorSpace", *, copy: bool = True) -> _Feature:
         return self
--- a/torchvision/prototype/features/_image.py
+++ b/torchvision/prototype/features/_image.py
@@ -99,6 +99,21 @@ class Image(_Feature):
         else:
             return ColorSpace.OTHER
 
+    def to_color_space(self, color_space: Union[str, ColorSpace], copy: bool = True) -> Image:
+        # import at runtime to avoid cyclic imports
+        from torchvision.prototype.transforms.functional import convert_color_space_image_tensor
+
+        if isinstance(color_space, str):
+            color_space = ColorSpace.from_str(color_space.upper())
+
+        return Image.new_like(
+            self,
+            convert_color_space_image_tensor(
+                self, old_color_space=self.color_space, new_color_space=color_space, copy=copy
+            ),
+            format=color_space,
+        )
+
     def show(self) -> None:
         # TODO: this is useful for developing and debugging but we should remove or at least revisit this before we
         #  promote this out of the prototype state
--- a/torchvision/prototype/features/_image.py
+++ b/torchvision/prototype/features/_image.py
@@ -111,7 +111,7 @@ class Image(_Feature):
             convert_color_space_image_tensor(
                 self, old_color_space=self.color_space, new_color_space=color_space, copy=copy
             ),
-            format=color_space,
+            color_space=color_space,
         )
 
     def show(self) -> None:
--- a/torchvision/prototype/features/_image.py
+++ b/torchvision/prototype/features/_image.py
@@ -100,15 +100,14 @@ class Image(_Feature):
             return ColorSpace.OTHER
 
     def to_color_space(self, color_space: Union[str, ColorSpace], copy: bool = True) -> Image:
-        # import at runtime to avoid cyclic imports
-        from torchvision.prototype.transforms.functional import convert_color_space_image_tensor
+        from torchvision.prototype.transforms import functional as _F
 
         if isinstance(color_space, str):
             color_space = ColorSpace.from_str(color_space.upper())
 
         return Image.new_like(
             self,
-            convert_color_space_image_tensor(
+            _F.convert_color_space_image_tensor(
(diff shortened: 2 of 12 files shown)


torchvision/prototype/features/_feature.py
```diff
@@ -1,11 +1,14 @@
...
 
-from typing import Any, Callable, cast, List, Mapping, Optional, Sequence, Tuple, Type, TypeVar, Union
+from typing import Any, Callable, cast, List, Mapping, Optional, Sequence, Tuple, Type, TYPE_CHECKING, TypeVar, Union
 
 import torch
 from torch._C import _TensorBase, DisableTorchFunction
 from torchvision.transforms import InterpolationMode
 
+if TYPE_CHECKING:
+    from ._image import ColorSpace
```
pmeier, line 10: We need this to avoid cyclic imports here. A better way would be to put all enums in a separate module from all others. For an example see #6398.

torchvision/prototype/transforms/_color.py
```diff
@@ -84,30 +84,6 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
         return output
 
 
-class _RandomChannelShuffle(Transform):
```
pmeier, line 87: Since this PR removes the last remnants of the former idiom to use primitive transforms in other transforms, I've moved this functionality into a method of `RandomPhotometricDistort`.

torchvision/prototype/transforms/functional/_meta.py
```diff
@@ -154,3 +154,21 @@ def convert_image_color_space_pil(
         return image
 
     return image.convert(new_mode)
+
+
+def convert_color_space(
```
pmeier, line 159: I think we botched the naming of the other kernels a little. It should be `convert_color_space_image_{tensor, pil}` rather than `convert_image_color_space_{tensor, pil}`. Naively following the old scheme would give us `convert_image_color_space` for the dispatcher, which is misleading since all inputs are allowed.

datumbox (commented):
Looks mostly OK @pmeier. A couple of comments to discuss but nothing too problematic. Let me know below to agree. 

@vfdev-5 Please have also a look.

Also let's always include all 3 of us in the reviews to make sure we remain aligned.

torchvision/prototype/transforms/functional/_meta.py
pmeier: Nothing, just wanted confirmation to do it.

torchvision/prototype/features/_feature.py
```diff
@@ -202,3 +205,6 @@ def invert(self) -> _Feature:
 
     def gaussian_blur(self, kernel_size: List[int], sigma: Optional[List[float]] = None) -> _Feature:
         return self
+
+    def convert_color_space(self, color_space: "ColorSpace", *, copy: bool = True) -> _Feature:
```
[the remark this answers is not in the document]
vfdev-5: Can't we just have this method on Image, `Image.convert_color_space` ?
I think we are not supposed to do something like:
```python
inpt: _Feature = ...
out = inpt.convert_color_space("RGB"
```
but only on images

torchvision/prototype/transforms/functional/_meta.py
datumbox: It's a good idea. Let's do it.

torchvision/prototype/features/_feature.py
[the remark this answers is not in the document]
pmeier: We have something like that for bounding boxes with regard to the format:

https://github.com/pytorch/vision/blob/6e535db255cee3ce878dd7a54dda01d4ec8932c1/torchvision/prototype/features/_bounding_box.py#L62-L64

Happy to resolve the comment and put the same onto images. The dispatcher would then look something like

```py
if isinstance(inpt, features.Image):
    return inpt.to_color_space(...)
elif isinstance(inpt, features._Feature):
    return self
elif ...
```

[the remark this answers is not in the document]
vfdev-5: Sounds good. Yes, that's the point of my comment == to have only `Image.convert_color_space` method and not `_Feature.convert_color_space`.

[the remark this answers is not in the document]
datumbox: Yes we all agree. Adding it only to Image makes sense.

torchvision/prototype/features/_image.py
```diff
@@ -99,6 +99,21 @@ def guess_color_space(data: torch.Tensor) -> ColorSpace:
         else:
             return ColorSpace.OTHER
 
+    def to_color_space(self, color_space: Union[str, ColorSpace], copy: bool = True) -> Image:
+        # import at runtime to avoid cyclic imports
+        from torchvision.prototype.transforms.functional import convert_color_space_image_tensor
```
datumbox, line 104: Nit: Can you use the same style imports as everywhere else?
```python
from torchvision.prototype.transforms import functional as _F
```

torchvision/prototype/transforms/functional/_meta.py
```diff
@@ -154,3 +154,23 @@ def convert_image_color_space_pil(
...
+    inpt: Any, *, color_space: ColorSpace, old_color_space: Optional[ColorSpace] = None, copy: bool = True
+) -> Any:
+    if isinstance(inpt, Image):
+        return inpt.to_color_space(color_space, copy=copy)
+    if isinstance(inpt, _Feature):
+        return inpt
```
datumbox, line 165: I don't think the kernels should offer pass through functionality here. No other kernel does as far as I recall. Do we really need this?
```suggestion
```

datumbox (approved):
Thanks @pmeier. I've added 2 nits but overall looks good.

@vfdev-5 any thoughts?

torchvision/prototype/transforms/functional/_meta.py
pmeier: > No other kernel 

You mean _dispatcher_, right? This is not a kernel. 

Plus, all current dispatchers implement a pass through. The calls look like

https://github.com/pytorch/vision/blob/4db84b04838523594c7fa2250649492e0d41e150/torchvision/prototype/transforms/functional/_geometry.py#L51-L52

and the default implementation is

https://github.com/pytorch/vision/blob/4db84b04838523594c7fa2250649492e0d41e150/torchvision/prototype/features/_feature.py#L89-L90

I'm all for changing this default behavior, but so far this PR is consistent with the rest of the API.

torchvision/prototype/transforms/functional/_meta.py
```diff
@@ -154,3 +154,23 @@ def convert_image_color_space_pil(
...
 
     return image.convert(new_mode)
+
+
+def convert_color_space(
+    inpt: Any, *, color_space: ColorSpace, old_color_space: Optional[ColorSpace] = None, copy: bool = True
+) -> Any:
+    if isinstance(inpt, Image):
+        return inpt.to_color_space(color_space, copy=copy)
+    if isinstance(inpt, _Feature):
```
pmeier, line 164: For consistency

```suggestion
    elif isinstance(inpt, _Feature):
```

torchvision/prototype/transforms/functional/_meta.py
pmeier: As discussed offline, Vasilis is right to remove this branch here. There are two places where a passthrough should be defined:

1. The transform knows that a specific type cannot be handled and thus doesn't pass it to the dispatcher in the first place.
2. The feature does not overwrite the dispatch methods which defaults to passthrough.

The dispatcher itself should not do this.

pmeier:
As discussed offline, this PR will also replace the `apply_recursively` calls with `torch.utils._pytree.tree_map` to avoid custom code.

pmeier:
`query_recursively` can also be removed, but it is a little more complicated. I'll do that in a separate PR when we have finished porting the transforms.

torchvision/prototype/transforms/_geometry.py
```diff
@@ -194,31 +194,22 @@ def forward(self, *inputs: Any) -> Any:
 
 
 class BatchMultiCrop(Transform):
+    def _batch(self, crops: Any) -> Any:
```
datumbox, line 197: Nit: Why not name it `_transform()` as with every other class? It just happens that this Transform is doing batching.

datumbox (approved):
LGTM, thanks! Just an optional nit for your attention.

torchvision/prototype/transforms/_geometry.py
pmeier: Just my misconception that we can't use the regular idiom for `MultiCropResult`. But `pytree` treats it differently than regular `list`'s or `tuple`'s and thus we can use it after all. Thanks for the heads-up.

PR #6409, opened 2022-08-12
Fix M1 worklfow comments

pmeier:
I found this while getting familiar with the new workflows. `shlex` does not have a `which` function, but [`shutil` does](https://docs.python.org/3/library/shutil.html#shutil.which):

https://github.com/pytorch/vision/blob/6e535db255cee3ce878dd7a54dda01d4ec8932c1/setup.py#L105


datumbox (approved):
LGTM, just waiting for @atalman to confirm they didn't mean something different.

atalman (approved):
LGTM

PR #6410, opened 2022-08-12
[proto] Another attemp to rewrite RandomCrop

vfdev-5:


torchvision/prototype/transforms/_geometry.py
```diff
@@ -460,31 +476,29 @@ def _get_params(self, sample: Any) -> Dict[str, Any]:
...
         top = torch.randint(0, height - output_height + 1, size=(1,)).item()
         left = torch.randint(0, width - output_width + 1, size=(1,)).item()
-        return dict(top=top, left=left, height=output_height, width=output_width)
 
-    def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
-        return F.crop(inpt, **params)
+        return dict(
+            top=top, left=left, height=output_height, width=output_width,
+            input_width=width, input_height=height
+        )
```
vfdev-5, line 483: In other words we see that `_get_params` precomputed image size for further usage with images, bboxes etc

datumbox:
This approach makes sense to me. I didn't check that the entire logic is correct (I know you are looking into correcting some minor issue) but the overall strategy looks good.

torchvision/prototype/transforms/_geometry.py
```diff
@@ -448,43 +449,62 @@ def __init__(
...
+        # We have to store maybe padded image size for pad_if_needed branch in _transform
+        input_height, input_width = height, width
+
+        if self.pad_if_needed:
+            # pad width if needed
+            if width < output_width:
+                width += 2 * (output_width - width)
+            # pad height if needed
+            if height < output_height:
+                height += 2 * (output_height - height)
```
vfdev-5, line 472: @datumbox to explain why I didn't implemented that earlier. What I do not like in this approach is that we have to precompute image size before doing any manipulations on the data...

datumbox: Yeap understood. At the same time, this re-computation issue was discussed previously with the rotate and we decided it was OK to do it. So I'm also OK doing it here. We can review if this turns out to be a slowdown.

torchvision/prototype/transforms/_geometry.py
```diff
@@ -448,43 +449,62 @@ def __init__(
...
-                sample = F.pad(sample, padding=padding, fill=self.fill, padding_mode=self.padding_mode)
-
-        return super().forward(sample)
+            input_width, input_height = params["input_width"], params["input_height"]
+            if input_width < self.size[1]:
+                padding = [self.size[1] - input_width, 0]
+                inpt = F.pad(inpt, padding=padding, fill=self.fill, padding_mode=self.padding_mode)
+            if input_height < self.size[0]:
+                padding = [0, self.size[0] - input_height]
+                inpt = F.pad(inpt, padding=padding, fill=self.fill, padding_mode=self.padding_mode)
```
datumbox, line 505: Isn't it possible to combine some of these 3 pads together? I'm afraid that doing 3 separate ones here will end up being slow.

torchvision/prototype/transforms/_geometry.py
vfdev-5: > At the same time, this re-computation issue was discussed previously with the rotate and we decided it was OK to do it

If you talk about this:
https://github.com/pytorch/vision/blob/c31db6d640fc92fcabbd8ae7212eb979096ca0be/torchvision/prototype/features/_bounding_box.py#L174-L182
IMO it is a bit another degree of recomputed things

datumbox: > IMO it is a bit another degree of recomputed things

Just to make sure I get your point, what you mean another degree? Which one in your opinion is worse/more-costly?

torchvision/prototype/transforms/_geometry.py
vfdev-5: Yes, maybe it could be possible. Without correctness tests do you think we should try to do that right now ?

torchvision/prototype/transforms/_geometry.py
vfdev-5: In case of rotate, we recompute due to low-level ops that do not provide meta-data.
In case of random crop, we show that if combined transformation needs dynamic params (next op's params depend on previous op's result), we have to precompute the whole pipeline in `_get_params` before executing actual ops on real data.

torchvision/prototype/transforms/_geometry.py
datumbox: You are right. Let's just add a TODO to optimize. This is going to give us a nice speed improvement over the existing API, as we use this practice on the old API. We should be on the look out for such patterns that we can optimize and mark them with performance TODOs for the next milestone.

torchvision/prototype/transforms/_geometry.py
datumbox: OK so the reason for the recomputation is different but the practice remains the same. Does one of the two practices concerns you more? I just want to understand if you have concerns over the use of this idiom.

vfdev-5: No blocking concerns right now. Let's move on :)

datumbox (approved):
LGTM, thanks.

@pmeier could you please also verify?

pmeier (approved):
LGTM if CI is green.

PR #6411, opened 2022-08-12
port `ScaleJitter` from detection reference to prototype transforms

pmeier:
```py
import torch


image = features.EncodedImage.from_path(
    "test/assets/fakedata/logos/rgb_pytorch.png"
).decode()

bounding_boxes = features.BoundingBox(
    [[60, 30, 15, 15]],
    format=features.BoundingBoxFormat.CXCYWH,
    image_size=(100, 100),
    dtype=torch.float,
)

segmentation_masks = torch.zeros((1, 100, 100), dtype=torch.bool)
segmentation_masks[..., 24:36, 55:66] = True
segmentation_masks = features.SegmentationMask(segmentation_masks)

target = dict(
    boxes=bounding_boxes,
    masks=segmentation_masks,
)
sample = image, target

transform = transforms.ScaleJitter(target_size=(100, 100))

torch.manual_seed(0)

for i in range(5):
    transformed_image, transformed_target = transform(sample)
    transformed_bounding_boxes = transformed_target["boxes"]
    transformed_segmentation_masks = transformed_target["masks"]
```

    

I've verified that plugging `image` and `target` in the current reference transform yields the same results.

### DIFF
--- a/test/test_prototype_transforms.py
+++ b/test/test_prototype_transforms.py
@@ -1148,17 +1148,17 @@ class TestScaleJitter:
         assert int(target_size[1] * scale_range[0]) <= width <= int(target_size[1] * scale_range[1])
 
     def test__transform(self, mocker):
-        interpolation_sentinel = object()
+        interpolation_sentinel = mocker.MagicMock()
 
         transform = transforms.ScaleJitter(target_size=(16, 12), interpolation=interpolation_sentinel)
-        transform._transformed_types = (object,)
+        transform._transformed_types = (mocker.MagicMock,)
 
-        size_sentinel = object()
+        size_sentinel = mocker.MagicMock()
         mocker.patch(
             "torchvision.prototype.transforms._geometry.ScaleJitter._get_params", return_value=dict(size=size_sentinel)
         )
 
-        inpt_sentinel = object()
+        inpt_sentinel = mocker.MagicMock()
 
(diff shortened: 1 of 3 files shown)


datumbox (approved):
LGTM, thanks!

vfdev-5:
How about adding tests like for other transforms ?

test/test_prototype_transforms.py
```diff
@@ -1102,3 +1102,42 @@ def test_ctor(self, trfms):
...
+
+        transform = transforms.ScaleJitter(target_size=(16, 12), interpolation=interpolation_sentinel)
+        transform._transformed_types = (object,)
+
+        size_sentinel = object()
+        mocker.patch(
+            "torchvision.prototype.transforms._geometry.ScaleJitter._get_params", return_value=dict(size=size_sentinel)
+        )
+
+        inpt_sentinel = object()
```
vfdev-5, line 1161: Is it better to use `object()` instead of `MagicMock()` ?

pmeier: For this purpose they are equivalent. `MagicMock` is useful if the object should have a specific structure or will be transformed by transform. Both things are not the case here.

vfdev-5: For consistency I'd use `MagicMock`

vfdev-5 (approved):
LGTM, thanks @pmeier !

Issue #5319, opened 2022-01-31
New Model Architectures - Implementation and Documentation Details


jdsgomes:
### 🚀 The feature

When adding a new model architecture there are some design/implementation details and documentation requirements that need to be taken into account. This issue intents to track such details in a dynamic manner, as it is possible to change over time.

### Model development and training steps

When developing a new model there are some details not to be missed:

- Implement a model factory function for each of the model variants

- in the module constructor, [pass layer constructor instead of instance](https://github.com/pytorch/vision/blob/47bd962069ba03f753e7ba711cb825317be0b00a/torchvision/models/efficientnet.py#L88) for configurable layers like norm, activation, and log the api usage with `_log_api_usage_once(self)`

- fuse layers together with existing common blocks if possible; For example consecutive conv, bn, activation layers could be replaced by [ConvNormActication](https://github.com/pytorch/vision/blob/47bd962069ba03f753e7ba711cb825317be0b00a/torchvision/ops/misc.py#L104)

- define `__all__` in the beginning of the model file to expose model factory functions; import model public APIs (e.g. factory methods) in `torchvision/models/__init__.py`

- create the model builder using the new API and add it to the prototype area. Here is an [example](https://github.com/pytorch/vision/pull/4784/files) on how to do this. The new API requires adding more information about the weights such as the preprocessing transforms necessary for using the model, meta-data about the model, etc

- Make sure you write tests for the model itself (see `_check_input_backprop`, `_model_params` and `_model_params` in `test/test_models.py`) and for any new operators/transforms or important functions that you introduce

Note that this list is not exhaustive and there are details here related to the code quality etc, but these are rules that apply in all PRs (see [Contributing to TorchVision](https://github.com/pytorch/vision/blob/main/CONTRIBUTING.md)).

Once the model is implemented, you need to train the model using the reference scripts. For example, in order to train a classification resnet18 model you would:

1. go to `references/classification`

2. run the train command (for example `torchrun --nproc_per_node=8 train.py --model resnet18`)

After training the model, select the best checkpoint and estimate its accuracy with a batch size of 1 on a single GPU. This helps us get better measurements about the accuracy of the models and avoid variants introduced due to batch padding (read [here](https://github.com/pytorch/vision/pull/4609/commits/5264b1a670107bcb4dc89e83a369f6fd97466ef8) for more details).

Finally, run the model test to generate expected model files for testing. Please include those generated files in the PR as well.:

[...]

PR #6412, opened 2022-08-12
Add the S3D architecture to TorchVision

### LINKED ISSUE #6402: S3D feature request
### 🚀 The feature

S3D model for video classification ([paper](https://arxiv.org/abs/1712.04851), [paperswithcode](https://paperswithcode.com/paper/rethinking-spatiotemporal-feature-learning), [implementation](https://github.com/facebookresearch/multimodal/blob/main/examples/mugen/retrieval/s3d.py) in torchmultimodal).

### Motivation, pitch

I've onboarded a [model](https://github.com/facebookresearch/multimodal/blob/main/examples/mugen/retrieval/video_clip.py) to TorchMultimodal that depends on S3D. Right now the S3D implementation is in torchmultimodal, but as a purely vision model (and a fairly popular model) it would fit nicely in torchvision. There is some discussion on [this PR in torchmultimodal](https://github.com/facebookresearch/multimodal/pull/135) about the changes needed to onboard S3D to torchvision. 

### Alternatives

Right now, S3D is implemented in torchmultimodal.

### Additional context

I started working on the [suggested refactoring](https://github.com/facebookresearch/multimodal/pull/135) to move S3D from multimodal to vision. However, the process of onboarding myself to the torchvision library and implementing the shared base with the inception model will likely take more time than I have, and my torchvision POC is on PTO. Are there any steps I can take within the next week (my internship is ending), or can the model be onboarded by the torchvision team?

sophiazhi:
Fixes #6402 

Add S3D model for video classification.


torchvision/models/video/s3d.py
```diff
@@ -0,0 +1,173 @@
...
+            pre-trained weights are used.
+        progress (bool): If True, displays a progress bar of the download to stderr. Default is True.
+        **kwargs: parameters passed to the ``torchvision.models.video.S3D`` base class.
+            Please refer to the `source code
+            <https://github.com/pytorch/vision/blob/main/torchvision/models/video/s3d.py>`_
+            for more details about this class.
+
+    """
+    weights = S3D_Weights.verify(weights)
+
```
datumbox, line 167: You need the following bit. In case someone asks for a pretrained model, the num_classes is predetermined.
```suggestion

    if weights is not None:
        _ovewrite_named_param(kwargs, "num_classes", len(weights.meta["categories"]))
```

torchvision/models/video/s3d.py
```diff
@@ -0,0 +1,173 @@
...
+        logits = torch.mean(y, 2)
+
+        return logits
+
+
+class S3D_Weights(WeightsEnum):
+    pass
+
+
+@register_model()
```
datumbox, line 147: Once you make any necessary change on the architecture, you would need to produce expected files so that all the tests work. This is very easy, you basically run the test for the specific model one by passing a specific flag:
```
EXPECTTEST_ACCEPT=1 pytest test/test_models.py -k s3d
```

You can read more at #5319

torchvision/models/video/s3d.py
```diff
@@ -0,0 +1,173 @@
...
+        y = F.avg_pool3d(y, (2, y.size(3), y.size(4)), stride=1)
+        y = self.fc(y)
+        y = y.view(y.size(0), y.size(1), y.size(2))
+        logits = torch.mean(y, 2)
+
+        return logits
+
+
+class S3D_Weights(WeightsEnum):
+    pass
```
datumbox, line 144: Again once everything is inplace, we would need to offer weights. Let's leave this last. Are there official paper weights we can use?

datumbox (commented):
Thanks @sophiazhi, the PR looks good overall. I have added a few comments, please let me know what you think.

datumbox:
It might be worth looking into the [original TF implementation](https://github.com/tensorflow/models/blob/master/research/slim/nets/s3dg.py#L568-L584) to confirm you implement S3D as expected. I know you don't do the Gated variant but this reference is still valid. As you can see the final `AvgPool_0a_7x7` layer uses a `kernel=(2,7,7)` and performs the reduction as I described at https://github.com/pytorch/vision/pull/6412#discussion_r945581645.

It's also worth noting the are using a dropout layer at the end. See this [idiom](https://github.com/pytorch/vision/blob/main/torchvision/models/alexnet.py#L18) on how to pass the parameter to the network.

Finally is there any reason we implement S3D and not it's gated variant S3D-G which is marginally more accurate?

torchvision/models/video/s3d.py
langong347: Noob Q: Are the expected files just the tensor outputs for running the forward pass? (like [here](https://github.com/pytorch/vision/blob/main/test/test_models.py#L622-L623)) 

I wonder how does this testing strategy handle the following cases:
- The forward signature contains optional (e.g., attention masks) or boolean args which can trigger different code paths / logic
- The output of the forward pass is compounded, e.g., a namedtuple with multiple fields of different types like [here](https://github.com/facebookresearch/multimodal/blob/main/torchmultimodal/models/gpt.py#L18-L22)
- Error raising statements

The usual way is to write test cases manually for each of the codepaths within the same model. Looks like TorchVision just focuses on the default codepath?

langong347: Also noticed that you set random seed just once before calling the model? [here](https://github.com/pytorch/vision/blob/main/test/test_models.py#L606)

If at some point the model gets refactored, e.g., by changing the order of the layers containing randomness, it will break the test due to change in the sequence of random numbers?

torchvision/models/video/s3d.py
sophiazhi: I used [these weights](https://github.com/kylemin/S3D#weight-file--sample-code) pretrained on kinetics400. they were used in the project i onboarded to torchmultimodal, but they're not the official weights (which don't exist as far as i know) from the paper

torchvision/models/video/s3d.py
datumbox: @langong347 Good points. Here are some thoughts:

- This testing infra discussed here only covers the forward for every task. Typically all tasks receive the same inputs, so we have separate tests for them. 
- If you have a set of masked models, you would create tests for all of these as well. 
- In our case, we have also various [other tests](https://github.com/pytorch/vision/blob/main/test/test_models.py#L411-L571) to cover other codepaths, including testing for specific [error raising](https://github.com/pytorch/vision/blob/main/test/test_models.py#L800-L825) statements.

> If at some point the model gets refactored, e.g., by changing the order of the layers containing randomness, it will break the test due to change in the sequence of random numbers?

Correct. If the order gets changed, then the developer would have to update the expected file and explain on the PR why that is the case. I agree that this is problematic and our plan is to refactor our tests to load actual weights going forwards. This will give us more stable results (especially for Object Detection models), test the loading of the weights and fix other similar corner-cases.

torchvision/models/video/s3d.py
datumbox: It's not clear from the repo what is the size of the video they use. I don't see any resize statement so I assume it's `384 × 256 px` as the images on the repo. The specific aspect ratio is a bit odd/non-standard. As part of adding the weights, we would have to verify the accuracy on Kinetics400 and guesstimate the central-crop/resize they are using to reproduce the results.

torchvision/models/video/s3d.py
langong347: @datumbox Thanks! I notice your pattern of having one type of test for a class of similar models. Wonder how TorchVision tests have evolved to this high degree of uniformity? Is there a test design doc or wiki that I may take a look?

torchvision/models/video/s3d.py
```diff
@@ -0,0 +1,214 @@
...
+
+__all__ = [
+    "S3D",
+    "S3D_Weights",
+    "s3d",
+]
+
+
+class TemporalSeparableConv(nn.Sequential):
+    def __init__(self, in_planes: int, out_planes: int, kernel_size: int, stride: int, padding: int = 0):
```
datumbox, line 22: It's a bit weird that we don't use the same `norm_layer` here as the rest of the network. I understand you do this because of different eps/momentum values? Worth clarifying that the values are indeed different by looking at the original TF implementation and on the other implementations. 

If it turns out this is needed, it might be worth we retrain the network to use the default eps/momentum and have a simpler/better API. Quite often we can simplify hyperparams without losing accuracy or compensate for the tiny drop by doing a newer optimization (augmentation, loss etc).

datumbox (commented):
Minus reproducing the accuracy, I think we are almost there. Just minor comments left.

torchvision/models/video/s3d.py
sophiazhi: it's actually the same `norm_layer` as the rest of the network, I just missed this module when i was making `norm_layer` configurable :)

torchvision/models/video/s3d.py
```diff
@@ -0,0 +1,240 @@
...
+    """
+
+    def __init__(
+        self,
+        in_planes: int,
+        out_planes: int,
+        kernel_size: int,
+        stride: int,
+        padding: int = 0,
+        norm_layer: Optional[Callable] = None,
```
datumbox, line 41:
```suggestion
        padding: int,
        norm_layer: Callable[..., torch.nn.Module],
```

torchvision/models/video/s3d.py
```diff
@@ -0,0 +1,240 @@
...
+    ):
+        if norm_layer is None:
+            norm_layer = partial(nn.BatchNorm3d, eps=0.001, momentum=0.001)
+
```
datumbox, line 45:
```suggestion
```

torchvision/models/video/s3d.py
```diff
@@ -0,0 +1,240 @@
...
+        b1_mid: int,
+        b1_out: int,
+        b2_mid: int,
+        b2_out: int,
+        b3_out: int,
+        norm_layer: Optional[Callable] = None,
+    ):
+        super().__init__()
+        if norm_layer is None:
+            norm_layer = partial(nn.BatchNorm3d, eps=0.001, momentum=0.001)
```
datumbox, line 95:
```suggestion
        norm_layer: Callable[..., torch.nn.Module],
    ):
        super().__init__()
```

torchvision/models/video/s3d.py
```diff
@@ -0,0 +1,240 @@
...
+        norm_layer: Optional[Callable] = None,
+    ) -> None:
+        super().__init__()
+        _log_api_usage_once(self)
+
+        if norm_layer is None:
+            norm_layer = partial(nn.BatchNorm3d, eps=0.001, momentum=0.001)
+
+        self.features = nn.Sequential(
+            TemporalSeparableConv(3, 64, kernel_size=7, stride=2, padding=3),
```
datumbox, line 146: Nit: why do we pass named params for some value arguments and for some we don't? Also we should pass explicitly the `norm_layer` that the user provided everywher to ensure consistency.
```suggestion
            TemporalSeparableConv(3, 64, 7, 2, 3, norm_layer),
```

torchvision/models/video/s3d.py
```diff
@@ -0,0 +1,240 @@
...
+        norm_layer (Optional[Callable]): Module specifying the normalization layer to use.
+
+    Inputs:
+        x (Tensor): batch of videos with dimensions (batch, channel, time, height, width)
+    """
+
+    def __init__(
+        self,
+        num_classes: int = 400,
+        dropout: float = 0.2,
```
datumbox, line 136: Your dropout might be different from the one used on your pre-trained weights. It's worth checking if that's the source of the discrepancy. You can set it to 0.0 here temporarily and I can help you overwrite it and fix it more permanently just before merging.

torchvision/models/video/s3d.py
```diff
@@ -0,0 +1,240 @@
...
+        norm_layer: Optional[Callable] = None,
```
datumbox, line 137:
```suggestion
        norm_layer: Optional[Callable[..., torch.nn.Module]] = None,
```

datumbox:
Porting weights from @kylemin's [repo](https://github.com/kylemin/S3D) yields lower accuracy than the expected 72.1%:
```
torchrun --nproc_per_node=8 train.py --data-path="/datasets01/kinetics/070618/400/" --batch-size=16 \
--test-only --cache-dataset --clip-len 128 --clips-per-video 1 --model s3d --weights="S3D_Weights.DEFAULT"
 * Clip Acc@1 67.315 Clip Acc@5 87.593
```

Our setup is a quite similar to what described at the [paper](https://arxiv.org/pdf/1712.04851.pdf) (page 5, section 3.2):
> During training, we densely sample 64 frames from a video and resize input frames to 256 × 256 and then take random crops of size 224×224. During evaluation, we use all frames and take 224×224 center crops from the resized frames.

The only difference is that we don't take all available frames because that's not easily supported on our scripts. I did create an experimental PR at #6445 that made it possible to score all frames and the accuracy remains around 67%. We also tested the original implementation from the repo and the accuracy remains the same. Finally we noticed that the original repo uses a different input size (height=224, weigth=384), but that doesn't make much of a difference either.

So it seems that there is some discrepancy on how the preprocessing is done or something is different on the weights. Perhaps @kylemin could provide some additional details on the 72.1% accuracy was achieved. We have some details from issue https://github.com/kylemin/S3D/issues/6 but so far we think our approach is aligned with the description of the author.

datumbox (approved):
@sophiazhi Thanks a lot for your work. LGTM.

I am approving this PR on the basis that the model achieves decent accuracy for its size and that unblocks Multimodal. On the cons side, we are using ported weights from an unofficial repo which seem to be lagging by 5 accuracy points comparing to the paper. To mitigate this I have:
1. [Pinged](https://github.com/pytorch/vision/pull/6412#issuecomment-1219687434) the author of the repo to clarify the exact inference setup they use.
2. Started an internal job (id 55651) to train the network from scratch and potentially replace the weights if we get better accuracy.

Since the gap in accuracy is on the borderline of what we usually consider acceptable and because I've made multiple commits on the PR, I would like a second opinion on how we should progress here.

@jdsgomes @langong347 Please let me know your thoughts.

kylemin:
@datumbox There might be a few differences:
1) Instead of using the full-resolution frames, please try performing a 5-crop (each crop is 224x224) inference
2) Instead of processing all the frames (250) at once, please try dividing a video into multiple segments (each segment consists of 64 frames) and taking the average for the final prediction
3) We used the highest value of jpeg quality (minimal compression) both in the training and inference processes
4) I think the interpolation also does matter. Please try bicubic resizing.

Thank you,
Kyle

kylemin:
Plus, the transform function might be different than ours. Please refer to [this link](https://github.com/kylemin/S3D/blob/master/main.py#L63)

jdsgomes:
> 2\. Instead of processing all the frames (250) at once, please try dividing a video into multiple segments (each segment consists of 64 frames) and taking the average for the final prediction
> 3\. We used the highest value of jpeg quality (minimal compression) both in the training and inference processes
> 4\. I think the interpolation also does matter. Please try bicubic resizing.
> 
> Thank you,



> @sophiazhi Thanks a lot for your work. LGTM.
> 
> I am approving this PR on the basis that the model achieves decent accuracy for its size and that unblocks Multimodal. On the cons side, we are using ported weights from an unofficial repo which seem to be lagging by 5 accuracy points comparing to the paper. To mitigate this I have:
> 
> 1. [Pinged](https://github.com/pytorch/vision/pull/6412#issuecomment-1219687434) the author of the repo to clarify the exact inference setup they use.
> 2. Started an internal job (id 55651) to train the network from scratch and potentially replace the weights if we get better accuracy.
> 
> Since the gap in accuracy is on the borderline of what we usually consider acceptable and because I've made multiple commits on the PR, I would like a second opinion on how we should progress here.
> 
> @jdsgomes @langong347 Please let me know your thoughts.

tldr; from my point of view this can be merged to unblock but ideally we would follow up and try to close the gap.

This is a difficult one, since the gap is larger that we would normally accept (at least from recent contributions, we might find some cases where this was different).  I see that @kylemin suggested some changes in the testing setup so might be worth trying them to see if we can close the gap (point 1 you already did as far as I know, but its not clear if you did the others).

Another option would be to make it internal only to unblock until we can close the gap.

torchvision/models/video/s3d.py
```diff
@@ -0,0 +1,216 @@
...
+        x = self.avgpool(x)
+        x = self.classifier(x)
+        x = torch.mean(x, dim=(2, 3, 4))
+        return x
+
+
+class S3D_Weights(WeightsEnum):
+    KINETICS400_V1 = Weights(
+        url="https://download.pytorch.org/models/s3d-1bd8ae63.pth",
+        transforms=partial(
```
kylemin, line 157: I think the transforms function should be modified to be consistent with ours. Please refer to [this link](https://github.com/kylemin/S3D/blob/master/main.py#L63)

datumbox:
@kylemin Thank you very much for the prompt reply. I have some additional questions to ensure we are doing the right thing on our side:

> I think the transforms function should be modified to be consistent with ours. Please refer to [this link](https://github.com/kylemin/S3D/blob/master/main.py#L63)

I understand that your method scales the input to [-1, 1]. It's not immediate obvious but that's also what we do. Is there anything other inconsistency you spotted that we should fix?

> Instead of using the full-resolution frames, please try performing a 5-crop (each crop is 224x224) inference

So instead of resizing the video to 256x256, you propose to do a 5-crop of 224x224 each. Correct?

> Instead of processing all the frames (250) at once, please try dividing a video into multiple segments (each segment consists of 64 frames) and taking the average for the final prediction

Sounds good. Do you remember how many 64-frame clips you produce for each Video?

> I think the interpolation also does matter. Please try bicubic resizing.

Sounds good, I'll make the change. Thanks for pointing this out.

kylemin:
@datumbox I am not sure there is another inconsistency, but I believe that modifying the transforms function would improve the accuracy and mostly resolve the issue of performance discrepancy. Yes, I meant 5-crop of 224x224. I remember that this improved the inference accuracy a little bit. I think the rest (dividing a video into multiple segments, jpeg quality, bicubic interp) can wait and are not in high priority because they might not be the reason why there is a non-negligible performance discrepancy. I am unsure if I processed 250 frames at once or not, but I remember that I once tested using four 64-frame clips for each video. Sorry for this uncertainty, but I hope the problem is resolved by modifying the transforms function and scaling the input to [-1, 1].

datumbox:
I am going to revert the 5/10 crop update I did on transforms because this leads to incompatible batch & collate logic on the reference scripts. The issue is that the aforementioned augmentations make new videos "popup" and thus the length of labels and ids no longer match. It would require far more work to make the references work with this that I don't think it's worth it at this point. Since I plan to hard reset the branch, you can find the full changes [here](https://github.com/sophiazhi/vision/blob/6c7da4744905cd6734ab2824f26a5d6044c6ce89/torchvision/transforms/_presets.py#L84-L156). Part of the reason I'm not making the rest of the required changes is that the original paper explicitly mentions they do single central crops for inference.

I'm still going to test the proposed change on interpolation because that can indeed cause issues.

jdsgomes (approved):
Checked the latest changes and LGTM!
thanks

langong347 (commented):
Sorry didn't get a chance to review yesterday. Had a question about the test input. The code looks good to me.

As for upstreaming to torchvision: we can use our local version S3D for the moment while the weights are getting finalized. Personally, I don't think you need to make this implementation internal as long as the discrepancy of 5% in accuracy is temporarily acceptable.

test/test_models.py
```diff
@@ -312,6 +312,9 @@ def _check_input_backprop(model, inputs):
     "mvit_v2_s": {
         "input_shape": (1, 3, 16, 224, 224),
     },
+    "s3d": {
+        "input_shape": (1, 3, 16, 224, 224),
```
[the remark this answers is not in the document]
datumbox: The architecture doesn't support the default input size (which is smaller) so here we define the "smallest reasonable" input to make the test work.

datumbox:
@langong347 We have merged the PR and it should appear on the next nightly tomorrow. The weights we use here are identical to the ones you use on your local repo. This means that you don't have to wait for anything to adopt the version contributed by Sophia.

kylemin:
@datumbox Could we test if loading images in BGR instead of RGB can recover the performance? I found this log of S3D experiments and I think that how each video is scored on the temporal dimension does not affect the performance that much.
image
About the highest value of jpeg quality, I meant when you extract frames from videos. FFmpeg extracts frames of low quality by default so I think I changed the setup. However, I don't think that explains such a large performance gap.. I'll try to log in to my previous school server to see if the information about how I extracted the frames and how I performed the inference is still there. Thanks!

kylemin:
So I've found a script that was used to extract the frames using ffmpeg and save them in jpeg format. I am not sure this helps at this point.. but anyway I'll leave it here. Thanks

```
import os
import subprocess

from joblib import delayed
from joblib import Parallel
import pandas as pd
import numpy as np
#import jpeg4py as jpeg
import cv2
import csv
import h5py


def download_clip(video_identifier, start_time, end_time, sp, lstr, label):
    tmp_filename = os.path.join('../Kinetics-400', sp, '%s'%lstr, '%s_%06d_%06d.mp4' % (video_identifier, start_time, end_time))

    proc = subprocess.Popen('ffprobe -v error -select_streams v:0 -show_entries stream=width,height -of csv=p=0 "' + tmp_filename + '"', stdout=subprocess.PIPE, shell=True)
    (out, err) = proc.communicate()
    if err is not None:
        return None, None, None, 0
    else:
        if out:
            w, h = out.decode().strip().split(',')
            w = int(w)
            h = int(h)
            scale = ''
            if w >= h:
                w = int(float(w)*256./float(h))
                w = w - w%2
                h = 256
                if w > 384:
                    w = 384
            else:
                h = int(float(h)*256./float(w))
                h = h - h%2
                w = 256
                if h > 384:
                    h = 384
            scale = '%s:%s' % (w,h)
        else:
            return None, None, None, 0

    command = ['ffmpeg', '-hide_banner','-loglevel', 'quiet', '-debug', '0',#'panic'
               '-i', '"%s"' % tmp_filename,
               '-f', 'image2pipe', '-qscale:v', '1',
               '-filter:v', "scale=%s"%scale, '-r','25', #'25'
               '-threads', '1', '-pix_fmt', 'bgr24','-vcodec','rawvideo','-']
    command = ' '.join(command)

    max_num = 250
    num_frame = max_num
    np_frame = np.ndarray((1, max_num), dtype='object')
    pipe = subprocess.Popen(command, shell=True, stdout=subprocess.PIPE, bufsize=10**7)
    for i in range(max_num):
        try:
            if i < num_frame:
                raw_image = pipe.stdout.read(h*w*3)
                img = np.fromstring(raw_image, dtype='uint8')
                img = img.reshape((h, w, 3))
                jpg = cv2.imencode('.jpg', img, [cv2.IMWRITE_JPEG_QUALITY, 99])[1]
                jpg = jpg.squeeze()
                np_frame[0, i] = np.fromstring(jpg.tostring(), dtype='uint8')
                pipe.stdout.flush()
            else:
                np_frame[0, i] = np.array([], dtype=np.uint8)
        except:
            if num_frame > i:
                num_frame = i
            np_frame[0, i] = np.array([], dtype=np.uint8)

    np_label = None
    if label >= 0:
        np_label = np.zeros((1,), dtype=int)
        np_label[0] = label
    np_shape = np.zeros((1, 4), dtype=int)
    np_shape[0] = [num_frame, h, w, 3]

    return np_label, np_shape, np_frame, num_frame


def download_clip_wrapper(i, row, sp, lstr='', label=-1):
    np_label, np_shape, np_frame, num_frame = download_clip(row['video-id'], row['start-time'], row['end-time'], sp, lstr, label)

    status = tuple([i, np_label, np_shape, np_frame, num_frame])
    return status


def parse_kinetics_annotations(input_csv, ignore_is_cc=False):
    df = pd.read_csv(input_csv)
    if 'youtube_id' in df.columns:
        columns = OrderedDict([
            ('youtube_id', 'video-id'),
            ('time_start', 'start-time'),
            ('time_end', 'end-time'),
            ('label', 'label-name')])
        df.rename(columns=columns, inplace=True)
        if ignore_is_cc:
            df = df.loc[:, df.columns.tolist()[:-1]]
    return df


def main():
    sp = 'val'
    target = '/z/home/kylemin/dataset/Kinetics-new/Kinetics-400h/'
    class_label = tuple(csv.reader(open('../class_label.csv', 'r')))
    video_list = list(csv.reader(open(os.path.join(target, sp+'_list.csv'), 'r')))

    with h5py.File(os.path.join(target, sp+'.h5'), 'w', libver='latest') as hf:
        list_input = []
        if sp == 'test':
            list_input.append(['youtube_id','time_start','time_end','split'])
        else:
            list_input.append(['label','youtube_id','time_start','time_end','split'])

        for vs in video_list:
           cls = vs[0].split('_')
           if sp == 'test':
               list_input.append([vs[0][:11],int(cls[-2]),int(cls[-1])])
           else:
               list_input.append([vs[1],vs[0][:11],int(cls[-2]),int(cls[-1]),sp])

        input_csv = 'input.csv'
        with open(input_csv, 'w') as f:
            writer = csv.writer(f)
            writer.writerows(list_input)

        dataset = parse_kinetics_annotations(input_csv)

        if sp == 'test':
            status_lst = Parallel(n_jobs=10)(delayed(download_clip_wrapper)(
                j, row, sp) for j, row in dataset.iterrows())
        else:
            status_lst = Parallel(n_jobs=10)(delayed(download_clip_wrapper)(
                j, row, sp, row['label-name'], class_label.index([row['label-name']])) for j, row in dataset.iterrows())

        #status_lst = []
        #for j, row in dataset.iterrows():
        #    status_lst.append(download_clip_wrapper(j, row, sp, row['label-name'], class_label.index([row['label-name']])))

        idx_list = []
        for k, l, s, f, nf in status_lst:
            if s is not None and f is not None and nf > 15:
                idx_list.append(k)

        num_video = len(idx_list)
        if sp != 'test':
            dset_labels = hf.create_dataset('labels', shape=(num_video,), dtype=int)
        dset_shapes = hf.create_dataset('shapes', shape=(num_video,4), dtype=int)
        dt = h5py.special_dtype(vlen=np.dtype('uint8'))
        dset_frames = hf.create_dataset('videos', shape=(num_video,250), dtype=dt)

        if sp != 'test':
            np_labels = np.zeros((num_video,), dtype=int)
        np_shapes = np.zeros((num_video, 4), dtype=int)
        np_frames = np.ndarray((num_video, 250), dtype='object')
        for k, l, s, f, nf in status_lst:
            if s is not None and f is not None and nf > 15:
                idx = idx_list.index(k)
                if sp != 'test':
                    np_labels[idx] = l
                np_shapes[idx] = s
                np_frames[idx] = f

        if sp != 'test':
            dset_labels[...] = np_labels
        dset_shapes[...] = np_shapes
        dset_frames[...] = np_frames


if __name__ == '__main__':
    main()
```

datumbox:
@kylemin Thanks so much for digging into the logs and files to help out. I really appreciate the help!

> Could we test if loading images in BGR instead of RGB can recover the performance? I

I've tested that but it doesn't help, the accuracy is lower. I think you were scoring RGB frames because your do the conversion from BGR to RGB on your original repo in [this line](https://github.com/kylemin/S3D/blob/master/main.py#L47).

> FFmpeg extracts frames of low quality by default so I think I changed the setup. However, I don't think that explains such a large performance gap.

I think I understand why you say this now that I can see your preprocessing script at https://github.com/pytorch/vision/pull/6412#issuecomment-1221240288. It's because after decoding you store the frames as JPEG files. We don't do that on our side, we just directly use the decoded picture.

Another difference that I noticed is that you actually resize videos to 256 on the smallest dimension (with a hard max of `384`) instead of 256x256 on the original paper. I tested this out but didn't significantly affect the accuracy. 

So the accuracy gap remains strange to me. If you have any other hypothesis, I'm happy to explore it. Otherwise, thanks a lot for your help!

datumbox:
Training the network from scratch yields a network with the following accuracy using single crops with 128 frames:
```
Checkpoint: job55651/model_34.pth
Clip Acc@1 68.206 Clip Acc@5 87.645
```

Unfortunately the above leads to severe overfitting on the Training set, as we see below:
```
acc1: 100.0000 (99.4226)  acc5: 100.0000 (99.9334)
```

I think we need to add some higher regularization. The paper is a bit unclear on the hyper-params used but on the TF code-base I see the default value for Dropout=0.2. I'll give that a try and if I still see overfitting I'll tune the weight decay as well.

YosuaMichael:
@datumbox I notice from @kylemin script that in the ffmpeg command it use `-r 25` which mean it use frame rate of 25 fps. I am currently running the test using `--frame-rate 25` to check if this improve the result.

YosuaMichael:
I rand with a test with `BICUBIC` and following parameter:
```
>     --batch-size=16 --test-only \
>     --data-path="/datasets01/kinetics/070618/400/" \
>     --clip-len 64 --frame-rate 25 --clips-per-video 5 \
>     --cache-dataset \
>     --model s3d --weights="S3D_Weights.DEFAULT"
```

And the result I got is:
```
Test: Total time: 0:42:11
 * Clip Acc@1 60.389 Clip Acc@5 82.375
 * Video Acc@1 68.261 Video Acc@5 88.393
 ```
 Seems like it still can't reach `~72%` accuracy.

**Comment by datumbox** [comment]
@YosuaMichael Can you try with `--clip-len 128` and `--clips-per-video 1`? Supposedly that's close to how the paper does it.

**Comment by YosuaMichael** [comment]
@datumbox using `--clip-len 128 --clips-per-video 1` I got worse result:
```
Test: Total time: 0:55:10
 * Clip Acc@1 65.626 Clip Acc@5 86.332
```

Here are the full params:
```
>     --batch-size=16 --test-only \
>     --data-path="/datasets01/kinetics/070618/400/" \
>     --clip-len 128 --frame-rate 25 --clips-per-video 1 \
>     --cache-dataset \
>     --model s3d --weights="S3D_Weights.DEFAULT"
```

**Comment by datumbox** [comment]
@YosuaMichael My understanding form reading the documentation of [ffprobe](https://ffmpeg.org/ffprobe-all.html) is that the default interpolation is bilinear not bicubic. Worth checking that as well to see if you get an improvement if you have the bandwidth.

**Comment by datumbox** [comment]
Thanks @YosuaMichael. I've been training a new model the whole week but due to infra problems I don't have a new model. The accuracy you report is on par with the model that I have currently trained from scratch. I still have overfitting problems so I'm sure there is more to be done here to improve it. I'll report results here as I get them.

**Comment by datumbox** [comment]
Training a network with Dropout=0.2 yields the following results:

```
Train Params:  --ngpus 8 --nodes 8 --cache-dataset --batch-size=12 --lr 0.2 --clip-len 64 --clips-per-video 5 --sync-bn --model s3d --train-resize-size 256 256 --train-crop-size 224 224 --val-resize-size 256 256 --val-crop-size 224 224
Val Params: --batch-size=16 --test-only --cache-dataset --clip-len 128 --clips-per-video 1 
Checkpoint: job59255/model_43.pth
* Clip Acc@1 68.345 Clip Acc@5 88.050
```

Which is similar to Dropout=0.0. This model is about +1 Acc@1 point higher than the one we deployed but still have roughly 4 points gap from the original paper. I'm still observing significant amount of overfitting as we don't provide currently offer an good way to do augmentations for Video (coming up). I could keep trying to tweak the configuration to avoid overfitting but this job requires lots of resources and currently want to dedicate them on the testing of Transforms.

@YosuaMichael Any thoughts on whether we should deploy these weights instead of the ported ones?

YosuaMichael:
Hi @datumbox, overall I prefer to deploy the new weight as 1 point difference is quite significant. However I have no strong opinion on this since we may also update the weight later once we have more augmentation on video.

PR #6419, opened 2022-08-15
[proto] Clean-up Label.to_categories

vfdev-5:


datumbox (approved):
LGTM, thanks!

torchvision/prototype/features/_label.py
```diff
@@ -43,10 +43,10 @@ def from_category(
         return cls(categories.index(category), categories=categories, **kwargs)
 
     def to_categories(self) -> Any:
-        if not self.categories:
-            raise RuntimeError()
+        if self.categories is None:
+            raise RuntimeError("Label does not have categories")
 
-        return apply_recursively(lambda idx: cast(Sequence[str], self.categories)[idx], self.tolist())
+        return tree_map(lambda idx: self.categories[idx], self.tolist())
```
NicolasHug, line 49: We have a lot of calls to `apply_recursively`, should we update them all?

NicolasHug: but also, it looks private - should we actually use it ?

vfdev-5: Yes, in transforms I already replaced them.

pmeier (approved):
LGTM if CI is green.

NicolasHug:
@vfdev-5 @pmeier @datumbox any thought on https://github.com/pytorch/vision/pull/6419#discussion_r945772800 ?

vfdev-5:
It is used in functorch eager ops (e.g. vmap: https://github.com/pytorch/functorch/blob/518fecce963846679add02ad36a65d36d59a362d/functorch/_src/vmap.py#L11), so I would expect to be able to use it 
here as well. cc @zou3519

NicolasHug:
I believe `functorch` is bound to be integrated within core eventually, so they don't operate under the same constraints as we do.

We probably want to be careful about using private stuff from core because it will not just break us on GitHub; most importantly, if a Meta engineer modifies that private API internally, they'll start seeing the torchvision tests fail for no good reason. Nobody wants to see their code fail just because someone else has been using a private API from their repo.

zou3519:
functorch isn't the only library that uses `_pytree`, PyTorch Core features (like [torch.fx](https://github.com/pytorch/pytorch/blob/master/torch/fx/_pytree.py)) use `_pytree` utilities as well. 

We haven't exposed `_pytree` publicly yet, but if you want to use it and exposing it would make you feel more comfortable with using that, then please let us know. The reason why it is prefixed with an underscore is because we have not done work to expose it as a public API.

NicolasHug:
> We haven't exposed _pytree publicly yet, but if you want to use it and exposing it would make you feel more comfortable with using that, then please let us know

Great, thanks @zou3519 

@vfdev-5 would you mind submitting a feature request to core for `tree_map()` to be public?

zou3519:
There's an issue already over at https://github.com/pytorch/pytorch/issues/65761

NicolasHug:
@vfdev-5  Out of curiosity what was the original reason to go from `apply_recursively` to `tree_map()`? Looking at https://github.com/pytorch/pytorch/issues/65761 `tree_map()` is said to be slow (not sure if that's still the case, or whether that's something critical for us)

vfdev-5:
Performance issue may be resolved from core side. From torchvision point of view keeping our own implementations (e.g. `apply_recursive` and related methods) would increase codebase maintenance.

PR #6420, opened 2022-08-15
[proto] OneHotLabel subclass Label and implements `from_category` and `to_categories`

vfdev-5:
Description:

- OneHotLabel subclass Label
- Implemented `OneHotLabel.from_category` and `OneHotLabel.to_categories`

```python
import torch
from torchvision.prototype.features import Label, OneHotLabel

Label.from_category("dog", categories=["dog", "cat", "boat"])

l = Label(torch.tensor([[[0, 0, 2], [0, 1, 2]], [[0, 0, 1], [1, 1, 2]]]), categories=["dog", "cat", "boat"])
print(l.to_categories())

l = OneHotLabel(torch.eye(3, 3, dtype=torch.long)[None, None, ...].expand(1, 2, 3, 3), categories=["dog", "cat", "boat"])
print(l.to_categories())

print(OneHotLabel.from_category("cat", categories=["dog", "cat", "boat"]))
```
Output:
```
[[['dog', 'dog', 'boat'], ['dog', 'cat', 'boat']], [['dog', 'dog', 'cat'], ['cat', 'cat', 'boat']]]
[[['dog', 'cat', 'boat'], ['dog', 'cat', 'boat']]]
OneHotLabel([0, 1, 0])
```

Note: This PR is based upon https://github.com/pytorch/vision/pull/6419

### DIFF
--- a/torchvision/prototype/features/_label.py
+++ b/torchvision/prototype/features/_label.py
@@ -49,7 +49,7 @@ class Label(_Feature):
         return tree_map(lambda idx: self.categories[idx], self.tolist())
 
 
-class OneHotLabel(_Feature):
+class OneHotLabel(Label):
     categories: Optional[Sequence[str]]
 
     def __new__(
@@ -77,3 +77,21 @@ class OneHotLabel(_Feature):
         return super().new_like(
             other, data, categories=categories if categories is not None else other.categories, **kwargs
         )
+
+    @classmethod
+    def from_category(
+        cls,
+        category: str,
+        *,
+        categories: Sequence[str],
+        **kwargs: Any,
+    ) -> Label:
+        ohe_tensor = torch.zeros(len(categories), dtype=torch.long)
+        ohe_tensor[categories.index(category)] = 1
+        return cls(ohe_tensor, categories=categories, **kwargs)
+
+    def to_categories(self) -> Any:
+        if self.categories is None:
+            raise RuntimeError("OneHotLabel does not have categories")
+        label = self.argmax(dim=-1)
+        return tree_map(lambda idx: self.categories[idx], label.tolist())
--- a/torchvision/prototype/features/_label.py
+++ b/torchvision/prototype/features/_label.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Any, cast, Optional, Sequence, Union
+from typing import Any, Optional, Sequence, Union
 
 import torch
 from torch.utils._pytree import tree_map
--- a/torchvision/prototype/features/_label.py
+++ b/torchvision/prototype/features/_label.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 from typing import Any, cast, Optional, Sequence, Union
 
 import torch
-from torchvision.prototype.utils._internal import apply_recursively
+from torch.utils._pytree import tree_map
 
 from ._feature import _Feature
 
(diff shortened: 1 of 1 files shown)


torchvision/prototype/features/_label.py
```diff
@@ -77,3 +77,21 @@ def new_like(
...
+        **kwargs: Any,
+    ) -> Label:
+        ohe_tensor = torch.zeros(len(categories), dtype=torch.long)
+        ohe_tensor[categories.index(category)] = 1
+        return cls(ohe_tensor, categories=categories, **kwargs)
+
+    def to_categories(self) -> Any:
+        if self.categories is None:
+            raise RuntimeError("OneHotLabel does not have categories")
+        label = self.argmax(dim=-1)
```
pmeier, line 96: Is this an assumption we can safely make? IMO, that is very case specific. Imagine we have 60% dog and 40% cat. Do we really want to say this is 100% a dog for all cases? I left this method out when implementing `OneHotLabel` due to this issue.

torchvision/prototype/features/_label.py
```diff
@@ -43,13 +43,13 @@ def from_category(
...
+        if self.categories is None:
+            raise RuntimeError("Label does not have categories")
 
-        return apply_recursively(lambda idx: cast(Sequence[str], self.categories)[idx], self.tolist())
+        return tree_map(lambda idx: self.categories[idx], self.tolist())
 
 
-class OneHotLabel(_Feature):
+class OneHotLabel(Label):
     categories: Optional[Sequence[str]]
```
pmeier, line 53: This can probably go if we subclass.

torchvision/prototype/features/_label.py
vfdev-5: Ground truth is rarely presented as percentages of one class and another. If we have multilabel task (vs multiclass), think of music tags for songs for example, then I'd create a MultiLabel class for that. OneHotLabel is one-hot encoded label which is produced as int (class index) -> vector (0, ..., 0, 1, 0, ... 0), thus we can safely pass back from vector to int. Am I missing something ?

vfdev-5: So, technically it is not a one-hot-encoded label :)

datumbox: Good points.

Philip is right that this won't work very well for multi-class or soft-labels. Victor is also right to say "those 2 cases are not one-hot". 

But isn't our intention to use the specific class to store the info both for multi and soft-label cases? Isn't this what we do with MixUp and CutMix already?

So here I see two problems:
1. Probably Philip is right, those methods make assumptions that we might not make.
2. Victor is right, the name on this class is misleading.

pmeier: I agree, the naming is not optimal. Any ideas on how to improve? Do these concepts, i.e. representing the label through either an index or a percentage, have names in the literature?

datumbox: K-hot encoded for multi-label is used sometimes. Soft-assignments/labels for things that add to 100%. Off the top of my head not great names. I'm open to suggestions

vfdev-5:
Closing the PR as no-go. OneHotLabels should be handled differently than Labels, so it was decided not to subclass OneHotLabels from Labels.

pmeier:
One possible option to avoid code duplication is to have a `_LabelBase` base class. That way neither `isinstance(Label(), OneHotLabel)` nor `isinstance(OneHotLabel(), Label)` will be true, but they can still share some functionality, e.g the categories logic.

datumbox:
I'm OK if you wanna do this as a code quality exercise. We shouldn't expose the `_LabelBase` in the transforms though to replace the `Label` and `OneHotLabel` checks.

PR #6423, opened 2022-08-15
Use dist path when uploading using anaconda

atalman:
Fix the binary location path
Refer to: https://github.com/pytorch/audio/pull/2581/files


weiwangmeta (approved):
The fix looks good, thx!

Issue #6426, opened 2022-08-16
Cannot install from git using pip


vedantroy:
### 🐛 Describe the bug

I'm trying to install torchvision from source following this thread: https://github.com/pytorch/vision/issues/3283.
I tried doing `pip install 'git+https://github.com/pytorch/vision.git@v0.13.0'`, but it gives me the error:

```
Collecting git+https://github.com/pytorch/vision.git@v0.13.0
  Cloning https://github.com/pytorch/vision.git (to revision v0.13.0) to /tmp/pip-req-build-32tizbmz
  Running command git clone --filter=blob:none --quiet https://github.com/pytorch/vision.git /tmp/pip-req-build-32tizbmz
  Running command git checkout -q da3794e90c7cf69348f5446471926729c55f243e
  Resolved https://github.com/pytorch/vision.git to commit da3794e90c7cf69348f5446471926729c55f243e
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  error: subprocess-exited-with-error
  
  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [17 lines of output]
      Traceback (most recent call last):
        File "/home/vedant/miniconda3/envs/video-rec/lib/python3.9/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 363, in <module>
          main()
        File "/home/vedant/miniconda3/envs/video-rec/lib/python3.9/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 345, in main
          json_out['return_val'] = hook(**hook_input['kwargs'])
        File "/home/vedant/miniconda3/envs/video-rec/lib/python3.9/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 130, in get_requires_for_build_wheel
          return hook(config_settings)
        File "/tmp/pip-build-env-8an9cnsq/overlay/lib/python3.9/site-packages/setuptools/build_meta.py", line 338, in get_requires_for_build_wheel
          return self._get_build_requires(config_settings, requirements=['wheel'])
        File "/tmp/pip-build-env-8an9cnsq/overlay/lib/python3.9/site-packages/setuptools/build_meta.py", line 320, in _get_build_requires
          self.run_setup()
        File "/tmp/pip-build-env-8an9cnsq/overlay/lib/python3.9/site-packages/setuptools/build_meta.py", line 482, in run_setup
          super(_BuildMetaLegacyBackend,
[...]

PR #6427, opened 2022-08-16
Add note about --no-build-isolation flag in readme

NicolasHug:
Added a note about the need for `--no-build-isolation` when installing from source with `pip`. This just happened in https://github.com/pytorch/vision/issues/6426 but this isn't the first issue of that kind.


jdsgomes (approved):
LGTM

PR #6432, opened 2022-08-17
Free avPacket on EAGAIN decoder error

jdsgomes:
Summary:
According to the documentation the packet has to be freed after `av_read_frame()` call.
```
 If pkt->buf is NULL, then the packet is valid until the next
 av_read_frame() or until avformat_close_input(). Otherwise the packet
 is valid indefinitely. In both cases the packet must be freed with
 av_packet_unref when it is no longer needed.
```

Differential Revision: D38747612


NicolasHug:
@jdsgomes I assume this PR was from the "Export to GitHub" interface on fbcode, maybe it can be closed? The corresponding diff was landed and the sync PR is https://github.com/pytorch/vision/pull/6443

datumbox:
@NicolasHug That's that it is indeed. I spoke earlier with Joao and this came from the internal diff. Closing as you suggested.

PR #6433, opened 2022-08-17
Prototype references/classification

pmeier:
I don't want to merge this PR. This more like a feature branch that we can discuss on. For the actual port we can either cleanup this PR or use it as a starting point for another.


torchvision/prototype/transforms/__init__.py
```diff
@@ -3,7 +3,7 @@
 from ._transform import Transform  # usort: skip
 
 from ._augment import RandomCutmix, RandomErasing, RandomMixup
-from ._auto_augment import AugMix, AutoAugment, RandAugment, TrivialAugmentWide
+from ._auto_augment import AugMix, AutoAugment, AutoAugmentPolicy, RandAugment, TrivialAugmentWide
```
vfdev-5, line 6: If this is missing in `main`, please add this in a separate PR

references/classification/train.py
```diff
@@ -7,14 +7,20 @@
...
-import transforms
 import utils
 from sampler import RASampler
 from torch import nn
 from torch.utils.data.dataloader import default_collate
+from torchvision.prototype import features, transforms
 from torchvision.transforms.functional import InterpolationMode
 
 
+class WrapIntoFeatures(nn.Module):
```
datumbox, line 18: Nit: I would move this to the old local `transforms.py` file you just deleted. We might need more like this.

references/classification/train.py
```diff
@@ -200,14 +208,21 @@ def main(args):
...
     if args.cutmix_alpha > 0.0:
-        mixup_transforms.append(transforms.RandomCutmix(num_classes, p=1.0, alpha=args.cutmix_alpha))
-    if mixup_transforms:
-        mixupcutmix = torchvision.transforms.RandomChoice(mixup_transforms)
-        collate_fn = lambda batch: mixupcutmix(*default_collate(batch))  # noqa: E731
+        mixup_or_cutmix.append(transforms.RandomCutmix(alpha=args.cutmix_alpha))
+    if mixup_or_cutmix:
+        batch_transform = transforms.Compose(
+            [
+                WrapIntoFeatures(),
```
datumbox, line 219: Don't we have to WrapIntoFeatures unconditionally from whether we use mixup/cutmix?

torchvision/prototype/transforms/__init__.py
datumbox: Yes please, let's fix this elsewhere. It's likely we won't actually merge this PR but rather use it only for training.

datumbox (commented):
Looks good overall. Just a few comments. Feel free to ignore if it's too early.

references/classification/train.py
pmeier: The problem is that `default_collate` does not respect tensor subclasses. Since we use this transform afterwards, we need to wrap here. Of course we can also wrap before, but it is not necessary since the input is a plain tensor that defaults to images and an integer which is completely ignored.

torchvision/prototype/transforms/__init__.py
pmeier: #6411.

references/classification/train.py
datumbox: My question is, what if I don't use mixup or cutmix? Shouldn't we wrap the data into `features.Image` anyway? I might be missing something here. My main point is that since we are testing the new API, we should probably wrap all inputs using their appropriate types and see how the new kernels behave (Rather than relying on their default/legacy pure tensor implementations).

pmeier: We can only wrap after we have converted from PIL. This happens fairly late in the transform:

https://github.com/pytorch/vision/blob/b83d5f7c4b657235ff94da52970330399f344048/references/classification/presets.py#L33

I remember @vfdev-5 noting that on the CPU PIL kernels are faster (I don't remember if there was a special case or other constraints; please fill the blanks). Thus, if we want to optimize for speed, we should probably leave it as is. No strong opinion though.

pmeier:
I've added the needed changes for the detection references.

references/detection/coco_utils.py
```diff
@@ -1,30 +1,12 @@
...
 import torch
 import torch.utils.data
 import torchvision
-import transforms as T
+
 from pycocotools import mask as coco_mask
 from pycocotools.coco import COCO
-
-
-class FilterAndRemapCocoCategories:
```
pmeier, line 12: This was not used anywhere.

references/classification/train.py
```diff
@@ -200,14 +205,21 @@ def main(args):
...
-        mixupcutmix = torchvision.transforms.RandomChoice(mixup_transforms)
-        collate_fn = lambda batch: mixupcutmix(*default_collate(batch))  # noqa: E731
+        mixup_or_cutmix.append(transforms.RandomCutmix(alpha=args.cutmix_alpha, p=1.0))
+    if mixup_or_cutmix:
+        batch_transform = transforms.Compose(
+            [
+                WrapIntoFeatures(),
+                transforms.LabelToOneHot(num_categories=num_classes),
+                transforms.ToDtype(torch.float, features.OneHotLabel),
+                transforms.RandomChoice(*mixup_or_cutmix),
```
datumbox, line 219:
```suggestion
                transforms.RandomChoice(mixup_or_cutmix),
```

Otherwise you get:
```
  File "./vision/references/classification/train.py", line 219, in main
    transforms.RandomChoice(*mixup_or_cutmix),
  File "./torchvision/prototype/transforms/_container.py", line 47, in __init__

```

references/detection/coco_utils.py
```diff
@@ -100,6 +83,27 @@ def __call__(self, image, target):
...
+        wrapped_target = dict(
+            boxes=features.BoundingBox(
+                target["boxes"],
+                format=features.BoundingBoxFormat.XYXY,
+                image_size=(image.height, image.width),
+            ),
+            # TODO: add categories
+            labels=features.Label(target["labels"], categories=None),
+            masks=features.SegmentationMask(target["masks"]),
+            image_id=int(target["image_id"]),
```
datumbox, line 99: This wont work with:
https://github.com/pytorch/vision/blob/59ec1dfd550652a493cb99d5704dcddae832a204/references/detection/engine.py#L27-L29

and leads to:
Unfortunately using the scripts leads to errors:
```
  File "./vision/references/detection/train.py", line 283, in main
    train_one_epoch(model, optimizer, data_loader, device, epoch, args.print_freq, scaler)
  File "./vision/references/detection/engine.py", line 29, in train_one_epoch
    targets = [{k: v.to(device) for k, v in t.items()} for t in targets]
  File "./vision/references/detection/engine.py", line 29, in <listcomp>
    targets = [{k: v.to(device) for k, v in t.items()} for t in targets]
  File "./vision/references/detection/engine.py", line 29, in <dictcomp>
    targets = [{k: v.to(device) for k, v in t.items()} for t in targets]
AttributeError: 'int' object has no attribute 'to'
```

datumbox (commented):
@pmeier There seem to be a few issues with the scripts. See below. Might be worth doing a dummy run with very little data to confirm they work.

pmeier:
I've run the references with the following parameters to confirm they work:

- Classification:

    ```py
    [
        "--device=cpu",
        "--batch-size=2",
        "--epochs=1",
        "--workers=2",
        "--mixup-alpha=0.5",
        "--cutmix-alpha=0.5",
        "--auto-augment=ra",  # "ra", "ta_wide", "augmix", "imagenet", "cifar10", "svhn"
        "--random-erase=1.0",
    ]
    ```

- Detection:

    ```py
    [
        "--device=cpu",
        "--batch-size=2",
        "--epochs=1",
        "--workers=2",
        "--data-augmentation=hflip",  # "hflip", "lsj", "multiscale", "ssd", "ssdlite"
        # "--use-copypaste",  # if data_augmention == "lsj"
    ]
    ```

pmeier:
Detection references are affected by #6528. Do not train before this is merged.

datumbox:
I'm doing the following runs to confirm the validity.

## Classification

### Augmentation: ta_wide + random erasing + mixup + cutmix

Using githash 4d73fe7:

```
PYTHONPATH=$PYTHONPATH:`pwd` python -u run_with_submitit.py --ngpus 8 --nodes 1 --model resnet50 --batch-size 128 --lr 0.5 --lr-scheduler cosineannealinglr --lr-warmup-epochs 5 --lr-warmup-method linear --auto-augment ta_wide --epochs 600 --random-erase 0.1 --label-smoothing 0.1 --mixup-alpha 0.2 --cutmix-alpha 1.0 --weight-decay 0.00002 --norm-weight-decay 0.0 --train-crop-size 176 --model-ema --val-resize-size 232 --ra-sampler --ra-reps 4 --data-path /datasets01_ontap/imagenet_full_size/061417/
```

## Detection

### [DONE] Augmentation: multiscale

Using githash 7cb08d5:

```
PYTHONPATH=$PYTHONPATH:`pwd` python -u run_with_submitit.py --ngpus 8 --nodes 1 --weights-backbone ResNet50_Weights.IMAGENET1K_V2 --dataset coco --model retinanet_resnet50_fpn_v2 --opt adamw --lr 0.0001 --epochs 26 --lr-steps 16 22 --weight-decay 0.05 --norm-weight-decay 0.0 --data-augmentation multiscale --sync-bn --data-path /datasets01_ontap/COCO/022719/

IoU metric: bbox
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.414
 Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=100 ] = 0.616
 Average Precision  (AP) @[ IoU=0.75      | area=   all | maxDets=100 ] = 0.438
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.258
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.454
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.535
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=  1 ] = 0.338
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets= 10 ] = 0.545
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.588
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.419
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.623
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.728
```

The accuracy is as expected.

### Augmentation: lsj + copypaste

Using githash ec120ff:

```
PYTHONPATH=$PYTHONPATH:`pwd` python -u run_with_submitit.py --ngpus 8 --nodes 2 --dataset coco --model maskrcnn_resnet50_fpn_v2 --epochs 400 --lr-steps 352 384 --lr 0.1 --batch-size 4 --weight-decay 0.00004 --sync-bn --data-augmentation lsj --use-copypaste --data-path /datasets01_ontap/COCO/022719/
```

### Augmentation: ssdlite

Using githash ec120ff:

```
PYTHONPATH=$PYTHONPATH:`pwd` python -u run_with_submitit.py --ngpus 8 --nodes 1 --weights-backbone MobileNet_V3_Large_Weights.IMAGENET1K_V1 --dataset coco --model ssdlite320_mobilenet_v3_large --aspect-ratio-group-factor 3 --epochs 660 --lr-scheduler cosineannealinglr --lr 0.15 --batch-size 24 --weight-decay 0.00004 --data-augmentation ssdlite --data-path /datasets01_ontap/COCO/022719/
```

### Augmentation: ssd

Using githash ec120ff:

```
PYTHONPATH=$PYTHONPATH:`pwd` python -u run_with_submitit.py --ngpus 8 --nodes 1 --weights-backbone VGG16_Weights.IMAGENET1K_FEATURES --dataset coco --model ssd300_vgg16 --epochs 120 --lr-steps 80 110 --aspect-ratio-group-factor 3 --lr 0.002 --batch-size 4 --weight-decay 0.0005 --trainable-backbone-layers 5 --data-augmentation ssd --data-path /datasets01_ontap/COCO/022719/
```

references/classification/train.py
```diff
@@ -200,14 +205,21 @@ def main(args):
...
     if args.cutmix_alpha > 0.0:
-        mixup_transforms.append(transforms.RandomCutmix(num_classes, p=1.0, alpha=args.cutmix_alpha))
-    if mixup_transforms:
-        mixupcutmix = torchvision.transforms.RandomChoice(mixup_transforms)
-        collate_fn = lambda batch: mixupcutmix(*default_collate(batch))  # noqa: E731
+        mixup_or_cutmix.append(transforms.RandomCutmix(alpha=args.cutmix_alpha, p=1.0))
+    if mixup_or_cutmix:
+        batch_transform = transforms.Compose(
+            [
+                WrapIntoFeatures(),
```
datumbox, line 216: Right now we are testing the PIL backend. In order to test the `features.Image` one, we will need to move this at the beginning (aka on the transforms defined for ImageFolder), right?

datumbox:
Due to the bug discovered at #6541, we need to repeat the experiments that involve the LSJ.

### Augmentation: lsj + copypaste

Target Acc: 0.473 / 0.417 - time: 3 days, 19:00:54 - jobid: experiments/PR5825/22644

Using githash 4699e55:

```
PYTHONPATH=$PYTHONPATH:`pwd` python -u run_with_submitit.py --ngpus 8 --nodes 4 --dataset coco --model maskrcnn_resnet50_fpn_v2 --epochs 600 --lr-steps 540 570 585 --lr 0.32 --batch-size 8 --weight-decay 0.00004 --sync-bn --data-augmentation lsj --use-copypaste --data-path /datasets01_ontap/COCO/022719/
Submitted job_id: 61614
```

### Augmentation: lsj

Target Acc: 0.473 / 0.416 - time: 4 days, 6:43:29 - jobid: experiments/PR5444/34922

Using githash 4699e55:

```
PYTHONPATH=$PYTHONPATH:`pwd` python -u run_with_submitit.py --ngpus 8 --nodes 4 --dataset coco --model maskrcnn_resnet50_fpn_v2 --epochs 400 --lr-steps 352 384 --lr 0.1 --batch-size 2 --weight-decay 0.00004 --sync-bn --data-augmentation lsj --data-path /datasets01_ontap/COCO/022719/
Submitted job_id: 61615
```

datumbox:
@pmeier @vfdev-5 I think we just confirmed that the v2 produces the same model accuracy as v1. GGs! :)

references/segmentation/train.py
```diff
@@ -134,8 +135,8 @@ def main(args):
     else:
         torch.backends.cudnn.benchmark = True
 
-    dataset, num_classes = get_dataset(args.data_path, args.dataset, "train", get_transform(True, args))
-    dataset_test, _ = get_dataset(args.data_path, args.dataset, "val", get_transform(False, args))
+    dataset, num_classes = get_dataset(args.data_path, args.dataset, "train", get_transform(train=True, args=args))
+    dataset_test, _ = get_dataset(args.data_path, args.dataset, "val", get_transform(train=False, args=args))
```
pmeier, line 139: This has nothing to do with the new transforms, but I despise calling functions with positional bool arguments if it isn't abundantly what it means. That is not the case here. Using keyword arguments makes this much easier to understand.

references/segmentation/train.py
```diff
@@ -36,7 +36,8 @@ def get_transform(train, args):
         weights = torchvision.models.get_weight(args.weights)
         trans = weights.transforms()
 
-        def preprocessing(img, target):
+        def preprocessing(sample):
+            img, target = sample
```
pmeier, line 40: I didn't want to touch this any more than this, because that would mean I would need to mess with the presets transform on the new model weights. IIUC this is only for testing against pre-trained weights and thus out of scope of this preliminary test anyways.

references/segmentation/presets.py
```diff
@@ -1,39 +1,40 @@
...
+        image, segmentation_mask = sample
+        return image, features.SegmentationMask(segmentation_mask.squeeze(0), dtype=torch.int64)
+
 
-        trans = [T.RandomResize(min_size, max_size)]
+class SegmentationPresetTrain(T.Compose):
+    def __init__(self, *, base_size, crop_size, hflip_prob=0.5, mean=(0.485, 0.456, 0.406), std=(0.229, 0.224, 0.225)):
+        transforms = [
+            T.ToImageTensor(),
+            WrapIntoFeatures(),
```
pmeier, line 15: In contrast to the classification and detection references, we need to wrap before all transformation. This is now needed because the new transformation would take the mask as image otherwise.

pmeier (commented):
I've run the segmentation script with the following args for a few iterations without issues:

```py
[
    "--dataset=voc_aug",  # "coco", "voc", "voc_aug"
    "--device=cpu",
    "--batch-size=2",
    "--epochs=1",
    "--workers=4",
]
```

references/segmentation/presets.py
```diff
@@ -1,39 +1,40 @@
...
 
 
-class SegmentationPresetTrain:
-    def __init__(self, *, base_size, crop_size, hflip_prob=0.5, mean=(0.485, 0.456, 0.406), std=(0.229, 0.224, 0.225)):
-        min_size = int(0.5 * base_size)
-        max_size = int(2.0 * base_size)
+class WrapIntoFeatures(T.Transform):
+    def forward(self, sample):
+        image, segmentation_mask = sample
+        return image, features.SegmentationMask(segmentation_mask.squeeze(0), dtype=torch.int64)
```
datumbox: Why aren't we wrapping Image? This is what we currently do on the Classification references.

references/segmentation/presets.py
datumbox: Wrapping isn't the problem. Calling `ToImageTensor` first is I think. This is because if we run this check, we will be hitting the Tensor backend and not the PIL like we did with the previous tests. Everything should be tested obviously, but I wonder whether we can move the tensor conversion at the end?

references/segmentation/train.py
datumbox: Do you then mean that this wont work any more? If it works, I'm fine with it.

references/segmentation/train.py
datumbox: Sounds good.

references/segmentation/transforms.py
```diff
@@ -1,100 +1,20 @@
...
+        if not isinstance(inpt, features.SegmentationMask):
+            return super()._transform(inpt, params)
+
+        # `SegmentationMask`'s should be padded with 255 to indicate an area that should not be used in the loss
+        # calculation. See
+        # https://stackoverflow.com/questions/49629933/ground-truth-pixel-labels-in-pascal-voc-for-semantic-segmentation
+        # for details.
+        # FIXME: Using different values for `fill` based on the input type is not supported by `transforms.RandomCrop`.
+        #  Thus, we emulate it here. See https://github.com/pytorch/vision/issues/6568.
+        fill = self.fill
```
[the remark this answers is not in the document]
datumbox: Oh boy that's ugly... But it works... For now... :D

datumbox (commented):
@pmeier just a couple of comments on Segmentation. Otherwise it looks good.

references/segmentation/presets.py
pmeier: The `ToImageTensor` already wraps into a `features.Image`:

https://github.com/pytorch/vision/blob/24890d718f5a73586ef093371912b5b37a5b0d46/torchvision/prototype/transforms/functional/_type_conversion.py#L30

Thus, `segmentation_mask` here is already a `features.Image` and we just re-wrap in the correct type. Maybe `WrapIntoFeatures` is a bad name. Let's resolve this together with https://github.com/pytorch/vision/pull/6433#discussion_r970876270.

references/segmentation/presets.py
pmeier: Yes, that should be doable. I'll send a patch.

references/segmentation/presets.py
datumbox: Sounds good. Ping me when you are OK to check and start using it. Thanks for the work!

references/segmentation/train.py
```diff
@@ -11,15 +11,18 @@
...
 from torch.optim.lr_scheduler import PolynomialLR
-from torchvision.transforms import functional as F, InterpolationMode
+from torchvision.prototype.transforms import Compose, functional as F, InterpolationMode
 
 
 def get_dataset(dir_path, name, image_set, transform):
-    def sbd(*args, **kwargs):
-        return torchvision.datasets.SBDataset(*args, mode="segmentation", **kwargs)
+    def voc(*args, transforms, **kwargs):
+        return torchvision.datasets.VOCSegmentation(*args, transforms=Compose([transforms]), **kwargs)
```
pmeier, line 19: *This comment references some other parts of this PR. I'll leave some pointers below.*

---

This `Compose` looks redundant, but it isn't. The old, i.e. custom segmentation compose, was implemented like

```py
def forward(image, target):
    for transform in self.transforms:
        image, target = transform(image, target)
```

while the new roughly does

```py
def forward(sample):
    for transform in self.transforms:
        sample = transform(sample)
```

This in itself is not an issue for the new transforms since `transform(image, target)` and `transform((image, target))` are equivalent. However, if we run with `--weights` and `--test-only`, we define a `preprocessing` function that acts like a transformation. This function cannot deal with both types of inputs and thus we have to align the outputs of the different datasets.

Since COCO uses a `Compose` internally, I've opted to also use one here.

references/segmentation/transforms.py
```diff
@@ -1,100 +1,19 @@
...
-
-
-class Compose:
-    def __init__(self, transforms):
-        self.transforms = transforms
-
-    def __call__(self, image, target):
-        for t in self.transforms:
-            image, target = t(image, target)
-        return image, target
```
pmeier, line 26: Pointer for comment above. Resolve together.

references/segmentation/train.py
```diff
@@ -36,11 +39,12 @@ def get_transform(train, args):
         weights = torchvision.models.get_weight(args.weights)
         trans = weights.transforms()
 
-        def preprocessing(img, target):
+        def preprocessing(sample):
+            img, target = sample
```
pmeier, line 43: Pointer for comment above. Resolve together.

references/segmentation/coco_utils.py
```diff
@@ -90,7 +94,6 @@ def get_coco(root, image_set, transforms):
     PATHS = {
         "train": ("train2017", os.path.join("annotations", "instances_train2017.json")),
         "val": ("val2017", os.path.join("annotations", "instances_val2017.json")),
-        # "train": ("val2017", os.path.join("annotations", "instances_val2017.json"))
     }
     CAT_LIST = [0, 5, 2, 16, 9, 44, 6, 3, 17, 62, 21, 67, 18, 19, 4, 1, 64, 20, 63, 7, 72]
```
pmeier, line 99: Pointer for comment above. Referenced `Compose` is one line below this. Resolve together.

pmeier (commented):
I've resolved the early PIL to tensor conversion by only wrapping the segmentation mask early on. This way, the image stays as PIL image until we hit the `ToImageTensor` transform similar to what we had before.

references/segmentation/train.py
```diff
@@ -36,11 +39,12 @@ def get_transform(train, args):
...
             img = trans(img)
-            size = F.get_dimensions(img)[1:]
+            size = F.get_image_size(img)
```
datumbox, line 45: Reminder to not use the `get_image_size()` transform. It returns (w, h) instead (h, w). See #6581
```suggestion
            size = F.get_dimensions(img)[1:]
```

references/segmentation/train.py
```diff
@@ -36,11 +39,12 @@ def get_transform(train, args):
...
             target = F.resize(target, size, interpolation=InterpolationMode.NEAREST)
-            return img, F.pil_to_tensor(target)
+            return img, F.to_image_tensor(target)
```
datumbox, line 47:
```suggestion
            return img, F.pil_to_tensor(target)
```

datumbox:
## Segmentation

Target Acc: 91.1 / 57.9 - time: 3:12:36 - jobid: jobs/PR3276/2nd_training/35343756

Using githash a2893a1:
```
PYTHONPATH=$PYTHONPATH:`pwd` python -u run_with_submitit.py --ngpus 8 --nodes 1 --dataset coco -b 4 --model lraspp_mobilenet_v3_large --lr 0.01 --wd 0.000001
Submitted job_id: 64765
Submitted job_id: 64766
Submitted job_id: 64768
```

references/segmentation/train.py
```diff
@@ -36,11 +39,12 @@ def get_transform(train, args):
...
         trans = weights.transforms()
 
-        def preprocessing(img, target):
+        def preprocessing(sample):
+            img, target = sample
             img = trans(img)
             size = F.get_dimensions(img)[1:]
             target = F.resize(target, size, interpolation=InterpolationMode.NEAREST)
-            return img, F.pil_to_tensor(target)
+            return img, F.to_image_tensor(target)
```
datumbox, line 47: Oops, this doesn't look right. This will return target as Image.
```suggestion
            return img, F.pil_to_tensor(target)
```

references/segmentation/transforms.py
```diff
@@ -1,100 +1,19 @@
...
-        self.mean = mean
-        self.std = std
-
-    def __call__(self, image, target):
-        image = F.normalize(image, mean=self.mean, std=self.std)
-        return image, target
+from torchvision.prototype import features, transforms
+
+
+class RandomCrop(transforms.RandomCrop):
```
datumbox, line 4: @vfdev-5 Let's remove this after you merge https://github.com/pytorch/vision/pull/6586

references/segmentation/presets.py
```diff
@@ -1,39 +1,42 @@
...
 
 
-class SegmentationPresetTrain:
-    def __init__(self, *, base_size, crop_size, hflip_prob=0.5, mean=(0.485, 0.456, 0.406), std=(0.229, 0.224, 0.225)):
-        min_size = int(0.5 * base_size)
-        max_size = int(2.0 * base_size)
+class WrapIntoFeatures(T.Transform):
+    def forward(self, sample):
+        image, mask = sample
+        return image, features.Mask(F.to_image_tensor(mask).squeeze(0), dtype=torch.int64)
```
datumbox, line 10: @pmeier Similarly here, I don't think we should be calling `to_image_tensor` which casts the input to an Image and then rewrap it to a Mask. I'll send a PR with a potential correction, to review.

references/segmentation/train.py
datumbox: Potential fix at https://github.com/pytorch/vision/pull/6588

references/segmentation/presets.py
```diff
@@ -1,39 +1,42 @@
...
+        ]
         if hflip_prob > 0:
-            trans.append(T.RandomHorizontalFlip(hflip_prob))
-        trans.extend(
+            transforms.append(T.RandomHorizontalFlip(hflip_prob))
+        transforms.extend(
             [
-                T.RandomCrop(crop_size),
-                T.PILToTensor(),
+                RandomCrop(crop_size, pad_if_needed=True),
```
datumbox, line 23: @pmeier I think that's not equivalent with before.

Previously we padded as follows:
https://github.com/pytorch/vision/blob/d367a01a18a3ae6bee13d8be3b63fd6a581ea46f/references/segmentation/transforms.py#L9-L16

Now we pad differently:
https://github.com/pytorch/vision/blob/08743385d93a0aa4145da6f6db49bfa00df148a3/torchvision/prototype/transforms/_geometry.py#L438-L444

This inconsistency could be the reason we get accuracy issues.

pmeier: Here is how the padding is done:

1. references: perform a minimum bottom-right pad. Meaning, only pad the minimum amount so that `crop` doesn't complain. If both dimensions need to be padded, the `RandomCrop` transform is a no-op.

2. transforms (stable or prototype): perform a pad on all directions with twice the minimum amount. For example, an image of size `(16, 16)` with a crop size of `(18, 18)` get padded with `2 * (18 - 16) == 4` on all sides to `(24, 24)` and only then the crop will be performed. This behavior is intentional, see #564, to always enable random cropping even if there is padding.

I don't think we can unify both transforms in a way that supports both behaviors. My suggestion is to add `_compute_padding(...)` method on `transforms.RandomCrop` that we can overwrite in the references.

Thoughts? @datumbox @vfdev-5

datumbox: @pmeier Thanks for your debugging, I agree with your findings. This change of behaviour is likely to be causing the accuracy regression reported above.

In an offline chat, @vfdev-5 suggested decomposing Pad and Crop to achieve a similar effect, which is a great idea. Your proposal to overwrite the `RandomCrop` in references is a viable approach as well. As you both seem to suggest, we should avoid making the `transforms.RandomCrop` more complicated than it is.

pmeier: I've introduced a `PadIfSmaller` transform in e912976a5b30018fda5cc50d9169fc1b697ac84b as discussed in #6591.

datumbox:
Due to a discrepancy discovered on the padding of `RandomCrop`, we rerun the experiments:

## Segmentation

Target Acc: 91.2 / 57.9 - time: 3:13:11 - jobid: jobs/PR3276/3rd_training/35354083

Using githash 5a311b3:
```
PYTHONPATH=$PYTHONPATH:`pwd` python -u run_with_submitit.py --ngpus 8 --nodes 1 --dataset coco -b 4 --model lraspp_mobilenet_v3_large --lr 0.01 --wd 0.000001
Submitted job_id: 67061
Submitted job_id: 67062
Submitted job_id: 67063
```

**Result:** PENDING

torchvision/transforms/functional_tensor.py
```diff
@@ -880,7 +880,7 @@ def _scale_channel(img_chan: Tensor) -> Tensor:
     if img_chan.is_cuda:
         hist = torch.histc(img_chan.to(torch.float32), bins=256, min=0, max=255)
     else:
-        hist = torch.bincount(img_chan.view(-1), minlength=256)
```
datumbox, line 883: Without this we get:
```
    hist = torch.bincount(img_chan.view(-1), minlength=256)
```

references/classification/train.py
```diff
@@ -210,16 +215,22 @@ def main(args):
...
+        mixup_or_cutmix.append(transforms.RandomMixup(alpha=args.mixup_alpha, p=1.0))
     if args.cutmix_alpha > 0.0:
-        mixup_transforms.append(transforms.RandomCutmix(num_classes, p=1.0, alpha=args.cutmix_alpha))
-    if mixup_transforms:
-        mixupcutmix = torchvision.transforms.RandomChoice(mixup_transforms)
+        mixup_or_cutmix.append(transforms.RandomCutmix(alpha=args.cutmix_alpha, p=1.0))
+    if mixup_or_cutmix:
+        batch_transform = transforms.Compose(
+            [
+                transforms.LabelToOneHot(num_categories=num_classes),
```
datumbox, line 226: I removed from this line the `WrapIntoFeatures()` call and now get:
```
TypeError: RandomCutmix() is only defined for tensor images and one-hot labels.
```

I expected that the `Label` is already annotated because the dataset is defined with:
```python
target_transform=lambda target: features.Label(target),
```

I'll check it out further on Monday.

datumbox:
# Tensor Backend

## Classification

### Augmentation: ta_wide + random erasing + mixup + cutmix

Target Acc: 80.854 / 95.428 - time: 1 day, 18:06:05 - jobid: experiments/PR5201/resnet50_sota2/13576

Using githash e9c480e:
```
PYTHONPATH=$PYTHONPATH:`pwd` python -u run_with_submitit.py --ngpus 8 --nodes 1 --model resnet50 --batch-size 128 --lr 0.5 --lr-scheduler cosineannealinglr --lr-warmup-epochs 5 --lr-warmup-method linear --auto-augment ta_wide --epochs 600 --random-erase 0.1 --label-smoothing 0.1 --mixup-alpha 0.2 --cutmix-alpha 1.0 --weight-decay 0.00002 --norm-weight-decay 0.0 --train-crop-size 176 --model-ema --val-resize-size 232 --ra-sampler --ra-reps 4 --data-path /datasets01_ontap/imagenet_full_size/061417/
Submitted job_id: 67798
```

### Augmentation: aa + random erasing

Target Acc: 67.620 / 87.404 - time: 2 days, 17:09:20 - jobid: jobs/PR3354/classification/35753749

Using githash e9c480e:
```
PYTHONPATH=$PYTHONPATH:`pwd` python -u run_with_submitit.py --ngpus 8 --nodes 1 --model mobilenet_v3_small --epochs 600 --opt rmsprop --batch-size 128 --lr 0.064 --wd 0.00001 --lr-step-size 2 --lr-gamma 0.973 --auto-augment imagenet --random-erase 0.2 --data-path /datasets01_ontap/imagenet_full_size/061417/
Submitted job_id: 67799
```

## Detection

### Augmentation: multiscale

Target Acc: 0.415 - time: 9:54:20 - jobid: experiments/PR5444/34702

Using githash e9c480e:
```
PYTHONPATH=$PYTHONPATH:`pwd` python -u run_with_submitit.py --ngpus 8 --nodes 1 --weights-backbone ResNet50_Weights.IMAGENET1K_V2 --dataset coco --model retinanet_resnet50_fpn_v2 --opt adamw --lr 0.0001 --epochs 26 --lr-steps 16 22 --weight-decay 0.05 --norm-weight-decay 0.0 --data-augmentation multiscale --sync-bn --data-path /datasets01_ontap/COCO/022719/
Submitted job_id: 67794
```

### Augmentation: ssdlite

Target Acc: 0.212 - time: 1 day, 4:11:22 - jobid: jobs/PR3757/2nd_training/41046786

Using githash e9c480e:
```
PYTHONPATH=$PYTHONPATH:`pwd` python -u run_with_submitit.py --ngpus 8 --nodes 1 --dataset coco --model ssdlite320_mobilenet_v3_large --aspect-ratio-group-factor 3 --epochs 660 --lr-scheduler cosineannealinglr --lr 0.15 --batch-size 24 --weight-decay 0.00004 --data-augmentation ssdlite --data-path /datasets01_ontap/COCO/022719/
Submitted job_id: 67795
```

### Augmentation: ssd

Target Acc: 0.251 - time: 1 day, 3:40:14 - jobid: jobs/PR3403/4th_training/40773612

Using githash e9c480e:
```
PYTHONPATH=$PYTHONPATH:`pwd` python -u run_with_submitit.py --ngpus 8 --nodes 1 --weights-backbone VGG16_Weights.IMAGENET1K_FEATURES --dataset coco --model ssd300_vgg16 --epochs 120 --lr-steps 80 110 --aspect-ratio-group-factor 3 --lr 0.002 --batch-size 4 --weight-decay 0.0005 --trainable-backbone-layers 5 --data-augmentation ssd --data-path /datasets01_ontap/COCO/022719/
Submitted job_id: 67796
```

### Augmentation: lsj + copypaste

Target Acc: 0.473 / 0.417 - time: 3 days, 19:00:54 - jobid: experiments/PR5825/22644

Using githash e9c480e:
```
PYTHONPATH=$PYTHONPATH:`pwd` python -u run_with_submitit.py --ngpus 8 --nodes 4 --dataset coco --model maskrcnn_resnet50_fpn_v2 --epochs 600 --lr-steps 540 570 585 --lr 0.32 --batch-size 8 --weight-decay 0.00004 --sync-bn --data-augmentation lsj --use-copypaste --data-path /datasets01_ontap/COCO/022719/
Submitted job_id: 67791
```

## Segmentation with pretrained backbones

Target Acc: 90.7 / 55.5 - time: 2:25:39 - jobid: experiments/PR6433/67312

Using githash e9c480e:
```
PYTHONPATH=$PYTHONPATH:`pwd` python -u run_with_submitit.py --ngpus 8 --nodes 1 --dataset coco -b 4 --model lraspp_mobilenet_v3_large --lr 0.01 --wd 0.000001 --weights-backbone MobileNet_V3_Large_Weights.IMAGENET1K_V1
Submitted job_id: 67797
```

references/classification/presets.py
```diff
@@ -18,10 +18,11 @@ def __init__(
...
     ):
-        trans = (
-            [transforms.RandomResizedCrop(crop_size, interpolation=interpolation)]
+        trans = [transforms.ToImageTensor()]
+        trans.append(
+            transforms.RandomResizedCrop(crop_size, interpolation=interpolation, antialias=True)
             if center_crop
-            else [transforms.CenterCrop(crop_size)]
+            else transforms.CenterCrop(crop_size)
         )
```
datumbox, line 26: Adapted cherrypick from #6642

references/classification/train.py
pmeier: Yeah, that is a known issue. The `default_collate` internally uses [`torch.stack`](https://github.com/pytorch/pytorch/blob/9c036aa112b0a8fd9afb824d1fda058e2b66ba1d/torch/utils/data/_utils/collate.py#L143). Since we only automatically wrap on `.to()` and `.clone()` this will drop the feature type.

Three options here:

1. Allow `torch.stack` to retain the feature type. I would not go that route, since we cannot know in general if the result of a stack is still the feature we started out with. We at least need more elaborate checks that make sure this is happening.
2. Fix `default_collate` to handle our feature types. It is a possible solution, but won't happen, because the feature types are `torchvision` only and `default_collate` is a general solution.
3. Provide a custom collate like `vision_collate`. This would mean that depending on the taks, users can no longer go with the default value in the `DataLoader`, but judging from our references, this seems to be somewhat common. Plus, this `vision_collate` could also include more functionality that we need:
    1. Don't `torch.stack` images if we find `BoundingBox`'es.
    2. Support for `None` that some of our datasets already return (#5233) and will be heavily utilized by the new datasets for missing data in a configuration.

My vote is out for 3. To avoid rewriting `default_collate` from scratch in `torchvision` and thus risking it going out of sync with the core version, I made a push to make `default_collate` more extensible in [this gist](https://gist.github.com/pmeier/da684e84283113ed6e427d6519d21efe). @ejguan wanted to handle this from the `torchdata` side. Any progress there?

datumbox: It's a hard problem. I agree option 2 is not possible. What I don't like about option 3 is that if TorchVision has it's own collate, it will miss out on potential improvements from Core/Data. I also wouldn't put it inside the library but rather in the examples.

Option 1 sounds promising only if we feel we can do all the necessary checks to avoid gotchas (+ think of whether this is a performance bottleneck). If that's the case, we could offer implementations that whitelist the `stack` similary to `to()` and `clone()`. But if checking all the TV meta-data + Tensor meta-data is too slow, then we might leave it to the user to decide how to handle it efficiently on a case by case basis (as you have it here). I'm also open exploring other ideas to rewrap the types, though nothing elegant pops to my mind right now.

references/classification/presets.py
datumbox: Looks like the `center_crop` was unused and buggy. We'll need to remove it later.

datumbox: The issue is now fixed on main. I've cleaned up the code and merged it here. Unfortunately due to this bug, I had to restart the runs on the Classification models, so it's going to take longer to verify that the accuracies match.

references/classification/train.py
ejguan: Talked to Vitaly about it a while ago and we thought adding argument to `default_collate` might not be the ideal option because users can always do it in a reversed way by providing custom collate function and call `default_collate` if needed.

Would it be possible if we let `default_collate` to do collate by checking if the object is attached with a specific method (tentative: `collate_fn`)?
```py
class CustomObject:
   @staticmethod
    def collate_fn(batch):
        return ...

def default_collate(batch:
    elem = batch[0]
    ...
    if hasattr(elem, "collate_fn"):
        return elem.collate_fn(batch)
```
This would provide the [same mechanism](https://github.com/pytorch/pytorch/blob/6a04df3ac85714b33dcb2af20d0eeea96d131d75/torch/utils/data/_utils/pin_memory.py#L72-L73) as `pin_memory`

pmeier: > users can always do it in a reversed way by providing custom collate function and call `default_collate` if needed.

That is not true. `default_collate` does two things:

1. Handle containers such as lists, tuples, dictionaries, and so on, by recursively calling itself.
2. Handle "primitive batches", by `torch.stack`'ing tensors, putting integers and floats in a list and so on.

What we need is to modify 2. without touching 1. Since these two parts are inseparable at the moment, we cannot call `default_collate` to do 1. for us, because it would do the wrong "primitive batch" handling. Thus, we have to rewrite the functionality of 1. from scratch and can only call `default_collate` for the "primitive batches" that we don't want to handle differently.

> we thought adding argument to `default_collate` might not be the ideal option

My proposal is not advocating for that. I introduced a `collate` function that only does 1. from above and is flexible on 2. The old `default_collate` will just `functools.partial` of from it and thus not need / expose any extra parameters.

pmeier: As for adding a `collate_fn` to our feature type, I'm not really a fan. It feels a little weird to have this logic on the object, but I guess it is easier than to allow `torch.stack` to retain the feature type (option 1. from above).

Still, it can't be `@staticmethod` due to JIT, right @datumbox? But in @ejguan's example above, I don't see a reason why it needs to be, other than it makes little sense to have this as an instance method from the theoretical perspective.

ejguan: Ohhh you are right. So, those `Feature`s are primitive objects? I thought it's going to be a collection object that needs to handle dict or list specially at the outer level.

> I introduced a `collate` function that only does 1

I agree on adding this `collate` function to handle all collection objects before the primitive objects. But, in terms of how to register collate function for a certain primitive type of objects, I still don't agree on suppressing `TypeError`. IMHO, It's error-prone. Like [here](https://github.com/pytorch/pytorch/blob/f64857189d514a662dac09ca6421e1e95e87e843/torch/utils/data/_utils/collate.py#L148-L149), we do want to raise `TypeError` that will be silently suppressed by the `with contextlib.suppress(TypeError):`.

So, I think prefer the following two options:
1. Add this `collate` function with  an argument of `Dict[Type, collate_fn]` to handle each object based on the type
    - First check if `type(elem)` in the dict
    - Second go through each key in dict and check `isinstance` for derived class
2. Keeping the existing `default_collate`, but before any collate on primitive objects, we try to use `collate_fn` attached to the class.

In terms of JIT, I don't know if any of the above option would work. To make JIT working, a jittable custom collate function might be the best option to you.

pmeier: > * Second go through each key in dict and check `isinstance` for derived class

Shouldn't it be an `OrderedDict` then? Otherwise if an input is a subtype of multiple types in the map, two runs may yield different results making the collation non-deterministic.

--- 

I would prefer 1. since otherwise an object would need to know how it wants to be collated, but doesn't have all the context. For example, for classification and segmentation tasks, we want to `torch.stack` images, but for detection tasks they should be kept in a list. We can figure this out from the full sample passed to the collation function, but he object itself can't.

Plus, we don't have to worry about JIT and that can be quite a pain.

If you send a PR, please ping me there so I can see if it actually is extensible enough for us. @datumbox can you confirm that you would also be ok with handling this in the collation function if we only need to provide custom behavior for our stuff, but otherwise can depend on the core functionality?

ejguan: > Shouldn't it be an `OrderedDict` then?

If the `type` exists in the dict, it should invoke the corresponding collate function. And, `dict` is actually ordered in Python. So, if an input is a subtype of multiple types, and the type of the input doesn't equal to any of key, it will follow the order.

Will open a PR today. It should be a simple change.

pmeier: > And, `dict` is actually ordered in Python.

Well, not really. Since Python 3.6, `dict`'s are insertion ordered. Meaning, if you create a dictionary like

```py
dct = {"a": 1, "b": 2}
```

you can be sure that `"a"` is always the first key. However, `dict`'s are not ordered

```py
assert {"a": 1, "b": 2} == {"b": 2, "a": 1}
```

ejguan: If you try iterate over those dicts, you will find it's ordered.
```py
d1 = {"a": 1, "b": 2}
for k in d1:
    print(k)  # a, b

d2 = {"b": 2, "a": 1}
for k in d2:
    print(k)  # b, a
```

datumbox: >  @datumbox can you confirm that you would also be ok with handling this in the collation function if we only need to provide custom behavior for our stuff, but otherwise can depend on the core functionality?

Just to confirm, is your proposal to handle custom behaviour on our reference scripts? If yes, then that sounds good. I wouldn't want to add this on main TorchVision but rather leave it to the users to handle those special bases based on their use-case/dataset. Does that make sense?

references/video_classification/datasets.py
```diff
@@ -1,14 +1,18 @@
...
-from torch import Tensor
+import torch
+from torchvision import datasets
+from torchvision.prototype import features
 
 
-class KineticsWithVideoId(torchvision.datasets.Kinetics):
-    def __getitem__(self, idx: int) -> Tuple[Tensor, Tensor, int]:
+class KineticsWithVideoId(datasets.Kinetics):
+    def __getitem__(self, idx: int) -> Tuple[torch.Tensor, torch.Tensor, features.Label, int]:
```
pmeier, line 9: Not sure why we have annotations here, but they were wrong before. Note that type of the first returned value is not `features.Video`, since we normalize before.

references/video_classification/presets.py
```diff
@@ -19,7 +19,13 @@ def __init__(
         ]
         if hflip_prob > 0:
             trans.append(transforms.RandomHorizontalFlip(hflip_prob))
-        trans.extend([transforms.Normalize(mean=mean, std=std), transforms.RandomCrop(crop_size), ConvertBCHWtoCBHW()])
+        trans.extend(
```
pmeier, line 22: No change here other than to explode this and make it readable.

torchvision/transforms/functional_tensor.py
pmeier: Is this still relevant after #6726?

references/video_classification/datasets.py
datumbox: We don't need the annotations, you are right.

datumbox (commented):
Shall we add MixUp/CutMix wrappers @pmeier? Or even better, we could proceed with the Type Wrapping transform we discussed last week (the one configured with a map on the constructor) and remove all the custom wrapping classes from this PR.

torchvision/transforms/functional_tensor.py
datumbox: No idea. I'll need to check.

datumbox:
# Video

## Original recipe

Target Acc: 68.368 / 88.050 - time: 2 days, 13:07:05 - jobid: experiments/PR6412/59255

Using githash 669b1ba:
```
PYTHONPATH=$PYTHONPATH:`pwd` python -u run_with_submitit.py --ngpus 8 --nodes 8 --cache-dataset --batch-size=12 --lr 0.2 --clip-len 64 --clips-per-video 5 --sync-bn --model s3d --train-resize-size 256 256 --train-crop-size 224 224 --val-resize-size 256 256 --val-crop-size 224 224 --data-path="/datasets01_ontap_isolated/kinetics/070618/400/"
Submitted job_id: 71369
```

Validated using 128 clip lengths as the original model:
```
torchrun --nproc_per_node=8 train.py --data-path="/datasets01_ontap/kinetics/070618/400/" --batch-size=16 --test-only --cache-dataset --clip-len 128 --clips-per-video 1 --model s3d --resume /PATH/TO/MODEL
```

**Result:** TODO

torchvision/transforms/functional_tensor.py
datumbox: yeap same issue on the new implemantation:
```
  File "./torchvision/prototype/transforms/functional/_color.py", line 237, in equalize_video
    return equalize_image_tensor(video)
  File "./torchvision/prototype/transforms/functional/_color.py", line 230, in equalize_image_tensor
    return _equalize_image_tensor_vec(image.view(-1, height, width)).view(image.shape)
  File "./torchvision/prototype/features/_feature.py", line 94, in __torch_function__
    output = func(*args, **kwargs or dict())
```

datumbox:
# Tensor Backend

Reverifying the new API after the speed optimizations. Reference runs at https://github.com/pytorch/vision/pull/6433#issuecomment-1256741233 and https://github.com/pytorch/vision/pull/6433#issuecomment-1274585805

## Classification

Using githash d8b5202:
```
PYTHONPATH=$PYTHONPATH:`pwd` python -u run_with_submitit.py --ngpus 8 --nodes 1 --model resnet50 --batch-size 128 --lr 0.5 --lr-scheduler cosineannealinglr --lr-warmup-epochs 5 --lr-warmup-method linear --auto-augment ta_wide --epochs 600 --random-erase 0.1 --label-smoothing 0.1 --mixup-alpha 0.2 --cutmix-alpha 1.0 --weight-decay 0.00002 --norm-weight-decay 0.0 --train-crop-size 176 --model-ema --val-resize-size 232 --ra-sampler --ra-reps 4 --data-path /datasets01_ontap/imagenet_full_size/061417/
# Target Acc: 80.626 / 95.310 - time: 2 days, 6:04:18 - jobid: experiments/PR6433/68029
Submitted job_id: 75703
TODO
```

Using githash d8b5202:
```
PYTHONPATH=$PYTHONPATH:`pwd` python -u run_with_submitit.py --ngpus 8 --nodes 1 --model mobilenet_v3_small --epochs 600 --opt rmsprop --batch-size 128 --lr 0.064 --wd 0.00001 --lr-step-size 2 --lr-gamma 0.973 --auto-augment imagenet --random-erase 0.2 --data-path /datasets01_ontap/imagenet_full_size/061417/
# Target Acc: 66.044 / 86.338 - time: 2 days, 4:55:57 - jobid: experiments/PR6433/68030
Submitted job_id: 75704
TODO
```

## Detection

Using githash d8b5202:
```
PYTHONPATH=$PYTHONPATH:`pwd` python -u run_with_submitit.py --ngpus 8 --nodes 1 --weights-backbone ResNet50_Weights.IMAGENET1K_V2 --dataset coco --model retinanet_resnet50_fpn_v2 --opt adamw --lr 0.0001 --epochs 26 --lr-steps 16 22 --weight-decay 0.05 --norm-weight-decay 0.0 --data-augmentation multiscale --sync-bn --data-path /datasets01_ontap/COCO/022719/
# Target Acc: 0.415 - time: 9:49:04 - jobid: experiments/PR6433/67794
Submitted job_id: 75705
TODO
```

Using githash d8b5202:
```
PYTHONPATH=$PYTHONPATH:`pwd` python -u run_with_submitit.py --ngpus 8 --nodes 1 --dataset coco --model ssdlite320_mobilenet_v3_large --aspect-ratio-group-factor 3 --epochs 660 --lr-scheduler cosineannealinglr --lr 0.15 --batch-size 24 --weight-decay 0.00004 --data-augmentation ssdlite --data-path /datasets01_ontap/COCO/022719/
# Target Acc: 0.212 - time: 1 day, 16:06:26 - jobid: experiments/PR6433/67795
Submitted job_id: 75706
TODO
```

Using githash d8b5202:
```
PYTHONPATH=$PYTHONPATH:`pwd` python -u run_with_submitit.py --ngpus 8 --nodes 1 --weights-backbone VGG16_Weights.IMAGENET1K_FEATURES --dataset coco --model ssd300_vgg16 --epochs 120 --lr-steps 80 110 --aspect-ratio-group-factor 3 --lr 0.002 --batch-size 4 --weight-decay 0.0005 --trainable-backbone-layers 5 --data-augmentation ssd --data-path /datasets01_ontap/COCO/022719/
# Target Acc: 0.252 - time: 17:28:42 - jobid: experiments/PR6433/67796
Submitted job_id: 75707
TODO
```

Using githash d8b5202:
```
PYTHONPATH=$PYTHONPATH:`pwd` python -u run_with_submitit.py --ngpus 8 --nodes 4 --dataset coco --model maskrcnn_resnet50_fpn_v2 --epochs 600 --lr-steps 540 570 585 --lr 0.32 --batch-size 8 --weight-decay 0.00004 --sync-bn --data-augmentation lsj --use-copypaste --data-path /datasets01_ontap/COCO/022719/
# Target Acc: 0.474 / 0.416 - time: 3 days, 15:09:55 - jobid: experiments/PR6433/67791
Submitted job_id: 75708
TODO
```

## Segmentation

Using githash d8b5202:
```
PYTHONPATH=$PYTHONPATH:`pwd` python -u run_with_submitit.py --ngpus 8 --nodes 1 --dataset coco -b 4 --model lraspp_mobilenet_v3_large --lr 0.01 --wd 0.000001 --weights-backbone MobileNet_V3_Large_Weights.IMAGENET1K_V1
# Target Acc: 90.5 / 54.7 - time: 2:20:45 - jobid: experiments/PR6433/67797
Submitted job_id: 75709
TODO 
```

## Video

### New Recipe

Using githash d8b5202:
```
PYTHONPATH=$PYTHONPATH:`pwd` python -u run_with_submitit.py --ngpus 8 --nodes 8 --cache-dataset --batch-size=12 --lr 0.2 --clip-len 64 --clips-per-video 5 --sync-bn --model s3d --auto-augment ta_wide --mixup-alpha 0.8 --cutmix-alpha 1.0 --random-erase 0.25 --train-resize-size 256 320 --train-crop-size 224 224 --val-resize-size 256 256 --val-crop-size 224 224 --data-path="/datasets01_ontap_isolated/kinetics/070618/400/"
# Target Acc: 70.903 / 90.434 - time: 3 days, 6:00:05 - jobid: experiments/PR6433/72508
Submitted job_id: 75701
TODO
```
```
trainrun torchrun --nproc_per_node=8 train.py --data-path="/datasets01_ontap/kinetics/070618/400/" --batch-size=16 --test-only --cache-dataset --clip-len 128 --clips-per-video 1 --model s3d --val-resize-size 256 256 --val-crop-size 224 224 --resume /path/to/model.pth
TODO
```

PR #6434, opened 2022-08-17
replace query_recursively with pytree implementation

pmeier:
Addresses https://github.com/pytorch/vision/pull/6408#issuecomment-1216308721.


torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -43,29 +32,31 @@ def _get_random_item(self, dct: Dict[K, V]) -> Tuple[K, V]:
...
-        images = list(query_recursively(fn, sample))
         if not images:
             raise TypeError("Found no image in the sample.")
         if len(images) > 1:
             raise TypeError(
                 f"Auto augment transformations are only properly defined for a single image, but found {len(images)}."
             )
-        return images[0]
+        idx_flat, image = images[0]
+        return sample_flat, sample_spec, idx_flat, image
```
pmeier, line 55: The first three parameters are never touched by the transforms, but are needed to put the new image in the right place before we unflatten later. Should we wrap them in a `dataclass` or something? Or maybe just in a tuple like

```suggestion
        return (sample_flat, sample_spec, idx_flat), image
```

to indicate their cohesion?

vfdev-5 (approved):
Thanks @pmeier

torchvision/prototype/transforms/_auto_augment.py
vfdev-5: Its an internal method, so any convention is ok for me

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -361,7 +352,7 @@ def __init__(
     def forward(self, *inputs: Any) -> Any:
         sample = inputs if len(inputs) > 1 else inputs[0]
 
-        id, image = self._extract_image(sample)
+        sample_flat, sample_spec, idx_flat, image = self._flatten(sample)
```
vfdev-5, line 355: I appologize if I'm missing the context, If I understand correctly the code, you fetch the first image from sample with `_flatten` and below in the code the image is transformed and written back to sample.
Why do we transform only the first image ?

vfdev-5: More generally question, why can't we restructure the implementation such that we use `forward` to 
inspect sample and ensure that only images (and/or only one image in sample) go further into `_transform`. In `_transform` we do the necessary AA ops. What do you think ?

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -43,29 +32,31 @@ def _get_random_item(self, dct: Dict[K, V]) -> Tuple[K, V]:
         key = keys[int(torch.randint(len(keys), ()))]
         return key, dct[key]
 
-    def _extract_image(
+    def _flatten(
```
datumbox, line 35: I understand that this is an internal method but I feel its name and return types can lead to confusion to someone reading the code. This method now ends up doing too many things (flattening and extract image); similar for unflatten. What is the reason for this?

torchvision/prototype/transforms/_auto_augment.py
datumbox: I agree with Victor. It's best if we keep things as simple as possible. For me it's still unclear why we need to keep the 3 items and not just return as previously the image and its location and append it at the end. I'm always supportive when we reuse methods from core and reduce the utility methods on our side but in this case it seems that we are complicating things. I wonder if it's possible to still reuse parts of pytree but keep our implementation closer to the "extract image" logic.

pmeier: > For me it's still unclear why we need to keep the 3 items and not just return as previously the image and its location and append it at the end. 

We can't append it, but rather put it in the right place. If the input is not a flat dict or list, we need more than one index / key. The `query_recursively` gave us and `id` which is this collection of indices and keys. In the new implementation the equivalent is `idx_flat`. 

The difference is that the `id` of the previous implementation corresponds to the original sample, whereas `idx_flat` corresponds to the flattened sample.

torchvision/prototype/transforms/_auto_augment.py
pmeier: > This method now ends up doing too many things (flattening and extract image)

Basically it does the same thing as before. As explained in https://github.com/pytorch/vision/pull/6434#discussion_r948738057 `idx_flat` corresponds to the flattened sample. Thus, if we don't returned the flattened sample, we need to flatten it again later when we put the augmented image in. 

> I understand that this is an internal method but I feel its name and return types can lead to confusion to someone reading the code.

Agreed. I'll come up with better names.

datumbox: So this is more of a speed optimization? I don't know if flattening and unflattening can be a bottleneck. If we believe it is, we should measure it.

My concern here is that the previous code, though it was performing extra ops, it was much clearer. The new one, due to its inconsistent naming and complexity, is much harder to read.

torchvision/prototype/transforms/_auto_augment.py
datumbox: Thanks for the explanation. That makes sense.

Does it make sense not to replace our previous implementation or to replace smaller components of it? It seems what we did previously was both efficient and easier to read. Here we try to reduce the LOC (which I always like) but seems that this isn't for free both in terms of speed and readability.

PR #6436, opened 2022-08-17
Fix minor bug on `PolynomialLR` invocation

datumbox:
Resolves issue reported at https://github.com/pytorch/vision/commit/6e535db255cee3ce878dd7a54dda01d4ec8932c1#commitcomment-81409388

There seems to be a misspelling on the name of the parameter. This PR updates `total_steps` to `total_iters` which is the [correct argument](https://github.com/pytorch/pytorch/blob/4618371da56c887195e2e1d16dad2b9686302800/torch/optim/lr_scheduler.py#L734).

### DIFF
--- a/references/segmentation/train.py
+++ b/references/segmentation/train.py
@@ -186,7 +186,7 @@ def main(args):
 
     iters_per_epoch = len(data_loader)
     main_lr_scheduler = PolynomialLR(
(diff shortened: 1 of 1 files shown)


jdsgomes (approved):
LGTM, thanks for the fix

PR #6442, opened 2022-08-18
[FBcode->GH] Add an option to skip packets with empty data

NicolasHug:
Cherypick of a671bfff97770e15a98a16fdd194d906c9531a0a

Summary:
This diff adds **`skipOperationNotPermittedPackets`** to `DecoderParameters`.

Once that is set to `True`, it allows decoder to skip through those packets, that are causing `EPERM` errors and result into stoppage of consuming a stream.

Reviewed By: jdsgomes

Differential Revision: D38732706

fbshipit-source-id: a5cb64e14edda376e6ba240089ebee2e3a9865d0


datumbox (approved):
Stamping

PR #6443, opened 2022-08-18
[FBcode->GH] Free avPacket on EAGAIN decoder error (#6432)

NicolasHug:
Cherry-pick of 6bfca028d35d784fb537b06e103e8e10347614fe

Summary:
Pull Request resolved: https://github.com/pytorch/vision/pull/6432

According to the documentation the packet has to be freed after `av_read_frame()` call.
```
 If pkt->buf is NULL, then the packet is valid until the next
 av_read_frame() or until avformat_close_input(). Otherwise the packet
 is valid indefinitely. In both cases the packet must be freed with
 av_packet_unref when it is no longer needed.
```

Differential Revision: D38747612

fbshipit-source-id: 2e4ccc8365d0d97e5da756ff9c1dcdf27ed323f0


datumbox (approved):
stamping

PR #6444, opened 2022-08-18
Improve the Video Reference scripts

datumbox:
Supports the work on #6412

- Remove hardcoded resize/crop values and add extra arguments like we do for the classification reference
- Make `--lr` estimation similar to every other reference script (avoid multiplying by world-size)
- Update and centralize the docs. Make sure they are up-to-date.


jdsgomes (approved):
LGTM!

PR #6445, opened 2022-08-18
[EXPERIMENTAL] Add support for dynamic frames per clip

datumbox:
The [S3D](https://arxiv.org/pdf/1712.04851.pdf) model (see #6412) uses a less common validation protocol which requires passing all the available frames during inference. Currently this is not possible.

This PR attempts to change this (untested) by making the specific option optional. The "solution" is quite messy, not sure we should merge or if we should leave as a reference patch for people who want to reproduce the accuracy of the model.

### DIFF
--- a/references/video_classification/train.py
+++ b/references/video_classification/train.py
@@ -155,6 +155,8 @@ def main(args):
     train_crop_size = tuple(args.train_crop_size)
 
     if args.clip_len == 0:
+        if args.batch_size != 1:
+            raise ValueError("Dynamic clip length is supported only for batch-size 1.")
         args.clip_len = None
 
     traindir = os.path.join(args.data_path, "train")
--- a/references/video_classification/train.py
+++ b/references/video_classification/train.py
@@ -154,6 +154,9 @@ def main(args):
     train_resize_size = tuple(args.train_resize_size)
     train_crop_size = tuple(args.train_crop_size)
 
+    if args.clip_len == 0:
+        args.clip_len = None
+
     traindir = os.path.join(args.data_path, "train")
     valdir = os.path.join(args.data_path, "val")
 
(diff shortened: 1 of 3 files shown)


torchvision/datasets/video_utils.py
```diff
@@ -100,7 +101,7 @@ class VideoClips:
     def __init__(
         self,
         video_paths: List[str],
-        clip_length_in_frames: int = 16,
+        clip_length_in_frames: Optional[int] = 16,
```
datumbox, line 104: Perhaps I should keep it `int` with `0` encoding the dynamic mode?

references/video_classification/train.py
```diff
@@ -154,6 +154,11 @@ def main(args):
     train_resize_size = tuple(args.train_resize_size)
     train_crop_size = tuple(args.train_crop_size)
 
+    if args.clip_len == 0:
+        if args.batch_size != 1:
+            raise ValueError("Dynamic clip length is supported only for batch-size 1.")
```
datumbox, line 159: Another gotcha. Batching won't work for variable temporal dimension. So this can only used for batch-size = 1

torchvision/datasets/video_utils.py
```diff
@@ -213,6 +214,8 @@ def compute_clips_for_video(
         total_frames = len(video_pts) * (float(frame_rate) / fps)
         _idxs = VideoClips._resample_video_idx(int(math.floor(total_frames)), fps, frame_rate)
         video_pts = video_pts[_idxs]
+        if num_frames is None:
+            num_frames = int(total_frames)
```
datumbox, line 218: This can lead to videos so short that don't make the minimum temporal dimension of the model, leading to errors such as:
```
RuntimeError: input image (T: 1 H: 7 W: 7) smaller than kernel size (kT: 2 kH: 7 kW: 7)
```

references/video_classification/train.py
```diff
@@ -346,7 +351,13 @@ def get_args_parser(add_help=True):
     )
     parser.add_argument("--model", default="r2plus1d_18", type=str, help="model name")
     parser.add_argument("--device", default="cuda", type=str, help="device (Use cuda or cpu Default: cuda)")
-    parser.add_argument("--clip-len", default=16, type=int, metavar="N", help="number of frames per clip")
```
jdsgomes, line 349: One option could be to have `None` as default value and avoid passing 0s around. This will be nicer in terms of code IMO but as @datumbox pointed out in a conversation offline it has several downsides:
it doesn't play along with Batching, it causes issues with very short videos, etc

datumbox:
Using this PR, S3D achieves the following accuracy:
```
torchrun --nproc_per_node=8 train.py --data-path="/datasets01/kinetics/070618/400/" --batch-size=1 --test-only --cache-dataset --clip-len 0 --clips-per-video 1 --model s3d --weights="S3D_Weights.DEFAULT"
 * Clip Acc@1 67.112 Clip Acc@5 87.588
```

cc @sophiazhi

datumbox:
Closing as this doesn't improve a significant improvement from our previous result and it's quite messy and non-standard to support.

PR #6447, opened 2022-08-18
expand has_any and has_all to also accept check callables

pmeier:
This is useful if the type checking involves more than plain types. A common example is to check for an image in the sample which is done with the types `features.Image` and `PIL.Image.Image` as well as the check function `is_simple_tensor`. With this PR, we can pass all three to `has_any` and `has_all`:

```py
import PIL.Image
import torch
from torchvision.prototype.transforms._utils import has_any, is_simple_tensor
from torchvision.prototype.transforms.functional import to_image_pil

image_tensor = torch.rand(3, 16, 16)
image_pil = to_image_pil(image_tensor)
image = features.Image(image_tensor)

for input in [image_tensor, image_pil, image]:
    assert has_any(input, PIL.Image.Image, features.Image, is_simple_tensor)
```

This PR also uses this new idiom to add support for simple tensors to `CutMix` and `MixUp` as well as `RandomIoUCrop` as discussed in https://github.com/pytorch/vision/pull/6401#discussion_r948749012.


test/test_prototype_transforms_utils.py
```diff
@@ -0,0 +1,51 @@
...
+        ((0.0, ""), (int, float), False),
+        ((0.0, ""), (int, str), False),
+        ((0, ""), (float, str), False),
+        ((0, 0.0, ""), (int, float, str), True),
+        ((0.0, ""), (int, float, str), False),
+        ((0, ""), (int, float, str), False),
+        ((0, 0.0), (int, float, str), False),
+        ((0, 0.0, ""), (lambda obj: isinstance(obj, (int, float, str)),), True),
+        ((0, 0.0, ""), (lambda _: False,), False),
+        ((0, 0.0, ""), (lambda _: True,), True),
```
datumbox, line 47: Can we add also a few test cases for common cases? BBoxes, Masks, Images, Tensors etc.

torchvision/prototype/transforms/_utils.py
```diff
@@ -39,14 +39,27 @@ def get_image_dimensions(image: Union[PIL.Image.Image, torch.Tensor, features.Im
...
-def has_any(sample: Any, *types: Type) -> bool:
+def _parse_types_or_checks(
+    types_or_checks: Tuple[Union[Type, Callable[[Any], bool]], ...]
+) -> List[Callable[[Any], bool]]:
+    return [
+        cast(Callable[[Any], bool], lambda obj, typ=type_or_check: isinstance(obj, typ))
+        if isinstance(type_or_check, type)
+        else type_or_check
+        for type_or_check in types_or_checks
+    ]
```
datumbox, line 50: Nit: This is correct but really densely written. I had to eyeball it for 2 minutes to understand that the if/else is after the loop. Might be worth refactoring with loops to make it clearer, even if that's less pythonic.

torchvision/prototype/transforms/_utils.py
```diff
@@ -39,14 +39,27 @@ def get_image_dimensions(image: Union[PIL.Image.Image, torch.Tensor, features.Im
...
+
+
+def has_any(sample: Any, *types_or_checks: Union[Type, Callable[[Any], bool]]) -> bool:
     flat_sample, _ = tree_flatten(sample)
-    return any(issubclass(type(obj), types) for obj in flat_sample)
+    checks = _parse_types_or_checks(types_or_checks)
+    return any(any([check(obj) for obj in flat_sample]) for check in checks)
```
datumbox, line 56: Nit: Similar to above, I think you can make this easier to read and potentially faster with loops (because you can `break` early.

datumbox (commented):
@pmeier This might be a matter of styles and preferences, but I have hard time reading code that is so densely written as this. I understand that some people like it as it's more pythonic.


@vfdev-5 thoughts?

torchvision/prototype/transforms/_utils.py
NicolasHug: Unsolicited opinion: I find it hard to read too. The heavy type annotations, the call to `cast()` and the use of `lambda` with a default value also impair readability IMHO.

pmeier: I agree, the lambda with annotations is hard to read. I think we can safely replace the `cast` with an ignore comment. Plus, we can simplify the function annotations a little if we accept `*types_or_checks`.

However, I don't have a more readable way to create callables in a loop other than the default value. Happy to take suggestions here.

torchvision/prototype/transforms/_utils.py
```diff
@@ -39,14 +39,34 @@ def get_image_dimensions(image: Union[PIL.Image.Image, torch.Tensor, features.Im
...
-    return any(issubclass(type(obj), types) for obj in flat_sample)
+    for type_or_check in types_or_checks:
+        passed_check = False
+        for obj in flat_sample:
+            if isinstance(obj, type_or_check) if isinstance(type_or_check, type) else type_or_check(obj):
+                passed_check = True
+                break
 
+        if passed_check:
+            return True
```
datumbox, line 52: Nit: Isn't this the same as:
```suggestion
        for obj in flat_sample:
            if type_or_check(obj) if callable(type_or_check) else isinstance(obj, type_or_check):
                return True
```

torchvision/prototype/transforms/_utils.py
```diff
@@ -39,14 +39,34 @@ def get_image_dimensions(image: Union[PIL.Image.Image, torch.Tensor, features.Im
...
-    return not bool(set(types) - set([type(obj) for obj in flat_sample]))
+    for type_or_check in types_or_checks:
+        passed_check = False
+        for obj in flat_sample:
+            if isinstance(obj, type_or_check) if isinstance(type_or_check, type) else type_or_check(obj):
+                passed_check = True
+                break
+
+        if not passed_check:
+            return False
```
datumbox, line 67: If my above comment is correct, this might be possible to be simplified as well.

datumbox (approved):
LGTM. I've added a few optional comments above, I haven't thoroughly checked it as I am in between meetings. Feel free to ignore.

torchvision/prototype/transforms/_utils.py
```diff
@@ -39,14 +39,24 @@ def get_image_dimensions(image: Union[PIL.Image.Image, torch.Tensor, features.Im
...
+            if isinstance(obj, type_or_check) if isinstance(type_or_check, type) else type_or_check(obj):
+                return True
+    return False
 
 
-def has_all(sample: Any, *types: Type) -> bool:
+def has_all(sample: Any, *types_or_checks: Union[Type, Callable[[Any], bool]]) -> bool:
     flat_sample, _ = tree_flatten(sample)
-    return not bool(set(types) - set([type(obj) for obj in flat_sample]))
+    for type_or_check in types_or_checks:
```
pmeier, line 53: They are asymmetric because the behavior we want for `has_any` is `any(any(...))` and for `has_all` is `any(all(...))`. If `has_all` would be `all(all(...))` they would be symmetric.

PR #6448, opened 2022-08-18
add support for apply probability to CutMix and MixUp

pmeier:


torchvision/prototype/transforms/_augment.py
```diff
@@ -97,9 +97,9 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
             return inpt
 
 
-class _BaseMixupCutmix(Transform):
-    def __init__(self, *, alpha: float) -> None:
-        super().__init__()
+class _BaseMixupCutmix(_RandomApplyTransform):
+    def __init__(self, *, alpha: float, p: float = 0.5) -> None:
```
pmeier, line 101: Default for other transforms is `p=0.5`:

https://github.com/pytorch/vision/blob/2a0eea8222687bb93c11a23c95c19c69ebf96587/torchvision/prototype/transforms/_augment.py#L15-L18

https://github.com/pytorch/vision/blob/2a0eea8222687bb93c11a23c95c19c69ebf96587/references/classification/transforms.py#L22

In the classification reference we hardcode `p=1.0` for `CutMix` and `MixUp`:

https://github.com/pytorch/vision/blob/2a0eea8222687bb93c11a23c95c19c69ebf96587/references/classification/train.py#L204-L207

I would keep `p=0.5` here and set `p=1.0` in the references as before. No strong opinion though.

PR #6449, opened 2022-08-18
add cleanup transformation for bounding boxes

pmeier:
```py
import torch


image = features.EncodedImage.from_path("test/assets/fakedata/logos/rgb_pytorch.png").decode()

bounding_boxes = features.BoundingBox(
    [[60, 30, 15, 15], [30, 60, 15, 15]],
    format=features.BoundingBoxFormat.CXCYWH,
    image_size=(100, 100),
    dtype=torch.float,
)

segmentation_masks = torch.zeros((2, 100, 100), dtype=torch.bool)
segmentation_masks[0, 24:36, 55:66] = True
segmentation_masks[0, 55:66, 24:36] = True
segmentation_masks = features.SegmentationMask(segmentation_masks)

labels = features.Label([0, 1], categories=["foo", "bar"])

# Increase the size for easier visual comparison
image, bounding_boxes, segmentation_masks = transforms.Resize(400)(image, bounding_boxes, segmentation_masks)
```

before

```py
torch.manual_seed(123)
    [
        transforms.RandomCrop(size=(200, 200)),
        transforms.CleanupBoxes(),
    ]
)

transformed_image, transformed_bounding_boxes, transformed_segmentation_masks, transformed_labels = transform(
    image,
    bounding_boxes,
    segmentation_masks,
    labels,
)

assert transformed_bounding_boxes.shape[0] == 1
assert transformed_segmentation_masks.shape[0] == 1
assert transformed_labels.shape[0] == 1
```

after

### Todo

- [ ] add test
- [ ] find a good name for this operation. Help wanted!


torchvision/prototype/transforms/_misc.py
```diff
@@ -79,3 +81,28 @@ def __init__(self, dtype: torch.dtype, *types: Type) -> None:
 
     def extra_repr(self) -> str:
         return ", ".join([f"dtype={self.dtype}", f"types={[type.__name__ for type in self.types]}"])
+
+
+class CleanupBoxes(Transform):
+    def _get_params(self, sample: Any) -> Dict[str, Any]:
+        bounding_boxes = query_bounding_box(sample)
+        bounding_boxes_clamped = F.clamp_bounding_box(
```
datumbox, line 89: Shouldn't we be doing this clamp in the transform to ensure all values are valid? Without this clamp, we might have x1,y1 values that are negative on the final boxes. Or we might end up with x2,y2 that are beyond the size of the image.

torchvision/prototype/transforms/_misc.py
```diff
@@ -79,3 +81,28 @@ def __init__(self, dtype: torch.dtype, *types: Type) -> None:
...
+            bounding_boxes, format=bounding_boxes.format, image_size=bounding_boxes.image_size
+        )
+        bounding_boxes_xywh = F.convert_bounding_box_format(
+            bounding_boxes_clamped, old_format=bounding_boxes.format, new_format=features.BoundingBoxFormat.XYWH
+        )
```
datumbox, line 94: This might be slow. The clamp already changes format twice internally, so that's the 3rd format change in a row.

torchvision/prototype/transforms/_misc.py
```diff
@@ -79,3 +81,28 @@ def __init__(self, dtype: torch.dtype, *types: Type) -> None:
...
+        is_valid = torch.all(bounding_boxes_xywh[..., 2:] > 0, dim=-1)
+
+        return dict(is_valid=is_valid)
```
datumbox, line 97: I have the feeling that this transform might not fit nicely in the API. You effectively want to clamp the bboxes first and modify their values and then based on this outcome decide that is invalid sync the rest of the items in the output.

vfdev-5: Yes, I would also implement only forward for this transform

pmeier:
To avoid duplication of computation, I see two ways:

1. We perform the clamping in the `forward`, i.e. before the actual transform happens.
2. We make the cleanup transform an actual `Compose` and put the clamping transform in front of the invalid box removal.

Both are new idioms of which I like 2. a lot better. I've implemented it in 108db6205352b451b8854191f280adf711d49797 to have a better basis for discussion. LMK if you agree or if you prefer 1. or are ok with duplicate computation.

torchvision/prototype/features/_bounding_box.py
```diff
@@ -69,6 +69,16 @@ def to_format(self, format: Union[str, BoundingBoxFormat]) -> BoundingBox:
...
 
+    @property
+    def area(self) -> torch.Tensor:
+        if self.format in {BoundingBoxFormat.XYWH, BoundingBoxFormat.CXCYWH}:
+            _, _, width, height = torch.unbind(self, dim=-1)
+        else:  # self.format == BoundingBoxFormat.XYXY
+            x1, y1, x2, y2 = torch.unbind(self, dim=-1)
+            width = x2 - x1
+            height = y2 - y1
+        return width * height
```
datumbox, line 80: I don't think we should be implementing an op here. Those `_Feature` objects should be extremely thin wrappers of existing ops. Note there is already an operator for box area estimation called `box_area()`.

BTW what you do here is introduce the idiom of attaching custom methods and properties to specific `_Feature` subclasses but not to all of them. We got a couple of utility methods that do this already (like one for formatting, plotting etc) but it's not clear if we wanna add those methods to the objects. I don't object to that, just flagging it as something "new".

torchvision/prototype/transforms/_misc.py
```diff
@@ -79,3 +81,27 @@ def __init__(self, dtype: torch.dtype, *types: Type) -> None:
 
     def extra_repr(self) -> str:
         return ", ".join([f"dtype={self.dtype}", f"types={[type.__name__ for type in self.types]}"])
+
+
+# This transform is not exposed under `torchvision.prototype.transforms` since it should never be used without
+# `ClampBoundingBox`
+class RemoveInvalid(Transform):
```
datumbox, line 88: If we consider them internal, let's prefix both transforms with `_`.

datumbox (approved):
I like the approach, only minor nits from my side.

torchvision/prototype/features/_bounding_box.py
pmeier: > I don't think we should be implementing an op here. Those `_Feature` objects should be extremely thin wrappers of existing ops.

You are right. Let's not introduce new idioms for now.

> Note there is already an operator for box area estimation called `box_area()`.

For reference: 

https://github.com/pytorch/vision/blob/9c3e2bf46bc49997679785d76b7d0a9fea0223c7/torchvision/ops/boxes.py#L219

I'm not going to use that since it would require to always convert to `XYXY` although the other formats are even easier for this computation. 

> BTW what you do here is introduce the idiom of attaching custom methods and properties to specific _Feature subclasses but not to all of them. We got a couple of utility methods that do this already (like one for formatting, plotting etc) but it's not clear if we wanna add those methods to the objects.

I think this is a case that makes sense to have it on the `features.BoundingBox`. Compared to the other kernels this is does not return a somehow changed bounding box, but rather a new quantity. Thus, this should not be a transform.

torchvision/prototype/transforms/_misc.py
```diff
@@ -79,3 +81,33 @@ def __init__(self, dtype: torch.dtype, *types: Type) -> None:
 
     def extra_repr(self) -> str:
         return ", ".join([f"dtype={self.dtype}", f"types={[type.__name__ for type in self.types]}"])
+
+
+class _RemoveInvalid(Transform):
```
pmeier, line 86: We need a better name for this. Help wanted!

torchvision/prototype/transforms/_misc.py
```diff
@@ -79,3 +81,33 @@ def __init__(self, dtype: torch.dtype, *types: Type) -> None:
...
+            height = y2 - y1
+        area = width * height
+
+        return dict(is_valid=area > 0)
+
+    def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
+        return inpt.new_like(inpt, inpt[params["is_valid"]])
+
+
+class CleanupBoxes(Compose):
```
pmeier, line 106: Same as above.

torchvision/prototype/transforms/_meta.py
```diff
@@ -66,3 +66,11 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
         return F.convert_color_space(
             inpt, color_space=self.color_space, old_color_space=self.old_color_space, copy=self.copy
         )
+
+
+class ClampBoundingBox(Transform):
```
datumbox, line 71: Why isn't this private as well?

pmeier: Clamping boxes is useful in a general context, is it not? `torchvision.prototype.transforms.functional.clamp_bounding_box` is public, so shouldn't we also have a public transformation for it?

datumbox: Clamping is certainly useful but I would expect that one would do it by using directly the proposed clean up transform. It really depends on what idiom you would like to push for. I was imagining that we would be promoting the use of `CleanupBoxes` instead of individual components. WDYT?

torchvision/prototype/transforms/_misc.py
datumbox: `_RemoveEmptyBoundingBoxes`

torchvision/prototype/transforms/_meta.py
datumbox: `_ClampBoundingBoxes`? Plural?

pmeier:
As discussed offline, we opted to split the `Compose` into its parts and expose them since we didn't find a descriptive name for the composed transform. In addition, we changed `RemoveEmptyBoundingBoxes` to `RemoveSmallBoundingBoxes` and use the `remove_small_boxes` op `torchvision.ops` internally. For the default value of `min_size=1`, the behavior is the same, i.e. only empty boxes will be removed.

torchvision/prototype/transforms/_misc.py
```diff
@@ -141,40 +141,26 @@ def extra_repr(self) -> str:
...
-    def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
-        output = F.clamp_bounding_box(inpt, format=inpt.format, image_size=inpt.image_size)
-        return features.BoundingBox.new_like(inpt, output)
-
-
-class _RemoveEmptyBoundingBoxes(Transform):
+class RemoveSmallBoundingBoxes(Transform):
     _transformed_types = (features.BoundingBox, features.SegmentationMask, features.Label, features.OneHotLabel)
 
+    def __init__(self, min_size: int = 1) -> None:
```
datumbox, line 145: This has to be a float because the underlying op uses floats:
```suggestion
    def __init__(self, min_size: float = 1.0) -> None:
```

datumbox (approved):
LGTM, with one minor change. Feel free to merge after correcting it and upon a green-ish CI.

PR #6451, opened 2022-08-18
[proto] Ported SimpleCopyPaste transform

vfdev-5:
- Ported SimpleCopyPaste
- Add tests

Example with MSCoco (batchsize = 2):

image

image


torchvision/prototype/transforms/_augment.py
```diff
@@ -180,3 +184,190 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
...
+        self,
+        p: float = 0.5,
+        blending: bool = True,
+        resize_interpolation: InterpolationMode = F.InterpolationMode.BILINEAR,
+    ) -> None:
+        super().__init__(p=p)
+        self.resize_interpolation = resize_interpolation
+        self.blending = blending
+
+    def _copy_paste(
```
vfdev-5, line 200: This method is just copied from ref implementation

torchvision/prototype/transforms/_augment.py
```diff
@@ -180,3 +184,190 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
...
+        images, bboxes, masks, labels = [], [], [], []
+        for obj in flat_sample:
+            if isinstance(obj, PIL.Image.Image, features.Image) or is_simple_tensor(obj):
+                images.append(F.to_image_tensor(obj))
+            elif isinstance(obj, features.BoundingBox):
+                bboxes.append(obj)
+            elif isinstance(obj, features.SegmentationMask):
+                masks.append(obj)
+            elif isinstance(obj, (features.Label, features.OneHotLabel)):
+                labels.append(obj)
```
vfdev-5, line 318: As our input is a non-collated batch of datapoint, we need to fetch images, bboxes, masks, labels.
We then transform them and finally put into their places.
I'm not a fan of this code but this helps to do what we want. Maybe we could specify a bit more the input structure and thus code `forward` in a more elegant way.

torchvision/prototype/transforms/_augment.py
```diff
@@ -180,3 +184,190 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
...
+                paste_image,
+                paste_target,
+                blending=self.blending,
+                resize_interpolation=self.resize_interpolation,
+            )
+            output_images.append(output_image)
+            output_targets.append(output_data)
+
+        # Insert updated images and targets into input flat_sample
+        c0, c1, c2, c3 = 0, 0, 0, 0
```
pmeier, line 352: Can't we save these indices the first time we iterate over the flat sample? If yes, maybe we can get away with only doing `flat_sample[i] = output[i]` and only looping over the output.

torchvision/prototype/transforms/_augment.py
```diff
@@ -180,3 +184,190 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
...
+            paste_boxes = paste_boxes.view(-1, 2, 2).mul(ratios).view(paste_boxes.shape)
+
+        paste_alpha_mask = paste_masks.sum(dim=0) > 0
+
+        if blending:
+            paste_alpha_mask = F.gaussian_blur(
+                paste_alpha_mask.unsqueeze(0),
+                kernel_size=(5, 5),
+                sigma=[
+                    2.0,
```
pmeier, line 248:
```suggestion
                    2.0
```

torchvision/prototype/transforms/_augment.py
```diff
@@ -180,3 +184,190 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
...
+        out_target["masks"] = torch.cat([masks, paste_masks])
+
+        # Copy-paste boxes and labels
+        boxes = masks_to_boxes(masks)
+        out_target["boxes"] = torch.cat([boxes, paste_boxes])
+
+        labels = target["labels"][non_all_zero_masks]
+        out_target["labels"] = torch.cat([labels, paste_labels])
+
+        # Update additional optional keys: area and iscrowd if exist
```
pmeier, line 272: Is this an actual `TODO`? I don't think we can unless we make them attributes or the like on `features.BoundingBox`. We can easily compute `area` at runtime, i.e. we could have a `.area()` method. 

`is_crowd` will be a lot harder since it is metadata that is associated with every box individually. Meaning , a `BoundingBox.shape == (10, 4)` with a boolean `.is_crowd is True` makes little sense. A naive way would be to make the `.is_crowd` attribute a list or tuple, but than we get into hot waters whenever an op changes the shape of the tensor. For example, indexing would effect the tensor data, but would leave the `.is_crowd` attribute as is.

My suggestion is to remove all these occurrences until we have figured out if and how we want to handle them.

pmeier (commented):
Not a fan of `forward` behavior here, but I don't think we can really improve it unless we pose some constraints on the input.

torchvision/prototype/transforms/_augment.py
vfdev-5: yes, we can do that. I just found that passing indices everywhere would be a bit bulky...

torchvision/prototype/transforms/_augment.py
```diff
@@ -180,3 +184,187 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
...
+            # Random paste targets selection:
+            num_masks = len(paste_target["masks"])
+
+            if num_masks < 1:
+                # Such degerante case with num_masks=0 can happen with LSJ
+                # Let's just return (image, target)
+                output_image, output_target = image, target
+            else:
+                random_selection = torch.randint(0, num_masks, (num_masks,), device=paste_image.device)
+                random_selection = torch.unique(random_selection).to(torch.long)
```
datumbox, line 353: Reading the previous code, I see that the casting to `long` was done for JIT. Does it make sense to maintain here given we won't be JIT-scriptable? Up to you.

datumbox (dismissed):
LGTM, thanks! I left 2 comments below nothing blocking.

torchvision/prototype/transforms/_augment.py
```diff
@@ -180,3 +184,187 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
...
+        out_target = {k: v for k, v in target.items()}
+
+        out_target["masks"] = torch.cat([masks, paste_masks])
+
+        # Copy-paste boxes and labels
+        bbox_format = target["boxes"].format
+        xyxy_boxes = masks_to_boxes(masks)
+        # masks_to_boxes produces bboxes with x2y2 inclusive but x2y2 should be exclusive
+        # we need to add +1 to x2y2
+        xyxy_boxes[:, 2:] += 1
```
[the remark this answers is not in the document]
vfdev-5: Yes, it is exactly related to that. OK, I'll put a TODO to mark this place.

torchvision/prototype/transforms/_augment.py
vfdev-5: Yes, I can remove that.

vfdev-5:
Seems like git merge has failed as on main we removed PIL and here I'm still using it.

datumbox (approved):
LGTM again. Github was showing me random stuff on the views BTW. I had to check the whole file to ensure I get the right version. Feel free to merge on green CI.

Issue #5297, opened 2022-01-27
pil_to_tensor / to_tensor fails when converting zero-sized PIL images


vadimkantorov:
### 🐛 Describe the bug

```python
import torchvision
import numpy as np
from PIL import Image

img = Image.fromarray(np.empty((496, 0), dtype=np.uint8))

torchvision.transforms.functional.pil_to_tensor(img) # *** SystemError: tile cannot extend outside image
torchvision.transforms.functional.to_tensor(img) # *** SystemError: tile cannot extend outside image

np.asarray(img) # same error, because this upstream bug https://github.com/python-pillow/Pillow/issues/5992
```

torchvision could special-case conversion of zero-sized PIL images while the upstream bug is getting fixed. zero-sized images can sometimes occur as result of not super-well guarded cropping (that's how I hit this bug, the above is just a minimally reproducing example). IMO it's much better to propagate zero-sized images at least through these basic conversion functions. At the very least it should provide a good error message and links to the github issues about the bug

### Versions

(Pdb) torchvision.__version__
'0.10.1+cu111'
(Pdb) PIL.__version__
'8.4.0'
(Pdb) np.__version__
'1.21.2'

PR #6454, opened 2022-08-19
[proto] Fixed to_image_tensor and to_image_pil

vfdev-5:
- Fixed to_image_tensor and to_image_pil
- Replaced copy arg by mode in to_image_pil


torchvision/prototype/transforms/functional/_type_conversion.py
```diff
@@ -27,20 +27,25 @@ def label_to_one_hot(label: torch.Tensor, *, num_categories: int) -> torch.Tenso
 
 
 def to_image_tensor(image: Union[torch.Tensor, PIL.Image.Image, np.ndarray], copy: bool = False) -> torch.Tensor:
+    if isinstance(image, np.ndarray):
+        image = torch.from_numpy(image)
```
datumbox, line 31: This makes the kernel non-jit scriptable.

vfdev-5: was this kernel scriptable before ?

datumbox: Hmm I guess not because of the invocation of `pil_to_tensor`.

datumbox (approved):
LGTM!

@pmeier thoughts?

vadimkantorov:
Related: https://github.com/pytorch/vision/issues/5297

Maybe at least add a test on empty PIL images? (if it doesn't exist)

PR #6457, opened 2022-08-20
feat: Added support of Poly Loss

frgfm:
As discussed in #6439, this PR adds support for the poly loss. A few things are left to decide / do:
- [ ] decide whether a module API should also be added (easy to add in this PR)
- [ ] specify the depths of the unittests (I've checked the one of the focal loss, it's quite extensive. how much should we do? checking trivial vaues?)

cc @datumbox


torchvision/ops/poly_loss.py
```diff
@@ -0,0 +1,66 @@
...
+from ..utils import _log_api_usage_once
+
+
+def poly_loss(
+    x: Tensor,
+    target: Tensor,
+    eps: float = 2.0,
+    weight: Optional[Tensor] = None,
+    ignore_index: int = -100,
+    reduction: str = "mean",
```
oke-aditya, line 15: I also had a thought of making this into an Enum say LossReductionMode.

torchvision/ops/poly_loss.py
```diff
@@ -0,0 +1,66 @@
...
+
+from ..utils import _log_api_usage_once
+
+
+def poly_loss(
+    x: Tensor,
+    target: Tensor,
+    eps: float = 2.0,
+    weight: Optional[Tensor] = None,
+    ignore_index: int = -100,
```
oke-aditya, line 14: this is just to have default ignore as false right? Why not set it to None and check if its not None?

torchvision/ops/poly_loss.py
datumbox: Though I do prefer enums to strings, keeping this a string will ensure strong alignment with Core and will assist on a possible upstreaming on the future.

torchvision/ops/poly_loss.py
datumbox: My understanding is he is following very closely the default values of established losses in Core ([example](https://pytorch.org/docs/stable/generated/torch.nn.CrossEntropyLoss.html)). I think that's a good idea.

oke-aditya (commented):
Sorry for early poke but a few thoughts @frgfm

torchvision/ops/poly_loss.py
oke-aditya: aah I didn't know about this. Nice

torchvision/ops/poly_loss.py
oke-aditya: I'm suggesting to support both strings and LossReductionMode. But that's ok,

frgfm: I hesitated as well, but went for the same option as core

torchvision/ops/poly_loss.py
frgfm: Yup that's what I was going for

torchvision/ops/poly_loss.py
```diff
@@ -0,0 +1,66 @@
...
+        if weight.type() != x.data.type():
+            weight = weight.type_as(x.data)
+        logpt = weight.gather(0, target.data.view(-1)) * logpt
+
+    # Loss reduction
+    if reduction == "sum":
+        loss = loss[valid_idxs].sum()
+    elif reduction == "mean":
+        loss = loss[valid_idxs].mean()
+    else:
```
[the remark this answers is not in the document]
frgfm: I think a `ValueError` in the constructor is much better indeed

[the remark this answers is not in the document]
oke-aditya: We don't have any class based module. Only functional ones for losses. 

cc @datumbox for this this point. I think we can throw value error here too.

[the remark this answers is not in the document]
datumbox: Just being mindful that @frgfm has marked the PR as draft, so I don't want to snoop around too much before he is ready. 

But I think it would be OK to throw a value exception here as this is aligned with a [practice](https://github.com/pytorch/pytorch/blob/c371542efc31b1abfe6f388042aa3ab0cef935f2/torch/nn/_reduction.py#L7-L20) from PyTorch core. We possibly try to go for similar look-and-feel so that if we end up moving it on Core, the implementation behaves similarly.

datumbox:
@frgfm Just wanted to check in and ensure you are not blocked or need any assistance. Let me know if you do. :)

torchvision/ops/poly_loss.py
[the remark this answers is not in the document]
frgfm: Just added a ValueError!

frgfm:
Hello there :) 

My bad, I just caught up with the PR! I left it as a draft because the missing part is the unit test: I saw that the ones for the focal loss are quite extensive. Do I need to be that thorough?

datumbox:
@frgfm Ah yes the unit-tests. I would recommend following the pattern that we have on existing tests and have a similar coverage. Hopefully following the existing infra will make your work easier. The reason we really need them is because previously we spotted a few nasty bugs on some of the losses. We are only couple of weeks prior cutting the branch for the upcoming release so if we want to include this feature we need to have elevated degree of confidence.

PR #6458, opened 2022-08-22
[proto] Ported LinearTransformation

### LINKED ISSUE #5538: Port `transforms.LinearTransformation` to `prototype.transforms`

vfdev-5:
- Ported LinearTransformation
- Added tests

Fixes #5538

### DIFF
--- a/test/test_prototype_transforms.py
+++ b/test/test_prototype_transforms.py
@@ -18,7 +18,7 @@ from test_prototype_transforms_functional import (
 )
 from torchvision.ops.boxes import box_iou
 from torchvision.prototype import features, transforms
-from torchvision.transforms.functional import InterpolationMode, pil_to_tensor, to_pil_image
+from torchvision.transforms.functional import get_image_num_channels, InterpolationMode, pil_to_tensor, to_pil_image
 
 
 def make_vanilla_tensor_images(*args, **kwargs):
@@ -1500,18 +1500,25 @@ class TestLinearTransformation:
     @pytest.mark.parametrize(
         "inpt",
         [
-            122 * torch.ones(1, 3, 32, 32),
-            features.Image(122 * torch.ones(1, 3, 32, 32)),
-            PIL.Image.new("RGB", (32, 32), 122),
+            122 * torch.ones(1, 3, 8, 8),
+            122.0 * torch.ones(1, 3, 8, 8),
+            features.Image(122 * torch.ones(1, 3, 8, 8)),
+            PIL.Image.new("RGB", (8, 8), (122, 122, 122)),
+            PIL.Image.new("F", (8, 8), 122.0),
         ],
     )
     def test__transform(self, inpt):
 
-        v = 121 * torch.ones(3 * 32 * 32)
-        m = torch.ones(3 * 32 * 32, 3 * 32 * 32)
+        c = get_image_num_channels(inpt)
+        v = 121 * torch.ones(c * 8 * 8)
+        m = torch.ones(c * 8 * 8, c * 8 * 8)
         transform = transforms.LinearTransformation(m, v)
 
         output = transform(inpt)
 
         assert isinstance(output, type(inpt))
-        
+        if isinstance(output, PIL.Image.Image):
+            assert np.unique(np.asarray(output)) == c * 8 * 8
+        else:
+            assert output.unique() == c * 8 * 8
+            assert output.dtype == inpt.dtype
--- a/test/test_prototype_transforms.py
+++ b/test/test_prototype_transforms.py
@@ -1487,3 +1487,31 @@ class TestFixedSizeCrop:
         transform(bounding_box)
 
         mock.assert_called_once()
+
+
+class TestLinearTransformation:
+    def test_assertions(self):
+        with pytest.raises(ValueError, match="transformation_matrix should be square"):
+            transforms.LinearTransformation(torch.rand(2, 3), torch.rand(5))
+
+        with pytest.raises(ValueError, match="mean_vector should have the same length"):
+            transforms.LinearTransformation(torch.rand(3, 3), torch.rand(5))
+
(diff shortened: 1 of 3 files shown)


test/test_prototype_transforms.py
```diff
@@ -18,7 +18,7 @@
 )
 from torchvision.ops.boxes import box_iou
 from torchvision.prototype import features, transforms
-from torchvision.transforms.functional import InterpolationMode, pil_to_tensor, to_pil_image
+from torchvision.transforms.functional import get_image_num_channels, InterpolationMode, pil_to_tensor, to_pil_image
```
datumbox, line 21: We shouldn't be using `get_image_num_channels` kernels from `torchvision.transforms.functional` here. Instead we should only be using from the prototype area.

torchvision/prototype/transforms/_misc.py
```diff
@@ -32,6 +35,68 @@ def extra_repr(self) -> str:
...
+        flat_tensor = inpt.view(-1, n) - self.mean_vector
+        transformed_tensor = torch.mm(flat_tensor, self.transformation_matrix)
+        output = transformed_tensor.view(shape)
+
+        if pil_mode is not None:
+            # cast output to input dtype:
+            output = output.to(inpt.dtype)
+            output = F.to_image_pil(output, mode=pil_mode)
+        elif isinstance(inpt, features.Image):
+            output = features.Image.new_like(inpt, output)
```
datumbox, line 95: Given that the linear transform is going to move the values away from the standard scale, making it impossible for us to tell what is the maximum value (needed for many transforms), I think we shouldn't wrap things back as images (PIL or `_feature.Image`) but return things as tensors. Thoughts?

vfdev-5: Actually, I think I should drop PIL Image support as stable LinearTransformation does not support it: https://github.com/pytorch/vision/blob/a409c3fef429076168747d375614e10e72cf035f/torchvision/transforms/transforms.py#L1072-L1079

torchvision/prototype/transforms/_misc.py
```diff
@@ -32,6 +35,68 @@ def extra_repr(self) -> str:
...
+            inpt = F.to_image_tensor(inpt).contiguous()
+
+        shape = inpt.shape
+        n = shape[-3] * shape[-2] * shape[-1]
+        if n != self.transformation_matrix.shape[0]:
+            raise ValueError(
+                "Input tensor and transformation matrix have incompatible shape."
+                + f"[{shape[-3]} x {shape[-2]} x {shape[-1]}] != "
+                + f"{self.transformation_matrix.shape[0]}"
+            )
```
pmeier, line 78: Shouldn't this happen in `_get_params`?

torchvision/prototype/transforms/_misc.py
```diff
@@ -32,6 +35,68 @@ def extra_repr(self) -> str:
...
+
+        if inpt.device.type != self.mean_vector.device.type:
+            raise ValueError(
+                "Input tensor should be on the same device as transformation matrix and mean vector. "
+                f"Got {inpt.device} vs {self.mean_vector.device}"
+            )
```
pmeier, line 84: Given that this is easy to fix, should we maybe only warn and move `mean` and `std` to the same device?

vfdev-5: This is original implementation. Yes, we can move class attributes to provided device.

torchvision/prototype/transforms/_misc.py
vfdev-5: There are no params to generate in this transform. I think moving this code to `_get_params` and returning empty dict will look a bit strange, IMO

pmeier: Well, the `n` is parameter, no?

datumbox (approved):
LGTM on my side, thanks!

torchvision/prototype/transforms/_misc.py
vfdev-5: No, `n` is not a parameter, we only check if input is compatible with `self.transformation_matrix` and that's it. `n` wont change for the inputs.

pmeier: `n` is the product of the number of channels, height, and width. Since it 

> wont change for the inputs.

we should have it in `_get_params` with the 

```py
image = query_image(sample)
num_channels, height, width = F.get_image_dimensions(image)
n = num_channels * height * width

....

return dict(n=n)
```

idiom.

torchvision/prototype/transforms/_misc.py
```diff
@@ -32,6 +35,59 @@ def extra_repr(self) -> str:
...
+
+        self.transformation_matrix = transformation_matrix
+        self.mean_vector = mean_vector
+
+    def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
+
+        if isinstance(inpt, features._Feature) and not isinstance(inpt, features.Image):
+            return inpt
+        elif isinstance(inpt, PIL.Image.Image):
+            raise TypeError("Unsupported input type")
```
pmeier, line 66: Idiomatic this should happen in `forward` with a `has_any` check.

vfdev-5:
@pmeier I do not agree with your `_get_params` comment. Your second idiomatic usage of `has_any` comment seems like an overhead to me. I let you send a follow-up PR with these suggestions if these changes are important. Thanks

torchvision/prototype/transforms/_misc.py
datumbox: @vfdev-5 I think that's worth adding here to be consistent with other implementations. For tiny changes like this and the TODO below (https://github.com/pytorch/vision/pull/6458#discussion_r951206479) it might be worth doing them on the same PR. Otherwise we just increasing the TODO list which makes it harder to track and easier to miss.

PR #6459, opened 2022-08-22
replace `query_image` with `query_image_dimensions`

pmeier:
Compared to the idiom `get_image_dimensions(query_image(...))`, `query_image_dimensions` iterates the whole sample and fails in case if we find different shapes. Thus, we get protection in cases that before silently did the wrong thing or failed with a nondescript error message.

There was one case (`RandomErasing`) where we actually needed the old idiom, since the queried image was used for something else than just the dimension extraction. I was able to refactor it so we don't need to that anymore. LMK if these commits can stay in this PR or should be send individually.


torchvision/prototype/transforms/functional/_augment.py
```diff
@@ -14,3 +17,13 @@ def erase_image_pil(
     t_img = pil_to_tensor(img)
     output = erase_image_tensor(t_img, i=i, j=j, h=h, w=w, v=v, inplace=inplace)
     return to_pil_image(output, mode=img.mode)
+
+
+def erase(inpt: Any, i: int, j: int, h: int, w: int, v: torch.Tensor, inplace: bool = False):
```
pmeier, line 22: This dispatcher was missing. We agreed offline that we should always have a dispatcher in case we have more than one kernel. Thus, although this is a image-only function, a dispatcher is still warranted since we have a tensor and PIL kernel.

torchvision/prototype/transforms/_augment.py
```diff
@@ -79,22 +78,22 @@ def _get_params(self, sample: Any) -> Dict[str, Any]:
 
             i = torch.randint(0, img_h - h + 1, size=(1,)).item()
             j = torch.randint(0, img_w - w + 1, size=(1,)).item()
+
+            needs_erase = True
+            erase_params = dict(i=i, j=j, h=h, w=w, v=v)
+
             break
         else:
-            i, j, h, w, v = 0, 0, img_h, img_w, image
```
pmeier, line 84: We return `image` here to replace the image with itself in the `erase` kernels. Instead of doing that, I've added the `needs_erase` parameter and don't call the dispatcher in case it is not set. This will also give us a performance boost while also remove the need to actually retrieve the image.

vfdev-5: Why can't we return `v = None` and thus do not erase ?

pmeier: You mean use `v is None` as check whether we want to erase or not instead of `needs_erase`? We used the `needs_x` idiom before

https://github.com/pytorch/vision/blob/a409c3fef429076168747d375614e10e72cf035f/torchvision/prototype/transforms/_geometry.py#L850

https://github.com/pytorch/vision/blob/a409c3fef429076168747d375614e10e72cf035f/torchvision/prototype/transforms/_geometry.py#L866

so I thought we can reuse it again to make it very clear what is happening. No strong opinion though.

vfdev-5: I do not think that this idiom is needed here, you define `erase_params = None` for non-erase case and `needs_erase` is useless variable, IMO.
On the other hand, params can remain `dict(i=i, j=j, h=h, w=w, v=v)` and thus to distinguish if we need erase or not, we can check for `v is None`

torchvision/prototype/transforms/_augment.py
```diff
@@ -38,8 +38,7 @@ def __init__(
         self.value = value
 
     def _get_params(self, sample: Any) -> Dict[str, Any]:
-        image = query_image(sample)
-        img_c, img_h, img_w = get_image_dimensions(image)
+        img_c, img_h, img_w = query_image_dimensions(sample)
```
datumbox, line 41: Nit: It might be worth renaming this to to indicate that the method can be used not only for images but also videos. Spatial dims could be an option, though channels are not usually considered part of the spatial dims.

torchvision/prototype/transforms/_utils.py
```diff
@@ -39,6 +31,20 @@ def get_image_dimensions(image: Union[PIL.Image.Image, torch.Tensor, features.Im
     return channels, height, width
 
 
+def query_image_dimensions(sample: Any) -> Tuple[int, int, int]:
+    flat_sample, _ = tree_flatten(sample)
+    image_dimensionss = {
```
datumbox, line 36: Typo:
```suggestion
    image_dimensions = {
```

torchvision/prototype/transforms/_augment.py
pmeier: Well, it is not needed in the cited cases either, but we went for it anyway. We could also use some sentinel values there as well. Let's have @datumbox break the tie :innocent:

torchvision/prototype/transforms/_utils.py
pmeier: Not quite. `image_dimensionss` is a set of `image_dimensions`. Thus, the extra `s` denotes the plural the same way `images` could be a set of `image`. No strong opinion though.

torchvision/prototype/transforms/_augment.py
datumbox: If I were to write this piece of code, I would also probably be passing `None`. But this approach is equally valid, so I'll leave it to the author to decide.

pmeier: I'll go with `None` then.

torchvision/prototype/transforms/_utils.py
datumbox: Given we need to avoid the word `image` all together, you could name this `dim_set` and be explit.

torchvision/prototype/transforms/_augment.py
pmeier: We agreed offline to `{query, get}_chw` since it is concise, factual, and does not "bias" the name towards images.

datumbox (commented):
LGTM 😄

pmeier:
Oops sorry for the premature merge after the offline discussion :smile:

datumbox:
No worries, I verified everything we talked about is good. No need to do anything else.

PR #6460, opened 2022-08-22
remove BatchMultiCrop

### LINKED ISSUE #6455: The prototype `FiveCrop`/`TenCrop`/`BatchMultiCrop` produce unexpected results
### 🐛 Describe the bug

The current implementation of the transforms produce unexpected results for Classification.

- If a single image comes in (no batch&collated), a batched result comes out with a single label
- if a batch of images comes in, we get a new image batch with 5x the original size and a label vector with length equal to the original size

In both cases, the `X` and the `Y` lengths don't match. This creates issues on validation pipelines. 

One approach would be to duplicate the length of Labels to ensure they match. It's quite likely the same approach should be considered for meta-data (like ids) and other content included in the record. Things get even more complex if we consider extending the transform to Detection as we would need to crop the Masks/BBoxes, Cleanup the BBoxes and sync the Labels for each of the new crops.

This complexity is possible the reason why existing the `FiveCrop` implementation on stable don't offer a way to stack the result for the user but instead it just provides an [example](https://github.com/pytorch/vision/blob/main/torchvision/transforms/transforms.py#L992-L1000) in the documentation on how it can be done. Due to the above, I believe that the current implementation of `BatchMultiCrop` is flawed and should be either removed or redeveloped.

### Versions

Latest main branch

cc @vfdev-5 @datumbox @bjuncek @pmeier

pmeier:
Closes #6455.

### DIFF
--- a/torchvision/prototype/transforms/_geometry.py
+++ b/torchvision/prototype/transforms/_geometry.py
@@ -144,7 +144,7 @@ class FiveCrop(Transform):
         ...         images, labels = sample
         ...         batch_size = len(images)
         ...         images = features.Image.new_like(images[0], torch.stack(images))
-        ...         labels = features.Label.new_like(labels, labels.repeat(batch_size, *[1] * labels.ndim))
+        ...         labels = features.Label.new_like(labels, labels.repeat(batch_size))
         ...         return images, labels
         ...
         >>> image = features.Image(torch.rand(3, 256, 256))
@@ -162,6 +162,8 @@ class FiveCrop(Transform):
         self.size = _setup_size(size, error_msg="Please provide only two dimensions (h, w) for size.")
 
     def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
+        # TODO: returning a list is technically BC breaking since FiveCrop returned a tuple before. We switched to a
+        #  list here to align it with TenCrop.
         if isinstance(inpt, features.Image):
             output = F.five_crop_image_tensor(inpt, self.size)
             return [features.Image.new_like(inpt, o) for o in output]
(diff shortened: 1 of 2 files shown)


torchvision/prototype/transforms/_geometry.py
```diff
@@ -136,30 +136,39 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
...
         self.size = _setup_size(size, error_msg="Please provide only two dimensions (h, w) for size.")
 
     def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
         if isinstance(inpt, features.Image):
             output = F.five_crop_image_tensor(inpt, self.size)
-            return MultiCropResult(features.Image.new_like(inpt, o) for o in output)
+            return [features.Image.new_like(inpt, o) for o in output]
         elif is_simple_tensor(inpt):
-            return MultiCropResult(F.five_crop_image_tensor(inpt, self.size))
+            return list(F.five_crop_image_tensor(inpt, self.size))
```
pmeier, line 169: `five_crop_*` currently returns a tuple whereas `ten_crop_*` returns a list. We could full BC here and also return a tuple, but I think later resolving https://github.com/pytorch/vision/issues/6455#issuecomment-1222114489 in a BC manner would be harder due to the different types. Since `MultiCropResult` already was a list, we somewhat agreed to this BC break before.

torchvision/prototype/transforms/_geometry.py
```diff
@@ -136,30 +136,39 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
...
-    :class:`~torchvision.prototype.transforms.BatchMultiCrop`.
+class FiveCrop(Transform):
+    """
+    Example:
+        >>> class BatchMultiCrop(transforms.Transform):
+        ...     def forward(self, sample: Tuple[List[features.Image], features.Label]):
+        ...         images, labels = sample
+        ...         batch_size = len(images)
+        ...         images = features.Image.new_like(images[0], torch.stack(images))
+        ...         labels = features.Label.new_like(labels, labels.repeat(batch_size, *[1] * labels.ndim))
```
pmeier, line 147: If we stick with this minimal example, we could also remove this

```suggestion
        ...         labels = features.Label.new_like(labels, labels.repeat(batch_size))
```

since within this scope the label has a fixed shape, namely a scalar.

datumbox: Nice, let's lose this then.

torchvision/prototype/transforms/_geometry.py
datumbox: Could you please add a comment to highlight that this is potentially BC-breaking and worth revising before releasing it?

torchvision/prototype/transforms/_geometry.py
```diff
@@ -171,6 +180,26 @@ def forward(self, *inputs: Any) -> Any:
 
 
 class TenCrop(Transform):
+    """
+    Example:
```
datumbox, line 184: I would just point to the `FiveCrop` example here to avoid duplicating the example.

datumbox (approved):
LGTM, thanks @pmeier. Just a few nits but nothing too serious.

torchvision/prototype/transforms/_geometry.py
pmeier: Should we maybe have this change on `F.five_crop` rather than on the transform? If we accept this potential BC break on the transform we should accept the same on the functional.

datumbox: That might be a bigger change. Given where this PR is right now, is there a reason why we can't use Tuples and keep BC?

pmeier: Yeah, let's go full BC for now.

PR #6462, opened 2022-08-22
[proto] warning instead of error on device mismatch in LinearTransformation

vfdev-5:
- Warning instead of error on device mismatch in LinearTransformation
- Added a test

### DIFF
--- a/torchvision/prototype/transforms/_misc.py
+++ b/torchvision/prototype/transforms/_misc.py
@@ -1,4 +1,5 @@
 import functools
+import warnings
 from typing import Any, Callable, Dict, List, Sequence, Type, Union
 
 import PIL.Image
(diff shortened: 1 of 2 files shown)


torchvision/prototype/transforms/_misc.py
```diff
@@ -77,15 +78,20 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
                 + f"{self.transformation_matrix.shape[0]}"
             )
 
-        if inpt.device.type != self.mean_vector.device.type:
-            raise ValueError(
-                "Input tensor should be on the same device as transformation matrix and mean vector. "
-                f"Got {inpt.device} vs {self.mean_vector.device}"
+        input_device = inpt.device
+        if input_device.type != self.mean_vector.device.type:
```
pmeier, line 82: Maybe we can move the `mean` and `std` rather than the `inpt`? My guess is that people will rather forget to move `mean` and `std` on GPU than the images. If that is the case, we will move the images to CPU and thus slow down the whole pipeline.

Maybe we don't even need to move the `mean` and `std` back since my guess is that the inputs will always be on the same device. Thus, we only need to move whenever the device changes, which is probably only once.

datumbox: I agree that we should probably be moving mean/std.

Though I have to admit that this sounded like a good idea originally but now we need to make more and more assumptions about it. Perhaps that's the reason originally the author of the transform took the path of throwing an error rather than trying to move things around.

vfdev-5: So, what option we take finally:
- do not move devices around and throw an error ?
- move input to mean/transformation_matrix's device ?
- move mean/transformation_matrix to input's device ?

datumbox: I would vote for option 1 or 3. I would avoid option 2. If you end up doing option 3, don't overwrite the `self` attributes for mean/matrix.

vfdev-5: > If you end up doing option 3, don't overwrite the self attributes for mean/matrix.

so, if all images are on GPU and mean/transformation_matrix are on CPU, then we move mean/transformation_matrix to GPU all the time instead of once ?

vfdev-5:
As discussed offline we can close this PR and come back to that once have more clear vision of what we would like.

PR #6463, opened 2022-08-22
cleanup prototype auto augment transforms

pmeier:
While going through the codebase with @vfdev-5 and @datumbox, we noticed that the auto augment transforms were not using the `_get_params` / `_transform` idiom. Looking into it, I don't think there is currently a reason to not do that. I guess this was a limitation of the past design.

### DIFF
--- a/torchvision/prototype/transforms/_auto_augment.py
+++ b/torchvision/prototype/transforms/_auto_augment.py
@@ -1,4 +1,5 @@
 import math
+import numbers
 from typing import Any, Callable, cast, Dict, List, Optional, Sequence, Tuple, TypeVar, Union
 
 import PIL.Image
@@ -10,7 +11,6 @@ from torchvision.transforms.autoaugment import AutoAugmentPolicy
 from torchvision.transforms.functional import InterpolationMode, pil_to_tensor, to_pil_image
 
 from ._utils import is_simple_tensor, query_chw
-
 
 K = TypeVar("K")
 V = TypeVar("V")
@@ -25,6 +25,9 @@ class _AutoAugmentBase(Transform):
     ) -> None:
         super().__init__()
         self.interpolation = interpolation
+
+        if not isinstance(fill, (numbers.Number, tuple, list)):
+            raise TypeError("Got inappropriate fill arg")
         self.fill = fill
 
     def _get_random_item(self, dct: Dict[K, V]) -> Tuple[K, V]:
@@ -32,25 +35,9 @@ class _AutoAugmentBase(Transform):
         key = keys[int(torch.randint(len(keys), ()))]
         return key, dct[key]
 
-    def _parse_fill(
-        self, image: Union[PIL.Image.Image, torch.Tensor, features.Image], num_channels: int
-    ) -> Union[int, float, Sequence[int], Sequence[float]]:
-        fill = self.fill
-
-        if isinstance(image, PIL.Image.Image) or fill is None:
-            return fill
-
-        if isinstance(fill, (int, float)):
-            fill = [float(fill)] * num_channels
-        else:
-            fill = [float(f) for f in fill]
-
-        return fill
-
     def _get_params(self, sample: Any) -> Dict[str, Any]:
-        num_channels, height, width = query_chw(sample)
-        fill = self._parse_fill(image, num_channels)
-        return dict(num_channels=num_channels, height=height, width=width, fill=fill)
+        _, height, width = query_chw(sample)
+        return dict(height=height, width=width)
 
     def _apply_image_transform(
         self,
@@ -279,7 +266,7 @@ class AutoAugment(_AutoAugmentBase):
                 magnitude = 0.0
 
             inpt = self._apply_image_transform(
-                inpt, transform_id, magnitude, interpolation=self.interpolation, fill=params["fill"]
+                inpt, transform_id, magnitude, interpolation=self.interpolation, fill=self.fill
             )
 
         return inpt
@@ -344,7 +331,7 @@ class RandAugment(_AutoAugmentBase):
                 magnitude = 0.0
 
             inpt = self._apply_image_transform(
-                inpt, transform_id, magnitude, interpolation=self.interpolation, fill=params["fill"]
+                inpt, transform_id, magnitude, interpolation=self.interpolation, fill=self.fill
             )
 
         return inpt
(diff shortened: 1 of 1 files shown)


torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -38,27 +31,6 @@ def _get_random_item(self, dct: Dict[K, V]) -> Tuple[K, V]:
...
-                images.append((id, inpt))
-            elif isinstance(inpt, unsupported_types):
-                raise TypeError(f"Inputs of type {type(inpt).__name__} are not supported by {type(self).__name__}()")
-
-        if not images:
-            raise TypeError("Found no image in the sample.")
-        if len(images) > 1:
-            raise TypeError(
-                f"Auto augment transformations are only properly defined for a single image, but found {len(images)}."
-            )
```
pmeier, line 59: This check will be temporarily removed by this PR, but will be re-added by #6459.

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -478,16 +451,13 @@ def _sample_dirichlet(self, params: torch.Tensor) -> torch.Tensor:
...
-        num_channels, height, width = get_image_dimensions(orig_image)
-        fill = self._parse_fill(orig_image, num_channels)
-
-        if isinstance(orig_image, torch.Tensor):
-            image = orig_image
-        else:  # isinstance(inpt, PIL.Image.Image):
-            image = pil_to_tensor(orig_image)
+    def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
+        if isinstance(inpt, features.Image) or is_simple_tensor(inpt):
+            image = inpt
```
datumbox, line 456: Nit: We should start renaming `image` to `inpt`. Videos will be passed here soon.

datumbox (approved):
LGTM, thanks!

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -74,6 +47,11 @@ def _parse_fill(
 
         return fill
 
+    def _get_params(self, sample: Any) -> Dict[str, Any]:
+        num_channels, height, width = query_chw(sample)
+        fill = self._parse_fill(image, num_channels)
```
pmeier, line 52: ARGH, we need the actual image here due to 

https://github.com/pytorch/vision/blob/857c03031c4bedeea34e489fd9b1b7427b03f4bc/torchvision/prototype/transforms/_auto_augment.py#L67-L68

I'll try to refactor so we can use the new `query_chw`.

pmeier: The AutoAugment transforms are an outlier of how `fill` is handled. Internally they call `F.affine`. Our `RandomAffine` does the same but only checks the values statically with

https://github.com/pytorch/vision/blob/1a1fea3453c6f329168f91c7cdfe95e043661256/torchvision/prototype/transforms/_geometry.py#L211-L213

Inside `F.affine` we do 

https://github.com/pytorch/vision/blob/1a1fea3453c6f329168f91c7cdfe95e043661256/torchvision/prototype/transforms/functional/_geometry.py#L374-L384

but only for tensor images

https://github.com/pytorch/vision/blob/1a1fea3453c6f329168f91c7cdfe95e043661256/torchvision/prototype/transforms/functional/_geometry.py#L412-L413

datumbox: @vfdev-5 has changed quite a few things to handle fill easier. Do we still need this handling or can we move it elsewhere?

pmeier: I think we can get rid of the special handling in the auto augment transforms entirely. Both there and in `F.affine` don't do anything for PIL images. The only difference on the surface is that the auto augment version expands `fill` to match the number of channels in case we get a scalar

https://github.com/pytorch/vision/blob/1a1fea3453c6f329168f91c7cdfe95e043661256/torchvision/prototype/transforms/_auto_augment.py#L70-L71

which is not done for `F.affine`

https://github.com/pytorch/vision/blob/1a1fea3453c6f329168f91c7cdfe95e043661256/torchvision/prototype/transforms/functional/_geometry.py#L383

But that is only a surface difference, since later in the kernel we do 

https://github.com/pytorch/vision/blob/1a1fea3453c6f329168f91c7cdfe95e043661256/torchvision/transforms/functional_tensor.py#L560

`fill.expand_as(image)` will broadcast to the right shape, since a singleton dimension is broadcastable with any number of channels.

pmeier: Regardless of the image type and the (valid) type of `fill`, whether or not we call `AutoAugment._parse_fill` makes no difference:

```py
import functools
import itertools

import torch
from torchvision.prototype.transforms._utils import get_chw

affine = functools.partial(F.affine, angle=30, shear=[0, 0], translate=[0, 0], scale=1)

image_tensor = torch.rand(3, 16, 16)
image_pil = F.to_image_pil(image_tensor)
image_feature = features.Image(image_tensor)

for image, fill_plain in itertools.product(
    [image_tensor, image_pil, image_feature],
    [None, 0, 0.5, [0.1, 0.5, 0.9], (0.1, 0.5, 0.9)],
):
    output_plain = affine(image, fill=fill_plain)

    num_channels, _, _ = get_chw(image)
    fill_aa = transforms.AutoAugment(fill=fill_plain)._parse_fill(image, num_channels)
    output_aa = affine(image, fill=fill_aa)

    if fill_aa != fill_plain:
        print(f"Mismatch for {type(image)}: fill_plain={fill_plain}, fill_aa={fill_aa}")
    torch.testing.assert_close(F.to_image_tensor(output_aa), F.to_image_tensor(output_plain), atol=0, rtol=0)
```

```
Mismatch for <class 'torch.Tensor'>: fill_plain=0, fill_aa=[0.0, 0.0, 0.0]
Mismatch for <class 'torch.Tensor'>: fill_plain=0.5, fill_aa=[0.5, 0.5, 0.5]
Mismatch for <class 'torch.Tensor'>: fill_plain=(0.1, 0.5, 0.9), fill_aa=[0.1, 0.5, 0.9]
Mismatch for <class 'torchvision.prototype.features._image.Image'>: fill_plain=0, fill_aa=[0.0, 0.0, 0.0]
Mismatch for <class 'torchvision.prototype.features._image.Image'>: fill_plain=0.5, fill_aa=[0.5, 0.5, 0.5]
Mismatch for <class 'torchvision.prototype.features._image.Image'>: fill_plain=(0.1, 0.5, 0.9), fill_aa=[0.1, 0.5, 0.9]
```

We should remove.

vfdev-5: Yes, I think you can remove AA._parse_fill, `F.affine` and `Image.affine` should be able to handle channels expansion.

datumbox (approved):

PR #6464, opened 2022-08-22
fix MixUp and CutMix

pmeier:
I dropped the ball for two separate things in #6447:

- PIL is not supported by `MixUp` and `CutMix` since we need a batch of images (read batched and collated). 
- Just adding plain tensors to the allowed types is not sufficient and `_transforms` has to be adapted.


torchvision/prototype/transforms/_augment.py
```diff
@@ -105,9 +105,7 @@ def __init__(self, *, alpha: float, p: float = 0.5) -> None:
 
     def forward(self, *inpts: Any) -> Any:
         sample = inpts if len(inpts) > 1 else inpts[0]
-        if not (
-            has_any(sample, features.Image, PIL.Image.Image, is_simple_tensor) and has_any(sample, features.OneHotLabel)
-        ):
+        if not (has_any(sample, features.Image, is_simple_tensor) and has_any(sample, features.OneHotLabel)):
```
vfdev-5, line 108: To understand better, PIL Image was not and was bypassed previously and now we have an explicit error on PIL Image input ?

PR #6474, opened 2022-08-23
allow grayscale tensor images in `rgb_to_grayscale`

### LINKED ISSUE #5581: RandomGrayscale fails for grayscale tensor inputs

pmeier:
Fixes #5581.


torchvision/prototype/transforms/functional/_deprecated.py
```diff
@@ -0,0 +1,71 @@
...
+import PIL.Image
+import torch
+
+from torchvision.prototype import features
+
+from ._meta import convert_color_space, convert_color_space_image_pil
+
+
+# TODO: this was copied from `torchvision.prototype.transforms._utils`. Given that this is not related to pytree / the
+#  Transform object, we should probably move it to `functional._utils`.
```
vfdev-5, line 13: I do not understand this TODO. Can we resolve it right now ?

torchvision/prototype/transforms/functional/_deprecated.py
```diff
@@ -0,0 +1,71 @@
...
+def is_simple_tensor(inpt: Any) -> bool:
+    return isinstance(inpt, torch.Tensor) and not isinstance(inpt, features._Feature)
```
datumbox, line 15: Perhaps you are right that the original location at `torchvision.prototype.transforms._utils` might not be the best, but it's better than placing it here. Have you considered moving it to the `features` package? It's literally something that checks the type of the input. At any case, I wouldn't move this in this PR.

torchvision/transforms/functional_tensor.py
```diff
@@ -145,16 +145,19 @@ def crop(img: Tensor, top: int, left: int, height: int, width: int) -> Tensor:
 def rgb_to_grayscale(img: Tensor, num_output_channels: int = 1) -> Tensor:
     if img.ndim < 3:
         raise TypeError(f"Input image tensor should have at least 3 dimensions, but found {img.ndim}")
```
datumbox, line 147: Question: What if input is a 2d grayscale image `img.ndim == 2`? Should we unsqueeze and treat it as 1xHxW? We handle this in some limited kernels. Thoughts?

torchvision/transforms/functional_tensor.py
```diff
@@ -145,16 +145,19 @@ def crop(img: Tensor, top: int, left: int, height: int, width: int) -> Tensor:
...
-    l_img = (0.2989 * r + 0.587 * g + 0.114 * b).to(img.dtype)
-    l_img = l_img.unsqueeze(dim=-3)
+    if img.shape[-3] == 3:
+        r, g, b = img.unbind(dim=-3)
+        # This implementation closely follows the TF one:
+        # https://github.com/tensorflow/tensorflow/blob/v2.3.0/tensorflow/python/ops/image_ops_impl.py#L2105-L2138
+        l_img = (0.2989 * r + 0.587 * g + 0.114 * b).to(img.dtype)
+        l_img = l_img.unsqueeze(dim=-3)
+    else:
+        l_img = img
```
datumbox, line 160: Should we be cloning?

torchvision/prototype/transforms/functional/_deprecated.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def to_grayscale(inpt: PIL.Image.Image, num_output_channels: int = 1) -> PIL.Image.Image:
```
datumbox, line 18: I don't understand, why aren't we calling the legacy kernel from `_F` after throwing the warnings?

torchvision/prototype/transforms/functional/_deprecated.py
```diff
@@ -0,0 +1,71 @@
...
+    # TODO: Can we even deprecate in favor of `convert_color_space` in such a case? Should we tell the users to call
+    #  `image.convert("RGB")` on their own?
+    output = inpt.convert("RGB")
+
+    if num_output_channels == 3:
+        output = convert_color_space_image_pil(inpt, color_space=features.ColorSpace.RGB)
+    return output
+
+
+def rgb_to_grayscale(inpt: Any, num_output_channels: int = 1) -> Any:
```
datumbox, line 45: Same here, why not call `_F.rgb_to_grayscale`?

datumbox (commented):
Thanks @pmeier, I have a few questions:

torchvision/transforms/functional_tensor.py
vfdev-5: Yes, we could support 2d image tensors as we accept 2d tensors as images:
https://github.com/pytorch/vision/blob/fb35989df6e94f5dd0fbbf11027646f521e7bbbe/torchvision/transforms/functional_tensor.py#L9-L10

torchvision/prototype/transforms/functional/_deprecated.py
pmeier: See https://github.com/pytorch/vision/pull/6474#pullrequestreview-1082060405.

torchvision/prototype/transforms/functional/_deprecated.py
pmeier: > Have you considered moving it to the `features` package?

SGTM.

torchvision/transforms/functional_tensor.py
pmeier: Do we want to add new features to a deprecated functional that are not bug fixes?

torchvision/prototype/transforms/functional/_deprecated.py
pmeier: In the past we opted to not use deprecated behavior. The latest of these was #6454, but #5675 is the same as this PR but for the transforms.

torchvision/prototype/transforms/functional/_deprecated.py
pmeier: Let's discuss this in https://github.com/pytorch/vision/pull/6474#discussion_r952554861.

torchvision/prototype/transforms/functional/_deprecated.py
pmeier: As discussed offline, we will just call the legacy kernels.

torchvision/transforms/functional_tensor.py
pmeier: As discussed offline, we are not going to add support for 2D images here, since we are not sure if they are supported throughout the API in general. Consensus was that we either need to accept them everywhere or only in the "to tensor image" logic.

torchvision/prototype/transforms/functional/_deprecated.py
```diff
@@ -0,0 +1,53 @@
...
+        f"The function `to_grayscale(...{call})` is deprecated in will be removed in a future release. "
+        f"Instead, please use `{replacement}`.",
+    )
+
+    return _F.to_grayscale(inpt, num_output_channels=num_output_channels)
+
+
+def rgb_to_grayscale(inpt: Any, num_output_channels: int = 1) -> Any:
+    if num_output_channels not in (1, 3):
+        raise ValueError("num_output_channels should be either 1 or 3")
```
datumbox, line 27: I think this can be removed because it's being checked on the  `_F` kernel.

datumbox (approved):
LGTM, just a minor optional nit:

PR #6475, opened 2022-08-23
replace _Feature.new_like with _Feature().copy_metadata_to

pmeier:
This was proposed offline by @vfdev-5 as possible simplification. Instead of the idiom `features.Image.new_like(input, output)`, we could have an instance rather than a class method like `input.copy_metadata_to(output)`.

Although `_Feature.new_like` is a class method, we always need an instance as reference. Thus, there will never be a case where we need to instantiate an object just to use the instance method were before we could have used the classmethod.

I quite like this. The only issue I have is naming. `features.Image.new_like(ref, data)` is clear, but is the same true for `ref.copy_metadata_to(data)`?


datumbox:
+1 on the for making it an instance method. I agree the name is an issue; I feel like `new_like()` was very clear while the new one is not.

I like the fact we tried `*_like` naming style which is similar to `zeros_like()` and `full_like()` for the Class method. On core the tensor instances use the `new_*` naming style (see [here](https://pytorch.org/docs/stable/tensors.html#tensor-class-reference)); examples `new_full()`, `new_empty()`, `new_zeroes()` etc. Perhaps `new_instance()`? In all previous examples from core, meta-data are implied the same so things like types and devices are inherited form the tensor.

vfdev-5:
Yes, naming is hard. Here are my attempts to find an appropriate name : 
```
output = inpt.feature_from(tensor)
# or
output = inpt.make_feature(tensor)
# or
output = inpt.make_feature_from(tensor)
# or
output = inpt.create_feature_from(tensor)
# or 
output = inpt.create_from(tensor)
```

datumbox:
I think we should avoid the word `feature`. The `_feature` name has been flagged multiple times are problematic. Also the problem with the `_from` naming is that it implies that it copies the meta-data from the right side (argument) not from the left side (instance). I believe that's why Philip proposes the `_to` naming, which implies the opposite direction.

vfdev-5:
We may also see this op like `torch.from_numpy(nparray) -> tensor`. In our case we give a tensor and output Image, BBox etc: `inpt.from_tensor(output) -> feature` ...

datumbox:
@vfdev-5 I like it. Though to be honest, I would expect it to be a static method and not copy meta-data. Still I think it's a viable option if we don't find something else.

@pmeier how do you feel about it?

pmeier:
> Though to be honest, I would expect it to be a static method and not copy meta-data.

In that case we are back at the start. Calling a static method (or in our case a class method) with a reference and some data.

> `inpt.from_tensor(output) -> feature`

I think that has two caveats:

1. What we named `output` does not have to be a tensor. Something like `features.BoundingBox.new_like(ref, [0, 0, 5, 5])` is perfectly fine. That would translate to `ref.from_tensor([0, 0, 5, 5])` which looks weird.
2. As Vasilis mentioned above, the "from" naming does not make it clear that we are going to keep the metadata that we already have.

---

IMO the best names we have right now are

- `copy_metadata_to` and
- `new_instance`

Both have their issues though. If we don't find a compelling name, I don't think we should force the instance method if the old way has the better naming.

datumbox:
> I don't think we should force the instance method if the old way has the better naming.

+1 

Though I totally see how the instance method is less verbose and looks better, an improper naming will do more harm. 


> That would translate to ref.from_tensor([0, 0, 5, 5]) which looks weird.

Can you elaborate on why this looks weird? Not trying to debate, just trying to understand what you dont like. Is it that it's not immediately obvious what ref is?

pmeier:
> Can you elaborate on why this looks weird?

The method is called `from_tensor` and you are not putting in a tensor.

datumbox:
> The method is called from_tensor and you are not putting in a tensor.

Is this true? I think in the vast majority of cases (if not all?) it's a tensor. It became a tensor after passing it through the low-level kernel that strips subclassing or by applying an operator on it other than to/clone.

pmeier:
It is true for the example I gave:

```py
new = ref.from_tensor([0, 0, 5, 5]) 
```

I agree that for most of our cases on the transforms side we only use tensors as input. Still, this is a public method, so we should try not to have a confusing UI.

That being said, I see very few use cases for using `from_tensor` for stuff that is not coming out of a previous computation, which most likely involves a tensor. The fact that the method works with other inputs as well, is somewhat irrelevant if no one uses it that way.

pmeier:
Closing this for now since we couldn't find a fitting name.

PR #6476, opened 2022-08-23
Eliminate runtime cyclic dependencies

datumbox:

### DIFF
--- a/torchvision/prototype/features/_feature.py
+++ b/torchvision/prototype/features/_feature.py
@@ -18,7 +18,7 @@ class _Feature(torch.Tensor):
         device: Optional[Union[torch.device, str, int]] = None,
         requires_grad: bool = False,
     ) -> F:
-        feature = cast(
+        return cast(
             F,
             torch.Tensor._make_subclass(
                 cast(_TensorBase, cls),
@@ -26,13 +26,6 @@ class _Feature(torch.Tensor):
                 requires_grad,
             ),
         )
-
-        # To avoid circular dependency between features and transforms
-        from ..transforms import functional
-
-        feature._F = functional  # type: ignore[attr-defined]
-
-        return feature
 
     @classmethod
     def new_like(
@@ -98,6 +91,13 @@ class _Feature(torch.Tensor):
         # If that ever gets implemented, remove this in favor of the solution on the `torch.Tensor` class.
         extra_repr = ", ".join(f"{key}={value}" for key, value in kwargs.items())
         return f"{super().__repr__()[:-1]}, {extra_repr})"
+
+    @property
+    def _F(self):
+        if not hasattr(self, "__F"):
+            from ..transforms import functional
+            self.__F = functional
+        return self.__F
 
     def horizontal_flip(self) -> _Feature:
         return self
--- a/torchvision/prototype/features/_feature.py
+++ b/torchvision/prototype/features/_feature.py
@@ -18,7 +18,7 @@ class _Feature(torch.Tensor):
         device: Optional[Union[torch.device, str, int]] = None,
         requires_grad: bool = False,
     ) -> F:
-        return cast(
+        feature = cast(
             F,
             torch.Tensor._make_subclass(
                 cast(_TensorBase, cls),
@@ -26,6 +26,13 @@ class _Feature(torch.Tensor):
                 requires_grad,
             ),
         )
+
+        # To avoid circular dependency between features and transforms
+        from ..transforms import functional
+
+        feature._F = functional  # type: ignore[attr-defined]
+
+        return feature
 
     @classmethod
     def new_like(
--- a/torchvision/prototype/features/_bounding_box.py
+++ b/torchvision/prototype/features/_bounding_box.py
@@ -63,7 +63,7 @@ class BoundingBox(_Feature):
         )
 
     def to_format(self, format: Union[str, BoundingBoxFormat]) -> BoundingBox:
-        from torchvision.prototype.transforms import functional as _F
+        _F = self._F
 
         if isinstance(format, str):
             format = BoundingBoxFormat.from_str(format.upper())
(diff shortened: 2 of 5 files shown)


pmeier:
Given that there were problems with the `DataLoader` before, here are the tests that we use for the prototype datasets:

https://github.com/pytorch/vision/blob/b4b246a52ebbe29bff9099c6fbabff0b82d7edd3/test/test_prototype_builtin_datasets.py#L104-L135

Here is a version to run the same tests for different transforms:

```py
import pickle

import pytest
import torch

from torch.utils.data import DataLoader
from torchdata.datapipes.iter import IterableWrapper
from torchvision.prototype import transforms, features

# TODO: fill me up for more comprehensive tests!
TRANSFORMS = [
    transforms.Resize((8, 8)),
]

transforms = pytest.mark.parametrize("transform", TRANSFORMS, ids=lambda transform: type(transform).__name__)


@pytest.mark.parametrize("only_datapipe", [False, True])
@transforms
def test_traversable(transform, only_datapipe):
    dp = IterableWrapper([]).map(transform)

    traverse(dp, only_datapipe=only_datapipe)


@transforms
def test_serializable(transform):
    dp = IterableWrapper([]).map(transform)

    pickle.dumps(dp)


def _collate_fn(batch):
    return batch


@pytest.mark.parametrize("num_workers", [0, 1])
@transforms
def test_data_loader(transform, num_workers):
    dp = IterableWrapper([features.Image(torch.rand(3, 16, 16)) for _ in range(5)], deepcopy=False).map(transform)

    dl = DataLoader(
        dp,
        batch_size=2,
        num_workers=num_workers,
        collate_fn=_collate_fn,
    )

    list(dl)
```

So far the PR seems good.

pmeier:
You need at least two transforms

```py
transforms.Compose(
    [
        transforms.DecodeImage(),
        transforms.Resize(8, 8),
    ]
)
```

Otherwise, you will have no images inside the sample.

I'm ok with adding these there given that our current transforms compatibility tests are quite "minimal":

https://github.com/pytorch/vision/blob/b4b246a52ebbe29bff9099c6fbabff0b82d7edd3/test/test_prototype_builtin_datasets.py#L98-L102

@NicolasHug do you have an opinion here?

test/test_prototype_builtin_datasets.py
```diff
@@ -44,6 +44,14 @@ def test_coverage():
 
 @pytest.mark.filterwarnings("error")
 class TestCommon:
+
+    _TRANSFORMS = transforms.Compose(
```
pmeier, line 48: Please add a comment that these transforms are a proxy for a preprocessing pipeline that we use to have an "end-to-end" test.

test/test_prototype_builtin_datasets.py
```diff
@@ -104,13 +112,13 @@ def test_transformable(self, dataset_mock, config):
     @pytest.mark.parametrize("only_datapipe", [False, True])
     @parametrize_dataset_mocks(DATASET_MOCKS)
     def test_traversable(self, dataset_mock, config, only_datapipe):
-        dataset, _ = dataset_mock.load(config)
+        dataset = dataset_mock.load(config)[0].map(self._TRANSFORMS)
```
pmeier, line 115: Could you do this in two lines, to have the same loading call on all tests?

```suggestion
        dataset, _ = dataset_mock.load(config)
        dataset = dataset.map(self._TRANSFORMS)
```

pmeier (commented):
I'm fine with the changes to the datasets tests minus some small nits. There is `test_transformable` which is outside of the scope that GH let's me comment. It will be obsolete after your changes and you can delete it.

 Given that only the `test_data_loader` actually executes the transform, I don't think we should see a significant slow down. Still, let's wait for CI to complete to compare the timings. @NicolasHug any objections?

datumbox:
@pmeier I'll remove the extra tests to give you and @NicolasHug to finalize the strategy. They are not strictly needed for this work.

BTW I already see some unrelated [failures](https://github.com/pytorch/vision/runs/7991142663?check_suite_focus=true) related to the resize on bboxes that might be worth looking into.

torchvision/prototype/features/_feature.py
```diff
@@ -97,6 +97,10 @@ def _make_repr(self, **kwargs: Any) -> str:
 
     @property
     def _F(self) -> ModuleType:
+        # This implements a lazy import of the functional to get around the cyclic import. This import is deferred
+        # until the first time we need reference to the functional module and it's shared across all instances of
+        # the class. The deferred lazy import avoids the DataLoader issue described at
+        # https://github.com/pytorch/vision/pull/6219#discussion_r912721945
```
pmeier, line 103: I'm kind of anxious about this since we don't have a repro. @vfdev-5 could you try to replicate the error you saw? All our current test indicate that the lazy loading indeed works with the DataLoader, but maybe we also have a gap in the tests.

pmeier (approved):
Apart from my paranoia in https://github.com/pytorch/vision/pull/6476#discussion_r953588621, I think this is GTG.

torchvision/prototype/features/_feature.py
datumbox: You are right to be anxious. I did reproduce the issue first to ensure I'm not breaking anything. See commit 692d6b9459f14092771c8229531a93a0a6ad6f6a which restored an earlier approach that @vfdev-5 and I came up with and confirmed that it was still failing on my setup. I've also ran the new tests that you proposed (and rolled back) and all passed with exception to the unrelated voc issue. This doesn't actually guarantees we don't have a gap on the tests, but at least we covered the basics. 

I'm happy to do more testing to make you more comfortable with this. We can also revert this PR and reinstate the runtime imports if it turns out to be causing issues. Thoughts?

pmeier: >  I did reproduce the issue first to ensure I'm not breaking anything. 

Ah, I didn't know that. In that case I'm fine now, especially since we can always revert.

pmeier: Maybe we can link this discussion instead?

datumbox:
Merging, the failing tests are unrelated and are caused by dependencies issues from core.

PR #6480, opened 2022-08-24
use the same base class for Label and OneHotLabel

pmeier:
Follow-up to https://github.com/pytorch/vision/pull/6420#issuecomment-1225374947. TL;DR: having a common base class for `Label` and `OneHotLabel` let's us reuse some common logic without risking one being mistaken for the other in the transforms.

### DIFF
--- a/torchvision/prototype/features/_label.py
+++ b/torchvision/prototype/features/_label.py
@@ -8,7 +8,7 @@ from torch.utils._pytree import tree_map
 from ._feature import _Feature
 
 
-F = TypeVar("F", bound="_Feature")
+F = TypeVar("F", bound="_LabelBase")
 
 
 class _LabelBase(_Feature):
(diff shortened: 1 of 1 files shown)


torchvision/prototype/features/_label.py
```diff
@@ -1,57 +1,60 @@
...
+from typing import Any, Optional, Sequence, Type, TypeVar, Union
 
 import torch
 from torch.utils._pytree import tree_map
 
 from ._feature import _Feature
 
 
-class Label(_Feature):
+F = TypeVar("F", bound="_Feature")
```
datumbox, line 11: Did you mean the following?
```suggestion
F = TypeVar("F", bound="_LabelBase")
```

If yes should it be `L` instead of `F`?

datumbox (approved):
LGTM, just one question:

PR #6482, opened 2022-08-24
fix RandomPhotometricDistort

pmeier:
Found this while using the transforms for the detection references:

https://github.com/pytorch/vision/blob/c3ec69482a5b9359306c58f0f5c587e1b3bb11cb/references/detection/presets.py#L39

### DIFF
--- a/torchvision/prototype/transforms/_color.py
+++ b/torchvision/prototype/transforms/_color.py
@@ -105,7 +105,7 @@ class RandomPhotometricDistort(Transform):
         return dict(
             zip(
                 ["brightness", "contrast1", "saturation", "hue", "contrast2"],
-                (torch.rand(6) < self.p).tolist(),
+                (torch.rand(5) < self.p).tolist(),
             ),
             contrast_before=bool(torch.rand(()) < 0.5),
             channel_permutation=torch.randperm(num_channels) if torch.rand(()) < self.p else None,
--- a/torchvision/prototype/transforms/_color.py
+++ b/torchvision/prototype/transforms/_color.py
@@ -105,9 +105,9 @@ class RandomPhotometricDistort(Transform):
         return dict(
             zip(
                 ["brightness", "contrast1", "saturation", "hue", "contrast2"],
-                torch.rand(6) < self.p,
+                (torch.rand(6) < self.p).tolist(),
             ),
-            contrast_before=torch.rand(()) < 0.5,
+            contrast_before=bool(torch.rand(()) < 0.5),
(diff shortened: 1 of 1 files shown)


torchvision/prototype/transforms/_color.py
```diff
@@ -105,9 +105,9 @@ def _get_params(self, sample: Any) -> Dict[str, Any]:
         return dict(
             zip(
                 ["brightness", "contrast1", "saturation", "hue", "contrast2"],
-                torch.rand(6) < self.p,
+                (torch.rand(6) < self.p).tolist(),
             ),
-            contrast_before=torch.rand(()) < 0.5,
+            contrast_before=bool(torch.rand(()) < 0.5),
```
pmeier, line 110: This and the one above are mostly cosmetic to have an actual bool rather than a boolean tensor scalar. This has no effect on the runtime since `if x` invokes `bool(x)` anyway if `x` is not a boolean.

datumbox (approved):
Looks good. Shall we start adding some tests to avoid similar bugs?

torchvision/prototype/transforms/_color.py
```diff
@@ -105,9 +105,9 @@ def _get_params(self, sample: Any) -> Dict[str, Any]:
         return dict(
             zip(
                 ["brightness", "contrast1", "saturation", "hue", "contrast2"],
-                torch.rand(6) < self.p,
+                (torch.rand(6) < self.p).tolist(),
```
datumbox, line 108:
```suggestion
                (torch.rand(5) < self.p).tolist(),
```

torchvision/prototype/transforms/_color.py
```diff
@@ -147,7 +147,7 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
             inpt = F.adjust_contrast(
                 inpt, contrast_factor=ColorJitter._generate_value(self.contrast[0], self.contrast[1])
             )
-        if params["channel_permutation"]:
+        if params["channel_permutation"] is not None:
```
pmeier, line 150: This is an actual bug. While `if x` works correctly for `None`, i.e. no channel permutation, it fails if we actually have one:

```py
>>> bool(None)
False
>>> bool(torch.tensor([2, 0, 1]))
RuntimeError: Boolean value of Tensor with more than one value is ambiguous
```

datumbox:
@pmeier I'm OK if you don't want to do this in this PR but we need to start thinking about the strategy soon. We should not rely on references for testing this.

PR #6483, opened 2022-08-24
[proto] Removed F.label_to_one_hot and added tests for LabelToOneHot

vfdev-5:


datumbox (approved):
LGTM, thanks!

torchvision/prototype/transforms/_type_conversion.py
```diff
@@ -20,19 +21,18 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
 
 
 class LabelToOneHot(Transform):
+    _transformed_types = (features.Label,)
```
datumbox, line 24: Good catch for using the new idiom. We should look for other transforms that use old styles and refactor.

pmeier (approved):

PR #6484, opened 2022-08-24
check full sample in query_bounding_box

pmeier:
This is complementary to #6459. 

---

Currently we have three transformations that don't support batched inputs:

- https://github.com/pytorch/vision/blob/b6feccbc4387766b76a3e22b13815dbbbfa87c0f/torchvision/prototype/transforms/_geometry.py#L618
- https://github.com/pytorch/vision/blob/b6feccbc4387766b76a3e22b13815dbbbfa87c0f/torchvision/prototype/transforms/_geometry.py#L777
- https://github.com/pytorch/vision/blob/b6feccbc4387766b76a3e22b13815dbbbfa87c0f/torchvision/prototype/transforms/_misc.py#L142

Here "batched" means "list of data points" rather than "tensor with batch dimensions", since the latter is not possible in a general way if each data point does not have a fixed number of bounding boxes. 

One thing all have in common is that they call `query_bounding_box` early on in `_get_params`. Currently this just extracts the first bounding box it finds in the sample and returns it. This PR changes this to go over the whole sample and only return if we find exactly one bounding box. Otherwise we fail. Thus, if the input is batched and thus contains more than one bounding box, we now fail gracefully. 

This is almost on par what we do for image dimensions with `query_chw`. The caveat is that `query_chw` allows multiple images if they have the same dimensions. However, `query_bounding_box` returns an actual bounding box and not its shape, so IMO it is justified to require exactly one.

### DIFF
--- a/torchvision/prototype/transforms/_utils.py
+++ b/torchvision/prototype/transforms/_utils.py
@@ -15,7 +15,7 @@ def query_bounding_box(sample: Any) -> features.BoundingBox:
     if not bounding_boxes:
         raise TypeError("No bounding box was found in the sample")
     elif len(bounding_boxes) > 2:
-        raise TypeError("Found multiple bounding boxes in the sample")
+        raise ValueError("Found multiple bounding boxes in the sample")
     return bounding_boxes.pop()
 
 
@@ -42,7 +42,7 @@ def query_chw(sample: Any) -> Tuple[int, int, int]:
     if not chws:
         raise TypeError("No image was found in the sample")
     elif len(chws) > 2:
-        raise TypeError(f"Found multiple CxHxW dimensions in the sample: {sequence_to_str(sorted(chws))}")
+        raise ValueError(f"Found multiple CxHxW dimensions in the sample: {sequence_to_str(sorted(chws))}")
     return chws.pop()
 
 
--- a/torchvision/prototype/transforms/_utils.py
+++ b/torchvision/prototype/transforms/_utils.py
@@ -11,11 +11,12 @@ from .functional._meta import get_dimensions_image_pil, get_dimensions_image_ten
 
 def query_bounding_box(sample: Any) -> features.BoundingBox:
     flat_sample, _ = tree_flatten(sample)
-    for i in flat_sample:
-        if isinstance(i, features.BoundingBox):
-            return i
-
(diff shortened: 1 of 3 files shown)


torchvision/prototype/transforms/_utils.py
```diff
@@ -11,11 +11,12 @@
...
-    for i in flat_sample:
-        if isinstance(i, features.BoundingBox):
-            return i
-
-    raise TypeError("No bounding box was found in the sample")
+    bounding_boxes = {item for item in flat_sample if isinstance(item, features.BoundingBox)}
+    if not bounding_boxes:
+        raise TypeError("No bounding box was found in the sample")
+    elif len(bounding_boxes) > 2:
+        raise TypeError("Found multiple bounding boxes in the sample")
```
vfdev-5, line 18: Maybe, we can add some hints on how to solve the issue in both error messages ?
Especially for "Found multiple bounding boxes in the sample", where the error means that user passes batched sample instead of a single datapoint

datumbox (commented):
As discussed offline, `FixedSizeCrop` shouldn't fail if `query_bounding_box()` returns no bboxes. The transform is valid with and without bboxes, but if bboxes are found one needs to filter invalid bboxes and labels.

pmeier:
With `FixedSizeCrop` being the first transformation that conditionally needs a bounding box, my preferred way to communicate that from `query_bounding_box` is through different error types. I would stick to a `TypeError` for "nothing was found", but switch to a `RuntimeError` or `ValueError` if we find more than one.

With that we can do something like

```py
try:
    bounding_box = query_bounding_box(sample)
    is_valid = ...
except TypeError:
    is_valid = None

return dict(is_valid=is_valid)
```

The transform will still error out in case we find more than one bounding box, but we don't require a bounding box anymore. Thoughts?

datumbox:
@pmeier I'm not a huge fan of this pythonic idiom but it's very popular and often a recommended approach so yeah, that's going to work fine.

datumbox (dismissed):
LGTM, thanks!

datumbox:
@pmeier You will need to update the tests as some are [failing](https://github.com/pytorch/vision/runs/8033243497?check_suite_focus=true).

PR #6487, opened 2022-08-24
[proto] Fixes Transform._transformed_types and torch.Tensor

vfdev-5:
- Fixes unexpected behaviour with Transform._transformed_types and torch.Tensor
- Added a test for color space conversion


torchvision/prototype/transforms/_transform.py
```diff
@@ -6,9 +6,18 @@
...
+from torchvision.prototype.transforms._utils import is_simple_tensor
 from torchvision.utils import _log_api_usage_once
 
 
+def _isinstance(obj: Any, types: Tuple[Type, ...]) -> bool:
+    has_tensor = torch.Tensor in types
+    if not has_tensor:
+        return isinstance(obj, types)
+    types_ = tuple(t for t in types if t != torch.Tensor)
+    return isinstance(obj, types_) or is_simple_tensor(obj)
```
datumbox, line 18: In other places, we are very careful on when we use `torch.Tensor` and when we use `is_simple_tensor()`. Here handle it differently. Why not use a similar approach as for `has_any` and `has_all` and accept a `*types_or_checks: Union[Type, Callable[[Any], bool]]`?

torchvision/prototype/transforms/_transform.py
```diff
@@ -1,18 +1,19 @@
...
 from torchvision.prototype.features import _Feature
+from torchvision.prototype.transforms._utils import _isinstance, is_simple_tensor
 from torchvision.utils import _log_api_usage_once
 
 
 class Transform(nn.Module):
 
     # Class attribute defining transformed types. Other types are passed-through without any transformation
-    _transformed_types: Tuple[Type, ...] = (torch.Tensor, _Feature, PIL.Image.Image)
+    _transformed_types: Tuple[Union[Type, Callable[[Any], bool]], ...] = (is_simple_tensor, _Feature, PIL.Image.Image)
```
datumbox, line 16: I think that's correct. @pmeier thoughts?

datumbox (approved):
LGTM, thanks! You got some [unused](https://app.circleci.com/pipelines/github/pytorch/vision/19731/workflows/47babeff-3d1e-4279-96a5-7eb518299d7c/jobs/1599053) imports.

torchvision/prototype/transforms/_transform.py
vfdev-5: We need to transform data as pure torch.Tensor or Image or BBox, Mask, Labels (all _Features) and PIL images. We could either keep `torch.Tensor` and all `_Feature` then would pass automatically and also other user tensor subclasses which we are not aware of.

datumbox: We discussed offline, that's correct indeed.

torchvision/prototype/transforms/_deprecated.py
```diff
@@ -18,7 +17,7 @@
 class ToTensor(Transform):
 
     # Updated transformed types for ToTensor
-    _transformed_types = (torch.Tensor, features._Feature, PIL.Image.Image, np.ndarray)
+    _transformed_types = (is_simple_tensor, features._Feature, PIL.Image.Image, np.ndarray)
```
pmeier, line 20: Should we have `features._Feature` here? For `Image`'s it would unwrap IIUC, which is not what want. For all other features, this transform will fail, no?

Plus, having `is_simple_tensor` *and* `features._Feature` in this list is equivalent to `torch.Tensor`.

pmeier, line 20: Should we have `features._Feature` here? For `Image`'s it would unwrap IIUC, which is not what want. For all other features, this transform will fail, no?

Plus, having `is_simple_tensor` *and* `features._Feature` in this list is equivalent to `torch.Tensor`.

torchvision/prototype/transforms/_deprecated.py
```diff
@@ -52,7 +51,7 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
 class ToPILImage(Transform):
 
     # Updated transformed types for ToPILImage
-    _transformed_types = (torch.Tensor, features._Feature, PIL.Image.Image, np.ndarray)
+    _transformed_types = (is_simple_tensor, features._Feature, PIL.Image.Image, np.ndarray)
```
pmeier, line 54: Only `features.Image` should be allowed here instead of `features._Feature`. Everything else will fail.

torchvision/prototype/transforms/_meta.py
```diff
@@ -42,7 +42,7 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
 
 class ConvertColorSpace(Transform):
     # F.convert_color_space does NOT handle `_Feature`'s in general
-    _transformed_types = (torch.Tensor, features.Image, PIL.Image.Image)
```
pmeier, line 45: Maybe we should have an alias for these three, since they will often happen together?

torchvision/prototype/transforms/_utils.py
```diff
@@ -45,12 +45,18 @@ def query_chw(sample: Any) -> Tuple[int, int, int]:
     return chws.pop()
 
 
+def _isinstance(obj: Any, types_or_checks: Tuple[Union[Type, Callable[[Any], bool]], ...]) -> bool:
```
pmeier, line 48: Could you also use this in `has_all`?

torchvision/prototype/transforms/_deprecated.py
vfdev-5: I think @datumbox wanted to fix all incorrect `_transformed_types` usages and remove if/else inside `_transform`. So, here I just replace `torch.Tensor` by `is_simple_tensor`

torchvision/prototype/transforms/_utils.py
vfdev-5: I think we can change easily for-loops types -> objects into objects -> types.

torchvision/prototype/transforms/_deprecated.py
datumbox: @pmeier Victor is doing changes only on the is_simple_tensor to replace existing torch.Tensor. I'll be following up with a broader clean up here.

torchvision/prototype/transforms/_deprecated.py
datumbox: You are right. I will follow up on a PR to update these. The scope of this PR is just to change the `torch.Tensor` with `is_simple_tensor`.

torchvision/prototype/transforms/_meta.py
datumbox: We can do on a follow up. Good idea for when we review the whole input validation mechanism as a whole.

torchvision/prototype/transforms/_deprecated.py
pmeier: Ah, my bad.

torchvision/prototype/transforms/_utils.py
pmeier: True, I didn't see that. Could you add a comment to `has_all` why we are not using it, since that is not obvious?

datumbox:
@vfdev-5 This `gaussian_blur_image_tensor` [test](https://github.com/pytorch/vision/runs/8013743839?check_suite_focus=true) is flaky. It's been failing a lot lately. Could you look into it on a separate PR?

datumbox:
I'm merging this because I really need this to progress other work. Let's follow up with the non-blocking comments.

PR #6486, opened 2022-08-24
[NOMERGE] Transforms V2 API overview

datumbox:
Not to be merged. Adding comments over the overall API.

PR #6490, opened 2022-08-25
[proto] Fix kernel passthrough and types of Normalize

datumbox:
Addresses some of the remarks at #6486

Depends on #6487

This PR:
- Ensures `Normalize` uses the right `_transformed_types`
- Removes the passthrough functionality from the `F.normalize()` kernel
- Updates the error message to the one on main branch

### DIFF
--- a/torchvision/prototype/transforms/_misc.py
+++ b/torchvision/prototype/transforms/_misc.py
@@ -10,6 +10,8 @@ from torchvision.prototype.transforms import functional as F, Transform
 from torchvision.prototype.transforms._utils import query_bounding_box
 from torchvision.transforms.transforms import _setup_size
 
+from ._utils import is_simple_tensor
+
 
 class Identity(Transform):
     def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
@@ -91,10 +93,12 @@ class LinearTransformation(Transform):
 
 
 class Normalize(Transform):
-    def __init__(self, mean: List[float], std: List[float]):
+    _transformed_types = (PIL.Image.Image, features.Image, is_simple_tensor)
+
+    def __init__(self, mean: Sequence[float], std: Sequence[float]):
         super().__init__()
-        self.mean = mean
-        self.std = std
+        self.mean = list(mean)
+        self.std = list(std)
 
     def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
         return F.normalize(inpt, mean=self.mean, std=self.std)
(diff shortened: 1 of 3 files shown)


torchvision/prototype/transforms/_misc.py
```diff
@@ -91,10 +93,12 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
 
 
 class Normalize(Transform):
-    def __init__(self, mean: List[float], std: List[float]):
+    _transformed_types = (PIL.Image.Image, features.Image, is_simple_tensor)
```
datumbox, line 96: This won't work currently, it requires changes proposed at #6487

torchvision/prototype/transforms/functional/_misc.py
```diff
@@ -15,10 +15,8 @@
 
 
 def normalize(inpt: DType, mean: List[float], std: List[float], inplace: bool = False) -> DType:
```
pmeier, line 17: Can we also have `Sequence[float]` here and convert to list again before calling the kernel? Since we already do this conversion in the transform, there will be no extra conversion when repeatedly calling this from there. On the other hand, users that only want the functional get the benefit of not only using lists.

datumbox: That's a bigger change to be discussed because you are changing the signature of a public method. `F.normalize()` exists already on the old API. Perhaps we can do this because we already change the type of inpt. But I wouldn't do this change right now.

datumbox: Let me add a bit more info here to have a bit more context.

If we had a way to detect that a certain Tensor object is a "simple tensor" while JIT-scripting, then we would be able refactor the entire `F` API to keep it JIT-scriptable and maintain BC. This could require some nasty tricks like:
- Faking types
- Splitting the non-jit scriptable part of `F` high-level kernels in separate methods that are marked with @torch.jit.unused. 
- Extensive use of `torch.jit.is_scripting()` and `torch.jit.is_tracing()`

Another thing we could do if we want to avoid BC breakages is to provide a jit-scriptable `F` that has exactly the same signatures as the high-level kernels but only supports pure tensors.

To keep the above a possibility, we shouldn't modify prematurely the existing signatures to receive parameters that are not JIT-scriptable (like `Sequence`). We might end up doing it at the end, but let's hold off to that until we know for sure.

vfdev-5 (approved):
lgtm, thanks @datumbox !

datumbox:
The [failing](https://github.com/pytorch/vision/runs/8020015949?check_suite_focus=true) test is unrelated.

PR #6493, opened 2022-08-25
Try to fix mypy issue with calculate_md5

vfdev-5:
- https://app.circleci.com/pipelines/github/pytorch/vision/19736/workflows/98540a35-693d-4f68-92e4-cbdcc2601d81/jobs/1599504

and passing locally on python 3.9 without errors like:
```
# mypy --config-file mypy.ini
torchvision/datasets/utils.py:68: error: Argument 1 to "md5" has incompatible type "**Dict[str, bool]"; expected
"Union[bytes, Union[bytearray, memoryview, array[Any], mmap, _CData, PickleBuffer]]"  [arg-type]
        md5 = hashlib.md5(**dict(usedforsecurity=False) if sys.version_info >= (3, 9) else dict())
```

### DIFF
--- a/torchvision/datasets/utils.py
+++ b/torchvision/datasets/utils.py
@@ -65,7 +65,10 @@ def calculate_md5(fpath: str, chunk_size: int = 1024 * 1024) -> str:
     # Setting the `usedforsecurity` flag does not change anything about the functionality, but indicates that we are
     # not using the MD5 checksum for cryptography. This enables its usage in restricted environments like FIPS. Without
     # it torchvision.datasets is unusable in these environments since we perform a MD5 check everywhere.
-    md5 = hashlib.md5(**dict(usedforsecurity=False) if sys.version_info >= (3, 9) else dict())
+    if sys.version_info >= (3, 9):
+        md5 = hashlib.md5(usedforsecurity=False)
+    else:
+        md5 = hashlib.md5()
     with open(fpath, "rb") as f:
         for chunk in iter(lambda: f.read(chunk_size), b""):
             md5.update(chunk)


NicolasHug:
IIRC this only fails on 3.10. @pmeier had an attempt to fix this at https://github.com/pytorch/vision/pull/5755 but we weren't too enthusiastic about the solution

vfdev-5:
@NicolasHug yes, @pmeier pointed out to me his PR.

Let me recap the findings from my local tests using conda env on macosx :

code | python version | mypy check
---|---|---
main branch | 3.8 | failure with `md5 = hashlib.md5(**dict(usedforsecurity=False) if sys.version_info >= (3, 9) else dict())`
main branch | 3.9 | failure with `md5 = hashlib.md5(**dict(usedforsecurity=False) if sys.version_info >= (3, 9) else dict())`
main branch | 3.10 | failure with `md5 = hashlib.md5(**dict(usedforsecurity=False) if sys.version_info >= (3, 9) else dict())`
This PR | 3.8 | passing
This PR | 3.9 | passing
This PR | 3.10 | passing

mypy version: mypy 0.971 (compiled: yes)

datumbox (approved):
LGTM! It simplifies the code as well while not specifying Python to 3.10 as the previous attempt.

pmeier:
It seems that `mypy` indeed recognizes the `sys.version_info` branch: https://mypy-play.net/?mypy=latest&python=3.10&gist=3f2a763676db35f08960e3bad08b62f3

> while not specifying Python to 3.10 as the previous attempt.

My PR is not primarily about fixing this issue, but rather making these issues visible in CI. The change made in this PR cannot be verified by CI, but has to be done manually. Meaning, if someone introduces a change like I did in #5717, we will not know until someone on Python >= 3.7 is getting errors on main.

NicolasHug (approved):
`mypy` will always surprise me at how it behaves so differently on 2 pieces of code which look like the exact same thing externally...

Thanks @vfdev-5, I agree we can merge as this issue has been an inconvenience for local development

PR #6497, opened 2022-08-25
Move output to CPU before assert_expected

datumbox:


vfdev-5 (dismissed):
Thanks ! 
Lgtm as prototype tests are now passing on GPU

pmeier:
In case we don't care about the device and just want to compare the values, we can set `check_device=False` here:

https://github.com/pytorch/vision/blob/becaba0e0546f434be64de4e9603ab66e4b160b2/test/test_models.py#L116

datumbox:
I'm added both cpu and check_device=False. Ideally the expected values for new models should be stored in CPU (to avoid issues for people who don't have specialized hardware) but the idea of avoid checking the device is good, so I'm included this.

test/test_models.py
```diff
@@ -893,7 +895,7 @@ def test_quantized_classification_model(model_fn):
     out = model(x)
 
     if model_name not in quantized_flaky_models:
-        _assert_expected(out, model_name + "_quantized", prec=2e-2)
```
pmeier, line 896: Do we still need the `.cpu()` call here, if this is done anyway in the comparison logic by `check_device=False`? This applies to all occurrences of this.

pmeier (approved):
One question, otherwise LGTM if CI is green!

test/test_models.py
datumbox: If we don't add it, it will take something estimated in cuda and produce an expected file. This might cause issues to users who load it without cuda and might require to map it to CPU. That's why I put it everywhere it should be. Might be an overkill.

PR #6498, opened 2022-08-25
[proto] Consistent supported/unsupported types handling in LinearTransformation, other perf comments

vfdev-5:
- Consistent supported/unsupported types handling in LinearTransformation, 
- Fixed BC on arg in RandomChoice
- Added todo perf comments


torchvision/prototype/transforms/_container.py
```diff
@@ -33,9 +34,21 @@ def extra_repr(self) -> str:
 
 
 class RandomChoice(Transform):
-    def __init__(self, transforms: Sequence[Callable], probabilities: Optional[List[float]] = None) -> None:
+    def __init__(
+        self,
+        transforms: Sequence[Callable],
+        probabilities: Optional[List[float]] = None,
+        p: Optional[List[float]] = None,
```
datumbox, line 41: Question: Why not send extra things on `kwargs`? This way we expose/advertise on the documentation. 

Also, why do we change to `probabilities`? `p` is quite common in PyTorch.

torchvision/prototype/transforms/_container.py
```diff
@@ -33,9 +34,21 @@ def extra_repr(self) -> str:
...
+    ) -> None:
         if not isinstance(transforms, Sequence):
             raise TypeError("Argument transforms should be a sequence of callables")
+        if p is not None:
+            warnings.warn(
+                "Argument p is deprecated and will be removed in the next release. "
```
datumbox, line 47:
```suggestion
                "Argument p is deprecated and will be removed in a future release. "
```

torchvision/prototype/transforms/_misc.py
```diff
@@ -62,13 +62,14 @@ def __init__(self, transformation_matrix: torch.Tensor, mean_vector: torch.Tenso
         self.transformation_matrix = transformation_matrix
         self.mean_vector = mean_vector
 
-    def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
+    def forward(self, *inputs: Any) -> Any:
+        sample = inputs if len(inputs) > 1 else inputs[0]
+        if has_any(sample, PIL.Image.Image):
+            raise TypeError("LinearTransformation does not work on PIL Images")
```
datumbox, line 68: This is not going to work. If you want PIL images to be supported you would also need to add them on `_transformed_types`

torchvision/prototype/transforms/_misc.py
```diff
@@ -62,13 +62,14 @@ def __init__(self, transformation_matrix: torch.Tensor, mean_vector: torch.Tenso
...
 
-        if isinstance(inpt, features._Feature) and not isinstance(inpt, features.Image):
-            return inpt
-        elif isinstance(inpt, PIL.Image.Image):
-            raise TypeError("Unsupported input type")
+        return super().forward(sample)
 
+    def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
```
datumbox, line 72: Nit:
```suggestion
    def _transform(self, inpt: Any, params: Dict[str, Any]) -> torch.Tensor:
```

torchvision/prototype/transforms/_misc.py
pmeier: Not sure what you mean. This specifically says that they are not supported. If we remove this error, they will silently pass through.

torchvision/prototype/transforms/_misc.py
pmeier: I don't think `mypy` will complain here since we are making the type more concrete, but it won't have any effect either. This is an internal method and the user code will only "see" 

https://github.com/pytorch/vision/blob/becaba0e0546f434be64de4e9603ab66e4b160b2/torchvision/prototype/transforms/_transform.py#L28

torchvision/prototype/transforms/_container.py
vfdev-5: @pmeier I think it was you who introduced `probabilities` to `RandomChoice`.

IMO, `probabilities` emphasizes that it is a list instead of single value like `p`.

> Why not send extra things on kwargs?

We can, but this will look untypical in the code. 

> This way we expose/advertise on the documentation.

I do not understand what you mean by that.

torchvision/prototype/transforms/_misc.py
datumbox: @pmeier I agree. I think it's worth it for someone who reads the code to see what we are currently returning.

torchvision/prototype/transforms/_misc.py
datumbox: Right now it passes through. He needs to add `_transformed_types = (is_simple_tensor, features.Image, PIL.Image.Image)`

torchvision/prototype/transforms/_container.py
datumbox: > I do not understand what you mean by that.

I mean that the addition of the `p` will get it exposed and advertised in the documentation as a field.

I'm OK with what you do, feel free to keep as you have.

torchvision/prototype/transforms/_misc.py
datumbox: @pmeier and @vfdev-5 pointed out to me offline that this is unnecessary because the error is triggered before the use of `_transformed_types`

PR #6500, opened 2022-08-26
More cleanup for prototype transforms

pmeier:
Addresses https://github.com/pytorch/vision/pull/6486#discussion_r955226040.

- Since we already have a `ConvertImageDtype` transform, I kept the name and exposed `convert_image_dtype` as is. This also aligns well with other conversion ops like `convert_bounding_box_format`.
- I reduced the number of imports of `torchvision.transforms` inside the `torchvision.prototype.transforms` module. There are 3 places left that I don't think we can or should get rid of:
    - `torchvision.prototype.transforms.functional.*`: We have quite a few kernels that are just imported from the old API.
    - `torchvision.prototype.transforms._deprecated`: We opted to call the legacy kernels instead of reproducing the behavior with the new ones.
    - `torchvision.prototype.transforms._utils`: We use a number of utilities of the old transforms in the new ones. I moved them to this module to have them in one central place to avoid having each other module import them directly.
- I left a couple of `FIXME` comments for minor stuff that I found during cleaning up. Maybe we can address them as part of this PR

### Todo

- [x] get_dimensions
- [x] get_image_num_channels (new tmp name `get_channels_dim()` + alias for the old)
- [x] get_image_size (new tmp name `get_spatial_dims()` + alias for the of)


torchvision/prototype/transforms/_type_conversion.py
```diff
@@ -14,6 +14,7 @@ class DecodeImage(Transform):
     _transformed_types = (features.EncodedImage,)
 
     def _transform(self, inpt: Any, params: Dict[str, Any]) -> features.Image:
+        # FIXME: should we move this wrapping into the functional?
```
vfdev-5, line 17: Let's fix these fixme here ?

pmeier: I would prefer that as stated above. Let's see if we all agree with the proposals. I'll send a new commit after.

pmeier: This was done by @datumbox in #6502.

datumbox: I didn't modify `decode_image_with_pil()`. Should I?

torchvision/prototype/transforms/_color.py
```diff
@@ -117,14 +116,14 @@ def _permute_channels(self, inpt: Any, *, permutation: torch.Tensor) -> Any:
 
         image = inpt
         if isinstance(inpt, PIL.Image.Image):
-            image = _F.pil_to_tensor(image)
+            image = F.to_image_tensor(image)
```
datumbox, line 119: Github won't let me put the comment on the right place but 5 lines above we are doing a `not (isinstance(inpt, (features.Image, PIL.Image.Image)) or is_simple_tensor(inpt))` check to pass through. Instead we should declare in this class a:
`_transformed_types = (features.Image, PIL.Image.Image, is_simple_tensor)`

And avoid this.

torchvision/prototype/transforms/_deprecated.py
```diff
@@ -18,6 +18,7 @@ class ToTensor(Transform):
     _transformed_types = (PIL.Image.Image, np.ndarray)
 
     def __init__(self) -> None:
+        # FIXME: should the replacement be a `Compose` with `ConvertImageDtype`?
```
datumbox, line 21: Let's remove this fixme. Previously we have intentionally pushed people away from ToTensor which does 2 things at the same time (rescaling to 0-1 and tensor conversion). We asked people to explicitly use 2 transforms instead, so by design we didn't want to offer a single transform that does both. The reason is, we don't have to scale the image to 0-1 prematurely.

The new API allows people to keep the images on the original 0-255 scale even more, so adding this convenience method doesn't align with this. Let's remove the FIXME.

torchvision/prototype/transforms/_type_conversion.py
datumbox: I've done it on #6502.
```suggestion
```

torchvision/prototype/transforms/_type_conversion.py
```diff
@@ -43,6 +44,7 @@ class ToImageTensor(Transform):
     _transformed_types = (is_simple_tensor, PIL.Image.Image, np.ndarray)
 
     def _transform(self, inpt: Any, params: Dict[str, Any]) -> features.Image:
+        # FIXME: should we move this wrapping into the functional?
```
datumbox, line 47: Done it.
```suggestion
```

torchvision/prototype/transforms/functional/__init__.py
```diff
@@ -68,6 +68,7 @@
     five_crop,
     five_crop_image_pil,
     five_crop_image_tensor,
+    hflip,
```
datumbox, line 71: Question: Do we want to define all aliases in one single place? Perhaps all aliases should leave here instead on the submodules. No strong opinion. Thoughts @pmeier @vfdev-5 ?

torchvision/prototype/transforms/functional/_deprecated.py
```diff
@@ -41,3 +42,12 @@ def rgb_to_grayscale(inpt: Any, num_output_channels: int = 1) -> Any:
     )
 
     return _F.rgb_to_grayscale(inpt, num_output_channels=num_output_channels)
+
+
+def to_tensor(inpt: Any) -> torch.Tensor:
+    # FIXME: should we keep the "if needed" phrase or unconditionally recommend `convert_image_dtype`?
```
datumbox, line 48: To get the `to_tensor()` behaviour you need both `to_image_tensor()` and `convert_image_dtype()` calls. We should change the language below to indicate this.

datumbox (commented):
Some thoughts below:

torchvision/prototype/transforms/_color.py
datumbox: On a second thought, it might be worth applying this proposal to all Colour kernels to avoid the dispatching to the no-op. This is going to make things faster. Since this is a bigger change, I'll send a separate PR and discuss there if we want to do it.

torchvision/prototype/transforms/functional/__init__.py
vfdev-5: I think they can live in appropriate submodules

datumbox: OK keeping as is then

torchvision/prototype/transforms/_deprecated.py
datumbox: After an offline discussion with @pmeier, he clarified that the addition of the Copose would be just on the example inside the warning. I thought the proposal was for a new shortcut in our code. We should do what Philip suggested here to ensure we inform the users about the best practice.

torchvision/prototype/transforms/functional/_utils.py
```diff
@@ -0,0 +1,43 @@
...
+import PIL.Image
+import torch
+from torchvision.prototype import features
+from torchvision.transforms import functional_pil as _FP, functional_tensor as _FT
+
+get_dimensions_image_tensor = _FT.get_dimensions
+get_dimensions_image_pil = _FP.get_dimensions
+
+
+def get_chw(image: Union[PIL.Image.Image, torch.Tensor, features.Image]) -> Tuple[int, int, int]:
```
datumbox, line 12: Nit: I expected all these to be in `_meta` because they deal with meta-data stored in the Tensor.

datumbox (approved):
A few questions but LGTM.

torchvision/prototype/transforms/_type_conversion.py
```diff
@@ -52,3 +52,15 @@ def __init__(self, *, mode: Optional[str] = None) -> None:
...
+
+# We changed the names to align them with the new naming scheme. Still, `PILToTensor` and `ToPILImage` are
+# prevalent and well understood. Thus, we just alias them without deprecating the old names.
+
+
+class PILToTensor(ToImageTensor):
+    pass
+
+
+class ToPILImage(ToImagePIL):
```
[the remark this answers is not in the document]
pmeier: Honestly, I didn't think of that. Happy to directly alias.  

> Is it to ensure isinstance and other methods work?

If we have equality, `isinstance` will continue to work.

torchvision/prototype/transforms/functional/_utils.py
pmeier: Can move.

datumbox:
You need to patch the tests as well. See [this](https://github.com/pytorch/vision/runs/8037374696?check_suite_focus=true).

PR #6501, opened 2022-08-26
update roipool to make it torch fx traceable

inisis:
update roipool to make it torch fx traceable

### DIFF
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -135,7 +135,7 @@ class RoIOpTester(ABC):
         output_fx = graph_module(x, rois)
         assert output_fx.dtype == x.dtype
         tol = 1e-5
-        torch.testing.assert_close(output_gt.to(output_fx), output_fx, rtol=tol, atol=tol)
+        torch.testing.assert_close(output_gt, output_fx, rtol=tol, atol=tol)
 
     @pytest.mark.parametrize("seed", range(10))
     @pytest.mark.parametrize("device", cpu_and_gpu())
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -118,6 +118,25 @@ class RoIOpTester(ABC):
         assert len(graph_node_names[0]) == len(graph_node_names[1])
         assert len(graph_node_names[0]) == 1 + op_obj.n_inputs
 
+    @pytest.mark.parametrize("device", cpu_and_gpu())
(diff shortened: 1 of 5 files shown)


datumbox:
@inisis Thanks for the contribution. Can you do the same for `ps_roi_align`, `ps_roi_pool` and `roi_align`? Also we should add a simple test at [test_ops.py](https://github.com/pytorch/vision/blob/7cc2c95ad90f18f4e15005b49a3636d6ea3135b4/test/test_ops.py#L81) to confirm it's FX traceable.

inisis:
@datumbox Thanks for your reply, I modified the other three funtions, and now all of them can be traced. However, I'm not familiar with your test script, can you show me an example. I teseted on my working env, and following two images are the results for PSRoIAlign module with/without modification, the other three are the same.

# with modification
image


# without modification
image

inisis:
@datumbox op test is added

datumbox:
@inisis Thanks a lot! sorry for the delay. I'm currently on my annual leave but I'll review as soon as I come back.

If anyone else from the maintaining team has bandwidth to review, this is also very welcome.

test/test_ops.py
```diff
@@ -118,6 +118,25 @@ def test_is_leaf_node(self, device):
...
+            [[0, 0, 0, 9, 9], [0, 0, 5, 4, 9], [0, 5, 5, 9, 9], [1, 0, 0, 9, 9]],  # format is (xyxy)
+            dtype=rois_dtype,
+            device=device,
+        )
+        output_gt = op_obj(x, rois)
+        assert output_gt.dtype == x.dtype
+        output_fx = graph_module(x, rois)
+        assert output_fx.dtype == x.dtype
+        tol = 1e-5
+        torch.testing.assert_close(output_gt.to(output_fx), output_fx, rtol=tol, atol=tol)
```
datumbox, line 138: I'm not sure why we do `.to(output_fx)` here:
```suggestion
        torch.testing.assert_close(output_gt, output_fx, rtol=tol, atol=tol)
```

datumbox (commented):
LGTM, thanks for the @inisis. Only one comment on the test:

datumbox:
The [failing](https://app.circleci.com/pipelines/github/pytorch/vision/19931/workflows/880e2e16-1993-4407-850d-4dbf3e013512/jobs/1617853) jobs are unrelated to this PR.

datumbox (approved):
LGTM, thanks!

PR #6502, opened 2022-08-26
More prototype Transforms cleanups

datumbox:
This PR:
- Makes Normalize verify input on forward instead on the kernel
- The kernel `to_image_tensor` returns `features.Image` instead of plain tensor.
- The kernel `decode_image_with_pil` returns `features.Image` instead of plain tensor.

### DIFF
--- a/torchvision/prototype/features/_encoded.py
+++ b/torchvision/prototype/features/_encoded.py
@@ -9,7 +9,6 @@ import torch
 from torchvision.prototype.utils._internal import fromfile, ReadOnlyTensorBuffer
 
 from ._feature import _Feature
-from ._image import Image
 
 D = TypeVar("D", bound="EncodedData")
 
--- a/torchvision/prototype/features/_encoded.py
+++ b/torchvision/prototype/features/_encoded.py
@@ -46,11 +46,6 @@ class EncodedImage(EncodedData):
 
         return self._image_size
 
-    def decode(self) -> Image:
-        # TODO: this is useful for developing and debugging but we should remove or at least revisit this before we
-        #  promote this out of the prototype state
-        return self._F.decode_image_with_pil(self)
-
 
 class EncodedVideo(EncodedData):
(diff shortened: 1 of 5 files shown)


vfdev-5 (approved):
LGTM, thanks

pmeier (approved):
LGTM, thanks!

vfdev-5 (approved):
Still good to me

torchvision/prototype/features/_encoded.py
```diff
@@ -46,11 +46,6 @@ def image_size(self) -> Tuple[int, int]:
 
         return self._image_size
 
-    def decode(self) -> Image:
```
datumbox, line 49: Mypy was complaining with:
```
torchvision/prototype/features/_encoded.py:52: error: Returning Any from function declared to return "Image"  [no-any-return]
            return self._F.decode_image_with_pil(self)
```

It's unclear why it couldn't see that the return type is indeed `Image` but given it's not used anywhere in the codebase and it's for debugging only, I'm removing it. Feel free to add back on the future,

torchvision/prototype/transforms/_misc.py
```diff
@@ -104,6 +104,11 @@ def __init__(self, mean: Sequence[float], std: Sequence[float]):
     def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
         return F.normalize(inpt, mean=self.mean, std=self.std)
 
+    def forward(self, *inpts: Any) -> Any:
+        if has_any(inpts, PIL.Image.Image):
+            raise TypeError(f"{type(self).__name__}() does not support PIL images.")
+        return super().forward(*inpts)
```
datumbox, line 110: Adopting @vfdev-5's proposal to eliminate the `sample = inputs if len(inputs) > 1 else inputs[0]` idiom in favour of `super().forward(*inpts)`

PR #6503, opened 2022-08-26
introduce _check method for type checks on prototype transforms

pmeier:
Most of our prototype transforms that override `forward` do this only to perform some input checks before passing them on to `_get_params` and `_transform`. This PR adds a `_check` method that makes this regular behavior, whereas overriding `forward` means some "special" behavior the API can't handle.

with `_check` in place, there are only three* transforms that still need to override `forward`:

- https://github.com/pytorch/vision/blob/7cc2c95ad90f18f4e15005b49a3636d6ea3135b4/torchvision/prototype/transforms/_augment.py#L188
- https://github.com/pytorch/vision/blob/7cc2c95ad90f18f4e15005b49a3636d6ea3135b4/torchvision/prototype/transforms/_auto_augment.py#L139
- https://github.com/pytorch/vision/blob/7cc2c95ad90f18f4e15005b49a3636d6ea3135b4/torchvision/prototype/transforms/_container.py#L10

These are objectively outlier and thus it is justified for them to override `forward`.

Using `_check` will also get rid of the boilerplate

```py
def forward(self, *inputs):

    ...

```

idiom.


torchvision/prototype/transforms/_geometry.py
```diff
@@ -202,15 +201,12 @@ def __init__(self, size: Union[int, Sequence[int]], vertical_flip: bool = False)
...
-        return F.ten_crop(inpt, self.size, vertical_flip=self.vertical_flip)
-
-    def forward(self, *inputs: Any) -> Any:
-        if has_any(inputs, features.BoundingBox, features.Mask):
+    def _check(self, sample: Any) -> None:
+        if has_any(sample, features.BoundingBox, features.Mask):
             raise TypeError(f"BoundingBox'es and Mask's are not supported by {type(self).__name__}()")
-        return super().forward(*inputs)
+
+    def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
```
datumbox, line 208: Is this change intentional or the result of copy-paste?
```suggestion
    def _transform(
        self, inpt: Union[features.ImageType, features.VideoType], params: Dict[str, Any]
    ) -> Union[List[features.ImageTypeJIT], List[features.VideoTypeJIT]]:
```

torchvision/prototype/transforms/_transform.py
```diff
@@ -32,6 +35,8 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
     def forward(self, *inputs: Any) -> Any:
         sample = inputs if len(inputs) > 1 else inputs[0]
 
+        self._check(sample)
```
datumbox, line 38: Why aren't we flattening before this call? Also in this approach we can avoid the multiple flattening calls within `has_any`, `has_all` and the rest of the utility methods. Do you plan to do this on a follow up?

torchvision/prototype/transforms/_geometry.py
pmeier: The latter. Happened during fixing merge conflicts. Thanks for pointing it out!

torchvision/prototype/transforms/_augment.py
```diff
@@ -107,17 +107,16 @@ def __init__(self, alpha: float, p: float = 0.5) -> None:
         self.alpha = alpha
         self._dist = torch.distributions.Beta(torch.tensor([alpha]), torch.tensor([alpha]))
 
-    def forward(self, *inputs: Any) -> Any:
+    def _check(self, sample: Any) -> None:
```
vfdev-5, line 110: Let's call `_check_inputs` instead of `_check`

vfdev-5 (approved):
lgtm, thanks!

PR #6504, opened 2022-08-26
[proto] Simplified code in overridden transform forward methods

vfdev-5:

### DIFF
--- a/torchvision/prototype/transforms/_augment.py
+++ b/torchvision/prototype/transforms/_augment.py
@@ -101,15 +101,14 @@ class _BaseMixupCutmix(_RandomApplyTransform):
         self.alpha = alpha
         self._dist = torch.distributions.Beta(torch.tensor([alpha]), torch.tensor([alpha]))
 
-    def forward(self, *inpts: Any) -> Any:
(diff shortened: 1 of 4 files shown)


datumbox (approved):
LGTM, let's wait for the tests to confirm.

Should we apply the changes to `_RandomApplyTransform` and `_auto_augment.py` file? cc @pmeier

pmeier:
> Should we apply the changes to `_RandomApplyTransform` and `_auto_augment.py` file?

Yes, we should be consistent where we can. AA might be not possible though, because we don't do the regular processing and need to put the augmented image back into the sample.

datumbox:
@vfdev-5 The [failing](https://github.com/pytorch/vision/runs/8034716688?check_suite_focus=true) test is irrelevant; the flaky gaussian blur again. So not a blocker.

@pmeier Sounds good. Could you review this one to see if all the places changed are valid? I think Victor said that he changed everything that is currently safe to do.

datumbox:
The [failure](https://github.com/pytorch/vision/runs/8036191934?check_suite_focus=true) is unrelated, merging.

PR #6509, opened 2022-08-26
[proto] Added consistency check for classification training preset

vfdev-5:


test/test_prototype_transforms.py
```diff
@@ -1643,3 +1643,118 @@ def test__transform(self):
         assert isinstance(ohe_labels, features.OneHotLabel)
         assert ohe_labels.shape == (4, 3)
         assert ohe_labels.categories == labels.categories == categories
+
+
+class TestConsistencyClassificationPreset:
```
datumbox, line 1648: It's OK if you wanna do this for your own development or temporarily but just noted that the code on References will be massively modified and is not something we need to check compatibility against. You are right wanting to ensure that invocations should remain the same. I just don't think we should care about the Training Classification preset.

PR #6510, opened 2022-08-26
[proto] Fix bug with Compose and PR 6504

vfdev-5:
This isssue introduced in #6504 is discovered in https://github.com/pytorch/vision/pull/6509

- Added a test


datumbox (approved):
LGTM thanks @vfdev-5

PR #6512, opened 2022-08-26
Prototype Transform cleanups and bugfixing

datumbox:
This PR does a series of bug fixes and clean ups on the prototype Transforms API. Each commit is more-or-less atomic.

Feel free to modify or push corrections to this PR.

### DIFF
--- a/torchvision/prototype/features/_bounding_box.py
+++ b/torchvision/prototype/features/_bounding_box.py
@@ -122,7 +122,7 @@ class BoundingBox(_Feature):
         if not isinstance(padding, int):
             padding = list(padding)
 
-        output = self._F.pad_bounding_box(self, padding, format=self.format)
+        output = self._F.pad_bounding_box(self, padding, format=self.format, padding_mode=padding_mode)
 
         # Update output image size:
         left, top, right, bottom = self._F._geometry._parse_pad_padding(padding)
--- a/torchvision/prototype/features/_bounding_box.py
+++ b/torchvision/prototype/features/_bounding_box.py
@@ -4,9 +4,7 @@ from typing import Any, List, Optional, Sequence, Tuple, Union
 
 import torch
 from torchvision._utils import StrEnum
-from torchvision.transforms import InterpolationMode
-from torchvision.transforms.functional import _get_inverse_affine_matrix
-from torchvision.transforms.functional_tensor import _compute_output_size
+from torchvision.transforms import InterpolationMode  # TODO: this needs to be moved out of transforms
 
 from ._feature import _Feature
 
@@ -120,9 +118,6 @@ class BoundingBox(_Feature):
         fill: Optional[Union[int, float, Sequence[int], Sequence[float]]] = None,
         padding_mode: str = "constant",
     ) -> BoundingBox:
-        if padding_mode not in ["constant"]:
-            raise ValueError(f"Padding mode '{padding_mode}' is not supported with bounding boxes")
-
         # This cast does Sequence[int] -> List[int] and is required to make mypy happy
         if not isinstance(padding, int):
             padding = list(padding)
@@ -130,10 +125,7 @@ class BoundingBox(_Feature):
         output = self._F.pad_bounding_box(self, padding, format=self.format)
 
         # Update output image size:
-        # TODO: remove the import below and make _parse_pad_padding available
-        from torchvision.transforms.functional_tensor import _parse_pad_padding
-
-        left, top, right, bottom = _parse_pad_padding(padding)
+        left, top, right, bottom = self._F._geometry._parse_pad_padding(padding)
         height, width = self.image_size
         height += top + bottom
         width += left + right
@@ -158,8 +150,10 @@ class BoundingBox(_Feature):
             # - points dot matrix (_compute_output_size)
             # Alternatively, we could return new image size by self._F.rotate_bounding_box
             height, width = image_size
-            rotation_matrix = _get_inverse_affine_matrix([0.0, 0.0], angle, [0.0, 0.0], 1.0, [0.0, 0.0])
-            new_width, new_height = _compute_output_size(rotation_matrix, width, height)
+            rotation_matrix = self._F._geometry._get_inverse_affine_matrix(
+                [0.0, 0.0], angle, [0.0, 0.0], 1.0, [0.0, 0.0]
+            )
+            new_width, new_height = self._F._geometry._compute_output_size(rotation_matrix, width, height)
             image_size = (new_height, new_width)
 
         return BoundingBox.new_like(self, output, dtype=output.dtype, image_size=image_size)
--- a/torchvision/prototype/features/_bounding_box.py
+++ b/torchvision/prototype/features/_bounding_box.py
@@ -147,13 +147,13 @@ class BoundingBox(_Feature):
         if expand:
             # The way we recompute image_size is not optimal due to redundant computations of
             # - rotation matrix (_get_inverse_affine_matrix)
-            # - points dot matrix (_compute_output_size)
+            # - points dot matrix (_compute_affine_output_size)
             # Alternatively, we could return new image size by self._F.rotate_bounding_box
             height, width = image_size
             rotation_matrix = self._F._geometry._get_inverse_affine_matrix(
                 [0.0, 0.0], angle, [0.0, 0.0], 1.0, [0.0, 0.0]
             )
-            new_width, new_height = self._F._geometry._FT._compute_output_size(rotation_matrix, width, height)
+            new_width, new_height = self._F._geometry._FT._compute_affine_output_size(rotation_matrix, width, height)
             image_size = (new_height, new_width)
 
         return BoundingBox.new_like(self, output, dtype=output.dtype, image_size=image_size)
--- a/torchvision/prototype/features/_bounding_box.py
+++ b/torchvision/prototype/features/_bounding_box.py
@@ -153,7 +153,7 @@ class BoundingBox(_Feature):
             rotation_matrix = self._F._geometry._get_inverse_affine_matrix(
                 [0.0, 0.0], angle, [0.0, 0.0], 1.0, [0.0, 0.0]
             )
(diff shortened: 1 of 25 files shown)


torchvision/prototype/transforms/_meta.py
```diff
@@ -28,9 +28,11 @@ def __init__(self, dtype: torch.dtype = torch.float32) -> None:
         super().__init__()
         self.dtype = dtype
 
-    def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
+    def _transform(
+        self, inpt: Union[torch.Tensor, features.Image], params: Dict[str, Any]
+    ) -> Union[torch.Tensor, features.Image]:
         output = F.convert_image_dtype(inpt, dtype=self.dtype)
-        return output if features.is_simple_tensor(inpt) else features.Image.new_like(inpt, output, dtype=self.dtype)
+        return output if features.is_simple_tensor(inpt) else features.Image.new_like(inpt, output, dtype=self.dtype)  # type: ignore[arg-type]
```
datumbox, line 35: Mypy really couldn't understand that this can be a union.

datumbox (commented):
Some additional context:

torchvision/prototype/features/_bounding_box.py
```diff
@@ -158,8 +150,10 @@ def rotate(
             # - points dot matrix (_compute_output_size)
             # Alternatively, we could return new image size by self._F.rotate_bounding_box
             height, width = image_size
-            rotation_matrix = _get_inverse_affine_matrix([0.0, 0.0], angle, [0.0, 0.0], 1.0, [0.0, 0.0])
-            new_width, new_height = _compute_output_size(rotation_matrix, width, height)
+            rotation_matrix = self._F._geometry._get_inverse_affine_matrix(
+                [0.0, 0.0], angle, [0.0, 0.0], 1.0, [0.0, 0.0]
+            )
+            new_width, new_height = self._F._geometry._FT._compute_output_size(rotation_matrix, width, height)
```
datumbox, line 156: This sucks but apparently `self._F._geometry._compute_output_size` won't do it. We have to use the `self._F._geometry._FT._compute_output_size` version. Not sure if that's expected and if the two methods can be aligned.

torchvision/prototype/features/_bounding_box.py
```diff
@@ -120,20 +118,14 @@ def pad(
         fill: Optional[Union[int, float, Sequence[int], Sequence[float]]] = None,
         padding_mode: str = "constant",
     ) -> BoundingBox:
-        if padding_mode not in ["constant"]:
-            raise ValueError(f"Padding mode '{padding_mode}' is not supported with bounding boxes")
-
```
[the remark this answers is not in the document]
vfdev-5: Padding an image with other non-const modes will create new objects, but we won't be able to create corresponding bboxes, that's why we raise an error. Ignoring this case may lead to silent issues, I think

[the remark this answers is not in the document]
datumbox: You are right to say that if you use reflect, you won't be able to create bboxes for the new edges (maybe we should find a way to do it on the future). This might cause issues or not, we can't tell for sure as in Object Detection datasets there are often many unlabelled objects in the scene already.

The reason I believe this needs to be fixed is because the methods on the _Feature objects are just thin wrappers and should add no new functionality. There are there to redirect to the `F` kernel, rewrap the result and update the meta-data. Adding new functionality means diverging from the behaviour of `F` equivalent kernels. This is why we agreed that the method should receive maximally all parameters and use those that make sense without throwing errors. You can see this in action on how SegmentationMask handles wrong interpretation values. Here we should do the same here.

If we want to detect this and throw an error we need to put the check on the Pad transform. There we can ensure that specific modes aren't used when bboxes are present.

torchvision/prototype/transforms/_deprecated.py
```diff
@@ -14,6 +14,9 @@
 from ._utils import query_chw
 
 
+DType = Union[torch.Tensor, PIL.Image.Image, features._Feature]
```
pmeier, line 17: Why was the name `DType`chosen? I think it is quite misleading. If we are going to spread this to multiple files, we should rethink the name.

torchvision/prototype/transforms/_geometry.py
```diff
@@ -15,6 +15,9 @@
 from ._utils import _check_sequence_input, _setup_angle, _setup_size, has_all, has_any, query_bounding_box, query_chw
 
 
+DType = Union[torch.Tensor, PIL.Image.Image, features._Feature]
```
pmeier, line 18: Seeing that this is defined in every module, can we maybe put this in `_utils`?

torchvision/prototype/transforms/_meta.py
pmeier: Not sure what the error is, but `Union[torch.Tensor, features.Image]` is redundant. Basically this says "the output can be a `torch.Tensor` **or** a `features.Image`". But since a `features.Image` is a `torch.Tensor`, `Union[torch.Tensor, features.Image]` should be equivalent to simply annotating with `torch.Tensor`.

pmeier (commented):
I think there are a few redundant annotations. I've marked one such instance in the inline comments, but this pattern was used more often.

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -347,7 +347,7 @@ def forward(self, *inputs: Any) -> Any:
 
             magnitudes = magnitudes_fn(self.num_magnitude_bins, height, width)
             if magnitudes is not None:
-                magnitude = float(magnitudes[int(torch.randint(self.num_magnitude_bins, ()))])
+                magnitude = float(magnitudes[self.magnitude])
```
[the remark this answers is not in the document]
pmeier: Is there a PR that fixes this on main?

torchvision/prototype/transforms/_meta.py
datumbox: I believe earlier you introduced the usage of `DType = Union[torch.Tensor, PIL.Image.Image, features._Feature]`, which is also redundant for exactly the same reason.

torchvision/prototype/transforms/_deprecated.py
datumbox: `DType` is already used across the codebase. I was under the impression that you introduced but probably is @vfdev-5.

torchvision/prototype/transforms/_geometry.py
datumbox: Yeah fine by me.

torchvision/prototype/transforms/_deprecated.py
vfdev-5: It is me who introduced it

torchvision/prototype/features/_bounding_box.py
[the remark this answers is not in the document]
datumbox: As discussed offline with @vfdev-5, this check should move to the Kernel. I have sent a PR that does this but feel free to split across subPRs.

For reference what we decided is to validate input always on the kernel side. We also propose to ignore params only if they don't make sense for a given type (for example interpolation for SegmentationMask). Params which are unsupported should throw the right message on the lowlevel kernel side

torchvision/prototype/transforms/_auto_augment.py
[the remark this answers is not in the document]
pmeier: My bad, I was under the impression that there was a bug fix PR on `main` that we forgot to port to prototype. But IIUC, there was never a bug on `main` and we introduced this with the initial porting of `RandAugment()`, correct?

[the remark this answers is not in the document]
vfdev-5: Yes, there is a porting bug in prototype and it is not fully fixed even here. There remains few tweaks I'm working on.

torchvision/prototype/transforms/functional/_augment.py
```diff
@@ -19,7 +19,15 @@ def erase_image_pil(
     return to_pil_image(output, mode=img.mode)
 
 
-def erase(inpt: Any, i: int, j: int, h: int, w: int, v: torch.Tensor, inplace: bool = False) -> Any:
+def erase(
+    inpt: Union[torch.Tensor, PIL.Image.Image, features.Image],
```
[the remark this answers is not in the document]
vfdev-5: Shouldn't we also check inside `F.erase` for invalid types like BBoxes, Masks, Labels explicitly ?
As it is a public functional API, this code can be used without Transforms and thus can have a similar issue trying erase boxes...

torchvision/prototype/features/_bounding_box.py
vfdev-5: It is bad naming issue as two methods are not doing the same thing. 

`self._F._geometry._compute_output_size` refers to https://github.com/pytorch/vision/blob/97e0ea9c6ebc454538b3fa505e1d199547b0feed/torchvision/transforms/functional.py#L363-L383
and is used in `F.resize` to compute output size given specific options.

`self._F._geometry._FT._compute_output_size` refers to https://github.com/pytorch/vision/blob/97e0ea9c6ebc454538b3fa505e1d199547b0feed/torchvision/transforms/functional_tensor.py#L613-L643
and is used in `F.rotate` with `expand=True`. I suggest to rename them for clarity

torchvision/prototype/transforms/functional/_augment.py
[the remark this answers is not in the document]
datumbox: Right now, the new `F` follows the same idiom/approach as the old `F`. It checks for specific types in a series of if/elif and the else redirects to the default kernel (whichever this is). This can be error prone, I agree. What we could do is send a separate refactoring PR that changes the idiom on the new `F` to throw an exception on the `else` indicating that the input is not recognised. I wouldn't explicitly look for things we don't support but instead check for those that we do and throw the error on the `else` statement.

vfdev-5 (approved):
LGTM, thanks @datumbox !

PR #6516, opened 2022-08-30
Expand prototype transforms consistency tests

pmeier:
With this we'll have consistency checks for all deterministic transformations that retain type. The next steps will be:

- Add transformations that do not retain the type, e.g. `ToTensor`
- Add non-deterministic transformations that use the same random process for parameter generation. I think we can get away by simply setting `torch.manual_seed(123)` before the calls. For those where this is not sufficient, we maybe can align the parameter generation to avoid writing consistency checks manually.

In addition, this improves the framework a little for easier comparisons and better error messages.

### DIFF
--- a/test/test_prototype_transforms_consistency.py
+++ b/test/test_prototype_transforms_consistency.py
@@ -82,7 +82,18 @@ CONSISTENCY_CONFIGS = [
         [
             ArgsKwargs(32),
             ArgsKwargs((32, 29)),
-            ArgsKwargs((31, 28), interpolation=prototype_transforms.InterpolationMode.BICUBIC),
+            ArgsKwargs((31, 28), interpolation=prototype_transforms.InterpolationMode.NEAREST),
+            ArgsKwargs((33, 26), interpolation=prototype_transforms.InterpolationMode.BICUBIC),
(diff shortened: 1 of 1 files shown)


test/test_prototype_transforms_consistency.py
```diff
@@ -68,7 +96,18 @@ def parametrization(self):
         [
             ArgsKwargs(32),
             ArgsKwargs((32, 29)),
-            ArgsKwargs((31, 28), interpolation=prototype_transforms.InterpolationMode.BICUBIC),
+            ArgsKwargs((31, 28), interpolation=prototype_transforms.InterpolationMode.NEAREST),
+            ArgsKwargs((33, 26), interpolation=prototype_transforms.InterpolationMode.BICUBIC),
+            # FIXME: these are currently failing, since the new transform only supports the enum. The int input is
```
pmeier, line 101: Open question. We can also postpone a decision and potentially fix later.

vfdev-5: I think we should have dropped int values years ago. Let's keep your comment here as a reminder.

vfdev-5 (approved):
OK for me to merge

PR #6518, opened 2022-08-30
Expand prototype transforms consistency tests to all deterministic transformations

pmeier:
Follow-up to #6516. In addition to :point_up: this also fixes a bug in `Lambda`.

### DIFF
--- a/test/test_prototype_transforms_consistency.py
+++ b/test/test_prototype_transforms_consistency.py
@@ -1,3 +1,4 @@
+import enum
 import functools
 import itertools
 
@@ -8,6 +9,7 @@ import torch
 from test_prototype_transforms_functional import make_images
 from torch.testing._comparison import assert_equal as _assert_equal, TensorLikePair
 from torchvision import transforms as legacy_transforms
+from torchvision._utils import sequence_to_str
 from torchvision.prototype import features, transforms as prototype_transforms
 from torchvision.prototype.transforms.functional import to_image_pil, to_image_tensor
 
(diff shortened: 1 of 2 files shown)


test/test_prototype_transforms_consistency.py
```diff
@@ -174,9 +178,83 @@ def parametrization(self):
...
+        "AugMix",
+        "TrivialAugmentWide",
+        "GaussianBlur",
+        "RandAugment",
+        "AutoAugment",
+        "ColorJitter",
+        "ElasticTransform",
+    }
+
+    prototype = {config.legacy_cls.__name__ for config in CONSISTENCY_CONFIGS}
```
vfdev-5, line 248: Shouldn't this be ?
```suggestion
    prototype = {config.prototype_cls.__name__ for config in CONSISTENCY_CONFIGS}
```

pmeier: Nope. In some cases we just aliased some prototype transforms to their "legacy" name, for example 

https://github.com/pytorch/vision/blob/97e0ea9c6ebc454538b3fa505e1d199547b0feed/torchvision/prototype/transforms/_type_conversion.py#L60

```py
>>> from torchvision.prototype.transforms import ToPILImage
>>> ToPILImage.__name__
'ToImagePIL'
```

If we now check against the names of the transforms in the `torchvision.transforms` namespace, we get a naming mismatch. By also using the name of the legacy transform here, this goes away.

vfdev-5 (approved):
Looks ok to me

PR #6521, opened 2022-08-30
Add Video SwinTransformer

### LINKED ISSUE #6499: Port SwinTransformer3d from torchmultimodal
### 🚀 The feature

The main Idea is to port the SwinTransformer3d model from torchmulitmodal to torchvision.

Need to keep in mind the nuances and code structure of torchvision

https://github.com/facebookresearch/multimodal/blob/main/torchmultimodal/modules/encoders/swin_transformer_3d_encoder.py

https://github.com/facebookresearch/multimodal/blob/main/examples/omnivore/LoadOriginalPretrainedWeightAndCompare.ipynb

We need to port the implementation as well as the weights.

### Motivation, pitch

The idea is to first port SwinTransformer3dV1 and port its weights successfully. Once done we can then think of having SwinTransformer3dV2 (there is no such paper or implementation but maybe it will benefit like the 2d case)

### Alternatives

_No response_

### Additional context

Additionally in discussion with @YosuaMichael the paper also mentioned that SwinTransformerV2 can be used for object detection tasks. If possible we should explore it (but only after we finish previous things)

oke-aditya:
Closes #6499 

I think ready for a initial round of review. Will love to understand a bit more and then continue.

 I will mostly to wrap up this before the end of month. Little bit more work is pending :smile:


oke-aditya:
I got the hang of SwinTransformer2d. Soon I will take this up.

oke-aditya:
This was mega learning and refresher for me as I haven't at all worked on models / attention. 
(My ML skills probably are waning as I don't much work on it these days)

OK so after a long wait, I have completely read both the papers swin image and swin video. I'm currently working to align the image and video codebases and would then write down the model configurations. I have a few comments though

torchvision/models/video/swin_transformer.py
```diff
@@ -0,0 +1,409 @@
...
+
+__all__ = ["SwinTransformer3d"]
+
+
+def _compute_pad_size_3d(size_dhw: Tuple[int, int, int], patch_size: Tuple[int, int, int]) -> Tuple[int, int, int]:
+    pad_size = [(patch_size[i] - size_dhw[i] % patch_size[i]) % patch_size[i] for i in range(3)]
+    return pad_size[0], pad_size[1], pad_size[2]
+
+
+def _compute_attention_mask_3d(
```
oke-aditya, line 21: I think there is no such function in 2d case. This is just in 3d

torchvision/models/video/swin_transformer.py
```diff
@@ -0,0 +1,409 @@
...
+        num_heads (int): Number of attention heads.
+        shift_size (List[int]): Shift size for shifted window attention (T, H, W).
+        attention_dropout (float): Dropout ratio of attention weight. Default: 0.0.
+        dropout (float): Dropout ratio of output. Default: 0.0.
+        qkv_bias (Tensor[out_dim], optional): The bias tensor of query, key, value. Default: None.
+        proj_bias (Tensor[out_dim], optional): The bias tensor of projection. Default: None.
+    Returns:
+        Tensor[B, T, H, W, C]: The output tensor after shifted window attention.
+    """
+    b, t, h, w, c = input.shape
```
oke-aditya, line 93: one major change my IDE (mostly) and I did was rename D -> T and d -> t. As this is what used to describe temporal dimension in our codebase. I checked this with Vasilis. (I think T is correct and Temporal is more appropriate wording than Depth D)

torchvision/models/video/swin_transformer.py
```diff
@@ -0,0 +1,409 @@
...
+        self.shift_size = shift_size
+        self.num_heads = num_heads
+        self.attention_dropout = attention_dropout
+        self.dropout = dropout
+
+        self.qkv = nn.Linear(dim, dim * 3, bias=qkv_bias)
+        self.proj = nn.Linear(dim, dim, bias=proj_bias)
+
+        # define a parameter table of relative position bias
+        self.relative_position_bias_table = nn.Parameter(
```
oke-aditya, line 199: IMO this should be refactored as functions as done in image case.

torchvision/models/video/swin_transformer.py
```diff
@@ -0,0 +1,409 @@
...
+            torch.zeros(
+                (2 * window_size[0] - 1) * (2 * window_size[1] - 1) * (2 * window_size[2] - 1),
+                num_heads,
+            )
+        )  # 2*Wd-1 * 2*Wh-1 * 2*Ww-1, nH
+
+        # get pair-wise relative position index for each token inside the window
+        coords_dhw = [torch.arange(self.window_size[i]) for i in range(3)]
```
oke-aditya, line 207: This too same as above.

torchvision/models/video/swin_transformer.py
```diff
@@ -0,0 +1,409 @@
...
+        window_size, shift_size = self.window_size.copy(), self.shift_size.copy()
+        # Handle case where window_size is larger than the input tensor
+        for i in range(3):
+            if size_dhw[i] <= window_size[i]:
+                # In this case, window_size will adapt to the input size, and no need to shift
+                window_size[i] = size_dhw[i]
+                shift_size[i] = 0
+
+        window_vol = window_size[0] * window_size[1] * window_size[2]
+        relative_position_bias = self.relative_position_bias_table[
```
oke-aditya, line 237: And this too can be rewritten

torchvision/models/video/swin_transformer.py
```diff
@@ -0,0 +1,409 @@
...
+        self.register_buffer("relative_position_index", relative_position_index)
+
+        nn.init.trunc_normal_(self.relative_position_bias_table, std=0.02)
+
+    def forward(self, x: Tensor) -> Tensor:
+        _, d, h, w, _ = x.shape
+        size_dhw = (d, h, w)
+        window_size, shift_size = self.window_size.copy(), self.shift_size.copy()
+        # Handle case where window_size is larger than the input tensor
+        for i in range(3):
```
oke-aditya, line 230: Cab this somehow come out of loop and be similar to image case? I'm curios, my intuition is no coz we need to loop over.

torchvision/models/video/swin_transformer.py
```diff
@@ -0,0 +1,409 @@
...
+        self.proj = nn.Conv3d(
+            in_channels,
+            embed_dim,
+            kernel_size=self.tuple_patch_size,
+            stride=self.tuple_patch_size,
+        )
+        if norm_layer is not None:
+            self.norm = norm_layer(embed_dim)
+        else:
+            self.norm = None
```
oke-aditya, line 289: mypy doesn't love this

torchvision/models/video/swin_transformer.py
```diff
@@ -0,0 +1,409 @@
...
+        self,
+        patch_size: List[int],
+        embed_dim: int,
+        depths: List[int],
+        num_heads: List[int],
+        window_size: List[int],
+        mlp_ratio: float = 4.0,
+        dropout: float = 0.0,
+        attention_dropout: float = 0.0,
+        stochastic_depth_prob: float = 0.1,
```
oke-aditya, line 334: I changed this to 0.1 from 0.0 to align with Swin2d

torchvision/models/video/swin_transformer.py
```diff
@@ -0,0 +1,409 @@
...
+        num_classes: int = 400,
```
oke-aditya, line 335: As discussed with @YosuaMichael 
As we aren't doing any feature extraction, I have removed the optional head and changed the num_classes.

torchvision/models/video/swin_transformer.py
```diff
@@ -0,0 +1,409 @@
...
+        super().__init__()
+        self.num_classes = num_classes
+
+        if block is None:
+            block = partial(SwinTransformerBlock, attn_layer=ShiftedWindowAttention3d)
+
+        if norm_layer is None:
+            norm_layer = partial(nn.LayerNorm, eps=1e-5)
+
+        if patch_embed is None:
```
oke-aditya, line 350: Somehow I am not convinced with this style of writing both in 2d and 3d, why can't we give defaults in the function definition? Is there any reason?

torchvision/models/video/swin_transformer.py
```diff
@@ -0,0 +1,409 @@
...
+            attention_dropout=self.attention_dropout,
+            dropout=self.dropout,
+            qkv_bias=self.qkv.bias,
+            proj_bias=self.proj.bias,
+        )
+
+
+# Modified from:
+# https://github.com/SwinTransformer/Video-Swin-Transformer/blob/master/mmaction/models/backbones/swin_transformer.py
+class PatchEmbed3d(nn.Module):
```
oke-aditya, line 260: Just curious what was equivalent to this in 2d case and where is the codeblock for same?

torchvision/models/video/swin_transformer.py
```diff
@@ -0,0 +1,409 @@
...
+        input (Tensor[B, T, H, W, C]): The input tensor, 5-dimensions.
+        qkv_weight (Tensor[in_dim, out_dim]): The weight tensor of query, key, value.
+        proj_weight (Tensor[out_dim, out_dim]): The weight tensor of projection.
+        relative_position_bias (Tensor): The learned relative position bias added to attention.
+        window_size (List[int]): 3-dimensions window size, T, H, W .
+        num_heads (int): Number of attention heads.
+        shift_size (List[int]): Shift size for shifted window attention (T, H, W).
+        attention_dropout (float): Dropout ratio of attention weight. Default: 0.0.
+        dropout (float): Dropout ratio of output. Default: 0.0.
+        qkv_bias (Tensor[out_dim], optional): The bias tensor of query, key, value. Default: None.
```
oke-aditya, line 88: I believe we don't need `logit_scale` as we are adopting the swinv1 version from 2d?

oke-aditya (commented):
I think couple of things should be added.

1. Logging.
2. torch.fx.wrap() 

I would love to understand what torch.fx.wrap() does. :smile:

torchvision/models/video/swin_transformer.py
YosuaMichael: I think the only way to remove the loop is to do comparison one by one (3 if statement).

Btw, take note that on this specific logic, it is actually difference between video swin transformer and the 2d.
On [our 2d implementation](https://github.com/pytorch/vision/blob/main/torchvision/models/swin_transformer.py#L157), we dont change the window size but only set the shift to be 0.

On [the original 3d implementation](https://github.com/SwinTransformer/Video-Swin-Transformer/blob/master/mmaction/models/backbones/swin_transformer.py#L217), we actually resize the window_size like how it is implemented here.

torchvision/models/video/swin_transformer.py
YosuaMichael: Yeah, for V2 we should separate this into separate function

torchvision/models/video/swin_transformer.py
YosuaMichael: This actually implemented directly in the main class in the 2d version:
https://github.com/pytorch/vision/blob/main/torchvision/models/swin_transformer.py#L548-L557

In the 3d version we need to separate this into separate class because when `Omnivore` use this swin3d as encoder, it need to use different way for creating patch embedding: https://github.com/facebookresearch/multimodal/blob/main/torchmultimodal/models/omnivore.py#L144

torchvision/models/video/swin_transformer.py
YosuaMichael: maybe we should use `nn.Identity` to make mypy happy?

torchvision/models/video/swin_transformer.py
YosuaMichael: Actually I follow the 2d version style for this, I guess for the previous one it kinda make sense if we use `partial(...)`, but I agree that we might as well put this as default directly in the function definition for `patch_embed`. Can't really think the advantage using this style now.

YosuaMichael (commented):
Hi @oke-aditya , nice work! For this first review, I address some of your question and concern. I will do further review tomorrow :)

torchvision/models/video/swin_transformer.py
oke-aditya: I think we can refactor the 2d implementation a bit, the 3d looks much better here :smile:

torchvision/models/video/swin_transformer.py
oke-aditya: I wish we can separate in the 2d version too.

torchvision/models/video/swin_transformer.py
oke-aditya: I hope we can refactor the implementation, but it might be BC breaking. (minor bc break though)

torchvision/models/video/swin_transformer.py
oke-aditya: I need to understand this better, as why is this the case

oke-aditya:
Ok so this seems almost done. I guess we are ready to port weights. Can you verify the configurations @YosuaMichael ?

oke-aditya:
Some bug with relative position bias . Running locally I can trace error here.
My suspicion is that we are doing something wrong in 

```
def _get_relative_position_bias(
    relative_position_bias_table: torch.Tensor, relative_position_index: torch.Tensor, window_size: List[int]
) -> Tensor:
    window_vol = window_size[0] * window_size[1] * window_size[2]
    # In 3d case we flatten the relative_position_bias
    relative_position_bias = relative_position_bias_table[
        relative_position_index[:window_vol, :window_vol].flatten()  # type: ignore[index]
    ]
    relative_position_bias = relative_position_bias.view(window_vol, window_vol, -1)
    relative_position_bias = relative_position_bias.permute(2, 0, 1).contiguous().unsqueeze(0)
    return relative_position_bias
```

Error log, 

```
  File "/home/oke/Aditya/Pytorch/vision/torchvision/models/video/swin_transformer.py", line 167, in shifted_window_attention_3d
    attn = attn + relative_position_bias
RuntimeError: The size of tensor a (49) must match the size of tensor b (392) at non-singleton dimension 3

```

YosuaMichael:
> Some bug with relative position bias . Running locally I can trace error here. My suspicion is that we are doing something wrong in
> 
> ```
> def _get_relative_position_bias(
>     relative_position_bias_table: torch.Tensor, relative_position_index: torch.Tensor, window_size: List[int]
> ) -> Tensor:
>     window_vol = window_size[0] * window_size[1] * window_size[2]
>     # In 3d case we flatten the relative_position_bias
>     relative_position_bias = relative_position_bias_table[
>         relative_position_index[:window_vol, :window_vol].flatten()  # type: ignore[index]
>     ]
>     relative_position_bias = relative_position_bias.view(window_vol, window_vol, -1)
>     relative_position_bias = relative_position_bias.permute(2, 0, 1).contiguous().unsqueeze(0)
>     return relative_position_bias
> ```
> 
> Error log,
> 
> ```
>   File "/home/oke/Aditya/Pytorch/vision/torchvision/models/video/swin_transformer.py", line 167, in shifted_window_attention_3d
>     attn = attn + relative_position_bias
> RuntimeError: The size of tensor a (49) must match the size of tensor b (392) at non-singleton dimension 3
> ```

Hi @oke-aditya can you give a snippet on how to trigger this error? I assume you try a forward pass on the model, and knowing the input you give to the model will be helpful

oke-aditya:
I just ran the test_models.py

oke-aditya:
Hey @YosuaMichael can you check from your side.

torchvision/models/video/swin_transformer.py
```diff
@@ -0,0 +1,671 @@
...
+    """
+    weights = Swin3D_S_Weights.verify(weights)
+
+    return _swin_transformer3d(
+        patch_size=[2, 4, 4],
+        embed_dim=96,
+        depths=[2, 2, 18, 2],
+        num_heads=[3, 6, 12, 24],
+        window_size=[8, 7, 7],
+        stochastic_depth_prob=0.2,
```
YosuaMichael, line 629: NIT, The [original weights config](https://github.com/SwinTransformer/Video-Swin-Transformer/blob/master/configs/recognition/swin/swin_small_patch244_window877_kinetics400_1k.py#L4) seems to have `drop_path_rate=0.1`

torchvision/models/video/swin_transformer.py
```diff
@@ -0,0 +1,671 @@
...
+    """
+    weights = Swin3D_T_Weights.verify(weights)
+
+    return _swin_transformer3d(
+        patch_size=[2, 4, 4],
+        embed_dim=96,
+        depths=[2, 2, 6, 2],
+        num_heads=[3, 6, 12, 24],
+        window_size=[8, 7, 7],
+        stochastic_depth_prob=0.2,
```
YosuaMichael, line 591: NIT, the [original weights config ](https://github.com/SwinTransformer/Video-Swin-Transformer/blob/master/configs/recognition/swin/swin_tiny_patch244_window877_kinetics400_1k.py#L4)seems to have `drop_path_rate=0.1`

YosuaMichael (commented):
Thanks @oke-aditya  for the changes and fix. I have a few nits regarding the config. Other than that, I think we are ready to port the weight.

Just for context for others (or me in the future), @oke-aditya and me has a call to debug the bug mentioned previously and we find out that the problem is on the function `get_relative_position_bias` that using `self.window_size` but it should use the updated `window_size`.

YosuaMichael:
Also looking at the failing test, seems like it has something to do with different result on the expected file

oke-aditya:
Also @YosuaMichael Did you check with Omnivore team about the minor differences we noticed, apart from patch_size which is now [2, 4, 4].

YosuaMichael:
> Also @YosuaMichael Did you check with Omnivore team about the minor differences we noticed, apart from patch_size which is now [2, 4, 4].

I think the weight provided by the original author use patch_size = [2,4,4] @oke-aditya 
example: https://github.com/SwinTransformer/Video-Swin-Transformer/blob/master/configs/recognition/swin/swin_tiny_patch244_window877_kinetics400_1k.py#L4

torchvision/models/video/swin_transformer.py
```diff
@@ -0,0 +1,682 @@
...
+    "Swin3D_B_Weights",
+    "swin3d_t",
+    "swin3d_s",
+    "swin3d_b",
+]
+
+
+def _get_window_and_shift_size(
+    self, shift_size: List[int], size_dhw: List[int], window_size: List[int]
+) -> Tuple[List[int], List[int]]:
```
oke-aditya, line 34:
```suggestion
) -> Tuple[List[int], List[int]]:  # type: ignore[index]
```

oke-aditya:
Do we have `"min_size": (256, 256),` min_video_size or something equivalent?

oke-aditya:
Can we somehow get this into this release due? cc @datumbox @YosuaMichael ?
Its bit touch and go We are just left to port the weights. The code and tests are done.

oke-aditya:
Well the CI fails due to some gap in expected files which I generated locally. Those computations have a slight gap on CI vs my system. Any way to fix them ?

oke-aditya:
Just a quick update. 
We plan to continue working on this after the release. This will give us sufficient time to port all the weights!  

cc @datumbox

oke-aditya:
@datumbox 
Quick update, @YosuaMichael and I have resumed work on this. @YosuaMichael might kick off a run to check for initial set of weights. Hopefully we are in luck.

YosuaMichael:
Thanks @oke-aditya for the ported weight. I [modify](https://github.com/YosuaMichael/vision/blob/swin3d_validation/torchvision/models/video/swin_transformer.py#L508) the code from this branch a bit by adding the weight and transforms to mimic the original paper. Then I ran the weight to do testing on Kinetics-400 validation data using the following command:
```
    --timeout 3000 --ngpus 8 --nodes 1 \
    --data-path="/datasets01/kinetics/070618/400/" \
    --batch-size=16 --test-only \
    --clip-len 32 --frame-rate 15 --clips-per-video 12 \
    --cache-dataset \
    --model {model_name} --weights="{weight_name}"
```
Note: I use clips-per-video==12 to simulate the paper that use 4 spatial x 3 temporal clips since as of now we dont have the spatial clips capability. Thanks @datumbox suggestion for this.

Where the pair of the model_name and weight_name with their corresponding accuracy are as follow:

1. model_name=**swin3d_t**, weight_name=**Swin3D_T_Weights.KINETICS400_V1**
```
 * Video Acc@1 77.715 Video Acc@5 93.519
```
2. model_name=**swin3d_s**, weight_name=**Swin3D_S_Weights.KINETICS400_V1**
```
 * Video Acc@1 79.521 Video Acc@5 94.158
```
3. model_name=**swin3d_b**, weight_name=**Swin3D_B_Weights.KINETICS400_V1**
```
 * Video Acc@1 79.427 Video Acc@5 94.386
```
4. model_name=**swin3d_b**, weight_name=**Swin3D_B_Weights.IN22K_KINETICS400_V1**
```
 * Video Acc@1 81.643 Video Acc@5 95.574
```

datumbox:
@YosuaMichael Looks good. We should add the info about the input parameters passed to the ref script on the docs similar to what we do for MViT:
https://github.com/pytorch/vision/blob/cffb7f7f5015f09d12c585c2566b034cdd92f583/torchvision/models/video/mvit.py#L616-L619

oke-aditya:
Nice, I will wrap this over the weekend!

oke-aditya:
I kind of disagree to the name `IN22K_KINETICS400_V1` Since for Swag models we put the names for other pretrained dataset as last.
E.g
Default one

```
 IMAGENET1K_V2 = Weights(
```

Swag one

```
IMAGENET1K_SWAG_E2E_V1 = Weights(
```

I would suggest for KINETICS we do

```
KINETICS400_V1 
```

and 

```
KINETICS400_IMAGENET22K_V1 = Weights(
```

oke-aditya:
Btw also how do I get these 3 fields? 

```
 "min_size": (224, 224),
 "min_temporal_size": 16,
"num_params": 36610672,
```

datumbox:
@oke-aditya Thanks for the update.

> KINETICS400_IMAGENET22K_V1

Sounds good to me.

For `min_size` its the minimum spatial size that can be sent to your architecture without throwing error. Sometimes maxpooling or conv layers impose a minimum size below which the network returns an error. I usually try a couple of values until I find the one that breaks the architecture and then set the `min_size` to the next available permitted value.

For `min_temporal_size` it's exactly the same but for the temporal `T` dimension.

Finally for `num_params`, just do `sum(p.numel() for p in model.parameters())` to estimate them.

YosuaMichael:
Looks good @oke-aditya , I agree with the naming `KINETICS_IMAGENET22K_V1`!
For `min_size, temporal_size, and num_params` it is as @datumbox explained.
I think swin transformer 3d dont really have `min_size` and `min_temporal_size` since we do [padding in PatchEmbed3d](https://github.com/oke-aditya/vision/blob/0cd56abb844b04a19fdcfa44fc05b89ab654ea43/torchvision/models/video/swin_transformer.py#L356) (this means the min_size=(1,1) and min_temporal_size=1), however please give it a try first.

oke-aditya:
So I checked the min_size and temporal size. It seems even  (1,1) would work fine. And 1 for temporal size is fine too

```
>>> from torchvision.models.video import swin3d_b
>>> v_tensor = torch.randn(1, 3, 1, 1, 1) # b c t h w
>>> v_tensor = v_tensor.cuda()
>>> model = swin3d_b()
>>> out = model(v_tensor)
```

oke-aditya:
@YosuaMichael @datumbox  I have added the FLOPs and weights size by running the script in #6936 

I think we should be good to merge.

oke-aditya:
This should be good to go. CI failure is unrelated

YosuaMichael (approved):
Thanks for the hard work @oke-aditya, it looks good now!

YosuaMichael:
The test failures are unrelated so we merged.

PR #6522, opened 2022-08-30
Expand prototype transforms consistency tests to most random tests

pmeier:
Follow-up to #6516. Fortunately, we didn't touch the random parameter generation during porting and so setting the random seed if sufficient for most transformations. The only ones left are the AA transforms, for which @vfdev-5 is adding checks in #6519.

Apart from the tests, this PR also fixes some bugs in `RandomAffine`, `RandomPerspective`, and `GaussianBlur`.


torchvision/prototype/transforms/_geometry.py
```diff
@@ -389,7 +389,7 @@ def _get_params(self, sample: Any) -> Dict[str, Any]:
                 shear_y = float(torch.empty(1).uniform_(self.shear[2], self.shear[3]).item())
 
         shear = (shear_x, shear_y)
-        return dict(angle=angle, translations=translations, scale=scale, shear=shear)
+        return dict(angle=angle, translate=translate, scale=scale, shear=shear)
```
vfdev-5, line 392: Nice catch! Seems like I just copied the name from stable RandomAffine.get_params but unfortunately previous tests didn't revealed that.

vfdev-5 (approved):
Thanks @pmeier, lgtm!

PR #6525, opened 2022-08-31
add consistency tests for prototype container transforms

pmeier:
Per title. 

When designing the prototype API, we intentionally only used `torch`'s random capabilities to generate parameters everywhere. However, on the legacy API there are two transforms that don't:
There are two outliers: 

- `RandomChoice`:
    https://github.com/pytorch/vision/blob/cc9ceb54d404f8c6416fb71306881f0d26569bea/torchvision/transforms/transforms.py#L565-L567
- `RandomOrder`:
    https://github.com/pytorch/vision/blob/cc9ceb54d404f8c6416fb71306881f0d26569bea/torchvision/transforms/transforms.py#L548-L553

For the former we can at least test for `p=0` and `p=1`, but for the latter I have no idea how to test other than manually. Given that code is very simple, I want to avoid complicated mocking like we are doing in #6519.

The equivalent logic on the prototype transforms is:

- `RandomChoice`:
    https://github.com/pytorch/vision/blob/cc9ceb54d404f8c6416fb71306881f0d26569bea/torchvision/prototype/transforms/_container.py#L66-L69
- `RandomOrder`:
    https://github.com/pytorch/vision/blob/cc9ceb54d404f8c6416fb71306881f0d26569bea/torchvision/prototype/transforms/_container.py#L79-L84

Basically we need to make sure that they generate the same distribution. I confirmed that they are consistent [visually](https://gist.github.com/pmeier/6d7666b0843d6a9149ce89f3f6a3e9dc).

Apart from the testing, I also fixed a consistency bug in `RandomApply`.

### DIFF
--- a/test/test_prototype_transforms_consistency.py
+++ b/test/test_prototype_transforms_consistency.py
@@ -464,38 +464,18 @@ def test_automatic_coverage_deterministic():
         )
 
 
-@pytest.mark.parametrize(
-    ("prototype_transform_cls", "legacy_transform_cls", "args_kwargs", "make_images_kwargs", "supports_pil"),
-    itertools.chain.from_iterable(config.parametrization() for config in CONSISTENCY_CONFIGS),
-)
-def test_consistency(prototype_transform_cls, legacy_transform_cls, args_kwargs, make_images_kwargs, supports_pil):
-    args, kwargs = args_kwargs
+def check_consistency(prototype_transform, legacy_transform, images=None, supports_pil=True):
+    if images is None:
+        images = make_images(**DEFAULT_MAKE_IMAGES_KWARGS)
 
-    try:
-        legacy = legacy_transform_cls(*args, **kwargs)
-    except Exception as exc:
-        raise pytest.UsageError(
-            f"Initializing the legacy transform failed with the error above. "
-            f"Please correct the `ArgsKwargs({args_kwargs})` in the `ConsistencyConfig`."
-        ) from exc
-
-    try:
-        prototype = prototype_transform_cls(*args, **kwargs)
-    except Exception as exc:
-        raise AssertionError(
-            "Initializing the prototype transform failed with the error above. "
-            "This means there is a consistency bug in the constructor."
-        ) from exc
-
-    for image in make_images(**make_images_kwargs):
-        image_tensor = torch.Tensor(image)
-        image_pil = to_image_pil(image) if image.ndim == 3 and supports_pil else None
-
+    for image in images:
         image_repr = f"[{tuple(image.shape)}, {str(image.dtype).rsplit('.')[-1]}]"
+
+        image_tensor = torch.Tensor(image)
 
         try:
             torch.manual_seed(0)
-            output_legacy_tensor = legacy(image_tensor)
+            output_legacy_tensor = legacy_transform(image_tensor)
         except Exception as exc:
             raise pytest.UsageError(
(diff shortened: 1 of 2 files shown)


test/test_prototype_transforms_consistency.py
```diff
@@ -563,23 +545,116 @@ def test_consistency(prototype_transform_cls, legacy_transform_cls, args_kwargs,
...
+            ],
+            p=p,
+        )
+
+        check_consistency(prototype_transform, legacy_transform)
+
+
 class TestToTensorTransforms:
     def test_pil_to_tensor(self):
+        prototype_transform = prototype_transforms.PILToTensor()
```
pmeier, line 643: We don't need to create the transform for ever image.

torchvision/prototype/transforms/_container.py
```diff
@@ -21,16 +19,21 @@ def forward(self, *inputs: Any) -> Any:
         return sample
 
 
-class RandomApply(_RandomApplyTransform):
-    def __init__(self, transform: Transform, *, p: float = 0.5) -> None:
-        super().__init__(p=p)
-        self.transform = transform
+class RandomApply(Compose):
```
pmeier, line 22: I don't really understand why `RandomApply` needs to support multiple transforms. I've re-added it here for BC, but I think we should deprecate this in favor of what our internal `_RandomApplyTransform` does. If the user needs multiple transforms, they can simply wrap them in a `Compose` before passing it to `RandomApply`.

torchvision/prototype/transforms/_container.py
```diff
@@ -21,16 +19,21 @@ def forward(self, *inputs: Any) -> Any:
...
+    def __init__(self, transforms: Sequence[Callable], *, p: float = 0.5) -> None:
```
vfdev-5, line 23: Is it BC that `p` is kwargs only now ?

torchvision/prototype/transforms/_container.py
vfdev-5: I agree. Maybe, we can extend `Compose` with `p` argument to cover `RandomApply` feature...

torchvision/prototype/transforms/_container.py
datumbox: I have corrected this at https://github.com/pytorch/vision/pull/6512. I'm not sure what was the rational of splitting and cherrypicking my commits into multiple PRs but please make sure you won't miss any of the commits that fixes this.

torchvision/prototype/transforms/_container.py
datumbox: I think for now we should keep it and review on the future what we want to deprecate. It's true that many of the transforms can be written as a combination of other transforms. For example, the majority of the `Random*` transforms that support a `p` probability could have used the `RandomApply`. Unfortunately dropping some of these functionalities not only breaks BC but also leads to more verbose code. So we should have a separate discussion over what should be deprecated and why.

torchvision/prototype/transforms/_container.py
vfdev-5: Well, maybe it is not the best approach. I did that as RandAug was still broken in your PR...

pmeier: Since we have merged #6512 before this one, it created a merge conflict.

vfdev-5 (approved):
Looks ok to me, thanks @pmeier

PR #6526, opened 2022-08-31
add test for signature consistency of prototype and legacy transforms

pmeier:
Per title. Follow-up to https://github.com/pytorch/vision/pull/6525#discussion_r959369318 to avoid manually checking everything. In particular, the test includes

- a check that no parameter that is available on the legacy transform is missing on the prototype counterpart
- a check that potentially new parameters on the prototype transform have a default value and thus the transform can be created without passing it
- a check that parameters that are available on both transforms are of the same kind, e.g. keyword or positional

This test found two consistency bugs on `CenterCrop` and `Lambda` that were also fixed. 

Left over are two categories of signature divergences:

1. Some parameters are not included on the prototype transform since they are already deprecated and scheduled for removal before the roll-out. This includes:

    - https://github.com/pytorch/vision/blob/a5b3118f4650918d82819e80144fbdb91929f953/torchvision/transforms/transforms.py#L1408-L1414
    - https://github.com/pytorch/vision/blob/a5b3118f4650918d82819e80144fbdb91929f953/torchvision/transforms/transforms.py#L1292-L1295

    I've opted to simply ignore them in the test, since I see no reason to add them to the prototype transforms now to just remove them again before the release.
2. Some parameters are not included on the prototype transform since we no longer want the behavior, but we haven't started an official deprecation yet. This includes:

    - https://github.com/pytorch/vision/blob/a5b3118f4650918d82819e80144fbdb91929f953/torchvision/transforms/transforms.py#L251
    - https://github.com/pytorch/vision/blob/a5b3118f4650918d82819e80144fbdb91929f953/torchvision/transforms/transforms.py#L1651

    Since inplace operations can improve the performance, we might walk back on that unofficial decision and put these back after all. For now, I also added them to the ignore list for the test, but we need to decide if we want that or not.

### DIFF
--- a/test/test_prototype_transforms_consistency.py
+++ b/test/test_prototype_transforms_consistency.py
@@ -1,11 +1,9 @@
 import enum
 import functools
-
 import inspect
 import itertools
 
 import numpy as np
-
 import PIL.Image
 import pytest
 
@@ -427,8 +425,8 @@ CONSISTENCY_CONFIGS = [
 ]
 
 
-def test_automatic_coverage_deterministic():
-    legacy = {
+def test_automatic_coverage():
+    available = {
         name
         for name, obj in legacy_transforms.__dict__.items()
         if not name.startswith("_")
@@ -454,9 +452,9 @@ def test_automatic_coverage_deterministic():
         }
     }
 
-    prototype = {config.legacy_cls.__name__ for config in CONSISTENCY_CONFIGS}
+    checked = {config.legacy_cls.__name__ for config in CONSISTENCY_CONFIGS}
 
-    missing = legacy - prototype
+    missing = available - checked
     if missing:
         raise AssertionError(
             f"The prototype transformations {sequence_to_str(sorted(missing), separate_last='and ')} "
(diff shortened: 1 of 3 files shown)


vfdev-5 (approved):
Thanks @pmeier !

test/test_prototype_transforms_consistency.py
```diff
@@ -93,6 +87,7 @@ def parametrization(self):
         ],
         supports_pil=False,
         make_images_kwargs=dict(DEFAULT_MAKE_IMAGES_KWARGS, dtypes=[torch.float]),
+        removed_params=["inplace"],
```
pmeier, line 90: Open question if we want to remove the `inplace` parameter. That would be BC breaking without a proper deprecation cycle.

test/test_prototype_transforms_consistency.py
```diff
@@ -319,6 +314,7 @@ def parametrization(self):
             ArgsKwargs(p=1, value="random"),
         ],
         supports_pil=False,
+        removed_params=["inplace"],
```
pmeier, line 317: Same as above.

test/test_prototype_transforms_consistency.py
pmeier: We decided to put it back. See #6530.

test/test_prototype_transforms_consistency.py
pmeier: We decided to put it back. See #6530.

PR #6527, opened 2022-08-31
test all transform signatures for consistency

pmeier:
This does three things:

1. Test the signature of all transforms for consistency
2. Fix the signature of `RandAugment` and `TrivialAugmentWide`
3. Port the AA consistency test to the "correct" module


test/test_prototype_transforms.py
```diff
@@ -1665,205 +1665,3 @@ def test__transform(self):
...
-    @pytest.mark.parametrize("antialias", [True, False])
-    @pytest.mark.parametrize(
-        "inpt",
-        [
-            torch.randint(0, 256, size=(1, 3, 256, 256), dtype=torch.uint8),
-            PIL.Image.new("RGB", (256, 256), 123),
-            features.Image(torch.randint(0, 256, size=(1, 3, 256, 256), dtype=torch.uint8)),
-        ],
-    )
-    def test_random_resized_crop(self, antialias, inpt):
```
pmeier, line 1680: I've removed this while moving this class, since this was already present in the automatic consistency checks.

test/test_prototype_transforms_consistency.py
```diff
@@ -422,34 +423,54 @@ def __init__(
...
 
 def test_automatic_coverage():
     available = {
         name
         for name, obj in legacy_transforms.__dict__.items()
-        if not name.startswith("_")
-        and isinstance(obj, type)
-        and not issubclass(obj, enum.Enum)
-        and name
-        not in {
```
pmeier, line 436: We no longer need an exclude list, since all transforms are now have a `ConsistencyConfig` :tada:

test/test_prototype_transforms_consistency.py
```diff
@@ -480,16 +501,22 @@ def test_signature_consistency(config):
         )
 
     extra = prototype_params.keys() - legacy_params.keys()
-    extra_without_default = {param for param in extra if prototype_params[param].default is not inspect.Parameter.empty}
+    extra_without_default = {
+        param
+        for param in extra
+        if prototype_params[param].default is inspect.Parameter.empty
+        and prototype_params[param].kind not in {inspect.Parameter.VAR_POSITIONAL, inspect.Parameter.VAR_KEYWORD}
```
pmeier, line 508: The check was faulty in that it also took `*args` or `**kwargs` as required. These are now explicitly excluded.

test/test_prototype_transforms_consistency.py
```diff
@@ -480,16 +501,22 @@ def test_signature_consistency(config):
...
-            f"but the legacy transform does not. Please add a default value."
+            f"The prototype transform requires the parameters "
+            f"{sequence_to_str(sorted(extra_without_default), separate_last='and ')}, but the legacy transform does "
+            f"not. Please add a default value."
         )
 
-    for name, legacy_param in legacy_params.items():
-        prototype_param = prototype_params[name]
-        assert prototype_param.kind is legacy_param.kind
+    legacy_kinds = {name: param.kind for name, param in legacy_params.items()}
```
pmeier, line 517: Instead of going through the parameters one by one and fail on the first mismatch, we now compute everything upfront. This has the benefit that all errors are visible and we don't have to re-run the test.

test/test_prototype_transforms_consistency.py
```diff
@@ -693,3 +720,165 @@ def test_to_tensor(self):
...
+        randint_values = iter(randint_values)
+
+        mocker.patch("torch.randint", side_effect=lambda *arg, **kwargs: torch.tensor(next(randint_values)))
+        mocker.patch("torch.rand", return_value=1.0)
+
+        for i in range(le):
+            expected_output = t_ref(inpt)
+            output = t(inpt)
+
+            assert_equal(expected_output, output)
```
pmeier, line 762: As explained in https://github.com/pytorch/vision/pull/6519#discussion_r958834371, we no longer manually need to convert to tensor before we compare. Plus, we compare without tolerances.

vfdev-5 (approved):
Thanks @pmeier !

PR #6536, opened 2022-09-04
Fix incorrect recipe for SSDlite320

datumbox:
The information on `ssdlite320_mobilenet_v3_large` recipe is wrong. I checked my notes and the network was trained without pre-training from ImageNet. The issue was detected while trying to reproduce the results at https://github.com/pytorch/vision/pull/6433#issuecomment-1235269893


NicolasHug (approved):
Thanks @datumbox

PR #6537, opened 2022-09-05
Update S3D weights

datumbox:
Continuation from work at #6412

Trained from scratch using the following recipe:
```
PYTHONPATH=$PYTHONPATH:`pwd` python -u run_with_submitit.py --ngpus 8 --nodes 8 --cache-dataset --batch-size=12 --lr 0.2 --clip-len 64 --clips-per-video 5 --sync-bn --model s3d --train-resize-size 256 256 --train-crop-size 224 224 --val-resize-size 256 256 --val-crop-size 224 224 --data-path="/datasets01_ontap_isolated/kinetics/070618/400/"
Submitted job_id: 59255
```

Validated with:
```
torchrun --nproc_per_node=8 train.py --data-path="/datasets01_ontap/kinetics/070618/400/" --batch-size=16 --test-only --cache-dataset --clip-len 128 --clips-per-video 1 --model s3d --weights S3D_Weights.KINETICS400_V1
 * Clip Acc@1 68.368 Clip Acc@5 88.050
```

### DIFF
--- a/references/video_classification/README.md
+++ b/references/video_classification/README.md
@@ -81,6 +81,7 @@ Video resnet models:
 ```
 # number of frames per clip
 --clip_len 16 \ 
+--frame-rate 15 \
 # allow for temporal jittering
 --clips_per_video 5 \
 --batch-size 24 \
--- a/references/video_classification/README.md
+++ b/references/video_classification/README.md
@@ -97,6 +97,21 @@ Video resnet models:
 --val-crop-size 112 112
 ```
 
+### S3D
+
+The S3D model was trained similarly to the above but with the following changes on the default configuration:
+```
+--batch-size=12 --lr 0.2 --clip-len 64 --clips-per-video 5 --sync-bn \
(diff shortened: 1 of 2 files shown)


references/video_classification/README.md
```diff
@@ -97,6 +97,21 @@ Video resnet models:
 --val-crop-size 112 112
 ```
 
+### S3D
+
+The S3D model was trained similarly to the above but with the following changes on the default configuration:
```
YosuaMichael, line 102: NIT: I know this is a default value, but I think we may want to put `--frame-rate 15` in the previous section (resnet) since I think this is a quite important parameters for video classification.

YosuaMichael (approved):
Thanks for the PR @datumbox ! Overall looks good to me, just a single minor NIT.

references/video_classification/README.md
datumbox: Sounds good, let me add it above.

PR #6539, opened 2022-09-05
Skip big models per platform/device

datumbox:
Addressing [broken CI](https://app.circleci.com/pipelines/github/pytorch/vision/19938/workflows/eb804ae7-3645-4e61-bbaa-4b11e0887027/jobs/1618516) caused by memory issues on Windows GPU.

It might be worth specifying if possible the skips for `vit_h_14` and `regnet_y_128gf`. Perhaps skipping only windows will do the trick (see #6189).

### DIFF
--- a/test/test_models.py
+++ b/test/test_models.py
@@ -7,7 +7,6 @@ import platform
 import sys
 import warnings
 from collections import OrderedDict
-from itertools import product
 from tempfile import TemporaryDirectory
 from typing import Any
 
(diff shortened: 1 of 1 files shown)


test/test_models.py
```diff
@@ -343,12 +345,27 @@ def _check_input_backprop(model, inputs):
...
 
-# skip big models to reduce memory usage on CI test
+# skip big models to reduce memory usage on CI test. We can exclude combinations of (platform-system, device).
+_all_platforms = ("Darwin", "Linux", "Windows")
+_all_devices = ("cpu", "cuda")
 skipped_big_models = {
-    "vit_h_14",
-    "regnet_y_128gf",
+    "vit_h_14": set(product(_all_platforms, _all_devices)),
+    "regnet_y_128gf": set(product(_all_platforms, _all_devices)),
```
pmeier, line 353: I see we already did this before. Is this a temporary thing? If not, how are we going to make sure that we don't introduce bugs in the future?

pmeier (approved):
LGTM, thanks Vasilis!

test/test_models.py
YosuaMichael: I have chatted with @datumbox, for vit_h_14 and regnet_y_128gf, previously it breaks on windows for both [CPU](https://github.com/pytorch/vision/issues/6189) and [GPU](https://app.circleci.com/pipelines/github/pytorch/vision/17065/workflows/85192520-7d29-4f6f-a850-8082ea4eb3f8/jobs/1382986).

I think this PR is a good idea to loosen the test for the big model so we can still test them on Linux and Mac if possible :)

I have looked at the code and it looks good to me too.

datumbox: Perfect thank for the references. I just pushed skipping only Windows for these tests as well. Let's see how this will go. If green, I'll clean up the PR and merge using the more specific skips as you both proposed.

PR #6540, opened 2022-09-06
[proto] Added extra_repr to Compose

vfdev-5:
Description:
- Added extra_repr to Compose to print composed transforms like in stable API

Currently:
```
Compose(transforms=[<function copy_targets at 0x134443dc0>, <__main__.WrapIntoFeatures object at 0x13444c5e0>, ScaleJitter(target_size=(1024, 1024), scale_range=(0.1, 2.0), interpolation=InterpolationMode.BILINEAR), FixedSizeCrop(crop_height=1024, crop_width=1024, fill=0, padding_mode=constant), RandomHorizontalFlip(p=1.0), <function get_detection_transforms_v2.<locals>.friendly_to_image_tensor at 0x134447a60>, ConvertImageDtype()])
```

This PR:
```
Compose(
      <function copy_targets at 0x13a568dc0>
      <__main__.WrapIntoFeatures object at 0x13a571640>
      ScaleJitter(target_size=(1024, 1024), scale_range=(0.1, 2.0), interpolation=InterpolationMode.BILINEAR)
      FixedSizeCrop(crop_height=1024, crop_width=1024, fill=0, padding_mode=constant)
      RandomHorizontalFlip(p=1.0)
      <function get_detection_transforms_v2.<locals>.friendly_to_image_tensor at 0x13a56ca60>
      ConvertImageDtype()
)
```

Stable API prints:
```
Compose(
    RandomResizedCrop(size=(224, 224), scale=(0.08, 1.0), ratio=(0.75, 1.3333), interpolation=bilinear), antialias=None)
    RandomHorizontalFlip(p=1.0)
    <function get_classification_transforms_stable_api.<locals>.friendly_pil_to_tensor at 0x134441af0>
    ConvertImageDtype()
    Normalize(mean=(0.485, 0.456, 0.406), std=(0.229, 0.224, 0.225))
)
```


pmeier:
That was one of the reasons I wanted to use register the callables as modules, because printing will be handles by `nn.Module` in that case. 

In #6391 you reverted that for BC. Maybe we can keep it and wrap the general callables into a `transforms.Lambda`? Something like

```py
class Compose(Transform):
    def __init__(self, transforms: Sequence[Callable]) -> None:
        super().__init__()
        if not isinstance(transforms, Sequence):
        for idx, transform in enumerate(transforms):
            self.add_module(str(idx), transform if isinstance(transform, nn.Module) else transforms.Lambda(transform))
```

That would align our transform with [`nn.Sequential`](https://github.com/pytorch/pytorch/blob/207a5a8fa9bfd9361038d46636c0440290c171bb/torch/nn/modules/container.py#L90-L91) as it also doesn't define a custom `__repr__`.

vfdev-5:
I think wrapping any callable with `transforms.Lambda` may be a BC-breaking in certain rare cases. On the other hand we want only to improve visual representation, so I do not think it would make sense to reintroduce submodules just for that...

datumbox (approved):
LGTM thanks!

PR #6541, opened 2022-09-06
[proto] Fixed bug in ScaleJitter with params

vfdev-5:
Description:
- Fixed bug in ScaleJitter with params


pmeier (approved):
LGTM, thanks Victor!

PR #6542, opened 2022-09-07
fix padding for degenerate segmentation masks

pmeier:
Before this PR, padding degenerate segmentation masks failed:

```py
import torch

mask = features.SegmentationMask(torch.randint(0, 2, (0, 937, 1024), dtype=torch.bool))
F.pad(mask, 5)
```

```
RuntimeError: cannot reshape tensor of 0 elements into shape [-1, 0, 937, 1024] because the unspecified dimension size -1 can be any value and is ambiguous
```

I fixed the kernel and extended the kernel tests for all input types (images, segmentation masks, bounding boxes, ...) to also check the degenerate case.

### DIFF
--- a/test/test_prototype_transforms.py
+++ b/test/test_prototype_transforms.py
@@ -86,8 +86,6 @@ class TestSmoke:
         transforms.RandomHorizontalFlip(),
         transforms.Pad(5),
         transforms.RandomZoomOut(),
-        transforms.RandomRotation(degrees=(-45, 45)),
-        transforms.RandomAffine(degrees=(-45, 45)),
         transforms.RandomCrop([16, 16], padding=1, pad_if_needed=True),
         # TODO: Something wrong with input data setup. Let's fix that
         # transforms.RandomEqualize(),
@@ -95,6 +93,8 @@ class TestSmoke:
         # transforms.RandomPosterize(bits=4),
         # transforms.RandomSolarize(threshold=0.5),
         # transforms.RandomAdjustSharpness(sharpness_factor=0.5),
+        # transforms.RandomRotation(degrees=(-45, 45)),
+        # transforms.RandomAffine(degrees=(-45, 45)),
     )
     def test_common(self, transform, input):
         transform(input)
@@ -1206,9 +1206,9 @@ class TestRandomIoUCrop:
         bboxes = make_bounding_box(format="XYXY", image_size=(32, 24), extra_dims=(6,))
         label = features.Label(torch.randint(0, 10, size=(6,)))
         ohe_label = features.OneHotLabel(torch.zeros(6, 10).scatter_(1, label.unsqueeze(1), 1))
-        masks = make_segmentation_mask((32, 24))
-        ohe_masks = features.SegmentationMask(torch.randint(0, 2, size=(6, 32, 24)))
-        sample = [image, bboxes, label, ohe_label, masks, ohe_masks]
+        masks = make_segmentation_mask((32, 24), num_objects=6)
+
+        sample = [image, bboxes, label, ohe_label, masks]
 
         fn = mocker.patch("torchvision.prototype.transforms.functional.crop", side_effect=lambda x, **params: x)
         is_within_crop_area = torch.tensor([0, 1, 0, 1, 0, 1], dtype=torch.bool)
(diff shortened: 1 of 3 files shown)


torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -586,15 +592,7 @@ def _pad_with_vector_fill(
 def pad_segmentation_mask(
     segmentation_mask: torch.Tensor, padding: Union[int, List[int]], padding_mode: str = "constant"
 ) -> torch.Tensor:
-    num_masks, height, width = segmentation_mask.shape[-3:]
```
pmeier, line 589: Not sure why we replicated `pad_image_tensor` here if the only difference is that we set a fixed `fill=0`.

test/test_prototype_transforms.py
```diff
@@ -86,15 +86,15 @@ class TestSmoke:
...
-        transforms.RandomAffine(degrees=(-45, 45)),
         transforms.RandomCrop([16, 16], padding=1, pad_if_needed=True),
         # TODO: Something wrong with input data setup. Let's fix that
         # transforms.RandomEqualize(),
         # transforms.RandomInvert(),
         # transforms.RandomPosterize(bits=4),
         # transforms.RandomSolarize(threshold=0.5),
         # transforms.RandomAdjustSharpness(sharpness_factor=0.5),
+        # transforms.RandomRotation(degrees=(-45, 45)),
+        # transforms.RandomAffine(degrees=(-45, 45)),
```
pmeier, line 97: They don't work since the new input data generation also adds degenerate shapes and that is not supported yet by the underlying kernels.

test/test_prototype_transforms.py
```diff
@@ -1206,9 +1206,9 @@ def test__transform(self, mocker):
         bboxes = make_bounding_box(format="XYXY", image_size=(32, 24), extra_dims=(6,))
         label = features.Label(torch.randint(0, 10, size=(6,)))
         ohe_label = features.OneHotLabel(torch.zeros(6, 10).scatter_(1, label.unsqueeze(1), 1))
-        masks = make_segmentation_mask((32, 24))
-        ohe_masks = features.SegmentationMask(torch.randint(0, 2, size=(6, 32, 24)))
-        sample = [image, bboxes, label, ohe_label, masks, ohe_masks]
+        masks = make_segmentation_mask((32, 24), num_objects=6)
```
pmeier, line 1209: We currently only work with object detection masks. They were called `ohe_masks` here before. I've renamed them and removed the tests for what was `masks` before.

test/test_prototype_transforms_functional.py
```diff
@@ -142,21 +145,26 @@ def make_one_hot_labels(
         yield make_one_hot_label(extra_dims_)
 
 
-def make_segmentation_mask(size=None, *, num_categories=80, extra_dims=(), dtype=torch.long):
-    size = size or torch.randint(16, 33, (2,)).tolist()
-    shape = (*extra_dims, 1, *size)
-    data = make_tensor(shape, low=0, high=num_categories, dtype=dtype)
+def make_segmentation_mask(size=None, *, num_objects=None, extra_dims=(), dtype=torch.uint8):
```
pmeier, line 148: The functionality generated masks for segmentation tasks, which we currently don't support.

test/test_prototype_transforms_functional.py
```diff
@@ -896,6 +917,13 @@ def test_correctness_affine_bounding_box_on_fixed_input(device):
     torch.testing.assert_close(output_boxes.tolist(), expected_bboxes)
 
 
+incorrect_expected_segmentation_mask_setup = pytest.mark.xfail(
+    reason="This test fails because the expected result computation is wrong. Fix ASAP.",
+    strict=False,
+)
```
pmeier, line 923: This is to get the CI green for now. As the text implies we should get to this ASAP. But doing it now would slow down the reference training.

pmeier (commented):
Apart from the ones I fixed, there are more operators that use the `.view(-1, ...)` idiom. 

- https://github.com/pytorch/vision/blob/4c073b09521604e410bac8a0b0fe0f2680724e0d/torchvision/prototype/transforms/functional/_geometry.py#L227
- https://github.com/pytorch/vision/blob/4c073b09521604e410bac8a0b0fe0f2680724e0d/torchvision/prototype/transforms/functional/_geometry.py#L448

I've xfailed some tests for them. Plus, there are more bounding box ops that do this. But we don't need to worry about them, because `.view(-1, ...)` only fails if there is a `0` in the `...` part. For bounding boxes we only do 

```py
>>> t = torch.empty(0, 4)
>>> t.view(-1, 4)
tensor([], size=(0, 4))
```

The ambiguity only arises for segmentation masks, which use the channel dimension as number of objects and that can be `0`.

```py
>>> t = torch.empty(4, 0, 16, 16)
>>> t.view(-1, 0, 16, 16)
RuntimeError: cannot reshape tensor of 0 elements into shape [-1, 0, 16, 16] because the unspecified dimension size -1 can be any value and is ambiguous
```

That might be a good reason to have something in the kernel other than just calling the image kernel. But we can discuss this later.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -550,11 +556,17 @@ def pad_image_tensor(
     num_channels, height, width = img.shape[-3:]
     extra_dims = img.shape[:-3]
 
-    padded_image = _FT.pad(
-        img=img.view(-1, num_channels, height, width), padding=padding, fill=fill, padding_mode=padding_mode
-    )
+    left, right, top, bottom = _FT._parse_pad_padding(padding)
```
vfdev-5, line 559: Nit: this part of code can be put into `else` part to avoid parsing twice for normal case

PR #6544, opened 2022-09-08
Cleanup prototype kernels for degenerate inputs

pmeier:
Follow-up to #6542. Resolves

- https://github.com/pytorch/vision/pull/6542#discussion_r965006822
- https://github.com/pytorch/vision/pull/6542#pullrequestreview-1099421668
  Although cloning here is a `O(1)` operation since there is no data to copy, it also does nothing useful. Whether or not we clone or not, the `Tensor.data_ptr()` stays the same and we only create another Python object.
- https://github.com/pytorch/vision/pull/6542#pullrequestreview-1099391108

The only thing left are the xfailed reference tests introduced in #6542. As explained there, these are not bugs in the kernel, but rather in the test setup specifically the expected output computation. @vfdev-5 Could you look into these?

### DIFF
--- a/torchvision/prototype/transforms/functional/_geometry.py
+++ b/torchvision/prototype/transforms/functional/_geometry.py
@@ -450,28 +450,34 @@ def rotate_image_tensor(
     fill: Optional[List[float]] = None,
     center: Optional[List[float]] = None,
 ) -> torch.Tensor:
-    if img.numel() == 0:
-        return img
-
     num_channels, height, width = img.shape[-3:]
     extra_dims = img.shape[:-3]
-    img = img.view(-1, num_channels, height, width)
 
     center_f = [0.0, 0.0]
     if center is not None:
         if expand:
             warnings.warn("The provided center argument has no effect on the result if expand is True")
         else:
-            _, height, width = get_dimensions_image_tensor(img)
             # Center values should be in pixel coordinates but translated such that (0, 0) corresponds to image center.
             center_f = [1.0 * (c - s * 0.5) for c, s in zip(center, [width, height])]
 
     # due to current incoherence of rotation angle direction between affine and rotate implementations
     # we need to set -angle.
     matrix = _get_inverse_affine_matrix(center_f, -angle, [0.0, 0.0], 1.0, [0.0, 0.0])
-    output = _FT.rotate(img, matrix, interpolation=interpolation.value, expand=expand, fill=fill)
-    new_height, new_width = output.shape[-2:]
-    return output.view(extra_dims + (num_channels, new_height, new_width))
+
+    if img.numel() > 0:
+        img = _FT.rotate(
+            img.view(-1, num_channels, height, width),
+            matrix,
+            interpolation=interpolation.value,
+            expand=expand,
+            fill=fill,
+        )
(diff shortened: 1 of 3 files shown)


torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -108,17 +108,14 @@ def resize_image_tensor(
...
             size=[new_height, new_width],
             interpolation=interpolation.value,
             antialias=antialias,
         )
-    else:
-        # TODO: the cloning is probably unnecessary. Review this together with the other perf candidates
-        resized_image = image.clone()
 
-    return resized_image.view(extra_dims + (num_channels, new_height, new_width))
+    return image.view(extra_dims + (num_channels, new_height, new_width))
```
vfdev-5, line 118: We do not care if it is an inplace op for empty tensor, right ?

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -450,6 +450,9 @@ def rotate_image_tensor(
     fill: Optional[List[float]] = None,
     center: Optional[List[float]] = None,
 ) -> torch.Tensor:
+    if img.numel() == 0:
+        return img
```
vfdev-5, line 454: If expand is True, output can have larger size than the input. I mean output shape is changed, so probably we should do something like in bboxes: https://github.com/pytorch/vision/pull/6544/files#diff-8d715258c77dea35427ef8b4605c5dca131717b7e2f6ca1ee3afa99392813105R322

torchvision/prototype/transforms/functional/_geometry.py
pmeier: Depending on your definition of "inplace" I don't think this is an inplace op. For me "inplace" means, the returned tensor has the same `.data_ptr()` but it's values have changed. Since a degenerate input has no data to begin with, the data cannot be changed in any way. Thus, I think there are no inplace operation for degenerate tensors in general. What did you have in mind?

torchvision/prototype/transforms/functional/_geometry.py
pmeier: Good catch!

torchvision/prototype/transforms/functional/_geometry.py
vfdev-5: Oh, sorry, I was thinking view is an inplace op, my thoughts was that `x.view(n, c, new_h, new_w)` will also update x.

pmeier:
Test failure is unrelated:

```
____________ test_eager_vs_scripted[gaussian_blur_image_tensor-50] _____________
  File "/home/runner/work/vision/vision/test/test_prototype_transforms_functional.py", line 751, in test_eager_vs_scripted
    msg=msg,

Mismatched elements: 1 / 1024 (0.1%)
Greatest absolute difference: 1 at index (2, 0, 13, 9)
Greatest relative difference: 0.011627906933426857 at index (2, 0, 13, 9)
```

datumbox (approved):
LGTM, thanks @pmeier. Nice clean ups.

The failing tests seems unrelated to me, I think this is known to be flaky. Might be worth fixing.

PR #6549, opened 2022-09-08
Add stereo preset transforms

TeodorPoncu:
These are the transforms used for Stereo Matching. The resize and rescale is mostly a pure copy paste from the Optical Flow one, with the exception that it performs scaling just on the X flow channel.


references/depth/stereo/transforms.py
```diff
@@ -0,0 +1,572 @@
...
+import numpy as np
+import PIL.Image
+import torch
+import torchvision.transforms as T
+import torchvision.transforms.functional as F
+from torch import Tensor
+
+T_FLOW = Union[Tensor, np.ndarray, None]
+T_MASK = Union[Tensor, np.ndarray, None]
+T_STEREO_TENSOR = Tuple[Tensor, Tensor]
```
TeodorPoncu, line 13: My main curiosity is wether or not we should wrap this in a custom data type that can handle the transform operations in order to skip littering the code with IF-ELSE logic blocks when dealing with datasets that miss that specific piece of information in one of the pairs.

references/depth/stereo/transforms.py
```diff
@@ -0,0 +1,572 @@
...
+        if self.scaling_type == "exponential":
+            scale = 2 ** torch.empty(1, dtype=torch.float32).uniform_(self.min_scale, self.max_scale).item()
+        elif self.scaling_type == "linear":
+            scale = torch.empty(1, dtype=torch.float32).uniform_(self.min_scale, self.max_scale).item()
+
+        scale_x = scale
+        scale_y = scale
+
+        scale_x = max(scale_x, min_scale)
+        scale_y = max(scale_y, min_scale)
```
YosuaMichael, line 473: If `scale_x` always equal to `scale_y`, I think we can just use one variable `scale`

references/depth/stereo/presets.py
```diff
@@ -0,0 +1,101 @@
...
+
+class StereoMatchingEvalPreset(torch.nn.Module):
+    def __init__(self, size=None) -> None:
+        super().__init__()
+
+        transforms = [
+            T.ToTensor(),
+            T.MakeValidDisparityMask(512),  # we keep this transform for API consistency
+            T.ConvertImageDtype(torch.float32),
+            T.Normalize(mean=0.5, std=0.5),
```
YosuaMichael, line 15: I think it is better to put `mean` and `std` as part of the `__init__` param as well.

references/depth/stereo/transforms.py
```diff
@@ -0,0 +1,572 @@
...
+        # avoid occlusion starting in the corners of the right image
+        crop_x, crop_y = (random.randint(crop_w, img_w - crop_w), random.randint(crop_h, img_h - crop_h))
+
+        occlusion_value = img[..., crop_y - crop_h : crop_y + crop_h, crop_x - crop_w : crop_x + crop_w].mean(
+            dim=(-2, -1), keepdim=True
+        )
+
+        return (
+            crop_x - crop_w,
+            crop_y - crop_h,
```
YosuaMichael, line 307: Why do we need to substract with `crop_h`?

references/depth/stereo/transforms.py
```diff
@@ -0,0 +1,572 @@
...
+    # because of the way the random parameters are sampled in both transforms,
+    # which leads to fairly different resuts (and different epe). For more details see
+    # https://github.com/pytorch/vision/pull/5026/files#r762932579
+    def __init__(
+        self,
+        crop_size,
+        min_scale=-0.2,
+        max_scale=0.5,
+        resize_prob=0.8,
+        scaling_type="exponential",
```
YosuaMichael, line 438: For resize, usually we can let the user input the interpolation algorithm (especially for the image), since they might want to use BICUBIC instead of BILINEAR

references/depth/stereo/transforms.py
```diff
@@ -0,0 +1,572 @@
...
+        cropped_masks = ()
+        for mask in masks:
+            if mask is not None:
+                mask = F.crop(mask, y0, x0, self.crop_size[0], self.crop_size[1])
+            cropped_masks += (mask,)
+
+        return ((img_left, img_right), (dsp_left, dsp_right), cropped_masks)
+
+
+def _resize_sparse_flow(flow: Tensor, valid_flow_mask: Tensor, scale_x: float = 1.0, scale_y: float = 0.0):
```
YosuaMichael, line 528: Forgot to add type hint for the output

references/depth/stereo/transforms.py
```diff
@@ -0,0 +1,572 @@
...
+    ii_valid = ii_valid[within_bounds_mask]
+    jj_valid = jj_valid[within_bounds_mask]
+    ii_valid_new = ii_valid_new[within_bounds_mask]
+    jj_valid_new = jj_valid_new[within_bounds_mask]
+
+    valid_flow_new = flow[:, ii_valid, jj_valid]
+    valid_flow_new *= scale_x
+
+    flow_new[:, ii_valid_new, jj_valid_new] = valid_flow_new
+    valid_new[ii_valid_new, jj_valid_new] = 1
```
YosuaMichael, line 558: If I understand correctly, `valid_new` is the `mask`. In this case, the mask will be integer instead of boolean, is this intended?

YosuaMichael (commented):
Hi @TeodorPoncu , thanks for the PR. I have several comments, please let me know what you think.

references/depth/stereo/transforms.py
```diff
@@ -0,0 +1,572 @@
...
+        disparities: T_STEREO_TENSOR,
+        masks: T_STEREO_TENSOR,
+    ) -> Tuple[T_STEREO_TENSOR, Tuple[T_FLOW, T_FLOW], Tuple[T_MASK, T_MASK]]:
+
+        if torch.rand(1) > self.p:
+            return images, disparities, masks
+
+        image_left, image_right = images
+        mask_left, mask_right = masks
+        for _ in range(torch.randint(self.max_erase, size=(self.max_erase,)).item()):
```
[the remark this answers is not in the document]
TeodorPoncu: Yes, that was a mistype on my side. Size should be always equal to 1 similarly to https://github.com/pytorch/vision/blob/8acf1ca2e4f8bd2b94ccc0296a02dcd8ddd7da53/references/optical_flow/transforms.py#L124

references/depth/stereo/transforms.py
```diff
@@ -0,0 +1,572 @@
...
+        masks: Tuple[T_MASK, T_MASK],
+    ) -> Tuple[T_STEREO_TENSOR, Tuple[T_FLOW, T_FLOW], Tuple[T_MASK, T_MASK]]:
+
+        img_left, img_right = images
+        dsp_left, dsp_right = disparities
+        mask_left, mask_right = masks
+
+        # randomly sample scale
+        h, w = img_left.shape[-2:]
+        # Note: in original code, they use + 1 instead of + 8 for sparse datasets (e.g. Kitti)
```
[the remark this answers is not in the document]
TeodorPoncu: Almost 1-to-1 copy with what we have in the optical flow reference script:

https://github.com/pytorch/vision/blob/8acf1ca2e4f8bd2b94ccc0296a02dcd8ddd7da53/references/optical_flow/transforms.py#L157-L183

references/depth/stereo/transforms.py
TeodorPoncu: I am not entirely sure, this is taken from the Optical Flow reference scripts.

references/depth/stereo/transforms.py
```diff
@@ -0,0 +1,572 @@
...
+                mask_left = F.erase(mask_left, x, y, h, w, False, self.inplace)
+            if mask_right is not None:
+                mask_right = F.erase(mask_right, x, y, h, w, False, self.inplace)
+
+        return (image_left, image_right), disparities, (mask_left, mask_right)
+
+    def _get_params(self, img: torch.Tensor) -> Tuple[int, int, int, int, float]:
+        img_h, img_w = img.shape[-2:]
+        crop_h, crop_w = (
+            random.randint(self.min_px_erase, img_h - self.max_px_erase),
```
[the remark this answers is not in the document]
TeodorPoncu: The reasoning is such that it was consistent with the original way it was implemented in CREStereo: https://github.com/megvii-research/CREStereo/blob/ad3a1613bdedd88b93247e5f002cb7c80799762d/dataset.py#L149-L158

references/depth/stereo/transforms.py
YosuaMichael: Just to make sure we have a consistent type on mask, I think we should use boolean here as well. The reason is because in training we may do this resize then do random erase and in random erase we set the mask into boolean False.

TeodorPoncu: There's a conversion to bool() in the return

references/depth/stereo/transforms.py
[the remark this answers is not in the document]
TeodorPoncu: Going over your comments and the large number of parameters we provide in the transforms I believe it's just best to align the two (i.e. have the same cropping behaviour with no weird hacky / math tricks)

i.e. always sample anchor between 0 and size - crop_size and make a crop of crop_size.

In order to inform users of the behaviour and that if they want to replicate some specific paper (i.e. CREstereo) I would just add a comment / description in the docs.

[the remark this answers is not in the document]
YosuaMichael: Hi @TeodorPoncu, can you double check whether the upper bound `img_h - self.max_px_erase` is correct here? I think it should be `self.max_px_erase`.

references/depth/stereo/transforms.py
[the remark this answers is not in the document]
YosuaMichael: Hi @TeodorPoncu , I think you forgot to fix the bug so I unresolve this thread for now.

references/depth/stereo/transforms.py
```diff
@@ -0,0 +1,617 @@
...
+
+        return (image_left, image_right), disparities, (mask_left, mask_right)
+
+    def _get_params(self, img: torch.Tensor) -> Tuple[int, int, int, int, float]:
+        img_h, img_w = img.shape[-2:]
+        crop_h, crop_w = (
+            random.randint(self.min_px_erase, self.max_px_erase),
+            random.randint(self.min_px_erase, self.max_px_erase),
+        )
+        crop_x, crop_y = (random.randint(0, img_w - crop_w), random.randint(0, img_h - crop_h))
```
YosuaMichael, line 273: I think there is a bug here, if I understand correctly `crop_x` in here should correspond to `crop_h` instead of `crop_w`. So the correct code should be:
```
crop_x, crop_y = (random.randint(0, img_h - crop_h), random.randint(0, img_w - crop_w))
```

references/depth/stereo/presets.py
```diff
@@ -0,0 +1,101 @@
...
+
+        if do_flip:
+            transforms += [T.RandomHorizontalFlip()]
+
+        transforms += [
+            # occlusion after flip, otherwise we're occluding the reference image
+            T.RandomOcclusion(p=occlusion_prob, min_px=occlusion_min_px, max_px=occlusion_max_px),
+            T.RandomErase(p=erase_prob, min_px=erase_min_px, max_px=erase_max_px, num_repeats=erase_num_repeats),
+            T.Normalize(mean=0.5, std=0.5),
+            T.MakeValidDisparityMask(max_disparity),
```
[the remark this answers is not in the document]
YosuaMichael: @TeodorPoncu just wondering on this. Is having different order of transform between eval and train is okay?

references/depth/stereo/presets.py
```diff
@@ -0,0 +1,111 @@
+from typing import Tuple, Union
+
+import torch
+import transforms as T
+
+
+class StereoMatchingEvalPreset(torch.nn.Module):
+    def __init__(
+        self, mean: float = 0.5, std: float = 0.5, resize_size=None, interpolation_type: str = "bilinear"
```
YosuaMichael, line 9: Please add type hint for resize_size (I think it should be `Optional[Tuple[int, int]]` in this case)

references/depth/stereo/presets.py
```diff
@@ -0,0 +1,111 @@
...
+        crop_size: Tuple[int, int],
+        resize_prob: float = 1.0,
+        scaling_type: str = "exponential",
+        scale_range: Tuple[float, float] = (-0.2, 0.5),
+        scale_interpolation_type: str = "bilinear",
+        # normalization params:
+        mean: float = 0.5,
+        std: float = 0.5,
+        # processing device
+        gpu_transforms=False,
```
YosuaMichael, line 44: Add type hint for gpu_transforms

references/depth/stereo/presets.py
```diff
@@ -0,0 +1,111 @@
...
+        spatial_shift_interpolation_type: str = "bilinear",
+        # AssymetricColorJitter
+        gamma_range: Tuple[float, float] = (0.8, 1.2),
+        brightness: Union[int, Tuple[int, int]] = (0.8, 1.2),
+        contrast: Union[int, Tuple[int, int]] = (0.8, 1.2),
+        saturation: Union[int, Tuple[int, int]] = 0.0,
+        hue: Union[int, Tuple[int, int]] = 0.0,
+        asymmetric_jitter_prob: float = 1.0,
+        # RandomHorizontalFlip
+        horizontal_flip_prob=0.5,
```
YosuaMichael, line 60: Add type hint for horizontal_flip_prob

YosuaMichael (commented):
Hi @TeodorPoncu , I have few comments

references/depth/stereo/presets.py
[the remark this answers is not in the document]
TeodorPoncu: There is no effect on the values produced by masking procedure (since the disparity is rescaled as well, not only the image). Slight rotations and shifts are added only to the original images and not to the disparity maps or masks, therefore they remain unscathed throughout the procedure.

Since this thresholded valid mask is computed from the resized bilinear interpolation of the disparity values that are converted to floating point values, an original pixel  if beneath `max_disparity * rescale_scale` will still lie end up being masked.

In terms of definitions, having the disparity at the end is more of a "I want the network to be able to output only disparities bellow a certain threshold".

For testing it was "I want to evaluate my raw-input disparities only up to some threshold that I know".

I can most definitely change things up so that they are consistent across both transforms

torchvision/transforms/_presets.py
```diff
@@ -206,3 +207,51 @@ def describe(self) -> str:
...
+    def __init__(
+        self,
+        *,
+        resize_size: Optional[int],
+        mean: Tuple[float, ...] = (0.5, 0.5, 0.5),
+        std: Tuple[float, ...] = (0.5, 0.5, 0.5),
+        interpolation: InterpolationMode = InterpolationMode.BILINEAR,
+    ) -> None:
+        super().__init__()
+        self.resize_size = [resize_size] if resize_size is not None else None
```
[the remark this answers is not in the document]
TeodorPoncu: `VideoClassification` seems to be the the only doing `Tuple[int, int]`. It also makes more sense to me to individually rescale each dimension, just didn't know which is the preferred way of doing this. Semantic segmentation which is also a dense prediction task has `Optional[int]` as well.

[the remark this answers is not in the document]
YosuaMichael: I think as long we are consistent between the type in `transforms/_presets.py` and `reference/transforms.py` then we are fine.

YosuaMichael (commented):
Thanks for the update @TeodorPoncu ! I think overall looks good already, just one last bit about the `MakeValidDisparityMask`

references/depth/stereo/transforms.py
```diff
@@ -61,6 +61,22 @@ def forward(self, images, disparities, masks):
...
+    def __init__(self) -> None:
+        super().__init__()
+
+    def forward(
+        self,
+        images: T_STEREO_TENSOR,
+        disparities: Tuple[T_FLOW, T_FLOW],
+        masks: Tuple[T_MASK, T_MASK],
+    ) -> Tuple[T_STEREO_TENSOR, Tuple[T_FLOW, T_FLOW], Tuple[T_MASK, T_MASK]]:
+        img_left = F.to_grayscale(images[0], num_output_channels=3)
```
YosuaMichael, line 74: Question, just wondering why we use `num_output_channels=3` ? Is it because we assume the model only accept 3-channel image?

references/depth/stereo/transforms.py
```diff
@@ -0,0 +1,627 @@
...
+    h, w = flow.shape[-2:]
+
+    h_new = int(round(h * scale_y))
+    w_new = int(round(w * scale_x))
+    flow_new = torch.zeros(size=[1, h_new, w_new], dtype=flow.dtype)
+    valid_new = torch.zeros(size=[h_new, w_new], dtype=valid_flow_mask.dtype)
+
+    jj, ii = torch.meshgrid(torch.arange(w), torch.arange(h), indexing="xy")
+
+    ii_valid, jj_valid = ii[valid_flow_mask], jj[valid_flow_mask]
```
YosuaMichael, line 597: [To discuss] I think this resize flow is meant for flow where the `valid_flow_mask` is sparse. Therefore this computation is efficient (only consider `ii[valid_flow_mask]`).

However, I think for most stereo matching problem this assumption does not hold. I think using same `Resize` as inference for the `flow / disparity` should be okay. Could you help me to confirm this?

YosuaMichael (commented):
Hi @TeodorPoncu , the code looks good. I just have few things I want to discuss or confirm with you for the `MakeValidDisparityMask` and `_resize_sparse_flow`.

references/depth/stereo/presets.py
```diff
@@ -12,20 +12,36 @@ def __init__(
         resize_size: Optional[Tuple[int, int]] = None,
         max_disparity: Optional[float] = 512,
```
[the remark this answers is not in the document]
TeodorPoncu: I think it would be best to make a full valid mask when receiving a `None`. Reason for this is that when merging a dataset that has masks and one that doesn't have masks then the dataloader will fail. My proposal for the `None` case would be to return a `torch.ones_like(img[0])`. 

Also this would have consistent mathematical behaviour with an inexistent masks since we'll be evaluating the loss on all pixels.

references/depth/stereo/transforms.py
TeodorPoncu: Yes, it's mainly for that reason and to not add additional boiler plate code in other sources. This way someone can easily finetune a model that was trained on RGB to Grayscale or Vice-Versa since the network deals with the exact same type of input.

references/depth/stereo/transforms.py
TeodorPoncu: The assumption holds mostly for training datasets, since you cannot know if the sparse region is due to an occlusion or other reasons (such as KITTI Lidar). Should one be training an invalid / occlusion task alongside the OF / SM tasks, then you could technically treat all flows as continuous flows irrespective of occlusions / sparse regions. Using nearest neighbour interpolation might alter the sparsity mask in a way that is not correlated with the real world model. 

[In this paper](https://arxiv.org/pdf/2203.15174.pdf) flows are refined using an occlusion aware cost volume, instead of a simple cost volume.

torchvision/transforms/_presets.py
```diff
@@ -16,6 +16,7 @@
     "VideoClassification",
```
YosuaMichael, line 16: @TeodorPoncu Sorry I was wrong before about this file, I think this should be put inside prototype area, i.e. `torchvision/prototype/transforms/_presets.py`

references/depth/stereo/transforms.py
```diff
@@ -0,0 +1,633 @@
...
+        disparities: Tuple[T_FLOW, T_FLOW],
+        masks: Tuple[T_MASK, T_MASK],
+    ) -> Tuple[T_STEREO_TENSOR, Tuple[T_FLOW, T_FLOW], Tuple[T_MASK, T_MASK]]:
+        valid_masks = tuple(
+            torch.ones(images[idx].shape[-2:], dtype=torch.bool, device=images[idx].device) if mask is None else mask
+            for idx, mask in enumerate(masks)
+        )
+
+        valid_masks = tuple(
+            torch.logical_and(mask, disparity > 0).squeeze(0) if disparity is not None else mask
```
YosuaMichael, line 97: Question, is `disparity > 0` is something that always be masked on all datasets?

references/depth/stereo/transforms.py
```diff
@@ -0,0 +1,633 @@
...
+        if masks[0] is not None:
+            if masks[0].shape != (h, w):
+                raise ValueError(f"masks[0].shape should be ({h}, {w}) instead of {masks[0].shape}")
+            if masks[0].dtype != torch.bool:
+                raise TypeError(f"masks[0] should be of dtype torch.bool instead of {masks[0].dtype}")
+
+        return images, disparities, masks
+
+
+class ConverToGrayscale(torch.nn.Module):
```
YosuaMichael, line 64: Typo:
`ConverToGrayscale` -> `ConvertToGrayscale` (missing one `t`)

references/depth/stereo/presets.py
```diff
@@ -0,0 +1,144 @@
...
+import transforms as T
+
+
+class StereoMatchingEvalPreset(torch.nn.Module):
+    def __init__(
+        self,
+        mean: float = 0.5,
+        std: float = 0.5,
+        resize_size: Optional[Tuple[int, int]] = None,
+        max_disparity: Optional[float] = 512,
```
YosuaMichael, line 13: I think for inference, it might be better to have `None` as default value for max_disparity (so we dont assume on the image's zoom level). What do you think?

YosuaMichael (commented):
Hey @TeodorPoncu , sorry I was wrong about the `transforms/_presets.py` before. It should be inside prototype area, hence the right file to put it in is in `torchvision/prototype/transforms/_presets.py`.

Other than this, it looks good , I just have a few question and NITs.

references/depth/stereo/presets.py
```diff
@@ -0,0 +1,144 @@
...
+import torch
+import transforms as T
+
+
+class StereoMatchingEvalPreset(torch.nn.Module):
+    def __init__(
+        self,
+        mean: float = 0.5,
+        std: float = 0.5,
+        resize_size: Optional[Tuple[int, int]] = None,
```
YosuaMichael, line 12: NIT, lets use `Tuple[int, ...]` like the others that has changed

references/depth/stereo/transforms.py
```diff
@@ -0,0 +1,633 @@
...
+        return mode in InterpolationStrategy._valid_modes
+
+    @property
+    def valid_modes() -> List[str]:
+        return InterpolationStrategy._valid_modes
+
+
+class ValidateModelInput(torch.nn.Module):
+    # Pass-through transform that checks the shape and dtypes to make sure the model gets what it expects
+    def forward(self, images, disparities, masks):
```
YosuaMichael, line 49: NIT, add type hint

references/depth/stereo/transforms.py
```diff
@@ -0,0 +1,633 @@
...
+        masks: Tuple[T_MASK, T_MASK],
+    ) -> Tuple[T_STEREO_TENSOR, Tuple[T_FLOW, T_FLOW], Tuple[T_MASK, T_MASK]]:
+        dev_images = tuple(image.cuda() for image in images)
+        dev_disparities = tuple(map(lambda x: x.cuda() if x is not None else None, disparities))
+        dev_masks = tuple(map(lambda x: x.cuda() if x is not None else None, masks))
+        return dev_images, dev_disparities, dev_masks
+
+
+class ConvertImageDtype(torch.nn.Module):
+    def __init__(self, dtype):
```
YosuaMichael, line 127: NIT, add type hint

references/depth/stereo/transforms.py
```diff
@@ -0,0 +1,633 @@
...
+        for idx in range(2):
+            disparity_tensors += (torch.from_numpy(disparities[idx]),) if disparities[idx] is not None else (None,)
+            mask_tensors += (torch.from_numpy(masks[idx]),) if masks[idx] is not None else (None,)
+
+        return (img_left, img_right), disparity_tensors, mask_tensors
+
+
+class AsymmetricColorJitter(T.ColorJitter):
+    # p determines the probability of doing asymmetric vs symmetric color jittering
+    def __init__(self, brightness=0, contrast=0, saturation=0, hue=0, p=0.2):
```
YosuaMichael, line 194: NIT, add type hint

references/depth/stereo/transforms.py
```diff
@@ -0,0 +1,633 @@
...
+
+        return (img_left, img_right), disparities, masks
+
+
+class RandomErase(torch.nn.Module):
+    # Produces multiple symetric random erasures
+    # these can be viewed as occlusions present in both camera views.
+    # Similarly to Optical Flow occlusion prediction tasks, we mask these pixels in the disparity map
+    def __init__(
+        self, p: float = 0.5, erase_px_range: Tuple[int, int] = (50, 100), value=0, inplace=False, max_erase=2
```
YosuaMichael, line 252: NIT, add type hint for value, inplace, max_erase

references/depth/stereo/transforms.py
```diff
@@ -0,0 +1,633 @@
...
+        crop_x, crop_y = (random.randint(0, img_w - crop_w), random.randint(0, img_h - crop_h))
+
+        return crop_y, crop_x, crop_h, crop_w, self.value
+
+
+class RandomOcclusion(torch.nn.Module):
+    # This adds an occlusion in the right image
+    # the occluded patch works as a patch erase where the erase value is the mean
+    # of the pixels from the selected zone
+    def __init__(self, p=0.5, occlusion_px_range: Tuple[int, int] = (50, 100), inplace=False):
```
YosuaMichael, line 310: NIT, add type hint for p, inplace

references/depth/stereo/transforms.py
```diff
@@ -0,0 +1,633 @@
...
+    # interpolation artifact at the borders of the output.
+    #
+    # The reason we don't rely on RandomResizedCrop is because of a significant
+    # difference in the parametrization of both transforms, in particular,
+    # because of the way the random parameters are sampled in both transforms,
+    # which leads to fairly different resuts (and different epe). For more details see
+    # https://github.com/pytorch/vision/pull/5026/files#r762932579
+    def __init__(
+        self,
+        crop_size,
```
YosuaMichael, line 485: NIT, add type hint for crop_size, rescale_prob, scaling_type, interpolation_type

references/depth/stereo/transforms.py
```diff
@@ -0,0 +1,633 @@
...
+    valid_flow_new *= scale_x
+
+    flow_new[:, ii_valid_new, jj_valid_new] = valid_flow_new
+    valid_new[ii_valid_new, jj_valid_new] = valid_flow_mask[ii_valid, jj_valid]
+
+    return flow_new, valid_new.bool()
+
+
+class Compose(torch.nn.Module):
+    def __init__(self, transforms):
```
YosuaMichael, line 625: NIT, add type hint

references/depth/stereo/transforms.py
```diff
@@ -0,0 +1,633 @@
...
+        super().__init__()
+        self.transforms = transforms
+
+    @torch.inference_mode()
+    def forward(self, images, disparities, masks):
```
YosuaMichael, line 630: NIT, add type hint

YosuaMichael (approved):
Hi @TeodorPoncu , I think this will be the last review iteration for me. Few nits on the typing.
I will approve first so it is not blocking. Thanks again for the PR!

PR #6551, opened 2022-09-09
Fix swapped width and height in DefaultBoxGenerator

### LINKED ISSUE #6550: Swapped width and height in DefaultBoxGenerator
In the DefaultBoxGenerator module, the image sizes in the image_size argument are given as (height, width) by SSD model but the code assumes (width, height).

https://github.com/pytorch/vision/blob/15a9a93b9e0e6173b936691a2cddced0ecfd271f/torchvision/models/detection/anchor_utils.py#L216-L217

I think this has gone unnoticed because the images are usually resized into squares.

cc @datumbox

nlgranger:
Fix #6550 

Changes image size convention of DefaultBoxGenerator to (h, w). I don't think this is used by models other than SSD which assumes this convention.

### DIFF
--- a/torchvision/models/detection/anchor_utils.py
+++ b/torchvision/models/detection/anchor_utils.py
@@ -213,8 +213,8 @@ class DefaultBoxGenerator(nn.Module):
         for k, f_k in enumerate(grid_sizes):
             # Now add the default boxes for each width-height pair
             if self.steps is not None:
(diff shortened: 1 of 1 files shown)


datumbox (approved):
Thanks for the fix, LGTM.

You are right this went unnoticed because SSD usually operates on H=W.

PR #6552, opened 2022-09-09
extract common utils for prototype transform tests

pmeier:
Per title. The new module `prototype_common_utils.py` now contains all functions for feature creation, comparison. Plus, I've unified `ArgsKwargs` from `test_prototype_transforms_consistency.py` with `SampleInput` from `test_prototype_transforms_functionals.py`.

I've done this in preparation for #5879, but this is a good change whether or not we go for the other PR in the end. With this we have all common utilities in one place rather than scattered and accidentally reinventing the wheel.


vfdev-5 (commented):
Looks ok to me, thanks @pmeier

vfdev-5 (approved):

PR #6554, opened 2022-09-09
Add stereo matching losses

TeodorPoncu:
Some of the Stereo Matching criterions that can be added to a training loop for regularisation purposes. Contains two candidates for more generic usage: respectively SSIM and PSNR losses


torchvision/models/optical_flow/_utils.py
```diff
@@ -19,8 +19,8 @@ def grid_sample(img: Tensor, absolute_grid: Tensor, mode: str = "bilinear", alig
     return F.grid_sample(img, normalized_grid, mode=mode, align_corners=align_corners)
 
 
-def make_coords_grid(batch_size: int, h: int, w: int):
-    coords = torch.meshgrid(torch.arange(h), torch.arange(w), indexing="ij")
+def make_coords_grid(batch_size: int, h: int, w: int, device: torch.device = torch.device("cpu")):
```
jdsgomes, line 22: I would prefer to just pass the device string here and instantiate the torch.device inside the method. Reason for this is that I am not sure if torch.device can return different values depending on the context.

references/depth/stereo/utils/losses.py
```diff
@@ -0,0 +1,470 @@
...
+        # reset the cached scale factor
+        self.weights = None
+
+
+def _ssim_loss_fn(
+    source: Tensor,
+    reference: Tensor,
+    kernel: Tensor,
+    eps: float = 1e-8,
+    C1: float = 0.01**2,
```
jdsgomes, line 99: since this is an argument I think it should be lowercase

jdsgomes (commented):
Thanks for this, just left a couple of comments

references/depth/stereo/utils/losses.py
```diff
@@ -0,0 +1,470 @@
...
+        flow_preds = (
+            F.interpolate(
+                flow_preds.view(N * B, C, H, W), scale_factor=resize_factor, mode=rescale_mode, align_corners=True
+            )
+        ) * rescale_factor
+        flow_preds = torch.stack(torch.chunk(flow_preds, N, dim=0), dim=0)
+
+    # force the next prediction to be similar to the previous prediction
+    abs_diff = (flow_preds[1:] - flow_preds[:-1]).square()
+    abs_diff = abs_diff.mean(axis=(1, 2, 3, 4))
```
YosuaMichael, line 312: Question, since we do `.square()` previously, shouldn't we take the `root` after the mean?

references/depth/stereo/utils/losses.py
```diff
@@ -0,0 +1,470 @@
...
+        )
+
+
+def _psnr_loss_fn(source: torch.Tensor, target: torch.Tensor, max_val: float) -> torch.Tensor:
+    torch._assert(
+        source.shape == target.shape,
+        "psnr_loss: source and target must have the same shape, but got {} and {}".format(source.shape, target.shape),
+    )
+
+    return 10 * torch.log10(max_val**2 / ((source - target).abs().pow(2).mean(axis=(-3, -2, -1))))
```
YosuaMichael, line 360: Just curious, why do we need `.abs()` before `.pow(2)` ? I assume `.pow(2)` will be always positive even without `.abs()`.

YosuaMichael (commented):
Hi @TeodorPoncu , thanks for the PR! In general, could you add reference (url to paper or other existing implementation) for the losses?
Also, I have few comments and questions.

references/depth/stereo/utils/losses.py
```diff
@@ -0,0 +1,470 @@
...
+        source.shape == reference.shape,
+        "SSIM: `source` and `reference` must have the same shape, but got {} and {}".format(
+            source.shape, reference.shape
+        ),
+    )
+
+    B, C, H, W = source.shape
+    kernel = kernel[None, None, ...].repeat(C, 1, 1, 1)
+    if use_padding:
+        pad_size = kernel.shape[2] // 2
```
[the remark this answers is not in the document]
TeodorPoncu: Yes, this is a symmetric gaussian kernel that's being used in SSIM.

references/depth/stereo/utils/losses.py
```diff
@@ -0,0 +1,470 @@
...
+            reference,
+            kernel=self.gaussian_kernel,
+            C1=self.C1,
+            C2=self.C2,
+            use_padding=self.use_padding,
+            eps=self.eps,
+        )
+
+
+def _smothness_loss_fn(img_gx: Tensor, img_gy: Tensor, depth_gx: Tensor, depth_gy: Tensor):
```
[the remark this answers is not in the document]
TeodorPoncu: It is the same formula that's being used for monocular depth smoothness. Since disparity is inverse depth, I assume in some contexts they could be interchangeable. Not sure what the most ideal naming scheme would be.

references/depth/stereo/utils/losses.py
TeodorPoncu: Not taking the root has a more severe impact on any interpolation artefacts / sudden changes the model might generate from a recursive iteration to another

references/depth/stereo/utils/losses.py
```diff
@@ -0,0 +1,470 @@
...
+    """Loss function defined over sequence of flow predictions"""
+    torch._assert(
+        gamma < 1,
+        "sequence_loss: `gamma` must be lower than 1, but got {}".format(gamma),
+    )
+
+    if exclude_large:
+        # exclude invalid pixels and extremely large diplacements
+        flow_norm = torch.sum(flow_gt**2, dim=1).sqrt()
+        valid_flow_mask = valid_flow_mask & (flow_norm < max_flow)
```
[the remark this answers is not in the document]
TeodorPoncu: `max_flow` would be enough if we assume users to use our transforms. If that assumption is correct than that term is not needed.

references/depth/stereo/utils/losses.py
```diff
@@ -0,0 +1,470 @@
...
+def _sequence_loss_fn(
+    flow_preds: List[Tensor],
+    flow_gt: Tensor,
+    valid_flow_mask: Tensor,
+    gamma: Tensor,
+    max_flow: int = 256,
+    exclude_large: bool = False,
+    weights: Optional[Tensor] = None,
+):
+    """Loss function defined over sequence of flow predictions"""
```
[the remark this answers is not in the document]
TeodorPoncu: This is actually the loss we're using in the reference Optical Flow as well.

references/depth/stereo/utils/losses.py
YosuaMichael: @TeodorPoncu So if I understand correctly, in this formula we intentionally not taking the root so it penalized more on interpolation artifacts?

references/depth/stereo/utils/losses.py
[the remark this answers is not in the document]
YosuaMichael: I think if it can be used for both `depth` and `disparity`, we might call it `val_gx` and `val_gy`, what do you think?

references/depth/stereo/utils/losses.py
TeodorPoncu: Yes, otherwise we'd be penalising very similarly small refinements / changes (which are to be expected from the network) as much as we'd penalising the drastic shifts in estimating how far / close the background might be or any other surface sudden changes

references/depth/stereo/utils/losses.py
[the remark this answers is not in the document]
TeodorPoncu: Unfortunately, because some datasets have `None` returns for the disparity mask, whilst others don't. We need a transform that fills in that `None` with something. Even if `max_disparity` is not needed, we still have to impute somehow the `None`.

references/depth/stereo/utils/losses.py
```diff
@@ -0,0 +1,477 @@
...
+    grid = (x**2 + y**2) / (2 * sigma**2)
+    kernel = torch.exp(-grid)
+    kernel = kernel / kernel.sum()
+    return kernel
+
+
+def _sequence_loss_fn(
+    flow_preds: List[Tensor],
+    flow_gt: Tensor,
+    valid_flow_mask: Tensor,
```
YosuaMichael, line 26: Just to follow up on my comment on https://github.com/pytorch/vision/pull/6549.
We may want the `valid_flow_mask` to be `Optional` here. But lets discuss this first

Note: I notice we already make `valid_mask` to be optional in `Photometric loss`

references/depth/stereo/utils/losses.py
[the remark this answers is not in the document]
YosuaMichael: Currently we allow `max_disparity` to be optional, and if it is `None` it will cause the problem you mention here. I think one possible solution is to let `valid_flow_mask` to be `Optional` as well here, and we treat `None` as if it is `True` everywhere. As indicated on previous comment, lets discuss on this particular topic first.

references/depth/stereo/utils/losses.py
```diff
@@ -0,0 +1,477 @@
...
+            reference,
+            kernel=self.gaussian_kernel,
+            c1=self.c1,
+            c2=self.c2,
+            use_padding=self.use_padding,
+            eps=self.eps,
+        )
+
+
+def _smothness_loss_fn(img_gx: Tensor, img_gy: Tensor, val_gx: Tensor, val_gy: Tensor):
```
YosuaMichael, line 201: NIT, just small typo I think (missing one `o`). 
```
smothness -> smoothness
```

references/depth/stereo/utils/losses.py
```diff
@@ -0,0 +1,477 @@
...
+        source: Tensor,
+        reference: Tensor,
+        flow_pred: Tensor,
+        valid_mask: Optional[Tensor] = None,
+    ):
+        """
+        Args:
+            source: tensor of shape (B, C, H, W)
+            reference: tensor of shape (B, C, H, W)
+            flow_pred: tensor of shape (B, 2, H, W)
```
YosuaMichael, line 428: Question, Is it possible to make this loss more generic so it can be used for 1-channel disparity, i.e flow_pred shape (B, 1, H, W) ?

Not sure if this is correct, but here is my idea for this:
```
Current:
  max_H = int(H * self._max_displacement_ratio)
  max_W = int(W * self._max_displacement_ratio)
New:
  if flow_pred.shape[1] == 1:
    max_displacements = [int(W * self._max_displacement_ratio)]
  elif flow_pred.shape[2] == 2:
    max_displacements = [int(H * self._max_displacement_ratio), int(W * self._max_displacement_ratio)]
  else:
    raise ValueError(f"flow_pred.shape[1] should be either 1 or 2, but got {flow_pred.shape[1]}")
```

```
Current:
 max_flow_mask = torch.logical_and(
     flow_pred[:, 0, :, :] < max_W,
     flow_pred[:, 1, :, :] < max_H,
  )

New:
  max_flow_mask = torch.logical_and(*[
      flow_pred[:, i, :, :] < max_displacements[i] 
      for i in range flow_pred.shape[1] 
  ])
```

references/depth/stereo/utils/losses.py
```diff
@@ -0,0 +1,503 @@
...
+        Args:
+            gamma: value for the exponential weighting of the loss across frames
+            max_flow: maximum flow value to exclude
+            exclude_large_flows: whether to exclude large flows
+        """
+
+        super().__init__()
+        self.max_flow = max_flow
+        self.excluding_large = exclude_large_flows
+        self.register_buffer("gamma", torch.tensor([gamma]))
```
YosuaMichael, line 78: Just curious, why do we need to do `register_buffer` compared to `self.gamma = torch.tensor([gamma])` ?

references/depth/stereo/utils/losses.py
```diff
@@ -0,0 +1,503 @@
...
+            use_padding: whether to pad the input tensor such that we have a score for each pixel
+        """
+        super().__init__()
+
+        self.kernel_size = kernel_size
+        self.max_val = max_val
+        self.sigma = sigma
+
+        gaussian_kernel = make_gaussian_kernel(kernel_size, sigma)
+        self.register_buffer("gaussian_kernel", gaussian_kernel)
```
YosuaMichael, line 180: Similar question, why do we need `register_buffer` here?

YosuaMichael (approved):
Hi @TeodorPoncu , overall looks good to me! Just have a question regarding register_buffer

references/depth/stereo/utils/losses.py
TeodorPoncu: This allows the loss to be instantiated as a module and be subsequently moved to a device. Registering it avoids having to reinstantiate the kernel / value / tensor at every forward pass when we do the "forward" through the loss.

YosuaMichael (approved):
LGTM!

jdsgomes (approved):
LGTM - thanks!

PR #6555, opened 2022-09-09
Added Cosine Annealing Warmup Restarts With Decay optimiser

TeodorPoncu:
This is an optimiser that is a variant of the Cosine Annealing Restart scheduler or the OneCycleLR optimiser. The added optimiser can decrease over time the maximum learning rate it warms up to and can increase the length of the decay cycle over time. If we set `T_warmup=0, gamma=1.0` we recover the base  [CosineAnnealingWarmRestarts](https://pytorch.org/docs/stable/generated/torch.optim.lr_scheduler.CosineAnnealingWarmRestarts.html).

This was based on the observation that CREStereo generally reacts better to a constantly changing learning rate as opposed to the default scheduler proposed in the paper. Bellow is an example of a schedule generated with a value of `gamma < 1.0` and a value of `T_warmup > 0`

301481165_797180951324735_4562113682048515098_n

### DIFF
--- a/references/depth/stereo/utils/optimizer.py
+++ b/references/depth/stereo/utils/optimizer.py
@@ -1,0 +1,114 @@
+import math
+
+import torch
+
+
+class ConsinAnnealingWarmupRestartsWithDecay(torch.optim.lr_scheduler._LRScheduler):
+    def __init__(
+        self,
+        optimizer: torch.optim.Optimizer,
+        T_0: int,
+        T_mult: float = 1.0,
+        eta_min: float = 0.0001,
+        T_warmup: int = 0,
+        gamma: float = 1.0,
+        last_epoch: int = -1,
+    ):
+        """
+
+        Args:
+            optimizer: the base optimizer
+            T_0: the base number of iterations for a cycle
+            T_mult: the scaling factor for how much a cycle lasts
+            eta_min: the minimum learning rate
+            T_warmup: the number of linear warmup iterations
+            gamma: the exponential decay factor for the maximum learning rate
+            last_epoch: the last epoch
+
+        """
+
+        if not isinstance(optimizer, torch.optim.Optimizer):
+            raise TypeError("{} is not an Optimizer".format(type(optimizer).__name__))
+        self.optimizer = optimizer
+
+        self.T_0 = T_0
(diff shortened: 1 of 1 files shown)


datumbox:
> This was based on the observation that CREStereo generally reacts better to a constantly changing learning rate as opposed to the default scheduler proposed in the paper. 

To avoid adding unnecessary complexity on the references, it's worth starting from the paper's proposal first. If you can't reproduce the results and you need an extra boost, we could consider adding new optimizers. wdyt?

TeodorPoncu:
There are multiple people [stating](https://github.com/megvii-research/CREStereo/issues/23) on the original repo that they have trouble reproducing the performance claimed by the authors with the training code provided by them.

This is something I have noticed as well during my training runs. It can however be ameliorated with much simpler scheduling. We can skip this PR all-together, I do not have any strong opinions on it.

datumbox:
@TeodorPoncu Thanks for the references! It's great that you tried reproducing it already.

From what I understand the `nan` value appears on the early epochs much before any restart. If that's the case, then the modified optimizer might have less effect on nans and it might be beneficial to do warmup at the beginning using one of the standard schedulers such as `LinearLR` or `ConstantLR`. If that doesn't work out then we should consider merging this PR.

In general it's not a big problem adding primitives in references, as long as they are necessary. For tasks such as Classification, we are more forgiving because some techniques are popular across multiple models and because we constantly do work on them. For Depth Perception that is still prototype, it would be great if we could keep things simple as long as we can reproduce the accuracies and we don't have a good reason to complicate things. So if you really need it, we can definitely add it. If you manage to work around it, that's a plus. :)

TeodorPoncu:
@datumbox I have already managed to work around both the NaN and some of the performance drop by simply adjusting the point at which the learning rate starts decaying, and by replacing the linear decay with a cosine. Therefor, the scheduler is just a minor nice-to-have / bonus.

The benefit of this schedule is in fine-tuning scenarios, solely for convergence time and a very minor performance boost.

datumbox:
@TeodorPoncu shall we close the PR given you managed to train the models and deploy the weights?

TeodorPoncu:
Yes, this could be closed.

PR #6560, opened 2022-09-11
Make get_model_builder public

datumbox:
This PR makes it possible to get a model builder method by its name. 

This is useful because a model might be registered under a different name than the method’s name (see quantization models, for example `quantized_resnet50`) and thus fetching it by indexing the module `__dict__` might not be possible. This PR addresses feedback received internally by [MobileVision](https://www.internalfb.com/diff/D39374209?dst_version_fbid=382397124079881&transaction_fbid=1195981907925443) and will simplify the integration with fastai (see https://github.com/fastai/fastai/issues/3789).

### DIFF
--- a/test/test_extended_models.py
+++ b/test/test_extended_models.py
@@ -30,6 +30,21 @@ def test_get_model(name, model_class):
 
 
 @pytest.mark.parametrize(
+    "name, model_fn",
+    [
+        ("resnet50", models.resnet50),
+        ("retinanet_resnet50_fpn_v2", models.detection.retinanet_resnet50_fpn_v2),
+        ("raft_large", models.optical_flow.raft_large),
+        ("quantized_resnet50", models.quantization.resnet50),
+        ("lraspp_mobilenet_v3_large", models.segmentation.lraspp_mobilenet_v3_large),
+        ("mvit_v1_b", models.video.mvit_v1_b),
+    ],
+)
+def test_get_model_builder(name, model_fn):
+    assert models.get_model_builder(name) == model_fn
+
+
+@pytest.mark.parametrize(
     "name, weight",
     [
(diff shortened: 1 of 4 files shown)


NicolasHug (commented):
Changing `find_model` into `get_model_builder` was originally suggested  in https://github.com/pytorch/vision/pull/6333#discussion_r934223094 and in https://github.com/pytorch/vision/pull/6333#discussion_r934299363 but it wasn't accepted at the time. What is the reason to change it now?

Also considering model buillders are now available, does it still make sense to expose `get_model()`? IIUC `get_model(name, args)` is now just `get_model_builder(name)(args)`, so we would have 2 ways of doing the same thing?

jdsgomes (approved):

datumbox:
@NicolasHug Just saw your message.

> What is the reason to change it now?

It was a private method named similarly to `[find()](https://github.com/pytorch/vision/blob/main/torchvision/prototype/datasets/_api.py#L38)` from datasets. Now that we publicly expose it, we can adopt your original proposal.

> so we would have 2 ways of doing the same thing?

I don't mind particularly offering both and the user feedback so far has been quite positive for the new `get_model()` method.

NicolasHug:
I assume that the user feedback about `get_model()` came before `get_model_builder()` was exposed? It's still worth re-considering IMO. 

> [There should be one-- and preferably only one --obvious way to do it.](https://peps.python.org/pep-0020/)


> @NicolasHug Just saw your message.

I'll try to accept that as true, but in all honesty, this is the exact same sequence of events as they happened in https://github.com/pytorch/vision/pull/6560. It makes me feel uncomfortable and I don't want to re-live that again. I'll disengage from this conversation.

PR #6561, opened 2022-09-12
port RandomResize from segmentation references

pmeier:
https://github.com/pytorch/vision/blob/cac4e228c9ca9e7564cb34406e7ebccfdd736976/references/segmentation/transforms.py#L29-L40

Although the reference transform does not, we could expose a few parameters that could be passed to `resize` like `antialias` and `interpolation`.

### DIFF
--- a/torchvision/prototype/transforms/_geometry.py
+++ b/torchvision/prototype/transforms/_geometry.py
@@ -869,14 +869,22 @@ class FixedSizeCrop(Transform):
 
 
 class RandomResize(Transform):
-    def __init__(self, min_size: int, max_size: int) -> None:
+    def __init__(
+        self,
+        min_size: int,
+        max_size: int,
+        interpolation: InterpolationMode = InterpolationMode.BILINEAR,
+        antialias: Optional[bool] = None,
+    ) -> None:
         super().__init__()
         self.min_size = min_size
         self.max_size = max_size
+        self.interpolation = interpolation
+        self.antialias = antialias
 
     def _get_params(self, sample: Any) -> Dict[str, Any]:
         size = int(torch.randint(self.min_size, self.max_size, ()))
         return dict(size=size)
 
     def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
-        return F.resize(inpt, params["size"])
+        return F.resize(inpt, params["size"], interpolation=self.interpolation, antialias=self.antialias)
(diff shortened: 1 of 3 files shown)


torchvision/prototype/transforms/_geometry.py
```diff
@@ -866,3 +866,20 @@ def forward(self, *inputs: Any) -> Any:
             )
 
         return super().forward(*inputs)
+
+
+class RandomResize(Transform):
+    def __init__(self, min_size: int, max_size: Optional[int] = None):
+        super().__init__()
+        self.min_size = min_size
+        self.max_size = max_size if max_size is not None else min_size
```
pmeier, line 875: Triggering this default actually makes this transform a no-op. In the reference scripts `max_size` is always passed:

https://github.com/pytorch/vision/blob/cac4e228c9ca9e7564cb34406e7ebccfdd736976/references/segmentation/presets.py#L10

https://github.com/pytorch/vision/blob/cac4e228c9ca9e7564cb34406e7ebccfdd736976/references/segmentation/presets.py#L31

(why are we using a random transformation in the evaluation transforms? Doesn't that mean we get different metrics every time we run this?)

We should probably get rid of this behavior, since it makes little sense to use this transform in the current state without passing a `max_size`.

datumbox: I would say it's not no-op but rather falls back to Resize. If I had to guess, the original author didn't want to provide another Resize wrapper to handle Images and Masks at the same time, so they hacked this one. I agree we can use the standard `Resize()` of the new API during inference. So I'm also in favour of getting rid of this behaviour.

datumbox (commented):
Thanks @pmeier. I agree with your comment to remove the old weird behaviour. Could you update the PR accordingly?

pmeier:
Test failure is unrelated:

```
=================================== FAILURES ===================================
____________ test_eager_vs_scripted[gaussian_blur_image_tensor-46] _____________
  File "/home/runner/work/vision/vision/test/test_prototype_transforms_functional.py", line 595, in test_eager_vs_scripted
    msg=msg,

Mismatched elements: 1 / 1116 (0.1%)
Greatest absolute difference: 1 at index (0, 3, 0)
Greatest relative difference: 0.0060240961611270905 at index (0, 3, 0)
```

datumbox (approved):
LGTM, thanks @pmeier! Shall we also add a test?

PR #6562, opened 2022-09-12
[proto] Few minor code improvements

vfdev-5:
Description:
- call torch.log once can save few ms
- all masks should be selected with is_within_crop_area in RandomIoUCrop


torchvision/prototype/transforms/_geometry.py
```diff
@@ -92,14 +92,16 @@ def __init__(
...
+        self._log_ratio = torch.log(torch.tensor(self.ratio))
+
     def _get_params(self, sample: Any) -> Dict[str, Any]:
         # vfdev-5: techically, this op can work on bboxes/segm masks only inputs without image in samples
         # What if we have multiple images/bboxes/masks of different sizes ?
         # TODO: let's support bbox or mask in samples without image
         _, height, width = query_chw(sample)
         area = height * width
 
-        log_ratio = torch.log(torch.tensor(self.ratio))
```
datumbox, line 102: Is it a problem that we did't send this to the right device? Also I wonder if the `log_ratio` needs to be on a tensor given it's provided as parameter on `unitform()`.

datumbox (approved):
LGTM, though I have a question below.

torchvision/prototype/transforms/_geometry.py
datumbox: Sounds good. Might be good testing math as well.

Issue #6564, opened 2022-09-12
Inconsistent use of `handle_legacy_interface()` across model builders


datumbox:
### 🐛 Describe the bug

We got feedback from some of our downstream frameworks (MMLabs, MobileCV, FastAI, Lightning, etc) that they are not yet ready to pin TorchVision to v0.13 or higher. This means that for compatibility reasons, they are forced to continue using the `pretrained=True` idiom. For the majority of the models, that's OK because we use the `handle_legacy_interface()` decorator to set the right weights. Unfortunately not all models support it and thus when they try to initialize the new models they get errors.

For example, the following at MobileCV:
```python
from mobile_cv.model_zoo.models import model_zoo_factory

model_zoo_factory.get_model("swin_t")
```

Raises an exception:
```
TypeError: SwinTransformer.__init__() got an unexpected keyword argument 'pretrained'
```

The use (or lack of use) of the decorator is not consistent. For example in v0.13 we released the `efficientnet_v2_s` which [uses](https://github.com/pytorch/vision/blob/0a919dbb825d3f33eefcaaaafd3a2bbf6ac91fb8/torchvision/models/efficientnet.py#L989-L992) the decorator and the `swin_t` which [doesn't](https://github.com/pytorch/vision/blob/0a919dbb825d3f33eefcaaaafd3a2bbf6ac91fb8/torchvision/models/swin_transformer.py#L519). Similarly `shufflenet_v2_x1_5` uses it but `resnext101_64x4d` doesn't. This lack of consistency across newly introduced models in v0.13 is probably a bug.

Adding it everywhere will ensure the behaviour is aligned across the library and will help the downstream frameworks transition smoother to the new idiom.

### Versions

latest main branch

cc @jdsgomes @YosuaMichael

datumbox:
> Does this mean we will have to support pretrained=True for new models that we'll release in 0.14 as well?

Correct, this would align our approach not only within v0.13 but across all versions. This will also ensure that the downstream libraries can use TorchVision without breakages. I saw you already commented on the PR #6565 that does this.

> We're adding support for pretrained=True on new model only to give more time and flexibility for downstream libs to migrate

Correct. The feedback both by internal and external stakeholders was that they didn't want to pin PyTorch (and as a result TorchVision) to the latest version as of yet. They will eventually but this needs to be coordinated. You can see more details by following the external issues and PRs linked to #6365.

> We're still committed to remove all support for pretrained=True in 0.15. Libraries must have migrated by then.

[...]

PR #6565, opened 2022-09-12
Add missing `handle_legacy_interface()` calls

datumbox:
Fixing #6564

### DIFF
--- a/test/test_extended_models.py
+++ b/test/test_extended_models.py
@@ -3,6 +3,7 @@ import os
 import pytest
 import test_models as TM
 import torch
+import torchvision.models
 from torchvision import models
 from torchvision.models._api import get_model_weights, Weights, WeightsEnum
 from torchvision.models._utils import handle_legacy_interface
@@ -338,13 +339,13 @@ class TestHandleLegacyInterface:
 
     @pytest.mark.parametrize(
         "model_fn",
-        TM.list_model_fns(models)
+        [fn for fn in TM.list_model_fns(models) if fn.__name__ not in {"vit_h_14", "regnet_y_128gf"}]
         + TM.list_model_fns(models.detection)
         + TM.list_model_fns(models.quantization)
         + TM.list_model_fns(models.segmentation)
         + TM.list_model_fns(models.video)
         + TM.list_model_fns(models.optical_flow),
     )
-    def test_all_models(self, model_fn):
+    def test_pretrained_deprecation(self, model_fn):
         with pytest.warns(UserWarning, match="deprecated"):
             model_fn(pretrained=True)
--- a/test/test_extended_models.py
+++ b/test/test_extended_models.py
@@ -335,3 +335,16 @@ class TestHandleLegacyInterface:
 
         with pytest.raises(ValueError, match="weights"):
             builder(pretrained=True, flag=False)
+
+    @pytest.mark.parametrize(
+        "model_fn",
+        TM.list_model_fns(models)
+        + TM.list_model_fns(models.detection)
+        + TM.list_model_fns(models.quantization)
+        + TM.list_model_fns(models.segmentation)
+        + TM.list_model_fns(models.video)
+        + TM.list_model_fns(models.optical_flow),
+    )
+    def test_all_models(self, model_fn):
+        with pytest.warns(UserWarning, match="deprecated"):
+            model_fn(pretrained=True)
--- a/test/test_extended_models.py
+++ b/test/test_extended_models.py
@@ -345,6 +345,7 @@ class TestHandleLegacyInterface:
         + TM.list_model_fns(models.video)
         + TM.list_model_fns(models.optical_flow),
     )
+    @run_if_test_with_extended
     def test_pretrained_deprecation(self, model_fn):
         with pytest.warns(UserWarning, match="deprecated"):
             model_fn(pretrained=True)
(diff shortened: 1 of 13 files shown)


test/test_extended_models.py
```diff
@@ -335,3 +335,16 @@ def builder(*, weights=None, flag):
...
+        TM.list_model_fns(models)
+        + TM.list_model_fns(models.detection)
+        + TM.list_model_fns(models.quantization)
+        + TM.list_model_fns(models.segmentation)
+        + TM.list_model_fns(models.video)
+        + TM.list_model_fns(models.optical_flow),
+    )
+    def test_all_models(self, model_fn):
+        with pytest.warns(UserWarning, match="deprecated"):
+            model_fn(pretrained=True)
```
datumbox, line 350: This will slowdown the tests. I can remove it but it can be used to avoid missing it on the future.

torchvision/models/regnet.py
```diff
@@ -1294,7 +1294,7 @@ def regnet_y_32gf(*, weights: Optional[RegNet_Y_32GF_Weights] = None, progress:
 
 
 @register_model()
-@handle_legacy_interface(weights=("pretrained", None))
+@handle_legacy_interface(weights=("pretrained", RegNet_Y_128GF_Weights.DEFAULT))
```
datumbox, line 1297: There are no `IMAGENET1K_V1` weights for this one. I'm not sure about this change; it was done purely to make the aforementioned test pass.

torchvision/models/vision_transformer.py
```diff
@@ -733,6 +733,7 @@ def vit_l_32(*, weights: Optional[ViT_L_32_Weights] = None, progress: bool = Tru
 
 
 @register_model()
+@handle_legacy_interface(weights=("pretrained", ViT_H_14_Weights.DEFAULT))
```
datumbox, line 736: Same as above. Here we could put `None` but this will lead to a "no weights available" exception.

YosuaMichael (commented):
Hi @datumbox thanks for the PR, overall it looks good. I notice there is a failure on the test, is this related to the changes?

torchvision/models/regnet.py
datumbox: There are no `IMAGENET1K_V1` weights for this one :(

I have the same concerns as you here. @jdsgomes expressed the same sentiment while speaking offline with him. I think I'll revert this to `None`.

test/test_extended_models.py
datumbox: Good question, it goes from [2:40](https://app.circleci.com/pipelines/github/pytorch/vision/20041/workflows/2e3bb4bc-ebe2-4a16-9fbb-72047860ba00/jobs/1628137) to [5:03](https://app.circleci.com/pipelines/github/pytorch/vision/20056/workflows/dc139173-7870-439f-8cb1-c3860e119759/jobs/1629547).

datumbox:
Concerning the [failing](https://app.circleci.com/pipelines/github/pytorch/vision/20056/workflows/dc139173-7870-439f-8cb1-c3860e119759/jobs/1629568) tests, they don't seem related. I'll open a ticket with RelEng in a bit.

NicolasHug:
FYI unless we are planning on a 0.13.2 bugfix release (I'm not sure it will happen), the deprecation message will be inaccurate as it will mention that `pretrained=True` was deprecated in 0.13, but in reality this will only start to to be available (and deprecated) in 0.14.

jdsgomes (approved):
LGTM thanks

test/test_extended_models.py
datumbox: Sounds good, I just added the `@run_if_test_with_extended` decorator to run only on the `unittest_linux_cpu` job.

YosuaMichael (approved):
Looks good!

NicolasHug:
Hi all,

Any feedback / follow-up on https://github.com/pytorch/vision/pull/6565#issuecomment-1243879186

YosuaMichael:
@NicolasHug sorry just noticed your comment, if I understand correctly the multi-weight API is (where we also deprecate `pretrained=True`) is released on v0.13: https://github.com/pytorch/vision/releases. So I think it should the handle_legacy text should be correct from `v0.13` in general.

However I agree that with this PR, some model that just got the decorator `handle_legacy_interface()` should not be deprecated on v0.13.

In my opinion, maybe we should bump the deprecation version for all model using `handle_legacy_interface` to v0.14.

WDYT @NicolasHug @datumbox @jdsgomes ?

NicolasHug:
>  if I understand correctly the multi-weight API is (where we also deprecate pretrained=True) is released on v0.13: https://github.com/pytorch/vision/releases. So I think it should the handle_legacy text should be correct from v0.13 in general.

Yes you understand correctly - the deprecation message is correct for most models, but not for the models tackled by this PR.

As mentioned in https://github.com/pytorch/vision/issues/6564, I think we should have a team discussion about whether it makes sense to postpone the removal of a deprecated API (I added an item for the next team meeting. This somewhat relates to this thread as well https://github.com/pytorch/vision/pull/6258#discussion_r977542950).

As far as the deprecation message for this PR's models though, a simple fix would be to change "deprecated in 0.13" to "deprecated in 0.14" (we can keep the planned removal for 0.15 IMHO, since this is a fairly exceptional case).

YosuaMichael:
@NicolasHug Agree with changing to `deprecated in 0.14` but in my opinion we should move back the planned removal to `0.16` just for consistency. But not really a strong opinion.

PR #6566, opened 2022-09-12
[proto] Added consistency tests for detection transforms

vfdev-5:


test/test_prototype_transforms_consistency.py
```diff
@@ -840,3 +840,681 @@ def test_aa(self, inpt, interpolation):
...
+    def make_datapoints(self, with_mask=True):
+        size = (600, 800)
+
+        pil_image = PIL.Image.new("RGB", size[::-1], 123)
+        target = {
+            "boxes": make_bounding_box(image_size=size, format="XYXY", extra_dims=(22,), dtype=torch.float),
+            "labels": features.Label(torch.randint(0, 81, size=(22,))),
+        }
+        if with_mask:
+            target["masks"] = features.SegmentationMask(torch.randint(0, 2, size=(22, *size), dtype=torch.long))
```
pmeier, line 855: See #6579

```suggestion
            target["masks"] = features.Mask(torch.randint(0, 2, size=(22, *size), dtype=torch.long))
```

test/test_prototype_transforms_consistency.py
```diff
@@ -840,3 +840,681 @@ def test_aa(self, inpt, interpolation):
...
+
+        yield (pil_image, target)
+
+        tensor_image = torch.randint(0, 256, size=(3, *size), dtype=torch.uint8)
+        target = {
+            "boxes": make_bounding_box(image_size=size, format="XYXY", extra_dims=(22,), dtype=torch.float),
+            "labels": features.Label(torch.randint(0, 81, size=(22,))),
+        }
+        if with_mask:
+            target["masks"] = features.SegmentationMask(torch.randint(0, 2, size=(22, *size), dtype=torch.long))
```
pmeier, line 865:
```suggestion
            target["masks"] = features.Mask(torch.randint(0, 2, size=(22, *size), dtype=torch.long))
```

test/test_prototype_transforms_consistency.py
```diff
@@ -840,3 +840,681 @@ def test_aa(self, inpt, interpolation):
...
+
+        yield (tensor_image, target)
+
+        feature_image = features.Image(torch.randint(0, 256, size=(3, *size), dtype=torch.uint8))
+        target = {
+            "boxes": make_bounding_box(image_size=size, format="XYXY", extra_dims=(22,), dtype=torch.float),
+            "labels": features.Label(torch.randint(0, 81, size=(22,))),
+        }
+        if with_mask:
+            target["masks"] = features.SegmentationMask(torch.randint(0, 2, size=(22, *size), dtype=torch.long))
```
pmeier, line 875:
```suggestion
            target["masks"] = features.Mask(torch.randint(0, 2, size=(22, *size), dtype=torch.long))
```

test/test_prototype_transforms_consistency.py
```diff
@@ -840,3 +840,681 @@ def test_aa(self, inpt, interpolation):
...
+        t = prototype_transforms.RandomShortestSize(
+            min_size=(480, 512, 544, 576, 608, 640, 672, 704, 736, 768, 800), max_size=1333
+        )
+        self._test_transform(t_ref, t)
+
+
+# -----
+# Dumped reference detection transforms here for consistency checks
+# torchvision/references/detection/transforms.py
+# -----
```
pmeier, line 927: Instead of dumping, we should just import them. This avoid duplication and enforces the tests to be in sync with the actual references:

```py
import pathlib
import importlib.util


def import_reference_module(*parts):
    *path_parts, name = parts

    here = pathlib.Path(__file__).parent
    project_root = here.parent
    location = project_root.joinpath("references", *path_parts, f"{name}.py")

    spec = importlib.util.spec_from_file_location(name, location=location)
    module = importlib.util.module_from_spec(spec)
    spec.loader.exec_module(module)
    return module


detection_transforms = import_reference_module("detection", "transforms")
t_ref = detection_transforms.RandomHorizontalFlip(p=1.0)
```

test/test_prototype_transforms_consistency.py
```diff
@@ -6,7 +6,7 @@
 import pytest
 
 import torch
-from prototype_common_utils import ArgsKwargs, assert_equal, make_images
+from prototype_common_utils import ArgsKwargs, assert_equal, make_bounding_box, make_images
```
pmeier, line 9:
```suggestion
from prototype_common_utils import ArgsKwargs, assert_equal, make_bounding_box, make_images, make_label, make_image, make_detection_mask
```

test/test_prototype_transforms_consistency.py
```diff
@@ -840,3 +840,681 @@ def test_aa(self, inpt, interpolation):
...
+
+
+class TestRefDetTransforms:
+    def make_datapoints(self, with_mask=True):
+        size = (600, 800)
+
+        pil_image = PIL.Image.new("RGB", size[::-1], 123)
+        target = {
+            "boxes": make_bounding_box(image_size=size, format="XYXY", extra_dims=(22,), dtype=torch.float),
+            "labels": features.Label(torch.randint(0, 81, size=(22,))),
```
pmeier, line 852: Is the number of categories important here? If not I would prefer to use the common utilities to create the label

```suggestion
            "labels": make_label(size=(22,)),
```

Note that #5879 introduces functionality to specify `categories=80`.

test/test_prototype_transforms_consistency.py
pmeier: Better yet, use the common utilities:

```suggestion
            target["masks"] = make_detection_mask(size=size, extra_dims=[(22,)], dtype=torch.long)
```

This also applies to the other comments like the one above.

test/test_prototype_transforms_consistency.py
pmeier, line 852: `22` seems to be the batch size here, correct? Can we please have this as a variable instead of hardcoding it multiple times? Maybe even call it `extra_dims` since that is what the parameter is called in the common utilities to generate the features.

test/test_prototype_transforms_consistency.py
```diff
@@ -840,3 +840,681 @@ def test_aa(self, inpt, interpolation):
         output = t(inpt)
 
         assert_equal(expected_output, output)
+
+
+class TestRefDetTransforms:
+    def make_datapoints(self, with_mask=True):
+        size = (600, 800)
+
+        pil_image = PIL.Image.new("RGB", size[::-1], 123)
```
pmeier, line 849:
```suggestion
        pil_image = F.to_pil_image(make_image(size=size, color_space=features.ColorSpace.RGB))
```

test/test_prototype_transforms_consistency.py
vfdev-5: I dumped it there as it was the easiest way to do especially that we wont keep this test file in the future...

test/test_prototype_transforms_consistency.py
vfdev-5: It is a number of objects, see that bbox has also 22

test/test_prototype_transforms_consistency.py
pmeier: > we wont keep this test file in the future...

True, but the two sources can still go out of sync until then. Plus, only ~100 lines in this PR is actually testing and the other ~700 lines are just copy pasted. This makes this file quite large and harder to work with for no reason.

vfdev-5: OK, i'll use importilb to get the code

test/test_prototype_transforms_consistency.py
pmeier: After #5879 the call should look like

```suggestion
            "labels": make_label(extra_dims=(22,), categories=80),
```

vfdev-5: Finally, I think it is unimportant to specify 80 categories instead of using default values.

test/test_prototype_transforms_consistency.py
```diff
@@ -1,16 +1,26 @@
...
+    make_detection_mask,
+    make_image,
+    make_images,
+    make_label,
+)
 from torchvision import transforms as legacy_transforms
 from torchvision._utils import sequence_to_str
 from torchvision.prototype import features, transforms as prototype_transforms
-from torchvision.prototype.transforms.functional import to_image_pil
+from torchvision.prototype.transforms.functional import to_image_pil, to_pil_image
```
pmeier, line 23: Why do we need both? Isn't the one just aliased to the other? I think `to_image_pil` is the new canonical name

test/test_prototype_transforms_consistency.py
```diff
@@ -840,3 +850,80 @@ def test_aa(self, inpt, interpolation):
...
+ref_det_filepath = Path(__file__).parent.parent / "references" / "detection" / "transforms.py"
+det_transforms = SourceFileLoader(ref_det_filepath.stem, ref_det_filepath.as_posix()).load_module()
+
+
+class TestRefDetTransforms:
+    def make_datapoints(self, with_mask=True):
+        size = (600, 800)
+        num_objects = 22
+
+        pil_image = to_pil_image(make_image(size=size, color_space=features.ColorSpace.RGB))
```
pmeier, line 866:
```suggestion
        pil_image = to_image_pil(make_image(size=size, color_space=features.ColorSpace.RGB))
```

test/test_prototype_transforms_consistency.py
pmeier:
```suggestion
from torchvision.prototype.transforms.functional import to_image_pil
```

test/test_prototype_transforms_consistency.py
```diff
@@ -840,3 +850,80 @@ def test_aa(self, inpt, interpolation):
         output = t(inpt)
 
         assert_equal(expected_output, output)
+
+
+# Import reference detection transforms here for consistency checks
+# torchvision/references/detection/transforms.py
+ref_det_filepath = Path(__file__).parent.parent / "references" / "detection" / "transforms.py"
+det_transforms = SourceFileLoader(ref_det_filepath.stem, ref_det_filepath.as_posix()).load_module()
```
pmeier, line 858: Why the `.as_posix()` here? Does that work on Windows? I'll guess CI will tell us. Did you mean to convert into a string?

```suggestion
det_transforms = SourceFileLoader(ref_det_filepath.stem, str(ref_det_filepath)).load_module()
```

pmeier (approved):
Looks a lot better, thank you! A few comments inline. Feel free to merge when resolved and CI is green.

test/test_prototype_transforms_consistency.py
vfdev-5: yes, it will work on windows as well.

PR #6569, opened 2022-09-12
[proto] Enable lazy loading for the data pipeline of CUB200 and Imagenet

ejguan:
Change the logic that will trigger pipeline iteration during the construction of the data pipeline. This PR would guarantee that all data or even meta data is loaded lazily until iteration starts.

### DIFF
--- a/torchvision/prototype/datasets/_builtin/cub200.py
+++ b/torchvision/prototype/datasets/_builtin/cub200.py
@@ -39,6 +39,10 @@ def _info() -> Dict[str, Any]:
     return dict(categories=read_categories_file(NAME))
 
 
+def image_filename_fn(rel_posix_path: str) -> str:
+    return rel_posix_path.rsplit("/", maxsplit=1)[1]
+
+
 @register_dataset(NAME)
 class CUB200(Dataset):
     """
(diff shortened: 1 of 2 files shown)


torchvision/prototype/datasets/_builtin/cub200.py
```diff
@@ -39,6 +39,10 @@ def _info() -> Dict[str, Any]:
     return dict(categories=read_categories_file(NAME))
 
 
+def image_filename_fn(rel_posix_path: str) -> str:
+    return rel_posix_path.rsplit("/", maxsplit=1)[1]
```
vadimkantorov, line 43: Will it work on Windows? should it use `os.pathsep` instead?

pmeier (commented):
Thanks Erjia, but I'm not sure if there is any benefit of this PR over #6076. Did I miss something? Otherwise, I'm inclined to close this and continue over there.

ejguan:
> Thanks Erjia, but I'm not sure if there is any benefit of this PR over #6076. Did I miss something? Otherwise, I'm inclined to close this and continue over there.

Oh, thanks pointing out! I completely forget the PR. Let me close this one, though.

Issue #6495, opened 2022-08-25
Stereo Depth Perception Transforms


YosuaMichael:
Currently we are in progress of revamping transforms API for torchvision library. During this revamp, we are considering the components and behavior for new use cases such as detection, segmentation, and video classification. In this document, we just want to raise awareness on several transforms in depth perception that might behave differently from the other use cases.

## Stereo Depth Perception Brief Introduction

In the Stereo Depth Perception problem, we are given **two images**, left and right, and for each pixel on the left image we want to guess the horizontal displacement of the pixel occurrence on the right image. The images are taken at the same time with two different cameras with the same upright orientation and their distance must be parallel with the horizontal axis / ground. The desired output is called **disparity map**, a 2d integer array with the same size as the image indicating how much displacement the pixel on that position is from left to right image (the unit of the displacement is in pixels). 

In a stereo depth perception dataset, they sometimes also give a **valid mask**. This is a 2d boolean array with the same size as the disparity map; the valid mask indicates whether the values in disparity maps are valid or not. Practically, we usually exclude the non-valid disparity maps from loss calculation during the training or validation.

## Components

1. Images: \
A pair of left and right image as input
2. Disparity map: \
2d integer array with same size as the image indicating the displacement of pixels on the image pair. Some dataset only give one disparity map which usually indicate the displacement from left to right image and some dataset provide a pair of left and right disparity maps.
3. Valid mask: \
2d boolean array with the same size as the disparity map indicating whether or not the disparity map value is valid and needs to be considered in loss calculation.

In general, a stereo depth perception dataset will have:

( 

(image_left, image_right), 

(disparity_map_left, [optional] disparity_map_right), 

([optional] valid_mask_left, [optional] valid_mask_right)

)

## Horizontal Flips

For horizontal flips, we do the horizontal flips on each component and then **we need to flip the order of left and right**. Supposing **HF** is an operation to do horizontal flip for a component, then:

horizontal_flips( (image_left, image_right), (disparity_map_left, disparity_map_right), (valid_mask_left, valid_mask_right) ) = 

(HF(image_right), HF(image_left)), (HF(disparity_map_right), HF(disparity_map_left)), (HF(valid_mask_right), HF(valid_mask_left))

The new behavior on this transform is that we need to flip the ordering of left and right which did not exist in previous use cases.

[...]

PR #6574, opened 2022-09-13
Fix prototype transforms for `(*, H, W)` segmentation masks

pmeier:
Most of our `*_segmentation_mask` kernels simply dispatch to the `*_image_tensor` kernel. However a segmentation mask (from the segmentation task) might have no channel dimension, i.e. shape `(*, H, W)`. Passing this directly into the image kernel will fail for some that require this dimension to be there. For example 

https://github.com/pytorch/vision/blob/cac4e228c9ca9e7564cb34406e7ebccfdd736976/torchvision/prototype/transforms/functional/_geometry.py#L566

This was not detected before because the CI only tested against "detection" masks, i.e. `(*; N, H, W)` where `N` denotes the number of objects.

All kernels work the same, but in case the input has no channel dimension, we `.unsqueeze(0)` before calling the image kernel and `.squeeze(0)` afterwards.

### DIFF
--- a/test/prototype_common_utils.py
+++ b/test/prototype_common_utils.py
@@ -178,7 +178,8 @@ def make_one_hot_labels(
         yield make_one_hot_label(extra_dims_)
 
 
-def make_segmentation_mask(size=None, *, num_objects=None, extra_dims=(), dtype=torch.uint8):
+def make_detection_mask(size=None, *, num_objects=None, extra_dims=(), dtype=torch.uint8):
+    # This produces "detection" masks, i.e. `(*, N, H, W)`, where `N` denotes the number of objects
     size = size if size is not None else torch.randint(16, 33, (2,)).tolist()
     num_objects = num_objects if num_objects is not None else int(torch.randint(1, 11, ()))
     shape = (*extra_dims, num_objects, *size)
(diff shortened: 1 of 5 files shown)


test/prototype_common_utils.py
```diff
@@ -178,22 +178,58 @@ def make_one_hot_labels(
         yield make_one_hot_label(extra_dims_)
 
 
-def make_segmentation_mask(size=None, *, num_objects=None, extra_dims=(), dtype=torch.uint8):
+def make_detection_mask(size=None, *, num_objects=None, extra_dims=(), dtype=torch.uint8):
```
pmeier, line 181: The naming is now quite weird. We are calling a `make_detection_mask` function and get a `features.SegmentationMask` back. Now that we have somewhat settled on the terms segmentation and detection masks, can we maybe rename the class to `features.Mask` since it represents both types?

test/test_prototype_transforms_functional.py
```diff
@@ -771,7 +778,8 @@ def _compute_expected_mask(mask, angle_, translate_, scale_, shear_, center_):
                         expected_mask[i, out_y, out_x] = mask[i, in_y, in_x]
         return expected_mask.to(mask.device)
 
-    for mask in make_segmentation_masks(extra_dims=((), (4,))):
+    # FIXME: `_compute_expected_mask` currently only works for "detection" masks. Extend it for "segmentation" masks.
+    for mask in make_detection_masks(extra_dims=((), (4,))):
```
pmeier, line 782: This is due to the `assert mask.ndim == 3`.  This was introduced in #6546 to make sure only detection masks are computed here. @vfdev-5 Could you fix the expected mask computation so both types of masks are supported? If that is done, we should use 

```suggestion
    for mask in make_detection_and_segmentation_masks(extra_dims=((), (4,))):
```

here.

This comment also applies to other occurrences of the same `FIXME` comment.

test/test_prototype_transforms_functional.py
```diff
@@ -1649,14 +1659,18 @@ def test_correctness_gaussian_blur_image_tensor(device, image_size, dt, ksize, s
...
+        # FIXME: This test currently only works for "detection" masks. Extend it for "segmentation" masks.
+        (F.elastic_segmentation_mask, make_detection_masks),
+    ],
 )
 def test_correctness_elastic_image_or_mask_tensor(device, fn, make_samples):
     in_box = [10, 15, 25, 35]
     for sample in make_samples(sizes=((64, 76),), extra_dims=((), (4,))):
         c, h, w = sample.shape[-3:]
         # Setup a dummy image with 4 points
-        print(sample.shape)
```
pmeier, line 1659: I took the liberty to remove this, since this seems like a debug statement that was forgotten before merge.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -367,15 +367,21 @@ def affine_bounding_box(
...
+    segmentation_mask: torch.Tensor,
     angle: float,
     translate: List[float],
     scale: float,
     shear: List[float],
     center: Optional[List[float]] = None,
 ) -> torch.Tensor:
-    return affine_image_tensor(
-        mask,
+    if segmentation_mask.ndim < 3:
```
pmeier, line 377: I would love to do that in a decorator, but `torch.jit.script` doesn't support it :cry: 

```py
def handle_segmentation_mask(mask_kernel):
    @functools.wraps(mask_kernel)
    def wrapper(mask, *other_args, **kwargs):
        if mask.ndim < 3:
            mask = mask.unsqueeze(0)
            needs_squeeze = True
        else:
            needs_squeeze = False

        output = mask_kernel(mask, *other_args, **kwargs)

        if needs_squeeze:
            output = output.squeeze(0)

        return output

    return wrapper


@handle_segmentation_mask
def affine_segmentation_mask(mask, ...):
    ...
```

datumbox: The story of our lives...

datumbox (approved):
The changes on the functional looks good to me.

pmeier:
Test failure is unrelated:

```
____________ test_eager_vs_scripted[gaussian_blur_image_tensor-50] _____________
  File "/home/runner/work/vision/vision/test/test_prototype_transforms_functional.py", line 602, in test_eager_vs_scripted
    msg=msg,

Mismatched elements: 1 / 1024 (0.1%)
Greatest absolute difference: 1 at index (1, 0, 15, 4)
Greatest relative difference: 0.006172839552164078 at index (1, 0, 15, 4)
```

test/prototype_common_utils.py
datumbox: Sounds good to me!

datumbox: On a second thought, I remember @YosuaMichael telling us that on Depth Perception they also have a sort of Valid Masks (see https://github.com/pytorch/vision/issues/6495).

@YosuaMichael could you clarify if they operate exactly like Segmentation/Detection masks and if we could just reuse this object?

pmeier: To give more context:

- segmentation mask is of shape `(*, H, W)` where each value corresponds to a class index
- detection mask is of shape `(*, N, H, W)` where `N` denotes the number of objects and the values are otherwise "boolean"

In both cases the `*` in the shape denotes arbitrary batch dimensions

YosuaMichael: @datumbox I think the valid_mask on depth perception will behave like detection_mask with 2 object (it is boolean, but can be implemented with 0/1 uint as well). So overall I think we can reuse it.

PR #6575, opened 2022-09-13
Update RAFT Stereo to be more sync with CREStereo implementation

YosuaMichael:
After discussing with @TeodorPoncu for on the CREStereo implementation, we plan to make the implementation more consistent. Here are the changes in RAFT Stereo:
1) Change the input name from `image1 and image2` to `left_image and right_image`
2) Enable using `flow_init` as input
3) Have `self.output_channel` as variable in RAFTStereo class

Here is the link to CREStereo implementation PR: https://github.com/pytorch/vision/pull/6310


TeodorPoncu (approved):
LGTM, just one question such that I know how to sync CREStereo to these changes.

torchvision/prototype/models/depth/stereo/raft_stereo.py
```diff
@@ -380,6 +380,10 @@ def __init__(
         super().__init__()
         _log_api_usage_once(self)
 
+        # This indicate that the disparity output will be only have 1 channel (represent horizontal axis).
+        # We need this because some stereo matching model like CREStereo might have 2 channel on the output
+        self.output_channel = 1
```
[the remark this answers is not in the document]
YosuaMichael: Ok, lets use `output_channels` then

torchvision/prototype/models/depth/stereo/raft_stereo.py
```diff
@@ -448,6 +455,10 @@ def forward(self, image1: Tensor, image2: Tensor, num_iters: int = 12) -> List[T
         coords0 = make_coords_grid(batch_size, Hf, Wf).to(fmap1.device)
         coords1 = make_coords_grid(batch_size, Hf, Wf).to(fmap1.device)
 
+        # We use flow_init for cascade inference
+        if flow_init is not None:
+            coords1 = coords1 + flow_init
```
datumbox, line 460: This is the only change for which I don't have context, but I understand it's a new feature.

datumbox (approved):
LGTM, I think the changes make sense and make the code much easier to read. Coming with basic knowledge of Depth Perception, the new naming allows me to understand what is happening and why.

torchvision/prototype/models/depth/stereo/raft_stereo.py
TeodorPoncu: When wanting to perform inference at a resolution significantly larger than that at which the model is trained, you can perform cascaded inference.

Cascaded inference first computes a flow at a low resolution, and uses that flow as a prior for the larger resolution image.

oke-aditya:
My doubt is do stereo models depend on order left and right input images. Can we not reverse them?
Like can we call the same model by changing left image to right and right image to left and still expect same output?

YosuaMichael:
@oke-aditya no we can't reverse the output since what the model does is guessing how much "displacement" of the pixels from left_image to the right_image, and this usually mean all pixels will be shifted to the right (not the other way around).

We have however augmentation that do horizontal flip, and in this case we will also flip the order of right and left image.

oke-aditya:
Perhaps we can document this.

YosuaMichael:
@oke-aditya do you have any recommendation on where we should document it? I think this behavior will be general on all depth stereo model (currently we also plan to add CREStereo).

oke-aditya:
Probably one root page for stereo models ? Also cc @datumbox for better ideas 💡

YosuaMichael:
The test error is not related to this PR changes

PR #6576, opened 2022-09-13
fix resize for segmentation masks without channel dimension

pmeier:
Follow-up to #6574.

```py
import torch

size = (16, 31)

segmentation_mask = features.SegmentationMask(torch.randint(0, 10, size))

print(segmentation_mask.shape, F.resize(segmentation_mask, size).shape)

```

```
torch.Size([16, 31]) torch.Size([1, 16, 31])
```

This was not detected in #6574 since kernel works without adding an explicit channel dimension, but it doesn't remove it again. This happens because

https://github.com/pytorch/vision/blob/b4686f2b7409d1783dfbb951492cd59bfed08bce/torchvision/prototype/transforms/functional/_geometry.py#L106

returns `1` for the number of channels if we have no channel dimension:

https://github.com/pytorch/vision/blob/b4686f2b7409d1783dfbb951492cd59bfed08bce/torchvision/transforms/functional_tensor.py#L24-L26

### DIFF
--- a/torchvision/transforms/functional_tensor.py
+++ b/torchvision/transforms/functional_tensor.py
@@ -132,7 +132,7 @@ def hflip(img: Tensor) -> Tensor:
 def crop(img: Tensor, top: int, left: int, height: int, width: int) -> Tensor:
     _assert_image_tensor(img)
 
-    w, h = get_image_size(img)
+    _, h, w = get_dimensions(img)
     right = left + width
     bottom = top + height
 
--- a/torchvision/transforms/functional_tensor.py
+++ b/torchvision/transforms/functional_tensor.py
@@ -132,7 +132,7 @@ def hflip(img: Tensor) -> Tensor:
 def crop(img: Tensor, top: int, left: int, height: int, width: int) -> Tensor:
     _assert_image_tensor(img)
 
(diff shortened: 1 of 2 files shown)


torchvision/transforms/functional_tensor.py
```diff
@@ -132,7 +132,7 @@ def hflip(img: Tensor) -> Tensor:
 def crop(img: Tensor, top: int, left: int, height: int, width: int) -> Tensor:
     _assert_image_tensor(img)
 
-    _, h, w = get_dimensions(img)
+    w, h = get_image_size(img)
```
pmeier, line 135: I found this while debugging `get_dimensions`. Since the number of channels is not used, we can get away with using `get_image_size`. Took me a while to figure out that it returns width first :roll_eyes: 

LMK if you don't want this change in this PR.

datumbox (approved):
LGTM, thanks!

torchvision/transforms/functional_tensor.py
datumbox: I will be removing the use of `get_image_size()` from stable `ElasticTransform` on a separate PR.

PR #6581, opened 2022-09-14
Fix issues with `get_image_size()`

datumbox:
This PR does 4 things:
- Fixes a bug on the prototype version of `get_image_size()`. We previously returned the result in the wrong order (we need test for this on a follow up).
- Moves the `get_image_size()` to the deprecated area of prototype.
- Removes the use of `get_image_size()` from `ElasticTransform` stable. This method is discouraged to be used in our own implementations and we missed its use during the reviews.
- Introduces equivalents for getting the spatial/channel sizes using more generic names (because videos are coming up)


torchvision/prototype/transforms/functional/__init__.py
```diff
@@ -125,4 +126,4 @@
     to_pil_image,
 )
 
-from ._deprecated import rgb_to_grayscale, to_grayscale  # usort: skip
+from ._deprecated import get_image_size, rgb_to_grayscale, to_grayscale, to_tensor  # usort: skip
```
datumbox, line 129: `to_tensor` was missing.

torchvision/prototype/transforms/functional/_deprecated.py
```diff
@@ -50,3 +52,12 @@ def to_tensor(inpt: Any) -> torch.Tensor:
...
     return _F.to_tensor(inpt)
+
+
+def get_image_size(image: Union[PIL.Image.Image, torch.Tensor, features.Image]) -> List[int]:
+    warnings.warn(
+        "The function `get_image_size(...)` is deprecated and will be removed in a future release. "
+        "Instead, please use `get_spatial_size(...)` which returns `(h, w)` instead of `(w, h)`."
+    )
+    _, h, w = get_chw(image)
+    return [w, h]
```
datumbox, line 63: Moving the method and fixing the bug. Previously it returned (h, w)

torchvision/transforms/transforms.py
```diff
@@ -2162,8 +2162,8 @@ def forward(self, tensor: Tensor) -> Tensor:
         Returns:
             PIL Image or Tensor: Transformed image.
         """
-        size = F.get_image_size(tensor)[::-1]
-        displacement = self.get_params(self.alpha, self.sigma, size)
+        _, height, width = F.get_dimensions(tensor)
```
datumbox, line 2165: Removing the use of `get_image_size()` from our transforms.

torchvision/prototype/transforms/functional/_meta.py
```diff
@@ -34,14 +34,17 @@ def get_dimensions(image: Union[PIL.Image.Image, torch.Tensor, features.Image])
     return list(get_chw(image))
 
 
-def get_image_num_channels(image: Union[PIL.Image.Image, torch.Tensor, features.Image]) -> int:
+def get_channel_size(image: Union[PIL.Image.Image, torch.Tensor, features.Image]) -> int:
```
pmeier, line 37: I agree to remove "image" from the name, but why not

```suggestion
def get_num_channels(image: Union[PIL.Image.Image, torch.Tensor, features.Image]) -> int:
```

torchvision/transforms/transforms.py
```diff
@@ -2162,8 +2162,8 @@ def forward(self, tensor: Tensor) -> Tensor:
...
+        displacement = self.get_params(self.alpha, self.sigma, [height, width])
```
pmeier, line 2166: I don't know if this works with JIT, but maybe we can use some syntax sugar

```suggestion
        _, *size = F.get_dimensions(tensor)
        displacement = self.get_params(self.alpha, self.sigma, size)
```

torchvision/prototype/transforms/functional/_meta.py
```diff
@@ -34,14 +34,17 @@ def get_dimensions(image: Union[PIL.Image.Image, torch.Tensor, features.Image])
...
     return num_channels
 
 
-def get_image_size(image: Union[PIL.Image.Image, torch.Tensor, features.Image]) -> List[int]:
-    _, *image_size = get_chw(image)
-    return image_size
+get_image_num_channels = get_channel_size
+
+
+def get_spatial_size(image: Union[PIL.Image.Image, torch.Tensor, features.Image]) -> List[int]:
```
pmeier, line 45: There was a push to also allow masks and boxes here. Are we going for that later?

torchvision/prototype/transforms/functional/_meta.py
datumbox: Sounds good to me.

torchvision/transforms/transforms.py
datumbox: I don't think this is supported by JIT. We also don't have a JIT test for this transform so I don't wanna keep adding stuff without proper testing. Is that OK?

torchvision/prototype/transforms/functional/_meta.py
datumbox: Yes we should add this. Perhaps also allow for handling bboxes independently from images as you proposed previously. Shall we do this on a follow up? The only reason I'm introducing the method right now is to deprecate the old `(w, h)` version.

torchvision/prototype/transforms/functional/_deprecated.py
```diff
@@ -50,3 +52,12 @@ def to_tensor(inpt: Any) -> torch.Tensor:
         "Instead, please use `to_image_tensor(...)` followed by `convert_image_dtype(...)`."
     )
     return _F.to_tensor(inpt)
+
+
+def get_image_size(image: Union[PIL.Image.Image, torch.Tensor, features.Image]) -> List[int]:
+    warnings.warn(
+        "The function `get_image_size(...)` is deprecated and will be removed in a future release. "
+        "Instead, please use `get_spatial_size(...)` which returns `(h, w)` instead of `(w, h)`."
```
pmeier, line 60: Nit 
```suggestion
        "Instead, please use `get_spatial_size(...)` which returns `[h, w]` instead of `[w, h]`."
```

torchvision/prototype/transforms/functional/_deprecated.py
pmeier: The other deprecated functions here emit a warning, but otherwise use the old code. Should we also do that here?

torchvision/prototype/transforms/functional/_meta.py
```diff
@@ -34,14 +34,17 @@ def get_dimensions(image: Union[PIL.Image.Image, torch.Tensor, features.Image])
...
-def get_image_num_channels(image: Union[PIL.Image.Image, torch.Tensor, features.Image]) -> int:
+def get_num_channels(image: Union[PIL.Image.Image, torch.Tensor, features.Image]) -> int:
     num_channels, *_ = get_chw(image)
     return num_channels
 
 
-def get_image_size(image: Union[PIL.Image.Image, torch.Tensor, features.Image]) -> List[int]:
-    _, *image_size = get_chw(image)
-    return image_size
+get_image_num_channels = get_num_channels
```
pmeier, line 42: Should we also properly deprecate this? If not, could you please add comment about the aliasing like we did for the other aliases?

https://github.com/pytorch/vision/blob/6b2e0a084536b7d7abced67839b52258d6e21712/torchvision/prototype/transforms/functional/_geometry.py#L93-L96

pmeier (approved):

datumbox:
The failing tests are unrelated. merging.

PR #6582, opened 2022-09-14
Rename `convert_bounding_box_format` => `convert_format_bounding_box`

datumbox:
Fully automatic update using my IDE. There shouldn't be anything worth reviewing. Should be fine if the tests pass.

The rational on the rename is that this is low-level Bounding Box kernel and as every such kernel, it should end with `_bounding_box`. I spotted it while refactoring the tests to catch non JIT-scriptable midlevels. 

If my understanding is incorrect and we consider this a midlevel, feel free to close the PR.

### DIFF
--- a/test/test_prototype_transforms_functional.py
+++ b/test/test_prototype_transforms_functional.py
@@ -19,7 +19,7 @@ from prototype_common_utils import (
 from torch import jit
 from torchvision.prototype import features
 from torchvision.prototype.transforms.functional._geometry import _center_crop_compute_padding
-from torchvision.prototype.transforms.functional._meta import convert_bounding_box_format
+from torchvision.prototype.transforms.functional._meta import convert_format_bounding_box
 from torchvision.transforms.functional import _get_perspective_coeffs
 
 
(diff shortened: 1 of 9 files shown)


oke-aditya:
Sounds bit bad grammatically `convert_format_bounding_box` :( and `convert_bounding_box_format` was more grammatically correct.

datumbox:
@oke-aditya  Though I agree that `convert_bounding_box_format` is the natural way to describe what the method does, it conflicts with literally all other naming conventions such as `horizontal_flip_bounding_box`, `resized_crop_bounding_box`, etc. It also makes it much harder to gather all the low-level kernels. See the tests at #6584 to see what I mean.

Not merging this is not going to be the end of the world, but this method is literally the only one what doesn't follow the naming convention and might require custom tricks to retrieve.

oke-aditya:
Yes. We can take the poetic license here I think 🤔

oke-aditya:
I mean it's OK let's rename.

datumbox:
> Yes. We can take the poetic license here I think

I like that. Very poetic. 😄 

I'll leave it to Victor and Philip to decide. No strong opinions on my side. :)

pmeier (approved):
This is the same change that we made with `convert_image_color_space_{tensor, pil}` -> `convert_color_space_image_{tensor, pil}` in #6408.

https://github.com/pytorch/vision/blob/321f655237e1dc6ed271b2f98a9aed909f7df0f4/torchvision/prototype/transforms/functional/_meta.py#L145

This weird naming was somewhat mitigated by introducing a dispatcher `convert_color_space`. Within in our current rules to need at least two kernels before we add a dispatcher, that is not possible here. Still, I agree, we should do this for consistency.

Thanks Vasilis. LGTM if CI is green!

vfdev-5:
I do not have a strong opinion on this rename, but yes, it reads less natural

PR #6584, opened 2022-09-14
Make prototype `F` JIT-scriptable

### LINKED ISSUE #6553: RFC: Make prototype `F` jit-scriptable again?
### 🚀 The feature

The mid-layer kernels `F` of the [new Transforms API](https://github.com/pytorch/vision/blob/main/torchvision/prototype/transforms/functional/__init__.py) are currently not JIT-scriptable because JIT doesn't support Tensor subclassing. This is a major BC-breaking change that potentially could be avoided by refactoring our mid-level kernels. 

Here is one way of how this could be achieved:
```python
def kernel(x: torch.Tensor) -> torch.Tensor:
    if isinstance(x, torch.Tensor) and (torch.jit.is_scripting() or not isinstance(x, features._Feature)):
        return _FT.kernel(x)  # image tensor dispatch
    elif isinstance(x, features._Feature):
        return x.kernel()  # _Feature dispatch
    else:
        return _FP.kernel(x)  # PIL dispatch
```

Effectively the above makes the `kernel()` operate exactly like before on the stable `F`: if a `torch.Tensor` kernel is provided then it is dispatched directly to the `_FT` kernel. The assumption is that while JIT-scripting, only image tensors are allowed. As long as the `_features.Image` and the `torch.Tensor` kernels are identical (which is true in our implementation), this means that we won't have any discrepancies between the two types. The extension on the new tensor subclasses (Image, BBox, Mask, Label etc) is only available in Python mode.

Unfortunately it's not possible to check the exact type of Tensor on run-time and throw the appropriate error if a user passes by accident a BBox or Label during JIT-scripting. But for these types, other checks (such as the dimension checks) will raise errors.

### Proof of Concept

The below implementation is a proof of concept. We intentionally make the kernels simulate different functionality in order to test that the dispatches work as expected.

```python
import numpy as np
import torch
from PIL import Image
from torchvision.transforms import functional_pil as _FP, functional_tensor as _FT


# Dummy low-level kernels, we only offer for image_* but we could have added for all types
def invertedmeanpix_image_tensor(x: torch.Tensor) -> float:
    return -_FT.invert(x).float().mean().item()  # return negative mean to simulate different functionality


@torch.jit.unused
def invertedmeanpix_image_pil(x: Image.Image) -> float:
    return np.mean(_FP.invert(x)) / 255.0  # return 0-1 scaled mean to simulate different functionality


# Dummy mid-level which fakes input type to make kernel JIT-scriptable
def invertedmeanpix(x: torch.Tensor) -> float:
    if isinstance(x, torch.Tensor) and (torch.jit.is_scripting() or not isinstance(x, features._Feature)):
        return invertedmeanpix_image_tensor(x)
    elif isinstance(x, features._Feature):
        return x.invert().float().mean().item()  # estimate standard mean to differentiate from the other 2 cases
    else:
        return invertedmeanpix_image_pil(x)


# JIT scripted kernel
invertedmeanpix_scripted = torch.jit.script(invertedmeanpix)
print(invertedmeanpix_scripted.code)

# Create dummy data
image_tensor = torch.randint(0, 256, (3, 30, 20), dtype=torch.uint8)
image_feature = features.Image(image_tensor)
image_pil = to_pil_image(image_tensor)

bbox_feature = features.BoundingBox(
    torch.tensor([[12, 13, 19, 18], [1, 15, 8, 19]]),
    format=features.BoundingBoxFormat.XYXY,
    image_size=image_tensor.shape[:-2],
)

label_data = torch.tensor([1, 2])
label_feature = features.Label(label_data)
ohelabel_data = torch.nn.functional.one_hot(label_data, 3)
ohelabel_feature = features.OneHotLabel(ohelabel_data)

segmask_data = torch.zeros(1, 30, 20)
segmask_data[13:19, 12:18] = 1
segmask_data[15:19, 1:8] = 2
segmask_feature = features.SegmentationMask(segmask_data)

detmask_data = torch.zeros(2, 30, 20)
detmask_data[0, 13:19, 12:18] = 1
detmask_data[1, 15:19, 1:8] = 1
detmask_feature = features.SegmentationMask(detmask_data)


# Assertions
value = invertedmeanpix(image_tensor)
assert value < 0
torch.testing.assert_close(invertedmeanpix(image_feature), -value)
torch.testing.assert_close(invertedmeanpix(image_pil), -value / 255.0)
torch.testing.assert_close(invertedmeanpix_scripted(image_tensor), value)
torch.testing.assert_close(invertedmeanpix(label_feature), label_data.float().mean().item())
torch.testing.assert_close(invertedmeanpix(ohelabel_feature), ohelabel_data.float().mean().item())
torch.testing.assert_close(invertedmeanpix(segmask_feature), segmask_data.mean().item())
torch.testing.assert_close(invertedmeanpix(detmask_feature), detmask_data.mean().item())
print("OK")
```

Output:
```
def invertedmeanpix(x: Tensor) -> float:
  _0 = __torch__.invertedmeanpix_image_tensor
  x0 = unchecked_cast(Tensor, x)
  x1 = unchecked_cast(Tensor, x0)
  return _0(x1, )

OK
```

cc @vfdev-5 @pmeier

datumbox:
Fixes #6553

- [x] Adds limited JIT tests for mid-level kernels
- [x] Makes mid-level kernels JIT-scriptable:
  - [x] augment
  - [x] colour
  - [x] deprecated
  - [x] geometry
  - [x] meta
  - [x] misc
  - [x] type conversion

co-authored with @vfdev-5


test/test_prototype_transforms_functional.py
```diff
@@ -426,27 +426,74 @@ def erase_image_tensor():
         yield ArgsKwargs(image, i=1, j=2, h=6, w=7, v=torch.rand(c, 6, 7))
 
 
+_KERNEL_TYPES = {"_image_tensor", "_image_pil", "_mask", "_bounding_box", "_label"}
+
+
+def _distinct_callables(callable_names):
```
datumbox, line 432: Quite a messy method to achieve want we want. The main problem is that not all aliases need to be excluded. Some are currently used to import things from `_FT` or `_FP` so we can't rely on `method.__name__`. So we need to have an over-engineered logic to filter things out. Perhaps that's not worth it and we should just  accept the dups. It's likely we will need a better way to discriminate between Mid-level and low-level kernels. Maybe a registration API or maybe a last minute restructuring of the definitions will do it.

test/test_prototype_transforms_functional.py
```diff
@@ -426,27 +426,74 @@ def erase_image_tensor():
...
+        if callable(f) and not name.startswith("_") and any(name.endswith(k) for k in _KERNEL_TYPES)
+    }
+    return _distinct_callables(kernel_names)
+
+
+def _get_distinct_midlevels():
+    midlevel_names = {
+        name
+        for name, f in F.__dict__.items()
+        if callable(f) and not name.startswith("_") and not any(name.endswith(k) for k in _KERNEL_TYPES)
```
datumbox, line 461: This definition of midlevel is a bit different. Basically here we consider as midlevel anything that doesn't follow the low-level kernel conventions. This is not technically accurate on the semantics but it's OK. All we want in these 2 tests to ensure we check all kernels (low-level and mid-levels) and doing the definition like this ensures that all kernels as checked.

test/test_prototype_transforms_functional.py
```diff
@@ -458,8 +505,8 @@ def test_scriptable(kernel):
     [
         pytest.param(func, id=name)
         for name, func in F.__dict__.items()
-        if not name.startswith("_")
-        and callable(func)
+        if not name.startswith("_") and callable(func)
+        # TODO: remove aliases
```
datumbox, line 509: Perhaps we can reuse `_get_distinct_midlevels()` here?

torchvision/prototype/transforms/functional/__init__.py
```diff
@@ -74,7 +74,7 @@
     five_crop,
     five_crop_image_pil,
     five_crop_image_tensor,
-    hflip,
+    hflip,  # TODO: Consider moving all pure alias definitions at the bottom of the file
```
datumbox, line 77: I know we have discussed this previously and decided to keep aliases defined in their files (aka geometry, color etc). One issue with this is that we are mixing "real kernels" with aliases. Of course once the migration completes, differentiating between the two will be easy. An alias will point to a method that its `__name__` won't match. But now we are not able to tell the difference between the two. So because everything is defined on the same file, which version of the mid-level kernel will be kept depends on the alphanumeric order. So `hflip` is the name selected instead of `horizontal_flip` and `vertical_flip` instead of `vflip`.

This could be resolved by moving the alias definition here at the end of this file and have them all in one place. No super strong opinion about this and we can leave the discussion for later. I just added a TODO, so that we can review again this decision in case we want to do something different.

torchvision/prototype/transforms/functional/_meta.py
```diff
@@ -10,6 +10,7 @@
 get_dimensions_image_pil = _FP.get_dimensions
 
 
+# TODO: Should this be prefixed with `_` similar to other methods that don't get exposed by init?
```
datumbox, line 13: Right now, all helper methods in all subpackages usually start with `_`. This is not about defining the scope, since all the subpackages are private. This is mostly about being consistent on the naming patterns. For example while validating that all kernels are being called by my utility on tests, I started debugging why `get_chw` wasn't selected only to realize it's because it's never imported and made public in `__init__.py`. Since this method is used in multiple files, this should happen in a separate PR if we decide to do it.

datumbox (commented):
Some early notes:

torchvision/prototype/transforms/functional/_color.py
```diff
@@ -1,152 +1,168 @@
...
-
 adjust_brightness_image_tensor = _FT.adjust_brightness
 adjust_brightness_image_pil = _FP.adjust_brightness
 
 
-def adjust_brightness(inpt: DType, brightness_factor: float) -> DType:
-    if isinstance(inpt, features._Feature):
+def adjust_brightness(inpt: torch.Tensor, brightness_factor: float) -> torch.Tensor:
+    if isinstance(inpt, torch.Tensor) and (
+        torch.jit.is_scripting() or torch.jit.is_tracing() or not isinstance(inpt, features._Feature)
```
datumbox, line 12: JIT doesn't like it if we move the `torch.jit.is_scripting() or torch.jit.is_tracing()` from the if and putting it on a separate variable before the if...

torchvision/prototype/features/_image.py
```diff
@@ -32,6 +32,31 @@ def from_pil_mode(cls, mode: str) -> ColorSpace:
         else:
             return cls.OTHER
 
+    @staticmethod
+    def from_tensor_shape(shape: List[int]) -> ColorSpace:
```
datumbox, line 36: I think this makes more sense to be on the Enum than the method. At any case it was moved because JIT didn't like it to be a class method. We don't have to keep it in the enum, we just need to have it as a standalone method.

test/test_prototype_transforms_functional.py
```diff
@@ -426,27 +426,74 @@ def erase_image_tensor():
...
+        pytest.param(midlevel, id=name)
+        for name, midlevel in _get_distinct_midlevels()
+        if name
         not in {
-            "to_image_tensor",
-            "get_num_channels",
-            "get_spatial_size",
-            "get_image_num_channels",
-            "get_image_size",
+            "InterpolationMode",
```
pmeier, line 485: Could we explicitly check for callables in `_get_distinct_midlevels` to avoid this enum here?

torchvision/prototype/features/_image.py
pmeier: So we can't even have classmethods on objects that are not covered by JIT? :exploding_head:

torchvision/prototype/features/_image.py
```diff
@@ -32,6 +32,31 @@ def from_pil_mode(cls, mode: str) -> ColorSpace:
...
+        return _from_tensor_shape(shape)
+
+
+def _from_tensor_shape(shape: List[int]) -> ColorSpace:
```
pmeier, line 40:
```suggestion
def _from_tensor_shape(shape: Tuple[int, ...]) -> ColorSpace:
```

But I'm guessing this also doesn't work with JIT?

torchvision/prototype/features/_image.py
pmeier: Plus, I intentionally put the term "guess" in the name, since the number of channels is not sufficient to pick the right colorspace. For example, CMYK also has 4 channels, but would be classified as RGBA. However, this is not a problem now since we don't support it yet and maybe never will.

pmeier (approved):
Few comments inline. Otherwise LGTM if CI is green.

test/test_prototype_transforms_functional.py
datumbox: We do check for callables but `InterpolationMode` is also a callable because it's a type.

torchvision/prototype/features/_image.py
datumbox: JIT says no: 😄 
```
RuntimeError: 
Expression of type dots cannot be used in a type expression:
  File "./vision/torchvision/prototype/features/_image.py", line 40
def _from_tensor_shape(shape: Tuple[int, ...]) -> ColorSpace:
                                         ~~~ <--- HERE
```

torchvision/prototype/transforms/functional/_augment.py
```diff
@@ -20,17 +19,17 @@ def erase_image_pil(
 
 
 def erase(
-    inpt: Union[torch.Tensor, PIL.Image.Image, features.Image],
+    inpt: torch.Tensor,
```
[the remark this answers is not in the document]
datumbox: It does, I like it. Let me see if JIT will like it as well.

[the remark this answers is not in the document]
datumbox: It probably worked. Have look.

vfdev-5:
@datumbox I let you review your PR once you are back :)

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -72,11 +72,10 @@ def _apply_image_transform(
 
         # Fill = 0 is not equivalent to None, https://github.com/pytorch/vision/issues/6517
         # So, we have to put fill as None if fill == 0
-        fill_: Optional[Union[int, float, Sequence[int], Sequence[float]]]
+        # This is due to BC with stable API which has fill = None by default
+        fill_ = F._geometry._convert_fill_arg(fill)
```
datumbox, line 76: We might want to consider making `_convert_fill_arg()` public on the future as this is regularly used in Transforms as a utility method.

datumbox (commented):
@vfdev-5 I might not be the best person to "review" this PR but I can confirm that any modification done on your side looks great. It was a good idea to do most of the work needed on `fill` on separate PRs. I had a look also at your PRs #6595 and #6599 and the direction looks great. Maintaining `F` JIT-scriptable is a major win, thanks a lot for your work on making it possible. Thank you for the excellent work.

@pmeier Could you please give us an independent review on this PR?

FYI: follow up work we should do is ensuring the signatures of `F` and `_Feature` are aligned. After merging this PR, the signatures of the dispatchers on the base class don't match (one uses Sequences and the other uses List).

vfdev-5:
> FYI: follow up work we should do is ensuring the signatures of F and _Feature are aligned.

@datumbox should they match ? `F` is now jit scriptable and `_Feature` is not, so the latter can also accept Sequence[int] or Sequence[float] for user API simplicity:
```python
Image.pad(..., fill=(1, 2, 3))
# instead of only List[float]
Image.pad(..., fill=[1.0, 2.0, 3.0])
```
Do we want that ? If yes, I can update this PR and put all `_Feature` fill typehint as `Optional[Union[int, float, List[float]]] = None`

datumbox:
@vfdev-5 One of the original proposals was that the `_Feature` class methods and `F` will have methods with aligned signatures. The first would receive the maximal possible parameters (even those not used) and do a simple redirection to the low-level kernel. The only reason I proposed to change them is to ensure this is uphold. We can certainly do this on a follow up PR where we confirm that all APIs are aligned overall. What do you think?

vfdev-5:
OK to me. @pmeier please review and we merge this PR and I'll update `_Feature` signatures in a follow-up PR

test/prototype_transforms_dispatcher_infos.py
```diff
@@ -113,14 +113,21 @@ def sample_inputs(self, *types):
...
-            features.Mask: F.perspective_mask,
-        },
-    ),
+    # FIXME:
+    # RuntimeError: perspective() is missing value for argument 'startpoints'.
+    # Declaration: perspective(Tensor inpt, int[][] startpoints, int[][] endpoints,
+    # Enum<__torch__.torchvision.transforms.functional.InterpolationMode> interpolation=Enum<InterpolationMode.BILINEAR>,
+    # Union(float[], float, int, NoneType) fill=None) -> Tensor
+    #
+    # This is probably due to the fact that F.perspective does not have the same signature as F.perspective_image_tensor
```
pmeier, line 122: Yes, if the signatures actually diverge, then this is the issue? I'm ok disabling this here for now and look at it later. 

Still, is this something we want? Shouldn't the public kernels be in sync with the dispatcher?

test/test_prototype_transforms_dispatchers.py
```diff
@@ -9,7 +9,6 @@
 
 
 class TestCommon:
-    @pytest.mark.xfail(reason="dispatchers are currently not scriptable")
```
pmeier, line 12: :tada:

test/test_prototype_transforms_functional.py
```diff
@@ -407,27 +407,74 @@ def erase_image_tensor():
...
+            "to_grayscale",
+            "to_pil_image",
+            "to_tensor",
         }
     ],
 )
-def test_scriptable(kernel):
-    jit.script(kernel)
+def test_scriptable_midlevel(midlevel):
+    jit.script(midlevel)  # TODO: pass data through it
```
pmeier, line 477: @vfdev-5 Isn't that superseded by the tests I've added?

torchvision/prototype/features/_feature.py
```diff
@@ -10,6 +10,11 @@
 F = TypeVar("F", bound="_Feature")
 
 
+# Due to torch.jit.script limitation we keep DType as torch.Tensor
+# instead of Union[torch.Tensor, PIL.Image.Image, features._Feature]
+DType = torch.Tensor
```
pmeier, line 15: What is the difference between this and `ImageType` in the next module? Should we have a cleanup PR that merges the two?

torchvision/prototype/features/_image.py
pmeier: If it works as is, I wouldn't mess with this class any further. If it is needed we can remove the `: Any` though.

torchvision/prototype/transforms/_geometry.py
```diff
@@ -252,7 +252,14 @@ def __init__(
 
     def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
         fill = self.fill[type(inpt)]
-        return F.pad(inpt, padding=self.padding, fill=fill, padding_mode=self.padding_mode)
+
+        # This cast does Sequence[int] -> List[int] and is required to make mypy happy
```
pmeier, line 256: Well, `mypy` is not at fault here. The kernel could support `Sequence[int]`, but we annotated `List[int]`. Since that is more specific, `mypy` is right to complain. We can always silence `mypy` with a `# type: ignore[...]` comment or use `cast(...)` since both have no runtime implications. Actually casting to a list while it is not needed hurts performance although probably not by a significant amount.

pmeier (approved):
Mostly questions from my side. LGTM if CI is green. Thanks @datumbox and @vfdev-5!

test/prototype_transforms_dispatcher_infos.py
datumbox: I believe @vfdev-5 made this change. As discussed previously there are other places where we need to align the signatures and this can happen on a follow up PR to avoid making this too long.

Personally I think it's worth aligning the signatures unless there is a good reason not to (perhaps the interpolation default value is one exception?). I'm open to discussing this and I think we should agree on the policy soon.

torchvision/prototype/features/_feature.py
datumbox: `DType = Union[torch.Tensor, PIL.Image.Image, features._Feature]` Aka can be bboxes, masks etc
`ImageType = Union[torch.Tensor, PIL.Image.Image, features.Image]` Aka only Image types (and soon Videos).

It's debatable if these aliases we use here make any sense IMO, especially if we can't define the actual types. I maintained this to keep track of what we support and where (else this info will be lost and we would have to review all signatures again) but I think we should make a decision before the cleanup PR. I would be OK removing this info in favour of `torch.Tensor` but I don't have strong opinion.

torchvision/prototype/features/_image.py
datumbox: @pmeier I'm also really flexible. I don't have context why the `tensor_contents` is introduced and what's supposed to be. Would you like to send a PR once this is merged to see if you can make it work in the original location?

torchvision/prototype/transforms/_geometry.py
datumbox: At this point where we try to improve the speed by any margin, I would prefer using ignore statements rather than casting. I don't know how negligible the introduction of `cast()` is but if there is no measurable slow-down I don't mind switching.

test/test_prototype_transforms_functional.py
datumbox: Maybe? I noticed that at one point early in the development, one of the kernels had a return type `Any` and this test didn't complaint. This is why usually to be safe, we take the strategy of:
1. JIT-scripting
2. Passing data through the kernel and compare it with the non-JIT version
3. Serialize/deserialize the method and confirm it still returns the right value.

See `_check_jit_scriptable()` from test_models.py for more info.

torchvision/prototype/features/_feature.py
pmeier: > `DType = Union[torch.Tensor, PIL.Image.Image, features._Feature]` Aka can be bboxes, masks etc
> `ImageType = Union[torch.Tensor, PIL.Image.Image, features.Image]` Aka only Image types (and soon Videos).

Thanks I missed that.

> I would be OK removing this info in favour of torch.Tensor (fake type) but I don't have strong opinion.

I think having such a type alias is a good thing, because it adds context. Without it we would either need to put such a comment onto each function or leave it out.

test/test_prototype_transforms_functional.py
vfdev-5: @datumbox in [test/test_prototype_transforms_dispatchers.py](https://github.com/pytorch/vision/pull/6584/files/d7e516cab2dd18f2d59cca91511b06f858fec779#diff-10032f2e1f2786df3d4ce4dc4c426fb875a7b4b12599f2047f022510b7ec2cb5) the test should script and execute midlevel op, so I think we can remove this `test_scriptable_midlevel`

torchvision/prototype/features/_feature.py
datumbox: That works for me. It just creates even more things we need to name properly. For example, `ImageType` is not a good name because soon we will add support of Videos. :( I do agree with you that it provides context on what we should be supporting and might make it easier to fix on the future if adds support of Subclassing. I don't know what's best here. @vfdev-5 thoughts?

test/test_prototype_transforms_functional.py
pmeier: I'll cleanup.

torchvision/prototype/features/_feature.py
vfdev-5: Let's keep these aliases

PR #6585, opened 2022-09-14
Fix the error message of `_ovewrite_value_param`

datumbox:
Originally reported by @pmeier:

The `_ovewrite_value_param()` method returns the wrong message.

```python
from torchvision.models.segmentation import fcn_resnet101, FCN_ResNet101_Weights

fcn_resnet101(weights=FCN_ResNet101_Weights.DEFAULT, aux_loss=False)
ValueError: The parameter 'False' expected value True but got False instead.
```

Instead we should have returned the error:
```
ValueError: The parameter 'aux_loss' expected value True but got False instead.
```

This PR fixes this.


pmeier (approved):
LGTM, thanks Vasilis!

PR #6586, opened 2022-09-15
[proto] Argument fill can accept dict of base types

vfdev-5:


torchvision/prototype/transforms/_geometry.py
```diff
@@ -196,9 +198,21 @@ def forward(self, *inputs: Any) -> Any:
...
+    else:
+        if not isinstance(fill, (numbers.Number, tuple, list)):
+            raise TypeError("Got inappropriate fill arg")
+
+
+def _setup_fill_arg(fill: Union[FillType, Dict[Type, FillType]]) -> Dict[Type, FillType]:
+    if isinstance(fill, dict):
+        return fill
+    else:
+        return defaultdict(lambda: fill, {features.Mask: 0})  # type: ignore[arg-type, return-value]
```
pmeier, line 215: Why does mask use a non-default value here? Is it because masks are always 1d and passing a RGB fill value will fail for it? If that is the case, I think we should push this to the user. Meaning, if they want to use an RGB fill value and masks at the same time, they should provide `fill` as a dict.

datumbox: Yeah I agree. Let's remove the default value for masks:
```suggestion
        return defaultdict(lambda: fill)
```

datumbox (approved):
LGTM, except only one remark before merging.

torchvision/prototype/transforms/_geometry.py
vfdev-5: I think this is not what we would like to have. If user does the following, I think they expect that only image is padded with 123 and mask is padded with a background value (=0)
```python
transform = transforms.Pad(1, fill=123, padding_mode="constant")
```
but if we remove `{features.Mask: 0}` then mask will be also padded with 123. So, to get correct behaviour users would need to do:
```python
transform = transforms.Pad(1, fill={features.Image: 123, features.Mask: 0}, padding_mode="constant")
```
and this does not look convenient.

datumbox: I understand what you are trying to fix. On the other hand I agree with Philip's point that it should be on user to define the right value. I don't mind either way, no blocking for me as-is.

vfdev-5: I suggest to keep things simple and instead of typing for code, adopt default behaviour as masks are padded with 0:
```diff
- transform = transforms.Pad(1, fill={features.Image: 123, features.Mask: 0}, padding_mode="constant")
+ transform = transforms.Pad(1, fill=123, padding_mode="constant")
```

PR #6587, opened 2022-09-15
reduce verbosity of prototype tests and split them into multiple steps

pmeier:
Running with `-vvv` emits one line per test / parametrization, which leads to enormous logs with very little information. They take some time to load and in some instances [GH even truncates the log](https://github.com/pytorch/vision/actions/runs/3058901939/jobs/4935646001). That means one needs to take extra steps just to see the failing test.

This PR reduces the verbosity to a normal level. Now tests will only be shown as a single character, i.e. `.` for passing or `F` for failing, while the total percentage is still visible. If one needs to debug an issue where the CI hangs and thus it is important to see which test causes this, the course of action is to temporarily increase the verbosity on the PR rather than global for everyone.

In addition to the verbosity decrease, the PR also splits the prototype tests into three steps namely testing datasets, transforms, and models individually to make it even easier to find failing tests.


vfdev-5 (approved):
OK, let's move on to enable logs :)

YosuaMichael (approved):
Thanks @pmeier , this looks good! hope it can help resolve truncated logs issue!

PR #6588, opened 2022-09-15
Remove aliases for `PILToTensor` and `pil_to_tensor()`

datumbox:
As noted in https://github.com/pytorch/vision/pull/6433#discussion_r971759989, sometimes PIL.Image is used to store masks. Consistently the `pil_to_tensor` shouldn't be an alias of `to_image_tensor`. More arguments can be made about this when considering that `to_image_tensor` is a generic method that handles numpy arrays and tensors.

In this PR I remove the aliases of `PILToTensor` and `pil_to_tensor()` and make them standalone. I also replace the use of `to_image_tensor` with `pil_to_tensor` when this is beneficial speedwise.


torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -487,7 +487,7 @@ def forward(self, *inputs: Any) -> Any:
         if isinstance(orig_image, torch.Tensor):
             image = orig_image
         else:  # isinstance(inpt, PIL.Image.Image):
-            image = F.to_image_tensor(orig_image)
+            image = F.pil_to_tensor(orig_image)
```
datumbox, line 490: Perf: There is no reason to wrap the PIL into an Image object because few lines below the meta-data will be lost due to the `view()` call.

torchvision/prototype/transforms/_color.py
```diff
@@ -114,7 +114,7 @@ def _get_params(self, sample: Any) -> Dict[str, Any]:
 
     def _permute_channels(self, inpt: Any, permutation: torch.Tensor) -> Any:
         if isinstance(inpt, PIL.Image.Image):
-            inpt = F.to_image_tensor(inpt)
+            inpt = F.pil_to_tensor(inpt)
```
datumbox, line 117: Perf: same here.

pmeier (approved):
Changes LGTM. Still, on a more general note: we are using the term `*_image_tensor` in our kernels for simple image tensors, i.e. no feature. Thus, we could argue that `to_image_tensor` should also return a simple tensor. IIUC, this would make this PR obsolete right?

datumbox:
@pmeier Yes it would make this PR obsolete but you would need many many other changes across the code-base to rewrap the objects to `features.Image`. Also a week ago or so you advocated to make `to_image_tensor()` return a `features.Image` to push users to the new API, which is the opposite of what you propose here.

Let me know if you have strong preference towards which direction.

pmeier:
> Also a week ago or so you advocated to make `to_image_tensor()` return a `features.Image` to push users to the new API, which is the opposite of what you propose here.

True. I didn't take this use case, i.e. using `to_image_tensor` into account. If we want to use it in the sense "Here is an image like object. Please give me a tensor back", using it for masks seems like a good thing to do. If we only want to use it for images, we need a canonical way how masks are handled. 

With the change in this PR we will have `pil_to_tensor` for masks and `to_image_tensor` for images, correct? Plus, the result for masks will also have to be wrapped into a `features.Mask` manually.

> you would need many many other changes across the code-base to rewrap the objects to `features.Image`

Does it actually? I mean, internally a `torch.Tensor` is treated the same as a `features.Image`. The only difference is that the former has no color space information attached to it. Basically all transformations apart from color space conversions should work out of the box without wrapping.

> Let me know if you have strong preference towards which direction.

Same here, but I'm leaning towards using `to_image_tensor` as common entry point for image like objects that afterwards have to be wrapped manually. The only thing bothering me is the term "image" in the name, but we can't use `to_tensor` for obvious reasons. Maybe we can name it `image_to_tensor`?

datumbox:
> Does it actually? I mean, internally a torch.Tensor is treated the same as a features.Image. The only difference is that the former has no color space information attached to it. Basically all transformations apart from color space conversions should work out of the box without wrapping.

What I mean is that if you make `to_image_tensor()` return a pure tensor, you will have to go everywhere in the code and do manual wrapping to `feature.Image`.

> Same here, but I'm leaning towards using to_image_tensor as common entry point for image like objects that afterwards have to be wrapped manually. The only thing bothering me is the term "image" in the name, but we can't use to_tensor for obvious reasons. Maybe we can name it image_to_tensor?

I think you have lots of unanswered questions. This PR addresses a real issue, aka the fact that PIL images have the need to be converted to tensors and that `to_image_tensor()` is not the right method for it. Maybe you are on to something, but your idea is not yet clear. I'll merge this PR to address the immediate issue as this aligns with what we have agreed so far and your past proposals. Happy to discuss a proposal that revises the name when you have it. It's OK if we end up removing parts of this PR in that process.

PR #6591, opened 2022-09-15
add segmentation reference consistency tests

pmeier:
Follow-up to #6566 for segmentation references.


test/test_prototype_transforms_consistency.py
```diff
@@ -873,7 +877,7 @@ def make_datapoints(self, with_mask=True):
 
         yield (pil_image, target)
 
-        tensor_image = torch.randint(0, 256, size=(3, *size), dtype=torch.uint8)
+        tensor_image = torch.Tensor(make_image(size=size, color_space=features.ColorSpace.RGB))
```
pmeier, line 880: I missed this before. Let's use the utilities everywhere.

test/test_prototype_transforms_consistency.py
```diff
@@ -927,3 +931,107 @@ def test_transform(self, t_ref, t, data_kwargs):
...
+                prototype_transforms.RandomHorizontalFlip(p=0.0),
+                dict(),
+            ),
+            # (
+            #     seg_transforms.RandomCrop(size=480),
+            #     prototype_transforms.RandomCrop(
+            #         size=480, pad_if_needed=True, fill=defaultdict(lambda: 0, {features.Mask: 255})
+            #     ),
+            #     dict(),
+            # ),
```
pmeier, line 995: This can only be added after #6586 was extended to `RandomCrop`

test/test_prototype_transforms_consistency.py
```diff
@@ -927,3 +931,107 @@ def test_transform(self, t_ref, t, data_kwargs):
...
+        mocker.patch(
+            "torchvision.prototype.transforms._geometry.torch.randint",
+            new=patched_randint,
+        )
+
+        t_ref = det_transforms.RandomResize(min_size=min_size, max_size=max_size)
+
+        self.check(t, t_ref)
+
+    def test_random_resize_eval(self):
```
pmeier, line 1029: Both this and the test above currently fail. The resized mask is off by quite a bit and so far I don't have an idea why or if that is expected. For the references the resize is performed as PIL image while the prototype transforms uses a `features.Mask` which internally is treated like a tensor image.

test/test_prototype_transforms_consistency.py
pmeier: See #6599.

test/test_prototype_transforms_consistency.py
```diff
@@ -927,3 +933,165 @@ def test_transform(self, t_ref, t, data_kwargs):
...
+
+
+seg_transforms = import_transforms_from_references("segmentation")
+
+
+# We need this transform for two reasons:
+# 1. transforms.RandomCrop uses a different scheme to pad images and masks of insufficient size than its name
+#    counterpart in the detection references. Thus, we cannot use it with `pad_if_needed=True`
+# 2. transforms.Pad only supports a fixed padding, but the segmentation datasets don't have a fixed image size.
+class PadIfSmaller(prototype_transforms.Transform):
```
pmeier, line 945: This implements the option suggested in https://github.com/pytorch/vision/pull/6433#discussion_r974500510. 

I've also played around with my proposal from https://github.com/pytorch/vision/pull/6433#discussion_r974033118, but it was a lot more complex. 

This PR demonstrates that this transform in combination with our `transforms.RandomCrop` can be used to mimic the behavior of the `RandomCrop` from the segmentation references. Thus, if we go through with this PR, I will port this transform to #6433.

test/test_prototype_transforms_consistency.py
```diff
@@ -927,3 +933,165 @@ def test_transform(self, t_ref, t, data_kwargs):
...
+    def __init__(self, size, fill=0):
+        super().__init__()
+        self.size = size
+        self.fill = prototype_transforms._geometry._setup_fill_arg(fill)
```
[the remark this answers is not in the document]
pmeier: We already do that here (see below). A transform has two touching points with `fill`:

1. At instantiation to setup the fill argument into the type dictionary, e.g.

    https://github.com/pytorch/vision/blob/841b9a19a5ae6fb92b517438061b9846153cc0c6/torchvision/prototype/transforms/_geometry.py#L443

2. At runtime to get the right fill value for the type and number of channels, e.g.

    https://github.com/pytorch/vision/blob/841b9a19a5ae6fb92b517438061b9846153cc0c6/torchvision/prototype/transforms/_geometry.py#L492-L493

test/test_prototype_transforms_consistency.py
```diff
@@ -927,3 +933,165 @@ def test_transform(self, t_ref, t, data_kwargs):
...
+        padding = [0, 0, max(self.size - width, 0), max(self.size - height, 0)]
+        needs_padding = any(padding)
+        return dict(padding=padding, needs_padding=needs_padding)
+
+    def _transform(self, inpt, params):
+        if not params["needs_padding"]:
+            return inpt
+
+        fill = self.fill[type(inpt)]
+        fill = F._geometry._convert_fill_arg(fill)
```
pmeier, line 962: Pointer for above.

datumbox: I missed it sorry 😞

pmeier:
Test failure is unrelated.

test/test_prototype_transforms_consistency.py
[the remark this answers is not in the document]
datumbox: Sorry Philip, I confused `_setup_fill_arg` with `_convert_fill_arg`. You are right, it looks good to me.

datumbox (approved):
The `PadIfSmaller` looks good to me. Let's unblock to continue the accuracy tests.

PR #6592, opened 2022-09-15
[FBcode->GH] Fix vit model assert text to be compatible with torchmultimodal

YosuaMichael:
During FBSync I notice that the PR https://github.com/pytorch/vision/pull/6583 break [torchmultimodal's test](https://github.com/facebookresearch/multimodal/blob/main/test/models/albef/test_image_encoder.py#L52).

Here is the part that break:
```
with pytest.raises(AssertionError, match="Wrong image height!"):
```

I modify the diffs internally and this PR is to make sure fbcode and github are in sync.

### DIFF
--- a/torchvision/models/vision_transformer.py
+++ b/torchvision/models/vision_transformer.py
@@ -268,8 +268,8 @@ class VisionTransformer(nn.Module):
     def _process_input(self, x: torch.Tensor) -> torch.Tensor:
         n, c, h, w = x.shape
         p = self.patch_size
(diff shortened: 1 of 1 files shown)


datumbox (approved):
LGTM, thanks!

datumbox:
@YosuaMichael I changed the classification tag for this. I don't think it's should be called out as a bug on the release notes. Let me know if you think otherwise or feel free to change the tags back.

Issue #6589, opened 2022-09-15
`fasterrcnn_resnet50_fpn` Windows GPU tests failing on CUDA 11.6


datumbox:
### 🐛 Describe the bug

After the removal of CUDA 11.3 and the setting of 11.6 as default the tests at `fasterrcnn_resnet50_fpn` started [failing](https://app.circleci.com/pipelines/github/pytorch/vision/20144/workflows/7dc88095-fe04-43a7-8c3f-64193a9f2661/jobs/1637810) across all python versions:
```
Traceback (most recent call last):
  File "C:\Users\circleci\project\test\test_models.py", line 775, in check_out
    _assert_expected(output, model_name, prec=prec)
  File "C:\Users\circleci\project\test\test_models.py", line 117, in _assert_expected
    torch.testing.assert_close(output, expected, rtol=rtol, atol=atol, check_dtype=False, check_device=False)
  File "C:\Users\circleci\project\env\lib\site-packages\torch\testing\_comparison.py", line 1342, in assert_close
    assert_equal(
  File "C:\Users\circleci\project\env\lib\site-packages\torch\testing\_comparison.py", line 1093, in assert_equal
    raise error_metas[0].to_error(msg)
AssertionError: Tensor-likes are not close!

Mismatched elements: 61 / 80 (76.2%)
Greatest absolute difference: 182.7935028076172 at index (9, 1) (up to 0.01 allowed)
Greatest relative difference: inf at index (1, 0) (up to 0.01 allowed)

The failure occurred for item [0]['boxes']

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\circleci\project\test\test_models.py", line 803, in test_detection_model
    full_validation &= check_out(out)
  File "C:\Users\circleci\project\test\test_models.py", line 783, in check_out
    torch.testing.assert_close(
  File "C:\Users\circleci\project\env\lib\site-packages\torch\testing\_comparison.py", line 1342, in assert_close
    assert_equal(
  File "C:\Users\circleci\project\env\lib\site-packages\torch\testing\_comparison.py", line 1093, in assert_equal
    raise error_metas[0].to_error(msg)
AssertionError: Tensor-likes are not close!

Mismatched elements: 1 / 20 (5.0%)
Greatest absolute difference: 0.019545435905456543 at index (19,) (up to 0.01 allowed)
Greatest relative difference: 0.026511636142229424 at index (19,) (up to 0.01 allowed)
```

[...]

PR #6593, opened 2022-09-16
Remove warnings pytest models

### LINKED ISSUE #6155: Fix warnings in the tests
Currently, there are following warnings when running tests:
```
test/test_models.py::test_quantized_classification_model[googlenet]
  /root/project/torchvision/models/googlenet.py:47: FutureWarning: The default weight initialization of GoogleNet will be changed in future releases of torchvision. If you wish to keep the old behavior (which leads to long initialization times due to scipy/scipy#11299), please set init_weights=True.
    warnings.warn(

test/test_models.py::test_quantized_classification_model[inception_v3]
  /root/project/torchvision/models/inception.py:43: FutureWarning: The default weight initialization of inception_v3 will be changed in future releases of torchvision. If you wish to keep the old behavior (which leads to long initialization times due to scipy/scipy#11299), please set init_weights=True.
    warnings.warn(

test/test_datasets.py::VOCDetectionTestCase::test_transforms
  /root/project/torchvision/datasets/voc.py:82: UserWarning: Accessing the test image set of the year 2007 with year='2007-test' is deprecated since 0.12 and will be removed in 0.14. Please use the combination year='2007' and image_set='test' instead.
    warnings.warn(
```

Source: https://app.circleci.com/pipelines/github/pytorch/vision/18272/workflows/46f9c352-ff53-4d87-a894-b124e49d0773/jobs/1478749

Let's fix these tests such that warnings are not shown anymore.

cc @pmeier

ambujpawar:
- [x] Closes #6155


test/test_models.py
```diff
@@ -316,6 +316,10 @@ def _check_input_backprop(model, inputs):
     "s3d": {
         "input_shape": (1, 3, 16, 224, 224),
     },
+    "googlenet": {
+        "num_classes": 50,
```
datumbox, line 320: No need to define default params:
```suggestion
```

datumbox:
@ambujpawar Thanks for the PR.

The tests are [failing](https://app.circleci.com/pipelines/github/pytorch/vision/20216/workflows/4caf16d9-af48-4b7f-b988-e73879060725/jobs/1644048) and look related. Could you please have a look? Thanks!

ambujpawar:
Thanks for the review. I tried to resolve those errors by just modifying the tests. But apparently, it was not possible for inceptionv3 model. Please try to have a look at the solution :)

torchvision/models/quantization/inception.py
```diff
@@ -128,6 +128,7 @@ def __init__(
         num_classes: int = 1000,
         aux_logits: bool = True,
         transform_input: bool = False,
+        init_weights: Optional[bool] = None,
```
pmeier, line 131: This would introduce a new API surface. Not sure if we want that. cc @datumbox

In case we don't, we can simply hardcode `init_weights=True` in the call to the superclass constructor.

test/test_models.py
```diff
@@ -316,6 +316,9 @@ def _check_input_backprop(model, inputs):
     "s3d": {
         "input_shape": (1, 3, 16, 224, 224),
     },
+    "googlenet": {
+        "init_weights": True,
+    },
```
pmeier, line 321: Depending on the discussion above, this maybe should be removed again. If we decide to keep `init_weights`, remove the trailing comma to keep this on a single line with our auto formatting:

```suggestion
    "googlenet": {"init_weights": True},
```

torchvision/models/quantization/inception.py
datumbox: @pmeier I believe the `init_weights` and `dropout` parameters are unintentionally missed here. This issue is caused because we didn't follow the normal idiom for `Inception3`. Here is how `GoogLeNet` does it:
https://github.com/pytorch/vision/blob/a89b1957a62e2f68f001d5d60268743edbe164d8/torchvision/models/quantization/googlenet.py#L76-L79


Normally we grab args and kwargs, overwrite the things we want and pass them to the super constructor. IMO `QuantizableInception3` should do exactly the same. Instead of trying to redefine all the parameters with the same defaults, we should just do the redirection.

@ambujpawar does the way I describe the above make sense to you? Would you be able to make the change? Happy to help

ambujpawar: Hi, yeah it makes sense to me. I wanted to the same thing initially but was not sure if I should progress that way. 
I am happy to work on that as well.

Shall it be done as a part of this PR? Or shall I create a new issue for it?
And wouldn't it also require changes in the API surface that @pmeier mentioned above?

datumbox: We should do it here. The changes on the API surface are OK. There are no BC breakages and this is exactly how every other model does it currently (GoogleNet, MobileNets, ResNets, ShuffleNet etc). This is literally the only model that doesn't follow the idiom and that's why we got on hot water here. Someone added after the fact the parameters on the base class but forgot to do so on the Quantized version.

ambujpawar: Cool, I will get started with it then! Thanks for making it clear for me 👍

pmeier (approved):
LGTM if CI is green. Thanks @ambujpawar!

ambujpawar:
Failing tests seems unrelated to the PR and I see that the related tasks are listed in [issues](https://github.com/pytorch/vision/issues)

pmeier:
The following could be related, but I'll leave that up to @datumbox since he knows the model tests better than I do.

```
_____________ test_detection_model[cuda-fasterrcnn_resnet50_fpn] ______________
  File "C:\Users\circleci\project\test\test_models.py", line 776, in check_out
  File "C:\Users\circleci\project\test\test_models.py", line 117, in _assert_expected
  File "C:\Users\circleci\project\env\lib\site-packages\torch\testing\_comparison.py", line 1342, in assert_close
    assert_equal(
  File "C:\Users\circleci\project\env\lib\site-packages\torch\testing\_comparison.py", line 1093, in assert_equal

Mismatched elements: 61 / 80 (76.2%)
Greatest absolute difference: 182.7935028076172 at index (9, 1) (up to 0.01 allowed)
Greatest relative difference: inf at index (1, 0) (up to 0.01 allowed)

The failure occurred for item [0]['boxes']

  File "C:\Users\circleci\project\test\test_models.py", line 804, in test_detection_model
    full_validation &= check_out(out)
  File "C:\Users\circleci\project\test\test_models.py", line 784, in check_out
    torch.testing.assert_close(
  File "C:\Users\circleci\project\env\lib\site-packages\torch\testing\_comparison.py", line 1342, in assert_close
    assert_equal(
  File "C:\Users\circleci\project\env\lib\site-packages\torch\testing\_comparison.py", line 1093, in assert_equal

Mismatched elements: 1 / 20 (5.0%)
Greatest absolute difference: 0.019545435905456543 at index (19,) (up to 0.01 allowed)
Greatest relative difference: 0.026511636142229424 at index (19,) (up to 0.01 allowed)
```

ambujpawar:
I think it is unrelated, as it was happening before this PR: https://github.com/pytorch/vision/issues/6589

torchvision/models/quantization/inception.py
```diff
@@ -142,6 +134,8 @@ def __init__(
                 QuantizableInceptionE,
                 QuantizableInceptionAux,
             ],
+            *args,
```
datumbox, line 137: @pmeier Since you are the most Pythonic person I know, any idea why this works? We pass the `inception_blocks` keyword parameter before *args. We do exactly the same thing on `QuantizableGoogLeNet`. Could this because we dont actually have any positional args in the classes? I was expecting that the class would emit an error for putting positional arguments after named params but it doesn't.

datumbox (approved):
The highlighted failure on the fasterrcnn model is not related. The changes LGTM.

Only one question which we could resolve elsewhere.

torchvision/models/quantization/inception.py
pmeier: This is one of the Python quirks that are syntactically allowed, but should never be used due to its very weird behavior. TL;DR: `foo(baz="baz", "bar")` is not valid syntax while `foo(baz="baz", *["bar"])` is. See python/cpython#82741.

I would advise to put the `*args` before the `inception_blocks=...` parameter to avoid confusion. Same for all other occurrences of this pattern. I don't know of a linter that checks this for us though.

ambujpawar: Done! 👍 
LGTM now?

pmeier: Not sure if we want to do it in this PR or split this off to a follow-up cc @datumbox. Nevertheless this bothered me a bit to much TBH. I wrote a small `flake8` plugin to find all occurrences of this. Put the following in a `nuakw.py` file in the root of the project

```py
import ast
import contextlib


class NoUnpackingAfterKeywordFinder(ast.NodeVisitor):
    def __init__(self):
        self.errors = []

    def visit_Call(self, node: ast.Call):
        with contextlib.suppress(StopIteration):
            starred = next(starred for starred in node.args if isinstance(starred, ast.Starred))
            if any(
                (keyword.value.lineno, keyword.value.col_offset) < (starred.lineno, starred.col_offset)
                for keyword in node.keywords
            ):
                self.errors.append(
                    (node.lineno, node.col_offset, "NUAKW Argument unpacking after keyword argument", type(self))
                )

        self.generic_visit(node)


class NoUnpackingAfterKeyword:
    name = "no_unpacking_after_keyword"
    version = "0.0.1"

    def __init__(self, tree: ast.AST):
        self._tree = tree

    def run(self):
        visitor = NoUnpackingAfterKeywordFinder()
        visitor.visit(self._tree)
        yield from visitor.errors

```

Now, put the following at the bottom of `setup.cfg`

```
[flake8:local-plugins]
extension =
    NUAKW = nuakw:NoUnpackingAfterKeyword
paths =
    ./

```

Running `flake8 torchvision` yields the following errors:

```
torchvision/models/quantization/inception.py:44:9: NUAKW Argument unpacking after keyword argument
torchvision/models/quantization/inception.py:55:9: NUAKW Argument unpacking after keyword argument
torchvision/models/quantization/inception.py:66:9: NUAKW Argument unpacking after keyword argument
torchvision/models/quantization/inception.py:77:9: NUAKW Argument unpacking after keyword argument
torchvision/models/quantization/inception.py:88:9: NUAKW Argument unpacking after keyword argument
torchvision/models/quantization/inception.py:122:9: NUAKW Argument unpacking after keyword argument
torchvision/models/quantization/googlenet.py:42:9: NUAKW Argument unpacking after keyword argument
torchvision/models/quantization/googlenet.py:53:9: NUAKW Argument unpacking after keyword argument
torchvision/models/quantization/googlenet.py:77:9: NUAKW Argument unpacking after keyword argument
torchvision/models/quantization/mobilenetv3.py:86:9: NUAKW Argument unpacking after keyword argument
```

If we fix one, we should fix all.

ambujpawar: This seemed to be a short fix, so I have implement this in this PR itself :)

ambujpawar: cc @pmeier @datumbox Shall we go to master now?

datumbox (approved):
LGTM (again) @ambujpawar. Thank you very much for being ultra patient and flexible with us. You did end up fixing many more issues than originally described. I do apologise for this but I think your PR leaves our codebase in much better state than before.

@pmeier Thanks a lot for the explanation you gave on the thread above. I agree that the aforementioned syntax is problematic and should be avoided. We can review whether its avoidance should be done via linter or via our code reviews, offline to avoid further delays on this work.

I just kicked the CI once more, we should be able to merge this once everything looks green. :)

ambujpawar:
Thanks for the kind words. Looking forward to solving more issues
Thanks for yours patience.

PR #6594, opened 2022-09-16
Update typehint for fill arg in rotate

vfdev-5:
Description:
- Update typehint for fill arg in rotate, mainly

```diff
- fill: Optional[List[float]] = None,
+ fill: Optional[Union[int, float, List[float]]] = None,
```
(we can't add `List[int]` into the typehint as JIT does not know how to iterate over the list: 'Union[List[float], List[int]]' object is not iterable)

Underlying functions are already almost supporting scalar values.

Next steps: 
- update fill arg to affine, perspective, elastic etc
- refactor pad_image_tensor to have the same typehint


datumbox (approved):
LGTM, thanks!

PR #6595, opened 2022-09-16
Updated fill arg typehint for affine, perspective and elastic ops

vfdev-5:
Follow-up to https://github.com/pytorch/vision/pull/6594

### DIFF
--- a/torchvision/prototype/features/_image.py
+++ b/torchvision/prototype/features/_image.py
@@ -177,12 +177,11 @@ class Image(_Feature):
         if not isinstance(padding, int):
             padding = list(padding)
 
-        # PyTorch's pad supports only scalars on fill. So we need to overwrite the colour
-        if isinstance(fill, (int, float)) or fill is None:
-            output = self._F.pad_image_tensor(self, padding, fill=fill, padding_mode=padding_mode)
-        else:
-            output = self._F._geometry._pad_with_vector_fill(self, padding, fill=fill, padding_mode=padding_mode)
+        # This cast does Sequence -> List and is required to make mypy happy
+        if not (fill is None or isinstance(fill, (int, float))):
+            fill = list(fill)
 
+        output = self._F.pad_image_tensor(self, padding, fill=fill, padding_mode=padding_mode)
         return Image.new_like(self, output)
 
     def rotate(
--- a/torchvision/prototype/features/_image.py
+++ b/torchvision/prototype/features/_image.py
@@ -177,9 +177,7 @@ class Image(_Feature):
         if not isinstance(padding, int):
             padding = list(padding)
 
-        # This cast does Sequence -> List and is required to make mypy happy
-        if not (fill is None or isinstance(fill, (int, float))):
-            fill = list(fill)
+        fill = self._F._geometry._convert_fill_arg(fill)
 
         output = self._F.pad_image_tensor(self, padding, fill=fill, padding_mode=padding_mode)
         return Image.new_like(self, output)
--- a/torchvision/prototype/transforms/functional/_geometry.py
+++ b/torchvision/prototype/transforms/functional/_geometry.py
@@ -590,7 +590,23 @@ pad_image_pil = _FP.pad
 def pad_image_tensor(
     img: torch.Tensor,
     padding: Union[int, List[int]],
-    fill: Optional[Union[int, float]] = 0,
+    fill: Optional[Union[int, float, List[float]]] = None,
+    padding_mode: str = "constant",
+) -> torch.Tensor:
+    if fill is None:
+        # This JIT workaround
+        return _pad_with_scalar_fill(img, padding, fill=None, padding_mode=padding_mode)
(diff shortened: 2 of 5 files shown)


torchvision/prototype/features/_image.py
```diff
@@ -177,12 +177,11 @@ def pad(
         if not isinstance(padding, int):
             padding = list(padding)
 
-        # PyTorch's pad supports only scalars on fill. So we need to overwrite the colour
-        if isinstance(fill, (int, float)) or fill is None:
-            output = self._F.pad_image_tensor(self, padding, fill=fill, padding_mode=padding_mode)
-        else:
-            output = self._F._geometry._pad_with_vector_fill(self, padding, fill=fill, padding_mode=padding_mode)
+        # This cast does Sequence -> List and is required to make mypy happy
+        if not (fill is None or isinstance(fill, (int, float))):
```
vfdev-5, line 181: I maybe need to replace this with `_convert_fill_arg`

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -591,7 +590,23 @@ def rotate(
 def pad_image_tensor(
     img: torch.Tensor,
     padding: Union[int, List[int]],
-    fill: Optional[Union[int, float]] = 0,
+    fill: Optional[Union[int, float, List[float]]] = None,
```
datumbox, line 593: I know this kernel is private on stable and I agree with the change to ensure consistency. Any BC breakages that can creep upwards due to the change on the default value?

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -591,7 +590,23 @@ def rotate(
...
+    padding_mode: str = "constant",
+) -> torch.Tensor:
+    if fill is None:
+        # This is a JIT workaround
+        return _pad_with_scalar_fill(img, padding, fill=None, padding_mode=padding_mode)
+    elif isinstance(fill, (int, float)) or len(fill) == 1:
+        fill_number = fill[0] if isinstance(fill, list) else fill
+        return _pad_with_scalar_fill(img, padding, fill=fill_number, padding_mode=padding_mode)
```
datumbox, line 601: I just wonder if we could do something like:
```suggestion
    if fill is None or isinstance(fill, (int, float)) or len(fill) == 1:
        if fill is None:
            fill = 0
        fill_number = fill[0] if isinstance(fill, list) else fill
        return _pad_with_scalar_fill(img, padding, fill=fill_number, padding_mode=padding_mode)
```

To avoid the duplicate dispatch to scalar. I assume that `fill-None` ends up becoming 0?

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -591,7 +590,23 @@ def rotate(
...
+    else:
+        return _pad_with_vector_fill(img, padding, fill=fill, padding_mode=padding_mode)
+
+
+def _pad_with_scalar_fill(
+    img: torch.Tensor,
+    padding: Union[int, List[int]],
+    fill: Optional[Union[int, float]] = None,
```
datumbox, line 609: If what I proposed above makes sense, you can avoid the optional here:
```suggestion
    fill: Union[int, float],
```

At any case I don't think we should pass a default value here `None`. We should always receive it as a positional arg, similar to what we do for `_pad_with_vector_fill`

datumbox (commented):
Overall looks good. A few questions/suggestions but nothing major. Let me know your thoughts.

torchvision/prototype/transforms/functional/_geometry.py
vfdev-5: I can change it to `fill: Union[int, float, None]` and remove default value.

torchvision/prototype/transforms/functional/_geometry.py
vfdev-5: I was thinking that below code is valid, but JIT does not agree
```
    if fill is None or isinstance(fill, (int, float)) or len(fill) == 1:
        fill_number = fill[0] if isinstance(fill, list) else fill
```

torchvision/prototype/transforms/functional/_geometry.py
vfdev-5: `pad_image_tensor` is low-level new API so I think there is no BC to keep. In any case fill=None will be transformed to fill=0 inside https://github.com/pytorch/vision/blob/c0911e31d367eca9600541bd634564c73753abbf/torchvision/transforms/functional_tensor.py#L381-L382

datumbox:
@vfdev-5 The work of making `F` JIT-scriptable (see https://github.com/pytorch/vision/pull/6584) depends on this PR. We should consider replacing Sequences for Lists and updating the `_convert_fill_arg()` to be JIT scriptable.

pmeier (approved):
Nothing stands out, so LGTM if CI is green.

PR #6596, opened 2022-09-16
improve prototype transforms kernel tests

pmeier:
This PR brings a few improvements to the new prototype transforms kernel tests and ports some to the new framework:

- Support PIL vs tensor mask comparison.
- Port common tests for `affine_mask` and `resize_mask` and reference tests for `affine_mask` and `affine_bounding_box`to new framework. This means the three "initial" functionals, i.e. `horizontal_flip_*`, `resize_*`, and `affine_*` are now completed for all features types and thus serve as a better reference for the remaining kernels.
- Add `KernelInfo`'s for `convert_format_bounding_box` and `convert_color_space_image_tensor` since they are already in use for the sample inputs of the other kernels and thus need to be working correctly.
- Add a dtype and device consistency test, i.e. output should have the same dtype and device as the input. As discussed offline, there are a number of bounding box kernels that don't do that. I've xfailed them for now and will send a follow-up PR to fix them.
- Some minor cleanup that I will add inline comments for

### DIFF
--- a/test/prototype_transforms_kernel_infos.py
+++ b/test/prototype_transforms_kernel_infos.py
@@ -67,7 +67,7 @@ KERNEL_INFOS = []
 
 
 def sample_inputs_horizontal_flip_image_tensor():
-    for image_loader in make_image_loaders(sizes=[None], dtypes=[torch.float32]):
+    for image_loader in make_image_loaders(sizes=["random"], dtypes=[torch.float32]):
         yield ArgsKwargs(image_loader)
 
 
@@ -86,7 +86,7 @@ def sample_inputs_horizontal_flip_bounding_box():
 
 
 def sample_inputs_horizontal_flip_mask():
-    for image_loader in make_mask_loaders(sizes=[None], dtypes=[torch.uint8]):
+    for image_loader in make_mask_loaders(sizes=["random"], dtypes=[torch.uint8]):
         yield ArgsKwargs(image_loader)
 
 
@@ -417,7 +417,7 @@ KERNEL_INFOS.append(
 
 def sample_inputs_convert_color_space_image_tensor():
     color_spaces = set(features.ColorSpace) - {features.ColorSpace.OTHER}
-    for image_loader in make_image_loaders(sizes=[None], color_spaces=color_spaces, constant_alpha=True):
+    for image_loader in make_image_loaders(sizes=["random"], color_spaces=color_spaces, constant_alpha=True):
         old_color_space = image_loader.color_space
         for params in combinations_grid(new_color_space=color_spaces - {old_color_space}, copy=(True, False)):
             yield ArgsKwargs(image_loader, old_color_space=old_color_space, **params)
--- a/test/prototype_transforms_kernel_infos.py
+++ b/test/prototype_transforms_kernel_infos.py
@@ -66,7 +66,7 @@ KERNEL_INFOS = []
 
 
 def sample_inputs_horizontal_flip_image_tensor():
-    for image_loader in make_image_loaders(dtypes=[torch.float32]):
+    for image_loader in make_image_loaders(sizes=[None], dtypes=[torch.float32]):
         yield ArgsKwargs(image_loader)
 
 
@@ -76,14 +76,16 @@ def reference_inputs_horizontal_flip_image_tensor():
 
 
 def sample_inputs_horizontal_flip_bounding_box():
-    for bounding_box_loader in make_bounding_box_loaders():
+    for bounding_box_loader in make_bounding_box_loaders(
+        formats=[features.BoundingBoxFormat.XYXY], dtypes=[torch.float32]
+    ):
         yield ArgsKwargs(
             bounding_box_loader, format=bounding_box_loader.format, image_size=bounding_box_loader.image_size
         )
 
 
 def sample_inputs_horizontal_flip_mask():
-    for image_loader in make_mask_loaders(dtypes=[torch.uint8]):
+    for image_loader in make_mask_loaders(sizes=[None], dtypes=[torch.uint8]):
         yield ArgsKwargs(image_loader)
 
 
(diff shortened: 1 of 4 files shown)


test/prototype_transforms_kernel_infos.py
```diff
@@ -62,7 +67,7 @@ def wrapper(image_tensor, *other_args, **kwargs):
 
 
 def sample_inputs_horizontal_flip_image_tensor():
-    for image_loader in make_image_loaders(dtypes=[torch.float32]):
+    for image_loader in make_image_loaders(sizes=[None], dtypes=[torch.float32]):
```
pmeier, line 70: This is an effort to keep the total number of tests moderate. Since the sample inputs should only cover all branches, we don't need to test multiple image sizes here. 

There are multiple places where I constrained the number of sample inputs below. I've checked the kernels and made sure that we cover all branches, but nothing more. The extensive tests should only happen in the reference inputs.

test/prototype_transforms_kernel_infos.py
```diff
@@ -72,21 +77,24 @@ def reference_inputs_horizontal_flip_image_tensor():
...
-    for image_loader in make_mask_loaders(dtypes=[torch.uint8]):
+    for image_loader in make_mask_loaders(sizes=[None], dtypes=[torch.uint8]):
         yield ArgsKwargs(image_loader)
 
 
 KERNEL_INFOS.extend(
     [
         KernelInfo(
             F.horizontal_flip_image_tensor,
+            kernel_name="horizontal_flip_image_tensor",
```
pmeier, line 97: Otherwise this appears as `hflip` and will mess up the collection.

test/prototype_transforms_kernel_infos.py
```diff
@@ -278,26 +323,38 @@ def transform(bbox):
...
+    for mask_loader, center in itertools.product(
+        make_mask_loaders(dtypes=[torch.uint8]),
+        [None, (0, 0)],
+    ):
+        yield ArgsKwargs(mask_loader, center=center, **_AFFINE_KWARGS[0])
+
+
+@pil_reference_wrapper
+def reference_affine_mask(*args, **kwargs):
+    return F.affine_image_pil(*args, interpolation=F.InterpolationMode.NEAREST, **kwargs)
```
pmeier, line 348: Previously we had a custom implementation for the reference (see below). @vfdev-5 was there a reason to not rely on PIL as reference here?

test/test_prototype_transforms_functional.py
```diff
@@ -653,60 +578,6 @@ def test_correctness_affine_bounding_box_on_fixed_input(device):
     torch.testing.assert_close(output_boxes.tolist(), expected_bboxes)
 
 
-@pytest.mark.parametrize("angle", [-54, 56])
-@pytest.mark.parametrize("translate", [-7, 8])
-@pytest.mark.parametrize("scale", [0.89, 1.12])
-@pytest.mark.parametrize("shear", [4])
-@pytest.mark.parametrize("center", [None, (12, 14)])
-def test_correctness_affine_mask(angle, translate, scale, shear, center):
```
pmeier, line 661: See above

test/test_prototype_transforms_functional.py
```diff
@@ -505,6 +471,41 @@ def test_eager_vs_scripted(functional_info, sample_input):
...
+                    }
+                    else []
+                )
+            ],
+        )
+        for functional_info in FUNCTIONAL_INFOS
+        for idx, sample_input in enumerate(functional_info.sample_inputs())
+    ],
+)
+def test_dtype_consistency(functional_info, sample_input):
```
pmeier, line 501: This test was also added on the old framework, because we haven't ported all kernels yet, but we need to make sure the behavior is consistent for the whole API.

test/test_prototype_transforms_kernels.py
```diff
@@ -92,7 +94,7 @@ class TestCommon:
     sample_inputs = pytest.mark.parametrize(
         ("info", "args_kwargs"),
         [
-            pytest.param(info, args_kwargs, id=f"{info.kernel.__name__}")
+            pytest.param(info, args_kwargs, id=f"{info.kernel_name}-")
```
pmeier, line 97: The extra `-` makes it more clear that the number is related to the sample input rather than to the kernel

`test_against_reference[resize_image_tensor-0]`

vs.


`test_against_reference[resize_image_tensor0]`

test/prototype_transforms_kernel_infos.py
vfdev-5, line 70: Looks strange `sizes=[None]`, what does that do ?

test/prototype_transforms_kernel_infos.py
```diff
@@ -104,23 +112,34 @@ def sample_inputs_horizontal_flip_mask():
...
+    height, width = image_size
+    yield height, width
+    yield int(height * 0.75), int(width * 1.25)
+
+
 def sample_inputs_resize_image_tensor():
     for image_loader, interpolation in itertools.product(
         make_image_loaders(dtypes=[torch.float32]),
         [
-            F.InterpolationMode.NEAREST,
```
vfdev-5, line 111: Why did you remove this mode ?

test/prototype_transforms_kernel_infos.py
vfdev-5: The reason is that it was a correctness test and not consistency. We do not call our library to produce reference output. You can imagine a situation if both methods have changed but ci still passing.

test/test_prototype_transforms_functional.py
```diff
@@ -528,82 +529,6 @@ def _compute_affine_matrix(angle_, translate_, scale_, shear_, center_):
     return true_matrix
 
 
-@pytest.mark.parametrize("angle", range(-90, 90, 56))
-@pytest.mark.parametrize("translate", range(-10, 10, 8))
-@pytest.mark.parametrize("scale", [0.77, 1.0, 1.27])
-@pytest.mark.parametrize("shear", range(-15, 15, 8))
-@pytest.mark.parametrize("center", [None, (12, 14)])
-def test_correctness_affine_bounding_box(angle, translate, scale, shear, center):
```
vfdev-5, line 536: Where do you put this correctness test in the new test framework ?

test/prototype_transforms_kernel_infos.py
pmeier: `sizes` is a sequence of image sizes, i.e. height and width of the images to be created. By default we use

https://github.com/pytorch/vision/blob/c0911e31d367eca9600541bd634564c73753abbf/test/prototype_common_utils.py#L165-L168

where `None` means "I don't care, just something small"

https://github.com/pytorch/vision/blob/c0911e31d367eca9600541bd634564c73753abbf/test/prototype_common_utils.py#L172-L173

Together, `sizes=[None]` means, we don't want to test the full matrix on square, landscape, and portrait images here, but only on random image_sizes.

test/prototype_transforms_kernel_infos.py
pmeier: Because the only branching is happening on `BICUBIC`

https://github.com/pytorch/vision/blob/c0911e31d367eca9600541bd634564c73753abbf/torchvision/transforms/functional_tensor.py#L466-L467

Note that we are talking about the sample inputs here. The full matrix with all interpolation modes is still there for the reference inputs. L127-L131

test/test_prototype_transforms_functional.py
pmeier: The sample generation is happening in `reference_affine_bounding_box` (L325) while the reference function is `reference_affine_bounding_box` (L281)

test/prototype_transforms_kernel_infos.py
vfdev-5: Sorry, I do not understand anymore how it works :)

pmeier: As discussed offline this should be nearest and bicubic.

vfdev-5 (approved):
LGTM, thanks @pmeier !

test/prototype_transforms_kernel_infos.py
pmeier: As discussed offline, Victor is technically right here. Still, since our PIL kernels, e.g. `affine_image_pil` mostly wrap the PIL functionaly and only juggle the parameters a bit, it is fine to use them directly as reference for now. We can always come back later and write proper reference tests.

PR #6597, opened 2022-09-16
retain input dtype in all prototype transforms kernels

pmeier:
This PR fixes all transforms kernels to always return the same dtype it was called with.

### DIFF
--- a/torchvision/prototype/transforms/functional/_geometry.py
+++ b/torchvision/prototype/transforms/functional/_geometry.py
@@ -829,7 +829,7 @@ def perspective_bounding_box(
     transformed_points = transformed_points.view(-1, 4, 2)
     out_bbox_mins, _ = torch.min(transformed_points, dim=1)
     out_bbox_maxs, _ = torch.max(transformed_points, dim=1)
-    out_bboxes = torch.cat([out_bbox_mins, out_bbox_maxs], dim=1)
+    out_bboxes = torch.cat([out_bbox_mins, out_bbox_maxs], dim=1).to(bounding_box.dtype)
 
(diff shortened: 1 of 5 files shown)


torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -153,7 +153,7 @@ def resize_bounding_box(
     old_height, old_width = image_size
     new_height, new_width = _compute_resized_output_size(image_size, size=size, max_size=max_size)
     ratios = torch.tensor((new_width / old_width, new_height / old_height), device=bounding_box.device)
-    return bounding_box.view(-1, 2, 2).mul(ratios).view(bounding_box.shape)
+    return bounding_box.view(-1, 2, 2).mul(ratios).to(bounding_box.dtype).view(bounding_box.shape)
```
datumbox, line 156: Optional Nit: For those of us with OCDs perhaps always do last? 😄 
```suggestion
    return bounding_box.view(-1, 2, 2).mul(ratios).view(bounding_box.shape).to(bounding_box.dtype)
```

datumbox (approved):

torchvision/prototype/transforms/functional/_geometry.py
pmeier: [Ackchyually](https://knowyourmeme.com/memes/ackchyually-actually-guy) this is consistent. The dtype conversion always happens before the last `view`:

- `resize_bounding_box` as well as the format conversion functions have no final `.view()` call
- `{affine, perspective, elastic}_bounding_box` have it on the line below the change

I thought about that when I was writing the patch :grin:

PR #6598, opened 2022-09-16
[draft] Pyav backend for VideoReader API

bjuncek:
Draft implementation for the VideoReader API. 
This would allow the usage of the new API without the necessity to have own FFMPEG installation, but instead would only rely on having PyAV installed.

### DIFF
--- a/test/test_video_gpu_decoder.py
+++ b/test/test_video_gpu_decoder.py
@@ -3,6 +3,7 @@ import os
 
 import pytest
 import torch
+import torchvision
 from torchvision.io import _HAS_GPU_VIDEO_DECODER, VideoReader
 
 try:
@@ -29,8 +30,9 @@ class TestVideoGPUDecoder:
         ],
     )
     def test_frame_reading(self, video_file):
+        torchvision.set_video_backend("cuda")
         full_path = os.path.join(VIDEO_DIR, video_file)
-        decoder = VideoReader(full_path, device="cuda")
+        decoder = VideoReader(full_path)
         with av.open(full_path) as container:
             for av_frame in container.decode(container.streams.video[0]):
                 av_frames = torch.tensor(av_frame.to_rgb(src_colorspace="ITU709").to_ndarray())
@@ -54,7 +56,8 @@ class TestVideoGPUDecoder:
         ],
     )
     def test_seek_reading(self, keyframes, full_path, duration):
-        decoder = VideoReader(full_path, device="cuda")
+        torchvision.set_video_backend("cuda")
+        decoder = VideoReader(full_path)
         time = duration / 2
         decoder.seek(time, keyframes_only=keyframes)
         with av.open(full_path) as container:
@@ -79,8 +82,9 @@ class TestVideoGPUDecoder:
         ],
     )
     def test_metadata(self, video_file):
+        torchvision.set_video_backend("cuda")
         full_path = os.path.join(VIDEO_DIR, video_file)
-        decoder = VideoReader(full_path, device="cuda")
+        decoder = VideoReader(full_path)
         video_metadata = decoder.get_metadata()["video"]
         with av.open(full_path) as container:
             video = container.streams.video[0]
--- a/torchvision/__init__.py
+++ b/torchvision/__init__.py
@@ -1,15 +1,23 @@
+from modulefinder import Module
 import os
 import warnings
 
 import torch
 from torchvision import datasets, io, models, ops, transforms, utils
 
-from .extension import _HAS_OPS
+from .extension import _HAS_OPS, _load_library
 
 try:
     from .version import __version__  # noqa: F401
 except ImportError:
     pass
+
+try:
+    _load_library("Decoder")
+    _HAS_GPU_VIDEO_DECODER = True
+except (ImportError, OSError, ModuleNotFoundError):
+    _HAS_GPU_VIDEO_DECODER = False
+
 
 # Check if torchvision is being imported within the root folder
 if not _HAS_OPS and os.path.dirname(os.path.realpath(__file__)) == os.path.join(
@@ -66,11 +74,16 @@ def set_video_backend(backend):
         backend, please compile torchvision from source.
     """
     global _video_backend
-    if backend not in ["pyav", "video_reader"]:
-        raise ValueError("Invalid video backend '%s'. Options are 'pyav' and 'video_reader'" % backend)
+    if backend not in ["pyav", "video_reader", "cuda"]:
+        raise ValueError("Invalid video backend '%s'. Options are 'pyav', 'video_reader' and 'cuda'" % backend)
     if backend == "video_reader" and not io._HAS_VIDEO_OPT:
+        #TODO: better messages
         message = "video_reader video backend is not available. Please compile torchvision from source and try again"
-        warnings.warn(message)
+        raise RuntimeError(message)
+    elif backend == "cuda" and not _HAS_GPU_VIDEO_DECODER:
+        #TODO: better messages
+        message = "cuda video backend is not available."
+        raise RuntimeError(message)
     else:
         _video_backend = backend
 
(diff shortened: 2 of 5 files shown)


test/test_video_gpu_decoder.py
```diff
@@ -79,8 +82,9 @@ def test_seek_reading(self, keyframes, full_path, duration):
         ],
     )
     def test_metadata(self, video_file):
+        torchvision.set_video_backend("cuda")
```
jdsgomes, line 85: can we run this just once for all tests in this file since they all supposed to run with the cud backend?

torchvision/__init__.py
```diff
@@ -1,16 +1,24 @@
...
-from .extension import _HAS_OPS
+from .extension import _HAS_OPS, _load_library
 
 try:
     from .version import __version__  # noqa: F401
 except ImportError:
     pass
 
+try:
+    _load_library("Decoder")
```
jdsgomes, line 16: Do you know why we didn't have this previously?

jdsgomes: I see you moved it from io

torchvision/io/video_reader.py
```diff
@@ -21,11 +18,37 @@ def _has_video_opt() -> bool:
         return False
 
 
+try:
```
jdsgomes, line 21: I wonder if this should be when we `set_video_backend`

torchvision/io/video_reader.py
```diff
@@ -82,30 +105,34 @@ class VideoReader:
...
-    def __init__(self, path: str, stream: str = "video", num_threads: int = 0, device: str = "cpu") -> None:
+    def __init__(self, path: str, stream: str = "video", num_threads: int = 0) -> None:
         _log_api_usage_once(self)
-        self.is_cuda = False
-        device = torch.device(device)
-        if device.type == "cuda":
-            if not _HAS_GPU_VIDEO_DECODER:
-                raise RuntimeError("Not compiled with GPU decoder support.")
-            self.is_cuda = True
+        from .. import get_video_backend
```
jdsgomes, line 112: I think its better to import at the module level

torchvision/io/video_reader.py
```diff
@@ -82,30 +105,34 @@ class VideoReader:
...
+
+        self.backend = get_video_backend()
+        print("Initiated the backend", self.backend)
```
jdsgomes, line 115: if we really need this message I would add something like `f"VideoReader initiated with {self.backend} backend"`

torchvision/io/video_reader.py
```diff
@@ -82,30 +105,34 @@ class VideoReader:
...
+        if self.backend == "cuda":
+            device = torch.device("cuda")
             self._c = torch.classes.torchvision.GPUDecoder(path, device)
             return
-        if not _has_video_opt():
```
jdsgomes, line 100: we still need this inside the `elif self.backend == "video_reader"` bellow right?

torchvision/io/video_reader.py
```diff
@@ -119,14 +146,29 @@ def __next__(self) -> Dict[str, Any]:
             and corresponding timestamp (``pts``) in seconds
 
         """
-        if self.is_cuda:
+        if self.backend == "cuda":
             frame = self._c.next()
             if frame.numel() == 0:
                 raise StopIteration
             return {"data": frame}
```
jdsgomes, line 153: I wonder if we should just return None as pts until we fix the cuda side, don't have strong opinion but seems more consistent to return a dictionary with the same keys for all backends

torchvision/io/video_reader.py
```diff
@@ -145,7 +187,18 @@ def seek(self, time_s: float, keyframes_only: bool = False) -> "VideoReader":
...
+        if self.backend in ["cuda", "video_reader"]:
+            self._c.seek(time_s, keyframes_only)
+        else:
+            # handle special case as pyav doesn't catch it
+            if time_s < 0:
+                time_s = 0
+            temp_str = self.container.streams.get(**self.pyav_stream)[0]
+            offset = int(round(time_s / temp_str.time_base))
+            if not keyframes_only:
+                warnings.warn("Accurate seek is not implemented for pyav backend")
```
jdsgomes, line 199: Not related to this PR, but do you have plans to add accurate seek in PyAV?

torchvision/io/video_reader.py
```diff
@@ -173,6 +237,6 @@ def set_current_stream(self, stream: str) -> bool:
         Returns:
             (bool): True on succes, False otherwise
         """
-        if self.is_cuda:
+        if self.backend == "cuda":
             print("GPU decoding only works with video stream.")
```
jdsgomes, line 241: I think this should be changed to a warning. But can be done on a different PR as it is out of scope

jdsgomes (commented):
Thanks for working on it. In general looks good, so I just left a few comments

datumbox:
FYI @YosuaMichael will be cutting the release branch on Monday. Is your intention to try to merge prior?

bjuncek:
@datumbox let's not do it, it is missing accurate seek so it would make more sense to do it for the next one (and hopefully sort some other woes along the way).

datumbox:
@bjuncek Agreed. Makes sense to me to delay to avoid release issues.

torchvision/io/video_reader.py
bjuncek: Probably a good idea, yeah.

torchvision/io/video_reader.py
bjuncek: Yeah, but in a roundabout way.
(I need to do it manually)

torchvision/__init__.py
bjuncek: yup

torchvision/io/video_reader.py
bjuncek: We could; I have no strong preference.

torchvision/io/video_reader.py
bjuncek: Yes, but this causes circular imports as `get_video_backend` is defined at the root level and it cannot instantiate the module prior to loading all submodules. 
This is a workaround

torchvision/io/video_reader.py
bjuncek: We don't, it was for debugging purposes only. Good catch

torchvision/io/video_reader.py
bjuncek: Done

test/test_video_gpu_decoder.py
bjuncek: Good question. Let me try out :)

torchvision/io/video_reader.py
jdsgomes: I think it makes more sense as this is only relevant if we set backend to pyav

torchvision/io/video_reader.py
```diff
@@ -154,6 +206,17 @@ def get_metadata(self) -> Dict[str, Any]:
...
+        if self.backend == "pyav":
+            metadata = {}  # type:  Dict[str, Any]
+            for stream in self.container.streams:
+                if stream.type not in metadata:
+                    metadata[stream.type] = {"fps": [], "duration": []}
+
+                rate = stream.average_rate if stream.average_rate is not None else stream.sample_rate
+
+                metadata[stream.type]["duration"].append(float(stream.duration * stream.time_base))
+                metadata[stream.type]["fps"].append(float(rate))
```
jdsgomes, line 218: for audio we use framerate instead of fps, I think this is the reason build_docs is failing

test/test_video_gpu_decoder.py
bjuncek: For some reason that failed (putting it in a class header) so I'll keep it as is, and then re-do the tests later.

bjuncek:
Hi @jdsgomes .

If this looks ok, let's try to merge it. 
I'll add the precise seek in later PR to avoid the clutter

jdsgomes (approved):
LGTM I agree that precise seek can be added in a follow up PR

jdsgomes:
Checked the latest changes looks good to me.

PR #6599, opened 2022-09-16
[proto] Added dict support for fill arg for remaining transforms

vfdev-5:
Description:
- fill arg can be int/float/sequence or dict of that.
- Added support for other ops

Blocked by https://github.com/pytorch/vision/pull/6595


datumbox:
It should be merged prior continuing with https://github.com/pytorch/vision/pull/6584.

pmeier (approved):
LGTM when CI is green. Thanks Victor!

PR #6605, opened 2022-09-19
Add stereo train loop

TeodorPoncu:
This is the proposed stereo matching reference script. The logger, distributed, padder utils are taken from our optical flow reference scripts. 

The losses and transforms are included separately in two previous PRs. (#6554 and #6549)


references/depth/stereo/train.py
```diff
@@ -0,0 +1,751 @@
+import argparse
+from email.policy import default
```
vfdev-5, line 2:
```suggestion
```

references/depth/stereo/cascade_evaluation.py
```diff
@@ -0,0 +1,242 @@
...
+    
+    effective_batch_size = batch_size * args.world_size
+    can_do_distributed_test = effective_batch_size < len(val_dataset)
+    
+    if args.distributed and can_do_distributed_test:
+        sampler = torch.utils.data.distributed.DistributedSampler(val_dataset, shuffle=False, drop_last=True)
+    else:
+        sampler = torch.utils.data.SequentialSampler(val_dataset)
+    
+    val_loader = torch.utils.data.DataLoader(
```
YosuaMichael, line 86: Why do we create `val_loader` inside `_evaluate` functions? Is there any reason we choose to do this instead of passing `val_loader`?

(same question in train.py)

references/depth/stereo/parsing.py
```diff
@@ -0,0 +1,122 @@
...
+        "middlebury2014-other",
+        "middlebury2014-train",
+        "middlebury2014-test",
+        "middlebury2014-train-ambient",
+    ]
+
+    if dataset_name not in valid_datasets:
+        raise ValueError(f"Invalid dataset name: {dataset_name}. Valid datasets are: {valid_datasets}")
+
+    if dataset_name == "crestereo":
```
YosuaMichael, line 86: NIT, we might want to consider using `dict[str, Dataset]` instead of a series of if-elif-else. But no strong opinion on this.

references/depth/stereo/train.py
```diff
@@ -0,0 +1,750 @@
...
+    
+    effective_batch_size = batch_size * args.world_size
+    can_do_distributed_test = effective_batch_size < len(val_dataset)
+    
+    if args.distributed and can_do_distributed_test:
+        sampler = torch.utils.data.distributed.DistributedSampler(val_dataset, shuffle=False, drop_last=True)
+    else:
+        sampler = torch.utils.data.SequentialSampler(val_dataset)
+    
+    val_loader = torch.utils.data.DataLoader(
```
YosuaMichael, line 184: same question in cascade_evaluation, why do we define val_loader inside `_evaluate` function instead of passing `val_loader` and define it outside?

references/depth/stereo/train.py
```diff
@@ -0,0 +1,750 @@
...
+    model.eval()
+    header = header or "Test:"
+    device = torch.device(args.device)
+    metric_logger = utils.MetricLogger(delimiter="  ")
+    
+    effective_batch_size = batch_size * args.world_size
+    can_do_distributed_test = effective_batch_size < len(val_dataset)
+    
+    if args.distributed and can_do_distributed_test:
+        sampler = torch.utils.data.distributed.DistributedSampler(val_dataset, shuffle=False, drop_last=True)
```
YosuaMichael, line 180: question, why we use `drop_last=True`?

references/depth/stereo/cascade_evaluation.py
```diff
@@ -0,0 +1,242 @@
...
+        # which will not be fulfilled by non rank 0 processes
+        module = model.module
+
+    num_processed_samples = 0
+    with torch.cuda.amp.autocast(enabled=args.mixed_precision, dtype=torch.float16):
+        if should_forward:
+            batch_idx = 0
+            for blob in metric_logger.log_every(val_loader, print_freq, header):
+                image_left, image_right, disp_gt, valid_disp_mask = (x.to(device) for x in blob)
+                padder = utils.InputPadder(image_left.shape, mode=padder_mode)
```
YosuaMichael, line 117: Question, can we put this `InputPadder` as part of transform (passed to dataset)?

references/depth/stereo/cascade_evaluation.py
```diff
@@ -0,0 +1,242 @@
...
+    checkpoint = torch.load(args.checkpoint, map_location=torch.device("cpu"))
+    experiment_args = checkpoint["args"]
+    model = torchvision.prototype.models.depth.stereo.__dict__[experiment_args.model](weights=None)
+    model.load_state_dict(checkpoint["model"])
+    return model
+
+    
+def main(args):
+    model = load_checkpoint(args)
+    evaluate(model, args, padder_mode="kitti")
```
YosuaMichael, line 237: Do padder_mode always be `kitti`?

references/depth/stereo/train.py
```diff
@@ -0,0 +1,750 @@
...
+    if args.weights and args.test_only:
+        weights = torchvision.models.get_weight(args.weights)
+        trans = weights.transforms()
+
+        def preprocessing(image_left, image_right, disp, valid_disp_mask):
+            image_left, image_right = trans(image_left, image_right)
+            if disp is not None and not isinstance(disp, torch.Tensor):
+                disp = torch.from_numpy(disp)
+            if valid_disp_mask is not None and not isinstance(valid_disp_mask, torch.Tensor):
+                valid_disp_mask = torch.from_numpy(valid_disp_mask)
```
YosuaMichael, line 257: I think we should handle the case where the image is resized, i.e. need to resize disp and valid_disp_mask as well.

references/depth/stereo/train.py
```diff
@@ -0,0 +1,750 @@
...
+        preprocessing = make_eval_transform(args)
+
+    for name in val_datasets:
+        val_dataset = make_dataset(name, args.dataset_root, transforms=preprocessing)
+        _evaluate(
+            model,
+            args,
+            val_dataset,
+            iterations=args.recurrent_updates,
+            padder_mode="kitti",
```
YosuaMichael, line 270: Same question in cascade inference, do we always use `padder_mode="kitti"` ?

references/depth/stereo/train.py
```diff
@@ -0,0 +1,750 @@
...
+            header=f"{name} evaluation",
+            batch_size=args.batch_size,
+            writter=writter,
+            step=step,
+        )
+
+
+def run(model, optimizer, scheduler, train_loader, logger, writer, scaler, args):
```
YosuaMichael, line 278: NIT, instead of run, maybe we should call it `train` instead

references/depth/stereo/train.py
```diff
@@ -0,0 +1,750 @@
...
+
+        with torch.no_grad():
+            metrics, _ = utils.compute_metrics(
+                disp_predictions[-1][:, :1, :, :], # predictions might have 2 channels
+                disp_mask[:, :1, :, :], # so does the ground truth
+                valid_disp_mask,
+                args.metrics
+            )
+        
+        metrics.pop("f1", None)
```
YosuaMichael, line 386: Why do we pop "f1" specifically here?

references/depth/stereo/train.py
```diff
@@ -0,0 +1,750 @@
...
+                torch.save(checkpoint, Path(args.checkpoint_dir) / f"{args.name}.pth")
+
+        if step % args.valid_frequency == 0:
+            evaluate(model, args, writer, step)
+            model.train()
+            if args.freeze_batch_norm:
+                if isinstance(model, nn.parallel.DistributedDataParallel):
+                    freeze_batch_norm(model.module)
+                else:
+                    freeze_batch_norm(model)
```
YosuaMichael, line 443: We should save the model once the training finished as well (after exiting the for loop)

references/depth/stereo/train.py
```diff
@@ -0,0 +1,750 @@
...
+        raise ValueError("The device must be cuda if we want to run in distributed mode using torchrun")
+    device = torch.device(args.device)
+
+    # select model architecture
+    model = torchvision.prototype.models.depth.stereo.__dict__[args.model](weights=args.weights)
+
+    # convert to DDP if need be
+    if args.distributed:
+        model = model.to(args.local_rank)
+        model = torch.nn.parallel.DistributedDataParallel(model, device_ids=[args.local_rank])
```
YosuaMichael, line 464: NIT, lets follow similar pattern to [train.py in classification](https://github.com/pytorch/vision/blob/main/references/classification/train.py#L301), where we have `args.gpu` that defined in [init_distributed_mode function](https://github.com/pytorch/vision/blob/main/references/classification/utils.py#L245).

references/depth/stereo/train.py
```diff
@@ -0,0 +1,750 @@
...
+    import argparse
+
+    parser = argparse.ArgumentParser(description="PyTorch Stereo Matching Training", add_help=add_help)
+    # checkpointing
+    parser.add_argument("--name", default="crestereo", help="name of the experiment")
+    parser.add_argument("--resume", type=str, default=None, help="from which checkpoint to resume")
+    parser.add_argument("--checkpoint-dir", type=str, default="checkpoints", help="path to the checkpoint directory")
+
+    # dataset
+    parser.add_argument("--dataset-root", type=str, default="/fsx/users/teodorponcu/datasets", help="path to the dataset root directory")
```
YosuaMichael, line 571: NIT, Avoid having your name in the file path.

YosuaMichael (commented):
Hi @TeodorPoncu , thanks for the PR! Here are some comments, let me know what you think!

references/depth/stereo/cascade_evaluation.py
TeodorPoncu: Because we might want to consume the specific dataset multiple times in the cascaded inference, therefore that's done to avoid any bugs where the dataloader might not be fully consumed and just be partially iterated over.

references/depth/stereo/train.py
TeodorPoncu: My only concern with calling it train was that in contains a call to the evaluation loop as well.

references/depth/stereo/train.py
TeodorPoncu: "f1" is not necessarily a good metric during training time, because it's based on the relative epe. Depending on the batch disparity distribution the relative might not give a good representation of how it actually transfers to the evaluation part as well.

references/depth/stereo/parsing.py
TeodorPoncu: I agree, that is a better alternative.

references/depth/stereo/cascade_evaluation.py
TeodorPoncu: That would unfortunately not work since we have to be able to unpad the out as well after we have passed through the model. I believe another alternative would be to bind the padding utility to the model forward pass somehow.

references/depth/stereo/train.py
TeodorPoncu: In order to avoid strange behaviours when the test size is sufficiently large to fit on a step of DDP, but not large enough to be distributed to all worker nodes.

references/depth/stereo/train.py
TeodorPoncu: In that case we would have to move into the `_presets.py` transform to accept optional disparity masks just in case. It still feels odd for me to have presets that are destined to be input only to handle that. 

My suggestion would  be to compute the image size / shape before and after and just scale it inside this function.

references/depth/stereo/train.py
YosuaMichael: Hmm... if the case that you mention happen, isn't drop last gonna drop the whole dataset in that case?
Also, I think you already try to handle this with `can_do_distributed_test` variable, do we need both `drop_last=True` and `can_do_distributed_test` to handle such case?

TeodorPoncu: For instance, you have an effective batch_size of 10 and a dataset size of 11. You're going to process the first batch, which will leave you with 1 sample remaining. That sample will be skipped.

`can_do_distributed_test` checks wether or not the dataset size is smaller than the effective batch_size.

references/depth/stereo/cascade_evaluation.py
YosuaMichael: I think for your alternative, if we can add the padding on the model without specifying `padder_mode` then that should be fine. But if we need to specify `padder_mode`, then I think it is better to put it here as it is now

references/depth/stereo/cascade_evaluation.py
YosuaMichael: If you think always using Kitti is fine, in this case I think we should remove `padder_mode` and always use the same padding mode without the need of input param.
Otherwise if we need different padder_mode depend on dataset, then I think we should add extra param in argparse.

references/depth/stereo/cascade_evaluation.py
TeodorPoncu: One way to avoid specifying `padder_mode`  would be to have a default variable in the model constructors, and include a setter for them in case the user wants to change the padder mode after instantiating the model.

references/depth/stereo/train.py
YosuaMichael: Although I agree that this may not be the best approach, as of now my recommendation is to follow the train script in segmentation: https://github.com/pytorch/vision/blob/main/references/segmentation/train.py#L42

Basically manually transform the mask and disparity after knowing the image size.

references/depth/stereo/cascade_evaluation.py
YosuaMichael: Lets discuss about this tomorrow, I didn't really get what you mean here

references/depth/stereo/train.py
YosuaMichael: From the docs: https://pytorch.org/docs/stable/data.html#torch.utils.data.distributed.DistributedSampler
Seems like if we set `drop_last=False` it will add extra indices to make sure the data is evenly divisible across replica.
So I think both have their own issue (not using the whole test data, or having duplicate test data).
In my personal preference though, I prefer to have duplicate test data compared to dropping some of the test data.

references/depth/stereo/cascade_evaluation.py
YosuaMichael: Lets keep it as it is now, since if we add it to the models it also meant we need to add this padding behaviour to all stereo matching models

references/depth/stereo/train.py
```diff
@@ -0,0 +1,754 @@
...
+        def preprocessing(image_left, image_right, disp, valid_disp_mask):
+            C_o, H_o, W_o = get_dimensions(image_left)        
+            image_left, image_right = trans(image_left, image_right)
+            
+            C_t, H_t, W_t = get_dimensions(image_left)
+            scale_factor = W_t / W_o
+            
+            if disp is not None and not isinstance(disp, torch.Tensor):                
+                disp = torch.from_numpy(disp)
+                if scale_factor != 1:
```
YosuaMichael, line 222: Since `scale_factor` is a float, I think this comparison `scale_factor != 1` is risky.
I recommend: `if W_t != W_o`

references/depth/stereo/train.py
```diff
@@ -254,12 +232,28 @@ def preprocessing(image_left, image_right, disp, valid_disp_mask):
 
     for name in val_datasets:
         val_dataset = make_dataset(name, args.dataset_root, transforms=preprocessing)
+        effective_batch_size = args.batch_size * args.world_size
+        can_do_distributed_test = effective_batch_size < len(val_dataset)
```
YosuaMichael, line 236: Can you try if there is error if we dont use `can_do_distributed_test`?

references/depth/stereo/train.py
```diff
@@ -254,12 +232,28 @@ def preprocessing(image_left, image_right, disp, valid_disp_mask):
...
+        
+        if args.distributed and can_do_distributed_test:
+            sampler = torch.utils.data.distributed.DistributedSampler(val_dataset, shuffle=False, drop_last=False)
+        else:
+            sampler = torch.utils.data.SequentialSampler(val_dataset)
+        
+        val_loader = torch.utils.data.DataLoader(
```
YosuaMichael, line 243: NIT, can we move `val_loader` creation to the `main` function? Similar to `train_loader` (but should be before the first evaluate)

references/depth/stereo/cascade_evaluation.py
```diff
@@ -182,6 +157,23 @@ def preprocessing(image_left, image_right, disp, valid_disp_mask):
...
+        
+        effective_batch_size = args.batch_size * args.world_size
+        can_do_distributed_test = effective_batch_size < len(val_dataset)
+        
+        if args.distributed and can_do_distributed_test:
+            sampler = torch.utils.data.distributed.DistributedSampler(val_dataset, shuffle=False, drop_last=False)
+        else:
+            sampler = torch.utils.data.SequentialSampler(val_dataset)
+        
+        val_loader = torch.utils.data.DataLoader(
```
YosuaMichael, line 169: NIT, Similar with my comment in train.py, we should consider moving val_loader creation to main

YosuaMichael (commented):
Few NITs

references/depth/stereo/cascade_evaluation.py
TeodorPoncu: @YosuaMichael it could be moved, only observation is that it should be converted to a dictionary of loaders in order to still know which datasets is being handled during evaluation in order to correctly log it

.gitignore
```diff
@@ -42,3 +42,6 @@ xcuserdata/
 # direnv
 .direnv
 .envrc
+
+slurm-*
+*checkpoint
```
jdsgomes, line 47:
```suggestion
*checkpoint

```

references/depth/stereo/cascade_evaluation.py
```diff
@@ -0,0 +1,234 @@
...
+import torchvision.prototype.models.depth.stereo
+import warnings
+from vizualization import make_prediction_image_side_to_side
+
+def get_args_parser(add_help=True):
+    import argparse
+    
+    parser = argparse.ArgumentParser(description="PyTorch Stereo Matching Evaluation", add_help=add_help)
+    parser.add_argument("--dataset", type=str, default="middlebury2014-train", help="dataset to use")
+    parser.add_argument("--dataset-root", type=str, default="/fsx/users/teodorponcu/datasets", help="root of the dataset")
```
jdsgomes, line 16: ups :) your path is hardcoded here

references/depth/stereo/cascade_evaluation.py
```diff
@@ -0,0 +1,234 @@
...
+    checkpoint_name = os.path.basename(args.checkpoint)
+    image_checkpoint_folder = os.path.join(args.img_folder, checkpoint_name)
+
+    if args.weights and args.test_only:
+        weights = torchvision.models.get_weight(args.weights)
+        trans = weights.transforms()
+
+        def preprocessing(image_left, image_right, disp, valid_disp_mask):
+            image_left, image_right = trans(image_left, image_right)
+            if disp is not None and not isinstance(disp, torch.Tensor):
```
jdsgomes, line 144: I might have missed this, but why do we need this special casting here?

jdsgomes (commented):
another round. I will try to finish the entire review today

references/depth/stereo/cascade_evaluation.py
```diff
@@ -0,0 +1,234 @@
...
+    
+    num_processed_samples = utils.reduce_across_processes(num_processed_samples) / args.world_size 
+    
+    print("Num_processed_samples: ", num_processed_samples)
+    if (
+        hasattr(val_loader.dataset, "__len__")
+        and len(val_loader.dataset) != num_processed_samples
+        and torch.distributed.get_rank() == 0
+    ):
+        warnings.warn(
```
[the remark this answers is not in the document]
TeodorPoncu: I've discussed with Yosua and have chosen to follow the reference in the classification script. If we want to change back to the optical flow route, we could do that.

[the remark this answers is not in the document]
jdsgomes: We can leave it as it then.

jdsgomes (commented):
Thanks for doing this. In general looks good so I don't have further comments. 
In order to merge there are only a few things left
- a few pending comments from Yosua
- several newlines missing 
- lint errors

references/depth/stereo/cascade_evaluation.py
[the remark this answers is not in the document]
YosuaMichael: Ah yes, just to add context that this is a design decision between `drop_last=True` where we remove some of the eval data vs `drop_last=False` where in this case the data loader will add indices to make sure the data is divisible by the number of GPU and cause duplicate data.

Both case are not ideal in distributed setting (one have duplicate data, one have missing data). But personally I prefer to have duplicate data in this case since the code is simpler (like the classification example). And I think this is more like design choice since both option is not ideal.

references/depth/stereo/train.py
```diff
@@ -0,0 +1,801 @@
...
+            "crestereo",
+            "eth3d-train",
+            "middlebury2014-train-ambient",
+            "middlebury2014-other",
+            "instereo2k",
+            "fallingthings",
+            "carla-highres",
+            "sintel",
+            "sceneflow-monkaa",
+            "sceneflow-driving",
```
YosuaMichael, line 605: NIT, can we add sceneflow flyingthings?

YosuaMichael (approved):
Thanks for your hard work @TeodorPoncu, this PR looks good! Just one minor nit from me.
I will approve first to unblock.

jdsgomes (approved):
LGTM. Thanks for the great contribution.

PR #6606, opened 2022-09-19
add JIT script tests for prototype functional dispatchers

pmeier:
Per title. I've put the test in `test_prototype_transforms_dispatchers.py`. This should be merged with `test_prototype_transforms_kernels.py` into `test_prototype_transforms_functional` after we have ported everything to the new framework.

### DIFF
--- a/test/common_utils.py
+++ b/test/common_utils.py
@@ -202,3 +202,47 @@ def _test_fn_on_batch(batch_tensors, fn, scripted_fn_atol=1e-8, **fn_kwargs):
         # scriptable function test
         s_transformed_batch = scripted_fn(batch_tensors, **fn_kwargs)
         torch.testing.assert_close(transformed_batch, s_transformed_batch, rtol=1e-5, atol=scripted_fn_atol)
+
+
+def cache(fn):
+    """Similar to :func:`functools.cache` (Python >= 3.8) or :func:`functools.lru_cache` with infinite buffer size,
+    but also caches exceptions.
+
+    .. warning::
+
+        Only use this on deterministic functions.
+    """
+    sentinel = object()
+    out_cache = {}
+    exc_cache = {}
+
+    @functools.wraps(fn)
+    def wrapper(*args, **kwargs):
+        key = args + tuple(kwargs.values())
+
+        out = out_cache.get(key, sentinel)
+        if out is not sentinel:
+            return out
+
+        exc = exc_cache.get(key, sentinel)
+        if exc is not sentinel:
+            raise exc
+
+        try:
+            out = fn(*args, **kwargs)
+        except Exception as exc:
+            exc_cache[key] = exc
+            raise exc
+
+        out_cache[key] = out
+        return out
+
+    return wrapper
+
+
+@cache
+def script(fn):
+    try:
+        return torch.jit.script(fn)
+    except Exception as error:
+        raise AssertionError(f"Trying to `torch.jit.script` '{fn.__name__}' raised the error above.") from error
--- a/test/prototype_transforms_dispatcher_infos.py
+++ b/test/prototype_transforms_dispatcher_infos.py
@@ -2,16 +2,36 @@ import dataclasses
 from typing import Callable, Dict, Type
 
 import pytest
+import torch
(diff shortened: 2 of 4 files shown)


test/common_utils.py
```diff
@@ -202,3 +202,47 @@ def _test_fn_on_batch(batch_tensors, fn, scripted_fn_atol=1e-8, **fn_kwargs):
...
+            raise exc
+
+        out_cache[key] = out
+        return out
+
+    return wrapper
+
+
+@cache
+def script(fn):
```
pmeier, line 244: On my local machine this made quite a difference (~2-3x) especially in the error case. Since we currently xfailing everything, I've included this here.

test/prototype_transforms_dispatcher_infos.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+KERNEL_SAMPLE_INPUTS_FN_MAP.update(
+    {
+        info.functional: lambda: (
+            PreloadedArgsKwargs(*args_kwargs.args, **args_kwargs.kwargs) for args_kwargs in info.sample_inputs()
+        )
+        for info in FUNCTIONAL_INFOS
+    }
+)
```
pmeier, line 34: This allows us to use the `FunctionalInfo`'s from the old framework for dispatcher tests. This update as well as the helper class above can be removed after everything has been ported to the new framework.

test/prototype_transforms_dispatcher_infos.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+DISPATCHER_INFOS = [
+    DispatcherInfo(
+        F.horizontal_flip,
+        kernels={
+            features.Image: F.horizontal_flip_image_tensor,
+            features.BoundingBox: F.horizontal_flip_bounding_box,
+            features.Mask: F.horizontal_flip_mask,
+        },
```
pmeier, line 57: Given that we are using a strict naming scheme, we could later also try to fill this dictionary automatically. For now let's be explicit.

test/test_prototype_transforms_dispatchers.py
```diff
@@ -0,0 +1,27 @@
+import pytest
+
+import torch.jit
+from common_utils import cpu_and_gpu, script
+from prototype_transforms_dispatcher_infos import DISPATCHER_INFOS
+from torchvision.prototype import features
+
+
+class TestCommon:
```
pmeier, line 9: I will add tests for the correct dispatch in a follow-up PR.

test/test_prototype_transforms_kernels.py
```diff
@@ -104,10 +104,7 @@ class TestCommon:
     @pytest.mark.parametrize("device", cpu_and_gpu())
     def test_scripted_vs_eager(self, info, args_kwargs, device):
         kernel_eager = info.kernel
-        try:
-            kernel_scripted = torch.jit.script(kernel_eager)
-        except Exception as error:
-            raise AssertionError("Trying to `torch.jit.script` the kernel raised the error above.") from error
+        kernel_scripted = script(kernel_eager)
```
pmeier, line 107: If we already have a cached scripting function, we can also use it here although I locally measured no real difference.

test/common_utils.py
```diff
@@ -202,3 +202,47 @@ def _test_fn_on_batch(batch_tensors, fn, scripted_fn_atol=1e-8, **fn_kwargs):
...
+    try:
+        return torch.jit.script(fn)
+    except Exception as error:
+        raise AssertionError(f"Trying to `torch.jit.script` '{fn.__name__}' raised the error above.") from error
```
vfdev-5, line 248: @pmeier do we really need to cache scripted functions ?

vfdev-5 (dismissed):
Looks ok to me, let's merge

test/common_utils.py
pmeier: See https://github.com/pytorch/vision/pull/6606#discussion_r974211565. Let's re-evaluate if we can actually test things rather than xfailing in the first step.

vfdev-5 (approved):
OK, let's make it merged with cache and 10 samples :)

PR #6608, opened 2022-09-19
[DONT MERGE] Test detection model with real weight and image

YosuaMichael:
I did experiment and try to test detection model with real weight and image.
This PR is meant for experiment on the CI and not meant to be merged.


test/test_models.py
```diff
@@ -723,19 +729,31 @@ def check_out(out):
...
+    # kwargs = {**defaults, **_model_params.get(model_name, {})}
+    # input_shape = kwargs.pop("input_shape")
 
-    model = model_fn(**kwargs)
+    # model = model_fn(**kwargs)
+
+    # Use real weight
+    # Note that we use DEFAULT and it meant we need to update expected file
+    # every time we change the default weight
+    weights = models.get_model_weights(model_name).DEFAULT
```
datumbox, line 746: FYI you don't need the name here. You can just pass the `model_fn` too

datumbox:
@YosuaMichael from [this](https://app.circleci.com/pipelines/github/pytorch/vision/20343/workflows/bfb598f5-4369-4911-a44f-e995e1656bb1/jobs/1654645), I see:
```
_________________ test_detection_model[cuda-fcos_resnet50_fpn] _________________
  File "/home/circleci/project/test/test_models.py", line 796, in check_out
  File "/home/circleci/project/test/test_models.py", line 124, in _assert_expected
  File "/home/circleci/project/env/lib/python3.8/site-packages/torch/testing/_comparison.py", line 1342, in assert_close
    assert_equal(
  File "/home/circleci/project/env/lib/python3.8/site-packages/torch/testing/_comparison.py", line 1093, in assert_equal
AssertionError: The values for attribute 'shape' do not match: torch.Size([15, 4]) != torch.Size([14, 4]).
```

This clearly shows that minor differences on the kernel implementations of PyTorch actually lead to slightly different results across architectures. I was really hoping that using real data and real weights would reduce the issues but that doesn't seem to be the case. We should definitely try increasing the nms score to lower the number of low-quality bboxes and potentially focus on the top X boxes (as you proposed offline) to see if you get more stable results. If we can't fix the flakiness this way, we might need to review the testing strategy. Happy to chat more.

YosuaMichael:
NOTE: I also include the flaky autocast on the commit: 
Fix increase score_thresh for FCOS

Issue #6224, opened 2022-06-30
Implement AutoAugment for Detection


lezwon:
### 🚀 The feature

Implement [Learning Data Augmentation Strategies for Object Detection](https://arxiv.org/abs/1906.11172)
Refers to: #3817 

### Motivation, pitch

Good to have augmentation in Torchvision

lezwon:
@datumbox @vfdev-5 Based on our last conversation I suppose it would be best to start with the `bbox_only_*` augmentations for AA right? I could get started with implementing these in `torchvision.prototype`. Do let me know if that's alright.

1. Rotate_Only_BBoxes
2. ShearX_Only_BBoxes
3. ShearY_Only_BBoxes
4. TranslateX_Only_BBoxes
5. TranslateY_Only_BBoxes
6. Flip_Only_BBoxes
7. Solarize_Only_BBoxes
8. Equalize_Only_BBoxes
9. Cutout_Only_BBoxes

datumbox:
@lezwon thanks a lot! It's worth coordinating with @vfdev-5 because he is literally landing changes at the same API.

Victor, how do you propose to do this? Would you prefer to do an implementation on References (similar to what we did with CopyPaste) and move it later on the new API?

vfdev-5:
@datumbox @lezwon if we see AA detection code similar to classification one <=> single transformation class calling functional ops inside, then `<op>_Only_BBoxes` could be implemented as a special function that produces a list of image crops and we can apply `<op>_image_tensor` to them. I think we can go directly with new API.

lezwon:
@vfdev-5 Yep. Cutout does refer to erasing patches of data from the target box.

ain-soph:
@vfdev-5 
Here is a sample Cutout Implementation for image classification.
https://github.com/ain-soph/trojanzoo/blob/85730f629eddc17dcbc48218274b74c31daf5f99/trojanvision/utils/transform.py#L198-L226

torchvision doesn’t implement this because there is already a nearly equivalence [RandomErasing](https://pytorch.org/vision/stable/generated/torchvision.transforms.RandomErasing.html). The only difference is for the boundary cases. And there are randomness in RandomErasing. The arguments are quite difference.

lezwon:
@vfdev-5 I'm sorry for the delay. I have been a bit occupied with other tasks and haven't been able to give any time to the implementation. I don't think I can pick this issue up anytime soon either. It would be best to assign someone else to it.

[...]

Issue #420, opened 2018-02-13
Feature proposal: Cutout


juneoh:
[Improved Regularization of Convolutional Neural Networks with Cutout](https://arxiv.org/abs/1708.04552) proposes a simple and effective method for regularization. I think it would fit greatly in the `torchvision.transforms` toolbox, and am willing to contribute an implementation with a proper test, if that's okay.

alykhantejani:
@juneoh is this a similar proposal to the one in #226?

juneoh:
@alykhantejani It is indeed similar, but the key difference is that the cited Cutout paper has successfully parameterized effective random occlusions: e.g. the shape, be it polygons to free shapes, matters less than the size, so square patches are sufficient; allowing portions of the occlusion to lie outside the image is critical to good performance; etc. There need only be two hyperparameters for an implementation, the patch size and the fill value.

juneoh:
Oh, but the paper is cited in #226. Perhaps I should have requested a reopen of that issue instead. I thought I've searched thoroughly enough to avoid a duplicate.

alykhantejani:
cc @fmassa who closed out the previous issue - who might have some other thoughts on this

hendrycks:
Both Cutout and Random Erasure are highly similar. If I recall correctly, random Erasure has more parameters, though these parameters can be left to the default setting as is done in the paper. In contrast, Cutout has fewer parameters and is easier to describe, and contrariwise Cutout often requires more tuning.

PR #6609, opened 2022-09-19
[DRAFT] Implement AutoAugment for Detection

ain-soph:
Implement #6224.

Implement [Learning Data Augmentation Strategies for Object Detection](https://arxiv.org/abs/1906.11172)
Refers to: https://github.com/pytorch/vision/issues/3817

### Motivation, pitch
Good to have augmentation in Torchvision

cc @vfdev-5 @datumbox

### DIFF
--- a/torchvision/prototype/transforms/_auto_augment.py
+++ b/torchvision/prototype/transforms/_auto_augment.py
@@ -533,3 +533,111 @@ class AugMix(_AutoAugmentBase):
             mix = F.to_image_pil(mix)
 
         return self._put_into_sample(sample, id, mix)
+
+
+class _AutoAugmentDetectionBase(_AutoAugmentBase):
+    @staticmethod
+    def _transform_image_in_bboxes(
+        fn: Callable[..., torch.Tensor], fn_kwrgs: dict, image: torch.Tensor, bboxes: torch.Tensor
+    ):
+        new_img = image.clone()
+        for bbox in bboxes:
+            bbox_img = new_img[..., bbox[1] : bbox[3], bbox[0] : bbox[2]]
+            out_bbox_img = fn(bbox_img, **fn_kwrgs)
+            new_img[..., bbox[1] : bbox[3], bbox[0] : bbox[2]] = out_bbox_img
+        return new_img
+
+    def _extract_image(
+        self,
+        sample: Any,
+        unsupported_types: Tuple[Type, ...] = (features.Mask),
+    ) -> List[Tuple[int, Union[PIL.Image.Image, torch.Tensor, features.Image, features.BoundingBox]]]:
+        sample_flat, _ = tree_flatten(sample)
+        images = []
+        for id, inpt in enumerate(sample_flat):
+            if _isinstance(inpt, (features.Image, PIL.Image.Image, features.is_simple_tensor)):
+                images.append((id, inpt))
+            elif _isinstance(inpt, features.BoundingBox):
+                images.append((id, inpt))
+            elif isinstance(inpt, unsupported_types):
+                raise TypeError(f"Inputs of type {type(inpt).__name__} are not supported by {type(self).__name__}()")
+
+        if not images:
+            raise TypeError("Found no image in the sample.")
+        if len(images) > 2:
+            raise TypeError(
+                f"Auto augment transformations are only properly defined for a single image and its bboxes, but found {len(images)}."
+            )
+        return images
+
+    def _apply_image_transform(
+        self,
+        image: Any,
+        bboxes: features.BoundingBox,
+        transform_id: str,
+        magnitude: float,
+        interpolation: InterpolationMode,
+        fill: Union[int, float, Sequence[int], Sequence[float]],
+    ) -> Tuple[Any, features.BoundingBox]:
+        # TODO: SolarizeAdd, Cutout, BBox_Cutout, Flip_Only_BBoxes, Cutout_Only_BBoxes
+
+        if transform_id.endswith("_Only_BBoxes"):
+            transform_id = transform_id[:-12]
+            fn_kwargs = dict(transform_id=transform_id, magnitude=magnitude, interpolation=interpolation, fill=fill)
+            image = self._transform_image_in_bboxes(super()._apply_image_transform, fn_kwargs, image, bboxes)
+        elif transform_id.endswith("_BBox"):
+            transform_id = transform_id[:-5]
+            image = super()._apply_image_transform(image, transform_id, magnitude, interpolation, fill)
+            format = features.BoundingBoxFormat.XYXY  # TODO
+            image_size = image  # TODO
+
+            if transform_id == "Rotate":
+                bboxes = F.rotate_bounding_box(bboxes, format, image_size, angle=magnitude)
+            elif transform_id == "TranslateX":
+                bboxes = F.affine_bounding_box(
+                    bboxes,
+                    format,
+                    image_size,
+                    angle=0.0,
+                    translate=[int(magnitude), 0],
+                    scale=1.0,
+                    shear=[0.0, 0.0],
+                )
+            elif transform_id == "TranslateY":
+                bboxes = F.affine_bounding_box(
+                    bboxes,
+                    format,
+                    image_size,
+                    angle=0.0,
+                    translate=[0, int(magnitude)],
+                    scale=1.0,
+                    shear=[0.0, 0.0],
+                )
+            elif transform_id == "ShearX":
+                bboxes = F.affine_bounding_box(
+                    bboxes,
+                    format,
+                    image_size,
+                    angle=0.0,
+                    translate=[0, 0],
+                    scale=1.0,
+                    shear=[math.degrees(math.atan(magnitude)), 0.0],
+                    center=[0, 0],
+                )
+            elif transform_id == "ShearY":
+                bboxes = F.affine_bounding_box(
+                    bboxes,
+                    format,
+                    image_size,
+                    angle=0.0,
+                    translate=[0, 0],
+                    scale=1.0,
+                    shear=[0.0, math.degrees(math.atan(magnitude))],
+                    center=[0, 0],
+                )
+        elif transform_id == "BBox_Cutout":
+            raise NotImplementedError()
+        else:
+            image = super()._apply_image_transform(image, transform_id, magnitude, interpolation, fill)
+
+        return image, bboxes


ain-soph:
@vfdev-5 

- I overload `_extract_image` to return tuples that contains images and bboxes, which requires modification of forward implementation in child classes.
- There is a place that I'm not quite sure about the codes.
  Is format always `XYXY`? And how can we fetch the `image_size` when image could be `torch.Tensor` or `PIL.Image`
  ```python3
            format = features.BoundingBoxFormat.XYXY  # TODO
            image_size = image.shape[-2:]  # TODO

            if transform_id == "Rotate":
                bboxes = F.rotate_bounding_box(bboxes, format, image_size, angle=magnitude)
  ```
- a

vfdev-5:
@ain-soph thanks for the PR, let's remove useless files:
- torchvision/image.pyd
- torchvision/_C.pyd

ain-soph:
> @ain-soph thanks for the PR, let's remove useless files:
> 
> * torchvision/image.pyd
> * torchvision/_C.pyd

Oops! Sorry for missing that... I'll change it soon.

vfdev-5:
@ain-soph do you have any working testing code to run it and visually check the output ? If yes, please share it here, it would be helpful. Thanks !

ain-soph:
@vfdev-5 Sorry for the latency. Was busy at ICLR for the past week. I'll add some test codes tmr.

ain-soph:
The recent auto-augment api modification forces me to re-design the api for detection. I'm still working on this now. Need to know how the new framework works.
Kinda confused if there is any case for such "video detection" compatibility?

datumbox:
@ain-soph The API should be more stable now. Which specific AA modification broke your code. Perhaps we can help you?

ain-soph:
@datumbox Sorry for dragging it so long. The main confusing part is how I can extract the bounding box from inputs, and (after augmentation) how I can insert the modified bounding box back to inputs.

Current image classification codes don't have this part and I can't find something for reference.
I'm trying to overload existing `_flatten_and_extract_image_or_video` and add the bounding box support, but I don't know whether it's recommended. I'll implement an ugly-but-work demo first for review. Hopefully it shall be finished today.

datumbox:
@ain-soph Apologies for the rough edges. You are basically brave for digging into the developer part of the API so early. The public part should be fairly stable. We will continue polishing the internal part to make the experience smoother.

The method you probably want for extracting the bounding boxes is `_utils.query_bounding_box()`. I think overriding temporarily the `_flatten_and_extract_image_or_video` is fine. An alternative would be to have a similar separate method for bboxes. At this point I would recommend to "just do what you need to do" without worrying about the recommended way to implement the transform. We can help you refactor it once you have it in a working condition. 

@vfdev-5 / @pmeier any recommendation on your side at this point?

ain-soph:
I've uploaded a version that shall work and make a very simple unittest. Please take a review and I'm expecting some guidance to improve, thanks!

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -516,3 +516,309 @@ def forward(self, *inputs: Any) -> Any:
             mix = F.to_image_pil(mix)
 
         return self._unflatten_and_insert_image_or_video(flat_inputs_with_spec, mix)
+
+
+class _AutoAugmentDetectionBase(_AutoAugmentBase):
+    def _flatten_and_extract_image_or_video_and_bboxes(
```
vfdev-5, line 522: I think we have to figure out a generic way to do this things: 
- _flatten_and_extract_image_or_video_and_bboxes
- _unflatten_and_insert_image_or_video_and_bboxes
- _apply_image_or_video_and_bboxes_transform

ain-soph: I agree. Currently it’s just a temporary solution. If we can get a generic structure, things would be much better.

ain-soph:
Any suggestion on the method names? Shall I consider modifying the AutoAugment class for classification to unify the methods? @datumbox @vfdev-5

vfdev-5 (commented):
Thanks for the updates @ain-soph . I left primary few comments before going deeper into the review. Can you please provide some images when applied this transformation for visual assessment ?

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -516,3 +516,309 @@ def forward(self, *inputs: Any) -> Any:
...
+                bboxes.data = F.affine_bounding_box(
+                    bboxes.data,
+                    bboxes.format,
+                    bboxes.spatial_size,
+                    angle=magnitude,
+                    translate=[0, 0],
+                    scale=1.0,
+                    shear=[0.0, 0.0],
+                )
+                # bboxes.data = F.rotate_bounding_box(bboxes.data, bboxes.format, bboxes.image_size, angle=magnitude)[0]
```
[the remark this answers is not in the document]
ain-soph: Do we have any preference for `affine` and `rotate`?

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -516,3 +516,309 @@ def forward(self, *inputs: Any) -> Any:
...
+                [("Sharpness", 1.0, 4), ("SolarizeAdd", 0.4, 4)],
+                [("Rotate_BBox", 1.0, 8), ("Sharpness", 0.2, 8)],
+                [("ShearY_BBox", 0.6, 10), ("Equalize_Only_BBoxes", 0.6, 8)],
+                [("ShearX_BBox", 0.2, 6), ("TranslateY_Only_BBoxes", 0.2, 10)],
+                [("SolarizeAdd", 0.6, 8), ("Brightness", 0.8, 10)],
+            ]
+        elif policy == "vtest":
+            return [
+                [("TranslateX_BBox", 1.0, 4), ("Equalize", 1.0, 10)],
+            ]
```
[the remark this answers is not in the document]
ain-soph: It's a policy for testing which exists in the original repo. I'm currently using it for test as well.

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -516,3 +516,309 @@ def forward(self, *inputs: Any) -> Any:
...
+            image = self._transform_image_or_video_in_bboxes(
+                self._apply_image_or_video_transform, fn_kwargs, image, bboxes
+            )
+        elif transform_id.endswith("_BBox"):
+            transform_id = transform_id[:-5]
+            image = self._apply_image_or_video_transform(image, transform_id, magnitude, interpolation, fill)
+
+            if transform_id == "Rotate":
+                bboxes.data = F.affine_bounding_box(
+                    bboxes.data,
```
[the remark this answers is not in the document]
ain-soph: Sorry that I'm previously not familiar with it. Will modify it soon.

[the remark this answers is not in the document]
ain-soph: @vfdev-5 Are you sure that `F.affine` could be used in this case? My mypy check raises exceptions, and `inpt` type annotation in method definition is `datapoints.InputTypeJIT` (`torch.Tensor`). I could not find any method related to type `datapoints.BoundingBox`.

ain-soph:
@vfdev-5  Here's a test example:

```python3

import torch
from torchvision.prototype.transforms._auto_augment import AutoAugmentDetection
from torchvision.prototype.datapoints import Image, BoundingBox


spatial_size = (224, 224)

bboxes = [
    [10, 15, 40, 85],
    [160, 5, 210, 122],
    [180, 186, 220, 202],
    [40, 150, 100, 200],
]
labels = [1, 2, 3, 4]

img = torch.zeros(1, 3, *spatial_size)
for in_box, label in zip(bboxes, labels):
    img[..., in_box[1]:in_box[3], in_box[0]:in_box[2]] = \
        (torch.arange(200, 200 + 3 * (in_box[3] - in_box[1]) * (in_box[2] - in_box[0])).div(5) % 200).reshape(1, 3, in_box[3] - in_box[1], in_box[2] - in_box[0])

img = Image(img.byte())
bboxes = BoundingBox(bboxes, format="XYXY", spatial_size=spatial_size)

sample = [img, bboxes]

t = AutoAugmentDetection(policy="vtest")
out = t(sample)

from torchvision.prototype.transforms import InterpolationMode
interpolation = InterpolationMode.NEAREST
fill_ = None
magnitude = 111.11111450195312
plt.figure()
plt.subplot(221)
plt.imshow(img[0, ...].permute(1, 2, 0))
plt.subplot(222)
plt.imshow(out[0][0, ...].permute(1, 2, 0))
plt.subplot(223)
plt.imshow(F.affine(
    F.equalize(img),
    angle=0.0,
    translate=[int(magnitude), 0],
    scale=1.0,
    interpolation=interpolation,
    shear=[0.0, 0.0],
    fill=fill_,
)[0, ...].permute(1, 2, 0))
plt.subplot(224)
plt.imshow(img.equalize().affine(
    angle=0.0,
    translate=[int(magnitude), 0],
    scale=1.0,
    interpolation=interpolation,
    shear=[0.0, 0.0],
    fill=fill_,
)[0, ...].permute(1, 2, 0))
```
image

ain-soph:
@vfdev-5 Any further comment for the PR? I've provided a visual example above.

vfdev-5:
@ain-soph sorry for delay ! I'm checking this with MSCoco data and from time to time there are following errors appear:
```
Cell In[76], line 7
      4 bboxes = BoundingBox(target["boxes"], format="XYXY", spatial_size=image.shape[-2:])
      6 sample = [image, bboxes]
----> 7 out = t(sample)

File /usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py:1488, in Module._call_impl(self, *args, **kwargs)
   1483 # If we don't have any hooks, we want to skip the rest of the logic in
   1484 # this function, and just call forward.
   1485 if not (self._backward_hooks or self._backward_pre_hooks or self._forward_hooks or self._forward_pre_hooks
   1486         or _global_backward_pre_hooks or _global_backward_hooks
   1487         or _global_forward_hooks or _global_forward_pre_hooks):
-> 1488     return forward_call(*args, **kwargs)
   1489 # Do not call functions when jit is used
   1490 full_backward_hooks, non_full_backward_hooks = [], []

File /vision/torchvision/prototype/transforms/_auto_augment.py:829, in AutoAugmentDetection.forward(self, *inputs)
    826     else:
    827         magnitude = 0.0
--> 829     image_or_video, bboxes = self._apply_image_or_video_and_bboxes_transform(
    830         image_or_video, bboxes, transform_id, magnitude, interpolation=self.interpolation, fill=self.fill
    831     )
    833 return self._unflatten_and_insert_image_or_video_and_bboxes(flat_inputs_with_spec, image_or_video, bboxes)

File /vision/torchvision/prototype/transforms/_auto_augment.py:620, in _AutoAugmentDetectionBase._apply_image_or_video_and_bboxes_transform(self, image, bboxes, transform_id, magnitude, interpolation, fill)
    618     transform_id = transform_id[:-12]
    619     fn_kwargs = dict(transform_id=transform_id, magnitude=magnitude, interpolation=interpolation, fill=fill)
--> 620     image = self._transform_image_or_video_in_bboxes(
    621         self._apply_image_or_video_transform, fn_kwargs, image, bboxes
    622     )
    623 elif transform_id.endswith("_BBox"):
    624     transform_id = transform_id[:-5]

File /vision/torchvision/prototype/transforms/_auto_augment.py:601, in _AutoAugmentDetectionBase._transform_image_or_video_in_bboxes(fn, fn_kwrgs, image, bboxes)
    599 new_img = image.clone()
    600 for bbox in bboxes:
--> 601     bbox_img = new_img[..., bbox[1] : bbox[3], bbox[0] : bbox[2]]
    602     out_bbox_img = fn(bbox_img, **fn_kwrgs)
    603     new_img[..., bbox[1] : bbox[3], bbox[0] : bbox[2]] = out_bbox_img

File /vision/torchvision/prototype/datapoints/_datapoint.py:93, in Datapoint.__torch_function__(cls, func, types, args, kwargs)
     90     return NotImplemented
     92 with DisableTorchFunctionSubclass():
---> 93     output = func(*args, **kwargs or dict())
     95     wrapper = cls._NO_WRAPPING_EXCEPTIONS.get(func)
     96     # Apart from `func` needing to be an exception, we also require the primary operand, i.e. `args[0]`, to be
     97     # an instance of the class that `__torch_function__` was invoked on. The __torch_function__ protocol will
     98     # invoke this method on *all* types involved in the computation by walking the MRO upwards. For example,
     99     # `torch.Tensor(...).to(datapoints.Image(...))` will invoke `datapoints.Image.__torch_function__` with
    100     # `args = (torch.Tensor(), datapoints.Image())` first. Without this guard, the original `torch.Tensor` would
    101     # be wrapped into a `datapoints.Image`.

TypeError: only integer tensors of a single element can be converted to an index
``` 

I think here the problem is with float bbox values.


Another error :

```
IndexError                                Traceback (most recent call last)
Cell In[78], line 7
      4 bboxes = BoundingBox(target["boxes"], format="XYXY", spatial_size=image.shape[-2:])
      6 sample = [image, bboxes]
----> 7 out = t(sample)

File /usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py:1488, in Module._call_impl(self, *args, **kwargs)
   1483 # If we don't have any hooks, we want to skip the rest of the logic in
   1484 # this function, and just call forward.
   1485 if not (self._backward_hooks or self._backward_pre_hooks or self._forward_hooks or self._forward_pre_hooks
   1486         or _global_backward_pre_hooks or _global_backward_hooks
   1487         or _global_forward_hooks or _global_forward_pre_hooks):
-> 1488     return forward_call(*args, **kwargs)
   1489 # Do not call functions when jit is used
   1490 full_backward_hooks, non_full_backward_hooks = [], []

File /vision/torchvision/prototype/transforms/_auto_augment.py:823, in AutoAugmentDetection.forward(self, *inputs)
    821 magnitudes = magnitudes_fn(10, height, width)
    822 if magnitudes is not None:
--> 823     magnitude = float(magnitudes[magnitude_idx])
    824     if signed and torch.rand(()) <= 0.5:
    825         magnitude *= -1

IndexError: index 10 is out of bounds for dimension 0 with size 10
```


I'm using default policy "v0".

Code:
```python
image = Image(torch.randint(0, 256, size=(3, 480, 640), dtype=torch.uint8))
boxes = torch.tensor([[388.3100,  38.8300, 638.5600, 480.0000],
        [ 82.6800, 314.7300, 195.4400, 445.7400],
        [199.1000, 214.1700, 316.4100, 399.2800],
        [159.8400, 183.6800, 216.8700, 273.1200],
        [ 15.8000, 265.4600,  93.2900, 395.1800],
        [ 88.2300, 266.0800, 222.9800, 371.2500],
        [176.9000, 283.8600, 208.4300, 292.3000],
        [537.6300, 230.8300, 580.7100, 291.3300],
        [421.1200, 230.4700, 580.6700, 350.9500],
        [427.4200, 185.4300, 494.0200, 266.3500]])
bboxes = BoundingBox(boxes, format="XYXY", spatial_size=image.shape[-2:])

sample = [image, bboxes]
out = t(sample)
```

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -526,3 +526,302 @@ def forward(self, *inputs: Any) -> Any:
...
+            image = self._apply_image_or_video_transform(image, transform_id, magnitude, interpolation, fill)
+
+            if transform_id == "Rotate":
+                bboxes = bboxes.rotate(angle=magnitude)
+                # bboxes = bboxes.affine(
+                #     angle=magnitude,
+                #     translate=[0, 0],
+                #     scale=1.0,
+                #     shear=[0.0, 0.0],
+                # )
```
vfdev-5, line 628: let's remove commented code as well

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -526,3 +526,302 @@ def forward(self, *inputs: Any) -> Any:
...
+    def _apply_image_or_video_and_bboxes_transform(
+        self,
+        image: Any,
+        bboxes: datapoints.BoundingBox,
+        transform_id: str,
+        magnitude: float,
+        interpolation: InterpolationMode,
+        fill: Dict[Type, datapoints.FillTypeJIT],
+    ) -> Tuple[Any, datapoints.BoundingBox]:
+        # TODO: SolarizeAdd, Cutout, BBox_Cutout, Flip_Only_BBoxes, Cutout_Only_BBoxes
```
vfdev-5, line 609: Do you still plan to add those ?

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -526,3 +526,302 @@ def forward(self, *inputs: Any) -> Any:
...
+
+
+class AutoAugmentDetection(_AutoAugmentDetectionBase):
+    _AUGMENTATION_SPACE = {
+        "AutoContrast": (lambda num_bins, height, width: None, False),
+        "Equalize": (lambda num_bins, height, width: None, False),
+        "Posterize": (
+            lambda num_bins, height, width: (8 - (torch.arange(num_bins) / ((num_bins - 1) / 6))).round().int(),
+            False,
+        ),  # TODO: tf repo is (0 -> 4) while this seems to be (8 -> 4)
```
vfdev-5, line 674: How about these TODOs ?

vfdev-5 (commented):
@ain-soph Thanks for your patience and working on this PR. Hopefully we can get it in a good shape soon and finally merge

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -526,3 +526,302 @@ def forward(self, *inputs: Any) -> Any:
...
+        image: Any,
+        bboxes: datapoints.BoundingBox,
+        transform_id: str,
+        magnitude: float,
+        interpolation: InterpolationMode,
+        fill: Dict[Type, datapoints.FillTypeJIT],
+    ) -> Tuple[Any, datapoints.BoundingBox]:
+        # TODO: SolarizeAdd, Cutout, BBox_Cutout, Flip_Only_BBoxes, Cutout_Only_BBoxes
+
+        if transform_id.endswith("_Only_BBoxes"):
```
vfdev-5, line 611: I think in terms of namings we can following paper's contention: `BBox_Only_X` instead `X_Only_BBoxes`

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -526,3 +526,302 @@ def forward(self, *inputs: Any) -> Any:
...
+            fn_kwargs = dict(transform_id=transform_id, magnitude=magnitude, interpolation=interpolation, fill=fill)
+            image = self._transform_image_or_video_in_bboxes(
+                self._apply_image_or_video_transform, fn_kwargs, image, bboxes
+            )
+        elif transform_id.endswith("_BBox"):
+            transform_id = transform_id[:-5]
+            image = self._apply_image_or_video_transform(image, transform_id, magnitude, interpolation, fill)
+
+            if transform_id == "Rotate":
+                bboxes = bboxes.rotate(angle=magnitude)
```
vfdev-5, line 622: Actually, I'm not quite sure to understand why should we call transforms separately on bboxes. Can't we group them together with image ?

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -526,3 +526,302 @@ def forward(self, *inputs: Any) -> Any:
...
+    ) -> Tuple[Any, datapoints.BoundingBox]:
+        # TODO: SolarizeAdd, Cutout, BBox_Cutout, Flip_Only_BBoxes, Cutout_Only_BBoxes
+
+        if transform_id.endswith("_Only_BBoxes"):
+            transform_id = transform_id[:-12]
+            fn_kwargs = dict(transform_id=transform_id, magnitude=magnitude, interpolation=interpolation, fill=fill)
+            image = self._transform_image_or_video_in_bboxes(
+                self._apply_image_or_video_transform, fn_kwargs, image, bboxes
+            )
+        elif transform_id.endswith("_BBox"):
```
vfdev-5, line 617: Ideally, I would like also to drop `_BBox` postfix.

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -526,3 +526,302 @@ def forward(self, *inputs: Any) -> Any:
...
+        "Contrast": (lambda num_bins, height, width: torch.linspace(0.0, 0.9, num_bins), True),
+        "Brightness": (lambda num_bins, height, width: torch.linspace(0.0, 0.9, num_bins), True),
+        "Sharpness": (lambda num_bins, height, width: torch.linspace(0.0, 0.9, num_bins), True),
+        "Cutout": (lambda num_bins, height, width: torch.linspace(0.0, 100.0, num_bins), False),  # TODO: int()?
+        "BBox_Cutout": (lambda num_bins, height, width: torch.linspace(0.0, 0.75, num_bins), False),
+        "Rotate_BBox": (lambda num_bins, height, width: torch.linspace(0.0, 30.0, num_bins), True),
+        "TranslateX_BBox": (
+            lambda num_bins, height, width: torch.linspace(0.0, 250.0, num_bins),
+            True,
+        ),  # TODO: check 250.0
```
vfdev-5, line 693: @ain-soph Here I do not know what should we follow. Paper says [-150, 150] but the code uses [250](https://github.com/tensorflow/tpu/blob/01b67067be64987e9519b35342ba91d3179daeed/models/official/detection/utils/autoaugment_utils.py#L1621)

@datumbox for image-only AA I see the same question:

- https://github.com/pytorch/vision/blob/135a0f9ea9841b6324b4fe8974e2543cbb95709a/torchvision/prototype/transforms/_auto_augment.py#L169-L171
- https://github.com/tensorflow/tpu/blob/b24729de804fdb751b06467d3dce0637fa652060/models/official/efficientnet/autoaugment.py#L667

So, it was decided to implement paper params instead of official code ones ?

Another question, with current linspace definition I have an impression that we pick a wrong value for TranslateX. Is it OK or is it a bug ? For example, policy defines `("TranslateX_BBox", 0.6, 4)` -> `magnitude_idx=4`. 

```
a = torch.linspace(0.0, 250.0, 10, dtype=torch.long)
a[4] 
> 111.11111450195312
```
but in TF it is 100:
```
level = 4
_MAX_LEVEL=10
translate_const = 250
level = (level/_MAX_LEVEL) * float(translate_const)
level
> 100
```
https://github.com/tensorflow/tpu/blob/01b67067be64987e9519b35342ba91d3179daeed/models/official/detection/utils/autoaugment_utils.py#L1420

torchvision/prototype/transforms/_auto_augment.py
ain-soph: Yes, I would if it allows. I previously saw claim from maintainers that: "Cutout is equivalent to RandomErasing so Cutout won't be added to TorchVision." I'm not sure if that's the case.

torchvision/prototype/transforms/_auto_augment.py
ain-soph: I decide to follow tf repo settings

torchvision/prototype/transforms/_auto_augment.py
ain-soph: Just to clarify that tf repo code has this postfix. I can remove such postfix if you determine to do so.

torchvision/prototype/transforms/_auto_augment.py
ain-soph: See #420 and #3817. However, as I mentioned in https://github.com/pytorch/vision/issues/6224#issuecomment-1171772726, there are some slight difference between `Cutout` and `RandomErasing`.

torchvision/prototype/transforms/_auto_augment.py
ain-soph: To confirm, "group them together with image" means modifying method `_AutoAugmentBase._apply_image_or_video_transform` by adding an optional `bbox` argument. Is that correct? (I previously avoid touching that class.)

ain-soph:
@vfdev-5  Sorry for being busy this week. I'll go over your review carefully during the weekend and continue working on it. Thanks for your patient guidance!

test/test_transforms_v2.py
```diff
@@ -82,6 +82,30 @@ def auto_augment_adapter(transform, input, device):
...
+def auto_augment_detection_adapter(transform, input, device):
+    adapted_input = {}
+    image_or_video_found = False
+    bounding_box_found = False
+    for key, value in input.items():
+        if isinstance(value, datapoints.Mask):
+            # AA detection transforms don't support masks
+            continue
+        elif isinstance(value, datapoints.BoundingBox):
+            # TODO: this will only test the first bbox
```
ain-soph, line 94: We may want to test multiple BoundingBox Formats

test/test_transforms_v2.py
```diff
@@ -82,6 +82,30 @@ def auto_augment_adapter(transform, input, device):
...
+            if bounding_box_found:
+                # AA detection transforms only support a single bounding box tensor
+                continue
+            bounding_box_found = True
+        elif check_type(value, (datapoints.Image, datapoints.Video, is_simple_tensor, PIL.Image.Image)):
+            # TODO: this will only test the first image_or_video
```
ain-soph, line 100: We may want to test different images and videos

test/test_transforms_v2.py
```diff
@@ -119,6 +143,10 @@ class TestSmoke:
             (transforms.AutoAugment(), auto_augment_adapter),
             (transforms.RandAugment(), auto_augment_adapter),
             (transforms.TrivialAugmentWide(), auto_augment_adapter),
+            (transforms.AutoAugmentDetection("v0"), auto_augment_detection_adapter),
+            (transforms.AutoAugmentDetection("v1"), auto_augment_detection_adapter),
+            (transforms.AutoAugmentDetection("v2"), auto_augment_detection_adapter),
```
ain-soph, line 148: We may only need to test one of these.

torchvision/transforms/v2/_auto_augment.py
```diff
@@ -15,6 +15,51 @@
 from .utils import check_type, is_simple_tensor
 
 
+def solarize_add(
+    image: Union[datapoints._ImageType, datapoints._VideoType], addition: int = 0, threshold: int = 128
+) -> Union[datapoints._ImageType, datapoints._VideoType]:
+    # TODO: ? threshold max value: bound = _FT._max_value(inpt.dtype)
```
ain-soph, line 21: Currently it's "255" as max and "128" as threshold. Is it generalized?

torchvision/transforms/v2/_auto_augment.py
```diff
@@ -15,6 +15,51 @@
...
+    if isinstance(image, PIL.Image.Image):
+        def pixel_fn(pixel: int) -> int:
+            return max(0, min(pixel + addition, 255)) if pixel < threshold else pixel
+        return PIL.Image.eval(image, pixel_fn)
+
+    added_image = image.add(addition).clip(0, 255)
+    result = torch.where(image < threshold, added_image, image)
+    # TODO: check if this is appropriate way
+    if isinstance(image, datapoints._datapoint.Datapoint):
```
ain-soph, line 30: Need to check code quality here

torchvision/transforms/v2/_auto_augment.py
```diff
@@ -82,10 +127,23 @@ def _apply_image_or_video_transform(
...
+
+    def _apply_transform(
+        self,
+        inpt: datapoints._InputType,
+        transform_id: str,
+        magnitude: float,
+        interpolation: InterpolationMode,
+        fill: Dict[Type, datapoints._FillTypeJIT],
+    ) -> Union[datapoints._ImageType, datapoints._VideoType]:
+        # TODO: magnitude is not always float, it could be int (e.g., posterize).
```
ain-soph, line 142: In `_AUGMENTATION_SPACE`, they are already converted to `int` format, so we may need to change magnitude type annotation to `Union[float, int]`

torchvision/transforms/v2/_auto_augment.py
```diff
@@ -137,26 +195,43 @@ def _apply_image_or_video_transform(
...
         elif transform_id == "Contrast":
-            return F.adjust_contrast(image, contrast_factor=1.0 + magnitude)
+            return F.adjust_contrast(inpt, contrast_factor=1.0 + magnitude)
         elif transform_id == "Sharpness":
-            return F.adjust_sharpness(image, sharpness_factor=1.0 + magnitude)
+            return F.adjust_sharpness(inpt, sharpness_factor=1.0 + magnitude)
         elif transform_id == "Posterize":
-            return F.posterize(image, bits=int(magnitude))
+            # TODO: we don't need to convert magnitude to int because it's already int format.
+            return F.posterize(inpt, bits=int(magnitude))
```
ain-soph, line 209: Is it still necessary since we have convert it in `_AUGMENTATION_SPACE`?

torchvision/transforms/v2/_auto_augment.py
```diff
@@ -137,26 +195,43 @@ def _apply_image_or_video_transform(
...
+                return inpt
+            return solarize_add(inpt, int(magnitude))
+        elif transform_id == "Flip":
+            return F.horizontal_flip(inpt)
+        elif transform_id == "Cutout":
+            if isinstance(inpt, datapoints.BoundingBox):
+                return inpt
+            return cutout(inpt, pad_size=int(magnitude))
+        elif transform_id == "BBox_Cutout":
+            # TODO: Is there a better way?
```
ain-soph, line 233: Need to check code quality here. It doesn't seem clean to have a separate `BBox_Cutout` case.

torchvision/transforms/v2/_auto_augment.py
```diff
@@ -621,3 +696,269 @@ def forward(self, *inputs: Any) -> Any:
...
+        if not image_or_videos:
+            raise TypeError("Found no image or video in the sample.")
+        if len(image_or_videos) > 1:
+            raise TypeError(
+                f"Auto augment transformations are only properly defined for a single image or video, "
+                f"but found {len(image_or_videos)}."
+            )
+        if not bboxes_list:
+            raise TypeError("Found no bounding box in the sample.")
+        if len(bboxes_list) > 1:
```
ain-soph, line 740: Should we assert that there is always exactly one `bboxes` tensor?

torchvision/transforms/v2/_auto_augment.py
```diff
@@ -621,3 +696,269 @@ def forward(self, *inputs: Any) -> Any:
...
+            raise TypeError(
+                f"Auto augment transformations are only properly defined for a single bboxes tensor, "
+                f"but found {len(bboxes_list)}."
+            )
+
+        idx1, image_or_video = image_or_videos[0]
+        idx2, bboxes = bboxes_list[0]
+        return (flat_inputs, spec, idx1, idx2), image_or_video, bboxes
```
ain-soph, line 748: Shall we keep the 2 `id`?

torchvision/transforms/v2/_auto_augment.py
```diff
@@ -621,3 +696,269 @@ def forward(self, *inputs: Any) -> Any:
...
+        flat_inputs, spec, idx1, idx2 = flat_inputs_with_spec
+        flat_inputs[idx1] = image_or_video
+        flat_inputs[idx2] = bboxes
+        return tree_unflatten(flat_inputs, spec)
+
+    @staticmethod
+    def _transform_image_or_video_in_bboxes(
+        fn: Callable[..., torch.Tensor], fn_kwrgs: dict, image: torch.Tensor, bboxes: torch.Tensor
+    ) -> torch.Tensor:
+        # TODO: ? Do we need to assert bboxes format is XYXY?
```
ain-soph, line 765: I think we need to add a convert sentence here.

torchvision/transforms/v2/_auto_augment.py
```diff
@@ -621,3 +696,269 @@ def forward(self, *inputs: Any) -> Any:
...
+        image: Union[datapoints._ImageType, datapoints._VideoType],
+        bboxes: datapoints.BoundingBox,
+        transform_id: str,
+        magnitude: float,
+        interpolation: InterpolationMode,
+        fill: Dict[Type, datapoints._FillTypeJIT],
+    ) -> Tuple[Any, datapoints.BoundingBox]:
+        if transform_id == "BBox_Cutout":
+            random_index = int(torch.randint(len(bboxes), ()))
+            # TODO: this wrap_like solution seems very complex
```
ain-soph, line 785: Need to check code quality here about the usage of `wrap_like`

torchvision/transforms/v2/_auto_augment.py
```diff
@@ -621,3 +696,269 @@ def forward(self, *inputs: Any) -> Any:
...
+            )
+        else:
+            image = self._apply_transform(image, transform_id, magnitude, interpolation, fill)
+            bboxes = self._apply_transform(bboxes, transform_id, magnitude, interpolation, fill)
+
+        return image, bboxes
+
+
+class AutoAugmentDetection(_AutoAugmentDetectionBase):
+    _AUGMENTATION_SPACE = {
```
ain-soph, line 805: I'm now exactly following [tf repo setting](https://github.com/tensorflow/tpu/blob/b24729de804fdb751b06467d3dce0637fa652060/models/official/efficientnet/autoaugment.py) to set the level ranges, which means level ranges from [0, 10] (0 and 10 included).

torchvision/transforms/v2/_auto_augment.py
```diff
@@ -621,3 +698,274 @@ def forward(self, *inputs: Any) -> Any:
...
+        "Equalize": (lambda num_bins, height, width: None, False),
+        "Posterize": (
+            lambda num_bins, height, width: torch.linspace(8, 4, num_bins + 1, dtype=torch.long),
+            False,
+        ),
+        "Solarize": (
+            # TODO: this is extremely ugly because 255 is multiplied in _apply_transform
+            lambda num_bins, height, width: torch.linspace(256 / 255, 0, num_bins + 1),
+            False,
+        ),  # TODO: it's previously [255, 0], now [256, 0]
```
ain-soph, line 818: The difference shall be negligible.

torchvision/transforms/v2/_auto_augment.py
```diff
@@ -621,3 +698,274 @@ def forward(self, *inputs: Any) -> Any:
...
+        "SolarizeAdd": (
+            lambda num_bins, height, width: torch.linspace(0, 110, num_bins + 1, dtype=torch.long),
+            False,
+        ),
+        # TODO: it's previously [0, 0.9] (10 piece) with random negate, now [-0.9, 0.9] (11 piece)
```
ain-soph, line 823: The meaning of each level quite differs.

torchvision/transforms/v2/_auto_augment.py
```diff
@@ -621,3 +698,274 @@ def forward(self, *inputs: Any) -> Any:
...
+        "Color": (lambda num_bins, height, width: torch.linspace(-0.9, 0.9, num_bins + 1), False),
+        "Contrast": (lambda num_bins, height, width: torch.linspace(-0.9, 0.9, num_bins + 1), False),
+        "Brightness": (lambda num_bins, height, width: torch.linspace(-0.9, 0.9, num_bins + 1), False),
+        "Sharpness": (lambda num_bins, height, width: torch.linspace(-0.9, 0.9, num_bins + 1), False),
+        "Cutout": (lambda num_bins, height, width: torch.linspace(0, 100, num_bins + 1, dtype=torch.long), False),
+        "BBox_Cutout": (lambda num_bins, height, width: torch.linspace(0.0, 0.75, num_bins + 1), False),
+        "TranslateX": (
+            lambda num_bins, height, width: torch.linspace(0.0, 250.0, num_bins + 1),
+            True,
+        ),  # TODO: it's previously [0, 150.0 / 331.0 * width], now [0, 250]
```
ain-soph, line 833: It's a big difference.

ain-soph:
@vfdev-5 I've finished all the coding part. Please take a review. I have commented all important parts.

test/test_transforms_v2.py
```diff
@@ -82,6 +82,30 @@ def auto_augment_adapter(transform, input, device):
...
+    for key, value in input.items():
+        if isinstance(value, datapoints.Mask):
+            # AA detection transforms don't support masks
+            continue
+        elif isinstance(value, datapoints.BoundingBox):
+            # TODO: this will only test the first bbox
+            if bounding_box_found:
+                # AA detection transforms only support a single bounding box tensor
+                continue
+            bounding_box_found = True
```
ain-soph, line 98: We may want to test multiple BoundingBox Formats

test/test_transforms_v2.py
```diff
@@ -119,6 +143,10 @@ class TestSmoke:
             (transforms.AutoAugment(), auto_augment_adapter),
             (transforms.RandAugment(), auto_augment_adapter),
             (transforms.TrivialAugmentWide(), auto_augment_adapter),
+            (transforms.AutoAugmentDetection("v0"), auto_augment_detection_adapter),
+            (transforms.AutoAugmentDetection("v1"), auto_augment_detection_adapter),
+            (transforms.AutoAugmentDetection("v2"), auto_augment_detection_adapter),
+            (transforms.AutoAugmentDetection("v3"), auto_augment_detection_adapter),
```
ain-soph, line 149: We may only need to test one of these.

torchvision/transforms/v2/_auto_augment.py
```diff
@@ -15,6 +15,53 @@
...
+        def pixel_fn(pixel: int) -> int:
+            return max(0, min(pixel + addition, 255)) if pixel < threshold else pixel
+
+        return PIL.Image.eval(image, pixel_fn)
+
+    added_image = image.add(addition).clip(0, 255)
+    result = torch.where(image < threshold, added_image, image)
+    # TODO: check if this is appropriate way
+    if isinstance(image, datapoints._datapoint.Datapoint):
+        result = image.wrap_like(image, result)
```
ain-soph, line 33: Need to check code quality here about the usage of "wrap_like"

torchvision/transforms/v2/_auto_augment.py
```diff
@@ -621,3 +698,274 @@ def forward(self, *inputs: Any) -> Any:
...
+        bboxes: datapoints.BoundingBox,
+        transform_id: str,
+        magnitude: float,
+        interpolation: InterpolationMode,
+        fill: Dict[Type, datapoints._FillTypeJIT],
+    ) -> Tuple[Any, datapoints.BoundingBox]:
+        if transform_id == "BBox_Cutout":
+            random_index = int(torch.randint(len(bboxes), ()))
+            # TODO: this wrap_like solution seems very complex
+            chosen_bbox = bboxes.wrap_like(bboxes, bboxes[random_index].unsqueeze(0))
```
ain-soph, line 788: Need to check code quality here about the usage of wrap_like

torchvision/transforms/v2/_auto_augment.py
```diff
@@ -621,3 +698,274 @@ def forward(self, *inputs: Any) -> Any:
...
+            # TODO: this is extremely ugly because 255 is multiplied in _apply_transform
+            lambda num_bins, height, width: torch.linspace(256 / 255, 0, num_bins + 1),
+            False,
+        ),  # TODO: it's previously [255, 0], now [256, 0]
+        "SolarizeAdd": (
+            lambda num_bins, height, width: torch.linspace(0, 110, num_bins + 1, dtype=torch.long),
+            False,
+        ),
+        # TODO: it's previously [0, 0.9] (10 piece) with random negate, now [-0.9, 0.9] (11 piece)
+        "Color": (lambda num_bins, height, width: torch.linspace(-0.9, 0.9, num_bins + 1), False),
```
ain-soph, line 824: The meaning of each level quite differs.

torchvision/transforms/v2/_auto_augment.py
```diff
@@ -621,3 +700,274 @@ def forward(self, *inputs: Any) -> Any:
...
+            )
+        elif transform_id.endswith("_Only_BBoxes"):
+            transform_id = transform_id.replace("_Only_BBoxes", "")
+            fn_kwargs = dict(transform_id=transform_id, magnitude=magnitude, interpolation=interpolation, fill=fill)
+            image = self._transform_image_or_video_in_bboxes(
+                self._apply_image_or_video_transform, fn_kwargs, image, bboxes
+            )
+        else:
+            image = self._apply_transform(image, transform_id, magnitude, interpolation, fill)
+            bboxes = self._apply_transform(bboxes, transform_id, magnitude, interpolation, fill)
```
ain-soph, line 803: There's a strange type linting error that I couldn't figure out the reason.
```
error: Incompatible types in assignment (expression has type "Union[Tensor, Any]", variable has type "BoundingBox") 
[assignment]
```

torchvision/transforms/v2/_auto_augment.py
```diff
@@ -137,26 +197,45 @@ def _apply_image_or_video_transform(
...
+            return F.invert(inpt)
+        elif transform_id == "Flip":
+            return F.horizontal_flip(inpt)
+        elif transform_id == "SolarizeAdd":
+            if isinstance(inpt, datapoints.BoundingBox):
+                return inpt
+            return solarize_add(inpt, int(magnitude))
+        elif transform_id == "Cutout":
+            if isinstance(inpt, datapoints.BoundingBox):
+                return inpt
```
ain-soph, line 232: Any way to avoid the BoundingBox check?

ain-soph:
There is "CUDA out of memory issue" in the unittest jobs.  
https://app.circleci.com/pipelines/github/pytorch/vision/23820/workflows/bcd925c4-0659-42bd-8e4c-160477a63527/jobs/1848329

ain-soph:
@vfdev-5 @datumbox It shall be ready to merge after reviewing all the comments I posted above.

torchvision/transforms/v2/_auto_augment.py
```diff
@@ -15,6 +15,59 @@
...
+    else:
+        bound = 255
+    if bound != 1:
+        addition = round(addition * bound)
+        threshold = round(threshold * bound)
+
+    if isinstance(image, PIL.Image.Image):
+
+        def pixel_fn(pixel: int) -> int:
+            return max(0, min(pixel + int(addition), bound)) if pixel < int(threshold) else pixel
```
ain-soph, line 32: `int()` convert is only for mypy type checking and unnecessary during execution.

torchvision/transforms/v2/_auto_augment.py
```diff
@@ -137,26 +202,39 @@ def _apply_image_or_video_transform(
...
         elif transform_id == "Equalize":
-            return F.equalize(image)
+            return F.equalize(inpt)
         elif transform_id == "Invert":
-            return F.invert(image)
+            return F.invert(inpt)
+        elif transform_id == "Flip":
+            return F.horizontal_flip(inpt)
+        elif isinstance(inpt, datapoints.BoundingBox):
+            return inpt
```
ain-soph, line 231: This `isinstance()` condition is not consistent with other `transform_id` conditions and makes code more difficult to understand.

ain-soph: This condition is to return original bbox for all following `"SolarizeAdd", "Cutout", "BBox_Cutout"` transforms

torchvision/transforms/v2/_auto_augment.py
```diff
@@ -621,3 +699,272 @@ def forward(self, *inputs: Any) -> Any:
...
+            transform_id = transform_id.replace("_Only_BBoxes", "")
+            fn_kwargs = dict(transform_id=transform_id, magnitude=magnitude, interpolation=interpolation, fill=fill)
+            image = self._transform_image_or_video_in_bboxes(
+                self._apply_image_or_video_transform, fn_kwargs, image, bboxes
+            )
+        else:
+            image = self._apply_transform(image, transform_id, magnitude, interpolation, fill)
+            result = self._apply_transform(bboxes, transform_id, magnitude, interpolation, fill)
+            assert isinstance(result, datapoints.BoundingBox)
+            bboxes = result
```
ain-soph, line 802: `result` variable is only for mypy type checking and unnecessary during execution.
`self._apply_transform(bboxes, transform_id, magnitude, interpolation, fill)` will return a variable with type `Union[Tensor, Any]`, and we need to annotate it to `BoundingBox`

ain-soph:
@vfdev-5  Please take a final review. I have annotated some detailed comments in codes.

Summary:
- Implement `solarize_add` and `cutout` transforms.
- Add 4 new `transform_id` in `apply_transform`: `"Flip", "SolarizeAdd", "Cutout", "BBox_Cutout"`
- In `_transform_image_or_video_in_bboxes`, convert `bbox` format to `XYXY`
- Strictly follow [tensorflow repo](https://github.com/tensorflow/tpu/blob/b24729de804fdb751b06467d3dce0637fa652060/models/official/efficientnet/autoaugment.py) setting to:
  - use 11 levels from 0 to 10. In previous AA codes, levels are from 0 to 9.  
    To implement this, `num_bins` is replaced by `num_bins+1` to keep correct intervals.
  - follow their magnitude settings:
    - `"Solarize"`: slightly different. It's now from 256/255 to 0. In previous AA codes, it's from 1.0 to 0.
    - `"Brightness", "Color", "Contrast", "Sharpness"`: the meaning of level quite differs from previous AA codes.  
      In previous AA codes, it's `[0, 0.9]` in 10 pieces with random negate (so level 0 refers to 0 and level 9 refers to either 0.9 or -0.9).  
      In current AAdet codes, it's `[-0.9, 0.9]` in 11 pieces (so level 0 refers to -0.9 and level 10 refers 0.9).
    - `"TransformX", "TransformY"`: magnitude ranges are different from AA settings.

--------------------------------------------------------------

Several small concerns remaining:
- test file
  - Current test (for both AA and AAdetection) will only use the first boundingbox and image/video in test dict, which makes test cases limited: https://github.com/pytorch/vision/blob/dcca679ad460710c1bb9cb7abe8a5904e7de2ab8/test/test_transforms_v2.py#L69-L82
  - I'm currently testing all 4 policies, which might be unnecessary.
- AutoAugmentDetection
  - There are unnecessary type convert in `solarize_add` and `_apply_image_or_video_and_bboxes_transform`.
  - In `_apply_transform`, I have a `isinstance(inpt, datapoints.BoundingBox)` condition in the middle, which might makes codes inconsistent with other conditions and difficult to understand (see comments for details).
  - Need to check the usage of `wrap_like` is correct:
    `image.wrap_like(image, result)`
    `chosen_bbox = bboxes.wrap_like(bboxes, bboxes[random_index].unsqueeze(0))`
  - `flat_inputs_with_spec` records 2 `idx`s in AAdet, which is different from AA codes.

ain-soph:
@vfdev-5 @datumbox Any feedback from maintainers? I think all code work has been finished and this PR is already ready for review.

vfdev-5:
@ain-soph thanks for pinging and sorry for the delay, I'll check the code this week and see how we can make it done finally

test/test_transforms_v2.py
vfdev-5: Why can't we have mutliple bboxes ?
As for bbox format yes we can test on multiple formats but bbox format it is a detail.

torchvision/transforms/v2/_auto_augment.py
```diff
@@ -15,6 +15,59 @@
...
+        assert pad_fraction > 0.0
+        pad_size = (int(pad_fraction * img_h), int(pad_fraction * img_w))
+
+    lower_pad = max(0, cutout_center_height - pad_size[0])
+    upper_pad = max(0, img_h - cutout_center_height - pad_size[0])
+    left_pad = max(0, cutout_center_width - pad_size[1])
+    right_pad = max(0, img_w - cutout_center_width - pad_size[1])
+
+    cutout_shape = [img_h - (lower_pad + upper_pad), img_w - (left_pad + right_pad)]
+    v = torch.tensor(replace)[:, None, None]
```
vfdev-5, line 67: @ain-soph this code seems to fail:
```
Cell In[16], line 67, in cutout(image, pad_size, pad_fraction, replace)
     64 right_pad = max(0, img_w - cutout_center_width - pad_size[1])
     66 cutout_shape = [img_h - (lower_pad + upper_pad), img_w - (left_pad + right_pad)]
---> 67 v = torch.tensor(replace)[:, None, None]
     68 return F.erase(image, lower_pad, left_pad, cutout_shape[0], cutout_shape[1], v)

IndexError: too many indices for tensor of dimension 0
```

torchvision/transforms/v2/_auto_augment.py
```diff
@@ -698,38 +763,43 @@ def _unflatten_and_insert_image_or_video_and_bboxes(
...
     @staticmethod
     def _transform_image_or_video_in_bboxes(
-        fn: Callable[..., torch.Tensor], fn_kwrgs: dict, image: torch.Tensor, bboxes: torch.Tensor
+        fn: Callable[..., torch.Tensor], fn_kwrgs: dict, image: torch.Tensor, bboxes: datapoints.BoundingBox
     ) -> torch.Tensor:
         new_img = image.clone()
-        int_bboxes = bboxes.to(torch.long)
-        for bbox in int_bboxes:
+        xyxy_bboxes = F.convert_format_bounding_box(bboxes, new_format=datapoints.BoundingBoxFormat.XYXY)
+        for bbox in xyxy_bboxes:
```
vfdev-5, line 770: Nice catch about xyxy format, however boxes should be also dtype long as I committed previously:
```
        int_bboxes = bboxes.to(torch.long)
        for bbox in int_bboxes:
```

torchvision/transforms/v2/_auto_augment.py
```diff
@@ -15,6 +15,59 @@
 from .utils import check_type, is_simple_tensor
 
 
+def solarize_add(
```
vfdev-5, line 18: Let's make this method private : `_solarize_add`
I'm not very sure about Pillow implementation and if we could not use other pillow methods like blend here...
Ideally, it would be better to introduce this method into functional part and test explicitly...

torchvision/transforms/v2/_auto_augment.py
vfdev-5: you can use `typing.cast` for mypy. For example:
```
scale = cast(Tuple[float, float], scale)
```

torchvision/transforms/v2/_auto_augment.py
vfdev-5: Let's remove it `isinstance(inpt, datapoints.BoundingBox)` and all AAdet related "SolarizeAdd", "Cutout", "BBox_Cutout" transforms from here. We can override `_AutoAugmentDetectionBase._apply_transform` where we can add these special cases.

vfdev-5:
> _apply_image_or_video_transform simply calls _apply_transform. Could we merge them?

OK, let's remove `_apply_image_or_video_transform` and keep only `_apply_transform`. `_apply_image_or_video_transform` is sort of alias saying that it works only on images/video.

> In _apply_transform, I have a isinstance(inpt, datapoints.BoundingBox) condition in the middle, which might makes codes inconsistent with other conditions and difficult to understand (see comments for details).

Commented in the code. Let's remove all AADet related stuff from AABase.

>  flat_inputs_with_spec records 2 idxs in AAdet, which is different from AA codes.

This is probably ok.

> Need to check the usage of wrap_like is correct

Seems like ok.


@ain-soph thanks for the work on this transformation! It is rather untrivial and especially that we need finally to train a model with this transform to see real benefit of using it.

Let's fix reported bugs and merge it and continue with follow-up PRs.

vfdev-5 (approved):
LGTM, thanks a lot for the PR @ain-soph and appologies for such long review. 

I pushed main class AutoAugmentDetection as `_AutoAugmentDetection` temporarily the time we validate everything and write proper non-regression tests.

I'll comment in this PR on the next steps.

torchvision/transforms/v2/__init__.py
```diff
@@ -5,7 +5,7 @@
 from ._transform import Transform  # usort: skip
 
 from ._augment import RandomErasing
-from ._auto_augment import AugMix, AutoAugment, RandAugment, TrivialAugmentWide
+from ._auto_augment import _AutoAugmentDetection, AugMix, AutoAugment, RandAugment, TrivialAugmentWide
```
NicolasHug, line 8: Is it intended to keep it private (leading underscore)?

torchvision/transforms/v2/_auto_augment.py
```diff
@@ -621,3 +678,309 @@ def forward(self, *inputs: Any) -> Any:
             mix = F.to_image_pil(mix)
 
         return self._unflatten_and_insert_image_or_video(flat_inputs_with_spec, mix)
+
+
+class _AutoAugmentDetectionBase(_AutoAugmentBase):
```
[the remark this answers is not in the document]
vfdev-5: Later we could simply add other AA techniques for detection.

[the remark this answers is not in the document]
NicolasHug: do we plan to?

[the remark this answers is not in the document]
vfdev-5: There is no clear plans at the moment. Do you want to simplify that and readd later ?

ain-soph:
@vfdev-5 @NicolasHug  Feel free to ping me when there is a plan from maintainers for following work. I could continue working on this.

ain-soph:
@vfdev-5  Any possibility to merge this PR first and leave other plans for new PRs? I'm flooded by hundreds of notifications from CI jobs every day...

ain-soph:
Never mind, those ci jobs are from PyTorch main repo for other PRs.

Issue #6610, opened 2022-09-19
[RFC] U-Net framework


TeodorPoncu:
### 🚀 The feature

A module-based approach of building **U-Nets** inside torchvision, similar to torchmultimodals sub-network approach. Mostly a food for though experiment given the similar nature of other popular vision Architectural Frameworks (namely [DETR](https://arxiv.org/abs/2005.12872), [Deformable-DETR](https://arxiv.org/abs/2010.04159), [Mask2Former](https://arxiv.org/abs/2112.01527) or [Mask DINO](https://arxiv.org/abs/2206.02777) which are just a composition of sub-networks that can be adapted to most vision tasks)

### Motivation, pitch

U-Nets are a good example given the rising popularity of Diffusion Models in which the U-Net paradigm is used (layers or merge strategies being the main difference between most implementations). 

Unlike DETR or Mask2Former, which can be broken done quite simply into a module with 2-4 sub-modules followed by a task head, the U-Net framework presents some more intricate challenges at configuration specification level given that we have to sync cross-level encoder and decoder configurations.

Shifting towards a more Framework / Block based approach for larger architectures (think of `nn.Modules` like experience but for present-time vision architectures) would be beneficial for users when it comes to sharing code, improvements or simply swapping out backbones or different components. For instance, if someone would want to grab a Mask2Former they would have to go and integrate themselves into [Detectron2](https://github.com/facebookresearch/Mask2Former/blob/main/INSTALL.md).

Similarly, if someone would want to jump in into doing diffusion, they would first have to find or make their own U-Net implementation even though what they most-likely want to do is to add a bottleneck with attention or some residual connection somewhere throughout the network or simply a different normalization layer in comparison to the original paper.

These classic paradigms should be easy to configure or specify (the same way `torch.nn.Transformer`  handles the transformer), and if more severe changes are wanted a user can have access to a code-base which they can copy-paste and apply minimal modification to (similar to how DETR [handles positional embeddings](https://github.com/facebookresearch/detr/blob/8a144f83a287f4d3fece4acdf073f387c5af387d/models/transformer.py#L154) in the decoder).

Even if they opt for modifying to code-base and later on share their work, there is the bonus of familiarity when others might want to work on top of their code since it's not entirely different than the base version with which they are already familiar with.

[...]

PR #6611, opened 2022-09-19
[DRAFT][DONT MERGE] U-net proposal

TeodorPoncu:
This is a draft PR supporting the RFC for issue #6610. This is a rough sketch of how an architecture frame-work might look if added into torchvision. Suggestion are most welcome as I believe specifying the individual configurations of certain layers in such frameworks requires some trade-offs in terms of where the code-complexity is offloaded.

### DIFF
--- a/torchvision/models/segmentation/unet.py
+++ b/torchvision/models/segmentation/unet.py
@@ -1,0 +1,313 @@
+from typing import Callable, List, Tuple
+
+import torch
+from torch import nn
+from torchvision.ops import Conv2dNormActivation
+
+
+class UnetFrameworkEncoderLayer(nn.Module):
+    def __init__(self, feature_layer: nn.Module, projection_layer: nn.Module, *args, **kwargs) -> None:
+        super().__init__()
+        self.feature_layer = feature_layer
+        self.projection_layer = projection_layer
+
+    def forward(self, x: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
+        # a residual layer which computes the features that are going to be passed to the decoder
+        residual = self.feature_layer(x)
+        # a projection layer which takes in the residual layer and projects it for the next encoder layer
+        x = self.projection_layer(residual)
+        return x, residual
+
+
+class UnetFrameworkDecoderLayer(nn.Module):
+    def __init__(
+        self,
+        feature_layer: nn.Module,
+        combination_layer: nn.Module,
+        projection_layer: nn.Module,
+        upsample_layer: nn.Module,
+        *args,
+        **kwargs,
+    ) -> None:
+        super().__init__()
+        self.feature_layer = feature_layer
+        self.projection_layer = projection_layer
+        self.upsample_layer = upsample_layer
+        self.combination_layer = combination_layer
+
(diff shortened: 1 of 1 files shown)


datumbox:
@TeodorPoncu Thanks for your work in TorchVision the last few months. It's unclear to me what you would like to continue slowly the work on this PR now. It depends on your bandwidth. We can support on our side with training etc. Let me know your thoughts.

TeodorPoncu:
@datumbox I'm actively working on Diffusion so this PR is of high interest to me. I would very much like a ton of feedback on it from the community in order for us to reach a version which can be as re-usable and flexible as `nn.Transformer`.

datumbox:
@TeodorPoncu if you are in, we are in. Let's do it. I'll let you progress a bit on it and then we can even schedule a call to discuss the details. Feel free to ping us on Slack if you need anything.

TeodorPoncu:
@datumbox Well, I guess the first thing I would need insight with, from TorchVisions's side at least is how would we handle a `"framework"` in which some certain module might or might not take in an additional input during forward. 

For instance one might use some sort of conditional diffusion where they would need to add the conditional information into either the `encoder` or the `bottleneck`, or both.

The only thing that comes to mind right now, without adding  custom input-types like Hugging Face does, would be for the network modules to expect dicts as inputs. But that feels very not pytorch-y / torchvision-y at the moment.

datumbox:
It's a good question, this has come up before in the context of SSL. The issue is that much of these models are very new and the techniques have not converged yet on the input that the specific task takes. This makes it harder to create stable APIs. Note that the detection models do receive a dictionary as input (perhaps not the best example in terms of API design but still a valid existing case in TorchVision), so that's not out of question. What makes it harder for the diffusion case is the fact that we are not yet certain what this input would look like 6 months down the line when 10 more papers are out. So maybe the question here is, can we come up with a U-net implementation that implements faithfully the original paper but has code structure that is reusable or a good starting point for more exotic tasks?

The issue as you are very well aware of is that we don't have the headcount at the moment to open such massive discussions. Perhaps that's something we could take for H1? But I'll be happy to work with you on that and see how we can progress.

TeodorPoncu:
In previous cases where dealing with multi-modal / multi-branch data I defaulted to dicts. I feel that's a very intuitive and fast way to set-up things / experiment with things. The only downside I have encountered so far in this paradigm is when you start needing to keep track of a lot of keys (for instance, 6-7 inputs).

As per faithfully implementing the original, there's one thing I have not followed through here - namely that the original does not produce the same spatial size in the output as in the input. That is a decision that is most-likely 100% domain driven since the original paper was destined for medical imaging.

Having worked in that field myself - the largest degree of network uncertainty is located on the input edges, which makes learning in those regions tricky. If you have a direct loss evaluation in those regions (i.e. a detection box, or a segmentation mask) you might sending bad signal throughout the network. Therefore, the easiest thing to do, is to take a much larger area around the section you want to run the loss / network predictions on and treat it as context. I can definitely reproduce this with the building blocks I have submitted in this PR. 

But do we want to add this behaviour to torchvision's implementation? Throughout my experience so far, I feel that people would expect out of a U-Net to generate the same-sized output as the input.

PR #6612, opened 2022-09-20
port prototype transforms kernel tests to new architecture

pmeier:
Per title. There are still some references to port, but I'll do that in a follow up PR. Currently there are also quite a few failing tests. I'll have a look if I botched the sample inputs generation or if the test actually uncovered some bugs.


pmeier:
> I'll have a look if I botched the sample inputs generation or if the test actually uncovered some bugs.

Both actually. 

1. Some kernels don't support more than one batch dimension and I messed that up when porting the `sample_inputs_fn`. Still, our kernels should support arbitrary batch shapes, so I left `# FIXME` comments. Here is an overview of those that don't

    - `perspective_{image_tensor, mask}`
    - `elastic_{image_tensor, mask}`
    - `gaussian_blur_image_tensor`
    - `equalize_image_tensor`
    - `adjust_sharpness_image_tensor`
2. Some PIL kernels don't support sequences of a single element for `fill` if the number of channels is larger. The tensor counterpart interprets this the same as if the number were passed outside a sequence. Not sure if that is a bug and thus I didn't open an issue yet. I left some `# FIXME` comments. Offenders:

    - `pad_image_pil`
    - `perspective_image_pil`
3. Behavior of PIL and tensor `F.pad` actually diverges. Since we only wrap the stable kernels here, this is not a prototype bug. See #6613.

test/prototype_transforms_kernel_infos.py
```diff
@@ -451,3 +451,778 @@ def reference_inputs_convert_color_space_image_tensor():
...
+            F.rotate_mask,
+            sample_inputs_fn=sample_inputs_rotate_mask,
+            reference_fn=reference_rotate_mask,
+            reference_inputs_fn=reference_inputs_rotate_mask,
+            closeness_kwargs=DEFAULT_IMAGE_CLOSENESS_KWARGS,
+        ),
+    ]
+)
+
+_CROP_PARAMS = combinations_grid(top=[-8, 0, 9], left=[-8, 0, 9], height=[12, 20], width=[12, 20])
```
pmeier, line 593: @vfdev-5 The issue detailed in #6613 stems from these parameters. I just ported them 1-to-1 from our previous tests:

https://github.com/pytorch/vision/blob/841b9a19a5ae6fb92b517438061b9846153cc0c6/test/test_prototype_transforms_functional.py#L128

Should we change them?

vfdev-5 (approved):
Looks good to me.
Let's enable `crop_image_tensor` and it is good to go

vfdev-5:
@pmeier can we port gaussian blur to new framework such that we may have less flaky issues with this test :)

PR #6621, opened 2022-09-21
[proto] Fixed fill type in AA

vfdev-5:
Description:
- Fixed fill type in AA


datumbox:
@vfdev-5 We missed the `fill: Union[int, float, Sequence[int], Sequence[float]] = 0,` in the constructors of the techniques. You only made the change on `_AutoAugmentBase`. Could you update the rest of the classes?

vfdev-5:
Thank! I should be blind

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -21,14 +20,11 @@ def __init__(
         self,
         *,
         interpolation: InterpolationMode = InterpolationMode.NEAREST,
-        fill: Union[int, float, Sequence[int], Sequence[float]] = 0,
+        fill: Union[FillType, Dict[Type, FillType]] = 0,
```
datumbox, line 23: For BC:
```suggestion
        fill: Optional[Union[FillType, Dict[Type, FillType]]] = None,
```

same everywhere else?

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -63,18 +59,19 @@ def _put_into_sample(self, sample: Any, id: int, item: Any) -> Any:
...
 
+        fill_ = fill[type(image)]
         # Fill = 0 is not equivalent to None, https://github.com/pytorch/vision/issues/6517
         # So, we have to put fill as None if fill == 0
         # This is due to BC with stable API which has fill = None by default
-        fill_ = F._geometry._convert_fill_arg(fill)
-        if isinstance(fill, int) and fill == 0:
+        fill_ = F._geometry._convert_fill_arg(fill_)
+        if isinstance(fill_, int) and fill_ == 0:
             fill_ = None
```
datumbox, line 75: As discuss on our call, this will revert the BC for when AutoAugment receives `fill=0` and `fill=None` and align the behaviour or prototype and stable.
```suggestion
```

I've raised a separate issue for discussing the potential options for addressing the problem at #6623. Let me know your thoughts.

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -63,16 +63,10 @@ def _apply_image_transform(
...
         interpolation: InterpolationMode,
-        fill: Dict[Type, FillType],
+        fill: Optional[Dict[Type, FillType]],
     ) -> Any:
-
-        fill_ = fill[type(image)]
-        # Fill = 0 is not equivalent to None, https://github.com/pytorch/vision/issues/6517
-        # So, we have to put fill as None if fill == 0
-        # This is due to BC with stable API which has fill = None by default
+        fill_ = fill[type(image)] if fill is not None else fill
```
datumbox, line 68: Can we rely on the default dictionary to return always `None`?

datumbox (approved):
LGTM, thanks!

PR #6624, opened 2022-09-22
set device in ToTensor and PILToTensor

hoonhan:
Since tensor operations in `ToTensor` and `PILToTensor` performs in CPU with no choice, processing a number of requests to convert to tensors consumes huge CPU resources. Particularly, we cannot get the benefit of GPU multiprocessing due to CPU overhead. I thus added `device` argument in both transforms.

You can check the huge CPU consumption increase with the code below.
```
class TestDataset():
    def __init__(self, len):
        self.dataset = []
        for i in range(len):
            imarray = np.random.rand(100,100,3) * 255
            im = Image.fromarray(imarray.astype('uint8')).convert('RGBA')
            self.dataset.append(im)

    def __getitem__(self, idx):
        return T.ToTensor()(self.dataset[idx])

if __name__ == '__main__':
    len = 10000
    dataset = TestDataset(len)
    for i in range(len):
        img = dataset[i]
```

### DIFF
--- a/torchvision/transforms/functional.py
+++ b/torchvision/transforms/functional.py
@@ -119,7 +119,7 @@ def _is_numpy_image(img: Any) -> bool:
     return img.ndim in {2, 3}
 
 
-def to_tensor(pic) -> Tensor:
+def to_tensor(pic, device: torch.device) -> Tensor:
     """Convert a ``PIL Image`` or ``numpy.ndarray`` to tensor.
     This function does not support torchscript.
 
@@ -127,6 +127,7 @@ def to_tensor(pic) -> Tensor:
 
     Args:
         pic (PIL Image or numpy.ndarray): Image to be converted to tensor.
+        device (torch.device): Device to send the tensor to.
 
     Returns:
         Tensor: Converted image.
@@ -146,7 +147,7 @@ def to_tensor(pic) -> Tensor:
         if pic.ndim == 2:
             pic = pic[:, :, None]
 
-        img = torch.from_numpy(pic.transpose((2, 0, 1))).contiguous()
+        img = torch.from_numpy(pic.transpose((2, 0, 1))).to(device).contiguous()
         # backward compatibility
         if isinstance(img, torch.ByteTensor):
             return img.to(dtype=default_float_dtype).div(255)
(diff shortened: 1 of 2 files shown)


datumbox (commented):
@hoonhan Thanks a lot for the contribution!

Your proposal makes sense. One issue is that currently we are trying to move off from the `ToTensor` idiom because it does too many things (converts the image to tensor and rescales the values). We are currently moving towards allowing the user to choose whether the images should remain on their original uint8 space as this can have memory and performance benefits. We also work on adopting meta-data-rich tensors that would give to the user the colour space of the image, the size and other useful info. This is part of our Transforms V2 initiative which is in progress. Due to this, I would rather to not make any further expansions to the specific methods until the new API is finalised. This will help the faster adoption of the next idioms.

@pmeier @vfdev-5 This is useful feedback, perhaps we should consider it as an expansion on the new API. WDYT?

vfdev-5:
I wonder how this would work with DataLoader using num_workers > 1 and collating tensor datapoint on GPU into a batch. Wont be any problem with that ?

pmeier:
> I wonder how this would work with DataLoader using num_workers > 1 and collating tensor datapoint on GPU into a batch. Wont be any problem with that ?

That was my concern as well. I remember that this was not possible in the past, but a quick check seems to confirm that this is working now:

```py
from torch.utils.data import DataLoader
from torchvision import datasets, transforms


class ToDevice:
    def __init__(self, device):
        self.device = device

    def __call__(self, tensor):
        return tensor.to(self.device)


    [
        transforms.PILToTensor(),
        ToDevice("cuda"),
        transforms.CenterCrop(32),
    ]
)
dataset = datasets.FakeData(transform=transform)

data_loader = DataLoader(dataset, batch_size=4, num_workers=4)

sample = next(iter(data_loader))
print(sample[0].shape)
```

@ejguan @NivekT @VitalyFedyunin is moving to GPU before the collation state officially supported now? If yes, is this even a good idea?

torchvision/transforms/functional.py
```diff
@@ -119,14 +120,15 @@ def _is_numpy_image(img: Any) -> bool:
     return img.ndim in {2, 3}
 
 
-def to_tensor(pic) -> Tensor:
+def to_tensor(pic, device: torch.device = "cpu") -> Tensor:
```
ambujpawar, line 123: Add typing info for pic as well please

torchvision/transforms/functional.py
```diff
@@ -173,7 +175,7 @@ def to_tensor(pic) -> Tensor:
         return img
 
 
-def pil_to_tensor(pic: Any) -> Tensor:
+def pil_to_tensor(pic: Any, device: torch.device = "cpu") -> Tensor:
     """Convert a ``PIL Image`` to a tensor of the same type.
```
ambujpawar, line 179: The docstring below mentions the type of pic is only PIL.Image but over here it is mentioned Any.
Please consider changing this

ambujpawar (changes requested):

vfdev-5:
> The CUDA runtime does not support the fork start method; either the spawn or forkserver start method are required to use CUDA in subprocesses.

https://pytorch.org/docs/stable/notes/multiprocessing.html#cuda-in-multiprocessing

PR #6627, opened 2022-09-22
[prototype] Align and Clean up transform types

datumbox:
This PR:
- Moves all types definitions on a single file, mirroring what we do for funtional
- Fixes incorrect type definitions on existing transforms
- Aligns the definitions of fill types across Transforms (excluding the actual default value which would break BC)


torchvision/prototype/transforms/_augment.py
```diff
@@ -110,8 +108,10 @@ def __init__(self, alpha: float, p: float = 0.5) -> None:
     def forward(self, *inputs: Any) -> Any:
         if not (has_any(inputs, features.Image, features.is_simple_tensor) and has_any(inputs, features.OneHotLabel)):
             raise TypeError(f"{type(self).__name__}() is only defined for tensor images and one-hot labels.")
-        if has_any(inputs, features.BoundingBox, features.Mask, features.Label):
-            raise TypeError(f"{type(self).__name__}() does not support bounding boxes, masks and plain labels.")
+        if has_any(inputs, PIL.Image.Image, features.BoundingBox, features.Mask, features.Label):
+            raise TypeError(
+                f"{type(self).__name__}() does not support PIL images, bounding boxes, masks and plain labels."
+            )
```
datumbox, line 114: I believe we previously missed PIL images which are not supported by this class.

torchvision/prototype/transforms/_augment.py
```diff
@@ -223,7 +223,7 @@ def _copy_paste(
         # This is something different to TF implementation we introduced here as
         # originally the algorithm works on equal-sized data
         # (for example, coming from LSJ data augmentations)
-        size1 = image.shape[-2:]
+        size1 = cast(List[int], image.shape[-2:])
```
datumbox, line 226: mypy started to randomly complain once the image type was defined, so I choose the cast it. Happy to ignore for performance reasons.

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -20,7 +20,7 @@ def __init__(
         self,
         *,
         interpolation: InterpolationMode = InterpolationMode.NEAREST,
-        fill: Optional[Union[FillType, Dict[Type, FillType]]] = None,
+        fill: Union[FillType, Dict[Type, FillType]] = None,
```
datumbox, line 23: I redefined `FillType` to include `None`. See below.

torchvision/prototype/transforms/_deprecated.py
```diff
@@ -52,7 +52,7 @@ def __init__(self, num_output_channels: Literal[1, 3] = 1) -> None:
         super().__init__()
         self.num_output_channels = num_output_channels
 
-    def _transform(self, inpt: DType, params: Dict[str, Any]) -> DType:
+    def _transform(self, inpt: ImageType, params: Dict[str, Any]) -> ImageType:
```
datumbox, line 55: Previously defined incorrect type.

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -59,13 +59,13 @@ def _put_into_sample(self, sample: Any, id: int, item: Any) -> Any:
...
+        image: ImageType,
         transform_id: str,
         magnitude: float,
         interpolation: InterpolationMode,
-        fill: Union[Dict[Type, FillType], Dict[Type, None]],
-    ) -> Any:
-        fill_ = fill[type(image)]
+        fill: Optional[Dict[Type, Optional[FillType]]],
+    ) -> ImageType:
+        fill_ = fill[type(image)] if fill is not None else None
```
datumbox, line 68: Restoring @vfdev-5's earlier approach due to the new definition of the input. We adopt the same idiom elsewhere.

torchvision/prototype/transforms/_misc.py
```diff
@@ -68,7 +68,7 @@ def forward(self, *inputs: Any) -> Any:
 
         return super().forward(*inputs)
 
-    def _transform(self, inpt: Union[torch.Tensor, features._Feature], params: Dict[str, Any]) -> torch.Tensor:
+    def _transform(self, inpt: TensorImageType, params: Dict[str, Any]) -> torch.Tensor:
```
datumbox, line 71: Incorrectly defined `features._Feature` previously. It should have been `features.Image`

torchvision/prototype/transforms/_utils.py
```diff
@@ -31,12 +33,10 @@ def _check_fill_arg(fill: Optional[Union[FillType, Dict[Type, FillType]]]) -> No
...
 
 
-def _setup_fill_arg(
-    fill: Optional[Union[FillType, Dict[Type, FillType]]]
-) -> Union[Dict[Type, FillType], Dict[Type, None]]:
+def _setup_fill_arg(fill: Union[FillType, Dict[Type, FillType]]) -> Optional[Dict[Type, FillType]]:
     _check_fill_arg(fill)
 
-    if isinstance(fill, dict):
+    if fill is None or isinstance(fill, dict):
```
datumbox, line 39: I allow for fill to be directly `None` instead of forcing it to be a dictionary that returns `None` as this simplifies the type declarations everywhere. This is aligned with an earlier proposal for @vfdev-5

torchvision/prototype/transforms/_utils.py
```diff
@@ -18,10 +18,12 @@
 
 # Type shortcuts:
 DType = Union[torch.Tensor, PIL.Image.Image, features._Feature]
-FillType = Union[int, float, Sequence[int], Sequence[float]]
+ImageType = Union[torch.Tensor, PIL.Image.Image, features.Image]
+TensorImageType = Union[torch.Tensor, features.Image]
+FillType = Union[int, float, Sequence[int], Sequence[float], None]
```
datumbox, line 23: Including `None` here ensures it works both when we provide directly a `None` value and when the value is in a dictionary.

datumbox (commented):
Some clarifications:

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -59,13 +59,13 @@ def _put_into_sample(self, sample: Any, id: int, item: Any) -> Any:
...
         self,
-        image: Union[torch.Tensor, PIL.Image.Image, features.Image],
+        image: ImageType,
         transform_id: str,
         magnitude: float,
         interpolation: InterpolationMode,
-        fill: Union[Dict[Type, FillType], Dict[Type, None]],
-    ) -> Any:
-        fill_ = fill[type(image)]
+        fill: Optional[Dict[Type, Optional[FillType]]],
```
vfdev-5, line 66: Shouldn't be 
```suggestion
        fill: Optional[Dict[Type, FillType]],
```
as `FillType` already has None ?

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -337,7 +337,7 @@ def __init__(
         magnitude: int = 9,
         num_magnitude_bins: int = 31,
         interpolation: InterpolationMode = InterpolationMode.NEAREST,
-        fill: Optional[Union[FillType, Dict[Type, FillType]]] = None,
+        fill: Union[FillType, Dict[Type, FillType]] = None,
```
vfdev-5, line 340: This type hint looks a bit strange, we expect to see optional somewhere but surprise FillType includes None and this is not obvious. OK with that as it is just a typehint

vfdev-5 (commented):
I'm ok with the approach. I think we can update fill type handling and avoid mutiple `fill = self.fill[type(inpt)] if self.fill is not None else None` -> `fill = self.fill[type(inpt)]`

torchvision/prototype/transforms/_auto_augment.py
datumbox: True. But if we don't add `None` in `FillType`, we would constantly need to do `Union[Optional[FillType], Dict[Type, Optional[FillType]]` which is very verbose. Would you prefer this idiom?

vfdev-5: I totally understand your poinrt and I'm OK with None inside FillType.

datumbox:
> I'm ok with the approach. I think we can update fill type handling and avoid mutiple fill = self.fill[type(inpt)] if self.fill is not None else None -> fill = self.fill[type(inpt)]

To do this, we need to force `self.fill` be a dictionary. It's possible but then the public `fill` and the `self.fill` will have different types and this will lead to different signatures between internal and public methods. Is this the direction you favour?

torchvision/prototype/transforms/_auto_augment.py
datumbox: Good spot. Accepted.

vfdev-5:
> It's possible but then the public fill and the self.fill

They will already different if you pass a scalar right ?

torchvision/prototype/transforms/_auto_augment.py
datumbox: > If FillType can be None, can't we avoid having fill=None and Optional[Dict] ?

No we can't because after the setup is called, the method returns directly `None` instead of just returning a dictionary that gets `None` everytime. Doing the latter is just an extra unnecessary step but if you prefer it I can make the change. See my comment below though about the misalignment of the internal/external signatures.

datumbox:
Sounds good, let me try your proposal to force `None` into the dictionary

torchvision/prototype/features/_bounding_box.py
```diff
@@ -7,6 +7,7 @@
 from torchvision.transforms import InterpolationMode  # TODO: this needs to be moved out of transforms
 
 from ._feature import _Feature
+from ._utils import FillType
```
datumbox, line 10: Centralized this one in a `_utils.py` file to avoid instances of misalignment with `functional`.

torchvision/prototype/features/_utils.py
```diff
@@ -0,0 +1,4 @@
+from typing import List, Union
+
+# Same definition as in `functional._utils.py`. Copied here to avoid cyclical dependencies.
+FillType = Union[int, float, List[float], None]
```
datumbox, line 4: Alternatively I can move all `functional._utils.py` here and import the other way around.

torchvision/prototype/transforms/_utils.py
```diff
@@ -31,15 +33,13 @@ def _check_fill_arg(fill: Optional[Union[FillType, Dict[Type, FillType]]]) -> No
...
-    fill: Optional[Union[FillType, Dict[Type, FillType]]]
-) -> Union[Dict[Type, FillType], Dict[Type, None]]:
+def _setup_fill_arg(fill: Union[FillType, Dict[Type, FillType]]) -> Dict[Type, FillType]:
     _check_fill_arg(fill)
 
     if isinstance(fill, dict):
         return fill
 
-    return defaultdict(lambda: fill)  # type: ignore[return-value]
+    return defaultdict(lambda: fill)  # type: ignore[return-value, arg-type]
```
datumbox, line 42: If we don't do the `arg-type` ignore we get:
```
torchvision/prototype/transforms/_utils.py:42: error: Argument 1 to
"defaultdict" has incompatible type
"Callable[[], Union[Union[int, float, Sequence[int], Sequence[float], None], Dict[Type[Any], Union[int, float, Sequence[int], Sequence[float], None]]]]";
expected "Optional[Callable[[], Union[float, Sequence[float], None]]]" 
[arg-type]
        return defaultdict(lambda: fill)  # type: ignore[return-value]
                           ^
torchvision/prototype/transforms/_utils.py:42: note: Error code "arg-type" not covered by "type: ignore" comment
Found 1 error in 1 file (checked 222 source files)
```

torchvision/prototype/transforms/functional/_utils.py
```diff
@@ -0,0 +1,24 @@
...
+
+# Real type: Union[torch.Tensor, PIL.Image.Image]
+LegacyImageType = torch.Tensor
+
+# Real type: Union[torch.Tensor, features.Image]
+TensorImageType = torch.Tensor
+
+# Similarly, JIT doesn't support Sequencies and can't support at the same time, Lists of floats and ints.
+# Ideal type: Union[int, float, Sequence[int], Sequence[float], None]
+FillType = Union[int, float, List[float], None]
```
datumbox, line 24: Like I said above, we can move all these into `features._utils.py`.

torchvision/prototype/transforms/_utils.py
pmeier: I think this is worth investigating. At some point `mypy` thinks that fill can also be an `int`. Although that is part of the correct type, it shouldn't be for the actual type. I guess we forgot to update something. @datumbox I can look into this if you want.

pmeier (approved):
One comment inline other LGTM. I'm kinda worried that we need more and more of these workaround to avoid import cycles in the future though.

torchvision/prototype/transforms/_utils.py
datumbox: I agree. I've also felt that this was worth noting. I have no idea if this is a failure or mypy or it's seeing something I dont. If you have any input, I would love to get an advice.

datumbox:
> I'm kinda worried that we need more and more of these workaround to avoid import cycles in the future though.

There is an organic way to solve this, which is leaving the definitions as part of `features` module (this is the case now on main). Let me give that a try and if that's not what we want, I can revert the commit.

torchvision/prototype/features/_feature.py
```diff
@@ -232,3 +230,7 @@ def invert(self) -> _Feature:
 
     def gaussian_blur(self, kernel_size: List[int], sigma: Optional[List[float]] = None) -> _Feature:
         return self
+
+
+InputType = Union[torch.Tensor, PIL.Image.Image, _Feature]
+InputTypeJIT = torch.Tensor
```
datumbox, line 236: Formally `DType`.

torchvision/prototype/features/_feature.py
```diff
@@ -3,16 +3,14 @@
...
 from torchvision.transforms import InterpolationMode
 
 F = TypeVar("F", bound="_Feature")
-
-
-# Due to torch.jit.script limitation we keep DType as torch.Tensor
-# instead of Union[torch.Tensor, PIL.Image.Image, features._Feature]
-DType = torch.Tensor
+FillType = Union[int, float, Sequence[int], Sequence[float], None]
+FillTypeJIT = Union[int, float, List[float], None]
```
datumbox, line 13: Could move to another file if we don't want it here.

torchvision/prototype/transforms/_augment.py
```diff
@@ -10,6 +10,8 @@
 from torchvision.prototype import features
 from torchvision.prototype.transforms import functional as F, InterpolationMode
 
+from ..features._image import ImageType, TensorImageType
```
pmeier, line 13: I would avoid relative imports that span more than one "major" namespace, i.e. relative imports within `torchvision.prototype.transforms` are ok, but importing something from `torchvision.prototype.features` should be done by an absolute import:

```suggestion
from torchvision.prototype.features._image import ImageType, TensorImageType
```

My argument is somewhat weakened by the fact that all of the imports are internally any, so no strong opinion here.

pmeier (approved):
I like  this a lot better now, thanks a lot @datumbox! One question inline. I'll have a look at the `mypy` ignore now.

torchvision/prototype/features/_feature.py
pmeier: I think it is ok there. I would prefer to have everything in a `_typing.py` file, but as you explained this will give us circular imports due to the method signatures. Thus, let's take the way of the least resistance.

torchvision/prototype/transforms/_augment.py
datumbox: Makes sense, done.

datumbox:
There seems to be a [breakage](https://github.com/pytorch/vision/actions/runs/3112529418/jobs/5046049095) from TorchData on our tests.

datumbox:
The tests are now [passing](https://github.com/pytorch/vision/actions/runs/3112718651/jobs/5046460827). I'll restore the TorchData tests and merge.

torchvision/prototype/transforms/_utils.py
pmeier: This is a false alarm by `mypy` right before the line `mypy` thinks that `fill` has type

```
Union[builtins.float, typing.Sequence[builtins.float], None]
```

Note that for `mypy` `float == Union[int, float]` or to illustrate this more: for `mypy` the following is true while this will fail for Python `isinstance(0, float)`. I'm not sure why they went for this, but generic numbers support is sketchy in general. This is why you don't see the `int` and `Sequence[int]` here because for `mypy` they are subtypes of `float` and `Sequence[float]` and therefore redundant.

However, at the time we arrive at the line, `mypy` suddenly thinks that `fill` can be a `Dict` again. I can't find the issue now, but this is longstanding. For some reason `mypy` doesn't forward the more concrete type to inner functions like this `lambda`.

Both

```suggestion
    return defaultdict(lambda: cast(FillType, fill))
```

and

```suggestion
    return defaultdict(cast(Callable[[], FillType], lambda: fill))
```

will eliminate both previously ignored errors, but will raise a new `redundant-cast` error :roll_eyes: 

Thus, we have to ignore one either way and not having `cast` in there makes it more readable.

PR #5793, opened 2022-04-08
Add SWAG model weight that only the linear head is finetuned to ImageNet1K

YosuaMichael:
subtask of https://github.com/pytorch/vision/issues/5708

### Model Description

This model has trunk weight from weakly supervised learning described in https://arxiv.org/pdf/2201.08371.pdf. The linear head is fine-tuned to IMAGENET1K dataset while the pre-trained trunk weights are frozen. 

_**This model is suitable for users that want to fine tune the pre-trained trunk on other downstream datasets**_

### Linear head Fine-tuning parameters on IMAGENET1K:

**Regnet model (for all size 16gf, 32gf, 128gf):**

- Num epochs: 28
- Trained on 1 nodes with 8 voltas GPU (32Gb) each
- Batch size per GPU: 32
- image size: 224
- SGD Optimizer with params:
  * weight decay: 0.001
  * momentum: 0.9
  * use Nesterov: True
- Learning Rate param:
  * scheduler: CosineAnnealingLR
  * Start value: 0.001 
- ImageAugmentation transforms:
  * RandomResizeCrop of size 224 with interpolation 3
  * RandomHorizontalFlip
  * Normalize
- Note: Trained with pytorch mixed precision

**VIsion Transformer (for all size b/16, l/16, h/14):**

- Num epochs: 28
- Trained on 4 nodes with 8 voltas GPU (32Gb) each
- Batch size per GPU: 32
- image size: 224
- SGD Optimizer with params:
  * weight decay: 1.00 E-09
  * momentum: 0.9
  * use Nesterov: True
- Learning Rate param:
  * scheduler: CosineAnnealingLR
  * Start value: 0.04
- ImageAugmentation transforms:
  * RandomResizeCrop of size 224 with interpolation 3
  * RandomHorizontalFlip
  * Normalize
- Note: Trained with pytorch mixed precision

### Validation script and result

```
## RegNet_Y_16GF
python -u ~/script/run_with_submitit.py --timeout 3000 --ngpus 1 --nodes 1 --partition train --model regnet_y_16gf --data-path="/datasets01_ontap/imagenet_full_size/061417" --test-only --batch-size=1 --weights="RegNet_Y_16GF_Weights.IMAGENET1K_SWAG_LINEAR_V1"
# Acc@1 83.976 Acc@5 97.244

## RegNet_Y_32GF
[...]

PR #6629, opened 2022-09-22
Add CREStereo weights.

TeodorPoncu:
This PR includes a CREStereo weight-template pointing towards the author provided weights and reference script trained weights.

Test commands are as following:

**Reference base weights**:
`torchrun --nproc_per_node 1 --nnodes 1 cascade_evaluation.py --dataset middlebury2014-train --batch-size 1 --dataset-root $dataset_root --model crestereo_base --weights CREStereo_Base_Weights.CRESTEREO_ETH_MBL_V1
` 

will yield for 10 recursive iterations and 1 cascade:
`'mae': 1.416, 'rmse': 3.53, '1px': 0.777, '3px': 0.896, '5px': 0.933, 'relepe': 0.148, 'fl-all': 78.388`

**Reference finetuned weights**:
`torchrun --nproc_per_node 1 --nnodes 1 cascade_evaluation.py --dataset middlebury2014-train --batch-size 1 --dataset-root $dataset_root --model crestereo_base --weights CREStereo_Base_Weights.CRESTEREO_FINETUNE_MULTI_V1
`

 will yield for 10 recursive iterations and 1 cascade: 
`'mae': 1.02, 'rmse': 3.073, '1px': 0.854, '3px': 0.938, '5px': 0.96, 'relepe': 0.129, 'fl-all': 86.042`

**Author weights**:
`
torchrun --nproc_per_node 1 --nnodes 1 cascade_evaluation.py --dataset middlebury2014-train --batch-size 1 --dataset-root $dataset_root --model crestereo_base --weights CREStereo_Base_Weights.MEGVII_V1
`

 will yield for recursive iterations and 1 cascade:
`'mae': 0.792, 'rmse': 2.765, '1px': 0.905, '3px': 0.958, '5px': 0.97, 'relepe': 0.114, 'fl-all': 90.429`


torchvision/prototype/models/depth/stereo/crestereo.py
```diff
@@ -1055,11 +1057,145 @@ def _crestereo(
...
 class CREStereo_Base_Weights(WeightsEnum):
-    pass
+    """The metrics reported here are as follows.
+
+    ``mae`` is the "mean-average-error" and indicates how far (in pixels) the
+    predicted disparity is from its true value (equivalent to ``epe``). This is averaged over all pixels
+    of all images. ``1px``, ``3px``, ``5px`` and indicate the percentage of pixels that have a lower
+    error than that of the ground truth. ``relepe`` is the "relative-end-point-error" and is the
+    average ``epe`` divided by the average ground truth disparity. ``f1`` corresponds to the average of pixels whose epe
+    is either >3px, or whom's ``relepe`` is higher than 0.05 (therefore lower is better).
```
YosuaMichael, line 1073: @TeodorPoncu Just curious, where this definition of `f1` comes from? It seems quite custom to me.
Also if I understand correctly this is something like the average value of the disparity that is considered "bad"?

torchvision/prototype/models/depth/stereo/crestereo.py
```diff
@@ -1055,11 +1057,145 @@ def _crestereo(
...
+        meta={
+            **_COMMON_META,
+            "num_params": 5432948,
+            "recipe": "https://github.com/megvii-research/CREStereo",
+            "_metrics": {
+                "Middlebury2014-train": {
+                    # number of cascade stages
+                    1: {
+                        # number of refininement interations
+                        2: {
```
YosuaMichael, line 1090: This double nested key `cascade-stages` -> `refinement iterations` might break the docs and it is not clear on docs just the number `1` and `2`.

@NicolasHug any suggestions on how to make this better?
Currently I am thinking of encode the string key like: `"cascade:1-iter:2" but this looks ugly...

torchvision/prototype/models/depth/stereo/crestereo.py
```diff
@@ -1055,11 +1057,145 @@ def _crestereo(
...
+                            "mae": 1.704,
+                            "rmse": 3.738,
+                            "1px": 0.738,
+                            "3px": 0.896,
+                            "5px": 0.933,
+                            "relepe": 0.157,
+                            "f1": 10.041,
```
YosuaMichael, line 1097: NIT, I think 7 metrics is a bit too much and I suggest to pick top 4 metrics instead.

torchvision/prototype/models/depth/stereo/crestereo.py
```diff
@@ -1055,11 +1057,145 @@ def _crestereo(
...
+    ``mae`` is the "mean-average-error" and indicates how far (in pixels) the
+    predicted disparity is from its true value (equivalent to ``epe``). This is averaged over all pixels
+    of all images. ``1px``, ``3px``, ``5px`` and indicate the percentage of pixels that have a lower
+    error than that of the ground truth. ``relepe`` is the "relative-end-point-error" and is the
+    average ``epe`` divided by the average ground truth disparity. ``f1`` corresponds to the average of pixels whose epe
+    is either >3px, or whom's ``relepe`` is higher than 0.05 (therefore lower is better).
+
+    """
+
+    MEGVII = Weights(
```
YosuaMichael, line 1077: Question, why the name is `MEGVII`?

torchvision/prototype/models/depth/stereo/crestereo.py
```diff
@@ -1055,11 +1057,145 @@ def _crestereo(
...
+                            "mae": 0.749,
+                            "rmse": 2.706,
+                            "1px": 0.907,
+                            "3px": 0.961,
+                            "5px": 0.972,
+                            "relepe": 0.113,
+                            "f1": 3.782,
+                        },
+                    },
+                    2: {
```
YosuaMichael, line 1127: It seems to me the cascade=2 result is worse than cascade=1, in this case we may not need to put cascade=2 here? (just need to explain on the docs that this result is obtained with cascade=1)

YosuaMichael (commented):
Hi @TeodorPoncu , thanks for the PR! Could you give the command that you use to test and get those numbers on the PR description? (see https://github.com/pytorch/vision/pull/5793 for example)

torchvision/prototype/models/depth/stereo/crestereo.py
TeodorPoncu: It's similar to the `f1` used in Optical Flow, main difference is the norm2 of the flow is just norm1 since the second channel is always 0.

torchvision/prototype/models/depth/stereo/crestereo.py
TeodorPoncu: These are the weights uploaded by authors in their repo, with a training recipe that is unknown.

torchvision/prototype/models/depth/stereo/crestereo.py
TeodorPoncu: I have no clear answer for this solution, it boils down to how the users might be training their weights.

torchvision/prototype/models/depth/stereo/crestereo.py
YosuaMichael: Just realized it comes from the repo name: `megvii-research`. Just a little NIT, we should name it `MEGVII_V1` just in case the author release new weight and we want to name it `MEGVII_V2` later on.

torchvision/prototype/models/depth/stereo/crestereo.py
YosuaMichael: Oh, I think there is a typo here. In optical flow it is called `fl-all` notice it use small `L` rather than `1`.
see https://github.com/pytorch/vision/blob/main/torchvision/models/optical_flow/raft.py#L682
From what I see on [KITTI](https://www.cvlibs.net/datasets/kitti/eval_scene_flow.php?benchmark=flow) it is meant for optical flow outlier.

Lets fix the naming (probably need to fix the one in reference as well) so people dont mixed this with the standard `f1`.

torchvision/prototype/models/depth/stereo/crestereo.py
jdsgomes: I actually discussed this offline with @TeodorPoncu but forgot about the way it would be rendered in docs, and indeed it would be problematic when this is moved away from prototype:
https://github.com/pytorch/vision/blob/3c9ae0ac5dda3d881a6ce5004ce5756ae7de7bc4/docs/source/conf.py#L395

One option would be to report only on the "standard" configuration and leave the detailed breakdown as private

```
"_metrics": {
    "Middlebury2014-train": {
          # metrics for x cascades and y iterations 
          "mae": 1.704,
          "rmse": 3.738,
          "1px": 0.738,
          "3px": 0.896,
          "5px": 0.933,
          "relepe": 0.157,
          "f1": 10.041,
          "_detailed":
               # number of cascade stages
               1: {
               # number of refininement interations
                   2: {      
                    "mae": 1.704,
                    "rmse": 3.738,
                    "1px": 0.738,
                    "3px": 0.896,
                    "5px": 0.933,
                    "relepe": 0.157,
                    "f1": 10.041,
                    },
                   5: {
                   ....
                   },
              }
          }
     }   
```

YosuaMichael: @jdsgomes yes, I agree on your suggestion. I think have the the standard one with other non-standard in private works. We just need to define what is the `metric_keys` for stereo_matching use case in the docs code later.

torchvision/prototype/models/depth/stereo/crestereo.py
YosuaMichael: Just curious, how should we train the model so that we can get better result in the cascaded inference?

TeodorPoncu: I am unaware if there is a recommended way of doing this given there's no hint of it neither in the paper nor the provided implementation.

My only remark so far is that cascaded inference can work better at very large resolution but that is not a given. Performance is tightly tied to the scene disparity distribution that is presented to the model throughout training.

Given that the model is not trained in cascaded manner, the attention operators at the lowest resolution become "overfitted" to the disparity distribution and aspect ratios encountered during training.

A `2X` change in resolution will most likely have a low impact on the evaluation given that for the base training setting resolution (`384x512`) the non-offseted correlation layer can account for over half of the entire image. (a pixel in the lowest feature map has a receptive `32x32`, and the default `1d search` is a window of `1x9`, therefore the horizontal search can account for `288` horizontal pixels in the original image).

If the offset values would push individual pixel by a value of `1` away from their previous centre (basically obtaining a dilation factor of `2`) we would end up with a correlation product that scanned almost the entire image horizontally. Hence, the importance of the training data distribution.

In this case, going up to a resolution of `768x1024` will not produce a significant change in the networks capabilities (as most of the training sets will not have pixels that are further away than a quarter of the entire image). As such, the cascaded inference can be seen as a way of bringing back the data in the original "search" distribution when faced with extremely large resolutions.

TeodorPoncu: To further exemplify the importance of training data one can improve the evaluation on `Middlebury2014` by almost `0.2 mae` with very minimal fine-tuning. Based on the datasets and scales which are fed into the model during fine-tuning, this improvement can appear after less than `10000` steps.

jdsgomes (approved):
LGTM! Thanks

torchvision/prototype/models/depth/stereo/crestereo.py
```diff
@@ -1055,11 +1057,371 @@ def _crestereo(
...
+                    # metrics for 10 refinement iterations and 1 cascade
+                    "mae": 0.792,
+                    "rmse": 2.765,
+                    "1px": 0.905,
+                    "3px": 0.958,
+                    "5px": 0.97,
+                    "relepe": 0.114,
+                    "fl-all": 90.429,
+                    "_detailed": {
+                        1: {
```
YosuaMichael, line 1096: NIT, lets give comment indicating that `1` is the number of cascade

YosuaMichael (approved):
Thanks @TeodorPoncu , the PR looks good to me!
I have a minor NIT, but I will approve to unblock.

PR #6632, opened 2022-09-22
Remove cpp model in v0.14 due to deprecation

YosuaMichael:
This PR is linked to #6258 for removing things that already in deprecation for 2 version.

We separate the cpp models removal to check separately if there is anything broken because of this.


YosuaMichael:
@datumbox @NicolasHug In this PR I remove all the cpp models, and I have also tried it internally in [D39747331](https://www.internalfb.com/diff/D39747331) and it seems okay (nothing break)

NicolasHug (commented):
Thank you Yosua

I think we'll also want to remove the `tests/test_models.cpp` file as it relies on `models.h`. Prior to merging, would you mind  making sure that our example in `examples/cpp` still builds and runs properly? Ideally we would also check `test/test_cpp_models.py`, but that thing has been broken for years, so I don't think that's possible/reasonable.

When this PR is merged I think we'll want to immediately trigger a sync, in order to catch any issues as early as possible (Your diff looks good indeed, but stuff can go wrong at land time too).

YosuaMichael:
Thanks for the suggestion @NicolasHug will do that.
And yeah, I agree things can go wrong during land time too, any suggestion how to check this without actually landing it?

datumbox:
As Nicolas said, the c++ tests are not currently run and there are some extra mechanisms that determine this. Look for the `WITH_CPP_MODELS_TEST` flag to eliminate occurrences. Those are turned off by default so nothing will break if you just straight delete it but it's worth cleaning up. Let me know if you need any help.

YosuaMichael:
I have check building `examples/cpp` and make sure the compiled `hello-world` binary work in this branch. Here is a summary on what I did:
```

# Install libtorch
# Download libtorch from https://download.pytorch.org/libtorch/nightly/cpu/libtorch-cxx11-abi-shared-with-deps-latest.zip
# Unzip the libtorch

# Build torchvision cpp binary
# In torchvision repo, checkout this PR branch, and create a folder named build and go inside it, and do the following [command](https://download.pytorch.org/libtorch/nightly/cpu/libtorch-cxx11-abi-shared-with-deps-latest.zip):
cmake -DCMAKE_PREFIX_PATH="<PATH_TO_LIBTORCH>" ..
make
sudo make install 

# Build the examples/cpp/hello_world
# Go to examples/cpp/hello_world, and create build folder, then do the following:
cmake -DCMAKE_PREFIX_PATH="<PATH_TO_LIBTORCH>;<PATH_TO_TORCHVISION>" ..
cmake --build .

# Test the hello-world binary works
# Go back to examples/cpp/hello_world
python trace_model.py
./build/hello-world 

# This will output model outputs that look like:
Loading model
Model loaded
Columns 1 to 10 0.0145 -0.0182 -0.0140 -0.0071 -0.0927  0.0303  0.0186  0.0226 -0.0336  0.0406

Columns 11 to 20-0.0938  0.0164  0.0052 -0.0542 -0.0935  0.0205  0.0456 -0.0628  0.0219  0.0488

...

Columns 991 to 1000-0.1112  0.0571  0.0439 -0.0282  0.0506  0.0244  0.0018  0.0244  0.0662  0.0135
[ CPUFloatType{1,1000} ]
```

datumbox:
@YosuaMichael Good to know. I think the quoted example is not related to the changes you make. I've already cleaned up any interactions with the C++ models when we deprecated it at #4375

NicolasHug:
Thanks for checking @YosuaMichael . What is the plan regarding the cpp tests? We still have a reference to the now-deleted header:

https://github.com/pytorch/vision/blob/7046e56fe4370e94339b3e8b6fd011e285294a3a/test/test_models.cpp#L5

I think it's best to avoid keeping the `WITH_CPP_MODELS_TEST` functionality until this can build properly, so we probably want to either  clean that up now, or just straight up delete these tests (even if temporarily).

YosuaMichael:
@NicolasHug For the test, I plan to clean it up

YosuaMichael:
@datumbox @NicolasHug I have cleanup the cpp model test. Here what I did:
- Remove `test/test_models.cpp` and `test/test_cpp_models.py`
- Cleanup the `setup.py` that relate to cpp models test

Could you help me take a look to make sure I am doing it right?

NicolasHug (approved):
Thanks @YosuaMichael , build job are (some) unittests jobs are passing properly, so I think this is safe to merge.

datumbox (approved):
@YosuaMichael all changes LGTM, thanks Yosua.

On the future it might be good to prioritise such clean ups immediately after the release to give a lot of time to catch issues related to removing things.

PR #6633, opened 2022-09-22
always run all prototype tests in CI

pmeier:
In #6587 we decided to split the transforms tests into three CI steps. This makes it easier to find failing tests, but has one glaring downside that we missed: if the first step fails, the other steps aren't run at all. Imagine the CI is toasted for some external reason and now it won't run the tests your PR touches.

This PR refines the step triggers to run even if the previous tests failed.


.github/workflows/prototype-tests.yml
```diff
@@ -40,13 +40,18 @@ jobs:
         run: pip install --progress-bar=off pytest pytest-mock
 
       - name: Run prototype datasets tests
+        id: datasets
         shell: bash
         run: pytest --durations=20 test/test_prototype_datasets*.py
 
       - name: Run prototype transforms tests
+        id: transforms
+        if: ${{ success() || steps.datasets.conclusion == 'failure' }}
```
pmeier, line 49: We cannot use

```suggestion
        if: always()
```

here for two reasons:

1. [`always()`](https://docs.github.com/en/actions/learn-github-actions/expressions#always) is true even if the workflow is cancelled. Meaning we can't the tests through the UI.
2. We only want to run the tests if the previous tests failed not for example the setup.

YosuaMichael: Hi @pmeier a noob question here, what is the reason we only want to run the test if the previous test failed?
Can't we do something like:
```
if: success() || failure()
```
(I get this from https://stackoverflow.com/questions/58858429/how-to-run-a-github-actions-step-even-if-the-previous-step-fails-while-still-f )

pmeier: `failure()` will be true if *any* step has failed. Imagine installing `torchvision` failed for some reason. If we use `failure()` the tests will now also be run and obviously fail. This is somewhat a waste of CI resources and also gives a false signal that there is also a problem with the tests (the step will be marked as failure rather than as skipped) although that is impossible to say.

NicolasHug (approved):
No opinion from my side, accepting to unblock (let's just make sure this is fine with @YosuaMichael  as well before merging)

As a side Q:

> In https://github.com/pytorch/vision/pull/6587 we decided to split the transforms tests into three CI steps. This makes it easier to find failing tests

Is there a way to tell GA to report the failing tests separately, something like what CircleCI does with a [dedicated "failing tests" tab](https://app.circleci.com/pipelines/github/pytorch/vision/20478/workflows/b888c7cc-3a5c-4bb0-a0c4-4a0b6d4c8704/jobs/1665997/tests) ?

.github/workflows/prototype-tests.yml
```diff
@@ -40,13 +40,18 @@ jobs:
...
         shell: bash
         run: pytest --durations=20 test/test_prototype_transforms*.py
 
       - name: Run prototype models tests
+        if: ${{ success() || steps.datasets.conclusion == 'failure' || steps.transforms.conclusion == 'failure' }}
```
YosuaMichael, line 54: @pmeier If I understand correctly, since we have:
```
if: ${{ success() || steps.datasets.conclusion == 'failure' }}
```
on the transforms test, therefore it will not be skipped and will have state of either `success` or `failure`.

In this case, I think we dont need `steps.datasets.conclusion == 'failure'` conditions here since it is kinda redundant. I might be wrong on this though, could you confirm this @pmeier ?

.github/workflows/prototype-tests.yml
YosuaMichael: I see, yeah I agree that it will give a wrong signal to the test which is not desired. Thanks for the explanation!

.github/workflows/prototype-tests.yml
pmeier: > therefore it will not be skipped and will have state of either `success` or `failure`.

`failure` is not concrete enough. It will not be skipped in case of failure of the datasets tests. If the failure happens in any other step, e.g. `torchvision` installation, the step will be skipped.

> In this case, I think we dont need `steps.datasets.conclusion == 'failure'` conditions here since it is kinda redundant.

They are not redundant. If we leave them out, we are left with `if: success()` which is the default behavior if you don't define `if:` at all. Meaning, the step will only be run if *all* previous steps have succeeded. 

We need this step to run if *all setup steps* (everything before the first test step) have succeeded, regardless of the state of the datasets tests. IMO `success() || steps.datasets.conclusion == 'failure'` is shortest expression that gets us this behavior, but if you can find a better one, I'm all ears.

NicolasHug, line 54: > They are not redundant. If we leave them out, we are left with if: success() which is the default behavior if you don't define if: at all. Meaning, the step will only be run if all previous steps have succeeded.

I don't think @YosuaMichael  is suggesting to remove all `XYZ == 'failure'` checks. From what I understand I think he's suggesting the  following change only at this specific line (which I agree seems like it should work):

 ```suggestion
         if: ${{ success() || steps.transforms.conclusion == 'failure' }}
```

Comment (NicolasHug) [reply .github/workflows/prototype-tests.yml:54]: Although I guess that

`success() || steps.transforms.conclusion == 'failure'`

will evaluate to False if `transforms` passed but `datasets` failed...? In which case I agree we need the long form.

Comment (pmeier) [reply .github/workflows/prototype-tests.yml:54]: What if `steps.datasets.conclusion == 'failure'` and `steps.transforms.conclusion == 'success'`? `success()` will be `false` since we have at least one failure. 

In this case `success() || steps.transforms.conclusion == 'failure'` boils down to `false || false` -> `false` and the step is not run. 

With my version we get `false || true || false` -> `true` and the step will be run.

In general we need an expression that makes sure that the setup has succeeded. This is achieved by checking `success()` or failure of any of the steps afterwards. 

Branching is a pain in CI in general, but I think with these three conditions it is somewhat tolerable. Given that this sparked quite a bit of confusion maybe we can refactor this a little. I see two options there:

1. Change the condition to `if: steps.$LAST_SETUP_STEP.conclusion == 'success'`.
2. Add a another dummy step after the setup that we can check for success

From these two I like 2. better since then we only need to remember to keep it as a last step of the setup but can do whatever we want before it.

**Comment by pmeier** [comment]
> Is there a way to tell GA to report the failing tests separately, something like what CircleCI does with a [dedicated "failing tests" tab](https://app.circleci.com/pipelines/github/pytorch/vision/20478/workflows/b888c7cc-3a5c-4bb0-a0c4-4a0b6d4c8704/jobs/1665997/tests) ?

I wondered this myself a few months back and I found three possible options:

1. https://github.com/dorny/test-reporter
2. https://github.com/EnricoMi/publish-unit-test-result-action
3. https://github.com/mikepenz/action-junit-report

Unfortunately all of these options were not exactly what I was looking for. Either they had no real `pytest` support, i.e. the output looked quite weird for what we want, or they added quite a bit of noise that we also don't want. 

I tried to implement [my own solution](https://github.com/pmeier/pytest-summary-gha) until I realized that I can only use container actions when running on `ubuntu`. For cross-platform support the action needs to be written in JS, but my skills with that are not good enough to get to a usable state in my limited free time. If you want this, I would be happy to pick it up though during work :innocent:

File: .github/workflows/prototype-tests.yml (hunk 2 above)
Comment (pmeier) [reply .github/workflows/prototype-tests.yml]: I've implemented your suggestion in 89f83067817f454bb5a7b919d38f8661ca23de20: https://github.com/pytorch/vision/actions/runs/3112539248/jobs/5046070450 prototype tests are skipped although they should be run.

Comment (pmeier) [reply .github/workflows/prototype-tests.yml]: I went for 2. in 7a9eb1a00a35a19ff82606c8130222cf61d9cdbe and it works out well. There is an outage on the datasets tests, but the others are still run

**Comment by pmeier** [comment]
It seems `if: steps.setup.conclusion == 'success'` is not sufficient if there is an failure earlier even if the condition is otherwise `true`. Thus, I went with `success() || ( failure() && steps.setup.conclusion == 'success' )` which seems to do the job.

A little more complex than I had hoped before, but avoid the condition chain for multiple steps and thus should still be easier than the initial proposal. LMK what you think.

File: .github/workflows/prototype-tests.yml (hunk 2 above)
Comment (YosuaMichael) [reply .github/workflows/prototype-tests.yml]: Sorry, was otw to the office just now. Yeah, what I meant is like what @NicolasHug suggest, and from @pmeier explanation I can understand why, thanks for that.

I also saw your changes now using:
```
success() || ( failure() && steps.setup.conclusion == 'success' )
```

I think this is good (and avoid chaining)! Thanks for this!

YosuaMichael (approved):

PR #6634, opened 2022-09-22
Fix typo in category info of EuroSAT.

kdexd:
EuroSAT has 10 categories, however the current version has a typo which results in 9 categories instead of 10 and a new category named `Industrial,Pasture` which is not a part of EuroSAT. This commit fixes the typo.


NicolasHug (approved):
Thank you @kdexd , good catch!

PR #6635, opened 2022-09-23
diversify parameter types for a couple of prototype kernels

pmeier:
This is preparation for #6636. It diversifies the parameter type inputs for the kernels that are handled there. 

This PR only skips some JIT tests for `int` or `float` inputs while we have annotated `List[int]` and `List[float]`. All inputs that otherwise provide a tuple instead of a list work.


test/prototype_transforms_dispatcher_infos.py
```diff
@@ -50,6 +50,13 @@ def maybe_skip(self, *, test_name, args_kwargs, device):
             features.BoundingBox: F.resize_bounding_box,
             features.Mask: F.resize_mask,
         },
+        skips=[
```
pmeier, line 53: These skips currently don't do anything since we don't have integer sizes, but will be necessary as soon that happens.

test/prototype_transforms_kernel_infos.py
```diff
@@ -129,6 +129,9 @@ def sample_inputs_horizontal_flip_mask():
 
 def _get_resize_sizes(image_size):
     height, width = image_size
+    length = max(image_size)
+    # yield length
```
pmeier, line 133: @datumbox uncomment this to see if your fix worked.

datumbox (approved):
@pmeier LGTM. Thanks for fixing the tests. It's going to be handy for checking #6636 more thoroughly.

test/prototype_transforms_kernel_infos.py
```diff
@@ -129,6 +129,9 @@ def sample_inputs_horizontal_flip_mask():
...
+    yield [length]
     yield height, width
```
datumbox, line 135: This will yield a tuple instead of a list. Currently my tests fail because of that. Should we patch it @pmeier?

pmeier: Not sure if we can. For all the input types that accept a `List`, I also added tuples as input arguments. This works fine from Python. We need to check if that is also possible from C++. If yes, we shouldn't remove.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -279,33 +279,26 @@ def _affine_bounding_box_xyxy(
...
-
-    if scale is None:
-        scale = 1.0
-
-    if shear is None:
-        shear = [0.0, 0.0]
+    # This is just a dummy value to avoid raising an error in `_affine_parse_args` although we don't have an
+    # interpolation mode for bounding boxes.
+    interpolation = InterpolationMode.NEAREST
+    angle, translate, shear, center = _affine_parse_args(angle, translate, scale, shear, interpolation, center)
```
datumbox, line 291: I have explicitly avoided calling this method on my other PR. See https://github.com/pytorch/vision/pull/6636#discussion_r978920646

test/prototype_transforms_kernel_infos.py
datumbox: I just need a way to work around this on the other PR to check JIT-scriptability for lists. A temporary hack will work, even on the actual test kernel invocation side.

torchvision/prototype/transforms/functional/_geometry.py
pmeier: Well, I used it, since the old parsing, i.e. 

https://github.com/pytorch/vision/blob/2d92728341bbd3dc1e0f1e86c6a436049bbb3403/torchvision/prototype/transforms/functional/_geometry.py#L291-L306

failed for some of the new sample inputs. Meaning, the bounding box kernel did not support some parameters that the other kernels supported. This is why I used this to make sure that every kernel does the same. If you have speed concerns, I can look into what exactly caused the failure and only fix this.

test/prototype_transforms_kernel_infos.py
pmeier: As discussed offline, this PR is doing what @datumbox wants, i.e. adding sample inputs for everything that currently works.

torchvision/prototype/transforms/functional/_geometry.py
pmeier: I've reverted the changes in 6dfc9657ce89fe9e018a11ee25a8e26c7d3d43c6 and added skips for the tests. They should actually be xfails, but the test framework currently doesn't support that. #6653 adds support for xfails. @datumbox you need to remove the skips in your PR, because the tests otherwise will not start to fail, even if you did something wrong.

datumbox (approved):

PR #6636, opened 2022-09-23
[prototype] Fix BC-breakages on input params of `F`

datumbox:
Depends on #6635

A number of dispatchers on stable `F` define one typing but handle multiple. This comes from the days where JIT didn't support `Union`. One such an example, is the `size` parameter of `resize` which is declared as `List[int]` but `int` is also handled internally. Due to the use of these fake types, the prototype `F` doesn't currently have those mitigations. This PR, tries to restore BC-breakages by correcting the types on prototype `F` and handling all possible inputs. The issue was first reported by @vfdev-5 for `resize`. I took a look and here is some additional dispatchers affected:
- [x] resize: size
- [x] center_crop: output_size
- [x] five_crop: size
- [x] ten_crop: size
- [x] gaussian_blur: kernel_size, sigma
- [x] affine: angle, shear

Here is the proposed changes for each method:
- Change the typing info on the prototype to reflect the true supported types. Do this for ~~`_Feature` methods, `F` dispatchers, the actual kernels and~~ the Transform constructors. Edit: Unfortunately this is not possible for the kernels due to JIT. See this https://github.com/pytorch/vision/pull/6636#pullrequestreview-1118828611 
- Add the actual code mitigations in the kernels instead on the dispatchers, so that the behaviour remains the same no matter what the user is calling (kernels are to become public in the new API).


torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -92,11 +92,13 @@ def vertical_flip(inpt: features.InputTypeJIT) -> features.InputTypeJIT:
 
 def resize_image_tensor(
     image: torch.Tensor,
-    size: List[int],
+    size: Union[List[int], int],
```
vfdev-5, line 95: nit:
```suggestion
    size: Union[int, List[int]],
```

torchvision/prototype/features/_bounding_box.py
```diff
@@ -78,13 +78,16 @@ def vertical_flip(self) -> BoundingBox:
...
         interpolation: InterpolationMode = InterpolationMode.BILINEAR,
         max_size: Optional[int] = None,
         antialias: bool = False,
     ) -> BoundingBox:
         output = self._F.resize_bounding_box(self, size, image_size=self.image_size, max_size=max_size)
-        image_size = (size[0], size[0]) if len(size) == 1 else (size[0], size[1])
+        if isinstance(size, int):
+            image_size = [size, size]
+        else:
+            image_size = (size[0], size[0]) if len(size) == 1 else (size[0], size[1])
```
pmeier, line 90: I dislike the nested `if` here. Maybe

```suggestion
        if isinstance(size, int):
            image_size = [size, size]
        elif len(size) == 1:
            image_size = [size[0], size[0]]
```

?

This is almost equivalent with what we had before with the only difference that the previous solution silently ignored `size[2:]`. Meaning, it would still work if we pass sizes of arbitrary size. My suggestion will now raise an exception in the kernel.

pmeier:
Plus, `mypy` seems to be complaining about the change somewhere.

datumbox:
@pmeier @vfdev-5 Thanks for the comments. The PR is half-baked so I'll make sure the nits and tests are in order before finalizing. I was hoping you can provide feedback around the overall approach described on the PR description (aka where we place the mitigations, how we align the typing etc). I take it from your comments we are aligned?

torchvision/prototype/features/_bounding_box.py
datumbox: I tried it but the bbox expected a tuple and this needs to be handled explicitly to make mypy happy. I'll revert the previous approach which just handles the integer and maintains the previous treatment on the else but I'm happy to review on the nits where and how this should be handled. We need to discuss where exactly the exception should be thrown in this case. Possibly on the `new_like` method itself?

torchvision/prototype/transforms/_geometry.py
```diff
@@ -545,23 +546,6 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
         )
 
 
-def _setup_float_or_seq(arg: Union[float, Sequence[float]], name: str, req_size: int = 2) -> Sequence[float]:
```
datumbox, line 548: Moving it to utils, to share across files.

torchvision/prototype/transforms/_misc.py
```diff
@@ -112,25 +112,25 @@ def forward(self, *inpts: Any) -> Any:
...
         for ks in self.kernel_size:
             if ks <= 0 or ks % 2 == 0:
                 raise ValueError("Kernel size value should be an odd and positive number.")
 
-        if isinstance(sigma, float):
+        if isinstance(sigma, (int, float)):
             if sigma <= 0:
                 raise ValueError("If sigma is a single number, it must be positive.")
-            sigma = (sigma, sigma)
+            sigma = float(sigma)
```
datumbox, line 126: Since the `_setup_float_or_seq()` will do the job for us, we only need to handle the integer case.

torchvision/prototype/transforms/_utils.py
```diff
@@ -16,6 +16,23 @@
 from typing_extensions import Literal
 
 
+def _setup_float_or_seq(arg: Union[float, Sequence[float]], name: str, req_size: int = 2) -> Sequence[float]:
```
datumbox, line 19: Moved as-is.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -296,6 +301,10 @@ def _affine_bounding_box_xyxy(
 
     if shear is None:
         shear = [0.0, 0.0]
+    elif isinstance(shear, (int, float)):
+        shear = [float(shear), 0.0]
+    elif len(shear) == 1:
+        shear = [shear[0], shear[0]]
```
datumbox, line 307: For the changes on the body of this method, I was tempted to reuse `_affine_parse_args()` but that would probably slow things down, so I opted for porting only the absolutely necessary changes.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -981,13 +990,16 @@ def elastic(
...
-    elif isinstance(output_size, (tuple, list)) and len(output_size) == 1:
-        return [output_size[0], output_size[0]]
+    elif isinstance(output_size, (tuple, list)):
+        if len(output_size) == 1:
+            return [output_size[0], output_size[0]]
+        else:
+            return list(output_size)
     else:
-        return list(output_size)
+        raise ValueError("Invalid output_size")  # Needed for jit
```
datumbox, line 1003: Normally, we should never hit this exception. After just changing the output type n the signature, neither mypy nor JIT can understand that the previous `else` statement activates only when in list. This is because it doesn't recognise that the first `if` statement activates for integers. So we have to refactor it and explicitly assert that we are dealing with lists.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -1150,11 +1162,16 @@ def resized_crop(
...
         size = [int(size), int(size)]
-    elif isinstance(size, (tuple, list)) and len(size) == 1:
-        size = [size[0], size[0]]
+    elif isinstance(size, (tuple, list)):
+        if len(size) == 1:
+            size = [size[0], size[0]]
+        else:
+            size = list(size)
+    else:
+        raise ValueError("Invalid size")  # Needed for jit
```
datumbox, line 1175: Same as above. Refactored because JIT and mypy wouldn't get it.

datumbox (commented):
The main changes are mostly done, though there are lots of styles / nits we might want to see.

Several JIT tests are failing but this might be due to #6635 not being merged yet. Perhaps the following error is related to my comment at https://github.com/pytorch/vision/pull/6635/files#r978934554:
```
test/test_prototype_transforms_functional.py:57 (TestKernels.test_scripted_vs_eager[cpu-resize_image_tensor-167])
  File "test/test_prototype_transforms_functional.py", line 66, in test_scripted_vs_eager
    actual = kernel_scripted(*args, **kwargs)
RuntimeError: resize_image_tensor() Expected a value of type 'Union[List[int], int]' for argument 'size' but instead found type 'tuple'.
Position: 1
Value: (23, 11)
Declaration: resize_image_tensor(Tensor image, Union(int[], int) size, Enum<__torch__.torchvision.transforms.functional.InterpolationMode> interpolation=Enum<InterpolationMode.BILINEAR>, int? max_size=None, bool antialias=False) -> Tensor
Cast error details: Expected a member of Union[List[int], int] but instead found type Tuple[int, int]
```

torchvision/prototype/transforms/functional/_geometry.py
datumbox: All these can be avoided if we decide to return meta-data from the kernels. But I would take this as a separate discussion and follow up PRs.

pmeier: I used it to align the parsing with all other `affine_*` kernels. By using this function we only have a single point for the parsing for all kernels. Unless we detect a bottleneck there, I think that outweighs minimal speed ups.

I've reverted some changes on my PR in 6dfc9657ce89fe9e018a11ee25a8e26c7d3d43c6. Be aware that this is the revert commit, i.e. the diff is inverted.

Apart from using the function, I also refactored it a little to avoid the `None` defaults for most arguments. They are only there since we are also calling this function from `rotate_bounding_box` and there we only have the angle. I opted to simply pass the the current default values explicitly from `rotate_bounding_box` so that `_affine_parse_args()` does not have default values.

datumbox: As discussed offline, I agree with your changes and I should probably examine bringing them here once you merge the test changes. :)

datumbox:
Unfortunately the moment we put all the real types in the signatures (for example `Union[int, List[int]]`) the JIT stops working when provided with a `Tuple[int, int]`. Currently there seems to be some mitigation in JIT that handles `Tuple[int, int]` even when `List[int]` is expected. This mitigation breaks the moment the integers are added. It's a shame we cant define the real types but maintaining JIT-scriptability is more important. I'll be changing back the typing modifications, keeping only other BC-fixes in this PR.

torchvision/prototype/transforms/functional/_geometry.py
datumbox: This is no longer strictly necessary. Perhaps it looks cleaner? Let me know if I should revert

torchvision/prototype/transforms/functional/_geometry.py
datumbox: same here

torchvision/prototype/transforms/functional/_geometry.py
pmeier: I would revert since it is dead code and can never be reached.

torchvision/prototype/transforms/functional/_geometry.py
pmeier: This is still a TODO, right?

datumbox: done

pmeier (approved):

PR #6638, opened 2022-09-23
Link for downloading SBU dataset has changed. Enable the test again

### LINKED ISSUE #6390: SBU download is failing
This is happening since roughly a month (first detection in #6274), but was masked by #6268. 

http://www.cs.virginia.edu/~vicente/sbucaptions/

The server is still up, but the path is no longer valid. I'll reach out to the author to see if this can be fixed. In the meantime we should probably disable the test.

cc @pmeier @YosuaMichael

ambujpawar:
- [x] Closes #6390 

SBU seems to be working again. Therefore, removing it from expected failure test cases


pmeier (approved):
Thanks @ambujpawar! The author never replied if we are allowed to publish the image archive rather than just the URL list. Thus, this is still the best we can do.

pmeier:
@ambujpawar you don't need to regularly update your PR unless GH is showing you merge conflicts. We will update it before merge. Reviews might take some time since we experience some heavy internal workloads.

NicolasHug (approved):
Thanks both

PR #462, opened 2018-04-05
A minor change for transforms.RandomCrop

anguelos:
I just added an optional parameter that pads an image if needed so that small errors don't raise exceptions.
The option is by default turned off so that API behavior is exactly the same.
As far as I saw the test cases don't test for Exceptions so i did not add a test for the feature being off but I tested that locally.

PR #6640, opened 2022-09-23
refactor `prototype.transforms,RandomCrop`

pmeier:
This PR eliminates two `F.pad` calls in `RandomCrop` by computing the full padding upfront. IMO, the implementation of `_get_params` is also more clear now, but you shouldn't trust that since I'm the author of the patch :stuck_out_tongue: 

Since we are only touching the padding here, we can compare the old and new transform by ignoring the cropping altogether. Plus, since the only random behavior happens for the crop coordinate selection, we also don't need to pay attention to random behavior.

```py
import itertools
import unittest.mock

import torch

from torchvision.prototype import transforms


input_size = (32, 17)

for size, padding, pad_if_needed in itertools.product(
    [64, 16],
    [None, 5],
    [True, False],
):
    if any(size > dim for dim in input_size) and not pad_if_needed:
        continue

    old_transform = OldRandomCropWithoutCrop(size, padding=padding, pad_if_needed=pad_if_needed)
    new_transform = transforms.RandomCrop(size, padding=padding, pad_if_needed=pad_if_needed)

    image = torch.rand(3, *input_size)

    old_output = old_transform(image)

    with unittest.mock.patch("torchvision.prototype.transforms._geometry.F.crop", new=lambda x, *_, **__: x):
        new_output = new_transform(image)

    torch.testing.assert_close(new_output, old_output, rtol=0, atol=0)
```

So everything works out perfectly? No :disappointed: 

Replacing multiple `F.pad` calls with a single one only works if the later ones don't depend on the earlier ones. In general, this is only given for `padding_mode="constant"` and `padding_mode="edge"`. 

We will use this "1D" gradient image for simplification:

input

Let's look at three examples for `padding_mode="reflect"`

### :heavy_check_mark: `padding` only

old: 

old_padding

new:

new_padding

### :heavy_check_mark: `pad_if_needed` only

old: 

old_pad_if_needed

new:

new_pad_if_needed

### :heavy_multiplication_x: `padding` and `pad_if_needed`

old: 

old_padding_pad_if_needed

new:

new_padding_pad_if_needed

Since the old implementation pads twice independently from each other, the second pad reflects the reflection created by the first pad. IMO, the new implementation is doing the right thing here and only reflecting once. 

My guess is that this was just an oversight in the original stable implementation. Given that this is an obscure use case anyway, i.e. setting a fixed padding as well as using the dynamic one, and is not a problem at all for the most dominant `padding_mode="constant"`, I would be ok to break BC here in favor of performance and simplified implementation. One could also argue that this can be considered a bug fix although I'm not keen on porting this change back.

### DIFF
--- a/torchvision/prototype/transforms/_geometry.py
+++ b/torchvision/prototype/transforms/_geometry.py
@@ -414,75 +414,92 @@ class RandomCrop(Transform):
             _check_padding_mode_arg(padding_mode)
 
         self.padding = padding
+        self._parsed_padding = F._geometry._parse_pad_padding(padding) if padding else None
         self.pad_if_needed = pad_if_needed
         self.fill = _setup_fill_arg(fill)
         self.padding_mode = padding_mode
 
     def _get_params(self, sample: Any) -> Dict[str, Any]:
-        _, height, width = query_chw(sample)
+        _, padded_height, padded_width = query_chw(sample)
 
         if self.padding is not None:
-            # update height, width with static padding data
-            padding = self.padding
-            if isinstance(padding, Sequence):
-                padding = list(padding)
-            pad_left, pad_right, pad_top, pad_bottom = F._geometry._parse_pad_padding(padding)
-            height += pad_top + pad_bottom
-            width += pad_left + pad_right
+            pad_left, pad_right, pad_top, pad_bottom = self._parsed_padding
+            padded_height += pad_top + pad_bottom
+            padded_width += pad_left + pad_right
+        else:
+            pad_left = pad_right = pad_top = pad_bottom = 0
 
-        output_height, output_width = self.size
-        # We have to store maybe padded image size for pad_if_needed branch in _transform
-        input_height, input_width = height, width
+        cropped_height, cropped_width = self.size
+        height_diff = padded_height - cropped_height
+        width_diff = padded_width - cropped_width
 
         if self.pad_if_needed:
-            # pad width if needed
-            if width < output_width:
-                width += 2 * (output_width - width)
-            # pad height if needed
-            if height < output_height:
-                height += 2 * (output_height - height)
+            if height_diff < 0:
+                # This is a micro optimization that avoids some duplicate computation down the line.
+                #
+                # Here, we want to pad the top and bottom by the difference of the `cropped_height` (output) and the
+                # `padded_height` (input). Since the definition of `height_diff` is inverted we check for a negative
+                # value.
+                #
+                # Since we are padding the height by exactly twice the difference, the resulting `height_diff`, if we
+                # were to compute it again after the padding, is its exact inverse:
+                #
+                # new_height_diff =   new_padded_height                                            - cropped_height
+                #                 =   old_padded_height + 2 * (cropped_height - old_padded_height) - cropped_height
+                #                 = - old_padded_height + cropped_height
+                #                 = -(old_padded_height - cropped_height)
+                #                 = - old_height_diff
+                #
+                # Without this trick, all `+=` would need to be changed to `-=` and the re-definition of `height_diff`
+                # would need to moved below the re-definition of `padded_height` and be:
+                #
+                # height_diff = padded_height - cropped_height
+                #
+                # Although this only replaces one subtraction with the unary `-` in this branch, this trick also avoids
+                # re-defining `height_diff` anywhere else.
+                height_diff = -height_diff
 
-        if height < output_height or width < output_width:
+                pad_top += height_diff
+                pad_bottom += height_diff
+                padded_height += 2 * height_diff
+
+            if width_diff < 0:
+                # Same as above by substituting 'width' for 'height'.
+                width_diff = -width_diff
+
+                pad_left += width_diff
+                pad_right += width_diff
+                padded_width += 2 * width_diff
+
+        if height_diff < 0 or width_diff < 0:
             raise ValueError(
-                f"Required crop size {(output_height, output_width)} is larger then input image size {(height, width)}"
+                f"Required crop size {(cropped_height, cropped_width)} is larger then "
+                f"{'padded ' if self.padding is not None else ''}input image size {(padded_height, cropped_width)}."
             )
 
-        if width == output_width and height == output_height:
-            return dict(top=0, left=0, height=height, width=width, input_width=input_width, input_height=input_height)
+        # We need a different order here than we have in self._parsed_padding, since this padding will be parsed again
+        # in self._transform()
+        padding = [pad_left, pad_top, pad_right, pad_bottom]
+        needs_pad = any(padding)
 
-        top = torch.randint(0, height - output_height + 1, size=(1,)).item()
-        left = torch.randint(0, width - output_width + 1, size=(1,)).item()
+        top = int(torch.randint(0, height_diff + 1, size=())) if height_diff > 0 else 0
+        left = int(torch.randint(0, width_diff + 1, size=())) if width_diff > 0 else 0
 
         return dict(
             top=top,
             left=left,
-            height=output_height,
-            width=output_width,
-            input_width=input_width,
-            input_height=input_height,
+            height=cropped_height,
+            width=cropped_width,
+            needs_pad=needs_pad,
+            padding=padding,
         )
 
     def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
-        # TODO: (PERF) check for speed optimization if we avoid repeated pad calls
         fill = self.fill[type(inpt)]
         fill = F._geometry._convert_fill_arg(fill)
 
-        if self.padding is not None:
-            # This cast does Sequence[int] -> List[int] and is required to make mypy happy
-            padding = self.padding
-            if not isinstance(padding, int):
-                padding = list(padding)
-
-            inpt = F.pad(inpt, padding=padding, fill=fill, padding_mode=self.padding_mode)
-
-        if self.pad_if_needed:
-            input_width, input_height = params["input_width"], params["input_height"]
-            if input_width < self.size[1]:
-                padding = [self.size[1] - input_width, 0]
-                inpt = F.pad(inpt, padding=padding, fill=fill, padding_mode=self.padding_mode)
-            if input_height < self.size[0]:
-                padding = [0, self.size[0] - input_height]
-                inpt = F.pad(inpt, padding=padding, fill=fill, padding_mode=self.padding_mode)
+        if params["needs_pad"]:
+            inpt = F.pad(inpt, padding=params["padding"], fill=fill, padding_mode=self.padding_mode)
 
         return F.crop(inpt, top=params["top"], left=params["left"], height=params["height"], width=params["width"])
 
--- a/torchvision/prototype/transforms/_geometry.py
+++ b/torchvision/prototype/transforms/_geometry.py
@@ -414,7 +414,7 @@ class RandomCrop(Transform):
             _check_padding_mode_arg(padding_mode)
 
         self.padding = padding
-        self._parsed_padding = F._geometry._parse_pad_padding(padding) if padding else None
+        self._parsed_padding = F._geometry._parse_pad_padding(padding) if padding else None  # type: ignore[arg-type]
         self.pad_if_needed = pad_if_needed
         self.fill = _setup_fill_arg(fill)
         self.padding_mode = padding_mode
@@ -422,7 +422,7 @@ class RandomCrop(Transform):
     def _get_params(self, sample: Any) -> Dict[str, Any]:
         _, padded_height, padded_width = query_chw(sample)
 
-        if self.padding is not None:
+        if self._parsed_padding is not None:
             pad_left, pad_right, pad_top, pad_bottom = self._parsed_padding
             padded_height += pad_top + pad_bottom
             padded_width += pad_left + pad_right
--- a/test/test_prototype_transforms.py
+++ b/test/test_prototype_transforms.py
@@ -715,30 +715,38 @@ class TestRandomCrop:
 
         if padding is not None:
             if isinstance(padding, int):
-                h += 2 * padding
-                w += 2 * padding
+                pad_top = pad_bottom = pad_left = pad_right = padding
             elif isinstance(padding, list) and len(padding) == 2:
-                w += 2 * padding[0]
-                h += 2 * padding[1]
+                pad_left = pad_right = padding[0]
+                pad_top = pad_bottom = padding[1]
             elif isinstance(padding, list) and len(padding) == 4:
-                w += padding[0] + padding[2]
-                h += padding[1] + padding[3]
+                pad_left, pad_top, pad_right, pad_bottom = padding
 
-        expected_input_width = w
-        expected_input_height = h
+            h += pad_top + pad_bottom
+            w += pad_left + pad_right
+        else:
+            pad_left = pad_right = pad_top = pad_bottom = 0
 
         if pad_if_needed:
             if w < size[1]:
-                w += 2 * (size[1] - w)
+                diff = size[1] - w
+                pad_left += diff
+                pad_right += diff
+                w += 2 * diff
             if h < size[0]:
-                h += 2 * (size[0] - h)
+                diff = size[0] - h
+                pad_top += diff
+                pad_bottom += diff
+                h += 2 * diff
+
+        padding = [pad_left, pad_top, pad_right, pad_bottom]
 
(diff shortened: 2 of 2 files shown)


datumbox:
@pmeier Nice work! I do agree that the behaviour you describe on the 3rd case looks more like a bug/omission on the original implementation rather than an intentional behaviour. I also never seen anyone in practice doing the corner-case transform you describe on the 3rd case. I was hoping if you could provide some measurements on the speed difference between the 2 implementations?

@fmassa / @vfdev-5 Do you have an opinion on whether the above is an intended behaviour or just an omission/bug on the cornercase of the previous implementation?

Another approach we could take if we think the previous behaviour was intentional is to apply the speed optimization only when possible (for example when we use `constant`) which will probably account for 99.999% of real-world use-cases. This would probably have implications on the code-quality (we would have a very complex and difficult to maintain implementation). To avoid this we could split the Transform into 2 private Transform classes and use internally the optimal one based on the constructor configuration (that's not an idiom we use currently but we can discuss the details).

pmeier:
> I was hoping if you could provide some measurements on the speed difference between the 2 implementations?

There are only two cases where we could hit more than one pad:

1. Setting `padding` and `pad_if_needed` at the same time and having an input that triggers `pad_if_needed` in at least one dimension.
2. Setting only `pad_if_needed` and having an input that triggers it in both dimensions.

Since 1. is quite unrealistic, I focused on 2. I also used `padding_mode="constant"`, because that should account for most of the real world usages.

<details><summary>Code</summary>
<p>

Run this snippet on the `main` branch as well on the branch of this PR just changing the `description`:

```py
import pickle
import itertools

import torch

from torchvision.prototype import transforms

description = "main"  # "main", "PR"

measurements = []
for input_size, output_size_factor in itertools.product(
    [64, 256, 1024],
    [1.0, 1.5, 2.0],
):
    image = torch.randint(0, 256, (3, input_size, input_size), dtype=torch.uint8)

    output_size = int(input_size * output_size_factor)
    transform = transforms.RandomCrop(output_size, pad_if_needed=True)

    timer = benchmark.Timer(
        stmt="transform(image)",
        globals=dict(
            transform=transform,
            image=image,
        ),
        label="RandomCrop with padding needed on both sides",
        sub_label=f"{input_size:4d} -> {output_size}",
        description=description,
    )

    measurements.append(timer.blocked_autorange(min_run_time=5))

with open(f"{description}.measurements", "wb") as fh:
    pickle.dump(measurements, fh)
```

Afterwards run this snippet for the results:

```py
import pathlib
import pickle

measurements = []
for file in pathlib.Path(".").glob("*.measurements"):
    with open(file, "rb") as fh:
        measurements.extend(pickle.load(fh))

comparison.trim_significant_figures()
```

</p>
</details>

```
[ RandomCrop with padding needed on both sides ]
                    |   main  |   PR 
1 threads: --------------------------
        64 -> 64    |      9  |     9
        64 -> 96    |     66  |    49
        64 -> 128   |     88  |    65
       256 -> 256   |     10  |     9
       256 -> 384   |    389  |   256
       256 -> 512   |    716  |   523
      1024 -> 1024  |      9  |     9
      1024 -> 1536  |   5390  |  3600
      1024 -> 2048  |  10500  |  7860

```

In the tested scenario, this PR gives us roughly a 30%-50% perf boost. As for the real impact of this, it is hard to say. Only the segmentation and video classification references are using it:

https://github.com/pytorch/vision/blob/7046e56fe4370e94339b3e8b6fd011e285294a3a/references/segmentation/presets.py#L15

https://github.com/pytorch/vision/blob/7046e56fe4370e94339b3e8b6fd011e285294a3a/references/video_classification/presets.py#L22

In case of the segmentation references, the `RandomCrop` is actually differs from what we have in our prototype namespace there by the padding performed before the cropping. See https://github.com/pytorch/vision/pull/6433#discussion_r971846663 for details.

However, we discussed offline to use it directly there since it uses a more reasonable padding strategy. This gives us the opportunity to actually compute an estimate boost in a real setting.

<details><summary>Code</summary>
<p>

```py
import torch

# https://github.com/pytorch/vision/blob/7046e56fe4370e94339b3e8b6fd011e285294a3a/references/segmentation/train.py#L34
base_size = 520
crop_size = 480

dataset = datasets.CocoDetection(
    "/home/philip/datasets/coco/train2017",
    "/home/philip/datasets/coco/annotations/instances_train2017.json",
)


input_sizes = torch.tensor([(image.height, image.width) for image, _ in dataset])

input_smaller_size = torch.min(input_sizes, dim=-1, keepdim=True).values
torch.manual_seed(0)
random_resize_factor = (
    torch.distributions.Uniform(0.5, 2.0).sample(input_smaller_size.shape) * base_size / input_smaller_size
)

random_resized_input_sizes = (input_sizes * random_resize_factor).int()

needs_pad = random_resized_input_sizes < crop_size

needs_one_pad = needs_pad.any(dim=-1)
print(f"{int(torch.sum(needs_one_pad)) / len(dataset):.1%} of the images need one padding")

needs_two_pads = needs_pad.all(dim=-1)
print(f"{int(torch.sum(needs_two_pads)) / len(dataset):.1%} of the images need two paddings")

input_shapes_two_pads = random_resized_input_sizes[needs_two_pads.unsqueeze(1).repeat(1, 2)].reshape(-1, 2)
print(f"Of those the median shape is {tuple(torch.median(input_shapes_two_pads.float(), dim=0).values.int().tolist())}")
```

</p>
</details>

```
28.1% of the images need one padding
10.9% of the images need two paddings
Of those the median shape is (318, 399)
```

Re-running our benchmarks from above with inputs of shape `(318, 399)` and `size=480` yields

```
[ RandomCrop with padding needed on both sides ]
                                |   main  |   PR
1 threads: -------------------------------------
      (318, 399) -> (480, 480)  |  500.2  |  336

```

So we are looking at an ~50% decrease in ~10% of the cases or a total of a 5% decrease. In total though we are saving about only 2 seconds per epoch.

To conclude, although this PR actually speeds up the transform, in a real world scenario the patch makes no significant difference. It is a nice code clean up though.

torchvision/prototype/transforms/_geometry.py
```diff
@@ -414,75 +414,92 @@ def __init__(
             _check_padding_mode_arg(padding_mode)
 
         self.padding = padding
+        self._parsed_padding = F._geometry._parse_pad_padding(padding) if padding else None  # type: ignore[arg-type]
```
vfdev-5, line 417: This can lead to unexpected behaviour if user updates dynamically `self.padding` : 
```
t = RandomCrop(...)
t.padding = 12
```

pmeier: True. Depending if we want the speed boost of not computing this in every call, we could also make padding a `@property` and recompute only if it is set, e.g.

```py
class RandomCrop:
    def __init__(self, padding):
        self._padding = None
        self._parsed_padding = None
        self.padding = padding

    @property
    def padding(self):
        return self._padding

    @padding.setter
    def padding(self, padding):
        self._padding = padding
        self._parsed_padding = _parse_pad_padding(padding)

```

fmassa:
@datumbox 
> Do you have an opinion on whether the above is an intended behaviour or just an omission/bug on the cornercase of the previous implementation?

This was a bug in behavior I believe, and would only happen on really very few cases, so I would be ok breaking BC here.
For most cases where users apply `T.Resize` beforehand they don't need this option, as discussed in https://github.com/pytorch/vision/pull/462

torchvision/prototype/transforms/_geometry.py
vfdev-5: Maybe, `_parse_pad_padding(padding)` is not that slow.
On the other hand we already overriding input arg when we assign them to `self`, e.g `_setup_size` thus we may parse padding in the ctor and directly assign the output to `self.padding` ?

datumbox: > This can lead to unexpected behaviour if user updates dynamically self.padding

This is true not only for this but also for most other transform properties. Perhaps it's worth making read-only (using properties without setters). Shall we have a separate RFC for future handling of this?

pmeier: > On the other hand we already overriding input arg when we assign them to `self`, e.g `_setup_size` thus we may parse padding in the ctor and directly assign the output to `self.padding` ?

I think that is a strong argument. We shouldn't special case `padding` here.

> This is true not only for this but also for most other transform properties. Perhaps it's worth making read-only (using properties without setters). Shall we have a separate RFC for future handling of this?

It would make the implementation more complex in general. IMO we should wait until someone actually requests this and otherwise not go for it. As @vfdev-5 mentioned, we are already overwriting user input for some parameters, so `padding` would just continue to do so.

torchvision/prototype/transforms/_geometry.py
```diff
@@ -414,75 +414,92 @@ def __init__(
...
+            padded_width += pad_left + pad_right
+        else:
+            pad_left = pad_right = pad_top = pad_bottom = 0
 
-        output_height, output_width = self.size
-        # We have to store maybe padded image size for pad_if_needed branch in _transform
-        input_height, input_width = height, width
+        cropped_height, cropped_width = self.size
+        height_diff = padded_height - cropped_height
+        width_diff = padded_width - cropped_width
```
vfdev-5, line 434: Reading all below comments and the code I still do not quite get why we could not define `height_diff = cropped_height - padded_height` such that we do not need to do `height_diff = -height_diff`

pmeier: In the sampling of `top` and `left` L485-L486 we need `padded - cropped`. So if we do `cropped - padded` here, we need to invert for the regular case if `pad_if_needed` is not set.

Since we need both definitions for the diff, we need to invert (or recompute) at least once. With the speed optimization in mind, I put the invert into the optional branch.

vfdev-5: Seems strange. Stable implementation does not do that. Also we have if raise clause for cases when top/left can't be sampled if difference is negative or zero...

pmeier: > Seems strange. Stable implementation does not do that.

Yes, it does. In the `pad_if_needed` branch we do

https://github.com/pytorch/vision/blob/06d9726d66f97ed8e3d3910da246a52be9c6e168/torchvision/prototype/transforms/_geometry.py#L440

and later for the `top` and `left` sampling we do

https://github.com/pytorch/vision/blob/06d9726d66f97ed8e3d3910da246a52be9c6e168/torchvision/prototype/transforms/_geometry.py#L454

> Also we have if raise clause for cases when top/left can't be sampled if difference is negative or zero...

1. If the difference is negative, we never get to the sampling and raise an error earlier the same as the stable transform.
2. If the difference is zero, we also have this branch as ternary statement in the sampling:

    ```py
    top = int(torch.randint(0, height_diff + 1, size=())) if height_diff > 0 else 0
    left = int(torch.randint(0, width_diff + 1, size=())) if width_diff > 0 else 0
    ```

torchvision/prototype/transforms/_geometry.py
```diff
@@ -414,75 +414,92 @@ def __init__(
...
+
+                pad_left += width_diff
+                pad_right += width_diff
+                padded_width += 2 * width_diff
+
+        if height_diff < 0 or width_diff < 0:
             raise ValueError(
-                f"Required crop size {(output_height, output_width)} is larger then input image size {(height, width)}"
+                f"Required crop size {(cropped_height, cropped_width)} is larger then "
+                f"{'padded ' if self.padding is not None else ''}input image size {(padded_height, cropped_width)}."
```
vfdev-5, line 477:
```suggestion
                f"{'padded ' if self.padding is not None else ''}input image size {(padded_height, padded_width)}."
```

torchvision/prototype/transforms/_geometry.py
vfdev-5: OK, i checked in details the implementation and now understand the trick with `diff=-diff`. Well, I think it would be better to keep code more transparent without this micro-optimization... I think we can afford computing once again the diff once pad_if_needed branch is applied...

pmeier: I've removed the trick in 5aff3b759, while keeping the large comment in tact for better comparability. The comment would need to be adapted if we want to keep the commit.

Removing the trick leads to three more branches and IMO does not really enhances readability. I'll let you be the judge.

vfdev-5: We can do something like:
```
        cropped_height, cropped_width = self.size        
                
        if self.pad_if_needed:
            if cropped_height > padded_width:
                height_diff = cropped_height - padded_height
                pad_top += height_diff
                pad_bottom += height_diff
                padded_height += 2 * height_diff

            if cropped_width > padded_width:
                width_diff = cropped_width - padded_width                        
                pad_left += width_diff
                pad_right += width_diff
                padded_width += 2 * width_diff

        height_diff = padded_height - cropped_height
        width_diff = padded_width - cropped_width
```

pmeier: Good call. Let me send a patch.

torchvision/prototype/transforms/_geometry.py
```diff
@@ -413,76 +413,69 @@ def __init__(
...
+        padding = [pad_left, pad_top, pad_right, pad_bottom]
+        needs_pad = any(padding)
 
-        top = torch.randint(0, height - output_height + 1, size=(1,)).item()
-        left = torch.randint(0, width - output_width + 1, size=(1,)).item()
+        if padded_height > cropped_height or padded_width > cropped_width:
+            top = int(torch.randint(0, padded_height - cropped_height + 1, size=()))
+            left = int(torch.randint(0, padded_width - cropped_width + 1, size=()))
+        else:
+            top = left = 0
```
pmeier, line 462: In a previous version I checked the height and width diff individually before sampling and just setting `0` if there is no diff. This is a speed-up since in this case the call looks like `int(torch.randint(0, 1, size=()))` which is always `0`.

However, checking individually is different from what we do now in prototype and in stable:

https://github.com/pytorch/vision/blob/2d92728341bbd3dc1e0f1e86c6a436049bbb3403/torchvision/transforms/transforms.py#L634-L639

Thus, this could be regarded as BC breaking change since we would change the sampling in some cases. For example, this triggered our consistency tests since there we don't have a height diff, but call `torch.randint` twice since there is a width diff. If we check separately, we only call `torch.randint` once and thus get a different parameter set even if we set the random seed before the call.

Thoughts?

datumbox: This is a tangent comment:
https://github.com/pytorch/vision/blob/2d92728341bbd3dc1e0f1e86c6a436049bbb3403/torchvision/transforms/transforms.py#L634-L635

@vfdev-5 These are some optimizations we can do on the new Transforms to avoid calling the resize kernel if not needed. The previous implementation reaches resize which is unnecessary. Not 100% sure if this is a no-op on the C++ side but we should make sure we don't do this on the next Transforms (maybe it's fixed already?).


For the question at hand, I understand that the new implementation does fewer unnecessary random calls. Though you are right to say that this is going to make our life harder for testing the transforms (because the seeds won't be the same), I agree with Victor that there is no BC issue here. If I misunderstood your position, please let me know.

pmeier: > Though you are right to say that this is going to make our life harder for testing the transforms (because the seeds won't be the same), I agree with Victor that there is no BC issue here.

No, this is correct. I'll fix the test and revert back to separate checking the diffs.

torchvision/prototype/transforms/_geometry.py
```diff
@@ -414,76 +414,68 @@ def __init__(
...
+
+        if padded_height < cropped_height or padded_width < cropped_width:
             raise ValueError(
-                f"Required crop size {(output_height, output_width)} is larger then input image size {(height, width)}"
+                f"Required crop size {(cropped_height, cropped_width)} is larger than "
+                f"{'padded ' if self.padding is not None else ''}input image size {(padded_height, padded_width)}."
             )
 
-        if width == output_width and height == output_height:
-            return dict(top=0, left=0, height=height, width=width, input_width=input_width, input_height=input_height)
```
datumbox, line 452: This logic allows us to avoid doing the crop when not necessary. Let's refactor to keep and perhaps introduce a `needs_crop` boolean to skip cropping all together if unnecessary. Other than that, I don't have other comments. It would be good for @vfdev-5 to have one last pass to ensure there are no surprises.

datumbox (approved):
LGTM, thanks!

PR #6641, opened 2022-09-23
Ignore deprecation warning from traverse function

ejguan:
Since https://github.com/pytorch/pytorch/pull/84079 is landed eventually, the CI for prototype tests will turn green tmrw when PyTorch nightly is updated. However, this PR introduces a `FutureWarning` that will deprecate `only_datapipe` and the default behavior of `traverse` will be changed as `only_datapipe=True`. And, this PR is suppressing this Warning turning to Error in the test.


pmeier (changes requested):
@ejguan there is only one test that currently uses `only_datapipe=False`:

https://github.com/pytorch/vision/blob/7046e56fe4370e94339b3e8b6fd011e285294a3a/test/test_prototype_datasets_builtin.py#L104-L106

This test together with 

https://github.com/pytorch/vision/blob/7046e56fe4370e94339b3e8b6fd011e285294a3a/test/test_prototype_datasets_builtin.py#L112

and

https://github.com/pytorch/vision/blob/7046e56fe4370e94339b3e8b6fd011e285294a3a/test/test_prototype_datasets_builtin.py#L125

are only there to make sure our datasets work seamlessly with all `torchdata` utilities and don't error or throw warnings. If `only_datapipe=False` is going to be deprecated, we can simply remove the parametrization here instead of ignoring the warning. cc @NicolasHug

Since you are already looking at this: do these three test still cover all utilities that might interact with our datasets from your side or should we add or remove something? :innocent:

NicolasHug:
>  If only_datapipe=False is going to be deprecated, we can simply remove the parametrization here instead of ignoring the warning. cc @NicolasHug

KI dl

NicolasHug:
> If only_datapipe=False is going to be deprecated, we can simply remove the parametrization here instead of ignoring the warning. cc @NicolasHug

I agree, ignoring the warning will make our tests break when the parameter is removed for good. We should remove its usage now instead.

ejguan:
> > If only_datapipe=False is going to be deprecated, we can simply remove the parametrization here instead of ignoring the warning. cc @NicolasHug
> 
> I agree, ignoring the warning will make our tests break when the parameter is removed for good. We should remove its usage now instead.

For sure,  I will do a quick update by removing `only_datapipe` shortly.

pmeier (approved):
@ejguan I took the liberty to make the changes to fix our CI.

NicolasHug (approved):
Thank you both, LGTM

PR #6642, opened 2022-09-25
Fixing inverted center_crop check on Classification preset

datumbox:
Resolves https://github.com/pytorch/vision/pull/6342/files#r979421762

### DIFF
--- a/references/classification/presets.py
+++ b/references/classification/presets.py
@@ -18,10 +18,11 @@ class ClassificationPresetTrain:
         random_erase_prob=0.0,
         center_crop=False,
     ):
-        trans = (
-            [transforms.RandomResizedCrop(crop_size, interpolation=interpolation)]
+        trans = []
+        trans.append(
+            transforms.CenterCrop(crop_size)
             if center_crop
-            else [transforms.CenterCrop(crop_size)]
+            else transforms.RandomResizedCrop(crop_size, interpolation=interpolation)
         )
         if hflip_prob > 0:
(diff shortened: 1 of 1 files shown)


NicolasHug (approved):
LGTM but since we use the `bug` label for release notes and this isn't a released bug, I'd suggest to remove the label to avoid confusion

TeodorPoncu (approved):
Looks good to me, this should replicate 1:1 the setting in which the model was trained.

datumbox:
@TeodorPoncu Thanks for the review and for providing the details to verify how the model was trained. I'll add a note to to the checkpoint area to clarify that the log incorrectly reports the training argument and that the correct command is fixed in this PR. On my side, I also independently verified that `--train-center-crop` wasn't used for this model by looking back on each commit prior training.

@NicolasHug I think the tag discussion was brought up before. I understand that by tagging this as a bug, our automated script that prepares the release notes will put it under the bugs section and that requires manual cleaning (cc @YosuaMichael who is the POC of the upcoming release). One issue is that the bug tag is used to track how many bugs we introduce and solve overall. This has implications on tracking the dashboards that TorchVision maintains, the Better Engineering work, the upstream Core breakages and the classification of cherrypicks of the release (which again might be introduced only on development and might not pre-exist on previous releases). Perhaps we can avoid this by introducing an additional tag to filter these out? If you feel strongly about the tag in this PR, feel free to remove it and we can discuss it on a future weekly.

PR #6646, opened 2022-09-26
[PoC] move metadata computation from prototype features into kernels

pmeier:
As discussed offline. Moving the computation to the kernels has several upsides:

1. We would have the computation in one place rather than scattered on the kernel as well as on the feature. In some cases that might also be a performance improvement since we don't need to recompute intermediate values.
2. Users that want to use the prototype kernels directly currently don't have access to the metadata we compute and thus have to do it themselves. This somewhat defeats the idea of providing kernels for all the transforms since we only provide part of the computation.

The only downside that I see is that now some kernels don't have a more complex output. Since there is no hard rule that the kernels need to return a single tensor, this should not effect anything but our test suite. The test suite currently at some places depends on a single tensor return, but this is trivial to change. I have not touched the tests yet, so they are currently failing.

@vfdev-5 also suggested that the kernels could have a `return_metadata: bool = False` flag. Using this, the users by default will always only get a tensor back, while they have the option to get the metadata as well. I'm usually against changing the return type through an input flag, but the consistent return type for all kernels is also appealing. However, if we want to go for that, we need to check if JIT is happy as well. The annotation, e.g. `Union[torch.Tensor, Tuple[torch.Tensor, Tuple[int, int]]]` for the bounding box kernels from this PR, should not be an issue, but I'm not sure if JIT can work with the output, i.e.

```py
output1 = kernel(...)
output2, metadata = kernel(..., return_metadata=True)
```

If it can't, I'm strongly against this variant, because we then need more workarounds or the like that make the API more complex than a few kernels that return more than just a tensor.


torchvision/prototype/features/_bounding_box.py
```diff
@@ -140,23 +131,10 @@ def rotate(
         fill: FillTypeJIT = None,
         center: Optional[List[float]] = None,
     ) -> BoundingBox:
-        output = self._F.rotate_bounding_box(
+        output, image_size = self._F.rotate_bounding_box(
             self, format=self.format, image_size=self.image_size, angle=angle, expand=expand, center=center
         )
-        image_size = self.image_size
-        if expand:
-            # The way we recompute image_size is not optimal due to redundant computations of
```
datumbox, line 148: And an argument for adopting this, is the potential performance gains (and the obvious improvement in code-quality) from avoid recomputing staff. It would be great to get a benchmark on this to see what kind of difference we see. Previously we noted that bbox operations are a bottleneck.

datumbox (commented):
Overall I'm in favour of this approach. I think the `return_metadata=True` idea might bring some further speed gains but I don't believe it's JIT compatible (we do a conditional return on Object Detection models an it's exceptionally hacky to make it JIT-scriptable).

I think before merging this, we need to fix all the JIT-scriptability related PRs we have to ensure nothing breaks it. 

@pmeier Are these the only kernels that need to return meta-data or it's just an example and there are more to be changed?

torchvision/prototype/features/_bounding_box.py
```diff
@@ -118,19 +118,10 @@ def pad(
...
-
-        # Update output image size:
-        left, right, top, bottom = self._F._geometry._parse_pad_padding(padding)
-        height, width = self.image_size
-        height += top + bottom
-        width += left + right
-
-        return BoundingBox.new_like(self, output, image_size=(height, width))
+        output, image_size = self._F.pad_bounding_box(
+            self, padding, format=self.format, image_size=self.image_size, padding_mode=padding_mode
```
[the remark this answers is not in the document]
pmeier: They already don't have aligned signatures. `pad` doesn't take `format`, which is specific for bounding boxes. So this just adds more metadata that needs to be passed.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -144,11 +144,14 @@ def resize_mask(mask: torch.Tensor, size: List[int], max_size: Optional[int] = N
 
 def resize_bounding_box(
     bounding_box: torch.Tensor, size: List[int], image_size: Tuple[int, int], max_size: Optional[int] = None
-) -> torch.Tensor:
+) -> Tuple[torch.Tensor, Tuple[int, int]]:
```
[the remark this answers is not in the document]
pmeier: `BoundinBox.image_size` is a `Tuple[int, int]`:

https://github.com/pytorch/vision/blob/c4c28dff8a68fae20eb7f2a82051ef7a388353bd/torchvision/prototype/features/_bounding_box.py#L20

Given that these kernels / features are new do we really want to introduce the wrong types here as well? The `List[int]` was previously only used for the "requested" size on some kernels like `resize` or `crop`. That is of course somewhat related to this, but not the same.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -530,9 +533,22 @@ def rotate_bounding_box(
...
-    ).view(original_shape)
+    if expand:
+        # The way we recompute image_size is not optimal due to redundant computations of
+        # - rotation matrix (_get_inverse_affine_matrix)
+        # - points dot matrix (_compute_affine_output_size)
+        # Alternatively, we could return new image size by self._F.rotate_bounding_box
+        height, width = image_size
+        rotation_matrix = _get_inverse_affine_matrix([0.0, 0.0], angle, [0.0, 0.0], 1.0, [0.0, 0.0])
+        new_width, new_height = _FT._compute_affine_output_size(rotation_matrix, width, height)
+        image_size = (new_height, new_width)
```
pmeier, line 544: @vfdev-5 we should be able to get the `new_width, new_height` out of `_affine_bounding_box_xyxy` without extra / with minimal computation, right?

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -673,15 +689,16 @@ def pad_mask(
 
 def pad_bounding_box(
     bounding_box: torch.Tensor,
-    padding: Union[int, List[int]],
```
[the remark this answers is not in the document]
pmeier: We are not really consistent throughout the API, but most of the times, the metadata is placed directly after the input:

https://github.com/pytorch/vision/blob/7046e56fe4370e94339b3e8b6fd011e285294a3a/torchvision/prototype/transforms/functional/_geometry.py#L29-L31

No strong opinion, but having them as close as possible to the input makes the most sense to me. That would mean that stuff that was previously stored in a feature is still "together" even for the individual variables. Plus, after the metadata the signature of the other type kernels is the same.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -718,7 +740,9 @@ def crop_bounding_box(
     format: features.BoundingBoxFormat,
     top: int,
     left: int,
-) -> torch.Tensor:
+    height: int,
+    width: int,
```
pmeier, line 744: This is a little awkward: we only pass `height` and `width` here to return them unchanged. Thus, although the `image_size` is "decomposed" here, this is mostly belongs in the category of kernels that shouldn't return the `image_size` since it is not changed inside the kernel.

However, `crop_bounding_box` was an outlier in the first place, since all other `crop_*` kernels take the `height` and `width` parameter:

https://github.com/pytorch/vision/blob/c4c28dff8a68fae20eb7f2a82051ef7a388353bd/torchvision/transforms/functional_tensor.py#L132

https://github.com/pytorch/vision/blob/7046e56fe4370e94339b3e8b6fd011e285294a3a/torchvision/prototype/transforms/functional/_geometry.py#L735

https://github.com/pytorch/vision/blob/7046e56fe4370e94339b3e8b6fd011e285294a3a/torchvision/prototype/transforms/functional/_geometry.py#L739

Thus, with this change we are actually aligning the kernels although it is a little awkward for the bounding box one. No strong opinion here.

pmeier (commented):
> Are these the only kernels that need to return meta-data or it's just an example and there are more to be changed?

No, these are all kernels that need to be changed.

torchvision/prototype/features/_bounding_box.py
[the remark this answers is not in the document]
datumbox: As I mention below, I am not against doing this. Just mentioning it as an argument against it. Also the real issue is not removing parameters but adding more. Our previous policy was that the `F` would have the maximal parameters and the kernels would just use what they need; also `F` aligns with the method dispatchers. This addition makes the rule more complex. Not much we can do about it though.

torchvision/prototype/transforms/functional/_geometry.py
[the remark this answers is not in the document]
datumbox: I think it makes sense to keep consistent the behaviour. In the signature of this method (which is a broader problem, not just on your changes), we define `size: List[int]` while `image_size: Tuple[int, int]`. That looks weird IMO. I think this was missed on the clean ups and is a separate discussion.

But I agree with your argument that the return type matches the input type of `image_size`.

torchvision/prototype/transforms/functional/_geometry.py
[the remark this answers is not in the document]
datumbox: Yeah it would be good to align these details. They are minor nits but polishing the solution and fixing this is worth it.

torchvision/prototype/transforms/functional/_geometry.py
datumbox: Good call. No strong opinions on my side.

torchvision/prototype/transforms/functional/_geometry.py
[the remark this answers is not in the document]
pmeier: Any preferences where the metadata should go in the signature?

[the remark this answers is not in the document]
datumbox: No preference, as long as it is consistent across signatures. Your call.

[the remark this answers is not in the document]
pmeier: See #6648.

test/prototype_transforms_kernel_infos.py
```diff
@@ -1552,8 +1554,6 @@ def reference_inputs_ten_crop_image_tensor():
             skips=[
                 skip_integer_size_jit(),
                 Skip("test_batched_vs_single", reason="Custom batching needed for five_crop_image_tensor."),
-                Skip("test_no_inplace", reason="Output of five_crop_image_tensor is not a tensor."),
```
pmeier, line 1555: I was able to remove these, since the test framework no longer expects the output to be a tensor.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -90,15 +90,21 @@ def vertical_flip(inpt: features.InputTypeJIT) -> features.InputTypeJIT:
 vflip = vertical_flip
 
 
+def _compute_resized_output_size(
+    image_size: Tuple[int, int], size: List[int], max_size: Optional[int] = None
+) -> List[int]:
+    if isinstance(size, int):
+        size = [size]
```
pmeier, line 97: I took the liberty to put this into the common function that is called from every kernel.

torchvision/prototype/transforms/functional/_geometry.py
pmeier: Given that there is no regression in performance here, i.e. I only moved code around, I'll not fix that here but rather open an issue to mark this as a potential performance improvement.

torchvision/prototype/features/_bounding_box.py
```diff
@@ -112,29 +108,19 @@ def resized_crop(
...
 
     def pad(
         self,
         padding: Union[int, Sequence[int]],
         fill: FillTypeJIT = None,
         padding_mode: str = "constant",
     ) -> BoundingBox:
-        # This cast does Sequence[int] -> List[int] and is required to make mypy happy
-        if not isinstance(padding, int):
-            padding = list(padding)
```
datumbox, line 127: Do we need this?

torchvision/prototype/transforms/functional/_geometry.py
datumbox: Could you please add an in-place performance TODO?

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -518,7 +521,7 @@ def rotate_bounding_box(
     angle: float,
     expand: bool = False,
     center: Optional[List[float]] = None,
-) -> torch.Tensor:
+) -> Tuple[torch.Tensor, Tuple[int, int]]:
```
datumbox, line 524: I'm not going to insist on this but it's kinda weird that sometimes we define `size` as `List[int]` and others `image_size` as `Tuple[int, int]`. I do agree that the latter would have been preferable but due to JIT we can't do it. Shall we align everywhere in a separate PR?

datumbox (approved):
LGTM, with a few minor comments.

@vfdev-5 could you please also have a look?

torchvision/prototype/features/_bounding_box.py
pmeier: I guess not since the comment states that this was only added for `mypy`, but it seems happy now.

torchvision/prototype/transforms/functional/_geometry.py
pmeier: > Shall we align everywhere in a separate PR?

As discussed offline, we are not sure whether we should or not. I'll open an issue soon with a proposal to drop the wrong `List[int]` from all kernels in favor of the correct annotations.

Since this is not blocking, I'm going ahead with `Tuple[int, int]` as return type and we can revisit after we have resolved the discussion in the issue to be created.

PR #6647, opened 2022-09-26
close streams in prototype datasets

pmeier:
Redo of #6128 with the changes suggested there by me:

- Instead of closing every single `EncodedImage.from_file(buffer); buffer.close()` individually, I moved the closing into `EncodedImage.from_file`. So far we we don't have a case where we need to read from the same stream twice. Even if we at some point do, e.g. optical flow (although we are probably better of with caching a single image), we can always introduce a keyword `close: bool = True` to overwrite the behavior.
- The patch above accounts for the largest chunk of changes in #6128. There are a few datasets that handle other files as well and I closed them directly. Plus, I also added a `.close()` call into `read_mat`
- I've added a test that checks that a dataset doesn't leave unclosed streams after a full iteration.

### DIFF
--- a/test/test_prototype_datasets_builtin.py
+++ b/test/test_prototype_datasets_builtin.py
@@ -1,6 +1,7 @@
 import functools
 import io
 import pickle
+from collections import deque
 from pathlib import Path
 
 import pytest
@@ -11,10 +12,11 @@ from torch.utils.data import DataLoader
 from torch.utils.data.graph import traverse
 from torch.utils.data.graph_settings import get_all_graph_pipes
 from torchdata.datapipes.iter import ShardingFilter, Shuffler
+from torchdata.datapipes.utils import StreamWrapper
 from torchvision._utils import sequence_to_str
-from torchvision.prototype import datasets, transforms
+from torchvision.prototype import datasets, features, transforms
 from torchvision.prototype.datasets.utils._internal import INFINITE_BUFFER_SIZE
-from torchvision.prototype.features import Image, Label
+
 
 assert_samples_equal = functools.partial(
     assert_equal, pair_types=(TensorLikePair, ObjectPair), rtol=0, atol=0, equal_nan=True
@@ -23,6 +25,17 @@ assert_samples_equal = functools.partial(
 
 def extract_datapipes(dp):
     return get_all_graph_pipes(traverse(dp, only_datapipe=True))
+
+
+def consume(iterator):
+    # Copied from the official itertools recipes: https://docs.python.org/3/library/itertools.html#itertools-recipes
+    deque(iterator, maxlen=0)
+
+
+def next_consume(iterator):
+    item = next(iterator)
+    consume(iterator)
+    return item
 
 
 @pytest.fixture(autouse=True)
@@ -66,7 +79,7 @@ class TestCommon:
         dataset, _ = dataset_mock.load(config)
 
         try:
-            sample = next(iter(dataset))
+            sample = next_consume(iter(dataset))
         except StopIteration:
             raise AssertionError("Unable to draw any sample.") from None
         except Exception as error:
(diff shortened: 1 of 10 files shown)


test/test_prototype_datasets_builtin.py
```diff
@@ -25,6 +27,17 @@ def extract_datapipes(dp):
     return get_all_graph_pipes(traverse(dp, only_datapipe=True))
 
 
+def consume(iterator):
+    # Copied from the official itertools recipes: https://docs.python.org/3/library/itertools.html#itertools-recipes
+    deque(iterator, maxlen=0)
+
+
+def next_consume(iterator):
```
pmeier, line 35: This is useful since we often just want the first sample, but need to make sure to still consume to avoid dangling streams. `list(iterator)` would also do the trick, but keeps everything in memory for no reason.

test/test_prototype_datasets_builtin.py
```diff
@@ -84,22 +97,53 @@ def test_num_samples(self, dataset_mock, config):
 
         assert len(list(dataset)) == mock_info["num_samples"]
 
+    @pytest.fixture
+    def log_session_streams(self):
+        debug_unclosed_streams = StreamWrapper.debug_unclosed_streams
```
pmeier, line 102: @VitalyFedyunin Although this is called "debug", can we depend on this being there and public?

test/test_prototype_datasets_builtin.py
```diff
@@ -84,22 +97,53 @@ def test_num_samples(self, dataset_mock, config):
...
+        consume(iter(dataset))
+
+        if StreamWrapper.session_streams:
+            raise AssertionError(make_msg_and_close("The following streams were not closed after a full iteration:"))
+
+    @parametrize_dataset_mocks(DATASET_MOCKS)
+    def test_no_simple_tensors(self, dataset_mock, config):
+        dataset, _ = dataset_mock.load(config)
+
+        simple_tensors = {key for key, value in next_consume(iter(dataset)).items() if features.is_simple_tensor(value)}
```
pmeier, line 133: This is a drive-by since I was already touching the line: the term "vanilla" tensor is no longer used. In the prototype transforms we use "simple tensor" now and also have `features.is_simple_tensor` to check for them.

torchvision/prototype/datasets/_builtin/sbd.py
```diff
@@ -103,8 +103,11 @@ def _datapipe(self, resource_dps: List[IterDataPipe]) -> IterDataPipe[Dict[str,
             buffer_size=INFINITE_BUFFER_SIZE,
             drop_none=True,
         )
-        if self._split == "train_noval":
-            split_dp = extra_split_dp
+        split_dp, to_be_closed_dp = (
+            (extra_split_dp, split_dp) if self._split == "train_noval" else (split_dp, extra_split_dp)
+        )
+        for _, file in to_be_closed_dp:
+            file.close()
```
pmeier, line 110: This is somewhat problematic in the `split == "train_noval"` case. We want to close all file handles that are coming from the `split_dp`. Unfortunately, `split_dp` comes from a `Demultiplexer`. Thus, by fully iterating over it here, we are loading everything into the demux buffer.

Is there an idiom to "mark" a datapipe to be closed at runtime even if we don't return the datapipe? The only thing I came up with is changing the classifier function of the `Demultiplexer` to drop the samples that would go into `split_dp` if `split == "train_noval"`.

torchvision/prototype/datasets/utils/_internal.py
```diff
@@ -40,10 +39,9 @@ def read_mat(buffer: BinaryIO, **kwargs: Any) -> Any:
     except ImportError as error:
         raise ModuleNotFoundError("Package `scipy` is required to be installed to read .mat files.") from error
 
-    if isinstance(buffer, StreamWrapper):
-        buffer = buffer.file_obj
```
pmeier, line 44: This is no longer needed for quite some time, since `StreamWrapper` no correctly exposes all methods of the `file_obj` and thus can be used directly in `sio.loadmat`.

torchvision/prototype/datasets/_builtin/sbd.py
VitalyFedyunin: It is incorrect to consume Datapipe in construction time. 

```python

        for _, file in to_be_closed_dp:
            file.close()

```

It will not affect executed graph.

Ideally, we want to have something like dp.close(), which will effectively remove dangling pieces of the graph.

But for now you can either use trick like `split_dp = split_dp.concatinate(to_be_closed_dp.filter(lambda x: False))` or do code branching before Demux and avoid creating dangling pieces.

pmeier: > It is incorrect to consume Datapipe in construction time.

I don't think that is fully correct for our case. At construction, we load a couple of files and weave them together into one datapipe. These files are always loaded unconditionally, but for some configurations not all of the files are needed. So we should be able to simply close them during the construction of the dataset datapipe, since they will never make it in the graph, correct?

I agree, we shouldn't do this for datapipes that stem for a `Demultiplexer` if the other parts make it into the final graph.

pmeier: > do code branching before Demux and avoid creating dangling pieces.

If possible, I think that is the better solution, since the other still iterates over all items when the loop should actually be done. I guess that could be irritating as well.

I implemented branching in afb0ec2e718b74d96f4641812e574601b9fbd39b. PTAL

VitalyFedyunin (approved):

torchvision/prototype/datasets/_builtin/sbd.py
```diff
@@ -93,18 +97,24 @@ def _prepare_sample(self, data: Tuple[Tuple[Any, Tuple[str, BinaryIO]], Tuple[st
...
+            archive_dp, split_dp = resource_dps
+            images_dp, anns_dp = Demultiplexer(
+                archive_dp,
+                2,
+                self._classify_archive,
+                buffer_size=INFINITE_BUFFER_SIZE,
+                drop_none=True,
+            )
+        else:
+            archive_dp = resource_dps[0]
```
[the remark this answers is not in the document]
pmeier: If `if self._split != "train_noval"`, we have only one element in `resource_dps`. Meaning, it will not be loaded at all and thus also does not need to be closed.

PR #6652, opened 2022-09-27
fix mypy errors after the 0.981 release

pmeier:
This fixes the ongoing `mypy` CI failure caused by yesterdays `0.981` release

### DIFF
--- a/torchvision/models/_api.py
+++ b/torchvision/models/_api.py
@@ -112,10 +112,7 @@ def get_weight(name: str) -> WeightsEnum:
     return weights_enum.from_str(value_name)
 
 
-W = TypeVar("W", bound=WeightsEnum)
-
-
-def get_model_weights(name: Union[Callable, str]) -> W:
+def get_model_weights(name: Union[Callable, str]) -> WeightsEnum:
     """
     Retuns the weights enum class associated to the given model.
 
@@ -125,10 +122,10 @@ def get_model_weights(name: Union[Callable, str]) -> W:
         name (callable or str): The model builder function or the name under which it is registered.
 
     Returns:
-        weights_enum (W): The weights enum class associated with the model.
+        weights_enum (WeightsEnum): The weights enum class associated with the model.
     """
     model = get_model_builder(name) if isinstance(name, str) else name
-    return cast(W, _get_enum_from_fn(model))
+    return _get_enum_from_fn(model)
 
 
 def _get_enum_from_fn(fn: Callable) -> WeightsEnum:
@@ -199,7 +196,7 @@ def list_models(module: Optional[ModuleType] = None) -> List[str]:
     return sorted(models)
 
 
-def get_model_builder(name: str) -> Callable[..., M]:
+def get_model_builder(name: str) -> Callable[..., nn.Module]:
     """
     Gets the model name and returns the model builder method.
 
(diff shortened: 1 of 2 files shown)


torchvision/models/_api.py
```diff
@@ -199,7 +196,7 @@ def list_models(module: Optional[ModuleType] = None) -> List[str]:
     return sorted(models)
 
 
-def get_model_builder(name: str) -> Callable[..., M]:
+def get_model_builder(name: str) -> Callable[..., nn.Module]:
```
NicolasHug, line 199: How of curiosity why do we need to change this one, and not all the other usages of `M` e.g. in `register_model()` above?

torchvision/prototype/datasets/utils/_internal.py
```diff
@@ -72,8 +72,8 @@ def _getattr_closure(obj: Any, *, attrs: Sequence[str]) -> Any:
     return obj
 
 
-def _path_attribute_accessor(path: pathlib.Path, *, name: str) -> D:
-    return cast(D, _getattr_closure(path, attrs=name.split(".")))
+def _path_attribute_accessor(path: pathlib.Path, *, name: str) -> Any:
+    return _getattr_closure(path, attrs=name.split("."))
```
NicolasHug, line 76: Why is this change needed?
Do you remember why we annotated the return with `D` in the first place?

NicolasHug (approved):

NicolasHug:
BTW do the changes still pass with older versions of mypy?

torchvision/models/_api.py
pmeier: `TypeVar` is used to create a "link" between parameters. For example I can do something like

```py
def do_nothing(foo: M) -> M:
    return foo
```

This tells `mypy` "whatever we have as input type, we return as output type as well". This is useful in case the function can handle multiple types, e.g.

```py
T = TypeVar("T", int, float)

def add(a: T, b: T) -> T:
    return a + b
``` 

In the case above, there is no input with the type `M` and so `mypy` complains that it can't figure out the `M` in the output. 

Correct me if I'm wrong @datumbox, but I guess the annotation was intended to mean "we return any kind of model" here. Since `M` is defined as 

https://github.com/pytorch/vision/blob/06d9726d66f97ed8e3d3910da246a52be9c6e168/torchvision/models/_api.py#L168

we can simply use `nn.Module` here.

pmeier:
> BTW do the changes still pass with older versions of mypy?

Most likely, but I'll check. The annotations were wrong to begin with, but older versions of `mypy` didn't flag them yet. The new annotations should be correct and thus older versions should also accept them.

torchvision/prototype/datasets/utils/_internal.py
pmeier: > Why is this change needed?

Same as above. We no longer can use a `TypeVar` as return annotation if we don't have any input with the same type.

> Do you remember why we annotated the return with `D` in the first place?

IIRC, this is a remnant when this function was a local one inside 

https://github.com/pytorch/vision/blob/06d9726d66f97ed8e3d3910da246a52be9c6e168/torchvision/prototype/datasets/utils/_internal.py#L83

In there the annotation was correct and ok, since we have a `D` in the input parameters. At some point we factored them out since `torchdata` doesn't like local functions, and I probably forgot to fix this since `mypy` didn't complain.

datumbox (approved):
LGTM, thanks!

PR #6653, opened 2022-09-27
use pytest markers instead of custom solution for prototype transforms functional tests

pmeier:
In #6626 we added support for skipping tests through the `KernelInfo` and `DispatcherInfo`. It was a custom solution that relied on some fixture magic

https://github.com/pytorch/vision/blob/55a436cb6bf1bb79111f6080fa6e087b6b1fbf01/test/test_prototype_transforms_functional.py#L29-L35

Plus, there were other downsides:

1. We were only able to skip a test, but not xfail for example.
2. The test to skip was only identified by its name, but multiple classes in the test module could have the same test function names, leading to clashes.

This PR changes this by replacing our custom skip logic with the ability to add `pytest.mark`'s to a specific `info` / `args_kwargs` combination.


test/prototype_transforms_dispatcher_infos.py
```diff
@@ -15,11 +17,24 @@
...
+
+    test_marks: Sequence[TestMark] = dataclasses.field(default_factory=list)
+    _test_marks_map: Dict[str, List[TestMark]] = dataclasses.field(default=None, init=False)
 
     def __post_init__(self):
-        self._skips_map = {skip.test_name: skip for skip in self.skips}
+        test_marks_map = defaultdict(list)
+        for test_mark in self.test_marks:
+            test_marks_map[test_mark.test_id].append(test_mark)
+        self._test_marks_map = dict(test_marks_map)
```
pmeier, line 30: This is the same on `KernelInfo` and here on the `DispatcherInfo`. I'm going to factor this out into a common base class in a follow-up PR since this PR is already quite large.

test/prototype_transforms_kernel_infos.py
```diff
@@ -2,13 +2,16 @@
...
+from collections import defaultdict
+from typing import Any, Callable, Dict, Iterable, List, Optional, Sequence, Tuple
 
 import numpy as np
 import pytest
 import torch.testing
 import torchvision.ops
 import torchvision.prototype.transforms.functional as F
+
+from _pytest.mark.structures import MarkDecorator
```
pmeier, line 14: This is only needed for annotating the attributes of the data classes below. I'll send a follow-up PR removing them, since the upside of `@dataclasses.dataclass` is outweighed by the downside of importing from a private namespace.

test/prototype_transforms_kernel_infos.py
```diff
@@ -17,11 +20,24 @@
 __all__ = ["KernelInfo", "KERNEL_INFOS"]
 
 
+TestID = Tuple[Optional[str], str]
```
pmeier, line 23: Same as above.

test/prototype_transforms_kernel_infos.py
```diff
@@ -17,11 +20,24 @@
...
+
+
+def mark_framework_limitation(test_id, reason):
+    # The purpose of this function is to have a single entry point for skip marks that are only there, because the test
+    # framework cannot handle the kernel in general or a specific parameter combination.
+    # As development progresses, we can change the `mark.skip` to `mark.xfail` from time to time to see if the skip is
+    # still justified.
+    # We don't want to use `mark.xfail` all the time, because that actually runs the test until an error happens. Thus,
+    # we are wasting CI resources for no reason for most of the time
+    return TestMark(test_id, pytest.mark.skip(reason=reason))
```
pmeier, line 40: By changing this to 

```suggestion
    return TestMark(test_id, pytest.mark.xfail(reason=reason))
```
 
temporarily, I actually found a unnecessary skip although we barely have any at the moment. See below for details.

test/prototype_transforms_kernel_infos.py
```diff
@@ -1449,15 +1471,16 @@ def reference_inputs_ten_crop_image_tensor():
             sample_inputs_fn=sample_inputs_five_crop_image_tensor,
             reference_fn=pil_reference_wrapper(F.five_crop_image_pil),
             reference_inputs_fn=reference_inputs_five_crop_image_tensor,
-            skips=[
-                Skip(
-                    "test_scripted_vs_eager",
-                    condition=lambda args_kwargs, device: isinstance(args_kwargs.kwargs["size"], int),
-                    reason="Integer size is not supported when scripting five_crop_image_tensor.",
+            test_marks=[
```
pmeier, line 1474: We no longer need to skip `TestKernels.test_no_inplace` since there was a bug in the test.

test/test_prototype_transforms_functional.py
```diff
@@ -125,7 +153,7 @@ def test_no_inplace(self, info, args_kwargs, device):
         input_version = input._version
         output = info.kernel(input, *other_args, **kwargs)
 
-        assert output is not input or output._version == input_version
+        assert output is not input or input._version == input_version
```
pmeier, line 156: We actually want to check that the `_version` of the `input` has not changed after executing the kernel.

vfdev-5 (approved):
Thanks @pmeier

PR #6658, opened 2022-09-27
Use real image in test_detection_model

### LINKED ISSUE #6655: fasterrcnn_resnet50_fpn Linux GPU tests failing on CUDA 11.6
### 🐛 Describe the bug

Similar to: https://github.com/pytorch/vision/issues/6589

After the removal of CUDA 10.2 and the setting of 11.6 as default the tests at `fasterrcnn_resnet50_fpn` started [failing](https://app.circleci.com/pipelines/github/pytorch/vision/20644/workflows/95f54863-15d6-43ac-8c12-306c9c7ecd7d/jobs/1679689/steps) for python 3.8:

```
  File "/home/circleci/project/test/test_models.py", line 777, in check_out
  File "/home/circleci/project/test/test_models.py", line 117, in _assert_expected
  File "/home/circleci/project/env/lib/python3.8/site-packages/torch/testing/_comparison.py", line 1342, in assert_close
    assert_equal(
  File "/home/circleci/project/env/lib/python3.8/site-packages/torch/testing/_comparison.py", line 1093, in assert_equal

Mismatched elements: 59 / 80 (73.8%)
Greatest absolute difference: 223.815927028656 at index (12, 2) (up to 0.012 allowed)
Greatest relative difference: inf at index (3, 1) (up to 0.012 allowed)

The failure occurred for item [0]['boxes']

  File "/home/circleci/project/test/test_models.py", line 805, in test_detection_model
    full_validation &= check_out(out, autocast_custom_prec.get(model_name, 0.01))
  File "/home/circleci/project/test/test_models.py", line 785, in check_out
    torch.testing.assert_close(
  File "/home/circleci/project/env/lib/python3.8/site-packages/torch/testing/_comparison.py", line 1342, in assert_close
    assert_equal(
  File "/home/circleci/project/env/lib/python3.8/site-packages/torch/testing/_comparison.py", line 1093, in assert_equal

Mismatched elements: 2 / 20 (10.0%)
Greatest absolute difference: 0.02667921781539917 at index (16,) (up to 0.012 allowed)
Greatest relative difference: 0.03139688501900038 at index (18,) (up to 0.012 allowed)
```

This failure occurs on Linux

### Versions

PR: https://github.com/pytorch/vision/pull/6649

cc @datum @malfet @ptrblck 

### Versions

latest nightly

Aidyn-A:
This is an attempt to fix #6655 and #6589.
The main idea behind this pull request is to use a real image instead of random tensor.
The use of real image decreases the accumulation of numerical errors in AMP (half precision) operations.
The following models were modified to use a real image in `test_classification_model` and `test_detection_model` on all devices:
```
retinanet_resnet50_fpn
retinanet_resnet50_fpn_v2
keypointrcnn_resnet50_fpn
fasterrcnn_resnet50_fpn
fasterrcnn_resnet50_fpn_v2
fcos_resnet50_fpn
maskrcnn_resnet50_fpn
maskrcnn_resnet50_fpn_v2
```


ptrblck:
Good idea to try to use a real image. However, I would also try to use a pretrained model (detection models should have been trained on a COCO dataset) to make sure the actual predictions are viable. 

CC @datumbox

datumbox:
@ptrblck This is a solid strategy that we are currently examining on our side. See #6608

In this case, you would need to re-estimate the expected files of the affected models and commit them. This is done by setting `EXPECTTEST_ACCEPT=1` and running the tests at question. This will overwrite the existing expected files.

@YosuaMichael I was wondering if you can share thoughts on your experiments for using real data and weights on the models. Have you managed to combat the flakiness? Is this a direction we will adopt? It seems aligned with the recommendation from our Nvidia partners.

YosuaMichael:
I have tried to do the test with real image and real weight in https://github.com/pytorch/vision/pull/6608 (currently it has a lot of changes because I did revert after updating the branch) . However unfortunately, I still see flakiness in the CI. 

Few things that I observed during the testing:
- At first I found that the test can be robust when we also increase the threshold of the score to `0.4` this will let the model only output around 3 boxes and it can pass all platform in CI
- However in the next day, I found some test failed (I tried to debug with reverting to the code that pass, but still failed). I am not too sure what happened on this and haven't spend too much time on investigating the reason

Aidyn-A:
Hi @datumbox and @YosuaMichael,

This PR is passing all of the CUDA unittests from [today](https://github.com/pytorch/vision/runs/8683668549) and [3 days ago](https://github.com/pytorch/vision/runs/8643391426).
Would it be safe to merge it? I know it will cause some conflicts with #6608, but these changes are pretty short and can be easily modified. We would rather having the fix for #6655 and #6589 sooner than later.

datumbox (approved):
The changes LGTM. Thanks a lot @Aidyn-A for your investigation and work.

There is an on-going work on another PR to add support of real weights and images but I'm OK merging this first to assist your further investigation. I'll leave it to @YosuaMichael to review and decide how to proceed.

YosuaMichael (approved):
Thanks @Aidyn-A , the PR looks good. And no worry about conflicting #6608 since thats only experimental (will create a new PR to try out using real weight combined with this PR later).

I am quite curious on why using the real image (without real weight) could ` decreases the accumulation of numerical errors in AMP (half precision) operations.`. Would be great if you can give more insight in this!

Aidyn-A:
Hi @YosuaMichael thanks for reviewing and approving this PR. 
> I am quite curious on why using the real image (without real weight) could decreases the accumulation of numerical errors in AMP (half precision) operations.

This is a great question. The simplest explanation I can provide is that the real image has more small values than the random one. If we take a look at the histogram of the random image, we will see that there are almost the same amount of pixels with the values of 0.9 and with the values 0.1 in all channels:

In the real image there are much more pixels with 0.1, than those with 0.9:

The rounded off part of the half precision number is getting smaller and as the result the accumulated error is smaller.

The use of real weights will improve the situation even further. Nevertheless, the random weights do not corrupt the overall error because they are initialized with non-uniform random distribution and their values are small enough to maintain the round off error small.

PR #6660, opened 2022-09-28
Temporarily disable the autocast test for `fasterrcnn_resnet50_fpn`

datumbox:
Related to https://github.com/pytorch/vision/issues/6655
Done to unblock https://github.com/pytorch/vision/pull/6649
Alternative approaches at https://github.com/pytorch/vision/pull/6658 https://github.com/pytorch/vision/pull/6608

This PR:
- Reverts a previous mitigation introduced by @malfet for Windows as now we also need it on Linux
- Adds the tests on the list that skips autocast tests

It's very strange that `fasterrcnn_resnet50_fpn` is impacted but `fasterrcnn_resnet50_fpn_v2` is not. There are differences between the models but only in terms of the number of CNNs each model has. So it's very likely the issue is on a kernel which is affected by a CUDA 11.6 change.


atalman (approved):
LGTM

datumbox:
The [failing](https://github.com/pytorch/vision/actions/runs/3143382981/jobs/5108056143) tests on prototype are unrelated and already patched on a previous PR. Merging.

Issue #6662, opened 2022-09-28
compatibility layer between stable datasets and prototype transforms?


pmeier:
The original plan was to roll out the datasets and transforms revamp at the same time since they somewhat depend on each other. However, it is becoming more and more likely that the prototype transforms will be finished sooner. Thus, we need some compatibility layer in the meantime. This issue explains how transforms are currently used with the datasets, what will or will not work without a compatibility layer, and how such a compatibility layer might look like.

## Status quo

Most of our datasets support the `transform` and `target_transform` idiom. These transformations are applied separately to the first and second item of the raw sample returned by the dataset. For classification tasks this usually sufficient although I've never seen a practical use for `target_transform`:

https://github.com/pytorch/vision/blob/0fcfaa13d1a1402cc9788dbaeeadc47032cf214b/references/classification/train.py#L129-L137

However, the separation of the transforms breaks down in case image and label need to be transformed at the same time, e.g. `CutMix` or `MixUp`. They are currently applied through a custom collation function for the dataloader:

https://github.com/pytorch/vision/blob/3c9ae0ac5dda3d881a6ce5004ce5756ae7de7bc4/references/classification/train.py#L209-L210

Since these transforms do not work with the standard idioms, they never made it out of our references into the library.

The need to transform input and target at the same time is not a special case for other tasks such as segmentation or detection. Datasets for these tasks support the `transforms` parameter. It will be called with the complete sample and thus is able to support all use cases. 

Since even datasets for the same task have very diverse outputs, there were only two options without revamping the APIs completely:

1. Unify the datasets outputs on the dataset itself.
2. Unify the datasets outputs through a compatibility layer.

When this first came up in the past, we went with option 2. On our references we unified the output for a few select datasets for a specific task, so we can apply custom joint transformations to them. Since we didn't want to commit to the interface, neither the minimal compatibility layer nor the transformations made it into the library. Thus, although some of our datasets in theory support joint transformations, the users have to implement them themselves.   

## Do we need a compatibility layer?

The new transformations support the joint use case out of the box. Meaning, all the custom transformations from our references are now part of the library. Plus, all transformations that previously only supported images, e.g. resizing or padding, now also support bounding boxes, masks and so on.

The information which part of the sample is what kind of type is not communicated through the sample structure, i.e. first element is an image and second one is a mask, but rather through the actual type of the object. We introduced several tensor subclasses that will be rolled out together with the transforms.

[...]

PR #6663, opened 2022-09-28
[PoC] compatibility layer between stable datasets and prototype transforms

pmeier:
This is the proof of concept implementation for #6662. Let's keep the discussion there unless there is need for discussion on a technical issue of the proposal.


torchvision/prototype/features/__init__.py
```diff
@@ -13,3 +14,5 @@
 )
 from ._label import Label, OneHotLabel
 from ._mask import Mask
+
+from ._dataset_wrapper import VisionDatasetFeatureWrapper  # usort: skip
```
pmeier, line 18: I've put it under `torchvision.prototype.features` for now, because I think that is the best place as long as the `features` are still in the prototype namespace. After roll-out, we could certainly also put this under `torchvision.datasets`

torchvision/prototype/features/_dataset_wrapper.py
```diff
@@ -0,0 +1,388 @@
...
+            return wrappers_fn
+
+        return register
+
+    @classmethod
+    def from_torchvision_dataset(
+        cls,
+        dataset: datasets.VisionDataset,
+        *,
+        keep_pil_image: bool = False,
```
pmeier, line 100: This flag gives the user the option to keep images as PIL if they like. I've use `False` as default to promote the new functionality. In our experiments we saw that doing everything as tensors is on par with PIL regarding speed and accuracy.

torchvision/prototype/features/_dataset_wrapper.py
```diff
@@ -0,0 +1,388 @@
...
+        bounding_box_format: Optional[features.BoundingBoxFormat] = None,
```
pmeier, line 101: This parameter gives users the option to select the format. It could be useful so users always get bounding boxes in the "right" format for them. However, we have a 

https://github.com/pytorch/vision/blob/b8af91ab78d9e95ac377e059c7c47249128d5bdb/torchvision/prototype/transforms/_meta.py#L10

transform that does the same. I've not made up my mind yet.

torchvision/prototype/features/_dataset_wrapper.py
```diff
@@ -0,0 +1,388 @@
...
+        dtypes: Optional[Dict[Type[features._Feature], Optional[torch.dtype]]] = None,
+    ) -> VisionDatasetFeatureWrapper:
+        dtypes: Dict[Type[features._Feature], Optional[torch.dtype]] = {
+            features.Image: torch.uint8,
+            features.Label: torch.int64,
+            features.Mask: torch.uint8,
+            features.BoundingBox: torch.float32,
+            features.GenericFeature: None,
+            **(dtypes or dict()),
+        }
```
pmeier, line 111: This gives users the ability to decide what dtype each feature should have. This is somewhat different than the `bounding_box_format` above, since the initial value is dictated by the dataset, whereas the dtype is often "free" since we are creating a tensor for the first time. By choosing the right dtype from the beginning, we can avoid copies later.

torchvision/prototype/features/_dataset_wrapper.py
NicolasHug: It does look a bit redundant with the transforms.

Perhaps I'm pushing it a little, but: why have a `bounding_box_format` but not a `image_size` (assuming we agree `image_size` isn't something we want :) ! )

torchvision/prototype/features/_dataset_wrapper.py
NicolasHug: > By choosing the right dtype from the beginning, we can avoid copies later.

For both Images and Masks which are the only datapoints that are somewhat expensive to convert, we still call `F.convert_image_dtype` under the hood so this isn't saving any copy.

In lack of clear performance benefit, I'd err towards letting users just use the `ToDtype()` transform?

torchvision/prototype/features/_dataset_wrapper.py
```diff
@@ -0,0 +1,388 @@
...
+    *, keep_pil_image: bool, dtypes: Dict[Type[features._Feature], Optional[torch.dtype]]
+) -> Callable[[PIL.Image.Image], Union[PIL.Image.Image, features.Image]]:
+    if keep_pil_image:
+        return identity_wrapper
+
+    return functools.partial(wrap_image, dtype=dtypes[features.Image])
+
+
+def wrap_label(label: Any, *, categories: Optional[Sequence[str]], dtype: Optional[torch.dtype]) -> features.Label:
+    return features.Label(label, categories=categories, dtype=dtype)
```
NicolasHug, line 142: Curious how this work for test sets where the label might be `None`?

torchvision/prototype/features/_dataset_wrapper.py
```diff
@@ -0,0 +1,388 @@
...
+def classification_wrappers(
+    dataset: datasets.VisionDataset,
+    keep_pil_image: bool,
+    bounding_box_format: Optional[features.BoundingBoxFormat],
+    dtypes: Dict[Type[features._Feature], Optional[torch.dtype]],
+) -> Any:
+    return (
+        make_image_wrapper(keep_pil_image=keep_pil_image, dtypes=dtypes),
+        make_label_wrapper(categories=CATEGORIES_GETTER[type(dataset)](dataset), dtypes=dtypes),
+    )
```
NicolasHug, line 187: I'm wondering how much of this machinery is needed. At its core all we need for classification datasets is something like this (ignoring dtype and keep_pil_image which hopefully don't add too much complexity)


```py
class ClassificationWrapper(VisionDataset):
    def __getitem__(self, idx):
        image, target = super().__getitem__(idx)
        return Image(image), Target(target)
```


Granted I'm cheating a bit and this doesn't support everything the PR supports. But I'm wondering where all the extra code / logic in the PR comes from. We have to jump through a bunch of builders, classes, methods, helpers and partials to do something that should seem pretty basic.

Is this so that we can handle more complex cases like segmentation similarly? Are there opportunities for simplification?

torchvision/prototype/features/_dataset_wrapper.py
pmeier: On a second thought, we don't want disrupt peoples workflows with this wrapper. So we should at least have `True` as default or remove the flag completely. Users can always put a PIL -> Tensor transform into their pipeline without any extra cost compared to doing it here.

torchvision/prototype/features/_dataset_wrapper.py
pmeier: > Perhaps I'm pushing it a little, but: why have a `bounding_box_format` but not a `image_size` (assuming we agree `image_size` isn't something we want :) ! )

I guess you mean `spatial_size`

https://github.com/pytorch/vision/blob/5dd95944c609ac399743fa843ddb7b83780512b3/torchvision/prototype/datapoints/_bounding_box.py#L18-L20

right? If so, we can't make this a free parameter, because it is tied to the image in the same sample. The format is not tied to anything.

That being said, we no longer need this. I'm not sure what the state was when I wrote the comment, but the API is now agnostic with regards to bounding box format. Plus, if the user actually wants to force a specific format, they can drop a

https://github.com/pytorch/vision/blob/5dd95944c609ac399743fa843ddb7b83780512b3/torchvision/prototype/transforms/_meta.py#L11

into their transforms pipeline.

torchvision/prototype/features/_dataset_wrapper.py
pmeier: Yes. The only times this makes some sense is for datapoints that don't have a fixed starting dtype, e.g. bounding boxes which are often `List[int]` or `List[float]`. But, I agree, an extra copy of a `(N, 4)` tensor with `N < 20` is probably not something worth some extra API complexity.

torchvision/prototype/features/_dataset_wrapper.py
pmeier: It doesn't so far, thanks for the heads-up!

torchvision/prototype/features/_dataset_wrapper.py
pmeier: > Is this so that we can handle more complex cases like segmentation similarly? Are there opportunities for simplification?

Most likely. Let me have another look now that I haven't done so in a few months. I tend to overcomplicate stuff in the development phase. Will ping you when I have addressed the comments above and made it simpler.

pmeier:
dbfac05fa75496937ad82ba7341c12a0e4dac565 is a complete of the internals. Here are the main points:

- Change the main entry point from the classmethod `VisionDatasetDatapointWrapper.from_torchvision_dataset` to a standalone function `wrap_dataset_for_transforms_v2` (naming TBD). With this we can make the class private and thus users should have no touching points with the internals. On top of that it simplified the implementation quite a bit.
- Drop support for additional parameters of the wrapper to specify the dtypes of the wrapped objects, the bounding box format, and whether we want to keep images as PIL. See
  - https://github.com/pytorch/vision/pull/6663#discussion_r1087146167
  - https://github.com/pytorch/vision/pull/6663#discussion_r1087153746
  - https://github.com/pytorch/vision/pull/6663#discussion_r1087151023
- Simplify the internal machinery. https://github.com/pytorch/vision/issues/6662#issuecomment-1275821215 established that there is no easy way to provide standalone wrappers for the individual datapoints of a sample. Thus, there is no need to keep the bits that try to do so. This addresses https://github.com/pytorch/vision/pull/6663#discussion_r1086923542 although not as proposed with a standalone class for a task like classification. @NicolasHug PTAL and LMK if my new implementation is simple enough.

torchvision/prototype/features/__init__.py
pmeier: Now under `torchvision.prototype.datapoints`, but the spirit is the same.

torchvision/prototype/features/_feature.py
```diff
@@ -234,3 +234,7 @@ def gaussian_blur(self, kernel_size: List[int], sigma: Optional[List[float]] = N
 
 InputType = Union[torch.Tensor, PIL.Image.Image, _Feature]
 InputTypeJIT = torch.Tensor
+
+
+class GenericFeature(_Feature):
```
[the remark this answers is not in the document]
pmeier: Now named `GenericDatapoint`, but the spirit is the same.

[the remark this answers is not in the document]
NicolasHug: IIUC, this extra class is needed solely because pure Tensors are currently treated as Images, for BC reason.

Perhaps we can rely on a different convention which is that if there are pure Tensors in a sample, only the *first* pure Tensor will be treated as an Image if there's not already an Image; and all the other Tensors will be passed-through.

I feel like this convention would work for every single datasets we currently provide (or we can manually enforce it in the dataset -> transforms wrapper.

For user-defined datasets it should work for the majority of cases. For the cases where it doesn't work, well... Those cases didn't work for our current transforms anyway - so there's no BC consideration to have here.

Whether we would still need a `GenericDatapoint` or allow the base `DataPoint` class to be pass-through is up for discussion. My initial gut feeling would be to not expose that right away and wait for users feedback, but happy to hear others' thoughts on this

torchvision/prototype/datapoints/_dataset_wrapper.py
```diff
@@ -0,0 +1,250 @@
...
+__all__ = ["wrap_dataset_for_transforms_v2"]
+
+_WRAPPERS = {}
+
+
+# TODO: naming!
+def wrap_dataset_for_transforms_v2(dataset: datasets.VisionDataset) -> _VisionDatasetDatapointWrapper:
+    wrapper = _WRAPPERS.get(type(dataset))
+    if wrapper is None:
+        raise TypeError
```
NicolasHug, line 27: NIT: sounds more like a `ValueError` IMHO, the actual type of the dataset has little do to with it

torchvision/prototype/datapoints/_dataset_wrapper.py
```diff
@@ -0,0 +1,250 @@
...
+from typing import Any, Callable, Dict, List, Optional, Tuple
+
+import PIL.Image
+import torch
+from torch.utils.data import Dataset
+from torchvision import datasets
+from torchvision.prototype import datapoints
+from torchvision.prototype.transforms import functional as F
+
+__all__ = ["wrap_dataset_for_transforms_v2"]
```
NicolasHug, line 18: Curious where we're planning to expose this? I guess it'll be in `torchvision.transforms.v2` like all the transforms? In which case the `for_transforms_v2` might be slightly redundant (although I tend to prefer long descriptive names myself...)

torchvision/prototype/datapoints/_dataset_wrapper.py
```diff
@@ -0,0 +1,250 @@
...
+
+    def __getitem__(self, idx: int) -> Any:
+        # This gets us the raw sample since we disabled the transforms for the underlying dataset in the constructor
+        # of this class
+        sample = self.vision_dataset[idx]
+
+        sample = self.wrapper(self.vision_dataset, sample)
+
+        # We don't need to care about `transform` and `target_transform` here since `VisionDataset` joins them into a
+        # `transforms` internally:
```
NicolasHug, line 56: So do we actually need these 2 lines above in `__init__`? 

```py
        self.transform, dataset.transform = dataset.transform, None
        self.target_transform, dataset.target_transform = dataset.target_transform, None
```

Looks like they'd still be available through our custom `__getattr__`?

torchvision/prototype/datapoints/_dataset_wrapper.py
```diff
@@ -0,0 +1,250 @@
...
+        # https://github.com/pytorch/vision/blob/2d92728341bbd3dc1e0f1e86c6a436049bbb3403/torchvision/datasets/vision.py#L52-L54
+        if self.transforms is not None:
+            sample = self.transforms(*sample)
+
+        return sample
+
+    def __len__(self) -> int:
+        return len(self.vision_dataset)
```
NicolasHug, line 64: Would this be covered by `__getattr__()` already?

torchvision/prototype/datapoints/_dataset_wrapper.py
```diff
@@ -0,0 +1,250 @@
...
+
+
+@functools.lru_cache(maxsize=None)
```
NicolasHug, line 67: Is a LRU really needed here? It's not like we're going to create thousands of dataset instances, and the categories lists are fairly small anyway?

torchvision/prototype/datapoints/_dataset_wrapper.py
```diff
@@ -0,0 +1,250 @@
...
+    datasets.FashionMNIST,
+]:
+    _WRAPPERS[dataset_type] = classification_wrapper
+
+
+def segmentation_wrapper(
+    dataset: datasets.VisionDataset, sample: Tuple[PIL.Image.Image, PIL.Image.Image]
+) -> Tuple[PIL.Image.Image, datapoints.Mask]:
+    image, mask = sample
+    return image, datapoints.Mask(F.to_image_tensor(mask))
```
NicolasHug, line 103: Curious why we need `F.to_image_tensor(mask)` before wrapping it into a mask. Does that mean `Image(some_pil_image)` doesn't work either?

torchvision/prototype/datapoints/_dataset_wrapper.py
```diff
@@ -0,0 +1,250 @@
...
+
+
+def caltech101_wrapper(
+    dataset: datasets.Caltech101, sample: Tuple[PIL.Image.Image, Any]
+) -> Tuple[PIL.Image.Image, Any]:
+    image, target = sample
+
+    target_type_wrapper_map: Dict[str, Callable] = {
+        "category": lambda label: datapoints.Label(label, categories=dataset.categories),
+        "annotation": datapoints.GenericDatapoint,
```
NicolasHug, line 119: Is this `annotation` a Tensor already, or is it available as a different type (like a numpy array or a list)? In the spirit of leaving the original types untouched, I wonder if we should just not wrap anything that isn't strictly needed?

NicolasHug (commented):
Thanks Philip, I haven't looked at the individual wrappers in depth, but the overall design looks great (I appreciate the simplifications!). I left some comments / Qs below

torchvision/prototype/datapoints/_dataset_wrapper.py
pmeier: See https://github.com/pytorch/vision/pull/6663#discussion_r982550714. At the time of writing it was not clear if datasets v2 was going to come or not. Since this is clearer now, I'm also happy to put it under transforms. My concern here was that this wrapper links the two API's and we might get cyclic dependencies and so I put it into the place that both `datasets` and `transforms` can safely depend on.

torchvision/prototype/datapoints/_dataset_wrapper.py
pmeier: Fair point. I guess we only need to disable `transforms`.

torchvision/prototype/datapoints/_dataset_wrapper.py
pmeier: The wrapper is called for every sample that is pulled out of the dataset. Since we need to perform some operation rather than jsut attribute access for some datasets, we should probably only do it once. Otherwise, we should at least measure if we take a performance hit here.

torchvision/prototype/datapoints/_dataset_wrapper.py
pmeier: Yes, so far that doesn't work either. So far we pass the input directly into `torch.as_tensor` and PIL images don't work here. IMO, it was preferable to support that directly, or at least have something like `Image.from_pil` for convenience. I proposed something like that in the past (can't find the issue now), but it was soft-rejected since we have this functionality as transform already.

torchvision/prototype/datapoints/_dataset_wrapper.py
pmeier: As discussed offline, `"annotation"` is a numpy array that holds the coordinates of the polygon that represents the outline of the object. Caltech101 is an old datasets, so there only is a single contiguous instance and thus we can represent that with a single array.

We also discussed offline that this is a good use case to avoid the temptation to guess and just fail. This use case is currently not supported by transforms v1 and we aren't using this dataset in our references either. So we can just fail loudly if we encounter this and as the user to create an issue if they need this.

torchvision/prototype/datapoints/_dataset_wrapper.py
pmeier: I've reviewed this again and it was correct to disable everything. At this stage we only need to care about `transforms` since it is guaranteed that it includes all functionality even if the user passed `transform` and `target_transform` individually. However, most datasets don't use `transforms` in `__getitem__`, but rather the individual ones. Since our goal is to get the untransformed sample, we need to disable everything that might change it before we get our hands on it.

I've expanded the comment to explain this better.

torchvision/prototype/datapoints/_dataset_wrapper.py
```diff
@@ -0,0 +1,411 @@
...
+
+for dataset_type in [
+    datasets.Caltech256,
+    datasets.CIFAR10,
+    datasets.CIFAR100,
+    datasets.ImageNet,
+    datasets.MNIST,
+    datasets.FashionMNIST,
+    datasets.GTSRB,
+    datasets.DatasetFolder,
```
pmeier, line 148: For other datasets we opted to fail in case the dataset would return something that our transforms v2 cannot handle or would mishandle. For `DatasetFolder` this could happen as well, since it wraps arbitrary data. I've put it here for completeness and since we can wrap it well with the default classification wrapper. Open to remove it if someone else favors that.

torchvision/prototype/datapoints/_dataset_wrapper.py
```diff
@@ -0,0 +1,411 @@
...
+]:
+    _WRAPPERS[dataset_type] = segmentation_wrapper
+
+
+def video_classification_wrapper(dataset, sample):
+    if dataset.video_clips.output_format == "THWC":
+        raise RuntimeError(
+            f"{type(dataset).__name__} with output_format='THWC' is not supported by this wrapper, "
+            f"since it is not compatible with the transformations. Please use output_format='TCHW' instead."
+        )
```
pmeier, line 170: I've opted to fail here. Another option would be to just perform the permutation here, but that seems disruptive if the user explicitly set `output_format="THWC"` on the dataset. Plus, with that parameter they have very simple remedy if they want the default wrapping. 

Note that the error message is slightly different from the other ones: since our plan is to not support `"THWC"` in the transforms, there is no suggestion to open an issue if someone wants this.

torchvision/prototype/datapoints/_dataset_wrapper.py
```diff
@@ -0,0 +1,411 @@
...
+        dataset,
+        target,
+        {
+            "identity": datapoints.Label,
+            "bbox": lambda item: datapoints.BoundingBox(
+                item, format=datapoints.BoundingBoxFormat.XYWH, spatial_size=(image.height, image.width)
+            ),
+        },
+        # FIXME: Failing on "attr" here is problematic, since it is the default
+        fail_on=["attr", "landmarks"],
```
pmeier, line 324: Without `GenericDatapoint` or a different handling as explored in https://github.com/pytorch/vision/pull/6663#discussion_r1093087483, these two would be treated as images since they are tensors. 

As the comment states, `"attr"` is the default, so we shouldn't fail here if possible. I'm not sure how exactly `CelebA` is used in the wild, but one option could be to wrap `attr` in `datapoints.OneHotLabel`. That would be only somewhat true, since `attr` can have multiple `1`'s depending on which facial features are present. That being said, we had some intense offline discussions about the `OneHotLabel` name before and weren't able to come up with a better one. The new `MixUp` and `CutMix` transforms also violate the "One" part of the name, since they blend labels and so only `sum(one_hot_label) == 1`.

`"landmarks"` are actually keypoints that we should support at some point in the future, given that we have models for that. There is #5326 for that, which had some positive feedback, but was deemed low priority and thus not followed through.

torchvision/prototype/datapoints/_dataset_wrapper.py
```diff
@@ -0,0 +1,411 @@
...
+        for id in data.unique():
+            masks.append(data == id)
+            label = id
+            if label >= 1_000:
+                label //= 1_000
+            labels.append(label)
+        masks = datapoints.Mask(torch.stack(masks))
+        # FIXME: without the labels, returning just the instance masks is pretty useless. However, we would need to
+        #  return a two-tuple or the like where we originally only had a single PIL image.
+        labels = datapoints.Label(torch.stack(labels), categories=[cls.name for cls in dataset.classes])
```
pmeier, line 383: I'm not sure how the instance segmentations are used in the wild, so this would be a prime candidate to fail loudly and ask for user feedback. However, it is the default

https://github.com/pytorch/vision/blob/135a0f9ea9841b6324b4fe8974e2543cbb95709a/torchvision/datasets/cityscapes.py#L109

and thus failing here would make the dataset in its default form unusable with the wrapper.

torchvision/prototype/datapoints/_dataset_wrapper.py
```diff
@@ -0,0 +1,411 @@
...
+    )
+
+
+_WRAPPERS[datasets.Cityscapes] = cityscapes_wrapper
+
+
+def widerface_wrapper(dataset, sample):
+    image, target = sample
+    if target is not None:
+        # FIXME: all returned values inside this dictionary are tensors, but not images
```
pmeier, line 404: Per comment. None of them are images, so we need to handle that.

torchvision/prototype/features/_feature.py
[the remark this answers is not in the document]
pmeier: To clarify, you propose to replace

https://github.com/pytorch/vision/blob/135a0f9ea9841b6324b4fe8974e2543cbb95709a/torchvision/prototype/transforms/_transform.py#L40-L42

with something along the lines:

```python
flat_outputs = []
image_or_video_found = False
for inpt in flat_inputs:
    if check_input(inpt, self._transformed_types):
        if check_input(inpt, datapoints.Image, is_simple_tensor, PIL.Image.Image):
            if image_found:
                output = inpt
            else:
                image_found = True
                output = self._transform(inpt, params)
        else:
            output = self._transform(inpt, params)
    else:
        output = inpt
        
    flat_outputs.append(output)
```

(maybe we can untangle that a little, but the functionality should be clear)

---

Assuming the above is what you meant, I can't say I'm a fan. I agree that it will work for our datasets and most of the current custom ones as well. However, part of the v2 API is the ability to perform the transformation on arbitrary input structures. And this somewhat violates that, since now the order plays a role again. So basically we'll from then on say

> Transforms v2 handles _almost_ arbitrary input structures

Again, I don't think there will be a lot of incompatibility, I just dislike that we need to walk back even slightly on this IMO very nice feature. But then I'm totally biased in that as well.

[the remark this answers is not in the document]
NicolasHug: TBH I don't really see how the "arbitrary input structure" idea is violated.

We just make an assumption on the order, just like we currently make the assumption that Tensor == Image (which, if there's more than one tensor, is the wrong assumption to make in most cases).

> Transforms v2 handles almost arbitrary input structures

I would still say that the proposal handles all input structures. What input structures aren't supported?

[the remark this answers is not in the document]
pmeier: > We just make an assumption on the order, just like we currently make the assumption that Tensor == Image (which, if there's more than one tensor, is the wrong assumption to make in most cases).

I agree, but when the user actually has such a case like `{"image_1": image_tensor_1, "image_2": image_tensor_2}` that won't work. Basically we are limiting the API to a single plain tensor image.

So unless I'm missing something, we need to decide whether we want to 

1. allow multiple tensor images, but users have to wrap every other tensor that is not an image in a `GenericDatapoint`, or we 
2. make the assumption that users can only pass a single plain tensor image and have to manually wrap into `datapoints.Image` if they have multiple ones.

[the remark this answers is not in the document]
NicolasHug: > but when the user actually has such a case like {"image_1": image_tensor_1, "image_2": image_tensor_2} that won't work. Basically we are limiting the API to a single plain tensor image.

Yes but I don't think this is more limiting than treating all tensors as image? For every user who wants to pass multiple tensor images, there's another user who wants to pass a single image tensor and pass-through the rest without having to bother with an extra class.

I really don't think one option is more limiting than the other. I do think however that we should avoid introducing new classes as much as possible (it bloats the users' mental model). And we should remember than any of those use-cases aren't supported by V1 right now anyway. Users who want to transforms more than one image should probably be encouraged to wrap their images as Images rather than pass them as tensors.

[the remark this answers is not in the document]
NicolasHug: We're taking about a use-case that:

- wasn't supported before; so users have to write new code, and they can probably wrap their stuff into Image as they go
- doesn't even lead to anything meaningful within torchvision right now: there's no model that cares about 2 images, apart from the OF ones which aren't supported by V2 rn

So I'm OK with not addressing this potential risk which is IMO extremely niche / low. The risk of exposing an additional class seems higher and would impact more users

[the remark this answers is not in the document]
NicolasHug: > Forgetting to wrap something into a GenericDatapoint will most likely fail loudly due to unmet assumptions that we have for tensor images and thus at least mitigates this problem.

BTW this is purely speculative but if we assume that there's a Resize in 99% of the transformation pipelines, we're almost guaranteed to get a loud error as well if only one of the input images is resized while the rest are passed-through.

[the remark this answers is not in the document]
pmeier: > So I'm OK with not addressing this potential risk which is IMO extremely niche / low. The risk of exposing an additional class seems higher and would impact more users

Fair enough. Last try from my side:

As mentioned in the second part of https://github.com/pytorch/vision/pull/6663#discussion_r982567437, instead of having an extra class that enables this, we could also use the `Datapoint` base class an thus not need any extra class. Given that it is currently not exposed, it would still increase our API surface. But then if we want to go for #7028 at some point, we need to expose it eventually anyway.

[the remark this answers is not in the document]
NicolasHug: > we could also use the Datapoint base class an thus not need any extra class

Trying to clarify that to make sure I understand: if we pass a DataPoint right now, it's already completely passed-through, right? I.e. we would just need to make the DataPoint class public (and document this behaviour)?

Regardless, I'd rather go with the new heuristic option right now, and wait for users feedback and see if some people find it confusing. After all, none of the options discussed so far are really mutually exclusive.

torchvision/prototype/datapoints/_dataset_wrapper.py
NicolasHug: I feel like this is precisely the kind of APIs we need to make as frictionless as possible if we want to optimize adoption / migration. If users have to make 2 calls to convert their PIL images into a DataPoint Image, that's one too many calls (IIUC the above you would agree?).

There might be good technical reasons of *why* this may be hard to enable, but this seems like a very natural and expected UX. Is this something we can revisit?

NicolasHug (commented):
Bunch of comments, nothing too critical

torchvision/prototype/datapoints/_dataset_wrapper.py
pmeier: > There might be good technical reasons of why this may be hard to enable

I don't think that is hard to enable. Looking at the constructor of `Image`, we have
https://github.com/pytorch/vision/blob/135a0f9ea9841b6324b4fe8974e2543cbb95709a/torchvision/prototype/datapoints/_image.py#L18-L26

I don't think it takes more than something like

```python
if isinstance(data, PIL.Image.Image):
    data = F.pil_to_tensor(data)
data = cls._to_tensor(...)
```

> Is this something we can revisit?

As stated above, I wanted to have it in the first place so no concerns from my side. @vfdev-5?

torchvision/prototype/features/_feature.py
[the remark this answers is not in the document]
pmeier: > Trying to clarify that to make sure I understand: if we pass a DataPoint right now, it's already completely passed-through, right? I.e. we would just need to make the DataPoint class public (and document this behaviour)?

Correct. Note that passing a plain `Datapoint` will still go through the regular dispatch mechanism, but since all methods on the base class just return `self`, every dispatcher is a no-op. This is different than for example a string that will be filtered out in 

https://github.com/pytorch/vision/blob/135a0f9ea9841b6324b4fe8974e2543cbb95709a/torchvision/prototype/transforms/_transform.py#L40-L42

and thus never hits a dispatcher.

> Regardless, I'd rather go with the new heuristic option right now, and wait for users feedback and see if some people find it confusing. After all, none of the options discussed so far are really mutually exclusive.

Ok then. Let me send another PR since that behavior is only partially related to this PR.

torchvision/prototype/datapoints/_dataset_wrapper.py
```diff
@@ -0,0 +1,411 @@
...
+
+def list_of_dicts_to_dict_of_lists(list_of_dicts):
+    if not list_of_dicts:
+        return {}
+
+    dict_of_lists = {key: [value] for key, value in list_of_dicts[0].items()}
+    for dct in list_of_dicts[1:]:
+        for key, value in dct.items():
+            dict_of_lists[key].append(value)
+    return dict_of_lists
```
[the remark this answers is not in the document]
pmeier: Had a `defaultdict` before, but for some reason favored that. Can't have been strong since I can't remember. Happy to revert.

torchvision/prototype/datapoints/_dataset_wrapper.py
```diff
@@ -0,0 +1,411 @@
...
+        datasets.CIFAR100: lambda dataset: dataset.classes,
+        datasets.FashionMNIST: lambda dataset: dataset.classes,
+        datasets.ImageNet: lambda dataset: [", ".join(names) for names in dataset.classes],
+        datasets.DatasetFolder: lambda dataset: dataset.classes,
+        datasets.ImageFolder: lambda dataset: dataset.classes,
+    }.get(type(dataset))
+    return categories_fn(dataset) if categories_fn is not None else None
+
+
+def classification_wrapper(dataset, sample):
```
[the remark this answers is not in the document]
pmeier: You are right that we only need to wrap the sample. However the canonical way to use transforms is by setting `dataset = datasets.ImageNet(..., transform=...)`. Since we need to wrap the raw data before it hits the transforms, we only have two options:

1. Wrap the dataset and monkeypatch `__getitem__` to inject our wrapping before we hit the transforms. 
2. Tell users to not use `transform`, `target_transform`, and `transforms` as before, and use them manually.

IMO, 2. is more disruptive to the workflow and thus I opted for 1. For completeness, if we are going for 2. after all, the whole chicken-and-egg problem goes away, since we then can instantiate the dataset before we instantiate the transform. Meaning, we could do something like

```python
dataset = datasets.ImageNet(...)

transform = transforms.Compose([
    V2WrapperTransform.from_vision_dataset(dataset),
    ...,
])

# showing the regular case of only using `transform`
# For other cases, the loop will have to be adjusted
for input, target in dataset:
    input = transform(input)
    ...
```

torchvision/prototype/datapoints/_dataset_wrapper.py
```diff
@@ -0,0 +1,411 @@
...
+        for key, value in dct.items():
+            dict_of_lists[key].append(value)
+    return dict_of_lists
+
+
+def wrap_target_by_type(dataset, target, type_wrappers, *, fail_on=(), attr_name="target_type"):
+    if target is None:
+        return None
+
+    target_types = getattr(dataset, attr_name)
```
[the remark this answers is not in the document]
pmeier: Not sure that simplifies stuff a lot, but I'm not going to oppose.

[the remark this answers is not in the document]
NicolasHug: less parameters, less assumptions to make when reading the code, less scary function (etc.)

torchvision/prototype/datapoints/_dataset_wrapper.py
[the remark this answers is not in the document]
NicolasHug: I'm not sure I follow. If we didn't care about the `categories` meta-data, we wouldn't need to pass the dataset, correct?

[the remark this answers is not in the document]
pmeier: Nope. The wrapper needs to be adjusted depending on the dataset parameter. Take the `target_type` for example. Without knowing its exact value, we cannot provide a wrapper, since that information is missing from the sample. So `V2WrapperTransform.from_vision_dataset(...)` can either take the instantiated dataset which has this information, or we need to basically replicate the signature of the class, e.g.

```python
dataset = MyDataset(root, foo=2, bar="bar", baz=[-1.0, 2.0])
transform = V2WrapperTransform.from_dataset_args(MyDataset, foo=2, bar="bar", baz=[-1.0, 2.0])
```

Of course there are datasets for which the wrapper can be instantiated by just knowing the class, e.g. the classification datasets. But this doesn't hold in general.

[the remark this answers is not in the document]
NicolasHug: OK, I guess what I'm trying to say is: we only need to know the dataset instance *once*, not every single time `__getitem__()` gets called? I.e. we patch `__getitem__()` once and for all when `wrap_dataset_for_transforms_v2` is called and then the inner wrapper in `__getitem__()` should only need to take samples as input?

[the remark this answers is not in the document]
pmeier: Correct. I've opted to have a new object that provides the custom `__getitem__` over monkey-patching the original object, but I don't see a downside to the latter either.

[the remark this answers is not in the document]
NicolasHug: We don't have to monkey-patch the original dataset we just have to create the inner wrapper in `__init__()` of `_VisionDatasetDatapointWrapper`. It sounds unefficient to do all the checks in  `wrap_target_by_type()` when they just need to be done once at instanciation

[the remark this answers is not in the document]
pmeier: Fair enough. Let me find a better way to implement this.

test/test_datasets.py
```diff
@@ -763,11 +763,19 @@ def _create_annotation_file(self, root, name, file_names, num_annotations_per_im
...
+        annotations = []
+        annotion_id = 0
+        for image_id in itertools.islice(itertools.cycle(image_ids), len(image_ids) * num_annotations_per_image):
+            annotations.append(
+                dict(
+                    image_id=image_id,
+                    id=annotion_id,
+                    bbox=torch.rand(4).tolist(),
+                    segmentation=[torch.rand(8).tolist()],
+                    category_id=int(torch.randint(91, ())),
```
pmeier, line 775: These keys were not present before, because they were irrelevant. However, the wrapper tries to access them so I added some dummy data.

pmeier (commented):
The latest commits add the following changes:

- Following #7170, we no longer need special handling for any non-image parts of the samples, since all our datasets fit the heuristic.
- Following https://github.com/pytorch/vision/pull/7171#issuecomment-1419241405 we no longer use the `datapoints.Label` class. Together with the pass-through for PIL images, this means that this wrapper is a no-op for classification datasets. This also resolves all discussions regarding the categories, since they are no longer present.
- Instead of having a wrapper that takes the dataset as well as the sample as inputs, i.e.
  
  ```python
  def wrapper(dataset, sample):
      ...
      return wrapped_sample
  ```

  the architecture was changed to a factory pattern:

  ```python
  def wrapper_factory(dataset):
      def wrapper(sample):
          ...
          return wrapped_sample
  
      return wrapper
  ```

  In addition to making the wrapper more "natural", this also enables us to raise on unsupported behavior on wrapping rather than when the first sample is drawn. That should improve UX.
- I've added automated smoke tests to our datasets v1 tests to make sure wrapping and drawing samples doesn't raise anything.

torchvision/prototype/datapoints/_dataset_wrapper.py
```diff
@@ -0,0 +1,408 @@
...
+class WrapperFactories(dict):
+    def register(self, dataset_cls):
+        def decorator(wrapper_factory):
+            self[dataset_cls] = wrapper_factory
+            return wrapper_factory
+
+        return decorator
+
+
+WRAPPER_FACTORIES = WrapperFactories()
```
NicolasHug, line 32: perhaps add a comment for those like myself who will forget and wonder why it's done this way:

> We need factories of wrappers and not just wrappers, because some of the wrappers are **instance-dependent** (as they use instance attributes as defined by users). The factories thus need to generate the proper wrapper at execution time.

torchvision/prototype/datapoints/_dataset_wrapper.py
```diff
@@ -0,0 +1,408 @@
...
+
+def list_of_dicts_to_dict_of_lists(list_of_dicts):
+    dict_of_lists = defaultdict(list)
+    for dct in list_of_dicts:
+        for key, value in dct.items():
+            dict_of_lists[key].append(value)
+    return dict(dict_of_lists)
+
+
+def wrap_target_by_type(target, *, target_types, type_wrappers=None):
```
NicolasHug, line 109: slightly confused here: if `type_wrappers` is None, isn't this a no-op?

NicolasHug (approved):
Thanks a lot Philip, LGTM. I just have minor comments (that we discussed offline and are probably already addressed) + one minor Q about `wrap_target_by_type`.

I'll admit I haven't taken a super deep look to the individual wrappers. We may have to do a bit of manual testing for those later.

test/datasets_utils.py
```diff
@@ -581,6 +581,28 @@ def test_transforms(self, config):
...
+        try:
+            from torchvision.prototype.datapoints import wrap_dataset_for_transforms_v2
+        except ImportError:
+            return
+
+        try:
+            with self.create_dataset(config) as (dataset, _):
+                wrapped_dataset = wrap_dataset_for_transforms_v2(dataset)
+                wrapped_sample = wrapped_dataset[0]
+                assert wrapped_sample is not None
```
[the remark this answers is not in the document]
pmeier: This currently fails for `DatasetFolder`, since it doesn't load anything:

https://github.com/pytorch/vision/blob/87ec8048085a5d892e78e60cd1fb0b4e37219c1e/test/test_datasets.py#L1551

But, I'll have to tackle that test anyway for another reason: https://github.com/pytorch/vision/pull/7155#issuecomment-1423267645. I think I can manage something that is sufficient for both cases.

torchvision/prototype/features/__init__.py
pmeier: We have decided to bikeshed over this later. This is tracked in #7092.

test/datasets_utils.py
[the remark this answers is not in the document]
pmeier: Fixed in #7216.

PR #6665, opened 2022-09-29
[proto] Enable GPU tests on prototype

vfdev-5:


.github/workflows/prototype-tests-gpu.yml
```diff
@@ -0,0 +1,79 @@
...
+
+on:
+  pull_request:
+
+jobs:
+  prototype:
+    strategy:
+      fail-fast: false
+
+    runs-on: [self-hosted, linux.4xlarge.nvidia.gpu]
```
huydhn, line 12: As you are using the shared `linux.4xlarge.nvidia.gpu`, I think you could try the same approach as  PyTorch CI to make this works.  

* Install NVIDIA driver (nvidia-smi) and docker nvidia support https://github.com/pytorch/pytorch/blob/master/.github/workflows/_linux-test.yml#L63-L85
* Run the test in docker https://github.com/pytorch/pytorch/blob/master/.github/workflows/_linux-test.yml#L156, but you would need to provide the GPU_FLAG `--gpus all` for docker to correct pick up NVIDIA driver (nvidia-smi)

I have a feeling that `nvidia-smi` doesn't show up because the GPU_FLAG is missing https://docs.github.com/en/actions/using-jobs/running-jobs-in-a-container.  So you could also try to see if the following works.  You would still need to install NVIDIA driver first though because the AMI for `linux.4xlarge.nvidia.gpu` doesn't have it (I remember we track this issue somewhere)

```
container:
  image: pytorch/conda-builder:cuda116
  options: --gpus all
```

osalpekar:
Also there might be an opportunity to simplify this using the generic Linux jobs (with GPU support) that @seemethere has built. You can find the documentation here: https://github.com/pytorch/test-infra/wiki/Writing-generic-linux-jobs

huydhn:
> Also there might be an opportunity to simplify this using the generic Linux jobs (with GPU support) that @seemethere has built. You can find the documentation here: https://github.com/pytorch/test-infra/wiki/Writing-generic-linux-jobs

Ohh, this is nice and a much better option. TIL

vfdev-5:
@huydhn @osalpekar thanks for the review, I'll try to add `--gpus=all` flag to see if this enables the CI. Using generic linux jobs could be potentially fine, the only limitation I see from the example is that we should now create a single `script` vs multiple steps...

vfdev-5:
It is expected to have failures in cuda vs cpu tests, I'll xfail them in the PR that enables cuda tests and we'll fix the inconsistency in a follow-up PR

pmeier:
Unless I have missed something, there are all just closeness related. Thus, we only have to adjust the tolerances in our test suite. Or to put it differently: there is likely no bug in our implementation. I'm ok with doing that in a follow-up.

test/test_prototype_transforms_functional.py
```diff
@@ -171,7 +171,10 @@ def test_cuda_vs_cpu(self, info, args_kwargs):
         output_cpu = info.kernel(input_cpu, *other_args, **kwargs)
         output_cuda = info.kernel(input_cuda, *other_args, **kwargs)
 
-        assert_close(output_cuda, output_cpu, check_device=False, **info.closeness_kwargs)
+        try:
+            assert_close(output_cuda, output_cpu, check_device=False, **info.closeness_kwargs)
+        except AssertionError:
+            pytest.xfail("CUDA vs CPU tolerance issue to be fixed")
```
pmeier, line 177: This effectively disables this test. Either we should add proper xfails to the `KernelInfo`'s or simply comment out this test with a `FXIME` note. Otherwise we are wasting resources.

.github/workflows/prototype-tests-gpu.yml
```diff
@@ -0,0 +1,80 @@
...
+
+      - name: Install other prototype dependencies
+        run: pip install --progress-bar=off scipy pycocotools h5py iopath
+
+      - name: Install test requirements
+        run: pip install --progress-bar=off pytest pytest-mock pytest-cov
+
+      - name: Mark setup as complete
+        id: setup
+        run: python -c "import torch; print(torch.__version__); print(torch.cuda.is_available())" && exit 0
```
pmeier, line 41: Is this needed? We are explicitly installing from the CU116 wheels above. What extra information are you trying to get here?

pmeier (commented):
Given that there are very few differences between the GPU workflow and the CPU one, can we maybe merge the files?

.github/workflows/prototype-tests-gpu.yml
vfdev-5: Well, this just some debug info. If I do not pass `--gpus=all` flag and skip `nvidia-smi` step, then everything will be installed and tests will run without cuda tests and we wont see that...

vfdev-5:
> Given that there are very few differences between the GPU workflow and the CPU one, can we maybe merge the files?

IMO, this is complicated to refactor configuration for self-hosted and GHA runners. Can be done by someone else with better GHA knowledge.

test/test_prototype_transforms_functional.py
vfdev-5: This is a temporary fix with 3 lines. What you suggest if I understand correctly is to mark specific tests which can vary on GPU etc. Taking into account that you wanted to fix the problem we can keep things like that.

.github/workflows/prototype-tests-gpu.yml
pmeier: Shouldn't we assert it then? Something like

```suggestion
        run: python -c "import sys, torch; sys.exit(not torch.cuda.is_available())
```

test/test_prototype_transforms_functional.py
pmeier: I agree, fixing the individual tests is overkill here. But as is, this test is running with no information gain. `assert_close` will either pass or raise an `AssertionError`. Since we catch that and turn it into an `xfail`, there is no way this test can fail at all. Thus, we are better off just disabling the test completely, e.g. by commenting it out as I suggested, to get the same information but without wasted CI resources.

pmeier:
> IMO, this is complicated to refactor configuration for self-hosted and GHA runners. Can be done by someone else with better GHA knowledge.

Agreed. I can take that up in a follow-up.

.github/workflows/prototype-tests-gpu.yml
vfdev-5: yes, if we need to assert that we can do
```bash
python -c "import torch; exit(not torch.cuda.is_available())"
```

test/test_prototype_transforms_functional.py
vfdev-5: I see your point but I think it is ok to keep like as here as it still shows that majority of ops are passing on cuda.
As for wasted resources, run cuda_vs_cpu tests takes around 7 seconds.

pmeier (approved):
Stamping to unblock.

vfdev-5:
@osalpekar I merged this PR to enable GPU tests for prototype module and now on another PR I have an issue with starting the container with `--gpus=all` option:
- https://github.com/pytorch/vision/actions/runs/3297067787/attempts/1
- https://github.com/pytorch/vision/actions/runs/3297067787/attempts/2

Do you have any ideas why this could happen ? Thanks

PR #6622, opened 2022-09-21
cleanup prototype transforms functional tests

pmeier:
This PR cleans up the recent changes to our prototype transforms functional test frameworks. It merges the files `test_prototype_transforms_kernels.py` and `test_prototype_transforms_dispatchers.py` into `test_prototype_transforms_functional.py` while removing any obsolete code in the latter.

From here on there are the following tasks that are left to do:

1. Add `KernelInfo`'s for the remaining kernels

    https://github.com/pytorch/vision/blob/56e707bfccb62ada836d21e431d6db0d10dd73a1/test/test_prototype_transforms_kernels.py#L33-L41

    Afterwards, this can also be used to add the `DispatcherInfo`'s.
2. Add dispatch tests for the dispatchers. Basically we need to make sure that
    1. if `PIL.Image.Image` gets passed, we dispatch to the PIL kernel
    2. if a simple tensor is passed, we dispatch to the image tensor kernel
    3. if a `_Feature` is passed we dispatch to the feature method
3. Reduce the number of sample inputs since they are used for multiple tests:

    <details><summary>Overview</summary>
    <p>

    | kernel                           |   # sample inputs |
    |----------------------------------|-------------------|
    | pad_image_tensor                 |              2448 |
    | pad_mask                         |               864 |
    | affine_image_tensor              |               480 |
    | rotate_image_tensor              |               432 |
    | center_crop_image_tensor         |               432 |
    | resize_image_tensor              |               384 |
    | crop_image_tensor                |               384 |
    | center_crop_mask                 |               324 |
    | resize_mask                      |               288 |
    | convert_color_space_image_tensor |               288 |
    | crop_mask                        |               288 |
    | resized_crop_image_tensor        |               192 |
    | convert_format_bounding_box      |               144 |
    | resized_crop_mask                |               144 |
    | pad_bounding_box                 |               108 |
    | center_crop_bounding_box         |               108 |
    | perspective_image_tensor         |                80 |
    | elastic_image_tensor             |                80 |
    | affine_mask                      |                72 |
    | rotate_mask                      |                72 |
    | crop_bounding_box                |                72 |
    | erase_image_tensor               |                48 |
    | horizontal_flip_mask             |                36 |
    | affine_bounding_box              |                36 |
    | vertical_flip_mask               |                36 |
    | rotate_bounding_box              |                36 |
    | resized_crop_bounding_box        |                36 |
    | perspective_bounding_box         |                36 |
    | elastic_bounding_box             |                36 |
[...]

Issue #4029, opened 2021-06-09
[feature request] rgb2lab and other color space conversions


vadimkantorov:
It would be nice to have them available in the standard library (both on CPU/GPU) rather than rolling your own. Kornia seems to have implemented it: https://kornia.readthedocs.io/en/latest/color.html, but I think it's basic enough to merit inclusion of similar functions in core.

(my usecase: SLIC superpixel extraction which requires rgb2lab)

datumbox:
Thanks @vadimkantorov. This is useful feedback as we are currently in discussions to improve our transforms.

vadimkantorov:
Maybe this is not even related to transforms per se, since useful also in non-augmentation contexts as well.

For transforms, I really hope, there is more native batched transforms, clear separation between function and random-sampling its parameters, and also that there is an easy way to do inverse transforms (where possible). Also getting padding-masks for direct and inverse transforms is useful

datumbox:
@vadimkantorov The tag transforms includes to anything under the transform module, including augmentations, functional and class transforms. Similar requests have been raised in the past (see #3224 for summary), so that's something we are looking into.

fmassa:
@vadimkantorov I think adding a `rgb2lab` would be nice in principle, but the current issue I see with it is that there is no way for the library to know if the input tensor that you are passing is indeed in `rgb` colorspace. So you can silently get wrong results if you are not careful.
As of now, torchvision assumes that the tensor is either in RGB or grayscale space, and don't try to make any further assumptions otherwise. This works fine because RGB and grayscale have different number of channels, but things break if we start supporting HSV / LAB / etc colorspaces.

Wondering if you think this is a problem per se, and that this is up to the user to make sure that the input tensors are in RGB colorspace?

vadimkantorov:
I think it's not a problem per se, as we almost do not encode semantics into the tensor type system in PyTorch, in the goal of being flexible and letting users quickly build on things. Same for NumPy.

This problem persists with OpenCV returning BGR images, but it's up to users to remember to convert if this is needed.

In my own practice, I use suffix notation to bring extra attention to semantics, e.g. I use names like `mean_rgb`, `std_rgb` in my input normalization functions, especially if I do image reading with OpenCV.

If these functions stay in extra namespace and are not discussed in every other tutorial, the use will be limited to those looking for them, so it should not confuse the users of the most common usecases.

oke-aditya:
[...]

PR #6667, opened 2022-09-29
add Video feature and kernels

pmeier:


torchvision/prototype/features/_video.py
```diff
@@ -0,0 +1,232 @@
...
+    @property
+    def image_size(self) -> Tuple[int, int]:
+        return cast(Tuple[int, int], tuple(self.shape[-2:]))
+
+    @property
+    def num_channels(self) -> int:
+        return self.shape[-3]
+
+    @property
+    def num_frames(self) -> int:
```
pmeier, line 64: This is new for videos. Is that the correct name? cc @bjuncek @jdsgomes

torchvision/prototype/transforms/functional/_augment.py
```diff
@@ -17,15 +17,21 @@ def erase_image_pil(
     return to_pil_image(output, mode=image.mode)
 
 
+def erase_video(
```
pmeier, line 20: The new video kernels are trivial for now, since they only wrap the image kernels. This works as long as the image kernel supports arbitrary batch sizes, since for all the current transformations videos can be regarded as a batch of images. 

However, there are some kernels that currently don't support arbitrary batch sizes:

https://github.com/pytorch/vision/blob/9cece405db3a8301463b219d9e0885142837ae4f/test/prototype_transforms_kernel_infos.py#L906-L907

We need to fix them in order for the video kernel to work properly.

In the future, if we have transforms that work on the temporal dimension, i.e. `video.shape[-4]`, we actually need to implement explicit video kernels.

test/prototype_transforms_kernel_infos.py
```diff
@@ -126,6 +132,11 @@ def sample_inputs_horizontal_flip_mask():
         yield ArgsKwargs(image_loader)
 
 
+def sample_inputs_horizontal_flip_video():
```
pmeier, line 135: I intentionally only added a single `KernelInfo` for videos here, since we already have quite a few tests and videos will add even more. Thus, I'll go for 3. of #6622 first before adding more tests here. As explained above, the video kernels are a trivial wrapper around the image kernels for now, so unless I made a mistake while implementing the signature, there should be no new bugs.

torchvision/prototype/features/_video.py
```diff
@@ -0,0 +1,232 @@
...
+        color_space: Optional[Union[ColorSpace, str]] = None,
+        dtype: Optional[torch.dtype] = None,
+        device: Optional[Union[torch.device, str, int]] = None,
+        requires_grad: bool = False,
+    ) -> Video:
+        data = torch.as_tensor(data, dtype=dtype, device=device)  # type: ignore[arg-type]
+        if data.ndim < 3:
+            raise ValueError
+        elif data.ndim == 3:
+            data = data.unsqueeze(0)
```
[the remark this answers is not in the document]
datumbox: I think for videos we can safely require exactly 4 dimensions.

torchvision/prototype/features/_video.py
```diff
@@ -0,0 +1,232 @@
...
+    @classmethod
+    def new_like(
+        cls, other: Video, data: Any, *, color_space: Optional[Union[ColorSpace, str]] = None, **kwargs: Any
+    ) -> Video:
+        return super().new_like(
+            other, data, color_space=color_space if color_space is not None else other.color_space, **kwargs
+        )
+
+    @property
+    def image_size(self) -> Tuple[int, int]:
```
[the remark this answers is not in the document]
datumbox: Could you add a TODO?

torchvision/prototype/features/_video.py
```diff
@@ -0,0 +1,232 @@
...
+        return Video.new_like(
+            self,
+            self._F.convert_color_space_video(
+                self, old_color_space=self.color_space, new_color_space=color_space, copy=copy
+            ),
+            color_space=color_space,
+        )
+
+    def horizontal_flip(self) -> Video:
+        output = self._F.horizontal_flip_Video_tensor(self)
```
datumbox, line 80: Why is Video capitalized?

torchvision/prototype/transforms/functional/_augment.py
datumbox: For those that dont support arbitrary dimensions, you are able to reshape the T dimension into N, send to the Image kernel and then reshape back as a temporarily workaround.

torchvision/prototype/transforms/functional/_augment.py
```diff
@@ -17,15 +17,21 @@ def erase_image_pil(
...
+    video: torch.Tensor, i: int, j: int, h: int, w: int, v: torch.Tensor, inplace: bool = False
+) -> torch.Tensor:
+    return erase_image_tensor(video, i=i, j=j, h=h, w=w, v=v, inplace=inplace)
+
+
 def erase(
-    inpt: features.ImageTypeJIT,
+    inpt: features.InputTypeJIT,
```
datumbox, line 27: The proper type for of  `InputType` is defined as `Union[torch.Tensor, PIL.Image.Image, _Feature]`. Note that we include the base class for `_Feature` here not `Image`. So here we imply that BBoxes can be transformed (though I understand you use the JIT-friendly version but things go hand in hand).

One potential approach here is to add `Video` in the `ImageType`  and rename accordingly.

torchvision/prototype/transforms/functional/_color.py
```diff
@@ -19,6 +23,10 @@ def adjust_brightness(inpt: features.InputTypeJIT, brightness_factor: float) ->
 adjust_saturation_image_pil = _FP.adjust_saturation
 
 
+def adjust_saturation_video(video: torch.Tensor, saturation_factor: float) -> torch.Tensor:
```
datumbox, line 26: Following the pattern with other kernels, shouldn't this be something like:
```suggestion
def adjust_saturation_video(video: features.VideoTypeJIT, saturation_factor: float) -> features.VideoTypeJIT:
```

torchvision/prototype/transforms/functional/_misc.py
```diff
@@ -9,15 +9,19 @@
 normalize_image_tensor = _FT.normalize
 
 
+def normalize_video(video: torch.Tensor, mean: List[float], std: List[float], inplace: bool = False) -> torch.Tensor:
+    return normalize_image_tensor(video, mean, std, inplace=inplace)
+
+
 def normalize(
     inpt: features.TensorImageTypeJIT, mean: List[float], std: List[float], inplace: bool = False
```
datumbox, line 17: All these types such as `features.TensorImageTypeJIT` need to be named to make sense.

torchvision/prototype/transforms/functional/_misc.py
```diff
@@ -9,15 +9,19 @@
...
 ) -> torch.Tensor:
     if not isinstance(inpt, torch.Tensor):
         raise TypeError(f"img should be Tensor Image. Got {type(inpt)}")
-    else:
-        # Image instance after normalization is not Image anymore due to unknown data range
-        # Thus we return Tensor for input Image
-        return normalize_image_tensor(inpt, mean=mean, std=std, inplace=inplace)
+
+    # Image or video instance after normalization is not Image or Video anymore due to unknown data range. Thus, we
```
datumbox, line 22: There are so many place that we need to phase out the word image. I've been complaining about this for a very long time 😄  Time to change them.

datumbox (commented):
I saw you put comments for us, so from that I assume you want us to have a look. If not the case, feel free to ignore.

torchvision/prototype/features/_video.py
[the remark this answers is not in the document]
pmeier: As in a video will always have shape `(F, C, H, W)` rather than arbitrary leading batch dimensions, i.e. `(*, F, C, H, W)`? If yes, what is the reason for this restriction given that all other features don't have this?

Or do you mean we can require the input of the constructor to always have *at least* 4 dimensions?

torchvision/prototype/transforms/functional/_augment.py
pmeier: I've added #6670. How urgent is it for this to work for videos but not for images? If not as urgent, I would fix the image kernels and benefit from that here.

torchvision/prototype/transforms/functional/_augment.py
pmeier: What do you think would be a good name? `ImageOrVideoType`?

torchvision/prototype/features/_video.py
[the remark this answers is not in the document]
datumbox: I was suggesting the latter. Require the input of the constructor to always have at least 4 dimensions.

torchvision/prototype/transforms/functional/_augment.py
datumbox: Not urgent, we can do it on a follow up.

torchvision/prototype/transforms/functional/_augment.py
datumbox: I can live with that for now. We will rethink names before releasing.

torchvision/prototype/features/_video.py
pmeier: Copy-paste mistake. Good catch!

test/prototype_common_utils.py
```diff
@@ -527,3 +529,42 @@ def make_mask_loaders(
...
+    return VideoLoader(fn, shape=(*extra_dims, num_frames, *size), dtype=dtype, color_space=color_space)
+
+
+def make_video_loaders(
+    *,
+    sizes=DEFAULT_IMAGE_SIZES,
+    color_spaces=(
+        features.ColorSpace.GRAY,
+        features.ColorSpace.RGB,
+    ),
```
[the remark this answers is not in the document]
jdsgomes: That's correct. We might support YUV in the future as this is something supported by ffmpeg but I don't think we have any use case for the time being

test/prototype_common_utils.py
```diff
@@ -527,3 +529,42 @@ def make_mask_loaders(
...
+    num_frames=(1, 0, "random"),
+    extra_dims=DEFAULT_EXTRA_DIMS,
+    dtypes=(torch.float32, torch.uint8),
```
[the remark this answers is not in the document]
jdsgomes: For now we only support uint8, what are the implications of having float32 in this tuple?

test/prototype_common_utils.py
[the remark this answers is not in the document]
pmeier: This also ties into #4029.

torchvision/prototype/features/_video.py
jdsgomes: Yes that is correct, its the same name used in video clips

test/prototype_common_utils.py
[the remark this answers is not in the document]
pmeier: It means all our transforms test that don't explicitly specify the `dtype` will test with videos in both `dtypes`. If we are only supporting `torch.uint8` now, we should remove this here to avoid unnecessary tests that might or might not pass.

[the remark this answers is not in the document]
pmeier:
```suggestion
    dtypes=(torch.uint8,),
```

torchvision/prototype/features/_video.py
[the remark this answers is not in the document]
YosuaMichael: @pmeier just curious what `F` stand for in `(F, C, H, W)` ? From what I know, usually the shape is either channel last, `(T, H, W, C)`, or channel first, `(C, T, H, W)`. Where `T` is `temporal` dimension.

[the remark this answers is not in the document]
datumbox: `F` here is the temporal dimension. I recommend using `T` in our comms to avoid misunderstandings.

test/prototype_common_utils.py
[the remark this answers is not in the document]
bjuncek: I do not think supporting YUV is quite necessary. YUV should only be used as an intermediate colour space whilst decoding

bjuncek (commented):
One high level comment:

At the moment, if i understand it correctly, `Video` feature class only deals with seq-of-images data. 
How hard (or easy) would it be to extend this to additional video modalities (say audio)?

torchvision/prototype/features/_video.py
```diff
@@ -0,0 +1,232 @@
...
+        return Video.new_like(self, output)
+
+    def resize(  # type: ignore[override]
+        self,
+        size: List[int],
+        interpolation: InterpolationMode = InterpolationMode.BILINEAR,
+        max_size: Optional[int] = None,
+        antialias: bool = False,
+    ) -> Video:
+        output = self._F.resize_Video_tensor(
```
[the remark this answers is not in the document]
pmeier: See https://github.com/pytorch/vision/pull/6667#discussion_r984360727.

pmeier:
> At the moment, if i understand it correctly, `Video` feature class only deals with seq-of-images data.
> How hard (or easy) would it be to extend this to additional video modalities (say audio)?

As is, `Video` is a tensor subclass and thus does not support more than one storage. Thus, I don't think it is possible to bundle video and audio into one tensor. There are ["nested" tensors](https://pytorch.org/docs/stable/nested.html), but the are in prototype state for quite some time now. Plus, the tensors ought to have the same dimensionality (not shape), but that will be quite awkward for audio. I'm guessing the actual shape should be something like `(*, K, L)` like we have in our video datasets now:

https://github.com/pytorch/vision/blob/07ae61bf9c21ddd1d5f65d326aa9636849b383ca/torchvision/datasets/kinetics.py#L72-L73

(Noob question: we don't see `T` here, because the audio is sampled differently, read a lot higher, than the video, right?)

Could you be more specific what kind of transformations on video with audio you have in mind?

torchvision/prototype/transforms/functional/_color.py
pmeier: No. The kernels take plain tensors and return them. As such annotating with `torch.Tensor` is correct as this is not a workaround for JIT, but what we actually want.

torchvision/prototype/transforms/_augment.py
```diff
@@ -92,7 +92,7 @@ def _get_params(self, sample: Any) -> Dict[str, Any]:
 
         return dict(i=i, j=j, h=h, w=w, v=v)
 
-    def _transform(self, inpt: features.ImageType, params: Dict[str, Any]) -> features.ImageType:
+    def _transform(self, inpt: features.ImageOrVideoType, params: Dict[str, Any]) -> features.ImageType:
```
datumbox, line 95:
```suggestion
    def _transform(self, inpt: features.ImageOrVideoType, params: Dict[str, Any]) -> features.ImageOrVideoType:
```

torchvision/prototype/transforms/_color.py
```diff
@@ -110,7 +110,7 @@ def _get_params(self, sample: Any) -> Dict[str, Any]:
             channel_permutation=torch.randperm(num_channels) if torch.rand(()) < self.p else None,
         )
 
-    def _permute_channels(self, inpt: features.ImageType, permutation: torch.Tensor) -> features.ImageType:
+    def _permute_channels(self, inpt: features.ImageOrVideoType, permutation: torch.Tensor) -> features.ImageType:
```
datumbox, line 113: We need this everywhere I think:
```suggestion
    def _permute_channels(self, inpt: features.ImageOrVideoType, permutation: torch.Tensor) -> features.ImageOrVideoType:
```

torchvision/prototype/transforms/functional/_augment.py
datumbox: @pmeier You missed this comment. This should be `ImageOrVideoType` I think. Same on the return type.

datumbox (commented):
Overall looks good. A few comments below. We also need to adapt the AutoAugment classes.

One concern I have is if there is a specific transform that doesn't support Videos (squashes dimensions for example) or does something unexpected. I'm happy to go overall to the API and do checks for things we need to do or missed. Another thing we need after this PR as clean-up is remove the `image` wording (both from internal method variables and from arguments like `image_size`).

bjuncek:
>As is, Video is a tensor subclass and thus does not support more than one storage. Thus, I don't think it is possible to bundle video and audio into one tensor.

Ah, ok, makes sense. 

> Noob question: we don't see T here, because the audio is sampled differently, read a lot higher, than the video, right?

Yeah. Basically, audio should be `num_channels` (`K`) x `1-D signal`
On average signal is sampled at 48kHZ, so for a second of stereo audio, you'd have 
`2*48000` size tensor. 

>  what kind of transformations on video with audio you have in mind?

Usually the transforms are handled separately (when I was last working in multimodal video field, we'd apply things separately). So for example in the dataset `__getitem__` we'd have:

```
...
video, audio, info = read_video(path)
if v_transform is not None:
    video = v_transform(video)
if a_transform is not None:
    audio = a_transform(audio)
...
```


This all makes sense though. Let's worry about audio (and if we'd even need it) down the line.

torchvision/prototype/features/_video.py
```diff
@@ -0,0 +1,240 @@
...
+VideoTypeJIT = torch.Tensor
+LegacyVideoType = torch.Tensor
+LegacyVideoTypeJIT = torch.Tensor
+TensorVideoType = Union[torch.Tensor, Video]
+TensorVideoTypeJIT = torch.Tensor
+
+ImageOrVideoType = Union[ImageType, VideoType]
+ImageOrVideoTypeJIT = Union[ImageTypeJIT, VideoTypeJIT]
+TensorImageOrVideoType = Union[TensorImageType, TensorVideoType]
+TensorImageOrVideoTypeJIT = Union[TensorImageTypeJIT, TensorVideoTypeJIT]
```
datumbox, line 240: Nit: This name kinda sucks and I don't have any better naming proposal. What if instead of having a single type that unions both, we just use union on the actual method? I suspect that this might be as informative but less weird. This is just an opinion though and a matter of preference, thoughts @pmeier and @vfdev-5?

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -31,26 +31,27 @@ def _get_random_item(self, dct: Dict[K, V]) -> Tuple[K, V]:
...
-    def _extract_image(
+    def _extract_image_or_video(
         self,
         sample: Any,
         unsupported_types: Tuple[Type, ...] = (features.BoundingBox, features.Mask),
-    ) -> Tuple[int, features.ImageType]:
+    ) -> Tuple[int, features.ImageOrVideoType]:
         sample_flat, _ = tree_flatten(sample)
-        images = []
+        image_likes = []
```
datumbox, line 40: Nit?
```suggestion
        inputs = []
```

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -471,13 +472,13 @@ def _sample_dirichlet(self, params: torch.Tensor) -> torch.Tensor:
...
+        id, orig_image_or_video = self._extract_image_or_video(sample)
+        _, height, width = get_chw(orig_image_or_video)
 
-        if isinstance(orig_image, torch.Tensor):
-            image = orig_image
+        if isinstance(orig_image_or_video, torch.Tensor):
+            image = orig_image_or_video
         else:  # isinstance(inpt, PIL.Image.Image):
-            image = F.pil_to_tensor(orig_image)
+            image = F.pil_to_tensor(orig_image_or_video)
```
datumbox, line 481: Nit?
```suggestion
            image_or_video = F.pil_to_tensor(orig_image_or_video)
```

torchvision/prototype/transforms/functional/_augment.py
datumbox: @pmeier You keep resolving this comment but it's still unresolved. This type is wrong, because BBoxes are not suppored. It needs to be `ImageOrVideoType`

torchvision/prototype/transforms/functional/_color.py
```diff
@@ -32,6 +40,10 @@ def adjust_saturation(inpt: features.InputTypeJIT, saturation_factor: float) ->
 adjust_contrast_image_pil = _FP.adjust_contrast
 
 
+def adjust_contrast_video(video: torch.Tensor, contrast_factor: float) -> torch.Tensor:
+    return adjust_contrast_image_tensor(video, contrast_factor=contrast_factor)
```
datumbox, line 44: Question: Should we alias it instead? On the pros side, if someone asks for the name of the method, they will get `adjust_contrast_video` which is what we want. On the cons side, we just have an extra unnecessary method call. No strong opinion, just highlighting it.

torchvision/prototype/transforms/functional/_meta.py
```diff
@@ -207,13 +207,23 @@ def convert_color_space_image_pil(
...
-    inpt: features.ImageTypeJIT,
+    inpt: features.ImageOrVideoTypeJIT,
     color_space: ColorSpace,
     old_color_space: Optional[ColorSpace] = None,
     copy: bool = True,
-) -> features.ImageTypeJIT:
-    if isinstance(inpt, torch.Tensor) and (torch.jit.is_scripting() or not isinstance(inpt, features.Image)):
+) -> features.ImageOrVideoTypeJIT:
+    if isinstance(inpt, torch.Tensor) and (
+        torch.jit.is_scripting() or not isinstance(inpt, (features.Image, features.Video))
```
datumbox, line 225: This specific `isinstance` check for image or video is correct, but I think we missed it in all other places on the dispatchers. I think we need to add this everywhere.

datumbox (approved):
LGTM, after addressing some of the comments below. I'll have an other look on the whole API immediately after merging this to ensure we didn't miss anything.

torchvision/prototype/transforms/functional/_augment.py
pmeier: True, it seems I keep missing it :facepalm:

torchvision/prototype/features/_video.py
vfdev-5: What's the difference between `VideoType` and `TensorVideoType` ?

pmeier: There is none. It is only there for parity with the image types.

torchvision/prototype/transforms/_auto_augment.py
pmeier: I renamed to `image_or_videos` to be on par with the `image_or_video` variable in the actual transforms. I didn't like `inputs` here, because that would imply `input = self._extra_image_or_video(...)` and that clashes with the name. Plus, `input` is already used as variable in the actual transformation.

torchvision/prototype/transforms/functional/_color.py
pmeier: >  we just have an extra unnecessary method call. 

That will hardly make a difference. I think we can afford an extra kernel for now. In the future we can revisit this. Especially, if we support 2d images everywhere, most mask kernels could also be aliased. The only exceptions that come to mind are the ones with an interpolation mode.

torchvision/prototype/transforms/functional/_meta.py
pmeier: I went through the whole API again and found a few places where we need to decide if we want video support there:

- classificaton transforms: `MixUp`, `CutMix`
- object detection transforms: `SimpleCopyPaste`, `RandomIoUCrop`
- deprecated transforms: `Grayscale`, `RandomGrayscale`
- outlier transforms: `FiveCrop`, `TenCrop` (we are not supporting anything besides images here so far)
- image only transforms (by name): `ConvertImageDtype`

For everything else this is fixed.

test/prototype_transforms_kernel_infos.py
```diff
@@ -142,6 +143,25 @@ def xfail_jit_list_of_ints(name, *, reason=None):
...
+            "test_scripted_vs_eager",
+            "test_batched_vs_single",
+            "test_no_inplace",
+            "test_cuda_vs_cpu",
+            "test_dtype_and_device_consistency",
+        ]
+    ]
+
+
+xfails_image_degenerate_or_multi_batch_dims = xfail_all_tests(
```
pmeier, line 159: I've added them here to properly test video kernels that have the squash / unsquash "wrappers" now and thus do support arbitrary batch dimensions.

datumbox:
We got breakages on some tests when we throw in batches:
```
TypeError: Input image tensor should have 3 or 4 dimensions, but found 5
_ TestSmoke.test_auto_augment[AugMix-torchvision.prototype.features._video.Video-20] _
  File "/home/runner/work/vision/vision/test/test_prototype_transforms.py", line 183, in test_auto_augment
    transform(input)
  File "/opt/hostedtoolcache/Python/3.7.14/x64/lib/python3.7/site-packages/torch/nn/modules/module.py", line 1190, in _call_impl
  File "/home/runner/work/vision/vision/torchvision/prototype/transforms/_auto_augment.py", line 517, in forward
    aug, transform_id, magnitude, interpolation=self.interpolation, fill=self.fill
  File "/home/runner/work/vision/vision/torchvision/prototype/transforms/_auto_augment.py", line 133, in _apply_image_or_video_transform
    return F.adjust_sharpness(image, sharpness_factor=1.0 + magnitude)
  File "/home/runner/work/vision/vision/torchvision/prototype/transforms/functional/_color.py", line 84, in adjust_sharpness
    return adjust_sharpness_image_tensor(inpt, sharpness_factor=sharpness_factor)
  File "/home/runner/work/vision/vision/torchvision/transforms/functional_tensor.py", line 845, in adjust_sharpness
    return _blend(img, _blurred_degenerate_image(img), sharpness_factor)
  File "/home/runner/work/vision/vision/torchvision/transforms/functional_tensor.py", line 825, in _blurred_degenerate_image
    result_tmp = conv2d(result_tmp, kernel, groups=result_tmp.shape[-3])
RuntimeError: Expected 3D (unbatched) or 4D (batched) input to conv2d, but got input of size: [1, 1, 1, 23, 15]
_ TestSmoke.test_auto_augment[AugMix-torchvision.prototype.features._video.Video-21] _
  File "/home/runner/work/vision/vision/test/test_prototype_transforms.py", line 183, in test_auto_augment
    transform(input)
  File "/opt/hostedtoolcache/Python/3.7.14/x64/lib/python3.7/site-packages/torch/nn/modules/module.py", line 1190, in _call_impl
  File "/home/runner/work/vision/vision/torchvision/prototype/transforms/_auto_augment.py", line 517, in forward
    aug, transform_id, magnitude, interpolation=self.interpolation, fill=self.fill
  File "/home/runner/work/vision/vision/torchvision/prototype/transforms/_auto_augment.py", line 133, in _apply_image_or_video_transform
    return F.adjust_sharpness(image, sharpness_factor=1.0 + magnitude)
  File "/home/runner/work/vision/vision/torchvision/prototype/transforms/functional/_color.py", line 84, in adjust_sharpness
    return adjust_sharpness_image_tensor(inpt, sharpness_factor=sharpness_factor)
  File "/home/runner/work/vision/vision/torchvision/transforms/functional_tensor.py", line 845, in adjust_sharpness
    return _blend(img, _blurred_degenerate_image(img), sharpness_factor)
  File "/home/runner/work/vision/vision/torchvision/transforms/functional_tensor.py", line 825, in _blurred_degenerate_image
    result_tmp = conv2d(result_tmp, kernel, groups=result_tmp.shape[-3])
RuntimeError: Expected 3D (unbatched) or 4D (batched) input to conv2d, but got input of size: [1, 1, 1, 26, 19]
_ TestSmoke.test_auto_augment[AugMix-torchvision.prototype.features._video.Video-22] _
  File "/home/runner/work/vision/vision/test/test_prototype_transforms.py", line 183, in test_auto_augment
    transform(input)
  File "/opt/hostedtoolcache/Python/3.7.14/x64/lib/python3.7/site-packages/torch/nn/modules/module.py", line 1190, in _call_impl
  File "/home/runner/work/vision/vision/torchvision/prototype/transforms/_auto_augment.py", line 517, in forward
    aug, transform_id, magnitude, interpolation=self.interpolation, fill=self.fill
  File "/home/runner/work/vision/vision/torchvision/prototype/transforms/_auto_augment.py", line 133, in _apply_image_or_video_transform
    return F.adjust_sharpness(image, sharpness_factor=1.0 + magnitude)
  File "/home/runner/work/vision/vision/torchvision/prototype/transforms/functional/_color.py", line 84, in adjust_sharpness
    return adjust_sharpness_image_tensor(inpt, sharpness_factor=sharpness_factor)
  File "/home/runner/work/vision/vision/torchvision/transforms/functional_tensor.py", line 845, in adjust_sharpness
    return _blend(img, _blurred_degenerate_image(img), sharpness_factor)
  File "/home/runner/work/vision/vision/torchvision/transforms/functional_tensor.py", line 825, in _blurred_degenerate_image
    result_tmp = conv2d(result_tmp, kernel, groups=result_tmp.shape[-3])
RuntimeError: Expected 3D (unbatched) or 4D (batched) input to conv2d, but got input of size: [1, 3, 3, 24, 22]
_ TestSmoke.test_auto_augment[AugMix-torchvision.prototype.features._video.Video-23] _
  File "/home/runner/work/vision/vision/test/test_prototype_transforms.py", line 183, in test_auto_augment
    transform(input)
  File "/opt/hostedtoolcache/Python/3.7.14/x64/lib/python3.7/site-packages/torch/nn/modules/module.py", line 1190, in _call_impl
  File "/home/runner/work/vision/vision/torchvision/prototype/transforms/_auto_augment.py", line 517, in forward
    aug, transform_id, magnitude, interpolation=self.interpolation, fill=self.fill
  File "/home/runner/work/vision/vision/torchvision/prototype/transforms/_auto_augment.py", line 141, in _apply_image_or_video_transform
    return F.equalize(image)
  File "/home/runner/work/vision/vision/torchvision/prototype/transforms/functional/_color.py", line 204, in equalize
    return equalize_image_tensor(inpt)
  File "/home/runner/work/vision/vision/torchvision/transforms/functional_tensor.py", line 900, in equalize
    raise TypeError(f"Input image tensor should have 3 or 4 dimensions, but found {img.ndim}")
TypeError: Input image tensor should have 3 or 4 dimensions, but found 5
```

We might need to review the mitigations.

torchvision/prototype/transforms/functional/_color.py
```diff
@@ -119,10 +181,31 @@ def autocontrast(inpt: features.InputTypeJIT) -> features.InputTypeJIT:
...
+    _FT._assert_channels(image, [1, 3])
+
+    if image.numel() == 0:
+        return image
+    elif image.ndim == 2:
+        return _FT._scale_channel(image)
+    else:
+        shape = image.shape
+        channels = image.view((-1,) + shape[-2:]).unbind()
+        return torch.stack([_FT._scale_channel(channel) for channel in channels]).view(shape)
```
datumbox, line 199: Why not?
```suggestion
        return _FT._equalize_single_image(channels).view(shape)
```

torchvision/prototype/transforms/functional/_color.py
```diff
@@ -41,10 +53,40 @@ def adjust_contrast(inpt: features.InputTypeJIT, contrast_factor: float) -> feat
...
 
 
-adjust_sharpness_image_tensor = _FT.adjust_sharpness
+def adjust_sharpness_image_tensor(image: torch.Tensor, sharpness_factor: float) -> torch.Tensor:
+    if sharpness_factor < 0:
+        raise ValueError(f"sharpness_factor ({sharpness_factor}) is not non-negative.")
+
+    _FT._assert_image_tensor(image)
+
+    _FT._assert_channels(image, [1, 3])
```
datumbox, line 62: Since we are copy-pasting code, let's take a moment to rewrite on our side these 2 checks. If you see what they do internally, they are pretty bad generating multiple duplicate calls. Could you just rewrite these 2 assertions here?

pmeier:
The breakages that we observed in `AugMix` were two-fold:

1. There was an error in the video generation for the tests. I thought I had fixed it in f1e2bfa407b721ef2edda888ee0a6dc7bc3ed1ed but I needed d8945e68914cdbaf5fcdc5848e45e68bf406657f as well.
2. The problem with `AugMix` is that we don't transform the extracted `image` directly, but perform some operations on it first:

    https://github.com/pytorch/vision/blob/7eb5d7fcab73afec976907a855d9e63fa31f5579/torchvision/prototype/transforms/_auto_augment.py#L485

  This effectively unwraps the `features.Video` into a `torch.Tensor`. When we call the dispatchers later 

  https://github.com/pytorch/vision/blob/7eb5d7fcab73afec976907a855d9e63fa31f5579/torchvision/prototype/transforms/_auto_augment.py#L514-L516

  the input gets dispatched to the image kernel. In theory, that is no issue since the video kernel added by this PR does the same. However, as detailed in #6670, `equalize_image_tensor` and `adjust_sharpness_image_tensor` only support 4d inputs. Batched videos are 5d and thus the call fails.

  Previously, we opted to only fix the video kernels in 0d2ad963e27616f53fc9ee24456e43c927b561da. To use this fix, we would need to wrap the inputs to the AA dispatchers into a `Video` again, which might entail a significant runtime cost until #6681 or #6718 are merged. Thus, I've opted to fix the image kernels here since we need to do this anyway at some point.

torchvision/prototype/transforms/functional/_color.py
pmeier: I wanted to avoid that someone copy-pastes `_equalize_single_image` as it can easily be inlined. I'll add a comment and use it here.

torchvision/prototype/transforms/functional/_color.py
pmeier: Sure.

datumbox (approved):

PR #6671, opened 2022-09-30
handle inplace operations in _Feature.__torch_function__

### LINKED ISSUE #6669: inplace operations on `features._Feature`'s keeps the wrapping in tact
Going by the logic of

https://github.com/pytorch/vision/blob/7f4c55b1dac2418c594267c314afa5353b35608f/torchvision/prototype/features/_feature.py#L74-L81

that should not be happening.

```pycon
>>> image = features.Image(torch.rand(3, 16, 16))
>>> type(image.add(0))
torch.Tensor
>>> type(image.add_(0))
torchvision.prototype.features._image.Image
```

cc @bjuncek

pmeier:
Fixes #6669. At least as far as it is possible. We can only unwrap the result of an inplace operation, i.e. what is returned by it. We cannot change the type inplace and thus we can't prevent the following:

```pycon
>>> image = features.Image(torch.rand(3, 16, 16))
>>> output = image.add_(0)
>>> type(output)
torch.Tensor
>>> type(image)
torchvision.prototype.features._image.Image
```

As you can see `output` is properly unwrapped, but we can't do that for `image` although we have no idea if the operation "invalidated" the type.

### DIFF
--- a/test/test_prototype_transforms_functional.py
+++ b/test/test_prototype_transforms_functional.py
@@ -907,15 +907,15 @@ def test_correctness_gaussian_blur_image_tensor(device, image_size, dt, ksize, s
     torch.testing.assert_close(out, true_out, rtol=0.0, atol=1.0, msg=f"{ksize}, {sigma}")
 
 
-def test_midlevel_normalize_output_type():
+def test_normalize_output_type():
     inpt = torch.rand(1, 3, 32, 32)
     output = F.normalize(inpt, mean=[0.5, 0.5, 0.5], std=[1.0, 1.0, 1.0])
-    assert isinstance(output, torch.Tensor)
+    assert type(output) is torch.Tensor
     torch.testing.assert_close(inpt - 0.5, output)
 
     inpt = make_image(color_space=features.ColorSpace.RGB)
     output = F.normalize(inpt, mean=[0.5, 0.5, 0.5], std=[1.0, 1.0, 1.0])
-    assert isinstance(output, torch.Tensor)
+    assert type(output) is torch.Tensor
     torch.testing.assert_close(inpt - 0.5, output)
 
 
--- a/torchvision/prototype/features/_feature.py
+++ b/torchvision/prototype/features/_feature.py
@@ -102,10 +102,11 @@ class _Feature(torch.Tensor):
             # There are two reasons we want to return the output without wrapping it:
             # 1. The `func` is not one of the exceptions. This implicitly also excludes wrapping of outputs of `func`'s
             #    that don't return a tensor in the first place, e.g. `.numpy()`, `.tolist()`, or `.max(dim=...)`.
-            # 2. The primary operand, i.e. `args[0]`, is the `_Feature`. The __torch_function__ protocol will invoke
+            # 2. The primary operand, i.e. `args[0]`, is not a `_Feature`. The __torch_function__ protocol will invoke
             #    this method on all types involved in the computation by walking the MRO upwards. For example,
-            #    `torch.Tensor(...).to(features.Image(...))` will invoke `features.Image.__torch_function__` first and
-            #    without this guard the original simple tensor would either be wrapped or we fail trying to do so.
+            #    `torch.Tensor(...).to(features.Image(...))` will invoke `features.Image.__torch_function__` with
+            #    `args = (torch.Tensor(), features.Image())`. Without this guard the original plain tensor would either
+            #    be wrapped or we fail trying to do so.
             if func not in cls._NO_WRAPPING_EXCEPTIONS or not isinstance(args[0], cls):
                 # Inplace `func`'s, canonically identified with a trailing underscore in their name like `.add_(...)`,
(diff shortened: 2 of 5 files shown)


torchvision/prototype/features/_feature.py
```diff
@@ -96,18 +102,20 @@ def __torch_function__(
...
-        if not isinstance(args[0], cls):
-            return output
+            # Apart from `func` needing to be an exception, we also require the primary operand, i.e. `args[0]`, to be
+            # an instance of the class that `__torch_function__` was invoked on. The __torch_function__ protocol will
+            # invoke this method on *all* types involved in the computation by walking the MRO upwards. For example,
+            # `torch.Tensor(...).to(features.Image(...))` will invoke `features.Image.__torch_function__` with
+            # `args = (torch.Tensor(), features.Image())` first. Without this guard, the original `torch.Tensor` would
+            # be wrapped into a `features.Image`.
+            if func in cls._NO_WRAPPING_EXCEPTIONS and isinstance(args[0], cls):
+                return cls._NO_WRAPPING_EXCEPTIONS[func](cls, args[0], output)  # type: ignore[no-any-return]
```
datumbox, line 112: Optional Nit: you can avoid the double lookup on the dictionary by using `.get(func, None)`.

datumbox (approved):
LGTM, thanks!

test/test_prototype_transforms_functional.py
```diff
@@ -907,15 +907,15 @@ def test_correctness_gaussian_blur_image_tensor(device, image_size, dt, ksize, s
...
     inpt = torch.rand(1, 3, 32, 32)
     output = F.normalize(inpt, mean=[0.5, 0.5, 0.5], std=[1.0, 1.0, 1.0])
-    assert isinstance(output, torch.Tensor)
+    assert type(output) is torch.Tensor
     torch.testing.assert_close(inpt - 0.5, output)
 
     inpt = make_image(color_space=features.ColorSpace.RGB)
     output = F.normalize(inpt, mean=[0.5, 0.5, 0.5], std=[1.0, 1.0, 1.0])
-    assert isinstance(output, torch.Tensor)
+    assert type(output) is torch.Tensor
```
pmeier, line 918: This test was wrong and this is why it didn't fail before fixing `normalize`: since `features.Image` is a subclass of `torch.Tensor` the check passes. We need to check for type identity here.

test/test_prototype_transforms_functional.py
```diff
@@ -907,15 +907,15 @@ def test_correctness_gaussian_blur_image_tensor(device, image_size, dt, ksize, s
     torch.testing.assert_close(out, true_out, rtol=0.0, atol=1.0, msg=f"{ksize}, {sigma}")
 
 
-def test_midlevel_normalize_output_type():
+def test_normalize_output_type():
```
pmeier, line 910: I renamed since

1. we agreed to use the kernel / dispatcher scheme rather than low / mid level kernels.
2. using `normalize` in the name is sufficient, since the kernel is called `normalize_image_tensor`

torchvision/transforms/functional_tensor.py
```diff
@@ -937,8 +937,7 @@ def normalize(tensor: Tensor, mean: List[float], std: List[float], inplace: bool
         mean = mean.view(-1, 1, 1)
     if std.ndim == 1:
         std = std.view(-1, 1, 1)
-    tensor.sub_(mean).div_(std)
-    return tensor
```
pmeier, line 941: Behavior is equivalent for plain tensors, but gives us the unwrapping behavior we want from this change. With this in place, explicit unwrapping in `normalize` is no longer needed. Let me know if I should keep it there. Runtime is the same, because here we would implicitly unwrap inside `__torch_function__`.

torchvision/prototype/transforms/functional/_misc.py
```diff
@@ -12,12 +12,17 @@
...
 ) -> torch.Tensor:
-    if not isinstance(inpt, torch.Tensor):
-        raise TypeError(f"img should be Tensor Image. Got {type(inpt)}")
+    if torch.jit.is_scripting():
+        correct_type = isinstance(inpt, torch.Tensor)
     else:
-        # Image instance after normalization is not Image anymore due to unknown data range
-        # Thus we return Tensor for input Image
-        return normalize_image_tensor(inpt, mean=mean, std=std, inplace=inplace)
+        correct_type = features.is_simple_tensor(inpt) or isinstance(inpt, features.Image)
```
pmeier, line 18: This gives is the correct error behavior in eager mode. @vfdev-5 noted that before, we could pass bounding boxes or the like to `normalize` and it would only fail in the computation rather than being caught before.

torchvision/transforms/functional_tensor.py
datumbox: Initially I thought this was a styling change. Then I realize you do this to avoid returning `tensor` (which might be an image) and instead return the unwrapped `torch.Tensor`. Might worth adding a short comment in a follow up PR.

PR #6672, opened 2022-09-30
fix bug in output format for pyav

jdsgomes:
fixes bug:

when using `video_reader` backend, `read_video` ignores the `output_format` parameter


bjuncek (approved):
This looks good to me.

YosuaMichael (approved):
Thanks @jdsgomes , the PR looks good to me!

PR #6675, opened 2022-10-01
Handle invalid reduction values

### LINKED ISSUE #6604: Throw ValueError for wrong reduction in losses
As discussed in #6457

## The current problem

We do not restrict the reduction mode in losses. There are only three cases possible, ["none", "mean", "sum"]. If the user passes something like "xyz" we are not throwing any error or warning.

## The solution

If users aren't passing one of the valid cases and using a random string instead. It's better to throw a ValueError suggesting them to use one of the correct choice.

We had thought of Enum based solution, but to keep better consistency with Pytorch core as discussed, we are adopting the above solution.

## Edits required

Need to edit losses. which are namely
[ciou_loss](https://github.com/pytorch/vision/blob/main/torchvision/ops/ciou_loss.py), [diou_loss](https://github.com/pytorch/vision/blob/main/torchvision/ops/diou_loss.py), [giou_loss](https://github.com/pytorch/vision/blob/main/torchvision/ops/giou_loss.py),  and [focal_loss](https://github.com/pytorch/vision/blob/main/torchvision/ops/focal_loss.py)

Also need to edit the relevant tests in [test_ops.py](https://github.com/pytorch/vision/blob/main/test/test_ops.py)

## Additional Context

I think it's good first issue and some new contributor can take this :smiley:

Feel free to ping the maintainers or me if you need help. :smile:

adityagandhamal:
Fixes #6604 

As stated, reduction methods are restricted to ["none", "mean",  "sum"]. 
This PR contains modifications to the code which prompts the user a ValueError if an anonymous value is passed.

Also, added respective unit tests to support the changes.


oke-aditya:
Think a bit more. Can we make this bit more efficient? Are we doing unnecessary computation?

torchvision/ops/focal_loss.py
```diff
@@ -32,6 +32,14 @@ def sigmoid_focal_loss(
         Loss tensor with the reduction option applied.
     """
     # Original implementation from https://github.com/facebookresearch/fvcore/blob/master/fvcore/nn/focal_loss.py
+
+    # Check reduction option
+    modes = ["none", "mean", "sum"]
+    if reduction not in modes:
+        raise ValueError(
+            f"Invalid Value for arg 'reduction': '{reduction} \n Supported reduction modes: 'none', 'mean', 'sum'"
+        )
```
datumbox, line 41: What is the reason for using a different idiom here? Why not do the same if/elif/else approach as with the other losses?

datumbox (commented):
Thanks for the PR @adityagandhamal. Glad you managed to solve the installation issues you phased. The overall approach looks good, I only have a couple of comments. See below:

torchvision/ops/focal_loss.py
oke-aditya: My initial suggestion was to check the valid reduction mode even before we compute the losses. Throw value error immediately if it's badly configured.

datumbox: That make sense but that idiom requires us to maintain 2 separate lists with accepted modes. It also deviates from what all the other kernels are doing. It might be worth adopting a common strategy that aligns closely with Core. Thoughts?

oke-aditya: While aligning with core is OK. But it's a waste of computation. Compute the entire loss just to throw a valueError? 

And we could possibly refactor this a utility.

adityagandhamal: > What is the reason for using a different idiom here? Why not do the same if/elif/else approach as with the other losses?

Before applying the reduction method to the final loss, we compute `ce_loss`  first (which requires a valid reduction value).
Were it not for using that specific code, we would have got repetitive blocks of if-else statements.

But, as far as the idiom is concerned we can stick with the if-else approach.

adityagandhamal: > While aligning with core is OK. But it's a waste of computation. Compute the entire loss just to throw a valueError?
> 
> And we could possibly refactor this a utility.

Agree

test/test_ops.py
```diff
@@ -1367,6 +1367,8 @@ def assert_empty_loss(iou_fn, dtype, device):
     loss = iou_fn(box1, box2, reduction="none")
     assert loss.numel() == 0, f"{str(iou_fn)} for two empty box should be empty"
 
+def assert_reduction_mode(iou_fn, box1, box2, reduction):
+    assert iou_fn(box1, box2, reduction) == ValueError
```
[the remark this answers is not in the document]
adityagandhamal: I referred to the common structure of all the tests. More so than this, your point is valid. I will make the necessary changes accordingly.

torchvision/ops/focal_loss.py
oke-aditya: I thought we can use just this idiom and it does the job. I'm not getting how we need 2 lists

datumbox: Thanks for the comments, a few clarifications on my end:

> I'm not getting how we need 2 lists

We keep track the available options twice because the options are recorded once in the list (which you constantly reinitialise and look serially) for every call and once on the series of if/elif/else statements.

> But it's a waste of computation.

Should we be optimizing for the case of returning the error sooner? This seems to me an unnecessary optimization which leads to an idiom that is not present anywhere else in the codebase (TorchVision or Core). This is why I proposed to remove it.

adityagandhamal: Fair enough. I should point out that maintaining a consistent structure overall will help new contributors like myself catch up with the codebase.

adityagandhamal:
@datumbox 
I've done the changes as advised.

torchvision/ops/focal_loss.py
```diff
@@ -32,6 +32,13 @@ def sigmoid_focal_loss(
         Loss tensor with the reduction option applied.
     """
     # Original implementation from https://github.com/facebookresearch/fvcore/blob/master/fvcore/nn/focal_loss.py
+
+    # Check reduction option
+    if reduction != "none" or reduction != "mean" or reduction != "sum":
+        raise ValueError(
+            f"Invalid Value for arg 'reduction': '{reduction} \n Supported reduction modes: 'none', 'mean', 'sum'"
+        )
```
datumbox, line 40: Could you please move this on lines 53-56 similar to the other loss functions?

adityagandhamal: Will do

adityagandhamal: I have also made sure that all the loss functions follow the same idiom at checking the reduction value.

datumbox (approved):
LGTM, thanks a lot for the contribution @adityagandhamal.

@oke-aditya thanks for the support!

I think the issue with the [linter](https://app.circleci.com/pipelines/github/pytorch/vision/20826/workflows/d18f3f0e-0cfa-4bcc-bc9a-1890f1fcc46e/jobs/1693898) is unrelated. I will check again tomorrow to see if it's fixed. cc @pmeier

pmeier:
> I think the issue with the [linter](https://app.circleci.com/pipelines/github/pytorch/vision/20826/workflows/d18f3f0e-0cfa-4bcc-bc9a-1890f1fcc46e/jobs/1693898) is unrelated. I will check again tomorrow to see if it's fixed.

It is and this was fixed in #6678. Updating the branch should be sufficient.

datumbox:
@pmeier Thanks a lot for having a look. I confirm the job runs now correctly.

@adityagandhamal Unfortunately there are indeed some linter issues. You can see them [here](https://app.circleci.com/pipelines/github/pytorch/vision/20829/workflows/6c60c97f-315e-42f3-a6cb-eba426242aad/jobs/1694121), by clicking on the `Required lint modifications` section. You don't need to fix them manually, just run `ufmt format .` from the root of the project and the utility will fix it for you. Then commit and push and we should be good to go. :)

oke-aditya:
While I agree with @datumbox and thanks a lot to @adityagandhamal for all the work! 

> Should we be optimizing for the case of returning the error sooner? This seems to me an unnecessary optimization which leads to an idiom that is not present anywhere else in the codebase.

IMO the silver bullet  was using Enums. Which automatically prevents invalid reduction modes and would have returned error sooner as well as kept the code clean.

I think Enum was not used in core due to JIT.

That's fine let's accept the current solution as it too works. Anyways let's keep things simple and move ahead :smiley

adityagandhamal:
> just run `ufmt format .` from the root of the project and the utility will fix it for you. Then commit and push and we should be good to go. :)

Okay

datumbox (approved):
LGTM (again)! Let's wait for green CI and merge. 😃

oke-aditya:
If They are not be caused by your PR you don't need to worry. There may be some other changes / many reasons for Failures.

datumbox:
All good. The CI failure is unrelated. Thanks for your work!

adityagandhamal:
Thank you!
Grateful to all for the support. Looking forward to contributing more in the future.

PR #6680, opened 2022-10-03
Vision collate

pmeier:
Addresses https://github.com/pytorch/vision/pull/6433#discussion_r979067780. I've intentionally left out the collation function, since we just use

https://github.com/pytorch/vision/blob/07ae61bf9c21ddd1d5f65d326aa9636849b383ca/references/detection/utils.py#L203-L204

which is completely separated from `default_collate`. We might provide it publicly later on as well.

### DIFF
--- a/torchvision/prototype/features/_collate.py
+++ b/torchvision/prototype/features/_collate.py
@@ -1,7 +1,7 @@
 from typing import Any, Callable, Dict, Optional, Sequence, Tuple, Type, Union
 
 from torch.utils.data._utils.collate import collate, collate_tensor_fn, default_collate_fn_map
-from torchvision.prototype.features import BoundingBox, Image, Label, Mask, OneHotLabel
+from torchvision.prototype.features import Image, Label, Mask, OneHotLabel
 
 
 def no_collate_fn(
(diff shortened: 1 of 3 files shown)


torchvision/prototype/features/_collate.py
```diff
@@ -0,0 +1,29 @@
...
+    batch: Sequence[Any], *, collate_fn_map: Optional[Dict[Union[Type, Tuple[Type, ...]], Callable]] = None
+) -> Any:
+    feature = batch[0]
+    tensor = collate_tensor_fn(batch, collate_fn_map=collate_fn_map)
+    return type(feature).new_like(feature, tensor)
+
+
+vision_collate_fn_map = {
+    (Image, Mask, Label, OneHotLabel): new_like_collate_fn,
+    type(None): no_collate_fn,
```
pmeier, line 23: We need this for #5233.

torchvision/prototype/features/_collate.py
```diff
@@ -0,0 +1,29 @@
+from typing import Any, Callable, Dict, Optional, Sequence, Tuple, Type, Union
```
pmeier, line 1: I've put this into `features` for now, put when everything is rolled out it should probably be in `datasets`.

ejguan (approved):
LGTM in terms of `vision_collate`

datumbox (changes requested):
Ideally we should avoid adding collate methods directly in TorchVision. One option is to keep it on the references. That's something that we plan to discuss with @pmeier once he is back from PTO. The discussion can go either way so, everything is on the table.

I'm going to mark the PR with "request changes" to avoid accidental merges (since 1 approval = merge). After we sync with Philip, we can unblock this. 😃

PR #6681, opened 2022-10-03
[proto] Reduce number of calls of __torch_function__

vfdev-5:
An attempt to improve API v2 performances by reducing the number of `_Feature.__torch_function__` calls.

CProfiling classification AA(imagenet) + RandomErasing ([code](https://github.com/vfdev-5/tvapiv2_benchmarks/tree/8d24dacbf981f5e9469a28c6c6795ac9428b4f62)) : 
```
# python -u main.py cprofile_pil_vs_feature --n=2000
Profile API v2 on Feature
Compose(
      ToImageTensor()
      RandomResizedCrop(size=(224, 224), scale=(0.08, 1.0), ratio=(0.75, 1.3333333333333333), interpolation=InterpolationMode.BILINEAR, antialias=True)
      RandomHorizontalFlip(p=1.0)
      AutoAugment(interpolation=InterpolationMode.BILINEAR, policy=AutoAugmentPolicy.IMAGENET)
      ConvertImageDtype()
      Normalize(mean=[0.485, 0.456, 0.406], std=[0.229, 0.224, 0.225], inplace=False)
      RandomErasing(p=1.0, scale=(0.02, 0.33), ratio=(0.3, 3.3), value=0, inplace=False)
)
```

[Main](https://github.com/vfdev-5/tvapiv2_benchmarks/blob/8d24dacbf981f5e9469a28c6c6795ac9428b4f62/output/20221003-132331_cprof_v2_feat_classification_imagenet_ra_main_07ae61bf9.log):
```
         1827041 function calls (1792534 primitive calls) in 6.310 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     2000    1.450    0.001    1.450    0.001 {built-in method torch._C._nn._upsample_bilinear2d_aa}
15251/13251    0.518    0.000    0.565    0.000 {method 'to' of 'torch._C._TensorBase' objects}
     2628    0.466    0.000    0.855    0.000 /vision/torchvision/transforms/functional_tensor.py:875(_scale_channel)
      309    0.233    0.001    0.233    0.001 {built-in method torch.grid_sampler}
    96931    0.208    0.000    0.997    0.000 /vision/torchvision/prototype/features/_feature.py:71(__torch_function__)
4000/2000    0.201    0.000    0.208    0.000 {method 'flip' of 'torch._C._TensorBase' objects}
     2628    0.175    0.000    0.175    0.000 {built-in method torch.bincount}
4062/4031    0.171    0.000    0.172    0.000 {method 'clone' of 'torch._C._TensorBase' objects}
...
```

[This PR](https://github.com/vfdev-5/tvapiv2_benchmarks/blob/8d24dacbf981f5e9469a28c6c6795ac9428b4f62/output/20221003-132803_cprof_v2_feat_classification_imagenet_ra_4767fef93.log):
```
         1269905 function calls (1243829 primitive calls) in 5.986 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     2000    1.454    0.001    1.454    0.001 {built-in method torch._C._nn._upsample_bilinear2d_aa}
15251/13251    0.527    0.000    0.551    0.000 {method 'to' of 'torch._C._TensorBase' objects}
     2628    0.477    0.000    0.875    0.000 /vision/torchvision/transforms/functional_tensor.py:875(_scale_channel)
      309    0.231    0.001    0.231    0.001 {built-in method torch.grid_sampler}
4000/2000    0.203    0.000    0.216    0.000 {method 'flip' of 'torch._C._TensorBase' objects}
     2628    0.177    0.000    0.177    0.000 {built-in method torch.bincount}
4062/4031    0.171    0.000    0.171    0.000 {method 'clone' of 'torch._C._TensorBase' objects}
...
    18305    0.084    0.000    0.757    0.000 /vision/torchvision/prototype/features/_feature.py:74(__torch_function__)
     4000    0.073    0.000    0.754    0.000 /vision/torchvision/prototype/transforms/_transform.py:66(forward)
     2000    0.067    0.000    0.154    0.000 /vision/torchvision/transforms/functional_tensor.py:943(erase)
...
```

Time measurements on all classification dataaug pipelines. Improvements are ~1ms:
- https://diffonline.net/8LY9axh4iO
- https://github.com/vfdev-5/tvapiv2_benchmarks/blob/8d24dacbf981f5e9469a28c6c6795ac9428b4f62/output/20221003-121623_output_classification_all_pil_vs_features_main_07ae61bf9.log
- https://github.com/vfdev-5/tvapiv2_benchmarks/blob/8d24dacbf981f5e9469a28c6c6795ac9428b4f62/output/20221003-121856_output_classification_all_pil_vs_features_4767fef93.log

cc @pmeier @datumbox


torchvision/prototype/features/_feature.py
```diff
@@ -137,6 +140,28 @@ def _F(self) -> ModuleType:
             _Feature.__F = functional
         return _Feature.__F
 
+    # Add properties for common attributes like shape, dtype, device, ndim etc
+    # this way we return the result without passing into __torch_function__
+    @property
+    def shape(self):
+        return self.as_subclass(torch.Tensor).shape
```
datumbox, line 147: Have we tried maintaining the original reference of the input tensor `data` prior getting wrapped in the subclassed tensor? @ezyang proposed this idea offline as a means to speed things up. He just cautioned to keep this private because if a user gets hold of the original reference and do `original.resize_()` this change will be available only on that tensor and not on the wrapped one.

torchvision/prototype/features/_image.py
```diff
@@ -97,9 +97,10 @@ def __repr__(self, *, tensor_contents: Any = None) -> str:  # type: ignore[overr
     def new_like(
         cls, other: Image, data: Any, *, color_space: Optional[Union[ColorSpace, str]] = None, **kwargs: Any
     ) -> Image:
-        return super().new_like(
-            other, data, color_space=color_space if color_space is not None else other.color_space, **kwargs
-        )
+        # Question: Is it safe to assume data to be a tensor ?
+        out = data.as_subclass(Image)
```
pmeier, line 101: No. `features.Label(0)` is a good counter example.

pmeier (commented):
IIUC, we are only using `.as_subclass(torch.Tensor)` here to avoid `__torch_function__` here, correct? If yes, why not simply use the context manager that disables this?

https://github.com/pytorch/vision/blob/07ae61bf9c21ddd1d5f65d326aa9636849b383ca/torchvision/prototype/features/_feature.py#L102-L103

like

```py
with DisableTorchFunction():
    return self.shape
```

Better yet, instead of adding properties for all these attributes manually, why not overwrite `__getattribute__` to do this for us?

```py
    _NO_TORCH_FUNCTION = {
        "shape",
    }

    def __getattribute__(self, item):
        with DisableTorchFunction() if item in type(self)._NO_TORCH_FUNCTION else contextlib.nullcontext():
            return object.__getattribute__(self, item)
```

Test:

```py
import unittest.mock

import torch

image = features.Image(torch.rand(3, 16, 16))

with unittest.mock.patch(
    "torchvision.prototype.features._Feature.__torch_function__",
    side_effect=AssertionError,
):
    image.shape
```

I have not benchmarked this. If it is as fast or faster than your suggestion, I would prefer this for several reasons:

1. We don't need to add properties manual, which is harder to maintain.
2. Subclasses can extend `_NO_TORCH_FUNCTION` or whatever we call it to also avoid `__torch_function__` calls on the metadata they are adding, i.e. `Image.color_space`.
3. Since we don't need to unwrap anymore, we can also avoid the issues @datumbox mentioned in https://github.com/pytorch/vision/pull/6681/files#r986001975.

torchvision/prototype/features/_image.py
vfdev-5: This code is inside `Image`, Label wont pass here IMO

vfdev-5:
> IIUC, we are only using .as_subclass(torch.Tensor) here to avoid __torch_function__ here, correct? If yes, why not simply use the context manager that disables this?

Yes, it is to disable torch function. I can benchmark but I assume that using the context manager will be slower as we should be calling more functions (e.g. enter/exit of CM). Let's see.

vfdev-5:
Here are some time measurements for different options:
```python
import torch
from torch._C import DisableTorchFunction


class ATensor(torch.Tensor):
    
    def __new__(
        cls,
        data,
        *,
        dtype=None,
        device=None,
        requires_grad: bool = False,
    ):
        data = torch.as_tensor(  # type: ignore[return-value]
            data,
            dtype=dtype,  # type: ignore[arg-type]
            device=device,  # type: ignore[arg-type]
        )
        
        output = data.as_subclass(cls).requires_grad_(requires_grad)
        output._data_tensor = data
        return output
    
    @classmethod
    def __torch_function__(cls, func, types, args=(), kwargs=None):
        if kwargs is None:
            kwargs = {}
        return super().__torch_function__(func, types, args, kwargs)
    
    @property
    def shape2(self):
        return self.as_subclass(torch.Tensor).shape

    @property
    def shape3(self):
        return self._data_tensor.shape
    
    @property
    def shape4(self):
        with DisableTorchFunction():
            return self.shape

a = ATensor(torch.tensor([0, 1, 2]))
```

```
%%timeit
s = 0
for _ in range(100):
    s += sum(a.shape)  # original

462 µs ± 9.4 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

%%timeit
s = 0
for _ in range(100):
    s += sum(a.shape2)  # self.as_subclass(torch.Tensor)

189 µs ± 1.4 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

%%timeit
s = 0
for _ in range(100):
    s += sum(a.shape3)  # self._data_tensor

33.2 µs ± 552 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

%%timeit
s = 0
for _ in range(100):
    s += sum(a.shape4)  # DisableTorchFunction

54.3 µs ± 661 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```


The fastest is using a ref to original torch tensor and the second fastest is `DisableTorchFunction`.
Between these two options I think `DisableTorchFunction` is better as there are no side-effects with inplace ops on underlying tensor.

datumbox:
@vfdev-5 Sounds good. Let's proceed with `DisableTorchFunction`. IMO moving with properties is OK. It makes the code readable and simpler comparing to using `__getattribute__`. I would assume it's also marginally faster, as it avoids additional ops. The cost of maintaining them is not that big, only few lines of code placed in the `_Feature` class.

Looking forward to seeing what impact this has on the overall performance.

vfdev-5:
> Looking forward to seeing what impact this has on the overall performance.

Overall impact of such changes is about ~0.1 ms vs `main`.

Using ref is a bit faster that using `DisableTorchFunction`: 
- WIth ref : https://github.com/vfdev-5/tvapiv2_benchmarks/blob/master/output/20221005-150028_output_classification_all_pil_vs_features_6435c095a.log
- With `DisableTorchFunction` : https://github.com/vfdev-5/tvapiv2_benchmarks/blob/master/output/20221005-150214_output_classification_all_pil_vs_features_5ba3c0d58.log
- Main : https://github.com/vfdev-5/tvapiv2_benchmarks/blob/master/output/20221005-145730_output_classification_all_pil_vs_features_a46c4f0cc.log 


However, there are few blockers/issues in the implementation (not yet pushed here)... WIP

ezyang:
By the way, the core team is open to optimizing the relevant apis to make them faster. We didn't really do any benchmarking on them. One easy win is to have subclasses preregister the set of `__torch_function__` they are interested in interposing on, and fastpathing when they are not.

vfdev-5:
> One easy win is to have subclasses preregister the set of __torch_function__ they are interested in interposing on, and fastpathing when they are not.

Yes, this could be helpful

datumbox:
@ezyang Thanks for offering help. Absolutely, we would love to work with you to make the solution faster. Perhaps we can start by a quick review on our practices (I believe this PR is a good place to start), to let us know if we make improper use of some of the Subclassing features and whether there is any low hanging fruits we can improve. Happy to chat more about dedicating some time to improve the solution.

torchvision/prototype/features/_feature.py
ezyang: Is this really faster than going into torch function? You have to do a tensor allocation here!

ezyang:
The lines edited by this PR don't look obviously wrong but I haven't looked at the whole thing.

@vfdev-5 are you open to implementing some of the core optimizations? We're pretty low bandwidth on our side, but I can definitely find time to give design and review

vfdev-5:
@ezyang we are right now very tight on the schedule for v2, but yes, I can be open to implement some core optimization a bit later once we are done with python optims. Thanks for proposing, let me ping you once we can put an effort there.

vfdev-5:
After merging https://github.com/pytorch/vision/pull/6718 perf improvements are less important in terms of time measurements (~ 0.01 ms), however we can reduce the number of `__torch_function__` calls.

Main: `65932    0.156    0.000    0.888    0.000 /vision/torchvision/prototype/features/_feature.py:62(__torch_function__)`
This PR: `16747    0.073    0.000    0.767    0.000 /vision/torchvision/prototype/features/_feature.py:67(__torch_function__)`

torchvision/prototype/features/_encoded.py
```diff
@@ -16,7 +16,9 @@
 class EncodedData(_Feature):
     @classmethod
     def _wrap(cls: Type[D], tensor: torch.Tensor) -> D:
-        return tensor.as_subclass(cls)
+        output = tensor.as_subclass(cls)
+        output._tensor = tensor
```
datumbox, line 20: Originally you had concerns adopting this approach in order to avoid issues with in-place operators. Assuming we keep this reference private, are we safe from in-place operators? Aka do the shapes of `img` and `img._tensor` match on the following?
```
x = torch.randn((10,1,1,1))
img = Image(x)
img.resize_((10, 1, 10, 10))
```

vfdev-5: No, we are not safe with `resize_`-like ops and the shapes wont match...
This is a drawback of using tensor ref...

datumbox: Then probably that's something we would like to avoid, right? Shall we proceed with your original proposal (https://github.com/pytorch/vision/pull/6681#issuecomment-1268703395) to use `DisableTorchFunction`? My understanding is that this would be safe for inplace ops with little performance penalty?

vfdev-5: Yes, it would be better to avoid that. I'll check DisableTorchFunction again but seeing that the gain is very tiny and around benchmarking noise maybe we can search for other optims now

datumbox: I'm open to it if we can guarantee it's safe. If we go ahead with keeping a reference, is it possible to do a bit of refactoring to use `__` and make it much harder for users to shoot themselves in the foot? I suspect the issue here is subclassing of our `_Feature` but I'm open to ideas.

@vfdev-5 WDYT of Philip's proposal?

vfdev-5: I check that later, but as commented previously, I do not see that this update brings much perf gain, just very tiny speed-up. Let me come back to this PR later. I'll put it as draft now.

vfdev-5:
With `DisableTorchFunction` ([3799ce7](https://github.com/pytorch/vision/pull/6681/commits/3799ce7228e83d3e2190461e96189ccd405dbfa7)) we have : 
```diff
- 58336    0.147    0.000    0.513    0.000 /vision/torchvision/prototype/features/_feature.py:62(__torch_function__)
+ 16747    0.070    0.000    0.376    0.000 /vision/torchvision/prototype/features/_feature.py:64(__torch_function__)
```

Sources: [before](https://github.com/vfdev-5/tvapiv2_benchmarks/blob/master/output/20221014-105254_cprof_v2_feat_classification_imagenet_ra_main_before_reducing_pytree.log), [after](https://github.com/vfdev-5/tvapiv2_benchmarks/blob/master/output/20221014-153131_cprof_v2_feat_classification_imagenet_ra_3799ce722.log)

Time diffs for classification pipelines, one example:
```diff
[ Classification AA=ra RE=1.0 transforms measurements ]
                          |    v2 
1 threads: -----------------------
      Tensor Image data   |  3.803
-      Feature Image data  |  3.986
+      Feature Image data  |  3.931
```
Sources: [before](https://github.com/vfdev-5/tvapiv2_benchmarks/blob/master/output/20221014-155643_output_classification_all_ten_vs_feat_v2_e3238e5af.log), [after](https://github.com/vfdev-5/tvapiv2_benchmarks/blob/master/output/20221014-160926_output_classification_all_ten_vs_feat_v2_3799ce722.log)

vfdev-5:
Results for `_tensor` approach ([38f8e21](https://github.com/pytorch/vision/pull/6681/commits/38f8e21242830fed46ddf31287edb67c1abd124a)):

```diff
[ Classification AA=augmix RE=1.0 transforms measurements ]
                          |    v2 
1 threads: -----------------------
-      Tensor Image data   |  7.846
-      Feature Image data  |  7.916
+      Tensor Image data   |  7.798     <---- this is quicker than the same test  <-> noise
+      Feature Image data  |  7.875
```


Source: [DisableTorchFunction approach](https://github.com/vfdev-5/tvapiv2_benchmarks/blob/master/output/20221014-160926_output_classification_all_ten_vs_feat_v2_3799ce722.log), [`_tensor` approach](https://github.com/vfdev-5/tvapiv2_benchmarks/blob/master/output/20221014-174128_output_classification_all_ten_vs_feat_v2_38f8e2124.log)

torchvision/prototype/features/_feature.py
```diff
@@ -42,7 +43,8 @@ def __new__(
         requires_grad: bool = False,
     ) -> _Feature:
         tensor = cls._to_tensor(data, dtype=dtype, device=device, requires_grad=requires_grad)
-        return tensor.as_subclass(_Feature)
+        output = tensor.as_subclass(_Feature)
+        return output
```
pmeier, line 47: Could you revert this?

```suggestion
        return tensor.as_subclass(_Feature)
```

pmeier (approved):
One nit inline. Otherwise LGTM, thanks Victor! 

One question though for my own understanding: we don't need this for the feature specific attributes like `Image.color_space` since they are not part of `__torch_function__` anyway, right?

vfdev-5:
> we don't need this for the feature specific attributes like Image.color_space since they are not part of __torch_function__ anyway

Seems like additional attributes like color_space are not routed to `__torch_function__`, so no need to make them as properties.

PR #6682, opened 2022-10-03
Fix error message typo

robert-perrotta:
Just a typo fix. There were two `n`s in `training` in an error message.


NicolasHug (approved):
Thanks @robert-perrotta

PR #6684, opened 2022-10-03
Fix missing f-string prefix in error message

robert-perrotta:
The affected error message was missing the prefix to make it a format string.


NicolasHug (approved):
Thanks @robert-perrotta

PR #6685, opened 2022-10-03
[ONNX] Support exporting RoiAlign align=True to ONNX with opset 16

BowenBao:
Long over due, but it's here.
ONNX upgrades its 'RoiAlign' spec in opset 16 to support `align=True` via 'coordinate_transformation_mode'.
This PR adds support to export 'RoiAlign' under opset 16, and re-enabled disabled tests.

Together with [TODO PT PR] will fix pytorch/pytorch#81121

### DIFF
--- a/torchvision/ops/_register_onnx_ops.py
+++ b/torchvision/ops/_register_onnx_ops.py
@@ -10,8 +10,8 @@ base_onnx_opset_version = _onnx_opset_version_11
 
 def _register_custom_op():
     from torch.onnx.symbolic_helper import parse_args
-    from torch.onnx.symbolic_opset9 import _cast_Long
     from torch.onnx.symbolic_opset11 import select, squeeze, unsqueeze
+    from torch.onnx.symbolic_opset9 import _cast_Long
 
     @parse_args("v", "v", "f")
     def symbolic_multi_label_nms(g, boxes, scores, iou_threshold):
--- a/torchvision/ops/_register_onnx_ops.py
+++ b/torchvision/ops/_register_onnx_ops.py
@@ -3,13 +3,15 @@ import warnings
 
 import torch
 
-_onnx_opset_version = 11
+_onnx_opset_version_11 = 11
+_onnx_opset_version_16 = 16
+base_onnx_opset_version = _onnx_opset_version_11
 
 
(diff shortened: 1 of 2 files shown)


datumbox (approved):
LGTM, thanks @BowenBao.

We can merge on green CI.

torchvision/ops/_register_onnx_ops.py
```diff
@@ -20,32 +22,56 @@ def symbolic_multi_label_nms(g, boxes, scores, iou_threshold):
...
+        return sampling_ratio
+
+    @parse_args("v", "v", "f", "i", "i", "i", "i")
+    def roi_align_opset11(g, input, rois, spatial_scale, pooled_height, pooled_width, sampling_ratio, aligned):
+        batch_indices = _process_batch_indices_for_roi_align(g, rois)
+        rois = _process_rois_for_roi_align(g, rois)
+        if aligned:
+            warnings.warn(
+                "ROIAlign with aligned=True is not supported in ONNX, but is supported in opset 16. "
+                "Please export with opset 16 or higher to use aligned=False."
```
justinchuby, line 49: @BowenBao does this mean users will still need to set it to False with version>=16? Is this intended?

BowenBao: omg copilot ...

BowenBao:
@datumbox Can this be included in 0.14 release such that PyTorch 1.13 release has the fix?

datumbox:
@BowenBao I just checked and the PR was merged after the release cut. As this is a feature and not a bug fix, we would probably need to understand a bit better the confidence you have over this feature (for full transparency we are not ONNX power-users). Do you think this is safe to merge on the release branch? Have you tried exporting any model that uses the op to confirm? Thanks in advance. 

cc @YosuaMichael who is the POC for this release.

YosuaMichael:
@BowenBao @datumbox Could you explain more on whether this PR is a bugfix (something will break if we dont include this PR), or is it more on feature (nothing broken if this PR is not included)?

I am not that familiar with ONNX as well, so need to know more detail on this. Also agree with @datumbox , would be great to know how safe is this and what test has been done to make sure it is safe.

PR #6690, opened 2022-10-04
Adding BibTeX entry for TorchVision

datumbox:
Resolving https://github.com/pytorch/pytorch/issues/4126#issuecomment-1266180365

Following the format as proposed [here](https://academia.stackexchange.com/questions/14010/how-do-you-cite-a-github-repository/14015#14015).


pmeier (commented):
GH has a custom format for defining the citation info: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files

TL;DR you add a `CITATION.cff` file to the root of the project and GH adds a button on the top right column of the project home like they have in PyTorch core

Screenshot from 2022-10-04 11-34-54

PyTorch core just defines a `CITATION` file, which is why there is no information rendered directly. If you follow their protocol, the prompt looks like 

Screenshot from 2022-10-04 11-36-39

I would prefer us to use this protocol. I can also send a PR to core. Of course we can also put the info in the README.

soumith (approved):

PR #6691, opened 2022-10-04
Bump main version to 0.15

YosuaMichael:
For release v0.14, we want to bump the main version to 0.15


datumbox:
Looks good @YosuaMichael. I think we should also update the README.srt file (table).

This is not for this PR, but as a reminder we should also update the `ios/LibTorchvision.podspec` file after the release.

datumbox (approved):
LGTM, thanks @YosuaMichael.

As Yosua said offline, we should probably update the tables after PyTorch has released the new binaries. This can be done after the release.

Issue #6696, opened 2022-10-04
VideoClips.subset() does not reflect output_format of the original VideoClips


Simcs:
### 🐛 Describe the bug

When new ```VideoClips``` is created using ```VideoClips.subset()``` function, ```output_format``` of the original ```VideoClips``` is not taken into account.

Because of this, passing ```output_format="TCHW"``` argument to ```torchvision.datasets.hmdb51.HMDB51``` does not work.
It seems that ```HMDB51``` creates ```full_video_clips``` and ```video_clips``` for each ```fold``` of the dataset using ```subset()``` function.

### Code snippet
Below is the code snippet for reproducing the bug:
```python
import os
from torchvision.datasets import HMDB51

dataset_root = '~/workspace/dataset/hmdb51'
dataset_root = os.path.expanduser(dataset_root)
annotation_path = '~/workspace/dataset/testTrainMulti_7030_splits'
annotation_path = os.path.expanduser(annotation_path)

hmdb51_thwc = HMDB51(
    root=dataset_root,
    annotation_path=annotation_path,
    frames_per_clip=16,
    step_between_clips=1,
    frame_rate=2,
    fold=1,
    train=False,
    num_workers=8,
    output_format="THWC",
)
video, audio, class_index = hmdb51_thwc[0]
print(video.shape)

hmdb51_tchw = HMDB51(
    root=dataset_root,
    annotation_path=annotation_path,
    frames_per_clip=16,
    step_between_clips=1,
    frame_rate=2,
    fold=1,
    train=False,
    num_workers=8,
    output_format="TCHW",
)
video, audio, class_index = hmdb51_tchw[0]
print(video.shape)
```

### Output:
```
torch.Size([16, 240, 320, 3])
torch.Size([16, 240, 320, 3])
```

### Possible solution:
Passing ```output_format``` of the original ```VideoClips``` when instantiating new ```VideoClips``` in ```VideoClips.subset()``` function.

### Versions

```
Collecting environment information...
PyTorch version: 1.12.1+cu113
Is debug build: False
CUDA used to build PyTorch: 11.3
ROCM used to build PyTorch: N/A

OS: Ubuntu 20.04.4 LTS (x86_64)
GCC version: (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0
Clang version: Could not collect
CMake version: Could not collect
Libc version: glibc-2.31

[...]

PR #6700, opened 2022-10-04
[bugfix] Fix the output format for VideoClips.subset

YosuaMichael:
Resolve the bug reported in https://github.com/pytorch/vision/issues/6696


datumbox (approved):
LGTM, thanks!

Issue #5766, opened 2022-04-06
Wrong label mapping in vision/torchvision/datasets/flowers102.py


YEthYuan:
### 🐛 Describe the bug

from line 70 in the file vision/torchvision/datasets/flowers102.py
```
self._labels = []
self._image_files = []
for image_id in image_ids:
    self._labels.append(image_id_to_label[image_id])  # Note: the bug is here, current the labels are ranged from 1 to 102, which should be ranged from 0 to 101, consider to change this line into self._labels.append(image_id_to_label[image_id] - 1) ?
    self._image_files.append(self._images_folder / f"image_{image_id:05d}.jpg")
```

### Versions

PyTorch version: 1.11.0
Is debug build: False
CUDA used to build PyTorch: 11.3
ROCM used to build PyTorch: N/A

OS: Ubuntu 20.04.4 LTS (x86_64)
GCC version: (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0
Clang version: Could not collect
CMake version: version 3.16.3
Libc version: glibc-2.31

Python version: 3.8.13 (default, Mar 28 2022, 11:38:47)  [GCC 7.5.0] (64-bit runtime)
Python platform: Linux-5.4.0-104-generic-x86_64-with-glibc2.17
Is CUDA available: True
CUDA runtime version: Could not collect
GPU models and configuration: GPU 0: NVIDIA GeForce RTX 3090
Nvidia driver version: 510.54
cuDNN version: Could not collect
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

Versions of relevant libraries:
[pip3] advertorch==0.2.3
[pip3] numpy==1.21.2
[pip3] torch==1.11.0
[pip3] torchaudio==0.11.0
[pip3] torchvision==0.7.0
[conda] advertorch                0.2.3                    pypi_0    pypi
[conda] blas                      1.0                         mkl    https://mirrors.tuna.tsinghua.edu.cn/anaconda/pkgs/free
[conda] cudatoolkit               11.3.1               h2bc3f7f_2    https://mirrors.tuna.tsinghua.edu.cn/anaconda/pkgs/main
[conda] ffmpeg                    4.3                  hf484d3e_0    pytorch
[conda] mkl                       2021.4.0           h06a4308_640    https://mirrors.tuna.tsinghua.edu.cn/anaconda/pkgs/main
[...]

PR #6702, opened 2022-10-04
[bugfix] Make CUB200 labels 0-indexed.

kdexd:
### Bug description:

CUB200 dataset in `torchvision.prototype.datasets` module formed labels using file paths. This resulted in labels being 1-indexed (1-200) instead of 0-indexed (0-199). A related issue from the past was with Flowers102 (`torchvision.datasets` module, #5766).

This PR simply shifts the labels and makes them zero-indexed for consistency with other datasets.

### DIFF
--- a/torchvision/prototype/datasets/_builtin/cub200.py
+++ b/torchvision/prototype/datasets/_builtin/cub200.py
@@ -177,7 +177,10 @@ class CUB200(Dataset):
         return dict(
             prepare_ann_fn(anns_data, image.image_size),
             image=image,
-            label=Label(int(pathlib.Path(path).parent.name.rsplit(".", 1)[0]), categories=self._categories),
+            label=Label(
+                int(pathlib.Path(path).parent.name.rsplit(".", 1)[0]) - 1,
+                categories=self._categories,
+            ),
         )
 
     def _datapipe(self, resource_dps: List[IterDataPipe]) -> IterDataPipe[Dict[str, Any]]:


pmeier (approved):
Thanks @kdexd!

kdexd:
The pleasure is mine! On a related note, my ongoing project requires evaluating large vision models on a suite of ~25 image classification datasets, specifically a big subset of datasets used to evaluate CLIP models (see Figure below).

image

Some datasets are not yet implemented in `torchvision.prototype.datasets`, but are present in `torchvision.datasets`. In my development code I simply added a wrapper class to have a consistent API with new datasets (`IterDataPipe`), like so:

```python
class TorchvisionWrapIterDataPipe(dpi.IterDataPipe):
    def __init__(self, name: str, root: str | Path, split: str, **kwargs):
        # Get the dataset from `torchvision.datasets` module.
        DatasetClass = getattr(torchvision.datasets, name)
        self._inner = DatasetClass(str(root), split, download=True, **kwargs)

        # Wrap the dataset as a MapDataPipe, then convert to iterable.
        _dp = MapToIterConverter(SequenceWrapper(self._inner))
        self._dp = hint_sharding(hint_shuffling(_dp))

    def __len__(self):
        return len(self._inner)

    def __iter__(self):
        for image, label in self._dp:
            yield {"image": image, "label": label}


FGVCAircraft = partial(TorchvisionWrapIterDataPipe, "FGVCAircraft")
Flowers102 = partial(TorchvisionWrapIterDataPipe, "Flowers102")
STL10 = partial(TorchvisionWrapIterDataPipe, "STL10", folds=0)
RenderedSST2 = partial(TorchvisionWrapIterDataPipe, "RenderedSST2")
```

Some other datasets like `RESISC45` and `HatefulMemes` are not present in either modules at all. I believe that these datasets will eventually be commonly used to evaluate CLIP-like models and large SSL models. Torchvision users would find these useful if they are ready for evaluation out-of-the-box. If there is sufficient interest, would you accept pull requests for these datasets? In that case I will actually re-implement a consistent API for these datasets rather than a quick wrapper like shown above.

pmeier:
> If there is sufficient interest, would you accept pull requests for these datasets? In that case I will actually re-implement a consistent API for these datasets rather than a quick wrapper like shown above.

Thanks a lot for the interest in contributing. Could you open an issue so this won't get lost? As of now, our new API is not finalized and thus it makes little sense to port more datasets now. However, we are working hard to get this done and would notify you when we can start adding new datasets.

PR #6703, opened 2022-10-04
fix bug in output format for pyav (#6672)

jdsgomes:
* fix bug in output format for videos in `read_video`
cherrypicked: https://github.com/pytorch/vision/pull/6672


YosuaMichael (approved):
LGTM!

PR #6704, opened 2022-10-04
[ONNX] Rephrase ONNX RoiAlign warning for aligned=True

BowenBao:
Rephrase warning message when exporting `RoiAlign` in ONNX opset < 16 with `aligned=True`.


datumbox (approved):
Thanks for the correction.

PR #6705, opened 2022-10-05
[WIP] Clearup from memory explicitly during test

YosuaMichael:
During the CI test, we sometimes found that the model test get OOM. This happen rather intermittenly, and usually the model that get error should be able to fit into the memory, hence it indicate there are some left over of the model still occupying memory.

In this PR, we attempt to explicitly remove the models and its input-output from the memory.

We should try to check the CI on this PR if the problem arise.

### DIFF
--- a/test/test_models.py
+++ b/test/test_models.py
@@ -1,5 +1,6 @@
 import contextlib
 import functools
+import gc
 import operator
 import os
 import pkgutil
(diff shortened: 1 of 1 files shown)


test/test_models.py
```diff
@@ -697,6 +698,11 @@ def test_classification_model(model_fn, dev):
 
     _check_input_backprop(model, x)
 
+    # Clearing from memory explicitly
+    del model, x, out
+    gc.collect()
+    torch.cuda.empty_cache()
```
YosuaMichael, line 704: Note: I tried putting this functionality into a function in `common_utils.py` but I realized that calling `del` inside another function doesn't really works. For illustration look at the following script:
```
def mydelete(xs):
  for x in xs:
    del x

obj_1 = "hello"
obj_2 = [1,2,3]

mydelete([obj_1, obj_2])
print(obj_1, obj_2) # This still works and the objects is not deleted properly!

del obj_1, obj_2
print(obj_1, obj_2) # This will trigger error
```

PR #6708, opened 2022-10-05
make pytest summary more concise

pmeier:
We are currently using 

https://github.com/pytorch/vision/blob/a46c4f0ccdb67d94c2ffc8b68b52693533a7683c/pytest.ini#L3-L4

which shows every test that is not passing in the summary. Meaning, skips and xfails are also shown although I'd would wager a guess that no one looks at them. Thus, this PR changes this to only show failures, errors, and xpasses.

Without this, adding markers to the parametrization will render the summary almost useless, since they all get a distinct mention:

```py
import pytest


@pytest.mark.xfail
def test_foo():
    assert False


@pytest.mark.parametrize("baz", [pytest.param(i, marks=[pytest.mark.xfail]) for i in range(10)])
def test_bar(baz):
    assert False
```

- `-ra`: 

  ```
  ======================================== short test summary info =========================================
  XFAIL main.py::test_foo
  XFAIL main.py::test_bar[0]
  XFAIL main.py::test_bar[1]
  XFAIL main.py::test_bar[2]
  XFAIL main.py::test_bar[3]
  XFAIL main.py::test_bar[4]
  XFAIL main.py::test_bar[5]
  XFAIL main.py::test_bar[6]
  XFAIL main.py::test_bar[7]
  XFAIL main.py::test_bar[8]
  XFAIL main.py::test_bar[9]
  ========================================== 11 xfailed in 0.73s ===========================================
  ```
- `-rfEX`: 

  ```
  ========================================== 11 xfailed in 0.72s ===========================================
  ```

AFAIK, there is currently no occurrence of such pattern in our test suite yet, but will be soon #6653.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,7 +1,7 @@
 [pytest]
 addopts =
     # show summary of all tests that did not pass
-    -ra
+    -rfEX
     # Make tracebacks shorter
     --tb=native
     # enable all warnings
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,6 +1,6 @@
 [pytest]
 addopts =
-    # show summary of all tests that did not pass
+    # show tests that (f)ailed, (E)rror, or (X)passed in the summary
     -rfEX
     # Make tracebacks shorter
     --tb=native


pytest.ini
```diff
@@ -1,7 +1,7 @@
 [pytest]
 addopts =
     # show summary of all tests that did not pass
```
YosuaMichael, line 3: minor NIT, lets update the comment to reflect what we do here

YosuaMichael (approved):
Thanks @pmeier , the change looks good and I agree normally we don't need summary for those that are skipped or expected to fail.

I just have a minor NIT regarding the comment. I will approve first to unblock.

PR #6712, opened 2022-10-06
Avoid recommuting the affine matrix in bbox rotate

datumbox:
Removing the TODO we had for eliminating the re-estimation of the rotation matrix.

### DIFF
--- a/torchvision/prototype/transforms/functional/_geometry.py
+++ b/torchvision/prototype/transforms/functional/_geometry.py
@@ -334,11 +334,15 @@ def _affine_bounding_box_xyxy(
     device = bounding_box.device
 
     affine_vector = _get_inverse_affine_matrix(center, angle, translate, scale, shear, inverted=False)
-    transposed_affine_matrix = torch.tensor(
-        affine_vector,
-        dtype=dtype,
-        device=device,
-    ).view(2, 3).T
+    transposed_affine_matrix = (
+        torch.tensor(
+            affine_vector,
+            dtype=dtype,
+            device=device,
+        )
+        .view(2, 3)
+        .T
+    )
     # 1) Let's transform bboxes into a tensor of 4 points (top-left, top-right, bottom-left, bottom-right corners).
     # Tensor of points has shape (N * 4, 3), where N is the number of bboxes
     # Single point structure is similar to
--- a/torchvision/prototype/transforms/functional/_geometry.py
+++ b/torchvision/prototype/transforms/functional/_geometry.py
@@ -279,9 +279,9 @@ def affine_image_tensor(
     center_f = [0.0, 0.0]
     if center is not None:
         # Center values should be in pixel coordinates but translated such that (0, 0) corresponds to image center.
-        center_f = [1.0 * (c - s * 0.5) for c, s in zip(center, [width, height])]
+        center_f = [(c - s * 0.5) for c, s in zip(center, [width, height])]
 
-    translate_f = [1.0 * t for t in translate]
+    translate_f = [float(t) for t in translate]
     matrix = _get_inverse_affine_matrix(center_f, angle, translate_f, scale, shear)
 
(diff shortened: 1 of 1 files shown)


torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -569,14 +573,6 @@ def rotate_bounding_box(
         expand=expand,
     )
 
-    if expand:
-        # TODO: Move this computation inside of `_affine_bounding_box_xyxy` to avoid computing the rotation and points
-        #  matrix twice
-        height, width = image_size
-        rotation_matrix = _get_inverse_affine_matrix([0.0, 0.0], angle, [0.0, 0.0], 1.0, [0.0, 0.0])
```
datumbox, line 576: Here we passed center=[0, 0] instead of the value of `center`. Was that a bug? None of the tests captured a difference locally. This might indicate a gap on our testing strategy. cc @pmeier

datumbox: Also the value of `inverted` was True, while on the actual kernel we pass explicitly `False`.

vfdev-5: Center and expand are orthogonal, if expand is True, center should be default = in the image center.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -333,19 +333,20 @@ def _affine_bounding_box_xyxy(
     dtype = bounding_box.dtype if torch.is_floating_point(bounding_box) else torch.float32
     device = bounding_box.device
 
-    affine_matrix = torch.tensor(
-        _get_inverse_affine_matrix(center, angle, translate, scale, shear, inverted=False),
+    affine_vector = _get_inverse_affine_matrix(center, angle, translate, scale, shear, inverted=False)
```
datumbox, line 336: Are we sure the right value for this is `inverted=False`? If that's the case, we will need another call with `inverted=True` for the image.

torchvision/prototype/transforms/functional/_geometry.py
datumbox: Should we explicitly enforce and align this in the `_affine_bounding_box_xyxy` method?

datumbox: OK sounds like there is no issue with `center`. How about the value of the `inverted`?

torchvision/prototype/transforms/functional/_geometry.py
datumbox: It seems the tests are failing and it's probably related to the [inverted](https://github.com/pytorch/vision/actions/runs/3196650769/jobs/5218862358).

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -377,8 +379,12 @@ def _affine_bounding_box_xyxy(
         # Translate bounding boxes
         out_bboxes[:, 0::2] = out_bboxes[:, 0::2] - tr[:, 0]
         out_bboxes[:, 1::2] = out_bboxes[:, 1::2] - tr[:, 1]
-        # Estimate meta-data for image with inverted=True
-        affine_vector = _get_inverse_affine_matrix(center, angle, translate, scale, shear)
+        # Estimate meta-data for image with inverted=True and with the original center
+        center_f = [0.0, 0.0]
+        if center is not None:
+            # Center values should be in pixel coordinates but translated such that (0, 0) corresponds to image center.
+            center_f = [1.0 * (c - s * 0.5) for c, s in zip(center, [width, height])]
```
vfdev-5, line 386: If center is defined it shouldn't have any effect if expand is True.

vfdev-5 (approved):
Thanks @datumbox , lgtm!

PR #6718, opened 2022-10-07
replace new_like with wrap_like

pmeier:
*Throughout this comment I'm using `Image` as proxy for all of our features for simplicity*

--- 

This is my take on reducing the overhead transforms v2 has. Currently, we are using the idiom 

https://github.com/pytorch/vision/blob/7eb5d7fcab73afec976907a855d9e63fa31f5579/torchvision/prototype/features/_image.py#L136

everywhere to wrap a plain tensor into the image feature. Doing so results in multiple `__torch_function__` calls as detailed in #6681. Similar to the constructor, the `Image.new_like` method accepts arbitrary `data: Any` as input and thus has to go through the constructor every time.

However, we never call it without a tensor input. Plus, whenever we pass `dtype` to `Image.new_like`, it is not to change the `dtype` of the tensor to be wrapped, but rather to retain it:

https://github.com/pytorch/vision/blob/7eb5d7fcab73afec976907a855d9e63fa31f5579/torchvision/prototype/features/_bounding_box.py#L158

Taking this one step further, this also means that the `new_like` name is somewhat misleading. Yes, one gets a new `features.Image` object, but unlike the `torch.*_like` methods, we don't get a new storage unless the `dtype` or `device` is changed.

This PR proposed to fix the above by refactoring `Image.new_like` to `Image.wrap_like`. Opposed to `new_like`, `wrap_like` only takes a tensor to be wrapped as well the metadata for the specific type, i.e. `color_space` for `features.Image`. This prevents the need to go through the constructor and results in no `__torch_function__` calls at all:

```py
import unittest.mock

import torch

image = features.Image(torch.rand(3, 16, 16))

with unittest.mock.patch(
    "torchvision.prototype.features._feature._Feature.__torch_function__", side_effect=AssertionError
):
    # This has to be `.new_like` on `main` and `.wrap_like` on the PR
    features.Image.wrap_like(image, torch.rand(3, 16, 16))
```

We can estimate the impact of this change on one classification training:

```py

import torch

# @datumbox, @vfdev-5: please let me know if these assumptions don't reflect reality
# This comes from @vfdev-5's benchmarks
num_calls_per_sample = 20
# number of samples in imagenet training set
num_samples_per_epoch = 1_200_000
num_epochs = 600
# the wrapping happens in the transforms pipeline, i.e. on each worker individually. 
Thus, each worker only has a fraction of samples to process
num_processes = 8

input = features.Image(torch.rand(3, 512, 512))
output = torch.rand(3, 512, 512)


time_diffs = []
for _ in range(1000):
    time_diff_per_sample = 0
    for _ in range(num_calls_per_sample):
        start = perf_counter_ns()
        # This has to be `.new_like` on `main` and `.wrap_like` on the PR
        features.Image.wrap_like(input, output)
        stop = perf_counter_ns()
        time_diff_per_sample += stop - start
    time_diffs.append(time_diff_per_sample)

overhead_per_sample = float(torch.tensor(time_diffs).to(torch.float64).median()) * 1e-9
print(f"Overhead per sample: {overhead_per_sample*1e6:5.1f} µs")

estimated_overhead_per_training = overhead_per_sample * num_samples_per_epoch * num_epochs / num_processes
print(f"Estimated overhead per training: {estimated_overhead_per_training / 60 / 60:.1f} h")
```

```
Overhead per sample:  21.2 µs
Estimated overhead per training: 0.5 h
```

Although the overhead is quite low with roughly 20 µs per sample or 1 µs per call, the enormous amount of calls during a full training blows this up to a significant increase. However, running the same benchmark on `main` yields

```
Overhead per sample: 208.3 µs
Estimated overhead per training: 5.2 h
```

To put it in words, this PR achieves roughly a 10x reduction of the overhead. And the only thing we lose is the ability to pass arbitrary data to the wrapping function or change the `dtype` and `device` in the process. Both of which we don't do and I currently don't see a use case for it either. 

Note that this doesn't affect the ability to pass arbitrary data to the constructor. This is still supported. Plus, in contrast to the proposed `wrap_like` function, the constructor may also process the metadata, like guessing the `color_space` if none is passed to `features.Image`, while `wrap_like` only takes the correct type or takes the value from the reference.

### DIFF
--- a/torchvision/prototype/features/_bounding_box.py
+++ b/torchvision/prototype/features/_bounding_box.py
@@ -19,6 +19,13 @@ class BoundingBox(_Feature):
     format: BoundingBoxFormat
     image_size: Tuple[int, int]
 
+    @classmethod
+    def _wrap(cls, tensor: torch.Tensor, *, format: BoundingBoxFormat, image_size: Tuple[int, int]) -> BoundingBox:
+        bounding_box = tensor.as_subclass(cls)
+        bounding_box.format = format
+        bounding_box.image_size = image_size
+        return bounding_box
+
     def __new__(
         cls,
         data: Any,
@@ -29,52 +36,46 @@ class BoundingBox(_Feature):
         device: Optional[Union[torch.device, str, int]] = None,
         requires_grad: bool = False,
     ) -> BoundingBox:
-        bounding_box = super().__new__(cls, data, dtype=dtype, device=device, requires_grad=requires_grad)
+        tensor = cls._to_tensor(data, dtype=dtype, device=device, requires_grad=requires_grad)
 
         if isinstance(format, str):
             format = BoundingBoxFormat.from_str(format.upper())
-        bounding_box.format = format
 
-        bounding_box.image_size = image_size
+        return cls._wrap(tensor, format=format, image_size=image_size)
 
-        return bounding_box
+    @classmethod
+    def wrap_like(
+        cls,
+        other: BoundingBox,
+        tensor: torch.Tensor,
+        *,
+        format: Optional[BoundingBoxFormat] = None,
+        image_size: Optional[Tuple[int, int]] = None,
+    ) -> BoundingBox:
+        return cls._wrap(
+            tensor,
+            format=format if format is not None else other.format,
+            image_size=image_size if image_size is not None else other.image_size,
+        )
 
     def __repr__(self, *, tensor_contents: Any = None) -> str:  # type: ignore[override]
         return self._make_repr(format=self.format, image_size=self.image_size)
-
-    @classmethod
-    def new_like(
-        cls,
-        other: BoundingBox,
-        data: Any,
-        *,
-        format: Optional[Union[BoundingBoxFormat, str]] = None,
-        image_size: Optional[Tuple[int, int]] = None,
-        **kwargs: Any,
-    ) -> BoundingBox:
-        return super().new_like(
-            other,
-            data,
-            format=format if format is not None else other.format,
-            image_size=image_size if image_size is not None else other.image_size,
-            **kwargs,
-        )
 
     def to_format(self, format: Union[str, BoundingBoxFormat]) -> BoundingBox:
         if isinstance(format, str):
             format = BoundingBoxFormat.from_str(format.upper())
 
-        return BoundingBox.new_like(
+        return BoundingBox.wrap_like(
             self, self._F.convert_format_bounding_box(self, old_format=self.format, new_format=format), format=format
         )
 
     def horizontal_flip(self) -> BoundingBox:
         output = self._F.horizontal_flip_bounding_box(self, format=self.format, image_size=self.image_size)
-        return BoundingBox.new_like(self, output)
+        return BoundingBox.wrap_like(self, output)
 
     def vertical_flip(self) -> BoundingBox:
         output = self._F.vertical_flip_bounding_box(self, format=self.format, image_size=self.image_size)
-        return BoundingBox.new_like(self, output)
+        return BoundingBox.wrap_like(self, output)
 
     def resize(  # type: ignore[override]
         self,
@@ -84,19 +85,19 @@ class BoundingBox(_Feature):
         antialias: bool = False,
     ) -> BoundingBox:
         output, image_size = self._F.resize_bounding_box(self, image_size=self.image_size, size=size, max_size=max_size)
-        return BoundingBox.new_like(self, output, image_size=image_size)
+        return BoundingBox.wrap_like(self, output, image_size=image_size)
 
     def crop(self, top: int, left: int, height: int, width: int) -> BoundingBox:
         output, image_size = self._F.crop_bounding_box(
             self, self.format, top=top, left=left, height=height, width=width
         )
-        return BoundingBox.new_like(self, output, image_size=image_size)
+        return BoundingBox.wrap_like(self, output, image_size=image_size)
 
     def center_crop(self, output_size: List[int]) -> BoundingBox:
         output, image_size = self._F.center_crop_bounding_box(
             self, format=self.format, image_size=self.image_size, output_size=output_size
         )
-        return BoundingBox.new_like(self, output, image_size=image_size)
+        return BoundingBox.wrap_like(self, output, image_size=image_size)
 
     def resized_crop(
         self,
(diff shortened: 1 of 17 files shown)


torchvision/prototype/features/_image.py
```diff
@@ -71,36 +77,33 @@ def __new__(
         device: Optional[Union[torch.device, str, int]] = None,
         requires_grad: bool = False,
     ) -> Image:
-        data = torch.as_tensor(data, dtype=dtype, device=device)
-        if data.ndim < 2:
+        tensor = cls._to_tensor(data, dtype=dtype, device=device, requires_grad=requires_grad)
```
vfdev-5, line 80: What's `_to_tensor` ?

torchvision/prototype/features/_bounding_box.py
```diff
@@ -19,6 +19,13 @@ class BoundingBox(_Feature):
     format: BoundingBoxFormat
     image_size: Tuple[int, int]
 
+    @classmethod
+    def _wrap(cls, tensor: torch.Tensor, *, format: BoundingBoxFormat, image_size: Tuple[int, int]) -> BoundingBox:
```
datumbox, line 23: Nit: Do we need another private method across all classes? Why not add this directly to `wrap_like` and call that method when necessary (for example from the `__new__`).

torchvision/prototype/transforms/_meta.py
```diff
@@ -30,7 +30,9 @@ def __init__(self, dtype: torch.dtype = torch.float32) -> None:
 
     def _transform(self, inpt: features.TensorImageType, params: Dict[str, Any]) -> features.TensorImageType:
         output = F.convert_image_dtype(inpt, dtype=self.dtype)
-        return output if features.is_simple_tensor(inpt) else features.Image.new_like(inpt, output, dtype=self.dtype)  # type: ignore[arg-type]
+        return (
+            output if features.is_simple_tensor(inpt) else features.Image.wrap_like(cast(features.Image, inpt), output)
```
datumbox, line 34: Nit: Why did we switched from ignore to cast here?

datumbox (approved):
LGTM, thanks @pmeier! I only have a couple of nit comments but nothing major.

@vfdev-5 my understanding is that it's still worth proceeding with some of the ideas from #6681 to reduce the `__torch_function__` calls. Could you confirm that this PR doesn't affect the approach you are favouring to solve this?

torchvision/prototype/features/_mask.py
```diff
@@ -9,13 +9,36 @@
...
+        tensor = cls._to_tensor(data, dtype=dtype, device=device, requires_grad=requires_grad)
+        return cls._wrap(tensor)
+
+    @classmethod
+    def wrap_like(
+        cls,
+        other: Mask,
+        tensor: torch.Tensor,
+    ) -> Mask:
+        return cls._wrap(tensor)
```
vfdev-5, line 33: Why can't we just call `tensor.as_subclass(cls)` here ? I'm not sure to understand the point of introducing `_wrap` function.

torchvision/prototype/features/_image.py
```diff
@@ -71,36 +77,33 @@ def __new__(
...
-        return self._make_repr(color_space=self.color_space)
+        return cls._wrap(tensor, color_space=color_space)
 
     @classmethod
-    def new_like(
-        cls, other: Image, data: Any, *, color_space: Optional[Union[ColorSpace, str]] = None, **kwargs: Any
-    ) -> Image:
-        return super().new_like(
-            other, data, color_space=color_space if color_space is not None else other.color_space, **kwargs
+    def wrap_like(cls, other: Image, tensor: torch.Tensor, *, color_space: Optional[ColorSpace] = None) -> Image:
```
vfdev-5, line 98: This signature seems strange. IMO, we should not override attributes defined by `other`. Let's remove color_space arg. Same for other `wrap_like` methods.

torchvision/prototype/features/_bounding_box.py
pmeier: > Why not add this directly to `wrap_like` and call that method when necessary (for example from the `__new__`).

As it stand, `wrap_like` needs a reference instance of that class. Since we don't have that inside the constructor, I've factored out the common parts into `_wrap`. 

If you are worried about the performance implications, they are minimal. Using `features.Image._wrap(output, color_space=input.color_space)` yields

```py
Overhead per sample:  18.8 µs
Estimated overhead per training: 0.47 h
```

This is the best we can do, i.e. we pass everything explicitly. The difference is roughly 100 ns per call or roughly 3 minutes for the whole training. If performance was the only concern, I would keep both methods for convenience and parity between the classes.

torchvision/prototype/transforms/_meta.py
pmeier: There were a bunch of `mypy` errors and it seems I was overly ambitious. Will revert.

torchvision/prototype/features/_mask.py
pmeier: You are right, if you only look at `Mask` it makes little sense. However, not having it would diverge the implementation from the other features. In https://github.com/pytorch/vision/pull/6718#discussion_r989999849 I've shown that it basically has no runtime implications and thus I would favor consistency over micro optimization, i.e. a single method call.

torchvision/prototype/features/_image.py
pmeier: > IMO, we should not override attributes defined by `other`

That is the whole point of this method. This should not only wrap the tensor, but also provide a convenient interface to change the metadata. If we wouldn't do that, why would we have the `other` reference in the first place? This is on par with the `torch.*_like` methods that let you override the `dtype` or `device` during creation:

```pycon
>>> t = torch.rand(3, dtype=torch.float64)
>>> torch.rand_like(t, dtype=torch.float16)
tensor([0.4985, 0.7021, 0.7393], dtype=torch.float16)
```

What you seem to be proposing here is to have a "bare metal" function. That is what `_wrap` is currently doing. We could just make it public.

datumbox: I think changing the signature is worth it. I missed it on the original review and I must agree with Victor this looks a bit weird.

pmeier: > I think changing the signature is worth it.

Changing the signature to what? If we remove all the metadata overwrites how are we going to do something like 

https://github.com/pytorch/vision/blob/7eb5d7fcab73afec976907a855d9e63fa31f5579/torchvision/prototype/features/_bounding_box.py#L87

where we need to retain the `format` of the current bounding box, but need to change the `image_size`?

vfdev-5 (approved):
lgtm, let's move on !
Thanks @pmeier

torchvision/prototype/features/_image.py
datumbox: As discuss offline, we should either remove it from everywhere or from none. We decided offline to keep as-is.

pmeier:
As discussed offline, there are multiple ways we can approach the interface design. Nothing is set in stone here. We'll move with what I have proposed with the strong possibility of refactoring later. The performance gain is too high to hold this up with bike shedding.

PR #6721, opened 2022-10-07
NEW Feature: Mixup transform for Object Detection

### LINKED ISSUE #6720: New Feature: Mixup Transform for Object Detection
### 🚀 The feature

Follow up to #6323 

Add a new transform: Mixup to the transforms in torchvision
For reference: [Mixup for Detection](https://arxiv.org/pdf/1902.04103v1.pdf) [[1](https://github.com/Megvii-BaseDetection/YOLOX/blob/a5f629a6d28fcc3742ce9483698b3376ce457533/yolox/data/datasets/mosaicdetection.py#L162-L234), [2](https://github.com/open-mmlab/mmdetection/blob/1376e77e6ecbaad609f6003725158de24ed42e84/mmdet/datasets/pipelines/transforms.py#L2347)]

### Motivation, pitch

Keeping torchvision updated with the latest SotA techniques

### Alternatives

_No response_

### Additional context

_No response_

ambujpawar:
Closes #6720 

Draft PR for now

### DIFF
--- a/torchvision/prototype/transforms/_mixup.py
+++ b/torchvision/prototype/transforms/_mixup.py
@@ -1,0 +1,3 @@
+# TODO: Consturct a Mixup transform
+
+class
--- a/torchvision/prototype/transforms/_mixup.py
+++ b/torchvision/prototype/transforms/_mixup.py
@@ -1,3 +1,4 @@
 # TODO: Consturct a Mixup transform
 
-class
+class _Mixup:
+    pass


pmeier:
Hey @ambujpawar and thanks a lot for the PR! I'll try to help you land it in the near future. As you might have noticed, this transform is not straight forward to implement since it requires a batch of detection samples. In this context this means a list of samples, whereas for classification "batch" usually means an extra batch dimension on a tensor. This makes this implementation a lot harder compared to regular `MixUp`. 

Still, we need to be able to support it. I'll look into how we can streamline the process for example by providing a `_DetectionBatchTransform` or standalone utilities that makes this easier. I'll get back to you when I found a solution or need your input. Is that ok with you?

pmeier:
> However, shouldn't it be similar to what we do in CopyPaste transform? Because in copyPaste we also expect a batch of images

Exactly. Before we operated under the assumption that `SimpleCopyPaste` is the only batch detection transform and thus a one-off solution for it was good enough. With `DetectionMixUp` in the picture our assumption is no longer true and we need to look how we can provide utilities to ease the implementation of these transforms. 

Right now the largest part of the implementation deals with the "infrastructure", i.e. extracting the right inputs and putting them back afterwards. Only a small part is spent on the actual algorithm. In a best case scenario, I find a solution so you can only write the algorithm and the remainder is handled by a base class or some high level utilities.

ambujpawar:
That clears up all the questions for me. Thanks!
 
Yeah, a base class is perhaps the best solution in those regards.

pmeier (commented):
@ambujpawar I took the liberty of pushing a patch to your PR. I've added two functions `flatten_and_extract` as well as `unflatten_and_insert` that implement what their name implies. I'm actively looking for your feedback so nothing is fixed yet. Two things that I already noticed:

1. Both `SimpleCopyPaste` as well `MixUpDetection` use a "split" layout for images and targets. Is that by design or could we use one container like a dictionary for both of them. Imagine something like `sample = {"image": ..., "boxes": ...}`.
2. The old extraction and insertion logic converted to tensor and back for images and (un-)wrapped the other features. Right now, the new logic does not do this. Instead this is moved inside the `_mixup` function. We could move that back into the logic as well. What do you prefer?

ambujpawar:
Thanks for adding the patch! :)

I think it looks nice. Regarding your questions:

> 1. Both SimpleCopyPaste as well MixUpDetection use a "split" layout for images and targets. Is that by design or could we use one container like a dictionary for both of them. Imagine something like sample = {"image": ..., "boxes": ...}.
Yes, they both use a "split" design, but Mixup Detection but Mixup does not use "Masks". Mixup is only used for Detection not Segmentation. 

> The old extraction and insertion logic converted to tensor and back for images and (un-)wrapped the other features. Right now, the new logic does not do this. Instead this is moved inside the _mixup function. We could move that back into the logic as well. What do you prefer?
If I had to choose one design, I would chose the former design but I dont have any strong arguments for it.

Shall we also include the developers of SimpleCopyPaste transform as well. They might also have some comments regarding it

pmeier (commented):
@ambujpawar I've refactored the helpers again. They are quite complicated now, but at least they make your implementation simpler. We can always refactor them to make it easier in the future. PTAL.

> Shall we also include the developers of SimpleCopyPaste transform as well?

Nope. We are only changing the infrastructure here. We only need to bring them in if something about the algorithm is unclear. I haven't reviewed that yet. Will send a review soon.

ambujpawar (commented):
This looks really neat and thanks for refactoring it :)

ambujpawar:
[Just a question] why did we have to do changes to "torchvision/prototype/datasets/_builtin/coco.py"? I did not get that...


So, I will just have to fix the tests and we are done?

torchvision/prototype/transforms/__init__.py
```diff
@@ -51,6 +51,7 @@
     ToDtype,
     TransposeDimensions,
 )
+from ._mixup_detection import MixupDetection
```
datumbox, line 54: Should we move this inside `transforms/_augment.py` along with the rest of MixUp/CutMix etc?

torchvision/prototype/transforms/_augment.py
```diff
@@ -190,6 +190,90 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
             return inpt
 
 
+def flatten_and_extract_data(
```
datumbox, line 193: Perhaps it's too early but worth benchmarking that this doesn't introduce a significant speed regression on SimpleCopyPaste.

torchvision/prototype/transforms/_augment.py
```diff
@@ -190,6 +190,90 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
...
+    inputs: Any, **target_types_or_checks: Tuple[Union[Type, Callable[[Any], bool]], ...]
+) -> Tuple[Tuple[List[Any], TreeSpec, List[Dict[str, int]]], List[features.TensorImageType], List[Dict[str, Any]]]:
+    # Images are special in the sense that they will always be extracted and returned
+    # separately. Internally however, they behave just as the other features.
+    types_or_checks: Dict[str, Tuple[Union[Type, Callable[[Any], bool]], ...]] = {
+        "images": (features.Image, PIL.Image.Image, features.is_simple_tensor),
```
datumbox, line 199: Should we avoid defaults and just explicitly pass images in the 2 cases? This might be useful if on the future we extended this for video. No strong opinions.

torchvision/prototype/transforms/_mixup_detection.py
```diff
@@ -0,0 +1,114 @@
...
+        if has_any(flat_inputs, features.Mask):
+            raise TypeError(f"Masks are not supported by {type(self).__name__}()")
+
+        if not has_any(flat_inputs, PIL.Image.Image, features.Image, features.is_simple_tensor):
+            raise TypeError(
+                f"{type(self).__name__}() requires input sample to contain an tensor or PIL image or a Video."
+            )
+
+        if not (
+            has_any(flat_inputs, features.Image, PIL.Image.Image, features.is_simple_tensor)
```
datumbox, line 42: We checked this already above. Can we simplify/cache to avoid repeated checks?

torchvision/prototype/transforms/_mixup_detection.py
```diff
@@ -0,0 +1,114 @@
...
+    def __init__(
+        self,
+        *,
+        alpha: float = 1.5,
+    ) -> None:
+        super().__init__()
+        self._dist = torch.distributions.Beta(torch.tensor([alpha]), torch.tensor([alpha]))
+
+    def _check_inputs(self, flat_inputs: List[Any]) -> None:
+        if has_any(flat_inputs, features.Mask):
```
datumbox, line 33: How about other unsupported types like Video?

datumbox (commented):
Looking good, so far. I've added a few early comments but since the PR is in development feel free to ignore.

ambujpawar:
Hi @pmeier , is there anything I can do to speed up the merging of this PR to master?

pmeier:
Sorry, for keep you waiting. Unfortunately, I'm caught up in something else until end of the week. 

Things you can do until then:

- move your implementation into the `_augment` module as suggested in https://github.com/pytorch/vision/pull/6721#discussion_r1017685091
- refactor the `_check_inputs` per https://github.com/pytorch/vision/pull/6721#discussion_r1017693484 and https://github.com/pytorch/vision/pull/6721#discussion_r1017693719
- improve the error messages that I either left out or just added dummys like https://github.com/pytorch/vision/pull/6721#discussion_r1016715204
- Think again whether or not we need to special case images in the whole logic. My initial version didn't do it and @datumbox also asked about it in https://github.com/pytorch/vision/pull/6721#discussion_r1017690068.

torchvision/prototype/transforms/_mixup_detection.py
ambujpawar: Resolved. Since, the current implementation doesn't support video yet. I am simply raising a TypeError if we encounter a video

ambujpawar:
Hi @pmeier , Sorry I was a bit stuck with something but now happy to be back to contribute.
I have tried to resolve the first 3 topics you mentioned in your previous comment. The tests are still breaking and I will take a look at it again. 
In the meantime, PTAL at the PR.

One request from my side, its still not clear to me what should we regarding the point you mentioned over here 

> Think again whether or not we need to special case images in the whole logic. My initial version didn't do it and @datumbox also asked about it in https://github.com/pytorch/vision/pull/6721#discussion_r1017690068.

 and [this comment](https://github.com/pytorch/vision/pull/6721#discussion_r1017690068). I dont have any strong preference for either case. I am happy to implement whatever is easier to implement and maintain :)

torchvision/prototype/transforms/_augment.py
```diff
@@ -380,7 +409,95 @@ def forward(self, *inputs: Any) -> Any:
...
+        c_1, h_1, w_1 = image_1.shape
+        c_2, h_2, w_2 = image_2.shape
+        h_mixup = max(h_1, h_2)
+        w_mixup = max(w_1, w_2)
+
+        if mixup_ratio >= 1.0:
+            return image_1, target_1
+
+        # mixup images and prevent the object aspect ratio from changing
+        mix_img = torch.zeros(c_1, h_mixup, w_mixup, dtype=torch.float32)
```
pmeier, line 486: Shouldn't we have a `fill` parameter here on the transform like other transform that add regions without any original data? `0` is a good default, but we should probably give users the option to use something else. You can have a look at `Pad` for example: https://github.com/pytorch/vision/blob/main/torchvision/prototype/transforms/_geometry.py#L222-L226

torchvision/prototype/transforms/_augment.py
```diff
@@ -380,7 +409,95 @@ def forward(self, *inputs: Any) -> Any:
...
+        mix_img[:, : image_1.shape[1], : image_1.shape[2]] = image_1 * mixup_ratio
+        mix_img[:, : image_2.shape[1], : image_2.shape[2]] += image_2 * (1.0 - mixup_ratio)
```
pmeier, line 488: Can we use the extracted values here to make it less verbose?

```suggestion
        mix_img[:, : h_1, : w_1] = image_1 * mixup_ratio
        mix_img[:, : h_2, : w_2] += image_2 * (1.0 - mixup_ratio)
```

torchvision/prototype/transforms/_augment.py
```diff
@@ -380,7 +409,95 @@ def forward(self, *inputs: Any) -> Any:
...
+        # mixup targets
+        mix_target = {**target_1, **target_2}
+        box_format = target_1["boxes"].format
+        mixed_boxes = {
+            "boxes": features.BoundingBox(
+                torch.vstack((target_1["boxes"], target_2["boxes"])),
```
pmeier, line 494: Why do we need to `vstack` here? This creates an extra dimension, doesn't it? Shouldn't it be

```suggestion
                torch.cat((target_1["boxes"], target_2["boxes"]), dim=-1),
```

like below for the labels?

pmeier (commented):
Hey @ambujpawar, I did another round. Don't worry about the tests for now. Let's get the implementation in shape first.

Regarding the concern that @datumbox and I have: we are currently special casing images in the extraction logic and we don't know if it is worth it at this point. I probably makes more sense to only do so before we call `_mixup` and leave the extraction function more generic. Let me send a patch to see if this is viable.

pmeier:
@ambujpawar I've refactored the extraction and insertion logic in ae9908b037f5c64334ddba5f61d378bcc699a458. This temporarily breaks `SimpleCopyPaste`, but that is fine as long as we are in the design phase. Let me know what you think. While doing this, I already fixed a few of the comments I had above. Please still have a look in case I misunderstood something.

ambujpawar:
Hi @pmeier, it looks good to me.
Thanks a lot for refactoring the code. It looks much neat and tidy now :)

Shall I now work on fixing tests and refactoring SimpleCopyPaste?

pmeier:
> Shall I now work on fixing tests and refactoring SimpleCopyPaste?

Yup, go ahead. Plus, you haven't addressed some of the comments above. Could you go through all open comments and mark the ones that have been addressed as resolved? That makes it much easier to assess at a glance what still needs some work and what is finished.

ambujpawar:
Sure! I'll work in on the weekend and update it :)

torchvision/prototype/transforms/_augment.py
```diff
@@ -190,6 +190,40 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
...
+    for idx, inpt in enumerate(flat_inputs):
+        for key, types_or_checks_ in types_or_checks.items():
+            if _isinstance(inpt, types_or_checks_):
+                inputs[key].append(inpt)
+                idcs[key].append(idx)
+                break
+
+    num_inputs = [len(inputs_) for inputs_ in inputs.values()]
+    if not all(num_inputs_ == num_inputs[0] for num_inputs_ in num_inputs[1:]):
+        raise TypeError("FIXME")
```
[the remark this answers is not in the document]
ambujpawar: Outdated

ambujpawar:
Sorry, for being a bit late with my part of the code.
But finally fixed up the tests :) (they were simple just had to remove extract_image_targets call)

Even though I could run the tests successfully. However, I could not replicate it as a user of the transform. For instance,
`
import PIL
from torchvision import io, utils
from torchvision.prototype import features, transforms as T
from torchvision.prototype import datapoints


# Defining and wrapping input to appropriate Tensor Subclasses
path = "/Users/ambujpawar/Desktop/Cat03.jpeg"
path2 = "/Users/ambujpawar/Desktop/dog_2.jpeg"

# img = features.Image(io.read_image(path), color_space=features.ColorSpace.RGB)
img = PIL.Image.open(path)
img2 = PIL.Image.open(path2)

bbox_1 = datapoints.BoundingBox(
    [[2, 0, 100, 100], [396, 92, 479, 241]],
    format=datapoints.BoundingBoxFormat.XYXY,
    spatial_size=F.get_spatial_size(img),
)
bbox_2 = datapoints.BoundingBox(
    [ [200, 100, 300, 300], [424, 38, 479, 250]],
    format=datapoints.BoundingBoxFormat.XYXY,
    spatial_size=F.get_spatial_size(img2),
)
label = datapoints.Label([59, 58])


# Defining and applying Transforms V2
trans = T.Compose(
    [
        T.MixupDetection(),
    ]
)

imgs = [img, img2]
bboxes = [bbox_1, bbox_2]
labels= [label, label]
sample = {"image": imgs, "boxes": bboxes, "label": labels}
imgs, bboxes, labels = trans(imgs, bboxes,labels)
`

results in 
`TypeError: Sample at index 0 in the batch is missing dict_keys(['boxes', 'labels'])`

This is based on the example listed in the blogpost rergarding Transforms v2.
How should we call this transform instead?

ambujpawar:
Figured out SimpleCopyPaste still doesn't work. Working on it

torchvision/prototype/transforms/_augment.py
```diff
@@ -380,7 +324,73 @@ def forward(self, *inputs: Any) -> Any:
...
+        h_mixup = max(h_1, h_2)
+        w_mixup = max(w_1, w_2)
+
+        # TODO: add the option to fill this with something else than 0
+        dtype = sample_1["image"].dtype if sample_1["image"].is_floating_point() else torch.float32
+        mix_image = F.pad_image_tensor(
+            sample_1["image"].to(dtype), padding=[0, 0, w_mixup - w_1, h_mixup - h_1], fill=None
+        ).mul_(ratio)
+        mix_image[..., :h_2, :w_2] = sample_2["image"] * (1.0 - ratio)
+        mix_image = mix_image.to(sample_1["image"])
```
pmeier, line 383: You commented this out and indeed there was a bug in my implementation. I fixed that so we can reinstate it?

torchvision/prototype/transforms/_transform.py
```diff
@@ -83,3 +86,85 @@ def forward(self, *inputs: Any) -> Any:
         ]
 
         return tree_unflatten(flat_outputs, spec)
+
+
+class _DetectionBatchTransform(Transform):
```
pmeier, line 91: I've opted to put this common functionality into a superclass that all other detection batch transforms can subclass from. Maybe we can even improve this further, but I'll get to that when I fix `SimpleCopyPaste`.

pmeier:
Hey @ambujpawar :wave: I hope you are all right. I wanted to check in on this PR. Are you planning on finishing it or should I take over?

ambujpawar:
Hi @pmeier, thanks for asking! I'm doing good :) just back from a super long christmas and new year vacation so did not have time to work on this PR. I would still like to work on if we are not running on a deadline or something.

I can work on it this weekend and can request you for re-review :)
Does that work with you?

pmeier:
No rush from my side. I thought I check on you after roughly one month of inactivity. In case you didn't plan to finish this, we still would like to have it and I would have taken over. This weekend sounds good.

ambujpawar:
Thanks! I'll update the PR this weekend then! :)

torchvision/prototype/transforms/_transform.py
ambujpawar: Ah Thanks! This is much better! 👍

torchvision/prototype/transforms/_augment.py
```diff
@@ -380,7 +324,73 @@ def forward(self, *inputs: Any) -> Any:
...
+    ) -> None:
+        super().__init__()
+        self._dist = torch.distributions.Beta(torch.tensor([alpha]), torch.tensor([alpha]))
+
+    def _check_inputs(self, flat_inputs: List[Any]) -> None:
+        if has_any(flat_inputs, datapoints.Mask, datapoints.Video):
+            raise TypeError(f"{type(self).__name__}() is only supported for images and bounding boxes.")
+
+    def _get_params(self, flat_inputs: List[Any]) -> Dict[str, Any]:
+        return dict(ratio=float(self._dist.sample()))
```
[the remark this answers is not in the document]
ambujpawar: I agree, it looks much tidier this way!
However, I have a question: we dont use flat_inputs? Shall we just remove it?

test/test_prototype_transforms.py
```diff
@@ -1924,3 +1868,49 @@ def test__transform(self, inpt):
         assert type(output) is type(inpt)
         assert output.shape[-4] == num_samples
         assert output.dtype == inpt.dtype
+
+
+class TestMixupDetection:
+    def create_fake_image(self, mocker, image_type):
```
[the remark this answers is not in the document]
ambujpawar: Removing it! :)

test/test_prototype_transforms.py
```diff
@@ -1436,63 +1437,6 @@ def create_fake_image(self, mocker, image_type):
             return PIL.Image.new("RGB", (32, 32), 123)
         return mocker.MagicMock(spec=image_type)
 
-    def test__extract_image_targets_assertion(self, mocker):
```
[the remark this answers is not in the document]
ambujpawar: I was using it to test `_extract_image_targets ` function. However, since we removed those functions I removed them from here as well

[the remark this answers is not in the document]
ambujpawar: Oh sorry! Realized this was for TestSimpleCopyPaste. Undoing the changes, sorry

ambujpawar:
Hi, I added the test cases for when the ratios for mixup are 0 and 1. However, I still think there is some bug when we are mixing the two images. I am not able to exactly point what causes it though. Perhaps after looking at the code something rings a bell for you.
Screenshot 2023-01-22 at 14 39 04
So, this is the expected output (or something similar). Notice the light background of cat in the background


However, after our latest changes it look like this. Notice the picture of cat is completely overwriting the picture of dog.
Screenshot 2023-01-22 at 14 42 38

I am not exactly sure but seems like something is going wrong when we are mixing images in `augment.py  Line 376-381`. I am not able to solve it but perhaps you can have a look at it please?

torchvision/prototype/transforms/_augment.py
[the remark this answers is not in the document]
pmeier: I would keep it for consistency with the other transformations. This is the basic protocol for all `Transform._get_params` calls. Although we call it manually here, there is some benefit by aligning it. Someone not familiar with this transform, but the others in general might trip over the fact that the parameter is not there.

pmeier:
> I am not exactly sure but I suspect something is going wrong when we are mixing images in `augment.py Line 376-381`. I am not able to solve it but perhaps you can have a look at it please?

Yup, the problem is that we replace the values in the first image with the ones from the second rather than adding them. To demonstrate, let's establish a visual benchmark first that we both can easily reproduce:

```python
import PIL.Image

import torch

from torchvision.io import read_image
from torchvision.prototype import datapoints, transforms
from torchvision.utils import make_grid


def read_sample(path, label):
    image = datapoints.Image(read_image(path))
    bounding_box = datapoints.BoundingBox(
        [[0, 0, *image.spatial_size[::-1]]], format="xyxy", spatial_size=image.spatial_size
    )
    label = datapoints.Label([label])
    return dict(
        path=path,
        image=image,
        bounding_box=bounding_box,
        label=label,
    )


batch = [
    read_sample("test/assets/encode_jpeg/grace_hopper_517x606.jpg", 0),
    read_sample("test/assets/fakedata/logos/rgb_pytorch.png", 1),
]

transform = transforms.MixupDetection()

torch.manual_seed(0)
output = transform(batch)

image = make_grid([sample["image"] for sample in output])
PIL.Image.fromarray(image.permute(1, 2, 0).numpy()).save("mixup_detection.jpg")
```

Output with the current implementation is

mixup_detection

So, in the left image, the PyTorch logo is the second image and thus we are just pasting it over Grace Hopper. On the right side the PyTorch logo is completely gone, since Grace Hopper is larger and thus completely paints over it.

Applying the first suggestion from below gives us

mixup_detection

And thus the behavior we want.

torchvision/prototype/transforms/_augment.py
```diff
@@ -380,7 +324,71 @@ def forward(self, *inputs: Any) -> Any:
...
+        h_2, w_2 = sample_2["image"].shape[-2:]
+        h_mixup = max(h_1, h_2)
+        w_mixup = max(w_1, w_2)
+
+        # TODO: add the option to fill this with something else than 0
+        dtype = sample_1["image"].dtype if sample_1["image"].is_floating_point() else torch.float32
+        mix_image = F.pad_image_tensor(
+            sample_1["image"].to(dtype), padding=[0, 0, w_mixup - w_1, h_mixup - h_1], fill=None
+        ).mul_(ratio)
+        mix_image[..., :h_2, :w_2] = sample_2["image"] * (1.0 - ratio)
```
pmeier, line 380:
```suggestion
        mix_image[..., :h_2, :w_2] += sample_2["image"] * (1.0 - ratio)
```

torchvision/prototype/transforms/_augment.py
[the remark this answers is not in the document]
pmeier: I just realized we are actually not passing anything here. Let's just pass the flat inputs for completeness.

pmeier:
I'm working on fixing `SimpleCopyPaste` now.

ambujpawar:
> And thus the behavior we want.

Yup, exactly! This is the behavior we want.
Thanks for fixing it, my eyes were not able to find it haha

ambujpawar:
> @ambujpawar is there anything left on your side that you want to do?

Nope. I think everything is done on my side and this mixupDetection feature is ready. :)

ambujpawar:
Hi @pmeier, Congrats on the torchvision v0.15 release. 
I just wanted to checkup on the future regarding the MixupDetection transform.
Is it still waiting on the topic of "_How to smoothly support "pairwise" transforms_" listed in #7319? 

Thanks in advance!! :)

pmeier:
Yes, unfortunately we are blocked by this. Sorry for not informing your earlier. We held off the batch transforms for now for the reason you listed above. I'll ping you here when we have figured it out. Thanks a lot for your patience!

ambujpawar:
Ah sure! No worries 
Thanks for the update! :)
BTW, the new transforms_v2 really look good. Thanks for them

mansour-b:
Hi everyone,

I was looking for an implementation of MixUp for object detection in `torchvision` and came across this PR.

I noticed it has not seen much activity for a while; could someone please clarify its current status or whether this is still planned to be merged?

I would be happy to help rebase or update it to align with the current API if that would be useful.

Thanks!

Issue #6491, opened 2022-08-25
query image size from bounding boxes or segmentation masks


pmeier:
Our current implementation of `query_chw` requires an image somewhere in the sample, since that is the only way to extract the number of channels. 

https://github.com/pytorch/vision/blob/f9966d228c5b809b0b68671ef168004f1e06d3fe/torchvision/prototype/transforms/_utils.py#L39

However, there are quite a few transformations that only require the image size, i.e. height and width:

https://github.com/pytorch/vision/blob/f9966d228c5b809b0b68671ef168004f1e06d3fe/torchvision/prototype/transforms/_geometry.py#L95

Although these transforms should technically work with `BoundingBox`'es and `SegmentationMask`'s as well, they will fail at the moment.

I see two possible solutions to this:

1. Split `query_chw` into `query_c` and `query_hw`. The former will only work with images, while the latter works with with images as well as `BoundingBox`'es and `SegmentationMask`'s. This was already implemented in a PR of mine, but I can't find it now. If someone does, feel free to link.
2. Option 1. requires us to go through the sample twice in case we need the number of channels and the image size. If we find that we need to reduce the number of times we do this, we could also allow `query_chw` to return `None` for the number of channels. In that case, I would introduce another flag `need_c: bool = False` that if set errors in case we don't find the number of channels. That would avoid much of duplicated error checking in the transformation like

    ```py
    c, h, w = query_chw(sample)
    if c is None:
        raise TypeError("I need number of channels, but found no image!")
    ```
    
    in favor of
    
    ```py
    c, h, w = query_chw(sample, needs_c=True)
    assert c is not None
    ```

cc @vfdev-5 @datumbox @bjuncek

pmeier:
@datumbox I think we have that now, right?

PR #6722, opened 2022-10-07
[prototype] Rewrite the meta dimension methods

datumbox:
Related to #6491

This PR:
- Eliminates the `get_chw` utility in favour of `get_dimensions`
- Favours the use of `get_num_channels` and `get_spatial_size`
- Extends `get_spatial_size` to support bboxes, masks etc

This is in preparation of removing the overuse of `query_chw` from the transforms to support single bboxes/masks without images/videos. The renaming of `image_size` to `spatial_size` will happen on a follow up PR.


vfdev-5 (approved):
lgtm, thanks @datumbox !

PR #6724, opened 2022-10-07
Adding support of Video to remaining Transforms and Kernels

datumbox:
Builds upon #6667 

This PR ensures that the following have support for Video:
- Updated Transforms: `Grayscale`, `RandomGrayscale`, `FiveCrop`, `TenCrop`, `ConvertImageDtype`
- Updated Dispatchers: `rgb_to_grayscale`, `get_image_size`, `five_crop`, `ten_crop`
- New Kernels: `five_crop_video`, `ten_crop_video`

Pending: `MixUp` and `CutMix` that might require additional thought on how to implement them.

### DIFF
--- a/torchvision/prototype/features/_video.py
+++ b/torchvision/prototype/features/_video.py
@@ -238,6 +238,7 @@ LegacyVideoTypeJIT = torch.Tensor
 TensorVideoType = Union[torch.Tensor, Video]
 TensorVideoTypeJIT = torch.Tensor
 
+# TODO: decide if we should do definitions for both Images and Videos or use unions in the methods
 ImageOrVideoType = Union[ImageType, VideoType]
 ImageOrVideoTypeJIT = Union[ImageTypeJIT, VideoTypeJIT]
 TensorImageOrVideoType = Union[TensorImageType, TensorVideoType]
--- a/torchvision/prototype/transforms/_auto_augment.py
+++ b/torchvision/prototype/transforms/_auto_augment.py
@@ -521,7 +521,7 @@ class AugMix(_AutoAugmentBase):
         mix = mix.view(orig_dims).to(dtype=image_or_video.dtype)
 
         if isinstance(orig_image_or_video, (features.Image, features.Video)):
-            mix = type(orig_image_or_video).wrap_like(orig_image_or_video, mix)  # type: ignore[arg-type]
+            mix = orig_image_or_video.wrap_like(orig_image_or_video, mix)  # type: ignore[arg-type]
         elif isinstance(orig_image_or_video, PIL.Image.Image):
             mix = F.to_image_pil(mix)
 
--- a/torchvision/prototype/transforms/_auto_augment.py
+++ b/torchvision/prototype/transforms/_auto_augment.py
@@ -483,7 +483,8 @@ class AugMix(_AutoAugmentBase):
         augmentation_space = self._AUGMENTATION_SPACE if self.all_ops else self._PARTIAL_AUGMENTATION_SPACE
 
         orig_dims = list(image_or_video.shape)
-        batch = image_or_video.view([1] * max(4 - image_or_video.ndim, 0) + orig_dims)
+        expected_dim = 5 if isinstance(orig_image_or_video, features.Video) else 4
+        batch = image_or_video.view([1] * max(expected_dim - image_or_video.ndim, 0) + orig_dims)
         batch_dims = [batch.size(0)] + [1] * (batch.ndim - 1)
 
         # Sample the beta weights for combining the original and augmented image or video. To get Beta, we use a
--- a/torchvision/prototype/transforms/_geometry.py
+++ b/torchvision/prototype/transforms/_geometry.py
@@ -155,12 +155,12 @@ class FiveCrop(Transform):
     """
     Example:
         >>> class BatchMultiCrop(transforms.Transform):
-        ...     def forward(self, sample: Tuple[Tuple[features.Image, ...], features.Label]):
-        ...         images, labels = sample
-        ...         batch_size = len(images)
-        ...         images = features.Image.wrap_like(images[0], torch.stack(images))
+        ...     def forward(self, sample: Tuple[Tuple[Union[features.Image, features.Video], ...], features.Label]):
+        ...         images_or_videos, labels = sample
+        ...         batch_size = len(images_or_videos)
+        ...         images_or_videos = features.Image.wrap_like(images_or_videos[0], torch.stack(images_or_videos))
         ...         labels = features.Label.wrap_like(labels, labels.repeat(batch_size))
-        ...         return images, labels
+        ...         return images_or_videos, labels
         ...
         >>> image = features.Image(torch.rand(3, 256, 256))
         >>> label = features.Label(0)
@@ -201,7 +201,7 @@ class TenCrop(Transform):
         self.size = _setup_size(size, error_msg="Please provide only two dimensions (h, w) for size.")
         self.vertical_flip = vertical_flip
 
-    def _transform(self, inpt: features.ImageType, params: Dict[str, Any]) -> List[features.ImageType]:
+    def _transform(self, inpt: features.ImageOrVideoType, params: Dict[str, Any]) -> List[features.ImageOrVideoType]:
         return F.ten_crop(inpt, self.size, vertical_flip=self.vertical_flip)
 
     def forward(self, *inputs: Any) -> Any:
--- a/torchvision/prototype/transforms/_geometry.py
+++ b/torchvision/prototype/transforms/_geometry.py
@@ -172,15 +172,15 @@ class FiveCrop(Transform):
         torch.Size([5])
     """
 
-    _transformed_types = (features.Image, PIL.Image.Image, features.is_simple_tensor)
+    _transformed_types = (features.Image, PIL.Image.Image, features.is_simple_tensor, features.Video)
 
     def __init__(self, size: Union[int, Sequence[int]]) -> None:
         super().__init__()
         self.size = _setup_size(size, error_msg="Please provide only two dimensions (h, w) for size.")
 
     def _transform(
-        self, inpt: features.ImageType, params: Dict[str, Any]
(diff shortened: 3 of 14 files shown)


pmeier:
From https://github.com/pytorch/vision/pull/6667#discussion_r989428697 only `MixUp` and `CutMix` are missing. Since there has been no discussion on the first PR, maybe we can do it here? Do we want those transforms for videos as well?

Plus, why are we only supporting images and videos in `{five, ten}_crop`? If we support more than images, i.e. making them more regular, shouldn't we also support bounding boxes and masks?

datumbox:
@pmeier Yes we should add Mixup/Cutmix. The PR is still in progress, I'll update it with more commits soon.

Concerning Five/Ten crop, this is usually an inference transform used in Classification. As discussed previously, supporting Detection/Segmentation will require updating the labels/bboxes not just the images. Thus we would need a BC break to support them. Extending support to them would require additional discussions. This PR aims to ensure we support all existing Video compatible kernels in V2 as we do in V1.

datumbox:
@pmeier As discussed offline, to unblock other pending PRs and reduce conflicts, we'll add MixUp/CutMix on a follow up.

torchvision/prototype/features/_video.py
```diff
@@ -238,6 +238,7 @@ def gaussian_blur(self, kernel_size: List[int], sigma: Optional[List[float]] = N
 TensorVideoType = Union[torch.Tensor, Video]
 TensorVideoTypeJIT = torch.Tensor
 
+# TODO: decide if we should do definitions for both Images and Videos or use unions in the methods
```
pmeier, line 241: Good point, but I agree to postpone this.

torchvision/prototype/transforms/_geometry.py
```diff
@@ -155,12 +155,12 @@ class FiveCrop(Transform):
...
     Example:
         >>> class BatchMultiCrop(transforms.Transform):
-        ...     def forward(self, sample: Tuple[Tuple[features.Image, ...], features.Label]):
-        ...         images, labels = sample
-        ...         batch_size = len(images)
-        ...         images = features.Image.wrap_like(images[0], torch.stack(images))
+        ...     def forward(self, sample: Tuple[Tuple[Union[features.Image, features.Video], ...], features.Label]):
+        ...         images_or_videos, labels = sample
+        ...         batch_size = len(images_or_videos)
+        ...         images_or_videos = features.Image.wrap_like(images_or_videos[0], torch.stack(images_or_videos))
```
pmeier, line 161:
```suggestion
        ...         image_or_video = images_or_videos[0]
        ...         images_or_videos = type(image_or_video).wrap_like(image_or_video, torch.stack(images_or_videos))
```

torchvision/prototype/transforms/_geometry.py
```diff
@@ -155,12 +155,12 @@ class FiveCrop(Transform):
...
         ...         labels = features.Label.wrap_like(labels, labels.repeat(batch_size))
-        ...         return images, labels
+        ...         return images_or_videos, labels
         ...
         >>> image = features.Image(torch.rand(3, 256, 256))
```
pmeier, line 165: Do we want to add a `Video` example here as well?

torchvision/prototype/transforms/_geometry.py
```diff
@@ -172,15 +172,21 @@ class FiveCrop(Transform):
...
-        self, inpt: features.ImageType, params: Dict[str, Any]
-    ) -> Tuple[features.ImageType, features.ImageType, features.ImageType, features.ImageType, features.ImageType]:
+        self, inpt: features.ImageOrVideoType, params: Dict[str, Any]
+    ) -> Tuple[
+        features.ImageOrVideoType,
+        features.ImageOrVideoType,
+        features.ImageOrVideoType,
+        features.ImageOrVideoType,
+        features.ImageOrVideoType,
+    ]:
```
pmeier, line 189: _Feel free to ignore if mypy is happy_

This is not accurate. We don't have `Tuple[features.ImageOrVideoType, ...]` here, but rather `Union[Tuple[features.ImageType],...], Tuple[features.VideoType, ...]]`. Meaning, the type will not vary inside the returned tuple. We either get a tuple of images or a tuple of videos.

torchvision/prototype/transforms/_geometry.py
```diff
@@ -194,14 +200,14 @@ class TenCrop(Transform):
...
-    _transformed_types = (features.Image, PIL.Image.Image, features.is_simple_tensor)
+    _transformed_types = (features.Image, PIL.Image.Image, features.is_simple_tensor, features.Video)
 
     def __init__(self, size: Union[int, Sequence[int]], vertical_flip: bool = False) -> None:
         super().__init__()
         self.size = _setup_size(size, error_msg="Please provide only two dimensions (h, w) for size.")
         self.vertical_flip = vertical_flip
 
-    def _transform(self, inpt: features.ImageType, params: Dict[str, Any]) -> List[features.ImageType]:
+    def _transform(self, inpt: features.ImageOrVideoType, params: Dict[str, Any]) -> List[features.ImageOrVideoType]:
```
pmeier, line 210: Same as above.

torchvision/prototype/transforms/functional/_meta.py
```diff
@@ -55,6 +55,10 @@ def get_spatial_size_image_pil(image: PIL.Image.Image) -> List[int]:
     return [height, width]
 
 
+# TODO: Should we have get_spatial_size_video here? How about masks/bbox etc? What is the criterion for deciding when
```
pmeier, line 58: `get_spatial_size` should apply to everything, right? That was the whole reason we have extracted it out, because bounding boxes and masks can provide this information, while `num_channels` is reserved for images and videos.

pmeier (approved):
Some minor comments inline. Otherwise LGTM if CI is green.

torchvision/prototype/transforms/_geometry.py
datumbox: This is why IMO we should avoid `features.ImageOrVideoType` and instead define it as `Union[Image, Video]`. Since this will be fixed in a follow up, there is no point messing with mypy (which is happy) here. I'll implement this on a follow up.

torchvision/prototype/transforms/functional/_meta.py
datumbox: I've already updated `get_spatial_size` to handle all inputs. 

I think you are trying to answer a different question from what I ask here. What I think we should discuss is whether there should be specific kernels for each type, unrelated to whether the dispatcher can handle everything. We already have kernels (like `erase_video`) that aren't necessarily used in the dispatcher. So here I'm asking, what should the convention over providing kernels for individual types should be.

torchvision/prototype/transforms/_geometry.py
datumbox: If you think it adds more info we can. But IMO it's going to be so similar that it's just going to make the example more verbose. WDYT?

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -483,7 +483,8 @@ def forward(self, *inputs: Any) -> Any:
         augmentation_space = self._AUGMENTATION_SPACE if self.all_ops else self._PARTIAL_AUGMENTATION_SPACE
 
         orig_dims = list(image_or_video.shape)
-        batch = image_or_video.view([1] * max(4 - image_or_video.ndim, 0) + orig_dims)
+        expected_dim = 5 if isinstance(orig_image_or_video, features.Video) else 4
```
datumbox, line 486: Bug fix.

torchvision/prototype/transforms/functional/_meta.py
pmeier: Ah, sorry, yes I was confused. That is a good question and I don't have an answer for it yet. My gut says that we should stay consistent and provide the kernels just as we do for the other transforms.

torchvision/prototype/transforms/_auto_augment.py
pmeier: Good spot!

torchvision/prototype/transforms/_geometry.py
pmeier: Agreed!

torchvision/prototype/transforms/functional/_meta.py
datumbox: Same feeling here. I'll leave the TODO for the follow up. I think we can answer this on the PR where we switch `image_size` to `spatial_size`

torchvision/prototype/transforms/_geometry.py
datumbox: Sorry I forced pushed and lost this change. Let me reapply.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -1401,16 +1401,27 @@ def five_crop_image_pil(
...
 ]:
-    # TODO: consider breaking BC here to return List[features.ImageTypeJIT] to align this op with `ten_crop`
+    # TODO: consider breaking BC here to return List[features.ImageOrVideoTypeJIT] to align this op with `ten_crop`
     if isinstance(inpt, torch.Tensor):
         output = five_crop_image_tensor(inpt, size)
-        if not torch.jit.is_scripting() and isinstance(inpt, features.Image):
-            output = tuple(features.Image.wrap_like(inpt, item) for item in output)  # type: ignore[assignment]
+        if not torch.jit.is_scripting() and isinstance(inpt, (features.Image, features.Video)):
+            tmp = (inpt.wrap_like(inpt, item) for item in output)  # type: ignore[arg-type]
+            output = tmp  # type: ignore[assignment]
```
datumbox, line 1424: Sorry for the delay on merging. JIT was driving nuts:
```
RuntimeError: expected type comment but found 'ident' here:
output = (inpt.wrap_like(inpt, item) for item in output)  # type: ignore[assignment,arg-type]
~~~~~~ <--- HERE
```

Something breaks if ignore contains both assignment and arg-type directives. I suspect JIT might be looking for an `assignment` directive and completely breaks when we apply the mypy workaround. This idiom passes both JIT and mypy.

PR #6729, opened 2022-10-10
Remove performance workaround for mask resize

datumbox:
Removing performance workaround now that the https://github.com/pytorch/pytorch/issues/83840 is resolved.

### DIFF
--- a/torchvision/prototype/transforms/functional/_geometry.py
+++ b/torchvision/prototype/transforms/functional/_geometry.py
@@ -14,12 +14,7 @@ from torchvision.transforms.functional import (
     pil_to_tensor,
     to_pil_image,
 )
-from torchvision.transforms.functional_tensor import (
-    _cast_squeeze_in,
-    _cast_squeeze_out,
-    _parse_pad_padding,
-    interpolate,
-)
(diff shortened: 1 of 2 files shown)


NicolasHug (approved):

pmeier (approved):
Could you also remove the sample inputs that tested this specific branch:

https://github.com/pytorch/vision/blob/af54e5645399bc67711155de2f8bb9cb1f4ebbe1/test/prototype_transforms_kernel_infos.py#L252-L260

Otherwise LGTM if CI is green.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -128,34 +123,12 @@ def resize_image_tensor(
...
-
-        else:
-            image = _FT.resize(
-                image,
-                size=[new_height, new_width],
-                interpolation=interpolation.value,
-                antialias=antialias,
-            )
+        image = _FT.resize(
+            image,
```
pmeier, line 127: Test failures are happening, because you missed this above

```suggestion
            image.view(-1, num_channels, old_height, old_width),
```

Issue #6728, opened 2022-10-10
not all prototype transforms are serializable


pmeier:
This is important for two reasons:

1. If we stick to `torchdata` for datasets v2, every object on the datapipe graph has to be serializable for multiprocessing. IIRC, this is only a requirement if forking from the main process is not possible, but that is the case on Windows.
2. Our classification references have the ability to cache the dataset which includes the transformations:

https://github.com/pytorch/vision/blob/6e203b44098c3371689f56abc17b7c02bd51a261/references/video_classification/train.py#L189

So far we don't have a unified test framework for our prototype transforms and thus it is not easy to add tests for all transforms. Note that we don't need tests for the functional kernels and dispatchers, since they are already tested for JIT scriptability and that protocol is a lot stricter than pickling.

PR #6730, opened 2022-10-10
make _setup_fill_arg serializable

pmeier:
Addresses #6728. The only other usages of `lambda` are happening in the AA transforms: 

https://github.com/pytorch/vision/blob/6e203b44098c3371689f56abc17b7c02bd51a261/torchvision/prototype/transforms/_auto_augment.py#L150

This was a perf optimization we did in v2. In v1 we have 

https://github.com/pytorch/vision/blob/6e203b44098c3371689f56abc17b7c02bd51a261/torchvision/transforms/autoaugment.py#L226

Since we want to use AA transforms for videos as well (we do, don't we?), we probably either need to revert this or use a similar technique I used here. I'll look into this next.

As explained in #6728, we don't have unified testing for transforms yet. Thus, I cannot guarantee that these two instances cover every non-serializable things on transforms v2. I will work on testing when video training is up and running.

### DIFF
--- a/torchvision/prototype/transforms/_utils.py
+++ b/torchvision/prototype/transforms/_utils.py
@@ -1,6 +1,6 @@
+import functools
 import numbers
 from collections import defaultdict
-
 from typing import Any, Callable, Dict, Sequence, Tuple, Type, Union
 
 import PIL.Image
@@ -43,13 +43,19 @@ def _check_fill_arg(fill: Union[FillType, Dict[Type, FillType]]) -> None:
             raise TypeError("Got inappropriate fill arg")
 
 
+def _default_fill(fill: FillType) -> FillType:
+    return fill
+
+
 def _setup_fill_arg(fill: Union[FillType, Dict[Type, FillType]]) -> Dict[Type, FillType]:
     _check_fill_arg(fill)
 
     if isinstance(fill, dict):
         return fill
 
-    return defaultdict(lambda: fill)  # type: ignore[return-value, arg-type]
+    # This weird looking construct only exists, since `lambda`'s cannot be serialized by pickle.
+    # If it were possible, we could replace this with `defaultdict(lambda: fill)`
+    return defaultdict(functools.partial(_default_fill, fill))
 
 
 def _check_padding_arg(padding: Union[int, Sequence[int]]) -> None:


datumbox:
@pmeier Looks good. Let's do the same for AA.

pmeier:
Not sure why, but it seems AA seems to work as is:

```py
import pickle

import torch
from torchvision.prototype import transforms


image = torch.randint(0, 256, (3, 512, 512), dtype=torch.uint8)

for transform in [
    transforms.AutoAugment(),
    transforms.AugMix(),
    transforms.AutoAugment(),
    transforms.RandAugment(),
    transforms.TrivialAugmentWide(),
]:
    serialized = pickle.dumps(transform)
    deserialized = pickle.loads(serialized)

    deserialized(image)
```

I suggested we don't make it more complicated until we actually hit a problem. Using the same technique that I used here, is not really possible for AA, since the return value is dependent on the input parameters.

datumbox:
> Using the same technique that I used here, is not really possible for AA, since the return value is dependent on the input parameters.

Isn't the return type of the lambda always `Optional[Tensor]`?

pmeier:
> Isn't the return type of the lambda always `Optional[Tensor]`?

Yeah, but the return value is different and depends on the inputs. `functools.partial` only works in this PR, because the return type is static and we know it in advance. For AA we would need to replace 


https://github.com/pytorch/vision/blob/6e203b44098c3371689f56abc17b7c02bd51a261/torchvision/prototype/transforms/_auto_augment.py#L150

with something like


```py
def _auto_augment_shearx(num_bins: int, height: int, width: int) -> torch.Tensor:
    return torch.linspace(0.0, 0.3, num_bins)
```

If we would know `num_bins` at class definition, we could do something like

```py
functools.partial(torch.linspace, 0.0, 0.3, STATIC_NUM_BINS)
```

This is indeed fixed for some of the AA transforms like

https://github.com/pytorch/vision/blob/af54e5645399bc67711155de2f8bb9cb1f4ebbe1/torchvision/prototype/transforms/_auto_augment.py#L291

but `RandAugment` for example, takes this as input parameter:

https://github.com/pytorch/vision/blob/af54e5645399bc67711155de2f8bb9cb1f4ebbe1/torchvision/prototype/transforms/_auto_augment.py#L339

Worse, although not used often, know anything about `height` and `width` at definition time. Thus, all definitions that use `height` or `width` would need to have a standalone function.

PR #6733, opened 2022-10-10
Add video support on MixUp and CutMix

### LINKED ISSUE #6731: Add support for Video on MixUp and CutMix
### 🚀 The feature

MixUp and CutMix don't currently support Videos. We might want to rethink how we combine the batch to avoid mixing frames from the same video. [Here](https://github.com/facebookresearch/pytorchvideo/blob/a5cc0c8a3e1c3f8c963b75e0bd5409650ad7616a/pytorchvideo/transforms/mix.py#L78) is a potential workaround.

cc @vfdev-5 @bjuncek @pmeier

datumbox:
Fixes #6731

### DIFF
--- a/torchvision/prototype/transforms/_augment.py
+++ b/torchvision/prototype/transforms/_augment.py
@@ -107,8 +107,8 @@ class _BaseMixupCutmix(_RandomApplyTransform):
         self._dist = torch.distributions.Beta(torch.tensor([alpha]), torch.tensor([alpha]))
 
     def forward(self, *inputs: Any) -> Any:
-        if not (has_any(inputs, features.Image, features.is_simple_tensor) and has_any(inputs, features.OneHotLabel)):
-            raise TypeError(f"{type(self).__name__}() is only defined for tensor images and one-hot labels.")
+        if not (has_any(inputs, features.Image, features.Video, features.is_simple_tensor) and has_any(inputs, features.OneHotLabel)):
+            raise TypeError(f"{type(self).__name__}() is only defined for tensor images/videos and one-hot labels.")
         if has_any(inputs, PIL.Image.Image, features.BoundingBox, features.Mask, features.Label):
             raise TypeError(
                 f"{type(self).__name__}() does not support PIL images, bounding boxes, masks and plain labels."
@@ -116,10 +116,10 @@ class _BaseMixupCutmix(_RandomApplyTransform):
         return super().forward(*inputs)
 
     def _mixup_onehotlabel(self, inpt: features.OneHotLabel, lam: float) -> features.OneHotLabel:
-        if inpt.ndim < 2:
+        if inpt.ndim != 2:
             raise ValueError("Need a batch of one hot labels")
         output = inpt.clone()
-        output = output.roll(1, -2).mul_(1 - lam).add_(output.mul_(lam))
+        output = output.flip(0).mul_(1.0 - lam).add_(output.mul_(lam))
         return features.OneHotLabel.wrap_like(inpt, output)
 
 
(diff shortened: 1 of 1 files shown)


torchvision/prototype/transforms/_augment.py
```diff
@@ -107,19 +107,19 @@ def __init__(self, alpha: float, p: float = 0.5) -> None:
...
+            raise TypeError(f"{type(self).__name__}() is only defined for tensor images/videos and one-hot labels.")
         if has_any(inputs, PIL.Image.Image, features.BoundingBox, features.Mask, features.Label):
             raise TypeError(
                 f"{type(self).__name__}() does not support PIL images, bounding boxes, masks and plain labels."
             )
         return super().forward(*inputs)
 
     def _mixup_onehotlabel(self, inpt: features.OneHotLabel, lam: float) -> features.OneHotLabel:
-        if inpt.ndim < 2:
+        if inpt.ndim != 2:
```
datumbox, line 119: I think it's better to specify the exact expected dimensions. I don't see any use-case where we expect anything other than 2.

torchvision/prototype/transforms/_augment.py
```diff
@@ -129,14 +129,15 @@ def _get_params(self, sample: Any) -> Dict[str, Any]:
 
     def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
         lam = params["lam"]
-        if isinstance(inpt, features.Image) or features.is_simple_tensor(inpt):
-            if inpt.ndim < 4:
-                raise ValueError("Need a batch of images")
+        if isinstance(inpt, (features.Image, features.Video)) or features.is_simple_tensor(inpt):
+            expected_dim = 5 if isinstance(inpt, features.Video) else 4
+            if inpt.ndim != expected_dim:
```
datumbox, line 134: Perhaps we could allow here `<`. Imagine that someone throws in videos but in Tensor format. By not checking for `!=` we could support this usecase.

torchvision/prototype/transforms/_augment.py
```diff
@@ -129,14 +129,15 @@ def _get_params(self, sample: Any) -> Dict[str, Any]:
...
+                raise ValueError("The transform expects a batched input")
             output = inpt.clone()
-            output = output.roll(1, -4).mul_(1 - lam).add_(output.mul_(lam))
+            output = output.roll(1, 0).mul_(1.0 - lam).add_(output.mul_(lam))
```
datumbox, line 137: By hardcoding the batch dimension on `0`, we can make sure we support cases where pure-tensor videos are provided.

torchvision/prototype/transforms/_augment.py
pmeier, line 134: You mean batched videos, i.e. `(B, T, C, H, W)`, right? Because

> videos but in Tensor format

would be `(T, C, H, W)` and that is already covered. Although I'm not sure if we want to apply this to the temporal dimension.

torchvision/prototype/transforms/_augment.py
```diff
@@ -129,14 +129,15 @@ def _get_params(self, sample: Any) -> Dict[str, Any]:
 
     def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
         lam = params["lam"]
-        if isinstance(inpt, features.Image) or features.is_simple_tensor(inpt):
-            if inpt.ndim < 4:
-                raise ValueError("Need a batch of images")
+        if isinstance(inpt, (features.Image, features.Video)) or features.is_simple_tensor(inpt):
+            expected_dim = 5 if isinstance(inpt, features.Video) else 4
```
pmeier, line 133: Can we call this

```suggestion
            expected_dims = 5 if isinstance(inpt, features.Video) else 4
```

or

```suggestion
            expected_ndim = 5 if isinstance(inpt, features.Video) else 4
```

to make it more clear what we are checking here?

torchvision/prototype/transforms/_augment.py
```diff
@@ -169,17 +170,18 @@ def _get_params(self, sample: Any) -> Dict[str, Any]:
...
 
     def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
-        if isinstance(inpt, features.Image) or features.is_simple_tensor(inpt):
+        if isinstance(inpt, (features.Image, features.Video)) or features.is_simple_tensor(inpt):
             box = params["box"]
-            if inpt.ndim < 4:
-                raise ValueError("Need a batch of images")
+            expected_dim = 5 if isinstance(inpt, features.Video) else 4
+            if inpt.ndim != expected_dim:
+                raise ValueError("The transform expects a batched input")
```
pmeier, line 177: Maybe we can have a small helper method to keep both `CutMix` and `MixUp` in sync here? Like `_check_image_or_video`.

pmeier (approved):
A few inline comments, but otherwise LGTM if CI is green.

torchvision/prototype/transforms/_augment.py
datumbox: IDK. Is it really worth saving 3 lines and introducing more private methods in the base class?

datumbox:
@pmeier Can you check again? I had to modify the tests

pmeier (approved):
Patch still looks good if CI is happy.

PR #6735, opened 2022-10-10
Add missing type hints to anchor_utils

vsuryamurthy:
This PR also adds the missing type hints.

### DIFF
--- a/torchvision/models/detection/anchor_utils.py
+++ b/torchvision/models/detection/anchor_utils.py
@@ -61,7 +61,7 @@ class AnchorGenerator(nn.Module):
         aspect_ratios: List[float],
         dtype: torch.dtype = torch.float32,
         device: torch.device = torch.device("cpu"),
-    ):
+    ) -> Tensor:
         sizes = torch.as_tensor(sizes, dtype=dtype, device=device)
         aspect_ratios = torch.as_tensor(aspect_ratios, dtype=dtype, device=device)
         h_ratios = torch.sqrt(aspect_ratios)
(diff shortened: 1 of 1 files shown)


torchvision/models/detection/anchor_utils.py
```diff
@@ -52,31 +52,31 @@ def __init__(
...
     # TODO: https://github.com/pytorch/pytorch/issues/26792
-    # For every (aspect_ratios, scales) combination, output a zero-centered anchor with those values.
-    # (scales, aspect_ratios) are usually an element of zip(self.scales, self.aspect_ratios)
+    # For every (aspect_ratios, sizes) combination, output a zero-centered anchor with those values.
+    # (sizes, aspect_ratios) are usually an element of zip(self.sizes, self.aspect_ratios)
     # This method assumes aspect ratio = height / width for an anchor.
     def generate_anchors(
         self,
-        scales: List[int],
+        sizes: List[int],
```
datumbox, line 60: This could be a BC-breaking change for users who do named arguments. Shall we keep the original names?

vsuryamurthy: I understand, however this inconsistency does result in a bit of confusion. I assume there is no work around for this but to stick with the old naming?

vsuryamurthy: I will close this PR if there is no work around, do you think the missing type hints are worthy of a separate PR or should I ignore it for now?
Further lines 42-46 (where the aspect_ratios and sizes are checked for a list or tuple) in this file is quite interesting as it makes of an assumption of what the user wants but I think it would be best to error out. What do you think?
All these are minor issues that I faced while playing with the code.

datumbox: These are all great points @vsuryamurthy. If we were to rewrite the code, we would probably be better naming conventions. The problem is TorchVision comes with very strong BC guarantees and because the method is public, the risk of renaming and breaking user code is quite large. I think you can keep the typing additions of this PR; just revert the variable rename and we should be good to go.

vsuryamurthy: Alright fair enough! I have restored the naming and only added the type hints.

datumbox (approved):
LGTM, thanks!

PR #6736, opened 2022-10-10
[prototype] Switch to `spatial_size`

datumbox:
This PR:
- [x] Renames `image_size` to `spatial_size` everywhere in the code-base
- [x] Adds `get_num_channels_video` and `get_spatial_size_*` kernels for videos, masks and bboxes
- [x] Adds `get_num_frames` dispatcher and `get_num_frames_video` kernel for JIT.
- [x] Eliminates the over-use of `query_chw` to make things work with bboxes and masks

### DIFF
--- a/torchvision/prototype/transforms/_augment.py
+++ b/torchvision/prototype/transforms/_augment.py
@@ -11,7 +11,7 @@ from torchvision.prototype import features
 from torchvision.prototype.transforms import functional as F, InterpolationMode
 
 from ._transform import _RandomApplyTransform
-from ._utils import has_any, query_chw, query_hw
+from ._utils import has_any, query_chw, query_spatial_size
 
 
 class RandomErasing(_RandomApplyTransform):
@@ -153,7 +153,7 @@ class RandomCutmix(_BaseMixupCutmix):
     def _get_params(self, sample: Any) -> Dict[str, Any]:
         lam = float(self._dist.sample(()))
 
-        H, W = query_hw(sample)
+        H, W = query_spatial_size(sample)
 
         r_x = torch.randint(W, ())
         r_y = torch.randint(H, ())
--- a/torchvision/prototype/transforms/_augment.py
+++ b/torchvision/prototype/transforms/_augment.py
@@ -11,7 +11,7 @@ from torchvision.prototype import features
 from torchvision.prototype.transforms import functional as F, InterpolationMode
 
 from ._transform import _RandomApplyTransform
-from ._utils import has_any, query_chw
+from ._utils import has_any, query_chw, query_hw
 
 
 class RandomErasing(_RandomApplyTransform):
@@ -153,7 +153,7 @@ class RandomCutmix(_BaseMixupCutmix):
     def _get_params(self, sample: Any) -> Dict[str, Any]:
         lam = float(self._dist.sample(()))
 
-        _, H, W = query_chw(sample)
+        H, W = query_hw(sample)
 
         r_x = torch.randint(W, ())
         r_y = torch.randint(H, ())
--- a/torchvision/prototype/transforms/_color.py
+++ b/torchvision/prototype/transforms/_color.py
@@ -100,7 +100,7 @@ class RandomPhotometricDistort(Transform):
         self.p = p
 
     def _get_params(self, sample: Any) -> Dict[str, Any]:
-        num_channels, _, _ = query_chw(sample)
+        num_channels, *_ = query_chw(sample)
         return dict(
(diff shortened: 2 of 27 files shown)


torchvision/prototype/transforms/functional/_meta.py
```diff
@@ -55,21 +59,26 @@ def get_spatial_size_image_pil(image: PIL.Image.Image) -> List[int]:
...
+    return get_spatial_size_image_tensor(video)
+
+
+def get_spatial_size_mask(mask: torch.Tensor) -> List[int]:
+    return get_spatial_size_image_tensor(mask)
+
+
+@torch.jit.unused
+def get_spatial_size_bounding_box(bounding_box: features.BoundingBox) -> List[int]:
+    return list(bounding_box.spatial_size)
```
datumbox, line 72: Addition of the `get_spatial_size_*` kernels. The one of BBox can't have a JIT-scriptable implementation as it relies on Tensor Subclassing to retrieve this info.

torchvision/prototype/transforms/functional/_meta.py
```diff
@@ -55,21 +59,26 @@ def get_spatial_size_image_pil(image: PIL.Image.Image) -> List[int]:
...
-        image_size = getattr(inpt, "image_size", None)
-        if image_size is not None:
-            return list(image_size)
-        else:
-            raise ValueError(f"Type {inpt.__class__} doesn't have spatial size.")
+    elif isinstance(inpt, (features.Image, features.Video, features.BoundingBox, features.Mask)):
+        return list(inpt.spatial_size)
     else:
-        return get_spatial_size_image_pil(inpt)
+        return get_spatial_size_image_pil(inpt)  # type: ignore[no-any-return]
```
datumbox, line 81: Refactoring to avoid the `getattr` idiom. After that, mypy complains for the PIL kernel. It's unclear to be why it thinks we return `Any`. The `get_spatial_size_video` returns a `List[int]`.

torchvision/prototype/transforms/_color.py
```diff
@@ -100,7 +100,7 @@ def __init__(
         self.p = p
 
     def _get_params(self, sample: Any) -> Dict[str, Any]:
-        num_channels, _, _ = query_chw(sample)
+        num_channels, *_ = query_chw(sample)
```
datumbox, line 103: I didn't introduce yet another method for extracting channels only. This is indeed less elegant but doesn't introduce any limitations as the input is required to have channels. This happens in one more place in the codebase.

torchvision/prototype/transforms/_utils.py
```diff
@@ -98,6 +98,22 @@ def query_chw(sample: Any) -> Tuple[int, int, int]:
...
+        for item in flat_sample
+        if isinstance(item, (features.Image, PIL.Image.Image, features.Video, features.Mask, features.BoundingBox))
+        or features.is_simple_tensor(item)
+    }
+    if not hws:
+        raise TypeError("No image, video, mask or bounding box was found in the sample")
+    elif len(hws) > 1:
+        raise ValueError(f"Found multiple HxW dimensions in the sample: {sequence_to_str(sorted(hws))}")
+    h, w = hws.pop()
+    return h, w
```
datumbox, line 114: Lots of code duplication with `query_chw`. The two methods differ on the callable, the checked types, the error messages and the return type. I was tempted to write something that passes a callable and tries to reduce duplicate code but it become unnecessarily complex. Happy to implement other approaches if you have better ideas.

datumbox:
I got a few [failures](https://github.com/pytorch/vision/actions/runs/3221012858/jobs/5268434562) on Windows. They don't look like related at a first glance but then this PR touches too many things, so I'm not 100% sure. I'll check again tomorrow.

vfdev-5 (approved):
OK to me, thanks @datumbox !
Just a minor suggestion to rename query_hw to query_spatial_size... (not blocking)

torchvision/prototype/transforms/_augment.py
```diff
@@ -153,7 +153,7 @@ class RandomCutmix(_BaseMixupCutmix):
     def _get_params(self, sample: Any) -> Dict[str, Any]:
         lam = float(self._dist.sample(()))
 
-        _, H, W = query_chw(sample)
+        H, W = query_hw(sample)
```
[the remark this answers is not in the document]
datumbox: The method was renamed to `query_spatial_size` after https://github.com/pytorch/vision/pull/6736#pullrequestreview-1136986870

PR #6739, opened 2022-10-11
Revert "Remove performance workaround for mask resize"

datumbox:
Reverts pytorch/vision#6729

We are putting the performance workaround back because the mitigation on PyTorch Core was reverted: https://github.com/pytorch/pytorch/pull/86361

Let's keep this workaround in until the mitigation is permanently landed on Core.


datumbox:
It seems that the original PR got relanded. I'm closing this for now.

PR #6741, opened 2022-10-11
Add seeds on Kernel Info and reduce randomness for Gaussian Blur

datumbox:
Trying to reduce the [flakiness](https://github.com/pytorch/vision/actions/runs/3225886257/jobs/5278734196) on the prototype `gaussian_blur` kernel. This PR:
- Adds an optional seed in all `KernelInfo` objects that can be set at the beginning of each test prior generating the data.
- Removes the random sampling from the sizes/frames from Gaussian Blur data because these random calls are not covered by the seed.

### DIFF
--- a/test/prototype_transforms_kernel_infos.py
+++ b/test/prototype_transforms_kernel_infos.py
@@ -49,7 +49,7 @@ class KernelInfo(InfoBase):
         test_marks=None,
         # See InfoBase
         closeness_kwargs=None,
-        seed=None
+        seed=None,
     ):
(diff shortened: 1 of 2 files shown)


vfdev-5:
Thanks @datumbox , I rerun ubuntu test one more time to check.

datumbox:
@vfdev-5 Yeah good call. I couldn't make the test fail locally at all. Tried every seed in [0,100] and also randomly sampled but nothing. I'm not 100% this resolves the issue, let's see what the rerun does.

PR #6742, opened 2022-10-11
Fix for windows and python 3.8 call to add_dll_directory

atalman:
Fix for windows and python 3.8 call to add_dll_directory
Similar to: https://github.com/pytorch/audio/pull/2747


datumbox (approved):
LGTM, thanks! Do we need to cherrypick this into the release?

atalman:
> LGTM, thanks! Do we need to cherrypick this into the release?
> 
> Let's wait for the tests to complete before merging.

Yes we do need to cherry-pick this. And for sure will wait for this to pass tests

PR #6748, opened 2022-10-12
cache traceback together with exceptions

pmeier:
We are caching the scripting of kernels inside our prototype tests:

https://github.com/pytorch/vision/blob/11a2eeda8fb127a7ad72b4c98ca918b93055c1e7/test/test_prototype_transforms_functional.py#L21-L22

We do this because the test 

https://github.com/pytorch/vision/blob/11a2eeda8fb127a7ad72b4c98ca918b93055c1e7/test/test_prototype_transforms_functional.py#L83

is run multiple times for the same kernel, but with different sample inputs. In fact, we only have 78 `KernelInfo`'s but the test is run 7826 times. Meaning, in 99% of the runs, we script the kernel without any information gain whatsoever.

However, with the current implementation of `@cache`, the traceback of re-raised cached exceptions is not cleaned properly by `pytest`. TL;DR: the traceback of an exception is mutable. By re-raising, the `pytest` frames are getting appended over and over again, but `pytest` only removes them once. For a detailed explanation see https://github.com/pytest-dev/pytest/issues/10363.

This leads to ginormous tracebacks that drown the actual information in the noise. For example, I intentionally made `pad_bounding_box` non-scriptable:

```sh
pytest -q test/test_prototype_transforms_functional.py::TestKernels::test_scripted_vs_eager -k pad_bounding_box | wc -l
```

- `main`: 499510
- PR: 3790
- `main` without `@cache`: 3358

The few extra lines stem from the fact that the `@cache` wrapper adds minimal traceback itself.

### DIFF
--- a/test/common_utils.py
+++ b/test/common_utils.py
@@ -210,7 +210,7 @@ def cache(fn):
     """
     sentinel = object()
     out_cache = {}
-    exc_cache = {}
+    exc_tb_cache = {}
 
     @functools.wraps(fn)
     def wrapper(*args, **kwargs):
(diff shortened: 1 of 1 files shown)


datumbox (approved):
Kind of ugly but we already use a similar logic for caching the output... I'm approving as I agree that seeing the actual exceptions is necessary for debugging (I was left quite a few times scratching my head to find out what went wrong on tests) but I'm hoping we can adopt a better solution soon.

PR #6749, opened 2022-10-12
Seed transform tests

pmeier:
Follow-up to and partial revert of #6741.

Although it is ok to have `"random"` values in the inputs definition, we don't want actual random values inside our tests. The former communicates that the actual values don't really matter, while the latter makes sure we don't have flaky tests and can reproduce CI behavior locally.

Thus, this PR reverts the seed changes of #6741 and adds a fixture that sets the RNG seed before every test. If #6750 is resolved, we can probably also put back the `"random"` sizes for the `gaussian_blur_*` kernels.

### DIFF
--- a/test/prototype_transforms_kernel_infos.py
+++ b/test/prototype_transforms_kernel_infos.py
@@ -1304,7 +1304,7 @@ KERNEL_INFOS.extend(
 
 def sample_inputs_gaussian_blur_image_tensor():
     make_gaussian_blur_image_loaders = functools.partial(
-        make_image_loaders, sizes=["random"], color_spaces=[features.ColorSpace.RGB]
+        make_image_loaders, sizes=[(7, 33)], color_spaces=[features.ColorSpace.RGB]
     )
 
     for image_loader, kernel_size in itertools.product(make_gaussian_blur_image_loaders(), [5, (3, 3), [3, 3]]):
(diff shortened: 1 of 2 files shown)


datumbox:
I don't mind reverting my changes.

The reason why I didn't follow this approach is because the population of the `KERNEL_INFOS` dictionary happens long before the tests actually run and thus before the seed is set. This leads to different sizes for the actual data tensors which leads to the seed setting not having any effect. Am I missing something?

pmeier:
> As a result the random calls to determine the sizes yield different results every time.

True. And I need to look into how we can fix that. This is why I put back your changes to the sample inputs definition in 7d25068f21387025fb1a7c49e6ae340ba80a6875. 

> Consequently the actual data tensors will be different from run-to-run making the seed setting to have no effect.

Yes, they will be different in shape. This is not desired and I will look into it as explained above. However, your PR also does not fix this as the setting of the seed only happens inside the test and thus at the same time as the fixture of this PR does.

datumbox:
> However, your PR also does not fix this as the setting of the seed only happens inside the test and thus at the same time as the fixture of this PR does.

My original PR circumvents the problem by not having random calls for the specific problematic kernel during the `KERNEL_INFOS` initialisation. So right now, on main branch, the random data are deterministic (I've checked).

I'm not sure I get how this PR fixes the issue as you say. Does `fix_rng_seed()` execute before `KERNEL_INFOS` get populated?

datumbox (approved):

Issue #6645, opened 2022-09-26
Allow "nearest-exact" interpolation mode?


pmeier:
Currently the following interpolation modes are allowed 

https://github.com/pytorch/vision/blob/7046e56fe4370e94339b3e8b6fd011e285294a3a/torchvision/transforms/functional.py#L21-L32

Since `torch==1.11.0` (pytorch/pytorch#64501 of @vfdev-5 to be exact), `torch.nn.functional` also supports `mode="nearest-exact"`:

> Mode `mode='nearest-exact'` matches Scikit-Image and PIL nearest neighbours interpolation algorithms and fixes known issues with `mode='nearest'`. This mode is introduced to keep backward compatibility. Mode `mode='nearest'` matches buggy OpenCV's `INTER_NEAREST` interpolation algorithm.

Given that we are aligning more with PIL and `"nearest"` is described as "bugger", can we add support for `"nearest-exact"`? 

If yes, we should also think about changing all our default values to it. That might be a bit cumbersome for the users, but we could also remap its name. Meaning after the whole deprecation period is through, `"nearest"` just maps to `"nearest-exact"` of `interpolate` and we have a `"nearest-legacy"` or the like that maps to `"nearest"`. We already do a name mapping for other interpolation modes:

https://github.com/pytorch/vision/blob/7046e56fe4370e94339b3e8b6fd011e285294a3a/torchvision/transforms/functional_tensor.py#L412-L414

Deprecation process could look like this where `r` denotes the current release.

- `r+1`: add `"nearest-exact"` as valid interpolation mode and deprecate not passing a value explicitly, i.e.

    ```py
    def foo(..., mode=None):
        if mode is None:
            warnings.warn(...)
            mode = "nearest"
    ```
    
    Plus, this should add a `"nearest-legacy"` that aliases the current `"nearest"`. We also need to warn if `"nearest"` is passed.
- `r+2`: fail if interpolation mode is not passed explicitly or `"nearest"` is passed.
- `r+3`: re-introduce `"nearest"` as default value for the interpolation mode, but map it internally to `"nearest-exact"`.

bhack:
Can we close this?

NicolasHug:
Looks like the first part of this issue, i.e. introducing nearest-exact mode, was addressed in https://github.com/pytorch/vision/pull/6754.

[...]

PR #6754, opened 2022-10-12
introduce nearest-exact interpolation

pmeier:
Addresses the first part of #6645. As discussed offline with @datumbox, exposing this mode is non-controversial. Whether or not we should make it the default and how the transition period should look like is another matter entirely. Let's do that if more pressing matters are finished.

@vfdev I would like to have your input on how to test this properly on the stable transforms test.

### DIFF
--- a/torchvision/transforms/functional.py
+++ b/torchvision/transforms/functional.py
@@ -20,10 +20,12 @@ from . import functional_pil as F_pil, functional_tensor as F_t
 
 class InterpolationMode(Enum):
     """Interpolation modes
-    Available interpolation methods are ``nearest``, ``bilinear``, ``bicubic``, ``box``, ``hamming``, and ``lanczos``.
+    Available interpolation methods are ``nearest``, ``nearest-exact``, ``bilinear``, ``bicubic``, ``box``, ``hamming``,
+    and ``lanczos``.
     """
 
     NEAREST = "nearest"
+    NEAREST_EXACT = "nearest-exact"
     BILINEAR = "bilinear"
     BICUBIC = "bicubic"
     # For PIL compatibility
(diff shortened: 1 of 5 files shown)


torchvision/transforms/functional.py
```diff
@@ -50,6 +52,7 @@ def _interpolation_modes_from_int(i: int) -> InterpolationMode:
     InterpolationMode.NEAREST: 0,
     InterpolationMode.BILINEAR: 2,
     InterpolationMode.BICUBIC: 3,
+    InterpolationMode.NEAREST_EXACT: 0,
```
pmeier, line 55: The only instance of where we want the reverse mapping, i.e. from `int` to `InterpolationMode`, comes from the `_interpolation_modes_from_int` function above. AFAIK, this is only needed until the deprecation period for passing `int`'s is over and will be removed afterwards. Is that true?

If yes, I think it is ok to also assign `0` as the Pillow interpolation mode here as we already do for `InterpolationMode.NEAREST`. For the Pillow side this makes sense, since they only have one nearest interpolation. 

The only downside to this is that users cannot access `InterpolationMode.NEAREST_EXACT` through and `int`. But since that is deprecated anyway, I don't think this is an issue.

pmeier:
A related and interesting question is: do we want to use `InterpolationMode.EXACT_NEAREST` for all of our mask kernels?

Right now, we ignore the input

https://github.com/pytorch/vision/blob/7d36d263a8356fac0bb363617b0c57c3bac6f89f/torchvision/prototype/features/_mask.py#L47-L55

and simply hardcode `InterpolationMode.NEAREST`:

https://github.com/pytorch/vision/blob/7d36d263a8356fac0bb363617b0c57c3bac6f89f/torchvision/prototype/transforms/functional/_geometry.py#L156

Since there were no mask kernels before, there are no BC concerns here for using `InterpolationMode.EXACT_NEAREST`.

torchvision/transforms/functional.py
datumbox: I think as long as `0` is assigned back to `InterpolationMode.NEAREST` for BC, we should be OK.

datumbox (approved):
LGTM! 

@vfdev-5 any concerns on your side?

NicolasHug:
> compared to the previous implementation of "nearest-exact"? 

This one. There's nothing to do on torchvision's side to benefit from it, it'll come from core

PR #6756, opened 2022-10-12
expand ToDtype to support multiple conversions at once

### LINKED ISSUE #6697: add support for multiple dtypes in prototype ToDtype
Right now, the transform takes a single dtype and one or multiple types to apply the dtype to:

https://github.com/pytorch/vision/blob/b482d896f448cc44fdadb030391ac12723a81546/torchvision/prototype/transforms/_misc.py#L143-L144

In case one needs multiple different dtypes, you need to use multiple transforms, e.g.

```py
    [
        transforms.ToDtype(torch.uint8, features.Image),
        transforms.ToDtype(torch.float32, features.BoundingBox),
        transforms.ToDtype(torch.int64, features.Label),
    ]
)
```

Not only has this runtime implications since we need to recurse three times through the same sample, it would also be better UI, if `ToDtype` accepted a mapping specifying the different dtypes

```py
transform = transforms.ToDtype(
    {
        features.Image: torch.uint8,
        features.BoundingBox: torch.float32,
        features.Label: torch.int64,
    }
)
```

This would also align this parameter with what we did to `fill`.

pmeier:
Closes #6697.

### DIFF
--- a/torchvision/prototype/transforms/_misc.py
+++ b/torchvision/prototype/transforms/_misc.py
@@ -1,5 +1,6 @@
-from collections import OrderedDict
-from typing import Any, Callable, Dict, OrderedDict as OrderedDictType, Sequence, Type, Union
+import functools
+from collections import defaultdict
+from typing import Any, Callable, Dict, Sequence, Type, Union
 
 import PIL.Image
 
@@ -141,24 +142,21 @@ class GaussianBlur(Transform):
 
 
 class ToDtype(Transform):
-    def __init__(self, dtype: Union[torch.dtype, OrderedDictType[Type, torch.dtype]]) -> None:
+    _transformed_types = (torch.Tensor,)
+
+    def _default_dtype(self, dtype: torch.dtype) -> torch.dtype:
+        return dtype
+
+    def __init__(self, dtype: Union[torch.dtype, Dict[Type, torch.dtype]]) -> None:
         super().__init__()
-        if isinstance(dtype, torch.dtype):
-            dtype = OrderedDict([(torch.Tensor, dtype)])
+        if not isinstance(dtype, dict):
+            # This weird looking construct only exists, since `lambda`'s cannot be serialized by pickle.
+            # If it were possible, we could replace this with `defaultdict(lambda: dtype)`
+            dtype = defaultdict(functools.partial(self._default_dtype, dtype))
         self.dtype = dtype
 
     def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
-        input_type = type(inpt)
-        if input_type in self.dtype:
-            dtype = self.dtype[input_type]
-        else:
-            for to_type, dtype in self.dtype.items():
-                if issubclass(input_type, to_type):
-                    break
-            else:
-                return inpt
-
-        return inpt.to(dtype)
+        return inpt.to(self.dtype[type(inpt)])
 
 
 class RemoveSmallBoundingBoxes(Transform):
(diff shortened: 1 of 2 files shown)


torchvision/prototype/transforms/_misc.py
```diff
@@ -140,14 +140,25 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
         return F.gaussian_blur(inpt, self.kernel_size, **params)
 
 
-# TODO: Enhance as described at https://github.com/pytorch/vision/issues/6697
-class ToDtype(Lambda):
-    def __init__(self, dtype: torch.dtype, *types: Type) -> None:
+class ToDtype(Transform):
+    def __init__(self, dtype: Union[torch.dtype, OrderedDictType[Type, torch.dtype]]) -> None:
```
pmeier, line 144: We need the `OrderedDict` here since the order matters in case we find no exact type match. See below for details.

torchvision/prototype/transforms/_misc.py
```diff
@@ -140,14 +140,25 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
...
+    def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
+        input_type = type(inpt)
+        if input_type in self.dtype:
+            dtype = self.dtype[input_type]
+        else:
+            for to_type, dtype in self.dtype.items():
+                if issubclass(input_type, to_type):
+                    break
+            else:
+                return inpt
```
pmeier, line 159: This firsts tests if we have an exact type match. If we don't, we go through the dictionary and see if we find a subclass match and use that. This is useful in case someone wants to just convert all tensors to a specific `dtype`. Since we can't know what exact types we might encounter at runtime, we use `torch.Tensor` as default type that will match all tensors at the second stage above even if the tensors are features.

torchvision/prototype/transforms/_misc.py
datumbox: Why is this not a problem for `fill`? Should we have `fill` also using an `OrderedDict`?

torchvision/prototype/transforms/_misc.py
```diff
@@ -140,14 +140,25 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
...
 
 
-# TODO: Enhance as described at https://github.com/pytorch/vision/issues/6697
-class ToDtype(Lambda):
-    def __init__(self, dtype: torch.dtype, *types: Type) -> None:
+class ToDtype(Transform):
+    def __init__(self, dtype: Union[torch.dtype, OrderedDictType[Type, torch.dtype]]) -> None:
+        super().__init__()
+        if isinstance(dtype, torch.dtype):
+            dtype = OrderedDict([(torch.Tensor, dtype)])
```
datumbox, line 147: `fill` uses default dicts here. Should we align?

torchvision/prototype/transforms/_misc.py
datumbox: Why not go for direct types as we have on `fill`? I'm a bit skeptical for adding this extra logic here as it diverges from what we do elsewhere. I think we should align by just checking for `type(inpt)` and leaving it to the user to define things properly. WDYT?

torchvision/prototype/transforms/_misc.py
pmeier: No. `fill` has a default value whereas `dtype` has none.

pmeier:
@datumbox You were right. I made it too complex for no apparent reason. I simplified. PTAL.

datumbox (approved):
LGTM, thanks!

PR #6757, opened 2022-10-12
[proto] Performance improvements for equalize op

vfdev-5:
Description:

- vectorized version of equalize_image_tensor. This is a great job done by @lezcano, thanks a lot for your help ! 🎉
- Speed up on CPU and CUDA
```
[---------------- equalize_image_tensor transforms measurements ---------------]
                                         |    main   |    new  
1 threads: ----------------------------------------------------
      <class 'torch.Tensor'> Image data  |  978.917  |  335.936

```
```
Timestamp: 20221012-160208
Torch version: 1.14.0.dev20221010+cu116
Torchvision version: 0.15.0a0
Num threads: 1

Time benchmark: RandomEqualize (1.0,) None
V2: RandomEqualize(p=1.0) torchvision.prototype.transforms._color
Stable: RandomEqualize(p=1.0) torchvision.transforms.transforms

[- RandomEqualize transforms measurements -]
                         |  stable  |    v2 
1 threads: ---------------------------------
      Tensor Image data  |  2.891   |  2.266

```

cc @datumbox @NicolasHug

### DIFF
--- a/test/test_prototype_transforms_functional.py
+++ b/test/test_prototype_transforms_functional.py
@@ -1037,3 +1037,14 @@ def test_to_image_pil(inpt, mode):
     assert isinstance(output, PIL.Image.Image)
 
     assert np.asarray(inpt).sum() == np.asarray(output).sum()
+
+
+def test_equalize_image_tensor_edge_cases():
+    inpt = torch.zeros(3, 200, 200, dtype=torch.uint8)
+    output = F.equalize_image_tensor(inpt)
+    torch.testing.assert_close(inpt, output)
+
+    inpt = torch.zeros(5, 3, 200, 200, dtype=torch.uint8)
+    inpt[..., 100:, 100:] = 1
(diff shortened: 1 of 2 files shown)


datumbox (approved):
LGTM, thank you @vfdev-5 and @lezcano for making this fast! It's interesting that we can achieve this kind of performance optimization on the frontend without calling the native histogram methods. There might be some speed improvements to be achieved on the future on Core.

I see that all the tests pass, so this looks good to go.

vfdev-5:
@datumbox speaking with Mario, we can improve the code a bit more, I'll send an update in a follow-up PR.

lezcano:
>It's interesting that we can achieve this kind of performance optimization on the frontend without calling the native histogram methods

For reference, note that the speed-ups do not come from the histogram method itself. I just used that way of implementing a histogram because it allows to have batches. The main speed-up comes from using batched operations all across vs having a for loop and running all these kernels once per channel.

Issue #6750, opened 2022-10-12
allow `closeness_kwargs` to be set more specifically


pmeier:
In our prototype transforms tests, we have the ability to set `closeness_kwargs` 

https://github.com/pytorch/vision/blob/11a2eeda8fb127a7ad72b4c98ca918b93055c1e7/test/prototype_common_utils.py#L637-L638

that will be passed to `assert_close` in addition to the values to be compared:

https://github.com/pytorch/vision/blob/11a2eeda8fb127a7ad72b4c98ca918b93055c1e7/test/test_prototype_transforms_functional.py#L94

However, there is currently no option to specify the `closeness_kwargs` more specifically. Meaning, they will be applied to all tests regardless of the `dtype` and `device`. Especially for the `dtype`, this is not sufficient: images in `torch.uint8` might be allowed an absolute tolerance of `1` since that is a minor difference. However, applying the same tolerance to an `torch.float32` image would mean that any difference in the valid value range of `[0, 1]` would be fine and thus rendering the test almost useless. Thus, we use the stricter `torch.float32` tolerance here, but that leads to flaky tests #6741.

We already have the `TestMark` functionality in, that allows us to apply `pytest.mark.*` to specific tests based on the inputs. We should extend this to also specify the tolerances or even the whole `closeness_kwargs`.

cc @vfdev-5 @datumbox @bjuncek

PR #6726, opened 2022-10-10
enable arbitrary batch size for all prototype kernels

pmeier:
Fixes #6670. This basically only moves the squashing that we did for the video kernels into the image kernels.

If we start to merge transforms v1 and v2, we should support this better. I haven't checked, but I guesstimate that 90% - 100% of all times we need to squash / unsquash is because we use these three ops internally:

1. https://github.com/pytorch/vision/blob/6e203b44098c3371689f56abc17b7c02bd51a261/torchvision/transforms/functional_tensor.py#L469
2. https://github.com/pytorch/vision/blob/6e203b44098c3371689f56abc17b7c02bd51a261/torchvision/transforms/functional_tensor.py#L562
3. https://github.com/pytorch/vision/blob/6e203b44098c3371689f56abc17b7c02bd51a261/torchvision/transforms/functional_tensor.py#L763

I propose, we implement `_nd_{interpolate, grid_sample, conv_2d}` as internal utilities, so we don't have to have the arbitrary batch compatibility code everywhere. This is somewhat already implemented for 2. with

https://github.com/pytorch/vision/blob/6e203b44098c3371689f56abc17b7c02bd51a261/torchvision/transforms/functional_tensor.py#L547

PR #6762, opened 2022-10-13
[proto] Small optimization for gaussian_blur functional op

vfdev-5:
Description:
- use softmax and multiplication to speed-up gaussian blur op

```
[-------------- Gaussian Blur cpu torch.uint8 --------------]
                     |  gaussian_blur v1  |  gaussian_blur v2
1 threads: --------------------------------------------------
      (3, 400, 400)  |        2.78        |        2.65
6 threads: --------------------------------------------------
      (3, 400, 400)  |        1.08        |        1.01

[-------------- Gaussian Blur cuda torch.uint8 -------------]
                     |  gaussian_blur v1  |  gaussian_blur v2
1 threads: --------------------------------------------------
      (3, 400, 400)  |        342         |        227
6 threads: --------------------------------------------------
      (3, 400, 400)  |        259         |        227

[------------- Gaussian Blur cpu torch.float32 -------------]
                     |  gaussian_blur v1  |  gaussian_blur v2
1 threads: --------------------------------------------------
      (3, 400, 400)  |        2160        |        2080
6 threads: --------------------------------------------------
      (3, 400, 400)  |         917        |         840

[------------- Gaussian Blur cuda torch.float32 ------------]
                     |  gaussian_blur v1  |  gaussian_blur v2
1 threads: --------------------------------------------------
      (3, 400, 400)  |        232         |        200
6 threads: --------------------------------------------------
      (3, 400, 400)  |        231         |        200

```
[Code source](https://github.com/vfdev-5/tvapiv2_benchmarks/blob/master/check_gaussian_blur.py)

---

This PR may also help with flaky gaussian blur tests.

Context:

Gaussian blur eager vs jit tests are flaky for uint8 input. This may be due to the fact that we cast to float32, perform the opertation in float32 and cast back to uint8. We generate conv kernels (created using exp op) in float32 and maybe accumulating precision errors.

Refs: https://github.com/pytorch/vision/pull/6755

### DIFF
--- a/test/test_prototype_transforms_consistency.py
+++ b/test/test_prototype_transforms_consistency.py
@@ -308,25 +308,28 @@ CONSISTENCY_CONFIGS = [
             ArgsKwargs(brightness=0.1, contrast=0.4, saturation=0.7, hue=0.3),
         ],
     ),
-    ConsistencyConfig(
-        prototype_transforms.ElasticTransform,
-        legacy_transforms.ElasticTransform,
-        [
-            ArgsKwargs(),
-            ArgsKwargs(alpha=20.0),
-            ArgsKwargs(alpha=(15.3, 27.2)),
-            ArgsKwargs(sigma=3.0),
-            ArgsKwargs(sigma=(2.5, 3.9)),
-            ArgsKwargs(interpolation=prototype_transforms.InterpolationMode.NEAREST),
-            ArgsKwargs(interpolation=prototype_transforms.InterpolationMode.BICUBIC),
-            ArgsKwargs(fill=1),
-        ],
-        # ElasticTransform needs larger images to avoid the needed internal padding being larger than the actual image
-        make_images_kwargs=dict(DEFAULT_MAKE_IMAGES_KWARGS, sizes=[(163, 163), (72, 333), (313, 95)]),
-        # We updated gaussian blur kernel generation with a faster and numerically more stable version
-        # This brings float32 accumulation visible in elastic transform -> we need to relax consistency tolerance
-        closeness_kwargs={"rtol": 1e-1, "atol": 1},
-    ),
+    *[
+        ConsistencyConfig(
+            prototype_transforms.ElasticTransform,
+            legacy_transforms.ElasticTransform,
+            [
+                ArgsKwargs(),
+                ArgsKwargs(alpha=20.0),
+                ArgsKwargs(alpha=(15.3, 27.2)),
+                ArgsKwargs(sigma=3.0),
+                ArgsKwargs(sigma=(2.5, 3.9)),
+                ArgsKwargs(interpolation=prototype_transforms.InterpolationMode.NEAREST),
+                ArgsKwargs(interpolation=prototype_transforms.InterpolationMode.BICUBIC),
+                ArgsKwargs(fill=1),
+            ],
+            # ElasticTransform needs larger images to avoid the needed internal padding being larger than the actual image
+            make_images_kwargs=dict(DEFAULT_MAKE_IMAGES_KWARGS, sizes=[(163, 163), (72, 333), (313, 95)], dtypes=[dt]),
+            # We updated gaussian blur kernel generation with a faster and numerically more stable version
+            # This brings float32 accumulation visible in elastic transform -> we need to relax consistency tolerance
+            closeness_kwargs=ckw,
+        )
+        for dt, ckw in [(torch.uint8, {"rtol": 1e-1, "atol": 1}), (torch.float32, {"rtol": 1e-3, "atol": 1e-5})]
+    ],
     ConsistencyConfig(
         prototype_transforms.GaussianBlur,
         legacy_transforms.GaussianBlur,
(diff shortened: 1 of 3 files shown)


vfdev-5:
Successful runs on ubuntu-latest:
- https://github.com/pytorch/vision/actions/runs/3241542307/attempts/1
- https://github.com/pytorch/vision/actions/runs/3241542307/attempts/2

test/test_prototype_transforms_consistency.py
```diff
@@ -323,6 +323,9 @@ def __init__(
         ],
         # ElasticTransform needs larger images to avoid the needed internal padding being larger than the actual image
         make_images_kwargs=dict(DEFAULT_MAKE_IMAGES_KWARGS, sizes=[(163, 163), (72, 333), (313, 95)]),
+        # We updated gaussian blur kernel generation with a faster and numerically more stable version
+        # This brings float32 accumulation visible in elastic transform -> we need to relax consistency tolerance
+        closeness_kwargs={"rtol": 1e-1, "atol": 1},
```
pmeier, line 328: The `atol` part is problematic here since it basically renders the `float32` test useless: with `atol=1`, every float image within the interval `[0.0, 1.0]` will match regardless of a working kernel or not. I currently don't have a good answer for this problem. See #6750.

I see two options now:

1. Accept that we can't test `float32` for now.
2. Duplicate this whole configuration and set `make_images_kwargs=dict(..., dtypes=[torch.uint8])` and `make_images_kwargs=dict(..., dtypes=[torch.float32])` with the correct tolerances respectively.

torchvision/prototype/transforms/functional/_misc.py
```diff
@@ -70,7 +87,18 @@ def gaussian_blur_image_tensor(
...
+    dtype = image.dtype if torch.is_floating_point(image) else torch.float32
+    kernel = _get_gaussian_kernel2d(kernel_size, sigma, dtype=dtype, device=image.device)
+    kernel = kernel.expand(image.shape[-3], 1, kernel.shape[0], kernel.shape[1])
+
+    image, need_cast, need_squeeze, out_dtype = _FT._cast_squeeze_in(image, [kernel.dtype])
+
+    # padding = (left, right, top, bottom)
+    padding = [kernel_size[0] // 2, kernel_size[0] // 2, kernel_size[1] // 2, kernel_size[1] // 2]
+    output = _FT.torch_pad(image, padding, mode="reflect")
+    output = _FT.conv2d(output, kernel, groups=output.shape[-3])
```
pmeier, line 99: These two are just imported from `torch`, right? If so, can we also do that here instead importing from the legacy transforms? 

Plus, regarding the `conv2d`, could we maybe implement a common `nd_conv2d` as proposed in https://github.com/pytorch/vision/pull/6726#issue-1402722696? Basically, `nd_conv2d` should handle the `_cast_squeeze_in` and `_cast_squeeze_out` so the actual kernel is cleaner. That way, all kernels that use `conv2d` now could use that utility and don't need to worry about shape or dtype.

torchvision/prototype/transforms/functional/_misc.py
```diff
@@ -32,6 +33,22 @@ def normalize(
...
+    kernel1d = torch.softmax(-x.pow_(2), dim=0)
+    return kernel1d
+
+
+def _get_gaussian_kernel2d(
+    kernel_size: List[int], sigma: List[float], dtype: torch.dtype, device: torch.device
+) -> torch.Tensor:
+    kernel1d_x = _get_gaussian_kernel1d(kernel_size[0], sigma[0]).to(device, dtype=dtype)
+    kernel1d_y = _get_gaussian_kernel1d(kernel_size[1], sigma[1]).to(device, dtype=dtype)
+    kernel2d = kernel1d_y.unsqueeze(-1) * kernel1d_x
```
pmeier, line 48: Doesn't this have to be a matmul?

vfdev-5: This one is faster and gives required output

pmeier: Yeah, since we do something like `(N, 1) @ (1, N)` there are no additions and thus regular broadcasting works here. Thanks!

torchvision/prototype/transforms/functional/_misc.py
vfdev-5: OK, I can import methods directly. 

As for `nd_conv(2d?)`, I do not think we always have a situation when we can hide _cast_squeeze_in, _cast_squeeze_out inside `nd_*` op... Anyway, let's see that elsewhere (not this PR).

test/test_prototype_transforms_consistency.py
vfdev-5: How about this https://github.com/pytorch/vision/pull/6762/files#diff-bf36f728ca93b501ac1ae0dd786c891c9c389cab57e4843cbd5d3b93a1b95068R311 ?

pmeier (approved):
LGTM if CI is green!

PR #6763, opened 2022-10-13
improve test id for consistency tests

pmeier:
As requested by @vfdev-5 offline. In the past the `ArgsKwargs` object implemented a `__repr__` method, so the old id made sense at the time.

### DIFF
--- a/test/test_prototype_transforms_consistency.py
+++ b/test/test_prototype_transforms_consistency.py
@@ -575,9 +575,11 @@ def check_call_consistency(prototype_transform, legacy_transform, images=None, s
 @pytest.mark.parametrize(
     ("config", "args_kwargs"),
     [
-        pytest.param(config, args_kwargs, id=f"{config.legacy_cls.__name__}({args_kwargs})")
+        pytest.param(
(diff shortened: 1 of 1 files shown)


test/test_prototype_transforms_consistency.py
```diff
@@ -575,9 +575,11 @@ def check_call_consistency(prototype_transform, legacy_transform, images=None, s
 @pytest.mark.parametrize(
     ("config", "args_kwargs"),
     [
-        pytest.param(config, args_kwargs, id=f"{config.legacy_cls.__name__}({args_kwargs})")
+        pytest.param(
+            config, args_kwargs, id=f"{config.legacy_cls.__name__}-{idx:0{len(str(len(config.args_kwargs)))}d}"
```
pmeier, line 579: The numerical id looks weird, but this actually makes sense. As implemented, we always get a zero-padded number depending on the number of runs. For example, `test_call_consistency[RandomCrop-00]` vs. `test_call_consistency[RandomCrop-0]`. Although this seems minor, without this, it is impossible to select the `0`th test with `-k 0` while at the same time pulling in `test_call_consistency[RandomCrop-10]` or any other test that also includes a `0` in its id. With this change you can do `-k00` to select only the one test you want.

vfdev-5 (approved):
Thanks!

PR #6767, opened 2022-10-14
only flatten a pytree once

### LINKED ISSUE #6760: Avoid multiple pytree flattenings inside the prototype transforms
By design, our transforms v2 can handle arbitrary input structures. Internally, we are using `torch.utils._pytree` for it.

## Status quo

The simplest transforms flatten / unflatten only once:

https://github.com/pytorch/vision/blob/54a2d4e8f7a4568823532d4342f6ba13e7339dce/torchvision/prototype/transforms/_transform.py#L37-L42

However, albeit hidden, most transforms flatten at least twice if not multiple times:

- If a transform needs to know the spatial size to compute its `params`

  https://github.com/pytorch/vision/blob/54a2d4e8f7a4568823532d4342f6ba13e7339dce/torchvision/prototype/transforms/_geometry.py#L108 

  the extraction logic flattens again

  https://github.com/pytorch/vision/blob/54a2d4e8f7a4568823532d4342f6ba13e7339dce/torchvision/prototype/transforms/_utils.py#L101-L102

- If a transform performs some checks on the sample before transforming 
  
  https://github.com/pytorch/vision/blob/54a2d4e8f7a4568823532d4342f6ba13e7339dce/torchvision/prototype/transforms/_geometry.py#L188-L190

  the checking utility flattens again

  https://github.com/pytorch/vision/blob/54a2d4e8f7a4568823532d4342f6ba13e7339dce/torchvision/prototype/transforms/_utils.py#L124-L125

This of course has some performance implications that can be avoided.

## Proposal

In all of the cases where we perform the extra flattening are happening in internal and thus not user-facing methods. Thus, instead of keeping the option to operate on arbitrary input structures on our utilities, we could have them just work on already flattened inputs. This would avoid the repeated `tree_flatten` calls inside of them.

For our transforms that means two changes:

1. The extra calls in `_get_params` can be avoided by simply flattening before its call

   https://github.com/pytorch/vision/blob/54a2d4e8f7a4568823532d4342f6ba13e7339dce/torchvision/prototype/transforms/_transform.py#L35-L37

2. The extra calls in overridden `forward`'s require more boilerplate code. Basically, each transform that overrides `forward` needs to perform the flattening / unflattening themselves since the check utilities are called before the `super().forward(...)` call. This also means that we are technically still flattening twice, although the second time inside `super().forward(...)` does nothing.

   However, there is #6503 that introduces a common interface for the checks. IIRC, we never followed up on it, since we eliminated some boilerplate in the overridden `forward` in #6504. Since this proposal would re-add some boilerplate for performance gains, we could pick #6503 up again. If we do, this leaves very few, objectively "outlier" transforms that would need to have this boiler plate. 

   If we go for the common check interface, it could receive the already flattened sample as well similar to what was proposed in 1.

## Conclusion

Flattening an input sample multiple times inside a single transformation has no benefits while slowing down execution. This issue proposes a way to avoid this while keeping the UI as convenient as it is.

pmeier:
Closes #6760 minus the container handling as stated in https://github.com/pytorch/vision/issues/6760#issuecomment-1278583221.

### DIFF
--- a/test/test_prototype_transforms.py
+++ b/test/test_prototype_transforms.py
@@ -437,7 +437,7 @@ class TestRandomZoomOut:
         image = mocker.MagicMock(spec=features.Image)
         h, w = image.spatial_size = (24, 32)
 
-        params = transform._get_params(image)
+        params = transform._get_params([image])
 
         assert len(params["padding"]) == 4
         assert 0 <= params["padding"][0] <= (side_range[1] - 1) * w
@@ -462,7 +462,7 @@ class TestRandomZoomOut:
         _ = transform(inpt)
         torch.manual_seed(12)
         torch.rand(1)  # random apply changes random state
-        params = transform._get_params(inpt)
+        params = transform._get_params([inpt])
 
         fill = transforms.functional._geometry._convert_fill_arg(fill)
         fn.assert_called_once_with(inpt, **params, fill=fill)
@@ -623,7 +623,7 @@ class TestRandomAffine:
         h, w = image.spatial_size
 
         transform = transforms.RandomAffine(degrees, translate=translate, scale=scale, shear=shear)
-        params = transform._get_params(image)
+        params = transform._get_params([image])
 
         if not isinstance(degrees, (list, tuple)):
             assert -degrees <= params["angle"] <= degrees
@@ -690,7 +690,7 @@ class TestRandomAffine:
         torch.manual_seed(12)
         _ = transform(inpt)
         torch.manual_seed(12)
-        params = transform._get_params(inpt)
+        params = transform._get_params([inpt])
 
         fill = transforms.functional._geometry._convert_fill_arg(fill)
         fn.assert_called_once_with(inpt, **params, interpolation=interpolation, fill=fill, center=center)
@@ -722,7 +722,7 @@ class TestRandomCrop:
         h, w = image.spatial_size
 
         transform = transforms.RandomCrop(size, padding=padding, pad_if_needed=pad_if_needed)
-        params = transform._get_params(image)
+        params = transform._get_params([image])
 
         if padding is not None:
             if isinstance(padding, int):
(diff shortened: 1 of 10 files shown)


test/test_prototype_transforms.py
```diff
@@ -437,7 +437,7 @@ def test__get_params(self, fill, side_range, mocker):
         image = mocker.MagicMock(spec=features.Image)
         h, w = image.spatial_size = (24, 32)
 
-        params = transform._get_params(image)
+        params = transform._get_params([image])
```
pmeier, line 440: All changes in this file stem from the fact that `_get_params` previously handled pytree objects, but now only handles flattened ones.

test/test_prototype_transforms.py
```diff
@@ -1197,6 +1197,7 @@ def test_assertions(self, transform_cls):
         [
             [transforms.Pad(2), transforms.RandomCrop(28)],
             [lambda x: 2.0 * x, transforms.Pad(2), transforms.RandomCrop(28)],
+            [transforms.Pad(2), lambda x: 2.0 * x, transforms.RandomCrop(28)],
```
pmeier, line 1200: I've accidentally included this while playing with the container flattening. It is useful nevertheless to have the "foreign" transform also in the middle of the pipeline. Let me know if I should revert.

test/test_prototype_transforms_consistency.py
```diff
@@ -639,7 +639,7 @@ def test_random_apply(self, p):
         prototype_transform = prototype_transforms.RandomApply(
             [
                 prototype_transforms.Resize(256),
-                legacy_transforms.CenterCrop(224),
```
pmeier, line 642: This was just plain wrong. It never triggered because the input is a single image. Found this while playing with the container flattening.

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -31,16 +31,17 @@ def _get_random_item(self, dct: Dict[K, V]) -> Tuple[K, V]:
         key = keys[int(torch.randint(len(keys), ()))]
         return key, dct[key]
 
-    def _extract_image_or_video(
+    def _flatten_and_extract_image_or_video(
```
pmeier, line 34: I've opted to group the flattening and extraction as well as the unflattening and insertion since that yielded the cleanest results.

datumbox (approved):
LGTM @pmeier. I like the approach. It keeps this detail internal and minimizes the overhead. Let's see how this improved our speed in real world applications. I'll do some training later to confirm.

PR #6770, opened 2022-10-14
Extend `RandomShortestSize` to support Video specific flavour of the augmentation

datumbox:
Our `RandomShortestSize` implementation on the references is developed with Object Detection in mind. Videos require a slight variation (see reference implementation for [detection](https://github.com/facebookresearch/detectron2/blob/cf35922d74559fd46928dc787c1c312eb4195b41/detectron2/data/transforms/augmentation_impl.py#L129) vs [video](https://github.com/facebookresearch/pytorchvideo/blob/f4242c29a95443f23128437c0c37378069c32ffb/pytorchvideo/transforms/transforms.py#L123)).

This PR extends the transform in a BC-way so that it can support both. In particular, we make the `max_size` optional and this allows us to reparameterize the transform for videos as such:
```python
from torchvision.prototype.transforms import RandomShortestSize
import math
import torch

x = torch.randn(7, 11, 3, 450, 800)
t = RandomShortestSize(list(range(256, 320+1)))
z = t(x)
print(z.shape)

size = min(z.shape[-2:])

_, t, c, h, w = x.shape
if w < h:
    new_h = int(math.floor((float(h) / w) * size))
    new_w = size
else:
    new_h = size
    new_w = int(math.floor((float(w) / h) * size))

print(new_h, new_w)

assert (new_h, new_w) == tuple(z.shape[-2:])
```

Though the names of `min_size` and `max_size` are confusing (better names would have been `shortside_min_size_range` and `longside_max_size`), their semantics align with the arguments that `F.resize()` has for `size` and and `max_size`. On the latter the default value of `max_size` (which again applies to the longest edge) is `None`, so this transform uses the same semantics and default values as in other places of the API.

### DIFF
--- a/test/test_prototype_transforms.py
+++ b/test/test_prototype_transforms.py
@@ -1379,10 +1379,9 @@ class TestScaleJitter:
 
 
 class TestRandomShortestSize:
-    def test__get_params(self, mocker):
+    @pytest.mark.parametrize("min_size,max_size", [([5, 9], 20), ([5, 9], None)])
+    def test__get_params(self, min_size, max_size, mocker):
         spatial_size = (3, 10)
-        min_size = [5, 9]
-        max_size = 20
 
         transform = transforms.RandomShortestSize(min_size=min_size, max_size=max_size)
 
@@ -1395,10 +1394,9 @@ class TestRandomShortestSize:
         assert isinstance(size, tuple) and len(size) == 2
 
         longer = max(size)
-        assert longer <= max_size
-
         shorter = min(size)
-        if longer == max_size:
+        if max_size is not None and longer == max_size:
+            assert longer <= max_size
             assert shorter <= max_size
         else:
             assert shorter in min_size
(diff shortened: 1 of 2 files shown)


vfdev-5 (approved):
OK to me, thanks @datumbox !

test/test_prototype_transforms.py
```diff
@@ -1395,10 +1394,9 @@ def test__get_params(self, mocker):
         assert isinstance(size, tuple) and len(size) == 2
 
         longer = max(size)
-        assert longer <= max_size
-
         shorter = min(size)
-        if longer == max_size:
+        if max_size is not None and longer == max_size:
```
vfdev-5, line 1398: This is condition seems strange `and longer == max_size`

datumbox: Yeap, not sure why this existed before. Fixed. Still passes locally on multiple runs.

Issue #6603, opened 2022-09-18
Read videos from memory


vedantroy:
### 🚀 The feature

I have a lot of small videos (3 million video files), in an attempt to reduce disk bottleneck, I store all the videos in LMDB using: https://lmdb.readthedocs.io/en/release/. I would love the ability to read videos from memory.

### Motivation, pitch

My training (8xA100 GPUs) process is data starved. I'm pretty sure the issue is disk throughput, hence my usage of LMDB. Allowing torchvision to read videos from memory would help me use LMDB for storing binary files.

### Alternatives

I am about to try manually doing this using pyav.

pmeier:
Could you clarify what you mean by "read videos from memory"? Do you want to a dataset that allows you to read data from a LMDB? 

If yes, you can have a look at the [LSUN dataset](https://github.com/pytorch/vision/blob/main/torchvision/datasets/lsun.py) that stores images inside multiple LMDBs. There is probably room for optimization, but this should get you started.

YosuaMichael:
cc @jdsgomes

jdsgomes:
Hi @vedantroy thank you for the feature request. I am actually planning to add this functionality to our VideoReader. I will keep you posted with progress but the ETA is about 2 weeks.

PR #6771, opened 2022-10-14
Read video from memory newapi

jdsgomes:
adds read from memory functionality in the new video API.

Addresses #6603

### DIFF
--- a/test/test_videoapi.py
+++ b/test/test_videoapi.py
@@ -76,8 +76,14 @@ class TestVideoApi:
 
                 # compare the frames and ptss
                 for i in range(len(vr_frames)):
-                    assert av_pts[i] == vr_pts[i]
-                    torch.test.assert_equal(av_frames[i], vr_frames[i])
+                    assert float(av_pts[i]) == approx(vr_pts[i], abs=0.1)
+
+                    mean_delta = torch.mean(torch.abs(av_frames[i].float() - vr_frames[i].float()))
+                    # on average the difference is very small and caused
+                    # by decoding (around 1%)
+                    # TODO: asses empirically how to set this? atm it's 1%
+                    # averaged over all frames
+                    assert mean_delta.item() < 2.55
 
                 del vr_frames, av_frames, vr_pts, av_pts
 
--- a/test/test_videoapi.py
+++ b/test/test_videoapi.py
@@ -130,11 +130,14 @@ class TestVideoApi:
             vr_pts.append(vr_frame["pts"])
 
         # get vr frames = read from memory
-        with open(full_path, "rb") as f:
-            video_reader_from_mem = VideoReader(f.read(), "video")
-            for vr_frame_from_mem in video_reader_from_mem:
-                vr_frames_mem.append(vr_frame_from_mem["data"])
-                vr_pts_mem.append(vr_frame_from_mem["pts"])
+        f = open(full_path, "rb")
+        fbytes = f.read()
+        f.close()
+        video_reader_from_mem = VideoReader(fbytes, stream)
+
+        for vr_frame_from_mem in video_reader_from_mem:
+            vr_frames_mem.append(vr_frame_from_mem["data"])
+            vr_pts_mem.append(vr_frame_from_mem["pts"])
 
         # same number of frames
         assert len(vr_frames) == len(vr_frames_mem)
@@ -143,8 +146,12 @@ class TestVideoApi:
         # compare the frames and ptss
         for i in range(len(vr_frames)):
             assert vr_pts[i] == vr_pts_mem[i]
-
-            torch.equal(vr_frames[i], vr_frames_mem[i])
+            mean_delta = torch.mean(torch.abs(vr_frames[i].float() - vr_frames_mem[i].float()))
+            # on average the difference is very small and caused
+            # by decoding (around 1%)
+            # TODO: asses empirically how to set this? atm it's 1%
+            # averaged over all frames
+            assert mean_delta.item() < 2.55
 
         del vr_frames, vr_pts, vr_frames_mem, vr_pts_mem
 
--- a/test/test_videoapi.py
+++ b/test/test_videoapi.py
@@ -57,8 +57,8 @@ class TestVideoApi:
         full_path = os.path.join(VIDEO_DIR, test_video)
         with av.open(full_path) as av_reader:
             if av_reader.streams.video:
-                av_frames, vr_frames = [], []
-                av_pts, vr_pts = [], []
+                av_frames, vr_frames, vr_frames_mem = [], [], []
+                av_pts, vr_pts, vr_pts_mem = [], [], []
                 # get av frames
                 for av_frame in av_reader.decode(av_reader.streams.video[0]):
                     av_frames.append(torch.tensor(av_frame.to_rgb().to_ndarray()).permute(2, 0, 1))
@@ -70,19 +70,31 @@ class TestVideoApi:
                     vr_frames.append(vr_frame["data"])
                     vr_pts.append(vr_frame["pts"])
 
+                # get vr frames = read from memory
+                with open(full_path, "rb") as f:
+                    video_reader_from_mem = VideoReader(f.read())
+                    for vr_frame_from_mem in video_reader_from_mem:
+                        vr_frames_mem.append(vr_frame_from_mem["data"])
+                        vr_pts_mem.append(vr_frame_from_mem["pts"])
+
                 # same number of frames
-                assert len(vr_frames) == len(av_frames)
-                assert len(vr_pts) == len(av_pts)
+                assert len(vr_frames) == len(av_frames) == len(vr_frames_mem)
+                assert len(vr_pts) == len(av_pts) == len(vr_pts_mem)
 
                 # compare the frames and ptss
                 for i in range(len(vr_frames)):
+                    assert vr_pts[i] == vr_pts_mem[i]
                     assert float(av_pts[i]) == approx(vr_pts[i], abs=0.1)
+
+                    torch.equal(vr_frames[i], vr_frames_mem[i])
+
                     mean_delta = torch.mean(torch.abs(av_frames[i].float() - vr_frames[i].float()))
                     # on average the difference is very small and caused
                     # by decoding (around 1%)
                     # TODO: asses empirically how to set this? atm it's 1%
                     # averaged over all frames
                     assert mean_delta.item() < 2.55
+
 
                 del vr_frames, av_frames, vr_pts, av_pts
 
--- a/test/test_videoapi.py
+++ b/test/test_videoapi.py
@@ -76,14 +76,8 @@ class TestVideoApi:
 
                 # compare the frames and ptss
                 for i in range(len(vr_frames)):
-                    assert float(av_pts[i]) == approx(vr_pts[i], abs=0.1)
-
-                    mean_delta = torch.mean(torch.abs(av_frames[i].float() - vr_frames[i].float()))
-                    # on average the difference is very small and caused
-                    # by decoding (around 1%)
-                    # TODO: asses empirically how to set this? atm it's 1%
-                    # averaged over all frames
-                    assert mean_delta.item() < 2.55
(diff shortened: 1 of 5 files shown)


torchvision/io/video_reader.py
```diff
@@ -72,7 +72,11 @@ class VideoReader:
 
     Args:
 
-        path (string): Path to the video file in supported format
+        src (string, bytes object, or tensor): The media source.
```
jdsgomes, line 75: Possibly it is a bad idea to change this, despite being a beta API, but leaving it as path because misleading too.. I am open for suggestions from the reviewers

YosuaMichael: Since our Beta API has no warning (which make few users aware of it), I think we should follow deprecation policy for BC changes. I suggest we follow similar approach to transform to deprecate `resample` param: https://github.com/pytorch/vision/blob/21790df9a4f77bd9ec4db44de04594cb539457a7/torchvision/transforms/transforms.py#L1275-L1278

Basically we allow both `path` and `src` (put `path` as optional and the last param), and give warning for those who use `path` that it will be removed on v 0.17.

jdsgomes: But in that case if someone is using  `VideoReader(path="bla.avi")` wont be able to do it. So I am thinking that we should do the other way around, i.e., we add `src` as optional for now and in two releases it replaces `path`. lmk your thoughts

jdsgomes: Actually I don't see any good option

**Option 1: Make path optional**
This is non BC because you no longer can do `VideoReader(path="bla.avi")`. 

**Option 2: Make src optional**
This option is BC and allows users to do `VideoReader(path="bla.avi")`, `VideoReader("bla.avi")` and `VideoReader(bytesobj)` but if one wants to use the src keyword then it looks weird: `VideoReader("", src="bla.avi")`. Furthermore later wen we deprecate path it will be a BC change



**Option 3: add an extra argument permanently**
Third option  will be to add `video_obj` as last optional argument, and when this was provided then we we would ignore the `path`. This is BC but a bit ugly when we want to construct a VideoReader from memory : `VideoReader("", video_obj=bytesobj)`

Given this I am inclined for option 3, unless I am missing something

YosuaMichael: My thinking is similar with Option 2, but we let path to be optional at the end, and we set default value of src as empty string.
```
VideoReader(src: Any = "", ..., path: Optional[str] = None)
```

and inside constructor we can handle like:
```
if src == "":
  if path is None:
    raise ValueError(...)
  src = path
```

what do you think?

jdsgomes: yeah that makes a lot of  sense.  Thanks

vedantroy:
This segfaults on certain videos. Here's a partial-reproduction script (the videos are stored in a pandas dataframe):

```
import pandas as pd

# df2 = pd.read_pickle("df2.pkl")
df = pd.read_pickle("df.pkl")
# print the # of rows in the df
print(len(df))
# print the keys in the df
print(df.keys())
# print the first row in the df
print(df.iloc[0])


# print the type of the 1st value in the 1st row
first_vid = df.iloc[0][0]
print(f"Length: {len(first_vid)}")
print(f"Type: {type(first_vid)}")

# write first_vid to a file
with open("test.mp4", "wb") as f:
    f.write(first_vid)

import itertools
import copy

import torch
from torchvision.io import VideoReader
import torchvision

def clip_from_start(buf: bytes, expected_frames: int):
    # import av
    # import io
    # buffer = io.BytesIO(buf)
    # container = av.open(buffer)
    # i = 0 
    # for frame in container.decode(video=0):
    #     print(type(frame))
    #     i += 1
    #     print(i)
    #     pass

    tensor = torch.frombuffer(buf, dtype=torch.uint8)
    tensor = copy.deepcopy(tensor)
    # torchvision.io.read_video()
    rdr = VideoReader(tensor)
    sampled_frames = list(itertools.islice(iter(rdr), expected_frames))
    if len(sampled_frames) != expected_frames:
        return None
    data = []
    for frame in sampled_frames:
        data.append(frame["data"])
    return torch.stack(data, dim=0)

clip = clip_from_start(first_vid, 2)
print(clip.shape)
```

I'm working to get approval of the public copy of the data.

torchvision/io/video_reader.py
```diff
@@ -105,7 +125,24 @@ def __init__(self, path: str, stream: str = "video", num_threads: int = 0, devic
                 + "build torchvision from source."
             )
 
-        self._c = torch.classes.torchvision.Video(path, stream, num_threads)
+        if src == "":
+            if path is None:
+                raise ValueError("src cannot be empty")
```
YosuaMichael, line 130: NIT: I just realized that missing argument should raise `TypeError` instead of `ValueError`. Can we change this to `TypeError` instead?

jdsgomes: In this case I disagree, we are checking for the value of the variable, not for the type. Would like to understand your interpretation in case I am missing something.

YosuaMichael: I just tried on my local python the following:
```
>>> def foo(x, y):
...  print(x, y)
... 
>>> foo("hello")
  File "<stdin>", line 1, in <module>
TypeError: foo() missing 1 required positional argument: 'y'
```

Notice that when we dont specify 1 required positional argument, python raise `TypeError` instead of `ValueError` (I actually surprised that python use `TypeError` for this). Therefore I think we should follow this convention since in this case we basically missing 1 required argument as well.

jdsgomes:
> This segfaults on certain videos. Here's a partial-reproduction script (the videos are stored in a pandas dataframe):
> 
> ```
> import pandas as pd
> 
> # df2 = pd.read_pickle("df2.pkl")
> df = pd.read_pickle("df.pkl")
> # print the # of rows in the df
> print(len(df))
> # print the keys in the df
> print(df.keys())
> # print the first row in the df
> print(df.iloc[0])
> 
> 
> # print the type of the 1st value in the 1st row
> first_vid = df.iloc[0][0]
> print(f"Length: {len(first_vid)}")
> print(f"Type: {type(first_vid)}")
> 
> # write first_vid to a file
> with open("test.mp4", "wb") as f:
>     f.write(first_vid)
> 
> import itertools
> import copy
> 
> import torch
> from torchvision.io import VideoReader
> import torchvision
> 
> def clip_from_start(buf: bytes, expected_frames: int):
>     # import av
>     # import io
>     # buffer = io.BytesIO(buf)
>     # container = av.open(buffer)
>     # i = 0 
>     # for frame in container.decode(video=0):
>     #     print(type(frame))
>     #     i += 1
>     #     print(i)
>     #     pass
> 
>     tensor = torch.frombuffer(buf, dtype=torch.uint8)
>     tensor = copy.deepcopy(tensor)
>     # torchvision.io.read_video()
>     rdr = VideoReader(tensor)
>     sampled_frames = list(itertools.islice(iter(rdr), expected_frames))
>     if len(sampled_frames) != expected_frames:
>         return None
>     data = []
>     for frame in sampled_frames:
>         data.append(frame["data"])
>     return torch.stack(data, dim=0)
> 
> clip = clip_from_start(first_vid, 2)
> print(clip.shape)
> ```
> 
> I'm working to get approval of the public copy of the data.
> 
> In the meantime, the error is:
> 
> ```
> test.py:41: UserWarning: The given buffer is not writable, and PyTorch does not support non-writable tensors. This means you can write to the underlying (supposedly non-writable) buffer using the tensor. You may want to copy the buffer to protect its data or make it writable before converting it to a tensor. This type of warning will be suppressed for the rest of this program. (Triggered internally at ../torch/csrc/utils/tensor_new.cpp:1563.)
>   tensor = torch.frombuffer(buf, dtype=torch.uint8)
> malloc(): corrupted top size
> Aborted (core dumped)
> ```

Thanks for reporting this. 
I cannot reproduce the same error with some test videos. Is there any possibility that the data is corrupt? 
Otherwise please let me know when you would be able to share the sample video for us to verify and debug. 

Thanks

test/test_videoapi.py
```diff
@@ -114,6 +115,46 @@ def test_frame_reading(self, test_video):
...
+        for vr_frame_from_mem in video_reader_from_mem:
+            vr_frames_mem.append(vr_frame_from_mem["data"])
+            vr_pts_mem.append(vr_frame_from_mem["pts"])
+
+        # same number of frames
+        assert len(vr_frames) == len(vr_frames_mem)
+        assert len(vr_pts) == len(vr_pts_mem)
+
+        # compare the frames and ptss
+        for i in range(len(vr_frames)):
```
bjuncek, line 147: Given that this is like-for-like decoding (using verifiably the same backend), we should probably insist on either a max delta or completely equal.

torchvision/csrc/io/decoder/defs.h
```diff
@@ -165,7 +165,7 @@ struct MediaFormat {
 struct DecoderParameters {
   // local file, remote file, http url, rtmp stream uri, etc. anything that
   // ffmpeg can recognize
-  std::string uri;
+  std::string uri{std::string()};
```
bjuncek, line 168: Why this? 
(don't know C++ well, so I'm just curious)

bjuncek (changes requested):
Looks good to me. I'd love to make the tests stricter but otherwise good to go :)

torchvision/csrc/io/decoder/defs.h
jdsgomes: just want to initialise  the uri value as empty string

test/test_videoapi.py
jdsgomes: actually there are two videos for which this is not true: `SOX5yA1l24A.mp4`
I verified with this code by reading twice from file and also the result is not the same:

```python
In [3]: reader1 = VideoReader('test/assets/videos/SOX5yA1l24A.mp4')
In [4]: reader2 = VideoReader('test/assets/videos/SOX5yA1l24A.mp4')
In [5]: reader1_frames = []
In [6]: reader2_frames = []

In [8]: for r1frame in reader1:
   ...:     reader1_frames.append(r1frame["data"])

In [9]: for r2frame in reader2:
   ...:     reader2_frames.append(r2frame["data"])

In [12]: for i in range(len(reader1_frames)):
    ...:     if torch.all(reader1_frames[i].eq(reader2_frames[i])):
    ...:         print(f"Frame {i} is not equal")

Frame 0 is not equal
Frame 2 is not equal
Frame 3 is not equal
Frame 4 is not equal
Frame 5 is not equal
Frame 6 is not equal
Frame 7 is not equal
Frame 8 is not equal
Frame 9 is not equal
Frame 10 is not equal
Frame 11 is not equal
Frame 12 is not equal
....
```

Do you know if for this particular format the decoding could be non deterministic?

test/test_videoapi.py
```diff
@@ -76,8 +76,14 @@ def test_frame_reading(self, test_video):
 
                 # compare the frames and ptss
                 for i in range(len(vr_frames)):
-                    assert av_pts[i] == vr_pts[i]
-                    torch.test.assert_equal(av_frames[i], vr_frames[i])
```
jdsgomes: This was added in the wrong section so reverted the commit. My comment regarding why we cannot make the tests strict from read_from_memory still holds

test/test_videoapi.py
YosuaMichael: This is actually surprising to me, is this non deterministic behaviour also occur on the stable API (if we specify same pts)?

bjuncek: Yup; we use the same backend.
@jdsgomes can you check if it happens on the `pyav`?

Maybe this is a cue that we should reconsider our selection of testing videos.
Specifically, I think we should re-sample them again from whatever are most used datasets these days, plus some FFMPEG fate situations.

jdsgomes: I confirmed that that is the case.

jdsgomes: please note that there is a bug in the test code (checking for equal instead of not equal) but still there are some frames that are different. This happens both in pyav and video_reader

bjuncek (approved):
Approving to unblock, and starting a new issue to track video assets.

PR #6772, opened 2022-10-14
Fix errors on non-contiguous tensors in `equalize()` Tensor-backend

datumbox:
The use of `view()` in `equalize()` leads to the following errors when the input is non-contiguous. The problem affects stable:
```
    hist = torch.bincount(img_chan.view(-1), minlength=256)
```
And prototype:
```
  File "./torchvision/prototype/transforms/functional/_color.py", line 237, in equalize_video
    return equalize_image_tensor(video)
  File "./torchvision/prototype/transforms/functional/_color.py", line 230, in equalize_image_tensor
    return _equalize_image_tensor_vec(image.view(-1, height, width)).view(image.shape)
  File "./torchvision/prototype/features/_feature.py", line 94, in __torch_function__
    output = func(*args, **kwargs or dict())
```
It was first discovered after starting doing end-to-end training with complex Augmentations using the Tensor Backend. The PR updates the use of `view()` to `reshape()` to ensure that both operators work as expected.

### DIFF
--- a/torchvision/prototype/transforms/functional/_color.py
+++ b/torchvision/prototype/transforms/functional/_color.py
@@ -227,7 +227,7 @@ def equalize_image_tensor(image: torch.Tensor) -> torch.Tensor:
     if image.numel() == 0:
         return image
 
-    return _equalize_image_tensor_vec(image.view(-1, height, width)).view(image.shape)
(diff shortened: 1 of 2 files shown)


pmeier (approved):
Thanks Vasilis!

PR #6773, opened 2022-10-14
Update `ToDtype` to avoid unnecessary `to()` calls and fixing types on `Transform`

datumbox:
As discussed offline this PR:
- Avoids the unnecessary `.to(None)` call
- Fixes the nit on the `_transformed_types` of `Transform` to avoid the superfluous `features.is_simple_tensor, features._Feature` idiom.

### DIFF
--- a/torchvision/prototype/transforms/_transform.py
+++ b/torchvision/prototype/transforms/_transform.py
@@ -13,11 +13,7 @@ from torchvision.utils import _log_api_usage_once
 class Transform(nn.Module):
 
     # Class attribute defining transformed types. Other types are passed-through without any transformation
-    _transformed_types: Tuple[Union[Type, Callable[[Any], bool]], ...] = (
-        features.is_simple_tensor,
-        features._Feature,
-        PIL.Image.Image,
-    )
+    _transformed_types: Tuple[Union[Type, Callable[[Any], bool]], ...] = (torch.Tensor, PIL.Image.Image)
 
     def __init__(self) -> None:
         super().__init__()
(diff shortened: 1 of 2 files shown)


pmeier (approved):
LGTM, thanks Vasilis!

torchvision/prototype/transforms/_transform.py
```diff
@@ -13,11 +13,7 @@
 class Transform(nn.Module):
 
     # Class attribute defining transformed types. Other types are passed-through without any transformation
-    _transformed_types: Tuple[Union[Type, Callable[[Any], bool]], ...] = (
-        features.is_simple_tensor,
-        features._Feature,
-        PIL.Image.Image,
-    )
+    _transformed_types: Tuple[Union[Type, Callable[[Any], bool]], ...] = (torch.Tensor, PIL.Image.Image)
```
pmeier, line 15: As discussed offline, we can use `torch.Tensor` here directly, since `isinstance(x, features._Feature) or is_simple_tensor(x)` is the same as `isinstance(x, torch.Tensor)`. 

We should probably update the comment in the future to explain we are annotating with `Callable` now that we don't have any callable in the default definition.

pmeier (approved):
As discussed offline.

torchvision/prototype/transforms/_transform.py
datumbox: Agreed. I've added a comment.

datumbox:
Fun behaviour of defaultdict:
```
d = defaultdict(1)
d.get(1) # None
d[1] # 1
d.get(1) # 1
```

pmeier (approved):
LGTM, **if CI is green**. Not making this mistake again :P

datumbox:
The [failing](https://github.com/pytorch/vision/actions/runs/3250690933/jobs/5334704616) test is unrelated (the flaky gaussian), we are good to go.

PR #6774, opened 2022-10-14
allow tolerances in transforms consistency checks

pmeier:
This PR is a preparation for #6762. We started out with checking exact match between transforms v1 and v2. Thus, we needed no configurable and simply used 

https://github.com/pytorch/vision/blob/c960273c131e41a06a7b47836fb5ee81c88ebc5d/test/prototype_common_utils.py#L167

Now that we are in optimization mode, there are cases where performance changes might also lead to slight numerical changes. This PR enables the option to set them on a `ConsistencyConfig` basis, while keeping `rtol = atol = 0` as the default.

### DIFF
--- a/test/test_prototype_transforms_consistency.py
+++ b/test/test_prototype_transforms_consistency.py
@@ -12,6 +12,7 @@ import pytest
 import torch
 from prototype_common_utils import (
     ArgsKwargs,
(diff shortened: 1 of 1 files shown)


vfdev-5 (approved):
Thanks !

PR #6775, opened 2022-10-14
[prototype] Reshape input before equalize

datumbox:
related to #6772

The input image also requires reshape to avoid errors such as:
```
```

We should consider on a follow up PR to update all `view()` to `reshape()` calls.


pmeier (approved):
Thanks!

> We should consider on a follow up PR to update all `view()` to `reshape()` calls.

+1. Will do.

vfdev-5 (approved):
Stamp ✉️

PR #6777, opened 2022-10-14
replace .view with .reshape

pmeier:
As suggested in https://github.com/pytorch/vision/pull/6775#issue-1409549850, this PR replaces all `.view(...)` calls with `.reshape(...)`. In the regular case, they do the same thing, but in case `.view()` would fail because the input is noncontiguous, `.reshape` fixes this and afterwards performs the specified operation.

Having this earlier would have prevented #6772 and #6775.

### DIFF
--- a/torchvision/prototype/transforms/_auto_augment.py
+++ b/torchvision/prototype/transforms/_auto_augment.py
@@ -484,7 +484,7 @@ class AugMix(_AutoAugmentBase):
 
         orig_dims = list(image_or_video.shape)
         expected_ndim = 5 if isinstance(orig_image_or_video, features.Video) else 4
-        batch = image_or_video.view([1] * max(expected_ndim - image_or_video.ndim, 0) + orig_dims)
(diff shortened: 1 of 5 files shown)


datumbox (commented):
LGTM. Should we do one run with @vfdev-5 script to confirm we didn't introduce any serious speed regression?

pmeier:
```diff
-     1208    0.005    0.000    0.005    0.000 {method 'reshape' of 'torch._C._TensorBase' objects}
-    10135    0.025    0.000    0.031    0.000 {method 'view' of 'torch._C._TensorBase' objects}
+     6725    0.023    0.000    0.031    0.000 {method 'reshape' of 'torch._C._TensorBase' objects}
+     4618    0.010    0.000    0.010    0.000 {method 'view' of 'torch._C._TensorBase' objects}
```

I think we are good. Even if the difference is not measuring tolerance, we are talking about `((0.023 + 0.010) - (0.005 + 0.025)) / 2000` ~= 1.5 µs per call here.

datumbox (approved):

PR #6783, opened 2022-10-18
Add tests and proper support for videos in `ConvertImageDtype`

pmeier:
#6724 shoehorned support for `features.Video` into `ConvertImageDtype`. The implementation is correct, but the naming is confusing since it explicitly mentions images, but now also supports videos. 

Furthermore, we don't have a canonical dispatcher as we do for all other kernels that support for two or more types. Again, this works since videos can be treated as images here but makes the naming confusing, e.g. `F.convert_image_dtype(video)`.

This PR does a few things:

- rename `ConvertImageDtype` and `convert_image_dtype` to `ConvertDtype` and `convert_dtype` while keeping aliases for the former names
- add `convert_dtype_image_tensor` and `convert_dtype_video` kernels and make `convert_dtype` a proper dispatcher of the former
- Add tests for the functionals

### DIFF
--- a/test/prototype_transforms_dispatcher_infos.py
+++ b/test/prototype_transforms_dispatcher_infos.py
@@ -31,6 +31,8 @@ class DispatcherInfo(InfoBase):
         *,
         # Dictionary of types that map to the kernel the dispatcher dispatches to.
         kernels,
+        # Name of the corresponding method on the features. If omitted, defaults to the dispatcher name
+        method_name=None,
         # If omitted, no PIL dispatch test will be performed.
         pil_kernel_info=None,
         # See InfoBase
@@ -41,6 +43,7 @@ class DispatcherInfo(InfoBase):
         super().__init__(id=dispatcher.__name__, test_marks=test_marks, closeness_kwargs=closeness_kwargs)
         self.dispatcher = dispatcher
         self.kernels = kernels
+        self.method_name = method_name or self.id
         self.pil_kernel_info = pil_kernel_info
 
         kernel_infos = {}
@@ -415,5 +418,13 @@ DISPATCHER_INFOS = [
         test_marks=[
             skip_dispatch_feature,
         ],
+    ),
+    DispatcherInfo(
+        F.convert_dtype,
+        kernels={
+            features.Image: F.convert_dtype_image_tensor,
+            features.Video: F.convert_dtype_video,
+        },
+        method_name="to_dtype",
     ),
 ]
--- a/test/prototype_transforms_dispatcher_infos.py
+++ b/test/prototype_transforms_dispatcher_infos.py
@@ -31,8 +31,6 @@ class DispatcherInfo(InfoBase):
         *,
         # Dictionary of types that map to the kernel the dispatcher dispatches to.
         kernels,
-        # Name of the corresponding method on the features. If omitted, defaults to the dispatcher name
-        method_name=None,
         # If omitted, no PIL dispatch test will be performed.
         pil_kernel_info=None,
         # See InfoBase
@@ -43,7 +41,6 @@ class DispatcherInfo(InfoBase):
         super().__init__(id=dispatcher.__name__, test_marks=test_marks, closeness_kwargs=closeness_kwargs)
         self.dispatcher = dispatcher
         self.kernels = kernels
-        self.method_name = method_name or self.id
         self.pil_kernel_info = pil_kernel_info
 
         kernel_infos = {}
@@ -425,6 +422,8 @@ DISPATCHER_INFOS = [
             features.Image: F.convert_dtype_image_tensor,
             features.Video: F.convert_dtype_video,
         },
-        method_name="to_dtype",
+        test_marks=[
+            skip_dispatch_feature,
+        ],
     ),
 ]
--- a/test/test_prototype_transforms_functional.py
+++ b/test/test_prototype_transforms_functional.py
@@ -178,7 +178,7 @@ class TestKernels:
 
     @sample_inputs
     @pytest.mark.parametrize("device", cpu_and_gpu())
-    def test_dtype_consistency(self, info, args_kwargs, device):
+    def test_dtype_and_device_consistency(self, info, args_kwargs, device):
         (input, *other_args), kwargs = args_kwargs.load(device)
 
         output = info.kernel(input, *other_args, **kwargs)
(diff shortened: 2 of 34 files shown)


test/prototype_transforms_dispatcher_infos.py
```diff
@@ -416,4 +419,12 @@ def xfail_all_tests(*, reason, condition):
             skip_dispatch_feature,
         ],
     ),
+    DispatcherInfo(
+        F.convert_dtype,
+        kernels={
+            features.Image: F.convert_dtype_image_tensor,
+            features.Video: F.convert_dtype_video,
+        },
+        method_name="to_dtype",
```
pmeier, line 428: This is needed since the method on the features is called `to_dtype` whereas the dispatcher is called `convert_dtype`. This is on par with the other conversion methods we already have, e.g.

https://github.com/pytorch/vision/blob/0610b13ac4af3717f538454a9c6b1f441cb386f3/torchvision/prototype/features/_image.py#L113

https://github.com/pytorch/vision/blob/0610b13ac4af3717f538454a9c6b1f441cb386f3/torchvision/prototype/transforms/functional/_meta.py#L264

This was not needed before, since the other conversion methods don't have `DispatcherInfo`'s yet.

test/test_prototype_transforms_functional.py
```diff
@@ -175,7 +175,7 @@ def test_cuda_vs_cpu(self, info, args_kwargs):
 
     @sample_inputs
     @pytest.mark.parametrize("device", cpu_and_gpu())
-    def test_dtype_and_device_consistency(self, info, args_kwargs, device):
+    def test_dtype_consistency(self, info, args_kwargs, device):
```
pmeier, line 178: I've opted to split dtype and device consistency tests. Not sure if this is the right way to go. If we don't, we can only skip this completely for the dtype conversion kernels.

torchvision/prototype/transforms/functional/_meta.py
```diff
@@ -282,3 +282,32 @@ def convert_color_space(
         return inpt.to_color_space(color_space, copy=copy)
     else:
         return convert_color_space_image_pil(inpt, color_space, copy=copy)
+
+
+def convert_dtype_image_tensor(
+    image: torch.Tensor, dtype: torch.dtype = torch.float, copy: bool = True
```
pmeier, line 288: I've opted to add a `copy` flag here in accordance with the other conversion transforms.

torchvision/prototype/transforms/functional/_type_conversion.py
```diff
@@ -41,5 +41,3 @@ def to_image_tensor(image: Union[torch.Tensor, PIL.Image.Image, np.ndarray]) ->
 # We changed the names to align them with the new naming scheme. Still, `to_pil_image` is
 # prevalent and well understood. Thus, we just alias it without deprecating the old name.
 to_pil_image = to_image_pil
-
-convert_image_dtype = _F.convert_image_dtype
```
pmeier, line 45: This was in here by mistake I think. The old transform `ConvertImageDtype` as well as the other conversion transform functionals were already under the `_meta` namespace. Thus, I moved this there.

pmeier:
I agree, naming is most pressing thing here. But then `convert_image_dtype` is not a good name to start with either. I thought about `ConvertValueRange`, but this brings its own set of problems. We could do `ConvertImageOrVideoDtype`, but this goes against our naming scheme. Open for discussions on this.

vfdev-5:
Maybe, emphasize somehow that we change dtype for sample (not targets) ?

pmeier:
This PR conflicts with #6795. However, #6795 has (hopefully) no controversial changes so my vote is out for merging that one first and afterwards coming back to this.

torchvision/prototype/transforms/functional/_meta.py
datumbox: I understand your intention to support `copy` is to align with other APIs but this introduces unnecessary clone calls in our existing Transforms. Under which conditions do we even want to this? If we don't have a real use-case for this, I would prefer to remove it.

torchvision/prototype/features/_image.py
```diff
@@ -122,6 +122,9 @@ def to_color_space(self, color_space: Union[str, ColorSpace], copy: bool = True)
             color_space=color_space,
         )
 
+    def to_dtype(self, dtype: torch.dtype = torch.float, copy: bool = True) -> Image:
```
datumbox, line 125: It's unclear when we are adding methods to specific `_Feature` subclasses and when we dont. In most cases, a method is added when available across all classes. This makes sense as it allows us to call these methods for every `_Feature` without checking its type. There are some limited cases (such as `BoundingBox.to_format`) where for convenience we offer specific methods to subtypes.

Why do we advocate for adding a method here? If we don't really need it, we should avoid it.

datumbox (commented):
@pmeier Thanks for the work.

I agree with the direction of the PR. I have 2 concerns over the use of `copy` and the introduction of more methods in some of the `_Feature` subclasses.

Concerning the conflicts with other PRs, this PR looks like it adds more tests and aligns the API which is less dangerous than rewriting the `convert_dtype_image_tensor` method. How confident are you about the changes on the other PR? Do you think we can benefit by these extra tests and thus prioritise merging this PR first?

torchvision/prototype/features/_image.py
pmeier: My rule so far has been to add methods to the features if something is directly related to the attributes of a tensor. The example that we have for this are

https://github.com/pytorch/vision/blob/211563fba461062268db7cdf5a83203ed9e83e6f/torchvision/prototype/features/_bounding_box.py#L64

https://github.com/pytorch/vision/blob/211563fba461062268db7cdf5a83203ed9e83e6f/torchvision/prototype/features/_image.py#L113

(same for `features.Video`)

https://github.com/pytorch/vision/blob/211563fba461062268db7cdf5a83203ed9e83e6f/torchvision/prototype/features/_label.py#L54

I would put something that changes the tensor based on its `dtype` inside the same category, which is why I added it as method there.

> Why do we advocate for adding a method here? If we don't really need it, we should avoid it.

Unless there is a good reason for `convert_dtype` to be different from `convert_color_space` and `convert_format_bounding_box`, we should be consistent. Either have them all as methods or none of them. As explained above I'm leaning towards having them as methods, but I'm not going to oppose having them only as kernels or dispatchers.

pmeier: See #6801.

datumbox (approved):
LGTM, thanks!

PR #6765, opened 2022-10-13
[NOMERGE] Add optimized `_blend`

datumbox:
This PR is only to show-case the differences against the previous code. It shouldn't be merged as we should do this optimization directly to V2.

Results:
```
adjust_brightness - Winner: v2(device=cpu, dtype=torch.uint8) - v1: 0.0017781441507395356, v2: 0.0015310867200605573 - Diff: -13.89%
adjust_brightness - Winner: v2(device=cpu, dtype=torch.float32) - v1: 0.0008344144199509174, v2: 0.0005974295809864998 - Diff: -28.40%
adjust_brightness - Winner: v2(device=cuda, dtype=torch.uint8) - v1: 5.151713805971667e-05, v2: 4.7857200680300596e-05 - Diff: -7.10%
adjust_brightness - Winner: v2(device=cuda, dtype=torch.float32) - v1: 4.2962464748416095e-05, v2: 3.8983459910377864e-05 - Diff: -9.26%
adjust_contrast - Winner: v2(device=cpu, dtype=torch.uint8) - v1: 0.0019489222508855164, v2: 0.001805254714563489 - Diff: -7.37%
adjust_contrast - Winner: v2(device=cpu, dtype=torch.float32) - v1: 0.0009198985202237964, v2: 0.0007649726495146752 - Diff: -16.84%
adjust_contrast - Winner: v2(device=cuda, dtype=torch.uint8) - v1: 0.00012647993885912, v2: 0.0001235444841440767 - Diff: -2.32%
adjust_contrast - Winner: v2(device=cuda, dtype=torch.float32) - v1: 9.984084800817073e-05, v2: 9.742333088070155e-05 - Diff: -2.42%
adjust_saturation - Winner: v2(device=cpu, dtype=torch.uint8) - v1: 0.0019948067096993327, v2: 0.0018562771938741207 - Diff: -6.94%
adjust_saturation - Winner: v2(device=cpu, dtype=torch.float32) - v1: 0.0009555038996040821, v2: 0.0007746961107477546 - Diff: -18.92%
adjust_saturation - Winner: v2(device=cuda, dtype=torch.uint8) - v1: 0.0001049026179825887, v2: 0.00010217939910944551 - Diff: -2.60%
adjust_saturation - Winner: v2(device=cuda, dtype=torch.float32) - v1: 8.710524649359286e-05, v2: 8.433246007189155e-05 - Diff: -3.18%
[...]

PR #6784, opened 2022-10-18
[proto] Speed up adjust color ops

vfdev-5:
Description:
- Rewritten `_blend` op in prototype space
- Simplified `adjust_brightness`
- Optimized `_rgb_to_gray`
- Removed redundant checks 
- Follow-up PR to https://github.com/pytorch/vision/pull/6765

Results:
```
[------------------ Adjust_brightness cpu torch.uint8 ------------------]
                     |  adjust_brightness stable  |  adjust_brightness v2
1 threads: --------------------------------------------------------------
      (3, 400, 400)  |            1080            |          483         

[----------------- Adjust_contrast cpu torch.uint8 -----------------]
                     |  adjust_contrast stable  |  adjust_contrast v2
1 threads: ----------------------------------------------------------
      (3, 400, 400)  |           1150           |         926        

[------------------ Adjust_saturation cpu torch.uint8 ------------------]
                     |  adjust_saturation stable  |  adjust_saturation v2
1 threads: --------------------------------------------------------------
      (3, 400, 400)  |            1200            |          935         

[------------------ Adjust_sharpness cpu torch.uint8 -----------------]
                     |  adjust_sharpness stable  |  adjust_sharpness v2
1 threads: ------------------------------------------------------------
      (3, 400, 400)  |            4.06           |          3.64       

[----------------- Adjust_brightness cpu torch.float32 -----------------]
                     |  adjust_brightness stable  |  adjust_brightness v2
1 threads: --------------------------------------------------------------
      (3, 400, 400)  |            665             |          206         
6 threads: --------------------------------------------------------------
      (3, 400, 400)  |            167             |           57         

[---------------- Adjust_contrast cpu torch.float32 ----------------]
                     |  adjust_contrast stable  |  adjust_contrast v2
1 threads: ----------------------------------------------------------
      (3, 400, 400)  |           691            |         455        
6 threads: ----------------------------------------------------------
      (3, 400, 400)  |           248            |         162        

[----------------- Adjust_saturation cpu torch.float32 -----------------]
                     |  adjust_saturation stable  |  adjust_saturation v2
1 threads: --------------------------------------------------------------
      (3, 400, 400)  |            728             |          466         

[----------------- Adjust_sharpness cpu torch.float32 ----------------]
                     |  adjust_sharpness stable  |  adjust_sharpness v2
1 threads: ------------------------------------------------------------
      (3, 400, 400)  |            3100           |          2860       

```

More results in [logs](https://github.com/vfdev-5/tvapiv2_benchmarks/blob/bf1dbe56e3c1962f8e7225cc529c36777b2be123/output/20221020-111409-output-adjust-color-ops_41179579f.log), [code](https://github.com/vfdev-5/tvapiv2_benchmarks/blob/bf1dbe56e3c1962f8e7225cc529c36777b2be123/check_adjust_color_ops.py)

Logs with #6765 code + simplified adjust brightness : https://github.com/vfdev-5/tvapiv2_benchmarks/blob/ba35a27f14fd77123ea96f0c54ac9e2fb22f0f15/output/20221018-115341-output-adjust-color-ops_datumbox.log

### DIFF
--- a/test/test_prototype_transforms_consistency.py
+++ b/test/test_prototype_transforms_consistency.py
@@ -755,7 +755,7 @@ class TestAATransforms:
             expected_output = t_ref(inpt)
             output = t(inpt)
 
-            assert_equal(expected_output, output)
+            assert_close(expected_output, output, atol=1, rtol=0.1)
 
     @pytest.mark.parametrize(
         "inpt",
@@ -803,7 +803,7 @@ class TestAATransforms:
             expected_output = t_ref(inpt)
             output = t(inpt)
 
-            assert_equal(expected_output, output)
+            assert_close(expected_output, output, atol=1, rtol=0.1)
 
     @pytest.mark.parametrize(
         "inpt",
--- a/test/test_prototype_transforms_consistency.py
+++ b/test/test_prototype_transforms_consistency.py
@@ -308,22 +308,28 @@ CONSISTENCY_CONFIGS = [
             ArgsKwargs(brightness=0.1, contrast=0.4, saturation=0.7, hue=0.3),
         ],
     ),
-    ConsistencyConfig(
-        prototype_transforms.ElasticTransform,
-        legacy_transforms.ElasticTransform,
-        [
-            ArgsKwargs(),
-            ArgsKwargs(alpha=20.0),
-            ArgsKwargs(alpha=(15.3, 27.2)),
-            ArgsKwargs(sigma=3.0),
-            ArgsKwargs(sigma=(2.5, 3.9)),
-            ArgsKwargs(interpolation=prototype_transforms.InterpolationMode.NEAREST),
-            ArgsKwargs(interpolation=prototype_transforms.InterpolationMode.BICUBIC),
-            ArgsKwargs(fill=1),
-        ],
-        # ElasticTransform needs larger images to avoid the needed internal padding being larger than the actual image
-        make_images_kwargs=dict(DEFAULT_MAKE_IMAGES_KWARGS, sizes=[(163, 163), (72, 333), (313, 95)]),
-    ),
+    *[
+        ConsistencyConfig(
+            prototype_transforms.ElasticTransform,
+            legacy_transforms.ElasticTransform,
+            [
+                ArgsKwargs(),
+                ArgsKwargs(alpha=20.0),
+                ArgsKwargs(alpha=(15.3, 27.2)),
+                ArgsKwargs(sigma=3.0),
+                ArgsKwargs(sigma=(2.5, 3.9)),
+                ArgsKwargs(interpolation=prototype_transforms.InterpolationMode.NEAREST),
+                ArgsKwargs(interpolation=prototype_transforms.InterpolationMode.BICUBIC),
+                ArgsKwargs(fill=1),
+            ],
+            # ElasticTransform needs larger images to avoid the needed internal padding being larger than the actual image
+            make_images_kwargs=dict(DEFAULT_MAKE_IMAGES_KWARGS, sizes=[(163, 163), (72, 333), (313, 95)], dtypes=[dt]),
+            # We updated gaussian blur kernel generation with a faster and numerically more stable version
+            # This brings float32 accumulation visible in elastic transform -> we need to relax consistency tolerance
+            closeness_kwargs=ckw,
+        )
+        for dt, ckw in [(torch.uint8, {"rtol": 1e-1, "atol": 1}), (torch.float32, {"rtol": 1e-2, "atol": 1e-3})]
+    ],
     ConsistencyConfig(
         prototype_transforms.GaussianBlur,
         legacy_transforms.GaussianBlur,
@@ -333,6 +339,7 @@ CONSISTENCY_CONFIGS = [
             ArgsKwargs(kernel_size=3, sigma=0.7),
             ArgsKwargs(kernel_size=5, sigma=(0.3, 1.4)),
         ],
+        closeness_kwargs={"rtol": 1e-5, "atol": 1e-5},
     ),
     ConsistencyConfig(
         prototype_transforms.RandomAffine,
@@ -506,7 +513,6 @@ def check_call_consistency(
         image_repr = f"[{tuple(image.shape)}, {str(image.dtype).rsplit('.')[-1]}]"
 
         image_tensor = torch.Tensor(image)
-
         try:
             torch.manual_seed(0)
             output_legacy_tensor = legacy_transform(image_tensor)
--- a/test/test_prototype_transforms_consistency.py
+++ b/test/test_prototype_transforms_consistency.py
@@ -253,9 +253,10 @@ CONSISTENCY_CONFIGS = [
         legacy_transforms.RandomAdjustSharpness,
         [
             ArgsKwargs(p=0, sharpness_factor=0.5),
-            ArgsKwargs(p=1, sharpness_factor=0.3),
+            ArgsKwargs(p=1, sharpness_factor=0.2),
             ArgsKwargs(p=1, sharpness_factor=0.99),
         ],
+        closeness_kwargs={"atol": 1e-6, "rtol": 1e-6},
     ),
     ConsistencyConfig(
         prototype_transforms.RandomGrayscale,
@@ -305,8 +306,9 @@ CONSISTENCY_CONFIGS = [
             ArgsKwargs(saturation=(0.8, 0.9)),
             ArgsKwargs(hue=0.3),
             ArgsKwargs(hue=(-0.1, 0.2)),
-            ArgsKwargs(brightness=0.1, contrast=0.4, saturation=0.7, hue=0.3),
+            ArgsKwargs(brightness=0.1, contrast=0.4, saturation=0.5, hue=0.6),
         ],
+        closeness_kwargs={"atol": 1e-5, "rtol": 1e-5},
     ),
     *[
         ConsistencyConfig(
--- a/torchvision/prototype/transforms/functional/_color.py
+++ b/torchvision/prototype/transforms/functional/_color.py
@@ -2,7 +2,7 @@ import torch
 from torchvision.prototype import features
 from torchvision.transforms import functional_pil as _FP, functional_tensor as _FT
 
-from ._meta import get_dimensions_image_tensor, get_num_channels_image_tensor, rgb_to_grayscale
+from ._meta import _rgb_to_gray, get_dimensions_image_tensor, get_num_channels_image_tensor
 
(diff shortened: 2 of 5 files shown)


test/test_prototype_transforms_consistency.py
```diff
@@ -254,9 +254,10 @@ def __init__(
         legacy_transforms.RandomAdjustSharpness,
         [
             ArgsKwargs(p=0, sharpness_factor=0.5),
-            ArgsKwargs(p=1, sharpness_factor=0.3),
+            ArgsKwargs(p=1, sharpness_factor=0.2),
```
pmeier, line 257: Why is this needed? Don't we tolerance for that?

torchvision/prototype/transforms/functional/_color.py
```diff
@@ -2,9 +2,29 @@
...
+
+def _blend(img1: torch.Tensor, img2: torch.Tensor, ratio: float) -> torch.Tensor:
+    ratio = float(ratio)
+    fp = img1.is_floating_point()
+    bound = 1.0 if fp else 255.0
+    output = img1.mul(ratio).add_(img2, alpha=(1.0 - ratio)).clamp_(0, bound)
+    return output if fp else output.to(img1.dtype)
+
+
+def adjust_brightness_image_tensor(img: torch.Tensor, brightness_factor: float) -> torch.Tensor:
```
pmeier, line 16: Please align the input names with all other `*_image_tensor` kernels.

```suggestion
def adjust_brightness_image_tensor(image: torch.Tensor, brightness_factor: float) -> torch.Tensor:
```

torchvision/prototype/transforms/functional/_color.py
```diff
@@ -38,7 +71,22 @@ def adjust_saturation(inpt: features.InputTypeJIT, saturation_factor: float) ->
...
+        raise ValueError(f"contrast_factor ({contrast_factor}) is not non-negative.")
+
+    c = get_num_channels_image_tensor(img)
+    if c not in [1, 3]:
+        raise TypeError(f"Input image tensor permitted channel values are {[1, 3]}, but found {c}")
+    dtype = img.dtype if torch.is_floating_point(img) else torch.float32
+    if c == 3:
+        mean = torch.mean(_FT.rgb_to_grayscale(img).to(dtype), dim=(-3, -2, -1), keepdim=True)
+    else:
+        mean = torch.mean(img.to(dtype), dim=(-3, -2, -1), keepdim=True)
```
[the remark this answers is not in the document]
vfdev-5: Not very readable, but why not

test/test_prototype_transforms_consistency.py
vfdev-5: This helps to fix tolerance without adjusting a/r tol :)

torchvision/prototype/transforms/functional/_color.py
```diff
@@ -21,7 +41,20 @@ def adjust_brightness(inpt: features.InputTypeJIT, brightness_factor: float) ->
...
+        raise ValueError(f"saturation_factor ({saturation_factor}) is not non-negative.")
+
+    c = get_num_channels_image_tensor(img)
+    if c not in [1, 3]:
+        raise TypeError(f"Input image tensor permitted channel values are {[1, 3]}, but found {c}")
+
+    if c == 1:  # Match PIL behaviour
+        return img
+
+    return _blend(img, _FT.rgb_to_grayscale(img), saturation_factor)
```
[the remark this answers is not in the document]
vfdev-5: It is slower by ~10us :)

test/test_prototype_transforms_consistency.py
pmeier: In that case, why do we need it do it 3 lines below?

torchvision/prototype/transforms/functional/_color.py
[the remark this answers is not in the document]
vfdev-5: I'll push the update here in an hour

test/test_prototype_transforms_consistency.py
vfdev-5: there are float tol and uint8 tol. `sharpness_factor=0.2` avoids uint8 tol issue. `closeness_kwargs={"atol": 1e-6, "rtol": 1e-6},` helps with float tol issue

torchvision/prototype/transforms/functional/_meta.py
```diff
@@ -184,7 +184,31 @@ def _gray_to_rgb(grayscale: torch.Tensor) -> torch.Tensor:
     return grayscale.repeat(repeats)
 
 
-_rgb_to_gray = _FT.rgb_to_grayscale
+def rgb_to_grayscale(image: torch.Tensor, num_output_channels: int = 1) -> torch.Tensor:
```
pmeier, line 187: This is actually deprecated:

https://github.com/pytorch/vision/blob/7a62a545ce76f43ccc5cfe0009131f7db14ae7b5/torchvision/prototype/transforms/functional/_deprecated.py#L25

Can't we merge the `c == 3` branch into `convert_color_space_image_tensor`? If there is actually a perf implication and we need the faster op as standalone, I'm ok to have a `_rgb_to_gray` as before that only does this and can be used from the other color kernels.

vfdev-5: I do not follow you about deprecated method. Deprecated method is inside `_deprecated.py`, here I replaced `_rgb_to_gray = _FT.rgb_to_grayscale`. If there is something to do with `convert_color_space_image_tensor`, I'd prefer to keep it for another PR.

pmeier: You defined a function `rgb_to_grayscale` here. There are several issues with this:

1. It clashes with the function of the same name in the deprecated namespace.
2. It doesn't adhere to the naming convention with `features.ColorSpace.GRAY` since it uses the term `grayscale`
3. It reintroduces the whole concept of `num_output_channels` which we also deprecated. `convert_color_space` no longer has this concept.

Basically what I'm asking here is that you implement your perf improvements into `_rgb_to_gray`. Since that is private we can also remove all of the error checking. Basically the whole function should look like

```py
def _rgb_to_gray(image: torch.Tensor) -> torch.Tensor:
    r, g, b = image.unbind(dim=-3)
    return (r * 0.2989).add_(g, alpha=0.587).add_(b, alpha=0.114)
```

torchvision/prototype/transforms/functional/_color.py
```diff
@@ -2,9 +2,29 @@
...
 
-from ._meta import get_dimensions_image_tensor
+from ._meta import _rgb_to_gray, get_dimensions_image_tensor, get_num_channels_image_tensor
+
+
+def _blend(image1: torch.Tensor, image2: torch.Tensor, ratio: float) -> torch.Tensor:
+    ratio = float(ratio)
+    fp = image1.is_floating_point()
+    bound = 1.0 if fp else 255.0
+    output = image1.mul(ratio).add_(image2, alpha=(1.0 - ratio)).clamp_(0, bound)
```
datumbox, line 12: I like this! Supersedes the work at #6765

datumbox (approved):
LGTM, looks great!

vfdev-5:
With the latest commit we reduce runtime for adjust_contrast:
```
[----------------- Adjust_contrast cpu torch.uint8 -----------------]
                     |  adjust_contrast stable  |  adjust_contrast v2
1 threads: ----------------------------------------------------------
      (3, 400, 400)  |           1160           |         831        

```

torchvision/prototype/transforms/functional/_color.py
```diff
@@ -38,7 +71,19 @@ def adjust_saturation(inpt: features.InputTypeJIT, saturation_factor: float) ->
...
+def adjust_contrast_image_tensor(image: torch.Tensor, contrast_factor: float) -> torch.Tensor:
+    if contrast_factor < 0:
+        raise ValueError(f"contrast_factor ({contrast_factor}) is not non-negative.")
+
+    c = get_num_channels_image_tensor(image)
+    if c not in [1, 3]:
+        raise TypeError(f"Input image tensor permitted channel values are {[1, 3]}, but found {c}")
+    dtype = image.dtype if torch.is_floating_point(image) else torch.float32
+    grayscale_image = _rgb_to_gray(image) if c == 3 else image
+    mean = torch.mean(grayscale_image.to(dtype), dim=(-3, -2, -1), keepdim=True)
```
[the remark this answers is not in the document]
vfdev-5: @pmeier actually, doing so the output of `mean` is not the same for uint8 image.
As we originally applied uint8 cast in the end of `_rgb_to_gray` we get rid of all floating point values. This is not the case if image is casted to float before `_rgb_to_gray`. 
So, finally consistency tests report for example:
```
Mismatched elements: 256 / 2772 (9.2%)                                                                                                                                      
Greatest absolute difference: 3 at index (1, 2, 4, 21) (up to 1e-05 allowed)   
Greatest relative difference: 0.25 at index (2, 0, 6, 21) (up to 1e-05 allowed)
```
and this is a real failure, IMO.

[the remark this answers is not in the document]
pmeier: I agree that the behavior changes, but IMO repeatedly converting to `uint8` in the computation and thus eliminating intermediate values sounds more like a missed opportunity in the original kernel than a bug now. Thus, I would consider this more like a "bug fix" rather than a BC breaking change. On the other hand, that is not a strong opinion. Not going to block over this.

torchvision/prototype/transforms/functional/_color.py
```diff
@@ -2,9 +2,29 @@
...
+
+def adjust_brightness_image_tensor(image: torch.Tensor, brightness_factor: float) -> torch.Tensor:
+    if brightness_factor < 0:
+        raise ValueError(f"brightness_factor ({brightness_factor}) is not non-negative.")
+
+    _FT._assert_channels(image, [1, 3])
+
+    fp = image.is_floating_point()
+    bound = 1.0 if fp else 255.0
+    output = image.mul(brightness_factor).clamp_(0, bound)
```
MiChatz, line 24: Issue:

This clamp_ implementation is enforcing the histogram to take values between 0 and 1 in my case is not working the way I will expect since my images are normalized to values around 0 (ex: -0.2 to _0.8) so the multiplication with the factor of 1 will not return an identical image. 

Suggestion:
    ```output = image.mul(brightness_factor).clamp_(image.min(), image.max())```

vfdev-5: @MiChatz thanks for the feedback. There is (yet unwritten) assumption for color transformations on float images that image range is between [0, 1].

PR #6786, opened 2022-10-18
Add raft_stereo weights

YosuaMichael:
Adding weights for raft stereo models

### DIFF
--- a/torchvision/prototype/models/depth/stereo/raft_stereo.py
+++ b/torchvision/prototype/models/depth/stereo/raft_stereo.py
@@ -1,3 +1,4 @@
+from functools import partial
 from typing import Callable, List, Optional, Tuple
 
 import torch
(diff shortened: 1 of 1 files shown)


torchvision/prototype/models/depth/stereo/raft_stereo.py
```diff
@@ -624,11 +626,93 @@ def _raft_stereo(
...
+                # Following metrics from paper: https://arxiv.org/abs/2109.07547
+                # Using standard metrics for each datasets
+                "Kitty2015": {
+                    # Ratio of pixels with difference less than 3px from ground truth
+                    "3px": 0.9426,
+                },
+                "Middlebury2014-val": {
+                    # Ratio of pixels with difference less than 2px from ground truth
+                    # on full, half, and quarter image resolution
+                    "full-2px": 0.8167,
```
jdsgomes, line 666: Nit: possibly we could just have 2px as key for the full-resolution so it matches directly what we have in other weights.

jdsgomes (approved):
LGTM, just left a NIT which is just a suggestion

YosuaMichael:
The CI breakage is unrelated, and the PR pass the prototype test which should be good enough.

PR #6791, opened 2022-10-18
[proto] Some fixes for crestereo

YosuaMichael:
- Add default `None` for the `flow_init` in the forward (so user no need to explicitly specify it)
- Fix mypy complain
- Do verify weights, so user can use string as input (use `model = crestereo_base(weights="DEFAULT")` )

### DIFF
--- a/torchvision/prototype/models/depth/stereo/crestereo.py
+++ b/torchvision/prototype/models/depth/stereo/crestereo.py
@@ -763,7 +763,7 @@ class CREStereo(nn.Module):
         return "1d" if iteration % 2 == 0 else "2d"
 
     def forward(
-        self, left_image: Tensor, right_image: Tensor, flow_init: Optional[Tensor], num_iters: int = 10
+        self, left_image: Tensor, right_image: Tensor, flow_init: Optional[Tensor] = None, num_iters: int = 10
     ) -> List[Tensor]:
         features = torch.cat([left_image, right_image], dim=0)
         features = self.feature_encoder(features)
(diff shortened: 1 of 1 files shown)


torchvision/prototype/models/depth/stereo/crestereo.py
```diff
@@ -781,10 +781,10 @@ def forward(
         ctx_pyramid = self.downsampling_pyramid(ctx)
 
         # we store in reversed order because we process the pyramid from top to bottom
-        l_pyramid: Dict[str, Tensor] = {res: l_pyramid[idx] for idx, res in enumerate(self.resolutions)}
-        r_pyramid: Dict[str, Tensor] = {res: r_pyramid[idx] for idx, res in enumerate(self.resolutions)}
-        net_pyramid: Dict[str, Tensor] = {res: net_pyramid[idx] for idx, res in enumerate(self.resolutions)}
-        ctx_pyramid: Dict[str, Tensor] = {res: ctx_pyramid[idx] for idx, res in enumerate(self.resolutions)}
```
YosuaMichael, line 787: My local mypy complains: 
```
crestereo.py:784: error: Name "l_pyramid" already defined on line 778
crestereo.py:785: error: Name "r_pyramid" already defined on line 779
crestereo.py:786: error: Name "net_pyramid" already defined on line 780
crestereo.py:787: error: Name "ctx_pyramid" already defined on line 781
Found 4 errors in 1 file (checked 1 source file)
```
and removing the type declaration seems to fix it.

jdsgomes (approved):
LGTM thanks

PR #6797, opened 2022-10-19
[Nova] Migrate Linux CPU job to Generic Job

osalpekar:
Migrating the Linux CPU Testing job to the Generic Linux Job format that will be used long-term for CI jobs.

### DIFF
--- a/.github/workflows/test-linux-cpu.yml
+++ b/.github/workflows/test-linux-cpu.yml
@@ -18,10 +18,6 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
-      - name: Set Channel
-        if: ${{ (github.event_name == 'pull_request' && startsWith(github.base_ref, 'release')) || startsWith(github.ref, 'refs/heads/release') }}
-        run: |
-          echo "CHANNEL=test" >> "$GITHUB_ENV"
   tests:
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/test-linux-cpu.yml
```diff
@@ -13,58 +13,50 @@ env:
   CHANNEL: "nightly"
 
 jobs:
+  setup:
```
seemethere, line 16: Do we still need this `setup` job?

.github/workflows/test-linux-cpu.yml
```diff
@@ -13,58 +13,50 @@ env:
...
-          ${CONDA_RUN} python3 -m pytest --junitxml=test-results/junit.xml -v --durations 20
-          conda env remove -p ${ENV_NAME}
+        # Set up Environment Variables
+        export ENV_NAME=conda-env-"${{ github.run_id }}"
+        export PYTHON_VERSION="${{ matrix.py_vers }}"
+        export VERSION="cpu"
+        export CUDATOOLKIT="cpuonly"
+
+        # Set CHANNEL
+        if [[${{ (github.event_name == 'pull_request' && startsWith(github.base_ref, 'release')) || startsWith(github.ref, 'refs/heads/release') }} ]]; then
```
seemethere, line 40: These should actually be available as regular environment variables so it might be better to use those instead, see: https://docs.github.com/en/actions/learn-github-actions/environment-variables

.github/workflows/test-linux-cpu.yml
```diff
@@ -13,58 +13,50 @@ env:
...
+          export CHANNEL=test
+        else
+          export CHANNEL=nightly
+        fi
+
+        # Create Conda Env
+        conda create -yp "${ENV_NAME}" python="${PYTHON_VERSION}" numpy libpng jpeg scipy
+        export CONDA_RUN="conda run -p ${ENV_NAME}"
```
seemethere, line 48: You actually don't necessarily need to do this, you can just do a `conda activate` and run regular conda commands in this

seemethere (commented):
Looks mostly good, just a few optimizations to make, and to clean up some un-needed jobs

.github/workflows/test-linux-cpu.yml
osalpekar: Oh I didn't see the repository input for the generic job. Yeah we can use that instead.

osalpekar: nit: feedback on the inputs for the linux job - if the primary target for this is CI/lint jobs as opposed to Release/Nightly builds (which are covered by our other Nova tooling), I feel the ref argument should default to main instead of nightly

Issue #6768, opened 2022-10-14
[prototype] Add temporal Jittering and Permute transforms and kernels


datumbox:
### 🚀 The feature

Though the sampling of the temporal dimension can be handled efficiently on the VideoClipper (work done by @YosuaMichael), some users still want to perform jittering on the temporal dimension as part of the transforms. This can be achieved by uniformly sampling the indeces of the available samples and selecting the appropriate frames.

Similarly we need an easy transform for permuting channel order.

### Motivation, pitch

Add a Temporal Sampler and a Permute method as transform along with the necessary kernels.

cc @vfdev-5 @bjuncek @pmeier

PR #6800, opened 2022-10-20
Add PermuteDimensions and TransposeDimensions transforms

datumbox:
Part of #6768

This PR adds the `PermuteDimensions` and `TransposeDimensions` transforms which are [useful](https://github.com/facebookresearch/pytorchvideo/blob/main/pytorchvideo/transforms/transforms.py#L317) for video training which is often needed to move the temporal dimension around.

Because the end tensors can have their dimensions completely messed, meta-data information from Images and Videos are not retrained and instead a simple tensor is assumed.

cc @vfdev-5 @bjuncek @pmeier

### DIFF
--- a/torchvision/prototype/transforms/_misc.py
+++ b/torchvision/prototype/transforms/_misc.py
@@ -173,14 +173,13 @@ class PermuteDimensions(Transform):
             dims = defaultdict(functools.partial(_default_arg, dims))
         self.dims = dims
 
-    def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
+    def _transform(
+        self, inpt: Union[features.TensorImageType, features.TensorVideoType], params: Dict[str, Any]
+    ) -> torch.Tensor:
         dims = self.dims[type(inpt)]
         if dims is None:
-            return inpt
-        output = inpt.permute(dims)
-        if isinstance(inpt, (features.Image, features.Video)):
-            output = inpt.wrap_like(inpt, output)
-        return output
+            return inpt.as_subclass(torch.Tensor)
+        return inpt.permute(*dims)
 
 
 class TransposeDimensions(Transform):
@@ -192,15 +191,13 @@ class TransposeDimensions(Transform):
             dims = defaultdict(functools.partial(_default_arg, dims))
         self.dims = dims
 
-    def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
+    def _transform(
+        self, inpt: Union[features.TensorImageType, features.TensorVideoType], params: Dict[str, Any]
+    ) -> torch.Tensor:
         dims = self.dims[type(inpt)]
         if dims is None:
-            return inpt
-        dim0, dim1 = dims
-        output = inpt.transpose(dim0, dim1)
-        if isinstance(inpt, (features.Image, features.Video)):
-            output = inpt.wrap_like(inpt, output)
-        return output
+            return inpt.as_subclass(torch.Tensor)
+        return inpt.transpose(*dims)
 
 
 class RemoveSmallBoundingBoxes(Transform):
--- a/torchvision/prototype/transforms/_misc.py
+++ b/torchvision/prototype/transforms/_misc.py
@@ -1,6 +1,6 @@
 import functools
 from collections import defaultdict
-from typing import Any, Callable, Dict, List, Sequence, Type, Union
+from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple, Type, Union
 
 import PIL.Image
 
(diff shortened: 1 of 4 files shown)


torchvision/prototype/transforms/_misc.py
```diff
@@ -163,6 +164,42 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
...
+            dims = defaultdict(functools.partial(_default_arg, dims))
+        self.dims = dims
+
+    def _transform(
+        self, inpt: Union[features.TensorImageType, features.TensorVideoType], params: Dict[str, Any]
+    ) -> torch.Tensor:
+        dims = self.dims[type(inpt)]
+        if dims is None:
+            return inpt.as_subclass(torch.Tensor)
+        return inpt.permute(*dims)
```
datumbox, line 182: The end result is permuted, so standard wrapper methods like num_channels etc won't work. This is why we return the result as a simple Tensor. Users would have to manually premute and rewrap if needed though they are advised to do this at the end.

torchvision/prototype/transforms/_misc.py
```diff
@@ -163,6 +164,42 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
...
+
+
+class TransposeDimensions(Transform):
+    _transformed_types = (features.is_simple_tensor, features.Image, features.Video)
+
+    def __init__(self, dims: Union[Tuple[int, int], Dict[Type, Optional[Tuple[int, int]]]]) -> None:
```
datumbox, line 188: Tuple of the two dimensions that are about to change.

torchvision/prototype/transforms/_misc.py
```diff
@@ -163,6 +164,42 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
...
+    def _transform(
+        self, inpt: Union[features.TensorImageType, features.TensorVideoType], params: Dict[str, Any]
+    ) -> torch.Tensor:
+        dims = self.dims[type(inpt)]
+        if dims is None:
+            return inpt.as_subclass(torch.Tensor)
+        return inpt.permute(*dims)
+
+
+class TransposeDimensions(Transform):
```
datumbox, line 185: The Transpose transform can work with negative values on the dimension tuples that we want to flip. This can be useful for writing pipelines that work both in batched and non-batched input.

torchvision/prototype/transforms/_misc.py
```diff
@@ -142,18 +142,19 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
...
+
+
 class ToDtype(Transform):
     _transformed_types = (torch.Tensor,)
 
-    def _default_dtype(self, dtype: torch.dtype) -> torch.dtype:
-        return dtype
-
-    def __init__(self, dtype: Union[torch.dtype, Dict[Type, torch.dtype]]) -> None:
+    def __init__(self, dtype: Union[torch.dtype, Dict[Type, Optional[torch.dtype]]]) -> None:
```
datumbox, line 152: Only within the dictionary we expect `None` values. Initiatilizing the transform with a single `None` value makes little sense. The same "policy" is applied to all transforms below.

torchvision/prototype/transforms/_misc.py
```diff
@@ -142,18 +142,19 @@ def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
         return F.gaussian_blur(inpt, self.kernel_size, **params)
 
 
+def _default_arg(value: Any) -> Any:
+    return value
```
pmeier, line 146: Since we now having this idiom multiple times, should we have this in `_utils`? Maybe even hide this ugly construct in a 

```py
def _get_defaultdict(default: T) -> Dict[Any, T]:
    return defaultdict(functools.partial(_default_arg, default))
```

Otherwise we need to add an explanation comment to every occurrence of this idiom since it is not obvious what is happening and the convolution is needed.

pmeier (commented):
I agree we need this functionality somehow, but I'm not sure these general transformations are the way to go. They have very generic names and one could reasonable deduce that `PermuteDimensions` also works for `features.BoundingBox`'es or the like. Worse, one can even put `features.BoundingBox` in the permutation dict just for it to be silently ignored.

Basically these transforms suffer from the same fate as `ConvertDtype` from #6783. We basically need a name for "image or video" that can identify these transforms better. I'm going to paddle one idea that @vfdev-5 started in https://github.com/pytorch/vision/pull/6783#issuecomment-1281990986 and I already mentioned in our offline conversation: right now we use the term "input" (or `inpt`) to be exact to mean `Union[features._Feature, _is_simple_tensor, PIL.Image.Image]`. However, "input" would be a good name for `Union[features.Image, features.Video, _is_simple_tensor, PIL.Image.Image]` and could clearly separate from "target" or "annotation" like `Union[features.Label, features.Mask, features.BoundingBox]`. We could change what we currently call `inpt` to `feature`. That would free the term "input" and we could use that for `ConvertInputDtype`, `PermuteInputDimensions`, ... Of course `torch.Tensor`'s and `PIL.Image.Image`'s are not `features._Feature`'s but they have the same purpose. We could even for `feature_like` as a name to avoid confusion. WDYT?

Regarding the way forward: do we actually need both transforms or is one of them sufficient? Given that our references currently only need to swap two dimensions

https://github.com/pytorch/vision/blob/246de0772c80eae435c0a562268d8d21ed7a27a2/references/video_classification/transforms.py#L6

wouldn't `TransposeDimensions` be sufficient here? Although it can only swap two dimensions at a time, it has one major upside: you don't have to know the whole shape. For example `image.transpose(-1, -2)` is valid for arbitrary batch sizes while I need to know the number of dimensions if I use `image.permute(...)` although I don't need to know.

datumbox:
@pmeier Let's take the naming discussion out side of this PR because these are transforms we need to add ASAP to reach parity of functionality for internal use-cases (ClassyVision). I confirm we need `PermuteDimensions` because often the video can be read in the following formats `"THWC", "TCHW", "CTHW"`. `TransposeDimensions` is optional but advised as it will allow us to handle arbitrary batch sizes.

I'll move the `_get_defaultdict` into utils as you suggested, let me know if there are any other blocking comments.

PR #6801, opened 2022-10-20
Cleanup conversion transforms

pmeier:
This PR stems from the discussion in https://github.com/pytorch/vision/pull/6783#discussion_r1000557482. @datumbox and I agreed to the following things:

- Let's not have a `copy` flag on our `convert_*` kernels. It is currently unused for `convert_color_space`. It is used extensively for `convert_bounding_box` in our kernels due the idiom of converting to a specific format, performing the computation, and converting back. In the future, we wanted to have a look at these kernels anyway for possible performance gains by implementing the computation for every format. After this PR the current implementation like

    ```python
    bounding_box = convert_format_bounding_box(
        bounding_box, old_format=format, new_format=features.BoundingBoxFormat.XYXY
    )
    ```
    
    changes to
    
    ```python
    bounding_box = (
        bounding_box.clone()
        if format == features.BoundingBoxFormat.XYXY
        else convert_format_bounding_box(bounding_box, old_format=format, new_format=features.BoundingBoxFormat.XYXY)
    )
    ```
- Let's not have any `.to_*()` methods on the features. While useful for debugging, they go against the other idioms we are using. We can always re-add them later if there is user demand for it. Note that `Label.to_categories()` is an exception since it has no transformation associated with it.


vfdev-5:
```python
bounding_box = (
    bounding_box.clone()
    if format == features.BoundingBoxFormat.XYXY
    else convert_format_bounding_box(bounding_box, old_format=format, new_format=features.BoundingBoxFormat.XYXY)
)
```
IMO, this is too bulky. I do not mind to have `copy` in `convert_format_bounding_box` and reduce lines of user's code...

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -394,16 +398,18 @@ def affine_bounding_box(
     center: Optional[List[float]] = None,
 ) -> torch.Tensor:
     original_shape = bounding_box.shape
-    bounding_box = convert_format_bounding_box(
-        bounding_box, old_format=format, new_format=features.BoundingBoxFormat.XYXY
+    bounding_box = (
+        bounding_box.clone()
```
datumbox, line 402: Are you sure we need clone here? I don't see any inplace ops in `_affine_bounding_box_xyxy`. Does it make sense to remove it?

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -583,8 +589,10 @@ def rotate_bounding_box(
         center = None
 
     original_shape = bounding_box.shape
-    bounding_box = convert_format_bounding_box(
-        bounding_box, old_format=format, new_format=features.BoundingBoxFormat.XYXY
+    bounding_box = (
+        bounding_box.clone()
```
datumbox, line 593: Same.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -896,8 +904,10 @@ def perspective_bounding_box(
         raise ValueError("Argument perspective_coeffs should have 8 float values")
 
     original_shape = bounding_box.shape
-    bounding_box = convert_format_bounding_box(
-        bounding_box, old_format=format, new_format=features.BoundingBoxFormat.XYXY
+    bounding_box = (
+        bounding_box.clone()
```
datumbox, line 908: Same.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -1061,8 +1071,10 @@ def elastic_bounding_box(
     displacement = displacement.to(bounding_box.device)
 
     original_shape = bounding_box.shape
-    bounding_box = convert_format_bounding_box(
-        bounding_box, old_format=format, new_format=features.BoundingBoxFormat.XYXY
+    bounding_box = (
+        bounding_box.clone()
```
datumbox, line 1075: Same.

torchvision/prototype/transforms/functional/_meta.py
```diff
@@ -238,47 +239,45 @@ def convert_color_space_image_tensor(
...
-            raise RuntimeError(
-                "In order to convert the color space of simple tensors, "
-                "the `old_color_space=...` parameter needs to be passed."
-            )
-        return convert_color_space_image_tensor(
-            inpt, old_color_space=old_color_space, new_color_space=color_space, copy=copy
-        )
-    elif isinstance(inpt, (features.Image, features.Video)):
-        return inpt.to_color_space(color_space, copy=copy)
+    if isinstance(inpt, torch.Tensor):
```
datumbox, line 266: I find this refactoring less clear than what we had earlier. Can you revert?

datumbox (commented):
Thanks @pmeier, looks good overall. Just a couple of minor comments.

torchvision/prototype/transforms/functional/_geometry.py
pmeier: I was going for parity, i.e. replicating the old behavior exactly. But you are right indeed that we don't need to clone here due to the absence of inplace ops. Since this is a perf improvement, let me add this for this and all other comments you added.

torchvision/prototype/transforms/functional/_meta.py
pmeier: Not only less clear, but also wrong. Ci caught that. I refactored. PTAL.

pmeier:
@vfdev-5 

> I do not mind to have `copy` in `convert_format_bounding_box` and reduce lines of user's code...

Same here with the focus on consistency. If one `convert_*` kernel or dispatcher has a `copy` flag, every should have it. Currently we have `copy: bool = True`, which has a runtime implication unless someone explicitly sets the `copy=False` in case no copy is needed. Meaning `convert_dtype` of https://github.com/pytorch/vision/pull/6783#discussion_r1000451773 will be slower by default than before for mostly no reason. This PR basically hard-codes `copy=False`. 

Vasilis argument was that besides for `convert_format_bounding_box` we aren't using the `copy` flag anywhere. Even that is only used a handful of times throughout our API. Thus, if we want consistency, it is fairly simple to remove and avoids dead code. I don't want to misrepresent the argument, so I let him chime in here. cc @datumbox 

I'm ok with or without the `copy` flag if we are consistent with it. I'll let the two of you hash it out.

datumbox (approved):
LGTM, only one comment below.

vfdev-5:
@datumbox @pmeier can we do the following:
- make all conversion ops like _xywh_to_xyxy inplace (without cloning internally) and 
- we can apply `.clone()` for the output of `convert_format_bounding_box` 
```python
bounding_box = convert_format_bounding_box(
    bounding_box, old_format=format, new_format=features.BoundingBoxFormat.XYXY
).clone()
```
?

datumbox:
@vfdev-5 that would make our only operator that does in-place ops (outside of the optimizations that have no effect). If we were to go down that route, purely for speed reasons, I would recommend adopting the `inplace` argument for this one kernel (because it's used so often in conversions). The default should be `inplace=False` in that case. I would do this outside of this PR though after measuring its speed.

datumbox (approved):

PR #6804, opened 2022-10-20
[Nova] GHA Linux GPU Job

osalpekar:
Adding a GHA job for Linux GPU. Only added python3.8 x cuda11.6 since this is the only config for Linux GPU jobs that's run at the moment in CircleCI, other python and CUDA versions are run on main/other branches only and not per-PR.

### DIFF
--- a/.github/workflows/test-linux-gpu.yml
+++ b/.github/workflows/test-linux-gpu.yml
@@ -21,11 +21,11 @@ jobs:
       fail-fast: false
     uses: pytorch/test-infra/.github/workflows/linux_job.yml@main
     with:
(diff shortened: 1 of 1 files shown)


DanilBaibak (approved):
LGTM

seemethere (approved):

vfdev-5:
@osalpekar can you please check if tests are really running on GPU ? 

I have an impression that installed torch does not have CUDA available : https://github.com/pytorch/vision/actions/runs/3396459351/jobs/5647549321#step:8:1021

If trying your setup for prototype tests where we expect tests to fail on GPU and seems like they were skipped as cuda is not available...

PR #6807, opened 2022-10-20
Unwrap features before passing them into a kernel

### LINKED ISSUE #6781: Remove remaining __torch_function__ calls from regular transform pipeline
On our journey to micro-optimize transforms v2, we recently merged #6681. This significantly reduced the number of `__torch_function__` calls inside the augmentation pipeline. However, we still have some unnecessary calls to it that originate from this idiom that we have on the feature methods:

https://github.com/pytorch/vision/blob/149edda463b54b3eabe989e260a839727c89d099/torchvision/prototype/features/_image.py#L125-L126

The kernel, here `horizontal_flip_image_tensor`, is called with `self`, i.e. a `features.Image`. Internally, the first operation will thus go through `__torch_function__` and unwrap it. Meaning, we are going through the whole protocol for no reason, since the kernel per definition works with plain tensors.

We could simply unwrap, i.e. `self.as_subclass(torch.Tensor)` before we call the kernel. That would make our implementation a little more verbose, but would prevent any `__torch_function__` calls when using our transformations whatsoever. Of course we could also have a `self.unwrap()` method to reduce the boilerplate. Another option is to move the kernel call under the `DisableTorchFunction` context manager, similar to what we did in #6681.

```py

import torch
from torch._C import DisableTorchFunction


tensor = torch.rand(3, 256, 256)
image = features.Image(tensor)

arbitrary_unary_image_tensor_kernel = F.autocontrast_image_tensor


def minimum():
    arbitrary_unary_image_tensor_kernel(tensor)


def unwrap_boilerplate():
    arbitrary_unary_image_tensor_kernel(image.as_subclass(torch.Tensor))


def _unwrap_helper(feature):
    return feature.as_subclass(torch.Tensor)


def unwrap_with_helper_method():
    arbitrary_unary_image_tensor_kernel(_unwrap_helper(image))


def context_manager():
    with DisableTorchFunction():
        arbitrary_unary_image_tensor_kernel(image)


def baseline():
    arbitrary_unary_image_tensor_kernel(image)


for fn in [minimum, unwrap_boilerplate, unwrap_with_helper_method, context_manager, baseline]:
    for _ in range(1_000):
        fn()

    time_diffs_ns = []
    for _ in range(100_000):
        start = perf_counter_ns()
        fn()
        stop = perf_counter_ns()
        time_diffs_ns.append(stop - start)

    time_diffs = torch.tensor(time_diffs_ns, dtype=torch.float64) * 1e-9
    print(f"{fn.__name__}: median={time_diffs.median() * 1e6:.2f} µs, std={time_diffs.std() * 1e6:.2f} µs")
```

```
minimum: median=69.79 µs, std=3.61 µs
unwrap_boilerplate: median=74.35 µs, std=5.30 µs
unwrap_with_helper_method: median=74.65 µs, std=5.57 µs
context_manager: median=76.62 µs, std=3.58 µs
baseline: median=94.84 µs, std=4.42 µs
```

Benchmarking shows that unwrapping is slightly faster, but the difference to the context manager is small enough to also opt for consistency with #6681 and use it instead.

pmeier:
Closes #6781 by going with the `self.as_subclass(torch.Tensor)` proposal. After this PR, there are no more `__torch_function__` calls in the whole pipeline.

In fact, it almost renders #6681 obsolete: `ndim`, `device`, and `dtype` are no longer accessed. Even `shape` is only accessed 4 times:

https://github.com/pytorch/vision/blob/246de0772c80eae435c0a562268d8d21ed7a27a2/torchvision/prototype/transforms/functional/_meta.py#L19-L22

https://github.com/pytorch/vision/blob/246de0772c80eae435c0a562268d8d21ed7a27a2/torchvision/prototype/transforms/functional/_meta.py#L40-L41

https://github.com/pytorch/vision/blob/246de0772c80eae435c0a562268d8d21ed7a27a2/torchvision/prototype/transforms/functional/_meta.py#L78-L79

https://github.com/pytorch/vision/blob/246de0772c80eae435c0a562268d8d21ed7a27a2/torchvision/prototype/transforms/functional/_meta.py#L91-L92

Since they all only indirectly access `.shape` through properties we have anyway

https://github.com/pytorch/vision/blob/246de0772c80eae435c0a562268d8d21ed7a27a2/torchvision/prototype/features/_image.py#L105-L111

https://github.com/pytorch/vision/blob/246de0772c80eae435c0a562268d8d21ed7a27a2/torchvision/prototype/features/_video.py#L57-L67

we could also opt to move the `with DisableTorchFunction()` context manager there and completely remove

https://github.com/pytorch/vision/blob/246de0772c80eae435c0a562268d8d21ed7a27a2/torchvision/prototype/features/_feature.py#L132-L152

### DIFF
--- a/torchvision/prototype/features/_bounding_box.py
+++ b/torchvision/prototype/features/_bounding_box.py
@@ -66,15 +66,23 @@ class BoundingBox(_Feature):
             format = BoundingBoxFormat.from_str(format.upper())
 
         return BoundingBox.wrap_like(
-            self, self._F.convert_format_bounding_box(self, old_format=self.format, new_format=format), format=format
+            self,
+            self._F.convert_format_bounding_box(
+                self.as_subclass(torch.Tensor).as_subclass(torch.Tensor), old_format=self.format, new_format=format
+            ),
+            format=format,
         )
 
     def horizontal_flip(self) -> BoundingBox:
-        output = self._F.horizontal_flip_bounding_box(self, format=self.format, spatial_size=self.spatial_size)
+        output = self._F.horizontal_flip_bounding_box(
+            self.as_subclass(torch.Tensor).as_subclass(torch.Tensor), format=self.format, spatial_size=self.spatial_size
+        )
         return BoundingBox.wrap_like(self, output)
 
     def vertical_flip(self) -> BoundingBox:
-        output = self._F.vertical_flip_bounding_box(self, format=self.format, spatial_size=self.spatial_size)
+        output = self._F.vertical_flip_bounding_box(
+            self.as_subclass(torch.Tensor).as_subclass(torch.Tensor), format=self.format, spatial_size=self.spatial_size
+        )
         return BoundingBox.wrap_like(self, output)
 
     def resize(  # type: ignore[override]
@@ -85,19 +93,19 @@ class BoundingBox(_Feature):
         antialias: bool = False,
     ) -> BoundingBox:
         output, spatial_size = self._F.resize_bounding_box(
-            self, spatial_size=self.spatial_size, size=size, max_size=max_size
+            self.as_subclass(torch.Tensor), spatial_size=self.spatial_size, size=size, max_size=max_size
         )
         return BoundingBox.wrap_like(self, output, spatial_size=spatial_size)
 
     def crop(self, top: int, left: int, height: int, width: int) -> BoundingBox:
         output, spatial_size = self._F.crop_bounding_box(
-            self, self.format, top=top, left=left, height=height, width=width
+            self.as_subclass(torch.Tensor), self.format, top=top, left=left, height=height, width=width
         )
         return BoundingBox.wrap_like(self, output, spatial_size=spatial_size)
 
     def center_crop(self, output_size: List[int]) -> BoundingBox:
         output, spatial_size = self._F.center_crop_bounding_box(
-            self, format=self.format, spatial_size=self.spatial_size, output_size=output_size
+            self.as_subclass(torch.Tensor), format=self.format, spatial_size=self.spatial_size, output_size=output_size
         )
         return BoundingBox.wrap_like(self, output, spatial_size=spatial_size)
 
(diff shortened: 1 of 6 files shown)


torchvision/prototype/features/_image.py
```diff
@@ -182,8 +195,8 @@ def rotate(
         fill: FillTypeJIT = None,
         center: Optional[List[float]] = None,
     ) -> Image:
-        output = self._F._geometry.rotate_image_tensor(
-            self, angle, interpolation=interpolation, expand=expand, fill=fill, center=center
+        output = self._F.rotate_image_tensor(
```
pmeier, line 198: The `_F._geometry` crept in with #6476. I only found this because my regex replace did not handle this case :smile:

torchvision/prototype/features/_bounding_box.py
```diff
@@ -66,15 +66,23 @@ def to_format(self, format: Union[str, BoundingBoxFormat]) -> BoundingBox:
             format = BoundingBoxFormat.from_str(format.upper())
 
         return BoundingBox.wrap_like(
-            self, self._F.convert_format_bounding_box(self, old_format=self.format, new_format=format), format=format
+            self,
+            self._F.convert_format_bounding_box(
+                self.as_subclass(torch.Tensor).as_subclass(torch.Tensor), old_format=self.format, new_format=format
```
datumbox, line 71: Why `as_subclass(torch.Tensor).as_subclass(torch.Tensor)`? Did you hit replace twice?

vfdev-5:
@pmeier can you provide time measurements on a datapipeline (e.g. image classification + AA + RandomErasing, https://github.com/pytorch/vision/pull/6681#issue-1394618006)

torchvision/prototype/features/_bounding_box.py
pmeier: Most likely when I was fixing https://github.com/pytorch/vision/pull/6807#discussion_r1001158327. Good catch though!

vfdev-5:
@pmeier by time measurements I meant torch benchmarks and not cprofile (which may be less reliable)

torchvision/prototype/features/_feature.py
```diff
@@ -92,6 +92,7 @@ def __torch_function__(
             return NotImplemented
 
         with DisableTorchFunction():
+            raise Exception(str(func), str(args), str(kwargs))
```
datumbox, line 95: This looks like something added for debugging?
```suggestion
```

datumbox (commented):
LGTM overall. Only one comment below. Also mypy seems to be [failing](https://app.circleci.com/pipelines/github/pytorch/vision/21315/workflows/98a1f3db-4760-4197-b15e-8e183914f64d/jobs/1728995) due to an unused ignore.

torchvision/prototype/features/_feature.py
pmeier: Ah yes, sorry too many things at once :facepalm:

datumbox (approved):
LGTM, thanks!

PR #6809, opened 2022-10-21
Revert "[proto] Enable GPU tests on prototype"

vfdev-5:
Reverts pytorch/vision#6665


datumbox (approved):
stamping to unblock. Though I don't understand why the previous PR [passed](https://github.com/pytorch/vision/actions/runs/3297036681/jobs/5437383454) the last time you merged on  main on it and now it fails. That's quite surprising to me.

PR #6811, opened 2022-10-21
[proto] Speed improvement for autocontrast op

vfdev-5:
Results:

```
[---------------- Autocontrast cpu torch.uint8 ---------------]
                     |  autocontrast stable  |  autocontrast v2
1 threads: ----------------------------------------------------
      (3, 400, 400)  |          969          |        844      
6 threads: ----------------------------------------------------
      (3, 400, 400)  |          320          |        274      

[--------------- Autocontrast cuda torch.uint8 ---------------]
                     |  autocontrast stable  |  autocontrast v2
1 threads: ----------------------------------------------------
      (3, 400, 400)  |          200          |        150      
6 threads: ----------------------------------------------------
      (3, 400, 400)  |          201          |        149      

[--------------- Autocontrast cpu torch.float32 --------------]
                     |  autocontrast stable  |  autocontrast v2
1 threads: ----------------------------------------------------
      (3, 400, 400)  |          529          |        418      
6 threads: ----------------------------------------------------
      (3, 400, 400)  |          201          |        157      

[-------------- Autocontrast cuda torch.float32 --------------]
                     |  autocontrast stable  |  autocontrast v2
1 threads: ----------------------------------------------------
      (3, 400, 400)  |          166          |        115      
6 threads: ----------------------------------------------------
      (3, 400, 400)  |          165          |        114      

```

cc @datumbox @bjuncek @pmeier

### DIFF
--- a/torchvision/prototype/transforms/functional/_color.py
+++ b/torchvision/prototype/transforms/functional/_color.py
@@ -211,7 +211,43 @@ def solarize(inpt: features.InputTypeJIT, threshold: float) -> features.InputTyp
         return solarize_image_pil(inpt, threshold=threshold)
 
 
-autocontrast_image_tensor = _FT.autocontrast
+def autocontrast_image_tensor(image: torch.Tensor) -> torch.Tensor:
+
+    if not (isinstance(image, torch.Tensor)):
+        raise TypeError("Input img should be Tensor image")
+
+    c = get_num_channels_image_tensor(image)
+
+    if c not in [1, 3]:
+        raise TypeError(f"Input image tensor permitted channel values are {[1, 3]}, but found {c}")
+
+    # if img.ndim < 3:
+    #     raise TypeError(f"Input image tensor should have at least 3 dimensions, but found {img.ndim}")
+
+    # _assert_channels(img, [1, 3])
+
+    bound = 1.0 if image.is_floating_point() else 255.0
+    dtype = image.dtype if torch.is_floating_point(image) else torch.float32
+
+    # let's squash spatial dims into a single dim, such that torch.aminmax support it
+    shape = image.shape[:-2] + (image.shape[-2] * image.shape[-1], )
+    print("shape:", shape, image.numel())
+
+    minimum = torch.amin(image.view(shape), dim=(-2, -1), keepdim=True)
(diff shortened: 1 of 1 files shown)


torchvision/prototype/transforms/functional/_color.py
```diff
@@ -211,7 +211,34 @@ def solarize(inpt: features.InputTypeJIT, threshold: float) -> features.InputTyp
...
+
+    if image.numel() == 0:
+        # exit earlier on empty images
+        return image
+
+    bound = 1.0 if image.is_floating_point() else 255.0
+    dtype = image.dtype if torch.is_floating_point(image) else torch.float32
+
+    minimum = image.amin(dim=(-2, -1), keepdim=True).to(dtype)
+    maximum = image.amax(dim=(-2, -1), keepdim=True).to(dtype)
```
vfdev-5, line 232: We can use `torch.aminmax` here as it does not support tuple of dims. Reshaping the input does not improve runtime.

torchvision/prototype/transforms/functional/_color.py
```diff
@@ -211,7 +211,34 @@ def solarize(inpt: features.InputTypeJIT, threshold: float) -> features.InputTyp
...
+
+    scale = bound / (maximum - minimum)
+    eq_idxs = maximum == minimum
```
datumbox, line 235: Reading back to original code that I wrote, I see I've used `eq_idxs = torch.isfinite(scale).logical_not()` for the same thing. To be honest, I don't understand why I did that (it looks kind of stupid). Just to be safe, you tested this logic with special cases where the values are the same across different channels and the result is identical right?

From code I expect that your approach is faster and better. Just being a bit paranoid.

datumbox (approved):
LGTM, thanks! Just one question:

torchvision/prototype/transforms/functional/_color.py
vfdev-5: Thanks for asking, I'll check with more attention to that

vfdev-5: I did few other checks and I think `eq_idxs = maximum == minimum` should be equivalent to `eq_idxs = torch.isfinite(scale).logical_not()`. So, we can keep PR's implementation.

PR #6812, opened 2022-10-21
[WIP] Adding Uniform temporal sampling

datumbox:
Adding temporal sampling Transforms similar to the ones on [PyTorch Video](https://github.com/facebookresearch/pytorchvideo/blob/a0a131e/pytorchvideo/transforms/functional.py#L19).

### DIFF
--- a/torchvision/prototype/transforms/functional/__init__.py
+++ b/torchvision/prototype/transforms/functional/__init__.py
@@ -172,3 +172,4 @@ from ._type_conversion import (
 )
 
 from ._deprecated import get_image_size, rgb_to_grayscale, to_grayscale, to_tensor  # usort: skip
+from ._temporal import uniform_temporal_subsample_video, uniform_temporal_subsample  # usort: skip
--- a/torchvision/prototype/transforms/functional/_temporal.py
+++ b/torchvision/prototype/transforms/functional/_temporal.py
@@ -1,0 +1,23 @@
+import torch
+
+from torchvision.prototype import features
+
+
+def uniform_temporal_subsample(
+    inpt: features.TensorVideoTypeJIT, num_samples: int, temporal_dim: int = -4
+) -> features.InputTypeJIT:
+    if isinstance(inpt, torch.Tensor) and (torch.jit.is_scripting() or not isinstance(inpt, features.Video)):
+        return uniform_temporal_subsample_video(inpt, num_samples, temporal_dim=temporal_dim)
+    else:  # isinstance(inpt, features.Video)
+        if temporal_dim != -4:
+            raise ValueError("Video inputs must have temporal_dim equal to -4")
+        output = uniform_temporal_subsample_video(
+            inpt.as_subclass(torch.Tensor), num_samples, temporal_dim=temporal_dim
+        )
+        return features.Video.wrap_like(inpt, output)
+
+
+def uniform_temporal_subsample_video(video: torch.Tensor, num_samples: int, temporal_dim: int = -4) -> torch.Tensor:
+    t_max = video.size(temporal_dim) - 1
+    indices = torch.linspace(0, t_max, num_samples, device=video.device).clamp_(0, t_max).long()
+    return torch.index_select(video, temporal_dim, indices)


torchvision/prototype/transforms/functional/_temporal.py
```diff
@@ -0,0 +1,24 @@
+import torch
+
+from torchvision.prototype import features
+
+
+def uniform_temporal_subsample(
```
pmeier, line 6: OCD nit: so far we placed the dispatchers below the kernels :innocent:

torchvision/prototype/transforms/functional/_temporal.py
```diff
@@ -0,0 +1,24 @@
...
+    inpt: features.VideoTypeJIT, num_samples: int, temporal_dim: int = -4
+) -> features.VideoTypeJIT:
+    if isinstance(inpt, torch.Tensor) and (torch.jit.is_scripting() or not isinstance(inpt, features.Video)):
+        return uniform_temporal_subsample_video(inpt, num_samples, temporal_dim=temporal_dim)
+    else:  # isinstance(inpt, features.Video)
+        if temporal_dim != -4:
```
pmeier, line 12: Do we want to force users to use negative indexing here? What if I do `temporal_dim=0`. That is the same as `temporal_dim=-4` for `(T, C, H, W)` inputs.

torchvision/prototype/transforms/functional/_temporal.py
```diff
@@ -0,0 +1,24 @@
...
+        output = uniform_temporal_subsample_video(
+            inpt.as_subclass(torch.Tensor), num_samples, temporal_dim=temporal_dim
+        )
+        return features.Video.wrap_like(inpt, output)
+
+
+def uniform_temporal_subsample_video(video: torch.Tensor, num_samples: int, temporal_dim: int = -4) -> torch.Tensor:
+    # Reference: https://github.com/facebookresearch/pytorchvideo/blob/a0a131e/pytorchvideo/transforms/functional.py#L19
+    t_max = video.size(temporal_dim) - 1
+    indices = torch.linspace(0, t_max, num_samples, device=video.device).clamp_(0, t_max).long()
```
pmeier, line 23: Not sure if this is just for parity, but is there a reason to create the linspace in `int32` to later convert it to `int64`?

pmeier: Plus, maybe this is just personal preference, but I would appreciate us to be "explicit" about the types

```suggestion
    indices = torch.linspace(0, t_max, num_samples, device=video.device).clamp_(0, t_max).to(torch.int64)
```

pmeier: Plusplus, the docs for [`torch.linspace`](https://pytorch.org/docs/stable/generated/torch.linspace.html) state:

> Creates a one-dimensional tensor of size `steps` whose values are evenly spaced from `start` to `end`, inclusive.

What is the clamp for if the function by definition does not return values outside this range?

torchvision/prototype/transforms/functional/_temporal.py
```diff
@@ -0,0 +1,24 @@
...
+            raise ValueError("Video inputs must have temporal_dim equal to -4")
+        output = uniform_temporal_subsample_video(
+            inpt.as_subclass(torch.Tensor), num_samples, temporal_dim=temporal_dim
+        )
+        return features.Video.wrap_like(inpt, output)
+
+
+def uniform_temporal_subsample_video(video: torch.Tensor, num_samples: int, temporal_dim: int = -4) -> torch.Tensor:
+    # Reference: https://github.com/facebookresearch/pytorchvideo/blob/a0a131e/pytorchvideo/transforms/functional.py#L19
+    t_max = video.size(temporal_dim) - 1
```
pmeier, line 22: `Tensor.size()` is a somewhat outdated method. Plus, we don't use it anywhere so far. Just use `shape`

```suggestion
    t_max = video.shape[temporal_dim] - 1
```

torchvision/prototype/transforms/functional/__init__.py
```diff
@@ -172,3 +172,4 @@
 )
 
 from ._deprecated import get_image_size, rgb_to_grayscale, to_grayscale, to_tensor  # usort: skip
+from ._temporal import uniform_temporal_subsample_video, uniform_temporal_subsample  # usort: skip
```
pmeier, line 175: Why did you put it at the end of the file? Imports from `._deprecated` are at the end since they potentially need to import from the other namespaces and thus should be the last module to load. This is not true for `_temporal`, right? If so, please remove the `# usort: skip` directive and put it in the large block above.

torchvision/prototype/transforms/_temporal.py
```diff
@@ -0,0 +1,19 @@
+from typing import Any, Dict
+
+from torchvision.prototype import features
+from torchvision.prototype.transforms import functional as F, Transform
+
+
+class UniformTemporalSubsample(Transform):
+    _transformed_types = (
+        features.is_simple_tensor,
+        features.Video,
```
pmeier, line 10:
```suggestion
        features.Video
```

to prevent `black` from exploding this tuple

pmeier (commented):
I'm ok with having custom tests for this now, but they only cover correctness at this point. Unless this is super urgent, we should really implement a `KernelInfo` and `DispatcherInfo` for it to also cover stuff like JIT. @datumbox if you don't have time to add that LMK and I will push a commit.

datumbox:
@pmeier Thanks for the review.

I've addressed the comments to unblock. We should review on what degree these nits can be covered by the linter or adopt simpler conventions on the future. Regardless I've made the necessary changes to cut down back-and-forth as we need this ASAP to assist the migration of a couple of internal teams. 

Concerning the tests, I was hoping for more pointers from you on where I'm supposed to add them (the current structure has a lot of abstraction and it's not obvious). Concerning JIT testing, I have a test but it's probably on the wrong place. You are welcome to push a commit on this branch to improve on the tests as you are more familiar with the new test infra but if that is going to take longer, we should do it on a separate PR.

torchvision/prototype/transforms/functional/_temporal.py
datumbox: Let's be careful on how we are porting this. Here is why casting later is not the same:
```python
t_max = 9
num_samples = 8
indices = torch.linspace(0, t_max, num_samples, dtype=torch.int64)
indices2 = torch.linspace(0, t_max, num_samples).clamp(0, t_max).long()
assert indices.equal(indices2), f"({t_max}, {num_samples})\n{indices}\n{indices2}"
```
Result:
```
    assert indices.equal(indices2), f"({t_max}, {num_samples})\n{indices}\n{indices2}"
AssertionError: (9, 8)
tensor([0, 1, 2, 3, 5, 6, 7, 8])
tensor([0, 1, 2, 3, 5, 6, 7, 9])
```

test/prototype_transforms_kernel_infos.py
```diff
@@ -2100,3 +2118,39 @@ def sample_inputs_convert_dtype_video():
         ),
     ]
 )
+
+
+def sample_inputs_uniform_temporal_subsample_video():
```
pmeier, line 2123: Defining a `sample_inputs_fn` for a `KernelInfo` is the only non-optional thing that you need to define. Per comment

https://github.com/pytorch/vision/blob/9c112935abe400222cca8f9fbc2d8386e0f25e80/test/prototype_transforms_kernel_infos.py#L42-L44

these inputs will be used for the following tests:

- https://github.com/pytorch/vision/blob/9c112935abe400222cca8f9fbc2d8386e0f25e80/test/test_prototype_transforms_functional.py#L92
- https://github.com/pytorch/vision/blob/9c112935abe400222cca8f9fbc2d8386e0f25e80/test/test_prototype_transforms_functional.py#L122
- https://github.com/pytorch/vision/blob/9c112935abe400222cca8f9fbc2d8386e0f25e80/test/test_prototype_transforms_functional.py#L157
- https://github.com/pytorch/vision/blob/9c112935abe400222cca8f9fbc2d8386e0f25e80/test/test_prototype_transforms_functional.py#L170
- https://github.com/pytorch/vision/blob/9c112935abe400222cca8f9fbc2d8386e0f25e80/test/test_prototype_transforms_functional.py#L181

This function should either `yield` individual `ArgsKwargs` or `return` a sequence of them. We have been using the `yield` variant for now since it is less verbose, but it really doesn't matter since the output will be converted to a `list` by the machinery anyway.

test/prototype_transforms_kernel_infos.py
```diff
@@ -2100,3 +2118,39 @@ def sample_inputs_convert_dtype_video():
...
+    for video_loader in make_video_loaders(sizes=["random"], num_frames=[4]):
+        for temporal_dim in [-4, len(video_loader.shape) - 4]:
+            yield ArgsKwargs(video_loader, num_samples=2, temporal_dim=temporal_dim)
```
pmeier, line 2126: As shown above, the inputs defined here should cover all valid code paths but should otherwise not be excessive. Since there no branches in the whole kernel, the definition is trivial. I simply used `make_video_loaders` and restricted the videos to random spatial sizes and a fixed `num_frames`. On top of that I also added positive `temporal_dim`'s to the mix.

In total we are looking at 

https://github.com/pytorch/vision/blob/9c112935abe400222cca8f9fbc2d8386e0f25e80/test/prototype_common_utils.py#L582-L592

1 `sizes`, 2 `color_space`, 1 `num_frames`, 6 `extra_dims`

https://github.com/pytorch/vision/blob/9c112935abe400222cca8f9fbc2d8386e0f25e80/test/prototype_common_utils.py#L216-L219

1 `dtype`, and 2 `temporal_dim`'s or 24 total sample inputs. That is totally fine and even on the lower end of sample inputs compared to more complicated kernels.

If you are wondering why we are using the `make_video_loaders` here instead of `make_videos`, the reason is rather simple. These sample inputs will be loaded at test collection time and are alive for the whole duration of running the test suite. With this PR, we instantiate ~55k tests. We don't want to instantiate that many tensors. Thus, we are using the `TensorLoader` concept that knows how to instantiate the tensor and already has some metadata about it. For example, you can see that I can access `video_loader.shape` during the sample generation, but the actual values are unknown at this point. The tensor will only be instantiated inside the test with this call

https://github.com/pytorch/vision/blob/9c112935abe400222cca8f9fbc2d8386e0f25e80/test/test_prototype_transforms_functional.py#L96

test/prototype_transforms_kernel_infos.py
```diff
@@ -2100,3 +2118,39 @@ def sample_inputs_convert_dtype_video():
...
+
+
+def reference_uniform_temporal_subsample_video(video, num_samples, temporal_dim=-4):
+    from pytorchvideo.transforms.functional import uniform_temporal_subsample
+
+    return uniform_temporal_subsample(video, num_samples=num_samples, temporal_dim=temporal_dim)
```
pmeier, line 2132: If there is a good reference function, we can optionally define a `reference_fn`

https://github.com/pytorch/vision/blob/9c112935abe400222cca8f9fbc2d8386e0f25e80/test/prototype_transforms_kernel_infos.py#L45-L49

In our case we have the perfect reference as we want to mirror the function of `pytorchvideo`. We import it lazily here since we shouldn't expect it to be available when this module is collected. I'll explain what else we need to do to skip this test in case we don't have `pytorchvideo` below.

.github/workflows/prototype-tests.yml
```diff
@@ -34,7 +34,7 @@ jobs:
         run: pip install --progress-bar=off --no-build-isolation --editable .
 
       - name: Install other prototype dependencies
-        run: pip install --progress-bar=off scipy pycocotools h5py iopath
+        run: pip install --progress-bar=off scipy pycocotools h5py iopath pytorchvideo
```
pmeier, line 37: Since we want to run the reference test in CI, we add `pytorchvideo` to the third party packages we install. Since it is a pure Python library, there is no need to install it with the other PyTorch packages like `torch` or `torchdata` above.

test/prototype_transforms_kernel_infos.py
```diff
@@ -2100,3 +2118,39 @@ def sample_inputs_convert_dtype_video():
...
+
+
+def reference_inputs_uniform_temporal_subsample_video():
+    for video_loader in make_video_loaders(sizes=["random"], color_spaces=[features.ColorSpace.RGB], num_frames=[10]):
+        for num_frames in range(1, 11):
+            yield ArgsKwargs(video_loader, num_frames)
```
pmeier, line 2138: The reference test as a separate set of inputs that will be tested

https://github.com/pytorch/vision/blob/9c112935abe400222cca8f9fbc2d8386e0f25e80/test/prototype_transforms_kernel_infos.py#L50-L52

That has two reasons:

1. Sometimes the reference doesn't support all the inputs our kernel supports. For example, using a PIL kernel as reference requires the input to be unbatched and thus you will often see `extra_dims=[()]` in the reference inputs for kernels that have a PIL reference.
2. There is only one reference test and thus all the considerations of keeping the number of inputs low that we made for the sample inputs are off the table here. This can be as comprehensive as we need it for us to be sure our kernel is doing the right thing.

Of course the kernel under test here is pretty simple and thus the reference inputs don't look that much different from the sample inputs. I again opted to use a fixed spatial size and number of frames. This time I also used only one color space since and one temporal dim (the default one) since any errors regarding that will already be caught by the other tests. In contrast, I here varied the number of frames that will be subsampled very fine grained to make sure our rounding is doing exactly the same as the reference.

test/prototype_transforms_kernel_infos.py
```diff
@@ -2100,3 +2118,39 @@ def sample_inputs_convert_dtype_video():
...
+        test_marks=[
+            skip_on_missing_third_party_reference("pytorchvideo"),
+            TestMark(
+                ("TestKernels", "test_batched_vs_single"),
+                pytest.mark.skip("Positive `temporal_dim` arguments are not equivalent for batched and single inputs"),
+                condition=lambda args_kwargs: args_kwargs.kwargs.get("temporal_dim") >= 0,
+            ),
+        ],
+    )
+)
```
pmeier, line 2156: Finally, we need to bring everything above together. The information is bundled inside a `KernelInfo` and appended to long `KERNEL_INFOS` list that holds all defined `KernelInfo`'s in this file. The only thing new here are the `test_marks`. This field lets us conditionally apply `pytest.mark.*`'s to specific tests. There are two defined here:

1. If `pytorchvideo` is not available, we want to skip the reference test. Otherwise, calling `reference_uniform_temporal_subsample_video` will fail. Instead of doing this through the `test_marks`, we could have also used `pytorchvideo = pytest.importorskip("pytorchvideo")` inside the reference function. That has ultimately the same effect since the test will be skipped, but has a major disadvantage here: the test will only be skipped when the reference function is called. Looking at the implementation of the test

   https://github.com/pytorch/vision/blob/9c112935abe400222cca8f9fbc2d8386e0f25e80/test/test_prototype_transforms_functional.py#L193-L197

   at this point we have already instantiated the inputs as well as already called our kernel. Thus, compared to just skipping before the test is run in the first place, we are wasting resources here.

2. The `test_batched_vs_single` tests whether the output is identical for batched inputs or the same inputs, but with unbound batch dimensions. This works fine here if we use negative indices for `temporal_dim` since the "data dimensions" 

   https://github.com/pytorch/vision/blob/9c112935abe400222cca8f9fbc2d8386e0f25e80/test/test_prototype_transforms_functional.py#L126-L137

   will not be touched by the test. However, using positive indices is not possible, since that would need to be fixed after unbatching. Thus, we skip this test conditionally in case we encounter positive indices.

test/prototype_transforms_dispatcher_infos.py
```diff
@@ -426,4 +426,13 @@ def xfail_all_tests(*, reason, condition):
...
     ),
+    DispatcherInfo(
+        F.uniform_temporal_subsample,
+        kernels={
+            features.Video: F.uniform_temporal_subsample_video,
+        },
+        test_marks=[
+            skip_dispatch_feature,
+        ],
+    ),
```
pmeier, line 437: After the `KernelInfo` is added, adding the corresponding `DispatcherInfo` is much simpler. We only need to define which kernel will be dispatched to for which feature type and the inputs will be automatically picked up from the corresponding `KernelInfo`'s. In our case we only need add one predefined `TestMark` to skip the test checking if the dispatch to the corresponding method on the feature class works, since this dispatcher does not have such a method.

test/test_prototype_transforms_functional.py
```diff
@@ -1060,3 +1061,13 @@ def test_equalize_image_tensor_edge_cases():
...
+
+
+@pytest.mark.parametrize("device", cpu_and_gpu())
+def test_correctness_uniform_temporal_subsample(device):
+    video = torch.arange(10, device=device)[:, None, None, None].expand(-1, 3, 8, 8)
+    out_video = F.uniform_temporal_subsample(video, 5)
+    assert out_video.unique().tolist() == [0, 2, 4, 6, 9]
+
+    out_video = F.uniform_temporal_subsample(video, 8)
+    assert out_video.unique().tolist() == [0, 1, 2, 3, 5, 6, 7, 9]
```
pmeier, line 1073: I've left this test in for now, but it is probably no longer needed. I tried to re-introduce my former suggestion of using `torch.linspace(..., dtype=torch.int64)` instead of `torch.linspace(...).to(torch.int64)` and the reference test triggers correctly.

test/test_prototype_transforms.py
```diff
@@ -1899,3 +1899,22 @@ def test_transpose_dimensions(dims):
             assert type(transformed_value) == torch.Tensor
         else:
             assert transformed_value is value
+
+
+class TestUniformTemporalSubsample:
```
pmeier, line 1904: We currently don't have automated tests for testing the transforms that wraps a kernel. Thus, I don't have any objections here. I'm going to look into more structured tests for this as soon as the we pass the tight deadlines.

pmeier (commented):
@datumbox I've added the tests I've wanted and over-explained them in my comments below. To run these tests, you can use

```sh
pytest test/test_prototype_transforms_functional.py -k "uniform_temporal_subsample"
```

This will run ~150 tests. LMK if any of my explanations is unclear or I missed something that you want to know.

torchvision/prototype/transforms/functional/_temporal.py
```diff
@@ -0,0 +1,24 @@
+import torch
+
+from torchvision.prototype import features
+
+
+def uniform_temporal_subsample_video(video: torch.Tensor, num_samples: int, temporal_dim: int = -4) -> torch.Tensor:
+    # Reference: https://github.com/facebookresearch/pytorchvideo/blob/a0a131e/pytorchvideo/transforms/functional.py#L19
+    t_max = video.shape[temporal_dim] - 1
+    indices = torch.linspace(0, t_max, num_samples, device=video.device).long()
```
pmeier, line 9: Can we do one of two things here?

1. Rename this to `num_frames` since that is the term we are using throughout the test suite and is more descriptive.
2. (preferred) Merge this into the line below since the variable is only used once after we removed the `.clamp_()` call

```suggestion
    indices = torch.linspace(0, video.shape[temporal_dim] - 1, num_samples, device=video.device).long()
```

test/test_prototype_transforms_functional.py
pmeier: If you decide to keep this test, could you add a new line to the end of the file here in case there is actually missing one? GH shows that there is none, but I'm a little confused why our linters didn't fix or trigger on that.

pmeier (approved):
There is one [relevant test failure](https://github.com/pytorch/vision/actions/runs/3307188627/jobs/5458625371#step:12:986):

```
_____________ TestUniformTemporalSubsample.test__transform[inpt2] ______________
  File "/home/runner/work/vision/vision/test/test_prototype_transforms.py", line 1917, in test__transform
    output = transform(inpt)
  File "/opt/hostedtoolcache/Python/3.7.15/x64/lib/python3.7/site-packages/torch/nn/modules/module.py", line 1363, in _call_impl
  File "/home/runner/work/vision/vision/torchvision/prototype/transforms/_transform.py", line 40, in forward
    for inpt in flat_inputs
  File "/home/runner/work/vision/vision/torchvision/prototype/transforms/_transform.py", line 40, in <listcomp>
    for inpt in flat_inputs
  File "/home/runner/work/vision/vision/torchvision/prototype/transforms/_temporal.py", line 16, in _transform
    return F.uniform_temporal_subsample(inpt, self.num_samples, temporal_dim=self.temporal_dim)
  File "/home/runner/work/vision/vision/torchvision/prototype/transforms/functional/_temporal.py", line 20, in uniform_temporal_subsample
    raise ValueError("Video inputs must have temporal_dim equivalent to -4")
ValueError: Video inputs must have temporal_dim equivalent to -4
```

Otherwise LGTM if CI is green. Thanks Vasilis!

torchvision/prototype/transforms/functional/_temporal.py
```diff
@@ -0,0 +1,24 @@
...
+    return torch.index_select(video, temporal_dim, indices)
+
+
+def uniform_temporal_subsample(
+    inpt: features.VideoTypeJIT, num_samples: int, temporal_dim: int = -4
+) -> features.VideoTypeJIT:
+    if isinstance(inpt, torch.Tensor) and (torch.jit.is_scripting() or not isinstance(inpt, features.Video)):
+        return uniform_temporal_subsample_video(inpt, num_samples, temporal_dim=temporal_dim)
+    else:  # isinstance(inpt, features.Video)
+        if not (temporal_dim == -4 or inpt.ndim - 4 == temporal_dim):
```
pmeier, line 19: Previously you did `temporal_dim != -4 or inpt.ndim - 4 != temporal_dim`, but since `or` triggers on the first true, any `temporal_dim != -4` will trigger this error.

test/test_prototype_transforms.py
```diff
@@ -1899,3 +1899,22 @@ def test_transpose_dimensions(dims):
...
+            torch.zeros(1, 10, 3, 8, 8),
+            features.Video(torch.zeros(1, 10, 3, 8, 8)),
+        ],
+    )
+    def test__transform(self, inpt):
+        num_samples = 5
+        transform = transforms.UniformTemporalSubsample(num_samples)
+
+        output = transform(inpt)
+        assert type(output) is type(inpt)
```
pmeier, line 1918: This was not strict enough before. You did `assert isinstance(output, type(inpt))`. This would also pass if the output is a `features.Video` in case the input was a `torch.Tensor`. This is not that far fetched of a mistake in the transform and thus I made the check more strict. With the new check, we assert that the type is retained.

pmeier (approved):
Fixed the bug mentioned above for you. PR is now ready to be merged from my side unless you have any objections about the stuff I added.

torchvision/prototype/transforms/functional/_temporal.py
vfdev-5: Can we pass dtype directly into linspace method ?

.github/workflows/prototype-tests.yml
datumbox: Let's remove the dependency. Adding this is a bit of an overkill; I understand that this is on the tests only but pytorchvideo is a research library which breaks from time to time and also has a cyclical dependency on TorchVision. If it's absolutely needed to have a reference implementation, we could copy it over on our references scripts.

torchvision/prototype/transforms/functional/_temporal.py
datumbox, line 19:
```suggestion
        if temporal_dim != -4 and inpt.ndim - 4 != temporal_dim:
```

torchvision/prototype/transforms/functional/_temporal.py
datumbox: Unfortunately no. See https://github.com/pytorch/vision/pull/6812#discussion_r1002437589

torchvision/prototype/transforms/functional/_temporal.py
vfdev-5: IMO, this looks like a bug in torch linspace as its docs states that the last element is `end` (=t_max)

.github/workflows/prototype-tests.yml
pmeier: It is optional as explained above, but I don't see a reason to not have it. I mean things like https://github.com/pytorch/vision/pull/6812#discussion_r1002437589 make it clear that there are subtle things that we need to watch out for when refactoring the kernel. Why not encode the "right" behavior in an unoptimized reference function?

pmeier: I've removed `pytorchvideo` in 82c0aeebd636c8bb75977e007b9bfb340246f5b8 and copy-pasted the implementation into the reference function.

Issue #6818, opened 2022-10-24
Performance improvements for transforms v2 vs. v1


pmeier:
In addition to a lot of other goodies that transforms v2 will bring, we are also actively working on improving the performance. This is a tracker / overview issue of our progress.

Performance was measured with this [benchmark script](https://gist.github.com/pmeier/e0f1ea77c9cf75b682d7f30366a89bf8). Unless noted otherwise, the performance improvements reported above were computed on uint8, RGB images and videos while running single-threaded on CPU. You can find the full benchmark results alongside the benchmark script. The results will be constantly updated if new PRs are merged that have an effect on the kernels.

Kernels: 

- color
  - [x] `adjust_brightness` #6784
  - [x] `adjust_contrast` #6784 #6933
  - [x] `adjust_gamma` #6820 #6903
  - [x] `adjust_hue` #6805 #6903 #6938
  - [x] `adjust_saturation` #6784 #6940
  - [x] `adjust_sharpness` #6784 #6930
  - [x] `autocontrast` #6811 #6935 #6942
  - [x] `equalize` #6738, #6757, #6776
  - [x] `invert` #6819
  - [x] `posterize` #6823, #6847
  - [x] `solarize` #6819
- geometry
  - [x] `affine` #6949
  - [x] `center_crop` #6880 #6949
  - [x] `crop` #6949
  - [x] `elastic` #6942
  - [x] `erase` #6983
  - [x] `five_crop`: Composite kernel #6949
  - [x] `pad` #6949
  - [x] `perspective` #6907 #6949
  - [x] `resize` #6892
  - [x] `resized_crop`: Composite kernel #6892 #6949
  - [x] `rotate` #6949
  - [x] `ten_crop`: Composite kernel #6949
- meta
  - [x] `convert_color_space` #6784 #6832
  - [x] `convert_dtype` #6795 #6903
    - There is still some performance gain left for `int` to `int` conversion. Currently, we are using a multiplication 
      but theoretically bit shifts are faster. However, on PyTorch core the CPU kernels for bit shifts are not 
      vectorized making them slower for regular sized images than a multiplication. pytorch/pytorch#88607
- misc
  - [x] `gaussian_blur` #6762 #6888
  - [x] `normalize` #6821

Transform Classes:

- [x] MixUp/CutMix #6835
- [x] ColorJitter, RandomPhotometricDistort #6837

C++ (PyTorch core):

[...]

PR #6819, opened 2022-10-24
improve performance of invert_image_tensor

pmeier:
Per title. The benchmark script from #6818 yields the following

```
[--------- invert @ torchvision==0.15.0a0+62da7d4 ---------]
                                            |   v1   |   v2 
1 threads: -------------------------------------------------
      (1, 512, 512)       / uint8   / cpu   |    84  |     8
      (1, 512, 512)       / uint8   / cuda  |    20  |     4
      (1, 512, 512)       / float32 / cpu   |    28  |    27
      (1, 512, 512)       / float32 / cuda  |    20  |     6
      (3, 512, 512)       / uint8   / cpu   |   230  |    18
      (3, 512, 512)       / uint8   / cuda  |    30  |     4
      (3, 512, 512)       / float32 / cpu   |    61  |    62
      (3, 512, 512)       / float32 / cuda  |    40  |    27
      (5, 3, 512, 512)    / uint8   / cpu   |  1100  |    75
      (5, 3, 512, 512)    / uint8   / cuda  |    91  |    34
      (5, 3, 512, 512)    / float32 / cpu   |   500  |   500
      (5, 3, 512, 512)    / float32 / cuda  |   144  |   134
      (4, 5, 3, 512, 512) / uint8   / cpu   |  4400  |   400
      (4, 5, 3, 512, 512) / uint8   / cuda  |   400  |   133
      (4, 5, 3, 512, 512) / float32 / cpu   |  8000  |  7100
      (4, 5, 3, 512, 512) / float32 / cuda  |   540  |   535
2 threads: -------------------------------------------------
      (1, 512, 512)       / uint8   / cpu   |    49  |     7
      (1, 512, 512)       / float32 / cpu   |    21  |    20
      (3, 512, 512)       / uint8   / cpu   |   100  |    13
      (3, 512, 512)       / float32 / cpu   |    40  |    38
      (5, 3, 512, 512)    / uint8   / cpu   |   587  |    42
      (5, 3, 512, 512)    / float32 / cpu   |   150  |   150
      (4, 5, 3, 512, 512) / uint8   / cpu   |  2300  |   140
      (4, 5, 3, 512, 512) / float32 / cpu   |  7000  |  6600
4 threads: -------------------------------------------------
      (1, 512, 512)       / uint8   / cpu   |    30  |     6
      (1, 512, 512)       / float32 / cpu   |    16  |    14
      (3, 512, 512)       / uint8   / cpu   |    70  |    10
      (3, 512, 512)       / float32 / cpu   |    26  |    25
      (5, 3, 512, 512)    / uint8   / cpu   |   300  |    25
      (5, 3, 512, 512)    / float32 / cpu   |    83  |    90
      (4, 5, 3, 512, 512) / uint8   / cpu   |  1170  |    80
      (4, 5, 3, 512, 512) / float32 / cpu   |  6000  |  6000

Performance increased for images by roughly 12.5x.
Performance increased for videos by roughly 14.9x.
```

`uint8` is much faster now, while `float32` is on par with what we had before.

### DIFF
--- a/torchvision/prototype/transforms/functional/_color.py
+++ b/torchvision/prototype/transforms/functional/_color.py
@@ -435,7 +435,14 @@ def equalize(inpt: features.InputTypeJIT) -> features.InputTypeJIT:
         return equalize_image_pil(inpt)
 
 
-invert_image_tensor = _FT.invert
+def invert_image_tensor(image: torch.Tensor):
+    num_channels, height, width = get_dimensions_image_tensor(image)
+    if num_channels not in (1, 3):
+        raise TypeError(f"Input image tensor can have 1 or 3 channels, but found {num_channels}")
+
+    return 1.0 - image if image.is_floating_point() else image.bitwise_not()
+
+
 invert_image_pil = _FP.invert
 
 


torchvision/prototype/transforms/functional/_color.py
```diff
@@ -435,7 +435,14 @@ def equalize(inpt: features.InputTypeJIT) -> features.InputTypeJIT:
         return equalize_image_pil(inpt)
 
 
-invert_image_tensor = _FT.invert
+def invert_image_tensor(image: torch.Tensor):
+    num_channels, height, width = get_dimensions_image_tensor(image)
+    if num_channels not in (1, 3):
+        raise TypeError(f"Input image tensor can have 1 or 3 channels, but found {num_channels}")
```
pmeier, line 441: Not sure why we have this check since the number of channels is irrelevant here.

torchvision/prototype/transforms/functional/_color.py
```diff
@@ -435,7 +435,14 @@ def equalize(inpt: features.InputTypeJIT) -> features.InputTypeJIT:
...
+
+    return 1.0 - image if image.is_floating_point() else image.bitwise_not()
```
pmeier, line 443: The bitwise invert only works for `uint8`, but so does the v1 kernel

https://github.com/pytorch/vision/blob/62da7d48872ef6b9a4a907bbb6b866b39c24120a/torchvision/transforms/functional_tensor.py#L778-L779

Should we maybe add a guard that only accepts floating point and uint8 inputs?

torchvision/prototype/transforms/functional/_color.py
vfdev-5: Yes, makes sense to remove that, just check for input as tensor ?

torchvision/prototype/transforms/functional/_color.py
vfdev-5: Maybe we can keep previous logic and accept all dtypes as previously and make uint8 as a special case ?
Accepting only float and uint8 will be bc breaking

datumbox: If I understand correctly, Victors remarks is in relation to other integer types other than uint8. I'm OK do what he proposes, though I'm not sure if the original method truly handles correctly such `int16` etc, as it assumes their maximum value is 255. No strong opinions on my side.

datumbox (approved):
It's looking good. Perhaps address @vfdev-5 remark below, to ensure the previous and new behaviours are identical?

Worth noting that you also improve `solarize` which depends on `invert`. To get this speed gain, you would have to implement solarize again on v2, to call the new kernel. Do you plan to do this now or separately?

torchvision/prototype/transforms/functional/_color.py
datumbox: I got a similar situation here: https://github.com/pytorch/vision/pull/6820/files#r1003210245

Should I remove it?

datumbox:
There is a [related](https://app.circleci.com/pipelines/github/pytorch/vision/21402/workflows/1be1783d-9eb0-476f-934d-a68e2b3c5289/jobs/1735725) linter issue.

torchvision/prototype/transforms/functional/_color.py
pmeier: > I'm not sure if the original method truly handles correctly such `int16` etc, as it assumes their maximum value is 255. No strong opinions on my side.

It does not. The behavior is hardcoded for uint8. I'll implement it properly which can be regarded as bug fix.

torchvision/prototype/transforms/functional/_color.py
```diff
@@ -435,7 +435,15 @@ def equalize(inpt: features.InputTypeJIT) -> features.InputTypeJIT:
...
 
 
-invert_image_tensor = _FT.invert
+def invert_image_tensor(image: torch.Tensor):
+    _FT._assert_image_tensor(image)
+
+    if image.dtype == torch.uint8:
+        return image.bitwise_not()
+    else:
+        return _FT._max_value(image.dtype) - image
```
pmeier, line 444: I couldn't believe it, but this seems to be fastest way:

```py
def scalar_sub(image):
    return _max_value(image.dtype) - image


def new_full(image):
    return image.new_full(image.shape, _max_value(image.dtype)).sub_(image)


def full_like(image):
    return torch.full_like(image, _max_value(image.dtype)).sub_(image)
```

```
[-------------------- invert float or signed integer tensors -------------------]
                                         |  scalar_sub  |  new_full  |  full_like
1 threads: ----------------------------------------------------------------------
      (3, 256, 256)    / int32   / cpu   |      20      |     28     |      26   
      (3, 256, 256)    / int32   / cuda  |       5      |      8     |       7   
      (3, 256, 256)    / int64   / cpu   |      88      |    109     |     107   
      (3, 256, 256)    / int64   / cuda  |      13      |     22     |      22   
      (3, 256, 256)    / float64 / cpu   |      35      |     49     |      49   
      (3, 256, 256)    / float64 / cuda  |      12      |     21     |      21   
      (5, 3, 256, 256) / int32   / cpu   |      76      |    112     |     110   
      (5, 3, 256, 256) / int32   / cuda  |      34      |     67     |      67   
      (5, 3, 256, 256) / int64   / cpu   |     439      |    510     |     524   
      (5, 3, 256, 256) / int64   / cuda  |      68      |    133     |     133   
      (5, 3, 256, 256) / float64 / cpu   |     140      |    215     |     220   
      (5, 3, 256, 256) / float64 / cuda  |      68      |    133     |     133   

```

pmeier:
Re-run a reduced version of the benchmark with the recent changes:

```
[-------- invert @ torchvision==0.15.0a0+62da7d4 --------]
                                            |   v1   |  v2
1 threads: -----------------------------------------------
      (3, 512, 512)       / uint8   / cpu   |   236  |  17
      (3, 512, 512)       / uint8   / cuda  |    28  |   4
      (5, 3, 512, 512)    / uint8   / cpu   |  1140  |  70
      (5, 3, 512, 512)    / uint8   / cuda  |    93  |  34
2 threads: -----------------------------------------------
      (3, 512, 512)       / uint8   / cpu   |   126  |  12
      (5, 3, 512, 512)    / uint8   / cpu   |   582  |  41
4 threads: -----------------------------------------------
      (3, 512, 512)       / uint8   / cpu   |    69  |   7
      (5, 3, 512, 512)    / uint8   / cpu   |   300  |  23

Performance increased for images by roughly 13.8x.
Performance increased for videos by roughly 16.2x.
```

```
[-------- solarize @ torchvision==0.15.0a0+62da7d4 --------]
                                            |   v1   |   v2 
1 threads: -------------------------------------------------
      (3, 512, 512)       / uint8   / cpu   |  1000  |  1000
      (3, 512, 512)       / uint8   / cuda  |    54  |    30
      (5, 3, 512, 512)    / uint8   / cpu   |  6000  |  5300
      (5, 3, 512, 512)    / uint8   / cuda  |   230  |   180
2 threads: -------------------------------------------------
      (3, 512, 512)       / uint8   / cpu   |  1000  |   600
      (5, 3, 512, 512)    / uint8   / cpu   |  4000  |  3000
4 threads: -------------------------------------------------
      (3, 512, 512)       / uint8   / cpu   |   668  |   582
      (5, 3, 512, 512)    / uint8   / cpu   |  3140  |  2900

Performance increased for images by roughly 21%.
Performance increased for videos by roughly 16%.
```

torchvision/prototype/transforms/functional/_color.py
```diff
@@ -312,7 +312,13 @@ def posterize(inpt: features.InputTypeJIT, bits: int) -> features.InputTypeJIT:
         return posterize_image_pil(inpt, bits=bits)
 
 
-solarize_image_tensor = _FT.solarize
+def solarize_image_tensor(image: torch.Tensor, threshold: float) -> torch.Tensor:
+    if threshold > _FT._max_value(image.dtype):
```
datumbox, line 316: As discussed offline with @pmeier, we will revert the use of `_max_value` here temporarily and hardcode back the 255 value. This change should happen across V1 and V2 and it's a bug fix. But this should happen on a separate PR that changes it from everywhere to avoid issues.

torchvision/prototype/transforms/functional/_color.py
```diff
@@ -456,7 +462,13 @@ def equalize(inpt: features.InputTypeJIT) -> features.InputTypeJIT:
         return equalize_image_pil(inpt)
 
 
-invert_image_tensor = _FT.invert
+def invert_image_tensor(image: torch.Tensor) -> torch.Tensor:
+    if image.dtype == torch.uint8:
+        return image.bitwise_not()
+    else:
+        return _FT._max_value(image.dtype) - image  # type: ignore[no-any-return]
```
datumbox, line 469: Same with this one.

PR #6821, opened 2022-10-24
[prototype] Speed improvement for normalize op

datumbox:
This PR:
- Avoids the `(std == 0).any()` idiom which caused synchronization between CPU and GPU (50% improvement in CUDA)
- Minimizes memory writes by avoiding cloning (25% improvement on CPU)

The combined result is:

```
Float mean/std:
[------------- Normalize cpu torch.float32 -------------]
                     |  normalize stable  |  normalize v2
1 threads: ----------------------------------------------
      (3, 400, 400)  |        364         |      264     
6 threads: ----------------------------------------------
      (3, 400, 400)  |        497         |      351     

[------------- Normalize cuda torch.float32 ------------]
                     |  normalize stable  |  normalize v2
1 threads: ----------------------------------------------
      (3, 400, 400)  |        118         |      55.6    
6 threads: ----------------------------------------------
      (3, 400, 400)  |        118         |      55.6    

List mean/std:
[------------- Normalize cpu torch.float32 -------------]
                     |  normalize stable  |  normalize v2
1 threads: ----------------------------------------------
      (3, 400, 400)  |        378         |      271     
6 threads: ----------------------------------------------
      (3, 400, 400)  |        513         |      360     

[------------- Normalize cuda torch.float32 ------------]
                     |  normalize stable  |  normalize v2
1 threads: ----------------------------------------------
      (3, 400, 400)  |        116         |      61.6    
6 threads: ----------------------------------------------
      (3, 400, 400)  |        116         |      61.6    

```

Modified benchmark script from [here](https://github.com/vfdev-5/tvapiv2_benchmarks/blob/0d081204477a88dd3ee4d887437bf59faf4f0f3c/check_adjust_hue.py)

cc @vfdev-5 @bjuncek @pmeier

### DIFF
--- a/torchvision/prototype/transforms/functional/_misc.py
+++ b/torchvision/prototype/transforms/functional/_misc.py
@@ -20,7 +20,13 @@ def normalize_image_tensor(
             f"Expected tensor to be a tensor image of size (..., C, H, W). Got tensor.size() = {image.size()}"
         )
 
-    if (isinstance(std, (tuple, list)) and not all(std)) or std == 0:
+    if isinstance(std, (tuple, list)):
+        divzero = not all(std)
+    elif isinstance(std, (int, float)):
+        divzero = std == 0
+    else:
+        divzero = False
+    if divzero:
         raise ValueError(f"std evaluated to zero, leading to division by zero.")
 
     dtype = image.dtype
--- a/torchvision/prototype/transforms/functional/_misc.py
+++ b/torchvision/prototype/transforms/functional/_misc.py
@@ -23,21 +23,24 @@ def normalize_image_tensor(
             f"Expected tensor to be a tensor image of size (..., C, H, W). Got tensor.size() = {image.size()}"
         )
 
-    if (isinstance(std, (tuple, list)) and any(std)) or std == 0:
+    if (isinstance(std, (tuple, list)) and not all(std)) or std == 0:
         raise ValueError(f"std evaluated to zero after conversion to {image.dtype}, leading to division by zero.")
 
-    if not inplace:
-        image = image.clone()
-
     dtype = image.dtype
-    mean = torch.as_tensor(mean, dtype=dtype, device=image.device)
-    std = torch.as_tensor(std, dtype=dtype, device=image.device)
+    device = image.device
+    mean = torch.as_tensor(mean, dtype=dtype, device=device)
+    std = torch.as_tensor(std, dtype=dtype, device=device)
     if mean.ndim == 1:
         mean = mean.view(-1, 1, 1)
     if std.ndim == 1:
         std = std.view(-1, 1, 1)
 
-    return image.sub_(mean).div_(std)
+    if inplace:
+        image = image.sub_(mean)
+    else:
+        image = image.sub(mean)
+
+    return image.div_(std)
 
 
 def normalize_video(video: torch.Tensor, mean: List[float], std: List[float], inplace: bool = False) -> torch.Tensor:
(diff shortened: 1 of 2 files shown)


torchvision/prototype/transforms/functional/_misc.py
```diff
@@ -8,7 +8,39 @@
 from torchvision.transforms import functional_tensor as _FT
 from torchvision.transforms.functional import pil_to_tensor, to_pil_image
 
-normalize_image_tensor = _FT.normalize
+
+def normalize_image_tensor(
+    image: torch.Tensor, mean: List[float], std: List[float], inplace: bool = False
+) -> torch.Tensor:
+    if not isinstance(image, torch.Tensor):
+        raise TypeError("Input img should be Tensor image")
```
pmeier, line 16: Do we really need that? Most of our kernels do not have such a check. I mean the kernel is named `*_image_tensor`.

torchvision/prototype/transforms/functional/_misc.py
```diff
@@ -8,7 +8,39 @@
...
+
+    if not image.is_floating_point():
+        raise TypeError(f"Input tensor should be a float tensor. Got {image.dtype}.")
+
+    if image.ndim < 3:
+        raise ValueError(
+            f"Expected tensor to be a tensor image of size (..., C, H, W). Got tensor.size() = {image.size()}"
+        )
+
+    if (isinstance(std, (tuple, list)) and not all(std)) or std == 0:
```
pmeier, line 26: Do we need the first part of the check? What input would fail `isinstance(std, (tuple, list))`? Do we actually allow scalars here? Otherwise, this should be sufficient

```suggestion
    if not all(std):
```

torchvision/prototype/transforms/functional/_misc.py
```diff
@@ -8,7 +8,39 @@
...
+        raise ValueError(f"std evaluated to zero after conversion to {image.dtype}, leading to division by zero.")
```
pmeier, line 27: The message needs at least to be adapted since there is no longer any conversion.

torchvision/prototype/transforms/functional/_misc.py
```diff
@@ -8,7 +8,39 @@
...
+
+    dtype = image.dtype
+    device = image.device
+    mean = torch.as_tensor(mean, dtype=dtype, device=device)
+    std = torch.as_tensor(std, dtype=dtype, device=device)
+    if mean.ndim == 1:
+        mean = mean.view(-1, 1, 1)
+    if std.ndim == 1:
+        std = std.view(-1, 1, 1)
```
pmeier, line 36: I was also looking into this earlier and one thing I asked myself, is when would this branch not trigger? The tensor should always have one dimensions unless we allow scalars. See above for that.

pmeier (approved):
I looked at this earlier and saw no possible optimizations. It seems you have better eyes :stuck_out_tongue: 

LGTM, if CI is green.

torchvision/prototype/transforms/functional/_misc.py
datumbox: We actually allow scalars. It's not visible due to the JIT-script types but if you pass `mean=0.5, std=0.5` it works. So I'm keeping this for BC and provide separate benchmarks.

torchvision/prototype/transforms/functional/_misc.py
datumbox: I think this is purely for broadcasting in case someone passes lists, not scalars. Aka [0.5, 0.5, 0.5]. This is needed else, the following div/sub fails.

torchvision/prototype/transforms/functional/_misc.py
datumbox: Agreed, I'll drop it from here. We'll need to remove it from other places as well. We can do the removal on other places on a separate PR as clean up.

torchvision/prototype/transforms/functional/_misc.py
datumbox: Good catch. I also had to rewrite the check because JIT couldn't understand the assertions were correct in one line... This version seems to pass. I've updated the benchmarks and we are still good.

PR #6824, opened 2022-10-24
test scalars for mean and std in normalize

pmeier:
Per title. Plus, I cleaned up the predefined xfails for JIT a little bit.


datumbox (approved):
LGTM, thanks!

PR #6826, opened 2022-10-24
revert 255 -> max_value fix

### LINKED ISSUE #6825: Don't hardcode 255 unless uint8 is enforced
Across our transformations we sometimes hardcode the value `255`. This is justified if we make sure that point only `torch.uint8` images are allowed, like

https://github.com/pytorch/vision/blob/788ad12e5dcec1dffc7d6311e59a2be2792799b8/torchvision/transforms/functional_tensor.py#L471-L472

However, there are a few instances where uint8 is implied but never enforced:

https://github.com/pytorch/vision/blob/788ad12e5dcec1dffc7d6311e59a2be2792799b8/torchvision/transforms/functional_tensor.py#L266-L267

https://github.com/pytorch/vision/blob/788ad12e5dcec1dffc7d6311e59a2be2792799b8/torchvision/transforms/functional_tensor.py#L778-L779

https://github.com/pytorch/vision/blob/788ad12e5dcec1dffc7d6311e59a2be2792799b8/torchvision/transforms/functional_tensor.py#L852

Instead of hardcoding `255` here, we should either use `_max_value(dtype)` instead or if uint8 is actually required, enforce it.

cc @vfdev-5 @datumbox

pmeier:
See https://github.com/pytorch/vision/pull/6819#discussion_r1003361959 and https://github.com/pytorch/vision/pull/6819#discussion_r1003362097. This will fix will be re-instantiated again with a PR that fixes #6825.

### DIFF
--- a/torchvision/prototype/transforms/functional/_color.py
+++ b/torchvision/prototype/transforms/functional/_color.py
@@ -313,7 +313,8 @@ def posterize(inpt: features.InputTypeJIT, bits: int) -> features.InputTypeJIT:
 
 
 def solarize_image_tensor(image: torch.Tensor, threshold: float) -> torch.Tensor:
-    if threshold > _FT._max_value(image.dtype):
+    bound = 1 if image.is_floating_point() else 255
+    if threshold > bound:
         raise TypeError(f"Threshold should be less or equal the maximum value of the dtype, but got {threshold}")
 
     return torch.where(image >= threshold, invert_image_tensor(image), image)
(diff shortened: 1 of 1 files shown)


datumbox (approved):
LGTM, thanks!

To clarify, the mitigation that @pmeier proposed on the previous PR is correct. We just decided to do this bugfix across all methods and all versions on the same PR. We expect to reland this change soon.

PR #6827, opened 2022-10-24
Add v0.14 release candidates docs and cleanup

YosuaMichael:
- Add v0.14 release candidates docs in versions.html
- Cleanup the docs to use v0.14.0


svekars (approved):
LGTM! But @NicolasHug should take a look too

versions.html
```diff
@@ -20,6 +20,9 @@ <h1> PyTorch Documentation </h1>
         <li class="toctree-l1">
           <a class="reference internal" href="main/">main (unstable)</a>
         </li>
+        <li class="toctree-l1">
+          <a class="reference internal" href="0.14/">v0.14 (release candidate)</a>
```
NicolasHug, line 24: We'll need to remember to update that when the release is out for good

NicolasHug (approved):
Thanks Yosua, LGTM although I would suggest to rename `0.14.0` into simply `0.14`, as we usually don't mention the "bugfix  part" in the docs

versions.html
YosuaMichael: got it, thanks @NicolasHug , will update this

NicolasHug: Sorry I just meant the `.0` part i.e. minor version

YosuaMichael: Ah okay, I have fixed it then

PR #6828, opened 2022-10-24
[prototype] Cleaning up the size dimension methods

datumbox:
This PR does a few last clean ups on the methods that estimate the size of Tensors. The new implementations avoid the repeated assertions of `F` and do things by accessing once the properties.

cc @vfdev-5 @bjuncek @pmeier

### DIFF
--- a/torchvision/prototype/transforms/functional/_meta.py
+++ b/torchvision/prototype/transforms/functional/_meta.py
@@ -16,7 +16,7 @@ def get_dimensions_image_tensor(image: torch.Tensor) -> List[int]:
         chw.insert(0, 1)
         return chw
     else:
-        raise TypeError("Invalid input tensor dimensions.")
+        raise TypeError(f"Input tensor should have at least two dimensions, but got {ndims}")
 
 
 get_dimensions_image_pil = _FP.get_dimensions
@@ -43,7 +43,7 @@ def get_num_channels_image_tensor(image: torch.Tensor) -> int:
     elif ndims == 2:
         return 1
     else:
-        raise TypeError("Invalid input tensor dimensions.")
+        raise TypeError(f"Input tensor should have at least two dimensions, but got {ndims}")
 
 
 get_num_channels_image_pil = _FP.get_image_num_channels
@@ -71,10 +71,11 @@ get_image_num_channels = get_num_channels
 
 def get_spatial_size_image_tensor(image: torch.Tensor) -> List[int]:
     hw = list(image.shape[-2:])
-    if len(hw) == 2:
+    ndims = len(hw)
+    if ndims == 2:
         return hw
     else:
-        raise TypeError("Invalid input tensor dimensions.")
+        raise TypeError(f"Input tensor should have at least two dimensions, but got {ndims}")
 
 
 @torch.jit.unused
--- a/torchvision/prototype/transforms/functional/_meta.py
+++ b/torchvision/prototype/transforms/functional/_meta.py
@@ -7,7 +7,18 @@ from torchvision.prototype.features import BoundingBoxFormat, ColorSpace
 from torchvision.transforms import functional_pil as _FP, functional_tensor as _FT
 
 
-get_dimensions_image_tensor = _FT.get_dimensions
+def get_dimensions_image_tensor(image: torch.Tensor) -> List[int]:
+    chw = list(image.shape[-3:])
+    ndims = len(chw)
+    if ndims == 3:
+        return chw
+    elif ndims == 2:
+        chw.insert(0, 1)
+        return chw
+    else:
+        raise TypeError("Invalid input tensor dimensions.")
+
+
 get_dimensions_image_pil = _FP.get_dimensions
 
 
@@ -24,7 +35,17 @@ def get_dimensions(image: Union[features.ImageTypeJIT, features.VideoTypeJIT]) -
         return get_dimensions_image_pil(image)
 
 
-get_num_channels_image_tensor = _FT.get_image_num_channels
+def get_num_channels_image_tensor(image: torch.Tensor) -> int:
+    chw = image.shape[-3:]
+    ndims = len(chw)
+    if ndims == 3:
+        return chw[0]
+    elif ndims == 2:
+        return 1
+    else:
+        raise TypeError("Invalid input tensor dimensions.")
+
+
 get_num_channels_image_pil = _FP.get_image_num_channels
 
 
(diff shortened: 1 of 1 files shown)


torchvision/prototype/transforms/functional/_meta.py
```diff
@@ -7,7 +7,18 @@
...
+def get_dimensions_image_tensor(image: torch.Tensor) -> List[int]:
+    chw = list(image.shape[-3:])
+    ndims = len(chw)
+    if ndims == 3:
+        return chw
+    elif ndims == 2:
+        chw.insert(0, 1)
+        return chw
+    else:
+        raise TypeError("Invalid input tensor dimensions.")
```
pmeier, line 19: Can we make this a little more precise? Invalid has a pretty broad meaning and we want to only state that the input has to have at least two dimensions.

torchvision/prototype/transforms/functional/_meta.py
```diff
@@ -24,7 +35,17 @@ def get_dimensions(image: Union[features.ImageTypeJIT, features.VideoTypeJIT]) -
         return get_dimensions_image_pil(image)
 
 
-get_num_channels_image_tensor = _FT.get_image_num_channels
+def get_num_channels_image_tensor(image: torch.Tensor) -> int:
```
pmeier, line 38: Not sure if I'm missing something here, but can't we use `get_dimensions_image_tensor` from above and just return the channel?

torchvision/prototype/transforms/functional/_meta.py
```diff
@@ -49,8 +70,11 @@ def get_num_channels(image: Union[features.ImageTypeJIT, features.VideoTypeJIT])
 
 
 def get_spatial_size_image_tensor(image: torch.Tensor) -> List[int]:
-    width, height = _FT.get_image_size(image)
-    return [height, width]
+    hw = list(image.shape[-2:])
```
pmeier, line 73: Same as above. Is there a reason to replicate the logic?

torchvision/prototype/transforms/functional/_meta.py
datumbox: The previous message on V1 was `Tensor is not a torch image.`. I wanted to avoid it because the kernels are reused for images/videos/masks etc. If you have a better proposal, I'm happy to adopt it.

torchvision/prototype/transforms/functional/_meta.py
datumbox: Because we do a billion assertions recursively without good reason. The new implementation cleans up the mess.

torchvision/prototype/transforms/functional/_meta.py
datumbox: Let's keep the discussion in one place above.

torchvision/prototype/transforms/functional/_meta.py
pmeier:
```suggestion
        raise TypeError(f"Input tensor should have at least two dimensions, but got {ndims}")
```

torchvision/prototype/transforms/functional/_meta.py
```diff
@@ -24,7 +35,17 @@ def get_dimensions(image: Union[features.ImageTypeJIT, features.VideoTypeJIT]) -
         return get_dimensions_image_pil(image)
 
 
-get_num_channels_image_tensor = _FT.get_image_num_channels
+def get_num_channels_image_tensor(image: torch.Tensor) -> int:
+    chw = image.shape[-3:]
+    ndims = len(chw)
+    if ndims == 3:
```
atalman, line 41: Can we define numbers as constants ? What if ndims = 4 ?

atalman (commented):
Looks good! could you please answer comments

torchvision/prototype/transforms/functional/_meta.py
```diff
@@ -49,8 +70,12 @@ def get_num_channels(image: Union[features.ImageTypeJIT, features.VideoTypeJIT])
 
 
 def get_spatial_size_image_tensor(image: torch.Tensor) -> List[int]:
-    width, height = _FT.get_image_size(image)
-    return [height, width]
+    hw = list(image.shape[-2:])
+    ndims = len(hw)
+    if ndims == 2:
```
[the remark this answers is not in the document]
pmeier: > Could we maybe define a constant instead ? Also should this be ndims >= 2 ? Just judging from the error message below

Since we do `shape[-2:]` there is no change for this being longer than two elements.

pmeier (approved):
Stamping if CI is green.

torchvision/prototype/transforms/functional/_meta.py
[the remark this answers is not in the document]
datumbox: Thanks @atalman. Because we explicitly ask for the last 2 dims via `image.shape[-2:]`, it can be only up to `2`. It's possible to define a local constant (within the method) but not a global one because of JIT limitations:
```
RuntimeError: 
python value of type 'int' cannot be used as a value. Perhaps it is a closed over global variable? If so, please consider passing it in as an argument or use a local varible instead.:
  File "./torchvision/prototype/transforms/functional/_meta.py", line 74
def get_spatial_size_image_tensor(image: torch.Tensor) -> List[int]:
    hw = list(image.shape[-TWO:])
                           ~~~ <--- HERE
    ndims = len(hw)
    if ndims == TWO:
'get_spatial_size_image_tensor' is being compiled since it was called from 'get_spatial_size'
  File "./torchvision/prototype/transforms/functional/_meta.py", line 103
def get_spatial_size(inpt: features.InputTypeJIT) -> List[int]:
    if isinstance(inpt, torch.Tensor) and (torch.jit.is_scripting() or not isinstance(inpt, features._Feature)):
        return get_spatial_size_image_tensor(inpt)
               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ <--- HERE
    elif isinstance(inpt, (features.Image, features.Video, features.BoundingBox, features.Mask)):
        return list(inpt.spatial_size)
```

Most of our code-base doesn't use constants for shape manipulations. Shall we leave it as-is?

torchvision/prototype/transforms/functional/_meta.py
datumbox: JIT doesn't like constants :( See error at: https://github.com/pytorch/vision/pull/6828#discussion_r1003613897

Issue #6683, opened 2022-10-03
"reflect" padding used in Gaussian Blur Transform throws error on smaller images


themurtazanazir:
### 🐛 Describe the bug

Because of the set `reflect` mode in `torch_pad`, the `GaussianBlur` throws exception for smaller images.

```python
from PIL import Image
import numpy as np
from torchvision import transforms

im  = Image.new("L", (1, 1), 255)
transforms.GaussianBlur(3, sigma=(0.1, 4.0))(im)
```

We should either have a parameter to set the padding mode or 'reflect' should not be used when image size is smaller than kernel size.

Thanks.

### Versions

PyTorch version: 1.12.1+cu102
Is debug build: False
CUDA used to build PyTorch: 10.2
ROCM used to build PyTorch: N/A

OS: Ubuntu 20.04.5 LTS (x86_64)
GCC version: (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0
Clang version: Could not collect
CMake version: Could not collect
Libc version: glibc-2.31

Python version: 3.8.13 (default, Mar 28 2022, 11:38:47)  [GCC 7.5.0] (64-bit runtime)
Python platform: Linux-5.15.0-1020-aws-x86_64-with-glibc2.17
Is CUDA available: True
CUDA runtime version: Could not collect
GPU models and configuration: GPU 0: Tesla T4
Nvidia driver version: 515.65.01
cuDNN version: Could not collect
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

Versions of relevant libraries:
[pip3] numpy==1.23.3
[pip3] torch==1.12.1
[pip3] torch-summary==1.4.5
[pip3] torchvision==0.13.1
[conda] numpy                     1.23.3                   pypi_0    pypi
[conda] torch                     1.12.1                   pypi_0    pypi
[conda] torch-summary             1.4.5                    pypi_0    pypi
[conda] torchvision               0.13.1                   pypi_0    pypi

PR #6829, opened 2022-10-24
pad_mode and pad_value for GaussianBlur

themurtazanazir:
Fix for #6683.

### DIFF
--- a/torchvision/transforms/functional.py
+++ b/torchvision/transforms/functional.py
@@ -1308,7 +1308,13 @@ def erase(img: Tensor, i: int, j: int, h: int, w: int, v: Tensor, inplace: bool 
     return F_t.erase(img, i, j, h, w, v, inplace=inplace)
 
 
-def gaussian_blur(img: Tensor, kernel_size: List[int], sigma: Optional[List[float]] = None) -> Tensor:
+def gaussian_blur(
+    img: Tensor,
+    kernel_size: List[int],
+    sigma: Optional[List[float]] = None,
+    pad_mode: str = "reflect",
+    pad_value: float = 0.0,
+) -> Tensor:
     """Performs Gaussian blurring on the image by given kernel.
     If the image is torch Tensor, it is expected
     to have [..., H, W] shape, where ... means an arbitrary number of leading dimensions.
@@ -1330,6 +1336,8 @@ def gaussian_blur(img: Tensor, kernel_size: List[int], sigma: Optional[List[floa
             .. note::
                 In torchscript mode sigma as single float is
                 not supported, use a sequence of length 1: ``[sigma, ]``.
+        pad_mode (str): padding mode to be passed to the padding function. Default: "reflect".
+        pad_value (float): value to be passed to the padding function. Default: 0.0.
 
     Returns:
         PIL Image or Tensor: Gaussian Blurred version of the image.
@@ -1368,7 +1376,7 @@ def gaussian_blur(img: Tensor, kernel_size: List[int], sigma: Optional[List[floa
 
         t_img = pil_to_tensor(img)
 
-    output = F_t.gaussian_blur(t_img, kernel_size, sigma)
+    output = F_t.gaussian_blur(t_img, kernel_size, sigma, pad_mode, pad_value)
 
     if not isinstance(img, torch.Tensor):
         output = to_pil_image(output, mode=img.mode)
--- a/torchvision/transforms/functional_tensor.py
+++ b/torchvision/transforms/functional_tensor.py
@@ -745,7 +745,9 @@ def _get_gaussian_kernel2d(
     return kernel2d
 
 
-def gaussian_blur(img: Tensor, kernel_size: List[int], sigma: List[float]) -> Tensor:
+def gaussian_blur(
+    img: Tensor, kernel_size: List[int], sigma: List[float], pad_mode: str = "reflect", pad_value: float = 0.0
+) -> Tensor:
     if not (isinstance(img, torch.Tensor)):
         raise TypeError(f"img should be Tensor. Got {type(img)}")
 
@@ -759,7 +761,7 @@ def gaussian_blur(img: Tensor, kernel_size: List[int], sigma: List[float]) -> Te
 
     # padding = (left, right, top, bottom)
     padding = [kernel_size[0] // 2, kernel_size[0] // 2, kernel_size[1] // 2, kernel_size[1] // 2]
-    img = torch_pad(img, padding, mode="reflect")
+    img = torch_pad(img, padding, mode=pad_mode, value=pad_value)
     img = conv2d(img, kernel, groups=img.shape[-3])
 
     img = _cast_squeeze_out(img, need_cast, need_squeeze, out_dtype)
--- a/torchvision/transforms/transforms.py
+++ b/torchvision/transforms/transforms.py
@@ -1754,7 +1754,7 @@ class GaussianBlur(torch.nn.Module):
 
     """
 
-    def __init__(self, kernel_size, sigma=(0.1, 2.0)):
+    def __init__(self, kernel_size, sigma=(0.1, 2.0), pad_mode="reflect", pad_value=0.0):
         super().__init__()
         _log_api_usage_once(self)
         self.kernel_size = _setup_size(kernel_size, "Kernel size should be a tuple/list of two integers")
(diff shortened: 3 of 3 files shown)


datumbox:
@themurtazanazir Thanks for the contribution! We are currently revamping the Transforms API and try to reduce the number of new features/params introduced. Let me get back to you concerning whether we can merge right now or wait for after the API of V2 is finalized.

Pinging @pmeier and @vfdev-5 to get their thoughts.

themurtazanazir:
@datumbox Sure. Is there any tracking on this revamping, the timeline or the changes? I'd like to be updated. Thanks.

datumbox:
@themurtazanazir We track the work on the project on this [dashboard](https://github.com/pytorch/vision/projects/5). Moreover we are about to publish a blogpost about the new API with all the details and examples. Some high-level examples can be seen on #6753.

themurtazanazir:
@datumbox Thanks.

datumbox (commented):
@themurtazanazir thanks for the updates.

This leaves the V2 of the API disconnected from the implementation on V1. We would need to add the same parameters on the V2 (code pointers [here](https://github.com/pytorch/vision/blob/main/torchvision/prototype/transforms/functional/_misc.py#L85) and [here](https://github.com/pytorch/vision/blob/main/torchvision/prototype/transforms/_misc.py#L115)).

@vfdev-5 Any concerns/feedback about the nature of this extension?

vfdev-5:
@themurtazanazir thanks for working on this PR. I was trying to understand better the context of the linked issue and I'm not sure that we need to update the API for cases when input image is smaller than gaussian blur kernel size. Can we just fix gaussian blur implementation for such cases by explicitly using constant mode if it makes sense ?

themurtazanazir:
@vfdev-5 yes, just to fix this issue, we can dynamically choose what pad_mode to use. The pro is that there is no change in API parameters. However, this new approach adds the flexibility of using even other types of pad_modes, i.e "replicate", "circular" and this is backward compatible (throw errors where it was throwing errors earlier). Furthermore, it would be complicated to explain the dynamic padding modes in in docs.

vfdev-5:
@themurtazanazir I think such flexibility can be achived in a composed way instead of adding new args to existing transformation. We can explicitly pad the image with `Pad` before calling `GaussianBlur`. On the other hand the use-cases when image is smaller than kernel size seems rather rare IMO.

> Furthermore, it would be complicated to explain the dynamic padding modes in in docs.

In this case we can just raise an error with more comprehensible message and suggest how to overcome it.

themurtazanazir:
> ... such flexibility can be achived in a composed way instead of adding new args to existing transformation

I don't think it would be _exactly_ the same. `GaussianBlur` will pad in "reflect" (or "constant", if we go dynamic!) mode regardless.

> On the other hand the use-cases when image is smaller than kernel size seems rather rare IMO.

I cannot speak for that. But for the context, this error will occur even when just one dimension of the image is smaller than the kernel.

e.g:
```python
from PIL import Image
import numpy as np
from torchvision import transforms 
im  = Image.new("L", (78, 3), 255)
transforms.GaussianBlur(7, sigma=(0.1, 4.0))(im)
```

This will also throw an error. 

A common usecase would be generating synthetic images for OCR related tasks.
  
> In this case we can just raise an error with more comprehensible message and suggest how to overcome it.

Here the suggestion would be to use `Pad` before `GaussianBlur`?

vfdev-5:
> A common usecase would be generating synthetic images for OCR related tasks.

So, in OCR task you can apply a blur larger than an image dimension ?

The main question here is what would be a right way to do gaussian blur in those cases. What do you think ?
In general case I'd assume that padding the input image with reflect mode to compensate boundary effects is acceptable for gaussian blur. In case of kernel size larger than one input image dimension, I think it is fair to let user pad this dimension before gaussian blur:
```python
from PIL import Image
import numpy as np
from torchvision import transforms 

im  = Image.new("L", (78, 3), 255)

out = transforms.Compose([
    transforms.Pad([0, 2], fill=127),
    transforms.GaussianBlur(7, sigma=(0.1, 4.0))
])(im)
```

themurtazanazir:
> So, in OCR task you can apply a blur larger than an image dimension ?

Yes it can be. Imagine a synthetic image generation pipeline using a random font and a random font size (from a range). It is possible an image height can be smaller than kernel size. Although not so common, but defintely likely.

> The main question here is what would be a right way to do gaussian blur in those cases. What do you think ?

To answer this, I'm going to go on a bit of an over-explanation, so bear with me.

As far as I understand, the `GaussianBlur` is a wrapper of `funtional_tensor.gaussian_blur` (hereafter just refernced as `gaussian_blur` ) with a random `sigma` chosen. The `gaussian_blur` accepts multiple padding types because it pads the image to **keep the image size same**. 

Now if we go on to have user pad the image first using `Pad` and use `GaussianBlur` on top, in effect it will change the image size.

e.g, in the code:

```python
from PIL import Image
import numpy as np
from torchvision import transforms 

im  = Image.new("L", (78, 3), 255)

out = transforms.Compose([
    transforms.Pad([0, 2], fill=127),
    transforms.GaussianBlur(7, sigma=(0.1, 4.0))
])(im)

print(im.size, out.size)
```
The output is `(78, 3) (78, 7)` and keeping the size same might be important for usecases, (which is why we must have added padding in the `gaussuan_blur` itself).

Secondly in this solution, the user again has to dynamically choose the padding given the difference in image size and kernel size.

e.g:

```python
from PIL import Image
import numpy as np
from torchvision import transforms 

im  = Image.new("L", (78, 3), 255)

out = transforms.Compose([
    transforms.Pad([0, 2], fill=127),
    transforms.GaussianBlur(21, sigma=(0.1, 4.0))
])(im)
```
will again throw the error.  Here the user needs to dynamically calculate the padding.

Third is, why are we restricting the function paramters of `gaussian_blur` in `GaussianBlur`? I mean "reflect" padding is considered "better" but that should be kept as the default parameter for `GaussianBlur` but not the only option, IMO.

For the first two problems, we can either let the user decide by adding the parameters `pad_mode` and `pad_value` (as is done in this PR) or the way you suggested to dynamically check if the any of the image dimension is smaller, then switch to any other mode than "reflect".

If we consider all three issues, then giving user the choice through parameters seems the only way at the moment.

Sorry for the long note :innocent: .

vfdev-5:
@themurtazanazir thanks for exposing your thoughts, I'm still not convinced about new args.

I checked opencv docs and they expose `borderType` arg for an equivalent purpose: https://docs.opencv.org/4.6.0/d4/d86/group__imgproc__filter.html#gaabe8c836e97159a9193fb0b11ac52cf1
However, their replicate mode seems like to use max possible kernel size instead of provided by user:
```python
import numpy as np
import cv2
i = np.arange(4 * 4, dtype="uint8").reshape(4, 4)

cv2.GaussianBlur(i, ksize=(7, 0), sigmaX=2.0, borderType=cv2.BORDER_REFLECT)

# array([[ 5,  6,  6,  6],
       [ 6,  7,  7,  7],
       [ 8,  8,  8,  9],
       [ 9,  9,  9, 10]], dtype=uint8)

cv2.GaussianBlur(i, ksize=(9, 0), sigmaX=2.0, borderType=cv2.BORDER_REFLECT)
# array([[ 5,  6,  6,  6],
       [ 6,  7,  7,  7],
       [ 8,  8,  8,  9],
       [ 9,  9,  9, 10]], dtype=uint8)

cv2.GaussianBlur(i, ksize=(11, 0), sigmaX=2.0, borderType=cv2.BORDER_REFLECT)
# array([[ 5,  6,  6,  6],
       [ 6,  7,  7,  7],
       [ 8,  8,  8,  9],
       [ 9,  9,  9, 10]], dtype=uint8)
```
Maybe, it worth to check their code to see what they exactly doing...

On the other hand, PIL GaussianBlur does not have such arg: https://pillow.readthedocs.io/en/stable/reference/ImageFilter.html#PIL.ImageFilter.GaussianBlur

So, we have the several options without introducing new args:

1) do similar thing as opencv (maybe clip kernel size to max possible value)
2) raise an error if kernel size > image size and suggest to pad and crop
```
from PIL import Image
import numpy as np
from torchvision import transforms 

im  = Image.new("L", (78, 3), 255)

out = transforms.Compose([
    transforms.Pad([0, 2], fill=127),
    transforms.GaussianBlur(7, sigma=(0.1, 4.0)),
    transforms.Pad([0, -2], fill=127),
])(im)

print(im.size, out.size)
```
3) replace padding mode with constant if kernel size > image size

PR #6830, opened 2022-10-24
Fix hardcoded 255

### LINKED ISSUE #6825: Don't hardcode 255 unless uint8 is enforced

pmeier:
Fixes #6825.


torchvision/prototype/transforms/functional/_color.py
```diff
@@ -226,19 +226,15 @@ def adjust_hue_image_tensor(image: torch.Tensor, hue_factor: float) -> torch.Ten
         return image
 
     orig_dtype = image.dtype
-    if image.dtype == torch.uint8:
-        image = image / 255.0
```
pmeier, line 230: Instead of doing the conversion manually, I've opted to use our kernel for this. Note that this also implicitly converts to `float32` since the divisor is a `float`.

torchvision/transforms/functional_tensor.py
```diff
@@ -15,12 +15,6 @@ def _assert_image_tensor(img: Tensor) -> None:
         raise TypeError("Tensor is not a torch image.")
 
 
-def _assert_threshold(img: Tensor, threshold: float) -> None:
```
pmeier, line 18: This was only used once so I inlined it.

test/test_functional_tensor.py
```diff
@@ -808,14 +808,14 @@ def test_solarize_threshold1_upper_bound(threshold, device):
...
-    img = torch.randint(0, 256, (3, 12, 23)).to(device)
+    img = torch.randint(0, 256, (3, 12, 23), dtype=torch.uint8, device=device)
     F_t.solarize(img, threshold)
 
 
 @pytest.mark.parametrize("device", cpu_and_gpu())
 @pytest.mark.parametrize("threshold", [260])
 def test_solarize_threshold2_upper_bound(threshold, device):
-    img = torch.randint(0, 256, (3, 12, 23)).to(device)
+    img = torch.randint(0, 256, (3, 12, 23), dtype=torch.uint8, device=device)
```
pmeier, line 818: `torch.randint` is a `int64` by default which will no longer trigger an error for a threshold of `260`.

pmeier: Instead of just fixing this here, I opted to make the tests more robust in e13613ab3e891ff0af0338e43e0a27d370ca4159.

datumbox:
@pmeier Thanks for the PR. Have you made any measurements on the V2 changes to check if there is a performance degradation? There are a few more 255 hardcoded values left but those are in places where we support only uint8. Is there a plan to update them?

Finally this change, though correct, has the potential of breaking existing code. Before merging we might need to cherrypick it and bring it on FBcode to see if there is any breakage.

pmeier:
> Have you made any measurements on the V2 changes to check if there is a performance degradation?

Not yet. For all ops except for `adjust_hue` we only replace one if-else with a function call that includes an if-elif-else. This change should be in the nano second range and thus well within our measuring tolerance. Do you also want me to benchmark all of them?

> There are a few more 255 hardcoded values left but those are in places where we support only uint8. Is there a plan to update them?

Nope, I don't see a point. Or do you mean have another look at the kernel whether or not we are doing the wrong thing, i.e. nothing, for other integer dtypes? In that case, yes that would be a good idea.

pmeier:
There are two places with hardcoded `255` left:

1. https://github.com/pytorch/vision/blob/7de68b0d316d98a16aec9ede901afea3f3e1b542/torchvision/transforms/functional_tensor.py#L471-L472

    Two things here:
    
    1. Although the `255` is fine here for `torch.uint8`, all other integer images will not be clamped and thus might fail the following conversion back to an integer dtype due to overflow. Thus, this should be something like 
    
      ```py
      if interpolation == "bicubic" and not out_dtype.is_floating_point:
          img = img.clamp(min=0, max=_max_value(out_dtype)) 
      ```
  
    2. That being said, I'm a little confused why we are only clamping only for integer dtypes in the first place. Bicubic interpolation can lead to overflowing values, so shouldn't we clamp regardless of the `dtype`? Otherwise, the value range `[0.0, 1.0]` is no longer guaranteed after this operation. Thus, I think this should be 
    
      ```py
      if interpolation == "bicubic":
          img = img.clamp_(0, _max_value(out_dtype))
      ```

    Maybe @vfdev-5 can shed some light here.

2. https://github.com/pytorch/vision/blob/7de68b0d316d98a16aec9ede901afea3f3e1b542/torchvision/transforms/functional_tensor.py#L870-L873
 
   which is guarded by 

   https://github.com/pytorch/vision/blob/7de68b0d316d98a16aec9ede901afea3f3e1b542/torchvision/transforms/functional_tensor.py#L896-L897

   Meaning, `equalize` does not work with any dtype other than `torch.uint8` and thus the hardcoded `255` is fine. 

   That being said, I think we need to have discussion whether or not we want to have kernels that categorically only work with a subset of dtypes, in the extreme case like here, only for a single dtype. That means for example that the AA transforms that use `equalize` internally cannot be used with floating point images. Their docstring states that much

   https://github.com/pytorch/vision/blob/7de68b0d316d98a16aec9ede901afea3f3e1b542/torchvision/transforms/autoaugment.py#L104-L107

   but we should discuss whether or not we are ok with this behavior going forward. This should not happen on this PR and I will open an issue soon about this.

datumbox:
@pmeier As discussed offline, checking only the methods that change significantly will do. No need to benchmark those that just fetch the value from the dictionary. Let's wait for Victor's thoughts on this.

@fmassa I was wondering if you could chime in as well. I'm supportive of Philip's change, just wanted to make sure we don't miss something important. The TLDR is, we replace the hardcoded 255 values with the ones for each dtype. Floats and uint8 remain unaffected but other integer types would change. This would align the behaviour of the kernels with `convert_image_dtype` which currently use different max values. I think this can be considered a bug but not 100% if it was originally intentional.

pmeier:
```
[--------------------------- adjust_hue ---------------------------]
                                       |  main  |  fix-hardcoded-255
1 threads: ---------------------------------------------------------
      (3, 512, 512), uint8, cpu        |   14   |          15       
      (3, 512, 512), uint8, cuda       |    1   |           1       
      (3, 512, 512), float32, cpu      |   14   |          14       
      (3, 512, 512), float32, cuda     |    1   |           1       
      (5, 3, 512, 512), uint8, cpu     |   94   |          90       
      (5, 3, 512, 512), uint8, cuda    |    7   |           8       
      (5, 3, 512, 512), float32, cpu   |   88   |          84       
      (5, 3, 512, 512), float32, cuda  |    7   |           7       

```

No changes that cannot be attributed to noise.

datumbox:
I've imported this PR on FBcode to check if there are any breakages at [D40752944](https://www.internalfb.com/diff/D40752944)

datumbox:
I ran all the tests internally and it seems the change didn't break anything. There are a lot of pre-existing failures and skipped tests, so we can't be 100% sure. But it looks like it's mostly OK.

@pmeier Do we need to update the PR to cover your recently changes on the 2 kernels?

torchvision/prototype/transforms/functional/_color.py
```diff
@@ -469,7 +469,7 @@ def equalize(inpt: features.InputTypeJIT) -> features.InputTypeJIT:
 
 def invert_image_tensor(image: torch.Tensor) -> torch.Tensor:
     if image.is_floating_point():
-        return _FT._max_value(image.dtype) - image  # type: ignore[no-any-return]
+        return 1.0 - image  # type: ignore[no-any-return]
```
datumbox, line 473: This is correct but if for some reason on the future we decide to have different treatment around `max_value` for floats, it would have to be replaced. No action needed but I would be OK with the extra unnecessary method call as well.

pmeier: Apart from here, this is at least hardcoded for `convert_dtype` for the float to int 

https://github.com/pytorch/vision/blob/436ff9a52a0c901aa1b9bb51c0e580ae7f7cd4c0/torchvision/prototype/transforms/functional/_meta.py#L355-L359

and int to float cases

https://github.com/pytorch/vision/blob/436ff9a52a0c901aa1b9bb51c0e580ae7f7cd4c0/torchvision/prototype/transforms/functional/_meta.py#L363

If we don't hardcode `1.0` here, we should also do the same there.

datumbox: Agreed, good spot. It's up to you how you will handle it TBH. Another way would be to just add a comment and say we hardcode to avoid the call. No strong opinions on my side.

fmassa:
Hi @datumbox 

I'm supportive of this change. We didn't have good support for other dtypes before, so assuming either `uint8` or `float` was okaish. Happy to see this being improved!

datumbox:
@pmeier Looks like we should be good to go once you finish the remaining hardcoded values. Ping me when you are happy to do one final review and merge. I've already ported this internally and it looks there are no issues.

torchvision/prototype/transforms/functional/_geometry.py
```diff
@@ -143,8 +143,9 @@ def resize_image_tensor(
         )
 
         if need_cast:
-            if interpolation == InterpolationMode.BICUBIC and dtype == torch.uint8:
-                image = image.clamp_(min=0, max=255)
+            # bicubic interpolation can overshoot
+            if interpolation == InterpolationMode.BICUBIC:
+                image = image.clamp_(min=0, max=_FT._max_value(dtype))
```
datumbox, line 148: I will need to check this internally before we merge. Shall we revert the commit and bring it separately form this PR to merge this faster?

torchvision/transforms/functional_tensor.py
```diff
@@ -458,8 +458,8 @@ def resize(
 
     img = interpolate(img, size=size, mode=interpolation, align_corners=align_corners, antialias=antialias)
 
-    if interpolation == "bicubic" and out_dtype == torch.uint8:
-        img = img.clamp(min=0, max=255)
+    if interpolation == "bicubic" and out_dtype not in (torch.float32, torch.float64):
+        img = img.clamp(min=0, max=_max_value(out_dtype))
```
datumbox, line 462: Same here, revert change and bring on a new PR?

pmeier:
There was some offline discussion whether or not we want to remove the implicit assumption that floating point images have the maximum value `1.0`. Here are some benchmarks:

```
[-------------- convert_dtype_image_tensor float32 -> float64 ---------------]
                                      |        main       |  fix-hardcoded-255
1 threads: -------------------------------------------------------------------
      (3, 512, 512), float32, cpu     |    81 (+-  0) us  |    151 (+-  1) us 
      (5, 3, 512, 512), float32, cpu  |  1406 (+- 58) us  |   2183 (+-143) us 

```

- float to float will be slower because we need to perform an additional (inplace) multiplication whereas before a dtype conversion was sufficient. However, if we assume that each float has the same value range, we don't need to touch this conversion at all.

```
[--------------- convert_dtype_image_tensor float32 -> uint8 ----------------]
                                      |        main       |  fix-hardcoded-255  
1 threads: -------------------------------------------------------------------
      (3, 512, 512), float32, cpu     |   414 (+- 37) us  |    408 (+-  3) us   
      (5, 3, 512, 512), float32, cpu  |  2326 (+-182) us  |   2208 (+- 34) us   

```

- If anything, the new version should be slower since we need an additional (Python scalar) division. Not sure where the measured difference comes from.

```
[-------------- convert_dtype_image_tensor uint8 -> float32 --------------]
                                    |       main       |  fix-hardcoded-255
1 threads: ----------------------------------------------------------------
      (3, 512, 512), uint8, cpu     |  133 (+-  1) us  |     95 (+-  0) us 
      (5, 3, 512, 512), uint8, cpu  |  640 (+-  3) us  |    438 (+-  6) us 

```

- If anything, the new version should be slower since we need an additional (Python scalar) division. The performance improvement here comes from a trick that I found while implementing this patch that is independent of this change. Instead of doing a tensor division, we can simply do a Python scalar division followed by a tensor multiplication. Effectively, this turns 

  https://github.com/pytorch/vision/blob/1502ed9ae21f35bcd5661cd3bb67fd7bc6b2b863/torchvision/prototype/transforms/functional/_meta.py#L370

  into

  ```py
  image.to(dtype).mul_(1 / _FT._max_value(image.dtype)) 
  ```

  Although it seems unrelated here, I've included this improvement in the benchmark, because the only thing that changes is the `1` turns into a `_FT._max_value(dtype)`.

torchvision/prototype/transforms/functional/_color.py
```diff
@@ -222,19 +222,15 @@ def adjust_hue_image_tensor(image: torch.Tensor, hue_factor: float) -> torch.Ten
...
     h, s, v = image.unbind(dim=-3)
     h.add_(hue_factor).remainder_(1.0)
     image = torch.stack((h, s, v), dim=-3)
     image_hue_adj = _hsv_to_rgb(image)
 
-    if orig_dtype == torch.uint8:
-        image_hue_adj = image_hue_adj.mul_(255.0).to(dtype=orig_dtype)
-
-    return image_hue_adj
+    return convert_dtype_image_tensor(image_hue_adj, orig_dtype)
```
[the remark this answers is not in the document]
datumbox: Thanks. Is that V2 main vs V2 PR?

torchvision/prototype/transforms/functional/_color.py
```diff
@@ -461,10 +461,13 @@ def equalize(inpt: features.InputTypeJIT) -> features.InputTypeJIT:
...
-    if image.dtype == torch.uint8:
+    if image.is_floating_point():
+        return 1.0 - image  # type: ignore[no-any-return]
+    elif image.dtype == torch.uint8:
         return image.bitwise_not()
-    else:
-        return (1 if image.is_floating_point() else 255) - image  # type: ignore[no-any-return]
+    else:  # signed integer dtypes
+        # We can't use `Tensor.bitwise_not` here, since we want to retain the leading zero bit that encodes the sign
+        return image.bitwise_xor((1 << _num_value_bits(image.dtype)) - 1)
```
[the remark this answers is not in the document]
pmeier:
```
[--------------------------- invert_image_tensor ---------------------------]
                                      |       main       |  fix-hardcoded-255
1 threads: ------------------------------------------------------------------
      (3, 512, 512), float32, cpu     |   61 (+-  0) us  |     57 (+-  0) us 
      (3, 512, 512), uint8, cpu       |   17 (+-  0) us  |     17 (+-  0) us 
      (3, 512, 512), int32, cpu       |   78 (+-  0) us  |     63 (+-  0) us 
      (5, 3, 512, 512), float32, cpu  |  461 (+- 33) us  |    445 (+- 31) us 
      (5, 3, 512, 512), uint8, cpu    |   98 (+-  1) us  |     79 (+-  1) us 
      (5, 3, 512, 512), int32, cpu    |  538 (+- 67) us  |    514 (+-  8) us 

```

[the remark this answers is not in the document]
datumbox: Nice! I like it when bug/code-quality fixing leads to speed improvements. What more can we ask? 😄

datumbox:
> There was some offline discussion whether or not we want to remove the implicit assumption that floating point images have the maximum value 1.0. 

Let's not adopt anything that slows us down.

> Instead of doing a tensor division, we can simply do a Python scalar division followed by a tensor multiplication.

What the heck! Well sounds good to me. Shall we try the trick in other places too?

Here are some places that we could replace it:
https://github.com/pytorch/vision/blob/aec38fc2f97086a843ec79cb376eb3f350d80a88/torchvision/prototype/transforms/functional/_color.py#L183
https://github.com/pytorch/vision/blob/aec38fc2f97086a843ec79cb376eb3f350d80a88/torchvision/prototype/transforms/functional/_color.py#L297
https://github.com/pytorch/vision/blob/aec38fc2f97086a843ec79cb376eb3f350d80a88/torchvision/prototype/transforms/functional/_meta.py#L370

Finally from what I understand, none of the changes you make here are expected to cause speed regressions, can you confirm?

PR #6832, opened 2022-10-25
[prototype] Minor improvements on functional

datumbox:
Minor last improvements on the functional kernels. I consider these mostly code-quality improvements rather that speed optimizations.

cc @vfdev-5 @bjuncek @pmeier

### DIFF
--- a/torchvision/prototype/transforms/functional/_color.py
+++ b/torchvision/prototype/transforms/functional/_color.py
@@ -188,7 +188,7 @@ def _hsv_to_rgb(img: torch.Tensor) -> torch.Tensor:
     h, s, v = img.unbind(dim=-3)
     h6 = h * 6
     i = torch.floor(h6)
-    f = (h6) - i
+    f = h6 - i
     i = i.to(dtype=torch.int32)
 
     p = (v * (1.0 - s)).clamp_(0.0, 1.0)
@@ -209,9 +209,6 @@ def _hsv_to_rgb(img: torch.Tensor) -> torch.Tensor:
 def adjust_hue_image_tensor(image: torch.Tensor, hue_factor: float) -> torch.Tensor:
     if not (-0.5 <= hue_factor <= 0.5):
         raise ValueError(f"hue_factor ({hue_factor}) is not in [-0.5, 0.5].")
-
-    if not (isinstance(image, torch.Tensor)):
-        raise TypeError("Input img should be Tensor image")
 
     c = get_num_channels_image_tensor(image)
 
@@ -258,9 +255,6 @@ def adjust_hue(inpt: features.InputTypeJIT, hue_factor: float) -> features.Input
 
 
 def adjust_gamma_image_tensor(image: torch.Tensor, gamma: float, gain: float = 1.0) -> torch.Tensor:
-    if not (isinstance(image, torch.Tensor)):
-        raise TypeError("Input img should be Tensor image")
-
     if gamma < 0:
         raise ValueError("Gamma should be a non-negative real number")
 
@@ -337,10 +331,6 @@ def solarize(inpt: features.InputTypeJIT, threshold: float) -> features.InputTyp
 
 
 def autocontrast_image_tensor(image: torch.Tensor) -> torch.Tensor:
-
-    if not (isinstance(image, torch.Tensor)):
-        raise TypeError("Input img should be Tensor image")
-
     c = get_num_channels_image_tensor(image)
 
     if c not in [1, 3]:
--- a/torchvision/prototype/transforms/functional/_meta.py
+++ b/torchvision/prototype/transforms/functional/_meta.py
@@ -183,12 +183,8 @@ def clamp_bounding_box(
     return convert_format_bounding_box(xyxy_boxes, BoundingBoxFormat.XYXY, format)
 
 
-def _split_alpha(image: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
-    return torch.tensor_split(image, indices=(-1,), dim=-3)
-
-
 def _strip_alpha(image: torch.Tensor) -> torch.Tensor:
-    image, alpha = _split_alpha(image)
+    image, alpha = torch.tensor_split(image, indices=(-1,), dim=-3)
     if not torch.all(alpha == _FT._max_value(alpha.dtype)):
         raise RuntimeError(
             "Stripping the alpha channel if it contains values other than the max value is not supported."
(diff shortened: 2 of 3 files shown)


torchvision/prototype/transforms/functional/_color.py
```diff
@@ -210,9 +210,6 @@ def adjust_hue_image_tensor(image: torch.Tensor, hue_factor: float) -> torch.Ten
     if not (-0.5 <= hue_factor <= 0.5):
         raise ValueError(f"hue_factor ({hue_factor}) is not in [-0.5, 0.5].")
 
-    if not (isinstance(image, torch.Tensor)):
-        raise TypeError("Input img should be Tensor image")
```
datumbox, line 214: As decided previously, removing the final tensor checks on tensor kernels.

torchvision/prototype/transforms/functional/_meta.py
```diff
@@ -183,12 +183,8 @@ def clamp_bounding_box(
     return convert_format_bounding_box(xyxy_boxes, BoundingBoxFormat.XYXY, format)
 
 
-def _split_alpha(image: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
-    return image[..., :-1, :, :], image[..., -1:, :, :]
```
datumbox, line 187: `tensor_split` is marginally faster because it avoids the double indexing:
```
[------------------------------- Convert_color_space_image_tensor cpu torch.uint8 ------------------------------]
                     |  convert_color_space_image_tensor v2 before  |  convert_color_space_image_tensor_ v2 after
1 threads: ------------------------------------------------------------------------------------------------------
      (4, 400, 400)  |                     217                      |                     213                    
6 threads: ------------------------------------------------------------------------------------------------------
      (4, 400, 400)  |                     297                      |                     292                    

[------------------------------ Convert_color_space_image_tensor cuda torch.uint8 ------------------------------]
                     |  convert_color_space_image_tensor v2 before  |  convert_color_space_image_tensor_ v2 after
1 threads: ------------------------------------------------------------------------------------------------------
      (4, 400, 400)  |                     56.4                     |                      52                    
6 threads: ------------------------------------------------------------------------------------------------------
      (4, 400, 400)  |                     56.4                     |                      52                    

[------------------------------ Convert_color_space_image_tensor cpu torch.float32 -----------------------------]
                     |  convert_color_space_image_tensor v2 before  |  convert_color_space_image_tensor_ v2 after
1 threads: ------------------------------------------------------------------------------------------------------
      (4, 400, 400)  |                     236                      |                     232                    
6 threads: ------------------------------------------------------------------------------------------------------
      (4, 400, 400)  |                     325                      |                     321                    

[----------------------------- Convert_color_space_image_tensor cuda torch.float32 -----------------------------]
                     |  convert_color_space_image_tensor v2 before  |  convert_color_space_image_tensor_ v2 after
1 threads: ------------------------------------------------------------------------------------------------------
      (4, 400, 400)  |                     56.9                     |                     52.4                   
6 threads: ------------------------------------------------------------------------------------------------------
      (4, 400, 400)  |                     56.7                     |                     52.4                   

```

torchvision/prototype/transforms/functional/_meta.py
```diff
@@ -183,12 +183,8 @@ def clamp_bounding_box(
...
-
-
 def _strip_alpha(image: torch.Tensor) -> torch.Tensor:
-    image, alpha = _split_alpha(image)
+    image, alpha = torch.tensor_split(image, indices=(-1,), dim=-3)
```
datumbox, line 187: IMO we don't need a private one liner method for something Torch has a built in method.

torchvision/prototype/transforms/functional/_misc.py
```diff
@@ -69,18 +69,18 @@ def normalize(
     return normalize_image_tensor(inpt, mean=mean, std=std, inplace=inplace)
 
 
-def _get_gaussian_kernel1d(kernel_size: int, sigma: float) -> torch.Tensor:
+def _get_gaussian_kernel1d(kernel_size: int, sigma: float, dtype: torch.dtype, device: torch.device) -> torch.Tensor:
     lim = (kernel_size - 1) / (2 * math.sqrt(2) * sigma)
-    x = torch.linspace(-lim, lim, steps=kernel_size)
+    x = torch.linspace(-lim, lim, steps=kernel_size, dtype=dtype, device=device)
```
datumbox, line 74: Following the approach of 2d, we should create directly the linspace on device/type rather than copying. Minor improvement for GPU:
```
[--------------------- Gaussian_blur cpu torch.uint8 ---------------------]
                     |  gaussian_blur v2 before  |  gaussian_blur_ v2 after
1 threads: ----------------------------------------------------------------
      (3, 400, 400)  |            4.0           |             3.97         
6 threads: ----------------------------------------------------------------
      (3, 400, 400)  |            5.3           |             5.28         

[--------------------- Gaussian_blur cuda torch.uint8 --------------------]
                     |  gaussian_blur v2 before  |  gaussian_blur_ v2 after
1 threads: ----------------------------------------------------------------
      (3, 400, 400)  |            165            |            153          
6 threads: ----------------------------------------------------------------
      (3, 400, 400)  |            165            |            154          

[-------------------- Gaussian_blur cpu torch.float32 --------------------]
                     |  gaussian_blur v2 before  |  gaussian_blur_ v2 after
1 threads: ----------------------------------------------------------------
      (3, 400, 400)  |            3.12           |            3.12         
6 threads: ----------------------------------------------------------------
      (3, 400, 400)  |            4.39           |            4.38         

[-------------------- Gaussian_blur cuda torch.float32 -------------------]
                     |  gaussian_blur v2 before  |  gaussian_blur_ v2 after
1 threads: ----------------------------------------------------------------
      (3, 400, 400)  |            134            |            122          
6 threads: ----------------------------------------------------------------
      (3, 400, 400)  |            135            |            122          

```

pmeier: Can `dtype` be an integer one here? Recall, that there were some differences in https://github.com/pytorch/vision/pull/6812#discussion_r1002089826.

torchvision/prototype/transforms/functional/_meta.py
pmeier: Non-blocking: I would prefer it since it is a little more expressive. Plus, since we have a `_add_alpha` function it would be consistent to have the complementary `_split_alpha` as well. With this change we have three identical, non-trivial calls of `torch.tensor_split`. At a glance, it is harder to see that these three actually do exactly the same.

torchvision/prototype/transforms/functional/_misc.py
datumbox: No it can only be float types. There is a conversion to float before the kernel is called:
https://github.com/pytorch/vision/blob/788ad12e5dcec1dffc7d6311e59a2be2792799b8/torchvision/prototype/transforms/functional/_misc.py#L126-L128

pmeier (approved):
LGTM, thanks Vasilis.

torchvision/prototype/transforms/functional/_meta.py
datumbox: OK I'll restore `_split_alpha.

datumbox: Restoring this causes more [harm](https://github.com/pytorch/vision/actions/runs/3320465913/jobs/5486881222) than good. JIT is complaining that it's not longer a tuple, which means I would have to split into 2 lines and return as tuple to please it. I don't think it's worth this.

PR #6835, opened 2022-10-25
[prototype] Speed up Augment Transform Classes

datumbox:
This PR does a few optimizations on the `_augment.py` Transform classes.

cc @vfdev-5 @bjuncek @pmeier

### DIFF
--- a/torchvision/prototype/transforms/_augment.py
+++ b/torchvision/prototype/transforms/_augment.py
@@ -119,8 +119,7 @@ class _BaseMixupCutmix(_RandomApplyTransform):
     def _mixup_onehotlabel(self, inpt: features.OneHotLabel, lam: float) -> features.OneHotLabel:
         if inpt.ndim < 2:
             raise ValueError("Need a batch of one hot labels")
-        output = inpt.clone()
-        output = output.roll(1, 0).mul_(1.0 - lam).add_(output.mul_(lam))
+        output = inpt.roll(1, 0).mul_(1.0 - lam).add_(inpt.mul(lam))
         return features.OneHotLabel.wrap_like(inpt, output)
 
 
@@ -134,8 +133,7 @@ class RandomMixup(_BaseMixupCutmix):
             expected_ndim = 5 if isinstance(inpt, features.Video) else 4
             if inpt.ndim < expected_ndim:
                 raise ValueError("The transform expects a batched input")
-            output = inpt.clone()
-            output = output.roll(1, 0).mul_(1.0 - lam).add_(output.mul_(lam))
+            output = inpt.roll(1, 0).mul_(1.0 - lam).add_(inpt.mul(lam))
 
             if isinstance(inpt, (features.Image, features.Video)):
                 output = type(inpt).wrap_like(inpt, output)  # type: ignore[arg-type]
@@ -241,11 +239,12 @@ class SimpleCopyPaste(_RandomApplyTransform):
         if blending:
             paste_alpha_mask = F.gaussian_blur(paste_alpha_mask.unsqueeze(0), kernel_size=[5, 5], sigma=[2.0])
 
+        inverse_paste_alpha_mask = ~paste_alpha_mask
         # Copy-paste images:
-        image = (image * (~paste_alpha_mask)) + (paste_image * paste_alpha_mask)
+        image = (image * inverse_paste_alpha_mask).add_(paste_image * paste_alpha_mask)
 
         # Copy-paste masks:
-        masks = masks * (~paste_alpha_mask)
+        masks = masks * inverse_paste_alpha_mask
         non_all_zero_masks = masks.sum((-1, -2)) > 0
         masks = masks[non_all_zero_masks]
 
--- a/torchvision/prototype/transforms/_augment.py
+++ b/torchvision/prototype/transforms/_augment.py
@@ -239,9 +239,9 @@ class SimpleCopyPaste(_RandomApplyTransform):
         if blending:
             paste_alpha_mask = F.gaussian_blur(paste_alpha_mask.unsqueeze(0), kernel_size=[5, 5], sigma=[2.0])
 
-        inverse_paste_alpha_mask = ~paste_alpha_mask
+        inverse_paste_alpha_mask = paste_alpha_mask.logical_not()
         # Copy-paste images:
-        image = (image * inverse_paste_alpha_mask).add_(paste_image * paste_alpha_mask)
+        image = image.mul(inverse_paste_alpha_mask).add_(paste_image.mul(paste_alpha_mask))
 
         # Copy-paste masks:
         masks = masks * inverse_paste_alpha_mask
--- a/torchvision/prototype/transforms/_augment.py
+++ b/torchvision/prototype/transforms/_augment.py
@@ -40,7 +40,14 @@ class RandomErasing(_RandomApplyTransform):
             raise ValueError("Scale should be between 0 and 1")
         self.scale = scale
         self.ratio = ratio
-        self.value = value
+        if isinstance(value, (int, float)):
+            self.value = [value]
+        elif isinstance(value, str):
+            self.value = None
+        elif isinstance(value, tuple):
+            self.value = list(value)
+        else:
+            self.value = value
         self.inplace = inplace
 
         self._log_ratio = torch.log(torch.tensor(self.ratio))
@@ -48,16 +55,7 @@ class RandomErasing(_RandomApplyTransform):
     def _get_params(self, flat_inputs: List[Any]) -> Dict[str, Any]:
         img_c, img_h, img_w = query_chw(flat_inputs)
 
-        if isinstance(self.value, (int, float)):
-            value = [self.value]
-        elif isinstance(self.value, str):
-            value = None
-        elif isinstance(self.value, tuple):
-            value = list(self.value)
-        else:
-            value = self.value
-
-        if value is not None and not (len(value) in (1, img_c)):
+        if self.value is not None and not (len(self.value) in (1, img_c)):
             raise ValueError(
                 f"If value is a sequence, it should have either a single value or {img_c} (number of inpt channels)"
             )
@@ -79,10 +77,10 @@ class RandomErasing(_RandomApplyTransform):
             if not (h < img_h and w < img_w):
                 continue
 
-            if value is None:
+            if self.value is None:
                 v = torch.empty([img_c, h, w], dtype=torch.float32).normal_()
             else:
-                v = torch.tensor(value)[:, None, None]
+                v = torch.tensor(self.value)[:, None, None]
 
             i = torch.randint(0, img_h - h + 1, size=(1,)).item()
             j = torch.randint(0, img_w - w + 1, size=(1,)).item()


torchvision/prototype/transforms/_augment.py
```diff
@@ -40,24 +40,22 @@ def __init__(
...
         self.ratio = ratio
-        self.value = value
+        if isinstance(value, (int, float)):
+            self.value = [value]
+        elif isinstance(value, str):
+            self.value = None
+        elif isinstance(value, tuple):
+            self.value = list(value)
+        else:
+            self.value = value
```
datumbox, line 50: Moving `value` handling from run-time to constructor. This is more code-quality than perf.

torchvision/prototype/transforms/_augment.py
```diff
@@ -121,8 +119,7 @@ def _check_inputs(self, flat_inputs: List[Any]) -> None:
     def _mixup_onehotlabel(self, inpt: features.OneHotLabel, lam: float) -> features.OneHotLabel:
         if inpt.ndim < 2:
             raise ValueError("Need a batch of one hot labels")
-        output = inpt.clone()
-        output = output.roll(1, 0).mul_(1.0 - lam).add_(output.mul_(lam))
+        output = inpt.roll(1, 0).mul_(1.0 - lam).add_(inpt.mul(lam))
```
datumbox, line 122: As we saw in previous optimizations (for example #6821), removing clone in favour of moving the copy on the first multiplication has speed benefits. Below I benchmark the following two alternatives for various inputs (images and labels) to showcase this:

```python
def withclone(inpt, lam=0.5):
    output = inpt.clone()
    return output.roll(1, 0).mul_(1.0 - lam).add_(output.mul_(lam))


def withoutclone(inpt, lam=0.5):
    return inpt.roll(1, 0).mul_(1.0 - lam).add_(inpt.mul(lam))
```

Below we benchmark the 2 above methods alone:

```
[-------------- Mixing cpu torch.float32 --------------]
                         |  withclone   |  withoutclone 
1 threads: ---------------------------------------------
      (16, 3, 400, 400)  |    41200     |      15200    
      (16, 1000)         |       22     |         20    
6 threads: ---------------------------------------------
      (16, 3, 400, 400)  |    17700     |      15400    
      (16, 1000)         |       22     |         20    

[------------- Mixing cuda torch.float32 --------------]
                         |  withclone   |  withoutclone 
1 threads: ---------------------------------------------
      (16, 3, 400, 400)  |    274.9     |       226     
      (16, 1000)         |     33.2     |        26     
6 threads: ---------------------------------------------
      (16, 3, 400, 400)  |    275.6     |       226     
      (16, 1000)         |     33.0     |        26     

```

torchvision/prototype/transforms/_augment.py
```diff
@@ -243,11 +239,12 @@ def _copy_paste(
         if blending:
             paste_alpha_mask = F.gaussian_blur(paste_alpha_mask.unsqueeze(0), kernel_size=[5, 5], sigma=[2.0])
 
+        inverse_paste_alpha_mask = ~paste_alpha_mask
         # Copy-paste images:
-        image = (image * (~paste_alpha_mask)) + (paste_image * paste_alpha_mask)
+        image = (image * inverse_paste_alpha_mask).add_(paste_image * paste_alpha_mask)
```
datumbox, line 244: Simply cache `~paste_alpha_mask` to avoid repeated estimations and use of in-place addition on the intermediate results. I don't have benchmarks for this because I thought it's quite uncontroversial.

pmeier: Nit: I would prefer not to mix the regular operators like `*` and `+` with the flow interface like `.add_` and `.mul_`. Since the regular operators don't support inplace operations on the right side of a statement, this means I would prefer using the flow style whenever inplace operations are present

```py
image = image.mul(inverse_paste_alpha_mask).add_(paste_image.mul(paste_alpha_mask))
```

pmeier: Non-blocking: We could perform inplace operations on `paste_image` in case we needed to resize it above. Otherwise it is one of the inputs and thus we need to do out of place operations like it is done here.

torchvision/prototype/transforms/_augment.py
```diff
@@ -243,11 +239,12 @@ def _copy_paste(
         if blending:
             paste_alpha_mask = F.gaussian_blur(paste_alpha_mask.unsqueeze(0), kernel_size=[5, 5], sigma=[2.0])
 
+        inverse_paste_alpha_mask = ~paste_alpha_mask
```
pmeier, line 242: I just thought of another "evil hack" that avoids the out of place inversion here. There are two places where `paste_alpha_mask` or its inverse is used:

1. `image = (image * inverse_paste_alpha_mask).add_(paste_image * paste_alpha_mask)`
2. `masks = masks * inverse_paste_alpha_mask`

By reordering the first computation to

```py
image = (paste_image * paste_alpha_mask).add_(image * inverse_paste_alpha_mask)
```

we no longer need the `paste_alpha_mask` after `inverse_paste_alpha_mask` is used the first time. Thus, we can perform inplace operations on it:

```py
image = (paste_image * paste_alpha_mask).add_(image * paste_alpha_mask.logical_not_())
```

That turns the 2. statement into

```py
`masks = masks * paste_alpha_mask`
```

Although that looks wrong, it is actually correct since `paste_alpha_mask` at this point contains it inverse. 

Going for this needs at least an extensive comment since it is not obvious. Not sure if it is worth it.

pmeier (approved):

torchvision/prototype/transforms/_augment.py
datumbox: @pmeier I'm not 100% sure but I sense that this will end up having the same number of operations as before. You indeed do an inplace operation but at the end you still need to reverse the bool an write the results back to memory. I think what you propose will lead to less memory footprint. Still I think your proposal is worth trying. Do you want to bring it as an optimization on a PR after this one? You could refactor it a bit to split this into multiple lines and avoid doing one non-inplace op with the inplace op on `paste_alpha_mask` to make things clearer to the reader (that's debatable, maybe it's clear already?). You could literally measure only those 2 lines on the new PR and show-case if the proposal is faster. WDYT?

pmeier: You are right:

```py
def current(*, image, masks, paste_image, paste_alpha_mask, size1_ne_size2):
    inverse_paste_alpha_mask = paste_alpha_mask.logical_not()

    # Copy-paste images:
    image = image.mul(inverse_paste_alpha_mask).add_(paste_image.mul(paste_alpha_mask))
    # Copy-paste masks:
    masks = masks * inverse_paste_alpha_mask


def proposal(*, image, masks, paste_image, paste_alpha_mask, size1_ne_size2):
    if size1_ne_size2:
        tmp = paste_image.mul_(paste_alpha_mask)
    else:
        tmp = paste_image.mul(paste_alpha_mask)
    inverse_paste_alpha_mask = paste_alpha_mask.logical_not_()

    # Copy-paste images:
    image = tmp.add_(image.mul(inverse_paste_alpha_mask))
    # Copy-paste masks:
    masks = masks.mul(inverse_paste_alpha_mask)
```

<details><summary>Benchmark script</summary>
<p>

```py
import itertools

import torch

shapes = [
    (3, 512, 512),  # single image
    (5, 3, 512, 512),  # single video
]
dtypes = [torch.uint8, torch.float32]
devices = ["cpu", "cuda"]


def current(*, image, masks, paste_image, paste_alpha_mask, size1_ne_size2):
    inverse_paste_alpha_mask = paste_alpha_mask.logical_not()

    # Copy-paste images:
    image = image.mul(inverse_paste_alpha_mask).add_(paste_image.mul(paste_alpha_mask))
    # Copy-paste masks:
    masks = masks * inverse_paste_alpha_mask


def proposal(*, image, masks, paste_image, paste_alpha_mask, size1_ne_size2):
    if size1_ne_size2:
        tmp = paste_image.mul_(paste_alpha_mask)
    else:
        tmp = paste_image.mul(paste_alpha_mask)
    inverse_paste_alpha_mask = paste_alpha_mask.logical_not_()

    # Copy-paste images:
    image = tmp.add_(image.mul(inverse_paste_alpha_mask))
    # Copy-paste masks:
    masks = masks.mul(inverse_paste_alpha_mask)


timers = [
    benchmark.Timer(
        stmt="fn(image=image, masks=masks, paste_image=paste_image, paste_alpha_mask=paste_alpha_mask, size1_ne_size2=size1_ne_size2)",
        globals=dict(
            fn=fn,
            image=torch.testing.make_tensor(
                shape,
                dtype=dtype,
                device=device,
                low=0,
                high=1 if dtype.is_floating_point else torch.iinfo(dtype).max,
            ),
            masks=torch.testing.make_tensor(
                (4, *shape),
                dtype=torch.uint8,
                device=device,
                low=0,
                high=2,
            ),
            paste_image=torch.testing.make_tensor(
                shape,
                dtype=dtype,
                device=device,
                low=0,
                high=1 if dtype.is_floating_point else torch.iinfo(dtype).max,
            ),
            paste_alpha_mask=torch.testing.make_tensor(
                shape,
                dtype=torch.bool,
                device=device,
                low=0,
                high=2,
            ),
            size1_ne_size2=size1_ne_size2,
        ),
        sub_label=f"{shape!s:16} / {str(dtype).replace('torch.', ''):7} / {device:4} / size1_ne_size2={size1_ne_size2}",
        description=fn.__name__,
    )
    for fn, shape, dtype, device, size1_ne_size2 in itertools.product(
        [
            current,
            proposal,
        ],
        shapes,
        dtypes,
        devices,
        [False, True],
    )
]

measurements = [timer.blocked_autorange(min_run_time=10) for timer in timers]

comparison.print()

```

</p>
</details>

```
                                                                |  current  |  proposal
1 threads: ----------------------------------------------------------------------------
      (3, 512, 512)    / uint8   / cpu  / size1_ne_size2=False  |   1964.5  |   2155.9 
      (3, 512, 512)    / uint8   / cpu  / size1_ne_size2=True   |   1956.6  |   2169.5 
      (3, 512, 512)    / uint8   / cuda / size1_ne_size2=False  |    142.8  |    140.8 
      (3, 512, 512)    / uint8   / cuda / size1_ne_size2=True   |    143.4  |    139.6 
      (3, 512, 512)    / float32 / cpu  / size1_ne_size2=False  |   2053.3  |   2270.9 
      (3, 512, 512)    / float32 / cpu  / size1_ne_size2=True   |   2052.9  |   2217.7 
      (3, 512, 512)    / float32 / cuda / size1_ne_size2=False  |    196.9  |    197.0 
      (3, 512, 512)    / float32 / cuda / size1_ne_size2=True   |    196.9  |    196.7 
      (5, 3, 512, 512) / uint8   / cpu  / size1_ne_size2=False  |  10575.0  |  10932.7 
      (5, 3, 512, 512) / uint8   / cpu  / size1_ne_size2=True   |  10541.8  |  10887.1 
      (5, 3, 512, 512) / uint8   / cuda / size1_ne_size2=False  |    695.4  |    696.4 
      (5, 3, 512, 512) / uint8   / cuda / size1_ne_size2=True   |    695.2  |    697.6 
      (5, 3, 512, 512) / float32 / cpu  / size1_ne_size2=False  |  14170.3  |  14315.7 
      (5, 3, 512, 512) / float32 / cpu  / size1_ne_size2=True   |  14110.5  |  13735.6 
      (5, 3, 512, 512) / float32 / cuda / size1_ne_size2=False  |    946.1  |    949.2 
      (5, 3, 512, 512) / float32 / cuda / size1_ne_size2=True   |    948.9  |    951.2 

```

- It makes no difference on CUDA
- There is a slowdown on CPU
- Only for the combination `(5, 3, 512, 512) / float32 / cpu  / size1_ne_size2=True` there is a speed-up, of ~400µs. It is likely that the speed-up comes from the `size1_ne_size2=True` condition that I suggested in https://github.com/pytorch/vision/pull/6835#discussion_r1005298828.

I would leave it as is.

PR #6837, opened 2022-10-25
[prototype] Minor speed and nit optimizations on Transform Classes

datumbox:
Further speed optimizations, nits and code-quality changes:
- Switching from `torch.distributions.Uniform()` to `tensor.uniform_()` speeds up CUDA by 20%. It also aligns with the implementation on V1.
- Moves the input sanitisation from the runtime to construction time for as many params as possible.
- Contains a few non performance refactorings to avoid unnecessary casting and moving methods around.

cc @vfdev-5 @bjuncek @pmeier

### DIFF
--- a/torchvision/prototype/transforms/_auto_augment.py
+++ b/torchvision/prototype/transforms/_auto_augment.py
@@ -1,5 +1,5 @@
 import math
-from typing import Any, Callable, cast, Dict, List, Optional, Tuple, Type, Union
+from typing import Any, Callable, Dict, List, Optional, Tuple, Type, Union
 
 import PIL.Image
 import torch
@@ -166,9 +166,7 @@ class AutoAugment(_AutoAugmentBase):
         "Contrast": (lambda num_bins, height, width: torch.linspace(0.0, 0.9, num_bins), True),
         "Sharpness": (lambda num_bins, height, width: torch.linspace(0.0, 0.9, num_bins), True),
         "Posterize": (
-            lambda num_bins, height, width: cast(torch.Tensor, 8 - (torch.arange(num_bins) / ((num_bins - 1) / 4)))
-            .round()
-            .int(),
+            lambda num_bins, height, width: (8 - (torch.arange(num_bins) / ((num_bins - 1) / 4))).round().int(),
             False,
         ),
         "Solarize": (lambda num_bins, height, width: torch.linspace(255.0, 0.0, num_bins), False),
@@ -323,9 +321,7 @@ class RandAugment(_AutoAugmentBase):
         "Contrast": (lambda num_bins, height, width: torch.linspace(0.0, 0.9, num_bins), True),
         "Sharpness": (lambda num_bins, height, width: torch.linspace(0.0, 0.9, num_bins), True),
         "Posterize": (
-            lambda num_bins, height, width: cast(torch.Tensor, 8 - (torch.arange(num_bins) / ((num_bins - 1) / 4)))
-            .round()
-            .int(),
+            lambda num_bins, height, width: (8 - (torch.arange(num_bins) / ((num_bins - 1) / 4))).round().int(),
             False,
         ),
         "Solarize": (lambda num_bins, height, width: torch.linspace(255.0, 0.0, num_bins), False),
@@ -379,9 +375,7 @@ class TrivialAugmentWide(_AutoAugmentBase):
         "Contrast": (lambda num_bins, height, width: torch.linspace(0.0, 0.99, num_bins), True),
         "Sharpness": (lambda num_bins, height, width: torch.linspace(0.0, 0.99, num_bins), True),
         "Posterize": (
-            lambda num_bins, height, width: cast(torch.Tensor, 8 - (torch.arange(num_bins) / ((num_bins - 1) / 6)))
-            .round()
-            .int(),
+            lambda num_bins, height, width: (8 - (torch.arange(num_bins) / ((num_bins - 1) / 6))).round().int(),
             False,
         ),
         "Solarize": (lambda num_bins, height, width: torch.linspace(255.0, 0.0, num_bins), False),
@@ -426,9 +420,7 @@ class AugMix(_AutoAugmentBase):
         "TranslateY": (lambda num_bins, height, width: torch.linspace(0.0, height / 3.0, num_bins), True),
         "Rotate": (lambda num_bins, height, width: torch.linspace(0.0, 30.0, num_bins), True),
         "Posterize": (
-            lambda num_bins, height, width: cast(torch.Tensor, 4 - (torch.arange(num_bins) / ((num_bins - 1) / 4)))
-            .round()
-            .int(),
+            lambda num_bins, height, width: (4 - (torch.arange(num_bins) / ((num_bins - 1) / 4))).round().int(),
             False,
         ),
         "Solarize": (lambda num_bins, height, width: torch.linspace(255.0, 0.0, num_bins), False),
--- a/torchvision/prototype/transforms/_auto_augment.py
+++ b/torchvision/prototype/transforms/_auto_augment.py
@@ -1,5 +1,5 @@
 import math
-from typing import Any, Callable, cast, Dict, List, Optional, Tuple, Type, TypeVar, Union
+from typing import Any, Callable, cast, Dict, List, Optional, Tuple, Type, Union
 
 import PIL.Image
 import torch
@@ -10,9 +10,6 @@ from torchvision.prototype.transforms import AutoAugmentPolicy, functional as F,
 from torchvision.prototype.transforms.functional._meta import get_spatial_size
 
 from ._utils import _isinstance, _setup_fill_arg
-
-K = TypeVar("K")
-V = TypeVar("V")
 
 
 class _AutoAugmentBase(Transform):
@@ -26,7 +23,7 @@ class _AutoAugmentBase(Transform):
         self.interpolation = interpolation
         self.fill = _setup_fill_arg(fill)
 
-    def _get_random_item(self, dct: Dict[K, V]) -> Tuple[K, V]:
+    def _get_random_item(self, dct: Dict[str, Tuple[Callable, bool]]) -> Tuple[str, Tuple[Callable, bool]]:
         keys = tuple(dct.keys())
         key = keys[int(torch.randint(len(keys), ()))]
         return key, dct[key]
--- a/torchvision/prototype/transforms/_auto_augment.py
+++ b/torchvision/prototype/transforms/_auto_augment.py
@@ -71,10 +71,9 @@ class _AutoAugmentBase(Transform):
         transform_id: str,
         magnitude: float,
         interpolation: InterpolationMode,
-        fill: Dict[Type, features.FillType],
+        fill: Dict[Type, features.FillTypeJIT],
     ) -> Union[features.ImageType, features.VideoType]:
         fill_ = fill[type(image)]
-        fill_ = F._geometry._convert_fill_arg(fill_)
 
         if transform_id == "Identity":
             return image
--- a/torchvision/prototype/transforms/_auto_augment.py
+++ b/torchvision/prototype/transforms/_auto_augment.py
@@ -508,7 +508,7 @@ class AugMix(_AutoAugmentBase):
             mix.add_(
                 # The multiplication below could become in-place provided `aug is not batch and aug.is_floating_point()`
                 # Currently we can't do this because `aug` has to be `unint8` to support ops like `equalize`.
-                # TODO: change this once all ops in `F` support float inputs.
+                # TODO: change this once all ops in `F` support floats. https://github.com/pytorch/vision/issues/6840
                 combined_weights[:, i].reshape(batch_dims)
                 * aug
             )
--- a/torchvision/prototype/transforms/_auto_augment.py
+++ b/torchvision/prototype/transforms/_auto_augment.py
@@ -505,7 +505,13 @@ class AugMix(_AutoAugmentBase):
                 aug = self._apply_image_or_video_transform(
                     aug, transform_id, magnitude, interpolation=self.interpolation, fill=self.fill
                 )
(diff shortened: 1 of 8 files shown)


torchvision/prototype/transforms/_color.py
```diff
@@ -51,7 +51,7 @@ def _check_input(
 
     @staticmethod
     def _generate_value(left: float, right: float) -> float:
-        return float(torch.distributions.Uniform(left, right).sample())
+        return torch.empty(1).uniform_(left, right).item()
```
datumbox, line 54: Switching to this random generator we get a performance boost on GPU. Moreover this option is JIT-scriptable (if on the future we decide to add support) and doesn't require to constantly initialize a distribution object as before:
```
[--------- ColorJitter cpu torch.float32 ---------]
                     |   old random  |   new random
1 threads: ----------------------------------------
      (3, 400, 400)  |       17      |       17    
6 threads: ----------------------------------------
      (3, 400, 400)  |       21      |       21    

[--------- ColorJitter cuda torch.float32 --------]
                     |   old random  |   new random
1 threads: ----------------------------------------
      (3, 400, 400)  |      1090     |      883    
6 threads: ----------------------------------------
      (3, 400, 400)  |      1090     |      882    

```

torchvision/prototype/transforms/_utils.py
```diff
@@ -55,13 +55,30 @@ def _get_defaultdict(default: T) -> Dict[Any, T]:
     return defaultdict(functools.partial(_default_arg, default))
 
 
-def _setup_fill_arg(fill: Union[FillType, Dict[Type, FillType]]) -> Dict[Type, FillType]:
+def _convert_fill_arg(fill: features.FillType) -> features.FillTypeJIT:
```
datumbox, line 58: Moved from `_geometry.py`. Nothing to review.

torchvision/prototype/transforms/_utils.py
```diff
@@ -55,13 +55,30 @@ def _get_defaultdict(default: T) -> Dict[Any, T]:
...
+
+def _setup_fill_arg(fill: Union[FillType, Dict[Type, FillType]]) -> Dict[Type, FillTypeJIT]:
     _check_fill_arg(fill)
 
     if isinstance(fill, dict):
-        return fill
+        fill_copy = {}
+        for k, v in fill.items():
+            fill_copy[k] = _convert_fill_arg(v)
+        return fill_copy
```
datumbox, line 79: Let's convert the input dict/value into a `Dict[Type, FillTypeJIT]` dictionary on the constructor of objects. This way we won't have to do this on the runtime ever time. This is less of a speed optimization and more of code-quality as it allows us to remove a lot of dup code.

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -26,7 +23,7 @@ def __init__(
         self.interpolation = interpolation
         self.fill = _setup_fill_arg(fill)
 
-    def _get_random_item(self, dct: Dict[K, V]) -> Tuple[K, V]:
+    def _get_random_item(self, dct: Dict[str, Tuple[Callable, bool]]) -> Tuple[str, Tuple[Callable, bool]]:
```
datumbox, line 26: Removing unnecessary `TypeVar`

torchvision/prototype/transforms/_auto_augment.py
```diff
@@ -170,9 +166,7 @@ class AutoAugment(_AutoAugmentBase):
         "Contrast": (lambda num_bins, height, width: torch.linspace(0.0, 0.9, num_bins), True),
         "Sharpness": (lambda num_bins, height, width: torch.linspace(0.0, 0.9, num_bins), True),
         "Posterize": (
-            lambda num_bins, height, width: cast(torch.Tensor, 8 - (torch.arange(num_bins) / ((num_bins - 1) / 4)))
-            .round()
-            .int(),
+            lambda num_bins, height, width: (8 - (torch.arange(num_bins) / ((num_bins - 1) / 4))).round().int(),
```
datumbox, line 169: Removed unnecessary casting.

torchvision/prototype/transforms/_geometry.py
```diff
@@ -223,20 +223,16 @@ def __init__(
         _check_padding_arg(padding)
         _check_padding_mode_arg(padding_mode)
 
+        # This cast does Sequence[int] -> List[int] and is required to make mypy happy
+        if not isinstance(padding, int):
+            padding = list(padding)
```
datumbox, line 228: Move on construction time.

torchvision/prototype/transforms/_geometry.py
```diff
@@ -300,12 +295,11 @@ def __init__(
         self.center = center
 
     def _get_params(self, flat_inputs: List[Any]) -> Dict[str, Any]:
-        angle = float(torch.empty(1).uniform_(float(self.degrees[0]), float(self.degrees[1])).item())
+        angle = torch.empty(1).uniform_(self.degrees[0], self.degrees[1]).item()
```
datumbox, line 298: Remove unnecessary float convertions. The output is guaranteed to be float. Same for all occurrences below.

torchvision/prototype/transforms/_geometry.py
```diff
@@ -512,21 +503,23 @@ def _get_params(self, flat_inputs: List[Any]) -> Dict[str, Any]:
 
         half_height = height // 2
         half_width = width // 2
+        bound_height = int(distortion_scale * half_height) + 1
+        bound_width = int(distortion_scale * half_width) + 1
```
datumbox, line 507: Cleaning up the code a bit to avoid repeated estimations and unnecessary castings.

torchvision/prototype/transforms/_geometry.py
```diff
@@ -512,30 +503,31 @@ def _get_params(self, flat_inputs: List[Any]) -> Dict[str, Any]:
...
         ]
         botleft = [
-            int(torch.randint(0, int(distortion_scale * half_width) + 1, size=(1,)).item()),
-            int(torch.randint(height - int(distortion_scale * half_height) - 1, height, size=(1,)).item()),
+            torch.randint(0, bound_width, size=(1,)).item(),
+            torch.randint(height - bound_height, height, size=(1,)).item(),
         ]
         startpoints = [[0, 0], [width - 1, 0], [width - 1, height - 1], [0, height - 1]]
-        endpoints = [topleft, topright, botright, botleft]
+        endpoints: List[List[int]] = [topleft, topright, botright, botleft]  # type: ignore[list-item]
```
datumbox, line 525: For some reason mypy can't see that `endpoints` is a List of integer lists. That explains the original duplicate castings. I choose to ignore instead but happy to revert.

torchvision/prototype/transforms/_type_conversion.py
```diff
@@ -27,7 +27,7 @@ def _transform(self, inpt: features.Label, params: Dict[str, Any]) -> features.O
         num_categories = self.num_categories
         if num_categories == -1 and inpt.categories is not None:
             num_categories = len(inpt.categories)
-        output = one_hot(inpt, num_classes=num_categories)
+        output = one_hot(inpt.as_subclass(torch.Tensor), num_classes=num_categories)
```
datumbox, line 30: This is the last `as_subclass` call that we missed as far as I can see.

torchvision/prototype/transforms/_utils.py
```diff
@@ -80,7 +97,7 @@ def _check_padding_mode_arg(padding_mode: Literal["constant", "edge", "reflect",
 
 
 def query_bounding_box(flat_inputs: List[Any]) -> features.BoundingBox:
-    bounding_boxes = {inpt for inpt in flat_inputs if isinstance(inpt, features.BoundingBox)}
+    bounding_boxes = [inpt for inpt in flat_inputs if isinstance(inpt, features.BoundingBox)]
```
datumbox, line 100: I think we can use a list here instead of a set.

torchvision/prototype/transforms/_type_conversion.py
pmeier: Thanks!

torchvision/prototype/transforms/_type_conversion.py
```diff
@@ -13,7 +13,7 @@ class DecodeImage(Transform):
     _transformed_types = (features.EncodedImage,)
 
     def _transform(self, inpt: torch.Tensor, params: Dict[str, Any]) -> features.Image:
-        return cast(features.Image, F.decode_image_with_pil(inpt))
+        return F.decode_image_with_pil(inpt)  # type: ignore[no-any-return]
```
pmeier, line 16: This has to be here, because it seems 

https://github.com/pytorch/vision/blob/c84dbfad97251271a789b252a2a1a52c73f623ff/torchvision/prototype/transforms/functional/_type_conversion.py#L13-L14

doesn't "forward" the type annotations :roll_eyes:

torchvision/prototype/transforms/_utils.py
pmeier: Is this a perf optimization? Otherwise I would prefer the set since

1. it makes it more clear we are looking for duplicates
2. it is aligned with the other `query_*` function below.

Functionally, the only difference is that with a `set`, passing the same bounding box twice in the same sample would not be caught. This indeed seems like a user error, but if we raise there, we should probably check for duplicates everywhere. Not sure if we want to single out bounding boxes here.

datumbox: It's both perf and code correctness. What we want here is to ensure there is only 1 bbox in the input. If there were two, even if they were the same exact copy, our transforms that rely on `query_bounding_box` would end up working incorrectly. Effectively they would end up modifying the first copy but not the second. This is a different use-case from the rest of the `query_*` methods where we are able to handle multiple entries but just want to ensure same size across.

torchvision/prototype/transforms/_utils.py
```diff
@@ -33,13 +33,15 @@ def _setup_float_or_seq(arg: Union[float, Sequence[float]], name: str, req_size:
 
 
 def _check_fill_arg(fill: Union[FillType, Dict[Type, FillType]]) -> None:
-    if isinstance(fill, dict):
+    if type(fill) == dict:
+        # Do exact type check to avoid accepting default dicts from the user. DefaultDict values can be verified only
+        # at runtime not at construction type.
```
datumbox, line 38: Here I'm checking for exact simple dict. The reason we can't accept defaultdicts after this change is because there is no great way to verify at constructor time their default value. We can only check their non-default values. This means that users need to specify explicitly all types rather than doing something like `fill=defaultdict(lambda: 0, {features.Mask: 255}))`. IMO though this idiom is convenient, it can lead to surprises. This will align the behaviour with `ConvertDtype`, `TransposeDimensions` etc where we expect the user to explicitly define the values.

torchvision/prototype/transforms/_type_conversion.py
datumbox: It's fine, in all other places we took the decision to silence with ignore rather than cast.

torchvision/prototype/transforms/_utils.py
datumbox: As discussed offline, `query_chw` and `query_spatial_size` must use sets because what we are interested in is enforcing that all images have a single size. Multiple images/videos are allowed in these cases, we just enforce the size is the same everywhere.

On the other hand the `query_bounding_box` is used to extract the single bbox. The transforms that use the specific util, handle only 1 bbox and this is why we need a list. Note that this is identical to the `_flatten_and_extract_image_or_video` from AA where we also use a list.

So this is primarily a bug fix to restore the right semantics on the function and also has secondary positive performance implications.

torchvision/prototype/transforms/_utils.py
```diff
@@ -37,9 +37,12 @@ def _check_fill_arg(fill: Union[FillType, Dict[Type, FillType]]) -> None:
         for key, value in fill.items():
             # Check key for type
             _check_fill_arg(value)
+        if isinstance(fill, defaultdict) and callable(fill.default_factory):
+            default_value = fill.default_factory()
+            _check_fill_arg(default_value)
```
datumbox, line 42: @pmeier Removing the support of defaultdicts would mean an extremely verbose API for users. They would need to define not only types that make sense but also those that dont (such as Labels, HotLabels and BBoxes) even for padding calls. Here I propose an alternative mitigation. This is part 1 to ensure we check things properly during the constructor. This is an attempt to "bug fix"/extend the validation of the existing method for default dicts.

torchvision/prototype/transforms/_utils.py
```diff
@@ -55,13 +58,33 @@ def _get_defaultdict(default: T) -> Dict[Any, T]:
...
 
     if isinstance(fill, dict):
-        return fill
+        for k, v in fill.items():
+            fill[k] = _convert_fill_arg(v)
+        if isinstance(fill, defaultdict) and callable(fill.default_factory):
+            default_value = fill.default_factory()
+            sanitized_default = _convert_fill_arg(default_value)
+            fill.default_factory = functools.partial(_default_arg, sanitized_default)
+        return fill  # type: ignore[return-value]
```
datumbox, line 85: @pmeier Second mitigation. This sanitizes the input of all existing keys and patches the default_factory to have a sanitized value. If we don't like this mitigation, then I think we will have to revert the change of doing the verification on constructor time instead on runtime. Let me know your preference.

torchvision/prototype/transforms/_utils.py
pmeier: As discussed offline, let's move forward with the current implementation and refactor later if needed.

pmeier (approved):
Thanks Vasilis! LGTM if CI is green.

torchvision/prototype/transforms/_type_conversion.py
pmeier: Nono, I was just explaining why we need the ignore for future me that is looking confused at the blame why we introduced it in the first place.

PR #6838, opened 2022-10-26
reinstate _split_alpha

pmeier:
Follow-up to https://github.com/pytorch/vision/pull/6832#discussion_r1004448236. Instead of following the potential solution there, which I agree would not be worth it, this PR simply changes the annotation of `_split_alpha` from `Tuple[torch.Tensor, torch.Tensor]` to `List[torch.Tensor]`. Both `mypy` and JIT are happy with that.

@datumbox I brought this PR because you initially agreed to keep `_split_alpha` and apparently the issue above is the thing that tipped you in the other direction. Feel free to close this PR if there is anything else why you prefer the current solution over `_split_alpha`.

### DIFF
--- a/torchvision/prototype/transforms/functional/_meta.py
+++ b/torchvision/prototype/transforms/functional/_meta.py
@@ -183,6 +183,10 @@ def clamp_bounding_box(
     return convert_format_bounding_box(xyxy_boxes, BoundingBoxFormat.XYXY, format)
 
 
+def _split_alpha(image: torch.Tensor) -> List[torch.Tensor]:
+    return torch.tensor_split(image, indices=(-1,), dim=-3)
+
+
 def _strip_alpha(image: torch.Tensor) -> torch.Tensor:
     image, alpha = torch.tensor_split(image, indices=(-1,), dim=-3)
     if not torch.all(alpha == _FT._max_value(alpha.dtype)):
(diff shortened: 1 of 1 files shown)


torchvision/prototype/transforms/functional/_meta.py
```diff
@@ -183,6 +183,10 @@ def clamp_bounding_box(
     return convert_format_bounding_box(xyxy_boxes, BoundingBoxFormat.XYXY, format)
 
 
+def _split_alpha(image: torch.Tensor) -> List[torch.Tensor]:
+    return torch.tensor_split(image, indices=(-1,), dim=-3)
+
+
 def _strip_alpha(image: torch.Tensor) -> torch.Tensor:
     image, alpha = torch.tensor_split(image, indices=(-1,), dim=-3)
```
datumbox, line 191: I think you missed it here.

torchvision/prototype/transforms/functional/_meta.py
```diff
@@ -183,6 +183,10 @@ def clamp_bounding_box(
     return convert_format_bounding_box(xyxy_boxes, BoundingBoxFormat.XYXY, format)
 
 
+def _split_alpha(image: torch.Tensor) -> List[torch.Tensor]:
+    return torch.tensor_split(image, indices=(-1,), dim=-3)
```
datumbox, line 187: Frankly I reluctantly agreed with this to unblock the other PR. IMO there is very little value having a private method that does a one liner for an existing method on Core. I don't really mind this, I just think of this from the perspective of our practices. IMO the idiom of:
```
image, alpha = torch.tensor_split(image, indices=(-1,), dim=-3)
```
already sufficiently shows that we split the image and the alpha channel. The additional one liner, wrapper method doesn't add anything to the readability of the code. From that perspective, my preference is not to merge this, but I'm not going to insist if this is important to you.

PR #6841, opened 2022-10-26
Add Python 3.11

alihassanijr:
This PR adds Python 3.11 (cp311) to the [CircleCI regenerate script](https://github.com/alihassanijr/vision/blob/main/.circleci/regenerate.py), and the package helper.

I noticed torch 1.13 is built with cp311 wheels, but not torchvision.
I would assume this will create issues for users on 3.11, since pip will probably fall back to an older wheel.


datumbox:
@alihassanijr Thanks for the PR.

Adding new python versions is a bit more complex as it requires coordination with PyTorch Core and the Release engineering team. @malfet / @atalman Any thoughts on this? I would assume the addition of Python 3.11 will happen after the release?

alihassanijr:
Thanks for the information @datumbox .
The only reason I started this was to avoid inconsistency with the new torch release, which does support 3.11.
I don't have access to the circleci here, but I was able to build from source on 3.11 and run the tests locally, and can repeat that in manylinux containers with the supported cuda builds (and cpu) as well.
Don't know if that's helpful or not.

malfet:
Continuous Integration for TorchVision is build around Anaconda, and python-3.11 is yet not available here: https://anaconda.org/anaconda/python (and I know that it is available on CondaForge, but CondaForge builds are very unstable)

malfet (changes requested):
Sorry, I have not realized that PR targets `release/0.14` branch. As @datumbox correctly mentioned we should start with default branch and may be later will pick it for release

alihassanijr:
@malfet Thanks for the information. So you'd recommend we close this, and I reopen one with `main`?

alihassanijr:
Closing this in favor of #6959 . Rebasing this from 0.14 is just more trouble than it's worth compared to the very minor commits.

PR #6845, opened 2022-10-27
Added typing annotations to datasets/_optical_flow

frgfm:
Following up on https://github.com/pytorch/vision/issues/2025, this PR adds typing annotations in models/datasets/_optical_flow

Any feedback is welcome!

cc @datumbox @pmeier @oke-aditya


torchvision/datasets/_optical_flow.py
```diff
@@ -462,9 +476,9 @@ def _read_flo(file_name):
         return data.reshape(h, w, 2).transpose(2, 0, 1)
 
 
-def _read_16bits_png_with_flow_and_valid_mask(file_name):
+def _read_16bits_png_with_flow_and_valid_mask(file_name: Union[str, Path]) -> Tuple[np.ndarray, np.ndarray]:
 
-    flow_and_valid = _read_png_16(file_name).to(torch.float32)
+    flow_and_valid = _read_png_16(str(file_name)).to(torch.float32)
```
NicolasHug, line 481: Why is the `str` call needed here? If a Path was passed this would have failed already. Is the annotation actually correct?

frgfm: I had to balance two things:
- the upstream function `_read_png_16` whose typing I wanted to avoid changing (only accepting `str` even though I don't see any reason for it to fail on a `pathlib.Path`)
- this function is being called in `_read_flow` which does accept argument that can be both `pathlib.Path` or `str` if I'm not mistaking. So I don't think the code would fail before then if a Path slips in there

But I might be wrong :shrug:

torchvision/datasets/_optical_flow.py
```diff
@@ -139,7 +146,7 @@ def __init__(self, root, split="train", pass_name="clean", transforms=None):
                 if split == "train":
                     self._flow_list += sorted(glob(str(flow_root / scene / "*.flo")))
 
-    def __getitem__(self, index):
+    def __getitem__(self, index: int) -> Tuple:
```
pmeier, line 149: Could you make this more precise? Same for all other occurrences of plain `Tuple`

torchvision/datasets/_optical_flow.py
pmeier: I agree with @frgfm here. `file_name` could be a `Path` here so we need one mitigation if `_read_png_16` does not accept `Path`'s. Either convert here or extend it to also accept it.

pmeier (dismissed):
Thanks Francois, LGTM! One question though that is more directed at @NicolasHug: given that we are talking mostly about private methods here, is there a benefit of allowing `Union[str, pathlib.Path]` over a plain `Path` or `str`?

torchvision/datasets/_optical_flow.py
NicolasHug: Sorry I'm missing something. Could you please point out where `_read_16bits_png_with_flow_and_valid_mask()` is called on a `Path` object?

NicolasHug: From what I understand `_read_16bits_png_with_flow_and_valid_mask()` takes its input directly from here: 

https://github.com/pytorch/vision/blob/b522aca275cb20d688bef23f0155227d4a610af4/torchvision/datasets/_optical_flow.py#L56:L56

And `_flow_list` is a list of string according to the annotations above.

pmeier (approved):
LGTM, thanks Francois!

